Issue #36791, opened 2016-10-05
x509 signing policies should support more fine-grained access control


stenstad:
The current implementation of the "minions" parameter in x509 signing policies only support match.glob. This means that you either allow all minions to get a signed certificate (I.e. match on "_" or "_.domain") , or have your minion ID relate to the minions role (i.e. "www*.domain").

It would be great if it could support a list or maybe pcre?

```
x509_signing_policies:
  my_policy:
    minions: ["minion1.domain","minion2.domain","otherminion.otherdomain"]

  my_other_policy:
    minions: '((minion1|minion2).domain|otherminion.otherdomain)'
```

Ch3LL:
@stenstad agreed this would be nice to add this feature. I'll aprove as a feature request

PR #45378, opened 2018-01-10
modules.x509: add support for compound matching

ghost:
This addresses issue #36791 to allow fine-grained access control for x509 signing policies, by supporting `match.compound` in addition to `match.glob`. If the argument contains `@` it will use `match.compound`, otherwise it will use `match.glob`. I believe it is safe to assume there will be no minion IDs with `@`. Otherwise, a deeper change would be required, like adding a new parameter to select which match function is needed.

### Tests written?

No. The entire module `x509` has no tests.

### Commits signed with GPG?

No


garethgreenaway (approved):

rallytime:
Thank you @hgfischer! Congrats on your first Salt PR. :)

PR #45308, opened 2018-01-06
Fix `integration.modules.test_state` for Windows

twangboy:
### What does this PR do?
Fixes failing tests on Windows
- Fixes Windows-specific issues in `test_state`
- Fixes an issue with the `prereq_simple2` state file that had some
strange characters that were causing failure in windows (`<C2><A0>`)
- Adds a jinja to `watch_any` to handle Windows

Fixes an issue in `file.managed` where a command execution error was
being raised if the target file did not exist.

@rallytime Need to backport this to 2017.7. Should I do a separate PR?

### What issues does this PR fix or reference?
https://github.com/saltstack/salt-jenkins/issues/439

### Tests written?
Yes

### Commits signed with GPG?
Yes

PR #45380, opened 2018-01-10
Backport changes from #45308

twangboy:
### What does this PR do?
Backports changes from #45308 to 2017.7 branch.

### Tests written?
Yes

### Commits signed with GPG?
Yes


salt/states/file.py
```diff
@@ -2332,9 +2332,17 @@ def managed(name,
                 )
 
                 if salt.utils.is_windows():
-                    ret = __salt__['file.check_perms'](
-                        name, ret, win_owner, win_perms, win_deny_perms,
-                        win_inheritance)
+                    try:
+                        ret = __salt__['file.check_perms'](
+                            path=name,
```
rallytime: @twangboy I don't think this way of calling the options in the function will work on 2017.7: https://github.com/saltstack/salt/blob/2017.7/salt/modules/file.py#L4180

rallytime (changes requested):

salt/states/file.py
twangboy: This won't actually call the `check_perms` function in `salt/modules/file.py` as it has been namespaced in `salt/modules/win_file.py`

rallytime: Ohhh OK. thank you. I was thinking that might be the case but I wasn't sure.

rallytime (approved):

PR #45381, opened 2018-01-10
fix module.run docs

### LINKED ISSUE #43130: module.run documentation issues
According to https://docs.saltstack.com/en/latest/ref/states/all/salt.states.module.html#module-salt.states.module the new `module.run` syntax is:
```
fetch_out_of_band:
  module.run:
    git.fetch:
      - cwd: /path/to/my/repo
      - user: myuser
      - opts: '--all'
```

This example is incorrect. It should read:
```
fetch_out_of_band:
  module.run:
    - git.fetch:
      - cwd: /path/to/my/repo
      - user: myuser
      - opts: '--all'
```

All the other examples on that page also have the same error, which I found quite confusing.

That entire section doesn't read very well, as the old and new syntax formats are mixed together. It would be clearer if it was broken into sections with separate headings such as "The new way" and "The old way (supported until Oxygen)". It would also allow anyone new to `module.run` to simply skip the section on the old format.

Finally, I'd like to see a more complicated example. eg.
```
Remove appserver server from HAProxy host:
  module.run:
    - publish.publish:
      - tgt: 'rproxy*'
      - fun: haproxy_cmds.disable_server
      - arg: >-
          [
          'server=appserver',
          'backend=bk_appservers'
          ]
    - require_in:
      - git: Deploy code
    - onchanges_in:
      - module: Add appserver server to HAProxy host
```
I tried experimenting with a YAML list for `arg`, but it didn't work. This might save someone time.

### Setup
N/A

### Steps to Reproduce Issue
N/A

### Versions Report
Applicable to the documentation in `2017.7.0+`.

gtmanfred:
### What does this PR do?
Make sure the new version of module.run has the correct format

Fixes #43130

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/states/module.py
+++ b/salt/states/module.py
@@ -24,7 +24,7 @@ for this the :mod:`module.wait <salt.states.module.wait>` state can be used:
 
     fetch_out_of_band:
       module.run:
-        git.fetch:
+        - git.fetch:
           - cwd: /path/to/my/repo
           - user: myuser
           - opts: '--all'
@@ -35,7 +35,7 @@ Another example:
 
     mine.send:
       module.run:
-        network.ip_addrs:
+        - network.ip_addrs:
           - interface: eth0
 
 And more complex example:
@@ -44,7 +44,7 @@ And more complex example:
 
     eventsviewer:
       module.run:
-        task.create_task:
+        - task.create_task:
           - name: events-viewer
           - user_name: System
           - action_type: Execute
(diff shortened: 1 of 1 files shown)


rallytime (changes requested):
Why are there kitchen and Gemfile changes in a PR for a doc update?

gtmanfred:
woops, because I am bad at Git.

Fixed.  Thanks,
Daniel

rallytime (approved):

PR #45382, opened 2018-01-10
Skip flaky test on 2017.7 branch

terminalmage:
The behavior of the set union used to combine ext_nodes and top file
matches is not consistent on Python 3 (well, 3.5 at least), and this
produces flaky test results.

Since we do not change things like the state compiler in point
releases, the method used to combine the ext_nodes and top file matches
has been changed for Oxygen and is now fully deterministic. This test
will therefore be skipped for the remainder of the 2017.7 release cycle
on Python 3.

Fixes https://github.com/saltstack/salt-jenkins/issues/686.

### DIFF
--- a/tests/integration/ssh/test_state.py
+++ b/tests/integration/ssh/test_state.py
@@ -9,6 +9,7 @@ import time
 # Import Salt Testing Libs
 from tests.support.case import SSHCase
 from tests.support.paths import TMP
+from tests.support.unit import skipIf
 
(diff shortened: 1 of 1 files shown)


rallytime (approved):

terminalmage:
integration.modules.test_state.StateModuleTest.test_exclude is failing on PY3, I've fixed it in https://github.com/saltstack/salt/pull/45388.

PR #45384, opened 2018-01-11
Instruct systemd to always restart services on failure

prymitive:
If salt-minion or any other service crashes it should be restarted by systemd

### What does this PR do?

Adds `Restart=on-failure` option to all systemd unit files. 

It allows systemd to restart all salt services if they experience a crash or any other even resulting in marking service as failed (like timeout on stopping the service).

### Tests written?

No

### Commits signed with GPG?

Yes

### DIFF
--- a/pkg/salt-api.service
+++ b/pkg/salt-api.service
@@ -9,6 +9,7 @@ NotifyAccess=all
 LimitNOFILE=8192
 ExecStart=/usr/bin/salt-api
 TimeoutStopSec=3
+Restart=on-failure
 
 [Install]
 WantedBy=multi-user.target
--- a/pkg/salt-api.service
+++ b/pkg/salt-api.service
@@ -10,6 +10,7 @@ LimitNOFILE=8192
 ExecStart=/usr/bin/salt-api
 TimeoutStopSec=3
 Restart=on-failure
+RestartSec=1m
 
 [Install]
 WantedBy=multi-user.target
--- a/pkg/salt-master.service
+++ b/pkg/salt-master.service
@@ -9,6 +9,7 @@ Type=notify
 NotifyAccess=all
 ExecStart=/usr/bin/salt-master
 Restart=on-failure
+RestartSec=1m
 
 [Install]
 WantedBy=multi-user.target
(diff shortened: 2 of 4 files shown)


gtmanfred (commented):
I am not certain this is a good idea.  The daemon may be failing to start for a reason, because of a bad config or something else.

I would think it would be better to include documentation to instruct people to put 
```
[Service]
Restart=on-failure
```

in `/etc/systemd/system/salt-minion.service.d/restart.conf` instead of including it in the default service unit.  That way they can enable it once they have salt configured to their liking.  And they don't even have to restart the salt services for it to take effect, just a file.managed and a `daemon-reload`

prymitive:
What problem will trying to restart the service cause in case of a bad config?

gtmanfred:
It will constantly restart, and never stop, but also never start, and will only use up cpu that it doesn't have to be because the error log will already have the information needed.

prymitive:
Added `RestartSec=1m`, does this address your concern?
Also note that `salt-proxy@.service` already has 
```
Restart=on-failure
RestartSec=5s
```

terminalmage (commented):
I second @gtmanfred's idea of adding something to the documentation to override the unit file we distribute if they want to do this.

rallytime:
@prymitive Would you be willing to change this to a documentation update instead?

prymitive:
`pkg/salt-proxy@.service` contains
```
Restart=on-failure
RestartSec=5s
```
Why is it a good idea to have it there but not anywhere else?
I think that salt should have sensible defaults and I think that restarting a critical service when it crashes is a sensible default.

bmandrade:
> It will constantly restart, and never stop, but also never start, and will only use up cpu that it doesn't have to be because the error log will already have the information needed.

This concern might be tackled by adding `StartLimitBurst=` and `StartLimitInterval=`, to `[unit]` to rate limiting the number of retries/restart the service is allowed before systemd blocks it from retrying a restart anymore.

Maybe a `StartLimitBurst` of 5, with a high interval, might be a good alternative.

Sorry to comment on a closed issue, but I end up going against the same problem and found this solution to avoid continuous retries without ever stopping.

gtmanfred:
That has been in since 2012, we should totally be able to use that for all the current supported systemd versions https://github.com/systemd/systemd/commit/4b9397474f2c80957fd1cedf885061052333fb25

Would you mind opening a PR to add that and pinging @dmurphy18 

Thanks!

PR #45385, opened 2018-01-11
Document the utils refactor

mirceaulinic:
Adding some notes regarding the utils functions refactor in the Oxygen release notes. This will be helpful so the users are aware of this change in advance (pretty sure many don't know yet), and encourage to start migrating the extension modules to use the new imports.


rallytime (approved):
Thank you so much @mirceaulinic - that will be very helpful indeed. :)

PR #45403, opened 2018-01-11
Fall back to utf-8 when to_str/to_unicode/to_bytes fail to encode/decode

terminalmage:
This fixes tracebacks in cases where the default encoding is not unicode.

### DIFF
--- a/salt/utils/stringutils.py
+++ b/salt/utils/stringutils.py
@@ -35,7 +35,14 @@ def to_bytes(s, encoding=None):
         if isinstance(s, bytearray):
             return bytes(s)
         if isinstance(s, six.string_types):
-            return s.encode(encoding or __salt_system_encoding__)
+            if encoding:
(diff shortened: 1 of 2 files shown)


twangboy (approved):

twangboy:
Fixes some failing tests on Windows

damon-atkins:
It should fail if it can not output the correct encoding.     Some wants a Greek file on the system, and then its filled with UTF-8 without any error? 

Maybe if its not None it should be more strict.

Maybe it should also log when using the fall back.

terminalmage:
@damon-atkins It fails when you _specify_ the encoding and it can't encode. If you read the code, this only falls back to-utf-8 if it fails to encode/decode to the `__salt_system_encoding__`.

rallytime (approved):

damon-atkins:
Yep.  Need to look at code in the morning, rather than late at night.

PR #45405, opened 2018-01-12
[oxygen] Updating various utils and related tests for Unicode goodness.

garethgreenaway:
### What does this PR do?
Updating various utils and related tests for Unicode goodness

### Tests written?
Yes

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/utils/error.py
+++ b/salt/utils/error.py
@@ -3,7 +3,7 @@
 Utilities to enable exception reraising across the master commands
 
 '''
-from __future__ import absolute_import
+from __future__ import absolute_import, unicode_literals
 
 # Import python libs
 try:
--- a/salt/utils/error.py
+++ b/salt/utils/error.py
@@ -3,7 +3,7 @@
 Utilities to enable exception reraising across the master commands
 
 '''
-from __future__ import absolute_import, unicode_literals
+from __future__ import absolute_import, unicode_literals, print_function
 
 # Import python libs
 try:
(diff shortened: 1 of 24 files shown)


rallytime (approved):

salt/utils/error.py
```diff
@@ -3,7 +3,7 @@
 Utilities to enable exception reraising across the master commands
 
 '''
-from __future__ import absolute_import
+from __future__ import absolute_import, unicode_literals
```
terminalmage: We should have `print_function` imported as well in all of these files (see [here](https://gist.github.com/terminalmage/7440c8d6e46ed4ee2f69c834d5363512#print_function-and-unicode_literals)).

salt/utils/functools.py
```diff
@@ -45,7 +46,7 @@ def alias_function(fun, name, doc=None):
     '''
     alias_fun = types.FunctionType(fun.__code__,
                                    fun.__globals__,
-                                   str(name),  # future lint: disable=blacklisted-function
+                                   salt.utils.stringutils.to_str(name),  # future lint: disable=blacklisted-function
```
terminalmage: `str()` is actually what we want here, `and in fact salt.utils.stringutils.to_str()` will raise an exception if you feed it value that isn't one of the string types.

terminalmage (changes requested):

terminalmage (approved):

PR #45408, opened 2018-01-12
salt/runners/manage.py: Fix case where _ping returns empty dict inste…

github-abcde:
…ad of tuple.

### What does this PR do?
Fix the case where manage.status is assigned a non-existing target, which causes a stacktrace.

None that I'm aware of

### Previous Behavior
`salt-run manage.status tgt="missing"` results in a stacktrace dump because _ping returns an empty dictionary.

### New Behavior
`salt-run manage.status tgt="missing"` results in the following dict: {'down': [], 'up': []}

### Tests written?
No

### Commits signed with GPG?
Yes


garethgreenaway (approved):

rallytime:
Thank you @github-abcde!

PR #45411, opened 2018-01-12
Cmd binary stdin

github-abcde:
### What does this PR do?
It adds a boolean option (stdin_binary) to cmdmod._run (which is passed to salt.utils.timed_subprocess.TimedProc) which will result in TimedProc not mangling stdin.

None that I'm aware of.

### Previous Behavior
salt.utils.timed_subprocess.TimedProc always replaces '\\n' by '\n' in the stdin stream.

### New Behavior
salt.utils.timed_subprocess.TimedProc does not change stdin when kwargs['stdin_binary'] is coerces to True.

### Tests written?
No

### Commits signed with GPG?
Yes (partial)


terminalmage (changes requested):
I don't really like `stdin_binary`, since there's nothing particularly "binary" about the stdin when newlines aren't being converted. How about `stdin_raw_newlines` (still with a default value of `False` in timed_subprocess.py)?

@rallytime what do you think?

github-abcde:
The reason I chose this name is because this change was inspired by a change in modules/archive.py:tar to support piping data from stdin and/or to stdout (to be PR'd after this is accepted). Though I wouldn't mind changing the name to something else, I do think it's better to specify what *does* happen instead of what *doesn't*.
As such, one option could be to rename the variable to `stdin_fix_newlines` with a default to `True` (and saving us a negation on the side).

rallytime:
@github-abcde I think that sounds good. Any other opinions on that @terminalmage?

terminalmage (commented):
Having `fix` in an argument name is a bad idea, in my opinion. What does `fix` even mean in this context? It's just not intuitive. At least with `stdin_raw_newlines`, it makes sense that `\n` in the stdin will be interpreted as the equivalent of a raw `\n` (i.e. `\\n`).

@rallytime thoughts?

github-abcde:
The general thought behind the variable is (at least, for me) to convey the messsage "Do not alter! stdin is to be handled as is." If you have any suggestions as to a name which better conveys this, (from both the viewpoint of `modules/cmdmod.py:run` as well as `utils/timed_subprocess.py`), please let me know.

github-abcde:
@rallytime Though 'handle' is also a bit unspecific (like 'fix') about what is going to happen with regard to newlines in stdin, it does specify that some processing is going to take place. I could add a bit of documentation in the function docstring to clear up what exactly `handle_stdin_newlines` will do.
Using 'handle' instead of 'fix' avoids the implication that something is wrong (and needs to be fixed).

rallytime:
> Using 'handle' instead of 'fix' avoids the implication that something is wrong (and needs to be fixed).

@github-abcde Yeah, that's basically what I was going for there. I agree it's not perfect, but I like it better than `fix` and it's the best one I could come up with. 😛 Documenting that would be ++.

What do you think @terminalmage?

terminalmage (commented):
I like it _less_ than I like I do `cmd_binary_stdin`, to be honest. Actual newlines can be passed on the CLI, by starting a quoted string and hitting `Enter` before closing the quotes. So not only is "handle" not very descriptive (handle how?), there's this distinction between actual newlines and a raw `\n` which it leaves confusion about which of them is being "handled".

github-abcde:
In that perspective, a more proper variablename would be `stdin_unescape_escaped_newlines`, since that's exactly what it triggers. It doesn't do anything regarding actual newlines. (This assumes that newlines are always 'escaped' as `\\n` though.)
Except that makes it a bit long.

damon-atkins:
This breaks the code where kwargs is passed along from one function to sub-function. The new_kwargs prevents this from happening.

github-abcde:
@damon-atkins How do you figure that?
Previously, `kwargs` would be assigned a new dict, thereby separating it from the kwargs passed to `_run`, however, this creates ambiguity for all code following that assignment, because it's not immediately clear that you're dealing with this new kwargs-dict instead of the kwargs passed to `_run` (which are, from that point on, inaccessible). That's why I introduced `new_kwargs`.

damon-atkins:
Your right `kwargs` passed around a bit but looks like most of the time something eventual does not pass it on any further.

rallytime:
Let's just go with `stdin_raw_newlines` on this. I think that one is better than the other suggestions here. 

@github-abcde It looks like there's a merge conflict on this now. Can you rebase and fix that up? Thanks for your work here!

github-abcde:
Ok, will do.

damon-atkins:
There is output_encoding which could be set to magic name of 'binary'

timed_subprocess contains are we doing the **encoding twice** once in timed_process and once in cmdmod._run
```
        elif self.stdin is not None:
            if not self.stdin_raw_newlines:
                # Translate a newline submitted as '\n' on the CLI to an actual
                # newline character.
                self.stdin = self.stdin.replace('\\n', '\n').encode(__salt_system_encoding__)
            kwargs['stdin'] = subprocess.PIPE
```

github-abcde:
@damon-atkins Those issues are outside the scope of this PR. You are welcome to submit a separate PR to fix the issues you have identified.

damon-atkins:
@github-abcde  sorry thought you added `self.stdin_raw_newlines`

github-abcde:
@damon-atkins As you can see in the [files changed]-tab, the double encode is already present in the original code at https://github.com/saltstack/salt/pull/45411/files#diff-12242d794e2d6ae105582c95f7a9fbd3L34 and is left there in my change at https://github.com/saltstack/salt/pull/45411/files#diff-12242d794e2d6ae105582c95f7a9fbd3R36

rallytime:
@github-abcde Thanks for updating that. Can you also add some docs about this new option?

salt/modules/cmdmod.py
```diff
@@ -1027,6 +1031,12 @@ def run(cmd,
 
       .. versionadded:: Fluorine
 
+    :param bool stdin_raw_newlines : False
+        Normally, newlines present in ``stdin`` as ``\\n`` will be 'unescaped',
+        i.e. replaced with a ``\n``. Set this parameter to ``True`` to leave
+        the newlines as-is. This should be used when you are supplying data
+        using ``stdin`` that should not be modified.
+
```
terminalmage: This needs a `.. versionadded:: Fluorine` with one line of separation from the above documentation. Additionally, we need the same documentation in _every_ function which invokes `_run()`.

terminalmage (changes requested):

salt/modules/cmdmod.py
github-abcde: Done

terminalmage (approved):

salt/modules/cmdmod.py
terminalmage: Thanks!

Issue #45056, opened 2017-12-17
Spurious SSDP errors on startup


terminalmage:
master.py is doing a boolean check on `self.opts['discovery]`, and assuming discovery is enabled if it resolves to `True`. But because there is a default value in `salt/config/__init__.py`, this option is therefore assumed to be enabled out-of-the-box. This doesn't seem correct, the description in the release notes mentions nothing about this feature being enabled by default, and doing so would add a hard dependency in Python 2, which I don't think was intended.

It looks like the best course of action is to set `discovery` to `False` so that it is not enabled by default.

Also, the types in [this block of code](https://github.com/saltstack/salt/blob/990c628/salt/config/__init__.py#L1165-L1182) can be removed, since `_validate_opts()` doesn't do type checking on sub-keys of a dict, and therefore these options are never going to be checked.

CC: @isbm @cachedout @thatch45

isbm:
@terminalmage Hi!

Right. Let's see what we can do here:

- While it is a hard dependency, the library is quite old, used widely in OpenStack and so is packaged quite everywhere (pip, rpms, dpks etc). [We did that for sure](https://build.opensuse.org/package/show/openSUSE:Leap:42.3/python-trollius) long time ago, and if [it is in Debian stable](https://packages.debian.org/stretch/python-trollius), then it is everywhere. :wink:  So I am thinking why not just include it until EOL of Python2 by default? Python 3 don't need it (as `trollius` is basically a port of `asyncio`). Then we can throw it away soon anyway if we drop Python2 entirely. Additionally, enabling `discovery: True` and then seeing "ERROR: foobar is not installed" is also kind of... meh... I have no much good strategy here to offer other than indeed `Requires: python-trollius` in a package.

- The default topic. I thought I mentioned that it is by default, seems forgot it... Well, this is a question. While now it requires clean networks (or you will give the root to a malicious Master), it would be nice to have this by clean default. This is very useful if you have your (usually isolated) network of bunch of VMs of, say, cloud, and you don't want to go over each node and say which is IP/host of the Master but just batch-bootstrap them at one go (as long as the image has a Minion, started by default).

- The type-checking. Yes, it doesn't checks this _yet_. As we agreed with @cachedout a bit earlier that this is subject to change anytime soon (as it is not a major feature that even needs to be mentioned in the press release other than "now it fail for sure if you configure something completely unexpected"). So I would happily implement type check and the whole config integrity, instead of removing this.

[...]

PR #45412, opened 2018-01-12
Isbm ssdp spurious errors fix

isbm:
### What does this PR do?

Fixes issue https://github.com/saltstack/salt/issues/45056

@terminalmage !

### DIFF
--- a/doc/topics/releases/oxygen.rst
+++ b/doc/topics/releases/oxygen.rst
@@ -294,13 +294,13 @@ as follows:
 
 **Master configuration**
 
-To simply enable it by default, accepting all kind of minions, is enough to set it to True:
+To simply enable it by default, accepting all kinds of minions, it is enough to set it to True:
 
 .. code-block:: yaml
 
        discovery: true
 
-Sub-option `mapping` allows autodiscovery help Minions to find a proper Master for them. The mapping contains
+Sub-option `mapping` allows autodiscovery to help Minions to find a proper Master for them. The mapping contains
 just an arbitrary set of key/value strings that would match either any of them or all of them. Default is empty.
 
 Example:
@@ -323,7 +323,7 @@ Additionally to the identical configuration as on a Master, there are two more o
 
 - ``attempts`` will set how many "ping" broadcast requests should be sent to the network, waiting for any Master
    response. Each attempt takes a couple of seconds, so raising it to the higher value will
-   result to a slower startup of the Minion. Generally, on a properly configured networks, the autodiscovery
+   result to a slower startup of the Minion. Generally, on a properly configured network, the autodiscovery is
    supposed to be all set right after the first attempt. Default is set to 3 attempts.
 - ``match`` can have values ``all`` or ``any``. Match is always coming together with ``mapping``, and essentially means
   "match any/all of mapping values to the target". Default is set to ``any``.
--- a/doc/topics/releases/oxygen.rst
+++ b/doc/topics/releases/oxygen.rst
@@ -276,7 +276,8 @@ The new grains added are:
 Salt Minion Auto-discovery
 ------------------------
 
-Salt Minion now no longer need to be configured against a specific DNS name or IP address of a Master.
+Using auto-discovery, the Salt Minion now no longer needs to be configured
+against a specific DNS name or IP address of a Master.
 
 For this feature Salt Master now requires port 4520 for UDP broadcast packets to be opened
 and the Salt Minion be able to send UDP packets to the same port.
@@ -290,19 +291,20 @@ By default, automatic discovery is disabled.
    Due to the current limitations that will be changing in a future, before you turn on auto-discovery,
    make sure your network is secured and trusted.
 
-Auto-discovery is configured on Master and Minion. Both of them are configured via ``discovery`` option
+Auto-discovery is configured on Master and Minion. Both of them are configured via the ``discovery`` option
 as follows:
 
 **Master configuration**
 
-To simply enable it by default, accepting all kinds of minions, it is enough to set it to True:
+To use the default configuration, which accepts any minion, simply set ``discovery`` to True:
 
 .. code-block:: yaml
 
        discovery: true
 
-Sub-option `mapping` allows auto-discovery to help Minions to find a proper Master for them. The mapping contains
-just an arbitrary set of key/value strings that would match either any of them or all of them. Default is empty.
+A sub-option called `mapping` allows auto-discovery to help find the proper
+Master. The mapping contains an arbitrary set of key/value pairs, which the
+Minion configuration can target. By default, no mappings are set.
 
 Example:
 
@@ -330,13 +332,17 @@ setting a ``port`` option under the Master's ``discovery`` configuration:
 
 In addition to the ``mapping`` and ``port`` options, the following additional options are available to Minions:
 
-- ``attempts`` will set how many "ping" broadcast requests should be sent to the network, waiting for any Master
-   response. Each attempt takes a couple of seconds, so raising it to the higher value will
-   result to a slower startup of the Minion. Generally, on a properly configured network, the autodiscovery is
-   supposed to be all set right after the first attempt. Default is set to 3 attempts.
-- ``match`` can have values ``all`` or ``any``. Match is always coming together with ``mapping``, and essentially means
-  "match any/all of mapping values to the target". Default is set to ``any``.
-- ``pause`` is an interval in seconds between attempts. Default is set to five (5) seconds.
+- ``attempts`` - This option specifies how many broadcast requests should be
+  sent to the network, waiting for any Master response. Each attempt takes a
+  couple of seconds, so raising this value may result in a slower Minion
+  startup. Note that, on a properly-configured network, autodiscovery should
+  succeed on the first attempt. By default, this value is set to ``3``.
+- ``match`` - This option can be set to either ``all`` or ``any``, and it
+  determines how the values configured in ``mapping`` are matched. If set to
+  ``all``, then all of the key/value pairs in the Minion's ``mapping`` must
+  match a given Master. If set to ``any`` (the default), then any match to a
+  key/value mapping will constitute a match.
+- ``pause`` - The interval in seconds between attempts (default: 5).
 
 Connection to a type instead of DNS
 ===================================
--- a/doc/topics/releases/oxygen.rst
+++ b/doc/topics/releases/oxygen.rst
@@ -273,10 +273,10 @@ The new grains added are:
 * ``iscsi_iqn``: Show the iSCSI IQN name for a host
 * ``swap_total``: Show the configured swap_total for Linux, \*BSD, OS X and Solaris/SunOS
 
-Salt Minion Autodiscovery
+Salt Minion Auto-discovery
 ------------------------
 
-Salt Minion now no longer need to be configured against a specifig DNS name or IP address of a Master.
+Salt Minion now no longer need to be configured against a specific DNS name or IP address of a Master.
 
 For this feature Salt Master now requires port 4520 for UDP broadcast packets to be opened
 and the Salt Minion be able to send UDP packets to the same port.
@@ -286,10 +286,11 @@ Configuration
 
 By default, automatic discovery is disabled.
 
-Warning: due to the current limitations that will be changing in a future, before you turn on autodiscovery,
-make sure your network is secured and trusted. See the limitations below.
+..warning::
+   Due to the current limitations that will be changing in a future, before you turn on auto-discovery,
+   make sure your network is secured and trusted.
 
-Autodiscovery is configured on Master and Minion. Both of them are configured via ``discovery`` option
+Auto-discovery is configured on Master and Minion. Both of them are configured via ``discovery`` option
 as follows:
 
 **Master configuration**
@@ -300,7 +301,7 @@ To simply enable it by default, accepting all kinds of minions, it is enough to 
 
        discovery: true
 
-Sub-option `mapping` allows autodiscovery to help Minions to find a proper Master for them. The mapping contains
+Sub-option `mapping` allows auto-discovery to help Minions to find a proper Master for them. The mapping contains
 just an arbitrary set of key/value strings that would match either any of them or all of them. Default is empty.
 
 Example:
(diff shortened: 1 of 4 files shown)


doc/topics/releases/oxygen.rst
```diff
@@ -281,6 +281,54 @@ Salt Minion now no longer need to be configured against a specifig DNS name or I
...
+
+Warning: due to the current limitations that will be changing in a future, before you turn on autodiscovery,
+make sure your network is secured and trusted. See the limitations below.
+
+Autodiscovery is configured on Master and Minion. Both of them are configured via ``discovery`` option
+as follows:
+
+**Master configuration**
+
+To simply enable it by default, accepting all kind of minions, is enough to set it to True:
```
rallytime: `kind` --> `kinds`, and `is enough` --> `it is enough`.

doc/topics/releases/oxygen.rst
```diff
@@ -281,6 +281,54 @@ Salt Minion now no longer need to be configured against a specifig DNS name or I
...
+It is also possible to change default ``4520`` port, by setting another integer value to the ``port`` option within
+the discovery system. However, this should be also pre-configured on the minions as well.
+
+**Minion configuration**
+
+Additionally to the identical configuration as on a Master, there are two more options available on a Minion:
+
+- ``attempts`` will set how many "ping" broadcast requests should be sent to the network, waiting for any Master
+   response. Each attempt takes a couple of seconds, so raising it to the higher value will
+   result to a slower startup of the Minion. Generally, on a properly configured networks, the autodiscovery
```
rallytime: `networks` --> `network` and `the autodiscovery` --> `the autodiscovery is`

doc/topics/releases/oxygen.rst
```diff
@@ -281,6 +281,54 @@ Salt Minion now no longer need to be configured against a specifig DNS name or I
...
+
+**Master configuration**
+
+To simply enable it by default, accepting all kind of minions, is enough to set it to True:
+
+.. code-block:: yaml
+
+       discovery: true
+
+Sub-option `mapping` allows autodiscovery help Minions to find a proper Master for them. The mapping contains
```
rallytime: `autodiscovery help` --> `autodiscovery to help`

rallytime (changes requested):
Hi @isbm! Thanks for getting this in. I just have a couple of small grammatical requests.

rallytime:
One lint error: https://jenkins.saltstack.com/job/PR/job/salt-pr-lint-n/18166/violations/file/salt/master.py/

doc/topics/releases/oxygen.rst
```diff
@@ -281,6 +281,54 @@ Salt Minion now no longer need to be configured against a specifig DNS name or I
...
 
+Configuration
+=============
+
+By default, automatic discovery is disabled.
+
+Warning: due to the current limitations that will be changing in a future, before you turn on autodiscovery,
+make sure your network is secured and trusted. See the limitations below.
+
+Autodiscovery is configured on Master and Minion. Both of them are configured via ``discovery`` option
```
terminalmage: For clarity, this should be "Autodiscovery is configured both on the Master and Minion".

gtmanfred (approved):

doc/topics/releases/oxygen.rst
terminalmage: Instead of "it", we should just say "discovery". Also, "accepting all kind of minions" should be "accepting all minions". Finally, "is enough to set it to True" should be:

```
simply set `discovery` to `True`
```

doc/topics/releases/oxygen.rst
```diff
@@ -281,6 +281,54 @@ Salt Minion now no longer need to be configured against a specifig DNS name or I
...
+**Master configuration**
+
+To simply enable it by default, accepting all kind of minions, is enough to set it to True:
+
+.. code-block:: yaml
+
+       discovery: true
+
+Sub-option `mapping` allows autodiscovery help Minions to find a proper Master for them. The mapping contains
+just an arbitrary set of key/value strings that would match either any of them or all of them. Default is empty.
```
terminalmage:
```
A sub-option called `mapping` allows autodiscovery to help find the proper
Master. The mapping contains an arbitrary set of key/value pairs, which the
Minion configuration can target. By default, no mappings are set.
```

doc/topics/releases/oxygen.rst
```diff
@@ -281,6 +281,54 @@ Salt Minion now no longer need to be configured against a specifig DNS name or I
...
+Example:
+
+.. code-block:: yaml
+
+       discovery:
+         mapping:
+           description: SES 5.0
+           node: 1
+
+More of this below.
```
terminalmage: More of what? Where?

doc/topics/releases/oxygen.rst
```diff
@@ -281,6 +281,54 @@ Salt Minion now no longer need to be configured against a specifig DNS name or I
...
+
+It is also possible to change default ``4520`` port, by setting another integer value to the ``port`` option within
+the discovery system. However, this should be also pre-configured on the minions as well.
```
terminalmage:
```
It is also possible to change the port used from the default of ``4520``, by
setting a ``port`` option under the Master's ``discovery`` configuration:

.. code-block:: yaml

    discovery:
      port: 4567

.. note::
    When using a port number other than the default, the Minion's ``discovery``
    configuraton must *also* have a port specified, otherwise the Minion will
    still attempt to contact the Master on port ``4520``.
```

doc/topics/releases/oxygen.rst
```diff
@@ -281,6 +281,54 @@ Salt Minion now no longer need to be configured against a specifig DNS name or I
...
+
+**Minion configuration**
+
+Additionally to the identical configuration as on a Master, there are two more options available on a Minion:
```
terminalmage: Are you trying to say that the minion supports both the ``mapping`` and ``port`` options? If so, we should just say that instead of "the identical configuration as on a Master". Also, you say there are two more options, but you have listed three. How about this?

```
In addition to the ``mapping`` and ``port`` options, the following additional
options are available to Minions:
```

isbm:
@rallytime done!

doc/topics/releases/oxygen.rst
```diff
@@ -281,6 +281,54 @@ Salt Minion now no longer need to be configured against a specifig DNS name or I
...
+
+- ``attempts`` will set how many "ping" broadcast requests should be sent to the network, waiting for any Master
+   response. Each attempt takes a couple of seconds, so raising it to the higher value will
+   result to a slower startup of the Minion. Generally, on a properly configured networks, the autodiscovery
+   supposed to be all set right after the first attempt. Default is set to 3 attempts.
+- ``match`` can have values ``all`` or ``any``. Match is always coming together with ``mapping``, and essentially means
+  "match any/all of mapping values to the target". Default is set to ``any``.
+- ``pause`` is an interval in seconds between attempts. Default is set to five (5) seconds.
```
terminalmage:
```
- ``attempts`` - This option specifies how many broadcast requests should be
  sent to the network, waiting for any Master response. Each attempt takes a
  couple of seconds, so raising this value may result in a slower Minion
  startup. Note that, on a properly-configured network, autodiscovery should
  succeed on the first attempt. By default, this value is set to ``3``.
- ``match`` - This option can be set to either ``all`` or ``any``, and it
  determines how the values configured in ``mapping`` are matched. If set to
  ``all``, then all of the key/value pairs in the Minion's ``mapping`` must
  match a given Master. If set to ``any`` (the default), then any match to a
  key/value mapping will constitute a match.
- ``pause`` - The interval in seconds between attempts (default: 5).
```

terminalmage (changes requested):

terminalmage:
While in German, compound words are usually just concatenated to each other, in English we sometimes use dashes for clarity. @rallytime would it make sense to use `auto-discovery` instead of `autodiscovery`?

rallytime:
@terminalmage Yeah, I personally like `auto-discovery` better. I think it's easier to read, at least for me.

rallytime (approved):

isbm:
@terminalmage @rallytime OK, last swipe. Please look it up and greenbutton that, if no more issues. :wink:

doc/topics/releases/oxygen.rst
```diff
@@ -273,14 +273,71 @@ The new grains added are:
 * ``iscsi_iqn``: Show the iSCSI IQN name for a host
 * ``swap_total``: Show the configured swap_total for Linux, *BSD, OS X and Solaris/SunOS
 
-Salt Minion Autodiscovery
+Salt Minion Auto-discovery
 ------------------------
 
-Salt Minion now no longer need to be configured against a specifig DNS name or IP address of a Master.
+Salt Minion now no longer need to be configured against a specific DNS name or IP address of a Master.
```
terminalmage: Sorry, missed this in my last review, I think because these lines were not part of the diff before your recent commits:

```
Using auto-discovery, the Salt Minion now no longer needs to be configured
against a specific DNS name or IP address of a Master.
```

doc/topics/releases/oxygen.rst
```diff
@@ -273,14 +273,71 @@ The new grains added are:
...
+Configuration
+=============
+
+By default, automatic discovery is disabled.
+
+..warning::
+   Due to the current limitations that will be changing in a future, before you turn on auto-discovery,
+   make sure your network is secured and trusted.
+
+Auto-discovery is configured on Master and Minion. Both of them are configured via ``discovery`` option
```
terminalmage: Should be "via the"

doc/topics/releases/oxygen.rst
```diff
@@ -273,14 +273,71 @@ The new grains added are:
...
+as follows:
+
+**Master configuration**
+
+To simply enable it by default, accepting all kinds of minions, it is enough to set it to True:
```
terminalmage: Can we re-word this?

```
To use the default configuration, which accepts any minion, simply set ``discovery`` to True:
```

doc/topics/releases/oxygen.rst
```diff
@@ -273,14 +273,71 @@ The new grains added are:
...
+
+.. code-block:: yaml
+
+       discovery: true
+
+Sub-option `mapping` allows auto-discovery to help Minions to find a proper Master for them. The mapping contains
+just an arbitrary set of key/value strings that would match either any of them or all of them. Default is empty.
```
terminalmage: This doesn't look like it was updated per my earlier review:

```
A sub-option called `mapping` allows autodiscovery to help find the proper
Master. The mapping contains an arbitrary set of key/value pairs, which the
Minion configuration can target. By default, no mappings are set.
```

doc/topics/releases/oxygen.rst
```diff
@@ -273,14 +273,71 @@ The new grains added are:
...
+
+In addition to the ``mapping`` and ``port`` options, the following additional options are available to Minions:
+
+- ``attempts`` will set how many "ping" broadcast requests should be sent to the network, waiting for any Master
+   response. Each attempt takes a couple of seconds, so raising it to the higher value will
+   result to a slower startup of the Minion. Generally, on a properly configured network, the autodiscovery is
+   supposed to be all set right after the first attempt. Default is set to 3 attempts.
+- ``match`` can have values ``all`` or ``any``. Match is always coming together with ``mapping``, and essentially means
+  "match any/all of mapping values to the target". Default is set to ``any``.
+- ``pause`` is an interval in seconds between attempts. Default is set to five (5) seconds.
```
terminalmage: This also does not appear to have been updated per my last review:

```
- ``attempts`` - This option specifies how many broadcast requests should be
  sent to the network, waiting for any Master response. Each attempt takes a
  couple of seconds, so raising this value may result in a slower Minion
  startup. Note that, on a properly-configured network, autodiscovery should
  succeed on the first attempt. By default, this value is set to ``3``.
- ``match`` - This option can be set to either ``all`` or ``any``, and it
  determines how the values configured in ``mapping`` are matched. If set to
  ``all``, then all of the key/value pairs in the Minion's ``mapping`` must
  match a given Master. If set to ``any`` (the default), then any match to a
  key/value mapping will constitute a match.
- ``pause`` - The interval in seconds between attempts (default: 5).
```

terminalmage (changes requested):
Just a few more changes to the docs

doc/topics/releases/oxygen.rst
isbm: Right, ...and it is "auto-discovery". :wink:

isbm:
@terminalmage @rallytime yep, seems things were missed, sorry! I've rebased + updated docs. Thanks for helping with docs, @terminalmage !

salt/config/__init__.py
```diff
@@ -4009,6 +3999,7 @@ def apply_master_config(overrides=None, defaults=None):
 
     # Check and update TLS/SSL configuration
     _update_ssl_config(opts)
+    _update_discovery_config(opts, minion=False)
```
terminalmage: The minion argument can be removed here (see above).

salt/config/__init__.py
```diff
@@ -3811,10 +3781,31 @@ def apply_minion_config(overrides=None,
 
     # Check and update TLS/SSL configuration
     _update_ssl_config(opts)
+    _update_discovery_config(opts)
 
     return opts
 
 
+def _update_discovery_config(opts, minion=True):
```
terminalmage: We actually don't need to use an argument here to distinguish master from minion. `apply_master_config()` sets `opts['__role'] = 'master'`, so we can just get rid of the `minion` argument here.

salt/config/__init__.py
```diff
@@ -3811,10 +3781,31 @@ def apply_minion_config(overrides=None,
...
+    :return:
+    '''
+    if opts.get('discovery') not in (None, False):
+        if opts['discovery'] is True:
+            opts['discovery'] = {}
+        discovery_config = {'attempts': 3, 'pause': 5, 'port': 4520, 'match': 'any', 'mapping': {}}
+        for key in opts['discovery'].keys():
+            if key not in discovery_config:
+                raise salt.exceptions.SaltConfigurationError('Unknown discovery option: {0}'.format(key))
+        if not minion:
```
terminalmage: We can replace this line with:

```python
if opts.get('__role') == 'master':
```

terminalmage (changes requested):
I took a second look at the actual fix code and found a few improvements we can make. Once these are done this should be good to merge.

salt/config/__init__.py
isbm: Ah, cool. Didn't know about that. Actually `!= 'minion'`, but point taken. Thanks!

salt/master.py
```diff
@@ -751,13 +750,16 @@ def start(self):
...
+                        port=self.opts['discovery']['port'],
                         listen_ip=self.opts['interface'],
                         answer={'mapping': self.opts['discovery'].get('mapping', {})}).run)
                 else:
                     log.error('Unable to load SSDP: asynchronous IO is not available.')
                     if sys.version_info.major == 2:
                         log.error('You are using Python 2, please install "trollius" module to enable SSDP discovery.')
+            self.process_manager.add_process(
+                FileserverUpdate,
+                args=(self.opts,))
```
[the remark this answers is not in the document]
isbm: Right. This was exactly the case. Removed.

isbm:
@terminalmage thanks for _actually_ looking at the code! Fixd.

terminalmage (approved):

rallytime:
Go Go Jenkins!

PR #45419, opened 2018-01-12
[oxygen] Oxygen updating utils and related tests for Unicode round three

garethgreenaway:
### What does this PR do?
Updating various utils and related tests for Unicode goodness

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/utils/napalm.py
+++ b/salt/utils/napalm.py
@@ -16,7 +16,7 @@ Utils for the NAPALM modules and proxy.
 '''
 
 # Import Python libs
-from __future__ import absolute_import, unicode_literals
+from __future__ import absolute_import, unicode_literals, print_function
 import traceback
 import logging
 import importlib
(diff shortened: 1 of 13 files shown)


salt/utils/napalm.py
```diff
@@ -316,7 +316,9 @@ def get_device(opts, salt_obj=None):
         try:
             provider_lib = importlib.import_module(network_device.get('PROVIDER'))
         except ImportError as ierr:
-            log.error('Unable to import {0}'.format(network_device.get('PROVIDER')), exc_info=True)
+            log.error('Unable to import %s',
+                      network_device.get('PROVIDER'),
+                      exc_info=True)
```
terminalmage: Just FYI, `log.error('something', exc_info=True)` and `log.exception('something')` are identical.

https://docs.python.org/2/library/logging.html#logging.exception

I don't think it necessarily needs to be changed though, this is just an FYI.

terminalmage (approved):

rallytime:
re-run py

PR #45424, opened 2018-01-12
Fix some issues with reg.py

twangboy:
### What does this PR do?
Was throwing a key not found in instances where the key was found but
the key/value pair inside was not found.
Use trace instead of debug to show this issue... very noisy

Found in testing

### Tests written?
No

### Commits signed with GPG?
Yes


skizunov (approved):

rallytime (approved):

rallytime:
@twangboy Can you submit this against 2017.7.3 as well?

PR #45425, opened 2018-01-12
Fix stack trace on machines without docker

twangboy:
### What does this PR do?
Fixes a stack trace that was occurring on machines without docker.

Found in testing

### Previous Behavior
The following stack trace would show in debug:
```
[DEBUG   ] Error loading module.swarm: __init__ failed
  File "c:\salt-dev\salt\salt\loader.py", line 1504, in _load_module
    module_init(self.opts)
  File "c:\salt-dev\salt\salt\modules\swarm.py", line 48, in __init__
    __context__['client'] = docker.from_env()
NameError: global name 'docker' is not defined
```

### New Behavior
No stacktrace

### Tests written?
No

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/modules/swarm.py
+++ b/salt/modules/swarm.py
@@ -29,10 +29,6 @@ import salt.utils.json
 try:
     import docker
     HAS_DOCKER = True
-    def __init__(self):
-        __context__['client'] = docker.from_env()
-        __context__['server_name'] = __grains__['id']
-
 except ImportError:
     HAS_DOCKER = False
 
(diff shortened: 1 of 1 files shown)


salt/modules/swarm.py
```diff
@@ -29,6 +29,10 @@
 try:
     import docker
     HAS_DOCKER = True
+    def __init__(self):
```
gtmanfred: Probably better to just check `HAS_DOCKER` in the init function.

gtmanfred (changes requested):

salt/modules/swarm.py
twangboy: @gtmanfred Done

gtmanfred (approved):

rallytime (approved):

PR #45426, opened 2018-01-12
[PY3] Add unicode_literals to salt.utils modules (P)

terminalmage:
@Ch3LL can you take a look at `test_log_created()` in `test/unit/utils/test_parsers.py`? I was getting failures, so I investigated it and found some things that I think may have been inadvertently implemented incorrectly. Firstly, the `is` operator was used to compare strings. The `is` operator doesn't do an `==` comparison, it checks that the two things being compared are the _same exact instance_. In Python 2, as an optimization, the memory address of two `str` types with the same content is the same (i.e. only one string is allocated in memory and both instances point to that same string).

```python
>>> x = 'log_file'
>>> x is 'log_file'
True
```

The behavior of `unicode` types is different though:

```python
>>> x = 'log_file'
>>> x is 'log_file'
False
```

After making my edits to this test, it now is successfully skipped in the salt-key instance of that test (it was no longer being skipped, and thus failed, when I turned on `unicode_literals`).

I believe my edits are correct, but I would appreciate it if you could check my work and make sure I didn't do anything wrong, since you were the original author of the test.

### DIFF
--- a/tests/unit/utils/test_parsers.py
+++ b/tests/unit/utils/test_parsers.py
@@ -4,7 +4,7 @@
 '''
 
 # Import python libs
-from __future__ import absolute_import
+from __future__ import absolute_import, print_function, unicode_literals
 import logging
 import os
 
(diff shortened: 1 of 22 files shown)


rallytime (approved):

terminalmage:
This has been rebased to include Friday's PR that fixes the test daemon timeouts.

rallytime:
@terminalmage There's a merge conflict on this one now.

tests/unit/utils/test_parsers.py
```diff
@@ -499,21 +499,20 @@ def test_log_created(self):
         log_file_name = self.logfile_config_setting_name
         opts = self.default_config.copy()
         opts.update({'log_file': log_file})
-        if log_file_name is not 'log_file':
-            opts.update({log_file_name:
-                         getattr(self, log_file_name)})
```
Ch3LL: I believe we still need to keep this getattr and keep it at `log_file_name` on line 516.. Essentially what this is doing is if `log_file_name` is not `log_file` it means we are expecting a different argument in the configs to set where we are going to log. For example for ssh it looks for `ssh_log_file` instead of `log_file`. So in the test we are grabbing that value with `getattr(self, log_file_name)` since log_file_name equals `ssh_log_file`. This ensures we are testing that we are creating the log file that we are setting intentionally in the config instead of the default. Hopefully that explanation makes sense and maybe there is a better way of doing it that you see. Thanks for catching the is part!!

Ch3LL (changes requested):

tests/unit/utils/test_parsers.py
terminalmage: Good call on the attribute, I didn't grok what you were doing there and thought the log file being tested was the one being set to the local `log_file` variable. Thanks for the review!

Ch3LL: Looks great :) thanks

Ch3LL (approved):

rallytime:
@terminalmage This still has a merge conflcit. Can you rebase?

terminalmage:
D'oh, forgot to rebase before. Sorry, it's rebased now.

rallytime:
re-run lint

PR #45454, opened 2018-01-15
Add use_bin_type to msgpack gate

techhat:
### What does this PR do?
When checking for msgpack in `payload.py`, this adds a `use_bin_type` parameter.

None known.

### Previous Behavior
In Python 3:
```
/usr/lib/python3.6/site-packages/msgpack/__init__.py:47: FutureWarning: use_bin_type option is not specified. Default value of the option will be changed in future version.
  return Packer(**kwargs).pack(o)
```
### New Behavior
`FutureWarning` goes away.

### Tests written?
No

### Commits signed with GPG?
Yes


cachedout (approved):
This should be safe so far as I can tell.

rallytime (approved):

PR #45459, opened 2018-01-16
[2016.11] Salt Cloud: fix loading UTF-8 cached data

### LINKED ISSUE #20823: unicode on ec2 cloud
Hello! EC2 operations fails when there is unicode tags on any of instances:

```
$ salt-call -l trace cloud.full_query
.....
[DEBUG   ] Failed to execute 'ec2.list_nodes_min()' while querying for running nodes: 'ascii' codec can't encode characters in position 70860-70863: ordinal not in range(128)
  File "/usr/lib/python2.7/dist-packages/salt/cloud/__init__.py", line 2198, in run_parallel_map_providers_query
    cloud.clouds[data['fun']]()
  File "/usr/lib/python2.7/dist-packages/salt/cloud/clouds/ec2.py", line 2720, in list_nodes_min
    instances = query(params)
  File "/usr/lib/python2.7/dist-packages/salt/cloud/clouds/ec2.py", line 345, in query
    result.text
UnicodeEncodeError: 'ascii' codec can't encode characters in position 70860-70863: ordinal not in range(128)
......
```

By the way, even with --hard-crash option this exception is caught somewhere, and I can't figure out how to run pdb to debug it.

vutny:
### What does this PR do?
It makes sure that msgpack loads previously cached cloud event data in UTF-8. According to the API it always dumps the data in the `utf-8` encoding, but reads as raw (read ASCII) by default.
https://msgpack-python.readthedocs.io/en/latest/api.html#msgpack.Packer
https://msgpack-python.readthedocs.io/en/latest/api.html#msgpack.Unpacker

Should finally fix #20823.

### Previous Behavior
Salt Cloud was unable to compare cached data which contains unicode symbols with actual data read from an event.

### New Behavior
Salt Cloud successfully loads and works with cached data stored in UTF-8.

### Tests written?
No

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/utils/cloud.py
+++ b/salt/utils/cloud.py
@@ -2508,7 +2508,7 @@ def cachedir_index_add(minion_id, profile, driver, provider, base=None):
 
     if os.path.exists(index_file):
         with salt.utils.fopen(index_file, 'r') as fh_:
-            index = msgpack.load(fh_)
+            index = msgpack.load(fh_, encoding='utf-8')
     else:
         index = {}
 
@@ -2540,7 +2540,7 @@ def cachedir_index_del(minion_id, base=None):
 
     if os.path.exists(index_file):
         with salt.utils.fopen(index_file, 'r') as fh_:
-            index = msgpack.load(fh_)
+            index = msgpack.load(fh_, encoding='utf-8')
     else:
         return
 
@@ -2637,7 +2637,7 @@ def change_minion_cachedir(
     path = os.path.join(base, cachedir, fname)
 
     with salt.utils.fopen(path, 'r') as fh_:
-        cache_data = msgpack.load(fh_)
+        cache_data = msgpack.load(fh_, encoding='utf-8')
 
     cache_data.update(data)
 
@@ -2713,7 +2713,7 @@ def list_cache_nodes_full(opts, provider=None, base=None):
                 fname = '{0}.p'.format(minion_id)
                 fpath = os.path.join(min_dir, fname)
                 with salt.utils.fopen(fpath, 'r') as fh_:
-                    minions[driver][prov][minion_id] = msgpack.load(fh_)
+                    minions[driver][prov][minion_id] = msgpack.load(fh_, encoding='utf-8')
 
     return minions
 
@@ -2992,7 +2992,7 @@ def diff_node_cache(prov_dir, node, new_data, opts):
 
     with salt.utils.fopen(path, 'r') as fh_:
         try:
-            cache_data = msgpack.load(fh_)
+            cache_data = msgpack.load(fh_, encoding='utf-8')
         except ValueError:
             log.warning('Cache for {0} was corrupt: Deleting'.format(node))
             cache_data = {}
--- a/salt/utils/cloud.py
+++ b/salt/utils/cloud.py
@@ -88,6 +88,10 @@ try:
 except ImportError:
     HAS_GETPASS = False
 
+# This is required to support international characters in AWS EC2 tags or any
+# other kind of metadata provided by particular Cloud vendor.
+MSGPACK_ENCODING = 'utf-8'
+
 NSTATES = {
     0: 'running',
     1: 'rebooting',
@@ -2508,7 +2512,7 @@ def cachedir_index_add(minion_id, profile, driver, provider, base=None):
 
     if os.path.exists(index_file):
         with salt.utils.fopen(index_file, 'r') as fh_:
-            index = msgpack.load(fh_, encoding='utf-8')
+            index = msgpack.load(fh_, encoding=MSGPACK_ENCODING)
     else:
         index = {}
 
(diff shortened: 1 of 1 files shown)


vutny:
@rallytime Please could somebody review this stuff?

damon-atkins (changes requested):
I understand you say its the default, however for eases of reading please add encoding='utf-8'  to the pack/dump side.   The it looks consistent.

However what a bad idea for msgpack not to have the same default for in and out.

salt/utils/cloud.py
```diff
@@ -88,6 +88,8 @@
 except ImportError:
     HAS_GETPASS = False
 
+MSGPACK_ENCODING = 'utf-8'
```
damon-atkins: Add a comment required to support international characters in tags. for example.

damon-atkins (changes requested):

vutny:
@damon-atkins Thanks for your review! Yeah, that's very strange design in msgpack as for me too.

damon-atkins:
Posted to cloud channel for someone to test this out  (saltstackcommunity.slack.com)

salt/utils/cloud.py
```diff
@@ -88,6 +88,10 @@
 except ImportError:
     HAS_GETPASS = False
 
+# This is required to support international characters in AWS EC2 tags or any
+# other kind of metadata provided by particular Cloud vendor.
+MSGPACK_ENCODING = 'utf-8'
```
gtmanfred: Can we just use `__salt_system_encoding__` instead?

gtmanfred (changes requested):

salt/utils/cloud.py
vutny: @gtmanfred I'm not sure, really.

In my understanding, the locale of the system that Salt is running on has nothing to do with any weird characters, which could potentially appear in a Cloud instance metadata. They could be filled in manually, or by another automation system. Anyhow, Salt Cloud should be able to write everything on disk and restore without loosing bits.
I think `utf-8` is best choice here and it seems that `msgpack` supports it well by default.

But if you think that it should be defined with `__salt_system_encoding__`, we probably need to update the docs that host (or user) running `salt-cloud` should have proper locale set up.

vutny: In contrast, if any unicode symbol somehow appears in some instance metadata, and you would not set the correct locale on the Salt host, the `salt-cloud` would entirely fail with `UnicodeDecodeError` exception.
That's the case when it's really hard to figure out what you should do with that.

@gtmanfred Please just let me know what do think would be the better way to handle this?

gtmanfred: So, the problem is that if the systems local is not setup to handle utf-8, we shouldn't handle it.

vutny: Okay. That makes sense. I will follow up with documentation update for Salt Cloud to cover necessary locale setup.

Thanks.

damon-atkins: utf-8 is correct as this is what is coming from AWS.    ``__salt_system_encoding__``  is wrong as it could be anything and does not represent the results of an AWS REST API.  The data must be stored in a Unicode compatible format so it can be store in variables within Salt as Unicode.

The salt cli is responsible to convert the output of any salt-cloud command to __salt_system_encoding__ from internal Unicode variables. (I suspect PY2 & PY3 already do this in the log library)

@terminalmage will be adding code to convert cli arguments to Unicode (PY2 only)

vutny:
Rebased against latest changes in the `2016.11` branch.

damon-atkins (approved):

gtmanfred (changes requested):
I still do not feel good about not using `__salt_system_encoding__`, because you could end up with data coming back into salt that cannot be used because the system encoding does not support it.

If people want to be able to use this data, they need to set their encoding to utf-8 anyway, otherwise it has a chance to break somewhere else, that won't be obvious what is at fault.

If you look at fileservers we use this for `s3fs` when pulling down possible unicode data.

It is fine if someone else wants to overrule me, but i think using the system encoding is correct.

damon-atkins:
hi @terminalmage could you drop by this PR.

rallytime:
I am in agreement with @gtmanfred on this one.

damon-atkins:
So if a Chines web pages is loaded, you wish to store it has US ASCII and expect to get the same contents back. mmm.

So if tags are mix of Chines, English, Japanese. And the master is LANG=C (__salt_system_encoding__) and therefore  you expect it to store ok and be retrieved ok.

damon-atkins:
A recent update from
https://github.com/msgpack/msgpack-python
Deprecating encoding option encoding and unicode_errors options are deprecated.
In case of packer, use UTF-8 always. Storing other than UTF-8 is not recommended.

gtmanfred:
Can you provide me a link to where it actually says that?

Thanks,
Daniel

damon-atkins:
https://github.com/msgpack/msgpack-python

gtmanfred (approved):
For anyone that comes by later to see this, here is the actual line in the msgpack page.

https://github.com/msgpack/msgpack-python/blob/master/README.rst#deprecating-encoding-option

Since it is recommended by msgpack to use unicode.

Daniel

vutny:
Thank you guys!

Issue #45413, opened 2018-01-12
Installing variant of mac port fails if no version is provided


KaiSzuttor:
### Description of Issue/Question
Calling ``sudo salt-call pkg.install boost variant='+openmpi'``
ends up with
```
Error running 'pkg.install': Problem encountered installing package(s). Additional info follows:

changes:
    ----------
errors:
    Command Failed: ['port', 'install', 'boost+openmpi']
    Return Code: 1
    Output:
    Error: Error: Port boost+openmpi not found
```
while running ``sudo salt-call pkg.install boost version='1.65.1_1' variant='+openmpi'``
ends up with
```
local:
    ----------
    boost:
        ----------
        new:
            1.65.1_1+no_single+no_static+openmpi+python27
        old:
            1.65.1_1+no_single+no_static+python27
```

### Steps to Reproduce Issue
(Include debug logs if possible and relevant.)
See above.

### Versions Report
(Provided by running `salt --versions-report`. Please also mention any differences in master/minion versions.)
```
Salt Version:
           Salt: 2017.7.0

Dependency Versions:
           cffi: Not Installed
       cherrypy: Not Installed
       dateutil: Not Installed
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.10
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.4.8
   mysql-python: Not Installed
      pycparser: Not Installed
       pycrypto: Not Installed
   pycryptodome: Not Installed
         pygit2: Not Installed
         Python: 2.7.14 (default, Sep 27 2017, 12:15:00)
   python-gnupg: Not Installed
         PyYAML: 3.12
          PyZMQ: 16.0.3
           RAET: Not Installed
          smmap: Not Installed
        timelib: Not Installed
        Tornado: 4.5.2
            ZMQ: 4.2.3

System Versions:
           dist:
         locale: UTF-8
        machine: x86_64
        release: 17.3.0
         system: Darwin
        version: 10.13.2 x86_64
```

aphor:
The module produces a command with a single port name and version (if given), and variants (if given) concatenated as a single positional arg to the "port install" command.

This is incorrect. The command should give version and variants as separate args eg. 
`port install portname @version +variant`
[...]

PR #45462, opened 2018-01-16
emit port cli version, variants as separate args

aphor:
### What does this PR do?
Changes macports pkg module behavior to conform to macports documentation for single pkg with optional version and variants arguments.

https://github.com/saltstack/salt/issues/45413

### Previous Behavior
The module produces a command with a single port name and version (if given), and variants (if given) concatenated as a single positional arg to the "port install" command.

This is incorrect. The command should give version and variants as separate args eg. 
`port install portname @version +variant`
and the current rev, where version kwarg is omitted and variant is given will try to execute\
`port install portname+variant`
which is incorrect.

### New Behavior
give version and variants as separate args eg. 
`port install portname @version +variant`

### Tests written?

No

### Commits signed with GPG?
Yes


KaiSzuttor:
@aphor have you checked your PR? I just tried and an exception occurred:
```
                File "xxx/git/salt/salt/state.py", line 1964, in call
                  **cdata['kwargs'])
                File "xxx/git/salt/salt/loader.py", line 1822, in wrapper
                  return f(*args, **kwargs)
                File "xxx/git/salt/salt/states/pkg.py", line 1663, in installed
                  **kwargs)
                File "xxx/git/salt/salt/modules/mac_ports.py", line 330, in install
                  if 'version' in pparams:
```

aphor:
@KaiSzuttor can you provide an example from which I can add a test?

I checked the code with 4 cases, a single port (from the bug example) with all combinations of version and variant given or not.

If you're getting an exception, I think it would be good to first add a test and then fix it.

salt/modules/mac_ports.py
```diff
@@ -323,7 +323,13 @@ def install(name=None, refresh=False, pkgs=None, **kwargs):
 
     formulas_array = []
     for pname, pparams in six.iteritems(pkg_params):
-        formulas_array.append(pname + (pparams or ''))
+        formulas_array.append(pname)
+
+        if 'version' in pparams:
```
gtmanfred: you should do a `if pparams and 'version' in pparams:`

because it is possible that pparams is a None value.

gtmanfred (changes requested):

aphor:
I think @gtmanfred is probably right, and I'll fix the PR with his suggestion. I'll have to look into setting up a test for the state module which would cover your case @KaiSzuttor.

aphor:
@gtmanfred I dropped both formulas_array.append() calls into an if block testing for True-ish-ness of pparams. It wasn't exactly the changes you requested, but I think it covers your request, and probably fixes the error @KaiSzuttor got.

KaiSzuttor:
Just checked, variant is still ignored for the state file.

gtmanfred:
@aphor did you see @KaiSzuttor last comment about this causing 'variant' to be ignored when passed from a state?

aphor:
@gtmanfred Yeah, I saw that. I've been sidetracked, but I'll be able to deal with this over the weekend because we're going to get snowed in here in Chicago.

gtmanfred:
cool, thanks!

aphor:
@KaiSzuttor @gtmanfred 

The pkg state module's pkg.installed() doesn't support a variant argument, so it makes calls to the mac_ports execution module to install packages without the (optional) variant kwarg.

I looked at adding support for this into the state module, but I think that should be a separate ticket/PR.

I submitted #45967 "pkg state module does not allow specifying variants for macports packages" to address this.

gtmanfred (approved):

PR #45464, opened 2018-01-16
grains/core.py: use ascii strings when reading init system information

adelcast:
grains/core.py has the default string enconding set to unicode, via:

from _future_ import unicode_literals

however, on the os_data function, we read binary data using the function
salt.utils.files.fopen, which returns an ascii string. In PY2, explicity
convert literals to str to avoid unicode decode exceptions on the binary
data.

Signed-off-by: Alejandro del Castillo <alejandro.delcastillo@ni.com>

### What does this PR do?

Fix broken os_data grains

### Tests written?

No

### Commits signed with GPG?

No

### DIFF
--- a/salt/grains/core.py
+++ b/salt/grains/core.py
@@ -1462,16 +1462,24 @@ def os_data():
                         try:
                             with salt.utils.files.fopen(init_bin, 'rb') as fp_:
                                 buf = True
-                                edge = six.b('')
+                                if six.PY3:
+                                    edge = six.b('')
+                                else:
+                                    edge = str('')
                                 buf = fp_.read(buf_size).lower()
                                 while buf:
                                     buf = edge + buf
                                     for item in supported_inits:
+                                        if six.PY2:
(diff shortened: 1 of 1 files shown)


rallytime:
Hi @adelcast - Thanks for submitting this change. We've actually been working very hard on fixing these typs of issues in the `oxygen` branch. It looks to me like this has been fixed downstream, and when it was merged-forward, it caused a merge conflict.

I think this PR should likely be closed since we're fixing this in a slightly different way. However, before closing, I wanted to double check with your first that this is fixed at the HEAD of the develop branch.

rallytime:
@adelcast Just checking in here - did you get a chance to test the latest develop to make sure this issue is fixed? If so, we can close this PR. If not, a rebase will be needed and we can continue the discussion.

adelcast:
yep, feel free to close this PR

rallytime:
Excellent - thanks again @adelcast

PR #45480, opened 2018-01-16
aptpkg: force additional environment variables to be str types

terminalmage:
This prevents tracebacks in subprocess due to unicode values in
the dict merged into os.environ via the module's `__init__` func.


terminalmage:
This may not actually be necessary since the code that expects str types is unique to Windows, due to the fact it uses CreateProcess to spawn the process.

terminalmage:
actually yes, it's still a good practice, because subprocess will use `ascii` to encode any actual unicode characters in the env dict.

rallytime:
re-run py

rallytime (approved):

rallytime:
re-run py

rallytime:
re-run py

Issue #45155, opened 2017-12-24
[PY3] salt/salt/utils/data.py need to support in and out of different langs


damon-atkins:
### Description of Issue/Question
data.py is hard coded to use `__salt_system_encoding__`

For example someone calling salt-call could have a different language to the language defined for salt-minion will result in different output.

Salt Team need to say UTF-8 is the standard for templates and sls files etc (except for binaries). Provided a checker/converter tool for the master in salt-common.   salt-call or salt-minion need to convert to the target lang encoding on output, to one provide by the user in an sls file within the state  or the default language of the OS.

UTF-8  SLS could output English and Japanese.  e.g it could have if statement to action/output  different text based on the minion language.  Suitable example might be Windows Group Policies which are defined in the language of the OS is installed in and also US English. (@twangboy pls confirm)

twangboy:
From what I'm reading, `__salt_system_encoding__` is set to `mbcs` in Windows... which is whatever the Windows system encoding is set to. That's basically what you're saying here right?

https://github.com/saltstack/salt/blob/oxygen/salt/__init__.py#L72

damon-atkins:
Nope.   
Here is a summary of comments made against recent PRs
-- 44877
Assume all PY2 input from files (assumed to be UTF-8 unless indicated otherwise) or minion configs (UTF-8) or command lines (`__salt_system_encoding__`) or jinja (UTF-8) will be converted to Unicode and os.getenv() or os.environ.get() (`__salt_system_encoding__`)
This needs to be utf-8 only. For example someone edits a file on Unix in UTF-8 and then applies it to a windows system with a default language of Japanese. Or the other way around someone create a file on Windows in Japanese and then puts in the salt master which is UTF-8, and applies it to Unix systems.

-- 44892
Salts python code is UTF-8, and therefore control/functions statements like pkg.installed are in UTF-8 used within sls files.

Compares some of the 0-255 8bit languages https://en.wikipedia.org/wiki/ISO/IEC_646 Take note of the following chars `[, ], \ {, }, ^`, they are displayed as different characters in different languages.

-- 44877
The idea is that, internally salt is Unicode/UTF-8. A template can be UTF-8 (I would like to see this as must be) on the master, and need to output to Greek language 8bit (non-unicode) on the local minion.

[...]

PR #45491, opened 2018-01-17
Improve PY3 compatibility of salt.utils.data.decode and encode

terminalmage:
This adds the ability to specify the encoding and error behavior. It
also adds the "errors" argument to the salt.utils.stringutils funcs for
converting between str, unicode, and bytes.

Tests have also been updated to properly test the recursive behavior of
salt.utils.data.decode/encode.

### DIFF
--- a/salt/utils/data.py
+++ b/salt/utils/data.py
@@ -67,143 +67,177 @@ def compare_lists(old=None, new=None):
     return ret
 
 
-def decode(data, preserve_dict_class=False, preserve_tuples=False):
+def decode(data, encoding=None, errors='strict', preserve_dict_class=False, preserve_tuples=False):
     '''
     Generic function which will decode whichever type is passed, if necessary
     '''
-    if isinstance(data, six.string_types):
-        if six.PY2 and isinstance(data, str):
-            return data.decode('utf-8')
-        else:
-            return data
-    elif isinstance(data, collections.Mapping):
-        return decode_dict(data, preserve_dict_class, preserve_tuples)
+    if isinstance(data, collections.Mapping):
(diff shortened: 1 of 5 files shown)


damon-atkins (approved):
Cannot see any issues in the code

rallytime:
@terminalmage There's a lint error

rallytime:
re-run py

rallytime (approved):

terminalmage:
It looks like a lot of the code already using to_unicode was taking advantage of the permissiveness of the function (it wasn't blowing up when non-string/bytes data was passed to it like to_str and to_bytes already did). I may need to tweak this.

terminalmage:
OH! It may be easier than I thought. The offending code looks to be in salt.utils.locales.sdecode. We should actually probably deprecate that in favor of `salt.utils.data.decode()`, but in the short term catching the new TypeError to_unicode raises when non-string/bytes types are passed to it may take care of all of the failures.

damon-atkins:
Do you need to add OrderDict?  `isinstance(abc, (dict, OrderDict))`

damon-atkins:
link #45155  `[PY3] salt/salt/utils/data.py need to support in and out of different langs`

terminalmage:
> Do you need to add OrderDict?  `if isinstance(data, (collections.Mapping, dict, OrderDict))`
Or does `if isinstance(data, collections.Mapping):` cover it inside `def decode(data....`

```python
>>> import collections
>>> from salt.utils.odict import OrderedDict
>>> isinstance(OrderedDict(), collections.Mapping)
True
>>>
```

PR #45498, opened 2018-01-17
Added support for mixed line ending in file

t0fik:
Adds support for mixed line endings in file.line method.

Test written: No

Not signed


terminalmage (commented):
I assume when you say "mixed line ending" you mean windows and unix line endings? If so, this is GTM.

t0fik:
@terminalmage You assume right. What stands for GTM?

salt/modules/file.py
```diff
@@ -1693,6 +1693,13 @@ def line(path, content=None, match=None, mode=None, location=None,
 
             salt '*' file.line /path/to/file content="CREATEMAIL_SPOOL=no" match="CREATE_MAIL_SPOOL=yes" mode="replace"
     '''
+    def _split(content):
+        split_content = []
+        for linesep_line in content.split(os.linesep):
+            for content_line in linesep_line.split('\n'):
```
sergeizv: This code makes a difference only being executed on platforms with os.linesep != '\n'... Why not use str.splitlines() instead?

salt/modules/file.py
```diff
@@ -1728,11 +1735,11 @@ def line(path, content=None, match=None, mode=None, location=None,
         log.warning('Cannot find text to {0}. File \'{1}\' is empty.'.format(mode, path))
         body = ''
     elif mode == 'delete':
-        body = os.linesep.join([line for line in body.split(os.linesep) if line.find(match) < 0])
+        body = os.linesep.join([line for line in _split(body) if line.find(match) < 0])
```
sergeizv: Joining with os.linesep and splitting by something else will lead to all the line breaks replaced in the file...

t0fik: Seems legit to me, to output file with line endings proper to OS.
Converting '\n' to os.linesep is default python behavior for not binary writes.

salt/modules/file.py
t0fik: I'm aware of that.
That specific part of code was tested in file.blockreplace.

sergeizv:
On Fri, Jan 19, 2018 at 10:49:01PM +0000, Jerzy Drozdz wrote:
> t0fik commented on this pull request.
> 
> 
> 
> > @@ -1728,11 +1735,11 @@ def line(path, content=None, match=None, mode=None, location=None,
>          log.warning('Cannot find text to {0}. File \'{1}\' is empty.'.format(mode, path))
>          body = ''
>      elif mode == 'delete':
> -        body = os.linesep.join([line for line in body.split(os.linesep) if line.find(match) < 0])
> +        body = os.linesep.join([line for line in _split(body) if line.find(match) < 0])
> 
> Seems legit to me, to output file with line endings proper to OS.

Might be proper for OS, but confusing for software consuming that
particular file. Function that modifies a line shouldn't silently change
all the file :)

t0fik:
@sergeizv What is your proposition, instead converting line endings? 
For now that particular function cannot handle line endings, which are not proper for the OS. For example: LF instead CRLF on Windows

sergeizv:
On Fri, Jan 19, 2018 at 11:44:16PM +0000, Jerzy Drozdz wrote:
> @sergeizv What is your proposition, instead converting line endings? 

Maybe detect line break first and then split and join by it instead of
os.linesep?

> For now that particular function cannot handle line endings, which are not proper for the OS. For example: LF instead CRLF on Windows

The same is true for blockreplace() as well at least

rallytime:
@t0fik and @sergeizv Where did we land on this PR?

t0fik:
I will rewrite line function to detect line endings. I hadn't time to do it.

salt/modules/file.py
```diff
@@ -1589,6 +1590,13 @@ def _get_line_indent(src, line, indent):
 
     return ''.join(idt) + line.strip()
 
+def _get_line_ending(src, line):
```
damon-atkins: This seems to add line ending, rather than get_line_ending consider a different name for the function.
e.g. _add_src_eol(src, line).

t0fik: @damon-atkins I wanted to be consistent with name scheme of _get_line_indent, which ads indentation to line.

rallytime:
@t0fik There's a lint error: https://jenkins.saltstack.com/job/PR/job/salt-pr-lint-n/19000/violations/

rallytime:
There's also some related test failures: https://jenkins.saltstack.com/job/PR/job/salt-pr-linode-cent7-py3/2006/

salt/modules/file.py
damon-atkins: Good idea, change it

damon-atkins (changes requested):

salt/modules/file.py
```diff
@@ -1587,7 +1588,16 @@ def _get_line_indent(src, line, indent):
...
-    return ''.join(idt) + line.strip()
+    return ''.join(idt) + line.lstrip()
+
+
+def _set_line_eol(src, line):
+    '''
+    Add line ending
+    '''
+    regex = re.compile('((?<!\r)\n|\r(?!\n)|\r\n)$')
+    line_ending = regex.search(src).group()
```
terminalmage: What happens on the final line of the file? Not all files end in a newline, and this regex wouldn't match a line without one:

```python
>>> regex = re.compile('((?<!\r)\n|\r(?!\n)|\r\n)$')
>>> regex.search('foo').group()
AttributeError: 'NoneType' object has no attribute 'group'
```

salt/modules/file.py
```diff
@@ -1549,18 +1549,19 @@ def _regex_to_static(src, regex):
...
 
     try:
-        src = re.search(regex, src)
+        compiled = re.compile(regex, re.DOTALL)
+        src = [res.group() for res in map(compiled.search, src) if res]
     except Exception as ex:
         raise CommandExecutionError("{0}: '{1}'".format(_get_error_message(ex), regex))
 
-    return src and src.group() or regex
+    return src and src[0] or regex
```
terminalmage: Why is this being done? I don't see this function being referenced in any of the changed code.

terminalmage (commented):
I'm not comfortable with something that changes this much to be included in a bugfix release, especially with all the changes to file I/O made for Oxygen. In my opinion this should be done in develop.

rallytime:
@t0fik I am making the rounds on PRs today :)

Did you get a chance to review the comments above? Those will need to be addressed and this change needs to be moved to develop before we can move forward here.

salt/modules/win_file.py
```diff
@@ -45,20 +45,20 @@
...
+                               get_hash, manage_file, file_exists, get_diff, line, list_backups,
+                               __clean_tmp, check_file_meta, _binary_replace,
+                               _splitlines_preserving_trailing_newline, restore_backup,
+                               access, copy, readdir, rmdir, truncate, replace, delete_backup,
+                               search, _get_flags, extract_hash, _error, _sed_esc, _psed,
+                               RE_FLAG_TABLE, blockreplace, prepend, seek_read, seek_write, rename,
+                               lstat, path_exists_glob, write, pardir, join, HASHES, HASHES_REVMAP,
+                               comment, uncomment, _add_flags, comment_line, _regex_to_static,
+                               _set_line_indent, apply_template_on_contents, dirname, basename,
+                               list_backups_dir, _assert_occurrence, _starts_till, _set_line_eol)
```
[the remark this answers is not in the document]
t0fik: It was more readable for me, when I was checking imports.

salt/modules/file.py
t0fik: I'll add guard statement.

t0fik:
@rallytime sorry, I was busy.

terminalmage:
This still needs to be rebased/resubmitted on top of develop.

t0fik:
I'll resubmit PR to develop on weekend.

t0fik:
PR resubmited to develop #46291

PR #45519, opened 2018-01-18
For py2 use io.open io.* which is same as py3 open etc.

damon-atkins:
Set a default buffer to 1 megabyte

### What does this PR do?
For py2 use io.open io.* which is same as py3 open etc.
Set a default buffer to 1 megabyte, improve salt performance the default buffer is badly calculated.

### Previous Behavior
Need to code differently for PY2 and PY3

### New Behavior
Can code the same

### Tests written?

No new tests

### Commits signed with GPG?

No

### DIFF
--- a/salt/utils/files.py
+++ b/salt/utils/files.py
@@ -18,6 +18,9 @@ import tempfile
 import time
 import urllib
 
+if six.PY2:
+    from io import *
+
 # Import Salt libs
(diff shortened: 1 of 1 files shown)


damon-atkins:
Are the test working in oxygen?

damon-atkins:
Got this working sort of working but would need other changes.

PR #45544, opened 2018-01-19
utils/files.py dynamic i/o size for fopen()

damon-atkins:
### What does this PR do?
utils/files.py fopen inteligent chooses an I/O size if the file exists or the call provides the size of the target file.  _utils/files.py does not have access to `__opt__` and best not increase the number of times config files are read, hence a dynamic function was created._

### Previous Behavior
fopen() did 8k on average, unless a programmer indicated otherwise.

### New Behavior
A new function dynamic_buf_size(filepath, max_size=1048576, size=-1) tries to determine the best I/O size.

### Tests written?

No New Tests

### Commits signed with GPG?

No


damon-atkins:
Go Go Jenkins!

damon-atkins:
from the doco check.
```
12:07:35 Exception occurred:
12:07:35   File "/var/jenkins/workspace/PR/salt-pr-docs-n/salt/utils/files.py", line 368, in fopen
12:07:35     kwargs['buffering'] = __opts__['file_buffer_size']
12:07:35 NameError: global name '__opts__' is not defined
```
Not sure how to fix it.

terminalmage (commented):
Changes like this to core I/O behavior need to be reviewed by @thatch45, I'm not sure if there is a reason the minion file_buffer_size is specifically what it is.

I'm also not familiar with the docs check, do you know what is causing that @rallytime ?

thatch45:
Making them the same for the sake of them being the same is wrong. But if we can see an improvement in performance then I am all for it.

damon-atkins:
Go Go Jenkins!

damon-atkins:
Large I/O size increase bandwidth and reduces cpu usage.  However a large buffer for a small file is pointless.

Given __opts___ is not available and it would need to be loaded. Adding to already large  number of times the minion/master configs just to get file_buffer_size did seem like a good idea.  I have decided to head down path, that a function determines the best I/O size based on the existing file size or a size provided by the user.

Two parts to this  have the call to open() use buffering setting.
Improvements on the default buffer size for files of just 8k.

I was surprised to see the master buffering at 1MB. It would be interesting to see the average file size is.

I thought the Jenkins test might be a good benchmark.

damon-atkins:
Something going on with os.path.isfile(filepath) and _isfile_side_effect KeyError. Not sure what the issues is?
```
12:16:48 ----------------------------------------------------------------------
12:16:48 Traceback (most recent call last):
12:16:48   File "/testing/tests/unit/states/test_archive.py", line 165, in test_tar_gnutar
12:16:48     keep=True)
12:16:48   File "/testing/salt/states/archive.py", line 977, in extracted
12:16:48     existing_cached_source_sum = _read_cached_checksum(cached)
12:16:48   File "/testing/salt/states/archive.py", line 120, in _read_cached_checksum
12:16:48     with salt.utils.files.fopen(checksum_file, 'r') as fp_:
12:16:48   File "/testing/salt/utils/files.py", line 402, in fopen
12:16:48     kwargs['buffering'] = dynamic_buf_size(filename)
12:16:48   File "/testing/salt/utils/files.py", line 78, in dynamic_buf_size
12:16:48     not os.path.isfile(filepath)):
12:16:48   File "/usr/local/lib/python2.7/dist-packages/mock.py", line 955, in __call__
12:16:48     return _mock_self._mock_call(*args, **kwargs)
12:16:48   File "/usr/local/lib/python2.7/dist-packages/mock.py", line 1018, in _mock_call
12:16:48     ret_val = effect(*args, **kwargs)
12:16:48   File "/testing/tests/unit/states/test_archive.py", line 46, in _isfile_side_effect
12:16:48     }[path]
12:16:48 KeyError: u'/tmp/archive_hash/foo.tar.gz.hash'
```

damon-atkins:
Go Go Jenkins!

rallytime:
@damon-atkins This looks like a feature addition to me and something that still might need to be discussed about the problem being solved here. Can you move this to develop?

damon-atkins:
Hi @rallytime where do I find stats on run times for the tests in Jenkins I was thinking the tests themselves might show improvements.

damon-atkins:
Go Go Jenkins!

rallytime:
@damon-atkins This needs to be moved to develop, please.

cachedout:
Since this has been moved to develop, I have rebased it and restarted the tests.

damon-atkins:
Thanks.    Seems the errors are not related. What I want to see is the time the tests take with and with out the change.  As a bit of a benchmark.

rallytime (changes requested):
I would rather see a more refined PoC on this rather than just relying on the test results. I don't think that is a very reliable way to demonstrate that this concept is making a noticeable difference.

In addition, the test results above are not showing any noticeable improvement. The tests on develop typically take about 3 hours each to run, depending on many factors. Each of the test results above took a little over 3 hours to run. This is not the way we should be measuring a change like this.

PR #45549, opened 2018-01-19
Add unicode_literals to modules, states, tests (I)

rallytime:


salt/modules/incron.py
```diff
@@ -103,7 +105,7 @@ def _write_incron_lines(user, lines):
     else:
         path = salt.utils.files.mkstemp()
         with salt.utils.files.fopen(path, 'w+') as fp_:
-            fp_.writelines(lines)
+            fp_.writelines(salt.utils.stringutils.to_str(lines))
```
terminalmage: `to_str()` has to operate on strings, not lists. If you want a one-liner that doesn't use a list comprehension, you can open the file using a mode of `wb` and use:

```python
fp_.writelines(salt.utils.data.encode(lines))
```

Just FYI, the `+` in the mode is entirely unnecessary, it opens the file in read-and-write mode and we're not reading from the filehandle at all. It was likely originally written that way by someone who didn't understand what the `+` was for.

salt/modules/incron.py
```diff
@@ -117,12 +119,11 @@ def _write_file(folder, filename, data):
     '''
     path = os.path.join(folder, filename)
     if not os.path.exists(folder):
-        msg = '{0} cannot be written. {1} does not exist'
-        msg = msg.format(filename, folder)
+        msg = '%s cannot be written. %s does not exist', filename, folder
```
terminalmage:
```python
msg = '{0} cannot be written. {1} does not exist'.format(filename, folder)
```

salt/modules/incron.py
```diff
@@ -134,7 +135,7 @@ def _read_file(folder, filename):
     path = os.path.join(folder, filename)
     try:
         with salt.utils.files.fopen(path, 'rb') as contents:
-            return contents.readlines()
+            return salt.utils.stringutils.to_unicode(contents.readlines())
```
terminalmage: `.readlines()` produces a list, not a string. This should be:

```python
return salt.utils.data.decode(contents.readlines())
```

salt/modules/ini_manage.py
```diff
@@ -397,7 +398,7 @@ def flush(self):
             with salt.utils.files.fopen(self.name, 'w') as outfile:
                 ini_gen = self.gen_ini()
                 next(ini_gen)
-                outfile.writelines(ini_gen)
+                outfile.writelines(salt.utils.stringutils.to_str(ini_gen))
```
terminalmage: `ini_gen` is a list here, not a string. Like above, either we need to run `salt.utils.stringutils.to_str()` in a list comprehension, or you can open this file with mode `wb` and use:

```python
outfile.writelines(salt.utils.data.encode(ini_gen))
```

terminalmage (changes requested):
Just a few changes to make. For a couple, you will need to import salt.utils.data, and perhaps remove the stringutils import if it is no longer necessary.

cachedout (approved):

terminalmage (approved):

PR #45574, opened 2018-01-20
Add unicode_literals to more modules, states, and tests

rallytime:


salt/modules/linux_sysctl.py
```diff
@@ -231,7 +234,7 @@ def persist(name, value, config=None):
         nlines.append('{0} = {1}\n'.format(name, value))
     try:
         with salt.utils.files.fopen(config, 'w+') as _fh:
-            _fh.writelines(nlines)
+            _fh.writelines(salt.utils.stringutils.to_str(nlines))
```
terminalmage: nlines is a list here, so this will raise a TypeError. This could be fixed using:

```python
with salt.utils.files.fopen(config, 'wb') as _fh:
    _fh.writelines(salt.utils.data.encode(nlines))
```

terminalmage (changes requested):
Just one small issue.

salt/modules/linux_sysctl.py
terminalmage: Note: this opens the filehandle for binary so that it works with `salt.utils.data.encode()`, which will produce a list of bytestrings.

terminalmage (approved):

PR #45578, opened 2018-01-20
Add unicode_literals to even more modules and states

rallytime:


salt/modules/openbsd_sysctl.py
```diff
@@ -126,7 +130,7 @@ def persist(name, value, config='/etc/sysctl.conf'):
     if not edited:
         nlines.append('{0}={1}\n'.format(name, value))
     with salt.utils.files.fopen(config, 'w+') as ofile:
-        ofile.writelines(nlines)
+        ofile.writelines(salt.utils.data.decode(nlines))
```
terminalmage: This should be `encode`, not `decode`, and the filehandle should be opened with a mode of `wb`.

salt/modules/puppet.py
```diff
@@ -246,7 +247,7 @@ def disable(message=None):
             try:
                 # Puppet chokes when no valid json is found
                 str = '{{"disabled_message":"{0}"}}'.format(message) if message is not None else '{}'
-                lockfile.write(str)
+                lockfile.write(salt.utils.stringutils.to_str(str))
```
terminalmage: `str` is being shadowed here, we should be using a different name for the variable. Both lines 249 and 250 need to be updated.

salt/modules/puppet.py
```diff
@@ -312,7 +313,9 @@ def summary():
 
     try:
         with salt.utils.files.fopen(puppet.lastrunfile, 'r') as fp_:
-            report = salt.utils.yaml.safe_load(fp_)
+            report = salt.utils.yaml.safe_load(
+                salt.utils.stringutils.to_unicode(fp_)
```
terminalmage: PyYAML doesn't need the input to be a unicode string, it's actually designed to read from file streams which are not pre-decoded. So this change can be reverted.

salt/states/pagerduty_escalation_policy.py
```diff
@@ -91,7 +97,7 @@ def present(profile='pagerduty', subdomain=None, api_key=None, **kwargs):
                 if schedule:
                     target_id = schedule['schedule']['id']
             if target_id is None:
-                raise Exception('unidentified target: {0}'.format(str(target)))
+                raise Exception('unidentified target: {0}'.format(six.text_type(target)))
```
terminalmage: `six.text_type()` is unnecessary here

salt/states/pagerduty_schedule.py
```diff
@@ -62,7 +68,7 @@ def present(profile='pagerduty', subdomain=None, api_key=None, **kwargs):
                                                         subdomain=subdomain,
                                                         api_key=api_key)
             if u is None:
-                raise Exception('unknown user: {0}'.format(str(user)))
+                raise Exception('unknown user: {0}'.format(six.text_type(user)))
```
terminalmage: `six.text_type()` is unnecessary here

salt/states/pcs.py
```diff
@@ -340,16 +343,20 @@ def _item_present(name, item, item_id, item_type, show='show', create='create',
...
         cibfile=cibfile)
 
-    log.trace('Output of pcs.item_create: {0}'.format(str(item_create)))
+    log.trace('Output of pcs.item_create: %s', item_create)
 
     if item_create['retcode'] in [0]:
-        ret['comment'] += 'Created {0} {1} ({2})\n'.format(str(item), str(item_id), str(item_type))
-        ret['changes'].update({item_id: {'old': '', 'new': str(item_id)}})
+        ret['comment'] += 'Created {0} {1} ({2})\n'.format(
+            six.text_type(item), six.text_type(item_id), six.text_type(item_type)
```
terminalmage: `six.text_type()` is unnecessary here, we can just pass the three variables as arguments to `.format()` without pre-converting them to strings.

salt/states/pcs.py
```diff
@@ -340,16 +343,20 @@ def _item_present(name, item, item_id, item_type, show='show', create='create',
...
+        )
+        ret['changes'].update({item_id: {'old': '', 'new': six.text_type(item_id)}})
     else:
         ret['result'] = False
-        ret['comment'] += 'Failed to create {0} {1} ({2})\n'.format(str(item), str(item_id), str(item_type))
+        ret['comment'] += 'Failed to create {0} {1} ({2})\n'.format(
+            six.text_type(item), six.text_type(item_id), six.text_type(item_type)
```
terminalmage: Same as above, we can remove all instances of `six.text_type()` on this line and just pass the variables.

salt/states/pcs.py
```diff
@@ -475,19 +482,19 @@ def cluster_setup(name, nodes, pcsclustername='pcscluster', extra_args=None):
...
+    log.trace('Output of pcs.config_show: %s', config_show)
 
     for line in config_show['stdout'].splitlines():
         if len(line.split(':')) in [2]:
             key = line.split(':')[0].strip()
             value = line.split(':')[1].strip()
             if key in ['Cluster Name']:
                 if value in [pcsclustername]:
-                    ret['comment'] += 'Cluster {0} is already set up\n'.format(str(pcsclustername))
+                    ret['comment'] += 'Cluster {0} is already set up\n'.format(six.text_type(pcsclustername))
```
terminalmage: `six.text_type()` is unnecessary here

salt/states/pcs.py
```diff
@@ -475,19 +482,19 @@ def cluster_setup(name, nodes, pcsclustername='pcscluster', extra_args=None):
...
                 else:
                     setup_required = True
                     if __opts__['test']:
-                        ret['comment'] += 'Cluster {0} is set to set up\n'.format(str(pcsclustername))
+                        ret['comment'] += 'Cluster {0} is set to set up\n'.format(six.text_type(pcsclustername))
```
terminalmage: `six.text_type()` is unnecessary here

terminalmage (changes requested):

salt/modules/puppet.py
rallytime: Nice catch

terminalmage (approved):

PR #45579, opened 2018-01-20
Test suite stability fixes

terminalmage:
This PR includes 2 fixes:

1. Fix error with event unpacking introduced in db41d9b
2. Fix breakage affecting pyzmq < 14.3.0 introduced in 00f31bf

It also makes a tweak to a log message which was not showing the correct information for the name of the engine that failed to initialize.


gtmanfred:
This needs to be backported to 2017.7.3

PR #45580, opened 2018-01-20
aptpkg.py: Only encode env vars on PY2

terminalmage:
PY2 can't have unicode env vars, but bytestring env vars cause problems
for subprocess.Popen on PY3. Therefore, only encode the dict on PY2.


terminalmage:
Go Go Jenkins!

rallytime (approved):

PR #45582, opened 2018-01-21
Two salt-ssh fixes

terminalmage:
1. Pass on `OSError` if thin tarball already removed

   This can happen in the test suite while running `integration.ssh.test_state.SSHStateTest.test_state_running` because the instance of salt-ssh running in the background has completed during the second salt-ssh run, and since the contents of the thin tarball haven't changed between runs, the checksum used to derive the path of the thin tarball hasn't changed, so it is unlinked before the 2nd salt-ssh run in that test completes.

2. Add `salt/utils/state.py` to thin tarball

   This is required since some of the code in salt/modules/state.py was moved into salt/utils/state.py so it could be reused.


gtmanfred:
Looks like this is still failing on

```
  File "/testing/tests/integration/ssh/test_state.py", line 172, in test_state_running
    self.assertIn(ret, ' '.join(get_sls))
AssertionError: 'The function "state.pkg" is running as' not found in ''
```

This is also broken in 2017.7.3

terminalmage:
@gtmanfred It looks like the failures only continued to crop up under heavy load. I was only able to reproduce this by starting two docker containers and simultaneously running integration tests on both, which kicked my laptop's load avg up to around 9. I just pushed a few tweaks that seem to have resolved the failures on my end, I did 10 consecutive simultaneous runs on two containers and they both passed each time.

terminalmage:
Tests passed on Ubuntu 14, all failures are unrelated and fixed in https://github.com/saltstack/salt/pull/45595

rallytime (approved):

PR #45592, opened 2018-01-21
Add unicode_literals to modules, states, and tests [n]

rallytime:


rallytime:
The failure `unit.modules.test_nfs3.NfsTestCase.test_list_exports ` is related. Looking into that now.

rallytime:
Ok - that test failure should be fixed now.

salt/modules/netbsd_sysctl.py
```diff
@@ -149,7 +153,9 @@ def persist(name, value, config='/etc/sysctl.conf'):
         nlines.append("{0}\n".format(newline))
 
     with salt.utils.files.fopen(config, 'w+') as ofile:
-        ofile.writelines(nlines)
+        ofile.writelines(
+            salt.utils.data.decode(nlines)
```
terminalmage: This needs to be an `encode`, and we should be opening the file using a mode of `'wb'`.

salt/modules/nexus.py
```diff
@@ -466,13 +467,15 @@ def __save_artifact(artifact_url, target_file, headers):
             result['comment'] = checksum_comment
             return result
 
-    log.debug('Downloading: {url} -> {target_file}'.format(url=artifact_url, target_file=target_file))
+    log.debug('Downloading: %s -> %s', url=artifact_url, target_file=target_file)
```
terminalmage: The `url=` and `target_file=` need to be removed, they were only there as part of the `.format()`.

salt/modules/nftables.py
```diff
@@ -330,7 +332,7 @@ def save(filename=None, family='ipv4'):
     try:
         with salt.utils.files.fopen(filename, 'w+') as _fh:
             # Write out any changes
-            _fh.writelines(rules)
+            _fh.writelines(salt.utils.data.decode(rules))
```
terminalmage: This needs to be an `encode`, and we should be opening the file using a mode of `'wb'`.

salt/states/netyang.py
```diff
@@ -152,12 +153,16 @@ def managed(name,
...
-    log.debug('Creating temp file: {0}'.format(temp_file))
+    log.debug('Creating temp file: %s', temp_file)
     if 'to_dict' not in data:
         data = {'to_dict': data}
     data = [data]
     with salt.utils.files.fopen(temp_file, 'w') as file_handle:
-        salt.utils.yaml.safe_dump(salt.utils.json.loads(salt.utils.json.dumps(data)), file_handle, encoding='utf-8')
+        salt.utils.yaml.safe_dump(
+            salt.utils.json.loads(salt.utils.json.dumps(data)),
+            salt.utils.stringutils.to_str(file_handle),
```
terminalmage: This will break, we're passing a filehandle to the dumper. This should just be `file_handle`, not `salt.utils.stringutils.to_str(file_handle)`.

terminalmage (changes requested):

terminalmage (approved):

PR #45593, opened 2018-01-21
Add unicode_literals to modules/states [z]

rallytime:


salt/modules/zcbuildout.py
```diff
@@ -994,7 +997,7 @@ def buildout(directory='.',
 
         salt '*' buildout.buildout /srv/mybuildout
     '''
-    LOG.info('Running buildout in {0} ({1})'.format(directory, config))
+    LOG.info('Running buildout in {0} ({1})'.format(directory, config))  # future lint: disable=str-format-in-logging
```
terminalmage: This can just be converted to a %s substitution, unless there is something I am missing. Also, `str-format-in-logging` is not a "future lint" thing, it's actually already present in the newest salt-pylint, which is installed in Jenkins (but this check is currently just disabled in the lint job). So if you find any legit uses in the future, you can just use a normal `# pylint: disable=str-format-in-logging`, and it will be skipped once we enable the check in the jenkins lint job.

salt/modules/zonecfg.py
```diff
@@ -201,7 +202,11 @@ def _dump_cfg(cfg_file):
     '''Internal helper for debugging cfg files'''
     if __salt__['file.file_exists'](cfg_file):
         with salt.utils.files.fopen(cfg_file, 'r') as fp_:
-            log.debug("zonecfg - configuration file:\n{0}".format("".join(fp_.readlines())))
+            log.debug(
+                "zonecfg - configuration file:\n{0}".format(
+                    "".join(salt.utils.data.decode(fp_.readlines()))
+                )
```
terminalmage: This log message should be converted to a %s substitution.

terminalmage (changes requested):

terminalmage (changes requested):
The pylint disables should be `# pylint:`, not `# lint:`. Otherwise, this is GTM.

rallytime:
Oh, derp. I thought that looked weird. Thanks!

terminalmage (approved):

PR #45594, opened 2018-01-21
[oxygen] Tier three Unicode modules state updates

garethgreenaway:
### What does this PR do?
Tier three Unicode modules state updates

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/modules/telemetry.py
+++ b/salt/modules/telemetry.py
@@ -24,7 +24,7 @@ https://github.com/mongolab/mongolab-telemetry-api-docs/blob/master/alerts.md
 
 '''
 # Import Python libs
-from __future__ import absolute_import
+from __future__ import absolute_import, unicode_literals, print_function
 import logging
 
 # Import Salt libs
@@ -137,7 +137,7 @@ def get_alert_config(deployment_id, metric_name=None, api_key=None, profile="tel
             get_url = _get_telemetry_base(profile) + "/alerts?deployment={0}".format(deployment_id)
             response = requests.get(get_url, headers=auth)
         except requests.exceptions.RequestException as e:
-            log.error(str(e))
+            log.error(six.text_type(e))
             return False
 
         http_result = {}
@@ -180,7 +180,7 @@ def get_notification_channel_id(notify_channel, profile="telemetry"):
     notification_channel_id = _retrieve_channel_id(notify_channel)
 
     if not notification_channel_id:
-        log.info("{0} channel does not exist, creating.".format(notify_channel))
+        log.info("%s channel does not exist, creating.", notify_channel)
 
         # create the notification channel and cache the id
         post_url = _get_telemetry_base(profile) + "/notification-channels"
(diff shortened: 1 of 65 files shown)


salt/modules/telemetry.py
```diff
@@ -277,9 +277,9 @@ def create_alarm(deployment_id, metric_name, data, api_key=None, profile="teleme
         _update_cache(deployment_id, metric_name, response.json())
     else:
         log.error(
-            str('Failed to create alarm on metric: %s in deployment %s: payload: %s'),  # future lint: disable=blacklisted-function
-            salt.utils.stringutils.to_str(metric_name),
-            salt.utils.stringutils.to_str(deployment_id),
+            six.text_type('Failed to create alarm on metric: %s in deployment %s: payload: %s'),  # future lint: disable=blacklisted-function
+            salt.utils.stringutils.to_unicode(metric_name),
+            salt.utils.stringutils.to_unicode(deployment_id),
```
terminalmage: This log message should be changed. This was originally done intentionally since we're including the results of a json.dumps (which is a `str` type on both Python 2 and Python 3). This was done before our log handler was modified to gracefully handle str/unicode mismatches, so this kind of stuff is no longer necessary. This log message can be rewritten as:

```python
log.error(
    'Failed to create alarm on metric: %s in deployment %s: payload: %s'
    metric_name, deployment_id, salt.utils.json.dumps(post_body)
)
```

You may also need to remove the `salt.utils.stringutils` import as part of this change.

salt/modules/telemetry.py
```diff
@@ -371,11 +371,12 @@ def delete_alarms(deployment_id, alert_id=None, metric_name=None, api_key=None,
...
             response = requests.delete(delete_url, headers=auth)
             if metric_name:
-                log.debug("updating cache and delete {0} key from {1}".format(metric_name, deployment_id))
+                log.debug("updating cache and delete %s key from %s",
+                          metric_name, deployment_id)
                 _update_cache(deployment_id, metric_name, None)
 
         except requests.exceptions.RequestException as e:
-            log.error('Delete failed: {0}'.format(str(e)))
+            log.error('Delete failed: %s', six.text_type(e))
```
terminalmage: `six.text_type` is unnecessary here

salt/states/solrcloud.py
```diff
@@ -162,15 +162,15 @@ def collection(name, options=None, **kwargs):
             ret['comment'] = 'The collection "{0}" will be created.'.format(name)
             ret['pchanges'] = {
                 'old': None,
-                'new': str('options=') + new_changes  # future lint: disable=blacklisted-function
+                'new': six.text_type('options=') + new_changes  # future lint: disable=blacklisted-function
```
terminalmage: This should be reverted, we're concatenating to the result of a json.dumps, which is a `str()` type.

salt/states/solrcloud.py
```diff
@@ -162,15 +162,15 @@ def collection(name, options=None, **kwargs):
...
             }
             ret['result'] = None
         else:
             __salt__["solrcloud.collection_create"](name, options, **kwargs)
             ret['comment'] = 'The collection "{0}" has been created.'.format(name)
             ret['changes'] = {
                 'old': None,
-                'new': str('options=') + new_changes  # future lint: disable=blacklisted-function
+                'new': six.text_type('options=') + new_changes  # future lint: disable=blacklisted-function
```
terminalmage: This should be reverted, we're concatenating to the result of a json.dumps, which is a `str()` type.

terminalmage (changes requested):

terminalmage (approved):

PR #45595, opened 2018-01-21
Fix failing boto/network tests

terminalmage:
These failures were recently introduced but went undiscovered due to
some recent instability in the test suite (which has now been fixed).

This commit fixes these failing tests.


rallytime (approved):
🎉

Issue #45539, opened 2018-01-18
states/boto3_route53.rr_present doesn't currently allow for AliasTarget creation for the APEX of a domain


nullify005:
### Description of Issue/Question

When attempting to use boto3_route53.rr_present it currently isn't possible to create an A record for the APEX of a zone pointing to an AliasTarget such as an ELB.

The state tries to iterate through the ResourceRecords passed in to do all sorts of magic things with EC2 instances etc. and perform a lookup on ResourceRecords, however when making A records in this manner you don't need to supply any ResourceRecords according to the boto3 API documentation.

My current workaround is to simply do null checking around ResourceRecords in a few places to skip over the validation & remove the param if it's empty when talking to the AWS API

Given also that you can't really create AliasTarget entries without additional information from AWS like HostedZoneId & CanonicalDNSName etc. from ELB's or other resources (like CF or S3) I'm wondering whether there should be some additional features in the state (or modules) to fetch that information at runtime rather than having it statically present in the sls.

I've written modules & states to wrap various bits of the boto_* functionality in salt thus far because statically defining Id's etc. in a pillar just seems unnatural when you can use the API to get that information from the names - an important bit if you are frequently destroying the AWS resources and recreating them and all their internal Id's are changing.

Should the modules / states be a little more feature rich in the way that they handle such things?

For instance states/boto_elb.py supports cname attributes, but this only works for creating route53 cname entries, should it be modified to support other dns record types too?

I digress ...

Should I make a PR for this?

### Setup

Unfortunately you can't just call the state as is based on names of things, you have to fetch id's etc. so bear with me:

the sls:
```
elb APEX fake.news.domain:
    boto_helper.route53_elb_apex_present:
        - name: fake.news.domain
        - DomainName: fake.news.domain.
        - ELBName: my-awesome-elb
        - keyid: xxx
        - key: xxx
        - region: us-west-1
```

a custom state which backs the sls:
```
#!/usr/bin/env python

from exceptions import IndexError
import logging
from salt.exceptions import SaltInvocationError
import traceback
from boto.ec2 import elb

log = logging.getLogger(__name__)

def __get_elb_zone_id(name):
    ELB = elb.connect_to_region(os.environ.get('AWS_DEFAULT_REGION'))
    res = ELB.get_all_load_balancers(load_balancer_names=[name])
    return res[0].canonical_hosted_zone_name_id

# a wrapper which will create an APEX record based upon an ELBName
[...]

PR #45598, opened 2018-01-22
Patch around ResourceRecords needing to be present for AliasTarget

nullify005:
### What does this PR do?

Removes any validation or attempted usage of ResourceRecords when managing entry types don't require a ResourceRecord entry to be present.

In my context the example of this is an AliasTarget entry for an A record at the domain APEX which points to an ELB.

The boto3_route53 API specifically states that ResourceRecords should be omitted when attempting to make an A record using AliastTarget.

I suppose that other record types which don't require ResourceRecords will also benefit, though I have not tested any myself

https://github.com/saltstack/salt/issues/45539

### Previous Behavior

Any attempting make an A record using AliasTarget would fail because the code requires ResourceRecord(s) to be passed in even though they aren't needed.

### New Behavior

Simply skips over any ResourceRecord validation or usage if it happens to be None, this then allows for my entries to be made

### Tests written?

No

### Commits signed with GPG?

No

### Comments

As a side note the AliasTarget entries require additional information in order to be useful from a programmatic point of view.

I suppose you can statically define them, but it would really be preferable to refer to them by name and have the code lookup the missing bits based on the names (I'm referring to things like an ELB Canonical DNS name, or the ZoneId for the ELB for instance)

Should some of the boto3_route53 state or modules allow for setting of such things by name & fill in the blanks where appropriate?

I've written modules & wrappers for other boto* states in order to cater for this, but I'm not sure whether the states / modules should be capable of doing this themselves without wrapping them in more complex / complete states.


ryan-lane:
+1 looks good. some tests here would be nice, if there's already some basic tests for this module. If the module doesn't have any tests, it can be skipped for now (don't want to block a small PR on making tests for the module).

nullify005:
Hi @ryan-lane, yeah there doesn't appear to be any tests for this module or the states which sit on top

ryan-lane:
No worries, then, that's a long-term project and writing out the initial tests is a bit painful.

rallytime:
@nullify005 There's a merge conflict on this change, unfortunately. Can you rebase this and fix up the conflict?

nullify005:
Hi @rallytime, I've rebased to develop & resolved the merge conflict(s)

salt/states/boto3_route53.py
```diff
@@ -595,54 +595,55 @@ def rr_present(name, HostedZoneId=None, DomainName=None, PrivateZone=False, Name
...
-                if res:
-                    log.debug('Found %s %s for instance %s', instance_attr, res, instance.id)
-                    fixed_rrs += [_to_aws_encoding(res)]
+    if ResourceRecords:
+        for rr in ResourceRecords:
+            if rr.startswith('magic:'):
+                fields = rr.split(':')
+                if fields[1] == 'ec2_instance_tag':
+                    if len(fields) != 5:
+                        log.warning("Invalid magic RR value seen: '{}'.  Passing as-is.".format(rr))
```
rallytime: Can you switch this back to using `%s` notation instead of `.format()`? This will break the unicode changes we've been making recently.

salt/states/boto3_route53.py
```diff
@@ -595,54 +595,55 @@ def rr_present(name, HostedZoneId=None, DomainName=None, PrivateZone=False, Name
...
+                    if len(r) > 1:
+                        ret['comment'] = 'Multiple EC2 instances with tag {} == {} found'.format(
+                                tag_name, tag_value)
+                        log.error(ret['comment'])
+                        ret['result'] = False
+                        return ret
+                    instance = r[0]
+                    res = getattr(instance, instance_attr, None)
+                    if res:
+                        log.debug('Found {} {} for instance {}'.format(instance_attr, res, instance.id))
```
rallytime: Same comment here about the `%s` vs `format` notation.

rallytime (changes requested):
Hi @nullify005 - I have a couple of minor requests here.

salt/states/boto3_route53.py
nullify005: Just for the logging lines, or everywhere?

rallytime:
@nullify005 There's a lint error on this one now: https://jenkins.saltstack.com/job/PR/job/salt-pr-lint-n/18718/violations/file/salt/states/boto3_route53.py/

rallytime (approved):

Issue #22412, opened 2015-04-07
pkgrepo.managed for APT systems creates duplicate entries


skrobul:
I needed to pre-configure APT repository that is always trusted. According to [the manual](http://manpages.ubuntu.com/manpages/precise/ja/man5/sources.list.5.html), this is done by specifying `trusted=yes` option after `deb` in the name. It is valid statement and it works perfectly with APT. Problem is, that salt creates this repository on every run.

Example state:

```
"my repo":
  pkgrepo.managed:
    - name: "deb [trusted=yes] http://apt.example.net/ binary/"
```
## Bug description

```
staging:/srv/salt# cat /etc/apt/sources.list
deb http://ftp.us.debian.org/debian wheezy main
deb-src http://ftp.us.debian.org/debian wheezy main
staging:/srv/salt#

staging:/srv/salt# salt-call --local state.sls example.sls
[...]
----------
          ID: strategic repo
    Function: pkgrepo.managed
        Name: deb [trusted=yes] http://apt.example.net/ binary/
      Result: True
     Comment: Configured package repo 'deb [trusted=yes] http://apt.example.net/ binary/'
     Started: 13:19:39.065776
    Duration: 12555.233 ms
     Changes:
              ----------
              repo:
                  deb [trusted=yes] http://apt.example.net binary/
[...]
staging:/srv/salt# cat /etc/apt/sources.list
deb http:[kk//ftp.us.debian.org/debian wheezy main
deb-src http://ftp.us.debian.org/debian wheezy main
deb [trusted=yes] http://apt.example.net binary/staging:/srv/salt#

```

Notice missing `\n` on the end. It's not the biggest problem, but if you run the state again, this is what you get:

```
staging:/srv/salt# cat /etc/apt/sources.list
deb http://ftp.us.debian.org/debian wheezy main
deb-src http://ftp.us.debian.org/debian wheezy main
deb [trusted=yes] http://apt.example.net binary/

deb [trusted=yes] http://apt.example.net binary/deb [trusted=yes] http://apt.example.net binary/staging:/srv/salt#
```

This obviously results in non-working repo. Is there any other way to specify APT repository options to avoid this bug?

rallytime:
@skrobul Can you post the output of `salt --versions-report`?

skrobul:
@rallytime: Sure:
```
staging:# salt-call --versions-report
                  Salt: 2014.7.2
                Python: 2.7.3 (default, Mar 13 2014, 11:03:55)
[...]

PR #45603, opened 2018-01-22
Fix for duplicate entries with pkrepo.managed

andreaspe:
### What does this PR do?
Remove file attributes in pkrepo.managed prior comparing if the repo already exists.
This shall prevent creating duplicate sources.list entries on every salt run on debian based systems.

https://github.com/saltstack/salt/issues/22412

### Previous Behavior
pkgrepo.managed adds duplicate sources.list entry on every run

### New Behavior
pkrepo.managed sets up a repository only once

### Tests written?
No

### Commits signed with GPG?
No


terminalmage (approved):

terminalmage:
@rallytime this needs a backport to 2017.7

PR #45604, opened 2018-01-22
Back-port #45582 to 2017.7.3

rallytime:
Back-port #45582 to 2017.7.3


gtmanfred:
re-run py

salt/utils/thin.py
```diff
@@ -508,6 +508,7 @@ def gen_min(cachedir, extra_mods='', overwrite=False, so_mods='',
         'salt/utils/aggregation.py',
         'salt/utils/yamlloader.py',
         'salt/utils/event.py',
+        'salt/utils/state.py',
```
gtmanfred: This file doesn't exist in 2017.7.3

gtmanfred (changes requested):
everything else looks good

salt/utils/thin.py
rallytime: Good catch. I've removed that line.

gtmanfred (approved):

PR #45605, opened 2018-01-22
Boto iam role fix pseudo deepdiff

github-abcde:
### What does this PR do?
Fixes the method used to differentiate between supplied and existing configuration of an AWS IAM role.
Resolves some pylint errors.

None that I am aware of.

### Previous Behavior
Both supplied and existing configurations are sorted (as far as they contain arrays) and compared using `!=` in order to determine if there are configuration differences to resolve.
This caused unequalness where none was present when data from YAML was provided (which contains OrderedDicts).

### New Behavior
Supplied and existing configurations are compared with salt.utils.dictdiffer.deep_diff().

### Tests written?
No

### Commits signed with GPG?
Partial


ryan-lane:
+1 thanks for the PR!

rallytime (approved):

github-abcde:
Oops, looks like I made a mistake in removing the list-sorter. Amazon does return lists in an arbitrary (though fixed) order based upon the contents, which I missed in my initial tests. `salt.utils.dictdiffer.deep_diff` does care about list order, so now unequalness is created (based on the order of list-items provided by the user and by Amazon) where none is supposed to be. Please do not merge this until I have pushed a fix for this.

rallytime:
@github-abcde Thanks for the update. I changed the title of the PR to have `[WIP]` at the beginning. Feel free to change that back when you're ready.

github-abcde:
Sorry for not signing the last commit...really need to set signing as a default :)

Issue #44957, opened 2017-12-12
Various issues with dockermod and docker_image


brianthelion:
### Description of Issue/Question

`dockermod` and `docker_image` [recently underwent some changes](https://github.com/saltstack/salt/commit/64aa4fbbecde21baca42da8fa3cd936514d823df) that appear to have resulted in various issues. Below is an executable example.

### Setup
```bash
#! /bin/bash

LOG_LVL=error

ROOT_DIR=$(mktemp -d)
DOCKER_DIR=$(mktemp -d)
DOWNLOAD_DIR=$(mktemp -d)

python3 -m venv $ROOT_DIR > /dev/null
source $ROOT_DIR/bin/activate
git clone https://github.com/saltstack/salt $DOWNLOAD_DIR/salt > /dev/null
pip install $DOWNLOAD_DIR/salt > /dev/null
pip install docker > /dev/null

CONFIG_DIR=$ROOT_DIR/etc/salt
FILE_ROOT_BASE=$ROOT_DIR/srv/salt/base
FILE_ROOT_FOO=$ROOT_DIR/srv/salt/foo
PILLAR_ROOT_BASE=$ROOT_DIR/srv/pillar/base
PILLAR_ROOT_FOO=$ROOT_DIR/srv/pillar/foo

mkdir -p $CONFIG_DIR
cat <<EOF > $CONFIG_DIR/minion
root_dir: $ROOT_DIR
file_roots:
  base:
    - $FILE_ROOT_BASE
  foo:
    - $FILE_ROOT_FOO
pillar_roots:
  base:
    - $PILLAR_ROOT_BASE
  foo:
    - $PILLAR_ROOT_FOO
pillar_source_merging_strategy: none
top_file_merging_strategy: none
pillarenv_from_saltenv: True
EOF

cat <<EOF > $CONFIG_DIR/minion_id
my_minion
EOF

mkdir -p $PILLAR_ROOT_BASE
cat <<EOF > $PILLAR_ROOT_BASE/top.sls
base:
  'my_minion':
    - base_pillar
EOF

mkdir -p $PILLAR_ROOT_FOO
cat <<EOF > $PILLAR_ROOT_FOO/top.sls
foo:
  'my_minion':
    - foo_pillar
EOF

cat <<EOF > $PILLAR_ROOT_BASE/base_pillar.sls
A:
 B: 'curl'
EOF

cat <<EOF > $PILLAR_ROOT_FOO/foo_pillar.sls
A:
[...]

PR #45606, opened 2018-01-22
Fix bug affecting salt-ssh when root_dir differs from the default

terminalmage:
H/T https://github.com/saltstack/salt/issues/44957#issuecomment-356122811

The join on the line below will cause the path to be incorrect due to leading slashes.


rallytime (approved):

rallytime:
re-run py

gtmanfred:
re-run py

PR #45613, opened 2018-01-22
cmdmod: Remove more unnecessary environment variable encoding

terminalmage:
This causes errors in subprocess.Popen in Python 3 since it expects the
environment to be strings and not bytes.

The encoding that was done before was likely done to keep unicode types
out of the env dict on Python 2, since encoding produces a `str` type on
Python 2, but it doesn't do that on Python 3.


salt/modules/cmdmod.py
```diff
@@ -424,28 +423,24 @@ def _get_stripped(cmd):
                 env_cmd = ('su', runas, '-c', sys.executable)
             else:
                 env_cmd = ('su', '-s', shell, '-', runas, '-c', sys.executable)
-            env_encoded = subprocess.Popen(
+            env_bytes = salt.utils.stringutils.to_bytes(subprocess.Popen(
```
damon-atkins: Do not have an issue with this change, however reading the code it looks like this should be
```
if runas and reset_system_locale and not salt.utils.platform.is_windows()
    # Save the original command before munging it
```
Not see a lot of value in capturing the current environment, before running a command.

terminalmage: We don't need a windows check because we will have exited early in the `if` block above that one (lines 385-397) if we're on windows and an alternate user has been specified with `runas`.

Similarly, I disagree with checking for `reset_system_locale`, it doesn't really have anything to do with this.

> Not see a lot of value in capturing the current environment, before running a command.

I don't understand the confusion. Maybe you misread? The whole point of the block preceding this change is to get the `runas` user's environment, so that we can pass it to `salt.utils.timed_subprocess.TimedProc()` and the command can be run with that user's environment. We manage setting the euid/egid/umask via passing a [preexec_fn](https://docs.python.org/2/library/subprocess.html#subprocess.Popen), but that does nothing to the environment, hence the above block of code to obtain the user's environment so it can be passed to the [subprocess.Popen()](https://docs.python.org/2/library/subprocess.html#subprocess.Popen) constructor via the `env` argument.

damon-atkins: I will raise it in a Issue.

rallytime (approved):

PR #45618, opened 2018-01-22
Fix salt.utils.path.os_walk for certain Python 2 versions

terminalmage:
This helper was designed to ensure that the paths returned were all
unicode types, but on a couple platforms having the path be a unicode
type causes a traceback deep in the stdlib within posixpath.

To work around this, this commit makes the helper a generator that
simply decodes each tuple returned from os.walk as we iterate.

### DIFF
--- a/salt/fileclient.py
+++ b/salt/fileclient.py
@@ -24,6 +24,7 @@ import salt.loader
 import salt.payload
 import salt.transport
 import salt.fileserver
+import salt.utils.data
 import salt.utils.files
 import salt.utils.gzip_util
 import salt.utils.hashutils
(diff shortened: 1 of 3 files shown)


gtmanfred:
Go Go Jenkins!

gtmanfred:
@terminalmage seems like https://jenkins.saltstack.com/job/PR/job/salt-pr-linode-ubuntu14-n/19243/testReport/junit/unit.fileserver.test_roots/RootsLimitTraversalTest/test_limit_traversal/ may be an actual broken test.

terminalmage:
@gtmanfred This has to do with the `del dirs[:]` line in fileclient.py, which is used to limit traversal by on-the-fly clearing the `dirs` list so that the `os.walk()` generator does not yield any files from those directories in later iterations.

I think the only way around this would be to just use regular `os.walk()` here and manually decode each root and set of files.

terminalmage:
@gtmanfred should be fixed now

gtmanfred (approved):

PR #45621, opened 2018-01-22
salt.utils.json: only encode data on PY2

terminalmage:
We need a str type in both Python 2 and 3, and encoding produces a bytes type on Python 3. This makes the encoding conditional, only doing so in Python 2.

EDIT: for clarification, I mean that the strings in the data structure being dumped must be `str` in both Python 2 and 3. So this PR will make sure that unicode strings in the data structure being dumped are converted to `str` types.

### DIFF
--- a/salt/utils/json.py
+++ b/salt/utils/json.py
@@ -88,11 +88,13 @@ def dump(obj, fp, **kwargs):
     '''
     .. versionadded:: Oxygen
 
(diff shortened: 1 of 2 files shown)


gtmanfred (approved):

terminalmage:
I'm going to add some unit tests, so hold off on merging.

terminalmage:
OK, tests added

Issue #45637, opened 2018-01-23
windows minion 2016.11.9. Warning and Errors


frogunder:
### Description of Issue/Question
See some Warnings and Errors when running `salt \* cmd.run "C://salt/salt-call --local sys.doc none"`
```
    [WARNING ] State for file: C:\\salt\test.txt - Neither 'source' nor 'contents' nor 'contents_pillar' nor 'contents_grains' was defined, yet 'replace' was set to 'True'. As there is no source to replace the file with, 'replace' has been set to 'False' to avoid reading the file unnecessarily.
    [ERROR   ] Exception raised when processing __virtual__ function for win_dsc. Module will not be loaded local variable 'proc' referenced before assignment
    [WARNING ] salt.loaded.int.module.win_dsc.__virtual__() is wrongly returning `None`. It should either return `True`, `False` or a new name. If you're the developer of the module 'win_dsc', please fix this.
    [ERROR   ] Exception raised when processing __virtual__ function for win_pki. Module will not be loaded local variable 'proc' referenced before assignment
    [WARNING ] salt.loaded.int.module.win_pki.__virtual__() is wrongly returning `None`. It should either return `True`, `False` or a new name. If you're the developer of the module 'win_pki', please fix this.
    [ERROR   ] Exception raised when processing __virtual__ function for win_psget. Module will not be loaded local variable 'proc' referenced before assignment
    [WARNING ] salt.loaded.int.module.win_psget.__virtual__() is wrongly returning `None`. It should either return `True`, `False` or a new name. If you're the developer of the module 'win_psget', please fix this.
```

### Versions Report
    Salt Version:
               Salt: 2016.11.9
     
    Dependency Versions:
               cffi: 1.10.0
           cherrypy: 7.1.0
           dateutil: 2.5.3
          docker-py: Not Installed
              gitdb: 0.6.4
          gitpython: 2.0.8
              ioflo: 1.5.5
             Jinja2: 2.9.4
            libgit2: Not Installed
            libnacl: 1.4.5
           M2Crypto: Not Installed
               Mako: 1.0.4
       msgpack-pure: Not Installed
     msgpack-python: 0.4.8
       mysql-python: Not Installed
          pycparser: 2.17
           pycrypto: 2.6.1
       pycryptodome: Not Installed
             pygit2: Not Installed
             Python: 2.7.12 (v2.7.12:d33e0cf91556, Jun 27 2016, 15:24:40) [MSC v.1500 64 bit (AMD64)]
[...]

PR #44193, opened 2017-10-20
Fix reg.py for use with LGPO module

twangboy:
### What does this PR do?
Modifies the `salt.utils.to_unicode` function to return `None` if passed `None`

Makes sure the `vdata` value type matches what is required by the type specified in `vtype`. The LGPO module was passing a unicode string 0 (`u'0'`) which was causing the RegSetValueEx function to fail because it was expecting an int type value.

This casts vdata to the following:
int for REG_DWORD and REG_QWORD
str for REG_SZ and REG_EXPAND_SZ
bin for REG_BINARY
list for REG_MULTI_SZ

Fixes many Unicode registry issues.
Uses `salt.utils.to_unicode` instead of the local `_unicode_to_mcbs` and `_mcbs_to_unicode` functions
Was using a mixture of _winreg and win32api... moved to win32api for better unicode handling

Tested with Unicode values
Tested with Unicode value names

Fixes `reg.py` state to handle Unicode value names and values
Can't pass Unicode values in the `name` portion of a state for some reason

### What issues does this PR fix or reference?
https://github.com/saltstack/salt/issues/44192

### Previous Behavior
LGPO would fail

### New Behavior
Now it works

### Tests written?
No

Issue #36031, opened 2016-09-02
Windows PY3 resolved issues outstanding for PY2, non-salt


damon-atkins:
### Description of Issue/Question

This is a list of issues which exist in python 2 which have been resolved in python  3.
Feel free to post any missing fixes.
- win reg api under PY2 is non-Unicode check a Unicode registry.  PY3 as full Unicode API support
- Fix islink and effectively os.walk   https://bugs.python.org/issue1578269
- Speed improvements to os.walk https://bugs.python.org/issue23605

Ch3LL:
ping @twangboy fyi. Also do you have an issue that already is tracking these issue. Just trying to figure out if its better to tack these issues here or if you already have somewhere else you are using to track. Thanks

damon-atkins:
Just waking this up again.
[damon_lang_test_reg.zip](https://github.com/saltstack/salt/files/1667403/damon_lang_test_reg.zip)
The above file can be used to test windows registry and salt.
Their is a large difference if the result between PY2 and PY3. 

e.g. The below is ZDamon Test0发~送时\间1™自动/化平台
```
python3 -B salt\salt\utils\pkg\win.py list system > win_pkg.out
    "zsalt language test\\\\zdamon test0\u53d1~\u9001\u65f6\\\u95f41\u2122\u81ea\u52a8/\u5316\u5e73\u53f0\\\\soft": [
        "0.NotANumber.3"
```

```
python2 -B salt\salt\utils\pkg\win.py list system > win_pkg.out
    "zsalt language test\\\\zdamon test0?~??\\?1\u2122??/???\\\\soft": [
        "0.NotANumber.3"
```

@thatch45   @twangboy  can we not release a community addition of PY2 and only have a Enterprise version of PY2.   If people complain too much then you can always release community addition of PY2.

thatch45:
In the long run yes, we would only ship a PY3 windows install

damon-atkins:
Python 3.6 14 months old and have released 3.6.5 vs the 3.5.3 ship with 2018.3rc1

https://docs.python.org/3.6/whatsnew/changelog.html
https://www.python.org/downloads/release/python-365/
PEP 528, Change Windows console encoding to UTF-8
PEP 529, Change Windows filesystem encoding to UTF-8
[...]

PR #45638, opened 2018-01-23
Win fix shell info

twangboy:
### What does this PR do?
Fixes an issue with `cmd.shell_info` where it wasn't detecting powershell on the system. Probably caused by changes made in https://github.com/saltstack/salt/pull/44193 as all values were being converted to unicode. This fix only converts string values to unicode (`REG_SZ, REG_EXPAND_SZ, REG_MULTI_SZ`)

https://github.com/saltstack/salt/issues/45637

### Tests written?
No

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/modules/reg.py
+++ b/salt/modules/reg.py
@@ -296,9 +296,15 @@ def list_values(hive, key=None, use_32bit_registry=False, include_default=True):
             value = {'hive':   local_hive,
                      'key':    local_key,
                      'vname':  _to_mbcs(vname),
-                     'vdata':  _to_mbcs(vdata),
                      'vtype':  registry.vtype_reverse[vtype],
                      'success': True}
+            # Only convert text types to unicode
+            if vtype == win32con.REG_MULTI_SZ:
+                value['vdata'] = [_to_mbcs(i) for i in vdata]
+            elif vtype in [win32con.REG_SZ, win32con.REG_EXPAND_SZ]:
+                value['vdata'] = _to_mbcs(vdata)
+            else:
+                value['vdata'] = vdata
             values.append(value)
     except pywintypes.error as exc:  # pylint: disable=E0602
         log.debug(exc)
--- a/salt/modules/reg.py
+++ b/salt/modules/reg.py
@@ -372,11 +372,14 @@ def read_value(hive, key, vname=None, use_32bit_registry=False):
             # RegQueryValueEx returns and accepts unicode data
             vdata, vtype = win32api.RegQueryValueEx(handle, local_vname)
             if vdata or vdata in [0, '']:
+                # Only convert text types to unicode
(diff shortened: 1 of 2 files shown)


twangboy:
@lomeroe Do you foresee any issues with this and the LGPO module?

lomeroe:
No, I don't think so.  The reg module is used on some of the "static" policies defined in lgpo, but almost all of those are writing 0/1 into a REG_SZ.  I don't see those that aren't doing that being affected by this either (legal notice text, legal notice caption, etc).

rallytime (approved):

salt/modules/reg.py
```diff
@@ -372,11 +378,14 @@ def read_value(hive, key, vname=None, use_32bit_registry=False):
             # RegQueryValueEx returns and accepts unicode data
```
damon-atkins: Conversion not required at all.....???

# RegQueryValueEx returns and accepts unicode data
Therefore only PY2 strings from arguments should be converted to Unicode.  As the result from RegQueryValueEx  should be Unicode.   I thought it was worked out that the win32reg returned mbcs and RegQueryValueEx return unicode

twangboy: It accepts and returns Unicode data... we only have to convert the return to mbcs to display properly on whatever encoding the system is using. Since it's not really supported that well in Py2, this seems to give the best results.

damon-atkins: > we only have to convert the return to mbcs to display properly on whatever encoding the system is using. 

It needs to return Unicode to the salt modules calling it PY2 and PY3, otherwise it will break win_pkg and other things.

When 2016.11 was released reg.py always to return Unicode, even it the parameters passed were PY2 strings.

The old mbcs function was used under win32reg to convert the result to Unicode, because the output from win32reg was mbcs and it needed conversion to Unicode.

You now have changed it to to convert Unicode to mbcs which is wrong.  !!!!!!!!!!

twangboy: Can you provide an example software package that is broken by this change so I can test this? I did quite a bit of testing on this to make sure it worked properly.

damon-atkins:
See the reg file which is attached to #36031 and if the  PY2 and PY3 results match, then their is no issue.

damon-atkins:
OK Re-visit this.
`# RegQueryValueEx returns and accepts unicode data`, caused me to thing `The above takes Unicode and coverts it to Unicode.` as said in early comment.
I suggest its updated to:
`# RegQueryValueEx accepts Unicode data returns PY2 mbcs encoded data and PY3 Unicode data` 

_to_mbcs() needs a better name. e.g. _mbcs_to_unicode or _from_mbcs()  Because variable contents is mbcs format/encoding already under PY2.

The above to items

PR #45640, opened 2018-01-23
Fix file.symlink backupname with basename path + file.move

### LINKED ISSUE #45631: file.symlink backupname with relative name has unclear behaviour
file.symlink's use of backupname is somewhat unclear in the docs, and the behaviour is also not intuitive at first when debugging.

Case: symlink name to target - if "name" exists, make a backup.

I see examples where the user _expects_ that where backupname == "filename" that file will be placed relative to the original symlinks path.

This is not the case:
A relative path or filename in "backupname" will be **_relative to the home directory of the user running salt_**, mostly "root".

Two cases:

1. The user overwrites a file/dir in roots home dir
2. The user gets an error if the root filesystem is on another file system than the file being backed up

A user trying to symlink in one file system, having av backupname with a relative path or just a file name will get the error:

```
salt-call state.single file.symlink /tmp/sas/bat/mat/trigger target=/tmp/sas/foo/bar/trigger backupname=trigger.old
...
          ID: envid-symlink
    Function: file.symlink
        Name: /tmp/sas/bat/mat/trigger
      Result: False
                File "/usr/lib/python2.7/site-packages/salt/state.py", line 1765, in call
                  **cdata['kwargs'])
                File "/usr/lib/python2.7/site-packages/salt/loader.py", line 1705, in wrapper
                  return f(*args, **kwargs)
                File "/usr/lib/python2.7/site-packages/salt/states/file.py", line 975, in symlink
                  os.rename(name, backupname)
              OSError: [Errno 18] Invalid cross-device link
     Started: 14:06:13.242857
    Duration: 1.215 ms
     Changes:
```

This error is somewhat counter intuitive - it fails since file.symlink tries to create a backup of "target" to ~root/trigger.old - which is on another file system, thus returning "Invalid cross-device link" from os.rename().

It is also possibly dangerous, since if the symlink happens on the same file system as root's and the backupname is an actual file/directory in roots home directory, and with force:True, it will be removed.

The docs do not mention that an absolute path is required.

It is possible to argue that defining a simple file name is something one could expect to work - the file name should only go to the same directory as "target".

### Setup
```
mkdir -p /tmp/sas/foo/bar/trigger /tmp/sas/bat/mat/trigger
```

### Steps to Reproduce Issue
Will _work_ if the "target" is on the same file system as roots home directory file system, but should not.

Will fail if root's home directory is on another file system than "target", but with an unclear error message.
```
salt-call state.single file.symlink /tmp/sas/bat/mat/trigger target=/tmp/sas/foo/bar/trigger backupname=trigger.old
```

### Versions Report
See that this applies to develop as pr 23.01.2018

```
Salt Version:
           Salt: 2016.11.8

Dependency Versions:
           cffi: Not Installed
       cherrypy: Not Installed
       dateutil: 1.5
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.7.2
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: 0.21.1
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.4.6
   mysql-python: Not Installed
      pycparser: Not Installed
       pycrypto: 2.6.1
   pycryptodome: 3.4.3
         pygit2: Not Installed
         Python: 2.7.5 (default, May  3 2017, 07:55:04)
   python-gnupg: Not Installed
         PyYAML: 3.11
          PyZMQ: 15.3.0
           RAET: Not Installed
          smmap: Not Installed
        timelib: Not Installed
        Tornado: 4.2.1
            ZMQ: 4.1.4

System Versions:
           dist: redhat 7.4 Maipo
        machine: x86_64
        release: 3.10.0-693.5.2.el7.x86_64
         system: Linux
        version: Red Hat Enterprise Linux Server 7.4 Maipo
```

sastorsl:
### What does this PR do?
Updates the salt/states/file.py - file.symlink - function.

Tests if backupname is an absolute path or a single file/dir name.
If it's a single file/dir name it will be placed relative to the parent
directory of the named symlink.

Updated sys.doc to underline that an abspath or single file/dir must be used.

Fixes #45631

### Previous Behavior
A non-absolute path for backupname would be relative to the HOME directory of the user running salt, which is counter-intuitive.

os.rename() in backupname would fail with an exception if the backup target was on another file system.

### New Behavior
backupname must be an absolute path or a single file/dir name. In the latter case the file/dir will be placed relative to the parent dir of the symlink.

Change from os.rename to "file.move" to prevent errors if the backupname
is on another filesystem (relevant to some flavours of *nix) to make for
more consistent behaviour.

### Tests written?

No

Tested locally:
- With backupname as an abspath, single file/dir, relative path (should fail), confirmed expected behaviour.
- With / without force option

### Commits signed with GPG?

No

### DIFF
--- a/salt/states/file.py
+++ b/salt/states/file.py
@@ -1411,7 +1411,10 @@ def symlink(
                 __salt__['file.move'](name, backupname)
             except Exception as exc:
                 ret['changes'] = {}
-                log.debug(traceback.format_exc())
+                log.debug(
+                    'Encountered error renaming %s to %s',
+                    name, backupname, exc_info=True
+                )
                 return _error(ret, ('Unable to rename {0} to backup {1} -> '
                                     ': {2}'.format(name, backupname, exc)))
         elif force:
--- a/salt/states/file.py
+++ b/salt/states/file.py
@@ -1257,6 +1257,9 @@ def symlink(
         renamed to the backupname. If the backupname already
         exists and force is False, the state will fail. Otherwise, the
         backupname will be removed first.
+        An absolute path OR a basename file/directory name must be provided.
+        The latter will be placed relative to the symlink destination's parent
+        directory.
 
     makedirs
         If the location of the symlink does not already have a parent directory
@@ -1388,15 +1391,29 @@ def symlink(
     elif os.path.isfile(name) or os.path.isdir(name):
         # It is not a link, but a file or dir
         if backupname is not None:
+            if not os.path.isabs(backupname):
+                if backupname == os.path.basename(backupname):
+                    backupname = os.path.join(
+                        os.path.dirname(os.path.normpath(name)),
+                        backupname)
+                else:
+                    return _error(ret, (('Backupname must be an absolute path '
+                                         'or a file name: {0}').format(backupname)))
             # Make a backup first
             if os.path.lexists(backupname):
                 if not force:
-                    return _error(ret, ((
-                                            'File exists where the backup target {0} should go'
-                                        ).format(backupname)))
+                    return _error(ret, (('Symlink & backup dest exists and Force not set.'
+                                         ' {0} -> {1} - backup: {2}').format(
+                                             name, target, backupname)))
                 else:
                     __salt__['file.remove'](backupname)
-            os.rename(name, backupname)
+            try:
+                __salt__['file.move'](name, backupname)
+            except Exception as exc:
+                ret['changes'] = {}
+                log.debug(traceback.format_exc())
+                return _error(ret, ('Unable to rename {0} to backup {1} -> '
+                                    ': {2}'.format(name, backupname, exc)))
         elif force:
             # Remove whatever is in the way
             if __salt__['file.is_link'](name):


salt/states/file.py
```diff
@@ -1388,15 +1391,29 @@ def symlink(
...
+                                         ' {0} -> {1} - backup: {2}').format(
+                                             name, target, backupname)))
                 else:
                     __salt__['file.remove'](backupname)
-            os.rename(name, backupname)
+            try:
+                __salt__['file.move'](name, backupname)
+            except Exception as exc:
+                ret['changes'] = {}
+                log.debug(traceback.format_exc())
```
terminalmage: This would be better-written as:

```python
                log.debug(
                    'Encountered error renaming %s to %s',
                    name, backupname, exc_info=True
                )
```

terminalmage (changes requested):
Just one small change that I would make for clarity. Also, we're going to want to get a test in for this.

salt/states/file.py
sastorsl: Will have a look ,thx

sastorsl: Added a commit @terminalmage 
Had a look at tests, but I've not written any before so I think I must either peak over your(?) shoulder or get some help.

rallytime:
HI @sastorsl  - have you had a chance to write a test for this? We'll need one written here before we can merge this.

rallytime:
Since we didn't hear back here, I am going to close this. Once that test is written, we will happily open this again, or you can submit a new PR. We can of course continue the discussion below, as well. Thanks!

rallytime:
@sastorsl Yes, that is looking good. :)

PR #45648, opened 2018-01-23
Add onlyifempty to group.absent

### LINKED ISSUE #45311: Conditional group.absent - Only remove when group is not in use
### group.absent should have an optional condition - remove only if not in use
If such an option existed you could safely use group.absent where many users might share the same group. If the group is in use by any user it should abort successfully _if_ this option was turned on.

salt '*' group.absent foo onlyifempty=True

The default could be discussed, but False will adhere to the current usage.

### Setup
N/A

### Steps to Reproduce Issue
N/A

### Versions Report
N/A

sastorsl:
### What does this PR do?
Adds the possibility to have a group.absent state conditionally remove a group
only if there are no members in the group.

Fixes #45311

### Previous Behavior
A group would be removed regardless if it had any members.

### New Behavior
An _option_ to only remove the group if it has no members has been added.

The state will return a True result if this is the case.

### Tests written?

No

Tested locally

### Commits signed with GPG?

No

### DIFF
--- a/salt/states/group.py
+++ b/salt/states/group.py
@@ -275,7 +275,7 @@ def present(name,
     return ret
 
 
-def absent(name, onlyifempty=False):
+def absent(name, only_if_empty=False):
     '''
     Ensure that the named group is absent
 
@@ -283,8 +283,8 @@ def absent(name, onlyifempty=False):
         name (str):
             The name of the group to remove
 
-        onlyifempty (bool):
-            If True only delete if there are no members in the group
+        only_if_empty (bool):
+            Delete only if there are no members in the group
 
     Example:
 
@@ -298,8 +298,8 @@ def absent(name, onlyifempty=False):
 
         # Removes the local group `db_admin`, but only if there are no more members
         db_admin:
-          group.absent
-            - onlyifempty: True
+          group.absent:
+            - only_if_empty: True
     '''
     ret = {'name': name,
            'changes': {},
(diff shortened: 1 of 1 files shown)


salt/states/group.py
```diff
@@ -275,21 +275,31 @@ def present(name,
     return ret
 
 
-def absent(name):
+def absent(name, onlyifempty=False):
```
rallytime: Can we change this name to be more pythonic/stateful? At the very least some underscores could be added, `only_if_empty`, but there's probably a better stateful name we could choose. But I am not coming up with anything good. What do you think @terminalmage?

rallytime (commented):
I have one question here, and I'd also like @terminalmage to take a look at this idea.

In addition, some tests need to be written for this change.

salt/states/group.py
terminalmage: I think `only_if_empty` is probably best here. We just need to make sure that all references to the new variable name are updated in the code below.

salt/states/group.py
```diff
@@ -299,10 +309,20 @@ def absent(name):
...
+                ret['result'] = True
+                ret['comment'] = ('Group {0} will not be removed since members are '
+                                  'present and onlyifempty is set').format(name)
+            else:
+                ret['result'] = None
+                ret['comment'] = 'Group {0} is set for removal'.format(name)
             return ret
-        ret['result'] = __salt__['group.delete'](name)
+        if onlyifempty and grp_info['members']:
+            ret['comment'] = ('Group {0} will not be removed since members are '
```
terminalmage: We shouldn't be using future tense here since the state will have completed. Instead we should say "was not removed".

salt/states/group.py
```diff
@@ -299,10 +309,20 @@ def absent(name):
     if grp_info:
         # Group already exists. Remove the group.
         if __opts__['test']:
-            ret['result'] = None
-            ret['comment'] = 'Group {0} is set for removal'.format(name)
+            if onlyifempty and grp_info['members']:
+                ret['result'] = True
+                ret['comment'] = ('Group {0} will not be removed since members are '
```
terminalmage: Using the future tense here implies that we are definitely going to make changes, we should be using the conditional tense here. Instead of "will not be removed" it should be "would not be removed".

terminalmage (changes requested):
Just a few changes, and we need a test case.

salt/states/group.py
sastorsl: Didn't really like the name either, but took one to move on. Very open for suggestions.

rallytime:
@sastorsl Same question here - once you write a test for this, we can get this merged in.

sastorsl:
@rallytime I've been pressed for time (who isn't...), but the major issue is that I've never written tests in this project before, so stumbling along on my own takes more time. I would really like to finish my contribution(s), so any pointers would be of great help.

rallytime:
Hey @sastorsl - yeah for sure, I totally understand! It looks like you've started some tests in your other PR, #45640, but in case you still want some pointers, there's some docs on writing tests for salt [here](https://docs.saltstack.com/en/latest/topics/development/tests/index.html) and docs for unit tests specifically are [here](https://docs.saltstack.com/en/latest/topics/development/tests/unit.html#unit-tests). Please let us know if you have any questions!

rallytime:
I am going to close this PR for now until the tests can be written. We are happy to open this again when you're ready. Or feel free to open a new PR. Whichever is easier for you. :)

Thank you!

PR #45656, opened 2018-01-24
add label_selector to kubernetes module list functions

deuscapturus:
### What does this PR do?

Add ability to perform label_selector filters on list functions in the kubernetes module

None,  The functions just aren't very useful if you can't use label selectors.

### Tests written?

No

### Commits signed with GPG?

No


salt/modules/kubernetes.py
```diff
@@ -236,18 +236,18 @@ def nodes(**kwargs):
...
     CLI Examples::
 
-        salt '*' kubernetes.node name='minikube'
+        salt '*' kubernetes.node name='minikube' label_selector='io.kompose.service=nginx'
     '''
     _setup_conn(**kwargs)
     try:
         api_instance = kubernetes.client.CoreV1Api()
-        api_response = api_instance.list_node()
+        api_response = api_instance.list_node(label_selector=label_selector)
```
isbm: In Salt default params are `None`, not an empty String. That said, it should be `None`. Now, from the description of the kubernetes CoreV1Api class:
```
label_selector: A selector to restrict the list of returned 
objects by their labels. Defaults to everything.
```
Therefore I would write it that way:

```python
def node(name, **kwargs):
    ....
    api_response = api_instance.list_node(**{'label_selector': kwargs.get('label_selector', '')})
```

So then it goes clean and no signature is changed. I don't know whether `label_selector` _has to be_ an empty string or `None` would also work there. In case `None` is fine, then `kwargs.get('label_selector)` is sufficient.

salt/modules/kubernetes.py
```diff
@@ -381,19 +381,19 @@ def namespaces(**kwargs):
         _cleanup()
 
 
-def deployments(namespace='default', **kwargs):
+def deployments(namespace='default', label_selector='', **kwargs):
```
isbm: Same as above. No need to change the API signature here. Please reuse `kwargs` for this purpose.

salt/modules/kubernetes.py
```diff
@@ -410,19 +410,19 @@ def deployments(namespace='default', **kwargs):
         _cleanup()
 
 
-def services(namespace='default', **kwargs):
+def services(namespace='default', label_selector='', **kwargs):
```
isbm: No signature change, please.

salt/modules/kubernetes.py
```diff
@@ -438,19 +438,19 @@ def services(namespace='default', **kwargs):
         _cleanup()
 
 
-def pods(namespace='default', **kwargs):
+def pods(namespace='default', label_selector='', **kwargs):
```
isbm: ...and here too.

isbm (changes requested):
Please change my remarks and also add `unicode_literals` in the first import line: `from __future__ import absolute_import`.

Thanks!

salt/modules/kubernetes.py
deuscapturus: The kubernetes client methods fail when label_selector is set to `None` or `False` the only unset value that will work is `""`.  I didn't want to add label_selector to `**kwargs` because that's passed into the client connection setup, but after looking it doesn't look like it will be a problem.

isbm (changes requested):
@deuscapturus could you please rephrase your last commit name? Because "refactor for merge approval" is [not really helping others](https://chris.beams.io/posts/git-commit/) when you will look at this few years later from now on. "Refactor for keeping function signatures untouched" would be probably a better idea.

And I have an idea for your concern, which is in fact quite valid. We have `**kwargs` that are containing both:

a) params for connection — always the same
b) "random" API params — differ from time to time

Then I would just pop a) out from `kwargs` in `_setup_conn` function (just replace `.get(...)` with `.pop(...)` in there). So connection params will disapper from `kwargs` right after you call it, so it will be safe to be thrown at any API function. Surely you might check their values, but that can be done in another function and then reused.

rallytime:
Hi @deuscapturus - Have you had a chance to review the comments on this PR?

rallytime:
@deuscapturus Any updates on this one?

PR #45660, opened 2018-01-24
Support generators/iterators in the encode/decode helpers

terminalmage:
This will convert iterable types which are not mapping types, tuples,
lists, or strings into lists and return encoded/decoded versions of
those lists. This allows for more graceful handling of these types if
one forgets to first convert the sequence to a list before invoking one
of these helpers.

### DIFF
--- a/salt/utils/data.py
+++ b/salt/utils/data.py
@@ -83,7 +83,9 @@ def decode(data, encoding=None, errors='strict', preserve_dict_class=False, pres
         try:
             return salt.utils.stringutils.to_unicode(data, encoding, errors)
         except TypeError:
-            return data
+            return data \
+                if not isinstance(data, collections.Iterable) \
+                else decode_list(list(data), encoding, errors, preserve_dict_class, preserve_tuples)
 
 
 def decode_dict(data, encoding=None, errors='strict', preserve_dict_class=False, preserve_tuples=False):
@@ -101,7 +103,8 @@ def decode_dict(data, encoding=None, errors='strict', preserve_dict_class=False,
             try:
                 key = salt.utils.stringutils.to_unicode(key, encoding, errors)
             except TypeError:
-                pass
+                if isinstance(key, collections.Iterable):
+                    key = decode_list(list(key), encoding, errors, preserve_dict_class, preserve_tuples)
 
         if isinstance(value, list):
             value = decode_list(value, encoding, errors, preserve_dict_class, preserve_tuples)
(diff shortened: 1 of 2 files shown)


salt/utils/data.py
```diff
@@ -101,7 +103,8 @@ def decode_dict(data, encoding=None, errors='strict', preserve_dict_class=False,
             try:
                 key = salt.utils.stringutils.to_unicode(key, encoding, errors)
             except TypeError:
-                pass
+                if isinstance(key, collections.Iterable):
+                    key = decode_list(list(key), encoding, errors, preserve_dict_class, preserve_tuples)
```
terminalmage: This bit here might be questionable, since generators are hashable, while lists are not. So, in the event that someone had a dictionary with keys that were generators, and they called either this (or its encode counterpart below), it would cause an exception. Thoughts? Perhaps automagic handling of generators should be toggleable via an argument to the function? If so, I would think we'd still want the default behavior be to turn this on by default.

rallytime:
@saltstack/team-core What do you all think about this change?

terminalmage:
Keep in mind I don't think this PR is 100% necessary. This was inspired by https://github.com/saltstack/salt-jenkins/issues/789, because we were running `salt.utils.data.encode()` on a generator and that was causing the test failures.

terminalmage:
This is not urgent, and the underlying code has diverged. Closing for now, I may reopen if it is necessary.

tests/unit/utils/test_data.py
```diff
@@ -311,6 +311,50 @@ def test_encode_fallback(self):
         with patch.object(builtins, '__salt_system_encoding__', 'CP1252'):
             self.assertEqual(salt.utils.data.encode('Ψ'), _b('Ψ'))
 
+    menu_item = ['спам', 'спам', 'спам', 'яйца', 'спам']
```
DmitryKuzmenko: woot? hahah
Why `spam` and `eggs`? =)))

DmitryKuzmenko: I propose a critical fix here. Please replace with
`menu_item = ['водка', 'матрёшка', 'медведь', 'балалайка']`

terminalmage: @DmitryKuzmenko Monty Python joke: https://www.youtube.com/watch?v=M_eYSuPKP3Y

Also the Python docs use these in their examples: https://docs.python.org/3.2/tutorial/introduction.html#lists

PR #45666, opened 2018-01-24
Fix failing pkg integration tests for releases with no '.'

terminalmage:
Fedora releases are integers, and Arch's osrelease is simply `rolling`, so https://github.com/saltstack/salt/pull/45636 caused these tests to regress. This fixes them by moving the osrelease check until after the MacOS check.

It also fixes the windows check, which would _always_ evaluate to `True` since it was not calling the function. Therefore, the `if` would just be a simple boolean which would resolve to `True` since it was just checking a function reference.

Resolves https://github.com/saltstack/salt-jenkins/issues/793


terminalmage:
Re-pointed at `2017.7.3` branch per @rallytime

rallytime (approved):

PR #45667, opened 2018-01-24
default to upgrading when refreshing on archlinux

gtmanfred:
### What does this PR do?
Archlinux does not do resolution for soname bumps, so if one set of packages is
updated, but something depended on a dependency package's soname, and it is
bumped, the other package will break.

When refreshing the package database, there should always be a -Su, run to
upgrade all other packages on the system.  But still allow sysupgrade to be set
to false to override this behavior.

http://gist.io/5660494

### Tests written?

No

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/modules/pacman.py
+++ b/salt/modules/pacman.py
@@ -438,7 +438,7 @@ def refresh_db(root=None):
 
 def install(name=None,
             refresh=False,
-            sysupgrade=False,
+            sysupgrade=None,
             pkgs=None,
             sources=None,
             **kwargs):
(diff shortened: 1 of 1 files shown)


gtmanfred:
I would really like to get this backported to 2017.7.3, so that we can use that release to bootstrap arch instances of machines that have updates available, for the test suite

@rallytime ^^

Thanks,
Daniel

rallytime:
@gtmanfred OK, we can get that in.

terminalmage (changes requested):
Please see my comments on the backport version of this PR: https://github.com/saltstack/salt/pull/45672#pullrequestreview-91345405

terminalmage (approved):

PR #45672, opened 2018-01-24
Back-port #45667 to 2017.7.3

rallytime:
Back-port #45667 to 2017.7.3

### DIFF
--- a/salt/modules/pacman.py
+++ b/salt/modules/pacman.py
@@ -438,7 +438,7 @@ def refresh_db(root=None):
 
 def install(name=None,
             refresh=False,
-            sysupgrade=False,
+            sysupgrade=None,
             pkgs=None,
             sources=None,
             **kwargs):
@@ -478,7 +478,8 @@ def install(name=None,
 
     sysupgrade
         Whether or not to upgrade the system packages before installing.
-
+        If refresh is set to ``True`` but sysupgrade is not specified, ``-u`` will be
+        applied
 
     Multiple Package Installation Options:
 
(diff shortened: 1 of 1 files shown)


salt/modules/pacman.py
```diff
@@ -686,7 +684,7 @@ def upgrade(refresh=False, root=None, **kwargs):
             and __salt__['config.get']('systemd.scope', True):
         cmd.extend(['systemd-run', '--scope'])
     cmd.extend(['pacman', '-Su', '--noprogressbar', '--noconfirm'])
-    if salt.utils.is_true(refresh):
+    if refresh:
```
terminalmage: If we're not using `salt.utils.is_true()` any more, then we should be using `if refresh is True` here.

salt/modules/pacman.py
```diff
@@ -478,7 +478,8 @@ def install(name=None,
 
     sysupgrade
         Whether or not to upgrade the system packages before installing.
-
+        If refresh is set to ``True`` but sysupgrade is not set, ``-u`` will be
```
terminalmage: I would prefer to use `specified` instead of `set` here.

terminalmage (changes requested):

terminalmage:
I didn't realize that this was a backport PR for https://github.com/saltstack/salt/pull/45667, I reviewed it first. We should probably have @gtmanfred make the changes in https://github.com/saltstack/salt/pull/45667 and then cherrypick them onto this PR once he's done that.

rallytime:
@terminalmage Yeah, sounds good. @gtmanfred LMK when that's updated on your end and I'll pull in your commit.

terminalmage (approved):

Issue #45472, opened 2018-01-16
Add auto tests for new auto-discovery feature


Ch3LL:
Need to add auto tests for the new auto-discovery feature added here: https://github.com/saltstack/salt/pull/44678

To note this has been turned off by default in this pr: https://github.com/saltstack/salt/pull/45412

ping @isbm any chance you can write some tests up for this feature you wrote?

PR #45685, opened 2018-01-25
Add unit tests for the SSDP (network discovery)

isbm:
### What does this PR do?

- Adds a set of Unit Tests for SSDP discovery mode
- Adds few minor bugfixes in SSDP

https://github.com/saltstack/salt/issues/45472

### Tests written?

Yes

@Ch3LL here, you have it!

### DIFF
--- a/tests/unit/utils/test_ssdp.py
+++ b/tests/unit/utils/test_ssdp.py
@@ -398,3 +398,21 @@ class SSDPClientTestCase(TestCase):
         assert 'foo' in clnt._config
         assert 'foo' not in config
 
+    def test_query(self):
+        '''
+        Test if client queries the broadcast
+        :return:
+        '''
+        config = {ssdp.SSDPBase.SIGNATURE: 'SUSE Enterprise Server',
+                  ssdp.SSDPBase.PORT: 4000}
+        f_time = 1111
+        _socket = MagicMock()
+        with patch('salt.utils.ssdp.socket', _socket),\
+             patch('salt.utils.ssdp.time.time', MagicMock(return_value=f_time)):
+            clnt = ssdp.SSDPDiscoveryClient(**config)
+            clnt._query()
+            assert clnt._socket.sendto.called
+            message, target = clnt._socket.sendto.call_args[0]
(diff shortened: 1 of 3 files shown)


tests/unit/utils/test_ssdp.py
```diff
@@ -0,0 +1,531 @@
...
+    NO_MOCK_REASON,
+    MagicMock,
+    patch)
+
+from salt.utils import ssdp
+import datetime
+
+try:
+    import pytest
+except ImportError as err:
```
terminalmage: This is a nitpick, but if we're not using `err`, we don't really need to save the `ImportError` instance. It can just be `except ImportError:`.

terminalmage (changes requested):
Just one minor nitpick.

terminalmage (approved):

tests/unit/utils/test_ssdp.py
```diff
@@ -14,6 +14,7 @@
 from salt.utils import ssdp
 import datetime
 from salt.ext.six.moves import zip
+from salt.ext.six import text_type as str
```
terminalmage: Since we will eventually be linting for `str()`, we'd prefer using six.text_type in the code instead of replacing the builtin like that.

terminalmage (changes requested):

tests/unit/utils/test_ssdp.py
isbm: @terminalmage OK, fixd.

terminalmage: We should be importing `six` and using `six.text_type` instead of importing `text_type` directly into the module's global namespace. It's not as important here in the tests, but it's very important in the salt code itself. The reason for this is because our loader will add any functions in the global namespace to the LazyDict. So, if this were in salt/utitls/ssdp.py we'd have a function called `__utils__['ssdp.text_type']()`.

This wasn't important enough to delay merging, though. Since the tests passed and everything else was fine, I took care of changing this import in 7115cd9d as part of a manual merge.

PR #45686, opened 2018-01-25
Use dbus-run-session instead of dbus-launch

bdrung:
As described in https://lists.debian.org/debian-devel/2016/08/msg00554.html Simon McVittie tries to reduce how much dbus-launch is used in Debian.

This change requires dbus 1.8 or newer, but does not require dbus-launch (shipped by dbus-x11) or X11.

Closes: https://bugs.debian.org/836297

### DIFF
--- a/salt/modules/gnomedesktop.py
+++ b/salt/modules/gnomedesktop.py
@@ -20,6 +20,7 @@ try:
 except ImportError:
     HAS_GLIB = False
 
+import salt.utils
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


salt/modules/gnomedesktop.py
```diff
@@ -62,7 +62,7 @@ def _get(self):
             log.info('User does not exist')
             return False
 
-        cmd = 'dbus-launch --exit-with-session gsettings get {0} {1}'.format(self.SCHEMA, self.KEY)
+        cmd = 'dbus-run-session -- gsettings get {0} {1}'.format(self.SCHEMA, self.KEY)
```
gtmanfred: can you add a check to make sure that this command is actually there?

Since it looks like it is easily interchangable, it would be nice to just make sure that older setups that only have dbus-launch can still work.

Thanks,
Daniel

gtmanfred (changes requested):

salt/modules/gnomedesktop.py
bdrung: Done.

bdrung:
Thanks. Added missing import.

gtmanfred (approved):

Issue #40173, opened 2017-03-20
Document the Open File limit issue better


gtmanfred:
### Description of Issue/Question
Document how to handle having the number of open files better for multiple init systems.

#39957

gtmanfred:
@markuskramerIgitt What are your thoughts about this document here?

https://docs.saltstack.com/en/latest/ref/configuration/master.html#max-open-files

At the bottom of it, it lists the recommended thing to google per distribution?  If we recommended checking out the documentation on this check_max_open_files error log, would that have lead you to find the answer faster?

Thanks,
Daniel

markuskramerIgitt:
Unfortunately no. For Debian, documentation of `max_open_files` must not refer to `ulimit –Hn`, or “the hard limit for max open files”, but to  the "max open files of the salt-master process". 
On Debian, "max open files allowed for the salt-master process" is much lower than “the hard limit for max open files”. 

On Debian, Salt sets the "max open files for the salt-master process" as `LimitNOFILE` in file salt-master.service, which is part of the Salt Git repository.  As this file is in the Git-Repository, the Salt documentation should reference it directly, not by offering a Google search.


Does Salt limits salt-master open files on other OS too?
Which OS do not limit the number of open files of a process?

In Python the "max open files per process” is named RLIMIT_NOFILE. 

The description should mention that operating systems can limit the number of open files of the salt-master and Salt-Configuration does so on Debian with `LimitNOFILE`. 


Additionally
On Debian, LimitNOFILE has the Salt-chosen default of 16384. Therefore a Debian salt-master on startup instantly logs an INFO that it must lower `max_open_files` from 100,000 down to 16,384. 
Do you think this is a misconfiguration and could be changed?

gtmanfred:
That is what it says in the documentation.

```
Remember that this value CANNOT be higher than the hard limit. Raising the hard limit depends on the OS and/or distribution, a good way to find the limit is to search the internet for something like this:

raise max open files hard limit debian
```

markuskramerIgitt:
Setting `max_open_files` (documented in https://docs.saltstack.com/en/latest/ref/configuration/master.html#max-open-files) unfortunately does not solve the issue of PR https://github.com/saltstack/salt/pull/39957

[...]

PR #45688, opened 2018-01-25
Raise LimitNOFILE to default max open files

bdrung:
The configuration option max_open_files defaults to 100,000, but the salt-master service is only started with a limit of 16,384 open files. Therefore the higher limit cannot be applied and there will be a log
message:

```
Current values for max open files soft/hard setting: 16384/16384
The value for the 'max_open_files' setting, 100000, is higher than what the user running salt is allowed to raise to, 16384. Defaulting to 16384.
```

This is related to #40173.


terminalmage (approved):

terminalmage:
CC: @dmurphy18

PR #45702, opened 2018-01-25
Endode results to utf-8

twangboy:
### What does this PR do?
Since the test is assuming 'utf-8', we need to make sure the `data.decode` is using 'utf-8'.

https://github.com/saltstack/salt-jenkins/issues/439

### Tests written?
Yes

### Commits signed with GPG?
Yes


tests/unit/utils/test_data.py
```diff
@@ -225,7 +225,8 @@ def test_decode(self):
         ]
 
         ret = salt.utils.data.decode(
-            self.test_data, preserve_dict_class=True, preserve_tuples=True)
+            self.test_data, encoding='utf-8', preserve_dict_class=True,
```
terminalmage: Can you help me understand why this would be necessary? As of https://github.com/saltstack/salt/pull/45491 strings are decoded using `salt.utils.stringutils.to_unicode()`, which falls back to `utf-8` if it encounters a decode error using `__salt_system_encoding__`. The only reason I can think of which would cause this test to fail on Windows is if it successfully decoded using `cp1252` and did not fall back to `utf-8`. But as far as I can see, `cp1252` doesn't decode these strings properly:

```python
>>> x = 'спам'
>>> x
'\xd1\x81\xd0\xbf\xd0\xb0\xd0\xbc'
>>> x.decode('cp1252')
  File "/usr/lib/python2.7/encodings/cp1252.py", line 15, in decode
    return codecs.charmap_decode(input,errors,decoding_table)
```

Is the test suite using something other than `cp1252` as its `__salt_system_encoding__`? If it isn't, then this test should pass on Windows as of when https://github.com/saltstack/salt/pull/45491 was merged. Could you try to run this test from the head of the `oxygen` branch and see what results you get?

twangboy: So, the problem was that it was NOT encountering an error and decoding the data to `cp1252`. But the test was written to check for `utf-8`. So, we're just forcing it to use `utf-8` for the purpose of the test.

terminalmage: OK, thanks for clearing that up.

terminalmage (approved):

Issue #45409, opened 2018-01-12
win_path.exists should not change given entry to lowercase


dhs-rec:
### Description of Issue/Question
Adding a directory to the Windows %PATH% using win_path.exists changes the given PATH entry to lowercase. Please stop doing that.

### Setup
```
add_to_path:
  win_path.exists:
    - name: 'C:\Some\Directory'
```
is added to %PATH% as *c:\some\directory*.

### Steps to Reproduce Issue
Apply above state.

### Versions Report
2017.7.2 on both master and (Windows) minion.

Ch3LL:
ping @saltstack/team-windows can i get some input here. Seems like an easy fix. Not sure if there is a reason we are moving it to lowercase though.

twangboy:
It's probably doing a comparison to see if the path already exists in %PATH%. Though Windows itself is not case sensitive, python is, so it probably uses `.lower()` to make the comparison. I can take a look at this. I don't like that it changes the capitalization either. It should be able to do that comparison and make the changes without modifying the existing entries.

dhs-rec:
@lorengordon, no it's not cosmetic. If Salt changes things at will, it can't be trusted. Also think about automated tests to verify that the provisioned box is in the desired state.
@twangboy, thanks.

twangboy:
@dhs-rec I don't understand your `it can't be trusted` comment. Since path is a single text field, the only way to add a path is to change it. Even if you aren't changing capitalization, you're still writing the entire path.

dhs-rec:
What I mean is: If it changes a simple thing like path capitalization at will, are there other places where it does unwanted changes?

terminalmage:
> Also think about automated tests to verify that the provisioned box is in the desired state.

There were tests, they were just written to confirm the incorrect path normalization behavior. This is fixed as part of https://github.com/saltstack/salt/pull/45703.

PR #45703, opened 2018-01-26
Improve/remove PATH environment modifications

terminalmage:
- `salt/modules/rbenv.py`: Forces use of str types in the custom env dict passed to `cmd.run_all`.
- `salt/modules/syslog_ng.py`: Ditches janky PATH munging in favor of the stable and long-existing support built into cmdmod.py
- `salt/modules/win_path.py`: Forces use of str types in path modification functions.
- `salt/states/win_path.py`: Completely rewritten. Duplicated code from the execution module removed in favor of calls to the execution module. Tests junked and 14 new tests written.
- `salt/utils/path.py`: The `which()` function was mistakenly modified in 20033ee to inject the directories in the POSIX default path into the PATH environment variable (even for Windows!). We never used the PATH to find the executables, we simply cycled through the dirs one by one and looked for an executable file matching the named path. The code that modifies the path is now removed. In addition, `which()` now uses `salt.utils.path.join()`, which gracefully handles mismatched str and unicode directory components to prevent decode errors.  `join()` has also been simplified to use `salt.utils.data.decode()` to normalize directory components to unicode.

### DIFF
--- a/salt/modules/rbenv.py
+++ b/salt/modules/rbenv.py
@@ -19,6 +19,7 @@ import logging
 # Import Salt libs
 import salt.utils.args
 import salt.utils.data
+import salt.utils.path
 import salt.utils.platform
 from salt.exceptions import SaltInvocationError
 
@@ -380,11 +381,14 @@ def do(cmdline, runas=None, env=None):
     if not env:
         env = {}
 
+    # NOTE: Env vars (and their values) need to be str type on both Python 2
+    # and 3. The code below first normalizes all path components to unicode to
+    # stitch them together, and then converts the result back to a str type.
     env[str('PATH')] = salt.utils.stringutils.to_str(   # future lint: disable=blacklisted-function
-        '{0}/shims:{1}'.format(
-            path,
+        os.pathsep.join((
+            salt.utils.path.join(path, 'shims'),
             salt.utils.stringutils.to_unicode(os.environ['PATH'])
-        )
+        ))
     )
 
     try:
--- a/salt/modules/rbenv.py
+++ b/salt/modules/rbenv.py
@@ -380,7 +380,12 @@ def do(cmdline, runas=None, env=None):
     if not env:
         env = {}
 
-    env['PATH'] = '{0}/shims:{1}'.format(path, os.environ['PATH'])
+    env[str('PATH')] = salt.utils.stringutils.to_str(   # future lint: disable=blacklisted-function
+        '{0}/shims:{1}'.format(
+            path,
+            salt.utils.stringutils.to_unicode(os.environ['PATH'])
+        )
+    )
 
     try:
         cmdline = salt.utils.args.shlex_split(cmdline)
--- a/salt/modules/syslog_ng.py
+++ b/salt/modules/syslog_ng.py
@@ -796,6 +796,9 @@ def _run_command_in_extended_path(syslog_ng_sbin_dir, command, params):
             log.error('syslog_ng_sbin_dir path %s is not a directory',
                       syslog_ng_sbin_dir)
         else:
+            # Custom environment variables should be str types. This code
+            # normalizes the paths to unicode to join them together, and then
+            # converts back to a str type.
             env = {
                 str('PATH'): salt.utils.stringutils.to_str(  # future lint: disable=blacklisted-function
                     os.pathsep.join(
--- a/salt/modules/syslog_ng.py
+++ b/salt/modules/syslog_ng.py
@@ -735,14 +735,14 @@ def get_config_file():
     return __SYSLOG_NG_CONFIG_FILE
 
 
-def _run_command(cmd, options=()):
+def _run_command(cmd, options=(), env=None):
     '''
     Runs the command cmd with options as its CLI parameters and returns the
     result as a dictionary.
     '''
     params = [cmd]
     params.extend(options)
-    return __salt__['cmd.run_all'](params, python_shell=False)
+    return __salt__['cmd.run_all'](params, env=env, python_shell=False)
 
 
 def _determine_config_version(syslog_ng_sbin_dir):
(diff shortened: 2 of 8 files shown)


salt/modules/rbenv.py
```diff
@@ -380,7 +380,12 @@ def do(cmdline, runas=None, env=None):
     if not env:
         env = {}
 
-    env['PATH'] = '{0}/shims:{1}'.format(path, os.environ['PATH'])
+    env[str('PATH')] = salt.utils.stringutils.to_str(   # future lint: disable=blacklisted-function
+        '{0}/shims:{1}'.format(
+            path,
+            salt.utils.stringutils.to_unicode(os.environ['PATH'])
+        )
+    )
```
terminalmage: NOTE: I know it may seem weird that this code converts the PATH env var to unicode and then back to a `str` type. The unicode conversion is to ensure that the `.format()` doesn't raise a decode error, and the `to_str()` ensures the result is a `str` type like we need it to be.

salt/modules/win_path.py
```diff
@@ -126,10 +130,16 @@ def add(path, index=0):
     if index > len(sysPath):
         index = len(sysPath)
 
-    localPath = os.environ["PATH"].split(os.pathsep)
+    pathsep = str(os.pathsep)  # future lint: disable=blacklisted-function
```
damon-atkins: Windows registry is Unicode and the PATH should be Unicode as well. As NTFS is Unicode filenames.
FAT based files systems do not support Unicode, but that would be unto the person to make sure the input is valid ASCII.
So the reg function read the current path will return Unicode, and you need to send Unicode back when setting it.

damon-atkins (changes requested):

salt/modules/rbenv.py
rallytime: Maybe you should add a comment above explaining this, just so someone doesn't see it in the code and remove it in the future.

salt/modules/syslog_ng.py
```diff
@@ -784,49 +784,27 @@ def set_parameters(version=None,
...
-    return ret
+    orig_path = os.environ.get('PATH', '')
+    env = None
+    if syslog_ng_sbin_dir:
+        if not os.path.isdir(syslog_ng_sbin_dir):
+            log.error('syslog_ng_sbin_dir path %s is not a directory',
+                      syslog_ng_sbin_dir)
+        else:
+            env = {
+                str('PATH'): salt.utils.stringutils.to_str(  # future lint: disable=blacklisted-function
```
rallytime: A comment about this `str()` and `to_str()` behavior would likely be good here as well.

rallytime (changes requested):
Overall, this is really well done. I have a couple of small comments.
This needs to be rebased and it also looks like this is causing the `unit.modules.test_syslog_ng` tests to fail.

I'd like to see what @twangboy finds during his testing as well.

salt/modules/win_path.py
terminalmage: It seems that `reg.set_value` would convert to unicode anyway, so I'm not sure that's too big of a concern, but I've changed the calls to it so that a unicode PATH string is passed to it.

You are wrong however, with respect to `os.environ` containing unicode data. On Python 2 this causes a traceback:

```
  (snip)
  ...
  File "c:\python27\lib\subprocess.py", line 390, in __init__
    errread, errwrite)
  File "c:\python27\lib\subprocess.py", line 640, in _execute_child
    startupinfo)
TypeError: environment can only contain strings
```

terminalmage:
@rallytime I still need to address your concerns, will do so shortly.

terminalmage:
@rallytime ready for re-review

rallytime (approved):

salt/modules/win_path.py
damon-atkins: Thanks for trying.  It just provides more proof that PY2 is no longer suitable for Windows, PY2 has too many hacks to work in modern versions of Windows which use Unicode.  Even PowerShell 6 has now moved to UTF-8 without BOM instead of UTF with BOM to be more cross platform friendly.

damon-atkins (approved):

terminalmage:
I've just added support for negative indexes to this PR, and added new tests.

twangboy (approved):

terminalmage:
@rallytime @twangboy please re-review, I've updated the state and execution module to properly handle case-insensitive paths without converting them to lowercase (see https://github.com/saltstack/salt/issues/45409).

rallytime (approved):
Nicely done, and thanks for adding tests!

salt/modules/win_path.py
```diff
@@ -68,9 +78,12 @@ def get_path():
 
         salt '*' win_path.get_path
     '''
-    ret = __salt__['reg.read_value']('HKEY_LOCAL_MACHINE',
-                                   'SYSTEM\\CurrentControlSet\\Control\\Session Manager\\Environment',
-                                   'PATH')['vdata'].split(';')
+    ret = salt.utils.stringutils.to_unicode(
```
damon-atkins: __salt__['reg.*'] always returns Unicode since 2016.11, if it does not its a bug.

rallytime:
@terminalmage I think the syslog_ng test failures are related here. See `unit.modules.test_syslog_ng` on Python 3 tests.

twangboy (approved):

rallytime:
@terminalmage There are some lint errors in the test file. Can you take a look?

rallytime:
@terminalmage A couple of related failures still: 
- unit.modules.test_syslog_ng.SyslogNGTestCase.test_end_to_end_statement_generation
- unit.modules.test_syslog_ng.SyslogNGTestCase.test_option_with_parameters
- unit.modules.test_syslog_ng.SyslogNGTestCase.test_value_with_arguments

https://jenkins.saltstack.com/job/PR/job/salt-pr-linode-ubuntu14-n/19528/#showFailuresLink

rallytime:
@terminalmage There's a lint error on this PR too: https://jenkins.saltstack.com/job/PR/job/salt-pr-lint-n/18938/violations/

I think some of the failures here are related as well: https://jenkins.saltstack.com/job/PR/job/salt-pr-rs-cent7-n/15983/#showFailuresLink

Can you take a look?

terminalmage:
Lint error fixed, and the Cent7 failures were due to an unrelated issue that was fixed by @garethgreenaway in a separate PR. I've rebased, they should pass now.

Issue #45682, opened 2018-01-25
state.low: KeyError: '__id__'


jasperla:
### Description of Issue/Question

`state.low` expects an explicit `__id__`.

### Setup

CentOS 7; salt installed with bootstrap script.

### Steps to Reproduce Issue

The [documentation](https://docs.saltstack.com/en/develop/ref/modules/all/salt.modules.state.html#salt.modules.state.low) of `state.low` suggests the following should work:

```
[root@salt salt]# salt salt\* state.low '{"state": "file", "fun": "exists", "name": "/etc/fstab"}'
salt.localdomain:
    The minion function caused an exception: Traceback (most recent call last):
      File "/usr/lib/python2.7/site-packages/salt/minion.py", line 1470, in _thread_return
        return_data = executor.execute()
      File "/usr/lib/python2.7/site-packages/salt/executors/direct_call.py", line 28, in execute
        return self.func(*self.args, **self.kwargs)
      File "/usr/lib/python2.7/site-packages/salt/modules/state.py", line 312, in low
        ret = st_.call(data)
      File "/usr/lib/python2.7/site-packages/salt/state.py", line 1901, in call
        ret['__id__'] = low['__id__']
    KeyError: '__id__'
[root@salt salt]#
```

However it does not because no explicit `__id__` is added. The state compiler explicitly references this key since https://github.com/saltstack/salt/pull/26780.

So either the documentation needs to be updated to reflect the need to explicitly pass `__id__`, or `State.call()` needs to handle it being absent.

Also, tests for `state.low` have not caught this for some reason.

### Versions Report

```
Salt Version:
           Salt: 2017.7.2

Dependency Versions:
           cffi: Not Installed
       cherrypy: Not Installed
       dateutil: Not Installed
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.7.2
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.4.8
   mysql-python: Not Installed
      pycparser: Not Installed
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: Not Installed
         Python: 2.7.5 (default, Aug  4 2017, 00:39:18)
   python-gnupg: Not Installed
         PyYAML: 3.11
          PyZMQ: 15.3.0
           RAET: Not Installed
[...]

PR #45705, opened 2018-01-26
check the lowstate __id__ before using it

jasperla:
it may not always be set, for example when coming in via state.low.

### What does this PR do?

Effectively unbreak `state.low` when `__id__` is not explicitly passed (as suggested by the documentation).

#45682

### Previous Behavior

```
[root@salt salt]# salt salt\* state.low '{"state": "file", "fun": "exists", "name": "/etc/fstab"}'
salt.localdomain:
      File "/usr/lib/python2.7/site-packages/salt/minion.py", line 1470, in _thread_return
      File "/usr/lib/python2.7/site-packages/salt/executors/direct_call.py", line 28, in execute
      File "/usr/lib/python2.7/site-packages/salt/modules/state.py", line 312, in low
        ret = st_.call(data)
      File "/usr/lib/python2.7/site-packages/salt/state.py", line 1901, in call
        ret['__id__'] = low['__id__']
    KeyError: '__id__'
[root@salt salt]#
```

### New Behavior

```
[root@salt salt]# salt salt\* state.low '{"state": "file", "fun": "exists", "name": "/etc/fstab"}'
salt.localdomain:
    ----------
    __run_num__:
        0
    __sls__:
        None
    changes:
        ----------
    comment:
        Path /etc/fstab exists
    duration:
        0.792
    name:
        /etc/fstab
    pchanges:
        ----------
    result:
        True
    start_time:
        07:48:24.983401
[root@salt salt]#
```

### Tests written?

No(t yet)

### Commits signed with GPG?

No

### DIFF
--- a/salt/state.py
+++ b/salt/state.py
@@ -1980,7 +1980,8 @@ class State(object):
         # duration in milliseconds.microseconds
         duration = (delta.seconds * 1000000 + delta.microseconds)/1000.0
         ret['duration'] = duration
-        ret['__id__'] = low['__id__']
+        if '__id__' in low:
+            ret['__id__'] = low['__id__']
         log.info(
             'Completed state [%s] at time %s (duration_in_ms=%s)',
             low['name'].strip() if isinstance(low['name'], six.string_types)


jasperla:
@gtmanfred I'm not quite sure how to create the test though, as currently the `State.call()` is [mocked](https://github.com/saltstack/salt/blob/develop/tests/unit/modules/test_state.py#L65) in the `state.low` tests. Should it be adjusted too like such:
```
        def call(data):
            '''
                Mock call method
            '''
            ret = {}
            if '__id__' in data:
              ret['__id__'] = data['__id__']
            return ret
```

and adjust the callers accordingly?

gtmanfred:
so, you can't do a mock like that, because they you aren't actually testing the `call()` function, and if a change is made, it won't be caught.

If you mocked the possible external function calls in the `call()` function, that would be what I would do, and just have one extra state.low test that does this.

garethgreenaway (approved):

jasperla:
@gtmanfred I'm sorry but while I understand what you mean, I have no idea how to go about this.
You mean to mock any (missing) external calls from `salt/salt/state.py#State.call()` in `test_state.py` and then one extra `state.low` test that that does what exactly?

gtmanfred:
since you are testing the code for state.low against salt.state:State.call(), you need to make sure you aren't mocking call(), otherwise what is the point, if someone changes State.call(), the test is not going to fail if they break state.low, because you have mocked the whole function.

rallytime:
@jasperla Did you see the comment above?

jasperla:
@rallytime I have, then I got distracted by `smartos_nictagadm`. I'll follow-up shortly!

rallytime:
Just another bump on this one @jasperla :)

thatch45:
Sorry I am so slow on this one. If a low chunk does not have an `__id__` then the `__id__` should be the name. So the right way to fix this would be to still return an `__id__` but copy it in from the `name` key

thatch45 (changes requested):
We should never return a `ret` from the state system without an `__id__`

thatch45:
You may have also noticed that in your example the outputter is the nested outputter and not the highstate outputter. This is because the highstate outputter is crashing when it looks for `__id__` and is falling back onto the nested outputter

rallytime:
Hi @jasperla - Were you able to come back to this regarding the comments above?

rallytime:
I am going to close this PR for now since we haven't hear back for a while on this. When you're ready, please let us know and we will happily open this up again. Or feel free to open a new PR. Whichever is easier for you. Thank you!

Issue #45301, opened 2018-01-05
Ctl+C causes stack trace on Windows


twangboy:
### Description of Issue/Question
When running salt-minion from the command line in Windows using `salt-minion -l debug`, hitting `ctl+c` to end the process causes the following stack trace:
```
Traceback (most recent call last):
  File "c:\dev\salt\salt\log\handlers\__init__.py", line 211, in emit
    self.enqueue(self.prepare(record))
  File "c:\dev\salt\salt\log\handlers\__init__.py", line 177, in enqueue
  File "c:\python27\lib\multiprocessing\queues.py", line 264, in _feed
    self.queue.put_nowait(record)
  File "c:\python27\lib\multiprocessing\queues.py", line 155, in put_nowait
    return self.put(obj, False)
  File "c:\python27\lib\multiprocessing\queues.py", line 100, in put
    assert not self._closed
AssertionError
Logged from file setup.py, line 973
    send(obj)
IOError: [Errno 232] The pipe is being closed
```

### Setup
Pull down the head of the Oxygen branch
Install a dev environment by running the `build_env_2.ps1` script in `pkg\windows` (`powershell -file build_env_2.ps1`)
Install salt by running the following command from the root salt directory: `pip install -e .`
Start a salt-minion in debug mode: `salt-minion -l debug`

### Steps to Reproduce Issue
Once the minion is started, hit `ctl+c`
Observe the stack trace

### Versions Report
Head of Oxygen

twangboy:
@skizunov Do you have any ideas on this?

PR #45707, opened 2018-01-26
Fix exception when shutting down logging listener

skizunov:
### What does this PR do?

`shutdown_multiprocessing_logging_listener()` shuts down
the logging process which closes the multiprocessing queue
used to shuttle logging data. Before shutting down the
logging process, prevent this process from trying to write
to the logging queue to prevent an exception from occurring when
trying to write to a multiprocessing queue that has been closed
by the other side.

#45301

### Tests written?

No

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/log/setup.py
+++ b/salt/log/setup.py
@@ -946,6 +946,12 @@ def shutdown_multiprocessing_logging_listener(daemonizing=False):
         # We're in the MainProcess and we're not daemonizing, return!
         # No multiprocessing logging listener shutdown shall happen
         return
+
+    if not daemonizing:
+        # Need to remove the queue handler so that it doesn't try to send
+        # data over a queue that was shut down on the listener end.
+        shutdown_multiprocessing_logging()
+
     if __MP_LOGGING_QUEUE_PROCESS is None:
         return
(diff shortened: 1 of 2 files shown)


skizunov:
@twangboy : I hope this will fix the problem you ran into described in #45301. Let me know.

twangboy:
@skizunov That stops the stacktrace, but it shows a message at the end:
```
[DEBUG   ] salt.crypt.get_rsa_pub_key: Loading public key
No handlers could be found for logger "salt.log.setup"
The Salt Minion is shutdown. Minion received a SIGINT. Exited.
```

twangboy (approved):

rallytime (approved):

terminalmage (commented):
@skizunov did you see @twangboy's message about the additional debug log message?

skizunov:
@twangboy : I believe I addressed your issue. Unfortunately, once you stop the Multiprocessing Logging Listener, there is no way to log to file (logging to file needs to use the Multiprocessing Logging Listener to support log file rotation on Windows). But it will still log errors to console. Hopefully, if shutting down the Multiprocessing Logging Listener is one of the last things that are done, the log file won't miss much.

Let me know if it does resolve your issue.

PR #45711, opened 2018-01-26
Fix Unicode tests when run with LC_ALL=POSIX

bdrung:
When running the unit tests with the locale set to POSIX, some Unicode tests fail:

```
$ LC_ALL=POSIX python3 ./tests/runtests.py --unit
[...]
ERROR: test_list_products (unit.modules.test_zypper.ZypperTestCase)
[CPU:0.0%|MEM:73.2%]
  File "tests/unit/modules/test_zypper.py", line 236, in
test_list_products
    'stdout': get_test_data(filename)
  File "tests/unit/modules/test_zypper.py", line 53, in get_test_data
    return rfh.read()
  File "/usr/lib/python3.6/encodings/ascii.py", line 26, in decode
    return codecs.ascii_decode(input, self.errors)[0]
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 828: ordinal not in range(128)

ERROR: test_non_ascii (unit.templates.test_jinja.TestGetTemplate)
[CPU:0.0%|MEM:73.2%]
  File "tests/unit/templates/test_jinja.py", line 341, in test_non_ascii
    result = fp.read()
  File "/usr/lib/python3.6/encodings/ascii.py", line 26, in decode
    return codecs.ascii_decode(input, self.errors)[0]
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 5: ordinal not in range(128)

ERROR: test_non_ascii_encoding
(unit.templates.test_jinja.TestGetTemplate)
[CPU:0.0%|MEM:73.2%]
  File "tests/unit/templates/test_jinja.py", line 303, in
test_non_ascii_encoding
    fp_.read(),
  File "/usr/lib/python3.6/encodings/ascii.py", line 26, in decode
    return codecs.ascii_decode(input, self.errors)[0]
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 5: ordinal not in range(128)
```

Therefore open files in binary mode and explicitly decode them with utf-8 instead of their default locale.

### DIFF
--- a/tests/unit/modules/test_zypper.py
+++ b/tests/unit/modules/test_zypper.py
@@ -49,8 +49,8 @@ def get_test_data(filename):
     with salt.utils.fopen(
             os.path.join(
                 os.path.join(
-                    os.path.dirname(os.path.abspath(__file__)), 'zypp'), filename)) as rfh:
-        return rfh.read()
+                    os.path.dirname(os.path.abspath(__file__)), 'zypp'), filename), mode='rb') as rfh:
+        return salt.utils.to_str(rfh.read(), 'utf-8')
 
 
 @skipIf(NO_MOCK, NO_MOCK_REASON)
--- a/tests/unit/templates/test_jinja.py
+++ b/tests/unit/templates/test_jinja.py
@@ -294,13 +294,13 @@ class TestGetTemplate(TestCase):
                                'file_roots': self.local_opts['file_roots'],
                                'pillar_roots': self.local_opts['pillar_roots']},
                          a='Hi', b='Sàlt', saltenv='test', salt=self.local_salt))
-            self.assertEqual(out, salt.utils.to_unicode('Hey world !Hi Sàlt !' + os.linesep))
+            self.assertEqual(out, u'Hey world !Hi Sàlt !' + os.linesep)
             self.assertEqual(fc.requests[0]['path'], 'salt://macro')
 
             filename = os.path.join(TEMPLATES_DIR, 'files', 'test', 'non_ascii')
-            with salt.utils.fopen(filename) as fp_:
+            with salt.utils.fopen(filename, mode='rb') as fp_:
                 out = render_jinja_tmpl(
-                    fp_.read(),
+                    salt.utils.to_unicode(fp_.read(), 'utf-8'),
                     dict(opts={'cachedir': TEMPLATES_DIR, 'file_client': 'remote',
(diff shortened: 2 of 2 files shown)


terminalmage (commented):
Why would Salt ever be run with a POSIX locale? I don't see the need or use for this, to be honest.

bdrung:
The Debian build daemons set `LC_ALL=POSIX`. Therefore salt 2017.7.2+dfsg1-1 failed to build: https://buildd.debian.org/status/fetch.php?pkg=salt&arch=all&ver=2017.7.2%2Bdfsg1-1&stamp=1516915304&raw=0

If the test case require UTF-8, it should enforce it instead of relying on the locale to decode the file content to Unicode.

damon-atkins:
It's a good test given how much  `__salt_system_encoding__`  is used.  I would have thought most Linux/Unix platforms would have the `LANG` set to a UTF-8 LANG. e.g. `en_US.UTF-8` or `en_AU.UTF-8` etc

Which is the same as `__salt_system_encoding__`  being UTF-8

However en_US (or en_US.iso88591) would be more like POSIX, i.e. not UTF-8.

`locale -va | strings | egrep -A 13 'locale: (en_AU|en_US|an_ES)'`

bdrung:
The failing tests should check that the Jinja rendering works correctly with Unicode input. The result of the Jinja rendering is compared with expected ouput which is stored in UTF-8 encoded files (like `files/test/non_ascii`). These UTF-8 test files should be explicitly decoded as UTF-8 files instead of using the user's locale (which will fail if set to something else than UTF-8).

terminalmage (approved):
@rallytime Please work with me to ensure that we undo these changes after merging forward into oxygen/oxygen.rc1. In these branches, the to_unicoode/to_str helpers fall back to utf-8 on their own, so this change wouldn't be necessary there.

Also, @bdrung plans to have these tests run successfully when building future Salt releases, this change would need to be propagated to the dot-release branches for 2017.7. @rallytime heads-up on this as well.

rallytime:
Will do. Thanks for the heads-up @terminalmage.

Since this in going into the 2017.7 branch, all future releases starting with 2017.7.5 will have this fix. 

I'll make sure these changes don't go through into oxygen.

rallytime:
@terminalmage This PR is merged-forward to oxygen in #46041, and I was sure to adjust the files that this PR changed. Thanks!

PR #45714, opened 2018-01-26
Move to Unicode (3rd tier, tests)

isbm:
Unicode operation.

### DIFF
--- a/tests/unit/states/test_jboss7.py
+++ b/tests/unit/states/test_jboss7.py
@@ -12,6 +12,7 @@ from tests.support.mock import NO_MOCK, NO_MOCK_REASON, MagicMock, patch
 # Import Salt libs
 import salt.states.jboss7 as jboss7
 from salt.exceptions import CommandExecutionError
+from six import text_type
 
 
 @skipIf(NO_MOCK, NO_MOCK_REASON)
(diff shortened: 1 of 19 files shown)


rallytime:
@isbm There's a small lint error on this one: https://jenkins.saltstack.com/job/PR/job/salt-pr-lint-n/18774/violations/

tests/unit/states/test_jboss7.py
```diff
@@ -12,6 +12,7 @@
 # Import Salt libs
 import salt.states.jboss7 as jboss7
 from salt.exceptions import CommandExecutionError
+from six import text_type
```
terminalmage: Please change this to `from salt.ext import six` and use `six.text_type()` instead of `text_type()`. It wouldn't cause any issues here, but I'd like to keep the code uniform through the codebase, so we don't run the risk of another contributor seeing your example here and then writing a state/execution/etc. module using the same technique. Our loader will gobble up all public functions in the module's global namespace, so doing this in an execution module would add `modname.text_type` to the `__salt__` dunder, and we don't want that.

terminalmage (changes requested):
One small change and then the lint fix, and this should be good to merge. Thanks!

isbm:
@rallytime @terminalmage done!

PR #43807, opened 2017-09-28
cmdmod: Don't list-ify string commands on Windows

terminalmage:
See https://github.com/saltstack/salt/issues/43522.

PR #45716, opened 2018-01-26
fixed quoting of script path in cmd.script

ciiqr:
### What does this PR do?
This fixes the quoting of the script path in cmd.script on windows. Basically, shlex.quote doesn't quote arguments correctly for use in the cmd shell.

Was introduced with #45134

### Example errors
```
PS C:\Users\william> salt-call cmd.script salt://any-powershell-script.ps1 shell=powershell
[ERROR   ] cmd.script: Unable to clean tempfile ''C:\Users\william\AppData\Local\Temp\__salt.tmp.o_a7jnl2.ps1'': Path not found: 'C:\Users\william\AppData\Local\Temp\__salt.tmp.o_a7jnl2.ps1'
local:
    ----------
    pid:
        6236
    retcode:
        4294770688
    stderr:
        Processing -File ''C:\Users\william\AppData\Local\Temp\__salt.tmp.o_a7jnl2.ps1'' failed: The given path's format is not supported. Specify a valid path for the -File parameter.
    stdout:
        Windows PowerShell
        Copyright (C) Microsoft Corporation. All rights reserved.

PS C:\Users\william> salt-call cmd.script salt://any-batch-script.bat
[ERROR   ] cmd.script: Unable to clean tempfile ''C:\Users\william\AppData\Local\Temp\__salt.tmp.73tymk82.bat'': Path not found: 'C:\Users\william\AppData\Local\Temp\__salt.tmp.73tymk82.bat'
local:
    ----------
    pid:
        6432
    retcode:
        1
    stderr:
        The filename, directory name, or volume label syntax is incorrect.
    stdout:
```

### Tests written?

Yes

### Commits signed with GPG?

No

### DIFF
--- a/salt/modules/cmdmod.py
+++ b/salt/modules/cmdmod.py
@@ -36,7 +36,6 @@ from salt.exceptions import CommandExecutionError, TimedProcTimeoutError, \
     SaltInvocationError
 from salt.log import LOG_LEVELS
 from salt.ext.six.moves import range, zip
-from salt.ext.six.moves import shlex_quote as _cmd_quote
 from salt.utils.locales import sdecode
 
 # Only available on POSIX systems, nonfatal on windows
@@ -47,9 +46,10 @@ except ImportError:
 
 if salt.utils.is_windows():
     from salt.utils.win_runas import runas as win_runas
-    from salt.utils.win_functions import escape_argument as win_cmd_quote
+    from salt.utils.win_functions import escape_argument as _cmd_quote
     HAS_WIN_RUNAS = True
 else:
+    from salt.ext.six.moves import shlex_quote as _cmd_quote
     HAS_WIN_RUNAS = False
 
 __proxyenabled__ = ['*']
@@ -2148,10 +2148,7 @@ def script(source,
         os.chmod(path, 320)
         os.chown(path, __salt__['file.user_to_uid'](runas), -1)
 
-    if salt.utils.is_windows():
-        path = win_cmd_quote(path)
-    else:
-        path = _cmd_quote(path)
+    path = _cmd_quote(path)
 
     ret = _run(path + ' ' + str(args) if args else path,
                cwd=cwd,
--- a/salt/modules/cmdmod.py
+++ b/salt/modules/cmdmod.py
@@ -47,6 +47,7 @@ except ImportError:
 
 if salt.utils.is_windows():
     from salt.utils.win_runas import runas as win_runas
+    from salt.utils.win_functions import escape_argument as win_cmd_quote
     HAS_WIN_RUNAS = True
 else:
     HAS_WIN_RUNAS = False
@@ -2147,7 +2148,10 @@ def script(source,
         os.chmod(path, 320)
         os.chown(path, __salt__['file.user_to_uid'](runas), -1)
 
-    path = _cmd_quote(path)
+    if salt.utils.is_windows():
+        path = win_cmd_quote(path)
+    else:
+        path = _cmd_quote(path)
 
     ret = _run(path + ' ' + str(args) if args else path,
                cwd=cwd,
--- a/salt/utils/win_functions.py
+++ b/salt/utils/win_functions.py
@@ -5,6 +5,7 @@ missing functions in other modules
 '''
 from __future__ import absolute_import
 import platform
+import re
 
 # Import Salt Libs
 from salt.exceptions import CommandExecutionError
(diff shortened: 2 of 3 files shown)


damon-atkins:
This has already been fixed please try the develop version of 2017.7.  i.e. shlex should not be used at all on windows.

ciiqr:
@damon-atkins
- I've tried on the most recent version of `2017.7`, `2017.7.3`, `develop`, and `oxygen`. To no avail. Maybe there's some part of my process that's wrong or misses something?
	```
	git fetch saltstack
	git checkout {BRANCH}
	git reset --hard saltstack/{BRANCH} {BRANCH}

	cd .\pkg\windows\
	powershell -file .\build_env_3.ps1
	cd ../..
	pip --no-cache-dir install -e .
	# NOTE: I also have PYTHONDONTWRITEBYTECODE set to 1 to further prevent files being cached
	```
- If you could point me to what I'm doing wrong or where precisely you believe this is fixed, that would be great, but otherwise I'm left under the impression that this is still broken

- Also, your point about 'manipulating behind the scenes' doesn't really make sense to me. This is a temporary path that is generated by salt when the source file is copied to the minion, documenting the limitations hardly seems helpful when there wouldn't be much for the user to do to their state/call to make it work. I suppose you could override the temp directory path by setting $env:TEMP (and hope the special character isn't in the generated file name), this just seems unnecessary and adds extra work for the user, with little to no benefit on their part. The other thing to consider I guess, is that most of the time this path is going to be fine as it is, so passing it in raw probably wouldn't break anything for the average user. At the same time though, there's a reason the quoting was added in the first place, and there's no reason imho to let this just randomly pop up for users when there's a solution that's inline with how salt operates on other platforms... Nonetheless, if that's the solution that you feel is best, I'm open to whatever get this issue fixed.

damon-atkins:
There is no need build your own version.
https://raw.githubusercontent.com/saltstack/salt/2017.7/salt/modules/cmdmod.py  
You just need to backup `C:\salt\bin\Lib\site-packages\salt\modules\cmdmod.py` and replace it with a new version.

The fix was in #43807
Restart the salt-minion service.

ciiqr:
- 'The fix was in #43807'
	Cool, so I just checked that PR (#43807) out and it's definitelly a different issue. Where as that was an issue with shlex.split in the _run function, the offending shlex call here is shlex.quote (aliased to _cmd_quote) in the script function, before the _run function is even called. I've confirmed that the given commit for that PR is in my branch, you can see the applicable code in my branch on github is as expected: https://github.com/ciiqr/salt/blob/fix_cmd_script_quoting/salt/modules/cmdmod.py#L538

- 'There is no need build your own version.'
	Maybe I was unclear about this part, in order to test on those different versions of salt, and because I already had salt installed in editable mode, that is the process I take when switching version to ensure all the dependencies are correct and there aren't any cached files laying around.

- 'Are you running salt-call in a cmd.exe/powershell as "Run As Administrator"?'
	Yes

ciiqr:
oh, to address your two new points on that comment, the reason to not wrap `'"'` around the path is simply that that is insufficient to properly encode a parameter in the cmd shell, it would certainly be an improvement over not escaping at all, but it will leaves room for cmd to interpret the string the wrong way. As for changing the `is_windows` check to be on the import, that would definitely make things easier (and less likely to break) going forward, because yes, windows needs `"` (at the very least) and not `'`

damon-atkins:
@twangboy  might be worth targeting this at 2016.11 and merge it forward.

Also consider salt/states/archive.py which also contains 
`from salt.ext.six.moves import shlex_quote as _cmd_quote`

rallytime:
@ciiqr There are some lint failures here: https://jenkins.saltstack.com/job/PR/job/salt-pr-lint-n/18776/violations/

damon-atkins:
Only issue I have with this, is should be targeted at 2016.11

ciiqr:
Hmm, I branched off of 2017.7 because the call to quote the path in cmd.script was introduced there (as part of #45134). So I just want to confirm, if I'm going to recreate these changes off of 2016.11, that we're okay with that PR essentially being merged into 2016.11 also? Because we'll want the same test's that that PR added, and obviously my changes will introduce quoting for all platforms where is wasn't before.

twangboy:
With the upcoming Oxygen release, I don't think there is another planned dot release for 2016.11. I'm OK with this going in to 2017.7.

twangboy (approved):

garethgreenaway (approved):

absolutejam:
Does this mean these fixes are going in to `2017.7.5`?

rallytime:
@absolutejam Yeah

MerickOWA:
This appears to be fixed in 2017.7.5 but didn't get merged into 2018.3.0?

terminalmage:
@MerickOWA Correct. We had frozen for the 2018.3.0 release candidate already, so that fix will be in 2018.3.1.

yplonggit:
when the next version 2018.3.1 will be released?
Thanks for your patience.

PR #45724, opened 2018-01-26
Typo (`Hellium` → `Helium`)

eliasp:
### What does this PR do?
Typo (`Hellium` → `Helium`)

None

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/modules/file.py
+++ b/salt/modules/file.py
@@ -3500,7 +3500,7 @@ def path_exists_glob(path):
     Expansion allows usage of ? * and character ranges []. Tilde expansion
     is not supported. Returns True/False.
 
-    .. versionadded:: Hellium
+    .. versionadded:: 2014.7.0
 
     CLI Example:
 


salt/modules/file.py
```diff
@@ -3500,7 +3500,7 @@ def path_exists_glob(path):
     Expansion allows usage of ? * and character ranges []. Tilde expansion
     is not supported. Returns True/False.
 
-    .. versionadded:: Hellium
+    .. versionadded:: Helium
```
rallytime: Technically this should be `2014.7`. I probably missed changing this codename reference to the version number because of the misspelling. Mind fixing that?

rallytime (changes requested):

salt/modules/file.py
eliasp: Done

rallytime (approved):
Thanks!

PR #45725, opened 2018-01-26
[oxygen] Argument to struct.pack must be a bytescode not unicode with Python 2.7.5

garethgreenaway:
### What does this PR do?
Fixing code in salt/utils/vt.py

https://github.com/saltstack/salt-jenkins/issues/806

### Previous Behavior
Test failed on CentOS 7

### New Behavior
Test passes.

### Tests written?
No, but existing tests pass as expected.

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/utils/vt.py
+++ b/salt/utils/vt.py
@@ -708,7 +708,7 @@ class Terminal(object):
         def __detect_parent_terminal_size(self):
             try:
                 TIOCGWINSZ = getattr(termios, 'TIOCGWINSZ', 1074295912)
-                packed = struct.pack(str('HHHH'), 0, 0, 0, 0)  # future lint: blacklisted-function
+                packed = struct.pack(b'HHHH', 0, 0, 0, 0)
                 ioctl = fcntl.ioctl(sys.stdin.fileno(), TIOCGWINSZ, packed)
                 return struct.unpack('HHHH', ioctl)[0:2]
             except IOError:
@@ -731,7 +731,7 @@ class Terminal(object):
                 )
 
             TIOCGWINSZ = getattr(termios, 'TIOCGWINSZ', 1074295912)
-            packed = struct.pack('HHHH', 0, 0, 0, 0)
+            packed = struct.pack(b'HHHH', 0, 0, 0, 0)
             ioctl = fcntl.ioctl(self.child_fd, TIOCGWINSZ, packed)
             return struct.unpack('HHHH', ioctl)[0:2]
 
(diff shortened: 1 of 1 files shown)


salt/utils/vt.py
```diff
@@ -708,7 +708,7 @@ def _recv(self, maxsize):
         def __detect_parent_terminal_size(self):
             try:
                 TIOCGWINSZ = getattr(termios, 'TIOCGWINSZ', 1074295912)
-                packed = struct.pack('HHHH', 0, 0, 0, 0)
+                packed = struct.pack(str('HHHH'), 0, 0, 0, 0)  # future lint: blacklisted-function
```
terminalmage: I am missing something. The original code doesn't fail on Python 2 or Python 3:

**Python 2**
```python
>>> import struct
>>> struct.pack('HHHH', 0, 0, 0, 0)
'\x00\x00\x00\x00\x00\x00\x00\x00'
```

**Python 3**
```python
>>> import struct
>>> struct.pack('HHHH', 0, 0, 0, 0)
b'\x00\x00\x00\x00\x00\x00\x00\x00'
```

I recently fixed an issue with `struct` that affected Python 3, due to `struct.caclcsize()` not being able to accept a unicode string, but `struct.pack()` and `struct.unpack()` don't seem to be similarly impacted. Could you offer more background on how this produced the failure in the linked issue?

garethgreenaway: # Using CentOS 7

## Python 2
[root@6af876f92c00 /]# python2.7
Python 2.7.5 (default, Aug  4 2017, 00:39:18) 
[GCC 4.8.5 20150623 (Red Hat 4.8.5-16)] on linux2
>>> import struct
>>> struct.pack('HHHH', 0, 0, 0, 0)
TypeError: Struct() argument 1 must be string, not unicode
>>> struct.pack(str('HHHH'), 0, 0, 0, 0)
'\x00\x00\x00\x00\x00\x00\x00\x00'

## Python 3
>>> import struct
>>> struct.pack('HHHH', 0, 0, 0, 0)
b'\x00\x00\x00\x00\x00\x00\x00\x00'

garethgreenaway: @terminalmage FYI, https://bugs.python.org/issue19099

terminalmage: AHH! OK, this was fixed between 2.7.5 and 2.7.14, because that is what I used to check this.

So, it appears that struct.pack/unpack can use bytestrings for formatting, both on Python 2 and 3. So, can you change this (as well as the other 4 calls to this function) to use `b'HHHH'` instead of `str()`. That will also allow us to get rid of that pylint disable.

terminalmage (approved):

rallytime:
Go Go Jenkins!

PR #45731, opened 2018-01-27
Fix groupadd tests for Arch Linux

terminalmage:
The default /etc/login.defs does not have a value assigned to MOTD_FILE,
which causes a traceback in 2 of the tests.

Resolves https://github.com/saltstack/salt-jenkins/issues/804


rallytime (approved):

rallytime:
Go Go Jenkins!

PR #45733, opened 2018-01-28
PY3-compatibility fix for RH networking module

terminalmage:
Encoding on Python 3 produces a bytestring, when what we need is a unicode string.


rallytime (approved):
Nice - much simpler.

rallytime:
Go Go Jenkins!

PR #45735, opened 2018-01-28
Add debug logging to ShellCase functions

terminalmage:
Some of these have been failing in ways which are not reproducible
locally. This adds debug logging so that we can check the
salt-runtests.log for the return data, to aid in troubleshooting.

NOTE: This also contains a fix for a bug in the CLI output handler that
I think may have a chance of fixing some of these odd failures.


DmitryKuzmenko (approved):
👍

rallytime (approved):

PR #45737, opened 2018-01-28
SSDP PY3 compatibility fixes

terminalmage:


rallytime (approved):

isbm (approved):
Seems OK, as `transport.sendto` is looking for bytes. Thanks!

PR #45742, opened 2018-01-28
list.copy() is not compatible with python 2.7

marccardinal:
### What does this PR do?

I doubt that a .copy() is required there since list() would already create one but since the previous committer added it I am improving by replacing the .copy() with a [:] which makes a copy in python 2.7 and 3+

```
                File "/usr/lib/python2.7/dist-packages/salt/state.py", line 1934, in call
                  **cdata['kwargs'])
                File "/usr/lib/python2.7/dist-packages/salt/loader.py", line 1799, in wrapper
                  return f(*args, **kwargs)
                File "/usr/lib/python2.7/dist-packages/salt/states/cloud.py", line 277, in profile
                  info = __salt__['cloud.profile'](profile, name, vm_overrides=kwargs, opts=opts)
                File "/usr/lib/python2.7/dist-packages/salt/modules/cloud.py", line 199, in profile_
                  info = client.profile(profile, names, vm_overrides=vm_overrides, **kwargs)
                File "/usr/lib/python2.7/dist-packages/salt/cloud/__init__.py", line 345, in profile
                  mapper = salt.cloud.Map(self._opts_defaults(**kwargs))
                File "/usr/lib/python2.7/dist-packages/salt/cloud/__init__.py", line 237, in _opts_defaults
                  for provider in list(_providers)[:]:
              AttributeError: 'list' object has no attribute 'copy'
```

### Tests written?

No

### Commits signed with GPG?

No

### DIFF
--- a/salt/cloud/__init__.py
+++ b/salt/cloud/__init__.py
@@ -240,7 +240,7 @@ class CloudClient(object):
                          if a.get('provider', '')]
             if providers:
                 _providers = opts.get('providers', {})
-                for provider in list(_providers).copy():
+                for provider in list(_providers)[:]:
                     if provider not in providers:
                         _providers.pop(provider)
         return opts
--- a/salt/cloud/__init__.py
+++ b/salt/cloud/__init__.py
@@ -240,7 +240,7 @@ class CloudClient(object):
                          if a.get('provider', '')]
             if providers:
                 _providers = opts.get('providers', {})
-                for provider in list(_providers)[:]:
+                for provider in _providers.copy():
                     if provider not in providers:
                         _providers.pop(provider)
         return opts


marccardinal:
Is the build broken? This fix works well on my installation of 2018.2

gtmanfred (approved):
we had this conversation when it was added, list() should make a new list, and not need to be copied.

But since he ran into the issue initially, i want to make sure @techhat agrees with this

gtmanfred:
Also, just for anyone else comes by, the copy was there because python thought that the list was shrinking inside of the loop, which imho there is no way that should have been happening.

marccardinal:
Thanks @gtmanfred !

terminalmage (approved):

gtmanfred:
@techhat can we actually move the `.copy()` to the `_providers` and get rid of the list()?

ipmb:
Just my 2 cents, `list(_providers.keys())` makes it clearer that you're iterating over the _keys_ and does the copy so the dictionary can change without messing up the iteration.

techhat (approved):

salt/cloud/__init__.py
```diff
@@ -240,7 +240,7 @@ def _opts_defaults(self, **kwargs):
                          if a.get('provider', '')]
             if providers:
                 _providers = opts.get('providers', {})
-                for provider in list(_providers).copy():
+                for provider in list(_providers)[:]:
```
gtmanfred: can you get rid of the list() and just make it .copy() on `_providers`

gtmanfred (changes requested):
So, we specifically removed all of the `.keys()` because they are not actually needed and because it returns a `dict_keys()` object in python3.

Since the for loop should iterate over them anyway, we passed it through a list to make a new object so that it doesn't get modified.

But joseph found a spot where it was modifying the size of the dict in python3 here, even with list().

I would rather have this just be a `.copy()` which will make a new object of all the top level keys, and get rid of the list()

salt/cloud/__init__.py
mcardinalupgrade: Since _providers is a dict then yes .copy() would work on both python 2.7 and 3+ and also iterating on a dict would give the same behaviour as list(_providers) so thats good

mcardinalupgrade:
Updated the PR to reflect the latest iteration

gtmanfred (approved):

PR #45744, opened 2018-01-29
fix for UnicodeEncodeError on debconf.get_selections

marccardinal:
### What does this PR do?

Ignore non ascii characters in cmd output and error

Fix for `UnicodeEncodeError: 'ascii' codec can't encode characters in position 12839-12840: ordinal not in range(128)` when calling the `debconf.get_selections` module

### Tests written?

No

### Commits signed with GPG?

No

### DIFF
--- a/salt/modules/cmdmod.py
+++ b/salt/modules/cmdmod.py
@@ -609,7 +609,7 @@ def _run(cmd,
             out = ''
         except UnicodeDecodeError:
             log.error('UnicodeDecodeError while decoding output of cmd {0}'.format(cmd))
-            out = proc.stdout.decode(__salt_system_encoding__, 'replace')
+            out = proc.stdout.decode(__salt_system_encoding__, 'ignore')
 
         try:
             err = proc.stderr.decode(__salt_system_encoding__)
@@ -617,7 +617,7 @@ def _run(cmd,
             err = ''
         except UnicodeDecodeError:
             log.error('UnicodeDecodeError while decoding error of cmd {0}'.format(cmd))
-            err = proc.stderr.decode(__salt_system_encoding__, 'replace')
+            err = proc.stderr.decode(__salt_system_encoding__, 'ignore')
 
         if rstrip:
             if out is not None:


damon-atkins:
I believe this fix is incorrect.  Their is no reason to force all uses of cmdmod into the same result.  Salt Team are currently working their way through a Unicode conversion. Please test again after an 2018.2 rc release.

terminalmage (commented):
What is the output that is generating this? Also, what is your locale?

terminalmage:
The output from `salt minion_id test.versions_report` would also be useful information.

terminalmage:
I honestly don't see how this would even fix the issue, have you tested it? If the `debconf-get-selections` command is triggering this error, it's likely due to an incorrectly set-up locale.

terminalmage:
I took a look at this and came up with something different based on our unicode efforts: https://github.com/saltstack/salt/pull/45834

I don't think we'll be able to merge this PR due to the fact that the code you're modifying would be replaced by that PR. I have however started a discussion in that PR's comments to decide whether we should continue to use `replace`, or if we should switch to `ignore`. Since the `.decode()` is new to the oxygen release branch and is not part of any currently-released code, I'm not sure that changing it to `ignore` would do anything to fix the issue you're reporting.

marccardinal:
I applied the change described on my local installation and it fixed the get selections. Note that I am running the minions and master on debug log level.

Here is the requested versions report:

```
               Salt: 2018.2.0
     
    Dependency Versions:
               cffi: Not Installed
           cherrypy: Not Installed
           dateutil: Not Installed
          docker-py: Not Installed
              gitdb: Not Installed
          gitpython: Not Installed
              ioflo: Not Installed
             Jinja2: 2.7.2
            libgit2: Not Installed
            libnacl: Not Installed
           M2Crypto: 0.21.1
               Mako: Not Installed
       msgpack-pure: Not Installed
     msgpack-python: 0.3.0
       mysql-python: 1.2.3
          pycparser: Not Installed
           pycrypto: 2.6.1
       pycryptodome: Not Installed
             pygit2: Not Installed
             Python: 2.7.6 (default, Nov 23 2017, 15:49:48)
       python-gnupg: Not Installed
             PyYAML: 3.10
              PyZMQ: 14.0.1
               RAET: Not Installed
              smmap: Not Installed
            timelib: Not Installed
            Tornado: 4.5.3
                ZMQ: 4.0.4
     
    System Versions:
               dist: Ubuntu 14.04 trusty
             locale: ANSI_X3.4-1968
            machine: x86_64
            release: 3.19.0-80-generic
             system: Linux
            version: Ubuntu 14.04 trusty
```

My locale is set to `en_US.UTF-8`

I looked at the other ticket and there is a possibility that it fixes this issue in a cleaner way. I'm OK with not merging and waiting for the other one to be merged.

terminalmage:
@marccardinal Your detected locale is not UTF-8 though. Are you by chance running this in Docker? The Ubuntu base images for Docker don't handle locales well, I had to do so using the following in my Dockerfile:

```
ENV LANG=C.UTF-8
ENV LC_ALL=C.UTF-8
```

Either way though, the fix I submitted in my PR would have decoded correctly, since our `salt.utils.stringutils.to_unicode()` will try the detected locale first and then fall back to UTF-8 if that fails.

Your PR would remove all unicode characters from the output, which is not a great idea.

rallytime:
Thanks for follow up here @marccardinal! I'll go ahead and close this in favor of #45834.

terminalmage:
@marccardinal Thanks for helping bring this to our attention so we could get a fix in.

PR #45758, opened 2018-01-29
support amazon linux 2 for service module

mchugh19:
### What does this PR do?
Disables rh_service.py on Amazon Linux 2

### Previous Behavior
service.running states ran via rh_service.py rather than systemd and failed. 

### New Behavior
Systemd is used to manage service calls and runs properly.

### Tests written?
No

### Commits signed with GPG?
No

### DIFF
--- a/salt/modules/rh_service.py
+++ b/salt/modules/rh_service.py
@@ -84,7 +84,6 @@ def __virtual__():
                     'as virtual \'service\''
                 )
             return __virtualname__
-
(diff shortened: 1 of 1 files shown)


mchugh19:
If possible, it would be great to have this backported to 2017

rallytime:
@mchugh19 It is possible - can you submit this fix directly to 2017.7?

PR #45763, opened 2018-01-29
Fix rehash function in win_path.py

twangboy:
### What does this PR do?
Uses `SendMessageTimeout` instead of `SendMessage` to broadcast an environment change. `SendMessage` was causing the system to hang in some scenarios where an open windows wasn't responding to the `SendMessage`.
Use `ctypes` to access the `user32.dll` directly instead of using pywin32 as this seems to actually work. `win32gui.SendMessageTimeout` would complete successfully but new windows would not reflect the path change.

Found the hanging issue while testing: https://github.com/saltstack/salt/pull/45703#issuecomment-361013728

### Previous Behavior
Sometimes a `win_path.add` or `win_path.remove` would fail

### New Behavior
The refresh will time out after 5 seconds if windows aren't responding

### Tests written?
No

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/modules/win_path.py
+++ b/salt/modules/win_path.py
@@ -12,12 +12,12 @@ from __future__ import absolute_import
 import logging
 import re
 import os
+import ctypes
 from salt.ext.six.moves import map
 
 # Third party libs
 try:
-    from win32con import HWND_BROADCAST, WM_SETTINGCHANGE
-    from win32api import SendMessage
+    from win32con import HWND_BROADCAST, WM_SETTINGCHANGE, SMTO_ABORTIFHUNG
     HAS_WIN32 = True
 except ImportError:
     HAS_WIN32 = False
@@ -47,7 +47,15 @@ def _normalize_dir(string):
 
 def rehash():
     '''
-    Send a WM_SETTINGCHANGE Broadcast to Windows to refresh the Environment variables
+    Send a WM_SETTINGCHANGE Broadcast to Windows to refresh the Environment
+    variables for new processes.
+
+    .. note::
+        This will only affect new processes that aren't launched by services. To
+        apply changes to the path to services, the host must be restarted. The
+        ``salt-minion``, if running as a service, will not see changes to the
+        environment until the system is restarted. See
+        `MSDN Documentation <https://support.microsoft.com/en-us/help/821761/changes-that-you-make-to-environment-variables-do-not-affect-services>`
 
     CLI Example:
 
@@ -55,7 +63,12 @@ def rehash():
 
         salt '*' win_path.rehash
     '''
-    return bool(SendMessage(HWND_BROADCAST, WM_SETTINGCHANGE, 0, 'Environment'))
+    broadcast_message = ctypes.create_unicode_buffer('Environment')
+    user32 = ctypes.windll('user32', use_last_error=True)
+    result = user32.SendMessageTimeoutW(HWND_BROADCAST, WM_SETTINGCHANGE, 0,
+                                        broadcast_message, SMTO_ABORTIFHUNG,
+                                        5000, 0)
+    return result == 1
 
 
 def get_path():
--- a/salt/modules/win_path.py
+++ b/salt/modules/win_path.py
@@ -12,18 +12,17 @@ from __future__ import absolute_import
 import logging
 import re
 import os
-import ctypes
 from salt.ext.six.moves import map
 
 # Third party libs
 try:
-    from win32con import HWND_BROADCAST, WM_SETTINGCHANGE, SMTO_ABORTIFHUNG
     HAS_WIN32 = True
 except ImportError:
     HAS_WIN32 = False
 
 # Import salt libs
 import salt.utils
+import salt.utils.win_functions
 
 # Settings
 log = logging.getLogger(__name__)
@@ -63,12 +62,7 @@ def rehash():
 
         salt '*' win_path.rehash
     '''
-    broadcast_message = ctypes.create_unicode_buffer('Environment')
-    user32 = ctypes.windll('user32', use_last_error=True)
-    result = user32.SendMessageTimeoutW(HWND_BROADCAST, WM_SETTINGCHANGE, 0,
-                                        broadcast_message, SMTO_ABORTIFHUNG,
-                                        5000, 0)
-    return result == 1
+    return salt.utils.win_functions.refresh_environment()
 
 
 def get_path():
--- a/salt/modules/win_path.py
+++ b/salt/modules/win_path.py
@@ -54,7 +54,7 @@ def rehash():
         apply changes to the path to services, the host must be restarted. The
         ``salt-minion``, if running as a service, will not see changes to the
         environment until the system is restarted. See
-        `MSDN Documentation <https://support.microsoft.com/en-us/help/821761/changes-that-you-make-to-environment-variables-do-not-affect-services>`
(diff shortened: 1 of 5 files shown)


salt/modules/win_path.py
```diff
@@ -47,15 +47,28 @@ def _normalize_dir(string):
...
 
         salt '*' win_path.rehash
     '''
-    return bool(SendMessage(HWND_BROADCAST, WM_SETTINGCHANGE, 0, 'Environment'))
+    broadcast_message = ctypes.create_unicode_buffer('Environment')
+    user32 = ctypes.windll('user32', use_last_error=True)
+    result = user32.SendMessageTimeoutW(HWND_BROADCAST, WM_SETTINGCHANGE, 0,
+                                        broadcast_message, SMTO_ABORTIFHUNG,
+                                        5000, 0)
+    return result == 1
```
damon-atkins: # https://msdn.microsoft.com/en-us/library/windows/desktop/ms644952(v=vs.85).aspx
    _, res = win32gui.SendMessageTimeout(
        win32con.HWND_BROADCAST, win32con.WM_SETTINGCHANGE, 0, 0,
        win32con.SMTO_ABORTIFHUNG, 5000)

twangboy: The win32gui one didn't work for me. We should move this into a salt.util if it's being used in two places.

damon-atkins: Move this and reg one into win_functions. 
i.e. win_functions should be common functions shared across the platform.

damon-atkins: I approve this change as is, or update it,

twangboy: Added `refresh_environment` to `salt.utils.win_functions`.
Pointed the `rehash` function in `win_path.py` to the new salt util
Pointed the `broadcast_change` function in `reg.py` to the new salt util

twangboy:
@rallytime Rebased

damon-atkins:
https://msdn.microsoft.com/en-us/library/windows/desktop/ms644927(v=vs.85).aspx#system_defined
Could not find 'Environment' in the doco for the function.  However I would make 
`def refresh_environment(): more generic`  e.g. `def broadcast_setting_change('Environment'):`   If that makes sense given what SendMessageTimeoutW does.

twangboy:
@damon-atkins Changed to a more generic `broadcast_setting_change` function. Please check this again.

salt/utils/win_functions.py
```diff
@@ -214,10 +214,24 @@ def escape_meta_chars(m):
...
+            updated and needs to be refreshed. Default is ``Environment``. These
+            are some common values:
+
+            - "Environment" : to effect a change in the environment variables
+            - "intl" : to effect a change in locale settings
+            - "Policy" : to effect a change in Group Policy Settings
+            - a leaf node in the registry
+            - the name of a section in the ``Win.ini`` file
+
+            `See here <https://msdn.microsoft.com/en-us/library/ms725497%28VS.85%29.aspx>`
```
damon-atkins: Suggested Changes:
def broadcast_setting_change(message='Environment')

Sends message to all top-level windows, only those application windows listing for the particular message sent will react.  Check with application developers for a particular application, however the following is a list of common broadcast messages some application windows may listen for:

See  lParam witin .... for sending and   GWL_WNDPROC within https://msdn.microsoft.com/en-us/library/windows/desktop/ms633591(v=vs.85).aspx for receiving.

May include this as a comment in the code
Python listener would involve old_function = win32gui.SetWindowLong(window_handle,win32con.GWL_WNDPROC, new_function
To give people a glue if they are trying to do something like this.

End of Suggestions.

It seems its just a message which is sent, and the expectation is something is listening for it.  The string could be anything and the app could be listening for anything.    You could even send "salt-minion" and have the salt-minion listen for it and do a high-state.

damon-atkins (changes requested):

salt/utils/win_functions.py
twangboy: Changed the name of the parameter to message. I left the default as "Environment". I believe that to be a reasonable default.

twangboy:
@damon-atkins Please review again.

salt/utils/win_functions.py
damon-atkins: Update the text place I think it explains it better.  Sends ...

twangboy:
@damon-atkins OK, I added the code comments and additional docs. Please review.

damon-atkins (approved):

rallytime:
@twangboy This is causing a related test to fail: https://jenkins.saltstack.com/job/PR/job/salt-pr-linode-ubuntu14-n/19705/

damon-atkins:
@rallytime  not sure how a windows change effects a Linux platform?

rallytime:
@damon-atkins None of these files are loaded on linux.

rallytime:
This change needs tests.

salt/utils/win_functions.py
```diff
@@ -210,3 +212,72 @@ def escape_meta_chars(m):
...
+            for information on how to retrieve those messages.
+
+    .. note::
+        This will only affect new processes that aren't launched by services. To
+        apply changes to the path or registry to services, the host must be
+        restarted. The ``salt-minion``, if running as a service, will not see
+        changes to the environment until the system is restarted. Services
+        inherit their environment from ``services.exe`` which does not respond
+        to messaging events. See
+        `MSDN Documentation <https://support.microsoft.com/en-us/help/821761/changes-that-you-make-to-environment-variables-do-not-affect-services>`
```
rallytime: This link syntax doesn't look quite right. I think you need a `_`. http://docutils.sourceforge.net/docs/user/rst/quickref.html#external-hyperlink-targets

This goes for all of the other places you have links in the docs for this function, and in the `rehash` note above.

salt/utils/win_functions.py
```diff
@@ -210,3 +212,72 @@ def escape_meta_chars(m):
...
+        import salt.utils.win_functions
+        salt.utils.win_functions.broadcast_setting_change('Environment')
+    '''
+    # Listen for messages sent by this would involve working with the
+    # SetWindowLong function. This can be accessed via the win32gui or through
+    # ctypes. You can find examples on how to do this by searching for
+    # `Accessing WGL_WNDPROC` on the internet. Here are some examples of how
+    # this might work:
+    #
+    # # using win32gui
```
rallytime: Why are all of these lines commented out? Do they need to be here at all?

damon-atkins:
@rallytime  please read code review comments regarding comments

salt/utils/win_functions.py
damon-atkins: Ping @rallytime

salt/utils/win_functions.py
twangboy: That's if you're linking to the links later on. I was following the sphinx documentation.
http://www.sphinx-doc.org/en/stable/rest.html#external-links

Does that not work for inline docs?

salt/utils/win_functions.py
twangboy: These are examples of how to access the broadcast message.  In case someone in the future needs to read the message broadcast by this function. @damon-atkins suggested that this be added,

salt/utils/win_functions.py
lorengordon: You still need the final `_` to create the external link: 

```
`Link text <http://example.com/>`_
```

twangboy:
@rallytime Fixed the hyperlink thing

rallytime (approved):

PR #45764, opened 2018-01-29
support amazon linux 2 for service module

mchugh19:
### What does this PR do?
Same as #45758, but for the 2017.7 branch. Disables rh_service.py on Amazon Linux 2

### Previous Behavior
service.running states ran via rh_service.py rather than systemd and failed. 

### New Behavior
Systemd is used to manage service calls and runs properly.

### Tests written?
No

### Commits signed with GPG?
No

### DIFF
--- a/salt/modules/rh_service.py
+++ b/salt/modules/rh_service.py
@@ -97,6 +97,15 @@ def __virtual__():
                     'RedHat-based distros >= version 7 use systemd, will not '
                     'load rh_service.py as virtual \'service\''
                 )
+        if __grains__['os'] == 'Amazon':
+            if int(osrelease_major) in (2016, 2017):
+                return __virtualname__
+            else:
+                return (
+                    False,
+                    'Amazon Linux >= version 2 use systemd, will not '
+                    'load rh_service.py as virtual \'service\''
+                )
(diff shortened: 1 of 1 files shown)


salt/modules/rh_service.py
```diff
@@ -97,6 +97,15 @@ def __virtual__():
                     'RedHat-based distros >= version 7 use systemd, will not '
                     'load rh_service.py as virtual \'service\''
                 )
+        if __grains__['os'] == 'Amazon':
+            if int(osrelease_major) in (2016, 2017):
+                return __virtualname__
+            else:
+                return (
+                    False,
+                    'Amazon Linux >= version 2 use systemd, will not '
```
rallytime: `use` --> `uses` and then maybe let's split the message up into 2 sentences with `systemd. Will not`.

rallytime (changes requested):
I have one small request

mchugh19:
Edited the copy/pasted bits!

rallytime (approved):

terminalmage (approved):

terminalmage:
@rallytime I think this should be added to the RC branch.

PR #42476, opened 2017-07-21
Fix possible exceptions in the loader for boto modules

cachedout:
If the minimum dependencies in the boto* utility modules are not met,
the execution modules will end up bubbling those exceptions up to the loader.

This checks to see if the needed utility functions are present and if they
are not, do not load the execution module.

### Previous Behavior
Remove this section if not relevant

### New Behavior
Remove this section if not relevant

### Tests written?

Yes/No

Please review [Salt's Contributing Guide](https://docs.saltstack.com/en/latest/topics/development/contributing.html) for best practices.

PR #45771, opened 2018-01-29
Add a central function to check for boto lib requirements

rallytime:
This was added to salt.utils.versions.py in order for the boto.py
and boto3.py utils files to use it and to avoid adding a kind of
circular check in various __utils__ files for the boto deps.

The arguments added to versions.check_boto_reqs() allow the individual
module to override the boto, boto3, and/or botocore minimum versions
where the default values are defined in versions.check_boto_reqs().

Replaces PR #42476

### DIFF
--- a/salt/modules/boto3_elasticache.py
+++ b/salt/modules/boto3_elasticache.py
@@ -52,9 +52,11 @@ import logging
 import time
 
 # Import Salt libs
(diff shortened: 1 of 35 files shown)


cachedout (approved):
Nicely done.

PR #45777, opened 2018-01-30
Fix an inconsistency in the docs

while0pass:
### What does this PR do?
Fixes an inconsistency in the docs, where there are 2 occurencies of the function ``clear_list_cache``. According to the code of ``salt.runners.fileserver`` its name should be ``clear_file_list_cache`` instead.

No issues.

### Tests written?
No

### Commits signed with GPG?
No

### DIFF
--- a/doc/ref/configuration/master.rst
+++ b/doc/ref/configuration/master.rst
@@ -2470,8 +2470,9 @@ on a large number of minions.
 
 .. note::
     Rather than altering this configuration parameter, it may be advisable to
-    use the :mod:`fileserver.clear_list_cache
-    <salt.runners.fileserver.clear_list_cache>` runner to clear these caches.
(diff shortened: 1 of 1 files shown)


doc/man/salt.7
```diff
@@ -7095,7 +7095,7 @@ on a large number of minions.
 .INDENT 0.0
 .INDENT 3.5
 Rather than altering this configuration parameter, it may be advisable to
-use the \fBfileserver.clear_list_cache\fP runner to clear these caches.
+use the \fBfileserver.clear_file_list_cache\fP runner to clear these caches.
```
rallytime: This is fine, but a note for the future that the man pages are autogenerated based on the changes in the docs. Updating the man pages is not necessary.

rallytime (approved):
Good catch @while0pass! Thank you.

PR #45778, opened 2018-01-30
Support comma separated minion list target in payload for TCP transport

DmitryKuzmenko:
### What does this PR do?
Support comma separated minion list target in payload for TCP transport.

We have no problems with it because `salt` cmdline transforms comma-separated target list (when `-L` is set) to the python list object and sets `tgt` in payload to the list.
But the failing `integration.client.test_standard.StdTest.test_missing_minion_list` uses `client.cmd` function and passes comma-separated string to the client. ZeroMQ transport handles this correctly with `ckminions.check_minions` when TCP transport lacks this and tries to iterate though the string during publishing the job over minions.

Fixes https://github.com/saltstack/salt-jenkins/issues/813

### Tests written?
Exists


cachedout (approved):
LGTM

PR #45779, opened 2018-01-30
SSH shell shim: Don't use $() for optimal support

The-Loeki:
Some shells don't support the `$()` syntax, so use the backticks instead.
### Previous Behavior
```
this-posix-compliant-shell-oldskool.example.com:
    ----------
    retcode:
        2
    stderr:
        SUDO=
        + [ -n  ] 
        SUDO_USER=
        + [  ] 
        + [  ] 
        EX_PYTHON_INVALID=10
        PYTHON_CMDS=python3 python27 python2.7 python26 python2.6 python2 python
        /bin/sh: syntax error at line 28: `cmdpath=$' unexpected
    stdout:
```

### DIFF
--- a/salt/client/ssh/__init__.py
+++ b/salt/client/ssh/__init__.py
@@ -157,17 +157,17 @@ do
         py_cmd_path=`"$py_cmd" -c \
                    'from __future__ import print_function;
                    import sys; print(sys.executable);'`
-        cmdpath=$(command -v $py_cmd 2>/dev/null || which $py_cmd 2>/dev/null)
+        cmdpath=`command -v $py_cmd 2>/dev/null || which $py_cmd 2>/dev/null`
         if file $cmdpath | grep "shell script" > /dev/null
         then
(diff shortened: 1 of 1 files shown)


salt/client/ssh/__init__.py
```diff
@@ -157,17 +157,17 @@
         py_cmd_path=`"$py_cmd" -c \
                    'from __future__ import print_function;
                    import sys; print(sys.executable);'`
-        cmdpath=$(command -v $py_cmd 2>/dev/null || which $py_cmd 2>/dev/null)
+        cmdpath=`command -v $py_cmd 2>/dev/null || which $py_cmd 2>/dev/null`
```
cachedout: The only gotcha I'm aware of here would be that the backticks would strip double-quotes but since we have a statically defined list of single-word strings, we should be just fine. Looks good!

cachedout (approved):

rallytime:
re-run lint

PR #45780, opened 2018-01-30
[DOC] Add missing gpgautoimport for pkgrepo.managed

vutny:
### What does this PR do?
It adds missing `gpgautoimport` parameter to the docstring for `pkgrepo.managed` state that is supported on Zypper-based distros via `zypper.mod_repo` function.

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/modules/zypper.py
+++ b/salt/modules/zypper.py
@@ -728,10 +728,10 @@ def mod_repo(repo, **kwargs):
     be created, so long as the following values are specified:
 
     repo or alias
-        alias by which the zypper refers to the repo
+        alias by which Zypper refers to the repo
 
     url, mirrorlist or baseurl
-        the URL for zypper to reference
+        the URL for Zypper to reference
 
     enabled
         Enable or disable (True or False) repository,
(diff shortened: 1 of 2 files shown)


salt/modules/zypper.py
```diff
@@ -728,10 +728,10 @@ def mod_repo(repo, **kwargs):
     be created, so long as the following values are specified:
 
     repo or alias
-        alias by which the zypper refers to the repo
+        alias by which the Zypper refers to the repo
```
rallytime: This is picky, but since you're in here, `the` isn't needed in this sentence before `Zypper`.

salt/modules/zypper.py
```diff
@@ -728,10 +728,10 @@ def mod_repo(repo, **kwargs):
...
 
     url, mirrorlist or baseurl
-        the URL for zypper to reference
+        the URL for the Zypper to reference
```
rallytime: Same comment here.

salt/modules/zypper.py
```diff
@@ -747,11 +747,12 @@ def mod_repo(repo, **kwargs):
         Enable or disable (True or False) GPG check for this repository.
 
     gpgautoimport
-        Automatically trust and import new repository.
+        If set to True, automatically trust and import public GPG key for
```
rallytime: This seems to imply that `gpgautimport` defaults to `False`, but maybe we should be explicit about that information and say directly what the default is.

vutny:
Thank for the review @rallytime ! The articles is really my weak point in English grammar 😄 
Fixed that.

rallytime (approved):
Looks great now. Thanks @vutny!

PR #45781, opened 2018-01-30
Allow gpg renderer to render multiple ciphertexts in a single string

ignatk:
### What does this PR do?
This PR allows to encode more than one secret in a single pillar value. Moreover, it prevents `gpg` renderer stripping surrounding unencrypted text.

With current behaviour it is not possible to use `gpg` renderer in the middle of a renderer pipe-chain:
* you either have to use it first (encrypting the whole file and chaining other renderers afterwards), but it is not very straightforward to modify the pillar structure as you have to decrypt the whole `sls`, do modifications and encrypt again
* or use it last in the chain, but then you cannot have pillar values, which contain multiple secrets in a single value. Also, you can't have secrets mixed with plaintext data, because all plaintext data is stripped upon decryption
* because of the above, you cannot have a renderer, which does additional processing of data chained after `gpg` renderer, for example:

```
#!pgp|xml

<foo>
  <bar>
    -----BEGIN PGP MESSAGE-----<super key here>-----END PGP MESSAGE-----
  </bar>
</foo>
```

### Previous Behavior
pillar:
```
#!yaml|gpg

foo:
  bar: |
    this is my complex pillar value
    it has a password:
    -----BEGIN PGP MESSAGE-----<encrypted password>-----END PGP MESSAGE-----
    and salt:
    -----BEGIN PGP MESSAGE-----<encrypted salt>-----END PGP MESSAGE-----
```
will be rendered as
```
foo:
  bar: |
    <decrypted password>
```
`gpg` will render only the first ciphertext and strip the rest of plaintext and the second ciphertext

### New Behavior
pillar:
```
#!yaml|gpg

foo:
  bar: |
    this is my complex pillar value
    it has a password:
    -----BEGIN PGP MESSAGE-----<encrypted password>-----END PGP MESSAGE-----
    and salt:
    -----BEGIN PGP MESSAGE-----<encrypted salt>-----END PGP MESSAGE-----
```
will be rendered as
```
  bar: |
    this is my complex pillar value
    it has a password:
    <decrypted password>
    and salt:
    <decrypted salt>
```

### Tests written?
Yes

### Commits signed with GPG?
No

### DIFF
--- a/salt/renderers/gpg.py
+++ b/salt/renderers/gpg.py
@@ -226,7 +226,8 @@ from salt.ext import six
 
 log = logging.getLogger(__name__)
 
-GPG_HEADER = re.compile(r'-----BEGIN PGP MESSAGE-----')
+GPG_CIPHERTEXT = re.compile(
+    r'-----BEGIN PGP MESSAGE-----.*?-----END PGP MESSAGE-----', re.DOTALL)
 
 
 def _get_gpg_exec():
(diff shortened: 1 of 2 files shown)


garethgreenaway (approved):

terminalmage (approved):

secumod:
rebased on latest `develop`

PR #39948, opened 2017-03-10
Fix race condition in Salt loader.

smarsching:
### What does this PR do?

There was a race condition in the salt loader when injecting global values (e.g. "__pillar__" or "__salt__") into modules. One effect of this race condition was that in a setup with multiple threads, some threads may see pillar data intended for other threads or the pillar data seen by a thread might even change spuriously.

There have been earlier attempts to fix this problem (#27937, #29397). These patches tried to fix the problem by storing the dictionary that keeps the relevant data in a thread-local variable and referencing this thread-local variable from the variables that are injected into the modules.

These patches did not fix the problem completely because they only work when a module is loaded through a single loader instance only. When there is more than one loader, there is more than one thread-local variable and the variable injected into a module is changed to point to another thread-local variable when the module is loaded again. Thus, the problem resurfaced while working on #39670.

This patch attempts to solve the problem from a slightly different angle, complementing the earlier patches: The value injected into the modules now is a proxy that internally uses a thread-local variable to decide to which object it points. This means that when loading a module
again through a different loader (possibly passing different pillar data), the data is actually only changed in the thread in which the loader is used. Other threads are not affected by such a change.

This means that it will work correctly in the current situation where loaders are possibly created by many different modules and these modules do not necessary know in which context they are executed. Thus it is much more flexible and reliable than the more explicit approach
used by the two earlier patches.

### What issues does this PR fix or reference?

This PR fixes problems that surfaced when developing the parallel runners feature (#39670), but is also related to problems reported earlier (#23373). The problems described in #29028 might also be related, though this needs further confirmation.

### Previous Behavior
Changes to pillar data in one thread might have influenced pillar data in a different thread. Whether or when this problem actually appeared was unpredictable (due to the nature of a race condition).

### New Behavior
Changes to pillar data in one thread will never affect the pillar data in a different thread.

### Tests written?

No

### Regression Potential

The change to the loader code itself is rather small, but it affects a central component. So if there is a problem, the impact might affect nearly all components. For this reason I dubmit this PR for the develop branch, even though it addresses a bug that is actually also present in older releases.

Having said this, I believe that these changes should not cause any problems because to single-threaded processes, everything behaves as before, and for multi-threaded processes, the behavior is now more predictable. I tested this patch on our Salt master (I backported it to 2011.3 for this purpose) and I did not experience any issues there.

PR #39670, opened 2017-02-26
Added the saltmod.parallel_runners state.

smarsching:
### What does this PR do?

This new state function is intended for use with the orchestrate runner. It is used in a way very similar to saltmod.runner, except that it executes multiple runners in parallel.

### What issues does this PR fix or reference?

This PR adds a general feature and is not particularly focused on fixing a single issue. However, I believe that at least some of the use cases discussed in #32488 and #32956 are covered by this new feature.

The use case that I wrote it for is orchestrating actions where some can run in parallel and some have to run after each other. In my case, I want to run a system upgrade on all VMs hosted on a number of VM hosts. As running a system upgrade generates a significant I/O load, I only want to upgrade one VM per host at a time. However, upgrades on different VM hosts shall run in parallel.

With the existing features of the orchestrate runner, this was not possible (except for launching several instances of the orchestrate runner from the shell, of course). With the new feature, added by this PR, I can now have an SLS file that triggers the execution of a separate runner for each VM host and waits until all of these runners have finished.

The state blocks, until all runners triggered by it have finished. Therefore, it is possible to run certain actions in parallel and then continuing with another action, which depends on all of these actions having finished. Therefore, I believe that with the addition of this new feature, orchestration scenarios with very complex dependencies between actions can be covered correctly.

### Previous Behavior
The `saltmod.parallel_runners` state function did not exist.

### New Behavior
The `saltmod.parallel_runners` state function has been added.

### Tests written?

No

Issue #29028, opened 2015-11-19
state.sls fails to render state with pillar data: Jinja variable 'dict object' has no attribute


kevins9:
This is only occurring on my minions running Salt 2015.5.5 and CentOS 7.  Issue is similar to #16727 and #16656.
##### Version info - master and minion

```
[root@master ~]# cat /etc/redhat-release
CentOS release 6.5 (Final)
[root@master ~]# salt --versions
           Salt: 2015.5.3
         Python: 2.6.6 (r266:84292, Jan 22 2014, 09:42:36)
         Jinja2: 2.7.3
       M2Crypto: 0.20.2
 msgpack-python: 0.1.13
   msgpack-pure: Not Installed
       pycrypto: 2.0.1
        libnacl: Not Installed
         PyYAML: 3.10
          ioflo: Not Installed
          PyZMQ: 2.2.0.1
           RAET: Not Installed
            ZMQ: 3.2.4
           Mako: Not Installed
        Tornado: 2.2.1

[root@centos7test-01 ~]# cat /etc/redhat-release
CentOS Linux release 7.1.1503 (Core)
[root@centos7test-01 ~]# salt-call --versions
           Salt: 2015.5.5
         Python: 2.7.5 (default, Jun 24 2015, 00:41:19)
         Jinja2: 2.7.2
       M2Crypto: 0.21.1
 msgpack-python: 0.4.6
   msgpack-pure: Not Installed
       pycrypto: 2.6.1
        libnacl: Not Installed
         PyYAML: 3.10
          ioflo: Not Installed
          PyZMQ: 14.3.1
           RAET: Not Installed
            ZMQ: 3.2.5
           Mako: 0.8.1
        Tornado: Not Installed
        timelib: Not Installed
       dateutil: 1.5
```
##### Pillar top.sls

```
base:
  'role:bugdemo':
    - match: grain
    - bugdemo
```
##### Pillar bugdemo.sls

```
bugdemo: problem
```
##### minion /etc/salt/grains

```
role: bugdemo
```
##### Master /srv/salt/dev/bugdemo.sls

```
/var/tmp/{{ pillar['bugdemo'] }}:
  file.exists:
    - name: /var/tmp/{{ pillar['bugdemo'] }}
```

(Note: master config has "dev" saltenv configured with root /srv/salt/dev)
##### Evidence

Pillar value is present:

```
[root@centos7test-01 ~]# salt-call pillar.get bugdemo
local:
    problem
```

[...]

Issue #23373, opened 2015-05-05
reactor/orchestrate race condition on salt['pillar.get']


tnypex:
Hi all,

with `2015.2.0rc2`

I have the following reactor sls that calls an orchestrate runner:
```jinja
{% set fun = data.get('fun', {}) %}
{% set id = data.get('id', {}) %}
{% if fun == "test.ping" %}
orchestrate_test_{{id}}:
  runner.state.orchestrate:
    - mods: orchestration.test
    - pillar: { target_server: {{id}} }
{% endif %}
```
When running test.ping on multiple minions (3 in this case)
the reactor sls is rendered fine 3 times :
```yaml
2015-05-05 15:52:40,038 [salt.template                              ][DEBUG   ][25815] Rendered data from file: /srv/salt/base/reactor/jobs_ret.sls:


orchestrate_test_minion1:
  runner.state.orchestrate:
    - mods: orchestration.test
    - pillar: { target_server: minion1 }
```
```yaml	
2015-05-05 15:52:40,128 [salt.template                              ][DEBUG   ][25815] Rendered data from file: /srv/salt/base/reactor/jobs_ret.sls:


orchestrate_test_minion2:
  runner.state.orchestrate:
    - mods: orchestration.test
    - pillar: { target_server: minion2 }
```
```yaml
2015-05-05 15:52:40,261 [salt.template                              ][DEBUG   ][25815] Rendered data from file: /srv/salt/base/reactor/jobs_ret.sls:


orchestrate_test_minion3:
  runner.state.orchestrate:
    - mods: orchestration.test
    - pillar: { target_server: minion3 }
```	
	
however, the orchestrate runners will not always get the correct pillar data (it will sometimes get an empty value, sometimes an other orchestrate runner's pillar data, sometimes the expected value) :

The orchestrate sls:
```jinja
{% set tgs = salt['pillar.get']('target_server') %}

test:
  salt.function:
    - name: test.arg
    - tgt: {{ tgs }}
```

```yaml
2015-05-05 15:52:45,024 [salt.template                              ][DEBUG   ][25815] Rendered data from file: /var/cache/salt/master/files/base/orchestration/test.sls:


test:
  salt.function:
    - name: test.arg
    - tgt: minion1
```
```yaml	
2015-05-05 15:52:45,025 [salt.template                              ][DEBUG   ][25815] Rendered data from file: /var/cache/salt/master/files/base/orchestration/test.sls:


test:
  salt.function:
    - name: test.arg
[...]

PR #45782, opened 2018-01-30
 Fix race condition in Salt loader.

smarsching:
### What does this PR do?

This PR is a new version of #39948, so the description has been copied from there. The only difference to the original PR is that this PR is targeted at the `oxygen` branch instead of the `develop` branch.

There was a race condition in the salt loader when injecting global values (e.g. "__pillar__" or "__salt__") into modules. One effect of this race condition was that in a setup with multiple threads, some threads may see pillar data intended for other threads or the pillar data seen by a thread might even change spuriously.

There have been earlier attempts to fix this problem (#27937, #29397). These patches tried to fix the problem by storing the dictionary that keeps the relevant data in a thread-local variable and referencing this thread-local variable from the variables that are injected into the modules.

These patches did not fix the problem completely because they only work when a module is loaded through a single loader instance only. When there is more than one loader, there is more than one thread-local variable and the variable injected into a module is changed to point to another thread-local variable when the module is loaded again. Thus, the problem resurfaced while working on #39670.

This patch attempts to solve the problem from a slightly different angle, complementing the earlier patches: The value injected into the modules now is a proxy that internally uses a thread-local variable to decide to which object it points. This means that when loading a module
again through a different loader (possibly passing different pillar data), the data is actually only changed in the thread in which the loader is used. Other threads are not affected by such a change.

This means that it will work correctly in the current situation where loaders are possibly created by many different modules and these modules do not necessary know in which context they are executed. Thus it is much more flexible and reliable than the more explicit approach
used by the two earlier patches.

This PR fixes problems that surfaced when developing the parallel runners feature (#39670), but is also related to problems reported earlier (#23373). The problems described in #29028 might also be related, though this needs further confirmation.

### Previous Behavior
Changes to pillar data in one thread might have influenced pillar data in a different thread. Whether or when this problem actually appeared was unpredictable (due to the nature of a race condition).

### New Behavior
Changes to pillar data in one thread will never affect the pillar data in a different thread.

### Tests written?

No

### Regression Potential

The change to the loader code itself is rather small, but it affects a central component. So if there is a problem, the impact might affect nearly all components.

### DIFF
--- a/salt/loader.py
+++ b/salt/loader.py
@@ -13,6 +13,7 @@ import time
 import logging
 import inspect
 import tempfile
+import threading
 import functools
 import types
 from collections import MutableMapping
@@ -28,6 +29,7 @@ import salt.utils.files
 import salt.utils.lazy
 import salt.utils.odict
 import salt.utils.platform
+import salt.utils.thread_local_proxy
 import salt.utils.versions
 from salt.exceptions import LoaderError
 from salt.template import check_render_pipe_str
@@ -1003,6 +1005,76 @@ def _mod_type(module_path):
     return 'ext'
 
 
+def _inject_into_mod(mod, name, value, force_lock=False):
+    '''
+    Inject a variable into a module. This is used to inject "globals" like
+    ``__salt__``, ``__pillar``, or ``grains``.
+
+    Instead of injecting the value directly, a ``ThreadLocalProxy`` is created.
+    If such a proxy is already present under the specified name, it is updated
+    with the new value. This update only affects the current thread, so that
+    the same name can refer to different values depending on the thread of
+    execution.
+
+    This is important for data that is not truly global. For example, pillar
+    data might be dynamically overriden through function parameters and thus
+    the actual values available in pillar might depend on the thread that is
+    calling a module.
+
+    mod:
+        module object into which the value is going to be injected.
+
+    name:
+        name of the variable that is injected into the module.
+
+    value:
+        value that is injected into the variable. The value is not injected
+        directly, but instead set as the new reference of the proxy that has
+        been created for the variable.
+
+    force_lock:
+        whether the lock should be acquired before checking whether a proxy
+        object for the specified name has already been injected into the
+        module. If ``False`` (the default), this function checks for the
+        module's variable without acquiring the lock and only acquires the lock
+        if a new proxy has to be created and injected.
+    '''
+    from salt.utils.thread_local_proxy import ThreadLocalProxy
+    old_value = getattr(mod, name, None)
+    # We use a double-checked locking scheme in order to avoid taking the lock
+    # when a proxy object has already been injected.
+    # In most programming languages, double-checked locking is considered
+    # unsafe when used without explicit memory barriers because one might read
+    # an uninitialized value. In CPython it is safe due to the global
+    # interpreter lock (GIL). In Python implementations that do not have the
+    # GIL, it could be unsafe, but at least Jython also guarantees that (for
+    # Python objects) memory is not corrupted when writing and reading without
+    # explicit synchronization
+    # (http://www.jython.org/jythonbook/en/1.0/Concurrency.html).
+    # Please note that in order to make this code safe in a runtime environment
+    # that does not make this guarantees, it is not sufficient. The
+    # ThreadLocalProxy must also be created with fallback_to_shared set to
+    # False or a lock must be added to the ThreadLocalProxy.
+    if force_lock:
+        with _inject_into_mod.lock:
+            if isinstance(old_value, ThreadLocalProxy):
+                ThreadLocalProxy.set_reference(old_value, value)
+            else:
+                setattr(mod, name, ThreadLocalProxy(value, True))
+    else:
+        if isinstance(old_value, ThreadLocalProxy):
+            ThreadLocalProxy.set_reference(old_value, value)
+        else:
+            _inject_into_mod(mod, name, value, True)
+
+
+# Lock used when injecting globals. This is needed to avoid a race condition
+# when two threads try to load the same module concurrently. This must be
+# outside the loader because there might be more than one loader for the same
+# namespace.
+_inject_into_mod.lock = threading.RLock()
+
+
 # TODO: move somewhere else?
 class FilterDictWrapper(MutableMapping):
     '''
@@ -1493,7 +1565,7 @@ class LazyLoader(salt.utils.lazy.LazyDict):
 
         # pack whatever other globals we were asked to
         for p_name, p_value in six.iteritems(self.pack):
-            setattr(mod, p_name, p_value)
+            _inject_into_mod(mod, p_name, p_value)
 
         module_name = mod.__name__.rsplit('.', 1)[-1]
 
--- a/salt/log/handlers/fluent_mod.py
+++ b/salt/log/handlers/fluent_mod.py
@@ -97,14 +97,17 @@ log = logging.getLogger(__name__)
 try:
     # Attempt to import msgpack
     import msgpack
+    import salt.utils.msgpack
     # There is a serialization issue on ARM and potentially other platforms
     # for some msgpack bindings, check for it
     if msgpack.loads(msgpack.dumps([1, 2, 3]), use_list=True) is None:
         raise ImportError
+    import salt.utils.msgpack
 except ImportError:
     # Fall back to msgpack_pure
     try:
         import msgpack_pure as msgpack
+        import salt.utils.msgpack
     except ImportError:
         # TODO: Come up with a sane way to get a configured logfile
         #       and write to the logfile when this error is hit also
@@ -456,7 +459,7 @@ class FluentSender(object):
         packet = (tag, timestamp, data)
         if self.verbose:
             print(packet)
-        return msgpack.packb(packet)
+        return salt.utils.msgpack.packb(packet, _msgpack_module=msgpack)
 
     def _send(self, bytes_):
         self.lock.acquire()
--- a/salt/utils/thread_local_proxy.py
+++ b/salt/utils/thread_local_proxy.py
@@ -1,0 +1,738 @@
+# -*- coding: utf-8 -*-
+'''
+Proxy object that can reference different values depending on the current
+thread of execution.
+
+..versionadded:: Oxygen
+
+'''
+
+# Import python libs
+from __future__ import absolute_import
+import threading
+
+# Import 3rd-party libs
+from salt.ext import six
+
+
+# There are certain types which are sequences, but actually represent string
+# like objects. We need a list of these types for the recursive unproxy code.
+_STRING_LIKE_TYPES = (six.binary_type, six.string_types, six.text_type)
+
+
+class ThreadLocalProxy(object):
+    '''
+    Proxy that delegates all operations to its referenced object. The referenced
+    object is hold through a thread-local variable, so that this proxy may refer
+    to different objects in different threads of execution.
+
+    For all practical purposes (operators, attributes, `isinstance`), the proxy
+    acts like the referenced object. Thus, code receiving the proxy object
+    instead of the reference object typically does not have to be changed. The
+    only exception is code that explicitly uses the ``type()`` function for
+    checking the proxy's type. While `isinstance(proxy, ...)` will yield the
+    expected results (based on the actual type of the referenced object), using
+    something like ``issubclass(type(proxy), ...)`` will not work, because
+    these tests will be made on the type of the proxy object instead of the
+    type of the referenced object. In order to avoid this, such code must be
+    changed to use ``issubclass(type(ThreadLocalProxy.unproxy(proxy)), ...)``.
+
+    If an instance of this class is created with the ``fallback_to_shared`` flag
+    set and a thread uses the instance without setting the reference explicitly,
+    the reference for this thread is initialized with the latest reference set
+    by any thread.
+
+    This class has primarily been designed for use by the Salt loader, but it
+    might also be useful in other places.
+    '''
+
+    __slots__ = ['_thread_local', '_last_reference', '_fallback_to_shared']
+
+    @staticmethod
+    def get_reference(proxy):
+        '''
+        Return the object that is referenced by the specified proxy.
+
+        If the proxy has not been bound to a reference for the current thread,
+        the behavior depends on th the ``fallback_to_shared`` flag that has
+        been specified when creating the proxy. If the flag has been set, the
+        last reference that has been set by any thread is returned (and
+        silently set as the reference for the current thread). If the flag has
+        not been set, an ``AttributeError`` is raised.
+
+        If the object references by this proxy is itself a proxy, that proxy is
+        returned. Use ``unproxy`` for unwrapping the referenced object until it
+        is not a proxy.
+
+        proxy:
+            proxy object for which the reference shall be returned. If the
+            specified object is not an instance of `ThreadLocalProxy`, the
+            behavior is unspecified. Typically, an ``AttributeError`` is
+            going to be raised.
+        '''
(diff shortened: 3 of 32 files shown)


terminalmage (approved):

DmitryKuzmenko (approved):
Awesome work!
Just one minor change in doc.

gtmanfred:
Does this pr need to be pointed at oxygen.rc1?

salt/utils/thread_local_proxy.py
```diff
@@ -0,0 +1,691 @@
+# -*- coding: utf-8 -*-
+'''
+Proxy object that can reference different values depending on the current
+thread of execution.
+
+..versionadded:: Nitrogen
```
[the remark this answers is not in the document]
smarsching: You are right. This simply points to how old origins of this PR are. 😃 I fixed it.

rallytime:
@gtmanfred No, when I talked to mike about this, we decided to have this go into oxygen for the .1 release since we're hardening the branch right now.

gtmanfred:
cool :+1: sounds good

gtmanfred:
@smarsching would you mind rebasing this? it looks like it should be pretty simple, just was adding encoding to the msgpack stuff in salt.utils.cloud

smarsching:
@gtmanfred I rebased it to the newest version of the `oxygen` branch and used the chance to squash the two commits, so that now there is only one single and clean commit.

cachedout:
Thus far I am OK with this but I want to sit down with @thatch45 and discuss this line-by-line before I approve this. These are substantial changes but overall, I think it looks pretty good.

salt/log/handlers/fluent_mod.py
```diff
@@ -97,14 +97,17 @@
 try:
     # Attempt to import msgpack
     import msgpack
+    import salt.utils.msgpack
```
s0undt3ch: Why import if we're not using it?

salt/log/handlers/fluent_mod.py
```diff
@@ -456,7 +459,7 @@ def _make_packet(self, label, timestamp, data):
         packet = (tag, timestamp, data)
         if self.verbose:
             print(packet)
-        return msgpack.packb(packet)
+        return salt.utils.msgpack.packb(packet, _msgpack_module=msgpack)
```
s0undt3ch: Oh, we are using it here...

s0undt3ch (approved):

cachedout (approved):

mbologna:
While setting up a salt-minion with this patch applied we encountered the following errors:

```
 (remote-exec): File "/usr/lib/python2.7/site-packages/salt/utils/thread_local_proxy.py", line 319, in __nonzero__
 (remote-exec): return bool(reference)
 (remote-exec): File "/usr/lib/python2.7/site-packages/salt/utils/thread_local_proxy.py", line 319, in __nonzero__
 (remote-exec): return bool(reference)
 (remote-exec): File "/usr/lib/python2.7/site-packages/salt/utils/thread_local_proxy.py", line 319, in __nonzero__
 (remote-exec): return bool(reference)
 (remote-exec): File "/usr/lib/python2.7/site-packages/salt/utils/thread_local_proxy.py", line 318, in __nonzero__
 (remote-exec): reference = ThreadLocalProxy.get_reference(self)
 (remote-exec): File "/usr/lib/python2.7/site-packages/salt/utils/thread_local_proxy.py", line 69, in get_reference
 (remote-exec): thread_local = object.__getattribute__(proxy, '_thread_local')
 (remote-exec): RuntimeError: maximum recursion depth exceeded while calling a Python object
```

smarsching:
@mbologna Good catch! I am not sure why this problem did not surface during my tests (maybe because I tested them with an older Salt version).

I have pushed two further commits. The first one fixes the issue that you found, the second one fixes a related issue. I am not sure whether the second one has any practical significance (it is quite possible that having such a recursive data structure would lead to the same problem in the serialization code), but I think it was still worth fixing.

I also included unit tests for both issues in order to ensure that they are actually fixed.

Could you please test it again with the revised patch?

rallytime:
@smarsching Ah, this needs a rebase. Can you fix that? Looks like there might be a lint error, too, but the build has timed out.

Once this is rebased and errors are fixed, we can see if @cachedout and @thatch45 can go through this one last time. :)

smarsching:
@rallytime I rebased the branch and fixed the lint problems.

rallytime:
re-run ubuntu

rallytime:
@smarsching Looks like something isn't _quite_ right here. I am seeing this stacktrace right when the test daemons start up:
```
20:10:49  * Starting salt-master ... 20:10:48,977 [salt.log.setup                           :1165][ERROR   ] An un-handled exception was caught by salt's global exception handler:
20:10:49 AttributeError: 'list' object has no attribute 'iteritems'
20:10:49 Traceback (most recent call last):
20:10:49   File "/tmp/salt-tests-tmpdir/scripts/cli_run.py", line 27, in <module>
20:10:49     salt_run()
20:10:49   File "/testing/salt/scripts.py", line 412, in salt_run
20:10:49     client.run()
20:10:49   File "/testing/salt/cli/run.py", line 40, in run
20:10:49     ret = runner.run()
20:10:49   File "/testing/salt/runner.py", line 271, in run
20:10:49     daemonize=False)
20:10:49   File "/testing/salt/client/mixins.py", line 485, in _proc_function
20:10:49     return self.low(fun, low, full_return=False)
20:10:49   File "/testing/salt/client/mixins.py", line 261, in low
20:10:49     return self._low(fun, low, print_event=print_event, full_return=full_return)
20:10:49   File "/testing/salt/client/mixins.py", line 418, in _low
20:10:49     mminion=self.mminion,
20:10:49   File "/testing/salt/utils/job.py", line 107, in store_job
20:10:49     mminion.returners[savefstr](load['jid'], load)
20:10:49   File "/testing/salt/returners/local_cache.py", line 215, in save_load
20:10:49     serial.dump(clear_load, wfh)
20:10:49   File "/testing/salt/payload.py", line 263, in dump
20:10:49     fn_.write(self.dumps(msg))
20:10:49   File "/testing/salt/payload.py", line 225, in dumps
20:10:49     _msgpack_module=msgpack)
20:10:49   File "/testing/salt/utils/msgpack.py", line 48, in packb
20:10:49     o = salt.utils.thread_local_proxy.ThreadLocalProxy.unproxy_recursive(o)
20:10:49   File "/testing/salt/utils/thread_local_proxy.py", line 180, in unproxy_recursive
20:10:49     return ThreadLocalProxy._unproxy_recursive(obj)
20:10:49   File "/testing/salt/utils/thread_local_proxy.py", line 215, in _unproxy_recursive
20:10:49     or ThreadLocalProxy._needs_unproxy_recursive(obj)):
20:10:49   File "/testing/salt/utils/thread_local_proxy.py", line 290, in _needs_unproxy_recursive
20:10:49     for key, value in six.iteritems(obj))
20:10:49   File "/testing/salt/utils/thread_local_proxy.py", line 290, in <genexpr>
20:10:49     for key, value in six.iteritems(obj))
20:10:49   File "/testing/salt/utils/thread_local_proxy.py", line 290, in _needs_unproxy_recursive
20:10:49     for key, value in six.iteritems(obj))
20:10:49   File "/testing/salt/utils/thread_local_proxy.py", line 290, in <genexpr>
20:10:49     for key, value in six.iteritems(obj))
20:10:49   File "/testing/salt/utils/thread_local_proxy.py", line 298, in _needs_unproxy_recursive
20:10:49     for key, value in six.iteritems(obj))
20:10:49   File "/testing/salt/ext/six.py", line 606, in iteritems
20:10:49     return d.iteritems(**kw)
20:10:49 AttributeError: 'list' object has no attribute 'iteritems'
```

smarsching:
@rallytime Thank you very much. This was a bug that I introduced when fixing the problem reported by @mbologna. I added a unit test for detecting the bug and fixed it. Let's see whether the test suite runs successfully now.

salt/loader.py
```diff
@@ -1003,6 +1005,76 @@ def _mod_type(module_path):
...
+        module. If ``False`` (the default), this function checks for the
+        module's variable without acquiring the lock and only acquires the lock
+        if a new proxy has to be created and injected.
+    '''
+    from salt.utils.thread_local_proxy import ThreadLocalProxy
+    old_value = getattr(mod, name, None)
+    # We use a double-checked locking scheme in order to avoid taking the lock
+    # when a proxy object has already been injected.
+    # In most programming languages, double-checked locking is considered
+    # unsafe when used without explicit memory barries because one might read
```
Jo-Con-El: *barriers*

smarsching: Thanks, I fixed it.
It's good to see that someone actually reads the comments. 😃

smarsching:
@rallytime The Jenkins logs look better now, but there are still some errors (that seem to be related to timeouts). You probably know better whether this is expected or not.

rallytime:
re-run py

rallytime:
@smarsching The master test process is starting just fine now, but the minion process isn't. Can you take another look?
```
14:36:29  * Starting salt-master ... 
                                                                                                                                                                
 * Starting salt-master ... STARTED!
14:36:41  * Starting salt-minion ... 14:36:41,675 [salt.master                                :1056][ERROR   ] Received malformed command {'opts': {'publisher_acl_blacklist': {}}}
14:36:41 14:36:41,676 [salt.transport.zeromq                      :687 ][ERROR   ] Some exception handling a payload from minion
14:36:41 Traceback (most recent call last):
14:36:41   File "/testing/salt/transport/zeromq.py", line 683, in handle_message
14:36:41     ret, req_opts = yield self.payload_handler(payload)
14:36:41 ValueError: need more than 0 values to unpack
14:36:41 14:36:41,676 [tornado.general                            :445 ][ERROR   ] Uncaught exception, closing connection.
14:36:41 Traceback (most recent call last):
14:36:41   File "/usr/lib/python2.7/dist-packages/zmq/eventloop/zmqstream.py", line 437, in _handle_events
14:36:41     self._handle_send()
14:36:41   File "/usr/lib/python2.7/dist-packages/zmq/eventloop/zmqstream.py", line 480, in _handle_send
14:36:41     status = self.socket.send_multipart(msg, **kwargs)
14:36:41   File "/usr/lib/python2.7/dist-packages/zmq/sugar/socket.py", line 238, in send_multipart
14:36:41     return self.send(msg_parts[-1], flags, copy=copy, track=track)
14:36:41   File "socket.pyx", line 565, in zmq.backend.cython.socket.Socket.send (zmq/backend/cython/socket.c:5263)
14:36:41   File "socket.pyx", line 605, in zmq.backend.cython.socket.Socket.send (zmq/backend/cython/socket.c:4981)
14:36:41 TypeError: unicode not allowed, use send_string
14:36:41 14:36:41,677 [tornado.application                        :612 ][ERROR   ] Exception in callback None
14:36:41 Traceback (most recent call last):
14:36:41 
  File "/usr/lib/python2.7/dist-packages/tornado/ioloop.py", line 866, in start
14:36:41     handler_func(fd_obj, events)
14:36:41   File "/usr/lib/python2.7/dist-packages/tornado/stack_context.py", line 275, in null_wrapper
14:36:41     return fn(*args, **kwargs)
14:36:41   File "/usr/lib/python2.7/dist-packages/zmq/eventloop/zmqstream.py", line 437, in _handle_events
14:36:41     self._handle_send()
14:36:41   File "/usr/lib/python2.7/dist-packages/zmq/eventloop/zmqstream.py", line 480, in _handle_send
14:36:41     status = self.socket.send_multipart(msg, **kwargs)
14:36:41   File "/usr/lib/python2.7/dist-packages/zmq/sugar/socket.py", line 238, in send_multipart
14:36:41     return self.send(msg_parts[-1], flags, copy=copy, track=track)
14:36:41   File "socket.pyx", line 565, in zmq.backend.cython.socket.Socket.send (zmq/backend/cython/socket.c:5263)
14:36:41   File "socket.pyx", line 605, in zmq.backend.cython.socket.Socket.send (zmq/backend/cython/socket.c:4981)
14:36:41 TypeError: unicode not allowed, use send_string
14:36:41 
```

smarsching:
@rallytime: After working on this problem for almost a day, I am still puzzled. I have verified that the problem is indeed caused by my patch, but I cannot find the reason *why* it happens. Interestingly, it only appears when running the tests with `--transport=zeromq`.

The problems when starting the Minion seem to be related to the message
> TypeError: Frame 0 ('Some exception handling minion ...) does not support the buffer interface.

in the `master.log` file, but I suspect this message is rather a consequence of the failure rather than a hint to its cause.

I suspect that the root cause of the problem is that the ZeroMQ serialization code somehow sees a proxy object and does not know how to handle it. But instead of throwing a proper exception right there, the code does something strange which then results in the problems at a later point in time. Debugging the whole thing is not made any easier by the fact that most of the logic is asynchronous, so it is hard to track the (real) call paths.

I clearly need more time to understand the whole ZeroMQ transport logic, but at the moment I do not have that time due to some other urgent projects. I hope that I will be able to investigate these problems in a few weeks (maybe earlier if I can find some spare time).

rallytime:
@DmitryKuzmenko We'd like to get this in for a future 2018.3.x release, and @smarsching has spent quite a bit of time trying to get this finished up. @DmitryKuzmenko would you be willing to take a look at this and help?

@mattp- You might be interested in looking at this as well.

mattp-:
@rallytime sure i can take a stab at seeing if i can figure something out, though I'll admit this isn't my wheelhouse. 
On slack @cachedout said this PR might be fixed a different way with https://github.com/saltstack/salt/pull/46641 ; do you have any insight there mike or dmitry?

rallytime:
@mattp- That's totally fine, I just thought you might be interested in this one. That's a good point about #46641.

@smarsching Thoughts on that as well?

smarsching:
First of all, sorry you have not header anything from me about this PR for such a long time. I simply have not found the time to dig into the problems with ZeroMQ yet and at the moment it looks like I am not going to find the time for at least another two or three weeks.

Now more  to the point: @rallytime I do not think that #46641 addresses the same issue. I only took a short glance at the changes introduced by that PR, so I might have missed something, but it looks to me like it mainly addresses a race condition in the loader code itself (that can occur due to the GIL being released in certain situations). It does not address the issue that the loader may inject objects (like the pillar tree) that should be thread local, but effectively are not.

From an architectural point of view, my PR is a bit of a hack that tries to work around the fact that some of the objects injected into modules are basically global variables that represent a context. Obviously, this is not a very good design and when starting from scratch, one would certainly try to avoid something like this, but removing that now would mean to make changes to virtually every module of Salt (including custom modules which are not even available as part of the Salt distribution itself), so introducing the proxy object looks like a good trade-off.

In fact, as far as (pure) Python code is involved, the proxy objects behave (almost) like the original objects so it does not really matter to the code using them. Unfortunately, this not true for Python modules that use C code, because (depending on what the C code does) the concrete type actually does matter to the code.

So far I have (at least I believe so) ensured that proxies are "unpacked" before passing them to serialization code of the Json and MsgPack libraries. However, the same problem seems to apply to the serialization code of ZeroMQ. Unfortunately, I have not figured out yet, where the ZeroMQ code gets the proxied objects. Part of that problem is that I am not very familiar with neither Tornado nor ZeroMQ, so I do not fully understand the data flow yet. Maybe someone with a better understanding of these libraries might have a better chance to figure out how the proxied objects end up in the ZeroMQ serialization layer.

isbm:
@smarsching is there any kind of hope you are going to look into this? You've done lots of work here, but seems like PR is now stale. The problem is still here and either others would need to discard it all and go again themselves, or you just finish it. The World is waiting for you, hero! :wink: 

What you say?

damon-atkins:
Go Go Jenkins!

smarsching:
This PR has been replaced by #50655.

PR #45785, opened 2018-01-30
m/selinux.fcontext_get_policy allow long filespecs

### LINKED ISSUE #45784: SELinux module fcontext_get_policy fails with long regex

When an SELinux `fcontext` entry has a regular expression that is longer than 49 characters, the `modules/selinux.py` module is unable to find it. This is due to the output of `semanage fcontext --list` only containing a single space at the end of the regular expression.

We are expecting the output to contain 2 or more spaces between each item in the output. Here is an example output which we assume to always hold true (comes with the CentOS default policies):

```
/usr/bin/swift-account-reaper                      regular file       system_u:object_r:swift_exec_t:s0 
```

As you can see, after `/usr/bin/swift-account-reaper` there is a number of spaces. However, with longer regular expressions we only get a single space:

```
/usr/share/system-config-selinux/system-config-selinux-dbus\.py regular file       system_u:object_r:semanage_exec_t:s0
```

I can modify the script so that it matches one or more spaces (`[ ]+`) instead of the current two or more (`[ ]{2,}`). However, this then breaks the logic that follows which splits the string into its three component parts (as the file type itself can contain a space).

An idea is to not split the string using the spacer regex and instead split using this logic:

1. Everything up to the first space is the regular expression.
1. Everything after the final space is the context.
1. The leftover bit is the second field, the file type.

This appears to work in practice. The regular expression is not allowed to contain spaces, and I don't believe the context can either, so the logic should hold true in all situations.

I'll put in a PR containing these changes. I'd like some feedback as to whether this "strange" logic is acceptable and whether anyone anticipates this breaking any existing use cases. I've tested it on my state files (which contain regular expressions shorter than 49 characters which did previously work as well as longer ones which did not) and it works.

oarmstrong:
### What does this PR do?
The previous logic of matching the output of `semanage fcontext --list` did not
allow for filespecs that were longer than 49 characters. This was due to the
output of the semanage tool not conforming to the expected output.

We used to expect that the after the filespec would be at least two spaces.
However, with long filespecs there is only a single space separating it and the
next field (the file type).

This modifies the regular expression that we use to match the line to accept one
or more spaces as field delimeters.

However, this causes problems when we attempt to split the three fields into a
python dictionary. We cannot use the same logic as previously of using the field
delimeter as the file type field itself can contain a space. Instead we rely on
the fact that the first and last fields cannot contain spaces and pull those out
and the remaining text is the second field.

Fixes #45784.

### Previous Behavior
See #45784.

### New Behavior
See #45784.

### Tests written?
~~No.~~
Yes.

### Commits signed with GPG?
Yes.

### DIFF
--- a/salt/modules/selinux.py
+++ b/salt/modules/selinux.py
@@ -472,10 +472,10 @@ def fcontext_get_policy(name, filetype=None, sel_type=None, sel_user=None, sel_l
 
     parts = re.match(r'^({filespec}) +([a-z ]+) (.*)$'.format(**{'filespec': re.escape(name)}), current_entry_text)
     ret = {
-        'filespec': parts.group(1),
-        'filetype': parts.group(2),
+        'filespec': parts.group(1).strip(),
+        'filetype': parts.group(2).strip(),
     }
-    ret.update(_context_string_to_dict(parts.group(3)))
+    ret.update(_context_string_to_dict(parts.group(3).strip()))
 
     return ret
 
--- a/salt/modules/selinux.py
+++ b/salt/modules/selinux.py
@@ -456,7 +456,7 @@ def fcontext_get_policy(name, filetype=None, sel_type=None, sel_user=None, sel_l
     '''
     if filetype:
         _validate_filetype(filetype)
-    re_spacer = '[ ]{2,}'
+    re_spacer = '[ ]+'
     cmd_kwargs = {'spacer': re_spacer,
                   'filespec': re.escape(name),
                   'sel_user': sel_user or '[^:]+',
@@ -469,11 +469,14 @@ def fcontext_get_policy(name, filetype=None, sel_type=None, sel_user=None, sel_l
     current_entry_text = __salt__['cmd.shell'](cmd, ignore_retcode=True)
     if current_entry_text == '':
         return None
-    ret = {}
-    current_entry_list = re.split(re_spacer, current_entry_text)
-    ret['filespec'] = current_entry_list[0]
-    ret['filetype'] = current_entry_list[1]
-    ret.update(_context_string_to_dict(current_entry_list[2]))
+
+    parts = re.match(r'^({filespec}) +([a-z ]+) (.*)$'.format(**{'filespec': re.escape(name)}), current_entry_text)
+    ret = {
+        'filespec': parts.group(1),
+        'filetype': parts.group(2),
+    }
+    ret.update(_context_string_to_dict(parts.group(3)))
+
     return ret
 
 
--- a/tests/unit/modules/test_selinux.py
+++ b/tests/unit/modules/test_selinux.py
@@ -1,0 +1,88 @@
+# -*- coding: utf-8 -*-
+
+# Import Python libs
+import os
+
+# Import Salt Testing Libs
+from tests.support.mixins import LoaderModuleMockMixin
+from tests.support.unit import TestCase, skipIf
+from tests.support.mock import (
+    MagicMock,
+    patch,
+    NO_MOCK,
+    NO_MOCK_REASON
+)
+
+# Import Salt libs
+import salt.modules.selinux as selinux
+
+@skipIf(NO_MOCK, NO_MOCK_REASON)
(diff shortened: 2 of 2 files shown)


gtmanfred (approved):

salt/modules/selinux.py
```diff
@@ -469,11 +469,17 @@ def fcontext_get_policy(name, filetype=None, sel_type=None, sel_user=None, sel_l
     current_entry_text = __salt__['cmd.shell'](cmd, ignore_retcode=True)
     if current_entry_text == '':
         return None
+
+    current_entry_list = current_entry_text.split(' ')
+    filespec = current_entry_list.pop(0)
+    context_string = current_entry_list.pop()
+    filetype = ' '.join(current_entry_list).strip()
```
gtmanfred: I think there is a better way to do this.

```
filespec, ftype1, ftype2, context_string  = re.split(re_spacer, current_entry_text)
ret = {
    'filespec': filespec,
    'filetype': ' '.join((ftype1, ftype2)),
}
ret.update(_context_string_to_dict(context_string))
```

salt/modules/selinux.py
```diff
@@ -456,7 +456,7 @@ def fcontext_get_policy(name, filetype=None, sel_type=None, sel_user=None, sel_l
     '''
     if filetype:
         _validate_filetype(filetype)
-    re_spacer = '[ ]{2,}'
+    re_spacer = '[ ]+'
```
gtmanfred: this should probably just be changed to `'\s+'`? to grab all possible whitespace instead of just spaces?

gtmanfred (changes requested):

salt/modules/selinux.py
oarmstrong: Unfortunately this won't work, as the filetype _may_ contain a space, but doesn't always.

salt/modules/selinux.py
oarmstrong: Potentially. The output is always just space characters though and I'd prefer the parsing we do to reflect the format as specifically as possible.

salt/modules/selinux.py
oarmstrong: An alternative though (untested pseudocode):

```
parts = re.match('({filespec}) +([a-z ]+) +(.*)'.format({'filespec': re.escape(name)}))
ret = {
    'filespec': parts.group(1),
    'filetype': parts.group(2),
}
ret.update(_context_string_to_dict(parts.group(3)))
```

So parsing using a regular expression groups.

gtmanfred: This looks better to me :+1:

oarmstrong:
@gtmanfred changes made.

gtmanfred (approved):

oarmstrong:
Are we good now or did you want to talk about that other proposed change?

Also, I'm stupid and can't work out how to use Jenkins, could you point me to why the CI is failing?

rallytime:
@oarmstrong The Jenkins CI is having some issues we're working on lately. I don't see anything in here that is a new failure based on your changes.

However, I'd love to see some tests written for this change. Can you add some?

oarmstrong:
I was trying to get away with that :wink:.  Haven't looked at the test suite yet but I'll certainly give it a go!

rallytime:
Thanks @oarmstrong - let us know if you have any questions. :)

oarmstrong:
@rallytime I've added test cases. First time writing tests for Salt so let me know if it isn't quite up to scratch.

rallytime:
@oarmstrong Thank you! At first glance, they look good. I'll take a closer look once the tests have finished. Offhand, it looks like you've got a lint error. ;)

oarmstrong:
Sorry if you don't appreciate the ping, but I'm passing lints now @rallytime :).

rallytime:
No problem at all @oarmstrong :) I've requested a review from myself here to remind me to come back to this on Monday. Thanks for updating this!

rallytime (approved):
Thanks @oarmstrong!

oarmstrong:
:tada: thanks!

PR #45800, opened 2018-01-31
RAET transport Py3 support and shutdown fixes

DmitryKuzmenko:
### What does this PR do?
Fixes RAET transport:
* Fixed specific tests for raet
* Python 3 support
* No tracebacks on shutdown Py2/Py3
* Tested Py2-Py3 network level compatibility

This is the final part of RAET deprecation work.

### Tests written?
No

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/daemons/test/test_saltkeep.py
+++ b/salt/daemons/test/test_saltkeep.py
@@ -13,10 +13,11 @@ else:
 # pylint: enable=blacklisted-import
 
 import os
-import stat
-import time
-import tempfile
 import shutil
+import socket
+import stat
+import tempfile
+import time
 
 from ioflo.aid.odicting import odict
 from ioflo.aid.timing import StoreTimer
@@ -29,6 +30,7 @@ from raet.road import estating, stacking
 
 from salt.daemons import salting
 import salt.utils.kinds as kinds
+import salt.utils.stringutils
 
 
 def setUpModule():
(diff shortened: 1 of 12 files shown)


rallytime:
We have cut the `oxygen.rc1` branch so we can start hardening and testing. I will move this PR to the new branch once the test run that is already going is complete.

rallytime:
@DmitryKuzmenko Looks like there's some lint failures here. :)

DmitryKuzmenko:
Arrrggghhh!
@rallytime I'll fix them and rebase this to the RC branch. Thank you.

garethgreenaway (approved):

salt/daemons/test/test_saltkeep.py
```diff
@@ -232,20 +234,21 @@ def testBasic(self):
...
+                              'iha': None,
+                              'natted': None,
+                              'fqdn': socket.getfqdn('127.0.0.1'),
+                              'dyned': None,
+                              'sid': 0,
+                              'puid': 1,
+                              'aha': ['0.0.0.0', 7530],
+                              'role': mainData['role'],
+                              'sighex': salt.utils.stringutils.to_str(mainData['sighex']),
+                              'prihex': salt.utils.stringutils.to_str(mainData['prihex']),
```
terminalmage: Can you explain the use of `to_str` here (and elsewhere in this PR)? The `to_str` helper produces a `str` type on both Python 2 and 3. One common type of Python 3 incompatibility we've been cleaning up is when `to_str` is used instead of `to_bytes`. Since `str` and `bytes` are one and the same in Python 2, `to_str` will produce a `bytes` type on Python 2, but not on Python 3.

If what we need here is a `bytes` type, we need to be using the `to_bytes` helper instead.

terminalmage (commented):
Just have one question here @DmitryKuzmenko

salt/daemons/test/test_saltkeep.py
DmitryKuzmenko: It's because with raet transport Salt uses Ioflo and Raet under the hood and these packages take no care about `unicode` and use `str` type by default. Here I've tried to work it around in salt to not expand all the great unicode work to raet.

DmitryKuzmenko: If exactly about these lines: raet generates these keys as a binary hex blobs, but after it packed and unpacked for network all the data becomes strings of `str` type that is binary in Py2 and unicode in Py3.

terminalmage: OK, thanks for the explanation, this is good to merge then. I just wanted to make sure this value wasn't intended to be a bytes type, because this would break it on Python 3 if that were the case.

terminalmage (approved):

salt/daemons/test/test_saltkeep.py
DmitryKuzmenko: Thank you for pointing on this.

cachedout (approved):

Issue #40846, opened 2017-04-24
file.managed fails for long URLs


ghost:
### Description of Issue/Question

`file.managed` fails when using long URLs, such as signed URLs with Google Cloud:

    https://storage.googleapis.com/example-bucket/cat.jpeg?GoogleAccessId=example@example-project.iam.gserviceaccount.com&Expires=1458238630&Signature=VVUgfqviDCov%2B%2BKnmVOkwBR2olSbId51kSibuQeiH8ucGFyOfAVbH5J%2B5V0gDYIioO2dDGH9Fsj6YdwxWv65HE71VEOEsVPuS8CVb%2BVeeIzmEe8z7X7o1d%2BcWbPEo4exILQbj3ROM3T2OrkNBU9sbHq0mLbDMhiiQZ3xCaiCQdsrMEdYVvAFggPuPq%2FEQyQZmyJK3ty%2Bmr7kAFW16I9pD11jfBSD1XXjKTJzgd%2FMGSde4Va4J1RtHoX7r5i7YR7Mvf%2Fb17zlAuGlzVUf%2FzmhLPqtfKinVrcqdlmamMcmLoW8eLG%2B1yYW%2F7tlS2hvqSfCW8eMUUjiHiSWgZLEVIG4Lw%3D%3D

with this state:

    download_cat_jpeg:
    file.managed:
      - name: /tmp/cat.jpeg
      - source: https://storage.googleapis.com/example-bucket/cat.jpeg?GoogleAccessId=example@example-project.iam.gserviceaccount.com&Expires=1458238630&Signature=VVUgfqviDCov%2B%2BKnmVOkwBR2olSbId51kSibuQeiH8ucGFyOfAVbH5J%2B5V0gDYIioO2dDGH9Fsj6YdwxWv65HE71VEOEsVPuS8CVb%2BVeeIzmEe8z7X7o1d%2BcWbPEo4exILQbj3ROM3T2OrkNBU9sbHq0mLbDMhiiQZ3xCaiCQdsrMEdYVvAFggPuPq%2FEQyQZmyJK3ty%2Bmr7kAFW16I9pD11jfBSD1XXjKTJzgd%2FMGSde4Va4J1RtHoX7r5i7YR7Mvf%2Fb17zlAuGlzVUf%2FzmhLPqtfKinVrcqdlmamMcmLoW8eLG%2B1yYW%2F7tlS2hvqSfCW8eMUUjiHiSWgZLEVIG4Lw%3D%3D
[...]

PR #45808, opened 2018-01-31
[bugfix] Fix for oversized filenames in fileclient

b1naryth1ef:
### What does this PR do?

This PR fixes the bug reported in #40846, wherein salt will attempt (and fail) to write a file with a long filename (e.g. a URL we download that has long query params). This is caused by filesystem naming limitations, which are generally at or above 255 characters.

#40846

### Previous Behavior

Salt would crash when attempting to download and cache a file with a large URL or large set of query paramaters.

### New Behavior

Salt will now use a SHA256 hash of the file name if the file name meets or exceeds 255 characters.

### Tests written?

Yes

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/fileclient.py
+++ b/salt/fileclient.py
@@ -46,6 +46,7 @@ from salt.ext.six.moves.urllib.parse import urlparse, urlunparse
 # pylint: enable=no-name-in-module,import-error
 
 log = logging.getLogger(__name__)
+MAX_FILENAME_LENGTH = 255
 
 
 def get_file_client(opts, pillar=False):
@@ -830,6 +831,9 @@ class Client(object):
             file_name = '-'.join([url_data.path, url_data.query])
         else:
             file_name = url_data.path
+
+        if len(file_name) > MAX_FILENAME_LENGTH:
+            file_name = salt.utils.hashutils.sha256_digest(file_name)
 
         return salt.utils.path.join(
             cachedir,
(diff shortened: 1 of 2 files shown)


b1naryth1ef:
There is likely a pending discussion here about whether salt should do this implicitly (as implemented in this PR) or whether this should be a flag which changes the cache directory. I avoided the latter since this seemed relatively sane and I didn't wanna mess around with passing parameters all over the place, although I'm open to swapping it over.

terminalmage (changes requested):
This needs more work, because it would break things like `is_cached` and `_extrn_path` in the fileclient which would expect a full filename.

b1naryth1ef:
@terminalmage can you expand on what you mean by that a little bit? Based on my naive understanding and reading of the file, it looks like is_cached _uses_ `_extrn_path` (which is where my patch is applied).

terminalmage (approved):
You are right, I misread the function being modified when I was reviewing. Sorry!

gtmanfred:
re-run all

b1naryth1ef:
No problem! Rebased off the latest develop as was requested in slack.

terminalmage (approved):

terminalmage:
@rallytime I took care of the backport in https://github.com/saltstack/salt/pull/47673, so you don't need to. The call to `path_join` in the backport will however need to be fixed when merged into 2018.3, but this is something you already check for so we should be good.

PR #45820, opened 2018-02-01
Make vault session token reusable

iustina-m:
### What does this PR do?
Allow session token reuse in vault requests

n\a

### Previous Behavior
Any call to make_requests implies re-authentication. Session token is not being reused

### New Behavior
One can now reuse the session token by first getting the url+token returned by the runner and then calling make_request with token and url:
```
result, token, vault_url =__utils__['vault.make_request']('GET', url, get_token_url)
__utils__['vault.make_request']('GET', url, token, vault_url)
```
### Tests written?

No

### Commits signed with GPG?

No

### DIFF
--- a/salt/utils/vault.py
+++ b/salt/utils/vault.py
@@ -24,6 +24,8 @@ logging.getLogger("requests").setLevel(logging.WARNING)
 
 # Load the __salt__ dunder if not already loaded (when called from utils-module)
 __salt__ = None
+
+
 def __virtual__():  # pylint: disable=expected-2-blank-lines-found-0
     try:
         global __salt__  # pylint: disable=global-statement
@@ -50,27 +52,27 @@ def _get_token_and_url_from_master():
         log.debug('Running on minion, signing token request with key %s',
                   private_key)
         signature = base64.b64encode(salt.crypt.sign_message(
-                                                             private_key,
-                                                             minion_id
-                                                            ))
+            private_key,
+            minion_id
+        ))
         result = __salt__['publish.runner'](
-                                            'vault.generate_token',
-                                            arg=[minion_id, signature]
-                                           )
+            'vault.generate_token',
+            arg=[minion_id, signature]
+        )
     else:
         private_key = '{0}/master.pem'.format(pki_dir)
         log.debug('Running on master, signing token request for %s with key %s',
                   minion_id, private_key)
         signature = base64.b64encode(salt.crypt.sign_message(
-                                                             private_key,
-                                                             minion_id
-                                                            ))
+            private_key,
+            minion_id
+        ))
         result = __salt__['saltutil.runner'](
-                                             'vault.generate_token',
-                                             minion_id=minion_id,
-                                             signature=signature,
-                                             impersonated_by_master=True
-                                            )
+            'vault.generate_token',
+            minion_id=minion_id,
+            signature=signature,
+            impersonated_by_master=True
+        )
 
     if not result:
         log.error('Failed to get token from master! No result returned - '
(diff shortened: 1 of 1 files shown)


salt/utils/vault.py
```diff
@@ -126,24 +126,28 @@ def _get_vault_connection():
         return _get_token_and_url_from_master()
 
 
-def make_request(method, resource, profile=None, **args):
+def make_request(method, resource, token=None, vault_url=None, get_token_url=False, profile=None, **args):
```
rallytime: Can you move these new kwargs to the end fo the list? We don't want to cause an API change and cause regressions for calls to this function.

salt/utils/vault.py
```diff
@@ -126,24 +126,28 @@ def _get_vault_connection():
...
     '''
     if profile is not None and profile.keys().remove('driver') is not None:
         # Deprecated code path
         return make_request_with_profile(method, resource, profile, **args)
 
-    connection = _get_vault_connection()
-    token, vault_url = connection['token'], connection['url']
-    if 'verify' not in args:
-        args['verify'] = connection['verify']
+    if not token or not vault_url:
```
rallytime: What is the expected behavior here if one is passed in, but not the other? If you pass in a token, but not a vault_url, the passed in token will be overridden, which could be confusing. Is that expected?

salt/utils/vault.py
```diff
@@ -126,24 +126,28 @@ def _get_vault_connection():
...
+        if 'verify' not in args:
+            args['verify'] = connection['verify']
 
     url = "{0}/{1}".format(vault_url, resource)
     headers = {'X-Vault-Token': token, 'Content-Type': 'application/json'}
     response = requests.request(method, url, headers=headers, **args)
 
-    return response
+    if get_token_url:
+        return response, token, vault_url
```
rallytime: Do we need to handle this new return structure elsewhere in the code base?

rallytime (commented):
Thanks @iustinam - I have some comments/questions on this.

salt/utils/vault.py
iustinam: @rallytime The expected behavior from the code is to re-auth but I don't like that much either, if for example the url is not given.
It would have been better to have the vault url by other means, but I'm not sure how. The utils is stateless.
_get_token_and_url_from_master()  returns the token along with the url (which it gets from the master via the peer run). Would it be better to have a function in the vault runner that just returns the vault url?

salt/utils/vault.py
iustinam: No, because it's not used anywhere. 
If this gets merged it would make sense to set get_token_url to True to get the token and then reuse it but not needed.

salt/utils/vault.py
rallytime: Hrm. What do you think @gtmanfred? I think you've been involved in vault functionality lately.

gtmanfred: I am ok with this.  One of the things that sdb/vault should be able to do is reduce load on the master, and allow the minion to just look up secrets on its own.  And it would need to keep track of its token in order to do that.

But I would like to reserve the right to rip it out if it turns out to break vault + sdb in Fluorine.

I have made this issue to add tests before the Fluorine release https://github.com/saltstack/salt/issues/46052

rallytime: Wonderful! Thanks @gtmanfred :)

rallytime: That sound OK to you @iustinam, before we merge this in?

rallytime (approved):

PR #45824, opened 2018-02-01
decode file list for setup.py on python2

gtmanfred:
### What does this PR do?
Decode file list for setup.py

Unable to build `python setup.py sdist` with python2

### Tests written?

No

### Commits signed with GPG?

Yes

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -522,7 +522,8 @@ class Sdist(sdist):
             self.run_command('write_salt_ssh_packaging_file')
             self.filelist.files.append(os.path.basename(PACKAGED_FOR_SALT_SSH_FILE))
 
-        sdist.make_release_tree(self, base_dir, files)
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -522,7 +522,11 @@ def make_release_tree(self, base_dir, files):
             self.run_command('write_salt_ssh_packaging_file')
             self.filelist.files.append(os.path.basename(PACKAGED_FOR_SALT_SSH_FILE))
 
-        sdist.make_release_tree(self, base_dir, files)
+        def decode_files(files):
+            PY3 = sys.version_info[0] == 3
```
s0undt3ch: Better reverse the logic, if py2, because when py4 comes out this will break. Makes sense?

gtmanfred: that is a solid point

gtmanfred: actually, it turnes out that IS_PY3 is already a thing in setup.py, so I am going to use that instead.

It just checks if the version is > (3,) so it should still work with py4

s0undt3ch (approved):
:+1:

PR #45831, opened 2018-02-01
modules.postgres: cleanup handling of password kwargs

blueyed:
Remove it from _psql_cmd and _connection_defaults, where it is not used,
but only in _run_psql.


garethgreenaway (approved):

salt/modules/postgres.py
```diff
@@ -316,31 +315,29 @@ def _connection_defaults(user=None, host=None, port=None, maintenance_db=None,
...
 
 
 def _psql_cmd(*args, **kwargs):
     '''
     Return string with fully composed psql command.
 
-    Accept optional keyword arguments: user, host and port as well as any
-    number or positional arguments to be added to the end of command.
+    Accepts optional keyword arguments: user, host, port and maintenance_db,
+    as well as any number or positional arguments to be added to the end of
```
rallytime: `number or positional` --> `number of positional`

salt/modules/postgres.py
```diff
@@ -316,31 +315,29 @@ def _connection_defaults(user=None, host=None, port=None, maintenance_db=None,
...
+    the command.
     '''
     (user, host, port, maintenance_db, password) = _connection_defaults(
```
rallytime: If `_connection_defaults` doesn't return a password anymore, then it needs to be removed here as well.

rallytime (changes requested):

salt/modules/postgres.py
rallytime: @blueyed Were you able to come back to this PR?

blueyed: Apparently not.

Feel free to push a fixup here (maybe even through the GitHub UI), but since I am commenting already, I'll do it now.

blueyed:
Added fixup and rebased.

rallytime (approved):

PR #45834, opened 2018-02-02
2 unicode-compatibility fixes for cmdmod.py

terminalmage:
First, `shlex.split()` will raise an exception when passed a unicode
type with unicode characters in the string. This modifies our
`shlex.split()` helper to first convert the passed string to a `str`
type, and then return a decoded copy of the result of the split.

Second, this uses our `to_unicode` helper to more gracefully decode the
stdout and stderr from the command. Unit tests have been added to
confirm that the output is properly decoded, including instances where
decoding fails because the return from the command contains binary data.

`.format()` usage has also been removed from the logging in cmdmod.py.


terminalmage:
@thatch45 Should we be using the `'replace'` behavior to deal with decode errors in cmdmod.py? This means that characters which cannot be decoded will be replaced with U+FFFD. The other option would be to use `'ignore'` which would simply discard any characters which cannot be decoded.

Which would you prefer? Replacing or ignoring?

terminalmage:
Another option would be to make this toggleable via an argument to all the `cmd.run*` functions. This is probably the best way to go, though we'd still need to decide on which behavior to make the default. I'm not married to replace or ignore, and since the code that was using `'replace'` already is new to the oxygen branch, we don't have to worry about backward-compatibility.

damon-atkins:
cmd etc should have an decode option defaulting to __salt_system_encoding__
replace is better the ignore.   Because if the input return is a `/filename<japanlang>`  then the file name will be changed to `/filename`  and then it this is used for a delete the wrong file could be removed.

rallytime:
@terminalmage The test failure here is related: https://jenkins.saltstack.com/job/PR/job/salt-pr-rs-cent7-n/16011/

terminalmage:
@rallytime The test failure is fixed now

@damon-atkins I don't really see the need for such an option, but I've added it anyway.

I also made a number of changes to the docstrings in cmdmod.py. There was a lot of bad formatting and inaccurate information (e.g. SLS examples in the execution module) that was in need of a cleanup.

salt/modules/cmdmod.py
```diff
@@ -725,13 +748,13 @@ def _run_all_quiet(cmd,
                    saltenv='base',
                    pillarenv=None,
                    pillar_override=None,
-                   output_loglevel=None):
+                   output_encoding=None):
```
terminalmage: NOTE: This is intentional, not a mistake. We don't do anything with output_loglevel, since we're just passing it as `'quiet'` when we invoke `_run()`.

rallytime:
@terminalmage There's a lint error here: https://jenkins.saltstack.com/job/PR/job/salt-pr-lint-n/18991/violations/file/salt/modules/dracr.py/

rallytime (approved):
This looks good to me, pending the lint fix.

terminalmage:
@rallytime should be fixed now

salt/modules/cmdmod.py
```diff
@@ -569,21 +569,43 @@ def _get_stripped(cmd):
...
+                encoding=output_encoding)
+        except TypeError:
+            # stdout is None
             out = ''
         except UnicodeDecodeError:
-            log.error('UnicodeDecodeError while decoding output of cmd {0}'.format(cmd))
-            out = proc.stdout.decode(__salt_system_encoding__, 'replace')
+            out = salt.utils.stringutils.to_unicode(
+                proc.stdout,
+                encoding=output_encoding,
```
cachedout: Will we ever hit this with output_encoding set to None?

salt/modules/cmdmod.py
```diff
@@ -782,124 +807,115 @@ def run(cmd,
...
-                  - BATCH: 'yes'
-
-        .. warning::
-
-            The above illustrates a common PyYAML pitfall, that **yes**,
-            **no**, **on**, **off**, **true**, and **false** are all loaded as
-            boolean ``True`` and ``False`` values, and must be enclosed in
-            quotes to be used as strings. More info on this (and other) PyYAML
-            idiosyncrasies can be found :ref:`here <yaml-idiosyncrasies>`.
+        deliver it's results
```
cachedout: `its`

salt/modules/cmdmod.py
terminalmage: Yes, `output_encoding` defaults to None so it will always be None unless explicitly passed.

salt/modules/cmdmod.py
terminalmage: Fixed.

cachedout (approved):

PR #45835, opened 2018-02-02
Adds a missing return statement.

kstreee:
### What does this PR do?

I'm not sure that this is a bug or spec of the function, but based on the interface of the function and other implementations of the function, I think a return statement should be added.

N/A

### Previous Behavior

Returns `None` always.

### New Behavior

Returns something.

### Tests written?

No

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/fileclient.py
+++ b/salt/fileclient.py
@@ -1241,7 +1241,7 @@ class RemoteClient(Client):
         load = {'saltenv': saltenv,
                 'prefix': prefix,
                 'cmd': '_file_list_emptydirs'}
-        self.channel.send(load)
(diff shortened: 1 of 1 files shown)


terminalmage (approved):
Yes, this is correct, I fixed this in the Oxygen branch the other day, and I forgot to look at earlier branches to see where else this may have been broken. Thanks for the catch!

terminalmage:
@rallytime this should be backported if possible, unless we've cut 2017.7.3 already.

rallytime:
@terminalmage We've already cut 2017.7.3.

PR #45844, opened 2018-02-02
[oxygen] Track the next scheduled fire time and when jobs are skipped.

garethgreenaway:
### What does this PR do?
Track the next scheduled fire time and when jobs are skipped.

#45865

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/utils/schedule.py
+++ b/salt/utils/schedule.py
@@ -806,7 +806,7 @@ class Schedule(object):
 
             run = False
 
-            if job in _hidden and not data:
+            if job in _hidden:
                 continue
 
             if not isinstance(data, dict):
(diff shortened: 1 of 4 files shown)


cachedout (approved):

cachedout:
Tests?

garethgreenaway:
@cachedout I have some tests and some other changes to push to this PR.

cachedout:
@garethgreenaway Would it be a good idea to add a unit test specifically for the `_check_max_running` function?

alexpeay:
I am assuming this is the Max Parallel portion of the code, right? If there is a way to differentiate when a job is skipped due to a user request to skip vs a skip due to already running at the Max Parallel level (effectively a system Skip) that would be great.

garethgreenaway:
@alexpeay There are a few different skip reasons that this change introduces including the maxrunning reason, the others include when the job is in a time range specified by `skip_during_range` and when a job is set to skip at a specific time using `skip_explicit`.

Issue #38625, opened 2017-01-06
Add flag to cmd.run to allow overriding Salt's retcode field


whiteinge:
### Description of Issue/Question

It is a pretty well established pattern in the UNIX world that failures set non-zero exit codes -- but there are outliers (I'm looking at you, `diff`.) We are working to formalize Salt's `retcode` field in #18510 and right now we're setting the return code in Salt from the return code in the shell. These values may or may not line up.

@gtmanfred suggested following systemd's approach by adding an argument that accepts a list of exit codes that will also be treated as successes. E.g.:

```
salt cmd.run_all cmd='return 42' success_exit_codes='[42]'
```

If the returned code from the command matches a return in the argument then _Salt's_ `retcode` field will then correctly be set to `0`.

whiteinge:
The `success` and `retcode` fields were intended to communicate separate things. See the discussion in https://github.com/saltstack/salt/issues/38788#issuecomment-276495152.

whiteinge:
Reopening. We still need to the `success_exit_codes` flag for brain-dead utils (like `diff`) that return non-zero codes for successful operations.

whiteinge:
Wait, no. 😵  Brain-dead utils will return what they return and that will be reflected in Salt's `retcode` field and consumers of that util will have to know that util does that.

whiteinge:
Sigh, yes. Deleting a few noise comments and reopening.

rickh563:
ZD-1971

garethgreenaway:
The fix for this has been merged and will be in the Fluorine release.  Going to close this one out.  If the problem persists, please feel free to reopen the issue.  Thanks!

PR #45851, opened 2018-02-03
[develop] Ability to override retcode from various cmdmod functions

garethgreenaway:
### What does this PR do?
Adding a flag to various `cmdmod` functions to allow overriding the `retcode` that is returned.  This is useful when a command will return a non-zero `retcode` but the command is still considered successful.

#38625

### Tests written?
Yes

### Commits signed with GPG?
Yes


rallytime (approved):

salt/modules/cmdmod.py
```diff
@@ -951,6 +971,13 @@ def run(cmd,
         Be absolutely certain that you have sanitized your input prior to using
         python_shell=True
 
+    :param list success_exit_codes: This parameter will be allow a list of
+        non-zero return codes that should be considered a success.  If the
+        return code  returned from the run matches any in the provided list,
```
terminalmage: extra space between `code` and `returned`.

salt/modules/cmdmod.py
```diff
@@ -951,6 +971,13 @@ def run(cmd,
...
+        the return code will be overridden with zero.
```
terminalmage: "overridden" seems like the wrong word here. "replaced" maybe?

salt/modules/cmdmod.py
```diff
@@ -625,7 +626,15 @@ def _get_stripped(cmd):
...
-        ret['retcode'] = proc.process.returncode
+        _returncode = proc.process.returncode
+        ret['retcode'] = _returncode
+        if success_exit_codes:
+            if not isinstance(success_exit_codes, list):
+                success_exit_codes = list(success_exit_codes)
+            # Ensure all our retcodes are ints
+            success_exit_codes = [int(i) for i in success_exit_codes]
+            if _returncode in success_exit_codes:
+                ret['retcode'] = 0
```
terminalmage: In many places in salt we support both a Python and comma-separated list. We have a helper function for this, `salt.utils.args.split_input()`, which would be a good fit here.

Also , I noticed that there is a similar block below. How about moving the evaluation of `success_exit_codes` up higher in the function and doing something like this:

```python
if success_exit_codes is None:
    success_exit_codes = [0]
else:
    try:
        success_exit_codes = [int(i) for i in salt.utils.args.split_input(success_exit_codes)]
    except ValueError:
        raise SaltInvocationError(
            'success_exit_codes must be a list of integers'
        )
```

Then you could just do a simple membership check:

```python
if ret['retcode'] in success_exit_codes:
```

terminalmage: I'm also wondering if maybe this should be called `success_retcodes` instead of `success_exit_codes`. We already use `retcode` everywhere else in cmdmod.py to refer to the exit code of a command, so it may make sense to rename this for clarity/consistency.

salt/modules/cmdmod.py
```diff
@@ -951,6 +971,13 @@ def run(cmd,
         Be absolutely certain that you have sanitized your input prior to using
         python_shell=True
 
+    :param list success_exit_codes: This parameter will be allow a list of
+        non-zero return codes that should be considered a success.  If the
+        return code  returned from the run matches any in the provided list,
+        the return code will be overridden with zero.
+
+      .. versionadded:: Fluorine
+
```
terminalmage: NOTE: The above two comments should be addressed in all places in the docstrings where they were copied.

terminalmage (changes requested):

terminalmage (approved):

PR #45859, opened 2018-02-05
Support searching by 'groupattribute' when using freeipa client

Tanoti:
### What does this PR do?
Adds the attribute 'groupattribute' to the search path when using freeipa client to support configurations where authentication is by group.

N/A

### Previous Behavior
Only the 'accountattributename' attribute was searched.

### New Behavior
First 'accountattributename' is searched then 'groupattribute' to provide extended search functionality similar to that in the "else" section of the code block

### Tests written?
No

### Commits signed with GPG?
No

### DIFF
--- a/salt/auth/ldap.py
+++ b/salt/auth/ldap.py
@@ -362,10 +362,10 @@ def groups(username, **kwargs):
             search_results = bind.search_s(search_base,
                                            ldap.SCOPE_SUBTREE,
                                            search_string,
-                                           [_config('accountattributename'), 'cn'])
+                                           [_config('accountattributename'), 'cn', _config('groupattribute'), 'cn'])
 
             for entry, result in search_results:
-                for user in result[_config('accountattributename')]:
+                for user in result[_config('accountattributename'), _config('groupattribute')]:
                     if username == user.split(',')[0].split('=')[-1]:
                         group_list.append(entry.split(',')[0].split('=')[-1])
 


DmitryKuzmenko (approved):

gtmanfred (approved):

DmitryKuzmenko:
It looks I took less attention than it was needed on this.
@Tanoti could you please explain me how it works with freeipa? I have an openldap server only and can't test on freeipa.
With openldap `search_s` query's last argument means 'list of fields to be retrieved. And it returns the result in the format: `[(object_path, {values of requested keys}), (...), ...]`
So your commit adds a new value to the list of values to be retrieved. And the second part of this PR makes `for` loop will not work at all because it tries to retrieve a value from the dict by tuple of `accountattribute` and `groupattribute` when the dict keys are strings only.
Do freeipa treats the last argument in some another way? Do it returns results in another format?

Tanoti:
@DmitryKuzmenko I am not sure how this actually works. I just found that adding "_config('groupattribute'), 'cn'" to the array as additional elements meant that freeipa successfully found the record user via the group. I am sorry if this broke openldap, maybe we need a separate logic section for each?

DmitryKuzmenko:
@Tanoti no, you didn't break openldap, you broke just one test. Your changes touched the only freeipa part, I'm just don't understand how that changes could work.
If you have a machine working with freeipa where you can run a test it would be very helpful if you log the `search_results` with and without your fix and put it here to make me see the data structure.
I'll add an example as an inline comment.

salt/auth/ldap.py
```diff
@@ -362,10 +362,10 @@ def groups(username, **kwargs):
             search_results = bind.search_s(search_base,
                                            ldap.SCOPE_SUBTREE,
                                            search_string,
-                                           [_config('accountattributename'), 'cn'])
+                                           [_config('accountattributename'), 'cn', _config('groupattribute'), 'cn'])
```
DmitryKuzmenko: log.critical('SEARCH_RESULTS: %s', search_results)

Tanoti: @DmitryKuzmenko I'll set up a test server to talk to our freeipa server and get back to you soon,

Tanoti:
@DmitryKuzmenko A sanitised output can be seen below (I've just changed the usernames and company and reduced it to a handful of records).

```
[
  ('cn=admins,cn=groups,cn=accounts,dc=company,dc=com',
  {
    'member': [
      'uid=jhicks,cn=users,cn=accounts,dc=company,dc=com',
    ],
    'cn': [
      'admins'
    ]
  }),
  ('cn=ipausers,cn=groups,cn=accounts,dc=company,dc=com',
  {
    'member': [
      'uid=bbob,cn=users,cn=accounts,dc=company,dc=com'
      'uid=fflintstone,cn=users,cn=accounts,dc=company,dc=com'
    ],
    'cn': [
      'ipausers'
    ]
  }),
  ('cn=opusers,cn=groups,cn=accounts,dc=company,dc=com',
  {
    'member': [
      'uid=cdracula,cn=users,cn=accounts,dc=company,dc=com',
    ],
    'cn': [
      'opusers'
    ]
  }),
]
```

I think from looking at my code that what we need is the following:
```
# top of file
import itertools
...
# line 357
            search_results = bind.search_s(search_base,
                                           ldap.SCOPE_SUBTREE,
                                           search_string,
                                           [_config('accountattributename'), 'cn', _config('groupattribute'), 'cn'])
...
# line 360
                for user in itertools.chain(result.get(_config('accountattributename'), [])
                        , result.get(_config('groupattribute'), [])):
```
This gets the right results for a group search and allows the for loop to cope with either the user or group lists being empty.

DmitryKuzmenko:
Yes this looks more correct with just one correction for line 357: the list of strings you're passing says: "Put these attributes into the dictionary". So there's no need to duplicate the 'cn':
```python
            search_results = bind.search_s(search_base,
                                           ldap.SCOPE_SUBTREE,
                                           search_string,
                                           [_config('accountattributename'), 'cn', _config('groupattribute'), 'cn'])
```

Tanoti:
@DmitryKuzmenko I've pulled down the latest changes to develop and looked at my PR in relation to the current ldap.py and I'm not 100% sure how to handle the salt.utils.stringutils.to_str() calls? Should I just give it a go and let someone tell me what I've got wrong?

DmitryKuzmenko:
Here's the full text of our Unicode manifest. https://gist.github.com/terminalmage/7440c8d6e46ed4ee2f69c834d5363512

In two words: all string constants are now unicode by default, all strings salt uses internally have to be unicode too. If some external library or an IO operation needs non-unicode value it have to be translated.
In this case it looks ldap API expects a list of `str` objects as a last argument. `str` is unicode in py3 and a binary string in py2.
To translate a constant to `str` we use `str('constant')` appending the touched line with `  # future lint: disable=blacklisted-function` to say pylint it's ok to use `str()` right here.
To translate a non-constant to `str` we use `salt.utils.stringutils.to_str()`.

So all you need is to use this approach in your patch.

Tanoti:
@DmitryKuzmenko I've pushed an update to my branch in github but it has not updated this PR. Do I need to do something like raise a new PR?

gtmanfred:
This pr has already been merged, you will need to make a new pr.

Tanoti:
New PR opened as #47428

Issue #35070, opened 2016-07-29
postgres_user.present makes excessive SQL queries (regression)


notpeter:
### Description of Issue/Question

The changes made to modules.postgres.user_list() in saltstack/salt#20980 cause postgres_user.present states to issue many times more SQL queries (each in distinct psql subprocess).   This number of queries is directly proportional to the number roles present in your postgresql server. So in our case, our 30-odd server-side roles result in 30+ queries for a single postgres_user.present state.
### Steps to Reproduce Issue

On a box with a local copy of postgres create a handful of roles:

```
psql postgres -c "CREATE ROLE poop1; CREATE ROLE poop2; CREATE ROLE poop3;"
```

Then run a simple postgres_user.present state.

```
super_pete:
  postgres_user.present:
    - password: yumfudge
    - encrypted: True
    - user: postgres
    - superuser: True
```

Run salt-call with --log-level=INFO and notice there's a queries issued for each poop1-poop3 (and any other roles you might happen to have present). 

Reproduced with 2016.3.1.

Ch3LL:
ping @kiorky could I get your input on this one. Maybe there is a different way to approach this so we are not making as many calls to postgres?  What are your thoughts. Thanks

dglloyd:
I just encountered this issue as well, just on a larger scale (adding ~250 users). Process took over 30 minutes in total, as it executes multiple queries per user existing in the database, every create_user call, so time per create grows geometrically.  After some discussion in ##salt on freenode, I was asked to open a bug report.

```
09:24 < babilen> role_get should be a single query
09:26 < babilen> Holy shit ... That needs urgent refactoring
09:27 < babilen> I mean, that should be two queries per "add user" not all that madness
09:27 < babilen> And you should even be able to short circuit it when the user doesn't exist (you don't need the inherited roles and whatnot for a role that doesn't exist in the 
                 first place)
09:29 < babilen> Please file a bug, recommend a "user_exists" function in the module that runs for a specific user (SELECT 1 FROM pg_roles WHERE rolname=$USERNAME) and only if it 
                 exists starts getting info for that specific user and not *ALL OF THEM*
09:29 < babilen> Sorry, can't be of more help short of implementing that myself
09:29 < dlloyd> will do
09:31 < babilen> https://github.com/ansible/ansible/blob/devel/lib/ansible/modules/database/postgresql/postgresql_user.py#L245-L251 -- Ansible does this
[...]

Issue #20980, opened 2015-02-24
postgresql: user roles are never updated


kiorky:

notpeter:
@kiorky This change made it so that every postgres_user.present (even no change) can result in dozens of SQL queries. Your code queries for the list of roles, iterates through that list in python issuing one query per role present on the server.

So instead of querying like this inside the for loop (1 query for each server role)
````
select rolname from pg_user
join pg_auth_members on (pg_user.usesysid=pg_auth_members.member)
join pg_roles on (pg_roles.oid=pg_auth_members.roleid)
where pg_user.usename=\'{0}\'
````
It could instead issue a single query outside the loop and get the same results for all users:
````
select rolname from pg_user
join pg_auth_members on (pg_user.usesysid=pg_auth_members.member)
join pg_roles on (pg_roles.oid=pg_auth_members.roleid)
where pg_user.usename IN ['user1','user2','etc']
````

PR #45870, opened 2018-02-06
Remove quadratic behaviour from postgres.role_get

ferringb:
See #35070; in #20980, role_get was modified so that it constantly re-asked
the DB for what roles existed.  Caching should exist, but the intent isn't bad-
however the implementation was such that every single role required another
psql invocation.  If you're provisioning 100 users, this gets painfull quickly
since it's quadratic.

Instead, modify the "collect all users" query to pull role membership,
converting role_get back down to a single query.  The syntax used should be
compatible at least back to postgres 8.3 (array_ag usage specifically).

In terms of performance difference, with this change, a provisioning for
my local setup for 102 users takes under a minute- less than that, but my
test case has unrelated states.  Without this change, it takes well over 30
minutes (exact time, I don't know- I gave up waiting on it).


ferringb:
Unless I'm missing something, the various build tests you've got wired there all look fairly broke- and not by my change.

garethgreenaway:
Looks like there is some cross over with #45966 that might require a rebase depending on which is merged first.

garethgreenaway (approved):

rallytime:
Hi @ferringb - I know there are several test failures in the builds that are not related to your change, there are two test failures that are:
- unit.modules.test_postgres.PostgresTestCase.test_user_exists
- unit.modules.test_postgres.PostgresTestCase.test_user_list

Can you take a look at those, please? Thanks!

rallytime:
@ferringb Just a bump on the request above. :)

rallytime:
Since we haven't heard back here, I am going to close this PR. We will happily reopen this if traction picks up again on this change. Thanks!

PR #45874, opened 2018-02-06
fix for local client timeout bug

GwiYeong:
### What does this PR do?
This PR fixes a timeout bug that often occurs when calling api with local_client in salt-api
This bug occurs when the api process directly subscribes to asynchronous events and misses the event.
`all_return` function, which is called by `_disbatch_local` function, subscribes events  to create the result.
The problem is that this function is subscribing to events with a tag that only consists of jid.
When some minions return events while this function is processing an event that was raised before, 
that minions events are missed. And missed minions are still remain in `minions_remaining`. Finally timeout exception is occured

##### example situation
```
salt api : get_event(jid_tag)
minion 1 : return result
salt api : process minion 1 result data
minion 2 : return result (<- salt api misses this result)
minion 3 : return result (<- salt api misses this result)
salt api : get_event(jid_tag)
minion 4 : return result
salt api : process minion 4 result data
salt api : get_event(jid_tag) (<- waiting for minion2, minion3)
salt api : TimeoutException

```

None

### New Behavior

I update the function to subscribe events.
It subscribe to events for each minion, rather than subscribe to events for jid. [[code](https://github.com/GwiYeong/salt/blob/2017.7-local-client-hotfix/salt/netapi/rest_tornado/saltnado.py#L987)]
In order to this, It is necessary to subscribe to an event that exactly matches the tag. So I added it at `get_event`.
I also update the `job_not_running` function and the `all_returns` function to share on future to indicate that they ended each other,
This is to ensure that all generated futures are terminated gracefully to prevent memory leaks.

### Tests written?

No

### Commits signed with GPG?

Yes


salt/netapi/rest_tornado/saltnado.py
```diff
@@ -925,64 +947,83 @@ def _disbatch_local(self, chunk):
...
-                                           minions_remaining=minions_remaining,
-                                           )
 
-        raise tornado.gen.Return(chunk_ret)
+        all_return_future = self.all_returns(pub_data['jid'],
+                                             is_finished,
+                                             minions_remaining=list(minions_remaining),
+                                             )
+        yield job_not_running_future
+        raise tornado.gen.Return((yield all_return_future))
```
skizunov: Are the double parenthesis required here? Would this work:
```
raise tornado.gen.Return(yield all_return_future)
```

salt/netapi/rest_tornado/saltnado.py
```diff
@@ -925,64 +947,83 @@ def _disbatch_local(self, chunk):
...
+            except TimeoutException:
+                pass
+
+            # clear finished event future
             try:
-                minions_remaining.remove(event['data']['id'])
-            except ValueError:
+                minions_remaining.remove(minion_events[f])
+                del minion_events[f]
+            except ValueError as e:
```
skizunov: Since you don't make use of `e`, perhaps just have:
```
except ValueError:
```

skizunov (changes requested):

salt/netapi/rest_tornado/saltnado.py
GwiYeong: Inner parentheses is part of yield expressions. so it is required. 
you can see [doc](https://docs.python.org/3/reference/simple_stmts.html#the-yield-statement)

salt/netapi/rest_tornado/saltnado.py
GwiYeong: this has been fixed

skizunov (approved):

GwiYeong:
@skizunov @cachedout 
Hi everyone, is there any progress here?

cachedout (approved):
Nicely done.

PR #45878, opened 2018-02-06
ec2_pillar update to fix finding instance-id

damon-atkins:
### What does this PR do?
ec2_pillar update to fix finding instance-id. Current code did not work at all, their most of been
a change in salt some time ago.
Also add support to use any tag to compare minion id against to find the ec2 entry.
Also add the ability to return ec2 tags as a list instead of a string, split on a character of choice.

### Previous Behavior
Did not work, maybe the grain structure has been changed for ec2

### New Behavior
Works as documented as well as also add support to use any tag to compare minion id against.

### Tests written?

No Tests Added

### Commits signed with GPG?

No


damon-atkins:
@jbussdieker maybe you would like to test these changes.

garethgreenaway (approved):

damon-atkins:
This is ready to be merged.

gtmanfred (approved):

PR #45887, opened 2018-02-06
Remove special handling for PY2 escape function in zfs module

rallytime:
With the changes in oxygen to use `unicode_literals`, we shouldn't need
to encode these lines anymore specifically for PY2 in this way.

Fixes the zfs unit module test failures on develop:

https://github.com/saltstack/salt-jenkins/issues/824


terminalmage (approved):
Checked and confirmed earlier branches don't have this string_escape, this should be GTM.

PR #45890, opened 2018-02-06
Kubernetes: Support for modifying container image

mbhinder:
This commit adds support for setting the container image of a
kubernetes deployment.

### DIFF
--- a/salt/modules/kubernetes.py
+++ b/salt/modules/kubernetes.py
@@ -622,6 +622,58 @@ def show_namespace(name, **kwargs):
         _cleanup()
 
 
+def set_container_image(image, deployment,
+                         container,
+                         namespace="default",
+                         **kwargs):
+    '''
+    Set the image of a container in the kubernetes deployment defined
+    by the specified name.
+
+    CLI Examples::
+
+        salt '*' kubernetes.set_deployment_image nginx:latest nginx-deployment nginx-container namespace=prod
+
+    '''
+    _setup_conn(**kwargs)
+
+    try:
+        api_instance = kubernetes.client.ExtensionsV1beta1Api()
+        patch = {
+            "spec":{
+                "template":{
+                    "spec":{
+                        "containers":[
+                            {
+                                "name":container,
+                                "image":image
+                            }
+                        ]
+                    }
+                }
+            }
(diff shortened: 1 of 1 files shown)


rallytime:
@mbhinder There's a lint failure. Can you fix that up?

https://jenkins.saltstack.com/job/PR/job/salt-pr-lint-n/19076/violations/

salt/modules/kubernetes.py
```diff
@@ -622,6 +622,58 @@ def show_namespace(name, **kwargs):
...
+    except (ApiException, HTTPError) as exc:
+        if isinstance(exc, ApiException) and exc.status == 404:
+            return None
+        else:
+            log.exception(
+                'Exception when calling '
+                'ExtensionsV1beta1Api->patch_namespaced_deployment: '
+                '{0}'.format(exc)
+            )
+            raise CommandExecutionError(exc)
```
isbm: 1. There is no need to `return None`, in Python especially.
2. Your `log.exception` isn't done right: please use `%s` syntax.
3. Single exit is more welcome than multiple returns.
4. You use hard-coded 404 instead of constants from library.

So I would write it as follows:
```python
def set_container_image(.....
    ret = None
    try:
        ...
        ret = api_response.to_dict()
    except (ApiException, HTTPError) as err:
        if isinstance(err, HTTPError) and err.status != http.client.NOT_FOUND:
            log.exception('HTTP error: %s', six.text_type(err))
            raise CommandExecutionError(err)
    return ret
```

isbm (changes requested):
Please take a look at this. And could we also have a unit test for this? Thanks!

mbhinder:
Thanks for the reviews, I'll update the diff shortly.

isbm:
@mbhinder Thanks! I've updated my comment adding more issues. Please take a look at it.

rallytime:
Hi @mbhinder - Just checking in here to see if you were able to come back to this.

rallytime:
I am going to close this PR since we haven't heard back on this.

We will gladly re-open this once the comments above have been addressed and some tests are written for this. Or feel free to open a new PR. :)

Thanks!

Issue #33248, opened 2016-05-13
Execution module for FreeBSD Packet Filter


avb76:
### Description of Issue/Question

I searched for an existing execution module that would be an equivalent of `iptables` for FreeBSD's `pf`. I couldn't find one in the docs/repo ( I looked in the develop branch). Is there a planned `pf` execution module for the future? Or maybe there is one already but I've missed it. Thanks!
### Setup

(Please provide relevant configs and/or SLS files (Be sure to remove sensitive info).)
N/A
### Steps to Reproduce Issue

(Include debug logs if possible and relevant.)
N/A
### Versions Report

(Provided by running `salt --versions-report`. Please also mention any differences in master/minion versions.)
N/A

Ch3LL:
@avb76 i also did not see one available as well in the code. I will go ahead and label this as a feature request. i think this is a great idea :)

avb76:
@Ch3LL: great

jasperla:
@avb76 I just noticed this issue, you may want to have a look at the new module I've proposed in #45905 to see if it works on FreeBSD too.

backerman:
@jasperla Your pf module works just fine on FreeBSD; the only exception is that the available loglevels are different:

- none
- urgent
- misc
- loud

PR #45905, opened 2018-02-07
add execution module for OpenBSD Packet Filter

jasperla:
### What does this PR do?

Add a new module used to interact with the OpenBSD Packet Filter (pf). While it's possible to use pfctl on other platforms (e.g. FreeBSD or macOS) I have not yet verified it's working there.

#33248

### Tests written?

Yes

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/modules/pf.py
+++ b/salt/modules/pf.py
@@ -1,0 +1,349 @@
+# -*- coding: utf-8 -*-
+'''
+Control the OpenBSD packet filter (PF).
+
+:codeauthor: Jasper Lievisse Adriaanse <j@jasper.la>
+
+.. versionadded:: Fluorine
+'''
+
+from __future__ import absolute_import, print_function, unicode_literals
+
+# Import python libs
+import logging
+import re
+
+# Import salt libs
+import salt.utils.path
+from salt.exceptions import (CommandExecutionError, SaltInvocationError)
+
+log = logging.getLogger(__name__)
+
+
+def __virtual__():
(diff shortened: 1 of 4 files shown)


garethgreenaway (approved):

salt/modules/pf.py
```diff
@@ -0,0 +1,349 @@
+# -*- coding: utf-8 -*-
+'''
+Control the OpenBSD packet filter (PF).
+
+:codeauthor: Jasper Lievisse Adriaanse <j@jasper.la>
+
+.. versionadded:: Fluorine
+'''
+
+from __future__ import absolute_import
```
rallytime: For Py2/Py3 compatibility, we need this line to be:
```
```

tests/unit/modules/test_pf.py
```diff
@@ -0,0 +1,238 @@
+# -*- coding: utf-8 -*-
+
+# Import Python libs
+from __future__ import absolute_import
```
rallytime: Same comment here as well.

rallytime (changes requested):
Thanks @jasperla - I have one small request on this :)

salt/modules/pf.py
jasperla: Sure thing; I've just pushed the fix.

jasperla:
@rallytime this should be good to go now, right?

rallytime (approved):

Issue #45884, opened 2018-02-06
"TypeError: can't serialize <NodeImage" when calling salt-cloud with the dimensiondata driver


tintoy:
### Description of Issue/Question

Hi.

When running `salt-cloud -p ubuntu_dimensiondata test3`, I see the following error:

```
[ERROR   ] There was a profile error: can't serialize <NodeImage: id=42816eb2-9846-4483-95c3-7d7fbddebf2c, name=Ubuntu 14.04 2 CPU, driver=DimensionData  ...>
Traceback (most recent call last):
  File "/usr/local/src/saltstack/salt/salt/cloud/cli.py", line 281, in run
    self.config.get('names')
  File "/usr/local/src/saltstack/salt/salt/cloud/__init__.py", line 1454, in run_profile
    ret[name] = self.create(vm_)
  File "/usr/local/src/saltstack/salt/salt/cloud/__init__.py", line 1284, in create
    output = self.clouds[func](vm_)
  File "/usr/local/src/saltstack/salt/salt/cloud/clouds/dimensiondata.py", line 267, in create
    transport=__opts__['transport']
  File "/usr/local/src/saltstack/salt/salt/utils/cloud.py", line 1786, in fire_event
    event.fire_event(args, tag)
  File "/usr/local/src/saltstack/salt/salt/utils/event.py", line 718, in fire_event
    dump_data = self.serial.dumps(data)
  File "/usr/local/src/saltstack/salt/salt/payload.py", line 190, in dumps
    return msgpack.dumps(msg, use_bin_type=use_bin_type)
  File "/usr/local/lib/python2.7/dist-packages/msgpack/__init__.py", line 47, in packb
    return Packer(**kwargs).pack(o)
  File "msgpack/_packer.pyx", line 290, in msgpack._packer.Packer.pack
  File "msgpack/_packer.pyx", line 296, in msgpack._packer.Packer.pack
  File "msgpack/_packer.pyx", line 293, in msgpack._packer.Packer.pack
  File "msgpack/_packer.pyx", line 240, in msgpack._packer.Packer._pack
  File "msgpack/_packer.pyx", line 287, in msgpack._packer.Packer._pack
TypeError: can't serialize <NodeImage: id=42816eb2-9846-4483-95c3-7d7fbddebf2c, name=Ubuntu 14.04 2 CPU, driver=DimensionData  ...>
```

I'm sure I've seen this work before, but it fails reliably now (even, as mentioned, on a clean machine).

[...]

PR #45908, opened 2018-02-07
Fix for #45884 ("TypeError: can't serialize <NodeImage" when calling salt-cloud with the dimensiondata driver)

tintoy:
### What does this PR do?

* Fixes the behaviour of the `dimensiondata` cloud driver when deploying a new server by ensuring that event payloads are in a format that can be serialised by msgpack (i.e. only dicts, lists, and intrinsic data types; no raw Libcloud objects).
* Adds a couple of integration tests for the `dimensiondata` driver (there were none before) including one to deploy a server.

* "TypeError: can't serialize <NodeImage" error when calling salt-cloud to create a new server with the dimensiondata driver (saltstack/salt#45884)

### Tests written?

Yes (integration).

### Commits signed with GPG?

Yes. Using my key (C6AFE8C008E7F2B3).

### DIFF
--- a/salt/cloud/clouds/dimensiondata.py
+++ b/salt/cloud/clouds/dimensiondata.py
@@ -220,7 +220,6 @@ def create(vm_):
 
     log.info('Creating Cloud VM %s', vm_['name'])
     conn = get_conn()
-    rootPw = NodeAuthPassword(vm_['auth'])
 
     location = conn.ex_get_location_by_id(vm_['location'])
     images = conn.list_images(location=location)
@@ -251,7 +250,6 @@ def create(vm_):
     kwargs = {
         'name': vm_['name'],
         'image': image,
-        'auth': rootPw,
         'ex_description': vm_['description'],
         'ex_network_domain': network_domain,
         'ex_vlan': vlan,
@@ -259,7 +257,6 @@ def create(vm_):
     }
 
     event_data = _to_event_data(kwargs)
-    del event_data['auth']
 
     __utils__['cloud.fire_event'](
         'event',
@@ -269,6 +266,10 @@ def create(vm_):
         sock_dir=__opts__['sock_dir'],
         transport=__opts__['transport']
     )
+
+    # Initial password (excluded from event payload)
+    rootPw = NodeAuthPassword(vm_['auth'])
+    kwargs['auth'] = rootPw
 
     try:
         data = conn.create_node(**kwargs)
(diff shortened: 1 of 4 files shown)


tintoy:
CC: @gtmanfred

tintoy:
Ugh, this rebase is going to be harder than I thought. I'll ping you when it's ready to review.

tintoy:
Ok, ready to review (hooray for cherry-pick).

tintoy:
There's a separate issue where listing images doesn't work, but that's actually due to a bug in an upstream dependency (libcloud). I'll open a PR for that once the fix is in Libcloud.

tintoy:
Oh, and I ran flake8 and fixed anything relating to stuff I touched in this PR but that's it (I assumed you'd want to keep this PR focused on a single issue?).

tintoy:
(I suspect my rebase may have broken the CI build, but if it's possible to restart it then it should be good to go)

gtmanfred:
Go Go Jenkins!

tintoy:
Ok, signed too. That should be the last rebase :)

tintoy:
Ah, ok - my bad, looks like I missed one of the tests that's now broken. Will fix.

tintoy:
@gtmanfred - just realised that the dimensiondata driver has an existing (but indirect) dependency on `netaddr`, but it's not present when I install Salt or its dev dependencies.

Is it enough to add this to `pkg/*/req*.txt` or is there some other process for dependencies?

rallytime:
> I note that the docs for the driver indicate that this has to be installed for the driver to function so perhaps it's not supposed to be distributed by default, and instead installed only by those users that need it?

Yeah, that's exactly right. :)

rallytime:
@tintoy There's a lint error here: https://jenkins.saltstack.com/job/PR/job/salt-pr-lint-n/19100/violations/

gtmanfred:
A warning will be shown when that is not installed https://github.com/saltstack/salt/blob/oxygen/salt/cloud/clouds/dimensiondata.py#L133 so that those who need it will install it.

tintoy:
@rallytime oops sorry, thought I'd caught all of those last time around. Let's see if 1a75786b5a677c76d342385c05e0c540688ee57e does the trick (changed it to match the tests for the other providers).

tintoy:
@rallytime - it looks to me like the other failing tests are unrelated to this PR; am I correct?

rallytime (approved):
LGTM @tintoy - thanks for the fix. Yes, you're correct that the remaining failures are unrelated to your change.

gtmanfred (approved):

Issue #45868, opened 2018-02-05
LGPO No longer working


darrenjlowe:
After upgrading to 2017.7.3 on both master and minion, running salt "*MACHINENAME" lgpo.get machine return_full_policy_names=True brings back an error, on 2017.7.2 this was working

### Setup
(Please provide relevant configs and/or SLS files (Be sure to remove sensitive info).)

### Steps to Reproduce Issue
On salt master just run

salt "*MACHINENAME" lgpo.get machine return_full_policy_names=True

 The minion function caused an exception: Traceback (most recent call last):
     File "c:\salt\bin\lib\site-packages\salt\minion.py", line 1493, in _thread_return
       return_data = executor.execute()
     File "c:\salt\bin\lib\site-packages\salt\executors\direct_call.py", line 28, in execute
       return self.func(*self.args, **self.kwargs)
     File "c:\salt\bin\lib\site-packages\salt\modules\win_lgpo.py", line 5050, in get
       log.debug('Value {0} found for reg policy {1}'.format(class_vals[policy_name], policy_name))
   UnicodeDecodeError: 'ascii' codec can't decode byte 0x80 in position 0: ordinal not in range(128)

### Versions Report
Salt Minion

C:\salt>salt-minion --versions-report
Salt Version:
           Salt: 2017.7.3

Dependency Versions:
           cffi: 1.10.0
       cherrypy: unknown
       dateutil: 2.6.0
      docker-py: Not Installed
          gitdb: 2.0.3
      gitpython: 2.1.3
          ioflo: Not Installed
         Jinja2: 2.9.6
        libgit2: Not Installed
        libnacl: 1.6.1
       M2Crypto: Not Installed
           Mako: 1.0.6
   msgpack-pure: Not Installed
 msgpack-python: 0.4.8
   mysql-python: Not Installed
      pycparser: 2.17
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: Not Installed
         Python: 2.7.13 (v2.7.13:a06454b1afa1, Dec 17 2016, 20:42:59) [MSC v.1500 32 bit (Intel)]
   python-gnupg: 0.4.0
         PyYAML: 3.11
          PyZMQ: 16.0.2
           RAET: Not Installed
          smmap: 2.0.3
        timelib: 0.2.4
        Tornado: 4.5.1
            ZMQ: 4.1.6

System Versions:
           dist:
         locale: cp1252
        machine: AMD64
        release: 10
         system: Windows
[...]

Issue #45910, opened 2018-02-07
2016.11.9: UnicodeDecodeError traceback in reg.present


lorengordon:
### Description of Issue/Question

Was testing 2016.11.9 and am getting tracebacks when attempting to write binary data to the registry using reg.present. It was working fine in 2016.11.6 (haven't tested .7 or .8).

### Setup
It's a little complicated since the data is a certificate. Using the !py renderer is the only way I could get this to work in earlier versions.

cat test.sls:
```
#!py

import base64

CERTS = [
    {
        'id': 'SV-SV-78073r2_rule_02',
        'keys': [
            r'HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\SystemCertificates\Root\Certificates\73E8BB08E337D6A5A6AEF90CFFDD97D9176CB582',
            r'HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\EnterpriseCertificates\Root\Certificates\73E8BB08E337D6A5A6AEF90CFFDD97D9176CB582',
            r'HKEY_LOCAL_MACHINE\SOFTWARE\Policies\Microsoft\SystemCertificates\Root\Certificates\73E8BB08E337D6A5A6AEF90CFFDD97D9176CB582',
            r'HKEY_LOCAL_MACHINE\SOFTWARE\Policies\Microsoft\EnterpriseCertificates\Root\Certificates\73E8BB08E337D6A5A6AEF90CFFDD97D9176CB582',
        ],
        'vdata': base64.b16decode(
            '03000000010000001400000073e8bb08e337d6a5a6aef90cffdd97d9176cb582'
            '20000000010000005b030000308203573082023fa003020102020101300d0609'
            '2a864886f70d01010b0500304d310b3009060355040613025553311830160603'
            '55040a130f552e532e20476f7665726e6d656e74310c300a060355040b130345'
            '4341311630140603550403130d45434120526f6f742043412034301e170d3132'
            '303332303136313330345a170d3239313233303136313330345a304d310b3009'
            '06035504061302555331183016060355040a130f552e532e20476f7665726e6d'
            '656e74310c300a060355040b1303454341311630140603550403130d45434120'
            '526f6f74204341203430820122300d06092a864886f70d01010105000382010f'
            '003082010a0282010100b8706957bd805a41c019e695e7d48213dc2c384b8a82'
            'ca1c885d8e7e3c0c6f65e1dab81c7118b6de140879dbc68a99edfa21dd621800'
[...]

PR #45911, opened 2018-02-07
LGPO Module: Convert reg values to unicode for debug

twangboy:
### What does this PR do?
When the get function reads values from the registry it has the ability to return a `REG_BINARY` value. These values often cause the `log.debug` function to fail. This function converts the value to Unicode for the debug message. It also returns the actual value in the return when the data is invalid.

https://github.com/saltstack/salt/issues/45868

### Tests written?
No

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/modules/win_lgpo.py
+++ b/salt/modules/win_lgpo.py
@@ -2442,7 +2442,7 @@ class _policy_info(object):
                 elif ord(val) == 1:
                     return 'Enabled'
                 else:
-                    return 'Invalid Value'
+                    return 'Invalid Value: %s' % repr(val)
             else:
                 return 'Not Defined'
         except TypeError:
@@ -5047,7 +5047,10 @@ def get(policy_class=None, return_full_policy_names=True,
                     class_vals[policy_name] = __salt__['reg.read_value'](_pol['Registry']['Hive'],
                                                                          _pol['Registry']['Path'],
                                                                          _pol['Registry']['Value'])['vdata']
-                    log.debug('Value {0} found for reg policy {1}'.format(class_vals[policy_name], policy_name))
+                    log.debug(
+                        'Value {0} found for reg policy {1}'.format(
+                            salt.utils.to_unicode(class_vals[policy_name]),
+                            policy_name))
                 elif 'Secedit' in _pol:
                     # get value from secedit
                     _ret, _val = _findOptionValueInSeceditFile(_pol['Secedit']['Option'])
--- a/salt/modules/win_lgpo.py
+++ b/salt/modules/win_lgpo.py
@@ -2442,7 +2442,7 @@ class _policy_info(object):
                 elif ord(val) == 1:
                     return 'Enabled'
                 else:
-                    return 'Invalid Value: %s' % repr(val)
+                    return 'Invalid Value: {0!r}'.format(val)  # pylint: disable=repr-flag-used-in-string
             else:
                 return 'Not Defined'
         except TypeError:
--- a/salt/modules/win_lgpo.py
+++ b/salt/modules/win_lgpo.py
@@ -5048,9 +5048,9 @@ def get(policy_class=None, return_full_policy_names=True,
                                                                          _pol['Registry']['Path'],
                                                                          _pol['Registry']['Value'])['vdata']
                     log.debug(
-                        'Value {0} found for reg policy {1}'.format(
-                            salt.utils.to_unicode(class_vals[policy_name]),
-                            policy_name))
+                        'Value %r found for reg policy %s',
+                        class_vals[policy_name], policy_name
+                    )
                 elif 'Secedit' in _pol:
                     # get value from secedit
                     _ret, _val = _findOptionValueInSeceditFile(_pol['Secedit']['Option'])


damon-atkins:
under trace hexdump the binary data, under debug print a sha256 check sum.   See this #45910 issue why converting to unicode is a bad idea

rallytime:
@twangboy Lint: https://jenkins.saltstack.com/job/PR/job/salt-pr-lint-n/19101/violations/file/salt/modules/win_lgpo.py/

salt/modules/win_lgpo.py
```diff
@@ -2442,7 +2442,7 @@ def _binary_enable_zero_disable_one_conversion(cls, val, **kwargs):
                 elif ord(val) == 1:
                     return 'Enabled'
                 else:
-                    return 'Invalid Value'
+                    return 'Invalid Value: %s' % repr(val)
```
terminalmage: pylint won't like the percent formatting here. Also, if the string is unicode, repr will include the `u` before the string, is that what we really want?

What is it that we're trying to do here?

salt/modules/win_lgpo.py
```diff
@@ -5047,7 +5047,10 @@ def get(policy_class=None, return_full_policy_names=True,
                     class_vals[policy_name] = __salt__['reg.read_value'](_pol['Registry']['Hive'],
                                                                          _pol['Registry']['Path'],
                                                                          _pol['Registry']['Value'])['vdata']
-                    log.debug('Value {0} found for reg policy {1}'.format(class_vals[policy_name], policy_name))
+                    log.debug(
+                        'Value {0} found for reg policy {1}'.format(
+                            salt.utils.to_unicode(class_vals[policy_name]),
```
terminalmage: If this is just binary data, doing a `to_unicode` will fail. What content does this value contain?

```
>>> salt.utils.stringutils.to_unicode('\xaf')
  File "salt/utils/stringutils.py", line 109, in to_unicode
UnicodeDecodeError: 'utf8' codec can't decode byte 0xaf in position 0: invalid start byte
```

lomeroe:
I like @damon-atkins comment, hexdump and/or a checksum

salt/modules/win_lgpo.py
terminalmage: Chatted with @twangboy on this... the concern is that an invalid value is likely to be a bytestring, such as `\x80`, and we'd like to preserve the bytestring representation because it matches the hex value used in the registry. Instead of using `%` replacement though, we should be using the `!r` format flag and switching to a `.format()`. @twangboy is in the process of making this adjustment.

twangboy: Changed to `.format`

salt/modules/win_lgpo.py
twangboy: It looks like it's string data. `reg.read_value` will return unicode or str.

twangboy:
@lorengordon @damon-atkins I don't understand the hexdump/checksum recommendation.

lorengordon:
I think you mean @lomeroe... I don't understand the recommendation either.

twangboy:
Yeah, pardon me...

twangboy:
https://github.com/saltstack/salt/issues/45868#issuecomment-364571492

damon-atkins:
From https://github.com/saltstack/salt/issues/45910#issuecomment-364096193

'Value': salt.utils.stringutils.to_unicode(vdata, 'utf-8')} Needs to just set Value to SHA256 checksum of binary is {} when dealing with REG_BINARY

Output to log or screen should not contain binary data.
Maybe in this case do not set Value but set Checksum instead.

salt/modules/win_lgpo.py
terminalmage: Yeah, but on Python 2, binary blobs are represented in `str` types. The fact that it's `unicode` or `str` doesn't mean that it won't contain binary data.

terminalmage (commented):
See my recent comment [here](https://github.com/saltstack/salt/pull/45911#discussion_r167764504).

twangboy (commented):
@terminalmage So, should we handle it the same way you are here? https://github.com/saltstack/salt/pull/46000

salt/modules/win_lgpo.py
terminalmage: It's still a percent substitution, I don't think you updated it.

salt/modules/win_lgpo.py
terminalmage: If it's binary data, the `to_unicode` will fail, as I noted in my earlier comment. To prevent a decode error, we can just use `repr()`:

```python
log.debug(
    'Value %s found for reg policy %s',
    repr(class_vals[policy_name]), policy_name
)
```

terminalmage (changes requested):

salt/modules/win_lgpo.py
terminalmage: Oh, disregard... for some reason even though I refreshed the browser, when I clicked on "Files changed" it still had the old value there. After I submitted my review it magically updated.

salt/modules/win_lgpo.py
terminalmage: And actually I think you can use `%r` and not have to use `repr()`

terminalmage: Yep, that's true... So, this is more concise:

```python
log.debug(
    'Value %r found for reg policy %s',
    class_vals[policy_name], policy_name
)
```

terminalmage (changes requested):

twangboy:
@rallytime Updated the log statement as per @terminalmage 
Rebased...

terminalmage (approved):

PR #45921, opened 2018-02-08
yumpkg: Pass through kwargs from upgrade_available to latest_version

terminalmage:
The `upgrade_available` func is there as a convenience for people who are used to the yum/dnf syntax. However, it doesn't pass through kwargs and thus does not support the repo options that `latest_version` does. This commit makes the simple tweak of passing through the kwargs, allowing this convenience func to support all the extended options that `latest_version` does.

@rallytime up to you whether or not you think this merits a backport. It seems low-priority to me since one can just use `latest_version`.


rallytime (approved):

rallytime:
It's fine here, IMO.

PR #45925, opened 2018-02-08
Fix spelling error in docstring

terminalmage:


salt/modules/state.py
```diff
@@ -777,8 +777,8 @@ def highstate(test=None, queue=False, **kwargs):
 
         .. code-block:: bash
 
-            salt '*' state.higstate exclude=bar,baz
-            salt '*' state.higstate exclude=foo*
+            salt '*' state.hihgstate exclude=bar,baz
```
rallytime: These are still misspelled. :)

`hihg` --> `high`

The third example (which you didn't edit) is also misspelled. Can you fix all three?

rallytime (changes requested):

salt/modules/state.py
terminalmage: lol, epic fail. fixing now.

terminalmage: That third example looks like it was spelled properly in the first place.

rallytime (approved):

PR #45941, opened 2018-02-09
Fix incorrect translation of docker port_bindings -> ports

terminalmage:
The logic which ensures that we expose ports which are being bound, even
when not explicitly configured, was done incorrectly. UDP ports were being
passed to the API as '1234/udp' instead of (1234, 'udp'). This results in
the port not being exposed properly.

The logic has been corrected. Additionally both the "ports" input
translation function, as well as the post-processing code (where the port
numbers configured in port_bindings were being added) both contained code
to "fix" any ports which were configured using
'portnum/tcp', as these must be passed to the API simply as integers. To
reduce code duplication, this normalization is now only performed at the
very end of the post-processing function, after ports have been translated,
and any missing ports from the port_bindings have been added.

The unit test for the port_bindings input translation code, which was
written based upon the same incorrect reading of the API docs that resulted
in the incorrect behavior, have been updated to confirm the
(now) correct behavior. The unit test for the ports input translation code
has been updated to reflect the new normalization behavior.

Finally, an integration test has been added to ensure that we properly
expose UDP ports which are added as part of the post-processing function.

Note: This has been backported to the 2017.7 branch in https://github.com/saltstack/salt/pull/45942.

### DIFF
--- a/salt/modules/dockermod.py
+++ b/salt/modules/dockermod.py
@@ -630,7 +630,6 @@ def _client_wrapper(attr, *args, **kwargs):
         )
         ret = func(*args, **kwargs)
     except docker.errors.APIError as exc:
-        log.exception('Encountered error running API function %s', attr)
         if catch_api_errors:
             # Generic handling of Docker API errors
             raise CommandExecutionError(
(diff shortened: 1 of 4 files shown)


rallytime:
@terminalmage There are some related test failures: https://jenkins.saltstack.com/job/PR/job/salt-pr-linode-cent7-py3/2162/

rallytime (approved):

PR #45948, opened 2018-02-09
RFC: pkg.list_pkgs format change

isbm:
### What does this PR do?

This is an RFC for `pkg.list_pkgs` format change.
Please [read it here](https://github.com/isbm/salt/blob/9f2835e20e91d56997becdcc6742b074c18dad3a/rfcs/modules/0001-pkg:list_pkgs-output.md).


isbm:
@mirceaulinic @sjorge — optinions? Discussion about this?

damon-atkins:
```
python -B C:\salt\bin\Lib\site-packages\salt\utils\pkg\win.py
usage: C:\salt\bin\Lib\site-packages\salt\utils\pkg\win.py <detail|list> <system|system+user>
```
See the alternative  solution  publisher\software\type in oxygen

isbm:
@damon-atkins Could you please elaborate your opinion and discuss the RFC?

sjorge:
> It is still needs to know there were no slashes introduced in the naming of the packages to that particular operating system flavour.

IPS has slashes (/) in the package names. e.g. ```system/headers``` , so even if the delimiter was an argument to be passed, the caller would still need to know about the target operating systems. 

Sadly there is not something as a universal package manager.

> Structural example would be as follows: ...

I like this design, I think it is safe to make the assumption that a package will always have a name.
e.g. for the pkgsrc provider not much info is available aside from name, version and state (installed/not installed)

```
[isotope :: root][/workspace/salt][zfsutils ✗]
[.]# /opt/local/bin/pkgin ls
bash-4.4.012         The GNU Bourne Again Shell
binutils-2.26.1      GNU binary utilities
bmake-20150505       Portable (autoconf) version of NetBSD 'make' utility
...
```

Perhaps we can standardize this a bit more by defining a set rules for other fields, although I have no strong preference for this. As the package manager I usually deal with have a rather limited set of information available.

isbm:
@sjorge is there a package manager that has package that has no names?? I never heard of. As of `pkgsrc` yes, I know that. But there we just put all the _available_ attributes. And this is therefore better, because it will be agnostic to the underline PM.

sjorge:
@isbm I don't any PM that does not have names, hence the comment that having the name field be require should be the bare minimal. 

I will update the other comment to make this more clear.

damon-atkins:
Windows has no standard.
i.e. more than one company can use the same name for software (ignoring Trade Marks)
Hence publisher\\software name\\type   or vendor\\software name\\type
Type is anything.  e.g. x86, x64 ARM,  soft32, comp32, is determined by looking at the software properties.

Under windows you can have multiple version of the software installed.
So the data structure used is
`pkg['<publisher\\software name\\type>']['<version>']['<property>'] = value`

At this stage its double slash separator to avoid conflicts with a `/` being in software name. 
It might change however to a singe `\` and any `\` being escaped in the name.

At least other platforms have a standard, or do well at not having name clashes.

I disagree with as list of packages instead of a dict, as it would mean that data needs to be scanned all the time.  And then again what happens if a name is used twice, or combination of attributes is used twice except for the version string.   The windows specific data is prefixed with win_

```
    "google inc.\\\\google update helper\\\\comp32": [
        "1.3.21.169",
        "1.3.33.7"
    ]
```

```
    "google inc.\\\\google update helper\\\\comp32": {
        "arch": "x86",
        "vendor": "Google Inc.",
        "version": {
            "1.3.21.169": {
                "install_date": "2017-07-02T00:00:00",
                "install_date_time_t": 1498917600,
                "size": 25,
                "win_advertise_flags": 388,
                "win_display_name": "Google Update Helper",
                "win_install_src_dir": "c:\\Program Files (x86)\\Google\\Update\\1.3.21.169\\",
                "win_installer_type": "winmsi",
                "win_instance_type": 0,
                "win_language": "en_US",
                "win_package_code": "{59E40BF5-2D7D-4264-914F-FF663AF93FE4}",
                "win_package_name": "GoogleUpdateHelper.msi",
                "win_product_code": "{A92DAB39-4E2C-4304-9AB6-BC44E68B55E2}",
                "win_redeployment_flags": 3,
                "win_uninstall_cmd": "MsiExec.exe /I{A92DAB39-4E2C-4304-9AB6-BC44E68B55E2}",
                "win_upgrade_code": "{9515FFBD-40AB-4690-B983-4FA8E14EE7F5}",
                "win_version_binary": "1.3.21",
                "win_version_display": "1.3.21.169",
                "win_version_user": "1.3.21.169"
            },
            "1.3.33.7": {
                "install_date": "2017-11-14T00:00:00",
                "install_date_time_t": 1510578000,
                "size": 82,
                "win_advertise_flags": 388,
                "win_display_name": "Google Update Helper",
                "win_install_src_dir": "C:\\Program Files (x86)\\Google\\Update\\1.3.33.7\\",
                "win_installer_type": "winmsi",
                "win_instance_type": 0,
                "win_language": "en_US",
                "win_package_code": "{646E8264-633A-4689-A853-CE5123D8978A}",
                "win_package_name": "GoogleUpdateHelper.msi",
                "win_product_code": "{60EC980A-BDA2-4CB6-A427-B07A5498B4CA}",
                "win_redeployment_flags": 3,
                "win_uninstall_cmd": "MsiExec.exe /I{60EC980A-BDA2-4CB6-A427-B07A5498B4CA}",
                "win_upgrade_code": "{19B02465-7B0B-4816-9CB9-7CA4AB9BC078}",
                "win_version_binary": "1.3.33",
                "win_version_display": "1.3.33.7",
                "win_version_user": "1.3.33.7"
            }
        },
        "win_system_component": true,
        "win_version_src": "display-version"
    },
```

isbm:
> I disagree with as list of packages instead of a dict, as it would mean that data needs to be scanned all the time.

It needs to be scanned anyways. Except that you are iterating over the keys bound to sets, but now you will be iterating over the sets directly. Don't see much difference here.

> And then again what happens if a name is used twice, or combination of attributes is used twice except for the version string. 

Well, that is exactly the problem which this RFC is solving, if you read it carefully. Right now we are trying to be very smart and re-grouping data, hoping nothing wrong is happening. But on Linux we have a trouble with the different packages that has the same name: you cannot _assume_ that there is no dot or other symbols in the name(s) in a future of the same platform, as there is no official standard, just conventions _per distribution_ :warning: and hope nobody breaks them. And if you've got `foo.i686` you have no way to know it is a name of the package or something else, unless your client code is filled in with the assumptions and ad-hoc hard-coded `if/else` statements that matches particular OS family.

This is why I propose to keep it simple and stupid: just include/exclude data (`attr`) and that's it.

aphor:
I recently found myself wanting something like this for Macports packages.

Macports complicates the package architecture facet/attribute/column by adding an extensible set of "variants" which are usually boolean indicators of compile time options. Sometimes these are mandatory and defaults are provided. Sometimes they have conflicts (mutual exclusivity) defined between variants, and sometimes they are mutually compatible. Macports also supports installed but inactive (staged) packages as long as the tuple of the attributes are unique: (name, version, architecture, tuple(set(variants))).

I'm in favor of a flexible representation of package metadata. I think that the current salt/states/pkg.py logic could be done more elegantly, and this is a good opportunity to move it forward.

lucidd:
@damon-atkins a user does not have to scan it every time. They only have to scan it once to either find what they want directly or turn it into a dict indexed by what every key fits their use case best.
The using a attribute as key for efficient access argument breaks down pretty quick when you get into multi attribute key encodings and special cases. If you do stream processing for example via the netapi and a capable json parser you can even construct your dict as elements come in so no time lost compared to sending a dict.
The current dict/key style only complicates modules and user code alike for no added benefit.

meaksh:
I like more the proposal of changing the output structure than allowing the use of a delimiter, which is not really OS/distribution agnostic. As @sjorge mentioned, maybe `version` should be included as a fixed key on the output dictionary, since every PM, and of course the Salt itself, are meant to deal with packages that have, at least, name & version.

Would the following make more sense?
```json
[
	{
		"name": "package-name",
		"version": "package-version",
		"key": "value",
	}
]
```

Thanks for the proposal @isbm !

isbm:
We should have some sort of interface/validation of those keys and make them common. It is OK to extend the interface, once it is needed (say, Windows needs something that others do not even have). But those keys needs to be same across everywhere the same.

damon-atkins:
Two items common  `title of software` and support `multiple version` installed at the same time.
pkg.install requires a `title of software`  and `version`
Code which interprets the information this must be able to separate version if its included in the title, otherwise the title is not common amongst versions.

isbm:
> version if its included in the title

There is no version included in the title:
```
$ rpm -qa --queryformat='%{name}\n' | head
classpathx-mail
zypper
libX11-xcb1
libz1
apache-commons-httpclient
libnetfilter_conntrack3
libxslt-tools
libblkid1
snapper-zypp-plugin
auditlog-keeper-rdbms
```

And if it looks like it is, example here:
```
$ dpkg -l | grep yaml | awk '{print $2}'
libcpan-meta-yaml-perl
libyaml-0-2:amd64
libyaml-cpp0.3v5:amd64
libyaml-libyaml-perl
libyaml-tiny-perl
python-yaml
python3-yaml
yaml-mode
yamllint
```
...then it is most of the time a complete mess:
```
$ apt-cache show libyaml-0-2:amd64 | grep Version
Version: 0.1.6-3
$ apt-cache show libyaml-cpp0.3v5:amd64 | grep Version
Version: 0.3.0-1.2
```

And this example is another evidence why current way is bad: no dot as a separator  between the name and the arch, but a colon (unlike on SUSE, CentOS etc).

damon-atkins:
What is the command line to install `libyaml-0-2:amd64` and what would the data structure look like.
```
[
	{
		"name": "libyaml-0-2:amd64",
		"install_date": "2017-06-24T10:51:34Z",
		"version": "0.1.6-3",
		"arch": "amd64",
		"install_date_time_t": 1498301494
	},
```
 
Can not talk about the data structure and leave pkg.install or pkg.installed out of the conversation.
If you propose a data structure change then you are also proposing a pkg.install and pkg.installed change.

isbm:
@damon-atkins I am not sure you are on the same page. We are talking about `pkg.list_pkgs`. Not sure why you brought to the discussion all that above. And your example is a bit wrong. The name is this:
```
[
	{
		"name": "libyaml-0-2",
...
```

damon-atkins:
pkg.list_pkgs is used by pkg.install and pkg.installed to determine the current state. Hence its part of the discussion.  It could be we are on different pages.

aphor:
I agree with what I think @isbm is trying to say about the name.

    libyaml-0-2:amd64
and
    libssl2.0.0/x86

are strings packed with (I think) name, version, epoch, and arch metadata, which IMHO **should** be represented internally with something like
    [
      { 'name': 'libyaml',
        'version': (0,),
        'epoch': 2,
        'arch': 'amd64' },
      { 'name': 'libssl',
        'version': (2,0,0),
        'arch': 'x86' }
    ]

https://docs.python.org/3/library/functions.html#filter
I think it would be cleanest to implement a filter functor (a class with __call__(self, iterable) method, that returns a generator of booleans representing match or not), constructed with pkg.installed/removed or pkg.install/remove arguments, for example, which will filter an iterable of packages returned by pkg.list_packages, which match the criteria given in the constructor. It should operate exactly the same on installed packages list and available packages list.

isbm:
@aphor a priori, the name should not be a source for parsing version, revision and/or architecture. This is basically just a blob of characters that package manager should just "get it" as is. Everyone who is trying to _parse the name_ is asking for a trouble. Properly to do this is to actually ask particular Package Manager, like so:

```
$ rpm -qa | grep -i pyyaml
python-PyYAML-3.12-25.1.x86_64

$ rpm -qa --queryformat='%{name}\n' | grep -i pyyaml
python-PyYAML

$  rpm -qa --queryformat='%{name} -- %{version} -- %{release} -- %{arch} -- %{packager}\n' | grep -i pyyaml
python-PyYAML -- 3.12 -- 25.1 -- x86_64 -- https://www.suse.com/

```

damon-atkins:
```
rpm -q --queryformat='%{name}/%{arch} -- %{version} -- %{release} -- %{arch} -- %{packager}\n'  libxml2-python python-urllib3
libxml2-python/x86_64 -- 2.9.1 -- 6.el7_2.3 -- x86_64 -- CentOS BuildSystem <http://bugs.centos.org>
python-urllib3/noarch -- 1.10.2 -- 3.el7 -- noarch -- CentOS BuildSystem <http://bugs.centos.org>
```
seems to work

rallytime:
@isbm Where did we land on the discussion of this above? I know this hasn't been reviewed by any core members yet. It looks like there is still some discussion going on here.

isbm:
@rallytime yes, we still discuss this, so the core team should look at this.

gtmanfred (commented):
>>    I disagree with as list of packages instead of a dict, as it would mean that data needs to be scanned all the time.

>It needs to be scanned anyways. Except that you are iterating over the keys bound to sets, but now you will be iterating over the sets directly. Don't see much difference here.

Where did we end up on this, if it is a dict, why would we have to scan the whole list of packages, instead of having the name of the package as the top level of the dict, so we could at least only have to scan a subset of packages?

gtmanfred:
You example is just that we aren't currently handling the architecture as part of the name, if that was fixed, and the pkg.list_pkgs format was supposed to be just the name, and all other information went into that packages list, you should be able to determine the name.

The architecture can then be included in the list.

I think `you cannot reliably have a name of the package as a top level` is a little disingenuous, it isn't that you cannot do it, it is just that it is more difficult, because each package manager does it slightly differently.

damon-atkins:
FYI Solaris Pkgs:
```
 pkg info x11/xclock  # or just xclock will work as there is no other something/xclock
          Name: x11/xclock
       Summary: xclock - analog / digital clock for X
   Description: xclock is the classic X Window System clock utility.  It
                displays the time in analog or digital form, continuously
                updated at a frequency which may be specified by the user.
      Category: Applications/Accessories
         State: Installed
     Publisher: solaris
       Version: 1.0.5
 Build Release: 5.11
        Branch: 0.175.0.0.0.0.1215
Packaging Date: 27 September 2011 12:59:05 PM
          Size: 71.88 kB
          FMRI: pkg://solaris/x11/xclock@1.0.5,5.11-0.175.0.0.0.0.1215:20110927T125905Z
```
The FMRI is the full-name of the package is `pkg://<repo name>/<area>/<name>@<version>:<time>`
```
pkg list -Hv | egrep xclock
pkg://solaris/x11/xclock@1.0.5,5.11-0.175.0.0.0.0.1215:20110927T125905Z      i--
```

damon-atkins:
From DNF man page
SPECIFYING PACKAGES
       Many  commands  take  a  <package-spec> parameter that selects a package for the operation. DNF looks for interpretations of the parameter from the most commonly used meanings to the least, that is it  tries to see if the given spec fits one of the following patterns (in decreasing order of priority):
* name.arch
* name
* name-[epoch:]version-release.arch
* name-[epoch:]version-release
* name-[epoch:]version
Note that name can in general contain dashes (e.g. package-subpackage).

Failing to match the input argument to an existing package name based on the patterns above, DNF tries to see if the argument matches an existing provide.

By default, if multiple versions of the selected package exist in the repo, the most recent version suitable for the given operation is used. If the selected package exists for multiple architectures, the  packages  which best match the system's architecture will be preferred. The name specification is case-sensitive, globbing characters `"?, *` and `[` are allowed and trigger shell-like glob matching. If globbing character is present in name, DNF expands given name first and consequently selects all packages matching expanded `<package-spec>`.

isbm:
RPM and packaging is something that you never can properly predict even from the distro to distro. So saying "a little bit more difficult" usually implies involving some sort of guesswork and try/error approach. Which I propose to avoid _by design_.

damon-atkins:
What is important is the assumptions of yum, dnf etc   have about rpms'. Otherwise they would not work.

damon-atkins:
For DNF/YUM 
e.g.
```
{
  "name.arch": [
  {
    "name": "<full name>",
    "version": "1.5",
    "release": "fc27",
    "etc": "something"
    }
  ]
}
```
OR
```
{
  "name": [
  {
    "name": "<full name>",
    "version": "1.5",
    "release": "fc27",
    "arch": "x86_64"
    "etc": "something"
    }
  ],
[
  {
    "name": "<full name>",
    "version": "1.5",
    "release": "fc27",
    "arch": "x86"
    "etc": "something"
    }
  ]
}
```

isbm:
@aphor I tend to think the more we trying to get smarter, the worse it gets. If that would be only a silo OS like Windows or OSX, where strict laws exist, then we can hope that Tim Cook won't cook them in the next decade... But at SUSE over many years we hit countless cases where "being too smart" usually only brings problems, once you've hit a corner-case, of which you never thought of (and physically could not). Because design stays rigid for long, but the underneath distros around changing very fast.

And therefore I tend to see this problem as much generic as it can be: just work with the packages as it is, avoiding ameliorating/palliating here as much as one can do it. Besides, even if that would be a 50K entries in the list, it is still a fraction of a second to loop over it for a minion to find out right package.

rfcs/modules/0001-pkg:list_pkgs-output.md
```diff
@@ -0,0 +1,297 @@
...
+"xz-libs": ....
+```
+
+However, this does not solve two issues:
+
+1. There is no mention in the name what is the default architecture,
+   unless explicitly requested, since prepended architecture is only
+   different from the default. We only _assume_ that _most of the
+   time_ it is `x86_64` architecture which is capable running `i686`
+   or older.
```
terminalmage: The default arch for a minion is defined in the `osarch` grain, just FYI.

terminalmage (commented):
I still prefer a dict for this sort of thing. Consider the case of passing dozens of packages via the `pkgs` argument. You're forcing yourself to either:

1. Scan through the list once _for each package_.
2. Scan through the list once, but _check each desired package_ for each item in the list.

Neither of these sounds like a good idea.

I don't understand why we can't just assume we'll have a package name, and make that the dict key, with a list of dictionaries under each key. What is the reasoning for not doing this?

damon-atkins:
Ping @saltstack/team-core

rfcs/modules/0001-pkg:list_pkgs-output.md
```diff
@@ -0,0 +1,297 @@
...
+# Drawbacks
+[drawbacks]: #drawbacks
+
+- This change should not go to the prepared Oxygen release, but only in
+`develop` branch.
+
+- Implementation should be done using deprecation model system
+  (decorators we already have in Salt core).
+
+- The old version should stay for two releases (standard retention time)
```
cachedout: So there would be no changes out of the box until Magnesium, correct?

isbm: I'd say, we need to maintain at least two versions backward-compatible in everything we change so drastically.

lucidd:
@terminalmage while name with list of packages would probably work but there is really no reason for it.

- Your concern about scanning still applies to any user interested in any other grouping then by name. 
- Scanning itself is less of a problem then you make it out to be. Like i pointed out in my previous answer if you have to scan its only once to find what you are looking for or for building up a dict with keys for your use case. Also since salt is usually used remotely those data structures are usually serialized and must be deserialized before you can do anything. Modern serializes usually have an option to deserialize individual elements of a list and handle them right away in a streaming way. In this case scanning is the most efficient thing you can do since you never even need to have everything in memory at once and slow network does not block processing.
- any preprocessing / grouping you do and the user does not need/want is wasted computation.

In the end you are trying to abstract over all package managers salt supports now and potentially in the future while package managers themselves are also moving targets. A list of packages is the most basic and abstract representation that makes the least assumptions.

terminalmage:
>- Your concern about scanning still applies to any user interested in any other grouping then by name. 

Yeah, and so what? I don't care about a performance hit for someone's edge case on how they want to organize the data. My concern is how Salt will use it internally, as that will make up the vast majority of calls to `pkg.list_pkgs`.

>- Scanning itself is less of a problem then you make it out to be. Like i pointed out in my previous answer if you have to scan its only once to find what you are looking for or for building up a dict with keys for your use case. Also since salt is usually used remotely those data structures are usually serialized and must be deserialized before you can do anything. Modern serializes usually have an option to deserialize individual elements of a list and handle them right away in a streaming way. In this case scanning is the most efficient thing you can do since you never even need to have everything in memory at once and a slow network does not block processing.

I'm not sure what "running remotely" has to do with anything. First of all, most invocations of `pkg.list_pkgs` happen on the minion. That is, a state function running on the minion calls the function, and the results are returned directly to the calling function. No serialization. Secondly, even in cases where pkg.list_pkgs is run and sent back to the master (where we actually _do_ serialize/deserialize the data), _we are not deserializing them one element at a time_. We are using `msgpack.load()` to deserialize to a data structure. And in those cases, we are not performing additional processing on the returns, we're writing to the job cache and displaying the results on the CLI.

>- any preprocessing / grouping you do and the user does not need/want is wasted computation.

:laughing:

>In the end you are trying to abstract over all package managers salt supports now and potentially in the future while package managers themselves are also moving targets. A list of packages is the most basic and abstract representation that makes the least assumptions.

I don't think you really understand what I was explaining in the first place. I'm not talking about abstracting anything. All packages have names, this isn't some esoteric feature of package managers that has to be normalized.

I really wish everyone arguing for this function to return a list would go and read the source code for the pkg states that rely on `pkg.list_pkgs`. See how many times we look up what version of a desired package is installed from the `pkg.list_pkgs` return data, and then imagine every single one of these lookups needing to be replaced with a sequential scan of the `pkg.list_pkgs` return data.

This is madness. I don't know why it's still being entertained.

isbm:
> And in those cases, we are not performing additional processing on the returns, we're writing to the job cache and displaying the results on the CLI.

If I understand it correctly, @lucidd 's concern is when this happens over the API. And I thin the RFC refers to the API problems too. So when API returns current structure, it is probably would be a better when the result is not grouped, as one of the options. So from the 3rd party user perspective, accessing and processing un-grouped data is probably better.

But I also understand @terminalmage 's valid points that between minion and the master deserialisation is not streamed as well as having it _at least_ O(n^2) or even more nesting isn't the best idea indeed. This is something that was missing from the original proposal.

Therefore to me it boils down to the following:
- Salt should not burn a hole in CPUs while gathering this data for particular call
- The _mixed heterogeneous_ environment API-driven 3rd party client should not over-hammer Salt in order to get all needed info

Can we find compromise here? Let me iterate again over the [raised concerns in the original RFC](https://github.com/isbm/salt/blob/9f2835e20e91d56997becdcc6742b074c18dad3a/rfcs/modules/0001-pkg:list_pkgs-output.md#current-design):

1. *API caller should already know what system it is targeting (is this RPM based 
    or not). Unless this information is already known, another additional call for 
    grains is required.*

Looking at this again, I guess, this is not a big problem, because grains should be known prior anyway. In _any_ case it is better to get grains of the machine and cache it somewhere, instead of _every time_ rescan stuff on it. Even from that perspective this problem doesn't look to me very much as problematic.

2. *The format is not the same across all the distributions and operating systems,
    while it should be.*

But here is the problem. Format of the _name_ is not the same for the same software on different distributions, and so API client will need to prepare some set of `if/else`. E.g. some say `xclock` and some say `xclock.686` etc. Solution for this I would propose below.

3. *The fact that it is built on top of the mapping on top of keys as the name
    already brings by-design limitaions for the packages with the same name.
    To overcome it -- an architecture prepending.*

Which this is true. We should then think where possibly we can have two identical packages installed twice? Normally it is so:

- Different vendor and thus location, say in Solaris it is `/opt/usr` vs `/usr` etc. Something similar happens on Windows.
- Different versions (mostly only kernels)
- Different architecture (can happen to any arch-dependent)

@lucidd  how about this:

```
"xz-libs": [
	{
		"release": ["1.el7", "1.el9"],
		"install_date": ["2017-06-24T10:51:34Z", "2017-06-24T10:51:35Z"],
		"version": ["5.2.2", "5.2.3"],
		"arch": ["x86_64", "686"],
		"install_date_time_t": [1498301494, 1498301526]
	}
	],
```

As you can see, you've got the name without prepending anything. But you've got extra data for the second package (you can add third etc). Here data from the second package just added to the lists as a second element. Now, in case only arch is different, as the in original RFC example, that would look like so:

```
"xz-libs": [
	{
		"release": ["1.el7", "1.el7"],
		"install_date": ["2017-06-24T10:51:34Z", "2017-06-24T10:51:34Z"],
		"version": ["5.2.2", "5.2.2"],
		"arch": ["x86_64", "686"],
		"install_date_time_t": [1498301494, 1498301494]
	}
	],
```
As you can see above, if the data from the second package doesn't differ, it is still added anyway. By this API client can easily dissect it and regroup as they want. @lucidd does it solve your concern? Because I think it does solve @terminalmage 's concern not to rescan every time inside the minion in order to realise the package is there yet or not.

4. *No guarantee that there will not be any dots or colons in the names in a future in any possible distribution.*

This is just a name parsing problem, can be also discarded alongside with the third issue, since it solves the problem.

@lucidd would be users of API like yourself be happy, once we change type of the arch field to a list and so if there is more than one element, then different packages installed? You can also re-group it yourself then.

@terminalmage any corrections to my reasoning above?

damon-atkins:
Eventual list_pkgs on windows will call a class.  This object returns a point in time snapshot of the list of software installed, then you ask the object to return data in the format expected to be returned by list_pkgs. And the class provides a querying interface. So that knowledge of the internal structure is not required. 

I can see the benefit of being able to do
```
xz-libs:
  - pkg.installed
  - arch x86  # on a x86_64 platform
  - release: '>1'
```
e.g. instead of just a package name provide a set of filters on attributes to select the package to take action against.
notepad++  32bit and 64bit could both be removed with pkg.remove notepad++ because arch was not filtered on.

By states/pkg.py  using a class/object a lot of the internal structure of holding the data can be hidden.

isbm:
Another example could be done here (suggested by @terminalmage ):
```
"xz-libs": [
        {
            "version": "1.2.3-4.el7",
            "arch": "i686",
        },
        {
            "version": "1.2.3-5.el7",
            "arch": "x86_64",
        }
]
```
This is still key/value, but contains different values. @lucidd would that be sufficient for you?

terminalmage:
The option I discussed with @isbm (and described in his earlier comment [here](https://github.com/saltstack/salt/pull/45948#issuecomment-417336326)) is the best way to serve API users and keep the software as performant as possible.

damon-atkins:
Just need to define mandatory (e.g. name and version) and optional, and platform specific keys
e.g. do we prefix windows specific items with win_    rpm specific items with rpm_ you should the be sub dicts.  e.g.  #45948#issuecomment-364646934

Unfortunately some 64bit programs have poor installs which place install information in the 32 bit registry instead of the 64bit regsitry. Which makes it impossible to determine if the app is 32bit or 64bit.

pkg.install/remove would be good if they had a filter to select what software to install or remove?

rfcs/modules/0001-pkg:list_pkgs-output.md
thatch45: I thkn that the way to do this would be to make the new function `pkg.list` and place a deprecation warning on list_pkgs.

damon-atkins:
Maybe `pkg.list_installed`.   One thing I like about powershell is it has a well defined naming standard. i.e. https://docs.microsoft.com/en-us/powershell/developer/cmdlet/approved-verbs-for-windows-powershell-commands

dwoz:
@isbm We have standardized the RFC process and it has become "Salt Enhancement Proposals". Can you please create a PR for this one In the new repository and link back to the discussion here?
https://github.com/saltstack/salt-enhancement-proposals/pulls

dwoz:
@isbm Closing this due to inactivity. If this comes up again please open a new PR [here](https://github.com/saltstack/salt-enhancement-proposals/pulls) with a link back to this one.

PR #45961, opened 2018-02-09
Fix issue with reg.present

twangboy:
### What does this PR do?
When generating the changed item, it would crash when it encountered binary data because it was trying to convert to unicode (`salt.utils.to_unicode`).

This will use `salt.utils.to_str` instead.

https://github.com/saltstack/salt/issues/45910

### Tests written?
No

### Commits signed with GPG?
Yes


terminalmage (commented):
This would break in Python 3, bytestrings containing binary data do not decode to strings.

I recommend closing this in favor of #46000.

terminalmage:
More info here: https://github.com/saltstack/salt/issues/45910#issuecomment-365151568

twangboy:
@terminalmage Looks like you found a better way to fix it here: https://github.com/saltstack/salt/pull/46000

Issue #45924, opened 2018-02-08
pkg.upgrade with exclude does not exclude package on CentOS 7


Kaydub00:
According to documentation at https://docs.saltstack.com/en/latest/ref/modules/all/salt.modules.yumpkg.html using pkg.upgrade I should be able to exclude packages using the pkg.upgrade module.

## Setup
Salt-minions running from a Master

### Salt-Master:
Salt Version:
           Salt: 2017.7.3
System Versions:
           dist: centos 7.4.1708 Core
         locale: UTF-8
        machine: x86_64
        release: 3.10.0-693.17.1.el7.x86_64
         system: Linux
        version: CentOS Linux 7.4.1708 Core

### Salt-Minion
Salt Version:
           Salt: 2017.7.3
System Versions:
           dist: centos 7.4.1708 Core
         locale: UTF-8
        machine: x86_64
        release: 3.10.0-693.11.6.el7.x86_64
         system: Linux
        version: CentOS Linux 7.4.1708 Core

## Reproduce issue

`salt 'hostname' pkg.upgrade exclude='kernel*'`

Does not actually exclude the kernel as stated in the documentation.

perfecto25:
btw, I tested this w Salt salt 2018.3.4 (Oxygen)

on centos 7.5 target and its excluding kernel correctly

May need to upgrade your salt version to latest

vosscodes:
also working as expected with 2019.2.0 on centos 7.6

dwoz:
Fixed by ee6bf2e6345d (Feb 2018, in 2018.3); commenters confirmed in 2018.3.4 and 2019.2.0. Verified on 3006.x/3007.x/3008.x/master: 'pkg.upgrade exclude=kernel*' produces 'yum --quiet -y --exclude=kernel* upgrade'.

PR #45963, opened 2018-02-09
 [2017.7] Fixes to modules/yumpkg.py

garethgreenaway:
### What does this PR do?
When callilng cmd.run_all with a list the exclude argument was being ignored and packages were being upgraded.  This change converts the list to a string with all the elements before sending it along to cmd.run_all.

#45924 

### Commits signed with GPG?
Yes


terminalmage (commented):
This may actually cause problems if one uses a wildcard in fromrepo/disablerepo. I'm working on my own fix for this which preserves the use of the list of args for the command, and also adding some unit tests.

PR #45965, opened 2018-02-10
Simplify firewall rules

rjc:
As with `firewall-cmd`, use a single rule for ports in both, the `iptables` and `pf` examples. BTW, there's no need for `-m mport` or `-m multiport` (and these might not even be available if not compiled) for `iptables` or using a list or a macro in `pf.conf` if a contiguous range is being used - essentially, use colon ('`:`') for simplicity and consistency.

### What does this PR do?

Simplifies `iptables` and `pf` rules.

None - the changes simply make the rules shorter and cleaner, thus easier to adjust and debug.

### Tests written?

No need - this is documentation only.

### Commits signed with GPG?

No

### DIFF
--- a/doc/topics/tutorials/firewall.rst
+++ b/doc/topics/tutorials/firewall.rst
@@ -163,13 +163,12 @@ common locations, but your mileage may vary.
 
 Follow these instructions: https://wiki.debian.org/iptables
 
-Once you've found your firewall rules, you'll need to add the two lines below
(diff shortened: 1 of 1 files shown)


gtmanfred (approved):

garethgreenaway (approved):

rallytime:
Thanks @rjc and welcome!

Issue #43798, opened 2017-09-28
minion proxy settings causing http GETs to fail in consul module


campbellmc:
### Description of Issue/Question

Trivial consul module GET requests fail when proxy is enabled, **and** when run in a state after a file managed with remotefs:

`2017-09-27 15:34:04,283 [salt.state       ][ERROR   ][27010] 'consul.agent_info' failed: initialize() got an unexpected keyword argument 'max_body_size'`

I don't believe it's actually a consul module problem - it appears more to do with tornado not closing or re-initialising for subsequent http requests in a state file.

### Setup

Minions have proxy configuration (including no_proxy)
PR Feature **add no proxy support** #43764 has been monkey patched on 2017.7.1 

### Steps to Reproduce Issue

The API call works fine from the command line on the minion:
```
# curl -X GET --noproxy 127.0.0.1 127.0.0.1:8500/v1/agent/self
{"Config":{"DevMode":false,"Limits":{"RPCRate":1.7976931348623157e+308,"RPCMaxBurst":1000},"Performance":{"RaftMultiplier":0},"Bootstrap":false,"BootstrapExpect":0,"Server":false......
```
Minions have proxy configuration:
```
proxy_host: proxy.service.consul
proxy_port: 8080
no_proxy: ["127.0.0.1", "localhost",]
```

Run agent.self method in a state file:
```
get_agent_info:
  module.run:
    - consul.agent_self:
      - consul_url: http://127.0.0.1:8500
```
```
# salt kevin state.apply consul_test.agent_self
kevin:
13:44:48.715348 [ 51.189 ms]     module.run        Changed   Name: get_agent_info

Summary for kevin
------------
Succeeded: 1 (changed=1)
Failed:    0
------------
Total states run:     1
Total run time:  51.189 ms
```

Add a file.managed state with remotefs before the consul GET module run:
```
test_http_source:
  file.managed:
    - name: /tmp/test.txt
    - source: https://sentry.io
    - skip_verify: True

get_agent_info:
  module.run:
    - consul.agent_self:
      - consul_url: http://127.0.0.1:8500
```
consul module agent_self method now fails
```
# salt kevin state.apply consul_test.agent_self
kevin:
13:41:40.709597 [1195.605 ms]       file.managed    Changed   Name: /tmp/test.txt
13:41:41.906982 [ 11.617 ms]     module.run        Failed    Name: get_agent_info

Summary for kevin
------------
Succeeded: 1 (changed=1)
Failed:    1
------------
[...]

PR #45970, opened 2018-02-11
stop 'data' being assigned with serialised empty dict for GET requests

campbellmc:
### What does this PR do?
stops 'data' param being set for GET requests in module consul.py

https://github.com/saltstack/salt/issues/43798

### Previous Behavior
if 'data' was None, it would be set to json serialised empty dict which causes exception in http modules.

### New Behavior
For GET requests, force data = None.

### Tests written?
No

### Commits signed with GPG?
No

### DIFF
--- a/salt/modules/consul.py
+++ b/salt/modules/consul.py
@@ -61,7 +61,7 @@ def _query(function,
     :param api_version  The Consul api version
     :param function:    The Consul api function to perform.
     :param method:      The HTTP method, e.g. GET or POST.
-    :param data:        The data to be sent for POST method.
+    :param data:        The data to be sent for POST method. This param is ignored for GET requests.
     :return:            The json response from the API call or False.
     '''
 
--- a/salt/modules/consul.py
+++ b/salt/modules/consul.py
@@ -78,9 +78,12 @@ def _query(function,
     base_url = urllib.parse.urljoin(consul_url, '{0}/'.format(api_version))
     url = urllib.parse.urljoin(base_url, function, False)
 
-    if data is None:
-        data = {}
-    data = salt.utils.json.dumps(data)
+    if method == 'GET':
+        data = None
+    else:
+        if data is None:
+            data = {}
+        data = salt.utils.json.dumps(data)
 
     result = salt.utils.http.query(
         url,


salt/modules/consul.py
```diff
@@ -78,9 +78,12 @@ def _query(function,
     base_url = urllib.parse.urljoin(consul_url, '{0}/'.format(api_version))
     url = urllib.parse.urljoin(base_url, function, False)
 
-    if data is None:
-        data = {}
-    data = salt.utils.json.dumps(data)
+    if method == 'GET':
```
rallytime: Is there a chance that the `GET` method would ever be used with `data`?

campbellmc: There's nothing to say that a body payload cannot be sent in a `GET` request.
Your mileage may vary.
IMHO it's not a particularly smart thing to do.  Tornado is dogmatic about it and will complain.

```
>>> data=json.dumps({})
>>> data
'{}'
>>> res=salt.utils.http.query('https://github.com/saltstack/salt/blob/develop/salt/utils/http.py',method='GET',data=data)
  File "/usr/lib/python2.7/site-packages/salt/utils/http.py", line 515, in query
    **req_kwargs
  File "/usr/lib64/python2.7/site-packages/tornado/httpclient.py", line 102, in fetch
    self._async_client.fetch, request, **kwargs))
  File "/usr/lib64/python2.7/site-packages/tornado/ioloop.py", line 444, in run_sync
    return future_cell[0].result()
  File "/usr/lib64/python2.7/site-packages/tornado/concurrent.py", line 214, in result
  File "<string>", line 3, in raise_exc_info
ValueError: Body must be None for GET request
>>>
```
The following code solves the bug - and also permits a payload to be sent with `GET`:

```
    if method != 'GET' or data is not None:
        if data is None:
            data = {}
        data = salt.utils.json.dumps(data)
```
I'll commit that if you prefer.

rallytime: I just want to make sure we don't have a regression/change in expected behavior here for someone. The default for `method` is `GET` in the function signature, so it seems like we would have run into this error before. At the very least the docs should probably be updated.

@garethgreenaway It looks like you've been involved in this module in the past. I'd like to have you look this over as well.

garethgreenaway (approved):

salt/modules/consul.py
rallytime: @campbellmc Can you update the docs about this change as I mentioned above? After that is done, we can get this merged in.

rallytime:
Hi @campbellmc - I'd like to get this merged in. Can you add some docs as I mentioned above and then we can get this change in?

salt/modules/consul.py
campbellmc: Hi there!
Sorry for the delay.  I added a comment in docstring for _query.  I hope that's what you meant for me to do.
Campbell

rallytime (approved):

Issue #43385, opened 2017-09-07
salt-ssh minion logfile location is hard-coded


duk3luk3:
### Description of Issue/Question

Salt-SSH hardcodes the location of the minion logfile by overriding any options set in the config.

This means there is no way to get the minion logs when using --wipe, since the logs will be inside the wiped deployment directory.

The ideal solution here would be to transfer the minion log back to the master after execution, so that '--wipe' can actually be trusted not to leave data on the minion, but in the absence of that option, allowing that the log is stored in a custom location on the minion is necessary.

### Setup

Any

### Steps to Reproduce Issue

This config in the master config file should redirect the config:

```
ssh_minion_opts:
  log_file: /var/log/salt/minion
```

However, it is always overwritten by the salt-ssh initialization code (line 890 in the snippet below).

https://github.com/saltstack/salt/blob/1ddf818b4102aba43cb4775fcb3db9f0b6bc2fb1/salt/client/ssh/__init__.py#L878-L892

### Versions Report

```
$ salt-ssh --versions-report                                                                                                                                                                                                                   
Salt Version:
           Salt: 2017.7.0-539-g1ddf818
  
Dependency Versions:
           cffi: Not Installed
       cherrypy: Not Installed
       dateutil: 2.6.1
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.9.6
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.4.8
   mysql-python: Not Installed
      pycparser: Not Installed
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: Not Installed
         Python: 2.7.6 (default, Jun 22 2015, 17:58:13)
   python-gnupg: Not Installed
         PyYAML: 3.12
          PyZMQ: 16.0.2
           RAET: Not Installed
          smmap: Not Installed
        timelib: Not Installed
        Tornado: 4.5.1
            ZMQ: 4.1.6
 
System Versions:
           dist: Ubuntu 14.04 trusty
         locale: UTF-8
        machine: x86_64
        release: 3.13.0-117-generic
         system: Linux
        version: Ubuntu 14.04 trusty
```

garethgreenaway:
@duk3luk3 Thanks for the report.  Approving this for a future feature addition.

[...]

PR #45972, opened 2018-02-11
move log_file option to changeable defaults

mcalmer:
### What does this PR do?

Using salt-ssh with --wipe always remove the logfile. This patch makes it possible to overwrite
the log_file option and redirect it out of the thin dir which gets removed.

https://github.com/saltstack/salt/issues/43385

### Tests written?

No

### Commits signed with GPG?

No


rallytime (approved):
Thanks @mcalmer - this looks good to me, but I'm going to have @gtmanfred look here as well.

gtmanfred (approved):
LGTM

rallytime:
re-run py

Issue #45976, opened 2018-02-12
6a5e0f9 introduces regression that breaks Vault module for salt masterless


grobinson-blockchain:
### Description of Issue/Question

https://github.com/saltstack/salt/commit/6a5e0f9ac11e7731b759d7dcc873c07f6436ba1b introduces a regression that breaks the Vault module for salt masterless.

Since https://github.com/saltstack/salt/commit/6a5e0f9ac11e7731b759d7dcc873c07f6436ba1b, we get the following error message running `salt-call --local state.apply`:

        [CRITICAL] Rendering SLS 'foo' failed, render error:
        Problem running salt function in Jinja template: [Errno 2] No such file or directory: '/etc/salt/pki/minion/minion.pem'; line 9

        ---
        [...]
            foo:
              - "{{ salt.sdb.get('sdb://my-vault/secret/foo?value') }}"    <======================
        ---

### Setup

We use [Packer](https://packer.io) to generate machine images, with Salt masterless for configuration management. This change prevents us from being able to build new images because it breaks the Vault module for Salt masterless.

PR #45977, opened 2018-02-12
Fix regression in 6a5e0f

ghost:
### What does this PR do?

Fixes regression introduced in https://github.com/saltstack/salt/commit/6a5e0f9ac11e7731b759d7dcc873c07f6436ba1b.

https://github.com/saltstack/salt/issues/45976

### DIFF
--- a/salt/utils/vault.py
+++ b/salt/utils/vault.py
@@ -96,7 +96,7 @@ def _get_vault_connection():
     Get the connection details for calling Vault, from local configuration if
     it exists, or from the master otherwise
     '''
-    if 'vault' in __opts__ and __opts__.get('__role', 'minion') == 'master':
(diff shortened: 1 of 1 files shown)


lorengordon:
The change in 6a5e0f9 seems pretty intentional. Might want @gtmanfred to weigh in. If the test needs to be updated specifically to account for masterless mode, this will work:

```
masterless = __opts__['__role'] == 'minion' and __opts__['file_client'] == 'local'
```

gtmanfred (changes requested):
This should not be reverted, because then it will break vault/sdb in the master config and the minion config.

Instead you should probably change the _use_local_config to check all the `file_client: local` `local: true` or `master_type: disable` options.

gtmanfred:
I have addressed this in #45981

PR #45979, opened 2018-02-12
Unicode/logger fix + bugfix: salt/utils/parsers

isbm:
### What does this PR do?

Bugfix/refactoring

- Fixes logging with unicode
- Fixes PID file deletion as non-root (Master)
- Fixes unit tests for this topic

### Tests written?

Yes

### DIFF
--- a/salt/utils/parsers.py
+++ b/salt/utils/parsers.py
@@ -975,11 +975,11 @@ class DaemonMixIn(six.with_metaclass(MixInMeta, object)):
                 try:
                     os.unlink(self.config['pidfile'])
                 except OSError as err:
-                    logging.getLogger(__name__).info(
-                        'PIDfile could not be deleted: {0}'.format(
-                            self.config['pidfile']
-                        )
-                    )
+                    # Log error only when running salt-master as a root user.
+                    # Otherwise this can be ignored, since salt-master is able to
+                    # overwrite the PIDfile on the next start.
+                    if not os.getuid():
+                        log.info('PIDfile could not be deleted: %s', six.text_type(self.config['pidfile']))
 
     def set_pidfile(self):
         from salt.utils.process import set_pidfile
--- a/salt/utils/parsers.py
+++ b/salt/utils/parsers.py
@@ -261,12 +261,10 @@ class OptionParser(optparse.OptionParser, object):
             try:
                 mixin_before_exit_func(self)
             except Exception as err:  # pylint: disable=broad-except
-                logger = logging.getLogger(__name__)
                 logger.exception(err)
-                logger.error(
-                    'Error while processing %s: %s',
-                    mixin_before_exit_func, traceback.format_exc(err)
-                )
+                logger.error('Error while processing %s: %s',
+                             six.text_type(mixin_before_exit_func),
+                             traceback.format_exc(err))
         if self._setup_mp_logging_listener_ is True:
             # Stop logging through the queue
             log.shutdown_multiprocessing_logging()
--- a/salt/utils/parsers.py
+++ b/salt/utils/parsers.py
@@ -2815,13 +2815,13 @@ class SaltRunOptionParser(six.with_metaclass(OptionParserMeta,
             '--async',
             default=False,
             action='store_true',
-            help=('Start the runner operation and immediately return control.')
+            help='Start the runner operation and immediately return control.'
         )
         self.add_option(
             '--skip-grains',
             default=False,
             action='store_true',
-            help=('Do not load grains.')
+            help='Do not load grains.'
         )
         group = self.output_options_group = optparse.OptionGroup(
             self, 'Output Options', 'Configure your preferred output format.'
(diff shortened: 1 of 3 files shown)


isbm:
@rallytime @terminalmage please take a look.

salt/utils/parsers.py
```diff
@@ -1044,12 +1037,13 @@ def start(self):
         self._install_signal_handlers()
 
     def _handle_signals(self, signum, sigframe):  # pylint: disable=unused-argument
-        msg = self.__class__.__name__
+        msg = [six.text_type(self.__class__.__name__)]
         if signum == signal.SIGINT:
-            msg += ' received a SIGINT.'
+            msg.append('received a SIGINT.')
```
terminalmage: The string types do not have an `append` function.

salt/utils/parsers.py
```diff
@@ -1044,12 +1037,13 @@ def start(self):
...
         elif signum == signal.SIGTERM:
-            msg += ' received a SIGTERM.'
-        logging.getLogger(__name__).warning('%s Exiting.', msg)
+            msg.append('received a SIGTERM.')
```
terminalmage: Same thing here, this line needs to be reverted.

salt/utils/parsers.py
```diff
@@ -2739,31 +2733,31 @@ def _find_raet_minion(self, opts):
...
-                    not os.path.isfile(ha) and
-                    not os.path.isdir(ha)):  # minion manor yard
-                return True
+            try:
+                from raet.lane.yarding import Yard  # pylint: disable=3rd-party-module-not-gated
+                ha, dirpath = Yard.computeHa(dirpath, lanename, yardname)  # pylint: disable=invalid-name
+                if os.path.exists(ha) and not os.path.isfile(ha) and not os.path.isdir(ha):  # minion manor yard
+                    return True
+            except ImportError as ex:
+                logger.error("Error while importing Yard: %s", six.text_type(ex))
```
terminalmage: No need for `six.text_type(ex)`, this can just be `ex`. String replacement will automagically run `ex.__unicode__()` to get a string, because the string with the replacement placeholders is a unicode. For example:

```python
>>> u'The error is: %s' % TypeError(u'питон')
u'The error is: \u043f\u0438\u0442\u043e\u043d'
```

terminalmage (changes requested):

salt/utils/parsers.py
isbm: It is earlier an array... :wink:

salt/utils/parsers.py
terminalmage: I missed the brackets around msg when it was being initialized. Still, allocating a list for this is slower:

```python
>>> timeit.Timer('''
... x = ['foo']
... x.append('bar')
... x = ' '.join(x)''').repeat()
[0.22678208351135254, 0.21661591529846191, 0.20710206031799316]
>>> timeit.Timer('''
... x = 'foo'
... x += " bar"''').repeat()
[0.08382701873779297, 0.054234981536865234, 0.05179405212402344]
```

terminalmage (approved):

rallytime:
@isbm Looks like we have a related test failure here: https://jenkins.saltstack.com/job/PR/job/salt-pr-linode-ubuntu16-py3/6776/

Can you take a look?

isbm:
@rallytime Seems like your test suite on Ubuntu 16 is clashing with the Mock 2.0.0 from Pip, saying in logs it is installed, but actually using just a standard mock library from Python 3.5 where this method is actually absent. This method is [new to Python 3.6](https://docs.python.org/release/3.6.0/library/unittest.mock.html#unittest.mock.Mock.assert_called).

Seems we found a bug in test setup!

terminalmage:
Via https://github.com/saltstack/salt-jenkins/pull/840, https://github.com/saltstack/salt-jenkins/pull/841, and the commit just merged into this PR, we should now be using mock 2.0.0 from PyPI for everything but `mock_open`, which we have backported.

isbm:
@terminalmage @rallytime seems working now. I am wondering why Cent7/U14 still fails, but that's seems unrelated to here.

Thanks!

rallytime:
@isbm Ah, yeah the Ubuntu 14 errors are unrelated. But the cp test failures on the PY3 build are related to the concerns @gtmanfred brought up in https://github.com/saltstack/salt-jenkins/pull/840

@terminalmage Can you take another look into the changes you made in our custom `mock_open`?

gtmanfred:
Actually, those cp changes are unrelated.  mock_open was affecting the cp.push test

I have been seeing this same error with repo.saltstack.com, the 599 error, since yesterday when we switched the repo to be behind a DO load balancer instead of haproxy.

terminalmage:
I didn't change our backport, I just changed the bits that import directly from unittest.mock on Python 3. I'll look at the test failures.

rallytime:
Ah, OK. Thanks for the explanation everyone.

PR #45991, opened 2018-02-12
yumpkg: Fix a couple issues with _get_extra_opts

terminalmage:
`_get_extra_opts()` and `_get_branch_option()` were unnecessarily quoting the value, causing it to be interpreted as a literal quote by `subprocess.Popen()`.

Also, because there were separate helpers for repo options, disableexcludes, branch options, and extra options, and specifically because `_get_extra_opts()` parses *all* kwargs, any of the options from the other helper funcs would end up being added to the command line twice if `_get_extra_opts()` was used.

This commit consolidates all of the kwarg inspection and CLI opts construction to a single helper function. It also adds unit tests to make sure that we are formatting our commands properly.

Additionally, it makes a minor fix in `refresh_db()` which was not accounted for when we changed the osmajorrelease grain to an integer in 2017.7.0.

### DIFF
--- a/salt/modules/yumpkg.py
+++ b/salt/modules/yumpkg.py
@@ -204,25 +204,29 @@ def _check_versionlock():
         )
 
 
-def _get_repo_options(**kwargs):
+def _get_options(**kwargs):
     '''
-    Returns a list of '--enablerepo' and '--disablerepo' options to be used
-    in the yum command, based on the kwargs.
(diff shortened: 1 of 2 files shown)


garethgreenaway (approved):

rallytime:
@terminalmage This is causing some test failures. Can you take a look?

terminalmage:
@rallytime fixed these last night, passing now.

rallytime (approved):

PR #45994, opened 2018-02-13
Fix hosted zone Comment updates & quote TXT entries correctly

nullify005:
### What does this PR do?

1. Protects continual updates to the Comment field which isn't '' but rather is None when attempting to compare the local changes against it
2. Encapsulates ResourceRecords in quotes if they happen to be part of a TXT Type. The API requires that the entries are surrounded by quotes when being submitted in a ChangeBatch.

A wiser person might be able to figure out how to get the jinja templating engine to pass through the quotes are they are written in the state file, but I just couldn't figure it out.

Taking a clue from Ansible I'm just quoting ResourceRecord entries if the Type happens to be a TXT, when this happens everything is magical

* https://github.com/saltstack/salt/issues/45995
* https://github.com/saltstack/salt/issues/45996

### Previous Behavior

1. The Comment field in the Zone is updated even though you are leaving it blank
2. TXT records cannot be created at all

### New Behavior

1. Since the upstream value for Comment is actually None not '' when empty, we initialise the local variable with a None if you leave it out so that the comparison works correctly and the field isn't updated.
2. TXT records can be created, yay!

### Tests written?

No

### Commits signed with GPG?

No

### DIFF
--- a/salt/states/boto3_route53.py
+++ b/salt/states/boto3_route53.py
@@ -138,7 +138,7 @@ def _from_aws_encoding(string):  # XXX TODO
 
 
 def hosted_zone_present(name, Name=None, PrivateZone=False,
-                        CallerReference=None, Comment='', VPCs=None,
+                        CallerReference=None, Comment=None, VPCs=None,
                         region=None, key=None, keyid=None, profile=None):
     '''
     Ensure a hosted zone exists with the given attributes.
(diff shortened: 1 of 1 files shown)


Ch3LL:
ping @rallytime can this get back ported to 2017.7 when merged

rallytime:
@Ch3LL Totally. Thanks!

rallytime:
@nullify005 This has a lint error: https://jenkins.saltstack.com/job/PR/job/salt-pr-lint-n/19237/violations/file/salt/states/boto3_route53.py/

You should be able to use `format` instead of `%s` there.

rallytime (approved):

Issue #45790, opened 2018-01-31
Test with Tornado 5.0b1


bdarnell:
Tornado 5.0 is currently in beta, and I'm filing issues with prominent
users of Tornado to give them a chance to test it before release and
provide feedback on the changes and make adjustments as necessary. You
can install it with `pip install --pre tornado==5.0b1`. I expect to
release the final version of Tornado 5.0 in early March, about a month
away.

Full release notes are
[here](http://www.tornadoweb.org/en/latest/releases/v5.0.0.html). The
major change is that the `IOLoop` is now always backed by `asyncio` on
Python 3, which leads to various subtle changes at runtime but is
mostly compatible with previous versions. A more superficial but
source-incompatible change is that the `io_loop` argument to many
functions in Tornado has been removed and `IOLoop.current()` must be
used instead.

If you've already tested with Tornado 5.0b1 (or temporarily pinned
your dependency to 4.x so you can postpone dealing with this), feel
free to close this issue.

garethgreenaway:
@bdarnell Thanks for the report.  @gtmanfred is on the case.

gtmanfred:
@bdarnell is there a reason that you are just dropping the io_loop all together?  This will break all older versions of salt.

Would it be possible that yall could just swallow the value?  I tested with this, and it seems to work fine for us, the only problem is the error that there is no allowed io_loop variable anymore, which will break older versions.

Also, with not being able to pass in an io_loop, is there any possiblity to run an io_loop inside of an io_loop?

https://github.com/saltstack/salt/blob/2017.7/salt/engines/webhook.py#L84

We use multiprocessing to spin out these extra tasks called "engines".  This one in particular creates a new io_loop for a tiny web/api server, that doesn't really belong on the regular salt io_loop.

Without being able to pass in an io_loop, will it need to be rewritten to handle using IOLoop.current() or is there another way to make a seperate loop?

Thanks,
Daniel

bdarnell:
Accepting the `io_loop` argument but ignoring it would be very confusing if you were passing any other IOLoop than the current one. And if you were passing `IOLoop.current()`, you could simply remove the argument and be compatible with both old and new versions of Tornado.

Being able to run an IOLoop inside an IOLoop is unfortunately something that is lost with the move to asyncio, since asyncio doesn't allow this. If this is something you need, then I'm afraid there are only hacky solutions. The most straightforward is probably to spawn a second thread and create/run the IOLoop there.

[...]

PR #46002, opened 2018-02-13
Pyzmq 17.0.0 proper handling

isbm:
### What does this PR do?

ZeroMQ 4.1.6 and PyZMQ 17.0.0 no longer need extra-integration with Tornado. If used, deprecation warning appears all around the place.

- Removes code duplication for `zmq` import handling around Master/Minion and other places.
- Adds utility that handles all the discrepancies at one place. Other modules should just reuse it.
- Cleans up unnecessary checks and forgotten unused imports

### Tests written?

No

### DIFF
--- a/salt/utils/zeromq.py
+++ b/salt/utils/zeromq.py
@@ -26,7 +26,7 @@ try:
         if ZMQ_VERSION_INFO[0] > 16:  # 17.0.x+ deprecates zmq's ioloops
             ZMQDefaultLoop = tornado.ioloop.IOLoop
 except Exception as ex:
-    log.exception(ex)
+    log.exception('Error while getting PyZMQ library version')
 
 if ZMQDefaultLoop is None:
     try:
--- a/salt/utils/zeromq.py
+++ b/salt/utils/zeromq.py
@@ -38,6 +38,17 @@ if ZMQDefaultLoop is None:
     except ImportError:
         ZMQDefaultLoop = tornado.ioloop.IOLoop
 
+# Build version info
+if zmq:
+    if hasattr(zmq, 'zmq_version_info'):
+        zmq_version_info = zmq.zmq_version_info()
+    else:
+        # PyZMQ <= 2.1.9 does not have zmq_version_info, fall back to
+        # using zmq.zmq_version() and build a version info tuple.
+        zmq_version_info = tuple([int(v_el) for v_el in zmq.zmq_version().split('.')])
+else:
+    zmq_version_info = (-1, -1, -1)
+
 
 def install_zmq():
     '''
--- a/salt/utils/zeromq.py
+++ b/salt/utils/zeromq.py
@@ -27,7 +27,7 @@ try:
         LIBZMQ_VERSION_INFO = tuple([int(v_el) for v_el in zmq.zmq_version().split('.')])
         if ZMQ_VERSION_INFO[0] > 16:  # 17.0.x+ deprecates zmq's ioloops
             ZMQDefaultLoop = tornado.ioloop.IOLoop
-except Exception as ex:
+except Exception:
     log.exception('Error while getting LibZMQ/PyZMQ library version')
 
 if ZMQDefaultLoop is None:
--- a/salt/utils/zeromq.py
+++ b/salt/utils/zeromq.py
@@ -19,10 +19,12 @@ except ImportError:
 
 ZMQDefaultLoop = None
 ZMQ_VERSION_INFO = (-1, -1, -1)
+LIBZMQ_VERSION_INFO = (-1, -1, -1)
 
 try:
     if zmq:
-        ZMQ_VERSION_INFO = [int(v_el) for v_el in zmq.__version__.split('.')]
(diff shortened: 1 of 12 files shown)


gtmanfred:
Just fyi, tornado 5.0b removes being able to pass the IOLoop in.

https://github.com/saltstack/salt/issues/45790

isbm:
@gtmanfred yeah, but we are running on various tornado versions and zmq. So this combination is actually messy. I tried first step: collect all that stuff from around everywhere, place it on one side and use it. Sure we can add other exceptions there (utils) for newest tornado.

gtmanfred:
:+1: makes sense, just wanted to make a note since we are trying to deal with the tornado change as well.

isbm:
@gtmanfred then add those "dealings" in `salt.utils.zeromq` in a future and so it is valid everywhere. :wink:

salt/utils/zeromq.py
```diff
@@ -5,21 +5,65 @@
...
+ZMQDefaultLoop = None
+ZMQ_VERSION_INFO = (-1, -1, -1)
+
+try:
+    if zmq:
+        ZMQ_VERSION_INFO = [int(v_el) for v_el in zmq.__version__.split('.')]
+        if ZMQ_VERSION_INFO[0] > 16:  # 17.0.x+ deprecates zmq's ioloops
+            ZMQDefaultLoop = tornado.ioloop.IOLoop
+except Exception as ex:
+    log.exception(ex)
```
terminalmage: `log.exception()` will already log the exception message itself (including the traceback). Therefore, it's not necessary to capture the exception using `as ex`, and the log message should contain a descriptive message telling what led to the exception being raised. For example:

```python
log.exception('Error encountered doing XXXXX')
```

terminalmage (changes requested):
This looks good, just one minor nitpick.

salt/utils/zeromq.py
isbm: Ahh, right. Thanks! Fixd.

rallytime:
re-run py3

rallytime:
@isbm The test failures for the Ubuntu 16 builds look suspicious to me. I am not seeing those failures on the HEAD of the branch.

I am also seeing many of these stacktraces:
```
15:37:53 AttributeError: module 'zmq' has no attribute 'eventloop'
15:37:53 Traceback (most recent call last):
15:37:53   File "/tmp/salt-tests-tmpdir/scripts/salt-call", line 5, in <module>
15:37:53     salt_call()
15:37:53   File "/testing/salt/scripts.py", line 395, in salt_call
15:37:53     client.run()
15:37:53   File "/testing/salt/cli/call.py", line 47, in run
15:37:53     caller = salt.cli.caller.Caller.factory(self.config)
15:37:53   File "/testing/salt/cli/caller.py", line 79, in factory
15:37:53     return ZeroMQCaller(opts, **kwargs)
15:37:53   File "/testing/salt/cli/caller.py", line 291, in __init__
15:37:53     super(ZeroMQCaller, self).__init__(opts)
15:37:53   File "/testing/salt/cli/caller.py", line 102, in __init__
15:37:53     self.minion = salt.minion.SMinion(opts)
15:37:53   File "/testing/salt/minion.py", line 785, in __init__
15:37:53     io_loop = zmq.eventloop.ioloop.ZMQIOLoop()
```
Can you look into that?

isbm:
@rallytime yep, forgot to delete some more code. :wink:

salt/utils/zeromq.py
```diff
@@ -5,21 +5,56 @@
...
+ZMQ_VERSION_INFO = (-1, -1, -1)
+LIBZMQ_VERSION_INFO = (-1, -1, -1)
+
+try:
+    if zmq:
+        ZMQ_VERSION_INFO = tuple([int(v_el) for v_el in zmq.__version__.split('.')])
+        LIBZMQ_VERSION_INFO = tuple([int(v_el) for v_el in zmq.zmq_version().split('.')])
+        if ZMQ_VERSION_INFO[0] > 16:  # 17.0.x+ deprecates zmq's ioloops
+            ZMQDefaultLoop = tornado.ioloop.IOLoop
+except Exception as ex:
```
terminalmage: The `as ex` is still unnecessary here, since it won't be used. You don't need to assign the exception to a variable for log.exception to have access to the traceback. A simple `except Exception:` here is all that is needed.

terminalmage (changes requested):

salt/utils/zeromq.py
isbm: Fixd.

terminalmage (approved):

terminalmage (approved):

DmitryKuzmenko (approved):
Awesome!

rallytime (approved):
Well done - this will be great to have this code in one place.

cachedout (approved):

rallytime:
@isbm We'd like to get this fix into the 2017.7 branch as well, but this is (expectedly) not back-porting cleanly. Would you be so kind as to make this fix against the 2017.7 branch, too?

PR #46006, opened 2018-02-13
Remove obsolete unicode handling in pkg.info_installed

dincamihai:
### What does this PR do?

Removes obsolete unicode encode/decode handling

### Tests written?

No

### Commits signed with GPG?

No

### DIFF
--- a/salt/modules/zypper.py
+++ b/salt/modules/zypper.py
@@ -482,15 +482,6 @@ def info_installed(*names, **kwargs):
         t_nfo = dict()
         # Translate dpkg-specific keys to a common structure
         for key, value in six.iteritems(pkg_nfo):
-            if isinstance(value, six.string_types):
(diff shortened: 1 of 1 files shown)


dincamihai:
It seems that this might break something in zypper module. I have to investigate more.

rallytime:
@dincamihai Since you're still working on this PR, can I move the base branch here to `oxygen` instead of `oxygen.rc1`? We're getting close to moving away from the codenamed branch to a versioned branch and I don't consider this PR a blocker for the RC.

dincamihai:
@rallytime sorry for the late reply. i've changed the base and i have removed the WIP.
the change seem to be ok according to our tests.

rallytime:
@dincamihai Ok, thanks. Let's see how this run goes. :)

terminalmage (approved):

rallytime (approved):

PR #46011, opened 2018-02-13
cmdmod.py: runas workaround for platforms that don't set a USER env var

terminalmage:
Solaris doesn't set a USER env var in its default environment, that is if you run `su - someuser -c env` you get a HOME, PWD, LOGNAME, etc. env var, but not a USER. This commit makes sure that the USER env var is set to the runas user.


terminalmage:
@rallytime Let me know once this commit merges forward from 2017.7 --> oxygen, at that point I'll need to fix the run_env so we're not using unicode.

dmurphy18 (approved):
Looks good, same change could have been made for env_runas too but it works

dmurphy18:
Looks good

dmurphy18 (approved):

rallytime (approved):

Issue #44828, opened 2017-12-04
parallel=True invalidates duration values and as a result Total run time.


junster1:
### Description of Issue/Question
The duration of a parallel thread reflects the time it takes to spawn a thread, rather than the time it takes to execute the state.  It looks like we need to add the duration code to the spawned thread and have it return that data and if parallel=false, use the current logic.

It might also make sense to add a total duration time for the entire run.  aka something like this..

```
Summary for local
------------
Succeeded: 3 (changed=3)
Failed:    0
------------
Total states run:     3
Total state run time:   17.400 s
Total executing time: 12.400 s
```

### Setup

```
not_parallel:
  cmd.run:
    - name: sleep 2

my_parallel_test:
  cmd.run:
    - name: sleep 10
    - parallel: True
    - require: 
      - cmd: not_parallel

my_parallel_test2:
  cmd.run:
    - name: sleep 5
    - parallel: True
```

### Steps to Reproduce Issue

```
# time salt-call state.apply parallel -l warning 
local:
----------
          ID: not_parallel
    Function: cmd.run
        Name: sleep 2
      Result: True
     Comment: Command "sleep 2" run
     Started: 17:11:06.570068
    Duration: 2004.703 ms
     Changes:   
              ----------
              pid:
                  55756
              retcode:
                  0
              stderr:
              stdout:
----------
          ID: my_parallel_test
    Function: cmd.run
        Name: sleep 10
      Result: True
     Comment: Command "sleep 10" run
     Started: 17:11:08.575105
    Duration: 2.904 ms
     Changes:   
              ----------
              pid:
                  55759
              retcode:
                  0
              stderr:
              stdout:
----------
          ID: my_parallel_test2
    Function: cmd.run
        Name: sleep 5
      Result: True
     Comment: Command "sleep 5" run
     Started: 17:11:08.578268
    Duration: 1.541 ms
     Changes:   
              ----------
              pid:
                  55760
              retcode:
                  0
              stderr:
              stdout:

Summary for local
------------
Succeeded: 3 (changed=3)
Failed:    0
------------
Total states run:     3
Total run time:   2.009 s

real	0m13.163s
user	0m0.281s
sys	0m0.063s
```

### Versions Report
Salt Version:
           Salt: 2017.7.2
 
Dependency Versions:
           cffi: Not Installed
       cherrypy: Not Installed
       dateutil: 1.5
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
[...]

Issue #43668, opened 2017-09-21
salt-runner: RuntimeError: maximum recursion depth exceeded


schasi:
### Description of Issue/Question
When trying to use "parallel: True" with salt.runner and salt.state, I get a "RuntimeError: maximum recursion depth exceeded" error.

### Setup
orch.sls:
```
test-vm:
  salt.runner:
    - name: cloud.profile
    - prof: FreeBSD-2C-2G
    - provider: ovirt
    - instances:
      - Prod-Test-1.ocn.sh
    - opts:
        minion:
          master: Prod-Saltmaster-1.ocn.sh
    - parallel: True

test-setup:
  salt.state:
    - tgt: "Prod-Test-1.ocn.sh"
    - highstate: True
    - require:
      - salt: test-vm
```

### Steps to Reproduce Issue

Run the following command:
`salt-run -l debug state.orch orch.test saltenv=prod`

The following exception occurs:
```
[DEBUG   ] Runner return: Exception occurred in runner state.orch: Traceback (most recent call last):
  File "/usr/local/lib/python2.7/site-packages/salt/client/mixins.py", line 392, in _low
    data['return'] = self.functions[fun](*args, **kwargs)
  File "/usr/local/lib/python2.7/site-packages/salt/runners/state.py", line 82, in orchestrate
    orchestration_jid=orchestration_jid)
  File "/usr/local/lib/python2.7/site-packages/salt/modules/state.py", line 1109, in sls
    ret = st_.state.call_high(high_, orchestration_jid)
  File "/usr/local/lib/python2.7/site-packages/salt/state.py", line 2515, in call_high
    ret = self.call_chunks(chunks)
  File "/usr/local/lib/python2.7/site-packages/salt/state.py", line 2008, in call_chunks
    running = self.call_chunk(low, running, chunks)
  File "/usr/local/lib/python2.7/site-packages/salt/state.py", line 2245, in call_chunk
    status, reqs = self.check_requisite(low, running, chunks, pre=True)
  File "/usr/local/lib/python2.7/site-packages/salt/state.py", line 2148, in check_requisite
    return self.check_requisite(low, running, chunks, pre)
  File "/usr/local/lib/python2.7/site-packages/salt/state.py", line 2148, in check_requisite
    return self.check_requisite(low, running, chunks, pre)
...
  File "/usr/local/lib/python2.7/site-packages/salt/state.py", line 2148, in check_requisite
    return self.check_requisite(low, running, chunks, pre)
[...]

Issue #39832, opened 2017-03-04
Parallel mode crashes with "list index out of range"


ninja-:
### Description of Issue/Question
This is my first take at using the new parallel feature and it doesn't quite work in my setup.

SLS file
```
some_task:
  cloudflare.manage_zone_records:
    - zone: {{ pillar["xxx"]["dns"]["xxx.xyz"]|yaml }}
#    - parallel: True
```

```
    Function: cloudflare.manage_zone_records
      Result: False
     Comment: An exception occurred in this state: Traceback (most recent call last):
                File "/usr/lib/python3.6/site-packages/salt/state.py", line 1817, in call
                  ret = self.call_parallel(cdata, low)
                File "/usr/lib/python3.6/site-packages/salt/state.py", line 1699, in call_parallel
                  ret = {'name': cdata['args'][0],
              IndexError: list index out of range
     Started: 19:38:35.127274
    Duration: 3.431 ms
```

Any ideas?

@thatch45 would you have an idea as an author of this feature.

### Setup
(Please provide relevant configs and/or SLS files (Be sure to remove sensitive info).)
Latest Salt Develop, python3
### Steps to Reproduce Issue
(Include debug logs if possible and relevant.)

### Versions Report
(Provided by running `salt --versions-report`. Please also mention any differences in master/minion versions.)

ninja-:
This looks like a logical failure because the cdata[args][0] fallback should be duplicated to the second function. @thatch45 ?

https://github.com/saltstack/salt/blob/develop/salt/state.py#L1667

Ch3LL:
@ninja- thanks for the PR on this. Please feel free to close this issue once the PR has been reviewed/merged or we can.

thatch45:
Thanks for the cleanup here @ninja- ! I don't write features as much anymore because I just don't have the time to work on the cleanup and polish anymore. You assistance here is greatly appreciated!

nullify005:
Hey @ninja- @thatch45 this still looks to be a problem with 2017.7.1

I notice that the referenced PR also was closed due to inactivity, not because it got merged

Is it possible to get this fixed?

The parallel feature in the documentation appears to be entirely broken at the moment

ninja-:
[...]

PR #46023, opened 2018-02-14
add parallel support for orchestrations

mattp-:
### What does this PR do?
originally the parallel global state requisite did not work correctly when invoked under an orch - this fixes that; as well as running any other saltmod state (function, runner, wheel).

I'm opening this now as a review of the patch, as I'm not sure on the full repercussions of my changes. In particular, providing orchestration_jid for __pub_jid. Additionally, tests (submitting this now to see how jenkins responds).

Finally: I noticed there is a parallel_runners feature recently implemented by @smarsching. Wondering about the reasoning for a runner specific parallel versus the state engine feature?

### Previous Behavior
parallel: True did not work in orch.

### New Behavior
parallel: True works in orch; parallel: works on nested orch runners.

### Tests written?
No

### Commits signed with GPG?
No


mattp-:
added commit 30166e1 which should fix #43668

smarsching:
@mattp I think that your patch is very useful and can replace the `salt.parallel_runners` state for some, but not all use cases.

The `parallel = True` attribute has the effect that the state that is decorated with it does not block other states from running. In contrast, the `salt.parallel_runners` state does block other states from running, but executes a number of sub-tasks (in the form of runners) in parallel. This makes it possible to have another state in the orchestration sequence depend on the completion of **all** the parallel tasks in a previous state.

With your patch, defining several states with `parallel = True` should work fine when one only cares about the tasks being run in parallel. However, if one needs some sequential logic in combination with parallel sub-tasks, I do not think it will help. The only way to make such a scenario work would be to put the parallel tasks into a separate file and use `state.orchestrate` with that file. In that case, `state.orchestrate` would act as the synchronization point for all the parallel tasks defined in the other file. However, I think that it is less user-friendly (and also less readable) if one is forced to use a separate state file for the parallel tasks and cannot embed them into the same state file as the sequential tasks.

The reason, why I specifically added the feature for runners, is that this was the place where I needed it. Think of the following example:

You want to setup some clustered database service by first installing the necessary software on a number of nodes and then creating a database on the cluster. The software installation can happen on all nodes in parallel, but you can only configure the the database after the installation has finished on all nodes.

There might be similar cases in the regular state (non-orchestration) logic, but I have not encountered them so far. This might be due to the fact that the logic on a single node is typically state-based while runner / orchestration logic is more action-based (though it internally uses the state engine as well).

I hope this information gives you some insight into why I implemented the `salt.parallel_runners` feature.

mattp-:
@smarsching that was informative, thank you. I think you could maybe achieve similar with requires/requires_in and parallel: True'ing the statemod.runner(s), no? more options are usually better either way :)

austinpapp:
so i think they accomplish the same thing just differently 
https://gist.github.com/austinpapp/e2775271305731327ae335c5c7e0ad17

however, the distinction is really threading v forking @mattp-

regardless, i'd be happy to have either or both. it may prove itself useful down the road.

smarsching:
@mattp- & @austinpapp You are right, I did not think about combining the `parallel` attribute with `require`. This way, one should be able to get the desired behavior. I definitely like that idea. 👍

garethgreenaway (approved):

salt/state.py
```diff
@@ -2326,8 +2326,7 @@ def check_requisite(self, low, running, chunks, pre=False):
                     continue
                 if run_dict[tag].get('proc'):
                     # Run in parallel, first wait for a touch and then recheck
-                    time.sleep(0.01)
-                    return self.check_requisite(low, running, chunks, pre)
+                    run_dict[tag].get('proc').join()
```
cachedout: Eeeeeek. Will this block?

isbm: This will create a zombie until parent quits.

mattp-: @cachedout yes
@isbm according to  https://docs.python.org/2/library/multiprocessing.html#all-platforms join will waitpid and reap, unless im misreading

isbm: Yes. But `.join()` will block everything. And if you did not set it daemonic, it will anyway be joined after the start.

mattp-:
this is now rebased to target 2017.7.5.
@isbm I revisited the proc.join() behavior of this and found a solution that doesn't involve using it along the way. Also included a fix for measuring duration for parallel processes as well as another fix from #39832.
It's been a bit of work but I believe this should fix all outstanding parallel issues, whether that be minion or master side.

mattp-:
this also fixes #44828

mattp-:
accidentally introduced a regression when I revisited the proc.wait behavior ; should be fixed now but I'll wait to see what jenkins has to say.

rallytime:
@mattp- Thanks for fixing up those tests.

@cachedout and @isbm Can you guys review this again?

salt/state.py
```diff
@@ -1879,7 +1891,7 @@ def call(self, low, chunks=None, running=None, retries=1):
             # enough to not raise another KeyError as the name is easily
             # guessable and fallback in all cases to present the real
             # exception to the user
-            if len(cdata['args']) > 0:
+            if 'args' in cdata and cdata['args']:
```
isbm: Still not nice. Just:
```python
if cdata.get('args'):

File: salt/state.py [hunk 3]
```diff
@@ -2176,15 +2188,17 @@ def check_requisite(self, low, running, chunks, pre=False):
                 run_dict = self.pre
             else:
                 run_dict = running
+
+            while True:
+                if self.reconcile_procs(run_dict):
+                    break
+                time.sleep(0.01)
```
isbm: Why this `time.sleep` is doing here?

salt/utils/files.py
```diff
@@ -21,6 +20,12 @@
 # Import 3rd-party libs
 from salt.ext import six
 
+if six.PY2:
+    from urllib import quote
+elif six.PY3:
+    from urllib.parse import quote  # pylint: disable=no-name-in-module
```
isbm: This is wrong. Correct way is this:
```python
from salt.ext.six.moves.urllib.parse import quote
```

isbm (changes requested):

salt/state.py
```diff
@@ -1737,11 +1738,23 @@ def call_parallel(self, cdata, low):
...
+        # populated with what we might have expected, so just be smart
+        # enough to not raise another KeyError as the name is easily
+        # guessable and fallback in all cases to present the real
+        # exception to the user
+        if len(cdata['args']) > 0:
+            name = cdata['args'][0]
+        elif 'name' in cdata['kwargs']:
+            name = cdata['kwargs']['name']
+        else:
+            name = low.get('name', low.get('__id__'))
```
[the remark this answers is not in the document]
mattp-: i'll admit, this code is a replication of what occurs in the serial codepath earlier in the file (https://github.com/bloomberg/salt/blob/a9866c7a031ac6b7fe898c2380131e2a6de82c9f/salt/state.py#L1894). I also don't fully know all ways / types of low data that can be thrown at us here :) 
I'll update in both places

salt/state.py
isbm: Ah, right. The `self.reconcile_procs` still will go through checks and pass through so the CPU still in load. But you still can use `0.0001`, which will down the CPU load, but will do faster loops.

mattp-: It seems we do a similar polling loop in call_chunks() that uses 0.01; since these will be polling longrunning processes I think it makes sense to stick with 0.01 (or perhaps change them both); but i think 10ms latency is sufficient. if someone needed more we should perhaps consider not polling/using a child watcher that tornado provides.

mattp-: updated to 0.0001

cachedout (approved):
This looks right.

PR #46035, opened 2018-02-14
setup.py: Work around bad code in distutils

terminalmage:
While running an sdist, distutils logs when it creates paths. However, if the paths are unicode _and_ contain unicode code points, the log message results in a `UnicodeEncodeError` as distutils attempts to perform string replacement using a str format string and a unicode path.

Paths recently added in the test suite's files to test unicode compatibility have caused this issue to surface.

This fixes the resulting traceback by ensuring that Python 2 only passes utf-8 encoded bytestrings when making the release tree.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -522,7 +522,7 @@ class Sdist(sdist):
             self.run_command('write_salt_ssh_packaging_file')
             self.filelist.files.append(os.path.basename(PACKAGED_FOR_SALT_SSH_FILE))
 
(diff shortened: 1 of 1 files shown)


dmurphy18 (approved):
looks good

rallytime:
@terminalmage Lint: https://jenkins.saltstack.com/job/PR/job/salt-pr-lint-n/19294/violations/file/setup.py/ :)

terminalmage:
I disabled that check inline, WTF.

terminalmage:
@rallytime I pushed a commit to get rid of the lint error.

Issue #46018, opened 2018-02-13
[2017.7.3]: git.latest has regression for windows hosts where TypeError exception is now thrown.


poliva83:
### Description of Issue/Question

https://github.com/saltstack/salt/pull/43769 introduced with 2017.7.3 and has caused regression in the git module for windows hosts. This is because line https://github.com/saltstack/salt/pull/43769/files#diff-59a19c9338fb59c2d90efccd4a0d618fR298 calls `__salt__['file.remove']` passing in `None`. This triggers uncaught TypeError exception to occur here https://github.com/saltstack/salt/blob/v2017.7.3/salt/modules/win_file.py#L1059. You can produce the same error pretty easy on a windows host (see steps to produce).

### Setup

example state file:
```
git_checkout_serverspec_repo:
  git.latest:
    - name: {{ repo }}
    - rev: {{ branch }}
    - branch: {{ branch }}
    - target: {{ settings.serverspec_repo_dir }}
{% if repo.startswith('ssh') %}
    - identity: {{ test_repo_ssh_key_fpath }}
{% elif repo.startswith('https') %}
    - https_user: {{ salt['pillar.get']('salt:serverspec:test_repo:https_user', None) }}
    - https_pass: {{ salt['pillar.get']('salt:serverspec:test_repo:https_pass', None) }}
{% endif %}
    - force_reset: True
    - force_fetch: True
    - force_clone: {{ force_clone }}
```
example state failure:
```
"git_|-git_checkout_serverspec_repo_|-ssh://git@code.td.com/cloud/salt.git_|-latest": {
[...]

PR #45511, opened 2018-01-17
Add correct path to ssh.exe for Windows

twangboy:
### What does this PR do?
Adds the correct path to git's `ssh.exe` for Windows.
Includes the ssh state and execution modules in the Windows installer.

### What issues does this PR fix or reference?
https://github.com/saltstack/salt/issues/34237

### Tests written?
No

### Commits signed with GPG?
Yes

PR #46036, opened 2018-02-14
git.latest: Fix regression with identity file usage

terminalmage:
Unit test included.

Fixes https://github.com/saltstack/salt/issues/46018.


twangboy:
@rallytime You'll probably need to bring this PR in too: https://github.com/saltstack/salt/pull/45511 for Windows

terminalmage:
@twangboy that change you're referring to was already included in this PR, see here: https://github.com/saltstack/salt/pull/46036/files#diff-59a19c9338fb59c2d90efccd4a0d618fR182

I moved the check for the ssh.exe into a helper function to make unit testing on non-Windows platforms possible.

terminalmage:
Oh I see, you're referring to the batch file changes as well?

twangboy:
@terminalmage Yes, but there are also the changes to `build_pkg.bat` that makes sure the `ssh.py` module makes it into the build, otherwise, this will fail on packaged windows.

twangboy:
@rallytime Verified this fix on Windows. Looks good.

twangboy (approved):

rallytime (approved):

PR #46044, opened 2018-02-14
implement minimal fix for #45798

### LINKED ISSUE #45798: LIBCLOUD_VERSION_INFO tuple for rc versions
When running "salt-cloud -p myprofile.conf ourname" from release candidate 2.0.0rc2 
the following tuple fails because it expect each block to be of type int.

```
LIBCLOUD_VERSION_INFO = tuple([
    int(part) for part in libcloud.__version__.replace('-', '.').split('.')[:3]
])
```
.../site-packages/salt/cloud/libcloudfuncs.py [l=29]

Error

> ValueError: invalid literal for int() with base 10: '0rc2'

Installed via yum repo on Centos 7.4.1708

arnoschuring:
### What does this PR do?
Changes the libcloud version checker to not choke on -rc release versions.

This should fix #45798

### Previous Behavior
```
$ salt-cloud -V                                                         
[..]
  File "/usr/lib/python2.7/site-packages/salt/cloud/libcloudfuncs.py", line 28, in <module>
    int(part) for part in libcloud.__version__.replace('-', '.').split('.')[:3]
ValueError: invalid literal for int() with base 10: '0rc2'
```

### New Behavior
```
$ salt-cloud -V
Salt Version:
            Salt: 2018.2.0-2306-g2972204
 
Dependency Versions:
 Apache Libcloud: 2.0.0rc2
[..]
```

### Tests written?
No

### Commits signed with GPG?
No

### DIFF
--- a/salt/cloud/libcloudfuncs.py
+++ b/salt/cloud/libcloudfuncs.py
@@ -25,7 +25,7 @@ try:
     )
     HAS_LIBCLOUD = True
     LIBCLOUD_VERSION_INFO = tuple([
-        int(part) for part in libcloud.__version__.replace('-', '.').replace('rc', '.').split('.')[:3]
+        int(part) for part in re.split(r'[^\d]+', libcloud.__version__)[:3]
     ])
 
 except ImportError:


gtmanfred:
I was pretty certain this was fixed at one point...

gtmanfred:
Wait it does, Which version are you running into this problem with?
```
.replace('rc', '.')
```

This was added to specifically fix this problem, and it is in all current and latest released version.

gtmanfred (changes requested):
This shouldn't need to be done, because we are already replacing the rc with a .

arnoschuring:
ah, never mind, I ran into the issue with 2017.7.2 and forward-ported the check to current develop without verifying if the issue still existed there

gtmanfred:
This should be in 2017.7.2

https://github.com/saltstack/salt/blob/v2017.7.2/salt/cloud/libcloudfuncs.py#L28

PR #46057, opened 2018-02-16
Add unicode normalization support to data/stringutils helpers

terminalmage:
This allows for optional normalization of unicode strings, which will make testing more reliable. In Mac OS, os.listdir() will produce certain unicode glyphs using separate code points for the base character and diacritic mark, while others will use a single code point to represent the same unicode glyph. Adding this optional normalization will make our integration tests more reliable across different platforms.

```python
>>> a = u'\u0065\u0301'
>>> b = u'\u00e9'
>>> print(a)
é
>>> print(b)
é
>>> a == b
False
>>> import salt.utils.stringutils
>>> salt.utils.stringutils.to_unicode(a, normalize=True) == b
True
>>>
```

### DIFF
--- a/salt/utils/data.py
+++ b/salt/utils/data.py
@@ -111,6 +111,9 @@ def decode(data, encoding=None, errors='strict', keep=False,
             data = salt.utils.stringutils.to_unicode(
                 data, encoding, errors, normalize)
         except TypeError:
+            # to_unicode raises a TypeError when input is not a
+            # string/bytestring/bytearray. This is expected and simply means we
+            # are going to leave the value as-is.
             pass
         except UnicodeDecodeError:
             if not keep:
@@ -138,6 +141,9 @@ def decode_dict(data, encoding=None, errors='strict', keep=False,
                 key = salt.utils.stringutils.to_unicode(
                     key, encoding, errors, normalize)
             except TypeError:
+                # to_unicode raises a TypeError when input is not a
+                # string/bytestring/bytearray. This is expected and simply
+                # means we are going to leave the value as-is.
                 pass
             except UnicodeDecodeError:
                 if not keep:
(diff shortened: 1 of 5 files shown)


salt/utils/data.py
```diff
@@ -101,41 +119,46 @@ def decode(data, encoding=None, errors='strict', keep=False,
...
-                else decode_list(key, encoding, errors, keep,
+                else decode_list(key, encoding, errors, keep, normalize,
                                  preserve_dict_class, preserve_tuples)
         else:
             try:
-                key = salt.utils.stringutils.to_unicode(key, encoding, errors)
+                key = salt.utils.stringutils.to_unicode(
+                    key, encoding, errors, normalize)
             except TypeError:
                 pass
```
isbm: I would add here a comment that `pass` is only technical here as it is indicates that `to_unicode` did not processed correct type.

tests/unit/utils/test_stringutils.py
```diff
@@ -91,6 +94,23 @@ def test_to_bytes(self):
...
+        self.assertNotEqual(
+            salt.utils.stringutils.to_unicode(
+                EGGS,
+                encoding='utf=8',
+                normalize=False
+            ),
+            'яйца'
+        )
+
         if six.PY3:
```
isbm: This and below seems no longer relevant.

isbm (changes requested):
LGTM, except few nitpicks.

Ch3LL (approved):
Looks great! Tests that were previously failing are now passing :) Thanks once again @terminalmage :)

rallytime:
re-run py

tests/unit/utils/test_stringutils.py
terminalmage: At least part of it is relevant because it tests python2-specific code, but most of it can be combined in a single section.

terminalmage: I will open a separate PR for that.

rallytime (approved):

PR #46058, opened 2018-02-16
Properly remove package folder from modules

brad-alexander:
We are comparing `new_name` with some string, but then doing a `member.name.replace` instead of a `new_name.replace`.  This reverts a change previously done to the name `new_name = member.name.replace('{0}/'.format(package), '', 1)` so we are still left with `_modules`s going to `/srv/spm/salt/package/_modules`
or `_runners` going to `/var/cache/salt/master/extmods/PACKAGE/runners`

### What does this PR do?
Fixes module file installation via SPM

### Previous Behavior
modules would retain some path elements preventing them from being used properly

### New Behavior
modules files get properly installed at the top level of /srv/spm/salt or /var/cache/salt/master/extmods

### Tests written?

No

### Commits signed with GPG?

No - I will try to do this or just keep this PR around to convert into an issue


rallytime (approved):
This looks fine to me, but @techhat can you swing by here?

PR #46089, opened 2018-02-17
bugfix for auth_safemode

rongzeng54:
### What does this PR do?
bugfix for auth_safemode as the docs(https://docs.saltstack.com/en/develop/ref/configuration/minion.html) said: 
`AUTH_SAFEMODE
New in version 2014.7.0.

Default: False

If authentication fails due to SaltReqTimeoutError during a ping_interval, this setting, when set to True, will cause a sub-minion process to restart.`

No

### Tests written?
No

### Commits signed with GPG?
No


garethgreenaway (approved):

salt/minion.py
```diff
@@ -2624,7 +2624,7 @@ def tune_in(self, start=True):
             def ping_master():
                 try:
                     def ping_timeout_handler(*_):
-                        if not self.opts.get('auth_safemode', True):
+                        if self.opts.get('auth_safemode', True):
```
gtmanfred: Looks like this was implemented incorrectly, but we should not have it switch to True, if it is not specified, it should be `False` in the .get() as well.

gtmanfred (changes requested):

salt/minion.py
rallytime: @rongzeng54 Did you see the feedback here?

gtmanfred (approved):

rallytime (approved):

rallytime:
re-run ubuntu

rallytime:
re-run py3

PR #46098, opened 2018-02-19
Add initial work on gpfs execution and state modules

arif-ali:
### What does this PR do?

This add the ability to deploy gpfs clients on a HPC system. The saltstack formula is [here](https://github.com/ocfplc/salt-formula-gpfs) as a reference

### Tests written?

No, I do my own Ci testing before it goes in. Plus, it's difficult to write tests for the application, as it's a commercial application, that is not easily downloadable

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/modules/gpfs.py
+++ b/salt/modules/gpfs.py
@@ -1,0 +1,133 @@
+# -*- coding: utf-8 -*-
+'''
+Module to provide GPFS compatibility to Salt.
+'''
+from __future__ import absolute_import
+
+# Import python libs
+import logging
+import os.path
+
+log = logging.getLogger(__name__)
+
+
+def __virtual__():
+    '''
+    Verify gpfs is installed.
+    '''
+    return os.path.isfile('/usr/lpp/mmfs/bin/mmgetstate')
+
+
+def cluster_configured(runas=None):
+    '''
+    Return whether the node is configured in a cluster
+
+    CLI Example:
+
+    .. code-block:: bash
+
+        salt '*' gpfs.cluster_configured
+    '''
+    ret = True
+    res = __salt__['cmd.run']('/usr/lpp/mmfs/bin/mmlscluster',
+                              runas=runas, shell='/bin/bash')
+    for line in res.splitlines():
+        if 'node does not belong' in line:
+            ret = False
+
+    return ret
+
+
+def cluster_member(clustername, runas=None):
+    '''
+    Returns wheter the node is joined to the cluster
+
+    CLI Example:
+
+    .. code-block:: bash
+
+        salt '*' gpfs.cluster_joined gpfs.cluster
+    '''
+    ret = False
+    res = __salt__['cmd.run']('/usr/lpp/mmfs/bin/mmlscluster',
+                              runas=runas, shell='/bin/bash')
+    for line in res.splitlines():
+        if 'GPFS cluster name' in line:
+            parts = line.split(':')
+            if len(parts) < 2:
+                continue
+            cluster = parts[1].strip()
+            if cluster == clustername:
+                ret = True
+
+    return ret
+
+
+def join_cluster(master, runas=None):
+    '''
+    Joins a GPFS cluster via the specified master
+
+    CLI Example:
+
+    .. code-block:: bash
+        salt '*' gpfs.join_cluster server1
+    '''
+    ret = False
+    res = __salt__['cmd.run']('/usr/lpp/mmfs/bin/mmsdrrestore -p {0} -R /usr/bin/scp'.format(master),
+                              runas=runas, shell='/bin/bash')
+    for line in res.splitlines():
+        if 'successfully restored' in line:
+            ret = True
+            __salt__['cmd.run']('/usr/lpp/mmfs/bin/mmstartup',
+                                runas=runas, shell='/bin/bash')
+    if not ret:
+        res2 = __salt__['cmd.run']('/usr/bin/ssh {0} /usr/lpp/mmfs/bin/mmaddnode -N {1}'.format(master, __grains__['host']),
+                                   runas=runas, shell='/bin/bash')
+        for line in res2.splitlines():
+            if 'Command successfully completed' in line:
+                ret = True
+                __salt__['cmd.run']('/usr/lpp/mmfs/bin/mmchlicense client --accept -N {0}'.format(__grains__['host']),
+                                    runas=runas, shell='/bin/bash')
+                __salt__['cmd.run']('/usr/lpp/mmfs/bin/mmstartup',
+                                    runas=runas, shell='/bin/bash')
+    return ret
+
+
+def cluster_started(runas=None):
+    '''
+    Checks to see if the GPFS cluster has been started
+
+    CLI Example:
+
+    .. code-block:: bash
+        salt '*' gpfs.cluster_started
+    '''
+    ret = False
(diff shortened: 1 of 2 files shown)


salt/modules/gpfs.py
```diff
@@ -0,0 +1,137 @@
...
+import salt.utils
+
+log = logging.getLogger(__name__)
+
+
+def __virtual__():
+    '''
+    Verify gpfs is installed.
+    '''
+    os.environ["PATH"] += os.pathsep + '/usr/lpp/mmfs/bin'
```
gtmanfred: Is it possible to install this to a different location?

Could it be changed to be a config option?

Thanks,
Daniel

gtmanfred: Also, since this is the same with other services.  It is probably best to add docs that specify how to configure the PATH in an /etc/systemd/system dropin file, or in /etc/default/salt to add this path, instead of hard coding the path and explicitly adding it to the environment.

Because this will add this path to all users, even if they don't have this tool installed.

salt/states/gpfs.py
```diff
@@ -0,0 +1,114 @@
...
+
+log = logging.getLogger(__name__)
+
+
+def __virtual__():
+    '''
+    Only load if GPFS is installed.
+    '''
+    os.environ["PATH"] += os.pathsep + '/usr/lpp/mmfs/bin'
+    return salt.utils.which('mmlscluster') is not None
```
gtmanfred: instead of using the same check as the module, you can just check `if 'gpfs.start_cluster' in __salt__:`

gtmanfred (changes requested):

salt/modules/gpfs.py
arif-ali: This path is static, and cannot be changed, and has been like this for many years for this app. It cannot be customisable either.

I could instead check to see if that file exists, and not add it to PATH.

I am already using the full paths for the commands, so shouldn't make a difference there

gtmanfred: Yeah, i would rather just check if the full path is there, instead of mangling the PATH variable for all of salt just to use which to see if the a command is there.

arif-ali:
Have updated the modules/states as requested

gtmanfred (approved):
LGTM! Thanks!

LGTM

terminalmage (approved):

rallytime:
@terminalmage @gtmanfred and @arif-ali - If these modules/states are not something that we can easily add tests for, then I am wondering if this is a good candidate for the salt-contrib repo instead of  mainline salt. Thoughts?

damon-atkins:
Thanks for the contribution.

__virtual__ = 'ibm_gpfs'
Could you make the paths binaries Linux and AIX compatible.

Consider if any of these function name patterns are worth considering to keep function names consistent:
```
boto3_elasticache.py:def describe_cache_clusters(name=None, conn=None, region=None, key=None,
boto3_elasticache.py:def cache_cluster_exists(name, conn=None, region=None, key=None, keyid=None, profile=None):
boto3_elasticache.py:def create_cache_cluster(name, wait=600, security_groups=None,
boto3_elasticache.py:def modify_cache_cluster(name, wait=600, security_groups=None, region=None,
boto3_elasticache.py:def delete_cache_cluster(name, wait=600, region=None, key=None, keyid=None, profile=None, **args):
boto_apigateway.py:def create_api_deployment(restApiId, stageName, stageDescription='', description='', cacheClusterEnabled=False,
boto_elasticache.py:def create_replication_group(name, primary_cluster_id, replication_group_description,
boto_elasticsearch_domain.py:def create(DomainName, ElasticsearchClusterConfig=None, EBSOptions=None,
boto_elasticsearch_domain.py:def update(DomainName, ElasticsearchClusterConfig=None, EBSOptions=None,
ceph.py:def cluster_quorum(**kwargs):
ceph.py:def cluster_status(**kwargs):
deb_postgres.py:def cluster_create(version,
deb_postgres.py:def cluster_list(verbose=False):
deb_postgres.py:def cluster_exists(version,
deb_postgres.py:def cluster_remove(version,
deb_postgres.py:def _parse_pg_lscluster(output):
elasticsearch.py:def cluster_health(index=None, level='cluster', local=False, hosts=None, profile=None):
elasticsearch.py:def cluster_stats(nodes=None, hosts=None, profile=None):
pcs.py:def auth(nodes, pcsuser='hacluster', pcspasswd='hacluster', extra_args=None):
pcs.py:def cluster_setup(nodes, pcsclustername='pcscluster', extra_args=None):
pcs.py:def cluster_node_add(node, extra_args=None):
rabbitmq.py:def cluster_status(runas=None):
rabbitmq.py:def join_cluster(host, user='rabbit', ram_node=None, runas=None):
riak.py:def cluster_join(username, hostname):
riak.py:def cluster_leave(username, hostname):
riak.py:def cluster_plan():
riak.py:def cluster_commit():
solrcloud.py:def cluster_status(**kwargs):
trafficserver.py:def bounce_cluster():
trafficserver.py:def clear_cluster():
trafficserver.py:def restart_cluster():
trafficserver.py:def zero_cluster():
vsphere.py:def list_clusters(host, username, password, protocol=None, port=None):
vsphere.py:def list_datastore_clusters(host, username, password, protocol=None, port=None):
```

terminalmage:
> `__virtual__` = 'ibm_gpfs' and the name of the files ibm_gpfs.py

@arif-ali Please ignore this. I think @damon-atkins is referring to the `__virtualname__`, which is not required as our loader falls back to using the module name as the loaded module's name.

arif-ali:
@rallytime, what would it mean, if it was in salt-contrib, would it still be shipped as part of the salt package, or would we need to download these modules separately?

I've not really looked at salt-contrib, and didn't really know that existed

terminalmage:
@arif-ali salt-contrib is a separate repo, from which you would need to download and distribute using [saltutil.sync_modules](https://docs.saltstack.com/en/latest/ref/modules/all/salt.modules.saltutil.html#salt.modules.saltutil.sync_modules) and [saltutil.sync_states](https://docs.saltstack.com/en/latest/ref/modules/all/salt.modules.saltutil.html#salt.modules.saltutil.sync_states).

arif-ali:
OK, let me know either way, more than happy to drop this, if it better suites the salt-contrib repo

rallytime:
Since we're not able to write tests for this, then I think [salt-contrib](https://github.com/saltstack/salt-contrib) is the best place for it. Unless others disagree with me.

arif-ali:
OK, I'm happy to move. But it seems we don't have the same set of tests there, i.e. the salt-pr-lint-n. Which will ensure that the code is consistent the main modules?

Would it make sense to add these to the salt-contrib repo?

rallytime:
@arif-ali That's true that we don't have the same set of tests/lint checks over there. I do think it makes sense to add that lint check especially to the salt-contrib repo. @dwoz has recently taken over support of that repo and I know he has plans to standardize everything a bit there.

One of the reasons we can't enable a lint test right now for the repo is we need to fix all of the lint failures on the repo first. If that is something you'd like to help with, we'd certainly appreciate the help as it shouldn't be too hard to start cleaning it up.

rallytime:
As discussed above, this should be moved over to the salt-contrib repo. I know work has begun over there to get the repo up-to-snuff with the linter, but these states and modules can be added there before the project is complete. (Especially since we know these files are passing here.)

As such, I am going to close this PR. Thank you for your work here!

PR #46102, opened 2018-02-19
Allow configuring HTTP connect timeout

mephi42:
### What does this PR do?
Allows configuring HTTP connection timeout for Tornado back-end.

None

### Previous Behavior
HTTP connection timeout was always Tornado's default of 20 seconds.

### New Behavior
If not explicitly configured, HTTP connection timeout remains 20 seconds. Otherwise user-specified value is used.

### Tests written?

No

### Commits signed with GPG?

Yes

### DIFF
--- a/doc/ref/configuration/master.rst
+++ b/doc/ref/configuration/master.rst
@@ -1044,6 +1044,40 @@ cache events are fired when a minion requests a minion data cache refresh.
 
     minion_data_cache_events: True
 
+.. conf_master:: http_connect_timeout
+
+``http_connect_timeout``
+------------------------
+
+.. versionadded:: Fluorine
+
+Default: ``20``
+
+HTTP connection timeout in seconds.
+Applied when fetching files using tornado back-end.
+Should be greater than overall download time.
+
+.. code-block:: yaml
+
+    http_connect_timeout: 20
+
+.. conf_master:: http_request_timeout
+
+``http_request_timeout``
+------------------------
+
+.. versionadded:: 2015.8.0
+
+Default: ``3600``
+
+HTTP request timeout in seconds.
+Applied when fetching files using tornado back-end.
+Should be greater than overall download time.
+
+.. code-block:: yaml
+
+    http_request_timeout: 3600
+
 .. _salt-ssh-configuration:
 
 Salt-SSH Configuration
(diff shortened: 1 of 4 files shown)


DmitryKuzmenko (approved):
LGTM

salt/utils/http.py
```diff
@@ -532,6 +533,7 @@ def query(url,
                 allow_nonstandard_methods=True,
                 streaming_callback=streaming_callback,
                 header_callback=header_callback,
+                connect_timeout=connect_timeout,
```
rallytime: Can you move this to the bottom of the kwargs list? We don't want to cause an API change for people or other Salt functions using `query` who _might_ be calling this with args/kwargs in a certain order.

rallytime (changes requested):
I have one small request, and then I would also love to see docs added to the config docs.

salt/utils/http.py
mephi42: This is the call to fetch() - http://www.tornadoweb.org/en/stable/httpclient.html#tornado.httpclient.HTTPRequest (you need to expand the minimized lines to see this, otherwise this looks like query() parameter list), and there connect_timeout comes before request_timeout.

mephi42:
Added the docs for the new parameter and also for the existing http_request_timeout. I wasn't sure into which section to put them, so I created a new one.

doc/ref/configuration/master.rst
```diff
@@ -5142,6 +5142,46 @@ Sets ZeroMQ TCP keepalive interval. May be used to tune issues with minion disco
...
+
+HTTP Client Settings
+====================
+
+.. conf_master:: http_connect_timeout
+
+``http_connect_timeout``
+------------------------
+
+.. versionadded:: 2018.3.0
```
rallytime: This should be `Fluorine` for the develop branch.

rallytime (changes requested):
I like the new section in the docs - That looks good! I think adding this to the minion config docs would also be good.

salt/utils/http.py
rallytime: Oh! You're right. My apologies.

mephi42:
In minion docs the related proxy_* are in "Minion Primary Configuration", so I put the new items there as well. For that reason I also moved them to "Primary Master Configuration" in master docs.

rallytime (approved):

Issue #44923, opened 2017-12-11
Migrate from Crypto to M2Crypto


Ch3LL:
For oxygen we will be migrating to M2Crypto. This issue will track the work needed to complete this migration.

The-Loeki:
Without knowing the context, M2Crypto is Py2.7 only, right? I thought Salt even ditched it a while back for that reason?

DmitryKuzmenko:
@The-Loeki 
1. M2Crypto developers are working on Py3 support and now it's possible to build it for Py3 from the python3 branch.
2. Crypto support wasn't dropped. Salt will fallback to Crypto if M2Crypto isn't installed.

The-Loeki:
Well that's extremely recent news then ;) tnx 4 the update

damon-atkins:
This seems to be a bad choice, given open stack is now gain ground in Linux and Non-Linux, and uses a different library see https://github.com/saltstack/salt/issues/30287

thatch45:
We are making the crypto backend modular so that we will also be able to build in support for the cryptography library as well.

damon-atkins:
Thanks for the info.

rallytime:
#44962 has been merged. Closing.

PR #46105, opened 2018-02-19
M2Crypto Py3 unicode fix

DmitryKuzmenko:
### What does this PR do?
During checking some issues I've found that Py2 minion can't connect to Py3 master. It was surprise for me because I've checked this behavior on the moment I've done work on M2Crypto.
I've found that the issue is in Unicode handling so I've removed the unneeded tries to convert binary to binary in Py3 from M2Crypto related code.

Related to #44923

### Tests written?
No

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/crypt.py
+++ b/salt/crypt.py
@@ -806,7 +806,7 @@ class AsyncAuth(object):
             pubkey_path = os.path.join(self.opts['pki_dir'], self.mpub)
             pub = get_rsa_pub_key(pubkey_path)
             if HAS_M2:
-                payload['token'] = pub.public_encrypt(six.b(self.token), RSA.pkcs1_oaep_padding)
+                payload['token'] = pub.public_encrypt(self.token, RSA.pkcs1_oaep_padding)
             else:
                 cipher = PKCS1_OAEP.new(pub)
                 payload['token'] = cipher.encrypt(self.token)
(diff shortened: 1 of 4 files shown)


salt/crypt.py
```diff
@@ -806,7 +806,7 @@ def minion_sign_in_payload(self):
             pubkey_path = os.path.join(self.opts['pki_dir'], self.mpub)
             pub = get_rsa_pub_key(pubkey_path)
             if HAS_M2:
-                payload['token'] = pub.public_encrypt(six.b(self.token), RSA.pkcs1_oaep_padding)
+                payload['token'] = pub.public_encrypt(self.token, RSA.pkcs1_oaep_padding)
```
terminalmage: Does `public_encrypt` want a bytestring for the first argument? If so then we should be using `to_bytes` here. It turns out that `six.b()` is a liar and it doesn't produce a bytestring on Python 2:

```python
>>> six.b(u'foo')
u'foo'
```

terminalmage (commented):
This one question applies to all of the places where you have removed `six.b()`.

salt/crypt.py
DmitryKuzmenko: Yes, `public_encrypt` wants a bytestring but `self.token` is already a bytestring so we don't need to convert anything here.

terminalmage: OK, as long as the same holds true for the rest of the instances of six.b, we should be fine.

We should not be using six.b, since it doesn't do what it says it does.

terminalmage (approved):

salt/crypt.py
DmitryKuzmenko: Got it. Thanks @terminalmage !

Issue #46106, opened 2018-02-20
yumpkg.refresh_db hangs


amendlik:
### Description of Issue/Question
`salt.modules.yumpkg.refresh_db` can hang indefinitely if there is an repository signing key that has not yet been accepted on the system. This happens because the yum/dnf utility prompts the user to confirm importing the key, and the Salt Minion will hang because there is no way to respond to the prompt.

### Setup
Create a new YUM repository configuration in /etc/yum.repos.d that has a package signing key and has not yet been accepted on the system.

### Steps to Reproduce Issue
Run `salt-call -l pkg.refresh_db`

The log output will run to this point and hang:
```
[INFO    ] Executing command ['dnf', '--quiet', 'check-update'] in directory '/root'
```
Running the same command by hand will reveal that yum/dnf is waiting for user confirmation.

### Versions Report
```
Salt Version:
           Salt: 2017.7.3
 
Dependency Versions:
           cffi: 1.10.0
       cherrypy: Not Installed
       dateutil: 2.6.0
      docker-py: Not Installed
          gitdb: 2.0.3
      gitpython: 2.1.7
          ioflo: Not Installed
         Jinja2: 2.9.6
        libgit2: 0.26.0
        libnacl: Not Installed
       M2Crypto: 0.25.1
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.5.1
   mysql-python: Not Installed
      pycparser: 2.14
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: 0.26.3
         Python: 2.7.14 (default, Jan 17 2018, 14:28:32)
   python-gnupg: Not Installed
         PyYAML: 3.12
          PyZMQ: 16.0.2
           RAET: Not Installed
          smmap: 2.0.3
        timelib: Not Installed
        Tornado: 4.5.2
            ZMQ: 4.1.6
 
System Versions:
           dist: fedora 27 Twenty Seven
         locale: UTF-8
        machine: x86_64
        release: 4.14.16-300.fc27.x86_64
         system: Linux
        version: Fedora 27 Twenty Seven
```

gtmanfred:
Thanks for the report and the PR to fix it.

Daniel

PR #46107, opened 2018-02-20
Add --assumeyes on YUM/DNF commands

amendlik:
### What does this PR do?
Add `--assumeyes` on some YUM/DNF commands to bypass a prompt for user confirmation.

#46106

### Previous Behavior
`salt.modules.yumpkg.refresh_db` could hang indefinitely if there is an repository signing key that has not yet been accepted on the system.

### New Behavior
YUM/DNF command bypass user confirmation and "assume yes".

### Tests written?
Yes

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/modules/yumpkg.py
+++ b/salt/modules/yumpkg.py
@@ -1025,8 +1025,8 @@ def refresh_db(**kwargs):
 
     options = _get_options(**kwargs)
 
-    clean_cmd = [_yum(), '--quiet', 'clean', 'expire-cache']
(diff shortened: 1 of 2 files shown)


isbm (approved):
LGTM

terminalmage (approved):
I would prefer we standardize on either `-y` or `--assumeyes` for consistency (we use `-y` elsewhere), but this is a minor nitpick that shouldn't interfere with merging.

PR #46110, opened 2018-02-20
IQN check for iSCSII crash on non-root Master

isbm:
### What does this PR do?

Bugfix.

Salt Master will never start if runs on non-root but iSCSII initiator is on the system.

### Tests written?

Yes

### DIFF
--- a/tests/unit/grains/test_core.py
+++ b/tests/unit/grains/test_core.py
@@ -859,5 +859,7 @@ SwapTotal:       4789244 kB'''
         iscsii initiator is not there accessible or is not supported.
         :return:
         '''
-        assert core._linux_iqn() == []
+        with pytest.raises(IOError) as error:
+            assert core._linux_iqn() == []
         core.log.debug.assert_not_called()
+        assert 'No such file or directory' in six.text_type(error)
--- a/tests/unit/grains/test_core.py
+++ b/tests/unit/grains/test_core.py
@@ -815,6 +815,8 @@ SwapTotal:       4789244 kB'''
                             _check_type(key, value, ip4_empty, ip6_empty)
 
     @skipIf(not salt.utils.platform.is_linux(), 'System is not Linux')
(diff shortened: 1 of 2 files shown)


tests/unit/grains/test_core.py
```diff
@@ -809,6 +815,8 @@ def _check_type(key, value, ip4_empty, ip6_empty):
                             _check_type(key, value, ip4_empty, ip6_empty)
 
     @skipIf(not salt.utils.platform.is_linux(), 'System is not Linux')
+    @patch.object(salt.utils, 'is_windows', MagicMock(return_value=False))
```
terminalmage: This should be `salt.utils.platform` instead of `salt.utils`. I know this wasn't your code initially, and that you just moved an existing patch, but can you make this small edit? The code in `salt/grains/core.py` uses `salt.utils.platform.is_windows()`, so the test needs it too. It won't affect this test's accuracy on *NIX, but it will on Windows.

terminalmage (changes requested):
Just one small change, then this is ready for merge.

tests/unit/grains/test_core.py
isbm: @terminalmage fixd

terminalmage (approved):

rallytime:
@isbm Looks like a test failure here: https://jenkins.saltstack.com/job/PR/job/salt-pr-linode-cent7-py3/2419/testReport/junit/unit.grains.test_core/CoreGrainsTestCase/test_linux_iqn_non_root/

isbm:
@rallytime yep, fixed.

isbm:
@rallytime ah, so `IOError` instead of `OSError`. Looking.

isbm:
@rallytime OK, seems I just was catching wrong exception type. Now it looks much happier.

rallytime:
Thanks @isbm!

PR #46129, opened 2018-02-21
Make default pki directory configurable

bdrung:
The files in `/etc/salt/pki` are not configuration files in the sense of the FHS ("local file used to control the operation of a program"). Debian wants to change the default location to `/var/lib/salt/pki` (to properly follow FHS and to allow setting StateDirectory in the salt master systemd configuration).

Therefore introduce a `STATE_DIR` syspaths variable which defaults to `CONFIG_DIR`, but can be individually customized.

fixes #3396
Bug-Debian: https://bugs.debian.org/698898


garethgreenaway:
Seems the `develop` might be a better place for this PR.  The naming is also confusing as we have Salt states and a default location for Salt states, you use of state (I believe) is different?

bdrung:
I proposed this pull requests for 2017.7 because I have applied this patch to the Debian package. Is the `2018.3` branch open for this kind of pull requests?

You are right. The name chosen name is confusing. I changed it to `VARIABLE_STATE_DIR` to prevent confusing it with salt-states.

rallytime:
@bdrung We don't like making these kinds of changes to config file like this in dot releases. If you could move this to `develop`, that would be great. Adding this to the `Fluorine` release notes as a new feature would be great, too, when you move it over.

bdrung:
I'll rebase the merge proposal on `develop`.

PR #46137, opened 2018-02-21
[2017.7] update ec2 pillar arguments with better names

damon-atkins:
### What does this PR do?

Update the arguments to a more suitable name.

Relates to PR #45878 which has not been released.

### Previous Behavior

As per PR #45878

### New Behavior

PR #45878

### Commits signed with GPG?

No

### DIFF
--- a/salt/pillar/ec2_pillar.py
+++ b/salt/pillar/ec2_pillar.py
@@ -1,10 +1,10 @@
-#-*- coding: utf-8 -*-
+# -*- coding: utf-8 -*-
 '''
 Retrieve EC2 instance data for minions for ec2_tags and ec2_tags_list
 
-The minion id must be the AWS instance-id or value in 'tag_key'.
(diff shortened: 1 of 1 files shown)


gtmanfred (changes requested):
Unfortunately because this is in a release, i don't think that we can just change it.

It would need to go through a deprecation process and be merged into develop.

gtmanfred (approved):

gtmanfred:
Nevermind, i missed the part that it was added after 2017.7.3, and isn't in a release yet.

rallytime:
re-run py

damon-atkins:
@rallytime could you merge this before the next 2017.7 release pls if the original pr is in the next release.

gtmanfred:
The original one will not be in 2017.7.4, we only few cherry-picked bug fixes in 2017.7.4

damon-atkins:
I have another PR to come after this one for same file.    Heaps more doco...etc.

Issue #34423, opened 2016-07-01
oscodename wrong on Debian 8 (jessie)


bdrung:
On a Debian 8 (jessie) system where lsb-release is not installed, the oscodename is wrong (it should be just "jessie"):

```
$ sudo salt debian grains.items   
[...]
   lsb_distrib_codename:
        Debian GNU/Linux 8 (jessie)
    lsb_distrib_id:
        Debian GNU/Linux
    lsb_distrib_release:
        8
[...]
    os:
        Debian
    os_family:
        Debian
    osarch:
        amd64
    oscodename:
        Debian GNU/Linux 8 (jessie)
    osfullname:
        Debian GNU/Linux
    osrelease:
        8
```

When installing lsb-release, the output will be correct:

```
$ sudo salt debian grains.items   
[...]
    lsb_distrib_codename:
        jessie
    lsb_distrib_description:
        Debian GNU/Linux 8.4 (jessie)
    lsb_distrib_id:
        Debian
    lsb_distrib_os:
        GNU/Linux
    lsb_distrib_release:
        8.4
[...]
    os:
        Debian
    os_family:
        Debian
    osarch:
        amd64
    oscodename:
        jessie
    osfinger:
        Debian-8
    osfullname:
        Debian
    osmajorrelease:
        8
    osrelease:
        8.4
    osrelease_info:
        - 8
        - 4
```

salt 2016.3.0 and the develop branch are affected.

Currently `os_data()` is overly complex and has issues: When lsb-release is not found, it will fill out the lsb_\* values by falling back to various sources. Instead the lsb_\* values should only be set, when lsb-release is available.

Why don't you use `platform.linux_distribution()` if that provides useful information?

I am willing to fix this bug, but I don't want to add another quirk to make the code even less readable.

bdrung:
Maybe it is better to fix ```platform.linux_distribution()``` to provide useful data.

Ch3LL:
@bdrung I was able to easily replicate this and am seeing the same thing you are bringing attention to.

We don't want to refactor grain code too much to ensure we don't change values for others, particularly in a dot release. If you want to refactor a lot of it just make sure to add that to the develop branch and then we can review the PR. 

If ou do want to just approach this by just fixing just this bug you can add that to the 2016 branch. 

I'm not entirely sure if there is a specifc reason as to why linux_distrubtion has not been used in the paste instead. maybe @cachedout do you have any insight into why we aren't using platform.linux_distrubtion() instead?

[...]

PR #46139, opened 2018-02-21
Add os grains test cases for Debian/Ubuntu and fix oscodename on Ubuntu

bdrung:
This pull requests adds os grains test cases for Debian and Ubuntu (to address comments in #34423) and fixes an incorrect oscodename grain on Ubuntu (found the test case).

### DIFF
--- a/tests/unit/grains/test_core.py
+++ b/tests/unit/grains/test_core.py
@@ -209,7 +209,7 @@ class CoreGrainsTestCase(TestCase, LoaderModuleMockMixin):
         self.assertEqual(os_grains.get('os_family'), 'Suse')
         self.assertEqual(os_grains.get('os'), 'SUSE')
 
-    def _run_suse_os_grains_tests(self, os_release_map):
+    def _run_os_grains_tests(self, os_release_map):
         path_isfile_mock = MagicMock(side_effect=lambda x: x in os_release_map['files'])
         empty_mock = MagicMock(return_value={})
         osarch_mock = MagicMock(return_value="amd64")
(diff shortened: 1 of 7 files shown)


tests/unit/grains/test_core.py
```diff
@@ -243,10 +243,11 @@ def _import_mock(name, *args):
                                 # Mock linux_distribution to give us the OS
                                 # name that we want.
                                 distro_mock = MagicMock(
-                                    return_value=('SUSE test', 'version', 'arch')
+                                    return_value=os_release_map['linux_distribution']
                                 )
                                 with patch("salt.utils.fopen", mock_open()) as suse_release_file:
-                                    suse_release_file.return_value.__iter__.return_value = os_release_map.get('suse_release_file', '').splitlines()
+                                    suse_release_file.return_value.__iter__.return_value = \
+                                        os_release_map.get('suse_release_file', '').splitlines()
```
isbm: Doesn't looks like you've removed whole SUSE-related things from this function

isbm (changes requested):
I am not sure if this is a good approach to the unit tests, when you put pre-parsed data and simply comparing two dictionaries. It would make more sense to get a real data from Debian/Ubuntu machines, mock-up system reads and then pipe it through instead.

tests/unit/grains/test_core.py
bdrung: `_run_suse_os_grains_tests` also uses this `_run_os_grains_tests` function. Besides that this code part was also part of `_run_ubuntu_os_grains_tests`.

bdrung:
These are unit tests. Instead of using raw file content, we should add a unit tests for the `_parse_os_release` function.

bdrung:
Added unit tests for `_parse_os_release`.

isbm:
I mean, it would be nicer to have just a data samples somewhere (subdir?). I am sure this is different from distro to distro... Unfortunately.

bdrung:
Moving all the sample data into a subdirectory is an option (this could be done as additional commit / pull request)

isbm:
@bdrung since you're taking care of these unit tests, you could actually move that data right in this PR with no problems. It as a nice and complete unit test, so why not.

bdrung:
Implemented as requested. The os-release files are placed in a os-releases directory.

isbm (approved):
Great!

rallytime:
@bdrung A small lint error here: https://jenkins.saltstack.com/job/PR/job/salt-pr-lint-n/19475/violations/file/tests/unit/grains/test_core.py/

bdrung:
Rebased and fixed pylint error.

rallytime (approved):

PR #46147, opened 2018-02-22
Fix salt-ssh state.apply "'NoneType' object has no attribute 'count'" and unicode issue

dincamihai:
### What does this PR do?

Fixes "'NoneType' object has no attribute 'count'" and 

```
[ERROR   ] An Exception occurred while executing state.apply: 'ascii' codec can't decode byte 0xc2 in position 10: ordinal not in range(128) 
  File "/usr/lib/python2.7/site-packages/salt/client/ssh/__init__.py", line 1133, in run_wfunc
    result = self.wfuncs[self.fun](*self.args, **self.kwargs)
  File "/usr/lib/python2.7/site-packages/salt/client/ssh/wrapper/state.py", line 462, in apply_
    return sls(mods, **kwargs)
  File "/usr/lib/python2.7/site-packages/salt/client/ssh/wrapper/state.py", line 190, in sls
    stdout, stderr, _ = single.cmd_block()
  File "/usr/lib/python2.7/site-packages/salt/client/ssh/__init__.py", line 1272, in cmd_block
    stdout, stderr, retcode = self.shim_cmd(cmd_str)
  File "/usr/lib/python2.7/site-packages/salt/client/ssh/__init__.py", line 1218, in shim_cmd
    return self.shell.exec_cmd(cmd_str)
  File "/usr/lib/python2.7/site-packages/salt/client/ssh/shell.py", line 344, in exec_cmd
    ret = self._run_cmd(cmd)
  File "/usr/lib/python2.7/site-packages/salt/client/ssh/shell.py", line 391, in _run_cmd
    stdout, stderr = term.recv()
  File "/usr/lib/python2.7/site-packages/salt/utils/vt.py", line 310, in recv
    return self._recv(maxsize)
  File "/usr/lib/python2.7/site-packages/salt/utils/vt.py", line 679, in _recv
    os.read(self.child_fd, maxsize)
  File "/usr/lib/python2.7/site-packages/salt/utils/vt.py", line 343, in _translate_newlines
    return data.replace('\r\n', os.linesep)
```

### Tests written?

No

### Commits signed with GPG?

No

### DIFF
--- a/salt/utils/vt.py
+++ b/salt/utils/vt.py
@@ -642,7 +642,7 @@ class Terminal(object):
             if self.child_fde in rlist:
                 try:
                     stderr = self._translate_newlines(
-                        salt.utils.stringutils.to_str(
+                        salt.utils.stringutils.to_unicode(
                             os.read(self.child_fde, maxsize)
                         )
                     )
@@ -675,7 +675,7 @@ class Terminal(object):
             if self.child_fd in rlist:
                 try:
                     stdout = self._translate_newlines(
-                        salt.utils.stringutils.to_str(
+                        salt.utils.stringutils.to_unicode(
                             os.read(self.child_fd, maxsize)
                         )
                     )
--- a/salt/client/ssh/__init__.py
+++ b/salt/client/ssh/__init__.py
@@ -440,7 +440,7 @@ class SSH(object):
             if target.get('passwd', False) or self.opts['ssh_passwd']:
                 self._key_deploy_run(host, target, False)
             return ret
-        if ret[host].get('stderr', '').count('Permission denied'):
+        if (ret[host].get('stderr') or '').count('Permission denied'):
             target = self.targets[host]
             # permission denied, attempt to auto deploy ssh key
             print(('Permission denied for host {0}, do you want to deploy '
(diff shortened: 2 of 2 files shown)


terminalmage (changes requested):
We don't need the `u` on the string literals. Otherwise this is GTM.

dincamihai:
@terminalmage fixing that. thanks for looking

dincamihai:
@terminalmage Done.

salt/utils/vt.py
```diff
@@ -340,7 +340,7 @@ def _translate_newlines(self, data):
         if data is None or not data:
             return
         # PTY's always return \r\n as the line feeds
-        return data.replace('\r\n', os.linesep)
+        return data.decode('utf-8').replace('\r\n', os.linesep)
```
terminalmage: We are only invoking `self._translate_newlines()` on strings which have already been passed through `salt.utils.stringutils.to_str()`. This means that when run on Python 3, `data` will be a `str` type, which has no `decode` attribute:

```zsh
% python3 -c 'print("foo".decode("utf-8"))'
  File "<string>", line 1, in <module>
AttributeError: 'str' object has no attribute 'decode'
```

A better fix would be to revert line 343, and change the calls to `salt.utils.stringutils.to_str()` (lines 645 & 678) to use `salt.utils.stringutils.to_unicode()` instead.

terminalmage (changes requested):

salt/utils/vt.py
terminalmage: Sorry I did not catch this earlier, I did not see it in the diff in my initial review.

dincamihai: makes sense. doing it now

dincamihai:
@terminalmage fixed

terminalmage (approved):

rallytime:
re-run py3

PR #46145, opened 2018-02-22
3 small fixes for runners/orchestration

terminalmage:
This fixes the following issues:

1. The orchestration jid was not being passed to the `state.orchestrate` runner when invoked using the `state.orch` alias. This did not affect the runner's performance/results, but merely kept the jid from being included in the metadata for the job in the job cache.

2. Fix an edge case when orchestration attempts to compile pilllar data before making any other fileclient requests. (Resolves #46004)

3. Keeps salt-run's opts from being modified while running the `pillar.show_pillar` runner.

Additionally, it makes a small performance/memory improvement in the state compiler by extending lists rather than using `+=`, and documents some previously-undocumented behavior when orchestration is invoked using an SLS file containing state SLS instead of orchestration SLS.

PR #46149, opened 2018-02-22
Back-port #46145 to oxygen.rc1

rallytime:
Back-port #46145 to oxygen.rc1


terminalmage (approved):
Just needs the additional commit I added to #46145 to be cherrypicked and it'll be GTM. Should not cause a conflict when CP'ed.

terminalmage (approved):
:+1:

PR #46155, opened 2018-02-23
Add a KMS Envelope-Encryption Renderer

kojiromike:
### What does this PR do?

This provides a renderer to decrypt values stored using a KMS data key as per [KMS envelope encryption](https://docs.aws.amazon.com/kms/latest/developerguide/workflow.html). It is pretty similar to the gpg renderer -- it just uses a different approach to do the actual decryption. This is something that is useful to me. I hope it's useful to others, too.

### New Behavior

1. Provide a encrypted KMS data key in your master config under the name `aws_kms_data_key`.
2. Ensure that the salt master has appropriate iam permissions and configuration to access the KMS key needed to decrypt the data key.
3. Encrypt some values using the plaintext value of the data key with [`cryptography.fernet`](https://cryptography.io/en/latest/fernet/) and put them in a sls file to be rendered via the kms renderer provided in this PR.

### Notes and sales pitch

This is most useful for salt users already in an AWS environment who need shared secrets stored in pillars and prefer to put them e.g. in Git rather than in S3. With GPG you would need each salt user to have their own key, but encrypted PGP messages get linearly larger with each recipient key. Of course, you could use a single shared GPG key, but then you'd have to figure out how to securely share that key. KMS just handles the securely-sharing keys bit for us.

### Tests written?

Yes

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/renderers/aws_kms.py
+++ b/salt/renderers/aws_kms.py
@@ -188,6 +188,8 @@ def _plaintext_data_key():
         setattr(_plaintext_data_key, 'response', response)
     key_id = response['KeyId']
     plaintext = response['Plaintext']
+    if hasattr(plaintext, 'encode'):
+        plaintext = plaintext.encode(__salt_system_encoding__)
     log.debug('Using key %s from %s', key_id, 'cache' if cache_hit else 'api call')
     return plaintext
 
@@ -208,13 +210,13 @@ def _decrypt_ciphertext(cipher, translate_newlines=False):
     '''
     if translate_newlines:
         cipher = cipher.replace(r'\n', '\n')
-    if six.PY3:
+    if hasattr(cipher, 'encode'):
         cipher = cipher.encode(__salt_system_encoding__)
 
     # Decryption
     data_key = _base64_plaintext_data_key()
     plain_text = fernet.Fernet(data_key).decrypt(cipher)
-    if six.PY3 and isinstance(plain_text, bytes):
+    if hasattr(plain_text, 'decode'):
         plain_text = plain_text.decode(__salt_system_encoding__)
     return six.text_type(plain_text)
 
--- a/salt/renderers/aws_kms.py
+++ b/salt/renderers/aws_kms.py
@@ -70,7 +70,6 @@ import base64
 
 # Import salt libs
 import salt.utils.stringio
-import salt.utils.decorators as decorators
 
 # Import 3rd-party libs
 from salt.ext import six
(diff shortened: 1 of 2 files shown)


damon-atkins:
Please rename to aws_kms  as kms is to generic.

kojiromike:
OK, I did that.

kojiromike:
It occurs to me that the main difference between GPG and AWS_KMS is the configuration of a particular `decrypt` function. While I don't have time to work on a generic decrypt renderer _just_ now, I'm curious what kind of interest there would be in such a thing.

damon-atkins:
The trick is to make the salt interface generic so you can swap one in and swap another out. Bit like pkg almost does

kojiromike:
Updates: I've rewritten some of the code to facilitate testing, tweaked the docblock a bit, renamed the renderer to aws_kms and the config key to aws_kms_data_key and written unit tests.

rallytime:
@kojiromike Can you write some tests for these new files?

gtmanfred (approved):

kojiromike:
@rallytime I did write tests for these new files. Did I miss something?

rallytime:
@kojiromike Oh, you're right. I'm not sure how I missed that. My apologies. It does look like one of them is failing however: https://jenkins.saltstack.com/job/PR/job/salt-pr-rs-cent7-n/16770/.

There is also a small lint error: https://jenkins.saltstack.com/job/PR/job/salt-pr-lint-n/19717/violations/file/salt/renderers/aws_kms.py/

kojiromike:
@rallytime Oh, thanks, I fixed the whitespace error. When I run the tests locally (`./tests/runtests.py -n unit.renderers`) I get zero errors. But I can't view the result in Jenkins -- it seems to just load forever.

kojiromike:
I was able to see the Jenkins error now. I will attempt to reproduce locally and fix.

kojiromike:
Whew, that was a bigger task than I thought. Not only was it a legitimate error, but I was clearly letting some library calls slide through the mocks.

rallytime:
@kojiromike Looks like there's a small lint error here: https://jenkins.saltstack.com/job/PR/job/salt-pr-lint-n/19824/violations/file/salt/renderers/aws_kms.py/

rallytime:
@kojiromike It looks like your new tests are failing on the PY3 builds. Mind taking another look?

kojiromike:
Yep, fixed up. Sorry, not used to working in both major versions of Python at the same time.

PR #46162, opened 2018-02-23
Add team-suse to CODEOWNERS file for zypper files

rallytime:
This addition will enable automatic reviews from team-suse for PRs that touch the zypper files.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -52,6 +52,9 @@ salt/**/thin.py                     @saltstack/team-ssh
 # Team State
 salt/state.py                       @saltstack/team-state
 
+# Team SUSE
+salt/**/*zypper*                      @saltstack/team-suse
+
 # Team Transport
 salt/transport/                     @saltstack/team-transport
 salt/utils/zeromq.py                @saltstack/team-transport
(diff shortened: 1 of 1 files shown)


rallytime:
@saltstack/team-suse - this is just a the start of what you "own" and will be auto-requested for reviews. Please feel free to add your team to other files as you see fit. For example, I imagine the `yumpkg` module and the kubernetes files come to mind.

isbm:
While xfs/btrfs modules are also ours, as per conversation, I think we are fine to keep an eye on the entire `/*`.

isbm (commented):
While xfs/btrfs modules are also ours, as per conversation, I think we are fine to keep an eye on the entire /* being a backup of yours. Thanks for doing this!

rallytime:
OK, I added those files here as well. I need to implement the "mirroring" of the team-core files on the bot itself. I probably won't be able to do that until next week. So this change will just make sure `team-suse` is specifically requested to review changes to these files.

meaksh (commented):
Hey @rallytime , `salt/**/*snapper*` should be also set to `team-suse`. Thanks!

isbm:
And `*pkg*`.

rallytime:
Ah, yeah good call @isbm and @meaksh! How does that look now?

PR #46174, opened 2018-02-23
Fix a unicode issue with the git module on Windows

twangboy:
### What does this PR do?
Fixes an issue with the git module on Windows where the return of the git.exe is converted to Unicode using the `salt.utils.stringutils.to_unicode` salt util. This function defaults to the system codepage which is CP1252 on Windows, resulting in incorrectly decoded characters. This PR adds the `output_encoding` parameter and forces the git module to use UTF-8 unless otherwise specified in the `output_encoding` parameter.

Found while fixing tests:
https://github.com/saltstack/salt-jenkins/issues/439

### Tests written?
Yes

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/modules/git.py
+++ b/salt/modules/git.py
@@ -67,6 +67,7 @@ def _config_getter(get_opt,
                    user=None,
                    password=None,
                    ignore_retcode=False,
+                   output_encoding=None,
                    **kwargs):
     '''
     Common code for config.get_* functions, builds and runs the git CLI command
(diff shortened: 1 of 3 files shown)


terminalmage:
re-run all

twangboy:
Fixes https://github.com/saltstack/salt/issues/46360

rallytime:
re-run py

terminalmage (approved):

PR #46176, opened 2018-02-24
Add docker.prune - pruning for Docker subsystems

The-Loeki:
### What does this PR do?
Add `docker.prune`, which is capable of clearing Docker's subsystems of cruft. Which seems necessary, a lot ;)


The-Loeki:
> Function prune has a Cognitive Complexity of 6 (exceeds 5 allowed). Consider refactoring.

FOR REALZ ROFLMAO I could explain this function to my Cognitively Complexively Challenged neighbours

rallytime:
@The-Loeki Yeah, we need to tweak the rules for the code-climate build. Can you also add some tests for this new functions?

The-Loeki:
Hi, I'm afraid I'm going to have to pass on that one due to time constraints... I will however update the function once I get the same function PR'd in docker-py; that should get rid of the hidden function calls.

After that I'm not even sure what there is to test in this case. That is, I'm assuming Docker does it's testing on the API endpoints, `_client_wrapper` is tested as well, I'm not parsing / evalling any output, so what's left is to test is the if constructs?

salt/modules/dockermod.py
```diff
@@ -6786,3 +6786,58 @@ def sls_build(repository,
         stop(id_)
         rm_(id_)
     return ret
+
+
+def prune(system=True, containers=False, networks=False, images=False, build=False, volumes=False, **filters):
+    '''
+    Prune Docker's various subsystems
+
```
terminalmage: Doing a quick sweep of the codebase, these API functions all seem to have been added in docker-py 2.1.0. Can you please add the following here?

```rst
    .. note::
        This requires docker-py version 2.1.0 or later.
```

If pre-2.1.0 docker-py is installed, `_client_wrapper()` function will handle the API error from the unknown function call, so we won't have a traceback. This is just a bit of extra information that should help users who may be on earlier docker-py releases.

terminalmage (changes requested):

terminalmage (changes requested):
Actually there are a few things about this which don't actually work at all, it looks like it may not even have been tested since the `kwargs` argument doesn't exist for these API funcs and trying to run this function on my laptop immediately resulted in a tracecback.

I'm working on some modifications that I can submit against your fork that should clear this up.

terminalmage:
Another thing I noticed as I look deeper into the docker-py codebase is that negative filters don't seem to be supported. I've opened https://github.com/docker/docker-py/issues/1940 upstream to get clarification.

terminalmage:
@The-Loeki Have a look at https://github.com/The-Loeki/salt/pull/1, I made some fixes, changed how `system` works, and added unit tests.

The-Loeki:
@terminalmage thanks for all the help and advice, regarding the older api's: I can probably easily write up the same kind of workarounds for the other prunes?

terminalmage:
You mean posting directly to the API endpoint? Yeah, that should be doable, and I can also work to get them supported in docker-py. Which other prunes are yet to be supported?

The-Loeki:
AFAIK just the one I already opened a PR for, so that would be purely a backwards compat feature

terminalmage:
Oh, so you were talking about making backwards-compatibility changes for containers, images, etc? If so, I don't think it's necessary in this PR.

PR #46193, opened 2018-02-26
Bugfix: SSDP crash on Python3  environment (bytes vs unicode)

isbm:
### What does this PR do?

Bugfix

Unicode won't be sent over socket, but bytes. This was overlooked.

### Tests written?

Yes


terminalmage (approved):

rallytime:
re-run lint

PR #46197, opened 2018-02-26
Add mapper to salt.utils.args.split_input()

terminalmage:
This allows for one to pass a function reference as a second argumment
to split_input, which will be used to perform conversion on the split
input. This is useful for instances where we want to get a list of ints
out of a Salt CLI arg like foo=1,2,3. With the pre-existing code, this
would force you to either manually run something like:

```python
map(int, salt.utils.args.split_input(val))
```

With this PR, you can now run:

```python
salt.utils.args.split_input(val, int)
```

### DIFF
--- a/salt/utils/args.py
+++ b/salt/utils/args.py
@@ -14,7 +14,7 @@ import shlex
 # Import salt libs
 from salt.exceptions import SaltInvocationError
 from salt.ext import six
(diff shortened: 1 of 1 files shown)


terminalmage:
actually this needs to be modified, this will return an iterator, not a list.

rallytime:
re-run py

rallytime (approved):

PR #46206, opened 2018-02-26
adding loaded feature to mac_service

weswhet:
### What does this PR do?
Adds the ability for the mac_service module to recognize whether or not a mac service should have an active PID when checking the status, as some macOS services run on intervals or are initiated by other manners. This improves the support for the service state which previously would only work properly for services that are set be always running with an active PID.

None

### Previous Behavior
service.status on macOS would return an empty string for any service that did not have an active PID. Which would send a false positive of sorts to the service state module.

### New Behavior
Salt will now check the macOS service to see whether or not the service should actively be running by checking the `KeepAlive` key in service plist. If the service is not meant to be always running but the service is properly loaded in launchd service.status will now return a string 'loaded'.

CC @gtmanfred as we briefly talked about this addition. 

### Tests written?

No

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/modules/mac_service.py
+++ b/salt/modules/mac_service.py
@@ -162,7 +162,7 @@ def _always_running_service(name):
 
     :rtype: bool
 
-    .. versionadded::
+    .. versionadded:: Fluorine
     '''
 
     # get all the info from the launchctl service
--- a/salt/modules/mac_service.py
+++ b/salt/modules/mac_service.py
@@ -150,6 +150,37 @@ def _get_service(name):
     raise CommandExecutionError('Service not found: {0}'.format(name))
 
 
+def _always_running_service(name):
+    '''
+    Check if the service should always be running based on the KeepAlive Key
+    in the service plist.
+
+    :param str name: Service label, file name, or full path
+
+    :return: True if the KeepAlive key is set to True, False if set to False or
+        not set in the plist at all.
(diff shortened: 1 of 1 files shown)


gtmanfred (approved):

rallytime:
Welcome @weswhet! Thank you for this change. 

Looks like there is a lint error on this: https://jenkins.saltstack.com/job/PR/job/salt-pr-lint-n/19595/violations/file/salt/modules/mac_service.py/

Can you fix that up?

weswhet:
Thanks for the warm welcome @rallytime, lint check is passing now. Let me know if I need to do anything else. 😄

salt/modules/mac_service.py
```diff
@@ -150,6 +150,37 @@ def _get_service(name):
...
+    in the service plist.
+
+    :param str name: Service label, file name, or full path
+
+    :return: True if the KeepAlive key is set to True, False if set to False or
+        not set in the plist at all.
+
+    :rtype: bool
+
+    .. versionadded::
```
rallytime: Oops, this line should be `.. versionadded:: Fluorine`

weswhet: whoops, looks like you beat me to it. Thanks

rallytime (approved):

PR #46208, opened 2018-02-27
Blacklist os.umask

terminalmage:
We have for a while had a context manager to ensure that temporary umask changes are undone, but most of the umask usage is still done manually, and not all of it uses a try/finally block to ensure we undo the umask change.

This PR replaces `os.umask()` usage with either `salt.utils.files.set_umask()` or `salt.utils.files.get_umask()`, and blacklists `os.umask()` using our pylint check.

NOTE: @rallytime this will likely result in new lint failures as this is merged forward, in the event that any code new to the `2018.3` and/or `develop` branch introduces new calls to `os.umask()`.

### DIFF
--- a/salt/client/ssh/ssh_py_shim.py
+++ b/salt/client/ssh/ssh_py_shim.py
@@ -106,8 +106,10 @@ def need_deployment():
     if os.path.exists(OPTIONS.saltdir):
         shutil.rmtree(OPTIONS.saltdir)
     old_umask = os.umask(0o077)
-    os.makedirs(OPTIONS.saltdir)
-    os.umask(old_umask)
+    try:
+        os.makedirs(OPTIONS.saltdir)
+    finally:
+        os.umask(old_umask)
     # Verify perms on saltdir
     if not is_windows():
         euid = os.geteuid()
--- a/salt/client/ssh/ssh_py_shim.py
+++ b/salt/client/ssh/ssh_py_shim.py
@@ -105,11 +105,11 @@ def need_deployment():
     """
     if os.path.exists(OPTIONS.saltdir):
         shutil.rmtree(OPTIONS.saltdir)
-    old_umask = os.umask(0o077)  # pylint: disabe=blacklisted-function
+    old_umask = os.umask(0o077)  # pylint: disable=blacklisted-function
     try:
         os.makedirs(OPTIONS.saltdir)
     finally:
-        os.umask(old_umask)  # pylint: disabe=blacklisted-function
+        os.umask(old_umask)  # pylint: disable=blacklisted-function
     # Verify perms on saltdir
     if not is_windows():
         euid = os.geteuid()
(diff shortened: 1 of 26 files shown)


salt/utils/verify.py
```diff
@@ -218,12 +219,11 @@ def verify_env(dirs, user, permissive=False, pki_dir='', skip_extra=False, root_
             continue
         if not os.path.isdir(dir_):
             try:
-                cumask = os.umask(18)  # 077
-                os.makedirs(dir_)
+                with salt.utils.files.set_umask(0o022):
```
terminalmage: Note that while this seems like an incorrect change, 022 is octal for 18. The _original_ umask here was 077, but this was changed in 7eb909871f8ca3d31379d392c6c1d1f8fc106193 many years ago. The comment was just not modified to reflect the change.

salt/client/ssh/ssh_py_shim.py
```diff
@@ -106,8 +106,10 @@ def need_deployment():
     if os.path.exists(OPTIONS.saltdir):
         shutil.rmtree(OPTIONS.saltdir)
     old_umask = os.umask(0o077)
```
DmitryKuzmenko: Seems this `os.umask` has to be replaced too?
Maybe the correct one would be just:
```python
with salt.utils.files.set_umask(0o077):
    os.makedirs(OPTIONS.saltdir)
```

salt/modules/file.py
```diff
@@ -5032,30 +5032,25 @@ def _set_mode_and_make_dirs(name, dir_mode, mode, user, group):
...
-                    if touch(name):
-                        ret['changes']['diff'] = 'New file'
-                    else:
-                        return _error(
-                            ret, 'File {0} not created'.format(name)
-                        )
-
-            if mode:
-                os.umask(current_umask)
+            with salt.utils.files.set_umask(0o077 if mode else None):
```
DmitryKuzmenko: :+1: Very cute! Like it!

salt/client/ssh/ssh_py_shim.py
DmitryKuzmenko: Ah, I see, it's a thin module that shouldn't use utils. Am I right?
Don't we have to mark these lines with `# pylint: disable=...` or `# future pylint: disable=...`?

DmitryKuzmenko (approved):
:+1:

salt/client/ssh/ssh_py_shim.py
terminalmage: Yes, I need to disable this check here, thanks. And yeah, we don't want to use the utils in the shim because they won't necessarily be available.

rallytime:
@terminalmage There's some more lint errors on this.

terminalmage:
@rallytime Should be fixed with my latest push

rallytime (approved):

PR #46213, opened 2018-02-27
PY2: Remove conversion of values from config file to str types

terminalmage:
This removes a weird block that only affected string types loaded from a config file. We absolutely *do* want unicode types here, so we shouldn't be encoding them to str types. Also, this misguided code ignored strings in nested data structures, so it didn't even properly do what it was *trying* to do.


DmitryKuzmenko (approved):
Works!

terminalmage:
Thanks for confirming. I have tests to add, so I'll label his as "Pending Changes" for now.

rallytime:
@terminalmage There's a test failure here: https://jenkins.saltstack.com/job/PR/job/salt-pr-rs-cent7-n/16699/

rallytime (approved):

PR #46214, opened 2018-02-27
[DOC] Replace `note` rST block for GitHub

vutny:
### What does this PR do?
This PR removes `note` rST block in favor of just bold heading text in example of README file at formulas doc page.
It seems neither GitHub, nor Bitbucket does not properly support it and the rendered text does not look expressive. Same thing was done previously in saltstack/salt-bootstrap repository.

### Commits signed with GPG?
Yes

### DIFF
--- a/doc/topics/development/conventions/formulas.rst
+++ b/doc/topics/development/conventions/formulas.rst
@@ -1242,10 +1242,10 @@ A sample skeleton for the ``README.rst`` file:
 
     Install and configure the FOO service.
 
-    .. note::
+    **NOTE**
 
-        See the full `Salt Formulas installation and usage instructions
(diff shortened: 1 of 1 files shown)


rallytime:
@vutny I don't think we want to change that here because the file is an `.rst` file and we use that `.. note` syntax in our doc builds. 

I changed it in the salt-bootstrap file because we don't actually build those docs anywhere to use the rst syntax.

vutny:
@rallytime I think you'd not build that particular README file since it is embedded in own
```
.. code-block:: restructuredtext
```
block. So it does not affect building properly formatted formulas page.

rallytime (approved):
@vutny Ah, I see. You're right. Thanks!

PR #46216, opened 2018-02-27
Document out-of-date or wrong

edlane:
The URL of this document is the first result for Google search "salt-cloud azurearm" but it appears to be out of date or just plain wrong leading to dispair and suffering :-)
In particular, most of the named config parameters in this document are invalid.

**HOWEVER:** Another saltstack document appears to be correct and up-to-date: https://docs.saltstack.com/en/latest/ref/clouds/all/salt.cloud.clouds.azurearm.html

I suggest eliminating the incorrect doc alltogether or to fix it by incorporating the information from the "working" version.

### DIFF
--- a/doc/topics/cloud/azurearm.rst
+++ b/doc/topics/cloud/azurearm.rst
@@ -316,6 +316,7 @@ How this is used depends on the operating system that is being deployed. If
 used, any ``userdata`` setting will be ignored.
 
 wait_for_ip_timeout
+
 -------------------
 Optional. Default is ``600``. When waiting for a VM to be created, Salt Cloud
 will attempt to connect to the VM's IP address until it starts responding. This


rallytime:
Hi @edlane! Did you mean to submit other changes to this? Or maybe this should be an issue? We certainly want our documentation to be accurate, but the change you've supplied here is just adding a blank line, which we shouldn't do here because it will affect the headers in this document.

edlane:
@rallytime !!
Yes, you are right, it should probably be an issue.
I always wanted to click that "Edit This Page" link on the docs and see where it ended up :smile: 

Several people here at SUSE have gone down a "rat hole of dispair" caused by this doc so I would suggest just getting rid of it but I don't know all the implications of doing that in a PR.

If better I will submit an issue -- Please advise.

Great conversing with you in GH -- like old times :-)

rallytime:
@edlane Nice to see you again, as well! An issue sounds great! Then we can discuss what should be done to get these docs updated and go from there. :)

edlane:
Thanks!
done...
https://github.com/saltstack/salt/issues/46220

rallytime:
Thanks - I'll close this then.

PR #46226, opened 2018-02-27
Git Pillars: Fix floaty branch names stacktraces

The-Loeki:
### What does this PR do?
Set `git_pillar_base` to a floaty value, e.g.

```yaml
git_pillar_base: 2017.7
```
or even
```yaml
git_pillar_base: '2017.7'
```

will make the value be loaded as a `float`. GitFS then proceeds to stacktrace on it:

```sh
2018-02-27 22:00:56,296 [salt.pillar      :843 ][ERROR   ][7497] Execption caught loading ext_pillar 'git':
  File "/usr/lib/python2.7/dist-packages/salt/pillar/__init__.py", line 833, in ext_pillar
    key)
  File "/usr/lib/python2.7/dist-packages/salt/pillar/__init__.py", line 760, in _external_pillar_data
    pillar_dirs)
  File "/usr/lib/python2.7/dist-packages/salt/pillar/git_pillar.py", line 558, in ext_pillar
    pillar.checkout()
  File "/usr/lib/python2.7/dist-packages/salt/utils/gitfs.py", line 2753, in checkout
    cachedir = self.do_checkout(repo)
  File "/usr/lib/python2.7/dist-packages/salt/utils/gitfs.py", line 2409, in do_checkout
    return repo.checkout()
  File "/usr/lib/python2.7/dist-packages/salt/utils/gitfs.py", line 1227, in checkout
    local_ref = 'refs/heads/' + tgt_ref
```
### Previous Behavior
Stacktrace, don't checkout repo

### New Behavior
Don't stacktrace, checkout repo

### DIFF
--- a/salt/utils/gitfs.py
+++ b/salt/utils/gitfs.py
@@ -1224,9 +1224,9 @@ class Pygit2(GitProvider):
         Checkout the configured branch/tag
         '''
         tgt_ref = self.get_checkout_target()
-        local_ref = 'refs/heads/' + tgt_ref
-        remote_ref = 'refs/remotes/origin/' + tgt_ref
-        tag_ref = 'refs/tags/' + tgt_ref
+        local_ref = 'refs/heads/{0}'.format(tgt_ref)
+        remote_ref = 'refs/remotes/origin/{0}'.format(tgt_ref)
+        tag_ref = 'refs/tags/{0}'.format(tgt_ref)
 
         try:
             local_head = self.repo.lookup_reference('HEAD')


terminalmage (commented):
We have code in salt.utils.gitfs that is supposed to normalize config values so we have strings unless otherwise specified. This works in gitfs, and the correct fix here would be to ensure that it also is run for git_pillar too.

terminalmage:
Also, gitfs doesn't ever check out the repo, so `gitfs_base` wouldn't be affected anyway.

terminalmage:
Also, I see no issue number referenced, and there's no versions report, which would be helpful in troubleshooting this.

The-Loeki:
No issue, ran into it when I did that branching thing and wrote this quick fix.

terminalmage:
Well it's not the correct fix, and I can't reproduce the traceback anyway during checkout.

terminalmage:
I'm working on a proper fix that will ensure that global-only values (which is not a concept that exists in git_pillar but not gitfs) are also normalized.

The-Loeki:
@terminalmage I'm looking for the piece of you point out, but I'm having no luck.
AFAICT the value gets in there through `GitProvider.get_checkout_target`, which in turn takes from `self.opts`, which in turn gets filled with `opts` in `__init__`

terminalmage:
I opened an issue to track this, I was finally able to find a combination of configuration to reproduce the traceback: https://github.com/saltstack/salt/issues/46258

My fix for that and https://github.com/saltstack/salt/issues/46259 will be coming in the same PR, which I should be opening shortly.

Closing this in favor of my upcoming PR. Thanks for reporting this.

terminalmage:
@The-Loeki https://github.com/saltstack/salt/pull/46260

PR #46229, opened 2018-02-27
Add trust_failures option to svn.export and svn.latest.

smarsching:
### What does this PR do?

This PR adds the new `trust_failures` option to the `svn.export` and `svn.latest` states.

If set, the value of this option is passed to `svn` using the `--trust-server-cert-failures` option. This is useful in cases where using the `--trust-server-cert` option (as enabled by the `trust` option of the states) is not sufficient. One example of such a case is when the host name of the SVN server does not match the name in the certificate. In this case, specifying `trust: cn-mismatch` will make `svn` still accept the certificate.

None, it is a new feature.

### Tests written?

No

### Commits signed with GPG?

No

### DIFF
--- a/salt/states/svn.py
+++ b/salt/states/svn.py
@@ -48,7 +48,8 @@ def latest(name,
            password=None,
            force=False,
            externals=True,
-           trust=False):
+           trust=False,
+           trust_failures=None):
     '''
     Checkout or update the working directory to the latest revision from the
     remote repository.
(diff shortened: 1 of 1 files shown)


salt/states/svn.py
```diff
@@ -84,6 +85,12 @@ def latest(name,
 
     trust : False
         Automatically trust the remote server. SVN's --trust-server-cert
+
+    trust_failures : None
+        Comma-separated list of certificate trust failures, that shall be
+        ignored. This can be used if trust=True is not sufficient. The
+        specified string is passed to SVN's --trust-server-cert-failures
+        option as-is.
```
rallytime: Can you add a `.. versionadded:: Fluorine` tag to this new option?

salt/states/svn.py
```diff
@@ -222,6 +233,12 @@ def export(name,
...
```
rallytime: Same comment here, too.

rallytime (changes requested):
Thanks @smarsching! I just had two small requests.

salt/states/svn.py
smarsching: Sure, I just pushed the changes.

rallytime (approved):

PR #46242, opened 2018-02-28
Pass env_vars to pip.freeze

### LINKED ISSUE #46127: pip.installed does not pass env_vars when calling freeze to check if package is already installed
I have a formula for managing graphite installation which requires PYTHONPATH to be set to a custom directory (without changing the `bin_env`). The state looks something like

```
graphite-web:
  pip.installed:
    - name: 'graphite-web{% if graphite_web_version %}=={{graphite_web_version}}{% endif %}'
    - env_vars:
        PYTHONPATH: "/opt/graphite/lib/:/opt/graphite/webapp/"
    - require:
      - file: /opt/graphite
```

Running this state always results in an output like

```
----------
          ID: graphite-web
    Function: pip.installed
        Name: graphite-web==1.1.2
      Result: True
     Comment: There was no error installing package 'graphite-web==1.1.2' although it does not show when calling 'pip.freeze'.
     Started: 16:51:54.604922
    Duration: 6326.189 ms
     Changes:   
----------

```

or something like this with `test=True`

```
----------
          ID: graphite-web
    Function: pip.installed
        Name: graphite-web==1.1.2
      Result: None
     Comment: Python package graphite-web==1.1.2 is set to be installed
     Started: 16:55:14.297774
    Duration: 2413.703 ms
     Changes:   
----------

```

This creates problems in writing dependent states that must only be run in case pip updated the package.

Looking at the code, it looks like `pip freeze` is called before and after installation to check whether the package is installed/needs installation, but while `install` is passed env variables, `freeze` is not.

I am running salt-2017.7.3

I came across a similar bug #37038 but this one is a subset of the issues mentioned there.

redbaron4:
### What does this PR do?

`pip.installed` state calls pip.freeze to check for existing installation and to verify installation post-install. This patch propagates the `env_vars` passed to `pip.installed` to the `pip.freeze` function. It also modifies
existing pip unit tests to test new functionality and adds an integration test that verifies the expected correct behaviour.

Fixes #46127

### Previous Behavior

Calling pip.installed with custom `env_vars` would use the `env_vars` for install but would not use them for checking installation (using `pip freeze`). This lead to the problem of state showing output like `"There were no errors while installing XXX but it does not show up in pip freeze"`

### New Behavior

`env_vars` are also passed to `pip.freeze` so that freeze sees the same `env` as seen by install

### Tests written?

Yes

### Commits signed with GPG?

No


salt/modules/pip.py
```diff
@@ -335,6 +335,21 @@ def _process_requirements(requirements, cmd, cwd, saltenv, user):
     return cleanup_requirements, None
 
 
+def _pass_env_vars_to_cmd(env_vars, cmd_kwargs):
```
terminalmage: This function doesn't pass anything to a command, no command is being run here. We prefer that function names describe what they do.

I also don't like that we force one to pass through a dictionary to be modified, it makes the function less flexible. In my opinion this function should return a dictionary, and its result then added to cmd_kwargs separately. This also makes unit testing easier, FWIW.

I think this function can be entirely replaced with something like this:

```python
def _format_env(env_vars):
    ret = {}
    if env_vars:
        try:
            for key, val in six.iteritems(env_vars):
                if not isinstance(val, six.string_types):
                    val = str(val)
                ret[key] = val
        except AttributeError:
            raise CommandExecutionError(
                'env_vars {0} is not a dictionary'.format(env_vars))
    return ret
```
Then lines 821 and 1030 can be simply changed to:

```python
cmd_kwargs.update(_format_env(env_vars))
```

terminalmage (changes requested):

salt/modules/pip.py
redbaron4: Thanks for looking at this @terminalmage . I have made the changes you pointed out.

terminalmage (approved):

rallytime:
Welcome @redbaron4 and congrats on your first Salt PR! :)

redbaron4:
Thanks @rallytime ! Will this be auto merged forward into develop branch?

terminalmage:
@redbaron4 yes, it will: https://docs.saltstack.com/en/latest/topics/development/contributing.html#merge-forward-process

Issue #46086, opened 2018-02-17
Add support for data formated cluster API credentials overrides in salt.modules.kubernetes (feature request)


jryberg:
### Description of Issue/Question
It's possible to override API settings according to https://docs.saltstack.com/en/latest/ref/modules/all/salt.modules.kubernetes.html

> These settings can be also overrided by adding api_url, api_user, api_password, api_certificate_authority_file, api_client_certificate_file or api_client_key_file parameters when calling a function:

I would like to do exactly this but for the "data" format instead such as:
 - kubernetes.certificate-authority-data: '...'
 - kubernetes.client-certificate-data: '...'
 - kubernetes.client-key-data: '...'

For some reason this format was skipped but it's more usable since it's very easy to store the data as pillar instead of files on each minion. 

### Setup
N/A

### Steps to Reproduce Issue
N/A

### Versions Report
Tested with 2017.7.3

gtmanfred:
That would be great to have.  If you had time to submit a PR we would greatly appreciate it.

Otherwise, I have marked this as a feature request.

Thanks,
Daniel

jryberg:
This is still an issue, please don't close it

gtmanfred:
@saltstack/team-core can you reopen this?

waynew:
Bad bot!

Issue #44701, opened 2017-11-27
Kubernetes python client version 4.0 causes errors


rallytime:
### Description of Issue/Question
When the [Kubernetes python client](https://github.com/kubernetes-incubator/client-python) released version 4.0 last week, it caused some test failures in our test suite:
- unit.modules.test_kubernetes.KubernetesTestCase.test_create_deployments
- unit.modules.test_kubernetes.KubernetesTestCase.test_setup_client_key_file

After some initial investigation, this is because they have changed the way they're handling the the configuration.py attributes. From their CHANGELOG:

> Configuration is not a singleton object anymore. Please use Configuraion.set_default to change default configuration.

This causes issues whenever these attributes are called. Here's a stacktrace from last test listed above, as an example:
```
*** unit.modules.test_kubernetes.KubernetesTestCase.test_setup_client_key_file Tests  ********************************************
 --------  Tests with Errors  ----------------------------------------------------------------------------------------------------
   -> unit.modules.test_kubernetes.KubernetesTestCase.test_setup_client_key_file  ................................................
       Traceback (most recent call last):
         File "/root/SaltStack/salt/tests/unit/modules/test_kubernetes.py", line 161, in test_setup_client_key_file
           config = kubernetes._setup_conn()
         File "/root/SaltStack/salt/salt/modules/kubernetes.py", line 139, in _setup_conn
           kubernetes.client.configuration.host != host or
       AttributeError: 'module' object has no attribute 'host'
```

Salt's kubernetes execution module needs to handle the new configuration structure as well as for older versions.

The stacktrace above is from running the test on develop, but this will need to be fixed on 2017.7 and then merged-forward.

### Setup
Run the kubernetes execution module unit tests after upgrading to the newest version of the kubernetes python client.

### Versions Report
Relevant to the `2017.7` and `develop` branches.

@flavio @brejoc @meaksh @mcalmer Is this something you folks would be willing to look into fixing?

brejoc:
Thanks for the heads up, @rallytime. I've created an internal ticket and asked if we can schedule this for the next sprint.

rallytime:
Awesome! Thank you @brejoc :)

jryberg:
Hi,

Since Kubernetes 1.9 was released the latests version of client-python is 5.0

I'm concerned that SaltStack will not be able to keep up with Kubernetes release schedule with the result that it will be very hard to use SaltStack for Kubernetes management =(

mcalmer:
[...]

Issue #43514, opened 2017-09-14
Add kubeconfig flag to kubernetes module for configuration


SEJeff:
### Description of Issue/Question
The [kubernetes](https://docs.saltstack.com/en/latest/ref/modules/all/salt.modules.kubernetes.html) module is configured with data that is in a [kubeconfig](https://kubernetes.io/docs/concepts/configuration/organize-cluster-access-kubeconfig/) file already. The `kubeconfig` file format is how all kubernetes components are authorized including the `kubectl` client.

Please add support for just passing the kubeconfig path (e.g.: /etc/kubernetes/kubeconfig) and the context name (e.g.: `tectonic`) instead of having to duplicate all of the data inside the kubeconfig in pillar data or in the master config.

### Setup
Latest salt with a working kubernetes cluster.

### Steps to Reproduce Issue
Try to configure the [kubernetes module](https://docs.saltstack.com/en/latest/ref/modules/all/salt.modules.kubernetes.html) by passing in only the path to a kubeconfig AND context to use.

PR #46244, opened 2018-02-28
Kubernetes configured via kubeconfig

mcalmer:
### What does this PR do?

Newer versions of python-kubernetes (>2.0) has changed a lot in the way they want to have the configuration parameters. Also new types of authentication were added and more will come in near future.
All this make the current way to provide authentication data a bit useless.

The only future-proof way seems to be to provide the full `kubeconfig` file and let the python library
do the rest.

Mainly it support new python-kubernetes library as requested in https://github.com/saltstack/salt/issues/44701 .

Configuring via kubeconfig was requested in https://github.com/saltstack/salt/issues/43514 .

Last but not least this PR make it also possible to provide the full kubeconfig as data to overwrite configured kubeconfig like requested in https://github.com/saltstack/salt/issues/46086 .

### Previous Behavior

Configuring the cluster access and authentication was possible via single values:

- kubernetes.user
- kubernetes.password
- kubernetes.api_url
- kubernetes.certificate-authority-data/file
- kubernetes.client-certificate-data/file
- kubernetes.client-key-data/file

### New Behavior

This will be replaced by just configure via `kubeconfig` and `context`

- kubernetes.kubeconfig
- kubernetes.context

### Tests written?

Yes (adapted the existing)

### Commits signed with GPG?

No

### Related PR

2017.7 - add warning about config option change : https://github.com/saltstack/salt/pull/46320

### DIFF
--- a/salt/modules/kubernetes.py
+++ b/salt/modules/kubernetes.py
@@ -8,7 +8,7 @@ Module for handling kubernetes calls.
 
         kubernetes.kubeconfig: '/path/to/kubeconfig'
         kubernetes.kubeconfig-data: '<base64 encoded kubeconfig content'
-        kubernetes.contect: 'context'
+        kubernetes.context: 'context'
 
 These settings can be also overrided by adding `context and `kubeconfig` or
 `kubeconfig_data` parameters when calling a function.
--- a/salt/modules/kubernetes.py
+++ b/salt/modules/kubernetes.py
@@ -10,14 +10,14 @@ Module for handling kubernetes calls.
         kubernetes.kubeconfig-data: '<base64 encoded kubeconfig content'
         kubernetes.context: 'context'
 
-These settings can be also overrided by adding `context and `kubeconfig` or
+These settings can be overridden by adding `context and `kubeconfig` or
 `kubeconfig_data` parameters when calling a function.
 
 The data format for `kubernetes.kubeconfig-data` value is the content of
 `kubeconfig` base64 encoded in one line.
 
 Only `kubeconfig` or `kubeconfig-data` should be provided. In case both are
-provided `kubeconfig` entry is prefered.
+provided `kubeconfig` entry is preferred.
 
 .. code-block:: bash
 
@@ -126,7 +126,7 @@ def _cleanup(**kwargs):
                 os.unlink(kubeconfig)
             except (IOError, OSError) as err:
                 if err.errno != errno.ENOENT:
-                    log.exception('Removing kubeconfig failed: {0}'.format(err))
+                    log.exception(err)
 
 
 def ping(**kwargs):
--- a/salt/modules/kubernetes.py
+++ b/salt/modules/kubernetes.py
@@ -112,6 +112,80 @@ if not salt.utils.platform.is_windows():
     POLLING_TIME_LIMIT = 30
 
 
+def _setup_conn_old(**kwargs):
+    '''
+    Setup kubernetes API connection singleton the old way
+    '''
+    host = __salt__['config.option']('kubernetes.api_url',
+                                     'http://localhost:8080')
+    username = __salt__['config.option']('kubernetes.user')
+    password = __salt__['config.option']('kubernetes.password')
+    ca_cert = __salt__['config.option']('kubernetes.certificate-authority-data')
+    client_cert = __salt__['config.option']('kubernetes.client-certificate-data')
+    client_key = __salt__['config.option']('kubernetes.client-key-data')
+    ca_cert_file = __salt__['config.option']('kubernetes.certificate-authority-file')
+    client_cert_file = __salt__['config.option']('kubernetes.client-certificate-file')
+    client_key_file = __salt__['config.option']('kubernetes.client-key-file')
+
+    # Override default API settings when settings are provided
+    if 'api_url' in kwargs:
+        host = kwargs.get('api_url')
+
+    if 'api_user' in kwargs:
+        username = kwargs.get('api_user')
+
+    if 'api_password' in kwargs:
+        password = kwargs.get('api_password')
+
+    if 'api_certificate_authority_file' in kwargs:
+        ca_cert_file = kwargs.get('api_certificate_authority_file')
+
+    if 'api_client_certificate_file' in kwargs:
+        client_cert_file = kwargs.get('api_client_certificate_file')
+
+    if 'api_client_key_file' in kwargs:
+        client_key_file = kwargs.get('api_client_key_file')
+
+    if (
+            kubernetes.client.configuration.host != host or
+            kubernetes.client.configuration.user != username or
+            kubernetes.client.configuration.password != password):
+        # Recreates API connection if settings are changed
+        kubernetes.client.configuration.__init__()
+
+    kubernetes.client.configuration.host = host
+    kubernetes.client.configuration.user = username
+    kubernetes.client.configuration.passwd = password
+
+    if ca_cert_file:
+        kubernetes.client.configuration.ssl_ca_cert = ca_cert_file
+    elif ca_cert:
+        with tempfile.NamedTemporaryFile(prefix='salt-kube-', delete=False) as ca:
+            ca.write(base64.b64decode(ca_cert))
+            kubernetes.client.configuration.ssl_ca_cert = ca.name
+    else:
+        kubernetes.client.configuration.ssl_ca_cert = None
+
+    if client_cert_file:
+        kubernetes.client.configuration.cert_file = client_cert_file
+    elif client_cert:
+        with tempfile.NamedTemporaryFile(prefix='salt-kube-', delete=False) as c:
+            c.write(base64.b64decode(client_cert))
+            kubernetes.client.configuration.cert_file = c.name
+    else:
+        kubernetes.client.configuration.cert_file = None
+
+    if client_key_file:
+        kubernetes.client.configuration.key_file = client_key_file
+    elif client_key:
+        with tempfile.NamedTemporaryFile(prefix='salt-kube-', delete=False) as k:
+            k.write(base64.b64decode(client_key))
+            kubernetes.client.configuration.key_file = k.name
+    else:
+        kubernetes.client.configuration.key_file = None
+    return {}
+
+
 # pylint: disable=no-member
 def _setup_conn(**kwargs):
     '''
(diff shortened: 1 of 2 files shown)


rallytime:
@mcalmer Since Kubernetes is moving pretty quickly and this change will make it easier to maintain this module moving forward, I am fine with changing these configuration items in a feature release. However, I want to make sure that this is documented appropriately - there should be warnings at the top of the module that these configuration options have changed, and it probably would be good to add this to the `Fluorine` release notes.

It would also be prudent to add some deprecation warnings to the old configuration style in the 2017.7 branch.

mcalmer:
@rallytime do you have an example for such a text? Just to use a similar format and wording.

salt/modules/kubernetes.py
```diff
@@ -106,91 +98,36 @@ def _setup_conn(**kwargs):
...
-    if cert and os.path.exists(cert) and os.path.basename(cert).startswith('salt-kube-'):
-        salt.utils.files.safe_rm(cert)
-    if key and os.path.exists(key) and os.path.basename(key).startswith('salt-kube-'):
-        salt.utils.files.safe_rm(key)
-    if ca and os.path.exists(ca) and os.path.basename(ca).startswith('salt-kube-'):
-        salt.utils.files.safe_rm(ca)
+    if 'kubeconfig' in kwargs:
+        kubeconfig = kwargs.get('kubeconfig')
+        if os.path.exists(kubeconfig) and os.path.basename(kubeconfig).startswith('salt-kubeconfig-'):
+            salt.utils.files.safe_rm(kubeconfig)
```
isbm: No need to `stat` file here, you could just `try/except` and check for `errno.ENOENT`:
```python
try:
    os.unlink(kubeconfig)
except (IOError, OSError) as err:
    if err.errno != errno.ENOENT:
        log.exception()
```

isbm: BTW, interesting that you don't know what is inside `kubeconfig`... :wink:

tests/unit/modules/test_kubernetes.py
```diff
@@ -141,23 +140,49 @@ def test_create_deployments(self):
...
+        Test that provided `kubernetes.kubeconfig` configuration is overwritten
+        by provided kubeconfig_data in the command
+        :return:
+        '''
+        def settings(name, value=None):
+            data = {
+                'kubernetes.kubeconfig': '/home/testuser/.minikube/kubeconfig.cfg',
+                'kubernetes.context': 'minikube'
+            }
+            return data.get(name, value)
```
isbm: Why copypaste the same thing all over again? It can be just a static method and you could do something like:

```python
    @staticmethod
    def kube_settings(name, value=None):
        data = {....}
        return data.get(name, value)

    def test......:
        with mock_kubernetes_library() ....:
            with patch.dict(kubernetes.__salt__, {'config.option': Mock(side_effect=KubernetesTestCase.kube_settings)}):
                ...
```
...and then just reuse it twice. I am not fan with multiple nested `with` statements (which you still can have multiple conditions in just _one statement_) and I would use `@patch` decorator instead, but for now it is OK.

Please refactor this.

tests/unit/modules/test_kubernetes.py
```diff
@@ -141,23 +140,49 @@ def test_create_deployments(self):
...
             return data.get(name, value)
 
-        with patch.dict(kubernetes.__salt__, {'config.option': Mock(side_effect=settings)}):
-            config = kubernetes._setup_conn()
-            self.assertEqual(
-                settings('kubernetes.client-key-file'),
-                config['key_file'],
-            )
+        with mock_kubernetes_library() as mock_kubernetes_lib:
+            with patch.dict(kubernetes.__salt__, {'config.option': Mock(side_effect=settings)}):
```
isbm: It can be just one statement, BTW.

isbm (changes requested):
@mcalmer could you please take a look at these changes? Thanks!

salt/modules/kubernetes.py
mcalmer: No, this is not working. I only want to remove it when it is a temp file created by the module itself.
So just unlink is not an option. I can either just check if I have something in kubeconfig or call
path.exists().  I like to check if kubeconfig is a "path" and if it exists before I remove it.

salt/modules/kubernetes.py
```diff
@@ -106,91 +98,36 @@ def _setup_conn(**kwargs):
...
-    username = __salt__['config.option']('kubernetes.user')
-    password = __salt__['config.option']('kubernetes.password')
-    ca_cert = __salt__['config.option']('kubernetes.certificate-authority-data')
-    client_cert = __salt__['config.option']('kubernetes.client-certificate-data')
-    client_key = __salt__['config.option']('kubernetes.client-key-data')
-    ca_cert_file = __salt__['config.option']('kubernetes.certificate-authority-file')
-    client_cert_file = __salt__['config.option']('kubernetes.client-certificate-file')
-    client_key_file = __salt__['config.option']('kubernetes.client-key-file')
+    kubeconfig = __salt__['config.option']('kubernetes.kubeconfig')
+    kubeconfig_data = __salt__['config.option']('kubernetes.kubeconfig-data')
```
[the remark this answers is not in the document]
mcalmer: Your suggestion is not directly working like it should. But I modify it to go in that direction.

salt/modules/kubernetes.py
```diff
@@ -121,8 +121,12 @@ def _setup_conn(**kwargs):
...
     if 'kubeconfig' in kwargs:
         kubeconfig = kwargs.get('kubeconfig')
-        if os.path.exists(kubeconfig) and os.path.basename(kubeconfig).startswith('salt-kubeconfig-'):
-            salt.utils.files.safe_rm(kubeconfig)
+        if kubeconfig and os.path.basename(kubeconfig).startswith('salt-kubeconfig-'):
+            try:
+                os.unlink(kubeconfig)
+            except (IOError, OSError) as err:
+                if err.errno != errno.ENOENT:
+                    log.exception('Removing kubeconfig failed: {0}'.format(err))
```
isbm: In Salt core normally we just so: `log.exception(err)` —  and it will do all the job for you.

isbm (changes requested):

isbm:
@mcalmer justa little nitpick with the `log.exception(...)` but the rest is great and much simpler now! Thanks.

salt/modules/kubernetes.py
```diff
@@ -6,32 +6,30 @@
...
+The data format for `kubernetes.kubeconfig-data` value is the content of
+`kubeconfig` base64 encoded in one line.
 
-The data format for `kubernetes.*-data` values is the same as provided in `kubeconfig`.
-It's base64 encoded certificates/keys in one line.
-
-For an item only one field should be provided. Either a `data` or a `file` entry.
-In case both are provided the `file` entry is prefered.
+Only `kubeconfig` or `kubeconfig-data` should be provided. In case both are
+provided `kubeconfig` entry is prefered.
```
isbm: Typo: preferred.

salt/modules/kubernetes.py
```diff
@@ -6,32 +6,30 @@
...
-        kubernetes.client-certificate-data: '....n
-        kubernetes.client-key-data: '...'
-        kubernetes.certificate-authority-file: '/path/to/ca.crt'
-        kubernetes.client-certificate-file: '/path/to/client.crt'
-        kubernetes.client-key-file: '/path/to/client.key'
+        kubernetes.kubeconfig: '/path/to/kubeconfig'
+        kubernetes.kubeconfig-data: '<base64 encoded kubeconfig content'
+        kubernetes.context: 'context'
 
+These settings can be also overrided by adding `context and `kubeconfig` or
```
isbm: overridden

isbm (changes requested):

isbm (approved):
@mcalmer thanks!

rallytime:
@mcalmer I don't have any doc examples handy off the top of my head, but something along the lines of "The following configuration options have been removed in favor of x, y, z" and list out which items were removed (and why they were removed) and what to use instead. For example, if someone is using 2017.7.x or 2018.3.x and they upgrade and now their modules/states don't work, they need to know why and how to remediate the issue. This needs to be added to the Fluorine release notes.

I also would like to see a PR implementing the `warn_util` warnings in the 2017.7 branch so people will know that the old config options will be removed in Fluorine.

mcalmer:
@rallytime I have adapted the warning message here, but I am not sure if using warn_until in older versions is a good idea. These messages would just pollute the logfiles with tons of warnings and
the admin has no chance to remove them because the new options are not available in the old version.

So I would like to go with just a message in the docs.

An alternative would be to backport the feature and try to make both options available - let's say in oxygen. Then warn_until would make sense because the admin could switch to the new config style and prevent the warnings.

What do you prefer?

mcalmer:
@jryberg - yes. Overwrite is still possible. But now you use a full kubeconfig instead of single values out of it.

isbm:
@mcalmer I would say, we should have indeed some sort of warning and a proper deprecation process. For this we actually using [deprecation mechanism](https://github.com/saltstack/salt/blob/develop/salt/utils/decorators/__init__.py#L309-L322) and it is just a decorator on top of a function.

mcalmer:
@isbm - I fear it will not work as the signatures stay unchanged for all methods. Only configuration in the background will be different and you can provide different kwargs but they are optional.

isbm:
@mcalmer technically, surely it will. It is just in this case it is equivalent to:

```python
def foo(**kwargs):
    ... # all the real code

_foo = foo
```

The problem is here that due to the nature of the `**kwargs` that can be _anything_ inside, it does not matter which you use: new or old, since it is technically the same function reference. :disappointed: So the workaround would be a bit of refactoring:

```python
def _version_blocker(func, **kwargs):
    '''
    This will raise an exception if kubernetes is more than 2.0, for example.
    '''
    if _get_version_of_kubernetes() > (2, 0):  # Check what API we are linked to
        raise CommandExecutionError('Sticky bit become loose!')
    return func(**kwargs)

def foo(**kwargs):
    ... # all the real code

_foo = lambda **kwargs: _version_blocker(foo, **kwargs)

def bar(**kwargs):
    ... # all the real code

_bar = lambda **kwargs: _version_blocker(bar, **kwargs)
```

Well, you've got the idea. :wink:

rallytime:
@mcalmer Regarding your questions about a proper deprecation process - what if we keep the old way of doing things for now with a `warn_util` message when the old configs are found, and then users can  update over the next couple of feature releases and then we can remove the old behavior after the deprecation period has passed. This is what we usually do and I should have suggested doing that the first time around. My apologies for the run-around on that!

mcalmer:
@rallytime - It is not easy as new python-kubernetes client lib do not support this old way.
So I bring just the old code back and try to catch all exceptions which might get thrown because of invlaud use of the API. Old configuration can only be used upto v2.0.0 . New style works with all.

rallytime (approved):
👍

rallytime:
re-run ubuntu

Issue #46202, opened 2018-02-26
multimaster failover not working 


frogunder:
### Description of Issue/Question
In a multimaster failover setup on 2018.3.0rc1(oxygen.rc1) the minions stop.
It seems to be working fine/correctly for a little while. Tried a couple of times and seems when one master stops it starts to fail. 
See this in the minion log:
```
2018-02-26 18:02:15,266 [salt.minion      :204 ][WARNING ][16492] Master ip address changed from 10.1.51.25 to 10.1.51.24
2018-02-26 18:02:15,616 [salt.minion      :433 ][ERROR   ][16492] Exception unsupported operand type(s) for -: 'str' and 'datetime.datetime' occurred in scheduled job
2018-02-26 18:02:16,613 [salt.minion      :433 ][ERROR   ][16492] Exception unsupported operand type(s) for -: 'str' and 'datetime.datetime' occurred in scheduled job
2018-02-26 18:02:17,611 [salt.minion      :433 ][ERROR   ][16492] Exception unsupported operand type(s) for -: 'str' and 'datetime.datetime' occurred in scheduled job
2018-02-26 18:02:18,611 [salt.minion      :433 ][ERROR   ][16492] Exception unsupported operand type(s) for -: 'str' and 'datetime.datetime' occurred in scheduled job
2018-02-26 18:02:19,611 [salt.minion      :433 ][ERROR   ][16492] Exception unsupported operand type(s) for -: 'str' and 'datetime.datetime' occurred in scheduled job
2018-02-26 18:02:20,288 [salt.minion      :204 ][WARNING ][16492] Master ip address changed from 10.1.51.24 to 10.1.51.25
2018-02-26 18:02:20,611 [salt.minion      :433 ][ERROR   ][16492] Exception unsupported operand type(s) for -: 'str' and 'datetime.datetime' occurred in scheduled job
2018-02-26 18:02:21,611 [salt.minion      :433 ][ERROR   ][16492] Exception unsupported operand type(s) for -: 'str' and 'datetime.datetime' occurred in scheduled job
2018-02-26 18:02:22,610 [salt.minion      :433 ][ERROR   ][16492] Exception unsupported operand type(s) for -: 'str' and 'datetime.datetime' occurred in scheduled job
2018-02-26 18:02:23,619 [salt.minion      :433 ][ERROR   ][16492] Exception unsupported operand type(s) for -: 'str' and 'datetime.datetime' occurred in scheduled job
[...]

PR #46267, opened 2018-02-28
[2018.3.0rc1] Don't include internal schedule options

garethgreenaway:
### What does this PR do?
Filter out the internal schedule options when persisting the schedule to disk.  Some of these are "special" types and do not represent when when written out, eg. datetime objects. Also ensuring that after all the calculations, the seconds value is an int and does not include partial second bits.

#46202 

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/utils/schedule.py
+++ b/salt/utils/schedule.py
@@ -134,7 +134,8 @@ class Schedule(object):
 
     def _get_schedule(self,
                       include_opts=True,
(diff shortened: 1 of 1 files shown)


s0undt3ch (approved):

DmitryKuzmenko (approved):
:+1:

PR #46272, opened 2018-03-01
Remove dup start events

UtahDave:
### What does this PR do?
Remove duplicate `minion_start` and `syndic_start` events.

None

### Previous Behavior
When the minion started 2 events with the following event tags were sent up the event bus:

`minion_start`
`salt/minion/<minion id>/start`

The same behavior happened for the salt syndic daemon.

### New Behavior
Only `salt/minion/<minion id>/start` event is sent on minion startup.

The same behavior now happens for the salt syndic daemon.

### Tests written?

No, but I verified that current tests were only testing the new style event tags.
The old style event tags were untested. This PR removes a bit of untested code.

### Commits signed with GPG?

Yes

### DIFF
--- a/doc/ref/configuration/minion.rst
+++ b/doc/ref/configuration/minion.rst
@@ -208,6 +208,28 @@ minion event bus. The value is expressed in bytes.
 
     max_event_size: 1048576
 
+.. conf_minion:: enable_legacy_startup_events
+
+``enable_legacy_startup_events``
+--------------------------------
+
+.. versionadded:: Fluorine
+
+Default: ``True``
+
+When a minion starts up it sends a notification on the event bus with a tag
+that looks like this: `salt/minion/<minion_id>/start`. For historical reasons
+the minion also sends a similar event with an event tag like this:
+`minion_start`. This duplication can cause a lot of clutter on the event bus
+when there are many minions. Set `enable_legacy_startup_events: False` in the
+minion config to ensure only the `salt/minion/<minion_id>/start` events are
+sent. Beginning with the `Neon` Salt release this option will default to
+`False`
+
+.. code-block:: yaml
+
+    enable_legacy_startup_events: True
+
 .. conf_minion:: master_failback
 
 ``master_failback``
--- a/doc/topics/releases/fluorine.rst
+++ b/doc/topics/releases/fluorine.rst
@@ -3,3 +3,22 @@
 ======================================
 Salt Release Notes - Codename Fluorine
 ======================================
+
+
+Minion Startup Events
+---------------------
+
+When a minion starts up it sends a notification on the event bus with a tag
+that looks like this: `salt/minion/<minion_id>/start`. For historical reasons
+the minion also sends a similar event with an event tag like this:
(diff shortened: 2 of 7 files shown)


terminalmage (commented):
The old events are legacy and had been left for backward compatibility. Have we announced a deprecation plan for removing them?

CC: @cachedout

UtahDave:
Are these test failures happening in other PRs on develop?

rallytime:
Yeah, they are. I agree with @terminalmage though - I don't think we can just remove these kinds of events without some warning or putting these on a deprecation path.

UtahDave:
OK. What kind of warning do we want? A section in the release notes?

terminalmage:
@UtahDave yeah, we can put one in the 2018.3.0 release notes and give it the standard 2-release deprecation path, which would mean we'd target the Neon release for final removal.

We could also gate these legacy events behind a config option (defaulting that option to True), which would allow people who operate at large scale and are concerned with the extra overhead incurred by the legacy events to disable them until we can fully deprecate them.

UtahDave:
OK. I've added a gate with a new minion config option instead of just removing the events. I've added a note in the fluorine release notes and opened a github issue to remind us to change the default to False when the Neon release of Salt is made.

https://github.com/saltstack/salt/issues/46336

doc/ref/configuration/minion.rst
```diff
@@ -208,6 +208,28 @@ minion event bus. The value is expressed in bytes.
...
+`minion_start`. This duplication can cause a lot of clutter on the event bus
+when there are many minions. Set `enable_legacy_startup_events: False` in the
+minion config to ensure only the `salt/minion/<minion_id>/start` events are
+sent. Beginning with the `Neon` Salt release this option will default to
+`False`
+
+.. code-block:: yaml
+
+    enable_legacy_startup_events: True
+
```
terminalmage: :heart: 

Love that you added docs too :smile:

doc/topics/releases/fluorine.rst
```diff
@@ -3,3 +3,22 @@
...
+`minion_start`. This duplication can cause a lot of clutter on the event bus
+when there are many minions. Set `enable_legacy_startup_events: False` in the
+minion config to ensure only the `salt/minion/<minion_id>/start` events are
+sent.
+
+The new :conf_minion:`enable_legacy_startup_events` minion config option
+defaults to ``True``, but will be set to default to ``False`` beginning with
+the Neon release of Salt.
+
+The Salt Syndic currently sends an old style  `sydic_start` event as well. The
```
terminalmage: should be `syndic_start` here.

terminalmage (changes requested):
One minor spelling error in the release notes, otherwise this is GTM.

doc/topics/releases/fluorine.rst
UtahDave: Ah, great catch.  I'll fix that right now.

doc/ref/configuration/minion.rst
UtahDave: :)  More docs is always good!

rallytime (approved):

PR #46280, opened 2018-03-01
Remove unneeded checks for binaries in SmartOS modules

jasperla:
### What does this PR do?

This tidies the actual command string building and makes the code more readable in general. The required binaries are not likely to disappear after `__virtual__` and this is the common pattern.

Also use correct nomenclature when referring to 'compute nodes'.

It has no function change.

### Tests written?

No.

It was tested against a SmartOS compute node with `salt-ssh`.

### Commits signed with GPG?

No


sjorge (approved):
LGTM

sjorge:
Nice catch on the typo... that got copied around.

rallytime (approved):

PR #46284, opened 2018-03-01
Clouds vmware fqdn fix

ghost:
### What does this PR do?

Modifies the way a FQDN is handled in the vmware cloud provider.

None

### Previous Behavior

The module previously handled a FQDN by first splitting it on the first occurrence of a character which is neither a non-word character nor '-', after that the module proceeded to try to split the resulting string, which is guaranteed to have no periods, again from the first occurrence of the period.

The end result was that 'a.b.c.d.e', e.g., ended up being properly split into the host part, 'a', while failing to extract the domain part, 'b.c.d.e'.

### New Behavior

A FQDN gets properly handled ie. split into a hostname and a domain, which are then used appropriately. E.g. 'a.b.c.d.e' ends being parsed as 'a' for the hostname and 'b.c.d.e' for the domain.

### Tests written?

No.

### Commits signed with GPG?

No.


rallytime:
@samilaine Looks like you've got plenty of extra commits in here. Can you rebase this to 2016.11 if that is your intended branch?

samilaine:
Ahem. Another PR coming in.

Issue #44758, opened 2017-11-30
close_fds=True causes cmd.run very slow


angeloudy:
### Description of Issue/Question
salt is running very slow on my system.
It takes 25 seconds to run a test.ping.
after some cleanup on the system, I reduced it to 12 seconds, which, is still quite slow.
So I start tracking down the code.
And I found the the following code in /usr/local/lib/python2.7/site-packages/salt/utils/timed_subprocess.py is causing the problem.

```
if self.timeout and not isinstance(self.timeout, (int, float)):
            raise salt.exceptions.TimedProcTimeoutError('Error: timeout {0} must be a number'.format(self.timeout))

        try:
            self.process = subprocess.Popen(args, **kwargs)
        except TypeError:
```
It takes half a second to run subprocess.Popen, which is insane.
But when I run subprocess.Popen in a separate test script, it takes only 70ms

```
time python -c "import subprocess;subprocess.Popen(['/sbin/sysctl', '-n', 'hw.ncpu'])"
24

real	0m0.072s
user	0m0.040s
sys	0m0.039s

```
By adding some debug and tracing code finally I found that it was 'close_fds' in **kargs that caused it to run slow. If I changed close_fds to False, it will run very fast.
```
[...]

PR #46286, opened 2018-03-01
Bugfix: Py2 closing non-existing file descriptors slows down Salt run

isbm:
### What does this PR do?

Bugfix. This affects all Salt versions, running Python 2.7+ version. This bug _does not_ affects any Salt version as long as it is running Python 3+ version.

Salt by default comes with `close_fds=True` on `subprocess.Popen`, which causes it running very slow. Related issue: https://github.com/saltstack/salt/issues/44758

The cause of this is a pretty old issue of Python-2 stdlib: https://bugs.python.org/issue1663329 that was never properly fixed. Long story short: Python 3 is closing only file descriptors that are actually existing, while Python 2 is closing by a range, which is defined by `ulimit` of the environment. Closing non-existing file descriptors takes a while for _each_ call of the external process.

Please take a look.

@terminalmage 
@thatch45 
@cachedout 
@Ch3LL 
@rallytime 
@flavio 
@dincamihai 
@meaksh 
@brejoc 

### Previous Behavior

:cry: 

### New Behavior

:smile: 

### Tests written?

Yes

### Commits signed with GPG?

No


salt/utils/subprocess.py
```diff
@@ -4,14 +4,50 @@
...
+        def _close_fds(self, but):
+            '''
+            Close only file descriptors that are needed.
+            The issue is not new: https://bugs.python.org/issue1663329
+
+            :param but:
+            :return:
+            '''
+            maxfds = len(os.listdir('/proc/{}/fd'.format(os.getpid())))
+            log.debug('Closing {} file descriptors'.format(maxfds))
```
terminalmage: We should switch this to %s replacement

```
log.debug('Closing %s file descriptors', maxfds)
```

salt/utils/subprocess.py
```diff
@@ -4,14 +4,50 @@
...
+    if sys.platform.startswith('linux') and six.PY2:
+        def _close_fds(self, but):
+            '''
+            Close only file descriptors that are needed.
+            The issue is not new: https://bugs.python.org/issue1663329
+
+            :param but:
+            :return:
+            '''
+            maxfds = len(os.listdir('/proc/{}/fd'.format(os.getpid())))
```
terminalmage: We should have a fallback to the existing behavior for corner cases where /proc is not mounted.

terminalmage (changes requested):
This is excellent work, just a couple things that I caught that should be changed. We should also be blacklisting `subprocess.Popen` via pylint. We've been able to do this for a while now but we've never done it yet. If you would like to steal this large block comment from a pending PR, and just replace the `blacklisted-functions=` line with `blacklisted-functions=subprocess.Popen=salt.utils.subprocess.FdPopen`, that will blacklist Popen via pylint. If there are any legitimate places where we _call_ Popen (the place we use it as a parent class should be fine), then you can do an inline disable with `# pylint: disable=blacklisted-function`.

salt/utils/subprocess.py
```diff
@@ -4,12 +4,49 @@
...
+        def _close_fds(self, but):
+            '''
+            Close only file descriptors that are needed.
+            The issue is not new: https://bugs.python.org/issue1663329
+
+            :param but:
+            :return:
+            '''
+            proc_path = '/proc/{}/fd'.format(os.getpid())
+            maxfds = len(os.listdir(proc_path)) if os.path.exists(proc_path) else subprocess.MAXFD
```
terminalmage: This generates two stat syscalls for each command, it should use a try/except instead.

```python
try:
    maxfds = len(os.listdir(proc_path))
except OSError:
    maxfds = subprocess.MAXFD
```

EAFP.

terminalmage (changes requested):

isbm:
@terminalmage fixd!

salt/utils/subprocess.py
```diff
@@ -4,12 +4,53 @@
...
+                maxfds = len(os.listdir(proc_path))
+            except OSError:
+                maxfds = subprocess.MAXFD
+
+            log.debug('Closing %s file descriptors', maxfds)
+            if hasattr(os, 'closerange'):
+                os.closerange(3, but)
+                os.closerange(but + 1, maxfds)
+            else:
+                for i in range(3, maxfds):
```
DmitryKuzmenko: `salt.ext.six.moves.range` should be used here instead of `range`

salt/utils/subprocess.py
```diff
@@ -4,12 +4,53 @@
...
+            '''
+            Close only file descriptors that are needed.
+            The issue is not new: https://bugs.python.org/issue1663329
+
+            :param but:
+            :return:
+            '''
+            proc_path = '/proc/{}/fd'.format(os.getpid())
+            try:
+                maxfds = len(os.listdir(proc_path))
```
DmitryKuzmenko: `maxfds` calculated in this way will not represent the higher opened fd number as it's desired if I'm not wrong. To do it you have to look for max integer in the list. Simple example from a live Chromium process from my PC:
```
>>> ls = os.listdir('/proc/31554/fd'); len(ls); max([int(i) for i in ls])
157
173
```

DmitryKuzmenko (changes requested):

salt/utils/subprocess.py
terminalmage: This is a good point.

salt/utils/subprocess.py
terminalmage: Yeah, this makes range use xrange on py2 so that it's a generator. You can import it like so:

```python
```

isbm: @DmitryKuzmenko but this is what I am doing, if you had checked the imports part of this file:
```python
from salt.ext.six.moves import range
```

salt/utils/subprocess.py
isbm: No, this is not correct. Your Chrome is _currently_ has opened 157 files, while overall it _was opened_ 173. We are not interested how many files some session had opened in its life history, but we are only interested what we have to close after quit. So my code is correct here.

salt/utils/subprocess.py
isbm: @terminalmage I have no pylint disable there though. Is it needed?

DmitryKuzmenko: Ah. Sorry, I missed this fact. Just because usually we do `from salt.ext import six` and use `six.somehting` to not mess it with builtins.

salt/utils/subprocess.py
isbm: @DmitryKuzmenko I double-checked, I think we are both wrong: you will be hammering extra-fd's that are not there anymore, and I will be leaving some (which is not entirely bad, because kernel will kill them anyways). The correct fix is to ditch the `closerange` entirely and simply iterate over, like so:

```python
for fd in [int(x) for x in os.listdir(proc_path)]:
    try:
        os.close(fd)
    ...
```

DmitryKuzmenko: @isbm I took your function `_close_fds` and wrote a test for it. The test results shows that I'm right here. Sorry.

The test:

```python
#!/bin/env python

import os
import subprocess


def _close_fds(but):
    '''
    Close only file descriptors that are needed.
    The issue is not new: https://bugs.python.org/issue1663329

    :param but:
    :return:
    '''
    proc_path = '/proc/{}/fd'.format(os.getpid())
    try:
        maxfds = len(os.listdir(proc_path))
    except OSError:
        maxfds = subprocess.MAXFD

    print('Closing file descriptors from 3 to {}'.format(maxfds))
    if hasattr(os, 'closerange'):
        if 3 <= but <= maxfds:
            os.closerange(3, but)
            os.closerange(but + 1, maxfds)
        else:
            os.closerange(3, maxfds)
    else:
        for i in range(3, maxfds):
            if i != but:
                try:
                    os.close(i)
                except Exception:
                    pass


proc_path = '/proc/{}/fd'.format(os.getpid())
print('On start:           {}'.format(os.listdir(proc_path)))
files = []
for i in range(10):
    files.append(open('/tmp/f{}'.format(i), 'w'))
print('Opened 10 files:    {}'.format(os.listdir(proc_path)))
for i in range(5):
    print('Closing fd#{}'.format(files[i].fileno()))
    files[i].close()
print('After closing 3-7:  {}'.format(os.listdir(proc_path)))
_close_fds(0)
print('After _close_fds(): {}'.format(os.listdir(proc_path)))
```

Test output:

```
$ python test.py 
On start:           ['0', '1', '2', '3']
Opened 10 files:    ['0', '1', '2', '3', '4', '5', '6', '7', '8', '9', '10', '11', '12', '13']
Closing fd#3
Closing fd#4
Closing fd#5
Closing fd#6
Closing fd#7
After closing:      ['0', '1', '2', '3', '8', '9', '10', '11', '12']
Closing file descriptors from 3 to 9
After _close_fds(): ['0', '1', '2', '3', '9', '10', '11', '12']
```

As you can see after opening 10 files we have 13 files opened (14th is opened by `os.listdir` just while `os.listdir` works, I don't calc it here). After I close 5 files in the middle of this range we have 8. `_close_fds` tries to close range 3-9 (9 is excluded by `closerange` by design). So we have 9 to 12 opened as a result.

Also please take note that I've fixed how `_close_fds` handles `but` in case it use `closerange`. It's needed if `0` is passed meaning "no but, close all". In your implementation it closes 0-2 descriptors too if `0` passed as `but`.

DmitryKuzmenko: @isbm didn't see your comment. Yes, that is the best and simplest way as I see.

isbm: @DmitryKuzmenko I found exactly the same thing. :wink: So thanks for double-checking this as well. I forgot that `os.close()` is by descriptor num, and hence we need exact ones.

damon-atkins: Why not reduce the file limit.    ulimit setting being high and MAX_OPEN_FILES being high is only required for network connections.   Why not have the rest of salt stay at 256 fds and only the process which accepts connections have higher connection count. Might need to use dup2.  Have not looked at how salt handles in coming connections.

isbm:
@DmitryKuzmenko how about now? (tests fixed too)

salt/utils/subprocess.py
```diff
@@ -4,12 +4,50 @@
...
+            '''
+            Close only file descriptors that are needed.
+            The issue is not new: https://bugs.python.org/issue1663329
+
+            Do *not* use os.closerange (!)
+
+            :param but:
+            :return:
+            '''
+            proc_path = '/proc/{}/fd'.format(os.getpid())
```
damon-atkins: See also https://stackoverflow.com/questions/899038/getting-the-highest-allocated-file-descriptor?answertab=active#tab-top

damon-atkins: Check if 
FreeBSD, Darwin or OS X check /dev/fd/
else if exist dir /proc/self/fd 
else if exist dir  dir /proc/<pid>/fd

/proc/pid/fd/ or /proc/self/fd/ (Linux, Solaris, AIX, Cygwin, NetBSD)
(AIX does not support "self")

/dev/fd/ (FreeBSD, Darwin, OS X)

damon-atkins (changes requested):

salt/utils/subprocess.py
damon-atkins: @twangboy what is the os_family for FreeBSD, Darwin or OS X

isbm: @damon-atkins while thanks for the valuable info about other OSes, unfortunately I am not going to support them in this PR as I have none of them around. However, feel free to do another PR on top of this, which will do the following:

- Add another check for the OS platform
- Get an appropriate path of FDs
- Include unit tests for that

I am happy to assist you in your PR with any questions you will have.

isbm:
@DmitryKuzmenko is there a chance for you to look at the latest updates, please?

terminalmage (approved):

rallytime:
Hi @isbm :)

There are some related test failures here:

- Lint: https://jenkins.saltstack.com/job/PR/job/salt-pr-lint-n/19782/violations/file/salt/pillar/django_orm.py/

And the following tests in PY3 are related too:
- `unit.modules.test_openscap`
- `unit.modules.test_cmdmod`
- `unit.renderers.test_gpg`

Tests in https://jenkins.saltstack.com/job/PR/job/salt-pr-linode-cent7-py3/2783/#showFailuresLink

DmitryKuzmenko (approved):

DmitryKuzmenko:
@isbm now I'm good. Thank you for that work!

rallytime:
re-run py

rallytime:
re-run py

rallytime:
Something isn't right here, but I am not sure what. None of the test runs are running all of the tests.

rallytime:
re-run py

rallytime:
So the Py2 tests seem to be going slow specifically when the integration _module_ tests are running. Looking through the console logs, each test is taking about 6 minutes. The tests that run before that (integration runners, wheel, etc.) look fine.

Then eventually it hits the 3 hour timeout and Jenkins kills the tests. This is not happening on other PRs or on the HEAD of the branch, so something in this PR is causing something to slow down the module integration tests.

isbm:
re-run py

rallytime:
@isbm Ah, that's looking much better. I know some of these test failures are happening on the branch. But there are some new ones here that I am not seeing on the branch or in other PRs to develop:

- `integration.ssh.test_state.*`
- `integration.ssh.test_pillar.*`
- `integration.ssh.test_master.*`

Can you look into those failures? There's also a tiny lint error, too. 😄

gtmanfred (commented):
needs a rebase now

rallytime:
Hi @isbm - Are you planning on coming back to this one? Or should we close it for now until you're able to come back to it?

rallytime:
I am going to close this for now until you have a chance to come back to this and fix up the test failures. There are quite a few files that are conflicting so this will need a rebase as well. Let us know when you're ready and we can come back to this and review it again. :)

mattp-:
@isbm I'm going to be resuscitating this next week 👍

salt/utils/nb_popen.py
```diff
@@ -37,7 +38,37 @@
...
+    if sys.platform.startswith('linux') and salt.ext.six.PY2:
+        def _close_fds(self, but):
+            '''
+            Close only file descriptors that are needed.
+            The issue is not new: https://bugs.python.org/issue1663329
+
+            :param but:
+            :return:
+            '''
+            maxfds = len(os.listdir('/proc/{}/fd'.format(os.getpid())))
```
ferringb: This isn't actually safe; the number of fds open isn't guaranteed to equate to the highest FD open.
This should instead be:
`maxfds = max(int(x) for x in os.listdir('/proc/{}/fd'.format(os.getpid()))` with an appropriate check for if there is somehow no FDS open which would result in an exception from `max([])`

PR #46290, opened 2018-03-01
Add boto_rds.db_cluster_present

johnclyde:
### What does this PR do?
Fill in missing functionality to add an RDS db cluster. 
I couldn't find a way to use Aurora RDS without adding this feature to add a cluster.
### Previous Behavior
There didn't seem to be a way to add a db cluster with salt.
### New Behavior
Now you can add a db cluster in the intuitive fashion.
### Tests written?
No
### Commits signed with GPG?
No


rallytime (commented):
Thanks for adding these functions @johnclyde! Can you make sure each one has a `.. versionadded:: Fluorine` tag?

These new functions will need some unit tests added as well.

salt/states/boto_rds.py
```diff
@@ -733,3 +733,69 @@ def parameter_present(name, db_parameter_group_family, description, parameters=N
...
+        A dict with region, key and keyid, or a pillar key (string) that
+        contains a dict with region, key and keyid.
+    '''
+    ret = {'name': name,
+           'result': True,
+           'comment': '',
+           'changes': {}
+           }
+    res = __salt__['boto_rds.db_cluster_exists'](name=name, tags=tags, region=region, key=key,
+                                                 keyid=keyid, profile=profile)
```
ryan-lane: You're not doing any error checking here. You should be seeing if error is in the results, and if so, displaying the error message back and making the result as False.

johnclyde: Hi Ryan it looks like the error checking code is the same thing everywhere - do you know if it's been factored out in modules that receive a bit more love than boto_rds?

johnclyde:
I will probably be handing this work over to another developer on my team if it is blocked on tests and error handling. This is a heads up in case you see a new person start adding commits to the PR.

nhblong:
Hi,
The error checking and unittest cases have been added.
Could you please review? Thanks.

salt/states/boto_rds.py
nhblong: Hi Ryan, I have add the error checking as your suggestion.

rallytime:
There are a couple of lint errors here: https://jenkins.saltstack.com/job/PR/job/salt-pr-lint-n/19863/violations/file/tests/unit/states/test_boto_rds.py/

salt/modules/boto_rds.py
```diff
@@ -197,6 +197,30 @@ def parameter_group_exists(name, tags=None, region=None, key=None, keyid=None,
         return resp
 
 
+def db_cluster_exists(name, tags=None, region=None, key=None, keyid=None,
+                           profile=None):
+    '''
+    Check to see if an RDS db cluster exists.
+
+      .. versionadded:: Fluorine
+    CLI example::
```
rallytime: Can you change this to:
```
CLI Example:

.. code-block: bash
```
Instead of `CLI example::`? Our docs have some handy formatting that looks for the code-block type of CLI examples.

salt/modules/boto_rds.py
```diff
@@ -447,6 +471,43 @@ def create_parameter_group(name, db_parameter_group_family, description,
...
 
 
+def create_db_cluster(name, engine, master_username, master_user_password,
+                      tags=None, region=None, key=None, keyid=None,
+                      profile=None):
+    '''
+    Create an RDS db cluster
+
+      .. versionadded:: Fluorine
+    CLI example to create an RDS db cluster::
```
rallytime: Same comment here about doc formatting.

rallytime (commented):
I just have a couple of small comments, otherwise this looks good to me. Thank you for adding the error checking and the tests!

salt/modules/boto_rds.py
johnclyde: This seems reasonable but it's worth noting that what's in our PR is consistent with every other block inside this particular file (which granted, is probably way out of date).

rallytime (approved):

rallytime:
@ryan-lane How does this look to you now?

tkwilliams:
@nhblong  Looking at the code, I don't see any allowance for API Throttling retries.  This occurs a lot, often even at relatively low volumes, so it's a good thing to check for and special-case in the execution module `except` clauses.  If you take a look at e.g. .../modules/boto_elb.py you'll see some examples of the sort of thing I mean.  Something on the order of:

```    except boto.exception.BotoServerError as error:
            if error.error_code == 'Throttling':
                log.debug('Throttled by AWS API, will retry in 5 seconds.')
                time.sleep(5)
                retries -= 1
                continue
            log.error('Error fetching config for ELB {0}: {1}'.format(name, error.message))
            log.error(error)
```
... or similar :)

salt/modules/boto_rds.py
```diff
@@ -197,6 +197,33 @@ def parameter_group_exists(name, tags=None, region=None, key=None, keyid=None,
...
+
+        salt myminion boto_rds.db_cluster_exists mydbcluster \
+                region=us-east-1
+    '''
+    conn = _get_conn(region=region, key=key, keyid=keyid, profile=profile)
+
+    try:
+        rds = conn.describe_db_clusters(DBClusterIdentifier=name)
+        return {'exists': bool(rds), 'error': None}
+    except ClientError as e:
```
tkwilliams: I would recommend some sort of special handling for AWS API throttling errors - they're extremely common, and the general best practice is to wait-retry the call (a reasonable number of times) before failing the call entirely.

Something like this would be along the lines of what I mean:
```            if error.error_code == 'Throttling':
                log.debug('Throttled by AWS API, will retry in 5 seconds.')
                time.sleep(5)
                retries -= 1
                continue
            log.error('Error fetching config for ELB {0}: {1}'.format(name, error.message))
            log.error(error)
```  Check out the boto_ec2 or various other modules (grep for "Throttled") for good examples.

File: salt/modules/boto_rds.py [hunk 5]
```diff
@@ -447,6 +474,46 @@ def create_parameter_group(name, db_parameter_group_family, description,
...
+        conn = _get_conn(region=region, key=key, keyid=keyid, profile=profile)
+        if not conn:
+            return {'results': bool(conn)}
+
+        taglist = _tag_doc(tags)
+        rds = conn.create_db_cluster(DBClusterIdentifier=name,
+                                     Engine=engine,
+                                     MasterUsername=master_username,
+                                     MasterUserPassword=master_user_password,
+                                     Tags=taglist)
```
tkwilliams: Same thoughts about Throttle handling here :)

salt/modules/boto_rds.py
```diff
@@ -447,6 +474,46 @@ def create_parameter_group(name, db_parameter_group_family, description,
         return {'error': salt.utils.boto3.get_error(e)}
 
 
+def create_db_cluster(name, engine, master_username, master_user_password,
+                      tags=None, region=None, key=None, keyid=None,
+                      profile=None):
+    '''
+    .. versionadded:: Fluorine
```
tkwilliams: I'm also thinking some doc strings describing the input options (I very often just crib the info from the boto docu, changing where needed to make it sensible in the Salt context).

salt/states/boto_rds.py
```diff
@@ -733,3 +733,77 @@ def parameter_present(name, db_parameter_group_family, description, parameters=N
         else:
             ret['comment'] = os.linesep.join([ret['comment'], 'Parameters {0} for group {1} are present.'.format(params, name)])
     return ret
+
+
+def db_cluster_present(name, engine, master_username, master_user_password, parameters=None,
```
tkwilliams: Since the norm for Salt states is to enforce a given configuration, whether when creating new resources or updating existing ones, I think it would be good to call out somewhere in this function that it is unable to UPDATE existing clusters -- if a change is required, the cluster would need to be destroyed and recreated.  This is in no way your fault and TBH, it will require porting this module entirely to boto3 to make such a thing possible, but I do think that documenting this atypical behaviour would serve the principal of least surprise in this case.

tkwilliams (changes requested):

johnclyde:
I actually don't think we should land this PR. It seems like boto_rds needs a consistent pattern for error handling and it's beyond the scope of changes my team should be attempting to land in saltstack repo. If anything I would recommend landing my very first iteration - the one that copied the existing patterns in boto_rds without rolling our own error handling logic - to fill the feature gap outlined in the PR description.

rallytime:
@johnclyde Yes, that's true that this module (well, really all of the boto modules) need a consistent way to handling throttling. I agree that is a large project and beyond the scope of this PR/addition.

If you would like to make the throttling changes in the places requested above, we can certainly still get this in. Is that something you're interested in? If not, we can always close this and come back to it later as needed.

cachedout:
It sounds like the suggestion here is to abandon this and start over with a centralized approach. I'm going to close this for now and we can use it as a reference for follow-on work. If I misread that, please let me know and we can re-open and discuss. Thanks, all.

anitakrueger:
I'm surprised this was abandoned and there has been no follow-up whatsoever for more than a year. Is Aurora not used in the Salt community?
We created our other RDS instances with the boto_rds state and now wanted to create an Aurora instance failing miserably trying to use the current state (which for example requires the allocated_storage parameter which isn't used in Aurora, see Boto3 documentation).
Then I found this PR only to see that it implemented the feature and failed to be merged because of style changes that were out of scope of the feature.
The management of the boto modules/states is a bit disappointing. Is there so little usage for them? What does it take to get this feature implemented please?

anitakrueger:
@rallytime Can this be re-opened?

johnclyde:
My recollection was that the implementation at the end of the PR (after several commits) was actually functionally broken. The original commit I had pushed was sufficient to get the cluster created.

anitakrueger:
I'd be perfectly fine with that. At the moment, there is no way to create an Aurora cluster with boto_rds. I don't see that this would not be a wanted feature...

johnclyde:
You're welcome to spin up a PR and use my changes - this was something I was working on at a company where I am no longer employed.

PR #44624, opened 2017-11-21
Fix Traceback when using the `service.enabled` state on non-booted systems

eliasp:
### What does this PR do?

It fixes a traceback when using the `service.enabled` state on non-booted systems (e.g when bootstrapping plain filesystem images for containers using `systemd-nspawn -D …`).

### What issues does this PR fix or reference?
None

### Previous Behavior
`salt-call state.single service.enabled name=salt-minion`:

```
          ID: salt-minion
    Function: service.enabled
      Result: False
     Comment: An exception occurred in this state: Traceback (most recent call last):
                File "/usr/lib/python2.7/dist-packages/salt/state.py", line 1703, in call
                  **cdata['kwargs'])
                File "/usr/lib/python2.7/dist-packages/salt/loader.py", line 1649, in wrapper
                  return f(*args, **kwargs)
                File "/usr/lib/python2.7/dist-packages/salt/states/service.py", line 489, in enabled
                  ret.update(_enable(name, None, **kwargs))
                File "/usr/lib/python2.7/dist-packages/salt/states/service.py", line 97, in _enable
                  if not _available(name, ret):
                File "/usr/lib/python2.7/dist-packages/salt/states/service.py", line 275, in _available
                  avail = __salt__['service.available'](name)
                File "/usr/lib/python2.7/dist-packages/salt/modules/upstart.py", line 312, in available
                  return name in get_all()
                File "/usr/lib/python2.7/dist-packages/salt/modules/upstart.py", line 340, in get_all
                  return sorted(get_enabled() + get_disabled())
                File "/usr/lib/python2.7/dist-packages/salt/modules/upstart.py", line 274, in get_enabled
                  if _sysv_is_enabled(name):
                File "/usr/lib/python2.7/dist-packages/salt/modules/upstart.py", line 224, in _sysv_is_enabled
                  return not _sysv_is_disabled(name)
                File "/usr/lib/python2.7/dist-packages/salt/modules/upstart.py", line 216, in _sysv_is_disabled
                  return not bool(glob.glob('/etc/rc{0}.d/S*{1}'.format(_runlevel(), name)))
                File "/usr/lib/python2.7/dist-packages/salt/modules/upstart.py", line 151, in _runlevel
                  out = __salt__['cmd.run'](['runlevel', '{0}'.format(_find_utmp())], python_shell=False)
                File "/usr/lib/python2.7/dist-packages/salt/modules/upstart.py", line 100, in _find_utmp
                  return result[sorted(result).pop()]
[...]

PR #46293, opened 2018-03-02
Fix Python3 comparison `TypeError` in `salt.modules.upstart`

eliasp:
Use `len()` to determine the number of keys in a dictionary.

#44624 introduced a Python3 incompatible comparison where a dict is
compared against an integer, causing breakage on Ubuntu 18.04/bionic
which uses Python3 by default for SaltStack:

```
File "/usr/lib/python3/dist-packages/salt/modules/upstart.py", line 100, in _find_utmp
  if result > 0:
TypeError: '>' not supported between instances of 'dict' and 'int'
```

### What does this PR do?
See commit msg

https://github.com/saltstack/salt/pull/44624#discussion_r156496576

### Previous Behavior
```
File "/usr/lib/python3/dist-packages/salt/modules/upstart.py", line 100, in _find_utmp
  if result > 0:
TypeError: '>' not supported between instances of 'dict' and 'int'
```

### New Behavior
`salt.modules.upstart` doesn't cause a backtrace anymore

### Tests written?

No

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/modules/upstart.py
+++ b/salt/modules/upstart.py
@@ -97,7 +97,7 @@ def _find_utmp():
             result[os.stat(utmp).st_mtime] = utmp
         except Exception:
             pass
-    if result > 0:
+    if len(result):
         return result[sorted(result).pop()]
     else:
         return False


gtmanfred:
Just remove the > 0.  If the dictionary is empty it will be false. You don't have to check the length

eliasp:
Done

rallytime (approved):

PR #46294, opened 2018-03-02
Enhance Kapacitor module and state.

aabouzaid:
### What does this PR do?
Current Kapacitor module and state are missing some functionality.

This PR has following:
- TICK script could be defined to access more than 1 database/retention policy.
- Set Kapacitor URL (in case `kapacitord` doesn't listen to `localhost`).
- Support for URL using `https` and use `http` by default.
- Support for `-skipVerify` option to allow "Unsafe SSL".

### Tests written?
Yes

### Commits signed with GPG?
Yes


salt/modules/kapacitor.py
```diff
@@ -105,6 +121,7 @@ def _run_cmd(cmd):
 def define_task(name,
                 tick_script,
                 task_type='stream',
+                dbrps=None,
```
rallytime: Can you move this new kwarg to the end of the list? We don't want to change the API unnecessarily here.

salt/states/kapacitor.py
```diff
@@ -32,6 +34,7 @@ def __virtual__():
 def task_present(name,
                  tick_script,
                  task_type='stream',
+                 dbrps=None,
```
rallytime: Same comment here as above.

salt/states/kapacitor.py
```diff
@@ -47,6 +50,11 @@ def task_present(name,
     task_type
         Task type. Defaults to 'stream'
 
+    dbrps
+        A list of databases and retention policies in "dbname"."rpname" format
+        to fetch data from. For backward compatibility, the value of
+        'database' and 'retention_policy' will be merged as part of dbrps.
+
```
rallytime: Can you add a `.. versionadded:: Fluorine` tag to these docs (and do the same for the docs in the module)?

salt/states/kapacitor.py
```diff
@@ -63,6 +71,7 @@ def task_present(name,
 
     task = __salt__['kapacitor.get_task'](name)
     old_script = task['script'] if task else ''
+    task_dbrps = [{'db': dbrp[0], 'rp': dbrp[1]} for dbrp in (dbrp.split('.') for dbrp in dbrps)]
```
rallytime: How does this change things for people specifying `database` and `retention_policy` and not `dbrps`? It seems like this would change the default behavior.

rallytime (commented):
Hi @AAbouZaid - I have a couple of requests and one question on this change. Thanks!

salt/states/kapacitor.py
AAbouZaid: Sure, but how I can add it for mixed sections like this:

```
'''
Kapacitor state module.

:configuration: This module accepts connection configuration details either as
    parameters or as configuration settings in /etc/salt/minion on the relevant
    minions::

        kapacitor.unsafe_ssl: 'false'
        kapacitor.protocol: 'http'
        kapacitor.host: 'localhost'
        kapacitor.port: 9092

    This data can also be passed into pillar. Options passed into opts will
    overwrite options passed into pillar.

.. versionadded:: 2016.11.0
'''
```

Since `kapacitor.unsafe_ssl` and `kapacitor.protocol` are new, but rest are old (`2016.11.0`)

rallytime (approved):

salt/modules/kapacitor.py
```diff
@@ -117,6 +134,13 @@ def define_task(name,
...
 
+    dbrps
+        A list of databases and retention policies in "dbname"."rpname" format
+        to fetch data from. For backward compatibility, the value of
+        'database' and 'retention_policy' will be merged as part of dbrps.
+
+        .. versionadded:: Fluorine
+
     database
         Which database to fetch data from. Defaults to None, which will use the
```
AAbouZaid: @rallytime btw, I believe this has never been "really" true, it has been changed on [28.10.2015](https://github.com/influxdata/kapacitor/commit/19bf7e7b8b02fa8eafbe72c95f92645728dd14e2#diff-e58bf33a2f703ec10ac03af9342cd4f9R299) which means that happened even before the major change in v0.13.

Therefore, I will change this too.

rallytime: Ok sounds good.

PR #46315, opened 2018-03-02
add multi sysctl

basic-system:
### What does this PR do?
apply more than one sysctl at a time
### Previous Behavior
net.ipv4.tcp_synack_retries:
  sysctl.present:
    - config: /etc/sysctl.d/30-net.conf
     - value: 2

net.ipv4.tcp_syn_retries:
sysctl.present:
  - config: /etc/sysctl.d/30-net.conf
  - value: 2
### New Behavior
all_sysctls:
  sysctl.present:
    - sysctls:
      - net.ipv4.tcp_syn_retries: 2
      - net.ipv4.tcp_synack_retries: 2
    - config: /etc/sysctl.d/30-net.conf
### Tests written?

No

### Commits signed with GPG?

No

### DIFF
--- a/salt/modules/linux_sysctl.py
+++ b/salt/modules/linux_sysctl.py
@@ -2,22 +2,21 @@
 '''
 Module for viewing and modifying sysctl parameters
 '''
-from __future__ import absolute_import, print_function, unicode_literals
+from __future__ import absolute_import
 
 # Import python libs
 import logging
 import os
 import re
-import string
 
 # Import salt libs
-from salt.ext import six
+import salt.ext.six as six
+import salt.utils
 from salt.ext.six import string_types
 from salt.exceptions import CommandExecutionError
-import salt.utils.data
-import salt.utils.files
+from salt.utils.odict import OrderedDict
 import salt.utils.systemd
-import salt.utils.stringutils
+import string
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


salt/modules/linux_sysctl.py
```diff
@@ -2,22 +2,21 @@
...
+import salt.ext.six as six
+import salt.utils
 from salt.ext.six import string_types
 from salt.exceptions import CommandExecutionError
-import salt.utils.data
-import salt.utils.files
+from salt.utils.odict import OrderedDict
 import salt.utils.systemd
-import salt.utils.stringutils
+import string
```
rallytime: Why is this import being moved from above? That should stay within the "Python Imports" list.

salt/modules/linux_sysctl.py
```diff
@@ -190,11 +221,11 @@ def persist(name, value, config=None):
     # Read the existing sysctl.conf
     nlines = []
     try:
-        with salt.utils.files.fopen(config, 'r') as _fh:
+        with salt.utils.fopen(config, 'r') as _fh:
             # Use readlines because this should be a small file
             # and it seems unnecessary to indent the below for
             # loop since it is a fairly large block of code.
-            config_data = salt.utils.data.decode(_fh.readlines())
```
rallytime: Please do not remove this support. It is needed for PY2/PY3 compatibility.

rallytime (changes requested):
Hi @basic-system - Thanks for contributing here! I have left several review comments, but after going through this more, you are changing a lot of things in here that should not be changed.

Can you go back and change those back to the way they were? The `utils` paths need to stay the same, so do the calls using those functions. If you could just change the things that are needed specifically for your change, rather than all of the underlying calls, that would be greatly appreciated. Then we can review this again.

If you us the `.testing.pylintrc` file in this repo, that will help clean this up as well. 

Please let me know if you have any questions - we are happy to help.

rallytime:
@basic-system It looks like this also has a merge conflict, which will need to be rebased and fixed as well. :)

rallytime:
Since we haven't heard back on these changes, I am going to close this. Please feel free to open a new PR against the latest `develop` once you've had a chance to address the comments above and fix the merge conflicts. We can certainly continue the discussion here as needed. Thanks!

PR #46317, opened 2018-03-03
Ensure strings in grains are unicode

terminalmage:
This also ensures that PY3's grains contain str types and not bytes when the loader returns cached grains.


rallytime:
re-run py

rallytime (approved):

PR #46320, opened 2018-03-03
add warning about future config option change

mcalmer:
### What does this PR do?

PR https://github.com/saltstack/salt/pull/46244 introduce new config options for kubernetes.
This PR add a warning to let the user know that they have to change options in future.

### Tests written?

No

### DIFF
--- a/salt/modules/kubernetes.py
+++ b/salt/modules/kubernetes.py
@@ -31,6 +31,13 @@ In case both are provided the `file` entry is prefered.
     salt '*' kubernetes.nodes api_url=http://k8s-api-server:port api_user=myuser api_password=pass
 
 .. versionadded: 2017.7.0
+
+.. warning::
+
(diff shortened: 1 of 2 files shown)


rallytime:
Thanks for adding these documentation warnings @mcalmer. Can you also add calls in the code to using `salt.utils.warn_until` to make sure people using this module will see that the old configs will be removed in a future version?

mcalmer:
@rallytime - As I wrote in https://github.com/saltstack/salt/pull/46244 I think `warn_until` is not a good idea. All these warnings in the logs just hide really important messages and in this special case nobody can prevent them by changing the config to the new option as this is not implemented in this version.

Do you really want this warning producing megabytes of useless messages?

rallytime:
@mcalmer Ah, you're right. I didn't see that yet. I'll update there. Thanks!

rallytime (approved):

PR #46326, opened 2018-03-05
Fixes a timing bug of saltnado's client local.

kstreee:
### What does this PR do?

A complement fix of our previous pr (#45874). The previous pr wasn't perfect to patch the problem what we referred before.

#45874

### Previous Behavior

Subscribes minion returns after published job. Thus, still, the saltnado version of salt api client 'local' can miss minion returns by the timing issue what we described at former pr (#45874).

### New Behavior

Generates `jid` before `firing` a job to subscribe minion ret using the generated `jid`.

### Tests written?

No, but we checked the test case `test_simple_local_post_no_tgt` ( `/tests/integration/netapi/rest_tornado/test_app.py` ), which checkes 'no jid' case, works same.

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/netapi/rest_tornado/saltnado.py
+++ b/salt/netapi/rest_tornado/saltnado.py
@@ -371,7 +371,7 @@ class EventListener(object):
         for (tag, matcher), futures in six.iteritems(self.tag_map):
             try:
                 is_matched = matcher(mtag, tag)
-            except Exception as e:
+            except Exception:
                 logger.error('Failed to run a matcher.', exc_info=True)
                 is_matched = False
 
@@ -899,7 +899,7 @@ class SaltAPIHandler(BaseSaltAPIHandler):  # pylint: disable=W0223
             try:
                 chunk_ret = yield getattr(self, '_disbatch_{0}'.format(low['client']))(low)
                 ret.append(chunk_ret)
-            except EauthAuthenticationError as exc:
+            except EauthAuthenticationError:
                 ret.append('Failed to authenticate')
                 break
             except Exception as ex:
@@ -914,7 +914,11 @@ class SaltAPIHandler(BaseSaltAPIHandler):  # pylint: disable=W0223
         '''
         Dispatch local client commands
         '''
-        chunk_ret = {}
+        # Generate jid before triggering a job to subscribe all returns from minions
+        chunk['jid'] = salt.utils.jid.gen_jid()
+
+        # Subscribe returns from minions before firing a job
+        future_minion_map = self.subscribe_minion_returns(chunk['jid'], chunk['tgt'])
 
         f_call = self._format_call_run_job_async(chunk)
         # fire a job off
@@ -926,10 +930,12 @@ class SaltAPIHandler(BaseSaltAPIHandler):  # pylint: disable=W0223
         # if the job didn't publish, lets not wait around for nothing
         # TODO: set header??
         if 'jid' not in pub_data:
+            for future in future_minion_map:
+                try:
+                    future.set_result(None)
+                except Exception:
+                    pass
             raise tornado.gen.Return('No minions matched the target. No command was sent, no jid was assigned.')
-
-        # seed minions_remaining with the pub_data
-        minions_remaining = pub_data['minions']
 
         syndic_min_wait = None
         if self.application.opts['order_masters']:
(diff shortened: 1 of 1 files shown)


rallytime:
@cachedout and @skizunov - were you able to review this PR?

cachedout (approved):
This makes a lot of sense to me. I like it.

Moulde:
This pull request fixes an issue in the pull request #45874, and in that pull request (#45874), the saltnado.py started using the timeout configuration value from the master config.

self.application.opts['timeout']

The default value for which is 5 seconds, and it seems that this causes most of our api calls in salt-api when using tornado, to timeout, and just return [{}], even test.ping. But by increasing the timeout value, it returns the correct result.

But why does it even have a hard timeout like that? Shouldn't it just send the job to the minion to execute, and then repeatedly ask what the status is until the minion returns a result?

kstreee:
@cachedout Thanks for the comment, and congratulation for releasing the new version 2018.3  🎉

kstreee:
@Moulde Hi Moulde, The reason that we added `self.application.opts['timeout']` is to prevent waiting forever if even salt minion never return caused by some unknown reasons. Requests of Salt API can be terminated by the function `job_not_running` eventually, but the subscribing code in the function `subscribe_minion_returns` has a chance to run forever.

But what you pointed about the default timeout also should be resolved to prevent 'inconvenient experience in Salt API' for people who try to use the system. (In our case, we always set 'timeout' long enough to not be happened what you described, thus, unfortunately, we couldn't be aware of the issue.)
I think following examples (with a comment to explain the reason) could resolve the issue (in my opinion, the first one makes sense than others).

`timeout = self.application.opts['timeout'] + self.application.opts['gather_job_timeout']`
`timeout = self.application.opts['timeout'] if self.application.opts['timeout'] > 30 or 30`
`timeout = self.application.opts['NEW_OPTION_TO_RESOLVE_THE_ISSUE']`

After some testing, I will send a pr to resolve your issue. If you want to proceed this issue on your own, let me know your decision. Cause you discovered it, you deserve to get a credit :)

Thanks,

Moulde:
Hi

That's my current solution as well, just setting the timeout value to a larger value, but it's usually never a good idea to just increase timeouts, especially when I for one, has no idea where that somewhat generic 'timeout' configuration value is used other than in this case.

Regarding your first solution
`timeout = self.application.opts['timeout'] + self.application.opts['gather_job_timeout']`

So for default settings, that will be set to 5 + 10 = 15 seconds, so most commands should return in that timespan, but what about longer running jobs, in this case 15 seconds? Or will it have received a 'still running'-ping and keep running?

I changed our code to, if present, use a timeout parameter sent along with the jobs, that way we don't have to increase the timeout dramatically in the configuration. And if no timeout parameter is present, it will just fall back to the system configuration, which i increased to 30 seconds.
That way we can run very long-running jobs like highstate, by including a high timeout value for that specific request. But from what i understand, it may not even be necessary to set the timeout to anything more than 30 seconds?

Feel free to create the pr, but thank you! :)

kstreee:
@Moulde Hi, I read the code and tested the 'timeout' option, and I realize that I have misunderstood about it. I thought the 'timeout' is referring a timeout for minions' job but it isn't. (Actually, we set the 'timeout' option as 3600 to make sure that most of the minions' job can be done before the timeout.)

So, as you pointed out the timeout, removing 'timeout' as a timeout parameter in the code is a definitely reasonable opinion. Then, I found the method [`clean_timeout_futures`](https://github.com/kstreee/salt/blob/0f358a9c9ed4b79c875e79fb54bcece2258ec5d1/salt/netapi/rest_tornado/saltnado.py#L292) and the method is called to clean up every garbage future objects that are related to a request, [after finished handling the request](https://github.com/kstreee/salt/blob/0f358a9c9ed4b79c875e79fb54bcece2258ec5d1/salt/netapi/rest_tornado/saltnado.py#L486). Thus, removing timeout parameter is completely safe. To fix this issue, I will send a pr.
Thanks for pointing out the issue to fix correctly! :)

rallytime:
This does not merging forward cleanly to develop, so this fix will need to be made directly against the develop branch. (The merge forward to 2018.3 was fine, though.)

@kstreee Would you be willing to make this fix against the develop branch?

PR #46328, opened 2018-03-05
Fix openscap push

dincamihai:
### What does this PR do?

Replaces salt.client.Caller with __salt__

openscap module

### Tests written?

Yes (when the module was implemented)
I expect the tests to fail but I will keep the PR in WIP until I will fix them.

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/modules/openscap.py
+++ b/salt/modules/openscap.py
@@ -13,7 +13,6 @@ from subprocess import Popen, PIPE
 
 # Import Salt libs
 from salt.ext import six
-from salt.client import Caller
 
 
 ArgumentParser = object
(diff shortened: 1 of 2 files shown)


rallytime:
We're tagging the RC today, and I don't think this change should block that. Would you mind moving this to the `2018.3` branch so it can be included in 2018.3.1?

dincamihai:
@rallytime I have switched the base to 2018.3 and I've fixed the tests. No longed WIP. It can be merged if no objections. Thanks!

rallytime (approved):
👍

Issue #42438, opened 2017-07-20
Failed to send message: hipchat-message


ajoaugustine:
### Description of Issue/Question
Issue: Not able to pass the keyword color for  'hipchat.send_message'. Status shows failed. However the message is getting posted at the hip chat.
Pasting the error message:
[ERROR   ] Failed to send message: hipchat-message
local:
----------
          ID: hipchat-message
    Function: hipchat.send_message
      Result: False
     Comment: Failed to send message: hipchat-message
     Started: 19:11:07.306542
    Duration: 245.703 ms
     Changes:
    Warnings: 'color' is an invalid keyword argument for 'hipchat.send_message'.
              If you were trying to pass additional data to be used in a
              template context, please populate 'context' with 'key: value'
              pairs. Your approach will work until Salt Oxygen is out. Please
              update your state files.

Summary for local
------------
Succeeded: 1
Failed:    1
Warnings:  1
------------
Total states run:     2
Total run time: 748.277 ms

### Setup
(Please provide relevant configs and/or SLS files (Be sure to remove sensitive info).)
sls file:
  hipchat.send_message:
    - room_id: 40305
    - from_name: SuperAdmin
    - message: 'This state was executed successfully.'
    - api_url: https://api.hipchat.com
    - api_key: uDMxaRzJNyIuIOIobUhM7BgHCgOtKf
    - api_version: v2
    - color: 'red'
    - notify: True

### Steps to Reproduce Issue
execute the state with a valid api key and room ID.

To fix the 'color' related error, you need to replace color with 'message_color' in the sls file and then modify the documentation:
https://docs.saltstack.com/en/latest/ref/states/all/salt.states.hipchat.html

or update /usr/lib/python2.7/dist-packages/salt/states/hipchat.py

### Versions Report
Salt Version:
           Salt: 2017.7.0

Dependency Versions:
           cffi: Not Installed
       cherrypy: Not Installed
       dateutil: 2.4.2
      docker-py: Not Installed
          gitdb: 0.6.4
      gitpython: 1.0.1
          ioflo: Not Installed
         Jinja2: 2.8
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: 1.0.3
   msgpack-pure: Not Installed
 msgpack-python: 0.4.6
   mysql-python: Not Installed
      pycparser: Not Installed
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: Not Installed
[...]

PR #46333, opened 2018-03-05
Fixes color parameter mismatch and handles 204 responses correctly

danlsgiga:
### What does this PR do?
Fixes the hipchat module and state to handle API v2 204 (NO_CONTENT) responses correctly. Also fixes the `message_color` and `color` parameters mismatch between the docs and the code.

#42438

### Previous Behavior
When trying to send a message to hipchat using the state, the message is delivered sucessfully but the state returns as failed since it is considering a 204 HTTP response as invalid when it is actually the desired behaviour for the [Hipchat v2 API](https://www.hipchat.com/docs/apiv2/method/send_room_notification)

### New Behavior
The state returns True for such scenarios

### Tests written?

No

### Commits signed with GPG?

No

### Notes
The `develop` branch has this already fixed but it uses a code leveraged for the upcoming 2018 release  that contains components not present on 2017 like `salt.utils.json`. Pushing this PR in the hopes it can make to the next point release 2017.7.5

### DIFF
--- a/salt/states/hipchat.py
+++ b/salt/states/hipchat.py
@@ -45,7 +45,7 @@ def send_message(name,
                  api_url=None,
                  api_key=None,
                  api_version=None,
-                 color='yellow',
+                 message_color='yellow',
                  notify=False):
     '''
     Send a message to a Hipchat room.
@@ -60,7 +60,7 @@ def send_message(name,
             - api_url: https://hipchat.myteam.com
             - api_key: peWcBiMOS9HrZG15peWcBiMOS9HrZG15
             - api_version: v1
-            - color: green
+            - message_color: green
             - notify: True
 
     The following parameters are required:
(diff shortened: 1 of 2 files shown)


salt/states/hipchat.py
```diff
@@ -45,7 +45,7 @@ def send_message(name,
                  api_url=None,
                  api_key=None,
                  api_version=None,
-                 message_color='yellow',
+                 color='yellow',
```
gtmanfred: This should not be changed.  We do not like changing the actual code in patch releases.

The documentation should just be updated.

gtmanfred (changes requested):

salt/states/hipchat.py
danlsgiga: Ah, yeah. my bad! Will revert!

gtmanfred (approved):

Issue #46237, opened 2018-02-28
Python 3 API TypeError in salt-cloud 2017.7.4 on EC2


max-orhai:
### Description of Issue/Question

I'm using salt-cloud in EC2. I upgraded to 2017.7.4 hoping to use the Python 3 API, but encountered an error in the ec2.py availability zone listing code.

### Setup

```
ubuntu@ip-XXXXX:~$ cat /etc/salt/cloud.providers
ec2_public:
  minion:
    master: XXXXX
  ssh_interface: private_ips
  id: 'use-instance-role-credentials'
  key: 'use-instance-role-credentials'
  private_key: /etc/salt/saltmaster-key.pem
  keyname: saltmaster-key
  securitygroup: salty
  location: us-west-2
  availability_zone: us-west-2b
  driver: ec2

ubuntu@ip-XXXXX:~$ cat /etc/salt/cloud.profiles
ec2_test:
  provider: ec2_public
  # stock amd64 Ubuntu 16.04
  image: ami-0def3275
  size: t2.micro
  ssh_username: ubuntu
```

### Steps to Reproduce Issue

```
ubuntu@ip-XXXXX:~$ sudo python3
Python 3.5.2 (default, Nov 23 2017, 16:37:01) 
[GCC 5.4.0 20160609] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from salt.cloud import CloudClient
>>> cloudcli = CloudClient(path='/etc/salt/cloud')
>>> cloudcli.profile('ec2_test', names=['tmp0', 'tmp1'])
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/ubuntu/.local/lib/python3.5/site-packages/salt/cloud/__init__.py", line 348, in profile
    mapper.run_profile(profile, names, vm_overrides=vm_overrides)
  File "/home/ubuntu/.local/lib/python3.5/site-packages/salt/cloud/__init__.py", line 1448, in run_profile
    ret[name] = self.create(vm_)
  File "/home/ubuntu/.local/lib/python3.5/site-packages/salt/cloud/__init__.py", line 1278, in create
    output = self.clouds[func](vm_)
  File "/home/ubuntu/.local/lib/python3.5/site-packages/salt/cloud/clouds/ec2.py", line 2617, in create
    data, vm_ = request_instance(vm_, location)
[...]

PR #46382, opened 2018-03-06
Don't encode response string from role API

max-orhai:
Substituting a bytestring in the URL of the subsequent request leads to a malformed URL and thus 404.

Solves #46237. This may also be an issue elsewhere in here where Python 3 string semantics differ from Python 2. I haven't checked all the other uses of `encode` in this file, but this was enough to get me going.

### DIFF
--- a/salt/utils/aws.py
+++ b/salt/utils/aws.py
@@ -89,9 +89,7 @@ def creds(provider):
                 proxies={'http': ''}, timeout=AWS_METADATA_TIMEOUT,
             )
             result.raise_for_status()
-            role = result.text.encode(
-                result.encoding if result.encoding else 'utf-8'
-            )
+            role = result.text
         except (requests.exceptions.HTTPError, requests.exceptions.ConnectionError):
             return provider['id'], provider['key'], ''
 


salt/utils/aws.py
```diff
@@ -89,9 +89,7 @@ def creds(provider):
                 proxies={'http': ''}, timeout=AWS_METADATA_TIMEOUT,
             )
             result.raise_for_status()
-            role = result.text.encode(
-                result.encoding if result.encoding else 'utf-8'
-            )
+            role = result.text
```
gtmanfred: This is going to need more testing, because the result should report back the encoding, and it is used in other places fine.

They may be better to just run `salt.utils.stringutils.to_str` for places where it is needed, like 

```
"http://169.254.169.254/latest/meta-data/iam/security-credentials/{0}".format(salt.utils.stringutils.to_str(role)),
```

gtmanfred (changes requested):

gtmanfred (approved):

gtmanfred:
Actually, this is probably ok if it is only for the credentials. Did you verify that this works correctly on python2 as well?

max-orhai:
Yes, this works for me on Py2.7.

PR #46389, opened 2018-03-06
linux_lvm module: refactor lvresize

rkissos:
### What does this PR do?
linum_lvm module: lvresize refactoring

1. lvresize used by default --size, added --extents.
2. lvresize return value

### Previous Behavior
lvresize returns an empty dict

### New Behavior
lvresize returns the command output.

### Tests written?
Yes


garethgreenaway (approved):

salt/modules/linux_lvm.py
```diff
@@ -486,10 +486,21 @@ def lvresize(size, lvpath):
...
+
     '''
-    ret = {}
-    cmd = ['lvresize', '-L', six.text_type(size), lvpath]
-    cmd_ret = __salt__['cmd.run_all'](cmd, python_shell=False)
-    if cmd_ret['retcode'] != 0:
-        return {}
-    return ret
+    if size and extents:
+        return 'Error: Please specify only one of size or extents'
```
rallytime: This is changing the expected return structure it seems. Is that intended? Is that handled elsewhere?

salt/modules/linux_lvm.py
```diff
@@ -486,10 +486,21 @@ def lvresize(size, lvpath):
...
+
+    cmd = ['lvresize']
+
+    if size:
+        cmd.extend(['-L', '{0}'.format(size)])
+    elif extents:
+        cmd.extend(['-l', '{0}'.format(extents)])
+    else:
+        return 'Error: Either size or extents must be specified'
```
rallytime: Same question here.

rallytime (commented):
I have some questions about this change.

salt/modules/linux_lvm.py
rkissos: Yes, but this will be returned iff _extents_ and size will be used. And since _extents_ is introduced here, it won't affect others.

salt/modules/linux_lvm.py
rkissos: Needed to be defensive since I had to change _size_ to keyword argument.
But as mentioned, won't affect others who use lvextend('+12M', '/dev/mapper/vg1-test')

salt/modules/linux_lvm.py
rallytime: Yeah, sure. Mostly I am wondering if we can return a more consistent data structure and _log_ the error message. How would a user know when this function is successful or not without examining the string? If this function is going to be changed, it should be consistent with the other functions in this execution module. I don't think we should change the return data structure when people are expecting something else unless there is a good reason.

rkissos: So the retrun 'Error...' is the same style as the rest of the functhions in this module.
The return value type in line 506 is a dict, because this fucntion returned an (empty) dict consistenly.

Do you want me to change the error strings to dict? or the dict (line 506) to string?

rallytime: I think we should keep the return as a dict and log the error. Something like this:
```
if size and extents:
    log.error('Error: Please specify only one of size or extents')
    return {}
```

rkissos:
@rallytime pushed my changes. Thanks!

rallytime (approved):

rallytime:
@rkissos Thanks for making those changes. This looks good to me now, however one of the tests is failing now:

- unit.modules.test_linux_lvm.LinuxLVMTestCase.test_lvresize

https://jenkins.saltstack.com/job/PR/job/salt-pr-linode-ubuntu14-n/21636/testReport/junit/unit.modules.test_linux_lvm/LinuxLVMTestCase/test_lvresize/

rallytime:
@rkissos Thank you! And congrats on your first Salt PR! :D

PR #46391, opened 2018-03-06
Reworked modules.restartcheck: optimize code, generator, timeout

homolkad:
### What does this PR do?
Reworked restart check: used generator for _detete_files, added timeout in minutes for checking loops in restartcheck, added list dirs and using regex for valid deleted file. Optimize code.

None

### New Behavior
Add timeout to restartcheck, default 5 min.

### Tests written?
Yes

### Commits signed with GPG?
Yes


garethgreenaway:
@homolkad Great work.  Are there existing tests for this module?

homolkad:
@garethgreenaway I think, the tests never existed for this module.

garethgreenaway:
@homolkad Looks like you're right.  Including some tests to ensure the module is functioning as expected would be great.  Let us know if you need help getting going with adding them.

homolkad:
@garethgreenaway Ok, I am going to write the tests.

garethgreenaway (approved):

Issue #46052, opened 2018-02-15
[Fluorine] Add tests for vault and sdb


gtmanfred:
### Description of Issue/Question

Vault is probably the most used backend for sdb.  We should be testing this more thoroughly

PR #46393, opened 2018-03-06
make vault like all other sdb backends

gtmanfred:
### What does this PR do?
This is a dumb.  There is no reason we shouldn't just split the last value off and use it for the key name in the vault path.

### Previous Behavior
reference sdb

```
salt-call sdb.get sdb://myvault/path/to/value?key
```

### New Behavior
New format

```
salt-call sdb.get sdb://myvault/path/to/value/key
```

So that it looks like all other sdb backends

### Tests written?

https://github.com/saltstack/salt/issues/46052

### Commits signed with GPG?

Yes

### DIFF
--- a/doc/topics/sdb/index.rst
+++ b/doc/topics/sdb/index.rst
@@ -79,22 +79,12 @@ from the ``kevinopenstack`` profile above, you would use:
 
     salt-call sdb.get sdb://kevinopenstack/password
 
-Some drivers use slightly more complex URIs. For instance, the ``vault`` driver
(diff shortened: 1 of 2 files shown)


sumeetisp:
This ? in sdb always puzzled me :)

rallytime:
@gtmanfred Lint: https://jenkins.saltstack.com/job/PR/job/salt-pr-lint-n/19868/violations/file/salt/sdb/vault.py/

gtmanfred:
Sometimes python is just too close to writing sentences.

rallytime (approved):

PR #46398, opened 2018-03-06

rallytime:
Conflicts:
  - salt/cli/api.py
  - salt/client/ssh/wrapper/state.py
  - salt/config/__init__.py
  - salt/crypt.py
  - salt/daemons/masterapi.py
  - salt/key.py
  - salt/loader.py
  - salt/master.py
  - salt/modules/archive.py
  - salt/modules/hipchat.py
  - salt/modules/pip.py
  - salt/modules/state.py
  - salt/modules/virt.py
  - salt/modules/x509.py
  - salt/state.py
  - salt/transport/tcp.py
  - salt/transport/zeromq.py
  - salt/utils/__init__.py
  - salt/utils/event.py
  - salt/utils/extmods.py
  - salt/utils/files.py
  - salt/utils/parsers.py
  - salt/utils/verify.py
  - tests/integration/shell/test_call.py
  - tests/unit/utils/test_args.py


rallytime:
re-run py

terminalmage (approved):

DmitryKuzmenko (approved):

salt/crypt.py
```diff
@@ -146,12 +146,12 @@ def gen_keys(keydir, keyname, keysize, user=None, passphrase=None):
...
                     callback=lambda x: salt.utils.stringutils.to_bytes(passphrase))
         else:
-            with salt.utils.files.fopen(priv, 'wb') as f:
+            with salt.utils.files.fopen(priv, 'wb+') as f:
                 f.write(gen.exportKey('PEM', passphrase))
     if HAS_M2:
         gen.save_pub_key(pub)
     else:
-        with salt.utils.files.fopen(pub, 'wb') as f:
+        with salt.utils.files.fopen(pub, 'wb+') as f:
```
terminalmage: Why was the `+` needed here? We're not reading from this filehandle.

DmitryKuzmenko: @terminalmage 'w+' mode truncates the file so here we 100% sure it will be written from the beginning. And the second: that mode just was there. I haven't found a commit where it was dropped. I don't see why this merge request has this change. So I've just reverted it back.
Am I wrong here?

terminalmage: @DmitryKuzmenko According to the Python docs [here](https://docs.python.org/2/library/functions.html#open), both `w` and `w+` truncate the file. The only thing that adding the `+` does is open the filehandle in read-and-write mode, essentially adding funcs like `seek` to the filehandle object. Since we're not reading from the file here, it seems to me that `wb` is sufficient.

terminalmage: Actually, I looked at the head of 2018.3 and it has the `+` already so I'm not sure how it was removed. I still don't think it's necessary, but it's not worth standing in the way of merging this.

rallytime: I'm going to merge-this as-is since it's a merge-forward. If it needs to be changed outside of this PR, it can be discussed in a new one. :)

Issue #46341, opened 2018-03-05
Can't set default registry value in Windows (2018.3.0rc1)


twangboy:
### Description of Issue/Question

You can't set the default value of a registry key with salt.

### Setup
Given the following state:
```
reg_default_value:
  reg.present:
    - name: 'HKLM\SOFTWARE\Salt'
    - vdata: Some data
```

### Steps to Reproduce Issue
Apply the state with the following result:
```
          ID: reg_default_value
    Function: reg.present
        Name: HKLM\SOFTWARE\Salt
      Result: False
     Comment: An exception occurred in this state: Traceback (most recent call last):
                File "c:\salt\bin\lib\site-packages\salt\state.py", line 1876, in call
                  **cdata['kwargs'])
                File "c:\salt\bin\lib\site-packages\salt\loader.py", line 1823, in wrapper
                  return f(*args, **kwargs)
                File "c:\salt\bin\lib\site-packages\salt\states\reg.py", line 187, in present
                  use_32bit_registry=use_32bit_registry)
                File "c:\salt\bin\lib\site-packages\salt\modules\reg.py", line 367, in read_value
                  local_vname = _to_unicode(vname)
                File "c:\salt\bin\lib\site-packages\salt\modules\reg.py", line 86, in _to_unicode
                  return salt.utils.stringutils.to_unicode(vdata, 'utf-8')
                File "c:\salt\bin\lib\site-packages\salt\utils\stringutils.py", line 115, in to_unicode
                  raise TypeError('expected str or bytearray')
              TypeError: expected str or bytearray
     Started: 21:35:54.939000
    Duration: 0.0 ms
     Changes:
```

### Versions Report
```
# salt -t 30 's*d' test.versions
slee_winsrv2016_build:
    Salt Version:
               Salt: 2018.3.0rc1

    Dependency Versions:
               cffi: 1.10.0
           cherrypy: Not Installed
           dateutil: 2.6.1
          docker-py: Not Installed
              gitdb: 2.0.3
          gitpython: 2.1.7
              ioflo: Not Installed
             Jinja2: 2.10
            libgit2: Not Installed
            libnacl: 1.6.1
           M2Crypto: Not Installed
               Mako: 1.0.7
       msgpack-pure: Not Installed
     msgpack-python: 0.4.8
       mysql-python: Not Installed
          pycparser: 2.18
           pycrypto: 2.6.1
       pycryptodome: Not Installed
             pygit2: Not Installed
[...]

PR #46399, opened 2018-03-06
Fix reg module when vname=None

twangboy:
### What does this PR do?
Fixes an issue with None values being converted to unicode (or attempted).

To set the default value of a key in the registry, you set the `vname` parameter to `None`. This was causing a stacktrace when the module was attempting to convert `vname` to unicode. This PR adds an if statement to the `_to_unicode` function to just return `None` if the passed value is `None`.

Add `win32con.REG_TYPE` names to if statements to make it clearer what is being checked.

https://github.com/saltstack/salt/issues/46341

### Tests written?
No

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/modules/reg.py
+++ b/salt/modules/reg.py
@@ -83,6 +83,8 @@ def _to_unicode(vdata):
     Converts from current users character encoding to unicode. Use this for
     parameters being pass to reg functions
     '''
+    if vdata is None:
+        return None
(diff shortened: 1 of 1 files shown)


skizunov (approved):

lorengordon (approved):

rallytime:
re-run py3

rallytime (approved):

Issue #46335, opened 2018-03-05
[2018.0rc1] salt.states.x509 now broken


sjorge:
### Description of Issue/Question
```
----------
          ID: certificate.authority::private-key
    Function: x509.private_key_managed
        Name: /salt/pki/ca.key
      Result: False
     Comment: An exception occurred in this state: Traceback (most recent call last):
                File "/opt/local/lib/python2.7/site-packages/salt/state.py", line 1876, in call
                  **cdata['kwargs'])
                File "/opt/local/lib/python2.7/site-packages/salt/loader.py", line 1823, in wrapper
                  return f(*args, **kwargs)
                File "/opt/local/lib/python2.7/site-packages/salt/states/x509.py", line 303, in private_key_managed
                  name, bits=bits, passphrase=passphrase, new=new, overwrite=overwrite):
                File "/opt/local/lib/python2.7/site-packages/salt/states/x509.py", line 240, in _check_private_key
                  'The provided passphrase cannot decrypt the private key.')
              CommandExecutionError: The provided passphrase cannot decrypt the private key.
     Started: 19:28:45.695662
    Duration: 7.64 ms
     Changes:
----------
```
### Setup
```
certificate.authority::private-key:
  x509.private_key_managed:
    - name: {{ certcfg['config']['authority_dir'] }}/ca.key
    - bits: 8192
    - backup: True
    - require:
      - file: certificate.authority::directory
```

### Steps to Reproduce Issue
Try and apply a simple state to generate a private key

### Versions Report
```
cronos# salt-call --versions-report
Salt Version:
           Salt: 2018.3.0rc1

Dependency Versions:
           cffi: 1.9.1
       cherrypy: 8.1.0
       dateutil: 2.5.3
      docker-py: Not Installed
          gitdb: 2.0.3
      gitpython: 2.1.8
          ioflo: 1.7.4
         Jinja2: 2.8
        libgit2: Not Installed
        libnacl: 1.6.1
       M2Crypto: 0.22
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.4.6
   mysql-python: Not Installed
      pycparser: 2.17
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: Not Installed
         Python: 2.7.12 (default, Jan 26 2017, 18:07:59)
   python-gnupg: 2.3.1
         PyYAML: 3.12
          PyZMQ: 16.0.2
           RAET: 0.6.8
[...]

PR #46401, opened 2018-03-06
[2018.3.0rc1] fixes to x509 module

garethgreenaway:
### What does this PR do?
Recent unicode changes into Oxygen broke some aspects of the x509 module.  When processing PEM text, it needs to be treated as ascii not unicode.  Adding a couple tests for affected functions.

#46335 

### Tests written?
Yes

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/modules/x509.py
+++ b/salt/modules/x509.py
@@ -131,6 +131,10 @@ def _new_extension(name, value, critical=0, issuer=None, _pyfree=1):
         raise salt.exceptions.SaltInvocationError(
             'value must be precomputed hash')
 
+    # ensure name and value are bytes
+    name = salt.utils.stringutils.to_bytes(name)
+    value = salt.utils.stringutils.to_bytes(value)
+
     try:
         ctx = M2Crypto.m2.x509v3_set_nconf()
         _fix_ctx(ctx, issuer)
--- a/salt/modules/x509.py
+++ b/salt/modules/x509.py
@@ -316,7 +316,7 @@ def _text_or_file(input_):
     '''
     if os.path.isfile(input_):
         with salt.utils.files.fopen(input_) as fp_:
-            return salt.utils.stringutils.to_str(fp_.read())
+            return salt.utils.stringutils.to_bytes(fp_.read())
     else:
         return input_
 
--- a/salt/modules/x509.py
+++ b/salt/modules/x509.py
@@ -316,7 +316,7 @@ def _text_or_file(input_):
     '''
     if os.path.isfile(input_):
         with salt.utils.files.fopen(input_) as fp_:
-            return salt.utils.stringutils.to_unicode(fp_.read())
+            return salt.utils.stringutils.to_str(fp_.read())
     else:
         return input_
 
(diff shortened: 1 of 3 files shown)


rallytime (approved):
👍

terminalmage (commented):
Has this been confirmed on Python 3 under actual usage (i.e. not under unit testing)?

Some of the issues we've seen as we've updated for `unicode_literals` have been because code that needs a bytestring stopped working when the string became unicode. The issue though is that `to_str` returns a str type on both Python 2 and 3, and `str` is the same as `bytes` on Python 2 but is _not_ on Python 3. So, using `to_str` can end up fixing these sort of issues on Python 2 but not on Python 3.

I'm sure this is probably OK, but I just wanted to make sure that we didn't fix this only on Python 2.

garethgreenaway:
@terminalmage I'll verify that a real world example works using Python3.

rallytime:
@garethgreenaway It also looks like one related test is failing here: https://jenkins.saltstack.com/job/PR/job/salt-pr-rs-cent7-n/16938/ probably needs to be gated. The m2crypto packages are installed on branches, but I think the PRs don't have the newest version of bootstrap to install m2crypto presently.

garethgreenaway:
@terminalmage I think the issue might be moot since the module in question relies on M2Crypto, which isn't easily install-able for Py3.  That being said, everything works when using to_bytes to I swapped that out instead.

rallytime:
@garethgreenaway Some small lint here: https://jenkins.saltstack.com/job/PR/job/salt-pr-lint-n/19930/violations/file/tests/unit/modules/test_x509.py/

sjorge:
After this change I hit a new issue in x509, I have updates #46335

rallytime:
@garethgreenaway Looks like lint is still being a little tricky here: https://jenkins.saltstack.com/job/PR/job/salt-pr-lint-n/20004/violations/

@sjorge How does this look to you now?

sjorge:
@rallytime I havent tested the latest flurry of commit yet. I'm afraid I wont have time today to do so either. Hopefully tomorrow.

@garethgreenaway were you able to reproduce the 2nd error I was facing with the states @Ch3LL provided?

sjorge:
It also looks like something went wrong with rebasing on this, it seems to now also include some of the NaCL fixes!

sjorge:
@garethgreenaway Do you need me to give this another go on my setup?
I might have some time tomorrow evening to look at it again, if hotel WiFi is good enough.

garethgreenaway:
@sjorge if you could that would be great.

rallytime:
We're working on getting these PRs infor the .0 release. After discussing this with @garethgreenaway, I am going to merge this in as-is, since the x509 modules won't work without this change. If you have more feedback @sjorge, please let us know and we will follow up ASAP. :)

sjorge:
The internet was not usable at the hotel, I am back home now. I will spin up another test.

sjorge:
@rallytime sorry it took so long, the hotel did not have good wifi.

With this patch all is looking good. No more issues with rc1 on my salt-master.
I will slowly roll out my other minions later today.

PR #46403, opened 2018-03-06
Add env-support to composer modules/states

MaxBosse:
### What does this PR do?
Some packages, such as symfony, require additional environment variables during a composer run.
In Development-Environments this can be achieved using .env files however, this should not be used in production and therefore you were unable to run a composer install for symfony in a production-environment via salt until now.

This PR adds the possibility to add environment-variables to any composer run (state and modules), allowing you to set for example APP_ENV=prod for a production symfony installation.

symfony/symfony#11704

### Previous Behavior
Composer-Update would fail with the application asking you to add an environment-variable specifying the APP_ENV.

### New Behavior
Composer-Update runs successfully taking the APP_ENV into account.

### Tests written?

No

### Commits signed with GPG?

No


rallytime:
HI @MaxBosse  - Looks like there are some lint failures here: https://jenkins.saltstack.com/job/PR/job/salt-pr-lint-n/19885/violations/

salt/modules/composer.py
```diff
@@ -185,9 +189,14 @@ def _run_composer(action,
     if optimize is True:
         cmd.append('--optimize-autoloader')
 
+    if env is not None:
+        env.update({'COMPOSER_HOME': composer_home})
```
terminalmage: This assumes that the env vars are in a YAML dict:

```yaml
mystate:
  statemod.funcname:
    - name: foo
    - env:
        FOO: bar
        HELLO: world
```

But because of some weirdness in PyYAML which requires nested dicts to be double-indented to be properly loaded, in many places in Salt we also allow input to be a bit more user-friendly, like so:

```yaml
mystate:
  statemod.funcname:
    - name: foo
    - env:
      - FOO: bar
      - HELLO: world
```

We have a helper function, `salt.utils.data.repack_dictlist()` which you can use here:

```python
if env is not None:
    env = salt.utils.data.repack_dictlist(env)
    env['COMPOSER_HOME'] = composer_home
```

terminalmage (changes requested):

salt/modules/composer.py
terminalmage: That helper will load the YAML list of dicts into a single dict.

MaxBosse: Thank you for that information @terminalmage 
I was actually having some difficulties with exactly that 👍

terminalmage (approved):

max-arnold:
@MaxBosse You might want to refactor this using the new module environment framework: https://github.com/saltstack/salt/pull/48021

PR #46421, opened 2018-03-07
Skip SSHPasswordTests if ssh binary is not found

bdrung:
If the ssh binary is not found in path, SSHPasswordTests will fail:

```
ERROR: test_password_failure (unit.ssh.test_ssh.SSHPasswordTests)
[CPU:0.0%|MEM:54.9%]
  File "tests/unit/ssh/test_ssh.py", line 42, in test_password_failure
    client = ssh.SSH(opts)
  File "salt/client/ssh/__init__.py", line 226, in __init__
    raise salt.exceptions.SaltSystemExit('No ssh binary found in path -- ssh must be '
salt.exceptions.SaltSystemExit: None
```

Either SSHPasswordTests needs to be converted into a real unit test (i.e. mocking the salt call) or this tests needs to be skipped if ssh is not available.

### DIFF
--- a/tests/unit/ssh/test_ssh.py
+++ b/tests/unit/ssh/test_ssh.py
@@ -14,10 +14,12 @@ from tests.support.mock import NO_MOCK, NO_MOCK_REASON, patch, MagicMock
 
 # Import Salt libs
 import salt.config
+import salt.utils.path
 from salt.client import ssh
 
 
 @skipIf(NO_MOCK, NO_MOCK_REASON)
+@skipIf(not salt.utils.path.which('ssh'), "No ssh binary found in path")
 class SSHPasswordTests(ShellCase):
     def test_password_failure(self):
         '''


tests/unit/ssh/test_ssh.py
```diff
@@ -14,10 +14,12 @@
 
 # Import Salt libs
 import salt.config
+import salt.utils
 from salt.client import ssh
 
 
 @skipIf(NO_MOCK, NO_MOCK_REASON)
+@skipIf(not salt.utils.which('ssh'), "No ssh binary found in path")
```
rallytime: This should use the new utils location `salt.utils.path.which` (and the import should be changed as well). Otherwise 👍

bdrung: Fixed. Switching between the 2017.7 and 2018.3 branch made me do this mistake.

rallytime (approved):

PR #46423, opened 2018-03-07
cleanup getaddrinfo() usage in dns_check()

moench-tegeder:
We can pass the desired address family (AF_INET or AF_INET6) directly
to getaddrinfo() instead of filtering later. To preserve current behaviour,
not specifying the ipv6 flag (ipv6 is None) returns whatever the OS default
is (most possibly, both). Just don't blame me for python's syntax, sorry
about that.

Further, pass the service to connect to to getaddrinfo() and use
getaddrinfo()'s output as the port to connect to - getaddrinfo() also
resolves service names, so we can now names ("https") instead of port
numbers ("443") for free.

Lastly, remove the preference for IPv4 addresses: even if we wanted
to accept both IPv4 and IPv6 (thus setting ipv6=None) only IPv4-
addressed were saved in candidates - and only the first address from
candidates would be returned if the connect test failed. If the given
hostname resolved to only IPv6 addresses, no address would be returned
if the connect test failed. Instead, we just rely on getaddrinfo()
returning the desired address family(s) automatically.

In passing, clarify comment about returning an IP address from the input
without further ado.

### Previous Behavior

### New Behavior

### Tests written?

Yes/No

### Commits signed with GPG?

Yes/No


moench-tegeder:
This is not trivial to backpatch as the code in question was moved recently. I suggest leaving back branches as they are, as I couldn't find any complaints about this.

garethgreenaway (approved):

PR #46424, opened 2018-03-07
better reporting/handling of Linode API errors

rmcintosh:
### What does this PR do?
Raises an exception in the _query function of the Linode salt-cloud module when the Linode API reports back an error (or errors)

This isn't a massive improvement but an easy win, mostly just aiming to stop the silent failing/bizarre error messages.

### Previous Behavior
For example, in the case of an API authentication failure:
- certain calls like avail.locations would silent fail
- other things, like attempting to create a Linode would fail with confusing errors related to functions like avail_sizes() barfing on the lack of expected data

### New Behavior
Now, we'll raise a SaltCloudSystemExit exception if an API authentication failure occurs (a fatal error). Otherwise, we raise a plain SaltCloudException noting the error messages from the API.

I'm not certain, in context, that we actually again anything by raising SaltCloudSystemExit instead of a generic SaltCloudException - but it seems more right?

### Tests written?

No

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/cloud/clouds/linode.py
+++ b/salt/cloud/clouds/linode.py
@@ -1500,6 +1500,24 @@ def _query(action=None,
         hide_fields=['api_key', 'rootPass'],
         opts=__opts__,
     )
+
+    if 'ERRORARRAY' in result['dict']:
+        if len(result['dict']['ERRORARRAY']):
+            error_list = []
+
+            for error in result['dict']['ERRORARRAY']:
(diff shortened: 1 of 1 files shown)


gtmanfred (approved):

rmcintosh:
just realized I was raising before checking the length of the error array ... fixed that

rallytime:
Welcome @rmcintosh! Thank you for submitting this fix.

Issue #46334, opened 2018-03-05
[2018.3.0rc1] Stacktrace on call to nacl.dec


sjorge:
### Description of Issue/Question
```
2018-03-05 18:47:44,563 [salt.pillar                                                 :732 ][CRITICAL][339999] Rendering SLS 'common.account' failed, render error:
Traceback (most recent call last):
  File "/opt/local/lib/python2.7/site-packages/salt/utils/templates.py", line 553, in py
    data = mod.run()
  File "/salt/pillar/common/account/init.sls", line 121, in run
    'password': __salt__['shadow.gen_password'](__salt__['nacl.dec'](data=user['password'], key=nacl['db']), 'BLBnVC', 'sha256') if user['password'] else None,
  File "/opt/local/lib/python2.7/site-packages/salt/modules/nacl.py", line 365, in dec
    return sealedbox_decrypt(data, **kwargs)
  File "/opt/local/lib/python2.7/site-packages/salt/modules/nacl.py", line 436, in sealedbox_decrypt
    sk = _get_sk(**kwargs)
  File "/opt/local/lib/python2.7/site-packages/salt/modules/nacl.py", line 213, in _get_sk
    with salt.utils.files.fopen(sk_file, 'rb') as keyf:
  File "/opt/local/lib/python2.7/site-packages/salt/utils/files.py", line 366, in fopen
    f_handle = open(*args, **kwargs)  # pylint: disable=resource-leakage
IOError: [Errno 2] No such file or directory: u'/etc/salt/pki/master/nacl'
```
### Setup
I call nacl.dec from the __salt__ dict inside a pillar.

```__salt__['nacl.dec'](data=user['password'], key=nacl['db'])```
where user['password'] hold the encrypted password and nacl['db'] holds the key.

This is using the pyrendered as I pull this data from a database, it works fine on 2017.7.4 and has been working fine since somewhere 2016.x at least.

Which is interesting is the path it cannot find: ```/etc/salt/pki/master/nacl```
My salt package etc is /opt/local/etc/salt. looks like ```salt.modules.nacl._get_config``` has this hardcoded! Those should probably be updates to be relative to whatever pki dir is configured!

### Steps to Reproduce Issue
Call ```__salt__['nacl.dec']``` from a pillar.

### Versions Report
```
cronos# salt-call --versions-report
Salt Version:
           Salt: 2018.3.0rc1

Dependency Versions:
           cffi: 1.9.1
       cherrypy: 8.1.0
[...]

PR #46426, opened 2018-03-07
[2018.3.0rc1] fixes to nacl module & runner

garethgreenaway:
### What does this PR do?
Fixing a few bugs introduced following the move to unicode all the things.  Adding deprecation notices for the previously used `key` and `key_file` arguments.  Adding deprecation into Release notes.  Adding some tests for Naccl module and runner to test encryption and decryption.

#46334 

### Tests written?
Yes

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/modules/nacl.py
+++ b/salt/modules/nacl.py
@@ -164,7 +164,6 @@ import salt.utils.stringutils
 import salt.utils.win_functions
 import salt.utils.win_dacl
 
-
 REQ_ERROR = None
 try:
     import libnacl.secret
@@ -186,9 +185,9 @@ def _get_config(**kwargs):
     config = {
         'box_type': 'sealedbox',
         'sk': None,
-        'sk_file': '/etc/salt/pki/master/nacl',
+        'sk_file': os.path.join(__opts__['pki_dir'], 'master/nacl'),
         'pk': None,
-        'pk_file': '/etc/salt/pki/master/nacl.pub',
+        'pk_file': os.path.join(__opts__['pki_dir'], 'master/nacl.pub'),
     }
     config_key = '{0}.config'.format(__virtualname__)
     try:
@@ -233,7 +232,7 @@ def _get_pk(**kwargs):
     return base64.b64decode(pubkey)
 
 
-def keygen(sk_file=None, pk_file=None):
+def keygen(sk_file=None, pk_file=None, **kwargs):
     '''
     Use libnacl to generate a keypair.
 
@@ -253,6 +252,14 @@ def keygen(sk_file=None, pk_file=None):
         salt-call nacl.keygen sk_file=/etc/salt/pki/master/nacl pk_file=/etc/salt/pki/master/nacl.pub
         salt-call --local nacl.keygen
     '''
+    if 'keyfile' in kwargs:
+        salt.utils.versions.warn_until(
+            'Fluorine',
+            'The \'keyfile\' argument has been deprecated and will be removed in Salt '
+            '{version}. Please use \'sk_file\' argument instead.'
+        )
+        sk_file = kwargs['sk_file']
+
     if sk_file is None:
         kp = libnacl.public.SecretKey()
         return {'sk': base64.b64encode(kp.sk), 'pk': base64.b64encode(kp.pk)}
@@ -313,6 +320,25 @@ def enc(data, **kwargs):
 
     box_type: secretbox, sealedbox(default)
     '''
+    if 'keyfile' in kwargs:
+        salt.utils.versions.warn_until(
+            'Fluorine',
+            'The \'keyfile\' argument has been deprecated and will be removed in Salt '
+            '{version}. Please use \'sk_file\' argument instead.'
+        )
+        kwargs['sk_file'] = kwargs['keyfile']
+
+    if 'key' in kwargs:
+        salt.utils.versions.warn_until(
+            'Fluorine',
+            'The \'key\' argument has been deprecated and will be removed in Salt '
+            '{version}. Please use \'sk\' argument instead.'
+        )
+        kwargs['sk'] = kwargs['key']
+
+    # ensure data is in bytes
+    data = salt.utils.stringutils.to_bytes(data)
+
     box_type = _get_config(**kwargs)['box_type']
     if box_type == 'sealedbox':
         return sealedbox_encrypt(data, **kwargs)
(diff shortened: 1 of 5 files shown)


rallytime (approved):
Nicely done.

salt/modules/nacl.py
```diff
@@ -253,6 +252,14 @@ def keygen(sk_file=None, pk_file=None):
         salt-call nacl.keygen sk_file=/etc/salt/pki/master/nacl pk_file=/etc/salt/pki/master/nacl.pub
         salt-call --local nacl.keygen
     '''
+    if 'keyfile' in kwargs:
+        salt.utils.versions.warn_until(
+            'Fluorine',
+            'The \'keyfile\' argument has been deprecated and will be removed in Salt '
+            '{version}. Please use \'sk_file\' argument instead.'
+        )
+        sk_file = kwargs['sk_file']
```
sjorge: should this not be kwargs['keyfile'] ? if assume the idea to to warn and keep the old stuff working until Flourine.

sjorge (changes requested):
There is a small bug i think, I added a line comment.

garethgreenaway:
@sjorge Yes!  Thanks!  Thought I caught all of those.

rallytime:
@garethgreenaway Looks like a couple of test failures here: https://jenkins.saltstack.com/job/PR/job/salt-pr-rs-cent7-n/16998/

garethgreenaway:
@rallytime looks like I need to gate these to ensure libnacl is installed.  Thanks!

salt/runners/nacl.py
```diff
@@ -248,11 +211,20 @@ def keygen(sk_file=None, pk_file=None):
...
+        salt-run nacl.keygen
     '''
+
+    if 'keyfile' in kwargs:
+        salt.utils.versions.warn_until(
+            'Fluorine',
+            'The \'keyfile\' argument has been deprecated and will be removed in Salt '
+            '{version}. Please use \'sk_file\' argument instead.'
+        )
+        sk_file = kwargs['sk_file']
```
sjorge: This one is also still wrong!

sjorge (changes requested):

sjorge:
Out of scope for this fix, the code between the module and runner is nearly the same. Maybe we should have a salt.utils.nacl eventually?

salt/runners/nacl.py
```diff
@@ -411,12 +420,13 @@ def sealedbox_encrypt(data, **kwargs):
...
-        salt-call --local nacl.sealedbox_encrypt datatoenc pk_file=/etc/salt/pki/master/nacl.pub
-        salt-call --local nacl.sealedbox_encrypt datatoenc pk='vrwQF7cNiNAVQVAiS3bvcbJUnF0cN6fU9YTZD9mBfzQ='
     '''
+    # ensure data is bytes
+    data = salt.utils.stringutils.to_bytes(data)
+
     pk = _get_pk(**kwargs)
     b = libnacl.sealed.SealedBox(pk)
-    return base64.b64encode(b.encrypt(data))
+    return base64.b64encode(b.encrypt(salt.utils.stringutils.to_bytes(data)))
```
sjorge: Sorry to nitpick at this, while applying the changes to locally I noticed this line.
I don't think this one is present in salt.modules.nacl yet the function there is the same as here!

sjorge:
With the patch applied (with a few more sk_file -> keyfile) I still hit the error. With the following bit in my py-rendered pillar.

```
__salt__['nacl.dec'](data=user['password'], key=nacl['db'])
```

We also need to deprecate the 'key' parameter and have it copied to 'sk' until Flourine.

rallytime:
@sjorge:

> Out of scope for this fix, the code between the module and runner is nearly the same. Maybe we should have a salt.utils.nacl eventually?

Yeah, I think that is a good idea. Something like that would need to go into develop for a future feature release.

@garethgreenaway Can you follow up on the remaining comments?

garethgreenaway:
:+1: for salt.utils.nacl.  The code duplication between the two was one of the things I had noticed as well and should be fairly easy to move it into a shared location.

garethgreenaway:
All the references to `sk` or `sk_file` that should be `key` or `key_file` respectively have been updated.  The unnecessary duplicate `to_bytes` code has been removed.  There was a warning already in place mentioned `key` being deprecated.  And I'm unable to duplicate the issue of calling `nacl.dev` from a py rendered pillar file, are there any errors or additional information in the master logs?

sjorge:
```#!py
######
## PR 46426
## -----------------------------------
######

import os
import hashlib
import binascii
nacl_key = 'Ii5pTnanQPlo7Asp12NLdCsBd1hgo+PD5tEAPw8TW8Q='
def nthash(password):
  return binascii.hexlify(
    hashlib.new(
      'md4',
      password.encode('utf-16le')
    ).digest()
  ).upper()

def run():
  # salt-run nacl.enc data=password key=nacl_key
  enc_password = 'FZlbJiMU1jwJzQh/yLKPgmEIu56HhPg233PfZuvg6zFk8eFrJbc2kF72BNxJkJy30aGf'
  plain_password = __salt__['nacl.dec'](data=enc_password, key=nacl_key)
  return {
    'test': {
       'crypt': __salt__['shadow.gen_password'](plain_password, 'XeBnQC', 'sha256'),
       'lanman': nthash(plain_password),
    }
  }

# vim: tabstop=2 expandtab shiftwidth=2 softtabstop=2```

@garethgreenaway above is a stripped down version of my pillar, this should trigger it.

garethgreenaway:
@sjorge I was able to get your pillar file work by adding an additional argument:
```plain_password = __salt__['nacl.dec'](data=enc_password, key=nacl_key, box_type='secretbox')```

The default is to use the `sealedbox` approach which assumes the data was encrypted using the public key and then will be decrypted using the private key, versus the `secretbox` approach which assumes the data will be encrypted using the same key.

sjorge:
Perhaps we can set box_type to secretbox if box_type is unset and key is passed?

That way all existing code keep working for this release. (I actually already fixed my code to the new system, but probably best not to break peoples stares/pillars in oxygen just yet.)

rallytime:
@garethgreenaway Looks like those disable lint errors aren't quite right: https://jenkins.saltstack.com/job/PR/job/salt-pr-lint-n/19983/violations/

@sjorge Did you get a chance to follow up on the most recent changes?

sjorge:
@rallytime I manually applied the changes from 5a2913ff1c9659f2714640961a88937785d12c66 and that made my pillar current pillar work with a few warnings.

So that looks good to me.

Issue #46346, opened 2018-03-05
Build 449: unit.templates.test_jinja


twangboy:
[unit.templates.test_jinja.TestGetTemplate.test_non_ascii_encoding](https://jenkins.saltstack.com/job/2017.7/job/salt-dev-c7-windows-py2/449/testReport/junit/unit.templates.test_jinja/TestGetTemplate/test_non_ascii_encoding/)

PR #46432, opened 2018-03-08
Default to UTF-8 for templated files

twangboy:
### What does this PR do?
Uses the `SLS_ENCODING`, which is `utf-8`, when processing templated files. If this falls, falls back to system encoding.

https://github.com/saltstack/salt/issues/46346

### Tests written?
No

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/utils/locales.py
+++ b/salt/utils/locales.py
@@ -33,15 +33,12 @@ def get_encodings():
     return encodings
 
 
-def sdecode(string_, encoding=None):
+def sdecode(string_):
     '''
     Since we don't know where a string is coming from and that string will
     need to be safely decoded, this function will attempt to decode the string
     until it has a working string that does not stack trace
     '''
-    if encoding:
-        return salt.utils.to_unicode(string_, encoding)
-
     encodings = get_encodings()
     for encoding in encodings:
         try:
--- a/salt/utils/locales.py
+++ b/salt/utils/locales.py
@@ -17,7 +17,10 @@ def get_encodings():
     '''
     return a list of string encodings to try
     '''
-    encodings = [__salt_system_encoding__]
+    if salt.utils.is_windows():
+        encodings = ['utf-8', __salt_system_encoding__]
+    else:
+        encodings = [__salt_system_encoding__]
 
     try:
         sys_enc = sys.getdefaultencoding()
--- a/salt/utils/locales.py
+++ b/salt/utils/locales.py
@@ -17,10 +17,7 @@ def get_encodings():
     '''
     return a list of string encodings to try
     '''
-    if salt.utils.is_windows():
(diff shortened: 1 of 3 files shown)


twangboy:
Not sure this is the best fix... probably not actually... Ideas?

twangboy:
Here's how unicode is being rendered on Windows and Linux:
Windows
```
>>>'Sàlt'
'S\x85lt'
>>>u'Sàlt'
u'S\xe0lt'
```
Linux
```
>>>'Sàlt'
'S\xc3\xa0lt'
>>>u'Sàlt'
u'S\xe0lt'
```

twangboy:
@terminalmage I think this is a better fix. What do you think?

skizunov (approved):

damon-atkins:
All sls files need to be UTF-8 it can not work any other way, user needs to indicate output encoding.

However in line edits of a file on the minion are a bit harder. Just like working out the EOL character of a file on the minion, you also needs to work out the encoding of the file.

Had thoughts of changing the open function into a class when open for reading always determines the EOL and encoding type of a file.  It far to big of a change.

What is the use case for this test?

terminalmage:
@damon-atkins I think you're missing the point of this, the issue is that UTF-8-encoded bytestrings are decoding successfully (but incorrectly) using the default codepage.

terminalmage (commented):
@twangboy This looks like it would be enough to get the test to pass, but in actual usage we aren't passing the encoding when we render, so it seems like this would still break when a template is rendered on Windows.

twangboy:
@terminalmage But now you can specify an encoding if you need to. I've had to do it in several other tests for Windows, but the option to pass an encoding was already there. This just adds it, and I don't think it will break anything as currently implemented.

damon-atkins:
@twangboy You added that, but is it added up all the way to the top of the chain so the user has control.
Currently if your not using US-ASCII its hit or miss if things work (esp Windows). After 2018 release it should not be.

dwoz:
I'm wondering if we can be a bit more explicit here. Has anyone had a chance to look into detecting the default codepage of the command prompt? 

http://timgolden.me.uk/pywin32-docs/win32console__GetConsoleCP_meth.html

Or if we want to default to UTF-8 maybe we use the set codepage apis.

http://timgolden.me.uk/pywin32-docs/win32console__SetConsoleCP_meth.html

dwoz (approved):
I like this, We are always trying `utf-8` first, and falling back to the system encoding if `utf-8` fails.

twangboy:
@terminalmage Could you take a look... again...

terminalmage (approved):

rallytime:
re-run py3

PR #46441, opened 2018-03-08
Fix suse_apache os_family

bochi:
### What does this PR do?
Fixing os_family for SUSE Systems. This was changed from "Suse" to "SUSE" with commit c9d0cdb All systems I was able to check (SLES11SP4, SLES12SP2+3, openSUSE) have "Suse" as os_family, though I went for the ".upper()" way since they may be cases on SUSE systems that I was not able to check and this will work for both cases.
### Previous Behavior

### New Behavior

### Tests written?

Yes/No

### Commits signed with GPG?

Yes/No

### DIFF
--- a/salt/modules/suse_apache.py
+++ b/salt/modules/suse_apache.py
@@ -23,7 +23,7 @@ def __virtual__():
     '''
     Only load the module if apache is installed.
     '''
-    if salt.utils.path.which('apache2ctl') and __grains__['os_family'] == 'SUSE':
+    if salt.utils.path.which('apache2ctl') and __grains__['os_family'].upper() == 'SUSE':
         return __virtualname__
     return (False, 'apache execution module not loaded: apache not installed.')
 
--- a/salt/modules/suse_apache.py
+++ b/salt/modules/suse_apache.py
@@ -23,7 +23,7 @@ def __virtual__():
     '''
     Only load the module if apache is installed.
     '''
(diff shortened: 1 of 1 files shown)


salt/modules/suse_apache.py
```diff
@@ -23,7 +23,7 @@ def __virtual__():
     '''
     Only load the module if apache is installed.
     '''
-    if salt.utils.path.which('apache2ctl') and __grains__['os_family'] == 'SUSE':
+    if salt.utils.path.which('apache2ctl') and __grains__['os_family'].upper() == 'SUSE':
```
isbm: This is a wrong fix. It has to look for `Suse`, not modifying anything just because it is a module, like so:
```python
.... and __grains__['os_family'] == 'Suse':
```

isbm (changes requested):

bochi:
Done. It is looking for Suse now. I just did the .upper() thing because it was changed from Suse to SUSE in a previous commit: 
https://github.com/saltstack/salt/commit/c9d0cdb2c82c4c12b8c42580e2fb7545a9f0c272
...and figured there might be a reason for this.

rallytime (approved):

rallytime:
Thanks @bochi and congrats on your first PR to Salt! :)

isbm:
@bochi this is because @benediktwerner got it wrong too. :wink: Thanks!

PR #46442, opened 2018-03-08
Fix unicode decode error with salt-ssh

mcalmer:
### What does this PR do?

with salt-ssh, dumps() is called and the output went in find_json() where
it can produce a decode error. As we encode first to str() we should decode
the output back when we run with python2.

### Tests written?

No

### Commits signed with GPG?

No

### DIFF
--- a/salt/utils/json.py
+++ b/salt/utils/json.py
@@ -128,4 +128,7 @@ def dumps(obj, **kwargs):
         kwargs['ensure_ascii'] = False
     if six.PY2:
         obj = salt.utils.data.encode(obj)
-    return json_module.dumps(obj, **kwargs)  # future lint: blacklisted-function
+    ret = json_module.dumps(obj, **kwargs)  # future lint: blacklisted-function
+    if six.PY2:
+        return ret.decode(__salt_system_encoding__)
+    return ret


mcalmer:
I did not manage to write a test for it but it is easy to test when setting a grain in a roaster file
```
host1:
    host: <hostname>
    user: root
    passwed: secure
    grains:
        zz_unicode: öäü
```

The following command will not return expected data:
```
salt-ssh <host> grains.items
```

mcalmer:
@terminalmage - It seems that some tests fail now but I am not sure if the tests are correct. I fear somebody with more knowledge in this area need to find out what the right fix is for my problem.
An example how to reproduce it exists in the comment above.

With python2 you see an unexpected result (stdout key which full json as payload) and with python3 it parse the stuff correctly and you get grain output like expected.

mcalmer:
```
ssh/__init__.py (def handle_routine)   stdout, stderr, retcode = single.run()
ssh/__init__.py(1152)run_wfunc()  ret = salt.utils.json.dumps({'local': result['local']})
ssh/__init__.py (def handle_routine)   data = salt.utils.json.find_json(stdout)
salt/utils/json.py(29)find_json() working = '\n'.join(raw.splitlines()[ind:])
=> UnicodeDecodeError: UnicodeD...ge(128)')
```

terminalmage:
This is not the correct fix, nor the actual cause of the problem. The actual cause is _way_ weirder, as I just found out while debugging this.

The correct fix is in https://github.com/saltstack/salt/pull/46477. I'm going to close this in favor of that PR. @mcalmer can you test and confirm?

Thanks for helping to bring this to our attention.

mcalmer:
@terminalmage - I was finally able to test it and it works.

PR #46446, opened 2018-03-08
Fix various typos

bdrung:
Lintian found various typos. The first commit fixes the trivial typos. The second commit fixes the typo in the heat module/state parameter. The rename is documented and support for the misspelled version for backward compatibility is added.

### DIFF
--- a/salt/modules/heat.py
+++ b/salt/modules/heat.py
@@ -223,7 +223,7 @@ def _parse_template(tmpl_str):
     return tpl
 
 
-def _parse_enviroment(env_str):
+def _parse_environment(env_str):
     '''
     Parsing template
     '''
@@ -460,9 +460,9 @@ def delete_stack(name=None, poll=0, timeout=60, profile=None):
     return ret
 
 
-def create_stack(name=None, template_file=None, enviroment=None,
+def create_stack(name=None, template_file=None, environment=None,
                  parameters=None, poll=0, rollback=False, timeout=60,
-                 profile=None):
+                 profile=None, enviroment=None):
     '''
     Create a stack (heat stack-create)
 
@@ -472,8 +472,8 @@ def create_stack(name=None, template_file=None, enviroment=None,
     template_file
         File of template
 
-    enviroment
-        File of enviroment
+    environment
+        File of environment
 
     parameters
         Parameter dict used to create the stack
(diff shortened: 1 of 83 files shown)


salt/modules/heat.py
```diff
@@ -496,11 +496,19 @@ def create_stack(name=None, template_file=None, enviroment=None,
...
 
+    .. versionadded:: 2017.7.5,2018.3.1
+
+        The spelling mistake in parameter `enviroment` was corrected to `environment`.
+        The misspelled version is still supported for backward compatibility.
+
     '''
+    if environment is None:
+        # Support for misspelled parameter
+        environment = enviroment
```
rallytime: I think we should put the misspelled version on a deprecation path. Look for the `warn_util` function of how to do this and we can remove it in 2 feature releases.

salt/modules/heat.py
```diff
@@ -681,11 +689,19 @@ def update_stack(name=None, template_file=None, enviroment=None,
...
```
rallytime: Same comment here. :D

salt/states/heat.py
```diff
@@ -152,10 +157,18 @@ def deployed(name, template=None, enviroment=None, params=None, poll=5,
...
 
+    .. versionadded:: 2017.7.5,2018.3.1
+
+        The spelling mistake in parameter `enviroment` was corrected to `environment`.
+        The misspelled version is still supported for backward compatibility.
+
     '''
+    if environment is None and 'enviroment' in connection_args:
+        # Support for misspelled parameter
+        environment = connection_args.pop('enviroment')
```
rallytime: The `warn_util` deprecation notice should happen here as well.

rallytime (changes requested):
Thanks @bdrung! It's always nice to get these cleaned up.

I have one request, but once that is changed, this looks good to me.

bdrung:
Took me some seconds to figure out that you meant `warn_until` instead of `warn_util`. :smile: Implemented deprecation warning as requested.

rallytime:
Haha woops! Speaking of typos.... 😁 

This looks great now!

rallytime (approved):

bdrung:
Look at the diff and you will find typo fixes for commit messages that fix typos. :laughing:

terminalmage (approved):

PR #46447, opened 2018-03-08
Use integer parameter for sleep command

bdrung:
The bootstrap-salt.sh script uses a "bashism" for the sleep command. Bash allows real numbers as sleep time, but other POSIX shell interpreters only support an integer as parameter. Thus raise the sleep timeout from 0.1 to 1 second.

The git history does not explain why a sleep command is used in the code. You might be able to remove the sleep command completely.

Bug-Debian: https://bugs.debian.org/772406

### DIFF
--- a/salt/cloud/deploy/bootstrap-salt.sh
+++ b/salt/cloud/deploy/bootstrap-salt.sh
@@ -2791,7 +2791,7 @@ install_ubuntu_stable_post() {
                 /bin/systemctl preset salt-$fname.service > /dev/null 2>&1 &&
                 /bin/systemctl enable salt-$fname.service > /dev/null 2>&1
             )
-            sleep 0.1
+            sleep 1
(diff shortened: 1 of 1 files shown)


rallytime:
Hi @bdrung -  so it's not exactly intuitive what the process is for updating the bootstrap script, but actually any change made in salt will be overwritten when a new bootstrap release comes out. Can you make changes to the bootstrap script in [the bootstrap repo](https://github.com/saltstack/salt-bootstrap) instead?

Thanks again for all of these great contributions! :D

rallytime:
Oh, and for your question about the git history - there might be some more clues there in the bootstrap repo since the bootstrap script just gets completely replaced in Salt with the new stable release.

bdrung:
Done in https://github.com/saltstack/salt-bootstrap/pull/1205

PR #46459, opened 2018-03-09
New grains fetching detailed hypervisor info for Xen domains

ixs:
### What does this PR do?
Add new grains to parse hypervisor details

### New Behavior
Xen exposes a lot of detail about the hypervisor in /sys/hypervisor.

Parse the hypervisor version and the capabilities and expose these
as grains.

This allows for detailed targeting and appropriate actions, e.g. when
running on XenServer install the XenServer guest-tools.

Sample output:

    virtual_hv_features:
        000020f0
    virtual_hv_features_list:
        - pae_pgdir_above_4gb
        - mmu_pt_update_preserve_ad
        - gnttab_map_avail_bits
        - memory_op_vnode_supported
    virtual_hv_version:
        4.6.6-2.el7
    virtual_hv_version_info:
        - 4
        - 6
        - .6-2.el7

### Tests written?
No. It looked like the whole _virtual part has no tests at all.

### Commits signed with GPG?
No


damon-atkins:
Should these grains be documented?  Is there any page/doco describing grains?

ixs:
@damon-atkins I do not believe there is any documentation.

If you look at https://docs.saltstack.com/en/latest/topics/grains/, it links straight to https://github.com/saltstack/salt/blob/develop/salt/grains/core.py for "documentation". And otherwise it suggests to use the grains module to discover what is available.

garethgreenaway (approved):

ixs:
@rallytime done.

rallytime:
@ixs Looks like there are still some errors. You can run the lint check yourself if you like by following [these instructions](https://docs.saltstack.com/en/latest/topics/development/conventions/style.html#linting).

salt/grains/core.py
```diff
@@ -983,6 +984,59 @@ def _virtual(osdata):
...
+
+    # Bail early if we're not running on Xen
+    if 'xen' not in osdata['virtual']:
+        return grains
+
+    # Try to get the exact hypervisor version from sysfs
+    try:
+        version = {}
+        for file in ('major', 'minor', 'extra'):
+            with salt.utils.fopen('/sys/hypervisor/version/{}'.format(file), 'r') as fhr:
```
rallytime: Can you use the `salt.utils.files.fopen` path here, instead of the `salt.utils.fopen`? We've split out the utils init file into separate files in 2018.3 and develop.

salt/grains/core.py
```diff
@@ -983,6 +984,59 @@ def _virtual(osdata):
...
+    grains = {}
+
+    # Bail early if we're not running on Xen
+    if 'xen' not in osdata['virtual']:
+        return grains
+
+    # Try to get the exact hypervisor version from sysfs
+    try:
+        version = {}
+        for file in ('major', 'minor', 'extra'):
```
rallytime: `file` is a reserved keyword in Python. Can you change this to another variable name?

salt/grains/core.py
```diff
@@ -983,6 +984,59 @@ def _virtual(osdata):
...
+                         7: 'gnttab_map_avail_bits',
+                         8: 'hvm_callback_vector',
+                         9: 'hvm_safe_pvclock',
+                        10: 'hvm_pirqs',
+                        11: 'dom0',
+                        12: 'grant_map_identity',
+                        13: 'memory_op_vnode_supported',
+                        14: 'ARM_SMCCC_supported'}
+    try:
+        with salt.utils.fopen('/sys/hypervisor/properties/features', 'r') as fhr:
```
rallytime: Same comment here about the utils path. :)

rallytime (changes requested):
I have a couple of requested changes here, too.

salt/grains/core.py
ixs: Right. In that case you might also want to update your linter messages as that referes to salt.utils.fopen()

salt/grains/core.py
ixs: Done.

salt/grains/core.py
ixs: Done

salt/grains/core.py
```diff
@@ -984,6 +984,60 @@ def _virtual(osdata):
...
+
+    # Try to get the exact hypervisor version from sysfs
+    try:
+        version = {}
+        for fn in ('major', 'minor', 'extra'):
+            with salt.utils.files.fopen('/sys/hypervisor/version/{}'.format(fn), 'r') as fhr:
+                version[fn] = fhr.read().strip()
+        grains['virtual_hv_version'] = '{}.{}{}'.format(version['major'], version['minor'], version['extra'])
+        grains['virtual_hv_version_info'] = (version['major'], version['minor'], version['extra'])
+    except:
```
terminalmage: Bare `except`s are bad, at minimum `except Exception` should be used. But when you know the potenial exceptions, you should use them. In this case, `except (IOError, OSError, KeyError):`

salt/grains/core.py
```diff
@@ -984,6 +984,60 @@ def _virtual(osdata):
...
+        return grains
+
+    # Try to get the exact hypervisor version from sysfs
+    try:
+        version = {}
+        for fn in ('major', 'minor', 'extra'):
+            with salt.utils.files.fopen('/sys/hypervisor/version/{}'.format(fn), 'r') as fhr:
+                version[fn] = fhr.read().strip()
+        grains['virtual_hv_version'] = '{}.{}{}'.format(version['major'], version['minor'], version['extra'])
+        grains['virtual_hv_version_info'] = (version['major'], version['minor'], version['extra'])
```
terminalmage: We should be using a list instead of a tuple here, for a couple reasons:

1) The tuple will be lost (converted to list) anyway if/when the data is serialized using msgpack, json, etc.
2) We allow grains to be added via YAML, which does not support tuples.

salt/grains/core.py
```diff
@@ -984,6 +984,60 @@ def _virtual(osdata):
...
+    # Bail early if we're not running on Xen
+    if 'xen' not in osdata['virtual']:
+        return grains
+
+    # Try to get the exact hypervisor version from sysfs
+    try:
+        version = {}
+        for fn in ('major', 'minor', 'extra'):
+            with salt.utils.files.fopen('/sys/hypervisor/version/{}'.format(fn), 'r') as fhr:
+                version[fn] = fhr.read().strip()
```
terminalmage: This needs to be decoded to unicode using:

```python
version[fn] = salt.utils.stringutils.to_unicode(fhr.read().strip())
```

This may also require that `salt.utils.stringutils` is added to the imports.

salt/grains/core.py
```diff
@@ -984,6 +984,60 @@ def _virtual(osdata):
...
+                         8: 'hvm_callback_vector',
+                         9: 'hvm_safe_pvclock',
+                        10: 'hvm_pirqs',
+                        11: 'dom0',
+                        12: 'grant_map_identity',
+                        13: 'memory_op_vnode_supported',
+                        14: 'ARM_SMCCC_supported'}
+    try:
+        with salt.utils.files.fopen('/sys/hypervisor/properties/features', 'r') as fhr:
+            features = fhr.read().strip()
```
terminalmage: Same as above with decoding to unicode.

salt/grains/core.py
```diff
@@ -984,6 +984,60 @@ def _virtual(osdata):
...
+        enabled_features = []
+        for bit, feat in xen_feature_table.items():
```
terminalmage: Instead of `xen_feature_table.items()`, please use `six.iteritems(xen_feature_table)`. This ensures that Python 2 uses a generator over the key/value pairs.

salt/grains/core.py
```diff
@@ -984,6 +984,60 @@ def _virtual(osdata):
...
+            if int(features, 16) & (1 << bit):
+                enabled_features.append(feat)
+        grains['virtual_hv_features'] = features
+        grains['virtual_hv_features_list'] = enabled_features
+    except:
```
terminalmage: Same as above with the bare exception. Try to narrow this down to the exceptions you'd expect to be raised.

terminalmage (changes requested):

cachedout:
@ixs Are you planning to implement the review feedback on this?

cachedout:
@ixs We did not hear back from you on the review feedback so we are going to close this. When you're ready to come back to it, please leave a comment here and we will happily re-open it. Thanks!

ixs:
@cachedout thanks for the review and the suggestions. Finally gotten around to finish this. Please reopen.

ixs:
@rallytime Thanks. The new update is already up on github and I believe this PR has the latest changes with all requested work done.

rallytime (approved):

terminalmage (approved):

rallytime:
Hi @ixs - This change is causing some test failures in the `test_core.py` file. Can you take a look?

https://jenkins.saltstack.com/job/PR/job/salt-pr-linode-cent7-py3/4121/#showFailuresLink

ixs:
@rallytime Ohh, that's new. Fixed.
Turns out, the virtual grain is of course not set when running _not_ in a virtual environment... But who would ever do such a thing, right?

PR #46477, opened 2018-03-09
salt-ssh: fix JSON load of return data when it contains non-ascii

terminalmage:
For reasons I can't explain, in `salt.utils.json.find_json()` using `.splitlines()` will sometimes convert a unicode string into a list of str types on Python 2. So, that's weird. This can be triggered in salt-ssh whenever there are non-ascii chars in the return data.

```
[DEBUG] raw = u'{"local": {"return": {"foo": "\xf6\xe4\xfc"}}}'
[DEBUG] raw.splitlines() = ['{"local": {"return": {"foo": "\xc3\xb6\xc3\xa4\xc3\xbc"}}}']
```

To resolve this, the UnicodeDecodeError is caught and reattempted with a decoded list.

This fix can be confirmed using a Docker container I have provided:

```
% docker run --rm -it -v $PWD:/testing terminalmage/issues:46477 sh -c '/usr/sbin/sshd -D & salt-ssh foobar grains.item foo'
foobar:
    ----------
    foo:
        öäü
```

This command should be run from the root of a git clone of Salt, as it will mount the git repo into that container and run salt-ssh from the mounted repo. [This walkthrough](https://help.github.com/articles/checking-out-pull-requests-locally/) can be used to check out this PR into your clone.

### DIFF
--- a/salt/utils/json.py
+++ b/salt/utils/json.py
@@ -25,8 +25,13 @@ def find_json(raw):
     string to start with garbage and end with json but be cleanly loaded
     '''
     ret = {}
-    for ind, _ in enumerate(raw):
-        working = '\n'.join(raw.splitlines()[ind:])
+    lines = raw.splitlines()
+    for ind, _ in enumerate(lines):
+        try:
+            working = '\n'.join(lines[ind:])
+        except UnicodeDecodeError:
+            working = '\n'.join(salt.utils.data.decode(lines[ind:]))
+
         try:
             ret = json.loads(working)  # future lint: blacklisted-function
         except ValueError:
--- a/salt/utils/json.py
+++ b/salt/utils/json.py
@@ -19,13 +19,22 @@ from salt.ext import six
 log = logging.getLogger(__name__)
 
 
+def __split(raw):
+    '''
+    Performs a splitlines on the string. This function exists to make mocking
+    possible in unit tests, since the member functions of the str/unicode
+    builtins cannot be mocked.
+    '''
+    return raw.splitlines()
+
+
 def find_json(raw):
     '''
     Pass in a raw string and load the json when it starts. This allows for a
     string to start with garbage and end with json but be cleanly loaded
     '''
     ret = {}
-    lines = raw.splitlines()
+    lines = __split(raw)
     for ind, _ in enumerate(lines):
         try:
             working = '\n'.join(lines[ind:])
--- a/tests/support/helpers.py
+++ b/tests/support/helpers.py
@@ -24,6 +24,7 @@ import signal
 import socket
 import string
 import sys
+import tempfile
 import threading
 import time
 import tornado.ioloop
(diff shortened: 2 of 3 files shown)


terminalmage:
@thatch45 Is there a reason why we are joining a splitlines with a newline? It looks like you initially added this function.

cachedout (commented):
Would a quick unit test be appropriate here?

mcalmer:
I tested this fix in our environment and it works.

salt/utils/json.py
```diff
@@ -26,7 +26,12 @@ def find_json(raw):
     '''
     ret = {}
     for ind, _ in enumerate(raw):
```
DmitryKuzmenko: I want to say more here.
This `for` loops iterates the `raw` that is a string according the docstring. So `ind` here is a string character number. But below it's used as a line number in `raw` string data.
So as for me overall logic in this function should look like this:
```python
def find_json(raw):
    '''
    Pass in a raw string and load the json when it starts. This allows for a
    string to start with garbage and end with json but be cleanly loaded
    '''
    ret = {}
    for ind, _ in enumerate(raw):
        working = raw[ind:]
        try:
            ret = json.loads(raw[ind:])  # future lint: blacklisted-function
        except ValueError:
            continue
        if ret:
            return ret
    if not ret:
        # Not json, raise an error
        raise ValueError
```

DmitryKuzmenko:
Potential fix for centos 6 py2 SSHState tests failures: https://github.com/saltstack/salt-jenkins/issues/805

salt/utils/json.py
terminalmage: I talked with @thatch45 and confirmed the code should try to load json one line at a time, not one character at a time. This is the reason for splitting and joining based on a line index. Using the `enumerate()` on the entire string appears to have been a simple oversight. In reality, this would have resulted in the for loop doing a _lot_ of extra `json.loads()` calls on empty strings, since slicing returns an empty list when the slice indices are outside of the actual size of the list:

```python
>>> x = ['foo', 'bar']
>>> x[5:]
[]
```

Joining an empty list like this would produce an empty string, and loading an empty string using `json.loads()` raises a `ValueError`, which we are catching (hence this not being caught before).

So, a more correct change would be:

```python
ret = {}
lines = raw.splitlines()
for ind, _ in enumerate(lines):
    try:
        working = '\n'.join(lines[ind:])
    except UnicodeDecodeError:
        working = '\n'.join(salt.utils.data.decode(lines[ind:]))

    try:
        ret = json.loads(working)  # future lint: blacklisted-function
    except ValueError:
        continue
    if ret:
        return ret
if not ret:
    # Not json, raise an error
    raise ValueError
```

terminalmage: Working on this fix now, as well as a unit test.

terminalmage:
@mcalmer Thanks for the confirmation!

I've made the edits that @DmitryKuzmenko and I talked about [here](https://github.com/saltstack/salt/pull/46477#discussion_r174109774), and also added a unit test.

rallytime (approved):

DmitryKuzmenko (approved):
:tada:

rallytime:
@terminalmage Lint: https://jenkins.saltstack.com/job/PR/job/salt-pr-lint-n/20088/violations/file/tests/unit/utils/test_json.py/

terminalmage:
Fixed.

PR #46483, opened 2018-03-11
Fixes stdout user environment corruption

psyer:
### What does this PR do?
This fix will set a marker before the users env is created. Make the env var.
Then set a marker after it is done. Then will grab the real env info between
the markers, and throw away the markers. Any stdout noise is then mitigated.

When using PAM module pam_lastlog.so with config option "showfailed", it
outputs a "Last login ..." line to stdout everytime a shell is spawned.
This corrupts a random env var and causes havoc in paths and the like.

To reproduce: PAM option was put in /etc/pam.d/su file with line:
session required pam_lastlog.so showfailed

then: salt minionid cmd.run runas=username env

Which will show "Last login ..." in one of the env vars.

### Previous Behavior
    LOGNAME=username
    XDG_SESSION_ID=c4Last login: Sat Mar 10 23:49:26 EST 2018 from 192.168.0.55 on pts/3
    LC_COLLATE=C

### New Behavior
    LOGNAME=username
    XDG_SESSION_ID=c9
    LC_COLLATE=C

### Tests written?
No

### Commits signed with GPG?
No

### DIFF
--- a/salt/modules/cmdmod.py
+++ b/salt/modules/cmdmod.py
@@ -394,9 +394,13 @@ def _run(cmd,
         try:
             # Getting the environment for the runas user
             # There must be a better way to do this.
+            import uuid
+            marker = '<<<' + str(uuid.uuid4()) + '>>>'
             py_code = (
                 'import sys, os, itertools; '
-                'sys.stdout.write(\"\\0\".join(itertools.chain(*os.environ.items())))'
+                'sys.stdout.write(\"' + marker + '\\n\");'
+                'sys.stdout.write(\"\\0\".join(itertools.chain(*os.environ.items())));'
+                'sys.stdout.write(\"\\n' + marker + '\\n\");'
             )
             if __grains__['os'] in ['MacOS', 'Darwin']:
                 env_cmd = ('sudo', '-i', '-u', runas, '--',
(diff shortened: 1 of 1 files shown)


psyer:
The integration unit test is failing on the exception check code I added. I can not reproduce the error it is generating.  I tried running the command like the unit test does but it works without error.

# cd to salt install dir and start interpreter
>>> import modules.cmdmod
>>> modules.cmdmod.__grains__ = {'os': 'Darwin', 'os_family': 'Solaris'}
>>> modules.cmdmod._run('ls',runas='<user>',shell='/bin/bash')
{'pid': 23547, 'retcode': 0, 'stderr': '', 'stdout': 'Desktop\nDocuments\nDownloads\nMusic\nPictures\nPublic\nTemplates'}

rallytime:
Hi @psyer - Thanks for submitting this. 2016.11 is now in CVE-support only and not receiving regular bug fixes. Can you move this to the 2017.7 branch instead?

psyer:
Hey @rallytime, I sure can. I'll close this one and resubmit to 2017.  I actually have some new code for this that will hopefully expose this integration test build issue as well.

PR #46487, opened 2018-03-12
Pylint cleanups for salt/{modules,states}/state.py

ghost:
Pylint cleanups for salt/{modules,states}/state.py in preparation for another patch (a PR coming in swiftly if this one gets pulled). It'll be easier to add new stuff when pylint outputs less warnings ...

```
salt/modules/state.py:
Your code has been rated at 9.83/10 (previous run: 9.42/10, +0.41)

salt/runners/state.py:
Your code has been rated at 7.05/10 (previous run: 2.95/10, +4.10) 
```

### What does this PR do?

Cleans up Pylint warnings from salt/{modules,states}/state.py.

Code quality in terms of maintainability and readability.

### Tests written?

No

### Commits signed with GPG?

No


salt/modules/state.py
```diff
@@ -22,6 +22,8 @@
 import tempfile
 import time
 
+import msgpack
```
rallytime: If you're going to move this import anyway, can you move it to the list of Python imports after the `logging` import?

rallytime (commented):
Welcome to @samilaine!

Thanks for submitting this. I have one suggestion, otherwise looks good.

salt/modules/state.py
samilaine: I could, but SaltStack's own `.pylintrc` dictates that I shouldn't. Once `import msgpack` is placed where it most naturally -- in my opinion too -- would fit, pylint output gets pretty ugly:

```
C: 19, 0: standard import "import os" should be placed before "import msgpack"
(wrong-import-order)                                             
C: 20, 0: standard import "import shutil" should be placed before "import msgpa
ck" (wrong-import-order)                                         
C: 21, 0: standard import "import sys" should be placed before "import msgpack"
 (wrong-import-order)                                            
C: 22, 0: standard import "import tarfile" should be placed before "import msgp
ack" (wrong-import-order)                                        
C: 23, 0: standard import "import tempfile" should be placed before "import msg
pack" (wrong-import-order)                                       
C: 24, 0: standard import "import time" should be placed before "import msgpack
" (wrong-import-order)         
```

rallytime: Ah, we use the `.testing.pylintrc` for our Jenkins tests. It includes some salt-specific checks and ignores some of the checks we don't care about as much that can cause noise on PR submissions. (Import order is one of the ones ignored. So is line length.)

samilaine: I wonder if there is something wrong still or is the PR acceptable? I'd like to build on top of this instead of putting my other PRs against code generating more warnings from pylint.

I believe the move is justified because SaltStack's `.pylintrc` dictates that 3rd party imports should be _before_ imports of salt.* modules and that 3rd party imports should be _before_ modules shipped with Python. Or have I understood something wrong?

PR #46490, opened 2018-03-12
Managed os disk support and show instance action improvements

sumeetisp:
### What does this PR do?
Previously, supported only virtual machine with un-managed os disk. With this change virtual machine with managed os disk can be provisioned.

Secondly, show_instance listed data_disk value as blank because an object was returned. Now the object is unpacked to list all the properties and list it under data disks

Virtual machine provisioning with managed os disk and show instance value of data disks

### Previous Behavior
Provisioning virtual machine only with un-managed os disk and blank value for data disk in show_instance action

### New Behavior
Now virtual machine with both type of os disks i.e managed and unmanaged can be provisioned.
Show instance will now list all the data disks associated to virtual machine.

### Tests written?

Yes/No

### Commits signed with GPG?

Yes/No


gtmanfred (changes requested):
This pr should probably be made against the `develop` branch, since it is adding functionality.

sumeetisp:
Ok then I will close this pull request.

Thanks,
Sumeet Salvankar

Issue #46425, opened 2018-03-07
[2018.3.0rc1] Errors/Warnings in Windows packages 


frogunder:
### Description of Issue/Question
I see Errors/Warnings in the Windows packages for 2018.3.0rc1.

I see this in Win 2012, 2016, 8.1, and 10
```
    [ERROR   ] Exception raised when processing __virtual__ function for salt.loaded.int.module.win_snmp. Module will not be loaded: expected str or bytearray
    [WARNING ] salt.loaded.int.module.win_snmp.__virtual__() is wrongly returning `None`. It should either return `True`, `False` or a new name. If you're the developer of the module 'win_snmp', please fix this.
    [WARNING ] C:\salt\bin\lib\site-packages\salt\modules\win_update.py:92: DeprecationWarning: The 'win_update' module is being deprecated and will be removed in Salt Fluorine (Unreleased). Please use the 'win_wua' module instead.
```
I see this in Win 7 and 2008
```
    [ERROR   ] Command 'try {Get-WmiObject -ErrorAction Stop -class MSFC_FibrePortHBAAttributes -namespace "root\WMI" | Select -Expandproperty Attributes | %{($_.PortWWN | % {"{0:x2}" -f $_}) -join ""} | ConvertTo-JSON} catch { "{}" | ConvertTo-JSON}' failed with return code: 1
    [ERROR   ] output: The term 'ConvertTo-JSON' is not recognized as the name of a cmdlet, function, 
    
    script file, or operable program. Check the spelling of the name, or if a path 
    
    was included, verify that the path is correct and try again.
    
    At line:1 char:226
    
    + try {Get-WmiObject -ErrorAction Stop -class MSFC_FibrePortHBAAttributes -name
    
    space "root\WMI" | Select -Expandproperty Attributes | %{($_.PortWWN | % {"{0:x
    
    2}" -f $_}) -join ""} | ConvertTo-JSON} catch { "{}" | ConvertTo-JSON <<<< }
    
        + CategoryInfo          : ObjectNotFound: (ConvertTo-JSON:String) [], Comm 
    
       andNotFoundException
    
        + FullyQualifiedErrorId : CommandNotFoundException
    [ERROR   ] Error converting PowerShell JSON return
    Traceback (most recent call last):
      File "C:\salt\bin\lib\site-packages\salt\modules\cmdmod.py", line 3233, in powershell
        return salt.utils.json.loads(response)
[...]

PR #46501, opened 2018-03-12
Remove ConvertTo-JSON on Win 7 / 2008R2 and lower

twangboy:
### What does this PR do?
Fixes an issue where the grain that detects FibreChannel was failing on Windows 7 / Server 2008 R2. The `ConvertTo-JSON` powershell command is not available on those versions by default

https://github.com/saltstack/salt/issues/46425

### Tests written?
No

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/modules/cmdmod.py
+++ b/salt/modules/cmdmod.py
@@ -12,6 +12,7 @@ import functools
 import glob
 import logging
 import os
+import platform
 import shutil
 import subprocess
 import sys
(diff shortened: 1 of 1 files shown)


salt/modules/cmdmod.py
```diff
@@ -3184,9 +3185,14 @@ def powershell(cmd,
         python_shell = True
 
     # Append PowerShell Object formatting
-    cmd += ' | ConvertTo-JSON'
-    if depth is not None:
-        cmd += ' -Depth {0}'.format(depth)
+    # ConvertTo-JSON is only available on Versions of Windows greater than
```
damon-atkins: Would have thought ConvertTo-JSON is PowerShell Version dependant.

damon-atkins (changes requested):

rallytime (approved):

salt/modules/cmdmod.py
twangboy: You're right... should use the shell_info function to get the version of PowerShell that's installed...

PR #46503, opened 2018-03-13
Fixes stdout user environment corruption

psyer:
### What does this PR do?
This fix will set a marker before the users env is created. Make the env var.
Then set a marker after it is done. Then will grab the real env info between
the markers, and throw away the markers. Any stdout noise is then mitigated.

When using PAM module pam_lastlog.so with config option "showfailed", it
outputs a "Last login ..." line to stdout everytime a shell is spawned.
This corrupts a random env var and causes havoc in paths and the like.

To reproduce: PAM option was put in /etc/pam.d/su file with line:
session required pam_lastlog.so showfailed

then: salt <minion> cmd.run runas=<user> env

Which will show "Last login ..." in one of the env vars.

This fix will set a marker before the users env is created. Make the env var.
Then set a marker after it is done. Then will grab the real env info between
the markers, and throw away the markers. Any stdout noise is then mitigated.

### Previous Behavior
...
LOGNAME=username
XDG_SESSION_ID=c4Last login: Sat Mar 10 23:49:26 EST 2018 from 192.168.0.55 on pts/3
LC_COLLATE=C
...

### New Behavior
...
LOGNAME=username
XDG_SESSION_ID=c9
LC_COLLATE=C
...

### Tests written?
No

### Commits signed with GPG?
No

### DIFF
--- a/salt/modules/cmdmod.py
+++ b/salt/modules/cmdmod.py
@@ -462,12 +462,14 @@ def _run(cmd,
             env_encoded = env_encoded[env_mark + len(marker) + 1:]
             env_mark = env_encoded.find('\n' + marker + '\n')
             if env_mark < 0:
+                raise CommandExecutionError(
                     'Environment could not be retrieved for User \'{0}\':'
                     ' second marker, got err={1} out={2}'.format(
                         runas,
                         repr(env_encoded_err),
                         repr(env_encoded_org)
                     )
+                )
             # strip second marker line plus leading newline
             env_encoded = env_encoded[0:env_mark]
 
--- a/salt/modules/cmdmod.py
+++ b/salt/modules/cmdmod.py
@@ -475,7 +475,7 @@ def _run(cmd,
             else:
                 raise CommandExecutionError(
                     'Environment could not be retrieved for User \'{0}\':'
-                    ' missing a marker, got err={1!r} out={2!r}'.format(
+                    ' missing a marker, got err={1} out={2}'.format(
                         runas,
                         repr(env_encoded_err),
                         repr(env_encoded_org)
--- a/salt/modules/cmdmod.py
+++ b/salt/modules/cmdmod.py
@@ -445,9 +445,9 @@ def _run(cmd,
                 stdout=subprocess.PIPE,
                 stdin=subprocess.PIPE
             ).communicate(py_code.encode(__salt_system_encoding__))
+            env_encoded_err = env_encoded[1]
+            env_encoded = env_encoded[0]
             env_encoded_org = env_encoded
-            env_encoded = env_encoded_org[0]
-            env_encoded_err = env_encoded_org[1]
             env_mark = env_encoded.find(marker + '\n')
             if env_mark < 0:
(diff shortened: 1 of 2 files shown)


psyer:
The integration test that is failing for: unit.modules.test_cmdmod.CMDMODTestCase.test_os_environment_remains_intact

With error:
_CommandExecutionError: Environment could not be retrieved for User 'foobar': missing first marker, got err=None out=['{}', None]
	
Standard Error
/bin/sh: 1: foo: not found_

This error looks like it might have been happening before (I can't see the history from other builds) my code was added. It just might not have been caught,  it looks like the word 'foo' is trying to be executed by the shell during the test suite, which failed. Stderr was never exposed before, nor was checking to see if an error was happening during python code exec. Looking for the new unique tags wrapping the env code, and not finding them tripped the exception.

 When the code is figuring out how to set the cmd_env var,  sys.executable looks like it is getting set to 'foo' and then trying to be executed by the test system.  Not sure what is going on here, but that is just a guess.

terminalmage:
@psyer Please take a look at https://github.com/psyer/salt/pull/1

terminalmage (changes requested):
I've opened a PR against the OP's fork to simplify some of the logic. I'm working with the OP now to satisfy their concerns.

terminalmage:
@psyer No need to merge the 2017.7 branch into this branch again, let's let tests finish and we should be able to merge it without syncing again with 2017.7.

terminalmage (approved):

rallytime:
@terminalmage and @psyer this test failure looks related: https://jenkins.saltstack.com/job/PR/job/salt-pr-linode-cent7-py3/3322/testReport/

psyer:
@terminalmage I'm going to let you decide how you want to handle this. I believe the previous code dealt with something like this bytes to str error if you need an example https://github.com/saltstack/salt/pull/46503/commits/e8d3d017f9dade103c48050526f91be0ec1d7f87. If you want me to take a crack at it let me know.

terminalmage:
That's weird, GitHub appears to have pulled my latest commit from my fork into this PR, even though I didn't open this PR. I just added e8582e8 to fix the test.

terminalmage:
I had meant to open another PR against @psyer's fork to give them a chance to review. This seems like a potential security issue with GitHub.

terminalmage:
I've filed a support request upstream with GitHub. It's possible that because I have admin access to this repo, and I had already contributed with @psyer, there is a "feature" which allows me to directly push to the PR.

rallytime:
@psyer and @terminalmage Is this ready to go, then?

psyer:
@rallytime  I'm good with the current code. Let me know if you need anything else from me.

terminalmage:
This should be good to merge. I found that the same feature we enable to allow us to sync a PR with the base branch also allows maintainer's to push to the head branch for a given PR. So, somehow my push to my local branch also ended up in @psyer's, or else I mistyped when I pushed. Either way, that explains that.

rallytime:
Thanks y'all - merged. :)

PR #46522, opened 2018-03-13
Handle some of the state deprecations for Fluorine

### LINKED ISSUE #46470: Handle deprecated code in docker files for Fluorine
The following files have deprecation notices for functions set for removal in Fluorine:

- [x] salt/states/docker_network.py
- [x] salt/states/docker.py

### LINKED ISSUE #46471: Handle deprecated code in git files for Fluorine
The following files have deprecation notices for functions set for removal in Fluorine:

- [x] salt/states/git.py

### LINKED ISSUE #46472: Handle deprecated code in mongodb files for Fluorine
The following files have deprecation notices for functions set for removal in Fluorine:

- [ ] salt/states/mongodb_user.py
- [ ] salt/states/mongodb_database.py

rallytime:
There are several deprecations marked for removal in the Fluorine release. This PR handles several of them. All changes have been documented in the Fluorine release notes. 

Fixes #46470
Fixes #46471
Fixes #46472


doc/topics/releases/fluorine.rst
```diff
@@ -50,3 +84,17 @@ instead:
...
+- The ``mongodb_database.present`` function was removed. Please update applicable
+  SLS files to use the ``mongodb.database_absent`` function instead.
+
+The ``mongodb_user`` state has been removed. The following functions should be
+used instead:
+
+- The ``mongodb_user.present`` function was removed. Please update applicable SLS
+  files to use the ``mongodb.user_present`` function instead.
+- The ``mongodb_user.absent`` function was removed. Please update applicable SLS
+  files to use the ``mongodb.user_absent`` function instead.
```
terminalmage: Wait.... I split up the states in the docker state module to make it more like things like mongodb, etc. where we have separate state modules for different types of things, and now these state modules are being merged?

I think it'd look kinda silly for us to have split up the docker states and also merged the mongodb state modules into one module. There should be some consistency, here.

rallytime: @terminalmage That's likely true. But I didn't do the original deprecations in this case. These warnings have been there for a while and the states have already been changed in previous releases.

I'm don't think the mongo* states are as robust as the docker* states. That probably has something to do with the original deprecations.

rallytime: @terminalmage What do you think we should do here?

terminalmage: I don't know, I feel like we should have consistent behavior, so I would prefer not to consolidate the states into a single module. Like I said, that's why I split up the docker states into their own separate modules in 2017.7.0. We do it this way for mysql, and docker where we use separate state modules for distinct _things_ that are managed, and we _were_ doing it this way for mongodb as well. It also looks way cleaner in SLS files, especially in requisites. For example:

```
foo:
  mongodb_database.present:
    - require:
      - mongodb_user: foo
```

With the states consolidated, you end up with:

```
foo:
  mongodb.database_present:
    - require:
      - mongodb: foo
```

See what I mean?

rallytime:
re-run all

terminalmage (approved):

Issue #44870, opened 2017-12-07
Add Tests for #44642


Ch3LL:
### Description of Issue/Question
Add tests around this PR: https://github.com/saltstack/salt/issues/43796

PR #46526, opened 2018-03-13
Add tests for new source_* minion options

Ch3LL:
### What does this PR do?
Adds tests for https://github.com/saltstack/salt/issues/44870

Also fixes an issue found while writing tests. If check_dns is False and source_interface_name is set then this error appears:

```
*** unit.test_minion Tests  *******************************************************************************************************************************************
 --------  Tests with Errors  -----------------------------------------------------------------------------------------------------------------------------------------
   -> unit.test_minion.MinionTestCase.test_source_address  ............................................................................................................
         File "/home/ch3ll/git/salt/tests/unit/test_minion.py", line 98, in test_source_address
           assert salt.minion.resolve_dns(__opts__) == {'source_publish_port': 49018,
         File "/home/ch3ll/git/salt/salt/minion.py", line 224, in resolve_dns
           ret['source_ip'] = salt.utils.network.dns_check(
       UnboundLocalError: local variable 'salt' referenced before assignment
   ....................................................................................................................................................................
   -> unit.test_minion.MinionTestCase.test_source_int_name_local  .....................................................................................................
         File "/home/ch3ll/git/salt/tests/unit/test_minion.py", line 50, in test_source_int_name_local
           assert salt.minion.resolve_dns(__opts__) == {'master_ip': '127.0.0.1',
         File "/home/ch3ll/git/salt/salt/minion.py", line 209, in resolve_dns
           interfaces = salt.utils.network.interfaces()
       UnboundLocalError: local variable 'salt' referenced before assignment
```

https://github.com/saltstack/salt/issues/44870

### Tests written?

Yes

### DIFF
--- a/salt/minion.py
+++ b/salt/minion.py
@@ -144,10 +144,10 @@ def resolve_dns(opts, fallback=True):
     if (opts.get('file_client', 'remote') == 'local' and
             not opts.get('use_master_when_local', False)):
         check_dns = False
+    # Because I import salt.log below I need to re-import salt.utils here
+    import salt.utils
 
     if check_dns is True:
-        # Because I import salt.log below I need to re-import salt.utils here
-        import salt.utils
         try:
             if opts['master'] == '':
                 raise SaltSystemExit
--- a/salt/minion.py
+++ b/salt/minion.py
@@ -144,8 +144,8 @@ def resolve_dns(opts, fallback=True):
     if (opts.get('file_client', 'remote') == 'local' and
             not opts.get('use_master_when_local', False)):
         check_dns = False
-    # Because I import salt.log below I need to re-import salt.utils here
-    import salt.utils
+    # Since salt.log is imported below, salt.utils.network needs to be imported here as well
+    import salt.utils.network
 
     if check_dns is True:
         try:
--- a/tests/unit/transport/test_zeromq.py
+++ b/tests/unit/transport/test_zeromq.py
@@ -317,22 +317,24 @@ class ZMQConfigTest(TestCase):
         s_ip = '111.1.0.1'
         s_port = 4058
 
-        # source ip and source_port empty
-        assert salt.transport.zeromq._get_master_uri(master_ip=m_ip,
-                                                     master_port=m_port) == 'tcp://{0}:{1}'.format(m_ip, m_port)
+        with patch('salt.transport.zeromq.LIBZMQ_VERSION_INFO', (4, 1, 6)), \
+            patch('salt.transport.zeromq.ZMQ_VERSION_INFO', (16, 0, 1)):
+            # pass in both source_ip and source_port
+            assert salt.transport.zeromq._get_master_uri(master_ip=m_ip,
+                                                         master_port=m_port,
+                                                         source_ip=s_ip,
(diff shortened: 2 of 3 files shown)


salt/minion.py
```diff
@@ -144,10 +144,10 @@ def resolve_dns(opts, fallback=True):
     if (opts.get('file_client', 'remote') == 'local' and
             not opts.get('use_master_when_local', False)):
         check_dns = False
+    # Because I import salt.log below I need to re-import salt.utils here
```
rallytime: Perhaps this comment could be worded a little better. Maybe something like: `Since salt.log is imported below, salt.utils.network needs to be imported here as well`?

salt/minion.py
```diff
@@ -144,10 +144,10 @@ def resolve_dns(opts, fallback=True):
...
+    import salt.utils
```
rallytime: I think this should be `salt.utils.network`, and not `salt.utils`. We've moved away from importing `salt.utils` directly in 2018.3 and should be using direct imports to the needed files.

rallytime (changes requested):
I know the lines I am commenting on were not yours originally, but since you're moving them, can you clean those up?

rallytime:
re-run py2

rallytime (approved):
Nice.

tests/unit/transport/test_zeromq.py
```diff
@@ -305,3 +305,34 @@ def test_send(self):
...
+        '''
+        test _get_master_uri method
+        '''
+        m_ip = '127.0.0.1'
+        m_port = 4505
+        s_ip = '111.1.0.1'
+        s_port = 4058
+
+        # source ip and source_port empty
+        assert salt.transport.zeromq._get_master_uri(master_ip=m_ip,
```
DmitryKuzmenko: Why `assert` and not `self.assertEqual` is here? The difference in the output if a failure case will be sensitive, for instance:
```
AssertionError
```
vs
```
AssertionError: u'A brown fox jumped over the red hen.' != u'Abrownfoxjumpedovertheredhen.'
       - A brown fox jumped over the red hen.
       ?  -     -   -      -    -   -   -
       + Abrownfoxjumpedovertheredhen.
```

DmitryKuzmenko (changes requested):
Could you please change all `assert A == B` to `self.assertEqual(A, B)`? Or please correct me if there is a reason to use `assert`.

tests/unit/transport/test_zeromq.py
rallytime: `assert` is the syntax pytest uses and `self.assertEqual` is the syntax `unittest2` uses. We can support both ways. It is true that the unittest2 syntax makes it easier to read the failures, but since we will be moving towards pytest in the future, I am OK with this as-is.

rallytime:
re-run py2

rallytime:
re-run py2

rallytime:
re-run ubuntu

rallytime:
@Ch3LL It looks like one of your new tests is failing in the Ubuntu 14 PY2 run:

- unit.transport.test_zeromq.ZMQConfigTest.test_master_uri

Can you take a look?

cachedout:
I talked to @Ch3LL who will look at this either today or tomorrow.

PR #46527, opened 2018-03-13
Create reg salt util

twangboy:
### What does this PR do?
Use the reg salt util to get grains
Use the reg salt util in the reg module
Use the reg salt util in the reg state
Use the reg salt util to get the PowerShell version
Detect the version of PowerShell instead of the version of Windows to determine ConvertTo-JSON usage (The whole purpose for all of this)

https://github.com/saltstack/salt/pull/46501

### Previous Behavior
Basing `ConvertTo-JSON` on the version of Windows is incorrect. Although correct on a fresh install of Windows 7, PowerShell can be upgraded, at which point `ConvertTo-JSON` would be available.

### Tests written?
No

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/grains/core.py
+++ b/salt/grains/core.py
@@ -76,9 +76,8 @@ if salt.utils.platform.is_windows():
         import wmi  # pylint: disable=import-error
         import salt.utils.winapi
         import win32api
-        import salt.modules.reg
+        import salt.utils.win_reg
         HAS_WMI = True
-        __salt__['reg.read_value'] = salt.modules.reg.read_value
     except ImportError:
         log.exception(
             'Unable to import Python wmi module, some core grains '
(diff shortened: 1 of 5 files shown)


rallytime:
Wait, why is a change this large going in to the RC branch? Can you provide a little more explanation here, please?

twangboy:
This is the fix the `ConvertTo-JSON` issue that @frogunder was seeing on Windows 7 and Server 2008 R2.

One of the grains uses PowerShell to populate the grain. It uses `salt.modules.cmdmod.powershell` to do so, which needs to detect if the `ConvertTo-JSON` command is available, otherwise, it will stack trace. This needs to be done by checking values in the registry instead of launching PowerShell. Since this happens when loading grains, the reg module isn't available, so it needs to be moved into a salt.util.

rallytime:
I am extremely hesitant to make this large of a change in the RC 1 branch when we are trying to release soon.

This doesn't have any tests either.

The issue in question looks to be #46425, which was not deemed a blocker. Just something to fix and only occurs on the `sys.doc` call.  Can this wait for the `.1` release?

twangboy:
Submitted against 2018.3 branch. Will be in 2018.3.1

rallytime:
Ok, thanks @twangboy. I'll close this one then.

damon-atkins:
Hi @twangboy did not notice this PR when it was raised. Sort of results in the doco being maintain in two places and the lost of years of history for modules/reg.py when people look at utils/win_reg.py. Might be worth putting a comment in utils/win_reg.py that it was originally modules/reg.py

What was the reason for pulling the code out if modules/reg and placing it in different file and calling the same code?

```
def list_keys(hive, key=None, use_32bit_registry=False):
    '''
doco cut
    '''
    return __utils__['reg.list_keys'](hive=hive, key=key, use_32bit_registry=use_32bit_registry)
```

twangboy:
@damon-atkins Some of the grains needed access to the registry and those happen before the loader populates `__salt__`. So they were moved into this salt util to standardize the way we access the registry.

damon-atkins:
Thanks for the info.

Issue #44886, opened 2017-12-08
Azure managed storage


sumeetisp:
### Description of Issue/Question

How to provision azure vm's with managed disks/storage?

Ch3LL:
Are you using azure or azure arm salt-cloud driver?

sumeetisp:
Yes

Thanks,
Sumeet Salvankar

sumeetisp:
Sorry... I am using azure arm cloud driver

Thanks,
Sumeet Salvankar

On 08-Dec-2017 23:47, "sumeet salvankar" <sumeetisp@gmail.com> wrote:

Yes

Thanks,
Sumeet Salvankar

Ch3LL:
Thanks for the information. parsing the azure arm docs i'm only seeing the ability to clean up the disks?

ping @saltstack/team-cloud do you know of a way to manage the disks with azure arm specifically?

gtmanfred:
There is not a way to directly manage the volumes, only in reference to building the vm and volume together.

sumeetisp:
I am not talking abt managing volumes.

My question was about managed and unmanaged volumes provided by azure.

When we specify storage account in azure salt profile the VM that is
provisioned gets os volume from unmanaged storage.

And when we provision from azure console without specifying any storage
account gets azure managed disk.

I tried excluding storage account from profile but storage account is a
mandatory parameter.

So my question is how can I provision VM with managed disk's.
Thanks,
Sumeet Salvankar

gtmanfred:
@techhat can you comment on this since you are more familiar with the azurearm driver?

Thanks,
Daniel

SJern:
Hi all,

The Azure documentation says, "We recommend that you use Azure Managed Disks for new VMs, and that you convert your previous unmanaged disks to managed disks, to take advantage of the many features available in Managed Disks."
https://docs.microsoft.com/en-us/azure/virtual-machines/linux/about-disks-and-vhds

It would be great if we can figure out how to provision managed disks in salt. Thanks!

techhat:
Looks like in `volumes` you can set `create_option` to `image`, `attach`, or `empty`. But not all of the required options for managed disks are available in the drive right now. For future reference: https://docs.microsoft.com/en-us/python/azure/python-sdk-azure-samples-managed-disks?view=azure-python

sumeetisp:
According to docs point to by @techhat ,
[...]

PR #46534, opened 2018-03-14
Changes to determine os_type

sumeetisp:
What does this PR do?

Fixes the way to determine os type. It introduces a new profile parameter 'os_type' which can have value as Windows or Linux

#44886

Previous Behavior

os_type was determined from presence of win_installer. But when deploy is set to false there is not need to mention win_installer in salt profile. Because of this for custom images irrespective of type of os, os_type was being set to linux and hence provisioning used to fail (when deploy is set to false).
Market place images worked fine

New Behavior

os_type i made mandatory to be specified in case of custom images or vhd.

Tests written?

Yes/No

Commits signed with GPG?

Yes/No

Please review Salt's Contributing Guide for best practices.

See GitHub's page on GPG signing for more information about signing commits with GPG.


salt/cloud/clouds/azurearm.py
```diff
@@ -1173,10 +1180,6 @@ def request_instance(vm_):
                 sku=img_sku,
                 version=img_ver,
             )
-        if win_installer:
-            os_type = 'Windows'
-        else:
-            os_type = 'Linux'
```
gtmanfred: I am ok with the other change, but if os_type is not specified, I would like to leave this as the default, and log a debug message instead of raising a system exit error.

gtmanfred (changes requested):

sumeetisp:
What happens when os_type is not specified, deploy is set to false and
win_installer is not specified?

Thanks,
Sumeet Salvankar

sumeetisp:
In this case for custom image, Linux will be set by default. This causes
failure if custom image is of windows

Thanks,
Sumeet Salvankar

On 14-Mar-2018 20:00, "sumeet salvankar" <sumeetisp@gmail.com> wrote:

What happens when os_type is not specified, deploy is set to false and
win_installer is not specified?

Thanks,
Sumeet Salvankar

gtmanfred:
That is fine, that is what the log message is for.

The problem is you are changing the default behavior without a need for it, adding a log message if os_type is not specified would be fine, but exiting i think is over doing it.

sumeetisp:
Ok, I will make the changes.

Thanks,
Sumeet Salvankar

On 14-Mar-2018 20:31, "Daniel Wallace" <notifications@github.com> wrote:

That is fine, that is what the log message is for.

The problem is you are changing the default behavior without a need for it,
adding a log message if os_type is not specified would be fine, but exiting
i think is over doing it.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/saltstack/salt/pull/46534#issuecomment-373052725>, or mute
the thread

salt/cloud/clouds/azurearm.py
```diff
@@ -1164,6 +1167,12 @@ def request_instance(vm_):
         if vm_['image'].startswith('http'):
             source_image = VirtualHardDisk(vm_['image'])
             img_ref = None
+            if not os_type:
+                log.debug("os_type not specified")
+                if win_installer:
+                    os_type = 'Windows'
+                else:
+                    os_type = 'Linux'
```
gtmanfred: This is going to cause an issue, can we put this statement back where it was before?

gtmanfred (changes requested):

PR #46544, opened 2018-03-14
Allow running the schedule jobs to run in the foreground.

s0undt3ch:
### What does this PR do?
Allows the scheduler to run jobs in the foreground.

### Previous Behavior
Schedule jobs always ran in the background either on a new process or a thread.

### Tests written?

No

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/utils/schedule.py
+++ b/salt/utils/schedule.py
@@ -426,23 +426,27 @@ class Schedule(object):
 
         # Grab run, assume True
         run = data.get('run', True)
+        run_schedule_jobs_in_background = self.opts.get('run_schedule_jobs_in_background', True)
         if run:
-            multiprocessing_enabled = self.opts.get('multiprocessing', True)
-            if multiprocessing_enabled:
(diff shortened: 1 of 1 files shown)


salt/utils/schedule.py
```diff
@@ -408,23 +408,27 @@ def run_job(self, name):
 
         # Grab run, assume True
         run = data.get('run', True)
+        run_jobs_in_background = self.opts.get('run_jobs_in_background', True)
```
s0undt3ch: should it be `run_schedule_jobs_in_background`? On the salt side, `opts` is the **whole** set of options....

garethgreenaway: Should probably be `run_schedule_jobs_in_background`.

s0undt3ch: Updated

garethgreenaway (approved):

rallytime:
re-run py3

PR #46554, opened 2018-03-15
Regular expression matching possibility for state.event()

ghost:
A new parameter, tagrematch, for state.event() (in both modules and runners). This allows the caller of the function use regular expression matching against event tags in addition to previously available fnmatch matching.
    
The old behaviour of the function is retained, with no visible changes to code/users who are fine with fnmatch-style matching.

### What does this PR do?

Adds a keyword parameter, tagrematch, to state.event() in modules and runners. The parameter allows tags to be matched with regular expressions in addition to previously available fnmatch style matching.

The PR does not change the function interface towards previous code. All callers using it as they used to will work as expected.

Sometimes it gets rather tedious to debug event-related problems (ie. broken orchestrator states and so on) on a master which is really busy (200+ events per second) without a decent matching capability to filter events. Using regular expressions to filter events displayed makes the life of an administrator a tad easier.

### Tests written?

No

### Commits signed with GPG?

No

### DIFF
--- a/salt/modules/state.py
+++ b/salt/modules/state.py
@@ -16,6 +16,7 @@ from __future__ import absolute_import, print_function, unicode_literals
 import fnmatch
 import logging
 import os
+import re
 import shutil
 import sys
 import tarfile
@@ -2290,6 +2291,7 @@ def _disabled(funs):
 
 
 def event(tagmatch='*',
+        tagrematch=None,
         count=-1,
         quiet=False,
         sock_dir=None,
@@ -2307,8 +2309,12 @@ def event(tagmatch='*',
 
     :param tagmatch: the event is written to stdout for each tag that matches
         this pattern; uses the same matching semantics as Salt's Reactor.
+    :param tagrematch: the event is written to stdout for each tag that matches
+        this regular expression; if both ``tagmatch`` and ``tagrematch``
+        are defined, ``tagrematch`` takes precedence.
     :param count: this number is decremented for each event that matches the
-        ``tagmatch`` parameter; pass ``-1`` to listen forever.
+        ``tagmatch`` or ``tagrematch`` parameter; pass ``-1``, which is also
+        the default value, to listen forever.
     :param quiet: do not print to stdout; just block
     :param sock_dir: path to the Salt master's event socket file.
     :param pretty: Output the JSON all on a single line if ``False`` (useful
@@ -2328,12 +2334,17 @@ def event(tagmatch='*',
             opts=__opts__,
             listen=True)
 
+    if tagrematch is not None:
+      tagmatch_fun = lambda tag, pat = re.compile(tagrematch): pat.search(tag)
+    else:
+      tagmatch_fun = lambda tag: fnmatch.fnmatch(tag, tagmatch)
+
     while True:
         ret = sevent.get_event(full=True, auto_reconnect=True)
         if ret is None:
             continue
 
-        if fnmatch.fnmatch(ret['tag'], tagmatch):
+        if tagmatch_fun(ret['tag']):
             if not quiet:
                 salt.utils.stringutils.print_cli(
                     str('{0}\t{1}').format(  # future lint: blacklisted-function
--- a/salt/runners/state.py
+++ b/salt/runners/state.py
@@ -225,6 +225,7 @@ orch_show_sls = salt.utils.functools.alias_function(orchestrate_show_sls, 'orch_
 
 
 def event(tagmatch='*',
+        tagrematch=None,
         count=-1,
         quiet=False,
         sock_dir=None,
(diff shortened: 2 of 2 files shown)


rallytime:
Lint: https://jenkins.saltstack.com/job/PR/job/salt-pr-lint-n/20166/violations/file/salt/modules/state.py/

isbm (changes requested):
Hmm. As you say yourself: "without a decent matching capability to filter events". So if this is no decent matching, why bother? Let's remove that.

I see your point, but I am personally not in favour adding another parameter. First, `tagrematch` for unknown reasons overrides `tagmatch`, even if `tagmatch` is not `None`. Second, this is the same thing, just from the different perspective. Backward compatibility probably counts here, though.

Options are:
- Move everything to regexps (my preference)
- Write some sort of `tagrematch` and `tagmatch` recognising function that would use `fnmatch` or `re` automatically, if one still needs `fnmatch`. But this might be buggy and why we would use both engines to achieve the same thing?

So I would rather move to regexps, but moving everyone slowly away. This can be done by by writing two another functions and renaming-and-hiding the very original (keeping your changes for now):

```python
def  event(tagmatch='*', count=-1, quiet=False, sock_dir=None):
    'This uses regular expression'
    return _actual_event(tag_re_match=tagmatch, .....

def _event(tagmatch='*', count=-1, quiet=False, sock_dir=None):
    'This uses a deprecated version with fnmatch'
    return _actual_event(tagmatch=tagmatch....

def _actual_event(tagmatch='*', tag_re_match=None, count=-1, quiet=False, sock_dir=None):
    'This is an actual function'
```

To use old behaviour one would need to add into config:
```
use_deprecated:
    - state.event
```
But user will see a warning in the log "Please migrate me!". Then after Fluorine this will start banning `_event` version entirely and so we can just rename `_actual_event` into `event` and throw away `tagmatch` entirely along with the `fnmatch` engine.

You can also set default old one by so:

```python
@with_deprecated(globals(), "Fluorine", policy=with_deprecated.OPT_IN)
```

Then it will start "crying" into the logs "migrate me!" right away. To turn that off and start using regexps, one would need to configure so:
```
use_superseded:
  - state.event
```

And I would add Unit Tests. :wink:

samilaine:
Fair enough, backwards compatibility in this case isn't much of an issue anyway.

One question, though. It seems that @with_deprecated breaks salt-call. I'd presume this isn't what should be happening.

For example, tossing the example below into `salt/modules/state.py` works nicely (ie. `salt-call state.fun` returns `True`):

```
def fun():
    return True

def _fun():
    return False
```

Adding an import for the decorator and putting the decorator in produces something rather unexpected ...

```
from salt.utils.decorators import with_deprecated

def fun():
    return True

def _fun():
    return False
```

... namely what's below.

```
[DEBUG   ] LazyLoaded state.fun

Passed invalid arguments: coercing to Unicode: need string or buffer, function found.

Usage:
None
  File "/Volumes/User/sjl/work/salt/salt/salt/cli/caller.py", line 212, in call
  File "/Volumes/User/sjl/work/salt/salt/salt/utils/decorators/__init__.py", line 558, in _decorate
  File "/Volumes/User/sjl/work/salt/salt/salt/utils/decorators/__init__.py", line 252, in _call_function
    error = six.text_type(error).replace(self._function, self._orig_f_name)  # Hide hidden functions
TypeError: coercing to Unicode: need string or buffer, function found
```

```
Salt Version:
           Salt: 2018.3.0-225-gc88c20e
 
Dependency Versions:
           cffi: Not Installed
       cherrypy: Not Installed
       dateutil: Not Installed
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.10
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.5.6
   mysql-python: Not Installed
      pycparser: Not Installed
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: Not Installed
         Python: 2.7.14 (default, Nov 20 2017, 12:01:37)
   python-gnupg: Not Installed
         PyYAML: 3.12
          PyZMQ: 17.0.0
           RAET: Not Installed
          smmap: Not Installed
        timelib: Not Installed
        Tornado: 4.5.3
            ZMQ: 4.1.6

System Versions:
           dist:   
         locale: UTF-8
        machine: x86_64
        release: 17.4.0
         system: Darwin
        version: 10.13.3 x86_64
```

Comments `salt/utils/decorators/__init__.py` in hint that I might be using the decorator correctly, but I wouldn't bet on that any more than I'd bet on Betsy the Horse.

I should probably be asking this on IRC, but I guess this might come handy for someone who'll be landing on this page with a search engine in 2021.

samilaine:
For search engines: See #46811.

PR #46559, opened 2018-03-15
Add netbox pillar

kris-steinhoff:
### What does this PR do?

Add a pillar module that adds data to the Pillar structure from a [NetBox](http://netbox.readthedocs.io/en/latest/) API. Optionally creates the `proxy` structure.

### Tests written?

No

### Commits signed with GPG?

No

### DIFF
--- a/salt/pillar/netbox.py
+++ b/salt/pillar/netbox.py
@@ -1,0 +1,129 @@
+# -*- coding: utf-8 -*-
+'''
+A module that adds data to the Pillar structure from a NetBox API.
+
+
+Configuring the NetBox ext_pillar
+====================================
+
+.. code-block:: yaml
+
+  ext_pillar:
+    - netbox:
+        api_url: http://netbox_url.com/api/
+
+The following are optional, and determine whether or not the module will
+attempt to configure the ``proxy`` pillar data for use with the napalm
+proxy-minion:
+
+.. code-block:: yaml
+
+        proxy_return: True
+        proxy_username: admin
+        api_token: 123abc
+
+Create a token in your NetBox instance at
+http://netbox_url.com/user/api-tokens/
+
+By default, this module will query the NetBox API for the platform associated
+with the device, and use the 'NAPALM driver' field to set the napalm
+proxy-minion driver. (Currently only 'napalm' is supported for drivertype.)
+
+This module assumes you will use SSH keys to authenticate to the network device
+If password authentication is desired, it is recommended to create another
+``proxy`` key in pillar_roots (or git_pillar) with just the ``passwd`` key and
+use :py:func:`salt.renderers.gpg <salt.renderers.gpg>` to encrypt the value.
+If any additional options for the proxy setup are needed they should also be
+configured in pillar_roots.
+'''
+
+
+from __future__ import absolute_import, print_function, unicode_literals
+import logging
+
+try:
+    import requests
+    import ipaddress
+    _HAS_DEPENDENCIES = True
+except ImportError:
+    _HAS_DEPENDENCIES = False
+
+log = logging.getLogger(__name__)
+
+
+def __virtual__():
+    return _HAS_DEPENDENCIES
+
+
+def ext_pillar(minion_id, pillar, *args, **kwargs):
+    '''
+    Query NetBox API for minion data
+    '''
+
+    # Pull settings from kwargs
+    api_url = kwargs['api_url'].rstrip('/')
+
+    api_token = kwargs.get('api_token', None)
+    proxy_username = kwargs.get('proxy_username', None)
+    proxy_return = kwargs.get('proxy_return', True)
+
+    ret = {}
+
+    headers = {}
+    if api_token:
+        headers['Authorization'] = 'Token ' + api_token
+
+    # Fetch device from API
+    device_results = requests.get(
+        api_url + '/dcim/devices/',
+        params={'name': minion_id, },
+        headers=headers,
+    )
+
+    # Check status code for API call
+    if device_results.status_code != requests.codes.ok:
+        log.warn('API query failed for "%s", status code: %d',
+                 minion_id, device_results.status_code)
+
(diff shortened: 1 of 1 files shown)


salt/pillar/netbox.py
```diff
@@ -0,0 +1,121 @@
...
+
+This module currently only supports the napalm proxy minion and assumes
+you will use SSH keys to authenticate to the network device.  If password
+authentication is desired, it is recommended to create another ``proxy``
+key in pillar_roots (or git_pillar) with just the ``passwd`` key and use
+:py:func:`salt.renderers.gpg <salt.renderers.gpg>` to encrypt the value.
+If any additional options for the proxy setup are needed they should also be
+configured in pillar_roots.
+'''
+
```
terminalmage: Please add the following here:

```python
```

terminalmage (changes requested):
We need some imports from `__future__` here. Also, please update the logging here so that you pass the args to the log handler instead of performing string replacement manually? Our logging handler normalizes str/unicode mismatches on Python 2, so you should just do your logging like this:

```python
log.debug('foo = %s', foo)
```

Doing the logging the way you have it now will result in the string replacement happening before our logging handler has a chance to normalize any str/unicode mismatches.

ksofa2:
@terminalmage Thanks. We've made the requested changes. Also squashed the commits to clean up a mess I made in merging. (Sorry about that.)

salt/pillar/netbox.py
```diff
@@ -0,0 +1,122 @@
...
+        log.warn('API query failed for "%s", status code: %d',
+            minion_id, device_results.status_code)
+
+    # Assign results from API call to "netbox" key
+    try:
+        devices = device_results.json()['results']
+        if len(devices) == 1:
+            ret['netbox'] = devices[0]
+        elif len(devices) > 1:
+            log.error('More than one device found for "%s"' % minion_id)
```
terminalmage: This log entry still needs to be changed per my earlier comments.

terminalmage (changes requested):

ksofa2:
Fixed.

salt/pillar/netbox.py
```diff
@@ -0,0 +1,122 @@
...
+            if platform_results.status_code != requests.codes.ok:
+                log.info('API query failed for "%s", status code: %d',
+                         minion_id, platform_results.status_code)
+
+            # Assign results from API call to "proxy" key
+            ret['proxy'] = {
+                'host': str(ipaddress.IPv4Interface(
+                            ret['netbox']['primary_ip4']['address']).ip),
+                'driver': platform_results.json()['napalm_driver'],
+                'proxytype': 'napalm',
```
mirceaulinic: This should not be hardcoded like that. Maybe at least have it as an option or something.

salt/pillar/netbox.py
```diff
@@ -0,0 +1,122 @@
...
+A module that adds data to the Pillar structure from a NetBox API.
+
+
+Configuring the NetBox ext_pillar
+====================================
+
+.. code-block:: yaml
+
+  ext_pillar:
+    - netbox:
```
mirceaulinic: I would argue that this Pillar module should not be named as `netbox`, as its scope is very, very specific to retrieving data from NetBox and populating the `proxy` key into the Pillar. While this is very useful and nice, maybe we should emphasise the scope and rename it, say `netbox_proxy` or something similar.

To clarify: a `netbox` external Pillar I would expect to provide me _all_ the NetBox data related to that device (i.e, from the  `dcim/devices/<device ID>/` endpoint).

mirceaulinic (changes requested):
Nice work @ksofa2. I think we should rename this module appropriate to its scope, and document what it does exactly (the docstring doesn't currently reflect that).

salt/pillar/netbox.py
ksofa2: It's creating the `proxy` key if requested, but this isn't just doing that. It also is returning the device info in the `netbox` key. It's a raw dump of what it gets back from the API call.

mirceaulinic: Ah, okay, I misread that - please disregard.

salt/pillar/netbox.py
ksofa2: Good point, but given the structures we get back from NetBox, I'd like to leave it hardcoded and update the documentation so say that only napalm proxtypes are supported.

I added a code change to only produce the `proxy` key if the napalm_driver is defined in the device's platform entry.

salt/pillar/netbox.py
```diff
@@ -0,0 +1,126 @@
...
+
+
+def ext_pillar(minion_id, pillar, *args, **kwargs):
+    '''
+    Query NetBox API for minion data
+    '''
+
+    # Pull settings from kwargs
+    api_url = kwargs['api_url'].rstrip('/')
+    api_token = kwargs['api_token']
```
mirceaulinic: I believe the token should not be mandatory, as this Pillar module only fetches data (the token is required only when writing data into NetBox) - see https://docs.saltstack.com/en/develop/ref/modules/all/salt.modules.netbox.html as an example.

mirceaulinic (changes requested):

mirceaulinic (approved):
I wouldn't have any further comments on this, just to ask you to rebase and squash (but make sure to leave aside commits such as https://github.com/saltstack/salt/pull/46559/commits/ed6cdb775904e1efcfbdbbe95428e18be191f58f or https://github.com/saltstack/salt/pull/46559/commits/2c6d44ad51b777ab50765f4c1cd6aa04776752a2 which are unrelated to this addition). So I will defer to @terminalmage for more input.

Thanks for your contribution @ksofa2.

terminalmage (approved):
:+1:

ksofa2:
Thanks @mirceaulinic and @terminalmage!

I've squashed and rebased onto develop. Please let me know if there's anything else I need to do.

PR #46563, opened 2018-03-15
virtualenv version too old for python3.6

gtmanfred:
### What does this PR do?
This test is testing an old version of virtualenv with pip_env, and the version
of pip is 6.x.x, which is too old and doesn't work with the importlib in python
3.6. Skipping this test if python is newer then 3.5.

Fixes saltstack/salt-jenkins#886

### Tests written?

Yes

### Commits signed with GPG?

Yes

### DIFF
--- a/tests/integration/states/test_pip.py
+++ b/tests/integration/states/test_pip.py
@@ -14,6 +14,7 @@ import os
 import pwd
 import glob
 import shutil
+import sys
 
 # Import Salt Testing libs
 from tests.support.mixins import SaltReturnAssertsMixin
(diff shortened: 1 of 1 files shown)


.kitchen.yml
```diff
@@ -217,6 +217,8 @@ verifier:
   transport: zeromq
   types:
     - ssh
+  tests:
+    - integration.states.test_pip.PipStateTest.test_46127_pip_env_vars
```
rallytime: This shouldn't be in here. ;)

rallytime (changes requested):
@gtmanfred One change needed here.

gtmanfred:
Fixed :)

rallytime (approved):

PR #46566, opened 2018-03-15
Break out of ~infinite loop in boto_asg.get_instances

yred:
### What does this PR do?

Fix a minor ~infinite loop bug in `boto_asg.get_instances`.

Described above.

### Previous Behavior

`boto_asg.get_instances` could loop ~infinitely. In practice, it will be throttled by AWS, and will switch back and forth between both states until it fails.
 
### New Behavior

Break out of the retry loop as soon as AWS returns a successful response.

### Tests written?

No.

### Commits signed with GPG?

No.

### DIFF
--- a/salt/modules/boto_asg.py
+++ b/salt/modules/boto_asg.py
@@ -829,6 +829,7 @@ def get_instances(name, lifecycle_state="InService", health_status="Healthy",
     while True:
         try:
             asgs = conn.get_all_groups(names=[name])
+            break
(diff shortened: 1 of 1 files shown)


rallytime (approved):

ryan-lane:
+1 🤦‍♂️

rallytime:
Thanks @yred! And congrats on your first Salt PR :)

Issue #27898, opened 2015-10-13
[salt-ssh] Salt-ssh Crashes Silently When Executed More Than Twice


srkunze:
In our setup, we currently run more than two salt-ssh simultaneously:

```
salt:/srv # salt-ssh minion state.sls host& 
[1] 13915
salt:/srv # salt-ssh minion state.sls host& 
[2] 13927
salt:/srv # salt-ssh minion state.sls host& 
[3] 13944
salt:/srv # salt-ssh minion state.sls host& 
[4] 13951
salt:/srv # salt-ssh minion state.sls host& 
[5] 13968

salt:/srv # minion:

Summary for minion
-----------
Succeeded: 0
Failed:   0
-----------
Total states run:    0
Total run time:  0.000 ms

minion:

Summary for minion
-----------
Succeeded: 0
Failed:   0
-----------
Total states run:    0
Total run time:  0.000 ms

minion:

Summary for minion
-----------
Succeeded: 0
Failed:   0
-----------
Total states run:    0
Total run time:  0.000 ms

minion:
----------
          ID: necessary_packages
    Function: pkg.installed
      Result: True
        [...]

minion:
----------
          ID: necessary_packages
    Function: pkg.installed
      Result: True
        [...]
```

Note that two salt-ssh calls in parallel work perfectly. More than 3 calls "fails" all of them except the last 2.

jfindlay:
@srkunze, thanks for the report.

srkunze:
Is there something I can do here?

srkunze:
Still an issue with 2016.3.1:

```
salt:/srv # salt-ssh --versions-report
Salt Version:
           Salt: 2016.3.1
 
Dependency Versions:
           cffi: 0.8.6
       cherrypy: Not Installed
       dateutil: 2.1
          gitdb: 0.5.4
      gitpython: 0.3.2 RC1
          ioflo: Not Installed
         Jinja2: 2.8
        libgit2: 0.21.0
        libnacl: Not Installed
       M2Crypto: 0.22
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.4.7
   mysql-python: Not Installed
      pycparser: 2.10
       pycrypto: 2.6.1
         pygit2: 0.21.2
         Python: 2.7.3 (default, Apr 14 2012, 08:58:41) [GCC]
   python-gnupg: Not Installed
         PyYAML: 3.11
          PyZMQ: 15.2.0
           RAET: Not Installed
[...]

PR #46570, opened 2018-03-15
salt-ssh: randomize name of state archive

mephi42:
### What does this PR do?
Fixes the issue when using more than one instance of salt-ssh at a
time resulted in 0 states being applied.

#27898

### Previous Behavior
When two salt-ssh instances run concurrently, one of them ends with 0 states ran.

### New Behavior
Both instances complete successfully.

### Tests written?
No

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/modules/state.py
+++ b/salt/modules/state.py
@@ -2057,21 +2057,13 @@ def clear_cache():
     return ret
 
 
-def pkg(pkg_path,
-        pkg_sum,
-        hash_type,
-        test=None,
-        **kwargs):
+def _pkg(pkg_path,
+         pkg_sum,
+         hash_type,
+         test=None,
+         **kwargs):
     '''
-    Execute a packaged state run, the packaged state run will exist in a
-    tarball available locally. This packaged state
-    can be generated using salt-ssh.
-
-    CLI Example:
-
-    .. code-block:: bash
-
-        salt '*' state.pkg /tmp/salt_state.tgz 760a9353810e36f6d81416366fc426dc md5
+    Main logic of pkg() without package file removal functionality.
     '''
     # TODO - Add ability to download from salt master or other source
     popts = salt.utils.state.get_sls_opts(__opts__, **kwargs)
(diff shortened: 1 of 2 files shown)


cachedout:
I think this is fine but would also like @thatch45 to take a look.

thatch45 (changes requested):
If I am not mistaken the cleanup routine could create another collision because it cleans up all state pkgs and not just the one it needs.
I would argue that we likely need to make the uuid up in the pkg_ssh function and send it down to pkg so that we have a consistent track of what the package name is that we are cleaning up. It might make sense to make the whole path higher up and pass it down so that we have less copy pasta here.

mephi42:
@thatch45 Thanks for looking at this!

My reasoning was that we must make up the name on salt-ssh master, because otherwise collisions may occur on upload. Can this be avoided?

`_cleanup_ssh_pkgs` can indeed lead to problems when >10 concurrent salt-ssh instances are running. While unlikely in practice, I would prefer the design that would avoid this altogether. Since the ability to look at N state archives looks valuable, what do you think about the following scheme?

* Upload archive as `salt_state{1}.tgz`
* Pass it to `pkg_ssh` and `pkg` under this name
* When `pkg` exits, rename to `salt_state_debug{1}.tgz`
* Run cleanup on `salt_state_debug*.tgz` to keep the last N ones

I also agree that getting rid of copy-paste in `salt/client/ssh/wrapper/state.py` would be nice. I can do this once we decide how to proceed here.

thatch45:
I think your solution sounds great because it allows existing apis to remain consistent without mandating that higher level interfaces implement the uuid creation. It also cleanly moves those into a location that can be cleaned effectively.

mephi42:
@cachedout, yep, I pushed a new patch a few days ago and currently waiting for @thatch45 's comments.

rallytime:
@thatch45 When you get a chance, can you take a look at the changes here?

thatch45 (approved):

thatch45:
I am really sorry I missed this! Looks great to me!

salt/modules/state.py
```diff
@@ -2133,6 +2133,34 @@ def pkg(pkg_path,
...
+        if not (s.startswith('debug_salt_state') and s.endswith('.tgz')):
+            continue
+        pkg_path = os.path.join(pkg_dir, s)
+        pkgs.append((pkg_path, os.stat(pkg_path).st_mtime))
+    n_to_keep = 10
+    for pkg_path, _ in sorted(pkgs, key=lambda pkg: pkg[1])[:-n_to_keep]:
+        os.unlink(pkg_path)
+
+
+def pkg_ssh(pkg_path, *args, **kwargs):
```
rallytime: Can you add some documentation to this new function?

mephi42: Ok, will do.

mephi42: I realized that I don't, strictly speaking, need a new function here.

So I introduced a new parameter remove_pkg for the existing pkg(), which is False by default. salt-ssh sets it to True, and this triggers the new behavior.

gtmanfred (approved):

gtmanfred:
This has been confirmed to not break the `docker.call` stuff that uses this same wrapper state stuff right?

re-run py

gtmanfred:
re-run py

salt/modules/state.py
```diff
@@ -2133,6 +2125,46 @@ def pkg(pkg_path,
...
+        salt '*' state.pkg /tmp/salt_state.tgz 760a9353810e36f6d81416366fc426dc md5
+    '''
+    try:
+        return _pkg(pkg_path, *args, **kwargs)
+    finally:
+        if remove_pkg:
+            # Keep the copy for debugging purposes.
+            pkg_dir = os.path.dirname(pkg_path)
+            debug_pkg_path = os.path.join(
+                pkg_dir, 'debug_%s' % os.path.basename(pkg_path))
```
gtmanfred: please use format strings here.

https://jenkins.saltstack.com/job/PR/job/salt-pr-lint-n/22650/violations/file/salt/modules/state.py/

Or just `'debug_' + os.path.basename(pkg_path)` which is faster.

gtmanfred (changes requested):
One lint failure, and wanting to make sure the docker tests that I added that use some of this stuff for docker.highstate and docker.sls don't fail.

rallytime:
Hi @mephi42 - We haven't heard from you in a while. Are you able to make the changes as discussed above?

mephi42:
Hi @rallytime, sorry, I'm a tad busy with other things at the moment, but this is still on my list of TODOs.

rallytime:
No problem - I am going to close this for now. But please let us know when you're ready to come back to this and we will gladly re-open! Or you can open a new PR and link back to this one. Whichever is easier for you.

Do note however that there is a merge-conflict in one file presently. That will need to be fixed as well when you're ready to come back here.

The conflict is in `salt/client/ssh/wrapper/state.py`. Thanks!

Issue #46552, opened 2018-03-15
State with require requisite executes despite onfail requisite on another state.  


JeffLee123:
### Description of Issue/Question
A state with onfail requisite results in True if onfail target is successful and does not need to run. 
This would cause another state with a require requisite on the first state to always run regardless if the onfail target succeeds or not. 

Here, if pip is already installed, I do not want to run ensure_dir, source_download, or install_from_source.
However, if pip_installed is successful, ensure_dir would also result in True because it does not have to run because onfail req did not change. 
source_download has a require requisite on ensure_dir, so  it would run despite having a onfail requisite on pip_installed and pip_install is successful.   

Is there a way to have ensure_dir, source_download, and install_from_source to have depencies on each other, while at the same time, only execute if pip_installed fails? 

### Setup
(Please provide relevant configs and/or SLS files (Be sure to remove sensitive info).)
```
pip_installed:
  pip.installed:
    - name: pip

ensure_dir:
  file.directory:
    - name: /tmp/tmp_dir
    - onfail:
      - pip_installed

source_download:
  cmd.run:
    - name: download from source to dir
    - onfail:
      - pip_installed
    - require:
      - ensure_dir

install_from_source:
  cmd.run:
    - name: install from source
    - require: 
      - source_download
    - onfail:
      - pip_installed
```
### Steps to Reproduce Issue
(Include debug logs if possible and relevant.)
pip_installed results in True
ensure_dir does not run due to onfail requisite. Results in True.
source_download runs because it has require requisite on ensure_dir, despite having a onfail requite on pip_installled. 

### Versions Report
(Provided by running `salt --versions-report`. Please also mention any differences in master/minion versions.)
Salt Version:
           Salt: 2016.11.6

garethgreenaway:
@JeffLee123 Thanks for the report.  Looks like you found a use case that we previous hadn't considered.  Using `onfail` and `require` at the same time.  We'll get this fixed.

PR #46571, opened 2018-03-16
[2017.7] fixes to state.py

garethgreenaway:
### What does this PR do?
Accounting for a case when multiple onfails are used along with requires.  

#46552 

### Previous Behavior
Previously if you have multiple states using 'onfail' and two of those states using a 'require' against the first one state, the last two will run even if the 'onfail' isn't met because the 'require' is met because the first state returns true even though it didn't execute.

### New Behavior
This change adds an additional hidden variable that is used when checking requisities to determine if the state actually ran.

### Tests written?
Yes

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/state.py
+++ b/salt/state.py
@@ -2207,7 +2207,8 @@ class State(object):
                 if r_state == 'prereq' and not run_dict[tag]['result'] is None:
                     fun_stats.add('pre')
                 else:
-                    fun_stats.add('met')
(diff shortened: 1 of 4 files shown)


garethgreenaway:
re-run all

rallytime:
re-run all

rallytime (approved):

PR #46588, opened 2018-03-16
Don't crash when saltwinshell is missing

UtahDave:
### What does this PR do?
Don't crash when saltwinshell is missing
### Tests written?

No

### Commits signed with GPG?

Yes


thatch45 (approved):
Wow, that was a lot easier then we thought at first eh?

rallytime (approved):

rallytime:
re-run py3

PR #46602, opened 2018-03-19
Fix the @with_deprecated decorator.

ghost:
### What does this PR do?

Fixes the `@with_deprecated decorator`, which at the moment (at least in the development branch) breaks.

The steps to reproduce this issue:

1. Add the following snipped at the end of salt/modules/apache.py:

```
from salt.utils.decorators import with_deprecated

def foo(bar='foobar', quiet=False):
     return True


def _foo(bar='foobar', quiet=False):
      return False

```

2. Run `salt-call apache.foo` and get:

```
Passed invalid arguments: coercing to Unicode: need string or buffer, function found.

Usage:
None
  File "/Volumes/User/sjl/work/salt/salt/salt/cli/caller.py", line 212, in call
  File "/Volumes/User/sjl/work/salt/salt/salt/utils/decorators/__init__.py", line 558, in _decorate
  File "/Volumes/User/sjl/work/salt/salt/salt/utils/decorators/__init__.py", line 252, in _call_function
    error = six.text_type(error).replace(self._function, self._orig_f_name)  # Hide hidden functions
TypeError: coercing to Unicode: need string or buffer, function found
```

The correct result should be instead the value True.

The problem seems to be stemming from the fact that `_DeprecationDecorator._get_args()` in `salt/utils/decorators/__init__.py` assumes module calls to have a key `__pub_arg` in `kwargs`, which is not the case.

### What does this PR do?

Replaces `_DeprecationDecorator._get_args()` in `salt/utils/decorators/__init__.py` with a function which calls `salt.utils.args.clean_kwargs` to remove other than function specific keywords (and their values) from `kwargs` (which will then be passed to `_call_function()` etc.).

See the main description above. This is also indirectly related is #46554.

### Previous Behavior

Functions decorated with `@with_deprecated` produced an error (see above) when called with `salt-call mod.func`.

### New Behavior

Functions decorated with `@with_deprecated` work as expected when called with `salt-call mod.func`.

### Tests written?

No

### Commits signed with GPG?

No


samilaine:
This is more or less a precision labeled with @isbm (indirectly related to #46554).

salt/utils/decorators/__init__.py
```diff
@@ -216,22 +216,13 @@ def __init__(self, globals, version):
...
-
-        if '__pub_arg' in kwargs:  # For modules
-            for arg_item in kwargs.get('__pub_arg', list()):
-                if type(arg_item) == dict:
-                    _kwargs.update(arg_item.copy())
-                else:
-                    _args.append(arg_item)
-        else:
-            _kwargs = kwargs.copy()  # For states
+        _kwargs = salt.utils.args.clean_kwargs(**kwargs)
```
isbm: I am happy this is now finally an utility. :wink:

isbm (approved):
Thanks!

rallytime:
re-run all

samilaine:
I'm sorry to ask this, but I don't have a faintest idea what "re-run all" means. All I could do was to push "Update branch" -button, which I hope is what was asked for.

isbm:
@samilaine this is just telling Jenkins to rerun all the tests and see if that breaks anything around. :wink:

cachedout (approved):
Much cleaner.

rallytime (approved):
Yes, much cleaner. 👍

rallytime:
Thanks @samilaine!

PR #46604, opened 2018-03-19
Send only used log records to MP Logging Listener

skizunov:
### What does this PR do?

Currently, the architecture sends all log records to the Multiprocessing
Logging Listener through a multiprocessing queue. On production systems,
this will send many records that ultimately will not be used, including
debug and trace level log records from Salt and any dependencies of Salt
and its plugins. Such log records are numerous and some may be very large
due to containing packet information. This creates an unneeded burden on a
production system.

This change will determine the minimum needed logging level based on
the configuration and use that for the Multiprocessing Logging clients.

### Tests written?

No

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/engines/__init__.py
+++ b/salt/engines/__init__.py
@@ -69,11 +69,11 @@ class Engine(SignalHandlingMultiprocessingProcess):
     '''
     Execute the given engine in a new process
     '''
-    def __init__(self, opts, fun, config, funcs, runners, proxy, log_queue=None):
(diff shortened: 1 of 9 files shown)


cachedout:
@s0undt3ch do you have feelings about this?

s0undt3ch (approved):

s0undt3ch:
Looks good.

cachedout (approved):

PR #46408, opened 2018-03-07
zabbix.host_get should use host instead of name to get host information

angeloudy:
host_get should use host to get the zabbix host info.
This was already explained in the doc.
host means the technical name of the host while name means the visible name.

So here it should use host, not name.

```
    Retrieve hosts according to the given parameters.
    NOTE: This function accepts all optional host.get parameters: keyword argument names differ depending on your
    zabbix version, see: https://www.zabbix.com/documentation/2.4/manual/api/reference/host/get

    .. versionadded:: 2016.3.0

    :param host: technical name of the host
    :param name: visible name of the host
    :param hostids: ids of the hosts
    :param _connection_user: Optional - zabbix user (can also be set in opts or pillar, see module's docstring)
    :param _connection_password: Optional - zabbix password (can also be set in opts or pillar, see module's docstring)
    :param _connection_url: Optional - url of zabbix frontend (can also be set in opts, pillar, see module's docstring)
```

PR #46614, opened 2018-03-19
zabbix should use hostname instead of visible name to get host info

angeloudy:
Resubmit for #46408


rallytime (approved):
Thanks @angeloudy for submitting this cleaned up PR. :)

PR #46621, opened 2018-03-20
Reverted PR#45859 breaking the test

DmitryKuzmenko:
### What does this PR do?
As I can see PR #45859 not only breaks the test but possibly breaks freeipa ldap authorization.
I don't understand how that PR works at all.
I've opened discussion with the author I'll return it back in a proper
way after we'll figure out how it should work.
Possibly I'm wrong.

Fix saltstack/salt-jenkins#881
Reverts #45859 

### Tests written?
No

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/auth/ldap.py
+++ b/salt/auth/ldap.py
@@ -361,7 +361,7 @@ def groups(username, **kwargs):
                                            [salt.utils.stringutils.to_str(_config('accountattributename')), str('cn')])  # future lint: disable=blacklisted-function
 
             for entry, result in search_results:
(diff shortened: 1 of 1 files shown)


terminalmage (commented):
It looks like the PR you referenced actually modified two lines, not just the one. Does the other line also need to be reverted?

DmitryKuzmenko:
@terminalmage the other line is already reverted by somebody else. Probably by some of merge forwards.

terminalmage (approved):
OK, thanks.

DmitryKuzmenko:
tty

PR #46625, opened 2018-03-20
Fix Azure ARM catching the wrong exception type for image references

nicholasmhughes:
### What does this PR do?
This PR fixes part of the Azure ARM Cloud Module which catches the wrong exception type when handling image references in node listings.

N/A

### Previous Behavior
Exception thrown when using a custom image and calling the list_nodes_full function (directly or indirectly).

### New Behavior
Exception handled properly.

### Tests written?
No

### Commits signed with GPG?
Yes


rallytime (approved):

salt/cloud/clouds/azurearm.py
```diff
@@ -533,10 +533,10 @@ def _get_node_info(node):
                 image_ref['sku'],
                 image_ref['version'],
             ])
-        except TypeError:
+        except KeyError:
```
cachedout: Are we certain we don't want to catch both instead of changing from one to the other?

nicholasmhughes: I don't see how it would throw TypeError ever... but I can change it if you think it's best.

nicholasmhughes:
@cachedout updated to catch both Type and Key errors.

PR #46629, opened 2018-03-20
Fix symlink loop when file_roots/pillar_roots is a string instead of a list

terminalmage:
While strings are not the correct data type for these values, using a string instead of a list will cause file_list to try to iterate over each element in the _string_. This will cause it to try to walk the entire filesystem starting at `/`, which will end up causing a symlink loop when walking through `/proc/<PID>`.

This catches instances where a string was passed and ensures we are iterating over a list instead of a string.

### DIFF
--- a/salt/fileclient.py
+++ b/salt/fileclient.py
@@ -863,7 +863,10 @@ class LocalClient(Client):
         if saltenv not in self.opts['file_roots']:
             return ret
         prefix = prefix.strip('/')
-        for path in self.opts['file_roots'][saltenv]:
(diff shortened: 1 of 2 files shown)


rallytime (approved):
👍

rallytime:
@terminalmage There are some test failures here that look related. There's 5 tests failing in `integration.runners.test_fileserver` as seen here: https://jenkins.saltstack.com/job/PR/job/salt-pr-linode-ubuntu16-py3/8054/

Issue #36802, opened 2016-10-05
using clean=True parameter in file.recurse causes python process to spin out of control


rmarcinik:
### Description of Issue/Question

As the title says, I was managing the salt-minion config files with file.recurse.  When I added the clean=True flag, the python process would go crazy and eat up all the cpu.  

Starting the salt-minion-debug service, I would see messages like this repeated over and over

```
[DEBUG   ] In saltenv 'base', ** considering ** path
'c:\salt\var\cache\salt\minion\files\base\salt-minion-config\minion.d\win.conf' to resolve
'salt://salt-minion-config/minion.d/win.conf'
[DEBUG   ] Fetching file from saltenv 'base', ** attempting ** 'salt://salt-minion-config/minion.d/win.conf'
[DEBUG   ] No dest file found
[INFO    ] Fetching file from saltenv 'base', ** done ** 'salt-minion-config/minion.d/win.conf'
[DEBUG   ] Jinja search path: '['c:\\salt\\var\\cache\\salt\\minion\\files\\base']'
[DEBUG   ] In saltenv 'base', looking at rel_path 'salt-minion-config/minion.d/win.conf' to resolve
'salt://salt-minion-config/minion.d/win.conf'
```

The files in question were present on the minion, in the cached files and in the target directory, and on the master.  Without the clean=True flag the python process behaves normally, and there are no errors like above.
### Setup

```
#!pyobjects

File.recurse(
  'salt-minion-config',
  name     = 'C:/salt/conf/minion.d',
  source   = 'salt://salt-minion-config/minion.d',
  defaults = pillar('minion:defaults'),
  template = 'jinja',
  clean      = True
)
```
### Steps to Reproduce Issue

Add/Remove "clean=True" argument
### Versions Report

```
Salt Version:
           Salt: 2016.3.0

Dependency Versions:
           cffi: 1.6.0
       cherrypy: 5.3.0
       dateutil: 2.5.3
          gitdb: 0.6.4
      gitpython: 2.0.2
          ioflo: 1.5.3
         Jinja2: 2.8
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: 1.0.4
   msgpack-pure: Not Installed
 msgpack-python: 0.4.7
   mysql-python: 0.7.2.None
      pycparser: 2.14
       pycrypto: 2.6.1
         pygit2: Not Installed
[...]

PR #46632, opened 2018-03-20
Fix file.recurse w/ clean=True #36802

dwoz:
### What does this PR do?

Fix an endless looping but in file.recurse when clean is set to true.
This patch normalizes paths and handles the root correctly.

#36802 

### Previous Behavior

Using file.recurse with clean = True on windows could result in an endless loop iteration.

### New Behavior

Keep files processed correctly with a loop that always terminates.

### Tests written?

Yes

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/states/file.py
+++ b/salt/states/file.py
@@ -635,6 +635,7 @@ def _clean_dir(root, keep, exclude_pat):
     '''
     real_keep = _find_keep_files(root, keep)
     removed = set()
+
     def _delete_not_kept(nfn):
         if nfn not in real_keep:
             # -- check if this is a part of exclude_pat(only). No need to
--- a/salt/states/file.py
+++ b/salt/states/file.py
@@ -263,6 +263,7 @@ For example:
 
 # Import python libs
 from __future__ import absolute_import
+import collections
 import difflib
 import itertools
 import logging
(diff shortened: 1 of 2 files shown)


salt/states/file.py
```diff
@@ -607,29 +608,34 @@ def _check_file(name):
...
     '''
-    Clean out all of the files and directories in a directory (root) while
-    preserving the files in a list (keep) and part of exclude_pat
+    Compile a list of valid keep files (and directories).
     '''
-    removed = set()
     real_keep = set()
     real_keep.add(root)
-    if isinstance(keep, list):
+    if isinstance(keep, collections.Iterable):
```
terminalmage: This would make this code iterate over strings:

```python
>>> import collections
>>> isinstance('foo', collections.Iterable)
True
```

salt/states/file.py
```diff
@@ -607,29 +608,34 @@ def _check_file(name):
...
                     break
+    return real_keep
+
 
+def _clean_dir(root, keep, exclude_pat):
+    '''
+    Clean out all of the files and directories in a directory (root) while
+    preserving the files in a list (keep) and part of exclude_pat
+    '''
+    real_keep = _find_keep_files_old(root, keep)
```
terminalmage: This looks like a typo, this function doesn't appear to exist.

terminalmage (changes requested):

salt/states/file.py
dwoz: Yes, I guess the question is why would we get a string here. My intent was to not require a list excluding some other iterator or generator.

dwoz: Since this is a bugfix and this change is not essential to the fix, I've reverted it.

salt/states/file.py
dwoz: Fixed.

salt/states/file.py
terminalmage: This is true, I think I was more concerned with changing from list to something that would be more inclusive to types as a potential source of future bugs, and didn't look at the greater context.

Is there a reason to think that other iterable types could be used here? At the moment it does not appear that is the case, but if so, I think this would be a good change to make.

terminalmage (approved):

rallytime:
@dwoz There is a lint error here: https://jenkins.saltstack.com/job/PR/job/salt-pr-lint-n/20427/violations/file/salt/states/file.py/

And a related test failure: https://jenkins.saltstack.com/job/PR/job/salt-pr-linode-ubuntu14-n/21126/

Can you take a look?

dwoz:
re-run all

rallytime:
re-run py2

rallytime (approved):

PR #46634, opened 2018-03-21
Add `fail-fast` option to test runner

dwoz:
### What does this PR do?

Adds `--fail-fast (-f)` option to test suite. This will cause the test runner to exit on the first failure. 

### Commits signed with GPG?

Yes

### DIFF
--- a/tests/support/parser/__init__.py
+++ b/tests/support/parser/__init__.py
@@ -242,6 +242,14 @@ class SaltTestingParser(optparse.OptionParser):
             self, 'Output Options'
         )
         self.output_options_group.add_option(
+            '-f',
+            '--fail-fast',
+            dest='failfast',
+            default=False,
+            action='store_true',
+            help='Stop on first failure'
+        )
+        self.output_options_group.add_option(
             '-v',
             '--verbose',
             dest='verbosity',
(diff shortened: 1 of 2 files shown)


tests/support/parser/__init__.py
```diff
@@ -241,6 +241,14 @@ def __init__(self, testsuite_directory, *args, **kwargs):
         self.output_options_group = optparse.OptionGroup(
             self, 'Output Options'
         )
+        self.output_options_group.add_option(
+            '-f',
```
cachedout: I wonder if this should be `-F`? I always think of the lowercase as `force`.

dwoz: I'm cool with that and will push this minor change.

rallytime:
re-run py

rallytime:
@dwoz Something in here isn't quite right. There are only ~20 tests running with this change. It must be conflicting somewhere. Can you take a look?

dwoz:
@rallytime Yes, I'll check it out. Thanks for the help here.

tests/support/parser/__init__.py
dwoz: @cachedout Pytest uses `-X` not sure if that changes anything on your thoughts here or not. :)

https://docs.pytest.org/en/latest/usage.html#stopping-after-the-first-or-n-failures

rallytime (approved):

Issue #46628, opened 2018-03-20
FreeBSD pkg latest issue


abednarik:
### Description of Issue/Question

We have our custom FreeBSD Repository, where we store updated version of some packages. We just update salt to 2017.7.4 and we found and issue while using `pkg.latest`

This is what salt executes, to get package version, in this case for Nginx. There are a few messages with *DEBUG*, those are some lines added by me to get more info while debugging the issue

```
[DEBUG   ] DEBUG - Command: ['pkg', 'search', '-S', 'name', '-Q', 'version', '-e', '-q', 'nginx']
[INFO    ] Executing command ['pkg', 'search', '-S', 'name', '-Q', 'version', '-e', '-q', 'nginx'] in directory '/home/abednarik'
[DEBUG   ] DEBUG - Package: nginx, Version: 1.12.2_3,2
[DEBUG   ] Could not LazyLoad pkg.normalize_name: 'pkg.normalize_name' is not available.
[INFO    ] Executing command ['pkg', 'install', '-r', 'Perceptyx', '-y', 'nginx'] in directory '/home/abednarik'
[INFO    ] Executing command ['pkg', 'info', '-ao'] in directory '/home/abednarik'
[ERROR   ] Made the following changes:
'nginx' changed from 'absent' to '1.12.2_10,2'
```

As you can see, the search command returns version *1.12.2_3,2* while *1.12.2_10,2* is the latest one from our repos and is the one installed.

State fails, I think [here](https://github.com/saltstack/salt/blob/v2017.7.4/salt/states/pkg.py#L2417-L2420) because installed version is not the same as the one salt stores as the latest one.

```yml
----------
          ID: nginx
    Function: pkg.latest
      Result: False
     Comment: The following packages failed to update: nginx
     Started: 12:25:14.197295
    Duration: 1451.617 ms
     Changes:
              ----------
              nginx:
                  ----------
                  new:
                      1.12.2_10,2
                  old:
```

If I change `reverse=False` [here](https://github.com/saltstack/salt/blob/v2017.7.4/salt/modules/pkgng.py#L318-L320) things work as expected.

### Setup

This is the state

```yml
nginx:
  pkg.latest:
    - name: {{ nginx_pkg }}
{%- if grains['os'] == 'FreeBSD' %}
    - require:
      - cmd: repository_update
    - fromrepo: Company
{% endif %}
  service.running:
    - enable: True
    - watch:
      - file: {{ nginx_config_dir }}
    - require:
[...]

PR #46638, opened 2018-03-21
FreeBSD pkgng module search order.

abednarik:
### What does this PR do?

Do not reverse order when searching for latest version of a package in FreeBSD

#46628

### Previous Behavior

While search for the latest version of a package, since *reverse=True*  is used,  salt doesn't grab the newest version.

### New Behavior

/now the newest version should be the one used in pkg.latest

### Tests written?

No

### Commits signed with GPG?

No


salt/modules/pkgng.py
```diff
@@ -317,7 +317,7 @@ def latest_version(*names, **kwargs):
             name,
             sorted(
                 __salt__['cmd.run'](cmd, python_shell=False, output_loglevel='trace').splitlines(),
-                reverse=True
+                reverse=False
```
isbm: @abednarik but this is not what your PR description says. While it is OK for BSD, it is not OK for others. Therefore I would do it differently:

```python
# Is this BSD-like packaging system?
reverse = not _is_bsd_packaging_system()
...
    sorted(__salt__['cmd.run'](.....
           reverse=reverse)
```

isbm (changes requested):

salt/modules/pkgng.py
abednarik: Hi @isbm  thanks for your comments. 

 Let me try to understand here. As far as I understand pkgng.py is only loaded for FreeBSD >= 10,  DragonFly or if someone manually set pkgng. I know we could use a better approach, but at the same time is hard to me to test it outside FreeBSD.
 So maybe I can set a `reverse` var as you suggest, but I will leave it True as default, since is the way it is now and just set it to False when is FreeBSD.

Open to your feedback.
Thanks.

isbm: @abednarik ahhh, right. I've mistakenly mixed it up with other close-name module. Sorry. Then you could go as this, but please double-check if this does not hurt any other versions.

I am removing the blocker, but please double-check this.

isbm (approved):
@abednarik OK, I am off this, as I am not FreeBSD expert. But please double-check if permanent reverse does not affect other BSD versions.

It is all on you now. :wink:

abednarik:
Ok, thanks. Yes I already tested in FreeBSD and worked. I added some info in the issue Icreated related to this PR.

terminalmage (commented):
Is sorting alphanumerically really the best solution here? Why not just use `LooseVersion`?

```python
        candidates = []
        for line in __salt__['cmd.run'](
                cmd,
                python_shell=False,
                output_loglevel='trace').splitlines():
            pkgver = _get_version(name, line)
            if pkgver is not None:
                candidates.append(LooseVersion(pkgver))

        try:
            pkgver = sorted(candidates, reverse=True)[0].vstring
        except IndexError:
            # No matches
            pkgver = None
```

Note that this would require you to add `from salt.utils.versions import LooseVersion` to the imports of pkgng.py.

abednarik:
Thanks @terminalmage. Will do what you suggest once I have some time.

terminalmage:
Thanks. Specifically I'm worried about the fact that `pkg search` returns all partial matches, so messing with the order may mean you end up with a package name that doesn't exactly match your desired package, which means that `_get_version()` incorrectly returns `None`. With the snippet I posted above (which is untested, I should note), we at least make sure we're only considering exact package name matches in our results.

terminalmage:
Note that you could also do do the sorting like this:

```python
pkgver = sorted(candidates)[-1].vstring
```

This would keep the version numbers in ascending order instead of reversing, and then you'd just grab the _last_ element in the list instead of the _first_.

cachedout:
@abednarik Are you able to return back to this?

abednarik:
Hope so. Busy this days, hope tomorrow I can try to implement this.

Sorry for the delay.

cachedout:
@abednarik Cool. Thanks! Please keep us posted.

abednarik:
Hi

 Yeah, sorry for the delay. Need to finish some tasks at work first.

rallytime:
Hey @abednarik - I am going to close this for now until you have a chance to follow up on the feedback above. Let us know when you're ready and we will be happy to open this again! Or feel free to open a new PR, whatever is easiest for you. 🙂

abednarik:
Sure.

PR #46639, opened 2018-03-21
Use the correct path for nacl certificates in Windows

terminalmage:
This prevents the path from being `C:\path\to\pki\master/nacl` on
Windows.

This is a follow-up to #46426.


rallytime (approved):

rallytime:
re-run py

PR #46641, opened 2018-03-21
Make LazyLoader thread safe

skizunov:
### What does this PR do?

Even when `multiprocessing` is set to `True`, there is a case where
multiple threads in the same process attempt to use the same LazyLoader
object. When using the reactor and reacting to an event that will call a
runner, `salt.utils.reactor.ReactWrap.runner` will invoke
`self.pool.fire_async(self.client_cache['runner'].low, args=(fun, kwargs))`
potentially multiple times, each time using a thread from
`salt.utils.process.ThreadPool`. Each thread will invoke
`salt.client.mixins.SyncClientMixin.low` which in turn will invoke its
`_low` and call `salt.utils.job.store_job`. `salt.utils.job.store_job`
will invoke the LazyLoader object for the returner.

Since the LazyLoader object is not thread safe, occasional failures will
occur which will reduce the reliability of the overall system.

Let's examine why a function such as `LazyLoader._load` is not thread safe.
Any time the GIL is released, it allows another thread to run. There are
various types of operations that could release the GIL, but in this
particular case they are file operations that happen in both
`refresh_file_mapping` and `_load_module`. Note that if you add `print`
statements, those also release the GIL (and make the problem more
frequent). In the failure case, `refresh_file_mapping` releases the
GIL, another thread loads the module, and then when the original thread
runs again it will fail when `_inner_load` runs the second time (after
`refresh_file_mapping`). The failure is because the module is already in
`self.loaded_files`, so it is skipped over and `_inner_load` returns
`False` even though the required `key` is already in `self._dict`. Since
adding in stuff like `print` statements, or other logic also adds points in
the code that allow thread switches, the most robust solution to such a
problem is to use a mutex (as opposed to rechecking if `key` now appears
in `self._dict` at certain checkpoints).

This solution adds such a mutex and uses it in key places to ensure
integrity.

### Tests written?

No

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/loader.py
+++ b/salt/loader.py
@@ -14,6 +14,7 @@ import logging
 import inspect
 import tempfile
 import functools
+import threading
 import types
 from collections import MutableMapping
 from zipimport import zipimporter
@@ -1100,7 +1101,8 @@ class LazyLoader(salt.utils.lazy.LazyDict):
         self.disabled = set(self.opts.get('disable_{0}{1}'.format(
             self.tag, '' if self.tag[-1] == 's' else 's'), []))
 
-        self.refresh_file_mapping()
+        self._lock = threading.RLock()
+        self._refresh_file_mapping()
 
         super(LazyLoader, self).__init__()  # late init the lazy loader
         # create all of the import namespaces
@@ -1167,7 +1169,7 @@ class LazyLoader(salt.utils.lazy.LazyDict):
                 else:
                     return '\'{0}\' __virtual__ returned False'.format(mod_name)
 
-    def refresh_file_mapping(self):
+    def _refresh_file_mapping(self):
         '''
         refresh the mapping of the FS on disk
         '''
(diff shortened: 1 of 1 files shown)


DmitryKuzmenko (approved):

cachedout:
Go Go Jenkins!

rallytime:
@cachedout The latest run looks better. Those current failures happen sometimes on the PR runs. Do you want to review this again?

cachedout (approved):

mattp-:
@DmitryKuzmenko looks like this supersedes and fixes whats referenced in 45782 can you confirm? I'll close #48598 if so. thanks

DmitryKuzmenko:
@mattp- I don't think so. The corner idea of #45782 is using of a thread local store to keep the injected values. But this one just takes care of the race conditions.

PR #46643, opened 2018-03-21
fix docker return

mcalmer:
### What does this PR do?

Make it possible to use login, pull and push from module.run and get result correct

when using state.apply module.run doing docker operations retcode
is tracked to find out if the call was successful or not.

Example:
```
mgr_registries_login:
  module.run:
    - name: dockerng.login
    - registries: {{ pillar.get('docker-registries', {}).keys() }}

mgr_image_profileupdate:
  module.run:
    - name: dockerng.sls_build
    - m_name: "{{ container_name }}"
    - base: "{{ pillar.get('imagename') }}"
    - mods: packages.profileupdate
    - dryrun: True
    - kwargs:
        entrypoint: ""
    - require:
      - module: mgr_registries_login
```

The require can only work if a failed login is seen as failed (result: false).
This is the case when setting `retcode` correctly when a dictionary is returned.

### Tests written?

No

### Commits signed with GPG?

No

### DIFF
--- a/tests/unit/modules/test_dockermod.py
+++ b/tests/unit/modules/test_dockermod.py
@@ -81,8 +81,8 @@ class DockerTestCase(TestCase, LoaderModuleMockMixin):
             with patch.object(docker_mod, '_get_client', get_client_mock):
                 with patch.dict(docker_mod.__salt__, {'cmd.run_all': MagicMock(return_value=ref_out)}):
                     ret = docker_mod.login('portus.example.com:5000')
-                    self.assertTrue('retcode' in ret)
-                    self.assertTrue(ret['retcode'] > 0)
+                    self.assertIn('retcode', ret)
+                    self.assertNotEqual(ret['retcode'], 0)
 
     def test_ps_with_host_true(self):
         '''
--- a/tests/unit/modules/test_dockermod.py
+++ b/tests/unit/modules/test_dockermod.py
@@ -64,6 +64,26 @@ class DockerTestCase(TestCase, LoaderModuleMockMixin):
         '''
         docker_mod.__context__.pop('docker.client', None)
 
+    def test_failed_login(self):
+        '''
+        Check that when docker.login failed a retcode other then 0
+        is part of the return.
+        '''
+        client = Mock()
+        get_client_mock = MagicMock(return_value=client)
+        ref_out = {
+            'stdout': '',
+            'stderr': 'login failed',
+            'retcode': 1
+        }
+        with patch.dict(docker_mod.__pillar__, {'docker-registries': {'portus.example.com:5000':
+                {'username': 'admin', 'password': 'linux12345', 'email': 'tux@example.com'}}}):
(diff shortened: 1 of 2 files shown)


terminalmage (changes requested):
Please point this at the `2018.3` branch instead, `2018.3.0rc1` will soon be converted to the 2018.3.0 release and is not getting any new commits that aren't release blockers. This means that the fix will end up in 2018.3.1.

mcalmer:
@terminalmage rebased to `2018.3` branch

DmitryKuzmenko (approved):

tests/unit/modules/test_dockermod.py
```diff
@@ -64,6 +64,26 @@ def setUp(self):
...
+            'stdout': '',
+            'stderr': 'login failed',
+            'retcode': 1
+        }
+        with patch.dict(docker_mod.__pillar__, {'docker-registries': {'portus.example.com:5000':
+                {'username': 'admin', 'password': 'linux12345', 'email': 'tux@example.com'}}}):
+            with patch.object(docker_mod, '_get_client', get_client_mock):
+                with patch.dict(docker_mod.__salt__, {'cmd.run_all': MagicMock(return_value=ref_out)}):
+                    ret = docker_mod.login('portus.example.com:5000')
+                    self.assertTrue('retcode' in ret)
```
terminalmage: Can you change this to `self.assertIn('retcode', ret)`? It tests the same condition, but the output in case of a failure will contain additional context that isn't there in a simple True/False check.

tests/unit/modules/test_dockermod.py
```diff
@@ -64,6 +64,26 @@ def setUp(self):
...
+                    self.assertTrue(ret['retcode'] > 0)
```
terminalmage: Can you change this to `self.assertNotEqual(ret['retcode'], 0)`? Again, this provides additional context (i.e. the actual value of `ret['retcode']`) in the event that the condition is not met.

terminalmage (changes requested):
Just a couple minor changes in the tests to make the failures provide additional context. Otherwise this is good to merge.

terminalmage:
I disagree that the return was "broken", it worked just fine. It just did not provide a retcode, which module.run can use if present to signal success or failure. If one were to define any function which didn't provide a retcode as "broken", then probably over 90% of functions in Salt are "broken".

This is still an excellent feature addition, though. :+1:

tests/unit/modules/test_dockermod.py
mcalmer: yes, was not aware that this exists

tests/unit/modules/test_dockermod.py
mcalmer: yes, makes sense

mcalmer:
@terminalmage do not take "broken" too serious. I had to say it to have a chance to get it in 2018.3 and not be pushed to develop :-)

terminalmage (approved):

Issue #46348, opened 2018-03-05
Build 449: unit.returners.test_local_cache


twangboy:
[unit.returners.test_local_cache.LocalCacheCleanOldJobsTestCase.test_clean_old_jobs_empty_jid_dir_removed](https://jenkins.saltstack.com/job/2017.7/job/salt-dev-c7-windows-py2/449/testReport/junit/unit.returners.test_local_cache/LocalCacheCleanOldJobsTestCase/test_clean_old_jobs_empty_jid_dir_removed/)
[unit.returners.test_local_cache.LocalCacheCleanOldJobsTestCase.test_clean_old_jobs_jid_file_is_cleaned](https://jenkins.saltstack.com/job/2017.7/job/salt-dev-c7-windows-py2/449/testReport/junit/unit.returners.test_local_cache/LocalCacheCleanOldJobsTestCase/test_clean_old_jobs_jid_file_is_cleaned/)

PR #46646, opened 2018-03-21
Fix `unit.returners.test_local_cache` for Windows

twangboy:
### What does this PR do?
Fixes an issue with the `unit.returners.test_local_cache` test on Windows. `time.time()` is only precise to 3 decimal places in Windows. The test passes a `keep_jobs` time of `.0000000001` seconds. The module compares the date/time stamp on the file to the current time from `time.time()`. When the file time stampe is subtracted from `time.time()` it produces a negative number due to the in-precision of the `time.time()` function on Windows. This PR adds a sleep for .25 seconds on Windows to allow the test to pass.
Additionally, this PR fixes a lint error.

https://github.com/saltstack/salt/issues/46348

### Tests written?
Yes

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/returners/local_cache.py
+++ b/salt/returners/local_cache.py
@@ -420,7 +420,7 @@ def clean_old_jobs():
                     shutil.rmtree(t_path)
                 elif os.path.isfile(jid_file):
                     jid_ctime = os.stat(jid_file).st_ctime
-                    hours_difference = (time.time()- jid_ctime) / 3600.0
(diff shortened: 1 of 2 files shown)


cachedout:
I was a little worried about this but in looking at this, I think this still preserves the behavior of this test is preserved. It would be good if somebody else took a look at this as well.

rallytime (approved):

Issue #46351, opened 2018-03-05
Build 449: integration.modules.test_grains


twangboy:
[integration.modules.test_grains.GrainsAppendTestCase.test_grains_append_val_is_list](https://jenkins.saltstack.com/job/2017.7/job/salt-dev-c7-windows-py2/449/testReport/junit/integration.modules.test_grains/GrainsAppendTestCase/test_grains_append_val_is_list/)
[integration.modules.test_grains.GrainsAppendTestCase.test_grains_append_val_already_present](https://jenkins.saltstack.com/job/2017.7/job/salt-dev-c7-windows-py2/449/testReport/junit/integration.modules.test_grains/GrainsAppendTestCase/test_grains_append_val_already_present/)

twangboy:
Not failing on solo test run

PR #46647, opened 2018-03-21
Fix the tear down function in integration.modules.test_grains

twangboy:
### What does this PR do?
`integration.modules.test_grains.GrainsAppendTestCase`
Fix the `tearDown` function to just set the grain to an empty list instead of removing each individual item.

https://github.com/saltstack/salt/issues/46351

### Tests written?
Yes

### Commits signed with GPG?
Yes

### DIFF
--- a/tests/integration/modules/test_grains.py
+++ b/tests/integration/modules/test_grains.py
@@ -152,8 +152,7 @@ class GrainsAppendTestCase(ModuleCase):
     GRAIN_VAL = 'my-grain-val'
 
     def tearDown(self):
-        for item in self.run_function('grains.get', [self.GRAIN_KEY]):
-            self.run_function('grains.remove', [self.GRAIN_KEY, item])
+        self.run_function('grains.setval', [self.GRAIN_KEY, []])
 
     def test_grains_append(self):
         '''


tests/integration/modules/test_grains.py
```diff
@@ -152,8 +152,7 @@ class GrainsAppendTestCase(ModuleCase):
     GRAIN_VAL = 'my-grain-val'
 
     def tearDown(self):
-        for item in self.run_function('grains.get', [self.GRAIN_KEY]):
-            self.run_function('grains.remove', [self.GRAIN_KEY, item])
```
cachedout: OK, but why is this necessary? I don't understand what's being fixed here. Is `grains.remove` not doing the right thing?

twangboy: In rare cases the `grains.get` function was returning a boolean value and the following stack trace was thrown:
```
  File "c:\testing\tests\integration\modules\test_grains.py", line 155, in tearDown
    for item in self.run_function('grains.get', [self.GRAIN_KEY]):
TypeError: 'bool' object is not iterable
```
So rather than loop over the grains just to remove each item, we decided to just set it to an empty list.

rallytime: @cachedout Can you swing back around here when you get a moment?

cachedout (approved):

PR #46651, opened 2018-03-21
[develop] schedule clean up and error tests

garethgreenaway:
### What does this PR do?
Removing some duplicate code when handling "when" schedule items. 
Moving tests looking for errors in test_errors.  
Adding new tests to account for other error situations.

### Tests written?
Yes

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/utils/schedule.py
+++ b/salt/utils/schedule.py
@@ -1125,8 +1125,8 @@ class Schedule(object):
                 return data
             else:
                 if isinstance(data['skip_during_range'], dict):
-                    start = dateutil_parser.parse(data['skip_during_range']['start'])
-                    end = dateutil_parser.parse(data['skip_during_range']['end'])
+                    start = data['skip_during_range']['start']
+                    end = data['skip_during_range']['end']
                     if not isinstance(start, datetime.datetime):
                         try:
                             start = dateutil_parser.parse(start)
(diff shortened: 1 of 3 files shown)


salt/utils/schedule.py
```diff
@@ -911,38 +911,27 @@ def _handle_when(job, data, loop_interval):
...
-                            when__ = dateutil_parser.parse(i)
-                        except ValueError:
-                            data['_error'] = ('Invalid date string {0}. '
-                                              'Ignoring job {1}.'.format(i, job))
-                            log.error(data['_error'])
-                            return data
+                        __when = i
+
+                    try:
+                        when__ = dateutil_parser.parse(__when)
```
s0undt3ch: First check if its already a date time object, if not parse.

s0undt3ch (changes requested):

garethgreenaway:
@s0undt3ch :+1: Grabbed the other `isinstance` checks from #46540

s0undt3ch (approved):
Looks right now :+1:

s0undt3ch:
We did find another issue though. Happening on 2018.3.
If you what 2 when entries, the second never fires because _next_start_time is set to the first fire time and in the checks, since that date is way in the past, compared to now - interval, it never fires the 2nd entry.

I'll submit a PR against 2018.3

s0undt3ch:
Never mind my last comment. I was not updating all of the `_` prefixed fields on our end causing the scheduler to misbehave.

Issue #46354, opened 2018-03-05
Build 449: unit.test_state


twangboy:
[unit.test_state.StateCompilerTestCase.test_render_error_on_invalid_requisite](https://jenkins.saltstack.com/job/2017.7/job/salt-dev-c7-windows-py2/449/testReport/junit/unit.test_state/StateCompilerTestCase/test_render_error_on_invalid_requisite/)

Issue #46350, opened 2018-03-05
Build 449: unit.test_pyobjects.RendererTests


twangboy:
[unit.test_pyobjects.RendererTests.test_basic](https://jenkins.saltstack.com/job/2017.7/job/salt-dev-c7-windows-py2/449/testReport/junit/unit.test_pyobjects/RendererTests/test_basic/)
[unit.test_pyobjects.RendererTests.test_extend](https://jenkins.saltstack.com/job/2017.7/job/salt-dev-c7-windows-py2/449/testReport/junit/unit.test_pyobjects/RendererTests/test_extend/)
[unit.test_pyobjects.RendererTests.test_import_random_password](https://jenkins.saltstack.com/job/2017.7/job/salt-dev-c7-windows-py2/449/testReport/junit/unit.test_pyobjects/RendererTests/test_import_random_password/)
[unit.test_pyobjects.RendererTests.test_import_scope](https://jenkins.saltstack.com/job/2017.7/job/salt-dev-c7-windows-py2/449/testReport/junit/unit.test_pyobjects/RendererTests/test_import_scope/)
[unit.test_pyobjects.RendererTests.test_include](https://jenkins.saltstack.com/job/2017.7/job/salt-dev-c7-windows-py2/449/testReport/junit/unit.test_pyobjects/RendererTests/test_include/)
[unit.test_pyobjects.RendererTests.test_invalid_function](https://jenkins.saltstack.com/job/2017.7/job/salt-dev-c7-windows-py2/449/testReport/junit/unit.test_pyobjects/RendererTests/test_invalid_function/)
[unit.test_pyobjects.RendererTests.test_random_password](https://jenkins.saltstack.com/job/2017.7/job/salt-dev-c7-windows-py2/449/testReport/junit/unit.test_pyobjects/RendererTests/test_random_password/)
[unit.test_pyobjects.RendererTests.test_requisite_implicit_list](https://jenkins.saltstack.com/job/2017.7/job/salt-dev-c7-windows-py2/449/testReport/junit/unit.test_pyobjects/RendererTests/test_requisite_implicit_list/)
[unit.test_pyobjects.RendererTests.test_sls_imports](https://jenkins.saltstack.com/job/2017.7/job/salt-dev-c7-windows-py2/449/testReport/junit/unit.test_pyobjects/RendererTests/test_sls_imports/)

twangboy:
I believe @terminalmage said he needed to fix this one

twangboy:
[...]

Issue #46349, opened 2018-03-05
Build 449: unit.test_pydsl


twangboy:
[unit.test_pydsl.PyDSLRendererTestCase.test_cmd_call](https://jenkins.saltstack.com/job/2017.7/job/salt-dev-c7-windows-py2/449/testReport/junit/unit.test_pydsl/PyDSLRendererTestCase/test_cmd_call/)
[unit.test_pydsl.PyDSLRendererTestCase.test_pipe_through_stateconf](https://jenkins.saltstack.com/job/2017.7/job/salt-dev-c7-windows-py2/449/testReport/junit/unit.test_pydsl/PyDSLRendererTestCase/test_pipe_through_stateconf/)
[unit.test_pydsl.PyDSLRendererTestCase.test_repeat_includes](https://jenkins.saltstack.com/job/2017.7/job/salt-dev-c7-windows-py2/449/testReport/junit/unit.test_pydsl/PyDSLRendererTestCase/test_repeat_includes/)

twangboy:
Related issues: #46350 #46354 

This is what @s0undt3ch said in regards to this issue
"I haven't touched that in far too long, nor I have time current to mess with that.
But mock's patch function is not unpatching and that's the reason those changes are leaking..."

Issue #46345, opened 2018-03-05
Build 449: unit.test_pyobjects.MapTests (Manual Pass)


twangboy:
[unit.test_pyobjects.MapTests.test_map](https://jenkins.saltstack.com/job/2017.7/job/salt-dev-c7-windows-py2/449/testReport/junit/unit.test_pyobjects/MapTests/test_map/)
[unit.test_pyobjects.MapTests.test_map_with_priority](https://jenkins.saltstack.com/job/2017.7/job/salt-dev-c7-windows-py2/449/testReport/junit/unit.test_pyobjects/MapTests/test_map_with_priority/)

PR #46655, opened 2018-03-22
Fixing cleanUp method to restore environment

dwoz:
### What does this PR do?

The clean-up method was not reliabliy restoring sys.environment on
windows. Using Mock's patch to cleanup more reliably.

#46350, #46345, #46349, #46354

### Previous Behavior

cleanUp method was not restoring sys.environ properly (at least on windows)

### New Behavior

Mock's patch method restores environ properly

### Tests written?

No - This fixes a hand full of existing tests.

### Commits signed with GPG?

Yes

### DIFF
--- a/tests/unit/states/test_environ.py
+++ b/tests/unit/states/test_environ.py
@@ -27,12 +27,11 @@ class TestEnvironState(TestCase, LoaderModuleMockMixin):
         return {envstate: loader_globals, envmodule: loader_globals}
 
     def setUp(self):
-        original_environ = os.environ.copy()
-        os.environ = {'INITIAL': 'initial'}
-
-        def reset_environ(original_environ):
-            os.environ = original_environ
-        self.addCleanup(reset_environ, original_environ)
+        patcher = patch.dict(os.environ, {'INITIAL': 'initial'})
+        patcher.start()
+        def reset_environ(patcher):
+            patcher.stop()
+        self.addCleanup(reset_environ, patcher)
 
     def test_setenv(self):
         '''test that a subsequent calls of setenv changes nothing'''
--- a/tests/unit/states/test_environ.py
+++ b/tests/unit/states/test_environ.py
@@ -6,7 +6,7 @@ import os
 
 # Import Salt Testing libs
 from tests.support.mixins import LoaderModuleMockMixin
-from tests.support.unit import TestCase
+from tests.support.unit import TestCase, skipIf
 from tests.support.mock import (
     MagicMock,
     patch
--- a/tests/unit/states/test_environ.py
+++ b/tests/unit/states/test_environ.py
@@ -14,8 +14,8 @@ from tests.support.mock import (
 # Import salt libs
 import salt.states.environ as envstate
 import salt.modules.environ as envmodule
-import salt.modules.reg as regmodule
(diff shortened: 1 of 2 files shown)


dwoz:
Looks like this is breaking some other tests in unit.states.test_environ. Making this a WIP until I resolve those.

dwoz:
Adding `clean=True` to the patch call seems to have done the trick. I think this is good to go.

dwoz:
I ended up modifying `states.environ.setenv` to handle windows environments a bit more reasonably. The crux of the issue is that environment variables on windows are not case sensitive. Keys are always represented as upper case however.

Without my patch the `tests.unit.states.test_environ.TestEnvironState.test_setenv` test will always fail on the third assertion because `setenv` was in-correctly updating the key `test` which already exists in the environment.

twangboy (approved):

rallytime:
@dwoz Looks like there's a related test failure here and some lint. Can you take another look?

rallytime:
@dwoz This is still failing a related test on linux: https://jenkins.saltstack.com/job/PR/job/salt-pr-linode-ubuntu14-n/21135/

dwoz:
I'm hoping we'll be good after the last commit.

dwoz:
re-run all

rallytime:
@dwoz One more lint fix :)

https://jenkins.saltstack.com/job/PR/job/salt-pr-lint-n/20489/violations/file/tests/unit/states/test_environ.py/

dwoz:
@rallytime That linter is doing it's job.

rallytime:
re-run py

tests/unit/states/test_environ.py
```diff
@@ -6,14 +6,16 @@
...
+from tests.support.unit import TestCase, skipIf
 from tests.support.mock import (
     MagicMock,
     patch
 )
 # Import salt libs
 import salt.states.environ as envstate
 import salt.modules.environ as envmodule
+import salt.modules.reg as regmodule
+import salt.utils as utils
```
rallytime: I know the `environ` state above was doing it this way, but we are trying to make these imports more uniform. Can you change this to use the full path? That makes it easier to lint/check for these references in newer branches that have diverged from using the funcs in the `salt.utils.__init__.py` file.

tests/unit/states/test_environ.py
```diff
@@ -47,9 +53,10 @@ def test_setenv(self):
         ret = envstate.setenv('test', 'other')
         self.assertEqual(ret['changes'], {})
 
+    @skipIf(not utils.is_windows(), 'Windows only')
     def test_setenv_permanent(self):
-        with patch.dict(envmodule.__salt__, {'reg.set_value': MagicMock(), 'reg.delete_value': MagicMock()}), \
-                patch('salt.utils.is_windows', MagicMock(return_value=True)):
+        '''test that we can set perminent environment variables (requires pywin32)'''
```
rallytime: Can you split up this doc line into 3 lines?

rallytime (changes requested):
Just some stylist things that need to be changed, otherwise 👍

rallytime:
re-run py2

rallytime (approved):

PR #46656, opened 2018-03-22
Replace fnmatch with re for tag matching in state.event()

ghost:
### What does this PR do?

This PR replaces the old fnmatch-style matching for tags in `state.event()` with re-style matching.

The merits of replacing `fnmatch` with `re` are discussed in #46554. The general, albeit sparse, consensus seemed to be that this is the way to go. It's definitely a cleaner way than using try/expect to figure out whether or not `tagmatch` parameter contains a regular expression, and if it doesn't, to fall back to `fnmatch` -- it's hard to see how the resulting behaviour would be obvious for a user simply typoing a regular expression.

This PR contains code using the `@with_deprecated` decorator, which is broken in the branch `develop` at the time of opening this PR. So, this PR depends on #46602 being merged before or `@with_deprecated` getting fixed in some other way.

Sometimes it gets rather tedious to debug event-related problems (ie. broken orchestrator states and so on) on a master which is really busy (200+ events per second) without a decent matching capability to filter events. Using regular expressions to filter events displayed makes the life of an administrator a tad easier.

### Previous Behavior

The function `state.event()` used fnmatch to filter events.

### New Behavior

The function `state.event()` uses re to filter events.

### Tests written?

No (because I couldn't make heads or tails of the testing documentation -- any/all tutors and mentors are welcome to share their wisdom on the unit-testing
system should use mocks and whatnot).

### Commits signed with GPG?

No


salt/modules/state.py
```diff
@@ -2286,7 +2288,85 @@ def _disabled(funs):
...
+        )
+
+    sevent = salt.utils.event.get_event(
+            node,
+            sock_dir or __opts__['sock_dir'],
+            __opts__['transport'],
+            opts=__opts__,
+            listen=True)
+
+    we_are_counting = True if count > -1 else False
```
isbm: uhm... why not just:
```python
my_quite_strange_variable_name = count > -1
```
...? :wink:

salt/modules/state.py
```diff
@@ -2286,7 +2288,85 @@ def _disabled(funs):
...
+                            ret['data'],
+                            sort_keys=pretty,
+                            indent=None if not pretty else 4)
+                    )
+                )
+                sys.stdout.flush()
+
+            if we_are_counting:
+                count -= 1
+                if count == 0:
```
isbm: First, you can just:
```python
if not count:
    break
```
But this is still wrong. The `-1` is "listen forever". Can be `count=0`? If yes, what does it mean then other, than "do not count"? Still, you do not need this `we_are_watching_some_other_variable`. Just this:

```python
if count > 0:
    count -= 1
if not count:
    break
```

salt/modules/state.py
```diff
@@ -2286,7 +2288,85 @@ def _disabled(funs):
...
+                    break
+                log.debug('Remaining event matches: %s', count)
+
+        else:
+            log.debug('Skipping event tag: %s', ret['tag'])
+            continue
```
isbm: It continues anyway, no need for this.

salt/modules/state.py
```diff
@@ -2286,7 +2288,85 @@ def _disabled(funs):
...
+
+    while True:
+        ret = sevent.get_event(full=True, auto_reconnect=True)
+        if ret is None:
+            continue
+
+        if patt.search(ret['tag']) is not None:
+            if not quiet:
+                salt.utils.stringutils.print_cli(
+                    str('{0}\t{1}').format(  # future lint: blacklisted-function
```
isbm: Why `str` here?
```python
str('{0}\t{1}')
```

salt/modules/state.py
```diff
@@ -2286,7 +2288,85 @@ def _disabled(funs):
...
+    sevent = salt.utils.event.get_event(
+            node,
+            sock_dir or __opts__['sock_dir'],
+            __opts__['transport'],
+            opts=__opts__,
+            listen=True)
+
+    we_are_counting = True if count > -1 else False
+
+    while True:
```
isbm: What happens if `count == -1`?

isbm: If you would write a unit test for this thing, you would quickly find that it contains this bug. :wink: You can look at [`test_ansiblegate.py`](https://github.com/samilaine/salt/blob/develop/tests/unit/modules/test_ansiblegate.py) as a reference how to write them.

Generally, you can just run them with `pytest`, like so:

```
$ cd $GIT/salt/tests/unit/modules
$ pytest test_state.py
```

Just add your test(s) to the `test_state.py` file.

isbm (changes requested):

terminalmage (commented):
Have you considered the number of users who are likely relying on the current behavior? This could be quite disruptive to them.

We have a function in `salt.utils.stringutils` called `expr_match()` which tries a full string match, then a glob, then a regex. This seems like it would be a great compromise, and _much_ less invasive than this PR. All that would need to be done would be to replace the `fnmatch.fnmatch()` call with one to `salt.utils.stringutils.expr_match()`, and then add in a `.. versionchanged:: Fluorine` after the `.. versionadded::` line, which explains that regex matches are now supported:

```
.. versionadded:: Fluorine
    ``tagmatch`` can now be either a glob or regex
```

And actually, having looked at `expr_match()`, I noticed some improvements that could be made to it, which I did in https://github.com/saltstack/salt/pull/46729.

samilaine:
As @terminalmage wrote, this is rather disruptive indeed.

A previous attempt, #46554, headed on that direction. Albeit failing to note `salt.utils.stringutils.expr_match()` exists, that attempt added a parameter which seemed like a bad idea during the review -- a conclusion I tend to agree with.

So, going full circle ...

samilaine:
For search engines: See #46811.

Issue #46150, opened 2018-02-22
With chocolately.version some packages don't work with check_remote=True


whytewolf:
### Description of Issue/Question

Looking into this issue it looks like some packages have different capitalization when checking the remote version. so when used with `check_remote=True` the package is not matching any of the contents of the available dict. Which results in the following error. 

```
Traceback (most recent call last):
  File "C:\salt\bin\Scripts\salt-call", line 11, in <module>
    salt_call()
  File "C:\salt\bin\lib\site-packages\salt\scripts.py", line 391, in salt_call
    client.run()
  File "C:\salt\bin\lib\site-packages\salt\cli\call.py", line 58, in run
    caller.run()
  File "C:\salt\bin\lib\site-packages\salt\cli\caller.py", line 134, in run
    ret = self.call()
  File "C:\salt\bin\lib\site-packages\salt\cli\caller.py", line 204, in call
    ret['return'] = func(*args, **kwargs)
  File "C:\salt\bin\lib\site-packages\salt\modules\chocolatey.py", line 889, in version
    'available': available[pkg]}
KeyError: 'freesshd'
```
I have only tested with `freesshd` which to the remote system shows up as `freeSSHD`. 

This issue also stops `salt.state.chocolatey.installed` from testing already installed programs that this issue affects.  

### Setup
Install chocolatey on a windows machine. install freesshd then check it's version with `salt-call chocolatey.version freesshd check_remote=True` 

### Steps to Reproduce Issue
install freesshd then check it's version with `salt-call chocolatey.version freesshd check_remote=True`

### Versions Report
```
Salt Version:
           Salt: 2017.7.3

Dependency Versions:
           cffi: 1.10.0
       cherrypy: unknown
       dateutil: 2.6.0
      docker-py: Not Installed
          gitdb: 2.0.3
      gitpython: 2.1.3
          ioflo: Not Installed
         Jinja2: 2.9.6
        libgit2: Not Installed
        libnacl: 1.6.1
       M2Crypto: Not Installed
           Mako: 1.0.6
   msgpack-pure: Not Installed
 msgpack-python: 0.4.8
   mysql-python: Not Installed
      pycparser: 2.17
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: Not Installed
         Python: 2.7.13 (v2.7.13:a06454b1afa1, Dec 17 2016, 20:53:40) [MSC v.1500 64 bit (AMD64)]
   python-gnupg: 0.4.0
[...]

PR #46661, opened 2018-03-22
Chocolatey - Lowered name of local and remote packages before comparing versions.

Kimol:
### What does this PR do?
Fixes issue #46150
#46150

### Previous Behavior

### New Behavior

### Tests written?

No

### Commits signed with GPG?

No

### DIFF
--- a/salt/modules/chocolatey.py
+++ b/salt/modules/chocolatey.py
@@ -920,7 +920,7 @@ def version(name, check_remote=False, source=None, pre_versions=False):
     '''
     installed = list_(narrow=name, local_only=True)
     installed = {k.lower(): v for k, v in installed.items()}
-    
+
     packages = {}
     lower_name = name.lower()
     for pkg in installed:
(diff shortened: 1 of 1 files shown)


twangboy:
@Kimol So is the issue here that the `list_` function returns package names all in lowercase?

Kimol:
@twangboy list_ returns original package names (camelcase in many cases). And the issue in function version() is that only the local package name is lowered (probably to provide case insensitivity). Remote package name is left without conversion, so in this function it's always comparing lowercase local package name with original remote package name. Which will only work when remote package name is originally also lowercase.

twangboy:
yeah, I was going to suggest lowering the case of both.

twangboy (approved):

twangboy:
Looks like the failing tests are unrelated

rallytime:
Welcome @Kimol! Thanks for submitting this fix.

PR #46664, opened 2018-03-22
[develop][schedule] Add support for Job dry runs

s0undt3ch:
### What does this PR do?
Adds dry run support to the schedule

### Tests written?

No

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/utils/schedule.py
+++ b/salt/utils/schedule.py
@@ -426,27 +426,8 @@ class Schedule(object):
 
         # Grab run, assume True
         run = data.get('run', True)
-        run_schedule_jobs_in_background = self.opts.get('run_schedule_jobs_in_background', True)
         if run:
-            if run_schedule_jobs_in_background:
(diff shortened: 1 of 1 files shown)


garethgreenaway (approved):

garethgreenaway:
@s0undt3ch This would good to include in the execution module & state module as options too.

thatch45 (approved):

s0undt3ch:
There's some more code added to this PR.
* Factor out the code that actually runs the schedule
* Fix inversed logic
* Fix a missed `isinstance()` check
* Allow getting a new `Schedule` class instance instead of the singleton instance. Includes test.
* Don't change the process name if not running in a new process
* Fixed the code that allows running in the foreground. It wasn't calling the actual function since we pass the function name, not the instance.

@garethgreenaway Can I ask you the favour of taking care of the salt's function and state modules side of this?

s0undt3ch:
rebased

PR #45025, opened 2017-12-15
Fix pillar include merge order

steverweber:
### What does this PR do?
fix Pillar includes so they don't override whats in the current sls file.

### possible breakage
this change could cause breakage... so perhaps should add a configuration option and have a migration strategy.

### What issues does this PR fix or reference?
https://github.com/saltstack/salt/issues/39860

### Tests written?
Yes, however could use another test for pillars with multiple layers of includes

PR #46669, opened 2018-03-22
Add option to return to pre-2017.7.3 pillar include merge order

terminalmage:
https://github.com/saltstack/salt/pull/45025 changed the pillar include merge order. This adds a new config option which allows a user to revert to the pre-2017.7.3 behavior.


steverweber:
thanks for handling this.

rallytime (approved):

Issue #46673, opened 2018-03-22
Salt forces Cumulus linux to init.d scripts and not systemd


adam-smith-sundance:
### Description of Issue/Question
Cumulus Linux (based on Debian 8.9) seems to fall back to systemv/init scripts for services in salt though it uses systemd and grains show systemd as the init type.  When checking for a service it reports the service is not available if the service is not in /etc/init.d/

### Setup
(Please provide relevant configs and/or SLS files (Be sure to remove sensitive info).)
Using default configuration for the minion (this works just fine in my current environment) and the ntp formula https://github.com/saltstack-formulas/ntp-formula  general ntp state

pillar data
minion:
    ----------
    ntp:
        ----------
        lookup:
            ----------
            service:
                ntpd@mgmt
        servers:
            - ntp.local.server

grains 

minion:
    ----------
    init:
        systemd
    ----------
    systemd:
        ----------
        features:
            +PAM +AUDIT +SELINUX +IMA +SYSVINIT +LIBCRYPTSETUP +GCRYPT +ACL +XZ -SECCOMP -APPARMOR
        version:
            215

### Steps to Reproduce Issue
(Include debug logs if possible and relevant.)

sudo salt minion state.apply ntp

output

----------
          ID: ntp
    Function: pkg.installed
      Result: True
     Comment: All specified packages are already installed
     Started: 15:00:40.188452
    Duration: 1318.864 ms
     Changes:
----------
          ID: ntp_conf
    Function: file.managed
        Name: /etc/ntp.conf
      Result: True
     Comment: File /etc/ntp.conf is in the correct state
     Started: 15:00:41.515548
    Duration: 152.252 ms
     Changes:
----------
          ID: ntp_running
    Function: service.running
        Name: ntp@mgmt
      Result: False
     Comment: The named service ntp@mgmt is not available
     Started: 15:00:41.670473
    Duration: 2.384 ms
     Changes:

sudo salt minion service.restart ntp@mgmt

minion:
[...]

PR #46674, opened 2018-03-22
Update service.py

absmith82:
add Cumulus to the disable list in def __virtual__()

### What does this PR do?
This adds Cumulus Linux into the to the disable list in def __virtual__() to allow it access to systemd.py

Salt forces Cumulus linux to init.d scripts and not systemd #46673 

### Tests written?

Yes/No

### Commits signed with GPG?

Yes/No

### DIFF
--- a/salt/modules/service.py
+++ b/salt/modules/service.py
@@ -34,6 +34,7 @@ def __virtual__():
         'Gentoo',
         'Ubuntu',
         'Debian',
+        'Cumulus',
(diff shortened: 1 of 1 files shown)


rallytime:
Since this same fix is made in #46707 and will be merged-forward, I'm going to close this PR so we don't have to do 2 reviews. Thanks @adam-smith-sundance for fixing this. :)

Issue #46347, opened 2018-03-05
Buid 449: unit.modules.test_inspect_collector


twangboy:
[unit.modules.test_inspect_collector.InspectorCollectorTestCase.test_file_tree](https://jenkins.saltstack.com/job/2017.7/job/salt-dev-c7-windows-py2/449/testReport/junit/unit.modules.test_inspect_collector/InspectorCollectorTestCase/test_file_tree/)

twangboy:
Passing

PR #46675, opened 2018-03-23
Skip test when git symlinks are not configured

dwoz:
### What does this PR do?

Skips symlink tests when git configuration does not support symlinks.

#46347

### Previous Behavior

Tests would fail if git was cloned without 'core.symlinks=true'

### New Behavior

Tests pass when the config supports them and are skipped with a nice message otherwise.

### Tests written?

No - Fixes existing tests

### Commits signed with GPG?

Yes

### DIFF
--- a/tests/unit/modules/test_inspect_collector.py
+++ b/tests/unit/modules/test_inspect_collector.py
@@ -48,9 +48,12 @@ def no_symlinks():
     output = ''
     try:
         output = subprocess.check_output('git config --get core.symlinks', shell=True)
-    except OSError as ex:
-        if es.errno != errno.ENOENT:
+    except OSError as exc:
+        if exc.errno != errno.ENOENT:
             raise
+    except subprocess.CalledProcessError:
+        # git returned non-zero status
+        pass
     HAS_SYMLINKS = False
     if output.strip() == 'true':
         HAS_SYMLINKS = True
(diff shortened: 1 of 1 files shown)


tests/unit/modules/test_inspect_collector.py
```diff
@@ -33,7 +34,23 @@
 from salt.modules.inspectlib.collector import Inspector
 
 
+HAS_SYMLINKS = None
+
+
+def no_symlinks():
```
dwoz: There may be a more 'salty' way of doing this. I looked around a bit but didn't see one.

twangboy: There's also an option in the git installer for Windows where you have to enable symlinks on Windows. I don't know how you would check for that.

dwoz: I'm sure that option will just set the global config option for symlinks. This check will encompass that.

twangboy (approved):

rallytime:
@dwoz Looks like there are some related test failures here: 

- https://jenkins.saltstack.com/job/PR/job/salt-pr-linode-cent7-py3/3402/
- https://jenkins.saltstack.com/job/PR/job/salt-pr-linode-ubuntu14-n/21069/

rallytime:
@dwoz There's still a test failure on this and a lint error.

dwoz:
re-run py

dwoz:
re-run py

rallytime:
re-run py

rallytime (approved):

PR #46687, opened 2018-03-23
[develop][schedule] Fix run in background when it's set to false

s0undt3ch:
### What does this PR do?
Fix a bug when the schedule is configured to run in the foreground

### Tests written?

No

### Commits signed with GPG?

Yes

This should be merged after https://github.com/saltstack/salt/pull/46664


garethgreenaway (approved):

s0undt3ch:
rebased

PR #46692, opened 2018-03-23
saltnado bugfixes for ldap & syndics

mattp-:
### What does this PR do?
fix syndic_wait handling in saltnado in syndic topologies:

 as is, yielding sleep before registering the event listener as actually causing saltnado to _miss_ the event entirely. instead of waiting before registering  instead we just add the syndic_wait to the futures base timeout. This issue only affected synchronous localclient calls

also included, a fix for ldap group based perms that mimics cherrypy compatibility.

### New Behavior
- saltnado now works with syndic topologies
- saltnado now works with group perms with ldap

### Tests written?
Yes

### Commits signed with GPG?

No

### DIFF
--- a/salt/netapi/rest_tornado/saltnado.py
+++ b/salt/netapi/rest_tornado/saltnado.py
@@ -740,7 +740,17 @@ class SaltAuthHandler(BaseSaltAPIHandler):  # pylint: disable=W0223
 
         # Grab eauth config for the current backend for the current user
         try:
-            perms = self.application.opts['external_auth'][token['eauth']][token['name']]
+            eauth = self.application.opts['external_auth'][token['eauth']]
+            # Get sum of '*' perms, user-specific perms, and group-specific perms
+            perms = eauth.get(token['name'], [])
+            perms.extend(eauth.get('*', []))
+
+            if 'groups' in token and token['groups']:
+                user_groups = set(token['groups'])
+                eauth_groups = set([i.rstrip('%') for i in eauth.keys() if i.endswith('%')])
+
+                for group in user_groups & eauth_groups:
+                    perms.extend(eauth['{0}%'.format(group)])
 
         # If we can't find the creds, then they aren't authorized
         except KeyError:
--- a/salt/netapi/rest_tornado/saltnado.py
+++ b/salt/netapi/rest_tornado/saltnado.py
@@ -752,6 +752,7 @@ class SaltAuthHandler(BaseSaltAPIHandler):  # pylint: disable=W0223
                 for group in user_groups & eauth_groups:
                     perms.extend(eauth['{0}%'.format(group)])
 
+            perms = sorted(list(set(perms)))
         # If we can't find the creds, then they aren't authorized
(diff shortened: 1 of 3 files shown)


salt/netapi/rest_tornado/saltnado.py
```diff
@@ -740,7 +740,17 @@ def post(self):
...
+            # Get sum of '*' perms, user-specific perms, and group-specific perms
+            perms = eauth.get(token['name'], [])
+            perms.extend(eauth.get('*', []))
+
+            if 'groups' in token and token['groups']:
+                user_groups = set(token['groups'])
+                eauth_groups = set([i.rstrip('%') for i in eauth.keys() if i.endswith('%')])
+
+                for group in user_groups & eauth_groups:
+                    perms.extend(eauth['{0}%'.format(group)])
```
DmitryKuzmenko: I see a behavior change here. If not user name nor groups are in eauth list before we were getting `send_error(401)`. Now we're returning with empty perms list.
If this change is expected I'm OK. If not it's better to detect match with a flag keeping `send_error` if not name nor group match found.

DmitryKuzmenko (changes requested):
@mattp-  could you please take a look to my inline comment?

salt/netapi/rest_tornado/saltnado.py
mattp-: good catch, i just copied this from rest_cherrypy. I'll restore send_error(401)

mattp-: actually on further review it looks like this is expected behavior in cherrypy. I am indifferent here, is there any policy of attempting to make saltnado compatible with cherrypy? is cherrypy the canonical implementation?

mattp-: on further thought, I think aligning with cherrypy makes sense here. the reason I hit this, and likely why anyone else would, is theyve transitioned from cherrypy with pepper and are moving to tornado. functionality should be 1 to 1 as close as possible. so working as designed :)

DmitryKuzmenko: I'm OK then. Thank you for your work.

DmitryKuzmenko:
re-run all

DmitryKuzmenko (approved):

skizunov (approved):

cachedout (approved):

cachedout:
We hit a test failure here: https://jenkins.saltstack.com/job/PR/job/salt-pr-rs-cent7-n/17790/testReport/junit/integration.netapi.rest_tornado.test_app/TestSaltAPIHandler/test_simple_local_post_only_dictionary_request_with_order_masters/

mattp-:
rerun all

mattp-:
rerun all

cachedout:
re-run all

rallytime:
@mattp- This is causing some related tests to fail: https://jenkins.saltstack.com/job/PR/job/salt-pr-linode-ubuntu16-py3/8619/

Issue #46352, opened 2018-03-05
Build 449: unit.test_client


twangboy:
[unit.test_client.LocalClientTestCase.test_pub](https://jenkins.saltstack.com/job/2017.7/job/salt-dev-c7-windows-py2/449/testReport/junit/unit.test_client/LocalClientTestCase/test_pub/)

PR #46696, opened 2018-03-23
Windows `unit.test_client` fixes

dwoz:
### What does this PR do?

- Skip IPC transport test on windows
- Add TCP transport test for windows

### What issues does this PR fix or reference

#46352

### Previous Behavior

IPC test failed with exception

### New Behavior

Skip IPC test and instead test TCP transport which is the default on windows.

### Tests written?

Yes - New test for windows behavior.

### Commits signed with GPG?

Yes

### DIFF
--- a/tests/unit/test_client.py
+++ b/tests/unit/test_client.py
@@ -96,7 +96,7 @@ class LocalClientTestCase(TestCase,
     @skipIf(not salt.utils.is_windows(), 'Windows only test')
     def test_pub_win32(self):
         '''
-        Tests that the cleint raises a timeout error when using TCP transport
+        Tests that the client raises a timeout error when using TCP transport
         and publisher is not running.
 
         Note: Requires TCP transport, this is only the default on Windows.
--- a/tests/unit/test_client.py
+++ b/tests/unit/test_client.py
@@ -13,7 +13,10 @@ from tests.support.mock import patch, NO_MOCK, NO_MOCK_REASON
 
 # Import Salt libs
 from salt import client
-from salt.exceptions import EauthAuthenticationError, SaltInvocationError, SaltClientError
+import salt.utils
+from salt.exceptions import (
+    EauthAuthenticationError, SaltInvocationError, SaltClientError, SaltReqTimeoutError
+)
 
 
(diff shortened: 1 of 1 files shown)


twangboy (approved):

tests/unit/test_client.py
```diff
@@ -83,3 +92,27 @@ def test_pub(self):
                 self.assertRaises(SaltInvocationError,
                                   self.client.pub,
                                   'non_existent_group', 'test.ping', tgt_type='nodegroup')
+
+    @skipIf(not salt.utils.is_windows(), 'Windows only test')
+    def test_pub_win32(self):
+        '''
+        Tests that the cleint raises a timeout error when using TCP transport
```
rallytime: `cleint` --> `client` :)

tests/unit/test_client.py
```diff
@@ -83,3 +92,27 @@ def test_pub(self):
                 self.assertRaises(SaltInvocationError,
                                   self.client.pub,
                                   'non_existent_group', 'test.ping', tgt_type='nodegroup')
+
+    @skipIf(not salt.utils.is_windows(), 'Windows only test')
```
rallytime: We have tests that run the TCP transport in Jenkins. Perhaps this check should run on windows and those TCP tests as well?

https://jenkinsci.saltstack.com/job/2017.7/view/Transport/

rallytime (changes requested):
This looks good, I just have one small comment and a question here. Thanks @dwoz!

tests/unit/test_client.py
dwoz: Maybe I should be more clear in the docs that this is dealing with ZeroMQ's (TCP vs ICP). Not to be confused with Salt's TCP transport. Does that change things?

rallytime: Ah, yes. Thanks!

rallytime (approved):

tests/unit/test_client.py
dwoz: I've added "ZeroMQ" to the doc strings in order to be more clear about what is being tested.

Issue #46672, opened 2018-03-22
[2018.3.0] Running a state file with a non-ascii char in fails 


frogunder:
### Description of Issue/Question
Running `salt \* state.sls test` when the state file have a non-ascii char in it fails.

```
[ERROR   ] An un-handled exception was caught by salt's global exception handler:
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 27: ordinal not in range(128)
Traceback (most recent call last):
  File "/usr/bin/salt", line 10, in <module>
    salt_main()
  File "/usr/lib/python2.7/site-packages/salt/scripts.py", line 485, in salt_main
    client.run()
  File "/usr/lib/python2.7/site-packages/salt/cli/salt.py", line 197, in run
    self._output_ret(ret_, out, retcode=retcode)
  File "/usr/lib/python2.7/site-packages/salt/cli/salt.py", line 376, in _output_ret
    _retcode=retcode)
  File "/usr/lib/python2.7/site-packages/salt/output/__init__.py", line 94, in display_output
    display_data = try_printout(data, out, opts, **kwargs)
  File "/usr/lib/python2.7/site-packages/salt/output/__init__.py", line 41, in try_printout
    printout = get_printout(out, opts)(data, **kwargs)
  File "/usr/lib/python2.7/site-packages/salt/output/highstate.py", line 146, in output
    for host, hostdata in six.iteritems(data)
  File "/usr/lib/python2.7/site-packages/salt/output/highstate.py", line 251, in _format_host
    comps = tname.split('_|-')
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 27: ordinal not in range(128)
```

Running the same state file with `salt-ssh` or `salt-call --local` it does not fail.

```
[root@li432-250 ~]# cat /srv/salt/test.sls 
echo:
  cmd.run:
    - name: "echo 'This is Æ test!'"
[root@li432-250 ~]# salt-ssh \* state.sls test
ssh:
----------
          ID: echo
    Function: cmd.run
        Name: echo 'This is Æ test!'
      Result: True
     Comment: Command "echo 'This is Æ test!'" run
     Started: 21:12:22.057709
    Duration: 8.363 ms
     Changes:   
              ----------
              pid:
                  29401
              retcode:
                  0
              stderr:
              stdout:
                  This is Æ test!

Summary for ssh
------------
Succeeded: 1 (changed=1)
Failed:    0
------------
Total states run:     1
[...]

PR #46705, opened 2018-03-24
[2018.3.0rc1] Ensure event payloads are sent as Unicode

garethgreenaway:
### What does this PR do?
Ensure payloads for events are using Unicode.  Prior to this change, states that included Unicode characters would error out.

#46672 

### Previous Behavior
See issue #46672 for details

### New Behavior
Allow unicode/non-ascii characters in state file.

### Tests written?
Yes

### Commits signed with GPG?
Yes

### DIFF
--- a/tests/integration/cli/test_batch.py
+++ b/tests/integration/cli/test_batch.py
@@ -18,7 +18,7 @@ class BatchTest(ShellCase):
         '''
         Tests executing a simple batch command to help catch regressions
         '''
-        ret = 'Executing run on [\'sub_minion\']'
+        ret = 'Executing run on [u\'sub_minion\']'
 
         cmd = self.run_salt('\'*minion\' test.echo \'batch testing\' -b 50%')
         self.assertIn(ret, cmd)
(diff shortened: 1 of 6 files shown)


garethgreenaway:
re-run all

tests/integration/cli/test_batch.py
```diff
@@ -18,7 +18,7 @@ def test_batch_run(self):
         '''
         Tests executing a simple batch command to help catch regressions
         '''
-        ret = 'Executing run on [\'sub_minion\']'
+        ret = 'Executing run on [u\'sub_minion\']'
```
terminalmage: This fails on Python 3, since the `u` prefix is not used for unicode strings. The way I've gotten around this in other tests is to use `repr()`:

```python
ret = 'Executing run on [{0}]'.format(repr('sub_minion'))
```

This would apply to the other instances where lines have been changed below as well.

terminalmage (changes requested):

tests/integration/cli/test_batch.py
terminalmage: Fixed in 71a1d2b.

terminalmage (approved):

rallytime (approved):

PR #46707, opened 2018-03-24
Update service.py

absmith82:
add Cumulus to the disable list in def virtual()

### What does this PR do?
This adds Cumulus Linux into the to the disable list in def virtual() to allow it access to systemd.py

Salt forces Cumulus linux to init.d scripts and not systemd #46673 

### Previous Behavior
running service.* on Cumulus minion would force it to look up services in the init.d scripts

### New Behavior
running service.* on Cumulus minion will correctly use systemd to find services

### Tests written?

No

### Commits signed with GPG?

Yes/No

### DIFF
--- a/salt/modules/service.py
+++ b/salt/modules/service.py
@@ -32,6 +32,7 @@ def __virtual__():
         'Gentoo',
         'Ubuntu',
         'Debian',
+        'Cumulus',
         'Devuan',
         'Arch',
         'Arch ARM',


rallytime:
Hi @adam-smith-sundance and welcome!

Can you change the base branch of this PR to 2017.7 instead of 2017.7.5? We've already tagged 2017.7.5 and it is going through some QA processes right now. This will be included in 2017.7.6 if you change the base branch.

This is also the same fix as #46674, right? So in the future, if you submit to a bug-fix branch, those changes will be merged forward to develop so you don't need to submit 2 PRs. Just for your future information. :)

adam-smith-sundance:
Yes I will change this to the base branch.  Sorry about the develop PR, I
wasn't aware of the rules when I made the first PR.  I wasn't sure what the
process was for changing the PR from develop.  I figured there would be
more changes flagged from all the differences if I had submitted the change
from develop. So I just submitted a new PR.  I also noticed that 2018 is on
its way. Do I need to submit a PR for the 2018 and 2016 branches as well?

rallytime:
Hi @adam-smith-sundance - Not a problem at all! Thanks for switching the base branch.

Once this PR goes into 2017.7, it will be merged-forward into 2018.3 and develop. You don't need to submit any other PRs to any other branches for this to happen. :)

> I figured there would be more changes flagged from all the differences if I had submitted the change
from develop.

You're totally right about that, so submitting the new PR to the earlier branch is exactly right. 👍

gtmanfred (approved):

garethgreenaway:
Duplicate of https://github.com/saltstack/salt/pull/46547 which was merged into 2017.7.

rallytime:
@garethgreenaway Ah, you're right. Thanks!

@adam-smith-sundance Since this change has already been made in 2017.7, then I am going to close this PR. Apologies for the branch run-around on this. If I have closed this prematurely, please comment and we can always reopen. Thank you!

PR #46715, opened 2018-03-26
Salt Cloud EC2: Image Name

pakdel:
### What does this PR do?
Add the ability to use Amazon EC2 AMI name, instead of Image ID

### Previous Behavior
`image` had to be AMI ID in form of `ami-xxxxxxxx`

### New Behavior
`image` can be and AMI name, like `CentOS Linux 7 x86_64 HVM EBS 1801_01`, or a glod like `CentOS Linux 7 *`

### Tests written?

No

### Commits signed with GPG?

No


salt/cloud/clouds/ec2.py
```diff
@@ -1191,6 +1191,33 @@ def get_tenancy(vm_):
...
 
 
+def get_imageid(vm_):
+    '''
+    Returns the ImageId to use
+    '''
+    image = config.get_cloud_config_value(
+        'image', vm_, __opts__, search_global=False
+    )
+    if image.startswith('ami-'):
```
pakdel: Should we use a regex instead?

doc/topics/cloud/aws.rst
```diff
@@ -721,6 +721,24 @@ them have never been used, much less tested, by the Salt Stack team.
 .. __: https://aws.amazon.com/marketplace
 
 
+NOTE: If ``image`` of a profile does not start with ``ami-``, latest
+image with that name will be used.
```
cachedout: Could we add an example of this in the documentation just so it's extremely clear?

salt/cloud/clouds/ec2.py
cachedout: I'm not sure why we would need to.

doc/topics/cloud/aws.rst
cachedout: If we can get these requested docs added, then I think we can get this merged.

cachedout (approved):

pakdel:
Thank you @cachedout

PR #46720, opened 2018-03-26
Bump deprecation notices in nacl module & runner to Neon

rallytime:
These were initially marked for removal in "Fluorine", but we need to give users 2 feature releases to update. The version that the options will be removed in is "Neon". The related references have been updated, as well as the relevant documentation.

Refs #46426


cachedout (approved):

rallytime:
re-run py

Issue #46716, opened 2018-03-26
[2018.3.0] salt.states.x509 error on py3 setup


frogunder:
See this error when trying to generate key using x509 on a py3 setup.

```
An exception occurred in this state: Traceback (most recent call last):
  File "/root/salt/salt/state.py", line 1878, in call
    **cdata['kwargs'])
  File "/root/salt/salt/loader.py", line 1823, in wrapper
    return f(*args, **kwargs)
  File "/root/salt/salt/states/x509.py", line 311, in private_key_managed
    ret = __states__['file.managed'](**file_args)
  File "/root/salt/salt/loader.py", line 1823, in wrapper
    return f(*args, **kwargs)
  File "/root/salt/salt/states/file.py", line 2267, in managed
    isinstance(use_contents, six.string_types) and '\0' in use_contents
TypeError: 'str' does not support the buffer interface
```

### Setup
Install M2Crypto and this state file:
```
certificate.authority::private-key:
  x509.private_key_managed:
    - name: /root/ca.key
    - bits: 8192
    - backup: True
```

### Steps to Reproduce Issue
Run the state file.
Seems to work correctly on py2 but fails on py3 setup

py2:
```
li693-177.members.linode.com:
----------
          ID: certificate.authority::private-key
    Function: x509.private_key_managed
        Name: /root/ca.key
      Result: True
     Comment: File /root/ca.key updated
     Started: 17:19:03.942054
    Duration: 12414.535 ms
     Changes:   
              ----------
              new:
                  New private key generated

Summary for li693-177.members.linode.com
------------
Succeeded: 1 (changed=1)
Failed:    0
------------
Total states run:     1
Total run time:  12.415 s
```

py3:
```
          ID: certificate.authority::private-key
    Function: x509.private_key_managed
        Name: /root/ca.key
      Result: False
     Comment: An exception occurred in this state: Traceback (most recent call last):
                File "/root/salt/salt/state.py", line 1878, in call
                  **cdata['kwargs'])
                File "/root/salt/salt/loader.py", line 1823, in wrapper
                  return f(*args, **kwargs)
                File "/root/salt/salt/states/x509.py", line 311, in private_key_managed
                  ret = __states__['file.managed'](**file_args)
                File "/root/salt/salt/loader.py", line 1823, in wrapper
                  return f(*args, **kwargs)
                File "/root/salt/salt/states/file.py", line 2267, in managed
                  isinstance(use_contents, six.string_types) and '\0' in use_contents
[...]

PR #46723, opened 2018-03-26
[2018.3.0rc1] fixes to x509 module and state for PY3

garethgreenaway:
### What does this PR do?
A couple changes to ensure that when we're using the contents of keys that we're trying them as strings and not bytes.

#46716 

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/states/x509.py
+++ b/salt/states/x509.py
@@ -308,6 +308,9 @@ def private_key_managed(name,
         file_args['contents'] = __salt__['x509.create_private_key'](
             text=True, bits=bits, passphrase=passphrase, cipher=cipher, verbose=verbose)
 
+    # Ensure the key contents are a string before passing it along
+    file_args['contents'] = salt.utils.stringutils.to_str(file_args['contents'])
+
     ret = __states__['file.managed'](**file_args)
     if ret['changes'] and new_key:
         ret['changes'] = {'new': 'New private key generated'}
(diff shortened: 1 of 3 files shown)


cachedout:
@akissa Did you want to look at this? We usually have you take a peek at changes to this module. :]

akissa:
I do not have py3 at the moment so i will not be able to test.

rallytime:
Since this fixes a bug that @frogunder found, I have asked him to test this fix.

frogunder:
@rallytime working fine for me now with this fix, both on py2 and py3.

salt/states/x509.py
```diff
@@ -308,6 +308,9 @@ def private_key_managed(name,
         file_args['contents'] = __salt__['x509.create_private_key'](
             text=True, bits=bits, passphrase=passphrase, cipher=cipher, verbose=verbose)
 
+    # Ensure the key contents are a string before passing it along
+    file_args['contents'] = salt.utils.stringutils.to_str(file_args['contents'])
```
terminalmage: Is there a specific reason (like maybe x509 not working with unicode strings in PY2) that we're using `to_str()` here?

terminalmage (approved):
Just one question before merging, but otherwise this gets a :+1:

salt/states/x509.py
garethgreenaway: Exactly that. It's expecting it to be a string and not bytes.

terminalmage: :+1:

PR #46724, opened 2018-03-27
Reserve public IP block

edevenport:
### What does this PR do?

Provides a CLI function allowing IP blocks to be reserved. The public IPs can then be used in the cloud profile.

### Previous Behavior

IP blocks had to be reserved via the provider UI or other tools.

### New Behavior

IP blocks can be reserved from the `salt-cloud` CLI.

### Tests written?

No

### Commits signed with GPG?

No


salt/cloud/clouds/profitbricks.py
```diff
@@ -624,6 +626,40 @@ def list_nodes_full(conn=None, call=None):
     return ret
 
 
+def reserve_ipblock(call=None, kwargs=None):
+    '''
+
```
rallytime: Can you add docs here, please?

doc/topics/cloud/profitbricks.rst
```diff
@@ -151,6 +151,14 @@ command:
 
     # salt-cloud --list-sizes my-profitbricks-config
 
+.. versionadded:: Fluorine
+One or more public IP address can be reserved with the following command:
+command:
```
cachedout: duplicate `command`?

salt/cloud/clouds/profitbricks.py
```diff
@@ -120,7 +120,7 @@
     import profitbricks
     from profitbricks.client import (
         ProfitBricksService, Server,
-        NIC, Volume, FirewallRule,
+        NIC, Volume, FirewallRule, IPBlock,
```
cachedout: Is `IPBlock` a new addition to this upstream library or has it been present for a while? We want to make sure we preserve backward compatibility.

rallytime: Is `IPBlock` new to the latest release of profitbricks? The minimum version required is `3.1.0` according to the docs in this file. This import needs to be handled more elegantly if someone is using profitbricks with an older version.

rallytime (changes requested):

salt/cloud/clouds/profitbricks.py
denza: Doc comment added

doc/topics/cloud/profitbricks.rst
denza: Fixed

denza:
@rallytime @cachedout Could you please let us know if you require any further changes.

salt/cloud/clouds/profitbricks.py
```diff
@@ -626,6 +626,40 @@ def list_nodes_full(conn=None, call=None):
...
+    if kwargs.get('location') is None:
+        raise SaltCloudExecutionFailure('The "location" parameter is required')
+    location = kwargs.get('location')
+
+    size = 1
+    if kwargs.get('size') is not None:
+        size = kwargs.get('size')
+
+    block = conn.reserve_ipblock(IPBlock(size=size, location=location))
+    pprint.pprint(block)
```
rallytime: Can you remove this print statement? You can use logging if necessary, but we don't like to print directly this way.

rallytime (changes requested):
Hi @edevenport and @denza I just have one small request on this, otherwise this LGTM.

salt/cloud/clouds/profitbricks.py
denza: @rallytime The print statement removed.

rallytime (approved):

PR #46729, opened 2018-03-27
Performance improvement/error catching in expr_match

terminalmage:
`fnmatch.fnmatch()` does exact matches, so checking for an exact match is an unnecessary extra step.

Additionally, while regex errors are caught, TypeErrors are not, so the use of non-string expressions/values will result in a traceback. This prevents those tracebacks.

### DIFF
--- a/salt/utils/stringutils.py
+++ b/salt/utils/stringutils.py
@@ -294,15 +294,24 @@ def build_whitespace_split_regex(text):
 
 def expr_match(line, expr):
     '''
-    Evaluate a line of text against an expression. First try a full-string
-    match, next try globbing, and then try to match assuming expr is a regular
+    Evaluate a line of text against an expression. Tries to match expr first as
+    a glob using fnmatch.fnmatch(), and then tries to match expr as a regular
     expression. Originally designed to match minion IDs for
     whitelists/blacklists.
+
+    Note that this also does exact matches, as fnmatch.fnmatch() will return
+    ``True`` when no glob characters are used and the string is an exact match:
+
+    .. code-block:: python
+
+        >>> fnmatch.fnmatch('foo', 'foo')
+        True
     '''
-    if line == expr:
-        return True
-    if fnmatch.fnmatch(line, expr):
-        return True
+    try:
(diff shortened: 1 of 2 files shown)


terminalmage:
@saltstack/team-core actually I am undecided on the `TypeError` catching. Is this something we want to cause the `expr_match()` to silently fail? I feel like the `TypeError` may actually be useful in that it will make issues with the value and expresssion being the wrong type more apparent, instead of just silently returning `False` and requiring one to perhaps resort to stepping through in a debugger to discover that what was passed to `expr_match()` was an int instead of a str.

rallytime:
@terminalmage For your question about the `TypeError` catching, I can see what you mean. There's pros/cons to each approach. I am leaning towards leaving the `TypeError` catching in (as that is what we're doing on `re` errors as well) and then maybe logging something to `debug`. What do you think?

terminalmage:
If we pass bad data to it, something is very wrong, so a noisy log message is OK. I went ahead and did log.exception (which logs at ERROR and adds the traceback to the log message.

I also removed inaccurate logging from `check_whitelist_blacklist()`. The TypeError that would have been raised before would not have meant that the whitelist/blacklist is not iterable, and we're no longer raising the TypeError anyway.

The other refs to `expr_match()` in the codebase do not have any TypeError catching, so this should help with troubleshooting if any non-string data makes it into those areas of the code as well.

salt/utils/stringutils.py
```diff
@@ -292,22 +292,31 @@ def __build_parts(text):
     return r'(?m)^{0}$'.format(regex)
 
 
-def expr_match(line, expr):
+def expr_match(val, expr):
```
rallytime: What's the reasoning behind changing this variable name? This is a public facing function so I don't think we should change it unnecessarily.

rallytime (approved):

rallytime:
re-run py

rallytime:
re-run py

cachedout (approved):

rallytime:
re-run py2

PR #46743, opened 2018-03-27
Workaround getpwnam in auth test for MacOSX

Ch3LL:
### What does this PR do?
As explained in comment https://github.com/saltstack/salt-jenkins/issues/504#issuecomment-328961942 there is an issue on macosx when running `pwd.getpwnam` in the same python instance if the user is deleted it still detects the user. 
Fixes https://github.com/saltstack/salt-jenkins/issues/504


rallytime (approved):
Nice investigation on this one.

Issue #46668, opened 2018-03-22
Jinja2 filter strftime stopped working in salt-ssh 2018.3


anlutro:
I'm using salt-ssh from a virtualenv and a checked out git repository. It worked fine using 2017.7, but 2018.3 fails because at least one Jinja filter no longer seems to be present:

```
$ salt-ssh odin state.apply
[ERROR   ] Can't find varstack master_top
[ERROR   ] Rendering exception occurred
Traceback (most recent call last):
  File "/home/andreas/code/contrib/salt/salt/utils/templates.py", line 378, in render_jinja_tmpl
    template = jinja_env.from_string(tmplstr)
  File "/home/andreas/code/contrib/salt/.venv/lib/python3.5/site-packages/jinja2/environment.py", line 880, in from_string
    return cls.from_code(self, self.compile(source), globals, None)
  File "/home/andreas/code/contrib/salt/.venv/lib/python3.5/site-packages/jinja2/environment.py", line 591, in compile
    self.handle_exception(exc_info, source_hint=source_hint)
  File "/home/andreas/code/contrib/salt/.venv/lib/python3.5/site-packages/jinja2/environment.py", line 780, in handle_exception
    reraise(exc_type, exc_value, tb)
  File "/home/andreas/code/contrib/salt/.venv/lib/python3.5/site-packages/jinja2/_compat.py", line 37, in reraise
    raise value.with_traceback(tb)
  File "<unknown>", line 4, in template
  File "/home/andreas/code/contrib/salt/.venv/lib/python3.5/site-packages/jinja2/environment.py", line 543, in _generate
    optimized=self.optimized)
  File "/home/andreas/code/contrib/salt/.venv/lib/python3.5/site-packages/jinja2/compiler.py", line 82, in generate
    generator.visit(node)
  File "/home/andreas/code/contrib/salt/.venv/lib/python3.5/site-packages/jinja2/visitor.py", line 38, in visit
    return f(node, *args, **kwargs)
  File "/home/andreas/code/contrib/salt/.venv/lib/python3.5/site-packages/jinja2/compiler.py", line 754, in visit_Template
    self.blockvisit(node.body, frame)
  File "/home/andreas/code/contrib/salt/.venv/lib/python3.5/site-packages/jinja2/compiler.py", line 378, in blockvisit
    self.visit(node, frame)
  File "/home/andreas/code/contrib/salt/.venv/lib/python3.5/site-packages/jinja2/visitor.py", line 38, in visit
[...]

PR #46744, opened 2018-03-27
[2018.3] Ensure salt.utils.dateutils is available for templates via salt-ssh

garethgreenaway:
### What does this PR do?
Including salt.utils.dateutils so various jinja_filters are available when using salt-ssh.

#46668 

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/utils/templates.py
+++ b/salt/utils/templates.py
@@ -28,6 +28,7 @@ else:
 
 # Import Salt libs
 import salt.utils.data
+import salt.utils.dateutils
 import salt.utils.http
 import salt.utils.files
 import salt.utils.platform


terminalmage (approved):
This is a good catch. I wonder which other jinja filters may have this issue.

rallytime:
re-run py

rallytime:
re-run py

lorengordon:
Ohhhh, with the utils modules getting reorganized, the decorator approach might be insufficient as is.

* [decorators](https://github.com/saltstack/salt/blob/develop/salt/utils/decorators/jinja.py)
* [jinja env updates](https://github.com/saltstack/salt/blob/develop/salt/utils/templates.py#L359-L361)

May want instead to import all the functions providing jinja filters/tests/globals in a central place, and create the mapping there...? A little less intuitive for a new dev when adding a new jinja filter/test/global, but more sustainable over time...

PR #46748, opened 2018-03-28
Add serializer options passing

nocturo:
### What does this PR do?
Addresses a problem with file.serialize state which cannot pass additional options over to lower name.serializer module which can actually understand what's being passed.

### New Behavior
Adding to new `serializer_opts` state config list will pass down dicts to lower modules.

### Tests written?

No

### Commits signed with GPG?

Yes


terminalmage (changes requested):
This is an excellent contribution! There are a few changes we need to get it across the finish line though, I opened a PR against your fork: https://github.com/nocturo/salt/pull/1

If you could review that, and then merge it and push, I think this PR will be ready to merge.

Thanks again!

nocturo:
Awesome, thanks for the comments on the PR!

I've merged the PR and it looks to be automatically pushed here too.

terminalmage (approved):
Looks good, the test failures do not appear to be related.

Issue #46660, opened 2018-03-22
top file merging same does produce conflicting ids with gitfs


mruepp:
### Description of Issue/Question
We use gitfs as backend. We have three environments: dev-qa-staging
We use top file merging strategy same. Our Top files for every environment contains {{ saltenv }}:

We want to create and test states by branching away for instance from dev -> feature/branch, add changes, test them by using salt .. state.apply saltenv=feature/branch and after successfully highstating the changes, Pull Request and Merge.

So everything is fine except, as long as the feature branch exists, we are no longer able to highstate ALL environments at the same time without using saltenv. This will result in conflicting ids.

Also we are not able to move states from dev to qa for instance by checking out specific states or cherrypicking them from e.g. dev to qa to stage to prod because we will always having conflicting ids.

So I would expect with top file merging "same" for the environments to only account for the own files.

Do I miss something?

### Setup
Our gitfs config: 

```
fileserver_backend:
  - git

gitfs_provider: pygit2

gitfs_user: salt
gitfs_password: password

top_file_merging_strategy: same

default_top: base

gitfs_remotes:
  - https:/gitrepo/scm/devops/salt.git:
      - refspecs:
          - '+refs/heads/*:refs/remotes/origin/*'
          - '+refs/tags/*:refs/tags/*'
          - '+refs/pull/*/head:refs/remotes/origin/pr/*'
          - '+refs/pull/*/merge:refs/remotes/origin/merge/*'
      - root: states
      - base: master
      - saltenv:
          - dev:
              - ref: dev
          - qa:     
              - ref: qa
          - stage:
              - ref: stage

  #List the formula repos with respective base
  - https://gitrepo/scm/devops/salt-formulas.git:
      - name: vault-formula
      - root: vault-formula
      - saltenv:
          - dev:
              - ref: master

  - https://gitrepo/scm/devops/salt-formulas.git:
      - name: consul-formula
      - root: consul-formula
      - saltenv:
          - dev:
              - ref: master
```

### Steps to Reproduce Issue
Using Salt config as posted and try to create a feature branch to implement a change

### Versions Report
`Salt Version:
           Salt: 2017.7.4

Dependency Versions:
           cffi: 1.6.0
       cherrypy: Not Installed
       dateutil: Not Installed
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.7.2
        libgit2: 0.24.6
        libnacl: Not Installed
[...]

PR #46751, opened 2018-03-28
top file merging strategy 'same' works again

folti:
### What does this PR do?
Fix the 'top_file_merging_strategy: same' behavior, broken since 82bf7fe5e

See Previous Behavior

#46660 

### Previous Behavior
In case there are multiple environments with same or very similar file hierarchy, and merging
strategy set to 'same', running highstate without adding the saltenv=base (or whatever set as
default_top) argument caused merge conflicts, as get_tops() actually worked in merge mode, 
when collecting the environments, and thus highstate runs have been failed.

### New Behavior
Same as with 2016.3, the previous major version.

Running highstate without a saltenv=&lt;env> argument, would result in using the environment set in the 'default_top' config option.

### Tests written?
No

### Commits signed with GPG?
No

### DIFF
--- a/salt/state.py
+++ b/salt/state.py
@@ -2817,6 +2817,7 @@ class BaseHighState(object):
                     'top_file_merging_strategy set to \'same\', but no '
                     'default_top configuration option was set'
                 )
+            self.opts['environment'] = self.opts['default_top']
 
         if self.opts['environment']:
             contents = self.client.cache_file(


terminalmage (approved):

rallytime:
re-run py

terminalmage:
This is actually an incorrect fix and needs to be reverted.

PR #46766, opened 2018-03-28
Change the way we're cleaning up after some tests

twangboy:
### What does this PR do?
Uses the `@with_tempfile` decorator instead of `try/finally` blocks to clean up temp files.

Sometimes the `os.unlink` function was failing to clean up, causing the test to fail. The decorator cleans up and fails gracefully, without nullifying the results of the test.

### Tests written?
Yes

### Commits signed with GPG?
Yes

### DIFF
--- a/tests/integration/modules/test_cp.py
+++ b/tests/integration/modules/test_cp.py
@@ -4,7 +4,6 @@
 from __future__ import absolute_import
 import os
 import uuid
-import shutil
 import hashlib
 import logging
 import psutil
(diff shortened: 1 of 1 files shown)


terminalmage (changes requested):
To add a little more background, the os.unlink in the try/finally was raising an Error 2 (i.e. file doesn't exist). Since several of the tests were using the same destination path, this may have been caused by a race condition in which the unlink triggered by another test's `finally` clause deleted the file in the middle of a subsequent test, causing it to be gone already by the end of that test.

@twangboy Just one minor change here.

terminalmage (approved):

Issue #46433, opened 2018-03-08
Creating Directories takes a long time on an AD Network


mxrss2:
### Description of Issue/Question

I am running a salt pilot and its taking forever to create a directory (90secs+) when on a AD Domain. I am attaching the logs for this and the configuration. This is set up in a masterless setup.

I am running on the following.

```
Salt Version:
           Salt: 2017.7.4

Dependency Versions:
           cffi: 1.10.0
       cherrypy: unknown
       dateutil: 2.6.0
      docker-py: Not Installed
          gitdb: 2.0.3
      gitpython: 2.1.3
          ioflo: Not Installed
         Jinja2: 2.9.6
        libgit2: Not Installed
        libnacl: 1.6.1
       M2Crypto: Not Installed
           Mako: 1.0.6
   msgpack-pure: Not Installed
 msgpack-python: 0.4.8
   mysql-python: Not Installed
      pycparser: 2.17
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: Not Installed
         Python: 2.7.13 (v2.7.13:a06454b1afa1, Dec 17 2016, 20:5
0 64 bit (AMD64)]
   python-gnupg: 0.4.0
         PyYAML: 3.12
          PyZMQ: 16.0.2
           RAET: Not Installed
          smmap: 2.0.3
        timelib: 0.2.4
        Tornado: 4.5.1
            ZMQ: 4.1.6

System Versions:
           dist:
         locale: cp1252
        machine: AMD64
        release: 2012ServerR2
         system: Windows
        version: 2012ServerR2 6.3.9600  Multiprocessor Free

```

```
[TRACE   ] Loading core.ip_interfaces grain
[TRACE   ] Loading core.linux_distribution grain
[TRACE   ] Loading core.locale_info grain
[TRACE   ] Loading core.os_data grain
[TRACE   ] Loading core.path grain
[TRACE   ] Loading core.pythonexecutable grain
[TRACE   ] Loading core.pythonpath grain
[TRACE   ] Loading core.pythonversion grain
[TRACE   ] Loading core.saltpath grain
[TRACE   ] Loading core.saltversion grain
[TRACE   ] Loading core.saltversioninfo grain
[TRACE   ] Loading core.zmqversion grain
[TRACE   ] Loading disks.disks grain
[TRACE   ] Disk grain does not support OS
[TRACE   ] Loading extra.config grain
[TRACE   ] Loading extra.shell grain
[...]

PR #46770, opened 2018-03-28
Change the order of SID Lookup

twangboy:
### What does this PR do?
Changes the order in which `win_dacl.get_name` obtains a PySID object.
- Check for a PySID object first (least expensive)
- Get PySID object from String SID next
- Get PySID object from principal name last (most expensive)

This function is used by the `_ace_to_dict` function to match SIDs in an ACL with usernames. The function already converts the SID to a PySID object, so there's no need to go through all the other steps.

https://github.com/saltstack/salt/issues/46433

### Previous Behavior
Even if a PySID object was passed to the function, it would still try lookup by name and by string sid before returning the name. This was greatly slowing down file state functions.

### New Behavior
Checks least expensive first.

### Tests written?
No

### Commits signed with GPG?
Yes


rallytime:
@twangboy lint: https://jenkins.saltstack.com/job/PR/job/salt-pr-lint-n/20527/violations/file/salt/utils/win_dacl.py/

salt/utils/win_dacl.py
```diff
@@ -1115,34 +1121,37 @@ def get_name(principal):
...
+        # String SID
+        # Try Converting String SID to SID Object first as it's least expensive
+        try:
+            sid_obj = win32security.ConvertStringSidToSid(principal)
+        except pywintypes.error:
+
+            # Name
+            # Try Getting the SID Object by Name Lookup last
+            # This is expensive, especially on large AD Domains
+            try:
```
skizunov: Right after the `except` statement, the blank line and `# Name` probably need to be removed (or `# Name` elaborated on).

salt/utils/win_dacl.py
```diff
@@ -1115,34 +1121,37 @@ def get_name(principal):
...
+                sid_obj = win32security.LookupAccountName(None, principal)[0]
+            except pywintypes.error:
+                # Just pass the object and let the LookupAccountSid function try
```
skizunov: Shouldn't this read `# Just pass the string` instead of `# Just pass the object`?

salt/utils/win_dacl.py
```diff
@@ -1115,34 +1121,37 @@ def get_name(principal):
...
+                sid_obj = principal
 
+    # By now we should have a valid PySID object
     try:
         return win32security.LookupAccountSid(None, sid_obj)[0]
     except (pywintypes.error, TypeError) as exc:
+        message = 'Error resolving "{0}"'.format(principal)
         if type(exc) == pywintypes.error:
             win_error = win32api.FormatMessage(exc.winerror).rstrip('\n')
```
skizunov: Shouldn't this be `message =` instead of `win_error =` (since `win_error` is not used after this line). If so, then you can move the `message` that is currently above the `if` into the `else` (unless you want to use it in this `if` case in addition to the message from `FormatMessage`).

twangboy: It's used on the following line to add more detail to the error message if it is a `pywintypes.error`

skizunov: I see what you are saying. I got confused because the `else` case was red (removed), not green (added). So there is no `else` case at all anymore. Ignore that comment.

skizunov (approved):

rallytime (approved):
Nice

rallytime:
re-run py

cachedout (approved):

PR #46772, opened 2018-03-29
fix container removal if auto_remove was enabled

bmiguel-teixeira:
### What does this PR do?
Corrects a bug in the container stop & removal flow. It the container was started with the AutoRemove flag  (--rm), when stopped, the docker engine will automatically remove it, and salt's rm function will throw a 404 when removing.

Did not open an issue.

### Previous Behavior
`salt myminion docker.rm mycontainer stop=True`
This would throw 404 if `mycontainer` had the AutoRemove flag enabled as stopping the container would imply auto removal.

### New Behavior
`salt myminion docker.rm mycontainer stop=True`
If `stop=True`, we check the container's AutoRemove flag and do not remove the container if the flag is set to True. The docker engine will take care of that.

No longer throws 404 if `mycontainer` had the AutoRemove flag enabled.

### Tests written?

No

### Commits signed with GPG?

No

### DIFF
--- a/salt/modules/dockermod.py
+++ b/salt/modules/dockermod.py
@@ -3798,6 +3798,7 @@ def rm_(name, force=False, volumes=False, **kwargs):
     kwargs = __utils__['args.clean_kwargs'](**kwargs)
     stop_ = kwargs.pop('stop', False)
     timeout = kwargs.pop('timeout', None)
+    auto_remove = False
     if kwargs:
         __utils__['args.invalid_kwargs'](kwargs)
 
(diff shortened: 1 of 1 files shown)


salt/modules/dockermod.py
```diff
@@ -3807,9 +3808,12 @@ def rm_(name, force=False, volumes=False, **kwargs):
             'remove this container'.format(name)
         )
     if stop_ and not force:
+        auto_remove = inspect_container(name)['HostConfig']['AutoRemove']
```
terminalmage: To future-proof us against Docker changing inspect return data structure, we should catch on `KeyError` here:

```python
        inspect_results = inspect_container(name)
        try:
            auto_remove = inspect_results['HostConfig']['AutoRemove']
        except KeyError:
            log.error(
                'Failed to find AutoRemove in inspect results, Docker API may '
                'have changed. Full results: %s', inspect_results
            )
            # Assume False and attempt to remove the container below
            auto_remove = False
```

terminalmage (changes requested):
@rallytime once this is merged we'll need a backport to 2017.7.

salt/modules/dockermod.py
bmiguel-teixeira: agreed with eafp

terminalmage: Good point, I forgot you had set it above. :+1:

terminalmage (approved):

PR #46774, opened 2018-03-29
Add opkg/NILinuxRT 'reboot required' witnessing feature

10ne1:
### What does this PR do?

Add  new reboot required witnessed logic to opkg operations on NILRT systems similar to how win_system.py does it, using the same system API as on windows.

Doesn't have an issue assigned.

### New Behavior

A new Linux system API is added for getting/setting reboot witness events which is called by the opkg module. The witness events are stored in volatile memory such that they're cleared upon reboot.

### Tests written?

No

### Commits signed with GPG?

No

### DIFF
--- a/salt/modules/opkg.py
+++ b/salt/modules/opkg.py
@@ -22,6 +22,7 @@ import copy
 import os
 import re
 import logging
+import errno
 
 # Import salt libs
 import salt.utils.args
@@ -55,12 +56,68 @@ log = logging.getLogger(__name__)
 # Define the module's virtual name
 __virtualname__ = 'pkg'
 
+NILRT_MODULE_STATE_PATH = '/var/lib/salt/kernel_module_state'
+
+
+def _update_nilrt_module_dep_info():
+    '''
+    Update modules.dep timestamp & checksum.
+
+    NILRT systems determine whether to reboot after kernel module install/
+    removals/etc by checking modules.dep which gets modified/touched by
+    each kernel module on-target dkms-like compilation. This function
+    updates the module.dep data after each opkg kernel module operation
+    which needs a reboot as detected by the salt checkrestart module.
+    '''
+    __salt__['cmd.shell']('stat -c %Y /lib/modules/$(uname -r)/modules.dep >{0}/modules.dep.timestamp'
+                          .format(NILRT_MODULE_STATE_PATH))
+    __salt__['cmd.shell']('md5sum /lib/modules/$(uname -r)/modules.dep >{0}/modules.dep.md5sum'
+                          .format(NILRT_MODULE_STATE_PATH))
+
+
+def _get_restartcheck_result(errors):
+    '''
+    Return restartcheck result and append errors (if any) to ``errors``
+    '''
+    rs_result = __salt__['restartcheck.restartcheck'](verbose=False)
+    if isinstance(rs_result, dict) and 'comment' in rs_result:
+        errors.append(rs_result['comment'])
+    return rs_result
+
+
+def _process_restartcheck_result(rs_result):
+    '''
+    Check restartcheck output to see if system/service restarts were requested
+    and take appropriate action.
+    '''
+    if 'No packages seem to need to be restarted' in rs_result:
+        return
+    for rstr in rs_result:
+        if 'System restart required' in rstr:
+            _update_nilrt_module_dep_info()
+            __salt__['system.set_reboot_required_witnessed']()
+        else:
+            service = os.path.join('/etc/init.d', rstr)
+            if os.path.exists(service):
+                __salt__['cmd.run']([service, 'restart'])
+
 
 def __virtual__():
     '''
     Confirm this module is on a nilrt based system
     '''
-    if __grains__.get('os_family', False) == 'NILinuxRT':
+    if __grains__.get('os_family') == 'NILinuxRT':
+        try:
+            os.makedirs(NILRT_MODULE_STATE_PATH)
+        except OSError as exc:
+            if exc.errno != errno.EEXIST:
+                return False, 'Error creating {0} (-{1}): {2}'.format(
+                    NILRT_MODULE_STATE_PATH,
+                    exc.errno,
+                    exc.strerror)
+        if not (os.path.exists(os.path.join(NILRT_MODULE_STATE_PATH, 'modules.dep.timestamp')) and
+                os.path.exists(os.path.join(NILRT_MODULE_STATE_PATH, 'modules.dep.md5sum'))):
+            _update_nilrt_module_dep_info()
         return __virtualname__
     return (False, "Module opkg only works on nilrt based systems")
 
@@ -412,11 +469,15 @@ def install(name=None,
             ret.update({pkgname: {'old': old.get(pkgname, ''),
                                   'new': new.get(pkgname, '')}})
 
+    rs_result = _get_restartcheck_result(errors)
+
     if errors:
         raise CommandExecutionError(
             'Problem encountered installing package(s)',
             info={'errors': errors, 'changes': ret}
         )
+
+    _process_restartcheck_result(rs_result)
 
     return ret
 
@@ -475,11 +536,15 @@ def remove(name=None, pkgs=None, **kwargs):  # pylint: disable=unused-argument
     new = list_pkgs()
     ret = salt.utils.data.compare_dicts(old, new)
 
+    rs_result = _get_restartcheck_result(errors)
+
     if errors:
         raise CommandExecutionError(
             'Problem encountered removing package(s)',
             info={'errors': errors, 'changes': ret}
         )
+
+    _process_restartcheck_result(rs_result)
 
     return ret
 
@@ -536,6 +601,8 @@ def upgrade(refresh=True, **kwargs):  # pylint: disable=unused-argument
            'comment': '',
            }
 
+    errors = []
+
     if salt.utils.data.is_true(refresh):
         refresh_db()
 
@@ -550,10 +617,17 @@ def upgrade(refresh=True, **kwargs):  # pylint: disable=unused-argument
     ret = salt.utils.data.compare_dicts(old, new)
 
     if result['retcode'] != 0:
+        errors.append(result)
+
+    rs_result = _get_restartcheck_result(errors)
+
+    if errors:
         raise CommandExecutionError(
             'Problem encountered upgrading packages',
-            info={'changes': ret, 'result': result}
+            info={'errors': errors, 'changes': ret}
         )
+
+    _process_restartcheck_result(rs_result)
 
     return ret
 
--- a/salt/modules/restartcheck.py
+++ b/salt/modules/restartcheck.py
@@ -547,10 +547,10 @@ def restartcheck(ignorelist=None, blacklist=None, excludepid=None, **kwargs):
             service = __salt__['service.available'](packages[package]['process_name'])
 
             if service:
-                packages[package]['systemdservice'].append(packages[package]['process_name'])
-            else:
                 if os.path.exists('/etc/init.d/' + packages[package]['process_name']):
                     packages[package]['initscripts'].append(packages[package]['process_name'])
+                else:
+                    packages[package]['systemdservice'].append(packages[package]['process_name'])
 
     restartable = []
     nonrestartable = []
--- a/salt/modules/restartcheck.py
+++ b/salt/modules/restartcheck.py
@@ -358,9 +358,9 @@ def restartcheck(ignorelist=None, blacklist=None, excludepid=None, **kwargs):
         timeout: int, timeout in minute
 
     Returns:
-        True if no packages for restart found.
-        False on failure.
-        String with checkrestart output if some package seems to need to be restarted.
+        Dict on error: { 'result': False, 'comment': '<reason>' }
+        String with checkrestart output if some package seems to need to be restarted or
+        if no packages need restarting.
 
     .. versionadded:: 2015.8.3
 
--- a/salt/modules/restartcheck.py
+++ b/salt/modules/restartcheck.py
@@ -345,6 +345,36 @@ def _check_timeout(start_time, timeout):
         raise salt.exceptions.TimeoutError('Timeout expired.')
 
 
+def _kernel_modules_changed_nilrt(kernelversion):
+    '''
+    Once a NILRT kernel module is inserted, it can't be rmmod so systems need
+    rebooting (some modules explicitely ask for reboots even on first install),
+    hence this functionality of determining if the module state got modified by
+    testing if depmod was run.
+
+    Returns:
+             - True/False depending if modules.dep got modified/touched
+    '''
+    depmodpath_base = '/lib/modules/{0}/modules.dep'.format(kernelversion)
+    depmodpath_timestamp = "/var/lib/salt/kernel_module_state/modules.dep.timestamp"
+    depmodpath_md5sum = "/var/lib/salt/kernel_module_state/modules.dep.md5sum"
+
+    # nothing can be detected without these dependencies
+    if (kernelversion is None or
+            not os.path.exists(depmodpath_timestamp) or
+            not os.path.exists(depmodpath_md5sum)):
+        return False
+
+    prev_timestamp = __salt__['file.read'](depmodpath_timestamp).rstrip()
+    # need timestamp in seconds so floor it using int()
+    cur_timestamp = str(int(os.path.getmtime(depmodpath_base)))
+
+    if prev_timestamp != cur_timestamp:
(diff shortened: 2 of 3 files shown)


skizunov (approved):
There are a few instances of inconsistent use of single-quotes (') vs doublequotes ("), but I don't know if that qualifies as an issue. It looks fine overall.

salt/modules/opkg.py
```diff
@@ -412,12 +453,19 @@ def install(name=None,
             ret.update({pkgname: {'old': old.get(pkgname, ''),
                                   'new': new.get(pkgname, '')}})
 
+    rs_result = __salt__['restartcheck.restartcheck'](verbose=False)
+    if isinstance(rs_result, dict) and 'comment' in rs_result:
```
cachedout: We have this duplicated over on L527. Could we refactor it to a single block that's called?

10ne1: Done.

salt/modules/system.py
```diff
@@ -608,3 +608,54 @@ def get_computer_name():
...
+    the event gets cleared upon reboot.
+
+    Returns:
+        bool: ``True`` if successful, otherwise ``False``
+
+    .. code-block:: bash
+
+        salt '*' system.set_reboot_required_witnessed
+    '''
+    if __grains__.get('os_family') == 'NILinuxRT':
```
cachedout: Hrm. I don't like that this function is essentially a no-op for everything _but_ the `NILinuxRT` OS family but exists in the `system` module which is available and consumed by all types of systems. Is this the best place for this?

salt/modules/system.py
```diff
@@ -608,3 +608,54 @@ def get_computer_name():
...
+    Returns:
+        bool: ``True`` if the a reboot request was witnessed, ``False`` otherwise
+
+    CLI Example:
+
+    .. code-block:: bash
+
+        salt '*' system.get_reboot_required_witnessed
+    '''
+    if __grains__.get('os_family') == 'NILinuxRT':
```
cachedout: Same concern as above here.

salt/modules/system.py
skizunov: @cachedout : I think the idea behind this was to copy from the architecture already put in place in `salt/modules/win_system.py`, which is also a "system" module (by virtual name). So the function names and location are not new. I believe this exact same feature could apply to other Linux distributions, I would assume that the author was not motivated to implement and test it on other distributions. Although, by looking at things, it seems that it wouldn't be difficult to do so.

Perhaps the author should add at least one more distribution (eg CentOS) so that your automated tests would cover this new feature?

10ne1: I can move it somewhere else, no problem, we just need to figure out where is the best place to put them. How about the opkg module? It was written for the NILinuxRT family and it's the only place which calls these witness functions.

terminalmage: I think that opkg.py is probably the correct place to put it. I don't see any code here that actually calls these new functions, and since they only apply to NILinuxRT they shouldn't be in a more general module.

skizunov: @terminalmage : I disagree, because I think these new functions are not private and really meant to be called at the job level. Eg `system.get_reboot_required_witnessed`. How else would the master find out if a reboot is required on a given minion? Because they are required to be called at the job level, putting them into a module with virtual name `pkg` doesn't make that much sense, because all modules with virtual name `pkg` should adhere to a predefined set of APIs, which this falls outside of.

Since there already is precedent for `system.get_reboot_required_witnessed` from the Windows side, it seems like the logical place to put these functions where they are right now. This feature can easily be extended to other Linux distributions, because the main logic that checks if a reboot is required is in `salt/modules/restartcheck.py`, which already supports Debian and Red Hat based systems.

The `pkg` modules for such systems could be modified in a similar way to check if a restart is required after a package operation, and invoke `system.set_reboot_required_witnessed` in such a case. Then, for such systems, a user has the option of checking if a reboot is required after a package operation (eg if a new kernel is installed). It would actually be a pretty useful feature for general use.

10ne1: I'm ok either way, I think we just need a consensus so I know what to do further.

terminalmage: @skizunov Thanks, I overlooked your earlier comment regarding the `win_system` module. Sorry about that! I agree, system.py is a good place for it.

cachedout (approved):

cachedout:

terminalmage (approved):
@cachedout sorry, my comment [here](https://github.com/saltstack/salt/pull/46774#discussion_r179514086) was meant to signify that my concerns were satisfied, but I neglected to submit an "Approved" review. Sorry for holding things up.

isbm (changes requested):

salt/modules/system.py
```diff
@@ -608,3 +608,54 @@ def get_computer_name():
...
+        if os.path.exists(NILRT_REBOOT_WITNESS_PATH):
+            return True
+    return False
```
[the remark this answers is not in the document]
terminalmage: Actually, I like the idea of such decorators even if we don't use them here, so I am looking into writing them now.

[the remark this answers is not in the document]
isbm: @terminalmage :+1: for the decorator. I would still map it separately into its own namespace. Because it is "special" stuff. Then in your state you can better segregate these calls, like:

```
salt * somemodule.special.do_something
```
Then you will also construct your states/commands differently, avoiding hitting systems that aren't in this space.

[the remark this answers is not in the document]
isbm: @skizunov are we on the same page? Sounds like you're talking about an interface. And it is yet to be implemented. What I am talking is a principal of an actual implementation of _any_ module that on some OSes even do not have to exist. That said, if other systems needs to implement this, then exception raise "not yet implemented" or "cannot be at all supported" is needed here.

But right now it is returning one of an identical return data of the actual implementation. And so this is a bug and confusion.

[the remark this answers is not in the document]
skizunov: @isbm : This interface has already been implemented for Windows systems (in `win_system.py` which also implements the `system` virtual interface). But I understand your point about the behavior on other Linux distributions that may not necessarily implement the interface. So perhaps there are a few options:
1) Use the solution that @terminalmage suggests (requires infrastructure not yet present)
2) Just implement for the other Linux distributions (should be very easy, just need to confirm the path to volatile storage is valid for each Linux distribution)
3) On other Linux distributions that don't (yet) support this, throw an exception instead of returning a value.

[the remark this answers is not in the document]
isbm: @skizunov I'd prefer @terminalmage 's suggestion as it is case-agnostic.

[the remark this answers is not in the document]
terminalmage: Implementing function namespaces is not going to happen right now. It may in a future release, but it's extremely unlikely that it will happen in time for Fluorine. So, we should adjust expectations here accordingly.

Late last week I added the needed support to the `depends` decorator to enable the running of a shell command (https://github.com/saltstack/salt/pull/47040). In the process of doing so, I discovered that this decorator _already_ supports interpreting an expression that evaluates to a boolean (with `True` meaning that the function should be loaded and `False` meaning that it should not).

There are two approaches we could use:

**OPTION 1 (depends on https://github.com/saltstack/salt/pull/47040, would have to wait for that PR to be merged and then rebase to gain access to this functonality)**

```python
from salt.utils.decorators import depends

@depends('some shell command', retcode=0)
def set_reboot_required_witnessed():
    '''
```

**OPTION 2**

```python
from salt.utils.decorators import depends

@depends(some python expression)
def set_reboot_required_witnessed():
    '''
```

Option 2 could be done right now without waiting for my decorator modifications to be merged, but note that whatever is passed to the depends decorator must evaluate as boolean. So it has to be an expression or function that returns a bool.

The question is, I guess, what would be the best way to confirm that the minion is NILinuxRT? Keep in mind that we won't have access to the grains.

cachedout:
I think we're still waiting on some feedback for some of the concerns raised by @isbm here.

isbm:
@10ne1 any updates here?

10ne1:
Sorry for the delay, I've been overwhelmed with other work + conference travel, I'll get back to this ASAP (I hope to update the PR tomorrow).

salt/modules/opkg.py
```diff
@@ -55,12 +56,62 @@
...
 def __virtual__():
     '''
     Confirm this module is on a nilrt based system
     '''
-    if __grains__.get('os_family', False) == 'NILinuxRT':
+    if __grains__.get('os_family') == 'NILinuxRT':
+        if not (os.path.exists(os.path.join(NILRT_MODULE_STATE_PATH, 'modules.dep.timestamp')) and
+                os.path.exists(os.path.join(NILRT_MODULE_STATE_PATH, 'modules.dep.md5sum'))):
+            try:
+                os.makedirs(NILRT_MODULE_STATE_PATH)
```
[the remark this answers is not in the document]
10ne1: Done.

salt/modules/opkg.py
```diff
@@ -55,12 +56,62 @@
...
+            except OSError as exc:
+                if exc.errno != errno.EEXIST:
+                    msg = 'Error creating {0} (-{1}): {2}'.format(NILRT_MODULE_STATE_PATH, exc.errno, exc.strerror)
+                    return (False, msg)
```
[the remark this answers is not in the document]
10ne1: Done.

salt/modules/restartcheck.py
```diff
@@ -298,6 +298,13 @@ def _kernel_versions_redhat():
     return kernel_versions
 
 
+def _is_older_nilrt():
+    '''
+    If this is an older version of NILinuxRT, return True. Otherwise, return False.
+    '''
+    return bool(os.path.exists('/usr/local/natinst/bin/nisafemodeversion'))
```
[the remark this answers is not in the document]
10ne1: Done.

salt/modules/restartcheck.py
```diff
@@ -307,14 +314,61 @@ def _kernel_versions_nilrt():
             as they are probably interpreted in output of `uname -a` command.
     '''
     kernel_versions = []
-    kernel = os.readlink('/boot/bzImage')
-    kernel = os.path.basename(kernel)
-    kernel = kernel.strip('bzImage-')
+
+    if __grains__.get('os_family') == 'NILinuxRT' and _is_older_nilrt():
```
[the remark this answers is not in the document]
10ne1: Done.

salt/modules/restartcheck.py
```diff
@@ -307,14 +314,61 @@ def _kernel_versions_nilrt():
...
+        # bzImage is copied in the rootfs without any package management or
+        # version info. We also can't depend on kernel headers like
+        # include/generated/uapi/linux/version.h being installed. Even if
+        # we fix this in newer versions of "old NILRT" we still need to be
+        # backwards compatible so it'll just get more complicated.
+        kpath = '/boot/runmode/bzImage'
+        kvregex = r'[0-9]+\.[0-9]+\.[0-9]+-rt'
+        kernel = __salt__['cmd.shell']('strings {0} | awk \'$1 ~ /{1}/ {{print $1}}\''
```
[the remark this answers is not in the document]
10ne1: Done (ported the awk logic to python re)

salt/modules/restartcheck.py
```diff
@@ -307,14 +314,61 @@ def _kernel_versions_nilrt():
...
+                                       .format(kpath, kvregex))
+    else:
+        kernel = os.readlink('/boot/bzImage')
+        kernel = os.path.basename(kernel)
+        kernel = kernel.strip('bzImage-')
```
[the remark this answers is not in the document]
10ne1: Done.

salt/modules/restartcheck.py
```diff
@@ -307,14 +314,61 @@ def _kernel_versions_nilrt():
...
+    cur_timestamp = str(int(os.path.getmtime(depmodpath_base)))
+
+    if prev_timestamp != cur_timestamp:
+        return True
+
+    retcode = __salt__['cmd.retcode']('md5sum -cs {0}'.format(depmodpath_md5sum), output_loglevel="quiet")
+    if retcode != 0:
+        return True
+
+    return False
```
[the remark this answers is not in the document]
10ne1: Done.

salt/modules/restartcheck.py
```diff
@@ -359,8 +413,14 @@ def restartcheck(ignorelist=None, blacklist=None, excludepid=None, verbose=True)
     kernel_current = __salt__['cmd.run']('uname -a')
     for kernel in kernel_versions:
         if kernel in kernel_current:
-            kernel_restart = False
-            break
+            if __grains__.get('os_family') == 'NILinuxRT':
```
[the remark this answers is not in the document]
10ne1: Done.

salt/modules/system.py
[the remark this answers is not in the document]
10ne1: I've rebased on top of master and have the code from #47040 

Can I run any shell commands in the depends decorator? I can check if we're on NILRT using 'lsb-release -i | grep -q nilrt ' if yes.

isbm:
@10ne1 do not forget to push. :wink:

10ne1:
@isbm Sure, I'll push, I just need to do a little more testing (got preempted again but I'm back on it :smiley: )

salt/modules/restartcheck.py
```diff
@@ -298,6 +300,13 @@ def _kernel_versions_redhat():
     return kernel_versions
 
 
+def _is_older_nilrt():
+    '''
+    If this is an older version of NILinuxRT, return True. Otherwise, return False.
+    '''
+    return os.path.exists('/usr/local/natinst/bin/nisafemodeversion')
```
isbm: This is strange. You're judging _version_ by just a mere place of the binary installation. Why you're so sure that the `nisafemodeversion` isn't up to the latest _version_ but just is a symlink for backward compatibility? If you want to look for version, then look for the version by actually calling it or refer to some data that would actually fetch the real version and compare it.

salt/modules/system.py
```diff
@@ -608,3 +609,54 @@ def get_computer_name():
...
+    dir_path = os.path.dirname(NILRT_REBOOT_WITNESS_PATH)
+    if not os.path.exists(dir_path):
+        try:
+            os.makedirs(dir_path)
+        except OSError as ex:
+            raise SaltInvocationError('Error creating {0} (-{1}): {2}'
+                                      .format(dir_path, ex.errno, ex.strerror))
+
+        __salt__['cmd.run']('touch {0}'.format(NILRT_REBOOT_WITNESS_PATH))
+        return True
```
isbm: And if that fails for whatever reason with error code non zero, you're still saying "all perfect". I would not return it here, but at the end, explicitly that the exit code is 0 and then bool it:

```python
errcode = -1
...
if not os.path.exists(dir_path):
    ...
    errcode = ....

return errcode == 0
```

salt/modules/system.py
```diff
@@ -608,3 +609,54 @@ def get_computer_name():
         salt '*' network.get_hostname
     '''
     return __salt__['network.get_hostname']()
+
+
+NILRT_REBOOT_WITNESS_PATH = '/var/volatile/tmp/salt/reboot_witnessed'
+
+NILRT_FAMILY_VERIFY_CMD = 'sh -c \'lsb_release -d | grep -qi "NI Linux Real-Time"\''
```
isbm: It has to be better than this...
```
$ lsb_release -d
-bash: lsb_release: command not found
```

isbm (changes requested):
Sorry, I found some more concerns. Could you look at them?

salt/modules/restartcheck.py
10ne1: We have two big incompatible versions (usually we call them older and newer NILRT) of the distro supported side by side and there is no way to distinguish them from the rootfs (no versioning info to make that distinction in any package). I could implement a machanism to do that but we'd still have to support already released images...

That nisafemodeversion file is not present t all in newer NILRT, as we have a unified bootflow with no "safe mode", so even though it looks ugly that's the most reliable way I could come up to distinguish older vs newer :sweat_smile:

salt/modules/system.py
10ne1: Done.

salt/modules/system.py
10ne1: Done, I'm using the os family grain now.

10ne1:
Rebased (resolved 1 conflict with develop branch) and added 1 more fix.

salt/modules/restartcheck.py
isbm: @10ne1 I hear you. It is still would be nicer if your distro would start using `/etc/os-release` like everyone else are doing.

isbm (approved):

Issue #46757, opened 2018-03-28
salt-minion.pid boot issue leads to high CPU load (Windows)


landergate:
### Description of Issue/Question.
After several weeks of successful usage, salt-minion service started to crash on booting up:
```
2018-03-28 13:17:15,009 [salt.log.setup   :1133][ERROR   ][20736] An un-handled exception was caught by salt's global exception handler:
TypeError: unorderable types: NoneType() < int()
Traceback (most recent call last):
  File "c:\salt\bin\Scripts\salt-minion", line 26, in <module>
    salt_minion()
  File "c:\salt\bin\lib\site-packages\salt\scripts.py", line 168, in salt_minion
    minion.start()
  File "c:\salt\bin\lib\site-packages\salt\cli\daemons.py", line 342, in start
    super(Minion, self).start()
  File "c:\salt\bin\lib\site-packages\salt\utils\parsers.py", line 1039, in start
    self.prepare()
  File "c:\salt\bin\lib\site-packages\salt\cli\daemons.py", line 299, in prepare
    if self.check_running():
  File "c:\salt\bin\lib\site-packages\salt\utils\parsers.py", line 1022, in check_running
    if self.check_pidfile() and self.is_daemonized(pid):
  File "c:\salt\bin\lib\site-packages\salt\utils\parsers.py", line 1028, in is_daemonized
    return os_is_running(pid)
  File "c:\salt\bin\lib\site-packages\salt\utils\process.py", line 187, in os_is_running
    return psutil.pid_exists(pid)
  File "c:\salt\bin\lib\site-packages\psutil\__init__.py", line 1438, in pid_exists
    if pid < 0:
TypeError: unorderable types: NoneType() < int()
```

For some reason, `C:\salt\var\run\salt-minion.pid` contains exactly 4 null bytes:

![8912741](https://user-images.githubusercontent.com/904839/38034215-71b481fa-32aa-11e8-851c-be91b4ede2eb.png)

I suppose a couple of cold system reboots could lead to such file content, but as a result salt-minion can't start and crashing here.

But the issue itself is not about starting salt-minion.
I've detected this situation only after some time, and all that time salt-minion windows service attempted to start minion endlessly, which leads to continuous high CPU usage:

![468346854](https://user-images.githubusercontent.com/904839/38035116-5c2065e6-32ac-11e8-8f0d-8c3d1392f00e.png)

On a long uptime this was leading to 100% cpu and further overall system perf degradation.

### Setup
[...]

PR #46786, opened 2018-03-29
Return int(-1) when pidfile contains invalid data

twangboy:
### What does this PR do?
DISCLAIMER: I know this may not be the right fix. Please recommend the appropriate solution. Perhaps the bug is upstream in `psutil`.

Returns `int(-1)` instead of `None` when the `pidfile` contains invalid data such as null bytes. This was an issue on Windows when salt-minion is running as a service. The issue doesn't occur with salt-minion running in debug mode.

Here's a pidfile with null bytes. Save the file as `salt-minion.pid` and place it in the pidfile location. (`C:\salt\var\run\salt-minion.pid`):
salt-minion.pid

Also uses salt.utils.win_functions.get_parent_pid function to maintain sync with Linux.

https://github.com/saltstack/salt/issues/46757

### Previous Behavior
The salt minion service would fail to start. The NSSM would continue attempting to start the service causing it to burn a lot of CPU cycles.

The minion log showed repeated instances of this stacktrace:
```
2018-03-28 13:17:15,009 [salt.log.setup   :1133][ERROR   ][20736] An un-handled exception was caught by salt's global exception handler:
TypeError: unorderable types: NoneType() < int()
  File "c:\salt\bin\Scripts\salt-minion", line 26, in <module>
    salt_minion()
  File "c:\salt\bin\lib\site-packages\salt\scripts.py", line 168, in salt_minion
    minion.start()
  File "c:\salt\bin\lib\site-packages\salt\cli\daemons.py", line 342, in start
    super(Minion, self).start()
  File "c:\salt\bin\lib\site-packages\salt\utils\parsers.py", line 1039, in start
    self.prepare()
  File "c:\salt\bin\lib\site-packages\salt\cli\daemons.py", line 299, in prepare
    if self.check_running():
  File "c:\salt\bin\lib\site-packages\salt\utils\parsers.py", line 1022, in check_running
    if self.check_pidfile() and self.is_daemonized(pid):
  File "c:\salt\bin\lib\site-packages\salt\utils\parsers.py", line 1028, in is_daemonized
    return os_is_running(pid)
  File "c:\salt\bin\lib\site-packages\salt\utils\process.py", line 187, in os_is_running
    return psutil.pid_exists(pid)
  File "c:\salt\bin\lib\site-packages\psutil\__init__.py", line 1438, in pid_exists
    if pid < 0:
TypeError: unorderable types: NoneType() < int()
```

### New Behavior
The salt-minion starts successfully.

### Tests written?
No

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/utils/parsers.py
+++ b/salt/utils/parsers.py
@@ -35,6 +35,7 @@ import salt.utils.args
 import salt.utils.xdg
 import salt.utils.jid
 import salt.utils.files
+import salt.utils.win_functions
 from salt.utils import kinds
 from salt.defaults import DEFAULT_TARGET_DELIM
 from salt.utils.validate.path import is_writeable
@@ -1020,8 +1021,8 @@ class DaemonMixIn(six.with_metaclass(MixInMeta, object)):
                 if self.check_pidfile() and self.is_daemonized(pid) and not os.getppid() == pid:
                     return True
             else:
-                # We have no os.getppid() on Windows. Best effort.
-                if self.check_pidfile() and self.is_daemonized(pid):
+                # We have no os.getppid() on Windows. Use salt.utils.win_functions.get_parent_pid
+                if self.check_pidfile() and self.is_daemonized(pid) and not salt.utils.win_functions.get_parent_pid() == pid:
                     return True
         return False
 
--- a/salt/utils/parsers.py
+++ b/salt/utils/parsers.py
@@ -1018,11 +1018,11 @@ class DaemonMixIn(six.with_metaclass(MixInMeta, object)):
         if self.check_pidfile():
             pid = self.get_pidfile()
             if not salt.utils.is_windows():
-                if self.check_pidfile() and self.is_daemonized(pid) and not os.getppid() == pid:
+                if self.check_pidfile() and self.is_daemonized(pid) and os.getppid() != pid:
                     return True
             else:
                 # We have no os.getppid() on Windows. Use salt.utils.win_functions.get_parent_pid
-                if self.check_pidfile() and self.is_daemonized(pid) and not salt.utils.win_functions.get_parent_pid() == pid:
+                if self.check_pidfile() and self.is_daemonized(pid) and salt.utils.win_functions.get_parent_pid() != pid:
                     return True
         return False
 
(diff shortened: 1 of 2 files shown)


terminalmage (approved):
This looks like a good fix. I don't think that this should be considered a bug in `psutil.pid_exists()`, it seems reasonable for that function to expect that the pid being passed to it is an integer. Given that `psutil.pid_exists()` returns `False` when fed a negative integer, returning -1 looks to be a good way to deal with this case.

salt/utils/process.py
```diff
@@ -147,7 +147,7 @@ def get_pidfile(pidfile):
             pid = pdf.read().strip()
         return int(pid)
     except (OSError, IOError, TypeError, ValueError):
-        return None
+        return int(-1)
```
skizunov: Just curious, isn't `int(-1)` and `-1` equivalent in Python? Why the extra `int()`?

twangboy: Removed int()

skizunov (approved):

cachedout (approved):
I agree. This looks good.

salt/utils/parsers.py
```diff
@@ -1020,8 +1021,8 @@ def check_running(self):
                 if self.check_pidfile() and self.is_daemonized(pid) and not os.getppid() == pid:
                     return True
             else:
-                # We have no os.getppid() on Windows. Best effort.
-                if self.check_pidfile() and self.is_daemonized(pid):
+                # We have no os.getppid() on Windows. Use salt.utils.win_functions.get_parent_pid
+                if self.check_pidfile() and self.is_daemonized(pid) and not salt.utils.win_functions.get_parent_pid() == pid:
```
DmitryKuzmenko: I think `!=` is better than `not ==`.

salt/utils/process.py
DmitryKuzmenko: Do we need `int` here? Isn't `int(-1) == -1`?

DmitryKuzmenko (changes requested):
As for me early detection looks a bit better. I mean if `get_pidfile` returned `None` we shouldn't continue. Just check `if pid is None: return False` would be good here, it will avoid the following logic to work having a non-int pid value and do further unneeded checks.

DmitryKuzmenko (approved):
But if everybody OK, I'm ok too =)

salt/utils/process.py
DmitryKuzmenko: Sorry, I haven't refreshed the page from tomorrow.

salt/utils/parsers.py
terminalmage: This is a good point.

PR #46789, opened 2018-03-30
Add support to only return failed states to slack via slack_retuner.

mwerickso:
### What does this PR do?
Add support to only return failed states to slack via slack_retuner.

N/A

### Previous Behavior
Previously, using `changes: true` returned both states that had changes and states that failed.

### New Behavior
Now, you can still use `changes: true` to return changed and failed states, or you can use `failed: true` to only return failed states.

### Tests written?
No

### Commits signed with GPG?
Yes


salt/returners/slack_returner.py
```diff
@@ -14,6 +14,7 @@
     slack.as_user (required to see the profile picture of your bot)
     slack.profile (optional)
     slack.changes(optional, only show changes and failed states)
+    slack.failed(optional, only show failed states)
```
cachedout: Would this be more clear if the option was `only_show_failed`?

mwerickso: That is more clear as to what the option does and I have implemented that suggestion.

rallytime:
Hi @mwerickso - It looks like something went sideways here on your PR. Can you rebase this PR so that only your commits are included?

PR #46787, opened 2018-03-29
[schedule][develop] Fix another missed `isinstance` check

s0undt3ch:
### Tests written?

No

### Commits signed with GPG?

Yes

PR #46790, opened 2018-03-30
Fixed _handle_range

pizzapanther:
### What does this PR do?

Fixes `_handle_range` for range schedules.

None

### Previous Behavior

`_handle_range` returned `None` in certain situations thus overwriting the `data` dictionary with `None`. Specifically, my range was hitting the else block on line 1222 and returning `None`.

### New Behavior

Never return anything and update the dictionary in place like it was doing before anyways. This makes the code less prone to errors if you forget to do a `return`.

### Tests written?

No

### Commits signed with GPG?

Maybe

### DIFF
--- a/salt/utils/schedule.py
+++ b/salt/utils/schedule.py
@@ -1191,7 +1191,6 @@ class Schedule(object):
                 data['_error'] = ('Missing python-dateutil. '
                                   'Ignoring job {0}'.format(job))
                 log.error(data['_error'])
-                return data
             else:
                 if isinstance(data['range'], dict):
                     start = data['range']['start']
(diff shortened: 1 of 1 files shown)


salt/utils/schedule.py
```diff
@@ -1203,15 +1202,15 @@ def _handle_range(job, data):
                             data['_error'] = ('Invalid date string for start. '
                                               'Ignoring job {0}.'.format(job))
                             log.error(data['_error'])
-                            return data
+                            return None
```
s0undt3ch: Just `return` no need to `return None`. Linters will likely complain.

garethgreenaway: Or just remove the returns.

pizzapanther: Returns are needed to exit early, I'll clean up the nones

s0undt3ch:
These commits are now included in https://github.com/saltstack/salt/pull/46787. Closing.

PR #46801, opened 2018-04-01
rename jenkins to jenkinsmod

yagnik:
### What does this PR do?
Rename jenkins module from jenkins.py to jenkinsmod.py

No issue

### Previous Behavior
On first try, even if state mentioned reload_modules, salt would load "jenkins" execution module instead of jenkins package. 

### New Behavior
This forces module to load right jenkins from package. Same way problem with dockermod was solved

### Tests written?
No

### Commits signed with GPG?
Yes


cachedout:
@terminalmage Were you involved with the change to `dockermod` that's referenced here? I wasn't aware this was the reason and I'd like to hear more about that.

yagnik:
@cachedout I assumed that was the reason to move to dockermod given that the library probably was importing "docker". Don't quote me on it!
Having said that, the above problem still exists. When we try to import jenkins, the module finds itself instead of the pip package.

terminalmage:
The reason for renaming docker.py to dockermod.py was because our loader temporarily monkeypatches sys.path and adds the module directory at the beginning (see [here](https://github.com/saltstack/salt/blob/2c4668a/salt/loader.py#L1377) and [here](https://github.com/saltstack/salt/blob/2c4668a/salt/loader.py#L1487)). This means that, in `salt/utils/docker.py`, when we try to do an `import docker` to import docker-py, we instead import `salt/modules/docker.py`.

Given that the python bindings for jenkins are named `jenkins`, this does look like the correct solution. In fact, it should be made in 2017.7 and not develop, since this module exists as far back as 2017.7.

cachedout (approved):
Approved per @terminalmage 's recommendation

terminalmage:
Hmm, I just thought of something. We append to sys.path, so the `site-packages` dir still comes first. One would assume then the jenkins module from pypi would take precedence.

Either way though, this is a good change since we don't want our loader to shadow the name of a python module that may potentially be imported.

@yagnik can you re-do this against 2017.7?

terminalmage:
Or we could just merge and backport, since it's already been approved :sunglasses:

PR #46809, opened 2018-04-01
Fix sharedsecret authentication

ezh:
### What does this PR do?

Fix sharedsecret authentication

#46808

### Previous Behavior
for `salt -a sharedsecret test.ping` for example
1. cli ask username
2. cli skip password because of this https://github.com/saltstack/salt/blob/develop/salt/auth/__init__.py#L699
3. minion send request to master with username only
4. master compare sharedsecret with sharedsecret which is always true
Authentication could fail only if there was no such user. 

### New Behavior
1. cli ask username
2. cli ask password
3. minion send request to master with username and password
4. master compare password with sharedsecret which is always true

### Tests written?

No

### Commits signed with GPG?

No

### DIFF
--- a/salt/auth/sharedsecret.py
+++ b/salt/auth/sharedsecret.py
@@ -37,8 +37,8 @@ import logging
 log = logging.getLogger(__name__)
 
 
-def auth(username, sharedsecret, **kwargs):
+def auth(username, password):
     '''
     Shared secret authentication
     '''
-    return sharedsecret == __opts__.get('sharedsecret')
+    return password == __opts__.get('sharedsecret')


terminalmage (approved):

thatch45 (approved):
Yes, this is correct.
While I understand the nature of this module, it would be a good idea to allow the shared secret to be a proper password viable hash with a salt, iterations etc.

thatch45:
For the record, a function to verify and gen secure passwords would look SOMETHING LIKE this:

```
class Passwd():
    def __init__(self):
        self.algo = 'sha512'

    def gen_phash(self, passwd):
        '''
        Return a bytestring to be verified in verify phash
        '''
        salt = os.urandom(32)
        raw = hashlib.pbkdf2_hmac(self.algo, passwd, salt, self.iterations)
        return b':'.join([
            self.algo.encode('utf-8'),
            binascii.hexlify(raw),
            binascii.hexlify(salt),
            str(self.iterations).encode('utf-8')
])

    def verify_phash(self, passwd, phash):
        '''
        Verify the given phash and password combination
        '''
        passwd = passwd.encode('utf-8')
        phash = phash.encode('utf-8')
        comps = phash.split(b':')
        if len(comps) != 4:
            return False
        algo = comps[0].decode('utf-8')
        raw = binascii.unhexlify(comps[1])
        salt = binascii.unhexlify(comps[2])
        iterations = int(comps[3])
        new_raw = hashlib.pbkdf2_hmac(algo, passwd, salt, iterations)
        good = True
        for idx, part in enumerate(raw):
            if part != new_raw[idx]:
                good = False
        if not good:
            rsleep()
return good
```

PR #46811, opened 2018-04-02
Prefer salt.utils.stringutils.expr_match() over re.search() in state.event.

ghost:
### What does this PR do?

Replaces `fnmatch.match()` in `state.event` (`modules/state.py`) with `salt.utils.stringutils.expr_match()`.

A few merits of this approach have been discussed in #46554 and #46656: while @isbm initially supported a disruptive approach, @terminalmage pointed out that breaking the expectations of users is not desired if it can be avoided. So, the enlightened public seems to gravitate towards a non-disruptive approach, which is what this PR now tries to realise.

Sometimes it gets rather tedious to debug event-related problems (ie. broken orchestrator states and so on) on a master which is really busy (200+ events per second) without a decent matching capability to filter events. Giving a possibility to use regular expressions to filter events displayed makes the life of an administrator a tad easier.

### Previous Behavior

The function `state.event()` used `fnmatch.match()` to filter events.

### New Behavior

The function `state.event()` uses `salt.utils.stringutils.expr_match()` to filter events.

### Tests written?

No (`state.event()` neither returns a value nor does it have any side effects).

### Commits signed with GPG?

No


salt/modules/state.py
```diff
@@ -2319,21 +2321,21 @@ def event(tagmatch='*',
...
         ret = sevent.get_event(full=True, auto_reconnect=True)
         if ret is None:
             continue
 
-        if fnmatch.fnmatch(ret['tag'], tagmatch):
+        if salt.utils.stringutils.expr_match(ret['tag'], tagmatch):
             if not quiet:
                 salt.utils.stringutils.print_cli(
-                    str('{0}\t{1}').format(  # future lint: blacklisted-function
+                    '{0}\t{1}'.format(  # future lint: blacklisted-function
```
terminalmage: Can you explain your reasoning for removing the `str()` here? It was done for a very good reason. With `unicode_literals` imported, and `json.dumps()` producing a `str` type, the format string has to also be a `str` type to avoid a decode error when the dumped data includes unicode characters:

```python
>>> import salt.utils.json
>>> data = {'спам': 'яйца'}
>>> data
{u'\u0441\u043f\u0430\u043c': u'\u044f\u0439\u0446\u0430'}
>>> '{0}'.format(salt.utils.json.dumps(data))
UnicodeDecodeError: 'ascii' codec can't decode byte 0xd1 in position 2: ordinal not in range(128)
>>> str('{0}').format(salt.utils.json.dumps(data))
'{"\xd1\x81\xd0\xbf\xd0\xb0\xd0\xbc": "\xd1\x8f\xd0\xb9\xd1\x86\xd0\xb0"}'
>>>
```

This should be reverted back to the way it was.

salt/modules/state.py
```diff
@@ -2343,9 +2345,9 @@ def event(tagmatch='*',
                 )
                 sys.stdout.flush()
 
-            count -= 1
-            log.debug('Remaining event matches: %s', count)
-
+            if count > -1:
+                count -= 1
+                log.debug('Remaining event matches: %s', count)
```
terminalmage: This is a good change, it prevents a bunch of confusing debugging messages that I've seen from time to time as `state.event` is used.

terminalmage (changes requested):

salt/modules/state.py
samilaine: I'm ashamed to admit that the only reason for removing `str()` because @isbm asked rather suggestively why it's there. Ie. the reason is that more attention got paid for who is saying that what was being said.

isbm: @samilaine @terminalmage *\*ahemmm\**, I didn't really meant "please remove it and never use it again". I just thought that `salt.utils.stringutils.to_str` is more correct to use here instead of just `str`.

samilaine: Communication between humans usually fails and when it doesn't, it's an accident. By no means I tried to lay blame on anyone else, as it was yours truly who failed to comprehend that "please remove it and never use it again" wasn't what was meant.

isbm: :wink: no problem, take it easy. The `str` is OK, but my perfectionist guts thinks that using standard utility for this would be more "kosher" here, since it does things according to the current Python version. I would leave this to @terminalmage — he is the Unicode boss here.

terminalmage: `to_str()` is only necessary when it's possible that the string may contain unicode, as it will perform the necessary `.encode()` on PY2 and do nothing on PY3.

Given that we know that the format string is ascii-only, `str()` is fine here.

terminalmage: @samilaine If you can get that line reverted back to what it was before, then this should be good to merge.

terminalmage (changes requested):

samilaine:
Failing to see how to paddle out of the swampy mess created by an untimely rebase, a new PR 46861 has been made to replace this one.

Issue #44243, opened 2017-10-23
Exception handling in module vault.read_secret


wojciechwroblewski:
### Description of Issue/Question
Currently, module `vault.read_secret` raises exception when there is no secret present at given location (endpoint returns http status code 404). This behavior does not allow much flexibility when used in jinja templates as there is no possibility to catch exceptions, causing sls rendering to fail. I'd suggest changing this behavior to handle exception and return `None` instead.

Example below showcases potential benefit of using this module if it returned `None`.

### Setup
Example state:
```
{% set secret = salt.vault.read_secret('/secrets/secret') %}
/tmp/mysecret:
  file.managed:
    - contents:
      - {{ secret or 'hunter2' }}
```

### Steps to Reproduce Issue
`salt-call state.apply file-with-state-above`

### Versions Report
```
Salt Version:
           Salt: 2017.7.1
 
Dependency Versions:
           cffi: Not Installed
       cherrypy: Not Installed
       dateutil: 2.4.2
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.8
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: 0.21.1
           Mako: 1.0.3
   msgpack-pure: Not Installed
 msgpack-python: 0.4.6
   mysql-python: Not Installed
      pycparser: Not Installed
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: Not Installed
         Python: 2.7.12 (default, Nov 19 2016, 06:48:10)
   python-gnupg: Not Installed
         PyYAML: 3.11
          PyZMQ: 15.2.0
           RAET: Not Installed
          smmap: Not Installed
        timelib: Not Installed
        Tornado: 4.2.1
            ZMQ: 4.1.4
 
System Versions:
           dist: Ubuntu 16.04 xenial
         locale: ANSI_X3.4-1968
        machine: x86_64
        release: 4.13.7-1-macbook
         system: Linux
        version: Ubuntu 16.04 xenial
 ```

gtmanfred:
This seems reasonable, would you mind submitting a PR to make this change?

Thanks,
Daniel

PR #46829, opened 2018-04-03
Exception handling in vault module

beornf:
### What does this PR do?
Adds behaviour to handle all exceptions in the vault module meaning jinja templates can be used without breaking.

https://github.com/saltstack/salt/issues/44243

### Previous Behavior
Exceptions are raised that causes sls rendering to sometimes fail.

### New Behavior
Exceptions are logged and execution of sls templates can continue.

### Tests written?

No

### Commits signed with GPG?

Yes


rallytime:
@beornf There's a lint error on this change: https://jenkins.saltstack.com/job/PR/job/salt-pr-lint-n/20627/violations/file/salt/modules/vault.py/

cachedout:
I'm not sure this is the right place for this fix. Many modules raise exceptions, as they should be able to, and they should be handled appropriately. IMHO, this should be fixed inside of the jinja renderer instead.

I'd also like @terminalmage to weigh in on this.

terminalmage (commented):
@cachedout I think that the reasoning is that Jinja provides no means of performing a try/except (at least none of which I know). I'm not sure there's even a way to fix this inside the Jinja renderer, given that the exception would be generated within jinja itself, at runtime.

cachedout (approved):
Approved after talking to @terminalmage

gpaggi:
@beornf / @cachedout this doesn't look like a good idea to me.

In case Vault is unavailable for any reason, Jinja will render the string None in place of the secret. This caused a production incident where passwords were replaced by None and the rendered config file was hot-reloaded by the database (Clickhouse), causing connections to fail.

Have you considered making this behavior configurable? default_none=True, or something like that?
Silently swallowing and ignoring exceptions while rendering a template it's not safe.
For the time being I switched to a self-written Vault module.

cachedout:
@gpaggi Sorry to hear of the production issue. :-/ TBH, I don't remember the specifics of the conversation with @terminalmage as it was a long time ago and I'm not totally sure why we agreed on this approach.

IMHO, a revert of this commit would be reasonable to return to the previous behavior and a new option could then be introduced to continue to render the template even if errors were encountered.

beornf:
@gpaggi Sorry I didn't expect this would happen. I added checks for secrets to ensure they aren't None in Jinja which is extra work and shouldn't be necessary by the principle of least astonishment.

terminalmage:
Yeah, I think it makes sense to treat this like pillar.get works, when `pillar_raise_on_missing` is set. By default we can raise, but an added argument called `default` can be used to suppress the error and return some other value.

terminalmage:
I just opened a pull request, let me know what you think: https://github.com/saltstack/salt/pull/56311

PR #46838, opened 2018-04-03
use http registry for npm

gtmanfred:
### What does this PR do?
Older versions of npm can't validate the newer certificate.

We don't really need to worry about this, and it would be better to just make
sure the tests run correctly that to use https for the tests.

Fixes broken ub14 and centos6 tests

### Tests written?

Yes

### Commits signed with GPG?

Yes


rallytime (approved):

rallytime:
re-run py

cachedout (approved):

PR #46839, opened 2018-04-03
match tuple for targets as well

### LINKED ISSUE #46826: grain modules using tuples affect targeting
when a grain module returns a tuple as opposed to a list it can't be targeted by value as expected

for example `osrelease_info` returns a tuple and as a result cannot be targeted based on a value eg `osrelease_info:6`

what is confusing is that a master will match a minion ok however the minion itself will not match, so running `salt -G "osrelease_info:6" grains.get saltversion` will yield `Minion did not return. [No response]` from all of the minions that did actually match for `osrelease_info:6`

changing the core.py grain definition for osrelease_info to return a list fixes the issue for me but possibly osrelease_info was made a tuple for performance reasons - #13822

probably finding out why the minion can't access `:0` on the tuple would be the right way to fix

### Setup
nothing special

### Steps to Reproduce Issue
salt -G "osrelease_info:6" grains.get saltversion
salt-call grains.get osrelease_info:6

### Versions Report
confirmed issue on 2015.5 and 2017.7

```Salt Version:
           Salt: 2017.7.4

Dependency Versions:
           cffi: Not Installed
       cherrypy: Not Installed
       dateutil: 2.6.1
      docker-py: Not Installed
          gitdb: 2.0.3
      gitpython: 2.1.9
          ioflo: Not Installed
         Jinja2: 2.8.1
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.4.6
   mysql-python: Not Installed
      pycparser: Not Installed
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: Not Installed
         Python: 2.7.14 (default, Jan 31 2018, 02:12:13)
   python-gnupg: Not Installed
         PyYAML: 3.11
          PyZMQ: 14.5.0
           RAET: Not Installed
          smmap: 2.0.3
        timelib: Not Installed
        Tornado: 4.2.1
            ZMQ: 4.0.5

System Versions:
           dist: redhat 6.9 Santiago
         locale: UTF-8
        machine: x86_64
        release: 2.6.32-696.1.1.el6.x86_64
         system: Linux
        version: Red Hat Enterprise Linux Server 6.9 Santiago
```

gtmanfred:
### What does this PR do?
We should be matching tuples in the same way as lists in the targeting (especially since some grains are now tuples)

Fixes #46826

### Tests written?

No

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/utils/__init__.py
+++ b/salt/utils/__init__.py
@@ -1586,7 +1586,7 @@ def subdict_match(data,
                            exact_match=exact_match):
                 return True
             continue
-        if isinstance(match, list):
+        if isinstance(match, (list, tuple)):
             # We are matching a single component to a single list member
             for member in match:
                 if isinstance(member, dict):


terminalmage (approved):
This is good just from a matching perspective, since we shouldn't let a type mismatch cause us to fail to match, but I personally think that we just need to _not be using tuples in grains_. The reason for this is because serialization converts them to lists anyway:

```python
>>> import msgpack
>>> msgpack.loads(msgpack.dumps((1, 2, 3)))
[1, 2, 3]
>>>
```

This would make comparisons fail:

```python
>>> (1, 2, 3) == [1, 2, 3]
False
```

I think this is good to merge and provides a necessary fallback, but an analysis of the core grains to ensure that we aren't using tuples is a good idea, as well as some new documentation for those writing custom grains modules or contributing new grains to Salt.

PR #46840, opened 2018-04-03
Add support to only return failed states to slack.

mwerickso:
### What does this PR do?
Add support to only return failed states to slack via slack_retuner.

N/A

### Previous Behavior
Previously, using `changes: true` returned both states that had changes and states that failed.

### New Behavior
Now, you can still use `changes: true` to return changed and failed states, or you can use `only_show_failed: true` to only return failed states.

### Tests written?
No

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/returners/slack_returner.py
+++ b/salt/returners/slack_returner.py
@@ -215,10 +215,10 @@ def returner(ret):
 
     returns = ret.get('return')
     if changes is True:
-        returns = {key, value for key, value in returns.items() if value['result'] is not True or value['changes']}
+        returns = {(key, value) for key, value in returns.items() if value['result'] is not True or value['changes']}
 
     if only_show_failed is True:
-        returns = {key, value for key, value in returns.items() if value['result'] is not True}
+        returns = {(key, value) for key, value in returns.items() if value['result'] is not True}
 
     if yaml_format is True:
         returns = salt.utils.yaml.safe_dump(returns)
--- a/salt/returners/slack_returner.py
+++ b/salt/returners/slack_returner.py
@@ -215,10 +215,10 @@ def returner(ret):
 
     returns = ret.get('return')
     if changes is True:
-        returns = dict((key, value) for key, value in returns.items() if value['result'] is not True or value['changes'])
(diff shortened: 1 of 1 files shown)


salt/returners/slack_returner.py
```diff
@@ -231,4 +241,4 @@ def returner(ret):
                           username,
                           as_user,
                           api_key)
-    return slack
+    return slack
```
cachedout: Please add a newline below this.

cachedout (changes requested):

salt/returners/slack_returner.py
```diff
@@ -206,10 +209,17 @@ def returner(ret):
...
+    if only_show_failed and changes:
+        log.error('cannot define both slack.changes and slack.only_show_failed in salt config')
+        return
+
     returns = ret.get('return')
     if changes is True:
         returns = dict((key, value) for key, value in returns.items() if value['result'] is not True or value['changes'])
 
+    if only_show_failed is True:
+        returns = dict((key, value) for key, value in returns.items() if value['result'] is not True)
```
gtmanfred: This should be switched to dictionary interpretation, since we no longer support py2.6.

`{key, value for key, value in returns.items() if value['result'] is not True}`

This is significantly faster.

gtmanfred (changes requested):

salt/returners/slack_returner.py
mwerickso: @gtmanfred Good to know, updated lines 218 and 221 with new method.

salt/returners/slack_returner.py
mwerickso: @cachedout Sorry I didnt notice it got deleted. Added back in.

gtmanfred (approved):

rallytime:
@mwerickso Lint: https://jenkins.saltstack.com/job/PR/job/salt-pr-lint-n/20708/violations/file/salt/returners/slack_returner.py/

cachedout (approved):

Issue #32500, opened 2016-04-12
Salt modules depends on utils are not loaded via salt loader


acgupt:
### Description of Issue/Question

Salt Modules depends on utils are not loaded via salt loader. 
### Setup

We have a custom module and custom utils create at location _modules and _utils on master /srv/salt.
Modules are dependent on utils and loaded first in modules
After we do saltutil.sync_all to all minions . And modules and utils synced in the minion cache but failed to load.
### Steps to Reproduce Issue

salt 'minion' sys.list_modules test\* 

or

salt 'minion' test_mod.echo

error in log:   2016-04-11 08:13:32,126 [salt.loaded.ext.module.test_mod][ERROR   ][554][140137025505024] Got exception No module named test_utils: Traceback (most recent call last):
    ImportError: No module named test_utils

This issue is reproducible in  Salt version  2015.5.10 and 2015.8.8.2.
### Versions Report

Salt Version:
           Salt: 2015.8.8.2

Dependency Versions:
         Jinja2: unknown
       M2Crypto: 0.20.2
           Mako: Not Installed
         PyYAML: 3.11
          PyZMQ: 14.5.0
         Python: 2.6.6 (r266:84292, Jul 23 2015, 05:13:40)
           RAET: Not Installed
        Tornado: 4.2.1
            ZMQ: 4.0.5
           cffi: Not Installed
       cherrypy: 3.2.2
       dateutil: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
        libgit2: Not Installed
        libnacl: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.4.6
   mysql-python: Not Installed
      pycparser: Not Installed
       pycrypto: 2.6.1
         pygit2: Not Installed
   python-gnupg: Not Installed
          smmap: Not Installed
        timelib: Not Installed

System Versions:
           dist: oracle 6.7 
        machine: x86_64
        release: 3.8.13-68.3.4.el6uek.x86_64
         system: Oracle Linux Server 6.7 
###### 

   Salt: 2015.5.10
         Python: 2.6.6 (r266:84292, Jul 23 2015, 05:13:40)
         Jinja2: unknown
       M2Crypto: 0.20.2
 msgpack-python: 0.4.6
   msgpack-pure: Not Installed
       pycrypto: 2.6.1
        libnacl: Not Installed
         PyYAML: 3.11
          ioflo: Not Installed
[...]

PR #46841, opened 2018-04-03
Make raw imports from _utils work

mephi42:
This is a quick hack to make it work with salt-ssh.
It would be great if anyone could outline the proper solution, since I got lost in module loading logic.

#32500

### Previous Behavior
Importing `_utils` module resulted in `ImportError`.

### New Behavior
Importing `_utils` module works as described in https://docs.saltstack.com/en/latest/topics/utils/index.html

### Tests written?

No

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/client/ssh/__init__.py
+++ b/salt/client/ssh/__init__.py
@@ -1503,6 +1503,7 @@ def mod_data(fsclient):
             'grains',
             'renderers',
             'returners',
+            'utils',
             ]
     ret = {}
     envs = fsclient.envs()
--- a/salt/loader.py
+++ b/salt/loader.py
@@ -221,7 +221,7 @@ def minion_mods(
     if not whitelist:
         whitelist = opts.get('whitelist_modules', None)
     ret = LazyLoader(
-        _module_dirs(opts, 'modules', 'module'),
+        _module_dirs(opts, 'modules', 'module') + _module_dirs(opts, 'utils'),
         opts,
         tag='module',
         pack={'__context__': context, '__utils__': utils, '__proxy__': proxy},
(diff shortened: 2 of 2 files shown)


terminalmage (commented):
This would not be a good idea, it would add utils functions to both `__salt__` and `__states__`.

Can you show an example of what you're trying to do, what you expect to happen, and what is happening instead? Your most recent comment in the linked issue doesn't really offer much information.

mephi42:
I'm trying to make the following example from the doc work:

`_utils/mymodule.py`:
```
class Foo(object):
    def bar(self):
        return 'baz'
```

`_modules/awesome.py`:
```
import mymodule
def observe_the_awesomeness():
    foo = mymodule.Foo()
    return foo.bar()
```

In v2017.7.3 it fails with an `ImportError`.

terminalmage:
OK, I'm not familiar with this functionality, I will investigate.

mephi42:
I see now what you meant in your comment:

```
[ERROR   ] Module/package collision: '<redacted>/lib/python2.7/site-packages/salt/utils/docker' and '<redacted>/lib/python2.7/site-packages/salt/states/docker.py'
[ERROR   ] Module/package collision: '<redacted>/lib/python2.7/site-packages/salt/utils/pkg' and '<redacted>/lib/python2.7/site-packages/salt/states/pkg.py'
```

I did not use those states when I tested the patch, so it looked okay.

cachedout:
We definitely don't want to merge this. I'm going to close this but we can discuss a path forward in #32500.

mephi42:
No problem, let's discuss the possible alternatives there.

PR #46849, opened 2018-04-03
Unbreak creating etherstubs on SmartOS

jasperla:
### What does this PR do?

SmartOS (no longer?) handles setting an MTU for etherstub devices, so silently drop it in order to maintain the current API.

Also fixup two examples while here which failed to run.

### Previous Behavior

```
salt:2066 salt % doas salt-ssh smartos-gz nictagadm.add storage0 etherstub
smartos-gz:
    ----------
    Error:
        Error: Cannot specify MTU for local nic tag
```

### Tests written?

No (Tested on SmartOS 20160204T080230Z and 20180203T031130Z though)

### Commits signed with GPG?

Yes

@sjorge could you review this too please?

### DIFF
--- a/salt/modules/smartos_nictagadm.py
+++ b/salt/modules/smartos_nictagadm.py
@@ -143,14 +143,14 @@ def add(name, mac, mtu=1500):
     mac : string
         mac of parent interface or 'etherstub' to create a ether stub
     mtu : int
-        MTU
+        MTU (ignored for etherstubs)
 
     CLI Example:
 
     .. code-block:: bash
 
-        salt '*' nictagadm.add storage etherstub
-        salt '*' nictagadm.add trunk 'DE:AD:OO:OO:BE:EF' 9000
+        salt '*' nictagadm.add storage0 etherstub
+        salt '*' nictagadm.add trunk0 'DE:AD:OO:OO:BE:EF' 9000
     '''
     ret = {}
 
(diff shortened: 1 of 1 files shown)


salt/modules/smartos_nictagadm.py
```diff
@@ -163,10 +163,7 @@ def add(name, mac, mtu=1500):
             return {'Error': '{0} is not present on this system.'.format(mac)}
 
     if mac == 'etherstub':
-        cmd = 'nictagadm add -l -p mtu={mtu} {name}'.format(
-            mtu=mtu,
-            name=name
-        )
+        cmd = 'nictagadm add -l {0}'.format(name)
```
sjorge: Can we keep this consistent within the file? and use ether ```{name}``` here too or correct the other ones to use this unnamed format.

sjorge (changes requested):

sjorge:
Aside from a small nit, this looks fine.

salt/modules/smartos_nictagadm.py
jasperla: Thanks, I've chosen to use the unnamed format as that appears to have been more common in the rest of the file.

sjorge (approved):
LGTM

rallytime:
re-run py

cachedout (approved):

PR #46852, opened 2018-04-03
fix creating a nic tag on a link with double 0 in the MAC

jasperla:
### What does this PR do?

Otherwise we get a message like '12:34:00:56:78:90 is not present on this system.' as dladm prints the '00' as '0' and we fail to correctly identify the existing interface.

### Tests written?

No

### Commits signed with GPG?

Yes

@sjorge could you please review this too?


sjorge (approved):
LGTM

cachedout (approved):

PR #46854, opened 2018-04-03
git.latest: add option to hard-reset only when there are remote changes

### LINKED ISSUE #35888: #34869 prevent updating the repository only when the remote changed with local changes

I have a git repository updated with `git.latest`, and I do local changes.

Before that, my state was like this:

```
git:
  git.latest:
    - rev: master
    - branch: master
    - force_fetch: True
    - force_checkout: True
    - target: /git/
```

Everything was ok, and the repository was updated when the remote repository changed.
Now with 2016.3.3 (and dues to #34869 I think), if I do local change to the project, he complains he can't do a fast-forward and I must use `force_reset`. 

But if I use `force_reset`, my local changes are erased each time the state is applied, and I don't want this, I want to erase them only when the remote changed.

To fix this I added this

```
    - force_reset: True
    - unless: "cd /git/ && git remote update && git status -uno | grep 'Your branch is up-to-date with'"
```

but it feel hackish and bad.

Is it possible to have an intermediate `force_reset` option who will reset the repository only when the remote changed ?

Thanks !
### Versions Report

Salt 2016.3.3. Was ok in Salt 2016.3.2.

terminalmage:
This adds a third potential value for the `force_reset` argument, `'remote-changes'` which if used will not discard uncommitted changes unless the local clone is not up-to-date with the remote repository.

Resolves #35888.

### DIFF
--- a/salt/modules/git.py
+++ b/salt/modules/git.py
@@ -2140,13 +2140,9 @@ def discard_local_changes(cwd,
         Windows only. Required when specifying ``user``. This parameter will be
         ignored on non-Windows platforms.
 
-      .. versionadded:: 2016.3.4
-
     ignore_retcode : False
         If ``True``, do not log an error to the minion log if the git command
         returns a nonzero exit status.
-
-        .. versionadded:: 2015.8.0
 
     output_encoding
         Use this option to specify which encoding to use to decode the output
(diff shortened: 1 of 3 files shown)


salt/modules/git.py
```diff
@@ -2115,6 +2115,82 @@ def diff(cwd,
...
+            repositories containing UTF-8 filenames on Windows unreliable.
+            Therefore, Windows will default to decoding the output from git
+            commands using UTF-8 unless this option is explicitly used to
+            specify the encoding.
+
+            On non-Windows platforms, the default output decoding behavior will
+            be observed (i.e. the encoding specified by the locale will be
+            tried first, and if that fails, UTF-8 will be used as a fallback).
+
+        .. versionadded:: 2018.3.1
```
rallytime: Also probably not needed, I think. :)

rallytime (commented):
I just have a couple of really picky comments, otherwise this looks really nice and ++ for tests. :)

terminalmage:
@rallytime good catch, updated. Thanks!

rallytime (approved):

PR #46863, opened 2018-04-04
fixed top function which was broken since commit 002aa88a97e

tamcore:
### What does this PR do?
fixed top function which was broken since commit 002aa88a97e

```
2018-04-04 12:49:51,312 [salt.daemons.masterapi:540 ][ERROR   ][11108] Top function ext_nodes failed with error 'int' object has no attribute 'communicate' for minion random.system.local
```

### Tests written?
No

### Commits signed with GPG?
No

### DIFF
--- a/salt/tops/ext_nodes.py
+++ b/salt/tops/ext_nodes.py
@@ -80,7 +80,7 @@ def top(**kwargs):
         subprocess.Popen(
             cmd,
             shell=True,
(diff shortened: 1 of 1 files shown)


terminalmage (approved):

terminalmage:
Backported to 2018.3 in https://github.com/saltstack/salt/pull/46872 and unit test added.

PR #46873, opened 2018-04-04
Attempt UTF-8 first when decoding/encoding

terminalmage:
This fixes a nagging issue on Windows where the detected locale (cp1252 when running from the command prompt) decodes without error, but produces incorrect results. For example:

```python
>>> u'Ψ'
u'\u03a8'
>>> u'Ψ'.encode('utf-8').decode('utf-8')
u'\u03a8'
>>> u'Ψ'.encode('utf-8').decode('cp1252')
u'\xce\xa8'
>>> u'Д'
u'\u0414'
>>> u'Д'.encode('utf-8').decode('utf-8')
u'\u0414'
>>> u'Д'.encode('utf-8').decode('cp1252')
u'\xd0\u201d'
```

Attempting UTF-8 first keeps Windows from producing incorrectly decoded/encoded results, while still letting the code fall back to using cp1252 if the decode/encode fails.

This also allows us to remove a number of hacky workarounds designed to deal with the above issue on Windows platforms in 2018.3.0, and should keep these sort of problems from cropping up elsewhere in places we've not yet discovered them.

### DIFF
--- a/salt/modules/cmdmod.py
+++ b/salt/modules/cmdmod.py
@@ -610,7 +610,7 @@ def _run(cmd,
             ret['retcode'] = 1
             return ret
 
-        if output_encoding is not None:
+        if output_loglevel != 'quiet' and output_encoding is not None:
             log.debug('Decoding output from command %s using %s encoding',
                       cmd, output_encoding)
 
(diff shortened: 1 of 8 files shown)


rallytime (approved):
This looks great.

skizunov (approved):

terminalmage:
I've also added a commit that keeps the command from being logged when `output_encoding` is used and `output_loglevel == 'quiet'`.

rallytime (approved):

isbm (approved):
@terminalmage great! Thanks.

cachedout (approved):

PR #46888, opened 2018-04-05
Declare set literally

remyleone:
### What does this PR do?

It declares non-empty sets literally.


isbm (changes requested):
While in Python3 sets are fast, in Python2 they are still horribly slow. I'd say I would go with tuple/lists in those declarations instead, so we are supporting both Python major versions equally. These actually can be tuples, because they are only for checking "if X in a set".

@terminalmage what you think?

cachedout:
I'm not sure what we're getting at here. Is this change purely syntactical? I have a vague recollection of there being a tiny performance difference in a literal definition versus a constructed one but I don't recall the details. What else am I missing here?

isbm:
@cachedout same here, seems this is just syntactical, so I am also doubt this is needed at all. And like I said, if one is already doing this, performance-wise pure tuple/list is better:

```bash
$ python -m timeit '"a" in {"d", "c", "r", "b", "a"}'
10000000 loops, best of 3: 0.123 usec per loop

$ python -m timeit '"a" in set(["d", "c", "r", "b", "a"])'
1000000 loops, best of 3: 0.266 usec per loop

$ python -m timeit '"a" in ["d", "c", "r", "b", "a"]'                         
10000000 loops, best of 3: 0.0559 usec per loop
```

salt/utils/jinja.py
```diff
@@ -749,8 +749,7 @@ class SerializerExtension(Extension, object):
     .. _`import tag`: http://jinja.pocoo.org/docs/templates/#import
     '''
 
-    tags = set(['load_yaml', 'load_json', 'import_yaml', 'import_json',
-                'load_text', 'import_text'])
+    tags = {'load_yaml', 'load_json', 'import_yaml', 'import_json', 'load_text', 'import_text'}
```
terminalmage: Looking at the upstream Jinja source code, sets are used for this sort of attribute, so I think the changes made here (and below in this file) are fine.

salt/transport/tcp.py
```diff
@@ -1206,7 +1206,7 @@ def _add_client_present(self, client):
             clients = self.present[id_]
             clients.add(client)
         else:
-            self.present[id_] = set([client])
+            self.present[id_] = {client}
```
terminalmage: As for this, I'm not certain it should be a tuple. The set is never modified, but I don't grok the TCP transport well enough to know whether or not `self.present[id_]` is supposed to be mutable.

salt/state.py
```diff
@@ -1506,24 +1506,8 @@ def requisite_in(self, high):
...
-        req_in_all = req_in.union(
-                set([
-                    'require',
-                    'watch',
-                    'onfail',
-                    'onfail_stop',
-                    'onchanges',
-                    ]))
+        req_in = {'require_in', 'watch_in', 'onfail_in', 'onchanges_in', 'use', 'use_in', 'prereq', 'prereq_in'}
+        req_in_all = req_in.union({'require', 'watch', 'onfail', 'onfail_stop', 'onchanges'})
```
terminalmage: These should not be tuples because we are using set operations (e.g. `union`) here.

salt/runner.py
```diff
@@ -256,7 +256,7 @@ def run(self):
                 # otherwise run it in the main process
                 if self.opts.get('eauth'):
                     ret = self.cmd_sync(low)
-                    if isinstance(ret, dict) and set(ret) == set(('data', 'outputter')):
+                    if isinstance(ret, dict) and set(ret) == {'data', 'outputter'}:
```
terminalmage: Since `set(ret)` and `tuple(ret)` would both produce objects containing just the keys, this can be done just fine with tuples:

```python
if isinstance(ret, dict) and tuple(ret) == ('data', 'outputter'):

File: salt/utils/jinja.py [hunk 5]
```diff
@@ -889,7 +888,7 @@ def load_text(self, value):
 
         return value
 
-    _load_parsers = set(['load_yaml', 'load_json', 'load_text'])
+    _load_parsers = {'load_yaml', 'load_json', 'load_text'}
```
terminalmage: Same as above, this seems to be fine based on upstream Jinja code.

salt/modules/pkg_resource.py
```diff
@@ -311,8 +311,7 @@ def format_pkg_list(packages, versions_as_list, attr):
     '''
     ret = copy.deepcopy(packages)
     if attr:
-        requested_attr = set(['epoch', 'version', 'release', 'arch',
-                              'install_date', 'install_date_time_t'])
+        requested_attr = {'epoch', 'version', 'release', 'arch', 'install_date', 'install_date_time_t'}
```
terminalmage: This needs to stay as a set since we use the `set` type's `&=` operator to modify the set. Changing to a tuple would break this.

terminalmage (changes requested):
@cachedout There is a performance benefit to using the syntactical declaration rather than invoking a function which returns an instance. The reason we were using `set(['foo', 'bar', 'baz'])` instead of `{'foo', 'bar', 'baz'}` has to do with the fact that when most (if not all) of this code was written, we had not yet dropped support for Python 2.6, which doesn't support the curly brace syntax:

```
% docker run --rm -it centos:6 python -c '{1, 2, 3}'
  File "<string>", line 1
    {1, 2, 3}
      ^
SyntaxError: invalid syntax
```

@isbm is right that in many of these cases a tuple is even faster, and it seems that in most of these cases we're literally only using sets to construct something from which we later just do a membership check. These should all be changed to tuples.

I've noted some exceptions below where I feel that the changes made in this PR are correct, but the rest should be updated to use tuples instead of sets.

I would appreciate someone (@isbm?) checking my work on this to make sure I didn't misidentify (or neglect) anything.

salt/modules/pkg_resource.py
isbm: That one yes. I've missed the part. But other than that, if the set is immutable, I would just prefer tuples instead.

salt/runner.py
sieben: Does order matters? Because set are unordered but tuples are.

sieben:
@isbm I've changed to tuples. Except when asked explicitly not to. Could you tell me what you think?

terminalmage (approved):

PR #26803, opened 2015-08-31
Added check for PyMySQL if MySQLdb import fails

twangboy:
- Added PyMySQL to dependencies for windows so that it will be bundled with Salt. See the following PRs:
  https://github.com/saltstack/salt-windows-dev/pull/24
  https://github.com/saltstack/salt-windows-dev/pull/25

Fixes #26754

PR #46890, opened 2018-04-05
fall back to PyMySQL

MaximilianMeister:
### What does this PR do?
adds a fallback to a drop in library. the same way is already done in modules (see #26803)

### DIFF
--- a/salt/auth/mysql.py
+++ b/salt/auth/mysql.py
@@ -55,10 +55,29 @@ import logging
 log = logging.getLogger(__name__)
 
 try:
+    # Trying to import MySQLdb
     import MySQLdb
-    HAS_MYSQL = True
+    import MySQLdb.cursors
+    import MySQLdb.converters
+    from MySQLdb.connections import OperationalError
 except ImportError:
-    HAS_MYSQL = False
+    try:
+        # MySQLdb import failed, try to import PyMySQL
+        import pymysql
+        pymysql.install_as_MySQLdb()
+        import MySQLdb
+        import MySQLdb.cursors
+        import MySQLdb.converters
+        from MySQLdb.err import OperationalError
+    except ImportError:
+        MySQLdb = None
+
+
+def __virtual__():
+    '''
+    Confirm that a python mysql client is installed.
+    '''
+    return bool(MySQLdb), 'No python mysql client installed.' if MySQLdb is None else ''
 
 
 def __get_connection_info():
@@ -95,7 +114,7 @@ def auth(username, password):
                                _info['username'],
                                _info['password'],
                                _info['database'])
-    except MySQLdb.OperationalError as e:
+    except OperationalError as e:
         log.error(e)
         return False
 
--- a/salt/cache/mysql_cache.py
+++ b/salt/cache/mysql_cache.py
@@ -46,11 +46,24 @@ value to ``mysql``:
 from __future__ import absolute_import, print_function, unicode_literals
 from time import sleep
 import logging
+
 try:
+    # Trying to import MySQLdb
     import MySQLdb
-    HAS_MYSQL = True
+    import MySQLdb.cursors
+    import MySQLdb.converters
+    from MySQLdb.connections import OperationalError
 except ImportError:
-    HAS_MYSQL = False
+    try:
+        # MySQLdb import failed, try to import PyMySQL
+        import pymysql
+        pymysql.install_as_MySQLdb()
+        import MySQLdb
+        import MySQLdb.cursors
+        import MySQLdb.converters
+        from MySQLdb.err import OperationalError
+    except ImportError:
+        MySQLdb = None
 
 from salt.exceptions import SaltCacheError
 
@@ -71,12 +84,9 @@ __func_alias__ = {'ls': 'list'}
 
 def __virtual__():
     '''
-    Confirm that python-mysql package is installed.
+    Confirm that a python mysql client is installed.
     '''
-    if not HAS_MYSQL:
-        return (False, "Please install python-mysql package to use mysql data "
-                "cache driver")
-    return __virtualname__
+    return bool(MySQLdb), 'No python mysql client installed.' if MySQLdb is None else ''
 
 
 def run_query(conn, query, retries=3):
@@ -84,13 +94,13 @@ def run_query(conn, query, retries=3):
     Get a cursor and run a query. Reconnect up to `retries` times if
     needed.
     Returns: cursor, affected rows counter
-    Raises: SaltCacheError, AttributeError, MySQLdb.OperationalError
+    Raises: SaltCacheError, AttributeError, OperationalError
     '''
     try:
         cur = conn.cursor()
         out = cur.execute(query)
         return cur, out
-    except (AttributeError, MySQLdb.OperationalError) as e:
+    except (AttributeError, OperationalError) as e:
         if retries == 0:
             raise
         # reconnect creating new client
--- a/salt/pillar/mysql.py
+++ b/salt/pillar/mysql.py
@@ -59,16 +59,27 @@ log = logging.getLogger(__name__)
 
 # Import third party libs
 try:
+    # Trying to import MySQLdb
     import MySQLdb
-    HAS_MYSQL = True
+    import MySQLdb.cursors
+    import MySQLdb.converters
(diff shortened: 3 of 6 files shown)


cachedout (commented):
IMHO, it would be a good idea to structure this so that we can provide logging about which provider ended up getting loaded.

cachedout (approved):

cachedout:
@MaximilianMeister Can you please fix the lint errors below so that we can merge this?

https://jenkins.saltstack.com/job/PR/job/salt-pr-lint-n/20778/violations/

MaximilianMeister:
> Can you please fix the lint errors below so that we can merge this?

@cachedout updated

> Can you please fix the lint errors below so that we can merge this?

@cachedout updated

rallytime:
@MaximilianMeister Actually, I misspoke. It looks like this is failing at the head of the branch. Thanks for submitting this fix. :)

salt/auth/mysql.py
```diff
@@ -55,10 +55,27 @@
 log = logging.getLogger(__name__)
 
 try:
+    log.debug('Trying to import MySQLdb')
```
rallytime: @cachedout Should we be logging in the namespace like this? Auth modules don't have a `__virtual__` method, but will these debug lines _always_ should up in logs?

gtmanfred: Yes they do.

https://github.com/saltstack/salt/blob/2017.7/salt/auth/pam.py#L155

rallytime: Ah, nice. I only checked this auth file and didn't see one. However, @gtmanfred rightly pointed out that anything that is loaded by the LazyLoader should/can have a `__virtual__()`.

@MaximilianMeister Perhaps this module should have one as well and then only log the type of connector once it has been detected. We don't want those log messages flooding people's logs if they're not using these files.

Can you update these files so that the log lines are inside a `__virtual__` function and gated by the appropriate checks? It can be something simple like:

```
def __virtual__():
    '''
    <whatever doc>
    '''
    if HAS_MYSQL_DB:
        log.debug('MySQLdb loaded')
        return True
    elif HAS_PYMYSQL:
        log.debug('PyMySQL loaded')
        return True
    return False, 'Either the \'MySQLdb\' or \'PyMySQL\' libraries must be installed.
```

You can add whatever other log lines you want there, but nothing can be logged outside of a check (`__virtual__` + `if` statement) so we don't flood the logs. Each file in this PR will need the same type of checks applied.

salt/auth/mysql.py
```diff
@@ -55,10 +55,39 @@
...
+        import pymysql
+        pymysql.install_as_MySQLdb()
+        import MySQLdb
+        import MySQLdb.cursors
+        import MySQLdb.converters
+        import MySQLdb.constants
+        from MySQLdb.err import OperationalError
+        HAS_PYMYSQL = True
+    except ImportError:
+        HAS_MYSQLDB = HAS_PYMYSQL = False
```
isbm: I honestly do not like these `UGLY_UPPERCASE_COVARIABLES` all around everywhere, because they are unnecessarily overcomplicating the code so it gets also bulky and hard to read, while value of that is basically zero. I would ask myself: "Do we _really_ care to know what MySQL flavour of driver was loaded? No. We don't really care. So I would not really log that. Therefore:

```python
try:
    import MySQLdb
    ....
except ImportError:
    try:
        import pymysql
        pymysql.install_as_MySQLdb()
        ...
    except ImportError:
        MySQLdb = None
```

At this point you either have a non-`None` object or actually `None`. Then in `__virtual__` function you do:

```python
def __virtual__():
    return bool(MySQLdb), 'No MySQL driver dependency found' if MySQLdb is None else ''
```

Alas, we tend not to log anything in `__virtual__` unless this is really needed. Because you already return the error message as a second element of tuple in `__virtual__` or you just quiet, because it works.

isbm (changes requested):
@MaximilianMeister could you just a bit delete your code? :wink:

isbm:
@cachedout do we really care which provider loaded? Because as long as one of them, both supposed to be supportd _equally_. And therefore it is any of none.

gtmanfred:
Yeah, i don't think we care about which provider loaded.

But, if we wanted to make this more usable, we could combine all the mysql modules everywhere, returners and execution modules, and do kind of like what boto does, and put the basic connection module in salt/utils/mysql.py, then just check and see if that is loaded, and if it is then do a `mysql.assign_funcs` to namespace the functions from utils into the mysql functions, then it would be consistent everywhere.

isbm:
@gtmanfred right! That I would be second on this: put this somewhere like we did to ZMQ or `...moves` in `six`. So we can import `MySQLdb` from something like `salt.utils.providers` (?) and that's it.

MaximilianMeister:
@isbm updated to your suggestions :+1: 

i still dont like the DRY violations either, but as @gtmanfred said it could perhaps be done in a follow up

cachedout:
This all seems right. Good discussion!

I am all right with merging this as it sits but I do agree that consolidation is a very worthwhile future goal.

@isbm Are you willing to approve this PR at this point? If so, could you change your review?

salt/auth/mysql.py
```diff
@@ -55,10 +55,31 @@
...
+        from MySQLdb.err import OperationalError
+    except ImportError:
+        MySQLdb = None
+
+
+def __virtual__():
+    '''
+    Confirm that a python mysql client is installed.
+    '''
+    return (bool(MySQLdb), "No python mysql client installed." if MySQLdb is None else '')
```
isbm: These parenthesis are not needed. Just so:
```python
return bool(MySQLdb), "..."
```

salt/cache/mysql_cache.py
```diff
@@ -71,26 +86,23 @@
...
 def __virtual__():
     '''
-    Confirm that python-mysql package is installed.
+    Confirm that a python mysql client is installed.
     '''
-    if not HAS_MYSQL:
-        return (False, "Please install python-mysql package to use mysql data "
-                "cache driver")
-    return __virtualname__
+    return (bool(MySQLdb), "No python mysql client installed." if MySQLdb is None else '')
```
isbm: These parenthesis not needed too.

salt/pillar/mysql.py
```diff
@@ -59,16 +59,29 @@
...
 
 
 def __virtual__():
-    if not HAS_MYSQL:
-        return False
-    return True
+    '''
+    Confirm that a python mysql client is installed.
+    '''
+    return (bool(MySQLdb), "No python mysql client installed." if MySQLdb is None else '')
```
isbm: And these too. Also for consistency sake, use double-quotes only when needed, say you want show inside the message single quotes etc. But normally we use single quotes all the time. :wink:

salt/returners/mysql.py
```diff
@@ -168,10 +183,10 @@
...
 
 def __virtual__():
-    if not HAS_MYSQL:
-        return False, 'Could not import mysql returner; ' \
-                      'mysql python client is not installed.'
-    return True
+    '''
+    Confirm that a python mysql client is installed.
+    '''
+    return (bool(MySQLdb), "No python mysql client installed." if MySQLdb is None else '')
```
isbm: And here.

isbm (changes requested):
@cachedout Basically it is OK with me, but if this is not burning, then just one more very simple commit and we're done. :wink:

MaximilianMeister:
@isbm @cachedout updated, thanks for the reviews :+1:

isbm (approved):
@cachedout at this point I am fine now with how it is.

rallytime (approved):

rallytime:
@MaximilianMeister There are some lint errors here, however: https://jenkins.saltstack.com/job/PR/job/salt-pr-lint-n/20901/violations/file/salt/modules/mysql.py/

Can you take another look?

MaximilianMeister:
> Can you take another look?

@rallytime fixed thanks

MaximilianMeister:
@rallytime i think i fixed this specific issue now. are the other failures in the same job related? sorry i have no setup to run the tests locally...

rallytime:
Hi @MaximilianMeister - those are not related. Thanks for fixing that!

PR #46892, opened 2018-04-05
Working Python examples for the event documentation.

ghost:
### What does this PR do?

Updates the Python examples in the event system documentation.

The former examples were somewhat antiquated, e.g. they mention `salt.sminion.functions` over which `salt.client.Caller` has been preferred 2015.8.0 according to [the docs](https://docs.saltstack.com/en/latest/ref/clients/#salt-caller).

### Tests written?

No

### Commits signed with GPG?

No

### DIFF
--- a/doc/topics/event/events.rst
+++ b/doc/topics/event/events.rst
@@ -174,17 +174,20 @@ YAML can be used at the CLI in function arguments:
     salt-call event.send 'myco/mytag/success' '{success: True, message: "It works!"}'
 
 If a process is listening on the minion, it may be useful for a user on the
-master to fire an event to it:
+master to fire an event to it. An example of listening local events on
+a minion:
 
 .. code-block:: python
 
     # Job on minion
     import salt.utils.event
 
-    event = salt.utils.event.MinionEvent(**__opts__)
+    opts = salt.config.minion_config('/etc/salt/minion')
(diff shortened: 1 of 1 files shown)


doc/topics/event/events.rst
```diff
@@ -174,17 +174,20 @@ YAML can be used at the CLI in function arguments:
...
+master to fire an event to it. An example of listening local events on
+a minion:
 
 .. code-block:: python
 
     # Job on minion
     import salt.utils.event
 
-    event = salt.utils.event.MinionEvent(**__opts__)
+    opts = salt.config.minion_config('/etc/salt/minion')
```
cachedout: We should provide an example for Windows as well, IMHO. One way to do this might be to use `salt.minion.DEFAULT_MINION_OPTS`.

samilaine: That, however, would then break the example on a *nix system having a non-default configuration file.

On the one hand most readers would be able to figure out that `/etc/salt/minion` isn't available on a Windows system.

On the other hand loading grains just to make an if-clause to handle *nix systems and Windows systems separately would make the example convoluted with stuff slightly if not completely irrelevant material.

However, I made a separate example for a windows system and a *nix system, though I'd actually say this is now worse than the original PR was.

PR #46895, opened 2018-04-05
Add support to skip checking if DB exists in mysql_query

oldmantaiter:
### What does this PR do?
- Some database servers do not support the command used, and instead
  of changing the command we can just skip that check altogether.
- For example, ProxySQL's admin interface has a SHOW DATABASES command
  but not SHOW DATABASES LIKE x.

### Previous Behavior
`mysql_query.run` and `mysql_query.run_file` checked that the database selected for the query exists

### New Behavior
`mysql_query.run` and `mysql_query.run_file` can take a boolean flag `check_db_exists`, which defaults to `True`, to disable the `db_exists` check for each state

### Tests written?

No

### Commits signed with GPG?

No

### DIFF
--- a/salt/states/mysql_query.py
+++ b/salt/states/mysql_query.py
@@ -57,6 +57,7 @@ def run_file(name,
         grain=None,
         key=None,
         overwrite=True,
+        check_db_exists=True,
         **connection_args):
     '''
     Execute an arbitrary query on the specified database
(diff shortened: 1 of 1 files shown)


cachedout (approved):

oldmantaiter:
Thanks @cachedout

PR #46908, opened 2018-04-05

rallytime:
Conflicts:
  - salt/grains/core.py
  - salt/utils/__init__.py


salt/log/setup.py
```diff
@@ -338,6 +338,13 @@ def makeRecord(self, name, level, fn, lno, msg, args, exc_info,
             # If nothing else is in extra, make it None
             extra = None
 
+        salt_system_encoding = __salt_system_encoding__
+        if salt_system_encoding == 'ascii':
+            # Encoding detection most likely failed, let's use the utf-8
+            # value which we defaulted before __salt_system_encoding__ was
+            # implemented
+            salt_system_encoding = 'utf-8'
+
```
terminalmage: This section is redundant now. It's part of the code I used to backport the log argument normalization and the same block exists directly under here. This entire green block should be removed.

terminalmage (changes requested):
One minor change.

terminalmage (approved):

PR #46930, opened 2018-04-06
Clean up bad public key headers

dwoz:
### What does this PR do?

Remove the "RSA" delaration from public key headers generated by
pycryptodome<=3.4.6

#46868

### Previous Behavior

Trying to load public keys with an invalid rsa header cause `salt.crypt.get_rsa_pub_key` to raise an `RSAError: no start line` exception

### New Behavior

Public keys with the invalid rsa header load properly 

### Tests written?

Yes

### Commits signed with GPG?

Yes


rallytime:
@dwoz Looks like that test you wrote isn't super happy in PY2: https://jenkins.saltstack.com/job/PR/job/salt-pr-linode-ubuntu14-n/21609/

dwoz:
@rallytime thanks, I will look into why.

dwoz:
@rallytime the interesting thing is this is not one of the tests I added.

dwoz:
@rallytime Okay, my changes did break that test. I've pushed a fix.

mtorromeo:
Shouldn't we also handle better the exceptions that might be raised here?
Having the master crash because one minion has an invalid key shouldn't be the accepted outcome.
It would be a lot better if the key was just rejected.

dwoz:
@mtorromeo I agree with what you are suggesting. However, I think that change could go into a separate issue. This PR sums up the bug with a regression tests succinctly as it stands right now.

@cachedout, do you have thought's on @mtorromeo's concerns?

cachedout (approved):
After speaking with @thatch45, I am good with this.

garethgreenaway (approved):

Issue #46903, opened 2018-04-05
zfs module broken in 2018.3.0


DeadlyMercury:
My salt minion failed after upgrade to 2018:

```
# salt-call saltutil.sync_all
[ERROR   ] Command '/sbin/zpool list -H -p -o name,size' failed with return code: 2
[ERROR   ] output: invalid option 'p'
usage:
        list [-gHLPv] [-o property[,...]] [-T d|u] [pool] ... [interval [count]]
...

[CRITICAL] Failed to load grains defined in grain file zfs.zfs in function <function zfs at 0x7fa3241bf2a8>, error:
Traceback (most recent call last):
  File "/usr/lib/python2.7/dist-packages/salt/loader.py", line 756, in grains
    ret = funcs[key]()
  File "/usr/lib/python2.7/dist-packages/salt/grains/zfs.py", line 116, in zfs
    grains = salt.utils.dictupdate.update(grains, _zfs_pool_data(), merge_lists=True)
  File "/usr/lib/python2.7/dist-packages/salt/grains/zfs.py", line 102, in _zfs_pool_data
    grains['zpool'][zpool[0]] = _conform_value(zpool[1], True)
IndexError: list index out of range
...

```
Problem is in:
`/usr/lib/python2.7/dist-packages/salt/grains/zfs.py`

```
    98      for zpool in __salt__['cmd.run']('{zpool} list -H -p -o name,size'.format(zpool=zpool_cmd)).splitlines():
```

Must be -P, not -p

Also problems with lxd-formula (but I didn't dig it):
```
[ERROR   ] Failed to import module lxd, this is due most likely to a syntax error:
Traceback (most recent call last):
  File "/usr/lib/python2.7/dist-packages/salt/loader.py", line 1459, in _load_module
    mod = imp.load_module(mod_namespace, fn_, fpath, desc)
  File "/var/cache/salt/minion/extmods/modules/lxd.py", line 114, in <module>
    @salt.utils.decorators.which('lxd')
AttributeError: 'module' object has no attribute 'which'
```

garethgreenaway:
@DeadlyMercury Thanks for the report.  

The zpool issue appears to be related to options potentially changing in the zpool binary.  @sjorge Thoughts on this one?

The lxd formula issue seems to be using a custom module that might need to be updated for some changes around utils that went into 2018.3.

sjorge:
The entire zpool/zfs stuff got a big update that just missed the boot for 2018.3.

[...]

PR #44796, opened 2017-12-02
Fix for Issue #44404

sjorge:
### What does this PR do?
This PR fixes the issue described in #44404

This is done by exposing the parsable output from ```zfs list```, ```zfs get```, ```zpool list```, and ```zpool get``` via the salt zfs and zpool module to the zfs state.

We then conver the user input to what the zfs parsable output is so we can properly compare them.

### What issues does this PR fix or reference?
#44404

### Previous Behavior
A user could set a property to '1024G' without parsable output zfs would return this as '1T' on the next run, resulting in the property always being updates.

### New Behavior
We convert the user's input to the zfs parsable format, we then compare it to the output of zfs.

### Tests written?
No

### Commits signed with GPG?
No

PR #46933, opened 2018-04-06
Only very recent zfs support parsable output

sjorge:
### What does this PR do?
~~On the latest LTS version of ubuntu there is not support for the parsable output in zfs/zpool.~~
Only recent zfs on linux supports parsable output.
We can simply strip the -p flag if we are running on this version.

It also looks like Oracle ZFS does not have the ```-p``` flag at all.

Since we now do value normalization, we can deal with not having parsable output.
However we still request parsable output when we can, as this is faster.

#46903

### Previous Behavior
On older zfs on linux, zpool/zfs commands do not yet support parsable output (-p), it would fail.
Likewise for Oracle ZFS

### New behavior
We drop the ```-p``` flag.

### Tests written?
No

### Commits signed with GPG?
No

### DIFF
--- a/salt/grains/zfs.py
+++ b/salt/grains/zfs.py
@@ -57,14 +57,14 @@ def _zfs_pool_data():
     # collect zpool data
     zpool_list_cmd = __utils__['zfs.zpool_command'](
         'list',
-        flags=['-H', '-p'],
+        flags=['-H'],
         opts={'-o': 'name,size'},
     )
     for zpool in __salt__['cmd.run'](zpool_list_cmd).splitlines():
         if 'zpool' not in grains:
             grains['zpool'] = {}
         zpool = zpool.split()
-        grains['zpool'][zpool[0]] = __utils__['zfs.to_size'](zpool[1], True)
+        grains['zpool'][zpool[0]] = __utils__['zfs.to_size'](zpool[1], False)
 
     # return grain data
     return grains
--- a/salt/modules/zfs.py
+++ b/salt/modules/zfs.py
@@ -322,7 +322,7 @@ def list_(name=None, **kwargs):
 
     ## Configure command
     # NOTE: initialize the defaults
-    flags = ['-H', '-p']
+    flags = ['-H']
     opts = {}
 
     # NOTE: set extra config from kwargs
@@ -1198,7 +1198,7 @@ def get(*dataset, **kwargs):
     '''
     ## Configure command
     # NOTE: initialize the defaults
-    flags = ['-H', '-p']
+    flags = ['-H']
     opts = {}
 
     # NOTE: set extra config from kwargs
--- a/salt/modules/zpool.py
+++ b/salt/modules/zpool.py
@@ -405,7 +405,7 @@ def list_(properties='size,alloc,free,cap,frag,health', zpool=None, parsable=Tru
     res = __salt__['cmd.run_all'](
         __utils__['zfs.zpool_command'](
             command='list',
-            flags=['-H', '-p'],
+            flags=['-H'],
             opts={'-o': ','.join(properties)},
             target=zpool
         ),
@@ -476,7 +476,7 @@ def get(zpool, prop=None, show_source=False, parsable=True):
     res = __salt__['cmd.run_all'](
         __utils__['zfs.zpool_command'](
             command='get',
-            flags=['-H', '-p'],
+            flags=['-H'],
             opts={'-o': ','.join(value_properties)},
             property_name=prop if prop else 'all',
             target=zpool,
--- a/tests/unit/modules/test_zfs.py
+++ b/tests/unit/modules/test_zfs.py
@@ -304,15 +304,15 @@ class ZfsTestCase(TestCase, LoaderModuleMockMixin):
         '''
         res = OrderedDict([
             ('myzpool', OrderedDict([
-                ('used', 905792561152),
-                ('avail', 1024795238400),
+                ('used', 849329782784),
+                ('avail', 1081258016768),
                 ('refer', 98304),
                 ('mountpoint', '/myzpool'),
             ])),
         ])
         ret = {}
         ret['retcode'] = 0
-        ret['stdout'] = 'myzpool\t905792561152\t1024795238400\t98304\t/myzpool'
+        ret['stdout'] = 'myzpool\t791G\t1007G\t96K\t/myzpool'
         ret['stderr'] = ''
         mock_cmd = MagicMock(return_value=ret)
         with patch.dict(zfs.__salt__, {'cmd.run_all': mock_cmd}), \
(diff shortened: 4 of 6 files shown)


sjorge:
@Ch3LL this only fixes it in the develop branch.

It's not really easy to fix in 2018.3, we switched to including -p because people were reporting bugs because 1g == 1G,... so using ```-p``` in those cases fixed all that.

Would it be possible to backport salt.utils.zfs, salt.grains.zfs, salt.modules.zfs, salt.modules.zpool, salt.states.zfs and salt.states.zpool from develop to 2018.3? It's a rather big update though.

myii:
@sjorge This is not limited to 16.04 Xenial.  I've got a number of boxes still running 14.04 Trusty, which is still supported until 2019.  What the issue here appears to be the introduction of the `-p` switch, which appeared after ZFS `v0.7.0` at some point:

https://github.com/zfsonlinux/zfs/commit/d2f3e292dccab23e47ade3c67677a10f353b9e85

> ...
> Those changes have been merged, re-factored, **a -p option added and extended to all relevant zpool subcommands**.
> ...

* Emphasis above is mine

In terms of versions, then 14.04 is on `v0.6.5` (`zfsutils: 0.6.5.11-1~trusty`), just like 16.04 (`zfsutils: 0.6.5.6-0ubuntu19`) -- that's quoted from the linked issue.

Instead of checking for 16.04, please check for ZFS earlier than `v0.7.0`.

myii:
@sjorge Some more good/bad news for you.

I've tested a change on my `2018.3` master and _for my use case_, the following is working.

```diff
+++ /usr/lib/python2.7/dist-packages/salt/grains/zfs.py
@@ -95,7 +95,7 @@
 
     # collect zpool data
     zpool_cmd = salt.utils.path.which('zpool')
-    for zpool in __salt__['cmd.run']('{zpool} list -H -p -o name,size'.format(zpool=zpool_cmd)).splitlines():
+    for zpool in __salt__['cmd.run']('{zpool} list -H -o name,size'.format(zpool=zpool_cmd)).splitlines():
         if 'zpool' not in grains:
             grains['zpool'] = {}
         zpool = zpool.split()
```

* That is: https://github.com/saltstack/salt/blob/2018.3/salt/grains/zfs.py#L98

Of course, the `salt/utils/zfs.py` file doesn't exist at all.

sjorge:
As mentioned here https://github.com/saltstack/salt/pull/46933#comment-379583681, I will need to update this PR to drop -p on Linux in all cases.

myii:
@sjorge The issue isn't Linux; it is the version of ZFS.  I've been testing an early release of 18.04 Bionic:

```bash
# salt -Cv 'G@lsb_distrib_release:18.04' cmd.run 'zpool list -H -p -o name,size'
Executing job with jid 20180409080336821223

E6530-02:
    ZFSroot     246960619520
```

And the versions of the ZFS packages:

```bash
# salt -Cv 'G@lsb_distrib_release:18.04' cmd.run 'dpkg -l | grep zfs'
Executing job with jid 20180409080511274357

E6530-02:
    ii  libzfs2linux                          0.7.5-1ubuntu7                      amd64        OpenZFS filesystem library for Linux
    ii  zfs-auto-snapshot                     1.2.4-1                             all          ZFS automatic snapshot service
    ii  zfs-zed                               0.7.5-1ubuntu7                      amd64        OpenZFS Event Daemon
    ii  zfsutils-linux                        0.7.5-1ubuntu7                      amd64        command-line tools to manage OpenZFS filesystems
```

sjorge:
Yes, I know the issue is not the zfs version.
However in salt.utils.zfs we do not have grains or much of salt available to us.

That would either mean adding a whole lot of code to properly detect it, or just disable usage of the '-p' flag when running on linux. The impact should be rather minor as salt.utils.zfs internally normalizes the values.

cachedout (approved):

sjorge:
@cachedout based on more report of this not working like https://github.com/saltstack/salt/commit/a64d81d2cdaf47a471ad2a0cd9ef242e6a42ed45#commitcomment-28491782

How do you feel about dropping the -p flag altogether? The values get normalized anyway so it will only include a small performance hit, probably not noticeable at all in most cases?

dmurphy18:
The -p is causing issues on Solaris 10 and 11.3

sjorge:
So to summarize ```-p``` does not work on:
- Oracle ZFS
- ZFS on Linux (unless 0.7.x)
- A rather old illumos vm I had laying around

I'm reworking this PR to strip it out completely.

sjorge:
There is one tiny thing I need to fix in this, but it will have to be for tomorrow. I will re-add the wil tag.

rallytime:
Thank you for your quick work on resolving this issue @sjorge! It is very much appreciated.

sjorge:
@rallytime can we get these backported into 2018.3 too, else a different fix is needed for that branch and we do not have the value normalization in there yet, which will make it much more difficult.

rallytime:
@sjorge I have marked this for back-porting to 2018.3. Is something more than the back-port needed?

sjorge:
As long as we grab:
- salt/utils/zfs.py
- salt/grains/zfs.py
- salt/modules/zfs.py
- salt/states/zfs.py
- salt/modules/zpool.py
- salt/states/zpool.py
...

I think we should be good, well and the tests too. I already used to verify this fix. (aka the ones I expected to break, broke)

rallytime:
Can you explain what you mean by "as long as we grab"?

sjorge:
My big PR with fixes for the zfs/zpool stuff that introduced salt.utils.zfs landed just after 2018.3 got tagged, so just merging in this PR will probably not be enough.

rallytime:
@sjorge There isn't a way to fix that bug on the 2018.3 branch separately from pulling in those files or the PR with those massive changes? We don't usually like to do that in dot releases.

sjorge:
There were a few issues discovered in 2017.x and I fixed them by using the ```-p``` flag and minimal value normalization (_conform_value). While doing so I discovered a few other issues, so I did the big rewrite moving all the duplicate code into utils and doing the value normalization.

It might be possible to selectively move bits from salt.utils.zfs into 2018.3 but that would take me a weekend or two at minimal.

rallytime:
@sjorge OK, thanks for the clarification. That makes more sense now. I'd like to bring @cachedout in on this discussion and see what he thinks.

sjorge:
@dmurphy18 sounds good, I think you need:
Original PR:
8c7b3e530c3e2ae285dcabdbf180ecb5231613e1
dd29778ed5337919b2601a04a5d1504cf9bf7f21
01da93ec1f88054dd914978b5d0b246838ad7205
8517e876e4d8b40b9ca6c21795054474a3b763b4
ba45fcca31d34972e14295b0b282655a47d2d865
f4bcf5fa71bd9ecd9c96159601867c19ce133f34
ce5c9792ea602191cfaa6402b4dde186365d9629
9d18da9e2ea300ab37bffc54f2b0daeec34d94c3
fce3920b07afe28cf2945687d2668c77865ce8f4
5ca342494f6e976638172c68b875698f79d86e54
e7e902662bbcc82200ddd7ac48a3b07d7d48143d
a64d81d2cdaf47a471ad2a0cd9ef242e6a42ed45
505b816914a29306b0a05d9c137ede9f4bf4da03
7b32faa0ced71817e88ab49fb83176bd0941a581

Other commits that touch these files in between (these touch other files too):
97a39b8b0a4cc5358e490a553fe65d5764b1541c
ed46ce3c746b9c9bba77240ad4a9a39cd3a11aad (merge forward from 2018.3, so not needed?)
76603042622576eeb5f14c5cf8c33effa95b5075 (fix for merge conflict caused by ed46ce3c746b9c9bba77240ad4a9a39cd3a11aad)


Fix for ```-p```:
36819172d6d8c0ce493c009bfa12fea067e223d3

Sorry for the mess with the ```-p``` It was worked fine on all the platforms I tested on (granted, they were all relatively recent), I used the same flag in the fix from #44796 and that was verified by the other person too.

dmurphy18:
@sjorge Thanks for the list.  
Some us have to deal with ancient stuff but another few years and it will be gone.

sjorge:
You might also want to grab #47078

dmurphy18:
@sjorge With 97a39b8, I should only be concerned with smartos_virt.py and utils/zfs/py ?

sjorge:
Yeah looks right, although smartos_virt.py is probably not needed.

Issue #46881, opened 2018-04-05
Cron.file - source file not found


SynPrime:
### Description of Issue/Question

Getting an error, when I try to run cron.file function. The source file isn't found.
The file is present at base/cron/starscream/root = salt://cron/starscream/root

### Setup
(Please provide relevant configs and/or SLS files (Be sure to remove sensitive info).)

```
root_crontab:
  cron.file:
    - name: salt://cron/starscream/root
    - user: root
```

### Steps to Reproduce Issue
(Include debug logs if possible and relevant.)

`salt-call state.apply cron.starscream`

```
[ERROR   ] Source file salt://cron/starscream/root not found
local:
----------
          ID: root_crontab
    Function: cron.file
        Name: salt://cron/starscream/root
      Result: False
     Comment: Source file salt://cron/starscream/root not found
     Started: 03:57:06.065633
    Duration: 498.876 ms
     Changes:

Summary for local
------------
Succeeded: 0
Failed:    1
```

### Versions Report
(Provided by running `salt --versions-report`. Please also mention any differences in master/minion versions.)
```
Salt Version:
           Salt: 2018.3.0

Dependency Versions:
           cffi: Not Installed
       cherrypy: Not Installed
       dateutil: 2.5.3
      docker-py: Not Installed
          gitdb: 2.0.0
      gitpython: 2.1.1
          ioflo: Not Installed
         Jinja2: 2.9.4
        libgit2: 0.24.5
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: 1.0.6
   msgpack-pure: Not Installed
 msgpack-python: 0.4.8
   mysql-python: Not Installed
      pycparser: Not Installed
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: 0.24.2
         Python: 2.7.13 (default, Nov 24 2017, 17:33:09)
   python-gnupg: Not Installed
         PyYAML: 3.12
          PyZMQ: 16.0.2
           RAET: Not Installed
          smmap: 2.0.1
        timelib: Not Installed
        Tornado: 4.4.3
            ZMQ: 4.2.1

System Versions:
           dist: debian 9.4
         locale: UTF-8
        machine: x86_64
        release: 4.9.0-6-amd64
         system: Linux
        version: debian 9.4
```

garethgreenaway:
@SynPrime Thanks for the report.  Are your state files stored under `/srv/salt/` on the Salt master?  Have you changed the default file roots at all?

[...]

PR #46944, opened 2018-04-08
[2018.3] cron.file with salt source URL

garethgreenaway:
### What does this PR do?
Fixing bug that made cron.file unable to use a file via a Salt URL.

#46881 

### Tests written?
Yes

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/states/cron.py
+++ b/salt/states/cron.py
@@ -606,6 +606,7 @@ def file(name,
             user,
             group,
             mode,
+            [],
             __env__,
             context,
             defaults,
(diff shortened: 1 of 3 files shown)


salt/states/cron.py
```diff
@@ -606,6 +606,7 @@ def file(name,
             user,
             group,
             mode,
+            [],
```
rallytime: I know this wasn't your call originally, but maybe we should change all of these kwargs to use their `kwarg=thing` syntax to prevent regressions in the future.

garethgreenaway: This would be helpful.  I'll update the PR.

rallytime (approved):

Issue #44597, opened 2017-11-19
implement ignore_missing for /srv/salt/top.sls as well


tampakrap:
### Description of Issue/Question

The `ignore_missing: True` option at the pillar top.sls file allows us to include a file only if it exists. Such an option would be useful to have for the salt top.sls file as well.

### Setup
(Please provide relevant configs and/or SLS files (Be sure to remove sensitive info).)

`/srv/salt/top.sls`

```salt
{% set roles = salt['grains.get']('roles', []) %}

base:
  '*':
    - role.base
  {% for role in roles %}
  'role:{{ role }}':
    - match: grain
    #- ignore_missing: True
    - role.{{ role }}
  {% endfor %}
```

garethgreenaway:
@tampakrap Thanks for the report.  This would be a useful feature to have.

mchugh19:
This seems to be a dupe for #16687 
(I found this issue when also looking for this functionality)

mchugh19:
This is still a valid feature request

mchugh19:
Still a valid feature request

cdalvaro:
Please, don't forget this request

Issue #39186, opened 2017-02-06
Add support for 'ignore_missing: True' option to states


ksoviero:
### Description of Issue/Question

Our pillar/top.sls file looks like this:

```yaml
{% set role = grains['ec2_tags']['role'] %}
{% set host = grains['id'].replace('.', '_') %}
{% set env = grains['ec2_tags']['env'] %}

base:
  '*':
    - roles.all

  'ec2_tags:role:{{ role }}':
    - match: grain
    - roles.{{ role }}
    - ignore_missing: True

  '{{ grains['id'] }}':
    - hosts.{{host}}
    - ignore_missing: True

  'ec2_tags:env:{{ env }}':
    - match: grain
    - envs.{{ env }}
    - ignore_missing: True
```

I'd like to do something very similar for states, however the states/top.sls file does not support the 'ignore_missing: True' attribute. As such, anytime we create a host with a new role, we have to make sure that the role is already defined in Salt. Otherwise, users won't even be able to login to the new hosts since the roles.all state will not have applied.

On top of that, host-specific states (edge cases can require this) are impossible since any host that doesn't need one will throw an error regardless since it will try to import it.

### Versions Report

```
# salt --versions-report
Salt Version:
           Salt: 2016.11.1
 
Dependency Versions:
           cffi: Not Installed
       cherrypy: Not Installed
       dateutil: 2.4.2
          gitdb: 0.6.4
      gitpython: 1.0.1
          ioflo: Not Installed
         Jinja2: 2.8
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: 1.0.3
   msgpack-pure: Not Installed
 msgpack-python: 0.4.6
   mysql-python: Not Installed
      pycparser: Not Installed
       pycrypto: 2.6.1
         pygit2: Not Installed
         Python: 2.7.12 (default, Nov 19 2016, 06:48:10)
   python-gnupg: Not Installed
         PyYAML: 3.11
          PyZMQ: 15.2.0
           RAET: Not Installed
          smmap: 0.9.0
        timelib: Not Installed
        Tornado: 4.2.1
            ZMQ: 4.1.4
 
System Versions:
           dist: Ubuntu 16.04 xenial
        machine: x86_64
        release: 4.4.0-59-generic
         system: Linux
        version: Ubuntu 16.04 xenial
```

ksoviero:
Marked as duplicate.

PR #19429, opened 2015-01-07
Add new ignore_missing option to pillar top

ryan-lane:

Issue #16687, opened 2014-10-16
Add module/function and top.sls syntax to check for sls existence


ryan-lane:
Currently, if I want to conditionally include an sls file, if the file exists, I need to do some ugly jinja magic:

``` yaml
base:
  # For each directory that can have pillar files
  {% for root in opts['pillar_roots']['base'] -%}
  # Set a variable to the path to check
  {% set service_group_sls = '{0}/{1}.sls'.format(root, grains['service_group']) -%}
  # Check to see if the file exists, if so, include that pillar file.
  {% if salt['file.file_exists'](service_group_sls) %}
  'service_group:{{ grains["service_group"] }}':
    - match: grain
    - {{ grains['service_group'] }}
  {% endif %}
  {% endfor -%}
```

What I really want to do is:

``` yaml
base:
  'service_group:{{ grains["service_group"] }}':
    - match: grain
    - ignore_missing: True
    - {{ grains['service_group'] }}
```

I also want to see if a pillar or state sls file exists from the cli, or from inside of a state, for conditional include:

``` bash
salt-call state.sls_exists mystatssls
salt-call pillar.sls_exists mypillarsls
```

This way I can call individual states, but only if they actually exist.

ryan-lane:
Fixed in #19429

syphernl:
I'm working on something that requires a similar approach and running into an issue that Salt doesn't ignore it if I try to `include` a state that does not exist.

It looks like for states the `cp.show_states` could come in handy. 
The proposed `sls_exists` can most likely be a wrapper around that probably?

invsblduck:
Too bad this was only implemented for Pillar!  Generic support for SLS inclusion in `top.sls` (as the title states!) or SLS inclusion from within another SLS would have been epic.  But that is no small patch, I'm sure.  Thanks @ryan-lane for getting the Pillar part done! :+1:

ryan-lane:
Yeah... The code for pillar and state top files are separate for some reason, even though they do the exact same thing. I was hoping to do both at the same time, but I only needed pillar and the state top file code looked more difficult to modify. I'm still hoping someone from SaltStackInc will finish this work up :)

hartloosli:
Bumping this, as it would really clean up a tedious top.sls in our code.

dgorissen:
Linking this to duplicate #20010

ksoviero:
#39186 is a duplicate as well.

zerthimon:
+1
I really need this functionality:
```
[...]

PR #46952, opened 2018-04-09
ignore_missing parameter for state top files

ghost:
### What does this PR do?

Adds a new keyword parameter, `ignore_missing`, for the state top file entries.

The way a new keyword parameter handling has been added appears, at least to me, to be the least intrusive way of doing it. The previous behaviour is unchanged, and only when the new parameter is present is the behaviour changed in any way.

This is not the cleanest possible way to implement this feature; however, considering the slight performance penalty together with the fact that all other ways of handling appear to require a much larger rewrite, I'd say the price is acceptable, if not actually negligible.

Some sther options for handling a new keyword parameter for a top file entry might be:

1. Incorporate similar changes from the pillar top file handling from #19429.
2. Write from scratch a new mechanism to handle both pillar and state top files.
3. Change all functions handling the top file matches to include a list of states *and* a list of parameters, eg. render_highstate(self, matches) would turn into render_highstate(self, states, parameters).

First two options would require quite a bit refactoring, which I think is out of the scope of implementing just a new parameter. The refactoring probably should be done independently.

The third option would require changing things outside the current module. For example, `salt.minion.Matcher` gets called, and it takes a list of states and keyword parameters (ie. a list of strings and dictionaries). Thus, I believe that changing function fingerprints would amount to nothing because a list of states and keyword parameters would have to be combined again for calls to `salt.minion.Matcher` (and possibly others).

#16687
#39186
#44597

### Previous Behavior

Any missing state files in a top file entry matching for a minion produce an error when rendering the higher state.

### New Behavior

Any missing state files in a top file entry matching for a minion produce an error when rendering the higher state, unless the top file entry includes the parameter `ignore_missing: True`.

For example, assuming that the `missing.state` is nowhere to be found in the fileserver backend, this will produce an error:

```
base:
  '^dev[0-9]+.*':
    - match: pcre
    - some.state
    - missing.state
```

Whereas this will happily compile the higher state, using the states found from the fileserver backend:

```
base:
  '^dev[0-9]+.*':
    - match: pcre
    - some.state
    - missing.state
    - ignore_missing: True
```

### Tests written?

No

### Commits signed with GPG?

No

### DIFF
--- a/salt/state.py
+++ b/salt/state.py
@@ -3259,19 +3259,46 @@ class BaseHighState(object):
                             raise SaltRenderError('Unable to render top file. No targets found.')
         return top
 
+    @staticmethod
+    def _get_topmatch_params(lst):
+        '''
+        Take a list of state names and dictionaries of keyword parameters,
+        and produce a dictionary of all keyword parameters.
+
+        Input:
+        { 'env': [ 'state1', { 'prm1': 'v1' }, 'state2', { 'prm2': 'v2' } ] }
+
+        Returns:
+        { 'prm1': 'v1', 'prm2': 'v2' }
+        '''
+        ret = {}
+        for elem in lst:
+            if isinstance(elem, dict):
+                ret.update(elem)
+        return ret
+
+    @staticmethod
+    def _get_topmatch_states(lst):
+        '''
+        Take a list of state names and dictionaries of keyword parameters,
+        and produce a list of just state names.
+
+        Input:
+        { 'env': [ 'state1', { 'prm1': 'v1' }, 'state2', { 'prm2': 'v2' } ] }
+
+        Returns:
+        [ 'state1', 'state2' ]
+        '''
+        return [x for x in lst if isinstance(x, six.string_types)]
+
     def _merge_tops_merge_all(self, tops):
         '''
-        Merge the top files into a single dictionary
+        Merge the top files into a single dictionary and return a dictionary
+        of keyword parameters and a list of states to be applied.
         '''
         def _read_tgt(tgt):
-            match_type = None
-            states = []
-            for item in tgt:
-                if isinstance(item, dict):
-                    match_type = item
-                if isinstance(item, six.string_types):
-                    states.append(item)
-            return match_type, states
+            return (self._get_topmatch_params(tgt),
+                    self._get_topmatch_states(tgt))
 
         top = DefaultOrderedDict(OrderedDict)
         for ctops in six.itervalues(tops):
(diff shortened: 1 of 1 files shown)


salt/state.py
```diff
@@ -3284,12 +3311,11 @@ def _read_tgt(tgt):
...
-                            m_type1, m_states1 = _read_tgt(top[saltenv][tgt])
-                            m_type2, m_states2 = _read_tgt(ctop[saltenv][tgt])
+                            kargs1, m_states_1 = _read_tgt(top[saltenv][tgt])
+                            kargs2, m_states_2 = _read_tgt(ctop[saltenv][tgt])
                             merged = []
-                            match_type = m_type2 or m_type1
-                            if match_type is not None:
-                                merged.append(match_type)
+                            merged.extend(kargs1)
+                            merged.extend(kargs2)
```
terminalmage: The whole point of the three lines you're replacing with the above two lines, is to only append the `match_type` dict if it is specified. Hence the if statement, which only appends the dict if a `match` keyword param was specified. Appending twice here doesn't seem right to me, can you explain your thought process here? It would seem to me that if you wanted to support multiple keyword parameters, you'd want to do something like this:

```python
kargs = {}
kargs.update(kargs1)
kargs.update(kargs2)
if kargs:
    merged.append(kargs)
```

This would only append if at least one keyword param was used.

Then again, I may be misunderstanding, so I'd like to know what your intent was here. Another thing that doesn't make sense to me is that extending a list with a dict only gives you the key names:

```python
>>> x = []
>>> y = {'foo': 'bar', 'abc': 'def'}
>>> x.extend(y)
>>> x
['foo', 'abc']
>>>
```
So it would seem like we lose the actual values here.

salt/state.py
```diff
@@ -3391,13 +3419,9 @@ def _filter_matches(_match, _data, _opts):
...
-                            elif isinstance(item, dict):
-                                env_key, inc_sls = item.popitem()
-                                if env_key not in self.avail:
-                                    continue
-                                if env_key not in matches:
-                                    matches[env_key] = []
-                                matches[env_key].append(inc_sls)
+                            # Pass through any keyword parameters.
+                            if isinstance(item, dict):
+                                matches[saltenv].append(item)
```
terminalmage: I also don't understand this. The code you're replacing looks to be designed to include matches from a specific environment, and your replacement code removes this completely. Can you explain your thinking here?

terminalmage (commented):
I'm confused by this. You only seem to be modifying the merging behavior for a single merging strategy, for one, and I don't really understand the reasoning behind the changes you are making. I'm hoping you can help clarify some of this for me.

samilaine:
I'll try to tackle the comments one at a time.

> You only seem to be modifying the merging behavior for a single merging strategy[.]

That indeed is true. Tottering on the outer limits of my abilities while trying to understand what the somewhat spaghettish contents of `salt/state.py` tries to do, I shot out a PR, knowing well that it's not a complete solution, to get input about whether or not this general direction is at all desirable.

Since someone took the time to read the code and comment it, that seems to me a hint that the way I was heading with this isn't one completely wrong; so, it makes sense to try and implement/correct what's missing.

salt/state.py
samilaine: To me it *appeared* to do that as well, however, I'm not really sure if it works as expected (I don't claim to have knowledge or even justified belief in this case).

When the code goes through a list of matches produced by an instance of `salt.minion.Matcher` in a for loop (`for item in _data`), each loop variable should be either a string naming a state or a dictionary stating a key-value pair or pairs to specify keyword parameters (e.g. `{ 'match': 'pcre' }`).

If that is really the case, in what situation the loop variable would have a value suitable for popping `env_key, inc_sls`? That'd require a line like `someenv:somestate`, which seems like an entry not suitable for the top file.

I'm not claiming that I've exhaustively *tested* every possible scenario, and I may have misunderstood the format of the top file, in which case there might very well be other kind of values than just the two I mentioned above.

But, should I've grasped it correctly (and I doubt I am, that is why sent the PR to beg some comments), lines from `env_key, inc_sls = item.popitem()` to `matches[env_key].append(inc_sls)` don't do what it appears to be trying to accomplish, instead a single line `matches[saltenv].append(item)` would then do what is desired ie. take whatever is in the dictionary and pass it forward untouched.

samilaine:
An attempt to incorporate ideas from the discussion above have been sent in #47094.

PR #46960, opened 2018-04-09
Added clone_images to OpenNebula template_clone kwargs

Sikerdebaard:
### What does this PR do?
Added clone_images to OpenNebula template_clone kwargs. If set to True the images attached to the template will be cloned as well.

Link to the OpenNebula docs: https://docs.opennebula.org/5.4/integration/system_interfaces/api.html#one-template-clone
It's the 4th param in the one.template.clone call which reads as follows: "true to clone the template plus any image defined in DISK. The new IMAGE_ID is set into each DISK."

Unknown

### Previous Behavior
Previously only the template would be cloned and the cloned template would be linked to the original image.

### New Behavior
With this code the option to clone the images along with the template in one RPC call has been added.

### Tests written?

No

### Commits signed with GPG?

No

### DIFF
--- a/salt/cloud/clouds/opennebula.py
+++ b/salt/cloud/clouds/opennebula.py
@@ -2387,6 +2387,7 @@ def template_clone(call=None, kwargs=None):
     name = kwargs.get('name', None)
     template_id = kwargs.get('template_id', None)
     template_name = kwargs.get('template_name', None)
+    clone_images = kwargs.get('clone_images', False)
 
     if name is None:
         raise SaltCloudSystemExit(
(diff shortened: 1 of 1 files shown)


gtmanfred (approved):

gtmanfred (changes requested):
Actually, this should be made against develop, since it would be a new feature, and it needs to be added to the documentation for opennebula.

Sikerdebaard:
@gtmanfred I've changed the base to develop. I'm not sure how to change the documentation.

gtmanfred:
Can you rebase your PR to develop, so that it only has your changes in it?

The documentation is just at the top of the opennebula.py file.

https://github.com/saltstack/salt/blob/develop/salt/cloud/clouds/opennebula.py#L2

and also here

https://github.com/saltstack/salt/blob/develop/doc/topics/cloud/opennebula.rst

If you could just reference the new option somewhere in there, that would do.

gtmanfred (approved):

Issue #46935, opened 2018-04-06
Salt not reading Azure subscription ID?


esell:
### Description of Issue/Question

I am trying to simply list the available images and/or deploy a VM to Azure using the `azurearm` driver but am running into the following error (snipped):

```
  [ERROR   ] Failed to get the output of 'azurearm.avail_images()': Parameter 'subscription_id' must be str.
Traceback (most recent call last):
  File "/usr/lib/python2.7/dist-packages/salt/cloud/__init__.py", line 874, in image_list
    data[alias][driver] = self.clouds[fun]()
  File "/usr/lib/python2.7/dist-packages/salt/cloud/clouds/azurearm.py", line 381, in avail_images
    compconn = get_conn(client_type='compute')
  File "/usr/lib/python2.7/dist-packages/salt/cloud/clouds/azurearm.py", line 323, in get_conn
    client_type=client_type, **conn_kwargs
  File "/usr/lib/python2.7/dist-packages/salt/utils/azurearm.py", line 176, in get_client
    base_url=cloud_env.endpoints.resource_manager,
  File "/usr/local/lib/python2.7/dist-packages/azure/mgmt/compute/compute_management_client.py", line 75, in __init__
    self.config = ComputeManagementClientConfiguration(credentials, subscription_id, base_url)
  File "/usr/local/lib/python2.7/dist-packages/azure/mgmt/compute/compute_management_client.py", line 41, in __init__
    raise TypeError("Parameter 'subscription_id' must be str.")
TypeError: Parameter 'subscription_id' must be str.
[DEBUG   ] LazyLoaded nested.output
azure:
    ----------
```

### Setup
My `/etc/salt/cloud.providers.d/azure.conf` looks like:

```
azure:
  driver: azurearm
  master: salt
  subscription_id: my_sub_id
  client_id: my_client_id
  secret: my_client_secret
  tenant: my_tenant_id
  location: westus
  resource_group: saltFun
  network: salt-Fun-vnet
```

### Steps to Reproduce Issue
* Install dev version of salt using bootstrap.sh script (`sh bootstrap-salt.sh -L -M -N git develop`)
* create `/etc/salt/cloud.providers.d/azure.conf`
* run `salt-cloud -l debug --list-images azure`

I did try the various quote tricks on the values in the config file, none of which seemed to do anything.

### Versions Report

Salt Version:
           Salt: 2018.3.0-300-ged8e94a

Dependency Versions:
           cffi: 1.11.5
[...]

PR #46963, opened 2018-04-09
convert azure sub id to byte string vs UTF

esell:
### What does this PR do?
convert azure subscription_id to byte string vs UTF

[46935](https://github.com/saltstack/salt/issues/46935)

### Previous Behavior
Using the azurearm driver would throw a "subscription_id must be a str" error

### New Behavior
Error no longer is thrown

### Tests written?

No

### Commits signed with GPG?

No

### DIFF
--- a/salt/utils/azurearm.py
+++ b/salt/utils/azurearm.py
@@ -125,7 +125,7 @@ def _determine_auth(**kwargs):
 
     subscription_id = salt.utils.stringutils.to_str(kwargs['subscription_id'])
 
-    return credentials, subscription_id.encode(), cloud_env
+    return credentials, subscription_id, cloud_env
 
 
 def get_client(client_type, **kwargs):
--- a/salt/utils/azurearm.py
+++ b/salt/utils/azurearm.py
@@ -124,7 +124,7 @@ def _determine_auth(**kwargs):
 
     subscription_id = kwargs['subscription_id']
 
-    return credentials, subscription_id, cloud_env
+    return credentials, subscription_id.encode(), cloud_env
 
 
 def get_client(client_type, **kwargs):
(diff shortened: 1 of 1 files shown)


salt/utils/azurearm.py
```diff
@@ -124,7 +124,7 @@ def _determine_auth(**kwargs):
 
     subscription_id = kwargs['subscription_id']
 
-    return credentials, subscription_id, cloud_env
+    return credentials, subscription_id.encode(), cloud_env
```
terminalmage: This doesn't seem correct, since the documentation for the subscription ID in the azure mgmt modules says that it's supposed to be a `str`. Encoding would give you a bytestring, meaning this would likely break on Python 3. If what we need is a `str` type, then what we should do is:

1. Revert line 127 back to its original value.
2. Change line 125 to `subscription_id = kwargs['subscription_id']`

You'd also need to import `salt.utils.stringutils`.

terminalmage (changes requested):

esell:
@terminalmage I had not updated the other files only because I thought that the one in `salt/utils` was the area that the main Azure client connection was initiated. From there, all of those other referenced files pulled from `salt/utils/azurearm.py`. The main issue, at least as it appears to me, is the act of passing the `subscription_id` value from salt over to the Azure lib(s). Depending on the path taken to fix this, I can certainly update the other files.

Thanks for the help with this. As I mentioned before, salt and python are not my world so there are things I'm likely doing wrong haha.

salt/utils/azurearm.py
terminalmage: Hey, sorry I just realized I mistyped in my earlier comment. It should have been "Change line 125 to `subscription_id = salt.utils.stringutils.to_str(kwargs['subscription_id'])`."

esell:
Ok, I've removed the `encode()` stuff and replaced it with the salt utility method, thanks for pointing that out. Does it make sense to run through the other files you listed and repeat the same process?

terminalmage (approved):

PR #46907, opened 2018-04-05
[develop] Merge forward from 2018.3 to develop

rallytime:
Conflicts:
  - doc/ref/proxy/all/salt.proxy.cimc.rst
  - doc/ref/proxy/all/salt.proxy.panos.rst
  - doc/topics/releases/2018.3.0.rst
  - salt/modules/cmdmod.py
  - salt/modules/state.py
  - salt/utils/schedule.py
  - tests/integration/modules/test_state.py

PR #46964, opened 2018-04-09
Add new utils arg to schedule init

rallytime:
This was missed in the last merge-forward and causes lint to fail.

Refs #46907


cachedout:
@rallytime It looks like lint failed again though?

rallytime:
Bah. I will fix.

rallytime:
This isn't the right fix. Closing.

PR #46990, opened 2018-04-10

rallytime:
Conflicts:
  - .kitchen.yml


.kitchen.yml
```diff
@@ -38,7 +38,7 @@ provisioner:
     - 139
   max_retries: 2
   remote_states:
-    name: git://github.com/saltstack/salt-jenkins.git
+    name: git://github.com/gtmanfred/salt-jenkins.git
```
gtmanfred: crap, this one should be saltstack

gtmanfred: https://github.com/saltstack/salt/pull/46991

rallytime: OK - just in the 2018.3 branch? Or a fix to 2017.7 is needed as well?

gtmanfred (approved):

PR #46995, opened 2018-04-10
[schedule][develop] Minor fixes and enhancements

s0undt3ch:
### What does this PR do?
The main fix of this PR is that, for a time element schedule, if the job did not run, we couldn't get the next start date to advance

### Tests written?

No

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/utils/schedule.py
+++ b/salt/utils/schedule.py
@@ -91,7 +91,12 @@ class Schedule(object):
             # it in a WeakValueDictionary-- which will remove the item if no one
             # references it-- this forces a reference while we return to the caller
             instance = object.__new__(cls)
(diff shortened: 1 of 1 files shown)


salt/utils/schedule.py
```diff
@@ -1320,10 +1327,15 @@ def _handle_until(job, data):
                     del data[item]
             run = False
 
+            if 'name' in data:
+                job_name = data['name']
+            else:
+                job_name = data['name'] = job
```
garethgreenaway: The call to `_handle_until` was updated to not pass `job` so wouldn't this fail?

garethgreenaway: Never mind.  Makes more sense when looking at the full code.

garethgreenaway (approved):

Issue #46968, opened 2018-04-09
win_servermanager state missing on 2018.3.0


vernondcole:
### Description of Issue/Question

win_servermanager state no longer works on the new Windows version.

### Setup

install the learn-salt environment on a workstation from https://github.com/vernondcole/learn-salt
as documented in README.md.

### Steps to Reproduce Issue

on the workstation...
```bash
vagrant up bevymaster
vagrant up win12
vagrant ssh bevymaster
```
on bevymaster...
```
sudo salt win12 state.apply windows.install_server_features
```
That fails with an 'invalid keyword argument' error -- which is a bug,
as the install_server_features.sls is an exact cut-and-paste from 
https://docs.saltstack.com/en/latest/ref/states/all/salt.states.win_servermanager.html#salt.states.win_servermanager.installed

So let's upgrade the Windows VM to the current version of Salt...

Log in to the win12 machine virtual display using username `vagrant` password `vagrant`.
In the `Server Manager` click on `Local Server`
click on `IE Enhanced Security Configuration` and turn it `off`.
Start `Internet Explorer` and download and install https://repo.saltstack.com/windows/Salt-Minion-2018.3.0-Py2-x86-Setup.exe

Start a Windows PowerShell session as Administrator...
```
PS C:\Users\vagrant> salt-call state.apply windows.install_server_features
[WARNING ] C:\salt\bin\lib\site-packages\salt\modules\win_update.py:92: DeprecationWarning: The 'win_update' module is being deprecated and will be removed in Salt Fluorine (Unreleased). Please use the 'win_wua' module instead.

[ERROR   ] State 'win_servermanager.installed' was not found in SLS 'windows.install_server_features'
Reason: 'win_servermanager' __virtual__ returned False

local:
----------
          ID: install_multiple_features
    Function: win_servermanager.installed
      Result: False
     Comment: State 'win_servermanager.installed' was not found in SLS 'windows.install_server_features'
              Reason: 'win_servermanager' __virtual__ returned False
     Changes:

Summary for local
------------
Succeeded: 0
Failed:    1
------------
Total states run:     1
Total run time:   0.000 ms
```
So the command which was valid, but buggy, in 2017.7 is missing in 2018.3.

### Versions Report

```
PS C:\Users\vagrant> salt-call --versions-report
Salt Version:
           Salt: 2018.3.0

Dependency Versions:
           cffi: 1.10.0
       cherrypy: 10.2.1
       dateutil: 2.6.1
[...]

PR #47007, opened 2018-04-11
Fix some issues with the win_servermanager module

twangboy:
### What does this PR do?
Puts some error checking in the `_pshell_json`  helper function
Use `cmd.run_all` instead of `cmd.shell`
Raise `CommandExecutionError`s when it finds a problem
Account for instances where there's an empty string returned that will cause the JSON load to fail
Add `try/except` blocks to handle the `CommandExecutionError`s

Fix the example in the docs for the state
Add `try/except` blocks to the state

https://github.com/saltstack/salt/issues/46968

### Tests written?
No

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/modules/win_servermanager.py
+++ b/salt/modules/win_servermanager.py
@@ -60,6 +60,7 @@ def _pshell_json(cmd, cwd=None):
     log.debug('PowerShell: %s', cmd)
     ret = __salt__['cmd.run_all'](cmd, shell='powershell', cwd=cwd)
 
+    ret.pop()
     if 'pid' in ret:
         del ret['pid']
 
@@ -123,10 +124,7 @@ def list_installed():
           '-ErrorAction SilentlyContinue ' \
           '-WarningAction SilentlyContinue ' \
           '| Select DisplayName,Name,Installed'
-    try:
-        features = _pshell_json(cmd)
-    except CommandExecutionError:
-        raise
+    features = _pshell_json(cmd)
 
     ret = {}
     for entry in features:
@@ -225,10 +223,7 @@ def install(feature, recurse=False, restart=False, source=None, exclude=None):
           .format(command, _cmd_quote(feature), management_tools,
                   '-IncludeAllSubFeature' if recurse else '',
                   '' if source is None else '-Source {0}'.format(source))
-    try:
-        out = _pshell_json(cmd)
-    except CommandExecutionError:
-        raise
+    out = _pshell_json(cmd)
 
     # Uninstall items in the exclude list
     # The Install-WindowsFeature command doesn't have the concept of an exclude
--- a/salt/modules/win_servermanager.py
+++ b/salt/modules/win_servermanager.py
@@ -60,7 +60,6 @@ def _pshell_json(cmd, cwd=None):
     log.debug('PowerShell: %s', cmd)
     ret = __salt__['cmd.run_all'](cmd, shell='powershell', cwd=cwd)
 
-    ret.pop()
     if 'pid' in ret:
         del ret['pid']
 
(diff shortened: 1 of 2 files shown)


salt/modules/win_servermanager.py
```diff
@@ -57,11 +58,29 @@ def _pshell_json(cmd, cwd=None):
     if 'convertto-json' not in cmd.lower():
         cmd = '{0} | ConvertTo-Json'.format(cmd)
     log.debug('PowerShell: %s', cmd)
-    ret = __salt__['cmd.shell'](cmd, shell='powershell', cwd=cwd)
+    ret = __salt__['cmd.run_all'](cmd, shell='powershell', cwd=cwd)
+
+    if 'pid' in ret:
+        del ret['pid']
```
rallytime: Can you explain why using a `del` here is needed? I would imagine a `pop` would be fine instead, but I might be missing something here.

salt/modules/win_servermanager.py
```diff
@@ -104,7 +123,10 @@ def list_installed():
           '-ErrorAction SilentlyContinue ' \
           '-WarningAction SilentlyContinue ' \
           '| Select DisplayName,Name,Installed'
-    features = _pshell_json(cmd)
+    try:
+        features = _pshell_json(cmd)
+    except CommandExecutionError:
```
rallytime: Can you explain the purpose of this? I don't know why catching and then raising the same exception would be needed here.

salt/modules/win_servermanager.py
```diff
@@ -203,7 +225,10 @@ def install(feature, recurse=False, restart=False, source=None, exclude=None):
           .format(command, _cmd_quote(feature), management_tools,
                   '-IncludeAllSubFeature' if recurse else '',
                   '' if source is None else '-Source {0}'.format(source))
-    out = _pshell_json(cmd)
+    try:
+        out = _pshell_json(cmd)
+    except CommandExecutionError:
```
rallytime: Same question here as above. :)

salt/states/win_servermanager.py
```diff
@@ -298,8 +299,11 @@ def removed(name, features=None, remove_payload=False, restart=False):
         return ret
 
     # Remove the features
-    status = __salt__['win_servermanager.remove'](
-        features, remove_payload=remove_payload, restart=restart)
+    try:
+        status = __salt__['win_servermanager.remove'](
+            features, remove_payload=remove_payload, restart=restart)
+    except CommandExecutionError:
+        raise
```
rallytime: And the same question as above on this one, too.

rallytime (changes requested):
Hiya @twangboy - I have some questions on this one. Can you take a look and address them?

salt/modules/win_servermanager.py
twangboy: del is supposed to be faster than pop... and I don't need to know the deleted value

salt/modules/win_servermanager.py
```diff
@@ -57,11 +58,30 @@ def _pshell_json(cmd, cwd=None):
     if 'convertto-json' not in cmd.lower():
         cmd = '{0} | ConvertTo-Json'.format(cmd)
     log.debug('PowerShell: %s', cmd)
-    ret = __salt__['cmd.shell'](cmd, shell='powershell', cwd=cwd)
+    ret = __salt__['cmd.run_all'](cmd, shell='powershell', cwd=cwd)
+
+    ret.pop()
```
skizunov: Did this `ret.pop` slip in by mistake? If not, would you explain what this is for?

twangboy: You know it did!

skizunov (approved):

salt/modules/win_servermanager.py
rallytime: Ah, right. Thanks.

rallytime (approved):

Issue #40146, opened 2017-03-20
Should file.patch state require a hash?


MikeH-Halo:
### Description of Issue/Question
The file.patch state requires a hash today. In my case I have patch file templates which are applied using certain criteria (grains, nodegroups, what have you). The hash is impossible to calculate beforehand, and I am currently setting the hash to zero just get it to run (even if the state fails). 

My suggestion is that if the hash is omitted then file.patch should succeed if the underlying patch command succeeds.

### Setup
(Please provide relevant configs and/or SLS files (Be sure to remove sensitive info).)

### Steps to Reproduce Issue
(Include debug logs if possible and relevant.)

### Versions Report
(Provided by running `salt --versions-report`. Please also mention any differences in master/minion versions.)

gtmanfred:
@terminalmage do you have any opinions about this?

I am pretty sure that file.patch can be set to verify=False, and the default behavior of all the other file states is to have a source_hash.

Also, most patches IMHO are going to be pulled down and just need to be verified, and not generated from templates, so they should have hash sums.

I disagree with the statement that just because it succeeds to be applied it shouldn't need to have a source hash.  We could make the same argument about package managers, just because they have to do more work, and get the source hash into the database file in the repository, doesn't mean we shouldn't still have gpg signatures.

Thanks,
Daniel

MikeH-Halo:
Just to be clear; I never made any statements that the hash isn't necessary. What I said was; if the hash _isn't provided_ then file.patch should succeed if the underlying patch command succeeds. That is something different.
However, if I can set verify=False then that suffices for me..

terminalmage:
@gtmanfred it looks like flie.patch has no ``verify`` argument, in fact the state will refuse to run if there is no hash.

gtmanfred:
Ahh, my mistake.

I thought file.patch has a skip_verify options, but it does not appear that it does.  We should totally add one.

Thanks,
Daniel

terminalmage:
The proper fix here, I think, would be to normalize the arguments in the ``file.patch`` state to make them consistent with ``file.managed``. That is, we would rename ``hash`` to ``source_hash`` (and deprecate ``hash``), and add a ``skip_verify``. We would also not require hashes for local files (or those served up via the Salt fileserver, i.e. those with ``salt://`` URIs).

@gtmanfred, @MikeH-Halo thoughts?

gtmanfred:
:+1: from me

terminalmage:
[...]

PR #47010, opened 2018-04-11
Rewrite file.patch state

terminalmage:
This is a full rewrite. Previously, this state required the user to provide the hash of the patched file, but this is no longer necessary as the state is now able to check whether or not the patch will apply cleanly without actually making any changes. The following is a summary of the changes made in this rewrite:

- The `hash` and `dry_run_first` arguments are now ignored and generate a warning in the state return if used.
- The path being patched can now be a directory or a file, instead of just a file.
- The patch file is now downloaded by calling `file.managed`. This expands the potential file sources to include remote files (with all the hash verification features provided by `file.managed`), and also adds templating support.
- Instead of blindly running the `file.patch` remote-execution function and then checking that the patched result matches a pre-provided result, the state will run a non-dry-run `file.patch` with all patch output redirected to a temp file. This does 2 things: 1) It keeps any changes from being made, and 2) creates a reject file (which is not done when `--dry-run` is used with `patch`). This initial `file.patch` will tell us whether or not the patch would apply cleanly. If it does not, then we attempt to reverse-apply the rejected hunks using `--dry-run`, and if they apply cleanly then we know that the patch has already been applied and can exit with a `True` result. This means that we only actually attempt to apply the patch if we know if will apply cleanly, and prevents us from leaving the file(s) being patched in a partially-patched state, improving greatly the idempotence of the state.

Resolves https://github.com/saltstack/salt/issues/40146


salt/states/file.py
```diff
@@ -5155,122 +5160,415 @@ def prepend(name,
...
+        if not isinstance(option, six.string_types):
+            option = six.text_type(option)
+
+        for blacklisted in ('-N', '--forward', '-r', '--reject-file',
+                            '-o', '--output'):
+            if option.startswith(blacklisted):
+                ret['comment'] = 'The \'{0}\' CLI option is not allowed'.format(
+                    blacklisted
+                )
+                return ret
```
cachedout: Maybe instead of returning straight away just collect bad options and then deliver the full list in the error? That way we better handle the case where there is more than one.

cachedout (commented):
This is very good. One minor nitpick.

salt/states/file.py
terminalmage: Good idea. Done, and rebased.

cachedout (approved):

aarnaud:
Badly this new `file.patch` generate errors in log 

```
Mar 22 20:14:33 elasticsearch-worker salt-minion[13232]: [ERROR   ] Command '['/usr/bin/patch', '-N', '-r', '/tmp/__salt.tmp.fdbyaflo', '-o', '/tmp/__salt.tmp.pt62h1c2', '-i', '/tmp/__salt.tmp.t3i7qtxv', '/usr/lib/python3/dist-packages/salt/transport/zeromq.py']' failed with return code: 1
Mar 22 20:14:33 elasticsearch-worker salt-minion[13232]: [ERROR   ] stdout: patching file /tmp/__salt.tmp.pt62h1c2 (read from /usr/lib/python3/dist-packages/salt/transport/zeromq.py)
Mar 22 20:14:33 elasticsearch-worker salt-minion[13232]: Reversed (or previously applied) patch detected!  Skipping patch.
Mar 22 20:14:33 elasticsearch-worker salt-minion[13232]: 2 out of 2 hunks ignored -- saving rejects to file /tmp/__salt.tmp.fdbyaflo
Mar 22 20:14:33 elasticsearch-worker salt-minion[13232]: [ERROR   ] retcode: 1
```

terminalmage:
You should probably open a bug report then, instead of leaving a comment in a PR that was merged a year ago.

PR #47014, opened 2018-04-11
Test is now also working with Python2

brejoc:
### What does this PR do?

Since assertCountEqual was Python>=3.2 only, we backported this
to Python2.

### Previous Behavior

Testing with Python3 would work and Python2 would fail.

### New Behavior

Testing with Python2 and Python3 would work.

### Commits signed with GPG?

Yes


dincamihai:
+1

rallytime (approved):

Issue #46977, opened 2018-04-09
[2018.3.0] Backwards compatibilty breaking change in 2018.3.0


gtmanfred:
### Description of Issue/Question

I know we reserve the right to do this, but there is a breaking change in `salt/state.py` that breaks using `state.highstate` on minions that have 2018.3.0 installed on them, if your master is 2017.7.

### Steps to Reproduce Issue
Salt master on 2017.7.5
Salt minion on 2018.3.0

```
==> /srv/salt/top.sls <==
base:
  '*':
    - test

==> /srv/salt/test.sls <==
test:
  test.succeed_without_changes
```

Run `salt-call state.apply` and it fails with 

```
Traceback (most recent call last):
  File "/usr/lib/python2.7/site-packages/salt/cli/caller.py", line 212, in call
    ret['return'] = func(*args, **kwargs)
  File "/usr/lib/python2.7/site-packages/salt/modules/state.py", line 749, in apply_
    return highstate(**kwargs)
  File "/usr/lib/python2.7/site-packages/salt/modules/state.py", line 1055, in highstate
    orchestration_jid=orchestration_jid)
  File "/usr/lib/python2.7/site-packages/salt/state.py", line 3839, in call_highstate
    matches = self.top_matches(top)
  File "/usr/lib/python2.7/site-packages/salt/state.py", line 3333, in top_matches
    for saltenv in ext_matches:
TypeError: 'bool' object is not iterable
```

### Versions Report
```
[root@salt salt]# salt --versions-report
Salt Version:
           Salt: 2017.7.5

Dependency Versions:
           cffi: Not Installed
       cherrypy: unknown
       dateutil: Not Installed
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.7.2
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.5.1
   mysql-python: Not Installed
      pycparser: Not Installed
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: Not Installed
         Python: 2.7.5 (default, Aug  4 2017, 00:39:18)
   python-gnupg: Not Installed
         PyYAML: 3.11
          PyZMQ: 15.3.0
           RAET: Not Installed
          smmap: Not Installed
        timelib: Not Installed
        Tornado: 4.2.1
[...]

PR #47038, opened 2018-04-12
[2018.3] fix to fileclient.py

garethgreenaway:
### What does this PR do?
The _ext_nodes master function has been renamed to _master_tops. To ensure compatibility when using older Salt masters we continue to pass the function as _ext_nodes until the Magnesium release.

#46977

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/fileclient.py
+++ b/salt/fileclient.py
@@ -1391,7 +1391,17 @@ class RemoteClient(Client):
         '''
         Return the metadata derived from the master_tops system
         '''
-        load = {'cmd': '_master_tops',
+        salt.utils.versions.warn_until(
+            'Magnesium',
+            'The _ext_nodes master function has '
+            'been renamed to _master_tops. To ensure '
+            'compatibility when using older Salt masters '
+            'we continue to pass the function as _ext_nodes.'
+        )
+
+        # TODO: Change back to _master_tops
+        # for Magnesium release
+        load = {'cmd': '_ext_nodes',
                 'id': self.opts['id'],
                 'opts': self.opts}
         if self.auth:
--- a/salt/version.py
+++ b/salt/version.py
@@ -101,9 +101,8 @@ class SaltStackVersion(object):
         'Fluorine'      : (MAX_SIZE - 100, 0),
         'Neon'          : (MAX_SIZE - 99, 0),
         'Sodium'        : (MAX_SIZE - 98, 0),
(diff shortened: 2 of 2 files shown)


rallytime:
I am not familiar with this part of the code, or why that function was renamed, but do we _need_ to rename it? I think that providing a warning is good, but we'll still have this problem with older minions down the road. Is there a way we can alias the function names?

garethgreenaway:
@rallytime The function is aliased on the master side, so older minions will still be able to function without issue.  The issue popped up when a newer minion connects to an older minion, new minion is expecting that the master side function is named one thing but on the older master it's named something else.  If we pass the older name in the newer minion it will use the alias, until we can deprecate the function name in later versions where we no longer need to support those versions expecting that particular name.

rallytime:
@garethgreenaway Ah, OK. Thanks for the explanation. That makes more sense.

There's some test failure related to this it looks like. `Magnesium` doesn't exist yet. You'll need to uncomment it [here](https://github.com/saltstack/salt/blob/develop/salt/version.py#L106 ) and move the `pylint: disable` line.

(Looks like that file isn't quite right either. There are 2 Sodiums and there should only be one. Can you remove the commented out Sodium too while you're in there?)

gtmanfred:
Would it make more sense to just catch an error if the new function doesn't exist, and use the old function?  Then we don't ever have to deprecate it?

@thatch45 is Magnesium a long enough time to document that this will break backwards compatibility? for newer minions with older masters?

terminalmage (approved):
The change to `master_tops` only affects cases where the minion is a newer release than the master, which is not a supported use of Salt. That said, in the interest of reducing errors, I think this is OK. Magnesium may even be excessive, since we only support going back 2 full release cycles.

gtmanfred (approved):
ok, cool :+1:

thatch45:
The only reason I say Magnesium is because this does not really hurt anything to keep it in there for a little longer, if we can make deprecation windows large enough that they should not bother anyone then we should do it that way.

rallytime:
@garethgreenaway Can you make the changes above about making the `Magnesium` variable available so we can get this in? Seems like all of the discussion has been resolved. :)

rallytime (changes requested):
The `Magnesium` variable needs to be exposed in `version.py`.

rallytime (approved):

PR #47040, opened 2018-04-12
Add ability to run a shell command to depends decorator

terminalmage:
This adds the ability to suppress loading of a function based on the
retcode of a command.

### DIFF
--- a/salt/utils/decorators/__init__.py
+++ b/salt/utils/decorators/__init__.py
@@ -5,8 +5,10 @@ Helpful decorators for module writing
 
 # Import python libs
 from __future__ import absolute_import, print_function, unicode_literals
+import errno
 import inspect
 import logging
+import subprocess
 import sys
 import time
 from functools import wraps
@@ -28,7 +30,7 @@ class Depends(object):
     '''
     This decorator will check the module when it is loaded and check that the
     dependencies passed in are in the globals of the module. If not, it will
-    cause the function to be unloaded (or replaced)
+    cause the function to be unloaded (or replaced).
     '''
     # kind -> Dependency -> list of things that depend on it
     dependency_dict = defaultdict(lambda: defaultdict(dict))
@@ -40,6 +42,8 @@ class Depends(object):
 
         An example use of this would be:
 
+        .. code-block:: python
+
             @depends('modulename')
             def test():
                 return 'foo'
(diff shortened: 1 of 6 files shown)


salt/utils/decorators/__init__.py
```diff
@@ -49,14 +53,36 @@ def test():
             @depends('modulename', fallback_function=function)
             def test():
                 return 'foo'
-        '''
 
+        .. code-block:: python
+
+        This can also be done with the retcode of a command, using the
+        `retcode` argument:
```
rallytime: This is _very_ picky, but 
```
`retcode`
```
Needs to be:
```
``retcode``
```
In order for the docs to render that as a code section. RST needs double backticks, while markdown only needs single backticks.

salt/utils/decorators/__init__.py
```diff
@@ -49,14 +53,36 @@ def test():
...
+
+            @depends('/opt/bin/check_cmd', retcode=0)
+            def test():
+                return 'foo'
+
+        It is also possible to check for any nonzero retcode using the
+        `nonzero_retcode` argument:
```
rallytime: Same doc comment here.

tests/integration/files/file/base/_modules/runtests_decorators.py
```diff
@@ -90,3 +98,63 @@ def missing_depends_will_fallback():
     ret = {'ret': True,
            'time': time.time()}
     return ret
+
+
+@salt.utils.decorators.depends(_exit_code(42), retcode=42)
+def command_success_retcode():
+    '''
+    CLI Example:
```
rallytime: I think these docs aren't quite right. :)

Can you fill this doc section out better?

Same thing is happening on all of the new functions in this file.

rallytime (changes requested):
This is really nice. I just have a couple of minor nitpicks. Nice tests, too!

tests/integration/files/file/base/_modules/runtests_decorators.py
terminalmage: All of the functions in that file have the same docs, I copied and pasted from them. I believe that the only reason they look the way they do is so that they pass the `test_valid_docs` test.

salt/utils/decorators/__init__.py
terminalmage: Fair enough, though I'm pretty certain we don't build docs for this page. However, if we do start doing so, we should at least have it correct.

tests/integration/files/file/base/_modules/runtests_decorators.py
terminalmage: I updated the `test_valid_docs` test to ignore all tests in `runtests_decorators.py` and removed all the docstrings in that module.

rallytime (approved):

rallytime:
@terminalmage This is failing one of the tests you wrote: 

- integration.modules.test_decorators.DecoratorTest.test_missing_depends

https://jenkins.saltstack.com/job/PR/job/salt-pr-linode-cent7-py3/4043/

terminalmage:
@rallytime This was actually one of the old tests, it just failed because of the removal of the docstrings in the previous commit. Should've tried running tests before I pushed last time!

Test should pass now.

rallytime:
re-run py

Issue #47043, opened 2018-04-12
lsattr slowing down archive.extracted


xBytez:
### Description of Issue/Question
When using `archive.extracted` with user and group set, it loops through every file checking it's attributes, this is both ineffecient and slow. It makes my CPU usage peak and my states takes a solid 10 minutes to run.

What I'd like:
- An option to turn the lsattr stuff off and just enforce the chown.
- Rewrite the lsattr stuff to be more efficient (example: `lsattr -R /var/folder`)

### Setup
(Please provide relevant configs and/or SLS files (Be sure to remove sensitive info).)
```
  archive.extracted:
    - name: /var/folder
    - source: s3://bucket/filename
    - skip_verify: True
    - user: user
    - group: user
```

### Steps to Reproduce Issue
(Include debug logs if possible and relevant.)
```
2018-04-12 23:11:01,923 [salt.loaded.int.module.cmdmod:282 ][DEBUG   ][21519] output: --------------e---- /var/folder/node_modules/resolve-pkg/package.json
2018-04-12 23:11:02,108 [salt.loaded.int.module.cmdmod:385 ][INFO    ][21519] Executing command [u'lsattr', u'/var/folder/node_modules/resolve-pkg/index.js'] in directory '/root'
2018-04-12 23:11:02,517 [salt.loaded.int.module.cmdmod:282 ][DEBUG   ][21519] output: --------------e---- /var/folder/node_modules/resolve-pkg/index.js
2018-04-12 23:11:02,519 [salt.loaded.int.module.cmdmod:385 ][INFO    ][21519] Executing command [u'lsattr', u'/var/folder/node_modules/engine.io-client/History.md'] in directory '/root'
2018-04-12 23:11:03,133 [salt.loaded.int.module.cmdmod:282 ][DEBUG   ][21519] output: --------------e---- /var/folder/node_modules/engine.io-client/History.md
2018-04-12 23:11:03,135 [salt.loaded.int.module.cmdmod:385 ][INFO    ][21519] Executing command [u'lsattr', u'/var/folder/node_modules/engine.io-client/engine.io.js'] in directory '/root'
2018-04-12 23:11:03,704 [salt.loaded.int.module.cmdmod:282 ][DEBUG   ][21519] output: --------------e---- /var/folder/node_modules/engine.io-client/engine.io.js
[...]

PR #47058, opened 2018-04-13
Fix calls to file.lsattr when lsattr is not installed

terminalmage:
This was partially fixed before 2018.3.0, but some careless programming since then has caused problems to return. This fixes those by making `file.lsattr` return `None` when lsattr is not installed, and then fixes all references to take this into account.


rallytime (approved):

rallytime:
Refs #47043

Issue #46979, opened 2018-04-09
2018 breaks environment configuration parameter


onlyanegg:
### Description of Issue/Question
The [release notes](https://docs.saltstack.com/en/latest/topics/releases/2018.3.0.html#environment-config-option-renamed-to-saltenv) say that the `environment` configuration parameter would not be deprecated, so I was surprised when running a state outside the base environment failed. Running the state with the saltenv command line parameter succeeded, but running with out it failed.

### Setup
Minion config:
```
master:
  - mgmt101.pod1.local.ca.tyler.com

verify_master_pubkey_sign: true
mine_functions:
  network.interfaces: []
environment: "common"
file_client: "remote"
file_roots:
  common:
    - /srv/salt/common/states
    - /srv/salt/common/files
fileserver_backend:
  - roots
  - git
gitfs_remotes:
    - https://github.com/certain-ops/salt-formula.git
    - https://github.com/certain-ops/zookeeper-formula.git
pillar_roots:
  common:
    - /srv/salt/common/pillar
pillar_merge_lists: true
state_output: "changes"
log_level: "debug"
return: "rawfile_json"
rawfile_json.filename: "/data02/logs/minion_events"
gitfs_saltenv: [{"common": [{"ref": "master"}]}]
top_file_merging_strategy: "same"
hadoop: {"driver": "zookeeper", "hosts": ["if101.pod1.local.ca.tyler.com"], "memcached.host": "mgmt101.pod1.local.ca.tyler.com"}
consul_sdb: {"host": "mgmt101.pod1.local.ca.tyler.com"}
docker_swarm: {"memcached.host": "mgmt101.pod1.local.ca.tyler.com"}
```

### Steps to Reproduce Issue
```
[root@if101.pod1.local minion.d]# salt-call state.apply salt.minion
[DEBUG   ] Configuration file path: /etc/salt/minion
[WARNING ] Insecure logging configuration detected! Sensitive data may be logged.
[DEBUG   ] Grains refresh requested. Refreshing grains.
[DEBUG   ] Missing configuration file: /etc/salt/minion
[DEBUG   ] Including configuration from '/etc/salt/minion.d/_schedule.conf'
[DEBUG   ] Reading configuration from /etc/salt/minion.d/_schedule.conf
[DEBUG   ] Including configuration from '/etc/salt/minion.d/beacons.conf'
[DEBUG   ] Reading configuration from /etc/salt/minion.d/beacons.conf
[...]

PR #47066, opened 2018-04-13
Fix regression in handling of environment/saltenv

terminalmage:
This was broken sometime between when the change was made and 2018.3.0 was released. This fixes the issue and adds a regression test.

Fixes https://github.com/saltstack/salt/issues/46979.


cachedout (approved):
Good catch here.

Issue #26396, opened 2015-08-17
'Smart' pillar_source_merging_strategy does not use aggregate for yamlex


carlpett:
I'm trying to use `#!yamlex` and `!aggregate` in some of my pillar files. As I understand the [documentation](http://docs.saltstack.com/en/latest/ref/configuration/master.html#pillar-source-merging-strategy) of `pillar_source_merging_strategy`, the default 'smart' strategy should then switch to the aggregate when it finds a yamlex renderer. However, it doesn't. If I set the merge strategy to aggregate in the master config, it does work.

I've created files like this:

```
pillar/top.sls
'test-node':
  - merging.testbase
  - merging.test1
  - merging.test2
```

```
pillar/merging/testbase.sls
#!yamlex
foo: !aggregate baz
foo: !aggregate base
```

```
pillar/merging/test1.sls
#!yamlex
foo: !aggregate bar
```

```
pillar/merging/test2.sls
#!yamlex
foo: !aggregate qux
```

```
# salt-call --local pillar.items foo
local:
    ----------
    foo:
        - qux
```

Am I misunderstanding the documentation when I think that the smart strategy should switch to aggregate here, or is this a bug?

```
# salt --versions-report
           Salt: 2015.5.3
         Python: 2.7.5 (default, Jun 17 2014, 18:11:42)
         Jinja2: 2.8
       M2Crypto: 0.21.1
 msgpack-python: 0.4.2
   msgpack-pure: Not Installed
       pycrypto: 2.6.1
        libnacl: Not Installed
         PyYAML: 3.10
          ioflo: Not Installed
          PyZMQ: 14.3.1
           RAET: Not Installed
            ZMQ: 3.2.4
           Mako: Not Installed
        Tornado: Not Installed
```

jfindlay:
@carlpett, thanks for the report.

mchugh19:
This still seems to be the case.

```
# salt --versions
Salt Version:
           Salt: 2016.11.4

Dependency Versions:
           cffi: 1.6.0
       cherrypy: unknown
       dateutil: 2.6.0
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.7.2
        libgit2: 0.21.4
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.4.8
   mysql-python: Not Installed
[...]

PR #47084, opened 2018-04-15
GH-26396, add a clarification on pillar merge strategy

mattp-:
### What does this PR do?
make it clear what is required for yamlex behavior cross-sls.

doc patch only.

### DIFF
--- a/doc/ref/configuration/master.rst
+++ b/doc/ref/configuration/master.rst
@@ -4500,6 +4500,11 @@ strategy between different sources. It accepts 5 values:
 
   Guesses the best strategy based on the "renderer" setting.
 
+.. note::
+    In order for yamlex based features such as ``!aggregate`` to work as expected
+    across documents using the default ``smart`` merge strategy, the :conf_master:`renderer`
+    config option must be set to ``jinja|yamlex`` or similar.
+
 .. conf_master:: pillar_merge_lists
 
 ``pillar_merge_lists``


doc/ref/configuration/master.rst
```diff
@@ -4500,6 +4500,11 @@ strategy between different sources. It accepts 5 values:
 
   Guesses the best strategy based on the "renderer" setting.
 
+.. note::
+    In order for yamlex based features such as ``!aggregate`` to work as expected
+    across documents using the default ``smart`` merge strategy, the :conf_master:`renderer`
+    config option must be set to ``yamlex_jinja`` or similar.
```
terminalmage: While this is technically correct, this is a long-outdated method of specifying multiple renderers. We've used a pipe syntax for a while now (the old syntax is translated in `salt/template.py`), so our documentation should suggest using a renderer of `jinja|yamlex`.

For that matter, we should probably change the default renderer in `salt/config/__init__.py` from `yaml_jinja` to `jinja|yaml`, and make the same updates both `doc/ref/configuration/master.rst` and in in the config stubs (`conf/minion`, `conf/master`, et al).

Having said all of that, I'm fine with merging this as-is and making all of the rest of the changes in a future PR.

terminalmage (approved):

doc/ref/configuration/master.rst
thatch45: I agree!

terminalmage: @mattp- if you would like to change the example to `jinja|yamlex` then I can make all of the other changes I noted above.

terminalmage: Otherwise we can just merge this and I can update this reference myself.

mattp-: @terminalmage was going to say I got incepted by the dictupdate code, but I just checked again:
```
105     ¦   if renderer == 'yamlex' or renderer.startswith('yamlex_'):
```
so it seems you actually do need the legacy yamlex_jinja as it stands :)
I'm not actually sure it even makes sense to be relying on the default opts['renderer'] attribute here, but trying to bubble up the pillar/state shebang from the template for each sls seemed ... not easy, as an alternative. any thoughts on how to proceed? guessing renderer.contains('yamlex').

mattp-: also, happy to update the other doc references once decision on ^ is confirmed

terminalmage: @mattp- I took care of the other doc references and also rewrote a lot of the renderer docs in #47111. As for the dictupdate code, the proper way to support templated yamlex would be to split on a pipe and confirm that the last element is `yamlex`, which I've also done in #47111. We would still need to keep the second half of the if clause as-is since we need to support the legacy syntax.

If you update line 4506 here to use `jinja|yamlex` instead of `yamlex_jinja`, then that should be all that is needed.

terminalmage (changes requested):

mattp-:
@terminalmage thanks for doing the hard bits 👍 pushed the s/yamlex_jinja/jinja\|yamlex/ change on my commit.

terminalmage (approved):

Issue #46959, opened 2018-04-09
Intermittent problems with boto states - Probably adding "Name" tag too fast


jryberg:
### Description of Issue/Question
I have huge problems with boto and to create resources in Amazon. I'm currently creating one VPC and a couple of subnets. 

The state works, the VPC and subnets are rarely created without problem but it's normally one resource that fails to be created. 

The resource are actually created but it seems boto tries to set the Name tag too fast, before the API can return object id and the newly created object becomes orphaned without Name tag, example a subnet with cidr, assigned to the VPC but nameless.

It's also not possible to recover since boto detects that the cidr already exist and since it has no Name tag it don't understand it was created by boto.

This is a typical error message:
```
----------
          ID: Ensure labvpc01 subnet Test 1 exists
    Function: boto_vpc.subnet_present
        Name: labvpc01
      Result: False
     Comment: Failed to create subnet: Bad Request: The subnet ID 'subnet-123456' does not exist
     Started: 12:34:10.170200
    Duration: 332.45 ms
     Changes:
----------
```
If I check AWS console I can see the subnet id without a Name.

Not only subnets fails this way, sometimes the VPC is created but boto fails to set Name tag and the entire state fails and it's unrecoverable.

#### State
**vpc.sls**
```
{% set vpc_pillar_name = "vpc_configurations" -%}
{%- if salt['pillar.get'](vpc_pillar_name) -%}
  # Create VPC
  {%- set vpc = salt['pillar.get'](vpc_pillar_name, {}) -%}
  {%- for vpc, settings in vpc.items()|sort %}

Ensure VPC {{ vpc }} exists:
  boto_vpc.present:
    - name: {{ vpc }}
    - cidr_block: {{ settings['cidr'] }}
    - dns_hostnames: True
    - dns_support: True

    # Create subnets
    {%- set subnets = settings['subnets'] %}
    {% for subnet, subnetsettings in subnets.items()|sort %}
Ensure {{ vpc }} subnet {{ subnet }} exists:
  boto_vpc.subnet_present:
    - name: {{ subnet }}
    - vpc_name: {{ vpc }}
    - cidr_block: {{ subnetsettings['cidr'] }}
    - availability_zone: {{ subnetsettings['az'] }}
    {%- endfor %}
  {%- endfor -%}
{%- endif -%}
```
#### Pillar
```
vpc.keyid: [[ REDACTED ]]
vpc.key: [[ REDACTED ]]
vpc.region: eu-west-1

[...]

PR #47095, opened 2018-04-16
Wait until resource exists before attempting to add a Name tag

j-flowers:
### What does this PR do?
Added a wait until resource is ready when adding a Name tag directly after creating the resource. 
Due to Amazon API using eventual consistency model. When a salt state is applied it attempts to add a Tag Name and Value to a resource that has not been created. The state errors out and does not apply the tag.  

https://github.com/saltstack/salt/issues/46959
### Previous Behavior
Salt would create the resource, but fail on adding the Name tag, leaving the subnet without any Name tag. This would break any dependancies that reference the subnet by name.  

### New Behavior
Using  resource_exists method to check if the resource exists, if false sleep 5 seconds.  

### Tests written?

No

### Commits signed with GPG?

No


cachedout:
Hi @jflowers84. It appears you have submitted this against a different branch than the one your local copy was branched from. I am going to close this. Could you please re-submit against the correct branch? Thanks.

PR #47104, opened 2018-04-16
yamlloader: Properly handle colons in inline dicts

### LINKED ISSUE #46929: 2018.3 regression file.managed.context parsing

With 2018.3 release, there is regression in file.managed states WRT  [found unexpected ':'](http://pyyaml.org/wiki/YAMLColonInFlowContext) issue.  
Sorry I have no logs available.

### Setup

These values have embedded Colon" and file.managed context parsing fails.
```
etcd:
  service:
    initial_cluster: etcd0=https://127.0.0.1:2380
    initial_cluster_state: new
    initial_cluster_token: etcd-cluster-1
    initial_advertise_peer_urls: https://127.0.0.1:2380
    listen_peer_urls: https://127.0.0.1:2380
    listen_client_urls: https://127.0.0.1:2379
    advertise_client_urls: https://127.0.0.1:2379


..

  file.managed:
    - template: jinja
    - context:
      etcd: {{ etcd }}
```

**Workaround:**

```
    - context:
      etcd: {{ etcd|json }}
```

### Steps to Reproduce Issue

A colleague verbally reported same issue with **different formula** so looks like regression.

Reverting to 2017.7 stable resolved any problems.

### Versions Report

2018.3  (not available, I have downgraded to 2017.7).

terminalmage:
PyYAML's scanner is not written to handle Python's unicode literal strings, since they are not part of the YAML spec. With all strings loaded in the YAML context now being unicode types, this causes errors when scanning the document to construct the next node. The unicode string literal is misidentified as a plain scalar and any colons within it are interpreted as delimiting a mapping node. Since no whitespace appears after the colon, this causes scanning error.

To fix it, we check for this scanning error and reset the scanner's position and check if the first char is the "u" prefix. If so, and if the following char is a quote, then we attempt to scan for a string literal instead.

Resolves #46929.

### DIFF
--- a/salt/utils/yamlloader.py
+++ b/salt/utils/yamlloader.py
@@ -127,6 +127,47 @@ class SaltYamlSafeLoader(yaml.SafeLoader):
         value = self.construct_scalar(node)
         return salt.utils.stringutils.to_unicode(value)
 
+    def fetch_plain(self):
+        '''
+        Handle unicode literal strings which appear inline in the YAML
+        '''
+        orig_line = self.line
+        orig_column = self.column
+        orig_pointer = self.pointer
+        try:
+            return super(SaltYamlSafeLoader, self).fetch_plain()
+        except yaml.scanner.ScannerError as exc:
+            problem_line = self.line
+            problem_column = self.column
+            problem_pointer = self.pointer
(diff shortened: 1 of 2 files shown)


rallytime (approved):
Nice

PR #47106, opened 2018-04-16
Tornado50 compatibility fixes

### LINKED ISSUE #45790: Test with Tornado 5.0b1

DmitryKuzmenko:
### What does this PR do?
This is not a full moving to the new Tornado 5.0 approach which corner idea for our design is to not pass io_loop instance everywhere and use the current one everywhere it's possible.
Here I've just adopted the Salt code to the new API dropped `io_loop` kwarg. This approach uses `salt.utils.async.current_ioloop` in cases where the code can be possibly run with another instance of the io_loop. In some cases where the current io_loop is == expected io_loop I've just dropped the kwarg.

Anyway it seems we have to rework our io_loop management in the future in `develop` branch. I've started this work initially but got that it doesn't look like a minor compatibility fix...

Fixes #45790

### Previous Behavior
io_loop kwarg is not supported by Tornado 5.0

### New Behavior
Should work good with Tronado 5.0

### Tests written?
No

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/engines/ircbot.py
+++ b/salt/engines/ircbot.py
@@ -94,14 +94,15 @@ class IRCClient(object):
         self.allow_nicks = allow_nicks
         self.disable_query = disable_query
         self.io_loop = tornado.ioloop.IOLoop(make_current=False)
+        self.io_loop.make_current()
         self._connect()
 
     def _connect(self):
         _sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM, 0)
         if self.ssl is True:
-            self._stream = tornado.iostream.SSLIOStream(_sock, ssl_options={'cert_reqs': ssl.CERT_NONE}, io_loop=self.io_loop)
+            self._stream = tornado.iostream.SSLIOStream(_sock, ssl_options={'cert_reqs': ssl.CERT_NONE})
         else:
-            self._stream = tornado.iostream.IOStream(_sock, io_loop=self.io_loop)
+            self._stream = tornado.iostream.IOStream(_sock)
         self._stream.set_close_callback(self.on_closed)
         self._stream.connect((self.host, self.port), self.on_connect)
 
--- a/requirements/dev_python27.txt
+++ b/requirements/dev_python27.txt
@@ -1,4 +1,4 @@
--r base.txt
+-r base-py2.txt
 
 mock>=2.0.0
 apache-libcloud>=0.14.0
@@ -6,6 +6,6 @@ boto>=2.32.1
 boto3>=1.2.1
 moto>=0.3.6
 SaltPyLint>=v2017.3.6
-pytest
+pytest>=3.5.0
 git+https://github.com/eisensheng/pytest-catchlog.git@develop#egg=Pytest-catchlog
 git+https://github.com/saltstack/pytest-salt.git@master#egg=pytest-salt
--- a/requirements/dev_python34.txt
+++ b/requirements/dev_python34.txt
@@ -1,4 +1,4 @@
--r base.txt
+-r base-py3.txt
 
 mock>=2.0.0
 apache-libcloud>=0.14.0
@@ -11,5 +11,5 @@ moto>=0.3.6
 # prevent it from being successfully installed (at least on Python 3.4).
 httpretty
 SaltPyLint>=v2017.2.29
-pytest
+pytest>=3.5.0
 git+https://github.com/saltstack/pytest-salt.git@master#egg=pytest-salt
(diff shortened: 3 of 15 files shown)


salt/engines/ircbot.py
```diff
@@ -94,14 +94,15 @@ def __init__(self, nick, host, port=6667, username=None, password=None, channels
         self.allow_nicks = allow_nicks
         self.disable_query = disable_query
         self.io_loop = tornado.ioloop.IOLoop(make_current=False)
+        self.io_loop.make_current()
```
terminalmage: I'm not familiar with `make_current`, but is there a reason why we are instantiating the `IOLoop` instance and then starting it as a separate action?

DmitryKuzmenko: if `make_current` is `True` and there is one current IOLoop it raises an exception. But here we want to have an own one because it's an engine that is an independent process.
I should probably have a comment for this in the code.

shish:
Using this patch, things *mostly* work, but I'm getting this error after a lot of commands (ie, the actual work happens successfully, then this happens during shutdown):

```
  File "/usr/bin/salt", line 10, in <module>
    salt_main()
  File "/usr/lib/python3/dist-packages/salt/scripts.py", line 476, in salt_main
    client.run()
  File "/usr/lib/python3/dist-packages/salt/cli/salt.py", line 173, in run
    for full_ret in cmd_func(**kwargs):
  File "/usr/lib/python3/dist-packages/salt/client/__init__.py", line 825, in cmd_cli
    self.event.close_pub()
  File "/usr/lib/python3/dist-packages/salt/utils/event.py", line 390, in close_pub
    self.subscriber.close()
  File "/usr/lib/python3/dist-packages/salt/transport/ipc.py", line 753, in close
    self._read_stream_future.exc_info()
AttributeError: '_asyncio.Future' object has no attribute 'exc_info'
```

salt/engines/ircbot.py
terminalmage: Yes, that would be helpful for future people that maintain this code, to prevent regressions. Thanks for the explanation!

DmitryKuzmenko:
Sorry, I see no related failures but I'm not familiar how stable our test suite at this moment? @rallytime can you please take a look?

DmitryKuzmenko:
@shish wow. Thank you! I'll check

rallytime:
@DmitryKuzmenko I think the beacons tests here are related:

- unit.test_minion.MinionTestCase.test_beacons_before_connect
- unit.test_minion.MinionTestCase.test_scheduler_before_connect

https://jenkins.saltstack.com/job/PR/job/salt-pr-linode-ubuntu14-n/21884/

But I only see those failing on that build. I'll start these again to see if anything changes.

rallytime:
re-run py

abednarik:
Hello

 Tested this one on FreeBSD, which uses Tornado 5.0.2 and it worked great! Salt is broken in FreeBSB, hope this one is merged and release ASAP.

Thanks

s0undt3ch:
Made some minimal changes to the failing unit tests just because they forced us to run as root. But they pass with tornado 5.0.2

rallytime:
The tests look good here now and I don't think that adding those comments should block merging this PR, but should be added later for sure for clarity.

gtmanfred:
@shish can you double check if you are still seeing that error with the newest changes?

Thanks,
Daniel

gtmanfred (approved):

DmitryKuzmenko:
@s0undt3ch thank you for help!

s0undt3ch:
I didn't do much :)

shish:
@gtmanfred even without the latest version of the patch, the error I was seeing last week has mysteriously fixed itself, so everything seems to be working now...

gtmanfred:
@shish perfect :+1: from me then.

Working on getting tornado >= 5.0.0 tests running in jenkins branch tests right now.

DmitryKuzmenko:
@gtmanfred awesome! Thank you man!

cachedout (approved):

Issue #46932, opened 2018-04-06
2018.3.0 reg.present catching wrong exception when vdata is a list.


whytewolf:
### Description of Issue/Question

with 2018.3.0 the `salt.utils.stringutils.to_unicode` function was changed to throw `TypeError` instead of `UnicodeEncodeError` the reg state is still looking for `UnicodeEncodeError` so when a list is added as as vdata. instead of running it's code to handle it. the `TypeError` trace gets output to the user. and the registery is never updated.

### Setup
Try this state in 2018.3.0 on windows
```
test_reg:
  reg.present:
    - name: HKEY_LOCAL_MACHINE\SOFTWARE\salttest\salty\salted
    - vname: Test
    - REG_MULTI_SZ
    - vdata: ['test'.'test2']
```

### Steps to Reproduce Issue
Run the above state in 2018.3.0

result:
```
local:
---------
              ID: test_reg 
   Function: reg.present 
        Name: HKEY_LOCAL_MACHINE\SOFTWARE\salttest\salty\salted 
       Result: False 
 Comment: An exception occurred in this state: Traceback (most recent call last): 
                     File "c:\salt\bin\lib\site-packages\salt\state.py", line 1878, in call 
                       **cdata['kwargs']) 
                     File "c:\salt\bin\lib\site-packages\salt\loader.py", line 1823, in wrapper 
                       return f(*args, **kwargs) 
                     File "c:\salt\bin\lib\site-packages\salt\states\reg.py", line 196, in present 
                       vdata_decoded = salt.utils.stringutils.to_unicode(vdata, 'utf-8') 
                     File "c:\salt\bin\lib\site-packages\salt\utils\stringutils.py", line 99, in to_unicode 
                       raise TypeError('expected str, bytes, or bytearray') 
                     TypeError: expected str, bytes, or bytearray 
```

Expected result: a new registry REG_MULTI_SZ registry item inserted. 

### Versions Report
```
Salt Version:
           Salt: 2018.3.0

Dependency Versions:
           cffi: 1.10.0
       cherrypy: 10.2.1
       dateutil: 2.6.1
      docker-py: Not Installed
          gitdb: 2.0.3
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.10
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: 1.0.7
   msgpack-pure: Not Installed
 msgpack-python: 0.4.8
[...]

PR #47107, opened 2018-04-16
Fix issues with reg state, add tests

twangboy:
### What does this PR do?
Fixes some issues with the reg state and salt util
Fixes issue when `vdata` is a list and you're passing `vtype: REG_MULTI_SZ` by creating a `cast_vdata` function in `salt.utils.win_reg` for handling the different data types
Fixes an issue with binary data and unicode in the same function. Unicode data gets additional unwanted characters when written to the registry as binary. It is now decoded first.
Adds some tests for `reg.present`.

https://github.com/saltstack/salt/issues/46932

### Tests written?
Yes

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/states/reg.py
+++ b/salt/states/reg.py
@@ -192,11 +192,8 @@ def present(name,
                                    salt.utils.stringutils.to_unicode(name, 'utf-8'))
         return ret
 
-    try:
-        vdata_decoded = salt.utils.stringutils.to_unicode(vdata, 'utf-8')
-    except UnicodeDecodeError:
-        # vdata contains binary data that can't be decoded
-        vdata_decoded = vdata
+    vdata_decoded = __utils__['reg.cast_vdata'](vdata=vdata, vtype=vtype)
+
     add_change = {'Key': r'{0}\{1}'.format(hive, key),
                   'Entry': '{0}'.format(salt.utils.stringutils.to_unicode(vname, 'utf-8') if vname else '(Default)'),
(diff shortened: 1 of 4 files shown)


twangboy:
Need to test this on PY3 to see how the encoding thing effects the test

twangboy:
Fixed Binary test for Py3

dwoz (approved):

gtmanfred (approved):

hrumph:
Thanks for fixing this @Twangboy. I just noticed this today and as about to report an issue about this.

hrumph:
One other thing. Doesn't the documentation for reg.present need a fix. It still says __:param str vdata:__, but shouldn't it acutally say something like __:param (str, int, list) vdata:__?

hrumph:
When I modified the reg module I wrote the following documentation:
```
 :param object vdata: The value data to be set.
        What the type of this parameter
        should be is determined by the value of the vtype
        parameter. The correspondence
        is as follows:
        .. glossary::
           REG_BINARY
               binary data (i.e. str in python version < 3 and bytes in version >=3)
           REG_DWORD
               int
           REG_EXPAND_SZ
               str
           REG_MULTI_SZ
               list of objects of type str
           REG_SZ
               str
    :param str vtype: The value type.
        The possible values of the vtype parameter are indicated
        above in the description of the vdata parameter.
```

twangboy:
@hrumph The above PR updates the docs

PR #47114, opened 2018-04-17
[schedule][develop] Also advance `_next_scheduled_fire_time`

s0undt3ch:
### Tests written?

No

### Commits signed with GPG?

Yes


s0undt3ch:
Not the right fix

PR #47115, opened 2018-04-17
salt-cloud/vmware if specified use ssh_host to deploy

### LINKED ISSUE #36543: Salt Cloud / vmware: Allow 'ssh_host' in VM profile

In my situation, I'm using Salt Cloud with VMWare to create VMs in vCenter. Some VMs need to have two NICs defined, but due to business reasons, only the IP on "Network adapter 1" (eth0) will be functional on first boot. Unfortunately, Salt Cloud (or its vmware module) simply queries vCenter for the VM's IPs and takes the first one it gets back. For whatever reason, this is consistenly "Network adapter 2" (eth1), which isn't functional yet. Salt Cloud then tries to connect to the not-connectable IP to do the deploy, and eventually times out and fails.
If this process would pass along the 'ssh_host' value for Salt Cloud to eventually connect to, I could get around this whole problem.
### Example

Here's an example dictionary of what I'd _like_ to be able to pass in to CloudClient.

```
        vmdict = {
            'names': [vmname],
            'provider': myprovider,
            'clonefrom': mytemplate.name,
            'cluster': mycluster,
            'datacenter': mydatacenter,
            'datastore': myname,
            'ssh_host': eth0_IP,
            'ssh_username': mytemplate.username,
            'password': mytemplate.password,
            etc, etc,
        }
```

The ssh_host value would be new. @nmadhok said this is something he might be able to look at.
### Versions Report

salt-cloud 2016.3.2 (Boron)

cvedel:
Fixes #36543

### What does this PR do?
In the salt-cloud vmware provider, use ssh_host when specified to deploy salt.

#36543

### Previous Behavior
Deploy salt through the first discovered IP address.

### New Behavior
If ssh_host is defined in the profile, prefer it to the discovered IP address.

### Tests written?

No

### Commits signed with GPG?

No

### DIFF
--- a/salt/cloud/clouds/vmware.py
+++ b/salt/cloud/clouds/vmware.py
@@ -2804,7 +2804,8 @@ def create(vm_):
             if deploy:
                 vm_['key_filename'] = key_filename
                 # if specified, prefer ssh_host to the discovered ip address
-                vm_['ssh_host'] = vm_.get('ssh_host', ip)
+                if 'ssh_host' not in vm_:
+                    vm_['ssh_host'] = ip
                 log.info("[ {0} ] Deploying to {1}".format(vm_name, vm_['ssh_host']))
 
                 out = __utils__['cloud.bootstrap'](vm_, __opts__)
--- a/salt/cloud/clouds/vmware.py
+++ b/salt/cloud/clouds/vmware.py
@@ -2803,7 +2803,9 @@ def create(vm_):
             # ssh or smb using ip and install salt only if deploy is True
             if deploy:
                 vm_['key_filename'] = key_filename
-                vm_['ssh_host'] = ip
+                # if specified, prefer ssh_host to the discovered ip address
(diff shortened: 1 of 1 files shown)


salt/cloud/clouds/vmware.py
```diff
@@ -2803,7 +2803,9 @@ def create(vm_):
             # ssh or smb using ip and install salt only if deploy is True
             if deploy:
                 vm_['key_filename'] = key_filename
-                vm_['ssh_host'] = ip
+                # if specified, prefer ssh_host to the discovered ip address
+                vm_['ssh_host'] = vm_.get('ssh_host', ip)
```
gtmanfred: why not just do
```
if 'ssh_host' not in vm_:
    vm_['ssh_host'] = ip
```

cvedel: Indeed, changed. Thanks

gtmanfred (approved):

PR #44723, opened 2017-11-28
Implements #9183

blhsing:
Support glob pattern in file.absent state as requested in issue #9183.

Currently file.absent can only remove one specific file or directory at a time. This PR will enable file.absent to remove all files matching a given glob pattern.

The way I implemented it is with a decorator that wraps any existing function in states/file.py and makes it accept an addition keyword argument of resolve_glob such that when it is set True, the decorator will treat the given file name as a glob pattern and calls the wrapped function for each file that matches the glob pattern, and for the returning dict, the result will only be True if every call to the wrapped function returns result as True, and changes, pchanges and comment will all be made into dicts indexed by matching file names, with values set to those of the corresponding values returned by the individual calls to the wrapped function, so that the changes and comments for each matching file can be individually tracked in case they're different. And if resolve_glob is set False or if no file is found, the decorator will simply call the wrapped function directly for its original behavior.

Issue #9183, opened 2013-12-11
Feature Request : Support Glob pattern in file.absent state


JeremieCharest:
The file.absent state (and some others) need an absolute path to work. It will be nice if they can support glob or regex patterns.

As a side note; the file module contains a lot more interesting stuff than is state counterpart. Do you have plan to expose more of these methods in the file state?

KumarRajnish:
All,

Due to a misconfiguration on our Office365 tenant, all email addressed to your Ruckus Wireless email address (you@ruckuswireless.com) were delivered to a catch all mailbox.  This incident occurred between 11 PM to 8 AM (PST).  IT is delivering these emails with a “Reply-all” rule to all individuals on the email chain.  If you feel you are missing email attachments, please send email to helpdesk@ruckuswireless.com.

We apologize for any inconvenience this may have caused.

Regards,

IT

blhsing:
My team was in need of this exact same feature so I went ahead to implement it and am now seeking to push it into the develop branch as part of a possible future release.

The way I implemented it is with a decorator that wraps any existing function in states/file.py and makes it accept an addition keyword argument of _resolve_glob_ such that when it is set True, the decorator will treat the given file name as a glob pattern and calls the wrapped function for each file that matches the glob pattern, and for the returning dict, the _result_ will only be True if every call to the wrapped function returns _result_ as True, and _changes_, _pchanges_ and _comment_ will all be made into dicts indexed by matching file names, with values set to those of the corresponding values returned by the individual calls to the wrapped function, so that the changes and comments for each matching file can be individually tracked in case they're different. And if _resolve_glob_ is set False or if no file is found, the decorator will simply call the wrapped function directly for its original behavior.

Let me know if I can go forward with pushing my changes. Thanks.

waynew:
Looks like a couple of PRs failed to actually make their way to completion, I'm going to reopen this issue, since it looks like the functionality is still absent... from file.absent :trollface:

waynew:
bad bot!

heini:
BTW: As of 2019.2.0, there's `file.tidied`, which may do what you want. Was inspired by Puppet's `tidy`.

heini:
Ping.

sharkbruhaha:
Can we please get globbing/regex enabled for this already?

dmurphy18:
@JeremieCharest Closing this due to age, the old version of Salt and Python 2.
[...]

PR #47132, opened 2018-04-17
Support glob pattern in file.absent state

ixs:
This is a rework of the older PR #44723. The original author never
finished the PR and the feature is actually worthwhile.

Changed glob import
Added simple test-case.

### What does this PR do?
Add globbing support to the file.absent state.

#9183

### DIFF
--- a/salt/states/file.py
+++ b/salt/states/file.py
@@ -268,6 +268,7 @@ For example:
 from __future__ import absolute_import, print_function, unicode_literals
 import copy
 import difflib
+import glob
 import itertools
 import logging
 import os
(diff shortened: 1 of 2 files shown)


ixs:
As a general note: I find your process of closing PRs when they haven't been worked on a while a bit suboptimal as this seems to also close the referenced issue.
This is kinda unexpected as the issue actually hasn't been fixed.

Maybe there's a better way of dealing with such a situation?

cachedout:
@ixs You are totally right. This was a mistake on our part. We'll find a better way to handle that situation.

terminalmage (changes requested):
Can you please do this without the decorator? For one, I don't see the benefit of using one here when it's only being used on the one function. Secondly, using the decorator negates the docstring and will remove all documentation for this state when docs are built.

ixs:
@terminalmage Just to clarify: You would prefer _not_ using the decorator but instead you would like the globbing support folded into the main absent state function?

terminalmage:
Correct, that logic can be placed near the top of the function.

rallytime:
@ixs Were you able to come back to this?

ixs:
@rallytime I've been traveling/doing the conferenc-circuit the last four weeks. Give me a bit of time to get back to my own timezone and I'll look at implementing the requested changes.

rallytime:
Hey @ixs - I'm going to close this for now until you've had a chance to come back here. :)

When you're ready to incorporate the feedback, this will also need a rebase as there are merge-conflicts in `file.py`.

Let us know when you're ready and we will be very happy to open this up again. Or, if it is easier for you, feel free to open a new PR and link it to think one.

Thanks again!

PR #47140, opened 2018-04-17
[Bugfix] Add proper (and explicit) ordering to s3fs

basepi:
### What does this PR do?

Adds proper ordering to s3fs so that salt will always use the version of a file from the first listed bucket with that file (as in gitfs and other fileservers).

### Previous Behavior

Previously, while processing the buckets, s3fs would use dictionaries, which are unordered in python 2. This could result in buckets being searched in unexpected order.

### New Behavior

The internal data structure now uses lists of dictionaries to make sure order is maintained, and buckets will properly be searched in their defined order.

### Tests written?

No

### Commits signed with GPG?

Yes

----------

I consider this a serious bug in s3fs and think this should be backported to all supported versions. However, it's not critical for me as we have already included this updated version [in hubble](https://github.com/hubblestack/hubble/pull/355).

### DIFF
--- a/salt/fileserver/s3fs.py
+++ b/salt/fileserver/s3fs.py
@@ -57,6 +57,25 @@ structure::
 
 .. note:: This fileserver back-end requires the use of the MD5 hashing algorithm.
     MD5 may not be compliant with all security policies.
+
+.. note:: This fileserver back-end is only compatible with MD5 ETag hashes in
+    the S3 metadata. This means that you must use SSE-S3 or plaintext for
+    bucket encryption, and that you must not use multipart upload when
+    uploading to your bucket. More information here:
+    https://docs.aws.amazon.com/AmazonS3/latest/API/RESTCommonResponseHeaders.html
+
+    Objects without an MD5 ETag will be fetched on every fileserver update.
+
+    If you deal with objects greater than 8MB, then you should use the
+    following AWS CLI config to avoid mutipart upload:
+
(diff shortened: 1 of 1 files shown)


basepi:
I'm going to add another doc note around the config you have to use when uploading to your s3 bucket. If you have large files and accidentally use multipart upload, the cache checking will break and it will always download your large files (because this module assumes the etag is md5, and it is not when multipart is used). I'll put that in later today.

basepi:
This is ready for review and merge.

salt/fileserver/s3fs.py
```diff
@@ -57,6 +57,25 @@
...
+    Objects without an MD5 ETag will be fetched on every fileserver update.
+
+    If you deal with objects greater than 8MB, then you should use the
+    following AWS CLI config to avoid mutipart upload:
+
+    .. code-block::
+
+        s3 =
+          multipart_threshold = 1024MB
+
```
terminalmage: Was the line break and indentation after the `=` intentional? Just making sure as I am not familiar with this syntax.

basepi: I think so. You can see example config here: https://docs.aws.amazon.com/cli/latest/topic/s3-config.html

I just realized that I accidentally used the same link twice -- the second instance should be that link above. I'll fix that momentarily.

terminalmage (approved):

PR #47143, opened 2018-04-18
pillar.netbox: Rewrite of Netbox external pillar

avolmensky:
@ksofa2 @mirceaulinic This is my version of the Netbox external pillar. I had already completed 90% of it before realizing that @ksofa2 has written his own. Thinking we can combine the best features of both? I'm using pynetbox so I don't have to worry about different Netbox API versions and also to decrypt secrets from Netbox.

### What does this PR do?

External pillar for Netbox. Adds device information, interfaces and IP addresses from Netbox to the pillar. Will also get device login credentials from Netbox if they exit.

### Tests written?

No

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/pillar/netbox.py
+++ b/salt/pillar/netbox.py
@@ -51,6 +51,13 @@ except ImportError:
     HAS_PYNETBOX = False
     log.error('pynetbox must be installed')
 
+try:
+    import netaddr
+    HAS_NETADDR = True
+except ImportError:
+    HAS_NETADDR = False
+    log.error('netaddr must be installed')
+
 # Pull the following from Netbox
 DEVICE_FIELDS = [
     'id',
@@ -76,9 +83,19 @@ INTERFACE_FIELDS = [
     'lag'
 ]
 
+IP_FIELDS = [
+    'id',
+    'address',
+    'role',
+    'description',
+    'tenant',
+    'vrf.name',
+    'vrf.rd'
+]
+
 
 def __virtual__():
-    if HAS_PYNETBOX:
+    if HAS_PYNETBOX and HAS_NETADDR:
         return True
     return False
 
@@ -103,13 +120,6 @@ def ext_pillar(minion_id,
         # a napalm driver is assigned to the device platform then add it to salt
         device_info['netbox'] = {}
         if hasattr(device, 'primary_ip') and device.status.label == 'Active' and napalm_driver:
-            for device_field in DEVICE_FIELDS:
-                device_info['netbox'][device_field] = getattr(device, device_field)
-                if isinstance(
-                            device_info['netbox'][device_field],
-                            (pynetbox.ipam.IpAddresses, pynetbox.lib.response.Record)
-                            ):
-                    device_info['netbox'][device_field] = str(device_info['netbox'][device_field])
 
             # If proxy_config enabled in config, then add proxy pillar information
             if proxy_config:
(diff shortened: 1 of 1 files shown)


gtmanfred (commented):
If this changes the pillar output of the netbox pillar, we should go through a deprecation process of the old netbox pillar, and probably name this rewrite netboxng.py, and use `__utils__['versions.warn_until']()` to warn that the other external pillar is deprecated.

avolmensky:
@gtmanfred The other version of the netbox pillar is has only just been added to develop and is not included in a released version, do we still have to go through the depreciation process?

gtmanfred (approved):
Oh, ok, then if it is not in a release yet, :+1:

cachedout:
Let's keep this on hold until @mirceaulinic and @ksofa2 can weigh in.

cachedout:
@ksofa2 Did you see the ping above?

ksofa2:
@cachedout Thanks for the reminder. I've been swamped, will look soon.

(Sorry @avolmensky)

salt/pillar/netbox.py
```diff
@@ -1,129 +1,198 @@
...
-    api_token = kwargs.get('api_token', None)
-    proxy_username = kwargs.get('proxy_username', None)
-    proxy_return = kwargs.get('proxy_return', True)
-
+    nb_api = pynetbox.api(url, token=token, private_key_file=keyfile)
+    device = nb_api.dcim.devices.get(name=minion_id)
+    device_info = {}
+    if device:
+        device.platform.full_details()
+        napalm_driver = device.platform.napalm_driver
```
ksofa2: I think these should be wrapped in a `try` and set napalm_driver to None on exception. This will handle the case of platform being None. (The NetBox interface doesn't require platform, and in my test case I hadn't set it, so I got an exception here.)

kris-steinhoff (commented):
I sat down with @ngrundler (who's done most of the work with our Salt and NetBox integrations) to look at this in detail today.

We like the idea of using pynetbox instead of requests to fetch the data from the API.

We're also OK with changing the structure of the data returned, but I have some concerns about where that work is done. Our original design simply passed through data from the API, and we intentionally held off on adding other structures like interfaces because we wanted to have a discussion with the community about how best to do that.

I suggest that this would be best in pynetbox.

One reason for this is that if someone were to build an Ansible Dynamic Inventory to do something similar, it would be great to have the same data structures to make templates for a device config much more portable.

avolmensky:
@ksofa2 I agree that the structure of the data returned has to change. I have been thinking it would be better to have this data returned in the OpenConfig format. Once this information is in OpenConfig we could use napalm_yang configure an interface:

https://docs.saltstack.com/en/latest/ref/states/all/salt.states.netyang.html

```
{%- set expected_config =  pillar.get('openconfig_interfaces_cfg') -%}
interfaces_config:
  napalm_yang.managed:
    - data: {{ expected_config | json }}
    - models:
      - models.openconfig_interfaces
    - debug: true
```

@mirceaulinic thoughts?

ngrundler:
@avolmensky you're thinking along the same lines @ksofa2 and I are, but we weren't sure whether or not it made more sense to see if netbox should return the yang model, or if salt should compile the data.  Were you thinking of using something like pyangbind to build/validate the model based on the data returned by netbox?  I was playing around with this idea a few months ago, and it was a not insignificant amount of time to generate the OpenConfig models for 48 individual interfaces, so I was concerned about scaling this to 4000+ devices without totally killing the salt master.  

In principal I really like the idea, though.  How would you propose handling data that isn't currently modeled in OpenConfig?  One use case we have to work around is an upcoming data center deployment using EVPN/VXLAN, so we need some way to configure the ESI values on the interfaces for dual-attached hosts.

avolmensky:
Hey @ngrundler, I haven't played around with building the yang model yet. And I have the same problem with data not modelled in OpenConfig or stored in Netbox, I don't really have a solution there either. Maybe store that data in a git pillar or something? Custom fields on an interface in Netbox would possibly fix this. If you could chime in on https://github.com/digitalocean/netbox/issues/2055 it would give some weight to the issue. 

I agree with @ksofa2 that this would probably be best done within pynetbox

zachmoody:
:wave: Hello folks, @avolmensky asked if I could take a look at this PR as it related to pynetbox. Unless I'm misunderstanding something, having pynetbox return openconfig data structures _seems_ a little too specific for something like pynetbox, but it's entirely possible I'm missing something. Feel free to open up an issue over there and we can talk about it.

I do think it's a great idea for a package of its own though. There's probably enough to do in getting data from NetBox (via pynetbox) and into openconfig to warrant it.

rallytime:
@avolmensky and @ksofa2 Do you have anything to add concerning @zachmoody's comment above? It seems like this PR may not be the best approach as discussed above. Please let us know what you'd like to do with this one. :)

ngrundler:
Hey @avolmensky, @ksofa2 and I have been chatting on and off again about this, I think we're in agreement that we should be using pynetbox for this module, but we just haven't had time to refactor things.  We're getting ready to deploy and manage a new datacenter using a combination of Salt/NetBox, and our focus lately has just been on getting things to work.  My hope is that once that effort is complete we'll be able to sit down and get this in shape; part of the problem is in order to build the configuration templates in a fully automated way, I had to extend what we'd already written to include additional calls to the API to pull in things like IP addresses/VLAN tags assigned to interfaces, connected devices, and the IP addresses assigned to the remote interface of the connected devices, etc.  Rewriting this in a way that is useful to others is absolutely our goal here, so I'm hopeful that we will have some cycles to devote to this in early July.

rallytime:
@ngrundler Thank you so much for the update and explanation! I think that sounds like a good plan. I am going to close this for now, and we will be happy to open it again when you three are ready to come back to this. Or, feel free to open a new PR. Whichever is easier for you. :)

Issue #45939, opened 2018-02-09
user.present with hash_password: True detects change on every state.apply/highstate


andygabby:
### Description of Issue/Question
user.present with  hash_password: True detects change on every highstate. The password is successfully set to what was intended, but upon subsequent highstates, salt always thinks it needs to change the root password again.

### Setup
```
# salt/root_pw.sls
root:
  user.present:
    - password: {{ salt['pillar.get']('root_pw') }}
    - hash_password: True
```

```
# pillar/root_pw.sls
root_pw: supersecret
```

### Steps to Reproduce Issue
Run state multiple times against same node. The password is set successfully without error, but always detects a change is necessary.

### Versions Report
Minions and masters all running same version. Tested on 2017.7.2 and 2016.11.2

```
Salt Version:
           Salt: 2017.7.2
 
Dependency Versions:
           cffi: Not Installed
       cherrypy: Not Installed
       dateutil: 1.5
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.7.2
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.5.1
   mysql-python: Not Installed
      pycparser: Not Installed
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: Not Installed
         Python: 2.7.5 (default, Aug  4 2017, 00:39:18)
   python-gnupg: Not Installed
         PyYAML: 3.11
          PyZMQ: 15.3.0
           RAET: Not Installed
          smmap: Not Installed
        timelib: Not Installed
        Tornado: 4.2.1
            ZMQ: 4.1.4
 
System Versions:
           dist: centos 7.4.1708 Core
         locale: UTF-8
        machine: x86_64
        release: 3.10.0-693.17.1.el7.x86_64
         system: Linux
        version: CentOS Linux 7.4.1708 Core
```
Also not working on:
```
Salt Version:
           Salt: 2016.11.2
 
Dependency Versions:
           cffi: 1.6.0
       cherrypy: Not Installed
       dateutil: 1.5
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.7.2
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: 0.21.1
           Mako: Not Installed
[...]

PR #47147, opened 2018-04-18
states.user.present: Make usage of `hash_password` idempotent

eliasp:
### What does this PR do?

It fixes the non-idempotent behavior of `states.user.present` when `hash_password: True` described in #45939 
It targets `2018.3`, but due to it's trivial nature, it should be easily cherry-picked for backports to older releases.

#45939

### Previous Behavior
Every time `states.user.present` with `hash_password: True` was used, a new shadow hash was generated based on a new randomly generated salt value for hashing which made this state's behavior idempotent.

### New Behavior
Before hashing a password, it checks now whether an existing salt can be retrieved.
If yes, it will be re-used for generating the hash.
If not, it defaults to using a new randomly generated salt.

### Tests written?

No

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/states/user.py
+++ b/salt/states/user.py
@@ -438,7 +438,24 @@ def present(name,
     # hash_password is True, then hash it.
     if password and hash_password:
         log.debug('Hashing a clear text password')
-        password = __salt__['shadow.gen_password'](password)
+        # in case a password is already set, it will contain a Salt
+        # which should be re-used to generate the new hash, other-
+        # wise the Salt will be generated randomly, causing the
+        # hash to change each time and thereby making the
+        # user.present state non-idempotent.
+        algorithms = {
+            '1':  'md5',
+            '2a': 'blowfish',
+            '5':  'sha256',
+            '6':  'sha512',
+        }
+        try:
+            _, algo, shadow_salt, shadow_hash = __salt__['shadow.info'](name)['passwd'].split('$', 4)
+            log.debug('Re-using existing shadow salt for hashing password using {}'.format(algorithms.get(algo)))
+            password = __salt__['shadow.gen_password'](password, crypt_salt=shadow_salt, algorithm=algorithms.get(algo))
+        except ValueError:
+            log.info('No existing shadow salt found, defaulting to a randomly generated new one')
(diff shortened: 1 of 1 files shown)


eliasp:
Those failures aren't related to this PR:
- [jenkins/PR/salt-pr-linode-cent7-py3#4162](https://jenkins.saltstack.com/job/PR/job/salt-pr-linode-cent7-py3/4162/)
- [jenkins/PR/salt-pr-linode-ubuntu16-py3#9000](https://jenkins.saltstack.com/job/PR/job/salt-pr-linode-ubuntu16-py3/9000/)

cachedout:
I totally see the case for this but would love to have some tests written around this behavior. Any chance you'd be willing to write some?

cachedout:
@eliasp Did you see my comment above regarding tests?

eliasp:
Sorry, I initially planned on adding some tests, but won't have time right now to complete this.
Feel free to take it from here for now:
- needs tests
- when the result of hashing the salt+password differs, re-generate using a new random hash as suggested by @tigpas in https://github.com/saltstack/salt/issues/45939#issuecomment-382813691

cachedout (approved):

salt/states/user.py
```diff
@@ -438,7 +438,24 @@ def present(name,
...
     if password and hash_password:
         log.debug('Hashing a clear text password')
-        password = __salt__['shadow.gen_password'](password)
+        # in case a password is already set, it will contain a Salt
+        # which should be re-used to generate the new hash, other-
+        # wise the Salt will be generated randomly, causing the
+        # hash to change each time and thereby making the
+        # user.present state non-idempotent.
+        algorithms = {
+            '1':  'md5',
```
isbm: But this is insecure. I think we should either kill this option or at least scream all around the place in logs if anyone will use this.

isbm (changes requested):

salt/states/user.py
eliasp: I do agree that "screaming all around the place in logs" is the most reasonable reaction here…

Added a corresponding warning.

isbm (approved):

PR #47158, opened 2018-04-18
Change check_db_exists default to True

rallytime:
This option was added in PR #46895 where the default in the `run_file` function was set to `True`, but in the `run` function it was set to `False`. This is an error and the default should be `True`.

With the default set to False, it causes a test failure and would also be a change in behavior from previous releases. In addition, the docs also say that default is True. Therefore, this looks to be a typo. This PR sets the default to `True`.

Fixes the following test failure:

- unit.states.test_mysql_query.MysqlQueryTestCase.test_run

### DIFF
--- a/salt/states/mysql_query.py
+++ b/salt/states/mysql_query.py
@@ -221,7 +221,7 @@ def run(name,
         grain=None,
         key=None,
         overwrite=True,
-        check_db_exists=False,
+        check_db_exists=True,
         **connection_args):
     '''
     Execute an arbitrary query on the specified database


cachedout:
It would be good to get insight from @oldmantaiter here.

oldmantaiter:
@rallytime @cachedout I think it was definitely in error that it was different for each of the functions, sorry about that.

rallytime:
@oldmantaiter All good! We missed it initially, too! :D

PR #47159, opened 2018-04-18
Fix for whitelist/blacklist checking for non-list iterables

terminalmage:
A recent PY3-compatibility fix was made to `check_whitelist_blacklist`.  The function was previously checking for the `__iter__` attribute to check if the blacklist and/or whitelist was a sequence of expressions or a single expression. However, the `str` type has the `__iter__` attribute in PY3 while it does not in PY2. To fix this, the check was changed to see if the blacklist and/or whitelist were `list` types. This however causes problems in `salt.daemons.masterapi` when we pass `set` types.

To fix this, we just check if the whitelist and/or blacklist are string types, rather than that they are _not_ some sort of sequence type.

### DIFF
--- a/salt/utils/stringutils.py
+++ b/salt/utils/stringutils.py
@@ -364,6 +364,12 @@ def check_whitelist_blacklist(value, whitelist=None, blacklist=None):
     if blacklist is not None:
         if isinstance(blacklist, six.string_types):
             blacklist = [blacklist]
+        if not hasattr(blacklist, '__iter__'):
+            raise TypeError(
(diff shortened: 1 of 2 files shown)


garethgreenaway (approved):

rallytime:
@terminalmage There are some test failures here: https://jenkins.saltstack.com/job/PR/job/salt-pr-linode-cent7-py3/4479/

terminalmage:
re-run all

terminalmage:
OK, test failures should be fixed now. The tests exposed a flaw in my fix, which I addressed and for which I added additional asserts in the new unit test added in this PR.

PR #47191, opened 2018-04-19
salt-ssh: Do not attempt to match host/ip to minion ID if reverse lookup fails

### LINKED ISSUE #47150: [Regression] ip_to_host and SSH._expand_target require missing reverse-lookup
Hey folks, we got an interesting problem. We fixed it by temporarily updating our DNS entries but this won't last long.

So, this problem occurred after upgrading to 2018.3.0.

This is what we saw at the first glance

```
salt:/srv # salt-ssh minionname state.sls deploy.system+root......
AssertionError: 
  File "/usr/bin/salt-ssh", line 11, in <module>
    sys.exit(salt_ssh())
  File "/usr/lib/python2.7/site-packages/salt/scripts.py", line 425, in salt_ssh
    client.run()
  File "/usr/lib/python2.7/site-packages/salt/cli/ssh.py", line 23, in run
  File "/usr/lib/python2.7/site-packages/salt/client/ssh/__init__.py", line 235, in __init__
    self.tgt_type)
  File "/usr/lib/python2.7/site-packages/salt/roster/__init__.py", line 105, in targets
  File "/usr/lib/python2.7/site-packages/salt/roster/flat.py", line 49, in targets
    rmatcher = RosterMatcher(conditioned_raw, tgt, tgt_type, 'ipv4')
  File "/usr/lib/python2.7/site-packages/salt/roster/flat.py", line 58, in __init__
    assert tgt
AssertionError
  File "/usr/bin/salt-ssh", line 11, in <module>
    sys.exit(salt_ssh())
  File "/usr/lib/python2.7/site-packages/salt/scripts.py", line 425, in salt_ssh
    client.run()
  File "/usr/lib/python2.7/site-packages/salt/cli/ssh.py", line 23, in run
  File "/usr/lib/python2.7/site-packages/salt/client/ssh/__init__.py", line 235, in __init__
    self.tgt_type)
  File "/usr/lib/python2.7/site-packages/salt/roster/__init__.py", line 105, in targets
  File "/usr/lib/python2.7/site-packages/salt/roster/flat.py", line 49, in targets
    rmatcher = RosterMatcher(conditioned_raw, tgt, tgt_type, 'ipv4')
  File "/usr/lib/python2.7/site-packages/salt/roster/flat.py", line 58, in __init__
    assert tgt
AssertionError
```

We assert-debugged salt and found that the issue at hand is caused by the two following functions:

https://github.com/saltstack/salt/blob/68aeb62a950666/salt/client/ssh/__init__.py#L339 SSH._expand_target
```python
...
            hostname = salt.utils.network.ip_to_host(hostname)  # which returns None
...
                            if hostname != self.opts['tgt']:  # None != minionname
                                self.opts['tgt'] = hostname  # overrides valid target name
...
```

https://github.com/saltstack/salt/blob/68aeb62a/salt/utils/network.py#L214 ip_to_host
```python
    try:
        hostname, aliaslist, ipaddrlist = socket.gethostbyaddr(ip)
    except Exception:
        hostname = None
    return hostname
```

### Why does gethostbyaddr fail?
Because the minion "minionname" hadn't a reverse DNS lookup (IP to Name). Temporary adding it fixed it for now but it won't work for long as those are dynamic IP addresses.

### Can you ping the host?
Yep, we can even ssh to minionname perfectly fine.

### Why does the reverse-lookup matter for salt?
That is this issue about. We cannot even use blank IP addresses. They fail as well.

### Last question: can we have `ip_to_host` reraising it's original exception?
That would be great for debugging next time.
Because as you can see the catching assert is far away from the original place where None was introduced into the variables. And backtracing took a while to figure out.

terminalmage:
When the reverse lookup fails, `salt.utils.network.ip_to_host()` returns `None`. Later, when we attempt to look through the roster(s) for a minion ID matching the target's host, a roster entry without a `host` key will mean that `roster_data.get('host')` evaluates as None. This causes the target expression to be expanded as `None`, when what we should do is simply carry on without attempting to find a match in the roster(s). This commit does just that.

Fixes #47150.

### DIFF
--- a/salt/client/ssh/__init__.py
+++ b/salt/client/ssh/__init__.py
@@ -353,6 +353,9 @@ class SSH(object):
         needs_expansion = '*' not in hostname and salt.utils.network.is_reachable_host(hostname)
         if needs_expansion:
             hostname = salt.utils.network.ip_to_host(hostname)
+            if hostname is None:
+                # Reverse lookup failed
+                return
             self._get_roster()
             for roster_filename in self.__parsed_rosters:
                 roster_data = self.__parsed_rosters[roster_filename]
--- a/salt/utils/network.py
+++ b/salt/utils/network.py
@@ -218,7 +218,8 @@ def ip_to_host(ip):
     '''
     try:
         hostname, aliaslist, ipaddrlist = socket.gethostbyaddr(ip)
-    except Exception:
+    except Exception as exc:
+        log.debug('salt.utils.network.ip_to_host(%r) failed: %s', ip, exc)
         hostname = None
     return hostname
 


gtmanfred (approved):

isbm:
@terminalmage great catch, thanks!

PR #47102, opened 2018-04-16
dont allow using no_use_wheel for pip 10.0.0 or newer

gtmanfred:
### What does this PR do?
The --no-use-wheel flag has been deprecated in pip 10.0.0

### Tests written?

Yes, there is a test that uses this right now, which is failing, and will stop failing once this has been merged.

### Commits signed with GPG?

Yes

PR #47207, opened 2018-04-20
Fix pip_state with pip3 if no changes occourred

benediktwerner:
### What does this PR do?
This fixes the behavior of `states.pip.installed` when used with `pip3` (via `bin_env`) in the case that no changes occurred (all packages are already installed and up-to-date).

### Previous Behavior
If packages specified via the `name` parameter were already installed the function returned with an error because in this case `salt` tells `pip` to install no packages which produces an error. For older versions of `pip` that error was caught via a `startswith()`-call on `stdout`. This fails when using `pip3` because it now outputs the error on `stderr` and adds an additional `ERROR`-prefix.

If packages were instead specified via a `requirements`-file salt always reported that something changed because `pip3` uses a slightly different wording to indicate that a package wasn't changed.

### New Behavior
If no package needs to be installed the function returns without calling `pip install` and reports that all packages are installed (and up-to-date, if `upgrade` was specified).

### Commits signed with GPG?

Yes

Also does this count as a bugfix and needs to go into a different branch? If so just tell me into which branch I should merge.

### DIFF
--- a/salt/states/pip_state.py
+++ b/salt/states/pip_state.py
@@ -725,6 +725,13 @@ def installed(name,
                 ret['comment'] = out['comment']
                 return ret
 
+        if not target_pkgs:
+            ret['result'] = True
+            aicomms = '\n'.join(already_installed_comments)
+            last_line = 'All specified packages are already installed' + (' and up-to-date' if upgrade else '')
+            ret['comment'] = aicomms + ('\n' if aicomms else '') + last_line
+            return ret
+
     # Construct the string that will get passed to the install call
     pkgs_str = ','.join([state_name for _, state_name in target_pkgs])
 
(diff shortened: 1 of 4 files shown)


benediktwerner:
Ok, I think I adjusted most of the tests, but I don't get why this is failing: https://jenkins.saltstack.com/job/PR/job/salt-pr-linode-cent7-py3/4230/testReport/junit/integration.states.test_virtualenv/VirtualenvTest/test_issue_1959_virtualenv_runas/

It looks related but I don't see how my changes could cause this, as this is only related to the return value which shouldn't be different.

benediktwerner:
I just realized #47102 fixes part of the issue but isn't merged forward yet. But the issue with the requirements file still remains and it also looks like there would be some leftover code in pip_state. I will have another look at this.

Issue #43185, opened 2017-08-25
cmd.run runas does not properly change user on macOS


odormond:
### Description of Issue/Question

On macOS, tools like https://github.com/mas-cli/mas, uses some Apple specific internal call to get user specific informations. When invoked through `cmd.run` with the `runas` option set, they will not get the correct user specific data.

The problem can be pinpointed to the use of `sudo` instead of `su` as it can be reproduced directly with these two without involving salt at all (assuming testuser is logged into the AppStore and root is not):
```
mac-minion:~ root# sudo -i -u testuser /usr/local/bin/mas account
Not signed in
Error: Not signed in
mac-minion:~ root# su - testuser -c '/usr/local/bin/mas account'
testuser@example.com
```

### Setup

On the mac minion, install https://github.com/mas-cli/mas and log in as `testuser`. Do a `mas signin testuser@example.com t3stPassw0rd`.

### Steps to Reproduce Issue

Actual behavior:
```
root@salt-master:~# salt mac-minion cmd.run '/usr/local/bin/mas account' runas=testuser
mac-minion:
    Not signed in
    Warning: Not signed in
ERROR: Minions returned with non-zero exit code
```

Expected behavior:
```
root@salt-master:~# salt mac-minion cmd.run '/usr/local/bin/mas account' runas=testuser
mac-minion:
    testuser@example.com
```

The expected result can be obtained with this workaround:
```
root@salt-master:~# salt mac-minion cmd.run "su - testuser -c '/usr/local/bin/mas account'"
mac-minion:
    testuser@example.com
```

### Versions Report
==== master ====

Salt Version:
           Salt: 2016.11.5
 
Dependency Versions:
           cffi: Not Installed
       cherrypy: Not Installed
       dateutil: Not Installed
      docker-py: Not Installed
          gitdb: 0.6.4
      gitpython: 1.0.1
          ioflo: Not Installed
         Jinja2: 2.9.6
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.4.8
   mysql-python: Not Installed
      pycparser: Not Installed
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: Not Installed
         Python: 2.7.12 (default, Nov 19 2016, 06:48:10)
   python-gnupg: Not Installed
         PyYAML: 3.12
          PyZMQ: 16.0.2
[...]

PR #47212, opened 2018-04-20
fix macOS running as user.

weswhet:
### What does this PR do?

Fixes an issue on macOS where running certain commands as a user would fail. This also fixes the service state on macOS for starting/stoping LaunchAgents.

https://github.com/saltstack/salt/issues/43185

### Previous Behavior
On macOS `cmdmod` would try to load the user environment with something like `'sudo', '-i', '-u', runas, '--', sys.executable`. This method of running commands as user fails for some commands like launchctl.

### New Behavior
When attempting to run a command as a user `cmdmod` will now inject `su -l {user} -c` into the command. This method works for properly simulating commands as a user when run from root for every scenario I could think of. 

### Tests written?

No

### Commits signed with GPG?

Yes


cachedout (approved):

gtmanfred (commented):
Has this been tested on any other BSD platforms?

I am wondering if it needs to be more restrictive and only be applied on mac, and not BSD OSs

@cedwards would you mind taking a look at this?

weswhet:
This should only be on macOS unless `salt.utils.platform.is_darwin()` returns True on something other than macOS.

gtmanfred (approved):
After reading about darwin more closely, it is only Apple stuff, for some reason I thought that apple still used the freebsd kernel, but they have done their own stuff with it and renamed it darwin.

salt/modules/cmdmod.py
```diff
@@ -404,6 +404,19 @@ def _get_stripped(cmd):
 
         return win_runas(cmd, runas, password, cwd)
 
+    if runas and salt.utils.platform.is_darwin():
+        # we need to insert the user simulation into the command itself and not
+        # just run it from the environment on macOS as that
+        # method doesn't work properly when run as root for certain commands.
+        if isinstance(cmd, (list, tuple)):
+            cmd = ' '.join(cmd)
```
terminalmage: This will break if any arguments contain spaces. You'll need to use shlex.quote from six which we've already imported into cmdmod.py as `_cmd_quote`. For example:

```python
```

Note that you'll also need to add `map` to the `from salt.ext.six.moves import` in cmdmod.py to ensure that the itertools version is used in Python 2:

```python
from salt.ext.six.moves import range, zip, map
```

salt/modules/cmdmod.py
```diff
@@ -404,6 +404,19 @@ def _get_stripped(cmd):
...
+
+        cmd = 'su -l {0} -c \'{1}\''.format(runas, cmd)
```
terminalmage: Since shlex.quote uses single quotes, you'll want to double-quote `{1}` here.

terminalmage (changes requested):

weswhet:
@terminalmage Thanks for taking a look at this PR. I made the requested changes. Let me know if there is anything else.

rallytime:
re-run lint

terminalmage (approved):

Issue #38671, opened 2017-01-10
zpool.get error


MikeSpaceG:
### Description of Issue/Question
zpool.get fails as '-o' is not a valid option with ZFS on Linux.

### Setup
zfs: 0.6.5.8-1.el7_3.centos

### Steps to Reproduce Issue

```
# salt nas zpool.get pool/users
nas:
    ----------
    error:
        invalid option 'o'
        usage:
        	get [-pH] <"all" | property[,...]> <pool> ...
        
        the following properties are supported:
        
        	PROPERTY         EDIT   VALUES
        
        	allocated          NO   <size>
        	capacity           NO   <size>
        	dedupratio         NO   <1.00x or higher if deduped>
        	expandsize         NO   <size>
        	fragmentation      NO   <percent>
        	free               NO   <size>
        	freeing            NO   <size>
        	guid               NO   <guid>
        	health             NO   <state>
        	leaked             NO   <size>
        	size               NO   <size>
        	altroot           YES   <path>
        	ashift            YES   <ashift, 9-13, or 0=default>
        	autoexpand        YES   on | off
        	autoreplace       YES   on | off
        	bootfs            YES   <filesystem>
        	cachefile         YES   <file> | none
        	comment           YES   <comment-string>
        	dedupditto        YES   <threshold (min 100)>
        	delegation        YES   on | off
        	failmode          YES   wait | continue | panic
        	listsnapshots     YES   on | off
        	readonly          YES   on | off
        	version           YES   <version>
        	feature@...       YES   disabled | enabled | active
        
        The feature@ properties must be appended with a feature name.
        See zpool-features(5).
    pool/users:
        ----------
ERROR: Minions returned with non-zero exit code
```

### Versions Report
Master:
```
Salt Version:
           Salt: 2016.3.4
 
Dependency Versions:
           cffi: 1.5.2
       cherrypy: Not Installed
       dateutil: 2.4.2
          gitdb: 0.6.4
      gitpython: 1.0.1
          ioflo: Not Installed
         Jinja2: 2.8
        libgit2: 0.24.0
        libnacl: Not Installed
       M2Crypto: 0.21.1
           Mako: 1.0.3
   msgpack-pure: Not Installed
 msgpack-python: 0.4.6
   mysql-python: Not Installed
      pycparser: 2.14
       pycrypto: 2.6.1
         pygit2: 0.24.0
         Python: 2.7.12 (default, Nov 19 2016, 06:48:10)
   python-gnupg: Not Installed
[...]

PR #47224, opened 2018-04-21
Zpool zol fixes

sjorge:
### What does this PR do?
Was going though some of the zfs/zpool tickets, looks liked I did not get tagged on #38671

This is also because the ZFS on Linux version on Ubuntu 16.04 is rather old.
We fix this by not using `-o` to request the fields we want, we simple requests all the fields.

Then manually strip 'name'.

#38671

### Previous Behavior
The call to ```zpool``` used the newish ```-o``` flag, some minor docs update are also included.

### New Behavior
We do not use ```-o```

### Tests written?
Tests update to include new command output.

### Commits signed with GPG?
No

### Backports
- 2018.3


pruiz:
IMHO, this is a bad idea, as not filtering by properties names makes zfs states/module quite slow on systems with lots of datasets (mor specifically snapshots), as I already stated at issue #47225.

A better option would be to only filter by property name if/when the installed zfs utilities allow for it (something which can be tested by pseudo-parsing "zfs/zpool -h" output).

sjorge:
@pruiz we're still limiting property names!

```
[tachyon :: sjorge][/tmp]
[.]$ zpool get -H size tank
tank	size	112M	-
[tachyon :: sjorge][/tmp]
[.]$ zpool get -H -o property,value,source size tank
size	112M	-
```

The older ZFS does not allow limiting of the returned fields for the property we request. This should not introduce a noticable performance impact.

sjorge:
@rallytime test failures seem not related to this change.

pruiz:
@sjorge yeah, I understand older zfs versions do not allow for '-o' argument. However, on new zfs versions supporting it, not passing '-o' (and sometimes -s too) when querying/filtering datasets with lots of snapshots, the command takes minutes to complete.

So my request was: instead of removing the '-o' usage here. Try to guess wether the installed zfs command supported it or not (for example by parsing zfs help output with a regex), so you can dynamically decide wether to pass '-o' or not.

sjorge:
@pruiz Do you have any examples I have not seen any performance impact of this. I only had up to about a thousand datasets though.

```
time zpool get -H size tank
time zpool get -H -o property,value,source size tank
```

Difference would be nice data to have. (Now mind you dropping the 'size' one and fetching all properties does have an impact but that is not what this particular change does.

pruiz:
Hi @sjorge,

My main concern is not with 'zpool', but with 'zfs' command, see this for example:

```
# zfs list -t all |wc -l
15058

# zfs list -t snapshot |wc -l
15051

# time zfs list -t all > /dev/null

real	1m22.034s
user	0m6.107s
sys	1m20.927s

# time zfs list -t all -o name > /dev/null

real	1m2.041s
user	0m9.097s
sys	0m51.944s

# time zfs list -t all -o name -s name > /dev/null

real	0m3.067s
user	0m0.040s
sys	0m2.630s
```

This machine is somewhat old (Xeon Server from 2012), with 4xSAS disks on raidz1. But still not uncommon, nor specially slow on anything else, but listing zfs datasets..

sjorge:
Hi @pruiz 

This PR does not touch the zfs list. I went over all the commands and zpool list was the only one that is currently broken on ZoL due to the use of ```-o```.

Not as large as your pool, but here is a largish one. Same as you it is slowish when using zfs list.
```
[root@carbon ~]# time zfs list -t snap -r tank | wc -l
   50011

real	0m14.096s
user	0m4.348s
sys	0m9.667s
[root@carbon ~]# time zfs list -r tank | wc -l
    5002

real	0m1.210s
user	0m0.571s
sys	0m0.644s
```

For ```zpool get``` there does not a significant difference. It's also not possible to use recursion with get. So the total count of dataset/snapshot should not not make a difference.
```
[root@carbon ~]# time zpool get -H -o name,property,value,source size tank
tank	size	1.88G	-

real	0m0.016s
user	0m0.005s
sys	0m0.011s
[root@carbon ~]# time zpool get -H  size tank
tank	size	1.88G	-

real	0m0.018s
user	0m0.006s
sys	0m0.012s
```

pruiz:
Then, my bad, I thought the PR was against zfs command, instead of zpool. I should have digged a little bit deeper..  ;)

cachedout:
@pruiz Are you all right with this PR as it sits then?

pruiz:
@cachedout yep, I myself have not seen any perf issues with zpool commands with vs without -o/-s (unlike zfs command).

sjorge:
@cachedout bump

sjorge:
@rallytime can you take a look at this? It should probably also get backported to 2018.3 eventually.
I actually started on this before the ZFS stuff got backported from develop.

rallytime (approved):

rallytime:
@sjorge Yeah, sure thing. Looks good to me and i've marked this for back-porting.

Issue #47225, opened 2018-04-21
zfs.filesystem_present takes forever on a dataset with lots (10k+) of snapshots


pruiz:
### Description of Issue/Question

zfs.filesystem_present (when specifying properties) takes forever on a dataset with lots (10k+) of snapshots, this is due to underlying zfs.dataset_present function not limiting properties query to only the filesystem actually being tested for presence, 
but instead allowing query to ask for snapshots and other sub-datasets too.

### Setup

Given a zfs dataset with lots of snapshots, a state like the follwing will take minutes to complete:

```yaml
zfs::dataset::data:
  zfs.filesystem_present:
    - name: TANK/data
    - properties:
        mountpount: /data
```

### Versions Report

```
Salt Version:
           Salt: 2017.7.1

Dependency Versions:
           cffi: Not Installed
       cherrypy: Not Installed
       dateutil: Not Installed
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.7.2
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.5.1
   mysql-python: Not Installed
      pycparser: Not Installed
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: Not Installed
         Python: 2.7.5 (default, Aug  4 2017, 00:39:18)
   python-gnupg: 0.3.8
         PyYAML: 3.11
          PyZMQ: 15.3.0
           RAET: Not Installed
          smmap: Not Installed
        timelib: Not Installed
        Tornado: 4.2.1
            ZMQ: 4.1.4

System Versions:
           dist: centos 7.4.1708 Core
         locale: UTF-8
        machine: x86_64
        release: 4.4.128-1.el7.elrepo.x86_64
         system: Linux
        version: CentOS Linux 7.4.1708 Core
```

PR #47226, opened 2018-04-21
Fix issue #47225: avoid zfs.filesystem_present slowdown when dataset has lots of snapshots

pruiz:
### What does this PR do?

Changes zfs.dataset_present so it 1) takes into consideration the intended type of dataset while filtering for properties. 2) limits the properties query depth to 0 (ie. just the referenced dataset)

### What issues does this PR fix or reference?

Fixes #47225 (against develop)

### Previous Behavior

Existing dataset's properties query did not filter by dataset type, and (incorrectly) included fetching of sub-dataset properties.

### New Behavior

Now only really needed information is requested to zfs, thus speeding up the query (a lot) on affected scenarios.

### Tests written?

No

### Commits signed with GPG?

No

PR #47228, opened 2018-04-21
Fix issue #47225: avoid zfs.filesystem_present slowdown when dataset has lots of snapshots (2018.3 branch)

pruiz:

This is the same pull-request as #47226, but against branch 2018.3.

zfs.filesystem_present (when specifying properties) takes forever on a dataset with lots (10k+) of snapshots, this is due to underlying zfs.dataset_present function not limiting properties query to only the filesystem actually being tested for presence, 
but instead allowing query to ask for snapshots and other sub-datasets too.

### Setup

Given a zfs dataset with lots of snapshots, a state like the follwing will take minutes to complete:

```yaml
zfs::dataset::data:
  zfs.filesystem_present:
    - name: TANK/data
    - properties:
        mountpount: /data
```

### DIFF
--- a/salt/states/zfs.py
+++ b/salt/states/zfs.py
@@ -454,8 +454,9 @@ def _dataset_present(dataset_type, name, volume_size=None, sparse=False, create_
         ## NOTE: fetch current volume properties
         properties_current = __salt__['zfs.get'](
             name,
+            type=dataset_type,
             fields='value',
-            depth=1,
+            depth=0,
             parsable=True,
         ).get(name, OrderedDict())
 


rallytime:
Hi @pruiz - There is a merge conflict on this change now. Can you rebase and fix the conflicts? Thanks for your work here. :)

dmurphy18 (changes requested):
The develop branch changes for zfs were backported to 2018.3 branch, hence would expect that the changes to states/zfs.py should be the same as in #47226

rallytime (approved):

isbm (approved):

PR #47230, opened 2018-04-22
Add functionality and tests for managing selinux port policy

### LINKED ISSUE #42635: Selinux : add handling for ports

Selinux can restrict access to ports for programs (typically, Apache/Nginx would be restricted to ports 80 and 443 (or the class http_port_t).

It would be nice to be able to handle ports contexts in addition to files in the selinux module/state.

An example semanage command for doing so would be : `semanage port -a -t http_port_t -p tcp 6543`.

If this feature is accepted, I could help with the implementation.

leeclemens:
### What does this PR do?

Allow usage of `semanage port` via modules and states to view/add/delete SELinux port policies.

Fixes #42635

### New Behavior
Modules:

- `selinux.port_get_policy`
- `selinux.port_add_or_delete_policy`

States:

- `selinux.port_policy_present`
- `selinux.port_policy_absent`

### Tests written?

Yes (for validating arguments and parsing semanage output)

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/modules/selinux.py
+++ b/salt/modules/selinux.py
@@ -387,6 +387,28 @@ def _validate_filetype(filetype):
     return True
 
 
+def _parse_protocol_port(name, protocol, port):
+    '''
+    .. versionadded:: Fluorine
+
+    Validates and parses the protocol and port/port range from the name
+    if both protocol and port are not provided.
+
+    If the name is in a valid format, the protocol and port are ignored if provided
+
+    Examples: tcp/8080 or udp/20-21
+    '''
+    protocol_port_pattern = r'^(tcp|udp)\/(([\d]+)\-?[\d]+)$'
+    name_parts = re.match(protocol_port_pattern, name)
+    if not name_parts:
+        name_parts = re.match(protocol_port_pattern, '{0}/{1}'.format(protocol, port))
+    if not name_parts:
+        raise SaltInvocationError(
+            'Invalid name "{0}" format and protocol and port not provided or invalid: "{1}" "{2}".'.format(
+                name, protocol, port))
+    return name_parts.group(1), name_parts.group(2)
+
+
 def _context_dict_to_string(context):
     '''
     .. versionadded:: 2017.7.0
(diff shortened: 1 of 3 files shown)


salt/modules/selinux.py
```diff
@@ -604,3 +626,99 @@ def fcontext_apply_policy(name, recursive=False):
                 del new[key]
             ret['changes'].update({filespec: {'old': old, 'new': new}})
     return ret
+
+
+def port_get_policy(name, sel_type=None, protocol=None, port=None):
+    '''
+    .. versionadded:: develop
```
gtmanfred: This should be Fluorine.

salt/states/selinux.py
```diff
@@ -480,3 +480,110 @@ def fcontext_policy_applied(name, recursive=False):
             ret.update({'result': True})
             ret.update({'changes': apply_ret.get('changes')})
     return ret
+
+
+def port_policy_present(name, sel_type, protocol=None, port=None, sel_range=None):
+    '''
+    .. versionadded:: develop
```
gtmanfred: Fluorine here too

salt/modules/selinux.py
```diff
@@ -604,3 +626,99 @@ def fcontext_apply_policy(name, recursive=False):
...
+
+def port_add_or_delete_policy(action, name, sel_type=None, protocol=None, port=None, sel_range=None):
+    '''
+    .. versionadded:: develop
+
+    Adds or deletes the SELinux policy for a given protocol and port.
+
+    Returns the result of the call to semanage.
+
+    name
```
gtmanfred: Can you document the action variable as well? including that it is `add` or `delete`.

gtmanfred (changes requested):
Just some doc fixes, and this should be good.

salt/modules/selinux.py
leeclemens: Should it be set to literally "Fluorine"? https://docs.saltstack.com/en/latest/topics/releases/version_numbers.html shows the version number as `TBD` - I wasn't sure how/when these get updated when released.

salt/modules/selinux.py
leeclemens: Good catch, adding that as well

salt/modules/selinux.py
gtmanfred: Yes, it should be just the word `Fluorine` we do a search and replace on the code word once we know when the release date will be.

leeclemens: Thanks, updated

salt/states/selinux.py
leeclemens: Thanks, updated

gtmanfred (approved):

rallytime:
re-run py

salt/modules/selinux.py
```diff
@@ -604,3 +626,102 @@ def fcontext_apply_policy(name, recursive=False):
...
+        return None
+
+    parts = re.match(r'^(\w+)[ ]+(\w+)[ ]+([\d\-, ]+)', port_policy)
+    return {
+        'sel_type': parts.group(1).strip(),
+        'protocol': parts.group(2).strip(),
+        'port': parts.group(3).strip(), }
+
+
+def port_add_or_delete_policy(action, name, sel_type=None, protocol=None, port=None, sel_range=None):
```
terminalmage: This seems like a cumbersome name for a function, and also redundant since you're making the user specify the action as an argument. How about changing this function name to `_port_add_or_delete_policy()` and then making two separate functions to add and delete, each of which is a one-liner that calls the common function. For example:

```python
def port_add_policy(name, sel_type=None, protocol=None, port=None, sel_range=None):
    '''
    <docstring goes here>
    '''
    return _port_add_or_delete_policy('add', name, sel_type, protocol, port, sel_range)


def port_delete_policy(name, sel_type=None, protocol=None, port=None, sel_range=None):
    '''
    <docstring goes here>
    '''
    return _port_add_or_delete_policy('delete', name, sel_type, protocol, port, sel_range)
```

gtmanfred: This is the same naming scheme as the rest of the modules in this execution module.

https://github.com/saltstack/salt/blob/develop/salt/modules/selinux.py#L483

Probably good to rename all of them honestly, but i think that can be done in a different pr.

terminalmage (approved):

salt/modules/selinux.py
leeclemens: Yes, I pretty much copied the fcontext naming/flow and altered it accordingly. I'll make a note to put together another PR - thanks for the feedback from both of you!

leeclemens: I submitted PR #47622

gtmanfred: Awesome thanks @leeclemens

PR #47232, opened 2018-04-22
Fixed usage of ipaddress

brejoc:
### What does this PR do?

ipaddress is imported either directly or from salt.ext. If we use it, we shouldn't address it with salt.ext.ipaddress.

### Previous Behavior

AttributeError: module 'salt.ext' has no attribute 'ipaddress'

### New Behavior

No more AttributeError.

### Tests written?

No

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/modules/network.py
+++ b/salt/modules/network.py
@@ -1140,7 +1140,7 @@ def convert_cidr(cidr):
     ret = {'network': None,
            'netmask': None}
     cidr = calc_net(cidr)
-    network_info = salt.ext.ipaddress.ip_network(cidr)
+    network_info = ipaddress.ip_network(cidr)
     ret['network'] = six.text_type(network_info.network_address)
     ret['netmask'] = six.text_type(network_info.netmask)
     return ret


salt/modules/network.py
```diff
@@ -1140,7 +1140,7 @@ def convert_cidr(cidr):
     ret = {'network': None,
            'netmask': None}
     cidr = calc_net(cidr)
-    network_info = salt.ext.ipaddress.ip_network(cidr)
```
mirceaulinic: This is `salt.ext.ipaddress`: https://github.com/saltstack/salt/blob/develop/salt/ext/ipaddress.py - it would suggest there's a different issue if you're seeing an error when importing this module. Can you elaborate?

brejoc: Depending on the Python version we either import directly or from salt.ext. In both cases we have to use ipaddress directly, not via salt.ext.ipaddress.

mirceaulinic: Ah alright, that makes sense. Thanks!

cachedout (approved):

PR #47240, opened 2018-04-23
Add autoremove, force-removal-of-dependent-packages for opkg.remove

chotea:
### What does this PR do?
This PR adds support for opkg remove in terms of giving the possibility of specifying two extra parameters: force-removal-of-dependent-packages, autoremove 

### Previous Behavior
The implementation was always calling opkg remove without any extra parameters.

### New Behavior
Now we check these 2 force-options in kwargs and append them to the command if necessary:
- --force-removal-of-dependent-packages - Remove package and all dependencies
- --autoremove - Remove packages that were installed automatically to
satisfy dependencies

### Tests written?
No

### Commits signed with GPG?
No

### DIFF
--- a/salt/modules/opkg.py
+++ b/salt/modules/opkg.py
@@ -435,6 +435,15 @@ def remove(name=None, pkgs=None, **kwargs):  # pylint: disable=unused-argument
         A list of packages to delete. Must be passed as a python list. The
         ``name`` parameter will be ignored if this option is passed.
 
+    remove_dependent_packages
+        Remove package and all dependencies
+
+        .. versionadded:: Fluorine
+
+    auto_remove
+        Remove packages that were installed automatically to satisfy dependencies
+
+        .. versionadded:: Fluorine
 
     Returns a dict containing the changes.
 
@@ -445,6 +454,7 @@ def remove(name=None, pkgs=None, **kwargs):  # pylint: disable=unused-argument
         salt '*' pkg.remove <package name>
         salt '*' pkg.remove <package1>,<package2>,<package3>
         salt '*' pkg.remove pkgs='["foo", "bar"]'
+        salt '*' pkg.remove pkgs='["foo", "bar"]' remove_dependent_packages=True auto_remove=True
     '''
     try:
         pkg_params = __salt__['pkg_resource.parse_targets'](name, pkgs)[0]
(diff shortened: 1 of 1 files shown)


salt/modules/opkg.py
```diff
@@ -435,6 +435,15 @@ def remove(name=None, pkgs=None, **kwargs):  # pylint: disable=unused-argument
         A list of packages to delete. Must be passed as a python list. The
         ``name`` parameter will be ignored if this option is passed.
 
+    remove_dependent_packages
```
isbm: Maybe "remove_dependencies" would be better. What else than "packages" can be possibly removed in the context of "dependency"?

salt/modules/opkg.py
```diff
@@ -456,6 +466,10 @@ def remove(name=None, pkgs=None, **kwargs):  # pylint: disable=unused-argument
     if not targets:
         return {}
     cmd = ['opkg', 'remove']
+    if kwargs.get('remove_dependent_packages', False):
+        cmd.append('--force-removal-of-dependent-packages')
+    if kwargs.get('auto_remove', False):
```
isbm: You actually not need to write default `False` to `.get(...)`. In this case perfectly enough to:
```python
if kwargs.get('auto_remove'):
    cmd.append...
```
It will either return `None` (no param) or any value. You can validate it though, if you want to _force_ people pass `True/False` instead of `1/0` etc.

isbm (changes requested):
...and adding a unit test would be a great wrap-up, alongside with the suggestions. :wink:

salt/modules/opkg.py
```diff
@@ -435,6 +435,15 @@ def remove(name=None, pkgs=None, **kwargs):  # pylint: disable=unused-argument
         A list of packages to delete. Must be passed as a python list. The
         ``name`` parameter will be ignored if this option is passed.
 
+    remove_dependent_packages
+        Remove package and all dependencies
+
+        .. versionadded:: Fluorine
+
+    auto_remove
```
cachedout: `auto_remove_deps` perhaps?

cachedout:
@chotea Have you had a chance to review the feedback on this PR?

rallytime:
@chotea Any chance you were able to come back to this one?

chotea:
Yes. I implemented the feedback and wrote some tests. I wasn't able to update this PR because I was shifted to something with more priority. I will come back ASAP.

rallytime:
Since we haven't heard back here for a while, I'm closing this PR.

We'd still love to have this change, so when you're ready to implement the feedback, let us know and we will gladly open this up again. If it's easier for you to just submit a new PR, that's fine too.

Thank you!

chotea:
@rallytime  Can you please reopen this PR? I have implemented the feedback and wrote unit tests for opkg module.

PR #47242, opened 2018-04-23
PY3 fix for zeromq setsockopt

aesposito91:
What does this PR do?

Resolves the following traceback, appears related to zmq changes -

[CRITICAL] Engine 'napalm_syslog' could not be started!
Traceback (most recent call last):
File "/usr/lib/python2.7/dist-packages/salt/engines/init.py", line 128, in run
self.engineself.fun
File "/usr/lib/python2.7/dist-packages/salt/engines/napalm_syslog.py", line 317, in start
port=port)
File "/usr/lib/python2.7/dist-packages/salt/engines/napalm_syslog.py", line 234, in _get_transport_recv
return TRANSPORT_FUN_MAP[name](address, port, **kwargs)
File "/usr/lib/python2.7/dist-packages/salt/engines/napalm_syslog.py", line 223, in _zmq
socket.setsockopt(zmq.SUBSCRIBE, '')
File "zmq/backend/cython/socket.pyx", line 426, in zmq.backend.cython.socket.Socket.set
TypeError: unicode not allowed, use setsockopt_string
Tests written?

No.
Commits signed with GPG?

No.

Please review Salt's Contributing Guide for best practices.

See GitHub's page on GPG signing for more information about signing commits with GPG.

PR #47245, opened 2018-04-23
Ensure we pass hexid as bytes when zmq_filtering enabled

terminalmage:
When using hashlib and running `.hexdigest()`, a `str` type is returned.  It needs to be converted to a bytes type for the setsockopt a few lines below, if zmq_filtering is enabled and the master is running Python 3.

Inspired by https://github.com/saltstack/salt/pull/47242.


rallytime (approved):

cachedout (approved):

pengyao:
#48881

PR #47253, opened 2018-04-23
More reliable virtual method

dwoz:
### What does this PR do?

Fixes failing test `integration.grains.test_core.TestGrainsCore.test_win_cpu_model`

### Tests written?

No - Fixes Test

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/modules/reg.py
+++ b/salt/modules/reg.py
@@ -33,6 +33,11 @@ import logging
 # Import Salt libs
 import salt.utils.platform
 from salt.exceptions import CommandExecutionError
+try:
+    import salt.utils.win_reg
+    HAS_WINDOWS_MODUELS = salt.utils.win_reg.HAS_WINDOWS_MODULES
+except ImportError:
+    HAS_WINDOWS_MODULES = False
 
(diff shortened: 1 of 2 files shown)


salt/modules/reg.py
```diff
@@ -33,6 +33,11 @@
 # Import Salt libs
 import salt.utils.platform
 from salt.exceptions import CommandExecutionError
+try:
+    import salt.utils.win_reg
+    HAS_WINDOWS_MODUELS = salt.utils.win_reg.HAS_WINDOWS_MODULES
```
twangboy: Should probably be `HAS_WINDOWS_MODULES`... unless we're dueling...

twangboy (changes requested):

salt/modules/reg.py
dwoz: @twangboy good catch, thanks!

dwoz:
This doesn't seem to fully fix the issue. I saw other tests fail when running the full suite. Closing this PR until I have a solid fix.

PR #47256, opened 2018-04-23
add macOS mac_service update.

weswhet:
### What does this PR do?
This PR adds support for using the latest macOS `launchctl` subcommands as well as the ability to manage user specific services. 

### Previous Behavior
Previously salt would use the legacy subcommands like `/bin/launchctl load/unload` to start and stop services. It was also not capable of modifying any user specific services located in `/Users/foo/Library/LaunchAgents/`

### New Behavior
Salt now leverages the latest subcommands `/bin/launchctl bootstrap/bootout` to start and stop services. Using these subcommands allows us to properly load and unload LauchAgents from root without having to run commands as the user as this is handled properly by `launchctl` when specifying to appropriate domain/system targets like `gui/501/` or `gui/501/com.apple.example.`

Salt also now looks for user specific services located in `/Users/foo/Library/LaunchAgents/` and is now capable of modifying them.

### Tests written?

No

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/modules/mac_service.py
+++ b/salt/modules/mac_service.py
@@ -113,6 +113,45 @@ def _always_running_service(name):
     return False
 
 
+def _get_domain_target(name, service_target=False):
+    '''
+    Returns the domain/service target and path for a service. This is used to
+    determine whether or not a service should be loaded in a user space or
+    system space.
+
+    :param str name: Service label, file name, or full path
+
+    :param bool service_target: Whether to return a full
+    service target. This is needed for the enable and disable
+    subcommands of /bin/launchctl. Defaults to False
+
+    :return: Tuple of the domain/service target and the path to the service.
+
+    :rtype: tuple
+    '''
+
+    # Get service information
+    service = _get_service(name)
+
+    # get the path to the service
+    path = service['file_path']
+
+    # most of the time we'll be at the system level.
+    domain_target = 'system'
+
+    # check if a LaunchAgent as we should treat these differently.
+    if 'LaunchAgents' in path:
+        # Get the console user so we can service in the correct session
+        uid = salt.utils.mac_utils.console_user()
+        domain_target = 'gui/{}'.format(uid)
+
+    # check to see if we need to make it a full service target.
+    if service_target is True:
+        domain_target = '{}/{}'.format(domain_target, service['plist']['Label'])
+
+    return (domain_target, path)
+
+
 def show(name):
     '''
     Show properties of a launchctl service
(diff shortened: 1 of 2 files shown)


salt/modules/mac_service.py
```diff
@@ -113,6 +113,45 @@ def _always_running_service(name):
...
+    # get the path to the service
+    path = service['file_path']
+
+    # most of the time we'll be at the system level.
+    domain_target = 'system'
+
+    # check if a LaunchAgent as we should treat these differently.
+    if 'LaunchAgents' in path:
+        # Get the console user so we can service in the correct session
+        uid = salt.utils.mac_utils.console_user()
```
gtmanfred: can you switch these to using `__utils__['mac_utils.console_user']()` instead of using the import `salt.utils.mac_utils`?

It allows for people that use the dropin dynamic modules to actually replace the modules that use newer salt utils.

Thanks,
Daniel

gtmanfred (approved):
Just one change I would like to see, but it can be merged without it.

salt/modules/mac_service.py
weswhet: @gtmanfred would like me to convert the other mac_util calls as well?

weswhet:
after testing the requested changes by Daniel, I noticed that the service state doesn't pass in the `runas` parameter when checking the status of a LaunchAgent. I've added a patch for this behavior in https://github.com/saltstack/salt/pull/47256/commits/ea6c18cef3b3d4ae35fea6dbbc208ce840ff51c9

salt/modules/mac_service.py
gtmanfred: Eventually we should go through and change them all to using `__utils__`  i was just think about these ones since you are adding new ones.

weswhet: Sounds good, I changed all my additions to using `__utils__`

weswhet:
@rallytime I'd like to update the documentation with some notes and version added information for these changes. Would this PR make it into the Fluorine release?

rallytime:
@weswhet Yeah, definitely. We're not branching for the Fluorine release for a while.

rallytime:
Hi @weswhet - It looks like this change is causing the following related tests to fail:

- unit.utils.test_mac_utils.MacUtilsTestCase.test_available_services
- unit.utils.test_mac_utils.MacUtilsTestCase.test_available_services_broken_symlink
- unit.utils.test_mac_utils.MacUtilsTestCase.test_available_services_non_xml
- unit.utils.test_mac_utils.MacUtilsTestCase.test_available_services_non_xml_malformed_plist

https://jenkins.saltstack.com/job/PR/job/salt-pr-linode-ubuntu14-n/22418/

Can you take a look?

twangboy (approved):

salt/modules/mac_service.py
twangboy: I'm still seeing `salt.utils.mac_utils.console_user()` need change it to `__utils__['mac_utils.console_user']`

weswhet:
@rallytime, for future reference what state are the test Macs in? I’m assuming headless With no users and sitting at the login window? Just curious for any future code changes.

PR #47257, opened 2018-04-23
Role is not a list but a dictionary

jeroennijhof:
### What does this PR do?
Checks if there is a Role key and make sure it's a dictionary so it could be parsed.

Fixes a bug in salt cloud msazure:
```
input_endpoint = role['ConfigurationSets']['ConfigurationSet'].get('InputEndpoints', {}).get('InputEndpoint')
TypeError: string indices must be integers
``` 

### Previous Behavior
Not able to list or add input endpoints for azure classic

### New Behavior
Able to list and add input endpoints for azure classic

### Tests written?

No

### Commits signed with GPG?

No

### DIFF
--- a/salt/cloud/clouds/msazure.py
+++ b/salt/cloud/clouds/msazure.py
@@ -2043,9 +2043,10 @@ def list_input_endpoints(kwargs=None, conn=None, call=None):
 
     ret = {}
     for item in data:
-        if 'Role' not in item:
-            continue
-        for role in item['Role']:
+        if 'Role' in item:
+            role = item['Role']
+            if not isinstance(role, dict):
+                return ret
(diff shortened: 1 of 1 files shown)


salt/cloud/clouds/msazure.py
```diff
@@ -2043,16 +2043,18 @@ def list_input_endpoints(kwargs=None, conn=None, call=None):
 
     ret = {}
     for item in data:
-        if 'Role' not in item:
-            continue
-        for role in item['Role']:
```
jeroennijhof: This will contain RoleType since it's the first element in the dictionary and not the planned dictionary itself.

gtmanfred:
@techhat can you take a look at this for me? I am not familiar with roles in azure at all.

Thanks,
Daniel

techhat (approved):

rallytime:
re-run py

PR #47261, opened 2018-04-24
Bash completion log level and function cache

shtrom:
### What does this PR do?

1. add `--log-level error` to bash and zsh completion commands, so the `salt` and `salt-call` ran in the bacground don't pollute the terminal when logs are set to more verbose levels system-wide
2. factorise the function cache into a separate function in the bash completion, and use it for both completion for `salt` (as before) and `salt-call` (new)

None? Sorry, I should have opened one, but it was such a quick fix.

### Tests written?

No

### Commits signed with GPG?

Yes

### DIFF
--- a/pkg/salt.bash
+++ b/pkg/salt.bash
@@ -34,14 +34,41 @@ _salt_get_keys(){
     done
 }
 
-_salt(){
-    CACHE_DIR="$HOME/.cache/salt-comp-cache_functions"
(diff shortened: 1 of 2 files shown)


meaksh (approved):
Looks good to me. Thanks for contributing!

rallytime:
Thank you @shtrom!

Issue #47092, opened 2018-04-16
[2018.3.0] pkg.installed breaks with pseudo packages


syphernl:
### Description of Issue/Question

There are a few psuedo packages in the repo: `php7.0-redis`, `php7.1-redis` etc are all actually `php-redis`. 

With versions prior to 2018.3.0 this worked fine. However, a highstate with a 2018.3.0 minion results in:

```
          ID: php7.0-redis
    Function: pkg.installed
      Result: False
     Comment: The following packages failed to install/update: php7.0-redis
     Started: 11:55:47.567707
    Duration: 422.623 ms
     Changes:   
```

The package is however (already) installed. Installing it by hand shows no errors:

```
root@testing01:~# apt install php7.0-redis
Reading package lists... Done
Building dependency tree       
Reading state information... Done
Note, selecting 'php-redis' instead of 'php7.0-redis'
php-redis is already the newest version (4.0.0-1+ubuntu16.04.1+deb.sury.org+1).
0 upgraded, 0 newly installed, 0 to remove and 4 not upgraded.
```

Why is Salt giving an error? I couldn't find anything in the changelogs regarding to this.

### Setup
```
php7.0-redis:
  pkg.installed
```

### Steps to Reproduce Issue
```
root@testing01:~# salt-call --local pkg.install php7.0-redis -l debug
[DEBUG   ] Reading configuration from /etc/salt/minion
[DEBUG   ] Including configuration from '/etc/salt/minion.d/99-master-address.conf'
[DEBUG   ] Reading configuration from /etc/salt/minion.d/99-master-address.conf
[DEBUG   ] Including configuration from '/etc/salt/minion.d/_schedule.conf'
[DEBUG   ] Reading configuration from /etc/salt/minion.d/_schedule.conf
[DEBUG   ] Including configuration from '/etc/salt/minion.d/mysql.conf'
[DEBUG   ] Reading configuration from /etc/salt/minion.d/mysql.conf
[DEBUG   ] Using cached minion ID from /etc/salt/minion_id: testing01.office.amazon.enrise.net
[DEBUG   ] Configuration file path: /etc/salt/minion
[WARNING ] Insecure logging configuration detected! Sensitive data may be logged.
[DEBUG   ] Grains refresh requested. Refreshing grains.
[DEBUG   ] Reading configuration from /etc/salt/minion
[DEBUG   ] Including configuration from '/etc/salt/minion.d/99-master-address.conf'
[...]

PR #47263, opened 2018-04-24
aptpkg: Drop virtual package support for Fluorine

terminalmage:
See https://github.com/saltstack/salt/issues/47092#issuecomment-383669573

This also adds a new `pkg.show` function for aptpkg.py which wraps `apt-cache show <pkgname>`.

### DIFF
--- a/doc/topics/releases/fluorine.rst
+++ b/doc/topics/releases/fluorine.rst
@@ -5,6 +5,232 @@ Salt Release Notes - Codename Fluorine
 ======================================
 
 
+"Virtual Package" Support Dropped for APT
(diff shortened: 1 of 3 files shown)


rallytime (approved):
👍 

If this support is being removed, we probably need to put a deprecation path in somewhere in the older release branches, right?

isbm (approved):
This seems indeed brings more problems than goods. Maybe in a future APT will be a bit better than this...

terminalmage:
This will need to be rebased when https://github.com/saltstack/salt/pull/47309/files is merged.

Issue #46445, opened 2018-03-08
scheduler run job return data does not include retcode


golmaal:
Event generated when salt-call test.ping is called - 
```
salt/job/20180308090940010139/ret/salt-master  {
    "_stamp": "2018-03-08T14:09:40.011277",
    "arg": [],
    "cmd": "_return",
    "fun": "test.ping",
    "fun_args": [],
    "id": "xoas",
    "jid": "20180308090940010139",
    "retcode": 0,
    "return": true,
    "tgt": "salt-master",
    "tgt_type": "glob"
}
```

Event generated when a test.ping is added to schedule and run by scheduler - 

Added schedule as following - 
```
salt-call schedule.add test_ping function='test.ping' seconds=10
```
Generated event - 
```
salt/job/20180308091048430549/ret/salt-master {
    "_stamp": "2018-03-08T14:10:48.431949",
    "arg": [],
    "cmd": "_return",
    "fun": "test.ping",
    "fun_args": [],
    "id": "xoas",
    "jid": "20180308091048430549",
    "pid": 3630,
    "return": true,
    "schedule": "test_ping",
    "success": true,
    "tgt": "salt-master",
    "tgt_type": "glob"
}
```

@gtmanfred thought this issue was reported already. I searched a combination of terms "job, schedule[r] and retcode" and didn't find this exact issue. 

### Versions Report
Salt Version:
           Salt: 2017.7.4

Dependency Versions:
           cffi: Not Installed
       cherrypy: unknown
       dateutil: 1.5
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.7.2
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: 0.21.1
           Mako: 0.8.1
   msgpack-pure: Not Installed
 msgpack-python: 0.4.6
   mysql-python: Not Installed
      pycparser: Not Installed
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: Not Installed
         Python: 2.7.5 (default, Aug  4 2017, 00:39:18)
   python-gnupg: Not Installed
         PyYAML: 3.11
          PyZMQ: 15.3.0
           RAET: Not Installed
          smmap: Not Installed
        timelib: Not Installed
        Tornado: 4.2.1
            ZMQ: 4.1.4

System Versions:
           dist: centos 7.4.1708 Core
         locale: UTF-8
        machine: x86_64
[...]

Issue #36466, opened 2016-09-21
Scheduled task doesn't return 'retcode'


sebastian-cb:
```
schedule:
  test01:
    enabled: True
    function: cmd.shell
    args:
      - uptime
    minutes: 1
    maxrunning: 1
    returner: test
```

_returners/test.py

```
# Import python libs
from __future__ import absolute_import, print_function
import logging
log = logging.getLogger(__name__)

def returner(ret):
    '''
    Print the return data to the terminal to verify functionality
    '''
    log.debug('RET from returner: {0}'.format(ret))
    print(ret)
```

Testing - I added some extra log entries for debugging

When I run this manually there is a `retcode` in the output:

```
# salt test01.local --return test cmd.shell uptime
test01.local:
     13:52:52 up 2 days,  3:18,  1 user,  load average: 0.05, 0.03, 0.05
```

```
2016-09-21 13:55:57,759 [salt.loaded.int.returner.test][DEBUG   ][10594] RET from returner: {'fun_args': ['uptime'], 'jid': '20160921135557718108', 'return': ' 13:55:57 up 2 days,  3:21,  1 user,  load average: 0.00, 0.01, 0.05', 'retcode': 0, 'success': True, 'fun': 'cmd.shell', 'id': 'test01.local'}
```

No `retcode` from returner when job is scheduled:

```
2016-09-21 13:56:06,797 [salt.utils.lazy  ][DEBUG   ][10630] LazyLoaded test.returner
2016-09-21 13:56:06,797 [salt.loaded.int.returner.test][DEBUG   ][10630] RET from returner: {'fun_args': ['uptime'], 'jid': '20160921135606759346', 'return': ' 13:56:06 up 2 days,  3:21,  1 user,  load average: 0.00, 0.01, 0.05', 'success': True, 'schedule': 'test01', 'pid': 10630, 'fun': 'cmd.shell', 'id': 'test01.local'}
```

although it's available in scheduler itself:

```
2016-09-21 13:56:06,797 [salt.utils.schedule][DEBUG   ][10630] RET from scheduler: {'fun_args': ['uptime'], 'jid': '20160921135606759346', 'return': ' 13:56:06 up 2 days,  3:21,  1 user,  load average: 0.00, 0.01, 0.05', 'retcode': 0, 'success': True, 'schedule': 'test01', 'pid': 10630, 'fun': 'cmd.shell', 'id': 'test01.local'}
```

gtmanfred:
[...]

PR #47270, opened 2018-04-24
Fix minion scheduler to return 'retcode' from executed functions

meaksh:
### What does this PR do?
Currently, the `retcode` is not part of the response from some functions executions when these are executed via the minion scheduler, i.a. `test.ping`, `state.apply`, etc.

When these functions are called directly (without using the schedule module), the `retcode` is successfully gathered from the `__context__` and it's added to the response.

If the functions are executed by the minion scheduler, then the `retcode` is missing from the response. It seems this is happening because the `__context__` has not been successfully initialized, like we already do on https://github.com/saltstack/salt/blob/2017.7/salt/minion.py#L1474 for functions called directly.

https://github.com/saltstack/salt/issues/36466
https://github.com/saltstack/salt/issues/46445

### Previous Behavior
```
salt/job/20180424152111538991/ret/min-sles12sp3.tf.local	{
    "_stamp": "2018-04-24T13:21:11.540941", 
    "arg": [], 
    "cmd": "_return", 
    "fun": "test.ping", 
    "fun_args": [], 
    "id": "min-sles12sp3.tf.local", 
    "jid": "20180424152111538991", 
    "pid": 15427, 
    "return": true, 
    "schedule": "job1", 
    "success": true, 
    "tgt": "min-sles12sp3.tf.local", 
    "tgt_type": "glob"
}
```

### New Behavior
```
salt/job/20180424152247090134/ret/min-sles12sp3.tf.local	{
    "_stamp": "2018-04-24T13:22:47.090948", 
    "arg": [], 
    "cmd": "_return", 
    "fun": "test.ping", 
    "fun_args": [], 
    "id": "min-sles12sp3.tf.local", 
    "jid": "20180424152247090134", 
    "pid": 15930, 
    "retcode": 0, 
    "return": true, 
    "schedule": "job1", 
    "success": true, 
    "tgt": "min-sles12sp3.tf.local", 
    "tgt_type": "glob"
}
```

### Tests written?

No

### Commits signed with GPG?

Yes


salt/utils/schedule.py
```diff
@@ -889,6 +889,7 @@ def handle_func(self, multiprocessing_enabled, func, data):
                     for global_key, value in six.iteritems(func_globals):
                         self.functions[mod_name].__globals__[global_key] = value
 
+            self.functions.pack['__context__']['retcode'] = 0
```
s0undt3ch: So it will always be `0`?

meaksh: No exactly, `self.functions.pack['__context__']['retcode']` will get updated by `self.functions[func](*args, **kwargs)` execution placed on the next line of code. In case the function called does not provide any `retcode`, then it would be `0` by default. Here I mimic the behavior already existing here:  https://github.com/saltstack/salt/blob/2017.7/salt/minion.py#L1474

s0undt3ch (approved):

garethgreenaway (approved):

PR #47279, opened 2018-04-24
Gracefully shutdown worker threads

dwoz:
### What does this PR do?

Gracefully shutdown worker threads when TCPReqServerChannel is closed or destroyed.

### Previous Behavior

Running thread causes processes not to terminate.

### New Behavior

Threads will stop gracefully and processes will terminate as they should. This fixes the windows python 3 Jenkins build.

### Tests written?

No - Fixes test suite

### Commits signed with GPG?

No

### DIFF
--- a/salt/transport/tcp.py
+++ b/salt/transport/tcp.py
@@ -32,7 +32,7 @@ import salt.transport.client
 import salt.transport.server
 import salt.transport.mixins.auth
 import salt.ext.six as six
-from salt.ext.six.moves import queue
+from salt.ext.six.moves import queue  # pylint: disable=import-error
 from salt.exceptions import SaltReqTimeoutError, SaltClientError
 from salt.transport import iter_transport_opts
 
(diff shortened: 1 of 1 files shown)


DmitryKuzmenko (approved):

gtmanfred (approved):

salt/transport/tcp.py
```diff
@@ -556,6 +557,12 @@ def close(self):
                     raise exc
             self._socket.close()
             self._socket = None
+        if hasattr(self.req_server, 'stop'):
+            try:
+                self.req_server.stop()
+            except Exception as exc:
+                log.debug("TCPReqServerChannel close generated an exception: %s" str(exc))
```
rallytime: This has a syntax error: there is a missing comma after `%s"`.

salt/transport/tcp.py
```diff
@@ -556,6 +557,11 @@ def close(self):
                     raise exc
             self._socket.close()
             self._socket = None
+        if hasattr(self.req_server, 'stop'):
+            try:
+                self.req_server.stop()
+            except Exception as exc:
+                log.debug("TCPReqServerChannel close generated an exception: %s", str(exc))
```
cachedout: I think `log.exception` would be better here.

cachedout (approved):
I'll approve this but with a comment that can be handled in a follow-up PR.

salt/transport/tcp.py
dwoz: I think I can get rid of that completely by checking is_alive in the `stop` method. If not, I'll change this to log exception.

Issue #44998, opened 2017-12-14
Add Windows Firewall Tests to Auto Test Suite


Ch3LL:
This task is to monitor the addition of six tests to the auto test suite:
Windows:

- [ ] firewall.get_config
- [ ] firewall.disable
- [ ] firewall.enable
- [ ] firewall.get_rule
- [ ] firewall.add_rule
- [ ] firewall.delete_rule

PR #47311, opened 2018-04-25
Add firewall execution modules tests for windows

Ch3LL:
### What does this PR do?
Add firewall execution modules tests for windows.

https://github.com/saltstack/salt/issues/44998

### DIFF
--- a/tests/integration/modules/test_firewall.py
+++ b/tests/integration/modules/test_firewall.py
@@ -1,0 +1,110 @@
+# -*- coding: utf-8 -*-
+
+# Import Python libs
+from __future__ import absolute_import
+
+# Import Salt Testing libs
+from tests.support.case import ModuleCase
+from tests.support.unit import skipIf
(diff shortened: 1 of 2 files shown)


tests/integration/modules/test_firewall.py
```diff
@@ -0,0 +1,110 @@
...
+# Import Salt Libs
+import salt.utils
+
+
+@skipIf(not salt.utils.is_windows(), 'Tests for only Windows')
+class FirewallTest(ModuleCase):
+    '''
+    Validate windows firewall module
+    '''
+    def _pre_firewall_status(self, pre_run):
```
cachedout: So, given this I almost wonder if we should wrap some of the calls below in try/except/finally statements where the finally calls this. That way, if an exception is raised by salt, we'll at least reset the FW for the next test before we go on.

cachedout (approved):

twangboy (approved):

PR #47313, opened 2018-04-25
Add Poky/Yocto support

Circuitsoft:
### What does this PR do?
Adds Poky as a distribution, and supports opkg

Haven't made an issue for this, but it's pretty simple.

### Previous Behavior
Salt does not use opkg on Poky/Yocto images

### New Behavior
Salt uses opkg on Poky/Yocto images

### Tests written?
No

### Commits signed with GPG?
No


salt/modules/opkg.py
```diff
@@ -56,9 +56,11 @@ def __virtual__():
     '''
     Confirm this module is on a nilrt based system
     '''
-    if __grains__.get('os_family', False) == 'NILinuxRT':
+    try:
+        os.listdir(OPKG_CONFDIR)
```
isbm: Why would you `os.listdir` it only in order to know that the directory is there? This generates lots of stats for literally nothing. Why not just:

```python
def __virtual__():
    return not os.path.isdir(OPKG_CONFDIR) and (False, "Not OpenEmbedded system") or True

File: salt/states/pkgrepo.py [hunk 2]
```diff
@@ -344,7 +344,7 @@ def managed(name, ppa=None, **kwargs):
             if disabled is not None \
             else salt.utils.is_true(enabled)
 
-    elif os_family == 'nilinuxrt':
+    elif os_family in ('nilinuxrt', 'poky'):
```
isbm: It is "NILinuxRT" or "lilinuxrt"?

isbm (changes requested):

salt/states/pkgrepo.py
Circuitsoft: Previously in that code (not by me) the os_family variable is "normalized" with .lower().

salt/modules/opkg.py
Circuitsoft: Ah. Forgot about that method. Will do.

salt/states/pkgrepo.py
isbm: Could you please fix this? Such "normalisation" usually leads to a nasty bugs than later we have all over the place.

cachedout (changes requested):
I'd like to see the changes recommended by @isbm go in before we merge this.

salt/states/pkgrepo.py
Circuitsoft: So, it seems that the .lower() is originally from e7fb3095ceeecc1e5507754515a34e2d498a6e75. Was there a reason for it then?

cachedout:
@isbm Re-review requested.

salt/modules/opkg.py
```diff
@@ -56,9 +56,9 @@ def __virtual__():
     '''
     Confirm this module is on a nilrt based system
     '''
-    if __grains__.get('os_family', False) == 'NILinuxRT':
+    if os.path.isdir(OPKG_CONFDIR):
         return __virtualname__
-    return (False, "Module opkg only works on nilrt based systems")
+    return (False, "Module opkg only works on OpenEmbedded based systems")
```
isbm: These parenthesis aren't needed.

salt/states/pkgrepo.py
```diff
@@ -333,7 +332,7 @@ def managed(name, ppa=None, **kwargs):
             if enabled is not None \
             else salt.utils.is_true(disabled)
 
-    elif os_family in ('redhat', 'suse'):
+    elif __grains__['os_family'] in ('RedHat', 'Suse', 'SUSE'):
```
isbm: The all uppercase `os_family` "SUSE" is invalid, it is only 'Suse'. If someone is using something else, then this is wrong.

salt/states/pkgrepo.py
```diff
@@ -412,14 +411,14 @@ def managed(name, ppa=None, **kwargs):
...
-            elif kwarg == 'comments' and os_family == 'redhat':
+            elif kwarg == 'comments' and __grains__['os_family'] == 'RedHat':
                 precomments = salt.utils.pkg.rpm.combine_comments(pre[kwarg])
                 kwargcomments = salt.utils.pkg.rpm.combine_comments(
                         sanitizedkwargs[kwarg])
                 if precomments != kwargcomments:
                     break
             else:
-                if os_family in ('redhat', 'suse') \
+                if __grains__['os_family'] in ('RedHat', 'Suse', 'SUSE') \
```
isbm: Same here. It is "Suse".

isbm (changes requested):

salt/states/pkgrepo.py
Circuitsoft: I was looking at the grains source to figure out what all possible values could be, because I didn't know the exact capitalization since the original code used .lower().

salt/states/pkgrepo.py
Circuitsoft: Ok. I'll change that, and hope I don't break anything.

cachedout:
@isbm Can you please take another look here?

isbm (approved):
@Circuitsoft thanks, well done! @cachedout sorry, I was on a short vacations. :wink:

PR #47317, opened 2018-04-25
Do not join a thread that is stopped

dwoz:
### What does this PR do?

Remove the need for exception handling by not raising exceptions in the
first place.

Follow up PR to #47279

### Previous Behavior

Multiple calls to `LoadBalancedWorker.stop` would raise exceptions

### New Behavior

The stop method can be called multiple times without raising exceptions. This remove the need for special exception handling elsewhere.

### Tests written?

No - Covered by multiple existing tests

### Commits signed with GPG?

Yes


rallytime:
re-run py

DmitryKuzmenko (approved):

dwoz:
Based on the tests failures in the py3 builds, I do not think this patch will work. I'm looking into tother options.

rallytime (approved):

PR #47326, opened 2018-04-26
Some Redis fixes

The-Loeki:
### What does this PR do?
Assortment of small fixes to the Redis cache & returner

### Previous Behavior
* `KeyError` stacktrace during module loading if `cluster_mode` not in conf
* doc generation skips certain codeblocks
* can't use UNIX sockets

### New Behavior
* `cluster_mode` defaults to False for `__virtual__`
* all doc blocks render correctly
* can use `unix_socket_path` to access Redis through socket 

For `redis_return` I had to refactor the pool code a bit, it now looks *amazingly* similar to the one in `redis_cache`. Fancy that ;) maybe it should be in `utils` so all Redis share the same codebase?

### DIFF
--- a/salt/cache/redis_cache.py
+++ b/salt/cache/redis_cache.py
@@ -205,6 +205,7 @@ def _get_redis_cache_opts():
     return {
         'host': __opts__.get('cache.redis.host', 'localhost'),
         'port': __opts__.get('cache.redis.port', 6379),
+        'unix_socket_path': __opts__.get('cache.redis.unix_socket_path', None),
         'db': __opts__.get('cache.redis.db', '0'),
         'password': __opts__.get('cache.redis.password', ''),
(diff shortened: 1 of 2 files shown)


salt/cache/redis_cache.py
```diff
@@ -205,6 +205,7 @@ def _get_redis_cache_opts():
     return {
         'host': __opts__.get('cache.redis.host', 'localhost'),
         'port': __opts__.get('cache.redis.port', 6379),
+        'unix_socket_path': __opts__.get('cache.redis.unix_socket_path', None),
```
gtmanfred: please document this option.

gtmanfred (changes requested):

salt/cache/redis_cache.py
The-Loeki: will do

gtmanfred (approved):

PR #47334, opened 2018-04-26
pillar_ldap: Fix cryptic errors when config file fails to load

terminalmage:
This fixes an UnboundLocalError when the jinja templating or yaml loading fails. It also re-adds some previously-deleted crucial documentation regarding the `search_order` option.

Finally, it fixes some incorrectly-formatted RST and makes the result example a python data structure.

We still need "split mode" documentation, but this was left undone with a TODO telling the user to reference a docstring that isn't visible in the docs. so I removed that part for now.

### DIFF
--- a/salt/pillar/pillar_ldap.py
+++ b/salt/pillar/pillar_ldap.py
@@ -50,60 +50,76 @@ possible to reference grains within the configuration.
     to trick the master into returning secret data.
     Use only the 'id' grain which is verified through the minion's key/cert.
 
+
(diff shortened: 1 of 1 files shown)


gtmanfred (approved):

cachedout (approved):

terminalmage:
I forgot to return an empty dict when no `search_order` was configured. This has been done in https://github.com/saltstack/salt/pull/47369

PR #47339, opened 2018-04-26
Use salt.utils.fopen for line ending consistancy

dwoz:
### What does this PR do?

Fixes failing python 3 test case on windows.

https://jenkinsci.saltstack.com/job/2017.7/view/Python3/job/salt-windows-2016-py3/4/testReport/junit/unit.modules.test_ssh/SSHAuthKeyTestCase/test_replace_auth_key/

### Tests written?

No - Fix existing test

### Commits signed with GPG?

Yes

### DIFF
--- a/tests/unit/modules/test_ssh.py
+++ b/tests/unit/modules/test_ssh.py
@@ -93,10 +93,8 @@ class SSHAuthKeyTestCase(TestCase, LoaderModuleMockMixin):
         comment_line = '# this is a comment \n'
 
         # Write out the authorized key to a temporary file
-        if salt.utils.is_windows():
-            temp_file = tempfile.NamedTemporaryFile(delete=False)
-        else:
-            temp_file = tempfile.NamedTemporaryFile(delete=False, mode='w+')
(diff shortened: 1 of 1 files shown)


tests/unit/modules/test_ssh.py
```diff
@@ -93,18 +93,17 @@ def test_replace_auth_key(self):
...
+        temp_file = tempfile.NamedTemporaryFile(delete=False, mode='w+')
         temp_file.close()
 
+        with salt.utils.fopen(temp_file.name, 'w') as _fh:
+            # Add comment
+            _fh.write(comment_line)
+            # Add empty line for #41335
+            _fh.write(empty_line)
+            _fh.write('{0} {1} {2} {3}'.format(options, enc, key, email))
+            _fh.close()
```
rallytime: I don't think you need this `close()` since the context handler should do that for you, but I don't think it hurts either.

rallytime (approved):

tests/unit/modules/test_ssh.py
dwoz: Good catch, I'll remove it.

PR #47356, opened 2018-04-27
Fix sysctl translate

robinro:
### What does this PR do?
Fix sysctl state failing on either unicode or non-unicode names.

This is a follow-up to 72e6b31314a3439a00f435081cd92dab70d9af45 and 61c5397805a26f2b3ec59bd9bb4eadfe68a3f7ad

Essentially both 2017.7 and 2018.3 don't work, since in one case unicode names are broken, in the other the non-unicode names. This PR should ensure all if-else branches work.

### Previous Behavior
sysctl.present broken

### New Behavior
sysctl.present works

### Tests written?
No

### Commits signed with GPG?
No

### DIFF
--- a/salt/modules/linux_sysctl.py
+++ b/salt/modules/linux_sysctl.py
@@ -129,13 +129,10 @@ def assign(name, value):
         tran_tab = name.translate(''.maketrans('./', '/.'))
     else:
         if isinstance(name, unicode):  # pylint: disable=incompatible-py3-code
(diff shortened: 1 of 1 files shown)


terminalmage (approved):
@rallytime this needs a backport

rallytime:
@robinro This won't back-port cleanly to 2017.7, so I have only back-ported it to 2018.3.

PR #47375, opened 2018-04-27
Warn on use of virtual packages in pkg.installed state

terminalmage:
See https://github.com/saltstack/salt/pull/47263

Note that this will need to be removed once merged forward to develop, I've opened https://github.com/saltstack/salt/issues/47396 to track this task.

### DIFF
--- a/salt/states/pkg.py
+++ b/salt/states/pkg.py
@@ -534,7 +534,6 @@ def _find_install_targets(name=None,
                 'result': False,
                 'comment': exc.strerror}
 
-
     if salt.utils.is_windows() and kwargs.pop('refresh', False):
         # We already refreshed when we called pkg.list_pkgs
         was_refreshed = True
--- a/salt/states/pkg.py
+++ b/salt/states/pkg.py
@@ -153,6 +153,16 @@ def __virtual__():
     return 'pkg.install' in __salt__
 
 
+def _warn_virtual(virtual):
+    return [
+        'The following package(s) are "virtual package" names: {0}. These '
+        'will no longer be supported as of the Fluorine release. Please '
+        'update your SLS file(s) to use the actual package name.'.format(
+            ', '.join(virtual)
+        )
+    ]
+
+
 def _get_comparison_spec(pkgver):
     '''
     Return a tuple containing the comparison operator and the version. If no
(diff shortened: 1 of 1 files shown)


salt/states/pkg.py
```diff
@@ -582,22 +601,28 @@ def _find_install_targets(name=None,
...
                 # The package is installed and is the correct version
-                return {'name': name,
-                        'changes': {},
-                        'result': True,
-                        'comment': 'Version {0} of package \'{1}\' is already '
-                                   'installed'.format(version, name)}
+                ret = {'name': name,
+                       'changes': {},
+                       'result': True,
+                       'comment': 'Version {0} of package \'{1}\' is already '
```
isbm: BTW, sometimes we use single quotes, sometimes double quotes, sometimes none at all in these situations. Maybe at least we can get rid of those ugly back-slashes in the code. So either one looks OK to me:

```python
'Version {0} of package {1} is already '  # This is your preferred, seems like
```

```python
"Version {0} of package '{1}' is already "
```

```python
'Version {0} of package "{1}" is already '
```

:wink:

salt/states/pkg.py
```diff
@@ -1720,6 +1753,22 @@ def installed(
...
+                    cver = new_pkgs.get(pkgname, [])
+                    if '1' in cver:
+                        virtual_pkgs.append(pkgname)
+                if virtual_pkgs:
+                    warnings.extend(_warn_virtual(virtual_pkgs))
+            except Exception:
+                # This is just some temporary code to warn the user about using
+                # virtual packages. Don't let an exception break the entire
+                # state.
+                pass
```
isbm: Can we at leaast `log.debug` here this until the code goes away?

isbm (changes requested):

salt/states/pkg.py
terminalmage: Good idea, I've added debug logging here.

rallytime (approved):

salt/states/pkg.py
isbm: Well, then sounds like  Option 1 is preferred for you indeed.

isbm (approved):

cachedout:
@terminalmage Quick lint fix requested please: https://jenkins.saltstack.com/job/PR/job/salt-pr-lint-n/21436/violations/file/salt/states/pkg.py/

Issue #47124, opened 2018-04-17
Vault module problem in 2018.3.0


mchugh19:
### Description of Issue/Question
vault.read_secret fails when 2017 minion runs against a 2018.3 salt-master. 

Minion:
```
[ERROR   ] Failed to get token from master! Response is not a dict: Exception occurred in runner vault.generate_token: Traceback (most recent call last):
      File "/usr/lib/python2.7/site-packages/salt/client/mixins.py", line 387, in _low
        data['return'] = self.functions[fun](*args, **kwargs)
      File "/usr/lib/python2.7/site-packages/salt/runners/vault.py", line 46, in generate_token
        _validate_signature(minion_id, signature, impersonated_by_master)
      File "/usr/lib/python2.7/site-packages/salt/runners/vault.py", line 129, in _validate_signature
        'Could not validate token request from {0}'.format(minion_id)
    AuthenticationError: Could not validate token request from **MINION-ID**
```

And master logs sometimes show 
```
2018-04-17 16:17:45,168 [salt.loaded.ext.utils.vault:45  ][ERROR   ][9979] Failed to get token from master! An error was returned: list indices must be integers, not unicode
2018-04-17 16:17:45,169 [salt.loaded.ext.module.vault:47  ][ERROR   ][9979] Failed to read secret! CommandExecutionError: {u'error': u'list indices must be integers, not unicode'}
```

In trying to troubleshoot the _validate_signature function, I added debug lines to print out the minion signature and got:
```
2018-04-17 16:26:06,114 [salt.loaded.int.runners.vault:126 ][ERROR   ][9990] Validating signature for **MINION-ID**
2018-04-17 16:26:06,115 [salt.loaded.int.runners.vault:128 ][ERROR   ][9990] Got signature q����r�"w��|��A����f�x�RO-����kc6�   ݨS6]c�֯�q�ӑ�70�@%5�"�\�=gE̓�/�fEp�?JI*����nݶQ";�߭��dU�9�#��C�[$� ���7�~(j�S�4#�4�M|L���?�u�.uT�T�~Y�~�
�4�ňʅhkH���E�          ,�j�`=   � b�!d$�t�O�.)sΩ,��H
```

At this stage I'm guessing there is an issue with the strings being passed, and the recent unicode updates from 2018. 

<del>Additionally, to get this far, I had to correct some typos in the vault runner code for which I'll submit a PR. </del>
[...]

PR #47379, opened 2018-04-27
Properly encode messages when creating/validating signatures with m2crypto

dwoz:
### What does this PR do?

Properly encode messages when creating/validating signatures with m2crypto

#47124

### Previous Behavior

M2Crypto sign and verify methods are not encoding messages while other crypto libraries are. This leads to issues validating signatures when M2Crypto is installed.

### New Behavior

Properly encode messages when creating and validating signatures. Test sign/verify functions with both unicode and bytes.

### Tests written?

Yes

### Commits signed with GPG?

Yes

### DIFF
--- a/tests/unit/test_crypt.py
+++ b/tests/unit/test_crypt.py
@@ -275,7 +275,6 @@ class TestM2CryptoRegression47124(TestCase):
         with patch('salt.utils.files.fopen', mock_open(read_data=PUBKEY_DATA)):
             salt.crypt.verify_signature('/keydir/keyname.pub', message, self.SIGNATURE)
 
-
     @skipIf(not HAS_M2, "Skip when m2crypto is not installed")
     def test_m2crypto_verify_unicode(self):
         message = salt.utils.stringutils.to_bytes('meh')
@@ -289,7 +288,6 @@ class TestM2CryptoRegression47124(TestCase):
         with patch('salt.crypt.get_rsa_key', return_value=key):
             signature = salt.crypt.sign_message('/keydir/keyname.pem', message, passphrase='password')
         self.assertEqual(signature, self.SIGNATURE)
-
 
     @skipIf(not HAS_M2, "Skip when m2crypto is not installed")
     def test_m2crypto_sign_unicode(self):
--- a/tests/unit/test_crypt.py
+++ b/tests/unit/test_crypt.py
@@ -282,3 +282,19 @@ class TestM2CryptoRegression47124(TestCase):
         with patch('salt.utils.files.fopen', mock_open(read_data=PUBKEY_DATA)):
             salt.crypt.verify_signature('/keydir/keyname.pub', message, self.SIGNATURE)
 
+    @skipIf(not HAS_M2, "Skip when m2crypto is not installed")
+    def test_m2crypto_sign_bytes(self):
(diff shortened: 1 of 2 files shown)


garethgreenaway (approved):

tests/unit/test_crypt.py
```diff
@@ -250,3 +250,49 @@ def test_crypto_bad_key(self):
...
+        '\xd0\xbd\x1a7\xfc\xde\xe6\xb0\xad!\x12#\xc86Y\xea\xc5\xe3\xe2\xb3'
+        '\xc9\xaf\xfa\x0c\xf2?\xbf\x93w\x18\x9e\x0b\xa2a\x10:M\x05\x89\xe2W.Q'
+        '\xe8;yGT\xb1\xf2\xc6A\xd2\xc4\xbeN\xb3\xcfS\xaf\x03f\xe2\xb4)\xe7\xf6'
+        '\xdbs\xd0Z}8\xa4\xd2\x1fW*\xe6\x1c"\x8b\xd0\x18w\xb9\x7f\x9e\x96\xa3'
+        '\xd9v\xf7\x833\x8e\x01'
+    )
+
+    @skipIf(not HAS_M2, "Skip when m2crypto is not installed")
+    def test_m2crypto_verify_bytes(self):
+        message = salt.utils.stringutils.to_unicode('meh')
```
DmitryKuzmenko: `to_unicode` is intended to convert runtime data. If we operate with a constant string it's better to use `u` and `b` modifier.
But the best is to use `from __future__ import unicode_literals` as we do in the code and adopt this test for default unicode constant stirngs.

tests/unit/test_crypt.py
```diff
@@ -250,3 +250,49 @@ def test_crypto_bad_key(self):
...
+
+    @skipIf(not HAS_M2, "Skip when m2crypto is not installed")
+    def test_m2crypto_verify_unicode(self):
+        message = salt.utils.stringutils.to_bytes('meh')
+        with patch('salt.utils.files.fopen', mock_open(read_data=PUBKEY_DATA)):
+            salt.crypt.verify_signature('/keydir/keyname.pub', message, self.SIGNATURE)
+
+    @skipIf(not HAS_M2, "Skip when m2crypto is not installed")
+    def test_m2crypto_sign_bytes(self):
+        message = salt.utils.stringutils.to_unicode('meh')
```
DmitryKuzmenko: The same is here

tests/unit/test_crypt.py
DmitryKuzmenko: @terminalmage could you please advance what's the proper way. I'm just not sure do we apply the same unicode guideline to tests as we do for the code?

DmitryKuzmenko (commented):
Sorry for late review, but I have some comments (minor though)

tests/unit/test_crypt.py
terminalmage: We should be importing `unicode_literals`, but we'd need to make sure that tests that need bytes use `b` or else that'd break Python 2. Though, without the `b` it'd already be failing on Python 3, so we're probably good there.

rallytime: @dwoz I think this is fine that the PR was merged already, but can you address the concerns raised above in a follow-up PR?

PR #47398, opened 2018-04-30
Add state modules for Zabbix templates, actions and valuemaps

slivik:
### What does this PR do?
Adds functions to existing Zabbix module, test and docs related to newly added state modules for Zabbix templates, actions and value maps.

None

### Tests written?

Yes

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/modules/zabbix.py
+++ b/salt/modules/zabbix.py
@@ -23,22 +23,79 @@ Support for Zabbix
 
 :codeauthor: Jiri Kotlin <jiri.kotlin@ultimum.io>
 '''
-from __future__ import absolute_import, print_function, unicode_literals
+from __future__ import absolute_import
+from __future__ import print_function
+from __future__ import unicode_literals
+
 # Import python libs
 import logging
 import socket
+import json
+import os
 
 # Import salt libs
-from salt.ext import six
-import salt.utils.http
-import salt.utils.json
-import salt.utils.path
-from salt.utils.versions import LooseVersion as _LooseVersion
-from salt.ext.six.moves.urllib.error import HTTPError, URLError  # pylint: disable=import-error,no-name-in-module
+try:
+    import salt.utils
+    from salt.ext import six
+    from salt.utils.versions import LooseVersion as _LooseVersion
+    # pylint: disable=import-error,no-name-in-module,unused-import
+    from salt.ext.six.moves.urllib.error import HTTPError, URLError
+    from salt.exceptions import SaltException
+    IMPORTS_OK = True
+except ImportError:
+    IMPORTS_OK = False
 
 log = logging.getLogger(__name__)
 
 INTERFACE_DEFAULT_PORTS = [10050, 161, 623, 12345]
+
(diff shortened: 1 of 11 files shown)


cachedout:
@jirikotlin Can you look some of this over? I know you've done a lot of work in this area.

salt/modules/zabbix.py
```diff
@@ -2100,10 +2333,86 @@ def run_query(method, params, **connection_args):
...
             raise KeyError
     except KeyError:
         return ret
+
+
+def configuration_import(config_file, rules=None, file_format='xml', **connection_args):
+    '''
+    .. versionadded:: 2017.7
+
+    Imports Zabbix configuration sepcified in file to Zabbix server.
```
mchugh19: Spelling: specified

salt/modules/zabbix.py
```diff
@@ -846,8 +1070,8 @@ def host_create(host, groups, interfaces, **connection_args):
     .. code-block:: bash
 
         salt '*' zabbix.host_create technicalname 4
-        interfaces='{type: 1, main: 1, useip: 1, ip: "192.168.3.1", dns: "", port: 10050}'
-        visible_name='Host Visible Name' inventory_mode=0 inventory='{"alias": "something"}'
+        interfaces='{if_type: 1, main: 1, useip: 1, ip_: "192.168.3.1", dns: "", port: 10050}'
```
mchugh19: What is the gain from the change in user facing syntax? This line also clobbers an example of setting inventory information.

slivik: Sorry, my mistake. Will fix it in next commit.

mchugh19: The Zabbix API uses the following for interfaces in host.create:
```
"interfaces": [
            {
                "type": 1,
                "main": 1,
                "useip": 1,
                "ip": "192.168.3.1",
                "dns": "",
                "port": "10050"
            }
        ]
```

Why the change to if_type and ip_? This will affect user configurations and existing states, so is the rename necessary?

slivik: You are right, it was a typo. Please, see my last [commit](https://github.com/saltstack/salt/pull/47398/commits/612f2a5f207db90254ff6259e01356d017ab62e3#diff-48649b036902f9e3cd75a774e29e1b91R1072) where it is fixed.

salt/modules/zabbix.py
```diff
@@ -1466,7 +1689,7 @@ def hostinterface_get(hostids, **connection_args):
         return ret
 
 
-def hostinterface_create(hostid, ip, dns='', main=1, type=1, useip=1, port=None, **connection_args):
+def hostinterface_create(hostid, ip_, dns='', main=1, if_type=1, useip=1, port=None, **connection_args):
```
mchugh19: Should this still be ip_ and if_type?

slivik: Yes, in this case it is correct since Salt coding style (and pylint) says at least three characters for variable name and "type" is reserved Python keyword.

PR #47408, opened 2018-04-30
Skip status.diskusage integration tests on macsox

Ch3LL:
### What does this PR do?
The execution modules `status.diskusage` currently only returns `{}` on macosx because there is no logic in the code currently to support this operating system. Skipping the test for now until functionality is added for mac.

https://github.com/saltstack/salt-jenkins/issues/939


rallytime (approved):

tests/integration/modules/test_status.py
```diff
@@ -44,6 +44,7 @@ def test_status_saltmem(self):
         ret = self.run_function('status.saltmem')
         self.assertTrue(isinstance(ret, int))
 
+    @skipIf(salt.utils.is_darwin(), 'status.diskusage not currently supported on macosx')
```
cachedout: Is the module itself properly accounting for this and returning an appropriate warning? If not, shouldn't it be?

Ch3LL: agreed it should return an error. Currently it just returns {} if its not supported. Will add a warning.

cachedout (approved):

Issue #47125, opened 2018-04-17
Runas Causing "invalid directory" error on Windows.


doesitblend:
### Description of Issue/Question
When using `cmd.script` with the `runas` option I am getting an error "The current directory is invalid". 

### Setup
Install Salt 2017.7.5 on Windows 2016. 

Create a test user using `salt '<test minion>' user.add borneo <some password>`. 

### Steps to Reproduce Issue
Run the following state on the newly configured minion from above.

```
run_test_script:
  cmd.script:
    - source: salt://mytestscript.ps1
    - shell: powershell
    - args: '"this is a string"'
    - runas: {{ pillar['temp-db']['username'] }}
    - password: {{ pillar['temp-db']['password'] }}
```

[mytestscript.ps1.tar.gz](https://github.com/saltstack/salt/files/1921430/mytestscript.ps1.tar.gz)
[testme.sls.tar.gz](https://github.com/saltstack/salt/files/1921431/testme.sls.tar.gz)

### Versions Report
```
    Salt Version:
               Salt: 2017.7.5
     
    Dependency Versions:
               cffi: 1.10.0
           cherrypy: 10.2.1
           dateutil: 2.6.0
          docker-py: Not Installed
              gitdb: 2.0.3
          gitpython: 2.1.3
              ioflo: Not Installed
             Jinja2: 2.9.6
            libgit2: Not Installed
            libnacl: 1.6.1
           M2Crypto: Not Installed
               Mako: 1.0.6
       msgpack-pure: Not Installed
     msgpack-python: 0.4.8
       mysql-python: Not Installed
          pycparser: 2.17
           pycrypto: 2.6.1
       pycryptodome: Not Installed
             pygit2: Not Installed
             Python: 2.7.13 (v2.7.13:a06454b1afa1, Dec 17 2016, 20:53:40) [MSC v.1500 64 bit (AMD64)]
       python-gnupg: 0.4.0
             PyYAML: 3.12
              PyZMQ: 16.0.2
               RAET: Not Installed
              smmap: 2.0.3
            timelib: 0.2.4
            Tornado: 4.5.1
                ZMQ: 4.1.6
     
    System Versions:
               dist:   
             locale: cp1252
            machine: AMD64
            release: 2016Server
             system: Windows
            version: 2016Server 10.0.14393  Multiprocessor Free
```

doesitblend:
ZD-2428

[...]

PR #47412, opened 2018-04-30
Fix issue where the cwd was being removed

twangboy:
### What does this PR do?
Fixes an issue where the passed `cwd` was being deleted by cmd.script. The only `cwd` we want to delete is the temporary one created by salt if `cwd` is not passed on Windows. The runas functionality requires a `cwd` on Windows.

https://github.com/saltstack/salt/issues/47125

### Previous Behavior
Any directory passed in cwd was being deleted on Windows

### New Behavior
Only deletes the temp cwd created by salt

### Tests written?
No

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/modules/cmdmod.py
+++ b/salt/modules/cmdmod.py
@@ -2077,7 +2077,9 @@ def script(source,
         # Create a temp working directory
         cwd = tempfile.mkdtemp(dir=__opts__['cachedir'])
         win_cwd = True
-        salt.utils.win_dacl.set_permissions(obj_name=cwd, principal=runas, permissions='full_control')
+        salt.utils.win_dacl.set_permissions(obj_name=cwd,
+                                            principal=runas,
+                                            permissions='full_control')
 
     path = salt.utils.files.mkstemp(dir=cwd, suffix=os.path.splitext(source)[1])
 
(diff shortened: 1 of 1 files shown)


salt/modules/cmdmod.py
```diff
@@ -2089,10 +2093,10 @@ def _cleanup_tempfile(path):
                                           saltenv,
                                           **kwargs)
         if not fn_:
-            if salt.utils.is_windows() and runas:
+            _cleanup_tempfile(path)
+            # If a temp working directory was created (Windows), let's remove that
+            if win_cwd:
                 _cleanup_tempfile(cwd)
```
cachedout: If I'm reading this right, doesn't `cwd` include `path`? And if so, do we really need two calls to `_cleanup_tempfile` in that case?

twangboy: It's a temporary directory created by the code above (line 2078) for the purposes of this function. After which it needs to be deleted

twangboy: If you don't pass `cwd` on a Windows box, this will create one under the `cachedir` with a temporary name. It will give the user permissions to that directory. The full path to the directory is in the `cwd` variable. The `path` variable is to a location inside the `cwd`, so if anything, we would only need to delete the `cwd` which would contain the `path`

cachedout (approved):

Issue #47416, opened 2018-05-01
slack engine running from salt-master tries to access /etc/salt/pki/minion/


viq:
### Description of Issue/Question
Slack engine, running from salt-master, tries to access `/etc/salt/pki/minion` and complains loudly when it cannot.

### Setup
`salt-master` running as a dedicated, non-root user, simple slack engine configured to run from salt-master.
https://raw.githubusercontent.com/saltstack/salt/bcdef641e8943f7a2aaeb88ad25cd3d23aac80f9/salt/engines/slack.py put in `/srv/salt/_engines/` and synced.

### Steps to Reproduce Issue
```
may 01 16:10:26 salt.master.example.com salt-master[22969]: Traceback (most recent call last):                                                                                                                               
may 01 16:10:26 salt.master.example.com salt-master[22969]:   File "/srv/deploy/salt/.venv/lib/python3.6/site-packages/salt/engines/__init__.py", line 128, in run
may 01 16:10:26 salt.master.example.com salt-master[22969]:     self.engine[self.fun](**kwargs)                                                                                                                              
may 01 16:10:26 salt.master.example.com salt-master[22969]:   File "/tmp/pip-build-1jhjfqa0/salt/salt/engines/slack.py", line 777, in start                                                                                  
may 01 16:10:26 salt.master.example.com salt-master[22969]: Exception: Traceback (most recent call last):                                                                                                                    
may 01 16:10:26 salt.master.example.com salt-master[22969]:   File "/tmp/pip-build-1jhjfqa0/salt/salt/engines/slack.py", line 775, in start
may 01 16:10:26 salt.master.example.com salt-master[22969]:   File "/tmp/pip-build-1jhjfqa0/salt/salt/engines/slack.py", line 652, in run_commands_from_slack_async                                                          
may 01 16:10:26 salt.master.example.com salt-master[22969]:   File "/tmp/pip-build-1jhjfqa0/salt/salt/engines/slack.py", line 464, in generate_triggered_messages                                                            
may 01 16:10:26 salt.master.example.com salt-master[22969]:   File "/tmp/pip-build-1jhjfqa0/salt/salt/engines/slack.py", line 205, in get_config_groups
may 01 16:10:26 salt.master.example.com salt-master[22969]:   File "/tmp/pip-build-1jhjfqa0/salt/salt/engines/slack.py", line 238, in _groups_from_pillar                                                                    
[...]

PR #47417, opened 2018-05-01
revert instantiating a Caller Client in the engine

gtmanfred:
### What does this PR do?
This Caller client prevents running the salt master as a non root user.  Also,
master processes are not assigned pillars so it is using whatever minion get
instantiated on the master from a caller client.

Instead, to use pillars, the engine should be started on in a minion process.

Fixes https://github.com/saltstack/salt/issues/47416

### Tests written?

No

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/engines/slack.py
+++ b/salt/engines/slack.py
@@ -88,6 +88,13 @@ In addition, other groups are being loaded from pillars.
 
 :depends: slackclient
 
+
+.. note:: groups_pillar_name
+
+    In order to use this, the engine must be running as of a minion running on
+    the master, so that the ``Caller`` client can be used to retrieve that
+    minions pillar data, because the master process does not have pillars.
+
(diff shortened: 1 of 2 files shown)


gtmanfred:
@pcn fyi

garethgreenaway (approved):

salt/engines/slack.py
```diff
@@ -88,6 +88,13 @@
 
 :depends: slackclient
 
+
+.. note:: groups_pillar_name
+
+    In order to use this, the engine must be running as of a minion running on
```
pcn: What does `as of a minion` mean here?

gtmanfred: put your `engines:` settings in /etc/salt/minion, and start a salt-minion process with them.

PR #47440, opened 2018-05-02
fix for the race condition, details are here: https://github.com/salt…

kt97679:
…stack/salt/issues/33223#issuecomment-386117236

### Previous Behavior

### New Behavior

### Tests written?

Yes/No

### Commits signed with GPG?

Yes/No


salt/fileclient.py
```diff
@@ -1207,11 +1213,12 @@ def get_file(self,
                             saltenv,
                             cachedir=cachedir) as cache_dest:
                         dest = cache_dest
+                        dest_tmp = "{0}/{1}".format(os.path.dirname(dest), str(uuid.uuid4()))
                         # If a directory was formerly cached at this path, then
                         # remove it to avoid a traceback trying to write the file
                         if os.path.isdir(dest):
                             salt.utils.files.rm_rf(dest)
-                        fn_ = salt.utils.files.fopen(dest, 'wb+')
+                        fn_ = salt.utils.files.fopen(dest_tmp, 'wb+')
```
terminalmage: Instead of using `dest_tmp` here and renaming manually, you should use `salt.utils.atomicfile.atomic_open()`, which was designed exactly for this sort of use case. You can change this line to:

```python
fn_ = salt.utils.atomicfile.atomic_open(dest, 'wb+')
```

Then, when the filehandle is closed, the tempfile will be automatically renamed to the `dest` path.

Note that using this method will require two additional changes:

1. `salt.utils.atomicfile` will need to be imported here in fileclient.py.
2. `salt/utils/atomicfile.py` will need to be added to salt/utils/thin.py in the `min_files` tuple so that it is part of the thin tarball.

terminalmage (changes requested):
Rather than manually creating

salt/fileclient.py
kt97679: Thanks for your comments @terminalmage , please see changes in the latest commit.

kt97679:
@terminalmage @rallytime please let me know if I need to apply any other changes to this PR.

terminalmage (approved):

PR #47442, opened 2018-05-02
Add attachment support to SMTP state and module

jeduardo:
### What does this PR do?

This PR adds support to sending out email messages containing one or more attachments specified through a new parameter called `attachments`. 

The acceptable value for this parameter is a list of files present in the local filesystem that will be added as attachments to the outgoing email sent by the Salt state/module for SMTP.

When trying to deploy some automation bits where Salt generates a file periodically (for example, a client certificate) I have faced the issue that I cannot easily send out the generated file to a recipient through email. 

The enhancement on this PR makes this particular use case possible.

### Previous Behavior

The `attachments` key is not present and is ignored by Salt. 

### New Behavior

All files in the minion filesystem specified as a list under the `attachments` key are added as individual attachments to an outgoing email message.

### Tests written?

No

### Commits signed with GPG?

No


jeduardo:
@cachedout hey! You've helped me with my last PR when Jenkins was failing. This time even the reviewer bot is not making an appearance. Any idea about what might be going on here?

jeduardo:
@cachedout it took some time specially after the linting task was not accessible for a couple days but linting problems are finally resolved! What else?

jeduardo:
@cachedout hey! Builds seem to be failing now because of issues not related to the changes suggested in this PR. Could you please review it and send forward any other suggestions? I'm really keen to get this functionality added to Salt, either as I'm proposing or in any other way. Thanks!

salt/modules/smtp.py
```diff
@@ -76,6 +82,7 @@ def send_msg(recipient,
              use_ssl='True',
              username=None,
              password=None,
+             attachments=None,
```
rallytime: Can you move this to the end of the list so we don't cause an API error?

salt/states/smtp.py
```diff
@@ -27,7 +27,13 @@ def __virtual__():
     return 'smtp' if 'smtp.send_msg' in __salt__ else False
 
 
-def send_msg(name, recipient, subject, sender=None, profile=None, use_ssl='True'):
+def send_msg(name,
+             recipient,
+             subject,
+             sender=None,
+             profile=None,
+             attachments=None,
```
rallytime: Same comment here. :)

rallytime (changes requested):

salt/modules/smtp.py
jeduardo: Ah, of course! Moved to the end, thanks for spotting it. :)

rallytime (approved):
LGTM

jeduardo:
@rallytime thanks for approving it! :)

cachedout (approved):

PR #47454, opened 2018-05-03
[WIP] enable passing a unix_socket for mysql returners

MaximilianMeister:
### What does this PR do?
enable the `unix_socket` config option for returners

### DIFF
--- a/salt/returners/mysql.py
+++ b/salt/returners/mysql.py
@@ -18,6 +18,7 @@ config. These are the defaults:
     mysql.pass: 'salt'
     mysql.db: 'salt'
     mysql.port: 3306
+    mysql.unix_socket: '/tmp/mysql.sock'
 
 SSL is optional. The defaults are set to None. If you do not want to use SSL,
 either exclude these options or set them to None.
@@ -43,6 +44,7 @@ optional. The following ssl options are simply for illustration purposes:
     alternative.mysql.ssl_ca: '/etc/pki/mysql/certs/localhost.pem'
     alternative.mysql.ssl_cert: '/etc/pki/mysql/certs/localhost.crt'
     alternative.mysql.ssl_key: '/etc/pki/mysql/certs/localhost.key'
+    alternative.mysql.unix_socket: '/tmp/mysql.sock'
 
 Should you wish the returner data to be cleaned out every so often, set
 `keep_jobs` to the number of hours for the jobs to live in the tables.
@@ -198,7 +200,8 @@ def _get_options(ret=None):
                 'port': 3306,
                 'ssl_ca': None,
                 'ssl_cert': None,
-                'ssl_key': None}
+                'ssl_key': None,
+                'unix_socket': '/tmp/mysql.sock'}
 
     attrs = {'host': 'host',
              'user': 'user',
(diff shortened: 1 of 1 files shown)


salt/returners/mysql.py
```diff
@@ -198,7 +200,8 @@ def _get_options(ret=None):
                 'port': 3306,
                 'ssl_ca': None,
                 'ssl_cert': None,
-                'ssl_key': None}
+                'ssl_key': None,
+                'unix_socket': None}
```
cachedout: Above it says that `/tmp/mysql.sock` is the default but here it appears to be `None`?

MaximilianMeister: good point, i think the default can safely be set here, because both [MySQLdb](https://github.com/dccmx/mysqldb/blob/master/doc/MySQLdb.txt#L137) and [PyMySQL](https://github.com/PyMySQL/PyMySQL/blob/master/pymysql/connections.py#L944) will only use a `unix_socket` when host == localhost

so the default `unix_socket` will just be discarded for someone who wants to use a remote connection

salt/returners/mysql.py
```diff
@@ -261,7 +265,8 @@ def _get_serv(ret=None, commit=False):
                                    passwd=_options.get('pass'),
                                    db=_options.get('db'),
                                    port=_options.get('port'),
-                                   ssl=ssl_options)
+                                   ssl=ssl_options,
+                                   unix_socket=_options.get('unix_socket'))
```
brejoc: From glimpsing at the lib I can't really tell how MySQLdb is dealing with the kwargs here. What's happening if you pass `host`/`port` and `unix_socket` - which one is getting priority?

MaximilianMeister: from `this` explanation i get that `host` has always priority over `unix_socket`, only not when it is `localhost`. then `unix_port` snaps either with the default or the passed argument.

brejoc (approved):

MaximilianMeister:
rebased

isbm (changes requested):
@MaximilianMeister can we also standardise [this](https://github.com/saltstack/salt/blob/2018.3/salt/auth/mysql.py#L64-L100), [this](https://github.com/saltstack/salt/blob/2018.3/salt/modules/mysql.py#L329-L338), [this](https://github.com/saltstack/salt/blob/2018.3/salt/pillar/mysql.py#L103-L119) and [this](https://github.com/saltstack/salt/blob/2018.3/salt/returners/mysql.py#L234-L258) together, so we have once and for all one way to configure it? Because if we continue with this PR, we will have again a long way migrating everyone to something better.

MaximilianMeister:
> @MaximilianMeister can we also standardise this, this, this and this together, so we have once and for all one way to configure it? Because if we continue with this PR, we will have again a long way migrating everyone to something better.

@isbm agree this is a good idea, but can we do a follow up for the standardisation? i can create an issue and work on it when i have some more time

isbm:
@cachedout @rallytime I just spoke with @MaximilianMeister and we agreed to WP this PR. We will remove MySQL configuration inconsistencies, refactor the code (mostly delete redundant), add migration procedure we typically do in Salt and also provide a proper unit testing.

:warning:  Please do not merge this yet.

rallytime:
OK, that sounds fine. I am going to close this for now. We can of course open this again when you're ready to come back to it and the changes discussed above have been implemented. Thanks!

PR #47490, opened 2018-05-04
Support Jinja imports with relative paths. Issue #13889

plastikos:
Jinja imports will be interpreted as originating from the top of each
of the directories in the searchpath when the template name does not
begin with './' or '../'.  When a template name begins with './' or
'../' then the import will be relative to the importing file.

Prior practices required the following construct:

```jinja
{% from tpldir ~ '/foo' import bar %}
```

This patch supports a more "natural" construct:

```jinja
{% from './foo' import bar %}
```

Comparatively when importing from a parent directory - prior practice:

```jinja
{% from tpldir ~ '/../foo' import bar %}
```

Construct supported by this patch:

```jinja
{% from '../foo' import bar %}
```

### What does this PR do?

Supports relative Jinja imports.

#13889

### Previous Behavior

Unfortunately ``{% from './foo' import bar %}`` does not work.  This super-secret technique was required: ``{% from tpldir ~ '/foo' import bar %}``

### New Behavior

This works: ``{% from './foo' import bar %}``

### Tests written?

Yes

### Commits signed with GPG?

No


salt/utils/jinja.py
```diff
@@ -98,28 +98,60 @@ def check_cache(self, template):
...
+        Salt-specific loader to find imported jinja files.
+
+        Jinja imports will be interpreted as originating from the top
+        of each of the directories in the searchpath when the template
+        name does not begin with './' or '../'.  When a template name
+        begins with './' or '../' then the import will be relative to
+        the importing file.
+        
+        '''
+        # FIXME: somewhere do seprataor replacement: '\\' => '/'
```
plastikos: Translation of `\` => `/` might be broken.  Help me, Obi Juan, you're my only hope!

salt/utils/jinja.py
```diff
@@ -98,28 +98,60 @@ def check_cache(self, template):
...
-        if '..' in template:
-            log.warning(
-                'Discarded template path \'%s\', relative paths are '
-                'prohibited', template
-            )
-            raise TemplateNotFound(template)
+        '''
+        Salt-specific loader to find imported jinja files.
+
+        Jinja imports will be interpreted as originating from the top
```
plastikos: This import behavior needs to be in the documentation.  Where is the best location for it?

plastikos (commented):
There are a few details that should be cleaned-up before this is merged.

plastikos:
RETEST

plastikos:
Help!  Suggestions welcome!

tests/support/parser/__init__.py
```diff
@@ -332,7 +332,7 @@ def parse_args(self, args=None, values=None):
                         os.path.basename(fpath).startswith('test_'):
                     self.options.name.append(fpath)
                     continue
-                self.exit(status=1, msg='\'{}\' is not a valid test module'.format(fpath))
+                self.exit(status=1, msg='\'{}\' is not a valid test module\n'.format(fpath))
```
terminalmage: Can you help me understand this change? I don't know what adding a newline here fixes.

salt/utils/jinja.py
```diff
@@ -98,28 +98,60 @@ def check_cache(self, template):
...
+        # checks for relative '..' paths that step-out of file_roots
+        if is_relative:
+            # Starts with a relative path indicator
+
+            if not environment or 'tpldir' not in environment.globals:
+                log.warning(
+                    'Relative path "%s" cannot be resolved without an environment',
+                    template
+                )
+                print('Relative path "{0}" cannot be resolved without an environment'.format(template))
```
terminalmage: We should not be printing to the CLI here.

salt/utils/jinja.py
```diff
@@ -98,28 +98,60 @@ def check_cache(self, template):
...
+                raise TemplateNotFound
+            base_path = environment.globals['tpldir']
+            _template = os.path.normpath('/'.join((base_path, _template)))
+            if _template.split('/', 1)[0] == '..':
+                log.warning(
+                    'Discarded template path "%s": attempts to'
+                    ' ascend outside of file roots', template
```
terminalmage: Since we have multiple backends, we can't assume people will be using `file_roots`, so this log message should reference `salt://` instead.

salt/utils/jinja.py
terminalmage: @dwoz @twangboy When using jinja imports on Windows, are Windows-specific path-separators supported? If so, are they the only ones supported, or are forward slashes also supported? I would assume both are supported, since they are in Python proper, but I don't want to assume without knowing for sure.

terminalmage (changes requested):

salt/utils/jinja.py
twangboy: I use Windows separators in my Jinja. I don't even have to escape them. Jinja apparently does some escaping for you.

tests/support/parser/__init__.py
plastikos: Without the newline then the error message is not on a line by itself and the command prompt is left on the same line as the error message.

salt/utils/jinja.py
plastikos: Well that's embarrassing - that should have been removed.

salt/utils/jinja.py
plastikos: Okay.

salt/utils/jinja.py
plastikos: The ``tpldir = os.path.dirname(template).replace('\\', '/')`` construct was in the original code and appears to do some path translation on Windows.  I'm not sure how to test it to ensure that I haven't broken the functionality.  Is there a unit test?

plastikos:
@terminalmage - review changes, please.

twangboy (approved):

terminalmage (approved):

rallytime:
@plastikos Is this ready to go from your side?

If so, can you also add a mention of this addition to the [Fluorine release notes](https://github.com/saltstack/salt/blob/develop/doc/topics/releases/fluorine.rst)?

plastikos:
@rallytime - I'll add the mention in the release notes and it will be ready to go.

PR #47499, opened 2018-05-06
Move kill process tree and re-use it

dwoz:
### What does this PR do?

Handle another case where the test suite needs to time out a subprocess on Windows.

### Tests written?

No - Fixes existing tests

### Commits signed with GPG?

Yes


gtmanfred (approved):

gtmanfred:
@dwoz does this need to be added to 2017.7.6?

dwoz:
@gtmanfred This is some preparatory work for when we run the windows suite without `whitelist.txt`. This won't affect anything in the 2017.7.6 branch.

Issue #47501, opened 2018-05-06
salt-ssh doesn't propagate pillar when overriding pillar in module.run + state.apply


mateiw:
### Description of Issue/Question

When applying a state using `salt-ssh` the pillar data is replaced completely inside the second state that is applied with `module.run` + `state.apply` and explicit pillar overriding.
The initial pillar data is available in the state that is applied directly but not in the second state that's applied using `module.run` and `state.apply`. The overriding pillar is available in the second state when referenced explicitly but not in `pillar.items`.

### Setup

Patch from #47497 is needed in order to reproduce this because `--extra-filerefs` doesn't work properly in this case as the state to be applied doesn't reference the second state file using the `salt://...` prefix.

Master config:
```yaml
pillar_roots:
  base:
    - /srv/pillar

file_roots:
  base:
    - /srv/salt
```
Roster:
`/tmp/roster`:
```
dev-minsles12sp2-1.tf.local:
    host: dev-minsles12sp2-1.tf.local
    user: root
    port: 22
    timeout: 180
```

### Steps to Reproduce Issue

Pillar definition:
`/srv/pillar/top.sls`:
```yaml
base:
  '*':
    - test
```
`/srv/pillar/test.sls`:
```yaml
somekey: xyz
```

States:
`/srv/salt/testpillar.sls`:
```yaml
call_show:
  module.run:
    - name: state.apply
    - mods:
      - showpillar
    - kwargs:
          pillar:
              test: "foo bar"
```

`/srv/salt/showpillar.sls`:
```yaml
show_all_pillars:
    module.run:
    -   name: pillar.items
test_somekey:
    test.check_pillar:
    -   present:
         -  somekey
test_test:
    test.check_pillar:
    -   present:
         -  test
```
Displaying the pillar works fine:
```bash
salt-ssh  --roster-file=/tmp/roster dev-minsles12sp2-1.tf.local pillar.items
```
```
dev-minsles12sp2-1.tf.local:
    ----------
    somekey:
        xyz

```
but when applying the `testpillar` state the pillar is not propagated across `module.run` + `state.apply`:
```bash
salt-ssh --roster-file=/tmp/roster --extra-filerefs="salt://showpillar.sls" dev-minsles12sp2-1.tf.local state.apply testpillar
```

```
dev-minsles12sp2-1.tf.local:                                                                                                                                                        [44/1830]
----------
          ID: call_show
    Function: module.run
        Name: state.apply
      Result: False
[...]

PR #47504, opened 2018-05-06
Fix pillar propagation in salt-ssh when overriding pillar in module.run + state.apply

mateiw:
### What does this PR do?

Add option to merge pillar with `opts['pillar']` during pillar compilation. The options is `False` by default and only set by `salt-ssh` in order to allow for the pillar to be propagated when applying a state using `module.run` + `state.apply` and overriding the pillar explicitly. 

This problem doesn't happen with regular, non-ssh minions.

#47501

### Previous Behavior
If `pillar` is specified in `module.run` + `state.apply` it replaces any previous pillar data completely in the state that is applied instead of merging the value with the existing data.
E.g.
```yaml
call_show:
  module.run:
    - name: state.apply
    - mods:
      - showpillar
    - kwargs: 
          pillar:    # <= replaces existing pillar data completely in the showpillar state
              test: "foo bar"
```

### New Behavior
`salt-ssh` sets the option `pillar_compile_merge_global` to `True` in order to for the pillar compilation to merge the new data with the existing one.

### Tests written?

No

### Commits signed with GPG?

No


gtmanfred (approved):

salt/client/ssh/__init__.py
```diff
@@ -911,6 +911,7 @@ def __init__(
                     'id': self.id,
                     'sock_dir': '/',
                     'fileserver_list_cache_time': 3,
+                    'pillar_compile_merge_global': True
```
isbm: This gets set to `True` all the time and ignores _any kind_ of settings. I would be OK to set it to `True` by default, but completely ignoring the opposite value of it isn't the great idea. And maybe shorter name? like `merge_pillar`.

isbm (changes requested):

mateiw:
Thanks for the suggestion @isbm. Renamed to `pillar_merge_global` and made it possible to override the default value using `minion_opts` in the roster file or globally using `ssh_minion_opts` in the master config.

salt/client/ssh/__init__.py
```diff
@@ -901,6 +901,7 @@ def __init__(
         self.minion_opts = {
                     'grains_cache': True,
                     'log_file': 'salt-call.log',
+                    'pillar_merge_global': True,
```
isbm: I'd remove this and put it do the default `ssh_.....` options that comes from `ssh_minion_opts`. Besides, I would use conventional name `ssh_<do>_<what>`. That would be `ssh_merge_pillar`, and set it to `True` by default. Anyone who wants it void can set it to `False` for Salt SSH.

isbm (changes requested):
@mateiw some more updates.

isbm (approved):
@mateiw thanks, perfect! Now we're good to go.

terminalmage (changes requested):
This looks good, but we should document the new config option in doc/ref/configuration/minion.rst (with a `.. versionadded:: 2018.3.2` to note when the option was added).

mateiw:
@terminalmage I added the documentation for this new option.

doc/ref/configuration/minion.rst
```diff
@@ -3219,3 +3219,31 @@ URL of the repository:
 Replace ``<commit_id>`` with the SHA1 hash of a commit ID. Specifying a commit
 ID is useful in that it allows one to revert back to a previous version in the
 event that an error is introduced in the latest revision of the repo.
+
+``ssh_merge_pillar``
+----------------------
```
terminalmage: Can you delete the last two dashes here? Sphinx will warn on the docs build when the dashed line is longer than the line above it.

terminalmage (changes requested):
Thanks for adding the docs! Just one minor tweak needed here.

terminalmage (approved):

Issue #47364, opened 2018-04-27
ZFS grain/module unable to be reloaded during state run


ryanwalder:
### Description of Issue/Question

Unable to reload the zfs module during a state run.

#### Expected behaviour

0. run `state.apply`
0. Install zfsutils-linux package
0. reload zfs module
0. create zpool
0. `state.apply` finishes

#### Actual behaviour

0. Installs zfsutils-linux package
0. Reloads module but does not enable (because `zfs_available` grain is not updated)
0. Fails the create zpool due to module not being loaded corectly
0. `state.apply` finishes
0. run `state.apply` again and it works

Debug log: https://www.hastebin.com/himapiladu

Looks like it's unable to reload and activate the module due to relying on the grain `zfs_support` which is not being updated, changing the order of `reload_modules` and `reload_grains` in the below formula has no effect.

### Setup
```
zfs:
  pkg.installed:
    - pkgs: ['zfsutils-linux']
    - reload_grains: True
    - reload_modules: True

zfs-zpool-foo:
  zpool.present:
    - name: foo
    - layout: [{'mirror': ['/dev/sdb', '/dev/sdc']}, {'mirror': ['/dev/sdd', '/dev/sde']}, {'mirror': ['/dev/sdf', '/dev/sdg']}, {'log': ['mirror', '/dev/sdh', '/dev/sdi']}]
    - config: {'device_dir': 'None', 'force': True}
    - require:
      - pkg: zfs
```

### Steps to Reproduce Issue
Use the above

### Versions Report
```
Salt Version:
           Salt: 2018.3.0-n/a-7390b72 (installed via bootstrap `git 2018.3`)
 
Dependency Versions:
           cffi: Not Installed
       cherrypy: Not Installed
       dateutil: Not Installed
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.8
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: 0.21.1
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.4.6
   mysql-python: Not Installed
      pycparser: Not Installed
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: Not Installed
         Python: 2.7.12 (default, Dec  4 2017, 14:50:18)
   python-gnupg: Not Installed
         PyYAML: 3.11
          PyZMQ: 15.2.0
[...]

PR #47518, opened 2018-05-07
Fix 47364: ensure we are not caching zfs.is_supported

Ch3LL:
https://github.com/saltstack/salt/issues/47364

### Previous Behavior
When attempting to reload the grains after installing the zfs package during a state run it would not work properly because the decorator `@real_memoize` was caching the previous query. 

### New Behavior
refreshing grains during a state run works now for zfs_supported grain.

### Tests written?

Yes

### Commits signed with GPG?

Yes

### DIFF
--- a/tests/unit/utils/test_zfs.py
+++ b/tests/unit/utils/test_zfs.py
@@ -891,6 +891,18 @@ class ZfsUtilsTestCase(TestCase):
     This class contains a set of functions that test salt.utils.zfs utils
     '''
     ## NOTE: test parameter parsing
+    def test_is_supported(self):
+        '''
+        Test zfs.is_supported method
+        '''
+        for value in [False, True]:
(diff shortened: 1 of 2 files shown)


tests/unit/utils/test_zfs.py
```diff
@@ -891,6 +891,18 @@ class ZfsUtilsTestCase(TestCase):
     This class contains a set of functions that test salt.utils.zfs utils
     '''
     ## NOTE: test parameter parsing
+    def test_is_supported(self):
+        '''
+        Test zfs.is_supported method
+        '''
+        for value in [False, True]:
+            with patch.object(zfs, '_zpool_cmd', MagicMock(return_value=value)):
```
sjorge: We probably don’t need this one as we removed it from the is_supported function.

sjorge (approved):
LGTM

rallytime (approved):

PR #47524, opened 2018-05-07
Gracefully handle blank lines in whitelist.txt

dwoz:
### What does this PR do?

Handle the case of having extra spaces and/or a blank lines in whitelist.txt. This can happen if one manually edits whitelist.txt during development. In those cases the test suite ends with an Exception which can be fairly annoying.

### Commits signed with GPG?

Yesgning commits with GPG.


gtmanfred (approved):

dwoz:
rerun py

dwoz:
re-run py

Issue #47178, opened 2018-04-19
TypeError: makedirs_() [...] for file.symlink and file.shortcut


asymetrixs:
Dear All,

I have a problem with file.symlink and file.shortcut:
```
file.symlink:
  - name: 'C:\ProgramData\Microsoft\Windows\Start Menu\test\link.lnk'
  - target: '{{ salt['environ.get']('PROGRAMFILES') }}\test\app.exe'
  - makedirs: True
```
Same happens for `file.shortcut`
```
Returns
TypeError: makedirs_() got an unexpected keyword argument 'user'
```
Running salt `2017.7.4` on Windows 7

Also referencing #42421

gtmanfred:
Can you run this with `salt-call -l debug` from the minion so that we can get the full stacktrace.

Also, please provide the versions report like the issues template requested.

Thanks,
Daniel

gtmanfred:
@twangboy can you take a look at this?

It seems like file.symlink and file.shortcut need the same treatment as file.directory and others.

```
                if salt.utils.is_windows():
                    # Make sure the drive is mapped before trying to create the
                    # path in windows
                    drive, path = os.path.splitdrive(name)
                    if not os.path.isdir(drive):
                        return _error(
                            ret, 'Drive {0} is not mapped'.format(drive))
                    __salt__['file.makedirs'](name, win_owner, win_perms,
                                              win_deny_perms, win_inheritance)
                else:
                    __salt__['file.makedirs'](name, user=user, group=group,
                                              mode=dir_mode)

```

gtmanfred:
@asymetrixs for now, the work around would be to use a file.directory before the file.symlink or file.shortcut, and not use makedirs.

Thanks,
Daniel

PR #47533, opened 2018-05-08
Fix issues with functions that use makedirs on Windows

twangboy:
### What does this PR do?
Fixes issues with the file state module when `makedirs` is True on Windows

https://github.com/saltstack/salt/issues/47178

### Previous Behavior
Functions would stack trace on Windows due to differences in the parameters expected by the `file.makedirs` and `file.makedirs_perms` functions on Windows and Linux.

### New Behavior
It passes... on Windows...

### Tests written?
No

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/states/file.py
+++ b/salt/states/file.py
@@ -663,13 +663,13 @@ def _error(ret, err_msg):
 
 
 def _check_directory(name,
-                     user,
-                     group,
-                     recurse,
-                     mode,
-                     clean,
-                     require,
-                     exclude_pat,
+                     user=None,
+                     group=None,
+                     recurse=False,
+                     mode=None,
+                     clean=False,
+                     require=False,
+                     exclude_pat=None,
                      max_depth=None,
                      follow_symlinks=False):
     '''
@@ -754,7 +754,7 @@ def _check_directory(name,
 
 
 def _check_directory_win(name,
-                         win_owner,
+                         win_owner=None,
                          win_perms=None,
                          win_deny_perms=None,
                          win_inheritance=None):
@@ -767,9 +767,10 @@ def _check_directory_win(name,
         changes = {name: {'directory': 'new'}}
     else:
         # Check owner
-        owner = salt.utils.win_dacl.get_owner(name)
-        if not owner.lower() == win_owner.lower():
-            changes['owner'] = win_owner
+        if win_owner is not None:
+            owner = salt.utils.win_dacl.get_owner(name)
+            if not owner.lower() == win_owner.lower():
+                changes['owner'] = win_owner
 
         # Check perms
         perms = salt.utils.win_dacl.get_permissions(name)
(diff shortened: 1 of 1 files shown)


salt/states/file.py
```diff
@@ -663,13 +663,13 @@ def _error(ret, err_msg):
...
 
 def _check_directory(name,
-                     user,
-                     group,
-                     recurse,
-                     mode,
-                     clean,
-                     require,
-                     exclude_pat,
+                     user=None,
```
cachedout: This scares me a little bit. Why is it necessary to set all the defaults suddenly?

salt/states/file.py
```diff
@@ -1200,6 +1214,57 @@ def _shortcut_check(name,
                        'should be. Did you mean to use force?'.format(name)), pchanges
 
 
+def _makedirs(name,
+              user=None,
+              group=None,
+              dir_mode=None,
+              win_owner=None,
```
cachedout: I don't understand here what distinguishes `win_owner` from `user`. Why do these need to be distinct?

rallytime:
@twangboy does this really need to go in for 2017.7.6 and block that release? Or is this a nice-to-have and can go in for 2017.7.7?

salt/states/file.py
twangboy: Maybe if you have a state that is managing a file on both Windows and Linux... Honestly I don't know why. I'm happy to use user. If `win_owner` isn't passed, it will use `user`. But we've been using `win_owner` for a while. Just keeping with the paradigm I guess.

salt/states/file.py
twangboy: There are two instances where this function is called. It only specifies the name of the directory. All other values are hard values (no variables). Those are here:
- https://github.com/saltstack/salt/pull/47533/files#diff-3b447dc835943beb1326f98f684fe783L4706
  and here:
- https://github.com/saltstack/salt/pull/47533/files#diff-3b447dc835943beb1326f98f684fe783L4899
Since both calls to this function were passing the exact same things, I made them the defaults. This allows you to only pass the items you wish to check instead of having to pass all options every time.

salt/states/file.py
twangboy: I believe it was a design decision Dave and I came up with... but I don't remember the reasons.

cachedout: @UtahDave Do you know why?

UtahDave: I just went back and looked through the commit history a little bit. If `win_owner` isn't specified, then `user` is used.

My memory is a little hazy on the specifics since this was committed two years ago, but I believe it's because it was getting too difficult to manage Windows file system permissions by shoehorning it into Salt's  Linux file system permission management functions.

cachedout (approved):
I think I'm all right with this now but would like another core reviewer to have a look to double-check my work.

terminalmage (approved):

PR #47540, opened 2018-05-08
Fix corner case where runas user's HOME env value is incorrect

terminalmage:
In the test suite, two of the failing `pip.installed` tests were failing because when we shelled out to get the user's environment, the `HOME` environment variable's value was incorrectly showing the root user's home dir.

This change ensures that we use the correct value for the `HOME` environment variable when running a command as another user.

### DIFF
--- a/salt/modules/cmdmod.py
+++ b/salt/modules/cmdmod.py
@@ -478,10 +478,18 @@ def _run(cmd,
 
             env_runas = dict((sdecode(k), sdecode(v)) for k, v in six.iteritems(env_runas))
             env_runas.update(env)
+
             # Fix platforms like Solaris that don't set a USER env var in the
             # user's default environment as obtained above.
             if env_runas.get('USER') != runas:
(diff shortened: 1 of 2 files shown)


cachedout (approved):

terminalmage:
As a bit of context here, this test failures seemed to have been related to the fact that the nonprivileged user created by the `@with_system_user` decorator was created immediately before we shelled out in cmdmod.py. The `USER` env var was not present in the results from shelling out, and the `HOME` env var was set to the root user's homedir. However, if I stopped the test using remote pudb and then shelled out via the python shell in pudb, both `USER` and `HOME` were present in the results, and had the correct values. So, this appears to be a bit of a race condition.

Note: We already had code in cmdmod.py to fix `USER` if it wasn't present, since it was not showing up when using `runas` in Solaris, so the fact that `USER` was missing when this test shelled out to get the environment was fixed by existing code.

Issue #47529, opened 2018-05-07
salt-ssh doesn't work with lets than perfect dns servers


edgan:
### Description of Issue/Question
I am trying to setup some DNS servers with salt-ssh. Which are also configured as in my /etc/resolv.conf. dns-dns-01 is already configured, but dns-dns-03 is the not setup. The dns-dns-03 instance is first in the resolv.conf.

In ```salt/client/ssh/__init__.py``` on line 355 it tries to use the ip_to_host function to test if it can resolve the target name, not the host entry from the roster. When it fails the hostname is then set to None, and you get the traceback below.

In my opinion this line is useless, because I could set the target name in the roster to be whatever I want, and there shouldn't be an assumption that it will resolve via DNS. Also if I comment out that line, salt-ssh works without error.

Also the whole expand_target function is new, and I question if it does anything useful. It is new in 2018.3.0.

```
salt/client/ssh/__init__.py:
355: hostname = salt.utils.network.ip_to_host(hostname)
```
```
dns-dns-01 = 10.200.32.10
dns-dns-02 = 10.200.40.10
dns-dns-03 = 10.200.24.10
```

### Setup
roster:
```
dns-dns-01.inf.us-west-2.aws.acme.com:
  host: 10.200.32.10
  sudo: True
```
/etc/resolv.conf:
```
nameserver 10.200.24.10
nameserver 10.200.32.10
nameserver 100.200.40.10
```

### Steps to Reproduce Issue
```
salt-ssh -i dns-dns-01.inf.us-west-2.aws.acme.com
[ERROR   ] An un-handled exception was caught by salt's global exception handler:
TypeError: object of type 'NoneType' has no len()
Traceback (most recent call last):
  File "/usr/bin/salt-ssh", line 10, in <module>
    salt_ssh()
  File "/usr/lib/python2.7/site-packages/salt/scripts.py", line 425, in salt_ssh
    client.run()
  File "/usr/lib/python2.7/site-packages/salt/cli/ssh.py", line 23, in run
    ssh = salt.client.ssh.SSH(self.config)
  File "/usr/lib/python2.7/site-packages/salt/client/ssh/__init__.py", line 235, in __init__
    self.tgt_type)
  File "/usr/lib/python2.7/site-packages/salt/roster/__init__.py", line 105, in targets
    targets.update(self.rosters[f_str](tgt, tgt_type))
  File "/usr/lib/python2.7/site-packages/salt/roster/flat.py", line 50, in targets
[...]

PR #47541, opened 2018-05-08
switch skip-roster to update-roster

### LINKED ISSUE #47532: roster auto-add feature in salt-ssh-2018.3.0
I don't want this feature, and if it should stay it should be disabled by default.

```salt/client/ssh/__init__.py```:
```
                    roster_fp.write('# Automatically added by "{s_user}" at {s_time}\n{hostname}:\n    host: '
                                    '{hostname}\n    user: {user}'
                                    '\n    passwd: {passwd}\n'.format(s_user=getpass.getuser(),
                                                                      s_time=datetime.datetime.utcnow().isoformat(),
                                                                      hostname=self.opts.get('tgt', ''),
                                                                      user=self.opts.get('ssh_user', ''),
                                                                      passwd=self.opts.get('ssh_passwd', '')))
                log.info('The host {0} has been added to the roster {1}'.format(self.opts.get('tgt', ''),
                                                                                roster_file))
        else:
log.error('Unable to update roster {0}: access denied'.format(roster_file))
```

It causes things like in my roster:
```
# Automatically added by "edgan" at 2018-05-07T22:18:44.047773
None:
    host: None
    user: ubuntu
    passwd: 
```

Even if the None worked right, which is another bug, I don't want it editing my roster file.

### Setup
roster:
```
dns-dns-01.inf.us-west-2.aws.acme.com:
  host: 10.200.32.10
  sudo: True
```

### Steps to Reproduce Issue
```salt-ssh -i dns-dns-02.inf.us-west-2.aws.acme.com state.highstate```

### Versions Report
```
Salt Version:
           Salt: 2018.3.0
 
Dependency Versions:
           cffi: 1.11.2
       cherrypy: Not Installed
       dateutil: 2.6.1
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.10
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: 1.0.6
   msgpack-pure: Not Installed
 msgpack-python: 0.5.5
   mysql-python: Not Installed
      pycparser: 2.14
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: Not Installed
         Python: 2.7.14 (default, Mar 14 2018, 16:45:33)
   python-gnupg: Not Installed
         PyYAML: 3.12
          PyZMQ: 16.0.2
           RAET: Not Installed
          smmap: Not Installed
        timelib: Not Installed
        Tornado: 4.5.2
            ZMQ: 4.1.6
 
System Versions:
           dist: fedora 28 Twenty Eight
         locale: UTF-8
        machine: x86_64
        release: 4.16.6-302.fc28.x86_64
         system: Linux
        version: Fedora 28 Twenty Eight
```

### PR DESCRIPTION
**Pull request by gtmanfred**
### What does this PR do?
Writing to the /etc/salt/roster file by default was agreed by the SaltStack core engineers to be an undesirable action.  So I have switched the flag to be `--update-roster` and it defaults to False.

Fixes #47532

### Tests written?

No

### Commits signed with GPG?

Yes

### DIFF
--- a/conf/master
+++ b/conf/master
@@ -538,6 +538,10 @@
 # targeted with the normal -N argument to salt-ssh.
 #ssh_list_nodegroups: {}
 
+# salt-ssh has the ability to update the flat roster file if a minion is not
+# found in the roster.  Set this to True to enable it.
+#ssh_update_roster: False
+
 #####    Master Module Management    #####
 ##########################################
 # Manage how master side modules are loaded.
--- a/doc/topics/releases/2018.3.1.rst
+++ b/doc/topics/releases/2018.3.1.rst
@@ -13,3 +13,13 @@ used as part of a salt-minion process running on the master.  This will allow
 the minion to have pillars assigned to it, and will still allow the engine to
 create a LocalClient connection to the master ipc sockets to control
 environments.
+
+Changes to Automatically Updating the Roster File
+-------------------------------------------------
+
+In ``2018.3.0`` salt-ssh was configured to automatically update the flat roster
+file if a minion was not found for salt-ssh. This was decided to be
+undesireable as a default.  The ``--skip-roster`` flag has been removed and
+replaced  with ``--update-roster``, which will enable salt-ssh to add minions
+to the flat roster file.  This behavior can also be enabled by setting
+``ssh_update_roster: True`` in the master config file.
--- a/salt/client/ssh/__init__.py
+++ b/salt/client/ssh/__init__.py
@@ -410,7 +410,7 @@ class SSH(object):
                 'host': hostname,
                 'user': user,
             }
-            if not self.opts.get('ssh_skip_roster'):
+            if self.opts.get('ssh_update_roster'):
                 self._update_roster()
 
     def get_pubkey(self):
--- a/salt/utils/parsers.py
+++ b/salt/utils/parsers.py
@@ -3078,11 +3078,11 @@ class SaltSSHOptionParser(six.with_metaclass(OptionParserMeta,
             help='Run command via sudo.'
         )
         auth_group.add_option(
-            '--skip-roster',
-            dest='ssh_skip_roster',
+            '--update-roster',
+            dest='ssh_update_roster',
             default=False,
             action='store_true',
-            help='If hostname is not found in the roster, do not store the information'
+            help='If hostname is not found in the roster, store the information'
                  'into the default roster file (flat).'
         )
         self.add_option_group(auth_group)

### CODE REVIEW

**Review by cachedout — APPROVED** [review]

**Review by isbm — CHANGES_REQUESTED** [review]
I am honestly not in a favour disabling CLI sugar-things that can help attract newcomers. Otherwise this has no much deal here and we shooting ourselves in the foot with this.

The main idea here was that everyone who wants to call salt-ssh can call it directly first time (think "hello world" in PHP or RoR), so then automatically roster gets updated with the credentials and so the next call is smooth. Now you've simply removed this, turning it into again roster-editing configuration curve, which everyone of newcomer hates.

What would be actually improved here is so:

- `ssh_autoupdate_roster: False` into `/etc/salt/master` to disable it explicitly (because the user is already familiar with Salt)
- Add message "host has been added to the roster X" once it is automatically added.

This way we can keep happy newcomers who hates _any sort of configuration_ as well as keep happy advanced users.

**Comment by edgan** [comment]
@isbm This feature is broken out of the box. It assumes the title in the roster is a DNS resolvable FQDN. Which breaks existing assumptions that go back years, like title can be whatever, and the actual answer is host:. If host: didn't exist it would make sense, but it does. Breaking existing setups for new user sugar is bad.

Example:
```
title:
  host: 1.2.3.4
```

Real world example:
```
foo:
  host: 10.10.1.10
```

**Comment by isbm** [comment]
@edgan thanks for the update!

Well, actually it is [Salt ID, not a title](https://docs.saltstack.com/en/latest/topics/ssh/roster.html). While this can be indeed anything, why would I put a quote of Oscar Wilde (or anything else) there instead of a short hostname? That said, it supposed to work that if you have something like:

```
my_world_simplest_id_i_could_imagine_to_enter_here:
    foo.somedomain.biz: 123.234.123.234
```

...then nothing bad happens having another record here, once you access `salt-ssh root@foo`:
```
foo:
    foo.somedomain.biz: 123.234.123.234
```

But seems like it should have a better improvement here, especially on poor DNS. So I am all ears how we can do this even less editor dependent. Or, rephrasing, how to automatically add host data to the roster in a better way.

Second, let us stop mixing here two completely different topics: feature itself and its default state. While you think the feature is not working as it should, please use `--skip-roster` option, while we're together making things better.

How do we still keep "autoroster" alongside with the smooth work for advanced users? Having new syntax like `salt-ssh user@host` helps a lot for newcomers and also daily users who dislike to edit roster every time new docker or temporary VM pops up. The initial goal was to make Salt SSH act something like SSH's `known_hosts`, which you do not manually edit at all. And so I see no reason why SaltSSH would not do this automatically, yet keeping the possibility to add hosts data manually or import them from somewhere.

Summarising, I do agree on improvement and listening to the ideas and compromises. But I disagree with the attitude to provide workarounds by moving unresolved questions away by default and avoid solving them and just move on.

rallytime:
I agree with @gtmanfred and @edgan here. I don't think we should be writing to `/etc/salt/roster` by default, especially since this is a change in default behavior. Passing `--update-roster` to enable the behavior instead of remembering to pass `--skip-roster` for existing installs seems like a good compromise to me.

rallytime:
@isbm Yes, after looking more closely here, I see there are 2 issues (as you stated in your previous comment - I can understand this a little better now)

1. The `None` syntax bug. This has already has been fixed in #47191. (Original report in #47529)
2. What we do with the default behavior (or another solution?) still needs to be decided. Which is what I think @thatch45 can chime in about.

We can definitely keep this conversation going. I don't think we should rush to get something in for the 2018.3.1 release. We should consider this carefully as we go forward, as you said @isbm. Thanks for chiming in so we can make sure we have a good solution moving forward.

isbm:
@rallytime absolutely. I am all ears how to improve this as best for everyone, yet keeping it "sweet", i.e. automatic. So, to recap:

- @gtmanfred properly noted that it is not good when things are added somewhere quietly with no message. I completely agree with it. Solution: add a verbosity to what just had happened.

- @edgan dislikes that some garbage is added to the main roster. I actually dislike that too. :wink: Thanks for the bug discovery. @edgan , if you could please write here clear steps to reproduce it, I will fix it and write some tests for this. Thanks!

- Following SSH itself, I'd say adding stuff in background to some cached analogy of like `~/.ssh/known_hosts` would be a correct way. Yes, `known_hosts` serves for a bit different purposes, but still we can leverage this. Probably not to the main `/etc/salt/roster`, but nothing wrong to add that to the _local_ roster, mentioning it in the CLI and have it totally volatile. So next time Salt SSH would look into the main roster _and_ in the local roster. And if nothing found, only then ask for passwords and other options.

Any other suggestions are highly welcome.

thatch45 (approved):

thatch45:
In a nutshell, I find that it is appalling that we added something in the first place that would modify the flat roster, this change is a MUST.

isbm (approved):
@thatch45 OK, let's indeed leave main flat roster untouched.
@gtmanfred let's this go in first, then I will propose a follow-up PR on top of this that cleans up the cache story part.

Issue #18510, opened 2014-11-26
salt exit codes


Deshke:
is there a reason why test.\* returns False and exits with bash exit code 0 instead of 1 ? 

``` bash
~ # salt some-minion file.access  /var/run/reboot-required f; echo $?
<>:
    False
0

~ #  salt some-other-minion  file.access  /var/run/reboot-required f; echo $?
<>:
    True
0

```

inthecloud247:
+1

simple10:
+1

uvsmtid:
+1

I've just raised/tested few other similar issues particularly important for me. Searching for ["zero exit"](https://github.com/saltstack/salt/search?p=1&q=zero+exit&type=Issues) gives a list of issues related to lack of error indication for CLI.

## Should it be a critical issue? ##

"Critical" not in the sense that it does not allow using Salt. It is "critical" in the sense that it effectively makes automation unusable - the very reason why Salt is used in the first place.

For example, even if the output can be parsed, it is normally _never_ parsed on non-zero error code given by CLI. Otherwise the multitude of special cases to deal with "explodes into impracticality".

jfindlay:
relates to #2417.

ealphonse:
Hi,

+1 @uvsmtid for any jenkins jobs, and anything using state.orchestrate

uvsmtid:
I've just revisited this place due to related problems and remembered that I wrote a ad-hoc script to sniff failures from Salt output as part of Salt-based automation project. The script, however, is independent/single/standalone Python file - see links with details on [updated stackoverflow's answer][1].

[1]: http://stackoverflow.com/a/30000380/441652

jfindlay:
@uvsmtid, do you know if there are other github issues that document the issues you mention in that stackoverflow answer?  There are many annoying inconsistencies in salt's plugin systems, especially the modules and the states because those are the most used.  I have some ideas to automate plugin conformation with unit and possibly integration testing.  This is a big project that I've been wanting to do for a while and need to set aside time for it, but first I need to collect the scattered wisdom and user experience on the details for which no clear convention has been either implemented or documented.  I think most users would agree that a consistent and a standard plugin behavior is much more important than a favorite plugin behavior.

jfindlay:
[...]

PR #47544, opened 2018-05-08
Multi-command salt calls return highest exit

KaiSforza:
### What does this PR do?

Running a multiple command salt call will currently make salt think that
all of the minions returned with a failed state regardless of the return
values of the modules actually run. Right now it just returns the value
of `retcode` in the return dictionary if the return value is a
dictionary.

This checks if the `retcode` value is a dictionary itself, and then
returns the maximum return code that we recieved. This means that we
will continue to fail if the jobs did fail, but will return a zero if
all the jobs did so.

#18510 

### Previous Behavior

The salt summary would say that any multi-module salt call had failed, because it did not know how to parse returns with a dictionary `retcode`.

### New Behavior

Salt now handles dictionary `retcode` values by taking the highest return value from multi-module call and using that to determine if it failed.

### Tests written?

No

### Commits signed with GPG?

Yes


KaiSforza:
Updated the default value from the `.get()` to be a `dict` so it actually has `.values()`

KaiSforza:
Fixed code lint check failure

`18:16:33 salt/cli/salt.py:405: [C0325(superfluous-parens), ] Unnecessary parens after u'return' keyword`

isbm (approved):
@rallytime the code looks good dealing with the different exit code types and structures :wink: but it would be still good to run a tests over it.

gtmanfred (approved):

KaiSforza:
Okay it looks like the tests are failing outside of my changes.

terminalmage (approved):

Issue #47432, opened 2018-05-02
SysLogHandler stack traces when using incorrect log_file location in python3


Ch3LL:
### Description of Issue/Question
When using a syslog log_file location in your config that does not exist you will see a stacktrace telling you the file cannot be found:

```
--- Logging error ---                                                                                                                                                                   
Traceback (most recent call last):                                                                                                                                                      
  File "/usr/lib/python3.6/logging/handlers.py", line 936, in emit                                                                                                                      
    self.socket.send(msg)                                                                                                                                                               
OSError: [Errno 9] Bad file descriptor                                                                                                                                                  
                                                                                                                                                                                        
During handling of the above exception, another exception occurred:                                                                                                                     
                                                                                                                                                                                        
Traceback (most recent call last):                                                                                                                                                      
  File "/usr/lib/python3.6/logging/handlers.py", line 857, in _connect_unixsocket                                                                                                       
    self.socket.connect(address)                                                                                                                                                        
FileNotFoundError: [Errno 2] No such file or directory                                                                                                                                  
                                                                                                                                                                                        
During handling of the above exception, another exception occurred:                                                                                                                     
                                                                                                                                                                                        
Traceback (most recent call last):                                                                                                                                                      
  File "/usr/lib/python3.6/logging/handlers.py", line 939, in emit                                                                                                                      
    self._connect_unixsocket(self.address)                                                                                                                                              
  File "/usr/lib/python3.6/logging/handlers.py", line 868, in _connect_unixsocket                                                                                                       
    self.socket.connect(address)                                                                                                                                                        
FileNotFoundError: [Errno 2] No such file or directory
```



### Setup
1. syslog is installed and running
2. minion config:
```
log_file: file:///dev/log2/LOG_LOCAL3
```
(make sure its a file that does not exist)
3. `salt-call --local cmd.run "echo foo" -lwarning`

and you will see the stacktrace. 

This works just fine on python2 and exits out with an error:

```
➜  salt git:(e745639763) sudo salt-call --local cmd.run "echo foo" -lwarning
[ERROR   ] Failed to setup the Syslog logging handler: [Errno 2] No such file or directory
➜  salt git:(e745639763) 
```

python3:

```
salt-call --local cmd.run "echo foo" -lwarning
--- Logging error ---
Traceback (most recent call last):
  File "/usr/lib/python3.6/logging/handlers.py", line 936, in emit
    self.socket.send(msg)
OSError: [Errno 9] Bad file descriptor

[...]

PR #47548, opened 2018-05-08
Catch Sysloghandler errors when log file does not exist

Ch3LL:
https://github.com/saltstack/salt/issues/47432
https://github.com/saltstack/salt-jenkins/issues/936

### Previous Behavior
In python3 this PR was added https://github.com/python/cpython/pull/663  which makes sure it just passes when there is an OSERROR. The problem is when running salt with a `log_file` set to `file://doesnotexist` or syslog service hasn't started yet it prints the entire traceback and is hard to digest.

### New Behavior
Prints out a new warning: `[WARNING ] The log_file does not exist. Logging not setup correctly or syslog service not started.`

### Tests written?

Yes

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/log/handlers/__init__.py
+++ b/salt/log/handlers/__init__.py
@@ -17,7 +17,6 @@ import logging.handlers
 
 # Import salt libs
 from salt.log.mixins import NewStyleClassMixIn, ExcInfoOnLogLevelFormatMixIn
-from salt.ext import six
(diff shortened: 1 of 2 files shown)


rallytime:
re-run py3

rallytime (approved):
Really nice work tracking this one down. 👍

s0undt3ch (approved):

Issue #46981, opened 2018-04-09
win_snmp module bypasses group policy settings


doesitblend:
### Description of Issue/Question
When retrieving values from the SNMP configuration `win_snmp` always retrieves values from the registry which is not correct. Windows allows configuring the daemon through the service agent and through group policy, in my case local group policy. 

The problem comes when you configure the agent as above and then query the settings with `salt '<target minion>' win_snmp.get_community_names`. Even though the agent settings have been overwritten by the group policy Salt still returns the original configured community name pulling it directly from a registry setting. 

### Setup

1. Enable the SNMP service in Windows Server manager [Instructions here](https://support.microsoft.com/en-us/help/324263/how-to-configure-the-simple-network-management-protocol-snmp-service-i)
2. Follow above instructions to configure a community string
3. Configure a new community string through group policy 
4. Check the agent settings through the windows service manager
  * These should now read the new community string set through group policy
  * May require a `gpupdate /force` from powershell to update immediately
5. Query the community name with the command provided above
  * This will return the originally configured community string set manually through the agent configuration

### Versions Report
```
    Salt Version:
               Salt: 2017.7.2
     
    Dependency Versions:
               cffi: 1.10.0
           cherrypy: 10.2.1
           dateutil: 2.6.0
          docker-py: Not Installed
              gitdb: 2.0.3
          gitpython: 2.1.3
              ioflo: Not Installed
             Jinja2: 2.9.6
            libgit2: Not Installed
            libnacl: Not Installed
           M2Crypto: Not Installed
               Mako: 1.0.6
       msgpack-pure: Not Installed
     msgpack-python: 0.4.8
       mysql-python: Not Installed
          pycparser: 2.17
           pycrypto: 2.6.1
       pycryptodome: Not Installed
             pygit2: Not Installed
             Python: 2.7.13 (v2.7.13:a06454b1afa1, Dec 17 2016, 20:53:40) [MSC v.1500 64 bit (AMD64)]
       python-gnupg: 0.4.0
             PyYAML: 3.11
              PyZMQ: 16.0.2
               RAET: Not Installed
              smmap: 2.0.3
            timelib: 0.2.4
            Tornado: 4.5.1
                ZMQ: 4.1.6
     
    System Versions:
               dist:   
             locale: cp1252
            machine: AMD64
            release: 2012ServerR2
             system: Windows
            version: 2012ServerR2 6.3.9600  Multiprocessor Free
```

doesitblend:
[...]

PR #47552, opened 2018-05-09
Show GPO settings, raise error if trying to set gpo managed settings

twangboy:
### What does this PR do?
Fixes issues with conflicts between GPO managed SNMP settings and those managed by the SNMP gui. In the GUI, the SNMP settings are greyed out when they are being managed by Group Policy.

https://github.com/saltstack/salt/issues/46981

### Previous Behavior
`get_community_names` would always return the SNMP gui settings
`set_community_names` would successfully set SNMP settings without error but would not apply if Group Policy was also managing those settings

### New Behavior
`get_community_names` will now return the Group Policy settings if they are applied. Otherwise, they will return the normal SNMP settings.
```
GPO Managed
----------
test_community:
    Managed by GPO

SNMP Managed
----------
test_community:
    Read Only
```

`set_community_names` now raises a `CommandExecutionError` if the settings are being managed by Group Policy.

### Tests written?
Yes

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/modules/win_snmp.py
+++ b/salt/modules/win_snmp.py
@@ -9,16 +9,20 @@ from __future__ import absolute_import
 import logging
 
 # Import salt libs
-from salt.exceptions import SaltInvocationError
+from salt.exceptions import SaltInvocationError, CommandExecutionError
 import salt.utils
 
 # Import 3rd party libs
 from salt.ext import six
 
 _HKEY = 'HKLM'
+
 _SNMP_KEY = r'SYSTEM\CurrentControlSet\Services\SNMP\Parameters'
 _AGENT_KEY = r'{0}\RFC1156Agent'.format(_SNMP_KEY)
 _COMMUNITIES_KEY = r'{0}\ValidCommunities'.format(_SNMP_KEY)
+
+_SNMP_GPO_KEY = r'SOFTWARE\Policies\SNMP\Parameters'
+_COMMUNITIES_GPO_KEY = r'{0}\ValidCommunities'.format(_SNMP_GPO_KEY)
 
 _PERMISSION_TYPES = {'None': 1,
                      'Notify': 2,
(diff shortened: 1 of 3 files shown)


rallytime:
@twangboy This is causing the following 2 tests to fail:

- unit.modules.test_win_snmp.WinSnmpTestCase.test_get_community_names
- unit.modules.test_win_snmp.WinSnmpTestCase.test_set_community_names

Can you take a look?

twangboy:
@rallytime Fixed the failing tests... added new tests to test for instances where the settings are found in gpo

dwoz (approved):

rallytime:
@lomeroe and @lorengordon Can you review this PR? Since you were involved in the original discussion in #46981?

salt/modules/win_snmp.py
```diff
@@ -340,6 +412,11 @@ def set_community_names(communities):
     '''
     values = dict()
 
+    if __salt__['reg.key_exists'](_HKEY, _COMMUNITIES_GPO_KEY):
+        _LOG.debug('Communities on this system are managed by Group Policy')
+        raise CommandExecutionError(
```
lorengordon: I agree it is better to `raise` here, but I think it is at least worth a release note that this is a change in behavior (if not a deprecation path). Someone may be using this function on a system now where Group Policy happens to be managing the communities  Salt would succeed previously (if a little pointlessly), and this change will cause a bit of breakage when they update.

twangboy: The state wouldn't fail, as it checks for the CommandExecutionError. Only a Command line run or a salt api run on the execution module would fail.

twangboy:
@lomeroe @lorengordon Could you look at this again.

salt/modules/win_snmp.py
lorengordon: Right... The point remains, yes? The behavior is changing in a way that breaks prior usage of this function. Certainly salt states are designed to handle that, but salt states are not the sole user of execution modules.

lorengordon (approved):

salt/modules/win_snmp.py
twangboy: Right. I added this to the release notes.

PR #47557, opened 2018-05-09
fix broken rabbitmq list policies in rabbitmq version 3.7

L4rS6:
### What does this PR do?
This PR fixes the `list_policies` module output in rabbitmq version 3.7 and later. In rabbitmqctl version 3.7 the order of the `list_policies` output has changed:

**Version 3.6**
`/	ha-all	all	^(?!amq[.]).*	{"ha-mode":"all"}	0`

**Version 3.7**
`/	ha-all	^(?!amq[.]).*	all	{"ha-mode":"all"}	0`

In fact that we need the rabbitmq version I used salt's `pkg.version` module to detect its version. Due to different package names on different OS (freebsd, ubuntu) this check has also been implemented.

### Previous Behavior
On minions with rabbitmq version 3.7 the order of `apply_to` and `pattern` was wrong:
```
    ----------
    /:
        ----------
        ha-all:
            ----------
            apply_to:
                ^(?!amq[.]).*
            definition:
                {"ha-mode":"all"}
            pattern:
                all
            priority:
                0
```

### New Behavior
Order has been fixed for `apply_to` and `pattern`:
```
    ----------
    /:
        ----------
        ha-all:
            ----------
            apply_to:
                all
            definition:
                {"ha-mode":"all"}
            pattern:
                ^(?!amq[.]).*
            priority:
                0
```
### Tests written?
No

### Commits signed with GPG?
No


salt/modules/rabbitmq.py
```diff
@@ -828,24 +828,46 @@ def list_policies(vhost="/", runas=None):
         python_shell=False)
     _check_response(res)
     output = res['stdout']
+
+    if __salt__['pkg.version']('rabbitmq-server'):
```
rallytime: Would using something like `salt.utils.versions.LooseVersion` be beneficial here?

rallytime (changes requested):
Hi @L4rS6 - Thanks for submitting this change, and apologies for not responding sooner. I have one question.

In addition, the following tests are failing:
- unit.modules.test_rabbitmq.RabbitmqTestCase.test_list_policies
- unit.modules.test_rabbitmq.RabbitmqTestCase.test_policy_exists

Can you take a look at that?

salt/modules/rabbitmq.py
L4rS6: @rallytime Yes, I took a look at that. I'm thinking about to split up the rabbitmq.py module to a freebsd_rabbitmq.py and a default rabbitmq.py module, because of the different package names. If I do it like this I can set the package name in the freebsd module and don't have to make an ugly if-else construct. At the moment I'm not sure if I should copy all functions into the freebsd_rabbitmq.py module. If I do this we have a lot of duplicate code (freebsd_rabbitmq.py and rabbitmq.py). Do you have an idea how to solve this nicely? 

You're right, I should use LooseVersion to compare the versions.

rallytime: The if/else is fine, but to answer your question, we use the files in the `utils` directory to hold code that is used across multiple modules/states.

rallytime (approved):
Welcome @L4rS6! Thank you for fixing this and for writing these additional tests. ++

L4rS6:
@rallytime Will this change also be merged into the master branch? I just want to be sure that those changes will be in future releases as well.

rallytime:
@L4rS6 Yep! It will be merged forward to `develop`. We've got some doc on that [here](https://docs.saltstack.com/en/latest/topics/development/contributing.html#salt-s-branch-topology), if you're interested.

PR #47513, opened 2018-05-07
Fix #46546 eos napalm's grains issue 

frankiexyz:
### What does this PR do?
For eos minion (not proxy minon), grains are not able to retrieve due to connection is not opened.
By changing the device cache update to function get_device, it will setup a connection first.
### What issues does this PR fix or reference?
#46546
#47203
### Previous Behavior
➜   docker exec -t d03bd4c21b37 salt '*' grains.get vendor
supervisor1:
    None
### New Behavior
➜  docker exec -t d03bd4c21b37 salt '*' grains.get hostname
test:
    test
➜   docker exec -t d03bd4c21b37 salt '*' grains.get vendor
test:
    Arista
After change hostname to test2
➜  frankie docker exec -t d03bd4c21b37 salt '*' saltutil.refresh_grains                                                                                                       
test:
    True
➜   docker exec -t d03bd4c21b37 salt '*' grains.get hostname
test:
    test2

### Tests written?

No

### Commits signed with GPG?

No

Please review [Salt's Contributing Guide](https://docs.saltstack.com/en/latest/topics/development/contributing.html) for best practices.

See GitHub's [page on GPG signing](https://help.github.com/articles/signing-commits-using-gpg/) for more information about signing commits with GPG.

PR #47562, opened 2018-05-09
Back-port #47513 to 2017.7

rallytime:
Back-port #47513 to 2017.7


mirceaulinic (approved):
Thanks @rallytime. To help you with these backports: would it help you if I backport them myself, or would you rather prefer to have everything on your radar?

rallytime:
@mirceaulinic You can back-port them if you like. It's not hard to do them when they back-port cleanly. As long as you link them back to the original PR, then that's great if you want to help out. Either was is good. :)

PR #47567, opened 2018-05-09
Clean up import blocks in zabbix module

rallytime:
The salt imports were moved to a try/except block and they shouldn't be.

As a result of cleaning up that import block, this PR does the following:
- moves the salt imports out of the try/except
- updates the __virtual__() function to no longer rely on the status of the salt imports
- updates the old utils paths with the new utils paths

Refs #47398

Additionally, #47398 added autodoc files for several zabbix states, but those need references in the `index.rst` file as well. The second commit adds those references.

### DIFF
--- a/doc/ref/states/all/index.rst
+++ b/doc/ref/states/all/index.rst
@@ -314,11 +314,14 @@ state modules
     winrepo
     x509
     xmpp
+    zabbix_action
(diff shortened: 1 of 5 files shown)


cachedout (approved):

rallytime:
re-run py

PR #47568, opened 2018-05-09
salt.serializers.yaml/yamlex: remove invalid multi_constructor

terminalmage:
The first argument to add_multi_constructor must be a YAML tag. This is because when PyYAML looks for a constructor to match the tag (in order to get the function to use for deserialization) it matches the tag using `.startswith()`. Thus, when this is attempted using a constructor with a tag value of `None`, an error is raised.

The ordering of the list of multi_constructors in the Loader object appears to differ from platform to platform, which explains why this only caused the unit tests to fail on one or two platforms. If a tag match is found, then PyYAML stops iterating through the list of multi_constructors, so this invalid one has been able to stay in Salt without causing any noticeable trouble until now.

Additionally, this PR adds some exception logging to aid in troubleshooting, since we re-raise our own exception classes using the caught exception message, which means we lose the context of what actually raised the original exception.

Refs: https://github.com/saltstack/salt-jenkins/issues/971

### DIFF
--- a/salt/serializers/yaml.py
+++ b/salt/serializers/yaml.py
@@ -11,6 +11,7 @@
 
 from __future__ import absolute_import, print_function, unicode_literals
 import datetime
+import logging
 
 import yaml
 from yaml.constructor import ConstructorError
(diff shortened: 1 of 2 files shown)


rallytime (approved):

cachedout (approved):
Sheesh. Amazing find.

cachedout:
Hrm .That failure of a core grains test is odd.

rallytime:
@cachedout That grains test is failing on the branch and should be fixed with #47569.

PR #47569, opened 2018-05-09
Update salt.utils.path mock in virtual core test

Ch3LL:
### What does this PR do?
change mock in `test_core_virtual` from `salt.utils.which` to `salt.utils.path.which` as this was updated in 2018.3.

https://github.com/saltstack/salt-jenkins/issues/969

### DIFF
--- a/tests/unit/grains/test_core.py
+++ b/tests/unit/grains/test_core.py
@@ -866,7 +866,7 @@ SwapTotal:       4789244 kB'''
         virt = 'kvm'
         with patch.object(salt.utils, 'is_windows',
                           MagicMock(return_value=False)):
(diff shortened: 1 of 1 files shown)


rallytime (approved):

tests/unit/grains/test_core.py
```diff
@@ -866,7 +866,7 @@ def test_core_virtual(self):
         virt = 'kvm'
         with patch.object(salt.utils, 'is_windows',
```
rallytime: This line should be `salt.utils.platform` as well.

rallytime (changes requested):
There's another path update needed for this test. Can you update that as well?

rallytime (approved):

Issue #47546, opened 2018-05-08
archive.extracted fails when archive includes files with non-ascii names with 2018.3.0


ender8282:
### Description of Issue/Question
I have a .tgz archive which fails to extract.  I believe that it is due to a non-ascii character in the name of one of the files.  I believe that this is a regression because the state functioned properly when using 2017.7.4 but now fails with 2018.3.0.

In searching for help with this issue I found a few issues that appeared related but none of those issues specifically dealt with archive.extracted.

### Setup
```
ibm-java:
  archive.extracted:
    - name: /opt/java
    - source: salt://roles/csf_app_host/ibm-jvm.tgz
    - archive_format: tar
    - if_missing: /opt/java/ibm-java-80
```

I believe the problematic file is:
`ibm-java-80/_uninstall/Logs/IBM®_64-bit_SDK_for_Linux®,_v8.0_Install_01_06_2016_09_55_41.log`

The error is:
```
----------
          ID: ibm-java
    Function: archive.extracted
        Name: /opt/java
      Result: False
     Comment: An exception occurred in this state: Traceback (most recent call last):
                File "/usr/lib/python2.7/site-packages/salt/state.py", line 1878, in call
                  **cdata['kwargs'])
                File "/usr/lib/python2.7/site-packages/salt/loader.py", line 1823, in wrapper
                  return f(*args, **kwargs)
                File "/usr/lib/python2.7/site-packages/salt/states/archive.py", line 1013, in extracted
                  verbose=True)
                File "/usr/lib/python2.7/site-packages/salt/modules/archive.py", line 419, in list_
                  if x.count('/') == 0]
              UnicodeDecodeError: 'ascii' codec can't decode byte 0xc2 in position 31: ordinal not in range(128)
     Started: 11:45:42.568472
    Duration: 3533.573 ms
     Changes:

```

### Versions Report
```
Salt Version:
           Salt: 2018.3.0

Dependency Versions:
           cffi: Not Installed
       cherrypy: Not Installed
       dateutil: Not Installed
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.8
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.5.1
   mysql-python: Not Installed
      pycparser: Not Installed
       pycrypto: 2.6.1
[...]

PR #47572, opened 2018-05-09
[2018.3] archive module, filenames with Unicode characters

garethgreenaway:
### What does this PR do?
Accounting for when files in an archive contain non-ascii characters
Updating integration/modules/test_archive to include filenames with unicode characters.

#47546 

### Tests written?
Yes

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/modules/archive.py
+++ b/salt/modules/archive.py
@@ -413,12 +413,20 @@ def list_(name,
             ret = {'dirs': sorted(dirs),
                    'files': sorted(files),
                    'links': sorted(links)}
-            ret['top_level_dirs'] = [x for x in ret['dirs']
-                                     if x.count('/') == 1]
-            ret['top_level_files'] = [x for x in ret['files']
-                                      if x.count('/') == 0]
-            ret['top_level_links'] = [x for x in ret['links']
-                                      if x.count('/') == 0]
+            if six.PY2:
+                ret['top_level_dirs'] = [x for x in ret['dirs']
+                                         if x.decode('utf-8').count('/') == 1]
+                ret['top_level_files'] = [x for x in ret['files']
+                                          if x.decode('utf-8').count('/') == 0]
+                ret['top_level_links'] = [x for x in ret['links']
+                                          if x.decode('utf-8').count('/') == 0]
+            else:
+                ret['top_level_dirs'] = [x for x in ret['dirs']
+                                         if x.count('/') == 1]
+                ret['top_level_files'] = [x for x in ret['files']
+                                          if x.count('/') == 0]
+                ret['top_level_links'] = [x for x in ret['links']
+                                          if x.count('/') == 0]
         else:
             ret = sorted(dirs + files + links)
         return ret
--- a/salt/modules/archive.py
+++ b/salt/modules/archive.py
@@ -410,23 +410,15 @@ def list_(name,
                 item.sort()
 
         if verbose:
-            ret = {'dirs': sorted(dirs),
-                   'files': sorted(files),
-                   'links': sorted(links)}
-            if six.PY2:
-                ret['top_level_dirs'] = [x for x in ret['dirs']
-                                         if x.decode('utf-8').count('/') == 1]
-                ret['top_level_files'] = [x for x in ret['files']
-                                          if x.decode('utf-8').count('/') == 0]
-                ret['top_level_links'] = [x for x in ret['links']
-                                          if x.decode('utf-8').count('/') == 0]
-            else:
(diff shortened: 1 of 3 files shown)


rallytime:
@garethgreenaway The archive state integration tests didn't like this change very much: https://jenkins.saltstack.com/job/PR/job/salt-pr-linode-cent7-py3/4803/

Can you take a look?

rallytime:
re-run lint

rallytime (approved):

salt/modules/archive.py
```diff
@@ -413,12 +413,20 @@ def _unsupported_format(archive_format):
...
+                                          if x.decode('utf-8').count('/') == 0]
+                ret['top_level_links'] = [x for x in ret['links']
+                                          if x.decode('utf-8').count('/') == 0]
+            else:
+                ret['top_level_dirs'] = [x for x in ret['dirs']
+                                         if x.count('/') == 1]
+                ret['top_level_files'] = [x for x in ret['files']
+                                          if x.count('/') == 0]
+                ret['top_level_links'] = [x for x in ret['links']
+                                          if x.count('/') == 0]
```
terminalmage: It looks like the `getmembers` method within `tarfile` is returning paths as `str` types on Python 2. If this is the case, then the specific handling for Python 2 and 3 won't be necessary if you use `salt.utils.data.decode_list()` on each list when initializing `ret`:

```python
ret = {'dirs': sorted(salt.utils.data.decode_list(dirs)),
       'files': sorted(salt.utils.data.decode_list(files)),
       'links': sorted(salt.utils.data.decode_list(links))}
```

This will ensure that all paths are unicode types, and the list comprehensions which generate the `top_level_*` keys will therefore not cause a `UnicodeDecodeError`.

`salt.utils.data.decode()` would also work here, but it is a general-purpose function that does type checking and would end up simply calling `salt.utils.data.decode_list()` under-the-hood. Since we know that `dirs`, `files`, and `links` are all lists, using `salt.utils.data.decode_list()` would be marginally faster here.

terminalmage (changes requested):
There are a few places that this fix can be simplified.

:+1: for the additional test coverage.

terminalmage (approved):

garethgreenaway:
@junovitch No need for a new issue, we can just re-open the existing issue and reference the new PR in there.  I believe I've got this one figured out.  The issue is over in the archive module, here is a diff if you wouldn't mind testing it for me:

```
diff --git a/salt/modules/archive.py b/salt/modules/archive.py
index dc49fd252f..76cd3eeb97 100644
--- a/salt/modules/archive.py
+++ b/salt/modules/archive.py
@@ -186,12 +186,13 @@ def list_(name,
                 else {'fileobj': cached.stdout, 'mode': 'r|'}
             with contextlib.closing(tarfile.open(**open_kwargs)) as tar_archive:
                 for member in tar_archive.getmembers():
+                    _member = salt.utils.data.decode(member.name)
                     if member.issym():
-                        links.append(member.name)
+                        links.append(_member)
                     elif member.isdir():
-                        dirs.append(member.name + '/')
+                        dirs.append(_member + '/')
                     else:
-                        files.append(member.name)
+                        files.append(_member)
             return dirs, files, links
 
         except tarfile.ReadError:
```

If everything looks good then I'll prep a new PR.

junovitch:
@garethgreenaway thanks for the quick work.  This is an improvement; the tar file now doesn't hit that uncaught exception.  Both the tar and zip in the referenced sls snippet above fail at the same point:

```
  File "/usr/local/lib/python2.7/site-packages/salt/states/archive.py", line 1105, in extracted                   
    path_mode = os.lstat(full_path.rstrip(os.sep)).st_mode                                                        
UnicodeEncodeError: 'ascii' codec can't encode characters in position 120-130: ordinal not in range(128)  
```

What do you think?  Let me know and I will sanity check things.  Certainly tag me in the new PR when we get there as well.  Thanks!

junovitch:
On my device network at work the above patch fixed my issue but I have also pulled a few more unicode related patches back from the 2018.3 branch on to my 2018.3.0 deployment.

garethgreenaway:
@junovitch I ran your test state with the tar.gz file in question and it worked as expected.  This was using the HEAD of 2018.3 so it installed all the other Unicode related fixes.  I'm going to add some additional tests and submit a PR with the above fix.

junovitch:
Great! I just didn't have $LANG set on my test jail at home and it was set at work. Looks good now! Thanks for the quick response and getting these fixed in the next release.

Issue #47274, opened 2018-04-24
Hosts File - Windows - "Cannot perform string replacements on a binary file"


Silvenga:
### Description of Issue/Question

When attempting to modify the Window host file as a work-around for bug #47215, `file.blockreplace` fails with the following error:

```
          ID: Ensure host entries exists
    Function: file.blockreplace
        Name: C:\Windows\System32\drivers\etc\hosts
      Result: False
     Comment: An exception occurred in this state: Traceback (most recent call last):
                File "C:\salt\bin\lib\site-packages\salt\state.py", line 1878, in call
                  **cdata['kwargs'])
                File "C:\salt\bin\lib\site-packages\salt\loader.py", line 1823, in wrapper
                  return f(*args, **kwargs)
                File "C:\salt\bin\lib\site-packages\salt\states\file.py", line 4378, in blockreplace
                  show_changes=show_changes
                File "C:\salt\bin\lib\site-packages\salt\modules\file.py", line 2518, in blockreplace
                  .format(path)
              SaltInvocationError: Cannot perform string replacements on a binary file: C:\Windows\System32\drivers\etc\hosts
     Started: 08:22:54.611000
    Duration: 16.0 ms
     Changes:
```

### Setup

```
Ensure host entries exists:
  file.blockreplace:
    - name: C:\Windows\System32\drivers\etc\hosts
    - content: "127.0.0.1 example.com"
    - marker_start: "# Start host entries - managed by Salt."
    - marker_end: "# End host entries."
    - append_if_not_found: True
```

### Steps to Reproduce Issue

Run the above state on a Windows minion.

### Versions Report

```
Salt Version:
           Salt: 2018.3.0

Dependency Versions:
           cffi: 1.10.0
       cherrypy: 10.2.1
       dateutil: 2.6.1
      docker-py: Not Installed
          gitdb: 2.0.3
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.10
        libgit2: Not Installed
        libnacl: 1.6.1
       M2Crypto: Not Installed
           Mako: 1.0.7
   msgpack-pure: Not Installed
 msgpack-python: 0.4.8
   mysql-python: Not Installed
      pycparser: 2.18
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: Not Installed
         Python: 2.7.14 (v2.7.14:84471935ed, Sep 16 2017, 20:25:58) [MSC v.1500 64 bit (AMD64)]
[...]

PR #47581, opened 2018-05-09
Add get_encoding salt util

twangboy:
### What does this PR do?
Created the `get_encoding` function `salt.util.files` to try to detect the encoding of the file.
Uses `io.open` to open the file with the proper encoding.
Writes the lines to the new file using the proper encoding.

https://github.com/saltstack/salt/issues/47274

### Previous Behavior
Some files were being incorrectly detected as binary files by the `is_binary` function (utf-16 encoded files for example). Additionally, once they were properly detected as not binary, the read line function would incorreclty parse the lines ignoring a trailing null byte (`\x00`) causing the file to become corrupted.

### New Behavior
Detects the encoding of the file and opens the file with the detected encoding so that the read line function works correctly.

### Tests written?
No

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/utils/files.py
+++ b/salt/utils/files.py
@@ -616,6 +616,31 @@ def safe_filepath(file_path_name, dir_sep=None):
     return path
 
 
+def is_encoding(path, encoding="utf-8"):
+    '''
+    Detect if the file can be successfully decoded by the passed encoding
+
+    Args:
+
+        fp_ (pointer): A pointer to the file to check
+        encoding (str): The encoding to test
+
+    Return:
+        bool: True if successful, otherwise False
+    '''
+    if not os.path.isfile(path):
+        return False
+    try:
+        with fopen(path, 'rb') as fp_:
+            try:
+                data = fp_.read(2048)
+                data.decode(encoding)
+            except UnicodeDecodeError:
+                return True
+    except os.error:
+        return False
+
+
 @jinja_filter('is_text_file')
 def is_text(fp_, blocksize=512):
     '''
--- a/salt/utils/files.py
+++ b/salt/utils/files.py
@@ -826,7 +826,7 @@ def get_encoding(path):
         ]
         for _encoding, bom in boms:
             if _data.startswith(bom):
-                log.debug('Found BOM for {0}'.format(_encoding))
+                log.debug('Found BOM for %s', _encoding)
                 return _encoding
         return False
 
--- a/salt/utils/files.py
+++ b/salt/utils/files.py
@@ -6,6 +6,7 @@ Functions for working with files
 from __future__ import absolute_import, unicode_literals, print_function
 
 # Import Python libs
+import codecs
 import contextlib
 import errno
 import logging
@@ -616,31 +617,6 @@ def safe_filepath(file_path_name, dir_sep=None):
     return path
 
 
-def is_encoding(path, encoding="utf-8"):
-    '''
-    Detect if the file can be successfully decoded by the passed encoding
-
-    Args:
-
-        fp_ (pointer): A pointer to the file to check
-        encoding (str): The encoding to test
-
-    Return:
-        bool: True if successful, otherwise False
-    '''
-    if not os.path.isfile(path):
-        return False
-    try:
-        with fopen(path, 'rb') as fp_:
-            try:
-                data = fp_.read(2048)
-                data.decode(encoding)
-            except UnicodeDecodeError:
-                return True
-    except os.error:
-        return False
-
-
 @jinja_filter('is_text_file')
 def is_text(fp_, blocksize=512):
     '''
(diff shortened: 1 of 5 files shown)


damon-atkins:
I suggest a function which returns the file information.
e.g.
{
open_type: <can be used by open>
type: <text string>
eol: <end of line type>
size: <int>
}

Or salt/utils/files contains a "open" class which manages files  open, backups, read, write, eol char, etc. It just takes care of it all.

salt/utils/files.py
```diff
@@ -616,6 +616,31 @@ def safe_filepath(file_path_name, dir_sep=None):
     return path
 
 
+def is_encoding(path, encoding="utf-8"):
+    '''
+    Detect if the file can be successfully decoded by the passed encoding
+
+    Args:
+
+        fp_ (pointer): A pointer to the file to check
```
terminalmage: Shouldn't the arg name be `path` here? Also, there is no such type in Python as `pointer`, this should be `str` instead.

salt/utils/files.py
```diff
@@ -616,6 +616,31 @@ def safe_filepath(file_path_name, dir_sep=None):
...
+        encoding (str): The encoding to test
+
+    Return:
+        bool: True if successful, otherwise False
+    '''
+    if not os.path.isfile(path):
+        return False
```
terminalmage: I feel like we should probably be attempting to read from the file and simply raising the exception here, but it looks like `is_binary`, et al have already established the pattern that we just return `False` in these cases, so I guess this is OK.

salt/utils/files.py
```diff
@@ -616,6 +616,31 @@ def safe_filepath(file_path_name, dir_sep=None):
...
+    try:
+        with fopen(path, 'rb') as fp_:
+            try:
+                data = fp_.read(2048)
+                data.decode(encoding)
+            except UnicodeDecodeError:
+                return True
```
terminalmage: Shouldn't `False` be returned here? The decode is failing, so that would suggest that the file is not in the specified encoding.

terminalmage (changes requested):

damon-atkins:
Can not see any issue with the new function.

However the reason I suggest a function to return several bits of information. Is so the file examined once. For example to determine EOL the last part of the file is best place to read (e.g. last two characters of the file). If the file is under 2k then last part of the file has already been read.
```
{
open_type: <can be used by open>
type: <text string>
eol: <end of line type>
size: <int>
etc
}
```

twangboy:
@damon-atkins Could you explain what you mean by open_type and type?

twangboy:
The following tests are failing:
integration.states.test_file.BlockreplaceTest.test_append_auto_line_separator
integration.states.test_file.BlockreplaceTest.test_prepend_auto_line_separator
unit.utils.test_schedule.ScheduleTestCase.test_eval_schedule_cron_splay
unit.utils.test_schedule.ScheduleTestCase.test_eval_schedule_cron
unit.modules.test_file.FileBlockReplaceTestCase.test_backup
unit.modules.test_file.FileBlockReplaceTestCase.test_dry_run
unit.modules.test_file.FileBlockReplaceTestCase.test_no_modifications
unit.modules.test_file.FileBlockReplaceTestCase.test_replace_append
unit.modules.test_file.FileBlockReplaceTestCase.test_replace_append_newline_at_eof
unit.modules.test_file.FileBlockReplaceTestCase.test_replace_multiline
unit.modules.test_file.FileBlockReplaceTestCase.test_replace_partial_marked_lines
unit.modules.test_file.FileBlockReplaceTestCase.test_replace_prepend
unit.modules.test_file.FileBlockReplaceTestCase.test_show_changes
unit.modules.test_file.FileBlockReplaceTestCase.test_unfinished_block_exception

damon-atkins:
> Could you explain what you mean by open_type and type?   

Basically open_type information on how python open should open the file, or params for python open to use.  e.g. open in text or binary mode or some thing else.   type would be binary or text or text-bom or something along these lines.


The orginal orginal idea I had was to have a utils.filemgt (manage files) class.

- opens the file and determines determines its characteristics, and cache this info. e.g. EOL  etc. So it ensures the file structure is not changes i.e. of it has a BOM head the BOM head is kept.
- manages disk full issues
- manages not replacing the file if it has not changed
- keeps a record of all files open for change but not changed
- keeps a record of all files changed.
- backup files if they are changed into backup location.
- etc.
i.e. try to pull most of the logic out of all the modules and states.

twangboy:
I think that's a cool idea, but it's a major change and doesn't relate to this issue directly.

twangboy:
Now the tests are just failing on Py3... I'll start digging

salt/utils/files.py
```diff
@@ -777,3 +778,102 @@ def backup_minion(path, bkroot):
...
+            ('UTF-8', salt.utils.stringutils.to_bytes(codecs.BOM_UTF8)),
+            ('UTF-7', salt.utils.stringutils.to_bytes('\x2b\x2f\x76\x38\x2D')),
+            ('UTF-7', salt.utils.stringutils.to_bytes('\x2b\x2f\x76\x38')),
+            ('UTF-7', salt.utils.stringutils.to_bytes('\x2b\x2f\x76\x39')),
+            ('UTF-7', salt.utils.stringutils.to_bytes('\x2b\x2f\x76\x2b')),
+            ('UTF-7', salt.utils.stringutils.to_bytes('\x2b\x2f\x76\x2f')),
+        ]
+        for _encoding, bom in boms:
+            if _data.startswith(bom):
+                log.debug('Found BOM for {0}'.format(_encoding))
```
terminalmage: We should be using the %s string substitution in log messages:

```python
log.debug('Found BOM for %s', _encoding)
```

tests/unit/modules/test_file.py
```diff
@@ -235,7 +235,10 @@ def setup_loader_modules(self):
                     'grains': {},
                 },
                 '__grains__': {'kernel': 'Linux'},
-                '__utils__': {'files.is_text': MagicMock(return_value=True)},
+                '__utils__': {
+                    'files.is_binary': MagicMock(return_value=False),
+                    'files.get_encoding': MagicMock(return_value=None)
```
terminalmage: You're mocking thiis to return `None`, but that function never actually returns `None`, might that cause problems in terms of the accuracy of our testing?

In addition, I'm not sure it's a great idea to mock these functions at the module level, rather than just patching in individual tests, unless a given function is used in a lot of tests and needs to be consistently mocked with the same result.

terminalmage (changes requested):

terminalmage (approved):

Issue #47574, opened 2018-05-09
Documentation error in description of watch_any


cfxegbert:
https://docs.saltstack.com/en/latest/ref/states/requisites.html

The example yaml for watch_any is:
````
apache2:
  service.running:
    - watch_any:
      - file: /etc/apache2/sites-available/site1.conf
      - file: /etc/apache2/sites-available/site2.conf
  file.managed:
    - name: /etc/apache2/sites-available/site1.conf
    - source: salt://apache2/files/site1.conf
  file.managed:
    - name: /etc/apache2/sites-available/site2.conf
    - source: salt://apache2/files/site2.conf
````

This yaml has a dictionary "apache2" with the key "file.managed" listed twice.

Ch3LL:
fixed in https://github.com/saltstack/salt/pull/47589  thanks for bringing this to our attention :)

PR #47589, opened 2018-05-10
Remove duplicate file id in watch_any doc example

Ch3LL:
### What does this PR do?
Update docs to remove duplicate ID. As far as I know this is still not possible within salts rendering to allow multiple calls of the same state module under one ID. 

https://github.com/saltstack/salt/issues/47574

### DIFF
--- a/doc/ref/states/requisites.rst
+++ b/doc/ref/states/requisites.rst
@@ -410,16 +410,12 @@ exactly like the ``require`` requisite (the watching state will execute if
       service.running:
         - watch_any:
           - file: /etc/apache2/sites-available/site1.conf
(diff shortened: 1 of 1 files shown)


rallytime (approved):

garethgreenaway (approved):

cfxegbert (commented):
Removing one of the files sort of defeats the purpose of this example.  It needs two or more states in the watch_any statement to be a useful example.  Change one of the file.managed to some other state or move one of the file.managed out of the state.

Ch3LL:
thanks for catching that @cfxegbert . i updated it with a more appropriate example if you want to re-review :)

cfxegbert (approved):
Looks good to me

PR #47601, opened 2018-05-10
Skip tests when we can not use runas

dwoz:
### What does this PR do?

Skip tests that require runas when we are running under a limited winrm shell.

### Commits signed with GPG?

Yes


cachedout (approved):

dwoz:
re-run py

dwoz:
re-run all

twangboy (approved):

gtmanfred (approved):

dwoz:
re-run lint

rallytime:
re-run lint

tests/integration/states/test_pip_state.py
```diff
@@ -37,13 +37,34 @@
...
+    '''
+    Detect if we are running in a limited shell (winrm) and are un-able to use
+    the runas utility method.
+    '''
+    if salt.utils.is_windows():
+        try:
+            salt.utils.win_runas.run_as(
+                'cmd.exe /c echo 1', 'noexistuser', 'n0existp4ss',
+            )
+        except WindowsError as exc:  # pylint: disable=E060
```
rallytime: Looks like the linter doesn't like `E060`. You should be able to replace that just with `undefined-variable` instead.

rallytime (commented):
I linted this by hand since the lint run is down. Can you make that change really quickly? Then we can get this in.

PR #47298, opened 2018-04-25
ec2: Use exponential backoff for describeinstance

ghost:
### What does this PR do?
Be more conservative when running describeInstance on ec2 API, use exponential backup and sleep up to one minute.

### What issues does this PR fix or reference?
Fixes fails in creating instances, where the instance creation takes over 5 seconds and is therefore not done before the function runs out of retries fails.

### Previous Behavior
linear back-off with 5 retires spaced 1 second

### New Behavior
exponential back-off with 7 retries from 1 second to 64 seconds

### Tests written?

No

### Commits signed with GPG?

No

PR #47211, opened 2018-04-20
Fix for AWS EC2 Rate Limit Failures in salt-cloud

edlane:
### What does this PR do?
Implements these fixes to "rate limit failures" suggested by AWS referenced **here:**

https://docs.aws.amazon.com/AWSEC2/latest/APIReference/query-api-troubleshooting.html (scan down to "Calculating the sleep interval")

**also:**
https://docs.aws.amazon.com/general/latest/gr/api-retries.html

**and also for the really nerdy:**
https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter

### What issues does this PR fix or reference?
Experiencing sporadic failures while deploying large (~50 node clusters) in Amazon EC2 when using
salt-cloud libraries in parallel (asynchronously).
This behavior has been demonstrated to occur to a greater or lesser degree with these salt-cloud APIs:
- salt-api,
- salt-runner,
- salt-cloud (using map files and "--parallel" option)

### Previous Behavior
Starting a cluster of 50 VMs in rapid succession using salt-cloud or salt-api
will experience sporadic failures with the following EC2 error:

```
[ERROR   ] AWS Response Status Code and Error: [503 503 Server Error: Service Unavailable for url: https://ec2.us-west-1.amazonaws.com/?Action=RunInstances&ImageId=ami-02d0d962&InstanceType=t2.micro&KeyName=edlane&MaxCount=1&MinCount=1&NetworkInterface.0.AssociatePublicIpAddress=true&NetworkInterface.0.DeleteOnTermination=true&NetworkInterface.0.DeviceIndex=0&NetworkInterface.0.SubnetId=subnet-36e8e570&Version=2014-10-01] {'Errors': {'Error': {'Message': 'Request limit exceeded.', 'Code': 'RequestLimitExceeded'}}, 'RequestID': 'e70bd27a-f596-464e-ac42-321a810aa4fb'}; Attempts remaining: 0
```

Of 50 instances  started by:

- salt-cloud "--parallel" only 36/50 were successful
- salt-api-client "async" 47/50 were successful
- salt-runner-client "async" 40/50 were successful

Code snippet to create 50 VMs using the salt-api with this python code snippet:

```
for i in xrange(50):
    utime = str(int(time.time()*100000))
    instance_name = "ed-deletemenow-{}-{}".format(i, utime)
    resp = session.post('http://localhost:8000',
                        json=[{
                            'client': 'runner_async',
                            'fun': 'cloud.profile',
[...]

PR #47609, opened 2018-05-11
Unify retries for AWS API

ghost:
### What does this PR do?

Unify the retries for AWS API calls
This is a follow-up to #47211 and #47298

* Use exponential backoff (instead of linear) in all relevant places
* Share MAX_RETRIES and code for exponential_backoff
* Remove old (now duplicate) retry loops

### Previous Behavior
Linear retries, double retry loops

### New Behavior
Identical retry setup (exponential with same MAX_RETRIES) for all relevant AWS API calls.

### Tests written?
No

### Commits signed with GPG?
No

### DIFF
--- a/salt/utils/aws.py
+++ b/salt/utils/aws.py
@@ -52,6 +52,8 @@ AWS_RETRY_CODES = [
 ]
 AWS_METADATA_TIMEOUT = 3.05
 
+AWS_MAX_RETRIES = 7
+
 IROLE_CODE = 'use-instance-role-credentials'
 __AccessKeyId__ = ''
 __SecretAccessKey__ = ''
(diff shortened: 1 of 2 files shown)


rrroo:
`aws.query` does retries correctly since #47211 
With this PR more cases rely on aws.query directly, but in some cases additional retry logic is needed, which would imo overcomplicate aws.query, so I left it in ec2.py.

rrroo:
cc @gtmanfred @edlane @rallytime 
Since you were involved in the previous PRs you might have an opinion here.

gtmanfred (approved):

salt/utils/aws.py
```diff
@@ -61,6 +63,20 @@
...
+    """
+    backoff an exponential amount of time to throttle requests
+    during "API Rate Exceeded" failures as suggested by the AWS documentation here:
+    https://docs.aws.amazon.com/AWSEC2/latest/APIReference/query-api-troubleshooting.html
+    and also here:
+    https://docs.aws.amazon.com/general/latest/gr/api-retries.html
+    Failure to implement this approach results in a failure rate of >30% when using salt-cloud with
+    "--parallel" when creating 50 or more instances with a fixed delay of 2 seconds.
+    A failure rate of >10% is observed when using the salt-api with an asyncronous client
+    specified (runner_async)."""
```
rallytime: Can you move these closing quotes to it's own line? It's a little tricky to see where those end. Would you also mind changing those to single quotes instead of double? That's what we use in Salt's [style guide](https://docs.saltstack.com/en/latest/topics/development/conventions/style.html).

rallytime (commented):
I just have one small request on this, otherwise this looks great.

rallytime (approved):

rallytime:
@rrroo This looks good to me now. There are a couple of small lint errors. Once those are fixed, we should be able to get this in.

@edlane Did you get a chance to look over this?

rrroo:
@rallytime The linter is green.

PR #47610, opened 2018-05-11
restartcheck: add NILRT ARM kernel version detection

10ne1:
### What does this PR do?
 
Adds the capability to detect ARM kernel versions to restartcheck on NI Linux RT platforms.

### Tests written?

No

### Commits signed with GPG?

No


rallytime:
@isbm You came up on GitHub's reviewers suggestions list, so I've requested a review from you here. Would you mind taking a look?

salt/modules/restartcheck.py
```diff
@@ -314,23 +314,40 @@ def _kernel_versions_nilrt():
...
+    kver = None
+
+    def _get_kver_from_bin(kbin):
+        '''Get kernel version from a binary image or None if detection fails'''
+        kvregex = r'[0-9]+\.[0-9]+\.[0-9]+-rt.*(?=\s\()'
+        kernel_strings = __salt__['cmd.run']('strings {0}'.format(kbin))
+        re_result = re.search(kvregex, kernel_strings)
+        return None if re_result is None else re_result.group(0)
+
+    if _is_older_nilrt():
```
isbm: Oh, very nice! `if distribution_i_use_at_my_office():` anyone? :wink:

@10ne1 since seems NILRT does not have a proper versioning, then can we come up with something generic instead, like property checker etc, which will do decision based on that (say, "foo is installed", "bar is located there" etc) instead of hardcoding "older version" or "even_older_version" right into the code?

isbm (changes requested):

isbm:
@rallytime seems your bot works very well! Last time I was reviewing this indeed. You should also probably take to the account number of comments posted by a reviewer... :laughing:

salt/modules/restartcheck.py
10ne1: I think I can modify the distro rootfs without breaking backwards compatibility such that the lsb_release DISTRIB_ID field reflects the difference between these older/newer versions and that would automatically get picked up by the existing lsb_distrib_id grain, so then we'd just have to test for grains[lsb_distrib_id] == 'nilrt'.

Are you ok with this approach?

salt/modules/restartcheck.py
```diff
@@ -314,23 +314,40 @@ def _kernel_versions_nilrt():
...
-        # backwards compatible so it'll just get more complicated.
-        kpath = '/boot/runmode/bzImage'
-        kernel_strings = __salt__['cmd.run']('strings {0}'.format(kpath))
-        re_result = re.search(r'[0-9]+\.[0-9]+\.[0-9]+-rt.*(?=\s\()', kernel_strings)
-        if re_result is not None:
-            kernel_versions.append(re_result.group(0))
+    kver = None
+
+    def _get_kver_from_bin(kbin):
+        '''Get kernel version from a binary image or None if detection fails'''
```
rallytime: Can you split this doc string into 3 lines instead of 1?

salt/modules/restartcheck.py
10ne1: @isbm A collegue of mine posted a PR implementing my suggestion, if that gets merged we can test the lsb grain in this one https://github.com/saltstack/salt/pull/47951

salt/modules/restartcheck.py
10ne1: @rallytime Yes, done.

10ne1:
Kind reminder. Ping?

isbm (approved):
@10ne1 oh, sorry! Thanks for the reminder, it was really needed. Yeah, looks better. I still don't like that you need to re-check the path to the image and _in my personal_ opinion I would move it away to a separate function, like `_get_kernel_path`. If you would be kind to do that, go ahead.

Otherwise you can leave it as is for now.

rallytime (approved):

PR #47621, opened 2018-05-12
Runas any user even when shell is limited like winrm

dwoz:
### What does this PR do?

- Fix runas when running under winrm.
- Support for LOCAL SERVICE and NETWORK SERVICE system accounts.
- Runas can now use system accounts from salt-call. 
   (SYSTEM, LOCAL SERVICE and NETWORK SERVICE)
- Runas can launch processes on behalf of users **without a password**. 

- Some tests fail when run over winrm because of restrictions.

### Tests written?

No

### Commits signed with GPG?

Yes


damon-atkins:
Please use a class for constants like in win_reg.py

dwoz:
This should fix

#47621

salt/utils/win_runas.py
```diff
@@ -44,422 +40,136 @@ def __virtual__():
...
-
-    # Get Security Attributes
     security_attributes = win32security.SECURITY_ATTRIBUTES()
     security_attributes.bInheritHandle = 1
 
-    # Create a pipe to set as stdout in the child. The write handle needs to be
-    # inheritable.
     stdin_read, stdin_write = win32pipe.CreatePipe(security_attributes, 0)
-    stdin_read = make_inheritable(stdin_read)
+    stdin_read = salt.utils.winutil.make_inheritable(stdin_read)
```
twangboy: And here

salt/utils/win_runas.py
```diff
@@ -44,422 +40,136 @@ def __virtual__():
...
 
     stdout_read, stdout_write = win32pipe.CreatePipe(security_attributes, 0)
-    stdout_write = make_inheritable(stdout_write)
+    stdout_write = salt.utils.winutil.make_inheritable(stdout_write)
```
twangboy: same here

salt/utils/win_runas.py
```diff
@@ -44,422 +40,136 @@ def __virtual__():
...
-    # Get Standard Out
-    fd_out = msvcrt.open_osfhandle(stdout_read, os.O_RDONLY | os.O_TEXT)
-    with os.fdopen(fd_out, 'r') as f_out:
-        ret['stdout'] = f_out.read()
-
-    # Get Standard Error
-    fd_err = msvcrt.open_osfhandle(stderr_read, os.O_RDONLY | os.O_TEXT)
-    with os.fdopen(fd_err, 'r') as f_err:
-        ret['stderr'] = f_err.read()
+    stderr_write = salt.utils.winutil.make_inheritable(stderr_write)
```
twangboy: and here

salt/utils/win_runas.py
```diff
@@ -44,422 +40,136 @@ def __virtual__():
...
+    )
+
+    hProcess = process_info.hProcess
+    hThread = process_info.hThread
+    dwProcessId = process_info.dwProcessId
+    dwThreadId = process_info.dwThreadId
+
+    salt.utils.winutil.kernel32.CloseHandle(stdin_write.handle)
+    salt.utils.winutil.kernel32.CloseHandle(stdout_write.handle)
+    salt.utils.winutil.kernel32.CloseHandle(stderr_write.handle)
```
twangboy: And these

salt/utils/win_runas.py
```diff
@@ -44,422 +40,136 @@ def __virtual__():
...
     if '@' in username:
         username, domain = username.split('@')
     if '\\' in username:
         domain, username = username.split('\\')
+    return username, domain
+
+
+def runas(cmdLine, username, password=None, cwd=None, elevated=True):
+
+    impersonation_token = salt.utils.winutil.impersonate_sid(
```
twangboy: Shouldn't this be `salt.winutil` now?

salt/utils/win_runas.py
```diff
@@ -44,422 +40,136 @@ def __virtual__():
...
+            domain,
+            '',
+            win32con.LOGON32_LOGON_SERVICE,
+            win32con.LOGON32_PROVIDER_DEFAULT,
+        )
+    elif password:
+        log.warn("Logon user with password: %s")
+        #user_token = salt.utils.winutil.logon_msv1(
+        #    username, password, domain=domain,
+        #).Token
```
twangboy: Could you remove these commented lines here?

salt/utils/win_runas.py
```diff
@@ -44,422 +40,136 @@ def __virtual__():
...
+        user_token = win32security.LogonUser(
+            username,
+            domain,
+            password,
+            win32con.LOGON32_LOGON_INTERACTIVE,
+            win32con.LOGON32_PROVIDER_DEFAULT,
+        )
+    else:
+        log.warn("Logon user without password: %s")
+        user_token = salt.utils.winutil.logon_msv1_s4u(username).Token
```
twangboy: And here

salt/utils/win_runas.py
```diff
@@ -44,422 +40,136 @@ def __virtual__():
...
+    )
+    if elevation_type > 1:
+        user_token = win32security.GetTokenInformation(
+            user_token,
+            win32security.TokenLinkedToken
+        )
+        salt.utils.winutil.elevate_token(user_token)
+
+    handle_reg = win32profile.LoadUserProfile(user_token, {'UserName': username})
+    salt.utils.winutil.grant_winsta_and_desktop(user_token)
```
twangboy: And here

salt/utils/win_runas.py
```diff
@@ -44,422 +40,136 @@ def __virtual__():
...
-        ret['stderr'] = f_err.read()
+    stderr_write = salt.utils.winutil.make_inheritable(stderr_write)
+
+    creationflags = (
+        win32process.CREATE_NO_WINDOW |
+        win32process.CREATE_NEW_CONSOLE |
+        win32process.CREATE_SUSPENDED
+    )
+
+    startup_info = salt.utils.winutil.STARTUPINFO(
```
twangboy: You get the idea...

salt/utils/win_runas.py
```diff
@@ -44,422 +40,136 @@ def __virtual__():
...
+        dwFlags=win32con.STARTF_USESTDHANDLES,
+        hStdInput = stdin_read.handle,
+        hStdOutput = stdout_write.handle,
+        hStdError = stderr_write.handle,
+    )
+
+    env = win32profile.CreateEnvironmentBlock(user_token, False)
+
+    process_info = salt.utils.winutil.CreateProcessWithTokenW(
```
twangboy: and here

salt/utils/win_runas.py
```diff
@@ -44,422 +40,136 @@ def __virtual__():
...
+        stderr = f_err.read()
+        ret['stderr'] = stderr
+
+    win32profile.UnloadUserProfile(user_token, handle_reg)
+#    salt.utils.winutil.kernel32.CloseHandle(handle_reg)
+    salt.utils.winutil.kernel32.CloseHandle(hProcess)
+#    salt.utils.winutil.kernel32.CloseHandle(user_token)
+    if impersonation_token:
+        win32security.RevertToSelf()
+#    salt.utils.winutil.kernel32.CloseHandle(impersonation_token)
```
twangboy: Can we remove these comments?
And the `salt.winutil` thing

salt/winutil.py
```diff
@@ -0,0 +1,1165 @@
+# -*- coding: utf-8 -*-
+'''
+Windows specific utility functions, this module should be imported in a try,
+except block because it is only applicable on Windows platforms.
+
+
+Much of what is here was adappted from the following:
```
twangboy: adapted

twangboy (changes requested):

damon-atkins:
Some random thoughts.
Not sure module salt.win is the correct place.   Maybe `salt.utils.platform.win.somthing` (if you break the file up).    Maybe along the lines of how powershell is broken up  e.g. `salt.utils.platform.win.credential` class.
`Get-<Name Of Python Class>` e.g. Get-Credentials
`Get-Module -ListAvailable | Format-Table Name, Description`

There are also a few win_*.py in salt.utils

For example we have salt.utils.pkg.win

dwoz:
@damon-atkins, thanks so much for reviewing this. :)

`salt.win` came from discussions with @twangboy. The main driver for `salt.win` is to have a place for Windows specific stuff. For now `salt.win` only contains things related to logon. The plan is to expand on this module to make `salt.win.logon`, `salt.win.service`, `salt.win.registry` ect. I did not put the module under `salt.utils` to avoid having the loader try to import this platform specifics automatically. I think anything under `salt.utils` gets touched by the loader automatically (I could be wrong on this). I'm down with `salt.utils.platform` provided the loader plays nicely. If not, maybe `salt.platform` works as a namespace for these things. Maybe @s0undt3ch or @cachedout have some opinions?

salt/utils/win_runas.py
```diff
@@ -44,422 +39,135 @@ def __virtual__():
...
+    impersonation_token = salt.win.impersonate_sid(
+        salt.win.SYSTEM_SID,
+        session_id=0,
+        privs=['SeTcbPrivilege'],
+    )
+
+    username, domain = split_username(username)
+    sid, domain, sidType = win32security.LookupAccountName(domain, username)
+    if domain == 'NT AUTHORITY':
+        log.warn("Logon system account: %s", username)
```
s0undt3ch: Why warning?
Also, switch to `log.warning`, `log.warn` is deprecated in Py3.

salt/win.py
```diff
@@ -0,0 +1,1164 @@
+# -*- coding: utf-8 -*-
+'''
+Windows specific utility functions, this module should be imported in a try,
```
s0undt3ch: If this is a utilities module it should be moved to `salt.utils`.

s0undt3ch (changes requested):

salt/utils/win_runas.py
dwoz: I agree warning is overkill. Should these 'logon user' log statements go away completely?

s0undt3ch: If you think these will make sense when trying to debug an issue, leave then, but perhaps, at debug log level? Info?

dwoz: Windows creates an event log for the logons so I've just removed these log statements.

salt/utils/win_runas.py
```diff
@@ -1,34 +1,29 @@
...
 '''
-from __future__ import absolute_import, print_function, unicode_literals
+from __future__ import absolute_import, unicode_literals
 
 # Import Python Libraries
 import os
 import logging
 
 # Import Third Party Libs
+import psutil
```
cachedout: Achtung! This needs to either use the psutil compat lib in `salt.utils` or it needs to be conditionally imported.

cachedout (dismissed):

damon-atkins:
I am happy with salt.platform `salt/platform/<something>/win.py`
I do not understand why the loader is used with utils. A lot of the code imports utils.   
Some of the code I have developed is way to big for a single file, need to split it into parts.

dwoz:
@damon-atkins I'd prefer not to go to far down the rabbit hole of future proofing here. Are you cool with `salt/platform/win.py` for now?

damon-atkins:
Given its not an interface for users can access then it can be rename/moved without impact. Seems fine.

rallytime:
@dwoz Bump. :) Have you been able to come back to this and respond to the feedback above?

dwoz:
@damon-atkins, @cachedout, @twangboy 

Does everything look good now?

salt/utils/win_runas.py
```diff
@@ -38,428 +38,135 @@ def __virtual__():
...
-        privileges = win32security.GetTokenInformation(
-            elevated_token, win32security.TokenPrivileges)
-
-        # Create a set of all privileges to be enabled
-        enable_privs = set()
-        for luid, flags in privileges:
-            enable_privs.add((luid, win32con.SE_PRIVILEGE_ENABLED))
-
-        # Enable the privileges
-        win32security.AdjustTokenPrivileges(elevated_token, 0, enable_privs)
```
twangboy: You might still need this for some scripts to run (303 - 309). Even though it has the token, not all privileges are enabled. This just enables all of them.

dwoz: @twangboy Based on our Slack discussion. I agree and will put it back.

twangboy (approved):

salt/utils/win_runas.py
dwoz: Turns out this was moved to the `elevate_token` function.

twangboy: I guess I missed that.

salt/platform/win.py
```diff
@@ -0,0 +1,1171 @@
...
+from salt.ext.six.moves import range
+from salt.ext.six.moves import zip
+
+import win32con
+import win32con
+import win32api
+import win32process
+import win32security
+import win32service
+import ntsecuritycon
```
damon-atkins: Consider importing these are using these constants, rather than defining your own (assumed)   
```
 import pywintypes
 import winerror
```

damon-atkins (changes requested):
See Comment

salt/platform/win.py
dwoz: I'd like to get this merged. I'm committed to a lot of other work for the Fluorine release. If needed, let's make a separate issue about cleaning up the constants and assign it to me. I'll tackle that as time permits.

damon-atkins (approved):

twangboy:
I'm getting the following error on Py3 when I do a `cmd.run` running as a normal (non-admin) user. I haven't tried with an admin user. Here's the command I ran and the result on both Py2 and Py3.

```
root@shanelee-master:/srv/salt/test# salt -t 300 dev* cmd.run whoami runas=testuser01 password=PassWord1!
dev(py2):
    win-8fgt3e045se\testuser01
dev3(py3):
      File "c:\dev\salt\salt\minion.py", line 1619, in _thread_return
      File "c:\dev\salt\salt\executors\direct_call.py", line 12, in execute
        return func(*args, **kwargs)
      File "c:\dev\salt\salt\modules\cmdmod.py", line 1161, in run
        **kwargs)
      File "c:\dev\salt\salt\modules\cmdmod.py", line 411, in _run
        return win_runas(cmd, runas, password, cwd)
      File "c:\dev\salt\salt\utils\win_runas.py", line 62, in runas
        privs=['SeTcbPrivilege'],
      File "c:\dev\salt\salt\platform\win.py", line 1125, in impersonate_sid
        raise WindowsError("Impersonation failure")  # pylint: disable=undefined-variable
    OSError: Impersonation failure
root@shanelee-master:/srv/salt/test#
```

doc/topics/releases/fluorine.rst
```diff
@@ -870,3 +870,13 @@ for viewing minions, runners, and jobs as well as running execution modules
...
+
+Security
+========
+
+Windows runas changes
+---------------------
+
+A password is no longer required with ``runas`` under normal circumstances.
+The password option is only needed if the minion process is run under a
+restricted (non-administrator) account.
```
cachedout: As I recall, password is only needed if it's under a non-admin account AND it's requesting priv escalation, right? If so, I think that should be called out specifically.

tests/support/helpers.py
```diff
@@ -685,7 +685,7 @@ def wrap(cls):
                             username
                         )
                     )
-                    create_user = cls.run_function('user.add', [username])
+                    create_user = cls.run_function('user.add', [username], **kwargs)
```
cachedout: Is the intention here just to pass `groups`? I'm wary of passing all kwargs unless that's truly needed.

doc/topics/releases/fluorine.rst
dwoz: @cachedout If the minion happens to be running under a non admin. They will need to use a password anytime the use the `runas` argument to `cmd.run`. Maybe this is better?

```
 A password is no longer required with ``runas`` under normal circumstances.
The password option is only needed if the minion process is run under a
restricted (non-administrator) account. In the aforementioned case, a password
is only required when using the ``runas`` argument to run command as a different 
user.
```

tests/support/helpers.py
dwoz: @cachedout Yes, this is to pass groups. I'll change it to be more specific.

dwoz: Actually, looking at this again. We are making the kwargs dictionary for the sole purpose of passing it to `user.add`. It is being used to pass timeout, groups, and password.

rallytime:
@dwoz This has a merge conflict with the release notes. Can you fix that up?

dwoz:
@rallytime fixed.

PR #47626, opened 2018-05-13
Implement Pipfile and remove extra requirements.txt files

gtmanfred:
### What does this PR do?
Instead of having extra requirements.txt files, we can just specify only install dependencies based on which version of python we are using.

Also, add a Pipfile for pipenv, because it is super awesome, and gives us lockfiles/virtualenvs.

Based on what we saw at PyCon 2018, this will be a super awesome way to manage pip stuff and will be included in pip in the future.

https://us.pycon.org/2018/schedule/presentation/243/

### Commits signed with GPG?

Yes


.gitignore
```diff
@@ -11,6 +11,7 @@ MANIFEST
 *.wpr
 *.wpu
 *.DS_Store
+Pipfile.lock
```
terminalmage: The creator of pipenv recommended keeping the Pipfile and lockfile both in version control.

gtmanfred: Agreed, once I learned that you can still install newer stuff even if the lock file is there, it makes sense that you should put the lock file in VCS.

gtmanfred:
Thanks @kennethreitz for pipenv

Pipfile
```diff
@@ -0,0 +1,33 @@
...
+
+[dev-packages]
+mock = ">=2.0.0"
+apache-libcloud = ">=0.14.0"
+boto = ">=2.32.1"
+boto3 = ">=1.2.1"
+moto = ">=0.3.6"
+SaltPyLint = ">=v2017.3.6"
+pytest = ">=3.5.0"
+Pytest-catchlog = {git = "git://github.com/eisensheng/pytest-catchlog.git", ref = "develop", markers = "python_version < '3.0'"}
```
s0undt3ch: Drop this one, if it `py.test` explodes, I have a fix

requirements/dev.txt
```diff
@@ -0,0 +1,17 @@
+-r base.txt
+
+mock>=2.0.0
+apache-libcloud>=0.14.0
+boto>=2.32.1
+boto3>=1.2.1
+moto>=0.3.6
+SaltPyLint>=v2017.3.6
+pytest>=3.5.0
+git+https://github.com/eisensheng/pytest-catchlog.git@develop#egg=Pytest-catchlog; python_version < '3.0'
```
s0undt3ch: same here

s0undt3ch (approved):
:tada:

Pipfile
gtmanfred: Done

requirements/dev.txt
gtmanfred: Done

gtmanfred: looks like pytest still works.

kennethreitz:
This is awesome! ✨🍰✨

requirements/dev.txt
s0undt3ch: :applause:

tests/conftest.py
```diff
@@ -102,6 +102,14 @@ def pytest_addoption(parser):
              'SSH server on your machine. In certain environments, this '
              'may be insecure! Default: False'
     )
+    test_selection_group.addoption(
```
rallytime: Is this change intentional? This seems unrelated to the rest of the changes here.

requirements/dev_python27.txt
```diff
@@ -1,11 +1,2 @@
...
-mock>=2.0.0
-apache-libcloud>=0.14.0
-boto>=2.32.1
-boto3>=1.2.1
-moto>=0.3.6
-SaltPyLint>=v2017.3.6
-pytest>=3.5.0
-git+https://github.com/eisensheng/pytest-catchlog.git@develop#egg=Pytest-catchlog
-git+https://github.com/saltstack/pytest-salt.git@master#egg=pytest-salt
+# This is for legacy purposes
```
rallytime: I know this is picky, but it might make sense for this comment to match the one in `dev_python34.txt` (and I liked that one the best of the two). 😄

requirements/dev_python27.txt
```diff
@@ -1,11 +1,2 @@
--r base-py2.txt
-
-mock>=2.0.0
-apache-libcloud>=0.14.0
-boto>=2.32.1
-boto3>=1.2.1
-moto>=0.3.6
-SaltPyLint>=v2017.3.6
-pytest>=3.5.0
-git+https://github.com/eisensheng/pytest-catchlog.git@develop#egg=Pytest-catchlog
```
rallytime: I don't see the pytest-catchlog in the new dev.txt file. I am not sure what that particular addition to the Py27 requirements file is for since there's no comment about it, but I wanted to double check that this was intentional to leave that out.

rallytime (commented):
This is neat! I have a couple of very picky suggestions and a question on this change.

tests/conftest.py
gtmanfred: It was intentional, it fixes pytest for salt, so you can run integration tests.

I can move it to a new PR if you want.

requirements/dev_python27.txt
gtmanfred: Will do

requirements/dev_python27.txt
gtmanfred: pytest-catchlog has been merged into pytest upstream, so we no longer needed it specified.

s0undt3ch (approved):

requirements/dev_python27.txt
rallytime: Perfect!

tests/conftest.py
rallytime: Nah, it's good. Just making sure.

rallytime (approved):

PR #47641, opened 2018-05-14
fix _create_stream and tornado 5.0

gtmanfred:
### What does this PR do?
This should be the last fix for tornado 5.0

### Tests written?

Yes

### Commits signed with GPG?

Yes


gtmanfred:
Hrm, yeah it is still hanging on the whatever happens to be the second test_tcp test that it runs.

I will look at this tomorrow as well.

rallytime (approved):

PR #47645, opened 2018-05-14
query the pip path for test test_issue_2087_missing_pip

Ch3LL:
### What does this PR do?
The following tests are failing on macosx python3:

```
integration.modules.test_pip.PipModuleTest.test_issue_2087_missing_pip                                                                                                              
integration.states.test_pip.PipStateTest.test_issue_2087_missing_pip
```

This is because when the virtualenv is created its created with python2 because the default `python` is 2 on the macosx hosts. My understanding of the virtualenv.create module is that it will use the default python version installed unless `python=python3` is passed to the command. I could be wrong about this code analysis as I am not certain if its suppose to grab the version of python that salt is running.

This PR changes the test to query what the site-packages path is for the pip package before we try to remove it, instead of assuming the path is the same version of python as the test runner.

### DIFF
--- a/tests/integration/modules/test_pip.py
+++ b/tests/integration/modules/test_pip.py
@@ -11,7 +11,6 @@ from __future__ import absolute_import
 import os
 import re
 import shutil
-import sys
 import tempfile
 
 # Import Salt Testing libs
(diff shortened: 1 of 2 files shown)


rallytime (approved):

twangboy:
@Ch3LL Actually I think the version of is bundled within virtualenv itself and does not rely on the System python. At least that's what I think @terminalmage and I worked out.

twangboy (approved):

Issue #47651, opened 2018-05-14
states/boto3_route53 issue retrieving wildcard records


nullify005:
### Description of Issue/Question

The upstream is here: https://github.com/boto/boto/issues/818

Essentially when you go searching for '*.domain.com' resource records you won't get anything back from the API.

If, however, you look for '\052.domain.com' you will end up getting the resource records matching the wildcard entry.

As such any time you try to look for '*.whatever' you need to replace it with '\052.whatever', then do whatever processing you want to the records

### Setup
(Please provide relevant configs and/or SLS files (Be sure to remove sensitive info).)

### Steps to Reproduce Issue

Looking for '*' yields nothing
```
(aws-ami-cyan) Madison:aws-ami-cyan lwebb$ ./bin/salt-call --local -c `pwd` boto3_route53.get_resource_records HostedZoneId=BLAHBLAHBLAH StartRecordName='*.domain.com.' StartRecordType=A
local:
```

Looking for '\052' gets the record we actually want
```
(aws-ami-cyan) Madison:aws-ami-cyan lwebb$ ./bin/salt-call --local -c `pwd` boto3_route53.get_resource_records HostedZoneId=BLAHBLAHBLAH StartRecordName='\052.domain.com.' StartRecordType=A
local:
    |_
      ----------
      AliasTarget:
          ----------
          DNSName:
              BLAHBLAHBLAH.ap-southeast-2.elb.amazonaws.com.
          EvaluateTargetHealth:
              False
          HostedZoneId:
              BLAHBLAHBLAH
      Name:
          \052.domain.com.
      Type:
          A
```

### Versions Report

Note that the salt version I'm using is a patched 2017.7.2 ... this happens in develop too though
```
Salt Version:
           Salt: 2017.7.2-3-gf0473af
 
Dependency Versions:
           cffi: 1.8.3
       cherrypy: Not Installed
       dateutil: 2.6.0
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.10
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.5.1
   mysql-python: Not Installed
      pycparser: 2.17
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: Not Installed
[...]

PR #47652, opened 2018-05-14
Use the unicode version of the '*' character for route53

nullify005:
### What does this PR do?

Basically when you are looking for resource records which are for wildcard entries in a domain you have to search for the unicode representation of the '*' character.

If you do that then the subsequent matching of the object for the state works correctly & skips any modifications if the RRSET matches what you want, if not then you have to literally wait a thousand years for route53 to report that it's made the changes every time you run the state

https://github.com/saltstack/salt/issues/47651

### Previous Behavior

Any state changes for a wildcard entry would always result in a state change even if the record was already there in the desired state

### New Behaviour

Correctly finds the wildcard recordset so that comparisons can be made & therefore skip any changes if the state is correct

### Tests written?

No

### Commits signed with GPG?

No


rallytime:
re-run all

rallytime (approved):
LGTM

rallytime:
This has a small lint violation: https://jenkins.saltstack.com/job/PR/job/salt-pr-lint-n/21821/violations/file/salt/states/boto3_route53.py/

@nullify005 Can you fix that?

cachedout (approved):

cachedout:
@ryan-lane Anything to add here?

nullify005:
Ok, so I've just discovered that there is a `_to_aws_encoding` & an unimplemented `_from_aws_encoding` methods in the state which looks like it's trying to do what I want but doesn't seem to get there

Rather than putting my hack PR in like it is, I think I'll take the time to fix both those methods to work properly for this case (& possibly others?)

nullify005:
I've moved my 'hack' into the module instead of doing it in the state, which to me feels better
IMO the get_resource_records method should get the records I want, regardless of the encoding & it shouldn't be the states responsibility to do this

As such there is now a _to & _from_aws_encoding method which works on the names & the returns to allow for matching against strings instead of pre-encode them because you know there will be an issue.

Note that if anyone is specifically doing any encoding or un-encoding on their own (either in sls or a module or whatever) then this may break things for them, so a note should be made somewhere about this.

I believe that these changes should cater for all the different characters the the Route53 API messes with rather than just '*' ... 🤞

damon-atkins:
`Amazon Route 53 stores alphabetic characters as lowercase letters (a-z), regardless of how you specify them: as uppercase letters, lowercase letters, or the corresponding letters in escape codes.`

This would mean if you input HELLO you would get back HELLO I assume. Does the state  internally convert to lower for comparison if required?

This might me of interest
https://docs.python.org/3/library/codecs.html#module-encodings.idna PY3 may do it already.
https://tools.ietf.org/html/rfc3490  # Read a few RFC, this one seems like a hard read.

damon-atkins:
@cachedout   Suggest python inbuilt  x.encode('idna') functions are used for the conversion.

salt/modules/boto3_route53.py
```diff
@@ -746,13 +814,13 @@ def get_resource_records(HostedZoneId=None, Name=None, StartRecordName=None,
         args.update({'PrivateZone': PrivateZone}) if PrivateZone is not None else None
         zone = find_hosted_zone(**args)
         if not zone:
-            log.error("Couldn't resolve domain name %s to a hosted zone ID.", Name)
+            log.error("Couldn't resolve domain name {} to a hosted zone ID.".format(Name))
```
terminalmage: Please revert this back to how it was before, we use % subsitution in logging for good reason.

salt/modules/boto3_route53.py
```diff
@@ -714,6 +714,74 @@ def delete_hosted_zone_by_domain(Name, PrivateZone=None, region=None, key=None,
...
+    REALLY need this, just embed '\056' directly in your strings.  I can't promise it won't break
+    your DNS out in the real world though.
+    '''
+    instring = instring.lower()                                           # Always lowercase
+    send_as_is = list(range(97, 123))                   # a-z is 97-122 inclusive
+    send_as_is += [ord(n) for n in ['0', '1', '2', '3', '4', '5', '6', '7', '8',
+                                    '9', '-', '_', '.']]
+    send_as_octal = [ord(n) for n in ['!', '"', '#', '$', '%', '&', "'", '(', ')', '*',
+                                      '+', ',', '/', '\\', ':', ';', '<', '=', '>', '?',
+                                      '@', '[', ']', '^', '`', '{', '|', '}']]
```
terminalmage: There's no need to define these as lists, Python will happily iterate over each character in a string. Given that there are quotes in `send_as_octal`, you'd want to use a triple-quote to surround that string. For example:

```python
send_as_is += [ord(n) for n in '0123456789-_.']
send_as_octal = [ord(n) for n in '''!"#$%&'()*+,/\\:;<=>?'@[]^`{|}''']
```

salt/modules/boto3_route53.py
```diff
@@ -714,6 +714,74 @@ def delete_hosted_zone_by_domain(Name, PrivateZone=None, region=None, key=None,
...
+            outlist += [char]
+        elif point in send_as_octal:
+            outlist += [octal]
+        elif point in acceptable_points_which_must_be_octalized:
+            outlist += [octal]
+        else:
+            raise SaltInvocationError("Invalid Route53 domain character seen (octal {0}) in string "
+                                      "{1}.  Do you need to punycode it?".format(octal, instring))
+    ret = ''.join(outlist)
+    log.debug('Name after massaging is: {}'.format(ret))
```
terminalmage: Please use % formatting here. Also, keep in mind that log messages should be distinctive. "Name after massaging" won't necessarily mean anything to someone looking at debug logging. My suggestion is to use something like:

```python
log.debug('AWS-encoded result for %s: %s', instring, ret)
```

salt/modules/boto3_route53.py
```diff
@@ -714,6 +714,74 @@ def delete_hosted_zone_by_domain(Name, PrivateZone=None, region=None, key=None,
...
+    return ret
+
+def _octalReplace(x):
+    '''
+    https://github.com/boto/boto/pull/1216
+    '''
+    c = int(x.group(1), 8)
+    if c > 0x20 and c < 0x2e or c > 0x2e and c < 0x7f:
```
terminalmage: This code needs to be better documented as to why the given hex ranges are being used.

terminalmage (changes requested):

nullify005:
Ok, so not withstanding the requested changes for the logging, I'm honestly at a loss as to how to convert extended ASCII characters which have the potential to come back from AWS as ordinal.

Specifically I was attempting to cater for AWS's example of:

"For example, to create a hosted zone for exämple.com, you specify ex\344mple.com."

This is basically the same issue as _to_aws_encoding() which is present in the state which I cooped for the module too in that it just doesn't care of the int value is above 256 (ie. no extended ASCII) which chr() can cater for. I suppose that I could unichr() but that might not be a happy thing from a Python 3 point of view.

Either I need more time to research it (I'm not a full time python coder) or need a little bit of a hand on how to deal with this.

I can go back to the original hack of just converting '\052' to '*' but that doesn't really attend to the wider problem here

damon-atkins:
Suggest python inbuilt x.encode('idna') functions are used for the conversion. See https://github.com/saltstack/salt/pull/47652#issuecomment-391212430

nullify005:
I'm getting closer, still needs some testing though

nullify005:
Okies, so I believe I'm ready for review this time.

- as much as possible the module (I think) should be responsible for the encode / decode where appropriate, as such this has been stripped from the state everywhere it could be
- the module will return decoded strings for comparisons upstream
- the encode tries to do an ascii encode & failing that assumes that it's idna, rather than implement all the AWS cases they specify, we try to encode everything
- similarly the decode looks for the existence of escaped strings which indicate octal is coming back & if not then assumes idna
- we now support idna & therefore non-english language domain names which wasn't possible before

salt/modules/boto3_route53.py
```diff
@@ -717,6 +719,94 @@ def delete_hosted_zone_by_domain(Name, PrivateZone=None, region=None, key=None,
...
+    '''
+    ret = None
+    try:
+        x.encode('ascii')
+        ret = re.sub(r'\\x([a-f0-8]{2})',
+                      _hexReplace, x.encode('unicode_escape'))
+    except UnicodeEncodeError:
+        ret = x.encode('idna')
+    except Exception as e:
+        log.error("Couldn't encode %s using either 'unicode_escape' or 'idna' codecs" % (x))
```
terminalmage: You don't need to manually perform the % substitution, you can simply pass `x` as a positional argument directly after the log message. The substitution happens within our logging handler.

terminalmage (changes requested):

salt/modules/boto3_route53.py
```diff
@@ -719,24 +720,20 @@ def delete_hosted_zone_by_domain(Name, PrivateZone=None, region=None, key=None,
...
     return delete_hosted_zone(Id=Id, region=region, key=key, keyid=keyid, profile=profile)
 
+
 def aws_encode(x):
     '''
     An implementation of the encoding required to suport AWS's domain name
-    rules defined here:
-
-        .. __: http://docs.aws.amazon.com/Route53/latest/DeveloperGuide/DomainNameFormat.html
+    rules defined `here <http://docs.aws.amazon.com/Route53/latest/DeveloperGuide/DomainNameFormat.html>`_.
```
terminalmage: Sphinx will not permit links defined using the format `` `link text <url>`_`` when there are more than one link in the same file with the same link text but different target URLs. Instead, you can do:

```
rules defined here__:

.. __: http://docs.aws.amazon.com/Route53/latest/DeveloperGuide/DomainNameFormat.html
```

terminalmage (changes requested):

salt/modules/boto3_route53.py
nullify005: Sure thing, sorry for all the iterations here

salt/modules/boto3_route53.py
```diff
@@ -718,6 +721,91 @@ def delete_hosted_zone_by_domain(Name, PrivateZone=None, region=None, key=None,
...
+
+    .. __: http://docs.aws.amazon.com/Route53/latest/DeveloperGuide/DomainNameFormat.html
+
+    While AWS's documentation specifies individual ASCII characters which need
+    to be encoded, we instead just try to force the string to one of
+    escaped unicode or idna depending on whether there are non-ASCII characters
+    present.
+
+    This means that we support things like ドメイン.テスト as a domain name string
+    per the idna documentation `here <https://pypi.org/project/idna>`_, in
```
terminalmage: Sorry, I thought my last review was clear enough to let you know to update the other similar "here" links. This will also need to be changed like in your last commit.

salt/modules/boto3_route53.py
```diff
@@ -718,6 +721,91 @@ def delete_hosted_zone_by_domain(Name, PrivateZone=None, region=None, key=None,
...
+        if 'AliasTarget' in o['Changes'][change_idx]['ResourceRecordSet']:
+            o['Changes'][change_idx]['ResourceRecordSet']['AliasTarget']['DNSName'] = aws_encode(o['Changes'][change_idx]['ResourceRecordSet']['AliasTarget']['DNSName'])
+        change_idx += 1
+    return o
+
+
+def _aws_decode(x):
+    '''
+    An implementation of the decoding required to suport AWS's domain name
+    rules defined `here <http://docs.aws.amazon.com/Route53/latest/DeveloperGuide/DomainNameFormat.html>`_.
```
terminalmage: This one too.

terminalmage (changes requested):

terminalmage (approved):

rallytime:
Thanks for your work on this @nullify005!

PR #47638, opened 2018-05-14
Collect all versions of installed packages on SUSE and RHEL systems

meaksh:
### What does this PR do?
At the moment, Salt `pkg.info_installed` method (for zypper & yum modules) only returns one version information per installed package name:

```
$ rpm -qa libgcc
libgcc-4.4.7-18.el6_9.2.i686
libgcc-4.4.7-18.el6_9.2.x86_64
```

Now we ask Salt for information of the installed `libgcc` package:

```
$ salt-call pkg.info_installed libgcc attr=version,arch
local:
    ----------
    libgcc:
        ----------
        arch:
            i686
        version:
            4.4.7
```

In case of multiple installed versions of the same package name, Salt is currently returning the latest installed package.

This PR also fixes an issue with the `pkg` state on SUSE systems when targeting packages like `glibc.x86_64`. Zypper actually install/remove the packages but the comparison made by `pkg` state is setting the packages as `failed` because zypper response does not include the "arch" as part of the package name.

### Previous Behavior
No way to get information for all package versions installed on the minion.

### New Behavior
This PR introduces the `all_versions` parameter into `zypper.info_installed`, `yum.info_installed` and also to `rpm.info` methods which is set to `False` by default. No changes on the returned structure.

Only when using `all_versions=True` it would change the output structure to produce a list of package information which actually reflects all installed versions for `libgcc`:

```
$ salt-call pkg.info_installed libgcc attr=version,arch all_versions=True
local:
    ----------
    libgcc:
        |_
          ----------
          arch:
              i686
          edition:
              4.4.7-18.el6_9.2
          version:
              4.4.7
        |_
          ----------
          arch:
              x86_64
          edition:
              4.4.7-18.el6_9.2
          version:
              4.4.7
```

### Tests written?

Yes

### Commits signed with GPG?

Yes

PR #47678, opened 2018-05-16
[develop] Collect all versions of installed packages on SUSE and RHEL systems

meaksh:
### What does this PR do?
This PR targets `develop` branch as discussed on https://github.com/saltstack/salt/pull/47638 

@terminalmage - I've also addressed your comment about using `ret.setdefault()` instead `if-else` statement. Thanks

<hr>

At the moment, Salt `pkg.info_installed` method (for zypper & yum modules) only returns one version information per installed package name:

```
$ rpm -qa libgcc
libgcc-4.4.7-18.el6_9.2.i686
libgcc-4.4.7-18.el6_9.2.x86_64
```

Now we ask Salt for information of the installed `libgcc` package:

```
$ salt-call pkg.info_installed libgcc attr=version,arch
local:
    ----------
    libgcc:
        ----------
        arch:
            i686
        version:
            4.4.7
```

In case of multiple installed versions of the same package name, Salt is currently returning the latest installed package.

This PR also fixes an issue with the `pkg` state on SUSE systems when targeting packages like `glibc.x86_64`. Zypper actually install/remove the packages but the comparison made by `pkg` state is setting the packages as `failed` because zypper response does not include the "arch" as part of the package name.

### Previous Behavior
No way to get information for all package versions installed on the minion.

### New Behavior
This PR introduces the `all_versions` parameter into `zypper.info_installed`, `yum.info_installed` and also to `rpm.info` methods which is set to `False` by default. No changes on the returned structure.

Only when using `all_versions=True` it would change the output structure to produce a list of package information which actually reflects all installed versions for `libgcc`:

```
$ salt-call pkg.info_installed libgcc attr=version,arch all_versions=True
local:
    ----------
    libgcc:
        |_
          ----------
          arch:
              i686
          edition:
              4.4.7-18.el6_9.2
          version:
              4.4.7
        |_
          ----------
          arch:
              x86_64
          edition:
              4.4.7-18.el6_9.2
          version:
              4.4.7
```

### Tests written?

Yes

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/modules/yumpkg.py
+++ b/salt/modules/yumpkg.py
@@ -1040,11 +1040,8 @@ def info_installed(*names, **kwargs):
                     t_nfo[key] = value
             if not all_versions:
                 ret[pkg_name] = t_nfo
-            elif not pkg_name in ret:
-                ret[pkg_name] = [t_nfo]
             else:
-                ret[pkg_name].append(t_nfo)
-
+                ret.setdefault(pkg_name, []).append(t_nfo)
     return ret
 
 
--- a/salt/modules/yumpkg.py
+++ b/salt/modules/yumpkg.py
@@ -1009,7 +1009,7 @@ def list_downloaded():
     return ret
 
 
-def info_installed(*names):
+def info_installed(*names, **kwargs):
     '''
     .. versionadded:: 2015.8.1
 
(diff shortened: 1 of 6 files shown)


terminalmage (approved):

isbm (changes requested):
@meaksh but this is different than [RFC you've agreed on](https://github.com/isbm/salt/blob/9f2835e20e91d56997becdcc6742b074c18dad3a/rfcs/modules/0001-pkg:list_pkgs-output.md)

meaksh:
@isbm - this PR is actually fixing the `pkg.info_installed` output, there is no changes on `pkg.list_pkgs` format. I didn't even change the default behavior for `pkg.info_installed` but just allowing multiple versions based on a parameter.

isbm (approved):
@meaksh OK. But we still need to go back to the RFC and think of/implement it. The 2018.3 is already released.

meaksh:
@isbm :+1:

damon-atkins:
OK this changes the standard for all platforms.

damon-atkins:
Link to https://github.com/saltstack/salt/pull/45948

salt/modules/yumpkg.py
```diff
@@ -1957,7 +1965,24 @@ def remove(name=None, pkgs=None, **kwargs):  # pylint: disable=W0613
...
 
     old = list_pkgs()
-    targets = [x for x in pkg_params if x in old]
+    targets = []
+    for target in pkg_params:
+        # Check if package version set to be removed is actually installed:
+        # old[target] contains a comma-separated list of installed versions
+        if target in old and not pkg_params[target]:
+            targets.append(target)
+        elif target in old and pkg_params[target] in old[target].split(','):
```
damon-atkins: call pkgs_list with versions_as_list=True    then no need to split

meaksh: If I add `versions_as_list=True` then it would also affect to the structure of returning `new` and `old` attributes of the `changes` dict as it would be a list of string instead of a comma-separated string. I didn't want to change that behavior here.

PR #47700, opened 2018-05-17
fix roots modification time check

yannj-fr:
### What does this PR do?
Fix cache modification time comparison. When converting the modification from os.path.getmtime to string, the number is stripped and the comparison is always false. the hash is recomputed everytime
### Previous Behavior
hash was compute everytime like if the cache was not present
### New Behavior
cache is used

### Tests written?

No

### Commits signed with GPG?

No

### DIFF
--- a/salt/fileserver/roots.py
+++ b/salt/fileserver/roots.py
@@ -255,7 +255,7 @@ def file_hash(load, fnd):
                     except OSError:
                         pass
                     return file_hash(load, fnd)
-                if os.path.getmtime(path) == mtime:
+                if "{}".format(os.path.getmtime(path)) == mtime:
                     # check if mtime changed
                     ret['hsum'] = hsum
                     return ret
--- a/salt/fileserver/roots.py
+++ b/salt/fileserver/roots.py
@@ -255,7 +255,7 @@ def file_hash(load, fnd):
                     except OSError:
                         pass
                     return file_hash(load, fnd)
-                if os.path.getmtime(path) == mtime:
+                if str(os.path.getmtime(path)) == mtime:
                     # check if mtime changed
                     ret['hsum'] = hsum
                     return ret


salt/fileserver/roots.py
```diff
@@ -255,7 +255,7 @@ def file_hash(load, fnd):
                     except OSError:
                         pass
                     return file_hash(load, fnd)
-                if os.path.getmtime(path) == mtime:
+                if "{}".format(os.path.getmtime(path)) == mtime:
```
terminalmage: Why not just use `str()`?

yannj-fr:
The mtime is created using format for the cache file , I prefer to use the exact same method in case implementation varies.

salt/fileserver/roots.py
yannj-fr: The mtime is created using format for the cache file , I prefer to use the exact same method in case implementation varies.

terminalmage: That's incorrect. The mtime is read from the cache file: https://github.com/saltstack/salt/pull/47700/files#diff-ea643453ed766af4296bffc374c40c0dR249

yannj-fr: Sorry my explanation was not clear enough.
The mtime is read from the cache.
but:
- when the mtime was written in the file it used this line:
```
   cache_object = '{0}:{1}'.format(ret['hsum'], os.path.getmtime(path))
    with salt.utils.flopen(cache_path, 'w') as fp_:
        fp_.write(cache_object)
    return ret
```

Because format will convert a float into string with only 2 decimals, like this:

```
>>> import os 
>>> os.path.getmtime('/dev/null')
>>> ts = os.path.getmtime('/dev/null')
>>> ts
1526395491.048
>>> "{}".format(ts)
'1526395491.05'

```
the timestamp cached in the cache file will be rounded to 2 decimals, and this will lead to a not matching comparison with the original line:

` if os.path.getmtime(path) == mtime:`

You can check this behaviour with a simple test:

```
>>> os.path.getmtime('/dev/null') == "{}".format(os.path.getmtime('/dev/null'))
False

```
The solution in to format the timestamp the same way it was formated when cached. my choice here is to you the format method as well as in the code I quoted in the beginning of this message.
str() may give exactly the same result, but I feel safer in using the exact same method than the one used to write the cache file

terminalmage: Thanks for the explanation. However, as noted in the [Python docs](https://docs.python.org/3/library/string.html#formatspec) an empty format specifier is equivalent to calling `str()` on the value.

yannj-fr: do you want me to use str() instead?

terminalmage: Yes, that would avoid the unnecessary overhead of allocating a string and invoking its `.format()` method.

terminalmage: Yes, that would avoid the unnecessary overhead of allocating a string and invoking its `.format()` method.

yannj-fr: just changed the PR with str()

rallytime (approved):

Issue #47484, opened 2018-05-04
Windows: pkg.latest state not updating packages.


whytewolf:
### Description of Issue/Question
This affects both py2 and py3

Installed Firefox 42.0, and used a state with pkg.latest to attempt to upgrade to the latest release. 

state returns 

```
winion-1:
----------
          ID: update-firefox
    Function: pkg.latest
        Name: firefox
      Result: False
     Comment: Package firefox failed to update.
     Started: 21:17:51.152000
    Duration: 19071.0 ms
     Changes:

Summary for winion-1
------------
Succeeded: 0
Failed:    1
------------
Total states run:     1
Total run time:  19.071 s
```

### Setup

Windows minion running 2018.3.0 with a state with pkg.latest and a package that needs updating. 

```
[root@master salt-winrepo-ng]# salt -G 'os:windows' pkg.list_pkgs
winion-1:
    ----------
    Microsoft Silverlight:
        5.1.50907.0
    Mozilla Maintenance Service:
        42.0
    Oracle VM VirtualBox Guest Additions 5.1.22:
        5.1.22.0
    firefox:
        42.0
    npp:
        7.5.6
    salt-minion:
        2018.3.0
```

```
[root@master salt-winrepo-ng]# salt -G 'os:windows' pkg.latest_version firefox
winion-1:
    59.0.3
```

```
[root@master salt-winrepo-ng]# cat /srv/salt/update.sls
update-firefox:
  pkg.latest:
    - name: firefox
```

### Steps to Reproduce Issue
See section above for setup. 

in the debug log line was all the rendering of the win_repo pacakges
afterwords was this 

```
[DEBUG   ] Latest installed version of firefox is 42.0
[DEBUG   ] Latest available version of package firefox is 59.0.3
[DEBUG   ] Upgrade of firefox from 42.0 to 59.0.3 is available
[DEBUG   ] Using existing pkg metadata db for saltenv 'base' (age is 0:00:00.047126)
[DEBUG   ] Could not LazyLoad pkg.normalize_name: 'pkg.normalize_name' is not available.
[DEBUG   ] Using existing pkg metadata db for saltenv 'base' (age is 0:00:00.062777)
[DEBUG   ] A version (None) already installed for package firefox
[DEBUG   ] Using existing pkg metadata db for saltenv 'base' (age is 0:00:00.078435)
[ERROR   ] Package firefox failed to update.
[...]

PR #47702, opened 2018-05-17
State pkg.latest called win pkg.install with list of pkgs and the required versions

damon-atkins:
### What does this PR do?
If the software is already installed windows pkg.install does not upgrade the software, as design.
However the state pkg.latest expected this behaviour.

The state pkg.latest determines already the software which needs to be updated and the version required. However does not use the version information.

This fix applied only to windows passes on this information (pkg + version) to pkg.installed.

Windows: pkg.latest state not updating packages. (#47484)

### Previous Behavior
The state pkg.latest did not work unless the "Package Definition" used latest as the version number.

### New Behavior
Works with normal version number and "latest"

Given the way Windows pkg system works.  If pkg.latest is called and the current version of the software is not "Package Definition" the software may be downgraded to the latest version defined in the "Package Definition" 

### Tests written?
No New Tests

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/states/pkg.py
+++ b/salt/states/pkg.py
@@ -2449,15 +2449,14 @@ def latest(
                     'result': None,
                     'comment': '\n'.join(comments)}
 
-
         # No need to refresh, if a refresh was necessary it would have been
         # performed above when pkg.latest_version was run.
         try:
             if salt.utils.is_windows():
                 # pkg.install execution module on windows ensures the
                 # software package is installed when no version is
-                # specified, it does not upgrade the software to the 
-                # latest. This is per the design. 
+                # specified, it does not upgrade the software to the
+                # latest. This is per the design.
                 # Build updated list of pkgs *with verion number*, exclude
                 # non-targeted ones
                 targeted_pkgs = [{x: targets[x]} for x in targets.keys()]
--- a/salt/states/pkg.py
+++ b/salt/states/pkg.py
@@ -2449,18 +2449,33 @@ def latest(
                     'result': None,
                     'comment': '\n'.join(comments)}
 
-        # Build updated list of pkgs to exclude non-targeted ones
-        targeted_pkgs = list(targets.keys()) if pkgs else None
 
+        # No need to refresh, if a refresh was necessary it would have been
+        # performed above when pkg.latest_version was run.
         try:
-            # No need to refresh, if a refresh was necessary it would have been
-            # performed above when pkg.latest_version was run.
-            changes = __salt__['pkg.install'](name,
-                                              refresh=False,
-                                              fromrepo=fromrepo,
-                                              skip_verify=skip_verify,
-                                              pkgs=targeted_pkgs,
-                                              **kwargs)
+            if salt.utils.is_windows():
+                # pkg.install execution module on windows ensures the
+                # software package is installed when no version is
+                # specified, it does not upgrade the software to the 
+                # latest. This is per the design. 
+                # Build updated list of pkgs *with verion number*, exclude
+                # non-targeted ones
+                targeted_pkgs = [{x: targets[x]} for x in targets.keys()]
+                changes = __salt__['pkg.install'](name=None,
+                                                  refresh=False,
+                                                  fromrepo=fromrepo,
+                                                  skip_verify=skip_verify,
+                                                  pkgs=targeted_pkgs,
+                                                  **kwargs)
+            else:
+                # Build updated list of pkgs to exclude non-targeted ones
+                targeted_pkgs = list(targets.keys()) if pkgs else None
+                changes = __salt__['pkg.install'](name,
+                                                  refresh=False,
+                                                  fromrepo=fromrepo,
+                                                  skip_verify=skip_verify,
+                                                  pkgs=targeted_pkgs,
+                                                  **kwargs)
         except CommandExecutionError as exc:
             return {'name': name,
                     'changes': {},
--- a/salt/states/pkg.py
+++ b/salt/states/pkg.py
@@ -2449,32 +2449,26 @@ def latest(
                     'result': None,
                     'comment': '\n'.join(comments)}
 
+        if salt.utils.is_windows():
+            # pkg.install execution module on windows ensures the software
+            # package is installed when no version is specified, it does not
+            # upgrade the software to the latest. This is per the design.
+            # Build updated list of pkgs *with verion number*, exclude
+            # non-targeted ones
+            targeted_pkgs = [{x: targets[x]} for x in targets.keys()]
+        else:
+            # Build updated list of pkgs to exclude non-targeted ones
+            targeted_pkgs = list(targets.keys()) if pkgs else None
+
         # No need to refresh, if a refresh was necessary it would have been
         # performed above when pkg.latest_version was run.
         try:
-            if salt.utils.is_windows():
-                # pkg.install execution module on windows ensures the
-                # software package is installed when no version is
-                # specified, it does not upgrade the software to the
-                # latest. This is per the design.
-                # Build updated list of pkgs *with verion number*, exclude
-                # non-targeted ones
-                targeted_pkgs = [{x: targets[x]} for x in targets.keys()]
-                changes = __salt__['pkg.install'](name=None,
-                                                  refresh=False,
-                                                  fromrepo=fromrepo,
-                                                  skip_verify=skip_verify,
-                                                  pkgs=targeted_pkgs,
-                                                  **kwargs)
-            else:
-                # Build updated list of pkgs to exclude non-targeted ones
-                targeted_pkgs = list(targets.keys()) if pkgs else None
-                changes = __salt__['pkg.install'](name,
-                                                  refresh=False,
-                                                  fromrepo=fromrepo,
-                                                  skip_verify=skip_verify,
-                                                  pkgs=targeted_pkgs,
-                                                  **kwargs)
+            changes = __salt__['pkg.install'](name=None,
+                                              refresh=False,
+                                              fromrepo=fromrepo,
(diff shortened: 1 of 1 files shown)


damon-atkins:
Results of testing, showing the software is now upgraded to last (left side)
```
#### pkg.install octopus-tentacle version=3.3.16 [pkg.install_octopus-tentacle_version=3.3.16_pkg_state_latest]
#### state.high {octopus-tentacle:{pkg:[latest]}} [state.high_{octopus-tentacle:{pkg:[latest]}}_pkg_state_latest]
[INFO    ] state.call Completed state [octopus-tentacle]     [INFO    ] state.call Completed state [octopus-tentacle]
at time 02:09:33.091000 duration_in_ms=37335.0             | at time 01:52:17.968000 duration_in_ms=34777.0
{                                                            {
    "local": {                                                   "local": {

"pkg_|-octopus-tentacle_|-octopus-tentacle_|-latest": {      "pkg_|-octopus-tentacle_|-octopus-tentacle_|-latest": {
            "comment": "The following packages were        |             "comment": "Package octopus-tentacle failed
successfully installed/upgraded: octopus-tentacle",        | to update.",
            "name": "octopus-tentacle",                                  "name": "octopus-tentacle",
            "start_time": "02:08:55.756000",               |             "start_time": "01:51:43.191000",
            "result": true,                                |             "result": false,
            "duration": 37335.0,                           |             "duration": 34777.0,
            "__run_num__": 0,                                            "__run_num__": 0,
            "__sls__": null,                                             "__sls__": null,
            "changes": {                                   |             "changes": {},
                "octopus-tentacle": {                      <
                    "new": "3.20.1",                       <
                    "old": "3.3.16"                        <
                }                                          <
            },                                             <
            "__id__": "octopus-tentacle"                                 "__id__": "octopus-tentacle"
        }                                                            }
    }                                                            }
}                                                            }
```

salt/states/pkg.py
```diff
@@ -2449,18 +2449,32 @@ def latest(
...
+                                                  **kwargs)
+            else:
+                # Build updated list of pkgs to exclude non-targeted ones
+                targeted_pkgs = list(targets.keys()) if pkgs else None
+                changes = __salt__['pkg.install'](name,
+                                                  refresh=False,
+                                                  fromrepo=fromrepo,
+                                                  skip_verify=skip_verify,
+                                                  pkgs=targeted_pkgs,
+                                                  **kwargs)
```
terminalmage: We should only need one call to `pkg.install` here. The only difference I see is that we are passing `name=None` on the Windows side, and `name` on non-Windows. But the pkg virtuals already ignore `name` when `pkgs` is not None, so we should safely be able to pass `name=None` on non-Windows platforms.

terminalmage (changes requested):

terminalmage (approved):

rallytime:
@damon-atkins This is causing 2 package tests to fail. Can you look into that please?

https://jenkins.saltstack.com/job/PR/job/salt-pr-linode-ubuntu16-py3/9968/

damon-atkins:
re-run ubuntu

damon-atkins:
https://jenkins.saltstack.com/job/PR/job/salt-pr-linode-cent7-py3/5024/
https://jenkins.saltstack.com/job/PR/job/salt-pr-rs-cent7-n/19081/
Seems to be fine now.

damon-atkins:
@terminalmage  Please have another look.

salt/states/pkg.py
```diff
@@ -2329,12 +2329,16 @@ def latest(
     '''
     refresh = salt.utils.pkg.check_refresh(__opts__, refresh)
 
+    if not name:
+        # Note state checks name is a string, this should only happen if their is a bug
+        raise CommandExecutionError('Parameter name must contain label or package name.')
+
```
terminalmage: I don't understand why you added this. It doesn't seem like it is necessary, for one thing, and it would result in an ugly traceback in the state results if it were ever triggered.

salt/states/pkg.py
```diff
@@ -2329,12 +2329,16 @@ def latest(
...
     if kwargs.get('sources'):
         return {'name': name,
                 'changes': {},
                 'result': False,
                 'comment': 'The "sources" parameter is not supported.'}
-    elif pkgs:
+    elif pkgs is not None:
```
terminalmage: Again, why would this be necessary? We shouldn't be invoking this state with an empty list as input, so why do we need to specifically check if it is not None?

salt/states/pkg.py
```diff
@@ -2344,15 +2348,7 @@ def latest(
...
-        if isinstance(pkgs, list) and len(pkgs) == 0:
-            return {
-                'name': name,
-                'changes': {},
-                'result': True,
-                'comment': 'No packages to install provided'
-            }
-        else:
-            desired_pkgs = [name]
+        desired_pkgs = [name]
```
terminalmage: Again, this doesn't make sense. Invoking the state with empty `pkgs` is not a supported use, we should be returning the error dict here.

terminalmage (changes requested):

salt/states/pkg.py
damon-atkins: It covers the following case
`salt-call state.high {'strace':{'pkg':['latest']}}`  where `pkgs` is `None`

damon-atkins:
`if pkgs is not None` covers off pkgs=0 which would result in a  "Invalidly formatted" message vs trying to upgrade what is contain in variable name.
I have restored it back to `if pkgs`
```
# salt-call state.single pkg.latest name=vim pkgs=0
[ERROR   ] pkg.latest latest - name: vim, pkgs: 0
[ERROR   ] pkg.latest No information found for 'vim'.
[ERROR   ] state.format_log No information found for 'vim'.
local:
----------
          ID: vim
    Function: pkg.latest
      Result: False
     Comment: No information found for 'vim'.
     Started: 04:30:07.624110
    Duration: 37397.384 ms
     Changes:

Summary for local
------------
Succeeded: 0
Failed:    1
------------
Total states run:     1
Total run time:  37.397 s
# salt-call state.single pkg.latest name=vim pkgs=strace
[ERROR   ] pkg.latest latest - name: vim, pkgs: strace
[ERROR   ] __init__.repack_dictlist Invalid input for repack_dictlist, data passed is not a list (strace)
[ERROR   ] state.format_log Invalidly formatted "pkgs" parameter. See minion log.
local:
----------
          ID: vim
    Function: pkg.latest
      Result: False
     Comment: Invalidly formatted "pkgs" parameter. See minion log.
     Started: 04:31:18.924653
    Duration: 10.632 ms
     Changes:

Summary for local
------------
Succeeded: 0
Failed:    1
------------
Total states run:     1
Total run time:  10.632 ms
# salt-call state.single pkg.latest name=vim
[ERROR   ] pkg.latest latest - name: vim, pkgs: None
[ERROR   ] pkg.latest No information found for 'vim'.
[ERROR   ] state.format_log No information found for 'vim'.
local:
----------
          ID: vim
    Function: pkg.latest
      Result: False
     Comment: No information found for 'vim'.
     Started: 04:33:15.291855
    Duration: 37435.706 ms
     Changes:

Summary for local
------------
Succeeded: 0
Failed:    1
------------
Total states run:     1
Total run time:  37.436 s
```

terminalmage (changes requested):
Please perform a `git reset --hard 83b4224c` and apply the following patch:

```diff
diff --git a/salt/states/pkg.py b/salt/states/pkg.py
index 37ac5e5966..f58453837b 100644
--- a/salt/states/pkg.py
+++ b/salt/states/pkg.py
@@ -2455,10 +2455,10 @@ def latest(
             # upgrade the software to the latest. This is per the design.
             # Build updated list of pkgs *with verion number*, exclude
             # non-targeted ones
-            targeted_pkgs = [{x: targets[x]} for x in targets.keys()]
+            targeted_pkgs = [{x: targets[x]} for x in targets]
         else:
             # Build updated list of pkgs to exclude non-targeted ones
-            targeted_pkgs = list(targets.keys()) if pkgs else None
+            targeted_pkgs = list(targets)
 
         # No need to refresh, if a refresh was necessary it would have been
         # performed above when pkg.latest_version was run.
```

terminalmage:
Passing `name=None` requires that `pkgs` has contents, which it should since we've pre-determined the install targets. So the fix was as simple as getting rid of the `if pkgs else None`.

Also, there is no need to call `.keys()` because iterating over a dictionary (or calling `list()` on one) will simply iterate over the keys. Running `list()` allocates a list first on Python 2 and then iterates over that list, rather than just iterating over the keys.

damon-atkins:
FYI:
```
        if isinstance(pkgs, list) and len(pkgs) == 0:
            return {
                'name': name,
                'changes': {},
                'result': True,
                'comment': 'No packages to install provided'
            }
```
I can not see how the above if will ever be true. That can be fixed in a different PR.

terminalmage:
That code is there for a reason. That `if` will evaluate as True in the case that someone intentionally passes invalid SLS:

```yaml
foo:
  pkg.installed:
    - pkgs: []
```

Yes, some people actually have done this in the past and complained that we didn't properly report on their misconfiguration.

terminalmage (approved):

damon-atkins:
Thanks

damon-atkins:
Failed with non related issues.
https://jenkins.saltstack.com/job/PR/job/salt-pr-linode-ubuntu16-py3/10027/
https://jenkins.saltstack.com/job/PR/job/salt-pr-rs-cent7-n/19114/

Passed
https://jenkins.saltstack.com/job/PR/job/salt-pr-linode-cent7-py3/5057/
https://jenkins.saltstack.com/job/PR/job/salt-pr-linode-ubuntu14-n/22991/

PR #47710, opened 2018-05-17
salt.loader: add error logging when whitelist lookup fails

terminalmage:
Rather than just raising a bare KeyError with no additional information, this raises a proper KeyError with the key that failed to be looked up. It also logs the key that failed to load, as well as the whitelist, to aid in troubleshooting.

Additionally, this resolves failed lazydict lookups in the tests due to improperly-set whitelists.

### DIFF
--- a/salt/loader.py
+++ b/salt/loader.py
@@ -1629,7 +1629,11 @@ class LazyLoader(salt.utils.lazy.LazyDict):
                 return True
             # if the modulename isn't in the whitelist, don't bother
             if self.whitelist and mod_name not in self.whitelist:
-                raise KeyError
+                log.error(
+                    'Failed to load key %s because its module (%s) is not in '
+                    'the whitelist: %s', key, mod_name, self.whitelist
+                )
+                raise KeyError(key)
 
             def _inner_load(mod_name):
                 for name in self._iter_files(mod_name):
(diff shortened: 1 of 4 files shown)


salt/loader.py
```diff
@@ -1629,7 +1629,11 @@ def _load(self, key):
                 return True
             # if the modulename isn't in the whitelist, don't bother
             if self.whitelist and mod_name not in self.whitelist:
-                raise KeyError
+                log.error(
+                    'Failed to load key %s because its module (%s) is not in '
```
isbm: If someone reading the log won't start panicking about SSH pr SSL keys... Can we rephrase this to something else?

isbm (changes requested):

rallytime:
@terminalmage The dockermod tests are failing on this change. Can you take a look?

https://jenkins.saltstack.com/job/PR/job/salt-pr-linode-cent7-py3/5009/

salt/loader.py
terminalmage: Sure, that is a good point.

terminalmage: Done.

terminalmage:
@rallytime Should be fixed now. I though I ran all the dockermod unit tests after editing the whitelist, but it looks like I only ran a single test.

cachedout (approved):

rallytime (approved):

PR #47721, opened 2018-05-18
Update yumpkg.py to include update(-minimal)

solhuebner:
### What does this PR do?
Added functions to the yum/dnf pkg module to offer an update or update-minimal as upgrade is forcing the removal of obsolete packages and this might not be desired.

### New Behavior
new function pkg.update (for yum/dnf)
new function pkg.update_minimal (for yum/dnf)
new option pkg.update update_minimal=True (for yum/dnf)

### Tests written?
No

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/modules/yumpkg.py
+++ b/salt/modules/yumpkg.py
@@ -1742,6 +1742,269 @@ def upgrade(name=None,
 
     return ret
 
+def update_minimal(name=None,
+            pkgs=None,
+            refresh=True,
+            skip_verify=False,
+            normalize=True,
+            **kwargs):
+    '''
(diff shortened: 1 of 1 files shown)


solhuebner:
If this would be merged I will open a separate PR for the 2018.3.1 branch as there are quite a few changes in the involved modules and it would not be a simple copy and paste job :)

cachedout:
Hi @solhuebner 

First, welcome to the Salt community! We're glad you're here.

We only accept new features into the `develop` branch of Salt, however. Could you please close this PR and re-open it against `develop`? We'll review it from there. Thanks!

solhuebner:
Thanks. Will do :)

PR #47728, opened 2018-05-18
Update yumpkg.py to include update(-minimal)

solhuebner:
### What does this PR do?
Added functions to the yum/dnf pkg module to offer an update or update-minimal as upgrade is forcing the removal of obsolete packages and this might not be desired.

### New Behavior
new function pkg.update (for yum/dnf)
new function pkg.update_minimal (for yum/dnf)
new option pkg.update update_minimal=True (for yum/dnf)

### Tests written?
No

### Commits signed with GPG?
Yes


solhuebner:
Rebased it to devel (#47721)

isbm (approved):

terminalmage (changes requested):
A lot of this new code is redundant. We already have a function called `upgrade` which performs a full system upgrade.

Also, yum supports `upgrade` and `upgrade-minimal` as aliases to `update` and `update-minimal`.

This new functionality can be implemented by simply adding a `minimal=False` argument to the `upgrade` function (right after `normalize` [here](https://github.com/saltstack/salt/blob/d909665/salt/modules/yumpkg.py#L1750)), and then by changing [this line](https://github.com/saltstack/salt/blob/d909665/salt/modules/yumpkg.py#L1903) like so:

```python
cmd.append('upgrade' if not minimal else 'upgrade-minimal')
```

terminalmage:
themoreyouknow

So, thanks for correcting me on that, I had never seen how it worked and just assumed they did the same thing.

The issue is that we already use the `upgrade()` function for a given `pkg` module in the `pkg.uptodate` state. So renaming the function would change how this state works. Maybe the best thing to do would be:

1. Add the `minimal` argument to `upgrade()`.
2. Also add an `obsoletes` argument to `upgrade()` (defaulting to `True`).

Have the `update()` function be a one-line which calls `upgrade()` with `obsoletes=False`.

What do you think?

solhuebner:
@terminalmage great suggestions!
I implemented them and hope you like the changes :)

salt/modules/yumpkg.py
```diff
@@ -1769,7 +1771,8 @@ def upgrade(name=None,
     .. _`systemd-run(1)`: https://www.freedesktop.org/software/systemd/man/systemd-run.html
     .. _`systemd.kill(5)`: https://www.freedesktop.org/software/systemd/man/systemd.kill.html
 
-    Run a full system upgrade, a yum upgrade
+    .. versionchanged:: Fluorine
+        Added obsoletes and minimal arguments
```
terminalmage: For emphasis, can you surround "obsoletes" and "minimal" with double-backticks? This will make them stand out in our documentation.

```rst
Added ``obsoletes`` and ``minimal`` arguments 
```

salt/modules/yumpkg.py
```diff
@@ -1853,6 +1856,27 @@ def upgrade(name=None,
...
+        Goes to the 'newest' package match which fixes a problem that affects your system.
+
+        .. code-block:: bash
+
+            salt '*' pkg.upgrade minimal=True
+
+        .. versionadded:: Fluorine
+
+    obsoletes : True
+        Controls wether dnf/yum should take obsoletes into account and remove them.
```
terminalmage: Could you make this say "yum/dnf" instead? Just to be consistent with other similar references elsewhere in the docstrings.

salt/modules/yumpkg.py
```diff
@@ -1853,6 +1856,27 @@ def upgrade(name=None,
...
+        If set to False yum will use update instead of upgrade
+        and dnf will be run with --obsoletes=False
```
terminalmage: The above two lines would be a good place to add more emphasis, can you surround "update", "upgrade", and "--obsoletes=False" in double-backticks?

salt/modules/yumpkg.py
```diff
@@ -1919,6 +1953,28 @@ def upgrade(name=None,
...
+            refresh=True,
+            skip_verify=False,
+            normalize=True,
+            minimal=False,
+            obsoletes=False,
+            **kwargs):
+    '''
+    Uses "upgrade" but defaults to obsoletes=False
+    Introduced to mirror expected CLI usage of yum update
+    For options see "upgrade"
```
terminalmage: We can link to the docs for `upgrade()` to make things even better. Also, while we place `.. versionadded::` and `.. versionchanged::` inline to denote new arguments, for new _functions_ we tend to put this at the top before the function description, so that it appears right next to the function name in the docs. So, how about something like this.

```rst
    .. versionadded:: Fluorine

    Calls :py:func:`pkg.upgrade <salt.modules.yumpkg.upgrade>` with
    ``obsoletes=False``. Mirrors the CLI behavior of ``yum update``. See
    :py:func:`pkg.upgrade <salt.modules.yumpkg.upgrade>` for further
    documentation.
```

terminalmage (changes requested):
The code looks great!

There are just some minor tweaks we can make in the docstrings for consistency and emphasis. Could you take a look at these?

solhuebner:
@terminalmage good point with the link to the upgrade() doc. I will try to keep the formatting in mind for future submissions :) And thank you for your patience with me.

salt/modules/yumpkg.py
```diff
@@ -1919,6 +1953,29 @@ def upgrade(name=None,
...
+            skip_verify=False,
+            normalize=True,
+            minimal=False,
+            obsoletes=False,
+            **kwargs):
+    '''
+    .. versionadded:: Fluorine
+
+    Calls :py:func:`pkg.upgrade <salt.modules.yumpkg.upgrade>` with
+    ``obsoletes=False``. Mirrors the CLI behavior of ``yum update``.See
```
terminalmage: I hate to nitpick, but can you put a space before "See"?

terminalmage (changes requested):
Just _one_ more change, and this will be ready to merge (pending test results).

terminalmage (approved):
:+1: Excellent. Now we wait for tests :smile:

solhuebner:
Is there a chance that this PR will get backported or are "additions" only for future releases?

terminalmage:
@solhuebner Feature additions like this are not backported. You can however copy yumpkg.py into `salt://_modules/` and run [saltutil.sync_modules](https://docs.saltstack.com/en/latest/ref/modules/all/salt.modules.saltutil.html#salt.modules.saltutil.sync_modules), and that yumpkg.py will override the one we distribute with Salt. You'd just need to remember to remove the module and re-sync once you upgrade to the next feature release (i.e. Fluorine).

Issue #35523, opened 2016-08-17
Come up with a reasonable alternative for lxc.edited_conf


rallytime:
### Description of Issue/Question

The `edited_conf` function in the lxc state module is not idempotent and makes changes every time it is run. The function is marked as deprecated as of 2015.5.0 and has a huge warning in the docs and the deprecation message about this problem. However, we haven't come up with a reasonable alternative to replace this function yet.

This issue is to track the need to replace this `edited_conf` so it can be properly removed in a  future feature release.

ping @terminalmage

PR #47731, opened 2018-05-18
Remove deprecation warning until #35523 is resolved

rallytime:
Refs #35523

I discussed this issue with @terminalmage offline, but basically this function was marked for deprecation a very long time ago and we cannot fully remove the function and complete the deprecation process until issue #35523 is resolved.

Instead of just continually bumping the deprecation version, the warning has been removed from the code, but the warning about the function's bug remains in the documentation. Once we have more resources to dedicate to this issue, the documentation warning can suffice.

### DIFF
--- a/salt/states/lxc.py
+++ b/salt/states/lxc.py
@@ -8,7 +8,6 @@ from __future__ import absolute_import, print_function, unicode_literals
 __docformat__ = 'restructuredtext en'
 
 # Import salt libs
-import salt.utils.versions
 from salt.exceptions import CommandExecutionError, SaltInvocationError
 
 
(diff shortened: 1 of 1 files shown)


salt/states/lxc.py
```diff
@@ -681,8 +681,10 @@ def edited_conf(name, lxc_conf=None, lxc_conf_unset=None):
         This state is unsuitable for setting parameters that appear more than
         once in an LXC config file, or parameters which must appear in a
         certain order (such as when configuring more than one network
-        interface). It is slated to be replaced, and as of version 2015.5.0 it
-        is deprecated.
+        interface).
+
+        Issue #35523 was opened to track the addition of a suitable replacement
```
terminalmage: You can link to this issue by making an RST link like so:

```
`Issue #35523`_ was opened to track the addition of a suitable replacement
or fix

.. _`Issue #35523`: https://github.com/saltstack/salt/issues/35523
```

terminalmage (changes requested):

salt/states/lxc.py
rallytime: Done.

terminalmage (approved):

Issue #47689, opened 2018-05-16
Poor performance of pip.installed when given a list of packages


OrlandoArcapix:
### Description of Issue/Question

When using pip.installed to manage the presence of a list of packages, the performance is quite poor, with each package in the list adding ~1-2s to the runtime of the state.

This is due to the re-execution of "pip freeze" for each package, which is an expensive operation. It could potentially be avoided by running "pip freeze" once and storing the output for re-use, refreshing it only as required.

### Setup

Start from a CentOS7 system. Install pip and salt 2017.7.

Note - the issue is also present in 2018.3 and dev.

```
yum -y install python2-pip
./bootstrap.sh stable 2017.7
mkdir -p /srv/salt
```

Set up a pip state to install the current python pip packages:
```
cat > /srv/salt/pip.sls << EOF
pip packages:
  pip.installed:
    - pkgs:
EOF

pip freeze  | gawk '{print "      - "$1}' >> /srv/salt/pip.sls 
```

### Steps to Reproduce Issue

Apply the state - note that "pip freeze" is called once for each package.

```
salt-call --local state.apply pip test=true --log-level=info
```
Run time is about 26s:
```
    Duration: 26203.069 ms
```

Example patch that shows an improvement is included below, as well as debug level logs. With that change, run time is about 1.2s:
```
    Duration: 1265.095 ms
```

### Versions Report
```
# salt-call --versions-report
Salt Version:
           Salt: 2017.7.5

Dependency Versions:
           cffi: Not Installed
       cherrypy: Not Installed
       dateutil: Not Installed
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.7.2
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.5.1
   mysql-python: Not Installed
      pycparser: Not Installed
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: Not Installed
         Python: 2.7.5 (default, Apr 11 2018, 07:36:10)
   python-gnupg: Not Installed
         PyYAML: 3.11
          PyZMQ: 15.3.0
           RAET: Not Installed
          smmap: Not Installed
        timelib: Not Installed
        Tornado: 4.2.1
            ZMQ: 4.1.4

System Versions:
           dist: centos 7.5.1804 Core
[...]

PR #47734, opened 2018-05-18
#47689 improve run-speed of pip package state 

Orlando-Richards:
### What does this PR do?

Only load the current package list once when checking multiple packages, rather than once for each package.

#47689 

### Previous Behavior

pip freeze is executed once for each package in a list when checking multiple pip packages.

### New Behavior

pip freeze is only executed once, and the result is re-used for each package check.

### Tests written?

No

### Commits signed with GPG?

No


gtmanfred:
There are some lint failures that will need to be cleaned up https://jenkins.saltstack.com/job/PR/job/salt-pr-docs-n/17344/violations/

OrlandoArcapix:
Lint errors sorted

gtmanfred (approved):

salt/states/pip_state.py
```diff
@@ -186,16 +186,19 @@ def _check_pkg_version_format(pkg):
 
 def _check_if_installed(prefix, state_pkg_name, version_spec, ignore_installed,
                         force_reinstall, upgrade, user, cwd, bin_env, env_vars,
-                        **kwargs):
+                        pip_list=False, **kwargs):
```
cachedout: IMHO we need to have some documentation about what this kwarg does and the type of data expected here.

cachedout (changes requested):
Some docs, please

salt/states/pip_state.py
OrlandoArcapix: I was surprised by that kwargs too! It wasn't in the release code I initially patched on my systems, but it was there when I pulled the 2017.7 branch to do the pull request against.

Looks like it came from here: https://github.com/saltstack/salt/commit/7c46d9d0d457a686afb12860a655c0ecc15e427b

rallytime: @OrlandoArcapix I think we just want some documentation for your new kwarg, `pip_list`. I know the terminology was a little confusing there. :)

OrlandoArcapix:
Ahh, I see @rallytime ! 

Not sure where to put that - the extra argument I've added is only used in an internal function, and it's arguments are not currently documented anywhere that I can see - can you point me to the right place to add the extra argument info?

```
def _check_if_installed(prefix, state_pkg_name, version_spec, ignore_installed,
                        force_reinstall, upgrade, user, cwd, bin_env, env_vars,
                        pip_list=False, **kwargs):
    # result: None means the command failed to run
    # result: True means the package is installed
    # result: False means the package is not installed
```

rallytime:
@OrlandoArcapix Ah, you're right. I misread the code and thought it was an exposed variable, but it is not. I suppose it would be nice to have it documented somewhere maybe inline what the point of the variable is? I know none of the other kwargs are documented in `_check_if_installed`, but adding this would be a good place to start.

You can either do a doc string, or just a comment. The doc string won't populate the API documentation, but at least there will be something in the code to reference.

OrlandoArcapix:
Docblock added with new kwarg described!

rallytime:
Thanks @OrlandoArcapix! There are some lint errors on this now: https://jenkins.saltstack.com/job/PR/job/salt-pr-lint-n/22663/violations/file/salt/states/pip_state.py/

Can you fix those up?

OrlandoArcapix:
Oops - sorry! All sorted.

rallytime:
@OrlandoArcapix This change is failing the following test:

- integration.states.test_pip_state.PipStateTest.test_46127_pip_env_vars

https://jenkins.saltstack.com/job/PR/job/salt-pr-linode-ubuntu14-n/23706/

Can you take a look?

rallytime:
Hi @OrlandoArcapix - Just a bump here. Were you able to take a look at that failing test?

OrlandoArcapix:
Hi @rallytime, sorry, I've not had much of a chance to dig into it. I had a brief look at first, and I thought it might just need a change to the string checks in the tests. I'd worry that there was something more fundamental though, in case the test really is flagging a bad behaviour - I'm not familiar with the pip module code to know for sure that my quick fix was sensible!

Having said that, I've been using it in production now for hundreds of system builds and it hasn't let me down yet!

I can have a shot at changing the test's string check if you think that's a good way to proceed  - let me know.

gtmanfred:
As long as the test still is valid, that should be fine to change the test string.

OrlandoArcapix:
I've updated the test to match the string currently returned. 

For what it's worth, I have been using it for a couple of months now, across a lot of different pip installation situations, and it's been working fine - but I wouldn't say no to someone with knowledge of the code workings doing a review of the change!

rallytime (approved):

PR #47741, opened 2018-05-20
Add module for executing yum-complete-transaction

### LINKED ISSUE #47685: yum-complete-transaction module/state
It may be helpful for some to be able to call `yum-complete-transaction` as a module/state, instead of directly from `cmd.run`. The available option is `--cleanup-only`, and it may be useful to allow for it to be called recursively via the config.

### Setup
Feature request, would require a system using yum and with `yum-complete-transaction` (`yum-utils`) installed.

### Steps to Reproduce Issue
None, feature request

### Versions Report
Not available in 2018.3.1

Discussion/Question: Should this be implemented under the virtual package `salt.modules.pkg` (I'm not sure how many other package management systems have a similar functionality) or only by directly addressing `salt.modules.yumpkg` directly or ___?

leeclemens:
### What does this PR do?

Adds a module ~~and state~~ for executing `yum-complete-transaction`

Fixes #47685

### New Behavior
Adds a module `pkg.complete_transaction` ~~and state `pkg.complete_transaction`~~, which executes `yum-complete-transaction`.

### Tests written?

No

### Commits signed with GPG?

Yes


leeclemens:
Sample output from `state`:
```
minion1:
----------
          ID: foo-complete-transaction-recursive
    Function: pkg.complete_transaction
      Result: True
     Comment: Loaded plugins: fastestmirror
              Loading mirror speeds from cached hostfile
               * base: mirrors.abc.com
               * epel: mirrors.xyz.com
               * extras: mirrors.extraextra.com
               * updates: mirrors.updates.com
              No unfinished transactions left.
     Started: 17:15:03.540321
    Duration: 862.494 ms
     Changes:   

Summary for minion1
------------
Succeeded: 1
Failed:    0
------------
Total states run:     1
Total run time: 862.494 ms
```

I was thinking of including completed transactions details in `changes`, but it may involve a bit more string parsing - so I just stuffed all of the output in `comment` for now.

terminalmage (changes requested):
The state is not idempotent. A well-written state should follow the following logic:

1. Check if changes need to be made.

   This is typically done via some sort of information-gathering function, preferably one which would be useful outside of the context of the state. In this context, such a function would list any incomplete transactions. When run from the state, an empty return would let the state know there are no unfinished transactions.

2. If no changes need to be made, return from the state with a `True` result and a proper comment explaining that nothing needed to be done.

   In this state, such a comment would be something like. "No incomplete transactions found".

3. If changes _do_ need to be made, do an `__opts__['test']` check to ensure that we A) mention what changes would be made, and B) do not actually proceed to the point where we make those changes.

   This has already been done here. :+1: 

4. Make the needed changes.

   This also looks to have been done here.

5. Perform a repeat of the check from step 1. If it shows that changes still need to be made, then the result is `False`, otherwise it is True.

6. Prepare and return the return dict. Set the proper `comment`, `result`, and also set the `changes` key if any changes were made.

It looks like you're just setting the comment to the return of the command. I've not seen the return from this command, so I don't know how good an idea that is. Also, you're not setting the `changes` at all, which is problematic.


Keep in mind also that not all remote-execution functions lend themselves to idempotent states. If you can't satisfy the above, that's OK! This is why we have the `module.run` state, which is designed to just always run a remote-exec function and not worry about idempotency.

terminalmage:
Yeah, I think it would make sense to remove the state for now. It can always be added later, and there is plenty of time for that as we are not planning to make a new feature release until this autumn.

salt/modules/yumpkg.py
```diff
@@ -3217,3 +3217,70 @@ def list_installed_patches():
...
+        python_shell=False
+    )
+    if recursive and 'No unfinished transactions left' not in ret:
+        if run_count >= max_attempts:
+            ret += '\nAttempt {0}/{1} exceeded max'.format(run_count, max_attempts)
+            log.warning('Attempt %s/%s exceeded max for command: `%s`', run_count, max_attempts, ' '.join(cmd))
+        else:
+            return _complete_transaction(cleanup_only, recursive, max_attempts, run_count + 1, ret)
+
+    return ret
```
isbm: I am not sure if this is a good idea to return string, deliberately losing the actual error code. This is all over the place for RH module though, which is even worse.

leeclemens:
I was looking for a similar recursive example, but I don't see (and can't readily test) where `zypper` would return 5 `pid'`s, for example. The latest commit returns a list of dictionaries (each representing one execution of the command), should it return a dictionary with lists for each key (e.g. `{'pid': [1, 2, 3], 'retcode': [0, 0, 0]}`)? Or squash the outputs and use the last execution's pid/retcode?

salt/modules/yumpkg.py
```diff
@@ -3217,3 +3217,76 @@ def list_installed_patches():
...
+
+    cmd_ret_list.append(__salt__['cmd.run_all'](
+        cmd,
+        output_loglevel='trace',
+        python_shell=False
+    ))
+
+    if cmd_ret_list[-1]['retcode'] == 0 \
+        and recursive \
+            and 'No unfinished transactions left.' not in cmd_ret_list[-1]['stdout']:
```
isbm: Please use `(` and `)` instead of `\` for [line continuation](http://docs.python-guide.org/en/latest/writing/style/#line-continuations).

salt/modules/yumpkg.py
```diff
@@ -3217,3 +3217,76 @@ def list_installed_patches():
...
+        if run_count >= max_attempts:
+            cmd_ret_list[-1]['retcode'] = 1
+            if cmd_ret_list[-1]['stderr']:
+                # this shouldn't happen, since retcode was 0 and we had no unfinished transactions
+                cmd_ret_list[-1]['stderr'] += '\n'
+            cmd_ret_list[-1]['stderr'] += 'Command succeeded but attempt {0}/{1} exceeded max.' \
+                                          ' Unfinished transactions remain.'.format(run_count, max_attempts)
```
isbm: Usually this is `comment` or `changes` section for `STDOUT` and `errors` for `STDERR`. And the [line continuation](http://docs.python-guide.org/en/latest/writing/style/#line-continuations) applies here too.

salt/modules/yumpkg.py
```diff
@@ -3217,3 +3217,76 @@ def list_installed_patches():
...
+
+    if cmd_ret_list[-1]['retcode'] == 0 \
+        and recursive \
+            and 'No unfinished transactions left.' not in cmd_ret_list[-1]['stdout']:
+        if run_count >= max_attempts:
+            cmd_ret_list[-1]['retcode'] = 1
+            if cmd_ret_list[-1]['stderr']:
+                # this shouldn't happen, since retcode was 0 and we had no unfinished transactions
+                cmd_ret_list[-1]['stderr'] += '\n'
+            cmd_ret_list[-1]['stderr'] += 'Command succeeded but attempt {0}/{1} exceeded max.' \
```
isbm: Could you rephrase this message?

isbm (changes requested):

isbm:
@terminalmage yes. :wink:

salt/modules/yumpkg.py
terminalmage: Yeah, `stderr` is usually reserved for the output of a command. What we should be doing here actually is raising a `CommandExecutionError`. This exception class provides for an optional `info` parameter, which will then have its results rendered in the return:

```python
>>> # assuming max_attempts == 1
...
>>> from salt.exceptions import CommandExecutionError
>>> cmd_ret_list = [{'pid': 12345, 'retcode': 0, 'stdout': 'Applied transaction', 'stderr': ''}, {'pid': 23456, 'retcode': 0, 'stdout': 'Applied transaction', 'stderr': ''}]
>>> try:
...     raise CommandExecutionError('Command succeeded but attempts exceeded max', info={'results': results})
... except CommandExecutionError as exc:
...     pass
...
>>> print(exc)
Command succeeded but attempts exceeded max. Additional info follows:

results:
    |_
      ----------
      pid:
          12345
      retcode:
          0
      stderr:
      stdout:
          Applied transaction
    |_
      ----------
      pid:
          23456
      retcode:
          0
      stderr:
      stdout:
          Applied transaction
>>> exc.info
{'results': [{'pid': 12345, 'retcode': 0, 'stderr': '', 'stdout': 'Applied transaction'}, {'pid': 23456, 'retcode': 0, 'stderr': '', 'stdout': 'Applied transaction'}]}
```
So the string that I printed there is what you'd see in the return from that minion, but you'd also have the ability to catch that exception and access the results. That's not something that would be necessary in this case, but it does leave that option open to any code which invokes this function.

terminalmage (changes requested):
I'd like to see us raising an exception instead of munging the stderr of the last command. Also, @isbm's line continuation suggestions are a good idea.

leeclemens:
Hi @rallytime yes, I have been a little sidetracked, but I agree with and can implement the recommendations. I ran into a potential issue though that I'm having trouble setting up test cases for, where `yum-complete-transaction` will require input (`yum -y`, but not sure if how best to pass `y` to it since it's executed indirectly).

rallytime:
Can you not pass that information through in a mock? I'm not sure what your test cases look like.

leeclemens:
I was referring to the actual execution. Previously I was testing where `yum-complete-transaction` did not require additional input, but I have since seen it require input (but have not been able to test that with the salt module specifically). I am happy to make the requested changes/improvements, but there may be real life cases where `yum-complete-transaction` will require input which could otherwise be overridden with `yum -y`, but I see no similar way (other than piping in `yes`) to auto-confirm with `yum-complete-transaction`.

cachedout:
@terminalmage and @isbm I think we might be ready for re-reviews here. Thanks.

terminalmage (approved):

salt/modules/yumpkg.py
```diff
@@ -3219,3 +3219,75 @@ def list_installed_patches():
...
+    if cleanup_only:
+        cmd.append('--cleanup-only')
+
+    cmd_ret_list.append(__salt__['cmd.run_all'](
+        cmd,
+        output_loglevel='trace',
+        python_shell=False
+    ))
+
+    if (cmd_ret_list[-1]['retcode'] == 0 and
```
isbm: Either do it implicit (not good, BTW):
```python
if (not cmd_ret_list[-1]['retcode'] and ...
```
...or explicit:
```python
if (cmd_ret_list[-1]['retcode'] == salt.defaults.exitcodes.EX_OK and ...
```

But please do not hard-code things just like that.

salt/modules/yumpkg.py
```diff
@@ -3219,3 +3219,75 @@ def list_installed_patches():
...
+            recursive and
+            'No unfinished transactions left.' not in cmd_ret_list[-1]['stdout']):
+        if run_count >= max_attempts:
+            cmd_ret_list[-1]['retcode'] = 1
```
isbm: Same here:
```python
cmd_ret_list[-1]['retcode'] == salt.defaults.exitcodes.EX_GENERIC
```

isbm (changes requested):
Sorry, I was on vacations. Here are some minor things left.

salt/modules/yumpkg.py
```diff
@@ -3219,3 +3219,75 @@ def list_installed_patches():
...
+            log.warning('Attempt %s/%s exceeded `max_attempts` for command: `%s`',
```
isbm: I guess, this is probably a `log.error`, right? Because you're raising an exception then.

leeclemens: Yes, good catch!

salt/modules/yumpkg.py
leeclemens: Was unaware of these variables, thank you!

salt/modules/yumpkg.py
```diff
@@ -3219,3 +3220,75 @@ def list_installed_patches():
...
+
+    cmd_ret_list.append(__salt__['cmd.run_all'](
+        cmd,
+        output_loglevel='trace',
+        python_shell=False
+    ))
+
+    if (cmd_ret_list[-1]['retcode'] == salt.defaults.exitcodes.EX_OK and
+            recursive and
+            'No unfinished transactions left.' not in cmd_ret_list[-1]['stdout']):
```
isbm: I think there is some problem with the indent, isn't it?

isbm (changes requested):
I think this is fine, except I believe some issue with the indent. While it works, it still could be fixed.

salt/modules/yumpkg.py
leeclemens: pep8 seems okay with it either way. Are you thinking the "r" in `recursive` should be under the "c" in `cmd_ret_list` and "'No unfinished" be left where it is?

leeclemens: I don't see anything overriding in Salt's guide, but pep8 states:
> This PEP takes no explicit position on how (or whether) to further visually distinguish such conditional lines from the nested suite inside the if-statement.

I'm happy to un-indent `recursive`, or simply drop it to a new line and leave `if (` and indent below that ("Vertical is better").

isbm: @leeclemens yeah, if that works, fine. Maybe github shows it that way strange, so I just thought that Linter didn't captured that. :wink: 

As of just sharing experiences, I personally would format "operator first", once I use parenthesis. That said, I would do:

```python
if (cmd_ret_list[-1]['retcode'] == salt.defaults.exitcodes.EX_OK
    and recursive and 'No unfinished transactions left.' not in cmd_ret_list[-1]['stdout']):
```
Or (probably better):
```python
if (cmd_ret_list[-1]['retcode'] == salt.defaults.exitcodes.EX_OK and recursive
    and 'No unfinished transactions left.' not in cmd_ret_list[-1]['stdout']):
```
But your miles may wary. Your call. :wink:

isbm (approved):

terminalmage (approved):

salt/modules/yumpkg.py
leeclemens: @isbm I personally agree with "operator first", but I have seen the "operator last" in formatting examples and a bit in salt's own codebase (although only ~1/4 as often after a quick search in hindsight) so I went that route. Merged now, not sure why github didn't show a notification when I checked earlier.

Issue #40975, opened 2017-04-30
Feat: read Salt-Cloud Maps from the Pillar


khaije1:
FEATURE REQUEST

Cloud Profiles and Cloud Providers can already be defined in the Pillar.
https://docs.saltstack.com/en/latest/topics/cloud/config.html#pillar-configuration

Could Cloud Maps also be setup to similarly read from the Pillar?

gtmanfred:
Absolutely, this would be great to have.

It is worth noting that the cloud providers and cloud profiles are only readable when using salt-call right now.  It would be great to add that into the regular salt-cloud or salt-run cloud.create.

Thanks for the request.
Daniel

khaije1:
FYI for those looking for this functionality, it's alternatively available by using the absolutely flippin' brilliant Salt-Formula, visible here: https://github.com/saltstack-formulas/salt-formula/blob/master/pillar.example#L203

Consequently I'd consider this request either very low priority or simply closable at your preference.

Warm regards!

gtmanfred:
Yeah, it is on my low priority list right now, but I do want to do some
things that would make this possible soon.

khaije1:
I've used this recently (via salt-call) and just wanted to say thanks!

nickgarber:
I'd be happy to help implement it, though I'd require some tutelage and direction.

PR #47742, opened 2018-05-20
feat: read Salt-Cloud Maps from the Pillar

nickgarber:
### What does this PR do?
Enable use of cloud map data sourced from the Pillar.
Uses the map data found in the pillar at `cloud:maps:` + the key-name provided with `map_pillar` argument.

Example
```bash
salt-call cloud.map_run map_pillar=my-prd-map
```

Where
```yaml
cloud:
  maps:
    my-prd-map:
      named-profile-01:
        - named-instance-01
        - named-instance-02
```

To preserve backwards compatibility the `map_pillar` option is ignored in the presence of either `map_data` or `path` arguments.

The implementation is in two parts: (1) add maps data from the pillar at `CloudClient.__init__()`  (alongside the existing similarly loaded `profiles` and `providers` info), and (2) allow for selecting and reading from this data in `Map.read()`, which is the function that map data is sourced from.

#40975 

### Previous Behavior
Allow loading map data from a file or from a JSON data structure via CLI.

### New Behavior
Add a pillar-based option to the existing map data.
 
### Tests written?
No so far, but what sort of tests should this have?

### Commits signed with GPG?
No

### DIFF
--- a/salt/modules/cloud.py
+++ b/salt/modules/cloud.py
@@ -204,11 +204,21 @@ def map_run(path=None, **kwargs):
     '''
     Execute a salt cloud map file
 
+    Cloud Map data can be retrieved from several sources:
+      - a local file (provide the path to the file to the 'path' argument)
+      - a JSON-formatted map directly (provide the appropriately formatted to using the 'map_data' argument)
+      - the Salt Pillar (provide the map name of under 'pillar:cloud:maps' to the 'map_pillar' argument)
+      Note: Only one of these sources can be read at a time. The options are listed in their order of precedence.
+
+
     CLI Examples:
 
     .. code-block:: bash
 
         salt minionname cloud.map_run /path/to/cloud.map
+        salt minionname cloud.map_run path=/path/to/cloud.map
+        salt minionname cloud.map_run map_pillar='<map_pillar>'
+          .. versionchanged:: 2018.3.1
         salt minionname cloud.map_run map_data='<actual map data>'
     '''
     client = _get_client()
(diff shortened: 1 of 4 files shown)


salt/modules/cloud.py
```diff
@@ -210,6 +210,8 @@ def map_run(path=None, **kwargs):
 
         salt minionname cloud.map_run /path/to/cloud.map
         salt minionname cloud.map_run map_data='<actual map data>'
+        # to use map data in pillar under salt:cloud:maps:<map_pillar>'
```
gtmanfred: I believe this should be `cloud:maps` not `salt:cloud:maps`

salt/modules/cloud.py
```diff
@@ -210,6 +210,8 @@ def map_run(path=None, **kwargs):
...
+        salt minionname cloud.map_run map_pillar='<map_pillar>'
```
gtmanfred: Can you add a section above this for map_pillar and really describe what it is and how to use it?

gtmanfred (changes requested):
This looks great!

Just a couple docs issue that need to be cleaned up.

salt/modules/cloud.py
nickgarber: Heya @gtmanfred, you're absolutely right! I was referencing it's location in the salt-formula pillar by accident. I have the correction for this on my laptop already - will push ASAP.

cachedout (changes requested):
Once we get docs pushed that will be reflected here, I think this looks good:

https://docs.saltstack.com/en/latest/topics/cloud/config.html#pillar-configuration

nickgarber:
I've added docs and am just watching for tests. Does this change need to include tests? Any other feedback to consider?

gtmanfred (approved):

gtmanfred:
Tests would be great.

You can put them in unit tests, or if you want to write an integration test, we put them in salt/integration/cloud, and add the @expensiveTest decorator to them.

Thanks!
Daniel

PR #47751, opened 2018-05-21
Allow value 0(zero) for updatables in ResourceRecordSet.

### LINKED ISSUE #47693: Cannot set ResourceRecord Weight to 0(zero)
https://github.com/saltstack/salt/blob/9ee2786b622412ec1d79c861d12ce44e4eacbfcf/salt/states/boto3_route53.py#L704-L705

The linked `if` statement results in `Weight` being completely omitted from the `ResourceRecordSet` when defined as `0`. `0` is a completely valid `Weight`, but omitting it from the API call results in an AWS API error.

ezekiel:
### What does this PR do?
Allow value 0(zero) for updatables in ResourceRecordSet.

Fixes #47693 

### Previous Behavior
Python treats zero as false:
  https://docs.python.org/2/library/stdtypes.html#truth-value-testing

False causes the updatable to be omitted from the API request, which causes AWS API
errors in some cases (e.g. defining Weight: 0).

### New Behavior
Successfully updates with values of zero in the ResourceRecordSet.

### Tests written?
No

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/states/boto3_route53.py
+++ b/salt/states/boto3_route53.py
@@ -702,7 +702,10 @@ def rr_present(name, HostedZoneId=None, DomainName=None, PrivateZone=False, Name
         if ResourceRecords:
             ResourceRecordSet['ResourceRecords'] = ResourceRecords
         for u in updatable:
-            ResourceRecordSet.update({u: locals().get(u)}) if locals().get(u) else None
+            if locals().get(u) or (locals().get(u) == 0):
+                ResourceRecordSet.update({u: locals().get(u)})
+            else:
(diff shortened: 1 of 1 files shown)


rallytime (approved):
LGTM

rallytime:
@ryan-lane Do you want to review this?

ezekiel:
Rebased, re-signed, and pushed to be sure I don't have conflicts with current development.

rallytime:
Welcome @ezekiel - Thank you for fixing this. :)

PR #47753, opened 2018-05-21
Add stderr launchctl helper class and fix service mac tests

Ch3LL:
### What does this PR do?
Improve the error reporting for when the `service is disabled` message shows up in the launchctl stderr. Currently launchctl does not return a bad exit code so catching this error to report to the user they cannot start a service since its disabled. Also fixing the mac service tests. Currently this test is failing: `integration.states.test_service.ServiceTest.test_service_running` because i believe a test in modules.test_mac_service is  disabling the service, so I improved the tearDown there as well to ensure we are setting the disable/enabled status back to what it was initially.

### DIFF
--- a/salt/utils/mac_utils.py
+++ b/salt/utils/mac_utils.py
@@ -119,6 +119,18 @@ def _run_all(cmd):
     return ret
 
 
+def _check_launchctl_stderr(ret):
+    '''
+    helper class to check the launchctl stderr.
+    launchctl does not always return bad exit code
+    if there is a failure
+    '''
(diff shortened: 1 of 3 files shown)


rallytime (approved):

rallytime:
@Ch3LL Looks like this change caused a related failure:

- integration.states.test_service.ServiceTest.test_service_running

https://jenkins.saltstack.com/job/PR/job/salt-pr-linode-ubuntu14-n/23036/

rallytime:
@Ch3LL This will need a rebase as well.

Issue #47760, opened 2018-05-22
salt-cloud: openstack driver: userdata is not configured on the instance


zerthimon:
### Description of Issue/Question
userdata is configured in cloud profile, gets rendered, but doesn't end up in the new instance

### Setup
from cloud provider:
```
myopenstack:
  driver: openstack
  ...
```

from cloud profile:
```
ubuntu:
  provider: myopenstack
  ...
  userdata: /etc/salt/cloud.init/userdata.jinja
  userdata_template: jinja
```

### Steps to Reproduce Issue
`salt-cloud -p ubuntu myinstance`

### Versions Report
```
# salt-cloud -V
Salt Version:
            Salt: 2018.3.0
 
Dependency Versions:
 Apache Libcloud: 1.5.0
            cffi: 1.11.5
        cherrypy: 3.5.0
        dateutil: 2.4.2
       docker-py: 3.3.0
           gitdb: 0.6.4
       gitpython: 1.0.1
           ioflo: Not Installed
          Jinja2: 2.8
         libgit2: Not Installed
         libnacl: Not Installed
        M2Crypto: Not Installed
            Mako: 1.0.3
    msgpack-pure: Not Installed
  msgpack-python: 0.5.6
    mysql-python: Not Installed
       pycparser: 2.18
        pycrypto: 2.6.1
    pycryptodome: Not Installed
          pygit2: Not Installed
          Python: 2.7.12 (default, Dec  4 2017, 14:50:18)
    python-gnupg: Not Installed
          PyYAML: 3.12
           PyZMQ: 15.2.0
            RAET: Not Installed
           smmap: 0.9.0
         timelib: Not Installed
         Tornado: 4.2.1
             ZMQ: 4.1.4
 
System Versions:
            dist: Ubuntu 16.04 xenial
          locale: UTF-8
         machine: x86_64
         release: 4.4.0-124-generic
          system: Linux
         version: Ubuntu 16.04 xenial
```

zerthimon:
userdata parameter is not allowed in `_clean_create_kwargs` function in cloud/clouds/openstack.py (line 564) adding in to the `VALID_OPTS` fixes the issue
```
--- _openstack.py_	2018-05-22 11:51:12.685595485 +0300
+++ openstack.py	2018-05-22 12:28:14.827846504 +0300
@@ -597,6 +597,7 @@
         'volume_size': int,
         'nat_destination': six.string_types,
         'group': six.string_types,
+        'userdata': six.string_types,
     }
[...]

PR #47761, opened 2018-05-22
Fix userdata in openstack instances

zerthimon:
### What does this PR do?
Fix instance userdata in openstack instances

https://github.com/saltstack/salt/issues/47760

### Previous Behavior
metadata is not created for openstack instance

### New Behavior
metadata is created for openstack instance

### Tests written?
No

### Commits signed with GPG?
No

### DIFF
--- a/salt/cloud/clouds/openstack.py
+++ b/salt/cloud/clouds/openstack.py
@@ -597,6 +597,7 @@ def _clean_create_kwargs(**kwargs):
         'volume_size': int,
         'nat_destination': six.string_types,
         'group': six.string_types,
+        'userdata': six.string_types,
     }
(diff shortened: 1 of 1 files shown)


gtmanfred (approved):

gtmanfred:
This needs to be backported to 2018.3 as well.

gtmanfred:
@zerthimon Thank you for the PR, for future reference, bug fixes should be made against the major branch of the oldest release where it is broken, in this case it would be 2018.3, and would be merged forward.  https://docs.saltstack.com/en/latest/topics/development/contributing.html#merge-forward-process

We will get this one backported though.

Thanks!
Daniel

Issue #45016, opened 2017-12-14
Add pkg state windows Tests to Auto Test Suite


Ch3LL:
This task is to monitor the addition of one test to the auto test suite:

- [ ] pkg.installed
- [ ] pkg.latest

this is using state pkg modules

PR #47778, opened 2018-05-22
Add test_pkg integration state tests to windows

Ch3LL:
### What does this PR do?
Allow the state test_pkg integration tests to run on windows.

https://github.com/saltstack/salt/issues/45016

### DIFF
--- a/tests/integration/files/file/base/win/repo-ng/7zip.sls
+++ b/tests/integration/files/file/base/win/repo-ng/7zip.sls
@@ -1,0 +1,24 @@
+{% set versions = {'18':['05', '03', '01'], '16':['04', '03', '02', '00'], '9':['20']} %}
+
+7zip:
+{% for major, subversions in versions.items() %}
+{% for minor in subversions %}
+  '{{major}}.{{minor}}.00.0':
+    {% if grains['cpuarch'] == 'AMD64' %}
+    full_name: '7-Zip {{major}}.{{minor}} (x64 edition)'
+    installer: 'http://d.7-zip.org/a/7z{{major}}{{minor}}-x64.msi'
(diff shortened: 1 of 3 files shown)


damon-atkins:
Instead of hard coding a sls file I suggest you pull them down from/clone 
https://github.com/saltstack/salt-winrepo
Given your fetching the software from the internet.

tests/integration/files/file/base/win/repo-ng/7zip.sls
```diff
@@ -0,0 +1,24 @@
+{% set versions = {'18':['05', '03', '01'], '16':['04', '03', '02', '00'], '9':['20']} %}
```
damon-atkins: See comments

Ch3LL: I think this is fine as this test isn't testing the salt-winrepo repo but only the functionality of `pkg.install` just as we are also doing for `putty` in the test repo. I believe it would be more useful to add tests specifically for win-repo if we want to test the repo files themselves. Hopefully we can move towards using our own build packages as well in the future. But would like to get @twangboy and @dwoz input here as well.

damon-atkins:
> testing the salt-winrepo repo 

One item which most likely is not being tested is the settings in this section https://docs.saltstack.com/en/latest/ref/configuration/master.html#windows-software-repo-settings

If you were to pull down the salt-repo-ng you could use the above settings, which would increase the test coverage

tests/integration/files/file/base/win/repo-ng/7zip.sls
dwoz: @Ch3LL If the intent is to test `pkg.install` I am fine with this. I agree adding tests specifically for the win-repo functionality is a good future step.

dwoz (approved):

PR #47779, opened 2018-05-22
Increase ec2 cloud timeouts to 1000

dwoz:
### What does this PR do?

Increase all the ec2 could test timeouts, again :)

### Commits signed with GPG?

Yes


tests/integration/cloud/providers/test_ec2.py
```diff
@@ -45,7 +45,6 @@ class EC2Test(ShellCase):
     '''
     Integration tests for the EC2 cloud provider in Salt-Cloud
     '''
-    TIMEOUT = 500
```
rallytime: Is there a reason you're not just changing this variable and setting it everywhere in this file?

dwoz: Only because that was not being used consistently within the file.

rallytime (approved):

PR #47780, opened 2018-05-22
External WAL directory and datapage checksums for PostgreSQL

moench-tegeder:
Since almost forever, PostgreSQL allows storing it's transaction log
(WAL, write ahead log) outside of the data directory (this improves
performance on certain setups, or may just be required by local
policy. The newly added "waldir" option on states.postgres_initdb.present()
and modules.postgresql.initdb() is passed to initdb (the default is
still to keep the transaction log inside the data directory).
As part of the big renaming in PostgreSQL 10 the old "xlog directory"
has been renamed to "wal directory", this patch uses the new name only.

Sind PostgreSQL 9.3 (which is the oldest supported release as of
time of writing), there is support for checksumming the database
on the block level. This can only be enabled when creating the
cluster (this is, at initdb time), so it should be available in
modules and states. This patch adds it in both places.

Documentation for both mini-features has been added to the doc strings.

### Previous Behavior

### New Behavior

### Tests written?

Yes/No

### Commits signed with GPG?

Yes/No


moench-tegeder:
The versionadded tag has been set to XXX, as it's up to your release planning if and when you merge this.

salt/modules/postgres.py
```diff
@@ -3109,8 +3122,21 @@ def datadir_init(name,
...
+    checksums
+        If True, the cluster will be created with data page checksums.
+
     runas
         The system user the operation should be performed on behalf of
+
+    .. note:
+
+        The ``waldir`` and ``checksum`` options have been added in
+        version XXX. The ``checksum`` option requires PostgreSQL 9.3
```
gtmanfred: This will be added in `Fluorine`, you can use the `.. versionadded: Fluorine` macro for rst.

gtmanfred (changes requested):

moench-tegeder:
Ah right. I shouldn't fix documentation while overtired. Gave it a little improvement, too. Thanks for noticing.

gtmanfred (changes requested):
Just need to fix the spelling and this should be good

salt/modules/postgres.py
```diff
@@ -3109,8 +3122,22 @@ def datadir_init(name,
...
 
+    waldir
+        The transaction log (WAL) directory (default is to keep WAL
+        inside the data directory)
+
+        .. versionadded:: Flourine
+
+    checksums
+        If True, the cluster will be created with data page checksums.
+
```
gtmanfred: Should be Fluorine, like the element in the periodic table

moench-tegeder:
Drats. One of the few words spelled with "uo".

salt/states/postgres_initdb.py
```diff
@@ -61,6 +63,19 @@ def present(name,
     locale
         The default locale for new databases
 
+    waldir
+        The transaction log (WAL) directory (default is to keep WAL
+        inside the data directory)
+
+        .. versionadded:: Flourine
```
gtmanfred: One last thing :P these guys need to be fixed too.

gtmanfred (approved):

salt/modules/postgres.py
```diff
@@ -199,6 +199,8 @@ def _run_initdb(name,
         password=None,
         encoding='UTF8',
         locale=None,
+        waldir=None,
```
rallytime: Can you move these new kwargs to the end of the list? We don't want to cause an API change in case users are passing positional args.

salt/modules/postgres.py
```diff
@@ -3079,6 +3090,8 @@ def datadir_init(name,
...
+        checksums=False,
```
rallytime: Same comment here.

rallytime (changes requested):
A couple of small requests from me.

moench-tegeder:
sure, while I'm here.
And while grep failed me yesterday, today it found even more misspellings: https://github.com/saltstack/salt/pull/48052 :)

moench-tegeder:
Hi again,
is there anything I can do here? Unless I'm missing something, thw kwargs are now in order?

rallytime (approved):

rallytime:
Thanks for the bump. It looks great now. Thank you!

Issue #47559, opened 2018-05-09
Setting Timezone to Berlin sets it to Longyearbyen


soner89:
### Description of Issue/Question
I want to set the Timezone of a Windows server via Salt to Europe/Berlin. Setting the value returns "True" (=correct behaviour).
Retrieving the Timezone of this server would return "Arctic/Longyearbyen" which is the exact the same time as in Berlin. I expected to retrieve "Europe/Berlin" instead of "Arctic/Longyearbyen".

### Setup
Salt Master: 2016.11.3
Salt Minion: 2016.11.3

### Steps to Reproduce Issue
On the Salt-Master:
```
# salt my_minion timezone.set_zone 'Europe/Berlin'
my_minion:
    True

# salt my_minion timezone.get_zone
my_minion:
    Arctic/Longyearbyen
```

### Versions Report
```
Salt Version:
           Salt: 2016.11.3
 
Dependency Versions:
           cffi: 1.6.0
       cherrypy: Not Installed
       dateutil: 2.6.1
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.7.2
        libgit2: 0.24.6
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.4.6
   mysql-python: Not Installed
      pycparser: 2.14
       pycrypto: 2.6.1
         pygit2: 0.24.2
         Python: 2.7.5 (default, Aug  4 2017, 00:39:18)
   python-gnupg: Not Installed
         PyYAML: 3.11
          PyZMQ: 15.3.0
           RAET: Not Installed
          smmap: Not Installed
        timelib: Not Installed
        Tornado: 4.2.1
            ZMQ: 4.1.4

System Versions:
           dist: centos 7.4.1708 Core
        machine: x86_64
        release: 3.10.0-229.11.1.el7.x86_64
         system: Linux
        version: CentOS Linux 7.4.1708 Core

```

garethgreenaway:
@soner89 Thanks for the report.  
@dwoz @twangboy Thoughts on this one?

terminalmage:
[...]

PR #47788, opened 2018-05-23
Use pytz to calculate timezones

twangboy:
### What does this PR do?
Fixes issue where erroneous Timezones were being returned. Uses pytz module to get timzone information. Fixes `timezone.get_offset` so it actually returns an offset.

https://github.com/saltstack/salt/issues/47559

### Previous Behavior
Everything was in a giant dictionary so the `get_timezone` just returned the first item it found in the dictionary. `get_offset` always returned false.

### New Behavior
The dictionary only has one possible option for the timezone it finds, so `get_zone` and `get_zonecode` will be mostly right.
The `get_offset` function actually works.

### Tests written?
No

### Commits signed with GPG?
Yes

### DIFF
--- a/doc/topics/releases/2018.3.2.rst
+++ b/doc/topics/releases/2018.3.2.rst
@@ -1,0 +1,16 @@
+===========================
+In Progress: Salt 2018.3.2 Release Notes
+===========================
+
+Version 2018.3.2 is an **unreleased** bugfix release for :ref:`2018.3.0 <release-2018-3-0>`.
+This release is still in progress and has not been released yet.
+
+Changes to win_timezone
+-----------------------
+
+Improves timezone detection by using the pytz module.
+
+``timezone.get_offset`` and ``timezone.get_zonecode`` now work properly.
+
+Adds ``timezone.list`` to list supported timezones in either Windows or Unix
+format.
--- a/doc/topics/releases/2018.3.2.rst
+++ b/doc/topics/releases/2018.3.2.rst
@@ -1,12 +1,12 @@
-===========================
+========================================
 In Progress: Salt 2018.3.2 Release Notes
-===========================
+========================================
 
 Version 2018.3.2 is an **unreleased** bugfix release for :ref:`2018.3.0 <release-2018-3-0>`.
 This release is still in progress and has not been released yet.
 
 Changes to win_timezone
------------------------
+=======================
 
 Improves timezone detection by using the pytz module.
 
--- a/salt/modules/win_timezone.py
+++ b/salt/modules/win_timezone.py
@@ -11,9 +11,6 @@ from datetime import datetime
 
 # Import Salt libs
 from salt.exceptions import CommandExecutionError
-import salt.utils.path
-import salt.utils.platform
-import salt.utils.win_reg
 
 log = logging.getLogger(__name__)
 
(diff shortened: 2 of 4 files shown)


dwoz (approved):

rallytime:
@twangboy The `test_win_timezone` tests are failing with this change.

https://jenkins.saltstack.com/job/PR/job/salt-pr-linode-ubuntu14-n/23102/

salt/modules/win_timezone.py
```diff
@@ -6,590 +6,387 @@
 
 # Import Python libs
 import logging
-import re
+import pytz
```
terminalmage: This is not a hard dep for Salt, is it? If not, then this will cause a lot of tracebacks on non-Windows boxes which don't have pytz installed, unless we gate this in a try/except to catch the `ImportError`.

Also, we'd need to make sure that we add pytz to the freezer_includes or whatever we use now when making the Windows release.

What about the gist I posted in [this comment](https://github.com/saltstack/salt/issues/47559#issuecomment-388572450)? It seems like there was a way to get this to work using a different flag for tzutil. That would be the optimal solution as it would not require us to dep on another module.

doc/topics/releases/2018.3.2.rst
```diff
@@ -0,0 +1,16 @@
+===========================
+In Progress: Salt 2018.3.2 Release Notes
+===========================
```
terminalmage: The overline and underline must match the length of the title line to avoid Sphinx warnings.

doc/topics/releases/2018.3.2.rst
```diff
@@ -0,0 +1,16 @@
...
+
+Version 2018.3.2 is an **unreleased** bugfix release for :ref:`2018.3.0 <release-2018-3-0>`.
+This release is still in progress and has not been released yet.
+
+Changes to win_timezone
+-----------------------
```
terminalmage: Please change this underline to use `====` instead of `----`, to match the pending changes to the changelog generator. The order of use of different underline chars affects the section names generated by Sphinx.

terminalmage (changes requested):
A couple changes need to be made to the release notes you added.

@twangboy To respond to [this comment](https://github.com/saltstack/salt/pull/47788#discussion_r190960058), if you read the gist closely it talks about having gotten the CLDR time zone information (that's the format with the offsets) from tzutil on Windows. This can apparently be done with the `/l` flag. Did you try that?

Having said all this though, pytz is better suited to this sort of thing, and as long as it is A) only used on Windows, and B) already included in the embedded Python we distribute with Windows, then we should use pytz instead of trying to parse tzutil.

doc/topics/releases/2018.3.2.rst
twangboy: Sorry, just copied the file from 2018.3.1 and changed the 1 to a 2. Shall I go through and fix all of them?

twangboy: I think it's because the `In Progress:` eventually gets removed when they're ready for release.

twangboy:
@terminalmage I did look at the gist. The `/l` option returns a list of all Windows Timezones with their offset. The `/g` option returns the name of the timezone without the offset. The gist you're referring to just creates a dictionary you can use to marry the Windows timezone to the iana standard in a 1 to 1 relationship... which is what we're doing in this PR. Though I got the dictionary from somewhere else.

terminalmage (approved):

terminalmage:
Yeah, I know what the gist did... My original point was that we could have perhaps used tzutil to get a more accurate timezone mapping than we were already using. But given what I mentioned in my last comment, pytz is definitely a better solution.

PR #47808, opened 2018-05-23
Updated AIX support for grains reported

dmurphy18:
### What does this PR do?
Increased the grains reported and corrects the values reported for machien_id, cpuarch, cpu_model,  os information, product name, manufacturer, serialn umber and firmware

https://github.com/saltstack/zh/issues/1341

### Previous Behavior
Some of the grains were not reported or with incomplete information

### New Behavior
AIX now reports grains fully, similar to Redhat

### Tests written?
No, check by hand

### Commits signed with GPG?
No

### DIFF
--- a/salt/grains/core.py
+++ b/salt/grains/core.py
@@ -409,62 +409,19 @@ def _aix_cpudata():
     #   cpu_model
     #   cpu_flags
     grains = {}
-    data = ""
     cmd = salt.utils.path.which('prtconf')
     if cmd:
-        data += __salt__['cmd.run']('{0}'.format(cmd))
-        data += '\n'
-
-        cpuarch_regexes = [
-            re.compile(r) for r in [
-                r'(?im)^\s*Processor\s+Type:\s+(\S+)',  # prtconf
-            ]
-        ]
-
-        cpuflags_regexes = [
-            re.compile(r) for r in [
-                r'(?im)^\s*Processor\s+Version:\s+(\S+)',  # prtconf
-            ]
-        ]
-
-        cpumodel_regexes = [
-            re.compile(r) for r in [
-                r'(?im)^\s*Processor\s+Implementation\s+Mode:\s+(.*)',  # prtconf
-            ]
-        ]
-
-        numcpus_regexes = [
-            re.compile(r) for r in [
-                r'(?im)^\s*Number\s+Of\s+Processors:\s+(\S+)',  # prtconf
-            ]
-        ]
-
-        for regex in cpuarch_regexes:
-            res = regex.search(data)
-            if res and len(res.groups()) >= 1:
-                grains['cpuarch'] = res.group(1).strip().replace("'", "")
-                break
-
-        for regex in cpuflags_regexes:
-            res = regex.search(data)
-            if res and len(res.groups()) >= 1:
-                grains['cpu_flags'] = res.group(1).strip().replace("'", "")
-                break
-
-        for regex in cpumodel_regexes:
-            res = regex.search(data)
-            if res and len(res.groups()) >= 1:
-                grains['cpu_model'] = res.group(1).strip().replace("'", "")
-                break
-
-        for regex in numcpus_regexes:
-            res = regex.search(data)
-            if res and len(res.groups()) >= 1:
-                grains['num_cpus'] = res.group(1).strip().replace("'", "")
-                break
+        data = __salt__['cmd.run']('{0}'.format(cmd)) + os.linesep
+        for dest, regstring in (('cpuarch', r'(?im)^\s*Processor\s+Type:\s+(\S+)'),
+                                ('cpu_flags', r'(?im)^\s*Processor\s+Version:\s+(\S+)'),
+                                ('cpu_model', r'(?im)^\s*Processor\s+Implementation\s+Mode:\s+(.*)'),
+                                ('num_cpus', r'(?im)^\s*Number\s+Of\s+Processors:\s+(\S+)')):
+            for regex in [re.compile(r) for r in [regstring]]:
+                res = regex.search(data)
+                if res and len(res.groups()) >= 1:
+                    grains[dest] = res.group(1).strip().replace("'", '')
     else:
         log.error('The \'prtconf\' binary was not found in $PATH.')
-
     return grains
 
 
@@ -571,9 +528,9 @@ def _aix_memdata():
     prtconf = salt.utils.path.which('prtconf')
     if prtconf:
         for line in __salt__['cmd.run'](prtconf, python_shell=True).splitlines():
-            comps = line.split(' ')
-            if 'Memory' == comps[0].strip() and 'Size:' == comps[1].strip():
-                grains['mem_total'] = int(comps[2].strip())
+            comps = [x for x in line.strip().split(' ') if x]
+            if len(comps) > 2 and 'Memory' in comps[0] and 'Size' in comps[1]:
+                grains['mem_total'] = int(comps[2])
                 break
     else:
         log.error('The \'prtconf\' binary was not found in $PATH.')
@@ -582,13 +539,12 @@ def _aix_memdata():
     if swap_cmd:
         swap_data = __salt__['cmd.run']('{0} -s'.format(swap_cmd)).split()
         try:
-            swap_total = int(swap_data[-2]) * 4 + int(swap_data[-6]) * 4
+            swap_total = (int(swap_data[-2]) + int(swap_data[-6])) * 4
         except ValueError:
             swap_total = None
         grains['swap_total'] = swap_total
     else:
         log.error('The \'swap\' binary was not found in $PATH.')
-
     return grains
 
 
@@ -2597,47 +2553,21 @@ def _hw_data(osdata):
                 break
 
     elif osdata['kernel'] == 'AIX':
-        data = ""
         cmd = salt.utils.path.which('prtconf')
-        if cmd:
-            data += __salt__['cmd.run']('{0}'.format(cmd)) + os.linesep
+        if data:
+            data = __salt__['cmd.run']('{0}'.format(cmd)) + os.linesep
+            for dest, regstring in (('serialnumber', r'(?im)^\s*Machine\s+Serial\s+Number:\s+(\S+)'),
+                                    ('systemfirmware', r'(?im)^\s*Firmware\s+Version:\s+(.*)')):
+                for regex in [re.compile(r) for r in [regstring]]:
+                    res = regex.search(data)
+                    if res and len(res.groups()) >= 1:
+                        grains[dest] = res.group(1).strip().replace("'", '')
 
-            sn_regexes = [
-                re.compile(r) for r in [
-                    r'(?im)^\s*Machine\s+Serial\s+Number:\s+(\S+)',  # prtconf
-                ]
-            ]
-
-            fw_regexes = [
-                re.compile(r) for r in [
-                    r'(?im)^\s*Firmware\s+Version:\s+(.*)',  # prtconf
-                ]
-            ]
-
-            product_regexes = [
-                re.compile(r) for r in [
-                    r'(?im)^\s*System\s+Model:\s+(\S+)',  # prtconf
-                ]
-            ]
-
-            for regex in sn_regexes:
-                res = regex.search(data)
-                if res and len(res.groups()) >= 1:
-                    grains['serialnumber'] = res.group(1).strip().replace("'", "")
-                    break
-
-            for regex in fw_regexes:
-                res = regex.search(data)
-                if res and len(res.groups()) >= 1:
-                    fw_string = res.group(1).strip().replace("'", "")
-                    _, grains['systemfirmware'] = fw_string.split(",")
-                    break
-
+            product_regexes = [re.compile(r'(?im)^\s*System\s+Model:\s+(\S+)')]
             for regex in product_regexes:
                 res = regex.search(data)
                 if res and len(res.groups()) >= 1:
-                    sn_string = res.group(1).strip().replace("'", "")
-                    grains['manufacturer'], grains['productname'] = sn_string.split(",")
+                    grains['manufacturer'], grains['productname'] = res.group(1).strip().replace("'", "").split(",")
                     break
         else:
             log.error('The \'prtconf\' binary was not found in $PATH.')
--- a/salt/grains/core.py
+++ b/salt/grains/core.py
@@ -408,60 +408,61 @@ def _aix_cpudata():
     #   num_cpus
     #   cpu_model
     #   cpu_flags
+
     grains = {}
-    cmd = '/usr/sbin/prtconf'
     data = ""
-    if salt.utils.path.which(cmd):  # Also verifies that cmd is executable
+    cmd = salt.utils.path.which('prtconf')
+    if cmd:
         data += __salt__['cmd.run']('{0}'.format(cmd))
         data += '\n'
 
-    cpuarch_regexes = [
-        re.compile(r) for r in [
-            r'(?im)^\s*Processor\s+Type:\s+(\S+)',  # prtconf
+        cpuarch_regexes = [
+            re.compile(r) for r in [
+                r'(?im)^\s*Processor\s+Type:\s+(\S+)',  # prtconf
+            ]
         ]
-    ]
 
-    cpuflags_regexes = [
-        re.compile(r) for r in [
-            r'(?im)^\s*Processor\s+Version:\s+(\S+)',  # prtconf
+        cpuflags_regexes = [
+            re.compile(r) for r in [
+                r'(?im)^\s*Processor\s+Version:\s+(\S+)',  # prtconf
+            ]
         ]
-    ]
 
-    cpumodel_regexes = [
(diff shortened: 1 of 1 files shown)


salt/grains/core.py
```diff
@@ -2098,6 +2202,10 @@ def get_machine_id():
     '''
     # Provides:
     #   machine-id
+
+    if platform.system() == 'AIX':
+        return {'machine_id': platform.machine()}
```
brejoc: Isn't `platform.machine()` returning the architecture of the machine? At least on Linux it does. I think `get_machine_id` should return something else here.

salt/grains/core.py
```diff
@@ -2434,6 +2542,51 @@ def _hw_data(osdata):
...
+    elif osdata['kernel'] == 'AIX':
+        data = ""
+        cmd = '/usr/sbin/prtconf'
+        if salt.utils.path.which(cmd):  # Also verifies that cmd is executable
+            data += __salt__['cmd.run']('{0}'.format(cmd))
+            data += '\n'
+
+        sn_regexes = [
+            re.compile(r) for r in [
+                r'(?im)^\s*Machine\s+Serial\s+Number:\s+(\S+)',  # prtconf
```
brejoc: Would this still work, if the locale of the installation would change?

salt/grains/core.py
```diff
@@ -399,6 +399,73 @@ def _sunos_cpudata():
...
+    #   cpuarch
+    #   num_cpus
+    #   cpu_model
+    #   cpu_flags
+    grains = {}
+    cmd = '/usr/sbin/prtconf'
+    data = ""
+    if salt.utils.path.which(cmd):  # Also verifies that cmd is executable
+        data += __salt__['cmd.run']('{0}'.format(cmd))
+        data += '\n'
```
isbm:
```python
data = __salt__['cmd.run']('{0}'.format(cmd)) + os.linesep
```

salt/grains/core.py
```diff
@@ -399,6 +399,73 @@ def _sunos_cpudata():
...
+        res = regex.search(data)
+        if res and len(res.groups()) >= 1:
+            grains['cpu_model'] = res.group(1).strip().replace("'", "")
+            break
+
+    for regex in numcpus_regexes:
+        res = regex.search(data)
+        if res and len(res.groups()) >= 1:
+            grains['num_cpus'] = res.group(1).strip().replace("'", "")
+            break
```
isbm: OK, how about compacting all this into just couple of lines like so?
```python
for dest, regstring in (('cpuarch', r'(?im)^\s*Processor\s+Type:\s+(\S+)'),
                        ('cpu_flags', r'(?im)^\s*Processor\s+Version:\s+(\S+)'),
                        ('cpu_model', r'(?im)^\s*Processor\s+Implementation\s+Mode:\s+(.*)'),
                        ('num_cpus', r'(?im)^\s*Number\s+Of\s+Processors:\s+(\S+)')):
    for regex in [re.compile(r) for r in [regstring]]:
        res = regex.search(data)
        if res and len(res.grouns()) >= 1:
            grains[dest] = res.group(1).strip().replace("'", '')
```
Unit tests are, of course, welcome. :wink:

salt/grains/core.py
```diff
@@ -494,6 +561,31 @@ def _sunos_memdata():
     return grains
 
 
+def _aix_memdata():
+    '''
+    Return the memory information for AIX systems
+    '''
+    grains = {'mem_total': 0, 'swap_total': 0}
+
```
isbm: Please remove this empty line.

salt/grains/core.py
```diff
@@ -494,6 +561,31 @@ def _sunos_memdata():
...
+    prtconf = '/usr/sbin/prtconf 2>/dev/null'
+    for line in __salt__['cmd.run'](prtconf, python_shell=True).splitlines():
+        comps = line.split(' ')
+        if comps[0].strip() == 'Memory' and comps[1].strip() == 'Size:':
```
isbm: Unreliable. I would rather do `if foo in comps[x]` instead.

salt/grains/core.py
```diff
@@ -494,6 +561,31 @@ def _sunos_memdata():
...
+            grains['mem_total'] = int(comps[2].strip())
+            break
+
+    swap_cmd = salt.utils.path.which('swap')
+    swap_data = __salt__['cmd.run']('{0} -s'.format(swap_cmd)).split()  # 4k blocks
+    try:
+        swap_avail = int(swap_data[-2]) * 4
+        swap_used = int(swap_data[-6]) * 4
+        swap_total = (swap_avail + swap_used)
```
isbm: Why not just:
```python
swap_total = int(swap_data[-2]) * 4 + int(swap_data[-6]) * 4
```

isbm: And I am not sure if this is the entirely right way, but my AIX knowledge greatly outdated. Back in times I was using `lsps` instead, since it shows disks that are swap allocated, but some of them might be inactive at this particular moment. Maybe you would like consider to take this to the account? Correct me if I am wrong.

isbm (changes requested):
@dmurphy18 could you please look at these changes? Thanks!

salt/grains/core.py
dmurphy18: I agree, platform.machine() usually returns the machine architecture, but apparently on AIX it does not, both with the Salt contained version of Python 2.7.5 that ships with Salt on AIX and with IBM's own Python 2.7.13.  On AIX the command uname -m displays the machine ID number of the hardware running the system, see https://www.ibm.com/support/knowledgecenter/ssw_aix_71/com.ibm.aix.cmds5/uname.htm

Hence the following sample:
root@l3ferp005_pub:~/test# python
Python 2.7.13 (default, Oct 17 2017, 05:37:56) [C] on aix6
>>> import platform
>>> platform.machine()
'00F7F3FE4C00'
>>> platform.uname()  
('AIX', 'l3ferp005_pub', '1', '7', '00F7F3FE4C00', 'powerpc')
>>> 
root@l3ferp005_pub:~/test# uname -m
00F7F3FE4C00
root@l3ferp005_pub:~/test# 

Given Salt uses a self contained version of Python 2.7.5 which produces as above, believe it is safe to use for now, until AIX corrects their use of uname -m which normally returns the machine architecture on Linux.  It will have to be rechecked when Salt AIX supports Python 3 which gets more attention than Python 2.

salt/grains/core.py
```diff
@@ -399,6 +399,73 @@ def _sunos_cpudata():
...
+    '''
+    Return CPU information for AIX systems
+    '''
+    # Provides:
+    #   cpuarch
+    #   num_cpus
+    #   cpu_model
+    #   cpu_flags
+    grains = {}
+    cmd = '/usr/sbin/prtconf'
```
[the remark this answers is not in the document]
dmurphy18: Revised to use salt.utils.path.which, which attempts to use decorator for which caused issues in the loader, see approx line 734 /salt/loader.py, problem with decorator requiring an argument.

salt/grains/core.py
```diff
@@ -494,6 +561,31 @@ def _sunos_memdata():
     return grains
 
 
+def _aix_memdata():
+    '''
+    Return the memory information for AIX systems
+    '''
+    grains = {'mem_total': 0, 'swap_total': 0}
+
+    prtconf = '/usr/sbin/prtconf 2>/dev/null'
```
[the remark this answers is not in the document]
dmurphy18: Revised to use salt.utils.path.which, which attempts to use decorator for which caused issues in the loader, see approx line 734 /salt/loader.py, problem with decorator requiring an argument.

salt/grains/core.py
```diff
@@ -2434,6 +2542,51 @@ def _hw_data(osdata):
                 grains['product'] = res.group(1).strip().replace("'", "")
                 break
 
+    elif osdata['kernel'] == 'AIX':
+        data = ""
+        cmd = '/usr/sbin/prtconf'
+        if salt.utils.path.which(cmd):  # Also verifies that cmd is executable
+            data += __salt__['cmd.run']('{0}'.format(cmd))
+            data += '\n'
```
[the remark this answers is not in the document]
dmurphy18: will add logging, but prtconf is not a symlink on AIX v7.1

root@l490vp024_pub:/opt/freeware/lib/python2.7/site-packages/salt/grains# ls -al /usr/sbin/prtconf 
-r-xr-xr-x    2 root     system        14265 Mar 17 2015  /usr/sbin/prtconf

salt/grains/core.py
```diff
@@ -1759,6 +1853,16 @@ def os_data():
         grains.update(_bsd_cpudata(grains))
         grains.update(_osx_gpudata())
         grains.update(_osx_platform_data())
+    elif grains['kernel'] == 'AIX':
+        osrelease = __salt__['cmd.run']('oslevel')
```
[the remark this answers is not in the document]
dmurphy18: yes, there is always an oslevel command on AIX, part of base operating system

salt/grains/core.py
dmurphy18: We are talking UNIX here, \n usage is similar to other usage in the file for UNIX system, for example: Solaris.  However I shall update to use os.linesep.

salt/grains/core.py
dmurphy18: Following style used for Solaris in the file.

salt/grains/core.py
dmurphy18: Following style used for _sunos_memdata, _bsd_memdata, etc. already used in the file.

salt/grains/core.py
dmurphy18: code changed
prefer to use similar commands across the various UNIX's as much as possible, Solaris is using swap.

salt/grains/core.py
dmurphy18: changed it anyway

gtmanfred (approved):

salt/grains/core.py
```diff
@@ -1759,6 +1853,16 @@ def os_data():
...
+        osrelease_techlevel = __salt__['cmd.run']('oslevel -r')
+        osname = __salt__['cmd.run']('uname')
+        grains['os'] = 'AIX'
+        grains['os_family'] = 'AIX'
```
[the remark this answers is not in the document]
dmurphy18: done

salt/grains/core.py
brejoc: Isn't there perhaps a third way to get the machine ID that doesn't have the potential to suddenly break with an update?

isbm: @dmurphy18 no, please. This is wrong. The `uname -m` does not give you a system ID, but the _hardware_ ID. Given AIX running on LPAR this simply won't work. Besides, it is deprecated. You should use `lsattr -El sys0` instead looking for system ID, which is typically appears as `os_uuid` if I remember it correctly after so many years.

salt/grains/core.py
```diff
@@ -567,22 +568,29 @@ def _aix_memdata():
...
-        comps = line.split(' ')
-        if comps[0].strip() == 'Memory' and comps[1].strip() == 'Size:':
-            grains['mem_total'] = int(comps[2].strip())
-            break
+    prtconf = salt.utils.path.which('prtconf')
+    if prtconf:
+        for line in __salt__['cmd.run'](prtconf, python_shell=True).splitlines():
+            comps = line.split(' ')
+            if comps[0].strip() == 'Memory' and comps[1].strip() == 'Size:':
+                grains['mem_total'] = int(comps[2].strip())
```
isbm: Like I already mentioned, this sort of parsing is very unreliable and fragile. You also constantly stripping things as you are unsure what is inside. My proposed version instead:
```python
comps = [x for x in line.strip().split(' ') if x]  # Or use filter() if you like
if len(comps) > 2 and 'Memory' in comps[0] and 'Size' in comps[1]:
    grains['mem_total'] = int(comps[2])

File: salt/grains/core.py [hunk 14]
```diff
@@ -408,60 +408,61 @@ def _aix_cpudata():
...
-    for regex in numcpus_regexes:
-        res = regex.search(data)
-        if res and len(res.groups()) >= 1:
-            grains['num_cpus'] = res.group(1).strip().replace("'", "")
-            break
+        for regex in numcpus_regexes:
+            res = regex.search(data)
+            if res and len(res.groups()) >= 1:
+                grains['num_cpus'] = res.group(1).strip().replace("'", "")
+                break
```
isbm: Like I already suggested earlier, this should be flattened into one loop instead of writing a very long repeated code. The example how to do this was also there but gone after you force-pushed.

isbm (changes requested):

salt/grains/core.py
```diff
@@ -399,6 +399,75 @@ def _sunos_cpudata():
...
+    #   cpuarch
+    #   num_cpus
+    #   cpu_model
+    #   cpu_flags
+    grains = {}
+    data = ""
+    cmd = salt.utils.path.which('prtconf')
+    if cmd:
+        data += __salt__['cmd.run']('{0}'.format(cmd))
+        data += '\n'
```
isbm: Please do not use `+=` here.

isbm (changes requested):

salt/grains/core.py
isbm: Please flatten this.

salt/grains/core.py
```diff
@@ -2434,6 +2548,52 @@ def _hw_data(osdata):
                 grains['product'] = res.group(1).strip().replace("'", "")
                 break
 
+    elif osdata['kernel'] == 'AIX':
+        data = ""
+        cmd = salt.utils.path.which('prtconf')
+        if cmd:
+            data += __salt__['cmd.run']('{0}'.format(cmd)) + os.linesep
+
```
isbm: Again `+=`. This is not needed here, especially that `data` is not used after `if cmd` returns `False`.

isbm (changes requested):

salt/grains/core.py
dmurphy18: The Solaris code from which the AIX support was derived had previously been limited to Python 2.6, hence no list comprehensions. Corrected now as per your suggestion, since base level is Python 2.7.

salt/grains/core.py
dmurphy18: done

dmurphy18:
@rallytime The codeclimate might need a little adjustment, running into issues, which are not part of the changes made, for example: _linux_cpudata, see https://codeclimate.com/github/saltstack/salt/salt/grains/core.py/source?from_sha=cf1e5d4c&to_sha=d2baabff

Would hate to have to refactor complete files to have some changes accepted.

salt/grains/core.py
isbm: @dmurphy18 are we speaking of the same thing? The unique machine ID we are about, is the ID of **the local system** that is set during installation or boot, not a reference to the hardware (!!). That said, every time an admin reinstalls an OS on the _same_ hardware, he still supposed to get a different ID. As well as if one installs AIX on LPAR. And so this gives the whole different tracking of the entities in your asset ledger.

Now, if you want to use `uname -m`, you returning not a _system ID_, but _hardware ID_. This is why: its format is `xxyyyyyymmss` for a reason, where `xx` and `ss` are always zeroes and `mm` almost always set to `4C` as it is just a  ID of the CPU Planar. The rest `yyyyyy` is a Model ID of the hardware. Just like in your own example above. So it is kind of unique per a hardware (what we **do not** need), but per a software (what we **do** need). Now, if you will install AIX on LPAR, it will no longer return _the unique_ ID (while it _absolutely_ has to be as such), because LPAR will reference the same hardware, and thus you will render the whole feature **broken**. I am sure you already read the official manual from IBM and found the remark about this regarding LPAR, didn't you? Then you already got to the details why it is so and you already figured out that this is because it is the hardware id, not an OS id. And so you probably by now understood that you're returning the wrong type and thus got to the conclusion that `uname -m` is a bad idea after all.

Regarding `lsattr -El sys0` by "buggy", I mean that sometimes in rare cases you would need to use a different device for this to fetch the `os_uuid`. But sorry I do not remember AIX details, as it was really long years ago... But the `os_uuid` is exactly what you are looking for, if you care Salt's feature work as it supposed to.

Like I also said, alternatively, we can simply help almighty IBM to generate `/etc/machine-id` just like systemd does, as simply as using Python's internals:
```python
import uuid
print(uuid.uuid4().hex)
```

rallytime:
@dmurphy18 We do not block on codeclimate violations right now, so you're good. It does need some tuning and it always runs into trouble with the core grains file. Mostly it's a helpful suggestion for new changes so we can start addressing tech debt.

salt/grains/core.py
dmurphy18: @isbm Checked with Daniel internally since he originated this code and got clarification on the intent of this value for machine id (given documentation did not state equivocally what it represented), and now understand it as a value representing the machine/virtualization, similar to systemd/nspawn.  Using this interpretation the code needs to change, and possibly for Solaris too.  Will alter AIX code to do as advised.

terminalmage (approved):

isbm (approved):
Thanks for all the refactoring and fixing this, @dmurphy18! And sorry for those 52 comments of heated-up discussion. But we've made it! :sunglasses:

Issue #45010, opened 2017-12-14
Add win_dns_client.add_dn windows Test to Auto Test Suite


Ch3LL:
This task is to monitor the addition of one test to the auto test suite:

- [ ] win_dns_client.add_dns

`salt Win64Builder win_dns_client.add_dns 4.2.2.2 Ethernet index=42`

PR #47823, opened 2018-05-24
Add win_dns module integration tests for windows

Ch3LL:
### What does this PR do?
Add integration tests for the win_dns module

https://github.com/saltstack/salt/issues/45010

### DIFF
--- a/tests/integration/modules/test_win_dns_client.py
+++ b/tests/integration/modules/test_win_dns_client.py
@@ -1,0 +1,37 @@
+# -*- coding: utf-8 -*-
+
+# Import Python libs
+from __future__ import absolute_import
+
+# Import Salt Testing libs
+from tests.support.case import ModuleCase
+from tests.support.unit import skipIf
+from tests.support.helpers import destructiveTest
+
+# Import Salt libs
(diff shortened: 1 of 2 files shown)


tests/integration/modules/test_win_dns_client.py
```diff
@@ -0,0 +1,37 @@
...
+    '''
+    Test for salt.modules.win_dns_client
+    '''
+    @destructiveTest
+    def test_add_remove_dns(self):
+        '''
+        Test add and removing a dns server
+        '''
+        dns = '8.8.8.8'
+        interface = 'Ethernet'
```
twangboy: This test may not work on all machines as it's impossible to know what the name of the interface will be. Depending on the version of Windows and what adapters are already present on the machine. Windows doesn't have the clear `eth0`, `eth1` thing like linux... Some system default to an interface name such as `Local Area Connection`. I think `Ethernet` is newer. It will probably work fine in our situation where we're spinning up a new VM on a modern version of Windows. I doubt it would work on a Windows 7 box for example.

twangboy (approved):

PR #47825, opened 2018-05-24
Allow environment variables to be set for Salt system services for Debian platforms

jeduardo:
### What does this PR do?

This PR allows each systemd service file managing a Salt service on Debian-based platforms to source environment variables from /etc/default/salt-<service>. (e.g. /etc/default/salt-minion, /etc/default/salt-master, etc)

This PR was created to enhance the behaviour of the Salt HTTP module when using the urllib backend on a system that is behind a proxy server, allowing the specification of the NO_PROXY environment variable.

Alternatively it also provides a way to pass configuration to any potential external command executed by a Salt minion on an environment that requires it. 

### Previous Behavior

No environment variables can be specified for a Salt system service.

### New Behavior
Every system service sources environment variables from a /etc/default/salt-<service> file.

### Tests written?

No

### Commits signed with GPG?

No


gtmanfred (commented):
That is not true.

You can specify all the environment variables that you want using systemd drop-in files.

create /etc/systemd/system/salt-minion.service.d/env.conf

then add

```
[Service]
Environment=PATH=/bin/
EnvironmentFile=/etc/default/salt-api
```

I don't think there is a need to add this, because we do not use those environment variables in the systemd service.

jeduardo:
@gtmanfred hey, thanks for the review. I was unaware of the existence of the drop-in files with systemd. I've just checked whether doing this would help with my problem and it sure did, the value of NO_PROXY is present in the environment and passed down to the urllib2 HTTP backend should I choose to use it. 

While I can see in the old upstart and init files some references to sourcing configuration from /etc/default/$service or /etc/sysconfig/$service (on RedHat systems) I think that the drop-in file you suggested fix the problem nicely on systemd-based systems in a way that I feel comfortable enough to not push to get this merged anymore.

PR #47828, opened 2018-05-25
Add allowing for wheel publish jobs

gtmanfred:
### What does this PR do?
This is required to allow minions to use the wheel clients on the master.

Specifically this will be useful with the direction of allowing minions to bootstrap other minions and then tell the master to accept the new minions key.

### Tests written?

Yes

### Commits signed with GPG?

Yes

### DIFF
--- a/doc/ref/peer.rst
+++ b/doc/ref/peer.rst
@@ -108,6 +108,12 @@ Peer Wheel Communication
 
 .. versionadded: Fluorine
 
+.. warning::
+
+    Any of the below configurations allow the minion to be able to control the
+    master.  Any minion that has the ability to use these functions will be
+    **fully trusted** by the master.
+
 Configuration to allow minions to execute wheel functions from the master is
 done via the ``peer_wheel`` option on the master. The ``peer_wheel``
 configuration follows the same logic as the ``peer`` option. The only
@@ -120,11 +126,6 @@ To open up access to all minions to all wheel:
     peer_wheel:
       .*:
         - .*
-
-.. warning::
-
-    The above configuration allows all the minions to control the master,
-    including changing master configuration files.
 
 This configuration will allow minions with IDs ending in example.com access
 to the key wheel functions.
--- a/doc/ref/peer.rst
+++ b/doc/ref/peer.rst
@@ -99,6 +99,36 @@ to the manage and jobs runner functions.
 .. note::
     Functions are matched using regular expressions.
 
+.. _peer-wheel:
+
+Peer Wheel Communication
+=========================
+
+Configuration to allow minions to execute wheel from the master is done via
+the ``peer_wheel`` option on the master. The ``peer_wheel`` configuration follows
+the same logic as the ``peer`` option. The only difference is that access is
+granted to wheel modules.
+
+To open up access to all minions to all wheel:
+
+.. code-block:: yaml
+
+    peer_wheel:
+      .*:
+        - .*
+
+This configuration will allow minions with IDs ending in example.com access
+to the key wheel functions.
+
+.. code-block:: yaml
+
+    peer_wheel:
+      .*example.com:
+        - key.*
+
+.. note::
+    Functions are matched using regular expressions.
+
 Using Peer Communication
 ========================
 
@@ -113,12 +143,6 @@ To execute test.ping on all minions:
 
     # salt-call publish.publish \* test.ping
 
-To execute the manage.up runner:
-
-.. code-block:: bash
-
-    # salt-call publish.runner manage.up
-
 To match minions using other matchers, use ``tgt_type``:
 
 .. code-block:: bash
@@ -127,3 +151,15 @@ To match minions using other matchers, use ``tgt_type``:
 
 .. note::
     In pre-2017.7.0 releases, use ``expr_form`` instead of ``tgt_type``.
+
+To execute the manage.up runner:
+
+.. code-block:: bash
+
+    # salt-call publish.runner manage.up
+
+To execute the key.accept wheel module:
+
+.. code-block:: bash
+
+    # salt-call publish.wheel key.accept arg='match=<minionid>'
--- a/doc/ref/peer.rst
+++ b/doc/ref/peer.rst
@@ -68,6 +68,7 @@ allow minions ending with foo.org access to the publisher.
         - pkg.*
 
 .. note::
+
     Functions are matched using regular expressions.
 
 Peer Runner Communication
@@ -97,6 +98,7 @@ to the manage and jobs runner functions.
         - jobs.*
 
 .. note::
+
     Functions are matched using regular expressions.
 
 .. _peer-wheel:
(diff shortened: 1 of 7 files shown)


gtmanfred:
re-run py

gtmanfred:
hrm, seems like py2 doesn't work, investigating.

salt/daemons/masterapi.py
```diff
@@ -836,6 +836,46 @@ def minion_runner(self, load):
...
+    def minion_wheel(self, load):
+        '''
+        Execute a wheel from a minion, return the wheel's function data
+        '''
+        if 'peer_wheel' not in self.opts:
+            return {}
+        if not isinstance(self.opts['peer_wheel'], dict):
+            return {}
+        if any(key not in load for key in ('fun', 'arg', 'id')):
+            return {}
```
isbm: @gtmanfred can we [avoid multiple returns](http://docs.python-guide.org/en/latest/writing/style/#returning-values), sticking to [SESE](https://en.wikipedia.org/wiki/Single-entry_single-exit) instead? Because other starts copying this and then the code isn't fun to maintain.

isbm (changes requested):

doc/ref/peer.rst
```diff
@@ -99,6 +99,36 @@ to the manage and jobs runner functions.
 .. note::
     Functions are matched using regular expressions.
 
+.. _peer-wheel:
+
+Peer Wheel Communication
+=========================
+
+Configuration to allow minions to execute wheel from the master is done via
```
rallytime: execute wheel `commands`? `functions`?

doc/ref/peer.rst
```diff
@@ -99,6 +99,36 @@ to the manage and jobs runner functions.
...
+the ``peer_wheel`` option on the master. The ``peer_wheel`` configuration follows
+the same logic as the ``peer`` option. The only difference is that access is
+granted to wheel modules.
+
```
rallytime: I think this section could use a note about the security implications of enabling this option. Similar to the Peer Configuration section.

doc/ref/peer.rst
```diff
@@ -99,6 +99,36 @@ to the manage and jobs runner functions.
...
+to the key wheel functions.
+
+.. code-block:: yaml
+
+    peer_wheel:
+      .*example.com:
+        - key.*
+
+.. note::
+    Functions are matched using regular expressions.
```
rallytime: I think you need a blank line below the `.. note::` for this to display as a proper "note" section.

doc/ref/peer.rst
```diff
@@ -113,17 +143,23 @@ To execute test.ping on all minions:
...
     # salt-call publish.publish \* test.ping
 
+To match minions using other matchers, use ``tgt_type``:
+
+.. code-block:: bash
+
+    # salt-call publish.publish 'webserv* and not G@os:Ubuntu' test.ping tgt_type='compound'
+
+.. note::
+    In pre-2017.7.0 releases, use ``expr_form`` instead of ``tgt_type``.
```
rallytime: Same comment here about the `note` syntax.

rallytime (commented):
I have a couple of doc requests, otherwise this LGTM. 👍

doc/ref/peer.rst
rallytime: Sure, but it wouldn't hurt to make that extra clear.

cachedout:
This has some pretty intense security implications. I see the benefit here and I understand that it has to be explicitly configured but I'm worried about what we might be getting ourselves into here.

I'd be interested in what @thatch45 has to say here.

thatch45 (approved):

doc/topics/releases/fluorine.rst
```diff
@@ -157,6 +157,16 @@ The syntax is as follows:
...
     - :py:mod:`zypper <salt.modules.zypper>`
 
+Wheel Peering Options
+=====================
+
+A ``publish.wheel`` module has been added, allowing access to the wheel
+functions on the master.
+
+The peer configuration can be found :ref:`here <peer-wheel>`.
+>>>>>>> add docs
```
rallytime: Left over from a merge conflict. :)

doc/ref/peer.rst
```diff
@@ -97,6 +98,40 @@ to the manage and jobs runner functions.
...
+=========================
+
+.. versionadded: Fluorine
+
+Configuration to allow minions to execute wheel functions from the master is
+done via the ``peer_wheel`` option on the master. The ``peer_wheel``
+configuration follows the same logic as the ``peer`` option. The only
+difference is that access is granted to wheel modules.
+
+To open up access to all minions to all wheel:
```
rallytime: I still think there should be a more defined security note about the implications of this setup.

thatch45: Yes, very much so! We need to be clear that this allows the minion to alter the master configuration etc.

gtmanfred: done

rallytime (approved):

gtmanfred:
I think we are just waiting on @cachedout to rereview this one?

Thanks,
Daniel

doc/ref/peer.rst
```diff
@@ -97,6 +98,45 @@ to the manage and jobs runner functions.
...
+
+.. code-block:: yaml
+
+    peer_wheel:
+      .*:
+        - .*
+
+.. warning::
+
+    The above configuration allows all the minions to control the master,
```
cachedout: I have to be particular about the way this is phrased. What this says is that *this* configuration allows all minions to control the configuration files of the master. I think we need to be even more clear that _any_ minion that is allowed to control wheel functions on the master is effectively being *fully trusted* by the master.

doc/topics/releases/fluorine.rst
```diff
@@ -180,6 +180,15 @@ The syntax is as follows:
     - :py:mod:`yumpkg <salt.modules.yumpkg>`
     - :py:mod:`zypper <salt.modules.zypper>`
 
+Wheel Peering Options
+=====================
+
+A ``publish.wheel`` module has been added, allowing access to the wheel
+functions on the master.
```
cachedout: Again, I think we need to be extremely explicit about what this means from a security perspective. I am already extremely wary of this feature and I want us to go way overboard on making sure people understand what turning this on would do.

salt/daemons/masterapi.py
```diff
@@ -836,6 +836,46 @@ def minion_runner(self, load):
...
+                if isinstance(self.opts['peer_wheel'][match], list):
+                    perms.update(self.opts['peer_wheel'][match])
+        good = False
+        for perm in perms:
+            if re.match(perm, load['fun']):
+                good = True
+        if not good:
+            # The minion is not who it says it is!
+            # We don't want to listen to it!
+            log.warning('Minion id %s is not who it says it is!', load['id'])
```
cachedout: Eh? This is a perms problem and not an identify problem, isn't it?

cachedout (changes requested):
I left some additional comments. For the record, I'm really worried about this feature. I think the security risk is far too high and that the benefit is far too low. I don't believe we should have a security model that works like this at all.

doc/ref/peer.rst
gtmanfred: I tried to make the warning better.

doc/topics/releases/fluorine.rst
gtmanfred: I added a warning to this.

salt/daemons/masterapi.py
gtmanfred: yup, fixed.

gtmanfred:
This is too much of a security issue.

I am going to close this issue, and work on a smaller scoped change that just allows designated minions to pass keys for new minions back to the master.

Issue #47794, opened 2018-05-23
salt-cloud azurearm: VM deployed to network_resource_group


cskowronnek:
### Description of Issue/Question
With the latest develop code my VM is deployed to the network_resource_group. I use different resource_group and network_resource_group because we have the VNET in the NRG. After deployment, all my resources end up in the network_resource_group instead of the resource_group.
### Setup
(Please provide relevant configs and/or SLS files (Be sure to remove sensitive info).)

### Steps to Reproduce Issue
(Include debug logs if possible and relevant.)

### Versions Report
(Provided by running `salt --versions-report`. Please also mention any differences in master/minion versions.)

```
Salt Version:
           Salt: 2018.3.0

Dependency Versions:
           cffi: 1.11.2
       cherrypy: Not Installed
       dateutil: 2.6.1
      docker-py: Not Installed
          gitdb: 0.6.4
      gitpython: 1.0.1
          ioflo: Not Installed
         Jinja2: 2.8
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: 1.0.3
   msgpack-pure: Not Installed
 msgpack-python: 0.4.6
   mysql-python: Not Installed
      pycparser: 2.18
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: Not Installed
         Python: 2.7.12 (default, Dec  4 2017, 14:50:18)
   python-gnupg: Not Installed
         PyYAML: 3.11
          PyZMQ: 15.2.0
           RAET: Not Installed
          smmap: 0.9.0
        timelib: Not Installed
        Tornado: 4.2.1
            ZMQ: 4.1.4

System Versions:
           dist: Ubuntu 16.04 xenial
         locale: UTF-8
        machine: x86_64
        release: 4.4.0-121-generic
         system: Linux
        version: Ubuntu 16.04 xenial
```

Ch3LL:
ping @saltstack/team-cloud any ideas here?

nicholasmhughes:
I just got pinged on this off GitHub yesterday... I'm looking into it right now. It seems like the logic there has been jacked for a while. I'm on it.

nicholasmhughes:
@cskowronnek , can you test this patch (created off the latest develop code):
```
diff --git a/salt/cloud/clouds/azurearm.py b/salt/cloud/clouds/azurearm.py
index 336cb9c115..336c72de92 100644
--- a/salt/cloud/clouds/azurearm.py
[...]

Issue #47506, opened 2018-05-07
Profile Error when trying to generate an Azure VM from GCP


justin-w-long:
### Description of Issue/Question
My Salt Master is in GCP and I am trying to get vm generation in Azure to work from a GCP master. I am able to communicate with Azure and it begins the process of spinning up the VM but it fails during the creation with a profile error.

### Setup
Provider Configuration:
```
azure-ubuntu:
  driver: azurearm
  subscription_id: SUBSCRITION_ID

  tenant: TENANT_ID
  client_id: CLIENT_ID
  secret: MY_SECRET

  location: eastus
  resource_group: AzureDemo
  network_resource_group: AzureDemo
  storage_account: MY_STORAGE_ACCOUNT
  storage_key: MY_STORAGE_KEY

  cleanup_disks: True
  cleanup_vhds: True
  cleanup_data_disks: True
  cleanup_interfaces: True

  # Set up the location of the salt master
  #
  minion:
    master: MASTER_IP
```
Profile Configuration:
```
azure-ubuntu:
  provider: azure-ubuntu
  image: Canonical|UbuntuServer|18.04-LTS|18.04.201804262
  size: Standard_DS1
  ssh_username: MY_USERNAME
  ssh_password: MY_PASSWORD
  location: eastus

  network: AzureDemo-Network
  public_ip: False
```

### Steps to Reproduce Issue
Run the command: "salt-cloud -l debug -p azure-ubuntu test-demo"

[ERROR   ] There was a profile error: __init__() takes exactly 1 argument (2 given)
Traceback (most recent call last):
  File "/usr/lib/python2.7/dist-packages/salt/cloud/cli.py", line 281, in run
    self.config.get('names')
  File "/usr/lib/python2.7/dist-packages/salt/cloud/__init__.py", line 1454, in run_profile
    ret[name] = self.create(vm_)
  File "/usr/lib/python2.7/dist-packages/salt/cloud/__init__.py", line 1284, in create
    output = self.clouds[func](vm_)
  File "/usr/lib/python2.7/dist-packages/salt/cloud/clouds/azurearm.py", line 1279, in create
    'wait_for_ip_interval_multiplier', vm_, __opts__, default=1),
  File "/usr/lib/python2.7/dist-packages/salt/utils/cloud.py", line 2389, in wait_for_ip
    data = update_callback(*update_args, **update_kwargs)
[...]

PR #47839, opened 2018-05-25
Fix Azure RM cloud module network_resource_group handling

nicholasmhughes:
### What does this PR do?
Fixes the behavior of `network_resource_group` to ensure that the virtual network is expected to be in that resource group while the instance and all associated objects created by the cloud module end up in `resource_group`.

 #47794 

### Previous Behavior
Details in the issue.

### New Behavior
The virtual network is expected to be in `network_resource_group` while the instance and all associated objects created by the cloud module end up in `resource_group`.

### Tests written?
No

### Commits signed with GPG?
Yes

### DIFF
--- a/doc/topics/cloud/azurearm.rst
+++ b/doc/topics/cloud/azurearm.rst
@@ -235,9 +235,10 @@ etc) will be created in.
 
 network_resource_group
 ----------------------
-Optional. If specified, then the VM will be connected to the network resources
-in this group, rather than the group that it was created in. The VM interfaces
-and IPs will remain in the configured ``resource_group`` with the VM.
+Optional. If specified, then the VM will be connected to the virtual network
+in this resource group, rather than the parent resource group of the instance.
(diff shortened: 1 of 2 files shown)


rallytime (approved):

gtmanfred:
@nicholasmhughes does this need to be backported to 2018.3?

Thanks,
Daniel

nicholasmhughes:
Probably... it seemed like the `network_resource_group` logic was broken for a while. However, the capabilities are pretty different between `2018.3` and `develop`. The exact same code changes might not work. That's what [prompted me to ask](https://github.com/saltstack/salt/issues/47506#issuecomment-387114357) when we might see the `azurearm` module merge from `develop` into a stable release.

gtmanfred:
We do not merge new features into stable releases, we do releases from develop.

So the next major release will have the azurearm modules in it.

Are you on the saltstack community slack?  I have something that I have been working on for a minute. That might be of interest to you.

Issue #30367, opened 2016-01-14
salt-run orchestration via state.sls always successful; it is ignoring state return value


paulfanelli:
In the process of writing some salt orchestration, no matter which options I used, the salt-run orchestration state.sls module/function always ignores the pass/fail return output from the state (that runs on the minion) and the salt-run orchestration is always successful.

I'm using the latest salt released version - 2015.8.3, on both the master and minion (all version info at the bottom).

I first wrote some pass and fail state code:

``` yaml
# /srv/salt/utils/fail.sls
cmd-run-false:
  cmd.run:
    - name: /bin/false
```

``` yaml
# /srv/salt/utils/pass.sls
cmd-run-true:
  cmd.run:
    - name: /bin/true
```

I next wrote some test orchestration code using state.function:

``` yaml
# /srv/salt/orch/test/orch-by-func-true.sls
orch-by-func-true:
  salt.function:
    - name: cmd.run
    - tgt: mfg-test-iws1
    - arg:
      - 'true'
```

Here is the test run and output for orch-by-func-true.sls:

``` bash
[root@salt test]# salt-run state.orch orch.test.orch-by-func-true
salt.corp.mutualink.net_master:
----------
          ID: orch-by-func-true
    Function: salt.function
        Name: cmd.run
      Result: True
     Comment: Function ran successfully. Function cmd.run ran on mfg-test-iws1.
     Started: 18:39:34.214854
    Duration: 396.539 ms
     Changes:
              mfg-test-iws1:


Summary for salt.corp.mutualink.net_master
------------
Succeeded: 1 (changed=1)
Failed:    0
------------
Total states run:     1
Total run time: 396.539 ms
```

``` yaml
# /srv/salt/orch/test/orch-by-func-false.sls
orch-by-func-false:
  salt.function:
    - name: cmd.run
    - tgt: mfg-test-iws1
    - arg:
      - 'false'
```

Here is the test run and output for orch-by-func-false.sls:

``` bash
[root@salt test]# salt-run state.orch orch.test.orch-by-func-false
[ERROR   ] {'ret': {'mfg-test-iws1': ''}, 'out': 'highstate'}
salt.corp.mutualink.net_master:
----------
          ID: orch-by-func-false
    Function: salt.function
        Name: cmd.run
      Result: False
     Comment: Running function cmd.run failed on minions: mfg-test-iws1 Function cmd.run ran on mfg-test-iws1.
     Started: 18:42:51.675954
    Duration: 397.494 ms
[...]

PR #47843, opened 2018-05-25
[2018.3] orchestration results False when function is False

garethgreenaway:
### What does this PR do?
Updating function in saltmod to ensure that the result is a failure if the function being run returns as False.

#30367 

### Previous Behavior
Previously functions that returned False would still return in a True result in the orchestration results.

### New Behavior
If the function being run in orchestration returns as False then the orchestration result is False.

### Tests written?
Yes

### Commits signed with GPG?

Yes/No

### DIFF
--- a/salt/states/saltmod.py
+++ b/salt/states/saltmod.py
@@ -581,6 +581,9 @@ def function(
             m_ret = mdata['ret']
             m_func = (not fail_function and True) or __salt__[fail_function](m_ret)
 
+            if m_ret is False:
+                m_func = False
+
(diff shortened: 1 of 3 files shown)


gtmanfred (approved):

cachedout:
LGTM but I would like to see some independent confirmation on this fix from the people in the OP issue.

rallytime:
This was confirmed to be working in https://github.com/saltstack/salt/issues/30367#issuecomment-392937110

👍

PR #47851, opened 2018-05-26
Fixup! add master.py:FileserverUpdate **kwargs

rares-pop:
Fixup! add master.py:FileserverUpdate **kwargs

All classes inherited from salt/utils/process.py:MultiprocessingProcess 
must use **kwargs after commit 90edc69

### What does this PR do?
It fixes the master on Windows platform.

### Previous Behavior
salt-master was broken

### Tests written?

No

### Commits signed with GPG?

No

### DIFF
--- a/salt/master.py
+++ b/salt/master.py
@@ -349,8 +349,8 @@ class FileserverUpdate(salt.utils.process.SignalHandlingMultiprocessingProcess):
     '''
     A process from which to update any dynamic fileserver backends
     '''
-    def __init__(self, opts, log_queue=None):
-        super(FileserverUpdate, self).__init__(log_queue=log_queue)
+    def __init__(self, opts, **kwargs):
+        super(FileserverUpdate, self).__init__(**kwargs)
         self.opts = opts
         self.update_threads = {}
         # Avoid circular import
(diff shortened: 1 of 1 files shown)


salt/master.py
```diff
@@ -349,8 +349,8 @@ class FileserverUpdate(salt.utils.process.SignalHandlingMultiprocessingProcess):
     '''
     A process from which to update any dynamic fileserver backends
     '''
-    def __init__(self, opts, log_queue=None):
-        super(FileserverUpdate, self).__init__(log_queue=log_queue)
+    def __init__(self, opts, **kwargs):
```
gtmanfred: I do not like the idea of using kwargs for something like this, especially without documenting what can be passed as kwargs.

gtmanfred (changes requested):
One major grip.

Can you also provide us with more information than 'FIXUP! add master.py:FileserverUpdate **kwargs`  This doesn't tell us anything on why it is needed or what is "broken"

salt/master.py
rares-pop: All classes inherited from salt/utils/process.py:MultiprocessingProcess must use **kwargs after commit 90edc697c2a2dd21fbc5f7a139139e227fd025eb

gtmanfred: Perfect, thank you for the explanation.  Now it makes sense on why you are using `**kwargs`

gtmanfred (approved):

salt/master.py
rares-pop: Sure! I'll add the explanation in the description. Sorry for the misunderstanding.

gtmanfred: Awesome!  Thank you very much for this addition!

Daniel

terminalmage (approved):
@rallytime the `FileserverUpdate` process was added in 2018.3, so this will need a backport.

rares-pop:
As a followup - should I backport this to 2018.3? I can totally do that.

gtmanfred:
It is marked to be backported, we will backport it.

PR #47814, opened 2018-05-24
Bugfixes and unit tests for pkgin module

travispaul:
Corrects pkg.latest_version and pkg.file_dict behavior

### What does this PR do?

Fixes 2 bugs in the pkgin module:
- pkg.latest_version doesn't return a version for an uninstalled package.
- pkg.file_dict crashes.

### What issues does this PR fix or reference?

None that I have found.

### Previous Behavior

**pkg.latest_version**

pkgin reports that nginx version 1.14.0 is available but not yet installed:
```
salt# pkgin se nginx
nginx-1.14.0         Lightweight HTTP server and mail proxy server

=: package is installed and up-to-date
<: package is installed but newer version is available
: installed package has a greater version than available package
```

However, pkg.latest_version returns nothing:
```
salt# salt \* pkg.latest_version nginx
salt.n7test.local:

```

**pkg.file_dict**

After installing a package, pkg.file_dict throws an exception:
 
```
# salt \* pkg.install nginx
salt.n7-test.local:
   ----------
   nginx:
       ----------
       new:
           1.14.0
       old:
   pcre:
       ----------
       new:
           8.42
       old:

salt# pkg_info -qL nginx
/opt/pkg/current/man/man8/nginx.8
/opt/pkg/current/sbin/nginx
/opt/pkg/current/share/examples/nginx/conf/fastcgi.conf
/opt/pkg/current/share/examples/nginx/conf/fastcgi_params
/opt/pkg/current/share/examples/nginx/conf/koi-utf
/opt/pkg/current/share/examples/nginx/conf/koi-win
/opt/pkg/current/share/examples/nginx/conf/mime.types
/opt/pkg/current/share/examples/nginx/conf/nginx.conf
/opt/pkg/current/share/examples/nginx/conf/win-utf
/opt/pkg/current/share/examples/nginx/html/50x.html
/opt/pkg/current/share/examples/nginx/html/inde

salt# salt \* pkg.file_dict nginx

salt.n7-test.local:
   The minion function caused an exception: Traceback (most recent call last):
     File "/opt/pkg/current/lib/python2.7/site-packages/salt/minion.py", line 1600, in _thread_return
       return_data = minion_instance.executors[fname](opts, data, func, args, kwargs)
     File "/opt/pkg/current/lib/python2.7/site-packages/salt/executors/direct_call.py", line 12, in execute
       return func(*args, **kwargs)
[...]

PR #47600, opened 2018-05-10
Prevent crash on NetBSD and OpenBSD when no swap is configured.

travispaul:
### What does this PR do?

Prevents the ``salt`` and ``salt-master`` commands from crashing on NetBSD and OpenBSD when no swap space is configured.

### What issues does this PR fix or reference?
I am not aware of an existing open issue.

### Previous Behavior
On a NetBSD 7.1.2 system with no swap configured:

```
salt# swapctl -sk
no swap devices configured

salt# salt-master --help
Traceback (most recent call last):
  File "/opt/pkg/bin/salt-master", line 22, in <module>
    salt_master()
  File "/opt/pkg/lib/python2.7/site-packages/salt/scripts.py", line 95, in salt_master
    import salt.cli.daemons
  File "/opt/pkg/lib/python2.7/site-packages/salt/cli/daemons.py", line 48, in <module>
    import salt.utils.parsers
  File "/opt/pkg/lib/python2.7/site-packages/salt/utils/parsers.py", line 27, in <module>
    import salt.config as config
  File "/opt/pkg/lib/python2.7/site-packages/salt/config/__init__.py", line 98, in <module>
    _DFLT_IPC_WBUFFER = _gather_buffer_space() * .5
  File "/opt/pkg/lib/python2.7/site-packages/salt/config/__init__.py", line 91, in _gather_buffer_space
    grains = salt.grains.core._memdata(os_data)
  File "/opt/pkg/lib/python2.7/site-packages/salt/grains/core.py", line 509, in _memdata
    grains.update(_bsd_memdata(osdata))
  File "/opt/pkg/lib/python2.7/site-packages/salt/grains/core.py", line 458, in _bsd_memdata
    grains['swap_total'] = int(swap_total) // 1024 // 1024
ValueError: invalid literal for int() with base 10: 'swap'
```

### New Behavior

Salt commands work without crashing.

### Tests written?

No

### Commits signed with GPG?

Yes

PR #47866, opened 2018-05-29
Back-port #47814 to 2018.3

rallytime:
Back-port #47814 to 2018.3


cachedout (approved):

travispaul:
#47814 and #47600 have been absent from both the 2018.3.1 and 2018.3.2 releases. When can we expect these fixes to make it into the next release? I'm maintaining patches for these in pkgsrc and trying to gauge when they will no longer be needed.

Thanks

PR #47871, opened 2018-05-29
Fixing "installed:" in pkg.info

bochi:
In some cases, ie. for packages that have been installed as dependency, zypper info's output will be "Installed      : Yes (automatically)" which will result in "Installed: False" when using pkg.info

Random example:

sles12sp3-7:~ # zypper info xev
Loading repository data...
Reading installed packages...

Information for package xev:
----------------------------
Repository     : @System
Name           : xev
Version        : 1.2.1-2.15
Arch           : x86_64
Vendor         : SUSE LLC <https://www.suse.com/>
Support Level  : unknown
Installed Size : 49.9 KiB
Installed      : Yes (automatically)

suma31:~ # salt sles12sp3-7 pkg.info xev
sles12sp3-7:
    ----------
    xev:
        ----------
        arch:
            x86_64
        installed:
            False

### Previous Behavior
pkg.info shows wrong information for some packages

### New Behavior
pkg.info shows correct information for these packages ;)

### Tests written?

No

### Commits signed with GPG?

No

### DIFF
--- a/salt/modules/zypper.py
+++ b/salt/modules/zypper.py
@@ -555,7 +555,7 @@ def info_available(*names, **kwargs):
         if nfo.get('status'):
             nfo['status'] = nfo.get('status')
         if nfo.get('installed'):
-            nfo['installed'] = nfo.get('installed').lower() == 'yes' and True or False
+            nfo['installed'] = nfo.get('installed').lower().startswith( 'yes' ) and True or False
 
     return ret
 
--- a/salt/modules/zypper.py
+++ b/salt/modules/zypper.py
@@ -555,7 +555,7 @@ def info_available(*names, **kwargs):
         if nfo.get('status'):
             nfo['status'] = nfo.get('status')
         if nfo.get('installed'):
-            nfo['installed'] = nfo.get('installed').lower().startswith( 'yes' ) and True or False
+            nfo['installed'] = nfo.get('installed').lower().startswith('yes')
 
     return ret
 


salt/modules/zypper.py
```diff
@@ -555,7 +555,7 @@ def info_available(*names, **kwargs):
         if nfo.get('status'):
             nfo['status'] = nfo.get('status')
         if nfo.get('installed'):
-            nfo['installed'] = nfo.get('installed').lower() == 'yes' and True or False
+            nfo['installed'] = nfo.get('installed').lower().startswith( 'yes' ) and True or False
```
cachedout: Isn't the half of this statement after the `and` now redundant?

terminalmage: Yes, `startswith()` already returns a bool.

damon-atkins:
Hi Guys,
pkg needs to be consistent across platforms as much as possible.
If you add a new attribute it should be added to all package managers or prefixed with a platform specific key. e.g. win_ or suse_ To indicate it will not be available anywhere else.

Something like status would be more appropriate if it was going to be across the board.
`status: installed` or `status: delay-installed` or something.

The other choice is to report it not installed. i.e. ignore it.   Does pkgs_list report it.

For example Windows also has first time use install.  e.g. Have Word in your start menu and click on it and word starts installing and then word starts running.

isbm (approved):

isbm:
@rallytime this is good to go.

PR #47899, opened 2018-05-30
salt-jenkins issue 1000

terminalmage:
This adds logging of the output and retcode when the output file fails to be created. As I have thus far been unable to reproduce this locally, this is the best shot at seeing what is causing the issue, since ShellCase's run_script is rather opaque about what happens in the script itself. We do know from the salt-runtests.log that the `salt-call -g` is successfully returning, what we don't know is what happened after that data was returned, and this should provide that insight.


terminalmage:
re-run py2

rallytime (approved):

PR #47909, opened 2018-05-31
Fix salt.states.pkg.latest for IPS packages on Solaris

gmsoft-tuxicoman:
### What does this PR do?
Fix salt.states.pkg.latest for IPS packages on Solaris.

None.

### Previous Behavior
Currently the state function pkg.latest is broken with the solarisips provider because the modules functions solarisips.install() and solarisips.latest_version() don't return the expected value in some cases.

### New Behavior
The module install() now returns an empty dict if the package is already installed.
The latest_version() now returns an empty string for packages already at the latest version.
The version() function ignores the return code of pkg list to avoid errors when checking the version of non installed packages.

### Tests written?

No

### Commits signed with GPG?

No

### DIFF
--- a/salt/modules/solarisips.py
+++ b/salt/modules/solarisips.py
@@ -466,7 +466,7 @@ def install(name=None, refresh=False, pkgs=None, version=None, test=False, **kwa
     '''
     if not pkgs:
         if is_installed(name):
-            return 'Package already installed.'
+            return {}
 
     if refresh:
         refresh_db(full=True)
--- a/salt/modules/solarisips.py
+++ b/salt/modules/solarisips.py
@@ -355,6 +355,10 @@ def latest_version(*names, **kwargs):
 
     installed = version(*names)
 
+    if len(names) == 1:
+        # Convert back our result in a dict if only one name is passed
+        installed = { list(ret.keys())[0] if len(ret) > 0 else names[0]: installed }
+
     for name in ret:
         if not name in installed:
             continue
--- a/salt/modules/solarisips.py
+++ b/salt/modules/solarisips.py
@@ -310,7 +310,7 @@ def version(*names, **kwargs):
         ret[_ips_get_pkgname(line)] = _ips_get_pkgversion(line)
 
     # Append package names which are not installed/found
-    unmatched = list(filter(lambda name: not reduce(lambda x, y : x or name in y, ret.keys(), False), names))
+    unmatched = list(filter(lambda name: not reduce(lambda x, y : x or name in y, ret), False), names))
     ret.update(zip(unmatched, [ '' ] * len(unmatched)))
 
     # Return a string if only one package name passed
@@ -357,7 +357,7 @@ def latest_version(*names, **kwargs):
 
     if len(names) == 1:
         # Convert back our result in a dict if only one name is passed
-        installed = { list(ret.keys())[0] if len(ret) > 0 else names[0]: installed }
+        installed = { list(ret)[0] if len(ret) > 0 else names[0]: installed }
 
     for name in ret:
         if not name in installed:
@@ -366,7 +366,7 @@ def latest_version(*names, **kwargs):
             ret[name] = ''
 
     # Append package names which are not found
-    unmatched = list(filter(lambda name: not reduce(lambda x, y : x or name in y, ret.keys(), False), names))
+    unmatched = list(filter(lambda name: not reduce(lambda x, y : x or name in y, ret, False), names))
     ret.update(zip(unmatched, [ '' ] * len(unmatched)))
 
     # Return a string if only one package name passed
--- a/salt/modules/solarisips.py
+++ b/salt/modules/solarisips.py
@@ -332,7 +332,7 @@ def latest_version(*names, **kwargs):
     Please use pkg.latest_version as pkg.available_version is being deprecated.
 
     .. versionchanged:: Fluorine
-        Argument changed from 'name' to '*names' tu support multiple packages.
+        Support for multiple package names added.
 
     CLI Example:
 
--- a/salt/modules/solarisips.py
+++ b/salt/modules/solarisips.py
@@ -314,7 +314,7 @@ def version(*names, **kwargs):
     ret.update(zip(unmatched, [ '' ] * len(unmatched)))
 
     # Return a string if only one package name passed
-    if len(ret) == 1:
+    if len(names) == 1:
         return list(ret.values())[0]
 
     return ret
@@ -366,7 +366,7 @@ def latest_version(*names, **kwargs):
     ret.update(zip(unmatched, [ '' ] * len(unmatched)))
 
     # Return a string if only one package name passed
-    if len(ret) == 1:
+    if len(names) == 1:
         return list(ret.values())[0]
 
     return ret
--- a/salt/modules/solarisips.py
+++ b/salt/modules/solarisips.py
@@ -314,7 +314,7 @@ def version(*names, **kwargs):
         ret[_ips_get_pkgname(line)] = _ips_get_pkgversion(line)
 
     # Append package names which are not installed/found
-    unmatched = list(filter(lambda name: not reduce(lambda x, y : x or name in y, ret), False), names))
+    unmatched = list(filter(lambda name: not reduce(lambda x, y : x or name in y, ret, False), names))
     ret.update(zip(unmatched, itertools.cycle(('',))))
 
     # Return a string if only one package name passed
--- a/salt/modules/solarisips.py
+++ b/salt/modules/solarisips.py
@@ -314,8 +314,8 @@ def version(*names, **kwargs):
     ret.update(zip(unmatched, [ '' ] * len(unmatched)))
 
     # Return a string if only one package name passed
-    if len(names) == 1:
-        return ret[names[0]]
+    if len(ret) == 1:
+        return list(ret.values())[0]
 
     return ret
 
@@ -366,8 +366,8 @@ def latest_version(*names, **kwargs):
     ret.update(zip(unmatched, [ '' ] * len(unmatched)))
 
     # Return a string if only one package name passed
-    if len(names) == 1:
-        return ret[names[0]]
+    if len(ret) == 1:
+        return list(ret.values())[0]
 
     return ret
 
--- a/salt/modules/solarisips.py
+++ b/salt/modules/solarisips.py
@@ -342,9 +342,11 @@ def latest_version(*names, **kwargs):
         if ret[name] == installed[name]:
             ret[name] = ''
 
-    if ret:
-        return ret
-    return ''
+    # Return a string if only one package name passed
+    if len(names) == 1:
+        return ret[names[0]]
+
+    return ret
 
 # available_version is being deprecated
 available_version = salt.utils.functools.alias_function(latest_version, 'available_version')
--- a/salt/modules/solarisips.py
+++ b/salt/modules/solarisips.py
@@ -288,6 +288,7 @@ def version(*names, **kwargs):
     '''
     Common interface for obtaining the version of installed packages.
     Accepts full or partial FMRI. If called using pkg_resource, full FMRI is required.
+    Partial FMRI is returned if the package is not installed.
 
     CLI Example:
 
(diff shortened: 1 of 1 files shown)


damon-atkins:
Some examples
```
salt-call pkg.latest_version glibc abc123 --out json
{
    "local": {
        "glibc": "2.26-28.fc27",
        "abc123": ""
    }
}
salt-call pkg.latest_version glibc --out json
{
    "local": "2.26-28.fc27"
}
salt-call pkg.latest_version abc123 --out json
{
    "local": ""
}
```

latest_version:
    Returns a string representing the package version or an empty string if not
    installed. If more than one package name is specified, a dict of
    name/version pairs is returned. (if a package not installed the dict value should also contain empty string for that package name)
```
# salt-call pkg.version glibc abc123 --out json
{
    "local": {
        "glibc": "2.26-27.fc27",
        "abc123": ""
    }
}
salt-call pkg.version abc123 --out json
{
    "local": ""
}
```

dmurphy18 (changes requested):
The documentation for latest_version should probably have a versionchanged line since the definition is changed, and an improved example.

salt/modules/solarisips.py
```diff
@@ -301,7 +301,7 @@ def version(*names, **kwargs):
     if names:
         cmd = ['/bin/pkg', 'list', '-Hv']
         cmd.extend(names)
-        lines = __salt__['cmd.run_stdout'](cmd).splitlines()
+        lines = __salt__['cmd.run_stdout'](cmd, ignore_retcode=True).splitlines()
```
terminalmage: Can you exmplain why you added this? `ignore_retcode` just suppresses Salt's normal behavior of logging an error (as well as the command's output) when a command returns a non-zero exit status. Does this command return a non-zero exit status in cases where there is no error?

gmsoft-tuxicoman: Yes, as described in the New Behavior of the PR : 
The version() function ignores the return code of pkg list to avoid errors when checking the version of non installed packages.

This is because lastest_version() calls version(). The package might not be installed and version will error out.

gmsoft-tuxicoman:
@dmurphy18 What should the 'versionchanged' line look like ?

Please let me know and I'll commit an update with updated doc.

damon-atkins:
If software package name does not exist (even in the remote repo) like the abc123 example you output package version as an empty string.

damon-atkins:
versionchanged  grep the other files. Heaps of examples.

dmurphy18:
@gmsoft-tuxicoman 

.. versionchanged:: Fluorine

example taken from salt/runners/state.py

Fluorine since that is the code name for the next major release.  Salt is working its way through the Periodic Table.

gmsoft-tuxicoman:
@damon-atkins 
From aptpkg.py I understood that if the latest version was already installed, an empty string should be returned :

> Return the latest version of the named package available for upgrade or
> installation. If more than one package name is specified, a dict of
> name/version pairs is returned.
> If the latest version of a given package is already installed, an empty
> string will be returned for that package.


Also on solaris, package names usually have to be resolved. When you specify 'sendmail', you are actually talking about 'pkg://solaris/service/network/smtp/sendmail'. If I supply 'abc123', the pkg command won't be able to resolve the package name and will not output anything.

According the documentation of salt.modules.solarisips, the function is only returning matched package :
> The available version of the package in the repository. In case of multiple matches, it returns list of all matched packages. Accepts full or partial FMRI.

damon-atkins:
In the examples provided abc123 package does not exist.  Although RHEL and Solaris IPS have different package names the output data structure needs to be the same.  If more than one package name is provided, a dict must be return even if only one package name is valid. If only one package name is provided a string is returned.

salt/modules/solarisips.py
```diff
@@ -366,8 +366,8 @@ def latest_version(*names, **kwargs):
     ret.update(zip(unmatched, [ '' ] * len(unmatched)))
 
     # Return a string if only one package name passed
-    if len(names) == 1:
-        return ret[names[0]]
+    if len(ret) == 1:
```
damon-atkins: This is wrong.   The number of results does not determine the output.   The number of inputs determines the output type.   One Package name provided return a string, more than one package name return a dict. The len(name) == 1 was correct.

damon-atkins (changes requested):

damon-atkins (changes requested):

damon-atkins:
Looks like you need to use `pkg list -Hvfa`  which will give you all pkg names and all versions. Cache this as a pkg and version available lookup table for every package.  Which would allow you to do wildcard lookup and latest version lookup.

gmsoft-tuxicoman:
There is no need to use a wildcard. You can simply specify 'sendmail' and the package will be installed. So I don't think caching all the version will be useful.

I've also silenced out the retcode check in latest_version() to make it consistent with version().

Here is the output with different test cases. The package sendmail is current and installed on the system, postfix is not installed and abc123 of course doesn't exists :`

```
# salt-call pkg.latest_version sendmail postfix abc123 --out json
{
    "local": {
        "pkg://solaris/service/network/smtp/sendmail": "", 
        "pkg://solaris/service/network/smtp/postfix": "3.2.2,5.11-0.175.3.23.0.1.0:20170712T212555Z", 
        "abc123": ""
    }
}
```

```
# salt-call pkg.version sendmail postfix abc123 --out json
{
    "local": {
        "pkg://solaris/service/network/smtp/sendmail": "8.15.1,5.11-0.175.3.29.0.4.0:20180206T160949Z", 
        "postfix": "", 
        "abc123": ""
    }
}
```
```
# salt-call pkg.latest_version sendmail --out json
{
    "local": ""
}
```

```
# salt-call pkg.version sendmail --out json
{
    "local": "8.15.1,5.11-0.175.3.29.0.4.0:20180206T160949Z"
}
```
```
# salt-call pkg.latest_version postfix --out json
{
    "local": "3.2.2,5.11-0.175.3.23.0.1.0:20170712T212555Z"
}
```
```
# salt-call pkg.version postfix --out json
{
    "local": ""
}
```
```
# salt-call pkg.latest_version abc123 --out json
{
    "local": ""
}
```
```
# salt-call pkg.version abc123 --out json
{
    "local": ""
}
```

YanChii:
Thank you guys for taking this module further. I wrote it about 5 years ago and maybe I didn't test enough the `pkg.latest` state. Or something changed in the salt itself (I didn't follow everything).
I agree with @gmsoft-tuxicoman that (if possible) we should let the `pkg` command to do the resolving of partial FMRI so we don't have to reimplement it.
If you want, I can test your patch also on OmniOS.
YanChii

gmsoft-tuxicoman:
Sorry I'm not sure what you expect from me at this point.
Do you want me to review pkg.py and find out if there could be any issue with the fact that package names get resolved ?

damon-atkins:
As salt/modules/solarisips.py is called by states/pkg.py its worth reading through at least once. I do not think the package name will be an issue, as I suspect that code which does do a comparison is not engaged for the code in solarisips.py.  It would be worth your time to write a few pkg states to upgrade and downgrade software, install latest, shortnames, long names, halfnames etc, if you have not already, as their is no Salt test suite for Solaris.

As I said the output is looking a lot more standard now. Maybe @YanChii  has some tests sls files from the original development. I don't have a Solaris system with Salt on it.

Its fantastic effort to improve this. 👍

gmsoft-tuxicoman:
I've spent some time testing and reviewing states/pkg.py. I had some doubts about the _uninstall() function just by reading it but it works perfectly. I couldn't find any problem with updated solarisips.py and the current states/pkg.py.

salt/modules/solarisips.py
```diff
@@ -298,39 +299,81 @@ def version(*names, **kwargs):
...
+        if ret[name] == installed[name]:
+            ret[name] = ''
+
+    # Append package names which are not found
+    unmatched = list(filter(lambda name: not reduce(lambda x, y : x or name in y, ret.keys(), False), names))
+    ret.update(zip(unmatched, [ '' ] * len(unmatched)))
+
+    # Return a string if only one package name passed
+    if len(names) == 1:
+        return list(ret.values())[0]
```
terminalmage: This can also be optimized, since `.values()` allocates a list. Here's how this same action is done elsewhere in Salt:

```python
if len(names) == 1:
    try:
        return next(six.itervalues(ret))
    except StopIteration:
        return ''
```

Note that this would also require that you add `from salt.ext import six` to the module's imports.

salt/modules/solarisips.py
```diff
@@ -298,39 +299,81 @@ def version(*names, **kwargs):
...
+
+    for name in ret:
+        if not name in installed:
+            continue
+        if ret[name] == installed[name]:
+            ret[name] = ''
+
+    # Append package names which are not found
+    unmatched = list(filter(lambda name: not reduce(lambda x, y : x or name in y, ret.keys(), False), names))
+    ret.update(zip(unmatched, [ '' ] * len(unmatched)))
```
terminalmage: This can be optimized using `itertools.cycle()`.

```python
ret.update(zip(unmatched, itertools.cycle(('',))))
```

You will also need to add the following two imports:

```python
import itertools
from salt.ext.six.moves import zip  # pylint: disable=redefined-builtin
```

The second import there ensures that Python 2 uses a `zip` which returns an iterator instead of a list.

terminalmage (changes requested):

dmurphy18 (changes requested):
Have a typo 'tu' should be 'to' in 
+    .. versionchanged:: Fluorine
+        Argument changed from 'name' to '*names' tu support multiple packages.

damon-atkins (approved):
Once other review changes are complete.

rallytime:
Hi @gmsoft-tuxicoman - Just a bump here to see if you've had a chance to review the latest feedback? It looks like this is close with the exception of getting the last couple of change requests across the line.

gmsoft-tuxicoman:
Sorry for the delay.
I have commited all the changes. All the test cases above still work like before.

rallytime:
Thanks @gmsoft-tuxicoman! There are some lint errors taht need to be cleaned up. Can you fix them?

https://jenkins.saltstack.com/job/PR/job/salt-pr-lint-n/23127/violations/file/salt/modules/solarisips.py/

dmurphy18 (approved):
apart from pylint error of extra blank line at line number 53, (nop in operation) looks like other lint errors fixed.

terminalmage:
re-run lint

salt/modules/solarisips.py
```diff
@@ -47,6 +48,9 @@
 import salt.utils.path
 import salt.utils.pkg
 from salt.exceptions import CommandExecutionError
+from salt.ext import six
+from salt.ext.six.moves import zip  # pylint: disable=redefined-builtin
+
```
terminalmage: The linter doesn't like this extra blank line, it needs to be removed.

salt/modules/solarisips.py
```diff
@@ -298,39 +303,87 @@ def version(*names, **kwargs):
...
+    cmd = ['/bin/pkg', 'list', '-Hv']
+    cmd.extend(names)
+    lines = __salt__['cmd.run_stdout'](cmd, ignore_retcode=True).splitlines()
+    ret = {}
+    for line in lines:
+        ret[_ips_get_pkgname(line)] = _ips_get_pkgversion(line)
 
+    # Append package names which are not installed/found
+    unmatched = list(filter(lambda name: not reduce(lambda x, y: x or name in y, ret, False), names))
```
terminalmage: The linter is marking `name` as not defined here, but this may be a false positive due to the fact that you are calling a lambda within another lambda.

salt/modules/solarisips.py
```diff
@@ -298,39 +303,87 @@ def version(*names, **kwargs):
...
+        installed = {list(ret)[0] if len(ret) > 0 else names[0]: installed}
+
+    for name in ret:
+        if name not in installed:
+            continue
+        if ret[name] == installed[name]:
+            ret[name] = ''
+
+    # Append package names which are not found
+    unmatched = list(filter(lambda name: not reduce(lambda x, y: x or name in y, ret, False), names))
```
terminalmage: Same thing as above: the linter is marking `name` as not defined here, but this may be a false positive due to the fact that you are calling a lambda within another lambda.

terminalmage (changes requested):

terminalmage:
@gmsoft-tuxicoman Did you run the code on a Solaris box to confirm that those two other lint failures are false-positives? If so, then you'll need to disable those lint checks by adding the following at the end of the line:

```
# pylint: disable=undefined-variable
```

Also, you'll need to put _two_ spaces before the `#` or else Pylint will give you a whole _different_ error :smile:

terminalmage:
Actually, pylint doesn't like the use of `filter` for Python 3 compatibility reasons.

From https://jenkinsci.saltstack.com/job/pr-lint/job/PR-47909/12/console

```
[salt linting] ************* Module salt.modules.solarisips
[salt linting] salt/modules/solarisips.py:53: [W1699(incompatible-py3-code), ] Incompatible Python 3 code found. Proposed fix:
[salt linting]  from salt.exceptions import CommandExecutionError
[salt linting]  from salt.ext import six
[salt linting]  from salt.ext.six.moves import zip  # pylint: disable=redefined-builtin
[salt linting] +from functools import reduce
[salt linting]  
[salt linting]  # Define the module's virtual name
[salt linting]  __virtualname__ = 'pkg'
[salt linting] salt/modules/solarisips.py:316: [W1699(incompatible-py3-code), ] Incompatible Python 3 code found. Proposed fix:
[salt linting]          ret[_ips_get_pkgname(line)] = _ips_get_pkgversion(line)
[salt linting]  
[salt linting]      # Append package names which are not installed/found
[salt linting] -    unmatched = list(filter(lambda name: not reduce(lambda x, y: x or name in y, ret, False), names))
[salt linting] +    unmatched = list([name for name in names if not reduce(lambda x, y: x or name in y, ret, False)])
[salt linting]      ret.update(zip(unmatched, itertools.cycle(('',))))
[salt linting]  
[salt linting]      # Return a string if only one package name passed
[salt linting] salt/modules/solarisips.py:375: [W1699(incompatible-py3-code), ] Incompatible Python 3 code found. Proposed fix:
[salt linting]              ret[name] = ''
[salt linting]  
[salt linting]      # Append package names which are not found
[salt linting] -    unmatched = list(filter(lambda name: not reduce(lambda x, y: x or name in y, ret, False), names))
[salt linting] +    unmatched = list([name for name in names if not reduce(lambda x, y: x or name in y, ret, False)])
```

gmsoft-tuxicoman:
I have fixed the issues as suggested. I reran all the test cases and the output is still the same.

terminalmage:
Awesome, pending lint results this looks good to merge.

PR #47923, opened 2018-06-01
Prevent crash if files in file.recurse

isbm:
### What does this PR do?

Bugfix.

```yaml
foo:
  file.recurse:
    - name: /tmp/blah
    - source: salt://foo
```

If we have `$SALT/foo/здоров-був.txt` file, crash happens. This PR fixes it.

### DIFF
--- a/salt/modules/state.py
+++ b/salt/modules/state.py
@@ -41,7 +41,6 @@ import salt.utils.versions
 from salt.exceptions import CommandExecutionError, SaltInvocationError
 from salt.runners.state import orchestrate as _orchestrate
 from salt.utils.odict import OrderedDict
-from salt.utils import to_unicode
 
 # Import 3rd-party libs
 from salt.ext import six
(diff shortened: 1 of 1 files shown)


salt/modules/state.py
```diff
@@ -41,6 +41,7 @@
 from salt.exceptions import CommandExecutionError, SaltInvocationError
 from salt.runners.state import orchestrate as _orchestrate
 from salt.utils.odict import OrderedDict
+from salt.utils import to_unicode
```
terminalmage: `salt.utils` is deprecated, this would need to be imported from `salt.utils.stringutils`.

terminalmage (changes requested):
One minor change, and then this also needs a backport to 2018.3 (CC @rallytime)

isbm:
bart

terminalmage (approved):

PR #47939, opened 2018-06-02
Make postgres version compare more robust

steveno:
### What does this PR do?
It attempts to make the version comparison function in the PostgreSQL module more robust.

I did not see any open issues for this. It was labeled as a TODO in the code.

### Previous Behavior
The previous behavior was a simple string comparison.

### New Behavior
The new behavior breaks the PostgreSQL version string into chunks splitting on periods and attempts to compare the versions piecemeal. 

### Tests written?

Yes

### Commits signed with GPG?

Yes


salt/modules/postgres.py
```diff
@@ -1465,10 +1465,21 @@ def is_available_extension(name,
...
 def _pg_is_older_ext_ver(a, b):
     '''Return true if version a is lesser than b
-    TODO: be more intelligent to test versions
-
+    
+    Compare versions by breaking the version number into
+    its major, minor, and patch components.
     '''
-    return a < b
+    a, b = a.split('.'), b.split('.')
```
rares-pop: why not using distutils.version?

moench-tegeder:
Hm, some points: this is the comparison is only used for comparing PostgreSQL extension versions. The PostgreSQL version itself is handled as LooseVersion - so why not move the extension check to the same functions? That would also avoid the slightly hacking "pad version with zeros"-thing.
Then again I'd make sure the restrictions you're imposing on extension version numbers match what PostgreSQL does internally - e.g. I'm rather not sure that you have to have a "major.minor" structure (version "42" would be fully valid). In fact, from a quick glance, I gather PostgreSQL does not impose many restrictions, and even "1.2a" could be considered valid (or did I miss it?). I'd recommend being careful, to avoid breakage when some extension author tries to be clever/funny/whatever.

salt/modules/postgres.py
steveno: I was unaware of distutils.version. That's a good suggestion. Thank you.

steveno:
@moench-tegeder Looking through the [list of versions](https://www.postgresql.org/ftp/source/) I didn't see any with letters in them. There is the beta version, but I will admit I don't know what version it reports back once installed. I will update my code.

moench-tegeder:
But this is not about the version of the PostgreSQL server: you modified function _pg_is_older_ext_ver() which is only called from create_metadata() - and it's used to compare versions of PostgreSQL extensions. Extensions have their own versioning (and every extension on their own), and those version numbers can be assigned rather freely by the extension's author. Moreover, these extensions often exist as separate projects from PostgreSQL itself (famous, large examples: PL/R, PostGIS, ... but many small extensions exist, too; and even some "in house" extensions which have never been published for whatever reason); so it's impossible to gather a list of all extensions and their versions.

Even if you were thinking about PostgreSQL versions: current development versions self-identify as "11beta1 (Debian 11~beta1-2.pgdg90+1)" (obviously this one was installed from a Debian package).

salt/modules/postgres.py
rallytime: Or even `salt.utils.versions.LooseVersion` would perhaps work here as well.

steveno:
@rallytime I used your suggestion. That salt.utils.versions.LooseVersion was already an import so it seemed like an obvious choice.

rallytime (approved):
👍

moench-tegeder:
This is still... barking up the wrong tree? The code in question does not compare PostgreSQL versions but extension versions (by the way, the comment in the test get's this wrong, too). And extension versions do not follow the "PostgreSQL release version number scheme".
Here are two extensions (no functionality) with "perfectly valid" version numbers (as far as PostgreSQL is concerned): https://github.com/moench-tegeder/pg_ext_nonsense (these exists for demo purposes only, but such schemes ("1.2.3.4", "2.0a", "1_2", "2") do exist in the wild).

"Improving the PostgreSQL version comparison" would be moving all the handling of "server_version" to "server_version_num" https://www.postgresql.org/docs/10/static/runtime-config-preset.html#GUC-SERVER-VERSION-NUM (except where it's for display purposes only).

steveno:
@moench-tegeder I'm not sure what you're expecting? I will gladly change the comment that's an easy fix. I can also add more examples to the test that are contrived like in your demo repository. Is there something else I'm missing?

moench-tegeder:
My main concern is that you're trying to improve a function for a purpose for which it is never used. If you look closely at modules/postgres.py, _pg_is_older_ext_ver() is called in exactly one place in create_metadata() - which returns "lifecycle information about an extension". So all the discussion about PostgreSQL versions is moot - the extensions have their own versions, which have nothing to do with the PostgreSQL version (compare "SELECT * FROM pg_available_extension_versions" on a PostgreSQL instance of your choice (remember to install the -contrib package, if available)).

Inside PostgreSQL, the extension version is really only handled and compared as a string - see https://github.com/postgres/postgres/blob/master/src/backend/commands/extension.c for the gory details (I reckon this detail wil get us into trouble on the PostgreSQL side at some point in the future, but for now it is as it is). And the decision wether an extension is upgradeable should be the same on the saltstack side and the PostgreSQL side, else the information returned by saltstack will not be helpful at all. (Still with me? I know this is long...). Unfortunately, PostgreSQL doesn't really care about the extension versions at all - all it needs is a chain of one or more "extension upgrade files" - see find_update_path() in PostgreSQL's extension.c. So, basically: I'm not even sure if it's really possible to "improve" _pg_is_older_ext_ver() - trying for the simple string comparison we have now could very well be the only thing we can do at all. And from the whole thread here and your commits I guess you're trying to build a better comparison for PostgreSQL versions - that's what I'm trying to express with "barking up the wrong tree". (Language barrier may make this conversation harder than it should be... I'm trying my best).

But the other thing is: modules/postgres.py has a function _parsed_version() - "Returns the server version properly parsed and int casted for internal use" - which retrieves the PostgreSQL version via version() and uses _LooseVersion() to turn that into something "comparable". And this is something which should be improved: since almost forever (2006, to be precise) PostgreSQL not only has "server_version" (a string, for human consumption) but also "server_version_num", an integer especially made for comparing PostgreSQL server versions. The "internal" representation _parsed_version() is talking about should just use server_version_num.

Oh, and about the "contrived" examples: PostGIS routinely uses versions as "2.4.4dev" and "2.4.4next" - but PostGIS is somewhat large for "just a little test".

steveno:
> Inside PostgreSQL, the extension version is really only handled and compared as a string - see https://github.com/postgres/postgres/blob/master/src/backend/commands/extension.c for the gory details (I reckon this detail wil get us into trouble on the PostgreSQL side at some point in the future, but for now it is as it is). And the decision wether an extension is upgradeable should be the same on the saltstack side and the PostgreSQL side, else the information returned by saltstack will not be helpful at all.

@moench-tegeder That was the piece I didn't understand. Now we're on the same page I think.

I don't see any less then or greater than comparisons in the code you linked. I only see a lot `strcmp`s. Where exactly is PostgreSQL doing the comparisons similar to salt?

moench-tegeder:
The scary thing is: PostgreSQL does not compare extension versions, at all (I was surprised, too). There are only a few rules:
- the default version of an extension is defined in it's control file. that's the version you get when you do a simple "CREATE EXTENSION foo".
- a file "foo--$ver.sql" is the SQL file for creating the objects of extension "foo" version "$version"
- a file "foo--$vera--$verb.sql" updates extension "foo" from $vera to $verb - thus defining $verb to be larger than $vera (that's the real scary thing) - it's up to the developer not to mess this up too much
- when upgrading an extension when there's no direct upgrade file, PostgreSQL uses Dijkstra's algorithm to build a chain of updates. This really works, when you think a little about it...

I had thought there were some comparisons going on - but about the only comparison is used for breaking ties when building the update path. Wen I learned that I concluded that it's perhaps best to just throw our hand into the air (it's a mess , even when it's a well-working mess) - or at least make sure that all the "funny" versions and perhaps some more do work with the new system.

steveno:
I just pushed an update that clarifies the comments that we're testing postgres extension versions, not versions of postgres. I also added a couple of more tests. 

I think this is ready for another review!

steveno:
@rallytime or @rares-pop Is there anything else this needs to be merged? Please let me know and I'll attempt to get it fixed as soon as I can.

rallytime:
Hi @steveno - Thanks for the bump. :)

Looks good to me!

moench-tegeder:
OK, let's see if this works out or if we get fallout - I think it shouldn't be worse than before :)

PR #47951, opened 2018-06-04
Fix NI Linux RT distribution core grains

rares-pop:
NI fixed lsb_distrib_id to correctly differentiate
between older distribution vs current distribution
which have different behavior for multiple module
functionality, but it should continue to populate
'osfullname' as 'NILinuxRT' not to break existing code

This should also be backported in 2018.3

### What does this PR do?
National Instruments used to have a custom set of commits in-house, for the older Linux RT distribution, but it makes more sense to upstream that too and expand salt's support. In order to do that, we have to differentiate between those two platforms, and we do that now through this 'lsb_distrib_id' grain.

It also clean up the nilrt-xfce distribution that was never a distribution.

### Previous Behavior
grains['lsb_distrib_id'] was always showing 'nilrt' previously and it was some custom code that differentiate between older distribution and the new distribution of Linux RT. That had to be spread around multiple modules that needed that logic.

### New Behavior
Modules can use the grains['lsb_distrib_id'] info to differentiate between newer vs. older linux distribution for National Instruments Linux RT platform
For the older distribution is showing 'nilrt', as for the newer distribution 'nilrt-nxg'.

### Tests written?

No

### Commits signed with GPG?

No

### DIFF
--- a/salt/grains/core.py
+++ b/salt/grains/core.py
@@ -1350,7 +1350,6 @@ _OS_NAME_MAP = {
     'scientific': 'ScientificLinux',
     'synology': 'Synology',
     'nilrt': 'NILinuxRT',
-    'nilrt-xfce': 'NILinuxRT-XFCE',
     'poky': 'Poky',
     'manjaro': 'Manjaro',
     'manjarolin': 'Manjaro',
@@ -1423,7 +1422,6 @@ _OS_FAMILY_MAP = {
     'Cumulus': 'Debian',
     'Deepin': 'Debian',
     'NILinuxRT': 'NILinuxRT',
-    'NILinuxRT-XFCE': 'NILinuxRT',
     'KDE neon': 'Debian',
     'Void': 'Void',
     'IDMS': 'Debian',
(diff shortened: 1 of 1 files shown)


salt/grains/core.py
```diff
@@ -1788,8 +1786,12 @@ def os_data():
         # so that linux_distribution() does the /etc/lsb-release parsing, but
         # we do it anyway here for the sake for full portability.
         if 'osfullname' not in grains:
-            grains['osfullname'] = \
-                grains.get('lsb_distrib_id', osname).strip()
+            # If NI Linux RT distribution, set the grains['osfullname'] to 'nilrt'
+            if grains.get('lsb_distrib_id', '').startswith('nilrt'):
```
terminalmage: Are you sure this will always startwith a _lowercase_ `nilrt`? Most distribution IDs are not all lowercase. Maybe it would make sense to use `.lower()` here.

```python
if grains.get('lsb_distrib_id', '').lower().startswith('nilrt'):
```

rares-pop: Okay, that makes sense! NI's linux distribution relies on OpenEmbedded, and that's lower-case based, but it doesn't hurt comparing with lower() case.

Done!

isbm (changes requested):

salt/grains/core.py
```diff
@@ -1788,8 +1786,12 @@ def os_data():
...
         # we do it anyway here for the sake for full portability.
         if 'osfullname' not in grains:
-            grains['osfullname'] = \
-                grains.get('lsb_distrib_id', osname).strip()
+            # If NI Linux RT distribution, set the grains['osfullname'] to 'nilrt'
+            if grains.get('lsb_distrib_id', '').lower().startswith('nilrt'):
+                grains['osfullname'] = 'nilrt'
+            else:
+                grains['osfullname'] = \
+                    grains.get('lsb_distrib_id', osname).strip()
```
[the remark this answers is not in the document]
rares-pop: It was like that before my change. I fixed it though!

terminalmage (approved):

isbm (approved):

PR #47964, opened 2018-06-05
Fix 47921

### LINKED ISSUE #47921: [develop] file.serialized with json and merge_if_exists doesn't work due to wrong options being passed to deserialize function
I was looking forward to using `file.serialize`'s new `serializer_opts` argument on the develop branch, but ran into a problem when trying it out.

Given a state like:

```
/path/to/file.json:
  file.serialize:
    - dataset:
        foo: bar
    - formatter: json
    - serializer_opts:
      - indent: 4
    - merge_if_exists: true
    - replace: false
```

Produces the following error:

```
Function: file.serialize
  Result: False
            File "/var/tmp/.salt_08831c_salt/pyall/salt/serializers/json.py", line 41, in deserialize
              stream_or_string, _json_module=_json, **options)
            File "/var/tmp/.salt_08831c_salt/pyall/salt/utils/json.py", line 77, in load
              return kwargs.pop('_json_module', json).load(fp, **kwargs)
            File "/usr/lib/python3.5/json/__init__.py", line 268, in load
              parse_constant=parse_constant, object_pairs_hook=object_pairs_hook, **kw)
            File "/usr/lib/python3.5/json/__init__.py", line 332, in loads
              return cls(**kw).decode(s)
          TypeError: __init__() got an unexpected keyword argument 'indent'

            File "/var/tmp/.salt_08831c_salt/pyall/salt/state.py", line 1928, in call
              **cdata['kwargs'])
            File "/var/tmp/.salt_08831c_salt/pyall/salt/loader.py", line 1837, in wrapper
              return f(*args, **kwargs)
            File "/var/tmp/.salt_08831c_salt/pyall/salt/states/file.py", line 6309, in serialize
              existing_data = __serializers__[deserializer_name](fhr, **options.get(serializer_name, {}))
            File "/var/tmp/.salt_08831c_salt/pyall/salt/serializers/json.py", line 48, in deserialize
              raise DeserializationError(error)
          salt.serializers.DeserializationError: __init__() got an unexpected keyword argument 'indent'
 Started: 15:26:49.300034
Duration: 4.08 ms
 Changes:
```

This solution makes the error go away, but might break other use cases.

```diff
diff --git a/salt/states/file.py b/salt/states/file.py
index 920d7bbfad..67b7cb52c7 100644
--- a/salt/states/file.py
+++ b/salt/states/file.py
@@ -6306,7 +6306,7 @@ def serialize(name,
                         'result': False}
 
             with salt.utils.files.fopen(name, 'r') as fhr:
-                existing_data = __serializers__[deserializer_name](fhr, **options.get(serializer_name, {}))
+                existing_data = __serializers__[deserializer_name](fhr)
 
             if existing_data is not None:
                 merged_data = salt.utils.dictupdate.merge_recurse(existing_data, dataset)
```

```
$ salt-ssh --versions
Salt Version:
           Salt: 2018.3.0-473-g2bac7b2
 
Dependency Versions:
           cffi: Not Installed
       cherrypy: Not Installed
       dateutil: Not Installed
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.10
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.5.6
   mysql-python: Not Installed
      pycparser: Not Installed
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: Not Installed
         Python: 3.5.3 (default, Jan 19 2017, 14:11:04)
   python-gnupg: Not Installed
         PyYAML: 3.12
          PyZMQ: 17.0.0
           RAET: Not Installed
          smmap: Not Installed
        timelib: Not Installed
        Tornado: 5.0.2
            ZMQ: 4.1.6
 
System Versions:
           dist: debian 9.4 
         locale: UTF-8
        machine: x86_64
        release: 4.9.0-6-amd64
         system: Linux
        version: debian 9.4
```

nocturo:
### What does this PR do?
Attempts to use same `serializer_opts` for deserialization during `merge_if_exists`, if it's not working (`JSONDecode` class for example doesn't accept kwargs it doesn't know about) try deserializing without custom options.

Ideally we would also have `deserializer_opts` for this but I think it would just clutter the state. If there is a legitimate use when you need to pass something to both serializer and deserializer, but for example deserializer doesn't accept only one of the options, this would all fail. I'm not sure if there is a legitimate case for that yet.

Also fixes `serializer_opts` not being passed down if the serialize_name doesn't exist in options dict.

Fixes #47921

### Tests written?

No

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/states/file.py
+++ b/salt/states/file.py
@@ -6298,6 +6298,9 @@ def serialize(name,
                 }
 
     if serializer_opts:
+        if not options.get(serializer_name, {}):
+            options[serializer_name] = {}
+
         options.get(serializer_name, {}).update(
             salt.utils.data.repack_dictlist(serializer_opts)
         )
@@ -6312,7 +6315,11 @@ def serialize(name,
                         'result': False}
 
             with salt.utils.files.fopen(name, 'r') as fhr:
-                existing_data = __serializers__[deserializer_name](fhr, **options.get(serializer_name, {}))
+                try:
+                    existing_data = __serializers__[deserializer_name](fhr, **options.get(serializer_name, {}))
+                except (TypeError, salt.serializers.DeserializationError):
+                    fhr.seek(0)
+                    existing_data = __serializers__[deserializer_name](fhr)
 
             if existing_data is not None:
                 merged_data = salt.utils.dictupdate.merge_recurse(existing_data, dataset)
--- a/salt/states/file.py
+++ b/salt/states/file.py
@@ -6318,6 +6318,7 @@ def serialize(name,
                 try:
                     existing_data = __serializers__[deserializer_name](fhr, **options.get(serializer_name, {}))
                 except (TypeError, salt.serializers.DeserializationError):
+                    log.debug('DeserializationError exception caught, trying to merge without serializer_opts: %s', options.get(serializer_name, {}))
                     fhr.seek(0)
                     existing_data = __serializers__[deserializer_name](fhr)
 


salt/states/file.py
```diff
@@ -6312,7 +6315,11 @@ def serialize(name,
                         'result': False}
 
             with salt.utils.files.fopen(name, 'r') as fhr:
-                existing_data = __serializers__[deserializer_name](fhr, **options.get(serializer_name, {}))
+                try:
+                    existing_data = __serializers__[deserializer_name](fhr, **options.get(serializer_name, {}))
```
gtmanfred: instead of just passing in all the args, i would do something like.

```
goodopts = salt.utils.args.get_function_argspec(__serializers__[deserializer_name]).args
opts = opts.get(serializer_name, {})
for opt in list(opts):
    if opt not in goodopts:
        opts.pop(opt, None)
```

And only pass in the key word arguments if the deserializer actually accepts those arguments.

gtmanfred (changes requested):

salt/states/file.py
nocturo: That was my first instinct but this is not the inability of serializer/deserializer to accept args, but actual python stdlib JSONDecode doesn't accept it. I don't know how could I test if some random deserializer will be able to accept our kwargs except in a try block. Do you have an idea if that's possible?

gtmanfred: ahh yes, in that case, this is the best option.  But it would be good to add a debug log that says that this exception was caught and that the function is being run without the options.

nocturo: Added debug output that we caught the exception with what options were passed to it.

gtmanfred (approved):

terminalmage (approved):

PR #47966, opened 2018-06-05
Fix postfix.set_main's assumption of prefix-free key names

### LINKED ISSUE #47888: postfix.set_main overwrites wrong keys

The `postfix` module's `set_main` function overwrites wrong keys in addition to the key it should overwrite. The reason is that Postfix' keys aren't prefix-free, but the `set_main` function assumes they are.

### Setup

First create a dummy `main.cf` file, e.g. `/tmp/test.cf`, and add the following content:

```
virtual_transport_rate_delay = $default_transport_rate_delay
virtual_transport = virtual
```

Those are default values for two keys from an Ubuntu installation as given by `postconf | grep virtual`.

Next, try setting `virtual_transport` while leaving `virtual_transport_rate_delay` as-is:

```
[0 mbunkus@reliant ~] sudo salt-call postfix.set_main virtual_transport testme path=/tmp/test.cf
local:
    - virtual_transport = testme
    - virtual_transport = testme
```

The output already shows that two keys are replaced, not just one. Looking at the file via `cat /tmp/test.cf` confirms it:

```
[0 mbunkus@reliant ~] cat /tmp/test.cf
virtual_transport = testme
virtual_transport = testme
```

### Analysis

The problem is that `set_main` assumes that keys are prefix-free. While iterating over the list of lines, it uses `line.startswith(key)` as the test whether or not to replace the line. This test is obviously true for `virtual_transport = …`, but unfortunately it's also true for any other key whose name starts with `virtual_transport`, such as the aforementioned `virtual_transport_rate_delay = …`

Remedy: don't use `startswith(key)`; rather use something like "starts with the key followed by a whitespace or an equal", e.g. `re.match("^{0}[\\s=]".format(re.escape(key)), line)`. The regex could be compiled earlier in the function, of course.

### Versions Report

```
Salt Version:
           Salt: 2017.7.5

Dependency Versions:
           cffi: Not Installed
       cherrypy: Not Installed
       dateutil: 2.6.1
      docker-py: Not Installed
          gitdb: 2.0.3
      gitpython: 2.1.8
          ioflo: Not Installed
         Jinja2: 2.10
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: 1.0.7
   msgpack-pure: Not Installed
 msgpack-python: 0.5.6
   mysql-python: Not Installed
      pycparser: Not Installed
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: Not Installed
         Python: 2.7.15rc1 (default, Apr 15 2018, 21:51:34)
   python-gnupg: Not Installed
         PyYAML: 3.12
          PyZMQ: 16.0.2
           RAET: Not Installed
          smmap: 2.0.3
        timelib: Not Installed
        Tornado: 4.5.3
            ZMQ: 4.2.5

System Versions:
           dist: Ubuntu 18.04 bionic
         locale: UTF-8
        machine: x86_64
        release: 3.13.0-88-generic
         system: Linux
        version: Ubuntu 18.04 bionic
```

Yeah, these are the 16.04 Salt packages on Ubuntu 18.04. However, the same code is still present in [today's git](https://github.com/saltstack/salt/blob/develop/salt/modules/postfix.py#L298); therefore the problem still exists.

mbunkus:
Salt's "postfix.set_main" module function used "line.startswith()"
for determining which entries from "main.cf" to keep and which to
overwrite. This matched more lines that intended, though, as the names
of Postfix' configuration options aren't prefix-free.

For example, both "virtual_transport" and
"virtual_transport_rate_delay" are valid and known keys. Therefore
trying to set "virtual_transport" would lead to
"virtual_transport_rate_delay" being replaced, too, by another
instance of "virtual_transport".

The new method matches the line based on the key name being terminated
by whitespace, an equal sign or the end of the line.

Fixes #47888.

### Tests written?

No

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/modules/postfix.py
+++ b/salt/modules/postfix.py
@@ -288,9 +288,10 @@ def set_main(key, value, path=MAIN_CF):
     pairs, conf_list = _parse_main(path)
 
     new_conf = []
+    key_line_match = re.compile("^{0}([\\s=]|$)".format(re.escape(key)))
     if key in pairs:
         for line in conf_list:
-            if line.startswith(key):
+            if re.match(key_line_match, line):
                 new_conf.append('{0} = {1}'.format(key, value))
             else:
                 new_conf.append(line)


rallytime (approved):
Thanks @mbunkus!

Issue #47937, opened 2018-06-01
Add a git.cloned state for cases where updating is unnecessary after cloning


viraptor:
### Description of Issue/Question

In some configurations of `git.latest`, there's no point in running `ls-remote` to find what HEAD (or other `rev`) is pointing at.

This matters for large repos, because running `git ls-remote` against them can take minutes on a slow connection.
It also prevents offline runs when the local state is sufficient.

### Setup

```
some_repo:
  git.latest:
    - name: git@github.com:....
    - target: /some/local/path
    - update_head: false
```

In this case, if the repo is already cloned and checked out (HEAD exists locally), there's no point running anything against the remote repository.

### Versions Report
```
Salt Version:
           Salt: 2018.3.0

Dependency Versions:
           cffi: 1.11.5
       cherrypy: Not Installed
       dateutil: Not Installed
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.10
        libgit2: 0.27.0
        libnacl: Not Installed
       M2Crypto: 0.29.0
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.5.6
   mysql-python: Not Installed
      pycparser: 2.18
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: 0.27.0
         Python: 2.7.15 (default, May  1 2018, 16:44:08)
   python-gnupg: Not Installed
         PyYAML: 3.12
          PyZMQ: 17.0.0
           RAET: Not Installed
          smmap: Not Installed
        timelib: Not Installed
        Tornado: 4.5.3
            ZMQ: 4.2.5

System Versions:
           dist:
         locale: UTF-8
        machine: x86_64
        release: 17.5.0
         system: Darwin
        version: 10.13.4 x86_64
```

terminalmage:
I disagree, given the following from the documentation:

```
    update_head : True
        If set to ``False``, then the remote repository will be fetched (if
        necessary) to ensure that the commit to which ``rev`` points exists in
        the local checkout, but no changes will be made to the local HEAD.

        .. versionadded:: 2015.8.3
```

The `ls-remote` is needed to determine whether the local checkout has the commit to which the current remote points.

[...]

PR #47975, opened 2018-06-05
Add a new git.cloned state

terminalmage:
See https://github.com/saltstack/salt/issues/47937


cachedout:
2018.3 and not develop?

terminalmage:
Yes, I made the choice to put it in 2018.3 because it doesn't touch any existing code (i.e. no code was changed in the git module, and code was only _added_ to the state module), and it satisfied use case from someone who wanted `git.latest` to not check for updates if the repo was already cloned.

cachedout (approved):

andreasnuesslein:
just a heads up, for others who might land here after trying to run `git.cloned`:
although it's already in the [docs](https://docs.saltstack.com/en/latest/ref/states/all/salt.states.git.html#salt.states.git.cloned), it refers to 2018.3.**3** . the latest release as of now is 2018.3.**2**.

terminalmage:
Yes, 2018.3.3 is nearing release, so within a few weeks it will be available.

PR #47985, opened 2018-06-05
Fix `unit.states.test_reg` test failure

twangboy:
### What does this PR do?
Fixes failing test by mocking `reg.cast_vdata`

Found by Jenkins

### Tests written?
Yes

### Commits signed with GPG?
Yes

### DIFF
--- a/tests/unit/states/test_reg.py
+++ b/tests/unit/states/test_reg.py
@@ -18,6 +18,7 @@ from tests.support.mock import (
 # Import Salt Libs
 import salt.states.reg as reg
 import salt.utils.platform
+import salt.utils.stringutils
 
 
 @skipIf(NO_MOCK, NO_MOCK_REASON)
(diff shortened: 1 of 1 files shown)


tests/unit/states/test_reg.py
```diff
@@ -48,8 +49,12 @@ def test_present(self):
                                            {'vdata': 'a', 'success': True},
                                            {'vdata': 'a', 'success': True}])
         mock_t = MagicMock(return_value=True)
+        mock_cast = MagicMock(
+            return_value=salt.utils.stringutils.to_unicode(vdata)
+        )
```
skizunov: Could this simply be:
```
mock_cast = MagicMock(return_value=vdata)
```
Because vdata is already unicode (on PY2) due to `from __future__ import unicode_literals`?

Or am I missing something that requires the explicit cast?

twangboy: No, it's my own idiocy. I saw that that's what `cast_vdata` was returning, so I thought I should make the mock the same. The tests pass without the `to_unicode`. I'll remove it.

skizunov (approved):

gtmanfred (approved):

gtmanfred (approved):

PR #47987, opened 2018-06-05
Fix `unit.utils.test_hashutils` unicode issues

twangboy:
### What does this PR do?
Fixes an issue where the hash was being returned as unicode, but was being compared with a bytestring, so the assert was failing... I think.

Found in Jenkins testing

### Tests written?
Yes

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/utils/hashutils.py
+++ b/salt/utils/hashutils.py
@@ -13,6 +13,7 @@ import random
 # Import Salt libs
 from salt.ext import six
 import salt.utils.files
+import salt.utils.platform
 import salt.utils.stringutils
 
 from salt.utils.decorators.jinja import jinja_filter
@@ -27,7 +28,8 @@ def base64_b64encode(instr):
     newline ('\\n') characters in the encoded output.
     '''
     return salt.utils.stringutils.to_unicode(
-        base64.b64encode(salt.utils.stringutils.to_bytes(instr))
+        base64.b64encode(salt.utils.stringutils.to_bytes(instr)),
+        encoding='utf8' if salt.utils.platform.is_windows() else None
     )
 
 
@@ -38,7 +40,10 @@ def base64_b64decode(instr):
     '''
     decoded = base64.b64decode(salt.utils.stringutils.to_bytes(instr))
     try:
-        return salt.utils.stringutils.to_unicode(decoded)
+        return salt.utils.stringutils.to_unicode(
+            decoded,
+            encoding='utf8' if salt.utils.platform.is_windows() else None
+        )
     except UnicodeDecodeError:
         return decoded
 
@@ -52,7 +57,8 @@ def base64_encodestring(instr):
     at the end of the encoded string.
     '''
     return salt.utils.stringutils.to_unicode(
-        base64.encodestring(salt.utils.stringutils.to_bytes(instr))
+        base64.encodestring(salt.utils.stringutils.to_bytes(instr)),
+        encoding='utf8' if salt.utils.platform.is_windows() else None
     )
 
 
@@ -68,7 +74,10 @@ def base64_decodestring(instr):
         # PY2
         decoded = base64.decodestring(b)
     try:
-        return salt.utils.stringutils.to_unicode(decoded)
+        return salt.utils.stringutils.to_unicode(
+            decoded,
+            encoding='utf8' if salt.utils.platform.is_windows() else None
+        )
     except UnicodeDecodeError:
         return decoded
 
--- a/tests/unit/utils/test_hashutils.py
+++ b/tests/unit/utils/test_hashutils.py
@@ -8,8 +8,6 @@ from tests.support.unit import TestCase
 
 # Import Salt libs
 import salt.utils.hashutils
-import salt.utils.stringutils
-import salt.utils.platform
 
 
 class HashutilsTestCase(TestCase):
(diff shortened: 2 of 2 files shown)


gtmanfred (approved):

rallytime:
I don't think this is quite right. There are 2 tests failing here: https://jenkins.saltstack.com/job/PR/job/salt-pr-linode-ubuntu14-n/23455/

- unit.utils.test_hashutils.HashutilsTestCase.test_base64_b64decode
- unit.utils.test_hashutils.HashutilsTestCase.test_base64_decodestring

terminalmage (commented):
No, this is incorrect. Decoding bytes that were base64-encoded will result in a bytestring: Python 3 example:

```python
>>> import base64
>>> base64.b64decode('Yhn2hg4aHHMM2iZ6dvyi3Q==')
b'b\x19\xf6\x86\x0e\x1a\x1cs\x0c\xda&zv\xfc\xa2\xdd'
```

terminalmage:
It's not clear what this is actually trying to fix, can you elaborate?

terminalmage:
> However, because we started using unicode_literals the value for self.bytes is being converted to:
>
    u'b\x19\xf6\x86\x0e\x1a\x1cs\x0c\xda&zv\xfc\xa2\xdd

No, `self.bytes` is definitely not a `unicode` type. It's defined as a bytestring, see [this line](https://github.com/saltstack/salt/blob/7266c99/tests/unit/utils/test_hashutils.py#L28). 

If anything, `base64.b64decode()` is acting differently on Windows. I think I know how we can get to the bottom of this. Please run the following sets of commands, each on fresh Python shells, and post the results. I'll do the same on Linux, and we can compare.

### Test 1 (Python 2, *with* `unicode_literals`)
```python
Python 2.7.5 (default, Aug  4 2017, 00:39:18)
[GCC 4.8.5 20150623 (Red Hat 4.8.5-16)] on linux2
>>> import salt.utils.hashutils
>>> salt.utils.hashutils.base64_b64decode('Yhn2hg4aHHMM2iZ6dvyi3Q==')
'b\x19\xf6\x86\x0e\x1a\x1cs\x0c\xda&zv\xfc\xa2\xdd'
>>>
```

### Test 2 (Python 2, *without* `unicode_literals`, *unicode* input to `b64decode`)
```python
Python 2.7.5 (default, Aug  4 2017, 00:39:18)
[GCC 4.8.5 20150623 (Red Hat 4.8.5-16)] on linux2
>>> import salt.utils.hashutils
>>> salt.utils.hashutils.base64_b64decode(u'Yhn2hg4aHHMM2iZ6dvyi3Q==')
'b\x19\xf6\x86\x0e\x1a\x1cs\x0c\xda&zv\xfc\xa2\xdd'
>>>
```

### Test 3 (Python 2, *without* `unicode_literals`, *str* input to `b64decode`)
```python
Python 2.7.5 (default, Aug  4 2017, 00:39:18)
[GCC 4.8.5 20150623 (Red Hat 4.8.5-16)] on linux2
>>> import salt.utils.hashutils
>>> salt.utils.hashutils.base64_b64decode('Yhn2hg4aHHMM2iZ6dvyi3Q==')
'b\x19\xf6\x86\x0e\x1a\x1cs\x0c\xda&zv\xfc\xa2\xdd'
>>>
```

### Test 4 (Python 3, *with* `unicode_literals`)
```python
Python 3.4.8 (default, Mar 23 2018, 10:04:27)
[GCC 4.8.5 20150623 (Red Hat 4.8.5-16)] on linux
>>> import salt.utils.hashutils
>>> salt.utils.hashutils.base64_b64decode('Yhn2hg4aHHMM2iZ6dvyi3Q==')
b'b\x19\xf6\x86\x0e\x1a\x1cs\x0c\xda&zv\xfc\xa2\xdd'
>>>
```

### Test 5 (Python 3, *without* `unicode_literals`)
```python
Python 3.4.8 (default, Mar 23 2018, 10:04:27)
[GCC 4.8.5 20150623 (Red Hat 4.8.5-16)] on linux
>>> import salt.utils.hashutils
>>> salt.utils.hashutils.base64_b64decode('Yhn2hg4aHHMM2iZ6dvyi3Q==')
b'b\x19\xf6\x86\x0e\x1a\x1cs\x0c\xda&zv\xfc\xa2\xdd'
>>>
```

On Linux, we get a bytestring return in all cases.

twangboy:
The problem is with this portion of code at the end of the `hashutils.base64_decodestring` function:
```
    try:
        return salt.utils.stringutils.to_unicode(decoded)
    except UnicodeDecodeError:
        return decoded
```
It tries a `to_unicode` and if it fails it just returns the decoded value. On Linux, `to_unicode` fails because the system encoding is `utf-8`, so it returns a byte string. On Windows, `to_unicode` success because the system encoding is `cp1252`, which for some reason succeeds and returns Unicode instead of a byte string. Is that the desired output? Do we want a Unicode output here? Or do we want a byte string, in which case we should remove the `to_unicode` function from the `hashutils.base64_encoding` function. Or we change the test to test for Unicode value on Windows and byte string otherwise.

terminalmage:
I believe the `to_unicode` is being used here because these values are being used in jinja filters.

> On Linux, `to_unicode` fails because the system encoding is `utf-8`, so it returns a byte string. On Windows, `to_unicode` success because the system encoding is `cp1252`, which for some reason succeeds and returns Unicode instead of a byte string.

I think you're mistaken about this. Prior to the 2018.3.0 release, the `to_unicode`/`to_str`/`to_bytes` helpers were all modified to try `utf-8` first.

terminalmage:
OK, I think I understand what you were getting at before... the `utf-8` attempt fails, but the `cp1252` fallback does not. Rather than modifying the test, we should just `utf-8` on Windows.

terminalmage:
@twangboy how about https://github.com/saltstack/salt/pull/48041?

Issue #47955, opened 2018-06-04
2018.3.1 Creating Windows machine in Amazon using salt-cloud fails.


frogunder:
### Description of Issue/Question
In 2018.3.1 when trying to create a windows machine in amazon from a linux master with salt-cloud it fails with this error: 
```
[ERROR   ] There was a profile error: data must be bytes or None, not unicode
Traceback (most recent call last):
  File "/usr/lib/python2.7/site-packages/salt/cloud/cli.py", line 281, in run
    self.config.get('names')
  File "/usr/lib/python2.7/site-packages/salt/cloud/__init__.py", line 1454, in run_profile
    ret[name] = self.create(vm_)
  File "/usr/lib/python2.7/site-packages/salt/cloud/__init__.py", line 1284, in create
    output = self.clouds[func](vm_)
  File "/usr/lib/python2.7/site-packages/salt/cloud/clouds/ec2.py", line 2705, in create
    vm_, data, ip_address, display_ssh_output
  File "/usr/lib/python2.7/site-packages/salt/cloud/clouds/ec2.py", line 2333, in wait_for_instance
    call='action',
  File "/usr/lib/python2.7/site-packages/salt/cloud/clouds/ec2.py", line 4765, in get_password_data
    key = RSA.load_key_string(rsa_key)
  File "/usr/lib64/python2.7/site-packages/M2Crypto/RSA.py", line 412, in load_key_string
    bio = BIO.MemoryBuffer(string)
  File "/usr/lib64/python2.7/site-packages/M2Crypto/BIO.py", line 187, in __init__
    "data must be bytes or None, not %s" % (type(data).__name__, ))
TypeError: data must be bytes or None, not unicode
```

### Setup
providers.d
```
amazon:
  driver: ec2
  id: xxxxx
  key: xxxxx
  securitygroup: xxxxx
  private_key: xxxxx
  keyname: xxxxx
  ssh_interface: public_ips
  ssh_username: ec2-user
```

profiles.d
```
windows:
  provider: amazon
  image: ami-50e87938
  size: m3.medium
  userdata_file: /etc/salt/windows-firewall.ps1
  win_installer: /root/Salt-Minion-2018.3.1-Py2-AMD64-Setup.exe
  win_username: Administrator
  win_password: auto
```

### Versions Report
```
Salt Version:
           Salt: 2018.3.1
 
Dependency Versions:
           cffi: Not Installed
       cherrypy: Not Installed
       dateutil: Not Installed
[...]

PR #47989, opened 2018-06-06
Properly decode password from aws using m2crypto

dwoz:
M2Crypto requires bytes when loading keys. Fixes issue #47955

### Tests written?

Yes

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/cloud/clouds/ec2.py
+++ b/salt/cloud/clouds/ec2.py
@@ -4762,7 +4762,7 @@ def get_password_data(
             rsa_key = kwargs['key']
             pwdata = base64.b64decode(pwdata)
             if HAS_M2:
-                key = RSA.load_key_string(rsa_key)
+                key = RSA.load_key_string(rsa_key.encode())
                 password = key.private_decrypt(pwdata, RSA.pkcs1_padding)
             else:
                 dsize = Crypto.Hash.SHA.digest_size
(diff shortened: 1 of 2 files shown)


gtmanfred (approved):

rallytime (commented):
This looks good to me but there were several lint errors. I have fixed them here: https://github.com/dwoz/salt/pull/1

(I wanted to keep this PR moving, so the back-port to 2018.3.1 has the original fix, and my lint fix.)

Once that lint fix is merged in, this looks good to me.

salt/cloud/clouds/ec2.py
```diff
@@ -4762,7 +4762,7 @@ def get_password_data(
             rsa_key = kwargs['key']
             pwdata = base64.b64decode(pwdata)
             if HAS_M2:
-                key = RSA.load_key_string(rsa_key)
+                key = RSA.load_key_string(rsa_key.encode())
```
terminalmage: When you do not pass an explicit encoding to `.encode()`, Python 2 will use its default encoding (i.e. `ascii`). Are we certain that the key will only include ascii? If not, we should be using `salt.utils.stringutils.to_bytes(rsa_key)`.

dwoz: Yes, Keys are only ever ascii. We could actually make that an explicit ascii decode.

rallytime (approved):

PR #48035, opened 2018-06-08
Provide python version mismatch solutions

dincamihai:
### What does this PR do?

Improves the error message when Python version mismatch detected by providing solutions.

See below.

### Previous Behavior (Python3 origin - Python2.7 target)

```bash
bash-4.4# salt-ssh -l quiet -i --out json --key-deploy --passwd admin123 container__wjQFc test.ping
{
    "container__wjQFc": {
        "stdout": "ERROR: salt requires python 2.6 or newer on target hosts, must have same major version as origin host",                                                                                                                  
        "stderr": "",
        "retcode": 10
    }
}
```

### New Behavior (Python3 origin - Python2.7 target)

```bash
bash-4.4# salt-ssh -l quiet -i --out json --key-deploy --passwd admin123 container__wjQFc test.ping
{
    "container__wjQFc": {
        "stdout": "ERROR: Depending on the Python version on the target, you need to install python2-salt on origin to add support for Python2.7 targets or install py26-compat-salt on origin to add support for Python2.6 targets or upgrade to Python==3.x on target",
        "stderr": "",
        "retcode": 10
    }
}
```

### Tests written?

No

### Commits signed with GPG?

No

### DIFF
--- a/salt/client/ssh/__init__.py
+++ b/salt/client/ssh/__init__.py
@@ -1393,6 +1393,26 @@ ARGS = {arguments}\n'''.format(config=self.minion_config,
         perm_error_fmt = 'Permissions problem, target user may need '\
                          'to be root or use sudo:\n {0}'
 
+        python_mismatch_options = {
+            '3.x': (
+                'Depending on the Python version on the target, you need to' +
+                ' install Python2.7 compatible salt on origin to add support for Python2.7 targets or' +
+                ' install Python2.6 compatible salt on origin to add support for Python2.6 targets or' +
+                ' upgrade to Python==3.x on target'
+            ),
+            '2.7': (
+                'Depending on the Python version on the target, you need to' +
+                ' upgrade to Python3 on origin to match Python3 on target or' +
+                ' install Python2.7 compatible Salt on origin to add support for Python2.7 targets or' +
+                ' install Python2.6 compatible Salt on origin to add support for Python2.6 targets'
+            ),
+            '2.6': (
+                'Upgrade Python on origin to match the Python version available on target' +
+                ' and install the corresponding Salt package(Python2 compatible Salt or Python3 compatible Salt).'
+            ),
+            'default': 'Matching Python>=2.6 version needed both on origin and target.'
+        }
+
         errors = [
             (
                 (),
@@ -1401,8 +1421,15 @@ ARGS = {arguments}\n'''.format(config=self.minion_config,
             ),
             (
                 (salt.defaults.exitcodes.EX_THIN_PYTHON_INVALID,),
-                'Python interpreter is too old',
-                'salt requires python 2.6 or newer on target hosts, must have same major version as origin host'
+                'Python version mismatched',
+                python_mismatch_options.get(
+                    # check for major.minor first
+                    '%s.%s' % sys.version_info[0:2]
+                ) or python_mismatch_options.get(
+                    # check for major.any
+                    '%s.x' % sys.version_info[0]
+                    # show the default
+                ) or python_mismatch_options['default']
             ),
             (
                 (salt.defaults.exitcodes.EX_THIN_CHECKSUM,),
--- a/salt/client/ssh/__init__.py
+++ b/salt/client/ssh/__init__.py
@@ -1421,13 +1421,13 @@ ARGS = {arguments}\n'''.format(config=self.minion_config,
             ),
             (
                 (salt.defaults.exitcodes.EX_THIN_PYTHON_INVALID,),
-                'Python version mismatched',
+                'Python interpreter is too old',
                 python_mismatch_options.get(
                     # check for major.minor first
-                    '%s.%s' % sys.version_info[0:2]
+                    '{ver[0]}.{ver[1]}'.format(ver=sys.version_info)
                 ) or python_mismatch_options.get(
                     # check for major.any
-                    '%s.x' % sys.version_info[0]
+                    '{ver[0]}.x'.format(ver=sys.version_info)
                     # show the default
                 ) or python_mismatch_options['default']
             ),


cachedout:
@dincamihai There are a few lint errors here: https://jenkins.saltstack.com/job/PR/job/salt-pr-lint-n/22487/violations/file/salt/client/ssh/__init__.py/

terminalmage (commented):
I'm not sure what this solves. @isbm already has done a lot of work on cross-version support (see https://github.com/saltstack/salt/pull/46684).

gtmanfred (approved):
This looks like it just fixes the version mismatch error, and should probably be backported to 2017.7 and 2018.3

salt/client/ssh/__init__.py
```diff
@@ -1401,8 +1421,15 @@ def categorize_shim_errors(self, stdout_bytes, stderr_bytes, retcode):
...
-                'salt requires python 2.6 or newer on target hosts, must have same major version as origin host'
+                'Python version mismatched',
+                python_mismatch_options.get(
+                    # check for major.minor first
+                    '%s.%s' % sys.version_info[0:2]
+                ) or python_mismatch_options.get(
+                    # check for major.any
+                    '%s.x' % sys.version_info[0]
+                    # show the default
+                ) or python_mismatch_options['default']
```
isbm: Please do not use `'%s.%s' % foo`, but `'{}.{}'.format(...)` instead, as per internal Salt agreements. As well as we probably do not need "3.x" but can be just "3". Also I would do it less complex, but use `.get(...)` as intended:
```python
py_mismatch.get('{}.{}'.format(*sys.version_info[:2]),
                py_mismatch.get(str(sys.version_info[0]),
                py_mismatch['default']))```

File: salt/client/ssh/__init__.py [hunk 2]
```diff
@@ -1393,6 +1393,26 @@ def categorize_shim_errors(self, stdout_bytes, stderr_bytes, retcode):
...
+                'Depending on the Python version on the target, you need to' +
+                ' upgrade to Python3 on origin to match Python3 on target or' +
+                ' install Python2.7 compatible Salt on origin to add support for Python2.7 targets or' +
+                ' install Python2.6 compatible Salt on origin to add support for Python2.6 targets'
+            ),
+            '2.6': (
+                'Upgrade Python on origin to match the Python version available on target' +
+                ' and install the corresponding Salt package(Python2 compatible Salt or Python3 compatible Salt).'
+            ),
+            'default': 'Matching Python>=2.6 version needed both on origin and target.'
```
isbm: Generally I am OK with this, except this will be like an info on the CLI. So maybe either rephrase this not to repeat the same thing twice, or make it as a list of actions (and less verbose). Thus calling `salt-ssh` would return something looking like this on the terminal:

```
Python version error. Your options:
- Install Salt for Python 2.7 on a Master
- Install Salt for Python 2.6 on a Master
- Upgrade to Python 3.x on the client machine
```

...etc. It would be easier to comprehend and quickly understand the problem. Like we do with Zypper. :wink:

isbm (changes requested):

isbm:
@terminalmage I've forgot the error message, which is "invalid Python". Salt SSH still needs to have both library sets on the Master, so the `.tar.gz` is carrying over those Python specific versions.

@dincamihai NOTE: actually if you are running Salt on a Master from the specific version, you _do not_ need to install Salt again for the alternative version, as it is anyway works for both versions. This is just packaging convenience. But in fact you need only couple of version-specific libraries to be installed so they will be picked up by the `thin` creation procedure. That is, probably we should not say "install Salt for alternative Python X" (which implies you will get all the needed dependencies automatically), but "install Salt _dependencies_ only for the alternative Python X".

dincamihai:
@cachedout i've fixed the lint errors but now, after rebase, it seems to have some unrelated lint errors in the tests.

@terminalmage Bo and Daniel already provided an explanation. sorry, for not being clear enough about this in the description.

isbm:
@dincamihai I would still minimise the text according to the example I showed above. As well as the "3.x" thing.

isbm:
@dincamihai the "3.x" is the same as "3", as I see it. I also suggested to use integers as in the `version_info` instead of just strings. And so if we have some issue with the 3.8 (e.g.) one would just add a minor version key. Otherwise default should go. But "3.x" is more to me looks like a hack.

dincamihai:
@isbm i'm checking first for major.minor, so if we add 3.8 it will go to that first

isbm:
@dincamihai correct. So that means there **is something** that needs to be shown _instead of_ the default message. And if you have none of '8', then you will get the default one:

```python
msg = {
    2: {
        6: 'Too old',
        7: 'Soon to be dead',
        'default': 'Problem between the keyboard and the chair',
    },
    3: {
        4: 'Many things are missing in this release',
        6: 'Many things are incompatible with the previous release, ha-ha!',
        8: 'Larry Wall was hired to design new Python language syntax',
        'default': 'I like you begging, do it again',
    }
}

import sys
mj, mn = sys.version_info[:2]
print(msg[mj].get(mn, msg[mj]['default']))
```

dincamihai:
@isbm if there is no 3.8 specific message, 3.x would be shown. is this not enough?

isbm:
The code I exampled above does this. I am just in favour of your structure that keeps strings that needs to be parsed instead of map to versions.

dincamihai:
@isbm ok. thanks for the suggestion, but if the upstream is fine with my version (changes approved) please merge the PR. thanks!

isbm:
@terminalmage you like the proposed error messages and the structure?

brejoc (approved):
Using the terms master and minion is imho misleading, since we don't necessarily have a master here. Just a computer that targets some others. So I would say origin and target makes more sense.  I also think we should be fine regarding the version numbers here.

meaksh (approved):
Looks good to me. FMPOV it's ok either keeping the current structure (sparce) or changing it as Bo is suggesting (condensed).

About using "origin" or "Master" I would prefer to keep using "origin/target" which is more generic than assuming the salt-ssh host is necessary the Salt Master.

isbm:
OK, as per speaking with @dincamihai it would be also an option to:

- Add all the options into the documentation.
- Add short minimally intrusive to the target machine option (i.e. update the SaltSSH side w/o touching the target)

That said, CLI would return something like:
```
The "my-to-be-minion.greatdomain.com" machine is running Python 2.6 version.
Please install Salt for 2.6 Python on the Salt SSH machine and set it up.
```

or:
```
The "my-to-be-minion.greatdomain.com" machine is running Python 2.7 version.
Please install Salt for 2.7 Python on the Salt SSH machine.
```

So such info on the CLI would suggest what to do now, quickly and easiest way. And the following would go to the documentation (rephrased into more extended version):

"Depending on the Python version on the target, you need to install Python2.7 compatible salt on origin to add support for Python2.7 targets or install Python2.6 compatible salt on origin to add support for Python2.6 targets or upgrade to Python==3.x on target"

terminalmage (changes requested):
I don't understand the reasoning of "this will prevent confusion between a traditional Master and Minion". How will it do that? by using a different nomenclature ("origin" and "target"), which is used _nowhere else in Salt_? How is that any _clearer_?

If you're using salt-ssh, I think it's pretty well understood that you're not using a "traditional master-and-minion".

Even if it isn't 100% clear, _no one has explained why that even matters_. salt-ssh reads its configuration from /etc/salt/master, it targets machines like a master, and it stores results in a job cache.... _like a master_. I don't understand what end is served by introducing new terminology, when we could just call it what it is.

Also, I prefer @isbm's solution to storing the warning messages. Why construct a string using a `.format()` to make a dictionary key, when you can organize the dictionary containing the messages in such a way to make it unnecessary?

dincamihai:
One note:
sometimes, when targeting 1000 machines, maybe 999 are ok but one has an old python. there are two options and we don't know what it is better for the user:
 - update the origin machine (that works on 999 of the clients) or
 - update the one client that doesn't work

Other than that, this is open source, so I don't have any problem with adapting my PR to whatever is a better fit for saltstack/salt.

isbm:
I would only say that "target" is that part between `salt` and the `command`... More here: https://docs.saltstack.com/en/latest/topics/targeting/ So calling it "target" and "origin" brings lots of confusion from this POV.  I would definitely keep it Master, especially if SaltSSH is "to execute salt commands and states over ssh without installing a salt-minion". Which implies "Master" at first place!

dincamihai:
@terminalmage i am not an expert in salt terminology. I just proposed what I thought it would be appropriate and so far saltstack did not disagree. sure, please propose better messages.
the changes were already approved so why should i make more changes to the PR or invest more time into it?

dincamihai:
@terminalmage by the way https://cse.google.com/cse?cx=004624818632696854117:yfmprrbw3pk&q=target

isbm:
@dincamihai actually "approved" here means _everyone_ agrees. JFYI.

dincamihai:
@isbm and @terminalmage I have tested my changes on different combinations.
I would not make any sense to make additional changes and test the whole thing again with python2.6, 2.7 and 3 and combinations of them just because someone likes it more is some way. (i'm not talking here about the formulation of the messages, that is a valid change that is worth doing). but changing some structure to another structure just because... is not enough reason to redo the whole testing again.

dincamihai:
@isbm what about target in the context of rosters https://docs.saltstack.com/en/latest/topics/ssh/roster.html

brejoc:
😉🍿

terminalmage:
I would have loved to work with you, but since you were not willing to accept any recommendations, I am closing this and we will do any remaining work in https://github.com/saltstack/salt/pull/48058.

PR #48041, opened 2018-06-08
salt.utils.hashutils: Only decode to utf-8 on Windows

terminalmage:
We only decode to unicode so that we can use some of these functions in Jinja filters. Since CP1252 will incorrectly decode certain bytestrings, force utf8 decoding on Windows.

### DIFF
--- a/salt/utils/hashutils.py
+++ b/salt/utils/hashutils.py
@@ -13,6 +13,7 @@ import random
 # Import Salt libs
 from salt.ext import six
 import salt.utils.files
+import salt.utils.platform
 import salt.utils.stringutils
 
 from salt.utils.decorators.jinja import jinja_filter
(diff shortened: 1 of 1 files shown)


twangboy (approved):

damon-atkins:
Keep in mine https://en.wikipedia.org/wiki/Base64 restrictions. i.e. Character Set Allowed, hence its not just Windows.

terminalmage:
@damon-atkins Can you be a little more clear? I don't know what you're trying to sayh.

terminalmage:
I don't think this is needed on non-Windows platforms. The reason for doing this is because CP1252 is awful and will successfully decode bytestrings that it shouldn't, with corrupted results. UTF-8, latin-1, ascii, all don't have this problem. Forcing use of UTF-8 when decoding the bytestring on Windows keeps us from trying CP1252.

rallytime (approved):

PR #48046, opened 2018-06-09
SmartOS imgadm module and state did not like docker images

sjorge:
### What does this PR do?
This PR fixes the importing and managing of docker images on SmartOS.

N/a

### Previous Behavior
Exceptions were encountered when trying to import docker images.

### New Behavior
Docker images can now be imported and managed properly.

### Tests written?
No

### Commits signed with GPG?
No

### Backporting
It's a rather large fix so I would prefer to keep it in develop only, although there is no functional change compared to 2018.3.x so it may be worth while pulling this in.

### DIFF
--- a/salt/modules/smartos_imgadm.py
+++ b/salt/modules/smartos_imgadm.py
@@ -25,38 +25,6 @@ __func_alias__ = {
 __virtualname__ = 'imgadm'
 
 
-def _exit_status(retcode):
-    '''
-    Translate exit status of imgadm
-    '''
-    ret = {0: 'Successful completion.',
-           1: 'An error occurred.',
-           2: 'Usage error.',
-           3: 'Image not installed.'}[retcode]
-    return ret
-
-
-def _parse_image_meta(image=None, detail=False):
-    if not image:
-        return {}
-
-    if detail:
-        return {
-            'name': image['manifest']['name'],
-            'version': image['manifest']['version'],
-            'os': image['manifest']['os'],
-            'description': image['manifest']['description'],
-            'published': image['manifest']['published_at'],
-            'source': image['source']
-        }
-    else:
-        return '{name}@{version} [{date}]'.format(
-                name=image['manifest']['name'],
-                version=image['manifest']['version'],
-                date=image['manifest']['published_at'],
-            )
-
-
 def __virtual__():
     '''
     Provides imgadm only on SmartOS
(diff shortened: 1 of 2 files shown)


sjorge:
I will be super busy the coming week/weekend, but smartos_vmadm should also be updated to not error on docker images. Fixing this is a requirement for smartos_vmadm, I will deal with smartos_vmadm once this is merged and I am less busy.

sjorge:
Lint and test failures seem to be unrelated.

salt/modules/smartos_imgadm.py
```diff
@@ -150,25 +232,33 @@ def list_installed(verbose=False):
     verbose : boolean (False)
         toggle verbose output
 
+    .. versionchanged:: 2016.3.2
```
rallytime: This should be `Fluorine`, yes?

salt/modules/smartos_imgadm.py
```diff
@@ -89,6 +153,22 @@ def version():
     return ret[-1]
 
 
+def docker_to_uuid(uuid):
+    '''
+    Get the image uuid from an imported docker image
+
+    .. versionadded:: Neon
```
rallytime: This should be Fluorine.

salt/modules/smartos_imgadm.py
```diff
@@ -308,4 +415,91 @@ def vacuum(verbose=False):
         return list(result.keys())
 
 
+def sources(verbose=False):
+    '''
+    Return a list of available sources
+
```
rallytime: This new function, as well as all of the the other new functions below, should have a versionadded tag for `Fluorine`. :)

rallytime (changes requested):
Some minor documentation requests from me. :)

salt/modules/smartos_imgadm.py
sjorge: Done

salt/modules/smartos_imgadm.py
sjorge: Done

salt/modules/smartos_imgadm.py
sjorge: Done

sjorge:
Done and rebased on develop

cachedout (approved):
Very nice.

rallytime (approved):
👍

PR #48049, opened 2018-06-11
modules.ebuild Use standard emaint module for pkg sync

jfindlay:
### What does this PR do?
`eix-sync` does not support all of the functionality available in the relatively new sync plugin architecture, such as only syncing auto-sync enabled repos.  Also add a minion config to prevent syncing the portage source tree if the last sync was done less than 24 hours prior.

References:
- [New portage plug-in sync system](https://gentoo.org/support/news-items/2015-02-04-portage-sync-changes.html)
- [Project:Portage/Sync](https://wiki.gentoo.org/wiki/Project:Portage/Sync#Operation)
- [Handbook:Gentoo ebuild repository](https://wiki.gentoo.org/wiki/Handbook:AMD64/Full/Installation#Gentoo_ebuild_repository)
- [eix:Managing the cache](https://wiki.gentoo.org/wiki/Eix#Managing_the_cache)
- [Rsync etiquette OR How often should I sync?](https://forums.gentoo.org/viewtopic-t-133158.html)

None

### Tests written?

No

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/modules/ebuild.py
+++ b/salt/modules/ebuild.py
@@ -16,6 +16,7 @@ i.e. ``'vim'`` will not work, ``'app-editors/vim'`` will.
 from __future__ import absolute_import, print_function, unicode_literals
 
 # Import python libs
+import os
 import copy
 import logging
 import re
@@ -437,7 +438,13 @@ def list_pkgs(versions_as_list=False, **kwargs):
 
 def refresh_db():
     '''
-    Updates the portage tree (emerge --sync). Uses eix-sync if available.
+    Update the portage tree using the first available method from the following
+    list:
+
+    - emaint sync
+    - eix-sync
+    - emerge-webrsync
+    - emerge --sync
 
     CLI Example:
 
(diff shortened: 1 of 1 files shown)


salt/modules/ebuild.py
```diff
@@ -437,36 +439,59 @@ def list_pkgs(versions_as_list=False, **kwargs):
...
+    - emerge-webrsync
+    - emerge --sync
+
+    If the ``portage.sync_wait`` minion config is set, the portage tree will
+    not be synced within one day of the previous sync.  To enable this feature,
+    place the following config into /etc/salt/minion:
+
+    .. code-block:: yaml
+
+        portage.sync_wait: True
```
cachedout: I wonder if it would be more flexible to use a pillar variable here instead?

jfindlay: Good idea.

cachedout (approved):

PR #48058, opened 2018-06-11
Show recommendations for salt-ssh cross-version python errors

terminalmage:
This shows more accurate information on how to resolve version issues (e.g. master only has Salt deps installed for Python 3 but remote host has no Python 3 installed).

### DIFF
--- a/salt/client/ssh/__init__.py
+++ b/salt/client/ssh/__init__.py
@@ -1393,6 +1393,28 @@ ARGS = {arguments}\n'''.format(config=self.minion_config,
         perm_error_fmt = 'Permissions problem, target user may need '\
                          'to be root or use sudo:\n {0}'
 
+        def _version_mismatch_error():
+            messages = {
+                2: {
+                    6: 'Install Python 2.7 / Python 3 Salt dependencies on the Salt SSH master \n'
+                       'to interact with Python 2.7 / Python 3 targets',
+                    7: 'Install Python 2.6 / Python 3 Salt dependencies on the Salt SSH master \n'
+                       'to interact with Python 2.6 / Python 3 targets',
+                },
+                3: {
+                    'default': '- Install Python 2.6/2.7 Salt dependencies on the Salt SSH \n'
+                               '  master to interact with Python 2.6/2.7 targets\n'
+                               '- Install Python 3 on the target machine(s)',
+                },
+                'default': 'Matching major/minor Python release (>=2.6) needed both on the Salt SSH \n'
+                           'master and target machine',
+            }
+            major, minor = sys.version_info[:2]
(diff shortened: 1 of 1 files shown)


dincamihai:
@terminalmage why now the term `target` is ok and here it was not? https://github.com/saltstack/salt/pull/48035

dincamihai:
@terminalmage just for the record, i was open to adapt my PR but if there was a good reason. why is "your structure" (proposed by @isbm) better?

dincamihai (commented):
I would like clarifications for closing my PR (https://github.com/saltstack/salt/pull/48035) (approved by 3 engineers 1 Saltstack, 2 Suse) before merging this one because the proposed actions are not really accurate all the time. see https://github.com/saltstack/salt/pull/48035#issuecomment-396248698

salt/client/ssh/__init__.py
```diff
@@ -1393,6 +1393,28 @@ def categorize_shim_errors(self, stdout_bytes, stderr_bytes, retcode):
...
+                               '  master to interact with Python 2.6/2.7 targets\n'
+                               '- Install Python 3 on the target machine(s)',
+                },
+                'default': 'Matching major/minor Python release (>=2.6) needed both on the Salt SSH \n'
+                           'master and target machine',
+            }
+            major, minor = sys.version_info[:2]
+            help_msg = messages.get(major, {}).get(minor) \
+                or messages.get(major, {}).get('default') \
+                or messages['default']
```
isbm: I would remove that ~ugly~ backslash here, as per [line continuation style guide](http://docs.python-guide.org/en/latest/writing/style/#line-continuations), but it is up to you.

isbm (approved):

terminalmage:
@dincamihai Actually, I had meant that referring to an "origin and target" is something we don't use together anywhere. But you never made an effort to clarify, you just argued.

gtmanfred (approved):

isbm:
@dincamihai I am not sure that your situation is actually correct, as any admin should anyways go with the most less intrusive default option to the targeted machine, as it is the most important entity, running business UAT or production application, regardless if is such and just a development environment instead. Normally the Master machine intended to be a "workhorse" station that can change its state frequently, according to the infrastructure needs without harming anything important. So if one of the managed systems is still running an old Python, an admin should choose temporarily use compatibility package, upgrade the Master machine in order to adjust it and so then later he can dispose the compatibility package, once it is no longer needed (as per ISO 9001 requirements, actually).

The rest of the choices are up to the policies of the admin who is looking after the infrastructure and can be described in a documentation.

@terminalmage what you think?

dincamihai:
@terminalmage ok, i'm glad i've started the discussion and that the problem is solved. great teamwork.

PR #48060, opened 2018-06-11
mark test as expensive

gtmanfred:
### What does this PR do?
This tests takes a large amount of cpu to run, and is causing tests to
hang.

Mark it as an expensive test so it only runs with the cloud tests.

Trying to clean up some tests that might be causing 2018.3 to take a long time to run.

### Commits signed with GPG?

Yes

### DIFF
--- a/tests/integration/runners/test_state.py
+++ b/tests/integration/runners/test_state.py
@@ -20,7 +20,7 @@ from salt.ext.six.moves import queue
 from tests.support.case import ShellCase
 from tests.support.unit import skipIf
 from tests.support.paths import TMP
-from tests.support.helpers import flaky
+from tests.support.helpers import flaky, expensiveTest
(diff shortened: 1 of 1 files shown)


rallytime (approved):

cachedout:
Is this just looping through the `while True:` loop without sleeping? It probably just needs a 50ms sleep or something in there and it should be fine.

gtmanfred:
I tried that, and it did not reduce the cpu/memory consumption.

It is actually the same way that the rest of the event listening stuff is running.

https://github.com/saltstack/salt/blob/2017.7/salt/modules/state.py#L2018

rallytime:
I will say too that this is the best test result we've seen on this branch for a while, FWIW.

garethgreenaway (approved):

PR #48062, opened 2018-06-12
salt/utils/cloud: Allow to customize ssh gateway command/options

icy:
### What does this PR do?

By default, nc command (`nc -q0 %h %p`) is used to create gateway. This is not
always possible in some restricted environments. An alternative is to use
native support from ssh: `-W %h:%p`.

This commit allows user to provide a customize option for gateway. Please note
the (wait_for_port) process may need another patch.

### Previous Behavior

Users are forced to use `nc -q0 %h %p` command to create ssh proxy for `salt-cloud`

### New Behavior

Users can use a customized command to create ssh proxy for `salt-cloud`

### Tests written?

NO

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/cloud/clouds/ec2.py
+++ b/salt/cloud/clouds/ec2.py
@@ -48,6 +48,10 @@ To use the EC2 cloud module, set up the cloud configuration at
       # Optional
       ssh_gateway_username: root
 
+      # Default to nc -q0 %h %p
+      # Optional
+      ssh_gateway_command: "-W %h:%p"
+
       # One authentication method is required. If both
       # are specified, Private key wins.
 
--- a/salt/cloud/clouds/ec2.py
+++ b/salt/cloud/clouds/ec2.py
@@ -1082,6 +1082,12 @@ def get_ssh_gateway_config(vm_):
         search_global=False
     )
 
+    # ssh_gateway_command
+    ssh_gateway_config['ssh_gateway_command'] = config.get_cloud_config_value(
+        'ssh_gateway_command', vm_, __opts__, default=None,
+        search_global=False
+    )
+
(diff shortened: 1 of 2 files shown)


doc/man/salt.7
```diff
@@ -80063,6 +80063,10 @@ my\-ec2\-config:
   # Optional
   ssh_gateway_username: root
 
+  # Default to nc -q0 %h %p
+  # Optional
+  ssh_gateway_command: "-W %h:%p"
+
```
gtmanfred: Can you remove this change, we compile these every time we build, so this will not remain.

gtmanfred (changes requested):
This should be documented in salt/cloud/clouds/ec2.py, as well as adding the gateway stuff to ./doc/topics/cloud/misc.rst

doc/man/salt.7
icy: thanks a lot. I will update `ec2.py` and `misc.rst` instead :)

gtmanfred (approved):

cachedout (approved):

icy:
Thanks @gtmanfred and @cachedout  for your reviews and approvals.

rallytime:
Thanks @icy and congrats on your first salt PR. :)

icy:
Hi @rallytime and @gtmanfred,

Is there any good start to write and improve unit tests (e.g., in my PRs)?

At my laptop, the tests are passed as seen here https://gist.github.com/icy/a59f975ba54224b7d941b2e38ccc74a1 , I don't think these tests are working fine. I did some smoke tests actually and it's not a recommended way (expensive?)

PR #48068, opened 2018-06-12
Fix {set/get}_zone NILinuxRT CurrentGen

rares-pop:
On legacy version of NILinuxRT, /etc/localtime is a
symlink to /etc/natinst/share/localtime.
On this systems, {set/get}_zone must work directly on
/etc/natinst/share/localtime, instead of default file.

### What does this PR do?
National Instruments supports two Linux RT distributions, and the former one, also known as CurrentGen works directly with /etc/natinst/share/localtime for the timezone.

### Previous Behavior
Timezone wasn't working correctly

### Tests written?

No

### Commits signed with GPG?

No

### DIFF
--- a/doc/topics/cloud/aws.rst
+++ b/doc/topics/cloud/aws.rst
@@ -57,6 +57,10 @@ parameters are discussed in more detail below.
       id: 'use-instance-role-credentials'
       key: 'use-instance-role-credentials'
 
+      # If 'role_arn' is specified the above credentials are used to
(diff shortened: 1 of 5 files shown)


cachedout (approved):

rallytime:
@rares-pop The timezone tests didn't love this change. Can you take a look?

rares-pop:
Yes, I will look on Monday.

PR #48070, opened 2018-06-12
[RFC] Refactor error messages to have better documented errors and to allow packagers to customize them in an easy way

dincamihai:
### What does this PR do?

Proposes an idea to refactor the error messages in order to provide an easy way for packagers to customize them.

### DIFF
--- a/rfcs/0001-error_messages.md
+++ b/rfcs/0001-error_messages.md
@@ -1,0 +1,48 @@
+- Feature Name: Allow packagers to have custom error/suggestion messages
+- Start Date: Tue 12 Jun, 2018
+- RFC PR:
+- Salt Issue:
+
+# Summary
+[summary]: #summary
+
+Extract error messages in separate files to allow packagers to customize them
+
+# Motivation
+[motivation]: #motivation
+
+Sometimes we need to have custom error/suggestions messages in order to point to the right package names depending on the context (eg: OS)
+
+# Design
+[design]: #detailed-design
+
+### Current Design
+
+At the moment, the errors are spread within the code.
+
+### Proposed Design
+
+The simplest solution would be to extract all of them in a separate file.
+They could be also be organized by category.
+We could also assign unique IDs in order to facilitate documenting them.
+
+### Notes
+
+There might be cases where the messages need to contain some dynamic parts in which case we could store them as string templates and populate them before displaying or find a better solution if needed.
+
+## Alternatives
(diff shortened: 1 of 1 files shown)


brejoc:
How about using something like gettext? This way the error messages could also be localized.

dincamihai:
@cachedout the simplest thing could be to have tuples of strings but maybe I better idea would be to implement a parent exception class and override some of the member functions so that it would ask for the right parameters (if needed) in order to render the output message.

tuples are simple and straight forward to implement but also starting with a simple exception class would not be that complicated (and extend it later when we know more about what is needed)
What do you think it would be best?

isbm:
I would use just [gettext](https://docs.python.org/3/library/gettext.html) for this. We would be able to shoot two rabbits with one bullet: i18n and vendor supported messages.

gtmanfred (approved):
I am for centralizing error messages.

isbm (changes requested):
@dincamihai looks good, but we need to specify here how exactly vendors can override the messages. And not just error messages, but all of 'em, if they need to. That way we could also get i18n/l10n done almost for free.

But I strongly suggest to use gettext, since not it is just century old reliable stuff, but also it sucks to write IDs instead of real English in the code.

terminalmage (approved):

rallytime:
@dincamihai Did you want to make the change above about the specifications @isbm noted above? Once that is done, I think we can get this merged in.

dincamihai:
@rallytime please give me a couple of days. I was on vacation. I would like to incorporate the feedback and improve the RFC a bit. Maybe even a proposed implementation.
I'll add the WIP.
Thanks!

dincamihai:
@isbm @brejoc @cachedout I have updated my proposal to use gettext. I'm also providing an example of how this would be implemented and how it would work. Please take a look when you have time. Thanks!

isbm:
We are using throwaways all around the code. Maybe:

```python
from salt.utils.gettext import gt
...
print(gt('something'))
```

...?

dincamihai:
@isbm that makes sense. we don't really need to keep `_` in the real implementation. I can also change the examples in the RFC.
Do we agree to have this indirect import from salt.utils?

dincamihai:
I have updated the example to incorporate the suggestion from @isbm 
@rallytime this could be merged if there is no objection. thank you!

cachedout (approved):

PR #48072, opened 2018-06-12
Fix system.set_computer_name on NILinuxRT CurrentGen

rares-pop:
On Current-Gen, we also need to modify the bootloader variable
`hostname`. This is due to `populateconfig` calling `nirtcfg --migrate` on
startup so that the bootloader variables end up in `ni-rt.ini`.
`ni-rt.ini` is the source used to set the hostname when the `hostname`
init script is invoked.

Signed-off-by: Sergey Kizunov <sergey.kizunov@ni.com>

### What does this PR do?
National Instruments supports two Linux RT distributions, and the former one, also known as CurrentGen is using an utility 'nirtcfg' for storing system settings in order to be persisted across reboots.

Without this change, any system setting change won't be persisted across reboots.

### Previous Behavior
Any system setting change isn't persisted across reboots.

### New Behavior
Settings are persisted.

### Tests written?

No

### Commits signed with GPG?

No

### DIFF
--- a/salt/modules/network.py
+++ b/salt/modules/network.py
@@ -1354,6 +1354,12 @@ def mod_hostname(hostname):
     elif __grains__['os_family'] in ('Debian', 'NILinuxRT'):
         with salt.utils.files.fopen('/etc/hostname', 'w') as fh_:
             fh_.write(salt.utils.stringutils.to_str(hostname + '\n'))
+        if __grains__['lsb_distrib_id'] == 'nilrt':
+            str_hostname = salt.utils.stringutils.to_str(hostname)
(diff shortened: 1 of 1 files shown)


garethgreenaway:
Looks good.  The only question I had was the line with the path:
`nirtcfg_cmd = '/usr/local/natinst/bin/nirtcfg'`
Is that location in the path of the user that will be running Salt?  Can we use `salt.utils.path.which` so that it's not hard coded?

rares-pop:
/usr/local/natinst/bin isn't always in the PATH, that's why it's safer to stick to the hardcoded string. It would be very difficult to have that always in the PATH especially for older targets

cachedout (approved):

PR #48077, opened 2018-06-12
Fix issue with `salt.utils.parsers` on Windows

twangboy:
### What does this PR do?
Fixes an issue with `salt.utils.parsers` on Windows. There is no `os.getuid` function in Windows. The `_mixin_before_exit` function uses `os.getuid` to check for root which fails on Windows. This function detects root/admin differently if it's on a Windows box. Both methods are also mocked properly in the tests.

Found via failing tests on Jenkins

### Previous Behavior
The following tests were failing:
unit.utils.test_parsers.DaemonMixInTestCase.test_pid_deleted_oserror_as_non_root
unit.utils.test_parsers.DaemonMixInTestCase.test_pid_deleted_oserror_as_root

### New Behavior
The tests now pass.

### Tests written?
Yes

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/utils/parsers.py
+++ b/salt/utils/parsers.py
@@ -968,17 +968,17 @@ class DaemonMixIn(six.with_metaclass(MixInMeta, object)):
                     # Log error only when running salt-master as a root user.
                     # Otherwise this can be ignored, since salt-master is able to
                     # overwrite the PIDfile on the next start.
-                    def log_error():
-                        logger.info('PIDfile could not be deleted: %s',
-                                    six.text_type(self.config['pidfile']))
-                        logger.debug(six.text_type(err))
+                    err_msg = ('PIDfile could not be deleted: %s',
+                               six.text_type(self.config['pidfile']))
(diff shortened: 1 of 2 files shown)


salt/utils/parsers.py
```diff
@@ -968,9 +968,17 @@ def _mixin_before_exit(self):
                     # Log error only when running salt-master as a root user.
                     # Otherwise this can be ignored, since salt-master is able to
                     # overwrite the PIDfile on the next start.
-                    if not os.getuid():
-                        logger.info('PIDfile could not be deleted: %s', six.text_type(self.config['pidfile']))
+                    def log_error():
```
cachedout: Instead of the overhead of a function call here, I think it would be better just to define the error string and then call logger.info in each case, IMHO.

twangboy:
@cachedout Moved the log calls as requested. Got rid of the function

cachedout (approved):
Very nice. This is quite clean.

Issue #47984, opened 2018-06-05
x509 module/state writing wrong certificate serial number to CRL 


jeduardo:
### Description of Issue/Question

CRL files created through the Salt x509 module store a wrong serial number for the certificates to be revoked, causing revoked certificates to be recognized as valid and removing the effectiveness of the CRL created. 

### Setup

* Master peer configuration:

```yaml
peer:
  .*:
    - x509.sign_remote_certificate
```

* Minion signing policies:

```yaml
x509_signing_policies:                                                       
  # Default policy for server certificates.                                  
  # Here we allow a certificate to be used both as a server and              
  # machine-client certificate.                                              
  server:                                                                    
    - minions: '*'                                                           
    - signing_private_key: /srv/pki/intermediateca.key                       
    - signing_cert: /srv/pki/intermediateca.pem                              
    - C: DE                                                                  
    - ST: City                                                            
    - L: City                                                             
    - basicConstraints: "critical CA:false"                                  
    - keyUsage: "critical digitalSignature, keyEncipherment, dataEncipherment, keyAgreement"
    - extendedKeyUsage: "critical serverAuth"                                
    - subjectKeyIdentifier: hash                                             
    - authorityKeyIdentifier: keyid,issuer:always                            
    - days_valid: 90                                                         
    - copypath: /srv/pki/issued_certs/                                       
    - nsCertType: server                                                     
    - crlDistributionPoints: "URI:http://pki.mysite.org/intermediate.crl"
```

* State to issue Root CA:

```yaml
/srv/pki/rootca.key:
  x509.private_key_managed:
    - bits: 4096
    - backup: True
    - require:
      - file: /srv/pki    
    {%- if salt['file.file_exists']('/srv/pki/rootca.key') %}
    - prereq:
      - x509: /srv/pki/rootca.pem
    {%- endif %}

/srv/pki/rootca.pem:
  x509.certificate_managed:
    - signing_private_key: /srv/pki/rootca.key
    - CN: Root CA
    - C: DE
    - ST: City
    - L: City
    - basicConstraints: "critical CA:true"
    - keyUsage: "critical digitalSignature, cRLSign, keyCertSign"
    - subjectKeyIdentifier: hash
    - authorityKeyIdentifier: keyid,issuer:always
    - crlDistributionPoints: "URI:http://pki.mysite.org/root.crl"
    - days_valid: 3650
    - days_remaining: 30
    - backup: True

publish updated salt mine with root ca:
  module.run:
[...]

PR #48080, opened 2018-06-13
Bugfix #47984 messed up cert serial

lusche:
### What does this PR do?

Fix the Problem described in #47984 in the 2017 branch

#47984

### Tests written?

No

### Commits signed with GPG?

No

### DIFF
--- a/salt/modules/x509.py
+++ b/salt/modules/x509.py
@@ -942,7 +942,6 @@ def create_crl(  # pylint: disable=too-many-arguments,too-many-locals
             rev_item['not_after'] = rev_cert['Not After']
 
         serial_number = rev_item['serial_number'].replace(':', '')
-        serial_number = str(serial_number)
 
         if 'not_after' in rev_item and not include_expired:
             not_after = datetime.datetime.strptime(
--- a/salt/modules/x509.py
+++ b/salt/modules/x509.py
@@ -942,7 +942,7 @@ def create_crl(  # pylint: disable=too-many-arguments,too-many-locals
             rev_item['not_after'] = rev_cert['Not After']
 
         serial_number = rev_item['serial_number'].replace(':', '')
-        serial_number = str(int(serial_number, 16))
+        serial_number = str(serial_number)
 
         if 'not_after' in rev_item and not include_expired:
             not_after = datetime.datetime.strptime(


cachedout:
I would rather we just backport #47984 instead. @rallytime what do you think?

lusche:
I'm sure that is possible but some parts of the PR are not available in the 2017 branch.

For example: ```import salt.utils.stringutils``` 

To be honest my fix is quick an not so "nice" but should do the job. But if anyone is willing to backport the original fix that's ok to.

 I hope that this issue gets a fix soon. At this time we had a ugly workaround to make it work.

jeduardo:
Hey all, chiming in here :)

My original fix actually targeted at 2017.7 and was even simpler than the one suggested here. Removing the offending line `serial_number = str(int(serial_number, 16))` was enough to get the correct serial into the CRL. This is what is running on my 2017.7 environment though the magic of `_modules/x509.py`.

However as current Salt has unicode strings everywhere (which is a _Good Thing(tm)_) simply removing the line ended up not being enough.

cachedout:
@jeduardo Thanks for the info. Are you in favor of this being merged then? I couldn't quite get your opinion from your comment.

jeduardo:
@cachedout I've meant to say that the patch is in the right direction but can be simplified by removing the buggy line altogether instead of replacing it by `serial_number = str(serial_number)`. 

The `serial_number` variable will already be a string at that point in the code (as its value is returned from string.replace() as a string already), thus making the `str(serial_number)` call redundant.

rallytime:
Ah, I missed my original ping somehow. I can backport #47986 to the 2018.3 branch, but as stated above, it won't back-port nicely to 2017.7. I'll do the 2018.3 back-port right now so it will be included in 2018.3.3.

In the mean time, @lusche - what do you think about @jeduardo's latest comment.

lusche:
@rallytime i pushed the suggested changes. @jeduardo hint sounds plausible.

garethgreenaway:
In 2017.7 we should be fine just removing that line since we're not enforcing Unicode like we are in 2018.3.  This looks good to me.

garethgreenaway (approved):

jeduardo:
@rallytime well, @garethgreenaway already approved it but indeed it does look just like the patch I initially did when I've faced the problem on 2017.7. 

I'm also really glad this is getting in as we are not yet ready to migrate to 2018.3 at $WORKPLACE. :)

PR #48103, opened 2018-06-13
Fix for gitfs base env being pinned to commit ID

terminalmage:
This worked when done through per-saltenv configuration (which causes a
warning to be logged), but not when configured like it is supposed to.


rallytime (approved):

rallytime:
re-run py

cachedout (approved):

PR #48104, opened 2018-06-13
Fix `unit.modules.test_win_status`

twangboy:
### What does this PR do?
Fixes an issue with the test for `win_status`.

Jenkins

### Previous Behavior
In the transition to unicode, some of the log messages changed from the `'{0}'.format('something')` format to `'this %s', 'something'` format. In the test it was checking that the `log.warning` was called with a single item which works with the `.format` method. The new method considers each element and item, so the test would fail.

### New Behavior
Add's an additional `ANY` to the assertion in the test. Also adds the `self.expected_error_code` to the first test which removes the need for the second assertion (`assertIn`)

### Tests written?
Yes

### Commits signed with GPG?
Yes


rallytime (approved):

rallytime:
re-run py

cachedout (approved):
Seems reasonable.

PR #48116, opened 2018-06-13
Add jenkinsfiles to define tests in 2018

KaiSforza:
### What does this PR do?
Adds jenkinsfiles to the `.ci/` directory, allowing us to use Jenkins Pipelines.

### Previous Behavior
Manually specified jobs on our Jenkins node

### New Behavior
Automatically scans for branches with these files and runs a job defined in them (what file to look for still has to be defined on the jenkins master, though).

### Tests written?
Look at me. I am the tests now.

### Commits signed with GPG?
Yes

@gtmanfred @dubb-b


gtmanfred (approved):

cachedout:
Hmm. I guess my first question is whether the salt repo is the right place for this. I know very little about Jenkins pipelines but one thought is whether it might make sense to break this into a separate repo and then clone from it before cloning from Salt. Is something like that possible?

KaiSforza:
Currently this is the implementation that we've tested. I haven't tried splitting these out into another repo, but we would lose a lot of the automation that having them in this repo gives us (easy clones, easy github pr updates, easily knows about changes to the repo, etc.).

gtmanfred:
The ci files are supposed to be in the github repository that they are testing.

This is how Jenkins adapted to the paradigm created by drone/travisci/circleci, etc.

Issue #48026, opened 2018-06-07
Windows service.disabled does not disable a service if it is set to manual


getSurreal:
### Description of Issue/Question
The win_service module interprets a manual service the same as a disabled service.

### Setup
windowsservices.sls
Disable Windows Service:
  service.disabled:
    - name: Browser

or

Disable Windows Service:
  service.dead:
    - name: Browser
    - enable: False

### Steps to Reproduce Issue
Set a windows service (i.e. Computer Browser) to manual and run a state against it to disable the service.

### Versions Report
Salt Version:
           Salt: 2017.7.2

Dependency Versions:
           cffi: 1.6.0
       cherrypy: unknown
       dateutil: 2.6.1
      docker-py: Not Installed
          gitdb: 2.0.3
      gitpython: 2.1.10
          ioflo: Not Installed
         Jinja2: 2.7.2
        libgit2: 0.26.3
        libnacl: 1.6.1
       M2Crypto: 0.21.1
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.5.1
   mysql-python: Not Installed
      pycparser: 2.14
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: 0.26.4
         Python: 2.7.5 (default, Feb 20 2018, 09:19:12)
   python-gnupg: Not Installed
         PyYAML: 3.12
          PyZMQ: 15.3.0
           RAET: Not Installed
          smmap: 2.0.3
        timelib: Not Installed
        Tornado: 4.2.1
            ZMQ: 4.1.4

System Versions:
           dist: redhat 7.5 Maipo
         locale: UTF-8
        machine: x86_64
        release: 3.10.0-862.3.2.el7.x86_64
         system: Linux
        version: Red Hat Enterprise Linux Server 7.5 Maipo

PR #48117, opened 2018-06-13
service.disabled on Windows

twangboy:
### What does this PR do?
Disable services that are set to manual on Windows using `service.disabled` and `service.dead` when `enable: False`

https://github.com/saltstack/salt/issues/48026

### Previous Behavior
`service.dead` and `service.disabled` would not disable a service if the start_type was set to Manual.

### New Behavior
Now manual start types are properly disabled.

### Tests written?
Yes

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/states/service.py
+++ b/salt/states/service.py
@@ -255,7 +255,12 @@ def _disable(name, started, result=True, **kwargs):
         return ret
 
     # Service can be disabled
-    before_toggle_disable_status = __salt__['service.disabled'](name)
+    if salt.utils.is_windows():
+        # service.disabled in Windows returns True for services that are set to
+        # Manual start, so we need to check specifically for Disabled
+        before_toggle_disable_status = __salt__['service.info'](name)['StartType'] in ['Disabled']
+    else:
+        before_toggle_disable_status = __salt__['service.disabled'](name)
     if before_toggle_disable_status:
         # Service is disabled
         if started is True:
@@ -549,7 +554,12 @@ def dead(name,
     # command, so it is just an indicator but can not be fully trusted
     before_toggle_status = __salt__['service.status'](name, sig)
     if 'service.enabled' in __salt__:
-        before_toggle_enable_status = __salt__['service.enabled'](name)
+        if salt.utils.is_windows():
+            # service.enabled in Windows returns True for services that are set
+            # to Auto start, but services set to Manual can also be disabled
+            before_toggle_enable_status = __salt__['service.info'](name)['StartType'] in ['Auto', 'Manual']
+        else:
+            before_toggle_enable_status = __salt__['service.enabled'](name)
     else:
         before_toggle_enable_status = True
 


damon-atkins:
Please add some comments to the if is_windows which currently do not have comments.

Not this PR fault, it would be nice if private functions where documents as well as public functions, it would help with any review.

salt/states/service.py
```diff
@@ -255,7 +255,12 @@ def _disable(name, started, result=True, **kwargs):
         return ret
 
     # Service can be disabled
-    before_toggle_disable_status = __salt__['service.disabled'](name)
+    if salt.utils.is_windows():
+        # service.disabled in Windows returns True for services that are set to
+        # Manual start, so we need to check specifically for Disabled
+        before_toggle_disable_status = __salt__['service.info'](name)['StartType'] in ['Disabled']
```
damon-atkins: Can we create a function called `_salt__['service.disabled'](name) ` rather than call `__salt__['service.info'](name)['StartType']  in ['Disabled']`.  Keep in interface consistent across platforms.

salt/states/service.py
```diff
@@ -549,7 +554,12 @@ def dead(name,
     # command, so it is just an indicator but can not be fully trusted
     before_toggle_status = __salt__['service.status'](name, sig)
     if 'service.enabled' in __salt__:
-        before_toggle_enable_status = __salt__['service.enabled'](name)
+        if salt.utils.is_windows():
+            # service.enabled in Windows returns True for services that are set
+            # to Auto start, but services set to Manual can also be disabled
+            before_toggle_enable_status = __salt__['service.info'](name)['StartType'] in ['Auto', 'Manual']
```
damon-atkins: Can we create a function called `_salt__['service.enabled'](name)` rather than call `__salt__['service.info'](name)['StartType']  in 'Auto', Manual'`.  Keep in interface consistent across platforms.

damon-atkins (changes requested):
Please add functions to the execution module instead.

salt/states/service.py
cachedout: Such a function already exists, doesn't it?

twangboy: There is already a `service.disabled`, however as noted in the code comments, services that are set to start manually are considered disabled.

salt/states/service.py
twangboy: Same issue as noted above.

damon-atkins:
I think it comes down to definitions:
enable/disable Unix/POSIX is on boot starting, windows equivalent of disable is Manual
start(running)/stop(dead) is independent of enable/disable  i.e. which Manual is also.

I am not convinced of this change
` __salt__['service.info'](name)['StartType'] in ['Disabled']`
As its currently correct in terms of the definition from Unix/POSIX system.

I think the real need is to rewrite service in more open terms and not "specific platform terms" e.g. "service.dead" is bad name.

We have:
service.dead # Ensure that the named service is dead by stopping the service if it is running
service.disable # Disable Start on Boot
service.enabled # Enable Start on Boot
service.masked  # fully disable a unit, making it impossible to start it even manually. **(Windows Disable)**
service.unmasked undo above

I think service state should be changed to the following i.e. create the following, and place warning on the old methods. (or have both)
 ````
nginx:
  service.status
    - start_mode: '<automatic|manual|etc>'   Can add more modes for different service managers.
    - state: '<running|stopped>' 
 ````

**My thoughts are leave it like it is, close this PR and open one to introduce new state called service.status**

PS Second thought.
```
nginx:
  service.configure
    - win_service:
      - username:
      - password:
      - command:
      - description:
      - etc
    - init.d:
    - systemd:
    - etc
```
The something like the above. Where the platform implementation can decide what to use e.g. systemd platform would use the systemd settings above using init.d settings. If systemd is replaced, then it will also handle the next bright idea which comes along.

lorengordon:
I prefer the current semantics, where the states are adjectives. Much more salty.

damon-atkins:
A word normally used as a noun can function as an adjective, depending on its placement. :-)
I have changed configure to configured, in the example above.

lorengordon:
Right, but configure would have been a verb, much more suited to the execution module side of things. ;)

I'd probably recommend "service.managed" to follow other salt states/conventions...

dwoz:
This PR addresses an open issue #48026 and fixes some broken integration tests.

https://jenkinsci.saltstack.com/job/2017.7/view/Python3/job/salt-windows-2016-py3/70/

Lets merge it and discuss changes to the service semantics in a separate issue (not a PR).

dwoz (approved):

dwoz:
re-run py

rallytime (approved):

PR #48148, opened 2018-06-15
property managment in dracr module + state

johnybx:
### New feature:
Adds managment for properties in idrac and state for this case.

```  test:
    dracr.property:  
       - properties:
           System.ServerOS.HostName: "Pretty-server"
           System.ServerOS.OSName: "Ubuntu 16.04"
       - admin_password: calvin
       - admin_root: myuser
       - host: 10.10.10.1
```


salt/modules/dracr.py
```diff
@@ -150,6 +150,43 @@ def __execute_ret(command, host=None,
...
+    if property is None:
+        raise SaltException('No property specified!')
+    elif value is None:
+        raise SaltException('No value specified!')
+    ret = __execute_ret('set %s %s' % (property, value), host=host,
+                        admin_username=admin_username,
+                        admin_password=admin_password)
+    return ret
+
+def property_present(host=None, admin_username=None, admin_password=None, property=None, value=None):
```
rares-pop: isn't this method confusing? where would you use it?

if the 'set' is expensive and you want to avoid setting it if it's the same value, the set_property could do a get and do a proper set only if it makes sense.

johnybx: The idea was to distinguish set and get method and if you want to use it as module
```
salt 'server' idrac.property_present' key value
```
you can which could be practical. In state only get and set methods are used because we need to know if value is set or should be set because of test. Important is that get and set methods are equally expensive so this should not matter.

If the name is confusing I can rename it but I did not come up with better name (suggestions are welcome).

michalsuba:
This feature is great, very useful for our infrastructure.

salt/modules/dracr.py
rares-pop: I think it's confusing because it's inside a module. It would make more sense in a state?

Typically modules are 'actions', and that method doesn't really qualify with an action, but more with a state.

If you have an analogy with the drac.py module and state, you can see that the 'present' python method is inside the state.

rares-pop: I find it confusing because in case I want to check if a property is set, I do a get_property. If I want to do a set, I do a set_property.  I assume tests can and do that too.

The only 'present'-like methods I saw were in salt/state/*, which basically you do with the property().

In case you really want it, maybe it makes sense to rename it to something like: set_property_if_doesnt_exist_or_has_different_value, but how is that different than just using the normal set_property, except is more expensive? (potentially doubling the runtime)

If it's just me with this opinion, feel free to ignore it. I'm not requiring changes or approving. I'm still fairly new with salt.

cachedout: I also find the naming of this confusing but the implementation itself seems useful and seems like it makes sense in an execution module.

rallytime:
Hi @johnybx - thank you very much for his addition. Can you make the adjustments suggested in the feedback above about the naming of the `property_present` execution module function?

The lint violations will also need to be fixed before we can merge this: https://jenkins.saltstack.com/job/PR/job/salt-pr-lint-n/22801/violations/

johnybx:
Hi @rallytime, @cachedout - I changed name of function to ensure_property_set which should define / explain function much better. Also I repaired linter problems.

salt/modules/dracr.py
```diff
@@ -151,6 +151,47 @@ def __execute_ret(command, host=None,
     return cmd
 
 
+def get_property(host=None, admin_username=None, admin_password=None, property=None):
+    '''
+    .. versionadded:: 2018.3.2
```
rallytime: This version should be `Fluorine` instead.

Can you also document the various options (host, admin_username, etc) and provide a CLI Example? I think there are examples of how this should look in other functions in this file.

salt/modules/dracr.py
```diff
@@ -151,6 +151,47 @@ def __execute_ret(command, host=None,
...
+        raise SaltException('No property specified!')
+    ret = __execute_ret('get {0}'.format(property), host=host,
+                        admin_username=admin_username,
+                        admin_password=admin_password)
+    return ret
+
+
+def set_property(host=None, admin_username=None, admin_password=None, property=None, value=None):
+    '''
+    .. versionadded:: 2018.3.2
```
rallytime: Same comment here as above about the version and documentation.

salt/modules/dracr.py
```diff
@@ -151,6 +151,47 @@ def __execute_ret(command, host=None,
...
+        raise SaltException('No value specified!')
+    ret = __execute_ret('set {0} {1}'.format(property, value), host=host,
+                        admin_username=admin_username,
+                        admin_password=admin_password)
+    return ret
+
+
+def ensure_property_set(host=None, admin_username=None, admin_password=None, property=None, value=None):
+    '''
+    .. versionadded:: 2018.3.2
```
rallytime: And here as well.

salt/states/dracr.py
```diff
@@ -0,0 +1,87 @@
+# -*- coding: utf-8 -*-
+'''
+.. versionadded:: 2018.3.2
```
rallytime: `Fluorine` instead. :)

salt/states/dracr.py
```diff
@@ -0,0 +1,87 @@
...
+    '''
+    Ensure the racadm command is installed
+    '''
+    if salt.utils.which('racadm'):
+        return True
+
+    return False
+
+
+def property(properties, admin_username='root', admin_password='calvin', host=None, **kwargs):
```
rallytime: What about `property_present` instead of just `property`? I think that is more inline with our state syntax.

salt/states/dracr.py
```diff
@@ -0,0 +1,87 @@
...
+       - properties:
+           System.ServerOS.HostName: "Pretty-server"
+           System.ServerOS.OSName: "Ubuntu 16.04"
+       - admin_password: calvin
+       - admin_root: myuser
+       - host: 10.10.10.1
+
+'''
+
+from __future__ import absolute_import
```
rallytime: Can you change this line to:
```
```
We need that for Python 3 compatibility.

rallytime (commented):
I have a couple more small requests. Can you fix those up?

In addition to those, I am wondering if it makes sense to create a new `dracr` _state_ when we already have a `drac` state. I see we have a `drac` and `dracr` exectution modules, but after a quick glance, I am not sure if it make sense to add a new `dracr` state, or to add the `property_present` function to the `drac` state. I could be missing some subtle distinctions, but I wanted to at least ask.

Anyone have thoughts on that?

johnybx:
I did those changes so docs and version should be fine now.

The choice for dracr module and state over drac comes from the fact that drac module and states does not support authentication and remote host which are quite useful features. We use this already so the idea is that others could find it useful feature too. Also once it will be part of dracr module we do not need it in drac module because dracr module seem to be more flexible (support for authentication and remote host).

rallytime (approved):

rallytime:
re-run all

gtmanfred (approved):

salt/states/dracr.py
```diff
@@ -0,0 +1,87 @@
...
+    '''
+
+    ret = {'name': host,
+           'context': {'Host': host},
+           'result': True,
+           'changes': {},
+           'comment': ''}
+
+    if host is None:
+        host = __salt__['cmd.run_all']('ipmitool lan print | tr -d " " | grep "IPAddress:" | cut -f 2 -d ":"',
```
gtmanfred: I would rather pull all this data back, and do the processing in python.

gtmanfred (changes requested):

salt/states/dracr.py
johnybx: Hi @gtmanfred, is there special reason for this? I used shell mainly because I do not really see point in processing the output in python as it is plain text prefixed with plenty of spaces and no standard format. So we would basicly need to iterate over text and find regex with re module per line. 
Should I change it anyway?

gtmanfred: We are writing python, I would rather use the builtin python stuff, in the really weird case that tr, cut, or grep is not in the path for some reason, we know for certain that the python regex module will be there.

johnybx: Ok, it makes sense.

johnybx: @gtmanfred check now please.

salt/states/dracr.py
```diff
@@ -0,0 +1,95 @@
...
+    '''
+
+    ret = {'name': host,
+           'context': {'Host': host},
+           'result': True,
+           'changes': {},
+           'comment': ''}
+
+    if host is None:
+        output = __salt__['cmd.run_all']('ipmitool lan print', python_shell=True)
```
gtmanfred: you should not need python_shell=True anymore.

gtmanfred (changes requested):
Just one small change, you shouldn't need python_shell set to True anymore now that you are not using pipes.

Otherwise, this looks ready to go.

Thanks for making that change!
Daniel

salt/modules/dracr.py
```diff
@@ -151,6 +151,107 @@ def __execute_ret(command, host=None,
     return cmd
 
 
+def get_property(host=None, admin_username=None, admin_password=None, property=None):
```
cachedout: When it comes to passing password directly to execution modules, IMHO that should always be pulled from pillars. This is going to show up in a process list while it's running and in shell history. :-/

rallytime: @johnybx Did you see the feedback above?

johnybx: @rallytime I was on vacation without computer.

@cachedout I followed other functions in that module. I could put there some condition if password is none check pillar data if admin_password exists there. Would this work for you ?

johnybx:
@gtmanfred I changed it.

gtmanfred (approved):

rallytime:
@johnybx Can you fix these final lint errors? https://jenkinsci.saltstack.com/job/pr-lint/job/PR-48148/2/warnings52Result/

salt/modules/dracr.py
cachedout: If it's already in other functions then I'm fine with updating them all at once in another PR.

cachedout (approved):

PR #48152, opened 2018-06-15
[WIP] Add initial version of Salt Capabilities RFC

meaksh:
### What does this PR do?
This PR tries to bring up a discussion about how to implement a mechanism to check Salt capabilities on the minion.

Please read the RFC [here](https://github.com/saltstack/salt/blob/2a01cf47bdf1c4e9c49c439968460b3f6f372b41/rfcs/0001-salt-capabilities.md)

What do you think about this?

### Commits signed with GPG?

Yes

### DIFF
--- a/rfcs/0001-salt-capabilities.md
+++ b/rfcs/0001-salt-capabilities.md
@@ -1,0 +1,62 @@
+- Feature Name: Salt Capabilities on Minions
+- Start Date: 2018-06-15
+- RFC PR: (leave this empty)
+- Salt Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+Bring a mechanism to figure out if certain Salt feature/capability is available on the minion.
+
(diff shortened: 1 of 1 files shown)


isbm (changes requested):
@meaksh I guess, this is outdated in a whole and the brand new RFC with prototype implementation is already coming. Can we probably close this?

gtmanfred (changes requested):
This needs to be more specific.

We already have some of this discovery, you are able to know what it can do based on which modules are loaded.

What does this add that knowing which modules are loaded doesn't give us?

isbm:
@gtmanfred this is an old RFC and this idea is totally scrapped out. I think this can be just closed.

meaksh:
@gtmanfred, sorry for the noise. This can be actually closed since @isbm is preparing the new RFC in a different PR.

PR #48166, opened 2018-06-17
Add trace logging and optimize file handling in grains.core.os_data

terminalmage:
This is a continuation of the troubleshooting being done for https://github.com/saltstack/salt-jenkins/issues/1000. I've narrowed the timeout to the `os_data` function in the core grains. To that end, this PR does the following:

1. Adds the timestamp to `log_fmt_console` in the integration suite's minion config file.
2. Adds trace logging throughout the `os_data` function.

In the process of adding the trace logging, I discovered a couple bugs in my recent changes to `mock_open`, so I resolved those. I also found some optimizations that could be made in `os_data`, such as removing `os.path.isfile()` checks before attempting to read from a file (when we can simply catch the IOError). Removing the `os.path.isfile()` checks exposed some shortcomings of `mock_open`, most importantly that the same data is returned for all filehandles opened while the patch is in effect. To resolve this, I rewrote `mock_open` to support distinct filehandles per call to `salt.utils.files.fopen()`.

Additionally, I've added a bunch of unit tests to confirm that mock_open behaves as closely as possible to a real filehandle. One of the problems uncovered by my recent changes to `mock_open` has exposed the fact that a number of tests which use `mock_open` were written to conform to the limitations of `mock_open`, and thus those tests were not as accurate as they should have been. Having unit tests to confirm the behavior of `mock_open` should help ensure that tests which rely on it are more accurate.


terminalmage:
Some tests need updating to work with the new `MockFH` class I wrote. Also, a couple lint issues. Working on them now.

terminalmage:
The failing unit tests from before should pass now. Might have a few lint failures though.

rallytime:
@terminalmage Looks like there are still some unittests failing with this change, in addition to the lint failure.

terminalmage:
Yes, those were all from newly-added tests that came in when I rebased. They should be fixed now as well.

I still need to add some more unit tests for the new mock_open, and write some documentation

terminalmage:
Correct.

rallytime:
@terminalmage There's one more unittest failure that is related to this:

https://jenkins.saltstack.com/job/PR/job/salt-pr-linode-cent7-py3/5836/

rallytime (approved):
This LGTM, except for that last new test failure.

gtmanfred (approved):

terminalmage:
re-run py3

PR #48174, opened 2018-06-18
Add docker.update_mine config option

terminalmage:
This allows for the mine refresh to be disabled via configuring this option in the minion config, grains, or pillar.


rallytime (approved):

rallytime:
Another PR will need to be made to develop when this merges forward to change the default, right?

terminalmage:
Correct

Issue #47006, opened 2018-04-10
marathon & fx2 grain modules cause master and minion failure


cedwards:
### Description of Issue/Question
A couple of proxy minion grains are failing to load properly on 2018.3.0 (FreeBSD) causing both the master and minion to fail.

The offending modules are `grains/marathon.py` and `grains/fx2.py`.

I can provide the `sysutils/py-salt-2018.3.0` port & Makefile if someone else needs to replicate.

### Setup
`pkg install py36-salt`
`service salt_master start`
`service salt_minion start`

### Steps to Reproduce Issue
After starting the master/minion I see the following in the logs and the service does not work:

```
==> /var/log/salt/master <==
2018-04-10 16:42:43,610 [salt.loader                                                                              :763 ][CRITICAL][48291] Failed to load grains defined in grain file opt-1.fx2 in function <function fx2 at 0x80dfa5510>, error:
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/site-packages/salt/loader.py", line 756, in grains
    ret = funcs[key]()
  File "/usr/local/lib/python3.6/site-packages/salt/grains/fx2.py", line 90, in fx2
    return _grains()
  File "/usr/local/lib/python3.6/site-packages/salt/grains/fx2.py", line 72, in _grains
    (username, password) = _find_credentials()
  File "/usr/local/lib/python3.6/site-packages/salt/grains/fx2.py", line 41, in _find_credentials
    usernames.append(__pillar__['proxy'].get('admin_username', 'root'))
  File "/usr/local/lib/python3.6/site-packages/salt/utils/context.py", line 228, in __getitem__
    return self._dict()[key]
KeyError: 'proxy'
2018-04-10 16:42:43,613 [salt.loader                                                                              :763 ][CRITICAL][48291] Failed to load grains defined in grain file opt-1.marathon in function <function marathon at 0x80dfaeb70>, error:
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/site-packages/salt/loader.py", line 756, in grains
    ret = funcs[key]()
  File "/usr/local/lib/python3.6/site-packages/salt/grains/marathon.py", line 41, in marathon
    "{0}/v2/info".format(__opts__['proxy'].get(
KeyError: 'proxy'

==> /var/log/salt/minion <==
[...]

PR #48177, opened 2018-06-18
Exclude opt-1/opt-2 .pyc files in loader.py

angeloudy:
### What does this PR do?
Exclude opt-1/opt-2 pyc files

#47006

### Previous Behavior


isbm:
@angeloudy conflicts, please resolve them.

gtmanfred (changes requested):
This will need the extra commits removed from the PR.

PR #48178, opened 2018-06-18
Teach salt.utils.minions.CkMinions.connected_ids() about IPv6

dwfreed:
### What does this PR do?

This teaches salt.utils.minions.CkMinions.connected_ids() about IPv6, so that IPv6 minions show up in the manage.list_state and family runners, as well as presence events, when the master runs on a system with /proc/net/tcp{,6}.

While I was in this section of code, I fixed some logic issues around including any localhost minion (including IPv6 localhost minion), while avoiding including any minion that isn't connected.  This allows any localhost minion to always show up in presence events.

I also fixed salt.utils.network.local_port_tcp() and salt.utils.network.active_tcp() so they don't include listening ports when /proc/net/tcp is available, to match their OS-specific implementations, and made it so IPv6 connections don't overwrite IPv4 connections in active_tcp() results.

### Previous Behavior

IPv6 minions don't show up in manage.list_state and family runner results or presence events on masters with /proc/net/tcp{,6}.  salt.utils.network.local_port_tcp() and salt.utils.network.active_tcp() include listening ports when /proc/net/tcp{,6} is available.  salt.utils.network.active_tcp() is missing some IPv4 connections when there are IPv6 connections.

### New Behavior

IPv6 minions are included in manage.list_state and family runner results and presence events on masters with /proc/net/tcp{,6}.  salt.utils.network.local_port_tcp() and salt.utils.network.active_tcp() don't include listening ports when /proc/net/tcp{,6} is available.  salt.utils.network.active_tcp() includes all IPv4 and IPv6 connections.

### Tests written?

No

### Commits signed with GPG?

Yes

### Miscellaneous notes

I opened this PR against develop because I changed the API of CkMinions.connected_ids() as the include_localhost parameter was no longer needed.  I would like to see this backported to at least 2018.3, though, and preferably make it in before 2018.3.2 is released, so I don't have to continue to apply this fix locally.  If changing the API of this function in a stable release is an issue, then the function could still accept the parameter but just not do anything with it.

I'm still not happy with how CkMinions.connected_ids() is implemented, because there's one presumably common case and a few edge cases that are not handled:

* NATed minions won't be listed (presumably common case)
* If there are multiple minions with the same IP address, they will all be listed if at least one is connected from that address
* If there are multiple masters listening on the same port but different addresses, minions can be listed that aren't connected to the master being interrogated
* (possibly Linux-specific) If a localhost minion connects to the master using an address in 127.0.0.0/8 that either is not explicitly configured on the loopback interface or the master is not explicitly listening on, it won't be listed

Fixing any of these would require a lot more work.

### DIFF
--- a/salt/runners/manage.py
+++ b/salt/runners/manage.py
@@ -201,9 +201,24 @@ def up(tgt='*', tgt_type='glob', timeout=None, gather_job_timeout=None):  # pyli
     return ret
 
 
-def list_state(subset=None, show_ipv4=False, state=None):
+def _show_ip_migration(show_ip, show_ipv4):
+    if show_ipv4 is not None:
+        salt.utils.versions.warn_until(
+            'Sodium',
+            'The \'show_ipv4\' argument has been renamed to \'show_ip\' as'
+            'it now also includes IPv6 addresses for IPv6-connected'
+            'minions.'
+        )
+        return show_ipv4
+    return show_ip
+
+
+def list_state(subset=None, show_ip=False, show_ipv4=None, state=None):
     '''
     .. versionadded:: 2015.8.0
+    .. versionchanged:: Fluorine
+        The 'show_ipv4' argument has been renamed to 'show_ip' as it now
+        includes IPv6 addresses for IPv6-connected minions.
 
     Print a list of all minions that are up according to Salt's presence
     detection (no commands will be sent to minions)
@@ -211,7 +226,7 @@ def list_state(subset=None, show_ipv4=False, state=None):
     subset : None
         Pass in a CIDR range to filter minions by IP address.
 
-    show_ipv4 : False
+    show_ip : False
         Also show the IP address each minion is connecting from.
 
     state : 'available'
@@ -224,6 +239,7 @@ def list_state(subset=None, show_ipv4=False, state=None):
 
         salt-run manage.list_state
     '''
+    show_ip = _show_ip_migration(show_ip, show_ipv4)
     conf_file = __opts__['conf_file']
     opts = salt.config.client_config(conf_file)
     if opts['transport'] == 'raet':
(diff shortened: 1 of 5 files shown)


salt/utils/minions.py
```diff
@@ -643,16 +647,19 @@ def connected_ids(self, subset=None, show_ipv4=False, include_localhost=False):
...
-                        continue
                     if ipv4 in addrs:
                         if show_ipv4:
                             minions.add((id_, ipv4))
                         else:
                             minions.add(id_)
                         break
+                for ipv6 in grains.get('ipv6', []):
+                    if ipv6 in addrs:
+                        if show_ipv4:
```
isbm: This is request to show IPv4, and still you putting IPv6. I guess, this parameter needs either be renamed to "show_ip" (preferred) or we need to add another one (no idea why). In fact, this parameter/option is only used once, in manage runner, list_state function.

isbm (changes requested):

gtmanfred (changes requested):
Can the localhost changes be split out to a different PR so that we can make sure this gets reviewed correctly?

I just want to make sure something isn't missed while localhost is being changed.

Thanks,
Daniel

gtmanfred:
Ok, then can you split it into two commits, the first one fixing the localhost stuff for ipv4, and then the second one adding ipv6 support so that we can see the difference?

isbm (approved):

gtmanfred:
Can you revert the changes to `doc/man/salt.7`? these are automatically generated when packaged, they do not need to be modified.

salt/runners/manage.py
```diff
@@ -201,7 +201,7 @@ def up(tgt='*', tgt_type='glob', timeout=None, gather_job_timeout=None):  # pyli
     return ret
 
 
-def list_state(subset=None, show_ipv4=False, state=None):
+def list_state(subset=None, show_ip=False, state=None):
```
gtmanfred: This change will need to go through a deprecation process.

gtmanfred: Where you have show_ipv4, and show_ip for 2 releases, and then it can be removed.

gtmanfred (changes requested):

gtmanfred (approved):

rallytime (approved):

PR #48181, opened 2018-06-18
timezone: Fix {set/get}_zone for legacy NILinuxRT

rares-pop:
On legacy version of NILinuxRT, /etc/localtime is a
symlink to /etc/natinst/share/localtime.
On this systems, {set/get}_zone must work directly on
/ect/natinst/share/localtime, instead of default file.

Signed-off-by: Heghedus Razvan <razvan.heghedus@ni.com>

### What does this PR do?
National Instruments supports two Linux RT distributions, and the former one is using /etc/natinst/share/localtime instead of /etc/localtime for the locale.

### Previous Behavior
timezone doesn't work correctly on NILinuxRT legacy platform

### Tests written?

No

### Commits signed with GPG?

No


rallytime:
Ah, @rares-pop - The timezone (`unit.modules.test_timezone`) tests didn't like this change very much. Can you take a look?

https://github.com/saltstack/salt/pull/48181

rares-pop:
I fixed the tests! Sorry for overlooking that.

garethgreenaway (approved):

Issue #44366, opened 2017-10-31
salt-cloud fails to prepend root_dir to log_file


brianthelion:
### Description of Issue/Question

With non-standard `root_dir`, the `salt-cloud` CLI fails and throws and non-zero exit code if the `log_file` setting isn't specifically set in the `cloud` config. It's not immediately clear from the code where the problem lies.

### Setup

```bash
MINION_NAME='test_minion'
PROFILE_NAME='test_profile'

LOG_LVL=all

ROOT_DIR=$(mktemp -d)
CONFIG_DIR=$ROOT_DIR/etc/salt
mkdir -p $CONFIG_DIR
FILE_ROOT_BASE=$(mktemp -d)

cat <<EOF > $CONFIG_DIR/cloud
root_dir: $ROOT_DIR
#log_file: $ROOT_DIR/var/log/salt/cloud
EOF

salt-cloud --log-file-level=$LOG_LVL --config-dir=$CONFIG_DIR -p $PROFILE_NAME $MINION_NAME
echo "EXIT $?"
```

gives 

```
No permissions to access "/var/log/salt/cloud", are you running as the correct user?
EXIT 1
```

### Steps to Reproduce Issue
Run the script with and without the comment on the `log_file: ...` line.

### Versions Report
```
Salt Version:
           Salt: 2017.7.2
 
Dependency Versions:
           cffi: Not Installed
       cherrypy: Not Installed
       dateutil: Not Installed
      docker-py: Not Installed
          gitdb: 2.0.3
      gitpython: 2.1.7
          ioflo: Not Installed
         Jinja2: 2.9.6
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.4.8
   mysql-python: Not Installed
      pycparser: Not Installed
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: Not Installed
         Python: 3.5.3 (default, Sep 14 2017, 22:58:41)
   python-gnupg: Not Installed
         PyYAML: 3.12
          PyZMQ: 16.0.3
           RAET: Not Installed
          smmap: 2.0.3
        timelib: Not Installed
        Tornado: 4.5.2
            ZMQ: 4.1.6
 
System Versions:
           dist: Ubuntu 17.04 zesty
         locale: UTF-8
        machine: x86_64
        release: 4.12.0-041200rc6-generic
         system: Linux
[...]

PR #48182, opened 2018-06-18
PR 44366

ghost:
### What does this PR do?

Allows Salt Cloud to operate within a separate directory tree without having to define the configuration directive `log_file` separately (on the command line or in the main cloud configuration file).

 #44366

### Previous Behavior

A user has to specify `--log-file` separately, even if `root_dir` is specified in the main cloud configuration file and `-c` is being used on the command line.

### New Behavior

Running SaltCloud with non-default configuration file locations works as expected ie. the log file ends up under the directory hierarchy specified in 'root_dir' directive in the main salt cloud configuration file.

### Tests written?

No

### Commits signed with GPG?

No

### DIFF
--- a/salt/config/__init__.py
+++ b/salt/config/__init__.py
@@ -2787,7 +2787,7 @@ def cloud_config(path, env_var='SALT_CLOUD_CONFIG', defaults=None,
     apply_sdb(opts)
 
     # prepend root_dir
-    prepend_root_dirs = ['cachedir']
+    prepend_root_dirs = ['cachedir','log_file']
     prepend_root_dir(opts, prepend_root_dirs)
 
     # Return the final options
--- a/salt/config/__init__.py
+++ b/salt/config/__init__.py
@@ -2787,7 +2787,9 @@ def cloud_config(path, env_var='SALT_CLOUD_CONFIG', defaults=None,
     apply_sdb(opts)
 
     # prepend root_dir
-    prepend_root_dirs = ['cachedir','log_file']
+    prepend_root_dirs = ['cachedir']
+    if 'log_file' in opts and urlparse(opts['log_file']).scheme == '':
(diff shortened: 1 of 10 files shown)


gtmanfred (approved):

salt/config/__init__.py
```diff
@@ -2787,7 +2787,7 @@ def cloud_config(path, env_var='SALT_CLOUD_CONFIG', defaults=None,
     apply_sdb(opts)
 
     # prepend root_dir
-    prepend_root_dirs = ['cachedir']
+    prepend_root_dirs = ['cachedir','log_file']
     prepend_root_dir(opts, prepend_root_dirs)
```
isbm: @samilaine but can we maybe not to call this list "prepend_root_dirs", since it is path options? I understand that the variable was already there, but let's agree it is greatly misleading name. Maybe I would just do it so:

```python
prepend_root_dir(opts, ['cachedir','log_file'])
```

So whoever wants to know how this is working, can lookup the function and see that this is just a `path_options`.

isbm (approved):

salt/config/__init__.py
samilaine: However, it turns out that there might be a reasonable explanation for that funny variable `prepend_root_dirs` (see line [2358](https://github.com/saltstack/salt/blob/develop/salt/config/__init__.py#L2538) in `salt/config/__init__.py`) .

So, instead of getting rid of the funny variable, I opted for abusing it even more.

gtmanfred: for the record, i am for leaving the variable, as it is how all the other config functions work, master_config, minion_config, etc.

isbm: @samilaine :wink: Well, OK. I approved this anyway, and that one was up to you. Thanks!

PR #48191, opened 2018-06-18
Use the --disable-pip-version-check option

twangboy:
### What does this PR do?
Adds the `--disable-pip-version-check` option to all pip installs.

When installing the build environment for Python 3 the warning that we're using an older version of pip was causing the script to hang. This problematic when trying to automate builds.

### Tests written?
NA

### Commits signed with GPG?
Yes


KaiSforza (approved):
:+1:

rallytime (approved):

PR #48194, opened 2018-06-18
Add yaml_strict renderer

jacksontj:
The C versions of the loader/dumper are 5-10x faster than the python
versions. We already do this for the "unsafe" ones, this diff simply
adds it for the "safe" ones.

Supercedes #47422

TLDR; while attempting to make the current yaml loader use the CLoader it gets complicated very fast since we support some non-yaml things (which is scary!). So, to avoid that problem I've created a new renderer (cyaml) that just uses the csafeloader and has the "yaml" niceties we have. This way you can get the perf benefits on a per-sls file basis (through the shebang) or install-wide by setting the [default renderer](https://docs.saltstack.com/en/latest/ref/renderers/#multiple-renderers)

cc: @terminalmage

### DIFF
--- a/salt/renderers/cyaml.py
+++ b/salt/renderers/cyaml.py
@@ -1,0 +1,142 @@
+# -*- coding: utf-8 -*-
+'''
+C-YAML Renderer for Salt
+
+For YAML usage information see :ref:`Understanding YAML <yaml>`.
+'''
+
+from __future__ import absolute_import, print_function, unicode_literals
+
+# Import python libs
+import re
+import logging
+import warnings
+import yaml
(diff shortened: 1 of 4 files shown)


jacksontj:
After thinking about it some, I'm going to rename this to `yaml_strict`. The main goal here is to create an actual yaml parser (to remove the special casing etc.) -- and that itself enables the Cloader. Ideally we'd call this one `yaml` and move the other one to `salt_yaml` or something, but that will be for another time.

jacksontj:
After talking with @terminalmage more it looks like we can probably go back to the approach in #47422 as the one feature we are trying to preserve is actually a bug :)

Issue #46924, opened 2018-04-06
Stale custom roster from __pycache__ executed instead of changed .py file


chschmitt:
### Description of Issue

I implemented a custom roster module. After changing the module, `salt-ssh` executes the stale `.pyc` file from the `__pycache__` directory, instead of the the changed `.py` file.

### Setup

Dummy roster implementation `stale.py`:

```
from __future__ import print_function

def targets(tgt, tgt_type='glob', **kwargs):
    print(__file__)
    return {}
```
### Steps to Reproduce Issue

1. Create custom roster `/tmp/ext/roster/stale.py` and add `/tmp/ext` to `module_dirs` in master configuration.

2. Execute `salt-ssh '*' test.ping --roster=stale`, output:
   ```
   /tmp/ext/roster/stale.py
   [ERROR   ] No matching targets found in roster.
   ```

3. Change roster module, e.g. insert `print('file was changed')`
4. Execute `salt-ssh '*' test.ping --roster=stale`, **change in module is not refelected**. Output:
   ```
   /tmp/ext/roster/__pycache__/stale.cpython-35.pyc
   [ERROR   ] No matching targets found in roster.
   ```

5. Remove `__pycache__`: `rm -rfv /tmp/ext/roster/__pycache__`
4. Execute `salt-ssh '*' test.ping --roster=stale`, output:
   ````
   file was changed
   /tmp/ext/roster/stale.py
   [ERROR   ] No matching targets found in roster.
   ````

### Workaround ###

Delete the `__pycache__` directory and create a `__pycache__` file (!) in the roster directory instead. Generally speaking, create a condition that prevents the executing Python process from creating the `__pycache__` directory permanently.

### Versions Report

```
Salt Version:
           Salt: 2018.3.0
 
Dependency Versions:
           cffi: Not Installed
       cherrypy: Not Installed
       dateutil: Not Installed
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.10
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.5.6
   mysql-python: Not Installed
      pycparser: Not Installed
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: Not Installed
         Python: 3.5.2 (default, Nov 23 2017, 16:37:01)
[...]

PR #45095, opened 2017-12-20
PY3: Make loader ignore .pyc files not in __pycache__

terminalmage:
This keeps the loader from loading a PY2 .pyc file if it is present.

PR #48195, opened 2018-06-19
Exclude opt-1/opt-2 .pyc files in loader.py

angeloudy:
### What does this PR do?
Exclude *opt-1.pyc/*opt-2.pyc bytecode files generated by optimization option during compiling.

#47006


gtmanfred (commented):
If they aren't usable, what is the point of even compiling them?  Why not just remove -O1,2 from the build script for the package?

@cedwards can you comment on this since I have only ever seen this with freebsd packages?

Thanks,
Daniel

angeloudy:
@gtmanfred  I don't think it's a freebsd specific issue. The opt-1 pyc files were generated with optimization flags. It could be that on other platforms the default optimization level is 0 and the issue is hidden.

ari:
1. They are usable, its just that salt has its own class loader which isn't properly handling python3 compiled files with optimisation. Salt already strips out cpython-36 from the file strings in its special loader.
2. This isn't specific to FreeBSD. Python3 creates opt-1 and opt-2 files when optimisation is enabled on all platforms I think.
3. Optimisation is enabled on FreeBSD by default, so no FreeBSD user has been able to run salt since it was upgraded to 2018.3 unless they tweaked the python port makefile. @cedwards noted this, but the FreeBSD port hasn't yet been patched to work.
4. This is going to hit other platforms too as soon as someone decides to turn on the optimisation flag, so its best fixed in salt's loader. I don't understand why python decided to move from .pyo python2 files to -opt-1 naming in python3, but that's what they chose. I reckon an upstream (salt) fix is better than a FreeBSD specific hack.

https://www.python.org/dev/peps/pep-0488/

Perhaps a better patch here would be to pull in the highest opt level files it can find as the first priority. My undertanding is that -O1 removes all asserts in the bytecode, but perhaps it also does other things. -O2 removes docstrings.

isbm (changes requested):
I am not sure I would agree to this change. What it does is following:

- Eliminates the possibility to use optimised code (`-O` and `-OO` is now just gone). This is _against_ the PEP you're referring to.
- It leaves the garbage in `__pycache__` (well, the `.opt*pyc` are now just a garbage)
- It doesn't prevent these objects created (and so expected to be in use).
- It is a workaround, not a fix. And not even needed workaround for everyone.

Eventually, this is a packaging/distribution problem. If your package manager cannot cope with this, well, fix your package manager. Or switch the OS, e.g. SLES has no such problems (and we would like to have optimised code!!). :stuck_out_tongue: 

At your best, you should say "If this is FreeBSD, then workaround" but don't force _everyone_ live without optimised objects!

@terminalmage @cachedout what do you think?

ari:
I agree that the patch should be adjusted to preference optimised code. But the questions are:

1. Should salt die if more than one compiled optimisation version is found? That's what it does today.

2. Do you want to raise a bug upstream with python that their compile tool should never put two optimisation versions into the cache directory?

3. Even if python or the distribution put only the optimised files into the cache folder, how will you handle the opt-[12] in the filename?

isbm:
>  if more than one compiled optimisation version is found

I am wondering why this happens. Do you have an idea and suggestions?

> Do you want to raise a bug upstream with python that their compile tool should never put two optimisation versions into the cache directory?

Again, to me it more seems like a messy environment. You either use it with `-O` or `-OO` or not at all and just get `pyc`. But if you end up with all three together, it means you have an inconsistent invocation of your Salt installation that produces both. So I would ask myself why this is happening at first place.

And second, I would rather do a more smart way of loading that would _prefer_ files (taken to the account that we can have such messy environment). The [importlib](https://docs.python.org/3/library/importlib.html) actually has this [mechanism in place](https://www.python.org/dev/peps/pep-0488/#importlib).

> Even if python or the distribution put only the optimised files into the cache folder, how will you handle the opt-[12] in the filename?

I think we should "dance" [this direction](https://docs.python.org/3/library/importlib.html#importlib.util.cache_from_source) instead. I agree this is an issue, but I disagree for quick-n-wrong fixes.

gtmanfred:
I am interested in what @terminalmage thinks of this, since I believe he was involved with renaming the pyc files in __pycache__.

Thanks,
Daniel

terminalmage (commented):
Well, first of all, to my knowledge no files were renamed. The `.replace()` simply removes the `.cpython-3X` from the `.pyc` filename _solely_ for the purposes of telling the loader which module name to use for the module.

That said, I think that we need to have a greater discussion about how we want the loader to behave, because there are problems with the current implementation that affect both Python 2 _and_ Python 3.

I'd like both @cachedout and @thatch45's opinions on this. Note that @thatch45 is on vacation right now so we'll need to wait until he returns to get his input.

-----

So, the way we are currently handling things is we build a list of files in the module_dirs, and then we iterate over them. We build a mapping of the candidate module name to a tuple containing the full path to the file, as well as just the extension. Now, because we are iterating over directories containing both byte-compiled files and plain `.py` files, we are naturally going to iterate over both `modname.py` and `modname.pyc` (`modname.cpython-3X.pyc` on Python 3). The way the loader decides which of the two to prefer (and which to ignore) is based solely on the order in which they appear in `imp.get_suffixes()` (i.e. the `suffix_order` list in loader.py):

```python
>>> [x[0] for x in imp.get_suffixes()]
['.so', 'module.so', '.py', '.pyc']
```

The thing about this approach is that it means that ***we will never load a `.pyc` file if the corresponding `.py` file is also present*** (which duh, it should always be). I'm not sure if this was the intention, as I did not write the loader. ***EDIT: This is actually not true for Python 3, as you can see in [my comment below](https://github.com/saltstack/salt/pull/48195#issuecomment-398898394).***

Incidentally, when using `importlib` in Python 3, we create the same data structure on which `suffix_order` is based, but it seems to have been done specifically to match the existing Python 2.X/3.4 behavior (which was to derive the order from `imp.get_suffixes()` as noted above). Thus, Python 3 exhibits the same behavior of only ever loading a `.pyc` file if no corresponding `.py` file is present. ***EDIT: again, not actually true for Python 3, see [my comment below](https://github.com/saltstack/salt/pull/48195#issuecomment-398898394).***

That is what brings us to this issue. [This line](https://github.com/saltstack/salt/blob/7192c38/salt/loader.py#L1258) exists to ensure we consider `test.cpython-36.pyc` as the module `test`, but it of course does not take into account the PEP 488 naming conventions, which is why you get a module named `opt-1` or `opt-2` in the loader. So, rather than simply ignoring these files, if we were to maintain our current approach then what we should actually be doing is replacing that line with:

```python
PY3_PRE_EXT = re.compile(r'\.cpython-{0}{1}(?:\.opt-[12])?'.format(*sys.version_info[:2]))

if six.PY3:
    f_noext = PY3_PRE_EXT.sub('', f_noext)
```

However, once we make that change, we'd still be left with the fact that this would for all intents and purposes make the byte-compiled file "disappear" from the perspective of the loader. We'd still end up loading the `.py` file instead of the `.pyc`.

So, we need to make a decision here: do we want to prefer `.pyc` files if present? If we don't, then I don't know why we're bothering with them at all, we should just skip them. However, if we do want to load them, then it gets complicated:

- When someone updates a custom module in `salt://_modules` and syncs it over, we'd only be syncing the file(s) that changed. If the `.py` file changed, that's all we'd sync and the stale `.pyc` would remain in the minion's extmods cache. This means that the minion would never get the updated .pyc file unless the stale one was removed. This is something we would have to consider and perhaps modify our module syncing code to take care of. ***EDIT: Actually, the `.pyc` likely wouldn't exist in salt://_modules/, but the reasoning is still the same; the `.pyc` would need to be cleared in cases where the `.py` is updated***.

- For cases in which optimized `.pyc` files are present, which do we prefer? Should this be tunable via `__opts__`?

In my opinion, if we are going to prefer `.pyc` files when present, we should make sure that the `files` list over which we are iterating does not contain any `.pyc` files. Instead, when `ext == '.py'` we should look for a corresponding `.pyc` file. This should be done via another method in the `LazyLoader` class. Here's my (completely untested) idea for how that would be done, taking a cue from @isbm's idea to use importlib for this where possible:

```python

class LazyLoader(salt.utils.lazy.LazyDict):
    ...
    (snip)
    ...

    py3_pyc_suffix = '.cpython-{0}{1}.pyc'.format(sys.version_info[:2])

    def _find_pyc(self, fpath):
        if USE_IMPORTLIB:
            # Assuming that __opts__['optimization_order'] is a list of
            # integers, with "0" denoting no optimization.
            for optimization_level in __opts__['optimization_order']:
                pyc = importlib.util.cache_from_source(
                    fpath,
                    optimization=optimization_level or None)
                try:
                    os.stat(pyc)
                    return pyc
                except OSError:
                    # File does not exist
                    continue
        else:
            if six.PY3:
                # Python 3.4 (ignore optimized .pyc files? ¯\_(ツ)_/¯)
                dirname, fname = os.path.split(fpath)
                pyc = os.path.join(
                    dirname,
                    '__pycache__',
                    fname[:-3] + self.py3_pyc_suffix
                )
            else:
                pyc = fpath + 'c'
            try:
                os.stat(pyc)
                return pyc
            except OSError:
                pass

        return None
```

terminalmage:
I took another look and some of the information in my earlier comment is inaccurate. On Python 3, `.pyc` comes before `.py` in the suffix order. I must have been running PY2 when I thought I was using PY3 and checking for the suffix order while stepping through in a debugger. Here is the actual suffix_order for Python 3:

```python
['', '.cpython-36m-x86_64-linux-gnu.so', '.abi3.so', '.so', '.pyc', '.py']
```

So this actually means that if a `.py` file is updated, Salt will not load the updated `.py` file, which would likely explain https://github.com/saltstack/salt/issues/46924.

ari:
I think it also explains #48221 which is that you can no longer reference code in a subfolder of runner. I'm not sure what our workaround is there other than flattening out all our subfolders into one folder.

isbm:
There are two hard problems in computer science: cache invalidation, naming things, and off-by-1 errors. :confused: 

I've seen the same behaviour of `__pycache__` before where custom runners misbehave as well (no changes happens, unless you explicitly kill the entire cache directory).

The order is likely correct here, because there is no point to load `.py` file _before_ you have already compiled one. However, I am wondering if the `.pyc`s are updated alongside the source? If you do:

```python
import importlib
foo = importlib.import_module('blah')
foo.run()
```

Then the `__pycache__/blah.cpython36.py` will be updated. So it is crucial that the Python updates  the `pyc` files every time a module is triggered to be imported.

angeloudy:
@terminalmage 
I think the problem was introduced by #45095.
Before this fix, it seems the `.pyc` files in `__pycache__` would  not be loaded at all.

terminalmage:
@isbm using either `imp` or `importlib` to load the source file will update the `.pyc`, I even looked at the source to be sure :smile:

thatch45 (approved):

thatch45:
I think this approach works, I think that we should still load pyc files

isbm:
@terminalmage to me sounds like @thatch45 meant to load `pyc` files when source is not around.

terminalmage:
https://github.com/saltstack/salt/pull/48636 has been opened to provide a better solution for this. Closing this PR as https://github.com/saltstack/salt/pull/48636 supersedes it.

Thanks to @angeloudy for bringing this to our attention.

Issue #48184, opened 2018-06-18
Invalid DMI prevents salt-minion from running


mmulqueen:
### Setup
(Please provide relevant configs and/or SLS files (Be sure to remove sensitive info).)

Salt is setup with the default settings, except the master and master_finger settings.

### Steps to Reproduce Issue
(Include debug logs if possible and relevant.)

I've only been able to see this bug on one machine. I've deployed an identical setup on other machines without issue. 

When I try to start the salt-minion service or start salt-minion direct, it fails to start and give this traceback:

```
Process Process-1:
Traceback (most recent call last):
  File "/usr/lib/python2.7/multiprocessing/process.py", line 267, in _bootstrap
    self.run()
  File "/usr/lib/python2.7/multiprocessing/process.py", line 114, in run
    self._target(*self._args, **self._kwargs)
  File "/usr/lib/python2.7/dist-packages/salt/scripts.py", line 144, in minion_process
    minion.start()
  File "/usr/lib/python2.7/dist-packages/salt/cli/daemons.py", line 348, in start
    self.minion.tune_in()
  File "/usr/lib/python2.7/dist-packages/salt/minion.py", line 1029, in tune_in
    self._spawn_minions()
  File "/usr/lib/python2.7/dist-packages/salt/minion.py", line 978, in _spawn_minions
    jid_queue=self.jid_queue,
  File "/usr/lib/python2.7/dist-packages/salt/minion.py", line 959, in _create_minion_object
    jid_queue=jid_queue)
  File "/usr/lib/python2.7/dist-packages/salt/minion.py", line 1099, in __init__
    self.opts['grains'] = salt.loader.grains(opts)
  File "/usr/lib/python2.7/dist-packages/salt/loader.py", line 734, in grains
    ret = funcs[key]()
  File "/usr/lib/python2.7/dist-packages/salt/grains/core.py", line 1788, in os_data
    grains.update(_virtual(grains))
  File "/usr/lib/python2.7/dist-packages/salt/grains/core.py", line 872, in _virtual
    if 'VirtualBox' in output:
UnicodeDecodeError: 'ascii' codec can't decode byte 0xff in position 0: ordinal not in range(128)
```

I've traced this back and I see that the contents of /sys/devices/virtual/dmi/id/product_name is actually
[...]

PR #48216, opened 2018-06-19
Fix UnicodeDecodeError when reading file to determine virtual grain

Ch3LL:
### What does this PR do?
Fixes the UnicodeDecodeError when trying to read `/sys/devices/virtual/dmi/id/product_name` file when it contains non ascii characters.


rallytime:
re-run py3

rallytime (approved):

PR #48219, opened 2018-06-19
Fix: LXC legacy configuration key warnings falsely report errors during state change

zer0def:
### What does this PR do?

Fixes another case where a condition based on LXC command's output, instead of retcode could cause false error reporting and pre-emptive bailout on a task.

### Tests written?
No

### Commits signed with GPG?

No


zer0def:
Closing until I narrow down source of indeterministic behavior.

zer0def:
Forementioned indeterministic behavior was caused by other means, so I've reopened this for merging.

terminalmage (approved):

PR #48220, opened 2018-06-20
Explain how to use a grain in a state

mnieber:
This PR explains that that you can access a grain in the state file using `{{ grains['key'] }}`


rallytime (approved):
Thanks @mnieber!

Issue #46507, opened 2018-03-13
Failed to return clean data


nkv16786:
### Description of Issue/Question

State file is not being applied when thin_dir is set to /etc/salt.The error is as follows:-

Devpacktest:
    ----------
    _error:
        Failed to return clean data
    retcode:
        0
    stderr:
        Connection to xx.xx.xxx.xx closed.
    stdout:
        ERROR: Failure deploying ext_mods: + SUDO=
        + '[' -n sudo ']'
        + SUDO='sudo '
        + SUDO_USER=
        + '[' 'sudo ' ']'
        + '[' '' ']'
        + '[' 'sudo ' ']'
        + '[' -n '' ']'
        + EX_PYTHON_INVALID=10
        + PYTHON_CMDS='python3 python27 python2.7 python26 python2.6 python2 python'
        + for py_cmd in '$PYTHON_CMDS'
        + command -v python3
        + continue
        + for py_cmd in '$PYTHON_CMDS'
        + command -v python27
        + continue
        + for py_cmd in '$PYTHON_CMDS'
        + command -v python2.7
        + python2.7 -c 'import sys; sys.exit(not (sys.version_info >= (2, 6)
        and sys.version_info[0] == 2));'
        ++ python2.7 -c 'from __future__ import print_function;
        import sys; print(sys.executable);'
        + py_cmd_path=/usr/bin/python2.7
        ++ command -v python2.7
        + cmdpath=/usr/bin/python2.7
        + file /usr/bin/python2.7
        + grep 'shell script'
        + exec sudo /usr/bin/python2.7 -c 'import base64;
        exec(base64.b64decode("""IyBweWxpbnQ6IGRpc2FibGU9bm8tZW5jb2RpbmctaW4tZmlsZSxyZXNvdXJjZS1sZWFrYWdlCicn
        JwpUaGlzIGlzIGEgc2hpbSB0aGF0IGhhbmRsZXMgY2hlY2tpbmcgYW5kIHVwZGF0aW5nIHNhbHQg
        dGhpbiBhbmQKdGhlbiBpbnZva2luZyB0aGluLgoKVGhpcyBpcyBub3QgaW50ZW5kZWQgdG8gYmUg
        aW5zdGFudGlhdGVkIGFzIGEgbW9kdWxlLCByYXRoZXIgaXQgaXMgYQpoZWxwZXIgc2NyaXB0IHVz
[...]

PR #48234, opened 2018-06-20
Fix py2 thin dir issues

dwoz:
### What does this PR do?

Fixes a missing thin directory dependency when running under python2 and adds regression test for fix.

#46507

### Previous Behavior

Running salt-call in thin directory will fail due to missing concurrent.futures import which can break salt-ssh under Python 2.

### New Behavior

salt-ssh runs under both Python 2 and Python 3

### Tests written?

Yes

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/client/ssh/__init__.py
+++ b/salt/client/ssh/__init__.py
@@ -1106,7 +1106,10 @@ ARGS = {10}\n'''.format(self.minion_config,
             shim_tmp_file.write(salt.utils.to_bytes(cmd_str))
 
         # Copy shim to target system, under $HOME/.<randomized name>
-        target_shim_file = '.{0}.{1}'.format(binascii.hexlify(os.urandom(6)), extension)
+        target_shim_file = '.{0}.{1}'.format(
+            binascii.hexlify(os.urandom(6)).decode('ascii'),
+            extension
+        )
         if self.winrm:
             target_shim_file = saltwinshell.get_target_shim_file(self, target_shim_file)
         self.shell.send(shim_tmp_file.name, target_shim_file, makedirs=True)
--- a/salt/utils/thin.py
+++ b/salt/utils/thin.py
@@ -14,6 +14,7 @@ import tarfile
 import zipfile
 import tempfile
 import subprocess
+import concurrent
 
 # Import third party libs
 import jinja2
@@ -106,6 +107,8 @@ def get_tops(extra_mods='', so_mods=''):
             os.path.dirname(msgpack.__file__),
             ]
 
+    if _six.PY2:
+        tops.append(os.path.dirname(concurrent.__file__))
     tops.append(_six.__file__.replace('.pyc', '.py'))
     tops.append(backports_abc.__file__.replace('.pyc', '.py'))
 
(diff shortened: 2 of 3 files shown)


salt/client/ssh/__init__.py
```diff
@@ -1106,7 +1106,10 @@ def shim_cmd(self, cmd_str, extension='py'):
             shim_tmp_file.write(salt.utils.to_bytes(cmd_str))
 
         # Copy shim to target system, under $HOME/.<randomized name>
-        target_shim_file = '.{0}.{1}'.format(binascii.hexlify(os.urandom(6)), extension)
+        target_shim_file = '.{0}.{1}'.format(
+            binascii.hexlify(os.urandom(6)).decode('ascii'),
```
s0undt3ch: You likely only want to decode under py3 right?

dwoz: Good catch. Decoding on python 2 doesn't cause an issue because the unicode string is decoded by format.

```
>>> '{}'.format(binascii.hexlify(os.urandom(6)).decode('ascii'))
'1f84d3dc6284'
```

isbm:
@dwoz waitasec, how exactly you've concluded that the problem is lack of `concurrent.futures` import? And I fail to understand why this is affecting _only_ Python 2 if the same module is not anywhere added also to Python 3? Unless I am missing something, the `concurrent.futures` is only relevant to a transport under normal operation between Master and Minion, and so it is used by Tornado in `tornado.concurrent.Futures` class. Salt SSH is copying all the states to the target machine and executes all that locally without transport involved.

In the original issue I see no evidence that this is about this module. And if it is, I still would like to see the evidence that lack of import of this module causing this.

And a very clear reproducer.

isbm (changes requested):
@dwoz I have some questions before we move forward:

- How exactly you've concluded that the problem is lack of `concurrent.futures` import? 
- ~Why this is affecting _only_ Python 2 if the same module is not anywhere added also to Python 3?~
- ~Could you explain why Python 3 survives _without_ this module at all? Because this module has to be added the same way as in Python 2. Yet it is not added and yet it works. Why?~

Unless I am missing something, the `concurrent.futures` is only relevant to a transport under normal operation between Master and Minion, and so it is used by Tornado in `tornado.concurrent.Futures` class. Salt SSH is copying all the states to the target machine and executes all that locally without transport involved. So technically, this module is not supposed to be even triggered under Salt SSH operation.

In the original issue I see no evidence that this is about this module. And if it is, I still would like to see the evidence that lack of import of this module causing this (tracebacks, root of the cause etc).

And a very clear reproducer.

s0undt3ch:
Python 3 includes `concurrent.futures` in stdlib.

isbm:
@s0undt3ch right. Forgot to look at `.spec` that we explicitly require `python-futures` package for 2.x. :persevere:  That explains affection for Py2 only.

Any idea what exactly happening there and why?

dwoz:
@isbm Tornado is being included in the thin dir which is causing the need for `concurrent.futures`. I determined `concurrent.futures` is missing by running salt-call in a python 2 thin environment. The regression test in this PR confirms the issue and the fix.

rallytime (approved):

isbm (approved):
@dwoz Ok, this makes sense. Thanks!

`/me` quietly ports the same thing to `develop` (Fluorine): https://github.com/saltstack/salt/pull/48289

gtmanfred (approved):

Issue #48141, opened 2018-06-14
salt-api request causes exception


zerthimon:
```
Traceback (most recent call last):
  File "/usr/lib/python2.7/dist-packages/salt/netapi/rest_cherrypy/app.py", line 858, in hypermedia_handler
    ret = cherrypy.serving.request._hypermedia_inner_handler(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/cherrypy/_cpdispatch.py", line 34, in __call__
    return self.callable(*self.args, **self.kwargs)
  File "/usr/lib/python2.7/dist-packages/salt/netapi/rest_cherrypy/app.py", line 1286, in POST
    token=cherrypy.session.get('token')))
  File "/usr/lib/python2.7/dist-packages/salt/netapi/rest_cherrypy/app.py", line 1182, in exec_lowstate
    for i in ret:
  File "/usr/lib/python2.7/dist-packages/salt/client/__init__.py", line 562, in cmd_batch
    salt.utils.versions.warn_until(
UnboundLocalError: local variable 'salt' referenced before assignment
```

```
Salt Version:
           Salt: 2018.3.1
 
Dependency Versions:
           cffi: Not Installed
       cherrypy: 3.2.2
       dateutil: 2.4.2
      docker-py: 3.3.0
          gitdb: 0.6.4
      gitpython: 1.0.1
          ioflo: Not Installed
         Jinja2: 2.8
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: 1.0.3
   msgpack-pure: Not Installed
 msgpack-python: 0.4.6
   mysql-python: Not Installed
      pycparser: Not Installed
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: Not Installed
         Python: 2.7.12 (default, Dec  4 2017, 14:50:18)
   python-gnupg: 0.3.8
         PyYAML: 3.11
          PyZMQ: 15.2.0
           RAET: Not Installed
          smmap: 0.9.0
        timelib: Not Installed
        Tornado: 4.2.1
            ZMQ: 4.1.4
 
System Versions:
           dist: Ubuntu 16.04 xenial
         locale: UTF-8
        machine: x86_64
        release: 4.4.0-1061-aws
         system: Linux
        version: Ubuntu 16.04 xenial
```

zerthimon:
Seems to happen only for "batch" jobs.

garethgreenaway:
[...]

PR #48236, opened 2018-06-21
[2018.3] Fix to cmd_batch

garethgreenaway:
### What does this PR do?
Fixing cmd_batch to work correctly when called via salt-api.

#48141 

### Previous Behavior
When running from the salt-api this would result in an exception, presumably because salt.utils.versions was unavailable.

### New Behavior
import salt.utils.versions inside cmd_batch.

### Tests written?
Yes

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/client/__init__.py
+++ b/salt/client/__init__.py
@@ -559,6 +559,10 @@ class LocalClient(object):
             {'stewart': {...}}
         '''
         if 'expr_form' in kwargs:
+            # We need to report salt.utils.versions here
+            # even though it has already been imported.
+            # when cmd_batch is called via the NetAPI
+            # the module is unavailable.
             import salt.utils.versions
             salt.utils.versions.warn_until(
                 'Fluorine',
--- a/salt/client/__init__.py
+++ b/salt/client/__init__.py
@@ -559,7 +559,7 @@ class LocalClient(object):
             {'stewart': {...}}
         '''
         if 'expr_form' in kwargs:
-            # We need to report salt.utils.versions here
+            # We need to re-import salt.utils.versions here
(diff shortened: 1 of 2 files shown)


isbm:
@garethgreenaway how come? This module is imported already and is used all over the place.

isbm (changes requested):
This is really weird. The module is imported and is used all over the place. And only that one function fails?? How come?

gtmanfred (approved):

gtmanfred:
Probably because of multiprocessing

salt/client/__init__.py
```diff
@@ -559,6 +559,7 @@ def cmd_batch(
             {'stewart': {...}}
         '''
         if 'expr_form' in kwargs:
+            import salt.utils.versions
```
rallytime: Can you add a comment here about why this later import is needed so we don't accidentally remove it in the future?

rallytime (approved):
And a test to boot! Well done.

rallytime:
@garethgreenaway That test is failing, however. Can you take a look?

salt/client/__init__.py
```diff
@@ -559,6 +559,11 @@ def cmd_batch(
             {'stewart': {...}}
         '''
         if 'expr_form' in kwargs:
+            # We need to re-import salt.utils.versions here
+            # even though it has already been imported.
+            # when cmd_batch is called via the NetAPI
+            # the module is unavailable.
```
isbm: :+1:

isbm (approved):
@garethgreenaway thanks for the explanation comment.

PR #48239, opened 2018-06-21
Redis sdb mod

mirceaulinic:
### What does this PR do?

Adds an SDB module for Redis.

### Tests written?

No - but happy to provide: I'm unsure how to mock a connection to a Redis server, without having the Redis server.

### Commits signed with GPG?

No

### DIFF
--- a/salt/sdb/redis_sdb.py
+++ b/salt/sdb/redis_sdb.py
@@ -54,7 +54,7 @@ def set_(key, value, profile=None):
         return False
     redis_kwargs = profile.copy()
     redis_kwargs.pop('driver')
-    redis_conn = redis.Redis(**redis_kwargs)
+    redis_conn = redis.StrictRedis(**redis_kwargs)
     return redis_conn.set(key, value)
 
 
@@ -66,7 +66,7 @@ def get(key, profile=None):
         return False
     redis_kwargs = profile.copy()
     redis_kwargs.pop('driver')
-    redis_conn = redis.Redis(**redis_kwargs)
+    redis_conn = redis.StrictRedis(**redis_kwargs)
     return redis_conn.get(key)
 
 
@@ -78,5 +78,5 @@ def delete(key, profile=None):
         return False
     redis_kwargs = profile.copy()
     redis_kwargs.pop('driver')
-    redis_conn = redis.Redis(**redis_kwargs)
+    redis_conn = redis.StrictRedis(**redis_kwargs)
     return redis_conn.get(key)
--- a/salt/sdb/redis_sdb.py
+++ b/salt/sdb/redis_sdb.py
@@ -3,6 +3,8 @@
 Redis SDB module
 ================
 
+ .. versionadded:: Fluorine
+
 This module allows access to Redis  using an ``sdb://`` URI.
 
 Like all SDB modules, the Redis module requires a configuration profile to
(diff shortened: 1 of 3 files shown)


salt/sdb/redis_sdb.py
```diff
@@ -0,0 +1,82 @@
...
+
+def set_(key, value, profile=None):
+    '''
+    Set a value into the Redis SDB.
+    '''
+    if not profile:
+        return False
+    redis_kwargs = profile.copy()
+    redis_kwargs.pop('driver')
+    redis_conn = redis.Redis(**redis_kwargs)
```
gtmanfred: Fyi, we pack the minion modules into sdb.

https://github.com/saltstack/salt/blob/develop/salt/loader.py#L892

So, you should just be able to do `__salt__['redis.set'](key, value, **redis_kwargs)`

gtmanfred (changes requested):
Just a small change to use the already written redis module, instead of importing redis here.

Thanks,
Daniel

salt/sdb/redis_sdb.py
mirceaulinic: I tried on 2017.7 and it doesn't work like that. Did that change in 2018.3?

Either way, the Redis module only accepts `host`, `password`, `port`, and `db` as arguments, while there are many more available (and with the code I proposed we don't impose any restriction - some of them are very important!) - see https://redis-py.readthedocs.io/en/latest/#contents. What do you think?

gtmanfred: OK, those should also be added to the redis module, but we don't have to do that now, but this should still use the StrictRedis class

mirceaulinic: Yes, good catch, should use `StrictRedis` indeed.

mirceaulinic: Pushed https://github.com/saltstack/salt/pull/48239/commits/3f17add324361e177fc9eab8bcfb98cf91109e18 to solve this.

gtmanfred (approved):

salt/sdb/redis_sdb.py
```diff
@@ -0,0 +1,82 @@
...
+      host: 127.0.0.1
+      port: 6379
+      password: pass
+      db: 1
+
+The ``driver`` refers to the Redis module, all other options are optional.
+For option details see: https://redis-py.readthedocs.io/en/latest/.
+
+'''
+from __future__ import absolute_import
```
rallytime: We also need to import `print_function` and `unicode_literals` here for PY3 compatibility.

rallytime (changes requested):
One small request from me. :)

This could use a `.. versionadded:: Fluorine` tag as well.

salt/sdb/redis_sdb.py
mirceaulinic: Good points - added: https://github.com/saltstack/salt/pull/48239/commits/30cd544d7ebc2491db6034c4215d59b73db58def

Regarding `print_function` and `unicode_literals`: I've seen this has been added across all the modules, and I'm afraid I didn't understand that - could you please explain why is `print_function` (for example) needed although the module doesn't call the `print()` function? Thanks!

rallytime (approved):

rallytime:
Thanks @mirceaulinic!

Issue #38306, opened 2016-12-16
The `apache.configfile` state does not allow multiple same sections


vutny:
### Description of Issue/Question
Looks like the `apache.configfile` state function has limited capabilities in setting multiple sections of the same type at a single level of config hierarchy. 

### Setup
Taking the example from the docs:
```yaml
/etc/httpd/conf.d/website.com.conf:
  apache.configfile:
    - config:
      - VirtualHost:
          this: '*:80'
          ServerName:
            - website.com
          ServerAlias:
            - www.website.com
            - dev.website.com
          ErrorLog: logs/website.com-error_log
          CustomLog: logs/website.com-access_log combined
          DocumentRoot: /var/www/vhosts/website.com
          Directory:
            this: /var/www/vhosts/website.com
            ...
```

It is unclear how to add additional config options for other sub-directories to this virtual host.
To achive resulting config look like this:

```xml
<VirtualHost *:80>
ServerName website.com
ServerAlias www.website.com dev.website.com
ErrorLog logs/website.com-error_log
CustomLog logs/website.com-access_log combined
DocumentRoot /var/www/vhosts/website.com
<Directory /var/www/vhosts/website.com>
Require all granted
Options Indexes
</Directory>
<Directory /var/www/vhosts/website.com/admin>
AllowOverride None
Order Deny,Allow
Allow from 192.168.100.0/24
</Directory>
</VirtualHost>
```

Ch3LL:
@vutny i am able to replicate this behavior And when adding another `Directory` like below:

```
          Directory:
            this: /var/www/vhosts/website.com
          Directory:
            this: /var/www/vhosts/website2.com
```
or this:

```
          Directory:
            this: /var/www/vhosts/website.com
            this: /var/www/vhosts/website2.com
```

I receive a conflicting id error. Looks like we need to add the ability to add more then one directory. Thanks

Also here is a docker container for anyone that wants to quickly replicate this:

1. `docker run -it -v /home/ch3ll/git/salt/:/testing ch3ll/issues:38306 salt-call --local state.sls apache`

aaronlauterer:
Any news regarding this? I am still experiencing it with 2017.7.5 (Nitrogen).

boltronics:
I too would like to express interest in seeing this issue solved.

It seems to me that the syntax should be changed so that the items in the `VirtualHost` dict (and elsewhere) are also presented as a list to avoid duplicate keys when rendering the YAML. ie. Instead of

```
[...]

PR #48244, opened 2018-06-21
Fixed apache.configfile to work with multiple same sections.

slivik:
### What does this PR do?
Fixes `apache.configfile` so that it allows to create Apache config files with multiple same section containers. It is possible to define structure like before (through a dict) and newly also through lists which allows sections with same name, e.g. `<Directory>`. Behavior is backward compatible and the new one is described in updated doc section in `apache` state module.

As side-effect it also fixes issue that when two structures were used at the same salt state run, the second structure was wrong since the previous run removed all `this:` items. So intended `<Directory /some/path>` was always generated as `<Directory>`.

[#38306 - The `apache.configfile` state does not allow multiple same sections](https://github.com/saltstack/salt/issues/38306)

### Tests written?

Yes

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/states/apache.py
+++ b/salt/states/apache.py
@@ -36,7 +36,55 @@ the above word between angle brackets (<>).
                   - Indexes
                   - FollowSymlinks
                 AllowOverride: All
+
+.. versionchanged:: 2018.3
+
+Allows having the same section container multiple times (e.g. <Directory /path/to/dir>).
+
+YAML structure stays the same only replace dictionary with a list.
+
+When a section container does not have mandatory attribute, such as <Else>,
+it still needs keyword ``this`` with empty string (or "\b" if nicer output is required - without space).
+
+.. code-block:: yaml
+
+    /etc/httpd/conf.d/website.com.conf:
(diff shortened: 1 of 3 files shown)


salt/states/apache.py
```diff
@@ -36,7 +36,55 @@
                   - Indexes
                   - FollowSymlinks
                 AllowOverride: All
+
+.. versionchanged:: 2018.3
```
rallytime: This should be `Fluorine`.

rallytime (commented):
I realize you might be following what's already in the module, but is there some reason we're using `print` statements instead of `log`?

gtmanfred:
Print is used to write to the config file. I will look at this tomorrow or Monday and review it.

gtmanfred (changes requested):
I think this looks fine, but I would really like to see at least a unittest added for both this version and the old version to verify that we are not breaking backwards compatibility.

rallytime (approved):

gtmanfred (approved):

PR #48254, opened 2018-06-21
add ansible playbook runner to ansiblegate module

gtmanfred:
### What does this PR do?
Adds the ability to run ansible playbooks from salt execution modules or from states, specifically for orchestration states.

### Tests written?

Added

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/modules/ansiblegate.py
+++ b/salt/modules/ansiblegate.py
@@ -29,6 +29,7 @@ any Ansible module to respond.
 '''
 
 from __future__ import absolute_import, print_function, unicode_literals
+import json
 import os
 import sys
 import logging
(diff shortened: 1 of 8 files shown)


salt/states/ansiblegate.py
```diff
@@ -99,3 +103,85 @@ def __virtual__():
...
+            git_kwargs = {}
+        __states__['git.latest'](
+            name=gitrepo,
+            target=rundir,
+            **git_kwargs,
+        )
+    if ansible_kwargs is None:
+        ansible_kwargs = {}
+    checks = __salt__['ansible.playbooks'](name, rundir=rundir, check=True, diff=True, **ansible_kwargs)
+    if all(not check['changed'] for _, check in six.iteritems(checks['stats'])):
```
terminalmage: You can use `six.itervalues()` here and not need to use the "throwaway" variable.

terminalmage (changes requested):
Just one minor change I would make, otherwise this is ++.

salt/states/ansiblegate.py
gtmanfred: cool

gtmanfred:
Still need to write tests for this.

salt/modules/ansiblegate.py
```diff
@@ -276,3 +277,99 @@ def list(pattern=None):
...
+    :return: Playbook return
+
+    CLI Example:
+
+    .. code-block:: bash
+
+        salt 'ansiblehost'  ansible.playbook playbook=/srv/playbooks/play.yml
+    '''
+    command = ['ansible-playbook', playbook]
+    if check is True:
```
isbm: What is the reason to explicitly check the type alongside the value? I.e. why `if check:` is not sufficient?

salt/modules/ansiblegate.py
```diff
@@ -276,3 +277,99 @@ def list(pattern=None):
...
+        command.append('--check')
+    if diff is True:
+        command.append('--diff')
+    if extra_vars is not None:
+        command.append('--extra-vars={0}'.format(json.dumps(extra_vars)))
```
isbm: Hmm, still the same question though. I'd say `if extra_vars:`

salt/states/ansiblegate.py
```diff
@@ -99,3 +103,85 @@ def __virtual__():
...
+
+
+def _client():
+    '''
+    Get a fileclient
+    '''
+    return salt.fileclient.get_file_client(__opts__)
+
+
+def _changes(plays):
```
isbm: Docstring?

isbm (changes requested):
Generally, if we want to validate the params, we should do it explicitly. Otherwise I would use Pythonic checks (see the comment).

Otherwise, excellent, woohoo!! :sunglasses:

gtmanfred:
Test added

gtmanfred:
re-run py

salt/states/ansiblegate.py
```diff
@@ -99,3 +103,88 @@ def __virtual__():
...
+            - git_kwargs:
+                rev: master
+    '''
+    ret = {
+        'result': False,
+        'changes': {},
+        'comment': 'Running playbook {0}'.format(name),
+        'name': name,
+    }
+    if gitrepo is not None:
```
isbm: If `gitrepo` is an empty string this will fail.

salt/states/ansiblegate.py
```diff
@@ -99,3 +103,88 @@ def __virtual__():
...
+            rundir = _client()._extrn_path(gitrepo, 'base')
+        if git_kwargs is None:
+            git_kwargs = {}
+        __states__['git.latest'](
+            name=gitrepo,
+            target=rundir,
+            **git_kwargs,
+        )
+    if ansible_kwargs is None:
+        ansible_kwargs = {}
```
isbm: Maybe this is better way, unless we do not want to validate all that and crash with the error "You've passed an empty string instead of a parameters" (or something like that):

```python
if not ansible_kwargs:
    ansible_kwargs = {}
```

isbm (changes requested):
Cool, thanks! Some last little nitpicks though.

rallytime:
Lint: https://jenkins.saltstack.com/job/PR/job/salt-pr-lint-n/22834/violations/

salt/states/ansiblegate.py
gtmanfred: I am just going to check if it a dict and if it isn't then i will set it to {}.

gtmanfred:
This should be all set to go on the next test run.

gtmanfred:
https://jenkins.saltstack.com/job/PR/job/salt-pr-rs-cent7-n/20023/testReport/integration.states.test_ansiblegate/AnsiblePlaybooksTestCase/

terminalmage (commented):
Is `gitrepo` part of a naming convention used elsewhere? If not, it may make sense to change it. I'm just worried about confusion between `gitrepo` (no underscore) and `git_kwargs` (underscore). It seems like this could lead to confusion/typos.

isbm (changes requested):
@terminalmage actually a _very_ good catch, it slipped out of my radar: the `git_<something>` is better convention than `git<something>`.

@gtmanfred could you please appease BOFH (and terminalmage) by looking into name conventions and checking to `None` one more time? :wink: I would appreciate that.

Other then that, it is good to go as of my current views.

isbm (approved):
LOL :laughing:  :+1: good to go.

terminalmage (approved):

rallytime:
@gtmanfred Lint :)

salt/states/ansiblegate.py
```diff
@@ -99,3 +105,94 @@ def __virtual__():
...
+                           module besides the `name` and `target`
+
+    :return: Ansible playbook output.
+
+    .. code-block:: yaml
+
+        run nginx install:
+          ansible.playbooks:
+            - name: install.yml
+            - gitrepo: git://github.com/gituser/playbook.git
```
isbm: Docs :)

tests/integration/states/test_ansiblegate.py
```diff
@@ -0,0 +1,73 @@
...
+            'ansible.playbooks',
+            name='remove.yml',
+            gitrepo='git://github.com/gtmanfred/playbooks.git',
+            ansible_kwargs={'inventory': self.inventory}
+        )
+        self.assertSaltTrueReturn(ret)
+        ret = self.run_state(
+            'ansible.playbooks',
+            name='install.yml',
+            gitrepo='git://github.com/gtmanfred/playbooks.git',
```
isbm: There is no such param as `gitrepo` anymore (and so this should be fixed). But why the test is passing anyway, makes me wondering if the `run_state` is reliable enough reporting wrong parameter has been passed?

isbm (changes requested):
@gtmanfred docs, tests. :)

salt/states/ansiblegate.py
gtmanfred: bah... i did a search replace... apparently i am just terrible at regex.

tests/integration/states/test_ansiblegate.py
gtmanfred: It did fail this time https://jenkins.saltstack.com/job/PR/job/salt-pr-linode-ubuntu14-n/23939/testReport/junit/integration.states.test_ansiblegate/AnsiblePlaybooksTestCase/test_ansible_playbook/

With not having gitrepo in the keys.

isbm (approved):
@gtmanfred thanks! Hopefully now it is OK to go.

rallytime:
@gtmanfred You've got a test failure here: https://jenkins.saltstack.com/job/PR/job/salt-pr-linode-ubuntu14-n/24000/

gtmanfred:
re-run py

rallytime:
@gtmanfred The Py3 tests are failing your new test. Can you take a look?

https://jenkins.saltstack.com/job/PR/job/salt-pr-linode-cent7-py3/6163/

- integration.states.test_ansiblegate.AnsiblePlaybooksTestCase.test_ansible_playbook

gtmanfred:
Yup, i am working on it now.

It is failing because ansible can't be installed on py3.4, so i was installing it with only py2 in salt-jenkins.

But this makes it so the ansiblegate modules don't load because the module stuff stops if it can't import ansible, which this doesn't need.

gtmanfred:
Tests should be fixed on this run.

gtmanfred:
Ok, this should be ready.

rallytime:
@gtmanfred There is still a test failing here, unfortunately:

https://jenkins.saltstack.com/job/PR/job/salt-pr-linode-ubuntu14-n/24138/testReport/junit/unit.loader.test_globals/LoaderGlobalsTest/test_states/

gtmanfred:
re-run py

gtmanfred:
re-run all

rallytime:
re-run py

rallytime:
@gtmanfred I think the following test is related. I am not seeing this fail elsewhere:

- integration.modules.test_cmdmod.CMDModuleTest.test_run

https://jenkins.saltstack.com/job/PR/job/salt-pr-linode-ubuntu16-py3/11261/

gtmanfred:
🎉

PR #48260, opened 2018-06-22
Adding the initial version of the Netmiko Proxy and Execution modules

mirceaulinic:
### What does this PR do?

Adding the initial version of the Netmiko Proxy and Execution modules (based on a third party library called [netmiko](https://github.com/ktbyers/netmiko/tree/develop/netmiko)). This gates the support to be able to interact and manage a large number of network operating systems (though over SSH!).

While the user can always manage the remote device through the ``netmiko`` Proxy, the Execution Module is flexible enough to be able to run arbitrary commands when running under a regular Minion, or a different flavour of a Proxy Minion (I have expanded more on this in the documentation).

### DIFF
--- a/doc/ref/modules/all/index.rst
+++ b/doc/ref/modules/all/index.rst
@@ -283,6 +283,7 @@ execution modules
     netbox
     netbsd_sysctl
     netbsdservice
+    netmiko_mod
     netscaler
     network
     neutron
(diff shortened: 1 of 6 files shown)


ngrundler:
+1

This looks great, and will be particularly useful for us since we have many legacy devices that we want to automate using Salt, but for which we don't want to run a full proxy-minion(primarily campus access switches with relatively static configurations).

rallytime (approved):

gtmanfred (approved):

PR #39099, opened 2017-02-01
Log module

mirceaulinic:
### What does this PR do?

I am proposing this module... that is just logging :) Yeah, I find it funny too!

I don't know how you guys feel, but I find Jinja the most painful thing to debug.
Although the `show_full_context` function (https://docs.saltstack.com/en/latest/topics/jinja/index.html#debugging) is very handy, sometimes I feel like I would need to log elements and see what's going on inside a for loop, etc. 

So with this module, one would be able to:

```jinja
{%- do salt.log.error('testing jinja logging') -%}
```

inside their template and:

```
2017-02-01 01:24:40,728 [salt.loader.localhost.int.module.logmod][ERROR   ][3779] testing jinja logging
```

appears in the minion log.

If you find this sane, please let me know and I will add some tests (kidding), better doc + some notes under the Jinja debugging section - https://docs.saltstack.com/en/latest/topics/jinja/index.html#debugging

Cheers,
Mircea

PR #48267, opened 2018-06-22
Adding a little module as a helper for template rendering on the fly

mirceaulinic:
### What does this PR do?

Nothing really spectacular, this is sort of in-line with other little modules I proposed in the past, e.g., https://github.com/saltstack/salt/pull/39099 or https://github.com/saltstack/salt/pull/42018, so not mind blowing this time either. But I often need to render various templates, and rather than saving the rendered result into a file, I need the text. And here's exactly what this function does, aiming to avoid repeating this code over and over again, and hopefully saving other people's time as well. Being an execution module, can easily be used also from the CLI for checks, or endless alternatives. Hope you're also finding it helpful.

### DIFF
--- a/doc/ref/modules/all/index.rst
+++ b/doc/ref/modules/all/index.rst
@@ -426,6 +426,7 @@ execution modules
     system
     system_profiler
     systemd
+    template
     telegram
     telemetry
     temp
--- a/doc/ref/modules/all/salt.modules.template.rst
+++ b/doc/ref/modules/all/salt.modules.template.rst
@@ -1,0 +1,6 @@
+=====================
+salt.modules.template
+=====================
+
+.. automodule:: salt.modules.template
+    :members:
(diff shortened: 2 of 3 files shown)


gtmanfred:
This looks good, but what does it offer that slsutil.renderer doesn't? Or could the context and default stuff be added to slsutil?

mirceaulinic:
`slsutil.render` tries to return an object, the Python representation of some structured input (data), while this one returns text.

mirceaulinic:
Oh, damn, you are right @gtmanfred. Silly me... I had this module in my branch for other reasons actually (the support for the Salt fileserver has been added to 2018.3 only, and I'm still running 2017.7).

mirceaulinic:
Reopening as discussed in private: `slsutil.render` seems to try to parse the data instead of trying to render: `salt '*' slsutil.renderer https://bit.ly/2yuSs2Y hostname=aaa default_renderer=jinja`

gtmanfred (approved):
This adds the ability to test templates for other other stuff besides sls files

rallytime (approved):

PR #48273, opened 2018-06-22
More SMBIOS Information for mainboard and chassis.

ixs:
### What does this PR do?
 Try to not just fetch the system info from smbios but also separate
mainboard and chassis manufacturer, name and version.

In case of whitebox or built-to-order machines this can be rather
valueable as the system manufacturer might be very different from the
mainboard manufacturer.

Add chassis-type detection to clean and remove unknown values.

Drive-By: Use _clean_value to remove known invalid values from smbios
          grains: E.g. disregard "0123456789" as a serial number.
Drive-By: Refactor smbios/dmidecode calls to be in-line with the other
          OS variants and ensure returned values are _clean_value()'d.

### Previous Behavior
Limited SMBios information returned

### New Behavior
More SMBios information returned instead

### Tests written?
No. Grains seem to have no testing framework unfortunately.

### DIFF
--- a/salt/grains/core.py
+++ b/salt/grains/core.py
@@ -2357,35 +2357,35 @@ def _hw_data(osdata):
     # Chassis Type table taken from the System Management BIOS (SMBIOS)
     # Reference Specification (DSP0134) Version 2.7.1
     chassis_types = {
-      0x01: 'Other',
-      0x02: 'Unknown',
-      0x03: 'Desktop',
-      0x04: 'Low Profile Desktop',
-      0x05: 'Pizza Box',
-      0x06: 'Mini Tower',
-      0x07: 'Tower',
-      0x08: 'Portable',
-      0x09: 'LapTop',
-      0x0a: 'Notebook',
-      0x0b: 'Hand Held',
-      0x0c: 'Docking Station',
-      0x0d: 'All in One',
-      0x0e: 'Sub Notebook',
-      0x0f: 'Space-saving',
-      0x10: 'Lunch Box',
-      0x11: 'Main Server Chassis',
-      0x12: 'Expansion Chassis',
-      0x13: 'SubChassis',
-      0x14: 'Bus Expansion Chassis',
-      0x15: 'Peripheral Chassis',
-      0x16: 'RAID Chassis',
-      0x17: 'Rack Mount Chassis',
-      0x18: 'Sealed-case PC',
-      0x19: 'Multi-system chassis',
-      0x1a: 'Compact PCI',
-      0x1b: 'Advanced TCA',
-      0x1c: 'Blade',
-      0x1d: 'Blade Enclosure'
+        0x01: 'Other',
+        0x02: 'Unknown',
+        0x03: 'Desktop',
+        0x04: 'Low Profile Desktop',
+        0x05: 'Pizza Box',
+        0x06: 'Mini Tower',
+        0x07: 'Tower',
+        0x08: 'Portable',
+        0x09: 'LapTop',
+        0x0a: 'Notebook',
+        0x0b: 'Hand Held',
+        0x0c: 'Docking Station',
+        0x0d: 'All in One',
+        0x0e: 'Sub Notebook',
+        0x0f: 'Space-saving',
+        0x10: 'Lunch Box',
+        0x11: 'Main Server Chassis',
+        0x12: 'Expansion Chassis',
+        0x13: 'SubChassis',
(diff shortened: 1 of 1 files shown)


salt/grains/core.py
```diff
@@ -2343,6 +2353,41 @@ def _hw_data(osdata):
         return {}
 
     grains = {}
+
+    # Chassis Type table taken from the System Management BIOS (SMBIOS)
+    # Reference Specification (DSP0134) Version 2.7.1
+    chassis_types = {
+      0x01: 'Other',
```
terminalmage: We indent using 4 spaces

salt/grains/core.py
```diff
@@ -2352,22 +2397,35 @@ def _hw_data(osdata):
...
+                                grains['chassis_type'] = chassis_types[int(grains['chassis_type'])]
+                            except KeyError:
+                                grains['chassis_type'] = 'Unknown'
                 except (IOError, OSError) as err:
                     # PermissionError is new to Python 3, but corresponds to the EACESS and
                     # EPERM error numbers. Use those instead here for PY2 compatibility.
                     if err.errno == EACCES or err.errno == EPERM:
                         # Skip the grain if non-root user has no access to the file.
                         pass
+        grains = dict([(key, val) for key, val in grains.items() if val is not None])
```
terminalmage: You can use a dict comprehension here, it'll run quicker. We no longer support Python 2.6 so dict comprehensions are OK. Also, using six.iteritems(grains) will make it more efficient on Python 2:

```python
grains = {key: val for key, val in six.iteritems(grains) if val is not None}
```

terminalmage (changes requested):

ixs:
@terminalmage Right. This looks better I believe.

salt/grains/core.py
```diff
@@ -1168,6 +1168,9 @@ def _clean_value(key, val):
         # AssetTag0. Manufacturer04. Begone.
         if re.search(r'manufacturer|to be filled|available|asset|^no(ne|t)', val, flags=re.IGNORECASE):
             return None
+    elif key == 'chassis_type':
+        if val in [0x02, '0x02', '2', 'Unknown']:
+            return None
```
isbm: You don't need to explicitly write `2` in hexadecimal, as Python will implicitly convert it during the comparison, so just `2` is fine (the `2 == 0x02` or `2 in [0x02]` and vice versa returns `True`). However, this sort of list isn't reliable and prone to bugs, especially if `chassis_type` appears to be like `0xFF` or `0` or whatever proprietary hardware will return you — it also should be considered as unknown (it is above `0x1D`), etc. Hence I would write it this way:

```python
try:
    val = int(val, 16)
    if not val or val == 2 or val > max(chassis_types):
        val = None
except ValueError:
        val = None
```

Since now `val` is already either value or `None`, no need to explicitly `return None` as it will anyway fall off to the last `return`. I know you've just copied the previous code, but [multiple returns aren't nice](http://docs.python-guide.org/en/latest/writing/style/#returning-values) (and so the older code too).

isbm (changes requested):
@ixs thanks for a great job! Could you please look at my note and also update/write some unit tests for this?

salt/grains/core.py
ixs: @isbm Please have another look at the list of Values I consider unknown. I did not check for 2 or 0x02 which would indeed be duplicate. I specifically checked for 0x02 and 2 as string values in case someone somewhere forgot a cast. The case of int(2) is already covered with the first val 0x02.

And if you look at https://github.com/saltstack/salt/pull/48273/commits/b78ba338a42aa28f54f6dfb9a476f97610da4c22#diff-6e894b39b67c4bab67796752744ea9d5R2418 the case of unknown value like 0xff or -1 or 0 is already covered as well.

terminalmage (approved):

rallytime:
@ixs It looks like this PR has a merge-conflict in it. Can you fix that up?

cachedout:
In addition to the merge conflict, we also need a follow-up from @isbm here.

salt/grains/core.py
isbm: @ixs I would prefer you just cast everything, regardless who forgot what, instead of creating a special hard-coded list of things that you assume you've got right.

isbm (changes requested):

cachedout:
@ixs We'd love to move forward on this. Any chance you could fixup the merge conflict here?

PR #48282, opened 2018-06-24
The smartos state's vm_present should support docker

sjorge:
### What does this PR do?
In follow up of the PR that added docker uuid support to the imgadm module and states... this one allows VM's to be managed based on a docker image.

N/a

### Previous Behavior
Salt would error when specifying a docker uuid.

### New Behavior
Salt does the expected thing creates the vm from the docker image.

### Tests written?
No

### Commits signed with GPG?
No


sjorge:
Hit a weird intermittent issue, investigating that now...

rallytime (approved):

PR #48284, opened 2018-06-24
virt module: Allow NVRAM unlinking on DOM undefine

alexandruavadanii:
UEFI-enabled VMs usually have pflash (NVRAM) devices attached,
which require one additional libvirt flag to be passed at 'undefine'.

This is usually the case for AArch64 (arm64) VMs, where AAVMF (AA64
UEFI) is the only supported guest bootloader.

### What does this PR do?
Prepares virt module for AArch64 support by allowing undefine operations on UEFI-enabled VMs.

N/A

### Previous Behavior
undefine fails since VMs were booted with an NVRAM device attached.

### New Behavior
NVRAM devices are destroyed as well on VM undefine.

### Tests written?

No

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/modules/virt.py
+++ b/salt/modules/virt.py
@@ -2381,7 +2381,11 @@ def undefine(vm_, **kwargs):
     '''
     conn = __get_conn(**kwargs)
     dom = _get_domain(conn, vm_)
-    ret = dom.undefine() == 0
+    if getattr(libvirt, 'VIR_DOMAIN_UNDEFINE_NVRAM', False):
+        # This one is only in 1.2.8+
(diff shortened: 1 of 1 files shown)


rallytime (approved):

alexandruavadanii:
I'm interested in backporting this to 2016.11; according to the contribution guidelines, this will be handled automatically by the Salt team, but just in case you need any action from my side, don't hesitate to let me know.

rallytime:
Hi @alexandruavadanii - You're right, we won't be back-porting this to 2016.11 as we're not maintaining that branch anymore.

I attempted to backport this to 2017.7, which we are still maintaining, but since the `virt` module has been reworked significantly in `develop`, it wouldn't back-port cleanly.

PR #38339, opened 2016-12-19
Migrate NAPALM modules to be used in both proxy and regular minion

mirceaulinic:
### What does this PR do?

Transforms the existing NAPALM modules to be usable with both proxy minions and straight minions. The key of this PR is the ```proxy_napalm_wrap``` decorator which injects the global var ```napalm_device``` in the module function.

Direct implications:

- moved the logic from the proxy module in the utils, in order to be re-used
- allow **kwargs for all functions
- ```__virtual__()``` returns the  ```__virtualname__``` no matter if proxy or not (but still fails if NAPALM is not installed)
- configuration-related functions need to pass an argument ```inherit_napalm_device``` with the original ```napalm_device``` object containing the configuration session. If this is not sent, at each function underneath will be open another config DB which leads to failure, as network devices allow one single config session open at a time.

Otherwise, everything remains unchanged.

PR #48290, opened 2018-06-25
Enhance the proxy_napalm_wrap decorator to allow "proxyless" execution

mirceaulinic:
Usually the Salt proxies have been designed using a single methodology: for
each device you aim to manage, start one Proxy Minion. The NAPALM modules didn't
make an exception, however, beginning with https://github.com/saltstack/salt/pull/38339
(therefore starting with release Nitrogen), the functionality has been enhanced
in such a way that we can execute the code when we are able to install the
regular Minion directly on the network hardware.

There is another use case, for something, let's call it "proxyless" when we
don't particularly need to start a Proxy process for each device, but rather
simply invoke arbitrary functions. These changes make this possible, so with one
single Minion (whether Proxy or regular), one is able to execute Salt commands
going through the NAPALM library to connect to the remote network device by
specifying the connection details from the command line and/or opts/pillar, e.g.
``salt server1 bgp.neighbors driver=junos host=1.2.3.4 username=salt``.

If the ``server1`` Minion has the following block in the Pillar (for example):

```yaml
napalm:
  driver: junos
  username: salt
```

The user would only need to provide the rest of the credentials (depending on
each individual case):

``salt server1 bgp.neighbors host=1.2.3.4``.

-----

Having the following configuration in the opts/pillar, for example:

```yaml
napalm:
  username: salt
  password: test

napalm_inventory:
  1.2.3.4:
    driver: eos
  edge01.bzr01:
    driver: junos
```

With the above available in the opts/pillar for a Minion, say ``server1``, the
user would be able to execute: ``salt 'server1' bgp.neighbors host=1.2.3.4`` or
``salt 'server1' net.arp host=edge01.bzr01``, etc.

### DIFF
--- a/salt/utils/napalm.py
+++ b/salt/utils/napalm.py
@@ -17,6 +17,7 @@ Utils for the NAPALM modules and proxy.
 
 # Import Python libs
 from __future__ import absolute_import, unicode_literals, print_function
+import copy
 import traceback
 import logging
 import importlib
@@ -422,7 +423,7 @@ def proxy_napalm_wrap(func):
                 napalm_opts.update(inventory_opts)
                 log.debug('Merging the config for %s with the details found in the napalm inventory:', host)
                 log.debug(napalm_opts)
-            opts = opts.copy()  # make sure we don't override the original
+            opts = copy.deepcopy(opts)  # make sure we don't override the original
             # opts, but just inject the CLI args from the kwargs to into the
             # object manipulated by ``get_device_opts`` to extract the
             # connection details, then use then to establish the connection.
--- a/salt/utils/napalm.py
+++ b/salt/utils/napalm.py
@@ -405,12 +405,23 @@ def proxy_napalm_wrap(func):
             log.debug('Not running in a NAPALM Proxy Minion')
             _salt_obj = wrapped_global_namespace.get('__salt__')
             napalm_opts = _salt_obj['config.get']('napalm', {})
+            napalm_inventory = _salt_obj['config.get']('napalm_inventory', {})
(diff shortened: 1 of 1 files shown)


salt/utils/napalm.py
```diff
@@ -398,10 +397,36 @@ def func_wrapper(*args, **kwargs):
...
+                   napalm_opts.get('fqdn') or napalm_opts.get('ip')
+            if host and napalm_inventory and isinstance(napalm_inventory, dict) and\
+               host in napalm_inventory:
+                inventory_opts = napalm_inventory[host]
+                log.debug('Found %s in the NAPALM inventory:', host)
+                log.debug(inventory_opts)
+                napalm_opts.update(inventory_opts)
+                log.debug('Merging the config for %s with the details found in the napalm inventory:', host)
+                log.debug(napalm_opts)
+            opts = opts.copy()  # make sure we don't override the original
```
gtmanfred: you will probably want to do a `copy.deepcopy(opts)` here instead, to get any nested options.

gtmanfred (changes requested):

gtmanfred:
Other than one small change, this LGTM

salt/utils/napalm.py
mirceaulinic: Added https://github.com/saltstack/salt/pull/48290/commits/bf7baaebe20a392db0ae40d7a3478082dbf385b8 for fix this.

gtmanfred (approved):

rallytime (approved):

PR #48291, opened 2018-06-25
Mask rendered data (in logs)

isbm:
### What does this PR do?

- Fixes a security issue
- Adds an utility to further fix such cases

When you call `salt-ssh -l debug`, you will get the entire roster with the passwords directly in the log. This PR fixes this by matching `key: value` strings in the output data and replaces YAML values with the `** hidden **` string. This can be applied for passwords, user IDs, actual IP addresses etc.

NOTE: this is so far is limited only to `key: value` where `value` is replaced. Feel free to extend it for IP addresses etc.

### Tests written?

Yes

### DIFF
--- a/salt/utils/sanitizers.py
+++ b/salt/utils/sanitizers.py
@@ -74,9 +74,11 @@ def mask_args_value(data, mask):
     but you certainly do not want to put there an actual IP address,
     passwords, user names etc.
 
-    Note, this is working only when data is a single string,
-    ready for print or dump to the log. Also, when the data is formatted
-    as "key: value" in YAML syntax.
+    This can be used for cases where keys in your roster file may contain
+    sensitive data such as IP addresses, passwords, user names, etc.
+
+    Note that this works only when ``data`` is a single string (i.e. when the
+    data in the roster is formatted as ``key: value`` pairs in YAML syntax).
 
     :param data: String data, already rendered.
     :param mask: Mask that matches a single line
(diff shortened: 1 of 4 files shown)


salt/utils/sanitizers.py
```diff
@@ -62,3 +63,34 @@ def hostname(value):
...
+    :return:
+    '''
+    if not mask:
+        return data
+
+    out = []
+    for line in data.split(os.linesep):
+        if fnmatch.fnmatch(line.strip(), mask) and ':' in line:
+            key, value = line.split(':', 1)
+            out.append('{}: {}'.format(key.strip(), '** hidden **'))
```
terminalmage: Just as a precaution, `key.strip()` should probably be `salt.utils.stringutils.to_unicode(key.strip())`, to prevent a `UnicodeDecodeError` if `key` happens to be a `str` type with non-ascii unicode in it.

salt/utils/sanitizers.py
```diff
@@ -62,3 +63,34 @@ def hostname(value):
...
+    '''
+    Mask a line in the data, which matches "mask".
+
+    In case you want to put to the logs rosters or other data,
+    but you certainly do not want to put there an actual IP address,
+    passwords, user names etc.
+
+    Note, this is working only when data is a single string,
+    ready for print or dump to the log. Also, when the data is formatted
+    as "key: value" in YAML syntax.
```
terminalmage: Can we reword the above two paragraphs like this?

```
    This can be used for cases where keys in your roster file may contain
    sensitive data such as IP addresses, passwords, user names, etc.

    Note that this works only when ``data`` is a single string (i.e. when the
    data in the roster is formatted as ``key: value`` pairs in YAML syntax).
```

terminalmage (changes requested):

isbm:
@terminalmage done!

terminalmage (approved):

Issue #48078, opened 2018-06-13
fresh install result on mac = ValueError: invalid literal for float(): 1024,00


sugizo:
### Setup
no config, just started from scratch new python environment in conda

### Steps to Reproduce Issue
```
$ uname
Darwin
$ launchctl limit maxfiles
	maxfiles    8192           10240
$ python -V
Python 2.7.15 :: Anaconda, Inc.
$ pip install salt
$ salt-master --log-level=all
Traceback (most recent call last):
  File "/Users/MacBookPro/miniconda3/envs/python2_test/bin/salt-master", line 11, in <module>
    sys.exit(salt_master())
  File "/Users/MacBookPro/miniconda3/envs/python2_test/lib/python2.7/site-packages/salt/scripts.py", line 95, in salt_master
    import salt.cli.daemons
  File "/Users/MacBookPro/miniconda3/envs/python2_test/lib/python2.7/site-packages/salt/cli/daemons.py", line 48, in <module>
    import salt.utils.parsers
  File "/Users/MacBookPro/miniconda3/envs/python2_test/lib/python2.7/site-packages/salt/utils/parsers.py", line 27, in <module>
    import salt.config as config
  File "/Users/MacBookPro/miniconda3/envs/python2_test/lib/python2.7/site-packages/salt/config/__init__.py", line 98, in <module>
    _DFLT_IPC_WBUFFER = _gather_buffer_space() * .5
  File "/Users/MacBookPro/miniconda3/envs/python2_test/lib/python2.7/site-packages/salt/config/__init__.py", line 91, in _gather_buffer_space
    grains = salt.grains.core._memdata(os_data)
  File "/Users/MacBookPro/miniconda3/envs/python2_test/lib/python2.7/site-packages/salt/grains/core.py", line 511, in _memdata
    grains.update(_osx_memdata())
  File "/Users/MacBookPro/miniconda3/envs/python2_test/lib/python2.7/site-packages/salt/grains/core.py", line 433, in _osx_memdata
    swap_total = float(swap_total[:-1]) * _power
ValueError: invalid literal for float(): 1024,00
```

trying salt --versions-report and salt --help also thrown the same error : 
ValueError: invalid literal for float(): 1024,00

### Versions Report
can not generate salt --version-report, but seen in pip install salt = salt-2018.3.0

any idea ? thx

Ch3LL:
[...]

PR #48304, opened 2018-06-25
Fix macosx grains when swapusage returns comma

Ch3LL:
### What does this PR do?
Apparently macosx can return a comman when running `vm.swapusage`. This handles those cases and adds tests.

https://github.com/saltstack/salt/issues/48078

### Previous Behavior
salt-minion does not start due to valuerror

### New Behavior
salt-minion starts

### Tests written?

Yes

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/grains/core.py
+++ b/salt/grains/core.py
@@ -448,7 +448,7 @@ def _osx_memdata():
     sysctl = salt.utils.path.which('sysctl')
     if sysctl:
         mem = __salt__['cmd.run']('{0} -n hw.memsize'.format(sysctl))
-        swap_total = __salt__['cmd.run']('{0} -n vm.swapusage'.format(sysctl)).split()[2]
+        swap_total = __salt__['cmd.run']('{0} -n vm.swapusage'.format(sysctl)).split()[2].replace(',', '.')
         if swap_total.endswith('K'):
             _power = 2**10
         elif swap_total.endswith('M'):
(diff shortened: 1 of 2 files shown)


salt/grains/core.py
```diff
@@ -448,7 +448,7 @@ def _osx_memdata():
     sysctl = salt.utils.path.which('sysctl')
     if sysctl:
         mem = __salt__['cmd.run']('{0} -n hw.memsize'.format(sysctl))
-        swap_total = __salt__['cmd.run']('{0} -n vm.swapusage'.format(sysctl)).split()[2]
+        swap_total = __salt__['cmd.run']('{0} -n vm.swapusage'.format(sysctl)).split()[2].replace(',', '.')
```
rares-pop: Is this due to the locale?

Here you can find a more generic approach:
https://stackoverflow.com/questions/13362121/convert-python-strings-into-floats-explicitly-using-the-comma-or-the-point-as-se

I wonder if there are no better ways to do it though.

Ch3LL: yeah looks like you are right that this is related to locale. let me look into a better fix. thanks :)

cachedout:
@Ch3LL Any updates here?

rares-pop:
I just read the comments there, and probably is fine to go with the original fix and do something more on the next related open issue. I believe this will significantly reduce the hit-count.

rallytime (approved):

PR #48309, opened 2018-06-26
Drop support for Python 2 unicode string literals in YAML renderer

terminalmage:
In earlier releases, the below was considered valid usage in Python 2, assuming that `data` was a dictionary containing keys/values which are `unicode` types:

```yaml
/etc/foo.conf:
  file.managed:
    - source: salt://foo.conf.jinja
    - template: jinja
    - context:
        data: {{ data }}
```

Jinja will render the unicode string types in Python 2 with the "u" prefix (e.g. `{u'foo': u'bar'}`. While not valid YAML, earlier releases would successfully load these values.

This pull request drops this support, effective in the upcoming Fluorine release. Instead, it will now be necessary to use Jinja's ["tojson" filter](http://jinja.pocoo.org/docs/2.10/templates/#tojson).

```yaml
/etc/foo.conf:
  file.managed:
    - source: salt://foo.conf.jinja
    - template: jinja
    - context:
        data: {{ data|tojson }}
```

The ["tojson" filter](http://jinja.pocoo.org/docs/2.10/templates/#tojson) was added to Jinja in version 2.9, but several LTS releases (RHEL 7, Ubuntu 14.04, etc.) ship with earlier versions in their repositories. Therefore, this PR also adds an implementation of this filter, which will be used if Jinja doesn't already provide it (when it does, the upstream version of the filter will be used).

These changes allow Salt's custom YAML SafeLoader to use the libyaml versions of PyYAML's SafeLoader and SafeDumper (i.e. ``yaml.CSafeLoader`` and ``yaml.CSafeDumper``) if PyYAML was built with libyaml support.

Thanks to @jacksontj for working with me on this.

### DIFF
--- a/salt/utils/jinja.py
+++ b/salt/utils/jinja.py
@@ -309,6 +309,26 @@ def to_bool(val):
     return False
 
 
+@jinja_filter('tojson')
+def tojson(val, indent=None):
+    '''
+    Implementation of tojson filter (only present in Jinja 2.9 and later). If
+    Jinja 2.9 or later is installed, then the upstream version of this filter
+    will be used.
+    '''
+    options = {'ensure_ascii': True}
+    if indent is not None:
+        options['indent'] = indent
+    return (
+        salt.utils.json.dumps(
+            val, **options
+        ).replace('<', '\\u003c')
+         .replace('>', '\\u003e')
+         .replace('&', '\\u0026')
+         .replace("'", '\\u0027')
+    )
+
+
 @jinja_filter('quote')
 def quote(txt):
     '''
--- a/doc/topics/jinja/index.rst
+++ b/doc/topics/jinja/index.rst
@@ -886,7 +886,7 @@ Example:
     encoding (usually a ``unicode`` type). This filter was incorrectly-named
     when it was added. ``json_decode_list`` will be supported until the Neon
     release.
-.. deprecated:: Fluorine
+.. deprecated:: 2018.3.3,Fluorine
     The :jinja_ref:`tojson` filter accomplishes what this filter was designed
     to do, making this filter redundant.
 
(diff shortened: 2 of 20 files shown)


terminalmage:
My new integration test doesn't take into account dict iteration order differences. Derp. Fixing this as well as a lint violation, and there appear to be some other related failures.

salt/utils/jinja.py
```diff
@@ -309,6 +309,26 @@ def to_bool(val):
     return False
 
 
+@jinja_filter('tojson')
+def tojson(val, indent=None):
```
isbm: I would suggest to have a few of Unit Tests for this.

DmitryKuzmenko (approved):

isbm (changes requested):

salt/utils/jinja.py
terminalmage: I decided not to because A) Jinja 2.9+ has this filter built in, and B) the integration test I added already confirms that the content is dumped to JSON (before loading as YAML). I suppose I could add some tests just to be thorough.

terminalmage: OK, unit test added.

isbm (approved):
@terminalmage thanks :wink:

terminalmage:
I backported just the `tojson` filter to 2018.3 in https://github.com/saltstack/salt/pull/48339. I've updated the docs I wrote for this filter to reflect that.

This PR should be good to merge pending test results.

rallytime:
@terminalmage This is failing one test: https://jenkins.saltstack.com/job/PR/job/salt-pr-linode-cent7-py3/6075/

terminalmage:
@rallytime that should be fixed now. Apparently the jinja `sort` filter doesn't, you know, _sort_ anything.

rallytime (approved):

PR #48313, opened 2018-06-26
Add new netmiko and junos functions to the napalm module

mirceaulinic:
To gate functionality from the ``junos`` Execution Module, by reusing the
existing connection from NAPALM.
At the same time, add two more handy functions that leverage the connection via the Netmiko module by reusing the connection arguments: `send_commands` and `send_config`.

### DIFF
--- a/salt/modules/netmiko_mod.py
+++ b/salt/modules/netmiko_mod.py
@@ -493,6 +493,7 @@ def exit_config_mode(**kwargs):
 def send_config(config_file=None,
                 config_commands=None,
                 template_engine='jinja',
+                commit=False,
                 source_hash=None,
                 source_hash_name=None,
                 user=None,
@@ -538,6 +539,11 @@ def send_config(config_file=None,
         The template engine to use when rendering the source file. Default:
         ``jinja``. To simply fetch the file without attempting to render, set
         this argument to ``None``.
+
+    commit: ``False``
+        Commit the configuration changes before exiting the config mode. This
+        option is by default disabled, as many platforms don't have this
+        capability natively.
 
     source_hash
         The hash of the ``config_file``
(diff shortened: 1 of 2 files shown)


salt/modules/netmiko_mod.py
```diff
@@ -493,6 +493,7 @@ def exit_config_mode(**kwargs):
 def send_config(config_file=None,
                 config_commands=None,
                 template_engine='jinja',
+                commit=False,
```
rallytime: This looks like an API change -  can you add this to tend of the list, just in case?

salt/modules/netmiko_mod.py
```diff
@@ -501,7 +502,7 @@ def send_config(config_file=None,
                 attrs=None,
                 context=None,
                 defaults=None,
-                skip_verify=False,
+                skip_verify=True,
```
rallytime: Why is this default changing? That at least needs to be documented.

rallytime (changes requested):
A couple of comments/questions from me. :)

salt/modules/netmiko_mod.py
mirceaulinic: The netmiko module has been just added https://github.com/saltstack/salt/pull/48260 - there might be other upcoming changes if I notice anything misbehaving, sorry.

salt/modules/netmiko_mod.py
mirceaulinic: Indeed, it shouldn't. Reverted.

salt/modules/netmiko_mod.py
rallytime: Oh! I should have checked. Thanks! :)

rallytime (approved):

PR #48314, opened 2018-06-26
[ssh] support bootstrapping multiple remote nodes

### LINKED ISSUE #48224: please add a salt-ssh --bootstrap
Originally filed at https://bugzilla.opensuse.org/show_bug.cgi?id=1057772

### Setup
have a remote node without python installed

### Steps to Reproduce Issue
run `salt-ssh $NODENAME test.ping`

This will fail due to lack of python on the remote node.

### Versions Report
Salt: 2017.7.2

### Thoughts
One nice solution could be a new `salt-ssh --bootstrap $NODENAME`
that translates to something like
```bash
salt-ssh $NODENAME -r '
  if grep -q "SUSE" /etc/os-release ; then 
    zypper -n in python python-xml ;
  elif [ -e /etc/debian_version ] ; then 
    apt install ... ;
  else
    echo "please add support for your OS to salt-ssh bootstrap"
  fi'
```

bmwiedemann:
PoC patch to support bootstrapping multiple remote nodes
using salt-ssh raw mode

In case nodes do not have python we can only run raw commands
but we use OS-detection logic
to allow for diverse remote OSes to be bootstrapped in one simple command:
`salt-ssh --bootstrap '*'`

Fixes #48224

maybe it could be moved into salt/client/ssh/client.py
to be in a better place?

### DIFF
--- a/salt/cli/ssh.py
+++ b/salt/cli/ssh.py
@@ -16,6 +16,17 @@ class SaltSSH(salt.utils.parsers.SaltSSHOptionParser):
         if '-H' in sys.argv or '--hosts' in sys.argv:
             sys.argv += ['x', 'x']  # Hack: pass a mandatory two options
                                     # that won't be used anyways with -H or --hosts
+        if '--bootstrap' == sys.argv[1]:
+            sys.argv.remove('--bootstrap')
+            sys.argv += ['-r', '\
+                if grep -q "SUSE" /etc/os-release ; then\
+                    zypper -n in python python-xml ;\
+                elif [ -e /etc/debian_version ] ; then\
+                    apt --yes install python ;\
+                else\
+                    echo "please add support for your OS to salt-ssh bootstrap" ;\
+                    exit 6 ;\
+                fi']
         self.parse_args()
         self.setup_logfile_logger()
         verify_log(self.config)


gtmanfred (changes requested):
This feels really dirty.

I would rather have a separate process started to run through and install python, if we even consider going this route.

bmwiedemann:
'a separate process' like an exec to salt-ssh using the newly constructed argv?

bmwiedemann:
or I could make an extra salt-ssh-bootstrap script to construct the raw command

gtmanfred:
Yes, and or some other tool like using salt-run manage.bootstrap, that just manages logging in and running a script that manages all the distributions similar to how bootstrap.saltstack.com works.

salt/cli/ssh.py
```diff
@@ -16,6 +16,17 @@ def run(self):
...
+            sys.argv.remove('--bootstrap')
+            sys.argv += ['-r', '\
+                if grep -q "SUSE" /etc/os-release ; then\
+                    zypper -n in python python-xml ;\
+                elif [ -e /etc/debian_version ] ; then\
+                    apt --yes install python ;\
+                else\
+                    echo "please add support for your OS to salt-ssh bootstrap" ;\
+                    exit 6 ;\
+                fi']
```
isbm: I hate to say that, but no. In the [corresponding issue](https://github.com/saltstack/salt/issues/48224#issuecomment-400281487) I already explained that this will turn into a trimmed, poor's man `pkg.install` remake that will work only for that OS and probably only for that version.

That said, please write your own bootstrap script in whatever scripting language you are about: PowerShell for Windows, Korn Shell for Solaris, Bourne Again Shell for Linux, Shell for BSD etc.

stop

isbm (changes requested):

salt/cli/ssh.py
bmwiedemann: But then I'd have to get that bootstrap script onto a node first. Can salt-ssh copy files via raw mode, too?
Because state.apply wont work before python is installed.

isbm:
> 'a separate process' like an exec to salt-ssh using the newly constructed argv?

This is all what you would need to do:

```bash
#!/bin/bash

HOST="suma-refhead-srv.mgr.suse.de"
USER="root"

if [ "$(ssh $USER@$HOST 'which python 2>/dev/null')" == "" ]; then
    ssh $USER@$HOST 'zypper -n in python python-xml'
fi
```

salt/cli/ssh.py
isbm: Well, nothing will work at all, if no Python. I would just think it differently: build a script that would make sure your target nodes are requirements compliant. Say, you have 100 machines that needs to be operated with Salt SSH, so `cat list-of-machines` and `bootstrap &` with the script below I gave an example of.

bmwiedemann:
But that `ssh $USER@$HOST` would need to use `salt-ssh -r` to apply the settings (e.g. port) from the roster file
and it would only work for homogenous nodes, so no easy bootstrapping of Debian, Fedora and openSUSE nodes in one go.

isbm:
Then you just add that stuff like `if debian...` in _that_ script. Your approach here is that Salt should fit everywhere universally. My suggestion is to use it as a tool, one of those around, and "glue" it with Bash on your own.

It is also question which is your master machine? If it is Windows, likely all that bootstrap needs to be in PowerShell or something. Unlikely older Solaris will run on BASH either.

bmwiedemann:
I'd prefer if the admin does not need to find out that python-xml is needed, but let salt have that knowledge embedded *somewhere*. This PR might even work on a Windows without bash, because the shell code is run on the remote side only - likely even to work as is on Solaris, BSD, MacOSX (and of course Linux). Just because something only works on 9 of 10 target OSes does not mean it is useless.

Sure, it can be done differently and nicer and I'm all for it, but I object to having every admin reinvent this wheel.

isbm:
But this won't work if you are targeting Windows.

bmwiedemann:
yes.

isbm:
@bmwiedemann I think Salt SSH has documented [very clear requirements prior to go](https://docs.saltstack.com/en/latest/topics/ssh/#getting-started). And I don't see the reason of changing this. Python is almost everywhere by default in *nix systems nowadays. If it happens you have custom set of packages installed, then you probably should be tracking it on your own.

bmwiedemann:
The requirements just say "python is required" - but for openSUSE, the python interpreter binary is contained in the `python-base` package which is hard to uninstall anyway. So the `zypper in python` is actually about getting the **extra** python modules required for salt-ssh that I dont see documented there.

isbm:
@bmwiedemann what exactly packages are missing for you? Actually, Leap 15 minimal doesn't require anything special and Salt SSH just works there.

bmwiedemann:
It seems that on Leap-15/Tumbleweed, `zypper` recommends `zypper-log` which requires `python3` but when you build kiwi images in OBS, it usually ignores recommends to reduce the resulting image size (because few people actually know or use zypper-log).

OTOH, even when uninstalling python3, salt-ssh still works fine towards that node. Also without python*-xml ... so something seems to have changed there since I originally filed the bug. Then we can close this.

PR #48330, opened 2018-06-26
Add warnings plugin to replace violations

KaiSforza:
This plugin has no violations. There is **NO DOCUMENTATION** for this
entire plugin. And I still don't entirely understand how this being the
name of a step makes it work.

Writing these jenkinsfiles is making me go insane, I swear. 

@gtmanfred @dubb-b

### DIFF
--- a/.ci/lint
+++ b/.ci/lint
@@ -39,6 +39,16 @@ pipeline {
         }
     }
     post {
+        always {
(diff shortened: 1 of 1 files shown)


KaiSforza:
Note that the test I did had a threshold of 10, this one has the same as the old jenkins, at 999

gtmanfred (approved):

PR #48339, opened 2018-06-27
Backport tojson filter from #48309 to 2018.3

terminalmage:
This filter is a superior way to dump data structures into SLS and is less problematic than ``json_encode_dict``, which A) doesn't emit actual JSON, and B) emits strings as bytestrings, which would cause an error in Python 3.

### DIFF
--- a/doc/topics/jinja/index.rst
+++ b/doc/topics/jinja/index.rst
@@ -886,7 +886,7 @@ Example:
     encoding (usually a ``unicode`` type). This filter was incorrectly-named
     when it was added. ``json_decode_list`` will be supported until the Neon
     release.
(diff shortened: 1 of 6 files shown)


dijit:
Looks good, could use some more tests if this is replacing the current jinja JSON filter.

terminalmage:
@dijit It doesn't replace it, if the installed version of Jinja has a `tojson` filter then the upstream version of the filter is used. This is just implemented so people on LTS Linux distros can use the filter.

I'm curious what other tests you think could be added, though. The unit test I added was pretty straightforward.

rallytime (approved):

Issue #48169, opened 2018-06-17
win.powercfg documentation is incorrect 


adamsewell:
### Description of Issue/Question
The documentation located https://docs.saltstack.com/en/latest/ref/states/all/salt.states.win_powercfg.html states that the value for set_timeout is in minutes but it's actually in seconds.

Ch3LL:
ping @saltstack/team-windows can you confirm?

rallytime:
@adamsewell Can you give the fix in #48345 a try?

PR #48345, opened 2018-06-27
Fix behavior of powercfg module and state

twangboy:
### What does this PR do?
Improves documentation
Uses `'` instead of `"`
Uses minutes for the set function instead of seconds to be consistent with the documentation and the get function which returns everything in minutes.
Warn until Fluorine that the function expects seconds for settings as stated in the docs.
Adds the `test=True` option to the powercfg state module.

https://github.com/saltstack/salt/issues/48169

### Previous Behavior
The value that was passed was set directly without converting to seconds. For example, if the user was trying to set a timeout of 120 minutes it was actually setting 120 seconds.

### New Behavior
Now the value can be passed as minutes and it will be converted to seconds.

### Tests written?
Yes

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/modules/win_powercfg.py
+++ b/salt/modules/win_powercfg.py
@@ -7,8 +7,10 @@ powercfg.
 
 .. code-block:: bash
 
+    # Set monitor to never turn off
     salt '*' powercfg.set_monitor_timeout 0 power=dc
-    salt '*' powercfg.set_disk_timeout 120 power=ac
+    # Set disk timeout to 120 minutes
+    salt '*' powercfg.set_disk_timeout 7200 power=ac
 '''
 
 # Import Python Libs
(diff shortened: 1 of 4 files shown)


rallytime (approved):

rallytime:
@dwoz This look good to you?

salt/modules/win_powercfg.py
```diff
@@ -59,172 +64,309 @@ def _get_powercfg_minute_values(scheme, guid, subguid, safe_name):
...
+            'dc': int(raw_settings[1], 0) / 60}
 
 
 def _set_powercfg_value(scheme, sub_group, setting_guid, power, value):
     '''
-    Sets the value of a setting with a given power (ac/dc) to
-    the given scheme
+    Sets the AC/DC values of a setting with the given power for the given scheme
     '''
+    salt.utils.warn_until(
```
cachedout: Why would we change to something less exact? I'm not clear on the reasoning here. Would it be better just to change the documentation instead?

twangboy: Power settings in the GUI are displayed in Minutes and Hours. As a Windows admin, that's what I would expect to be setting from the CLI as well.

Since the return value is in Minutes and the documentation says minutes, I believe the original intent of the author was to set the time in Minutes.

cachedout (approved):

dwoz (approved):

PR #48356, opened 2018-06-28
Adding a pair of proxy and execution modules for managing Arista switches via the pyeapi

mirceaulinic:
### What does this PR do?

Adds the initial version of two new modules: one Proxy and one Execution module for managing Arista switches via [pyeapi](http://pyeapi.readthedocs.io/en/master/index.html) which is the official library from Arista, using the [eAPI](https://www.arista.com/assets/data/pdf/Whitepapers/Arista_eAPI_FINAL.pdf).

The execution module is flexible enough to execute the commands to the remote device without necessarily requiring to be running inside the `pyeapi` Proxy Minion. I named the modules `arista_pyeapi`, as I will provide a set of modules for managing Arista switches without having the `pyeapi` external dependency without necessarily having feature parity with `arista_pyeapi` (as that would imply rewriting a good part of the `pyeapi` library into Salt).

### DIFF
--- a/salt/modules/arista_pyeapi.py
+++ b/salt/modules/arista_pyeapi.py
@@ -1,0 +1,577 @@
+# -*- coding: utf-8 -*-
+'''
+Arista pyeapi
+=============
+
+.. versionadded:: Fluorine
+
+Execution module to interface the connection with Arista switches, connecting to
+the remote network device using the
+`pyeapi <http://pyeapi.readthedocs.io/en/master/index.html>`_ library. It is
+flexible enough to execute the commands both when running under an Arista Proxy
+Minion, as well as running under a Regular Minion by specifying the connection
+arguments, i.e., ``device_type``, ``host``, ``username``, ``password`` etc.
+
+:codeauthor: Mircea Ulinic <ping@mirceaulinic.net>
+:maturity:   new
+:depends:    pyeapi
+:platform:   unix
+
+.. note::
+
+    To understand how to correctly enable the eAPI on your switch, please check
+    https://eos.arista.com/arista-eapi-101/.
+
+Dependencies
+------------
+
+The ``pyeapi`` Execution module requires Netmiko to be installed:
+``pip install pyeapi``.
+
+Usage
+-----
+
+This module can equally be used via the :mod:`pyeapi <salt.proxy.arista_pyeapi>`
+Proxy module or directly from an arbitrary (Proxy) Minion that is running on a
+machine having access to the network device API, and the ``pyeapi`` library is
+installed.
+
+When running outside of the :mod:`pyeapi Proxy <salt.proxy.arista_pyeapi>`
+(i.e., from another Proxy Minion type, or regular Minion), the pyeapi connection
+arguments can be either specified from the CLI when executing the command, or
+in a configuration block under the ``pyeapi`` key in the configuration opts
+(i.e., (Proxy) Minion configuration file), or Pillar. The module supports these
+simultaneously. These fields are the exact same supported by the ``pyeapi``
+Proxy Module:
+
+transport: ``https``
+    Specifies the type of connection transport to use. Valid values for the
+    connection are ``socket``, ``http_local``, ``http``, and  ``https``.
+
+host: ``localhost``
+    The IP addres or DNS host name of the connection device.
+
+username: ``admin``
+    The username to pass to the device to authenticate the eAPI connection.
+
+password
+    The password to pass to the device to authenticate the eAPI connection.
+
+port
+    The TCP port of the endpoint for the eAPI connection. If this keyword is
+    not specified, the default value is automatically determined by the
+    transport type (``80`` for ``http``, or ``443`` for ``https``).
+
+enablepwd
+    The enable mode password if required by the destination node.
+
+Example (when not running in a ``pyeapi`` Proxy Minion):
+
+.. code-block:: yaml
+
+  pyeapi:
+    username: test
+    password: test
+
+In case the ``username`` and ``password`` are the same on any device you are
+targeting, the block above (besides other parameters specific to your
+environment you might need) should suffice to be able to execute commands from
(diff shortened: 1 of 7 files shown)


salt/modules/arista_pyeapi.py
```diff
@@ -0,0 +1,577 @@
...
+def _prepare_connection(**kwargs):
+    '''
+    Prepare the connection with the remote network device, and clean up the key
+    value pairs, removing the args used for the connection init.
+    '''
+    init_args = {}
+    fun_kwargs = {}
+    pyeapi_kwargs = __salt__['config.get']('pyeapi', {})
+    pyeapi_kwargs.update(kwargs)  # merge the CLI args with the opts/pillar
+    for karg, warg in six.iteritems(pyeapi_kwargs):
```
cachedout: We have nearly identical code over in netmiko_mod. Refactor for DRY?

salt/modules/arista_pyeapi.py
```diff
@@ -0,0 +1,577 @@
...
+    conn = pyeapi.client.connect(**init_args)
+    node = pyeapi.client.Node(conn, enablepwd=init_args.get('enablepwd'))
+    return node, fun_kwargs
+
+# -----------------------------------------------------------------------------
+# callable functions
+# -----------------------------------------------------------------------------
+
+
+def get_connection(**kwargs):
```
cachedout: Same here. Code is in the netmiko module. Refactor for DRY?

salt/modules/arista_pyeapi.py
mirceaulinic: Apologies for late reply @cachedout. Yes, there's a little overlap between them. We could have the following function in the `utils`:

```python
def prepare_kwargs(all_kwargs, class_init_kwargs):
    for karg, warg in six.iteritems(all_kwargs):
        if karg not in class_init_kwargs:
            if warg is not None:
                fun_kwargs[karg] = warg
            continue
        if warg is not None:
            init_args[karg] = warg
    return init_args, fun_kwargs
```

Then the code in this execution module would become:

```python
def _prepare_connection(**kwargs):
    '''
    Prepare the connection with the remote network device, and clean up the key
    value pairs, removing the args used for the connection init.
    '''
    init_args = {}
    fun_kwargs = {}
    pyeapi_kwargs = __salt__['config.get']('pyeapi', {})
    pyeapi_kwargs.update(kwargs)  # merge the CLI args with the opts/pillar
    init_args, fun_kwargs = salt.utils.args.prepare_kwargs(pyeapi_kwargs, PYEAPI_INIT_KWARGS)
    if 'transport' not in init_args:
        init_args['transport'] = 'https'
    conn = pyeapi.client.connect(**init_args)
    node = pyeapi.client.Node(conn, enablepwd=init_args.get('enablepwd'))
    return node, fun_kwargs
```

I'm not disagreeing, just that I'm not sure if it's really worth it, particularly as it's harder to port a module in an older Salt version due to the utils dependency.
On the other hand, I have a couple of more modules pending in the queued to be submitted, that would have the same code as well.

What do you think would be the best? Thanks!

gtmanfred: If you use `__utils__['args.prepare_kwargs']` they should just have to drop the newer args module into `salt://_utils/` and sync them and it should be easily portable.

mirceaulinic: Fantastic, I will do it like that then - will refactor and push tomorrow morning. Thanks guys!

mirceaulinic:
Hi @cachedout - I have pushed https://github.com/saltstack/salt/pull/48356/commits/dddef4996f45c9cbc4c399a428142e96309bd181 to remove the redundant code. I will follow up with another PR for Netmiko to use the new `args.prepare_kwargs` util function.

CC. @gtmanfred @rallytime

gtmanfred (approved):

salt/modules/arista_pyeapi.py
```diff
@@ -0,0 +1,718 @@
...
+    salt '*' pyeapi.send_commands 'show version' 'show interfaces'
+    salt '*' pyeapi.config 'ntp server 1.2.3.4'
+
+.. note::
+
+    Remember that the above applies only when not running in a ``pyeapi`` Proxy
+    Minion. If you want to use the :mod:`pyeapi Proxy <salt.proxy.arista_pyeapi>`,
+    please follow the documentation notes for a proper setup.
+'''
+from __future__ import absolute_import
```
rallytime: We also need to import `print_function` and `unicode_literals` here.

salt/proxy/arista_pyeapi.py
```diff
@@ -0,0 +1,178 @@
...
+
+.. code-block:: yaml
+
+    proxy:
+      proxytype: pyeapi
+      host: router1.example.com
+      username: example
+      password: example
+'''
+from __future__ import absolute_import
```
rallytime: Same comment here.

rallytime (commented):
I have two small changes that are needed.

cachedout (approved):

mirceaulinic:
Okay, added https://github.com/saltstack/salt/pull/48356/commits/2b61f5d1a47ced980f6ce23d856e1b169a6b5dc4 to solve the imports.

mirceaulinic:
Bump @cachedout @rallytime - apologies if I'm pushy, but I'd love to have this in so I can follow up with the next PR which depends on this (of course, if you are happy with the code), so I can finish before the 6th of Aug. Thanks!

rallytime (approved):

rallytime:
@mirceaulinic It's merged now. I was off yesterday so I hadn't see then notification yet. :)

Thanks for working on all of this for the feature release! We appreciate it, as always. :)

One thing - I know you've been adding a bunch of new modules - can you make sure you get those listed in the `Fluorine` release notes?

mirceaulinic:
Yes, thanks for reminder @rallytime. I will surely add them to the release notes too!

Issue #48320, opened 2018-06-26
m2crypto python3 corrupt public key


Ch3LL:
### Description of Issue/Question
When using m2crypto python3 alongside >=2018.3.1 python3 versions I see this error when accepting the minion key: 

```
2018-06-26 16:15:19,369 [salt.transport.mixins.auth:438 ][ERROR   ][9189] Corrupt public key "/etc/salt/pki/master/minions/qapkgtestrs-redhat7qavk": Can't convert 'bytes' object to str implicitly
```

And then going forward I cannot run any commands against that minion. 

```
[root@qapkgtestrs-redhat7qavk ~]# salt '*' test.ping
qapkgtestrs-redhat7qavk:
    Minion did not return. [No response]
```

### Steps to Reproduce Issue
1. install m2crypto: `yum install python34-m2crypto`
2. install python3 salt-master/salt-minion
3. point minion to master.
4. accept key `salt-key -a <keyname> -y`
5. look at logs on master: 

```
[root@qapkgtestrs-redhat7qavk ~]# grep -i corrup /var/log/salt/master 
2018-06-26 16:15:19,369 [salt.transport.mixins.auth:438 ][ERROR   ][9189] Corrupt public key "/etc/salt/pki/master/minions/qapkgtestrs-redhat7qavk": Can't convert 'bytes' object to str implicitly
```

This was due to this change https://github.com/saltstack/salt/pull/46930

You can also quickly replicate it by running this script:

```
import salt.utils.files
from M2Crypto import RSA, EVP, BIO
path = '/etc/salt/pki/master/minions/qapkgtestrs-redhat7qavk'
with salt.utils.files.fopen(path) as f:
    data = f.read().replace(b'RSA ', '')
bio = BIO.MemoryBuffer(data)
key = RSA.load_pub_key_bio(bio)
print(key)
```

Which grabs code from salt/crypt.py here: https://github.com/saltstack/salt/blob/v2018.3.1/salt/crypt.py#L208-L212

currently working on fix

### Versions Report

```
[root@qapkgtestrs-redhat7qavk ~]# salt --versions-report
Salt Version:
           Salt: 2018.3.1

Dependency Versions:
           cffi: Not Installed
       cherrypy: unknown
       dateutil: Not Installed
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.8
        libgit2: Not Installed
        libnacl: Not Installed
[...]

PR #48358, opened 2018-06-28
Fix corrupt public key with m2crypto python3

Ch3LL:
### What does this PR do?
This ensures we are using byte data with m2crypto which it requires. Also fixes some previous tests on python3 to ensure the SIGNATURE is byte data as well.

https://github.com/saltstack/salt/issues/48320

### Previous Behavior
could not communicate with minion because of a corrupt public key error.

### New Behavior
can communicate with minion and key is accepted without error

### Tests written?

Yes there are already tests written but we need to add a new jenkins job which install m2crypto and run them. Currently they are being skipped and we are only testing pycrypto which is the default crypto library that we use. Will be working on this to ensure we are testing both m2crypto and pycryptodomex in our auto test suite. I ran the unit.test_crypt tests with this fix for both python2/python3 but we will definitely want to get these tests running before next release.

### Commits signed with GPG?

Yes


dwoz (approved):

dwoz:
We can't use `salt.utils.fopen('foo', encoding='utf-8')` in python 2 because the underlying `open` call does not support the `encoding` keyword argument. There was a discussion in slack about migrating to `io.open` which does support the `encoding` argument.  @terminalmage I think this is a case that supports the idea of moving to `io.open`, Thoughts?

salt/crypt.py
```diff
@@ -207,7 +207,7 @@ def get_rsa_pub_key(path):
     log.debug('salt.crypt.get_rsa_pub_key: Loading public key')
     if HAS_M2:
         with salt.utils.files.fopen(path) as f:
-            data = f.read().replace(b'RSA ', '')
+            data = salt.utils.stringutils.to_bytes(f.read()).replace(b'RSA ', b'')
```
DmitryKuzmenko: Isn't it better to read the file in binary mode than converting it to bytes?
```python
    with salt.utils.files.fopen(path, 'rb') as f:
        data = f.read().replace(b'RSA ', b'')
```

dwoz (approved):

DmitryKuzmenko (approved):

Issue #47732, opened 2018-05-18
RFC: Standard for determining if a remote-execution function failed


terminalmage:
# Problem
While Salt does exit with a non-zero status when states fail, it does not have a reliable way of reporting that a remote-execution function failed. It is not feasible to determine this simply by a boolean evaluation of the function's return data, as many functions are informational in nature (e.g. `service.running`, `pkg.version`, etc.) and a return value that evaluates as `False` doesn't necessarily mean that the function failed.

# Proposed Solution

The minion is already setting the retcode for states based on `__context__['retcode']`. In minion.py, this value is checked [here](https://github.com/saltstack/salt/blob/ffe507d/salt/minion.py#L1628-L1631) and the job's return code is set accordingly.

Therefore, the following is proposed to standardize return code handling for remote-execution functions:

1. In [this exception handling code](https://github.com/saltstack/salt/blob/ffe507d/salt/minion.py#L1633-L1668), set `ret['retcode']` to a non-zero integer value in each `except` block. This will get us a non-zero exit status for free, for any remote-execution functions which raise an exception, and we already have a great deal of exception raising already built into Salt code for various error conditions.

2. Modify existing code to raise exceptions when the result of a function is a failure. **NOTE: This will not be done. The PR opened to address this will rely on either `__context__['retcode']` or for the return data to be a dictionary with either a `result` or `success` key, in cases where no exception was raised.**

   Note that there are some remote execution functions that simply return the result of a `cmd.run_all`, or even a state return dictionary (i.e. one with `name`, `result`, `changes`, and `comment` keys). 

   For cases where a `cmd.run_all` dictionary is returned, we can simply raise a `CommandExecutionError('some error msg', info=cmd_result)`, where `cmd_result` is the name of the dict we'd normally be returning. This may require that we do exception catching in any state which invokes that remote-execution function. When doing so, we would be able to obtain the original `cmd.run_all` dict from exception's `info` attribute:

   ```python
   try:
       result = __salt__['some_mod.some_func'](name, foo, bar, **kwargs)
   except CommandExecutionError as exc:
       # do some stuff here to gracefully return a False result for the state.
       # exc.info will contain the original cmd.run_all return dict
   ```

   For cases where the remote-execution function returns a state-like return dict, we could either take the same tactic as above, or we could simply set `__context__['retcode']` before returning.

terminalmage:
ping @saltstack/team-suse

mattp-:
[...]

PR #48361, opened 2018-06-28
Improve retcode reporting in salt/salt-call CLI commands

terminalmage:
## What this PR does

Improves how the `salt` and `salt-call` CLI commands set exit codes.

Refs: https://github.com/saltstack/salt/issues/47732

### For `salt`:

The following cases will result in a nonzero exit status for `salt`

- Any exception raised by the minion function will result in `salt`
- If the return data is a dictionary and contains either a `result` or `success` key, and its value is `False`

### For `salt-call`:

The following cases will result in a nonzero exit status for `salt-call` (assuming it was invoked using `--retcode-passthrough`):

- If the return data is a dictionary and contains either a `result` or `success` key, and its value is `False`

## Further points of discussion needed before merge

There is some asymmetry to how `salt` and `salt-call` handle exit codes:

For `salt`, when the master sees that one of its minions has set a nonzero retcode in the return data, it does a `sys.exit(11)`. When `salt-call` is invoked with `--retcode-passthrough`, it will return whatever retcode was set in `__context__['retcode']`, otherwise it just returns a `0` exit status (assuming no exception was raised).

For `salt`, when a minion catches an exception, it sets a retcode in the return event sent back to the master. As noted above, any one minion that generated a nonzero retcode will result in the master exiting with a return code of `11`. However, for `salt-call`, any exception raised just results in a `sys.exit(salt.defaults.exitcodes.EX_GENERIC)` (wth that retcode being `1`).

To represent this all in a table:

| Result                               | `salt` exit code | `salt-call` exit code | `salt-call --retcode-passthrough` exit code  |
| ------------------------------------ |:----------------:|:---------------------:|:--------------------------------------------:|
| Nonzero retcode set in `__context__` | 11               | 0                     | Whatever was set in `__context__['retcode']` |
| Exception raised                     | 11               | 1                     | 1                                            |
| `result` or `success` is `False`     | 11               | 0                     | 1                                            |

I feel like we should normalize this somehow. Here are my ideas:

- `--retcode-passthrough` should be deprecated and salt-call's default behavior should be to return a nonzero retcode when there is a failure case
- The same retcode should be returned in both `salt` and `salt-call` when an exception is raised by a minion function

This is somewhat complicated by the fact that `salt` has to set a CLI retcode based on potentially multiple minions having failed, while `salt-call` only has to set a CLI retcode based on one host's result. So perhaps _some_ asymmetry is called for. I think it would be worth creating a new exit code in `salt.defaults.exitcodes` which represents a caught exception, and then exiting with that code in both `salt` and `salt-call`. This gets around the fact that when running `salt` from the master, the master can't necessarily tell that the minion raised an exception or whether that retcode was set in `__context__['retcode']`. Alternatively, rather than trying to manage this simply via the retcode, it may be a good idea to set an optional key in the return event payload the minion sends back to the master, so that the master knows uneqivocally that an exception was caught. After all, if we just managed this using the `retcode` key in the return event payload, then one could just set `__context__['retcode']` to that number and it would fool the `salt` CLI command into thinking that an exception had been raised.

I'm less sure what to do when `__context__['retcode']` is set to a nonzero value. What if there are multiple distinct retcodes that come back to the master for different minions? Which does the `salt` CLI command use for its exit code?


damon-atkins:
Consider what I did with windows `pkg.refresh_db failhard=False --out json`   which was to output result as data even if it failed and the user is responsible for checking the data return with some other program.  The problem with raise CommandExecutionError etc is the output is no longer a valid data structure.

```
    if error_count > 0 and failhard:
        raise CommandExecutionError(
            'Error occurred while generating repo db',
            info=results
        )
    else:
        return results   # user needs to check the data returned for the error information.

```


```
$ salt -G os:Windows pkg.refresh_db failhard=False --out=json 2> /dev/null
{
    "testminion": {
        "failed": 2,
        "total": 242,
        "failed_list": {
            "dvscheduler2.sls": [
                "package 'dvschedulerX', version number 1 is not a string",
                "package 'dvschedulerX', version number 2 is not a string"
            ],
            "bad-file.sls": [
                "Compiled contents, not a dictionary/hash "
            ]
        },
        "success": 240
    }
}
{
    "downminion": "Minion did not return. [Not connected]"
}
```

```
$ salt -G os:Windows pkg.refresh_db failhard=True --out=json 2> /dev/null
{
    "testminion": "ERROR: Error occurred while generating repo db. Additional info follows:\n\nfailed:\n    2\nfailed_list:\n    ----------\n    bad-file.sls:\n        - Compiled contents, not a dictionary/hash \n    dvscheduler2.sls:\n        - package 'dvschedulerX', version number 1 is not a string\n        - package 'dvschedulerX', version number 2 is not a string\nsuccess:\n    240\ntotal:\n    242"
}
{
    "downminion": "Minion did not return. [Not connected]"
}
```

An alternative might be to change the structure of the output e.g.
```
salt \* --errors-as-data   state.apply stuff.sls
{
minion:
    data:   <normal output>
    errors: <list>
}
```

DmitryKuzmenko (approved):

cachedout:
This is great. Let me respond to few of your comments as best I can:

```Which does the salt CLI command use for its exit code?```

Frankly, I think we can be somewhat arbitrary about this. We could, for example, simply agree to return the highest integer. The more important notion is that it's non-zero and there's not a ton that we can do beyond that.

```--retcode-passthrough should be deprecated and salt-call's default behavior should be to return a nonzero retcode when there is a failure case```

I have never liked this flag. I think it's non-intuitive and I've always felt like salt-call should just "do the right thing" by default which would be to pass through the return code of the called function. That said, I'm pretty skittish about this change to behavior. We'd be talking about something that might bite a lot of people. I'd want to discuss how we'd communicate this impending change. That said, I do think that as much parity as we can achieve between `salt` and `salt-call` is a very worthy goal.

File: salt/minion.py [hunk 1]
```diff
@@ -1807,20 +1826,26 @@ def _thread_multi_return(cls, minion_instance, opts, data):
...
+                    # No nonzero retcode in __context__ dunder. Check if return
+                    # is a dictionary with a "result" or "success" key.
+                    try:
+                        func_result = all(ret['return'][key].get(x, True)
+                                          for x in ('result', 'success'))
+                    except Exception:
+                        # return data is not a dict
+                        func_result = True
+                    if not func_result:
+                        retcode = 1
```
cachedout: We've got a lot of things setting retcodes of one. Worth doing some debug logging here so at least the reason could be inspected?

salt/modules/test.py
```diff
@@ -628,3 +630,44 @@ def false_():
...
+
+    try:
+        if name.startswith('salt.exceptions.'):
+            exc = getattr(salt.exceptions, name[16:])
+        else:
+            exc = getattr(builtins, name)
+        if _is_exc(exc):
+            raise exc(*args, **salt.utils.args.clean_kwargs(**kwargs))
+        else:
+            log.error('%s is not an exception', name)
```
cachedout: This is not the most intuitive error message, IMHO. As a user, I wouldn't know what to do if I saw this.

salt/modules/test.py
```diff
@@ -628,3 +630,44 @@ def false_():
...
+            return False
+    except AttributeError:
+        log.error('No such exception: %s', name)
```
cachedout: Same as above.

isbm:
@terminalmage Right direction! Since the exit codes are always very important to any Unix scripting, I  would though say my few cents here:

- `salt-call` should not return `1` and `0` for the same thing. If it is `success`, then it is `0`. But then it is not, it should be something else (TBD).
- `sys.exit(11)` is still very odd and we should change that. POSIX it means "Try Again" (?), Salt Exit Codes it means "Thin redeployment failure" (??). I would propose to return an actual code, or at least stick to a generic failure and success.
- We probably need to either change exit codes normalising for every command or at least segregate one for Salt SSH and other for the rest.
- I would concentrate of _what_ is returned, not _how_ (though it is also very important). Because, again, if one gets `0` for `False` and `True`, then the whole thing makes no much sense.

isbm (commented):
@terminalmage Right direction! Since the exit codes are always very important to any Unix scripting, I would though say my few cents here:

- salt-call should not return 1 and 0 for the same thing. If it is success, then it is 0. But then it is not, it should be something else (TBD).
- sys.exit(11) is still very odd and we should change that. POSIX it means "Try Again" (?), Salt Exit Codes it means "Thin redeployment failure" (??). I would propose to return an actual code, or at least stick to a generic failure and success.
- We probably need to either change exit codes normalising for every command or at least segregate one for Salt SSH and other for the rest.
- I would concentrate of what is returned, not how (though it is also very important). Because, again, if one gets 0 for False and True, then the whole thing makes no much sense.
- Separate "CLI error codes" from "Salt call error codes". That is, whatever mechanism is chosen to determine what kind of "False" with whatever internal error code Minion just returned in the actual result inside JSON or so, to CLI it should be POSIX way.

I am still puzzled why we are overriding POSIX codes, instead of just allocating new ones? I.e. why we are overriding 11 with "Thin wrong deployment" instead of allocating 211 for it?

damon-atkins:
FYI
http://pubs.opengroup.org/onlinepubs/9699919799//utilities/V3_chap02.html#tag_18_08_02
http://tldp.org/LDP/abs/html/exitcodes.html

terminalmage:
My vote would be for a generic `1` for failure in most cases

@isbm what are your thoughts for signaling an exception raised on a remote minion? I personally like the idea of having an optional key in the return event payload rather than setting a `retcode` key in the return payload.

Perhaps `--retcode-passthrough` is worth keeping as-is. That way we can keep with the standardized retcode by default, and still let people keep the retcode from the context dunder if they want.

salt/modules/test.py
terminalmage: It would be accompanied by a line number in the log file. This is a function designed to be used for testing, so I didn't think we needed something that was all that verbose. The idea is that if you are running this function, you're doing it for a reason. I don't think it's something that 99.99% of non-developer users are ever going to see.

What did you have in mind?

salt/minion.py
terminalmage: Yes, that can be arranged. The `1`s I added are kind of just placeholders right now though, I would prefer to eventually be using constants defined in salt.defaults.exitcodes.

damon-atkins:
The other option is to assign an error to a bit from 0 to 255 there are 8 bits. Therefore 8 different types of errors.  If two different errors occur xor them together.
```
1 General Error
2 No Response from some minions
4 Some states failed
8 ?
16 ?
32 ?
64 ?
128 ?
```
So 6 would mean some minions did not respond and some states failed. Then a person could care about States failing and not care about minion not responding in their shell script which determines to alarm  or not based on the exit code.

damon-atkins:
Not sure if 20 plus errors codes are the way to go.

We need to known for `salt \* something` if:
* everything worked (0)
* we need to know if we should try again e.g. minion did not respond. (2)
* we need to know if something failed and we need to review the output. (4)
Anything else is a bit more details along the same lines.

Also people like to use cmd.run to get real error code results as well, which might be an exception to the rule which only makes sense for an action against a single server.  Maybe a new command called `salt-cmd <stdout_file_name> <stdin_file_name> <return_code_filename> <single minion only>  cmd.*  `

terminalmage:
I don't think we need a large number of error codes. At most we would need 2, in my opinion. As I see it, here are two good options:

- All failures receive the same error code

- Failures which result from something other than an exception get one error code, while any failures set via `__context__['retcode']` (which denote a job that ran to completion but failed without raising an uncaught exception) would get a different error code.

damon-atkins:
I am suggesting an extra command e.g. `salt-cmd` as ssh returns an error code of the remote command.   However the way ssh is implemented its hard to tell if ssh had the problem or remote command has the problem. Hence the
`salt-cmd <stdout_file_name> <stdin_file_name> <return_code_filename> <single minion only> cmd.*`
which allows the issues to be separated.

Which will give people the scripted remote execution they want, without impacting `salt \*` standard behaviour.

terminalmage:
After talking with @thatch45 earlier today on this, this is what we've decided:

- _All_ error cases will use `EX_GENERIC` (i.e. `1`).
- `--retcode-passthrough` will be left untouched, but salt-call will still exit with `1` when there is an error condition and `--retcode-passthrough` is not used. Essentially, the only function that `--retcode-passthrough` has will be to return what is in `__context__['retcode']`.

thatch45:
My name is @thatch45 and I endorse this conclusion.
:)

terminalmage:
OK, adjustments have been made to error handling, including adjusting some of the integration tests I added and adding a couple new ones. The updated behavior should be as follows:

| Result                               | `salt` exit code | `salt-call` exit code | `salt-call --retcode-passthrough` exit code  |
| ------------------------------------ |:----------------:|:---------------------:|:--------------------------------------------:|
| Nonzero retcode set in `__context__` | 1                | 1                     | Whatever was set in `__context__['retcode']` |
| Exception raised                     | 1                | 1                     | 1                                            |
| `result` or `success` is `False`     | 1                | 1                     | 1                                            |

salt/modules/test.py
cachedout: Ah, I failed to recognize that this was in the test module. Yes, your counter-point is sound.

salt/modules/test.py
cachedout: Issue resolved via above comment.

cachedout:
IMHO we absolutely need a page in the docs dedicated to explaining this added before we merge it.

rallytime:
@cachedout How does this look to you now?

cachedout (approved):

PR #48371, opened 2018-06-29
mongo_future_return: MongoDb Connection string support

jyurdal:
### What does this PR do?

PR adds MongoDB connection string support (URI), what enables using replica set and other connection options. 
Those features are already supported in pymongo so this PR only adds URI config and passes it to pymongo. 

### Tests written?

No

### Commits signed with GPG?

No

### DIFF
--- a/salt/returners/mongo_future_return.py
+++ b/salt/returners/mongo_future_return.py
@@ -176,8 +176,6 @@ def _get_conn(ret):
         if uri and host:
             raise salt.exceptions.SaltConfigurationError(
                     "Mongo returner expects either uri or host configuration. Both were provided")
-
-
         pymongo.uri_parser.parse_uri(uri)
         conn = pymongo.MongoClient(uri)
         mdb = conn.get_database()
--- a/salt/returners/mongo_future_return.py
+++ b/salt/returners/mongo_future_return.py
@@ -5,9 +5,10 @@ Return data to a mongodb server
 Required python modules: pymongo
 
 
-This returner will send data from the minions to a MongoDB server. To
-configure the settings for your MongoDB server, add the following lines
-to the minion config files:
+This returner will send data from the minions to a MongoDB server. MongoDB
+server can be configured by using host, port, db, user and password settings
+or by URI (for pymongo > 2.3). To configure the settings for your MongoDB
+server, add the following lines to the minion config files:
 
 .. code-block:: yaml
 
@@ -16,6 +17,12 @@ to the minion config files:
     mongo.user: <MongoDB username>
     mongo.password: <MongoDB user password>
     mongo.port: 27017
+
+Or single URI:
+
+.. code-block:: yaml
+
+   mongo.uri: URI
 
 You can also ask for indexes creation on the most common used fields, which
 should greatly improve performance. Indexes are not created by default.
@@ -36,6 +43,11 @@ the default location:
     alternative.mongo.password: <MongoDB user password>
     alternative.mongo.port: 27017
 
+Or single URI:
+
+.. code-block:: yaml
+
+   alternative.mongo.uri: URI
 
 This mongo returner is being developed to replace the default mongodb returner
 in the future and should not be considered API stable yet.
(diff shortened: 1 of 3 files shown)


salt/returners/mongo_future_return.py
```diff
@@ -141,15 +155,22 @@ def _get_conn(ret):
...
+        if uri and host:
+            warnlog = 'Specified both URI and host. URI {0} will be taken'.format(uri)
+            log.warn(warnlog)
+        pymongo.uri_parser.parse_uri(uri)
+        conn = pymongo.MongoClient(uri)
+        mdb = conn.get_database()
     else:
-        conn = pymongo.Connection(host, port)
-    mdb = conn[db_]
+        if float(version) > 2.3:
```
rares-pop: would it make sense to use salt.utils.versions.LooseVersion instead of float(), to be uniform with other version checks throughout salt's codebase?

that's valid for both mongo_future_return.py and mongo_returner.py

jyurdal: I can change it to the LooseVersion.

I did not want to change code not related to the URI feature but I could imagine installing salt 2018 requires lots of up to date python modules so it could be the time to remove support for pymongo < 2.3
In that case port, host and other configs could be removed and only connection string would be supported ( you can easily put all previously used options into connection string format)
If you and other salt contributors think it is OK, I can make the changes.

rares-pop: I'm not that familiar with mongodb and I don't know the users' existing code-base, so I can't recommend anything. Maybe is okay to deprecate it (a sudden removal seems.. sudden).

I thought it's a low-hanging-fruit and seems among the last places where salt has a custom version compare. I think it can easily be done independent of this change though.

I'm not an official reviewer, so it's not in my power to approve things.

jyurdal: Good idea, so all in all:
- depreciate host and port configuration ( but still support it)
- error on both uri and host
- any comments on removing 2.3 support anyone?

salt/returners/mongo_future_return.py
```diff
@@ -17,6 +18,12 @@
     mongo.password: <MongoDB user password>
     mongo.port: 27017
 
+Or single URI:
+
+.. code-block:: yaml
+
+   mongo.uri: URI
```
gtmanfred: Please provide an example of the uri format.

Thanks,
Daniel

salt/returners/mongo_future_return.py
gtmanfred: Yes, lets use LooseVersion.

We shouldn't remove support for older versions if we can help it.  And I would say we shouldn't remove support for older configurations when it isn't absolutely necessary.

salt/returners/mongo_future_return.py
```diff
@@ -141,15 +155,22 @@ def _get_conn(ret):
...
+            conn = pymongo.Connection(host, port)
 
-    if user and password:
-        mdb.authenticate(user, password)
+        mdb = conn[db_]
+        if user and password:
+            mdb.authenticate(user, password)
 
     if indexes:
         if float(version) > 2.3:
```
gtmanfred: If you wouldn't mind, it would be great to move all these other version comparisons to using LooseVersion as well.

gtmanfred (changes requested):
Just a couple changes.

Thanks!
Daniel

salt/returners/mongo_future_return.py
jyurdal: Done

jyurdal: I have updated doc, and changed LooseVersion in mongo_future_return and mongo_return. I did some more extra handling so that only uri or host can be specified in the config file.

gtmanfred: perfect, thanks!

gtmanfred (approved):
Thanks!

rallytime (approved):
Nice!

PR #48373, opened 2018-06-29
Additional (and previously missing) events for the VMWare cloud provider

ghost:
### What does this PR do?

Adds events `salt/cloud/../querying` and `salt/cloud/../waiting_for_ssh` in the VMWare cloud provider and fixes a few pylint warnings.

Please pardon the utter laziness which resulted in putting two completely unrelated things into the same patch.

None.

### Previous Behavior

SaltCloud VMWare provider didn't send (at all) events `salt/cloud/../querying` and `salt/cloud/../waiting_for_ssh`. The former might not of much use, but the latter, I believe, is definitely worth a patch.

### New Behavior

SaltCloud VMWare provider sends events `salt/cloud/../querying` and `salt/cloud/../waiting_for_ssh`.

### Tests written?

No

### Commits signed with GPG?

No

### DIFF
--- a/salt/cloud/clouds/vmware.py
+++ b/salt/cloud/clouds/vmware.py
@@ -224,6 +224,9 @@ def script(vm_):
 
 
 def _str_to_bool(var):
+    '''
+    Safely convert a string representation of a boolean to a real boolean.
+    '''
     if isinstance(var, bool):
         return var
 
@@ -235,7 +238,9 @@ def _str_to_bool(var):
 
 def _get_si():
     '''
-    Authenticate with vCenter server and return service instance object.
+    Authenticate with vCenter server and acquire a service instance object.
+
+    :returns: A service instance object.
     '''
 
     url = config.get_cloud_config_value(
@@ -621,6 +626,9 @@ def _add_new_cd_or_dvd_drive_helper(drive_label,
                                     device_type,
                                     mode,
                                     iso_path):
+    '''
+    Generate a new vim.vm.device.VirtualCDrom object with desired properties.
+    '''
     random_key = randint(-3025, -3000)
 
     device_type.strip().lower()
@@ -670,9 +678,11 @@ def _set_network_adapter_mapping(adapter_specs):
         gateway = adapter_specs['gateway']
         adapter_mapping.adapter.gateway = gateway
     if 'ip' in list(adapter_specs.keys()):
-        ip = six.text_type(adapter_specs['ip'])
+        ip_address = six.text_type(adapter_specs['ip'])
         subnet_mask = six.text_type(adapter_specs['subnet_mask'])
-        adapter_mapping.adapter.ip = vim.vm.customization.FixedIp(ipAddress=ip)
+        adapter_mapping.adapter.ip = vim.vm.customization.FixedIp(
+                                          ipAddress=ip_address
+                                     )
         adapter_mapping.adapter.subnetMask = subnet_mask
     else:
         adapter_mapping.adapter.ip = vim.vm.customization.DhcpIpGenerator()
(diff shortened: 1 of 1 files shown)


salt/cloud/clouds/vmware.py
```diff
@@ -1448,7 +1524,7 @@ def _get_hba_type(hba_type):
     raise ValueError('Unknown Host Bus Adapter Type')
 
 
-def test_vcenter_connection(kwargs=None, call=None):
```
gtmanfred: Why are we removing kwargs?

These are required for calling stuff from the commandline using `salt-cloud -f` or `salt-cloud -a`

gtmanfred (changes requested):
we need to leave kwargs on functions as they are required for using `salt-cloud -a/-f`

Also, there is a lot of extra clean up in here that isn't needed as the code already meets our pylint style guidelines.

salt/cloud/clouds/vmware.py
samilaine: Because kwargs aren't used in the function itself. I oscillated between putting pylint hints in place and removing kwargs when they're not used *in* a function, eventually settling with the latter for clarity.

> the code already meets our pylint style guidelines

Would you care to explain why running pylint with the `.pylintrc` (https://github.com/saltstack/salt/blob/develop/.pylintrc) in accompanying the SaltStack code spits out warnings and errors for things corrected? If the SaltStack Coding Style is not completely erroneous, I believe the pylint fixes introduced are in line with the coding standards and the code didn't meet the pylint style guidelines.

gtmanfred: We use .testing.pylintrc for our pylint tests, and it is currently passing on this branch.

https://jenkinsci.saltstack.com/job/develop/view/Lint/job/salt-develop-lint/

samilaine: In that case I suggest the relevant documentation would be amended to reflect that fact. Currently the documentation claims: "SaltStack recommends running the .pylintrc file on any files you are changing with your code contribution before submitting a pull request to Salt's repository."

Mentioning that against the branch "development" a separate pylintrc is being used would help to avoid misunderstandings in the future.

... that said, I still stand by my fabulous argument about not putting unused arguments in the function fingerprints (obviously I did not remove kwargs or any other parameters from functions which actually use it).

cachedout: @samilaine is right on here about the error on our docs. That's been there for years and I keep forgetting that it needs to be cleaned up.

cachedout:
Paging @gtmanfred to come back by here and finish the discussion, please. :)

salt/cloud/clouds/vmware.py
gtmanfred: Yeah, that needs to be cleaned up, over the last couple weeks, we have moved pylint to use `tox` so to use the same lint command that we are using, you can run `tox -e pylint-salt,pylint-tests`  You can also pass a noops operator, and pass what files you want to lint afterwards, and it should only test those files.

```
tox -e pylint-salt -- salt/cloud/clouds/vmware.py
```

As for removing kwargs, that still needs to be in there because of the way that we call those functions when we call them inside of the CloudClient (if someone passes in kwargs) when calling functions and actions.

https://github.com/saltstack/salt/blob/develop/salt/cloud/__init__.py#L1610
https://github.com/saltstack/salt/blob/develop/salt/cloud/__init__.py#L1534

rallytime:
@samilaine I think it's your turn to respond to the feedback/comments above, when you get a chance. :)

cachedout:
@samilaine I think we just need the kwargs concern addressed and then we can move forward here.

cachedout:
@samilaine Since we have not heard from you, I have to go ahead and close this. I agree with @gtmanfred on the need for `kwargs` in that function. If we can get that issue resolved, we can certainly reopen this and continue. Thanks.

Issue #35595, opened 2016-08-19
file.copy makedirs creates root dirs


srkunze:
Given the following, we get a directory with root-only permission.

```
my_file:
  file.copy:
    - name: /not-there/my_file
    - source: /exists/my_file
    - user: me        # that's the relevant user
    - makedirs: True
```

Creates the following:

`/not-there` with root permissions
`/not-there/my_file` with user permissions

We need `/not-there` to be with user-permissions.

Ch3LL:
@srkunze can you please include a versions report: `salt --versions-report`/`salt-minion --versions-report` where you are seeing this behavior? Thanks

Ch3LL:
apologies I think I actually found a duplicate here: #22711  If you agree please close and post what version you are using in that issue report and we can track the issue there. Thanks :)

srkunze:
@Ch3LL I don't see the duplicate here. Could you explain?

For me, the **(to be created) directories** have the **wrong** permissions. I was not talking about the file perms. Those are correct for me.

@bradthurber have the directories and file created with root perms in the first place and then wants the **file** perms changed to something else afterwards.

I have **directories** created with the wrong perms. I think that's the difference.

Furthermore, I am unsure if it's a security issue if such dirs get created with other than root perms. What do you think?

We use the latest salt-ssh version 2016.3.2.

srkunze:
@Ch3LL ping

srkunze:
Oh, I see that this issues hasn't been addressed and also the related (but not duplicate) #22711 has been marked as stale and been closed.

Is this still relevant?

Ch3LL:
@srkunze thanks for bumping this for me. Looks like i'm able to replicate this. I'm working on a fix right now.

PR #48375, opened 2018-06-29
Add user and group to makedirs cmd in file.copy

Ch3LL:
### What does this PR do?
Ensures we are setting user and group when creating the directories before copying. @terminalmage one thing I was not certain about was the case when a user would use `preserve: True`. Should I be grabbing the user perms of the original file and then create the directories with those permissions? Or would it be expected that the user would not user `preserve` with make_dirs?

https://github.com/saltstack/salt/issues/35595

### Previous Behavior
Created directories with root user/group owners, not the ones specified in the state.

### New Behavior
Creates directories with correct user/group owners 

### Tests written?

Yes
### Commits signed with GPG?

Yes

### DIFF
--- a/salt/states/file.py
+++ b/salt/states/file.py
@@ -5663,7 +5663,7 @@ def copy(
     if not os.path.isdir(dname):
         if makedirs:
             try:
-                _makedirs(name=name)
+                _makedirs(name=name, user=user, group=group)
             except CommandExecutionError as exc:
                 return _error(ret, 'Drive {0} is not mapped'.format(exc.message))
         else:
--- a/salt/states/file.py
+++ b/salt/states/file.py
@@ -5663,7 +5663,7 @@ def copy(
     if not os.path.isdir(dname):
         if makedirs:
             try:
-                _makedirs(name=name, user=user, group=group)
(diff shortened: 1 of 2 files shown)


salt/states/file.py
```diff
@@ -5663,7 +5663,7 @@ def copy(
     if not os.path.isdir(dname):
         if makedirs:
             try:
-                _makedirs(name=name)
+                _makedirs(name=name, user=user, group=group)
```
rares-pop: what about 'mode'? should that be added too?

srkunze: Additional question: salt/states/file.py also contains a lot more of those `_makedirs(name=name)` calls. Shouldn't those also be called with almost all parameters?

(provided that _makedirs does the right thing)

Ch3LL: @srkunze i looked at the other `_makedirs` calls and those functions do not include the `user/group/mode` arguments to the function so I think for this PR just adding it for `file.copy` will be fine and if we want to add the user/group arguments to the other functions that can be handled in another PR in develop.

also nice catch @rares-pop i'll work on adding that

rares-pop:
Looks good to me!

rallytime (approved):

PR #48377, opened 2018-06-29
Documentation - Fix the /srv/formulas dir for 'apache-formula' in the documentation

PetrusKiendys:
### What does this PR do?
https://docs.saltstack.com/en/2017.7/topics/development/conventions/formulas.html#adding-a-formula-directory-manually

N/A

### Tests written?

No

### Commits signed with GPG?

No

### DIFF
--- a/doc/topics/development/conventions/formulas.rst
+++ b/doc/topics/development/conventions/formulas.rst
@@ -77,9 +77,9 @@ or zip file of the repository. The directory structure is designed to work with
 
     .. code-block:: bash
 
-        mkdir -p /srv/formulas/apache-formula
-        cd /srv/formulas/apache-formula
-        git clone https://github.com/saltstack-formulas/apache-formula.git
+        mkdir -p /srv/formulas
+        cd /srv/formulas
+        git clone https://github.com/saltstack-formulas/apache-formula.git /srv/formulas/apache-formula
 
         # or
 
--- a/doc/topics/development/conventions/formulas.rst
+++ b/doc/topics/development/conventions/formulas.rst
@@ -77,14 +77,14 @@ or zip file of the repository. The directory structure is designed to work with
 
     .. code-block:: bash
 
-        mkdir -p /srv/formulas
-        cd /srv/formulas
+        mkdir -p /srv/formulas/apache-formula
+        cd /srv/formulas/apache-formula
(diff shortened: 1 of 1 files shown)


gtmanfred (changes requested):
This is not a problem if you cd into the directory.

The reason it was a problem when we discussed this in the community slack was you were doing a git clone to /srv/formulas.  If /srv/formulas is an empty directory in that case, git clone will just fill up the directory with the git clone.

If something is in the directory, then it would create a new directory called apache-formula.

The documentation you have added here would make /srv/formulas/apache-formula/apache-formula, which is not desired.

doc/topics/development/conventions/formulas.rst
```diff
@@ -79,12 +79,12 @@ or zip file of the repository. The directory structure is designed to work with
...
         cd /srv/formulas
-        git clone https://github.com/saltstack-formulas/apache-formula.git
+        git clone https://github.com/saltstack-formulas/apache-formula.git /srv/formulas/apache-formula
 
         # or
 
-        mkdir -p /srv/formulas
-        cd /srv/formulas
+        mkdir -p /srv/formulas/apache-formula
+        cd /srv/formulas/apache-formula
```
gtmanfred: This will still do the same thing, because `apache-formula` is the top level file in the archive.

doc/topics/development/conventions/formulas.rst
```diff
@@ -79,12 +79,12 @@ or zip file of the repository. The directory structure is designed to work with
 
         mkdir -p /srv/formulas
         cd /srv/formulas
-        git clone https://github.com/saltstack-formulas/apache-formula.git
+        git clone https://github.com/saltstack-formulas/apache-formula.git /srv/formulas/apache-formula
```
gtmanfred: This is not needed at all, because that is exactly what it is going to do after cding into the directory.

gtmanfred (changes requested):
I just don't think this actually clarifies anything if you actually do the `cd` command.

And I don't think there is a good reason to remove that from the docs.

gtmanfred:
Closing after discussion on community slack.

PetrusKiendys:
As per the discussion in the community Slack this commit doesn't add any value or clarification, thus closing it.

Issue #48277, opened 2018-06-23
init.sls with included states fails with more than one, 'Template was specified incorrectly: False'


dvenckus:
### Description of Issue/Question

This applies to a problem with the 2018.3.1 version and I have a 2nd server built identically except for having a different version of salt.  The same state files exist on each server.  Each server runs as a masterless, stand-alone minion.

I have 2 separate Virtualbox Ubuntu 18.04 VMs:  
    Server 1:  Salt 2018.3.1 installed
    Server 2:  Salt 2017.7.4 installed

I have a directory named 'common' containing an init.sls file and several sls files that are included by init.sls in the same directory.

```
common/init.sls:
     include: 
        - common.file1
        - common.file2
        - common.fille3
```

SERVER 1 & SERVER2:

I am able to run each of the included sls files individually with success

`salt-call state.sls common.file1`

However, when I call the directory for the init.sls file

`salt-call state.sls common`

it fails.

```
[INFO    ] Loading fresh modules for state activity
[INFO    ] Fetching file from saltenv 'base', ** done ** 'common/init.sls'
[ERROR   ] Template was specified incorrectly: False
local:
    Data failed to compile:
----------
    Specified SLS common.file2 in saltenv base is not available on the salt master or through a configured fileserver
    Specified SLS common.file3 in saltenv base is not available on the salt master or through a configured fileserver
```
I have tried changing the order of the included states.... no difference.
I commented out all but 1 included file (I rotated which state I kept).... then it worked.

ALL of these states work without issue on SERVER 2 with salt 2017.7.4

### Versions Report

**SERVER 1**

Salt Version:
           Salt: 2018.3.1

Dependency Versions:
           cffi: Not Installed
       cherrypy: unknown
       dateutil: 2.6.1
      docker-py: Not Installed
          gitdb: 2.0.3
      gitpython: 2.1.8
          ioflo: Not Installed
         Jinja2: 2.10
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: 1.0.7
   msgpack-pure: Not Installed
 msgpack-python: 0.5.6
   mysql-python: Not Installed
      pycparser: Not Installed
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: Not Installed
         Python: 2.7.15rc1 (default, Apr 15 2018, 21:51:34)
   python-gnupg: 0.4.1
         PyYAML: 3.12
          PyZMQ: 16.0.2
[...]

Issue #46986, opened 2018-04-10
opts file_roots gets overwritten with pillar_roots in orchestration run


github-abcde:
### Description of Issue/Question
Exactly as https://github.com/saltstack/salt/issues/46004, however the minimal reproduction scenario is slightly different.

### Setup
master config:
```
renderer: jinja | yaml | gpg
file_roots:
  base:
  - /srv/salt/states
pillar_roots:
  base:
  - /srv/salt/pillar/base
  test:
  - /srv/salt/pillar/test
  prod:
  - /srv/salt/pillar/prod
```

`/srv/salt/states/main.sls`:
```
include:
- one
- two
- three
```

`/srv/salt/states/one.sls`:
```
{%- set foo = salt['saltutil.runner']('pillar.show_pillar', minion='foo') %}
placeholder_one:
  test.succeed_without_changes
```

`/srv/salt/states/two.sls`:
```
placeholder_two:
  test.succeed_without_changes
```

`/srv/salt/states/three.sls`:
```
placeholder_three:
  test.succeed_without_changes
```

### Steps to Reproduce Issue
run `salt-run state.orchestrate main` and notice the error:
```
Specified SLS three in saltenv base is not available on the salt master or through a configured fileserver
```

I've tried to find the issue myself, but I'm stumped...tracing the contents of `__opts__['file_roots']` through the various files (`state.py`, `fileserver/roots.py`, `fileclient.py`, `renderers/jinja.py`, amongst others) seems to indicate that once the `pillar.show_pillar` has been fired, the `__opts__['file_roots']` in `renderers/jinja.py:render` stays filled with the `pillar_roots` instead of being restored to the 'regular' `file_roots`, which causes something to be unable to find `three.sls`.
I don't know why `two.sls` is unaffected by this.

### Versions Report
Salt Version:
           Salt: 2017.7.5
 
Dependency Versions:
           cffi: Not Installed
       cherrypy: Not Installed
       dateutil: Not Installed
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.7.2
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.4.6
   mysql-python: Not Installed
      pycparser: Not Installed
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: Not Installed
[...]

PR #48388, opened 2018-06-29
[2017.7] reset file_roots for renderers after compile_pillar

garethgreenaway:
### What does this PR do?
When pillar items are compiled a new render is instantiated but the file_roots is the pillar_roots.  This change forces the __opts__['file_roots'] to be set to what is set in actual_file_roots for all renderers once compile_pillar has finished.  Adding a test when this situation is run via a orchestration runner.

#48277 
#46986

### Tests written?
Yes

### Commits signed with GPG?
Yes

### DIFF
--- a/tests/integration/runners/test_state.py
+++ b/tests/integration/runners/test_state.py
@@ -386,7 +386,7 @@ class OrchEventTest(ShellCase):
     def test_orchestration_with_pillar_dot_items(self):
         '''
         Test to confirm when using a state file that includes other state file, if
-        one of those state files includes pillar related functions that will not 
+        one of those state files includes pillar related functions that will not
         be pulling from the pillar cache that all the state files are available and
         the file_roots has been preserved.  See issues #48277 and #46986.
         '''
--- a/tests/integration/runners/test_state.py
+++ b/tests/integration/runners/test_state.py
@@ -443,7 +443,7 @@ class OrchEventTest(ShellCase):
             __reload_config=True).get('jid')
 
         if jid is None:
-            raise Exception('jid missing from run_run_plus output')
+            raise salt.exceptions.SaltInvocationError('jid missing from run_run_plus output')
 
         signal.signal(signal.SIGALRM, self.alarm_handler)
         signal.alarm(self.timeout)
@@ -453,7 +453,7 @@ class OrchEventTest(ShellCase):
                 event = listener.get_event(full=True)
                 if event is None:
                     continue
-                if event['tag'] == 'salt/run/{0}/ret'.format(jid):
+                if event.get('tag', '') == 'salt/run/{0}/ret'.format(jid):
                     received = True
                     # Don't wrap this in a try/except. We want to know if the
                     # data structure is different from what we expect!
--- a/tests/integration/runners/test_state.py
+++ b/tests/integration/runners/test_state.py
@@ -382,3 +382,88 @@ class OrchEventTest(ShellCase):
             self.assertTrue(received)
             del listener
             signal.alarm(0)
+
+    def test_orchestration_with_pillar_dot_items(self):
+        '''
+        Test to confirm when using a state file that includes other state file, if
+        one of those state files includes pillar related functions that will not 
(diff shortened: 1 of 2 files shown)


tests/integration/runners/test_state.py
```diff
@@ -382,3 +382,88 @@ def test_parallel_orchestrations(self):
...
+            opts=self.master_opts)
+
+        start_time = time.time()
+        jid = self.run_run_plus(
+            'state.orchestrate',
+            'main',
+            __reload_config=True).get('jid')
+
+        if jid is None:
+            raise Exception('jid missing from run_run_plus output')
```
isbm: Why do we raise common exceptions, instead of AssertionError?
```python
assert bool(jid), 'Job ID missing from run_run_plus output'
```

tests/integration/runners/test_state.py
```diff
@@ -382,3 +382,88 @@ def test_parallel_orchestrations(self):
...
+
+        signal.signal(signal.SIGALRM, self.alarm_handler)
+        signal.alarm(self.timeout)
+        received = False
+        try:
+            while True:
+                event = listener.get_event(full=True)
+                if event is None:
+                    continue
+                if event['tag'] == 'salt/run/{0}/ret'.format(jid):
```
isbm: I'd do it more safe:
```python
if event.get('salt/run/{0}/ret'.format(jid)):
    ...
```

tests/integration/runners/test_state.py
```diff
@@ -382,3 +382,88 @@ def test_parallel_orchestrations(self):
...
+                    # data structure is different from what we expect!
+                    ret = event['data']['return']['data']['master']
+                    for state in ret:
+                        data = ret[state]
+                        # Each state should be successful
+                        self.assertEqual(data['comment'], 'Success!')
+                    break
+        finally:
+            self.assertTrue(received)
+            del listener
```
isbm: BTW, is this really needed?

isbm (changes requested):

tests/integration/runners/test_state.py
gtmanfred: If you do an assert, it gets removed when the optomized bytecode is compiled at build time.

https://stackoverflow.com/questions/9097706/why-assert-is-not-largely-used

isbm: Oh, right, my bad. Well, I use so often optimised mode that I totally forgot about that (last 20 years of Python and none of it at all). :wink: 

OK, how about raise other than generic Exception?

tests/integration/runners/test_state.py
garethgreenaway: Which part specifically?  The assertTrue or the clean up of the listener?

tests/integration/runners/test_state.py
garethgreenaway: This would trigger the `try...finally` as the key in the event dictionary is `tag` not the tag itself.
Something like this would work though:
```
if event.get('tag', '') == 'salt/run/{0}/ret'.format(jid):
```

gtmanfred (approved):

tests/integration/runners/test_state.py
isbm: The `del`. Seems you're basically calling `SaltEvent.__del__` in order to call `SaltEvent.destroy`. I would say, we should not do this, because `__del__` is not an opposite to `__init__`. I would rather refactor it a bit:

```python
class SaltEvent(event):
    def destroy(self):
        try:
            if self.subscriber is not None:
                self.subscriber.close()
            if self.pusher is not None:
                self.pusher.close()
            if self._run_io_loop_sync and not self.keep_loop:
                self.io_loop.close()
        except Exception as err:
            log.error('Unable to close event listener')
            log.debug(err)

    __del__ = destroy
```

And then we should never use it as `del obj` but always `obj.destroy()` instead. However, we still leave `__del__` to the garbage collector.

terminalmage (approved):

PR #48406, opened 2018-07-02
[RFC] add rfc for loader changes in Fluorine

gtmanfred:
### What does this PR do?
RFC for loader changes to include package information

### Commits signed with GPG?

Yes

@cachedout @terminalmage

### DIFF
--- a/rfcs/0001-adding-package-logic-to-loader.md
+++ b/rfcs/0001-adding-package-logic-to-loader.md
@@ -1,0 +1,51 @@
+- Feature Name: adding_package_logic_to_loader
+- Start Date: 2018-06-29
+- RFC PR:
+- Salt Issue:
+
+# Summary
+[summary]: #summary
+
+The loader is kind of dumb, and has no sense of heirarchy for modules other than the order that is hard coded into `_module_dirs`.  Salt should be able to be more configurable over what should be loaded.
+
+# Motivation
+[motivation]: #motivation
+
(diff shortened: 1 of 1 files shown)


rallytime (approved):
I like this idea.

cachedout:
This is good but IMHO it needs more information about the transition from a single Python module, which is abstracted in the salt dunder as `module.func` to a set of modules that can exist in a Python package, all files in which should be _combined_ to do the same.

For example, would a function defined in `my_salt_module/__init__.py` always take precedence over a function named the same in `my_salt_module/foo.py`?

gtmanfred:
I was not aware that that is what we were discussing.

Are you talking about taking all the functions from a python package, and putting them all under one module namespace of <packagename>.<functionname>?

Because what I was proposing that we do instead is add an extra layer to that abstraction, so that we could do <packagename>:<modulename>.<functionname>, so for your example, nothing could be put in __init__.py, but the rest of the modules in the same module package do not overwrite each other at all, and if you had another package that had the same module/function name combination, you could explicitly reach one by prefixing the package name, or setting the priority higher.

terminalmage (commented):
How would we establish the priority, and what would stop a user from setting a priority of 1 to their own modules? Or do we not care (i.e. "give them enough rope to hang themselves")?

gtmanfred:
@terminalmage after talking to mike, I am working on rewriting this rfc.  Some of the priority stuff is going to be split into a different rfc, which will be all the priority and seperate pip packages stuff.

PR #48420, opened 2018-07-03
Fix log.warn to log.warning

rares-pop:
log.warn got deprecated and this commit addresses the few places
where was still in use

### What does this PR do?
Fix the deprecation warnings

### Tests written?

No

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/engines/slack.py
+++ b/salt/engines/slack.py
@@ -216,8 +216,8 @@ class SlackClient(object):
         try:
             groups_gen = itertools.chain(self._groups_from_pillar(groups_pillar_name).items(), use_groups.items())
         except AttributeError:
-            log.warn('Failed to get groups from %s: %s', groups_pillar_name, self._groups_from_pillar(groups_pillar_name))
-            log.warn('or from config: %s', use_groups)
+            log.warning('Failed to get groups from %s: %s', groups_pillar_name, self._groups_from_pillar(groups_pillar_name))
+            log.warning('or from config: %s', use_groups)
             groups_gen = []
         for name, config in groups_gen:
             log.info('Trying to get %s and %s to be useful', name, config)
@@ -231,7 +231,7 @@ class SlackClient(object):
                 ret_groups[name]['default_target'].update(config.get('default_target', {}))
                 ret_groups[name]['targets'].update(config.get('targets', {}))
             except (IndexError, AttributeError):
-                log.warn("Couldn't use group %s. Check that targets is a dict and not a list", name)
+                log.warning("Couldn't use group %s. Check that targets is a dict and not a list", name)
 
         log.debug('Got the groups: %s', ret_groups)
         return ret_groups
@@ -683,11 +683,11 @@ class SlackClient(object):
                 # 10 times without taking a break.
                 log.trace('Got a message from the generator: %s', msg.keys())
                 if count > 10:
-                    log.warn('Breaking in getting messages because count is exceeded')
+                    log.warning('Breaking in getting messages because count is exceeded')
                     break
                 if len(msg) == 0:
                     count += 1
-                    log.warn('len(msg) is zero')
+                    log.warning('len(msg) is zero')
                     continue  # This one is a dud, get the next message
                 if msg.get('done'):
                     log.trace('msg is done')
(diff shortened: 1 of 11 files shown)


salt/engines/slack.py
```diff
@@ -216,8 +216,8 @@ def get_config_groups(self, groups_conf, groups_pillar_name):
         try:
             groups_gen = itertools.chain(self._groups_from_pillar(groups_pillar_name).items(), use_groups.items())
         except AttributeError:
-            log.warn('Failed to get groups from %s: %s', groups_pillar_name, self._groups_from_pillar(groups_pillar_name))
-            log.warn('or from config: %s', use_groups)
+            log.warning('Failed to get groups from %s: %s', groups_pillar_name, self._groups_from_pillar(groups_pillar_name))
+            log.warning('or from config: %s', use_groups)
```
isbm: I guess, that can be refactored to just one log entry.

salt/engines/slack.py
```diff
@@ -231,7 +231,7 @@ def get_config_groups(self, groups_conf, groups_pillar_name):
                 ret_groups[name]['default_target'].update(config.get('default_target', {}))
                 ret_groups[name]['targets'].update(config.get('targets', {}))
             except (IndexError, AttributeError):
-                log.warn("Couldn't use group %s. Check that targets is a dict and not a list", name)
+                log.warning("Couldn't use group %s. Check that targets is a dict and not a list", name)
```
isbm: Messages would be also expand to a end-user level language. "Dict" ? A dictionary or key/value structure etc.

salt/engines/slack.py
```diff
@@ -683,11 +683,11 @@ def run_commands_from_slack_async(self, message_generator, fire_all, tag, contro
                 # 10 times without taking a break.
                 log.trace('Got a message from the generator: %s', msg.keys())
                 if count > 10:
-                    log.warn('Breaking in getting messages because count is exceeded')
+                    log.warning('Breaking in getting messages because count is exceeded')
                     break
                 if len(msg) == 0:
                     count += 1
-                    log.warn('len(msg) is zero')
+                    log.warning('len(msg) is zero')
```
isbm: Please rephrase this. Without knowing the source code this message is just useless. What does it says?

salt/states/cryptdev.py
```diff
@@ -100,7 +100,7 @@ def mapped(name,
         if name not in active.keys():
             # Open the map using cryptsetup. This does not pass any options.
             if opts:
-                log.warn('passed cryptdev options are ignored when mapping immediately')
+                log.warning('passed cryptdev options are ignored when mapping immediately')
```
isbm: Could we rephrase this?

salt/utils/smb.py
```diff
@@ -426,7 +426,7 @@ def _delete_directory_smbprotocol(path, share='C$', conn=None, host=None, userna
         conn = get_conn(host, username, password)
     if conn is False:
         return False
-    log.warn("PATH: %s %s", share, path)
+    log.warning("PATH: %s %s", share, path)
```
isbm: Path for what? Either this needs to be a debug (and then we know its context) or rephrased with more expanded explanation and a context.

isbm (changes requested):
@rares-pop since you're touching warning log messages, could you please also take care of the messages itself? They are not always descriptive. Thanks!

salt/utils/smb.py
rares-pop: I touched these because it was polluting my output (seeing warnings about warnings :) ).

I can try to rephrase some of them but that should be reviewed, since I'm really out of context for all of them.

rallytime (approved):

gtmanfred (approved):

salt/utils/smb.py
isbm: @rares-pop we can also do the other PR though. But it wouldn't hurt if you would offer less work for others. :wink: 

I leave it open to you.

isbm (approved):

salt/utils/smb.py
rares-pop: I meant, I will change the warning messages in this PR, but you or someone who has more/better context should double-check that they make sense after rephrasing, since I'm not really familiar with the code-base.

isbm: You bet! OK, you've getting my :x: now again. :laughing:

isbm (changes requested):
Re-check messages.

isbm (approved):
@rares-pop see, it just worked. :wink: Thanks!

rares-pop:
@isbm  - sure!

I feel I might have been misunderstood previously. :)

Issue #46689, opened 2018-03-23
`pkg.installed`: `hold: True` not applied to a package which is already installed.


mxork:
### Description of Issue/Question

`hold: True` is not applied to a package which is already installed. (may be specific to compound `pkg.installed`)

### Setup

```
some-packages-please:
  pkg.installed:
      - pkgs:
          - "git": "1.8.3.1"
          - "glusterfs-fuse": "3.8.4"
          - "socat": "1.7.3.2-2.el7"
          - "ceph-common": "1:0.94.5"
          - "nfs-utils": "1:1.3.0"
          - "iscsi-initiator-utils": "6.2.0.874"
          - "cifs-utils": "6.2"
      - hold: true
      - update_holds: true
```

### Steps to Reproduce Issue

Applying the above (`some-packages-please`) gives:

```
----------
          ID: some-packages-please
    Function: pkg.installed
      Result: True
     Comment: 5 targeted packages were installed/updated.
              Package glusterfs-fuse is now being held.
              Package socat is now being held.
              Package ceph-common is now being held.
              Package iscsi-initiator-utils is now being held.
              Package cifs-utils is now being held.
              The following packages were already installed: git=1.8.3.1, nfs-utils=1:1.3.0
```

and `nfs-utils`, `git` are indeed missing the version lock:

```
$ yum versionlock
Loaded plugins: fastestmirror, versionlock
0:telegraf-1.4.1-1.*
1:ceph-common-0.94.5-2.el7.*
0:cifs-utils-6.2-10.el7.*
0:glusterfs-fuse-3.8.4-18.4.el7.centos.*
0:iscsi-initiator-utils-6.2.0.874-4.el7.*
0:socat-1.7.3.2-2.el7.*
versionlock list done
```

### Versions Report

```
Salt Version:
           Salt: 2017.7.3
 
Dependency Versions:
           cffi: Not Installed
       cherrypy: Not Installed
       dateutil: Not Installed
      docker-py: 1.10.6
          gitdb: 0.6.4
      gitpython: 1.0.1
          ioflo: Not Installed
         Jinja2: 2.7.2
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.5.1
   mysql-python: Not Installed
      pycparser: Not Installed
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: Not Installed
[...]

PR #48426, opened 2018-07-03
[2017.7] fixes to states/pkg.py

garethgreenaway:
### What does this PR do?
Fixing a situation when a package is already installed via salt or manually and a state attempts to set that package to be held.  

#46689 

### Previous Behavior
Previously the holding/unholding logic was only being run against packages that were being installed.

### New Behavior
This change moves the holding logic outside and runs it against all desired packages.  Adding a new test to test holding logic.

### Tests written?
Yes

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/states/pkg.py
+++ b/salt/states/pkg.py
@@ -1690,45 +1690,45 @@ def installed(
             # checks reinstall targets works.
             pkg_ret = {}
 
-        if 'pkg.hold' in __salt__:
-            if 'hold' in kwargs:
-                try:
-                    if kwargs['hold']:
-                        hold_ret = __salt__['pkg.hold'](
-                            name=name, pkgs=pkgs, sources=sources
-                        )
-                    else:
-                        hold_ret = __salt__['pkg.unhold'](
-                            name=name, pkgs=pkgs, sources=sources
-                        )
-                except (CommandExecutionError, SaltInvocationError) as exc:
-                    comment.append(str(exc))
+    if 'pkg.hold' in __salt__:
+        if 'hold' in kwargs:
+            try:
+                if kwargs['hold']:
+                    hold_ret = __salt__['pkg.hold'](
+                        name=name, pkgs=desired, sources=sources
+                    )
+                else:
+                    hold_ret = __salt__['pkg.unhold'](
+                        name=name, pkgs=desired, sources=sources
+                    )
+            except (CommandExecutionError, SaltInvocationError) as exc:
+                comment.append(str(exc))
+                ret = {'name': name,
+                       'changes': changes,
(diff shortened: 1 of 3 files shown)


salt/states/pkg.py
```diff
@@ -1690,45 +1690,45 @@ def installed(
...
+        if 'hold' in kwargs:
+            try:
+                if kwargs['hold']:
+                    hold_ret = __salt__['pkg.hold'](
+                        name=name, pkgs=desired, sources=sources
+                    )
+                else:
+                    hold_ret = __salt__['pkg.unhold'](
+                        name=name, pkgs=desired, sources=sources
+                    )
```
isbm: Or shorter, so easier comprehend half-year later while debuggin and no same code repetitions:
```python
hold_ret = __salt__['pkg.{}hold'.format(not kwargs['hold'] and 'un' or '')](
        name=name, pkgs=desired, sources=sources
)
```

salt/states/pkg.py
```diff
@@ -1690,45 +1690,45 @@ def installed(
...
-                            name=name, pkgs=pkgs, sources=sources
-                        )
-                    else:
-                        hold_ret = __salt__['pkg.unhold'](
-                            name=name, pkgs=pkgs, sources=sources
-                        )
-                except (CommandExecutionError, SaltInvocationError) as exc:
-                    comment.append(str(exc))
+    if 'pkg.hold' in __salt__:
+        if 'hold' in kwargs:
```
isbm: No need in nested ifs and unnecessary large indent because of it:
```python
if 'pkg.hold' in __salt__ and 'hold' in kwargs:
```

isbm (changes requested):

salt/states/pkg.py
terminalmage: This doesn't look that much better. It's certainly a lot less readable. It would be more readable like this:

```python
action = 'pkg.hold' if kwargs['hold'] else 'pkg.unhold'
hold_ret = __salt__[action](
    name=name, pkgs=desired, sources=sources
)
```

salt/states/pkg.py
terminalmage: While @garethgreenaway didn't originally write this (he just dedented existing code), this would probably be a good idea, especially since there is no action taken when the `hold` argument is not passed.

terminalmage (changes requested):

rallytime:
@garethgreenaway Can you fix the lint here? https://jenkinsci.saltstack.com/job/pr-lint/job/PR-48426/1/warnings52Result/

terminalmage (approved):
LGTM

PR #48433, opened 2018-07-04
Fix issue with redismod.hmset method

discogestalt:
redis.StrictRedis.hmset expects a dict as its second arg, not a series of keyword args.

hmset(self, name, mapping) method of redis.client.StrictRedis instance
    Set key to value within hash ``name`` for each corresponding
    key and value from the ``mapping`` dict.

### What does this PR do?

Fixes the call to redis.StrictRedis.hmset in the redismod.hmset method

N/A

### Previous Behavior

Calling the redismod.hmset method either from the command line or a state/orchestration would return an invalid arguments error because the `fieldsvars` dict would be unpacked and passed as individual args:

`Passed invalid arguments: hmset() got an unexpected keyword argument 'test1'.`

### New Behavior

`fieldsvars` dict is passed as a dict, method runs successfully.

### Tests written?

No

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/modules/redismod.py
+++ b/salt/modules/redismod.py
@@ -395,7 +395,14 @@ def hmset(key, **fieldsvals):
     database = fieldsvals.pop('db', None)
     password = fieldsvals.pop('password', None)
     server = _connect(host, port, database, password)
-    return server.hmset(key, fieldsvals)
+    # orchestration passes several private key/var pairs in that we don't want
+    # to set in redis.
+    redispairs = {}
+    for field, value in fieldsvals.iteritems():
+        if not field.startswith('__'):
+            redispairs[field] = value
+
+    return server.hmset(key, redispairs)
 
 
 def hset(key, field, value, host=None, port=None, db=None, password=None):
--- a/salt/modules/redismod.py
+++ b/salt/modules/redismod.py
@@ -395,7 +395,7 @@ def hmset(key, **fieldsvals):
     database = fieldsvals.pop('db', None)
     password = fieldsvals.pop('password', None)
     server = _connect(host, port, database, password)
-    return server.hmset(key, **fieldsvals)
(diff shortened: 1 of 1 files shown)


gtmanfred (approved):

salt/modules/redismod.py
```diff
@@ -395,7 +395,14 @@ def hmset(key, **fieldsvals):
     database = fieldsvals.pop('db', None)
     password = fieldsvals.pop('password', None)
     server = _connect(host, port, database, password)
-    return server.hmset(key, fieldsvals)
+    # orchestration passes several private key/var pairs in that we don't want
+    # to set in redis.
+    redispairs = {}
+    for field, value in fieldsvals.iteritems():
+        if not field.startswith('__'):
+            redispairs[field] = value
```
gtmanfred: You don't need to do this, use `salt.utils.args.clean_kwargs`

```
return server.hmset(key, __utils__['args.clean_kwargs'](**fieldvals))
```

gtmanfred (changes requested):

salt/modules/redismod.py
discogestalt: That works in 2018.3, but I'm working from 2017.7 and `salt.utils.args.clean_kwargs` isn't available there...

gtmanfred: it will be in `salt.utils.clean_kwargs` in 2017.7, so you cannot use the __utils__ syntax, but you should still use this builtin.

discogestalt: Excellent, updating now.

gtmanfred (approved):

mirceaulinic (approved):

rallytime:
re-run all

rallytime:
Thank you for contributing @discogestalt and welcome!

PR #48435, opened 2018-07-04
Added support for using passwd to use for sudo authentication

### LINKED ISSUE #8882: salt-ssh fails without sudo and with sudo not using NOPASSWD
when i use salt-ssh without enabling sudo in the roster, it fails:

<pre>
pille@salt-master ~ % sudo salt-ssh '*' test.version
live3:
    ERROR: Permissions problem, target user may need to be root or use sudo:
      File "/tmp/.salt/salt-call", line 4, in <module>
        salt_call()
      File "/tmp/.salt/salt/scripts.py", line 77, in salt_call
        client.run()
      File "/tmp/.salt/salt/cli/__init__.py", line 303, in run
        caller.run()
      File "/tmp/.salt/salt/cli/caller.py", line 135, in run
        ret = self.call()
      File "/tmp/.salt/salt/cli/caller.py", line 72, in call
        with salt.utils.fopen(proc_fn, 'w+') as fp_:
      File "/tmp/.salt/salt/utils/__init__.py", line 872, in fopen
        fhandle = open(*args, **kwargs)
    IOError: [Errno 13] Permission denied: '/var/cache/salt/minion/proc/20131128102923233205'
</pre>

which is understandable, so i enabled sudo in the roster and got:

<pre>    
pille@salt-master ~ % sudo salt-ssh '*' test.version
live3:
    ERROR: sudo expected a password, NOPASSWD required
</pre>

it wasn't asking for my password, but trying something (probably nothing).
same, when i specify passwd in the roster.

i'm not fine using NOPASSWD for my sudo access, and like to be prompted for my password interactively. perhaps it's a good idea to implement this as a new cmdline option, so that it may fail when running automated or with a huge number of targets.

MovingEarth:
Fixes 
#8882
If sudo and passwd is set in the roster file, salt-ssh will generate
the sudo command with the -p parameter with the unique string. This
string is used to identify sudo asking for the sudo password. The
passwd provided is used.
This fix should work when passwd is used for ssh authentication and
when a seperate private key is used for ssh authentication.

### What does this PR do?

It allows a user to provide a sudo password using the passwd field in the roster file (or the command line overwrite I suppose)

#8882

### Previous Behavior

If the ssh user wanted to use sudo and had not NOPASSWD set, sudo would not succeed as a password should be provided, but salt-ssh was not expecting it. 

### New Behavior

If the passwd field is set, the -p parameter will be added to the sudo command, like
sudo -p [salt:sudo:d11bd4221135c33324a6bdc09674146fbfdf519989847491e34a689369bbce23]passwd:
if the sudo requires a password the prompt is returned. salt.ssh.client.shell will recognize the prompt and return the password.
The above allows a user to use sudo with a password and without NOPASSWD set.

### Tests written?

No
This should be tested with an integration test, which requires different accounts to be setup.

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/client/ssh/__init__.py
+++ b/salt/client/ssh/__init__.py
@@ -1175,7 +1175,12 @@ class Single(object):
         '''
         Prepare the command string
         '''
-        sudo = 'sudo' if self.target['sudo'] else ''
+        if self.target['sudo']:
+            sudo = 'sudo'
+            # set the sudo promtp parameter with the unique prompt if the password is set
+            sudo += ' -p \"{0}\"'.format(shell.SUDO_PROMPT) if self.target['passwd'] else ''
+        else:
+            sudo = ''
         sudo_user = self.target['sudo_user']
         if '_caller_cachedir' in self.opts:
             cachedir = self.opts['_caller_cachedir']
--- a/salt/client/ssh/__init__.py
+++ b/salt/client/ssh/__init__.py
@@ -1175,10 +1175,8 @@ class Single(object):
         '''
         Prepare the command string
         '''
-        if self.target['sudo']:
-            sudo = 'sudo'
-            # set the sudo promtp parameter with the unique prompt if the password is set
-            sudo += ' -p \"{0}\"'.format(shell.SUDO_PROMPT) if self.target['passwd'] else ''
+        if self.target.get('sudo'):
+            sudo = 'sudo -p \"{0}\"'.format(shell.SUDO_PROMPT) if self.target.get('passwd') else 'sudo'
         else:
             sudo = ''
         sudo_user = self.target['sudo_user']
--- a/salt/client/ssh/shell.py
+++ b/salt/client/ssh/shell.py
@@ -26,6 +26,11 @@ SSH_PASSWORD_PROMPT_RE = re.compile(r'(?:.*)[Pp]assword(?: for .*)?:', re.M)
 KEY_VALID_RE = re.compile(r'.*\(yes\/no\).*')
 SSH_PRIVATE_KEY_PASSWORD_PROMPT_RE = re.compile(r'Enter passphrase for key', re.M)
 
+# sudo prompt is used to recognize sudo prompting for a password and should therefore be
+# fairly recognizable and unique
+SUDO_PROMPT = '[salt:sudo:d11bd4221135c33324a6bdc09674146fbfdf519989847491e34a689369bbce23]passwd:'
+SUDO_PROMPT_RE = re.compile(r'{0}'.format(SUDO_PROMPT), re.M)
+
 # Keep these in sync with ./__init__.py
 RSTR = '_edbc7885e4f9aac9b83b35999b68d015148caf467b78fa39c05f669c0ff89878'
 RSTR_RE = re.compile(r'(?:^|\r?\n)' + RSTR + r'(?:\r?\n|$)')
@@ -427,6 +432,12 @@ class Shell(object):
                                       'auto accept run salt-ssh with the -i '
                                       'flag:\n{0}').format(stdout)
                         return ret_stdout, '', 254
+                elif buff and SUDO_PROMPT_RE.search(buff):
+                    if not self.passwd:
+                        return '', 'Sudo password is required but not provided', 254
+                    else:
+                        term.sendline(self.passwd)
+                        continue
                 elif buff and buff.endswith('_||ext_mods||_'):
                     mods_raw = salt.utils.json.dumps(self.mods, separators=(',', ':')) + '|_E|0|'
                     term.sendline(mods_raw)


salt/client/ssh/__init__.py
```diff
@@ -1175,7 +1175,12 @@ def _cmd_str(self):
         '''
         Prepare the command string
         '''
-        sudo = 'sudo' if self.target['sudo'] else ''
+        if self.target['sudo']:
+            sudo = 'sudo'
+            # set the sudo promtp parameter with the unique prompt if the password is set
+            sudo += ' -p \"{0}\"'.format(shell.SUDO_PROMPT) if self.target['passwd'] else ''
```
isbm: What's wrong avoid `+=` and do just `sudo = ' sudo -p \"{......` ? That said:

```python
if self.target.get('sudo'):
    sudo = self.target.get('passwd') and 'sudo -p \"{0}\"'.format(shell.SUDO_PROMPT) or ''
else:
    sudo = ''
```

salt/client/ssh/shell.py
```diff
@@ -26,6 +26,11 @@
 KEY_VALID_RE = re.compile(r'.*\(yes\/no\).*')
 SSH_PRIVATE_KEY_PASSWORD_PROMPT_RE = re.compile(r'Enter passphrase for key', re.M)
 
+# sudo prompt is used to recognize sudo prompting for a password and should therefore be
+# fairly recognizable and unique
+SUDO_PROMPT = '[salt:sudo:d11bd4221135c33324a6bdc09674146fbfdf519989847491e34a689369bbce23]passwd:'
+SUDO_PROMPT_RE = re.compile(r'{0}'.format(SUDO_PROMPT), re.M)
```
isbm: Can you explain a bit what is going on here?

isbm (changes requested):

salt/client/ssh/shell.py
MovingEarth: SUDO_PROMPT is the value sudo will output if the user should enter its password (when it is used with the -p parameter). You'll want it to be quite unlike an application will output, so you can recognize it in the output. Really almost any string would do, so I just made something up.

salt/client/ssh/__init__.py
MovingEarth: Is there a reason you use self.target.get('sudo') in your exmple in stead of self.target['sudo']?

salt/client/ssh/shell.py
isbm: Creative. :wink: Why not reuse existing `RSTR` instead?

```python
SUDO_PROMPT = 'salt:sudo:{}...'.format(RSTR)
```

salt/client/ssh/__init__.py
isbm: The code above I've made has less nested ifs and in case `sudo` key is not in `target` for whatever reason, it won't crash but will return `None`. The rest is the same as yours. Just my bad that I forgot to write `'sudo'` at the end and wrote `''`. Fixed.

salt/client/ssh/shell.py
MovingEarth: I don't think you can do that. SUDO_PROMPT_RE is used in shell._run_cmd() to recognize the sudo password request. In the same method RSTR_RE is used to recognize that we are receiving results. If RSTR was part of SUDO_PROMPT, RSTR_RE might trigger first in stead of SUDO_PROMPT_RE. It is a lot safer for SUDO_PROMPT to be a completely separate unique string.

isbm: If used the same regexps yes. OK, let's leave it for now as is, but I think this should be revisited a while later.

gtmanfred (changes requested):
I really like this, but it needs some testing added to make sure it actually works, and make sure it doesn't get broken.

salt/client/ssh/__init__.py
MovingEarth: I had to read your example a couple of times before I understood how it works. Are you sure you want a 

```python
x = test and a or b
```

in stead of 

```python
x = a if test else b
```
because it does not really help readability?
Your "and + or" construction functions as an "if else" only you depend on the execution order. In most code standards this is not allowed.

MovingEarth:
I have been looking at writing an integration for this feature, but I'm not completely sure how to continue. As far as I can tell, the current ssh integration tests, test functionality after the ssh connection has been established and the salt scripts are run on the target. The test is setup by running a ssh server on a (none default) port and using the credentials of the current user.

To test the functionality of this branch, you'll need specific user accounts setup on the target (the localhost if you don't run the integration tests in a docker container). If you want it tested fully, you would need an account without any sudo rights, an account with sudo rights and NOPASSWORD set, and an account with sudo rights what requires a password. Ideally, you'll want the above accounts with and without keybased authentication.

Then a new variant of SSHCase needs to be created, that allows the target to be defined, passwords to be set (or not) and make the use of the certificate for authentication optional. Then integration test can be created that test running priviliged and none-priviliged commands.

Am I overlooking something here?

The first hurdle I'm running into is, how to proceed with creating the accounts. To create the accounts on the fly you'll need to run the integration tests with root rights. Or should assume the accounts are there and just ignore the specific integration test when they are not?

gtmanfred:
All testing instances run `git.salt` from https://github.com/saltstack/salt-jenkins to setup, so you should be able to just add states to that for what you need to run.

You can also replicate how we are running tests by using the .kitchen.yml setup in the main salt repository.

Here is a getting started guide for test-kitchen https://kitchen.saltstack.com/docs/file/docs/gettingstarted.md

rallytime:
Hi @MovingEarth - Just a bump here. Did the comment above help?

rallytime:
Since we haven't heard back here for a while, I am going to close this. We can certainly continue the conversation and we will gladly open this again when you're ready. Or feel free to open a new PR. Whichever is easier for you. Thanks!

ichilton:
@MovingEarth Did this never get completed/merged?

mycodehost:
Bumping @Ichabond , @MovingEarth has this been merged? This issue can be a bit of a cumbersome thing to overcome when issuing salt minion installs to a large amount of linux machines via state.apply over salt-ssh or when committing a similar task that requires a ssh login when using a sudo user to perform the task.

MovingEarth:
No, this never got merged as far as I know.
Adding the tests was a lot of work as nodes with the correct set of credentials needed to be created AND the test framework needed to be extended as it did not provide ways to test this. It was just a runner that test functionality for salt-ssh after "the connection" was already made. None of the other methods were tested (at that time) and all together it was just to much work for the 10 line patch.
For my project I just patched my own instance and it worked very well.

sagetherage:
@MovingEarth 
Apologies! we no longer are using the `develop` branch please rebase this against `master`

nicholasmhughes:
Work covered in #61910. This PR can be closed.

PR #47726, opened 2018-05-18
rename boto to botomod

gtmanfred:
### What issues does this PR fix or reference?
Fixes #47694 

### Tests written?

Yes

### Commits signed with GPG?

Yes

PR #48436, opened 2018-07-04
Avoid logging inside the varstack __virtual__ function

max-arnold:
### What does this PR do?

Removes the `[ERROR   ] Can't find varstack master_top` from logs when varstack is not used.  This seems to be the recommended practice:  https://docs.saltstack.com/en/latest/ref/modules/#logging-restrictions

### Tests written?

No

### Commits signed with GPG?

No

### DIFF
--- a/salt/pillar/varstack_pillar.py
+++ b/salt/pillar/varstack_pillar.py
@@ -21,27 +21,20 @@ varstack on how this file is evaluated.
 
 from __future__ import absolute_import, print_function, unicode_literals
 
-# Import python libs
-import logging
-
-HAS_VARSTACK = False
 try:
     import varstack
-    HAS_VARSTACK = True
 except ImportError:
-    pass
-
-# Set up logging
-log = logging.getLogger(__name__)
+    varstack = None
 
 # Define the module's virtual name
 __virtualname__ = 'varstack'
 
 
 def __virtual__():
-    if not HAS_VARSTACK:
-        return False
-    return __virtualname__
+    return (
+        varstack and __virtualname__ or False,
+        'The varstack module could not be loaded: varstack dependency is missing.'
+    )
 
 
 def ext_pillar(minion_id,  # pylint: disable=W0613
--- a/salt/tops/varstack_top.py
+++ b/salt/tops/varstack_top.py
@@ -45,28 +45,20 @@ managed by salt as if given from a top.sls file.
 
 from __future__ import absolute_import, print_function, unicode_literals
 
-# Import python libs
-import logging
-
 try:
-    HAS_VARSTACK = False
     import varstack
-    HAS_VARSTACK = True
 except ImportError:
-    pass
-
-# Set up logging
-log = logging.getLogger(__name__)
+    varstack = None
 
 # Define the module's virtual name
 __virtualname__ = 'varstack'
 
 
 def __virtual__():
-    if not HAS_VARSTACK:
-        log.error("Can't find varstack master_top")
-        return False
-    return __virtualname__
+    return (
+        varstack and __virtualname__ or False,
+        'The varstack module could not be loaded: varstack dependency is missing.'
+    )
 
 
 def top(**kwargs):


salt/tops/varstack.py
```diff
@@ -45,27 +45,20 @@
...
-
 # Define the module's virtual name
 __virtualname__ = 'varstack'
 
 
 def __virtual__():
     if not HAS_VARSTACK:
-        log.error("Can't find varstack master_top")
-        return False
+        return (False, 'The varstack master_tops plugin could not be loaded: varstack module not found.')
```
isbm: Parenthesis are extra here. Please remove them.

salt/tops/varstack.py
```diff
@@ -45,27 +45,20 @@
...
 
-# Import python libs
-import logging
-
 try:
     HAS_VARSTACK = False
     import varstack
     HAS_VARSTACK = True
 except ImportError:
     pass
```
isbm: If you already doing this, then also please remove these `HAS_FOO` vars as well. As simple as that:

```python
try:
    import varstack
except ImportError:
    varstack = None

def __virtual__():
    return bool(varstack), 'The varstack module could not be loaded: varstack dependency is missing.'
```

Note, I've updated the error message. Feel free to just copypaste all that. :wink:

salt/tops/varstack.py
```diff
@@ -45,27 +45,20 @@
...
 
-# Set up logging
-log = logging.getLogger(__name__)
-
 # Define the module's virtual name
 __virtualname__ = 'varstack'
```
isbm: You don't need that.

salt/pillar/varstack_pillar.py
```diff
@@ -31,16 +29,14 @@
 except ImportError:
     pass
```
isbm:
```python
try:
    import varstack
except ImportError:
    varstack = None

def __virtual__():
        return bool(varstack), 'message here...'
```

salt/pillar/varstack_pillar.py
```diff
@@ -31,16 +29,14 @@
...
 
-# Set up logging
-log = logging.getLogger(__name__)
 
 # Define the module's virtual name
 __virtualname__ = 'varstack'
```
isbm: You don't need that.

salt/pillar/varstack_pillar.py
```diff
@@ -31,16 +29,14 @@
...
 
 
 def __virtual__():
     if not HAS_VARSTACK:
-        return False
+        return (False, 'The varstack_pillar module could not be loaded: varstack module not found.')
```
isbm: Extra parenthesis. Please lookup the example in my other comments.

isbm (changes requested):

salt/tops/varstack.py
max-arnold: @isbm Why? The module does not use any logging, so I removed it.

max-arnold:
Thanks for the feedback, I updated the PR (but not exactly as requested, see my questions above).

salt/pillar/varstack_pillar.py
max-arnold: The module does not use any logging at all (or I'm missing something).

isbm: You do not need `__virtualname__` if the filename is the same. You use it only in case the module is one of those like zypper/apt/yum/whatever but we refer to it as `pkg`. Please remove this.

salt/pillar/varstack_pillar.py
```diff
@@ -21,27 +21,21 @@
...
 __virtualname__ = 'varstack'
 
 
 def __virtual__():
-    if not HAS_VARSTACK:
-        return False
-    return __virtualname__
+    if varstack:
+        return __virtualname__
+    return False, 'The varstack_pillar module could not be loaded: varstack module not found.'
```
isbm: You do not need this. Please return just this:

```python
return bool(varstack), '......
```
Your module is not any virtual, just True is totally sufficient. In case `varstack` is `None`, the error message next to it will be reused.

salt/tops/varstack.py
```diff
@@ -45,28 +45,19 @@
...
-    HAS_VARSTACK = True
 except ImportError:
-    pass
-
-# Set up logging
-log = logging.getLogger(__name__)
+    varstack = None
 
 # Define the module's virtual name
 __virtualname__ = 'varstack'
```
isbm: You need this only in case you want to rename the module and join it to some other group of those, like `pkg` for example. Otherwise [keep it simple](https://en.wikipedia.org/wiki/KISS_principle).

salt/tops/varstack.py
```diff
@@ -45,28 +45,19 @@
...
 
 
 def __virtual__():
-    if not HAS_VARSTACK:
-        log.error("Can't find varstack master_top")
-        return False
-    return __virtualname__
+    if varstack:
+        return __virtualname__
+    return False, 'The varstack module could not be loaded: varstack dependency is missing.'
```
isbm: Same as above.

isbm (changes requested):

salt/pillar/varstack_pillar.py
max-arnold: The module filename is `varstack_pillar.py` and `__virtualname__ = 'varstack'`, so they are different.

isbm: Srsly, I never understand why it is called that way. It is `salt.pillar.varstack` and `salt.module.varstack`. Why on Earth it needs to be called "Pizza pepperoni with meat pepperoni".

isbm: I;ve just looked up. There are no imports, so it is safe.

max-arnold: Also see https://github.com/saltstack/salt/blob/develop/salt/loader.py#L1789 (virtualname should match the returned value)

isbm: @max-arnold this is in case when you do have it but misusing it:
https://github.com/saltstack/salt/blob/develop/salt/loader.py#L1770

Just remove it.

isbm: This is when you have it:
https://github.com/saltstack/salt/blob/develop/salt/loader.py#L1770

But if you return just True and your filename is properly named, it is fine.

isbm: If you have `$SALT/modules/example.py`:
```python
def __virtual__():
    return True, 'Some error message'

def hello():
    return 'hello'
```
Then:
```
salt-call --local example.hello -l error
local:
    hello
```

But:
```python
def __virtual__():
    return 'foo', 'Some error message'
```

...will render to this:

```
salt-call --local example.hello -l error
[ERROR   ] Failed to read the virtual function for module: example
```
And it won't even load the module.

max-arnold:
@isbm Done

doc/man/salt.7
```diff
@@ -288913,7 +288913,7 @@ data to return as pillar information. From there you can take a look at the
 varstack on how this file is evaluated.
 .INDENT 0.0
 .TP
-.B salt.pillar.varstack_pillar.ext_pillar(minion_id, pillar, conf)
+.B salt.pillar.varstack.ext_pillar(minion_id, pillar, conf)
 Parse varstack data and return the result
 .UNINDENT
```
gtmanfred: You do not need to make changes to this file.

We regenerate it with each release from all of the other doc strings.

salt/tops/varstack.py
```diff
@@ -45,28 +45,14 @@
...
-__virtualname__ = 'varstack'
+    varstack = None
 
 
 def __virtual__():
-    if not HAS_VARSTACK:
-        log.error("Can't find varstack master_top")
-        return False
-    return __virtualname__
+    return bool(varstack), 'The varstack module could not be loaded: varstack dependency is missing.'
```
gtmanfred: This will need to return the virtual name when renamed.

gtmanfred (changes requested):

salt/tops/varstack.py
terminalmage: Technically that is not true in this case. When the `__virtual__` returns `True`, the loader will fall back to the module name (i.e. without the .py). In this case, since it is called `varstack.py`, it would be loaded as `varstack`. So, it would work, though via a happy accident.

So it should still be changed to something like this:

```python
if varstack is None:
    return (False, 'The varstack module could not be loaded: varstack dependency is missing.')
return __virtualname__
```

Note that the same change should be made to the external pillar module.

gtmanfred: apparently my other comment didn't get posted.

I was also requesting that varstack.py gets renamed back to varstack_pillar.py, so that we do not run into the same problems we have seen with boto/boto3 and docker -> dockermod

isbm: But is really odd. Sounds to me like PHP problems: everything in one pile. Loader problems? Python has namespace, why is that now `foo.bar.spam` needs to clash with `foo.fred.spam`?

gtmanfred: Also, apparently I jumped over to the varstack tops file when reviewing and didn't realize it.

gtmanfred: I am just trying to avoid another problem like this https://github.com/saltstack/salt/pull/47726

isbm: But we should be able to have the same package names in different namespace. To me it sounds like a flaw in the Loader. Because renaming all around the place different parts into different names in order let them do not clash together sounds like a workaround, not really a fix.

Maybe we can leave this PR as is and then look into this topic?

terminalmage: @isbm Our loader temporarily monkeypatches `sys.path` while loading modules to include the directory containing the modules which are loaded. So, if the module is named `foo.py` and it does an `import foo` within it, then it will try to load itself instead of `/usr/lib/python3.6/site-packages/foo/__init__.py` and that will cause problems.

gtmanfred: it isn't a problem with namespaces for modules.

It is a problem that we are doing `import varstack` inside this module.

https://github.com/saltstack/salt/pull/48436/files#diff-caa464e4e9ee6e852e118724583cc313L29

isbm: @terminalmage yeah, in that way it will crash for sure, as it doesn't use the namespace. How we can fix this (without randomising names of the components)? Enforce full import? Reject modules that are so? Other ideas? Because it is nicer to have `salt.modules.docker` and `salt.pillar.docker` instead of `salt.modules.dockermodule` and `salt.pillar.docker_this_time_a_pillar`.

isbm: @gtmanfred well, then just fix the import?

max-arnold: > So it should still be changed to something like this:

```
if varstack is None:
    return (False, 'The varstack module could not be loaded: varstack dependency is missing.')
return __virtualname__
```

@terminalmage  @gtmanfred This is basically what I did in my initial PR, but then changed it based on the feedback.

If necessary, I can resubmit my initial PR which was more idiomatic, as far as I can tell by looking at existing modules.

isbm: @max-arnold OK, the loader seem has a problems that we need to tackle them later. The way it is now "fixed" is just that you don't have the same name in the module/pillars. And so when we have `varstack` in pillars, `varstack` in modules and `varstack` in Python, then things gets messy.

In that regard, lets do the following:

- Rename (or leave?) `salt.pillars.varstack_pillar`
- Rename `salt.modules.varstackmod`
- `varstack` — comes from the Python

```python
__virtualname__ = `varstack`

def __virtual__():
    return (varstack and __virtualname__ or False,
            'The varstack module could not be loaded: varstack dependency is missing.')
```

Then this should work. In a long run there should be either rules of the naming of the modules/pillars/states or a way to stop these sort of clashes.

max-arnold:
- Rebased to the latest develop branch
- Renamed `salt/tops/varstack.py` to `salt/tops/varstack_top.py` (this way it seems to be consistent with the corresponding `pillar/varstack_pillar.py`)
- Dropped all documentation changes

salt/pillar/varstack_pillar.py
```diff
@@ -21,27 +21,20 @@
...
 
 
 def __virtual__():
-    if not HAS_VARSTACK:
-        return False
-    return __virtualname__
+    return (
+        varstack and __virtualname__ or False,
+        'The varstack module could not be loaded: varstack dependency is missing.'
+    )
```
gtmanfred: This looks weird, what is it going to return?  Why are we not just using.
```
if varstack is None:
    return False, 'The varstack module could not be loaded: varstack dependency is missing.'
return __virtualname__

Comment (isbm) [reply salt/pillar/varstack_pillar.py]: It is going to return `__virtualname__` if `varstack` is a module, `False` otherwise. The error message is ignored.

Two returns are bad style in Python.

Comment (gtmanfred) [reply salt/pillar/varstack_pillar.py]: I would still contend that the other way is more readable, but whatever

**Review by gtmanfred — APPROVED** [review]

File: salt/tops/varstack.py (hunk 8 above)
Comment (cachedout) [reply salt/tops/varstack.py]: We should probably document a standard. There is a mix of both styles at present.

**Review by cachedout — APPROVED** [review]

**Comment by cachedout** [comment]
@isbm Are you OK with this as it sits right now?

File: salt/pillar/varstack_pillar.py (hunk 12 above)
Comment (max-arnold) [reply salt/pillar/varstack_pillar.py]: Do you want me to change this part?  I see several +1s, and actually, this is how it looked in my initial PR.

Comment (gtmanfred) [reply salt/pillar/varstack_pillar.py]: you are fine.  just waiting for @isbm to :+1: the review.

**Review by isbm — APPROVED** [review]
@max-arnold Thanks!
@cachedout While we _have to_ describe the `__virtual*` standard, we should make sure we are not enforcing workarounds for our monkeypatch problem in Loader, but we should first look there and fix it. We have import clashes and this is bad, and "just rename" is a workaround we would prefer to get rid of. Other then that yes, I am fine as it is now. But we should return to `__virtual*` topic soon.

**Comment by max-arnold** [comment]
@rallytime I'm getting a warning on the latest develop branch during `make -C doc clean html`:

```
WARNING: [autosummary] failed to import u'salt.tops.varstack': no module named salt.tops.varstack
```

Not sure if this could be fixed by running `apidoc-saltmods.sh` (it fails for me with `./apidoc-saltmods.sh: 59: ./apidoc-saltmods.sh: 1: Dirname to exclude is required.`), but it may be a good idea to add `SPHINXOPTS    = -W` to `doc/Makefile` to turn any sphinx wanings into errors.

PR #48440, opened 2018-07-04
Make core.py tolerant of invalid chars in DMI data

mmulqueen:
### What does this PR do?
Avoids unicode errors when DMI data has invalid characters in it. Solves issue #48184. Improves on PR #48216.

### Previous Behavior
UnicodeDecodeError raised and prevent salt-minion from running.

### New Behavior
These errors are not raised because invalid characters are replaced.

### Tests written?

No - but @Ch3LL has already added one test for this in PR #48216. I suppose a further test could be added with \xff as file contents.

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/grains/core.py
+++ b/salt/grains/core.py
@@ -947,7 +947,7 @@ def _virtual(osdata):
         if os.path.isfile('/sys/devices/virtual/dmi/id/product_name'):
             try:
                 with salt.utils.files.fopen('/sys/devices/virtual/dmi/id/product_name', 'r') as fhr:
-                    output = salt.utils.stringutils.to_unicode(fhr.read())
+                    output = salt.utils.stringutils.to_unicode(fhr.read(), errors='replace')
                     if 'VirtualBox' in output:
                         grains['virtual'] = 'VirtualBox'
                     elif 'RHEV Hypervisor' in output:
@@ -2288,7 +2288,7 @@ def _hw_data(osdata):
             if os.path.exists(contents_file):
                 try:
                     with salt.utils.files.fopen(contents_file, 'r') as ifile:
-                        grains[key] = ifile.read().strip()
+                        grains[key] = salt.utils.stringutils.to_unicode(ifile.read().strip(), errors='replace')
                         if key == 'uuid':
                             grains['uuid'] = grains['uuid'].lower()
                 except (IOError, OSError) as err:


terminalmage (commented):
Would `ignore` be better than `replace` here? Using `replace` would replace invalid chars with the [unicode replacement character](https://www.fileformat.info/info/unicode/char/fffd/index.htm). Using `ignore` will just leave them out of the grain value.

mmulqueen:
@terminalmage Whatever you think is best really, I picked replace over ignore arbitrarily, just anything but strict! The only thing I'd say is support of replace is at least if you get the unicode replacement character, it's an indication that you've lost something.

terminalmage:
Yeah, good point, I guess my argument against `replace` would be that it would make it harder to match on that grain's value.

@saltstack/team-core thoughts?

cachedout:
The counterpoint is that then wouldn't you then end up having to target this grain using the replacement character? That might make for an odd user experience but perhaps it isn't the biggest deal? What do you think @terminalmage and @mmulqueen ?

mmulqueen:
@cachedout @terminalmage I don't think it really matters either way. If a grain has an invalid or blank value like this, then I can't see anyone targeting it at all, they more likely going to find something that gives them a useful value to target on. The most important thing is that we stop this from crashing the rest of salt, whether that's replace or strip.

On reflection, I think it may be more user friendly to strip the invalid characters. If they're replaced then someone inspecting them may assume it's a bug in salt's unicode handing.

Shall I amend the PR to strip instead of replace?

terminalmage:
Well, I do think that when someone tries to target based on grains, they going to 
expect most of the time to be able to type out the value on which they're matching. However, if the replacement character does end up in the grain, and they have run a `grains.items`, one would presume they can copy and paste from the terminal and be able to match properly using a string containing the unicode replacement character. So, maybe not that big of an issue.

I can honestly see good reasons to do it either way, and I'm ambivalent about which we'd choose. If @cachedout doesn't think it's a big deal, I think we can merge this as-is.

cachedout:
After giving this some thought, I think the right thing to do here is to replace. There's an (admittedly) outside chance that if we strip it that we'll collide with a custom grain.

rallytime:
Sounds like this is good to go in now. Thank you @mmulqueen!

Issue #48442, opened 2018-07-05
elasticsearch.index_template_present fails on ES <6.x


babs:
### Description of Issue/Question

Index template definition fails for ES < 6.x

(patch is on the way)

### Setup

kind of same as #47701 but against ES 5.x for example

### Steps to Reproduce Issue
(Include debug logs if possible and relevant.)

### Versions Report
```
Salt Version:
           Salt: 2018.3.2
 
Dependency Versions:
           cffi: Not Installed
       cherrypy: unknown
       dateutil: 2.5.3
      docker-py: Not Installed
          gitdb: 2.0.0
      gitpython: 2.1.1
          ioflo: Not Installed
         Jinja2: 2.9.4
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.4.8
   mysql-python: Not Installed
      pycparser: Not Installed
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: Not Installed
         Python: 2.7.13 (default, Nov 24 2017, 17:33:09)
   python-gnupg: Not Installed
         PyYAML: 3.12
          PyZMQ: 16.0.2
           RAET: Not Installed
          smmap: 2.0.1
        timelib: Not Installed
        Tornado: 4.4.3
            ZMQ: 4.2.1
 
System Versions:
           dist: debian 9.4 
         locale: UTF-8
        machine: x86_64
        release: 4.9.0-6-amd64
         system: Linux
        version: debian 9.4 
 
```

Ch3LL:
Thanks for the PR!! Since the fix resolves this issue I will close here and keep discussion in PR, but let me know if you need it re-opened :)

PR #48445, opened 2018-07-05
Fix for elasticsearch.index_template_present fails on ES <6.x #48442

babs:
### What does this PR do?

It resolv a bug related to index template for Elasticsearch before 6.x

It fixes issue #48442.

### Tests written?

No

### Commits signed with GPG?

No

### DIFF
--- a/salt/states/elasticsearch.py
+++ b/salt/states/elasticsearch.py
@@ -275,10 +275,16 @@ def index_template_present(name, definition, check_definition=False):
                     ret['comment'] = 'Cannot create index template {0}, {1}'.format(name, output)
         else:
             if check_definition:
-                definition_to_diff = {'aliases': {}, 'mappings': {}, 'settings': {}}
-                definition_to_diff.update(definition)
+                if type(definition) == str:
+                    definition_parsed = salt.utils.json.loads(definition)
+                else:
+                    definition_parsed = definition
                 current_template = __salt__['elasticsearch.index_template_get'](name=name)[name]
-                diff = __utils__['dictdiffer.deep_diff'](current_template, definition_to_diff)
+                # Prune empty keys (avoid false positive diff)
+                for key in ("mappings", "aliases", "settings"):
+                    if current_template[key] == {}:
+                        del(current_template[key])
+                diff = __utils__['dictdiffer.deep_diff'](current_template, definition_parsed)
                 if len(diff) != 0:
                     if __opts__['test']:
                         ret['comment'] = 'Index template {0} exist but need to be updated'.format(name)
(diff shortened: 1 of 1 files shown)


salt/states/elasticsearch.py
```diff
@@ -275,10 +275,16 @@ def index_template_present(name, definition, check_definition=False):
                     ret['comment'] = 'Cannot create index template {0}, {1}'.format(name, output)
         else:
             if check_definition:
-                definition_to_diff = {'aliases': {}, 'mappings': {}, 'settings': {}}
-                definition_to_diff.update(definition)
+                if type(definition) == str:
```
rallytime: I think using an `isinstance` check here is more clear.

salt/states/elasticsearch.py
```diff
@@ -275,10 +275,16 @@ def index_template_present(name, definition, check_definition=False):
...
+                    definition_parsed = salt.utils.json.loads(definition)
+                else:
+                    definition_parsed = definition
                 current_template = __salt__['elasticsearch.index_template_get'](name=name)[name]
-                diff = __utils__['dictdiffer.deep_diff'](current_template, definition_to_diff)
+                # Prune empty keys (avoid false positive diff)
+                for key in ("mappings", "aliases", "settings"):
+                    if current_template[key] == {}:
```
rallytime: Same comment here. :)

rallytime (changes requested):
Thanks for this PR @babs 

I have a couple of suggestions. In additon there is a very small lint error that needs to be fixed:

https://jenkins.saltstack.com/job/PR/job/salt-pr-lint-n/23133/violations/file/salt/states/elasticsearch.py/

salt/states/elasticsearch.py
babs: The check here it not th know if it's a dict but and empty dict.
I find ```isinstance(current_template[key], dict) and len(current_template[key].key() == 0)``` a little bit too wordy for that purpose, any suggestion ?

salt/states/elasticsearch.py
babs: I'll take care of that

babs:
Hi @rallytime, should I squash the commits ?

salt/states/elasticsearch.py
rallytime: What about `if not current_template[key]:`?

cachedout:
Have we tested this change on EL 6 and above as well?

babs:
@cachedout yes I did, I've both in prod using it (5.x and 6+).
The definition is slightly different at the state level but this is due to ES representation of data and a type change.

salt/states/elasticsearch.py
babs: ```if not current_template[key]:``` would also match non dict typed values, I'll check if it might have any border effect as if an empty string was an expected value

rallytime: You're right and this is fine. It just felt that an awkward check to me.

rallytime (approved):

Issue #37670, opened 2016-11-14
influxdb_user.present state breaks authentication options


chrimi:
### Description of Issue/Question

The current develop version of influxdb_user.present breaks authentication options to influxdb due a change in parameters.

2016.3.4:
salt.states.influxdb_user.present(name, passwd, database=None, user=None, password=None, host=None, port=None)

develop (and 2016.11 rc2):
salt.states.influxdb_user.present(name, password, admin=False, **client_args)

### Setup
- installed influxdb with http authentification
- created admin user as described in influxdb documentation
- state (was valid until 2016.3.4):
```
influx-user:
  influxdb_user.present:
    - name: nonadminuser
    - passwd: theuserpass
    - user: admin
    - password: theadminpass
```

Because the state as of 2016.11 defines the `password` parameter itself `**client_args` cannot use it for authentication options (which are still defined in the influx module with *user* and *password*)

Working with the module itself works (for example):
```
salt '*' influxdb.list_users user=admin password=testpw
```

### Steps to Reproduce Issue
see *Setup*

### Versions Report
```
Salt Version:
           Salt: 2016.11.0rc2

Dependency Versions:
           cffi: Not Installed
       cherrypy: Not Installed
       dateutil: 2.6.0
          gitdb: 0.5.4
      gitpython: 0.3.2 RC1
          ioflo: Not Installed
         Jinja2: 2.7.2
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.4.6
   mysql-python: 1.2.3
      pycparser: Not Installed
       pycrypto: 2.6.1
         pygit2: Not Installed
         Python: 2.7.6 (default, Jun 22 2015, 17:58:13)
   python-gnupg: Not Installed
         PyYAML: 3.10
          PyZMQ: 14.0.1
           RAET: Not Installed
          smmap: 0.8.2
        timelib: Not Installed
        Tornado: 4.2.1
            ZMQ: 4.0.5

System Versions:
           dist: Ubuntu 14.04 trusty
        machine: x86_64
        release: 3.13.0-76-generic
         system: Linux
        version: Ubuntu 14.04 trusty
```

gtmanfred:
It looks like this is a change between the influx_user08.py and influx_user state modules.

@babilen can you take a look at this issue for us?

Thanks,
Daniel

babilen:
[...]

PR #48446, opened 2018-07-05
Influx module user fixes

eugene-davis:
### What does this PR do?
Removes a collision between the client_args parameter and the inputs to user creation and updating in the salt module and state for Influxdb. This allows user management on a system require authentication.

Also updates the documentation for the grants parameter to make its usage more clear.

https://github.com/saltstack/salt/issues/37670

### Tests written?

No

### Commits signed with GPG?

No

### DIFF
--- a/salt/states/influxdb_user.py
+++ b/salt/states/influxdb_user.py
@@ -20,7 +20,7 @@ def __virtual__():
 
 
 def present(name,
-            password,
+            passwd,
             admin=False,
             grants=None,
             **client_args):
@@ -30,7 +30,7 @@ def present(name,
     name
         Name of the user to manage
 
-    password
+    passwd
         Password of the user
 
     admin : False
@@ -52,7 +52,7 @@ def present(name,
         example user present in influxdb:
           influxdb_user.present:
             - name: example
-            - password: somepassword
+            - passwd: somepassword
             - admin: False
             - grants:
                 foo_db: read
@@ -72,7 +72,7 @@ def present(name,
             return ret
         else:
             if not __salt__['influxdb.create_user'](
-                    name, password, admin=admin, **client_args):
+                    name, passwd, admin=admin, **client_args):
                 ret['comment'] = 'Failed to create user {0}'.format(name)
                 ret['result'] = False
                 return ret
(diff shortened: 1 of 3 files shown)


rallytime:
@babilen or @hvnsweeting Would either of you like to take a look at this change and review?

salt/states/influxdb_user.py
```diff
@@ -52,11 +53,12 @@ def present(name,
         example user present in influxdb:
           influxdb_user.present:
             - name: example
-            - password: somepassword
+            - passwd: somepassword
             - admin: False
-            - grants:
-                foo_db: read
+            - grants: {
```
babilen: Why not rely on yaml parsing for the dictionary here?

babilen:
Change looks fine. Not entirely sure why the YAML example was changed though.

salt/states/influxdb_user.py
eugene-davis: As written it didn't work, I'm not super familiar with the way YAML parsing works though.

eugene-davis:
Ok I've dropped the documentation change

eugene-davis:
@babilen Anything additional I should add or someone else I should flag to get this set to merge?

babilen:
Looks great, thanks

rallytime:
@eugene-davis Can you add a note about this change to the [Fluorine Release notes](https://github.com/saltstack/salt/blob/develop/doc/topics/releases/fluorine.rst)? We need to document this kind of an API change so people can update their states appropriately.

gtmanfred (approved):
I am ok with this, since there is no way to pass in password twice (once for password= and once for client_kwargs)

`influxdb.foo_function admin supersecret password=whatever`

And there isn't a good way for there to be a deprecation path, it would take at least 4 releases to get to the desired state. (2 to deprecated password -> passwd, and then two more to make client_password in kwargs -> password).

So I am ok with this as long as there is a big note in the release docs for fluorine.

eugene-davis:
@rallytime @gtmanfred I've added to the release notes describing the change. Please let me know if this is sufficient.

gtmanfred (approved):

eugene-davis:
@gtmanfred @rallytime Is there something else I need to do prior to merge?
I see some of the checks aren't passing but this doesn't seem to be unusual looking at past merged PRs.

gtmanfred:
I was just waiting to see if any other core dev wanted to review it.

gtmanfred:
Thanks for the PR!

PR #48466, opened 2018-07-06
Fix return when ctrl-c used to gracefully exit

terminalmage:
This prevents a traceback from being printed to the screen when using using ctrl-c to exit the state.event runner after using it to watch the event bus.

### DIFF
--- a/salt/client/mixins.py
+++ b/salt/client/mixins.py
@@ -397,11 +397,15 @@ class SyncClientMixin(object):
             if isinstance(ex, salt.exceptions.NotImplemented):
                 data['return'] = six.text_type(ex)
             else:
-                data['return'] = 'Exception occurred in {0} {1}: {2}'.format(
-                    self.client,
-                    fun,
-                    traceback.format_exc(),
-                    )
+                exc_msg = six.text_type(ex)
+                if 'Exiting gracefully' in exc_msg:
+                    data['return'] = exc_msg
+                else:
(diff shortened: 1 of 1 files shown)


rallytime (approved):

salt/client/mixins.py
```diff
@@ -397,11 +397,15 @@ def _low(self, fun, low, print_event=True, full_return=False):
             if isinstance(ex, salt.exceptions.NotImplemented):
                 data['return'] = six.text_type(ex)
             else:
-                data['return'] = 'Exception occurred in {0} {1}: {2}'.format(
-                    self.client,
-                    fun,
-                    traceback.format_exc(),
-                    )
+                exc_msg = six.text_type(ex)
+                if 'Exiting gracefully' in exc_msg:
```
isbm: I think there is a better way of doing this, something like redefine exception class in `scripts/_handle_signals` like so:

```python
if signum == signal.SIGINT:
    exception_class = SaltSystemExit  # Or something else?
    exit_msg = '\nExiting gracefully on Ctrl-C'
    ...
else:
    exception_class = SystemExit

_handle_interrupt(exception_class(exit_msg)......
```

And then capture it properly in a separate catch?

isbm (changes requested):

salt/client/mixins.py
cachedout: Interesting thought. @terminalmage ?

Issue #48360, opened 2018-06-28
salt fileserver does not like cache with date in future


sjorge:
### Description of Issue/Question
I dropped a new beacon I am working on in file_root/_beacons (also tried _beacon).
```salt \* saltutil.sync_beacons``` or ```salt \* saltutil.sync_all``` are not picking it up.

Checking the local cache on the minion where they should en up, I do not see a _beacons directory. Other directories for _modules and _states show up and get synced correctly.

### Setup
n/a

### Steps to Reproduce Issue
Create a _beacons folder in your file_root, add a new beacon to it. Try to sync it to all minions.

### Versions Report
```Salt Version:
           Salt: 2018.3.1

Dependency Versions:
           cffi: 1.11.5
       cherrypy: 14.0.1
       dateutil: 2.7.2
      docker-py: Not Installed
          gitdb: 2.0.3
      gitpython: 2.1.9
          ioflo: Not Installed
         Jinja2: 2.10
        libgit2: Not Installed
        libnacl: 1.6.1
       M2Crypto: 0.29.0
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.5.6
   mysql-python: Not Installed
      pycparser: 2.18
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: Not Installed
         Python: 2.7.14 (default, Apr  7 2018, 14:51:22)
   python-gnupg: 0.4.2
         PyYAML: 3.12
          PyZMQ: 17.0.0
           RAET: Not Installed
          smmap: 0.9.0
        timelib: 0.2.4
        Tornado: 4.5.3
            ZMQ: 4.2.3

System Versions:
           dist:
         locale: UTF-8
        machine: i86pc
        release: 5.11
         system: SunOS
        version: Not Installed```

**Comment by gtmanfred** [comment]
I am unable to replicate this.


```
[root@salt ~]# cat /srv/salt/_beacons/this.py
'''
TEST
'''
[root@salt ~]# salt \* saltutil.sync_beacons
salt:
    - beacons.this
[root@salt ~]# cat /var/cache/salt/minion/extmods/beacons/this.py
'''
TEST
'''
```

Are you putting the beacon in the 'base' fileroot? or another file root?

sjorge:
I have 2 file roots, one just has all the extra modules.

Everything is picked up except the beacons.

sjorge:
[...]

PR #48478, opened 2018-07-07
Fileserver should handle list_cache from the future

sjorge:
### What does this PR do?
While debugging an issue where new beacon modules would not sync with `saltutil.sync_beacons` we discovered that the cause was because for a brief period the clock on the salt-master was incorrect. It was set to the year 2039.

This meant that salt.fileserver never invalidated the cache as it was always 'less old' than the max cache age.

#48360

### Previous Behavior
A list_cache with a modified date in the future does not expire until we catch up to that date.

### New Behavior
A list_cache with a modified date in the future will be marked invalid if more than the max cache age + 2h into the future.

The 2 hours was used to handle switches between daylight saving time to non daylight saving time.

Additional a small lint fix was added because my commit hooks prevented me from committing because of lint errors.

### Tests written?
No

### Commits signed with GPG?
No


salt/fileserver/__init__.py
```diff
@@ -129,6 +129,12 @@ def check_file_list_cache(opts, form, list_cache, w_lock):
                 else:
                     # if filelist does not exists yet, mark it as expired
                     age = opts.get('fileserver_list_cache_time', 20) + 1
+                if age < (opts.get('fileserver_list_cache_time', 20) - 60*60*2):
+                    # Cache is from the future! Warn and mark cache invalid.
+                    # NOTE: We subtract an additional 2h from fileserver_list_cache_time
+                    #       to avoid change between daylight saving time.
+                    log.warning('The file list_cache was created in the future!')
+                    age = opts.get('fileserver_list_cache_time', 20) + 1
```
DmitryKuzmenko: Why not to just change the following `if` condition to this?
```python
if age > 0 and age < opts.get('fileserver_list_cache_time', 20):
```

DmitryKuzmenko (changes requested):

salt/fileserver/__init__.py
sjorge: Because then we lose the warning. It would be good to log those case.

DmitryKuzmenko: Ok, but why not to just check that the age is positive? Because the condition "cache in the future" means `age < 0`, isn't it?

sjorge: ```if age > (-60*60*2) and age < opts.get('fileserver_list_cache_time', 20):``` might work and keep the 2 hour padding. However we would still need a separate if above to log the warning.

But honestly I am not even sure the fix proposed here is correct at all, it works for my edge case. But I am not super familiar with this area of the code.

DmitryKuzmenko: I completely agree about the warning. My discussion points here are the following.
1. The warning have to be shown each time we see the negative age.
2. We have to treat negative age as an `invalid cache` condition and recreate the cache in that case.

@terminalmage what do you think?

DmitryKuzmenko: BTW, here we're operating the unix timestamp that is seconds from the epoch in UTC so we shouldn't care about timezones and daily saving.

terminalmage (changes requested):
I think that @DmitryKuzmenko is correct here. The timestamps being used are epoch timestamps, so the DST logic should not be necessary. A negative age should be all we need to determine that the cache time is invalid, and in those cases we should re-gen the cache just the same as if it was too old.

sjorge:
Done, drop the DST bits and updated both if's to check for age < 0

salt/fileserver/__init__.py
```diff
@@ -129,7 +129,10 @@ def check_file_list_cache(opts, form, list_cache, w_lock):
                 else:
                     # if filelist does not exists yet, mark it as expired
                     age = opts.get('fileserver_list_cache_time', 20) + 1
-                if age < opts.get('fileserver_list_cache_time', 20):
+                if age < 0:
+                    # Cache is from the future! Warn and mark cache invalid.
+                    log.warning('The file list_cache was created in the future!')
+                if age < opts.get('fileserver_list_cache_time', 20) or age < 0:
```
DmitryKuzmenko: This line decides to return the existing data if the condition is `True`. So it shall be:
```python
if 0 < age < opts.get('fileserver_list_cache_time', 20):
```

DmitryKuzmenko (changes requested):

DmitryKuzmenko (approved):

DmitryKuzmenko:
@sjorge thank you!
@terminalmage your turn.

terminalmage (approved):
Excellent!

Issue #43138, opened 2017-08-23
npm.installed mistakenly throws error for packages which are "installed via remote"


F30:
### Description of Issue/Question
The `npm.installed` state throws an error when installing certain packages. I was able to track this down to packages where "is installed via remote" is printed to stdout. (Whatever that means, it's not really Google-able.) Examples for such packages are "sqlite3" or "bcrypt", and the exact line looks like this:

> [bcrypt] Success: "/tmp/npm-test/node_modules/bcrypt/lib/binding/bcrypt_lib.node" is installed via remote"

However, the packages acutally are successfully installed, so a subsequent highstate will succeed.

The cause for this appears to be that the `npm.install` module cannot parse the output of `npm install --silent --json` because of that line. It therefore returns a string, which the `npm.installed` state counts as failure.

Since #35075, there already is the special `_extract_json()` function in place, which tries to isolate the JSON from the other output like this:

    while lines and not lines[0].startswith('{') and not lines[0].startswith('['):
        lines = lines[1:]
    while lines and not lines[-1].startswith('}') and not lines[-1].startswith(']'):
        lines = lines[:-1]

Unfortunately, the line from above starts with a '[', although it's not part of the JSON.

NPM *appears* to always output a JSON object (not a list), so the checks could *probably* be limited to `startswith('{')`. But what do I know, I can't even find documentation for NPM's JSON output.

### Steps to Reproduce Issue
Create a target directory like "/tmp/npm-test" on the minion and try to install a NPM package like "sqlite3" or "bcrypt" using `npm.installed` like this:

    sqlite3:
      npm.installed:
        - dir: /tmp/npm-test

It will fail with the error "Could not install package(s) 'sqlite3'", but the package will still be installed.

### Versions Report
* Salt 2016.11.2, but the relevant code is also present on the current "develop" branch
* NPM 3.10.10

Ch3LL:
We also had to do something similar here: https://github.com/saltstack/salt/pull/41776

seems we will need to also add a check for this output as well.

F30:
Yeah, secial-casing all kinds of output feels rather ugly, but I suspect it's necessary.

Obviously, an alternative would be getting NPM to only output JSON to stdout in JSON mode. I suppose that's not really a solution for Salt, as older versions of NPM would still have to be supported? But maybe it is still worth trying for the future.

EvaSDK:
[...]

PR #48492, opened 2018-07-09
Improve identifying json out in npm module (#43138)

wyardley:
### What does this PR do?
Rather than look for [ or { at the beginning of a line, it looks for equality. Since the JSON output of `npm install` seems to have the first character on its own, this seems to work, at least for the npm version I looked at -- however, I'm open to better / other approaches that would be better. I think the previous approach was a bit too limited, since there could be other output in the future that matches `^[` or `^{`.

Not sure if we could just ignore either stdout or stderr (I think some versions of node send the json output to stdout), or if there's another way this could be fixed. I am guessing the json output is reasonably consistent, but haven't tested older versions. [edit, checked, and currently, the message is going to stdout as well as the json output, so that won't work here]

Idempotency issue - first run will "fail" with an error, even when packages get installed, if the output has a string like:
`[grpc] Success: "/usr/lib/node_modules/@google-cloud/monitoring/node_modules/grpc/src/node/extension_binary/node-v57-linux-x64-glibc/grpc_node.node" is installed via remote`, the first run will fail because the json parsing routine fails.

More details in #43138 

### Tests written?
No - but willing to write one if you let me know where it should be go and what it should be modeled on. I threw together a quick one, but not sure if it's right.
Presumably something like
https://github.com/saltstack/salt/blob/develop/tests/unit/modules/test_npm.py#L48-L49
only with that extra output? Or just testing `_extract_json()` directly?

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/modules/npm.py
+++ b/salt/modules/npm.py
@@ -162,7 +162,11 @@ def install(pkg=None,
             env.update({'SUDO_UID': uid, 'SUDO_USER': ''})
 
     cmd = ' '.join(cmd)
-    result = __salt__['cmd.run_all'](cmd, python_shell=True, cwd=dir, runas=runas, env=env)
+    result = __salt__['cmd.run_all'](cmd,
+                                     python_shell=True,
+                                     cwd=dir,
+                                     runas=runas,
+                                     env=env)
 
     if result['retcode'] != 0:
         raise CommandExecutionError(result['stderr'])
@@ -170,33 +174,9 @@ def install(pkg=None,
     # npm >1.2.21 is putting the output to stderr even though retcode is 0
     npm_output = result['stdout'] or result['stderr']
     try:
-        return salt.utils.json.loads(npm_output)
+        return salt.utils.json.find_json(npm_output)
     except ValueError:
-        pass
-
-    json_npm_output = _extract_json(npm_output)
-    return json_npm_output or npm_output
-
-
-def _extract_json(npm_output):
-    lines = npm_output.splitlines()
-    log.debug(lines)
-
-    # Strip all lines until JSON output starts
-    while lines and not lines[0].startswith('{') and not lines[0].startswith('['):
-        lines = lines[1:]
-    while lines and not lines[-1].startswith('}') and not lines[-1].startswith(']'):
-        lines = lines[:-1]
-    # macOS with fsevents includes the following line in the return
-    # when a new module is installed which is invalid JSON:
-    #     [fsevents] Success: "..."
-    while lines and (lines[0].startswith('[fsevents]') or lines[0].startswith('Pass ')):
-        lines = lines[1:]
-    try:
-        return salt.utils.json.loads(''.join(lines))
-    except ValueError:
-        pass
-    return None
+        return npm_output
 
 
 def uninstall(pkg, dir=None, runas=None, env=None):
--- a/tests/unit/modules/test_npm.py
+++ b/tests/unit/modules/test_npm.py
@@ -5,6 +5,7 @@
 
 # Import Python Libs
 from __future__ import absolute_import, unicode_literals, print_function
+import textwrap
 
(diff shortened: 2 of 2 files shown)


wyardley:
I think the tests will need to be adjusted more than I realized, I'll take a look at it

https://github.com/saltstack/salt/blob/develop/tests/unit/modules/test_npm.py#L48-L49 is mocking the output of the response, right? Or am I misunderstanding? Is there a convention for mocking a long payload? I see some stuff in the tests building an array and then doing `'\n'.join(array)`, but that would get pretty ugly (vs. a heredoc) for something like this. I don't see a fixtures directory either?

Even going pretty far back (e.g., npm 2), the format is somewhat different, and also pretty predictable.

```
root@16b37a02a7bc:~# ./node_modules/.bin/npm install --json --silent --global '@google-cloud/monitoring@^0.4.1' 
[grpc] Success: "/usr/lib/node_modules/@google-cloud/monitoring/node_modules/google-gax/node_modules/grpc/src/node/extension_binary/node-v57-linux-x64-glibc/grpc_node.node" is installed via remote
[
  {
    "name": "@google-cloud/monitoring",
    "version": "0.4.1",
```

tests/unit/modules/test_npm.py
```diff
@@ -38,39 +39,49 @@ def setup_loader_modules(self):
...
+            "salt": "SALT",
+          }
+        ]''')
+
+        # Successful run, expected output format
+        mock = MagicMock(return_value={'retcode': 0, 'stderr': '',
+                                       'stdout': mock_json_out})
         with patch.dict(npm.__salt__, {'cmd.run_all': mock}):
-            mock_err = MagicMock(return_value='SALT')
-            with patch.object(salt.utils.json, 'loads', mock_err):
```
wyardley: I'm not sure if dropping this second patch is better or not, seems like it shouldn't be needed here?

tests/unit/modules/test_npm.py
```diff
@@ -38,39 +39,49 @@ def setup_loader_modules(self):
...
-                self.assertEqual(npm.install('coffee-script'), 'SALT')
+            self.assertEqual(npm.install('coffee-script'),
+                '[\n  {\n    "salt": "SALT",\n  }\n]')
 
-        mock = MagicMock(return_value={'retcode': 0, 'stderr': 'error',
-                                       'stdout': '{"salt": ["SALT"]}'})
+        # Successful run, unexpected output format
+        mock = MagicMock(return_value={'retcode': 0, 'stderr': '',
```
wyardley: I think stderr should typically empty on the successful runs?

terminalmage (changes requested):
We already have a function that should make this much easier, `salt.utils.json.find_json()`: 

https://github.com/saltstack/salt/blob/develop/salt/utils/json.py#L31-L52

wyardley:
@terminalimage thanks, I’ll give that a shot. Do the changes to the tests look correct, and should I try to add any additional coverage?

wyardley:
@terminalmage: how's that look?

salt/modules/npm.py
```diff
@@ -162,41 +162,21 @@ def install(pkg=None,
             env.update({'SUDO_UID': uid, 'SUDO_USER': ''})
 
     cmd = ' '.join(cmd)
-    result = __salt__['cmd.run_all'](cmd, python_shell=True, cwd=dir, runas=runas, env=env)
+    result = __salt__['cmd.run_all'](cmd,
```
wyardley: I figure this is as good a time as any to switch this to be more vertical, but I can put it back if you'd prefer.

terminalmage (changes requested):
It doesn't look like there's a test case where there is output before the JSON, which is what your PR is designed to fix. Can you add one more test which does this so that we can confirm that using `salt.utils.json.find_json()` works for that use case?

Issue #44782, opened 2017-12-01
Random_master documentation issue


kubadawczynski:
### Description of Issue/Question
In documentation there is an info that random_master works only when master_type is set to failover
```
If :conf_minion:`master` is a list of addresses, and :conf_minion`master_type`
is set to ``failover`` shuffle them before trying to connect to distribute the
minions over all available masters. This uses Python's :func:`random.shuffle
<python2:random.shuffle>` method.

    random_master: True
```
But during salt-call if salt_master is set to default value (srt), this option also must be set to `True`, to shuffle between masters.

garethgreenaway:
@kubadawczynski Thanks for the report.  We would appreciate a PR to update the documentation to include this note.

rallytime:
I have updated the docs for this as part of the change needed in deprecating `master_shuffle`.

PR #40502, opened 2017-04-03
Add docs per #39807

thatch45:
Just add some clarifying documentation

Issue #39807, opened 2017-03-03
Incorrect Logging of info for multimaster setup


daswathn:
### Description of Issue/Question
The minion is setup for multi-master setup, but when you restart the minion the log says "random_master is True but there is only one master specified. Ignoring" which is not true, this error is being reported in the salt-call 2016.11.2 (Carbon) . It was working well with 2016.3.1.

But its working fine, communication between the multi-master and minion is fine. 

### Setup
cat /etc/salt/minion
```
master:
  - master01
  - master02
  - master03
random_master: True
```
### Steps to Reproduce Issue
Install the minion 2016.11.2 (Carbon)  with multi-master setup as show below, start the minion check the logs. 

> 2017-03-02[salt.minion][WARNING ][24488] random_master is True but there is only one master specified. Ignoring.
> 2017-03-02 [salt.minion][WARNING ][24488] random_master is True but there is only one master specified. Ignoring.
> 2017-03-02 [salt.minion][WARNING ][24488] random_master is True but there is only one master specified. Ignoring.

 
### Versions Report
Salt Master version 
salt-call --version
salt-call 2016.11.2 (Carbon)

Minion Version 
Salt Version:
           Salt: 2016.11.2

Dependency Versions:
           cffi: 1.4.2
       cherrypy: Not Installed
       dateutil: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.8
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.4.6
   mysql-python: 1.2.2
      pycparser: 2.14
       pycrypto: 2.6.1
         pygit2: Not Installed
         Python: 2.6.9 (unknown, Aug  5 2016, 11:15:31)
   python-gnupg: Not Installed
         PyYAML: 3.11
          PyZMQ: 15.2.0
           RAET: Not Installed
          smmap: Not Installed
        timelib: Not Installed
        Tornado: 4.3
            ZMQ: 4.1.3

System Versions:
           dist: SuSE 11 x86_64
        machine: x86_64
        release: 3.0.101-94-default
         system: Linux
        version: SUSE Linux Enterprise Server  11 x86_64

gtmanfred:
I have confirmed this.  It actually looks like it is not choosing a random master.

[...]

PR #48494, opened 2018-07-09
Deprecate 'master_shuffle' in favor of 'random_master'

### LINKED ISSUE #40532: Deprecate master_shuffle config param in favor of random_master
These two both do the same thing, with master_shuffle implementing an additional "failback" option

Per https://github.com/saltstack/salt/pull/40502#issuecomment-291269993, we will be deprecating ``master_shuffle`` in favor of ``random_master``, with ``random_master`` assuming the failback functionality currently unique to ``master_shuffle``.

rallytime:
This PR deprecates the `master_shuffle` minion config option in favor of the `random_master` minion config option.

Fixes #40532

As discussed in #40532, #40502, and #39807, the `master_shuffle` option and the `random_master` option are basically the same. This PR adds a deprecation warning to the user when `master_shuffle` is detected, and then sets `random_master` to the value used in `master_shuffle`.

It was determined/discussed in the issues listed above that the `random_master` option also needed to be used with `master_type: failover` and the documentation was updated accordingly in #40502. However, this is no longer the case as seen [here](https://github.com/saltstack/salt/blob/e8fc110b524d5b4407bfa9aecc46fd450be67d4e/salt/minion.py#L584). If `random_master` is set, the masters are shuffled, whether `master_type` is set to failover or not. Furthermore, if you were using `master_shuffle` and `random_master`, the masters would have been shuffled twice and the original `master_failback` shuffling would be overruled by the second shuffle (linked above).

This PR maintains the `random_master` behavior. It also allows the user to set `master_type: failover` in addition to `master_failback: True` to use the failback functionality previously only available to `master_shuffle`.

I have also updated all of the relevant documentation to reflect the fact that `random_master` does not require `master_type: failback` (see Issue #44782).

@thatch45 and @terminalmage You were involved the most in the discussion in those linked issues. Can you review this and let me know if you see any problems?


salt/minion.py
```diff
@@ -578,12 +581,19 @@ def eval_master(self,
...
+            # shuffle the masters and then loop through them
             if opts['random_master']:
-                shuffle(opts['local_masters'])
-            last_exc = None
-            opts['master_uri_list'] = list()
+                # master_failback is only used when master_type is set to failover
+                if opts['master_type'] == 'failover' and opts['master_failback']:
+                    secondary_masters = opts['local_masters'][1:]
+                    shuffle(secondary_masters)
+                    opts['local_masters'][1:] = secondary_masters
```
isbm: Maybe we should clarify so that only secondary masters are shuffled?

isbm (approved):
@rallytime question only, regarding the documentation. But it is up to you.

DmitryKuzmenko (approved):
LGTM

salt/minion.py
rallytime: Do you mean the [master_failback option](https://docs.saltstack.com/en/latest/ref/configuration/minion.html#master-failback) docs? It seems implied that the first master wouldn't be shuffled, but I can see how that might be confusing.

isbm: @rallytime yeah, that part. And they need to know that since `secondary_masters` are attached to the `opts['local_masters'][1:]` has implications too.

cachedout:
I'm wondering here if it would be better just to alias these together. This is going to end up being a breaking change for people and I'm not sure that's really necessary.

rallytime:
@cachedout Can you elaborate on what you mean by a breaking change? Other than needing to update the config option?

isbm:
@cachedout as I understand, there is none of breaking changes. All it does, just cries into logs that this particular option is deprecated and needs to be updated to another particular option. But it will still work for a couple of releases.

As of necessity, I _personally_ think it is a good idea. `random_master` came up in 2014, `master_shuffle` in 2013. But both doing the same thing. While it maybe could a better name in both cases, at least getting rid of one of those is certainly the right direction.

cachedout:
I should have clarified. Obviously, a deprecation warning is not a breaking change but _removal_ of this option would be, if somebody had it set.

rallytime:
@terminalmage and @thatch45 This request came from the discussion in #40532, which I was tackling here to help out with Fluorine tasks. Can one of you chime in on this when you get a moment and address @cachedout's concern?

I don't have a strong opinion either way since there are pros and cons to both.

rallytime:
Ah, ok. I chatted with @cachedout on another channel about this as I was misunderstanding the need. Since this is a sensitive area (minion config) and it is easy to maintain, instead of removing support for the config, let's alias it with a warning and update the docs accordingly. I'll make those updates today.

rallytime:
@cachedout Ready for re-review. :)

terminalmage (approved):

cachedout (approved):

PR #48511, opened 2018-07-10
pkg.installed version parameter support for multiple conditions

0w3w:
### What does this PR do?
salt.states.pkg.installed supporting multiple version conditions.

### Previous Behavior
Support for specifying a package condition in the version was recently added, this to enable more control over what versions will be installed.
E.G.
```
mypkgs:
  pkg.installed:
    - pkgs:
      - foo
      - bar: '>=1.2.3-4'
```
see [https://docs.saltstack.com/en/latest/ref/states/all/salt.states.pkg.html#salt.states.pkg.installed](url)

### New Behavior
This feature enables support for multiple conditions, this to allow version ranges and exclusions, which will enable more complex scenarios.
E.G.
```
mypkgs:
  pkg.installed:
    - pkgs:
      - foo
      - bar: ' > 12.0.0, < 15.0.0, != 14.0.1'
```

### Tests written?
Yes


meaksh (approved):
Looks good to me :+1:. Thanks @0w3w for contributing!

0w3w:
@rallytime pylint errors fixed!

tests/unit/utils/test_pkg.py
```diff
@@ -0,0 +1,47 @@
...
+        ("!=15.0.0", "!=", "15.0.0"),
+        ("<=>15.0.0", "<=>", "15.0.0"),
+        ("<>15.0.0", "<>", "15.0.0"),
+        ("=15.0.0", "=", "15.0.0"),
+        ("", "", "")
+    ]
+
+    def test_split_comparison(self):
+        '''
+        Test passing a jid on the command line
```
terminalmage: I don't think that this test has anything to do with JIDs. Bad copy-and-paste?

tests/unit/utils/test_pkg.py
```diff
@@ -0,0 +1,47 @@
...
+        '''
+        for test_parameter in PkgUtilsTestCase.test_parameters:
```
terminalmage: There is no need to refer to class attributes here are they are inherited by the instance. You can use `self.test_parameters` here.

terminalmage (changes requested):

0w3w:
@terminalmage Comments addressed, good catch, thanks for the review!

terminalmage (approved):

damon-atkins:
Hi 0w3w 
Just a few questions.

This change assumes comma and spaces are not used in a version string. Before it assumed prefix of a version string would not be `>=` etc.  What impact will this have? I suspect none? However might be worth a note in the doco.   Have you added the doco to pkg.installed() for this change?

What happens if 14.0.1 or 17.0 is already installed, is it removed or upgraded or down-graded?
`bar: ' > 12.0.0, < 15.0.0, != 14.0.1'`
If down or up graded which version does it move too.

I am assume that compare version() is coming from the execution module still.

`allow_updates` should be removed. i.e. time to add a warning the setting is going.

The following seems to be in only one place.
```
            if allow_updates and len(version_conditions) == 1 and operator == '==':
                operator = '>='
```

I would have thought that allow_updates bool check could be removed from other parts of the code  if version string is prefixed with `>=` when allow_updates is true at the start of the functions.

0w3w:
Hey @damon-atkins 

You are right, the change assumes version numbers following the debian standard (https://www.debian.org/doc/debian-policy/#version) as well as the classic standard version format (major.minor.update.build), which covers, I would say, all of the versioning schemas. The REGEX part to detect the version string was not change though, so whatever version string was previously supported, will still be supported. The list of operators was expanded to support libsolv-based package managers and debian operators.
I don't expect any impact on this change, as it is, in essence, backwards compatible.

I have not updated the DOCO as I expect the first pkg provider to claim support for this, to update it. (opkg can be a great provider for this, I'll talk to Alejandro del Castillo, the maintainer of opkg, to see if he can implement this)

"What happens if 14.0.1 or 17.0 is already installed, is it removed or upgraded or down-graded? bar: ' > 12.0.0, < 15.0.0, != 14.0.1'"
It will be downgraded to the highest available version that meets the condition.

Didn't make changes to compare_version, should come from the same module.

Regarding the allow_updates behavior, I purposefully maintained the existing logic to warranty backwards compatibility and didn't change existing behavior.

Thanks for the review!

rallytime:
Thanks for this addition @0w3w and congrats on your first salt PR! :)

PR #48518, opened 2018-07-11
[2017.7] fixed selinux configuration path for Redhat and Debian based distros

equilibri0:
### What does this PR do?
Fixing the selinux configuration path for Redhat based distributions. The selinux.py module is trying to use /etc/selinux/config in any case, the path under redhat/centos has to be /etc/sysconfig/selinux.

`       
----------
                 ID: permissive
           Function: selinux.mode
             Result: False
                File "/usr/lib/python2.7/site-packages/salt/state.py", line 1870, in call
                  **cdata['kwargs'])
                File "/usr/lib/python2.7/site-packages/salt/loader.py", line 1818, in wrapper
                  return f(*args, **kwargs)
                File "/usr/lib/python2.7/site-packages/salt/states/selinux.py", line 118, in mode
                  oldmode, mode = mode, __salt__['selinux.setenforce'](tmode)
                File "/usr/lib/python2.7/site-packages/salt/modules/selinux.py", line 180, in setenforce
                  raise CommandExecutionError(msg.format(str(exc)))
              CommandExecutionError: Could not read SELinux config file: [Errno 2] No such file or directory: '/etc/selinux/config'
            Started: 11:17:21.093402
           Duration: 1.284 ms
  `

### Tests written?

No

### Commits signed with GPG?

No


Equilibri0:
Hi @rallytime  sure . Thx!

salt/modules/selinux.py
```diff
@@ -164,7 +164,11 @@ def setenforce(mode):
             msg = 'Could not write SELinux enforce file: {0}'
             raise CommandExecutionError(msg.format(str(exc)))
 
-    config = '/etc/selinux/config'
+    if 'RedHat' in __grains__['os_family']:
```
gtmanfred: This should be `==` not an `in` comparison.

salt/modules/selinux.py
```diff
@@ -164,7 +164,11 @@ def setenforce(mode):
...
+        config = '/etc/sysconfig/selinux'
+
+    elif 'Debian' in __grains__['os_family']:
```
gtmanfred: same here.

gtmanfred (changes requested):
Just a small change.

Thanks for submitting this!

salt/modules/selinux.py
Equilibri0: Thx for the tip

salt/modules/selinux.py
Equilibri0: also already changed ;-)

gtmanfred (approved):

rallytime (approved):

cachedout:
Has his _ever_ worked on Redhat or is this a change that Redhat distros made recently? Also, any chance we can add some testing here?

gtmanfred:
Actually, it looks like /etc/sysconfig/selinux should just be a symlink to /etc/selinux/config

```
Note
The /etc/sysconfig/selinux contains a symbolic link to the actual configuration file, /etc/selinux/config. 
```

https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/5/html/deployment_guide/ch-selinux

RedHat 5.

Same note on Redhat 7

https://access.redhat.com/documentation/en-US/Red_Hat_Enterprise_Linux/4/html/Reference_Guide/s2-SELinux-files-etc.html

```
The /etc/sysconfig/selinux contains a symbolic link to the actual configuration file, /etc/selinux/config. 
```

gtmanfred:
This PR should not be needed based on the documentation from RedHat.

Equilibri0:
@gtmanfred you are right with the symbolic link, but redhat is also writing in the doc this : ( see quote)

For me the question would be -> If redhat recommend to change the setting in this file /etc/sysconfig/selinux .. should not the task be based on this recommendation?


> There are two ways to configure SELinux under Red Hat Enterprise Linux: using the Security Level Configuration Tool (system-config-securitylevel), or manually editing the configuration file (/etc/sysconfig/selinux).
The /etc/sysconfig/selinux file is the primary configuration file for enabling or disabling SELinux, as well as setting which policy to enforce on the system and how to enforce it.

gtmanfred:
sure, but the actual configuration file for selinux is /etc/selinux/config.

jeduardo:
@gtmanfred hey, joining the discussion here. Wouldn't it then be interesting to have the config parameters as a configurable one that defaults to the base config path defined as defaults for the software component but allow one to override the config in the case the file is located in a different place for a specific distribution?

Going by @Equilibri0's example, if RedHat ever decides to change how they deploy the actual config it seems that they will keep the `/etc/sysconfig/selinux` as a standard path as they centralize every configuration under the `/etc/sysconfig` directory. If the module (and later the state as well) allows one to override the hardcoded default it would make the module (and state) much more flexible for different deployments, wouldn't it?

gtmanfred:
If RedHat updates Fedora and makes this change, then I think it would be a good time to discuss changing this, and making it a configuration option.  But as it has been this way since selinux was written by the NSA, i am not inclined to make the change.

And if it does get changed, it would be an easy drop in fix for anyone using the selinux module to make until a new release of salt comes out fixing it.

The other thing that concerns me is trying to edit a file via the symlink.  See tools like `sed` which will actually just replace the symlink file with an actual file instead of modifying the file the symlink points to.  Until it is not a symlink, and selinux actually supports changing the config file, I don't think this is something we need to add.  https://selinuxproject.org/page/GlobalConfigurationFiles#.2Fetc.2Fselinux.2Fconfig_File

Issue #47954, opened 2018-06-04
ordering in includes is broken.


steverweber:
pillar:

x.sls
```
include:
    - x3
    - x2
    - x1
```
x3.sls
```
x: 3
```
x2.sls
```
x: 2
```
x1.sls
```
x: 1
```

```
salt-run --log-level=debug pillar.show_pillar system_with_state.test.com
```

```
[DEBUG   ] Rendered data from file: /var/cache/salt/master/git_pillar/...
include:
    - x3
    - x2
    - x1
[DEBUG   ] Rendered data from file: /var/cache/salt/master/git_pillar/...//x1.sls:
[DEBUG   ] Rendered data from file: /var/cache/salt/master/git_pillar/...//x2.sls:
[DEBUG   ] Rendered data from file: /var/cache/salt/master/git_pillar/...//x3.sls:

x: 3
```
where x should be `1`!

the bug seems to be:
https://github.com/saltstack/salt/blob/develop/salt/pillar/__init__.py#L764

it is changing the matched_pstates array to a set:
```
for sub_sls in set(matched_pstates):
```
perhaps it was changed to this to remove duplicates...

fixes the issue.
```
for sub_sls in matched_pstates:
```

steverweber:
this is kinda critical to get fixed and back-ported.

Ch3LL:
i'm not able to replicate this on develop which is the version of salt you pointed to in the code. can you include a versions-report so i can attempte to replicate on the same version you are using.

Here is my return from salt-run

```
➜  salt git:(7de5dbc755) sudo salt-run pillar.show_pillar the_cake_is_a_lie
x:
    1
```

Do you have anything set in master config to change the order of pillar rendering?

steverweber:
i'll need to do some more investigation on this..
noticed today that the fix didn't work ....
`salt-run pillar.show_pillar the_cake_is_a_lie`
is giving different ordering than
`salt the_cake_is_a_lie pillar.items`
really odd.

Ch3LL:
ping @terminalmage might have some ideas here as well to help with the fix

steverweber:
lets also add a test case along with the fix so we don't run into this one again.
I would create the test case myself a few months ago but had trouble figuring out how to write it.

steverweber:
@Ch3LL I'll create a pull request because this is a problem I don't want to patch on my end for each release.

cachedout:
[...]

PR #48524, opened 2018-07-11
keep the array order of includes

steverweber:
keep the array order of includes
https://github.com/saltstack/salt/issues/47954

Tests written?: No


salt/pillar/__init__.py
```diff
@@ -761,7 +761,7 @@ def render_pstate(self, sls, saltenv, mods, defaults=None):
                                      '\'{0}\' found'.format(saltenv)]
                                 )
 
-                        for sub_sls in set(matched_pstates):
+                        for sub_sls in matched_pstates:
```
gtmanfred: These will still need to be uniqued, because it is possible for multiple states to include the same state.

gtmanfred (changes requested):

isbm (changes requested):

cachedout:
We will want @RichardW42 to look at this since he made the original change. This seems like it should be safe but I would like to hear what he thinks.

salt/pillar/__init__.py
RichardW42: @isbm 
```
[_matched_pstates.append(el) for el in matched_pstates if el not in _matched_pstates]
```

Will keep the first item of duplications, but I think in isse #47954 , @steverweber expected keep the last one of duplications.

RichardW42:
@cachedout It seems safe to me. But the concerns about unique includes still needs discussion. If there are multiple states includes:

- A:no unique function applied
  -  A1: pillar_includes_override_sls set to false(which is defaut) or not set: those states will be executed multiple times in order.
  - A2: pillar_includes_override_sls set to true, states will be merged and executed.
- B: unique function applied
  - B1: keep the first item of duplications
  - B2: keep the last item of duplications

As for now, I prefer A1, which is what this mr does.

salt/pillar/__init__.py
RichardW42: I noticed that. Maybe I am over-thinking about this case:

```
>>> matched_pstates = ['a', 'b', 'a']
>>> _matched_pstates = []
>>> [_matched_pstates.append(el) for el in matched_pstates if el not in _matched_pstates]
>>> _matched_pstates
['a', 'b']
>>>
```
Which comes to a my consideration that this `a` state should be keep before `b` or after?

RichardW42: Yes, I was thinking about overring/merge policy. Later of this block of code theres was an option called "pillar_includes_override_sls", maybe this unique policy should also provided as an option with default value set, so users can choice to use unique or override or merge policy.

rallytime:
@steverweber Did you get a chance to review this feedback? Where are we on this PR?

steverweber:
yes i seen the feedback... I did do some thought before removing the `set()` yes duplication are not handled but they never prevented in the past releases.

removing duplicates will change the behaviour that was not in the previous releases. /So some caution/.

Personally I would merge this now and perhaps add removing duplication as a new PR. Feedback from @RichardW42 would be nice... lets look at it like this.  If someone had caught that the change from Richard would change the ordering of the includes it likely would not have been merged.

steverweber:
also duplicates might be important because i think salt allows something like
```
include:
   - .add_user
        defaults:
            user: steve
   - .add_user 
        defaults:
            user: joe
```

gtmanfred:
That makes sense.

gtmanfred (approved):

terminalmage (approved):
Looks good.

PR #48530, opened 2018-07-11
Walkthrough: define the SLS accronym

SimonSapin:
Some Googling suggests this is what it means.


cachedout (approved):

cachedout:
Great! Thanks, @SimonSapin!

PR #48532, opened 2018-07-11
Enhancements to slack engine and wtmp beacon

terminalmage:
For the `slack` engine, the reuturn from commands spawned using the engine is now formatted using the engine specified in the raw return. Previously this would use the highstate outputter for all functions from the `state` module, and the yaml outputter for everything else.

For the `wtmp` beacon, the following was added:

- `action` key added to the event data (either "login" or "logout")
- Configurable ut_type to distinguish login and logout events (because hey, why the hell not)
- Use of `__context__` to store user attached to a given tty so that we can report the user that logged out.

Additionally, the wtmp and btmp beacon docs have been modified to include a reactor use case for using the events that are fired by the beacons.

### DIFF
--- a/doc/topics/releases/fluorine.rst
+++ b/doc/topics/releases/fluorine.rst
@@ -453,6 +453,26 @@ proxy.
 
     no_proxy: [ '127.0.0.1', 'foo.tld' ]
 
+Changes to :py:mod:`slack <salt.engines.slack>` Engine
+======================================================
+
+The output returned to Slack from functions run using this engine is now
+formatted using that function's proper outputter. Earlier releases would format
+the output in YAML for all functions except for when states were run.
+
+Enhancements to :py:mod:`wtmp <salt.beacons.wtmp>` Beacon
+=========================================================
+
+A new key, `action`, has been added to the events fired by this beacon, which
+will contain either the string `login` or `logout`. This will simplify reactors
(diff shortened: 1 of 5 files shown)


cachedout:
Changelog docs?

terminalmage:
Added.

doc/topics/releases/fluorine.rst
```diff
@@ -453,6 +453,26 @@ proxy.
...
+======================================================
+
+The output returned to Slack from functions run using this engine is now
+formatted using that function's proper outputter. Earlier releases would format
+the output in YAML for all functions except for when states were run.
+
+Enhancements to :py:mod:`wtmp <salt.beacons.wtmp>` Beacon
+=========================================================
+
+A new key, `action`, has been added to the events fired by this beacon, which
```
rallytime: Double backticks for `code` demarcation? :)

There's some other examples below that also need double backticks instead of single.

rallytime (approved):
I have one picky comment, but it's not a big deal. Only something I noticed.

doc/topics/releases/fluorine.rst
terminalmage: Good catch. I have been writing a lot of markdown this week and my head was still in that mode. :smile:

PR #48538, opened 2018-07-11
Revamp ansible roster to use `ansible-inventory`

gtmanfred:
### What does this PR do?
Ansible now provides an ansible-inventory command which renders the inventory script/file.  Because of this, we can now use that command to render the inventory for ourselves, instead of having to maintain code to process the file as well.

### Tests written?

Yes

### Commits signed with GPG?

Yes


salt/roster/ansible.py
```diff
@@ -113,189 +105,46 @@
...
-        self.hostvars[key] = value
-
-    def _parse_parents(self, key, value):
-        '''
-        Parse children data from inventory_file
-        '''
-        if key not in self.parents:
-            self.parents[key] = []
-        self.parents[key].extend(value)
+    inventory = __runner__['salt.cmd']('cmd.run', 'ansible-inventory -i {0} --list'.format(get_roster_file(__opts__)))
```
isbm: I am really wondering why we would execute that in so heavy-weight fashion spinning off another Python interpreter etc, instead of just reuse the `ansible.cli.InventoryCLI` for this. As an example:

```python
import ansible.cli.inventory

c = ansible.cli.inventory.InventoryCLI(['', '-i', get_roster_file(__opts__), '--list'])
c.parse()
inventory = c.run()
```

isbm (changes requested):

salt/roster/ansible.py
gtmanfred: We still support python 3.4, which ansible does not, so in order to use ansible-inventory on centos 7, right now you have to pip2 install ansible, and then run the binary.

Until we drop python3.4 support i think this is the better setup.

isbm: Ah, you're right. CentOS 7 and family on 3.4. OK, in this case we should do the fallback then:

```
return (ansible_inventory or which('ansible-inventory')) and __virtualname__, '....'
```

I still don't want to spin-off another Python process on an expensive IBM Mainframe. :-)

gtmanfred:
```
.tox/py36/lib/python3.6/site-packages/ansible/cli/inventory.py:186: in run
    exit(0)
_ _ _ _ _ _ _ _
```

This is why we cannot use the ansible/cli/inventory.py script, because it calls a system exit.

isbm (approved):

terminalmage (approved):

PR #48541, opened 2018-07-12
Adding a pair of Proxy and Execution modules for managing Cisco Nexus switches via the NX-API

mirceaulinic:
### What does this PR do?

Adds the initial version of two new modules: one Proxy and one Execution module for managing Cisco Nexus switches via via [NX-API](https://www.cisco.com/c/en/us/td/docs/switches/datacenter/nexus9000/sw/6-x/programmability/guide/b_Cisco_Nexus_9000_Series_NX-OS_Programmability_Guide/b_Cisco_Nexus_9000_Series_NX-OS_Programmability_Guide_chapter_011.html). Well, the NX-API is not the most reliable API in this world, but it's there and it can be used up to some extent.

The execution module is flexible enough to execute the commands on the (remote) device without necessarily requiring to be running inside the `nxos_api` Proxy Minion. This goes well with its SSH-based counterpart https://docs.saltstack.com/en/latest/ref/modules/all/salt.modules.nxos.html providing good coverage for NXOS switches.

The most important side of these modules is the fact that it has no external dependencies, as everything is HTTP-based. Not only that it's easier to get it up and running, but that also means it's straight forward to have it running under a regular Minion: on Nexus switches it is possible to install the Salt Minion under a container running in a  what-is-called [OAC](https://www.cisco.com/c/en/us/td/docs/switches/datacenter/nexus5000/sw/programmability/guide/b_Cisco_Nexus_5K6K_Series_NX-OS_Programmability_Guide/open_agent_container.html) (why keep it simple when can you make it awkward, right?). Anyway, the idea is that the user only has to execute `yum install salt-minion` in that OAC, then can start using: `salt-call --local nxos_api.show 'show bgp sessions'` without any further headache.
If, for whatever reasons, the user can't / is not allowed / the hardware does not support the OAC thing, can always fallback to using the `nxos_api` Proxy instead.

### DIFF
--- a/doc/ref/modules/all/index.rst
+++ b/doc/ref/modules/all/index.rst
@@ -297,6 +297,7 @@ execution modules
     npm
     nspawn
     nxos
+    nxos_api
     omapi
     openbsd_sysctl
     openbsdpkg
--- a/doc/ref/modules/all/salt.modules.nxos_api.rst
+++ b/doc/ref/modules/all/salt.modules.nxos_api.rst
@@ -1,0 +1,6 @@
+salt.modules.nxos_api module
+============================
+
+.. automodule:: salt.modules.nxos_api
+    :members:
+
--- a/doc/ref/proxy/all/index.rst
+++ b/doc/ref/proxy/all/index.rst
@@ -22,6 +22,7 @@ proxy modules
     napalm
     netmiko_px
     nxos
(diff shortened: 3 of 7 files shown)


gtmanfred (approved):

rallytime (approved):

PR #48545, opened 2018-07-12
Fix unreachable ssh-id-wrapper template when root_dir is relative

bbinet:
Since the git commands will be executed from the user home directory, if
salt root_dir is set to a relative path (for example using `root_dir: ./.tmp/` in salt config),
then the git/ssh-id-wrapper template may be unreachable and we may
encouter the following error that this patch is fixing:

```
[ERROR   ] Command '[u'git', u'ls-remote', u'git@github.com:bbinet/salt-formula-linux.git']' failed with return code: 128
[ERROR   ] stderr: error: cannot run .tmp/thin/py2/salt/templates/git/ssh-id-wrapper: No such file or directory
fatal: unable to fork
[ERROR   ] retcode: 128
[ERROR   ] Failed to check remote refs: Unable to authenticate using identity file:

error: cannot run .tmp/thin/py2/salt/templates/git/ssh-id-wrapper: No such file or directory
fatal: unable to fork
```

### Previous Behavior

### New Behavior

### Tests written?

Yes/No

### Commits signed with GPG?

Yes/No

### DIFF
--- a/salt/modules/git.py
+++ b/salt/modules/git.py
@@ -265,10 +265,10 @@ def _git_run(command, cwd=None, user=None, password=None, identity=None,
 
             # copy wrapper to area accessible by ``runas`` user
             # currently no support in windows for wrapping git ssh
-            ssh_id_wrapper = os.path.join(
+            ssh_id_wrapper = os.path.abspath(os.path.join(
                 salt.utils.templates.TEMPLATE_DIRNAME,
                 'git/ssh-id-wrapper'
-            )
+            ))
             tmp_ssh_wrapper = None
             if salt.utils.platform.is_windows():
                 ssh_exe = _find_ssh_exe()


bbinet:
@rallytime Ok, will do

bbinet:
@rallytime done

terminalmage (approved):

terminalmage (commented):
Actually, I think we need to take a closer look here and make sure that `root_dir` can't be used to introduce a vulnerability.

bbinet:
@terminalmage But `root_dir` is a salt-master config param, so I guess the salt-master config file should be considered trusted (writable by admin only) ?

rallytime:
@thatch45 What do you think about this and @terminalmage's concerns?

thatch45:
I think that my main question here is that a relative root dir even works, that strikes me as a potential issue. I don't think that it will present a security issue.
All in all I am ok with this change, but I would rather pull support for a relative root dir.

bbinet:
@thatch45 relative root dir actually works perfectly (with this patch for version 2018.3): I'm using relative root dir in the following project: https://github.com/bbinet/salt-thin-setupify

I'm going to rebase this PR onto latest 2018.3 branch.

bbinet:
This is now rebased and ready to be merged.

cachedout (approved):

PR #48549, opened 2018-07-12
Add autoremove, force-removal-of-dependent-packages for opkg.remove

chotea:
What does this PR do?
This PR adds support for opkg remove in terms of giving the possibility of specifying two extra parameters: force-removal-of-dependent-packages, autoremove. It also contains unit tests for opkg module.

Previous Behavior
The implementation was always calling opkg remove without any extra parameters.

New Behavior
Now we check these 2 force-options in kwargs and append them to the command if necessary:

--force-removal-of-dependent-packages - Remove package and all dependencies
--autoremove - Remove packages that were installed automatically to
satisfy dependencies
Tests written?
Yes

Commits signed with GPG?
No

Old PR: https://github.com/saltstack/salt/pull/47240

### DIFF
--- a/salt/modules/opkg.py
+++ b/salt/modules/opkg.py
@@ -509,6 +509,15 @@ def remove(name=None, pkgs=None, **kwargs):  # pylint: disable=unused-argument
         A list of packages to delete. Must be passed as a python list. The
         ``name`` parameter will be ignored if this option is passed.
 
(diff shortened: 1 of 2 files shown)


rallytime (approved):
This looks great @chotea!

Thank you for this addition.

Issue #48318, opened 2018-06-26
state.sls_id executed locally with salt-ssh


xuhcc:
### Description of Issue/Question

`state.sls_id` executes `cmd.run` on local machine instead of remote when salt-ssh is used.

### Setup

maintenance.sls

```
test_state:
   cmd.run:
    - name: 'uname -a'

```

### Steps to Reproduce Issue

```
salt-ssh 'my-server' state.sls_id test_state maintenance
```

Returns local system info.

### Versions Report

```
$ salt --versions-report
Salt Version:
           Salt: 2018.3.1
 
Dependency Versions:
           cffi: 1.11.5
       cherrypy: Not Installed
       dateutil: 2.7.3
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.10
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: 0.30.1
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.5.6
   mysql-python: Not Installed
      pycparser: 2.18
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: Not Installed
         Python: 2.7.15 (default, May  1 2018, 20:16:04)
   python-gnupg: Not Installed
         PyYAML: 3.12
          PyZMQ: 16.0.3
           RAET: Not Installed
          smmap: Not Installed
        timelib: Not Installed
        Tornado: 4.5.3
            ZMQ: 4.2.2
 
System Versions:
           dist:   
         locale: UTF-8
        machine: x86_64
        release: 4.17.2-1-ARCH
         system: Linux
        version: Not Installed
```

Ch3LL:
i'll assign this to myself and take a look next week :)

Ch3LL:
apologies needed to bump this issue to this week.

PR #48555, opened 2018-07-12
Fix state.sls_id not running on ssh minion

Ch3LL:
### What does this PR do?
`state.sls_id` was running on the master and not the salt-ssh minion. This PR makes sure we are running the state on the salt-ssh minion. I noticed we are calling a lot of the same functions in the state.py wrapper for salt-ssh so I created this funciton: `_ssh_state`. I hope to move the other ones to this but in develop to make sure we don't ruin something in a point release once this is merged forward

https://github.com/saltstack/salt/issues/48318

### Tests written?

Yes

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/client/ssh/wrapper/state.py
+++ b/salt/client/ssh/wrapper/state.py
@@ -31,6 +31,56 @@ __func_alias__ = {
 }
 log = logging.getLogger(__name__)
 
+def _ssh_state(chunks, __opts__, __context__, __pillar__, __salt__, st_kwargs,
+              kwargs, test=False):
+    file_refs = salt.client.ssh.state.lowstate_file_refs(
+            chunks,
+            _merge_extra_filerefs(
+                kwargs.get('extra_filerefs', ''),
+                __opts__.get('extra_filerefs', '')
+                )
+            )
+    # Create the tar containing the state pkg and relevant files.
+    trans_tar = salt.client.ssh.state.prep_trans_tar(
+            __opts__,
+            __context__['fileclient'],
+            chunks,
+            file_refs,
+            __pillar__,
+            st_kwargs['id_'])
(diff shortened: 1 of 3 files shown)


rallytime (approved):
LGTM

gtmanfred (approved):

DmitryKuzmenko (approved):

salt/client/ssh/wrapper/state.py
```diff
@@ -32,6 +32,60 @@
 log = logging.getLogger(__name__)
 
 
+def _ssh_state(chunks, __opts__, __context__, __pillar__, __salt__, st_kwargs,
```
cachedout: I'm very confused by this. Why are we taking in `__opts__` and `__context__` as arguments for this function when they should already be available in this namespace?

salt/client/ssh/wrapper/state.py
```diff
@@ -32,6 +32,60 @@
...
+    file_refs = salt.client.ssh.state.lowstate_file_refs(
+            chunks,
+            _merge_extra_filerefs(
+                kwargs.get('extra_filerefs', ''),
+                __opts__.get('extra_filerefs', '')
+                )
+            )
+    # Create the tar containing the state pkg and relevant files.
+    trans_tar = salt.client.ssh.state.prep_trans_tar(
+            __opts__,
```
isbm: @Ch3LL  This is unused inside the function. I'd propose either remove the option entirely from `prep_trans_tar` or refactor that it is used.

salt/client/ssh/wrapper/state.py
isbm: And `__pillar__`. Can we fix this, @Ch3LL ?

PR #48562, opened 2018-07-12
Add timeouts to all s3 queries

basepi:
### What does this PR do?

Add timeouts to all s3 queries

n/a

### Previous Behavior

According to http://docs.python-requests.org/en/master/user/advanced/#timeouts:

> By default, requests do not time out unless a timeout value is set explicitly.

This is undesirable behavior.

### New Behavior

Will timeout after 5 minutes. This would be pretty trivial to make into a config option, but for now I figured an insanely long timeout is better than none.

### Tests written?

No

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/utils/s3.py
+++ b/salt/utils/s3.py
@@ -162,20 +162,23 @@ def query(key, keyid, method='GET', params=None, headers=None,
                                       headers=headers,
                                       data=data,
                                       verify=verify_ssl,
-                                      stream=True)
+                                      stream=True,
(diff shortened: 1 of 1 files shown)


gtmanfred (approved):

rallytime (approved):

basepi:
@rallytime I recommend that this be backported to all versions. Forgot to mention that in the original PR.

rallytime:
Sure thing @basepi

PR #48564, opened 2018-07-13
Fix failing prepend test

dwoz:
The prepend test is failing when jenkins runs kithchen but is passing in
my test environment. Make sure each test starts without the 'etc' path
existing.

### Tests written?

No

### Commits signed with GPG?

Yes

### DIFF
--- a/tests/unit/states/test_file.py
+++ b/tests/unit/states/test_file.py
@@ -61,6 +61,15 @@ class TestFileState(TestCase, LoaderModuleMockMixin):
             }
         }
 
+    def tearDown(self):
+        remove_dir = '/etc'
+        if salt.utils.is_windows():
+            remove_dir = 'c:\\etc'
+        try:
+            os.remove(remove_dir)
+        except OSError:
+            pass
+
     def test_serialize(self):
         def returner(contents, *args, **kwargs):
             returner.returned = contents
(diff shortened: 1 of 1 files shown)


gtmanfred (commented):
Just one question

tests/unit/states/test_file.py
```diff
@@ -61,6 +61,15 @@ def setup_loader_modules(self):
             }
         }
 
+    def tearDown(self):
+        remove_dir = '/etc'
+        if salt.utils.is_windows():
+            remove_dir = 'c:\\etc'
```
gtmanfred: This could be one line

tests/unit/states/test_file.py
```diff
@@ -1174,6 +1183,7 @@ def test_prepend(self):
         '''
         Test to ensure that some text appears at the beginning of a file.
         '''
```
gtmanfred: Does this work on linux?

dwoz: good catch.

gtmanfred: The tests looked like they passed, so I merged it.

dwoz: Oh, right. They would pass because 'c:\\etc' wont ever exist on linux. It's still buggy though so I'll clean it up in another PR (after making sure test_prepend actually passes now).

PR #48566, opened 2018-07-13
Fix #44954

### LINKED ISSUE #44954: SDB REST Module does not work without a URI Query Parameter
REST SDB module cannot be used without specifying a request parameter. The documentation indicates that the SDB URI can be used without requiring any query parameters and the SDB REST configuration allows for configuration without a template variable. When attempting to retrieve the value without specifying a query parameter, the following error is given:

```
Exception occurred in runner sdb.get: Traceback (most recent call last):
  File "/usr/lib/python2.7/dist-packages/salt/client/mixins.py", line 382, in _low
  File "/usr/lib/python2.7/dist-packages/salt/runners/sdb.py", line 27, in get
    return salt.utils.sdb.sdb_get(uri, __opts__, __utils__)
  File "/usr/lib/python2.7/dist-packages/salt/utils/sdb.py", line 46, in sdb_get
    return loaded_db[fun](query, profile=profile)
  File "/usr/lib/python2.7/dist-packages/salt/sdb/rest.py", line 94, in get
    return query(key, None, service, profile)
  File "/usr/lib/python2.7/dist-packages/salt/sdb/rest.py", line 115, in query
    for pair in comps[1].split('&'):
IndexError: list index out of range
```

Looking at the code for '/salt/sdb/rest.py', the logic is such that it always assumes there is a value from performing the `split('?')` and does not perform any safety checks to ensure that it is not there:

```
'''
    Get a value from the REST interface
    '''
    comps = key.split('?')
    key = comps[0]
    key_vars = {}
    for pair in comps[1].split('&'):
        pair_key, pair_val = pair.split('=')
        key_vars[pair_key] = pair_val
```

In order to get the SDB value to return properly, you have to pass any arbitrary query string to the sdb URI in order for the value to return:

Assuming the SDB configuration in the Setup section below:
This does not work: `salt-run sdb.get sdb://my-rest-api/urls`
This works: `salt-run sdb.get sdb://my-rest-api/urls?foo=bar`

### Setup
SDB Rest configuration from the rest example:

/etc/salt/master.d/sdb-github.sls
```
my-rest-api:
  driver: rest
  urls:
    url: https://api.github.com/
  keys:
    url: https://api.github.com/users/{{user}}/keys
    backend: requests
```

### Steps to Reproduce Issue
1. Add the above SDB configuration to /etc/salt/master.d/github.sls
2. Execute the following Salt Run:
```
salt-run sdb.get sdb://my-rest-api/urls
```

### Versions Report
```
Salt Version:
           Salt: 2017.7.2

Dependency Versions:
           cffi: Not Installed
       cherrypy: 3.5.0
       dateutil: 2.4.2
      docker-py: Not Installed
          gitdb: 0.6.4
      gitpython: 1.0.1
          ioflo: Not Installed
         Jinja2: 2.8
        libgit2: Not Installed
        libnacl: 1.6.1
       M2Crypto: Not Installed
           Mako: 1.0.3
   msgpack-pure: Not Installed
 msgpack-python: 0.4.6
   mysql-python: Not Installed
      pycparser: Not Installed
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: Not Installed
         Python: 2.7.12 (default, Nov 19 2016, 06:48:10)
   python-gnupg: Not Installed
         PyYAML: 3.11
          PyZMQ: 15.2.0
           RAET: Not Installed
          smmap: 0.9.0
        timelib: Not Installed
        Tornado: 4.2.1
            ZMQ: 4.1.4

System Versions:
           dist: Ubuntu 16.04 xenial
         locale: UTF-8
        machine: x86_64
        release: 4.4.0-87-generic
         system: Linux
        version: Ubuntu 16.04 xenial

```

Poil:
Try to fix 
#44954

### DIFF
--- a/salt/sdb/rest.py
+++ b/salt/sdb/rest.py
@@ -98,12 +98,15 @@ def query(key, value=None, service=None, profile=None):  # pylint: disable=W0613
     '''
     Get a value from the REST interface
     '''
-    comps = key.split('?')
-    key = comps[0]
-    key_vars = {}
-    for pair in comps[1].split('&'):
-        pair_key, pair_val = pair.split('=')
-        key_vars[pair_key] = pair_val
+    try:
+        key, query_string = key.split('?')
+        key_vars = {}
+        if query_string:
+           for pair in query_string.split('&'):
+               pair_key, pair_val = pair.split('=')
+               key_vars[pair_key] = pair_val
+    except ValueError:
+        key_vars = {}
 
     renderer = __opts__.get('renderer', 'jinja|yaml')
     rend = salt.loader.render(__opts__, {})


salt/sdb/rest.py
```diff
@@ -98,12 +98,15 @@ def query(key, value=None, service=None, profile=None):  # pylint: disable=W0613
...
-        pair_key, pair_val = pair.split('=')
-        key_vars[pair_key] = pair_val
+    try:
+        key, query_string = key.split('?')
+        key_vars = {}
+        if query_string:
+           for pair in query_string.split('&'):
+               pair_key, pair_val = pair.split('=')
+               key_vars[pair_key] = pair_val
+    except ValueError:
```
rares-pop: do you need this? 

why not use urllib and get rid of the custom parsing?

Poil: Not sure I've changed nothing here, it was before :)

rares-pop: I asked about the try/except? That wasn't there before, was it? Why do you need them?

But as I suggested earlier, why not replace the entire functionality custom with few urllib calls?

Poil: Today we raise an error when the given url have no query string

rares-pop: So, what inside the try/except you added will raise the ValueError you are trying to handle?

cachedout: I have the same question at @rares-pop here. I'm failing to see how this `except` ever gets hit. This sort of looks like something that was written to catch what would have been an `IndexError` in the old code, perhaps?

Poil: Yes it is only to fix when you give a SDB URL without a query string.
I'll try to migrate this module to parse the URL via urllib if I have some times at work this week.

rares-pop: Aren't you protected from that by this: if query_string: on line 104?

Poil: Today I've a stack trace so no :)

rallytime: I think there is a simpler way to handle this, rather than wrapping the whole thing in a try/except, which can be difficult to reason around when wrapping blocks of code like this. For example, the issue referenced above is looking for an `IndexError`, not a `KeyError`. I'd rather we be more precise about the try/except.

Instead, what about just checking for a `?` in the `key` first?

cachedout:
@Poil Did you see that last suggestion from @rallytime ?

rallytime:
Since we haven't heard back here for a bit, I am closing this. We will gladly open this again when you are ready to come back to it and implement the feedback above. Thanks!

PR #48567, opened 2018-07-13
Add OpenEmbedded ptest output support

rares-pop:
OpenEmbedded has its own integrated test infrastructure called ptest:
https://wiki.yoctoproject.org/wiki/Ptest

Add the `--ptest-out` option to `salt/tests/runtests.py` to
enable outputting in a format familiar to ptest.

Signed-off-by: Sergey Kizunov <sergey.kizunov@ni.com>

### Tests written?

No

### Commits signed with GPG?

Yes


tests/support/ptest.py
```diff
@@ -0,0 +1,96 @@
+# -*- coding: utf-8 -*-
+
+'''
+For use with OpenEmbedded ptest.
+Based on the API in Python's TextTestRunner.
+'''
+
+# Import Python libs
+from __future__ import absolute_import, print_function
```
rallytime: We should have `unicode_literals` here as well.

gtmanfred (approved):
I am ok with this, just know that we are in the process of moving away from runtests.py, and moving towards using pytest.

You can currently run the pytest testing setup using `tox -e <env>`

rallytime (approved):

cachedout:
I want to second @gtmanfred's point here. This runner will eventually be taken out of service. I have no problem merging this but please be aware that it won't be maintained in the future and may well be removed entirely.

rares-pop:
So, what would be the timeframe for moving to pytest?

Since our distribution is OpenEmbedded based, is using ptest as the main testing runner.

gtmanfred:
Well, we have just switched over to using our new jenkins, and that is what was taking up most of my time.

I am looking at working on moving over once the Fluorine release is branched in August.

PR #48585, opened 2018-07-13
replaced meta tag

astorath:
### What does this PR do?
Fix vault meta attribute as of documentation: https://www.vaultproject.io/api/auth/token/index.html#meta

### Tests written?
No

### Commits signed with GPG?
No

PR #48586, opened 2018-07-13
added named roles feature for module.vault

astorath:
### What does this PR do?
Added an option to define named role for creation of vault tokens. This is the preferred way to control access permissions and available policies of secondary minion tokens: https://www.vaultproject.io/api/auth/token/index.html#create-token

### Previous Behavior
Child token can have any policies main token have (this is only controlled by vault security mechanisms). Child token can have an option to create other tokens (as parent token does), leading to security risks.
Salt master token must have all the access policies minions have.

### New Behavior
User can define specific token named role for minion created tokens and explicitly define its behavior and access policies. Example: https://www.nomadproject.io/docs/vault-integration/index.html#vault-token-role-configuration

### Tests written?
Yes

### Commits signed with GPG?
No


salt/runners/vault.py
```diff
@@ -63,7 +63,12 @@ def generate_token(minion_id, signature, impersonated_by_master=False):
                     return {'error': response.reason}
                 config['auth']['token'] = response.json()['auth']['client_token']
 
-        url = '{0}/v1/auth/token/create'.format(config['url'])
+        role_name = config.get('role_name', None)
+        base_url = config['url']
+        if role_name is None
```
rallytime: Missing a `:` at the end of this line.

salt/runners/vault.py
```diff
@@ -63,7 +63,12 @@ def generate_token(minion_id, signature, impersonated_by_master=False):
...
+            url = '{0}/v1/auth/token/create'.format(base_url)
+        else
+            url = '{0}/v1/auth/token/create/{1}'.format(base_url, role_name)
```
rallytime: I think this whole block could be simplified:
```
url = '{0}/v1/auth/token/create'.format(config['url'])
role_name = config.get('role_name')
if role_name:
    url = url + '/' + role_name
```
What do you think?

salt/runners/vault.py
```diff
@@ -73,7 +78,7 @@ def generate_token(minion_id, signature, impersonated_by_master=False):
         payload = {
                     'policies': _get_policies(minion_id, config),
                     'num_uses': 1,
-                    'metadata': audit_data
+                    'meta': audit_data
```
rallytime: Can you explain why this key is changing? I didn't see a mention of this in your PR explanation. (Though I certainly could have missed it.)

rallytime (changes requested):
Thanks for submitting this. I have some python sytle/syntax things that need to be fixed.

I'd also like @gtmanfred to look at this since he is more familiar with the vault module than I am.

salt/runners/vault.py
gtmanfred: seconded

salt/runners/vault.py
gtmanfred: I am not sure why it was ever metadata, because the docs for vault v1 say `meta` https://www.vaultproject.io/api/auth/token/index.html#meta

gtmanfred (approved):
looks good to me.

salt/runners/vault.py
rallytime: Oh, this was mentioned/changed in this PR #48585.

salt/runners/vault.py
cachedout: Not that it matters too much but do we run a risk of doubling-up on a slash after URL using this method?

gtmanfred: We control the while url building in this case, the user cannot pass anything in that would cause an extra slash.

astorath:
@rallytime Sure, I just want to update docs and tests a bit.

cachedout:
@rallytime Re-review requested.

astorath:
@rallytime @cachedout, I've fixed linting errors and added some tests and docs. I don't know which version to address in docs though.

rallytime (approved):

rallytime:
Thanks @astorath!

Issue #46236, opened 2018-02-28
WIN_LGPO requires unsetting certain policies before configuring new policies


doesitblend:
### Description of Issue/Question
When trying to use the `lgpo.set` state I get the following error if some Group Policies are set (listed in Steps to Produce):
```
# sudo salt 'minion1' state.apply config_win_updates
minion1:
----------
          ID: My Local Group Policy
    Function: lgpo.set
      Result: False
     Comment: An exception occurred in this state: Traceback (most recent call last):
                File "C:\salt\bin\lib\site-packages\salt\state.py", line 1851, in call
                  **cdata['kwargs'])
                File "C:\salt\bin\lib\site-packages\salt\loader.py", line 1795, in wrapper
                  return f(*args, **kwargs)
                File "C:\salt\bin\lib\site-packages\salt\states\win_lgpo.py", line 301, in set_
                  adml_language=adml_language)
                File "C:\salt\bin\lib\site-packages\salt\modules\win_lgpo.py", line 5525, in set_
                  raise CommandExecutionError(msg)
              CommandExecutionError: Error while attempting to write Administrative Template Policy data.  Some changes may not be applied as expected
     Started: 10:30:07.663000
    Duration: 93120.0 ms
     Changes:   

Summary for minion1
------------
Succeeded: 0
Failed:    1
------------
Total states run:     1
Total run time:  93.120 s
```

The request is for the module to not fail when these settings are pre-configured. I thought of just adding these settings to my state to have the state set them, but they are not available for configuration in the `win_lgpo.py` module. 

### Setup
Configure a master on 2017.7.3 (Any Linux OS)
Configure a 2017.7.3 Python 2 AMD64 minion on Windows 10/2016

### Steps to Reproduce Issue
Manually configure any of the settings in this list to anything but "Not Configured". From what I can tell it can be any mix of these, but as long as at least one of them is configured the error will occur:
```
Prevent installation of devices using drivers 
Prevent installation of devices that match any of these device IDs 
Join Microsoft MAPS
Send file samples when further analysis is required
Let Windows apps access trusted devices
Let Windows apps access location
Let Windows apps access messaging
Let Windows apps access account information
Let Windows apps access calendar
Let Windows apps access motion
Let Windows apps access contacts
Let Windows apps access email
Let Windows apps access camera
Boot-Start Driver Initialization Policy
Configure Windows Defender Smartscreen
Prevent bypassing Windows Defender SmartScreen prompts for sites
[...]

PR #48587, opened 2018-07-13
Fix lgpo issue on Py3

twangboy:
### What does this PR do?
Open the gpt.ini file in normal mode (not binary)
Use log.exception so you can see where the actual problems are
Add TODOs to make the exceptions more specific

https://github.com/saltstack/salt/issues/46236

### Previous Behavior
Worked on Py2, but not Py3

### New Behavior
Works on Both

### Tests written?
No

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/modules/win_lgpo.py
+++ b/salt/modules/win_lgpo.py
@@ -4258,8 +4258,8 @@ def _write_regpol_data(data_to_write,
                             16),
                         gpt_ini_data[general_location.end():])
             if gpt_ini_data:
-                with salt.utils.files.fopen(gpt_ini_path, 'wb') as gpt_file:
-                    gpt_file.write(salt.utils.stringutils.to_bytes(gpt_ini_data))
+                with salt.utils.files.fopen(gpt_ini_path, 'w') as gpt_file:
+                    gpt_file.write(gpt_ini_data)
         # TODO: This needs to be more specific
         except Exception as e:
             msg = 'An error occurred attempting to write to {0}, the exception was {1}'.format(
--- a/salt/modules/win_lgpo.py
+++ b/salt/modules/win_lgpo.py
@@ -2588,7 +2588,9 @@ class _policy_info(object):
                     userSid = '{1}\\{0}'.format(userSid[0], userSid[1])
                 else:
                     userSid = '{0}'.format(userSid[0])
+            # TODO: This needs to be more specific
             except Exception:
+                log.exception('Handle this explicitly')
                 userSid = win32security.ConvertSidToStringSid(_sid)
             usernames.append(userSid)
         return usernames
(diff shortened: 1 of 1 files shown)


dwoz (approved):

salt/modules/win_lgpo.py
```diff
@@ -4237,20 +4250,26 @@ def _write_regpol_data(data_to_write,
...
+                gpt_ini_data = '{0}{1}={2}\r\n{3}'.format(
                         gpt_ini_data[general_location.start():general_location.end()],
                         'Version',
-                        int("{0}{1}".format(six.text_type(version_nums[0]).zfill(4), six.text_type(version_nums[1]).zfill(4)), 16),
+                        int("{0}{1}".format(six.text_type(version_nums[0]).zfill(4),
+                                            six.text_type(version_nums[1]).zfill(4)),
+                            16),
                         gpt_ini_data[general_location.end():])
             if gpt_ini_data:
                 with salt.utils.files.fopen(gpt_ini_path, 'wb') as gpt_file:
```
lomeroe: @twangboy should the file be written in ASCII mode as well, since it is not being read in binary mode?

lomeroe (approved):
looks good to me visually, haven't had a chance to fully test it

salt/modules/win_lgpo.py
```diff
@@ -2588,7 +2588,9 @@ def _sidConversion(cls, val, **kwargs):
                     userSid = '{1}\\{0}'.format(userSid[0], userSid[1])
                 else:
                     userSid = '{0}'.format(userSid[0])
+            # TODO: This needs to be more specific
```
rallytime: Is there a reason we're adding `TODO`s here instead of just making these all more specific now?

rallytime (changes requested):

salt/modules/win_lgpo.py
twangboy: Because I don't know what the actual failures may be.

dwoz: If we had some automated tests for this then I think we could remove the blanket exception handling all together.

rallytime: I guess I don't really get what this is adding. We're just logging a line that we hit the exception, and in many cases handling and then moving on. I don't see how that is very useful to the end user.

twangboy: `log.exception` gives us the actual location where the exception is occurring. Otherwise, the rest of the code catches the exceptions as they bubble up and it makes it hard to find where the actual issue is.

rallytime: OK, thanks for the explanation!

rallytime (approved):

Issue #48415, opened 2018-07-02
Event.send short-circuiting in multi-master mode


doesitblend:
### Description of Issue/Question
In a multi-master setup `event.send` appears to just go through the list of masters, open a channel and fire an event to that master's bus. This continues to work when one of these masters is down, logging `[DEBUG   ] SaltReqTimeoutError, retrying. (1/7)`. However, in the case where any one of these masters has not yet accepted the minion key, this process will die at whatever master has not accepted the key instead of continuing through the list of masters. 

The event appears on all the masters prior to the failure, but the overall command back from the CLI fails.

I've attached the output of my trace log from the CLI for reference. The first log will show that the second master is not attempted to be connected to. The second log shows that the command failed at the second master but sent the event to the first one.
[salt-call_trace_eventsend.log](https://github.com/saltstack/salt/files/2156857/salt-call_trace_eventsend.log)
[salt-call_trace_eventsend_master2_fail.log](https://github.com/saltstack/salt/files/2156868/salt-call_trace_eventsend_master2_fail.log)

My testing was primarily done in 2017.7.5, but I have also verified this is 2017.7.6.

### Setup
I'm using three Docker containers, one minion and two masters. I've installed the code with the bootstrap script on each container.

#### Minion Config
```
master:
  -  172.17.0.2
  -  172.17.0.4
```

### Steps to Reproduce Issue
* Connect minion to two masters
* Accept key on one of the masters
* Leave the key in unaccepted state on the other master
* From the minion run, `salt-call -ltrace event.send 'myco/mytag/someminion/testing4' "{'mydata':'this is neat'}" `

### Versions Report

```
Salt Version:
           Salt: 2017.7.5
 
Dependency Versions:
           cffi: Not Installed
       cherrypy: Not Installed
       dateutil: Not Installed
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.7.2
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.5.1
   mysql-python: Not Installed
      pycparser: Not Installed
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: Not Installed
         Python: 2.7.5 (default, Apr 11 2018, 07:36:10)
   python-gnupg: Not Installed
         PyYAML: 3.11
[...]

PR #48588, opened 2018-07-13
[2017.7] event send multi master

garethgreenaway:
### What does this PR do?
When using Salt multi-master, if we encounter a salt master that has not accepted the minion key yet we should not exit right away, rather continue on and try the next salt master available in the list.  If the channel.send is successful, ensure that `ret` is True in case previous attempts at contacting masters had set it to False.

#48415 

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/crypt.py
+++ b/salt/crypt.py
@@ -520,9 +520,18 @@ class AsyncAuth(object):
                     error = SaltClientError('Detect mode is on')
                     break
                 if self.opts.get('caller'):
-                    print('Minion failed to authenticate with the master, '
-                          'has the minion key been accepted?')
-                    sys.exit(2)
+                    # We have a list of masters, so we should break
+                    # and try the next one in the list.
+                    if self.opts.get('local_masters', None):
+                        log.debug('=== opts %s ===', self.opts['master'])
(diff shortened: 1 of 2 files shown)


rallytime (approved):
This is nice.

salt/crypt.py
```diff
@@ -520,9 +520,17 @@ def _authenticate(self):
                     error = SaltClientError('Detect mode is on')
                     break
                 if self.opts.get('caller'):
-                    print('Minion failed to authenticate with the master, '
-                          'has the minion key been accepted?')
-                    sys.exit(2)
+                    # We have a list of masters, so we should break
+                    # and try the next one in the list.
+                    if self.opts.get('local_masters', None):
```
cachedout: This is nit-picky but this is such a sensitive area that I would feel better if we checked explicitly against a list type.

cachedout: Oh! Oh! My fault. You're totally right.

mattp-:
should this logic also be applied at https://github.com/garethgreenaway/salt/blob/fab25af1a975e0dd6a44ad28c52c6a3e90f9fa62/salt/crypt.py#L1099 ?

PR #48605, opened 2018-07-16
acme: Add support for the CloudFlare DNS plugin

kyrias:
### What does this PR do?

Adds support for the CloudFlare DNS plugin to the acme module.

None.

### Tests written?

No

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/modules/acme.py
+++ b/salt/modules/acme.py
@@ -23,6 +23,16 @@ eventually falls back to /opt/letsencrypt/letsencrypt-auto
 
 Most parameters will fall back to cli.ini defaults if None is given.
 
+DNS plugins
+-----------
+
+This module currently supports the CloudFlare certbot DNS plugin.  The DNS
+plugin credentials file needs to be passed in using the
+``dns_plugin_credentials`` argument.
+
+Make sure the appropriate certbot plugin for the wanted DNS provider is
+installed before using this module.
+
 '''
 # Import python libs
 from __future__ import absolute_import, print_function, unicode_literals
--- a/salt/modules/acme.py
+++ b/salt/modules/acme.py
@@ -107,7 +107,9 @@ def cert(name,
          tls_sni_01_port=None,
          tls_sni_01_address=None,
          http_01_port=None,
-         http_01_address=None):
+         http_01_address=None,
+         dns_plugin=None,
+         dns_plugin_credentials=None):
     '''
     Obtain/renew a certificate from an ACME CA, probably Let's Encrypt.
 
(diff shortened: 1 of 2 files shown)


salt/modules/acme.py
```diff
@@ -114,6 +116,8 @@ def cert(name,
     :param aliases: subjectAltNames (Additional DNS names on certificate)
     :param email: e-mail address for interaction with ACME provider
     :param webroot: True or a full path to use to use webroot. Otherwise use standalone mode
+    :param dns_plugin: Name of a DNS plugin to use
+    :param dns_cloudflare_credentials: Path to CloudFlare credentials file if the CloudFlare DNS Plugin is in use
```
cachedout: I'm not a huge fan of adding an argument that's specific to a single provider. Can we make this something like `dns_plugin_credentials`?

salt/states/acme.py
```diff
@@ -62,6 +64,8 @@ def cert(name,
     :param aliases: subjectAltNames (Additional DNS names on certificate)
     :param email: e-mail address for interaction with ACME provider
     :param webroot: True or a full path to webroot. Otherwise use standalone mode
+    :param dns_plugin: Name of a DNS plugin to use
+    :param dns_cloudflare_credentials: Path to CloudFlare credentials file if the CloudFlare DNS Plugin is in use
```
cachedout: Same comment here as above.

kyrias:
Good idea, done.

salt/modules/acme.py
```diff
@@ -95,6 +95,8 @@ def cert(name,
          aliases=None,
          email=None,
          webroot=None,
+         dns_plugin=None,
```
rallytime: These new kwargs need to go at the end of the list so we don't cause an API change and break things for people if they're using positional arguments.

rallytime (changes requested):
Thanks @kyrias - One request from me. :)

salt/modules/acme.py
kyrias: Done.

salt/modules/acme.py
```diff
@@ -133,6 +135,8 @@ def cert(name,
                          the port Certbot listens on. A conforming ACME server
                          will still attempt to connect on port 80.
     :param https_01_address: The address the server listens to during http-01 challenge.
+    :param dns_plugin: Name of a DNS plugin to use
```
cachedout: We need documentation somewhere on possible values for this or at least some examples. Otherwise, nobody will know how to use this. :) Same for the plugin credentials.

rallytime:
Hi @kyrias - Can you add some docs for this? Then we can review this again.

kyrias:
I've added a short note to the module top-level doc comment.  Will something like this be acceptable?

rallytime (approved):
LGTM

rallytime:
@kyrias Oh, there is a pylint failure on this change. Can you fix it?

https://jenkinsci.saltstack.com/job/pr-lint/job/PR-48605/3/warnings52Result/

kyrias:
Hm, so it doesn't like `{!r}`.  I used it there so that it would be escaped properly in case the string contains some strange characters.  Replaced it with a dirtier `\'{0}\'` now.

PR #48617, opened 2018-07-17
modules/nilrt_ip.py: Add EtherCAT Support

alexvasiu:
* modules/nilrt_ip.py: Add EtherCAT Support

The possibilities for adapter mode are: `TCP/IP`, `EhterCAT` and `Disabled`.

Signed-off-by: Ovidiu-Adrian Vancea <ovidiu.vancea@ni.com>

* modules/nilrt_ip.py: Fix configparser compatibility

read_file is available only since python 3.2 so readfp should
be used instead

### DIFF
--- a/salt/modules/nilrt_ip.py
+++ b/salt/modules/nilrt_ip.py
@@ -274,7 +274,7 @@ def _load_config(section, options, default_value='', file=INI_FILE):
         return results
     with salt.utils.files.fopen(file, 'r') as config_file:
         config_parser = configparser.RawConfigParser(dict_type=CaseInsensitiveDict)
-        config_parser.read_file(config_file)
+        config_parser.readfp(config_file)
         for option in options:
             if six.PY2:
                 results[option] = _remove_quotes(config_parser.get(section, option)) \
@@ -789,7 +789,7 @@ def _configure_static_interface(interface, **settings):
     if os.path.exists(INTERFACES_CONFIG):
         try:
             with salt.utils.files.fopen(INTERFACES_CONFIG, 'r') as config_file:
-                parser.read_file(config_file)
+                parser.readfp(config_file)
         except configparser.MissingSectionHeaderError:
             pass
     hwaddr = interface.hwaddr[:-1]
--- a/salt/modules/nilrt_ip.py
+++ b/salt/modules/nilrt_ip.py
@@ -284,6 +284,57 @@ def _get_requestmode_info(interface):
     return 'dhcp_linklocal'
 
 
+def _get_adaptermode_info(interface):
+    '''
+    return adaptermode for given interface
+    '''
+    with salt.utils.files.fopen(INI_FILE, 'r') as config_file:
+        config_parser = configparser.RawConfigParser(dict_type=CaseInsensitiveDict)
+        config_parser.read_file(config_file)
+        mode = '' if not config_parser.has_option(interface, 'mode') else \
+            _remove_quotes(config_parser.get(interface, 'mode').lower())
+    return mode if mode in ['disabled', 'ethercat'] else 'tcpip'
+
+
+def _get_possible_adaptermodes(interface, blacklist):
+    '''
+    Return possible adapter modes for a given interface using a blacklist.
+
+    :param interface: interface name
+    :param blacklist: given blacklist
+    :return: list of possible adapter modes
+    '''
+    adapter_modes = []
+
+    with salt.utils.files.fopen(INI_FILE, 'r') as config_file:
+        config_parser = configparser.RawConfigParser(dict_type=CaseInsensitiveDict)
+        config_parser.read_file(config_file)
+        protocols = _remove_quotes(config_parser.get('lvrt', 'AdditionalNetworkProtocols').lower()) if \
+            config_parser.has_option('lvrt', 'AdditionalNetworkProtocols') else ''
+    sys_interface_path = os.readlink('/sys/class/net/{0}'.format(interface))
+    with salt.utils.files.fopen('/sys/class/net/{0}/uevent'.format(interface)) as uevent_file:
+        uevent_lines = uevent_file.readlines()
+    uevent_devtype = ""
+    for line in uevent_lines:
+        if line.startswith("DEVTYPE="):
+            uevent_devtype = line.split('=')[1].strip()
+            break
+
+    for adapter_mode in blacklist:
+        if adapter_mode == '_':
+            continue
+        value = blacklist.get(adapter_mode, {})
+        if value.get('additional_protocol') and adapter_mode not in protocols:
+            continue
+
+        if interface not in value['name'] \
+            and not any((blacklist['_'][iface_type] == 'sys' and iface_type in sys_interface_path) or
+                        (blacklist['_'][iface_type] == 'uevent' and iface_type == uevent_devtype)
+                        for iface_type in value['type']):
+            adapter_modes += [adapter_mode]
+    return adapter_modes
+
+
 def _get_static_info(interface):
     '''
     Return information about an interface from config file.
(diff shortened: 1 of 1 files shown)


salt/modules/nilrt_ip.py
```diff
@@ -284,6 +284,57 @@ def _get_requestmode_info(interface):
...
 
+def _get_adaptermode_info(interface):
+    '''
+    return adaptermode for given interface
+    '''
+    with salt.utils.files.fopen(INI_FILE, 'r') as config_file:
+        config_parser = configparser.RawConfigParser(dict_type=CaseInsensitiveDict)
+        config_parser.read_file(config_file)
+        mode = '' if not config_parser.has_option(interface, 'mode') else \
+            _remove_quotes(config_parser.get(interface, 'mode').lower())
```
rares-pop: This seems to be an anti-pattern - doing an 'has_option' then a 'get()'. Can we use a default for the get() and do only that?

More than that, the first 5 lines seems to be a pattern that is repeated throughout the code. Can/Should we create a higher-level wrapper over the configparser to hide all this boilerplate code?

salt/modules/nilrt_ip.py
```diff
@@ -515,10 +631,40 @@ def _persist_config(section, token, value):
...
+
+    CLI Example:
+
+    .. code-block:: bash
+
+        salt '*' ip.set_ethercat interface-label master-id
+    '''
+    if __grains__['lsb_distrib_id'] == 'nilrt':
+        initial_mode = _get_adaptermode_info(interface)
+        _set_adapter_mode(interface, 'EtherCAT')
```
rares-pop: why you spell 'EtherCAT' here, but do this in other places: if initial_mode == 'ethercat':

salt/modules/nilrt_ip.py
alexvasiu: `get()` throws an exception if the option doesn't exist 

We can make a class or a global variable where the config will be read only one time, but the problem is that if we write something with `nirtcfg` then we should reload config from **INI_FILE** in configparser, so a method like this one will be ok:
```python
def _get_option_configparser(section, option, default_value='', file=INI_FILE):
        with salt.utils.files.fopen(file, 'r') as config_file:
            config_parser = configparser.RawConfigParser(dict_type=CaseInsensitiveDict)
            config_parser.read_file(config_file)
            if config_parser.has_section(section) and config_parser.has_option(option):
                return config_parser.get(section, option)
            return default_value
```

salt/modules/nilrt_ip.py
```diff
@@ -530,10 +676,15 @@ def set_dhcp_linklocal_all(interface):
...
+        initial_mode = _get_adaptermode_info(interface)
+        _set_adapter_mode(interface, 'TCPIP')
         _persist_config(interface, 'dhcpenabled', '1')
         _persist_config(interface, 'linklocalenabled', '1')
-        disable(interface)
-        enable(interface)
+        if initial_mode == 'ethercat':
+            __salt__['system.set_reboot_required_witnessed']()
+        else:
+            disable(interface)
```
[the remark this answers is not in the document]
alexvasiu: Yep :+1:

salt/modules/nilrt_ip.py
rares-pop: There should be some option to pass defaults to the get, so you can avoid the has_option/get pair.

We shouldn't need a class, but only a function that can take a list of keys and returns a dict with keys/values

salt/modules/nilrt_ip.py
alexvasiu: I will make a constant:
```python
NIRTCFG_ETHERCAT = 'EtherCAT'
```

salt/modules/nilrt_ip.py
```diff
@@ -258,45 +259,106 @@ def _remove_quotes(value):
...
+        if link_local_enabled == 0:
+            return 'static'
 
     # some versions of nirtcfg don't set the dhcpenabled/linklocalenabled variables
     # when selecting "DHCP or Link Local" from MAX, so return it by default to avoid
     # having the requestmode "None" because none of the conditions above matched.
     return 'dhcp_linklocal'
 
 
+def _get_adapter_mode_info(interface):
```
rares-pop: can this benefit from the newly created function? - _get_config_options

salt/modules/nilrt_ip.py
```diff
@@ -309,50 +371,86 @@ def _get_static_info(interface):
...
+        'adapter_mode': _get_adapter_mode_info(interface.name),
         'up': False,
         'ipv4': {
             'supportedrequestmodes': ['dhcp_linklocal', 'dhcp_only', 'linklocal_only', 'static'],
-            'requestmode': _get_requestmode_info(interface.name)
+            'requestmode': _get_request_mode_info(interface.name)
         },
         'hwaddr': interface.hwaddr[:-1]
     }
+    with salt.utils.files.fopen(INI_FILE, 'r') as config_file:
```
rares-pop: Is this used anywhere?

salt/modules/nilrt_ip.py
alexvasiu: Forgot about that function ... I'll change it

salt/modules/nilrt_ip.py
alexvasiu: I will delete that

salt/modules/nilrt_ip.py
```diff
@@ -309,50 +371,86 @@ def _get_static_info(interface):
...
+        config_parser = configparser.RawConfigParser(dict_type=CaseInsensitiveDict)
```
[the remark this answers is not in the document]
alexvasiu: Maybe `_save_config` for `_persist_config` and `_load_config` for `_get_config_options`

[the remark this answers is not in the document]
rares-pop: That makes sense

salt/modules/nilrt_ip.py
```diff
@@ -258,45 +259,102 @@ def _remove_quotes(value):
...
     # when selecting "DHCP or Link Local" from MAX, so return it by default to avoid
     # having the requestmode "None" because none of the conditions above matched.
     return 'dhcp_linklocal'
 
 
+def _get_adapter_mode_info(interface):
+    '''
+    return adaptermode for given interface
+    '''
+    mode = _load_config(interface, ['mode'])['mode']
```
alexvasiu: I didn't put `default_value = 'tcpip'` because the file may be corrupted and mode should be one of these values:
* **disabled**
* **ethercat**
* **tcpip** (default value)

rares-pop: this will raise if 'mode' is not in the file.

Should you go with:
mode = _load_config(interface, ['mode']).get('mode') ?

alexvasiu: If mode is not in file, _load_config will return the default_value for it:
```python
mode = {
    'mode': ''
}
```

salt/modules/nilrt_ip.py
```diff
@@ -469,10 +584,15 @@ def down(interface, iface_type=None):  # pylint: disable=unused-argument
         salt '*' ip.down interface-label
     '''
     if __grains__['lsb_distrib_id'] == 'nilrt':
-        out = __salt__['cmd.run_all']('ip link set {0} down'.format(interface))
-        if out['retcode'] != 0:
-            msg = 'Couldn\'t disable interface {0}. Error: {1}'.format(interface, out['stderr'])
-            raise salt.exceptions.CommandExecutionError(msg)
+        initial_mode = _get_adapter_mode_info(interface)
```
rares-pop: up/down code is almost identical. Can you create a worker to share the logic?

salt/modules/nilrt_ip.py
rares-pop: Ah, forgot about that.

salt/modules/nilrt_ip.py
alexvasiu: A function called `_change_state(state)` and state can be `up` or `down` ?

rares-pop: Looks good to me!

rares-pop:
Things look good to me. 

@gtmanfred can you take a look on these? They are something we'd need for Fluorine.

gtmanfred (approved):

rares-pop (commented):
There is still a small cleanup to be made. I'm fine doing it in a different review since the functionality is correct as is.

salt/modules/nilrt_ip.py
```diff
@@ -504,7 +621,7 @@ def disable(interface):
     return down(interface)
 
 
-def _persist_config(section, token, value):
+def _save_config(section, token, value):
```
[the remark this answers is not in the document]
alexvasiu: for reading, we use config parser and for writing nirtcfg (is also used by the system) ... we should test if the file will be corrupted if we are writing at the same time with nirtcfg and config parser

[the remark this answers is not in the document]
rares-pop: Okay. It looks good!

rallytime (changes requested):
I just have a couple of small requests.

rallytime (approved):

rallytime:
@rares-pop How does this look to you now?

rares-pop:
@rallytime - the code looks good and works correctly.

In the near future we'll replace the _save_config to use the ConfigParser as opposed to a binary that's saving the things in the ini.. It's fine to leave it in a separate commit since it's not critical by any means and we have to check that the ConfigParser is 'thread-safe' in terms of not corrupting the ini file when called multiple times from multiple processes.

rallytime:
@alexvasiu There's one lint error that needs to be fixed here: https://jenkinsci.saltstack.com/job/pr-lint/job/PR-48617/10/

alexvasiu:
@rallytime Fixed

PR #48623, opened 2018-07-17
Get multiple mines in single mine.get request

amuhametov:
### What does this PR do?
With more than 10k minons it takes more than 30 seconds to get one mine for '*' target (using consul as cache backend).
The more mine.get calls I've inside a jinja template, the longer the template executes.
 
This PR allows multiple mines to be returned in a single mine.get request
Backward-compatible, tested with 2017.7

### Previous Behavior
mine.get was able to return single function value only

### New Behavior
Now you can pass a list as "fun" value to mine.get, set "as_dict=True" and get a dict as a result

### Tests written?

No

### Commits signed with GPG?

No


salt/daemons/masterapi.py
```diff
@@ -587,10 +619,15 @@ def _mine_get(self, load, skip_verify=False):
         minions = _res['minions']
         for minion in minions:
             fdata = self.cache.fetch('minions/{0}'.format(minion), 'mine')
+            if load.get('as_dict'): ret[minion] = {}
```
gtmanfred: This should be two lines.

gtmanfred (changes requested):
There is a lot of extra stuff in here that i think is just not needed.

Is there a reason to not just always pass in a list?

We could always just do a .split(',') on `fun` when it gets passed to mine.get, and then get rid of all of this extra code, and if it is just one function in the list, we return like always (to keep backwards compatibility, and if it is a list with more then one entry we always return a dictionary.

That would be simpler and remove a lot of this extra branching that I don't think is necessary.

Also this will need tests added

Otherwise, welcome, i am actually really excited about this idea!

Thanks,
Daniel

amuhametov:
The only reason is backwards-compatibility.
Removing branches (return as dict always) leads salt users to modify their states/templates/runners, etc. Also it becomes impossible to backport this feature into 2016/2017 releases.

We could make a deprecation warning for non-dict mine.get requests in next release, then remove those extra branches. This way may be a compromise for user experience.

amuhametov:
> We could always just do a .split(',') on fun when it gets passed to mine.get, and then get rid of all of this extra code, and if it is just one function in the list, we return like always (to keep backwards compatibility, and if it is a list with more then one entry we always return a dictionary.

Not exactly. It is wrong to rely on size of  the list. Example:
mine.get mine1 - fine, old-fashioned, return non-dict
mine.get '["mine1"]' - ambiguous - we can't be sure what type of result is being expected, dict or non-dict.

salt/daemons/masterapi.py
```diff
@@ -548,6 +548,30 @@ def _mine_get(self, load, skip_verify=False):
         if not skip_verify:
             if any(key not in load for key in ('id', 'tgt', 'fun')):
                 return {}
+
+        if isinstance(load['fun'], six.string_types):
+            functions = list(set(load['fun'].split(',')))
+        else:
+            if not load.get('as_dict'):
+                log.warning('Passed function as object to mine.get but "as_dict" is {0}, '
```
rares-pop: can you refactor those warnings to avoid duplicating the code?

I'm thinking to check if not load.get('as_dict') outside the else branch here and bundle its condition in the below check

amuhametov: Make a placeholder for the error text you mean?

gtmanfred:
Just only ever allow a `string` to be the input, don't allow passing in a 'list' object, we don't allow that now.

It should just be a comma separated list.

salt/daemons/masterapi.py
```diff
@@ -587,10 +619,17 @@ def _mine_get(self, load, skip_verify=False):
...
         for minion in minions:
             fdata = self.cache.fetch('minions/{0}'.format(minion), 'mine')
+            if load.get('as_dict'):
+                ret[minion] = {}
             if isinstance(fdata, dict):
-                fdata = fdata.get(load['fun'])
-                if fdata:
-                    ret[minion] = fdata
+                for fun in functions_allowed:
+                    if fdata.has_key(fun):
```
rares-pop: this is an anti-pattern - doing a fdata.has_key(fun) then fdata.get(fun). Do you really need it?

amuhametov: Just to keep ret[minion][fun] empty if no such mine exists for current minion. Any ideas to make it better?

amuhametov:
@gtmanfred 
I made it to be close to yumpkg.group_install method.

What if there is a python list of mines I'd get in my code?
Join them by comma? That seems to be an overhead from the user position for me. Isn't it?

gtmanfred:
Actually, we need to make sure we support older minions, so just make sure it always passes a string in the load.

Eitherway, i think a lot of the logic here is superfluous, and could be simplified.  `as_dict` just isn't needed.

amuhametov:
@gtmanfred 
> Actually, we need to make sure we support older minions, so just make sure it always passes a string in the load.
Do you really would exclude support of python lists here? What about runners and python renderer?

> Eitherway, i think a lot of the logic here is superfluous, and could be simplified. as_dict just isn't needed.
I can agree to refactor log messages here. But removing as_dict will make impossible to get one mine as dict :(

gtmanfred:
It is not impossible.

Here is what you do, allow lists and strings to be passed in, then do the `,`.join(list) if isinstance(fun, list), otherwise if it is a string, just pass it to the master.

Then do a `load['fun'].split(',')` on the master, and if the length is 1, return a dictionary for the one return like we do now, but if it has more than one entry in the list, then return a dictionary of all the returns.

I would also switch the return to be `data[function][minion]` so that your return doesn't really change, and will be the same format as now if you just look in the "function" key in the dictionary.  So your output if a list is passed would be

```
network.ip_addrs:
  minion1:
    - 192.168.0.1
  minion2:
    - 192.168.0.2
test.ping:
  minion1:
    True
  minion2:
    True
```

amuhametov:
> Here is what you do, allow lists and strings to be passed in, then do the ,.join(list) if isinstance(fun, list), otherwise if it is a string, just pass it to the master.

Then do a load['fun'].split(',') on the master, and if the length is 1, return a dictionary for the one return like we do now, but if it has more than one entry in the list, then return a dictionary of all the returns.

Length can't be used as a switch of the return type, see [https://github.com/saltstack/salt/pull/48623#issuecomment-405611024](url)
I'll remove the list object, ok.

> I would also switch the return to be data[function][minion] so that your return doesn't really change, and will be the same format as now if you just look in the "function" key in the dictionary. So your output if a list is passed would be

M'kay

gtmanfred:
Just add a note that if it is a one entry list that is passed in, that it will be the old format.

amuhametov:
> Just add a note that if it is a one entry list that is passed in, that it will be the old format.
No, the behavior must be predicted!

gtmanfred:
well, in the end you could just read that the input was a list, and use a different temporary variable to mark that, and just put it in the dict format at the end once the mine data comes back from the master.

amuhametov:
> well, in the end you could just read that the input was a list, and use a different temporary variable to mark that, and just put it in the dict format at the end once the mine data comes back from the master.

Both minion and runner do not reprocess data they get from the master. Master API is the only place for such logic.

Simplified the code as much as I can.

gtmanfred (approved):

salt/utils/minions.py
```diff
@@ -1156,11 +1156,28 @@ def mine_get(tgt, fun, tgt_type='glob', opts=None):
...
-        if fdata:
-            ret[minion] = fdata
+
+        if isinstance(mdata, dict):
+            for fun in functions:
+                if mdata.has_key(fun):
+                    if _ret_dict:
+                        ret.setdefault(fun, {})[minion] = mdata.get(fun)
+                    else:
+                        ret[minion] = mdata.get(fun)
```
rares-pop: If this works for multi-function jobs, isn't this overwriting itself, being valid only for the last function?

If it's just for single-function jobs, why is inside the for?

cachedout:
We're still waiting on feedback for some of the comments from @rares-pop before a final review.

salt/utils/minions.py
amuhametov: Moved single-function jib out of the loop. Is it better now?

salt/daemons/masterapi.py
```diff
@@ -609,13 +609,17 @@ def _mine_get(self, load, skip_verify=False):
         minions = _res['minions']
         for minion in minions:
             fdata = self.cache.fetch('minions/{0}'.format(minion), 'mine')
-            if isinstance(fdata, dict):
+
+            if fdata is None or not isinstance(fdata, dict):
```
rares-pop: I don't think you need the first comparison with None, since that's not an instance of dict.

Doing: if not isinstance(fdata, dict) should suffice

rares-pop:
It looks good to me!

amuhametov:
What do you think of code duplicates here? Sholdn't we optimize it?

cachedout (approved):

gtmanfred (changes requested):
Looks like there are some failing tests.

https://jenkinsci.saltstack.com/job/pr-kitchen-centos7-py3/job/PR-48623/12/

It would also be good to add some more tests around this feature

amuhametov:
@gtmanfred should be fixed now.

Tried to write a test.

gtmanfred:
Awesome, thanks, i will rereview once the tests are done running.

isbm (changes requested):

salt/daemons/masterapi.py
```diff
@@ -548,6 +548,20 @@ def _mine_get(self, load, skip_verify=False):
         if not skip_verify:
             if any(key not in load for key in ('id', 'tgt', 'fun')):
                 return {}
+
+        _ret_dict = False
+        if isinstance(load['fun'], six.string_types):
+            functions = list(set(load['fun'].split(',')))
+            if len(functions) > 1:
+                _ret_dict = True
```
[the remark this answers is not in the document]
amuhametov: Not exactly. If you pass a single-item list to mine.get the result will be formed as dict too.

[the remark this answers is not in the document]
isbm: @amuhametov huh? I mean  you _either_ set `_ret_dict` to a `bool` or return an empty `dict`. So what I am saying is:

```python
_ret_dict = False
...
if len(functions) > 1:
    _ret_dict = True
...
```
...replace with:

```python
_ret_dict = len(functions) > 1
```

...which you will pass forward or bail-out with your empty dict.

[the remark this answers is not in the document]
amuhametov: Ah. Fixed, thanks.

gtmanfred (changes requested):
I am still hesitant to approve this because it is still failing tests.

https://jenkinsci.saltstack.com/job/pr-kitchen-centos7-py3/job/PR-48623/7/

isbm (changes requested):
Tests must pass.

amuhametov:
Is Jenkins broken?

https://jenkinsci.saltstack.com/job/pr-kitchen-centos7-py2/job/PR-48623/14/console

rallytime:
@amuhametov We've been moving over to a new instance, so you may have seen some strangeness in between, but that should be all set now.

I am not able to update the branch here. Can you rebase this with the current `develop` so we can get an accurate test run? That will trigger all of the new tests.

amuhametov:
@rallytime Tried to rebase but there ara too many conflicts :(

Jenkins still has errors.

gtmanfred:
It looks like you have two lint warnings remaining.

https://jenkinsci.saltstack.com/job/pr-lint/job/PR-48623/17/warnings52Result/new/

amuhametov:
Jenkins is still broken?
https://jenkinsci.saltstack.com/job/pr-kitchen-centos7-py2/job/PR-48623/18/console

rallytime:
Let's wait and see how this latest run goes before restarting them again.

amuhametov:
https://jenkinsci.saltstack.com/job/pr-kitchen-centos7-py2/job/PR-48623/19/ is stuck :(

rallytime:
The tests that i ran before came back with results, so we don't need to keep running these.

Both of the failures that occurred in the results above have been fixed and are making their way upstream. 

From a tests perspective, this looks good to me now.

gtmanfred (approved):

PR #48631, opened 2018-07-17
[RFC] add rfc for loader changes in Fluorine

gtmanfred:
### What does this PR do?
Add rfc for python package loader changes

@cachedout @terminalmage ready to review this one.

### Commits signed with GPG?

Yes

### DIFF
--- a/rfcs/0001-adding-package-logic-to-loader.md
+++ b/rfcs/0001-adding-package-logic-to-loader.md
@@ -1,0 +1,37 @@
+- Feature Name: adding_python_package_logic_to_loader
+- Start Date: 2018-06-29
+- RFC PR:
+- Salt Issue:
+
+# Summary
+[summary]: #summary
+
+Add support for python packages to the loader. Instead of just having
+apache.py, we would have `salt/modules/apache/__init__.py`, and all the other
+modules under that directory loaded in the `apache.<function>` namespace that
+we already have.
+
+# Motivation
+[motivation]: #motivation
+
+Split modules out into different git repositories, and allow the community more control to help maintain them.
+
+# Design
+[design]: #detailed-design
+
+This should be pretty straight forward, we should be able to use the same
+_module_dirs that we are using now, and iterate over the files, and if they are
+a directory, then we load all of the modules underneath it into the namespace
+of the directory.
+
+Heirarchy would be alphabetical, and we would need to log messages if the
+function exists in an earlier file that has been loaded.
+
+## Alternatives
+[alternatives]: #alternatives
+
+## Unresolved questions
+[unresolved]: #unresolved-questions
+
+# Drawbacks
+[drawbacks]: #drawbacks


terminalmage (approved):
Had a discussion with @garethgreenaway and @cachedout today, and we came up with the following ideas/information:

1. The loader already supports looking into a subdir and loading the `__init__.py` within it. This means that we don't need to modify the `LazyLoader`, just the `_module_dirs()` helper func that determines which directory paths to pass to the `LazyLoader`'s constructor.

2. Instead of separately loading all .py files including the `__init__.py`, only load the `__init__.py`. Developers can split up logic into separate sub-modules, but any functions which they want to be made available _must_ be imported into the `__init__.py`. This allows us to maintain a 1-to-1 relationship between module name and associated filename, and will require no actual changes to the `LazyLoader` (see point 1 above).

3. Managing the distribution of modules into Salt proper can be done via an execution module (which would also allow one to invoke it via salt-api, which would be ++). This execution module would have functions to list/add/remove/pull your tracked git repos, with all information about the tracked repos stored in a JSON or msgpack-serialized file. That same file would then be read in by `_module_dirs()` in loader.py so it knows all of the locations of the cached git repos, and in turn the `LazyLoader` would know where to find all of the tracked modules. *(NOTE: This approach would require a runner module counterpart for managing things like runners, fileserver modules, etc. on the master.)*

   3a. There's actually nothing stopping us from also accepting `salt://` paths to dirs containing the modules, in addition to git repos.

4. (only discussed this with @garethgreenaway) If we're going to be moving to this git repository means of distributing modules, we don't necessarily need to hew too closely to this requirement of using subdirs with `__init__.py`. Imagine the following layout:

   ```
   modules
   ├── bar.py
   └── foo
       └── __init__.py
   ```

   In this example, `modules` is a top-level directory in a git repo called `reponame`. Somewhere in /var/cache/salt we'd have this cached, for the sake of argument let's call it `/var/cache/salt/minion/something_clever/reponame/modules`. If we passed each tracked git repo's `modules` (or `states`, etc.) directory path separately to the `LazyLoader`'s constructor, then when we create the file mapping we'd just be doing an `os.listdir('/var/cache/salt/minion/something_clever/reponame/modules')`, which would result in both the `foo` and `bar` modules being loaded. So, organizing them into a subdirectory and using an `__init__.py` is not a hard requirement in this scenario. @cachedout your thoughts? (you weren't there for that part of the discussion)

rfcs/0001-adding-package-logic-to-loader.md
```diff
@@ -0,0 +1,37 @@
...
+This should be pretty straight forward, we should be able to use the same
+_module_dirs that we are using now, and iterate over the files, and if they are
+a directory, then we load all of the modules underneath it into the namespace
+of the directory.
+
+Heirarchy would be alphabetical, and we would need to log messages if the
+function exists in an earlier file that has been loaded.
+
+## Alternatives
+[alternatives]: #alternatives
```
isbm: Categorise modules to at least three parts:

- Core modules. Modules, that should be working _everywhere_ (cross-platform) and are vital for a Salt primary usage (package installations, configuration changes).

- All 3rd party modules. Modules that requires additional and optional deps and aren't generic (e.g. Facebook's Cassandra etc).

- All cloud related modules.

This also valid to states and so the states that are related to the 2nd or 3rd category should be moved along with these modules as well.

First layer stays with the Salt's core within the same repo as they are essential. The other two categories should have their own repos.

rfcs/0001-adding-package-logic-to-loader.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Unresolved questions
+[unresolved]: #unresolved-questions
+
```
isbm: This seems give a major impact elsewhere:

- How `setup.py` should behave from now on?
- Since this is totally different Git repos, it impacts releases, tests and maintenance. How release process will look like here, sine we will get more than one source tarball?
- If we decide to keep a single tarball on the release, how do we make sure that the modules are properly tested and aren't broken?
- Splitting repos will also split packages and this should be taken to the account.

rfcs/0001-adding-package-logic-to-loader.md
```diff
@@ -0,0 +1,37 @@
...
+# Drawbacks
+[drawbacks]: #drawbacks
```
isbm: - Find out process release.
- Testing of each Git repo will go on their own.
- Branches/PRs/reviews will multiply on number of repos.
- Packaging will likely split too and turn into a dependencies.

isbm (changes requested):
I am adding couple of changes to the RFC that needs to be addressed prior to this.

rfcs/0001-adding-package-logic-to-loader.md
gtmanfred: So, this rfc is not about splitting the modules into new repositories yet.  That will be a different one, this is just about supporting the ability to load packages/submodules into the same namespace.

The rest of these questions will be answered in a different rfc.

isbm: Well, [you wrote it is about splitting](https://github.com/saltstack/salt/pull/48631/files#diff-983ac08a468477835dc9d9214f7d7c66R17) and that confused me. :wink: 

I would say "motivation" here is to have a namespace. And that makes totally lots of sense here (and I am all for it!).

PR #48610, opened 2018-07-16
only run pylint on files that change

gtmanfred:
### What does this PR do?
Only run pylint on files that change

### Commits signed with GPG?

Yes

PR #48633, opened 2018-07-17
Revert "only run pylint on files that change"

gtmanfred:
Reverts saltstack/salt#48610

Revert this until we can talk about how to set `shopt -s globstar` in the lint jenkinsfile, so that `**` recursively expands.


gtmanfred:
@KaiSforza fyi for when you get back.

PR #48593, opened 2018-07-15
Fix importlib pyc loading order

pritambaral:
### What does this PR do?

Fixes the order in `SUFFIXES` to place `'.py'` before `'.pyc'` when using `importlib` (i.e., on Python versions 3.5 and above), the same order that `imp.get_suffixes()` (deprecated from Python 3.5 onwards) returns its result in.

### What issues does this PR fix or reference?

Fixes #46924

Also see: https://github.com/saltstack/salt/pull/48195#issuecomment-398898394

### Previous Behavior

On Python 3.5 and newer, `.pyc` files are preferred by the loader to `.py` files.

### New Behavior

`.py` files are preferred by the loader to `.pyc` files.

### Tests written?

No

### Commits signed with GPG?

Yes

### Additional Notes

The patch in this PR has been verified to apply cleanly when cherry-picked onto `v2018.3`.

PR #48636, opened 2018-07-17
Properly handle optimized .pyc files in loader in PY3

terminalmage:
See https://github.com/saltstack/salt/pull/48195#pullrequestreview-130530059 for some background.

This PR properly handles optimized .pyc files in the loader. It also introduces a new config option, `optimization_order`, which allows for the priority to be set so that optimized .pyc files are preferred.

It also fixes issue #46924 (separately fixed in #48593) in which the loader preferred .pyc files over .py files of the same name in Python 3, causing updated .py files not to be loaded. Note that this means that `optimization_order` only comes into play when Salt is distributed without .py files. ***NOTE: since #48593 was merged, my commit that made the same fix was a no-op and has disappeared in my latest rebase.***

Unit tests added to confirm that the `optimization_order` option works as expected.

In addition to the above, this PR makes the following two changes:

1. Because we do an `os.listdir()` on each module directory processed by the loader, this means that the `__pycache__` dir was in the results. It would never end up being loaded, so this didn't break anything, but I have nonetheless excluded it from the `os.listdir()` results so that the loader doesn't waste any effort processing it.

2. The suffix map/order are now compiled in the LazyLoader's dunder init. The values which are used to build both are module-level globals, so there is no need to repeat this every time the file mapping is refreshed.

@cachedout due to the fact that the `optimization_order` only matters when there is no .py file for that module, I decided not to add release notes for this. It is the corner-iest of corner cases.

### DIFF
--- a/doc/ref/configuration/master.rst
+++ b/doc/ref/configuration/master.rst
@@ -1612,6 +1612,22 @@ the master will drop the request and the minion's key will remain accepted.
 
     allow_minion_key_revoke: False
 
+.. conf_master:: optimization_order
+
+``optimization_order``
+----------------------
+
+Default: ``[0, 1, 2]``
+
+In cases where Salt is distributed without .py files, this option determines
+the priority of optimization level(s) Salt's module loader should prefer.
+
+.. code-block:: yaml
+
+    optimization_order:
+      - 2
+      - 0
+      - 1
 
 Master Large Scale Tuning Settings
 ==================================
--- a/doc/ref/configuration/master.rst
+++ b/doc/ref/configuration/master.rst
@@ -1622,6 +1622,9 @@ Default: ``[0, 1, 2]``
 In cases where Salt is distributed without .py files, this option determines
 the priority of optimization level(s) Salt's module loader should prefer.
 
+.. note::
+    This option is only supported on Python 3.5+.
+
 .. code-block:: yaml
 
     optimization_order:
--- a/doc/ref/configuration/minion.rst
+++ b/doc/ref/configuration/minion.rst
@@ -1195,6 +1195,22 @@ option to ``False`` to keep Salt from updating the mine with this information.
 
     docker.update_mine: False
 
+.. conf_master:: optimization_order
+
+``optimization_order``
+----------------------
+
+Default: ``[0, 1, 2]``
+
+In cases where Salt is distributed without .py files, this option determines
+the priority of optimization level(s) Salt's module loader should prefer.
+
+.. code-block:: yaml
+
+    optimization_order:
+      - 2
+      - 0
+      - 1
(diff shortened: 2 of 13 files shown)


terminalmage:
If one wanted to get Salt to load anything but the default (unoptimized) .pyc file, they would need a way to tell Salt to do so. That's what this option is for.

ari:
So on  a platform that creates the optimised files (eg FreeBSD) by default, they will be ignored and the .pyc used instead? And if the .py source file is newer than the .pyc compiled file, then the source file will be used?

Is it too resource draining to set default optimization_order so that files are loaded from most optimised to least (opt-2, opt-2, pyc, py) unless the py file is newer that the others?

I haven't done any benchmarking, but anything to make salt go faster is welcome.

terminalmage:
As noted in the OP, the `optimization_order` value only comes into play when Salt is distributed without .py files for its modules.

ari:
Sorry to be dense. If there are source files, will this now mean compiled files are ignored or used in order from most optimised?

gtmanfred (approved):

isbm (approved):

rallytime:
@terminalmage Lint: https://jenkinsci.saltstack.com/job/pr-lint/job/PR-48636/2/warnings52Result/

terminalmage:
@ari If there are source files, it means that .pyc files will not be loaded. The normal means of loading source files will be used (which will also update the .pyc if needed).

Until https://github.com/saltstack/salt/pull/48593 was merged a couple days ago, Python 3 would prefer .pyc files of the same module name, but this would have the effect not picking up changes to the .py file if it were updated.

terminalmage:
I'm working on the lint failures now, as well as another Python 3-related loader issue.

terminalmage:
@ari No, there is no way to do that right now, and I'm not sure it would be a good idea to do that since stat'ing every module every time the loader is refreshed would consume additional resources, and I don't know that the net improvement gain by loading .pyc files would be worth it.

rallytime:
@terminalmage Something isn't quite right here with the Py3 tests. Can you take another swing at this?

rallytime (approved):

gtmanfred:
yes, we merge forward to 2018.3 before branching the 2018.3.3 release branch.

the target is by the end of this month.

cedwards:
@rhavenn I will be testing this patch in the FreeBSD port tonight. Fingers crossed we can patch this mid-stream in FreeBSD until it is available in 2018.3.3.

rhavenn:
@cedwards Sweet!!! Thanks.

cedwards:
I've been working on this, but I'm still having trouble. Need more time to troubleshoot.

PR #48654, opened 2018-07-18
[WIP] Move handling of file protocols to a new subsystem in salt.fileclient

gtmanfred:
### What does this PR do?
Move handling of protocols for urls to be a subsystem

### Tests written?

Yes

### Commits signed with GPG?

Yes


cachedout:
If we're going to add an entirely new subsystem, presumably in the interest of allowing people to write their own plugins, we need to document that fact and include an explainer on how to write one. If the intention is to allow these plugins to be synced, we need to think about that functionality as well.

isbm (approved):
:+1: :+1: :+1: (and current issues you know anyways, so I leave them out).

rallytime (commented):
Looks like this needs some ❤️ still.

PR #48658, opened 2018-07-18
Improve handling of json output (#43138)

wyardley:
### What does this PR do?
(This is #48492, but cherry-picked against 2017.7)
Rather than look for [ or { at the beginning of a line, it looks for equality. Since the JSON output of `npm install` seems to have the first character on its own, this seems to work, at least for the npm version I looked at -- however, I'm open to better / other approaches that would be better. I think the previous approach was a bit too limited, since there could be other output in the future that matches `^[` or `^{`.

Not sure if we could just ignore either stdout or stderr (I think some versions of node send the json output to stdout), or if there's another way this could be fixed. I am guessing the json output is reasonably consistent, but haven't tested older versions. [edit, checked, and currently, the message is going to stdout as well as the json output, so that won't work here]

Idempotency issue - first run will "fail" with an error, even when packages get installed, if the output has a string like:
`[grpc] Success: "/usr/lib/node_modules/@google-cloud/monitoring/node_modules/grpc/src/node/extension_binary/node-v57-linux-x64-glibc/grpc_node.node" is installed via remote`, the first run will fail because the json parsing routine fails.

More details in #43138; replaces #48492

### Tests written?
Yes

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/modules/npm.py
+++ b/salt/modules/npm.py
@@ -160,7 +160,11 @@ def install(pkg=None,
             env.update({'SUDO_UID': uid, 'SUDO_USER': ''})
 
     cmd = ' '.join(cmd)
-    result = __salt__['cmd.run_all'](cmd, python_shell=True, cwd=dir, runas=runas, env=env)
+    result = __salt__['cmd.run_all'](cmd,
+                                     python_shell=True,
+                                     cwd=dir,
+                                     runas=runas,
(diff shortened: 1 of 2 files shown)


terminalmage (approved):

rallytime (approved):

Issue #48599, opened 2018-07-16
module ldap3 'TypeError: ('expected a byte string in the list''


angeloudy:
### Description of Issue/Question
I think this pull request #47029 fixed the problem in python2.7 while broke it in python3.
From my experience with python-ldap with python3, when writing back to ldap, the attribute name should be `str`, while the value should be `bytes`. 

```
 Comment: An exception occurred in this state: Traceback (most recent call last):
                File "/usr/local/lib/python3.6/site-packages/salt/state.py", line 1905, in call
                  **cdata['kwargs'])
                File "/usr/local/lib/python3.6/site-packages/salt/loader.py", line 1830, in wrapper
                  return f(*args, **kwargs)
                File "/usr/local/lib/python3.6/site-packages/salt/states/ldap.py", line 327, in managed
                  __salt__['ldap3.change'](l, dn, o, n)
                File "/usr/local/lib/python3.6/site-packages/salt/modules/ldap3.py", line 595, in change
                  l.c.modify_s(dn, modlist)
                File "/usr/local/lib/python3.6/site-packages/ldap/ldapobject.py", line 600, in modify_s
                  return self.modify_ext_s(dn,modlist,None,None)
                File "/usr/local/lib/python3.6/site-packages/ldap/ldapobject.py", line 572, in modify_ext_s
                  msgid = self.modify_ext(dn,modlist,serverctrls,clientctrls)
                File "/usr/local/lib/python3.6/site-packages/ldap/ldapobject.py", line 569, in modify_ext
                  return self._ldap_call(self._l.modify_ext,dn,modlist,RequestControlTuples(serverctrls),RequestControlTuples(clientctrls))
                File "/usr/local/lib/python3.6/site-packages/ldap/ldapobject.py", line 294, in _ldap_call
                  result = func(*args,**kwargs)
              TypeError: ('expected a byte string in the list', 'tao')
```
### Setup
(Please provide relevant configs and/or SLS files (Be sure to remove sensitive info).)

### Steps to Reproduce Issue
(Include debug logs if possible and relevant.)

### Versions Report
(Provided by running `salt --versions-report`. Please also mention any differences in master/minion versions.)
```
Salt Version:
           Salt: 2018.3.1

Dependency Versions:
           cffi: 1.11.5
       cherrypy: unknown
       dateutil: 2.7.2
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.10
[...]

PR #48666, opened 2018-07-19
values need to be bytes when writing back to ldap

angeloudy:
fix for #48599


salt/modules/ldap3.py
```diff
@@ -572,19 +572,16 @@ def change(connect_spec, dn, before, after):
     # convert the "iterable of values" to lists in case that's what
     # modifyModlist() expects (also to ensure that the caller's dicts
     # are not modified)
-    before = dict(((attr, list(vals))
+    before = dict(((attr, salt.utils.data.encode(vals))
                    for attr, vals in six.iteritems(before)))
-    after = dict(((attr, list(vals))
+    after = dict(((attr, salt.utils.data.encode(vals))
```
terminalmage: I'm not familiar with this module, but `list()` has been there from the beginning (according to git log). Can you explain why you removed it here?

dwoz:
Should the encoding/decoding be specific to the ldap server? It looks like this PR is basing the encoding/decoding based on the client's system settings. Have you thought about what happens if the client system is using a non utf-8 encoding?

salt/modules/ldap3.py
angeloudy: In fact, it would be safe to use `salt.utils.data.encode(list(vals))` because vals could be type `OrderedSet`

angeloudy:
@dwoz 
That's a good point!
I have no idea how to get the encoding.
What do you suggest?

dwoz:
@angeloudy I think it's best to use utf-8 here (it looks like that is what LDAPv3 expects by default). We can make it configurable down the road if needed

cachedout:
@dwoz and @angeloudy There is also `__salt_system_encoding__` that should be available.

angeloudy:
@cachedout should `salt.utils.data.encode` use `__salt_system_encoding__` by default?
I did have any 'utf8' hardcoded in my changes.

terminalmage:
@angeloudy It would try utf8 first, and then try the detected encoding if that raises a `UnicodeEncodeError`.

rallytime:
@angeloudy Are you planning on making the requested changes here?

angeloudy:
@rallytime what requested changes.
I am fine with my current version.

angeloudy:
@terminalmage @rallytime 
if an `UnicodeEncodeError` exception catching is necessary, shouldn't it be done inside `salt.utils.data.encode`?

rallytime:
@terminalmage and @dwoz can you comment/re-review?

angeloudy:
Hi, any plans to merge this pull request so that I don't have to maintain my own patch?
I don't understand why are you guys worrying about UnicodeEncodeError, the encode method is being used everywhere in saltstack.

dwoz (approved):

salt/modules/ldap3.py
terminalmage: Then you should probably make that change.

terminalmage (changes requested):
I don't think anyone is worrying about an encoding error. You still haven't made the suggested changes from a few months ago, which is probably why this hasn't been merged.

terminalmage (approved):

PR #48670, opened 2018-07-19
Add the "traverse" jinja filter

bbinet:
This filter can be really useful in states to select some leaf pillar data or fallback with a default value.

The code is trivial as it is just a matter of exposing the existing `traverse_dict_and_list` function as a jinja filter through the `@jinja_filter('traverse')` decorator.
That is the reason why it seems safe to merge it in the 2018.3 branch instead of the develop branch.

### DIFF
--- a/doc/topics/jinja/index.rst
+++ b/doc/topics/jinja/index.rst
@@ -1152,7 +1152,7 @@ Returns:
 .. jinja_ref:: traverse
 
 ``traverse``
---------------
+------------
 
 .. versionadded:: 2018.3.3
 
--- a/doc/topics/jinja/index.rst
+++ b/doc/topics/jinja/index.rst
@@ -1148,6 +1148,40 @@ Returns:
     }'
   }
 
+
+.. jinja_ref:: traverse
+
+``traverse``
+--------------
+
+.. versionadded:: 2018.3.3
+
+Traverse a dict or list using a colon-delimited target string.
+The target 'foo:bar:0' will return data['foo']['bar'][0] if this value exists,
+and will otherwise return the provided default value.
+
+Example:
+
+.. code-block:: jinja
+
+  {{ {'a1': {'b1': {'c1': 'foo'}}, 'a2': 'bar'} | traverse('a1:b1', 'default') }}
+
+Returns:
+
+.. code-block:: python
+
+  {'c1': 'foo'}
+
+.. code-block:: jinja
+
+  {{ {'a1': {'b1': {'c1': 'foo'}}, 'a2': 'bar'} | traverse('a2:b2', 'default') }}
+
+Returns:
+
+.. code-block:: python
+
+  'default'
+
 .. _`builtin filters`: http://jinja.pocoo.org/docs/templates/#builtin-filters
 .. _`timelib`: https://github.com/pediapress/timelib/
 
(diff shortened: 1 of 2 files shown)


doc/topics/jinja/index.rst
```diff
@@ -1148,6 +1148,40 @@ Returns:
     }'
   }
 
+
+.. jinja_ref:: traverse
+
+``traverse``
+--------------
```
mirceaulinic: The doc checks might trigger a warning here (the line of `-` must be as long as the test above).

doc/topics/jinja/index.rst
```diff
@@ -1148,6 +1148,40 @@ Returns:
...
+
+.. versionadded:: 2018.3.3
```
mirceaulinic: `.. versionadded:: Fluorine`

doc/topics/jinja/index.rst
```diff
@@ -1148,6 +1148,40 @@ Returns:
...
+
+  {'c1': 'foo'}
+
+.. code-block:: jinja
+
+  {{ {'a1': {'b1': {'c1': 'foo'}}, 'a2': 'bar'} | traverse('a2:b2', 'default') }}
+
+Returns:
+
+.. code-block:: python
```
mirceaulinic: `.. code-block:: text`, I think (again, RST might complain about it).

mirceaulinic (approved):
Nice one. I ended up with a custom execution module, e.g,, `{{ salt.my_utils.traverse(my_dict) }}`, but this filter would be better IMO.

doc/topics/jinja/index.rst
bbinet: Ok

doc/topics/jinja/index.rst
bbinet: Ok

doc/topics/jinja/index.rst
bbinet: `.. code-block:: python` seems ok, see: https://github.com/saltstack/salt/pull/48670/files#diff-f819197957c800b37cc714918abe44f4L1140

doc/topics/jinja/index.rst
bbinet: In fact, would it be possible to include this patch in the next minor version of salt?

As I told in the PR description, the code is trivial as it is just a matter of exposing the existing traverse_dict_and_list function as a jinja filter through the @jinja_filter('traverse') decorator.
So it seems safe to me to merge it in the 2018.3 branch instead of the develop branch.

doc/topics/jinja/index.rst
```diff
@@ -1148,6 +1148,40 @@ Returns:
...
+
+.. code-block:: jinja
+
+  {{ {'a1': {'b1': {'c1': 'foo'}}, 'a2': 'bar'} | traverse('a1:b1', 'default') }}
+
+Returns:
+
+.. code-block:: python
+
+  {'c1': 'foo'}
```
rallytime: These code-block sections need to be a full indent for this to display properly in the docs. Can you fix this one, as well as the rest of the code-block sections?

bbinet: @rallytime these python code-block sections already exists in this `index.rst` file and renders correctly in the docs. See for example the  `compare_lists` filter doc: https://docs.saltstack.com/en/latest/topics/jinja/index.html#compare-lists

rallytime: Interesting. Usually sphinx generates a warning about that I thought. *shrug*

rallytime (approved):

Issue #47059, opened 2018-04-13
Some states incorrectly return None instead of an empty dict when there are no changes


OrlandoArcapix:
### Description of Issue/Question

Version 2018.3.0 fails state runs when no changes are made with modules which return "None" instead of an empty dict for the 'changes' data.

See ticket #46985 

### Steps to Reproduce Issue

See the following lines in files in `salt/states/`:
```
boto_cfn.py:        ret['changes'] = None
boto_cfn.py:        ret['changes'] = None
boto_cfn.py:        ret['changes'] = None
boto_iam.py:    ret['changes'] = None
boto_iot.py:            ret['changes'] = None
glassfish.py:            ret['changes'] = None
glassfish.py:            ret['changes'] = None
glassfish.py:        ret['changes'] = None
grafana_datasource.py:            ret['changes'] = None
influxdb_user.py:                ret['changes'] = None
vault.py:        ret['changes'] = None
zenoss.py:        ret['changes'] = None
zenoss.py:        ret['changes'] = None
```

OrlandoArcapix:
Hey @rallytime - I've put in a PR against issue #47059, which replaces those `None` entries with an empty dict`{}`m though I've not been able to test as I don't have an environment which uses the state modules I've updated in that pull request.

rallytime:
Awesome! Thanks @OrlandoArcapix! Let's see how the test run goes on that. Thanks for filing this issue and for the fix. :)

rallytime:
Closed via #47060

Issue #46985, opened 2018-04-10
grafana4_user.present and grafana4_org.present states not working in 2018.3.0


OrlandoArcapix:
### Description of Issue/Question

I'm testing 2018.3.0 and have run into an issue with the grafana4 states - I'm getting:
```  The State execution failed to record the order in which all states were executed. The state return missing data is:
  {u'changes': {},
   u'comment': u"An exception occurred in this state: 'Changes' should be a dictionary.",
   u'name': u'later',
   u'result': False}
```
from the grafana4_user.present and grafana4_org.present states.

### Setup

1. Set up grafana4.
2. Apply a state such as this:
```
test grafana org:
  grafana4_org.present:
    - name: 'Test'
    - users:
        admin: Admin

test grafana user:
  grafana4_user.present:
    - name: test
    - password: 'somepass'
    - is_admin: False
    - email: 'test@localhost'
```

### Steps to Reproduce Issue

On the first run, the above state will create the Test org (even with `test=true` on the `state.apply`). On a subsequent run, it will return:
```
# salt-call state.apply grafana.test test=false
local:
The State execution failed to record the order in which all states were executed. The state return missing data is:
{u'changes': {},
 u'comment': u"An exception occurred in this state: 'Changes' should be a dictionary.",
 u'name': u'later',
 u'result': False}
The State execution failed to record the order in which all states were executed. The state return missing data is:
{u'changes': {},
 u'comment': u"An exception occurred in this state: 'Changes' should be a dictionary.",
 u'name': u'later',
 u'result': False}
----------
          ID: test grafana org
    Function: grafana4_org.present
        Name: Test
      Result: False
     Comment: An exception occurred in this state: 'Changes' should be a dictionary.
     Changes:
----------
          ID: test grafana user
    Function: grafana4_user.present
        Name: test
      Result: False
     Comment: An exception occurred in this state: 'Changes' should be a dictionary.
     Changes:

Summary for local
------------
Succeeded: 0
Failed:    2
------------
Total states run:     2
Total run time:   0.000 ms
```

### Versions Report
```
# salt-call --versions-report
Salt Version:
           Salt: 2018.3.0

Dependency Versions:
           cffi: Not Installed
       cherrypy: Not Installed
       dateutil: 2.6.1
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.7.2
        libgit2: Not Installed
        libnacl: Not Installed
[...]

PR #48685, opened 2018-07-20
Use empty dict for 'changes' instead of None

bbinet:
See #47059 #46985 for previous similar changes.

This PR fixes the missing occurences of changes=None to replace None by an empty dict (this can result in state failures: "An exception occurred in this state: 'Changes' should be a dictionary."):

```
$ git grep "'changes': *None"
salt/states/esxvm.py:                       'changes': None,
salt/states/grafana4_datasource.py:    ret = {'name': name, 'result': None, 'comment': None, 'changes': None}
salt/states/grafana_datasource.py:    ret = {'name': name, 'result': None, 'comment': None, 'changes': None}
salt/states/grafana_datasource.py:    ret = {'result': None, 'comment': None, 'changes': None}
salt/states/vault.py:            'changes': None,
salt/states/vault.py:            'changes': None,
salt/states/vault.py:            'changes': None,
```

Note that I've not changed the `salt/states/grafana4_datasource.py` file in this PR as this is already fixed in upcoming PR #48671.

### DIFF
--- a/salt/states/esxvm.py
+++ b/salt/states/esxvm.py
@@ -434,10 +434,8 @@ def vm_created(name, vm_name, cpu, memory, image, version, interfaces,
               'comment': ''}
 
     if __opts__['test']:
-        result.update({'result': None,
-                       'changes': None,
-                       'comment': 'Virtual machine '
-                                  '{0} will be created'.format(vm_name)})
+        result['comment'] = 'Virtual machine {0} will be created'.format(
+                vm_name)
         return result
 
     service_instance = __salt__['vsphere.get_service_instance_via_proxy']()
--- a/salt/states/vault.py
+++ b/salt/states/vault.py
@@ -55,7 +55,7 @@ def policy_present(name, rules):
     except Exception as e:
         return {
             'name': name,
-            'changes': None,
+            'changes': {},
             'result': False,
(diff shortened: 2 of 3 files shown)


salt/states/esxvm.py
```diff
@@ -434,10 +434,8 @@ def vm_created(name, vm_name, cpu, memory, image, version, interfaces,
               'comment': ''}
 
     if __opts__['test']:
-        result.update({'result': None,
-                       'changes': None,
-                       'comment': 'Virtual machine '
-                                  '{0} will be created'.format(vm_name)})
+        result['comment'] = 'Virtual machine {0} will be created'.format(
+                vm_name)})
```
rallytime: You need to remove the `}` here.

salt/states/vault.py
```diff
@@ -111,7 +111,7 @@ def _handle_existing_policy(name, new_rules, existing_rules):
     if response.status_code != 204:
         return {
             'name': name,
-            'changes': None,
+            'changes': {}
```
rallytime: There needs to be a `,` at the end of this line.

rallytime (changes requested):

bbinet:
@rallytime These are now fixed, thank you!

rallytime (approved):

PR #48687, opened 2018-07-20
Adding a new execution module to manipulate JIRA tickets

mirceaulinic:
### What does this PR do?

Adding a new execution module to manipulate JIRA tickets - very useful to dynamically create tickets as in response to various events (e.g., create ticket on interface down, or on server reboot etc.).

### DIFF
--- a/salt/modules/jira_mod.py
+++ b/salt/modules/jira_mod.py
@@ -1,0 +1,288 @@
+# -*- coding: utf-8 -*-
+'''
+JIRA Execution module
+=====================
+
+.. versionadded:: Fluorine
+
+Execution module to manipulate JIRA tickets via Salt.
+
+This module requires the ``jira`` Python library to be installed.
+
+Configuration example:
+
+.. code-block:: yaml
+
+  jira:
+    server: https://jira.atlassian.org
+    username: salt
+    password: pass
+'''
+from __future__ import absolute_import, unicode_literals, print_function
+
+# Import python libs
+import logging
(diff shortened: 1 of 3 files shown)


damon-atkins:
Please rename jira_mod to atlassian_jira

salt/modules/jira_mod.py
```diff
@@ -0,0 +1,288 @@
...
+Configuration example:
+
+.. code-block:: yaml
+
+  jira:
+    server: https://jira.atlassian.org
+    username: salt
+    password: pass
+'''
+from __future__ import absolute_import
```
rallytime: `print_function` and `unicode_literals` need to be added here for imports

rallytime (changes requested):
One small thing to fix, otherwise this looks good to me. The name of the module is fine with me.

salt/modules/jira_mod.py
mirceaulinic: Fixed @rallytime

rallytime (approved):

damon-atkins:
Reason for the name change e.g. atlassian_jira is so other atlassian modules can have the name atlassian_* pattern.    And also mod not need in the name of a module as its in the module directory.

cachedout:
I don't agree with the proposed naming. We don't, for example, name the `apache` module `apache_foundation_apache` simply because the Apache Foundation produces other software. Jira is well-known on its own and I just don't see how including the vendor name adds any clarity or organizational value.

PR #48688, opened 2018-07-20
A new execution module: PeeringDB

mirceaulinic:
### What does this PR do?

Adds another execution module, for the basic interaction with the [PeeringDB](https://www.peeringdb.com/) API. This is useful to gather data about other networks you can potentially peer with, and automatically establish BGP sessions, e.g., given just a specific AS number , the rest of the information (i.e., IP addresses, places where the remote network is available, etc.) is retrieved from PeeringDB, and the session configuration is automated with minimum to no effort (typing manually IP addresses is both tedious and error prone).

### DIFF
--- a/salt/modules/peeringdb.py
+++ b/salt/modules/peeringdb.py
@@ -1,0 +1,316 @@
+# -*- coding: utf-8 -*-
+'''
+PeeringDB Module
+================
+
+.. versionadded:: Fluorine
+
+Execution module for the basic interaction with the
+`PeeringDB <https://www.peeringdb.com/>`_ API.
+
+While for GET operations (the functions prefixed by ``get_``) the credentials
+are optional, there are some specific details that are visible only to
+authenticated users. Moreover, the credentials are required when adding or
+updating information. That means, the module can equally work out of the box
+without any further configuration with the limitations imposed by the PeeringDB
+API.
+
+For complete API documentation, please refer to https://www.peeringdb.com/apidocs/.
+
+Configuration (in the opts or Pillar):
+
+.. code-block:: yaml
+
+    peeringdb:
+      username: salt
+      password: 5@1t
+'''
+from __future__ import absolute_import
+
+# Import python libs
+import logging
+log = logging.getLogger(__name__)
+
+# Import salt modules
+import salt.utils.http
+try:
+    from salt.utils import clean_kwargs
+except ImportError:
+    from salt.utils.args import clean_kwargs
+
+__virtualname__ = 'peeringdb'
+__proxyenabled__ = ['*']
+
+PEERINGDB_URL = 'https://www.peeringdb.com/api'
+
+
+def __virtual__():
+    return __virtualname__
+
+
+def _get_auth(username=None,
+              password=None):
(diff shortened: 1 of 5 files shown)


salt/modules/peeringdb.py
```diff
@@ -0,0 +1,316 @@
...
+    username, password = _get_auth(kwargs.pop('username', None),
+                                   kwargs.pop('password', None))
+    kwargs = clean_kwargs(**kwargs)
+    url = _build_url(endpoint, id=id)
+    ret = {
+        'comment': '',
+        'result': True,
+        'out': None
+    }
+    res = salt.utils.http.query(url,
```
cachedout: This will default to HTTP. Should we prefer HTTPS given that we're passing credentials in the clear here?

mirceaulinic: Definitely. Is there an option to `salt.utils.http.query` for that? I've never used it, I'm afraid. Though, the base URL is `PEERINGDB_URL = 'https://www.peeringdb.com/api'` and I have assumed that it would use that. Let me know and I'll update, thanks!

cachedout: Ah! You are right. Yup, we're good here. False alarm on my part.

cachedout (approved):

rallytime:
This is causing a related test to fail: https://jenkinsci.saltstack.com/job/pr-kitchen-ubuntu1604-py2/job/PR-48688/2/

@mirceaulinic Can you take a look?

mirceaulinic:
@rallytime it seem to me that is complain about the `clean_kwargs` utility function (https://github.com/saltstack/salt/blob/develop/salt/utils/args.py#L30) which doesn't have an example.

rallytime:
@mirceaulinic That'd be great! Thanks!

mirceaulinic:
I pushed https://github.com/saltstack/salt/pull/48688/commits/dd10817769be1a4503e7277c75e8cffe046ea644 - please let me know @rallytime if this don't suffice to pass the tests.

rallytime:
@mirceaulinic That test is picky and is specifically looking for `CLI Example`. I updated your PR here to get them running again.

Thanks again for his and all of your awesome work!

rallytime:
Hrm, I read the test failure incorrectly. Now it says:
```
No docstring:
  - peeringdb.clean_kwargs
```

mirceaulinic:
HI @rallytime - yes, `clean_kwargs` is imported from `salt.utils.args`. Bizarrely, it is imported in the exact same way in several other modules, but the tests didn't complain. Is there anything I can do? Thanks!

mirceaulinic:
It looks like a rebase helped, the lint check is passing now...

rallytime:
@mirceaulinic Hrm...still not passing. We just need to add that function to the list of functions to ignore. Can you add that [here](https://github.com/saltstack/salt/blob/fb5d63da164e1d816846f744f4cc92214789d15e/tests/integration/files/file/base/_modules/runtests_helpers.py#L51)?

mirceaulinic:
I pushed https://github.com/saltstack/salt/pull/48688/commits/fc000b8577a5ffe63f836c1d9f70fbedd8843697 let's see if that helps.

PR #48702, opened 2018-07-22
Correct inotify beacon description

ixs:
Correct outdated and misleading code comment about the config format.
Update the manpage to the current yaml format, rather than the
non-working 2015-release one.


cachedout (changes requested):
Hi @ixs. The man pages are actually built automatically so we don't need any changes made to them in a PR. You need to actually change the docs under the docs/ directory instead and the changes will be reflected in the man pages when Salt is prepared for release. Thanks.

rallytime:
Hi @ixs - Can you make these changes in the docs directory instead of in the man pages? Then we can get this change in.

Issue #48640, opened 2018-07-18
user.present group regression in 2017.7.5+ if gid_from_name is True


boltronics:
### Description of Issue/Question
There is a regression when upgrading from 2017.7.4 to 2017.7.5+ which results in a change in behaviour to the `user.present` functionality. My understanding is that this should never happen in a bugfix release unless it cannot be avoided due to a security issue.

I was building a new host with the following state applied (as rendered via `state.show_sls users`):

```
    myuser1:
        ----------
        __env__:
            base
        __sls__:
            users
        user:
            |_
              ----------
              fullname:
                  User 1
            |_
              ----------
              shell:
                  /bin/bash
            |_
              ----------
              gid_from_name:
                  True
            |_
              ----------
              uid:
                  None
            |_
              ----------
              gid:
                  myuser1
            |_
              ----------
              groups:
                  - adm
                  - sudo
                  - staff
            |_
              ----------
              home:
                  /home/myuser1
            |_
              ----------
              createhome:
                  None
            - present
            |_
              ----------
              order:
                  10018
```
Note I have replaced the id, fullname and gid arguments in all these examples to protect the innocent. I am focusing on just one specific user here for clarity.

This has worked for many years, and results in the following successful output on <=2017.7.4:

```
          ID: myuser1
    Function: user.present
      Result: True
     Comment: New user myuser1 created
     Started: 11:07:51.917005
    Duration: 82.221 ms
     Changes:   
              ----------
              fullname:
                  User 1
              gid:
                  1001
              groups:
                  - myuser1
                  - adm
                  - staff
                  - sudo
              home:
              homephone:
              name:
                  myuser1
              passwd:
                  x
              roomnumber:
              shell:
                  /bin/bash
              uid:
                  1001
              workphone:
```

The `gid_from_name` argument has historically functioned in more or less the same way as useradd's `-U` argument, which naturally makes sense given the module name is called `useradd`.

```
$ useradd --help | grep -- -U
  -U, --user-group              create a group with the same name as the user
```

The [original](https://docs.saltstack.com/en/2016.11/ref/states/all/salt.states.user.html#salt.states.user.present) documentation for this option in Salt said:
`If True, the default group id will be set to the id of the group with the same name as the user, Default is False.`
To me, this implies that it will create the user for the group - which has always been how useradd has functioned, and is the most logical behaviour since the user has already indicated they want the group.

[...]

PR #48704, opened 2018-07-23
Add usergroup support to user.present

boltronics:
### What does this PR do?
Replaces the seemingly redundant `gid_from_name` argument from `user.present` and replaces it with a `usergroup` argument that has the effect of adding the `--user-group` argument to the `useradd` command on GNU/Linux hosts.

I whitelisted it to GNU/Linux for now since I don't know which other operating systems support the `-U`/`--user-group` argument.

I tested it on an OpenSUSE 42.3 and Debian Stretch by creating a basic state sls:
```
test:
  user.present:
    - usergroup: True

test2:
  user.present:
    - usergroup: False
```

On both hosts the intended behaviour was observed.

https://github.com/saltstack/salt/issues/48640

### Previous Behavior
`gid_from_name` has been removed. AFAICT this command no longer has any practical benefit - please correct me if I'm overlooking something. The behaviour it intended to provide can be simulated just as easily by:
```
test:
  user.present:
    - gid: test
```
(where we want a `test` group to match a test user). This is much easier to read and understand. It's also clearer that the group will not be automatically created in this case.

Regardless of the `gid_from_name` option, the group will not be created since 2017.7.5 due to the regression referenced. This PR does not revert that regression (which is actually a pain to deal with on a stable release and IMO really should be fixed), but I believe this is a better approach for the long-term.

### New Behavior
We now mimic the existing `--user-group` option behaviour of the `useradd` command which I believe is closer to user expectations - even for users of distributions such as SuSE that don't typically use this.

If `usergroup: True` is set in the user state sls file, the `-U` argument will be passed to the `useradd` command causing the group with the same name as the user name to be created. If `usergroup: False` is set, the `-N`/`--no-user-group` argument is added instead. If left undefined, the defaults that the distribution or user typically defines in `/etc/default/useradd` and `/etc/login.defs` are used as usual.

### Tests written?

No. There are no existing tests for `gid_from_name` either. I'm not sure what the best way to do this would be, but I suspect it would involve having the test inspect the constructed `cmd` list in the `useradd.add` function to check for the presence of `-U` instead, as opposed to just inspecting a return value of a mocked command.

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/modules/useradd.py
+++ b/salt/modules/useradd.py
@@ -136,7 +136,8 @@ def add(name,
         createhome=True,
         loginclass=None,
         nologinit=False,
-        root=None):
+        root=None,
+        usergroup=None):
     '''
     Add a user to the minion
 
@@ -191,6 +192,9 @@ def add(name,
     root
         Directory to chroot into
 
+    usergroup
+        Create and add the user to a new primary group of the same name
+
     CLI Example:
 
     .. code-block:: bash
@@ -204,6 +208,10 @@ def add(name,
         cmd.extend(['-u', uid])
     if gid not in (None, ''):
         cmd.extend(['-g', gid])
+    elif usergroup:
+        cmd.append('-U')
+        if __grains__['kernel'] != 'Linux':
+            log.warning("'usergroup' is only supported on GNU/Linux hosts.")
     elif groups is not None and name in groups:
         defs_file = '/etc/login.defs'
         if __grains__['kernel'] != 'OpenBSD':
@@ -244,6 +252,11 @@ def add(name,
             except OSError:
                 # /etc/usermgmt.conf not present: defaults will be used
                 pass
+    # Setting usergroup to False adds the -N command argument. If
+    # usergroup is None, no arguments are added to allow useradd to go
+    # with the defaults defined for the OS.
+    if usergroup is False:
+        cmd.append('-N')
 
     if createhome:
         cmd.append('-m')
--- a/salt/modules/useradd.py
+++ b/salt/modules/useradd.py
@@ -148,7 +148,7 @@ def add(name,
         User ID of the new account
 
     gid
-        Name or ID of the primary group of the new accoun
+        Name or ID of the primary group of the new account
 
     groups
         List of supplementary groups of the new account
@@ -160,7 +160,7 @@ def add(name,
         Login shell of the new account
 
     unique
-        Allow to create users with duplicate
+        If not True, the user account can have a non-unique UID
 
     system
         Create a system account
--- a/salt/states/user.py
+++ b/salt/states/user.py
@@ -225,7 +225,7 @@ def _changes(name,
 def present(name,
             uid=None,
             gid=None,
-            gid_from_name=False,
+            usergroup=None,
             groups=None,
             optional_groups=None,
             remove_groups=True,
@@ -272,11 +272,6 @@ def present(name,
         or gid can be used. If not specified, and the user does not exist, then
         the next available gid will be assigned.
 
-    gid_from_name : False
-        If ``True``, the default group id will be set to the id of the group
-        with the same name as the user. If the group does not exist the state
-        will fail.
-
     allow_uid_change : False
         Set to ``True`` to allow the state to update the uid.
 
@@ -286,6 +281,17 @@ def present(name,
         Set to ``True`` to allow the state to update the gid.
 
         .. versionadded:: 2018.3.1
+
+    usergroup
+        If True, a group with the same name as the user will be created. If
+        False, a group with the same name as the user will not be created. The
+        default is distribution-specific. See the USERGROUPS_ENAB section of
+        the login.defs(5) man page.
+
+        .. note::
+            Only supported on GNU/Linux distributions
+
+        .. versionadded:: Fluorine
 
     groups
         A list of groups to assign the user to, pass a list object. If a group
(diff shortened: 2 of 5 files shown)


salt/modules/useradd.py
```diff
@@ -187,6 +192,8 @@ def add(name,
             except OSError:
                 # /etc/usermgmt.conf not present: defaults will be used
                 pass
+    if usergroup == False:
```
rares-pop: the default for 'usergroup' is None. Should you compare this to None too? I'm thinking to change this to: if not usergroup:

boltronics: There are three possibilities - we explicitly set `usergroup` to `False` to trigger `-N` to require no user group be created, we set it to `True` to trigger `-U` to require a user group to be created, or don't touch it at all which will go with the OS defaults - which differs depending on the distribution or user preferences. 

If you instead go with `if not usergroup`, you'll be force overriding the OS defaults (or overriding whatever the user configured the OS to do), so I imagine most people won't be too happy about that.

rares-pop: Ah, okay. So it's intended to use the OS defaults unless specified.

cachedout: Shouldn't we then document this so that a user knows when to set this to `False` and what behavior that will produce?

boltronics: @rares-pop Correct. I don't see any reason to change things the user didn't request. @cachedout Yep, fair call. Will do.

boltronics:
That has now been updated. I also rebased onto the latest develop branch.

rallytime:
@boltronics Can you fix these lint errors?

https://jenkinsci.saltstack.com/job/pr-lint/job/PR-48704/4/warnings52Result/

boltronics:
@rallytime Hopefully the lint error is sorted now. Thanks for pointing that out.

rallytime:
@boltronics Sure thing. The lint check is happy now, but now there are some related test failures with the user states and the git pillar tests: https://jenkinsci.saltstack.com/job/pr-kitchen-ubuntu1604-py3/job/PR-48704/5/#showFailuresLink

Can you take a look?

boltronics:
No worries, can do.

boltronics:
Sorry for the hold-up. I've been busy, but will try to get to this this week.

cachedout:
Hi @boltronics. Just checking in with you. :)

boltronics:
Sorry for the delay. I haven't forgotten, just taking care of higher priorities first. I hope to get to this soon.

cachedout:
Hi @boltronics. Just checking back in with you. :) Any updates here?

boltronics:
Hi @cachedout. Hopefully GitHub is working this time... I think I'm all good for fixing the tests (or I will be when GH is fixed so my pushed code doesn't go to `/dev/null`). The only one I'm concerned about is https://github.com/saltstack/salt/blob/develop/tests/integration/states/test_user.py#L131 `test_user_present_gid_from_name_default` as that doesn't make sense to me. Say I rename it to `test_user_present_usergroup_default`... it would only exist to test undefined behaviour. ie.

```
        if grains['os_family'] in ('Suse',):                                                                                                                                                       
            self.assertEqual(group_name, 'users')                                                                                                                                                    
        elif grains['os_family'] == 'MacOS':                                                                                                                                                         
            self.assertEqual(group_name, 'staff') 
```
Of course, Apple, SuSE, etc. could change their mind about their default behaviour at any point which would break the test.

If we set `usergroup` to `True` or `False` we can immediately see what we should expect, but if we set it to `None` (or just leave it unset) it'll do whatever the OS will do by default. We can try to guess what might happen, or we can try to get a better idea by parsing `/etc/login.defs`, etc. eg.

```
login_defs_path = '/etc/login.defs'                                   
# It looks like FreeBSD creates a usergroup by default. I'm           
# working on the assumption that this is usually true outside         
# of MacOS and Linux.                                                 
usergroup = True                                                      
# MacOS users' primary group defaults to staff (20), not the          
# name of user                                                        
if grains['os_family'] == 'MacOS':                                    
    usergroup = False                                                 
elif grains['kernel'] == 'Linux':                                     
    # It looks like login usually defaults to "no".                   
    try:                                                              
        with salt.utils.files.fopen(login_defs_path, 'r') as lfh:     
            for line in lfh:                                          
                if line.startswith('USERGROUPS_ENAB'):                
                    usergroup = line.strip().endswith('yes')          
                break                                                 
    except IOError:                                                   
        usergroup = False                                             
```

but I'm not sure I see a useful test here. We're only validating that we guessed correctly what the OS might do when the user didn't request anything specific, as opposed to testing the `usergroup` functionality worked as intended. We already have what will be a `test_user_present_usergroup` function, so I propose replacing `test_user_present_gid_from_name_default` with `test_user_present_usergroup_disabled` instead.

Please let me know your thoughts. In the meantime, I'll proceed with this proposal for this PR.

boltronics:
Hi @cachedout,

I'm trying to update my PR, but can't because GitHub is
completely broken right now. Unfortunately posting messages directly
to the PR is also broken, so I'm attempting one last time, this time
via e-mail.

This is my repo:

https://github.com/boltronics/salt/commits/develop-user_present-usergroup

Please let the tests run on it when you can. I think it fixes the
issues raised. Thanks.

boltronics:
I have just rebased against the current develop HEAD, and looks like Jenkins triggered successfully now.

...and it also looks like all the comments came through from yesterday that reportedly failed to submit. :smile:

salt/states/user.py
```diff
@@ -225,7 +225,7 @@ def _changes(name,
 def present(name,
             uid=None,
             gid=None,
-            gid_from_name=False,
```
cachedout: This is a breaking change. Is there a reason we can't preserve backward compatibility here?

boltronics: Because it doesn't actually do anything now IIRC. Just taking a quick read of https://github.com/saltstack/salt/issues/48640 to jog my memory.... the `gid_from_name` was already broken in a stable point release some months back, and AFAIK the bad commit that broke it was never reverted.

What do you think that argument should now do if we leave it in?

cachedout (approved):

cachedout:
@boltronics Sorry for the delay here. I think we're about ready to go with this. If you can resolve the merge conflict that snuck in then we'll get this merged.

thatch45:
Thanks @boltronics ! This looks great to me. And thanks for following the PR for so long!

PR #48705, opened 2018-07-23
Python 3.7 support

isbm:
### What does this PR do?

Adds Python 3.7 syntax changes support.

### Previous Behavior

Salt crashes on Python 3.7.

### New Behavior

Salt does not crash on Python 3.7.

### Tests written?

All tests should apply.


isbm:
@gtmanfred here. You have it. Let's see what happens. :wink:

dwoz (approved):
Very nice!

salt/client/api.py
```diff
@@ -120,7 +120,7 @@ def run(self, cmd):
         '''
         cmd = dict(cmd)  # make copy
         client = 'minion'  # default to local minion client
-        mode = cmd.get('mode', 'async')  # default to 'async'
```
thatch45: I don't think we need to change this line here, and won't this break the api?

salt/modules/mandrill.py
```diff
@@ -238,7 +238,7 @@ def send(message,
     data = {
         'key': params['api_key'],
         'message': message,
-        'async': async,
+        'asynchronous': asynchronous,
```
thatch45: Should the dict here still be 'async'?

thatch45 (changes requested):
There are a few places where I worry that the change will cause some api issues.

salt/modules/mandrill.py
isbm: Oh, you're right! And I didn't deprecated it properly as well.

salt/client/api.py
isbm: Yeah, you're right. I looked on it again, and this formats `something_async`, which we never changed to `something_asynchronous`. Fixed!

isbm:
@thatch45 your concerns are fixed now.

gtmanfred (changes requested):
This needs a big note at the top of the release notes for 2018.3.3 and develop about this change, since it is going to break backwards compatibility.

cachedout:
@gtmanfred Can you please elaborate on the nature of your concern about backward compatibility?

gtmanfred:
Sorry, i had talked to bo about it on irc, adding them now.

salt/client/mixins.py
```diff
@@ -516,7 +516,7 @@ def _gen_async_pub(self, jid=None):
         tag = salt.utils.event.tagify(jid, prefix=self.tag_prefix)
         return {'tag': tag, 'jid': jid}
 
-    def async(self, fun, low, user='UNKNOWN', pub=None):
+    def asynchronous(self, fun, low, user='UNKNOWN', pub=None):
```
gtmanfred: I was concerned about this one, but it looks like this is an internal api, what should be used is `cmd_async` which calls `self.async` or now `self.asynchronous`

gtmanfred: This should be fine, and not break anything.

salt/modules/cassandra_cql.py
```diff
@@ -361,7 +361,7 @@ def cql_query(query, contact_points=None, port=None, cql_user=None, cql_pass=Non
     return ret
 
 
-def cql_query_with_prepare(query, statement_name, statement_arguments, async=False,
+def cql_query_with_prepare(query, statement_name, statement_arguments, asynchronous=False,
```
gtmanfred: this will need to have `**kwargs` added, and pop `async` out of kwargs and add a warn_until message.

salt/modules/saltutil.py
```diff
@@ -986,15 +986,16 @@ def refresh_pillar():
         ret = False  # Effectively a no-op, since we can't really return without an event system
     return ret
 
+
 pillar_refresh = salt.utils.functools.alias_function(refresh_pillar, 'pillar_refresh')
 
 
-def refresh_modules(async=True):
+def refresh_modules(asynchronous=True):
```
gtmanfred: This will need the same kwargs added.

gtmanfred (changes requested):
just 2 changes, then we probably don't need the release note warning.

isbm:
@gtmanfred I've decided not to warn_until anything and just support `async` as it was before through the `kwargs` with the note inside. We should be fine.

isbm:
@thatch45 is there anything else of your concerns I can fix this, or you can approve this?

gtmanfred:
Cool, that works for me.

gtmanfred (approved):

gtmanfred:
is there anything extra in 2018.3 that needs to be covered, or does this backport?

isbm:
@gtmanfred since 2018.3 differs, I prefer here to be accurate and first agree on everything, then I will backport that myself there in another PR from this one.

@cachedout @rallytime @thatch45 @terminalmage any blocker here still? I'd like to proceed porting this to 2018.3 and get it done.

gtmanfred:
cool, looks good to me then.

You have one pylint warning https://jenkinsci.saltstack.com/job/pr-lint/job/PR-48705/1/warnings52Result/new/

gtmanfred:
And for some reason it does not like the version name, https://jenkinsci.saltstack.com/job/pr-kitchen-centos7-py2/job/PR-48705/2/testReport/junit/unit.utils.test_versions/VersionTestCase/test_spelling_version_name/

rallytime (approved):
Very nice.

cachedout (commented):
This looks to me like it changes the API for the RunnerClient method from `async` to `asynchronous`. If that's the case, we need to talk more about how to handle this since many, many people use that client in custom code.

salt/runner.py
```diff
@@ -236,17 +236,17 @@ def run(self):
                     low['kwarg']['orchestration_jid'] = async_pub['jid']
 
                 # Run the runner!
-                if self.opts.get('async', False):
```
gtmanfred: I think if we leave this as `async` and not switch it over to asyncronous, we shouldn't see anyone affected, because they should be using `self.cmd` and passing `async: True` to self.opts. which is what salt-run --async would do, or using `cmd_async`, which is how all the other python apis work.

rallytime:
@isbm Did you get a chance to see the most recent feedback here? I'd love to get this in soon.

isbm:
@cachedout @gtmanfred should be good now with the runner APIs

cachedout (approved):

rallytime:
@isbm There are some related test failures here. Can you take a look?

https://jenkinsci.saltstack.com/job/pr-kitchen-ubuntu1604-py2/job/PR-48705/5/

rallytime:
@isbm I have changed the base branch of this PR from develop to fluorine for inclusion in the Fluorine feature release.

gtmanfred (changes requested):
Looks like some of the things in the test suite are still not covered.

https://github.com/saltstack/salt/blob/fluorine/tests/support/case.py#L152

That will also need to be fixed.

rallytime (approved):

rallytime:
Those test failures are not related, so this is good to go!

PR #48707, opened 2018-07-23
Add several new functions to the napalm module

mirceaulinic:
This commit adds the functions required to gate functionality from the new
Fluorine modules [``nxos_api``](https://docs.saltstack.com/en/develop/ref/modules/all/salt.modules.nxos_api.html)
and [``pyeapi``](https://docs.saltstack.com/en/develop/ref/modules/all/salt.modules.arista_pyeapi.html).
NAPALM has all the underlying dependencies for all of these modules, which is
why we can reuse the code from the mentioned modules (plus ``junos``, ``netmiko``
etc.). This brings a lot of flexibility to the user, as without any further
configuration (or very little for the non-native NAPALM drivers) can execute
arbitrary RPC requests or show commands. This is due to the fact that the code
also forwards the already existing configuration details.
In other words, with this changes, starting with release Fluorine, NAPALM users
will be able to execute commands such as:
``salt '*' napalm.pyeapi_run_commands 'show version'`` with no other changes in
their environment.

These functions are important as they open the gates to extend Salt without
depending on the third party NAPALM library, which massively speeds up the
development of new features (particularly when we need features that wouldn't
have their place into the public NAPALM library).

### Previous Behavior

### New Behavior

### Tests written?

Yes/No

### Commits signed with GPG?

Yes/No


rallytime (approved):
👍

cachedout (approved):

Issue #48701, opened 2018-07-22
salt pkg.upgrade reports new packages despite installing none (zypper module)


miukumac:
### Description of Issue/Question
When issuing pkg.upgrade on any openSUSE or Enterprise Server, certain packages cause false reports of upgrading packages although nothing is actually installed. Such packages are always multiversion files such as kernel, crash-kmp and gpg-pubkey packages which many may be installed at the same time as well as custom packages that may have multiple versions such as nVidia or driver rpms.

Example:

```
zabbix:~ # salt '<sles12sp3host>' pkg.upgrade
sles12sp3host:
    ----------
    crash-kmp-default:
        ----------
        new:
            7.1.8_k4.4.92_6.30-4.6.2
        old:
            7.1.8_k4.4.131_94.29-4.8.1,7.1.8_k4.4.92_6.30-4.6.2
    gpg-pubkey:
        ----------
        new:
            edf0d733-54930778
        old:
            39db7c82-5847eb1f,50a3dd1c-50f35137,bf9a56ad-5758109f,e4e37c43-584f065d,e9e45fb5-57d42e04,edf0d733-54930778
    kernel-default:
        ----------
        new:
            4.4.138-94.39.1
        old:
            4.4.120-94.17.1,4.4.126-94.22.1,4.4.131-94.29.1,4.4.132-94.33.1,4.4.138-94.39.1
```

No new packages were installed, only reported as being "new" despite having already been installed on a previous run.

### Setup
(Please provide relevant configs and/or SLS files (Be sure to remove sensitive info).)
No special configuration is required - even out of the box configuration produces the same issue.

### Steps to Reproduce Issue
(Include debug logs if possible and relevant.)
Simply running salt 'host' pkg.upgrade reproduces the issue on all distribution versions that we currently use (SLE12SP3, SLE15)

### Versions Report
(Provided by running `salt --versions-report`. Please also mention any differences in master/minion versions.)
```
Salt Version:
           Salt: 2018.3.1
 
Dependency Versions:
           cffi: 1.5.2
       cherrypy: 3.6.0
       dateutil: Not Installed
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.8
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: 0.21.1
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.4.6
[...]

PR #48294, opened 2018-06-25
Handle packages with multiple versions properly with zypper 

meaksh:
### What does this PR do?
This PR fixes some issues introduced after merging https://github.com/saltstack/salt/pull/47678 . 

### What issues does this PR fix or reference?
- Exclude `gpg-pubkey` entries from `zypper.list_pkg`
- Fix `changes` on `pkg.installed` output when dealing with packages with multiple versions installed.
- Fix `changes` on `pkg.installed` output when packages name include "arch".

### Previous Behavior
On a SUSE systems, given this SLS state file: 
```yaml
my_tests:
  pkg.installed:
    - pkgs:
      - kernel-default.x86_64: 4.4.138-94.39.1
      - iotop.noarch: 0.6-3.10
      - apache2.x86_64: 2.4.23-29.18.2
      - apache2-doc: 2.4.23-29.18.2
      - perseus-dummy.i586: 1.1-1.1
      - orion-dummy: 1.1-1.1
```

Applying `my_tests` state will produce:
```yaml
local:
----------
          ID: my_tests
    Function: pkg.installed
      Result: True
     Comment: 6 targeted packages were installed/updated.
     Started: 16:05:56.124001
    Duration: 24153.705 ms
     Changes:   
              ----------
              apache2:
                  ----------
                  new:
                      2.4.23-29.18.2
                  old:
              apache2-doc:
                  ----------
                  new:
                      2.4.23-29.18.2
                  old:
              apache2-prefork:
                  ----------
                  new:
                      2.4.23-29.18.2
                  old:
              gpg-pubkey:
                  ----------
                  new:
                      e6e5a213-5b291a5f
                  old:
                      39db7c82-5847eb1f,8a7c64f9-5aaa93ca,e6e5a213-5b291a5f
              iotop:
                  ----------
                  new:
                      0.6-3.10
                  old:
              orion-dummy:
                  ----------
                  new:
                      1.1-1.1
                  old:

Summary for local
------------
Succeeded: 1 (changed=1)
Failed:    0
------------
Total states run:     1
Total run time:  24.154 s
```

Notice that `perseus-dummy.x586` package, as well as `kernel-default`, are actually installed but not reported on `changes` (and they should). Also the non-desired `gpg-pubkey` entry is not excluded from `changes` since it's not actually a real package.

### New Behavior
Given the same SLS file:

```yaml
local:
----------
          ID: my_tests
    Function: pkg.installed
      Result: True
     Comment: 6 targeted packages were installed/updated.
[...]

PR #48712, opened 2018-07-23
Remove old hack when reporting multiversion packages

meaksh:
### What does this PR do?
This PR fixes an issue with zypper `pkg.upgrade` after adding multiversion package support. 

https://github.com/saltstack/salt/issues/48701

### Previous Behavior
Even if the minion is upgraded, each call to `pkg.upgrade` report wrong output for multiversion packages:
```
# salt-call --local pkg.upgrade
local:
    ----------
    kernel-default:
        ----------
        new:
            4.4.21-69.1
        old:
            4.4.120-92.70.1,4.4.21-69.1
```

### New Behavior
The results are the expected since the minion is already upgraded:
```
# salt-call --local pkg.upgrade
local:
    ----------
```

### Tests written?

No

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/modules/zypper.py
+++ b/salt/modules/zypper.py
@@ -1481,11 +1481,6 @@ def upgrade(refresh=True,
     __zypper__(systemd_scope=_systemd_scope()).noraise.call(*cmd_update)
     _clean_cache()
     new = list_pkgs()
-
-    # Handle packages which report multiple new versions
-    # (affects only kernel packages at this point)
-    for pkg in new:
-        new[pkg] = new[pkg].split(',')[-1]
     ret = salt.utils.data.compare_dicts(old, new)
 
     if __zypper__.exit_code not in __zypper__.SUCCESS_EXIT_CODES:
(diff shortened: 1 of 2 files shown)


rallytime:
This should be against the 2018.3 branch, yes?

meaksh:
@rallytime - Not really. We added the "multiversion support" for Zypper into `develop` branch so we got rid of those hacks only on `develop` (like we see here: https://github.com/saltstack/salt/pull/48294/files#diff-0939df492f4987da56df53de1335fdb6L1355)

Let me also clarify that we backported the "multiversion support" to our Salt 2018.3.0 package and that's why it reaches openSUSE.

So this is good for `develop`. :+1:

rallytime:
@meaksh Looks like one of the tests needs to be fixed: https://jenkinsci.saltstack.com/job/pr-kitchen-ubuntu1604-py3/job/PR-48712/1/testReport/junit/unit.modules.test_zypper/ZypperTestCase/test_upgrade_success/

rallytime (approved):

PR #48723, opened 2018-07-24
Added check to ignore "-t" on unsupported versions of "zfs get"

clallen:
### What does this PR do?
Allows the zfs.get function to ignore the "type" argument on versions of Solaris that do not have the "-t" option for the "zfs get" subcommand.

None that I know of

### Previous Behavior
On recent versions of Oracle Solaris (within the last year or so), the zfs.get function will fail because at some point in that time Oracle removed the "-t" option from "zfs get".

### New Behavior
Causes the module to test for validity of "-t" and ignore the zfs.get function's "type" argument if "-t" is not supported, otherwise resume previous behavior.

### Tests written?
No

### Commits signed with GPG?
No

### DIFF
--- a/salt/modules/zfs.py
+++ b/salt/modules/zfs.py
@@ -1214,7 +1214,11 @@ def get(*dataset, **kwargs):
     fields.insert(1, 'property')
     opts['-o'] = ",".join(fields)
     if kwargs.get('type', False):
-        opts['-t'] = kwargs.get('type')
+        zfs_cmd = salt.utils.path.which('zfs')
+        res = __salt__['cmd.run_stderr']('{0} get -t'.format(zfs_cmd),
+                                         output_loglevel='quiet')
+        if 'invalid option' not in res:
+            opts['-t'] = kwargs.get('type')
     if kwargs.get('source', False):
         opts['-s'] = kwargs.get('source')
 


salt/modules/zfs.py
```diff
@@ -1214,7 +1214,11 @@ def get(*dataset, **kwargs):
     fields.insert(1, 'property')
     opts['-o'] = ",".join(fields)
     if kwargs.get('type', False):
-        opts['-t'] = kwargs.get('type')
+        zfs_cmd = salt.utils.path.which('zfs')
+        res = __salt__['cmd.run_stderr']('{0} get -t'.format(zfs_cmd),
+                                         output_loglevel='quiet')
+        if 'invalid option' not in res:
```
rares-pop: This seems like a hack. And a silent one too (no warnings/errors). I think it's fine to fail if the user tried to use a feature that's not available anymore.

Can you use grains to determine the version of the OS and not use the -t if it's not supported? Or the version of zfs (zfs get version).

Should you update the comments/docstrings for the method to specify when it works and when not?

clallen: "zfs get version" shows the version of the filesystem, not of the zfs binary.  The two can be different.
I'll try to find out what Oracle SRU this change was made in.

I agree there should be a warning and docstring updates, I'll get those in.

rares-pop: On my ubuntu box, I can find the version of zfs either by:
 - looking into /sys/module/zfs/version i.e. (cat /sys/module/zfs/version  and it replies 0.7.5-1ubuntu15)
 - doing a modinfo zfs

clallen: Right, however this is specifically targeted at Solaris, which does not have "/sys", and modinfo just shows output like this:

```
# modinfo
 Id Loadaddr   Size Info Rev Module Name
...
 13 12020250  eaed8   8   1  zfs (ZFS filesystem version 37)
 13 12020250  eaed8 286   1  zfs (ZFS storage pool)
...
```

If I can find the Oracle SRU where this change happened, I can add a check for that version, but it seems a lot more expedient just to test the zfs command directly.  This way it doesn't matter what OS you happen to be on, it will still work.

I suppose that since the "type" arg to zfs.get is optional it might suffice to just add a comment to the docstring stating that it won't work on versions of Solaris after whatever SRU this was introduced in.

I'll close this PR and see about updating the docstring.

PR #48724, opened 2018-07-24
test_vcenter_connection should raise system exception when it fails

guillermo-menjivar:
### What does this PR do?
This PR allows 

### Previous Behavior
The cloud.vmware.test_vscenter_connection function returned a exit 0 even when it failed to connect to vsphere

### New Behavior
The cloud.vmware.test_vscener_connection function now raises the SaltCloudSystemExit exception on failure. This allows for the shell to receive the correct "failed" exit code.

### Tests written?

No

### Commits signed with GPG?
No

### DIFF
--- a/salt/cloud/clouds/vmware.py
+++ b/salt/cloud/clouds/vmware.py
@@ -1470,7 +1470,7 @@ def test_vcenter_connection(kwargs=None, call=None):
         # Get the service instance object
         _get_si()
     except Exception as exc:
-        return 'failed to connect: {0}'.format(exc)
+        raise SaltCloudSystemExit('failed to connect: {0}'.format(exc))
 
     return 'connection successful'
 


rallytime (approved):

gtmanfred (changes requested):
can you set `__context__['retcode'] = 1` and see if it returns the retcode.

This function should be run with `salt-cloud -f test_vcenter_connection <something>` if someone chooses to use this from their CloudClient using client.action() and passing in the provider name, we don't want to raise a SystemExit, because then it will just exit fully.  It would be better to actually just set the return code for if salt-cloud exits after the failure, though i am not sure the __context__ retcode stuff is built into the command.

But overall, I am against this SystemExit.

videte47:
Ah cool thanks @gtmanfred for the feedback. I will amend the PR with your feedback.

cachedout:
@videte47 Were you going to amend this PR?

cachedout:
We didn't get a response from @videte47 so I'm going to close this. If @videte47 can come back to this, we can re-open it easily.

PR #29909, opened 2015-12-21
FreeBSD pkgng fix for non-interactive install.

abednarik:
In pkgng the way to install packages and assuming yes in pre and post install scripts is to use
-y, same as apt-get or yum, for example. Also env variable ASSUME_ALWAYS_YES=YES can be used, but
in this case, is a good idea to follow others package manager implementation.
Also remove the check where only -y is used if pkgng is executed in dryrun mode. In this case
there is no difference between dry run or not.

Fixes #27056

PR #29907, opened 2015-12-21
Presumably what was meant

bmcorser:
Otherwise it's invalid syntax

Issue #27056, opened 2015-09-11
pkgng provider on FreeBSD does not do BATCH=yes


oogali:
The pkgng provider on FreeBSD does not use the "BATCH=yes" environment setting when installing new packages.

This is problematic when installing packages that have an interactive post-install script (e.g. mail/postfix). The lack of the batch flag causes `pkg` to forever wait for a user response (which it will never receive), which in turn causes the minion to "hang".

```
# pkg install -qy postfix
===> Creating users and/or groups.
Using existing group 'mail'.
Creating group 'maildrop' with gid '126'.
Creating group 'postfix' with gid '125'.
Creating user 'postfix' with uid '125'.
Adding user 'postfix' to group 'mail'.
Would you like to activate Postfix in /etc/mail/mailer.conf [n]? ^C
```

But, because the minion is still reachable by the master, your highstate never completes or times out, and you then have to log in, kill the minion child, manually install the package in question, and then restart your highstate run.

```
# salt --versions-report
           Salt: 2015.5.2
         Python: 2.7.9 (default, Aug 27 2015, 01:28:02)
         Jinja2: 2.7.3
       M2Crypto: 0.22
 msgpack-python: 0.4.2
   msgpack-pure: Not Installed
       pycrypto: 2.6.1
        libnacl: Not Installed
         PyYAML: 3.11
          ioflo: Not Installed
          PyZMQ: 14.6.0
           RAET: Not Installed
            ZMQ: 4.1.3
           Mako: Not Installed
```

jfindlay:
@oogali, thanks for the report.

oogali:
Any traction? This is preventing me from fully using salt-cloud to orchestrate hosts in Amazon EC2.

jfindlay:
@oogali, not that I know of.  You could ask @cedwards or @kim0 if they think this would be an easy fix.

abednarik:
HI @oogali 

 II'm not expertise, but I'm trying to fix this. Would be great if you can test this, of course not in production. I do not have any FreeBSD around.,

#29707

abednarik:
Sorry for any inconvenience, but please read my last comment https://github.com/saltstack/salt/pull/29707
I was trying to fix the old pkd module and this one is related to the new one.

oogali:
@abednarik Adding -y is not enough. -y simply tells pkg not to prompt to install the package, but has no effect on interactive post-install scripts.

[...]

PR #48730, opened 2018-07-24
Fix batch install on FreeBSD using pkgng

fzipi:
For fully automated batch installing in FreeBSD you need to pass BATCH=true.

### What does this PR do?

Add a new option to `pkg install`, `batch=True`.

#27056, #29907, #29909

### Previous Behavior
When installing `postfix` in FreeBSD, it keeps asking for confirmation to change `mailer.conf`. When using the postfix formula, it does not work because installation is not automated (you can break things in production, etc.).

### New Behavior
Installation is succesful. You may require to configure additional steps, but at least the package is there.

### Tests written?

No

### Commits signed with GPG?

Yes


rallytime (approved):

rallytime:
Welcome @fzipi! Thank you for submitting this. This looks good to me, but I'm not super familiar with FreeBSD.

@abednarik You've been involved with FreeBSD support - can you review this?

abednarik:
Hi

 Sure  we can review this since we use FreeBSD. Will do it next Wensday.
I'm on vacations now.

Cheers

abednarik:
Hi @rallytime 

 This looks good to me.

Thanks.

PR #48733, opened 2018-07-24
Add ciscoconfparse execution module for various text config mainipulations

mirceaulinic:
Adding a simple execution module for various text manipulation when working with configuration style formatted according to a very poor "industry standard" Cisco IOS (space based hierarchy), but not limited to.

This helps identifying individual lines having a specific pattern, or if their children have a given pattern as well.

This can have unlimited use cases, and can be equally helpful from the command line, e.g., identify what interfaces are administratively shut down: `salt <net device> napalm.config_lines_w_child '^interface' '^\s?shutdown'`, as well as parsing various chunks of configuration from a Jinja template, etc.

The functions added to the `napalm` module are helpers to easier access the ciscoconfparse features applying directly on the running / startup / candidate config as preferred.

Also updating the autodoc for the `napalm` module (which I have renamed yesterday from `napalm.py` to `napalm_mod.py`: https://github.com/saltstack/salt/pull/48707, but forgot to update the RST build).


salt/modules/ciscoconfparse_mod.py
```diff
@@ -0,0 +1,443 @@
...
+                                                                  child_regex='stopbits')
+        for obj in objects:
+            print(obj.text)
+    '''
+    ccp = _get_ccp(config=config, config_path=config_path, saltenv=saltenv)
+    lines = ccp.find_objects_w_child(parent_regex, child_regex, ignore_ws=ignore_ws)
+    return lines
+
+
+def find_lines_w_child(config=None,
```
rallytime: Same comment here. Maybe `find_child_lines`?

salt/modules/ciscoconfparse_mod.py
```diff
@@ -0,0 +1,443 @@
...
+    lines = find_objects_w_child(config=config,
+                                 config_path=config_path,
+                                 parent_regex=parent_regex,
+                                 child_regex=child_regex,
+                                 ignore_ws=ignore_ws,
+                                 saltenv=saltenv)
+    return [line.text for line in lines]
+
+
+def find_objects_wo_child(config=None,
```
rallytime: What about `find_objects_no_child`? I'm not sure if that's better, but I think the `wo` is a tad confusing.

salt/modules/ciscoconfparse_mod.py
```diff
@@ -0,0 +1,443 @@
...
+                                                                   child_regex='stopbits')
+        for obj in objects:
+            print(obj.text)
+   '''
+    ccp = _get_ccp(config=config, config_path=config_path, saltenv=saltenv)
+    lines = ccp.find_objects_wo_child(parent_regex, child_regex, ignore_ws=ignore_ws)
+    return lines
+
+
+def find_lines_wo_child(config=None,
```
rallytime: Same comment here: `find_lines_no_child`?

salt/modules/napalm_mod.py
```diff
@@ -1286,3 +1293,156 @@ def rpc(command, **kwargs):
...
+
+        salt '*' napalm.config_find_lines '^interface Ethernet1\d'
+    '''
+    config_txt = __salt__['net.config'](source=source)['out'][source]
+    return __salt__['ciscoconfparse.find_lines'](config=config_txt,
+                                                 regex=regex)
+
+
+@depends(HAS_CISCOCONFPARSE)
+def config_lines_w_child(parent_regex, child_regex, source='running'):
```
rallytime: I think the functions in this file could use some new names as well, just like above. I won't comment on each one though, you get the idea.

rallytime (changes requested):
I like this in general, but I think some of the function names could be a little more direct. Let me know what you think.

salt/modules/ciscoconfparse_mod.py
```diff
@@ -0,0 +1,443 @@
...
+            -  no ip address
+    '''
+    lines = find_objects(config=config,
+                         config_path=config_path,
+                         regex=regex,
+                         saltenv=saltenv)
+    return [line.text for line in lines]
+
+
+def find_objects_w_child(config=None,
```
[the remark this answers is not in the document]
mirceaulinic: I used this name to preserve the nomeclature of the underlying library, e.g., http://www.pennington.net/py/ciscoconfparse/tutorial_parent_child.html#method-3-find-objects-w-child
I agree that name is a bit awkward, but I'm happy to change it as suggested. I'll let you reply with your thoughts given my reasoning above, then I can update if necessary.

salt/modules/napalm_mod.py
mirceaulinic: Please see https://github.com/saltstack/salt/pull/48733#discussion_r206498279 for my reasoning - happy to update if you think that's best - no problem at all!

salt/modules/ciscoconfparse_mod.py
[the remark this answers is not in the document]
rallytime: Ah, OK. If that's the way people are used to using these, then that is fine.

rallytime (approved):

PR #48739, opened 2018-07-24
Add a few functions for network configuration manipulation

mirceaulinic:
In this PR I am adding a few handy functions to the existing `net` module, for feature parity with the file module. It actually uses the corresponding functions from `file`, loading the result on the network device. As far as I know, besides Junos, no other network platform supports configuration replace natively. With this functions, we don't need to worry about that anymore, as the operation is done on the system side where we have (almost) unlimited flexibility: download the configuration from the network device, apply the patch, and finally load the result.

This helped me cleanup many inconsistent parts of the config (as very frequently a chunk of configuration may be referenced in multiple places), e.g., when converting the name of some prefix lists from upper to lower:

```bash
$ sudo salt 'junos-router' net.replace_pattern 'PREFIX-LIST' 'prefix-list'
junos-router:
    ----------
    already_configured:
        False
    comment:
        Configuration discarded.
    diff:
        [edit policy-options]
        +   prefix-list prefix-list {
        +       1.2.3.4/32;
        +   }
        [edit policy-options]
        -   prefix-list PREFIX-LIST {
        -       1.2.3.4/32;
        -   }
        [edit firewall filter in-filter term accept-icmp from prefix-list]
        +      prefix-list;
        -      PREFIX-LIST;
        [edit firewall filter filter-l0-in term accept-icmp from prefix-list]
        +      prefix-list;
        -      PREFIX-LIST;
    loaded_config:
    result:
        True
```

### DIFF
--- a/salt/modules/napalm_network.py
+++ b/salt/modules/napalm_network.py
@@ -32,6 +32,12 @@ import salt.utils.templates
 
 # Import 3rd-party libs
 from salt.ext import six
+try:
+    from salt.utils import fopen
+    from salt.utils import mkstemp
+except ImportError:
+    from salt.utils.files import fopen
+    from salt.utils.files import mkstemp
 
 # ----------------------------------------------------------------------------------------------------------------------
 # module properties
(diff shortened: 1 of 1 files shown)


rallytime (approved):
Nice 👍

Issue #45014, opened 2017-12-14
Add file windows Tests to Auto Test Suite


Ch3LL:
This task is to monitor the addition of five tests to the auto test suite:

- [ ] file.managed
- [ ] file.comment
- [ ] file.replace
- [ ] file.absent
- [ ] file.copy

There are unit tests but need to add integration tests

PR #48742, opened 2018-07-24
add windows integration tests for states.file

frogunder:

https://github.com/saltstack/salt/issues/45014


rallytime:
Hey @frogunder - There is a conflicting file in this change. Can you rebase and fix that?

tests/integration/states/test_file.py
```diff
@@ -3866,3 +3867,54 @@ def test_file_managed_http_source_skip_verify(self):
...
+
+
+@destructiveTest
+@skipIf(not salt.utils.is_windows(), 'windows test only')
+class WinFileTest(ModuleCase):
+    '''
+    Test for the file state on Windows
+    '''
+    def setUp(self):
+        self.test = self.run_state('file.managed', name='c:/testfile', makedirs=True, contents='Only a test')
```
Ch3LL: since we aren't using this `self.test` variable anywhere else in the test i think we can just remove this and remove the tearDown() `del self.test` and just call the `run_state`

tests/integration/states/test_file.py
```diff
@@ -3866,3 +3867,54 @@ def test_file_managed_http_source_skip_verify(self):
...
+
+    def tearDown(self):
+        del self.test
```
Ch3LL: since we are adding a file in the `setUp` we need to make sure are deleting that file in the `tearDown`

tests/integration/states/test_file.py
```diff
@@ -3866,3 +3867,54 @@ def test_file_managed_http_source_skip_verify(self):
...
+        '''
+        ret = self.run_state('file.copy', name='c:/testfile_copy', makedirs=True, source='c:/testfile')
+        self.assertTrue(ret)
+
+    def test_file_comment(self):
+        '''
+        Test file.comment on Windows
+        '''
+        self.run_state('file.comment', name='c:/testfile', regex='^Only')
+        file = 'c:/testfile'
```
Ch3LL: since it looks like you are using `c:/testfile` in a lot of cases here might be better to set as a class variable

Ch3LL (changes requested):

Ch3LL (approved):

PR #48762, opened 2018-07-25
Get rid of the template module and rework a couple of modules

mirceaulinic:
### What does this PR do?

The recently added modules for managing network platforms such as Arista EOS, Cisco Nexus, or a more generic Netmiko module, have a pretty common functionality to render a local or remote file and load the resulting contents on the device. To solve this, via https://github.com/saltstack/salt/pull/48267 I added a simple module named `template` for various template rendering on the fly.

That was a mistake on my side, as I overlooked existing features such as [`file.apply_template_on_contents`](https://docs.saltstack.com/en/develop/ref/modules/all/salt.modules.file.html#salt.modules.file.apply_template_on_contents) or [`cp.get_template`](https://docs.saltstack.com/en/develop/ref/modules/all/salt.modules.cp.html#salt.modules.cp.get_template) - the latter in particular does anything `template.render` would.
I think this is enough reason to revert my commits (https://github.com/saltstack/salt/pull/48267/commits/1afe33436a34af5e82ad81d141af6e7ce7deeeee and https://github.com/saltstack/salt/pull/48267/commits/e369357ff3f1b2daa4f0ed26f3ad528f301edea6), and similarly using the native Salt functions in the modules named above.

Sorry about that, but I think it's good getting rid of these right now as they are still in the develop branch, and this revert would hopefully simplify the maintenance workload.


rallytime (approved):
Oh, yes. Let's use those salt functions instead. :)

There are some lint failures that need to be fixed. Otherwise, 👍 

https://jenkinsci.saltstack.com/job/pr-lint/job/PR-48762/1/warnings52Result/

mirceaulinic:
Thanks @rallytime - should be fixed now.

Issue #48020, opened 2018-06-07
mine.get not returning data in a state.orchestrate sls 


calvinhp:
### Description of Issue/Question

`mine.get` is not working correctly from inside an orchestration.

### Setup

Add a `mine_function` to a pillar to return `network.ip_addrs`

Create an `orch.sls` with the following contents

```
{% set app_servers = salt.saltutil.runner('mine.get',
        tgt='*',
        fun='network.ip_addrs') %}

{% for app_server in app_servers.keys() %}

test-{{ app_server }}:
  salt.function:
    - tgt: {{ app_server }}
    - fun: test.ping

{% endfor %}
```

### Steps to Reproduce Issue

When running the following, `app_servers` will be an empty dictionary

`$ sudo salt-run state.orchestrate orch`

but if I run the same function via `saltutil.runner` it returns as I expect

`$ sudo salt-call saltutil.runner mine.get tgt='*' fun='network.ip_addrs'`

gives me this output

```
local:
    ----------
    minion01:
        - 10.192.2.58
    minion02:
        - 10.192.2.59
```

### Versions Report

```
Salt Version:
           Salt: 2018.3.0

Dependency Versions:
           cffi: Not Installed
       cherrypy: 3.5.0
       dateutil: 2.4.2
      docker-py: Not Installed
          gitdb: 0.6.4
      gitpython: 1.0.1
          ioflo: Not Installed
         Jinja2: 2.8
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: 1.0.3
   msgpack-pure: Not Installed
 msgpack-python: 0.4.6
   mysql-python: Not Installed
      pycparser: Not Installed
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: Not Installed
         Python: 2.7.12 (default, Nov 20 2017, 18:23:56)
   python-gnupg: Not Installed
         PyYAML: 3.11
          PyZMQ: 15.2.0
           RAET: Not Installed
          smmap: 0.9.0
        timelib: 0.2.4
        Tornado: 4.2.1
            ZMQ: 4.1.4

System Versions:
           dist: Ubuntu 16.04 xenial
         locale: UTF-8
        machine: x86_64
        release: 4.4.0-97-generic
         system: Linux
        version: Ubuntu 16.04 xenial
```

Ch3LL:
was this working previously? are you just seeing it in 2018.3.0?

[...]

PR #44890, opened 2017-12-08
Add grains (and optionally pillar) to salt.cmd runner

terminalmage:
Not having grains was an oversight, it causes a lot of errors in processing the `__virtual__` functions. Additionally, the `__pillar__` dunder did not exist, so anything that referenced it would result in a traceback.

PR #48765, opened 2018-07-25
[2018.3] Fix mine.get not returning data

FedericoCeratto:
### What does this PR do?
Implements a workaround for #48020
Ported from https://github.com/openSUSE/salt/pull/102

#48020

### Previous Behavior
mine.get would return no nodes

### New Behavior
mine.get returns the correct nodes

### Tests written?

No

### Commits signed with GPG?

No

### DIFF
--- a/salt/utils/minions.py
+++ b/salt/utils/minions.py
@@ -239,12 +239,12 @@ class CkMinions(object):
         Retreive complete minion list from PKI dir.
         Respects cache if configured
         '''
-        if self.opts.get('__role') == 'master' and self.opts.get('__cli') == 'salt-run':
-            # Compiling pillar directly on the master, just return the master's
-            # ID as that is the only one that is available.
-            return [self.opts['id']]
         minions = []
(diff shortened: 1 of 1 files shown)


gtmanfred (commented):
This was added in https://github.com/saltstack/salt/pull/44890

I would like @terminalmage to look at it, since that was originally added by him.

terminalmage (changes requested):
This is a good change. I mistook the cause for an error that is fired later on in the function when `salt.cmd` is used.

To fix the _actual_ issue I was trying to solve, can you please add the following after `pki_cache_fn` is defined? (line 243):

```python
try:
    os.makedirs(os.path.dirname(pki_cache_fn))
except OSError:
    pass
```

FedericoCeratto:
Updated and rebased.

terminalmage (approved):
Excellent!

cachedout:
@terminalmage This is a partial revert of #44890.

PR #48775, opened 2018-07-25
Log error when reactor tasks go to a full queue

### LINKED ISSUE #46431: Concurrency issues in reactor & orchestration with long running  orchestrations simultaneously targeting  different minions
It seems that there is some kind of limit in the concurrent number of either reactors or orchestrations regardless of the number of worker threads configured in the saltmaster configuration file
### Setup
I am showing the config used to reproduce the issue.
Reactor configured in master :

```
reactor: [ {"debug/minion-created": ["/srv/salt/reactor/debug_minion_creation.sls"]} ]

# cat /srv/salt/reactor/debug_minion_creation.sls
debug_orchestration:
  runner.state.orchestrate:
    - args:
      - mods: orch.debug
      - pillar:
          event_tag: {{ tag }}
          event_data: {{ data|json }}

###############################
# cat /srv/salt/orch/debug.sls 
{% import "orch/map.jinja" as v with context %}

debug_event_task_on_master:
  salt.function:
    - name: cmd.run
    - tgt: "saltmaster*"
    - ret: rawfile_json
    - arg:
      - "echo $HOSTNAME ORCHESTRATION minion={{v.minion_id}} >>/tmp/debug_orchestration.log"

debug_event_task_on_target:
  salt.function:
    - name: cmd.run
    - tgt: {{ v.minion_id }}
    - ret: rawfile_json
    - arg:
      - "echo $HOSTNAME ORCHESTRATION minion={{v.minion_id}} >>/tmp/debug_orchestration.log"



state_append_date:
  salt.state:
    - sls: append_date
    - tgt: {{ v.minion_id }}
    - ret: rawfile_json

# cat /srv/salt/append_date.sls 

debug_command_start:
  cmd.run:
    - name: "echo ORCH_STATE_START Minion $(date) >>/tmp/debug_orchestration.log"

just_wait:
  cmd.run:
    - name: sleep 1800

debug_command_end:
  cmd.run:
    - name: "echo ORCH_STATE_END Minion $(date) >>/tmp/debug_orchestration.log"

```

### Steps to Reproduce Issue
Generate the event on all minions
``` 
salt "*"  cmd.run "salt-call  event.send --out json 'debug/minion-created'"
```
At this point, usually a maximum of 10 minions will have their orchestrations running. This is verifiable by either checking all the /tmp/debug_orchestration.log files on all minions involved OR by counting the lines in the debug_orchestration_log in the saltmaster itself. To be noted, on the saltmaster we do an append to that file on each run of the orchestration of any minion.
It *seems on this simplified use case* that once the first batch finishes, another batch will eventually start, but in our more complex real life setup we often end up with orchestrations not starting at all not sure yet how to also reproduce this, too. Maybe a longer yet state that will trigger some timeout in the master to discard the event and the queue of orchestrations to run? In any case, if the saltmaster has activity, the orchestrations do not start any more. For us, this is a significant issue because I cannot seem to find what parameter controls this concurrency bottleneck and it's not even clear to me if this is a reactor or orchestration related limit. The fact that in our scenario some orchestrations never start is even more problematic (but let's not focus on that last part right now as I cannot reliably reproduce the total loss of an orchestration, I just hope that once I can adjust the concurrency the apparent event loss will also go away)

### Versions Report
salt-master-2017.7.4-1.el7.noarch
# salt --versions-report
```
Salt Version:
           Salt: 2017.7.4
 
Dependency Versions:
           cffi: Not Installed
       cherrypy: Not Installed
       dateutil: Not Installed
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.7.2
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.4.8
   mysql-python: Not Installed
      pycparser: Not Installed
       pycrypto: 2.6.1
   pycryptodome: 3.4.3
         pygit2: Not Installed
         Python: 2.7.5 (default, Nov  6 2016, 00:28:07)
   python-gnupg: 0.3.8
         PyYAML: 3.12
          PyZMQ: 15.3.0
           RAET: Not Installed
          smmap: Not Installed
        timelib: Not Installed
        Tornado: 4.2.1
            ZMQ: 4.1.4
 
System Versions:
           dist: centos 7.3.1611 Core
         locale: UTF-8
        machine: x86_64
        release: 3.10.0-514.16.1.el7.x86_64
         system: Linux
        version: CentOS Linux 7.3.1611 Core
```

geekinutah:
Fixes #46431

### What does this PR do?
Check return values of fire_async, the only possible return of False is if queue is full.

46431

### Previous Behavior
Silently ignore failure to inject a task into a full queue

### New Behavior
Only adds a log message

### Tests written?
No

### Commits signed with GPG?
No


salt/utils/reactor.py
```diff
@@ -417,7 +417,16 @@ def run(self, low):
             # and kwargs['kwarg'] contain the positional and keyword arguments
             # that will be passed to the client interface to execute the
             # desired runner/wheel/remote-exec/etc. function.
-            l_fun(*args, **kwargs)
+            ret = l_fun(*args, **kwargs)
+
+            if ret is False:
+                log.error('Reactor \'%s\' failed  to execute %s \'%s\': '
+                            'TaskPool queue is full!'
```
terminalmage: There is no space between this sentence and the next. A space should be added before `Consider`.

terminalmage (changes requested):
One minor change needed in the logging.

salt/utils/reactor.py
geekinutah: Done

terminalmage (approved):

salt/utils/reactor.py
geekinutah: Sure, from what I can see, False is only returned when the queue is full. As per current behavior, we will right now silently fail to insert something on the queue. The only behavior I am adding is a log message when the queue is full. Otherwise, continue as per today.

cachedout (approved):

Issue #35505, opened 2016-08-16
errors in minion logs if dmidecode is not installed


white-hat:
### Description of Issue/Question
I guess salt is trying to get BIOS info for grains even if dmidecode is not installed

```
2016-08-16 16:44:44,774 [tornado.application                                  ][ERROR   ][1452] Exception in callback <functools.partial object at 0x3889a48>
Traceback (most recent call last):
  File "/usr/lib64/python2.6/site-packages/tornado/ioloop.py", line 591, in _run_callback
    ret = callback()
  File "/usr/lib64/python2.6/site-packages/tornado/stack_context.py", line 274, in null_wrapper
    return fn(*args, **kwargs)
  File "/usr/lib64/python2.6/site-packages/tornado/ioloop.py", line 597, in <lambda>
    self.add_future(ret, lambda f: f.result())
  File "/usr/lib64/python2.6/site-packages/tornado/concurrent.py", line 214, in result
    raise_exc_info(self._exc_info)
  File "/usr/lib64/python2.6/site-packages/tornado/gen.py", line 230, in wrapper
    yielded = next(result)
  File "/usr/lib/python2.6/site-packages/salt/minion.py", line 1791, in handle_event
    self.module_refresh(notify=data.get('notify', False))
  File "/usr/lib/python2.6/site-packages/salt/minion.py", line 1635, in module_refresh
    self.functions, self.returners, _, self.executors = self._load_modules(force_refresh, notify=notify)
  File "/usr/lib/python2.6/site-packages/salt/minion.py", line 1082, in _load_modules
    self.opts['grains'] = salt.loader.grains(self.opts, force_refresh, proxy=proxy)
  File "/usr/lib/python2.6/site-packages/salt/loader.py", line 703, in grains
    ret = fun()
  File "/usr/lib/python2.6/site-packages/salt/grains/core.py", line 1468, in os_data
    grains.update(_hw_data(grains))
  File "/usr/lib/python2.6/site-packages/salt/grains/core.py", line 1897, in _hw_data
    'biosversion': __salt__['smbios.get']('bios-version'),
  File "/usr/lib/python2.6/site-packages/salt/modules/smbios.py", line 88, in get
    val = _dmidecoder('-s {0}'.format(string)).strip()
  File "/usr/lib/python2.6/site-packages/salt/modules/smbios.py", line 329, in _dmidecoder
    return salt.modules.cmdmod._run_quiet('{0} {1}'.format(DMIDECODER, args))
[...]

PR #48781, opened 2018-07-26
Check dmidecoder executable on each "smbios" call to avoid race condition

meaksh:
### What does this PR do?
This PR fixes a race condition happening during core grains generation in case `dmidecode` or `smbios` are installed/removed during `salt-minion` execution.

As described on #35505, `smbios.DMIDECODER` is fixed to the `dmidecode` or `smbios` executables, and this is happening only **once while the module is loaded**. Then there are some calls to `smbios.get` performed while calculating the "core" grains which produces a race condition in case the results of `salt.utils.path.which_bin(['dmidecode', 'smbios'])` changes during execution time.

Hopefully fixes https://github.com/saltstack/salt/issues/35505

### Previous Behavior
The situation appears in case neither `dmidecode` nor `smbios` are installed at the time of `salt-minion` is started, therefore `smbios.DMIDECODER` is set to `None`. Then, at some point, either `dmidecode` or `smbios` are installed on the minion and eventually the grains will be recalculated.
At this particular moment, when the "core" grains are being recalculated, https://github.com/saltstack/salt/blob/2018.3/salt/grains/core.py#L2310 is checking again if executables are now available, and if so, it would produce a call to `smbios.get` which will crash since `smbios.DMIDECODER` is set to `None`:

```
ERROR: Unable to run command '['None', '-s', 'bios-version']' with the context '{'timeout': None, 'with_communicate': True, 'shell': False, 'bg': False, 'stderr': -2, 'env': {'LC_NUMERIC': 'C', 'NOTIFY_SOCKET': '/run/systemd/notify', 'LC_MESSAGES': 'C', 'LC_IDENTIFICATION': 'C', 'LC_MONETARY': 'C', 'LC_COLLATE': 'C', 'LC_CTYPE': 'C', 'LC_ADDRESS': 'C', 'LC_MEASUREMENT': 'C', 'LC_TELEPHONE': 'C', 'LC_PAPER': 'C', 'LC_NAME': 'C', 'PATH': '/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin', 'LC_TIME': 'C'}, 'stdout': -1, 'close_fds': True, 'stdin': None, 'cwd': '/root'}', reason: [Errno 2] No such file or directory
```

### New Behavior
Now `smbios.DMIDECODER` is refreshed on each `smbios` module call to prevent the race condition with the "core" grains. Besides of that, this PR introduces also a comprehensive error message in case of crashing:

```
2018-07-26 17:04:40,063 [salt.modules.smbios:337 ][ERROR   ][16331] SMBIOS: neither dmidecode nor smbios found!
```

### Tests written?

No

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/modules/smbios.py
+++ b/salt/modules/smbios.py
@@ -32,11 +32,15 @@ log = logging.getLogger(__name__)
 
 DMIDECODER = salt.utils.path.which_bin(['dmidecode', 'smbios'])
 
+def _refresh_dmidecoder():
+    global DMIDECODER
+    DMIDECODER = salt.utils.path.which_bin(['dmidecode', 'smbios'])
 
 def __virtual__():
     '''
     Only work when dmidecode is installed.
     '''
+    _refresh_dmidecoder()
     if DMIDECODER is None:
         log.debug('SMBIOS: neither dmidecode nor smbios found!')
         return (False, 'The smbios execution module failed to load: neither dmidecode nor smbios in the path.')
@@ -328,6 +332,7 @@ def _dmidecoder(args=None):
     '''
     Call DMIdecode
     '''
+    _refresh_dmidecoder()
     if DMIDECODER is None:
         raise CommandExecutionError('SMBIOS: neither dmidecode nor smbios found!')
 
--- a/salt/modules/smbios.py
+++ b/salt/modules/smbios.py
@@ -19,6 +19,7 @@ import re
 
 # Import salt libs
 import salt.utils.path
+from salt.exceptions import CommandExecutionError
 
 # Solve the Chicken and egg problem where grains need to run before any
 # of the modules are loaded and are generally available for any usage.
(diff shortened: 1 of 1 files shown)


rallytime:
@meaksh Can you fix up the lint failures listed here?

https://jenkinsci.saltstack.com/job/pr-lint/job/PR-48781/1/warnings52Result/package.2029219534/

meaksh:
@rallytime - pylint issues fixed :+1:

salt/modules/smbios.py
```diff
@@ -32,10 +33,16 @@
 DMIDECODER = salt.utils.path.which_bin(['dmidecode', 'smbios'])
 
 
+def _refresh_dmidecoder():
+    global DMIDECODER
+    DMIDECODER = salt.utils.path.which_bin(['dmidecode', 'smbios'])
```
isbm: Why there is a need of this global variable at all? And this function is also not needed. I would discard this PR entirely and just refactor it the way that `which_bin` is just called.

isbm (changes requested):
I will follow-up on this PR differently.

isbm:
@rallytime Please close this. There is a better version of it here: https://github.com/saltstack/salt/pull/48814

rallytime:
Closed in favor of #48814

PR #48783, opened 2018-07-26
Only run lint checks against changed files

KaiSforza:
### What does this PR do?

Changes the behaviour of `pr-lint` to only check the files that git says changed between the merge target and the PR branch.

Re-does #48610 

### Previous Behavior
Runs pylint on every file in the repo

### New Behavior
Only runs on changed files using find and git diff

### Tests written?

These are the tests.

### Commits signed with GPG?

Yes

@gtmanfred @dubb-b

### DIFF
--- a/.ci/lint
+++ b/.ci/lint
@@ -29,13 +29,13 @@ pipeline {
             parallel {
                 stage('salt linting') {
                     steps {
-                        sh 'eval "$(pyenv init -)"; tox -e pylint-salt | tee pylint-report.xml'
+                        sh 'eval "$(pyenv init -)"; tox -e pylint-salt $(find salt/ -name "*.py" -exec git diff --name-only "origin/$CHANGE_TARGET" "origin/$BRANCH_NAME" setup.py {} +) | tee pylint-report.xml'
                         archiveArtifacts artifacts: 'pylint-report.xml'
                     }
                 }
                 stage('test linting') {
                     steps {
-                        sh 'eval "$(pyenv init -)"; tox -e pylint-tests | tee pylint-report-tests.xml'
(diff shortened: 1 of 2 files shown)


gtmanfred (changes requested):
can you fix this pylint error in the testprogram.py file?

```
16:56:23 [test linting] ************* Module tests.integration.utils.testprogram
16:56:23 [test linting] tests/integration/utils/testprogram.py:601: [W0702(bare-except), TestSaltProgram] No exception type(s) specified
```

rallytime:
I don't think this is quite right - the lint job says it passed, but it has a warning in the list: https://jenkinsci.saltstack.com/job/pr-lint/job/PR-48783/1/

My next question is why is this running on that `testprogram` file when you didn't change that file in this PR? Might be jumping the gun on that, but I am curious about how this should be working.

KaiSforza:
@rallytime the warning/pass is a whole other issue though. Currently the run isn't using the updated pylint script, so it's running against everything.

KaiSforza:
@rallytime also i think that if there are no changes to the files we're finding against, it'll run against everything, at least for the test run.

rallytime:
> the warning/pass is a whole other issue though.

I'm not quite following. What other issue? This is working in PRs as I would expect it to - we get failures when there are lint issues. See https://github.com/saltstack/salt/pull/48781 as an example. If this is changing those results, then that's not good.

KaiSforza:
@rallytime The failures happen if we have more than one issue, but with one it was below the threshold. I fixed it in a commit here.

rallytime:
@KaiSforza Perfect! Thank you :)

gtmanfred (approved):

Issue #45007, opened 2017-12-14
Add timezone windows Tests to Auto Test Suite


Ch3LL:
This task is to monitor the addition of four tests to the auto test suite:

- [x] timezone.get_zone
- [x] timezone.get_offset
- [ ] timezone.set_zone
- [ ] timezone.zone_compare

There are unit tests but need to add integration tests

PR #48788, opened 2018-07-26
Add timzeone windows integration tests and fix get_zone

Ch3LL:
### What does this PR do?
Adds timezone integration tests for windows and fixes an issue found while writing the tests.

Here is the issue:

```
PS C:\testing> salt-call --local timezone.set_zone 'America/Denver' -lerror
local:
    True
PS C:\testing> salt-call --local timezone.get_zone
[INFO    ] Executing command ['tzutil', '/g'] in directory 'C:\Users\Administrator'
local:
    America/Inuvik
PS C:\testing>
```

I believe this occurs because the mapping here: https://github.com/saltstack/salt/blob/v2017.7.7/salt/modules/win_timezone.py#L114-L121

has multiple 'Mountain Standard Times'

I changed it to use `tzlocal` but i'm not certain this is the *best* way to grab the windows timezone accurately and I am definitely up for other ideas on how to approach this on the windows side of things and was hoping to get feedback from @dwoz and @twangboy 

https://github.com/saltstack/salt/issues/45007

### DIFF
--- a/salt/modules/win_timezone.py
+++ b/salt/modules/win_timezone.py
@@ -10,7 +10,7 @@ import logging
 import re
 
 try:
-    import tzlocal  # pylint: disable=unused-import
+    import tzlocal
     HAS_TZLOCAL = True
 except ImportError:
     HAS_TZLOCAL = False
--- a/salt/modules/win_timezone.py
+++ b/salt/modules/win_timezone.py
@@ -10,7 +10,7 @@ import logging
 import re
 
 try:
-    import tzlocal  # pylint: disable=import-error
+    import tzlocal  # pylint: disable=unused-import
     HAS_TZLOCAL = True
 except ImportError:
     HAS_TZLOCAL = False
--- a/salt/modules/win_timezone.py
+++ b/salt/modules/win_timezone.py
@@ -10,7 +10,7 @@ import logging
 import re
 
 try:
-    import tzlocal
+    import tzlocal  # pylint: disable=import-error
     HAS_TZLOCAL = True
 except ImportError:
     HAS_TZLOCAL = False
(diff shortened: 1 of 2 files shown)


twangboy (approved):

rallytime:
@Ch3LL You've got a lint error here: https://jenkinsci.saltstack.com/job/pr-lint/job/PR-48788/1/warnings52Result/file.-580256476/

Don't worry about the one in the file you didn't change. That's fixed in another PR.

salt/modules/win_timezone.py
```diff
@@ -9,6 +9,12 @@
 import logging
 import re
 
+try:
+    import tzlocal
```
rares-pop: we seem to align on doing this like:
try:
import tzlocal
except ImportError:
tzlocal = None

And then instead of HAS_TZLOCAL just do - if tzlocal is not None:

Ch3LL: i see many instances in the code of using `HAS_*` in the same manner. Do you see an advantage to using None instead?

Ch3LL: well it seems to get rid of the pylint error. @rallytime do we have a standard for this? should i be using what @rares-pop recommended or just adding a pylint exception: `# pylint: disable=import-error`

rares-pop: I had to impression that this is the new way of doing it.
I have seen @isbm recommending that in some of the PRs he reviewed.

The only advantage I've seen is one less variable. Maybe @isbm can give his insight on this?

rallytime: We do not have a standard way of doing it, and I am fine with either way. It's really just a preference at this point.

rallytime (approved):

PR #48796, opened 2018-07-27
Remove fake su used in integration tests

Ch3LL:
### What does this PR do?
I found that this `su` in mockbin is causing some issues with mac tests in 2018.3 as documented in issue https://github.com/saltstack/salt-jenkins/issues/1033 . I just need to see the tests run on this. I tried running some tests with kitchen locally but I believe I have some slowness network issues and it was taking more then 3 hours. Instead of running locally at this point wanted to push as a PR to see the test results. I am not necessarily proposing this change yet.

https://github.com/saltstack/salt-jenkins/issues/1033


Ch3LL:
okay it looks like only this test failed: `integration.ssh.test_state.SSHStateTest.test_state_sls_id` which also fails without this commit. I have created an issue here: https://github.com/saltstack/salt-jenkins/issues/1043  and will look into this first thing next week. So this test failure is not related to removing this `su` command.

The cent7 py2 test failures seem unrelated because the build failed early so i'm re-running that job right now just to make sure: https://jenkinsci.saltstack.com/job/pr-kitchen-centos7-py2/job/PR-48796/2/

rallytime (approved):

PR #48803, opened 2018-07-27
Support for execution modules and states mount on AIX

dmurphy18:
### What does this PR do?
Provides support for mount on AIX for execution modules and states

https://github.com/saltstack/zh/issues/1407

### Previous Behavior
mount.active functioned on AIX

### New Behavior
Full support for mount functionality in execution modules and states on AIX

### Tests written?
Yes

### Commits signed with GPG?
No

### DIFF
--- a/salt/modules/mount.py
+++ b/salt/modules/mount.py
@@ -1646,7 +1646,7 @@ def filesystems(config='/etc/filesystems'):
 
     ret_dict = _filesystems(config)
     if ret_dict:
-        ret_key = ret_dict.keys()[0]
+        ret_key = next(iter(ret_dict.keys()))
         ret = {ret_key: dict(ret_dict[ret_key])}
 
     return ret
--- a/salt/modules/mount.py
+++ b/salt/modules/mount.py
@@ -5,7 +5,6 @@ Salt module to manage Unix mounts and the fstab file
 
 # Import python libs
 from __future__ import absolute_import, print_function, unicode_literals
-from collections import OrderedDict
 import os
 import re
 import logging
@@ -18,6 +17,7 @@ import salt.utils.path
 import salt.utils.platform
 import salt.utils.mount
 import salt.utils.stringutils
+from salt.utils.odict import OrderedDict
 from salt.exceptions import CommandNotFoundError, CommandExecutionError
 
 # Import 3rd-party libs
--- a/salt/modules/mount.py
+++ b/salt/modules/mount.py
@@ -5,6 +5,7 @@ Salt module to manage Unix mounts and the fstab file
 
 # Import python libs
 from __future__ import absolute_import, print_function, unicode_literals
+from collections import OrderedDict
 import os
 import re
 import logging
@@ -22,6 +23,7 @@ from salt.exceptions import CommandNotFoundError, CommandExecutionError
 # Import 3rd-party libs
 from salt.ext import six
 from salt.ext.six.moves import filter, zip  # pylint: disable=import-error,redefined-builtin
+
 
 # Set up logger
 log = logging.getLogger(__name__)
@@ -430,6 +432,138 @@ class _vfstab_entry(object):
             if entry[key] != value:
                 return False
         return True
+
+
+class _filesystems_entry(object):
+    '''
+    Utility class for manipulating filesystem entries. Primarily we're parsing,
+    formatting, and comparing lines. Parsing emits dicts expected from
+    fstab() or raises a ValueError.
+
+    Note: We'll probably want to use os.normpath and os.normcase on 'name'
+    '''
+
+    class ParseError(ValueError):
+        '''
+        Error raised when a line isn't parsible as an fstab entry
+        '''
+
+    filesystems_keys = ('device', 'name', 'fstype', 'vfstype', 'opts', 'mount')
+
+    # preserve data format
+    compatibility_keys = ('device', 'name', 'fstype', 'vfstype', 'opts', 'mount', 'account', 'boot', 'check', 'free', 'nodename', 'quota', 'size', 'vol', 'log')
+
+    @classmethod
+    def dict_from_lines(cls, lines, keys=filesystems_keys):
+        if len(lines) < 2:
+            raise ValueError('Invalid number of lines: {0}'.format(lines))
+
+        if len(keys) < 6:
+            raise ValueError('Invalid key array: {0}'.format(keys))
+
+        blk_lines = lines
+        orddict = OrderedDict()
+        orddict['name'] = blk_lines[0].split(':')[0].strip()
+        blk_lines.pop(0)
+        for line in blk_lines:
+            if line.startswith('#'):
+                raise cls.ParseError("Comment!")
+
+            comps = line.split('= ')
+            if len(comps) != 2:
+                raise cls.ParseError("Invalid Entry!")
+
+            orddict[comps[0].strip()] = comps[1].strip()
+
+        return orddict
+
+    @classmethod
+    def dict_from_cmd_line(cls, ipargs, keys=filesystems_keys):
+        if len(keys) < 1:
+            raise ValueError('Invalid key array: {0}'.format(keys))
+
+        cmdln_dict = ipargs
+        for key in keys:
+            # ignore unknown or local scope keys
+            if key.startswith('__'):
+                continue
+
+            if key not in _filesystems_entry.compatibility_keys:
+                cmdln_dict[key] = keys[key]
+
+        return cmdln_dict
+
+    @classmethod
+    def from_line(cls, *args, **kwargs):
+        return cls(** cls.dict_from_cmd_line(*args, **kwargs))
+
+    @classmethod
+    def dict_to_lines(cls, fsys_dict_entry):
+        entry = fsys_dict_entry
+        strg_out = entry['name'] + ':' + os.linesep
+        for k, v in six.viewitems(entry):
+            if 'name' not in k:
+                strg_out += '\t{0}\t\t= {1}'.format(k, v) + os.linesep
+        strg_out += os.linesep
+        return six.text_type(strg_out)
+
+    def dict_from_entry(self):
+        ret = OrderedDict()
+        ret[self.criteria['name']] = self.criteria
+        return ret
+
+    def __str__(self):
+        '''
+        String value, only works for full repr
+        '''
+        return self.dict_to_lines(self.criteria)
+
+    def __repr__(self):
+        '''
+        Always works
+        '''
+        return repr(self.criteria)
+
+    def pick(self, keys):
+        '''
+        Returns an instance with just those keys
+        '''
+        subset = dict([(key, self.criteria[key]) for key in keys])
+        return self.__class__(**subset)
+
+    def __init__(self, **criteria):
+        '''
+        Store non-empty, non-null values to use as filter
+        '''
+        items = [key_value for key_value in six.iteritems(criteria) if key_value[1] is not None]
+        items = [(key_value1[0], six.text_type(key_value1[1])) for key_value1 in items]
+        self.criteria = OrderedDict(items)
+
+    @staticmethod
+    def norm_path(path):
+        '''
+        Resolve equivalent paths equivalently
+        '''
+        return os.path.normcase(os.path.normpath(path))
+
+    def match(self, fsys_view):
+        '''
+        Compare potentially partial criteria against built filesystems entry dictionary
+        '''
+        evalue_dict = fsys_view[1]
+        for key, value in six.viewitems(self.criteria):
+            if key in evalue_dict:
+                if evalue_dict[key] != value:
+                    return False
+            else:
+                return False
+        return True
+
+    def __getitem__(self, key):
+        '''
+        Return value for input key
+        '''
+        return self.criteria[key]
 
 
 def fstab(config='/etc/fstab'):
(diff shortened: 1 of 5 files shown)


salt/modules/mount.py
```diff
@@ -5,6 +5,7 @@
 
 # Import python libs
 from __future__ import absolute_import, print_function, unicode_literals
+from collections import OrderedDict
```
cachedout: We generally prefer the use of `salt.utils.odict` instead.

salt/modules/mount.py
```diff
@@ -432,6 +434,141 @@ def match(self, line):
         return True
 
 
+class _filesystems_entry(object):
```
cachedout: Class names are typically camel-case: https://www.python.org/dev/peps/pep-0008/#class-names

salt/modules/mount.py
```diff
@@ -1422,3 +1569,287 @@ def delete_mount_cache(real_name):
...
+
+    config
+        File containing filesystem infomation
+
+    leading_key
+        True    return dictionary keyed by 'name' value and dictionary with other keys, values
+                { '/dir' : { 'dev': '/dev/hd8', .... }}
+
+        False   return dictionary  keyed by 'name' value and dictionary with all keys, values
+                { '/dir' : { 'name' : '/dir', 'dev': '/dev/hd8', .... }}
```
cachedout: What is the expected return of this function?

salt/modules/mount.py
```diff
@@ -1422,3 +1569,287 @@ def delete_mount_cache(real_name):
...
+
+    CLI Example:
+
+    .. code-block:: bash
+
+        salt '*' mount.filesystems
+    '''
+    ret = {}
+    if 'AIX' not in __grains__['kernel']:
+        return ret
```
cachedout: We can't just silently do this. It's going to appear completely broken to nearly all users. It either needs to be implemented for the operating systems supported by this module or not implemented at all.

salt/modules/mount.py
```diff
@@ -1422,3 +1569,287 @@ def delete_mount_cache(real_name):
...
+        mount='true',
+        config='/etc/filesystems',
+        test=False,
+        match_on='auto',
+        **kwargs):
+    '''
+    .. versionadded:: 2018.3.3
+
+    Verify that this mount is represented in the filesystems, change the mount
+    to match the data passed, or add the mount if it is not present on AIX
```
cachedout: I'm not sure how I follow that the functionality of this is different than the `remount` function

salt/utils/files.py
```diff
@@ -407,7 +407,10 @@ def flopen(*args, **kwargs):
     with fopen(*args, **kwargs) as f_handle:
         try:
             if is_fcntl_available(check_sunos=True):
-                fcntl.flock(f_handle.fileno(), fcntl.LOCK_SH)
+                lock_type = fcntl.LOCK_SH
+                if salt.utils.platform.is_aix() and ('a' in args[1] or 'w' in args[1]):
```
cachedout: I am not a huge fan of putting this lookup overhead into 100% of read calls in Salt. Can't we make this an argument instead?

cachedout (changes requested):

salt/modules/mount.py
terminalmage: Yeah, and they don't need to be private. It's a common misconception that it'll be picked up by Sphinx autodoc but it only picks up functions, not classes.

salt/utils/files.py
terminalmage: Yeah, we use this function everywhere, that would impact performance. It's better if we introduce a `lock_type` argument with a default value of `fcntl.LOCK_SH` and then rely on AIX-specific code to pass `lock_type=LOCK_EX` when opening files.

salt/modules/mount.py
```diff
@@ -432,6 +434,141 @@ def match(self, line):
...
+    formatting, and comparing lines. Parsing emits dicts expected from
+    fstab() or raises a ValueError.
+
+    Note: We'll probably want to use os.normpath and os.normcase on 'name'
+    '''
+
+    class ParseError(ValueError):
+        '''
+        Error raised when a line isn't parsible as an fstab entry
+        '''
```
terminalmage: I'm not sure why a separate exception class is needed here, specifically one that is private to this class.

salt/modules/mount.py
```diff
@@ -1422,3 +1569,287 @@ def delete_mount_cache(real_name):
                 if not cache_write:
                     raise CommandExecutionError('Unable to write mount cache.')
     return True
+
+
+def _filesystems(config='/etc/filesystems', leading_key=True):
+    '''
+    .. versionadded:: 2018.3.3
```
terminalmage: Sphinx directives aren't really useful in a private function.

salt/modules/mount.py
terminalmage: I think that this is meant as an AIX complement to the `fstab` function used on other platforms. It should be explicitly documented as such to distinguish itself.

I don't understand the reasoning for putting all the work in a separate, private function, though.

salt/modules/mount.py
```diff
@@ -1422,3 +1569,287 @@ def delete_mount_cache(real_name):
...
+    entry_ip = _filesystems_entry.from_line(entry_args, kwargs)
+    try:
+        criteria = entry_ip.pick(match_on)
+
+    except KeyError:
+        filterFn = lambda key: key not in _filesystems_entry.compatibility_keys
+        invalid_keys = filter(filterFn, match_on)
+
+        msg = 'Unrecognized keys in match_on: "{0}"'.format(invalid_keys)
+        raise CommandExecutionError(msg)
```
terminalmage: We don't need to assign this to a variable if all we're doing is raising an exception with it.

salt/modules/mount.py
dmurphy18: Will do, hang over from developing it.

salt/modules/mount.py
```diff
@@ -1422,3 +1569,287 @@ def delete_mount_cache(real_name):
...
+                else:
+                    ret = 'change'
+                    kv = entry_ip['name']
+                    view_lines.append((kv, entry_ip))
+            else:
+                view_lines.append(fsys_view)
+
+    except (IOError, OSError) as exc:
+        msg = 'Couldn\'t read from {0}: {1}'
+        raise CommandExecutionError(msg.format(config, exc))
```
terminalmage: We don't need to assign this to a variable if all we're doing is raising an exception with it.

salt/modules/mount.py
```diff
@@ -1422,3 +1569,287 @@ def delete_mount_cache(real_name):
...
+        try:
+            with salt.utils.files.fopen(config, 'wb') as ofile:
+                # The line was changed, commit it!
+                for fsys_view in view_lines:
+                    entry = fsys_view[1]
+                    mystrg = _filesystems_entry.dict_to_lines(entry)
+                    ofile.writelines(salt.utils.data.encode(mystrg))
+        except (IOError, OSError):
+            msg = 'File not writable {0}'
+            raise CommandExecutionError(msg.format(config))
```
terminalmage: We don't need to assign this to a variable if all we're doing is raising an exception with it.

salt/modules/mount.py
```diff
@@ -1422,3 +1569,287 @@ def delete_mount_cache(real_name):
...
+                    modified = True
+                else:
+                    view_lines.append(fsys_view)
+
+            except _filesystems_entry.ParseError:
+                view_lines.append(fsys_view)
+
+    except (IOError, OSError) as exc:
+        msg = "Couldn't read from {0}: {1}"
+        raise CommandExecutionError(msg.format(config, exc))
```
terminalmage: We don't need to assign this to a variable if all we're doing is raising an exception with it.

salt/modules/mount.py
```diff
@@ -1422,3 +1569,287 @@ def delete_mount_cache(real_name):
...
+    if modified:
+        try:
+            with salt.utils.files.fopen(config, 'wb') as ofile:
+                for fsys_view in view_lines:
+                    entry = fsys_view[1]
+                    mystrg = _filesystems_entry.dict_to_lines(entry)
+                    ofile.writelines(salt.utils.data.encode(mystrg))
+        except (IOError, OSError) as exc:
+            msg = "Couldn't write to {0}: {1}"
+            raise CommandExecutionError(msg.format(config, exc))
```
terminalmage: We don't need to assign this to a variable if all we're doing is raising an exception with it.

terminalmage (changes requested):

salt/modules/mount.py
dmurphy18: @terminalmage reason for separate functions for filesystems, set_filesystems, rm_filesystems is following established pattern, for example: automaster, set_automaster, rm_automaster and similar for vfstab, even if it basically call fstab equivalent with a different config file.  In this case more so, since /etc/filesystems (multi-line entries), is not even close to /etc/fstab in form (single line entries).

@cachedout the quick exit is similar to that already implemented, for example def automaster checks for a file (/etc/auto_salt) which will probably not exist on Linux and hence quickly exit. I could do the same check for file, however I feel the OS check is more inherently accurate as to functionality exists or not.  That said, I can change it to a file test similar to automaster.
Let me know your preference.

salt/modules/mount.py
dmurphy18: @terminalmage @cachedout I was following the current code's implementation, _fstab_entry, _vfstab_entry. I can certainly change the name, however it would appear inconsistent with existing code.  I prefer to maintain the style and usage already present in the file, but if preferred I can update my additions to be conformant with existing standards.  
Let me know which is the preference moving forward so that I can be consistent in the future.

salt/modules/mount.py
dmurphy18: OrderedDict, have to maintain the order of the entries found in the file /etc/filesystems as it is significant for correct operation. I shall update the doc string to better reflect this.

salt/modules/mount.py
dmurphy18: @cachedout so what is the reason for set_automaster, set_fstab, set_vfstab, just being consistent with existing functionality which is used internally in states/mount.py, where set_fstab is called.  Need to have AIX equivalents hence set_filesystems.

salt/modules/mount.py
dmurphy18: @terminalmage but automaster was written as a complement to fstab for Darwin.  As for msg, adjusted but similar usage throughout file

salt/modules/mount.py
dmurphy18: Due to cut & paste, as similar usage throughout existing file.

salt/modules/mount.py
dmurphy18: Due to cut & paste, as similar usage throughout existing file.

salt/modules/mount.py
dmurphy18: Due to cut & paste, as similar usage throughout existing file.

salt/modules/mount.py
dmurphy18: @terminalmage As a general review question, do we keep new code similar to existing code in existing file, or write in new style ?

salt/utils/files.py
dmurphy18: referring to https://github.com/saltstack/zh/issues/1409, 
Will attempt to leverage something at import time so that I can use that, but the issue being that Exclusive locks is only required if file opened for write, and I would not want to hurt performance for reads.  But if worried about performance, why is that is_fcntl_available constantly being checked when only really applies if Solaris.  

Could always rewrite such that AIX don't bother with is_fcntl_available and resultant minimized performance hit. Let me know your thoughts

salt/modules/mount.py
dmurphy18: done

salt/modules/mount.py
dmurphy18: done

salt/modules/mount.py
dmurphy18: done

salt/modules/mount.py
dmurphy18: done

salt/modules/mount.py
dmurphy18: done

salt/modules/mount.py
terminalmage: No that's not what I was talking about. I was talking about the fact that all the actual work was in a `_filesystems` function.

salt/modules/mount.py
dmurphy18: Updated to CamelCase

salt/modules/mount.py
dmurphy18: @terminalmage the use of _filesystems function is only to read in the contents /etc/filesystems, in order, and then allow manipulation as desired, here to write the new state to /etc/filesystems in the correct order.  Order is significant on AIX and should not be changed, otherwise operations are affected.
Hence the use of _filesystems is to obtain the current /etc/filesystems into an instance where it can be manipulated.  The function is used in plain output, setting and removing. A common internal function used, perhaps you could explain further your objection to its use.

Partial contents of /etc/filesystems for example:

/opt:
        dev             = /dev/hd10opt
        vfs             = jfs2
        log             = /dev/hd8
        mount           = true
        check           = true
        vol             = /opt
        free            = false
        quota           = no

/var/adm/ras/livedump:
        dev             = /dev/livedump
        vfs             = jfs2
        log             = /dev/hd8
        mount           = true
        account         = false
        quota           = no

dmurphy18:
Updated for review comments.  Note the changes to salt/utils/files.py flock and use of LOCK_EX.
We should be opening files for write with Exclusive Locks, rather than Shared. Default on Linux OS is LOCK_EX, on AIX OS it will object if attempt to open a file for write and not use LOCK_EX.

The use of LOCK_SH has been present in Salt since 2013, so this change may cause unanticipated affects. Please review this section carefully.

rallytime:
@dmurphy18 This is causing a test to fail: https://jenkinsci.saltstack.com/job/pr-kitchen-centos7-py2/job/PR-48803/3/testReport/junit/unit.modules.test_mount/MountTestCase/test_set_filesystems/

cachedout:
Once the comments from @terminalmage are resolved, I'm OK with this. The test still needs to be fixed though.

cachedout (approved):

salt/modules/mount.py
terminalmage: I did not catch the other calls to the function in my earlier review.

terminalmage (commented):
LGTM pending tests being resolved.

terminalmage (approved):
LGTM pending tests being resolved.

dmurphy18:
@rallytime Not see test_mount failures with:
python runtests.py -m
python runtests.py -m unit.modules.test_mount

On both Centos 7 7.3 and update to latest 7.5.1804

Do see 2 failures on both versions of Centos 7
integration.modules.test_cmdmod.CMDModuleTest.test_cmd_run_whoami
integration.modules.test_linux_acl.LinuxAclModuleTest.test_getfacl_w_single_file_without_acl

Running with VirtualBox rather than Docker and Kitchen etc.  Believe issues seen are related to environment than unit tests.

rallytime:
@dmurphy18 I can reproduce these with just `runtests.py -n unit.modules.test_mount`. I ran this using a docker container from [barnacle](https://github.com/cachedout/barnacle).

If you still can't reproduce, you should be able to get a matching environment with kitchen-salt. @gtmanfred should be able to point you in the right direction.

dmurphy18:
@rallytime I am very suspect that we are looking at a docker issue of some sort given VM's are a better abstraction of a machine.  However understand the need to get it working with testing and will check with Daniel to resolve the issue with automatic testing, noting that both Debian 8 and Redhat 7 (7.3 and 7.5) pass fine on a VM.

dmurphy18:
Problem with tests were due to Redhat/Arch having a /etc/filesystems file, Debian/Ubuntu do not.
Adjusted tests and fixed Python 3 issue

PR #48804, opened 2018-07-27
Use brew path from which cmd in mac_brew module

Ch3LL:
### What does this PR do?
see issue https://github.com/saltstack/salt-jenkins/issues/1033 for details. 

after this fix https://github.com/saltstack/salt/pull/47212  we aren't passing the PATH via the env anymore so we need to run `which brew` to get the entire path or else we see this error:

```
salt-call --local pkg.install vim
[ERROR   ] stderr: -bash: brew: command not found
[ERROR   ] retcode: 127
Error running 'pkg.install': Brew command failed. Additional info follows:

result:
    ----------
    pid:
        55236
    retcode:
        127
    stderr:
        -bash: brew: command not found
    stdout:
```

also note: brew tests still will not pass until https://github.com/saltstack/salt/pull/48796 is merged

fyi @weswhet

### DIFF
--- a/salt/modules/mac_brew.py
+++ b/salt/modules/mac_brew.py
@@ -48,7 +48,7 @@ def _list_taps():
     '''
     List currently installed brew taps
     '''
-    cmd = 'brew tap'
+    cmd = 'tap'
     return _call_brew(cmd)['stdout'].splitlines()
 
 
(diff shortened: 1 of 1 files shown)


weswhet:
LGTM thanks!

rallytime (approved):

PR #48812, opened 2018-07-28
Use Adler32 algorithm to compute string checksums

isbm:
### What does this PR do?

An alternative to https://github.com/saltstack/salt/pull/46649

### Previous Behavior

`server_id` is changing due to Python 3 is salting `hash` function. A workaround is to shell-out Python with environment variable `PYTHONHASHSEED=0`. Unfortunately, this is the only option.

However, what if we use Adler-32 instead of hash? This would avoid shell-out Python.
@terminalmage what do you think?

### Tests written?

Should apply existing

### DIFF
--- a/salt/config/__init__.py
+++ b/salt/config/__init__.py
@@ -1200,6 +1200,10 @@ VALID_OPTS = {
 
     # Thorium top file location
     'thorium_top': six.string_types,
+
+    # Use Adler32 hashing algorithm for server_id (default False until Sodium, "adler32" after)
+    # Possible values are: False, adler32, crc32
+    'server_id_use_crc': (bool, six.string_types),
 }
 
 # default configurations
--- a/salt/config/__init__.py
+++ b/salt/config/__init__.py
@@ -1508,7 +1508,8 @@ DEFAULT_MINION_OPTS = {
     },
     'discovery': False,
     'schedule': {},
-    'ssh_merge_pillar': True
+    'ssh_merge_pillar': True,
+    'server_id_use_crc': False,
 }
 
 DEFAULT_MASTER_OPTS = {
--- a/salt/grains/core.py
+++ b/salt/grains/core.py
@@ -20,6 +20,7 @@ import platform
 import logging
 import locale
 import uuid
+import zlib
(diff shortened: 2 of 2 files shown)


terminalmage (approved):
I am open to this change, but if we are going to make it then I think we need to at minimum be very explicit about the change in the release notes, due to the fact that this will change a grain on which users may be relying.

@thatch45 @cachedout @rallytime thoughts? Should we gate this behind a config item and make it the new default method after a couple release cycles? Or, should we make it the default right away?

cachedout:
I really don't like doing this sort of thing. We're going to change all of these hashes out from under people but for no reason other than it's only just a fraction of a bit faster and the code is slightly cleaner and I just don't feel like that's worth the disruption to users.

rallytime:
I agree with @cachedout here. I don't think this is something that we want to disrupt upgrades for with users, IMO.

cachedout:
We didn't get a response from @isbm so I'm closing this for now. Bo, if you like we can discuss this more on our next call. :)

isbm:
@cachedout @rallytime  I was on the vacations in the woods off the grid, and now this PR is just closed. :cry: 

The part that changing hashes probably isn't _that much_ problematic, but I've got the point. So why we indeed do not gate this behind the option and retire then it later, just like @terminalmage said? In the meanwhile firing up another interpreter for _just_ hash is really what has to be done due to impossibility of doing better. Which is fine (-ish, because on the Mainframes with the VM of Ferrari-price RAM), but still how we can go away from this hack in the future clean?

Another problem with the hack is that Adler32 is fixed, finished algorithm. It won't change. But are we sure Python next versions won't turn off that environment variable at all and we have them this problem back? I would still look for the option to gating this behind the option & press release this.

rallytime:
I can re-open this for now, certainly. We can discuss this tomorrow in our meeting. :)

isbm:
As per conversation, a valid concern from @cachedout for not to update defaults abruptly has been addressed as per suggestion from @terminalmage .

@rallytime what are the next steps here?

cachedout (approved):

PR #43288, opened 2017-08-31
allow pillar to be used in vault policies

gtmanfred:
### What does this PR do?
Allows the more secure pillar object to be used in vault policies.

### What issues does this PR fix or reference?
Closes 
#43287 

### Tests written?

No

PR #48829, opened 2018-07-30
Vault config pillar template

sizgiyaev:
### What does this PR do?
Adds vault module config policy templating by pillar.

### Previous Behavior
Now it's only possible with grains.

### Tests written?

Yes

### Commits signed with GPG?

No


mchugh19:
This likely has the same issue as was brought up in #43288

PR #48837, opened 2018-07-31
Check file_recv_max_size early

teselkin:
### What does this PR do?
Check size of the file transferred by cp.push early, preventing transferring files that exceeds file_recv_max_size

None

### Previous Behavior
Current behaviour of cp.push is that salt master checks file_recv_max_size
during file transfer process. Given that by default this limit is
about 100MB if we try to send 101MB then 100MB would be sent anyway,
and only after that master interrupts the transfer. Such useless
transfers should be avoided.

### New Behavior
This commit adds additional field 'size' to 'load' structure that
minion sends to master, transfrerring file size in every message,
so that master can check the size recieving just the first chunk
of data, saving the queue bandwith.

### Tests written?

No

### Commits signed with GPG?

No

### DIFF
--- a/salt/daemons/masterapi.py
+++ b/salt/daemons/masterapi.py
@@ -661,6 +661,13 @@ class RemoteFuncs(object):
             log.error('Invalid file pointer: load[loc] < 0')
             return False
 
+        if 'size' in load and load['size'] > file_recv_max_size:
+            log.error(
+                'Exceeding file_recv_max_size limit: %s',
+                file_recv_max_size
+            )
+            return False
+
         if len(load['data']) + load.get('loc', 0) > file_recv_max_size:
             log.error(
                 'Exceeding file_recv_max_size limit: %s',
--- a/salt/daemons/masterapi.py
+++ b/salt/daemons/masterapi.py
@@ -661,7 +661,7 @@ class RemoteFuncs(object):
             log.error('Invalid file pointer: load[loc] < 0')
             return False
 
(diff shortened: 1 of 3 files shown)


salt/daemons/masterapi.py
```diff
@@ -661,6 +661,13 @@ def _file_recv(self, load):
             log.error('Invalid file pointer: load[loc] < 0')
             return False
 
+        if 'size' in load and load['size'] > file_recv_max_size:
```
rares-pop: this is an anti-pattern - doing 'foo' in dict and dict['foo'], as described here:
https://docs.quantifiedcode.com/python-anti-patterns/correctness/not_using_get_to_return_a_default_value_from_a_dictionary.html

You should be using dict.get(), as in: if load.get('size', 0) > file_recv_max_size:

teselkin: Thanks, fixed.

gtmanfred (changes requested):
Looks like this breaks the test.push test case.

https://jenkinsci.saltstack.com/job/pr-kitchen-ubuntu1604-py2/job/PR-48837/2/testReport/junit/unit.modules.test_cp/CpTestCase/test_push/

gtmanfred:
looks like you got the unittest updated just before I could comment 😄

terminalmage (approved):
Looks good pending test results.

rallytime (approved):

PR #48842, opened 2018-07-31
modules/nilrt_ip.py: Fix _get_adapter_mode

alexvasiu:
The value returned by _load_config may have upper letters.


alexvasiu:
@gtmanfred May you take a look? It should be integrated into `Fluorine`.

gtmanfred (approved):

PR #48852, opened 2018-07-31
Record all the artifacts from the build

KaiSforza:
### What does this PR do?

Adds more of the build artifacts to the PR runs

None.

### Commits signed with GPG?

Yes

@gtmanfred

### DIFF
--- a/.ci/kitchen-centos7-py2
+++ b/.ci/kitchen-centos7-py2
@@ -46,6 +46,8 @@ pipeline {
                         }
                     }}
                     archiveArtifacts artifacts: 'artifacts/xml-unittests-output/*.xml'
+                    archiveArtifacts artifacts: 'artifacts/logs/minion'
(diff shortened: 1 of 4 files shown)


.ci/kitchen-centos7-py2
```diff
@@ -46,6 +46,8 @@ pipeline {
                         }
                     }}
                     archiveArtifacts artifacts: 'artifacts/xml-unittests-output/*.xml'
+                    archiveArtifacts artifacts: 'artifacts/logs/minion'
+                    archiveArtifacts artifacts: 'artifacts/coverage/coverage.xml
```
gtmanfred: coverage.xml should be pushed to the coverage plugin, we need `salt-runtests.log.

gtmanfred (changes requested):

gtmanfred (approved):

PR #48863, opened 2018-08-01
Decode file contents for python2 as well

admd:
### What does this PR do?
In file.blockreplace function, when user specify a source file which has some non-ascii characters. Then similar error as following was appearing
``` UnicodeDecodeError: 'ascii' codec can't decode byte 0xc2 in > position 17: ordinal not in range(128) ```

This happens with python2 only. There was already check for python3, not sure why it was limited to python3 though. With this PR, decoding will be performed no matter what version.

### Tests written?

No

### Commits signed with GPG?

Yes


terminalmage (approved):
@rallytime this needs a backport to 2018.3.

rallytime:
Thanks @admd and welcome!

Issue #44997, opened 2017-12-14
Add Windows cmd.run Test to Auto Test Suite


Ch3LL:
This task is to monitor the addition of one test to the auto test suite:

- [ ] cmd.run

`salt 'windows' cmd.run 'type c:\salt\conf\minion > nul'`
`salt 'windows' cmd.run 'type c:\salt\conf\minion' python_shell=False`

Just need to get integration/modules/test_cmdmod.py tests to run on windows

PR #48866, opened 2018-08-01
Add cmd module integration tests for windows and fix space in path issue

Ch3LL:
### What does this PR do?
Adds the cmdmod module integration tests to run on windows and fixes an issue on windows when there is a space in the path name of the `cwd` argument.

Given this state file:

```
test:
  cmd.script:
    - source: salt://test.py
    - cwd: "c:\\test 2"
    - args: 'saltine crackers biscuits=yes'
```

When run I see this error:

```
----------
          ID: test
    Function: cmd.script
      Result: False
     Comment: Command 'test' run
     Started: 12:11:28.997000
    Duration: 454.0 ms
     Changes:
              ----------
              pid:
                  38548
              retcode:
                  1
              stderr:
                  '"c:\test' is not recognized as an internal or external command,
                  operable program or batch file.
              stdout:

Summary for local
------------
Succeeded: 0 (changed=1)
Failed:    1
------------
Total states run:     1
Total run time: 454.000 ms
PS C:\testing>
```

If i use the powershell shell it works:

```
test:
  cmd.script:
    - source: salt://test.ps1
    - cwd: "c:\\test 2"
    - args: 'saltine crackers biscuits=yes'
    - shell: powershell
```

```
local:
----------
          ID: test
    Function: cmd.script
      Result: True
     Comment: Command 'test' run
     Started: 12:14:27.200000
    Duration: 954.0 ms
     Changes:
              ----------
              pid:
                  39572
              retcode:
                  0
              stderr:
              stdout:
                  hello

Summary for local
------------
Succeeded: 1 (changed=1)
Failed:    0
------------
```

The reason this is occurring is because it tries to escape the quotes around the command leaving hte space unescaped. So the resulting command is this:

`chcp 437 > nul & ^"c:\test 2\__salt.tmp.al96jv.py^"` and if run in cmd.exe:

```
C:\Users\Administrator>chcp 437 > nul & ^"c:\test 2\__salt.tmp.al96jv.py^"
'"c:\test' is not recognized as an internal or external command,
operable program or batch file.
```

Also fixes an issue when it it attempts to clean the temp file here: https://github.com/saltstack/salt/blob/v2017.7.7/salt/modules/cmdmod.py#L2141

it cannot delete the temp file because it quotes the path `'"<path>"'` and whne it checks to see if its a valid path it fails. So this ensures we are using the inital path created from here: https://github.com/saltstack/salt/blob/v2017.7.7/salt/modules/cmdmod.py#L2084

Also this PR changes the expected return in windows on the tests so I could use some review if those returns are expected behavior or not.

https://github.com/saltstack/salt/issues/44997

### Tests written?

Yes - adds tests and fixes something a test found

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/utils/win_functions.py
+++ b/salt/utils/win_functions.py
@@ -166,7 +166,7 @@ def get_sam_name(username):
     return '\\'.join([domain, username])
 
 
-def escape_argument(arg):
+def escape_argument(arg, escape=True):
     '''
     Escape the argument for the cmd.exe shell.
     See http://blogs.msdn.com/b/twistylittlepassagesallalike/archive/2011/04/23/everyone-quotes-arguments-the-wrong-way.aspx
@@ -183,6 +183,8 @@ def escape_argument(arg):
     if not arg or re.search(r'(["\s])', arg):
         arg = '"' + arg.replace('"', r'\"') + '"'
 
+    if not escape:
+        return arg
     return escape_for_cmd_exe(arg)
 
 
--- a/salt/utils/win_functions.py
+++ b/salt/utils/win_functions.py
@@ -177,6 +177,11 @@ def escape_argument(arg, escape=True):
     Args:
         arg (str): a single command line argument to escape for the cmd.exe shell
 
+    Kwargs:
+        escape (bool): True will call the escape_for_cmd_exe() function
+                       which escapes the characters '()%!^"<>&|'. False
+                       will not call the function and only quotes the cmd
+
     Returns:
         str: an escaped string suitable to be passed as a program argument to the cmd.exe shell
     '''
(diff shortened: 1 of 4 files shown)


salt/utils/win_functions.py
```diff
@@ -166,7 +166,7 @@ def get_sam_name(username):
     return '\\'.join([domain, username])
 
 
-def escape_argument(arg):
+def escape_argument(arg, escape=True):
     '''
```
twangboy: Please add this new parameter to the docs for this function

twangboy (changes requested):

damon-atkins:
Ping @ciiqr

damon-atkins:
Why do we need `chcp 437`?  What is the background with this?

twangboy (approved):

dwoz (approved):

PR #39754, opened 2017-03-01
terraform roster

dmacvicar:
### What does this PR do?

Implements a dynamic roster from terraform current state file.

Currently supports libvirt and AWS resources.
(For libvirt, see https://github.com/dmacvicar/terraform-provider-libvirt)

For libvirt resources, currently the first address of the first interface is chosen.
For AWS resources, the public_dns or the public_ip attributes are used.

Example:

Given a simple salt-ssh tree with a Saltfile

```yaml
salt-ssh:
  config_dir: etc/salt
  max_procs: 30
  wipe_ssh: True
```

and etc/salt/master

```yaml
root_dir: .
  file_roots:
    base:
      - srv/salt
  pillar_roots:
    base:
      - srv/pillar
  roster: terraform
```

We setup a terraform file that creates N virtual machines (libvirt_domain resources) setup
with cloud-init and the salt-ssh ssh key. In this case we set N to 2.

```hcl
provider "libvirt" {
  uri = "qemu:///system"
}

resource "libvirt_cloudinit" "init" {
  name = "test-init.iso"
  ssh_authorized_key = "${file("etc/salt/pki/master/ssh/salt-ssh.rsa.pub")}"
}

resource "libvirt_volume" "opensuse_leap" {
  name = "opensuse_leap"
  source = "http://r.opensu.se/Cloud:Images:Leap_42.3/images/openSUSE-Leap-42.3-OpenStack.x86_64.qcow2"
}

resource "libvirt_volume" "volume" {
  name = "volume-${count.index}"
  base_volume_id = "${libvirt_volume.opensuse_leap.id}"
  count = 2
}

resource "libvirt_domain" "domain" {
  name = "domain-${count.index}"
  memory = 1024
  disk {
       volume_id = "${element(libvirt_volume.volume.*.id, count.index)}"
  }

  network_interface {
    network_name = "default"
    hostname = "minion${count.index}"
    wait_for_lease = 1
  }
  cloudinit = "${libvirt_cloudinit.init.id}"
  count = 2
}
```

```console
$ terraform apply
libvirt_cloudinit.init: Creating...
...
libvirt_volume.opensuse_leap: Creating...
...
libvirt_domain.domain.1: Creating...
libvirt_domain.domain.0: Creating...
...
libvirt_domain.domain.0: Creation complete
...
libvirt_domain.domain.1: Creation complete

Apply complete! Resources: 5 added, 0 changed, 0 destroyed.

The state of your infrastructure has been saved to the path
[...]

PR #48873, opened 2018-08-01
[RFC] Terraform roster module to be used with terraform-provider-salt

dmacvicar:
- Feature Name: Roster module for Terraform to be used with salt-ssh
- Start Date: 2018-07-27
- RFC PR: 
- Salt Issue: 

# Summary
[summary]: #summary

This is a proposal for a roster module to be used with terraform. It solves the use-case of a self-contained folder in git with infrastructure defined with terraform, where one wants
to provision the infrastructure using salt-ssh. The idea is that the salt-ssh roster is automatically filled with terraform resources.

```console
$ cd myproject
$ terraform apply
$ salt-ssh '*' state.apply
```

# Motivation
[motivation]: #motivation

The goal is very simple: self-contained directories with infrastructure created by terraform and configuration using salt-ssh. [Example here](https://github.com/dmacvicar/terraform-provider-salt/tree/master/examples/libvirt-leap).

Some time ago I proposed a [solution for this](https://github.com/saltstack/salt/pull/39754), which never got merged (for different reasons).
While I did not know about it at the time, the design was very similar to [terraform-inventory](https://github.com/adammck/terraform-inventory) module for Ansible, and it suffers from the same problems.

It worked by lookin at terraform state file, and recognizing certain types of resources, like libvirt hosts and AWS instances, then doing things like looking at the public_ip and creating a roster based on that.

However, when I got feedback, the approach drawbacks were made evident:

* "Can you use the second ip instead of the first ip? what about the private one?
* "Azure is not supported!"

These drawbacks affected the motivation to get it merged. However it let me thinking on the problem, until I found [terraform-ansible-provider](https://github.com/nbering/terraform-provider-ansible/) and read the inspirational post about it.

I believe this is a much better approach for a terraform/salt-ssh integration roster module.

# Design
[design]: #detailed-design

The design, when compared to the original proposal, is to make the salt side (terraform.py) very agnostic. It does not know about all terraform resource types like AWS or Azure, but only one type of resource: `salt_host`.

the `salt_host` resource is provided by another provider: [terraform-provider-salt](https://github.com/dmacvicar/terraform-provider-salt). It is just a logical resource. It does not do much than gluing terraform resources to the roster:

So for a libvirt resource that is instantiated two times:

```hcl
resource "libvirt_domain" "domain" {
  name = "domain-${count.index}"
  memory = 1024
  disk {
    volume_id = "${element(libvirt_volume.volume.*.id, count.index)}"
  }
  
  network_interface {
    network_name = "default"
    wait_for_lease = true
  }
  count = 2
}
```

One could create one `salt_host` resource per libvirt domain, and bind the roster host to the first ip address:

```hcl
resource "salt_host" "example" {
  salt_id = "minion${count.index}"
  host = "${element(libvirt_domain.domain.*.network_interface.0.addresses[count.index], 0)}"
  user = "root"
  passwd = "linux"
  count = 2
}
```

As you can see, there are no limits on what you can glue into the roster. You have absolute freedom on how many hosts to expose to Salt, and what address to use in there.

This means `terraform.py` is much simpler. It scans the `salt_host` resources from the state, and collects the properties, which are 1:1 mapped to the roster attributes.
The roster module would not need to be updated as terraform adds features and resource types.

## Alternatives
[alternatives]: #alternatives

* A roster module that does not require a terraform provider. It was described in the motivation and my previous [PR](https://github.com/saltstack/salt/pull/39754). With the drawbacks described above: need to know about specific terraform resources.

* Another idea/alternative is to keep the terraform provider side but alter the `terraform.py` included in this script and make it a more generic "script" roster module. So that if the module finds a `salt-ssh.roster` executable, or it is mentioned in the Saltfile, it gets the roster from executing that script. In this case, the terraform-go-provider could implement the script and provide itself the roster (however this may require some knowledge of how to abuse a terraform provider executable to perform another task when called with a different option, and to guess where it is, as providers are not in PATH but in terraform.plugins.d).
In any case this is an improvement over this RFC, and could be done in a second iteration/step.
*Update*: this could be more feasible than I thought: https://twitter.com/dmacvicar/status/1024800245325213697 opinions?

## Unresolved questions
[unresolved]: #unresolved-questions

* Where to host `terraform-provider-salt`.

# Drawbacks
[drawbacks]: #drawbacks

The disadvantage is that the roster module needs and depends on a specific salt provider on the terraform side. But this provider ([prototype](https://github.com/dmacvicar/terraform-provider-salt)) is very simple, as it is pure logical, and will not need much changes if the roster architecture stays the same.


gtmanfred (changes requested):
This looks good, just one thing that I would like to see fixed, and another that is a nitpick.

dmacvicar:
Before moving onto fixing the issues in the roster code, I would focus on the RFC.

I would like some feedback on the 2nd alternative. Because if we instead of merging the roster module add a more generic script roster module, then for terraform one would only need the terraform salt provider, and the roster generic script module can be reused for other integrations.

In summary, instead of `terraform.py` and master having `roster: terraform`, having something like `roster_script: terraform-provider-salt --roster` and move all this functionality to the provider. Which would act as a terraform provider plugin when called by terraform and as a roster provider when called by Salt.

isbm:
@terminalmage @cachedout @rallytime any thoughs here?

isbm:
@dmacvicar whoa, I find this very nice idea (the main one) and the right step towards resources [re]use!

While the idea as it is _now_ is already great and can be left as is, but looking at this from mere user perspective, I find they likely would be even more happy to just call `salt-ssh \* state.apply` alone. As I would imagine, a particular Salt state could be/do the following:

- Describe resources in the SLS state something like Salt Cloud already doing, which is similar to how Terraform is describing resource. Maybe we could find one generic syntax for SLS on this regard and make sure it is applicable everywhere?
- Render terraform cf from that state structure (extend `terraform.py` with this ability, perhaps?) so no one needs to describe things in two different places. Also we could separate resource description from state processing, and so Salt Cloud part would leverage from this a lot (and vice versa).
- Plan/apply terraform to setup everything
- Apply the final state on top of it to end the infrastructure setup

Maybe we should also teach Salt to read Terraform state to let Salt know where it is.

dmacvicar:
>Describe resources in the SLS state something like Salt Cloud already doing, which is similar to how Terraform is describing resource. Maybe we could find one generic syntax for SLS on this regard and make sure it is applicable everywhere?

I think this is true for Salt users. I doubt Terraform users want to change the way they describe resources and that is completely out of the scope of this PR. It belongs to `salt-cloud` future.

In any case, it is not a simple problem. Terraform graph and Salt states/sls are different. Terraform automatically figures dependencies between resources as you use interpolation between them, unlike Salt `require:`.

For sake of moving the discussion I beg to keep this topic out. We don't want to saltify terraform. We want to let terraform users use Salt. Everything else is an effort that can be done independently of this PR and should be tackled by someone having a) deep knowledge about limitations of both Terraform and Salt b) some say in the future of `salt-cloud`.

dmacvicar:
Something that may be a nice addition, but unrelated to this PR would be a companion `ext_pillar` so that the `terraform-salt-provider` could interpolate and expose infrastructure details in the pillar.

To give an example: meta-salt. Create a salt cluster with salt. I use terraform to bring up 4 VMs. I use salt-ssh to configure them, 1 as a salt-master and 3 as a salt-minion. I need salt-ssh to know the ip of the VM that was configured as master to change `/etc/salt/minion` `master` key with salt-ssh on the minions. If I have access to the pillar, my "minion" state could use pillar data coming from terraform.

isbm:
> For sake of moving the discussion I beg to keep this topic out. We don't want to saltify terraform. We want to let terraform users use Salt. 

Yes, agreed. As I said, current PR perfectly go in as-is and to me current design makes lots of sense.

salt/roster/terraform.py
```diff
@@ -0,0 +1,171 @@
...
+}
+
+You can use the count attribute to create multiple roster entries with a single definition. Please refer to the `terraform salt provider <https://github.com/dmacvicar/terraform-provider-salt>`_ for more detailed examples.
+'''
+from __future__ import absolute_import
+import logging
+import json
+import os.path
+import salt.utils
+from salt.roster.flat import RosterMatcher
```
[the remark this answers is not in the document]
dmacvicar: I will refactor this in a next PR.

salt/roster/terraform.py
```diff
@@ -0,0 +1,171 @@
...
+      host = "${libvirt_domain.vm-db.network_interface.0.addresses.0}"
+      user = "root"
+      passwd = "linux"
+}
+
+You can use the count attribute to create multiple roster entries with a single definition. Please refer to the `terraform salt provider <https://github.com/dmacvicar/terraform-provider-salt>`_ for more detailed examples.
+'''
+from __future__ import absolute_import, unicode_literals
+import logging
+import salt.utils.json
```
rallytime: Can we break these imports up into 2 sections? We usually do something like:
```
# Import Python libs
from __future__ import absolute_import, unicode_literals
import logging
import os.path

# Import Salt libs
import salt.utils.json
```
etc.

That makes the import list easier to maintain moving forward.

salt/roster/terraform.py
```diff
@@ -0,0 +1,171 @@
...
+import os.path
+import salt.utils
```
rallytime: Rather than importing `salt.utils` can you import the individual files directly? Otherwise you'll see deprecation notices as the big utils init file was broken up in 2018.3.

rallytime (commented):
I have some small requests. :)

salt/roster/terraform.py
dmacvicar: Done

salt/roster/terraform.py
dmacvicar: Done

dmacvicar:
Rebased

salt/roster/terraform.py
```diff
@@ -0,0 +1,174 @@
...
+'''
+# Import Python libs
+from __future__ import absolute_import, unicode_literals
+import logging
+import os.path
+
+# Import Salt libs
+import salt.utils.files
+import salt.utils.json
+from salt.roster.flat import RosterMatcher
```
gtmanfred: Can we move this to `__utils__` and then call it from a `__utils__` function, like `roster.targets()` which calls RosterMatcher, instead of having to pass in `__opts__`?

gtmanfred (changes requested):

gtmanfred:
Overall, i really like this approach, i would just rather see the RosterMatcher stuff moved to __utils__, instead of importing from another roster, and having to pass __opts__ around.

dmacvicar:
@gtmanfred I am trying to get this refactored, but after moving it to utils, if I call it using `__utils__['roster_matcher.targets'] i get `KeyError` as it is not in `__utils__`, and if I import it and call it as `salt.utils.roster_matcher.targets` I get `__opts__` not defined inside it. So I am missing something.

```diff
diff --git a/salt/roster/flat.py b/salt/roster/flat.py
index 9ccb18bc97..8a7fa46d99 100644
--- a/salt/roster/flat.py
+++ b/salt/roster/flat.py
@@ -4,20 +4,6 @@ Read in the roster from a flat file using the renderer system
 '''
 from __future__ import absolute_import, print_function, unicode_literals
 
-# Import python libs
-import fnmatch
-import re
-import copy
-
-# Try to import range from https://github.com/ytoolshed/range
-HAS_RANGE = False
-try:
-    import seco.range
-    HAS_RANGE = True
-except ImportError:
-    pass
-# pylint: enable=import-error
-
 # Import Salt libs
 import salt.loader
 import salt.config
@@ -47,122 +33,4 @@ def targets(tgt, tgt_type='glob', **kwargs):
     conditioned_raw = {}
     for minion in raw:
         conditioned_raw[six.text_type(minion)] = salt.config.apply_sdb(raw[minion])
-    rmatcher = RosterMatcher(conditioned_raw, tgt, tgt_type, 'ipv4', opts=__opts__)
-    return rmatcher.targets()
-
-
-class RosterMatcher(object):
-    '''
-    Matcher for the roster data structure
-    '''
-    def __init__(self, raw, tgt, tgt_type, ipv='ipv4', **kwargs):
-        self.tgt = tgt
-        self.tgt_type = tgt_type
-        self.raw = raw
-        self.ipv = ipv
-        self.opts = kwargs.get('opts', {})
-
-    def targets(self):
-        '''
-        Execute the correct tgt_type routine and return
-        '''
-        try:
-            return getattr(self, 'ret_{0}_minions'.format(self.tgt_type))()
-        except AttributeError:
-            return {}
-
-    def ret_glob_minions(self):
-        '''
-        Return minions that match via glob
-        '''
-        minions = {}
-        for minion in self.raw:
-            if fnmatch.fnmatch(minion, self.tgt):
-                data = self.get_data(minion)
-                if data:
-                    minions[minion] = data.copy()
-        return minions
-
-    def ret_pcre_minions(self):
-        '''
-        Return minions that match via pcre
-        '''
-        minions = {}
-        for minion in self.raw:
-            if re.match(self.tgt, minion):
-                data = self.get_data(minion)
-                if data:
-                    minions[minion] = data.copy()
-        return minions
-
-    def ret_list_minions(self):
-        '''
-        Return minions that match via list
-        '''
-        minions = {}
-        if not isinstance(self.tgt, list):
-            self.tgt = self.tgt.split(',')
-        for minion in self.raw:
-            if minion in self.tgt:
-                data = self.get_data(minion)
-                if data:
-                    minions[minion] = data.copy()
-        return minions
-
-    def ret_nodegroup_minions(self):
-        '''
-        Return minions which match the special list-only groups defined by
-        ssh_list_nodegroups
-        '''
-        minions = {}
-        nodegroup = self.opts.get('ssh_list_nodegroups', {}).get(self.tgt, [])
-        if not isinstance(nodegroup, list):
-            nodegroup = nodegroup.split(',')
-        for minion in self.raw:
-            if minion in nodegroup:
-                data = self.get_data(minion)
-                if data:
-                    minions[minion] = data.copy()
-        return minions
-
-    def ret_range_minions(self):
-        '''
-        Return minions that are returned by a range query
-        '''
-        if HAS_RANGE is False:
-            raise RuntimeError("Python lib 'seco.range' is not available")
-
-        minions = {}
-        range_hosts = _convert_range_to_list(self.tgt, self.opts['range_server'])
-
-        for minion in self.raw:
-            if minion in range_hosts:
-                data = self.get_data(minion)
-                if data:
-                    minions[minion] = data.copy()
-        return minions
-
-    def get_data(self, minion):
-        '''
-        Return the configured ip
-        '''
-        ret = copy.deepcopy(self.opts.get('roster_defaults', {}))
-        if isinstance(self.raw[minion], six.string_types):
-            ret.update({'host': self.raw[minion]})
-            return ret
-        elif isinstance(self.raw[minion], dict):
-            ret.update(self.raw[minion])
-            return ret
-        return False
-
-
-def _convert_range_to_list(tgt, range_server):
-    '''
-    convert a seco.range range into a list target
-    '''
-    r = seco.range.Range(range_server)
-    try:
-        return r.expand(tgt)
-    except seco.range.RangeException as err:
-        log.error('Range server exception: %s', err)
-        return []
+    return __utils__['roster_matcher.targets'](conditioned_raw, tgt, tgt_type, 'ipv4')
diff --git a/salt/roster/terraform.py b/salt/roster/terraform.py
index d57f47f145..ebfae02cbc 100644
--- a/salt/roster/terraform.py
+++ b/salt/roster/terraform.py
@@ -52,7 +52,7 @@ import os.path
 # Import Salt libs
 import salt.utils.files
 import salt.utils.json
-from salt.roster.flat import RosterMatcher
+import salt.utils.roster_matcher
 
 log = logging.getLogger(__name__)
 
@@ -170,5 +170,4 @@ def targets(tgt, tgt_type='glob', **kwargs):  # pylint: disable=W0613
 
     raw = _parse_state_file(roster_file)
     log.debug('%s hosts in terraform state file', len(raw))
-    rmatcher = RosterMatcher(raw, tgt, tgt_type, 'ipv4', opts=__opts__)
-    return rmatcher.targets()
+    return salt.utils.roster_matcher.targets(raw, tgt, tgt_type, 'ipv4')
```

gtmanfred:
if you remove `salt.utils.roster_matcher` from the front of calling RosterTargets, does it fix the problem?

```
def targets(conditioned_raw, tgt, tgt_type, ipv='ipv4'):
    rmatcher = RosterMatcher(conditioned_raw, tgt, tgt_type, ipv)
    return rmatcher.targets()
```

gtmanfred:
Installing to check this.

gtmanfred:
With your patch, i get this
```
[root@salt salt]# salt-ssh \* test.ping
AttributeError: 'module' object has no attribute 'roster_matcher'
  File "/bin/salt-ssh", line 8, in <module>
    execfile(__file__)
  File "/root/src/salt/scripts/salt-ssh", line 10, in <module>
    salt_ssh()
  File "/root/src/salt/salt/scripts.py", line 425, in salt_ssh
    client.run()
  File "/root/src/salt/salt/cli/ssh.py", line 23, in run
  File "/root/src/salt/salt/client/ssh/__init__.py", line 231, in __init__
    self.tgt_type)
  File "/root/src/salt/salt/roster/__init__.py", line 105, in targets
  File "/root/src/salt/salt/roster/flat.py", line 36, in targets
    return __utils__['roster_matcher.targets'](conditioned_raw, tgt, tgt_type, 'ipv4')
  File "/root/src/salt/salt/utils/roster_matcher.py", line 32, in targets
    rmatcher = salt.utils.roster_matcher.RosterMatcher(conditioned_raw, tgt,
AttributeError: 'module' object has no attribute 'roster_matcher'
  File "/bin/salt-ssh", line 8, in <module>
    execfile(__file__)
  File "/root/src/salt/scripts/salt-ssh", line 10, in <module>
    salt_ssh()
  File "/root/src/salt/salt/scripts.py", line 425, in salt_ssh
    client.run()
  File "/root/src/salt/salt/cli/ssh.py", line 23, in run
  File "/root/src/salt/salt/client/ssh/__init__.py", line 231, in __init__
    self.tgt_type)
  File "/root/src/salt/salt/roster/__init__.py", line 105, in targets
  File "/root/src/salt/salt/roster/flat.py", line 36, in targets
    return __utils__['roster_matcher.targets'](conditioned_raw, tgt, tgt_type, 'ipv4')
  File "/root/src/salt/salt/utils/roster_matcher.py", line 32, in targets
    rmatcher = salt.utils.roster_matcher.RosterMatcher(conditioned_raw, tgt,
AttributeError: 'module' object has no attribute 'roster_matcher'
```

Adding 
```
diff --git a/salt/utils/roster_matcher.py b/salt/utils/roster_matcher.py
index 145a047..40e7533 100644
--- a/salt/utils/roster_matcher.py
+++ b/salt/utils/roster_matcher.py
@@ -29,8 +29,7 @@ log = logging.getLogger(__name__)


 def targets(conditioned_raw, tgt, tgt_type, ipv='ipv4'):
-    rmatcher = salt.utils.roster_matcher.RosterMatcher(conditioned_raw, tgt,
-                                                       tgt_type, ipv)
+    rmatcher = RosterMatcher(conditioned_raw, tgt, tgt_type, ipv)
     return rmatcher.targets()


```

And it works

```
[root@salt salt]# salt-ssh \* test.ping
minion:
    True
```

dmacvicar:
I fixed the spurious namespace when using `RosterMatcher`.

I still get undefined `__opts__`:

```
  File "/space/git/salt/salt/salt/utils/roster_matcher.py", line 130, in get_data
    ret = copy.deepcopy(__opts__.get('roster_defaults', {}))
NameError: global name '__opts__' is not defined
```

Which confuses me. Other util modules, like `azurerm` do a explicit definition of `__opts__`:

```python
__opts__ = salt.config.minion_config('/etc/salt/minion')
__salt__ = salt.loader.minion_mods(__opts__)
```

`namecheap` and `vault` in `salt.utils` attemps to manually define `__salt___` if missing, but it assumes `__opts__` exist:

```python
    global __salt__
    if not __salt__:
        __salt__ = salt.loader.minion_mods(__opts__)
```

This happens both in the testcase and runtime.
`napalm` module does a weird thing with

dmacvicar:
Adding
```python
__opts__ = salt.config.minion_config('/etc/salt/minion')
```
in `roster_matcher.py` fixes the problem for me. Here is my working diff attached.

[working.diff.txt](https://github.com/saltstack/salt/files/2259384/working.diff.txt)

gtmanfred:
switch it so you use `__utils__`

Try this diff

https://gist.github.com/gtmanfred/854724d07ae240005f326a52bdb62172

dmacvicar:
Done

gtmanfred (approved):
🎉

dmacvicar:
So, what do I need to do now? :-)

gtmanfred:
Just waiting on reviews, I moved this over to the fluorine branch to make sure it gets in.

terminalmage (commented):
I would rather this be reviewed by someone familiar with Terraform, as I know nothing about it.

dubb-b:
@dmacvicar I have tested this out using the examples from your GH repo using libvirt. I really like the way you implemented this. I am working on a final few items here and I will be done with my review. One thing that may be kind of cool is using the example code that you have in your GH repo is great, although a working example in a Docker container might also be nice. This would allow people to see how this is all wired up in case they have issues with any of the steps listed. Overall this is a very nice feature to Salt and thank you for making it. I should be ready to give the thumbs up tomorrow.

dmacvicar:
Thanks for the feedback @dubb-b .

As the roster plugin needs the provider counter-part, it will be quite important to make sure people can use it. I am already providing builds of the code in my repo, and once the Salt part is released I will consider doing stable/tagged releases of it.

The installation should be as as easy as copying the builds to `terraform.d/plugins`. As I am building everything in [OBS](https://build.opensuse.org) (including Ubuntu builds), and _OBS_ supports building containers, it should be feasible providing a more _plug and play_ example.

The only thing that bothers me is that the official roster module depends on a very specific provider that for now is unreleased and hosted in my home project, however:

* Anybody implementing a provider that uses the same resource names can provide alternative implementations of this provider
* I am open to discuss how to make this provider "official", eg: moving it to the _terraform-providers` project so that it is auto-installed by terraform once it is used.
* All the above steps are worthy only if the roster module is part of _Salt_
* With all the disadvantages, I still think the new design/approach outweighs any disadvantage vs the old PR that had resource knowledge.

I think I could also create some kind of diagram and add it to the documentation later. It would also help understanding all the _wiring_, as you called it.

dubb-b (approved):
I am finished with the testing, it is working quite well. @rallytime 👍

cachedout:
@dmacvicar There are a few issues identified by CodeClimate. It's not required that they be fixed but could you please take a look and see if any of them seem reasonable to you?

https://codeclimate.com/github/saltstack/salt/pull/48873

dmacvicar:
I have reviewed those.

I think all of the issues validate my view on why I don't use or like CodeClimate. They are false alarms and the code is fine.

MalloZup:
awesome :+1:

PR #48874, opened 2018-08-01
Mwilson/reactor stats WIP

geekinutah:
### What does this PR do?
Adds stat tracking to ThreadPool used for Reactor. Periodically fires a reactor_threadpool_stats event.

46431

### Previous Behavior
No stat tracking

### New Behavior
See What does this PR do

### Tests written?

Not yet

### Commits signed with GPG?

No


geekinutah:
@cachedout Pls read over and let me know if this is what you had in mind.

rallytime:
@geekinutah There's some lint issues that need to be addressed with this change.

https://jenkinsci.saltstack.com/job/pr-lint/job/PR-48874/1/warnings52Result/

salt/utils/process.py
```diff
@@ -299,13 +309,35 @@ def __init__(self,
 
         self._workers = []
 
+        # keep stats on overflows, puts, tasks_done
+        self._stats = {'overflows': 0,
+                       'puts': 0,
+                       'tasks_done': 0
+                      }
+
+        self.event = salt.utils.event.get_event(
```
cachedout: Definitely only want to do this if a configuration option is set to gather stats.

cachedout (commented):
This is exactly the sort of approach I was envisioning. Well played!

cachedout:
@geekinutah Just wondering if you saw the above feedback and if there are any other questions that I can answer.

cachedout (approved):

geekinutah:
@cachedout I saw your feedback, thanks. I will re-write this addressing lint and some other obvious issues. What are your thoughts on tying this stats reporting to the enable_master_stats and master_stats_interval configs?

cachedout:
@geekinutah I think that's a good idea but to accomplish that I think that we need to break apart `master_stats` so that a user can switch on and off different types of stats at-will. If we can achieve that, which shouldn't be hard, I think we'd be in awesome shape!

salt/utils/process.py
```diff
@@ -288,7 +290,15 @@ class ThreadPool(object):
     '''
     def __init__(self,
                  num_threads=None,
-                 queue_size=0):
+                 queue_size=0,
+                 opts=None):
+        if not opts:
```
cachedout: I almost think we should make `opts` a required arg here. We should already have it generated by a caller anyway.

cachedout:
@geekinutah Some small lint errors here:

https://jenkinsci.saltstack.com/job/pr-lint/job/PR-48874/3/warnings52Result/new/

salt/utils/process.py
geekinutah: Okay, I think it sounds long term'ish to get a nice consistent paradigm for stats. For now I will keep this discrete for simplicity. I should be able to push an update to this soon :-).

gtmanfred:
@geekinutah can you update these lint errors? it looks like that is what is preventing the tests from running.

salt/utils/process.py
cachedout: OK, sounds good!

rallytime:
@geekinutah It looks like you were planning on pushing an update to this, based on your most recent comment. Were you able to come back to this?

geekinutah:
Sorry @rallytime, I haven't been able to get back to it yet. Feel free to finish, it might be a while (weeks) before I can finish up.

rallytime:
@geekinutah No problem at all. I will close this for now, and then when you're ready to come back to it, we can open it up again. Just let us know when you're ready. :)

PR #48876, opened 2018-08-02
Make IP_LEARNING optional

Vaelatern:
### What does this PR do?
Not always put a certain piece of XML in place, depending on how we intend to learn the network address.

In a network with one VM serving dhcp addresses, and other VMs connecting to it, the following problems were observed:

- Issue observed where in libvirt 4.4.0 the filterref block would ignore dhcp if set to 'any'.
- Issue observed where in libvirt 4.5.0 (latest at time of writing) the filterref block would permit the client to get an address, but then not permit any more traffic.

### Previous Behavior
Always make sure the filterref xml clause is added to the network configuration.

### New Behavior
If `ip_source` is not `ip-learning` (or blank, defaults to ip-learning), then don't alter the vm network configuration beyond removing the MAC address.

### Tests written?

No

### Commits signed with GPG?

No


rallytime (approved):
👍

PR #48877, opened 2018-08-02
Wrap ElementTree.tostring to make strs, not bytes

Vaelatern:
This is in use on a setup running python3 right now.
 Without this patch, assertions were thrown by python3-libvirt code that expected strings, not bytes.

### What does this PR do?
Make salt work better with python3. Untested with python2.
I didn't look for issue numbers.
### Previous Behavior
Within "/usr/lib/python3/dist-packages/libvirt.py"
```
def defineXMLFlags(self, xml, flags=0):
    ret = libvirtmod.virDomainDefineXMLFlags(self._o, xml, flags)
    if ret is None:raise libvirtError('virDomainDefineXMLFlags() failed', conn=self)
```

### New Behavior
Within "/usr/lib/python3/dist-packages/libvirt.py"
```
def defineXMLFlags(self, xml, flags=0):
    ret = libvirtmod.virDomainDefineXMLFlags(self._o, xml, flags)
    if ret is None:raise libvirtError('virDomainDefineXMLFlags() failed', conn=self)
    __tmp = virDomain(self,_obj=ret)
    return __tmp
```
### Tests written?

No

### Commits signed with GPG?

No.

### DIFF
--- a/salt/cloud/clouds/libvirt.py
+++ b/salt/cloud/clouds/libvirt.py
@@ -422,7 +422,7 @@ def create(vm_):
                 else:
                     raise SaltCloudExecutionFailure("Disk type '{0}' not supported".format(disk_type))
 
-            clone_xml = ElementTree.tostring(domain_xml)
+            clone_xml = salt.utils.stringutils.to_str(ElementTree.tostring(domain_xml))
             log.debug("Clone XML '%s'", clone_xml)
 
             validate_flags = libvirt.VIR_DOMAIN_DEFINE_VALIDATE if validate_xml else 0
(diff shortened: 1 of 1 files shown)


rares-pop:
There are similar calls inside salt/modules/virt.py. Should they be updated too?

rallytime (approved):

rallytime:
@rares-pop Probably.

@Vaelatern Would you be interested in a follow-up PR addressing the comment above?

I think this change looks good as it stands, but a follow up PR would be greatly appreciated!

Vaelatern:
Sure, I'll add it to my queue.

PR #48879, opened 2018-08-02
Adding an execution module providing helpers for the NAPALM formulas

mirceaulinic:
### What does this PR do?

Some time ago I started preparing some formulas for network configuration management with respect to the OpenConfig standards. The repositories are created, part of the formulas are available, e.g, https://github.com/saltstack-formulas/napalm-ntp-formula

Generally, all these formulas (unfortunately) only generate a text blob that is eventually loaded on the device. Therefore the formulas will 90% be some Jinja template with a considerable size. To shrink this size, and make it more human readable, I thought about having a module, that transforms blocks such as:

```jinja
{%- if subif_cfg.config.get('description') %}
  description {{ subif_cfg.config.description }};
{%- endif %}
```

Into a more simplified (when reading, at least):

```jinja
{{ salt.napalm_formula.render_field(subif_cfg.config, 'description') }}
```

With the same rendering result:

```
description "Some description";
```

Which does all these repetitive checks such as above, which would appear tens of times in a template. Besides, it also add the field name.

This module is generic enough, and it will be used by all the future NAPALM Formulas, e.g., https://github.com/saltstack-formulas/napalm-bgp-formula, https://github.com/saltstack-formulas/napalm-interfaces-formula etc.

Together with this, adding some more little helpers for various manipulation of the OpenConfig-like Python objects.

### DIFF
--- a/salt/modules/napalm_formula.py
+++ b/salt/modules/napalm_formula.py
@@ -22,9 +22,9 @@ import salt.ext.six as six
 import salt.utils.dictupdate
 from salt.defaults import DEFAULT_TARGET_DELIM
 try:
-    from salt.utils import traverse_dict_and_list
+    from salt.utils.data import traverse_dict_and_list as _traverse_dict_and_list
 except ImportError:
-    from salt.utils.data import traverse_dict_and_list
+    from salt.utils import traverse_dict_and_list as _traverse_dict_and_list
 
 __proxyenabled__ = ['*']
 __virtualname__ = 'napalm_formula'
@@ -150,7 +150,7 @@ def traverse(data, key, default=None, delimiter=DEFAULT_TARGET_DELIM):
 
         salt '*' napalm_formula.traverse "{'foo': {'bar': {'baz': True}}}" foo:baz:bar
     '''
-    return traverse_dict_and_list(data, key, default=default, delimiter=delimiter)
+    return _traverse_dict_and_list(data, key, default=default, delimiter=delimiter)
 
 
 def dictupdate(dest, upd, recursive_update=True, merge_lists=False):
@@ -221,7 +221,7 @@ def defaults(model,
                 continue
             log.debug('%s matches %s', model_path, default_path)
             # If there's a match, it will build the dictionary from the top
-            devault_val = traverse_dict_and_list(defaults_,
+            devault_val = _traverse_dict_and_list(defaults_,
                                                  default_path,
                                                  delimiter=delim)
             merged = setval(model_path, devault_val, dict_=merged, delim=delim)
--- a/salt/modules/napalm_formula.py
+++ b/salt/modules/napalm_formula.py
@@ -1,4 +1,4 @@
--*- coding: utf-8 -*-
+# -*- coding: utf-8 -*-
 '''
 NAPALM Formula helpers
 ======================
--- a/salt/modules/napalm_formula.py
+++ b/salt/modules/napalm_formula.py
@@ -1,0 +1,367 @@
+-*- coding: utf-8 -*-
+'''
+NAPALM Formula helpers
+======================
+
+.. versionadded:: Fluorine
+
+This is an Execution Module providing helpers for various NAPALM formulas,
+e.g., napalm-interfaces-formula, napalm-bgp-formula, napalm-ntp-formula etc.,
+meant to provide various helper functions to make the templates more readable.
+'''
+from __future__ import absolute_import, unicode_literals, print_function
+
+# Import python libs
+import copy
+import logging
+import fnmatch
+
+# Import salt modules
+import salt.utils.napalm
+import salt.ext.six as six
+import salt.utils.dictupdate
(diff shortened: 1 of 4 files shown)


rallytime:
@mirceaulinic Lint :)

https://jenkinsci.saltstack.com/job/pr-lint/job/PR-48879/1/warnings52Result/

mirceaulinic:
Ooops! It should be fixed now, sorry!

salt/modules/napalm_formula.py
```diff
@@ -0,0 +1,367 @@
...
+
+# Import salt modules
+import salt.utils.napalm
+import salt.ext.six as six
+import salt.utils.dictupdate
+from salt.defaults import DEFAULT_TARGET_DELIM
+try:
+    from salt.utils import traverse_dict_and_list
+except ImportError:
+    from salt.utils.data import traverse_dict_and_list
```
terminalmage: Since this code is going into develop, there is no need to do this try/except, since the function lives in `salt.utils.data` now. Not only that, there are three problems with this approach:

1. Importing a function into the module's global scope will add it to the `__salt__` dunder (i.e. `__salt__['napalm_formula.traverse_dict_and_list']`)
2. This will also add this function to the docs, which will be confusing.
3. Since we have compatibility functions in `salt.utils` for all the funcs that were moved for 2018.3, the first import will succeed and this will result in deprecation warnings.

Instead, you should simply `import salt.utils.data` and invoke the function with its full name.

gtmanfred (approved):

terminalmage (changes requested):

salt/modules/napalm_formula.py
gtmanfred: When possible, it would be great to also use `__utils__['data.traverse_dict_list']()` because then it will allow people to include that data.py file in their `salt://_utils/` directory if they are unable to upgrade to the newer version of salt.

terminalmage: Doing it @gtmanfred's way will also prevent you from needing to import the module at all.

mirceaulinic: Thanks guys for reviewing this and your thoughts. As Daniel mentioned, I made these imports having backwards potability in mind; using `__utils__` sounds like a better idea, thought it would require backporting loads of util modules.

Re 2: Indeed, I thought about that too, but it seems like Sphinx is smart enough not to pick these functions (no idea why) and they don't appear in the documentation (tested with both Salt and out internal documentation).

Anyway, I will change to use `__utils__` as it seems a better tradeoff.

mirceaulinic: Updated, though I just wanted to point this out: porting `utils.data` means it also requires porting `utils.dictupdate`, `utils.stringutils`, and `utils.yaml` (as they are imported there).

This is fine, though when I need, e.g., `files.fopen`, or `args.clean_kwargs`, I would also need to port loads of others such as `utils.jid`, `utils.path`, `utils.platform`, `utils.data`... and here's the catch: `salt.utils.args` needs `salt.utils.data` (which obviously won't be available). So not only that one basically needs to port almost all the modules under `utils/` but also they need to be patched (i.e., replace imports like the one from `utils.args` from `import salt.utils.data` to `import _utils.data`. This sounds like a nightmare to me.

Looking at this vs. 1 & 3 I'm not sure which is the worst...

mirceaulinic: What about the following:

```python
from salt.version import __version_info__
if __version_info__ < (2018, 3):
    from salt.utils import traverse_list_and_dict as _traverse_list_and_dict
else:
    from salt.utils.data import traverse_list_and_dict as _traverse_list_and_dict
```

Or why not, even simpler:

```python
try:
    from salt.utils.data import traverse_dict_and_list as _traverse_dict_and_list
except ImportError:
    from salt.utils import traverse_dict_and_list as _traverse_dict_and_list
```

This way we can ensure that the function won't be exported (note that I flipped the import error to prevent deprecation warning thrown on >= Oxygen).

Does this sound like a better alternative?

gtmanfred:
I definitely think importing the new method first would be the correct
path.  And we should add `__utils__` to the loader for utils in the next
major release.

salt/modules/napalm_formula.py
terminalmage: This does seem like a better alternative.

```python
try:
    from salt.utils.data import traverse_dict_and_list as _traverse_dict_and_list
except ImportError:
    from salt.utils import traverse_dict_and_list as _traverse_dict_and_list
```

However, I'm not sure why this should be necessary at all in `develop`.

gtmanfred: Yes, the question just comes down to, "How friendly to we want to be for allowing people to backport these modules?"

mirceaulinic: Added https://github.com/saltstack/salt/pull/48879/commits/3e9ccdb397ba4fea30c97dba2084b91e44db7287 - please let me know if this is okay for the time being.

PR #48896, opened 2018-08-02
Back-port #48730 to 2017.7

rallytime:
Back-port #48730 to 2017.7

### DIFF
--- a/salt/modules/pkgng.py
+++ b/salt/modules/pkgng.py
@@ -794,7 +794,16 @@ def install(name=None,
 
         .. code-block:: bash
 
-            salt '*' pkg.install <extended regular expression> pcre=True
+
+    batch
+        Use BATCH=true for pkg install, skipping all questions.
+        Be careful when using in production.
+
+        CLI Example:
+
+        .. code-block:: bash
+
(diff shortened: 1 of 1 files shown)


amontalban:
Thank you for backporting this @rallytime. Do you think it will reach 2017.7.8? Thanks!

rallytime:
@amontalban Yeah :)

fzipi:
Hey @rallytime !

I think you needed to change this for backporting to work.

```
       opts += 'x'
    if salt.utils.is_true(pcre):
        opts += 'X'
    if salt.utils.data.is_true(batch): # <========= PROBLEM
        env = {
                "BATCH": "true",
                "ASSUME_ALWAYS_YES": "YES"
}
```
There is no `data` in version `v2017`, it was in `develop` when I made that patch. So for this to work you need to get rid of `.data` there. Do you need me to create a patch for this?

The corrct version should be:
```
       opts += 'x'
    if salt.utils.is_true(pcre):
        opts += 'X'
    if salt.utils.is_true(batch): # <========== CORRECT VERSION
        env = {
                "BATCH": "true",
                "ASSUME_ALWAYS_YES": "YES"
}
```

PR #48903, opened 2018-08-03
Wrap ElementTree.tostring to make strs, not bytes

Vaelatern:
### What does this PR do?

It's the requested followup to #48877 

See #48877 

### Previous Behavior
Unknown, but this seems to be another place where strings are wanted.

### New Behavior
Hopefully it works, have not been able to test outright.

### Tests written?

No

### Commits signed with GPG?
No

### DIFF
--- a/salt/modules/virt.py
+++ b/salt/modules/virt.py
@@ -2010,7 +2010,7 @@ def snapshot(domain, name=None, suffix=None):
     n_name = ElementTree.SubElement(doc, 'name')
     n_name.text = name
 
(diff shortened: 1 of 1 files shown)


rallytime (approved):

rallytime:
@rares-pop ^

rares-pop:
The change looks good, but do we want it in develop too? There are more instances to be changed.

rallytime:
@rares-pop It will be merged-forward into develop. I just realized the fix was against 2018.3. 👍

PR #48904, opened 2018-08-03
No rehashing in parallel

KaiSforza:
Rehashing in parallel causes pyenv to fail, if they're run at almost the
exact same time. It's breaking our lint tests pretty horribly.

### What does this PR do?

Allows our lint tests to pass

### Previous Behavior
Lint tests would fail in either `salt linting` or `test linting` during the rehash.

### New Behavior
Tests pass.

@gtmanfred @dubb-b

### DIFF
--- a/.ci/lint
+++ b/.ci/lint
@@ -29,13 +29,13 @@ pipeline {
             parallel {
                 stage('salt linting') {
                     steps {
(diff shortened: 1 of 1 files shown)


dubb-b (approved):

gtmanfred (changes requested):
PR lint is failing, rerunning.

gtmanfred (approved):

PR #48779, opened 2018-07-26
Two new features for the napalm configuration management: commit_in and commit_at

mirceaulinic:
### What does this PR do?

In the networking world, there are a very small number of platforms that allow you to commit a specific configuration change at a desired time. Although this is not particularly one of the best practices, turns out to be important sometimes (very rarely though). Nevertheless it is and will forever be optional.

With these changes, I am adding support for ``commit_in`` and ``commit_at`` to commit the config changes later. This simply leverages the abilities of the Salt scheduler, by adding a job to be executed once, at the time requested by the user. Therefore, this is available for any network platform managed through NAPALM, regardless if they do or don't support scheduled commit capabilities.

For safety reasons, I am adding the `net.cancel_commit` counterpart execution function to discard a scheduled commit.

Example:

```bash
# salt veos net.load_config text='ntp server 4.3.2.1' commit_in=2m
veos:
    ----------
    already_configured:
        False
    comment:
        Changes discarded for now, and scheduled commit at: 2018-07-26T10:06:33.
        The commit ID is: 20180726100429648495.
        To discard this commit, you can execute:

        salt veos net.cancel_commit 20180726100429648495
    diff:
        @@ -5,6 +5,7 @@
         transceiver qsfp default-mode 4x10G
         !
         ntp server 1.2.3.4
        +ntp server 4.3.2.1
         ntp server 5.6.7.8
         ntp server 10.10.10.1
         !
    loaded_config:
    result:
        True
```

PR #48909, opened 2018-08-03
Add napalm commit confirmed features

mirceaulinic:
### What does this PR do?

In line with my previous PR https://github.com/saltstack/salt/pull/48779, adding now two new features `revert_in` and `revert_at` that would schedule a configuration revert, similar to what we have on Junos (where is called "commit confirmed"). I opted to use these names as I find it makes it more obvious that the configuration will be reverted unless instructed otherwise.

The main problem is not the naming nevertheless, but the fact that not all the platforms support this natively (only Junos AFAIK). However, via Salt this would be available on any platform, but with some caveats and the user must be very cautious when using this (not that the feature itself can cause issues, but to have the right expectations when using this).

Usage example:

```bash
$ sudo salt veos-lab net.load_config text='ntp server 172.17.17.2' revert_in=2m
veos-lab:
    ----------
    already_configured:
        False
    comment:
        The commit ID is: 20180802145541882500.
        This commit will be reverted at: 2018-08-02T14:57:49, unless confirmed.
        To confirm the commit and avoid reverting, you can execute:

        salt veos-lab net.confirm_commit 20180802145541882500
    diff:
        @@ -5,6 +5,7 @@
         transceiver qsfp default-mode 4x10G
         !
         ntp server 172.17.17.1
        +ntp server 172.17.17.2
         !
         snmp-server location Tampa, FL
         !
    loaded_config:
    result:
        True
$ sudo salt veos-lab net.confirm_commit 20180802145541882500
veos-lab:
    ----------
    comment:
        Commit #20180802145541882500 confirmed.
    result:
        True
```

With this PR, I am also:

- Cleaning up the schedule after the job is executed (so we don't endup with tens of scheduled jobs)
- Store the config schedule to be loaded into a temporary file from where the Minion will read it, as in opposite to storing this in the schedule directly - inefficient especially as the config can be huge.
- Remove the temporary files after using them.
- Add the `commit_confirmed` and `commit_cancelled` State functions to the `netconfig` State.
- The functions would return a `commit_id` key which can be used using the `__slots__` feature to be sent to the state functions mentioned above.

### DIFF
--- a/salt/modules/napalm_mod.py
+++ b/salt/modules/napalm_mod.py
@@ -780,6 +780,55 @@ def junos_rpc(cmd=None, dest=None, format=None, **kwargs):
 
 
 @proxy_napalm_wrap
+def junos_commit(**kwargs):
+    '''
+    .. versionadded:: Fluorine
+
+    Commit the changes loaded in the candidate configuration.
+
+    dev_timeout: ``30``
+        The NETCONF RPC timeout (in seconds).
+
+    comment
+      Provide a comment for the commit.
+
+    confirm
+      Provide time in minutes for commit confirmation. If this option is
+      specified, the commit will be rolled back in the specified amount of time
+      unless the commit is confirmed.
+
+    sync: ``False``
+      When ``True``, on dual control plane systems, requests that the candidate
+      configuration on one control plane be copied to the other control plane,
+      checked for correct syntax, and committed on both Routing Engines.
+
+    force_sync: ``False``
+      When ``True``, on dual control plane systems, force the candidate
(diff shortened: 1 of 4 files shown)


cachedout (commented):
CodeClimate has a few ideas about some possible refactoring for DRY and breaking up some of the nested logic but otherwise this looks good!

mirceaulinic:
Hi @cachedout - I looked through the warnings from codeclimate, but I couldn't identify which one I could fix (some are not relevant to these changes though)...

cachedout (approved):

gtmanfred (approved):

PR #48910, opened 2018-08-03
Two more options added to the netbox Pillar, and use salt.utils.http

mirceaulinic:
### What does this PR do?

I am applying some small (or I think so) changes to the `netbox` external Pillar:

- Use `salt.utils.http` instead of `requests` to avoid external dependency.
- Add two new options: `site_details` and `site_prefixes`.
- `versionadded` tag.

@ksofa2 @ngrundler I hope you don't mind my changes, please let me know if there's anything you disagree / should be changed etc. Additionally, it would be great if you could test this out and confirm it doesn't change anything in your environment. Thanks!

### DIFF
--- a/salt/pillar/netbox.py
+++ b/salt/pillar/netbox.py
@@ -2,115 +2,159 @@
 '''
 A module that adds data to the Pillar structure from a NetBox API.
 
+.. versionadded:: Fluorine
 
 Configuring the NetBox ext_pillar
-====================================
+---------------------------------
 
 .. code-block:: yaml
 
   ext_pillar:
     - netbox:
         api_url: http://netbox_url.com/api/
-
-The following are optional, and determine whether or not the module will
-attempt to configure the ``proxy`` pillar data for use with the napalm
-proxy-minion:
-
-.. code-block:: yaml
-
-        proxy_return: True
-        proxy_username: admin
         api_token: 123abc
 
 Create a token in your NetBox instance at
 http://netbox_url.com/user/api-tokens/
 
-By default, this module will query the NetBox API for the platform associated
-with the device, and use the 'NAPALM driver' field to set the napalm
-proxy-minion driver. (Currently only 'napalm' is supported for drivertype.)
+The following options are optional, and determine whether or not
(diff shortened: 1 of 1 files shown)


ngrundler:
@mirceaulinic 

No issues on my end, this looks great.  A good learning example for me about how I can use features built into salt to remove dependencies and make things simpler! 

What do you think about nesting the different NetBox API endpoints into the `netbox` pillar key, e.g.

```
"netbox": {
    "device": { <device details> },
    "site": { <site details> },
    "interfaces": { <interface details> },
    "ip-addresses": { <ip address details> }
 }
```
I've since made some tweaks internally since we're trying to use netbox data to drive configuration templates and I needed this data available, so I have set up queries to interfaces, addresses, connections, etc.  I was planning on adding this to the module myself soon(ish), but I see no reason for that to impact this PR.

mirceaulinic:
> What do you think about nesting the different NetBox API endpoints into the netbox pillar key, e.g.

That would make sense to me, yes.

Ichabond:
> What do you think about nesting the different NetBox API endpoints into the netbox pillar key, e.g.

I think that's a great idea, but I'd nest them in the App endpoint as well. So instead of the flatter structure you proposed, I propose:
```
"netbox": {
    "dcim": {
        "device": { <device details> },
        "site": { <site details> },
        "interfaces": { <interface details> }        
    },
    "ipam": {
        "ip-addresses": { <ip address details> }
    }
 }
```

This will make sure there's no conflicts if/when Netbox has 2 apps with identical API endpoints.

mirceaulinic:
I think these proposals make sense to me, I would leave it though for another day to discuss this further. Wondering if @zachmoody has any thoughts on this?

zachmoody:
TIL that `salt.utils.http` doesn't use `requests`, by default anyways. That's impressive. :smile: 

I'd definitely nest under app though, ran into the same conundrum with earlier versions of pynetbox and in the end had to acquiesce to a two-tiered hierarchy.

rallytime (approved):
Nice

PR #48912, opened 2018-08-03
make jboss7.py compatible with 2018.3.0 and later

maxim-sermin:
### What does this PR do?
The functions `file.get_managed` and `file.manage_file` were updated in Release [2018.3.0](https://docs.saltstack.com/en/latest/ref/modules/all/salt.modules.file.html#salt.modules.file.get_managed) such that the new argument `attrs` was added to both. This PR updates the state `jboss7.py` so it calls those functions with a value of `None` for the newly added argument.

### Previous Behavior
Functions `file.get_managed` and `file.manage_file` could not be called by `jboss7.py` due to wrong number for arguments.

### New Behavior
Functions `file.get_managed` and `file.manage_file` are now correctly called by `jboss7.py` allowing deployment to succeed.

### Tests written?

No

### Commits signed with GPG?

No


rallytime (approved):
👍

cachedout (approved):

PR #48917, opened 2018-08-03
[develop] Merge forward from 2018.3 to develop

rallytime:
Conflicts:
  - salt/modules/nilrt_ip.py
  - salt/states/pip_state.py
  - tests/unit/modules/test_file.py


rares-pop:
@rallytime - please drop the salt/modules/nilrt_ip.py changes. What is in 'develop' is the correct code. I will double-check this on Monday with a colleague, and in case we need to port those changes, I will create a PR for that.

Some background: National Instruments are actively maintaining two different linux distributions. They are referred either by 'currentGen' and 'nextGen' NILinuxRT, or 'legacy' and 'current'. Before our changes in 'develop', we used to have upstream-ed only nextGen changes, and have currentGen changes in a private repo. That changed for Fluorine.

rallytime:
@rares-pop Ah, OK. That file did conflict, and it looked to me like the other changes were the "correct" ones. I chose poorly. :)

I'll fix that up.

rallytime:
@rares-pop Please review. :)

rares-pop:
Things look good! Thank you!

PR #48945, opened 2018-08-06
Rethinking the way configuration management for network devices is handled

mirceaulinic:
### What does this PR do?

There are two root causes for these changes:

1. The community asked several times to be able to render ore than one template in one go. This is due to the fact that many would prefer having the templates grouped logically and eventually reuse them in other contexts. In other words, break a big Jinja template into smaller and easier to digest parts, that can be later reused in other states.

I agree with this, and with these changes, this is going to be possible from now on. For example, a state such as:

```sls
hostname_and_ntp:
  netconfig.managed:
    - template_name:
        - https://bit.ly/2OhSgqP
        - https://bit.ly/2M6C4Lx
        - https://bit.ly/2OIWVTs
    - debug: true
    - context:
        hostname: {{ opts.id }}
        servers:
          - 172.17.17.1
          - 172.17.17.2
        peers:
          - 192.168.0.1
          - 192.168.0.2
```

Would totally be valid (previously only `template_name: https://bit.ly/2OhSgqP` was possible).

2. When I firstly implemented the `net.load_template` function more than 2 years ago, I still wasn't fully comfortable with Salt and I didn't have enough experience to evaluate what's the best way to render templates on the fly. This is why I ended up using `file.get_managed`: I was reading the template contents, dumping the rendered result into a temporary file, then reading the contents from there and load them into the device. Not only this is suboptimal, but Salt already has `file.apply_template_on_contents` for this exact purpose, without creating temporary files and all that mess. Using `file.get_managed` made me add a bunch of arguments such as `template_user`, `template_mode` etc., which never made sense, nobody actually used those as they refer to the temporary file which is anyway removed after this function is executed.

With this said, I'm swapping the usage of the `file.get_managed` with the more appropriate `file.apply_template_on_contents`. This also means that the `template_user`, `template_mode`, `template_group` et. al. arguments would no longer be used.

I have removed these arguments in https://github.com/saltstack/salt/commit/b5a642e1171dab75bff247ef22bfa27cf93b6969, but in case this should better be done in a later release I am happy to revert this commit and have the deprecated tags for the time being: https://github.com/saltstack/salt/commit/2ea0cbe159f4b7cbe0a5706b9488f00cff8c309a

CC @gtmanfred 

Notable byproducts:

- Faster template rendering.
- Fewer lines of code to maintain.

--------

After discussion with @gtmanfred, @rallytime and @terminalmage, (2) is not enough good reason to drop the template hash arguments, therefore we'll continue using `file.get_managed`. However, deprecating `template_user`, `template_group`, etc.

### DIFF
--- a/salt/modules/napalm_network.py
+++ b/salt/modules/napalm_network.py
@@ -2033,6 +2033,13 @@ def load_template(template_name=None,
                 }
             )
     else:
+        salt.utils.versions.warn_until(
+            'Sodium',
+            'Native NAPALM templates support will be removed in the Sodium '
+            'release. Please consider using the Salt rendering pipeline instead.'
+            'If you are using the \'netntp\', \'netsnmp\', or \'netusers\' Salt '
+            'State modules, you can ignore this message'
+        )
         # otherwise, use NAPALM render system, injecting pillar/grains/opts vars
         load_templates_params = defaults if defaults else {}
         load_templates_params.update(template_vars)
--- a/salt/modules/napalm_network.py
+++ b/salt/modules/napalm_network.py
@@ -1547,7 +1547,7 @@ def load_config(filename=None,
 
 
 @salt.utils.napalm.proxy_napalm_wrap
-def load_template(template_name,
+def load_template(template_name=None,
                   template_source=None,
                   template_hash=None,
                   template_hash_name=None,
--- a/salt/modules/napalm_network.py
+++ b/salt/modules/napalm_network.py
@@ -1373,7 +1373,7 @@ def load_config(filename=None,
             applies a manual configuration change, or a different process or
             command changes the configuration in the meanwhile).
 
-        .. versionadded: Fluorine
+        .. versionadded:: Fluorine
 
     commit_at: ``None``
         Commit the changes at a specific time. Example of accepted formats:
@@ -1397,7 +1397,7 @@ def load_config(filename=None,
             applies a manual configuration change, or a different process or
             command changes the configuration in the meanwhile).
 
-        .. versionadded: Fluorine
+        .. versionadded:: Fluorine
 
     revert_in: ``None``
         Commit and revert the changes in a specific number of minutes / hours.
@@ -1428,7 +1428,7 @@ def load_config(filename=None,
             commit and till the changes are reverted), these changes would be
             equally reverted, as Salt cannot be aware of them.
 
-        .. versionadded: Fluorine
+        .. versionadded:: Fluorine
 
     revert_at: ``None``
         Commit and revert the changes at a specific time. Example of accepted
@@ -1458,7 +1458,7 @@ def load_config(filename=None,
             commit and till the changes are reverted), these changes would be
             equally reverted, as Salt cannot be aware of them.
 
-        .. versionadded: Fluorine
+        .. versionadded:: Fluorine
 
     saltenv: ``base``
         Specifies the Salt environment name.
@@ -1715,7 +1715,7 @@ def load_template(template_name=None,
             applies a manual configuration change, or a different process or
             command changes the configuration in the meanwhile).
 
-        .. versionadded: Fluorine
+        .. versionadded:: Fluorine
 
     commit_at: ``None``
         Commit the changes at a specific time. Example of accepted formats:
@@ -1739,7 +1739,7 @@ def load_template(template_name=None,
             applies a manual configuration change, or a different process or
             command changes the configuration in the meanwhile).
 
-        .. versionadded: Fluorine
+        .. versionadded:: Fluorine
 
     revert_in: ``None``
         Commit and revert the changes in a specific number of minutes / hours.
@@ -1770,7 +1770,7 @@ def load_template(template_name=None,
             commit and till the changes are reverted), these changes would be
             equally reverted, as Salt cannot be aware of them.
 
-        .. versionadded: Fluorine
+        .. versionadded:: Fluorine
 
     revert_at: ``None``
         Commit and revert the changes at a specific time. Example of accepted
@@ -1800,7 +1800,7 @@ def load_template(template_name=None,
             commit and till the changes are reverted), these changes would be
             equally reverted, as Salt cannot be aware of them.
 
-        .. versionadded: Fluorine
+        .. versionadded:: Fluorine
 
     defaults: None
         Default variables/context passed to the template.
@@ -2219,7 +2219,7 @@ def cancel_commit(jid):
     Cancel a commit scheduled to be executed via the ``commit_in`` and
     ``commit_at`` arguments from the
     :py:func:`net.load_template <salt.modules.napalm_network.load_template>` or
-    :py:func:`net.load_config <salt.modules.napalm_network.load_config`
+    :py:func:`net.load_config <salt.modules.napalm_network.load_config>`
     execution functions. The commit ID is displayed when the commit is scheduled
     via the functions named above.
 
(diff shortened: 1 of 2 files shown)


salt/modules/napalm_network.py
```diff
@@ -1547,18 +1547,12 @@ def load_config(filename=None,
 
 
 @salt.utils.napalm.proxy_napalm_wrap
-def load_template(template_name,
+def load_template(template_name=None,
                   template_source=None,
-                  template_hash=None,
```
rallytime: We can't remove any of these older kwargs without putting them on a deprecation path first. We'll need to support these with a warning for 2 feature releases. (So, Sodium would be the version to wan until and then remove the kwargs.) You can use the `warn_until` function in `salt.utils.versions`.

rallytime (changes requested):

salt/modules/napalm_network.py
mirceaulinic: I understand that I should revert https://github.com/saltstack/salt/pull/48945/commits/6a70bb1d27a9141ca04d0a3519171c2b074b9588.
In that case, they will no longer be used (although defined in the function header). How would you suggest to proceed? Leave them defined and unused?
Sorry I know this is a bit controversial, it's also the first time I'm deprecating something so I'm not sure what's the best way. Also, please remember that the arguments in subject, as of today, are not used / don't make sense to be there in the first place. Thanks!

gtmanfred: I do not like removing the template_hash.  We should probably still keep this, even though cp.get_template does not support giving it a hash.  We should be verifying the hash manually as part of the load_template process.

gtmanfred: so, you have `**template_vars`, so you don't want anyone that is still passing stuff to this, to inject extra arguments to the template_vars.

I would put a for loop at the top of the function, that iterates over all of the arguments here, and just tells the user that they do not need to specify them, and in 2 releases, they will be passed to template_vars if they do not stop using them.

That should be good enough, then in 2 releases we will just delete all that, and delete these variables from the function definition.

salt/states/netconfig.py
```diff
@@ -454,16 +408,10 @@ def saved(name,
...
             template_source=None,
-            template_hash=None,
-            template_hash_name=None,
-            template_user='root',
-            template_group='root',
-            template_mode='755',
-            template_attrs='--------------e----',
-            saltenv=None,
             template_engine='jinja',
-            skip_verify=False,
```
gtmanfred: These should be handled the same way as they are in the load_template module.

gtmanfred (changes requested):

salt/modules/napalm_network.py
mirceaulinic: Thanks @gtmanfred - I will update this branch tomorrow.

rallytime:
@mirceaulinic I have changed the base branch of this PR from `develop` to `fluorine` for inclusion in the Fluorine feature release.

terminalmage:
I spoke with @mirceaulinic a few minutes ago and helped clear up how the file state/module work internally. I think the confusion comes from the fact that `file.get_managed` does not have default values for most of its arguments. This is because it's meant to be used internally by the file states, which _do_ have default values for these arguments. But if you're invoking the `file.managed` state via the `__states__` dunder, you don't actually need to specify the `user`, `group`, `mode`, or `attrs` arguments. When they are not specified in a `file.managed` state, they assume the defaults (which in most cases means we don't enforce the ownership/mode/whatever).

However, it would still be good to have a function in the `file` module which can accept

terminalmage:
I spoke with @mirceaulinic a few minutes ago and helped clear up how the file state/module work internally. I think the confusion comes from the fact that `file.get_managed` does not have default values for most of its arguments. This is because it's meant to be used internally by the file states, which _do_ have default values for these arguments. But if you're invoking the `file.managed` state via the `__states__` dunder, you don't actually need to specify the `user`, `group`, `mode`, or `attrs` arguments. When they are not specified in a `file.managed` state, they assume the defaults (which in most cases means we don't enforce the ownership/mode/whatever).

However, it would still be good to have a function in the `file` module which accepts mostly the same arguments as `file.get_managed` but instead of returning the path to a file, we can instead just return the file's contents. But that would be something to consider for a future feature release.

In the meantime, I think it would be good to do the following:

1. Continue to use `file.managed`.
2. Set the defaults for `template_user`, `template_group`, `template_mode`, and `template_attrs` to `None`. If any of these have a non-`None` value, we should then add a warning to the state return:

   ```python
   if template_user is not None:
       ret.setdefault('warnings', []).append(
           'The \'template_user\' argument is deprecated and '
           'has been ignored'
       )
   ```

I am also not a fan of using arguments like `template_name`, `template_source_name`, etc... If we are using the underlying file state/module support, then we should use the same argument names to provide a more uniform interface (similar to how we do things in the cron and archive state modules, which invoke `file` states to handle things like templating a source file).

salt/modules/napalm_network.py
```diff
@@ -2018,6 +2033,13 @@ def load_template(template_name,
                 }
             )
     else:
+        salt.utils.versions.warn_until(
```
mirceaulinic: I should have removed this in `Fluorine` (https://docs.saltstack.com/en/develop/topics/releases/fluorine.html#module-deprecations), but it turns out that it has more implications than I initially thoughts, postponing this till `Sodium` as well. Is that alright?

gtmanfred: :+1:

salt/modules/napalm_network.py
```diff
@@ -1899,6 +1892,14 @@ def load_template(template_name,
...
     loaded_config = None
     # prechecks
+    deprecated_args = ('template_user', 'template_attrs', 'template_group', 'template_mode')
+    for deprecated_arg in deprecated_args:
+        if template_vars.get(deprecated_arg):
+            salt.utils.versions.warn_until(
+                'Sodium',
+                ('The \'{arg}\' argument to \'net.load_template\' is deprecated '
+                 'and has been ignored').format(arg=deprecated_arg)
+            )
```
gtmanfred: Can we remove these args from template args?

Should be a simple `del template_vars[deprecated_arg]`

gtmanfred (changes requested):
One small change, otherwise this looks good.

salt/modules/napalm_network.py
mirceaulinic: Shall do.

salt/modules/napalm_network.py
```diff
@@ -1946,55 +1950,66 @@ def load_template(template_name,
...
-                _temp_tpl_file = _managed[0]
-                _temp_tpl_file_exists = __salt__['file.file_exists'](_temp_tpl_file)
-                if not _temp_tpl_file_exists:
+            if not isinstance(template_name, (list, tuple)):
+                template_name = [template_name]
+            if template_hash_name and not isinstance(template_hash_name, (list, tuple)):
+                template_hash_name = [template_hash_name]
+            elif not template_hash_name:
+                template_hash_name = [None] * len(template_name)
+            if template_hash and isinstance(template_hash, six.string_types) and not\
```
mirceaulinic: I was wondering what's @terminalmage's take on this? Does this make sense to you, or would you suggest doing this in a better way? Thanks.

rallytime (approved):

salt/modules/napalm_network.py
gtmanfred: Once this is added, this should be good to merge IMO

gtmanfred (approved):

terminalmage (approved):

PR #48946, opened 2018-08-06
Fixed Python 3 incompatibility in methods in nilrt_ip and debian_ip.

rares-pop:
This is a forward-port (cherry-pick) of this:
https://github.com/saltstack/salt/pull/48844

after I successfully mislead @rallytime here:
https://github.com/saltstack/salt/pull/48917

@rallytime - sorry man, I was wrong. I wasn't aware of the first PR and I tested our implementation fairly recently, but only with py2. Now I did test it on NILinuxRT with py2 with the new changes and works as a charm. So we should keep those.


rallytime (approved):
👍

PR #48948, opened 2018-08-06
Add force reconnection feature to NAPALM Proxy Minions

mirceaulinic:
### What does this PR do?

One of the major disadvantages of the Proxy Minions is their rigidity in terms of connectivity: the connection with the remote device is established with the remote device at the time of the Minion startup and that connection is going to be used forever.

If one would execute a command on the device using different parameters (due to various causes, such as unable to authenticate with the user specified in the Pillar and the authentication system - say TACACS+ is not available, or the user has a DNS in the Pillar and the resolver is currently down and would like to use temporarily the IP address instead, etc.), it requires a updating the Pillar data and Proxy Minion process restart. This is less than ideal.

Salt fortunately is as always flexible enough and we can add a magic argument, say `force_reconnect` which tells to establish a separate connection with the args specified in-line. Example:

Regular behaviour:

```bash
root@salt-master:/# salt napalm net.arp
napalm:
    ----------
    comment:
    out:
        |_
          ----------
          age:
              None
          interface:
              em1.0
          ip:
              128.0.0.16
          mac:
              02:00:00:00:00:10
        |_
          ----------
          age:
              1275.0
          interface:
              fxp0.0
          ip:
              172.31.0.1
          mac:
              02:8B:CA:89:F9:35
    result:
        True
```

With `force_reconnect`:

```
root@salt-master:/# salt napalm net.arp force_reconnect=True host=fake
napalm:
      File "/usr/lib/python2.7/dist-packages/salt/minion.py", line 1635, in _thread_return
        return func(*args, **kwargs)
      File "/usr/lib/python2.7/dist-packages/salt/utils/napalm.py", line 393, in func_wrapper
        wrapped_global_namespace['napalm_device'] = get_device(opts)
      File "/usr/lib/python2.7/dist-packages/salt/utils/napalm.py", line 340, in get_device
        network_device.get('DRIVER').open()
      File "/usr/local/lib/python2.7/dist-packages/napalm/junos/junos.py", line 107, in open
        self.device.open()
      File "/usr/local/lib/python2.7/dist-packages/jnpr/junos/device.py", line 1284, in open
        raise EzErrors.ConnectUnknownHostError(self)
    ConnectUnknownHostError: ConnectUnknownHostError(fake)
```

The above was used as an example proving that when `force_reconnect=True` it would indeed attempt to initiate a separate connection.

### DIFF
--- a/salt/proxy/napalm.py
+++ b/salt/proxy/napalm.py
@@ -3,6 +3,8 @@
 NAPALM: Network Automation and Programmability Abstraction Layer with Multivendor support
 =========================================================================================
 
+.. versionadded:: 2016.11.0
+
 Proxy minion for managing network devices via NAPALM_ library.
 
 :codeauthor: Mircea Ulinic <mircea@cloudflare.com> & Jerome Fleury <jf@cloudflare.com>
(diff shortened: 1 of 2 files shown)


rallytime (changes requested):
This needs to be documented somewhere and added to the release notes.

mirceaulinic:
Definitely @rallytime - I will add this to the NAPALM Proxy docs. For the release notes, I'm planning to finish some more things, and I'll add to the release notes everything I can add into Fluorine. Does this make sense?

rallytime (approved):

gtmanfred (approved):

PR #48954, opened 2018-08-06
updated doc string

rdutch:
Updated doc string to add an example.

### DIFF
--- a/salt/beacons/log.py
+++ b/salt/beacons/log.py
@@ -79,8 +79,7 @@ def beacon(config):
               - tags:
                   <tag>:
                     regex: <pattern>
-    '''
-    '''
+    
     The defined tag is added to the beacon event tag.
     This is not the tag in the log.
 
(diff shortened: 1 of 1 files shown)


salt/beacons/log.py
```diff
@@ -80,6 +80,20 @@ def beacon(config):
                   <tag>:
                     regex: <pattern>
     '''
+    '''
```
rallytime: Did you mean to replace the doc string above with this new one? As this sits presently, there are 2 docstring sections stacked one on top of the other. :)

rallytime (commented):
One quick question here.

salt/beacons/log.py
rdutch: No, I wanted to add a more clear example, I can make it all into 1 doc string if that is preferred.

rallytime (approved):

Issue #48952, opened 2018-08-06
document `regex_replace` Jinja filter


mbunkus:
### Description of Issue/Question

The ["Understanding Jinja" section](https://docs.saltstack.com/en/develop/topics/jinja/index.html#regex-match) documents and lists `regex_match` and `regex_search` but not `regex_replace`. However, that function is actually present in `salt/utils/jinja.py` and works just fine.

It should really be documented. I've been Googling for ways to regex-replace in Jinja several times now, seeing as the documentation stated such a function doesn't exist, and only a look under the hood (into `jinja.py`) revealed that it does exist after all.

### Steps to Reproduce Issue

Small test state that shows the function works just fine:

```
{% set a = '/tmp/hellely' %}
{{ a | regex_replace('l(.)', 'a\\1') }}:
  file.directory
```

Saving this as `testme.sls` & executing it (`salt-call testme`) will create `/tmp/healeay` as expected.

### Versions Report

```
Salt Version:
           Salt: 2018.3.2

Dependency Versions:
           cffi: Not Installed
       cherrypy: Not Installed
       dateutil: Not Installed
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.10
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: 0.30.1
           Mako: 1.0.7
   msgpack-pure: Not Installed
 msgpack-python: 0.5.6
   mysql-python: Not Installed
      pycparser: Not Installed
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: Not Installed
         Python: 2.7.15 (default, Jun 27 2018, 13:05:28)
   python-gnupg: Not Installed
         PyYAML: 3.12
          PyZMQ: 17.0.0
           RAET: Not Installed
          smmap: Not Installed
        timelib: Not Installed
        Tornado: 4.5.3
            ZMQ: 4.2.3

System Versions:
           dist:
         locale: UTF-8
        machine: x86_64
        release: 4.17.5-1-ARCH
         system: Linux
        version: Not Installed
```

garethgreenaway:
@mbunkus Thanks for the report.  It looks like the documentation for this function is available in salt/utils/jinja.py so the docstring for the `regex_search` function would simply need to be copied into salt/doc/topics/jinja/index.rst. We would definitely welcome a pull request with this addition.

mbunkus:
[...]

PR #48958, opened 2018-08-06
document `regex_replace` Jinja filter

mbunkus:
### What does this PR do?

It adds documentation for the existing `regex_replace` Jinja filter to the "Understanding Jinja" document.

#48952 

### Tests written?

No, but I did build the HTML docs and verified that the resulting HTML page looks correct.

### Commits signed with GPG?

Yes

### DIFF
--- a/doc/topics/jinja/index.rst
+++ b/doc/topics/jinja/index.rst
@@ -398,6 +398,29 @@ Returns:
   None
 
 
+.. jinja_ref:: regex_replace
+
+``regex_replace``
+-----------------
+
+.. versionadded:: 2017.7.0
+
+Searches for a pattern and replaces with a sequence of characters.
+
+Example:
+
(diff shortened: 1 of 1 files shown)


rallytime (approved):

lorengordon:
@rallytime It would be awesome to backport this simple doc patch back to 2018.3 and 2019.2, so it showed up in the current stable docs... Right now it only shows up in the "develop" docs...

rallytime:
@lorengordon For sure, good call.

@garethgreenaway or @Ch3LL can you help out here?

Ch3LL:
Thanks for the heads up @rallytime I added the label and it will get backported here soon. thanks!

sagetherage:
Notes from Community Port Jam: already in master-skipped

PR #48989, opened 2018-08-08
Add insert matching to file.blockreplace

nicholasmhughes:
### What does this PR do?
This PR adds the ability to regex match a line in the managed file in order to place the block above or below that line.

N/A

### Previous Behavior
Blocks could only be placed at the beginning or end of files if the block markers were not found.

### New Behavior
Blocks can now be placed at arbitrary places in files based upon pattern matching. The first match (from the top of the file) is used to place the block.

### Tests written?
Yes

### Commits signed with GPG?
Yes


salt/modules/file.py
```diff
@@ -2676,16 +2691,33 @@ def _add_content(linesep, lines=None, include_marker_start=True,
...
+                    if insert_before_match:
+                        match_idx += 1
+                block_found = True
+
+    if not block_found:
+        raise CommandExecutionError(
+            'Cannot edit marked block. Markers were not found in file.'
+        )
 
     if block_found:
```
DmitryKuzmenko: This `if` is redundant. `block_found` is definitely `True` if we're here.

DmitryKuzmenko (changes requested):
Looks good to me. Just one minor complaint.

salt/modules/file.py
nicholasmhughes: @DmitryKuzmenko you will hit this section if you don't have any of the `append...`, `prepend...`, or `insert...` parameters set.

nicholasmhughes: ah. got it.

nicholasmhughes: @DmitryKuzmenko please take a look when you get a chance and see if this is what you intended.

DmitryKuzmenko (approved):
:+1:

salt/modules/file.py
```diff
@@ -2449,6 +2449,8 @@ def blockreplace(path,
         content='',
         append_if_not_found=False,
         prepend_if_not_found=False,
+        insert_before_match=None,
```
rallytime: These need to go at the end of the list so we don't cause trouble with people calling these kwargs positionally.

salt/modules/file.py
```diff
@@ -2494,6 +2496,17 @@ def blockreplace(path,
         If markers are not found and set to ``True`` then, the markers and
         content will be prepended to the file.
 
+    insert_before_match
+        If markers are not found, this parameter can be set to a regex which will
+        insert the block before the first found occurrence in the file.
+
+        .. versionadded:: Fluorine
```
rallytime: This needs to be `Neon` now. :)

salt/states/file.py
```diff
@@ -4627,6 +4629,18 @@ def blockreplace(
         If markers are not found and this option is set to ``True``, the
         content block will be prepended to the file.
 
+    insert_before_match
+        If markers are not found, this parameter can be set to a regex which will
+        insert the block before the first found occurrence in the file.
+
+        .. versionadded:: Fluorine
```
rallytime: Same comment here about `Neon`.

salt/states/file.py
```diff
@@ -4499,6 +4499,8 @@ def blockreplace(
         content='',
         append_if_not_found=False,
         prepend_if_not_found=False,
+        insert_before_match=None,
```
rallytime: Same comment here about kwarg order as above.

salt/modules/file.py
```diff
@@ -2494,6 +2496,17 @@ def blockreplace(path,
...
+
+    insert_after_match
+        If markers are not found, this parameter can be set to a regex which will
+        insert the block after the first found occurrence in the file.
+
+        .. versionadded:: Fluorine
```
rallytime: `Neon` :)

salt/states/file.py
```diff
@@ -4627,6 +4629,18 @@ def blockreplace(
...
+        If markers are not found, this parameter can be set to a regex which will
+        insert the block before the first found occurrence in the file.
+
+        .. versionadded:: Fluorine
+
+    insert_after_match
+        If markers are not found, this parameter can be set to a regex which will
+        insert the block after the first found occurrence in the file.
+
+        .. versionadded:: Fluorine
```
rallytime: `Neon`.

rallytime (dismissed):
Some small requests from me on this change.

salt/modules/file.py
```diff
@@ -2676,56 +2691,72 @@ def _add_content(linesep, lines=None, include_marker_start=True,
...
-                        None,
-                        perms['user'],
-                        perms['group'],
-                        perms['mode'])
+        elif insert_before_match or insert_after_match:
+            match_regex = insert_before_match or insert_after_match
+            if not isinstance(match_regex, six.string_types):
+                raise CommandExecutionError(
+                    'RegEx expected in match parameter.'
+                )
```
terminalmage: Why wait until this late to check for a regex? This should be done much earlier, when we are checking the arguments passed to the function. Additionally, the error message doesn't make sense as there is no `match` parameter. We should be telling the user _which_ parameter is invalid, instead of making them figure it out.

salt/modules/file.py
terminalmage: We have already passed our feature freeze for Fluorine, so these `versionadded` items will need to be updated to `Neon`.

salt/states/file.py
terminalmage: We have already passed our feature freeze for Fluorine, so these `versionadded` items will need to be updated to `Neon`.

terminalmage (changes requested):

nicholasmhughes:
@rallytime @terminalmage please check the latest changes from review comments

terminalmage (approved):

cachedout:
@nicholasmhughes CodeClimate has one suggestion here. It's not required that you fix this but I'd like you to please take a look and see what you think:

https://codeclimate.com/github/saltstack/salt/pull/48989

nicholasmhughes:
@cachedout , I'm not sure how to reduce the nesting without playing around at re-writing a lot of the preceding code... which I'd rather not take on right now due to lack of time. I'm guessing this is one level deeper than the desired threshold. Do you think it will adversely affect operation or readability to the point that it requires changing?

PR #48990, opened 2018-08-08
modules/nilrt_ip.py: Fix static configuration

alexvasiu:
* `modules/nilrt_ip.py: Fix static configuration`

If request mode is static, then all settings will be in INI_FILE.

* `integration/test_nilrt_ip.py: Fix test_static_all`

On older nilrt, just one static dns is supported.

### DIFF
--- a/salt/modules/nilrt_ip.py
+++ b/salt/modules/nilrt_ip.py
@@ -129,16 +129,16 @@ def _space_delimited_list(value):
     '''
     validate that a value contains one or more space-delimited values
     '''
-    if isinstance(value, str):
+    if isinstance(value, six.string_types):
         items = value.split(' ')
         valid = items and all(items)
     else:
         valid = hasattr(value, '__iter__') and (value != [])
 
     if valid:
-        return (True, 'space-delimited string')
+        return True, 'space-delimited string'
     else:
-        return (False, '{0} is not a valid list.\n'.format(value))
+        return False, '{0} is not a valid list.\n'.format(value)
 
 
 def _validate_ipv4(value):
@@ -373,9 +373,9 @@ def _get_static_info(interface):
     return data
 
 
-def _get_interface_info(interface):
+def _get_base_interface_info(interface):
     '''
-    return details about given interface
+    return base details about given interface
     '''
     blacklist = {
         'tcpip': {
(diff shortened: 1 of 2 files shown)


rares-pop:
How did we catch this?

salt/modules/nilrt_ip.py
```diff
@@ -417,17 +417,16 @@ def _get_interface_info(interface):
...
-        data['ipv4']['netmask'] = interface.sockaddrToStr(interface.netmask)
-        data['ipv4']['gateway'] = '0.0.0.0'
-        data['ipv4']['dns'] = _get_dns_info()
-    elif data['ipv4']['requestmode'] == 'static':
+    if data['ipv4']['requestmode'] == 'static':
         data['ipv4']['address'] = settings['IP_Address']
         data['ipv4']['netmask'] = settings['Subnet_Mask']
         data['ipv4']['gateway'] = settings['Gateway']
         data['ipv4']['dns'] = [settings['DNS_Address']]
+    elif data['up']:
```
rares-pop: If this is 'up', we shouldn't _load_config at all, right?

alexvasiu: If the interface uses a **static** IP, then configs should be loaded using `_load_config`, no matter if the interface is up or not. If it doesn't use a **static** IP and it's up, then configs will be loaded using `pyiface` and `/etc/resolv.conf` for **DNS**

rares-pop: On this 'elif data['up']' you don't use the 'settings' at all. I was asking if you can move this before doing doing the _load_config().

I would also check for 'ethercat' adapter_mode before all this tcp-related logic and handle that and return.

alexvasiu: That was the problem before, that I checked data['up'] first ... So, it the adapter_mode is `ethercat`, data['up'] will be `False` and data['ipv4'] may contain other information, but since we know that's ethercat, that ipv4 information is not relevant and we ignore it.

rares-pop: OK, so the code looks right now, but I would refactor it to avoid confusion and the bugs that come out of it in the future, in a manner to have worker methods for 'disabled', 'ethercat', 'tcp' modes. I would have a get_base_interface_info() to return the common information for those cases to avoid duplicating the code. That way, it's going to be more maintainable in the future.

salt/modules/nilrt_ip.py
```diff
@@ -129,16 +129,16 @@ def _space_delimited_list(value):
     '''
     validate that a value contains one or more space-delimited values
     '''
-    if isinstance(value, str):
+    if isinstance(value, str) or isinstance(value, type(u'')):
```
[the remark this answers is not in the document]
alexvasiu: Works only in `Python 2`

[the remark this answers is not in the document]
rares-pop: what about replace all that with: if type(a) == six.text_type?

rares-pop (approved):
Looks good to me!

salt/modules/nilrt_ip.py
```diff
@@ -129,16 +129,16 @@ def _space_delimited_list(value):
     '''
     validate that a value contains one or more space-delimited values
     '''
-    if isinstance(value, str):
```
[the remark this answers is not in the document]
alexvasiu: Yes, it works. But I found a problem now in python2:
```python
a = 'abc'
type(a) == six.text_type # or isinstance(a, six.text_type) -> False
a = u'abc'
type(a) == six.text_type # or isinstance(a, six.text_type) -> True
```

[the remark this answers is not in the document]
rallytime: I think you should be using `six.string_types`:
```
>>> import salt.ext.six as six
>>> a = 'abc'
>>> isinstance(a, six.string_types)
True
>>> a = u'abc'
>>> isinstance(a, six.string_types)
True
```

rallytime (approved):

Issue #48661, opened 2018-07-18
win_lgpo: User Rights Assignment policies


mike2523:
When making changes or changes required, always returns true. But no change happened. User rights assignments exists in `Computer Configuration->Windows Settings->Security Settings->Local Policies->User Rights Assignent`

Before: (using lgpo.get machine)
```
Backup files and directories:
     - BUILTIN\Backup Operators
     - BUILTIN\Administrators
```
in SLS file:
```
user_rights_assignments:
  lgpo.set:
    - computer_policy:
        "Backup files and directories":
            - BUILTIN\Administrators
```
output:
```
----------
          ID: user_rights_assignments
    Function: lgpo.set
      Result: True
     Comment:
     Started: 13:20:04.467000
    Duration: 38200.0 ms
     Changes:

Summary for local
------------
Succeeded: 1
Failed:    0
------------
Total states run:     1
Total run time:  38.200 s
```
Result is same as before using lgpo.get
```
        Backup files and directories:
            - BUILTIN\Backup Operators
            - BUILTIN\Administrators
```

secedit /export looks the same as before:
```
SeBackupPrivilege = *S-1-5-32-544,*S-1-5-32-551
```
If an item is already in the desired state, STDOUT says the proper comment field:
```
local:
----------
          ID: user_rights_assignments
    Function: lgpo.set
      Result: True
     Comment: "Take ownership of files and other objects" is already set."Profile single process" is already set.
     Started: 11:31:31.307000
    Duration: 44233.0 ms
     Changes:

Summary for local
------------
```

```
Salt Version:
           Salt: 2018.3.2

Dependency Versions:
           cffi: 1.10.0
       cherrypy: 10.2.1
       dateutil: 2.6.1
      docker-py: Not Installed
          gitdb: 2.0.3
      gitpython: 2.1.3
          ioflo: Not Installed
         Jinja2: 2.9.6
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: 1.0.6
   msgpack-pure: Not Installed
 msgpack-python: 0.4.8
   mysql-python: Not Installed
      pycparser: 2.17
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: Not Installed
         Python: 2.7.14 (v2.7.14:84471935ed, Sep 16 2017, 20:25:58) [MSC v.1500 64 bit (AMD64)]
   python-gnupg: 0.4.1
         PyYAML: 3.12
[...]

PR #49025, opened 2018-08-08
Fix several issues with LGPO

twangboy:
### What does this PR do?
Fixes an issue with the state where the comments field wasn't being commented properly

Fixes an issue in the state module where it was detecting that it needed to change a policy that had a list of users. The two lists were the same but in differing orders.

Fixes an issue with the module where it was stack tracing when setting "Devices: Allowed to format..." to Not Defined

https://github.com/saltstack/salt/issues/48661

### Tests written?
No

### Commits signed with GPG?
Yes


cachedout:
@dwoz Can you review this today?

dwoz (approved):
Looks good, I'm going to re-run the py2 tests.

twangboy:
@lomeroe Would you mind taking a look at this?

PR #49031, opened 2018-08-09
states.archive add parameter named skip_same_file

rongzeng54:
### What does this PR do?
Used for extracting archives as needed.

Sometimes, replacement files will affect online services.
So we need a parameter to control if skip same files when extracting archives.

### Previous Behavior
Extract the archives to the directory, regardless of whether the file has changed.

### New Behavior
Firstly extract the archives to temporary directory, then only copy the changed file to the specified directory, finally delete the temporary directory.

### Tests written?
No

### Commits signed with GPG?
No


salt/states/archive.py
```diff
@@ -1387,6 +1443,68 @@ def extracted(name,
             ret['comment'] = exc.strerror
             return ret
 
+    if skip_same_file:
+        extracted_files = []
+        log.debug('Extracting %s to %s', name, origin_name)
+        mng_files, mng_dirs, mng_links = _file_lists(name, origin_name)
+        for dest_dir, src_dir in mng_dirs:
+            try:
```
cachedout: This looks like a block of code that could be refactored for DRY in this state.

salt/states/archive.py
```diff
@@ -1387,6 +1443,68 @@ def extracted(name,
...
+            except Exception as exc:
+                __salt__['file.remove'](name)
+                ret['changes'] = {}
+                log.debug(traceback.format_exc())
+                return _error(ret, 'Unable to manage dir: {0}'.format(exc))
+
+        for dest_file, src_file in mng_files:
+            try:
+                stats = __salt__['file.stats'](src_file)
+                ret = __salt__['file.manage_file'](
```
terminalmage: This will override the `ret` dict for the state with the results of `file.manage_file`. This is a _bad_ idea.

terminalmage (changes requested):
This is not commented at all, which needs to change if we are going to merge this.

If the goal is to only replace files which changed, then using `file.get_managed` seems like a _lot_ of extra overhead. Instead, why not just compare the checksums of the two files?

This also needs tests to be written for it if we're going to consider merging.

rongzeng54:
HI @terminalmage 
I use Salt in a CDN company. There are many lua files need to distributed to many servers located around the world.
Current we use file.recurse and its too slow, so we want use archives stored on CDN based on HTTP(s).

salt/states/archive.py
rongzeng54: In this scene, the main task is extracting archives, other tasks like manage_file dont need return too much information. Maybe fast failure is recommended.

cachedout:
@rongzeng54 CodeClimate has a few suggestions here. It is not required that you fix them but I'd at least like you to please take a look and see if any of them make sense to address:

https://codeclimate.com/github/saltstack/salt/pull/49031

salt/states/archive.py
```diff
@@ -1239,7 +1259,19 @@ def extracted(name,
             __states__['file.directory'](name, user=user, makedirs=True)
             created_destdir = True
 
-        log.debug('Extracting %s to %s', cached, name)
+        if skip_same_file:
+            origin_name = name
+            tmp_filename = salt.utils.hashutils.md5_digest(os.path.basename(cached))
+            name = os.path.join(os.path.dirname(cached), tmp_filename)
```
terminalmage: We should not be changing the value of `name` in a state. It's just not a good idea. The `name` parameter is something fundamental to the state being run. If we need to operate on a different path, then we should be assigning that alternate path to a different variable name, not overwriting `name`.

salt/states/archive.py
```diff
@@ -1387,6 +1419,104 @@ def extracted(name,
...
+                    __states__['file.directory'](
+                        dest_dir,
+                        user=stats.get('user'),
+                        group=stats.get('group'),
+                        file_mode=stats.get('mode')
+                    )
+            except Exception as exc:
+                __salt__['file.remove'](name)
+                ret['changes'] = {}
+                log.exception(traceback.format_exc())
```
terminalmage: This is redundant. `log.exception()` will already append the traceback to the log message, and log at the `error` loglevel. The proper use of `log.exception()` would be to log an error message.

salt/states/archive.py
terminalmage: This still hasn't been changed, and it _needs_ to be changed. Since `ret` holds the state's return dictionary, replacing it with the result of a remote execution function will cause errors in the state compiler if we then attempt to do a `return ret`.

salt/states/archive.py
```diff
@@ -1387,6 +1419,104 @@ def extracted(name,
...
+                    attrs=None,
+                    saltenv=__env__,
+                    backup=None,
+                    makedirs=True,
+                    follow_symlinks=True,
+                    show_changes=True)
+            except Exception as exc:
+                __salt__['file.remove'](name)
+                ret['changes'] = {}
+                log.exception(traceback.format_exc())
```
terminalmage: Again, this is redundant, since `log.exception()` will append the traceback to the log message.

salt/states/archive.py
```diff
@@ -1387,6 +1419,104 @@ def extracted(name,
...
+
+        for link_dest, link_src in mng_links:
+            try:
+                if os.path.exists(link_dest):
+                    __salt__['file.remove'](link_dest)
+                __salt__['file.symlink'](link_src, link_dest)
+            except Exception as exc:
+                __salt__['file.remove'](name)
+                ret['changes'] = {}
+                log.exception(traceback.format_exc())
```
terminalmage: Also redundant.

salt/states/archive.py
```diff
@@ -1387,6 +1419,104 @@ def extracted(name,
             ret['comment'] = exc.strerror
             return ret
 
+    def _file_lists(src_path, dest_path):
```
terminalmage: This seems like it was pulled wholesale from the `roots` fileserver. I don't like the idea of duplicating code like this. Assuming that this approach is necessary, we should create a shared function in `salt/utils/files.py` that both salt/fileserver/roots.py and this function can both use.

salt/states/archive.py
terminalmage: I also don't understand why this is using `file.manage_file`, which hasn't been explained since I asked a few weeks ago. This is a _lot_ of extra overhead when all we care about is whether or not the file is different. Simply comparing the checksums and replacing only when the checksum differs, should be sufficient. Perhaps an extra check to see if the paths are links.

terminalmage (changes requested):

salt/states/archive.py
rongzeng54: It seems that they have much more differences

cachedout:

rallytime:
@rongzeng54 Just checking in here. Did you see the feedback above?

terminalmage (commented):
We still need to see some integration tests here, but otherwise I think this is OK.

salt/states/archive.py
```diff
@@ -1239,153 +1440,205 @@ def extracted(name,
...
-                            return ret
-                else:
-                    if not salt.utils.path.which('tar'):
-                        ret['comment'] = (
-                            'tar command not available, it might not be '
-                            'installed on minion'
-                        )
-                        return ret
+        if skip_same_file:
+            tmp_filename = salt.utils.hashutils.md5_digest(os.path.basename(cached))
```
thatch45: We should never explicitly use md5, mainly because it is not enabled in all openssl libs so this will fail on many installs.

thatch45:
@rongzeng54 I really like what you have here but without integration tests and since you have gone dark for a long time I am going to to ahead and close this out. When you have time to circle back with this addition and can add some tests I would be happy to get it merged in! But for now I am going to close this out. Thanks!

PR #49036, opened 2018-08-09
minor doc fix in netyang.py

cspeidel:
### What does this PR do?
Fix a minor doc issue.

A minor doc issue. 

### Tests written?
Not needed

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/states/netyang.py
+++ b/salt/states/netyang.py
@@ -79,7 +79,7 @@ def managed(name,
             models,
             **kwargs):
     '''
-    Manage the device configuration given the input data strucuted
+    Manage the device configuration given the input data structured
     according to the YANG models.
 
     data


rallytime (approved):
Thanks!

FYI, the man pages are auto-generated from the docs, so you don't need to worry about fixing them. It doesn't hurt, but those changes will be overwritten on the next release with whatever is in the docs. :)

cspeidel:
@rallytime Thanks! Do I have to make some kind of request for the merge-forward process/or include a note when doing a back-ported request? Or do you do merge forwards every so often?

rallytime:
I do merge forwards every couple of days, so it will make it through to develop eventually. :)

PR #49037, opened 2018-08-09
Fix case where args or kwargs are not specified in loop state

mtlaird:

### What does this PR do?
Fixes an error with the loop state when the definition did not specify values for the m_args or m_kwargs parameters.

See above. I couldn't find an open issue for this, but it's an easy fix.

### Previous Behavior
If those values were not specified, the state would fail and log a stack trace.

### New Behavior
If those those values are not specified, the state will succeed without them.

### Tests written?

No

### Commits signed with GPG?

No

### DIFF
--- a/salt/states/loop.py
+++ b/salt/states/loop.py
@@ -94,7 +94,11 @@ def until(name,
         ret['comment'] = 'The execution module {0} will be run'.format(name)
         ret['result'] = None
         return ret
-
+    if not m_args:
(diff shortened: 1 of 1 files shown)


cachedout (approved):

rallytime (approved):

rallytime:
Welcome @mtlaird! Thank you for submitting this fix.

Issue #43164, opened 2017-08-24
Windows registry (reg state) not working as documented


enigmapsi:
### Description of Issue/Question
The following state **works**:

```
HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Services\Eventlog\Security:
      reg.present:
        - vname: retention
        - vdata: 0
        - vtype: REG_DWORD
```
The following state **does not work** - the ID was moved to `- name`:
```
reg_security_retention:
      reg.present:
        - name: 'HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Services\Eventlog\Security'
        - vname: retention
        - vdata: 0
        - vtype: REG_DWORD
```
It returns the following error:
```
[...]
MINDEV1:
----------
          ID: reg_security_retention
    Function: reg.present
        Name: HKEY_LOCAL_MACHINE\\SYSTEM\\CurrentControlSet\\Services\\Eventlog\\Security
      Result: False
     Comment: Failed to add HKEY_LOCAL_MACHINE\\SYSTEM\\CurrentControlSet\\Services\\Eventlog\\Security to HKEY_LOCAL_MACHINE\\SYSTEM\\CurrentControlSet\\Services\\Eventlog\\Security
     Started: 08:49:34.708000
    Duration: 0.0 ms
     Changes:

Summary for MINDEV1
------------
Succeeded: 0
Failed:    1
------------

```

It looks to me that Salt tries to escape the backslash (`\`) from the registry path even if it shouldn't (e.g. when the path is enclosed in single quotes). The same behaviour exists if I enclose it in double quotes and escape each backslash, like:

```
reg_security_retention:
      reg.present:
        - name: "HKEY_LOCAL_MACHINE\\SYSTEM\\CurrentControlSet\\Services\\Eventlog\\Security"
        - vname: retention
        - vdata: 0
        - vtype: REG_DWORD
```

The trace log below shows the same - backslashes escaped twice (?)

### Trace log
```
If I run salt with the debug log level (`-l trace`), the following is returned:
[...]

PR #49045, opened 2018-08-09
Fix docs for the registry module and state

twangboy:
### What does this PR do?
Addresses discrepancies about when you should double, single, and not quote the hive/key parameters

https://github.com/saltstack/salt/issues/43164

### Tests written?
No

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/modules/reg.py
+++ b/salt/modules/reg.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-'''
+r'''
 ===========================
 Manage the Windows registry
 ===========================
@@ -17,6 +17,17 @@ Keys
 ----
 Keys are the folders in the registry. Keys can have many nested subkeys. Keys
 can have a value assigned to them under the (Default)
+
+When passing key on the CLI it must be quoted correctly depending on the
+backslashes being used (``\`` vs ``\\``). The following are valid methods of
+passing the the key on the CLI:
+
+Using single backslashes:
+    ``"SOFTWARE\Python"``
+    ``'SOFTWARE\Python' (will not work on a Windows Master)``
+
+Using double backslashes:
+    ``SOFTWARE\\Python``
 
 -----------------
 Values or Entries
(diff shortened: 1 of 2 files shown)


salt/modules/reg.py
```diff
@@ -18,6 +18,17 @@
 Keys are the folders in the registry. Keys can have many nested subkeys. Keys
 can have a value assigned to them under the (Default)
 
+When passing key on the CLI it must be quoted correctly depending on the
```
rallytime: `When passing a key`

salt/modules/reg.py
```diff
@@ -18,6 +18,17 @@
...
+backslashes being used (``\`` vs ``\\``). The following are valid methods of
+passing the the key on the CLI:
+
+Using single backslashes:
+    ``"SOFTWARE\Python"``
+    ``'SOFTWARE\Python' (will not work on a Windows Master)``
```
rallytime: I think it would make more sense to move the parentheses part outside of the backticks.:
``'SOFWARE\Python'`` (will not work on a Windows Master)

rallytime (changes requested):
I have some small requests, otherwise 👍 

There are some lint errors that need to be fixed, too: https://jenkinsci.saltstack.com/job/pr-lint/job/PR-49045/1/warnings52Result/

salt/modules/reg.py
twangboy: Yeah, I'll fix that

salt/modules/reg.py
twangboy: OK

twangboy:
Getting a 502 error on the lint link. Do you happen to know what the issue was?

salt/modules/reg.py
```diff
@@ -18,6 +18,17 @@
 Keys are the folders in the registry. Keys can have many nested subkeys. Keys
 can have a value assigned to them under the (Default)
 
+When passing a key on the CLI it must be quoted correctly depending on the
+backslashes being used (``\`` vs ``\\``). The following are valid methods of
+passing the the key on the CLI:
+
+Using single backslashes:
+    ``"SOFTWARE\Python"``
+    ``'SOFTWARE\Python'`` (will not work on a Windows Master)
```
cachedout: Huh? Why won't it work on a master?

twangboy: The cmd line on Windows doesn't like single quotes.

rallytime (approved):

Issue #49016, opened 2018-08-08
add mine.get tests


Ch3LL:
Add some mine.get tests. This is related to this PR: https://github.com/saltstack/salt/pull/48765

rallytime:
Let's get these added to the 2018.3 branch as well.

PR #49056, opened 2018-08-10
Add orch runner with mine.get integration tests

Ch3LL:
### What does this PR do?
adds state.orchestration test with `mine.get` call in the sls file.

fixes https://github.com/saltstack/salt/issues/49016

### DIFF
--- a/tests/integration/runners/test_state.py
+++ b/tests/integration/runners/test_state.py
@@ -97,10 +97,18 @@ class StateRunnerTest(ShellCase):
         '''
         test salt-run state.orchestrate with mine.get call in sls
         '''
-        test = self.run_run('mine.update "*"')
-        ret = self.run_run('state.orchestrate orch.mine')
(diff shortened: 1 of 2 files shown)


rallytime (approved):

tests/integration/runners/test_state.py
```diff
@@ -93,6 +93,15 @@ def test_orchestrate_nested(self):
         assert os.path.exists('/tmp/ewu-2016-12-13') is False
         assert code != 0
 
+    def test_orchestrate_with_mine(self):
+        '''
+        test salt-run state.orchestrate with mine.get call in sls
+        '''
+        test = self.run_run('mine.update "*"')
```
cachedout: Mine updates are async. This might race.

Ch3LL: okay looking into a solution to make sure this doesn't occur.

PR #49092, opened 2018-08-13
skip gpg tests if not installed

Ch3LL:
### What does this PR do?
The test `unit.modules.test_gpg.GpgTestCase.test_delete_key` is failing on macosx because gpg is not installed. 

I added this salt-jenkins state: https://github.com/saltstack/salt-jenkins/pull/1063 to install it but figured it would be good to add the skipif to make sure we skip if its not installed for other OS's.

### DIFF
--- a/tests/unit/modules/test_gpg.py
+++ b/tests/unit/modules/test_gpg.py
@@ -21,6 +21,7 @@ from tests.support.mock import (
 
 # Import Salt Libs
 import salt.modules.gpg as gpg
+import salt.utils.path
 
 
 try:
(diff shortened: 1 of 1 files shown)


rallytime (approved):

tests/unit/modules/test_gpg.py
```diff
@@ -30,6 +31,7 @@
     HAS_GPG = False
 
 
+@skipIf(not salt.utils.path.which('gpg'), 'gpg not installed skipping')
```
cachedout: Should be `GPG not installed. Skipping.` (Yes, I know I'm being OCD.)

cachedout (approved):

PR #49097, opened 2018-08-13
Use the minion's region if use_grain is enabled

carsonoid:
This makes it possible for the master to be in a different region
than the minion and still fetch ec2 pillars. This requires the
use_grain configuration to be set to True and will fall back to the
master's region if there is no region grain.

Credit to John Nielsen for the code.

Signed-off-by: Carson Anderson <ca@carsonoid.net>

### Previous Behavior

The salt-master had to be in the same region as all salt-minions

### New Behavior

The master can use minion metadata to determine which api to call

### Tests written?

No

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/pillar/ec2_pillar.py
+++ b/salt/pillar/ec2_pillar.py
@@ -184,6 +184,10 @@ def ext_pillar(minion_id,
     # Get the Master's instance info, primarily the region
     (_, region) = _get_instance_info()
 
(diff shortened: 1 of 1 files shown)


gtmanfred (approved):

jnielsendotnet:
Thanks @carsonoid !

rallytime (approved):

PR #49121, opened 2018-08-14
Add pathstate to keepalive check in mac_service

Ch3LL:
### What does this PR do?
The following tests:

```
integration.states.test_service.ServiceTest.test_service_dead
integration.states.test_service.ServiceTest.test_service_dead_init_delay
integration.states.test_service.ServiceTest.test_service_running
```

are failing on macosx after this PR: https://github.com/saltstack/salt/pull/46206

The service we use in our test suite sets keepalive with PathState setting. From the launchd docs there are also other ways to set keepalive besides just setting it to be True. This PR only handles PathState so more cases will need to be added. I moved the `_always_running_service` check after we check for pids to try to retain backwards compatibility. If the service actually returns a pid we return that and not loaded if the check for KeepAlive is not added to the code yet.

### DIFF
--- a/salt/modules/mac_service.py
+++ b/salt/modules/mac_service.py
@@ -125,6 +125,15 @@ def _always_running_service(name):
         return False
 
     # check if KeepAlive is True and not just set.
+
+    if isinstance(keep_alive, dict):
+        # check for pathstate
+        for _file, value in six.iteritems(keep_alive.get('PathState', {})):
+            if value is True and os.path.exists(_file):
+                return True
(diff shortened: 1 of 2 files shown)


Ch3LL:
ping @weswhet tried to add you as a reviewer but doesn't seem i can. if you could give this a review :)

rallytime (approved):

weswhet:
This looks good to me. For more info on keepalive services checkout http://www.launchd.info > Configurations > KeepAlive

Issue #49100, opened 2018-08-13
salt.utils.yamldumper.dump/safe_dump incorrectly return 'str' instead of unicode on python2.7 (centos)


pruiz:
### Description of Issue/Question

After upgrading to 2018.3.2 (from 2017.7.2), I've found that when using pillar data (with non-ascii characters) while rendering yaml sls files, it fails with: UnicodeDecodeError: 'ascii' codec can't decode byte...

AFAIK it is due to PyYAML, as invoked at salt/utils/yamldumper.py:safe_dump, returning an 'str' object, which later on is being coerced into unicode as if it where an ascii-based string.

IMHO, this can be fixed by invoking salt.utils.stringutils.to_unicode() to yaml data before returning. I'll try to submit a pull-req next.

### Setup

First you will need some misc data at pillar (in my case git-based ext_pillar), like:

/tmp/k2.sls

```
data:
  question: Quieres Café?
```

Then a sample state like this one:

```
{%- set settings = salt['pillar.get']('herma:renderer:settings') %}

/tmp/k.settings:
  file.managed:
    - encoding: UTF-8
    - contents: |
      {{ settings|yaml }}
```

### Steps to Reproduce Issue

```
# salt-call  -l debug state.template /tmp/k2.sls
[...]
local:
    Data failed to compile:
----------
    Rendering SLS ':/tmp/k2.sls' failed: Jinja error: 'ascii' codec can't decode byte 0xc3 in position 398: ordinal not in range(128)
Traceback (most recent call last):
  File "/usr/lib/python2.7/site-packages/salt/utils/templates.py", line 390, in render_jinja_tmpl
    output = template.render(**decoded_context)
  File "/usr/lib/python2.7/site-packages/jinja2/environment.py", line 989, in render
    return self.environment.handle_exception(exc_info, True)
  File "/usr/lib/python2.7/site-packages/jinja2/environment.py", line 754, in handle_exception
    reraise(exc_type, exc_value, tb)
  File "<template>", line 10, in top-level template code
  File "/usr/lib/python2.7/site-packages/salt/utils/jinja.py", line 805, in format_yaml
    if yaml_txt.endswith('\n...'):
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 398: ordinal not in range(128)

; line 10

---
[...]
/tmp/k.settings:
  file.managed:
    - template: jinja
    - encoding: UTF-8
    - contents: |
      {{ settings|yaml }}    <======================

[...]

PR #49122, opened 2018-08-14
Fix for #49100: Avoid yamldumper's dump/safe_dump from returning a non-unicode object under python2.7, as this breaks jinja/yaml rendering.

pruiz:
### What does this PR do?

This change ensures yamldumper's dump/safe_dump methods, do not return non-unicode objects on python 2.7.

Hopefully this applies cleanly to salt 2018.3.3+.. as this is a reversion of salt 2018.3, previous versions (at least 2017.7.*) did not have this issue (when using the now deprecated yaml_utf8 setting)

This is a proposed fix for issue #49100. 

### Tests written?

No

### Commits signed with GPG?

No

### DIFF
--- a/salt/utils/yamldumper.py
+++ b/salt/utils/yamldumper.py
@@ -18,6 +18,8 @@ except ImportError:
 import yaml  # pylint: disable=blacklisted-import
 import collections
 
+import salt.utils.stringutils
+import salt.ext.six as six
 from salt.utils.odict import OrderedDict
 
 try:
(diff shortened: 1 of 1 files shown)


terminalmage (commented):
This is not the correct fix for this. A dumped value should by definition be a `str` type on Python 2, as serialized data should be directly able to be written to a filehandle. This would break anything that tries to write dumped content to a file, if it contains non-ascii unicode.

I've submitted a better fix [here](https://github.com/saltstack/salt/pull/49125), which also fixes an additional issue that affects both YAML and JSON filters.

terminalmage:
https://github.com/saltstack/salt/pull/49125 was merged. Closing this as redundant.

PR #49126, opened 2018-08-14
TLS cert_info fixes

The-Loeki:
### What does this PR do?
Fix multiple small problems with `tls.cert_info`
* Fix output of `extensions` keys  being bytes (see also  #49123)
* fix outputting errors for IP Addres subjectAltNames
* allow cert string as well as file path as input


salt/modules/tls.py
```diff
@@ -1575,12 +1575,15 @@ def create_pkcs12(ca_name, CN, passphrase='', cacert_path=None, replace=False):
     )
 
 
-def cert_info(cert_path, digest='sha256'):
+def cert_info(cert, digest='sha256'):
```
cachedout: Why are we changing this?

The-Loeki: `cert` is string|filepath
`cert_path` being a string would be rather confusing

Also it's the first arg which is nowhere used in Salt AFAICT, so pplz using `tls.get_cert_info(myfile)` wont notice a thing.

I can add some deprecation stuff around it if you're bothered by it?

cachedout (approved):

PR #49127, opened 2018-08-14
[develop] Merge forward from fluorine to develop

terminalmage:
This merge-forward resolves some loader unit test failures in develop.

**(Correction: I had cherrypicked some commits into https://github.com/saltstack/salt/pull/47337 which added new tests, for which the fix had not yet merged forward past `fluorine`. So, this really only affects https://github.com/saltstack/salt/pull/47337.**


cachedout:
@KaiSforza What's going on with these test runs in Jenkins?

cachedout (approved):

PR #49131, opened 2018-08-15
allow for using multiple on* requisites on the same state

### LINKED ISSUE #49120: Issue with Compound Requisites
Compound requisites in states are supported, however, the following use case shows unexpected behavior.

Per @gtmanfred :
"It goes through the for loop, sees that the state has an `onfail` check, and then does a `continue` which then jumps back to the next item in the for loop, and never runs that if check for `onchanges`."

Reference:
https://github.com/saltstack/salt/blob/2018.3/salt/state.py#L2342

### Setup and Steps to Reproduce Issue
Run this state and observe the behavior when changes should occur.  The observer will note that the onchanges requisite is ignored, in favor of onfail not having been triggered.
```
update_file:
  file.managed:
    - name: /tmp/foo
    - contents:
      - 'Test'

test_state:
  test.configurable_test_state:
    - name: Compound requisite test
    - changes: True
    - result: False
    - comment: Something changed or failed on {{ grains.id }}
    - onfail:
      - file: update_file
    - onchanges:
      - file: update_file
```

### Versions Report
Issue observed on 2017.7.7

gtmanfred:
### What does this PR do?
onchanges and onfail in this case

Fixes #49120

### Previous Behavior
If onchanges and onfail were both set, if the onfail state did not fail but had changes, checking onchanges was skipped.

### New Behavior
If onfail and onchanges are both specified, and either of them are `met` it will run the state.

### Tests written?

Yes

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/state.py
+++ b/salt/state.py
@@ -2391,7 +2391,7 @@ class State(object):
             status = 'onchanges'
         elif 'change' in fun_stats:
             status = 'change'
-        else:
+        if any(stat.endswith('met') for stat in fun_stats):
             status = 'met'
 
         return status, reqs
--- a/tests/unit/test_state.py
+++ b/tests/unit/test_state.py
@@ -18,7 +18,7 @@ from tests.support.mock import (
     MagicMock,
     patch)
 from tests.support.mixins import AdaptedConfigurationTestCaseMixin
-from tests.support.paths import BASE_FILES
+from tests.support.paths import BASE_FILES, TMP
 
 # Import Salt libs
 import salt.exceptions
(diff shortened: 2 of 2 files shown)


cachedout (changes requested):
Some lint errors snuck in here.

gtmanfred:
There are also some tests that need to be fixed.

thatch45 (approved):
This is the perfect fix for this

cachedout (approved):

gtmanfred (commented):
The problem is that now, the other requisites are going to behave differently because `met` has been returned, which is causing a bunch of tests to fail.

https://github.com/saltstack/salt/blob/2018.3/salt/state.py#L2462-L2662

I can change these tests so that they pass.  But I think this is actually going to require reworking how requisites resolution is handled in the develop branch.

thoughts @cachedout @thatch45 @saltstack/team-core ?

gtmanfred:
So, basically we have a whole set of if, elif, elif, else trees, and in order to handle this situation better, we need to change it to a series of if statements, so that more then one requisite can be on the same state chunk.

gtmanfred:
It can be closed.  I updated the issue, it will need to be reworked how we handle counting the different requisites so that more than one can be applied at the same time.

Issue #46942, opened 2018-04-06
Salt runners broken with master job cache and postgresql returner


breshead:
### Description of Issue/Question
When the master job cache is setup using postgres, calling salt-run <anything> will throw a 'TypeError' in the postgres_local_cache.py module.

My use case is `salt-run manage.[not_]present` but I tried 
```
    salt-run manage.list_state
    salt-run manage.alived
    salt-run manage.allowed
    salt-run manage.joined
    salt-run doc.execution
    salt-run saltutil.sync_all
    salt-run salt.cmd test.ping
```
To see if it was specific or not. It appears to be all runners.

When I disable the master job cache (remark out in the config file), the salt runner completes successfully.

### Setup
/etc/salt/master.d/master_job_cache.conf
```
    master_job_cache: postgres_local_cache
    event_return: postgres_local_cache
    master_job_cache.postgres.host: '192.168.8.14'
    master_job_cache.postgres.user: 'myuser'
    master_job_cache.postgres.passwd: 'mypass'
    master_job_cache.postgres.db: 'ejcache'
    master_job_cache.postgres.port: 5432
```

### Steps to Reproduce Issue
With master job cache setup..
On the command salt-master CLI run..

~# salt-run manage.list_state
---- RESULT ----
```
[ERROR   ] An un-handled exception was caught by salt's global exception handler:
TypeError: decoding Unicode is not supported
Traceback (most recent call last):
  File "/usr/bin/salt-run", line 10, in <module>
    salt_run()
  File "/usr/lib/python2.7/dist-packages/salt/scripts.py", line 412, in salt_run
    client.run()
  File "/usr/lib/python2.7/dist-packages/salt/cli/run.py", line 40, in run
    ret = runner.run()
  File "/usr/lib/python2.7/dist-packages/salt/runner.py", line 271, in run
    daemonize=False)
  File "/usr/lib/python2.7/dist-packages/salt/client/mixins.py", line 485, in _proc_function
    return self.low(fun, low, full_return=False)
  File "/usr/lib/python2.7/dist-packages/salt/client/mixins.py", line 261, in low
    return self._low(fun, low, print_event=print_event, full_return=full_return)
  File "/usr/lib/python2.7/dist-packages/salt/client/mixins.py", line 418, in _low
    mminion=self.mminion,
  File "/usr/lib/python2.7/dist-packages/salt/utils/job.py", line 110, in store_job
    mminion.returners[fstr](load)
[...]

PR #49133, opened 2018-08-15
Remove duplicate unicode encoding for bug #46942

breshead:
### What does this PR do?
Modifies postgres_local_cache.py to allow postgres returner to function.

#46942

### Previous Behavior

### Tests written?
No

### Commits signed with GPG?
No

[postgres_local_cache.zip](https://github.com/saltstack/salt/files/2290976/postgres_local_cache.zip)


salt/returners/postgres_local_cache.py
```diff
@@ -232,7 +232,7 @@ def returner(load):
     sql = '''INSERT INTO salt_returns
             (fun, jid, return, id, success)
             VALUES (%s, %s, %s, %s, %s)'''
-    job_ret = {'return': six.text_type(six.text_type(load['return']), 'utf-8', 'replace')}
+    job_ret = {'return': six.text_type(load['return'])}
```
terminalmage: This would break if the payload's return data contained encoded non-ascii unicode:

```python
>>> six.text_type('glücklich')
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 2: ordinal not in range(128)
```

Also, if for some reason the function returned non-decodable bytes (such as this example here where I'm reading from `/dev/urandom`), then we'd get a similar error:

```python
>>> with open('/dev/urandom', 'rb') as f:
...     data = f.read(16)
...
>>> six.text_type(data)
UnicodeDecodeError: 'ascii' codec can't decode byte 0xfa in position 0: ordinal not in range(128)
>>>
```

Therefore, how about something like this?

```python
try:
    ret = six.text_type(load['return'])
except UnicodeDecodeError:
    ret = str(load['return'])
job_ret = {'return': ret}
```

terminalmage (changes requested):

salt/returners/postgres_local_cache.py
breshead: So convert to unicode unless it will result in an error. In that case return the raw value. That makes sense to me.

terminalmage (approved):

breshead:
This commit seems to be not making it into any of the 'release' versions. I see it in develop but 2018.3.3, 3.4, 2019.2.0, 2.1 do not have it included.
How long do I have to keep my local fixes around before this makes it in?

terminalmage:
This pull request was never backported, and it was opened after the freeze for 2019.2. I just backported it in https://github.com/saltstack/salt/pull/53898.

salt/returners/postgres_local_cache.py
```diff
@@ -232,7 +232,11 @@ def returner(load):
     sql = '''INSERT INTO salt_returns
             (fun, jid, return, id, success)
             VALUES (%s, %s, %s, %s, %s)'''
-    job_ret = {'return': six.text_type(six.text_type(load['return']), 'utf-8', 'replace')}
+    try:
+        ret = six.text_type(load['return'])
+    except UnicodeDecodeError:
+        ret = str(load['return'])
```
waynew: Should this not use stringutils?

waynew (commented):
Did this have tests, or will we need to write those?

salt/returners/postgres_local_cache.py
terminalmage: @waynew This PR was merged 11 months ago. Maybe you meant to review the backport PR I opened this morning? I made the change to that PR.

waynew: I think GitHub must have gone sideways on me. But yes, I did 😂

PR #49146, opened 2018-08-16
Add FreeBSD support to pf executor

### LINKED ISSUE #33248: Execution module for FreeBSD Packet Filter

backerman:
### What does this PR do?
Enables FreeBSD support for the pf module; adds FreeBSD to OS whitelist and makes list of available loglevels OS-dependent.
This PR fixes #33248.
### Previous Behavior
Does not execute under FreeBSD.
### New Behavior
Will now execute on FreeBSD or OpenBSD.
### Tests written?
Yes
### Commits signed with GPG?
Yes

### DIFF
--- a/tests/unit/modules/test_pf.py
+++ b/tests/unit/modules/test_pf.py
@@ -66,14 +66,29 @@ class PfTestCase(TestCase, LoaderModuleMockMixin):
         with patch.dict(pf.__salt__, {'cmd.run_all': mock_cmd}):
             self.assertFalse(pf.disable()['changes'])
 
-    def test_loglevel(self):
+    def test_loglevel_freebsd(self):
         '''
         Tests setting a loglevel.
         '''
         ret = {}
         ret['retcode'] = 0
         mock_cmd = MagicMock(return_value=ret)
-        with patch.dict(pf.__salt__, {'cmd.run_all': mock_cmd}):
+        with patch.dict(pf.__salt__, {'cmd.run_all': mock_cmd}), \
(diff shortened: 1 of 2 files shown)


tests/unit/modules/test_pf.py
```diff
@@ -66,14 +66,29 @@ def test_disable_when_disabled(self):
         with patch.dict(pf.__salt__, {'cmd.run_all': mock_cmd}):
             self.assertFalse(pf.disable()['changes'])
 
-    def test_loglevel(self):
+    def test_loglevel_freebsd(self):
```
cachedout: Why are we changing an existing test to apply specifically to FreeBSD? This should be a separate test or this test should be extended to support both the existing and the new instead of modified like this.

backerman: The existing tests do it both possible ways; I was looking at e.g. `test/unit/modules/test_cron.py`, but there are also tests that have all the different OS variations in one test case (e.g. `./integration/states/test_pkg.py`). I'm not particularly attached to either, but I can definitely rewrite as one TC.

cachedout (approved):

s0undt3ch:
Port to master https://github.com/saltstack/salt/pull/56759

Issue #49152, opened 2018-08-16
problem with salt.wheel.error and raise_error


erwindon:
### Description of Issue/Question
When executing `wheel.error.error`, an unexpected internal error occurs.
The original purpose of this function is to raise an error. But it should be the specified error.

I believe the error is due to the code in file `salt/utils/error.py`.
There is exception handling around `import exceptions` which is triggered as under Python3 this package does not exist,
But on lines 28/29, the `exceptions` object is still used, this raising an unwanted exception.
When I comment these 2 lines, the function works as expected for a custom error.

Note that a similar situation is properly handled in `salt/_compat.py`.

### Setup
Default installation with just a salt-master and a salt-minion on the same machine.

### Steps to Reproduce Issue

```
$ salt HOSTNAME saltutil.wheel error.error "Exception" "This is an error."
HOSTNAME:
    ----------
    _stamp:
        2018-08-16T09:30:10.582247
    fun:
        wheel.error.error
    fun_args:
        - Exception
        - This is an error.
    jid:
        20180816113010567620
    return:
        Exception occurred in wheel error.error: Traceback (most recent call last):
          File "/usr/lib/python3.4/site-packages/salt/client/mixins.py", line 387, in _low
            data['return'] = self.functions[fun](*args, **kwargs)
          File "/usr/lib/python3.4/site-packages/salt/wheel/error.py", line 30, in error
            salt.utils.error.raise_error(name=name, message=message)
          File "/usr/lib/python3.4/site-packages/salt/utils/error.py", line 28, in raise_error
            elif hasattr(exceptions, name):
        NameError: name 'exceptions' is not defined
    success:
        False
    user:
        UNKNOWN
```

### Versions Report
Salt Version:
           Salt: 2018.3.2
 
Dependency Versions:
           cffi: Not Installed
       cherrypy: unknown
       dateutil: Not Installed
      docker-py: Not Installed
          gitdb: 2.0.3
      gitpython: 2.1.10
          ioflo: Not Installed
         Jinja2: 2.8
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.5.1
   mysql-python: 1.3.12
      pycparser: Not Installed
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: Not Installed
         Python: 3.4.8 (default, Mar 23 2018, 10:04:27)
[...]

PR #49162, opened 2018-08-16
Fixed unknown 'exceptions' under Python3 (#49152)

erwindon:
### What does this PR do?
This PR fixes the (wheel)error.error command, which fails under python3.
The solution is a small code transplant from file `_compat.py`.

This error was previously reported as #49152.

### Previous Behavior
Always exception about unknown identifier "exceptions"

### New Behavior
Behavior according to documentation.

### Tests written?
No
Tested using command `sudo salt HOSTNAME saltutil.wheel error.error "Exception" "This is an error."` after copying the updated file `salt/utils/error.py` into an installed system and restarting both the master and the minion of that system.
Manually verified that all `import exceptions` command are covered. I only found 2 cases: one in `_compat.py`, and this PR solves the issue for the other one (`salt/utils/error.py`).

### Commits signed with GPG?
No


salt/utils/error.py
```diff
@@ -5,11 +5,15 @@
...
 
-# Import python libs
-try:
+import sys
+# True if we are running on Python 3.
+PY3 = sys.version_info[0] == 3
+
+if PY3:
+    import builtins
+    exceptions = builtins
```
dwoz: This can be `from salt.ext.six.moves import builtins as exceptions`

dwoz (changes requested):

salt/utils/error.py
erwindon: @dwoz
Are you suggesting to just combine lines 13+14?
Then why not just `import builtins as exceptions`?
And should that also be applied to `_compat.py`?

dwoz: Using `salt.ext.six` also removes the need for the python 3 check. So 8-14 can become one line. Yes, the same can be done for lines 37-45 in `_compat.py`

erwindon: The code for `utils/error.py` is updated using this suggestion and re-tested in the same way as before.
I realize that changing `_compat.py` should not be done as part of this bugfix, as there is no bug there. So I did not update that one.

cachedout (approved):

erwindon:
@cachedout I was just thinking on how to force the regression test to re-run as older/failing jobs had finished later and github got confused about that... your merge action just solved that 👍

dwoz (approved):

PR #49167, opened 2018-08-17
Fix remaining file state integration tests (py3)

dwoz:
### What does this PR do?

Fix remaining file state integration tests (py3)

### Tests written?

No - Fixing tests

### Commits signed with GPG?

Yes

### DIFF
--- a/tests/integration/states/test_file.py
+++ b/tests/integration/states/test_file.py
@@ -267,7 +267,7 @@ class FileTest(ModuleCase, SaltReturnAssertsMixin):
 
         if IS_WINDOWS:
             expected = 'The \'mode\' option is not supported on Windows'
-            self.assertEqual(ret[ret.keys()[0]]['comment'], expected)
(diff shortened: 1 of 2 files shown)


gtmanfred (changes requested):
These should have been removed when we fixed all the other keys stuff,if you just pass a dict to list,it will give you back the keys

tests/integration/states/test_file.py
```diff
@@ -267,7 +267,7 @@ def test_managed_file_mode(self):
 
         if IS_WINDOWS:
             expected = 'The \'mode\' option is not supported on Windows'
-            self.assertEqual(ret[ret.keys()[0]]['comment'], expected)
+            self.assertEqual(ret[list(ret.keys())[0]]['comment'], expected)
```
gtmanfred: Just call list here,no reason for .keys

gtmanfred (approved):

Issue #24163, opened 2015-05-27
enable_gpu_grains default value confusion


tbaker57:
The docs says the default value of this config option is False
http://docs.saltstack.com/en/latest/ref/configuration/master.html#enable-gpu-grains
but the code says the default is True
https://github.com/saltstack/salt/blob/develop/salt/grains/core.py#L151

Which one should it be ?

As an aside, I'd love to be able to set enable_gpu_grains=False in salt-ssh to avoid #23342  using 'minion_opts' (or 'ssh_minion_opts' according to http://docs.saltstack.com/en/latest/ref/configuration/master.html#ssh-minion-opts ?) but I can't for the life of me get it to work (minion_opts seems to be ignored everywhere I stick it). Perhaps this is a separate issue...

cheers,
Tim

jtand:
@tbaker57 Thank you for pointing this out!

basepi:
@tbaker57 the `minion_opts` and `ssh_minion_opts` are only supported in 2015.5.1 and later. Sorry that I missed the versionadded there.

jacobhammons:
Updated the doc to match the code, http://docs.saltstack.com/en/latest/ref/configuration/master.html#enable-gpu-grains

@tbaker57 if you think this update is sufficient please close, thanks!

tbaker57:
@jacobhammons - looks great. Many thanks...

PR #49185, opened 2018-08-18
Update documentation to correctly state enable_gpu_grains default

ralish:
### What does this PR do?

While the relevant code in `grains/core.py` does enable GPU grains by default, the default in `config/__init__.py` for the **DEFAULT_MASTER_OPTS** dictionary is disabled. This takes precedence over the default in `grains/core.py` which would only take effect if the `enable_gpu_grains` setting was not specified at all.

This commit updates the documentation rather than change the defaults of Salt itself. Note that this commit effectively reverts https://github.com/saltstack/salt/issues/24163 which doesn't appear to have factored in the default set in **DEFAULT_MASTER_OPTS**.

None

### Tests written?

N/A

### Commits signed with GPG?

Yes

### DIFF
--- a/conf/master
+++ b/conf/master
@@ -125,7 +125,7 @@
 # The master can take a while to start up when lspci and/or dmidecode is used
 # to populate the grains for the master. Enable if you want to see GPU hardware
 # data for your master.
-# enable_gpu_grains: False
+# enable_gpu_grains: True
 
 # The master maintains a job cache. While this is a great addition, it can be
 # a burden on the master for larger deployments (over 5000 minions).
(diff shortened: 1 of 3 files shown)


cachedout (approved):

terminalmage:
This is an incorrect change. The config stub for the master governs _master_ grains, and `enable_gpu_grains` is set to to `False`, so this option would be disabled when compiling grains for the master. Grains are compiled on the master via the `MasterMinion` class, which provides an interface to things like execution modules, which normally run on a minion, to code that runs on the master (for example, in some runner functions).

It seems you have confused this with the minion, which because it does not have a default value set in the `DEFAULT_MINION_OPTS`, will default to enabling this config option.

terminalmage:
This PR has been reverted and corrected via https://github.com/saltstack/salt/pull/49375

ralish:
My apologies @terminalmage and thanks for correcting my mistake!

Issue #45620, opened 2018-01-22
Salt CLI is rounding floats to 2 decimal places


jmarinaro:
### Description of Issue/Question
When passing arguments via the command line, salt is rounding floats to 2 decimal places. I have verified this behavior exists in develop, 2017.7, and 2016.11. If I pass an argument I expect it to remain as is.

### Steps to Reproduce Issue
```
root@a0e930608610:/# salt-call test.arg_clean dec2='1516645352.09' dec3='1516645352.009' dec4='1516645352.0009'
local:
    ----------
    args:
    kwargs:
        ----------
        dec2:
            1516645352.09
        dec3:
            1516645352.01
        dec4:
            1516645352.0
```

### Versions Report
```
root@a0e930608610:/# salt-call --versions-report
Salt Version:
           Salt: 2017.7.2

Dependency Versions:
           cffi: Not Installed
       cherrypy: Not Installed
       dateutil: 2.4.2
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.8
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.4.6
   mysql-python: Not Installed
      pycparser: Not Installed
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: Not Installed
         Python: 2.7.12 (default, Dec  4 2017, 14:50:18)
   python-gnupg: Not Installed
         PyYAML: 3.11
          PyZMQ: 15.2.0
           RAET: Not Installed
          smmap: Not Installed
        timelib: Not Installed
        Tornado: 4.2.1
            ZMQ: 4.1.4

System Versions:
           dist: Ubuntu 16.04 xenial
         locale: ANSI_X3.4-1968
        machine: x86_64
        release: 4.10.0-42-generic
         system: Linux
        version: Ubuntu 16.04 xenial
```

jmarinaro:
Looks like it must be something with python2.7, I still see the issue with 2018.3.2
```
root@74ba8eb94a45:/# salt-call --versions-report
Salt Version:
           Salt: 2018.3.2
 
Dependency Versions:
           cffi: 1.5.2
       cherrypy: Not Installed
       dateutil: 2.4.2
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
[...]

PR #49187, opened 2018-08-18
Fix for #45620: "Salt CLI is rounding floats to 2 decimal places"

erwindon:
### What does this PR do?
This PR fixes a number representation issue under older Python versions.

#45620 

### Previous Behavior
Numbers were represented with a maximum of 12 digits of precision.

### New Behavior
Numbers are represented with the maximum precision.
For the DOUBLE format that is used, that is about 16 digits.

In python issue https://bugs.python.org/issue9337 it was decided that the artificial limitation to 12 digits should be removed. Therefore starting with Python3.2 it behaves differently (better).

### Tests written?
No
Test with `salt-call test.echo 1234567890.123456`
When using python27 the result previously was `1234567890.12`
When using python34 the result previously was `1234567890.123456`
Now the result is `1234567890.123456` in all cases.

### DIFF
--- a/salt/output/nested.py
+++ b/salt/output/nested.py
@@ -81,12 +81,14 @@ class NestDisplay(object):
             )
         # Number includes all python numbers types
         #  (float, int, long, complex, ...)
+        # use repr() to get the full precision also for older python versions 
+        # as until about python32 it was limited to 12 digits only by default
         elif isinstance(ret, Number):
             out.append(
                 self.ustring(
                     indent,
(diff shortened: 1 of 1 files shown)


cachedout (approved):
Very nice!

erwindon:
@cachedout build fails for `py3-centos-7`. it seems unrelated to my change. previous build was ok and only a small comment changed. the job complains about a salt-master being unreachable in 2 cases. what is the proper way forward?

rallytime (approved):

PR #49197, opened 2018-08-19
File state line ending fixes

dwoz:
### What does this PR do?

Fix tests in `integration.states.test_file`

### Tests written?

No - Fixing tests

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/modules/file.py
+++ b/salt/modules/file.py
@@ -4080,7 +4080,12 @@ def get_managed(
                 urlparsed_source.netloc, urlparsed_source.path).rstrip(os.sep)
         unix_local_source = parsed_scheme in ('file', '')
 
-        if unix_local_source:
+        if parsed_scheme == '':
+            parsed_path = sfn = source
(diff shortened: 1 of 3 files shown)


salt/modules/file.py
```diff
@@ -4080,7 +4080,12 @@ def _get_local_file_source_sum(path):
                 urlparsed_source.netloc, urlparsed_source.path).rstrip(os.sep)
         unix_local_source = parsed_scheme in ('file', '')
```
gtmanfred: is this line still needed?

gtmanfred (approved):
I would also like for @terminalmage to just double check this.

salt/modules/file.py
dwoz: Yes, It get's used later in the method.

gtmanfred: cool

terminalmage (approved):

PR #49199, opened 2018-08-20
Fix typeerror unicode

jacobweinstock:
fix TypeError: 'unicode' does not have the buffer interface and updates for python 3 support

`from __future__ import unicode_literals` - breaks this module in python 2. I found python 3 to not be supported as well. It appears that the pypureomapi library leverages sockets that are designed to only take byte strings. In python 2, strings are already byte strings and using unicode_literals causes the TypeError. In python 3 strings are unicode by default and thus need to be encoded to byte strings. This update encodes string literals into byte strings and provides support for both python 2 and 3.

### What does this PR do?
fix TypeError: 'unicode' does not have the buffer interface introduced by `from __future__ import unicode_literals` and handles python 3 support

n/a

### Tests written?

No

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/modules/omapi.py
+++ b/salt/modules/omapi.py
@@ -13,8 +13,7 @@ config or pillar:
 '''
 
 # Import python libs
-# Not importing unicode_literals as pypureomapi needs byte strings not unicode
-from __future__ import absolute_import, print_function
+from __future__ import absolute_import, print_function, unicode_literals
 import logging
 import struct
 
--- a/salt/modules/omapi.py
+++ b/salt/modules/omapi.py
@@ -13,9 +13,13 @@ config or pillar:
 '''
 
 # Import python libs
-from __future__ import absolute_import, print_function, unicode_literals
+# Not importing unicode_literals as pypureomapi needs byte strings not unicode
+from __future__ import absolute_import, print_function
 import logging
 import struct
+
+# Import salt libs
+import salt.utils.stringutils
 
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


salt/modules/omapi.py
```diff
@@ -13,7 +13,7 @@
 '''
 
 # Import python libs
-from __future__ import absolute_import, print_function, unicode_literals
+from __future__ import absolute_import, print_function
```
terminalmage: This needs a comment as to why we don't have `unicode_literals` imported, so that someone doesn't add it back later.

salt/modules/omapi.py
```diff
@@ -47,6 +47,10 @@ def _conn():
             __opts__.get('omapi.key', None))
     username = __pillar__.get('omapi.user',
             __opts__.get('omapi.user', None))
+    if key:
+        key = key.encode('UTF-8')
+    if username:
+        username = username.encode('UTF-8')
```
terminalmage: Instead of using `encode` here, you should be using `salt.utils.stringutils.to_bytes()`. This will avoid an `AttributeError` if this value happens to already be a bytestring (though this is very unlikely).

terminalmage (changes requested):

salt/modules/omapi.py
```diff
@@ -13,10 +13,14 @@
 '''
 
 # Import python libs
-from __future__ import absolute_import, print_function, unicode_literals
+# Not importing unicode_literals as pypureomapi needs byte strings not unicode
+from __future__ import absolute_import, print_function
```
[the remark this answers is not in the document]
terminalmage: This is a good point, actually.

terminalmage (approved):

gtmanfred (approved):

PR #49211, opened 2018-08-20
Fix diskusage

ProcenaFr:
### What does this PR do?
The variable "mount" could be in regex form, while _mount is the right existing form.
It fixes it.

### Tests written?

Yes

### Commits signed with GPG?

Yes


gtmanfred (changes requested):
@ProcenaFr can you rebase your branch to remove these extra commits?

Also, this should probably be submitted to an older an older branch, like 2017.7 or 2018.3, whichever is the oldest one that has this bug in it.

Thanks!
Daniel

ProcenaFr:
Sorry I'm completely new to Git and it's taking me time.
I've done this PR : https://github.com/saltstack/salt/pull/49219

Issue #49212, opened 2018-08-20
Feature: Suppress a configurable domain in utils.network._generate_minion_id()


markuskramerIgitt:
## Proposal
When the salt minion lacks an id, [utils.network.generate_minion_id()](https://github.com/saltstack/salt/blob/943e21f3e6e314a0fe1af7f08354a2b58303d224/salt/utils/network.py#L169) generates a fully qualified domain name, using a variety of sources.

I propose a new setting `minion_id_generate_no_domain`, to suppresses a single domain in the result of above function.

### Current Behavior
The function returns `king_bob.foo.bar`
### Proposed behavior:
Without `minion_id_generate_no_domain `, as currently.
With `minion_id_generate_no_domain = foo.baz`, as currently
With `minion_id_generate_no_domain = foo.bar`, the function returns `king_bob `

Ch3LL:
thanks for the PR!! As the pr resolves this issue I will close but please let me know if it needs to be reopened :)

markuskramerIgitt:
thank you, @Ch3LL

markuskramerIgitt:
@Ch3LL, could you please help me with PR https://github.com/saltstack/salt/pull/49378
Which are the requested changes?

markuskramerIgitt:
Hi @Ch3LL, I think the requested changes are from terminalmage, I asked him.

Ch3LL:
glad you got it figured out. let me know if  can help with anything else :)

markuskramerIgitt:
Yep, and terminalmarge already explained his request in all details.


Thank you for the offer.

I just looked at your PR https://github.com/saltstack/salt/pull/48866.
I will come back to twangboy and might come back to you because I would like that Jenkins creates the Windows msi installer. 

First I need to verify the installer.

Issue #48629, opened 2018-07-17
Prefer FQDN for minion id only by default    (Proposal) 


markuskramerIgitt:
### Proposal
The salt minion may lack an explicitly declared id.

Without id, [salt/utils/network._generate_minion_id()](https://github.com/saltstack/salt/blob/943e21f3e6e314a0fe1af7f08354a2b58303d224/salt/utils/network.py#L100) prefers FQDN (fully qualified domain name) over hostname. The function uses a variety of sources for the minion id.

I would like to introduce a new setting `minion_id_prefer_FQDN = True`, which leaves all logic as is by default, and otherwise prefers hostname over FQDN.

My goal is that minions automatically adapt their id to changing hostnames by setting `minion_id_caching = False` and `minion_id_prefer_FQDN = False`.  

### Implementation ideas
I think there are 3 changes required.

1) FQDN is prefered  [currently](https://github.com/saltstack/salt/blob/943e21f3e6e314a0fe1af7f08354a2b58303d224/salt/utils/network.py#L139) by the order of various id sources:
```
    hosts = DistinctList().append(socket.getfqdn()).append(platform.node()).append(socket.gethostname())
```
which, depending on the value of `minion_id_prefer_FQDN`, would then become
```
     node, gethostname, getfqdn 
```

2) The [test case salt.tests.unit.utils.test_network.test_generate_minion_id_platform_fqdn(self)](https://github.com/saltstack/salt/blob/943e21f3e6e314a0fe1af7f08354a2b58303d224/tests/unit/utils/test_network.py#L454) would need access to the new variable.

3) Documentation
E.g.  [the minion config file](https://github.com/saltstack/salt/blob/943e21f3e6e314a0fe1af7f08354a2b58303d224/conf/minion#L102):

```
# Explicitly declare the id for this minion to use, if left commented the id
# will be the hostname as returned by the python call: socket.getfqdn()
```

gtmanfred:
My biggest concern is that if the minion id changes with the hostname, then you are going to have problems with having to reaccept the key by the new name on the master, and any attempt to automate that feels like it is just going to cause security problems.

@saltstack/team-core anyone else have an opinion about this?

cro:
Historical context: every time we've changed the algorithm for determining minion_id, it has resulted in a world of pain.

DmitryKuzmenko:
Why not to leave the default behavior as is but make an optional way to determine the personal order as a `list` like it's done in `/etc/nsswitch.conf`?
[...]

PR #49213, opened 2018-08-20
Option to remove domain in the generated minion_id

marbx:
Replaced by PR https://github.com/saltstack/salt/pull/49378

### What does this PR do?
When a minion lacks its minion_id at startup, salt.config.\_\_init\_\_.py requests one from utils.network.generate_minion_id(), which returns the fully qualified domain name (FQDN) of the host.

When all minions are within one domain and named after their hostnames, generated a "FQDN minion id" would be an inconsistent naming.

salt.config.\_\_init\_\_.py has the option to lowercase the minion_id.
salt.config.\_\_init\_\_.py also has the option to append a domain to a generated "hostname minion id".

I hereby propose a new option, to generate a "hostname minion id", by removing a single domain from a FQDN.

The new option could be named:
`minion_id_remove_domain`
`minion_id_filter_domain`
`minion_id_domain_filter`
`remove_domain` 
`cloak_domain`
`minion_id_fqdn_filter` 
`minion_id_strip_generated_domain` 
`minion_id_generate_no_domain`

Issue #49212

### Previous Behavior (example)

The generated minion_id is `king_bob.foo.org`

### New Behavior (example)

With `minion_id_remove_domain = '' `, as previously.
With `minion_id_remove_domain = bar.org`, as previously.
With `minion_id_remove_domain = foo.org`, the generated minion_id becomes `king_bob `

### Tests written?
Yes


salt/config/__init__.py
```diff
@@ -944,6 +944,9 @@ def _gather_buffer_space():
     # Always generate minion id in lowercase.
     'minion_id_lowercase': bool,
 
+    # If set, suppress this domain in a generated minion id.
```
cachedout: It's not clear to me from this description that this needs to be a string containing the domain one wishes to suppress. Also, what if somebody wants to suppress multiple domains?

terminalmage (commented):
I'm not in love with the name of this config option. @saltstack/team-core do you have ideas of what else to call it?

salt/config/__init__.py
markuskramerIgitt: I added "single" and an example to the description.
Is it ok when I add more than one line description?

I think one should not suppress multiple domains because the hostname may not be unique in multiple domains.
Stripping multiple domains could lead to name clashes.

markuskramerIgitt:
@terminalmage, agreed.
Please consider the current name of this config option as a working title. 
Could you please name it consistently?

markuskramerIgitt:
What about `minion_id_strip_generated_domain`?

isbm:
`minion_id_short`.

isbm (commented):
`minion_id_stripped` But I dislike that option name. Why not `minion_id_fqdn_filter`? And why just one value, instead of a list?

But I'd drop that option entirely, and just live with `minion_id_stripped` that appears _all the time_ alongside the `minion_id`. If both are the same, then they are just same, why bother?

markuskramerIgitt:
@isbm, `minion_id_fqdn_filter` is fine.

In my opinion this setting should only be exactly one domain and not a list because the hostname may not be unique in multiple domains. Filtering multiple domains from a generated minion id could lead to name clashes.

Example:  With `minion_id_fqdn_filter = foo.bar`, generate_minion_id() would return king_bob in domain foo.bar and king_bob.foo.baz in domain foo.baz and prevent a name clash.


Of course you can use the generated fully qualified domain name if the domain does not bother you.

Our minions are named without domain name, and we want to keep it that way, but make use of generate_minion_id().

markuskramerIgitt:
I would assume that a fair share of users name their minions without domain, because the documentation is written that way and they happen to be in just one (big) domain.

Anyhow, `minon_id` may or may not contain the FQDN. 

That makes generate_minon_id() not generic enough.
This is an effort to make it more useful for domain-less minion id environments.

isbm:
@markuskramerIgitt exactly. So if you _always_ strip domain from `minion_id` and _always_ put it to e.g. `minion_id_short`, you  won't need any options/tumblers/tweaks, but just use it right away. If both grains are the same, then you just don't really care anyway.

So my question still here: why do we need an _option_ to strip some domain or not. Especially, if the "stripped" ID is not always stripped. This only brings even more confusion in my eyes.

markuskramerIgitt:
@isbm, because Salt only generates a FQDN minion_id.

isbm:
@markuskramerIgitt oh, dang!.. I think to my vacuum tube driven brains only now got that you want to have `minion_id` but control its value whether it is FQDN or just a hostname.

`minion_id_filter`. :laughing:

markuskramerIgitt:
@isbm, this is it.
Glad I'm not the only vacuum driven brain around. :-)

terminalmage (changes requested):
~You've only included a default value for this in the `DEFAULT_MASTER_OPTS`, we need one in the `DEFAULT_MINION_OPTS` as well.~ **Disregard this, apparently I can't read code today.**

Additionally, I would like to see some documentation of this in `doc/ref/configuration/minion.rst`. Right after `minion_id_caching`, around line 620, would be a good place.

markuskramerIgitt:
@terminalmage, sure.

While writing the documentation, I read the description of `append_domain`.

Now I find `append_domain` more readable than `minion_id_fqdn_filter`. What do you think?

markuskramerIgitt:
A symmetrical naming to `append_domain` would be nice, because of the symmetric working.

I like `remove_domain` .
What do you think of `cloak_domain`?

I no more would like `filter_domain` because it confuses me slightly: a filter can either block or let pass through.

isbm:
@markuskramerIgitt So you write in the description:

> I hereby propose a new option, to generate hostname for minion_id, by **filtering a single domain** from its FQDN.

And then:

>  a filter can either block or let pass through.

I am not sure you stay consistent to your own proposal. :wink: My point is that you actually **filtering** domain, as you write it by yourself in your own PR description (I just swapped to `remove_domain`):

Without `remove_domain`, as previously (i.e. `king_bob.foo.com`).
With `remove_domain = bar.com`, as previously (i.e. `king_bob.foo.com`, **domain still there!** )
With `remove_domain = foo.com`, the generated `minion_id` is `king_bob`

It gets confusing and inconsistent then. If you say "filter something" then it makes sense. But if you say "append" or "remove" but you're not really always appending or removing, then this is not a good idea.

[IMAGE ALT TEXT](http://www.youtube.com/watch?v=GmZiNLRVRL4 "Video Title")

markuskramerIgitt:
@isbm, very good points! 
Very cool film!

I agree that options can/should be longer than CLI commands, and they should be named so that ordering them (alphabetically) groups them logically.
Bottom line:  the option should start with `minion_id_`

I also agree that “filter” is better than “remove”.
Bottom line: the option should contain `filter`, and I will adapt the documentation.

As the filter applies to a domain, not a fqdn, I think it should contain `domain`.
Do you agree?

This leads (me) to two choices:
`minion_id_filter_domain`
`minion_id_domain_filter`
What do you think?

I personally like the first, because VERB OBJECT is easier to read than OBJECT VERB.
The second has the theoretical advantage of a future option `minion_id_domain_append`, but see the next paragraph.


How could we rename the current option `append_domain` (to either `minion_id_domain_append` or `minion_id_append_domain `)?
We cannot simply rename it because it would break the existing setups.
I guess this would involve an "alias" and deprecate the current option....
If we leave it like today, the options are named inconsistently and `append_domain` remains scattered from other `minion_id_*` options.

isbm:
@markuskramerIgitt I personally would prefer `minion_id_filter_domain` because it is not enabling/disabling some sort of filter that is still needs to be "optionised", but actually specifying a specific value "to do what?" right away.

terminalmage:
Honestly, the word `filter` doesn't sound right to me. When you filter something, you're usually narrowing down a larger set of things, keeping only the ones that _match_ the filter. This isn't what the option is designed to do. `remove` sounds better to me.

If we're going to use something like `minion_id_remove_domain`, then in another PR we should change `append_domain` to `minion_id_append_domain`, allowing `append_domain` to continue to be used for backward compatibility. I can take care of this in a separate PR.

salt/config/__init__.py
```diff
@@ -944,8 +944,8 @@ def _gather_buffer_space():
     # Always generate minion id in lowercase.
     'minion_id_lowercase': bool,
 
-    # By default '', suppress this single domain (example 'foo.org') in a generated minion id.
-    'remove_domain': six.string_types,
+    # Remove this single domain (e.g. 'foo.org') in a generated minion id.
+    'minion_id_remove_domain': six.string_types,
```
isbm: Can be also `False`. In case of `False` it equals to an empty string, and in case of `True` it can go and remove any domain. Just suggestion.

markuskramerIgitt:
@terminalmage, @isbm, thank you for the discussion (and video clip).
I changed code and documentation to `minion_id_remove_domain`.

More changes needed?

Who enters a version in versionadded of the documentation?

salt/config/__init__.py
```diff
@@ -944,6 +944,9 @@ def _gather_buffer_space():
     # Always generate minion id in lowercase.
     'minion_id_lowercase': bool,
 
+    # Remove this single domain (e.g. 'foo.org') in a generated minion id.
+    'minion_id_remove_domain': six.string_types,
```
isbm: This can also have `bool`. In case of `False`, it is as same as an empty string. In case of `True`, it removes _all_ domains. This is just a suggestion.

isbm: That would be if there are several domains. However, if someone has one domain network and wants it w/o domains at all, then why not?

isbm: > Also, it would confuse me to set a bool to 'foo.org', and I would half expect Python to reject it

I am not following. In case of boolean, you do:

`minion_id_remove_domain = true`

And so then every ID gets domain remove all the time, regardless. That is quite useful if you have one admin network and your machines are, say, mainly hardware and you want to access them via admin network, which sits behind another internal DNS. Or it is a VM, but it is not called `files.home.lan` but some coding scheme that tells you where exactly this machine hosted.

markuskramerIgitt: This sounds to me like a description of a new PR on a new variable `minion_id_remove_all_domains: bool`

markuskramerIgitt: If you would create such an option exists, I would 
leave `minion_id_remove_all_domains = false` and
set `minion_id_remove_domain = foo.bar` because I only want one single domain to be removed.


Lately, [there was an idea](https://github.com/saltstack/salt/issues/48629#issuecomment-414678059) for yet another options:

    minion_white_list_domain: good1.org, good2.org, good3.org
    minion_black_list_domain: evil1.org, evil2.org, evil3.org

DmitryKuzmenko: @isbm In just an abstract case in Linux for `mnion_id_remove_domain = True` I'd expect it'll remove the domain specified in `resolv.conf` ;)

isbm: @DmitryKuzmenko which is actually true (as well). This is exactly the point: clean away _any_ kind of domain, no matter where it comes from. At the end you end up only with the hostname. Pretty useful for small scale deployments.

If I have at all only 5-10 machines in a soho business office to look after, all of them are resolvable just by a name although might sit in a different domains, on what reason would I dislike such feature?

isbm: > This sounds to me like a description of a new PR on a new variable > `minion_id_remove_all_domains: bool`

Yes, it could be. But then it gets complicated, because "all domains" is like a list of possible domains, while "remove domain" sounds just as it is: remove a domain from the ID. And if it is `"true"` (or `"*"` if you want) then it will just match everything. Why I am not in favour of an asterisk, because it will look like probably other options available, e.g. `foo*`.

Hence I would prefer to have one option that can simply take different kind of options. If it is `true`, then just razor cut everything. If it is `false`, do not touch anything. If it is a specific domain, remove only it.

_MAYBE_, recalling to the black/white list, you want to add a simple matcher by an asterisk then. But that is indeed a different topic/PR and discussion, I would refrain from going there.

marbx (commented):
@isbm, I am glad that you proposed and argued for a possible **boolean** value of `minion_id_remove_domain`.

This just allowed me to workaround   issue https://github.com/saltstack/salt/issues/52162

PR #49216, opened 2018-08-20
Add the saltcli tests to the filename map salt/client/*

terminalmage:
This updates the filename map added in #47337 to ensure that the salt CLI tests are run when anything in salt.client files are updated.


gtmanfred (approved):

s0undt3ch:
Port to master https://github.com/saltstack/salt/pull/56760

PR #49219, opened 2018-08-21
Fix2 diskusage

ProcenaFr:
### What does this PR do?
The variable "mount" could be in regex form, while _mount is the right existing form.
It fixes it.

### Tests written?

Yes

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/beacons/diskusage.py
+++ b/salt/beacons/diskusage.py
@@ -89,10 +89,10 @@ def beacon(config):
         mount = mounts.keys()[0]
 
         try:
-            _current_usage = psutil.disk_usage(mount)
+            _current_usage = psutil.disk_usage(_mount)
         except OSError:
             # Ensure a valid mount point
-            log.warning('{0} is not a valid mount point, try regex.'.format(mount))
+            log.warning('{0} is not a valid mount point, try regex.'.format(_mount))
             for part in parts:
                 if re.match(mount, part.mountpoint):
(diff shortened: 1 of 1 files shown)


gtmanfred (commented):
I updated the branch to be merged into to 2017.7, now we just need to wait for the tests to run again.

salt/beacons/diskusage.py
```diff
@@ -89,10 +89,10 @@ def beacon(config):
         mount = mounts.keys()[0]
 
         try:
-            _current_usage = psutil.disk_usage(mount)
+            _current_usage = psutil.disk_usage(_mount)
```
cachedout: I don't understand this at all. You've set this to an undefined variable.

gtmanfred: in the proposed fix, the mount a couple lines above this was changed to `_mount` so it seems like that was not copied to the PR.

gtmanfred:
It looks like this change actually needs to only be made against the 2018.3 branch, and this pr was a change made in the 2017.7 branch

https://github.com/saltstack/salt/blob/2018.3/salt/beacons/diskusage.py#L99
is different from 
https://github.com/saltstack/salt/blob/2017.7/salt/beacons/diskusage.py#L89

gtmanfred:
This should be opened using the 2018.3 branch.

Issue #49222, opened 2018-08-21
Salt-key using external auth and token file is broken


msciciel:
### Description of Issue/Question
It's not possible to use token file with external auth:
```
$ salt-key -T --auth=pam
username: example
password:
Error: 'token_file'

$ ls salt_token
ls: cannot access salt_token: No such file or directory

### Setup
Enable external auth on salt-master:
```
external_auth:
  pam:
    users%:
      - '.*'
      - '@wheel'
      - '@runner'
      - '@jobs'
    limited%:
      - 'test.ping'
      - 'state.apply'
      - 'state.highstate'
      - 'grains.get'
      - 'grains.item'
      - 'grains.items'
      - 'pillar.get'
      - 'pillar.item'
      - 'pillar.items'
      - 'pkg.list_pkgs'
      - 'pkg.list_repos'
      - 'pkg.version'
```

### Steps to Reproduce Issue
$ salt-key -T --auth=pam

### Versions Report
```
Salt Version:
           Salt: 2017.7.6

Dependency Versions:
           cffi: Not Installed
       cherrypy: unknown
       dateutil: Not Installed
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.8.1
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.4.6
   mysql-python: Not Installed
      pycparser: Not Installed
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: Not Installed
         Python: 2.7.14 (default, Jan 31 2018, 02:12:13)
   python-gnupg: Not Installed
         PyYAML: 3.11
          PyZMQ: 14.5.0
           RAET: Not Installed
          smmap: Not Installed
        timelib: Not Installed
        Tornado: 4.2.1
            ZMQ: 4.0.5

System Versions:
           dist: centos 6.9 Final
         locale: UTF-8
        machine: x86_64
        release: 2.6.32-696.30.1.el6.x86_64
         system: Linux
        version: CentOS 6.9 Final
```

Ch3LL:
thanks for the PR! We really appreciate it :) since the PR resolves the issue I will close here and we can keep the discussion in the PR. Let me know if you want me to re-open here though :)

PR #49223, opened 2018-08-21
Fix token_file creation using external auth with salt-key

msciciel:
### What does this PR do?
Fix salt-key external auth using token file. salt-key with `--auth` and `--make-token` options needs to create token file. Options `token_file` which is used as a path for token file is provided by `config.client_config`.

#49222

### Previous Behavior
```
$ salt-key -T --auth=pam
username: example
password:
Error: 'token_file'

$ ls salt_token
ls: cannot access salt_token: No such file or directory
```

### New Behavior
```
$ salt-key -T --auth=pam
username: example
password:
Accepted Keys:
example1
example2
Denied Keys:
Unaccepted Keys:
Rejected Keys:

$ ls salt_token
salt_token
```

### Tests written?
No

### Commits signed with GPG?
No

### DIFF
--- a/salt/utils/parsers.py
+++ b/salt/utils/parsers.py
@@ -2522,7 +2522,7 @@ class SaltKeyOptionParser(six.with_metaclass(OptionParserMeta,
     process_config_dir._mixin_prio_ = ConfigDirMixIn._mixin_prio_
 
     def setup_config(self):
-        keys_config = config.master_config(self.get_config_file_path())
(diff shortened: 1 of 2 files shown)


cachedout (approved):

rallytime:
@msciciel This is causing a related test to fail. Can you take a look?

https://jenkinsci.saltstack.com/job/pr-kitchen-centos7-py3/job/PR-49223/2/

rallytime (approved):

Issue #48947, opened 2018-08-06
salt-cloud unable to create VM with external_ip: somename


dawidpogorzelski:
### Description of Issue/Question
salt-cloud is unable to create a VM on gce when `external_ip: somename` is provided regardless of whether that ip was pre-created or not.

```
ERROR   ] Caught Exception, terminating workers
TRACE: can't serialize <GCEAddress id="5717818400664030676" name="someipaddressname" address="someiphere" region="europe-west1">
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/site-packages/salt/cloud/__init__.py", line 67, in _call
    ret = func(*args, **kwargs)
  File "/usr/local/lib/python2.7/site-packages/salt/cloud/__init__.py", line 2317, in create_multiprocessing
    local_master=parallel_data['local_master']
  File "/usr/local/lib/python2.7/site-packages/salt/cloud/__init__.py", line 1284, in create
    output = self.clouds[func](vm_)
  File "/usr/local/lib/python2.7/site-packages/salt/cloud/clouds/gce.py", line 2581, in create
    ret = __utils__['cloud.bootstrap'](vm_, __opts__)
  File "/usr/local/lib/python2.7/site-packages/salt/utils/cloud.py", line 553, in bootstrap
    transport=opts.get('transport', 'zeromq')
  File "/usr/local/lib/python2.7/site-packages/salt/utils/cloud.py", line 1786, in fire_event
    event.fire_event(args, tag)
  File "/usr/local/lib/python2.7/site-packages/salt/utils/event.py", line 725, in fire_event
    dump_data = self.serial.dumps(data)
  File "/usr/local/lib/python2.7/site-packages/salt/payload.py", line 217, in dumps
    return msgpack.dumps(msg, default=ext_type_encoder, use_bin_type=use_bin_type)
  File "/usr/local/lib/python2.7/site-packages/msgpack/__init__.py", line 47, in packb
    return Packer(**kwargs).pack(o)
  File "msgpack/_packer.pyx", line 284, in msgpack._packer.Packer.pack
  File "msgpack/_packer.pyx", line 290, in msgpack._packer.Packer.pack
  File "msgpack/_packer.pyx", line 287, in msgpack._packer.Packer.pack
  File "msgpack/_packer.pyx", line 234, in msgpack._packer.Packer._pack
  File "msgpack/_packer.pyx", line 234, in msgpack._packer.Packer._pack
[...]

PR #49229, opened 2018-08-21
fixing https://github.com/saltstack/salt/issues/48947

dpogorzelski:
### What does this PR do?
it fixes https://github.com/saltstack/salt/issues/48947

I don't know the codebase enough so I'm not sure if this is the best way and place to do it especially the serialization part in utils/cloud.py so please review.

https://github.com/saltstack/salt/issues/48947

### Tests written?

No

### Commits signed with GPG?

No

### DIFF
--- a/salt/utils/cloud.py
+++ b/salt/utils/cloud.py
@@ -618,21 +618,13 @@ def bootstrap(vm_, opts=None):
     del event_kwargs['sudo_password']
     if 'password' in event_kwargs:
         del event_kwargs['password']
-    external_ip_dict = getattr(event_kwargs['vm_']['external_ip'], '__dict__', None)
-    if external_ip_dict:
-        if 'driver' in external_ip_dict:
-            del external_ip_dict['driver']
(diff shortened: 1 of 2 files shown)


salt/utils/cloud.py
```diff
@@ -618,6 +618,14 @@ def bootstrap(vm_, opts=None):
     del event_kwargs['sudo_password']
     if 'password' in event_kwargs:
         del event_kwargs['password']
+    external_ip_dict = getattr(event_kwargs['vm_']['external_ip'], '__dict__', None)
```
gtmanfred: IMHO, this should be done in the gce driver.

Then we should add an call to pass event_kwargs through salt.utils.data.simple_types_filter, so that it removes any none serializable objects from the dictionary.

gtmanfred (changes requested):

gtmanfred (approved):
Awesome, thank you for making those changes!

PR #49235, opened 2018-08-21
Use os.linesep.join instead of textwrap.dedent

twangboy:
### What does this PR do?
Uses `os.linesep.join` instead of `textwrap.dedent` to create the text strings. The problem with `textwrap.dedent` is that it always uses `posix-style` line endings.

https://github.com/saltstack/salt/pull/48960

### Tests written?
Yes

### Commits signed with GPG?
Yes

### DIFF
--- a/tests/support/helpers.py
+++ b/tests/support/helpers.py
@@ -25,6 +25,7 @@ import socket
 import subprocess
 import sys
 import tempfile
+import textwrap
 import threading
 import time
 import tornado.ioloop
--- a/tests/support/helpers.py
+++ b/tests/support/helpers.py
@@ -1574,5 +1574,5 @@ def dedent(text, linesep=os.linesep):
     if isinstance(text, six.text_type):
         linesep = salt.utils.stringutils.to_unicode(linesep)
     else:
-        linesep = salt.utils.stringutils.to_bytes(linesep)
+        linesep = salt.utils.stringutils.to_str(linesep)
     return linesep.join(textwrap.dedent(text).splitlines())
--- a/tests/support/helpers.py
+++ b/tests/support/helpers.py
@@ -1565,3 +1565,14 @@ def this_user():
     if salt.utils.is_windows():
         return salt.utils.win_functions.get_current_user(with_domain=False)
     return pwd.getpwuid(os.getuid())[0]
+
+
+def dedent(text, linesep=os.linesep):
+    '''
+    A wrapper around textwrap.dedent that also sets line endings.
+    '''
+    if isinstance(text, six.text_type):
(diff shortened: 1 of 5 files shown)


dwoz:
@twangboy I've combined this with the other fixes from 2017.7 branch:

https://github.com/twangboy/salt/pull/12

cachedout:
@dwoz Do you recommend that we close this then?

dwoz:
@cachedout No, @twangboy can merge my PR into his branch and then we can merge his branch into 2017.7.8 which will fix the remaining failing tests there.

terminalmage:
I would still prefer using `textwrap.dedent`, for ease of maintainability. It wouldn't be hard to write a helper that used platform-specific newlines.

terminalmage:
That is, one that would call `textwrap.dedent()` and _then_ make sure that the newlines matched `os.linesep`.

terminalmage:
OK, let me know if you need any assistance.

dwoz (approved):

tests/support/helpers.py
```diff
@@ -1565,3 +1566,14 @@ def this_user():
...
         return salt.utils.win_functions.get_current_user(with_domain=False)
     return pwd.getpwuid(os.getuid())[0]
+
+
+def dedent(text, linesep=os.linesep):
+    '''
+    A wrapper around textwrap.dedent that also sets line endings.
+    '''
+    if isinstance(text, six.text_type):
+        linesep = salt.utils.stringutils.to_unicode(linesep)
```
terminalmage: This needs to be `salt.utils.to_unicode(linesep)`, this function wasn't moved until 2018.3 and this PR has been opened against the 2017.7.8 release branch.

tests/support/helpers.py
```diff
@@ -1565,3 +1566,14 @@ def this_user():
...
+    else:
+        linesep = salt.utils.stringutils.to_str(linesep)
```
terminalmage: Similarly, this needs to be `salt.utils.to_str(linesep)`.

terminalmage (changes requested):

rallytime:
@twangboy and @dwoz Why is this submitted against 2017.7.8? I was under the impression we were only working on some last package fixes, not code changes to meet the release deadline.

dwoz:
@rallytime, I think @twangboy was just trying to make the 2017.7.8 test suite for windows start passing. If we don't need or want these test fixes in 2017.7.8 then we can close this.

rallytime:
Ah, OK sounds good. I spoke with @twangboy offline as well and he confirmed the same thing. I'll close this.

PR #49244, opened 2018-08-22
Test fixes: unit.fileserver.test_gitfs

dwoz:
### What does this PR do?

Multiple fixes for unit.fileserver.test_gitfs:
- file:// source repos need all forward slashes
- Use `salt.utils.files.rm_rf` for recursive removes
- Patch `salt.utils.files.rm_rf` to work with unicode filenames on windows python 2

Fixes:
```
unit.fileserver.test_gitfs.GitPythonTest.test_dir_list
unit.fileserver.test_gitfs.GitPythonTest.test_disable_saltenv_mapping_global_with_mapping_defined_globally
unit.fileserver.test_gitfs.GitPythonTest.test_disable_saltenv_mapping_global_with_mapping_defined_per_remote
unit.fileserver.test_gitfs.GitPythonTest.test_disable_saltenv_mapping_per_remote_with_mapping_defined_globally
unit.fileserver.test_gitfs.GitPythonTest.test_disable_saltenv_mapping_per_remote_with_mapping_defined_per_remote
unit.fileserver.test_gitfs.GitPythonTest.test_envs
unit.fileserver.test_gitfs.GitPythonTest.test_file_list
unit.fileserver.test_gitfs.GitPythonTest.test_ref_types_global
unit.fileserver.test_gitfs.GitPythonTest.test_ref_types_per_remote
```
https://jenkinsci.saltstack.com/job/2018.3/job/salt-windows-2016-py2/28/#showFailuresLink

### Tests written?

Yes - fixing tests

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/utils/files.py
+++ b/salt/utils/files.py
@@ -547,8 +547,11 @@ def rm_rf(path):
     if os.path.islink(path) or not os.path.isdir(path):
         os.remove(path)
     else:
-        if six.PY2 and salt.utils.platform.is_windows() and isinstance(path, six.string_types):
-            path = path.decode('utf-8')
+        if salt.utils.platform.is_windows():
+            try:
+                path = salt.utils.stringutils.to_unicode(path)
+            except TypeError:
+                pass
         shutil.rmtree(path, onerror=_onerror)
 
 
--- a/salt/utils/files.py
+++ b/salt/utils/files.py
@@ -547,6 +547,8 @@ def rm_rf(path):
     if os.path.islink(path) or not os.path.isdir(path):
         os.remove(path)
     else:
+        if six.PY2 and salt.utils.platform.is_windows() and isinstance(path, six.string_types):
+            path = path.decode('utf-8')
         shutil.rmtree(path, onerror=_onerror)
 
 
--- a/tests/unit/fileserver/test_gitfs.py
+++ b/tests/unit/fileserver/test_gitfs.py
@@ -62,6 +62,8 @@ log = logging.getLogger(__name__)
 
 TMP_SOCK_DIR = tempfile.mkdtemp(dir=TMP)
 TMP_REPO_DIR = os.path.join(TMP, 'gitfs_root')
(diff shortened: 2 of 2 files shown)


salt/utils/files.py
```diff
@@ -547,6 +547,8 @@ def _onerror(func, path, exc_info):
     if os.path.islink(path) or not os.path.isdir(path):
         os.remove(path)
     else:
+        if six.PY2 and salt.utils.platform.is_windows() and isinstance(path, six.string_types):
+            path = path.decode('utf-8')
```
terminalmage: `salt.utils.stringutils.to_unicode()` should be used here instead. The reason for this is that the `unicode` type's `decode()` method will attempt to encode the string before decoding, because reasons:

```
>>> u'Д'.decode('utf-8')
UnicodeEncodeError: 'ascii' codec can't encode character u'\u0414' in position 0: ordinal not in range(128)
```

We can skip the PY2 check completely because the `to_unicode()` helper already performs this check. So, these two lines can be alternately written as:

```python
if salt.utils.platform.is_windows():
    try:
        path = salt.utils.stringutils.to_unicode(path)
    except TypeError:
        pass
```

If `to_unicode()` is passed a non string/bytestring/bytearray type, it will raise a `TypeError`, hence the exception catching. This also allows us to avoid the `isinstance` check.

tests/unit/fileserver/test_gitfs.py
```diff
@@ -422,6 +424,9 @@ def setUpClass(cls):
 
             # Add a tag
             repo.create_tag(TAG_NAME, 'HEAD')
+            # Older GitPython versions do not have a close method.
+            if hasattr(repo, 'close'):
+                repo.close()
```
terminalmage: Good catch :+1:

terminalmage (changes requested):

terminalmage (approved):

PR #49250, opened 2018-08-22
add feature to let unchanged deployments stay deployed

maxim-sermin:
### What does this PR do?
This PR adds a feature to the state 'jboss7.py' which compares the already deployed artifact with the one that should be deployed and skips the deployment if both artifacts are the same (have an identical hash).

None, this is something new.

### Previous Behavior
Artifact was always undeployed and deployed again when `jboss7.deployed` was executed for it. For software which consists of many rather large deployments of which only some have changed, this took some unnecessary amount of time.

### New Behavior
When the parameter `undeploy_force` is set to `false` and the artifact that should be deployed is the same as the one which is already deployed, the deployment is skipped. When the parameter is set to `true` or omitted, the previous behaviour is executed.

### Tests written?
No

### Commits signed with GPG?
No


cachedout (changes requested):
@maxim-sermin Could you please fix the lint warnings shown here? https://jenkinsci.saltstack.com/job/pr-lint/job/PR-49250/1/warnings52Result/new/

maxim-sermin:
@cachedout done

maxim-sermin:
@cachedout Yeah good idea, I just have not done any unit testing in Python yet. So it could be some time before I have the chance to look into this - I will come back to you once I have written some tests.

cachedout:
@maxim-sermin Sounds good. Feel free to contact us in the community slack channel if you have questions about writing tests.

cachedout:
@maxim-sermin Just checking in with you here. Anything we can do to help?

maxim-sermin:
@cachedout Much appreciating you taking care of me 😄 Actually I just started to look into Python unit testing yesterday, managed to run some Salt tests - looks good so far. Hope to get this done during the next week, I will get in touch if I come across any issues

maxim-sermin:
@cachedout I was wrong last time, it seems like PyCharm just skipped all the test methods instead of actually running them - no idea why. Do you know a way to run the unit tests locally?

rallytime:
Hey @maxim-sermin - there are some docs about running salt's test suite here: https://docs.saltstack.com/en/latest/topics/tutorials/writing_tests.html

maxim-sermin:
@cachedout @rallytime Thanks a lot for your support regarding the tests. I am happy to report that I have written some unit tests now. The test methods are mostly copy-paste because that's how I found it to be in `test_jboss7.py` - if you think it would be better to avoid code repetition, please let me know and I will refactor them.

rallytime (approved):
I think this is fine since that was the pattern originally in the testing code. Thank you for writing some new tests!

If you feel inclined, it might be a good idea to refactor some of these tests in the module to be more DRY in another PR, but I think this PR is fine for now.

PR #49253, opened 2018-08-22
skip ID 7 for vmware hard drives

gtmanfred:
### What does this PR do?

Skip 7 as the id for vmware hard disks, because it is reserved to the SCSI Controller

Same from our friends.

https://github.com/ansible/ansible/blob/devel/lib/ansible/modules/cloud/vmware/vmware_guest.py#L670

### Tests written?

Yes

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/cloud/clouds/vmware.py
+++ b/salt/cloud/clouds/vmware.py
@@ -674,6 +674,19 @@ def _get_size_spec(device, size_gb=None, size_kb=None):
     return disk_spec
 
 
+def _iter_disk_unit_number(unit_number):
+    '''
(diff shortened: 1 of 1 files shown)


gtmanfred:
Waiting for @grichmond-salt to verify that this actually fixes the issue for $customer.

cachedout (changes requested):
Some lint fixup is needed on this.

grichmond-salt:
The suggested fix works.  Please proceed.  Would be nice for this to be back-ported to 2017.7, also.
Thanks, @gtmanfred!

PR #49232, opened 2018-08-21
fix HTTP method for acl_info

remijouannet:
### What does this PR do?

fixing consul.acl_info
the module was using PUT instead of GET (https://www.consul.io/api/acl.html#read-acl-token)

### Previous Behavior
acl_info was not working

### New Behavior
acl_info is now working

### Tests written?

No

### Commits signed with GPG?

No

Please review [Salt's Contributing Guide](https://docs.saltstack.com/en/latest/topics/development/contributing.html) for best practices.

See GitHub's [page on GPG signing](https://help.github.com/articles/signing-commits-using-gpg/) for more information about signing commits with GPG.

PR #49261, opened 2018-08-22
Add consul acl state

remijouannet:
### What does this PR do?

add a consul module state with the following function

- consul.acl_present
- consul.acl_absent

this PR depend from this one https://github.com/saltstack/salt/pull/49232
because it need consul.acl_info

```
acl_present:
  consul.acl_present:
    - id: 38AC8470-4A83-4140-8DFD-F924CD32917F
    - name: acl_name
    - rules: node "" {policy = "write"} service "" {policy = "read"} key "_rexec" {policy = "write"}
    - type: client
    - consul_url: http://localhost:8500

acl_delete:
   consul.acl_absent:
     - id: 38AC8470-4A83-4140-8DFD-F924CD32917F
```

### Tests written?

No

I'm already using this in production, the idea is the acl are fully manage by salt
all i have is this state consul/acl.sls

```
{% for acl, args in pillar['consul']['acl'].iteritems()|sort %}
add_consul_{{ acl }}:
  consul.acl_present:
    - name: {{ args['name'] }}
    - id: {{ args['id'] }}
    - type: {{ args['type'] }}
    - rules: {{ args['rules'] }}
{% endfor %}
```

and all the acl are define in pillar

```
consul:
  acl:
    consul-agent:
      name: Agent Token
      id: D537FB10-7C95-4EFF-BFE5-4CDBEFDCC418
      type: client
      rules: node "" {policy = "write"} service "" {policy = "read"} key "_rexec" {policy = "write"}
    admin:
      name: admin
      id: BF2D7FE2-65D6-43FE-ADA6-92A6522F90BD
      type: client
      rules: key "" { policy = "write" } service "admin" { policy = "write" } service "" { policy = "read" }
```

so when i have to update or create an acl, i just have to edit the pillar and use the state my consul.acl


cachedout (changes requested):
Hi @remijouannet 

Could you please fix these lint errors? https://jenkinsci.saltstack.com/job/pr-lint/job/PR-49261/1/warnings52Result/new/

remijouannet:
hello @cachedout done ;)

remijouannet:
https://github.com/saltstack/salt/pull/49232 has been merged :)

salt/states/consul.py
```diff
@@ -0,0 +1,169 @@
+# -*- coding: utf-8 -*-
+'''
+Consul Managemnt
+===========================
```
rallytime: This header (`====`) needs to match the number of characters in the `Consul Management` title, otherwise the doc build will output a warning.

There's also a typo in the heading. `Management`.

salt/states/consul.py
```diff
@@ -0,0 +1,169 @@
...
+       consul.acl_absent:
+         - id: 38AC8470-4A83-4140-8DFD-F924CD32917F
+'''
+
+
+def _acl_changes(name, id=None, type=None, rules=None, consul_url=None, token=None):
+    '''
+       return True if the acl need to be update, False if it doesn't need to be update
+    '''
+    r = __salt__['consul.acl_info'](id=id, token=token, consul_url=consul_url)
```
rallytime: I know this is picky, but can you change these variable references to be at least 3 letters? It's a stylistic standard we like to have at Salt. `ret` is fine.

This needs to be changed in some of the other functions below as well.

salt/states/consul.py
```diff
@@ -0,0 +1,169 @@
...
+        return True
+    elif r['res'] and r['data'][0]['Type'] != type:
+        return True
+    else:
+        return False
+
+
+def _acl_exists(name=None, id=None, token=None, consul_url=None):
+    '''
+       return True if acl exist, if False it not
```
rallytime: `If acl exists, return True. Otherwise, return False.` might be a little more clear here.

salt/states/consul.py
```diff
@@ -0,0 +1,169 @@
...
+    '''
+
+    res = {'result': False, 'id': None}
+
+    if id:
+        r = __salt__['consul.acl_info'](id=id, token=token, consul_url=consul_url)
+    elif name:
+        r = __salt__['consul.acl_list'](token=token, consul_url=consul_url)
+
+    if len(r['data']):
```
rallytime: Can `if r.get('data')` work here instead of `len`? That might be a little more clear here.

salt/states/consul.py
```diff
@@ -0,0 +1,169 @@
...
+            'changes': {},
+            'result': True,
+            'comment': 'ACL "{0}" exists and is up to date'.format(name)}
+
+    exists = _acl_exists(name, id, token, consul_url)
+
+    if not exists['result']:
+        if __opts__['test']:
+            ret['result'] = None
+            ret['comment'] = "the acl doesn't exist, it will be create"
```
rallytime: `create` --> `created`

salt/states/consul.py
```diff
@@ -0,0 +1,169 @@
...
+            ret['comment'] = "the acl has been created"
+        elif not r['res']:
+            ret['result'] = False
+            ret['comment'] = "failed to create the acl"
+    elif exists['result']:
+        changes = _acl_changes(name=name, id=exists['id'], token=token, type=type, rules=rules, consul_url=consul_url)
+        if changes:
+            if __opts__['test']:
+                ret['result'] = None
+                ret['comment'] = "the acl exist, but it need to be update"
```
rallytime: `the acl exists and will be updated`

salt/states/consul.py
```diff
@@ -0,0 +1,169 @@
...
+            'name': id,
+            'changes': {},
+            'result': True,
+            'comment': 'ACL "{0}" does not exist'.format(id)}
+
+    exists = _acl_exists(name, id, token, consul_url)
+    if exists['result']:
+        if __opts__['test']:
+            ret['result'] = None
+            ret['comment'] = "the acl exists, it will be delete"
```
rallytime: `delete` --> `deleted`

rallytime (changes requested):
I have some small change requests here.

remijouannet:
@rallytime thx for the review, i will take a look at it as soon as i can

rallytime (approved):

salt/states/consul.py
```diff
@@ -0,0 +1,169 @@
...
+        return True
+    elif info['res'] and info['data'][0]['Type'] != type:
+        return True
+    else:
+        return False
+
+
+def _acl_exists(name=None, id=None, token=None, consul_url=None):
+    '''
+       If acl exists, return True. Otherwise, return False.
```
cachedout: I think this is a little misleading. This function seems to return a dictionary and this makes it seem like the return itself is actually a boolean.

cachedout: This seems misleading. This makes it seem like the function returns a boolean when it actually returns a dictionary.

salt/states/consul.py
```diff
@@ -0,0 +1,169 @@
...
+    '''
+
+    ret = {'result': False, 'id': None}
+
+    if id:
+        info = __salt__['consul.acl_info'](id=id, token=token, consul_url=consul_url)
+    elif name:
+        info = __salt__['consul.acl_list'](token=token, consul_url=consul_url)
```
cachedout: This doesn't handle the case where neither `id` or `name` is set and then L52 will stacktrace.

cachedout (changes requested):

salt/states/consul.py
remijouannet: tell me if the description is more clear now

cachedout (approved):

remijouannet:
@cachedout @rallytime thx, is this possible to backport this PR in 2017.7 ?

rallytime:
Hi @remijouannet - We only accept/back-port bug fixes in the dot branches so this won't be backported.

You can use [dynamic module distribution](https://docs.saltstack.com/en/latest/ref/file_server/dynamic-modules.html) to use this earlier on your own, however.

remijouannet:
ho, ok no problem
in which release this module will be include ?

rallytime:
@remijouannet It will be in the release codenamed `Neon`.

remijouannet:
ok thx for the answer

PR #49266, opened 2018-08-22
Remove salt.utils convenience functions

terminalmage:
These are slated to be removed for the Neon release. Now that we have branched for Fluorine, these can be nuked.


terminalmage:
There were some failures, as expected, due to a few references to the old salt.utils paths to these functions. I've pushed a commit to this PR to fix these, and I've also backported these fixes to 2018.3 via https://github.com/saltstack/salt/pull/49276.

gtmanfred (approved):

rallytime (approved):

PR #49270, opened 2018-08-22
Add async helper to test_sock_path_len

dwoz:
### What does this PR do?

This test function is causing another test:

unit.utils.test_event.TestAsyncEventPublisher.test_event_subscription

To fail with `Async operation timed out after 5 seconds`, Adding the
tornado Async test helper so that both tests pass.

https://jenkinsci.saltstack.com/job/2018.3/job/salt-windows-2016-py2/lastCompletedBuild/testReport/unit.utils.test_event/TestAsyncEventPublisher/test_event_subscription/

### Tests written?

No

### Commits signed with GPG?

Yes

### DIFF
--- a/tests/unit/test_minion.py
+++ b/tests/unit/test_minion.py
@@ -19,7 +19,9 @@ import salt.utils.event as event
 from salt.exceptions import SaltSystemExit
 import salt.syspaths
 import tornado
+import tornado.testing
 from salt.ext.six.moves import range
+
 
 __opts__ = {}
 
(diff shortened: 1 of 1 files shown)


dwoz:
@rallytime This is one that came up before, the last time I tried adding `tornado.testing.AsyncTestCase` to all the tests in `test_minion.py`. IIRC that caused some other failure. Hopefully this works when splitting the async test out to it's own class.

rallytime (approved):

cachedout (approved):

PR #49285, opened 2018-08-23
INFRA-7574 - MakoStack support for failing on file parse errors

tkwilliams:
Lotsa stuff:
- add support for MakoStack to fail on various serious conditions (template parse error, template missing)
- above support is configurable, with defaults retaining the legacy behaviour
- extensible config stanza support
- major doc-string improvements
- code quality cleanup (rename 1-char vars to 3-char, normalize for string.format() syntax, etc).


tkwilliams:
AFAICT all the pylint warnings are for the ansible/roster module, not the files I touched..

gtmanfred:
After updating the repo, the pylint errors are gone.

tkwilliams:
Sweet!  Thanks @gtmanfred

salt/pillar/makostack.py
```diff
@@ -432,49 +452,57 @@ def ext_pillar(minion_id, pillar, *args, **kwargs):
...
-            log.warning('Ignoring Stack cfg "%s": file does not exist', cfg)
+            log.warning('Ignoring MakoStack cfg `{}`: file not found'.format(cfg))
             continue
-        stack = _process_stack_cfg(cfg, stack, minion_id, pillar, namespace)
+        stack = _process_stack_cfg(cfg, stack, minion_id, pillar, namespace, config)
     return stack
 
 
-def _process_stack_cfg(cfg, stack, minion_id, pillar, namespace):
+def _process_stack_cfg(cfg, stack, minion_id, pillar, namespace, config):
```
cachedout: The CodeClimate static analyzer is wondering if we might be able to split this function out even further to reduce its length and complexity. Thoughts?

tkwilliams: I'm looking at it now...  Other than compressing a couple of function calls which happen to be split over several lines - e.g. the two `lookup.get_template()` calls, I don't really see anything that is reasonably factorable.  It's a pretty simple try loop processing lines from a config file, with two (necessary) except clauses, and then a return.  After the above consolidation, the function is 36 LOC, but 8 of THOSE lines are logging, so practically speaking it's 28 LOC.

I don't really see any obvious lines of demarc for decomposing it further, but if you have a suggestion I'll happily implement it :)

isbm (changes requested):

salt/pillar/makostack.py
```diff
@@ -553,6 +575,6 @@ def _parse_top_cfg(content):
         obj = salt.utils.yaml.safe_load(content)
         if isinstance(obj, list):
             return obj
-    except Exception as e:
+    except Exception as err:
         pass
```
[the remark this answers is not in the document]
tkwilliams: Sure thing.  Though the data itself can be both sensitive and quite large, so unless you feel strongly about it I think it's best to simply note that the content was successfully loaded, not give the whole payload...

salt/pillar/makostack.py
cachedout: Totally fine. I just like it to bring it to the attention of contributors. As with all static analyzers, sometimes it makes sense and sometimes it doesn't. Thanks as always, @tkwilliams

PR #49291, opened 2018-08-23
Add dedent that sets line endings

dwoz:
### What does this PR do?

Add dedent helper that sets line endings

### Tests written?

No

### Commits signed with GPG?

Yes

### DIFF
--- a/tests/support/helpers.py
+++ b/tests/support/helpers.py
@@ -25,6 +25,7 @@ import socket
 import subprocess
 import sys
 import tempfile
+import textwrap
 import threading
 import time
 import tornado.ioloop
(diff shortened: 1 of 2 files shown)


twangboy (approved):

dwoz:
Looks like I need to circle back on this:
https://jenkinsci.saltstack.com/job/pr-kitchen-ubuntu1604-py2/job/PR-49291/2/

dwoz:
@terminalmage Does this one look good to you?

tests/support/helpers.py
```diff
@@ -1565,3 +1566,17 @@ def this_user():
...
+
+
+def dedent(text, linesep=os.linesep):
+    '''
+    A wrapper around textwrap.dedent that also sets line endings.
+    '''
+    linesep = salt.utils.to_unicode(linesep)
+    unicode_text = textwrap.dedent(salt.utils.to_unicode(text))
+    clean_text = linesep.join(unicode_text.splitlines())
+    if unicode_text.endswith(salt.utils.to_unicode('\n')):
```
terminalmage: `to_unicode` isn't really necessary here, since the newline character is part of the ASCII set. You can just use `u'\n'` instead.

tests/support/helpers.py
```diff
@@ -1565,3 +1566,17 @@ def this_user():
...
+        clean_text = clean_text + linesep
```
terminalmage: The `+=` operator can be used here.

terminalmage (changes requested):

Issue #48880, opened 2018-08-02
Can't run cmd.run with UTF-8 chars as arguments / parameters. With custom module.


damntoken:
### Description of Issue/Question

I'm trying to create a custom module because standard ones are not sufficient enought for managaging certain Windows management tastks. 

**But I can't run custom execution module to execute powershell with any Õ Ä Ö Ü character.**

First I googled pyton ascii unicode a lot and tried unicode() encode() decode(), because I used str(name) and other variable verifications, but this has been futile because the last stripped down simple example showed me that the error is somewhere further than in my custom sls file.

### Steps to Reproduce Issue
Sample file /srv/salt/_modules/str_aduser.py

	# -*- coding: utf-8 -*-
	from __future__ import absolute_import
	import logging
	LOG = logging.getLogger(__name__)
	__virtualname__ = 'str_aduser'

	def __virtual__():
		# # Determine whether or not to load this module
		if __salt__['grains.equals']('osfinger','Windows-2016Server'):
			return __virtualname__
		else:
			LOG.error('osfinger: '+ __salt__['grains.get']('osfinger') +' not Windows-2016Server')
		return False

	def test(name):
		runtest=__salt__['cmd.run_all'](name,shell='powershell')
		#runtest=__salt__['cmd.powershell_all'](name)
		ret = {
			'runtest retrun': str(runtest['retcode']) +'|'+ runtest['stderr'] +'|'+ runtest['stdout'],
			}
		return ret

Run with ASCII chars only..
`salt 'abelia' saltutil.sync_all; salt 'abelia' str_aduser.test 'echo ABC > C:\salt\var\log\salt\test.txt'`
Result:  runtest retrun: 0|| , Result file content: ABC
Run with UTF chars also..
`salt 'abelia' saltutil.sync_all; salt 'abelia' str_aduser.test 'echo ABCÄÖÜ > C:\salt\var\log\salt\test.txt'`
Result: 
```
    The minion function caused an exception: Traceback (most recent call last):
      File "c:\salt\bin\lib\site-packages\salt\minion.py", line 1606, in _thread_return
        return_data = minion_instance.executors[fname](opts, data, func, args, kwargs)
      File "c:\salt\bin\lib\site-packages\salt\executors\direct_call.py", line 12, in execute
        return func(*args, **kwargs)
      File "c:\salt\var\cache\salt\minion\extmods\modules\str_aduser.py", line 93, in test
        runtest=__salt__['cmd.powershell_all'](name)
[...]

PR #49299, opened 2018-08-23
Work around cmd.run unicode issues in test for now

dwoz:
Cat is missing on windows. I tired unsuccesfully to make this work with
the 'type' command on windows. This is maybe not ideal but it stablizes
the test suite for now.

### Tests written?

No

### Commits signed with GPG?

Yes


twangboy (approved):

damon-atkins:
Try 'type'

dwoz:
@damon-atkins For python 2 I think we still run into this:

https://github.com/saltstack/salt/issues/48880

The main purpose of this is to try and get the windows test suite stable so we can start running windows PR tests. We can tackle the `cmd.run` thing in another test specific to that issue. (or maybe we just drop py2 support ;))

PR #49303, opened 2018-08-24
use os._exit instead of sys.exit when daemonizing

gtmanfred:
### What does this PR do?
According to the docs [here](https://docs.python.org/3/library/os.html#os._exit)

> `os._exit` should should normally only be used in the child process after a fork().

This will exit the process with status n, without calling cleanup
handlers, flushing stdio buffers, etc.

do not merge, this is just testing to see if it breaks any other tests for python<3.7 at least.

Fixes saltstack/salt-jenkins#1075

### Tests written?

Yes

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/utils/process.py
+++ b/salt/utils/process.py
@@ -71,7 +71,7 @@ def daemonize(redirect_out=True):
         if pid > 0:
             # exit first parent
             salt.utils.crypt.reinit_crypto()
-            sys.exit(salt.defaults.exitcodes.EX_OK)
+            os._exit(salt.defaults.exitcodes.EX_OK)
     except OSError as exc:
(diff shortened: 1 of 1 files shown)


gtmanfred:
requesting reviews just in case someone knows why we might not want to use this function

dwoz (approved):
Looks good to me; the python-daemon package also does this:

https://github.com/arnaudsj/python-daemon/blob/master/daemon/daemon.py#L573

rallytime (approved):

gtmanfred:
So does the `daemonize` package

https://github.com/thesharp/daemonize/blob/master/daemonize.py#L111

terminalmage (approved):

Issue #49296, opened 2018-08-23
`salt-run manage.bootstrap` always raises `KeyError: 'list_hosts'`


RobRuana:
### Description of Issue/Question
I receive the following error every time I try to use the `manage.bootstrap` runner as [described in the documentation](https://docs.saltstack.com/en/latest/ref/runners/all/salt.runners.manage.html#salt.runners.manage.bootstrap).

```
$ salt-run manage.bootstrap hosts='host1.example.com,host2.example.com'
Exception occurred in runner manage.bootstrap: Traceback (most recent call last):
  File "/usr/lib/python2.7/dist-packages/salt/client/mixins.py", line 387, in _low
    data['return'] = self.functions[fun](*args, **kwargs)
  File "/usr/lib/python2.7/dist-packages/salt/runners/manage.py", line 824, in bootstrap
    salt.client.ssh.SSH(client_opts).run()
  File "/usr/lib/python2.7/dist-packages/salt/client/ssh/__init__.py", line 703, in run
    if self.opts['list_hosts']:
KeyError: 'list_hosts'
```

### Steps to Reproduce Issue
Run the following command (with valid hosts and a valid roster):
```
salt-run manage.bootstrap hosts='host1.example.com,host2.example.com'
```

Here is the output demonstrating I have a valid `/etc/salt/roster`:

```
$ salt-ssh '*' test.ping
host1.example.com:
    True
host2.example.com:
    True
$ cat 
archive.uber.magfest.org:
  host: archive.uber.magfest.org
  user: root
  priv: /root/.ssh/id_rsa

backups.uber.magfest.org:
  host: backups.uber.magfest.org
  user: root
  priv: /root/.ssh/id_rsa
```
### Versions Report
```
$ salt-run --version
salt-run 2018.3.2 (Oxygen)
```

```
$ salt --versions-report
Salt Version:
           Salt: 2018.3.2
 
Dependency Versions:
           cffi: 1.5.2
       cherrypy: Not Installed
       dateutil: Not Installed
      docker-py: 3.3.0
          gitdb: 0.6.4
      gitpython: 1.0.1
          ioflo: Not Installed
         Jinja2: 2.8
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: 0.21.1
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.4.6
   mysql-python: Not Installed
      pycparser: 2.14
       pycrypto: 2.6.1
[...]

PR #49312, opened 2018-08-24
Fix keyerror in manage.bootstrap

Ch3LL:
### What does this PR do?
since `list_hosts` is an optional arg passed through the parser we can change it to a .get() call as it should not be required to be in opts.

Fixes https://github.com/saltstack/salt/issues/49296


gtmanfred (approved):

rallytime (approved):

RobRuana:
Wow, thanks for the quick turnaround!

Issue #49234, opened 2018-08-21
Windows service.restart not restarting a process that is already stopped.


whytewolf:
### Description of Issue/Question
On windows when you issue a `service.restart` of the process is already stopped it will throw an error.

Would be nice if that error wasn't fatal and allowed the start function to run effectively starting a stopped process.

### Setup
Setup a service and stop it. Then run service.restart on the service.

### Steps to Reproduce Issue
```
c:\salt\salt-call -l debug --local service.restart postgresql-x64-9.6
winion-0:
    [DEBUG   ] Reading configuration from c:\salt\conf\minion
    [DEBUG   ] Including configuration from 'c:\salt\conf\minion.d\_schedule.conf'
    [DEBUG   ] Reading configuration from c:\salt\conf\minion.d\_schedule.conf
    [DEBUG   ] Using cached minion ID from c:\salt\conf\minion_id: winion-0
    [DEBUG   ] Configuration file path: c:\salt\conf\minion
    [WARNING ] Insecure logging configuration detected! Sensitive data may be logged.
    [DEBUG   ] Grains refresh requested. Refreshing grains.
    [DEBUG   ] Reading configuration from c:\salt\conf\minion
    [DEBUG   ] Including configuration from 'c:\salt\conf\minion.d\_schedule.conf'
    [DEBUG   ] Reading configuration from c:\salt\conf\minion.d\_schedule.conf
    [DEBUG   ] Determining pillar cache
    [DEBUG   ] LazyLoaded jinja.render
    [DEBUG   ] LazyLoaded yaml.render
    [DEBUG   ] LazyLoaded jinja.render
    [DEBUG   ] LazyLoaded yaml.render
    [DEBUG   ] LazyLoaded service.restart
    [DEBUG   ] LazyLoaded cmd.run
    [INFO    ] Executing command ['net', 'stop', '/y', 'postgresql-x64-9.6'] in directory 'C:\Windows\system32\config\systemprofile'
    [ERROR   ] Command '['net', 'stop', '/y', 'postgresql-x64-9.6']' failed with return code: 2
    [ERROR   ] stdout: The postgresql-x64-9.6 service is not started.



    More help is available by typing NET HELPMSG 3521.
    [ERROR   ] retcode: 2
    [ERROR   ] Command '['net', 'stop', '/y', 'postgresql-x64-9.6']' failed with return code: 2
    [ERROR   ] output: The postgresql-x64-9.6 service is not started.



[...]

Issue #48828, opened 2018-07-30
win_service removing qoutes from bin_path


josemine:
### Description of Issue/Question
The win_service module is removing quotes from bin_path for both the create and modify function. If one goes to add a service in windows that has spaces using this module their system will be vulnerable. 
For more information you can take a look at
https://www.commonexploits.com/unquoted-service-paths/ and
http://cwe.mitre.org/data/definitions/428.html

I suggest line numbers 716 and 1027 should be deleted in salt/modules/win_service.py. Another options is to quote bin_path no matter what.

### Versions Report
(Provided by running `salt --versions-report`. Please also mention any differences in master/minion versions.)
### Master
```
Salt Version:
           Salt: 2018.3.2
 
Dependency Versions:
           cffi: 1.6.0
       cherrypy: Not Installed
       dateutil: 1.5
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.7.2
        libgit2: 0.26.3
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.5.6
   mysql-python: Not Installed
      pycparser: 2.14
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: 0.26.4
         Python: 2.7.5 (default, Jul 13 2018, 13:06:57)
   python-gnupg: Not Installed
         PyYAML: 3.11
          PyZMQ: 15.3.0
           RAET: Not Installed
          smmap: Not Installed
        timelib: Not Installed
        Tornado: 4.2.1
            ZMQ: 4.1.4
 
System Versions:
           dist: centos 7.5.1804 Core
         locale: UTF-8
        machine: x86_64
        release: 3.10.0-693.21.1.el7.x86_64
         system: Linux
        version: CentOS Linux 7.5.1804 Core
```
### Minion
```
Salt Version:
           Salt: 2018.3.2

Dependency Versions:
           cffi: 1.10.0
       cherrypy: 10.2.1
       dateutil: 2.6.1
      docker-py: Not Installed
          gitdb: 2.0.3
      gitpython: 2.1.3
          ioflo: Not Installed
         Jinja2: 2.9.6
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: Not Installed

           Mako: 1.0.6
   msgpack-pure: Not Installed
[...]

PR #49327, opened 2018-08-25
Fix issues with win_service

twangboy:
### What does this PR do?
Add shlex.quote
Properly quote the path to the binary (`bin_path`) on `create` and `modify`. It was removing double quotes. Double quotes are needed for paths to binaries with spaces. Single quotes will fail also.
Don't crash on `service.stop` when the service is already stopped
Don't crash on `service.delete` when the service is not present
Fix issues with the exception handling

https://github.com/saltstack/salt/issues/48828
https://github.com/saltstack/salt/issues/49234

### Tests written?
No

### Commits signed with GPG?
Yes


twangboy:
I still need to write some tests for this

damon-atkins:
- Consider using from salt.utils.win_functions import **escape_argument**
- We should not convert single quotes as its a valid file name, even if its a bad idea for someone to use it in a filename.

damon-atkins:
ping @ciiqr

damon-atkins:
- Consider using from salt.utils.win_functions import **escape_argument** or **escape_for_cmd_exe** instead of shlex.quote
- We should not convert single quotes as its a valid file name, even if its a bad idea for someone to use it in a filename.
- Suggest you check for reporting `changes` impacts to states/service.py with the  start/stop returning true when already started/stopped etc.

The other alternative is to check for begins with " then leave it alone, otherwise wrap " around it after running  **escape_argument** or **escape_for_cmd_exe**

twangboy:
@damon-atkins 
Neither of the `win_functions` commands will work in this case as they actually escape the quotes at the beginning and end. For example, `C:\Temp Dir\test.exe` becomes `^"C:\\Temp Dir\test.exe^"`. The command needs to be passed with unescaped double quotes.

Since this is the service module, I can't think of an instance where the path to an exe would start or end with a single quote. It would always end in `.exe` and would always start with the drive letter.

The `_cmd_quote` function only strips quotes from the beginning and the end. It doesn't strip single quotes from anything inside. This produces a valid value no matter how the bin_path is quoted in the state file.

rallytime:
@dwoz can you take a look here?

dwoz (approved):

rallytime:
@twangboy It looks like there is a related windows test failing here:

https://jenkinsci.saltstack.com/job/pr-windows2016-py2/job/PR-49327/2/testReport/junit/integration.modules.test_service/ServiceModuleTest/test_service_status_running/

Yay for catching that now! 🎉

twangboy:
@rallytime This may be a flaky test. I can't replicate it manually. Is there a way to re run?

cachedout:
Might be related and might not, since this PR also whitelists that test for the first time. Still, though, we'll want to mark it as a flaky or fix the test before we merge this. I'll re-reun tests now to see what comes back.

PR #49330, opened 2018-08-25
Add a sys.utils_doc function to return docstrings for utils modules

terminalmage:
h/t Dominik Bessler in community Slack for the idea

### DIFF
--- a/salt/modules/sysmod.py
+++ b/salt/modules/sysmod.py
@@ -325,6 +325,55 @@ def renderer_doc(*args):
     return _strip_rst(docs)
 
 
+def utils_doc(*args):
+    '''
+    .. versionadded:: Neon
+
+    Return the docstrings for all utils modules. Optionally, specify a module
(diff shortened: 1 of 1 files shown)


gtmanfred:
we already have a sys.states_doc and a sys.runner_doc, why not just add a sys.utils_doc?

https://docs.saltstack.com/en/latest/ref/modules/all/salt.modules.sysmod.html#salt.modules.sysmod.state_doc

gtmanfred (commented):
we already have a sys.states_doc and a sys.runner_doc, why not just add a sys.utils_doc?

https://docs.saltstack.com/en/latest/ref/modules/all/salt.modules.sysmod.html#salt.modules.sysmod.state_doc

terminalmage:
Ah, I did not see those in sysmod.py, I did not scroll down far enough. Good call, I've modified the PR to do thia,.

gtmanfred (approved):

PR #49345, opened 2018-08-27
upgrade including linux kernels

gtmanfred:
### What does this PR do?
This is required for ubuntu 14.04 to pass.

### Tests written?

Yes

### Commits signed with GPG?

Yes


gtmanfred:
needs to be backported to 2017.7

PR #49349, opened 2018-08-27
INFRA-6930 - boto3_athena modules, first pass

tkwilliams:
Our devs told us they needed this, and then came back afterwards and said "oops, sorry, we meant GLUE, not Athena", so this is as much as I'm likely to do on this in the foreseeable future..

- exec module with pretty much everything available in the boto3 module
- state module with a couple of basic funtions (named_query_present() / named_query_absent())
- moderately tested but not exhaustively so

It IS very clean and elegant code, if I'm allowed to say so, so it might be useful as a template for later modules if nothing else...


rallytime (changes requested):
Can you fix the lint errors here?

https://jenkinsci.saltstack.com/job/pr-lint/job/PR-49349/2/warnings52Result/

Then we can do a more thorough review. :)

tkwilliams:
This sat for several months untouched - I /thought/ it was better polished than it is, so I'm gonna pull it back and buff it a bit more before PR again later on....

Please ignore :)

Issue #48608, opened 2018-07-16
user.list_users returns duplicates on MacOS


V3XATI0N:
### Description of Issue/Question

When running the `user.list_users` module on Mac OS minions, Salt returns two entries for each user account.

### Setup

No special setup

### Steps to Reproduce Issue

run `salt MacOS-Minion user.list_users`

### Versions Report
master:
```
Salt Version:
           Salt: 2018.3.0
 
Dependency Versions:
           cffi: 1.11.5
       cherrypy: 11.2.0
       dateutil: 2.4.2
      docker-py: Not Installed
          gitdb: 0.6.4
      gitpython: 1.0.1
          ioflo: Not Installed
         Jinja2: 2.8
        libgit2: 0.25.1
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: 1.0.3
   msgpack-pure: Not Installed
 msgpack-python: 0.4.6
   mysql-python: Not Installed
      pycparser: 2.18
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: 0.25.1
         Python: 2.7.12 (default, Dec  4 2017, 14:50:18)
   python-gnupg: Not Installed
         PyYAML: 3.11
          PyZMQ: 15.2.0
           RAET: Not Installed
          smmap: 0.9.0
        timelib: Not Installed
        Tornado: 4.2.1
            ZMQ: 4.1.4
 
System Versions:
           dist: Ubuntu 16.04 xenial
         locale: UTF-8
        machine: x86_64
        release: 4.2.0-042stab120.16
         system: Linux
        version: Ubuntu 16.04 xenial
```
minion:
```
Salt Version:
           Salt: 2018.3.0
 
Dependency Versions:
           cffi: 1.11.2
       cherrypy: 13.0.0
       dateutil: 2.6.1
      docker-py: Not Installed
          gitdb: 2.0.3
      gitpython: 2.1.7
          ioflo: Not Installed
         Jinja2: 2.10
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: 1.0.7
   msgpack-pure: Not Installed
 msgpack-python: 0.4.8
   mysql-python: Not Installed
      pycparser: 2.18
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: Not Installed
         Python: 2.7.14 (default, Mar 28 2018, 14:48:50)
   python-gnupg: 0.4.1
[...]

PR #49376, opened 2018-08-28
Fix issue with duplicates in user.list_users on OSX

twangboy:
### What does this PR do?
Fixes the issue where `user.list_users` was showing duplicate users names. Uses a set to do this as a set does not allow duplicates.

https://github.com/saltstack/salt/issues/48608

### Tests written?
No

### Commits signed with GPG?
Yes

### DIFF
--- a/pkg/osx/build_env.sh
+++ b/pkg/osx/build_env.sh
@@ -161,7 +161,7 @@ sudo -H $MAKE install
 ############################################################################
 echo -n -e "\033]0;Build_Env: libsodium\007"
 
-PKGURL="https://download.libsodium.org/libsodium/releases/libsodium-1.0.13.tar.gz"
+PKGURL="https://download.libsodium.org/libsodium/releases/old/libsodium-1.0.13.tar.gz"
 PKGDIR="libsodium-1.0.13"
 
 download $PKGURL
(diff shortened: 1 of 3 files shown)


cachedout:
This does not strike me as the right fix. IMHO, we need to figure out why the duplication is happening to begin with.

DmitryKuzmenko:
At least it's interesting why
```
>>> for p in pwd.getpwall():
...     print p[0]
```
gets a list of non-dups, but `[user.pw_name for user in pwd.getpwall()]` returns duplications. o_O

twangboy:
@DmitryKuzmenko There are still dups. The list is not sorted, so it's hard to tell

twangboy:
@cachedout I think the issue is with the `pwd.getpwall` command on osx. I don't get dupes on linux boxes. Is there a problem with filtering them out?

twangboy:
@cachedout @DmitryKuzmenko Changed it to use dscl

terminalmage (approved):
I usually see this as `/Users` and not `/users`. If dscl can use case-insensitive names, then I think this is fine.

twangboy:
@terminalmage Works fine with `/users` or `/Users`. Seems to be NOT case sensitive.

cachedout:
@twangboy We have a failed test here I'm afraid.

PR #35897, opened 2016-08-30
cachedir should be /cloud not /master

gtmanfred:
### What does this PR do?

Fix regression that moved the cloud cachedir to /master
### Tests written?

Not yet

PR #49377, opened 2018-08-28
Honor make_master also if master already exists

### LINKED ISSUE #49357: salt-cloud ignores make_master in map if master already exists
When having a map file containing a make_master: True statement, other nodes are only given this node as master node if the master node does not already exists.

### Setup
```
master:
  - master-0:
      make_master: True
normal:
  - node-0
  - node-1
```

### Steps to Reproduce Issue
Run salt-cloud with previously mentioned .map and notice that node-0 and node-1 get deployed with no master if master-0 already exists at the time of running salt-cloud.

```
# salt-cloud -m example.map
[WARNING ] 'master-0' already exists, removing from the create map.
The following virtual machines already exist:
  master-0
The following virtual machines are set to be created:
  node-0
  node-1

Proceed? [N/y] y
... proceeding
[WARNING ] There's no master defined on the 'node-0' VM settings.
```

### Versions Report
```
Salt Version:
           Salt: 2018.3.2

Dependency Versions:
           cffi: Not Installed
       cherrypy: Not Installed
       dateutil: Not Installed
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.7.2
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.5.6
   mysql-python: Not Installed
      pycparser: Not Installed
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: Not Installed
         Python: 2.7.5 (default, Aug  4 2017, 00:39:18)
   python-gnupg: Not Installed
         PyYAML: 3.11
          PyZMQ: 15.3.0
           RAET: Not Installed
          smmap: Not Installed
        timelib: Not Installed
        Tornado: 4.2.1
            ZMQ: 4.1.4

System Versions:
           dist: centos 7.4.1708 Core
         locale: ANSI_X3.4-1968
        machine: x86_64
        release: 3.10.0-693.11.6.el7.x86_64
         system: Linux
        version: CentOS Linux 7.4.1708 Core
```

TheLastProject:
### What does this PR do?
If make_master: True in a map, it'll be applied to other minions, even if the master already exists.

Fixes #49357

### Previous Behavior
See #49357

### New Behavior
See #49357 

### Tests written?

No

### Commits signed with GPG?

No


TheLastProject:
The pull request looks much larger than it is, due to indentation changes. The actual changes are fairly small.

salt/cloud/__init__.py
```diff
@@ -2083,128 +2083,142 @@ def run_map(self, dmap):
...
+        for name, profile in possible_master_list:
+            if profile.get('make_master', False) is True:
+                master_name = name
+                master_profile = profile
+
+        if master_name:
+            # If the master already exists, get the host
+            if master_name not in dmap['create']:
+                client = salt.client.get_local_client()
+                master_host = client.cmd(master_name, 'grains.get', ['host'])[master_name]
```
gtmanfred: I do not like this, we should do a query of the cloud provider to get this information instead of checking the grains.

There is always the possibility that the machine is not connected to this master.

gtmanfred (changes requested):
The rest of this looks fine, but salt-cloud should be able to work independent of the salt master.

salt/cloud/__init__.py
TheLastProject: Do you perhaps have some pointers on how to use salt cloud to query internally? I couldn't find a nice feature for that.

gtmanfred: You are already inside of salt-cloud, just use `self.query()` and find the instance.

salt/cloud/__init__.py
```diff
@@ -2083,128 +2085,145 @@ def run_map(self, dmap):
...
+        for name, profile in possible_master_list:
+            if profile.get('make_master', False) is True:
+                master_name = name
+                master_profile = profile
+
+        if master_name:
+            # If the master already exists, get the host
+            if master_name not in dmap['create']:
+                cloud_client = salt.cloud.CloudClient()
+                master_host = cloud_client.query()
```
gtmanfred: Why are you initializing a cloud client? you are already in one, just run self.query().

gtmanfred (changes requested):

salt/cloud/__init__.py
TheLastProject: There's no .query() on self.

gtmanfred: Oh, right, this is in the map, ok, what you should do is add

```
self.client = CloudClient(self.opts)
```

right here https://github.com/saltstack/salt/blob/develop/salt/cloud/__init__.py#L519

and then add self.client.query()

gtmanfred (approved):

salt/cloud/__init__.py
```diff
@@ -517,6 +519,7 @@ class Cloud(object):
     '''
     def __init__(self, opts):
         self.opts = opts
+        self.client = CloudClient(self.opts)
```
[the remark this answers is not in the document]
gtmanfred: this should be the last thing.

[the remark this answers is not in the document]
TheLastProject: Just checked on my dev box, you're indeed right, let me fix that real quick.

TheLastProject:
Hmm, it seems it fails a test. However, I sadly can't see why, because I get a 403 forbidden when trying to:
image

Could you perhaps tell me what is wrong so I can take a look at it?

TheLastProject:
Nevermind, found it in the logs. I'll try to take a look at this tomorrow:
```
16:42:21  --------  Tests with Errors  --------------------------------------------------
16:42:21    -> unit.test_cloud.MapConfTest.test_cloud_map_merge_conf  ...................
16:42:21        Traceback (most recent call last):
16:42:21          File "/tmp/kitchen/testing/tests/unit/test_cloud.py", line 120, in test_cloud_map_merge_conf
16:42:21            cloud_map = salt.cloud.Map(opts)
16:42:21          File "/tmp/kitchen/testing/salt/cloud/__init__.py", line 1678, in __init__
16:42:21            Cloud.__init__(self, opts)
16:42:21          File "/tmp/kitchen/testing/salt/cloud/__init__.py", line 522, in __init__
16:42:21            self.client = CloudClient(opts=self.opts)
16:42:21          File "/tmp/kitchen/testing/salt/cloud/__init__.py", line 193, in __init__
16:42:21            v_dirs = [self.opts['cachedir']]
16:42:21        KeyError: u'cachedir'
```

TheLastProject:
It seems to be caused by https://github.com/saltstack/salt/commit/d2db4ea7a207081ae50758e3e2cf627e32e5dd31#diff-f2223a6a0e1af2856634ba094e7f951bR1937. Simply removing the line should fix it, but I don't quite understand why the key gets removed in the first place, the comment doesn't explain it well enough for me, especially not because the defaults seem correct in of https://github.com/saltstack/salt/pull/35897/files#diff-f2223a6a0e1af2856634ba094e7f951bR1268.

If I understand correctly, that means the comment above https://github.com/saltstack/salt/pull/35897/files#diff-f2223a6a0e1af2856634ba094e7f951bR1943 is incorrect, otherwise the cachedir key would've been set.

Could it be we stumbled upon a regression introduced in that commit, or am I missing something?

cachedout:
@gtmanfred Can you address the above concern?

TheLastProject:
I was missing something. The reason the test failed is that it used a custom list of opts, which bypasses the configuration loading mechanism altogether. This custom list did not contain 'cachedir', which was fine up to now, because now the map instantiates a CloudClient which requires 'cachedir'. I believe my last commit should fix that, we'll know for sure when Jenkins completes (having some trouble getting them to run locally)

rallytime (approved):
Nice 👍 

Thanks for this contribution @TheLastProject and congrats on your first Salt PR!

gtmanfred:
Sorry about taking so long to get back to helping with this.

Nice Job :+1:

PR #49396, opened 2018-08-29
Add utility functions to slsutil module

amendlik:
### What does this PR do?
Add two utility functions to the `slsutil` execution module. These functions are for convenience when creating template files.

#### Banner
`slsutil.banner` creates a standard warning banner like this:
```
        ########################################################################
        #                                                                      #
        #              THIS FILE IS MANAGED BY SALT - DO NOT EDIT              #
        #                                                                      #
        # The contents of this file are managed by Salt. Any changes to this   #
        # file may be overwritten automatically and without warning.           #
        ########################################################################
```
The banner is fully customizable in its content and appearance.

#### Boolstr
`slsutil.boolstr` is for converting boolean values stored in Pillar to an appropriate string representation within templated files. This can obviously be done in Jinja, but it requires an entire if/else block and managing whitespace becomes difficult.

See embedded documentation for more details.

None

### Tests written?
Yes

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/modules/slsutil.py
+++ b/salt/modules/slsutil.py
@@ -336,3 +336,41 @@ def banner(width=72, commentchar='#', borderchar='#', blockstart=None, blockend=
         return result + '\n'
 
     return result
+
+
+def boolstr(value, true='true', false='false'):
+    '''
+    Convert a boolean value into a string. This function is
+    intended to be used from within file templates to provide
+    an easy way to take boolean values stored in Pillars or
+    Grains, and write them out in the apprpriate syntax for
+    a particular file template.
+
+    :param value: The boolean value to be converted
+    :param true: The value to return if ``value`` is ``True``
+    :param false: The value to return if ``value`` is ``False``
+
+    In this example, a pillar named ``smtp:encrypted`` stores a boolean
+    value, but the template that uses that value needs ``yes`` or ``no``
+    to be written, based on the boolean value.
+
+    *Note: this is written on two lines for clarity. The same result
+    could be achieved in one line.*
+
+    .. code-block:: jinja
+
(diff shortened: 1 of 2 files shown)


salt/modules/slsutil.py
```diff
@@ -246,3 +247,130 @@ def deserialize(serializer, stream_or_string, **mod_kwargs):
...
+        block.append(lgutter + line.center(textwidth) + rgutter)
+    block.append(spacer_line)
+    for line in wrapper.wrap(text):
+        block.append(lgutter + line + ' ' * (textwidth - len(line)) + rgutter)
+    block.append(border_line)
+    if blockend is not None:
+        block.append(blockend)
+
+    # Convert list to multiline string
+    result = '\n'.join(block)
```
cachedout: For better multi-platform support we should probably use `os.linesep` here instead

cachedout (changes requested):

salt/modules/slsutil.py
amendlik: That suggestion sounded right to me, but the [documentation](https://docs.python.org/3/library/os.html#os.linesep) is making me think not:

*Do not use os.linesep as a line terminator when writing files opened in text mode (the default); use a single '\n' instead, on all platforms.*

It seems that Python converts `\n` to `os.linesep` when it writes out the file.

cachedout: The files aren't ultimately opened in text mode, though. That's the thing. :)

s0undt3ch:
Port to master https://github.com/saltstack/salt/pull/56761

amendlik:
@s0undt3ch Please consider bringing #53499 along with this one. It includes a few fixes to the original PR.

s0undt3ch:
@amendlik Thank You! Will do.

Issue #49027, opened 2018-08-09
x509.certificate_managed certificate is mangled on process under py3 m2crypto


lachlanmunro:
### Description of Issue/Question
Having problems with the x509 certificate managed state under m2crypto/py3 on windows. We believe that the issue is not windows specific and is related to the charset changes in py3.

The below example is the passthrough of a certificate resulting in invalid encoding.

Although the demo state is a little contrite, we use pem_managed (we could just as easily just use file.managed?) to put ca certificates in place via the mine. The below reproduces the certificate correctly on the py2 minions that currently use states like the below.

### Setup
```
/testy.crt:
  x509.pem_managed:
    - text: |
        -----BEGIN CERTIFICATE-----
        MIIHjzCCBnegAwIBAgIIBzL2FMQfSVYwDQYJKoZIhvcNAQELBQAwVDELMAkGA1UE
        BhMCVVMxHjAcBgNVBAoTFUdvb2dsZSBUcnVzdCBTZXJ2aWNlczElMCMGA1UEAxMc
        R29vZ2xlIEludGVybmV0IEF1dGhvcml0eSBHMzAeFw0xODA3MjQxNjA4MjVaFw0x
        ODEwMDIxNjAwMDBaMGYxCzAJBgNVBAYTAlVTMRMwEQYDVQQIDApDYWxpZm9ybmlh
        MRYwFAYDVQQHDA1Nb3VudGFpbiBWaWV3MRMwEQYDVQQKDApHb29nbGUgTExDMRUw
        EwYDVQQDDAwqLmdvb2dsZS5jb20wWTATBgcqhkjOPQIBBggqhkjOPQMBBwNCAASs
        8tMhHKTNkKBHuyC9u0qbTibi9ZkpyvkFSPhBziOsLn7uDkU/PSKjHnSCswip07o9
        F0kYWilWXKKxB5w2QQ0qo4IFHDCCBRgwEwYDVR0lBAwwCgYIKwYBBQUHAwEwDgYD
        VR0PAQH/BAQDAgeAMIID4QYDVR0RBIID2DCCA9SCDCouZ29vZ2xlLmNvbYINKi5h
        bmRyb2lkLmNvbYIWKi5hcHBlbmdpbmUuZ29vZ2xlLmNvbYISKi5jbG91ZC5nb29n
        bGUuY29tghQqLmRiODMzOTUzLmdvb2dsZS5jboIGKi5nLmNvgg4qLmdjcC5ndnQy
[...]

Issue #49008, opened 2018-08-08
x509.certificate_managed state fails under py3 m2crypto with "int too large"


lachlanmunro:
### Description of Issue/Question
x509 certificate_managed fails on generated key length.

We have something like this working under py2, this change took place after an upgrade to py3 and compiling/adding m2crypto to it on windows.

### Setup
```
c:/test/test-ca.crt:
  x509.certificate_managed:
    - signing_private_key: c:/test/test-ca.key
    - CN: testy-mctest
    - basicConstraints: "critical CA:true"
    - keyUsage: "critical cRLSign, keyCertSign"
    - subjectKeyIdentifier: hash
    - authorityKeyIdentifier: keyid,issuer:always
    - days_valid: 1460
    - days_remaining: 0
    - backup: True
    - watch:
      - x509: c:/test/test-ca.key

c:/test/test-ca.key:
  x509.private_key_managed:
    - bits: 4096
    - backup: True
```

### Steps to Reproduce Issue
Running the above state results in:

```
local:
----------
          ID: c:/restic/test-ca.key
    Function: x509.private_key_managed
      Result: True
     Comment: File c:/restic/test-ca.key updated
     Started: 16:32:14.495656
    Duration: 1886.762 ms
     Changes:
              ----------
              new:
                  New private key generated
----------
          ID: c:/restic/test-ca.crt
    Function: x509.certificate_managed
      Result: False
     Comment: An exception occurred in this state: Traceback (most recent call last):
                File "C:\salt\bin\lib\site-packages\salt\state.py", line 1905, in call
                  **cdata['kwargs'])
                File "C:\salt\bin\lib\site-packages\salt\loader.py", line 1830, in wrapper
                  return f(*args, **kwargs)
                File "C:\salt\bin\lib\site-packages\salt\states\x509.py", line 496, in certificate_managed
                  new = __salt__['x509.create_certificate'](testrun=True, **kwargs)
                File "C:\salt\bin\lib\site-packages\salt\modules\x509.py", line 1440, in create_certificate
                  cert.set_serial_number(serial_number)
                File "C:\salt\bin\lib\site-packages\M2Crypto\X509.py", line 593, in set_serial_number
                  return m2.asn1_integer_set(asn1_integer, serial)
              OverflowError: Python int too large to convert to C long
     Started: 16:32:16.383417
    Duration: 5.005 ms
     Changes:
```

### Versions Report
```
Salt Version:
           Salt: 2018.3.2

Dependency Versions:
[...]

PR #49397, opened 2018-08-29
x509 fixes

isbm:
### What does this PR do?

Number of bugfixes.

- Python3: byte/unicode mistmatch on `text` parameter (crash)
- `None` type returned instead of an empty dictionary, once minions aren't matched or CA server unavailable (crash)
- Other minor issues

Test should pass as is.


terminalmage (approved):

garethgreenaway (approved):

cachedout:
Why are we modifying the publish module in a PR for x509 fixes?

isbm:
> Why are we modifying the publish module in a PR for x509 fixes?

Because people think it is cool to have [many returns inside the code](http://homepages.ihug.com.au/~ijoyner/Ian_Joyner/No_return.html) and wrap the way they do not understand what actually is going on. And then when it is expected to be at least an empty dictionary at the end, there is a corner case when it returns just `None` and then you get a crash, instead of expected no-op.

cachedout (approved):

cachedout:
@akissa or @clinta Do either of you want to review this?

dwoz:
The work done here might fix these two issues:

https://github.com/saltstack/salt/issues/49027
https://github.com/saltstack/salt/issues/49008

rnickle:
Is it easy to intercept these fixes earlier?  The PPA for Ubuntu is pointing to 2018.3.3 right now.

Issue #38209, opened 2016-12-12
Accepting a minion causes tornado to exit


limited:
### Description of Issue/Question
When running the Salt Python client inside of a tornado application, accepting a minion via Salt will terminate the global tornado IOLoop. 

### Setup
- Requires a minion in unaccepted state. 
```
[root@bf006b7b73a3 test]# salt-key -L
Accepted Keys:
Denied Keys:
Unaccepted Keys:
2f68db53cd12
Rejected Keys:
```

### Steps to Reproduce Issue
Run the below script to start the tornado server. Once started, make a curl request to the `/test/` endpoint. Tornado will exit. 

```
#!/usr/bin/python

import salt.config
import salt.key

import tornado.ioloop
import tornado.web

master_opts = salt.config.client_config('/etc/salt/master')
salt_key = salt.key.Key(master_opts)

class TestHandler(tornado.web.RequestHandler):
    def get(self):
        minions = salt_key.list_keys()
        salt_key.accept(minions['minions_pre'][0])
        self.write({'accepted_id': minions['minions_pre'][0]})

if __name__ == '__main__':
    application = tornado.web.Application([
        tornado.web.URLSpec(r"/test/",
                            TestHandler)])
    application.listen(8080)
    tornado.ioloop.IOLoop.current().start()
```

To test, run `curl http://localhost:8080/test/`

### Versions Report
Salt-Master:
```
Salt Version:
           Salt: 2016.3.4

Dependency Versions:
           cffi: Not Installed
       cherrypy: Not Installed
       dateutil: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.7.2
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.4.7
   mysql-python: Not Installed
      pycparser: Not Installed
       pycrypto: 2.6.1
         pygit2: Not Installed
         Python: 2.7.5 (default, Sep 15 2016, 22:37:39)
   python-gnupg: Not Installed
         PyYAML: 3.11
          PyZMQ: 15.3.0
           RAET: Not Installed
          smmap: Not Installed
        timelib: Not Installed
        Tornado: 4.2.1
            ZMQ: 4.1.4

System Versions:
           dist: centos 7.2.1511 Core
        machine: x86_64
        release: 4.4.27-moby
         system: Linux
[...]

PR #49398, opened 2018-08-29
[WIP] Tornado50

DmitryKuzmenko:
### What does this PR do?
Work is in progress, please ignore this for now.

Fix https://github.com/saltstack/salt-jenkins/issues/995

### Previous Behavior
Salt is non-functional on Python 3 with Tornado 5.0 

### New Behavior
Work in progress but at the end it should work better than before =)

### Tests written?
No

### Commits signed with GPG?
Yes


salt/config/__init__.py
```diff
@@ -1241,6 +1243,7 @@ def _gather_buffer_space():
     'id_function': {},
     'cachedir': os.path.join(salt.syspaths.CACHE_DIR, 'minion'),
     'append_minionid_config_dirs': [],
+    'worker_threads': 3,
```
cachedout: Are we certain about this? This default has been `5` for a long time. Could you explain this change more, please?

salt/grains/zfs.py
```diff
@@ -70,11 +70,12 @@ def _zfs_pool_data():
     return grains
 
 
-def zfs():
+def zfs(grains=None):
```
cachedout: I'm not following how this change is related to Tornado. Could you clarify, please?

salt/key.py
```diff
@@ -373,7 +373,7 @@ class Key(object):
     REJ = 'minions_rejected'
     DEN = 'minions_denied'
 
-    def __init__(self, opts, io_loop=None):
+    def __init__(self, opts):
```
cachedout: The reason we had this was to correct https://github.com/saltstack/salt/issues/38209

Are we certain that we won't regress here?

cachedout:
We should have @skizunov come by and look at this if he is available to as well.

salt/config/__init__.py
DmitryKuzmenko: It was so for master I'm not sure what's the best one for minion. I would change it to 5.

salt/grains/zfs.py
DmitryKuzmenko: There was a bug in the beginning of my work that I've fixed this way. I'll check how's it going now and will remove this if it's not needed anymore.

salt/key.py
DmitryKuzmenko: I'll check this case when I'll done here.
Anyway Tornado doesn't allow to play with IOLoops now and we shall use the current one everywhere.

rallytime:
This will need to be rebased as well when it's ready as there are some merge conflicts here now.

salt/key.py
isbm: > Anyway Tornado doesn't allow to play with IOLoops now and we shall use the current one everywhere.

But we should have the ability to use older ones.

salt/minion.py
```diff
@@ -1480,37 +1495,43 @@ def _handle_decoded_payload(self, data):
...
-                name=data['jid']
-            )
-
-        if multiprocessing_enabled:
-            with default_signals(signal.SIGINT, signal.SIGTERM):
-                # Reset current signals before starting the process in
-                # order not to inherit the current signal handlers
-                process.start()
+        if self.opts.get('ipc_mode', '') == 'tcp':
+            pull_uri = int(self.opts.get('tcp_minion_workers', 4516))
```
isbm: I think ports needs to be moved somewhere to `salt.utils.default.ports` or something similar. Just like we have exitcodes.

isbm (commented):
Keep in mind that we still need Tornado 4.x support backward compat.

salt/minion.py
DmitryKuzmenko: Agree.

salt/key.py
DmitryKuzmenko: I understand this, but If we use the only on IOLoop per thread/process, we should be OK with Tornado 4.
Anyway I'll take an additional attention on this.

isbm: @DmitryKuzmenko yeah, I think so too. But just keep in mind that. Sometimes you can have pretty old systems in the enterprise, but who want to have latest possible codebase (with outdated dependencies, of course)... :wink:

brejoc:
@DmitryKuzmenko Is there something we might try to help you with here? I'm asking because this is currently blocking the update to Python 3.7 in openSUSE Tumbleweed.

sysrich:
The lack of python 3.7 support in saltstack is leading to the consideration of salt being removed from openSUSE Tumbleweed and openSUSE Kubic - this is something which I don't want to see from both a personal and obviously political perspective, so if there is anything that can be done to expedite a solution I would greatly appreciate it.

gtmanfred:
You can use python 3.7 with 2018.3.3 and 2017.7.8.

You just cannot use tornado 5.0.

dwoz:
I've pointed this to the tornado50 branch. I'm wondering if we should close this issue (for now) and pull stuff into the tornado50 branch from here when needed?

Ch3LL:
@DmitryKuzmenko bump on @dwoz question above

DmitryKuzmenko:
@Ch3LL sorry, thanks for mention.
@dwoz yes I agree.

Issue #48792, opened 2018-07-26
Feature-Request:  add the ability to use Set-WebConfiguration command set to win_iis


whytewolf:
### Description of Issue/Question

The [win_iis](https://docs.saltstack.com/en/latest/ref/modules/all/salt.modules.win_iis.html) module needs the ability to use set-WebConfiguration from the [webadministration PowerShell module](https://docs.microsoft.com/en-us/powershell/module/webadministration/?view=win10-ps). 

The win_iis state module should also reflect this change.

This is to allow setting items like `/system.webServer/security/authentication/anonymouseAuthentication -name 'userName'` 

This will allow much wider ability to configure the web services within IIS.

twangboy:
ZD-2693

rickh563:
### Next Steps

* Discuss with Salt-Support before initiating development any development of this Feature-Request

* ZD-2693 has written a `state module` and a `custom execution module` to manage this for internal to their organization;

* It is possible that they my contribute this to the project;

* FOSS policy is under discussion;

rickh563:
See PR https://github.com/saltstack/salt/pull/49399

PR #49399, opened 2018-08-29
Add webconfiguration settings management (module + state).

tlemarchand:
### What does this PR do?
Add webconfiguration settings management, inside win_iis module ( get and set webconfigurations ) and win_iis state ( define webconfigurations ).
This is necessary to manage all possible IIS configurations.

https://github.com/saltstack/salt/issues/48792

### Tests written?
No

### Commits signed with GPG?
No

### DIFF
--- a/salt/modules/win_iis.py
+++ b/salt/modules/win_iis.py
@@ -14,6 +14,8 @@ from __future__ import absolute_import, print_function, unicode_literals
 import decimal
 import logging
 import os
+import re
+import yaml
 
 # Import salt libs
 import salt.utils.json
@@ -21,6 +23,7 @@ import salt.utils.platform
 from salt.ext.six.moves import range
 from salt.exceptions import SaltInvocationError, CommandExecutionError
 from salt.ext import six
+from salt.ext.six.moves import map
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


salt/modules/win_iis.py
```diff
@@ -160,6 +163,37 @@ def _srvmgr(cmd, return_json=False):
...
+    collection = get_webconfiguration_settings(pspath, [{'name': name, 'filter': colfilter}])[0]['value']
+    for idx, collect_dict in enumerate(collection):
+        if all(item in collect_dict.items() for item in match.items()):
+            return idx
+    return -1
+
+
+def _prepare_settings(pspath, settings):
+    '''
+    Prepare settings before execution wit get or set functions.
```
twangboy: with

twangboy (approved):

twangboy (changes requested):

salt/modules/win_iis.py
```diff
@@ -1985,3 +2019,171 @@ def set_webapp_settings(name, site, settings):
...
+
+    Returns:
+        dict: A list of dictionaries containing setting name, filter and value.
+
+    CLI Example:
+
+    .. code-block:: bash
+
+        salt '*' win_iis.get_webconfiguration_settings name='IIS:\' settings="[{'name': 'enabled', 'filter': 'system.webServer/security/authentication/anonymousAuthentication'}]"
+    '''
```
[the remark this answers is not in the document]
tlemarchand: r added line 2025?

twangboy (approved):

salt/modules/win_iis.py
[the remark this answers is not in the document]
twangboy: Oh, yeah, that's right

salt/modules/win_iis.py
```diff
@@ -160,6 +163,37 @@ def _srvmgr(cmd, return_json=False):
...
+
+
+def _prepare_settings(pspath, settings):
+    '''
+    Prepare settings before execution wit get or set functions.
+    Removes settings with a match parameter when index is not found.
+    '''
+    prepared_settings = []
+    for setting in settings:
+        match = re.search('Collection\[(\{.*\})\]', setting['name'])
```
twangboy: This line is failing lint as well... Might need to add an `r` here. Not sure how that would affect the regex though. You'll need to test.

twangboy: If it must be formatted like this, then you'll need to skip the lint test for this line by adding this to the end:
```
# pylint: disable=W1401
```
There should be two spaces between the end of the code and the `#`

twangboy (changes requested):

salt/modules/win_iis.py
tlemarchand: The backslashes here are not interpreted by Python in any way ( neither \[ nor \{  means something ). So it is safe to add the raw prefix. I tested it, it works.

twangboy (approved):

twangboy:
@rallytime Could we get this in to Fluorine? It fixes a customer issue.

twangboy:
https://github.com/saltstack/salt/issues/48792

twangboy:
@tlemarchand Great work! Thanks for the PR.

salt/modules/win_iis.py
cachedout: This was marked as resolved but doesn't seem to be. :)

cachedout (changes requested):

salt/modules/win_iis.py
tlemarchand: Good catch. Fixed now !

twangboy (approved):

cachedout (approved):

PR #49406, opened 2018-08-29
[2018.3] Merge forward from 2018.3.3 to 2018.3

rallytime:
Conflicts:
  - salt/utils/timed_subprocess.py
  - tests/integration/states/test_file.py


rallytime:
@cachedout That's because there were indeed conflicts, but on both of those files I needed to take the `HEAD` of 2018.3, so no changes would show up.

Issue #49269, opened 2018-08-22
2017 and 2018 grains.items not able to output with --out=yaml


whytewolf:
### Description of Issue/Question

Using 2017.7.x and 2018.3.x I'm unable to output with --out=yaml using salt

### Setup
Install salt and try to output the grains in yaml and run the command `salt-call grains.items --out=yaml`

### Steps to Reproduce Issue
Install salt and try to output the grains in yaml and run the command `salt-call grains.items --out=yaml`

Trace logs included in the attached file. 

[grains.items.yaml.2017.7.7.txt](https://github.com/saltstack/salt/files/2312082/grains.items.yaml.2017.7.7.txt)

### Versions Report
```
Salt Version:
           Salt: 2017.7.7

Dependency Versions:
           cffi: 1.6.0
       cherrypy: Not Installed
       dateutil: Not Installed
      docker-py: Not Installed
          gitdb: 0.6.4
      gitpython: 1.0.1
          ioflo: Not Installed
         Jinja2: 2.7.2
        libgit2: 0.26.3
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.5.6
   mysql-python: Not Installed
      pycparser: 2.14
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: 0.26.4
         Python: 2.7.5 (default, Jul 13 2018, 13:06:57)
   python-gnupg: Not Installed
         PyYAML: 3.11
          PyZMQ: 15.3.0
           RAET: Not Installed
          smmap: 0.9.0
        timelib: Not Installed
        Tornado: 4.2.1
            ZMQ: 4.1.4

System Versions:
           dist: centos 7.5.1804 Core
         locale: UTF-8
        machine: x86_64
        release: 3.10.0-862.11.6.el7.x86_64
         system: Linux
        version: CentOS Linux 7.5.1804 Core
```
```
Salt Version:
           Salt: 2018.3.2

Dependency Versions:
           cffi: 1.6.0
       cherrypy: Not Installed
       dateutil: Not Installed
      docker-py: Not Installed
          gitdb: 0.6.4
      gitpython: 1.0.1
          ioflo: Not Installed
         Jinja2: 2.7.2
        libgit2: 0.26.3
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: Not Installed
   msgpack-pure: Not Installed
[...]

PR #49408, opened 2018-08-29
Allow our custom yaml dumper to NamespacedDictWrapper objects

terminalmage:
This special dict class did not have a representer assigned to it, so PyYAML would raise an exception when attempting to dump it.

This also removes the class name overriding, as its implementation was ill-conceived and will be deprecated in a later branch (it's only used in this one place so deprecating should be of no-to-minimal impact).


rallytime (approved):
🎉

rallytime:
@terminalmage This fixes #49269, yeah?

terminalmage:
@rallytime that is correct

cachedout (approved):

PR #49411, opened 2018-08-29
Allow our custom yaml dumper to NamespacedDictWrapper objects (2018.3 branch)

terminalmage:
_(This is https://github.com/saltstack/salt/pull/49408 opened against the 2018.3 branch)_

This special dict class did not have a representer assigned to it, so PyYAML would raise an exception when attempting to dump it.

This also removes the class name overriding and deprecates the NamespacedDictWrapper's ability to do so, as it was A) incorrectly implmented and B) is no longer used by anything in Salt.


gtmanfred (approved):

rallytime (approved):

rallytime:
Refs #49269

Issue #49205, opened 2018-08-20
Compound targeting includes unexpected targets in the result


ereslibre:
### Description of Issue/Question

Compound targeting includes unexpected (even nonexistant) targets in the result.

### Setup

This was triggered by an orchestration that rejects a salt minion in the middle of the orchestration. The `tgt` used after the salt minion key is rejected included an explicit filter to not take this minion into consideration; even if salt guarantees this, it was nice to have this in a explicit way.

Consider the following minions:

```
# salt-key
Accepted Keys:
3db64d6583f548cd91336526b7733b90
6c1d7bdd4bb64196b3644429ed7f81e3
admin
ca
Denied Keys:
Unaccepted Keys:
Rejected Keys:
```

Now, we can write an orchestration:

```
do-something:
  salt.function:
    - tgt: 'not L@3db64d6583f548cd91336526b7733b90'
    - tgt_type: compound
    - name: grains.get
    - arg:
        - os
```

### Steps to Reproduce Issue

Run the orchestration. It works as expected:

```
# salt-run state.orchestrate orch.some_test
admin_master:
----------
          ID: do-something
    Function: salt.function
        Name: grains.get
      Result: True
     Comment: Function ran successfully. Function grains.get ran on admin, ca, 6c1d7bdd4bb64196b3644429ed7f81e3.
     Started: 10:40:03.119885
    Duration: 315.809 ms
     Changes:   
              admin:
                  SUSE
              ca:
                  SUSE
              6c1d7bdd4bb64196b3644429ed7f81e3:
                  SUSE

Summary for admin_master
------------
Succeeded: 1 (changed=1)
Failed:    0
------------
Total states run:     1
Total run time: 315.809 ms
```

Now, if we reject this minion:

```
# salt-key -r 3db64d6583f548cd91336526b7733b90 --include-accepted
The following keys are going to be rejected:
Accepted Keys:
3db64d6583f548cd91336526b7733b90
Proceed? [n/Y] Y
Key for minion 3db64d6583f548cd91336526b7733b90 rejected.
```

Now let's try to run the same orchestration again:

```
# salt-run state.orchestrate orch.some_test
[ERROR   ] {u'ret': {u'admin': u'SUSE', u'ca': u'SUSE', u'6c1d7bdd4bb64196b3644429ed7f81e3': u'SUSE', u'3db64d6583f548cd91336526b7733b90': False}, u'out': u'highstate'}
admin_master:
----------
          ID: do-something
    Function: salt.function
[...]

PR #49435, opened 2018-08-30
No longer passes missing in list compound engine

brejoc:
### What does this PR do?

Missing targets cause a very odd behaviour in combination with a "not" statement and don't seem to be used in any other case.

Having a compound list target like `not L@foominion` would cause a return for `foominion` even though it doesn't exist. Returning an empty list for "missing" fixes that.

https://github.com/saltstack/salt/issues/49205

### Previous Behavior

Given that there is only one minion registered called `minion1`.
```
$ salt -C 'not L@foominion' grains.get os
minion1:
    SUSE
foominion:
    Minion did not return. [No response]
```

### New Behavior
Given that there is only one minion registered called `minion1`.
```
$ salt -C 'not L@foominion' grains.get os
minion1:
    SUSE
```

### Tests written?
No

### Commits signed with GPG?

Yes


salt/utils/minions.py
```diff
@@ -224,7 +224,7 @@ def _check_list_minions(self, expr, greedy):  # pylint: disable=unused-argument
             expr = [m for m in expr.split(',') if m]
         minions = self._pki_minions()
         return {'minions': [x for x in expr if x in minions],
-                'missing': [x for x in expr if x not in minions]}
```
gtmanfred: This is not ok.

This was a specific feature added, and would rip out the behavior that shows regular list minions aren't specified.

`L@MINION_NOT_SPECIFIED` would not return that minion_not_specified is not a minion anymore.

gtmanfred (changes requested):

salt/utils/minions.py
brejoc: That is odd. With an unmodified version of Salt (2018.3.0-26.1 from SUSE) I see this:
```
salt -C 'L@foo' grains.get os
No minions matched the target. No command was sent, no jid was assigned.
ERROR: No return received
```
So you'd expect to get a return from the missing minion here or did I misunderstand you? Thanks!

brejoc:
@gtmanfred I think we found the use-case you where referring to. You where probably talking about a list like `L@minion1,foominion`, which should also return a response for `foominion` even though it is not existing. We've push an update that should reflect that.

cachedout:
@brejoc Right now, @gtmanfred is out for a little while on leave but he should be back in a few weeks to look at this. I'm not totally positive which case he's referring to so I'd rather wait and let him respond instead of guessing. Apologies for the delay.

brejoc:
Perhaps @garethgreenaway can take a look here, since he is afaik the original author of the change that introduced this behavior? Sorry for the impatience, but this is blocking the CaaSP release. Thanks!

salt/utils/minions.py
brejoc: See [new comment](https://github.com/saltstack/salt/pull/49435#issuecomment-417671177) please.

garethgreenaway (approved):

isbm (approved):

rallytime:
@brejoc This is causing some tests to fail: https://jenkinsci.saltstack.com/job/pr-kitchen-ubuntu1604-py2/job/PR-49435/5/

Can you take a look there? The matching and batch tests are related to this change.

brejoc:
@rallytime Test should be fixed now. Apart from `integration/shell/test_matcher.py::MatchTest::test_grain`. Looks like this was broken before.

cachedout:
@terminalmage Since you have been in the matchers code today, you might have an opinion on this.

terminalmage (commented):
I would wait until https://github.com/saltstack/salt/pull/48809 is merged, and then re-do this PR. The reason for this is that most of this code will no longer exist (or will exist in a different form, in a different file) once that PR is merged.

brejoc:
@terminalmage Sounds good. But are the changes from #48809 backported? Because this fixes a bug that exists down to at least `2018.3.0`.

gtmanfred:
This pr and that one is being proposed against develop.

So you will need to open a new pr against the older release to have it be backported.

rallytime:
#48809 is against fluorine as making the matchers pluggable is a new feature for fluorine.

terminalmage:
@brejoc Where possible, please open pull requests against the oldest release branch where the bug exists: https://docs.saltstack.com/en/latest/topics/development/contributing.html#which-salt-branch

In this case though we would need two PRs due to the massive difference in how matchers are architected beginning in the Fluorine release. You can open one PR against `2018.3` and another against our `fluorine` branch.

cachedout (approved):

cachedout:
@gtmanfred Re-review requested.

gtmanfred (approved):

PR #49453, opened 2018-08-30
Fix up py3 git config tests

dwoz:
### What does this PR do?

Fix issues with configparser and the default encoding on Windows.

### Tests written?

No

### Commits signed with GPG?

Yes


terminalmage (approved):
Good catch.

PR #49458, opened 2018-08-30
Clear cache for pip and virtualenv requirement files

Ch3LL:
### What does this PR do?
the tests: 

```
integration.states.test_virtualenv.VirtualenvTest.test_issue_1959_virtualenv_runas
integration.states.test_pip_state.PipStateTest.test_issue_6912_wrong_owner_requirements_file
```

are failing on macosx. I can't seem to replicate these failures anymore so this fix is a guess as to what is occurring. 

This is the error that is occuring:

```
Could not open requirements file: [Errno 13] Permission denied: '/var/folders/zz/zyxvpxvq6csfxvn_n0000000000000/T/tmpn4vhel/requirements.txt'
```

/var/folders/zz/ is a temporary location on macosx, so i'm guessing either a previous test is changing the permissions of the folder nested underneath zz, or it was previously cached under a different user. I'm trying to clear the cache so the `requirements.txt` files have to get re-created/cached.

https://github.com/saltstack/salt-jenkins/issues/1055


rallytime (approved):

Ch3LL:
k i'll go ahead and close and try to re-attack this another way.

Issue #46388, opened 2018-03-06
IPv4 & v6 network configuration on Debian


dynek:
### Description of Issue/Question
Unable to create working network configuration for DHCP IPv4 and static IPv6.
Instruction "iface eth0 inet dhcp" is missing for IPv4.

### Setup
**Simple setup for DHCP IPv4.**
SLS:
```
eth0:
  network.managed:
    - enabled: True
    - type: eth
    - proto: dhcp
```
/etc/network/interfaces content is OK:
```
auto lo
iface lo inet loopback
iface lo inet6 loopback
auto eth0
iface eth0 inet dhcp
```
**Adding Static IPv6 configuration:**
```
eth0:
  network.managed:
    - enabled: True
    - type: eth
    - proto: dhcp
    - enable_ipv6: True
    - ipv6proto: static
    - ipv6ipaddr: <ipv6ipaddr>
    - ipv6netmask: 64
    - ipv6gateway: <ipv6gateway>
```
/etc/network/interfaces content not OK (doesn't contain ```iface eth0 inet dhcp```).
Acting like "proto: dhcp" was not specified:
```
auto lo
iface lo inet loopback
iface lo inet6 loopback
auto eth0
iface eth0 inet6 static
    address <ipv6ipaddr>
    netmask 64
    gateway <ipv6gateway>
```

### Steps to Reproduce Issue
Use SLS declarations above

### Versions Report
Master version is 2017.7.2

And the minion:
```
Salt Version:                                                                                                                                                                                                                                                                                                                 
           Salt: 2017.7.3
 
Dependency Versions:
           cffi: 1.11.2
       cherrypy: Not Installed
       dateutil: 2.6.1
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.10
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.4.8
   mysql-python: Not Installed
      pycparser: 2.18
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: Not Installed
         Python: 2.7.14+ (default, Dec  5 2017, 15:17:02)
   python-gnupg: Not Installed
         PyYAML: 3.12
          PyZMQ: 16.0.2
           RAET: Not Installed
          smmap: Not Installed
        timelib: Not Installed
[...]

PR #49462, opened 2018-08-31
Some Debian network (debian_ip) refactoring

MTecknology:
### What does this PR do?

Commit a058e65 refactors the logic used in debian_ip._parse_settings_eth(). It re-works the logic used to put options into dictionaries and simplifies how some options are shared.

- Adds `-enable_ipv4:` which is meant for future deprecation of IPv4
- Makes better use of `-enable_ipv6:`
- Adds support for `-ipv4<opt>:` prefix (similar to `-ipv6<opt>:`)
- Adds support for sharing options between IPv4/v6
- Supports overriding any IPv4/v6 option
- Makes `-proto:` a required field

Commit 038261c adds new unit tests and updates/fixes existing tests. 

- Unit tests have been updated for modified behavior:
  + Make sure bond options are shared between IPv4/v6
  + Added tests for reported/discovered issues
- Added tests:
  + Check for v4/v6-only option sharing/overwriting
  + Check that netmask/gw is not added unless specified

Commit dad829e reworks the debian_eth.jinja template.

- Removes a lot of unique if's and turns them into a loop
- Removes extraneous newlines in interfaces file
- Cleans up redundancies
- Allows all options to be space-delimited list, line-delimited list, or string

- #46388
- Fixes other unreported issues

### Tests written?

Yes

### Commits signed with GPG?

Yes


MTecknology:
I have no idea how what I changed could have effected the tests that failed. Any hints?

cachedout (approved):

rallytime:
@MTecknology I think this is a pretty big change that I'd like to let sit in develop for a while.

garethgreenaway (approved):

dynek:
> @MTecknology I think this is a pretty big change that I'd like to let sit in develop for a while.

When wil this make it into a package available from http://repo.saltstack.com/apt/debian/9/amd64/?

MTecknology:
@dynek It'll show up in whatever comes after 2018.3.

dynek:
@MTecknology I missed your answer. I just checked it and I do not see any mention of neon: https://docs.saltstack.com/en/latest/topics/releases/version_numbers.html
Do you think this will make it into a version after Fluorine is released?

One more question, is this affecting server or client side? What if my server is running 2017 while clients run 2018? Just trying to grasp if I should be upgrading both or if client is sufficient.

Thank you!

rallytime:
@dynek Yes, `Neon` is the feature release after `Fluorine`. I have updated that document in #50354 to include a few more release codenames.

dynek:
Surprising that it takes so long to make it into a released version.

/me puzzled but will wait :-)

MTecknology:
There's two reasons for this particular change taking a long time to land. 1) It's a really massive change that includes non-minor changes in behavior. 2) The next feature release is getting ever closer and they're shooting for stability so major changes like this won't make it in.

MTecknology:
@dynek I guess I was wrong about those two reasons!

I think I missed the 3000.1 release, but it should wind up in the next one... this time.

dynek:
It's been a while :-) Just tested this and I confirm it fixes the problem I was having.

Thank you @MTecknology

PR #49464, opened 2018-08-31
Make opkg.del_repo be compatible with pkrepo state module

chotea:
### What does this PR do?
Pkgrepo state module is using a named argument when calling pkg.del_repo.
Change the name of the argument from 'alias' to 'repo' and update all
other functions to be consistent on naming.

### Previous Behavior
Applying a state containing pkgrepo.absent will crash on minions working with opkg module. 

### New Behavior
Now we can remove a repo using a state.

### Tests written?

No

### Commits signed with GPG?

No

### DIFF
--- a/salt/modules/opkg.py
+++ b/salt/modules/opkg.py
@@ -1168,7 +1168,7 @@ def list_repos(**kwargs):  # pylint: disable=unused-argument
     return repos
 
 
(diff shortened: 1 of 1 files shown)


cachedout (approved):

cachedout:
@adelcast do you have any comments here by chance?

adelcast:
@cachedout , nope it looks good to me. Thanks for checking!

PR #49477, opened 2018-08-31
Check if service exists first in service.status for mac

Ch3LL:
### What does this PR do?
this test `integration.modules.test_mac_service.MacServiceModuleTest.test_status` is failing because we are now returning a CommandExecutionError now instead of just `''` if a service does not exist like previous. This will change it back the previous behavior to return empty string and log an error that the service does not exist.

ping @weswhet can you review?


rallytime (approved):

weswhet:
LGTM!

Issue #48773, opened 2018-07-25
salt 2018.3.2 py3 on Ubuntu 18.04 takes a long time to run test.true


kartiksubbarao:
I'm running salt 2018.3.2+ds-1 with python 3 on Ubuntu 18.04, and am finding that it takes over 5 seconds to run `salt-call --local test.true`. If I use `--skip-grains`, it cuts the time down to 1 second (which is still somewhat slower than expected). While searching around I came across #26255, but in my case the number of file descriptors isn't the issue -- setting ulimit -n 1024 doesn't fix it. Here's the output from both runs. The numerous `Depends decorator instantiated...` messages seem to chew up most of the time.

### `time salt-call --local test.true -l trace`
```
[DEBUG   ] Reading configuration from /etc/salt/minion
[DEBUG   ] Changed git to gitfs in minion opts' fileserver_backend list
[DEBUG   ] Using cached minion ID from /etc/salt/minion_id: [...]
[TRACE   ] The required configuration section, 'fluent_handler', was not found the in the configuration. Not loading the fluent logging handlers module.
[TRACE   ] None of the required configuration sections, 'logstash_udp_handler' and 'logstash_zmq_handler', were found in the configuration. Not loading the Logstash logging handlers module.
[DEBUG   ] Configuration file path: /etc/salt/minion
[WARNING ] Insecure logging configuration detected! Sensitive data may be logged.
[DEBUG   ] Grains refresh requested. Refreshing grains.
[DEBUG   ] Reading configuration from /etc/salt/minion
[TRACE   ] 'drill' could not be found in the following search path: '['/usr/local/sbin', '/usr/local/bin', '/usr/sbin', '/usr/bin', '/sbin', '/bin', '/snap/bin']'
[TRACE   ] 'esxcli' could not be found in the following search path: '['/usr/local/sbin', '/usr/local/bin', '/usr/sbin', '/usr/bin', '/sbin', '/bin', '/snap/bin']'
[TRACE   ] Depends decorator instantiated with dep list of (False,)
[TRACE   ] Depends decorator instantiated with dep list of (False,)
[TRACE   ] Depends decorator instantiated with dep list of (False,)
[TRACE   ] Depends decorator instantiated with dep list of (False,)
[TRACE   ] Depends decorator instantiated with dep list of (False,)
[TRACE   ] Depends decorator instantiated with dep list of (False,)
[TRACE   ] Depends decorator instantiated with dep list of (False,)
[...]

PR #49481, opened 2018-09-01
Add parameter named grains_blacklist

rongzeng54:
### What does this PR do?
Add parameter named grains_blacklist to block the specific grains.

No

### Previous Behavior
The default grains list, seems getting longer and uncontrolled by the user.

### New Behavior
The user finally has a chance to say no.
Of course, only for advanced users because of the dependencies of module.

### Tests written?
No

### Commits signed with GPG?
No

### DIFF
--- a/doc/ref/configuration/minion.rst
+++ b/doc/ref/configuration/minion.rst
@@ -752,6 +752,30 @@ Statically assigns grains to the minion.
       cabinet: 13
       cab_u: 14-15
 
+.. conf_minion:: grains_blacklist
+
+``grains_blacklist``
+----------
+
+Default: ``[]``
+
+If :conf_minion:`grains_blacklist` is enabled then any core grains will be
+checked against the matches found in the `grains_blacklist` list and any
+matched core grains will be filtered.
+
+.. note::
+
+    Some modules may depends on core grains. Filtering may cause
+    the dependent module to run unusual.
+
+.. versionadded:: 2018.11.0
+
+.. code-block:: yaml
+
+    grains_blacklist:
+      - cpu_flags
+      - zmq*
+
 .. conf_minion:: grains_cache
 
 ``grains_cache``
--- a/doc/ref/configuration/minion.rst
+++ b/doc/ref/configuration/minion.rst
@@ -755,26 +755,26 @@ Statically assigns grains to the minion.
 .. conf_minion:: grains_blacklist
 
 ``grains_blacklist``
-----------
+--------------------
 
 Default: ``[]``
 
-If :conf_minion:`grains_blacklist` is enabled then any core grains will be
-checked against the matches found in the `grains_blacklist` list and any
-matched core grains will be filtered.
+Each grains key will be compared against each of the expressions in this list.
+Any keys which match will be filtered from the grains. Exact matches, glob
+matches, and regular expressions are supported.
 
 .. note::
+    Some states and execution modules depend on grains. Filtering may cause
+    them to be unavailable or run unreliably.
 
-    Some modules may depends on core grains. Filtering may cause
-    the dependent module to run unusual.
-
-.. versionadded:: 2018.11.0
+.. versionadded:: Neon
 
 .. code-block:: yaml
 
     grains_blacklist:
       - cpu_flags
       - zmq*
+      - ipv[46]
 
 .. conf_minion:: grains_cache
 
--- a/salt/loader.py
+++ b/salt/loader.py
@@ -756,6 +756,8 @@ def grains(opts, force_refresh=False, proxy=None):
             bkeys = set(ret.keys()).intersection(blist)
             for bkey in bkeys:
                 del ret[bkey]
+            if not ret:
+                continue
         if grains_deep_merge:
             salt.utils.dictupdate.update(grains_data, ret)
         else:
--- a/salt/loader.py
+++ b/salt/loader.py
@@ -736,6 +736,11 @@ def grains(opts, force_refresh=False, proxy=None):
         opts['grains'] = {}
 
     grains_data = {}
+    is_filter = False
+    blist = opts.get('grains_blacklist', None)
+    if isinstance(blist, list):
+        is_filter = True
+        blist = set(blist)
     funcs = grain_funcs(opts, proxy=proxy)
     if force_refresh:  # if we refresh, lets reload grain modules
         funcs.clear()
@@ -747,6 +752,10 @@ def grains(opts, force_refresh=False, proxy=None):
         ret = funcs[key]()
         if not isinstance(ret, dict):
             continue
+        if is_filter:
+            bkeys = set(ret.keys()).intersection(blist)
+            for bkey in bkeys:
+                del ret[bkey]
         if grains_deep_merge:
             salt.utils.dictupdate.update(grains_data, ret)
         else:
@@ -1106,7 +1115,6 @@ class LazyLoader(salt.utils.lazy.LazyDict):
         threadsafety = not opts.get('multiprocessing')
         self.context_dict = salt.utils.context.ContextDict(threadsafe=threadsafety)
         self.opts = self.__prep_mod_opts(opts)
-        self._filter_grains()
 
         self.module_dirs = module_dirs
         self.tag = tag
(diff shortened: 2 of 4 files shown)


cachedout:
Overall, I like the idea of this feature but I'm not certain about this implementation.

Would it not be more efficient to just filter at the point that the grains themselves are generated versus doing this on the instantiation of the LazyLoader? I wonder if we should move this logic to the `grains()` function in the loader. Let me know what you think, @rongzeng54

rongzeng54:
@cachedout 
Thank you for your reminder. Please check the new commit.

cachedout (approved):

cachedout:
This is much better. I like this a lot.

cachedout:
@terminalmage You feel good about this?

terminalmage (changes requested):
This needs to have documentation added in `doc/ref/configuration/minion.rst`.

In addition, all other whitelists/blacklists that we support in salt use `salt.utils.stringutils.expr_match()` to support exact matches, `fnmatch.fnmatch()` matches, and regex matches. We should be using it here as well for consistency.

Lastly, rather than doing type checking, we should be setting a default value for this in `DEFAULT_MINION_OPTS` in `salt/config/__init__.py`. For example:

```python
'grains_blacklist': [],
```

There should also be an entry for this option in `VALID_OPTS` as well, with the valid type being `list`.

doc/ref/configuration/minion.rst
```diff
@@ -752,6 +752,30 @@ Statically assigns grains to the minion.
...
+If :conf_minion:`grains_blacklist` is enabled then any core grains will be
+checked against the matches found in the `grains_blacklist` list and any
+matched core grains will be filtered.
+
+.. note::
+
+    Some modules may depends on core grains. Filtering may cause
+    the dependent module to run unusual.
+
+.. versionadded:: 2018.11.0
```
terminalmage: The version for this should be `Neon`. We haven't decided on a release number for the upcoming `Fluorine` release, but we've already passed the feature freeze for it anyway.

doc/ref/configuration/minion.rst
```diff
@@ -752,6 +752,30 @@ Statically assigns grains to the minion.
...
+----------
+
+Default: ``[]``
+
+If :conf_minion:`grains_blacklist` is enabled then any core grains will be
+checked against the matches found in the `grains_blacklist` list and any
+matched core grains will be filtered.
+
+.. note::
+
```
terminalmage: This blank line is unnecessary.

doc/ref/configuration/minion.rst
```diff
@@ -752,6 +752,30 @@ Statically assigns grains to the minion.
...
+    Some modules may depends on core grains. Filtering may cause
+    the dependent module to run unusual.
```
terminalmage: This should be re-worded like this:

```
    Some states and execution modules depend on grains. Filtering may cause
    them to be unavailable or run unreliably.
```

doc/ref/configuration/minion.rst
```diff
@@ -752,6 +752,30 @@ Statically assigns grains to the minion.
       cabinet: 13
       cab_u: 14-15
 
+.. conf_minion:: grains_blacklist
+
+``grains_blacklist``
+----------
```
terminalmage: The dashed line here needs to be the same length as the previous line to avoid a Sphinx warning.

salt/loader.py
```diff
@@ -745,10 +742,12 @@ def grains(opts, force_refresh=False, proxy=None):
         ret = funcs[key]()
         if not isinstance(ret, dict):
             continue
-        if is_filter:
-            bkeys = set(ret.keys()).intersection(blist)
-            for bkey in bkeys:
-                del ret[bkey]
+        if blist:
+            for key in ret.keys():
```
terminalmage: This will break on Python 3:

```python
>>> x = {'foo': 'bar', 'abc': 123}
>>> for key in x.keys():
...     del x[key]
...
RuntimeError: dictionary changed size during iteration
>>>
```

This is because on Python 3, `dict.keys()` is a view and not a separate list like in Python 2.

To fix this, use `list(ret)` instead of `ret.keys()`.

doc/ref/configuration/minion.rst
```diff
@@ -752,6 +752,30 @@ Statically assigns grains to the minion.
...
+
+Default: ``[]``
+
+If :conf_minion:`grains_blacklist` is enabled then any core grains will be
+checked against the matches found in the `grains_blacklist` list and any
+matched core grains will be filtered.
```
terminalmage: We don't need to use the `:conf_minion:` Sphinx role here, linking to the same config option is redundant. We also should mention that globs and regex matches are both supported. How about this?

```
Each grains key will be compared against each of the expressions in this list.
Any keys which match will be filtered from the grains. Exact matches, glob
matches, and regular expressions are supported.
```

doc/ref/configuration/minion.rst
```diff
@@ -752,6 +752,30 @@ Statically assigns grains to the minion.
...
+    Some modules may depends on core grains. Filtering may cause
+    the dependent module to run unusual.
+
+.. versionadded:: 2018.11.0
+
+.. code-block:: yaml
+
+    grains_blacklist:
+      - cpu_flags
+      - zmq*
```
terminalmage: Since regexes are supported, a regex example should be added here as well.

salt/loader.py
```diff
@@ -786,6 +785,14 @@ def grains(opts, force_refresh=False, proxy=None):
             continue
         if not isinstance(ret, dict):
             continue
+        if blist:
+            for key in ret.keys():
```
terminalmage: Same here, this should also be `list(ret)` instead of `ret.keys()`.

terminalmage (changes requested):

rongzeng54:
Hi @terminalmage 
Please check again.

isbm (commented):
Why this is even needed at all? If there is something in the default list, then it needs to be there _anyway_. If you want only specific keys, then you can already get them. But once you need something that you normally do not need and this "muter" exist, then you have a trouble.

I find this useless, sorry.

isbm:
> @isbm
> Useless data wastes server-side storage space, bandwidth caused by data synchronization, even speed.
> As described, It is not for everyone.

Then simply don't ask for all data. Ask only for what you need. But once you _do need_ it but at the same time _cannot get it_, then I find this a very bad idea.

rongzeng54:
> > @isbm
> > Useless data wastes server-side storage space, bandwidth caused by data synchronization, even speed.
> > As described, It is not for everyone.
> 
> Then simply don't ask for all data. Ask only for what you need. But once you _do need_ it but at the same time _cannot get it_, then I find this a very bad idea.

The following is the configuration of our production environment, we believe that we cannot need them, but we also believe that others may need.
```
grains_blacklist:
  - 'lsb_distrib_codename'
  - 'lsb_distrib_id'
  - 'lsb_distrib_release'
  - 'cpu_flags'
  - 'fqdn_ip4'
  - 'fqdn_ip6'
  - 'num_gpus'
  - 'gpus'
  - 'locale_info'
  - 'machine_id'
  - 'pythonpath'
  - 'saltversioninfo'
  - 'pythonversion'
  - 'server_id'
  - 'saltpath'
  - 'domain'
  - 'localhost'
  - 'ip4_interfaces'
  - 'ip6_interfaces'
  - 'zmqversion'
  - 'SSDs'
  - 'disks'
  - 'mdadm'
  - 'username'
  - 'groupname'
  - 'uid'
  - 'gid'
  - 'systemd'
```

rongzeng54:
Hi @terminalmage 
It seems that all the problems have been solved.

rallytime:
Can you write some tests for this new functionality? We wouldn't want this to regress in the future. :)

rongzeng54:
> Can you write some tests for this new functionality? We wouldn't want this to regress in the future. :)

Hi @rallytime . It is done.

rallytime:
Thanks @rongzeng54!

terminalmage (approved):

max-arnold:
This option seems to partially duplicate the existing `disable_grains` parameter (briefly mentioned in #42567).

IMO, having two different ways to do this is confusing.

max-arnold:
@rongzeng54 I plan to open a PR to revert this change and instead document existing loader options (see the comment above). For example, you can prevent loading of expensive `esxi` grain module by adding `disable_grains: [esxi]` option.  This option is not exactly the same as yours in terms of granularity (it works on a module level), but it can have positive performance impact, see https://github.com/saltstack/salt/issues/48773#issuecomment-443847089

Do you have any objections against that?

Issue #49203, opened 2018-08-20
mistake in file.not_cached documentation?


jose1711:
### Description of Issue/Question
https://docs.saltstack.com/en/latest/ref/states/all/salt.states.file.html mentions:

> salt.states.file.not_cached(name, saltenv=u'base')
Ensures that a file is saved to the minion's cache. This state is primarily invoked by other states to ensure that we do not re-download a source file if we do not need to.

Note that this is the same text as for ```salt.states.file.cached```. Where is the difference then?

I suspect there's a problem in documentation and it should be fixed.

gtmanfred:
Yup, looks like it was just copied and pasted and not updated.

Thanks for reporting, we would appreciate a PR to fix the docs.

Otherwise I am marking it as a documentation error.

Thanks!
Daniel

Issue #48778, opened 2018-07-26
Documenation suggests use of md5 password


nodermatt:
### Description of Issue/Question
The latest salt documentation for user states [0] suggests indirectly to use md5 for password hashes. In the section "password" it states that openssl passwd -1 can be used to create a has, however, checking the man pages for openssl password revealed that this uses md5 to hash the password.

This is confusing because the salt-minion/salt-master agent explicitly mention not to use md5 (with reason!):
`Jul 26 08:21:03 server1117 salt-minion[6702]: [WARNING ] IMPORTANT: Do not use md5 hashing algorithm! Please set "hash_type" to SHA256 in Salt Minion config!
`

### Fix
Suggest to use sha256 to hash password in the documentation. I can confirm that it is possible to use sha-256 hashes that can be generated using `mkpasswd -m sha-256 <password>`. Note that mkpasswd can be installed with the whois package.

### Versions Report
```
Salt Version:
           Salt: 2015.8.8

Dependency Versions:
         Jinja2: 2.8
       M2Crypto: Not Installed
           Mako: Not Installed
         PyYAML: 3.11
          PyZMQ: 15.2.0
         Python: 2.7.12 (default, Dec  4 2017, 14:50:18)
           RAET: Not Installed
        Tornado: 4.2.1
            ZMQ: 4.1.4
           cffi: Not Installed
       cherrypy: 3.5.0
       dateutil: 2.4.2
          gitdb: 0.6.4
      gitpython: 1.0.1
          ioflo: Not Installed
        libgit2: Not Installed
        libnacl: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.4.6
   mysql-python: Not Installed
      pycparser: Not Installed
       pycrypto: 2.6.1
         pygit2: Not Installed
   python-gnupg: Not Installed
          smmap: 0.9.0
        timelib: Not Installed

System Versions:
           dist: Ubuntu 16.04 xenial
        machine: x86_64
        release: 4.4.0-96-generic
         system: Ubuntu 16.04 xenial
```

[0] (https://docs.saltstack.com/en/latest/ref/states/all/salt.states.user.html)

Ch3LL:
seems you have a solid grasp of the changes that need to be made, want to submit a PR? also great find :)

nodermatt:
@Ch3LL I'll be happy to make the changes and submit a PR in the next couple of days.

Thanks for the encouragement :)

MTecknology:
[...]

Issue #46053, opened 2018-02-15
dns_check doesn't do what it says


JeffAbrahamson:
### Setup

Using the example from the [documentation](https://docs.saltstack.com/en/latest/topics/jinja/index.html#dns-check):

     {{ 'www.google.com' | dns_check }}

leads to an error (see below).  The following parses, but leads to a different error (see below).

     {{ 'www.google.com' | dns_check(port='443') }}

### Steps to Reproduce Issue

In the first case, as noted in the documentation, the following error is emitted:

```
[CRITICAL] Rendering SLS 'base:firewall' failed: Jinja error: dns_check() takes at least 2 arguments (1 given)
Traceback (most recent call last):
  File "/usr/lib/python2.7/dist-packages/salt/utils/templates.py", line 418, in render_jinja_tmpl
    output = template.render(**decoded_context)
  File "/usr/lib/python2.7/dist-packages/jinja2/environment.py", line 989, in render
    return self.environment.handle_exception(exc_info, True)
  File "/usr/lib/python2.7/dist-packages/jinja2/environment.py", line 754, in handle_exception
    reraise(exc_type, exc_value, tb)
  File "<template>", line 45, in top-level template code
TypeError: dns_check() takes at least 2 arguments (1 given)
```

In the second case, the code parses but results in the following output:

    Attempt to resolve address 'www.google.com' failed. Invalid or unresolveable address

even though a tcp connection to that host and port is accepted:

```
$ telnet www.google.com 443
Trying 216.58.209.228...
Connected to www.google.com.
Escape character is '^]'.
^]
telnet> c
Connection closed.
$ 
```

### Versions Report
(Provided by running `salt --versions-report`. Please also mention any differences in master/minion versions.)

```
t$ salt-call --versions-report
Salt Version:
           Salt: 2017.7.3
 
Dependency Versions:
           cffi: Not Installed
       cherrypy: Not Installed
       dateutil: 2.4.2
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.8
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: 1.0.3
   msgpack-pure: Not Installed
 msgpack-python: 0.4.6
   mysql-python: Not Installed
      pycparser: Not Installed
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: Not Installed
[...]

PR #49482, opened 2018-09-02
Various documentation updates

MTecknology:
### What does this PR do?

These commits fix documentation mistakes reported by users.

- #49203
- #48778
- #46053

### New Behavior

No, documentation correction.

### Tests written?

No

### Commits signed with GPG?

Yes


doc/topics/jinja/index.rst
```diff
@@ -1473,6 +1476,15 @@ Returns:
...
 ------------
 
+.. jinja_ref:: connection_check
+
+``connection_check``
+-------------
+
+.. versionadded:: Fluorine
+
+Return the ip resolved by dns. This is an alias of `dns_check`.
```
cachedout: `IP` and `DNS`, please.

salt/utils/network.py
```diff
@@ -1846,8 +1846,16 @@ def refresh_dns():
...
+def connection_check(addr, port=80, safe=False, ipv6=None):
+    '''
+    Provides a convenient alias for the dns_check filter.
+    '''
+    return dns_check(addr, port, safe, ipv6)
+
+
 @jinja_filter('dns_check')
-def dns_check(addr, port, safe=False, ipv6=None):
+def dns_check(addr, port=80, safe=False, ipv6=None):
```
cachedout: What's the reasoning behind defaulting this to port 80?

cachedout: Cool. That makes sense. This shouldn't cause any change in current behavior so I think we're fine here. Thanks, @MTecknology

cachedout (approved):

doc/topics/jinja/index.rst
```diff
@@ -1473,6 +1476,15 @@ Returns:
 File filters
 ------------
 
+.. jinja_ref:: connection_check
+
+``connection_check``
+-------------
```
rallytime: Can you make the `___` line match the number of characters in the line above? Otherwise a warning will be generated in the doc build.

doc/topics/jinja/index.rst
```diff
@@ -1473,6 +1476,15 @@ Returns:
...
+
+.. versionadded:: Fluorine
```
rallytime: This should be `Neon` now. :)

doc/topics/jinja/index.rst
```diff
@@ -1473,6 +1476,15 @@ Returns:
...
+
+Return the IP resolved by DNS. This is an alias of `dns_check`.
```
rallytime: The `dns_check` needs double backtcks instead of single backticks in order to display that text like you're intending. RST syntax like 2 ticks, whereas markdown can use one or two.

rallytime (commented):
I have some very picky changes on this, if you wouldn't mind updating it. :D

Thanks for this addition and for the doc updates!

salt/utils/network.py
```diff
@@ -1861,7 +1869,7 @@ def dns_check(addr, port, safe=False, ipv6=None):
     family = socket.AF_INET6 if ipv6 else socket.AF_INET if ipv6 is False else socket.AF_UNSPEC
     try:
         refresh_dns()
-        hostnames = socket.getaddrinfo(addr, port, family, socket.SOCK_STREAM)
+        hostnames = socket.getaddrinfo(addr, int(port), family, socket.SOCK_STREAM)
```
cachedout: What's the reasoning for this? Is this just a safety check or is there something specific you're concerned about?

MTecknology: I originally added this because of @Ch3LL's [comment](https://github.com/saltstack/salt/issues/46053#issuecomment-366271772) about ensuring port is an int. It felt like an arbitrary and easily solved limitation. However, after looking at the documentation for ``socket.getaddrinfo()``, I realized there are times where someone might want to pass a string (i.e. port='http'), so I reverted the change. Thanks for pointing it out!

rallytime (approved):

MTecknology:
@waynew This is another PR that should not find it's way into ``master``. The function in question was /very/ heavily modified. Additionally, this introduces test failures with new tests from ``master`` and investigation suggests the problem is these changes, not the tests.

If we find that the original issue still exists (bug reporter prodded), then we will need to find an alternative solution.

MTecknology:
@waynew [ignore the previous note from me] :)

PR #49486, opened 2018-09-02
add cmd_blacklist_regex and cmd_whitelist_regex

rongzeng54:
### What does this PR do?
Although we now have some ways to limit the dangerous commands, there seems to be a lack of a convenient and efficient way to really solve this problem.

No

### Previous Behavior
Maybe cmd_blacklist_glob and cmd_whitelist_glob can solve the dangerous command problems on the client side.
And publisher_acl can solve the dangerous command problems for non-root user on the server side.

### New Behavior
Now cmd_blacklist_regex and cmd_whitelist_regex can solve the dangerous command problems on the server side.

### Tests written?
No

### Commits signed with GPG?
No

### DIFF
--- a/salt/client/__init__.py
+++ b/salt/client/__init__.py
@@ -1684,7 +1684,7 @@ class LocalClient(object):
             )
             raise SaltClientError
 
-        self._check_cmd_avail(fun, tgt, expr_form, arg)
+        self._check_cmd_avail(fun, tgt, tgt_type, arg)
 
         payload_kwargs = self._prep_pub(
                 tgt,
@@ -1793,7 +1793,7 @@ class LocalClient(object):
             )
             raise SaltClientError
 
-        self._check_cmd_avail(fun, tgt, expr_form, arg)
+        self._check_cmd_avail(fun, tgt, tgt_type, arg)
 
         payload_kwargs = self._prep_pub(
                 tgt,
@@ -1860,7 +1860,7 @@ class LocalClient(object):
         raise tornado.gen.Return({'jid': payload['load']['jid'],
                                   'minions': payload['load']['minions']})
 
-    def _check_cmd_avail(self, fun, tgt, expr_form, arg):
+    def _check_cmd_avail(self, fun, tgt, tgt_type, arg):
         '''
         Check to see if the given command can be run
         '''
--- a/salt/client/__init__.py
+++ b/salt/client/__init__.py
@@ -1684,6 +1684,8 @@ class LocalClient(object):
             )
             raise SaltClientError
 
+        self._check_cmd_avail(fun, tgt, expr_form, arg)
+
         payload_kwargs = self._prep_pub(
                 tgt,
                 fun,
@@ -1791,6 +1793,8 @@ class LocalClient(object):
             )
             raise SaltClientError
 
+        self._check_cmd_avail(fun, tgt, expr_form, arg)
+
         payload_kwargs = self._prep_pub(
                 tgt,
                 fun,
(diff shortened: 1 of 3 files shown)


cachedout:
Hi @rongzeng54 

Thanks for this.

We've tried to address this concern with the Publisher ACL system:

https://docs.saltstack.com/en/develop/ref/publisheracl.html

Could you explain in more detail why that system isn't sufficient for these cases so that we can better understand what your concern is? Thanks. :)

cachedout:
In this case, though, couldn't the root user simply change the configuration to whitelist whatever she wants?

rongzeng54:
Hi @cachedout 
Yes, root means all.
Many times, we are only trying to raise the threshold for making mistakes, not for preventing enemies.
The fact that can't be ignored is that there are still a lot of people using root user just for convenience, we need to prevent them from making mistakes, especially when they think they have some perfect security policies like Publisher ACL system.

terminalmage (commented):
I don't really like the idea of limiting a blacklist/whitelist to a regex. This does not match the functionality used for every other blacklist/whitelist we have in Salt, in which exact matches, `fnmatch.fnmatch()` pattern matches, and regex matches are all supported under the same config option via `salt.utils.stringutils.expr_match()`.

rongzeng54:
Hi @terminalmage 
Just like we have both file_ignore_regex and file_ignore_glob, the important thing is not the designer's preference, but the user's needs.
Regex is more powerful in helping salt users filter out the command what they think is dangerous.

rongzeng54:
Hi @terminalmage 
Because the salt user may enter any command by the cmdmod module.
It seems that if we don't use the regex, we can't easily configure them.
Below is a partial configuration of our produce server.
If we don't use regex, we need to write a lot of commands to replace it.
```
cmd_blacklist_regex:
  - '.*(/|\b)rm\s+-[rf]+\s+/($|\b)'
  - '.*(/|\b)reboot($|\s+-.*)'
  - '.*(/|\b)shutdown($|\s+-.*)'
  - '.*(/|\b)rpm\s+-e\s+.*'
  - '.*(/|\b)yum.*\s+(remove|groupremove|erase)\s+.*'
  - '.*(/|\b)init\s+[0-9]($|\b)'
  - '.*(/|\b)find\s+/\s+.*-exec\s+rm\s+-.*'
```

rongzeng54:
Hi @cachedout  @terminalmage 
If i modify them to cmd_blacklist and cmd_whitelist, and let them match the functionality used for every other blacklist/whitelist we have in Salt, in which exact matches, fnmatch.fnmatch() pattern matches, and regex matches are all supported under the same config option via salt.utils.stringutils.expr_match().
Can you accept? or i need close this pr.

isbm (commented):
@rongzeng54  It is limited to only regex. :disappointed: It should be _at least_ the same as other functionality we have in Salt.

rongzeng54:
Hi @terminalmage 
Now it supports.

salt/client/__init__.py
```diff
@@ -1684,6 +1684,8 @@ def pub(self,
             )
             raise SaltClientError
 
+        self._check_cmd_avail(fun, tgt, expr_form, arg)
```
rallytime: `expr_form` has been deprecated and removed. This should be using `tgt_type` instead now.

rallytime (changes requested):
The comment I left below is applicable to any where `expr_form` was used. All of those references need to be adjusted.

Can you also write some tests for this new functionality?

isbm:
I took a look at this again. And this whole idea still looks to me like a duct-tape around questionable quality of production environment handling. I don't like this, to be honest, as it just makes Salt very messy. We have ACL for this. We should stick to that part. If something there is missing, we should look there.

But this feature is trying to patch certain commands over the wheel role by trying to "prevent mistakes", which in reality makes its use simply incommodious when actually needed.

Why not just better organise ACL roles and make sure root is used properly? I also don't like that we are keep continue developing this and moved to writing tests, while last question to @cachedout was still ignored. I'd say, we should look at ACL issues and make delegation better instead of developing cost next to nothing shortcuts that belongs to the ACL domain, but aren't part of it.

isbm (changes requested):
I took a look at this again. And this whole idea still looks to me like a duct-tape around questionable quality of production environment handling. I don't like this, to be honest, as it just makes Salt very messy. We have ACL for this. We should stick to that part. If something there is missing, we should look there.

But this feature is trying to patch certain commands over the wheel role by trying to "prevent mistakes", which in reality makes its use simply incommodious when actually needed. I don't find claim "We use root for convenience" as a valid reason to mess up our ACL with some strange another quick options, sorry.

Why not just better organise ACL roles and make sure root is used properly? I also don't like that we are keep continue developing this and moved to writing tests, while last question by @cachedout was still ignored to be answered.

I'd say, we should stop here, but look at ACL issues instead and make delegation better. But developing cost next to nothing shortcuts that belongs to the ACL domain but aren't part of it, is something odd to me from the usability and integrity perspective.

PR #44573, opened 2017-11-16
Adding the ability to specify requisites without a state module name

garethgreenaway:
### What does this PR do?
Adding the ability to specify requisites without a state module name.

### What issues does this PR fix or reference?
#44446 

### Previous Behavior
Requisites such as prereq, listen, onfail, etc. required the state module to be included.

### New Behavior
If the state module is not specified then we search for the appropriate state by `id` or `name`.

### Tests written?
Yes

### Commits signed with GPG?
Yes

Please review [Salt's Contributing Guide](https://docs.saltstack.com/en/latest/topics/development/contributing.html) for best practices.

See GitHub's [page on GPG signing](https://help.github.com/articles/signing-commits-using-gpg/) for more information about signing commits with GPG.

PR #49488, opened 2018-09-02
Allow service.mod_watch reload parameter to be a list of requisites. …

deuscapturus:
### What does this PR do?

Add new behavior to service.mod_watch reload parameter.  Change is non-breaking.

```
test-service:
  service.running:
    - name: test
    - reload:
      - file: application-config-file
    - watch:
      - file: application-executable
      - file: application-service-file
      - file: application-config-file
```

There are situations where you need a running service to reload or restart depending on which watched id changes.  Currently a service can only reload or restart for all watched changes.

### Previous Behavior

Consider this example.  Here I have an application that is configured to accept a reload signal to reload its configuration file.  However; I still need to perform a normal restart if `application-executable` or `application-service-file` changes.  Sadly, in any case, the service will only reload.
```
application-executable:
  file.managed:
    - name: /root/test.py
    - contents: |
        #!/usr/bin/python3
        import time, signal
        def reload_config(signum, frame):
            load_config()
        def load_config():
            print("loading configuration")
            global config
            config = open('/etc/test.config', 'r').read()
        load_config()
        signal.signal(signal.SIGHUP, reload_config)
        while True:
            time.sleep(2)
            print("{}".format(config))
application-config-file:
  file.managed:
    - name: /etc/test.config
    - contents: |
        SETTING=VALUE
application-service-file:
  file.managed:
    - name: /etc/systemd/system/test.service
    - contents: |
        [Service]
        Type=simple
        ExecStart=/usr/bin/python3 /root/test.py
        ExecReload=/usr/bin/kill -HUP $MAINPID
        
        [Install]
        WantedBy=multi-user.target
test-service:
  service.running:
    - name: test
    - reload: True
    - watch:
      - file: application-executable
      - file: application-service-file
      - file: application-config-file
```

### New Behavior

I can now get the behavior I want.  If `application-executable` or `application-service-file` changes the application will restart.  But if only `application-config-file` changes the application will reload.

```
application-executable:
  file.managed:
    - name: /root/test.py
    - contents: |
        #!/usr/bin/python3
        import time, signal
        def reload_config(signum, frame):
            load_config()
        def load_config():
            print("loading configuration")
            global config
            config = open('/etc/test.config', 'r').read()
        load_config()
        signal.signal(signal.SIGHUP, reload_config)
        while True:
            time.sleep(2)
            print("{}".format(config))
application-config-file:
  file.managed:
    - name: /etc/test.config
    - contents: |
        SETTING=NEWVALUE
application-service-file:
  file.managed:
    - name: /etc/systemd/system/test.service
    - contents: |
        [Service]
        Type=simple
        ExecStart=/usr/bin/python3 /root/test.py
        ExecReload=/usr/bin/kill -HUP $MAINPID
        
        [Install]
        WantedBy=multi-user.target
test-service:
  service.running:
    - name: test
    - reload:
      - file: application-config-file
    - watch:
      - file: application-executable
      - file: application-service-file
      - file: application-config-file
```

### Tests written?

No

### Commits signed with GPG?

Yes


terminalmage (changes requested):

rallytime:
Hey @deuscapturus did you see the feedback above?

cachedout:
@terminalmage I just went ahead and applied your requested change and then resolved a merge conflict. Can you come back by here and re-review this please?

terminalmage (commented):
There is some function shadowing I'm not thrilled with here, some that predated this PR and some that was added by it. I'll push some changes to this PR to take care of them.

terminalmage (approved):

terminalmage:
@cachedout can you sanity-check the changes I pushed?

deuscapturus:
Thank you. Paternity leave has kept me away.  I'm happy to see this still going forward.

terminalmage:
No problem, and congratulations!

salt/states/service.py
```diff
@@ -825,7 +826,6 @@ def unmasked(name, runtime=False):
 def mod_watch(name,
               sfun=None,
               sig=None,
-              reload=False,
```
cachedout: I am not following the reasoning for removing this from the list of defined kwargs. Could you help me understand please, @deuscapturus ?

terminalmage: @cachedout that was actually me. I removed it because it shadows a built-in. Instead, we pop it out of the kwargs below.

garethgreenaway (approved):

max-arnold:
It looks like this change does not support short requisites (without a module name): #44573

Why can't we have features in Salt that work consistently in all cases?

PR #49492, opened 2018-09-04
Fix AMD GPU vendor detection

mkuron:
### What does this PR do?

It fixes the detection of the vendor of AMD GPUs for the GPU grains. Previously the vendor would be reported as _unknown_.

No issue created for this.

### Previous Behavior
```
$ salt lama grains.get gpus
lama:
    |_
      ----------
      model:
          Vega [Radeon RX Vega]
      vendor:
          unknown
```
This happened because the `lspci`-reported vendor string was _Advanced Micro Devices, Inc. [AMD/ATI]_. When that is split at whitespace and each component is compared against the list of known vendors, no match is found.

### New Behavior
```
$ salt lama grains.get gpus
lama:
    |_
      ----------
      model:
          Vega [Radeon RX Vega]
      vendor:
          amd
```
I now split the vendor string at any non-alphanumeric characters instead of just whitespace. This ensures that one of the components is _AMD_, matching a known vendor.

### Tests written?

Yes, for GPU grains on Linux: different vendors, different device classes, skipping non-GPU devices, distinguishing AMD vs. ATI.

### Commits signed with GPG?

Yes


cachedout (approved):

cachedout:
A unit test would be very good here, if you have a few moments to put one together.

mkuron:
Do you have an example I could base my test on? I guess I’d need to mock `lspci` and check whether it correctly parses a couple different GPU‘s identifiers.

rallytime:
@mkuron There should be some examples in the [core grains unit tests](https://github.com/saltstack/salt/blob/develop/tests/unit/grains/test_core.py) file that might be helpful.

mkuron:
I figured out how to write and run a unit test. `python tests/runtests.py -v -n unit.grains.test_core.CoreGrainsTestCase.test_linux_gpus` suceeds with my patch applied and fails without it, so I think we are covered.

rallytime (approved):
Awesome! Thank you for the fix and for diving in and writing this test!

PR #49494, opened 2018-09-04
gpg-renderer: Update documentation

SteffenKockel:
### What does this PR do?

The documentation for the gpg renderer was missing a section that explains how to enable the gpg-keys on the salt-master if they are located in a different directory then `/etc/salt/gpgkeys`. Also added section that describes the import of existing gpg keys to the salt-master.

### Tests written?
No

### Commits signed with GPG?

No


salt/renderers/gpg.py
```diff
@@ -37,6 +37,55 @@
     increasing the system entropy. On virtualised Linux systems, this can often
     be achieved by installing the ``rng-tools`` package.
 
+Import keys to a master
+************************
+
+If the keys already exist and need to be imported to the salt master, run the
```
cachedout: It looks like we're assuming the key name here, aren't we? We should perhaps clarify the naming so that people don't just blindly run these commands and then get confused when they don't work.

salt/renderers/gpg.py
```diff
@@ -37,6 +37,55 @@
...
+
+If the salt master runs as normal user, become this user before importing the
+keys. The default target dir will be ``~/.gnupg``. This can be overridden by
+the ``--homedir`` option. The keys must be at least readable for the runuser.
+
+Adjust trust level of imported keys
+***********************************
+
+In some cases, importing existing keys may not be enough an the trust level of
+the key needs to be adjusted. This can be done by editing the key. The ``key_id``
```
cachedout: Is there a way that a person can detect this condition? Do we know what error(s) might be shown? That would be very helpful here, I think.

salt/renderers/gpg.py
```diff
@@ -37,6 +37,55 @@
...
+
+.. code-block:: bash
+
+    gpg --edit-key <key_id>
+
+This will open an interactive shell for the management of the gpg-key. Type
+``trust`` to be able to set the trust level for the key and then select
+``5 (I trust ultimately)``. Then quit the shell by typing ``save``.
+
+Enable usage of pgp keys on the master
```
cachedout: `PGP` and not `pgp`.

salt/renderers/gpg.py
```diff
@@ -37,6 +37,55 @@
...
+**************************************
+
+Generating or importing the keys is not enough to activate the ability to decrypt
+the pillars. Especially if the keys are generated/imported in a non-standard dir.
```
cachedout: These should be separated by a comma and not a period, I think.

cachedout (changes requested):

salt/renderers/gpg.py
SteffenKockel: Hmm, the example shows some filenames that may differ from case to case. One can choose the filename during export... Ideas how to make this more precise?

salt/renderers/gpg.py
SteffenKockel: Yes, this can be seen with `gpg --edit-key key_id`. I shall clarify that.

salt/renderers/gpg.py
SteffenKockel: ups. Fixing...

salt/renderers/gpg.py
SteffenKockel: this is easy :)

salt/renderers/gpg.py
SteffenKockel: Updated line in question. Should be fine imho.

salt/renderers/gpg.py
SteffenKockel: Changed key imort example to use `/path/to/key` syntax instead of just a filename.

rallytime (approved):

salt/renderers/gpg.py
```diff
@@ -37,6 +37,56 @@
...
+
+    gpg --edit-key <key_id>
+
+This will open an interactive shell for the management of the GPG encrypted key. Type
+``trust`` to be able to set the trust level for the key and then select
+``5 (I trust ultimately)``. Then quit the shell by typing ``save``.
+
+Enable usage of GPG keys on the master
+**************************************
+
```
MTecknology: This section is misleading. The ``gpg_keydir`` setting is only required if a non-standard path is used.

MTecknology (changes requested):

cachedout (approved):

rallytime:
Hi @SteffenKockel - Did you see the feedback above? We'd love to get this in!

salt/renderers/gpg.py
```diff
@@ -37,6 +37,56 @@
...
+    gpg --import /path/to/pubkey.gpg
+
+If the salt master runs as normal user, become this user before importing the
+keys. The default target dir will be ``~/.gnupg``. This can be overridden by
+the ``--homedir`` option. The keys must be at least readable for the runuser.
+
+Adjust trust level of imported keys
+***********************************
+
+In some cases, importing existing keys may not be enough an the trust level of
```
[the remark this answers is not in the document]
SteffenKockel: Done.

salt/renderers/gpg.py
SteffenKockel: Wontfix:

The salt tutorial requires one to put some keys into a non-standard location and then assumes this is the default location for gpg keys. If the salt-master is running as non-root user it would be fair to assume that the default dir is `/path/to/salt-master-homedir/.gpg`.

So, the default location is misleading if not outright wrong and the default value is documented as use case only. Now it becomes at least clear that the option exists and can be used. I assume people will be smart enough to omit the option, if it is not needed (documented location used).

MTecknology: The existing documentation uses ``/etc/salt/gpgkeys``. This is also the default location. I'm not sure how you can say this is either misleading or "outright wrong" when it's a well-documented and expected location for these keys.

This section is written as if the option is required, particularly the part where the heading is, "Enable usage of GPG keys on the master". This section should likely be titled something closer to, "Using a different path for keys" and written from that perspective.

MTecknology (changes requested):

salt/renderers/gpg.py
SteffenKockel: The existing documentation uses `/etc/salt/gpgkeys`, which is a non-standard location when it comes to gpg keys (one has to explicitly use the `homedir` option).
So, when it comes to standard locations, the default path should be `~/.gpg`. 

I have no idea why the gpg key handling with salt is so over complicated. People may already have (local) gpg keys and are used to the default gpg key location. 

Plus: Having the keys in /etc/ requires me to track another dir for correct permissions if I don't want to run the salt-master as root user.

PR #49497, opened 2018-09-04
Back-port #49473 to 2018.3

rallytime:
Back-port #49473 to 2018.3


cachedout:
@rallytime Hrm. We're missing a lint fix here, I think.

rallytime:
It should be fixed now :D

cachedout (approved):

PR #49499, opened 2018-09-04
Pin CherryPy version to < 18.0.0 in requirements files for PY2

rallytime:
Version 18.0 now requires Python 3.5, so we need to pin this for Py2.

### DIFF
--- a/requirements/opt.txt
+++ b/requirements/opt.txt
@@ -4,5 +4,6 @@ yappi>=0.8.2
 --allow-unverified python-novaclient>2.17.0
 --allow-unverified python-neutronclient>2.3.6
 python-gnupg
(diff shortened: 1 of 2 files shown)


cachedout (approved):

dmurphy18:
@rallytime REHL 7 for Python 3 currently ships with Python 3.4, cherrypy version check implemented for Python 3 is too limited here I feel. Especially since we supply cherrypy for Python 3 with v5.6.0.

rallytime:
We spoke offline a bit ago, but to clarify the concerns here: CherryPy 18 only supports Python 3.5 and newer. Since we support Python 3.4 as well, we need the check to be `< '3.5'`.

Issue #49523, opened 2018-09-06
[salt-cloud] Non-ascii chars cause an exception in salt-cloud


MTecknology:
### Description of Issue/Question

If a bootstrap script produces a non-ascii character, salt-cloud will produce an exception when trying to print that text to the console. I suspect this is a case of some missing `.encode('utf-8')`'s.

### Steps to Reproduce Issue

The following bootstrap script is the simplest example I can come up with that reproduces the issue.

```
#!/usr/bin/env python
print(u'\u2018')
```

It will produce an error similar to this:
```
UnicodeEncodeError: 'ascii' codec can't encode character u'\u2018' in position 89: ordinal not in range(128)
```

### Versions Report

root@prd-salt-01:/etc/salt# salt --versions-report
Salt Version:
           Salt: 2018.3.2
 
System Versions:
           dist: debian 9.5 
         locale: ANSI_X3.4-1968
        machine: x86_64
        release: 4.15.18-4-pve
         system: Linux
        version: debian 9.5

Ch3LL:
thanks for the PR. Looks like there are some comments on the PR there to change the encoding as he stated utf-8 will not work for everyone. I would also take a look at `salt.utils.data` there are some useful functions there that also handle cases for python2 and python3.

rbjorklin:
I had a quick look and all of these might be duplicates:

https://github.com/saltstack/salt/issues/49304
https://github.com/saltstack/salt/issues/48825
https://github.com/saltstack/salt/issues/49816

Which will all be solved once this has been completed: https://github.com/saltstack/salt/issues/37692

Ch3LL:
oh i commented before viewing the issue you referenced. We should keep this open as we probably just need to update in vt.py to handle this since we added a bunch of unicode work in the 2018.3 release.

Ch3LL:
i'm wondering if we need to do something similar to https://github.com/saltstack/salt/pull/49525 which was closed because utf-8 wont work for everyone. i'm wondering if right there we just need to do `salt.utils.to_unicode`.

MTecknology:
I'd be interesting in taking another look at this. Can you show me an example or two so I could make sure I implement this properly? It's not as simple as just replacing foo.encode('utf-8') with salt.utils.to_unicode(foo), is it?

Ch3LL:
[...]

PR #49525, opened 2018-09-06
[WIP] [salt-cloud] Improve handling of non-ascii characters. (Fixes: #49523)

MTecknology:
### What does this PR do?

Prevents exception when stdout contains non-ascii characters.

- #49523

### Tests written?

No

### Commits signed with GPG?

Yes


damon-atkins:
You need to have a look at cmdmod and output_encoding option and how its used. And determine the correct place to introduce/process it.

PR #49528, opened 2018-09-06
Fix merge wart

dwoz:
### What does this PR do?

Fix wart in `whitelist.txt`

### Tests written?

No

### Commits signed with GPG?

Yes

### DIFF
--- a/tests/whitelist.txt
+++ b/tests/whitelist.txt
@@ -40,7 +40,6 @@ integration.modules.test_sysmod
 integration.modules.test_system
 integration.modules.test_test
 integration.modules.test_useradd
-integration.modules.test_autoruns
 integration.modules.test_win_dns_client
 integration.modules.test_win_pkg
 integration.reactor.test_reactor


rallytime:
@dwoz Ah sorry, I mixed this up in my merge-forward. Thanks for fixing this again. :)

rallytime (approved):

rallytime:
@dwoz Any idea on why so many of the windows tests are failing on PRs, but not on the branch? I see only 6 tests failing on the Py3 windows tests for 2018.3 branch, for example, but there are 27 on this PR.

I have been seeing more tests fail on PRs than I think are failing on the branches. I just wanted to make sure that was something you're aware of.

dwoz:
@rallytime I'm not sure why there is a difference in the number of failures. So far I've been chalking things like that up to flaky test runs. I've re-launched the windows PR builds to see if the results are consistent.

rallytime:
@dwoz Sounds good. Let's get those flaky tests fixed or marked as such so we can make sure the signaling is as best as it can be, as long as that is on your radar!

PR #49537, opened 2018-09-06
add nsot external pillar module

wasabi222:
### What does this PR do?
Adds the nsot external pillar module, allowing pillar from nsot.

N/A

### Tests written?

No

### Commits signed with GPG?

No


salt/pillar/nsot.py
```diff
@@ -0,0 +1,181 @@
+# -*- coding: utf-8 -*-
+'''
+A module that adds data to the Pillar structure from the NSoT API.
+
+.. versionadded:: 2018.3.4
```
terminalmage: This must be changed to `Neon`, as the develop branch will not go into the 2018.3 release cycle.

terminalmage (changes requested):

terminalmage (approved):

rallytime:
Hi @wasabi222 - Can you rebase this PR against the latest HEAD of develop? There are several lint failures and test failures here that are suspicious and I suspect a stale branch. Then we can get a cleaner run.

salt/pillar/nsot.py
```diff
@@ -0,0 +1,181 @@
...
+    This is in case a minion_id contains a dot. Because NSoT
+    doesn't allow hostnmes with dots, this option allows you to look up a device
+    with this character replacing dots. For example, with this option set to
+     '-', the minion rtr1.lax1 would be looked up in NSoT as rtr1-lax1.
+
+all_device_info: ``list``
+    This is a list of minions that will receive all device
+    information in NSoT. For example, if this list contained the entry
+    'test_min', then the minion 'test_min' would have all devices in NSoT under
+    pillar['nsot']['devices']
```
isbm: How to give this to my 5K machines? This should support the same regexes as `minion_regex`.

salt/pillar/nsot.py
```diff
@@ -0,0 +1,181 @@
...
+    of whether or not to use the tailing slash, this module assumes that the
+    APPEND_SLASH option is set to True (the default)
+
+The following fields are optional:
+
+replace_dot: ``str``
+    This is in case a minion_id contains a dot. Because NSoT
+    doesn't allow hostnmes with dots, this option allows you to look up a device
+    with this character replacing dots. For example, with this option set to
+     '-', the minion rtr1.lax1 would be looked up in NSoT as rtr1-lax1.
```
isbm: `fqdn-separator`?

salt/pillar/nsot.py
```diff
@@ -0,0 +1,181 @@
...
+    :param secret_key: str
+    :return: str
+    '''
+    url += 'authenticate/'
+    data_dict = {"email": email, "secret_key": secret_key}
+    query = salt.utils.http.query(url, data=data_dict, method='POST',
+                                  decode=True)
+    error = query.get('error')
+    if error:
+        log.error('can\'t get auth_token from nsot! reason: %s', error)
```
isbm: What is "auth_token"? This is just a name of a key inside the structure inside JSON that end user should never even care about. Please do not put such stuff into the logs. Instead please rephrase it as it would be non-Salt-aware guy reading it, and also write it properly, from capital letter, something like "Unable to obtain authentication token from NSoT: %s".

salt/pillar/nsot.py
```diff
@@ -0,0 +1,181 @@
...
+        return False
+    else:
+        log.debug('successfully obtained token from nsot!')
+        return query['dict']['auth_token']
```
isbm: Unless you're _really sure_ that this structure never changes and _always_ contains these keys, use `.get` instead:

```python
return query.get('dict', {}).get('auth_token', False)
```

 But then debug should also say "oops, structure is different, no token has been found"

salt/pillar/nsot.py
```diff
@@ -0,0 +1,181 @@
...
+    :param minion_id: str
+    :param minion_regex: list
+    :return: bool
+    '''
+    get_pillar = False
+    for pattern in minion_regex:
+        log.debug('searching %s using %s', minion_id, minion_regex)
+        match = re.search(pattern, minion_id)
+        if match and match.string == minion_id:
+            log.debug('found match! %s: regex %s', minion_id, minion_regex)
```
isbm: Could you please be more verbose? What would you think if you have 100 DEBUG log records from all around everywhere and suddenly "found match!", then "regex!!"... How one supposed to know it is coming from exactly here? Debug is to help find problems. But now it raises questions "What is this stuff and where it comes from!?"

Please be move verbose and user-friendly.

salt/pillar/nsot.py
```diff
@@ -0,0 +1,181 @@
...
+            get_pillar = True
+            break
+        log.debug('unable to match %s using regex %s', minion_id, minion_regex)
```
isbm: Same here. Please refrain from programming slang in the messages that supposed to be read by people that might not be aware of Salt internals.

salt/pillar/nsot.py
```diff
@@ -0,0 +1,181 @@
...
+
+
+def _query_nsot(url, headers, device=None):
+    url += 'devices/'
+    if not device:
+        query = salt.utils.http.query(url, header_dict=headers, decode=True)
+    else:
+        device += '/'
+        query = salt.utils.http.query(url + device, header_dict=headers,
+                                      decode=True)
```
isbm: Please refer to a proper way of doing this, i.e. `os.path.join`, instead of `+=` on strings, and then refactor accordingly.

salt/pillar/nsot.py
```diff
@@ -0,0 +1,181 @@
...
+    error = query.get('error')
+    if error:
+        log.error('can\'t get device(s) from nsot! reason: %s', error)
+        return {}
+    else:
+        return query['dict']
```
isbm: Do we need double-returns here?

```python
error = query.get('error')
if error:
    log.error('....
    ret = {}
else:
    ret = query['dict']

return ret
```

salt/pillar/nsot.py
```diff
@@ -0,0 +1,181 @@
...
+            return ret
+
+    if replace_dot:
+        minion_id = minion_id.replace('.', replace_dot)
+
+    device_info = _query_nsot(api_url, headers, device=minion_id)
+    if device_info:
+        ret['nsot'] = device_info
+
+    return ret
```
isbm: This is not very nice. Please make sure we have only one return at the end, so we can continue maintaining this code in a future.

isbm (changes requested):
Thanks for the contribution! I have couple of little requests here.

salt/pillar/nsot.py
wasabi222: Fair point. I will change this.

salt/pillar/nsot.py
wasabi222: An auth_token is generated based on a valid email/secret_key. If it's unable to be obtained, the end user should care about it since we're expecting to obtain an auth_token if the proper creds are provided and the service is running.

salt/pillar/nsot.py
wasabi222: Good point. I'll change this.

salt/pillar/nsot.py
wasabi222: I'm unsure of the programming slang you're referring to here. I'll update this to reflect where it's coming from, though.

isbm: @wasabi222 e.g. "regex" is something that used among programmers/admins. But "regular expression" is something that is generic and official.

wasabi222:
@rallytime Is this looking better from a test perspective? I've followed a guide on how to rebase but this is all still very new to me so I'm not actually sure if it changed anything. If this looks correct, I'll continue with the suggested changes that @isbm suggested, test locally using the linter, and then update this PR. I just want to make sure tests look better after the rebase.

rallytime:
@wasabi222 Yes, much better! Lint is happy now. :)

Thanks!

salt/pillar/nsot.py
isbm: This was a clue question. :wink: My point is that an ordinary newcomer will just stare at this message, having no idea where this coming from, why and what to do next. So you'd better to be a bit more verbose here and rephrase it with generic terminology so it is clear what action to take next.

isbm (changes requested):
Looks good, except a little nitpick on the logging message, just to make it more clear.

wasabi222:
@rallytime This has less successful checks than my last commit before the refactoring :/ Do I need to rebase this again? This is all still new and unfamiliar to me, sorry! :)

rallytime:
Hi @wasabi222 - no worries. We're working on getting some of these errant tests fixed. This looks fine to me, test failures are unrelated.

Once @isbm has a chance to respond to your comment we can proceed from there.

Thank you!

isbm:
> @isbm My thinking with leaving that log message is that it should contain next steps for the user to take via the `%s` - for instance, `404 not found` or `403 unauthorized`, etc.

Ok. This is even better. :laughing: I'd personally would rather dig to the docs of nsot and see what the heck that would mean.

> Because this external pillar module requires a working nsot installation, my assumption is that a user will have a reasonable understanding of nsot and how it works, and thus the resulting log message should be sufficient.

Yes, usually this is the case of soho-size business. However even in a mid-size enterprise, once you are in a different department or even shift or even task-domain of the same department, this might be not the case anymore. I'd not assume that the user already knows what "nsot" is, but treat him as a total newcomer. It should not print there a complete documentation obviously, but I'd expect there a bit more clues and a bit more generic terminology. So it should be understandable where this comes from and  what to do next.

salt/pillar/nsot.py
```diff
@@ -92,7 +92,9 @@ def _get_token(url, email, secret_key):
                                   decode=True)
     error = query.get('error')
     if error:
-        log.error('can\'t get auth_token from nsot! reason: %s', error)
+        log.error('nsot ext pillar: can\'t get auth_token! reason: %s', error)
+        log.error('nsot ext pillar: verify that %s is reachable', url)
+        log.error('nsot ext pillar: verify that %s is a valid email', email)
```
isbm: OK, but we don't do it that way normally. How about this:

```python
log.error('Cannot obtain NSoT authentication token due to: %s.', error)
log.debug("Please verify NSoT URL '%s' is reachable and email is valid", url, email)
```

isbm (changes requested):

salt/pillar/nsot.py
wasabi222: This would put me over the 80 char line limit. I'm struggling to find a way to do this that fits within that constraint. Here's what I have:

```
if error:
        log.error('Cannot obtain NSoT authentication token due to: %s.', error)
        log.debug('verify NSoT URL %s is reachable', url)
        log.debug('verify email %s is valid', email)
```

salt/pillar/nsot.py
isbm: So do we?

salt/pillar/nsot.py
```diff
@@ -0,0 +1,220 @@
...
+    :param email: str
+    :param secret_key: str
+    :return: dict
+    '''
+    token = _get_token(api_url, email, secret_key)
+    if not token:
+        return
+    headers = {'Authorization': 'AuthToken {}:{}'.format(email, token)}
+    all_devices = _query_nsot(api_url, headers)
+    return all_devices
```
isbm: It would be better to return the same type. In both `_all_nsot_devices` and `_proxy_info` functions. Now you return `dictionary` or `None`. To fix this, please remove multuple returns:
```python
token = _get_token(api_url, email, secret_key)
if token:
    headers = {'Authorization': 'AuthToken {}:{}'.format(email, token)}
    all_devices = _query_nsot(api_url, headers)
else:
    all_devices = {}

return all_devices
```

isbm (changes requested):

salt/pillar/nsot.py
isbm: Great! Thanks.

salt/pillar/nsot.py
cachedout: @wasabi222 Are you able to resolve this concern? If so, we should be able to merge this!

wasabi222: @cachedout I just updated my PR to fix the last of these issues :)

PR #49538, opened 2018-09-06
Fix for SUSE Expanded Support detection

brejoc:
### What does this PR do?

A SUSE ES installation has both, the `centos-release` and `redhat-release`
file. Since `os_data()` only used the `centos-release` file to detect a
CentOS installation, this lead to SUSE ES being detected as CentOS.

This change also adds a check for `redhat-release` and then marks the
`lsb_distrib_id` as RedHat.

### Previous Behavior

```
$ salt-call --local grains.get os
local:
    CentOS
```

### New Behavior

```
$ salt-call --local grains.get os
local:
    RedHat
```

### Tests written?

No

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/grains/core.py
+++ b/salt/grains/core.py
@@ -1769,8 +1769,16 @@ def os_data():
                                     comps[3].replace('(', '').replace(')', '')
                 elif os.path.isfile('/etc/centos-release'):
                     log.trace('Parsing distrib info from /etc/centos-release')
-                    # CentOS Linux
-                    grains['lsb_distrib_id'] = 'CentOS'
+                    # Maybe CentOS Linux; could also be SUSE Expanded Support.
+                    # SUSE ES has both, centos-release and redhat-release.
+                    if os.path.isfile('/etc/redhat-release'):
+                        with salt.utils.files.fopen('/etc/redhat-release') as ifile:
+                            for line in ifile:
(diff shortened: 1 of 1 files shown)


salt/grains/core.py
```diff
@@ -1769,8 +1769,16 @@ def os_data():
                                     comps[3].replace('(', '').replace(')', '')
                 elif os.path.isfile('/etc/centos-release'):
                     log.trace('Parsing distrib info from /etc/centos-release')
-                    # CentOS Linux
-                    grains['lsb_distrib_id'] = 'CentOS'
+                    # Maybe CentOS Linux; could also be SUSE Expanded Support.
+                    # SUSE ES has both, centos-release and redhat-release.
+                    if os.path.isfile('/etc/redhat-release'):
+                        with salt.utils.fopen('/etc/redhat-release') as ifile:
```
terminalmage: This should be `salt.utils.files.fopen()` in develop now.

terminalmage (changes requested):

salt/grains/core.py
brejoc: Thanks for catching this. Fixed it and squashed the commits.

terminalmage (approved):

PR #48344, opened 2018-06-27
allow using proxy minion from cli

gtmanfred:
### What does this PR do?
This allows using proxy minions from the minion box, for troubleshooting, and also not needing a master.

```
[root@salt salt]# tail -c +0 /etc/salt/proxy /srv/pillar/*
==> /etc/salt/proxy <==
master: localhost

==> /srv/pillar/proxy.sls <==
proxy:
  proxytype: docker
  name: {{grains.id}}

==> /srv/pillar/top.sls <==
base:
  '*':
    - proxy
[root@salt salt]# salt-call --local grains.get virtual_subtype
local:
[root@salt salt]# docker ps
CONTAINER ID        IMAGE               COMMAND                  CREATED             STATUS              PORTS               NAMES
f0b4c86881cb        centos              "/usr/lib/systemd/sy…"   3 hours ago         Up 3 hours                              quirky_edison
[root@salt salt]# salt-call --local grains.get virtual_subtype --proxyid quirky_edison
local:
    Docker
```

### Tests written?

Not Yet

### Commits signed with GPG?

Yes

PR #49552, opened 2018-09-06
Remove redundant execution in salt-call

terminalmage:
Pull #48344 added logic to salt-call which runs the desired function using all
of the executors, or just the direct_call one if none are configured.

However, the previous code which ran the function was not removed, which
results in the function being run twice.

This removes the (now) redundant function call.

### DIFF
--- a/salt/cli/caller.py
+++ b/salt/cli/caller.py
@@ -223,7 +223,6 @@ class BaseCaller(object):
             if isinstance(executors, six.string_types):
                 executors = [executors]
             try:
-                ret['return'] = func(*args, **kwargs)
                 for name in executors:
                     fname = '{0}.execute'.format(name)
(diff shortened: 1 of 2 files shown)


gtmanfred (approved):

DmitryKuzmenko (approved):
:+1:

isbm (approved):

terminalmage:
Before this is merged, I will add an integration test. This seems like something important :smile:

terminalmage:
There was already a test that would have protected us from this regression but it was skipped for a long time now for reasons that are likely no longer valid. So I have removed the skip and rewritten the test to be a bit more robust.

rallytime:
@terminalmage Can you fix the lint errors?

rallytime (approved):

PR #49568, opened 2018-09-07
Skip reverse patch integration tests for version <2.6.1

Ch3LL:
### What does this PR do?
I found an issue in patch version < 2.6.1 and on macosx it has 2.5.9 installed. When attempting to use a reject file it ruins the header of the patch file from this:

```
diff -ur a/foo/bar/math.txt b/foo/bar/math.txt
--- a/foo/bar/math.txt  2018-04-09 18:43:52.883205365 -0500
+++ b/foo/bar/math.txt  2018-04-09 18:44:58.525061654 -0500
@@ -1,3 +1,3 @@
-Five plus five is ten
+5 + 5 = 10
```

to this:

```
********************************
********************************
@@ -1,3 +1,3 @@
-Five plus five is ten
+5 + 5 = 10
```

this is causing these tests:

```
integration.states.test_file.PatchTest.test_patch_directory
integration.states.test_file.PatchTest.test_patch_directory_remote_source
integration.states.test_file.PatchTest.test_patch_directory_remote_source_template
integration.states.test_file.PatchTest.test_patch_directory_template
integration.states.test_file.PatchTest.test_patch_strip_parsing
```

to fail on macosx, because it can't read the reject file correctly.

I noticed this was fixed in >=2.6.1 patch versions.

### DIFF
--- a/tests/integration/states/test_file.py
+++ b/tests/integration/states/test_file.py
@@ -42,6 +42,7 @@ import salt.utils.json
 import salt.utils.path
 import salt.utils.platform
 import salt.utils.stringutils
+from salt.utils.versions import LooseVersion as _LooseVersion
 
 HAS_PWD = True
 try:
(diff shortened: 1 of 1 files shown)


rallytime (approved):

cachedout:
We're certain this isn't an actual bug though, right?

Ch3LL:
@cachedout i ran the patch command outside of salt and it destroyed the header. I tried this on both linux and mac on patch versions <2.6 I glanced at patch's bug log and could not find a particular issue that was opened. Once i upgraded patch >=2.6.1 it did not destroy the header anymore, so i'm certain this is a patch issue and not a salt issue.

PR #49585, opened 2018-09-09
Squelch log message RE: server ID

terminalmage:
INFO is way too verbose since it this gets logged _every_ time the grains are compiled.


cachedout (approved):

cachedout:
cc: @isbm

PR #49593, opened 2018-09-10
Add more tests to whitelist for Windows

dwoz:
### What does this PR do?

Add more test file to `whitelist.txt` for Windows builds.

### Tests written?

No

### Commits signed with GPG?

Yes


tests/integration/modules/test_file.py
```diff
@@ -179,9 +179,14 @@ def test_remove_broken_symlink(self):
 
     def test_cannot_remove(self):
         ret = self.run_function('file.remove', arg=['tty'])
-        self.assertEqual(
-            'ERROR executing \'file.remove\': File path must be absolute: tty', ret
-        )
+        if salt.utils.platform.is_windows():
```
cachedout: Should we just normalize this error message in the module itself instead?

dwoz: Assuming we require an absolute path for both platforms. I think we do but I'll verify that and then change the message if needed.

rallytime (approved):

rallytime:
@dwoz Looks like these test failures are new: https://jenkinsci.saltstack.com/job/pr-windows2016-py3/job/PR-49593/7/

Issue #47431, opened 2018-05-02
Remove/update release warning banner on new/upcoming release note pages


rallytime:
### Description of Issue/Question
We always add a new release notes file after the next stable release launches. For example, [2018.3.1 release notes](https://docs.saltstack.com/en/latest/topics/releases/2018.3.1.html) exist. This is done so people can add changes to the file as we prepare for the 2018.3.1 release.

Since the 2018.3.1 release is presently not the current stable release, nor is it in the `develop` branch, a warning banner displays at the bottom of the page:

```
These release notes are for an old release of Salt. This release might contain known security and other issues that are fixed in the latest release.
```

This is incorrect and confusing. At the very least, we need to remove that banner for "future" release notes. 

It would be even better to have a warning there still, but change the text to say that it is an unreleased version presently. This approach would make it even more clear that those release notes are for a version that has not be released yet.

Ch3LL:
just verified this is working. closing for now :)

Ch3LL:
actually re-opening 2017.7.6 doesn't seem to be showing changes. will check into that

Ch3LL:
this is halfway fixed. It now does not show the warning on in progress releases. But we want to implement an approach that will automatically notfiy the user that this is an in progress release without having to manually add this to the release notes each time. So I will keep open until we can implement this.

PR #49613, opened 2018-09-11
Remove confusing footer banner from release notes

kaelaworthen:
### What does this PR do?
Removes warning banner at the bottom of the release notes.

https://github.com/saltstack/salt/issues/47431

In progress warnings already been added by @Ch3LL for new releases so this should cover what's left.

### Tests written?

No

### Commits signed with GPG?

No

### DIFF
--- a/doc/_themes/saltstack2/layout.html
+++ b/doc/_themes/saltstack2/layout.html
@@ -185,11 +185,7 @@
                     <div class="row">
                         <div class="col-sm-12 col-md-11 col-md-offset-1 col-lg-10 col-lg-offset-1">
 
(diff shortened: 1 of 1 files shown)


Ch3LL:
I don't think we want to actually remove this warning, because it needs to show up on older releases. Have you noticed another place where the warning is popping up when it should not be? 

I mostly resolved the issue you are associating this PR with but am keeping it open to make sure we are handling in progress releases automatically and not manually.

rallytime:
@kaelaworthen Did you see the comment above?

cachedout:
Hi @kaelaworthen. We'll have to close this if we don't hear back from you. Please let us know what you think about the comments from @Ch3LL that are above. Thanks!

PR #49617, opened 2018-09-11
Fix for #48915.

### LINKED ISSUE #48915: Unclear distinction between publisher_acl and external_auth in master.conf
When configuring salt for use by a non-root user or for use with the salt-api there is some confusion whether to use the external_auth or publisher_acl directives in the master.conf.

While there is a distinction at https://docs.saltstack.com/en/latest/topics/eauth/access_control.html#when-to-use-each-authentication-system the respective subpages https://docs.saltstack.com/en/latest/ref/publisheracl.html and https://docs.saltstack.com/en/latest/topics/eauth/index.html give no such clue.

I propose adding the respective section from the access_control page to either subpages top descriotion part to clarify things.

MKelley80:
### What does this PR do?
Fixes #48915. I added the 'When to use...' content from the comparison topic to the two discrete topics, with links back to the comparison topic.

#48915

### Tests written?

No

### Commits signed with GPG?

Yes/No

### DIFF
--- a/doc/topics/eauth/access_control.rst
+++ b/doc/topics/eauth/access_control.rst
@@ -28,6 +28,13 @@ case of the peer system.
     index
     ../../ref/peer
 
+.. The two paragraphs below (in the "When to use each authentication system"
+   heading) are copied in the doc/ref/publisheracl.rst and doc/topics/eauth/index.rst
+   topics as a note, at the top of the document. If you update the below
+   content, update it in the other two files as well.
+
+.. _acl_types:
+
 When to Use Each Authentication System
 ======================================
 ``publisher_acl`` is useful for allowing local system users to run Salt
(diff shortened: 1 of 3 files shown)


garethgreenaway (approved):

doc/topics/eauth/access_control.rst
```diff
@@ -28,6 +28,13 @@ case of the peer system.
     index
     ../../ref/peer
 
+.. The two paragraphs below (in the "When to use each authentication system"
```
rallytime: Did you mean for this section to be in a `.. note::` block?

MKelley80: No, that's a comment we don't want visible to the end-user. It's just telling the person who edits that particular content in the future that the content also exists in other places, and that they need to update the other instances.

rallytime: Ah, right. I don't think I've ever used an rst comment before! Thanks!

rallytime (approved):

PR #49622, opened 2018-09-12
fix command line options

angeloudy:
### What does this PR do?

python `subprocess.run` only accepts arguments as a list.
eg.
```
extract_something:
  archive.extracted:
    - name: /tmp/myfile.zip
    - options: -x myfile/test
    - archive_format: zip
```
In the state above, `options` wouldn't work because salt will only `subprocess.run(['unzip', '-x myfile/test', '/tmp/myfile.zip'])` instead of  `subprocess.run(['unzip', '-x',  'myfile/test', '/tmp/myfile.zip'])`


salt/modules/archive.py
```diff
@@ -936,7 +936,7 @@ def cmd_unzip(zip_file,
     if password:
         cmd.extend(['-P', password])
     if options:
-        cmd.append('{0}'.format(options))
+        cmd = cmd + options.split()
```
dwoz: I think we should we use `shlex.split` here.

dwoz (changes requested):

salt/modules/archive.py
terminalmage: Yes, we need to use `shlex.split()`, and we should also be using `cmd.extend()` instead of `cmd = cmd + ....`, because the former will add to the existing list while the 2nd will unnecessarily allocate an extra list.

terminalmage (changes requested):

terminalmage:
@rallytime This should be backported as far back as you feel comfortable.

rallytime:
@angeloudy Can you make the requested changes listed above? Then we can move this forward.

cachedout:
@angeloudy We would really like to get this in but we need a response from you. Will you have any free time to look at this?

angeloudy:
@cachedout I have made the suggested changes.

rallytime:
Hi @angeloudy - There is a test failing due to the changes here. Can you take a look?

https://jenkinsci.saltstack.com/job/pr-kitchen-ubuntu1604-py2/job/PR-49622/16/

angeloudy:
@rallytime I made the suggested changes `cmd.extend()` which break it because it seems cmd is not always initialized. Shall I change it back to `cmd + `

cachedout:
> @rallytime I made the suggested changes `cmd.extend()` which break it because it seems cmd is not always initialized. Shall I change it back to `cmd + `

I think we should just initialize cmd.

salt/modules/archive.py
```diff
@@ -936,7 +936,7 @@ def cmd_unzip(zip_file,
     if password:
         cmd.extend(['-P', password])
     if options:
-        cmd.append('{0}'.format(options))
+        cmd = cmd.extend(shlex.split(options))
```
thatch45: Just change this line to:
```
cmd.extend(shlex.split(options))
```
extend is an inline modifier so you are setting `cmd` to None here

thatch45 (changes requested):
Just the change I posted

terminalmage (approved):

terminalmage:
Backported in https://github.com/saltstack/salt/pull/50883

PR #49624, opened 2018-09-12
modules/nilrt_ip.py: Fix enable/disable function

alexvasiu:
The enable function shouldn't take any action if the service is already
    enabled and same for the disable function if the service is already
    disabled.
    
### DIFF
--- a/salt/modules/nilrt_ip.py
+++ b/salt/modules/nilrt_ip.py
@@ -536,7 +536,8 @@ def _change_state(interface, new_state):
     service = _interface_to_service(interface)
     if not service:
         raise salt.exceptions.CommandExecutionError('Invalid interface name: {0}'.format(interface))
-    if not _connected(service):
+    connected = _connected(service)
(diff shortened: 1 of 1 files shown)


rares-pop (commented):
do we need some test for this?

the _set_state() function should be renamed to what its comment says it's doing, because I would have expected set_ethercat to call into this too (now the logic is split among many functions).

The logic for CG vs NG should be split in two different functions because the have no common/shared code at all.

alexvasiu:
@rares-pop We already have a test for this, but no one runs that tests.

If we split the logic from this function, then we have to split all common functions used on both, CG and NG

rallytime (approved):

rares-pop:
It's sad that we are not running the test suite before *any* submit.

We shouldn't need to split the functionality in all functions, but when the size of the function exceeds a certain size, it's better to separate (see that codeclimante complains about it too).

PR #49639, opened 2018-09-13
Retire MD5 checksum for pkg mgmt plugins

isbm:
### What does this PR do?

Yum Notify and Zypper Notification plugins are still using MD5 algorithm, which is missing on FIPS-enabled Python builds.

Time to just say goodbye to MD5.

### DIFF
--- a/scripts/suse/yum/plugins/yumnotify.py
+++ b/scripts/suse/yum/plugins/yumnotify.py
@@ -32,7 +32,8 @@ def _get_checksum():
     Returns:
         hexdigest
     """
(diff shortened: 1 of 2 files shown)


garethgreenaway:
This should go in develop with a depreciation warning for Fluorine.

isbm:
@garethgreenaway no, it doesn't need to. It is just a cookie for RPM database which is constantly renewed, once something is installed/updated/removed.

cachedout:
I agree with @isbm. I think this is fine here.

rallytime (approved):

PR #49650, opened 2018-09-14
fix aliyun cloud typeerror

Yxnt:
### What does this PR do?
fix cloud aliyun TypeError

### Previous Behavior
```
[root@localhost]# salt-cloud --location=cn-shanghai -f list_nodes xxxx
........
TypeError: character mapping must return integer, None or unicode
```
### New Behavior
`[root@localhost]# salt-cloud --location=cn-shanghai -f list_nodes xxxx`
```
xxxx:
    ----------
    aliyun:
        ----------
        secret:
            ----------
            id:
                secret
            name:
                secret
            private_ips:
                - xxx.xxx.xxx.xxx
            public_ips:
                - xx.xxx.xxx.xxx
            size:
                TODO
            state:
                Running
```

### Tests written?

No

### Commits signed with GPG?

No


salt/cloud/clouds/aliyun.py
```diff
@@ -770,7 +770,7 @@ def percent_encode(line):
     # All aliyun API only support GET method
     stringToSign = 'GET&%2F&' + percent_encode(canonicalizedQueryString[1:])
 
-    h = hmac.new(access_key_secret + "&", stringToSign, sha1)
+    h = hmac.new(bytearray(access_key_secret + "&", 'utf-8'), bytearray(stringToSign, 'utf-8'), sha1)
```
rallytime: While I can't test this change, I think it would be better to use the `salt.utils.stringutils` functions instead of hard-coding `utf-8` here.

rallytime (changes requested):
Hello @Yxnt! Thank you for submitting this fix. I have one suggestion. 

Can you also move this change to the `2018.3` branch? We're finishing up some release tasks on the 2018.3.3 branch and it's closed to fixes currently.

salt/cloud/clouds/aliyun.py
Yxnt: Maybe i shuold move this change to develop branch ? (This is my first pr)
i will use `salt.utils.stringutils` functions instead of `utf8`

Thx:blush:

rallytime: Sure, that would be fine. Thank you so much for contributing. We are so happy to have you. :)

rallytime (approved):

Yxnt:
@rallytime I‘m so sorry for my last commit *b18095e*, 我今天太忙了可能操作错了

PR #49666, opened 2018-09-17
Add a test.master_ping function

### LINKED ISSUE #49621: Add a test.master_ping function
This would be used by salt-call to ensure that the minion can talk to the master.

The code to do this already exists as part of the `ping_interval` config option, and could easily be adapted for this purpose.

terminalmage:
Resolves #49621.

### DIFF
--- a/salt/modules/test.py
+++ b/salt/modules/test.py
@@ -119,7 +119,6 @@ def ping():
 
         salt '*' test.ping
     '''
-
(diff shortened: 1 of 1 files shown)


terminalmage:
@rares-pop good point. I actually was suggested to write this func at SaltConf and was not aware of `status.master`. That function seems to be better suited to this task, so I'll close this.

PR #49669, opened 2018-09-17
Documentation: fix typo in "equivalent"

aplanas:
Fix typo in the "equivalent" word in the tutorial and in the manual page


rallytime (approved):
Good catch!

In the future, you don't need to worry about updating the `man` pages as those are automatically generated off of the built docs. It's fine in this PR, but changes to those may be overwritten next time the man pages are built. :)

roaldnefs:
Already ported to master in #51405.

Issue #49675, opened 2018-09-17
win_system.get_pending_domain_join returning False after successful domain join


hrumph:
### Description of Issue/Question
win_system.get_pending_domain_join is returning False even though PC was just joined to an AD domain
### Setup
Here is a state file which renames and joins then attempts a reboot.
```
name computer:
    system.computer_name:
       - name: OUREXAMS2100008

{% set user_name = pillar['user_name'] %}
{% set p_word = pillar['p_word'] %}
join domain:   
    system.join_domain:
       - name: AD.QUEENSU.CA
       - username: {{user_name}}
       - password: {{p_word}}
       - account_exists: True
       - restart: False
       - require:
            - name computer
       
reboot computer:
     system.reboot:
        - timeout: 0
        - only_on_pending_reboot: True
        - order: last
        - require:
              - join domain
```

### Steps to Reproduce Issue
We can run the above state file with:
```
salt -t 120 '210008' state.apply our_join pillar='{"user_name": "garf", "p_word": "asdfasdfasdf"}'
```

We will then get the following output:
```
210008:
----------
          ID: name computer
    Function: system.computer_name
        Name: OUREXAMS210008
      Result: True
     Comment: Computer name successfully set to 'OUREXAMS210008' (reboot required for change to take effect)
     Started: 15:32:08.356468
    Duration: 31.577 ms
     Changes:   
              ----------
              new:
                  OUREXAMS210008
              old:
                  somehtingorother
----------
          ID: join domain
    Function: system.join_domain
        Name: AD.QUEENSU.CA
      Result: True
     Comment: Computer added to 'AD.QUEENSU.CA'
     Started: 15:32:08.403369
    Duration: 3955.51 ms
     Changes:   
----------
          ID: shutdown computer
    Function: system.reboot
      Result: True
     Comment: System reboot has been skipped because no reboot was pending
     Started: 15:32:12.358879
    Duration: 15.3 ms
     Changes:  
```

(Note that the changes dictionary in the "join domain" state is empty, which is a separate issue that also needs fixing)

The final reboot fails because it also turns out that when we run the following command:
```
salt -t 120 '210008' system.get_pending_domain_join
```

we then get the following output
```
210008:
    False
```
The result of the command is surely wrong.

### Versions Report
```
Salt Version:
           Salt: 2018.3.2
 
Dependency Versions:
           cffi: 1.10.0
       cherrypy: Not Installed
       dateutil: 2.6.0
      docker-py: Not Installed
[...]

PR #49684, opened 2018-09-17
Detect when system needs reboot after domain join

twangboy:
### What does this PR do?
Properly detects if the system requires reboot after Joining a domain.
Adds value_exists function
Uses key_exists instead of reading and failing

https://github.com/saltstack/salt/issues/49675

### Tests written?
No

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/modules/win_system.py
+++ b/salt/modules/win_system.py
@@ -1205,29 +1205,24 @@ def get_pending_domain_join():
 
         salt '*' system.get_pending_domain_join
     '''
-    vname = '(Default)'
     base_key = r'SYSTEM\CurrentControlSet\Services\Netlogon'
     avoid_key = r'{0}\AvoidSpnSet'.format(base_key)
     join_key = r'{0}\JoinDomain'.format(base_key)
 
     # If either the avoid_key or join_key is present,
     # then there is a reboot pending.
-
-    avoid_reg_ret = __salt__['reg.read_value']('HKLM', avoid_key, vname)
-
-    if avoid_reg_ret['success']:
-        log.debug('Found key: %s', avoid_key)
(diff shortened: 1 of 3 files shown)


twangboy:
@rallytime Probably need to backport to 2017.7, but I'll need to create separate PR.

hrumph:
@twangboy , thanks for the fix.
There are still two more instances of the same error in the module. You also need to fix the _get_pending_update_ and _get_pending_component_servicing_. You may as well fix everything at once.

hrumph:
@Twangboy I am now happy to see this committed.

rallytime:
@twangboy The following windows test failures look related to me:

- integration.modules.test_system.WinSystemModuleTest.test_get_computer_name
- integration.modules.test_service.ServiceModuleTest.test_service_disable_doesnot_exist
- integration.modules.test_service.ServiceModuleTest.test_service_enable

https://jenkinsci.saltstack.com/job/pr-windows2016-py3/job/PR-49684/3/

Can you fix those?

hrumph:
@rallytime, This is none of my business I suppose. I don't know how the testing works. Certainly, the _get_computer_name_ function works for me (on Windows 10).  (I haven't looked at the other two functions). How can one know the particulars of the test system in order to duplicate the environment? Can it be ascertained that the test system has no permissions issues and so forth?

The test issues may or may not be related to each other but I don't see how they're related to the bugs @Twangboy just fixed

twangboy:
@rallytime Yeah, I just relaunched the tests, but I don't think the failures are related.

rallytime:
Ok, the PR tests for windows still need some ❤️ so I was trying to make an educated guess so we don't cause more failures. Let's see how the next run goes. Thanks for chiming in!

rallytime (approved):

PR #49685, opened 2018-09-17
Fix heading line-lengths

rallytime:
This should clean up some of the warnings found by building the documentation:

```
/root/SaltStack/salt/doc/ref/pillar/all/index.rst:55:<autosummary>:1: WARNING: Unexpected section title or transition.

=================
```

I should check the doc build result before merging. I am not sure if I got all of them.

### DIFF
--- a/salt/pillar/cmd_yamlex.py
+++ b/salt/pillar/cmd_yamlex.py
@@ -1,14 +1,12 @@
 # -*- coding: utf-8 -*-
 '''
-Execute a command and read the output as YAMLEX. The YAMLEX data is then directly overlaid onto the minion's Pillar data
+Execute a command and read the output as YAMLEX.
+
+The YAMLEX data is then directly overlaid onto the minion's Pillar data
 '''
 
-# Don't "fix" the above docstring to put it on two lines, as the sphinx
-# autosummary pulls only the first line for its description.
-
-from __future__ import absolute_import
-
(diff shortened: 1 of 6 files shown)


dwoz (approved):

isbm (approved):

salt/pillar/cmd_yamlex.py
```diff
@@ -1,7 +1,6 @@
 # -*- coding: utf-8 -*-
 '''
-Execute a command and read the output as YAMLEX. The YAMLEX data is then
-directly overlaid onto the minion's Pillar data
+Execute a command and read the output as YAMLEX. The YAMLEX data is then directly overlaid onto the minion's Pillar data
```
DmitryKuzmenko: I understand other changes but is there a reason to do this line 121 characters long? If there's any I'm ok.

DmitryKuzmenko (approved):

salt/pillar/cmd_yamlex.py
rallytime: There's a note a couple of lines down about this which gave me a clue:
```
# Don't "fix" the above docstring to put it on two lines, as the sphinx
# autosummary pulls only the first line for its description.
```

Perhaps however we don't need that long of a basic description. I can split that long line into two separate lines instead like other docs.

PR #49688, opened 2018-09-17
Order the runtests suites

KaiSforza:
### What does this PR do?

Uses an ordered dictionary instead of a default python dict when
describing the `TEST_SUITES` variable. Sorts alphabetically based on the
name of the suite.

Fixes the test suite being run out of order, should make troubleshooting a bit easier for the tests, as they'll run in a consistent order (should see failures in consistent places).

### DIFF
--- a/tests/runtests.py
+++ b/tests/runtests.py
@@ -11,6 +11,7 @@ import os
 import sys
 import time
 import warnings
+import collections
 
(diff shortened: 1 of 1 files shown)


KaiSforza:
(had a typo in the commit)

dwoz (approved):

rallytime (approved):

PR #49696, opened 2018-09-18
loosen azure sdk dependencies in azurearm cloud driver

jgleissner:
### What does this PR do?
This PR removes the dependency on azure-cli from the azurearm cloud driver, which is not used at all. It also changes the driver to import the azure-storage sdk package in lieu of the multiapi storage package, in case the latter is not available.

Allows the azurearm cloud driver to be used on platforms that do not have the azure-cli or azure-multiapi-storage packages.

### Tests written?
Manually tested with azure-storage 0.36 and azure-multiapi-storage 0.1.4.

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/cloud/clouds/azurearm.py
+++ b/salt/cloud/clouds/azurearm.py
@@ -68,7 +68,6 @@ import salt.utils.data
 import salt.utils.files
 import salt.utils.stringutils
 import salt.utils.yaml
-from salt.utils.versions import LooseVersion
 from salt.ext import six
 import salt.version
 from salt.exceptions import (
--- a/salt/cloud/clouds/azurearm.py
+++ b/salt/cloud/clouds/azurearm.py
@@ -60,6 +60,7 @@ import logging
 import pprint
 import base64
 import collections
+import pkgutil
 import salt.cache
 import salt.config as config
 import salt.utils.cloud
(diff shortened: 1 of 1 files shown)


rallytime:
Welcome @jgleissner! Thank you for contributing.

Can you fix the lint error? https://jenkinsci.saltstack.com/job/pr-lint/job/PR-49696/1/warnings52Result/

salt/cloud/clouds/azurearm.py
```diff
@@ -117,9 +118,12 @@
     from azure.mgmt.storage import StorageManagementClient
     from azure.mgmt.web import WebSiteManagementClient
     from msrestazure.azure_exceptions import CloudError
-    from azure.multiapi.storage.v2016_05_31 import CloudStorageAccount
-    from azure.cli import core
-    HAS_LIBS = LooseVersion(core.__version__) >= LooseVersion("2.0.12")
+    if pkgutil.find_loader('azure.multiapi'):
```
rallytime: What about using the `salt.utils.path.which` function instead here?

rallytime (commented):
I have one small suggestion.

salt/cloud/clouds/azurearm.py
jgleissner: Hi @rallytime , if I'm not mistaken, `salt.utils.path.which` looks for an executable, whereas this code looks for an importable module. I don't see how `which` would be useful here, or am I missing something?

rallytime (approved):

rallytime:
Thank you @jgleissner!

PR #49697, opened 2018-09-18
Fix set_static_all, always display requestmode and refactor _change_state

alexvasiu:
* modules/nilrt_ip.py: Display requestmode even if service is offline
This will keep parity between older and newer nilrt distro.

* modules/nilrt_ip.py: Split _change_state function
Split the logic of _change_state function for older and newer disto.

* modules/nilrt_ip.py: Fix set_static_all
If there are no nameservers specified, we shouldn't set any property
for nameservers configuration to the service.

### DIFF
--- a/salt/modules/nilrt_ip.py
+++ b/salt/modules/nilrt_ip.py
@@ -510,6 +510,30 @@ def get_interfaces_details():
     return {'interfaces': list(map(_get_info, _interfaces))}
 
 
+def _change_state_legacy(interface, new_state):
+    '''
+    Enable or disable an interface on a legacy distro
+
+    Change adapter mode to TCP/IP. If previous adapter mode was EtherCAT, the target will need reboot.
+
+    :param interface: interface label
+    :param new_state: up or down
+    :return: True if the service was enabled, otherwise an exception will be thrown.
+    :rtype: bool
+    '''
+    initial_mode = _get_adapter_mode_info(interface)
+    _save_config(interface, 'Mode', 'TCPIP' if new_state == 'up' else 'Disabled')
+    if initial_mode == 'ethercat':
+        __salt__['system.set_reboot_required_witnessed']()
+    else:
+        out = __salt__['cmd.run_all']('ip link set {0} {1}'.format(interface, new_state))
+        if out['retcode'] != 0:
+            msg = 'Couldn\'t {0} interface {1}. Error: {2}'.format('enable' if new_state == 'up' else 'disable',
+                                                                   interface, out['stderr'])
+            raise salt.exceptions.CommandExecutionError(msg)
+    return True
+
+
 def _change_state(interface, new_state):
     '''
     Enable or disable an interface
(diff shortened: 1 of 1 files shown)


salt/modules/nilrt_ip.py
```diff
@@ -510,9 +513,33 @@ def get_interfaces_details():
     return {'interfaces': list(map(_get_info, _interfaces))}
 
 
-def _change_state(interface, new_state):
+def _change_state_older(interface, new_state):
```
rares-pop: we should go with 'legacy' instead of 'older', as it is in other modules.

I'd rename this to something like: _change_state_legacy()

salt/modules/nilrt_ip.py
```diff
@@ -510,9 +513,33 @@ def get_interfaces_details():
...
+    else:
+        out = __salt__['cmd.run_all']('ip link set {0} {1}'.format(interface, new_state))
+        if out['retcode'] != 0:
+            msg = 'Couldn\'t {0} interface {1}. Error: {2}'.format('enable' if new_state == 'up' else 'disable',
+                                                                   interface, out['stderr'])
+            raise salt.exceptions.CommandExecutionError(msg)
+    return True
+
+
+def _change_state_newer(interface, new_state):
```
rares-pop: if the other is _change_state_legacy(), this can be just _change_state()

rares-pop (changes requested):
we'd need some tests too for those

salt/modules/nilrt_ip.py
alexvasiu: Fixed

salt/modules/nilrt_ip.py
```diff
@@ -522,17 +549,7 @@ def _change_state(interface, new_state):
...
-        if initial_mode == 'ethercat':
-            __salt__['system.set_reboot_required_witnessed']()
-        else:
-            out = __salt__['cmd.run_all']('ip link set {0} {1}'.format(interface, new_state))
-            if out['retcode'] != 0:
-                msg = 'Couldn\'t {0} interface {1}. Error: {2}'.format('enable' if new_state == 'up' else 'disable',
-                                                                       interface, out['stderr'])
-                raise salt.exceptions.CommandExecutionError(msg)
-        return True
+        return _change_state_legacy(interface, new_state)
```
rares-pop: Please move this at the top of the function, so it's easier to follow what happens for legacy systems, vs the rest. Right now is in the middle of the function.

The code should look like: 

```
if legacy:
 return _change_state_legacy()
do the non-legacy behavior
```

rares-pop: Nevermind.. my bad!

rares-pop (approved):
Looks good!

alexvasiu:
@rallytime Can you take a look?

salt/modules/nilrt_ip.py
```diff
@@ -216,12 +216,15 @@ def _get_service_info(service):
...
 
-    if 'ipv4' in data:
-        data['ipv4']['supportedrequestmodes'] = [
-            'static',
-            'dhcp_linklocal'
-        ]
+    data['ipv4']['supportedrequestmodes'] = [
+        'static',
+        'dhcp_linklocal',
+        'disabled'
```
cachedout: Was switching the order of these intentional?

alexvasiu: Nope. I can sort them, if it's better

rallytime (approved):

PR #49704, opened 2018-09-18
INFRA-6465 - fix boto3_sns.subscribe() to accept Attributes as well

tkwilliams:
What the title says...  The previous patch I submitted supported ADDING attributes to existing subscriptions but missed the fact that the subscribe() function itself -- called when subscriptions are initially created -- had a hardcoded shortlist of allowed params, and Attributes wasn't in it.

Changed so that subscribe() now accepts any argument (present or future) valid to the underlying boto3 call.


salt/modules/boto3_sns.py
```diff
@@ -301,7 +302,7 @@ def set_subscription_attributes(SubscriptionArn, AttributeName, AttributeValue,
         return False
 
 
-def subscribe(TopicArn, Protocol, Endpoint, region=None, key=None, keyid=None, profile=None):
+def subscribe(region=None, key=None, keyid=None, profile=None, **args):
```
rallytime: This is going to cause an API change and should be put on a deprecation path to be removed later.

Since these are positional args, we don't want to break people's setups.

rallytime (changes requested):
Hey @tkwilliams! I have a request here so this doesn't break for people in the future.

The docs for deprecating code are [here](https://docs.saltstack.com/en/latest/topics/development/deprecations.html).

salt/modules/boto3_sns.py
tkwilliams: Ah, it didn't register that this would cause an API change...

I was only thinking in terms of keyword args (which are ALL I use, in both CLI and state defs) but for people passing them as ordered arguments I definitely see your concern.....

The alternative would be to add another explicit Attributes arg, as opposed to eating everything with **args.  This has the advantage of not breaking compat, but then it becomes a maintenance chore to add every new paramter as the underlying boto3 API expands...

This is exactly why I always use **args style in my own from-scratch boto modules :)

Anywho, I'll definitely follow up a.s.a.p. with your recommendation :)

tkwilliams: @rallytime 
What should I put in for the deprecation release version?

tkwilliams: @rallytime  grepping around in the code, it looked like Sodium was the likely choice here.  Let me know if something else is more appropriate :)

cachedout (approved):

cachedout:
@rallytime Re-review requested.

salt/modules/boto3_sns.py
```diff
@@ -301,7 +302,7 @@ def set_subscription_attributes(SubscriptionArn, AttributeName, AttributeValue,
         return False
 
 
-def subscribe(TopicArn, Protocol, Endpoint, region=None, key=None, keyid=None, profile=None):
+def subscribe(TopicArn, Protocol, Endpoint, region=None, key=None, keyid=None, profile=None, **kwargs):
```
rallytime: So since we're going to remove these in the future, shouldn't we make these optional now? Then check for their existence below?

tkwilliams: I can do that :)

rallytime (approved):

Issue #37949, opened 2016-11-29
Salt 2016.11.0 does not accept link-local IPv6 address in resolv.conf


Deewiant:
### Description of Issue/Question

Some servers of mine have link-local IPv6 addresses in `resolv.conf`, i.e. lines like `nameserver fe80::1234:5678:9abc:def0%eth0`. As of 2016.11.0 both master and minion log complaints like the following:

```
[ERROR   ] /etc/resolv.conf: 'fe80::21b:21ff:fe34:d66a%bond0' does not appear to be an IPv4 or IPv6 address
```

It doesn't seem to be breaking anything so far, though — just spamming logs. (Perhaps such nameserver addresses were never supported, but now it's more visible?)

### Setup
See description

### Steps to Reproduce Issue
See description

### Versions Report

```
Salt Version:
           Salt: 2016.11.0

Dependency Versions:
           cffi: Not Installed
       cherrypy: 8.1.2
       dateutil: 2.6.0
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.8
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: 0.24.0
           Mako: 1.0.6
   msgpack-pure: Not Installed
 msgpack-python: 0.4.8
   mysql-python: Not Installed
      pycparser: Not Installed
       pycrypto: 2.6.1
         pygit2: Not Installed
         Python: 2.7.12 (default, Nov  7 2016, 11:55:55)
   python-gnupg: Not Installed
         PyYAML: 3.12
          PyZMQ: 16.0.0
           RAET: Not Installed
          smmap: Not Installed
        timelib: Not Installed
        Tornado: 4.4.2
            ZMQ: 4.1.5

System Versions:
           dist:
        machine: x86_64
        release: 4.8.6-1-ARCH
         system: Linux
        version: Not Installed
```

Ch3LL:
okay it looks like it was added with this pr https://github.com/saltstack/salt/pull/22691

It seems that there is some issue with feature compatibiliy generally with using ipv6 local link addresses in /etc/resolv.conf as explained here: http://www.gossamer-threads.com/lists/nsp/ipv6/51213

So it seems we might need to add on our end the ability to support this format of an ipv6 address. Thanks

Deewiant:
I still see the exact same error message in the logs with 2018.3.2.

isbm:
https://github.com/saltstack/salt/pull/49705

PR #49705, opened 2018-09-18
Bugfix/rework: IPv6 scope errors

isbm:
### What does this PR do?

- Fixes IPv6 scope issues that aren't supported in Python `ipaddress`.
- Support IPv6 scope at `IPv6Address` object level.
- Reworks `salt._compat` for code duplication and other issues
- Fixes import of `ipaddress` across everywhere
- Uses proper `ipaddress` instance for Python 3
- Adds debug logging instead of just `pass`.

### Previous Behavior

If your `/etc/resolv.conf` has a scope in IPv6 address, something like this:

`nameserver fe80::20d:b9ff:fe01:ea8%eth0`

...then you are getting errors like this:
```
  2018-09-15 13:14:45,014 [salt.utils.dns   :1039][ERROR   ][845] 
  /etc/resolv.conf: u'fe80::20d:b9ff:fe01:ea8%eth0' does not
  appear to be an IPv4 or IPv6 address
```

This is hitting users since 2016: https://github.com/saltstack/salt/issues/37949

### New Behavior

:metal:  :sunglasses: :metal: 

### Tests written?

Existing should just pass.


isbm:
OK, Py2 seems unhappy, Py3 is just fine.

isbm:
@rallytime dat lint... JFYI: It should be muted, because we _have_ to shadow that method. But I can't find a number or a name of it to disable.

terminalmage (approved):

terminalmage:
Looks good aside from the lint failures.

cachedout (approved):
LGTM

terminalmage:
@isbm this should fix the lint failure. The failure was happening because you defined `_is_packed_binary` as a function and then separately also defined that attribute as a lambda in the `__init__()`.

```diff
diff --git a/salt/_compat.py b/salt/_compat.py
index c15cb84d80..3652655b7c 100644
--- a/salt/_compat.py
+++ b/salt/_compat.py
@@ -156,10 +156,6 @@ class IPv6AddressScoped(ipaddress.IPv6Address):
             ipaddress._BaseAddress.__init__(self, address)
             ipaddress._BaseV6.__init__(self, address)
         else:
-            # This method (below) won't be around for Python 3
-            # and this type is needed to be checked differently
-            self._is_packed_binary = lambda p: isinstance(p, bytes)
-
             # Python 3.4 fix. Versions higher are simply not affected
             # https://github.com/python/cpython/blob/3.4/Lib/ipaddress.py#L543-L544
             self._version = 6
@@ -185,6 +181,9 @@ class IPv6AddressScoped(ipaddress.IPv6Address):
         :param data:
         :return:
         '''
+        if sys.version_info.major != 2:
+            return isinstance(data, bytes)
+
         packed = False
         if len(data) == 16 and ':' not in data:
             try:
```

meaksh (approved):
Looks good :+1:

rallytime:
Hi @isbm! 😄 

There are some test failures that are related to this change:

- integration.ssh.test_jinja_filters.SSHJinjaFiltersTest.test_network_ipaddr
- unit.utils.test_jinja.TestCustomExtensions.test_ipaddr
- unit.utils.test_jinja.TestCustomExtensions.test_ipv4
- unit.utils.test_jinja.TestCustomExtensions.test_ipv6
- unit.utils.test_jinja.TestCustomExtensions.test_is_ip
- unit.utils.test_jinja.TestCustomExtensions.test_is_ipv4
- unit.utils.test_jinja.TestCustomExtensions.test_is_ipv6

Can you take another look? 😎 

https://jenkinsci.saltstack.com/job/pr-kitchen-centos7-py3/job/PR-49705/9/

gtmanfred (approved):

aphor:
FWIW, upstream in Python we have precedent to suggest upstream fixes in ipaddress module.
 * https://bugs.python.org/issue23448
 * https://bugs.python.org/issue29375

so I submitted https://bugs.python.org/issue34788

isbm:
@aphor this doesn't really help, I've been there already. First, the issue is like three years old. Second, even they will fix in some upcoming versions, we still need this even on 2.7 and SUSE on 2.6 that is just simply EOL.

aphor:
@isbm I agree completely salt should not wait for upstream fix, but this is technical debt which would ideally be factored out of salt, and FWIW, salt isn't the only Python utility that I need to support IPv6.

I put that stuff here, mostly because it will be useful in the future to have it here, for reference, not to imply we should not merge this PR into salt.

isbm:
@aphor after @rallytime puts that in (I am also going to backport that to 2018.3) let's look again at your PR. I am keeping an eye on yours and if you need any help, please ping me.

twangboy (approved):

avolmensky:
I have been playing around with 2019.02 and the netbox pillar module is broken with Python 3. The changes in this PR fix the problem I have been having. Will this be merged into the 2019.02 release?

PR #49706, opened 2018-09-18
[2017.7] Removing os.path.join mocking from tests.

garethgreenaway:
### What does this PR do?
Removing mocking of os.path.join from two tests, one of which was failing when used with coverage argument.  No reason we should be mocking os.path.join.

https://github.com/saltstack/salt-jenkins/issues/1120

### Previous Behavior
Test was failing.

### New Behavior
Test now succeeds.

### Tests written?
Yes.  Updated existing tests.

### Commits signed with GPG?

Yes/No


rallytime (approved):
👍

terminalmage (approved):

PR #49717, opened 2018-09-20
RFC: netapi support for asynchronous batching command execution

moio:
### What does this PR do?

RFC about supporting asynchronous batching command execution through the NetApi.

### DIFF
--- a/rfcs/0002-netapi-async-batch.md
+++ b/rfcs/0002-netapi-async-batch.md
@@ -29,8 +29,8 @@ Please note that the `Batch` class currently works synchronously only - this is 
 We propose to move the handling of batches into `salt-master`, so that long-term there is only one implementation and all clients/NetApi clients get support "for free".
 
 Implementation ideas:
- - a new parameter is added to the [publish method in `ClearFuncs`](https://github.com/saltstack/salt/blob/develop/salt/master.py#L2035) which is the receiving endpoint for commands from clients, or a new method is created (eg. `publish_batch`)
- - new code in `ClearFuncs` will implement the batching and call `publish` multiple times
+ - a new method is added to `ClearFuncs` that is called in place of the [publish](https://github.com/saltstack/salt/blob/develop/salt/master.py#L2035) for asynchronous batched command execution (eg. `publish_batch`). This method will receive batch-specific parameters and use `publish` internally
+ - it is assumed the [publish](https://github.com/saltstack/salt/blob/develop/salt/master.py#L2035) will not be changed. Further, separate discussion might be needed otherwise
  - implementation in principle follows from the existing `cli.Batch` class and makes it nonblocking/Tornado-aware, and removes the dependency from the client implementation
  - long term, all endpoints use this new code and `cli.Batch` is removed. Immediate goal would be to add asynchronous, batching mode to `rest_cherrypy` only (conforming to the now-deleted `rest_tornado` implementation noted above)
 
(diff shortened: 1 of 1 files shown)


rfcs/0002-netapi-async-batch.md
```diff
@@ -0,0 +1,55 @@
...
+
+Please note that the `Batch` class currently works synchronously only - this is of course OK for the CLI, it's acceptable for direct `LocalClient` use by third-party Python code, it's probably also acceptable for `rest_cherrypy` (although asynchronous calls are not supported) or direct `LocalClient` but it's definitely not adequate for `rest_tornado`.
+
+We propose to move the handling of batches into `salt-master`, so that long-term there is only one implementation and all clients/NetApi clients get support "for free".
+
+Implementation ideas:
+ - a new parameter is added to the [publish method in `ClearFuncs`](https://github.com/saltstack/salt/blob/develop/salt/master.py#L2035) which is the receiving endpoint for commands from clients, or a new method is created (eg. `publish_batch`)
```
cachedout: If we're going to do this, I think it's cleaner to split it out into multiple methods and keep `publish` as it is.

isbm: @cachedout I am second to this. Current publish method is already huge and begs refactoring. I'd say prior to this we should also refactor `publish` on its own. Only then we would be ready to go with this. And so I would indeed add another method than can publish in a batch.

Then we also could have an ability to Salt decide start "batching" after certain threshold without being explicitly configured to do so.

isbm (commented):
In general looks good. But like @cachedout pointed out, prior this needs to be separately refactored, so another method can reuse current publish bits.

rfcs/0002-netapi-async-batch.md
moio: If the implementation of this RFC requires changes to `publish`, I might agree that refactoring `publish` first could be a good idea.

If OTOH the implementation of this RFC uses `publish` as-is, I do not see a reason to withhold it until `publish` is refactored because of unrelated historical reasons. `publish` refactoring would be a separate task that could be equally be tackled before, after or even in parallel.

This is sometimes referred to as the "clean the kitchen, don't clean the garage" principle - is this acceptable?

isbm: You are going to reuse publish's parts. Without prior refactoring (splitting to reusable methods etc) you will either doing it at the same time (and it is not a good idea) or duplicate code (not a good idea, again).

Sure, if you think that new publish method won't have any of the code duplication at all, go for it.

moio: Clarified in new commit, marking discussion resolved

isbm: @cachedout are you OK with this?

rallytime (approved):
I like this idea, and this proposal looks good to me.

In terms of unresolved questions, we have been discussing a desire to improve the support/functionality of `rest_tornado` to have feature parity with `rest_cherrypy`. This is only in the early discussion stages though and we haven't made any firm plans yet.

rallytime:
@whiteinge You might be interested in this RCF proposal. :)

moio:
> we have been discussing a desire to improve the support/functionality of rest_tornado to have feature parity with rest_cherrypy. This is only in the early discussion stages

We have a similar/specular situation in Uyuni - on one hand we would like to migrate away from `rest_cherrypy` to `rest_tornado`, on the other hand we lack some important bits there - we could recently work around the absence of the Web Socket/event endpoint, the next biggest obstacle is Salt-SSH support. It would be awesome if we could make this happen!

whiteinge:
Solid proposal 👍.

Using NetapiClient for both sync and async interfaces (and in the future: non-blocking async) for all the main interfaces in Salt is very forward-thinking since there really needs to be a central place in Salt's Python API to facilitate easily building external APIs with eauth without having to wrap everything from scratch each time. E.g., if rest_tornado shared more of these internals it would be easier to switch off rest_cherrypy ('course that would've required non-blocking implementations, which we still don't have). Future netapi additions will also benefit from this work -- a raw websockets interrface would be useful, an external zeromq interface too, etc.

gtmanfred (approved):

PR #49719, opened 2018-09-20
Get proper endpoint for password verification in keystone module

msciciel:
Keystone module could be configured to use admin token which use
option keystone.endpoint and without admin token which use
keystone.auth_url. Function user_verify_password should distinguish
these cases and get endpoint from proper option. Otherwise password
will be always updated if admin token is not used causing tokens
revocation.

### What does this PR do?
Get proper keystone endpoint in function user_verify_password of keystone module.

### Previous Behavior
When admin token is not used then user password is always updated causing tokens revocation.

### New Behavior
When admin token is not used then user password is updated only if changed.

### Tests written?
No

### Commits signed with GPG?
No

### DIFF
--- a/salt/modules/keystone.py
+++ b/salt/modules/keystone.py
@@ -1124,11 +1124,15 @@ def user_verify_password(user_id=None, name=None, password=None,
     if 'connection_endpoint' in connection_args:
         auth_url = connection_args.get('connection_endpoint')
     else:
+        auth_url_opt = 'keystone.auth_url'
+        if __salt__['config.option']('keystone.token'):
+            auth_url_opt = 'keystone.endpoint'
+
         if _OS_IDENTITY_API_VERSION > 2:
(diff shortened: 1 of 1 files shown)


rallytime (approved):
This LGTM, but I'd like @gtmanfred to give the final review here.

gtmanfred (approved):
This makes sense, can you also check that the new keystoneng.py modules work correctly? because this keystone module is being deprecated in favor of the shade modules.

msciciel:
As I understand `keystoneng` module use `shade` module which handle user update. `update_user` function from https://github.com/openstack-infra/shade/blob/master/shade/openstackcloud.py#L856 propably will always update password and cause token revocation which is not good situation because could cause some errors on cloud operations ie. when nova user token are revoked. I will verify this next week.

rallytime:
@msciciel Thanks! I will get this one in now and you can make those other changes in a separate PR. Thank you for your work and contributions here!

Issue #49709, opened 2018-09-19
Useless requirement for prlsrvctl in Parallels module


cstarke:
We have machines running PVC 7 without the installation of prlsrvctl.
The module only requires prlsrvctl in one function, which is prlsrvctl(), but the whole module will error if there's no prlsrvctl binary found.

https://github.com/saltstack/salt/blob/ab7dc68c3b033fa46c38fe59ab77bf7d1df36e68/salt/modules/parallels.py#L52

IMHO it would be good to split those dependencies in different functions (like _has_prlctl() and _has_prlsrvctl()) and check within executed functions whether those requirements are met or to omit those checks completely and rely on sane output of the underlying cmd.run() call. ("tell don't ask")

Is that possible? I would create a PR for that if one of my idea is okey or you tell me about another solution.

Thanks :)

Ch3LL:
yeah we would accept that :) if only one module requires that dependency then we only need to do that check in that function.

PR #49720, opened 2018-09-20
Seperate prlctl and prlsrvctl checks into each requiring function

Pie4kaiz:
### What does this PR do?
Split and move the availability checks for prlctl or prlsrvctl to each function requiring it.

https://github.com/saltstack/salt/issues/49709

### Previous Behavior
Loading the Parallels module will fail if the prlsrvctl binary can't be found, whereas prlsrvctl is only required for one function: prlsrvclt().
All other functions can't be used.

### New Behavior
Only prlsrvctl() requires the prlsrvctl binary.
prlctl() (invoked by all other functions) require the prlctl binary.
They can fail with a CommandExecutionError.

### Tests written?
No

### Commits signed with GPG?
No

### DIFF
--- a/salt/modules/parallels.py
+++ b/salt/modules/parallels.py
@@ -29,7 +29,7 @@ import shlex
 import salt.utils.locales
 import salt.utils.path
 import salt.utils.yaml
-from salt.exceptions import SaltInvocationError
+from salt.exceptions import SaltInvocationError, CommandExecutionError
 
 # Import 3rd party libs
 from salt.ext import six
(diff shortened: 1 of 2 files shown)


cachedout (approved):

rallytime (commented):
This looks fine to me, but I think this would be a good opportunity to document the requirements more clearly as dependencies.

I see at the top of the file there are statements about what is or is not built with the various prlctl/prlsrvctl commands, but we don't say that either library is a dependency.

Can you add that to the docs?

rallytime:
@cstarke Welcome! I have a small request above.

It also looks like this change is causing some tests to fail. Can you take a look?

https://jenkinsci.saltstack.com/job/pr-kitchen-ubuntu1604-py2/job/PR-49720/2/

rallytime:
@cstarke Sorry, I was not being clear.

I just mean that we should document somewhere that some functions require one library or the other. 

That way people don't have to run the function and hit the `CommandExecutionError`. It should be stated clearly at the top. Something like:
```
This module requires the prlctl binary to be installed to run most functions.

To run parallels.prlsrvctl, the prlsrvctl binary is required.
```
Or something along those lines.

rallytime (approved):
Thank you @cstarke!

Issue #47614, opened 2018-05-11
[Bug] Scheduler for influxdb module fails on salt 2018.3


bbczeuz:
### Setup
* Debian9 on amd64
* salt-minion from saltstack's repo

### Steps to Reproduce Issue
* add a scheduler state:
```
job_fetch_influx_success:
    args: [telegraf, SELECT value FROM telegraf.apache2 WHERE time > now() -
        10m GROUP BY *]
    enabled: true
    function: influxdb.query
    jid_include: true
    maxrunning: 1
    name: job_fetch_influx_value
    seconds: 300
```
* Job will always fail with
```
2018-05-11 18:11:45,013 [salt.utils.schedule:737 ][ERROR   ][27136] Unhandled exception running influxdb.query
Traceback (most recent call last):
  File "/usr/lib/python2.7/dist-packages/salt/utils/schedule.py", line 704, in handle_func
    ret['return'] = self.functions[func](*args, **kwargs)
  File "/usr/lib/python2.7/dist-packages/salt/modules/influx.py", line 699, in query
    client = _client(**client_args)
  File "/usr/lib/python2.7/dist-packages/salt/modules/influx.py", line 73, in _client
    if __pub_fun_args in client_args:
NameError: global name '__pub_fun_args' is not defined
```
* Manually running 'salt minionname influx.query "SELECT *...."' always succeeds
* Dictionary ```client_args``` has elements  ```'__pub_id','__pub_fun_args','__pub_schedule'``` which python-influxdb's constructor does not like.
* I created a patch to modules/influx.py to delete the elements. I'm not sure where to fix this:
  * in modules/influx.py
  * in state.py (variable STATE_INTERNAL_KEYWORDS)
* I can open a pull request with a fix if I get an answer to the above question. Anyone?

### Versions Report
```
salt-call --versions-report
Salt Version:
           Salt: 2018.3.0
 
Dependency Versions:
           cffi: Not Installed
       cherrypy: Not Installed
       dateutil: 2.7.3
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.9.4
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.4.8
   mysql-python: Not Installed
      pycparser: Not Installed
       pycrypto: 2.6.1
   pycryptodome: Not Installed
[...]

PR #49721, opened 2018-09-20
influx: Drop unhandled kwargs for InfluxDBClient initializer

bbczeuz:
### What does this PR do?
It drops selected `kwargs` from the InfluxDBClient initializer. They are unhandled and InfluxDBClient throws an exception.
I'm not sure if filtering on a module level is the best thing to do, as this issue might happen with other modules, too. Therefore I provide an alternate diff that drops the kwargs at the central place:
```
diff --git a/salt/state.py b/salt/state.py
index c423066702..1785922262 100644
--- a/salt/state.py
+++ b/salt/state.py
@@ -115,8 +115,11 @@ STATE_RUNTIME_KEYWORDS = frozenset([
     '__orchestration_jid__',
     '__pub_user',
     '__pub_arg',
+    '__pub_id',
     '__pub_jid',
     '__pub_fun',
+    '__pub_fun_args',
+    '__pub_schedule',
     '__pub_tgt',
     '__pub_ret',
     '__pub_pid',
```

#47614

### Previous Behavior
influxdb module crashes during initialization

### New Behavior
influxdb module usable

### Tests written?

No

### Commits signed with GPG?

No


cachedout:
I don't think that filtering at the module level is the way to go here. This will almost assuredly break at some point in the future. I think we should try filtering at the state level as you suggest and see how the test suite likes it.

cachedout:
@bbczeuz Did you see my comment above?

bbczeuz:
@cachedout Sorry, I missed that. I'll rebase and adapt the PR to filter on the state level and push it for testing. Expect an update in the next days.

cachedout:
@bbczeuz Sounds great. Thank you!

bbczeuz:
Now, the params are filtered at STATE_INTERNAL_KEYWORDS in salt/state.py. This dict is used by several other states (nftables, iptables, pkgrepo) and modules (x509, nftables, iptables) for the same purpose. Let's see how this behaves in testing.

cachedout:
Doesn't look like things went too well with the test suite, sadly.

cachedout:
@bbczeuz We'll need you to please investigate the test failures here before we can merge this.

dwoz:
@bbczeuz Are you able to look at the test failures?

dwoz:
@cachedout The tests passed this go around. Do you approve this change?

cachedout (approved):

PR #49728, opened 2018-09-20
zsh-completion: strip out u'' unicode __repr__ of function names

llua:
The elements in the array returned by sys.list_{,runners_}functions
recently started to be enclosed by u'' so when the completer parses
salt-call --local --out txt sys.$func, the incorrect module is
presented.

### DIFF
--- a/pkg/zsh_completion.zsh
+++ b/pkg/zsh_completion.zsh
@@ -25,7 +25,7 @@ _modules(){
     fi
 
     if _cache_invalid salt/modules || ! _retrieve_cache salt/modules; then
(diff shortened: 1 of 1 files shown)


gtmanfred (approved):

gtmanfred:
this should be backported to 2018.3

rallytime:
@llua This wouldn't back-port cleanly to 2018.3, so I back-ported it to fluorine instead. Can you make the change directly against 2018.3 if you need the fix in a release sooner than fluorine?

roaldnefs:
Already ported to master in  #49775.

Issue #49734, opened 2018-09-21
[grafana4_user] Set user organization/role. Make email optional


bbczeuz:
### Description of Issue/Question
* The current grafana4_user state does not allow a user to be added to organizations. The current approach is to add users to organizations when creating an organization. This asks for two states to run, instead of doing everything in one state and keep all user relevant information at one place.
* The email parameter is required by the grafana4_user.present module. For Grafana itself if the logon name (name in grafana4_user state) is given, email is optional.
* There is no state function to set a user's default organization

### Setup
* Grafana 4.5.2
* Salt 2018.3.2
* Debian9

### Steps to Reproduce Issue
test/add_grafana_org.sls
```
test grafana org:
  grafana4_org.present:
    - name: 'TestOrg'
    - users:
        TestUser: Admin
```

test/add_grafana_user.sls
```
test grafana user:
  grafana4_user.present:
    - name: 'TestUser'
    - password: 123456789
    - fullname: blabla
```

### Versions Report
```
Salt Version:
           Salt: 2018.3.2
 
Dependency Versions:
           cffi: Not Installed
       cherrypy: Not Installed
       dateutil: 2.5.3
      docker-py: 1.9.0
          gitdb: 2.0.0
      gitpython: 2.1.1
          ioflo: Not Installed
         Jinja2: 2.9.4
        libgit2: 0.24.5
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.4.8
   mysql-python: Not Installed
      pycparser: Not Installed
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: 0.24.2
         Python: 2.7.13 (default, Nov 24 2017, 17:33:09)
   python-gnupg: Not Installed
         PyYAML: 3.12
          PyZMQ: 16.0.2
           RAET: Not Installed
          smmap: 2.0.1
        timelib: Not Installed
        Tornado: 4.4.3
            ZMQ: 4.2.1
 
System Versions:
           dist: debian 9.5 
         locale: UTF-8
        machine: x86_64
        release: 4.9.0-8-amd64
         system: Linux
        version: debian 9.5 
```

PR #49735, opened 2018-09-21
[grafana4_user] Support adding users to organizations, roles and set default organization

bbczeuz:
### What does this PR do?
* Catch multiple unhandled exceptions and return nicely formatted error messages
* Allow setting a user's default organization
* Allow setting user's member organizations and roles
* Make `'email` parameter optional

#49734 

### Previous Behavior
See issue #49734 

### New Behavior
```
test grafana org:
  grafana4_org.present:
    - name: 'TestOrg'
test grafana user:
  grafana4_user.present:
    - name: 'TestUser'
    - password: 1234567689
    - fullname: blabla
    - default_organization: TestOrg
    - organizations:
      - TestOrg: Admin
      - ReadonlyOrg
      - AnotherOrg: Viewer
      - Staging: Editor
```

### Tests written?
Manually tested. No unittests available

### Commits signed with GPG?

No


salt/modules/grafana4.py
```diff
@@ -183,7 +183,7 @@ def create_user(profile='grafana', **kwargs):
     return response.json()
 
 
-def update_user(userid, profile='grafana', **kwargs):
+def update_user(userid, profile='grafana', orgid=None, **kwargs):
```
cachedout: IMHO, we should consider making this `org_id` instead to make it easier to identify.

salt/states/grafana4_user.py
```diff
@@ -53,10 +54,12 @@ def __virtual__():
 
 def present(name,
             password,
-            email,
+            email=None,
```
cachedout: It looks like you're making this optional but I don't see this change reflected in the docstring below.

salt/modules/grafana4.py
bbczeuz: If I change it here, for consistency, I'd probably better change this anywhere. (orgid->org_id, userid->user_id, orgname->org_name)
This would lead to the 117 changes and is potential breaking change:
```
# git diff --numstat
105     105     salt/modules/grafana4.py
7       7       salt/states/grafana4_dashboard.py
6       6       salt/states/grafana4_datasource.py
5       5       salt/states/grafana4_user.py
```

go/no-go?

salt/states/grafana4_user.py
bbczeuz: fixed in bbczeuz/salt@5f63c595bba1ebe61bd26f73024ed370037509b5

salt/modules/grafana4.py
cachedout: Good point. I agree with you then. Let's leave it. :)

cachedout (changes requested):
We need these lint errors fixed before we can merge this, please: https://jenkinsci.saltstack.com/job/pr-lint/job/PR-49735/6/warnings52Result/new/

bbczeuz:
@cachedout : Some help needed: I fixed all but one linter errors. What is the recommended way to replace ```from requests.exceptions import HTTPError```. I only use this import to catch and nicely handle exceptions of the underlying grafana4 module. Does salt provide a python 2/3 compatible alternate or shall I use urllib instead?

bbczeuz:
I went through codeclimate's results and don't see much potential for improvement (other than what I did in the previous few commits). Please let me know if you want me to fix this.

cachedout (approved):

Issue #49604, opened 2018-09-11
Misunderstood file.prepend state behavior on dry-run


dzhukov89:
### Description of Issue
Hello

E.g. we have salt formula like this:

```
my_formula:
  file.prepend:
    - name: ~/output.txt
    - makedirs: True
    - text: |
        lorem ipsum blablabla
```

When I call " state.highstate test=True" I always get "Result: False":

```
ID: my_formula
Function: file.prepend
Name: ~/output.txt
Result: False
Comment: /root/output.txt: file not found
Started: 14:14:12.353606
Duration: 1.17 ms
Changes:   
```

But when I call only "  state.highstate  " I get:
```
ID: my_formula
Function: file.prepend
Name: ~/output.txt
Result: True
Comment: Prepended 1 lines
Started: 15:35:49.485290
Duration: 3.449 ms
Changes:   
              ----------
              diff:
                  ---  
                  +++  
                  @@ -1,0 +1,1 @@
                  +lorem ipsum blablabla
```

### Question
**Question**: Is it a bug or a feature?

### Report
```
Salt Version:
           Salt: 2016.3.3

Dependency Versions:
           cffi: Not Installed
       cherrypy: Not Installed
       dateutil: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.2.1
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: 0.20.2
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.4.6
   mysql-python: Not Installed
      pycparser: Not Installed
       pycrypto: 2.6.1
         pygit2: Not Installed
         Python: 2.6.6 (r266:84292, Aug 18 2016, 15:13:37)
   python-gnupg: Not Installed
         PyYAML: 3.10
          PyZMQ: 14.5.0
           RAET: Not Installed
          smmap: Not Installed
        timelib: Not Installed
        Tornado: 4.2.1
            ZMQ: 4.0.5

```

garethgreenaway:
@dzhukov89 Thanks for the report. This appears to be a bug. We'll get this fixed.

dzhukov89:
@garethgreenaway you're welcome :)
Don't forget to delete output.txt before reproduce (dry-run)

PR #49736, opened 2018-09-21
Fix test opts in append, prepend states/file.py

mattLLVW:
### What does this PR do?
This PR fixes the test option when trying to append or prepend a file

#49604 
### Previous Behavior

```
Result: False
Comment: /root/output.txt: file not found
```

### New Behavior

```
Result: None
Comment: File /root/output.txt is set to be created
```

### Tests written?

No

### Commits signed with GPG?

No

### DIFF
--- a/salt/states/file.py
+++ b/salt/states/file.py
@@ -4905,7 +4905,9 @@ def append(name,
     check_res, check_msg = _check_file(name)
     if not check_res:
         # Try to create the file
-        touch(name, makedirs=makedirs)
+        touch_ret = touch(name, makedirs=makedirs)
+        if __opts__['test']:
+            return touch_ret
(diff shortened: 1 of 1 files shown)


terminalmage (changes requested):
Test mode should never result in changes actually being made. Instead we should exit with a `None` result and say what we would have done.

mattLLVW:
@terminalmage  that's precisely what it does.
 
```
Result: None
Comment: File /root/output.txt is set to be created
```

terminalmage (approved):
Ahh yes, because `__opts__['test']` will still be `True` when `file.touch` is run.

This is just not how we normally handle this sort of thing, so it looked wrong.

Issue #49660, opened 2018-09-14
win_system.join_domain returning empty change dictionary


hrumph:
### Description of Issue/Question
win_system.join does not return a changes dictionary even if changes are made.

dwfreed:
I don't think you intended to tag me.  Perhaps you meant dwoz?  (I've never touched either the win_system execution module or the state module)

garethgreenaway:
@dwfreed Apologies. Wasn't paying attention to the auto-complete. :)

hrumph:
versions report is as follows:
```
Salt Version:
           Salt: 2018.3.2
 
Dependency Versions:
           cffi: 1.10.0
       cherrypy: Not Installed
       dateutil: 2.6.0
      docker-py: Not Installed
          gitdb: 2.0.3
      gitpython: 2.1.7
          ioflo: Not Installed
         Jinja2: 2.10
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: 0.28.2
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.5.6
   mysql-python: Not Installed
      pycparser: 2.14
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: Not Installed
         Python: 2.7.15 (default, May 15 2018, 15:37:31)
   python-gnupg: Not Installed
         PyYAML: 3.13
          PyZMQ: 17.1.2
           RAET: Not Installed
          smmap: 2.0.3
        timelib: Not Installed
        Tornado: 5.1
            ZMQ: 4.2.5
 
System Versions:
           dist: fedora 27 Twenty Seven
         locale: UTF-8
        machine: x86_64
        release: 4.17.14-102.fc27.x86_64
         system: Linux
        version: Fedora 27 Twenty Seven
```
the salt minion version is also 2018.3.2 (64 bit Python 3)

hrumph:
@Twangboy, I confess that I haven't looked closely at the code (yet). All I know is that when the state is ran I get output like this:

```
          ID: join domain
    Function: system.join_domain
        Name: AD.QUEENSU.CA
      Result: True
     Comment: Computer added to 'AD.QUEENSU.CA'
     Started: 15:32:08.403369
    Duration: 3955.51 ms
     Changes:   
```

Note that the  _Changes:_ subdictionary is empty, even though a change did occur. I would think that I should see the standard _New_ and _Old_ subsubdictionaries.

twangboy:
[...]

PR #49739, opened 2018-09-21
Populate changes dictionary on system.join_domain

twangboy:
### What does this PR do?
Fixes issue with the `system.join_domain` state where the changes dictionary was not being populated when changes were made to the system.

- Populate the changes dictionary
- Fix docs formatting
- Clean up code
- Use explicit function calls
- Add comment about reboot status

https://github.com/saltstack/salt/issues/49660

### Tests written?
No

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/states/win_system.py
+++ b/salt/states/win_system.py
@@ -162,36 +162,46 @@ def hostname(name):
     return ret
 
 
-def join_domain(name, username=None, password=None, account_ou=None,
-                account_exists=False, restart=False):
-
+def join_domain(name,
+                username=None,
+                password=None,
+                account_ou=None,
+                account_exists=False,
+                restart=False):
     '''
-    Checks if a computer is joined to the Domain.
-    If the computer is not in the Domain, it will be joined.
+    Checks if a computer is joined to the Domain. If the computer is not in the
+    Domain, it will be joined.
 
-    name:
-        The name of the Domain.
+    Args:
 
-    username:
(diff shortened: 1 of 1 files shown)


rallytime (approved):

hrumph:
Thanks @Twangboy. I have to point out that I think that the restart switch for this function should probably be deprecated, since we now have the system.reboot state (which is supposed to know about the pending domain join). Also the restart switch comes with no options, which means that it always has a five minute timeout. A sysadmin doing a domain join typically won't want a five minute timeout at that time since joining is usually a deployment step. Even if the reboot flag isn't deprecated it should result in an immediate reboot.

dwoz:
@twangboy Can you look into this test failure?

`integration.modules.test_groupadd.GroupModuleTest.test_members`

twangboy:
@dwoz The above failure is not related to this PR. However, I will look at the test to see if maybe it's flaky.

twangboy:
@dwoz I can't replicate it locally, but it's doing several `run_function`s. Maybe one of them is failing. I modified the test so that it actually checks the results of the `group.members` function on Windows.

https://github.com/saltstack/salt/pull/49794

rallytime:
@dwoz since @twangboy is addressing that test concern in another PR, how does this look to you now?

dwoz (approved):

dwoz:
@rallytime That seems okay to me.

Issue #49744, opened 2018-09-22
Group state cannot force empty group membership


MTecknology:
### Description of Issue/Question

As reported on IRC, the following state will produce no changes to group membership.

```
shadow:
  group.present:
    - members: []
```

This happens because the group state checks for false-like values and an empty list is false-like, producing a no-op change.

Desired Outcome: Group 'shadow' has no members.

Realized Outcome: Group 'shadow' member list in unchanged.

### Versions Report

```
root@saltmaster:/srv/salt/system/cis# salt-call --versions-report
Salt Version:
           Salt: 2018.3.2

System Versions:
           dist: Ubuntu 16.04 xenial
         locale: UTF-8
        machine: x86_64
        release: 4.4.0-116-generic
         system: Linux
        version: Ubuntu 16.04 xenial
```

Ch3LL:
Thanks for the PR!

PR #49745, opened 2018-09-22
Remove all users from group when members list is empty.

MTecknology:
### What does this PR do?

This patch allows an empty list to be supplied to the members argument for the ```group.present``` state.

- #49744

### Previous Behavior

An empty list would be interpreted as a False-like value and cause no-op during group members check.

### New Behavior

An empty members list will invoke the remove_user behavior for all list members.

### Tests written?

No

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/states/group.py
+++ b/salt/states/group.py
@@ -83,14 +83,13 @@ def _changes(name,
             ret['comment'] = 'Invalid gid'
             return ret
 
-    if members is not None:
-        if members:
-            # -- if new member list if different than the current
-            if set(lgrp['members']).symmetric_difference(members):
-                change['members'] = members
-        else:
-            if set(lgrp['members']).symmetric_difference(members):
-                change['delusers'] = set(lgrp['members'])
+    if members is not None and not members:
+        if set(lgrp['members']).symmetric_difference(members):
+            change['delusers'] = set(lgrp['members'])
+    elif members:
+        # if new member list if different than the current
+        if set(lgrp['members']).symmetric_difference(members):
+            change['members'] = members
 
     if addusers:
         users_2add = [user for user in addusers if user not in lgrp['members']]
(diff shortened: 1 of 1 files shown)


hemebond (approved):
Tested changes in 2017.7.7. Users correctly removed when members list is empty.

salt/states/group.py
```diff
@@ -72,21 +72,24 @@ def _changes(name,
...
     if gid:
         try:
             gid = int(gid)
             if lgrp['gid'] != gid:
                 change['gid'] = gid
         except (TypeError, ValueError):
-            ret['result'] = False
-            ret['comment'] = 'Invalid gid'
-            return ret
+            return {
```
cachedout: I'm not following the need for this change. Could you clarify, please?

MTecknology: Eh, similar to the elif: vs. if:continue change below, I should probably have held off until offering this for the develop branch. I'll revert both so that this is only a fix for the reported issue.

My logic for dropping ret={} was that ret is used nowhere else and logically wouldn't pop up. There's an empty dict created, manipulated, and returned instead of just returning the dict that needs to be returned.

salt/states/group.py
```diff
@@ -83,10 +83,14 @@ def _changes(name,
             ret['comment'] = 'Invalid gid'
             return ret
 
-    if members:
-        # -- if new member list if different than the current
-        if set(lgrp['members']).symmetric_difference(members):
-            change['members'] = members
+    if members is not None:
```
cachedout: Given L87, couldn't this just be `if members:` and then move the else block out to align with this conditional?

MTecknology: The first if checks whether any value was supplied to members, the second if checks whether it's an empty list.

I could probably turn it into something like:
```
if members is not None and not members:
    [...]
elif members:
    [...]
```

Would that be preferable?

gtmanfred (approved):

rallytime (approved):

PR #49750, opened 2018-09-23
Fix typo (exsiting -> existing)

jose1711:


MTecknology (approved):

gtmanfred (changes requested):
please revert all changes to doc/man/salt.7, we manage that from the other files, and it only gets updated with releases.

rallytime:
The man page changes are fine here, but just be aware that any changes made there will be overwritten the next time we generate the man pages (which we do before each release).

rallytime:
Thanks for contributing @jose1711!

PR #49757, opened 2018-09-24
Fix influxdb_user state when test=True

bbinet:
this fixes the error "ValueError: too many values to unpack"


rallytime (approved):
Good catch

Issue #49730, opened 2018-09-20
[Windows] network.managed cannot set interface with blank DNS servers


omgitsedwin:
### Description of Issue/Question

Windows allows leaving the DNS servers empty if DNS is not used. However, Salt requires DNS servers when using network.managed states. This seems to be an old issue #18513.

### Setup
The following .sls:
```
Local Area Connection:
  network:
    - managed
    - dns_proto: static
    - dns_servers: []
    - ip_proto: static
    - ip_addrs:
      - 10.8.1.32/24
```
results in:
```
          ID: Local Area Connection
    Function: network.managed
      Result: False
     Comment: The following SLS configuration errors were detected: The dns_servers param is required to set static DNS servers.
     Started: 19:38:09.632453
    Duration: 32.004 ms
     Changes:   
```
Likewise:
```
Local Area Connection:
  network:
    - managed
    - dns_proto: None
    - ip_proto: static
    - ip_addrs:
      - 10.8.1.32/24
```
results in:
```
          ID: Local Area Connection
    Function: network.managed
      Result: False
     Comment: dns_proto must be one of the following: static, dhcp.
     Started: 19:48:42.866864
    Duration: 0.5 ms
     Changes:   
```
### Steps to Reproduce Issue
Any way to get around putting in DNS servers for Windows minion in .sls file? 

### Versions Report
```
Salt Version:
           Salt: 2018.3.2
 
Dependency Versions:
           cffi: Not Installed
       cherrypy: Not Installed
       dateutil: Not Installed
      docker-py: Not Installed
          gitdb: 0.6.4
      gitpython: 1.0.1
          ioflo: Not Installed
         Jinja2: 2.7.2
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.5.6
   mysql-python: Not Installed
      pycparser: Not Installed
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: Not Installed
         Python: 2.7.5 (default, Aug  4 2017, 00:39:18)
   python-gnupg: Not Installed
         PyYAML: 3.11
          PyZMQ: 15.3.0
           RAET: Not Installed
          smmap: 0.9.0
        timelib: Not Installed
        Tornado: 4.2.1
            ZMQ: 4.1.4
 
System Versions:
           dist: centos 7.4.1708 Core
         locale: UTF-8
[...]

PR #49763, opened 2018-09-24
Add ability to clear DNS entries on static DNS

twangboy:
### What does this PR do?
Adds the ability to clear the list of DNS entries in Windows by passing an empty list.

Execution Module:
- Passing no DNS servers will clear the list because it is using *args.
- Update documentation
- Added a note about clearing DNS entries

State Module
- An empty list in `dns_servers` will clear the entires
- `None` in `dns_servers` will do nothing (default behavior)
- Update Documentation

https://github.com/saltstack/salt/issues/49730

### Tests written?
No

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/modules/win_ip.py
+++ b/salt/modules/win_ip.py
@@ -319,6 +319,19 @@ def set_static_dns(iface, *addrs):
     '''
     Set static DNS configuration on a Windows NIC
 
+    Args:
+
+        iface (str): The name of the interface to set
+
+        addrs (*):
+            One or more DNS servers to be added
+
+            .. note::
+                If no DNS servers are passed the list will be cleared.
+
+    Returns:
+        dict: A dictionary containing the new DNS settings
+
     CLI Example:
 
     .. code-block:: bash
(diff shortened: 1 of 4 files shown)


salt/modules/win_ip.py
```diff
@@ -319,13 +319,35 @@ def set_static_dns(iface, *addrs):
...
     Set static DNS configuration on a Windows NIC
 
+    Args:
+
+        iface (str): The name of the interface to set
+
+        addrs (*):
+            One or more DNS servers to be added
+
+            .. note::
```
cachedout: This does not seem like the right behavior to me. These sorts of side-effects are implicit and, in my opinion, should be explicit. IMHO, nothing passed should mean no change is applied and a list should only be cleared if, and only if, an empty list is explicitly passed in.

twangboy:
@cachedout Only clears the list if you specifically pass []
@dwoz Added some tests

dwoz (approved):

twangboy:
I don't think all the tests actually ran... 4 seems like a small number

rallytime:
@twangboy Yes, 4 is not the right number. The rest of the tests are waiting in the jenkins queue for an executor slot. Once they start, this will update with the remaining jobs.

rallytime:
@twangboy Your changes have caused a test to fail: https://jenkinsci.saltstack.com/job/pr-kitchen-ubuntu1604-py2/job/PR-49763/4/

Issue #49520, opened 2018-09-05
Multi-master DNS Issue - minion connect fail


doesitblend:
### Description of Issue/Question
When using multi-master and all keys have been accepted, a minion does not connect to the other masters when the first master in the list is down. Another symptom of this problem occurs simply when any master in the list is down

In a docker container test environment with 4 masters and 2 minions, each minion connecting to all masters via dns name, stop the first master. Then restart the two minions. Both of the minions will likely have the following error in the logs and be unresponsive from ay other master that is still up:

```
minion1_1  | [ERROR   ] DNS lookup or connection check of 'master1' failed.
minion1_1  | [ERROR   ] Master hostname: 'master1' not found or not responsive. Retrying in 30 seconds
minion1_1  | [ERROR   ] DNS lookup or connection check of 'master1' failed.
minion1_1  | [ERROR   ] Master hostname: 'master1' not found or not responsive. Retrying in 30 seconds
```

### Setup
My minion configuration for both minions is:

```
[root@minion1 /]# cat /etc/salt/minion
master: 
  - master1
  - master2
  - master3
  - master4

log_level_logfile: trace
master_alive_interval: 20
```

My master configurarion for all masters is:
```
auto_accept: True
log_level_logfile: trace
```

### Versions Report
```
Salt Version:
           Salt: 2018.3.2
 
Dependency Versions:
           cffi: Not Installed
       cherrypy: Not Installed
       dateutil: Not Installed
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.7.2
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.5.6
   mysql-python: Not Installed
      pycparser: Not Installed
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: Not Installed
         Python: 2.7.5 (default, Jul 13 2018, 13:06:57)
   python-gnupg: Not Installed
         PyYAML: 3.11
          PyZMQ: 15.3.0
           RAET: Not Installed
          smmap: Not Installed
        timelib: Not Installed
        Tornado: 4.2.1
            ZMQ: 4.1.4
 
System Versions:
           dist: centos 7.5.1804 Core
         locale: ANSI_X3.4-1968
        machine: x86_64
        release: 4.9.93-linuxkit-aufs
[...]

PR #49764, opened 2018-09-25
[2018.3] Adding retry_dns_count to minion

garethgreenaway:
### What does this PR do?
Updating the resolve_dns function in minion.py to include a new minion configuration option which will control how many attempts will be made when the master hostname is unable to be resolved before giving up.

#49520 

### Previous Behavior
When a including a master hostname that doesn't resolve, particular in a multi-master setup, the resolve_dns function in the minion will continue indefinitely and never move onto other minions.

### New Behavior
This change updates that function to look for a new minion configuration option, `retry_dns_count`, which will allow the minion to either move on or exit following a specified number of attempts.

### Tests written?
Yes

### Commits signed with GPG?
Yes

### DIFF
--- a/doc/ref/configuration/minion.rst
+++ b/doc/ref/configuration/minion.rst
@@ -307,10 +307,12 @@ Set to zero if the minion should shutdown and not retry.
 
     retry_dns: 30
 
-.. conf_minion:: retry_dns
+.. conf_minion:: retry_dns_count
 
 ``retry_dns_count``
--------------
+-------------------
+
+.. versionadded:: 2018.3.4
 
 Default: ``None``
 
--- a/doc/ref/configuration/minion.rst
+++ b/doc/ref/configuration/minion.rst
@@ -307,6 +307,21 @@ Set to zero if the minion should shutdown and not retry.
 
     retry_dns: 30
 
+.. conf_minion:: retry_dns
+
+``retry_dns_count``
+-------------
+
+Default: ``None``
+
+Set the number of attempts to perform when resolving
+the master hostname if name resolution fails.
+By default the minion will retry indefinitely.
+
+.. code-block:: yaml
+
+    retry_dns_count: 3
+
 .. conf_minion:: master_port
 
 ``master_port``
--- a/salt/config/__init__.py
+++ b/salt/config/__init__.py
@@ -518,6 +518,7 @@ VALID_OPTS = {
     # The number of seconds to sleep between retrying an attempt to resolve the hostname of a
     # salt master
     'retry_dns': float,
+    'retry_dns_count': int,
 
     # In the case when the resolve of the salt master hostname fails, fall back to localhost
     'resolve_dns_fallback': bool,
@@ -1397,6 +1398,7 @@ DEFAULT_MINION_OPTS = {
     'update_url': False,
     'update_restart_services': [],
     'retry_dns': 30,
+    'retry_dns_count': None,
     'resolve_dns_fallback': True,
     'recon_max': 10000,
     'recon_default': 1000,
--- a/salt/config/__init__.py
+++ b/salt/config/__init__.py
@@ -518,7 +518,7 @@ VALID_OPTS = {
     # The number of seconds to sleep between retrying an attempt to resolve the hostname of a
     # salt master
     'retry_dns': float,
-    'retry_dns_count': int,
+    'retry_dns_count': (type(None), int),
 
     # In the case when the resolve of the salt master hostname fails, fall back to localhost
     'resolve_dns_fallback': bool,
--- a/salt/minion.py
+++ b/salt/minion.py
@@ -682,7 +682,6 @@ class MinionBase(object):
                     yield tornado.gen.sleep(opts['acceptance_wait_time'])
                 attempts += 1
                 if tries > 0:
-
                     log.debug(
                         'Connecting to master. Attempt %s of %s',
                         attempts, tries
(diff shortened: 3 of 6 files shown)


gtmanfred (changes requested):
Just one change

salt/config/__init__.py
```diff
@@ -1397,6 +1398,7 @@ def _gather_buffer_space():
     'update_url': False,
     'update_restart_services': [],
     'retry_dns': 30,
```
gtmanfred: Seems like the default should be 0 since the type is int

gtmanfred (approved):

salt/config/__init__.py
garethgreenaway: Updated the default to include None, if the value is 0 then it will always count down and I wanted the default to remain the same, trying indefinitely.

isbm (changes requested):
I would rather set this default to a long enough, reasonable amount of times and bail out. Say, put 50 if you want. If then anyone wants to keep it until forever, should do it explicitly.

Also maybe `retry_dns_times` would be a better name.

salt/minion.py
```diff
@@ -984,8 +991,14 @@ def _spawn_minions(self):
...
-            self.minions.append(minion)
-            self.io_loop.spawn_callback(self._connect_minion, minion)
+
+            self._connect_minion(minion)
+
+        if not self.minions:
+            err = ('Minion unable to successfully connect to '
+                   'a Salt Master.  Exiting.')
+            log.error(err)
+            raise SaltSystemExit(code=42, msg=err)
```
isbm: We have constants for this. Please do not use hard-coded error codes.

salt/minion.py
```diff
@@ -157,8 +158,13 @@ def resolve_dns(opts, fallback=True):
                 True,
                 opts['ipv6'])
         except SaltClientError:
+            retry_dns_count = opts.get('retry_dns_count', None)
             if opts['retry_dns']:
                 while True:
+                    if retry_dns_count is not None:
+                        if retry_dns_count == 0:
+                            raise SaltMasterUnresolvableError
+                        retry_dns_count -= 1
```
isbm: Well, that's the thing. I'd set it default to something big enough. Any `None/0` or negative is "until forever".

isbm (changes requested):

salt/minion.py
```diff
@@ -676,6 +682,7 @@ def eval_master(self,
                     yield tornado.gen.sleep(opts['acceptance_wait_time'])
                 attempts += 1
                 if tries > 0:
+
```
DmitryKuzmenko: Unrelated/accidental edit?

salt/minion.py
```diff
@@ -984,8 +991,14 @@ def _spawn_minions(self):
                                                 loaded_base_name='salt.loader.{0}'.format(s_opts['master']),
                                                 jid_queue=self.jid_queue,
                                                )
-            self.minions.append(minion)
-            self.io_loop.spawn_callback(self._connect_minion, minion)
+
+            self._connect_minion(minion)
+
+        if not self.minions:
```
DmitryKuzmenko: `_connect_minion` is async operation. If it will not done fast enough the list could be empty here. It's better to run this check with `io_loop.call_later`.

DmitryKuzmenko (changes requested):

salt/minion.py
garethgreenaway: Yup.  I think there was a previous attempt here that resulted in that additional newline.

salt/minion.py
garethgreenaway: This particular return code is used in several other places in the code base, as is, not referenced by a constant, including in minion.py.

doc/ref/configuration/minion.rst
```diff
@@ -307,6 +307,21 @@ Set to zero if the minion should shutdown and not retry.
...
+
+Default: ``None``
+
+Set the number of attempts to perform when resolving
+the master hostname if name resolution fails.
+By default the minion will retry indefinitely.
+
+.. code-block:: yaml
+
+    retry_dns_count: 3
```
cachedout: Personally, I think this value should be 1.

DmitryKuzmenko (approved):

doc/ref/configuration/minion.rst
```diff
@@ -307,6 +307,21 @@ Set to zero if the minion should shutdown and not retry.
 
     retry_dns: 30
 
+.. conf_minion:: retry_dns
+
+``retry_dns_count``
+-------------
```
rallytime: I know this is picky, but the number of dashes needs to match the line above it or the doc build will have a warning. (We need to figure out a way to get that to bubble up in the PR tests still.)

Can you add a `versionadded` tag in here as well?

doc/ref/configuration/minion.rst
```diff
@@ -307,6 +307,21 @@ Set to zero if the minion should shutdown and not retry.
 
     retry_dns: 30
 
+.. conf_minion:: retry_dns
```
rallytime: This needs to match the value below: `retry_dns_count`.

salt/minion.py
rallytime: This looks correct to me. It should default "until forever" because that is the current behavior which we don't want to break for people.

rallytime (changes requested):
I have some small doc requests.

I addition to the inline comments, can you also add this setting to the sample [minion conf](https://github.com/saltstack/salt/blob/develop/conf/minion)?

Otherwise, 👍 from me.

rallytime (approved):

cachedout (approved):

PR #49771, opened 2018-09-25
modules/nilrt_ip.py: Fix set_dhcp_linklocal_only

alexvasiu:
* modules/nilrt_ip.py: Fix set_dhcp_linklocal_only

    It's safer to make dhcp_settings also through the config file and
    then restart the connman daemon.

    Also, after an interface was disabled, when it's changed to static,
    it should be removed from connman blacklist (same for dhcp, enable
    and disable function).

### DIFF
--- a/salt/modules/nilrt_ip.py
+++ b/salt/modules/nilrt_ip.py
@@ -567,12 +567,13 @@ def _remove_interface_section(interface):
     return True
 
 
-def _change_connman_backlist(interface, add=True):
+def _change_connman_blacklist(interface, add=True, enable_dhcp=True):
     '''
     Remove or add an interface to connman blacklist
 
     :param interface: interface label
     :param add: True to add interface to blacklist, False otherwise. Default is True.
+    :param enable_dhcp: True to enable dhcp for an interface after removing from blacklist
     '''
     parser = configparser.ConfigParser()
     parser.optionxform = str
@@ -600,7 +601,7 @@ def _change_connman_backlist(interface, add=True):
         parser.write(config_file)
     if add:
         return _remove_interface_section(interface)
-    return _enable_dhcp(interface)
+    return _enable_dhcp(interface) if enable_dhcp else True
 
 
 def _enable_dhcp(interface):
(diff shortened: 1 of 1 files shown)


cachedout:
I would like to see this documented in our changelog before we get this merged, please.

salt/modules/nilrt_ip.py
```diff
@@ -600,7 +601,7 @@ def _change_connman_backlist(interface, add=True):
         parser.write(config_file)
     if add:
```
rares-pop: this looks at least funny... if add remove()

salt/modules/nilrt_ip.py
```diff
@@ -625,7 +626,7 @@ def _enable_dhcp(interface):
     parser.add_section('service_{0}'.format(hwaddr_section_number))
     parser.set('service_{0}'.format(hwaddr_section_number), 'MAC', hwaddr)
     parser.set('service_{0}'.format(hwaddr_section_number), 'Name', 'ethernet_cable_{0}'.format(hwaddr_section_number))
-    parser.set('service_{0}'.format(hwaddr_section_number), 'IPv4.method', 'dhcp')
+    parser.set('service_{0}'.format(hwaddr_section_number), 'IPv4.Method', 'dhcp')
```
rares-pop: changing case from method to Method does affect anything? is it intentional?

salt/modules/nilrt_ip.py
```diff
@@ -645,23 +646,10 @@ def _change_state(interface, new_state):
...
-        try:
-            state = service.connect() if new_state == 'up' else service.disconnect()
-            _change_connman_backlist(interface, add=new_state == 'down')
-            return state is None
-        except Exception:
-            raise salt.exceptions.CommandExecutionError('Couldn\'t {0} service: {1}\n'
-                                                        .format('enable' if new_state == 'up' else 'disable', service))
-    return True
+    if interface in map(lambda x: x.name, pyiface.getIfaces()):
+        ret = _change_connman_blacklist(interface, add=new_state == 'down')
```
rares-pop: add=new_state == 'down' looks confusing. You could either get that an a different line, or add parentheses

rares-pop: In addition, there should be some tests to accompany this change.

salt/modules/nilrt_ip.py
alexvasiu: Yeah, I could refactor this

salt/modules/nilrt_ip.py
alexvasiu: Yes, it's intentional, because Connman recognizes `IPv4.Method` (it's case sensitive)

PR #49774, opened 2018-09-25
Move all pipelines to be fully scripted

KaiSforza:
This allows us to properly clean out the workspace before a job run.

### What does this PR do?

Changes the Jenkinsfiles to use scripted pipelines, not declarative pipelines. Should fix the failures we were getting with old workspaces being used. Not sure exactly how it's going to work if we are actually using old workspaces...but this should make new builds work.

Not cleaning up our workspace before the build. We now clean the workspace, clone the git repo, and then do the build.

### Commits signed with GPG?

Yes

@gtmanfred @dubb-b @rallytime

### DIFF
--- a/.ci/kitchen-centos7-py2
+++ b/.ci/kitchen-centos7-py2
@@ -1,73 +1,75 @@
-pipeline {
-    agent { label 'kitchen-slave' }
-    options {
-        timestamps()
-        ansiColor('xterm')
-    }
-    environment {
-        SALT_KITCHEN_PLATFORMS = "/var/jenkins/workspace/platforms.yml"
-        SALT_KITCHEN_DRIVER = "/var/jenkins/workspace/driver.yml"
-        PATH = "/usr/local/rbenv/shims/:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/root/bin:/root/bin"
-        RBENV_VERSION = "2.4.2"
-        TEST_SUITE = "py2"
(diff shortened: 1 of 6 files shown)


KaiSforza:
This jenkins job was run with the new scripted pipeline, only ran a single test for speed, though.

https://jenkinsci.saltstack.com/job/pr-kitchen-centos7-py2/job/PR-48809/43/console

gtmanfred (approved):

damon-atkins:
The `agent` statement cleans up when it checks out the scm.  For example if each stage had a `agent` line it would clean up between stages.

You can also set this in the stage
`options { skipDefaultCheckout() }`
and then do your own
```
cleanWS 
checkout scm
```
There is no need to convert to "fully scripted"

KaiSforza:
@damon-atkins Looking back at some of the tests, it was only cleaning up on successful builds if it got to the end, the `agent` statement wasn't doing anything to clean up. 

There are also other reasons why I want to convert to a scripted style, as well.

damon-atkins:
Maybe in post the junit was failing and it never reached the cleanWS

By the way the BlueOcean interface works well for users, makes the logs and unit tests easier to read as well.

KaiSforza:
@damon-atkins That's what I think was happening. Again, this isn't just because of the failed cleanups, but other stuff we want to do that *can't* be done in the declarative pipelines.

 Us devops people have been wanting to switch over to it, I personally want it, but there are reasons we haven't switched.

Issue #13340, opened 2014-06-09
Add an enforce option to the ssh_auth state


ryan-lane:
ssh_auth should be able to completely ensure the state of an authorized_keys file as well as adding and removing individual keys. It would be nice to have an enforce option (that defaults to false) that would fully manage an authorized_keys file.

I know it's possible to use file.managed for this, but then it's necessary to know the user's home directory, which may not be default.

baconz:
bump

hkbakke:
Is there any progress on this. In my opinion the lack of something like this makes ssh_auth almost useless for anything but the simplest use case.

jleroy:
Just a *bump* to keep the issue open.

basepi:
@gtmanfred A ping to keep this on employee radars.

mchugh19:
With the merge of #50141, I think this can be closed.

PR #49776, opened 2018-09-25
Add mod_aggregate to ssh_auth to manage authorized_keys

mchugh19:
Not sure if this is an appropriate way to handle this scenario. Comments requested!

### What does this PR do?
Adds a mod_aggregate function to the ssh_auth state. This then builds up a list of the ssh keys from this and subsequent ssh_auth.present declarations, for each user. The present function then uses this list to determine what ssh_keys are being managed by salt, and removes the rest.

#13340

### Previous Behavior
ssh_auth.present was only able to add keys. Any old keys not managed by salt need to be removed by specifying them in a ssh_auth.absent state.

### New Behavior
By specifying `aggregate: True` for the ssh_auth.present state, ssh keys not added by salt will be removed.

####  Example
When adding the Asomekeyh ssh key, another Aoldkey= was found for this user. Since it was not provisioned by salt, it was removed.

Here is a state to ensure that Asomekeyh is present. In the past Aoldkey= had been added to the host, but has been forgotten and is not managed by salt.
```
key1:
  ssh_auth.present:
    - user: testuser
    - enc: ssh-rsa
    - comment: testuser 1
    - aggregate: True
    - names:
      - Asomekeyh
      - MotherkeyB

#key2:
#  ssh_auth.present:
#    - user: testuser
#    - enc: ssh-rsa
#    - comment: 2
#    - names:
#      - Aoldkey=
```

By running this state, ssh_auth.present ensured the Asomekeyh was present, but also removed any keys found for the testuser user which were not managed by salt.
```
          ID: key1
    Function: ssh_auth.present
        Name: Asomekeyh
      Result: True
     Comment: The authorized host key Asomekeyh is already present for user testuser
     Started: 16:08:57.859383
    Duration: 9.498 ms
     Changes:
              ----------
              Aoldkey=:
                  Key removed
```
### Tests written?
No

### Commits signed with GPG?
No

### DIFF
--- a/salt/states/ssh_auth.py
+++ b/salt/states/ssh_auth.py
@@ -259,6 +259,11 @@ def present(
 
             The default value of the ``fingerprint_hash_type`` will change to
             ``sha256`` in Salt 2017.7.0.
+
+    aggregate
+        Removes any existing ssh keys except for those managed by subsequent ``ssh_auth.present``
+
+        .. versionadded:: Fluorine
     '''
     ret = {'name': name,
            'changes': {},
@@ -297,6 +302,13 @@ def present(
                 config,
                 fingerprint_hash_type)
         return ret
+
+    if 'ssh_keys' in kwargs:
+        existing_keys = __salt__['ssh.auth_keys'](user=user).keys()
+        remove_keys = set(existing_keys).difference(kwargs['ssh_keys'])
+        for ssh_key in remove_keys:
+            remove_comment = absent(ssh_key, user)['comment']
+            ret['changes'][ssh_key] = remove_comment
 
     # Get only the path to the file without env referrences to check if exists
     if source != '':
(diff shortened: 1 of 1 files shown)


salt/states/ssh_auth.py
```diff
@@ -259,6 +259,11 @@ def present(
 
             The default value of the ``fingerprint_hash_type`` will change to
             ``sha256`` in Salt 2017.7.0.
+
+    aggregate
+        Removes any existing ssh keys except for those managed by subsequent ``ssh_auth.present``
+
+        .. versionadded:: Fluorine
```
rallytime: Can you change this to `Neon`?

rallytime (commented):
I have one small comment here.

salt/states/ssh_auth.py
```diff
@@ -286,6 +291,16 @@ def present(
             if len(comps) == 3:
                 comment = comps[2]
 
+    if 'ssh_keys' in kwargs:
```
gtmanfred: Can we expose this so that it can be used directly as well, instead of only through mod_aggregate?

gtmanfred (changes requested):

salt/states/ssh_auth.py
mchugh19: Whoops! Yep, will do.

salt/states/ssh_auth.py
gtmanfred: Ok, removing the ssh keys is not the purpose of mod_aggregate.

The purpose of mod_aggregate is to combine all of the calls into just one call.

See how it is used in pkg.installed where it makes only one call to the system package manager, instead of once per package state.

I am not sure that we should make this change if it is going to start removing keys that are not managed by the aggregate.  @saltstack/team-core opinions?

mchugh19: Agreed. Hence the initial "I don't know if this is kosher" comment. 

In order to perform authorized_keys management where keys not managed by salt are removed, we would need to know about all the keys that salt is managing. Today that means a series of `ssh_auth.present` stanzas. Because they are indepent, it forces the use of mod_aggregate. This overloads the use of mod_aggregate. In the pkg state it aggregates calls while in this use it aggregates and allows the multiple calls to more completely manage the authorized_keys file. Possibly a hack, or possibly awesome?!

If there are cleaner ways of accomplishing this, I'm for that as well. But a definitely a review is needed before any potential merge.

gtmanfred: yeah, i just want to have more input from the rest of the core team to see if we are ok with this.

terminalmage (commented):
This is _definitely_ not how we want to use aggregate.

mchugh19:
@gtmanfred  @terminalmage @cachedout is there a recommendation for how to support #13340 functionality?

`ssh_auth.present` allows for passing various arguments and options, which is why this PR only attempted to handle removing unlisted keys and continued to use independent ssh_auth.present declarations to add. Perhaps for the simple use-case an "ssh-keys" key could be created which adds those keys, and removes all others? This would require all keys listed to have the same options, but perhaps that is an acceptable trade off?

mchugh19:
Closing this out in favor of a different approach

PR #49779, opened 2018-09-25
Remove deprecated options from the dockermod execution module

### LINKED ISSUE #49410: Handle deprecation notices in docker execution module for Neon release
- [x] salt/modules/dockermod.py

rallytime:
These options were deprecated and marked for removal in the Neon release.

Fixes #49410

### DIFF
--- a/salt/modules/dockermod.py
+++ b/salt/modules/dockermod.py
@@ -1915,7 +1915,7 @@ def resolve_image_id(name):
     return False
 
 
-def resolve_tag(name, tags=None, **kwargs):
+def resolve_tag(name, **kwargs):
     '''
     .. versionadded:: 2017.7.2
     .. versionchanged:: 2018.3.0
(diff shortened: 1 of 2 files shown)


salt/modules/dockermod.py
```diff
@@ -1945,7 +1945,6 @@ def resolve_tag(name, tags=None, **kwargs):
 
     tags
         .. deprecated:: 2018.3.0
-            Ignored if passed, will be removed in the Neon release.
```
terminalmage: We can probably also remove the rest of the documentation for this argument, what do you think?

terminalmage (approved):

salt/modules/dockermod.py
rallytime: I have done that in the past, but I was thinking it might be nice for people to know what happened to the option if they go looking. Thoughts?

PR #49781, opened 2018-09-25
Update pillar/grain master cache logic

chewvader:
### What does this PR do?

When executing _get_minion_grains and all nodes are loaded via cache, len(missed_minions) == 0.
This causes  ```lret = self._get_live_minion_grains(missed_minions)```  to execute with an empty list which results in the print statement ```No minions matched the target. No command was sent, no jid was assigned.``` but everything else executes fine...
This adds additional validation before executing the fallback statements.

This also ensures that directly queried minion grains/pillars "lret" overwrites any cached data "cret" before return.

No issue number.

### Tests written?

No

### Commits signed with GPG?

Yes


salt/utils/master.py
```diff
@@ -189,16 +189,16 @@ def _get_minion_grains(self, *minion_ids, **kwargs):
             cret = dict([(minion_id, mcache) for (minion_id, mcache) in six.iteritems(cached_grains) if mcache])
             missed_minions = [minion_id for minion_id in minion_ids if minion_id not in cret]
             log.debug('Missed cached minion grains for: %s', missed_minions)
-            if self.grains_fallback:
+            if self.grains_fallback and len(missed_minions) > 0:
```
gtmanfred: just do `if self.grains_fallback and missed_minions`

you do not need to check the length, if it is an empty list, it will evaluate as false, and if there are times in it it will evaluate as true

salt/utils/master.py
```diff
@@ -189,16 +189,16 @@ def _get_minion_grains(self, *minion_ids, **kwargs):
...
                 lret = self._get_live_minion_grains(missed_minions)
-            ret = dict(list(six.iteritems(dict([(minion_id, {}) for minion_id in minion_ids]))) + list(lret.items()) + list(cret.items()))
+            ret = dict(list(six.iteritems(dict([(minion_id, {}) for minion_id in minion_ids]))) + list(cret.items()) + list(lret.items()))
```
gtmanfred: can you convert this to using dict interpretation instead of dict()?

so instead of

`dict([x, y] for x, y in z.items())`

you would use 

`{x: y for x, y in z.items()}`

we no longer support python 2.6 which is what the dict() wrapper was used for.

gtmanfred (changes requested):
This looks good, just a couple pythonic/better way to do things requests.

salt/utils/master.py
gtmanfred: this needs to be applied to the other changes as well

salt/utils/master.py
gtmanfred: same for this one.

chewvader:
Updated requested changes.

salt/utils/master.py
```diff
@@ -189,16 +189,16 @@ def _get_minion_grains(self, *minion_ids, **kwargs):
             cret = dict([(minion_id, mcache) for (minion_id, mcache) in six.iteritems(cached_grains) if mcache])
             missed_minions = [minion_id for minion_id in minion_ids if minion_id not in cret]
             log.debug('Missed cached minion grains for: %s', missed_minions)
-            if self.grains_fallback:
+            if self.grains_fallback and missed_minions:
                 lret = self._get_live_minion_grains(missed_minions)
-            ret = dict(list(six.iteritems(dict([(minion_id, {}) for minion_id in minion_ids]))) + list(lret.items()) + list(cret.items()))
+            ret = dict([(minion_id, {}) for minion_id in minion_ids] + list(cret.items()) + list(lret.items()))
```
gtmanfred: no, this should be like.

```
ret = {key: value for key, value in [(minion_id, {}) for minion_id in minion_ids] + list(cret.items()) + list(lret.items())}
```

gtmanfred (approved):
Awesome, thanks! and welcome to the salt contributors community.

salt/utils/master.py
```diff
@@ -189,16 +189,16 @@ def _get_minion_grains(self, *minion_ids, **kwargs):
...
+            ret = {key: value for key, value in [(minion_id, {}) for minion_id in minion_ids] + list(cret.items()) + list(lret.items())}
         else:
             lret = self._get_live_minion_grains(minion_ids)
             missed_minions = [minion_id for minion_id in minion_ids if minion_id not in lret]
             log.debug('Missed live minion grains for: %s', missed_minions)
-            if self.grains_fallback:
+            if self.grains_fallback and missed_minions:
                 cret = dict([(minion_id, mcache) for (minion_id, mcache) in six.iteritems(cached_grains) if mcache])
-            ret = dict(list(six.iteritems(dict([(minion_id, {}) for minion_id in minion_ids]))) + list(lret.items()) + list(cret.items()))
+            ret = {key: value for key, value in [(minion_id, {}) for minion_id in minion_ids] + list(cret.items()) + list(lret.items())}
```
terminalmage: I know you didn't write the orignial code, but can you change the calls to `.items()` to use `six.iteritems()` (e.g. `list(six.iteritems(cret))`)? This prevents us from allocating two lists in Python 2.

The same goes for the dict comprehension on line 194.

salt/utils/master.py
```diff
@@ -214,16 +214,16 @@ def _get_minion_pillar(self, *minion_ids, **kwargs):
...
         else:
             lret = dict([(minion_id, self._get_live_minion_pillar(minion_id, grains.get(minion_id, {}))) for minion_id in minion_ids])
             missed_minions = [minion_id for minion_id in minion_ids if minion_id not in lret]
             log.debug('Missed live minion pillars for: %s', missed_minions)
-            if self.pillar_fallback:
+            if self.pillar_fallback and missed_minions:
                 cret = dict([(minion_id, mcache) for (minion_id, mcache) in six.iteritems(cached_pillar) if mcache])
-            ret = dict(list(six.iteritems(dict([(minion_id, {}) for minion_id in minion_ids]))) + list(lret.items()) + list(cret.items()))
+            ret = {key: value for key, value in [(minion_id, {}) for minion_id in minion_ids] + list(cret.items()) + list(lret.items())}
```
terminalmage: Same as the previous comment, can you please switch to using `list(six.iteritems(dictname))` here and on line 219?

terminalmage (changes requested):

isbm (commented):
To me it looks four to eight time the same code duplicated all over the place. I'd move this to a separate function, even it is a small one, even it is an internal function.

terminalmage (approved):

Issue #49766, opened 2018-09-25
Context key in Sentry Logger config causes hard exception.


weswhet:
### Description of Issue/Question
When specifying the configuration for the Sentry logger, if a `context` key is added to the config you end up with the following sadness. 

```sudo salt-call schedule.run_job tarball_update
[ERROR   ] u'grains.get'
Traceback (most recent call last):
  File "/opt/salt/lib/python2.7/site-packages/salt/utils/parsers.py", line 222, in parse_args
    mixin_after_parsed_func(self)
  File "/opt/salt/lib/python2.7/site-packages/salt/utils/parsers.py", line 852, in __setup_extended_logging
    log.setup_extended_logging(self.config)
  File "/opt/salt/lib/python2.7/site-packages/salt/log/setup.py", line 747, in setup_extended_logging
    handlers = get_handlers_func()
  File "/opt/salt/lib/python2.7/site-packages/salt/log/handlers/sentry_mod.py", line 194, in setup_handlers
    tag_value = __salt__['grains.get'](tag)
KeyError: u'grains.get'
Usage: salt-call [options] <function> [arguments]
```
### Setup
```
## MANAGED BY SALT DO NOT MODIFY ##
sentry_handler:
  dsn: https://foobar@sentry.io/somanynumbers
  context:
    - os
```

### Steps to Reproduce Issue
Add the above config to `/etc/salt/minion.d/sentry.conf`

run any `salt-call` action. Arrive at sadness

### Versions Report
```
sudo salt-call --versions-report

Salt Version:
           Salt: 2018.3.2

Dependency Versions:
           cffi: 1.11.2
       cherrypy: 13.0.0
       dateutil: 2.6.1
      docker-py: Not Installed
          gitdb: 2.0.3
      gitpython: 2.1.7
          ioflo: Not Installed
         Jinja2: 2.10
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: 1.0.7
   msgpack-pure: Not Installed
 msgpack-python: 0.4.8
   mysql-python: Not Installed
      pycparser: 2.18
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: Not Installed
         Python: 2.7.14 (default, Jun 18 2018, 15:52:11)
   python-gnupg: 0.4.1
         PyYAML: 3.12
          PyZMQ: 17.0.0
           RAET: Not Installed
          smmap: 2.0.3
        timelib: 0.2.4
[...]

PR #49790, opened 2018-09-26
fixing an issue where sentry logger would fail to get tags from grain…

weswhet:
### What does this PR do?
Fixes https://github.com/saltstack/salt/issues/49766

https://github.com/saltstack/salt/issues/49766

### Previous Behavior
Would fail when specifying any tags for sentry.

### New Behavior
Sends grains as tags, as expected. 

### Tests written?
No

### Commits signed with GPG?
Yes


weswhet:
Sorry for the mix up on PRs, git is hard 😞

rallytime (approved):
Not a problem. :)

Thanks for moving this to the 2018.3 branch!

PR #49794, opened 2018-09-26
Fix possible issue with test_groupadd

twangboy:
### What does this PR do?
In Windows, the `group.members` function returns a dicts where the `result` key returns true if successful. On Linux it just returns True or False. The way the test was written the check would always be True even on failure because an empty Dict results in True if you're just checking with `assertTrue`.

This checks the value in `result` on Windows

### Tests written?
Yes

### Commits signed with GPG?
Yes

### DIFF
--- a/tests/integration/modules/test_groupadd.py
+++ b/tests/integration/modules/test_groupadd.py
@@ -160,7 +160,8 @@ class GroupModuleTest(ModuleCase):
         self.run_function('user.add', [self._user])
         self.run_function('user.add', [self._user1])
         m = '{0},{1}'.format(self._user, self._user1)
-        self.assertTrue(self.run_function('group.members', [self._group, m]))
+        self.assertTrue(
+            self.run_function('group.members', [self._group, m])['result'])
         group_info = self.run_function('group.info', [self._group])
         self.assertIn(self._user, str(group_info['members']))
         self.assertIn(self._user1, str(group_info['members']))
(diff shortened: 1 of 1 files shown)


dwoz (approved):

rallytime (approved):

cachedout:
Why did these two platforms diverge in the first place? Would it make sense to just bring them in line with each other?

damon-atkins:
@cachedout agree they should be the same.   The state expects a true or false from all platforms, not a data structure.

Normal state outputs `comments`, `changes`, general modules do not.

damon-atkins:
`if __salt__['group.add'](name, gid=gid, system=system):`

However the above seems wrong.   The group.add or group.* on all platforms should raise a `CommandExecutionError('some error message') ` and the group state should trap it.

What should the standard be?  i.e. more like pkg state?  Windows `group.add` is sort of following.

twangboy:
Now returns bool instead of dict

rallytime:
@twangboy and @dwoz Is that change in structure something we want to do in a dot release? I think that might be better for a feature release. Thoughts?

cachedout:
I agree with @rallytime . I think we should consider a change in the structure in a major release but not change it here.

twangboy:
I reverted back to just the original test fix I did. I created another PR against the Fluorine branch with the fixes I had before. https://github.com/saltstack/salt/pull/50086

PR #49803, opened 2018-09-27
parted: fix _validate_partition_boundary

aplanas:
The current partition boundary validator checks that one of the
valid units is a suffix of the boundary, and if so consider it
a valid one. This validation have two problems:

* We can miss the numerical part of the boundary

  e.g: 's' or 'MB' will be valid units

* The unit provided can be invalid

  e.g: '10as', '1okB'

The current validator make sure that there is a valid number
and a valid unit in the boundary.


aplanas:
Is there anything here that I can do to understand the change?

rallytime:
Just waiting on a review from @garethgreenaway on this one, as well as some tests that didn't run completely. I've restarted them.

garethgreenaway (approved):

PR #49809, opened 2018-09-27
Add additional asserts to help investigate test_win_pkg failure

Ch3LL:
### What does this PR do?
Currently the `test_win_pkg` module integration tests is failing sometimes on windows. I cannot replicate this failure so I'm adding additional asserts and running the `refresh_db` one more time if it does initially fail.

### DIFF
--- a/salt/modules/win_pkg.py
+++ b/salt/modules/win_pkg.py
@@ -693,12 +693,13 @@ def refresh_db(**kwargs):
 
     # Cache repo-ng locally
     log.info('Fetching *.sls files from {0}'.format(repo_details.winrepo_source_dir))
-    __salt__['cp.cache_dir'](
+    ret = __salt__['cp.cache_dir'](
         path=repo_details.winrepo_source_dir,
         saltenv=saltenv,
         include_pat='*.sls',
         exclude_pat=r'E@\/\..*?\/'  # Exclude all hidden directories (.git)
     )
+    log.debug("refresh_db - Return from cache_dir %s", repr(ret))
 
     return genrepo(saltenv=saltenv, verbose=verbose, failhard=failhard)
 
--- a/salt/modules/win_pkg.py
+++ b/salt/modules/win_pkg.py
@@ -693,14 +693,12 @@ def refresh_db(**kwargs):
 
     # Cache repo-ng locally
     log.info('Fetching *.sls files from {0}'.format(repo_details.winrepo_source_dir))
-    ret = __salt__['cp.cache_dir'](
+    __salt__['cp.cache_dir'](
         path=repo_details.winrepo_source_dir,
         saltenv=saltenv,
         include_pat='*.sls',
         exclude_pat=r'E@\/\..*?\/'  # Exclude all hidden directories (.git)
     )
-    log.debug("refresh_db - Return from cache_dir %s", repr(ret))
-
     return genrepo(saltenv=saltenv, verbose=verbose, failhard=failhard)
 
 
--- a/tests/integration/modules/test_win_pkg.py
+++ b/tests/integration/modules/test_win_pkg.py
@@ -76,6 +76,7 @@ class WinPKGTest(ModuleCase):
                     reboot: False
                 '''))
             fp_.flush()
+            os.fsync(fp_.fileno())
         # now check if curl is also in cache and repo query
         pkgs.append('curl')
         for pkg in pkgs:
--- a/tests/integration/modules/test_win_pkg.py
+++ b/tests/integration/modules/test_win_pkg.py
@@ -75,8 +75,6 @@ class WinPKGTest(ModuleCase):
                     locale: en_US
                     reboot: False
                 '''))
-            fp_.flush()
-            os.fsync(fp_.fileno())
         # now check if curl is also in cache and repo query
         pkgs.append('curl')
         for pkg in pkgs:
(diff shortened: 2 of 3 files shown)


dwoz (approved):

tests/integration/modules/test_win_pkg.py
```diff
@@ -33,8 +34,24 @@ def test_adding_removing_pkg_sls(self):
         Test add and removing a new pkg sls
         in the windows software repository
         '''
-        def _check_pkg(pkgs, exists=True):
-            self.run_function('pkg.refresh_db')
+        def _check_pkg(pkgs, exists=True, check_refresh=None):
```
cachedout: Should `check_refresh` be a boolean by default?

tests/integration/modules/test_win_pkg.py
```diff
@@ -33,8 +34,24 @@ def test_adding_removing_pkg_sls(self):
...
+            if check_refresh:
+                count = 2
+                while count != 0:
+                    try:
+                        self.assertEqual(0, refresh['failed'],
+                             msg='failed returned {0}. Expected return: 0'.format(refresh['failed']))
+                        self.assertEqual(check_refresh, refresh['total'],
+                             msg='total returned {0}. Expected return {1}'.format(check_refresh, refresh['total']))
+                        self.assertEqual(check_refresh, refresh['success'],
+                             msg='success returned {0}. Expected return {1}'.format(check_refresh, refresh['success']))
```
damon-atkins: msg='success returned {0}. Expected return {1}'.format(refresh['success']), check_refresh)

tests/integration/modules/test_win_pkg.py
```diff
@@ -33,8 +34,24 @@ def test_adding_removing_pkg_sls(self):
...
+                        count = 0
+                    except AssertionError as err:
+                        if count == 1:
+                            raise AssertionError(err)
+                        count = count -1
+                        refresh = self.run_function('pkg.refresh_db')
             repo_data = self.run_function('pkg.get_repo_data', timeout=300)
```
damon-atkins: timeout=30   it just opens an existing file and reads it into ram.  30 seconds is heaps

damon-atkins (changes requested):

tests/integration/modules/test_win_pkg.py
Ch3LL: I could change it to boolean but we are passing in `check_refresh=2`, when using this so i kept it at None. Is it better practice to keep it a boolean even though we are not passing True/False into the kwarg.

tests/integration/modules/test_win_pkg.py
Ch3LL: looks like that change was made intentionally to also try to stabilize this flaky test so I dont want to remove it quit yet.

dwoz:
@Ch3LL I'vd made some additions to this: https://github.com/Ch3LL/salt/pull/4

salt/modules/win_pkg.py
```diff
@@ -693,12 +693,13 @@ def refresh_db(**kwargs):
...
     # Cache repo-ng locally
     log.info('Fetching *.sls files from {0}'.format(repo_details.winrepo_source_dir))
-    __salt__['cp.cache_dir'](
+    ret = __salt__['cp.cache_dir'](
         path=repo_details.winrepo_source_dir,
         saltenv=saltenv,
         include_pat='*.sls',
         exclude_pat=r'E@\/\..*?\/'  # Exclude all hidden directories (.git)
     )
+    log.debug("refresh_db - Return from cache_dir %s", repr(ret))
```
damon-atkins: Not sure why this is needed, it already spills out heaps of logs in debug. i.e. every file which is pulled down and every file skipped.

tests/integration/modules/test_win_pkg.py
damon-atkins: You could also call pkg.refresh_db with verbose=True and failhard=False
The data structure return will be this even if it fails (with data not like the example below). 
```
{
    "local": {
        "total": 0,
        "success": 0,
        "success_list": {},
        "failed_list": {},
        "failed": 0
    }
}
```

damon-atkins:
Just looking at the windows test and there are hundreds of these lines `00:39:41 unix2dos: converting file`
Which is not how it works when its running for real, being installed normally.

dwoz:
@damon-atkins We added the unix2dos conversion a month or two ago. This was done because, by default, git will perform newline translations and git clones on Windows will have CRLF line endings. By running unix2dos the tests run with the same line endings as if someone runs the tests from a git clone on windows.

Due to the fact that git can be configured to perform newline translations (or not), using unix2dos is not ideal. Since you could perform a git clone without newline translations and then see some failures in the test suite because of that.

In addition, as you have pointed out, using unix2dos does not cover the case where you have a linux master with windows minion. It means the test suite primarily testing a Windows master with a Windows minion.

I think long term we want to enumerate all possible line ending cases and test all of them. We're just not there yet. A couple other things to think about is the fact that YAML performs line ending translation so that values parsed by YAML always have LF line endings even if the source file has CRLF line endings. JSON on the other hand *should* always have LF line endings source files.

Hopefully this explains why we are using unix2dos for now and the direction we are heading.

damon-atkins:
Thanks.   And assume pip installs would have the original line endings as well from the source.

damon-atkins (approved):

PR #49812, opened 2018-09-28
[minion][scheduler] The master_alive job never deleted

kbespalov:
### What does this PR do?

Fixes removing of the `master_alive` job from scheduler on `disconnected` event.

### Issue

The master_alive job never deleted because of name of the master does not specified .
Each time when one of the master is lost,  the minion creates new one, but never deletes previous job.


isbm (approved):

DmitryKuzmenko (approved):
Good catch @kbespalov !

gtmanfred (approved):

cachedout (approved):

PR #49826, opened 2018-09-28
Remove warnings/support for force option for Neon

### LINKED ISSUE #49421: Handle deprecated code in win_servermanager state for Neon
- [x] salt/states/win_servermanager.py

rallytime:
The `win_servermanager.installed` function no longer supports the "force" option. This change removes the warning for Neon and updates the release notes.

Fixes #49421

### DIFF
--- a/doc/topics/releases/neon.rst
+++ b/doc/topics/releases/neon.rst
@@ -212,3 +212,13 @@ Module Deprecations
     - Support for the ``ssh.recv_known_host`` function has been removed. Please use the
       :py:func:`ssh.recv_known_host_entries <salt.modules.ssh.recv_known_host_entries>`
       function instead.
+
+State Deprecations
+------------------
+
+- The :py:mod:`win_servermanager <salt.states.win_servermanager>` state has been
+  changed as follows:
+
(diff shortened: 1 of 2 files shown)


doc/topics/releases/neon.rst
```diff
@@ -212,3 +212,13 @@ Module Deprecations
...
       :py:func:`ssh.recv_known_host_entries <salt.modules.ssh.recv_known_host_entries>`
       function instead.
+
+State Deprecations
+------------------
+
+- The :py:mod:`win_servermanager <salt.states.win_servermanager>` state has been
+  changed as follows:
+
+    - Support for the ``force`` kwarg has been removed from the
```
cachedout: IMHO, we need to note the option that it has been replaced with, per the deprecation warning notice.

rallytime: @cachedout It hasn't been replaced by anything. This option hasn't actually been used for a while. All this PR does is remove the warning and make it officially removed.

rallytime: Oh actually you're right. I totally missed that notice. I'll fix it right now. :)

Issue #49204, opened 2018-08-20
file.directory state fails if directory contains an invalid symlink


mchugh19:
### Description of Issue/Question
If a file.directory state is run on a directory which contains a broken symlink, the state crashes.

### Setup
test state:
```
/tmp/dirtest:
  file.directory:
    - user: root
    - group: root
    - mode: 755
    - recurse:
      - user
      - group
      - mode
```

### Steps to Reproduce Issue
Success
```
# salt-call state.sls dir
local:
----------
          ID: /tmp/dirtest
    Function: file.directory
      Result: True
     Comment: The directory /tmp/dirtest is in the correct state
     Started: 10:34:25.797989
    Duration: 11.287 ms
     Changes:

Summary for local
------------
Succeeded: 1
Failed:    0
------------
Total states run:     1
Total run time:  11.287 ms
```
Create bad symlink
```
ln -s /tmp/dirtest/thing /tmp/dirtest/nope
```
State now crashes
```
# salt-call state.sls dir
/usr/local/lib/python2.7/dist-packages/requests/__init__.py:83: RequestsDependencyWarning: Old version of cryptography ([1, 2, 3]) may cause slowdown.
  warnings.warn(warning, RequestsDependencyWarning)
[ERROR   ] An exception occurred in this state: Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/salt/state.py", line 1905, in call
    **cdata['kwargs'])
  File "/usr/local/lib/python2.7/dist-packages/salt/loader.py", line 1830, in wrapper
    return f(*args, **kwargs)
  File "/var/cache/salt/minion/extmods/states/file.py", line 3421, in directory
    full, ret, user, group, file_mode, None, follow_symlinks)
  File "/usr/local/lib/python2.7/dist-packages/salt/modules/file.py", line 4460, in check_perms
    lattrs = lsattr(name)
  File "/usr/local/lib/python2.7/dist-packages/salt/modules/file.py", line 564, in lsattr
    raise SaltInvocationError("File or directory does not exist.")
SaltInvocationError: File or directory does not exist.

local:
----------
          ID: /tmp/dirtest
    Function: file.directory
      Result: False
     Comment: An exception occurred in this state: Traceback (most recent call last):
                File "/usr/local/lib/python2.7/dist-packages/salt/state.py", line 1905, in call
                  **cdata['kwargs'])
[...]

PR #49827, opened 2018-09-28
fixing mis-spelling of lattrs in file.py, referncing #49204 

dgmorrisjr:
per request from @gtmanfred, creating a pull request

### What does this PR do?

Correct the mis-spelling of lattrs in the except block, line 4514

when running a file action that recurses (e.g. recurse owner, group, mode, etc), the state would fail.  Found this issue in 2018.1 and discovered that it was being fixed in latest develop branch.

lattrs is mis-spelled in this line. if the except condition fires, then the following `if lattrs is not None:` line will throw an `local variable 'lattrs' referenced before assignment` error.

### Previous Behavior
`local variable 'lattrs' referenced before assignment` error

### Tests written?

No

### Commits signed with GPG?

No

### DIFF
--- a/salt/modules/file.py
+++ b/salt/modules/file.py
@@ -4511,7 +4511,7 @@ def check_perms(name, ret, user, group, mode, attrs=None, follow_symlinks=False)
         try:
             lattrs = lsattr(name)
         except SaltInvocationError:
-            lsattrs = None
+            lattrs = None
         if lattrs is not None:
             # List attributes on file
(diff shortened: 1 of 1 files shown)


gtmanfred (approved):
Awesome, thanks for this @dgmorrisjr

gtmanfred:
great catch!

dgmorrisjr:
happy to help !

PR #49829, opened 2018-09-28
Adding support for cassandra's CREATE ROLE

The-Alchemist:
documentation at https://cassandra.apache.org/doc/4.0/cql/security.html

Needs some more testing

### What does this PR do?
Adds support for cassandra's `CREATE ROLE`

None
### Tests written?

No.  Not sure where the tests should go.  https://github.com/saltstack/salt/blob/develop/tests/unit/modules/test_cassandra_cql.py ?

### Commits signed with GPG?

No


cachedout (approved):

cachedout:
@The-Alchemist Thanks for this! There are some lint warnings which need to be corrected before we can merge this, however. They can be reviewed here: https://jenkinsci.saltstack.com/job/pr-lint/job/PR-49829/2/warnings52Result/new/

Once you've pushed those changes, let us know and we'll take another look. Thanks!

salt/modules/cassandra_cql.py
```diff
@@ -882,6 +882,60 @@ def create_user(username, password, superuser=False, contact_points=None, port=N
...
+    query = '''create role if not exists {0} with password '{1}' and {2} and {3} ;'''.format(username, password, superuser_cql, login_cql)
+    log.debug("Attempting to create a new role with username=%s superuser=%s login=%s", username, superuser, login)
+
+    # The create role query doesn't actually return anything if the query succeeds.
+    # If the query fails, catch the exception, log a messange and raise it again.
+    try:
+        cql_query(query, contact_points, port, cql_user, cql_pass)
+    except CommandExecutionError:
+        log.critical('Could not create role.')
+        raise
```
gtmanfred: Can we just return False here instead of raising the exception?

That way people can handle the return if it was unable to create the role.

salt/modules/cassandra_cql.py
```diff
@@ -882,6 +882,60 @@ def create_user(username, password, superuser=False, contact_points=None, port=N
...
+    except BaseException as e:
+        log.critical('Unexpected error while creating role: %s', e)
+        raise
```
gtmanfred: same here.

gtmanfred (changes requested):

salt/modules/cassandra_cql.py
The-Alchemist: thank you, @gtmanfred .  I wanted to be consistent with the rest of the code, and it seems every function raises some kind of exception on error, especially from `cql_query()`.

The-Alchemist: @gtmanfred : just to clarify, you'd like this PR to change ALL the `cql` functions to just return False?

gtmanfred:
In another PR that also adds handling that in the state, yes.

The-Alchemist:
> In another PR that also adds handling that in the state, yes.

@gtmanfred : I'm sorry, I'm not familiar with that part of the salt code.  What does "handling that in the state" mean?  Could you point me to an example?

Also, are there any other changes you'd like to see  in _this_ PR?

gtmanfred:
Looks like there is not a state module, so nevermind on that.

I would like to see the return changed to False here, instead of a stack trace, and then in another PR fix all the other states in this module, because we should be returning results that are possible to evaluate, instead of just exiting with an exception.

gtmanfred:
you could also return a dictionary or a message explaining why the failure happened, but raising an exception i don't think is appropriate.

cachedout:
@The-Alchemist Are you going to be able to come back to this one?

The-Alchemist:
@cachedout : Probably not this week, so if you wanna run with it, go for it. :)

cachedout:
@The-Alchemist That's fine. We'll check back with you next week. Thanks!

cachedout:
@The-Alchemist Just checking in with you again. :) Let us know if you're able to come back to this.

The-Alchemist:
@cachedout : sorry, I'm out of bandwidth!  It's all yours.

thatch45:
Thanks @The-Alchemist !

PR #49830, opened 2018-09-28
[WIP] remove stuff from pillar top file to see if it breaks anything

gtmanfred:
### What does this PR do?
This is causing a bunch of errors in the test suite.

### Tests written?

not sure

### Commits signed with GPG?

Yes

### DIFF
--- a/tests/integration/files/pillar/base/top.sls
+++ b/tests/integration/files/pillar/base/top.sls
@@ -11,7 +11,3 @@ base:
   'localhost':
     - generic
     - blackout
-  'N@mins not L@minion':
-    - ng1
-  'N@missing_minion':
-    - ng2


rallytime:
@gtmanfred Looks like it did break one test: https://jenkinsci.saltstack.com/job/pr-kitchen-ubuntu1604-py2/job/PR-49830/1/

I can't remember if that one is flaky or not. I know some of these are/were.

gtmanfred:
@mattp- can you comment on what this test is testing? and if it is really necessary? because it is the test is adding a lot of error output to runtests.py any time the pillar data is rendered for a test.

Thanks,
Daniel

mattp-:
the tests in question I added were actually exercising cases when a minion referenced in a node group doesn't exist, so it'd be a loss of some test coverage if we junked it. I think I can move the test specific master config into its own setup/teardown bits instead of it being global to the test suite. should have it ready for tomorrow.

gtmanfred:
Closing in favor of https://github.com/saltstack/salt/pull/49862/files

Issue #2747, opened 2012-11-29
file.recurse exclude_pat improvements (accept a list)


josmardias:
I tried to use exclude_pat but it seems not working for folders, only files. 
I think it will be very useful if we can ignore a subtree by matching a folder.
And i believe that exclude_pat can be an array, so we can match more than one file by glob.

So...

``` yaml
/home/josmar/tree:
  file.recurse:
    - source: salt://test/tree
    - include_empty: True
    - clean: True
    # not sure if this regex is right, but my intention here is 
    # to match all files inside these folders.
    - exclude_pat: 'E@(^excluded/*)|(^holder/excluded/*)'
    - user: josmar
    - group: josmar
    - dir_mode: 550
    - file_mode: 440


/home/josmar/tree/holder/excluded/managed:
  file.managed:
    - user: josmar
    - group: josmar
    - require:
      - file: /home/josmar/tree
```

Can be improved like:

``` yaml

/home/josmar/tree:
  file.recurse:
    - source: salt://test/tree
    - include_empty: True
    - clean: True
    - exclude_pat:
      - excluded
      - holder/excluded
    - user: josmar
    - group: josmar
    - dir_mode: 550
    - file_mode: 440
```

And the whole "/home/josmar/tree/excluded" folder will be ignored by recurse. No copying and no cleaning.

markuskramerIgitt:
@basepi as you gave the advise to use the exclude_pat: ``` "*.svn*" ``` in 2013, should it not work in Salt 2014.1.13 (Windows) and  Salt 2014.7.0 (Debian)?

Unfortunately, this expression copies all ``` .svn  ``` folders:
```
/localdir:
  file.recurse:
    - clean: True
    - source: salt://remotedir
    - exclude_pat: "*.svn*"
 ```

basepi:
@markuskramerIgitt Sounds like a bug. Open a new issue, please?

basepi:
@markuskramerIgitt Please include as much information as you can, including version information.

corywright:
Any idea on when we might see `file.recurse` be able to accept a list?  

I really want to be able to do this:

    /home/build:
      file.recurse:
        - source: salt://buildbox/home/build
        - exclude_pat:
          - .ssh/authorized_keys
          - .gnupg/pubring.gpg
          - .gnupg/secring.gpg

Thanks.

mchugh19:
I think this can be closed with the merged list support for include_pat/exclude_pat.

PR #49837, opened 2018-09-29
Support list for include_pat/exclude_pat in file.recurse

mchugh19:
### What does this PR do?
Allows a list syntax for file.recurse exclude_pat or include_pat
```
/home/build:
  file.recurse:
    - source: salt://buildbox/home/build
    - exclude_pat:
      - .ssh/authorized_keys
      - .gnupg/pubring.gpg
      - .gnupg/secring.gpg
```

#2747

### Previous Behavior
Only string supported

### New Behavior
Now string and list supported

### Tests written?
Yes

### Commits signed with GPG?
No


salt/states/file.py
```diff
@@ -3643,9 +3643,11 @@ def recurse(name,
           - include_pat: E@hello      :: regexp matches 'otherhello',
                                          'hello01' ...
 
+        .. versionchanged:: Neon
```
terminalmage: Please add on a separate line, indented 4 spaces from the `.. versionchanged::`, an explanation of _what_ changed (e.g. "A list of patterns is now supported.")

It would also be helpful to include some SLS examples here in the docstring.

terminalmage (changes requested):

salt/states/file.py
mchugh19: Done!

terminalmage (approved):

djsly:
@mchugh19  / @terminalmage / @rallytime  can someone explains to me why this was merged to `develop` almost 6 months ago and I don't see it in any of the recent tags? It is only present in `develop` :(

rallytime:
Hi @djsly - This was merged into `develop` after the feature branch for 2019.2 was created. As it stands right now, this change will be available in the next feature release, which I believe is codenamed `Neon`.

dynek:
Would love to see this coming to SaltStack! Neon?

mchugh19:
@dynek there was a change to the saltstack development process late 2019. So this will make neon once #54552 is merged. I suggest subscribing there to track status.

PR #49838, opened 2018-09-30
Option print_result should be true by default

dawidmalina:
### What does this PR do?
According documentation default value for print_result should be `true`
> Defaults to True

### Previous Behavior
Failed test result was not displayed by default as described in docs
```
status:
  Fail: Result is not equal
```

### New Behavior
Failed test now properly show status
```
status:
  Fail: True is not equal to False
```

### Tests written?

No

### Commits signed with GPG?

No

### DIFF
--- a/salt/modules/saltcheck.py
+++ b/salt/modules/saltcheck.py
@@ -570,7 +570,7 @@ class SaltCheck(object):
             else:
                 assertion = test_dict['assertion']
             expected_return = test_dict.get('expected-return', None)
-            assert_print_result = test_dict.get('print_result', None)
+            assert_print_result = test_dict.get('print_result', True)
             actual_return = self._call_salt_command(mod_and_func, args, kwargs, assertion_section)
             if assertion not in ["assertIn", "assertNotIn", "assertEmpty", "assertNotEmpty",
                                  "assertTrue", "assertFalse"]:


dawidmalina:
cc @mchugh19

mchugh19:
Looks good, but also shouldn't be necessary. Only some of the assertion types use the print_result flag, and those that do individually default to True.

Your example of assertEqual should default to True, but if that's not working then this looks good.

mchugh19 (commented):
assertEqual's print_result should already default to True

mchugh19 (approved):

dawidmalina:
@mchugh19 I have very simple test:
```
check if /etc/resolv.conf contains is 'nameserver 8.8.4.4':
  module_and_function: file.contains
  args:
    - /etc/resolv.conf
    - 'nameserver 8.8.4.4'
  assertion: assertEqual
  expected-return: True
```

Without this change test result looks like this:
```
          check if /etc/resolv.conf contains is 'nameserver 8.8.4.4':
              ----------
              duration:
                  0.0003
              status:
                  Fail: Result is not equal
```
after this change:
```
          check if /etc/resolv.conf contains is 'nameserver 8.8.4.4':
              ----------
              duration:
                  0.0003
              status:
                  Fail: True is not equal to False
```

I know that we can configure this on test level but according docs we should have this globally enabled:

> **print_result:**
>     (bool) Optional keyword to show results in the ``assertEqual``, ``assertNotEqual``,
>     ``assertIn``, and ``assertNotIn`` output. Defaults to True.

BTW I don't think that this change can result in windows test failure.

mchugh19:
Sounds good.

By the way, there is also an assertTrue, so you don't have to type as much
for the assertEqual

cachedout:
It sounds like everybody is in agreement here so I'll go ahead and get this merged. Thanks all!

Issue #49748, opened 2018-09-22
Should file.rename return success when target exists and force is not used?


DickerDackel:
### Description of Issue/Question

Use-case:
1. Move a pre-packaged configuration of software package-x to package-x.conf.example
2. Place my own package-x.conf there instead

Works correct for the first run, when package-x.conf.example doesn't exist yet.
Fails on consecutive runs, because package-x.conf.example already exists.

Using 'force=True' is not an option, because after the second state.apply, the packaged version would be lost.

Shouldn't the file.rename report a "success, renamed file already in correct state" instead of a failure when force is not used?

Misunderstanding?  Or am I doing something non-idiomatic here and there's a better way?

### Setup
```
# Backup distro provided config as example
/etc/package-x/package-x.conf.example:
    file.rename:
        - source: /etc/package-x/package-x.conf

# Push our own config up there
/etc/package-x/package-x.conf:
    file.managed:
        - source: salt://roles/package-x/package-x.conf
        - user: root
        - group: root
        - mode: 644
```

#### Initial run
```
salt:/srv/salt# salt $HOSTNAME state.apply roles/package-x
salt:
----------
          ID: /etc/package-x/package-x.conf.example
    Function: file.rename
      Result: True
     Comment: Moved "/etc/package-x/package-x.conf" to "/etc/package-x/package-x.conf.example"
     Started: 15:34:37.288072
    Duration: 0.507 ms
     Changes:
              ----------
              /etc/package-x/package-x.conf.example:
                  /etc/package-x/package-x.conf
----------
          ID: /etc/package-x/package-x.conf
    Function: file.managed
      Result: True
     Comment: File /etc/package-x/package-x.conf updated
     Started: 15:34:37.288700
    Duration: 22.224 ms
     Changes:
              ----------
              diff:
                  New file
              mode:
                  0644

Summary for salt
------------
Succeeded: 2 (changed=2)
Failed:    0
------------
Total states run:     2
Total run time:  22.731 ms


-------------------------------------------
Summary
-------------------------------------------
# of minions targeted: 1
# of minions returned: 1
# of minions that did not return: 0
# of minions with errors: 0
-------------------------------------------
salt:/srv/salt#
```

#### next run
````
salt:/srv/salt# salt $HOSTNAME state.apply roles/package-x
salt:
----------
          ID: /etc/package-x/package-x.conf.example
    Function: file.rename
      Result: False
[...]

PR #49843, opened 2018-09-30
Don't return False when target exists (file.rename)

MTecknology:
### What does this PR do?

This PR applies the change recommended by @DickerDackel in #49748.

### Previous Behavior

When using file.rename and then creating a new file in it's place, subsequent executions will fail due to a file existing in it's place.

### New Behavior

This patch removes the ``False`` return when a destination file exists if ``-force:False``.

### Tests written?

Yes

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/states/file.py
+++ b/salt/states/file.py
@@ -6375,7 +6375,6 @@ def rename(name, source, force=False, makedirs=False):
         if not force:
             ret['comment'] = ('The target file "{0}" exists and will not be '
                               'overwritten'.format(name))
-            ret['result'] = False
             return ret
         elif not __opts__['test']:
             # Remove the destination to prevent problems later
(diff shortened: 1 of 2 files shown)


rallytime:
@MTecknology Thanks for submitting this! There's a test failing here: https://jenkinsci.saltstack.com/job/pr-kitchen-ubuntu1604-py2/job/PR-49843/1/

Can you take a look?

terminalmage (approved):
This looks good and the failing test just needs to be updated to reflect the new return value.

damon-atkins:
Please doco the change in behavior in doc/topics/releases/neon.rst

On a side note, not an issue with this PR.   However file.copy() has been change to file.copy_() in develop branch?

terminalmage:
You can ignore that, @MTecknology, there is no need to document a bugfix in the release notes.

terminalmage:
We run a changelog script that adds sort of stuff for us.

terminalmage:
Ahh, after reading the original issue it's clear that this was not 100% bugfix, so I'll get the change documented.

MTecknology:
@terminalmage Thanks!

PR #49858, opened 2018-10-02
Edit requisite documentation

### LINKED ISSUE #49836: require_in doc has conflicting info when many states target same state
`require_in` documentation has seemingly conflicting information about behavior when multiple states `require_in` the same state. 
### Setup
From doc/ref/states/requisites.rst:

http.sls
```
httpd:
  pkg.installed: []
  service.running:
    - require:
      - pkg: httpd
```
php.sls
```
include:
  - http

php:
  pkg.installed:
    - require_in:
      - service: httpd
```
mod_python.sls
```
include:
  - http

mod_python:
  pkg.installed:
    - require_in:
      - service: httpd
```

> Now the httpd server will only start if php or mod_python are first verified to be installed. Thus allowing for a requisite to be defined "after the fact".  

This reads as: "httpd server will only start if _EITHER_ php _OR_ mod_python are verified to be installed"
But earlier in the doc we are told that `require_in` basically inserts a require into the target state. (These two points are the conflict. One says the state will **require** 2 other states to succeed, where as the other says it will **require** only one of the 2 states succeed.)

If this is the case then httpd should have 2 `require`. 
The outcome of this then follows that:
"httpd server will only start if _Both_ php _AND_ mod_python are verified to be installed"

Either the above is true or the documentation is still unclear on the behavior of `require`/`require_in`

Paulo-Nunes:
Fixes #49836

### What does this PR do?
Fixes conflicting information in documentation for `require_in`.

#49836 

### Tests written?

No

### Commits signed with GPG?
No

### DIFF
--- a/doc/ref/states/requisites.rst
+++ b/doc/ref/states/requisites.rst
@@ -632,7 +632,7 @@ mod_python.sls
         - require_in:
           - service: httpd
 
-Now the httpd server will only start if php or mod_python are first verified to
+Now the httpd server will only start if both php and mod_python are first verified to
 be installed. Thus allowing for a requisite to be defined "after the fact".
 
 


rallytime (approved):
Welcome @Paulo-Nunes!

Thanks for submitting this doc fix.

PR #49859, opened 2018-10-02
New Docker Compose module

sco0tr:
Provides functionality to interact directly with the docker-compose CLI.


salt/modules/composeng.py
```diff
@@ -0,0 +1,523 @@
...
+import sys
+import traceback
+import os
+import salt.utils
+from collections import namedtuple
+from contextlib import contextmanager
+from operator import attrgetter
+
+import yaml
+from docopt import docopt
```
gtmanfred: please do `import docopt` and refer to it as `docopt.docopt` otherwise, `composeng.docopt` will get added to the functions in the loader.

salt/modules/composeng.py
```diff
@@ -0,0 +1,523 @@
...
+from __future__ import absolute_import
+import io
+import logging
+import sys
+import traceback
+import os
+import salt.utils
+from collections import namedtuple
+from contextlib import contextmanager
+from operator import attrgetter
```
gtmanfred: same with the above 3 things, we do not want to use `from` because it pollutes the loader namespace

salt/modules/composeng.py
```diff
@@ -0,0 +1,523 @@
...
+
+import yaml
+from docopt import docopt
+from salt.exceptions import CommandExecutionError
+from salt.ext import six
+
+try:
+    from compose.cli.command import get_project
```
gtmanfred: This should also be treated like the above.

salt/modules/composeng.py
```diff
@@ -0,0 +1,523 @@
...
+import os
+import salt.utils
+from collections import namedtuple
+from contextlib import contextmanager
+from operator import attrgetter
+
+import yaml
+from docopt import docopt
+from salt.exceptions import CommandExecutionError
+from salt.ext import six
```
gtmanfred: `import salt.ext.six as six`

salt/modules/composeng.py
```diff
@@ -0,0 +1,523 @@
...
+
+    HAS_COMPOSE = True
+except ImportError as e:
+    HAS_COMPOSE = e
+
+DC_FILENAME = 'docker-compose.yml'
+
+__virtualname__ = 'composeng'
+
+from compose.cli.main import TopLevelCommand, project_from_options
```
gtmanfred: this should probably also be in the above compose lookup.

salt/modules/composeng.py
```diff
@@ -0,0 +1,523 @@
+# -*- coding: utf-8 -*-
+'''
+Work with containers managed by docker-compose
+
```
gtmanfred: can you add `:depends: docker-compose`

salt/modules/composeng.py
```diff
@@ -0,0 +1,523 @@
...
+@__catch_exception()
+def restart(project_dir, *args, **project_options):
+    '''
+    Restart container(s) in the docker-compose file
+
+    :param string project_dir: Path to Docker Compose directory.
+    :param args:
+    :param project_options:
+    :return:
```
gtmanfred: All of these need a CLI Example like other salt modules.

gtmanfred (changes requested):
This looks great! i just have some requests for salt specific things.

This module will also need to have an autodoc file added to `doc/ref/modules/all/salt.modules.composeng.rst` and then composeng will need to be added to the `doc/ref/modules/all/index.rst` file in that directory so it gets rendered

I would also prefer this be named dockercomposeng, instead of just composeng.

salt/modules/dockercomposeng.py
```diff
@@ -0,0 +1,611 @@
...
+    Get convergence plans for a project
+
+    :param project_dir:
+    :return:
+
+    CLI Example:
+
+    .. code-block:: bash
+
+    salt 'myminion' dockercomposeng.get_convergence_plans /path/to/project/dir
```
gtmanfred: This will need to be indented

gtmanfred (changes requested):
Thanks for making those changes, just a few more sphinx things

gtmanfred:
Ok, it all looks good now.

But as I have been looking at the other dockercompose module, I am not actually certain what this adds that makes it better?  and can you please include that in the description of the PR?

gtmanfred:
It would also be great to have some tests for this module.

cachedout:
I have the same question as @gtmanfred. It's unclear to me why we should replace what we have with something new. Could you please explain in detail why these changes are necessary and what advantages this move will have?

dwoz:
If we do replace what we have. I think it would be beneficial to have some tests to verify we are not loosing functionality.

terminalmage (commented):
I'm not familiar with docker-compose so I will defer to @gtmanfred on this. I will say though that unless it changes radically how our current docker-compose support works, I'm not sure it's a great idea to make a completely new module for this.

The reason we had a `dockerng` module for a while was because there were significant design flaws in the previous module, and we did not want to impact current users that relied on the old module and cause their states to break overnight once they upgraded.

cachedout:
@sco0tr We haven't heard from you since you submitted this. Did you see the feedback that came in?

sco0tr:
@gtmanfred @cachedout We wrote this module to increase the amount of functionality we could extract from docker-compose. This module allows us to use any number of the available command line arguments and provides better error handling.  The idea behind this module is to maintain as much feature parity as possible with the previous module while offering support for new functionality out of docker-compose if and when it's released.

gtmanfred:
That is awesome, but what new functionality does this add? And is there a
reason it can't be added to the other module?

sco0tr:
`dockercomposeng` exposes a more accurate and direct interface to the original Compose library.

As the `compose` package from Docker is [intentionally undocumented](https://github.com/docker/compose/issues/4542), the new module chose to interface with Compose through the same code it's CLI interface uses via docopt (the argument parser). This should mean this new module will be more resilient to internal changes in the Compose library. New parameters and options to any `docker-compose` command we have already wraped will be supported with no code changes to the module. The new module also returns unmodified output directly from the Compose library in a more consistent format and throws exceptions when necessary making it easier to work with.

While the core functionality is very similar, the interface with this new module changes how caller's calls and handles output from this module, breaking backwards compatibility. IIRC Salt has a 3 release deprecation notice for breaking changes, this module is here for the transition period. We have no issue of it eventually replacing the old `dockercompose` module.

gtmanfred:
If this added tests i would :+1: it.  outside of that, i don't know that it should be merged in.

leoxiong:
I will finish off writing the tests/additional changes to the new module to help get it merged in, the ETA will be 19th of November when I get some time. @scootr and I worked on this module together so I am familiar with its design.

cachedout:
Hi @leoxiong Any update on the work that was going to be done here?

dwoz:
@sco0tr It seems this PR has gone stale so I am going to close it for now. If you find time to pick up it up again we'll be happy to accept another PR with these changes. Thanks!

PR #49870, opened 2018-10-02
Actually catch the exception when we fail

KaiSforza:
This should actually cause builds to fail when any part in the try block
fails. We still delete instances and try junit rendering, but if things
fail then they get marked on github as failed.

@dubb-b @rallytime @gtmanfred


KaiSforza:
We're going to need to merge this forward soon for the PR builds to fail correctly.

gtmanfred (approved):

gtmanfred:
Can you just open prs for the other two branches?

PR #49874, opened 2018-10-03
allow pillar_cache always valid

rongzeng54:
### What does this PR do?
We set pillar_cache_backend to disk, and we want to control the cache expiration time by ourselves.
Allow pillar_cache always valid by setting pillar_cache_ttl to 0.

No

### Previous Behavior
pillar_cache_ttl is not allowed to never expire.

### New Behavior
pillar_cache_ttl is allowed to never expire.

### Tests written?
No

### Commits signed with GPG?
No


cachedout (approved):

doc/ref/configuration/master.rst
```diff
@@ -4597,7 +4597,7 @@ Default: ``3600``
 
 If and only if a master has set ``pillar_cache: True``, the cache TTL controls the amount
 of time, in seconds, before the cache is considered invalid by a master and a fresh
-pillar is recompiled and stored.
+pillar is recompiled and stored. The 0 means always valid.
```
terminalmage: Please change this added sentence from `The 0 means always valid.` to `A value of 0 will cause the cache to always be valid.`

Also, please make this change in all three config files that you modified.

terminalmage (changes requested):
Just a minor change in the config files is necessary, otherwise this is good to merge.

doc/ref/configuration/master.rst
rongzeng54: already fixed

terminalmage (approved):

Issue #43174, opened 2017-08-24
Get SPM Working on Windows


twangboy:
### Description of Issue/Question
SPM does not work in Windows
Need to get it running and add any batch scripts needed for the installer

### Setup
Clone the salt repo
Create a dev environment (`pkg\windows\build_env_2.ps1`)
Install salt in editable mode (`pip install -e .`)

### Steps to Reproduce Issue
Try to run `spm`. You'll get the following stack trace:
```
Traceback (most recent call last):
  File "C:\Python27\Scripts\spm-script.py", line 11, in <module>
    load_entry_point('salt', 'console_scripts', 'spm')()
  File "c:\salt-dev\salt\salt\scripts.py", line 485, in salt_spm
    import salt.cli.spm
  File "c:\salt-dev\salt\salt\cli\spm.py", line 15, in <module>
    import salt.spm
  File "c:\salt-dev\salt\salt\spm\__init__.py", line 17, in <module>
    import pwd
ImportError: No module named pwd
```

### Versions Report
2017.7 head (2017.7.1+)

twangboy:
@gtmanfred It's in my queue

gtmanfred:
@dwoz you also might be interested in getting this to work.

PR #49875, opened 2018-10-03
Add spm tests to Windows suite

dwoz:
### What does this PR do?

- Adds `integration.spm.*` tests to whitelist.txt for Windows.
- Close sqlite connections so test tearDown methods can remove temporary directories

### Tests written?

No

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/spm/pkgdb/sqlite3.py
+++ b/salt/spm/pkgdb/sqlite3.py
@@ -73,7 +73,9 @@ def info(package, conn=None):
     '''
     List info for a package
     '''
+    close = False
     if conn is None:
+        close = True
         conn = init()
 
     fields = (
@@ -94,6 +96,8 @@ def info(package, conn=None):
         (package, )
     )
     row = data.fetchone()
+    if close:
+        conn.close()
     if not row:
         return None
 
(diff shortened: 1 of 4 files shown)


gtmanfred:
nice! did you get spm to work on windows? https://github.com/saltstack/salt/issues/43174

gtmanfred (approved):

rallytime:
A couple of these new test additions are failing: https://jenkinsci.saltstack.com/job/pr-kitchen-windows2016-py3/job/PR-49875/1/

rallytime:
@dwoz Looks like there is an spm test failing on the windows test. Can you take another look?

salt/spm/pkgdb/sqlite3.py
```diff
@@ -107,32 +111,42 @@ def list_packages(conn=None):
...
     return ret
 
 
 def list_files(package, conn=None):
     '''
     List files for an installed package
     '''
+    close = False
     if conn is None:
+        close = True
```
isbm: You actually can just `None` the `conn` on `.close()` instead of having `close = False | True`:

```python
if conn is not None:
    conn.close()
    conn = None
```

isbm (changes requested):

salt/spm/pkgdb/sqlite3.py
dwoz: Not really, We do not want to close a connection passed in via the `conn` keyword argument.

Issue #47539, opened 2018-05-08
Reactor - No matching sls found for 'orch.minion-onBoot' in env 'base'


1ricardo:
### Description of Issue/Question
After my deep investigation, why I receiving randomly following message:
```sh 
2018-05-08 15:05:55,227 [salt.template    :62  ][ERROR   ][18028] Template was specified incorrectly: False
```
I figure out, that it happening only, when 2 or more minion hits in the reactor in the same time.
It looks like that some cache is cleared or something and not generated on the time, when second minion hit to the Reactor

```sh
salt/run/20180508150554686064/ret       {
    "_stamp": "2018-05-08T13:05:57.144721", 
    "fun": "runner.state.orchestrate", 
    "fun_args": [
        "orch.minion-onBoot", 
        {
            "pillar": {
                "event_data": {
                    "_stamp": "2018-05-08T13:05:54.546766", 
                    "cmd": "_minion_event", 
                    "data": "Minion web220c-stg-1.inet started at Tue May  8 13:05:53 2018", 
                    "id": "web220c-stg-1.inet", 
                    "pretag": null, 
                    "tag": "salt/minion/web220c-stg-1.inet/start"
                }, 
                "event_id": "web220c-stg-1.inet"
            }
        }
    ], 
    "jid": "20180508150554686064", 
    "return": {
        "data": {
            "salt01-tech.inet_master": [
                "No matching sls found for 'orch.minion-onBoot' in env 'base'"
            ]
        }, 
        "outputter": "highstate", 
        "retcode": 1
    }, 
    "success": false, 
    "user": "Reactor"
}
```

I have Auto-Scaling-Group from AWS, which deploys VM to cloud and make basic setup for minion(update master address, setup grains and start service).
After that SALT-Master make the highstate. The problem is that the Reactor works perfectly when there is only 1 new minion, but if I will deploy 2 or more new machine in the same time, then Reacotor will works only for first one which makes attempts, for others Reactor will throw error that some SLS was not found.

For me it looks like that after hit in the Reactor there is some cache clearing and not regenrated on the time.

Flow how looks like the "communication"
minion -> reactor -> orchestrate -> highstate for minion

##### Reactor config
```sh
reactor:
  - 'salt/minion/*/start':
    - /opt/salt/salt-pillar/reactor/minion-onStart.sls
```

##### Reactor state:
```sh
log_new_minion:
  local.cmd.run:
    - name: log new minion
    - tgt: 'apps:saltmaster'
[...]

PR #49888, opened 2018-10-03
[WIP] Allow reactor to react to more events quickly

gtmanfred:
### What does this PR do?
Fix reactor threads not being able to see files on the fileserver

https://github.com/saltstack/salt/issues/47539

### Previous Behavior
Uses Remote fileclient to get orchestration files

### New Behavior
Uses local file_client to get orchestration files

### Tests written?

Yes

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/state.py
+++ b/salt/state.py
@@ -4056,7 +4056,8 @@ class HighState(BaseHighState):
             mocked=False,
             loader='states',
             initial_pillar=None):
-        self.opts = opts
+        self.opts = copy.deepcopy(opts)
+        self.opts['file_client'] = 'local' if self.opts['__role'] == 'master' else self.opts['file_client']
         self.client = salt.fileclient.get_file_client(self.opts)
         BaseHighState.__init__(self, opts)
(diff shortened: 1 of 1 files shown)


gtmanfred:
@terminalmage ok, it looks like this does work fine according to tests, so I think the last thing that we need is a write lock for caching these files since multiple reactors could try to cache the same file at the same time.

salt/state.py
```diff
@@ -4056,7 +4056,8 @@ def __init__(
             mocked=False,
             loader='states',
             initial_pillar=None):
-        self.opts = opts
+        self.opts = copy.deepcopy(opts)
+        self.opts['file_client'] = 'local' if self.opts['__role'] == 'master' else self.opts['file_client']
```
cachedout: It would be good to put some inline docs here explaining why this is necessary so that we don't forget down the road.

gtmanfred:
This doesn't fix the issue, we are looking closer at the fileclient.

PR #49898, opened 2018-10-04
Skip ansiblegate state integration test on CentOS 6

rallytime:
This test is hanging the test suite when running on CentOS 6. I don't think this has ever worked properly (this test is new to fluorine) on CentOS 6, so this will need some further investigation.

Fixes https://github.com/saltstack/salt-jenkins/issues/1138

Skipping for now until we can come back to this.

```
04:45:15          test_install_set_and_remove (integration.states.test_alternatives.AlterantivesStateTest) ... OK (42.623s)
05:15:15 Build timed out (after 30 minutes). Marking the build as aborted.
```

The test that runs directly after the `integration.states.test_alternatives` tests are the `integration.states.test_ansiblegate` tests, which is where the build aborts above.

### DIFF
--- a/tests/integration/states/test_ansiblegate.py
+++ b/tests/integration/states/test_ansiblegate.py
@@ -16,7 +16,11 @@ import salt.utils.path
 
 # Import testing libraries
 from tests.support.case import ModuleCase
-from tests.support.helpers import destructiveTest, requires_sshd_server
+from tests.support.helpers import (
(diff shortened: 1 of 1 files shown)


cachedout (approved):

cachedout:
cc: @isbm

isbm:
@cachedout I didn't wrote that. :wink:

PR #49901, opened 2018-10-04
Fix issues with startup in Windows

twangboy:
### What does this PR do?
Don't load the kernelparams grains on Windows
Properly detect the hostname

Found in testing

### Previous Behavior
The following lines were showing up in the minion start up:
```
[ERROR   ] Having trouble getting a hostname.  Does this machine have its hostname and domain set properly?
[DEBUG   ] Failed to read /proc/cmdline: [Errno 2] No such file or directory: u'/proc/cmdline'
```

### Tests written?
No

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/utils/network.py
+++ b/salt/utils/network.py
@@ -208,6 +208,8 @@ def get_fqhostname():
         )
         for info in addrinfo:
             # info struct [family, socktype, proto, canonname, sockaddr]
+            # On Windows `canonname` can be an empty string
+            # This can cause the function to return `None`
             if len(info) >= 4 and info[3]:
                 l = [info[3]]
     except socket.gaierror:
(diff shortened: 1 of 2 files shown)


salt/utils/network.py
```diff
@@ -208,7 +208,7 @@ def get_fqhostname():
         )
         for info in addrinfo:
             # info struct [family, socktype, proto, canonname, sockaddr]
-            if len(info) >= 4:
+            if len(info) >= 4 and info[3]:
```
rallytime: I'm not quite sure what this check is for. Can you explain this a little more?

dwoz: `canonname` can be an empty string. We don't want to populate the `l` list with empty names. @twangboy , maybe add a comment about this?

twangboy: It's making sure there's a value in info[3]. It's sometimes `''` which causes the function to return None on Windows.

rallytime: OK, yes - let's get a comment in here so it doesn't get removed in the future.

rallytime (approved):

PR #49903, opened 2018-10-04
Improve consul_pillar configuration parsing

fraajad:
### What does this PR do?
Ensure consistency with how the pillar paths are parsed from the consul_pillar configuration.

1. Because of how the regex was, it was possible to get the wrong value for `root` if `pillar_root` was first in the list.

2. A user could be using a trailing slash or not in the config parameters and consul_pillar would perform differently.  The documentation does not use the trailing slash, but the test cases do.  This especially matters in the API call to Consul because if that call does not have a trailing slash, Consul returns all data for keys that the given key is a substring of.

3. Keys in Consul that contained special characters would get processed as regex ie `data.yaml`

### Tests written?

I moved `pillar_root` to the beginning of the config in `test_pillar_nest` to cover case #1.  `test_pillar_data` should cover case #2.  I also removed an unnecessary assertion from the `test_pillar_nest` test.

### Commits signed with GPG?

Yes


salt/pillar/consul_pillar.py
```diff
@@ -267,7 +267,7 @@ def fetch_tree(client, path, expand_keys):
     are folders. Take the remaining data and send it to be formatted
     in such a way as to be used as pillar data.
     '''
-    _, items = consul_fetch(client, path)
+    _, items = consul_fetch(client, path + '/')
```
isbm: The function isn't hidden. So I'd maybe rather check if path has training slash.

isbm (changes requested):

gtmanfred (approved):

salt/pillar/consul_pillar.py
```diff
@@ -267,7 +267,8 @@ def fetch_tree(client, path, expand_keys):
     are folders. Take the remaining data and send it to be formatted
     in such a way as to be used as pillar data.
     '''
-    _, items = consul_fetch(client, path)
+    absolute_path = path.rstrip('/') + '/'
+    _, items = consul_fetch(client, absolute_path)
```
isbm: Nooo, inside the `consul_fetch`. It should look on its own if the path is right or not. If you reuse it elsewhere, you don't know if trailing slash is needed or not, so you will do it every time you call it, and thus duplicate the code.

Please move that rstrip+/ inside the function. :wink:

isbm (changes requested):

salt/pillar/consul_pillar.py
```diff
@@ -258,7 +258,13 @@ def consul_fetch(client, path):
...
     Query consul for all keys/values within base path
     '''
-    return client.kv.get(path, recurse=True)
+    # Unless the root path is blank, it needs a trailing slash for
+    # the kv get from Consul to work as expected
+    if not path:
+        absolute_path = path
+    else:
+        absolute_path = path.rstrip('/') + '/'
+    return client.kv.get(absolute_path, recurse=True)
```
isbm: Almost. :wink:  If you are not sure `path` may be `None`, it might still crash. So:
```python
return client.kv.get(os.path.join(path or ''.rstrip(os.path.sep), ''), recurse=True)
```

Otherwise just `.rstrip` it.

isbm (changes requested):

FraaJad:
How about this for protecting against the None?  The path is a URL path rather than a file path and os lib might use a '\' on Windows.

salt/pillar/consul_pillar.py
isbm: Ah, right for back-slash. Still can be much less coding:
```python
return client.kv.get((path or '').rstrip('/') + '/', recurse=True)
```

isbm (changes requested):

cachedout:
@isbm Re-review requested.

isbm (approved):

Issue #49866, opened 2018-10-02
gem.installed is not recognizing version in dictionary output from gem.list


icycle77:
### Description of Issue/Question

`gem.installed` is not coping with mixed output (specifically the dictionaries that appear alongside strings) from `gem.list` and as a result it never recognizes that the desired gem is already installed, thus always reinstalling it and causing changes where none are needed.

I am in the process of switching from `rvm` to `rbenv` and this problem was discovered during this migration.

### Setup

[Code in question](https://github.com/saltstack/salt/blob/develop/salt/states/gem.py#L86):
```
    gems = __salt__['gem.list'](name, ruby, gem_bin=gem_bin, runas=user)
    if name in gems and version is not None and str(version) in gems[name]:
        ret['result'] = True
        ret['comment'] = 'Gem is already installed.'
        return ret
    elif name in gems and version is None:
        ret['result'] = True
        ret['comment'] = 'Gem is already installed.'
        return ret
```
which runs against the two flavors of output below:

`rbenv` system output (no strings, just a dictionary):
```
$ sudo salt-call --local gem.list json
local:
    ----------
    json:
        - default: 2.1.0
```
`rvm` system output (mix of strings and a dictionary):
```
$  sudo salt-call --local gem.list json
local:
    ----------
    json:
        - 1.8.6
        - 1.8.3
        - default: 1.8.1
```

### Steps to Reproduce Issue
It appears that the code is only able to interpret the strings and not the dictionaries. In the `rbenv` case, there is only a dictionary so that means when running this:
```
# cat /srv/salt/test1.sls
json Gem Installed:
  gem.installed:
    - name: json
    - version: '2.1.0'
```
I always see:
```
[INFO    ] Executing command ['gem', 'list', 'json'] in directory '/root'
[DEBUG   ] stdout: json (default: 2.1.0)
[INFO    ] Executing command ['/usr/local/rbenv/bin/rbenv', 'rehash'] in directory '/root'
[INFO    ] Executing command ['gem', 'install', 'json', '--version', '2.1.0', '--no-rdoc', '--no-ri'] in directory '/root'
[DEBUG   ] stdout: Building native extensions. This could take a while...
Successfully installed json-2.1.0
1 gem installed
```
There should not have been any install because the current version satisfies the desired end state.

### Versions Report
```
$ salt --versions-report
Salt Version:
           Salt: 2017.7.7

Dependency Versions:
           cffi: 1.6.0
       cherrypy: unknown
[...]

PR #49905, opened 2018-10-05
Fix bug in gem.installed related to rvm

kevinquinnyo:
### What does this PR do?
Fixes a bug in gem.installed in not stripping the string 'default: ' from the output of gem list when using rvm
#49866 
### Previous Behavior
It would always try to install the supplied version, say 3.0.0 because it would compare against 'default: 3.0.0' instead of doing nothing since it's already installed.

### New Behavior
Will not install the gem if it already exists

### Tests written?
yes
Yes/No

### Commits signed with GPG?

No

### DIFF
--- a/salt/states/gem.py
+++ b/salt/states/gem.py
@@ -86,7 +86,7 @@ def installed(name,          # pylint: disable=C0103
     gems = __salt__['gem.list'](name, ruby, gem_bin=gem_bin, runas=user)
 
     if name in gems:
-        versions = list(map(lambda x: x.replace('default: ', ''), gems[name]))
+        versions = list([x.replace('default: ', '') for x in gems[name]])
         if version is not None and str(version) in versions:
             ret['result'] = True
             ret['comment'] = 'Gem is already installed.'
--- a/salt/states/gem.py
+++ b/salt/states/gem.py
@@ -84,14 +84,17 @@ def installed(name,          # pylint: disable=C0103
             'Use of argument ruby found, but neither rvm or rbenv is installed'
         )
     gems = __salt__['gem.list'](name, ruby, gem_bin=gem_bin, runas=user)
-    if name in gems and version is not None and str(version) in gems[name]:
-        ret['result'] = True
-        ret['comment'] = 'Gem is already installed.'
-        return ret
-    elif name in gems and version is None:
-        ret['result'] = True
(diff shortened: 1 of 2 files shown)


kevinquinnyo:
I'm not a regular contributor, so someone might want to review this.  Also I only ran the specific tests:
```
 tests/runtests.py --name=unit.states.test_gem
```

Assuming I understood the issue, I *think* this resolves it.

rallytime:
Thanks @kevinquinnyo for contributing!

The linter caught a py2/py3 compatibility issue summarized here:
```
20:32:42 [salt linting] ************* Module salt.states.gem
20:32:42 [salt linting] salt/states/gem.py:89: [W1699(incompatible-py3-code), ] Incompatible Python 3 code found. Proposed fix:
20:32:42 [salt linting]      gems = __salt__['gem.list'](name, ruby, gem_bin=gem_bin, runas=user)
20:32:42 [salt linting]  
20:32:42 [salt linting]      if name in gems:
20:32:42 [salt linting] -        versions = list(map(lambda x: x.replace('default: ', ''), gems[name]))
20:32:42 [salt linting] +        versions = list([x.replace('default: ', '') for x in gems[name]])
20:32:42 [salt linting]          if version is not None and str(version) in versions:
20:32:42 [salt linting]              ret['result'] = True
20:32:42 [salt linting]              ret['comment'] = 'Gem is already installed.'
20:32:42 [salt linting] ERROR: InvocationError for command '/var/jenkins/workspace/pr-lint_PR-49905-UAL4R762YPURFIUO77YIAUWQPDS4IEIQALZBAO6AO7VAOHLJ5PMQ/.tox/pylint-salt/bin/pylint --rcfile=.testing.pylintrc --disable=I,W1307,C0411,C0413,W8410,str-format-in-logging salt/states/gem.py' (exited with code 4)
20:32:42 [salt linting] ___________________________________ summary ____________________________________
20:32:42 [salt linting] ERROR:   pylint-salt: commands failed
```

Can you fix that up?

kevinquinnyo:
@rallytime no problem, done

rallytime:
Hey @kevinquinnyo - we ran into some trouble with provisioning our test boxes the other day. I've restarted the tests here, and the we can see how the results look. :)

Thank you for fixing this!

cachedout (approved):

PR #49906, opened 2018-10-05
Update documentation about state ordering. (Fixes #45288)

### LINKED ISSUE #45288: Docs about Ordering States
I think the current docs of "Ordering states" could be improved.

> Salt always executes states in a finite manner, meaning that they will always execute in the same order regardless of the system that is executing them. But in Salt 0.17.0, the state_auto_order option was added. This option makes states get evaluated in the order in which they are defined in sls files, including the top.sls file.

>The evaluation order makes it easy to know what order the states will be executed in, but it is important to note that the requisite system will override the ordering defined in the files, and the order option described below will also override the order in which states are defined in sls files.

> If the classic ordering is preferred (lexicographic), then set state_auto_order to False in the master configuration file. Otherwise, state_auto_order defaults to True.

Source https://docs.saltstack.com/en/latest/ref/states/ordering.html#state-auto-ordering

The version 0.17.0 was released on 2013-09-26. I guess 99% of all users use a newer version.

I think it would be much more easier to read, if you just present the current state exclusively. Users of the four years old version should read the four years old docs :-)

This is current state: States get evaluated in the order in which they are defined in sls files.

MTecknology:
### What does this PR do?

This is a simple documentation update.

- #45288

### Commits signed with GPG?

Yes

### DIFF
--- a/doc/ref/states/ordering.rst
+++ b/doc/ref/states/ordering.rst
@@ -5,10 +5,10 @@ Ordering States
 ===============
 
 The way in which configuration management systems are executed is a hotly
-debated topic in the configuration management world. Two
-major philosophies exist on the subject, to either execute in an imperative
-fashion where things are executed in the order in which they are defined, or
-in a declarative fashion where dependencies need to be mapped between objects.
+debated topic in the configuration management world. Two major philosophies
+exist on the subject, to either execute in an imperative fashion where things
+are executed in the order in which they are defined, or in a declarative
+fashion where dependencies need to be mapped between objects.
 
 Imperative ordering is finite and generally considered easier to write, but
 declarative ordering is much more powerful and flexible but generally considered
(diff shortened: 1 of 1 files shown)


doc/ref/states/ordering.rst
```diff
@@ -24,23 +24,18 @@ dependencies via the `requisite` system.
 State Auto Ordering
 ===================
 
-.. versionadded: 0.17.0
```
rallytime: Why are you removing this?

doc/ref/states/ordering.rst
```diff
@@ -24,23 +24,18 @@ dependencies via the `requisite` system.
...
-But in Salt 0.17.0, the ``state_auto_order`` option was added. This option
-makes states get evaluated in the order in which they are defined in sls
-files, including the top.sls file.
-
-The evaluation order makes it easy to know what order the states will be
-executed in, but it is important to note that the requisite system will
-override the ordering defined in the files, and the ``order`` option described
-below will also override the order in which states are defined in sls files.
+execute in the same order regardless of the system that is executing them. This
+evaluation order makes it easy to know what order the states will beexecuted in,
```
rallytime: Typo `beexecuted` --> `be executed`

rallytime (commented):
Some small comments :)

doc/ref/states/ordering.rst
MTecknology: Thanks for catching that!

doc/ref/states/ordering.rst
MTecknology: According to (modified) version history, I would never ever have done such a silly thing! ;)

rallytime (approved):

PR #49914, opened 2018-10-05
Fix zypper.list_pkgs for armv7

matwey:
`__grains__['osarch']` returns 'armv7l'
`salt.utils.pkg.rpm.getosarch()` returns 'armv7hl'
openSUSE RPM packages for armv7 are '*.armv7hl.rpm'

Consequence of mismatched architecture is broken 'state.installed'.
salt.utils.pkg.rpm.resolve name() appends 'armv7hl' to package name and
installation check is failed.

### What does this PR do?

Fix state.installed at openSUSE armv7l installations

https://bugzilla.suse.com/show_bug.cgi?id=1109384

### Previous Behavior

```
# cat top.sls
base:
  '*':
    - git
# cat git.sls
git:
  pkg.installed: []
# rpm -q git
git-2.16.4-lp150.2.3.1.armv7hl
```

```
local:
----------
          ID: git
    Function: pkg.installed
      Result: False
     Comment: The following packages failed to install/update: git
     Started: 18:24:07.730221
    Duration: 169840.00000000015 ms
     Changes:   

Summary for local                                                                                                                      
------------                                                                                                                           
Succeeded: 0
Failed:    1
------------
Total states run:     1                                                                                                                
Total run time: 169.840 s
```

### New Behavior

```
local:
----------
          ID: git
    Function: pkg.installed
      Result: True
     Comment: All specified packages are already installed
     Started: 18:55:25.719224
    Duration: 108889.99999999942 ms
     Changes:   

Summary for local                                                                                                                      
------------                                                                                                                           
Succeeded: 1
Failed:    0
------------
Total states run:     1                                                                                                                
Total run time: 108.890 s
```

### Tests written?

No

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/modules/zypper.py
+++ b/salt/modules/zypper.py
@@ -751,10 +751,7 @@ def list_pkgs(versions_as_list=False, **kwargs):
                                      python_shell=False,
                                      output_loglevel='trace')
         for line in output.splitlines():
-            pkginfo = salt.utils.pkg.rpm.parse_pkginfo(
-                line,
-                osarch=__grains__['osarch']
-            )
+            pkginfo = salt.utils.pkg.rpm.parse_pkginfo(line)
             if pkginfo:
                 # see rpm version string rules available at https://goo.gl/UGKPNd
                 pkgver = pkginfo.version
@@ -1590,8 +1587,8 @@ def normalize_name(name):
             return name
     except ValueError:
         return name
-    if arch in (__grains__['osarch'], 'noarch') \
-            or salt.utils.pkg.rpm.check_32(arch, osarch=__grains__['osarch']):
+    if arch in (salt.utils.pkg.rpm.getosarch(), 'noarch') \
+            or salt.utils.pkg.rpm.check_32(arch):
         return name[:-(len(arch) + 1)]
     return name
 
--- a/salt/utils/pkg/rpm.py
+++ b/salt/utils/pkg/rpm.py
@@ -49,7 +49,7 @@ def get_osarch():
         shell=True,
         close_fds=True,
         stdout=subprocess.PIPE,
-        stderr=subprocess.PIPE).communicate()[0]
+        stderr=subprocess.PIPE).communicate()[0].decode().rstrip()
     return ret or 'unknown'
 
 


salt/modules/zypper.py
```diff
@@ -751,10 +751,7 @@ def list_pkgs(versions_as_list=False, **kwargs):
                                      python_shell=False,
                                      output_loglevel='trace')
         for line in output.splitlines():
-            pkginfo = salt.utils.pkg.rpm.parse_pkginfo(
-                line,
-                osarch=__grains__['osarch']
-            )
+            pkginfo = salt.utils.pkg.rpm.parse_pkginfo(line)
```
brejoc: I'm not sure if we can safely remove `osarch` here. Wouldn't it be better to check if `__grains__['osarch']` is `armv7l` and then adapt?

matwey: `osarch` is redundant here, check `salt.utils.pkg.rpm` sources

isbm (changes requested):
While `uname -m` might return something interesting on ARM platform, I created a fix for it: https://github.com/saltstack/salt/pull/49930
As of this very PR, it is no longer necessary therefore. Please close this one.

brejoc:
Your PR sparked some discussions here at SUSE this morning, @matwey. Thank you very much for bringing this up. But I have to agree with @isbm – this should be fixed at grains level like [this](https://github.com/saltstack/salt/pull/49930). Otherwise we'd always work around the core problem.

matwey:
It would be great if you fixed it for Leap 15.0

isbm:
@matwey this fix will be added to the packages, once it is merged here. Thanks!

brejoc:
> It would be great if you fixed it for Leap 15.0

You can check the changelog or spec file of the RPM file for Tumbleweed to check if the patch made it in. The pull requests are usually also linked in the spec file.

isbm:
Patch is always added [here](https://build.opensuse.org/package/show/systemsmanagement:saltstack:testing/salt). From [there](https://build.opensuse.org/package/show/systemsmanagement:saltstack:testing/salt) it gets all elsewhere.

cachedout:
IMHO, it looks like this should be closed. Any last-minute objections before we do so?

isbm:
@matwey [JFYI](https://build.opensuse.org/package/view_file/systemsmanagement:saltstack:testing/salt/get-os_family-for-rpm-distros-from-the-rpm-macros.-u.patch?expand=1)

Issue #49432, opened 2018-08-29
Handle RAET deprecation for Neon release


rallytime:
- [ ] salt/daemons/flo/__init__.py

PR #49916, opened 2018-10-05
Remove support for RAET

rallytime:
RAET support ends with the Neon release.

Fixes https://github.com/saltstack/salt/issues/49432

I want to see what the tests think about this big change first, hence the WIP.

NOTE: I did not remove any mentions of raet or it's associated packages (ioflow/libnacl/libsodium) from any files in the `pkg/*` directory yet. I will do that in a follow up PR.


isbm (approved):

gtmanfred (approved):

twangboy:
@rallytime So can we remove these as dependencies in Windows as well?
```
libnacl>=1.0.0
ioflo>=1.1.7
```

thatch45:
ioflo can go, but libnacl should probably stay. It is used in other places :)

salt/client/__init__.py
```diff
@@ -97,11 +97,8 @@ def get_local_client(
         # Late import to prevent circular import
         import salt.config
         opts = salt.config.client_config(c_path)
-    if opts['transport'] == 'raet':
-        import salt.client.raet
-        return salt.client.raet.LocalClient(mopts=opts)
     # TODO: AIO core is separate from transport
-    elif opts['transport'] in ('zeromq', 'tcp'):
+    if opts['transport'] in ('zeromq', 'tcp'):
```
DmitryKuzmenko: Redundant `if`.

salt/config/__init__.py
```diff
@@ -956,7 +956,7 @@ def _gather_buffer_space():
     # The size of key that should be generated when creating new keys
     'keysize': int,
 
-    # The transport system for this daemon. (i.e. zeromq, raet, etc)
+    # The transport system for this daemon. (i.e. zeromq, tcp, etc)
```
DmitryKuzmenko: Isn't it better to specify all 3 currently acceptable values here? `(i.e. zeromq, tcp or detect)`

salt/key.py
```diff
@@ -309,12 +295,9 @@ def __init__(self, opts):
         # Remove the key attribute set in KeyCLI.__init__
```
DmitryKuzmenko: It looks class `MultiKeyCLI` is not needed anymore.

salt/utils/event.py
```diff
@@ -143,12 +143,6 @@ def get_event(
                          io_loop=io_loop,
```
DmitryKuzmenko: Redundant `if` above.

DmitryKuzmenko (changes requested):

salt/cli/daemons.py
```diff
@@ -119,7 +119,7 @@ class Master(salt.utils.parsers.MasterOptionParser, DaemonsMixin):  # pylint: di
     Creates a master server
     '''
     def _handle_signals(self, signum, sigframe):  # pylint: disable=unused-argument
-        if hasattr(self.master, 'process_manager'):  # IofloMaster has no process manager
+        if hasattr(self.master, 'process_manager'):
```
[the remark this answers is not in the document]
rallytime: I wasn't _entirely_ sure what else might be using that, so I figured it didn't hurt to keep that there.

salt/auth/__init__.py
```diff
@@ -680,11 +680,6 @@ def _send_token_request(self, load):
                                                                 master_uri=master_uri)
```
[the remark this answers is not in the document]
rallytime: Yeah, it is. I was thinking about leaving those there in case we ever implemented some other transport options, but I can remove that for now and the others you mentioned below.

salt/config/__init__.py
rallytime: Yes, probably. Though I believe the `detect` functionality is broken. :(

I will fix that comment though. :D

salt/key.py
rallytime: You're right that it's not being used in Salt code anymore. I wasn't sure if I should remove that here in case other people were using it outside. But it's probably best to not maintain this since it's part of the raet support.

salt/cli/daemons.py
```diff
@@ -499,12 +472,6 @@ def prepare(self):
             if self.config.get('master_type') == 'func':
```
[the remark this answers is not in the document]
rallytime: This one is OK because we're logging that the transport type isn't found below.

DmitryKuzmenko (approved):

DmitryKuzmenko:
@rallytime sorry I blocked this accidentally. Just wanted to leave a review to point things I've found interesting.

rallytime:
@DmitryKuzmenko No problem at all! I am happy for the thorough review. Thank you :)

salt/auth/__init__.py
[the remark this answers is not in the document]
DmitryKuzmenko: I thought about it too. But then thought that not having that `if` will follow the developer of the new transport to a better path that is followed by both `tcp` and `zeromq` transports that don't need that `if` case. :smile:

PR #49920, opened 2018-10-06
Don't log in `__virtual__`, specially at the error level

s0undt3ch:
### What does this PR do?
Removes log calls under `__virtual__`


garethgreenaway (approved):

s0undt3ch:
Don't we have a lint rule for this yet?

rallytime:
@s0undt3ch We do not. I've created an issue so we can get that added. Thanks for fixing this!

PR #49930, opened 2018-10-08
Get os_family for RPM distros from the RPM macros

isbm:

Bugfix: sometimes `uname -m` is not what RPM macros returns. Therefore copying `cpuarch` over `osarch` is not the best idea, as it results to unwanted effects while installing packages on less used systems (ARM for example).

It is also that original `get_osarch` function returns `bytes` type with the `os.linesep` attached alongside Unicode string.

### Tests written?

No, should just apply existing

### DIFF
--- a/salt/grains/core.py
+++ b/salt/grains/core.py
@@ -46,6 +46,7 @@ import salt.utils.files
 import salt.utils.network
 import salt.utils.path
 import salt.utils.platform
+import salt.utils.pkg.rpm
 from salt.ext import six
 from salt.ext.six.moves import range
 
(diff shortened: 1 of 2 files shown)


isbm:
@rallytime this should be forwarded to `develop` too.

meaksh (approved):

brejoc (approved):
lgtm

rallytime (approved):

Poil:
Why not use the standard platform library from python ? Any case where we have a 32 bit interpreter on a 64 bit OS ?

isbm:
@Poil because I explained this in the description of the bug: `uname -m` is what Python's `platform` is doing and it is `cpuarch`. And on some platforms it returns different name than RPM database. In other words, `osarch` and `cpuarch` may differ.

Issue #25697, opened 2015-07-24
'virtual' grain returns `VirtualBox` when running `systemd-nspawn` nested inside VirtualBox


eliasp:
I have the following setup here:
- Fedora 22 in VirtualBox 5.0
- Ubuntu 15.04 inside a systemd-nspawn container (systemd 219)
- Salt 2015.5.2 in this Ubuntu/systemd-nspawn container

The `virtual` grain returns `VirtualBox` because it uses `osdata.get('productname', '')` first which returns `VirtualBox` and then doesn't even query `virt-what` anymore.
It should return `lxc` (as reported by `virt-what` instead).

```
                      Salt: 2015.5.2
                    Python: 2.7.9 (default, Apr  2 2015, 15:33:21)
                    Jinja2: 2.7.3
                  M2Crypto: 0.21.1
            msgpack-python: 0.4.2
              msgpack-pure: Not Installed
                  pycrypto: 2.6.1
                   libnacl: Not Installed
                    PyYAML: 3.11
                     ioflo: Not Installed
                     PyZMQ: 14.4.1
                      RAET: Not Installed
                       ZMQ: 4.0.5
                      Mako: 1.0.0
     Debian source package: 2015.5.2+ds-1utopic1
```

jfindlay:
@eliasp, thanks for the report.

eliasp:
Still applies on 2018.3.2, running on Ubuntu 18.04, inside systemd-nspawn inside Hyper-V. 

Running virt-what (1.18) returns:
hyperv
lxc

rallytime:
Thanks @eliasp for letting us know! I've reopend this issue.

PR #49937, opened 2018-10-08
Fix value of `virtual` grain when using `virt-what`

eliasp:
### What does this PR do?

`virt-what` might return multiple lines, depending on what it detected (e.g. `lxc` inside a `HyperV` VM), where the last line corresponds to the most likely match.

Until now, Salt evaluated the whole multi-line output as a single string, which caused quite inaccurate matches. This commit changes this behavior to only evaluate the last line of the `virt-what` output.

A similar issue is described in #25697, but might actually have a different root cause.

### Previous Behavior
The `virtual` grain reported `HyperV`, although Salt was executed in a `systemd-nspawn` container inside a HyperV-based Ubuntu VM.

### New Behavior

The `virtual` grain reports now `lxc` which reflects what `virt-what` determines.

### Tests written?

No

### Commits signed with GPG?

Yes

### DIFF
--- a/doc/topics/releases/fluorine.rst
+++ b/doc/topics/releases/fluorine.rst
@@ -488,6 +488,18 @@ in.  Because of the non-deterministic order that grains are rendered in, the
 only grains that can be relied upon to be passed in are ``core.py`` grains,
 since those are compiled first.
 
+More Precise ``virtual`` Grain
+==============================
+
+This release improves the accuracy of the ``virtual`` grain when running Salt in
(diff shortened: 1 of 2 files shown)


cachedout (commented):
I am hesitant to merge this without calling it out in the release notes. We should make it very clear whenever a grain that may be used for targeting is modified so that people can test their upgrade paths.

cachedout:
@eliasp Can you add this to the release notes? Then I think we can safely get this in.

cachedout:
Bump @eliasp

eliasp:
@cachedout Done.

PR #49940, opened 2018-10-08
Clean up documentation/language in service state. (Fixes #40819)

### LINKED ISSUE #40819: please clarify documentation on service watch
The document at https://docs.saltstack.com/en/latest/ref/states/all/salt.states.service.html is too confusing for its own good when it comes to the watching. The description at salt.states.service.mod_watch doesn't go into any nuance, it just says:

    "The service watcher, called to invoke the watch command."

If the user looks that up, they will miss some of the nuance from the general section:

    "By default if a service is triggered to refresh due to a watch statement the service is by default restarted."

This sentence should be moved or copied under mod_watch.

Also, the term "by default" is repeated twice in the same sentence, so one of those should be removed.

Furthermore, the descriptions of options are problematic:

    "Use reload instead of the default restart (exclusive option with full_restart, defaults to reload if both are used)"

Please rephrase the part in parenthesis, it sounds like it's trying to map boolean logic into simple English, but it's not actually succeeding, since the phrase "Y = Exclusive option with X" is a bad translation of XOR - just say "Y cannot be combined with X". Pseudo-code would have been more readable than this :)

    "Use service.full_restart instead of restart (exclusive option with reload)"

Ditto for the parenthesized part.

In context of a state named "service", the reference to "service.full_restart", which I assume is to the module named "service", is unclear. Especially when one can't actually find that at https://docs.saltstack.com/en/latest/ref/modules/all/salt.modules.service.html

Also, somewhere in there it wouldn't hurt to explain briefly what is meant by "full" restart - someone might think it's service(8)'s `--full-restart`, but it could also be assumed to be the module running a `stop` followed by a `start` itself, or maybe some third option I'm not currently acquainted with.

Please fix this. TIA.

MTecknology:
### What does this PR do?

Cleans up some potentially confusing documentation.

- #40819

### Commits signed with GPG?

Yes

Disclaimer: I had a hard time coming up with a simple way to describe mod_watch. Once reviewed, I would like to update the documentation for other mod_watch functions.


rallytime (approved):
Thank you :)

MTecknology:
@rallytime Thanks for the ping on this one! Is there any chance you could verify that the current mod_watch description is adequate? I feel like a cleaner requisites document might help the clarity of this note.

@shallot Any thoughts on these changes?

rallytime:
@MTecknology It looks good to me, but maybe @gtmanfred has some suggestions/comments.

gtmanfred (approved):
That looks good to me!

cachedout (approved):

cachedout:
@rallytime I think we could safely backport this and it might be helpful to get this into the current supported branches.

MTecknology:
Thanks for the reviews!

Issue #49927, opened 2018-10-08
File Tree external pillar fails to render templates in python 3


thetaurean:
### Description of Issue/Question
The File Tree external pillar fails to render templates in python 3. The root cause is the file_tree module passes bytes to the template renderer instead of a string.

 passes bytes instead of string to SLS_ENCODER.
### Setup
In your salt master config
```
  - file_tree:
      root_dir: /srv/pillar
      render_default: yaml
      renderer_whitelist:
        - yaml
      template: True
```
### Steps to Reproduce Issue
Simple add a properly formatted yaml hosts file to /srv/pillar/hosts/minion_id and try to refresh pillars.
You will see that salt fails with the following error.
```
2018-10-08 03:51:39,344 [salt.pillar      :972 ][ERROR   ][11949] Execption caught loading ext_pillar 'file_tree':
  File "/usr/lib/python3/dist-packages/salt/pillar/__init__.py", line 962, in ext_pillar
    key)
  File "/usr/lib/python3/dist-packages/salt/pillar/__init__.py", line 881, in _external_pillar_data
    ext = self.ext_pillars[key](self.minion_id, pillar, **val)
  File "/usr/lib/python3/dist-packages/salt/pillar/file_tree.py", line 443, in ext_pillar
    template)
  File "/usr/lib/python3/dist-packages/salt/pillar/file_tree.py", line 536, in _ext_pillar
    template)
  File "/usr/lib/python3/dist-packages/salt/pillar/file_tree.py", line 258, in _construct_pillar
    whitelist=renderer_whitelist)
  File "/usr/lib/python3/dist-packages/salt/template.py", line 142, in compile_template_str
    ofile.write(SLS_ENCODER(template)[0])
2018-10-08 03:51:39,345 [salt.pillar      :1019][CRITICAL][11949] Pillar render error: Failed to load ext_pillar file_tree: Can't convert 'bytes' object to str implicitly
2018-10-08 03:51:43,537 [salt.utils.parsers:1045][WARNING ][11928] Master received a SIGTERM. Exiting.
```

Changing lines 141-142 in template.py as follows is a temporary fix

```
with salt.utils.files.fopen(fn_, 'wb') as ofile:
        ofile.write(SLS_ENCODER(template.decode('utf-8'))[0])
```

The more appropriate fix probably belongs in file_tree.py:
```
else:
                data = contents # decode here
                if template is True:
                    data = salt.template.compile_template_str(template=contents,
                                                              renderers=renderers,
[...]

PR #49943, opened 2018-10-08
Coerce string  when calling compile_template_str

thetaurean:
### What does this PR do?
This fixes a bug in the file_tree pillar which causes it to fail to render templates in python 3. The caller for the template rendering function passes bytes instead of string.

Issue #49927 

### Previous Behavior
```
else:
                data = contents
                if template is True:
                    data = salt.template.compile_template_str(template=contents,
                                                              renderers=renderers,
                                                              default=render_default,
                                                              blacklist=renderer_blacklist,
                                                              whitelist=renderer_whitelist)
```
### New Behavior
```
else:
                data = contents
                if template is True:
                    data = salt.template.compile_template_str(template=salt.utils.stringutils.to_unicode(contents),
                                                              renderers=renderers,
                                                              default=render_default,
                                                              blacklist=renderer_blacklist,
                                                              whitelist=renderer_whitelist)
```
### Tests written?

No. New functionality not added

### Commits signed with GPG?

Yes


salt/pillar/file_tree.py
```diff
@@ -158,6 +158,7 @@
 import salt.utils.stringio
 import salt.template
 from salt.ext import six
+from salt.utils.stringutils import to_str
```
gtmanfred: please don't import functions like this, it will add it to the functions in the modules.

```
import salt.utils.stringutils
```

salt/pillar/file_tree.py
```diff
@@ -251,7 +252,7 @@ def _construct_pillar(top_dir,
             else:
                 data = contents
                 if template is True:
-                    data = salt.template.compile_template_str(template=contents,
+                    data = salt.template.compile_template_str(template=to_str(contents),
```
gtmanfred: `salt.utils.stringutils.to_str(contents)`

gtmanfred (changes requested):
just a styling change.

gtmanfred:
@terminalmage should this be to_str? or something else?

Thanks,
Daniel

thetaurean:
@gtmanfred  Addressed. Thank you.

gtmanfred (approved):
This looks good to me, still want @terminalmage or someone else to take a look.

gtmanfred:
I have also marked this to be backported, for future reference, PRs that fix bugs should be opened against the oldest feature release that the bug is present in, so in this case that would be the 2018.3 branch. (2018.3.3 is a release branch and not the main feature branch)

thetaurean:
@gtmanfred Good point, aww shucks!

I am happy to rebase the changes on 2018.3.3 and force push, thoughts?

gtmanfred:
There is 0 chance this makes it in 2018.3.3, it has already been tagged and is being packaged.

But i have marked it for backporting, so it will be in 2018.3.4

isbm (approved):

dwoz (approved):

terminalmage (changes requested):
`salt.templates.compile_template_str()` writes the template to a tempfile in binary mode, and therefore encodes to UTF-8 before doing so. This is why Python 3 raised the error.

However, because we are encoding, we want the contents to be unicode first, otherwise Python 2 will break when non-ascii characters are present in the string we pass:

```pytb
>>> 'groß'.encode('utf-8')
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 3: ordinal not in range(128)
```

Therefore, please change from `to_str()` to `to_unicode()`.

thetaurean:
@terminalmage Addressed. Excellent observation and appreciate the insight gained from your thorough explanation 👍

terminalmage (approved):
Well done, thank you for contributing!

roaldnefs:
Already ported to master in #50052.

PR #49946, opened 2018-10-09
osversion/osversion_info/osmajorrelease grains fixes

isbm:
### What does this PR do?

Bugfix for `osversion`, `osversion_info` and `osmajorrelease` grains on SUSE systems.

As per documentation of [`/etc/os-release`](https://www.linux.org/docs/man5/os-release.html) file:

> VERSION_ID= 
A lower-case string (mostly numeric, no spaces or other characters outside of 0-9, a-z, ".", "_" and "-") identifying the operating system version, excluding any OS name information or release code name, and suitable for processing by scripts or usage in generated filenames. This field is optional.
Example:
"VERSION_ID=17" or "VERSION_ID=11.04".

This in reality this so far always worked, because nobody adds there anything else but numbers and dots. However, it allows symbols. And thus if one is adding something like `VERSION_ID=foo.bar` which is still allowed, the grain `osmajorrelease` will disappear, since it is expected to be an integer.

`CPE_NAME`, in contrary, has a clear definition between version number and its phase (for "beta", "alpha" etc). Unfortunately, `CPE_NAME` is still optional. But it is provided on the systems where no `/etc/lsb-release` with `DISTRIB_RELEASE` like in Debian, for example. Therefore Salt should try look up for `CPE_NAME` as well and fall-back to `VERSION_ID` only if nothing else has been found to figure out distribution version. Also, on some distributions `CPE_NAME` is still broken (only major version specified).

### Previous Behavior

```
# cat /etc/os-release | grep VERSION_ID
VERSION_ID="foo.bar"

$ salt-call --local grains.get osrelease
local:
    foo.bar

$ salt-call --local grains.get osrelease_info
local:
    - foo
    - bar

$ salt-call --local grains.get osmajorrelease
local:
```

### New Behavior

```
# cat /etc/os-release | grep VERSION_ID
VERSION_ID="foo.bar"

$ salt-call --local grains.get osrelease
local:
    15.0

$ salt-call --local grains.get osrelease_info
local:
    - 15
    - 0

# salt-call --local grains.get osmajorrelease
local:
    15

# salt-call --local grains.get osmajorrelease --out json
{
    "local": 15
}
```

### Tests written?

Yes

### DIFF
--- a/salt/grains/core.py
+++ b/salt/grains/core.py
@@ -1679,7 +1679,9 @@ def os_data():
                         grains['lsb_distrib_codename'] = codename
                     if 'CPE_NAME' in os_release:
                         cpe = _parse_cpe_name(os_release['CPE_NAME'])
-                        if cpe['vendor'].lower() in ['suse', 'opensuse']:
+                        if not cpe:
+                            log.error('Broken CPE_NAME format in /etc/os-release!')
+                        elif cpe.get('vendor', '').lower() in ['suse', 'opensuse']:
                             grains['os'] = "SUSE"
                             # openSUSE `osfullname` grain normalization
                             if os_release.get("NAME") == "openSUSE Leap":
--- a/salt/grains/core.py
+++ b/salt/grains/core.py
@@ -1695,9 +1695,9 @@ def os_data():
                                 grains['osfullname'] = "Leap"
                             elif os_release.get("VERSION") == "Tumbleweed":
                                 grains['osfullname'] = os_release["VERSION"]
-                        # Override VERSION_ID, if CPE_NAME around
-                        if cpe.get('version'):
-                            grains['lsb_distrib_release'] = cpe['version']
+                            # Override VERSION_ID, if CPE_NAME around
+                            if cpe.get('version') and cpe.get('vendor') == 'opensuse':  # Keep VERSION_ID for SLES
+                                grains['lsb_distrib_release'] = cpe['version']
 
                 elif os.path.isfile('/etc/SuSE-release'):
                     log.trace('Parsing distrib info from /etc/SuSE-release')
--- a/salt/grains/core.py
+++ b/salt/grains/core.py
@@ -1466,6 +1466,27 @@ def _parse_os_release(*os_release_files):
     return ret
 
 
+def _parse_cpe_name(cpe):
+    '''
+    Parse CPE_NAME data from the os-release
+
+    Info: https://csrc.nist.gov/projects/security-content-automation-protocol/scap-specifications/cpe
+
+    :param cpe:
+    :return:
+    '''
+    ret = {}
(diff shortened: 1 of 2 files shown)


salt/grains/core.py
```diff
@@ -1657,13 +1678,20 @@ def os_data():
...
+                        elif cpe.get('vendor', '').lower() in ['suse', 'opensuse']:
                             grains['os'] = "SUSE"
                             # openSUSE `osfullname` grain normalization
                             if os_release.get("NAME") == "openSUSE Leap":
                                 grains['osfullname'] = "Leap"
                             elif os_release.get("VERSION") == "Tumbleweed":
                                 grains['osfullname'] = os_release["VERSION"]
+                        # Override VERSION_ID, if CPE_NAME around
+                        if cpe.get('version'):
+                            grains['lsb_distrib_release'] = cpe['version']
```
meaksh: IIUC this would make the `lsb_distrib_release` grain to change from current `12.3` (VERSION_ID) to `sp3` (phase on CPE) on SLE12SP3. Is that correct?

In that case, I wonder if we want to make this change a part of a bugfix release.

salt/grains/core.py
```diff
@@ -1466,6 +1466,27 @@ def _parse_os_release(*os_release_files):
     return ret
 
 
+def _parse_cpe_name(cpe):
```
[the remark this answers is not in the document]
isbm: @meaksh no, I do not agree with this at this point. We should refactor the whole grains, and move also `_parse_lsb_release`, `_parse_os_release` etc from the same place. This should be a separate PR.

salt/grains/core.py
isbm: Fixed.

cachedout (approved):

salt/grains/core.py
```diff
@@ -1695,9 +1695,9 @@ def os_data():
                                 grains['osfullname'] = "Leap"
                             elif os_release.get("VERSION") == "Tumbleweed":
                                 grains['osfullname'] = os_release["VERSION"]
-                        # Override VERSION_ID, if CPE_NAME around
-                        if cpe.get('version'):
-                            grains['lsb_distrib_release'] = cpe['version']
+                            # Override VERSION_ID, if CPE_NAME around
+                            if cpe.get('version') and cpe.get('vendor') == 'opensuse':  # Keep VERSION_ID for SLES
```
aplanas: As this is now moved inside the `elif`, we can remove the `cpe.get('version') and` part, as will be or 'suse' or 'opensuse' already.

isbm: @aplanas no, because if there is no `version` due to broken CPE, the whole thing will still go wrong.

rallytime (approved):

PR #48001, opened 2018-06-06
Add salt-api on Windows platform

rares-pop:
salt-api is broken on Windows platform
due to pickling issues on the salt.loader.netapi
Workaround that by creating a runner class
that can be pickled.

Signed-off-by: Rares POP <rares.pop@ni.com>

### What does this PR do?
It adds/fixes salt-api on the Windows platform, using rest_cherrypy module

### Previous Behavior
salt-api wasn't available on Windows

### New Behavior
The user is able to run the salt-api daemon on windows, using rest_cherrypy module

### Tests written?

No

### Commits signed with GPG?

No

Please review [Salt's Contributing Guide](https://docs.saltstack.com/en/latest/topics/development/contributing.html) for best practices.

See GitHub's [page on GPG signing](https://help.github.com/articles/signing-commits-using-gpg/) for more information about signing commits with GPG.

PR #49949, opened 2018-10-09
Add salt-api in windows installer

rares-pop:
Add salt-api in Windows installer

### What does this PR do?
This added salt-api support on Windows:
https://github.com/saltstack/salt/pull/48001
This change adds it in the installer too.

### Previous Behavior
salt-api wasn't in the Windows installer

### Tests written?

No

### Commits signed with GPG?

Yes


cachedout:
Before we merge this, I would like to have a discussion about the level at which we intend to support this going forward. @rares-pop since you've been working on this, do you believe it's production ready? Do we have adequate test coverage? Or, should we mark this as an experimental feature for Neon?

twangboy (approved):

rares-pop:
@cachedout - I tested a very specific case, using the cherrypy backend with eauth: auto. That worked pretty good - I created jobs with no parameters, with parameters, targeting both single and multiple minions. I didn't do stress testing, and I ran no salt tests on windows.
More than that, what I tested was on 2018.3, and while debugging, I hit a path where it was looking for a hardcoded path for PAM that didn't work on Windows (that's why I went with eauth - auto, which is enough for our use-cases).
Most of the netapi salt integrations tests are around login. I can try them tomorrow and see what works and what not, but if pam doesn't work, the tests will render useless.

Before saying about the state (production vs experimental), I'd like to have some more stress tests:
 - firing one or two jobs every 0.5 seconds and have that running for few hours
 - firing hundreds/thousands of parallel jobs and see nothing gets corrupted

I can do that tomorrow.

cachedout:
@rares-pop Sounds like a good plan. I'll look forward to hearing the results of your testing tomorrow. Thank you for your quick response!

rares-pop:
@cachedout - sorry for the late reply. Yesterday is was all meetings..

Here is a summary of what I did, and let's choose together between 'production ready' and 'experimental'.
 - as I was saying, the configuration I tested with is kind of limited:
[config here](https://pastebin.com/2kyGP2ZH)
I didn't manage to get 'pam' working on windows, that's why the above configuration
 - I did test single function jobs, multi function jobs, with and without arguments targeted to both single and multiple minions. Everything is working.
 - I did create several stress tests, and I got up to numbers like ~10.000 jobs working (both local and local_async clients), and everything worked.
[code here](https://pastebin.com/2R3YiDpJ)
 - SSE works

My impressions:
 - feature/configuration-wise, it's experimental (not a lot of options work)
 - stability wise, is pretty rock solid

However - I've noticed a memory leak pulled down my whole system after ~10.000 jobs. I don't know yet if it's salt-api, salt-master or a salt-minion I was running on the windows machine. The second minion was ubuntu and had no problems whatsoever. It's a leak we saw on LinuxRT but never seen it reproduced with a vanilla version of salt. So if the windows behavior is reproducible, I'll open an issue.

rares-pop:
LE: the leak does reproduce on windows pretty easily (and on LinuxRT). This happens when they are registered to Windows masters. However, it doesn't happen on the ubuntu minion (even though it's registered to the same Windows master).
salt_minion_leak

dwoz:
We have some existing tests that historically have not passed here:

https://github.com/saltstack/salt/tree/develop/tests/integration/netapi/rest_cherrypy

rares-pop:
@cachedout, @dwoz  - I'm not sure if I'm requested to do any further work. Should I run the windows tests? Or something else?

dwoz:
@rares-pop  Sounds like we should mark this as experimental for now. I'll add the netapi tests to `whitelist.txt` so they start running.

dwoz (approved):

cachedout:
@rares-pop Sorry for being a bit late to the discussion. I agree with what you've written here. @twangboy and @dwoz where would you like to see the caveats for this functionality documented?

rallytime:
@dwoz and @twangboy bump ^ :)

twangboy:
@rares-pop Please add a note in the following location that salt-api using cherrypy is experimental and has limitations, ie: PAM Authentication:

https://github.com/saltstack/salt/blob/develop/salt/netapi/rest_cherrypy/app.py

Please add anything else you find applicable.

Additionally, I ran the scripts you posted on one of My VM's and didn't see the memory leak you're seeing.  I was working directly off one of our branches though... I'll try a few more things to see if I can replicate that.

rares-pop:
@twangboy - salt-api is experimental only on Windows, right? I mean, I assume it works on other platforms. Should I add a 'windows support' category, and add details there?

The memory leak I see running those scripts are reproducing on Oxygen. I haven't tried them on 'develop'.

rares-pop:
@twangboy - I only tried with py3. I never tried any salt stuff on py2 on windows.

@rallytime - I'll come up with something and then refine it as you guys review it. I glanced through that file but I didn't find a concrete block to populate right now, but I'll do it tomorrow morning (it's close to 11pm here :) ).

rallytime (approved):

rallytime:
@dwoz and @twangboy is this ready to go in now?

twangboy (approved):

twangboy:
@rallytime Looks good

PR #49950, opened 2018-10-09
Fix file patch tests on CentOS 6

rallytime:
### What does this PR do?
Fix the `integration.states.test_file.PatchTest` tests that are failing on CentOS 6.

Some logic was added in PR #49568 to skip the patch tests on patch whose version is < 2.6.1. However, CentOS 6 has patch `2.6` installed (note the `2.6` notation instead of something like `2.6.0`)  and the logic was failing in the check like so:
```
  File "/tmp/kitchen/testing/tests/integration/states/test_file.py", line 4067, in test_patch_directory
    self._check_patch_version('2.6.1')
  File "/tmp/kitchen/testing/tests/integration/states/test_file.py", line 3940, in _check_patch_version
    version = re.search(r'\d\.\d\.\d', self.run_function('cmd.run', ['patch --version'])).group(0)
AttributeError: 'NoneType' object has no attribute 'group'
```

This PR changes the logic to not fail when the version installed reports either 2 (`2.6`) or 3 (`2.6.1`) version "pieces" of the notation.

While fixing this problem, I also ran the tests to see if they would pass on the `2.6` version, rather than requiring `2.6.1` or newer. They all pass just fine, so I loosened the minimum version from `2.6.1` to simply `2.6`. 

Fixes https://github.com/saltstack/salt-jenkins/issues/1154

Refs https://github.com/saltstack/salt/pull/49568

### Previous Behavior
Tests would fail with a stacktrace on CentOS 6.

### New Behavior
Stacktrace is no longer present and the tests run successfully.

### Tests written?

Yes - fixes tests and enables them again on CentOS 6

### Commits signed with GPG?

Yes

### DIFF
--- a/tests/integration/states/test_file.py
+++ b/tests/integration/states/test_file.py
@@ -3935,11 +3935,10 @@ class PatchTest(ModuleCase, SaltReturnAssertsMixin):
         '''
         patch version check
         '''
-        if not salt.utils.path.which('patch'):
-            self.skipTest('patch is not installed')
-        version = re.search(r'\d\.\d\.\d', self.run_function('cmd.run', ['patch --version'])).group(0)
+        version = self.run_function('cmd.run', ['patch --version']).splitlines()[0]
+        version = version.split()[1]
         if _LooseVersion(version) < _LooseVersion(min_version):
-            self.skipTest('Mininum patch version required: {0}.'
+            self.skipTest('Minimum patch version required: {0}. '
                           'Patch version installed: {1}'.format(min_version, version))
 
     @classmethod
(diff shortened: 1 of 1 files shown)


cachedout:
But doesn't this also mean that the underlying functionality of Salt itself is broken on RH 6?

rallytime:
No, this changes the minimum version from `2.6.1` to `2.6`, which allows the tests to run now on CentOS 6. They wouldn't have run before, if the original `_check_patch_version` had worked correctly against a `2.6` patch version.

Ch3LL (approved):

PR #49951, opened 2018-10-09
Fix test `unit.modules.test_kubernetes` on Windows

twangboy:
### What does this PR do?
Fixes the `test_kubernetes` unit tests in Windows. It's an issue with the path that's being checked. It's looking in `/tmp/`  by default... this needs to be os agnostic, so we'll use the `temp` environment variable to get the first portion.

### Tests written?
Yes

### Commits signed with GPG?
Yes


rallytime (approved):
👍

rallytime:
@twangboy It appears the linux tests didn't like this change very much: https://jenkinsci.saltstack.com/job/pr-kitchen-ubuntu1604-py2/job/PR-49951/2/

Can you fix that up?

PR #49969, opened 2018-10-10
Introduce 'ignore_deprecation_warnings' to avoid logging deprecation warnings

meaksh:
### What does this PR do?
This PR introduces `ignore_deprecation_warnings` as a minion settings in order to avoid logging deprecation warnings messages coming a particular function that makes use of the `@with_deprecated` decorator. This way, we can avoid excessive logging on the minion files.

### Previous Behavior
```
2018-10-10 11:41:23,020 [salt.utils.decorators:544 ][WARNING ][21264] The function "module.run" is using its deprecated version and will expire in version "Sodium".
```

### New Behavior
Having the following in our `/etc/salt/minion`:
```
...
ignore_deprecation_warnings:
  - module.run
...
```

No more deprecation warning messages from `module.run`.

### Tests written?

No

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/utils/decorators/__init__.py
+++ b/salt/utils/decorators/__init__.py
@@ -423,7 +423,8 @@ class _IsDeprecated(_DeprecationDecorator):
                                                                     version_name=self._exp_version_name)]
                 if self._successor:
                     msg.append('Use successor "{successor}" instead.'.format(successor=self._successor))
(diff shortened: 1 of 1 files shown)


meaksh:
I'm closing this PR since we actually want to reconsider the strategy here! Sorry for the noise

PR #49978, opened 2018-10-10
remove the incomplete exclude list.

whytewolf:
using an exclude list that does not list every operating system leaves open the possibility that it would work on those not listed

### What does this PR do?
Minor doc change to inotify beacon to remove exclude list as it is incomplete. Leaving the original definition as just those operating systems that have itnofy kernel 

### Tests written?

No - Doc change

### Commits signed with GPG?

No


rallytime (approved):

rallytime:
Thanks for the clarification!

Issue #49971, opened 2018-10-10
lgpo state module Unable to find machine policy "Devices: Prevent users from installing printer drivers"


amalaguti:
### Description of Issue/Question
state module lgpo.set Unable to find machine policy "Devices: Prevent users from installing printer drivers"

### Setup
```
set_policy_enable:
  lgpo.set:
    - computer_policy:
      "Devices: Prevent users from installing printer drivers": 'Enabled'
```
```
local:
----------
          ID: prevent_users_installing_printer_drivers
    Function: lgpo.set
      Result: False
     Comment: Unable to find Machine policy Devices: Prevent users from installing printer drivers
     Started: 12:56:38.350000
    Duration: 2968.0 ms
     Changes:
```

### Steps to Reproduce Issue
(windows): salt-call state.sls "state"

### Versions Report
salt 2018.3.2
Windows 2016 Server

Ch3LL:
ping @saltstack/team-windows any ideas here?

twangboy:
Some policies are defined in the ADMX/ADML files on the machine... others are not. It looks like security policies are not. Each of those need to be added to the dictionary inside the LGPO module to provide support for those. I added this one in the above PR.

lomeroe:
most (if not all) of the missing static policies were added to develop in #47278, it just wasn't ever backported (believe all the previously reported issues on missing policies had been tagged as features)

PR #49980, opened 2018-10-10
Add support for AddPrinterDrivers

twangboy:
### What does this PR do?
Adds support for the "Devices: Prevent users from installing printer drivers" Group Policy Object.

https://github.com/saltstack/salt/issues/49971

### Previous Behavior
```
[ERROR   ]  Unable to find Machine policy Devices: Prevent users from installing printer drivers
local:
----------
          ID: prevent_users_installing_printer_drivers
    Function: lgpo.set
      Result: False
     Comment: Unable to find Machine policy Devices: Prevent users from installing printer drivers
     Started: 20:01:55.460000
    Duration: 672.0 ms
     Changes:
```

### New Behavior
Successfully sets the Group Policy

### Tests written?
No

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/modules/win_lgpo.py
+++ b/salt/modules/win_lgpo.py
@@ -1437,6 +1437,21 @@ class _policy_info(object):
                         },
                         'Transform': self.enabled_one_disabled_zero_transform,
                     },
+                    'AddPrinterDrivers': {
+                        'Policy': 'Devices: Prevent users from installing '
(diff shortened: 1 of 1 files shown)


cachedout:
Why is this going into a point release?

twangboy:
It's not changing behavior... It just adding another policy that can be set. SecDev needs it.

PR #49981, opened 2018-10-10
Improved error logging for win_task.create_task_from_xml

Lesvek:
### What does this PR do?
Instead of only returning True or False when create_task_from_xml fail to add the scheduled task, the actual error message is returned to the user. If the error is unknown, the hex error code is returned. That way, it's easier to search for as this is the format Microsoft is using in their documentation.

### Previous Behavior
Previously, when there was an error, the module only returned False, only logging the error message in the debug logs.

### New Behavior
A proper exception containing the error message is now returned to the user when there is an error. New code will have to catch the exception.

### Tests written?
I added 2 new tests.

### Commits signed with GPG?
The commit was not signe with GPG.

### DIFF
--- a/salt/modules/win_task.py
+++ b/salt/modules/win_task.py
@@ -566,8 +566,8 @@ def create_task_from_xml(name,
         the task to run whether the user is logged in or not, but is currently
         not working.
 
-    :return: True if successful, False if unsuccessful
-    :rtype: bool
+    :return: True if successful, False if unsuccessful, A string with the error message if there is an error
+    :rtype: bool or str
 
     CLI Example:
 
@@ -609,6 +609,7 @@ def create_task_from_xml(name,
                 logon_type = TASK_LOGON_INTERACTIVE_TOKEN
     else:
         password = None
+        logon_type = TASK_LOGON_NONE
 
     # Save the task
     try:
--- a/salt/modules/win_task.py
+++ b/salt/modules/win_task.py
@@ -17,7 +17,7 @@ from datetime import datetime
 
 # Import Salt libs
 import salt.utils.platform
-import salt.exceptions
+from salt.exceptions import ArgumentValueError, CommandExecutionError
 
 # Import 3rd-party libraries
 try:
@@ -583,7 +583,7 @@ def create_task_from_xml(name,
         return '{0} already exists'.format(name)
 
     if not xml_text and not xml_path:
-        raise salt.exceptions.ArgumentValueError('Must specify either xml_text or xml_path')
+        raise ArgumentValueError('Must specify either xml_text or xml_path')
 
     # Create the task service object
     pythoncom.CoInitialize()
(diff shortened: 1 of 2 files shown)


twangboy (approved):

salt/modules/win_task.py
```diff
@@ -621,21 +622,38 @@ def create_task_from_xml(name,
...
-            failure_code = fc[exc[5]]
+            failure_code = fc[error_code]
         except KeyError:
-            failure_code = 'Unknown Failure: {0}'.format(error)
-
-        log.debug('Failed to create task: %s', failure_code)
+            failure_code = 'Unknown Failure: {0}'.format(error_code)
+        finally:
+            log.debug('Failed to create task: %s', failure_code)
+        return failure_code
```
twangboy: We should probably raise a CommandExecutionError here instead. If the Return type is Bool, it should always be Bool. It shouldn't return 2 different types.

twangboy (changes requested):

salt/modules/win_task.py
```diff
@@ -566,8 +566,8 @@ def create_task_from_xml(name,
         the task to run whether the user is logged in or not, but is currently
         not working.
 
-    :return: True if successful, False if unsuccessful
-    :rtype: bool
+    :return: True if successful, False if unsuccessful, A string with the error message if there is an error
+    :rtype: bool or str
```
twangboy: Add a `:raises: CommandExecutionError` for completenes

salt/modules/win_task.py
Lesvek: Oh, nice. I didn't know it existed.

twangboy (changes requested):

tests/integration/modules/test_win_task.py
```diff
@@ -0,0 +1,76 @@
+# -*- coding: utf-8 -*-
+from tests.support.case import ModuleCase
+from tests.support.unit import skipIf
+from tests.support.helpers import destructiveTest
+
+import salt.utils.platform
```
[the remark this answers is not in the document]
twangboy: And you need two blank lines between the imports and the class (lint)

tests/integration/modules/test_win_task.py
```diff
@@ -0,0 +1,76 @@
+# -*- coding: utf-8 -*-
```
twangboy: You probably need to add this to fix a lint error:
```
from __future__ import absolute_import, unicode_literals
```

tests/integration/modules/test_win_task.py
```diff
@@ -0,0 +1,76 @@
...
+        all_tasks = self.run_function('task.list_tasks')
+        self.assertIn('foo', all_tasks)
+
+    @destructiveTest
+    def test_adding_task_with_invalid_xml(self):
+        '''
+        Test adding a task using xml
+        '''
+        xml_text = r"""<Malformed"""
+        with self.assertRaises(salt.exceptions.CommandExecutionError):
```
twangboy:
```
with self.assertRaises(CommandExecutionError):
```

twangboy (approved):

twangboy:
@Lesvek Thanks for getting all those things fixed and thanks for the contribution! Looks good now.

PR #49984, opened 2018-10-10
Use prune option in Pygit2 provider when fetching

chotea:
Pygit2 version 0.26.2 added support for pruning when fetching.
In this way Pygit2 provider will no longer need to leverage
git command line utility for pruning the remote.

Signed-off-by: Cristian Hotea <cristian.hotea@ni.com>

### Tests written?

No

### Commits signed with GPG?

No

### DIFF
--- a/salt/utils/gitfs.py
+++ b/salt/utils/gitfs.py
@@ -1583,11 +1583,19 @@ class Pygit2(GitProvider):
         '''
         Clean stale local refs so they don't appear as fileserver environments
         '''
+        try:
+            if pygit2.GIT_FETCH_PRUNE:
+                # Don't need to clean anything, pygit2 can do it by itself
+                return []
+        except AttributeError:
+            # However, only in 0.26.2 and newer
+            pass
         if self.credentials is not None:
             log.debug(
-                'pygit2 does not support detecting stale refs for '
-                'authenticated remotes, saltenvs will not reflect '
-                'branches/tags removed from remote \'%s\'', self.id
+                'The installed version of pygit2 (%s) does not support '
+                'detecting stale refs for authenticated remotes, saltenvs '
+                'will not reflect branches/tags removed from remote \'%s\'',
+                PYGIT2_VERSION, self.id
             )
             return []
         return super(Pygit2, self).clean_stale_refs()
@@ -1699,6 +1707,11 @@ class Pygit2(GitProvider):
         else:
             if self.credentials is not None:
                 origin.credentials = self.credentials
+        try:
+            fetch_kwargs['prune'] = pygit2.GIT_FETCH_PRUNE
+        except AttributeError:
+            # pruning only available in pygit2 >= 0.26.2
+            pass
         try:
             fetch_results = origin.fetch(**fetch_kwargs)
         except GitError as exc:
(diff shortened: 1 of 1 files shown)


salt/utils/gitfs.py
```diff
@@ -1699,6 +1701,8 @@ def _fetch(self):
         else:
             if self.credentials is not None:
                 origin.credentials = self.credentials
+        if HAS_PYGIT2_FETCH_PRUNE:
+            fetch_kwargs['prune'] = pygit2.GIT_FETCH_PRUNE
```
terminalmage: Why set a global for this? EAFP.

```python
try:
    fetch_kwargs['prune'] = pygit2.GIT_FETCH_PRUNE
except AttributeError:
    # pruning only available in pygit2 >= 0.26.2
    pass
```

salt/utils/gitfs.py
```diff
@@ -2551,7 +2557,7 @@ def _recommend():
                     LIBGIT2_VERSION
                 )
             )
-        if not salt.utils.path.which('git'):
+        if not HAS_PYGIT2_FETCH_PRUNE and not salt.utils.path.which('git'):
```
terminalmage: How about this?

```python
if not getattr(pygit2, 'GIT_FETCH_PRUNE', False) \
        and not salt.utils.path.which('git'):
```

salt/utils/gitfs.py
```diff
@@ -1740,7 +1744,9 @@ def _fetch(self):
         else:
             log.debug('%s remote \'%s\' is up-to-date', self.role, self.id)
         refs_post = self.repo.listall_references()
-        cleaned = self.clean_stale_refs(local_refs=refs_post)
+        cleaned = []
+        if not HAS_PYGIT2_FETCH_PRUNE:
```
terminalmage: Instead of this defining `cleaned` twice, who about reverting this particular change and doing this?

```diff
diff --git a/salt/utils/gitfs.py b/salt/utils/gitfs.py
index 286657c72c..9e499b52a3 100644
--- a/salt/utils/gitfs.py
+++ b/salt/utils/gitfs.py
@@ -1583,11 +1583,19 @@ class Pygit2(GitProvider):
         '''
         Clean stale local refs so they don't appear as fileserver environments
         '''
+        try:
+            if pygit2.GIT_FETCH_PRUNE:
+                # Don't need to clean anything, pygit2 can do it by itself
+                return []
+        except AttributeError:
+            # However, only in 0.26.2 and newer
+            pass
         if self.credentials is not None:
             log.debug(
-                'pygit2 does not support detecting stale refs for '
-                'authenticated remotes, saltenvs will not reflect '
-                'branches/tags removed from remote \'%s\'', self.id
+                'The installed version of pygit2 (%s) does not support '
+                'detecting stale refs for authenticated remotes, saltenvs '
+                'will not reflect branches/tags removed from remote \'%s\'',
+                PYGIT2_VERSION, self.id
             )
             return []
         return super(Pygit2, self).clean_stale_refs()
```

terminalmage (changes requested):
I like this, but I'm trying to move away from version checks when I can, the reason being that if the versioning ends up being determined via some sort of `git describe` logic (for those installing pygit2 from the upstream project's `master` branch), it may interfere with this kind of version check. Here are some recommended changes that can help us avoid explicit version checking and instead rely on the attributes we know should be available.

Thanks very much!

cachedout:

terminalmage (approved):

PR #49987, opened 2018-10-10
Make Pillar no longer munge file_roots

terminalmage:
Pillar and the local fileclient used to use the same class, but this changed a few years ago with the introduction of FSClient (which enabled non-roots fileserver backends in masterless Salt). Since Pillar is the only thing that uses this client, there is no longer a reason to make Pillar modify file_roots.

Fixes https://github.com/saltstack/salt/issues/47539


gtmanfred (approved):
Confirmed fixes #47539

terminalmage:
Tests should be fixed with latest commit.

rallytime (approved):

PR #49989, opened 2018-10-10
Increase centos7-py2 kitchen pr timeout to 8 hours

KaiSforza:
The centos7-py2 tests have been consistently failing to finish before
the timeout. I am increasing the timeout for these tests to 8 hours in
order to see if we can finish a few of them. They should get much closer
to finishing as the branch tests have been finishing at around the 6:30
or 7:00 mark sometimes.

### DIFF
--- a/.ci/kitchen-centos7-py2
+++ b/.ci/kitchen-centos7-py2
@@ -1,4 +1,4 @@
-timeout(time: 6, unit: 'HOURS') {
+timeout(time: 8, unit: 'HOURS') {
     node('kitchen-slave') {
         timestamps {
             ansiColor('xterm') {


rallytime (approved):

dwoz:
@KaiSforza Looks like we might want to do this for the windows tests.. I've seen a few 2018.3 PY3 windows test suites timeout at 6 hours.
https://jenkinsci.saltstack.com/job/2018.3/view/Python3/job/salt-windows-2016-py3/11/

damon-atkins:
If its a load issue have you consider limiting the number of tests running.
Or even just fire off the lint only, and only run the others if the lint passes,  that way only more tests would run if they have half a chance of working, vs not working at all (if the lint fails)

damon-atkins:
Or even just fire off the lint only (28 min?), and only run the others (7 hours?) if the lint passes,  that way only more tests would run if they have half a chance of working, vs not working at all (if the lint fails)

If the lint fails most people would fix it, and then commit, and then bunch of more jobs are started, so a single PR may have multiple Jenkins slaves active.
For example https://jenkinsci.saltstack.com/view/Pull%20Requests/job/pr-kitchen-windows2016-py3/view/change-requests/job/PR-49995/ Has two run starting at the same time.
- 2 Oct 11, 2018 9:34 AM
- 1 Oct 11, 2018 9:03 AM

damon-atkins:
ping @dubb-b

Issue #38751, opened 2017-01-16
service.enabled does not enable service in chroot environment for systemd


geraldhansen:
service.enabled just say service is already enabled, but the symlink in **_/etc/systemd/system/multi-user.target.wants_**/ is still missing.
I prepare a new system installation in a chroot environment which is a physical partition and can be enabled as the initial boot partition for the next reboot (Debian 8/Ubuntu 16.10).

### Setup
In a chroot I run following state file test.sls:
```
install-apache:
    pkg.installed:
        - pkgs:
            - apache2

apache-service:
    service.enabled:
        - name: apache2
        - require:
            - pkg: install-apache

```
### Steps to Reproduce Issue

`salt-call state.sls test`

```
[INFO    ] Executing state service.enabled for apache2
[INFO    ] Executing command ['systemctl', 'status', 'apache2.service', '-n', '0'] in directory '/home/gerald'
[DEBUG   ] output: * apache2.service - LSB: Apache2 web server
   Loaded: loaded (/etc/init.d/apache2; generated; vendor preset: enabled)
  Drop-In: /lib/systemd/system/apache2.service.d
           `-apache2-systemd.conf
   Active: inactive (dead) since Mon 2017-01-16 12:28:35 CET; 10min ago
     Docs: man:systemd-sysv-generator(8)
[INFO    ] Executing command ['systemctl', 'is-enabled', 'apache2.service'] in directory '/home/gerald'
[DEBUG   ] output: apache2.service is not a native service, redirecting to systemd-sysv-install.
Executing: /lib/systemd/systemd-sysv-install is-enabled apache2
enabled
[INFO    ] Service apache2 is already enabled, and is in the desired state
[INFO    ] Completed state [apache2] at time 12:38:49.468597 duration_in_ms=18.652
```

### Versions Report

```
salt-call --versions-report
Salt Version:
           Salt: 2016.3.1
 
Dependency Versions:
           cffi: 1.7.0
       cherrypy: 3.5.0
       dateutil: 2.4.2
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.8
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.4.8
   mysql-python: Not Installed
      pycparser: 2.14
       pycrypto: 2.6.1
         pygit2: Not Installed
[...]

PR #50009, opened 2018-10-11
add skip_verify for service.enabled and .disabled

jigish:
### What does this PR do?

Adds the skip_verify option for service.enabled and service.disabled.

When running service.enabled or service.disabled, salt verifies that the service is 'available'. This option allows the user to skip that verification for cases that it doesn't make sense. For example, while running salt in a chroot environment (like packer's salt-masterless provisioner while using the amazon-chroot builder), `systemctl status` fails as it is invalid. Without skip_verify, this would cause the service.enabled state to fail as well.

#38751 -- Doesn't fix, just supplies a workaround.

### Tests written?

No

### Commits signed with GPG?

Yes


terminalmage (commented):
I'm curious how this would help, if the service name is invalid, wouldn't enabling/disabling it also fail?

```
% sudo systemctl enable foo
Failed to enable unit: Unit file foo.service does not exist.
% echo $?
1
% sudo systemctl disable foo
Failed to disable unit: Unit file foo.service does not exist.
% echo $?
1
```

It seems to me like this just kicks the can to a further point where it will also fail.

jigish:
You're absolutely right for cases in which the service name is invalid, which is why the option is defaulted to `False`.

The `skip_verify` option I've added is intended to be used in cases where the available check itself is not applicable and/or invalid but the service name *is* valid. In chroot environments, `systemctl status` calls fail even for valid service names as systemd seems to consider it an invalid use case. In those cases you could use `skip_verify` to skip that and directly `systemctl enable` which will work just fine. This is especially important while using salt to provision an AMI via amazon-chroot.

That being said, I've implemented this as a stopgap. It may be worth modifying the systemd module itself to not run `status` in chroot environments. I just have a lot less experience with the other service modules so I'm not sure if the same change is needed in them.

terminalmage (changes requested):
I don't like that you've inserted this `skip_verify` argument such that it is required to be passed every time the function is invoked. If we expect the vast majority of the invocations to be `False`, then the `skip_verify` argument should be moved until after `result` and should have a default value of `False`. Then we don't have to pass a `False` in most of the calls to `_enable()` and `_disable()`, and in the few cases where we are passing through the value from the state, we can simply pass `skip_verify=skip_verify`.

jigish:
@terminalmage - Just made the requested changes. Thanks for taking a look!

terminalmage (approved):

PR #50013, opened 2018-10-12
Passed kwargs to db_exists in db_remove method

Giandom:
### What does this PR do?
Add kwargs to the db_exists function call

The function db_remove fails because the call to the dbexists function needs kwargs to connect to the db if you're passing all connection parameters from cli.

### Previous Behavior
db_remove function fails passing connection parameters from cli

### New Behavior
db_remove works

### Tests written?
No

### Commits signed with GPG?

No


cachedout (approved):

cachedout:
Thank you, @Giandom

PR #50018, opened 2018-10-12
Add multi-file support and globbing to the filetree

isbm:
### What does this PR do?

- Feature for support conf
- Prevents crash, if file does not exists :fearful: 

### Previous Behavior

Before `filetree` collector could take only _one_ file or the entire directory. In order to take several files you would need to repeat it, like this:

```yaml
system.log:
   - filetree:
      info: Collect syslog
         - /var/log/syslog
   - filetree:
      info: Collect messages log
         - /var/log/messages
```

### New Behavior

Added support for:

- Several files at once
- Patterns (yay!)

Example
```yaml
boot.log:
   - filetree:
      info: Collect boot logs
         - /var/log/boot.*

system.log:
   - filetree:
      info: Collect system logs
         - /var/log/syslog
         - /var/log/messages
```

### Tests written?

No (should pass existing)

### DIFF
--- a/salt/cli/support/intfunc.py
+++ b/salt/cli/support/intfunc.py
@@ -6,6 +6,7 @@ Internal functions.
 
 from __future__ import absolute_import, print_function, unicode_literals
 import os
+import glob
 from salt.cli.support.console import MessagesOutput
 import salt.utils.files
 
@@ -21,9 +22,15 @@ def filetree(collector, *paths):
     :param path: File or directory
     :return:
     '''
+    _paths = []
+    # Unglob
     for path in paths:
+        _paths += glob.glob(path)
+    for path in set(_paths):
         if not path:
             out.error('Path not defined', ident=2)
+        elif not os.path.exists(path):
+            out.warning('Path {} does not exists'.format(path))
         else:
             # The filehandler needs to be explicitly passed here, so PyLint needs to accept that.
             # pylint: disable=W8470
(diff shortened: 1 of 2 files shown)


salt/cli/support/intfunc.py
```diff
@@ -6,37 +6,48 @@
...
-                out.put('Add {}'.format(filename), indent=2)
-                collector.add(filename)
-                collector.link(title=path, path=file_ref)
-            except Exception as err:
-                out.error(err, ident=4)
-        # pylint: enable=W8470
+    _paths = []
+    # Unglob
+    for path in paths:
+        _paths += glob.glob(path)
```
gtmanfred: Would this be better to just do a .extend()?

gtmanfred (approved):

salt/cli/support/intfunc.py
isbm: In this case there is no much difference, but let `timeit` rolls the dice:

```
$ python -m timeit 'a = []; a.extend([1, 2, 3])'
10000000 loops, best of 3: 0.173 usec per loop
$ python -m timeit 'a = []; a.extend([1, 2, 3])'
10000000 loops, best of 3: 0.172 usec per loop
$ python -m timeit 'a = []; a.extend([1, 2, 3])'
10000000 loops, best of 3: 0.172 usec per loop

$ python -m timeit 'a = []; a += [1, 2, 3]'
10000000 loops, best of 3: 0.136 usec per loop
$ python -m timeit 'a = []; a += [1, 2, 3]'
10000000 loops, best of 3: 0.135 usec per loop
$ python -m timeit 'a = []; a += [1, 2, 3]'
10000000 loops, best of 3: 0.136 usec per loop
```

Seems `+=` wins here `¯\_(ツ)_/¯`.

meaksh (approved):

cachedout (approved):

PR #50026, opened 2018-10-13
Jenkins CI Lint Improvements

damon-atkins:
### What does this PR do?
Improvements in .ci/lint
* Change the output extension from .xml to .log (plain text) so Jenkins sends the correct mime type to browsers.
* Remove the colour escape codes from the logs, so Artifact  does not contain them, but the console output record by Jenkins does. 
* Store an Artifact of which files have changed, and also the files which have been deleted, for debugging/verifying the files that have been reviewed by pylint.
*Remove the need to use find to scan the file system and call git multiple times. (small decrease in overhead)
*Only run a stage if required.

Some general improvements.

### DIFF
--- a/.ci/lint
+++ b/.ci/lint
@@ -24,11 +24,10 @@ pipeline {
                 sh '''
                 # Need -M to detect renames otherwise they are reported as Delete and Add, need -C to detect copies, -C includes -M
                 # -M is on by default in git 2.9+
-                git diff --name-status -C "origin/$CHANGE_TARGET" "origin/$BRANCH_NAME" > file-list-status.log
-                # '--diff-filter=ACMRTUXB' can be changed to '--diff-filter=d' when git version >=1.8.5 is available
-                # on renames --name-only outputs the destination only
-                git diff --name-only '--diff-filter=ACMRTUXB' -C "origin/$CHANGE_TARGET" "origin/$BRANCH_NAME" > file-list-changed.log
-                git diff --name-only '--diff-filter=D' -C "origin/$CHANGE_TARGET" "origin/$BRANCH_NAME" > file-list-deleted.log
+                git diff --name-status -l99999 -C "origin/$CHANGE_TARGET" "origin/$BRANCH_NAME" > file-list-status.log
+                # the -l increase the search limit, lets use awk so we do not need to repeat the search above.
+                gawk 'BEGIN {FS="\\t"} {if ($1 != "D") {print $NF}}' file-list-status.log > file-list-changed.log
+                gawk 'BEGIN {FS="\\t"} {if ($1 == "D") {print $NF}}' file-list-status.log > file-list-deleted.log
                 touch pylint-report-salt.log pylint-report-tests.log
                 eval "$(pyenv init -)"
                 pyenv --version
--- a/.ci/lint
+++ b/.ci/lint
@@ -22,6 +22,14 @@ pipeline {
         stage('setup') {
             steps {
                 sh '''
+                # Need -M to detect renames otherwise they are reported as Delete and Add, need -C to detect copies, -C includes -M
+                # -M is on by default in git 2.9+
+                git diff --name-status -C "origin/$CHANGE_TARGET" "origin/$BRANCH_NAME" > file-list-status.log
+                # '--diff-filter=ACMRTUXB' can be changed to '--diff-filter=d' when git version >=1.8.5 is available
+                # on renames --name-only outputs the destination only
+                git diff --name-only '--diff-filter=ACMRTUXB' -C "origin/$CHANGE_TARGET" "origin/$BRANCH_NAME" > file-list-changed.log
+                git diff --name-only '--diff-filter=D' -C "origin/$CHANGE_TARGET" "origin/$BRANCH_NAME" > file-list-deleted.log
+                touch pylint-report-salt.log pylint-report-tests.log
                 eval "$(pyenv init -)"
                 pyenv --version
                 pyenv install --skip-existing 2.7.14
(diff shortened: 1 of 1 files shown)


damon-atkins:
Best I could do is a syntax check to check if it will work.

KaiSforza (changes requested):

dubb-b:
@damon-atkins I like having all three log files archived. We could patch the server to allow/send the correct mime type. https://issues.jenkins-ci.org/browse/JENKINS-3803

What I would like to see here before we merge is some better testing ran against it.

@KaiSforza Can you please test this out on a few PR's by doing a replay or setting up some sort of test?

.ci/lint
```diff
@@ -22,6 +22,10 @@ pipeline {
         stage('setup') {
             steps {
                 sh '''
+                git diff --name-status "origin/$CHANGE_TARGET" "origin/$BRANCH_NAME" > file-list-status.log
+                git diff --name-only '--diff-filter=ACMRTUXB' "origin/$CHANGE_TARGET" "origin/$BRANCH_NAME" > file-list-changed.log
```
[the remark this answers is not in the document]
dubb-b: The line below should cover that correct? He is doing 2 logs 1 for status, 1 for changes, and one for deleted files.

KaiSforza:
Yep I'll get on it

.ci/lint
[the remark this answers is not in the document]
KaiSforza: It should, but I would prefer using the not-operator so if anything changes, we'll be fine.

damon-atkins:
Change the names to *.log as this is configured https://github.com/jenkinsci/jenkins/blob/master/war/src/main/webapp/WEB-INF/web.xml#L225

The file contents are not xml.

damon-atkins:
Should be right now. I am not a trusted user in Jenkins so its still running the old ci/lint.

.ci/lint
```diff
@@ -30,40 +38,38 @@ pipeline {
...
                         eval "$(pyenv init - --no-rehash)"
-                        _FILES="$(find salt/ -name "*.py" -exec git diff --name-only "origin/$CHANGE_TARGET" "origin/$BRANCH_NAME" {} +)"
-                        _FILES="$_FILES $(git diff --name-only "origin/$CHANGE_TARGET" "origin/$BRANCH_NAME" setup.py)"
-                        if [[ -z ${_FILES} ]]; then
-                            echo "No pylint run, no changes found in the files"
-                            echo "empty" pylint-reports.xml
-                        else
-                            tox -e pylint-salt ${_FILES} | tee pylint-report.xml
-                        fi
+                        egrep -Ei '^salt/.*\\.py$|^setup\\.py$' file-list-changed.log | xargs -r '--delimiter=\\n' tox -e pylint-salt | tee pylint-report-salt.log
```
KaiSforza: not `egrep -E`, just `grep -E`.

.ci/lint
```diff
@@ -30,40 +38,38 @@ pipeline {
...
                         sh '''
                         eval "$(pyenv init - --no-rehash)"
-                        _FILES="$(find tests/ -name "*.py" -exec git diff --name-only "origin/$CHANGE_TARGET" "origin/$BRANCH_NAME" setup.py {} +)"
-                        if [[ -z ${_FILES} ]]; then
-                            echo "No pylint run, no changes found in the files"
-                            touch pylint-report-tests.xml
-                        else
-                            tox -e pylint-tests ${_FILES} | tee pylint-report-tests.xml
-                        fi
+                        egrep -Ei '^tests/.*\\.py$' file-list-changed.log | xargs -r '--delimiter=\\n' tox -e pylint-tests | tee pylint-report-tests.log
```
KaiSforza: Same as above.

.ci/lint
```diff
@@ -22,6 +22,14 @@ pipeline {
         stage('setup') {
             steps {
                 sh '''
+                # Need -M to detect renames otherwise they are reported as Delete and Add, need -C to detect copies, -C includes -M
+                # -M is on by default in git 2.9+
+                git diff --name-status -C "origin/$CHANGE_TARGET" "origin/$BRANCH_NAME" > file-list-status.log
+                # '--diff-filter=ACMRTUXB' can be changed to '--diff-filter=d' when git version >=1.8.5 is available
```
KaiSforza: It actually saddens me that we're using a version of git that old, but oh well. :(

KaiSforza (changes requested):

KaiSforza:
I'd love these to be squashed down to one or two commits, honestly.

.ci/lint
```diff
@@ -47,7 +47,9 @@ pipeline {
                     steps {
                         sh '''
                         eval "$(pyenv init - --no-rehash)"
-                        egrep -i '^salt/.*\\.py$|^setup\\.py$' file-list-changed.log | xargs -r '--delimiter=\\n' tox -e pylint-salt | tee pylint-report-salt.log
+                        egrep -Ei '^salt/.*\\.py$|^setup\\.py$' file-list-changed.log | xargs -r '--delimiter=\\n' tox -e pylint-salt | tee pylint-report-salt.log
+                        # remove color escape coding
+                        sed -ri 's/\x1B\[([0-9]{1,2}(;[0-9]{1,2})?)?[m|K]//g' pylint-report-salt.log
```
KaiSforza: okay explain to me what this is actually needed for? the colors are automatically removed when pylint is piped to something, unless something is getting through.

damon-atkins: Save the current xml version and you will see them.   I suspect PY_COLOR forces it.   And is done so, so the console out put looks nice in Jenkins

.ci/lint
```diff
@@ -30,40 +38,38 @@ pipeline {
...
                         eval "$(pyenv init - --no-rehash)"
-                        _FILES="$(find salt/ -name "*.py" -exec git diff --name-only "origin/$CHANGE_TARGET" "origin/$BRANCH_NAME" {} +)"
-                        _FILES="$_FILES $(git diff --name-only "origin/$CHANGE_TARGET" "origin/$BRANCH_NAME" setup.py)"
-                        if [[ -z ${_FILES} ]]; then
-                            echo "No pylint run, no changes found in the files"
-                            echo "empty" pylint-reports.xml
-                        else
-                            tox -e pylint-salt ${_FILES} | tee pylint-report.xml
-                        fi
+                        egrep -E '^salt/.*\\.py$|^setup\\.py$' file-list-changed.log | xargs -r '--delimiter=\\n' tox -e pylint-salt | tee pylint-report-salt.log
```
KaiSforza: We're still doing `egrep` here, please remove that first `e`, it should be just `grep -E`.

.ci/lint
```diff
@@ -30,40 +38,38 @@ pipeline {
...
                         sh '''
                         eval "$(pyenv init - --no-rehash)"
-                        _FILES="$(find tests/ -name "*.py" -exec git diff --name-only "origin/$CHANGE_TARGET" "origin/$BRANCH_NAME" setup.py {} +)"
-                        if [[ -z ${_FILES} ]]; then
-                            echo "No pylint run, no changes found in the files"
-                            touch pylint-report-tests.xml
-                        else
-                            tox -e pylint-tests ${_FILES} | tee pylint-report-tests.xml
-                        fi
+                        egrep -E '^tests/.*\\.py$' file-list-changed.log | xargs -r '--delimiter=\\n' tox -e pylint-tests | tee pylint-report-tests.log
```
KaiSforza: Same as above.

KaiSforza (changes requested):

damon-atkins:
FYI  `egrep` is more common use than `grep -E`   as not all platforms support `grep -E`  i.e. Non-Linux/Non-GNU grep, but 99% of platforms support `egrep` which is the same as `egrep -E`

damon-atkins:
All updates ask for completed.

KaiSforza (approved):

damon-atkins:
Current Open Pull Requests against 2017.7 are
* 50128 Aug 19 Workaround for py2 builtin, =<3.3 imp and >=3.4 libimport quirks (Has Passed Lint Test)
* 46713 Mar 27 Don't refresh modules twice per sync or refresh ops
* 45839 Feb 2 Don't call close on singletons

So if this was merged into 2017.7 and issues occurred it would be manageable.

dubb-b:
@damon-atkins I ran this against a few PR's

I didn't dive too much into the reasoning. But there seems to be some discrepancies. 

https://jenkinsci.saltstack.com/view/Pull%20Requests/job/pr-lint/view/change-requests/job/PR-50124/

Notice #5 and #6 

5 is a replay with the new suggested changes and 6 is the old linting pipeline code. I don't have time to dig in at the moment, although it may give you an idea as to what's wrong.

damon-atkins:
I will have a look at then `when` clause, should be able to emulate it in any jenkins server, see why its not returning `true`.

damon-atkins:
@dubb-b that should do the job.

damon-atkins:
Go Go Jenkins!

dubb-b:
@damon-atkins I ran it against this: https://jenkinsci.saltstack.com/view/Pull%20Requests/job/pr-lint/view/change-requests/job/PR-50152/2/

Build #1 was successful and not it fails. Looking through the logs, it seems to have found 3 warnings. Can you validate that is what you'd expect. It looks right to me.

damon-atkins:
Looks Good.

damon-atkins:
So with all the recent changes to lint, the next step is only allow the other tests when the lint succeeds. This will reduce the testing load as per the example in https://github.com/saltstack/salt/pull/50026#issuecomment-431342611

dubb-b:
That is definitely on the roadmap. Not sure as to when we will get it in. But thanks for the work here. Much appreciated.

dubb-b (approved):
We have tested this on a number of PR's and it looks to be stable and working as it should.

PR #50028, opened 2018-10-13
Fix #50015, Windows grain defaults to physical as unix grain

### LINKED ISSUE #50015: Virtual grain in Windows
Hi there, we check if our hosts are virtual as such in our pillars:

``{% elif 'host' in grains['is_virtual'] %}``

and we have a problem on Windows where a physical host won't have a grain key for virtual and jinja errors out, because in grains/core.py the windows grain defaults to an empty dictionary rather than ''physical":

grains/core.py
```
def _windows_virtual(osdata):
    '''
    Returns what type of virtual hardware is under the hood, kvm or physical
    '''
    # Provides:
    #   virtual
    #   virtual_subtype
    grains = dict()
    if osdata['kernel'] != 'Windows':
        return grains

def _virtual(osdata):
    '''
    Returns what type of virtual hardware is under the hood, kvm or physical
    '''
    # This is going to be a monster, if you are running a vm you can test this
    # grain with please submit patches!
    # Provides:
    #   virtual
    #   virtual_subtype
    **grains = {'virtual': 'physical'}**
```
We worked around that creating a custom grain 'is_virtual' and wrapping around virtual as such:

```
import salt.config
import salt.loader
import platform
import os

def is_virtual():
    ## we put the only necessary dummy keys for loader.py
    ## if we used an actual config file, it would try to retrieve itself and enter an endless loop
    __opts__ = {}
    __opts__['cachedir'] = 'dummy'
    __opts__['extension_modules'] = 'dummy'
    ## Load core grains only
    __grains__ = salt.loader.grains(__opts__)
    try:
        virtual = __grains__['virtual']
    except KeyError, e:
        grains = {'is_virtual': 'physical'}
        return grains
    grains = {'is_virtual': virtual}
    return grains
```
But shouldn't the windows "virtual" default to "pyhisical" after the os check like this?

```
def _windows_virtual(osdata):
    '''
    Returns what type of virtual hardware is under the hood, kvm or physical
    '''
    # Provides:
    #   virtual
    #   virtual_subtype
    grains = dict()
    if osdata['kernel'] != 'Windows':
        return grains
    grains = {'virtual': 'physical'}
```
We are on 2017.7.7 but I can see the code is the same in the newer versions,

thanks

marco-kusa:
### What does this PR do?
Changes Windows "virtual" grain to default to physical. 

Issue #50015 
### Previous Behavior
No key for virtual is currently being returned and it causes errors if the grain is used in a multi-platform environment that includes windows.
### New Behavior
The Windows "virtual" assumes phyisical unless a specific type of vm is detected, matching the unix grain behavior.

### DIFF
--- a/salt/grains/core.py
+++ b/salt/grains/core.py
@@ -445,7 +445,7 @@ def _windows_virtual(osdata):
     grains = dict()
     if osdata['kernel'] != 'Windows':
         return grains
-    
+
     grains['virtual'] = 'physical'
 
     # It is possible that the 'manufacturer' and/or 'productname' grains
--- a/salt/grains/core.py
+++ b/salt/grains/core.py
@@ -445,6 +445,8 @@ def _windows_virtual(osdata):
     grains = dict()
     if osdata['kernel'] != 'Windows':
         return grains
+    
+    grains['virtual'] = 'physical'
 
     # It is possible that the 'manufacturer' and/or 'productname' grains
     # exist but have a value of None.


gtmanfred (approved):

damon-atkins (approved):
Seems a reasonable fix.

twangboy (approved):

PR #49878, opened 2018-10-03
[fluorine] Merge forward from 2018.3 to fluorine

rallytime:
Conflicts:
  - doc/man/salt.7

PR #50030, opened 2018-10-14
[fluorine] Updates to integration.sdb.test_vault

garethgreenaway:
### What does this PR do?
Updating the integration.sdb.test_vault setup function to attempt to recreate the container if the login test fails.  Also updating the internal counter to use VaultTestCase instead of self, following a successful run when all tests ran, the vault container was not stopped and deleted because the counter was not being incremented properly.

https://github.com/saltstack/salt/pull/49878

### Previous Behavior
The first test run in the integration.sdb.test_vault test would randomly fail because the vault container was not running.

### New Behavior
Try to recreate the vault container if the initial login attempt fails.  Try a maximum of three times before skipping the test.

### Tests written?
Yes. Existing tests updated.

### Commits signed with GPG?

Yes/No

### DIFF
--- a/tests/integration/sdb/test_vault.py
+++ b/tests/integration/sdb/test_vault.py
@@ -31,7 +31,7 @@ class VaultTestCase(ModuleCase, ShellCase):
         '''
         SetUp vault container
         '''
-        if self.count == 0:
+        if VaultTestCase.count == 0:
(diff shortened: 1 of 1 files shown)


gtmanfred (approved):

rallytime:
@garethgreenaway These tests are failing on python 3 now. Can you take a look?

garethgreenaway:
@rallytime on it.

cachedout (approved):

PR #50045, opened 2018-10-15
Don't throw exceptions just because tablespace exists

noelmcloughlin:
### What does this PR do?

Resolves 'KeyError' exception.

See https://github.com/saltstack-formulas/postgres-formula/issues/186

### Previous Behavior
'KeyError` if object already exists

### New Behavior
No  'KeyError' if object pre-exists.

### Tests written?

No

Manual tested:
https://github.com/saltstack-formulas/postgres-formula/issues/186#issuecomment-429872828

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/states/postgres_tablespace.py
+++ b/salt/states/postgres_tablespace.py
@@ -118,14 +118,14 @@ def present(name,
             return ret
 
     # already exists, make sure it's got the right config
-    if tblspaces[name]['Location'] != directory and not __opts__['test']:
+    if name in tblspaces and tblspaces[name]['Location'] != directory and not __opts__['test']:
         ret['comment'] = """Tablespace {0} is not at the right location. This is
             unfixable without dropping and recreating the tablespace.""".format(
                 name)
         ret['result'] = False
         return ret
 
-    if owner and not tblspaces[name]['Owner'] == owner:
+    if owner and name in tblspaces and not tblspaces[name]['Owner'] == owner:
         if __opts__['test']:
             ret['result'] = None
             ret['comment'] = 'Tablespace {0} owner to be altered'.format(name)


vutny (changes requested):
@noelmcloughlin @cachedout 

Looking at the original issue from the postgres-formula repo, I think these changes doesn't not resolve the issue at all. The state will try to create the same tablespace on each state run and we're going simply to mask that.
The "KeyError" exception looks deceptive, because it hides the original problem. That needs to be fixed though.

It appears that the `postgres.tablespace_list` execution module function is not able to read previously created tablespaces. That's why the tablespace name wasn't been found in the list.
It could be due insufficient permissions for the user, or more likely, some older/specific PostgreSQL versions or builds do not return existing tablespaces after this query:

```sql
'SELECT spcname as "Name", pga.rolname as "Owner", spcacl as "ACL",
spcoptions as "Opts", pg_tablespace_location(pgts.oid) as "Location"
FROM pg_tablespace pgts, pg_roles pga WHERE pga.oid = pgts.spcowner
```

Here we need to return proper failure state, and determine if the query above is portable across PG versions desired to be supported by Salt.

noelmcloughlin:
Hi @vutny 
Thanks once again for reviewing. I have raised https://github.com/saltstack/salt/issues/50081 and updated https://github.com/saltstack-formulas/postgres-formula/issues/186#issuecomment-430308978 with your comments.

The root cause is that `postgres_tablespace.list` cannot handle **selinux enforcing** mode.  I have an open PR to "fix" **postgres-formula** (https://github.com/saltstack-formulas/postgres-formula/pull/248) but I don't have time to fix Salt.

I would ask this PR get merged to hide the exception?  Thats my main gripe - end user experience.

Thanks

vutny (commented):
It is great that you found out the reason @noelmcloughlin 

However, to really improve user experience when dialing with such issues, I think instead of returning success result in subsequent state run, the `tablespace.present` function must **fail** with descriptive error message that it cannot create the tablespace with showing original error message from PostgreSQL.

Because even if the tablespace was created, it is unusable for both Salt and PG. Or even check the tablespace right after creation. This way a user will know that something went wrong.

noelmcloughlin:
Guys Sounds easy enough but I'll need to familiarize myself with the module. 

Will I close this PR meantime?

cachedout:
@noelmcloughlin I will go ahead and close this for now. You can either comment on this one when you are ready and we will re-open it or you can open a new PR. Thanks!

PR #50046, opened 2018-10-15
add http basis auth

arminhoh:
Adds username and password parameters for the http call.

### What does this PR do?

It adds the option to use http-basis-auth with the http_yaml pillar module. 

The username and password parameters are marked as TODO in the current Documentation and i tried to implement them with this PR.

### DIFF
--- a/salt/pillar/http_yaml.py
+++ b/salt/pillar/http_yaml.py
@@ -13,7 +13,6 @@ Set the following Salt config to setup an http endpoint as the external pillar s
   ext_pillar:
     - http_yaml:
         url: http://example.com/api/minion_id
-        ::TODO::
         username: username
         password: password
 
@@ -67,12 +66,15 @@ def __virtual__():
 def ext_pillar(minion_id,
                pillar,  # pylint: disable=W0613
                url,
-               with_grains=False):
+               with_grains=False,
+               **kwargs):
     '''
     Read pillar data from HTTP response.
 
     :param str url: Url to request.
     :param bool with_grains: Whether to substitute strings in the url with their grain values.
+    :param str username: Username for http basic auth
+    :param str password: Password for http basic auth
 
     :return: A dictionary of the pillar data to add.
     :rtype: dict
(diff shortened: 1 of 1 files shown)


salt/pillar/http_yaml.py
```diff
@@ -67,12 +66,15 @@ def __virtual__():
 def ext_pillar(minion_id,
                pillar,  # pylint: disable=W0613
                url,
-               with_grains=False):
+               with_grains=False,
+               **kwargs):
```
gtmanfred: I would prefer to accept `username` and `password` instead of adding `**kwargs`.

gtmanfred (changes requested):
There is no need to use `**kwargs`, it would be better to just pass the necessary variables through directly.

salt/pillar/http_yaml.py
```diff
@@ -97,7 +99,12 @@ def ext_pillar(minion_id,
             url = re.sub('<{0}>'.format(grain_name), grain_value, url)
 
     log.debug('Getting url: %s', url)
-    data = __salt__['http.query'](url=url, decode=True, decode_type='yaml')
+
+    if 'username' in kwargs and 'password' in kwargs:
+        data = __salt__['http.query'](url=url, decode=True, decode_type='yaml',
+                                      username=kwargs['username'], password=kwargs['password'])
+    else:
+        data = __salt__['http.query'](url=url, decode=True, decode_type='yaml')
```
gtmanfred: And if we just specify the defaults as none, then they can just be passed through as none anyway, and you don't need this if statement.

`data = __salt__['http.query'](url=url, decode=True, decode_type='yaml', username=username, password=password)`

vaskosdune: Thanks that would be way more elegant. I will try to get those changes in the pr asap.

salt/pillar/http_yaml.py
```diff
@@ -66,8 +66,9 @@ def __virtual__():
 def ext_pillar(minion_id,
                pillar,  # pylint: disable=W0613
                url,
-               with_grains=False,
-               **kwargs):
+               username=None,
+               password=None,
```
gtmanfred: One last thing, we have to maintain backwards compatibility, and if someone is passing in a True or False in to with_grains without the keyword, then this will break it.

So any new keyword arguments do need to be added to the end of the list.

Thanks!
Daniel

gtmanfred (changes requested):

salt/pillar/http_yaml.py
vaskosdune: Okey I did not think of that. Thanks for your quick reply and your patience.

Armin

gtmanfred (approved):
Looks great, thanks for the addition!

Welcome to the saltstack contributors club!

Thanks,
Daniel

DmitryKuzmenko (approved):

Issue #49295, opened 2018-08-23
New module.run syntax does not support keyword arguments


BenjaminSchiborr:
### Description of Issue
With Sodium a new syntax for the "module.run' function will be introduced as indicated by the following message: ``The function "module.run" is using its deprecated version and will expire in version "Sodium"``.
The required changes are described here: (https://docs.saltstack.com/en/develop/topics/releases/2017.7.0.html#state-module-changes)

This new syntax no longer supports the use of keyword arguments in a state. As such, converted states might show very confusing outcomes.
The first argument of the file.copy function is "src" and the second argument ist "dst" (https://docs.saltstack.com/en/latest/ref/modules/all/salt.modules.file.html#salt.modules.file.copy). In the example below this order is not followed.

### Setup
Given a (windows?) minion with two folders "Source" and "Dest". In each one of those folders there is a file.
We have two states. Both states copy content from folder "Source" to folder "Dest". State "old_copy" uses the old syntax and state "new_copy" uses the new syntax.

```
old_copy:
  module.run:
    - name: file.copy
    - dst: C:/Dest
    - src: C:/Source
    - recurse: True
    - remove_existing: False
```

```
new_copy:
  module.run:
    - file.copy:
      - dst: C:/Dest
      - src: C:/Source
      - recurse: True
      - remove_existing: False
```

Necessary changes to minion config to activate new syntax:
```
use_superseded:
- module.run
```

### Steps to Reproduce Issue
1. Execute state "old_copy" against the minion. The result will be that the folder "Dest" is containing both it's original files and those from folder "Source".
2. Revert the content of both folders back to its original state
3. Change the minion config to utilize the new module.run syntax
4. Execute state "new_copy" against the minion. The result will be that the folder "Source" is containing both it's original files and those from folder "Dest".

### Versions Report
```
Salt Version:
           Salt: 2017.7.5

Dependency Versions:
           cffi: Not Installed
       cherrypy: unknown
       dateutil: 2.7.3
      docker-py: Not Installed
          gitdb: 0.6.4
      gitpython: 1.0.1
          ioflo: Not Installed
         Jinja2: 2.8
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.4.6
   mysql-python: Not Installed
      pycparser: Not Installed
       pycrypto: 2.6.1
   pycryptodome: Not Installed
[...]

PR #50049, opened 2018-10-15
Preserving signature in "module.run" state

isbm:
### What does this PR do?

Bugfix for: https://github.com/saltstack/salt/issues/49295

### Previous Behavior

Keyword argument's ordering was still by order, not by key. That is, with the signature of `(src, dst)` these two statements were equal (and thus wrong):

```
foo:
  module.run
    - file.copy
      - dst: /tmp/dst
      - src: /tmp/src

bar:
  module.run
    - file.copy
      - /tmp/dst
      - /tmp/src
```

Expected behaviour is that `src` should be prior to `dst`, regardless how it is positioned, as long as it is designated with the key. Positional arguments only should be according to their order in the second example.

### New Behaviour

Function signature is now respected. For e.g. `file.copy(src, dst, **kw)`, these statements are all equally aligned to src/dst ordering:

```
foo:
  module.run
    - file.copy
      - dst: /tmp/dst
      - src: /tmp/src

foo:
  module.run
    - file.copy
      - src: /tmp/src
      - dst: /tmp/dst

fred:
  module.run
    - file.copy
      - /tmp/src
      - /tmp/dst
```

### Tests written?

No (existing should apply)

### DIFF
--- a/salt/states/module.py
+++ b/salt/states/module.py
@@ -323,7 +323,7 @@ def _call_function(name, returner=None, **kwargs):
 
     # func_args is initialized to a list of positional arguments that the function to be run accepts
     func_args = argspec.args[:len(argspec.args or []) - len(argspec.defaults or [])]
-    arg_type, na_type, kw_type = [], {}, False
+    arg_type, kw_to_arg_type, na_type, kw_type = [], {}, {}, False
     for funcset in reversed(kwargs.get('func_args') or []):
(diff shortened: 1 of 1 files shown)


isbm:
@rallytime @terminalmage @DmitryKuzmenko  *ping*

cachedout (approved):

PR #50053, opened 2018-10-15
introduce a concept of a reactor 'leader'

mattp-:
### What does this PR do?
in multimaster hot-hot environments, all events are streamed to all masters for a given minion. this introduces the problem of a reactor being executed on each hot master, regardless of if the reaction is idempotent or not. This patch introduces a simple is_leader bool that can be controlled via the reactor runner (or emitting the equivalent event to toggle) whether a reactor in hot-standby mode should be reacting to events coming into the bus. This provides a simple integration point for an external leader election tool such as consul-template. The feature itself is disconnected from multimaster, and is by default True, so it should be inert in most saltstack installs.

n/a.

### Tests written?
Yes

### Commits signed with GPG?
No


mattp-:
hmm. it occurs to me with this PR that any minion can use this means of events that reactor/manage provides to interact (possibly maliciously) with a reactor. is this something that should be enabled by default? is there a safer means of interacting with master processes?

cachedout:
@mattp- That's my immediate concern as well. We should discuss this at length but one thing that comes to mind is potentially using the master key as an auth token, similar to the way that the LocalClient checks to ensure a request for a publication is coming from the master itself.

austinpapp:
but this is the general, any minion could list the reactors and delete all reactors. that would be the same behavior as emitting set leader.

cachedout:
@DmitryKuzmenko Could you please comment on this proposal?

salt/utils/reactor.py
```diff
@@ -263,13 +264,21 @@ def run(self):
             # skip all events fired by ourselves
             if data['data'].get('user') == self.wrap.event_user:
                 continue
-
+            if data['tag'].endswith('salt/reactors/manage/is_leader'):
+                self.event.fire_event({'result': self.is_leader},
+                                       'salt/reactors/manage/leader/value')
+            if data['tag'].endswith('salt/reactors/manage/set_leader'):
```
DmitryKuzmenko: Let me to be a bit boring. =)
`elif`

salt/utils/reactor.py
```diff
@@ -263,13 +264,21 @@ def run(self):
...
+                # we only want to register events from the local master
+                if data['data'].get('id') == self.opts['id']:
+                    self.is_leader = data['data']['value']
+                self.event.fire_event({'result': self.is_leader},
+                                       'salt/reactors/manage/leader/value')
             if data['tag'].endswith('salt/reactors/manage/add'):
```
DmitryKuzmenko: `elif`

DmitryKuzmenko (approved):
Approving with hope to see the requested changes applied.

DmitryKuzmenko:
Overall I'm OK with the proposed feature. It doesn't change the default behavior and if customers able to effectively use that leading driven reactions it's great.

mattp-:
great :+1: I still need to sort out one last piece which is adding the master key as an auth token, will do that before merge alongside your recommended changes. thanks!

cachedout:
Hi @mattp-  

Just checking with you. :) We're waiting on your additional changes before we do more reviews here and I wanted to make sure that matched your expectation. :)

mattp-:
@cachedout this should be good to merge now, I added the master key auth thing. That reactor ping test is spurious.

PR #50061, opened 2018-10-15
pkg.install bugfix

rongzeng54:
### What does this PR do?
For example
the RPM package which current installed is 8:3.8.12-6.n.el7,
the RPM package which hope to install is 9:3.8.12-4.n.el7,
The yumpkg.py will run "yum downgrade" but not "yum install".

No

### Previous Behavior
Compare the package version without epoch.

### New Behavior
Compare the package version with epoch.

### Tests written?
No

### Commits signed with GPG?
No

### DIFF
--- a/salt/modules/yumpkg.py
+++ b/salt/modules/yumpkg.py
@@ -1502,7 +1502,8 @@ def install(name=None,
                     # provided. It could either be the OS architecture, or
                     # 'noarch', and we don't make that distinction in the
                     # pkg.list_pkgs return data.
-                    version_num = version_num.split(':', 1)[-1]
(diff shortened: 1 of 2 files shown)


terminalmage (commented):
Good catch, @rongzeng54. I've added a unit test for this to the PR.

@rallytime We should backport this fix.

terminalmage (approved):

rallytime (approved):
👍 

Nice team work @rongzeng54 and @terminalmage!

sagetherage:
Merge to Master https://github.com/saltstack/salt/pull/50114

Issue #50039, opened 2018-10-15
Windows: reg.set_value of type REG_QWORD causes traceback


dafyddj:
### Description of Issue/Question

Trying to set a Windows registry item of type REQ_QWORD causes traceback.

### Setup

Tried with Windows 8.1, Windows 10 1803. 
Salt 2018.3.2, develop branch

### Steps to Reproduce Issue
(Include debug logs if possible and relevant.)
```
PS C:\Python27\Scripts> ..\python.exe .\salt-call --local reg.set_value hive=HKLM key=SOFTWARE vname=test vdata=175 vtyp
e=REG_QWORD
[ERROR   ] Encountered error setting registry value
Traceback (most recent call last):
  File "C:\Python27\lib\site-packages\salt\utils\win_reg.py", line 708, in set_value
    win32api.RegSetValueEx(handle, local_vname, 0, vtype_value, local_vdata)
SystemError: ..\Objects\longobject.c:998: bad argument to internal function
local:
    False
```

### Versions Report
(Provided by running `salt --versions-report`. Please also mention any differences in master/minion versions.)
```
PS C:\Python27\Scripts> ..\python.exe .\salt-call --versions-report
Salt Version:
           Salt: 2018.11.0-234-ge57a110

Dependency Versions:
           cffi: 1.5.0
       cherrypy: 5.0.1
       dateutil: 2.4.2
      docker-py: Not Installed
          gitdb: 0.6.4
      gitpython: 1.0.1
         Jinja2: 2.8
        libgit2: Not Installed
       M2Crypto: Not Installed
           Mako: 1.0.3
   msgpack-pure: Not Installed
 msgpack-python: 0.4.7
   mysql-python: Not Installed
      pycparser: 2.14
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: Not Installed
         Python: 2.7.12 (v2.7.12:d33e0cf91556, Jun 27 2016, 15:24:40) [MSC v.1500 64 bit (AMD64)]
   python-gnupg: 0.3.8
         PyYAML: 3.11
          PyZMQ: 15.2.0
          smmap: 0.9.0
        timelib: Not Installed
        Tornado: 4.3
            ZMQ: 4.1.2

System Versions:
           dist:
         locale: cp1252
        machine: AMD64
        release: 8.1
         system: Windows
        version: 8.1 6.3.9600  Multiprocessor Free
```

Ch3LL:
[...]

PR #50063, opened 2018-10-15
Use long when setting REG_QWORD

twangboy:
### What does this PR do?
Uses long for REG_QWORD values in Py2

https://github.com/saltstack/salt/issues/50039

### Tests written?
Yes

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/utils/win_reg.py
+++ b/salt/utils/win_reg.py
@@ -767,6 +767,9 @@ def cast_vdata(vdata=None, vtype='REG_SZ'):
     # Make sure REG_MULTI_SZ is a list of strings
     elif vtype_value == win32con.REG_MULTI_SZ:
         return [_to_unicode(i) for i in vdata]
+    # Make sure REG_QWORD is a 64 bit integer
+    elif vtype_value == win32con.REG_QWORD:
+        return vdata if six.PY3 else long(vdata)
     # Everything else is int
     else:
         return int(vdata)
--- a/salt/utils/win_reg.py
+++ b/salt/utils/win_reg.py
@@ -769,7 +769,7 @@ def cast_vdata(vdata=None, vtype='REG_SZ'):
         return [_to_unicode(i) for i in vdata]
     # Make sure REG_QWORD is a 64 bit integer
     elif vtype_value == win32con.REG_QWORD:
-        return vdata if six.PY3 else long(vdata)
+        return vdata if six.PY3 else long(vdata)  # pylint: disable=W1699
     # Everything else is int
     else:
         return int(vdata)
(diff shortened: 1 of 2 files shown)


twangboy:
Regarding the lint failure... It is suggesting that I use `int` instead of `long` in Py2. It has to be `long` on Py2 for the Windows API call that occurs later on.

rallytime:
@twangboy Can you put a pylint disable for that check then? We can't merge this if lint is red. :)

lorengordon:
Conversion guide seems to suggest `from builtins import int`... Try that yet?

http://python-future.org/compatible_idioms.html#long-integers

dwoz:
@twangboy Lets see if @lorengordon 's comment works before merging this.

dwoz (changes requested):
@twangboy Lets see if @lorengordon's comment works before merging this.

twangboy:
@lorengordon Your suggestion requires adding another dependency to Salt. It doesn't work on Py2 by default. You have to `pip install future`. So, rather than add another dependency, I think I would rather just use this fix.

lorengordon:
@twangboy Ahh, that's true. I already use `future` everywhere I need 2/3 compatibility because it is so darn handy for these things (much like `six`). Figured you were also. Oh well.

salt/utils/win_reg.py
```diff
@@ -767,6 +767,9 @@ def cast_vdata(vdata=None, vtype='REG_SZ'):
     # Make sure REG_MULTI_SZ is a list of strings
     elif vtype_value == win32con.REG_MULTI_SZ:
         return [_to_unicode(i) for i in vdata]
+    # Make sure REG_QWORD is a 64 bit integer
+    elif vtype_value == win32con.REG_QWORD:
+        return vdata if six.PY3 else long(vdata)  # pylint: disable=W1699
```
rallytime: What about using `six.integer_types` here?

Seems nicer that disabling the lint error.

https://pythonhosted.org/six/#six.integer_types

twangboy: I tried that, but it seems to only be useful for detecting types not casting.
```
>>> from salt.ext import six
>>> long(7)
7L
>>> six.integer_types(7)
TypeError: 'tuple' object is not callable
>>> isinstance(7, six.integer_types)
True
>>> isinstance(7L, six.integer_types)
True
>>>
```

twangboy: @rallytime You want I should add `future` as a dependency instead? That works and is quite clean.

rallytime: Ah, right, OK. I think this is fine as-is. I'd rather not add the dependency. Your current check works.

rallytime (approved):

cachedout:
@dwoz Do you want to come back and look at this again?

dwoz (approved):

cachedout:
We didn't get Windows test results on that run so I've restarted the test suite here.

twangboy:
Rebased

PR #50078, opened 2018-10-16
Job retry RFC

cachedout:


isbm:
@cachedout Maybe description would help

rfcs/0003-job-retry.md
```diff
@@ -0,0 +1,57 @@
...
+
+Jobs to be retried are inserted into the cache by the LocalClient when it deterermines that a minion which it expected
+to return did not.
+
+Please see https://github.com/saltstack/salt/compare/develop...cachedout:job_retry for a POC.
+
+## Alternatives
+[alternatives]: #alternatives
+
+No alternatives considered but ideas welcome.
```
isbm: My alternative try here:

1. Master throws job at minion, but first it travels from Master's job cache to Minion's job cache.
2. Minion says "I've got it!"
3. Master deletes the job from the cache.
4. Minion is looking into the "job cache" and polls it.
5. If job is found, Minion performs it and commits its deletion only when guaranteed finished.
6. In case Minion has been restarted/crashed/killed/ripped by power down etc, job still there.
7. Master no longer needed here, so Minion can just wake up back, and continue from No.5 step.

Why I think this could be better than original proposal:

- It does not need to be explicitly dependent on the Master.
- It can work with LocalClient
- Great benefit for Salt SSH! We should definitely reuse this mechanism for Salt SSH. I am thinking something like in `Emacs <-> emacs-client` fashion for _concurrent_ Salt SSH calls. In this case, if already running SaltSSH process performing something, another SaltSSH spin won't create the whole isolated Minion within local Master like `--local` but see if that already running around. Then it should just open Minion Job Cache and put another job inside and quit. The previous, already running process, will continue polling Job Cache until empty, and thus will find a new job and will perform it! This will resolve chronic problem of simultaneously fired SaltSSH comments under the same user name.

```
[master]--->[job cache]---+
                          |
                          |       <<<< John Wick accidentally shoots the network cable. 
  +-----------------------+
  |
  V
[job cache]--->[minion]
```

So then "job cache" on the minion would have it.

isbm (changes requested):

rfcs/0003-job-retry.md
gtmanfred: What if we used the Queue subsystem instead? and queued up jobs on a per minion basis so that the jobs can be replayed in order when a minion comes online?

gtmanfred (approved):
I really like the use of the engine system, i would like to see us dogfood the queue system instead of using the cache subsystem, personally i think that is a better format to store the job information.

damon-atkins:
2c. You want to be able to filter this, so it needs a filter option, with the default being "states" only.
e.g. `salt \* test.ping` or `salt \* disk.usage` you do not want to be replayed etc.

rares-pop:
I concur with @damon-atkins, but more flexibility would be even nicer - any job have a way to specify if it wants to retry or not (maybe reuse the metadata mechanism)?

adelcast:
A use case of ours require that we keep a queue of jobs that need to be run sequentially, as well as a jobs that can run in parallel (read-only jobs). I believe the Oxygen option PROCESS_COUNT_MAX, when set to 1 can give the queuing mechanism for sequential jobs (combined with the retry option on this RFC). If jobs could be tagged in a way that will prevent retry, that could give us the flexibility to queue/retry only the jobs that are tagged as retry. @rares-pop

Issue #45314, opened 2018-01-07
RFE: master file server addition information e.g. ETag


damon-atkins:
### Description of Issue/Question
Can the backed be updated so instead of just returning simple information i.e. filename it always returns a dict
Now looks like
```
cp.list_master
- file1
- file2
- dir1\file3
```

Future look, however every command should receive a dict for each file name, may not use the info
```
cp.list_master detail=True
'file1':
   Time: seconds since 1970
   ETag: string
   Size: 
   Type: f(ile) or d(irectory)
'file2':
   Time: 7483278389
   ETag: jfds8jlfsjd8ereteghyrbvvdffeeejfljdl
   Size: 7184
   Type: f
'dir':
   Time: seconds since 1970
   ETag: <Maybe a checksum of all the filenames and modification str times>
   Size: <items in directory> or None
   Type: d
'dir1\file3':
   Time: 6367
   ETag: rfejlwjkldggdgfddgthdvhyhyytrfwe
   Size: 9373
   Type: f 
``` 

ETag its up to the client to record it so it knows if the file has change or not. The client does not care how it was generated and does not need to generate/calculate it itself. However the server must change it if the file changes on the server. ETag could be a commit tag or something else from git.

It's important all masters give out the same ETag for the same file. An Etag which uses inode number would fail as this would be different on different masters.

+1 if you like the idea.

The next stage would be for the clients to use this information as part of caching.

gtmanfred:
@saltstack/team-core does anyone have any opinions about this?

Thanks,
Daniel

thatch45:
I am fine with this as a feature addition as long as the default interface does not change. Any changes to the master/minion apis needs to be very carefully managed as we don't want to create and master/minion incompatibilities.

damon-atkins:
It's a change to the salt:// protocol.  Aim being that it will help make the cache on the minion faster as it can be smarter about what it does.  And opens the master to decided what to use the for ETag (e.g. checksum on a flat file system or git info from a git repo)
i.e. pulls a file down and saves the "file meta information", then to check if the file is out of date it compares save "file meta information" with the current "meta information" from the master.  (The client does not spend time using a checksum to determine if the file is out of date). Its basically how http proxy servers cache work.

damon-atkins:
Does the current salt:// protocol have version of the protocol in the request, or the minion version including in the request? So the master knows how to respond in a compatible way.

thatch45:
[...]

PR #50085, opened 2018-10-16
Fix `test_win_pkg` yet again

dwoz:
### What does this PR do?

We just fixed a bug in `check_file_list_cache` which should have resolved the `test_win_pkg` failure but this test still fails intermittently. This PR adds flushes and syncs the new file to make sure it's available. In addition to adding a debug log when file_list is returned from a cache.

### Tests written?

No

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/fileserver/__init__.py
+++ b/salt/fileserver/__init__.py
@@ -134,7 +134,17 @@ def check_file_list_cache(opts, form, list_cache, w_lock):
                     # st_time can have a greater precision than time, removing
                     # float precision makes sure age will never be a negative
                     # number.
-                    age = int(time.time()) - int(cache_stat.st_mtime)
+                    current_time = int(time.time())
+                    file_mtime = int(cache_stat.st_mtime)
+                    if file_mtime > current_time:
+                        log.debug(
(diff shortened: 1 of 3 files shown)


twangboy (approved):

damon-atkins:
Should not need to flush. Flush just ensures its on physical disk.   With out the flush, the file is read back from memory cache this memory cache is shared my everything, so all reads from any process will get the same results.. If their is a difference this would indicate a very large OS bug.   If you have python closed the file, the data should not be sitting in any python buffer.

dwoz:
The log from the last failing test run showed this...

```
Returning file list from cache: age=-3092 cache_time=0 c:\users\admini~1\appdata\local\temp\salt-tests-tmpdir\rootdir\cache\file_lists\roots\base.p
```

For some reason the cache file's modified time was 3092 seconds in the future. I patched to detect this case an log that it happened.

damon-atkins:
I see your looking at time. Are you running NTP? Virtual Machine are good at loosing or gaining time. I assume everything is processed as UTC time. I raise a PR to suggest some enhancements to the fileserver some time ago. #45314

PR #50087, opened 2018-10-16
Fix applying of attributes for returner rawfile_json

rbthomp:
Arguments are not getting applied to the rawfile_json returner. For example if you specify an alternate filename for the output the default "/var/log/salt/events" is always used. Passing the `ret` to `_get_options(ret) resolve this.


max-arnold:
There is a second occurrence of the `_get_options({})` here https://github.com/saltstack/salt/blob/0694cdf6df1676ad37037e17bd5afe504bc8c86b/salt/returners/rawfile_json.py#L76
I'm not sure if it also needs to be fixed.

@fdamstra Could you please take a look?

rbthomp:
I wasn't sure that the `event_returner` needed it as the event_returner is set in the master config and isn't called by the minion. I couldn't think of a way that `filename` or other args could be passed to `_get_options() for the event_returner other then from the master config.

Is my logic on this correct?

PR #50088, opened 2018-10-16
Fix incorrect user/group id handling in file state

vulnbe:
### What does this PR do?

It fixes error in salt.states.file. 

Issues messages are:
```
user not specified, but configured as a target for recursive ownership
group not specified, but configured as a target for recursive ownership
```

Issue occurs when user or group is set to 0 and 'recurse' is used: 

```
/var/log/test/:
  file.directory:
    - user:  0
    - group: 0
    - recurse:
      - user
      - group
```

### Tests written?

No

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/states/file.py
+++ b/salt/states/file.py
@@ -3188,7 +3188,7 @@ def directory(name,
             # NOTE: Should this be enough to stop the whole check altogether?
     if recurse_set:
         if 'user' in recurse_set:
-            if user:
+            if user or isinstance(user, int):
                 uid = __salt__['file.user_to_uid'](user)
                 # file.user_to_uid returns '' if user does not exist. Above
                 # check for user is not fatal, so we need to be sure user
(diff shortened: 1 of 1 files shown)


dwoz:
@vulnbe Does the state work when user is set to a string instead of int?

```
/var/log/test/:
  file.directory:
    - user:  '0'
    - group: '0'
    - recurse:
      - user
      - group
```

vulnbe:
@dwoz It doesn't work.

dwoz (approved):
I see, user/group can be an id.. `file.user_to_uid` just returns the int un-altered. I think the accepting both strings and integers here is a bit odd but that has nothing to do with your patch. Otherwise LGMT.

dwoz:
@vulnbe We noticed this PR was against 2018.3.3.. That release has already been 'frozen', although I merged this to the 2018.3.3, the changes won't be included in that release. We'll make sure they are in the next one though.

PR #49872, opened 2018-10-03
Fix issues with windows file permissions when using reset=True

twangboy:
### What does this PR do?
Fix issues with file permissions:
- Fix stacktrace when attempting to iterate over None
- Check permissions using the resolved username
- Set permissions using the resolved username
- Reload file permissions prior to resetting
- Compare usernames using lower()

### What issues does this PR fix or reference?
https://github.com/saltstack/salt/issues/49861

### Tests written?
Yes

### Commits signed with GPG?
Yes

PR #50100, opened 2018-10-17
[fluorine] Merge forward from 2018.3 to fluorine

rallytime:
Conflicts:
  - salt/fileclient.py
  - salt/grains/core.py
  - salt/modules/win_file.py
  - salt/pillar/__init__.py
  - salt/utils/http.py

The `test_win_file` unit tests might have some troubles here. Let's see how it goes and then I will address them later if they fail. They might need to be applied separately because #49872 would not merge forward cleanly as that file was heavily refactored in fluorine and the fix will need to be applied separately.


gtmanfred (approved):

garethgreenaway (approved):

rallytime:
This has a merge conflict now. I need to re-do this.

PR #50118, opened 2018-10-18
Add parameter named pillar_cache_env

rongzeng54:
### What does this PR do?
Considering the production environment, the difference in the frequency of use of different pillar environments is huge and should be treated differently.

No

### Previous Behavior
No matter what pillarenv is, always store the pillar cache.

### New Behavior
Only cache the pillar data for specified environment. Default is base.

### Tests written?
No

### Commits signed with GPG?
No

### DIFF
--- a/conf/master
+++ b/conf/master
@@ -1006,6 +1006,10 @@
 # pillar is recompiled and stored. A value of 0 will cause the cache to always be valid.
 #pillar_cache_ttl: 3600
 
+# If and only if a master has set ``pillar_cache: True``, the cache env controls the
+# pillarenv of pillar data, before the cache is considered stored by a master.
+#pillar_cache_env: 'base'
+
 # If and only if a master has set `pillar_cache: True`, one of several storage providers
 # can be utilized.
 #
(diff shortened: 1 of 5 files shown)


conf/master
```diff
@@ -1006,6 +1006,10 @@
 # pillar is recompiled and stored. A value of 0 will cause the cache to always be valid.
 #pillar_cache_ttl: 3600
 
+# If and only if a master has set ``pillar_cache: True``, the cache env controls the
+# pillarenv of pillar data, before the cache is considered stored by a master.
+#pillar_cache_env: 'base'
```
gtmanfred: This should probably be a list, so that you can provide more than just one environment if you wanted to.

gtmanfred (changes requested):

conf/master
rongzeng54: now dont support cache multi pillarenv, shoule we support ?

rongzeng54: yes. but seems pillar_cache can only work for one env at the same time.

gtmanfred: That should not be true, it should be caching the pillar dictionary on a per minion basis.

Looking at this closer, i am not sure this is a good approach to this, because any case that this is tried to be used without a pillarenv set, and just environments assigned in the top file, this will not work, which is bad because using environments from a top file is the default.

rongzeng54: hmm, i think i found a bug. i will commit another pr for fix it.

PR #50120, opened 2018-10-18
Fix bad merge conflict resolution

terminalmage:
b98df6d was incorrectly merged forward into 2018.3 (or into develop, before 2018.3 was branched). This causes failures to redeploy the thin tarball when it has changed, under certain circumstances.

This commit fixes the problem. I'm also leaving in some changes to the logging that I made in the course of troubleshooting this, as they make the border between stdout and stderr more explicit, as well as adding the shim command's retcode.

Fixes https://github.com/saltstack/salt-jenkins/issues/1137.


gtmanfred (approved):

isbm (approved):
@terminalmage ohwell, happens. @Ch3LL don't give up! :wink:

PR #49886, opened 2018-10-03
localclient: pass input args/kwargs through salt.utils.args.parse_input

mattp-:
### What does this PR do?
this functionality brings localclient in line with what happens on the
cli, as well as what runnerclient does. this ensures that any foo=bar
kwargs passed in as args are handled accordingly. right now localclient
relies on SaltCMDOptionParser to do this handling for us, but that
leaves a hole when salt-api interactions are brought in.

### What issues does this PR fix or reference?
should fix https://github.com/saltstack/pepper/issues/99

### Previous Behavior
kwargs passed in via low state via salt-api were not parsed correctly.
 
### New Behavior
yaml based kwargs are now handled correctly. see referenced issue for details.

### Tests written?
No, the existing tests should provide sufficient coverage for the behavior of localclient

### Commits signed with GPG?
Yes

Please review [Salt's Contributing Guide](https://docs.saltstack.com/en/latest/topics/development/contributing.html) for best practices.

See GitHub's [page on GPG signing](https://help.github.com/articles/signing-commits-using-gpg/) for more information about signing commits with GPG.

Issue #42491, opened 2017-07-23
Document cherrypy /token endpoint


mchugh19:
### Description of Issue/Question
The [2017.7 release notes](https://docs.saltstack.com/en/develop/topics/releases/2017.7.0.html#salt-api-changes) mention a /token endpoint. However, this endpoint is not mentioned in the [rest_cherrypy docs](https://docs.saltstack.com/en/latest/ref/netapi/all/salt.netapi.rest_cherrypy.html#login). 

So from this, there are a few questions:
1. Is this simply an alias for /login, or intended to replace it? The /login endpoint still appears to function and the /token [commit](https://github.com/saltstack/salt/pull/41352/commits/bec3da3a2f0d871f6a735c87b9c0abb8145bdae9) appears to show the same usage as /login.
2. If the rest interface is being updated, is there a way to query the version of the api to be able to support multiple versions?
3. Is there intent to support the same api endpoints in tornado?

mchugh19:
With the deprecation of expr_form in favor of the newer tgt_type in 2017.7, it is important to get version string from the salt-api in order to supporting multiple versions.

mchugh19:
Hey all. Seems this will continue to be an issue. jenkinsci/saltstack-plugin#105 and jenkinsci/saltstack-plugin#99 demonstrate salt-api api breaking changes. Is there any mechanism for querying the version of the salt-api in use in order to support both pre and post 2017 json formats?

rbjorklin:
@mchugh19 I can think of an ugly hack to find out which version is running in case there is no official support for it:

    #!/bin/bash

    # Write the cookie file on login:
    curl -sSk https://localhost:8000/login \
      -c ~/cookies.txt \
      -H 'Accept: application/x-yaml' \
      -d username=user \
      -d password=pass \
      -d eauth=pam
                                                                                   
    # Read the cookie file and query salt-api version:
    curl -sSk https://localhost:8000 \
      -b ~/cookies.txt \
      -H 'Accept: application/x-yaml' \
      -d client=runner \
      -d fun=salt.cmd \
      -d arg=cmd.run \
      -d arg="salt-api --version"

mchugh19:
As far as I know, this issue is still unresolved. There should still be documentation about the purpose and use for the /token API, broadcasted version numbers, and the other items listed above.

mattp-:
[...]

Issue #39783, opened 2017-03-02
Cannot pass pillar data to the slack engine


mdaffin:
### Description of Issue/Question

Pillar data cannot be passed to the slack engine from slack or the engine config resulting in `Pillar data must be formatted as a dictionary` error message.

I expected it to work the same way as the command line. ie you should be able to copy and paste a command from the command line into slack or the `cmd` line in the alias.

#### /srv/salt/orch/ping.sls
```
{% set tgt = pillar.get('target', '*') %}
test-ping:
    salt.function:
    - name: test.ping
    - tgt: '{{ tgt }}'
```

#### /etc/salt/master.d/slack-engine.conf
```
engines:
    slack:
       token: '<token>'
       control: True
       valid_users:
           - <someuser>
       valid_commands:
           - state.orchestrate
           - ping-web
       aliases:
           ping-web:
               cmd: |
                   state.orchestrate orch.ping pillar='{"target": "web-*"}'
```

### Steps to Reproduce Issue
From the salt master running `salt-run state.orchestrate orch.ping pillar='{"target": "web-*"}'` works as expected.

In slack running `!state.orchestrate orch.ping pillar='{"target":"web-*"}'` or `!ping-web` produces the following error in slack.
```
"Exception occurred in runner state.orchestrate: Traceback (most recent call last):\n  File \"/usr/lib/python2.7/site-packages/salt/client/mixins.py\", line 395, in _low\n    data['return'] = self.functions[fun](*args, **kwargs)\n  File \"/usr/lib/python2.7/site-packages/salt/runners/state.py\", line 54, in orchestrate\n    'Pillar data must be formatted as a dictionary'\nSaltInvocationError: Pillar data must be formatted as a dictionary\n"
```

I have tried a variety of different pillar data, including formatting is as a python dict `'{target: "web-*"}'`.

### Versions Report
We also saw this in the previous version (2016.11.2).
```
Salt Version:
           Salt: 2016.11.3
 
Dependency Versions:
           cffi: 1.6.0
       cherrypy: 3.2.2
       dateutil: 1.5
          gitdb: 0.6.4
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.7.3
        libgit2: 0.21.4
        libnacl: Not Installed
       M2Crypto: 0.21.1
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.4.8
   mysql-python: Not Installed
      pycparser: 2.14
[...]

PR #39472, opened 2017-02-17
Update _reformat_low to not run kwarg dicts through parse_input

whiteinge:
### What does this PR do?

Keyword args were incorrectly getting run through `parse_input`. E.g., `{'fun': 'jobs.list_jobs', 'jid': '1234'}` would run the `jid` kwarg through that function and then YAML would try to parse the argument as a number.

load_args_and_kwargs() call already calls parse_input() which already
calls condition_input() so I think we can do with just the one call.

This also removes a duplicate `{..., 'kwarg': {'kwarg': {...}}}` in the output.

### What issues does this PR fix or reference?

#38962

### New Behavior

I'm 43% sure this doesn't regress anything...

### Tests written?

No

@DmitryKuzmenko you've done some work here, do you mind giving this a look?

PR #50124, opened 2018-10-19
make salt-api arg/kwarg handling the same as cli for all clients

### LINKED ISSUE #99: pepper vs. salt has different `test.arg_repr` output types
When using `pepper` versus the standard `salt` command you get different results in terms of returned data types. 

This seems related to #57 but different in the details.

```bash
(envsalt)  >>> pepper -a pam --username=saltvirt --password=saltvirt 'sip-cannon' test.arg_repr trunktest2 2 '2048' image=/var/lib/libvit/images/centos66.qcow2 enable_vnc=True config='{"master": "goodboy.qa.sangoma.local"}' start=False enable_qcow=True                     
{
    "return": [
        {
            "sip-cannon": {
                "args": "('trunktest2', '2', '2048')",
                "kwargs": "{'__pub_user': 'saltvirt', '__pub_arg': ['trunktest2', '2', '2048', 'image=/var/lib/libvirt/images/centos66.qcow2', 'enable_vnc=True', 'config={\"master\": \"goodboy.qa.sangoma.local\"}', 'start=False', 'enable_qcow=True'], 'start': False, 'config
': {'master': 'goodboy.qa.sangoma.local'}, 'image': '/var/lib/libvirt/images/centos66.qcow2', '__pub_fun': 'test.arg_repr', 'enable_qcow$
: True, '__pub_jid': '20170301004836803665', '__pub_tgt': 'sip-cannon', '__pub_tgt_type': 'glob', 'enable_vnc': True, '__pub_ret': ''}"
            }
        }
    ]
}
 
(envsalt)  >>> sudo salt 'sip-cannon' test.arg_repr trunktest2 2 2048 image=/var/lib/libvirt/images/centos66.qcow2 enable_vnc=True con$
ig='{"master": "goodboy.qa.sangoma.local"}' start=False enable_qcow=True
sip-cannon:
    ----------
    args:
        ('trunktest2', 2, 2048)
    kwargs:
        {'__pub_user': 'sudo_tyler', '__pub_arg': ['trunktest2', 2, 2048, {'enable_qcow': True, 'image': '/var/lib/libvirt/images/centos66.qcow2', 'start': False, 'config': {'master': 'goodboy.qa.sangoma.local'}, 'enable_vnc': True}], 'enable_qcow': True, 'config': {'master': 'goodboy.qa.sangoma.local'}, 'image': '/var/lib/libvirt/images/centos66.qcow2', '__pub_fun': 'test.arg_repr', 'start': False, '__pub_id': '20170301004859437566', '__pub_tgt': 'sip-cannon', '__pub_tgt_type': 'glob', 'enable_vnc': True, '__pub_ret': ''}
```

Notably the `arg` array contains different types (in this case most obviously `int` versus `str`),
- `pepper`:
```python
('trunktest2', '2', '2048')
```

- `salt`: 
```python
('trunktest2', 2, 2048)
```

In my case this is turn causes a parsing problem downstream in the `salt.modules.virt.init()` function:
```
00:34:44 [ 2382] [DEBUG   ] Generating VM XML                                                                                            
libvirt: Domain Config error : XML error: Invalid value '20482048204820482048204820482048204820482048204820482048204820482048204820482048
20482048204820482048204820482048204820482048204820482048204820482048204820482048204820482048204820482048204820482048204820482048204820482
04820482048204820482048204820482048204820482048204820482048204820482048204820482048204820482048204820482048204820482048204820482048204820
48204820482048204820482048204820482048204820482048204820482048204820482048204820482048204820482048204820482048204820482048204820482048204
82048204820482048204820482048204820482048204820482048204820482048204820482048204820482048204820482048204820482048204820482048204820482048
20482048204820482048204820482048204820482048204820482048204820482048204820482048204820482048204820482048204820482048204820482048204820482
04820482048204820482048204820482048204820482048204820482048204820482048204820482048204820482048204820482048204820482048204820482048204820
482048204820482048204820482048204820482048204820482048204820482048204820482048204820482048204820482048204                                
```

mattp-:
### What does this PR do?
both cherrypy and saltnado now pass all args through parse_input to
allow for behavior equivalent to salt cli yaml based arg parsing. This
will allow for pepper to behave more like salt cli, or direct api users.

also included is a minor feature to saltnado/cherrypy, a new X-Salt-Version
header for usage in pepper to determine if this feature among others
is available to use.

will fix https://github.com/saltstack/pepper/issues/99 when equivalent pepper changes are made.

### Previous Behavior
yaml based foo=bar args are not passed through parse_input

### New Behavior
args are passed through parse_input

### Tests written?
yes

### Commits signed with GPG?
No


mattp-:
@rallytime I notice you disabled one of the other saltnado tests for being too flaky, the one I've added is essentially the same. It's not clear to me what race condition was the problem, but we can turn this test off for now as well if you like.

gtmanfred (changes requested):
looks like we just need some tests expected results updated for this change, and also pylint has one error, otherwise this looks perfect.

Thanks!
Daniel

mchugh19:
I'd suggest throwing something into the release notes as well. Modifying the args/kwargs handling of salt-api and the version header is a pretty large (and positive) change.

mattp-:
@gtmanfred ok, should have the tests/lint fixed up. will let jenkins chew on it for a bit
@mchugh19 seems reasonable, but i think generally the release note management is done by the release team when they cut releases, not in this PR :)

gtmanfred:
Nope, we do it in the PRs when we add features, it would be great to include this in the Neon release notes.

mattp-:
ah gotcha, added salt-api release notes to neon.rst

gtmanfred (approved):

rallytime:
@whiteinge Is this something you'd like to review?

whiteinge:
Thanks for the ping, @rallytime. This functionality should already be baked into Salt and I did not include this functionality in salt-api on purpose for the reasons described below. That said, I'll happily defer to the wishes of the core team and any contributors that want to put in the actual work. Please consider or disregard as you see fit.

Parsing the `arg=val` strings into kwargs is a central workflow in Salt and thus should live in a central place and work consistently regardless if you are calling Salt from the CLI, from a REST API, from a Python wrapper, HTML form fields, a Slack bot, and many, many other examples we've seen. The best central place to do that is Salt's Python API (`LocalClient`, `RunnerClient`, `WheelClient`, etc). That is what the above-linked Pepper issue is discussing. All those client interfaces **should** run the `arg` (singular) array values through the utils function we have that splits `arg=val` strings apart and parses the right-hand side of that as YAML. I know for certain this has worked correctly in the past for LocalClient and RunnerClient (I don't recall offhand where WheelClient stood), because I have used it many, many times and even wrote tests for some of it. However this functionality has also regressed more than once (again the Pepper discussion) and I can't speak to the current state of things.

https://github.com/saltstack/salt/blob/23e0d40d833c965fe573bee88c11be677f5b8e5c/tests/integration/client/test_runner.py#L108-L135

IMO the best fix is to figure out how to prevent future regressions, to keep this functionality in the `*client` interfaces and to _not_ to bake this into salt-api.

A concrete example of why this is best done in Salt instead of salt-api is [this discussion about the Slack engine](https://github.com/saltstack/salt/issues/39783#issuecomment-284116852). A Slack bot would be accepting `arg=val` strings because people are typing them into a chat interface -- just like the CLI, just like webhooks, just like Pepper, just like HTML forms, etc. It would be ideal for any module type that goes through the Python API (not just the REST API which of course uses the Python API, just like all the other module types) to not have to reinvent this over and over.

It is easier to put this into salt-api, for sure, and fixing this (and avoiding future regressions) in core may be quite difficult indeed (https://github.com/saltstack/salt/pull/39472). But I do feel that is the better solution.

My $0.02. I sincerely wish whoever takes either path good luck and godspeed. 😄

mattp-:
@whiteinge thanks for chiming in, appreciate your input. I actually _did_ start off with a separate PR to put parse_input in localclient directly (https://github.com/saltstack/salt/pull/49886), however after some chat with @cachedout / @terminalmage we were hesitant to put a change in localclient directly that manhandles args since its such a sensitive codepath (some tests surfaced expected behavior in the contrary). however if 'arg={yaml: foo}'  is a documented behavior of localclient at the *api layer*, I think the original change makes sense, and we can change those tests.
That said, I don't call the shots on what goes into salt either :) I lean towards your proposal since runnerclient already works that way. Anyone else have any thoughts/decisions?

rallytime:
I agree with @whiteinge on this one. I wasn't aware of that history. Thanks for filling that in!

I am definitely curious what @cachedout and @terminalmage may have to say, given @whiteinge's explanation above.

terminalmage (approved):

terminalmage:
I need to re-evaluate this, I did not see all the conversation here.

mattp-:
per slack convo with @terminalmage im going to close this in favor of #49886 once that is fixed up and ready for merge.

mchugh19:
WIth #49886 being merged, the release notes and X-Salt-Version header from this PR would still be very helpful and resolve part of #42491

PR #50125, opened 2018-10-19
Add root and no_recommends parameters in the public interface for Zypper and RPM (lowpkg)

aplanas:
### What does this PR do?

New feature

### Tests written?

For the low level RPM yes, for the zypper level I adjusted the current tests.

### DIFF
--- a/salt/modules/rpm.py
+++ b/salt/modules/rpm.py
@@ -145,7 +145,7 @@ def list_pkgs(*packages, **kwargs):
     '''
     pkgs = {}
     cmd = ['rpm']
-    if 'root' in kwargs:
+    if kwargs.get('root'):
         cmd.extend(['--root', kwargs['root']])
     cmd.extend(['-q' if packages else '-qa',
                 '--queryformat', r'%{NAME} %{VERSION}\n'])
@@ -208,7 +208,7 @@ def verify(*packages, **kwargs):
             verify_options = [x.strip() for x in six.text_type(verify_options).split(',')]
 
     cmd = ['rpm']
-    if 'root' in kwargs:
+    if kwargs.get('root'):
         cmd.extend(['--root', kwargs['root']])
     cmd.extend(['--' + x for x in verify_options])
     if packages:
@@ -281,7 +281,7 @@ def modified(*packages, **flags):
         salt '*' lowpkg.modified
     '''
     cmd = ['rpm']
-    if 'root' in flags:
+    if flags.get('root'):
         cmd.extend(['--root', flags.pop('root')])
     cmd.append('-Va')
     cmd.extend(packages)
@@ -358,7 +358,7 @@ def file_list(*packages, **kwargs):
         salt '*' lowpkg.file_list
     '''
     cmd = ['rpm']
-    if 'root' in kwargs:
+    if kwargs.get('root'):
         cmd.extend(['--root', kwargs['root']])
 
     cmd.append('-ql' if packages else '-qla')
@@ -394,7 +394,7 @@ def file_dict(*packages, **kwargs):
     ret = {}
     pkgs = {}
     cmd = ['rpm']
-    if 'root' in kwargs:
+    if kwargs.get('root'):
         cmd.extend(['--root', kwargs['root']])
     cmd.extend(['-q' if packages else '-qa',
                 '--queryformat', r'%{NAME} %{VERSION}\n'])
@@ -409,7 +409,7 @@ def file_dict(*packages, **kwargs):
         pkgs[comps[0]] = {'version': comps[1]}
     for pkg in pkgs:
         cmd = ['rpm']
-        if 'root' in kwargs:
+        if kwargs.get('root'):
             cmd.extend(['--root', kwargs['root']])
         cmd.extend(['-ql', pkg])
         out = __salt__['cmd.run'](
@@ -445,7 +445,7 @@ def owner(*paths, **kwargs):
     ret = {}
     for path in paths:
         cmd = ['rpm']
-        if 'root' in kwargs:
+        if kwargs.get('root'):
             cmd.extend(['--root', kwargs['root']])
         cmd.extend(['-qf', '--queryformat', '%{name}', path])
         ret[path] = __salt__['cmd.run_stdout'](cmd,
@@ -533,7 +533,7 @@ def info(*packages, **kwargs):
         size_tag = '%{SIZE}'
 
     cmd = ['rpm']
-    if 'root' in kwargs:
+    if kwargs.get('root'):
         cmd.extend(['--root', kwargs['root']])
     if packages:
         cmd.append('-q')
@@ -811,7 +811,7 @@ def checksum(*paths, **kwargs):
         raise CommandExecutionError("No package files has been specified.")
 
     cmd = ['rpm']
-    if 'root' in kwargs:
+    if kwargs.get('root'):
         cmd.extend(['--root', kwargs['root']])
     cmd.extend(['-K', '--quiet'])
     for package_file in paths:
--- a/salt/modules/rpm.py
+++ b/salt/modules/rpm.py
@@ -128,11 +128,14 @@ def bin_pkg_info(path, saltenv='base'):
     return ret
 
 
-def list_pkgs(*packages):
+def list_pkgs(*packages, **kwargs):
     '''
     List the packages currently installed in a dict::
 
         {'<package_name>': '<version>'}
+
+    root
+        use root as top level directory (default: "/")
 
     CLI Example:
 
@@ -141,8 +144,11 @@ def list_pkgs(*packages):
         salt '*' lowpkg.list_pkgs
     '''
     pkgs = {}
-    cmd = ['rpm', '-q' if packages else '-qa',
-           '--queryformat', r'%{NAME} %{VERSION}\n']
+    cmd = ['rpm']
+    if 'root' in kwargs:
+        cmd.extend(['--root', kwargs['root']])
+    cmd.extend(['-q' if packages else '-qa',
+                '--queryformat', r'%{NAME} %{VERSION}\n'])
     if packages:
         cmd.extend(packages)
     out = __salt__['cmd.run'](cmd, output_loglevel='trace', python_shell=False)
@@ -157,6 +163,9 @@ def list_pkgs(*packages):
 def verify(*packages, **kwargs):
     '''
     Runs an rpm -Va on a system, and returns the results in a dict
+
+    root
+        use root as top level directory (default: "/")
 
     Files with an attribute of config, doc, ghost, license or readme in the
     package header can be ignored using the ``ignore_types`` keyword argument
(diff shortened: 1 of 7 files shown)


salt/modules/rpm.py
```diff
@@ -338,10 +357,13 @@ def file_list(*packages):
...
     '''
+    cmd = ['rpm']
+    if 'root' in kwargs:
+        cmd.extend(['--root', kwargs['root']])
     if not packages:
-        cmd = ['rpm', '-qla']
+        cmd.append('-qla')
     else:
-        cmd = ['rpm', '-ql']
+        cmd.append('-ql')
```
isbm: A bit spaggettish, maybe we can always add `-ql` and only append another `-a` when no packages? Or so:

```python
cmd.append('-ql{}'.format(packages and 'a' or ''))
```

Or like you did later:

```python
cmd.append('-ql' if packages else '-qla')
```

salt/modules/zypper.py
```diff
@@ -2113,18 +2222,20 @@ def list_products(all=False, refresh=False):
...
     '''
     if refresh:
-        refresh_db()
+        refresh_db(root)
 
     ret = list()
-    OEM_PATH = "/var/lib/suseRegister/OEM"
+    oem_path = '/var/lib/suseRegister/OEM'
+    if root:
+        oem_path = os.path.join(root, oem_path)
```
isbm: Why constant became variable? This should be uppercase.

salt/modules/zypper.py
```diff
@@ -2206,22 +2322,27 @@ def download(*packages, **kwargs):
...
     CLI example:
 
     .. code-block:: bash
 
         salt '*' pkg.list_downloaded
     '''
-    CACHE_DIR = '/var/cache/zypp/packages/'
+    cache_dir = '/var/cache/zypp/packages/'
+    if root:
+        cache_dir = os.path.join(root, cache_dir)
```
isbm: Constant became variable. Please keep it upper-case.

salt/modules/zypper.py
```diff
@@ -2359,7 +2494,7 @@ def list_provides(**kwargs):
     return ret
 
 
-def resolve_capabilities(pkgs, refresh, **kwargs):
+def resolve_capabilities(pkgs, refresh=False, root=False, **kwargs):
```
isbm: Why sometimes `root` is `None` and sometimes `False`? This should be consistent.

isbm (changes requested):

salt/modules/zypper.py
isbm: @meaksh is that OK to default `refresh` here?

salt/modules/zypper.py
```diff
@@ -2206,22 +2322,27 @@ def download(*packages, **kwargs):
...
         salt '*' pkg.list_downloaded
     '''
-    CACHE_DIR = '/var/cache/zypp/packages/'
+    cache_dir = '/var/cache/zypp/packages/'
+    if root:
+        cache_dir = os.path.join(root, cache_dir)
 
     ret = {}
-    for root, dirnames, filenames in salt.utils.path.os_walk(CACHE_DIR):
+    for root, dirnames, filenames in salt.utils.path.os_walk(cache_dir):
```
isbm: Constant still variable...

isbm (changes requested):

salt/modules/zypper.py
aplanas: Looks like that github is a bit inconsistent today, but as commented I am changing CACHE_DIR, so is not a constant anymore. I would find confusing something like this:

```
def my_func():
  A = 10
  if cond():
    A = A + 5
```

I change A, but the reader do not expect that I do that based on the uppercase. Changing the the case will fix the expectation.

aplanas:
@isbm ready for a deeper review!

rallytime:
@aplanas There are some tests failing with this change. Can you take a look?

https://jenkinsci.saltstack.com/job/pr-kitchen-centos7-py2/job/PR-50125/12/

aplanas:
@rallytime indeed, thanks

salt/modules/rpm.py
```diff
@@ -128,21 +128,27 @@ def bin_pkg_info(path, saltenv='base'):
...
     .. code-block:: bash
 
         salt '*' lowpkg.list_pkgs
     '''
     pkgs = {}
-    cmd = ['rpm', '-q' if packages else '-qa',
-           '--queryformat', r'%{NAME} %{VERSION}\n']
+    cmd = ['rpm']
+    if 'root' in kwargs:
+        cmd.extend(['--root', kwargs['root']])
```
isbm: This is bug. Correctly would be:
```python
if kwargs.get('root'):
   ...
```

salt/modules/rpm.py
```diff
@@ -199,6 +208,8 @@ def verify(*packages, **kwargs):
             verify_options = [x.strip() for x in six.text_type(verify_options).split(',')]
 
     cmd = ['rpm']
+    if 'root' in kwargs:
```
isbm: Same bug.

salt/modules/rpm.py
```diff
@@ -266,10 +280,12 @@ def modified(*packages, **flags):
         salt '*' lowpkg.modified httpd postfix
         salt '*' lowpkg.modified
     '''
-    ret = __salt__['cmd.run_all'](
-        ['rpm', '-Va'] + list(packages),
-        output_loglevel='trace',
-        python_shell=False)
+    cmd = ['rpm']
+    if 'root' in flags:
```
isbm: Bug. Check for root _value_.

salt/modules/rpm.py
```diff
@@ -338,25 +357,31 @@ def file_list(*packages):
         salt '*' lowpkg.file_list httpd postfix
         salt '*' lowpkg.file_list
     '''
-    if not packages:
-        cmd = ['rpm', '-qla']
-    else:
-        cmd = ['rpm', '-ql']
+    cmd = ['rpm']
+    if 'root' in kwargs:
```
isbm: Here too.

isbm (changes requested):
@aplanas OK, I stopped, but there are more. Please first fix these. :wink:

salt/modules/rpm.py
aplanas: Oh, you are referring to the line up. I will change. But still I do not see the error. Is there a case where __getitem__ is not defined for kwargs?

aplanas: Ah nono. I get it now, the goal is to make sure that is not None , '' or 0. But this pattern can exclude 0, and sometimes is a valid value.

aplanas:
@isbm I am done with the 'in's. Thanks!

salt/modules/zypper.py
```diff
@@ -906,23 +950,25 @@ def _is_match(pkgname):
...
+def _get_configured_repos(root=None):
     '''
     Get all the info about repositories from the configurations.
     '''
 
+    repos = os.path.join(root, os.path.relpath(REPOS, start='/')) if root else REPOS
     repos_cfg = configparser.ConfigParser()
-    repos_cfg.read([REPOS + '/' + fname for fname in os.listdir(REPOS) if fname.endswith(".repo")])
+    if os.path.exists(repos):
+        repos_cfg.read([repos + '/' + fname for fname in os.listdir(repos) if fname.endswith(".repo")])
```
isbm: Add error logging here if `repos != REPOS` and not found.

salt/modules/zypper.py
```diff
@@ -1029,7 +1087,8 @@ def mod_repo(repo, **kwargs):
         salt '*' pkg.mod_repo alias url= mirrorlist=http://host.com/
     '''
 
-    repos_cfg = _get_configured_repos()
+    root = kwargs.get('root', None)
+    repos_cfg = _get_configured_repos(root=root)
```
isbm: This is a bug. 
```python
repos_cfg = _get_configured_repos(root=kwargs.get('root') or None)
```

salt/modules/zypper.py
```diff
@@ -1702,8 +1796,9 @@ def list_locks():
         salt '*' pkg.list_locks
     '''
     locks = {}
-    if os.path.exists(LOCKS):
-        with salt.utils.files.fopen(LOCKS) as fhr:
+    _locks = os.path.join(root, os.path.relpath(LOCKS, start='/')) if root else LOCKS
+    if os.path.exists(_locks):
+        with salt.utils.files.fopen(_locks) as fhr:
```
isbm: I'd rather try/except it instead of stat file twice.

salt/modules/zypper.py
```diff
@@ -1729,10 +1827,11 @@ def clean_locks():
     '''
     LCK = "removed"
     out = {LCK: 0}
-    if not os.path.exists("/etc/zypp/locks"):
+    locks = os.path.join(root, os.path.relpath(LOCKS, start='/')) if root else LOCKS
```
isbm: Technically, `os.path.sep`, not `'/'`. :wink:

isbm (changes requested):

salt/modules/zypper.py
aplanas: indeed!

salt/modules/zypper.py
```diff
@@ -127,18 +127,28 @@ def _reset(self):
...
+        # Reset after the call
+        if self.__called:
+            self._reset()
+
         # Ignore exit code for 106 (repo is not available)
         if 'no_repo_failure' in kwargs:
             self.__ignore_repo_failure = kwargs['no_repo_failure']
         if 'systemd_scope' in kwargs:
             self.__systemd_scope = kwargs['systemd_scope']
+        if 'root' in kwargs:
```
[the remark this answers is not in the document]
aplanas: On a second though, are you sure? This check if to assign self.__root later in the same fashion that 'no_repo_failure' and 'systemd_scope'

salt/modules/zypper.py
aplanas: I need to assign to root, as is used later. Also I am not able to see the bug, as root will be None in case that is not there.

I changed the code to some proposal that was before, also I am not able to see the bug.

salt/modules/zypper.py
aplanas: done, complicate the code a bit

aplanas:
@isbm any more reviews? Any thing that I can do?

salt/modules/zypper.py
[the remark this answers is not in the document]
isbm: Well, OK, copypaste. Maybe let's to as is. But I still think the other parts also needs to be type-safe.

isbm (approved):
@aplanas nope, good to go. :wink:

isbm (approved):

salt/modules/zypper.py
```diff
@@ -1186,6 +1250,9 @@ def install(name=None,
             skip_verify=False,
             version=None,
             ignore_repo_failure=False,
+            no_recommends=False,
+            auto_agree_with_licenses=False,
```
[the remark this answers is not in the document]
aplanas: Ah yes. Comes from a different place. Fixed.

aplanas:
This PR is reaching the 1 month old mark. If there anything that I can do the help here?

The code do not change the default behavior, as I did not change the semantic of the tests.

rallytime:
@terminalmage and @meaksh How does this look to you now?

terminalmage (approved):

aplanas:
thanks a lot for the reviews!

PR #50126, opened 2018-10-19
Fix lowpkg.diff documentation and parameter name

### LINKED ISSUE #50097: lowpkg.diff refers to a package name, but a package file is required
`lowpkg.diff` (rpm.py) have a parameter named `package` that refers to a package name (apache2, xz, etc), but internally is using `rpm2cpio` to see the content of the package. The problem is that this command requite a RPM file.

This can be a documentation issue, that affect the examples and all system dependent package managers too.

If is a documentation issue I can fix it, but first I would like to have a confirmation that the expectations are not wrong.

### Versions Report
develop branch

aplanas:
Package names are useless for lowpkg.diff, as to make a full
comparison we need to access the RPM

Fix #50097

### DIFF
--- a/salt/modules/rpm.py
+++ b/salt/modules/rpm.py
@@ -425,13 +425,13 @@ def owner(*paths):
 @salt.utils.decorators.path.which('rpm2cpio')
 @salt.utils.decorators.path.which('cpio')
 @salt.utils.decorators.path.which('diff')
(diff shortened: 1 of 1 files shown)


cachedout:
@isbm Can you please help review this?

isbm (approved):

Issue #50127, opened 2018-10-19
"NameError: name 'reload' is not defined" on python 3


mgomersbach:
### Description of Issue/Question
Running Salt under py3 gives a NameError :
```
An exception occurred in this state: Traceback (most recent call last):
                File "/usr/lib64/python3.6/site-packages/salt/state.py", line 1913, in call
                  **cdata['kwargs'])
                File "/usr/lib64/python3.6/site-packages/salt/loader.py", line 1898, in wrapper
                  return f(*args, **kwargs)
                File "/usr/lib64/python3.6/site-packages/salt/states/pkg.py", line 1548, in installed
                  **kwargs)
                File "/usr/lib64/python3.6/site-packages/salt/states/pkg.py", line 521, in _find_install_targets
                  cur_pkgs = __salt__['pkg.list_pkgs'](versions_as_list=True, **kwargs)
                File "/usr/lib64/python3.6/site-packages/salt/modules/ebuild.py", line 426, in list_pkgs
                  pkgs = _vartree().dbapi.cpv_all()
                File "/usr/lib64/python3.6/site-packages/salt/modules/ebuild.py", line 68, in _vartree
                  portage = reload(portage)
              NameError: name 'reload' is not defined
```

"reload" is a Python 2 builtin and not available under Python 3.
Going through the code I could find 3 occurrences:

modules/ebuild.py:68:    portage = reload(portage)
modules/ebuild.py:74:    portage = reload(portage)
utils/virtualbox.py:145:        reload(vboxapi)

### Versions Report
```
Salt Version:
           Salt: 2018.3.0-678-g84e2dcf
 
Dependency Versions:
           cffi: 1.11.5
       cherrypy: Not Installed
       dateutil: 2.7.3
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.10
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: 1.0.6
   msgpack-pure: Not Installed
 msgpack-python: 0.5.1
   mysql-python: Not Installed
      pycparser: 2.14
       pycrypto: Not Installed
   pycryptodome: Not Installed
         pygit2: Not Installed
         Python: 3.6.6 (default, Sep 11 2018, 11:42:02)
   python-gnupg: Not Installed
         PyYAML: 3.13
          PyZMQ: 16.0.2
           RAET: Not Installed
[...]

PR #50128, opened 2018-10-19
Workaround for py2 builtin, =<3.3 imp and >=3.4 libimport quirks

mgomersbach:
### What does this PR do?
Python 2 has a builtin for "reload"
Python lower than 3.3 has the "imp" module
And Python 3.4 and up switched to "importlib"
This patches to fall back through the different implementations

solves #50127

### Previous Behavior
NameError: name 'reload' is not defined

### New Behavior
Works

### Tests written?

No

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/modules/ebuild.py
+++ b/salt/modules/ebuild.py
@@ -61,13 +61,29 @@ def __virtual__():
 
 def _vartree():
     import portage  # pylint: disable=3rd-party-module-not-gated
-    portage = reload(portage)
+    try:
+        reload(portage)
+    except NameError:
+        try:
+            from importlib import reload
+            reload(portage)
+        except ImportError:
+            from imp import reload
+            reload(portage)
     return portage.db[portage.root]['vartree']
 
 
 def _porttree():
     import portage  # pylint: disable=3rd-party-module-not-gated
-    portage = reload(portage)
+    try:
+        reload(portage)
+    except NameError:
+        try:
+            from importlib import reload
+            reload(portage)
+        except ImportError:
+            from imp import reload
+            reload(portage)
     return portage.db[portage.root]['porttree']
 
 
--- a/salt/modules/ebuild.py
+++ b/salt/modules/ebuild.py
@@ -69,13 +69,13 @@ def __virtual__():
 
 def _vartree():
     import portage  # pylint: disable=3rd-party-module-not-gated
-    reload(portage)
+    portage = reload(portage)
     return portage.db[portage.root]['vartree']
 
 
 def _porttree():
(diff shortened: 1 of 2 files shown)


gtmanfred (changes requested):
Can these not be used as imports at the top of the file?

Check if `reload` is in globals(), if it is not, then import it with the ImportError exception?

mgomersbach:
Yeah my reasoning was to not evaluate it every time the module was called.
I'll change the PR!

mgomersbach:
The test result seems a fluke? Can it be retriggered?

gtmanfred:
Those failures are unrelated.

gtmanfred (changes requested):
Can we do something more like.

```
if not hasattr(globals()['__builtins__'], 'reload'):
    try:
        from importlib import reload
    except ImportError:
        from imp import reload
```

Also, have you verified that the builtins work on both python2 and py3? i know the handling of builtins was changed in python3

mgomersbach:
Yes 2.7, <=3.3 and >=3.4 have been tested for that code. (does salt run on <=3.3?, because we can then simplify it even more)

gtmanfred:
ok, perfect, then i would still ask that it be changed to be in an if statement instead of doing to try blocks.

mgomersbach:
Will do when I arrive at my destination!
Final question, does saltstack run on python 3.3 anywhere because I can leave out the except then

gtmanfred:
We do not run on 3.3, we support 2.7 and >= 3.4

gtmanfred:
But I do believe we use imp instead of importlib on python3.4 for our loader, so that would probably be good to use there.

gtmanfred:
Also, probably better to just check the python version instead of using try except blocks.

mgomersbach:
imp is deprecated since 3.4.
Version checking would exclude any other versioning from alternative pythons (unless I make a big if else tree).
When I have access to a desktop I will implement your suggestions

rallytime:
@mgomersbach Since we only support python == 2.7 and python >= 3.4, it shouldn't be too big of a tree. You can use the `six.PY2` and `six.PY3` variables, available from `salt.ext.six`.

gtmanfred:
if you want to just disable that check for those two lines, that should be good enough, since you are explicitly only running the import on versions where that import is there.

pylint runs on python2, so it doesn't know that it is wrong.

rallytime (approved):

rallytime:
@mgomersbach You can if you'd like to, but you don't have to. Let me know what you want to do.

mgomersbach:
Then let's not. Comments provide enough context I guess.

Issue #49957, opened 2018-10-09
salt-master scheduled items ignore maxrunning.


whytewolf:
### Description of Issue/Question
scheduled items in the master config ignore the maxrunning setting. So can have more than a single running schedule item at a time.

### Setup
setup a schedule on a master that ruins a task longer than the time it is scheduled for with a maxrunning setting, or using the default of 1. Wait and than parse ps for running tasks and see that more than one `/usr/bin/salt-master Maintenance Schedule <jid>` is running. setup setproctitle to know what processes are running.

### Steps to Reproduce Issue
Here is what I used for my configuration and test.

```
[root@salt00 test_orch]# cat /srv/salt/test_orch/test.sls
test_wait:
  salt.function:
    - tgt: '*'
    - name: test.sleep
    - arg:
      - 300
```

```
[root@salt00 test_orch]# cat /etc/salt/master.d/schedule.conf
schedule:
  test_sch:
    function: state.orch
    minutes: 1
    args:
      - test_orch.test
```

And the result

```
root     23908     1 23908  1    1 20:38 ?        00:00:03 /usr/bin/python /usr/bin/salt-master Maintenance Schedule 20181009203836745635
root     23974     1 23974  0    1 20:38 ?        00:00:00 /usr/bin/python /usr/bin/salt-minion KeepAlive MultiMinionProcessManager MinionProcessManager Minion._thread_return 20181009203837752085
root     24308     1 24308  1    1 20:39 ?        00:00:02 /usr/bin/python /usr/bin/salt-master Maintenance Schedule 20181009203938188486
root     24374     1 24374  0    1 20:39 ?        00:00:00 /usr/bin/python /usr/bin/salt-minion KeepAlive MultiMinionProcessManager MinionProcessManager Minion._thread_return 20181009203939210343
root     24710     1 24710  1    1 20:40 ?        00:00:01 /usr/bin/python /usr/bin/salt-master Maintenance Schedule 20181009204039865523
root     24776     1 24776  0    1 20:40 ?        00:00:00 /usr/bin/python /usr/bin/salt-minion KeepAlive MultiMinionProcessManager MinionProcessManager Minion._thread_return 20181009204040859635
root     25125     1 25125  1    1 20:41 ?        00:00:00 /usr/bin/python /usr/bin/salt-master Maintenance Schedule 20181009204141665150
[...]

PR #50130, opened 2018-10-19
[2018.3] Fixes to schedule maxrunning on master

garethgreenaway:
### What does this PR do?
Adding some master specific functions to uitls/masters.py to determine if a Salt process is running.  Updating utils/schedule.py to use the appropriate running function either from utils/master.py or utils/minion.py depending on where the scheduled job is running.  Adding tests to test maxrunning in scheduled jobs for both the minion and master.

#49957 

### Tests written?
Yes

### Commits signed with GPG?

Yes/No

### DIFF
--- a/salt/utils/master.py
+++ b/salt/utils/master.py
@@ -39,7 +39,7 @@ from salt.utils.zeromq import zmq
 log = logging.getLogger(__name__)
 
 
-def running(opts):
+def get_running_jobs(opts):
     '''
     Return the running jobs on this minion
     '''
@@ -56,9 +56,9 @@ def running(opts):
                 ret.append(data)
         except (IOError, OSError):
             # proc files may be removed at any time during this process by
-            # the minion process that is executing the JID in question, so
+            # the master process that is executing the JID in question, so
             # we must ignore ENOENT during this process
-            pass
+            log.trace('%s removed during processing by master process', path)
     return ret
 
 
@@ -126,8 +126,7 @@ def _check_cmdline(data):
         return False
     try:
         with salt.utils.files.fopen(path, 'rb') as fp_:
-            if b'salt' in fp_.read():
-                return True
+            return b'salt' in fp_.read()
     except (OSError, IOError):
         return False
 
--- a/salt/utils/master.py
+++ b/salt/utils/master.py
@@ -62,24 +62,6 @@ def running(opts):
     return ret
 
 
-def cache_jobs(opts, jid, ret):
-    '''
-    Write job information to cache
-    '''
-    serial = salt.payload.Serial(opts=opts)
-
-    fn_ = os.path.join(
-        opts['cachedir'],
-        'minion_jobs',
-        jid,
-        'return.p')
-    jdir = os.path.dirname(fn_)
-    if not os.path.isdir(jdir):
-        os.makedirs(jdir)
-    with salt.utils.files.fopen(fn_, 'w+b') as fp_:
-        fp_.write(serial.dumps(ret))
-
-
 def _read_proc_file(path, opts):
     '''
     Return a dict of JID metadata, or None
--- a/salt/utils/master.py
+++ b/salt/utils/master.py
@@ -63,6 +63,9 @@ def running(opts):
 
 
 def cache_jobs(opts, jid, ret):
+    '''
+    Write job information to cache
+    '''
     serial = salt.payload.Serial(opts=opts)
 
     fn_ = os.path.join(
@@ -92,7 +95,7 @@ def _read_proc_file(path, opts):
             try:
                 os.remove(path)
             except IOError:
-                pass
+                log.debug('Unable to remove proc file %s.', path)
             return None
     if not isinstance(data, dict):
         # Invalid serial object
@@ -103,7 +106,7 @@ def _read_proc_file(path, opts):
         try:
             os.remove(path)
         except IOError:
-            pass
+            log.debug('Unable to remove proc file %s.', path)
         return None
 
     if not _check_cmdline(data):
@@ -115,7 +118,7 @@ def _read_proc_file(path, opts):
         try:
             os.remove(path)
         except IOError:
-            pass
+            log.debug('Unable to remove proc file %s.', path)
         return None
     return data
 
--- a/salt/utils/master.py
+++ b/salt/utils/master.py
@@ -39,6 +39,114 @@ from salt.utils.zeromq import zmq
 log = logging.getLogger(__name__)
 
 
+def running(opts):
+    '''
+    Return the running jobs on this minion
+    '''
+
+    ret = []
+    proc_dir = os.path.join(opts['cachedir'], 'proc')
+    if not os.path.isdir(proc_dir):
+        return ret
+    for fn_ in os.listdir(proc_dir):
+        path = os.path.join(proc_dir, fn_)
+        try:
+            data = _read_proc_file(path, opts)
+            if data is not None:
+                ret.append(data)
+        except (IOError, OSError):
+            # proc files may be removed at any time during this process by
+            # the minion process that is executing the JID in question, so
+            # we must ignore ENOENT during this process
+            pass
+    return ret
+
+
+def cache_jobs(opts, jid, ret):
+    serial = salt.payload.Serial(opts=opts)
+
+    fn_ = os.path.join(
+        opts['cachedir'],
+        'minion_jobs',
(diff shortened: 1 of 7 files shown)


terminalmage:
@garethgreenaway Lint issues in the tests: https://jenkinsci.saltstack.com/job/pr-lint/job/PR-50130/1/warnings52Result/new/

terminalmage (approved):

gtmanfred (approved):

salt/utils/master.py
```diff
@@ -39,6 +39,114 @@
...
+    '''
+    In some cases where there are an insane number of processes being created
+    on a system a PID can get recycled or assigned to a non-Salt process.
+    On Linux this fn checks to make sure the PID we are checking on is actually
+    a Salt process.
+
+    For non-Linux systems we punt and just return True
+    '''
+    if not salt.utils.platform.is_linux():
+        return True
```
isbm: If I do not missing anything, then this is weird. First of all, why would you ban all other OSes that supports procfs? E.g. what's wrong with AIX or Solaris? Second, why would this function confirm process "is Salt" if the OS is not Linux at all?? This doesn't make sense.

salt/utils/schedule.py
```diff
@@ -173,7 +174,11 @@ def _check_max_running(self, func, data, opts):
         data['run'] = True
         if 'jid_include' not in data or data['jid_include']:
             jobcount = 0
-            for job in salt.utils.minion.running(self.opts):
+            if self.opts['__role'] == 'master':
+                current_jobs = salt.utils.master.running(self.opts)
+            else:
+                current_jobs = salt.utils.minion.running(self.opts)
```
isbm: This is not really nice approach, since e.g. Syndic is out here. Why not support structure like this:
```python
role = self.opts['__role']
if hasattr(self.utils, role):
    current_jobs = getattr(self.utils, role).running(self.opts)
```
Of course, the "role" should point in the object inside `salt/utils/__init__.py` then. But in that case everyone will easily add not yet implemented roles, for the same Syndic. Now you are just hard-coding things.

salt/utils/master.py
```diff
@@ -39,6 +39,114 @@
...
+        return False
+    if not os.path.isdir('/proc'):
+        return True
+    path = os.path.join('/proc/{0}/cmdline'.format(pid))
+    if not os.path.isfile(path):
+        return False
+    try:
+        with salt.utils.files.fopen(path, 'rb') as fp_:
+            if b'salt' in fp_.read():
+                return True
```
isbm: You could just return:
```python
return b'salt' in fp_.read()
```
But I still don't think this is correct way of "detecting" process.

salt/utils/master.py
```diff
@@ -39,6 +39,114 @@
...
+    serial = salt.payload.Serial(opts)
+    with salt.utils.files.fopen(path, 'rb') as fp_:
+        buf = fp_.read()
+        fp_.close()
+        if buf:
+            data = serial.loads(buf)
+        else:
+            # Proc file is empty, remove
+            try:
+                os.remove(path)
```
isbm: Should _this_ function take care of cleanup?

salt/utils/master.py
```diff
@@ -39,6 +39,114 @@
...
+        pid = data.get('pid')
+        if pid:
+            log.warning(
+                'PID %s exists but does not appear to be a salt process.', pid
+            )
+        try:
+            os.remove(path)
+        except IOError:
+            pass
+        return None
```
isbm: This is where it gets scattered. You are agree with cmdline for _any_ command that contains `salt` string, as well as it is not even Linux. This should not be done that way. This should say `True` only if the command line really is Salt command. But then you cannot guarantee it either, because you're banning all OSes that aren't Linux but still implements procfs.

isbm (changes requested):
I am not sure I like the whole approach and design of this code. And it is not portable. :disappointed:

isbm:
What I would propose here is different approach at this problem. Since our task here is to let `maxrunning` to be respected. Therefore the whole purpose of _this_ particular routine is to properly count maximum parallel running jobs and nothing else. To this, I would:

1. Avoid approach heuristically to figure out "what could be the process"? Instead I would solve that from the other direction: each job should be already prepared with the required in a future info/metadata about it and only _then_ fired to perform. That protocol might be very small, like just one byte attached to the data of job on boolean basis. This way gives us the following advantages:

    - No longer need to filter supported OSes. The whole thing is portable.
    - It is now guaranteed to know the kind of job/process.
    - More maintainable and structured code: it does only what it supposed to, being located in one place.

2. Remove any kind of PID cleanup helpers. This is simply another task, another place to do this. Either job scheduler should properly do it (again, the data from No.1 could be happily reused!) or job itself can "kill/cleanup" its post-mortem data, as that is done with typical threading/multiprocessing based on mutexes/signals. All that PIDs cleaning should be there, and be completely separate logical flow, thus:

    - Much easier to locate bugs, if any
    - No overlapping "spaghetti" code

So the **conceptually** in **pseudo-code** I would express the solution idea and its interface in the following way. I personally would prefer it as a class, so one can instantiate it and start working with the jobs. Or just refer an instance at module level (YMMV). It is so:

```python
class JobManagerOrTracker(object):
    def register(self, job):
        '''
        This registers job and works with the cache/serialiser.
        '''
        ...
        return True or False  # Registered or not

    def start_job(self, payload):
        '''
        We just start one job and register it.
        '''
        command, other_things = payload.parse_in_some_way()
        job = Job()  # Wrapper to a job in cache.
                     # Can be just a dict as well,
                     # but objects just more convenient.
        job.command = command
        job.is_parallel = other_things.get_is_countable()  # This is conceptual idea
        self.register(job)  # This job is not yet started!

    def get_all_running_jobs(self):
        '''
        This just reads what is in cache alongside
        with the whole metadata.
        '''
        # At this point we no longer care if this is
        # Windows or Linux or Solaris.
        ...
        return available_jobs

    def running_jobs(self):
        '''
        Get only jobs that we respect for maxrunning
        '''
        jobs_we_need = []
        for job in self.get_all_running_jobs():
            if job.we_are_interested_in():  # This is where we use beforehand prepared info
                jobs_we_need.append(job)

        return jobs_we_need

    def cleanup_jobs(self):
        '''
        Cleanup jobs.
        '''
        # This internally would get all running jobs,
        # see if job is actually running and cleanup PIDs or
        # leave as is.
        for job in self.get_all_running_jobs():
            if not job.is_running():
                job.cleanup()  # This Job() instance would remove pid
                               # check what is needed
                               # clean cache etc
```

salt/utils/schedule.py
isbm: :+1:

cachedout:
@garethgreenaway Review comments?

rallytime:
@garethgreenaway You're going to need to update [filename_map](https://github.com/saltstack/salt/blob/develop/tests/filename_map.yml) for this one. That is why that test is failing.

garethgreenaway:
@rallytime I'll take a look.

salt/utils/master.py
garethgreenaway: This is code that is only used by the Salt master, which is only supported on Linux based operating systems.  If/when that changes in the future, we'll definitely revisit this code so it's supported on other OSes.

cachedout (commented):
Why are there changes to doc/conf.py in here? Is that intentional?

garethgreenaway:
@cachedout Definitely not.  Fixing that up.

salt/utils/master.py
```diff
@@ -39,6 +39,98 @@
...
 
 
+def get_running_jobs(opts):
+    '''
+    Return the running jobs on this minion
+    '''
+
+    ret = []
+    proc_dir = os.path.join(opts['cachedir'], 'proc')
+    if not os.path.isdir(proc_dir):
```
cachedout: Wouldn't this be considered an error? I'm wondering if we should log this at at least info level if not error?

garethgreenaway: Perhaps a case where no jobs have run and the directory doesn't exist yet?

cachedout:
@isbm Are you all right with this going in as it sits? If so, could you please approve it? Thanks.

isbm (approved):

isbm:
@cachedout it still figures out running processes, instead of pre-registering them. But since you are asking to approve, I am approving.

cachedout:
@garethgreenaway There are test failures here.

PR #50131, opened 2018-10-19
cache signing file data

gtmanfred:
### What does this PR do?
This will make it so that we don't have to read from the disk every time
that a key reports, only if the mtime has changed.

### Previous Behavior
Every time a new minion key comes in, we read the autosign files from disk

### New Behavior
Only read the autosign files from disk if the modify time has changed.

### Tests written?

Existing tests for autosign files.

### Commits signed with GPG?

Yes


salt/daemons/masterapi.py
```diff
@@ -314,14 +315,16 @@ def check_signing_file(self, keyid, signing_file):
...
-                    continue
-                else:
-                    if salt.utils.stringutils.expr_match(keyid, line):
-                        return True
+        mtime = os.path.getmtime(signing_file)
+        if self.signing_file.get(signing_file, {}).get('mtime') != mtime:
+            self.signing_files.setdefault(signing_file, {})['mtime'] = mtime
+            with salt.utils.files.fopen(signing_file, 'r') as fp_:
+                self.signing_files[signing_file]['data'] = [
+                    line for line in [line.strip() for line in fp_] if not line.strip().startswith('#')
```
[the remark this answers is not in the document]
gtmanfred: :+1: good point.

yeah, this works fine in python3, i will double check python2

salt/daemons/masterapi.py
```diff
@@ -314,14 +315,16 @@ def check_signing_file(self, keyid, signing_file):
...
+                ]
+        if any(salt.utils.stringutils.expr_match(keyid, line) for line
+               in salt.signing_files[signing_file].get('data', [])):
+            return True
```
isbm:
```suggestion
            return bool(any(salt.utils.stringutils.expr_match(keyid, line) for line
                        in salt.signing_files[signing_file].get('data', [])))
```

isbm (changes requested):

salt/daemons/masterapi.py
terminalmage: Yes, good idea, except `any()` already returns a bool, so wrapping it in `bool()` is redundant.

terminalmage (approved):

salt/daemons/masterapi.py
isbm: Right...

rallytime:
@isbm bump :)

isbm (approved):
@rallytime hmm, wondering how did I missed for 19 days request from @cachedout ... Sorry holding this!

Issue #50067, opened 2018-10-15
Checking for multiple status with http.query state


DrPyser:
### Description of Issue/Question
When using the `http.query` state, I would need for the state to match multiple status. That is, I expect a successful response to be either a status 201 or 200. Is there a way to do that currently? If not, there are multiple ways to add support for this use case:
* Allow `status` to be a list of accepted values
* Make `match_type` also apply to the status matching, so that one could use a regular expression to match the status.

I prefer the second option, more flexible. 
 
### Setup
(Please provide relevant configs and/or SLS files (Be sure to remove sensitive info).)

```
{% set my_cert = salt['cmd.shell']('cat '+ pillar['my_cert_path']) %}
{% set my_cert_key = salt['cmd.shell']('cat '+ pillar['my_cert_key_path']) %}


kong_admin_api_available:
  http.query:
    - name: {{ pillar.get("kong_admin_url", "https://example.domain.com:8444")}}/
    - method: OPTIONS
    - status: 204

upload_certificate:
  http.query:
    - name: {{ pillar.get("kong_admin_url", "https://example.domain.com:8444")}}/certificates/{{ pillar["kong_certificate_hostname"] }}
    - method: PUT
    - data: |
        { "key": {{ my_cert_key|json }}, "cert": {{ my_cert|json }}}
    - header_dict:
        Content-Type: application/json
    - match_type: pcre
    - status: "(200)|(201)"
    - require:
        - kong_admin_api_available
```
Since the semantics of the `PUT` call is "create or update", the call might return a 201 on a first request and 200 on the nexts.

### Versions Report
<details><p>
```
Salt Version:                                                                                                                                              
           Salt: 2018.3.2                                                                                                                                  
                                                                                                                                                           
Dependency Versions:                                                                                                                                       
           cffi: Not Installed                                                                                                                             
       cherrypy: Not Installed                                                                                                                             
       dateutil: 2.6.1                                                                                                                                     
      docker-py: 2.5.1                                                                                                                                     
          gitdb: 2.0.3                                                                                                                                     
      gitpython: 2.1.8                                                                                                                                     
          ioflo: Not Installed                                                                                                                             
         Jinja2: 2.10                                                                                                                                      
        libgit2: Not Installed                                                                                                                             
        libnacl: Not Installed                                                                                                                             
       M2Crypto: Not Installed                                                                                                                             
           Mako: 1.0.7                                                                                                                                     
   msgpack-pure: Not Installed                                                                                                                             
[...]

PR #50150, opened 2018-10-21
Support regex status types in http.query state

mchugh19:
### What does this PR do?
Adds `status_type` to http.query state to support pcre expressions

#50067

### New Behavior
Supports states like the following (note `status_type` and `status`)
```
{% set statuses = [200, 201] %}
{% for status in statuses %}
verifty_status_{{ status }}:
  http.query:
    - name: http://the-internet.herokuapp.com/status_codes/{{ status }}
    - status: 200|201
    - status_type: pcre
    - match: "This page returned a"
{% endfor %}
```

### Tests written?
Yes

### Commits signed with GPG?
No

### DIFF
--- a/salt/states/http.py
+++ b/salt/states/http.py
@@ -20,7 +20,7 @@ __monitor__ = [
 log = logging.getLogger(__name__)
 
 
-def query(name, match=None, match_type='string', status=None, wait_for=None, **kwargs):
+def query(name, match=None, match_type='string', status=None, status_type='string', wait_for=None, **kwargs):
     '''
     Perform an HTTP query and statefully return the result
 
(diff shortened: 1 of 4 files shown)


salt/states/http.py
```diff
@@ -94,28 +107,40 @@ def query(name, match=None, match_type='string', status=None, wait_for=None, **k
...
-            ret['result'] = False
+        if status_type == 'string':
+            if data.get('status', '') == str(status):
+                ret['comment'] += ' Status {0} was found.'.format(status)
+                if ret['result'] is None:
+                    ret['result'] = True
+            else:
+                ret['comment'] += ' Status {0} was not found.'.format(status)
+                ret['result'] = False
+        elif status_type == 'pcre':
```
cachedout: Can we combine this with the check for the `string` status type? It looks like the blocks are basically identical.

cachedout (approved):

cachedout (approved):

Issue #50025, opened 2018-10-13
Default values in master config file - missing blank lines


rjc:
This report has been prompted by a [user comment](https://groups.google.com/d/msg/salt-users/wP5r2nStBtI/TMripp5sAgAJ).

The very top of the `/etc/salt/master` config file reads:

    ##### Primary configuration settings #####
    ##########################################
    # This configuration file is used to manage the behavior of the Salt Master.
    # Values that are commented out but have an empty line after the comment are
    # defaults that do not need to be set in the config. If there is no blank line
    # after the comment then the value is presented as an example and is not the
    # default.

However, as mentioned there, some defaults do **not** have an empty line immediately following them, i.e.:

    #file_roots:
    #  base:
    #    - /srv/salt
    #

or:

    #pillar_roots:
    #  base:
    #    - /srv/pillar
    #

The reverse is also true, i.e.:

    #file_ignore_regex:
    #  - '/\.svn($|/)'
    #  - '/\.git($|/)'
    

or:

    #ext_pillar:
    #  - hiera: /etc/hiera.yaml
    #  - cmd_yaml: cat /etc/salt/yaml
    

It is quite confusing as, despite a very clear comment at the beginning of the config file, one cannot be sure whether a particular value is the default or not.

Do you have any means of auto-generating the `master` config file with the defaults correctly followed by a blank line?

If not, please let me know and I'll provide a patch.

Ch3LL:
i dont believe this is auto generated as its directly in the code in conf/master, but @saltstack/team-core might be able to correct me if i'm wrong.

gtmanfred:
Yeah, this is not automatically created, the file is manually made in the source, so it can just be edited.

rjc:
The same information is included in said file, the `salt(7)` manual page, other docs.

Does that mean that **all** of them need to be updated individually?

Is any of them the canonical one from which all of the others are updated?

Better still, is there a way to have only one copy of the information and the remaining documentations "slurps up" the canonical version, i.e. in the style of `#include <file>`, etc.?

gtmanfred:
no, just the conf/<configfile> needs to be updated, the man 7 salt is autogenerated based on  the other files.

rjc:
I believe I covered everything in #50164, when it comes to `conf/master`.

rjc:
The fix has now been now been merged.

PR #50164, opened 2018-10-23
Fix spacing

rjc:
Ensure that:

- there is an empty line after a default commented out option
- default values are shown where appropriate
- sections are separated by two empty lines
- there is no space between the comment (#) sign and the commented out option
- there is a single space between the comment (#) sign and the comment text itself
- code indentation is correct

### Previous Behavior

### New Behavior

### Tests written?

Yes/No

### Commits signed with GPG?

Yes/No


conf/master
```diff
@@ -181,8 +184,9 @@
 #ipc_mode: ipc
 
 # Overwrite the default tcp ports used by the minion when ipc_mode is set to 'tcp'
-#tcp_master_pub_port: 4510
-#tcp_master_pull_port: 4511
+#tcp_master_pub_port: 4512
+
```
gtmanfred: i don't like adding this newline

conf/master
```diff
@@ -294,7 +299,8 @@
 # buffering.
 # Value: In bytes. Set to 'dynamic' to have Salt select
 # a value for you. Default is disabled.
-# ipc_write_buffer: 'dynamic'
+#ipc_write_buffer: 'dynamic'
+#
```
gtmanfred: why is this added, it seems inconsistent.

conf/master
```diff
@@ -303,11 +309,13 @@
 # of 8 will be used. If no batch_safe_limit is specified, then no automatic
 # batching will occur.
 #batch_safe_limit: 100
+#
 #batch_safe_size: 8
 
 # Master stats enables stats events to be fired from the master at close
 # to the defined interval
 #master_stats: False
+
```
gtmanfred: don't need the extra line

conf/master
```diff
@@ -303,11 +309,13 @@
 # of 8 will be used. If no batch_safe_limit is specified, then no automatic
 # batching will occur.
 #batch_safe_limit: 100
+#
```
gtmanfred: dont need this

conf/master
```diff
@@ -318,17 +326,18 @@
...
 # query is as follows.
-# masterkeyring:
+#masterkeyring:
 #  driver: keyring
 #  service: system
 #
-# key_pass: sdb://masterkeyring/key_pass
-
+#key_pass: sdb://masterkeyring/key_pass
+#
```
gtmanfred: why is this added?

conf/master
```diff
@@ -318,17 +326,18 @@
...
 # Enable passphrase protection of the Master signing_key. This only applies if
 # master_sign_pubkey is set to True.  This is disabled by default.
-# master_sign_pubkey: True
-# signing_key_pass: sdb://masterkeyring/signing_pass
+#master_sign_pubkey: False
```
gtmanfred: why are we adding this newline between

conf/master
```diff
@@ -318,17 +326,18 @@
...
 
+#signing_key_pass: sdb://masterkeyring/signing_pass
+#
```
gtmanfred: why are these being connected?

conf/master
```diff
@@ -1267,14 +1282,15 @@
...
-#runner_returns: True
+#runner_returns: False
 
 # Permanently include any available Python 3rd party modules into thin and minimal Salt
 # when they are generated for Salt-SSH or other purposes.
 # The modules should be named by the names they are actually imported inside the Python.
 # The value of the parameters can be either one module or a comma separated list of them.
 #thin_extra_mods: foo,bar
 #min_extra_mods: foo,bar,baz
+#
```
gtmanfred: why are we adding this extra comment?

gtmanfred (dismissed):
This just seems like you are adding arbitrary newlines and comments to connect stuff that doesn't always seem related, but also separating random configuration options.

removing the space after # makes sense, but a lot of this is just superfluous.

conf/master
rjc: As per comment at the top of the `conf/master` file:

    # This configuration file is used to manage the behavior of the Salt Master.
    # Values that are commented out but have an empty line after the comment are
    # defaults that do not need to be set in the config. If there is no blank line
    # after the comment then the value is presented as an example and is not the
    # default.

This is to highlight disambiguate option which **are** the default - otherwise, the file is all over the place and without further reading, user has no way of telling whether a particular option is the default or not. Please see #50025 what prompted both the issue and this PR.

conf/master
rjc: Because `dynamic` is **not** the default, hence shouldn't be followed by a blank line.

conf/master
rjc: `False` **is** the default so *should* be followed by a blank line.

conf/master
rjc: It isn't the default.

conf/master
rjc: Because it isn't the default.

conf/master
rjc: To remove ambiguity - this option is the default, hence a blank line after it.

conf/master
rjc: Not the default, hence add comment/remove blank line.

conf/master
rjc: Again, because according to the comment at the top of the file, options which are the default ones, should be followed by a blank line - the ones which are not, shouldn't be. Otherwise it is very confusing and the comment at the top doesn't make sense/is redundant.

cachedout:
I think the spacing changes are good here. One thing that I have some concern with are the cases where we remove sample values and replace them with single-quotes. While I understand that this is to better reflect the true default option, I also think that it makes it a bit harder for a new user to see the sort of value which could be set.

In these cases, could we move the previous value up into the accompanying documentation? So, for example, if we had a case where previously it was:

```
#foo: bar
```
and it was replaced with:
```
#foo: ''
```
could we then add something like to the description?:
```
example values: 'bar'
```

rjc:
@cachedout Thanks for the feedback - greatly appreciated!

Whilst I see your point, doing so would only *add* more confusion for new users - as it is the case now, at least for me.

There are exactly two occurrences of `#option: ''` in my proposed changes:

    #event_return: ''

and:

    #eauth_acl_module: ''

The former only gives `mysql` and the latter `django` as examples and, as a new user, without any other example values, I read this as "these are endorsed/suggested values". If that *is* indeed the case, then there should be an explanation why it is so. If it isn't, which I assume is true, then *none* should be given as an example, especially if other values aren't mentioned anywhere in the file.

It is surprisingly difficult, *especially* for a new user, to be able to know what *other* possible values can either of these options have. After a bit of digging - BTW, search the **whole** website for `eauth_acl_module` and see for yourself - I finally found what I was looking for: https://docs.saltstack.com/en/latest/ref/auth/all/

It was much easier with returners ;^) https://docs.saltstack.com/en/latest/ref/returners/all/

Personally, I'd like to see the two suggested changes as they are *but* add the links aforementioned links to their respective sections. How about that?

cachedout:
> Personally, I'd like to see the two suggested changes as they are _but_ add the links aforementioned links to their respective sections. How about that?

I think that is a very reasonable approach. Let's do that. Thanks, @rjc!

rjc:
@cachedout Updated with the above URLs. Sorry for the delay.

cachedout:
Resolved merge conflict.

sagetherage:
from the Community Port Jam notes: removing and adding some additional spaces into master was done a while ago, with one of branch mergings, no port to master needed-skipped

PR #50167, opened 2018-10-23
skip win_runas test on not windows

gtmanfred:
### What does this PR do?
allow win_runas test to be loaded by pytest, and not cause everything to fail

### Tests written?

Yes

### Commits signed with GPG?

Yes


gtmanfred:
closing for backports.

Issue #50105, opened 2018-10-18
Reasons for asynchronous Pillar refresh and behavior of modules/saltutil.sync_all


afischer-opentext-com:
Hi, 

I am currently attempting to understand issues I see in regards to not-up-to date pillar in combination with the use of saltenv/pillarenv. Currently I am not in the position to file a valid defect due to lack of understanding of how and why the code behaves (by purpose). 

Our setup is that we have all runners, states, modules, pillar and grain data in a git repository and highly rely on the saltenv/pillarenv feature to segregate different versions of the code base. As we highly rely on the saltenv feature (salt master configuration `pillarenv_from_saltenv: True`) we perform in front of any highstate a call [saltutil.sync_all(refresh=True, saltenv=some_value)](/saltstack/salt/blob/develop/salt/modules/saltutil.py#L988) to ensure the highstate is using the expected custom modules, grains, pillar etc, available in the mentioned saltenv. 

For some reason we observe from time to time that the pillar or grain data is not accurate and as expected. When reviewing the code of the saltutil.sync_all method I got some questions. The method includes this code
```
    if refresh:
        refresh_modules()
        refresh_pillar()
```

This means the `refresh_modules` call happens asynchronous and the method immediately returns before starting the `refresh_pillar`. Is this good? Wouldn't it make sense to be sure the modules and grains are refreshed before refreshing the pillar?
I am not sure what the methods are _exactly_ doing but I can imagine a situation in which Jinja code is rendering pillar data using custom module code or grains. If this module code is not confirmed to be current, pillar data might be not accurate. 

My other question is: Why is there no way to refresh pillar in a synchronous way? Currently we workaround this by adding a time out of some seconds but consider this only a hack which - even worse - still does not reliably work for some reasons we do not yet understand. This might be also caused by Jinja magic performing delaying calls but we are not yet sure about it. 

Any input which may increase the understanding is highly appreciated. 

The currently used Salt version is 2017.7.5

Ch3LL:
> I am not sure what the methods are exactly doing but I can imagine a situation in which Jinja code is rendering pillar data using custom module code or grains. If this module code is not confirmed to be current, pillar data might be not accurate.

`refresh_modules` is calling `self._load_modules(force_refresh, notify=notify)` which just returns the modules and returners which can be loaded. It is not actually rendering jinja at this step.

> Is the refresh_modules() call really needed here? Isn't refresh_modules a part of refresh_pillar already?

[...]

PR #50168, opened 2018-10-23
Allow to refresh pillar synchronously

ghost:
### What does this PR do?

This PR adds the optional ability to run `saltutil.refresh_pillar` Allow to refresh pillar synchronously, saying `saltutil.refresh_pillar async=False`. I tried to follow the explanatory comment from @gtmanfred and I believe that this works, still I am not sure about the mentioned use of the  `_gather_pillar()` method. I am happy to rework the PR if something is missing. 

This PR relates to issue #50105

### Previous Behavior
It was not possible to know when pillar got really refreshed and the minion is updated properly. 

### New Behavior
The user can ensure that pillar information is in a refreshed state.

### Tests written?

Yes

### Commits signed with GPG?

Yes


gtmanfred (approved):
This looks great, thanks!

salt/minion.py
```diff
@@ -2196,11 +2196,14 @@ def pillar_refresh(self, force_refresh=False):
                     self.opts['saltenv'],
                     pillarenv=self.opts.get('pillarenv'),
                 ).compile_pillar()
+                if notify:
+                    evt = salt.utils.event.get_event('minion', opts=self.opts, listen=False)
+                    evt.fire_event({'complete': True}, tag='/salt/minion/minion_pillar_complete')
```
isbm: I would move that thing into a constant. Since there is currently no nice home for this, I would keep it in this file, something like:

```python
PILLAR_COMPLETE = '/salt/minion/minion_pillar_complete'
```

And then I would [reuse it here](https://github.com/saltstack/salt/pull/50168/files#diff-a72c3e04b3df0479b574ff06020607efR1055).

isbm (changes requested):

salt/minion.py
afischer-opentext-com: I changed the code accordingly.

afischer-opentext-com: Okay, the last change went not well

```
  File "./tests/runtests.py", line 61, in <module>
    from tests.integration import TestDaemon  # pylint: disable=W0403
  File "/tmp/kitchen/testing/salt/tests/integration/__init__.py", line 36, in <module>
    from tests.support.case import ShellTestCase
  File "/tmp/kitchen/testing/salt/tests/support/case.py", line 36, in <module>
    from tests.support.mixins import AdaptedConfigurationTestCaseMixin, SaltClientTestCaseMixin
  File "/tmp/kitchen/testing/salt/tests/support/mixins.py", line 34, in <module>
    import salt.utils.event
  File "/tmp/kitchen/testing/salt/utils/event.py", line 79, in <module>
    import salt.payload
  File "/tmp/kitchen/testing/salt/payload.py", line 17, in <module>
    import salt.crypt
  File "/tmp/kitchen/testing/salt/crypt.py", line 64, in <module>
    import salt.utils.crypt
  File "/tmp/kitchen/testing/salt/utils/crypt.py", line 16, in <module>
    import salt.loader
  File "/tmp/kitchen/testing/salt/loader.py", line 39, in <module>
    from salt.minion import MinionEvents
  File "/tmp/kitchen/testing/salt/minion.py", line 66, in <module>
    import salt.client
  File "/tmp/kitchen/testing/salt/client/__init__.py", line 31, in <module>
    import salt.cache
  File "/tmp/kitchen/testing/salt/cache/__init__.py", line 16, in <module>
    from salt.payload import Serial
ImportError: cannot import name Serial
```
Is this due to a circular dependency?

cachedout:
I get where you're going with this but I can't help but wonder if we should just generalize the pattern of "blocking" on a return event somewhere that we can re-use more easily. It seems like this is something that could be used in other places.

Thoughts?

salt/minion.py
afischer-opentext-com: Moved the event constants into an own module `salt.events`, similar to  `salt.defaults`.

afischer-opentext-com:
Yeah, what we basically want is to be sure that minions work with the pillar, grain, state etc. information we expect them to work this, while dealing with an higher amount of saltenvs (using gitfs) which may change quickly. Still, this should be not part of this PR, as it sounds like a lot of planning and requires some deep knowledge of the code base, which I don't have :)

salt/minion.py
afischer-opentext-com: @isbm, how to we proceed? After moving the constants into a dedicated module it works again.

isbm: Not sure it should be a separate package though. If that so, I would rather put that thing to `salt/defaults/events.py` then. This would be a good home for this sort of things. Then we could gather those hard-coded tags all over the place and refer them there too.

@terminalmage any suggestions/objections?

afischer-opentext-com: @isbm, I changed the PR accordingly.

afischer-opentext-com:
To my understanding the `saltutil.refresh_pillar` method only guarantees to return False in case of an issue. The return value in case of no issue is not defined. (This does not change with the PR).

salt/minion.py
afischer-opentext-com: @isbm, how can we move this forward?

salt/defaults/events.py
```diff
@@ -0,0 +1,13 @@
...
+
+Do NOT, import any salt modules (salt.utils, salt.config, etc.) into this file,
+as this may result in circular imports.
+
+.. versionadded:: Neon
+'''
+
+# Constants for events on the minion bus
+EV_PILLAR_COMPLETE = '/salt/minion/minion_pillar_complete'
+EV_MOD_COMPLETE = '/salt/minion/minion_mod_complete'
```
isbm: Sorry being so much annoying here, but this is already in `events`, so no need to repeat yourself with `EV_` prefix. However, `MINION_` would help here better. :wink: You will be using it like:

```python
if salt.defaults.events.MINION_PILLAR_COMPLETE .....
```

This is perfectly readable & comprehensible.

isbm (changes requested):

salt/defaults/events.py
afischer-opentext-com: @isbm, I adapted the naming as requested.

isbm (approved):
@afischer-opentext-com OK, done. :+1:

afischer-opentext-com:
This may be not the smartest question, but what is the next step here, is there something I need to do?

salt/minion.py
```diff
@@ -2185,12 +2186,12 @@ def matchers_refresh(self):
...
     # TODO: only allow one future in flight at a time?
     @tornado.gen.coroutine
-    def pillar_refresh(self, force_refresh=False):
+    def pillar_refresh(self, force_refresh=False, notify=False):
         '''
         Refresh the pillar
         '''
         if self.connected:
-            log.debug('Refreshing pillar')
+            log.debug('Refreshing pillar. Notify: {}'.format(notify))
```
terminalmage: Please use `%s` replacement instead of `.format()` in log messages.

```python
log.debug('Refreshing pillar. Notify: %s', notify)
```

terminalmage (changes requested):

salt/minion.py
afischer-opentext-com: @terminalmage , I fixed it.

terminalmage (approved):

PR #50175, opened 2018-10-23
Add root parameter to useradd, shadow and groupadd

aplanas:
### What does this PR do?
Add root parameter to useradd and shadow execution modules

### DIFF
--- a/salt/modules/groupadd.py
+++ b/salt/modules/groupadd.py
@@ -12,6 +12,7 @@ Manage groups on Linux, OpenBSD and NetBSD
 # Import python libs
 from __future__ import absolute_import, print_function, unicode_literals
 import logging
+import functools
 import os
 
 from salt.ext import six
@@ -43,6 +44,18 @@ def add(name, gid=None, system=False, root=None):
     '''
     Add the specified group
 
+    name
+        Name of the new group
+
+    gid
+        Use GID for the new group
+
+    system
+        Create a system account
+
+    root
+        Directory to chroot into
+
     CLI Example:
 
     .. code-block:: bash
@@ -54,10 +67,11 @@ def add(name, gid=None, system=False, root=None):
         cmd.append('-g {0}'.format(gid))
     if system and __grains__['kernel'] != 'OpenBSD':
         cmd.append('-r')
-    cmd.append(name)
 
     if root is not None:
         cmd.extend(('-R', root))
+
+    cmd.append(name)
 
     ret = __salt__['cmd.run_all'](cmd, python_shell=False)
 
@@ -68,25 +82,39 @@ def delete(name, root=None):
     '''
     Remove the named group
 
+    name
+        Name group to delete
+
+    root
+        Directory to chroot into
+
     CLI Example:
 
     .. code-block:: bash
 
         salt '*' group.delete foo
     '''
-    cmd = ['groupdel', name]
+    cmd = ['groupdel']
 
     if root is not None:
         cmd.extend(('-R', root))
+
+    cmd.append(name)
 
     ret = __salt__['cmd.run_all'](cmd, python_shell=False)
 
     return not ret['retcode']
 
 
-def info(name):
+def info(name, root=None):
     '''
     Return information about a group
+
+    name
+        Name of the group
+
+    root
+        Directory to chroot into
 
     CLI Example:
 
@@ -94,8 +122,13 @@ def info(name):
 
         salt '*' group.info foo
     '''
+    if root is not None:
+        getgrnam = functools.partial(_getgrnam, root=root)
+    else:
+        getgrnam = functools.partial(grp.getgrnam)
+
     try:
-        grinfo = grp.getgrnam(name)
+        grinfo = getgrnam(name)
     except KeyError:
         return {}
     else:
@@ -112,9 +145,15 @@ def _format_info(data):
             'members': data.gr_mem}
 
 
-def getent(refresh=False):
+def getent(refresh=False, root=None):
     '''
     Return info on all groups
+
+    refresh
+        Force a refresh of group information
+
+    root
+        Directory to chroot into
 
     CLI Example:
 
@@ -126,15 +165,51 @@ def getent(refresh=False):
         return __context__['group.getent']
 
     ret = []
-    for grinfo in grp.getgrall():
+    if root is not None:
+        getgrall = functools.partial(_getgrall, root=root)
+    else:
+        getgrall = functools.partial(grp.getgrall)
+
+    for grinfo in getgrall():
         ret.append(_format_info(grinfo))
     __context__['group.getent'] = ret
     return ret
 
 
+def _chattrib(name, key, value, param, root=None):
+    '''
+    Change an attribute for a named user
+    '''
+    pre_info = info(name, root=root)
+    if not pre_info:
+        return False
+
+    if value == pre_info[key]:
+        return True
+
+    cmd = ['groupmod']
+
+    if root is not None:
+        cmd.extend(('-R', root))
+
+    cmd.extend((param, value, name))
+
+    __salt__['cmd.run'](cmd, python_shell=False)
+    return info(name, root=root).get(key) == value
+
+
 def chgid(name, gid, root=None):
     '''
     Change the gid for a named group
+
+    name
+        Name of the group to modify
+
+    gid
+        Change the group ID to GID
+
+    root
+        Directory to chroot into
 
     CLI Example:
 
@@ -142,24 +217,21 @@ def chgid(name, gid, root=None):
 
         salt '*' group.chgid foo 4376
     '''
-    pre_gid = __salt__['file.group_to_gid'](name)
-    if gid == pre_gid:
-        return True
-    cmd = ['groupmod', '-g', gid, name]
-
-    if root is not None:
-        cmd.extend(('-R', root))
-
-    __salt__['cmd.run'](cmd, python_shell=False)
-    post_gid = __salt__['file.group_to_gid'](name)
-    if post_gid != pre_gid:
-        return post_gid == gid
-    return False
+    return _chattrib(name, 'gid', gid, '-g', root=root)
 
 
 def adduser(name, username, root=None):
     '''
     Add a user in the group.
+
+    name
+        Name of the group to modify
+
+    username
+        Username to add to the group
+
+    root
+        Directory to chroot into
 
     CLI Example:
 
@@ -195,6 +267,15 @@ def adduser(name, username, root=None):
 def deluser(name, username, root=None):
     '''
     Remove a user from the group.
+
+    name
+        Name of the group to modify
+
+    username
+        Username to delete from the group
+
+    root
+        Directory to chroot into
 
     CLI Example:
 
@@ -241,6 +322,15 @@ def deluser(name, username, root=None):
 def members(name, members_list, root=None):
     '''
     Replaces members of the group with a provided list.
+
+    name
+        Name of the group to modify
+
+    members_list
+        Username list to set into the group
+
+    root
+        Directory to chroot into
 
     CLI Example:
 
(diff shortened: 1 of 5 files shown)


salt/modules/shadow.py
```diff
@@ -96,15 +96,19 @@ def set_inactdays(name, inactdays):
     pre_info = info(name)
     if inactdays == pre_info['inact']:
         return True
-    cmd = 'chage -I {0} {1}'.format(inactdays, name)
+    cmd = ['chage']
+    if root is not None and __grains__['kernel'] != 'AIX':
+        cmd.extend(('-R', root))
```
cachedout: I cannot find documentation for `-R` when used with `chage`. Could you please point me to some?

aplanas: Sure!

I can see it in my system (openSUSE) and in the source code:

https://github.com/shadow-maint/shadow/blob/master/src/chage.c#L159

According to blame was added 7 years ago:

https://github.com/shadow-maint/shadow/commit/0857837e64a5a45ef686806aa3f3694ff733d5b2

cachedout: Hrm. My worry here is that I'm seeing this in certain docs:

https://linux.die.net/man/1/chage

That said, perhaps we could just improve error checking down in L104 and we should be fine. Thoughts?

aplanas: I do not know what to say, looks like that was documented 7 years ago too, according to the previous commit.

I can see the same option in CentOS 7, that provides shadow-utils-4.1.5.1-24.

I do not know from where https://linux.die.net gets the information, or when was the last update, but I cannot find the same option in useradd, another tool form shadow-utils:

https://linux.die.net/man/8/useradd

Also note that there is a missing CONFIGURATION section at the end of the man page, that was added 9 years ago:

https://github.com/shadow-maint/shadow/commit/5d6c3143047c30c2c402405a98806bdd664b6bee

But more important, the same -R parameter is already in the salt code for useradd (that is also part of the shadow-utils package), added three years ago:

https://github.com/saltstack/salt/commit/1eb3b9c5e4ae3952c8198692e994997207733406

IMHO linux.dice.net is not a reliable source of information.

P.S according to https://www.die.net/ the man pages are from 2002

aplanas: > That said, perhaps we could just improve error checking down in L104 and we should be fine. Thoughts?

This makes me wonder what to do with the already -R in useradd, that was added the same date in a different commit:

https://github.com/shadow-maint/shadow/commit/cecae46ccfb85c2c4bf6e0fe0419d740ba1f2ecc

salt/modules/useradd.py
```diff
@@ -470,10 +473,10 @@ def chfullname(name, fullname):
 
         salt '*' user.chfullname foo "Foo Bar"
     '''
-    return _update_gecos(name, 'fullname', fullname)
+    return _update_gecos(name, 'fullname', fullname, root)
```
aplanas: Uhmm this is not correct. I realized that _update_gecos accept the root parameter, used to usermod, but this is deceptive as the return value use _get_gecos, that use pwd.getpwnam, that do not have a root parameter.

Maybe I can implement an alternative to pwd.* in case that root parameter is provided, and if not falback to pwd

aplanas:
Ready for review.

cachedout:
@aplanas Could you add docstrings for these new parameters, please?

aplanas:
@cachedout I added the documentation strings for root and for the missing parameters. Done.

aplanas:
Checking the logs looks like that user.getent fails in windows, but this PR do not touch this code (is in a different module)

cachedout:
@aplanas I've asked @twangboy from @saltstack/team-windows to swing by here and see if he can sort out why that test is failing before we get this merged.

aplanas:
any more reviews? Any thing that I can do?

twangboy:
I don't think this will cause any problems in Windows

salt/modules/groupadd.py
```diff
@@ -123,41 +165,74 @@ def getent(refresh=False):
         return __context__['group.getent']
 
     ret = []
-    for grinfo in grp.getgrall():
+    if root is not None:
+        _fun = functools.partial(_getgrall, root=root)
```
isbm: Even GvR himself [doesn't agree](https://twitter.com/gvanrossum/status/391769557758521345?lang=en) with you here! :laughing: Me too, BTW.

salt/modules/groupadd.py
```diff
@@ -123,41 +165,74 @@ def getent(refresh=False):
...
+        return True
+
+    cmd = ['groupmod']
+
+    if root is not None:
+        cmd.extend(('-R', root))
+
+    cmd.extend((param, value, name))
+
+    __salt__['cmd.run'](cmd, python_shell=False)
```
isbm: Some checks/logging would help here for debugging.

salt/modules/groupadd.py
```diff
@@ -284,3 +377,37 @@ def members(name, members_list, root=None):
         return False
 
     return not retcode
+
+
+def _getgrnam(name, root=None):
+    '''
+    Alternative implementation for getgrnam, that use only /etc/group
+    '''
+    root = '/' if not root else root
```
isbm: Safe to just say:

```python
root = root or '/'
```

salt/modules/groupadd.py
```diff
@@ -284,3 +377,37 @@ def members(name, members_list, root=None):
...
+    with salt.utils.files.fopen(passwd) as fp_:
+        for line in fp_:
+            line = salt.utils.stringutils.to_unicode(line)
+            comps = line.strip().split(':')
+            if comps[0] == name:
+                # Generate a getpwnam compatible output
+                comps[2] = int(comps[2])
+                comps[3] = comps[3].split(',') if comps[3] else []
+                return grp.struct_group(comps)
+    raise KeyError
```
isbm: Please add message here, so it gets logged later.

salt/modules/groupadd.py
```diff
@@ -284,3 +377,37 @@ def members(name, members_list, root=None):
...
+
+
+def _getgrall(root=None):
+    '''
+    Alternative implemetantion for getgrall, that use only /etc/group
+    '''
+    root = '/' if not root else root
```
isbm: Also can be just `or`.

salt/modules/groupadd.py
```diff
@@ -284,3 +377,37 @@ def members(name, members_list, root=None):
...
+    passwd = os.path.join(root, 'etc/group')
+    with salt.utils.files.fopen(passwd) as fp_:
+        for line in fp_:
+            line = salt.utils.stringutils.to_unicode(line)
+            comps = line.strip().split(':')
+            # Generate a getgrall compatible output
+            comps[2] = int(comps[2])
+            comps[3] = comps[3].split(',') if comps[3] else []
```
isbm: Scary. Here I would first check if `comps` is at least length of 4. Log at least debug if not, and include what was the value.

salt/modules/shadow.py
```diff
@@ -306,81 +392,145 @@ def set_password(name, password, use_usermod=False):
...
+
     CLI Example:
 
     .. code-block:: bash
 
         salt '*' shadow.list_users
     '''
-    return sorted([user.sp_nam for user in spwd.getspall()])
+    if root is not None:
+        _fun = functools.partial(_getspall, root=root)
```
isbm: I think, the same GvR's tweet still applies here. :wink:

salt/modules/shadow.py
```diff
@@ -306,81 +392,145 @@ def set_password(name, password, use_usermod=False):
...
+    else:
+        _fun = functools.partial(spwd.getspall)
+
+    return sorted([user.sp_namp if hasattr(user, 'sp_namp') else user.sp_nam
+                   for user in _fun()])
+
+
+def _getspnam(name, root=None):
```
isbm: This looks somewhat very similar to what is used in other module. You could join all that and make utility function instead, removing lots of code duplication from both places.

salt/modules/useradd.py
```diff
@@ -306,98 +381,132 @@ def getent(refresh=False):
         return __context__['user.getent']
 
     ret = []
-    for data in pwd.getpwall():
+    if root is not None and __grains__['kernel'] != 'AIX':
+        _fun = functools.partial(_getpwall, root=root)
```
isbm: :-)

salt/modules/useradd.py
```diff
@@ -306,98 +381,132 @@ def getent(refresh=False):
...
+    pre_info = info(name, root=root)
+    if not pre_info:
+        raise CommandExecutionError('User \'{0}\' does not exist'.format(name))
+
+    if value == pre_info[key]:
+        return True
+
+    cmd = ['usermod']
+
+    if root is not None and __grains__['kernel'] != 'AIX':
```
isbm: You mean, this is fine on Windows, OSX and Solaris?

salt/modules/useradd.py
```diff
@@ -546,25 +703,43 @@ def chloginclass(name, loginclass, root=None):
...
     '''
+    # If root is provided, we use a less portable solution that
+    # depends on analyzing /etc/passwd manually. Of course we cannot
+    # find users from NIS nor LDAP, but in those cases do not makes
+    # sense to provide a root parameter.
+    #
+    # Please, note that if the non-root /etc/passwd file is long the
+    # iteration can be slow.
+    if root is not None and __grains__['kernel'] != 'AIX':
+        _fun = functools.partial(_getpwnam, root=root)
```
isbm: And again...

salt/modules/useradd.py
```diff
@@ -654,43 +838,79 @@ def list_groups(name):
...
+
     CLI Example:
 
     .. code-block:: bash
 
         salt '*' user.list_users
     '''
-    return sorted([user.pw_name for user in pwd.getpwall()])
+    if root is not None and __grains__['kernel'] != 'AIX':
+        _fun = functools.partial(_getpwall, root=root)
```
isbm: Bump.

isbm (changes requested):
@aplanas I see also lots of code duplication and repeating yourself. Could you please make sure code is not duplicated?

salt/modules/groupadd.py
aplanas: Can I disagree here? The example is for partials that do not expect arguments (is done explicit two or three comments later), and this is not the case. I find this more explicit than this:

```python
    if root is not None:
        _fun = lambda name: _getgrnam(name, root=root)
    else:
        _fun = lambda name: grp.getgrnam(name)
```

In the lambda form is not explicit that the call is not evaluated (lazy evaluation), and that the free parameter is indeed name.

If you have an strong objection I can change it.

```

File: salt/modules/groupadd.py (hunk 5 above)
Comment (aplanas) [reply salt/modules/groupadd.py]: sure

File: salt/modules/groupadd.py (hunk 6 above)
Comment (aplanas) [reply salt/modules/groupadd.py]: yes

File: salt/modules/shadow.py (hunk 10 above)
Comment (aplanas) [reply salt/modules/shadow.py]: Ah indeed I was tempted to do this. Actually this patch removes a lot of old duplicated code. But in this case the interpretation of the columns in different for user, groups and shadow. The only generic part is open the file, read each line, and split the line with ':'

I will think what I can factor out.

File: salt/modules/useradd.py (hunk 12 above)
Comment (aplanas) [reply salt/modules/useradd.py]: nop, see `__virtual__`

File: salt/modules/groupadd.py (hunk 4 above)
Comment (aplanas) [reply salt/modules/groupadd.py]: Indeed, can we do that in a different change? The original code do not provide any check or log here for useradd, groupadd and shadow. I this this change for avoid code duplication, but I take care or not changing the semantic.

Adding some checks here can be normalized in a different commit, and this will require adjusting the tests.

**Comment by aplanas** [comment]
Any strong opinion to not merge this? There is one complain that can be fair: the new grp, pwd and spwd alternatives can be sharing some code (share the same structure of opening a file and reading the content line by line).

I do not want to merge them in the same code under the module utils, as they are very local. But I can revisit this opinion later if I see that this can be used outside their domain.

Also I removed a lot of [duplicated code](https://github.com/saltstack/salt/pull/50175/commits/1672f06d7e584c90ebed6bb24a9011f2a1a94470#diff-6ff7900860115f0b9224a20aca971b98L344) unrelated with my goal of introducing a new root parameter.

I am also not changing the semantic of the operations, validated with the old tests. This is something that we can revisit later, as makes sense to unify some fails (that sometimes are exceptions and other times return False)

File: salt/modules/groupadd.py (hunk 3 above)
Comment (aplanas) [reply salt/modules/groupadd.py]: Umh. E731 (do not assign a lambda expression, use a def) happens in this case.

I like `lambda`s as anyone, but in this case if some new user see the code, she/he will find documentation about [functools.partial](https://docs.python.org/3.7/library/functools.html#functools.partial) easily.

Tangentially about this topic, is there any set of WXXX or EXXX that the team agrees to ignore?

Comment (isbm) [reply salt/modules/groupadd.py]: Ahh, you're [running into this](https://www.python.org/dev/peps/pep-0008/#programming-recommendations).

```python
if root:
    def _fun(name): return _getgrnam(name, root=root)
else:
    def _fun(name): return grp.getgrnam(name)
```

aplanas: :confused:
What how is defining a local inner function be better than a standard library that is documented? Is there any plan to drop `functools.partial` that I am not aware of?

Also is used along the codebase of salt too:

* https://github.com/saltstack/salt/blob/develop/salt/utils/parsers.py#L24
* https://github.com/saltstack/salt/blob/develop/salt/utils/boto3mod.py#L40
* https://github.com/saltstack/salt/blob/develop/salt/utils/botomod.py#L40
* https://github.com/saltstack/salt/blob/develop/salt/utils/roster_matcher.py#L81
* https://github.com/saltstack/salt/blob/develop/salt/modules/rpmbuild.py#L557
...

isbm: @aplanas I am just fixing you a lint error according to as [PEP-8 says](https://www.python.org/dev/peps/pep-0008/#programming-recommendations). I still find this:

```python
def foo(x): f(x)
```

Much better than this:

```python
foo = functools.partial(f, x)
```

aplanas: For future reference. I found the pep about currying:

https://www.python.org/dev/peps/pep-0309/

aplanas:
I am still interested in this PR. Is there anything more that I can do?

aplanas:
* Rebased the code
* In the partial call I remove the ugly _fun name and make explicit the call to the correct function name in pwd, spwd and grp. I hope that this clarify why I currying over the real or the fake function (so I do not break the std interface)
* Refactor a bit _set_attrib to remove some extra returns. Still return early, as I am not able to find a way to avoid two nested indentations. But clarify the situation with a comment.

isbm (approved):
@cachedout I have no more issues with this PR. Unless someone else does, we're fine here.

aplanas:
Thanks for the reviews and the hard work here!

PR #50179, opened 2018-10-23
fix pillar_cache support for cacheing multi pillarenv

rongzeng54:
### What does this PR do?
fix pillar_cache support for cacheing multi pillarenv

No

### Previous Behavior
pillar_cache cannot support for cacheing multi pillarenv

### New Behavior
pillar_cache can support for cacheing multi pillarenv

### Tests written?
No

### Commits signed with GPG?
No

### DIFF
--- a/salt/pillar/__init__.py
+++ b/salt/pillar/__init__.py
@@ -322,7 +322,9 @@ class PillarCache(object):
             else:
                 # We found the minion but not the env. Store it.
                 fresh_pillar = self.fetch_pillar()
-                self.cache[self.minion_id][self.pillarenv] = fresh_pillar
+                self.cache[self.minion_id] = dict(self.cache[self.minion_id], **{
+                    self.pillarenv: fresh_pillar
+                })
                 log.debug('Pillar cache miss for pillarenv %s for minion %s', self.pillarenv, self.minion_id)
                 return fresh_pillar
         else:
--- a/tests/unit/test_pillar.py
+++ b/tests/unit/test_pillar.py
@@ -9,6 +9,7 @@
 
 # Import python libs
 from __future__ import absolute_import
+import os
 import shutil
 import tempfile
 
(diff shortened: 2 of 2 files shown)


gtmanfred (changes requested):
can you add a test for this?

salt/pillar/__init__.py
```diff
@@ -322,7 +322,9 @@ def compile_pillar(self, *args, **kwargs):  # Will likely just be pillar_dirs
             else:
                 # We found the minion but not the env. Store it.
                 fresh_pillar = self.fetch_pillar()
-                self.cache[self.minion_id][self.pillarenv] = fresh_pillar
+                self.cache[self.minion_id] = dict(self.cache[self.minion_id], **{
+                    self.pillarenv: fresh_pillar
+                })
```
isbm: I am not sure what I am looking at. The `self.cache[self.minion_id]` contains a dict, to which `fresh_pillar` variable is assigned. Now, all you do here is just create that dict over, with the same values, ending up with the same result.

What exactly this is helping?

isbm (changes requested):

salt/pillar/__init__.py
rongzeng54: @isbm 
Please check https://github.com/saltstack/salt/blob/develop/salt/utils/cache.py#L84

terminalmage: That is just a link to the CacheDisk class. Can you please explain better?

terminalmage (commented):
See my comment [here](https://github.com/saltstack/salt/pull/50179/files#r231671803)

salt/pillar/__init__.py
isbm: @rongzeng54 Thanks for giving me a link to a class. Now please explain what you are doing and how this helps.

rongzeng54: > @rongzeng54 Thanks for giving me a link to a class. Now please explain what you are doing and how this helps.

> self.cache[self.minion_id][self.pillarenv] = x
Will only call the `__getitem__` of CacheDisk class, and will not call the `__setitem__`

> self.cache[self.minion_id] = x
Will only call the `__setitem__` of CacheDisk class

In here, the second is correct

rongzeng54:
@terminalmage 
Please check my comment [here](https://github.com/saltstack/salt/pull/50179#discussion_r232007986)

terminalmage:
You haven't explained why this helps, you haven't explained well enough what problem this fixes, and you haven't written any tests for this.

rongzeng54:
@terminalmage @rallytime @gtmanfred 
I have tried to add a unit test for this fix. Please check.

tests/unit/test_pillar.py
```diff
@@ -895,3 +896,42 @@ def test_pillar_send_extra_minion_data_from_config(self):
...
+            with patch('salt.pillar.PillarCache.fetch_pillar', MagicMock(return_value={'env': 'test'})):
+                pillar.compile_pillar()
+                self.assertEqual(pillar.cache['mocked-minion'], {'dev': {'env': 'dev'},
+                                                                 'test': {'env': 'test'}})
+            # We have a cache hit! Send it back.
+            pillar = salt.pillar.PillarCache(opts, {}, 'mocked-minion', 'test', pillarenv='test')
+            with patch('salt.pillar.PillarCache.fetch_pillar', MagicMock(return_value={'k': 'v'})):
+                pillar.compile_pillar()
+                self.assertEqual(pillar.cache['mocked-minion'], {'dev': {'env': 'dev'},
+                                                                 'test': {'env': 'test'}})
```
isbm: The last test doesn't make sense. You should actually explicitly modify time so TTL changes and then you get back `k/v` instead if `env/test`. And you should explicitly hold time so the cache works.

But you continue developing this, **directly ignoring** requests to explain why this is even needed and what you're trying to solve here.

isbm (changes requested):

tests/unit/test_pillar.py
rongzeng54: Hi @isbm 
The last test is for testing whether the cache was actually hit but not for testing TTL, because the TTL has been tested in file https://github.com/saltstack/salt/blob/develop/tests/unit/utils/test_cache.py

And I think i have tried to explained why this helps, and what problem this fixes. 
```
self.cache[self.minion_id][self.pillarenv] = x
Will only call the __getitem__ of CacheDisk class, and will not call the __setitem__

self.cache[self.minion_id] = x
Will only call the __setitem__ of CacheDisk class
```

In here, the second is correct.

Maybe you should run the test first.

rongzeng54:
> @rongzeng54 You still haven't explained why this helps, you still haven't explained well enough what problem this fixes.

comment [here](https://github.com/saltstack/salt/pull/50179#discussion_r234644913)

tests/unit/test_pillar.py
isbm: @rongzeng54 but it should not. I am not sure you understand how this supposed to work. Here is an example:

```python
from salt.utils.cache import CacheDisk
import time
import os

cache_file = '/tmp/spam_eggs'
cache = CacheDisk(0, cache_file)

if not os.path.exists(cache_file):
    cache['foo'] = {}
    cache['foo']['bar'] = {}
    cache['foo']['bar']['spam'] = 'Fred'
    cache._write()
else:
    cache._read()
    print(cache['foo'])
    cache['foo'][str(time.time())] = 'one more time'
    cache._write()
```

So first time when you call this, it will  do nothing:
```
$ python cd.py
$
```

But then each time you call it, it will add more data:
```
$ python cd.py 
{'bar': {'spam': 'Fred'}}
$ python cd.py 
{'bar': {'spam': 'Fred'}, '1542639980.9734464': 'one more time'}
$ python cd.py 
{'bar': {'spam': 'Fred'}, '1542639980.9734464': 'one more time', '1542639982.2767625': 'one more time'}
```

What part is still missing you think?

rongzeng54: @isbm it should call __setitem__. Because the function _write used for Write out the pillar data to disk exists in __setitem__ but not __getitem__.

PR #47860, opened 2018-05-29
Clean up doc build warnings, add best practices for writing states

terminalmage:
This reduces the number of build warnings from 2986 to 1. We can't get rid of that last one because it is triggered due to our custom autofunction plugin (which overrides the default autofunction from Sphinx).

Additionally, this PR adds some best practices documentation for writing states.

PR #50180, opened 2018-10-23
Revert curly braces example for nested dictionaries

max-arnold:
### What does this PR do?

The snippet in the docs with curly braces for nested dictionaries is incorrect (almost identical to the previous one): https://docs.saltstack.com/en/latest/topics/troubleshooting/yaml_idiosyncrasies.html#nested-dictionaries. The change was introduced in #47860.

This PR reverts the snippet to previous state. I did a test build with `make html` and `sphinx==1.7.9` and found no new warnings.

### Commits signed with GPG?

No


cachedout (approved):
Good catch. Thanks, @max-arnold

PR #50184, opened 2018-10-23
Added handling for minion return exceptions containing the word "return"

ethanculler:
### What does this PR do?
This PR adds a check for the object type of the minion return to ensure we are dealing with a dict rather than a string containing 'return' (the key that the dict is looking for). This prevents a bad minion return from negatively impacting the master.

This bug was seen as far back as 2016.3 release

I didn't find any existing issues documenting this case, but there is certainly the chance I missed one.

### Tests written?

No

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/client/__init__.py
+++ b/salt/client/__init__.py
@@ -1283,6 +1283,13 @@ class LocalClient(object):
                 if raw['data']['return'] == {}:
                     continue
 
+                # if the minion throws an exception containing the word "return"
+                # the master will try to handle the string as a dict in the next
+                # step. Check if we have a string, log the issue and continue.
+                if isinstance(raw['data']['return'], six.string_types):
+                    log.error("unexpected return from minion: %s", raw)
+                    continue
+
                 if 'return' in raw['data']['return'] and \
                     raw['data']['return']['return'] == {}:
                     continue


salt/client/__init__.py
```diff
@@ -1283,6 +1283,14 @@ def get_iter_returns(
                 if raw['data']['return'] == {}:
                     continue
 
+                # if the minion throws an exception containing the word "return"
+                # the master will try to handle the string as a dict in the next
+                # step. Check if we have a string, log the issue and continue.
+                if isinstance(raw['data']['return'], six.string_types):
+                    log.error("unexpected return from minion: %s", raw)
+                    raw['data']['return'] = {}
```
gtmanfred: i don't think we want to overwrite this, because we still want to print out the exceptions from the minion return so people can troubleshoot issues.

gtmanfred (changes requested):

salt/client/__init__.py
```diff
@@ -1283,6 +1283,13 @@ def get_iter_returns(
                 if raw['data']['return'] == {}:
                     continue
 
+                # if the minion throws an exception containing the word "return"
+                # the master will try to handle the string as a dict in the next
+                # step. Check if we have a string, log the issue and continue.
+                if isinstance(raw['data']['return'], six.string_types):
+                    log.error("unexpected return from minion: %s", raw)
```
gtmanfred: I am not sure this should be a log.error...  maybe a log.info instead?  @rallytime do you have an opinion?

brejoc:
@ethanculler  Any chance, you could provide a test for this? Otherwise those little details might break again in the future with code changes.

ethanculler:
@brejoc no problem.

salt/client/__init__.py
cachedout: I believe this is the correct log level.

gtmanfred (approved):

dwoz (approved):

gtmanfred:
Thanks @ethanculler and welcome to the saltstack contributors club 🍾

gtmanfred:
bah, we probably do still need a test for this. sorry, that is on me.

ethanculler:
@gtmanfred no worries. This week has been busy, but i'll get a test in for it in another PR (will link to this one) hopefully this weekend. Cheers!

Issue #47771, opened 2018-05-22
salt-cloud always return error with Proxmox provider even when VM successfully created


hatifnatt:
### Description of Issue/Question
After creation of VM salt-cloud always returns an error.
I will check is this issue still preset in Salt 2018+ as soon as I will update to it.

### Setup
Proxmox provider config
```yaml
pve:
  minion:
    master: salt
  driver: proxmox
  url: pve.host.tld
  user: saltcloud@pve
  password: somepassword
```
Profile
```yaml
pve-kvm-debian9:
  provider: pve
  technology: qemu
  memory: 512
  host: pvenode
  size: ''
  ostype: l26
  image: image.iso
  scsi0: local:4,format=qcow2,size=4G
  cdrom: iso-storage:iso/debian-9.0.0-amd64-netinst.iso
  net0: virtio,bridge=vmbr1
  name: salt-provisioned-vm
```
Mapfile
```yaml
pve-kvm-debian9:
  - saltvm
```
### Steps to Reproduce Issue
Create VM from commandline without map file:
```bash
salt-cloud -p pve-kvm-debian9 saltvm
```
VM successfully created, but still useless error is returned:
```
Usage: salt-cloud [options] <-m MAP | -p PROFILE> <NAME> [NAME2 ...]
salt-cloud: error: There was a profile error:
```
Create VM from mapfile:
```bash
salt-cloud -m salvm.map
```
VM successfully created, but although error is returned:
```
The following virtual machines are set to be created:
  saltvm

Proceed? [N/y] y
... proceeding
[ERROR   ] Failed to deploy 'saltvm'. Error:
saltvm:
    ----------
    Error:
```
### Versions Report
```
Salt Version:
           Salt: 2017.7.5

Dependency Versions:
           cffi: Not Installed
       cherrypy: Not Installed
       dateutil: 2.5.3
      docker-py: Not Installed
          gitdb: 2.0.0
      gitpython: 2.1.1
          ioflo: Not Installed
         Jinja2: 2.9.4
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.4.8
   mysql-python: 1.3.7
      pycparser: Not Installed
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: Not Installed
         Python: 2.7.13 (default, Nov 24 2017, 17:33:09)
[...]

PR #50189, opened 2018-10-24
Allow IP detection via qemu-agent (proxmox salt-cloud)

MasonEdgel:
### What does this PR do?
Allows salt-cloud to create QEMU VMs and get their IP via the agent when using a Proxmox cloud.

https://github.com/saltstack/salt/issues/47771

### Previous Behavior
Would silently error out during a clone, as no IP address was determined for the new VM

### New Behavior
Checks if new VM is of type `qemu`, and uses the qemu agent to determine the IP.  This requires that qemu-guest-agent be installed to the source VM.

### Tests written?
No

### Commits signed with GPG?
No

This isn't really a solution so much as a workaround, looks like the develop branch may take an entirely different approach to the cloning process and I haven't had time to see if the issue persists in that branch.  These are simply the changes I made to my local installation in order to get it working for the time being.  

There is likely a more elegant way to detect whether the qemu-agent is running on the target VM than sleeping for 20 seconds, but I can't think of a way other than checking the URL repeatedly, which still won't work if qemu-agent isn't installed.

EDIT:  After a bit more review, looks like the develop branch doesn't actually change this behavior, so this may still be relevant there.

### DIFF
--- a/salt/cloud/clouds/proxmox.py
+++ b/salt/cloud/clouds/proxmox.py
@@ -595,7 +595,7 @@ def create(vm_):
     host = data['node']       # host which we have received
     nodeType = data['technology']  # VM tech (Qemu / OpenVZ)
 
-    if not nodeType == 'qemu':
+    if not 'agent_get_ip' in vm_ or vm_['agent_get_ip'] == 0:
         # Determine which IP to use in order of preference:
         if 'ip_address' in vm_:
             ip_address = six.text_type(vm_['ip_address'])
@@ -605,8 +605,6 @@ def create(vm_):
             ip_address = six.text_type(data['private_ips'][0])  # first IP
         else:
             raise SaltCloudExecutionFailure("Could not determine an IP address to use")
-
-        log.debug('Using IP address %s', ip_address)
 
     # wait until the vm has been created so we can start it
     if not wait_for_created(data['upid'], timeout=300):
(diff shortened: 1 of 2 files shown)


salt/cloud/clouds/proxmox.py
```diff
@@ -585,6 +586,23 @@ def create(vm_):
     if not wait_for_state(vmid, 'running'):
         return {'Error': 'Unable to start {0}, command timed out'.format(name)}
 
+    # For QEMU VMs, we can get the IP Address from qemu-agent
+    if nodeType == 'qemu':
```
gtmanfred: This should behind a configuration gate, we don't require people to have qemu-guest-agent installed now, so we shouldn't force them to do this.

Also, this should probably go against develop instead of 2018.3, as it would be a new feature.

salt/cloud/clouds/proxmox.py
```diff
@@ -585,6 +586,23 @@ def create(vm_):
...
+        # We have to wait for a bit for qemu-agent to start
+        log.debug("Waiting for qemu-agent to start...")
+        time.sleep(20)
```
gtmanfred: I don't like this, we should just use the `cloud.wait_for_fun` utils module to know when the node is available and ready to return.

https://github.com/saltstack/salt/blob/2018.3/salt/cloud/clouds/openstack.py#L713  Here is an example.

gtmanfred (changes requested):

MasonEdgel:
@gtmanfred, I've made the suggested changes.  I didn't want to hijack the `agent` config setting, in case someone wanted to enable it but still use static IP, so I added an `agent_get_ip` setting..  I've also moved from `time.sleep` to `wait_for_fun`.

rallytime:
Hi @MasonEdgel - There are some merge conflicts with this change. Can you rebase this against the current HEAD of develop and fix up the conflicts?

MasonEdgel:
Sorry all! I'm still getting the hang of this PR thing...  I _think_ I've fixed the history issues.  This PR should now only contain my current changes on top of the develop branch.

gtmanfred:
Awesome, we are almost there, this is looking really good now, we just need to add `agent_get_ip` to the getting started with proxmox doc and then you have a couple pylint tests.

https://jenkinsci.saltstack.com/job/pr-lint/job/PR-50189/9/warnings52Result/new/

https://github.com/saltstack/salt/blob/develop/doc/topics/cloud/proxmox.rst

Thanks!

gtmanfred (approved):

gtmanfred:
Thanks @MasonEdgel this looks great.

terminalmage:
Ported to master in https://github.com/saltstack/salt/pull/56710

PR #50229, opened 2018-10-25
runners.pillar: trigger a sys.reload_modules after compilation

mattp-:
### What does this PR do?
pillar can tentatively be destructive in terms of replacing dunders like __grains__ on certain modules. firing a sys.gen_modules/reload_modules sets the masterminion back into a known state.

### Previous Behavior
generating pillar clobbered grains

### New Behavior
grains is maintained across runner.pillar calls

### Tests written?
No

### Commits signed with GPG?
No


cachedout (approved):
This seems sensible. Nice find!

PR #50230, opened 2018-10-25
Add various ssh tests to the filemap for salt/utils/vt.py changes

rallytime:
Some of the salt-ssh tests use the salt/utils/vt.py functions. A recent change in that file caused about 90 tests to start failing on the branch.

We need to make sure these tests run when changes to the vt.py util are made in PRs.


cachedout (approved):

s0undt3ch:
Port to master https://github.com/saltstack/salt/pull/56767

PR #50231, opened 2018-10-25
Revert "Back-port #50146 to 2018.3"

rallytime:
Reverts saltstack/salt#50174

This change is causing about 90 salt-ssh tests to fail. Reverting for now so we can track this down better later.


rallytime:
Refs #50146

PR #50235, opened 2018-10-25
Don't encode the return values in utils/vt.py

rallytime:
These encodings were causing the salt-ssh tests to fail.

This is a partial revert of #50146.

In order to see if the tests succeed completely on a PR, PR #50230 will need to be merged first.


rallytime:
This issue is no longer blocked as indicated above. It seems I can't remove that `blocked` designation.

MTecknology (approved):
The originally reported bug is still fixed. Thanks!

rallytime:
@MTecknology Awesome! Thank you for confirming!

gtmanfred (approved):

PR #50252, opened 2018-10-26
[2017.7] backport - Do not allow age to be a negative number

dwoz:
### What does this PR do?

Backport file cache bug fix found while fixing 2018.3 tests

### Tests written?

No

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/fileserver/__init__.py
+++ b/salt/fileserver/__init__.py
@@ -120,7 +120,20 @@ def check_file_list_cache(opts, form, list_cache, w_lock):
                 if os.path.exists(list_cache):
                     # calculate filelist age is possible
                     cache_stat = os.stat(list_cache)
-                    age = time.time() - cache_stat.st_mtime
+                    # st_time can have a greater precision than time, removing
+                    # float precision makes sure age will never be a negative
+                    # number.
+                    current_time = int(time.time())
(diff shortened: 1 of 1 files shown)


salt/fileserver/__init__.py
```diff
@@ -120,7 +120,20 @@ def check_file_list_cache(opts, form, list_cache, w_lock):
                 if os.path.exists(list_cache):
                     # calculate filelist age is possible
                     cache_stat = os.stat(list_cache)
-                    age = time.time() - cache_stat.st_mtime
+                    # st_time can have a greater precision than time, removing
+                    # float precision makes sure age will never be a negative
+                    # number.
```
cachedout: Could we just do an absolute value here instead?

dwoz: Originally when we saw age was able to be a negative number I thought it was due to `st_mtime` having greater floating point precision (which could make age negative). Later we found that there are also cases where the system time changes enough to make age negative. I think that means with the `file_mtime > current_time` check we don't need to cast things to `int` in the first place and the comment can be updated.

Otherwise, I'd prefer not to use abs since that could result in more cache updates (rather than less). The code would be simpler but less explicit. That is just my $0.02, using abs would work as you say.

rallytime (approved):

PR #50261, opened 2018-10-27
[develop] Adding beacon_module to configure multiple instances of beacons

garethgreenaway:
### What does this PR do?
Initial work on adding a beacon_type parameter for configuring multiple instances of beacons.

N/A

### Tests written?
Yes.

### Commits signed with GPG?
Yes


cachedout:
The word "type" feels really weird to me. To me, that distinguishes a class of beacons, several of which can be of the same type whereas here what we are distinguishing is a specific beacon. Would something like `beacon_module` be better?

garethgreenaway:
I'm good with `beacon_module`.

gtmanfred (approved):

cachedout (approved):

PR #50263, opened 2018-10-27
Add support for zabbix services in zabbix module and state.

johnybx:
### What does this PR do?
Adds support for manipulating zabbix services which is useful for calculating SLA and overall availability.  

Example state:
```
create_service_icmp-server-1:
  zabbix_service.present:
    - host: server-1
    - service_root: Server-group/server icmp
    - service_name: server-1
    - trigger_desc: is unavailable by ICMP
```
Result of this state would be (if no services exists yet) that service Server-group would be created with child "server icmp" and than it would have child server-1. Then we may have on this server also apache2 or nginx which should listen on port 443 so we could add service like this:

```
create_service_https-server-1:
  zabbix_service.present:
    - host: server-1
    - service_root: Server-group/server https
    - service_name: server-1
    - trigger_desc: port 443 is closed
```
So this would not create service Server-group but would add service "server https" with child server-1. Leaf services have associated trigger which is identified by name and based on this trigger is calculated availability in zabbix. Also we could just add new server to existing services (as child):
```
create_service_icmp-server-2:
  zabbix_service.present:
    - host: server-2
    - service_root: Server-group/server icmp
    - service_name: server-2
    - trigger_desc: is unavailable by ICMP
```
After this states we would end up with structure like this:

- Server-group
    -  server icmp
        -  server-1 -> associated with trigger
        -  server-2 -> associated with trigger
    -  server https
        -  server-1 -> associated with trigger

There is no limit on how many levels of services we can do. Trigger description in combination with trigger priority (severity) should be unique because for one service there can be only one trigger and first found match will be used.

There are two states present and absent. In module there are functions which create service, get service or delete server.

### DIFF
--- a/salt/modules/zabbix.py
+++ b/salt/modules/zabbix.py
@@ -402,7 +402,7 @@ def compare_params(defined, existing, return_old_value=False):
                                                                                        defined))
 
 
-def get_object_id_by_params(obj, params=None, **connection_args):
+def get_object_id_by_params(obj, params=None, **kwargs):
     '''
     .. versionadded:: 2017.7
 
@@ -418,7 +418,7 @@ def get_object_id_by_params(obj, params=None, **connection_args):
     '''
     if params is None:
         params = {}
-    res = run_query(obj + '.get', params, **connection_args)
+    res = run_query(obj + '.get', params, **kwargs)
     if res and len(res) == 1:
         return six.text_type(res[0][ZABBIX_ID_MAPPER[obj]])
     else:
@@ -427,7 +427,7 @@ def get_object_id_by_params(obj, params=None, **connection_args):
                             'Result: {2}'.format(obj + '.get', params, res))
 
 
-def apiinfo_version(**connection_args):
+def apiinfo_version(**kwargs):
     '''
     Retrieve the version of the Zabbix API.
 
@@ -444,7 +444,7 @@ def apiinfo_version(**connection_args):
 
         salt '*' zabbix.apiinfo_version
     '''
-    conn_args = _login(**connection_args)
+    conn_args = _login(**kwargs)
     ret = {}
     try:
         if conn_args:
@@ -458,7 +458,7 @@ def apiinfo_version(**connection_args):
         return False
 
 
-def user_create(alias, passwd, usrgrps, **connection_args):
+def user_create(alias, passwd, usrgrps, **kwargs):
     '''
     .. versionadded:: 2016.3.0
 
@@ -488,7 +488,7 @@ def user_create(alias, passwd, usrgrps, **connection_args):
 
         salt '*' zabbix.user_create james password007 '[7, 12]' firstname='James Bond'
     '''
-    conn_args = _login(**connection_args)
+    conn_args = _login(**kwargs)
     ret = {}
     try:
         if conn_args:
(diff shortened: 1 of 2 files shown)


salt/modules/zabbix.py
```diff
@@ -2503,3 +2503,187 @@ def configuration_import(config_file, rules=None, file_format='xml', **connectio
...
+def triggerid_get(hostid=None, trigger_desc=None, priority=4, **connection_args):
+    '''
+    .. versionadded:: Fluorine
+
+    Retrieve trigger ID and description based in host ID and trigger description.
+
+    .. note::
+        https://www.zabbix.com/documentation/3.4/manual/api/reference/trigger/get
+
+    :param hostid: ID of the host which's trigger we want to find
```
cachedout: s/which's/whose/

salt/modules/zabbix.py
```diff
@@ -2503,3 +2503,187 @@ def configuration_import(config_file, rules=None, file_format='xml', **connectio
...
+    :param trigger_desc: Description of trigger (trigger name) which's we want to find
```
cachedout: s/which's/whose/

salt/modules/zabbix.py
```diff
@@ -2503,3 +2503,187 @@ def configuration_import(config_file, rules=None, file_format='xml', **connectio
                                                                 'called successfully.'}
     except SaltException as exc:
         return {'name': config_file, 'result': False, 'message': six.text_type(exc)}
+
+
+def triggerid_get(hostid=None, trigger_desc=None, priority=4, **connection_args):
```
cachedout: We typically prefer named kwargs instead of `**connection_args`)

salt/modules/zabbix.py
```diff
@@ -2503,3 +2503,187 @@ def configuration_import(config_file, rules=None, file_format='xml', **connectio
...
+        salt '*' zabbix.triggerid_get 1111 'trigger name to find' 5
+    '''
+    conn_args = _login(**connection_args)
+    ret = False
+    try:
+        if conn_args:
+            method = 'trigger.get'
+            if not hostid or not trigger_desc:
+                return {'result': False, 'comment': 'hostid and trigger_desc params are required'}
+            params = {'output': ["triggerid", "description"],
```
cachedout: Our style guide recommends single quotes here and on the line below.

salt/modules/zabbix.py
```diff
@@ -2503,3 +2503,187 @@ def configuration_import(config_file, rules=None, file_format='xml', **connectio
...
+
+    :return: Service details, False if service could not be added or on failure.
+
+    CLI Example:
+    .. code-block:: bash
+
+        salt '*' zabbix.service_add 11 'My service' 11111
+    '''
+    conn_args = _login(**connection_args)
+    ret = False
```
cachedout: It looks like here and in other places we're returning either a boolean or a dictionary. Should it instead be an empty dict or a popualated dict? That might avoid type errors for consumers of this module.

johnybx: good idea. I changed it in whole module.

salt/modules/zabbix.py
johnybx: good idea. I changed it in whole module.

johnybx:
Hi @cachedout, I did requested changes.

salt/modules/zabbix.py
```diff
@@ -427,7 +427,7 @@ def get_object_id_by_params(obj, params=None, **connection_args):
                             'Result: {2}'.format(obj + '.get', params, res))
 
 
-def apiinfo_version(**connection_args):
+def apiinfo_version(**kwargs):
```
cachedout: Could you clarify why this change and the others like it are necessary, please?

PR #50267, opened 2018-10-28
Fix missing aptly config paths

m03:
### What does this PR do?

* Passes the config_path that was provided to the function that is calling `get_repo` on to `get_repo`.
* Adds a missing CLI example comment.

There are no open issues for this. I noticed this omission while working on adding additional functionality to the aptly module.

### Previous Behavior

The calls to `get_repo` were always using the default config path instead of the one provided to the calling function.

### New Behavior

The calls to `get_repo` will now correctly use the config path provided to the calling function.

### Tests written?

No

### Commits signed with GPG?

Yes


m03:
The test failures on occurred on the Windows Jenkins jobs and don't appear to be related to my PR.

rallytime (approved):

PR #50270, opened 2018-10-29
apt: speedup installing packages with '>=' or '<=' and reduced memory…

terrible-broom:
… consumption

### What does this PR do?
Speedup salt work and reduce memory comsumption in aptpkg.py when installing packages with operators '>=' and '<='.

### Previous Behavior
For each pkg.installed with '>=' or '<=' salt saved output of "apt-cache dump" to variable and then read it. But this output could be huge. For example, on my system it took 47 Gb, but I didn't have 47 Gb of memory, so salt couldn't work.

### New Behavior
Salt just saves output of "apt-cache show <pkg>" if pkg is given, or full output: "apt-cache show .". It is much shorter than "apt-cache dump" (380 Mb on the same system) but contains the same useful information.

### Tests written?

No

### Commits signed with GPG?

No

### DIFF
--- a/salt/modules/aptpkg.py
+++ b/salt/modules/aptpkg.py
@@ -18,7 +18,6 @@ from __future__ import absolute_import, print_function, unicode_literals
 
 # Import python libs
 import copy
-import fnmatch
 import os
 import re
 import logging
@@ -1660,11 +1659,11 @@ def list_repo_pkgs(*args, **kwargs):  # pylint: disable=unused-import
         salt '*' pkg.list_repo_pkgs foo bar baz
     '''
     if args:
-        # get only information about packages in args
-        cmd = ['apt-cache', 'show' ] + [arg for arg in args]
+        # Get only information about packages in args
+        cmd = ['apt-cache', 'show'] + [arg for arg in args]
     else:
-        # get information about all available packages
-        cmd = ['apt-cache', 'show', '.']
+        # Get information about all available packages
+        cmd = ['apt-cache', 'dump']
 
     out = __salt__['cmd.run_all'](
         cmd,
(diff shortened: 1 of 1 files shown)


salt/modules/aptpkg.py
```diff
@@ -1659,8 +1659,15 @@ def list_repo_pkgs(*args, **kwargs):  # pylint: disable=unused-import
         salt '*' pkg.list_repo_pkgs
         salt '*' pkg.list_repo_pkgs foo bar baz
     '''
+    if args:
+        # get only information about packages in args
+        cmd = ['apt-cache', 'show' ] + [arg for arg in args]
+    else:
+        # get information about all available packages
+        cmd = ['apt-cache', 'show', '.']
```
isbm:
```python
cmd = ['apt-cache', 'show' ] + [arg for arg in args or ['.']]
```

But I don't like you use `apt-cache show .` here. It is super-slow, apparently, and doesn't speeds this up. If you use `apt-cache dump`, it will get all packages ~100Mb data about 2-3 seconds. The other one will do really long. The only reason I see to use `apt-cache show` is when you want to see specific packages, so no "skip thing" logic you've deleted before. I would then use both commands: "dump" for just everything and "show" only for specific packages:

```python
cmd = ['apt-cache']
if args:
    cmd.expand(['show'] + [arg for arg in args])
else:
    cmd.append('dump')
```

salt/modules/aptpkg.py
```diff
@@ -1671,29 +1678,19 @@ def list_repo_pkgs(*args, **kwargs):  # pylint: disable=unused-import
     skip_pkg = False
     new_pkg = re.compile('^Package: (.+)')
     for line in salt.utils.itertools.split(out['stdout'], '\n'):
+        if len(line.strip()) == 0:
+            continue
```
isbm:
```python
if not line.strip():
    continue
```

isbm (changes requested):

salt/modules/aptpkg.py
isbm: Here is the proof: I don't know how this is ever good, when the entire core of the CPU is 100% full for another couple of minutes.

```
$ time apt-cache show . > /dev/null

real    2m24.267s
user    0m32.576s
sys     1m49.792s

```
And other story with just dumping it:
```
$ time apt-cache dump > /dev/null                                                           

real    0m1.182s
user    0m0.852s
sys     0m0.324s
```

This is just no-go.

rallytime (approved):

PR #50272, opened 2018-10-29
Fix out of order docker Env comparison

jyurdal:
### What does this PR do?
I have experienced bug on windows only minions, regarding docker_container.running, where container is replaced on each state.apply whether it has changes or not.

If there are many "environment" variables specified, the environment list is having random order each time state is applied. As a result existing container and the temporarily created container are seen to have different environment variables ( although only order is different) and docker container is replaced each time salt is doing state.apply.

Argument reorder happens only on windows and it happens in salt/utils/docker/init.py function translate_input) . (Maybe it would make sense to see why it happens in translate_input. As for my system I just compare sorted list instead.) It has been introduced between 2017.7 and 2018.3.0

### Previous Behavior
docker container placed on each state.apply despite no configuration changes.

### New Behavior
docker container modified only on changes.

### Tests written?

No

### Commits signed with GPG?

No

### DIFF
--- a/salt/modules/dockermod.py
+++ b/salt/modules/dockermod.py
@@ -5589,6 +5589,7 @@ def pause(name):
                             .format(name))}
     return _change_state(name, 'pause', 'paused')
 
+
 freeze = salt.utils.functools.alias_function(pause, 'freeze')
 
 
@@ -5790,6 +5791,7 @@ def unpause(name):
                 'comment': ('Container \'{0}\' is stopped, cannot unpause'
                             .format(name))}
     return _change_state(name, 'unpause', 'running')
+
 
 unfreeze = salt.utils.functools.alias_function(unpause, 'unfreeze')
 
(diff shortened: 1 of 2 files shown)


cachedout:
@jyurdal Can you please fix these lint warnings? https://jenkinsci.saltstack.com/job/pr-lint/job/PR-50272/3/warnings52Result/new/

terminalmage (approved):
First of all, thanks for the pull request! This is a good catch.

I would however like to clear some things up. We're not actually changing any order when we translate input. We are placing the env vars into a dictionary to be passed to docker-py for container creation. Under the hood, it is likely dictionary iteration order which decides which order the environment variables are set, because the container's metadata contains the env vars as a list of `KEY=value` pairs.

I took the liberty of adding a unit test for this and pushed it to this PR.

terminalmage:
@rallytime this will need a backport to 2018.3.

PR #50290, opened 2018-10-29
Fix `test_state` for Windows

twangboy:
### What does this PR do?
Fixes the following state on Windows:
integration.modules.test_state.StateModuleTest.test_onchanges_requisite_no_state_module

The `atomicfile` close method was using `shutil.copymode` to copy permissions from the original file to the new one. This is useless in Windows. It was also crashing because one of the users specified in the security descriptor didn't exist on the machine.

This PR creates a salt util that will copy the Security information correctly. This salt util is then used in place of `shutil.copymode` on Windows.

Found in branch tests for 2018.3

### Tests written?
No

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/utils/atomicfile.py
+++ b/salt/utils/atomicfile.py
@@ -122,11 +122,11 @@ class _AtomicWFile(object):
             return
         self._fh.close()
         if os.path.isfile(self._filename):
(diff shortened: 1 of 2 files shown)


dwoz (approved):
LGTM

Issue #50286, opened 2018-10-29
Multiple issues with libcloud_dns


max-arnold:
### Description of Issue/Question

I found so many issues while trying to use `libcloud_dns`, that it looks like it was never tested at all.

1. Documentation doesn't mention that libcloud is not a cloud module. Using terms like `profile` and `providers` is really confusing, and it looks like I'm not the only one who spent hours on this: #44390

2. The example just doesn't work:

https://docs.saltstack.com/en/latest/ref/states/all/salt.states.libcloud_dns.html

```
my-zone:
  libcloud_dns.zone_present:
    - name: mywebsite.com
    - profile: profile1


local:
----------
          ID: my-zone
    Function: libcloud_dns.zone_present
        Name: mywebsite.com
      Result: False
     Comment: Missing parameter domain for state libcloud_dns.zone_present
              Missing parameter type for state libcloud_dns.zone_present
     Changes:
```

3. Once I replaced `name` with `domain` and added `type: master` I've got this:

```
  File "/Users/user/.virtualenvs/salt/lib/python3.6/site-packages/salt/states/libcloud_dns.py", line 94, in zone_present
    result = __salt__['libcloud_dns.create_zone'](domain, profile, type)
  File "/Users/user/.virtualenvs/salt/lib/python3.6/site-packages/salt/modules/libcloud_dns.py", line 213, in create_zone
    zone = conn.create_record(domain, type=type, ttl=ttl)
TypeError: create_record() got an unexpected keyword argument 'ttl'
```

This is because `create_zone` calls `conn.create_record` instead of `conn.create_zone`: https://github.com/saltstack/salt/blob/develop/salt/modules/libcloud_dns.py#L213

Unit tests are nominally present, but unable to catch this, because they just test for `True` and don't care which function was actually invoked: https://github.com/saltstack/salt/blob/develop/tests/unit/states/test_libcloud_dns.py#L123

4. I fixed this too, then I got the red warning while running it with `test=True`:

```
    Warnings: 'name' is an invalid keyword argument for
              'libcloud_dns.zone_present'. If you were trying to pass additional
              data to be used in a template context, please populate 'context'
              with 'key: value' pairs. Your approach will work until Salt
              Fluorine is out. Please update your state files.
```

Maybe it indeed was a bad idea to rename `name` to `domain`?

[...]

PR #49832, opened 2018-09-29
Replace pchanges with changes to make onchanges/prereq work in test mode

terminalmage:
Fixes https://github.com/saltstack/salt/issues/49753
Fixes https://github.com/saltstack/salt/issues/45895

PR #32694, opened 2016-04-19
Libcloud DNS execution module

tonybaloney:
### What issues does this PR fix or reference?

This execution module is a wrapper for libcloud's DNS abstraction library. http://libcloud.readthedocs.org/en/latest/dns/index.html

Libcloud currently supports 23 DNS providers across the globe, this module allows you to configure various providers and collect information on zones, records and create new zones or records. 
### Tests written?

No

CC @techhat

PR #50296, opened 2018-10-29
Address issues highlighted in #50286 for libcloud_dns

tonybaloney:
### What does this PR do?

Fixes a number of issues in the `libcloud_dns` state and execution modules

- [x] Documentation doesn't mention that libcloud is not a cloud module
- [x]  `create_zone` didn't work, called the wrong method and was missing a kwarg
- [x] Bug in Create_record
- [x] Incorrect test methods
- [x] 'name' is an invalid keyword argument for 'libcloud_dns.zone_present'
- [x] test argument not passed
- [x] TypeError: argument of type 'NoneType' is not iterable

#50286 

### Tests written?

Yes

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/states/libcloud_dns.py
+++ b/salt/states/libcloud_dns.py
@@ -210,7 +210,7 @@ def record_absent(name, zone, type, data, profile):
     if len(matching_records) > 0:
         result = []
         if __opts__['test']:
-            return state_result(True, 'Removed {0} records'.format(len(matching_records)))
+            return state_result(True, 'Removed {0} records'.format(len(matching_records)), name)
         else:
             for record in matching_records:
                 result.append(__salt__['libcloud_dns.delete_record'](
--- a/salt/states/libcloud_dns.py
+++ b/salt/states/libcloud_dns.py
@@ -72,7 +72,7 @@ def state_result(result, message, name, changes=None):
             'changes': changes}
 
 
-def zone_present(domain, type, profile):
+def zone_present(domain, type, profile, ttl=None, extra=None):
     '''
     Ensures a record is present.
 
@@ -84,6 +84,12 @@ def zone_present(domain, type, profile):
 
     :param profile: The profile key
     :type  profile: ``str``
+
+    :param ttl: TTL for new records. (optional)
+    :type  ttl: ``int``
+
+    :param extra: Extra data (optional)
+    :type  extra: ``dict``
     '''
     zones = __salt__['libcloud_dns.list_zones'](profile)
     if not type:
@@ -92,7 +98,8 @@ def zone_present(domain, type, profile):
     if len(matching_zone) > 0:
         return state_result(True, 'Zone already exists', domain)
     else:
-        result = __salt__['libcloud_dns.create_zone'](domain, profile, type)
+        result = __salt__['libcloud_dns.create_zone'](
+            domain=domain, profile=profile, type=type, ttl=ttl, extra=extra)
         return state_result(True, 'Created new zone', domain, result)
 
 
--- a/salt/states/libcloud_dns.py
+++ b/salt/states/libcloud_dns.py
@@ -128,7 +128,7 @@ def zone_absent(domain, profile):
             return state_result(result, 'Deleted zone', domain)
 
 
-def record_present(name, zone, type, data, profile):
+def record_present(name, zone, type, data, profile, extra=None):
     '''
     Ensures a record is present.
 
@@ -149,6 +149,9 @@ def record_present(name, zone, type, data, profile):
 
     :param profile: The profile key
     :type  profile: ``str``
+
+    :param extra: Extra data (optional)
+    :type  extra: ``dict``
     '''
     zones = __salt__['libcloud_dns.list_zones'](profile)
     try:
@@ -166,7 +169,7 @@ def record_present(name, zone, type, data, profile):
         else:
             result = __salt__['libcloud_dns.create_record'](
                 name, matching_zone['id'],
-                type, data, profile)
+                type, data, profile, extra=extra)
             return state_result(True, 'Created new record', name, result)
     else:
         return state_result(True, 'Record already exists', name)
--- a/salt/states/libcloud_dns.py
+++ b/salt/states/libcloud_dns.py
@@ -72,7 +72,7 @@ def state_result(result, message, name, changes=None):
             'changes': changes}
 
 
-def zone_present(domain, type, profile, ttl=None, extra=None):
+def zone_present(domain, profile, type=None, ttl=None, extra=None):
     '''
     Ensures a record is present.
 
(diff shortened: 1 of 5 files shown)


tonybaloney:
PR Complete pending tests

tonybaloney:
Tests and linter passing.
Ready for peer review

max-arnold:
I'll test this

max-arnold:
I didn't test this PR completely, but since I found a few issues I decided it would be better to provide some feedback for faster iteration. I did all my tests on 2018.3.3 with your patch applied.

1. Do you mind rebasing this patch against at least 2018.3 or even 2017.7 as the doc suggests? https://docs.saltstack.com/en/latest/topics/development/contributing.html#salt-s-branch-topology.  The `develop` branch will be released somewhere next year, and this means that the bug will be present in all current releases and upcoming `2018.11`.  More info on supported versions is here https://s.saltstack.com/product-support-lifecycle/

2. The documentation examples are still the same (`name` vs `domain`, missing `type` and the issue with `ID 'my-zone' in SLS 'dns' contains multiple state declarations of the same type`)

And maybe `type` should be optional.

3. `test=True` seems to work, but doesn't report expected changes in the same way as other states do (red/yellow/green)

Current output with `test=True`:

```
Summary for local
------------
Succeeded: 1
Failed:    0
------------
```

Expected output (ignore the numbers, what matters is `changed` and `unchanged`)

```
Summary for example.com
------------
Succeeded: 2 (unchanged=2, changed=1)
Failed:    0
------------
```

I'm not familiar with how it works internally, but have seen the PR which touches this (it looks like there is some refactoring going on and `pchanges` is removed).  You may find it useful: https://github.com/saltstack/salt/pull/49832. A quote:

```
TL;DR is that `changes` should show in test mode, if there are changes expected
```

https://docs.saltstack.com/en/latest/ref/states/writing.html#return-data

4. Number (7) from #50286 isn't fixed:

```
                File "/Users/user/.virtualenvs/salt/lib/python3.6/site-packages/salt/modules/libcloud_dns.py", line 289, in create_record
                  type=record_type, data=data, extra=extra))
                File "/Users/user/.virtualenvs/salt/lib/python3.6/site-packages/libcloud/dns/drivers/gandi.py", line 221, in create_record
                  if 'ttl' in extra:
```

Some providers expect `extra` to be a dict (not `None`), so you need to add the `extra = extra or {}` before passing it further down the stack.

5. And I still get this warning:

```
local:
----------
          ID: my-zone
    Function: libcloud_dns.zone_present
      Result: True
     Comment: Created new zone
     Started: 19:03:57.478894
    Duration: 1648.672 ms
     Changes:
              ----------
              domain:
                  mywebsite.com
              extra:
                  ----------
              id:
                  689357
              ttl:
                  None
              type:
                  master
    Warnings: 'name' is an invalid keyword argument for
              'libcloud_dns.zone_present'. If you were trying to pass additional
              data to be used in a template context, please populate 'context'
              with 'key: value' pairs. Your approach will work until Salt
              Fluorine is out. Please update your state files.
```

Not quite sure about it, will test later on the `develop` branch. 

Could you consider changing the `domain` state parameter back to `name`? It seems to be the common convention for most states. Also many people use implicit name (inherited from state id).

6. Libcloud versions

It would be good to figure out which versions of libcloud are compatible with new changes and communicate this to Salt packaging team. Otherwise, people may end up installing new salt with old libcloud and get all sorts of breakages.

Salt-bootstrap wants `0.14.0` or greater https://github.com/saltstack/salt-bootstrap/blob/develop/bootstrap-salt.sh#L253. It may be necessary to raise it for all `libcloud_*.py` modules (right now they want `REQUIRED_LIBCLOUD_VERSION = '2.0.0'`).

Also, it makes sense to implement the same `__virtual__` checks for the `states/libcloud_*.py` modules (because most other state modules do).

rallytime (approved):

salt/states/libcloud_dns.py
```diff
@@ -71,7 +72,7 @@ def state_result(result, message, name, changes=None):
             'changes': changes}
 
 
-def zone_present(domain, type, profile):
+def zone_present(domain, profile, type=None, ttl=None, extra=None):
```
gtmanfred: name is a required argument for all state functions, i would just have name as the first argument, and then add domain on at the end to keep people be able to use domain: in their states, and if domain is set, use that instead of name.

salt/states/libcloud_dns.py
```diff
@@ -83,16 +84,26 @@ def zone_present(domain, type, profile):
...
-        return state_result(True, 'Created new zone', domain, result)
+        if __opts__['test']:
+            return state_result(True, 'Created new zone', domain)
+        else:
+            result = __salt__['libcloud_dns.create_zone'](
+                domain=domain, profile=profile, type=type, ttl=ttl, extra=extra)
+            return state_result(True, 'Created new zone', domain, result)
 
 
 def zone_absent(domain, profile):
```
gtmanfred: name is required here too

gtmanfred (changes requested):
I agree that we should increase the minimum libcloud version to 2.0.0,. but that should be an issue opened against salt-bootstrap.

The `name` thing needs to be fixed, and we should do #4 from @max-arnold s comment.

max-arnold:
> I agree that we should increase the minimum libcloud version to 2.0.0,. but that should be an issue opened against salt-bootstrap.

I did some research, but haven't figured out when (i.e. exact version number) the libcloud API changed so dramatically that Salt modules became broken: https://github.com/apache/libcloud/tree/trunk/libcloud/dns/drivers

The salt/libcloud modules were introduced in April 2016 (#32694). Libcloud version around that time probably was either 1.5.0 or 2.0.0 https://github.com/apache/libcloud/releases

In my quick tests, this PR works (with the problems mentioned earlier in this thread) both with `libcloud==2.0.0` and `2.3.0`, so I think pinning '>=2.0.0` should be safe.

max-arnold:
Also, It could open up another rabbit hole, but some of the `libcloud_*` modules depend on `1.5.0` and some on `2.0.0`:

2.0.0: https://github.com/saltstack/salt/blob/develop/salt/modules/libcloud_compute.py#L48
2.0.0: https://github.com/saltstack/salt/blob/develop/salt/modules/libcloud_dns.py#L43
1.5.0: https://github.com/saltstack/salt/blob/develop/salt/modules/libcloud_loadbalancer.py#L47
1.5.0: https://github.com/saltstack/salt/blob/develop/salt/modules/libcloud_storage.py#L46

tonybaloney:
@max-arnold I'm going to open a new PR against 2018.3 instead of doing all sorts of git-foo to rebase without requesting a merge-backwards of develop

tonybaloney:
Replaced by https://github.com/saltstack/salt/pull/50321

tonybaloney:
> > I agree that we should increase the minimum libcloud version to 2.0.0,. but that should be an issue opened against salt-bootstrap.
> 
> I did some research, but haven't figured out when (i.e. exact version number) the libcloud API changed so dramatically that Salt modules became broken: https://github.com/apache/libcloud/tree/trunk/libcloud/dns/drivers
> 
> The salt/libcloud modules were introduced in April 2016 (#32694). Libcloud version around that time probably was either 1.5.0 or 2.0.0 https://github.com/apache/libcloud/releases
> 
> In my quick tests, this PR works (with the problems mentioned earlier in this thread) both with `libcloud==2.0.0` and `2.3.0`, so I think pinning '>=2.0.0` should be safe.

Basically, the difference between 1.5 and 2.0 of libcloud was the switch to requests as the HTTP framework instead of httplib. This shouldn't impact these modules

PR #50302, opened 2018-10-30
cmd: Add root parameter for wait and run states

aplanas:
### What does this PR do?
`run_chroot` module.

### DIFF
--- a/salt/states/cmd.py
+++ b/salt/states/cmd.py
@@ -403,6 +403,7 @@ def wait(name,
          unless=None,
          creates=None,
          cwd=None,
+         root=None,
          runas=None,
          shell=None,
          env=(),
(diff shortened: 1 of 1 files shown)


salt/states/cmd.py
```diff
@@ -917,8 +928,9 @@ def run(name,
 
     # Wow, we passed the test, run this sucker!
     try:
-        cmd_all = __salt__['cmd.run_all'](
-            name, timeout=timeout, python_shell=True, **cmd_kwargs
+        run_cmd = 'cmd.run_all' if not root else 'cmd.run_chroot'
+        cmd_all = __salt__[run_cmd](
```
cachedout: Given that we are calling into a few other modules if we chroot here, we may need to expand our exception handling.

aplanas: Right.

aplanas: I change it to Exception, not sure if is too generic.

cachedout (approved):

PR #50306, opened 2018-10-30
Add 'noaction' flag for install/remove in the opkg execution module 

rares-pop:
### What does this PR do?
Adding the 'noaction' flags when installing/removing software. Basically it's just testing if the operations will succeed or not without actually doing the operation.

### Previous Behavior
opkg wasn't providing this functionality.

### New Behavior
you can pass the noaction=True flag

### Tests written?

Yes

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/modules/opkg.py
+++ b/salt/modules/opkg.py
@@ -368,7 +368,7 @@ def install(name=None,
 
     if bool(kwargs.get('test') or __opts__.get('test')):
         cmd_prefix.append('--noaction')
-    
+
     if pkg_params is None or len(pkg_params) == 0:
         return {}
     elif pkg_type == 'file':
--- a/salt/modules/opkg.py
+++ b/salt/modules/opkg.py
@@ -366,6 +366,8 @@ def install(name=None,
     to_reinstall = []
     to_downgrade = []
 
+    if kwargs.get('noaction', False):
+        cmd_prefix.append('--noaction')
     if pkg_params is None or len(pkg_params) == 0:
         return {}
     elif pkg_type == 'file':
@@ -540,6 +542,8 @@ def remove(name=None, pkgs=None, **kwargs):  # pylint: disable=unused-argument
     if not targets:
         return {}
     cmd = ['opkg', 'remove']
+    if kwargs.get('noaction', False):
+        cmd.append('--noaction')
     if kwargs.get('remove_dependencies', False):
         cmd.append('--force-removal-of-dependent-packages')
     if kwargs.get('auto_remove_deps', False):
--- a/salt/modules/opkg.py
+++ b/salt/modules/opkg.py
@@ -366,8 +366,9 @@ def install(name=None,
     to_reinstall = []
     to_downgrade = []
 
-    if kwargs.get('noaction', False):
+    if bool(kwargs.get('test') or __opts__.get('test')):
         cmd_prefix.append('--noaction')
+    
     if pkg_params is None or len(pkg_params) == 0:
         return {}
     elif pkg_type == 'file':
@@ -542,7 +543,7 @@ def remove(name=None, pkgs=None, **kwargs):  # pylint: disable=unused-argument
     if not targets:
         return {}
     cmd = ['opkg', 'remove']
-    if kwargs.get('noaction', False):
+    if bool(kwargs.get('test') or __opts__.get('test')):
         cmd.append('--noaction')
     if kwargs.get('remove_dependencies', False):
         cmd.append('--force-removal-of-dependent-packages')
--- a/salt/modules/opkg.py
+++ b/salt/modules/opkg.py
@@ -269,6 +269,14 @@ def refresh_db(failhard=False, **kwargs):  # pylint: disable=unused-argument
     return ret
 
 
+def _append_noaction_if_testmode(cmd, **kwargs):
+    '''
+    Adds the --noaction flag to the command if it's running in the test mode.
+    '''
+    if bool(kwargs.get('test') or __opts__.get('test')):
+        cmd.append('--noaction')
+
+
 def install(name=None,
             refresh=False,
             pkgs=None,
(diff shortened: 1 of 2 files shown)


cachedout (approved):
LGTM

isbm (changes requested):
But why? Isn't that we have `test=True` for this instead?
@cachedout while code looks OK, it is conceptually weird. This should be incorporated with the `test=True` instead.

rares-pop:
@isbm - I am not aware of test=True. What is it?

I browsed dpkg.py and apk.py and I haven't seen anything interesting.

Are you saying that we shouldn't go with 'noaction' since that's very opkg specific and say 'test' instead?

isbm:
@rares-pop this is a test mode, like dry run. This is not really working if module does not respects that (and you also don't know it does respect it or not...). Basically, you're looking for `__opts__['test']`, and if it is there and `True`, then you no-op. So I would rather drop your own parameter and simply incorporate that path instead. Example you can fine [here](https://github.com/saltstack/salt/blob/develop/salt/modules/aptpkg.py#L1125-L1134), [here](https://github.com/saltstack/salt/blob/develop/salt/modules/aptpkg.py#L1200-L1208), [here](https://github.com/saltstack/salt/blob/develop/salt/states/aptpkg.py#L44-L53) or [here](https://github.com/saltstack/salt/blob/develop/salt/states/selinux.py#L117-L123). Or [here](https://github.com/saltstack/salt/blob/develop/salt/states/pkg.py#L1637-L1679).

rares-pop:
@isbm - __opts__ is the minion configuration, right? Can you specify that option (__opts__['test']) on a salt-call basis?

i.e. the sequence:
salt-call --local pkg.install 'foo'
salt-call --local pkg.install 'bar' test=True
salt-call --local pkg.install 'baz'

isbm:
> @isbm - optsis the minion configuration, right? Can you specify that option (**opts**['test']) on a salt-call basis?
> 
> i.e. the sequence:
> salt-call --local pkg.install 'foo'
> salt-call --local pkg.install 'bar' test=True
> salt-call --local pkg.install 'baz'

Generally, you should be able to incorporate this with the testing mode, described here:
https://docs.saltstack.com/en/latest/ref/states/testing.html

rares-pop:
@isbm - I don't want to go through state modules. I need to do a simple salt-call --local and pass that argument.

salt-call doesn't seem to put its arguments inside __opts__, but only in kwargs.

Can you be more specific on how I can achieve that with what you suggest?

isbm:
First of all, you _have_ to. :wink: It is a virtual module that is also used in states. So if `state.apply` gets it in test mode, your stuff **must** cope with it.

Second, if you want that _also_ in the module call it do nothing, then just add `**kwargs` to the signature (or `test=False` directly) and use it accordingly, so the `test=true` mode is still there. Problem is that there is no much of standard so far and we're working to sort this out. However, let's not do another ideological forks. :wink:

rares-pop:
@isbm - sorry for the back and forth, but just to be sure to implement the correct thing:
- salt/modules/opkg.py already has the kwargs in its signatures
- given the current PR, you want me to rename the 'noaction' thing to 'test' and in addition, also check the '__opkg__['test']

Is that correct?

If that's correct, if things contradict (__opts__['test'] = False but kwargs contain test=True'), which should have priority? Probably __opts__ if it's set to true, otherwise kwargs if present?

isbm:
> @isbm - sorry for the back and forth, but just to be sure to implement the correct thing:
> 
> * salt/modules/opkg.py already has the kwargs in its signatures

Then just pull "test" out of them:

```python
 if bool(kwargs.get('test') or __opt__.get('test')):
        cmd_prefix.append('--noaction')
```

tests/unit/modules/test_opkg.py
```diff
@@ -143,6 +143,23 @@ def test_install(self):
...
+            mock = MagicMock(return_value=ret_value)
+            patch_kwargs = {
+                '__salt__': {
+                    'cmd.run_all': mock,
+                    'pkg_resource.parse_targets': MagicMock(return_value=({'vim': '7.4'}, 'repository')),
+                    'restartcheck.restartcheck': MagicMock(return_value='No packages seem to need to be restarted')
+                }
+            }
+            with patch.multiple(opkg, **patch_kwargs):
+                self.assertEqual(opkg.install('vim:7.4', noaction=True), {})
```
isbm: I guess here should be also some sort of update then.

tests/unit/modules/test_opkg.py
```diff
@@ -160,6 +177,23 @@ def test_remove(self):
...
+            mock = MagicMock(return_value=ret_value)
+            patch_kwargs = {
+                '__salt__': {
+                    'cmd.run_all': mock,
+                    'pkg_resource.parse_targets': MagicMock(return_value=({'vim': '7.4'}, 'repository')),
+                    'restartcheck.restartcheck': MagicMock(return_value='No packages seem to need to be restarted')
+                }
+            }
+            with patch.multiple(opkg, **patch_kwargs):
+                self.assertEqual(opkg.remove('vim:7.4', noaction=True), {})
```
isbm: And here.

isbm (changes requested):

rares-pop:
Oops.. I kept the 'noaction' in the test name, because that's really what it should happen, even if the actual parameter is 'test=True'. More than that, test_install_test sounds weird.

salt/modules/opkg.py
```diff
@@ -269,6 +269,14 @@ def refresh_db(failhard=False, **kwargs):  # pylint: disable=unused-argument
     return ret
 
 
+def _append_noaction_if_testmode(cmd, **kwargs):
```
isbm: You can call it just `_noaction` :wink: but this is also OK.

isbm (approved):

PR #50317, opened 2018-10-30
Beacon to check on status endpoint

shaidar:
### What does this PR do?
Beacon to manage and report the status of a server status endpoint. It fires an event when specified values in pillar don't match returned response.

Our particular use case for this is to use it in conjunction with the [django-server-status](https://pypi.org/project/django-server-status/) and verify that specified services in pillar are functional. 

However, this is not restricted to that package and can be used with others.

Don't believe there's an existing issue. Wrote it to resolve a problem we were facing.

### Tests written?

No

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/beacons/http_status.py
+++ b/salt/beacons/http_status.py
@@ -1,0 +1,142 @@
+# -*- coding: utf-8 -*-
+'''
+Beacon to manage and report the status of a server status endpoint.
+Fire an event when specified values don't match returned response.
+
+'''
+
+# Import Python libs
+from __future__ import absolute_import, unicode_literals
+import logging
+import operator
+import re
+import requests
+import salt.utils.data
+from salt.ext.six.moves import map
+
+log = logging.getLogger(__name__)
+
+__virtualname__ = 'http_status'
+
+comparisons = {
+    '==': operator.eq,
+    '<': operator.lt,
+    '>': operator.gt,
+    '<=': operator.le,
+    '>=': operator.ge,
+    '!=': operator.ne,
+    'search': re.search
+}
+
+
+def __virtual__():
+    return __virtualname__
+
+
+def validate(config):
+    '''
+    Validate the beacon configuration
+    '''
+    if not isinstance(config, list):
+        return False, ('Configuration for %s beacon must '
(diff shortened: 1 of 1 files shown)


rares-pop (commented):
There should be more pythonic ways to do the validation - you shouldn't have to do if 'foo' in dict followed by if isinstance dict['foo'] ..., but rather if isinstance(dict.get('foo', None), ....
You do the validation but is partial (i.e. you don't check for 'comp' to be valid when you do the validation).
You say nothing about html_response.
Do you want to check if the whole configuration is valid (i.e. the 'comp' is valid in all cases)

cachedout:
This is one of those cases where I think that CodeClimate has some very valid points. I think it might be worth refactoring this a bit to avoid the deeply nested control statements.

salt/beacons/http_status.py
```diff
@@ -0,0 +1,158 @@
...
+            valid = False
+            messages.append('[-] Sites for {} beacon requires {}'.format(__virtualname__,
+                                                                         required_site_attributes))
+        log.debug('[+] site: %s', site)
+        log.debug('[+] settings: %s', settings)
+
+        for optional_attrs in itertools.chain(settings.get(attr, []) for attr in optional_site_attributes):
+            for item in optional_attrs:
+                cmp = item.get('comp')
+                if cmp and not cmp in comparisons:
```
cachedout: `if cmp and cmp not in comparisons` to satisfy the linter, please.

cachedout (changes requested):

cachedout (approved):

max-arnold:
@shaidar Interesting feature, however I found a couple of issues:

* The configuration example uses the `content-type` key, but the actual implementation contains `site_config.pop('content_type', 'json')` (note the underscore)
* The `content_type` variable seems to be unused (has no effect)
* It seems to be impossible to check if a html response contains some text, because it is always gets decoded as json
* It looks like the `if r.raise_for_status` check should be used with parentheses `()`, because it is a function, and also moved inside the `try/except` block
* No tests
* The event format is undocumented
* No release notes

PR #50318, opened 2018-10-30
salt/dpkg.py: fix path to package.list for different architectures

rpasche:
The 'install_date' is missing for many packages, because the <package>.list file sometimes
contains the architecture seperated by ':'. So first try this file location and
fall back, if that did not work.

### What does this PR do?
It fixes the ```install_date``` for debian packages that is missing for many packages because the file that is used to get the ```install_date``` is wrong.

None found. This issue has been found while trying to integrate Debian systems as Salt clients into Uyuni server.

### Previous Behavior
```install_date``` missing for many packages

### New Behavior
```install_date``` returned for every package

### Tests written?

No (sorry.)

### Commits signed with GPG?

No

### DIFF
--- a/salt/modules/dpkg.py
+++ b/salt/modules/dpkg.py
@@ -342,16 +342,23 @@ def _get_pkg_install_time(pkg, arch=None):
     :return:
     '''
     iso_time = None
+    loc_root = '/var/lib/dpkg/info'
     if pkg is not None:
+        locations = []
         if arch is not None and arch != 'all':
-            location = "/var/lib/dpkg/info/{0}:{1}.list".format(pkg, arch)
-            if os.path.exists(location):
-                iso_time = datetime.datetime.utcfromtimestamp(int(os.path.getmtime(location))).isoformat() + "Z"
+            locations.append(os.path.join(loc_root, '{0}:{1}.list'.format(pkg, arch)))
+
+        locations.append(os.path.join(loc_root, '{0}.list'.format(pkg)))
+        for location in locations:
+            try:
+                iso_time = datetime.datetime.utcfromtimestamp(
+                            int(os.path.getmtime(location))).isoformat() + 'Z'
+                break
+            except:
+                pass
 
         if iso_time is None:
-            location = "/var/lib/dpkg/info/{0}.list".format(pkg)
-            if os.path.exists(location):
-                iso_time = datetime.datetime.utcfromtimestamp(int(os.path.getmtime(location))).isoformat() + "Z"
+            log.debug('Unable to get package installation time for package "%s".', pkg)
 
     return iso_time
 
--- a/salt/modules/dpkg.py
+++ b/salt/modules/dpkg.py
@@ -307,9 +307,9 @@ def _get_pkg_info(*packages, **kwargs):
             key, value = pkg_info_line.split(":", 1)
             if value:
                 pkg_data[key] = value
-            install_date = _get_pkg_install_time(pkg_data.get('package'))
-            if install_date:
-                pkg_data['install_date'] = install_date
+        install_date = _get_pkg_install_time(pkg_data.get('package'), pkg_data.get('architecture'))
+        if install_date:
+            pkg_data['install_date'] = install_date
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: salt/modules/dpkg.py [hunk 1]
```diff
@@ -335,17 +335,23 @@ def _get_pkg_license(pkg):
...
-            iso_time = datetime.datetime.utcfromtimestamp(int(os.path.getmtime(location))).isoformat() + "Z"
+        if arch is not None and arch != 'all':
+            location = "/var/lib/dpkg/info/{0}:{1}.list".format(pkg, arch)
+            if os.path.exists(location):
+                iso_time = datetime.datetime.utcfromtimestamp(int(os.path.getmtime(location))).isoformat() + "Z"
+
+        if iso_time is None:
+            location = "/var/lib/dpkg/info/{0}.list".format(pkg)
+            if os.path.exists(location):
+                iso_time = datetime.datetime.utcfromtimestamp(int(os.path.getmtime(location))).isoformat() + "Z"
```
isbm: There are few issues here as I see it. First, the `iso_time` will be not None only if previous location exists. Therefore I would not repeat myself twice. Second, you do not need to call OS's stat three (!) times. Third, there is no logging involved, so if package installation time is not popping up while asked, we have no idea why. Also you repeating constants, which might be also needed to be fixed.

I would write this part this way instead:

```python
loc_root = '/var/lib/dpkg/info'
locations = []
if arch is not None and arch != 'all':
    locations.append(os.path.join(loc_root, '/{0}:{1}.list'.format(pkg, arch))
locations.append(os.path.join(loc_root, '/{}.list'.format(pkg))  # lower prio, if arch
# ... add here other locations on different Debian flavours, if any
for location in locations:
    try:
        iso_time = datetime.datetime.utcfromtimestamp(
            int(os.path.getmtime(location))).isoformat() + "Z"
        break
    except FileNotFoundError as err:
        log.debug('Unable to get package installation time: %s', err)
```

Also I am not sure if ISO format is properly generated here. I'd rather use time zone, something like (only `try/except` block):
```python
try:
    iso_time = datetime.datetime.utcfromtimestamp(
        int(os.path.getmtime(location))).replace(tzinfo=datetime.timezone.utc).isoformat()
except FileNotFoundError as err:
    log.debug('Unable to get package installation time: %s', err)
```

isbm (changes requested):

rpasche:
@isbm: I'm not sure, what the main problem is? I'm not calling ```os.path.exits``` 3 times. I'm calling it 2 times, because the ```.list``` file is not always in format ```<name>:<architecture>.list``` even if the package has an architecture other than ```all```. Often enough, it is only ```<package>.list```. I don't want to run another query againts the file list of the package to get the correct name of the ```.list``` file, so I just test, if the architecture is appended to the package name and if not, fall back to old style.

And in case, the file with architecture is not present, ```iso_time``` is not touched at all, so still ```None```

Also sorry for setting the resolved flag. Had some problems with the browser

salt/modules/dpkg.py
isbm: .

isbm:
@rpasche `os.path.getmtime` is also stat. Third one. Please take another look how I solved your case. You will run queries until you get iso_time. Sometimes it will be once, sometimes twice. In your case you do it _always_ twice because you [repeat yourself in the code, which is a bad practice](https://deviq.com/don-t-repeat-yourself/). And then third time you stat it again. We probably should also make sure that the ISO time stamp is right. "Z" append doesn't seem right after all for UTC time stamps, but we need to double-verify that.

The conversation was not resolved yet.

salt/modules/dpkg.py
rpasche: Even without testing your suggestion yet, for all packages with architecture ```!= 'all'```...in worst case, you log an error message, although you are able to retrieve the installation date correctly (from the other filename).

But I'll try to correct this and test again....

isbm (approved):
@rpasche Good to go! Thanks!

Issue #45626, opened 2018-01-23
Logs going to both file and syslog


hgfischer:
### Description of Issue/Question

SaltStack is logging to both `/var/log/salt/*` and `/var/log/syslog`. I haven't identified the reason, but I think it is because salt's daemons are still writing to `STDOUT` even when `log_file: /var/log/salt/...` is used.

### Setup

Using latest SaltMinion on Ubuntu 16.04.3 (systemd). With `log_file: /var/log/salt/minion`.

### Steps to Reproduce Issue

Run salt-minion and compare syslog logs with `/var/log/salt/minion`.

### Versions Report
```
Salt Version:
           Salt: 2017.7.2
 
Dependency Versions:
           cffi: 1.5.2
       cherrypy: Not Installed
       dateutil: 2.4.2
      docker-py: Not Installed
          gitdb: 0.6.4
      gitpython: 1.0.1
          ioflo: Not Installed
         Jinja2: 2.8
        libgit2: 0.24.0
        libnacl: Not Installed
       M2Crypto: 0.21.1
           Mako: 1.0.3
   msgpack-pure: Not Installed
 msgpack-python: 0.4.6
   mysql-python: Not Installed
      pycparser: 2.14
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: 0.24.0
         Python: 2.7.12 (default, Nov 20 2017, 18:23:56)
   python-gnupg: Not Installed
         PyYAML: 3.11
          PyZMQ: 15.2.0
           RAET: Not Installed
          smmap: 0.9.0
        timelib: Not Installed
        Tornado: 4.2.1
            ZMQ: 4.1.4
 
System Versions:
           dist: Ubuntu 16.04 xenial
         locale: UTF-8
        machine: x86_64
        release: 4.4.0-104-generic
         system: Linux
        version: Ubuntu 16.04 xenial
```

garethgreenaway:
@hgfischer Thanks for the report.  Have you set the `log_level` setting in the minion configuration?

hgfischer:
@garethgreenaway 

I am using these:
```
log_level: debug                                                                                                        
log_level_logfile: debug                                                                                                
```

PR #50326, opened 2018-10-31
Fix double logging to file and syslog.

aabouzaid:
### What does this PR do?
In Ubuntu/Debian with systemd (e.g. `Ubuntu 16.04`), salt minion and master were logging to custom file (`/var/log/salt/*`) and syslog (`/var/log/syslog`) at the same time.

So this PR disable send logs to `syslog` and keep the file managed in Salt config.

This a fix for issue no. #45626

### Previous Behavior
Salt minion and master were logging to custom file and syslog at the same time.
```
/var/log/syslog
/var/log/salt/master
/var/log/salt/minion
```

### New Behavior
Salt minion and master logging only to configured log file.
```
/var/log/salt/master
/var/log/salt/minion
```

### Tests written?
N/A

### Commits signed with GPG?
Yes

### Info
```
salt-minion -V
Salt Version:
           Salt: 2017.7.6
 
Dependency Versions:
           cffi: 1.5.2
       cherrypy: Not Installed
       dateutil: 2.4.2
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.8
        libgit2: 0.24.0
        libnacl: Not Installed
       M2Crypto: 0.21.1
           Mako: 1.0.3
   msgpack-pure: Not Installed
 msgpack-python: 0.4.6
   mysql-python: Not Installed
      pycparser: 2.14
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: 0.24.0
         Python: 2.7.12 (default, Nov 20 2017, 18:23:56)
   python-gnupg: Not Installed
         PyYAML: 3.11
          PyZMQ: 15.2.0
           RAET: Not Installed
          smmap: Not Installed
        timelib: Not Installed
        Tornado: 4.2.1
            ZMQ: 4.1.4
 
System Versions:
           dist: Ubuntu 16.04 xenial
         locale: UTF-8
        machine: x86_64
        release: 4.4.0-62-generic
         system: Linux
        version: Ubuntu 16.04 xenial
```

### DIFF
--- a/pkg/deb/salt-master.service
+++ b/pkg/deb/salt-master.service
@@ -6,6 +6,8 @@ After=network.target
 LimitNOFILE=16384
 Type=notify
 NotifyAccess=all
+StandardOutput=null                                                                                                     
+StandardError=null 
 ExecStart=/usr/bin/salt-master
 
 [Install]
--- a/pkg/deb/salt-minion.service
+++ b/pkg/deb/salt-minion.service
@@ -7,6 +7,8 @@ Type=notify
 KillMode=process
 NotifyAccess=all
 LimitNOFILE=8192
+StandardOutput=null                                                                                                     
+StandardError=null 
 ExecStart=/usr/bin/salt-minion
 
 [Install]


cachedout:
I'm not sure I follow the logic here. What if a person _wants_ to have it logged to syslog. Doesn't this prevent that?

AAbouZaid:
@cachedout The user can log to any place using [log_file](https://docs.saltstack.com/en/latest/ref/configuration/logging/#log-file) in master or minion.

The current behavior is that `systemd` also logs to `syslog` ... so the same data are logged twice on the same machine.

cachedout:
> @cachedout The user can log to any place using [log_file](https://docs.saltstack.com/en/latest/ref/configuration/logging/#log-file) in master or minion.
> 
> The current behavior is that `systemd` also logs to `syslog` ... so the same data are logged twice on the same machine.

OK. That makes sense. However, I do think we should heavily document this change in the release notes.

AAbouZaid:
@cachedout I think that too :-)
If there is anything I can do from my side (write that somewhere else to it will appear in the release notes) I will be glad to do :-)

rallytime:
@AAbouZaid If you could add some documentation to [this file](https://github.com/saltstack/salt/blob/develop/doc/topics/releases/neon.rst) that would be great. :)

gtmanfred (commented):
I disagree with this.  you lose all benefits of logging to journald, like seeing exactly when a log message got written to journald, what user did it, waht process and pid and during which boot cycle it was written.

If you do not want them also written to /var/log/syslog, you should stop the rsyslog service or whatever default service in ubuntu and redhat also writes disks out to files.  And set the log_file to /dev/null for salt-minion.

Also, this can be changed on a per userbases by adding a file at `/etc/systemd/system/salt-minion.service.d/logging.conf` that just has the following in it.

```
[Service]
StandardOutput=null                                                                                                     
StandardError=null 
```

I think that leaving it as is, is a sane default and anyone that doesn't care about journald can turn it off.

AAbouZaid:
@gtmanfred
You totally have a point, but there is no solution to disable the double logging by default? :thinking:
The same data are logged twice every time (syslog has more few data, but 99% are logged 2 times).

gtmanfred:
so, it is probably actually logged three times, once in journald, once in /var/log/syslog or messages depending on the OS, and once in /var/log/salt/*... we do not control turning off /var/log/syslog, that is the way the os is setup, to dump to disk by default, it is dumb, but it is what most distros chose to do instead of forcing people to use journalctl to query the logs.  Now, we could set the log file to not log to /var/log/salt/<service> but enough people look there that i don't think it would be good to just yank it out from under them, it would need a deprecation period.

And i do not support removing the logs from journald, because like i said, actually being able to introspect the logs, especially on a system that controls so much, is super valuable, and i don't like the idea of disabling it by default.

cachedout:
I don't think there is enough support among core developers to get this merged right now. I appreciate you sending this over, @AAbouZaid but I think we're going to close it.

PR #50332, opened 2018-10-31
nilrt_ip: _get_service_info(): Touch "ipv6" item in `data` map before use

harisokanovic:
Touch `data['ipv6']` before referencing it to avoid exception.

### New Behavior
I verified `salt-call --local ip.set_dhcp_linklocal_all ens3` now works on a QEMU x64 machine with usermode IPv6 networking.


cachedout (approved):

cachedout:
Thanks, @harisokanovic

PR #50337, opened 2018-11-01
Make x509 remote signing highly availability

aarnaud:
### What does this PR do?
Make x509 remote signing highly availability

### Previous Behavior
Only one CA server can be used

### New Behavior
Many CA server can be used to have highly availability

### Tests written?

No but use in production in `_modules`

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/modules/x509.py
+++ b/salt/modules/x509.py
@@ -1129,7 +1129,7 @@ def create_certificate(
     ca_server:
         Request a remotely signed certificate from ca_server. For this to
         work, a ``signing_policy`` must be specified, and that same policy
-        must be configured on the ca_server. See ``signing_policy`` for
+        must be configured on the ca_server (name or list of ca server). See ``signing_policy`` for
         details. Also the salt master must permit peers to call the
         ``sign_remote_certificate`` function.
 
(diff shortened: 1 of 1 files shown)


salt/modules/x509.py
```diff
@@ -1388,19 +1388,29 @@ def create_certificate(
         for ignore in list(_STATE_INTERNAL_KEYWORDS) + ['listen_in', 'preqrequired', '__prerequired__']:
             kwargs.pop(ignore, None)
 
-        certs = __salt__['publish.publish'](
-            tgt=ca_server,
-            fun='x509.sign_remote_certificate',
-            arg=six.text_type(kwargs))
+        ca_servers = []
+        if type(ca_server) is list:
```
gtmanfred: This needs to be `if isinstance(ca_server, list):`

salt/modules/x509.py
```diff
@@ -1388,19 +1388,29 @@ def create_certificate(
...
+                tgt=server,
+                fun='x509.sign_remote_certificate',
+                arg=six.text_type(kwargs))
+            if certs is None or not any(certs):
+              continue
+            else:
+              cert_txt = certs[server]
+              break
 
         if not any(certs):
```
gtmanfred: Does this need to be adjusted?

salt/modules/x509.py
```diff
@@ -1388,19 +1388,29 @@ def create_certificate(
...
-            tgt=ca_server,
-            fun='x509.sign_remote_certificate',
-            arg=six.text_type(kwargs))
+        ca_servers = []
+        if type(ca_server) is list:
+            ca_servers = ca_server
+            random.shuffle(ca_servers)
+        else:
+            ca_servers.append(ca_server)
+        for server in ca_servers:
```
gtmanfred: For this, can we instead do

```
if not isinstance(ca_server, list):
    ca_server = [ca_server]
random.shuffle(ca_server)
for server in ca_server:
```

Just to simplify it a bit, and remove extra assignments.

gtmanfred (changes requested):

salt/modules/x509.py
aarnaud: I change with your proposition:
```
if not isinstance(ca_server, list):
    ca_server = [ca_server]
random.shuffle(ca_server)
for server in ca_server:
```

salt/modules/x509.py
aarnaud: Applied

gtmanfred (approved):

PR #50341, opened 2018-11-01
ssdp discovery: multimaster support

mattp-:
### What does this PR do?
a user can now specify an accept: all (default 'any' for original behavior) to connect to all masters that match the ssdp config specified. this meant moving ssdp discovery from eval_master to tune_in phase of the MinionManager, as such discovery only happens once per process start, so reconnections don't collide (ie, if eval_master call remained, the list order is not guaranteed).

@isbm, I know you were the main author of this functionality. Was your original usecase relying on dynamic rehoming if a disconnection occurs? or is this behavior change ok. We could still support a discovery in eval_master, but only if 'accept: any' (ie single master) is configured.

n/a

### Previous Behavior
ssdp discovery happened at eval_master time

### New Behavior
ssdp discovery happened at minionmanager time, to allow for multimaster

### Tests written?
No

### Commits signed with GPG?
No


gtmanfred (approved):
This looks good, i would really like to see all these options documented, but as it isn't already, we should just make a documentation issue for that and do it in another pr.

isbm:
@mattp- This is interesting. But I would recommend join slack channel in saltstack community and let's discuss this a bit. This is basic feature that I wanted soon to extend. Currently it joins "first win" packet. But this should be extended to:

- Join to less loaded.
- Re-join elsewhere if current master goes offline.

Also this feature has one major security flaw: you can put to the circle malicious master and gain full control on the machine with the minion. Therefore there I would propose a mechanism of "trusted circle" of masters. If you put to the circle untrusted master, all trusted masters marks that one as "untrusted" and minions avoiding it automatically.

cachedout:
@isbm I'm not sure how I understand how multi-master support is functionally different than single-master support from a security perspective. If SSDP is advertising a malicious master, it seems like there is already a serious breach, no?

isbm:
@cachedout no, there is no difference.

cachedout:
@mattp- Can you please grab the lint errors described here? https://jenkinsci.saltstack.com/job/pr-lint/job/PR-50341/12/warnings52Result/new/

mattp-:
@cachedout woops, fixed.

isbm (approved):

cachedout:
So, given that we agree that there is no security threat _introduced_ by this PR, I think it is fine to go in. That said, certainly the ideas from @isbm are worthy of consideration and it would be good to see them go through the RFC process. Everybody cool with that?

mattp-:
sounds good to me :+1:

isbm:
This PR is not introducing problems, on top of existing ones. :grinning:

The security mechanism should be through the RFC without questions.

Issue #36502, opened 2016-09-22
Salt Text Handling of different Line Endings


damon-atkins:
### Description of Issue/Question

I will raise this as a separate issue as its bigger than this PR.
There are multiple competing Line Ending Formats:
- LF+CR
- LF
- CR
- CR+LF

It's a bad assumptions on Python part that only one type will be on the platform.
Application A on windows may use LF+CR and application B may use LF.   Or on Unix you may use salt to generate a file on Unix, which is provide to windows clients using Samba.
Text handling needs to hand all cases on a platform.   

Here is an example
C:\salt>salt-call --version
salt-call 2016.3.3 (Boron)

![image](https://cloud.githubusercontent.com/assets/9368124/18754668/d4c17ef0-812c-11e6-82a2-3be6188c15a4.png)

I think this will take a brain trust to decide on a way forward.

lorengordon:
Related issues:
* https://github.com/saltstack/salt/issues/12284 (should be closed though)
* https://github.com/saltstack/salt/issues/14002
* https://github.com/saltstack/salt/issues/14135
* https://github.com/saltstack/salt/issues/27368

Ch3LL:
I want to get a couple more people on this issue to determine how we could move forward these issues because obviously as @lorengordon points out there are multiple issues were this has caused issues. 

ping @twangboy @terminalmage @cachedout just wanted to get any of your inputs on whether you think there is an easy way for us to handle these situations within salt which we could track here. thanks

damon-atkins:
I see the option as 
- Editing in place, read the first line (or x bytes) of a text file to determine the line feed before performing any action. eg. regex (.[\r\n].) If file does not exist use **kargs newline='\r\n' or if newline=None use the platforms default newline  
- Use the same line feed as the original file which may come from http:// salt:// etc.

damon-atkins:
From def prepend(path, *args, **kwargs):
```
    try:
        with salt.utils.fopen(path) as fhr:
            contents = fhr.readlines()
    except IOError:
        contents = []
    preface = []
    for line in args:
        preface.append('{0}\n'.format(line))
    with salt.utils.fopen(path, "w") as ofile:
        contents = preface + contents
        ofile.write(''.join(contents))
    return 'Prepended {0} lines to "{1}"'.format(len(args), path)
```
I am adding this here as I suspect this is a general comment as well on handling text files.
[...]

PR #50349, opened 2018-11-02
Fixes `integration.states.test_file` on Windows in Py3

twangboy:
### What does this PR do?
Fixes `integration.states.test_file.FileTest.test_managed_unicode_jinja_with_tojson_filter` on Windows in Python 3. The `to_unicode` function wasn't handling the utf-8 conversion for some reason. This adds the ability to use the encoding option with open that's new to Py3.

Found in Jenkins tests

### Tests written?
Yes

### Commits signed with GPG?
Yes


dwoz (approved):

cachedout (approved):
This looks correct to me.

damon-atkins:
Python 2 io.open is the same as python 3 open & io.open
https://docs.python.org/2/library/io.html
See #36502 Salt Text Handling of different Line Endings

salt/utils/files.py
```diff
@@ -349,6 +349,10 @@ def fopen(*args, **kwargs):
                 )
         except IndexError:
             pass
+    if six.PY2:
+        # encoding is not available in Py2 and will cause an error later if passed
+        if kwargs.pop('encoding', False):
+            log.debug('Remove kwarg "encoding" on Py2')
```
terminalmage: There's really no context to this log message, so I feel like it would clutter up the debug logging. We should either improve it or remove it.

We should also look into using `io.open()` instead of `open()`, because that is the same function on both PY2 and PY3.

terminalmage (changes requested):

salt/utils/files.py
twangboy: Removed log message

twangboy:
@damon-atkins @terminalmage Now uses io.open by default.

terminalmage (approved):

twangboy:
@rallytime Rebased...

salt/loader.py
```diff
@@ -1535,7 +1535,7 @@ def _load_module(self, name):
                         # pylint: enable=no-member
                         sys.modules[mod_namespace] = mod
                     else:
-                        with salt.utils.files.fopen(fpath, desc[1]) as fn_:
+                        with salt.utils.files.fopen(fpath, desc[1], use_io_open=False) as fn_:
```
cachedout: Eeek. Why are we modifying the loader in an attempt to fix a failing state test?

twangboy: The loader needs to open using the old method (not io.open).

cachedout: OK, but why? This is a change to the core of how Salt operates. We need to understand exactly what's happening here.

twangboy: @terminalmage can you answer this?

twangboy: Nevermind... this has already been fixed

twangboy:
Closed in favor of https://github.com/saltstack/salt/pull/50713

Issue #40135, opened 2017-03-18
Consistent Params for Pkg Management


damon-atkins:
### Description of Issue/Question
Hi Salt team,
    What is the aim of the modules API/Interface. This is an example of Pkg Management Modules

Is its salt aim for it to be the same across all platforms?
Is its salt aim for params to be the same order?
It its salt aim that the params can be ignored if not valid for the platform?
Is its salt aim to use kwargs or not to use kwargs given changes in PR #40029 which makes win_pkg different to other pkgs modules which uses a lot of kwargs?

Cheers

### def install ### 12/13
aptpkg.py:def install(name=None, refresh=False, fromrepo=None, skip_verify=False, debconf=None, pkgs=None, sources=None, reinstall=False, ****kwargs**)
freebsdpkg.py:def install(name=None, refresh=False, fromrepo=None, pkgs=None, sources=None, ****kwargs**)
mac_pkgutil.py:def install(source, package_id)
openbsdpkg.py:def install(name=None, pkgs=None, sources=None, ****kwargs**)
opkg.py:def install(name=None, refresh=False, pkgs=None, sources=None, ****kwargs**)
pkgin.py:def install(name=None, refresh=False, fromrepo=None, pkgs=None, sources=None, ****kwargs**)
pkgng.py:def install(name=None, fromrepo=None, pkgs=None, sources=None, jail=None, chroot=None, root=None, orphan=False, force=False, glob=False, local=False,
pkgutil.py:def install(name=None, refresh=False, version=None, pkgs=None, ****kwargs**)
solarispkg.py:def install(name=None, sources=None, saltenv='base', ****kwargs**)
win_pkg.py:def install(name=None, refresh=False, pkgs=None, ****kwargs**)
xbpspkg.py:def install(name=None, refresh=False, fromrepo=None, pkgs=None, sources=None, ****kwargs**)
yumpkg.py:def install(name=None, refresh=False, skip_verify=False, pkgs=None, sources=None, reinstall=False, normalize=True, update_holds=False, ****kwargs**)

### def is_installed ### 1/13
mac_pkgutil.py:def is_installed(package_id)

### def latest_version ### 11/13
aptpkg.py:def latest_version(*names, **kwargs)
freebsdpkg.py:def latest_version(*names, **kwargs)
openbsdpkg.py:def latest_version(*names, **kwargs)
opkg.py:def latest_version(*names, **kwargs)
pkgin.py:def latest_version(*names, **kwargs)
[...]

PR #50353, opened 2018-11-02
Virtual module interface concept

isbm:
### What does this PR do?

Introduces virtual module interface concept (RFC).
[Read it here](https://github.com/isbm/salt/blob/isbm-api-unification-rfc/rfcs/0003-api-unification.md)

@terminalmage @cachedout @thatch45 @dwoz

### DIFF
--- a/rfcs/0003-api-unification.md
+++ b/rfcs/0003-api-unification.md
@@ -176,10 +176,12 @@ if __grains__['osfamily'] == 'RedHat':
 ## Return Structure Definition
 [returnstruct]: #returnstruct
 
-Return structure in virtual modules another pitfall, where dynamically
-replaced module suddenly returns "something else" than usually
-expected. This is widely affects API and integration. In this case
-integration code usually looks like this (pseudo-code):
+Return structure in virtual modules is another pitfall. Dynamically
+replaced module suddenly renders virtual module to return "something else"
+than is usually expected. This is widely affects API and integration.
+To the only way to avoid this, is to know what kind of platform minion
+is dealing with. In this case integration code usually looks like
+this (pseudo-code):
 
 ```
 if this_is_debian {
--- a/rfcs/0003-api-unification.md
+++ b/rfcs/0003-api-unification.md
@@ -6,151 +6,268 @@
 # Summary
 [summary]: #summary
 
-Modules with `__virtual__` should yield to the common interface.
+Salt Module Interface (SMI) concept introduction for virtual modules.
 
 # Motivation
 [motivation]: #motivation
 
-Virtual modules behaves differently on different Linux distributions
-and different operating systems. So if one is query `pkg` module, on
-one system there is `lock` package, on other is `hold` package, on
-another this feature is not even there. And so on...
+Any salt module has couple of specific properties we dealing every day
+with:
 
-To solve this chaos, virtual modules, that represents a module per se,
-should also have the same standard interface. So we no longer have a
-module that is _called_ differently on different operating systems,
-however we will have a module that _reports_ differently on
-heterogeneous environments.
+- Fixed set of functions
+- Known functions signatures
+- Known structure of return
+
+This is true but for virtual modules. The virtual module covering
+several "fixed" or "physical" modules and behaves like that would be
+one module. But differences between those physical modules on
+different platforms makes such virtual module a moving target and
+unpredictable.
+
+Virtual modules concept is missing crucial part in the design:
+interfaces. The interface should define how module looks like and what
+APIs can be called to it. Interface should move module that is
+_called_ differently on heterogeneous environments to a module that
+_reports_ differently on heterogeneous environments.
 
 
 # Design
 [design]: #detailed-design
 
-Interface is a list of functions in the module and is to make sure
-that any virtual module is always called exactly the same, regardless
+*DISCLAIMER*: The SMI is not that classic understanding of typical
+interface one may find in languages like Java. It is also not as same
+as Zope Interface package or Python Abstract Base Classes (ABC).
+
+The SMI should describe the following properties of the module:
+
+- Functions
+- Signatures
+- Lowest common denominator of the function output format (or minimum
+  required default output structure)
+
+SMI only describes functions of the module and is there to make sure
+that any virtual module is always called exactly the same way, regardless
 what operating system minion is running on.
 
-Interface does not define the function output format. It only defines
-the name of the function and its signature.
-
-
 ## Declaration
+[declaration]: #declaration
 
-Salt interfaces are declared just as classes. Salt's module to
-functions is Salt's interface to methods. Therefore `self` in the
-class is not a part of a function signature. Example:
+SMI are declared just as regular Python classes. Salt's "module to
+functions" map is Salt's "interface to methods" map. Therefore `self`
+parameter in the SMI class is not a part of a function signature.
+
+Example of SMI definition for module `pkg`:
 
 ```
 from salt.interfaces import Interface
 
 class PkgInterface(Interface):
+	__modulename__ = 'pkg'
+
     def list_installed(self, *names, **kwargs):
         '''
 	    List installed packages.
 	    '''
+		return {}
 
     def upgrade_available(self, name, **kwargs):
         '''
         List available upgrades.
 		'''
+		return {}
+
+    @Interface.supported('weirdlinux', 'beos', 'frogbsd')
+    def salute_fireworks(self, name):
+        '''
+        Launch some fireforks
+        '''
+		return {}
 ```
 
-Basically, the list of methods should reflect exact names and
-signatures as in the module, except `self` parameter.
+In above incomplete interface example, the list of methods should
+reflect exact names and signatures as in the module, except `self`
+parameter. Rules apply:
+
+- If a method is not in the SMI class, but function is implemented in
+  the module, then such function is marked as "deprecated".
+
+- If a method is in the SMI class but not in the module, then such
+  function is marked as "not implemented".
+
+- If a method has a decorator `@Interface.supported`, only on these
+  platforms unimplemented method will be reported as "not
+  implemented", otherwise "not supported".
 
 
 ## Usage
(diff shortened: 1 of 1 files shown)


isbm:
@sjorge what you think?

sjorge:
Maybe a decorator to mark a function so that they are only available  when called via 'somepkgmgr.fancy_stuff' and hidden when called via the 'pkg' virtual implementation?

I can see situations where the virtual interface is only providing a subset of shares functions but where it might be nice to have more implementation specific functions also available by calling it directly. It's a bit hard to express so let me try a mock example

VehicleInterface:
- move_forwards
- move_backwards
- turn_left
- turn_right

Both the 'targets' of 'boat', 'car', and 'plane' can provide the VehicleInterface.

But it might be desirable to also have 'plane.drop_landing_gear', 'plane.raise_landing_gear', 'plane.takeoff', and 'plane.land' accessible.

Generally via the VehicleInterface we can access the basic movements that will work but in some case if say 'plane' needs a bit of a kick to get going (.e.g plane.takeoff, plane.retract_landing_gear) those can still be called via 'plane' instead of the 'vehicle' interface. Does that make sense?

So..

Plane (implements VehicleInterface)
- move_forwards
- move_backwards
- turn_left
- turn_right
- takeoff (@non_virtual_interface)
- land  (@non_virtual_interface)
- drop_landing_gear  (@non_virtual_interface)
- retract_landing_gear  (@non_virtual_interface)

Where **non_virtual_interface** is the decorate to not expose the function via the VehicleIntercace (It needs a better decorator name for sure though)

I guess this can also be implemented as

Plane:
- move_forwards
- move_backwards
- turn_left
- turn_right
- takeoff
- land
- drop_landing_gear
- retract_landing_gear

PlaneVehicle (implements VehicleInterface):
- move_forwards
- move_backwards
- turn_left
- turn_right

Both would work, but the 2nd one would create more overhead.

isbm:
> Maybe a decorator to mark a function so that they are only available when called via 'somepkgmgr.fancy_stuff' and hidden when called via the 'pkg' virtual implementation?

Ha! I am actually already doing something similar. Soon comes another PR for `salt-support` available via modules/SaltSSH. I use class methods instead of functions, and methods are exported as functions exactly via decorator. We can definitely reuse that!

isbm:
@sjorge OK, so what you're basically saying can be outlined as:

> Virtual modules are only for the commonly available functions.

So as long as function is e.g. Solaris-only or Windows-only, it is not in the interface and is fancy.
I personally think exactly the same. Except maybe we do not need to kill the function, but only make it available when directly accessing the same module (not via virtual name).

That is something like `pkg.install` but `ips.get_svr4`, where `pkg.get_svr4` simply won't be found. I would also propose to forbid accessing `ips.install` and nag-screen it with "Function is not available, please call pkg.install instead" — because if we allow `ips.install` too, then you will have a lot of non-portable code around, which is bound only to `ips` instead of to `pkg`.

As of implements/inherits part, I think this would not be necessary. Interface is more like "gate" or "filter" here, rather then full-blown programming approach. All what we need is to dissect from what is common and what is proprietary. The very interface definition would be that "decorator" by itself automatically "for free".

But overall, this is a _very good point_, as it would even simplify the whole migration: just warn "You are accessing Solaris-only function, which will disappear from pkg function in N releases/years/days/minutes. Further use of this function via ips package".

P.S. I also want it minimum-to-none intrusive to other modules. :wink:

damon-atkins:
This ended up being a brain dump.

See also https://github.com/saltstack/salt/issues/40135  `Consistent Params for Pkg Management`

I have been pushing  a minimum interface standard. So for example you can read about pkg.installed state and it will work on all platforms, and then anything extra you need to read the win.pkg.installed.
Its faster to learn something consistent input and output.

The other thought is being able to pass data from one state to another.  Something like a standard dict called output.  For example a state which fetches a file into a cache, and returns the file location in the cache, which then can be used by the next state to do something with the cache file. At the moment you would have to hard code this path.

In an updated version of modules/win_pkg most of the backed code is now classes. A class for the old method and class for the new method, when pkg.install is called it determine which class too use, and hence the set of functions that back-ends the process of installing software. One interesting outcome of using classes in modules is less python imports.  (states. states/pkg.py has if statements for windows to determine what to import)

`__virtualname__ = PkgInterface(__name__)()`

The current issue with `__virtualname__` is there is a check to see if this is set,  I was thinking about raising a PR so you can set this to dynamic, so the error is not printed anymore.  The reason being if you want to have more than one package manger. e.g. Solaris IPS and PkgIn on the same platform, where IPS is considered `__salt__[pkg.install]` and PkgIn can exist as `__salt__[pkgin.install]`

I also like to see a standard where platform specific information can be passed
e.g.
```
file.perms
   linux:
   win:
   <os_family>:
```
So you do not need todo `if windows` do something different.

States should also take a 'provider' input, e.g. indicate what pkg manager to use, or maybe even a version. e.g. so we have a better method than dockerng (next gen) to replace docker.

Why does the virtual function need to be in the same file as the module?
Would it not be faster to have
salt/modules/something.py  # only loaded/defined if salt/modules_virtual/something.py does not return false.
salt/modules_virtual/something.py  
Or
salt/modules_virtual/pkg.py  # this has the smarts in it work out what is valid.

Maybe even
`__salt__[pkg.install]` and
`__salt_provider[<provider_name>][pkg.install]`

rfcs/0003-api-unification.md
```diff
@@ -0,0 +1,277 @@
...
+_called_ differently on heterogeneous environments to a module that
+_reports_ differently on heterogeneous environments.
+
+
+# Design
+[design]: #detailed-design
+
+*DISCLAIMER*: The SMI is not that classic understanding of typical
+interface one may find in languages like Java. It is also not as same
+as Zope Interface package or Python Abstract Base Classes (ABC).
```
cachedout: I think we should further explore why we _shouldn't_ use Python ABCs. Part of my concern here is how application-specific this proposal is. I think we would be best served by being as "Pythonic" as possible.

rfcs/0003-api-unification.md
```diff
@@ -0,0 +1,277 @@
...
+- Known functions signatures
+- Known structure of return
+
+This is true but for virtual modules. The virtual module covering
+several "fixed" or "physical" modules and behaves like that would be
+one module. But differences between those physical modules on
+different platforms makes such virtual module a moving target and
+unpredictable.
+
+Virtual modules concept is missing crucial part in the design:
```
cachedout: Is your proposal to just restrict the usage of this to virtual modules or do you see a wider usage being adopted?

rfcs/0003-api-unification.md
isbm: @cachedout The interface concept I am proposing is to dissect what should be for virtual modules (commonly available functions) from what is not everywhere possible.

We cannot use ABCs because first of all they stand for "abstract base **classes**". We do not have this by design, since all modules in Salt are functions at the module level. The ABC mechanism is pretty simple: as long as there is abstract method which is not overridden by an actual implementation, the `@abstractmethod` decorator will just trigger TypeError exception and block _abstract class_ from being instantiated with abstract methods in it.

But even if we would use classes, we still don't want that behaviour as well. We want to be very transparent and make sure that outside of the interface methods are perfectly usable, just they are not visible through the virtual access. That is that `mymodule.fancy_function` is accessible but `myvirtmodule.fancy_function` is not there, although this is call of the same module, just first is physical, the other is virtual. Because if someone needs something that is only possible on OSX or Solaris, we should not kill the function or force everyone to implement "Unsupported!" stub. But yes, some modules might be unsupported, in that case automatically added stub actually makes sense.

rfcs/0003-api-unification.md
isbm: I would not be very intrusive and have it for now only for virtual modules. But we should be able to always extend this.

cachedout (approved):

rfcs/0003-api-unification.md
```diff
@@ -0,0 +1,277 @@
...
+functions" map is Salt's "interface to methods" map. Therefore `self`
+parameter in the SMI class is not a part of a function signature.
+
+Example of SMI definition for module `pkg`:
+
+```python
+from salt.interfaces import Interface
+
+class PkgInterface(Interface):
+	__modulename__ = 'pkg'
```
terminalmage: Indentation level here should be fixed.

rfcs/0003-api-unification.md
```diff
@@ -0,0 +1,277 @@
...
+from salt.interfaces.pkg_module import PkgInterface
+
+__virtualname__ = PkgInterface(__name__)()
+
+```
+
+The code above does the following:
+
+- Ensures that the `__virtualname__` is properly set according to the
+  interface.
```
terminalmage: How? It looks like we're setting the `__virtualname__` to a class instance. Is the idea to have the loader look for `module.__virtualname__.__modulename__`, and fall back to `module.__virtualname__` if an `AttributeError` is raised?

isbm: > How? It looks like we're setting the `__virtualname__` to a class instance. 

Ah, no-no. Not the instance. This is doing the following:

```python
__virtualname__ = PkgInterface(__name__)()
```

The `__call__` of the PkgInterface will return just a string. The whole thing here is to:

- `PkgInstance` has already whatever name is inside, so `__call__` will just set it.
- But this call function will also introspect the current module and will readjust everything.

Think of it as just as this:

```python
auto_check_this_module(__name__)
__virtualname__ = 'humasn_writse_tpyos'
```

But I propose to use smart Class instance that also can "disappear" after its init, since it is no longer needed anyway. All is needed is to be a class as an interface carrier, get the module instance to itself, "pipe through" its checks, inject where needed stubs and return a proper name for the virtual module.

In fact, it can be even more simpler:

```python
PkgInterface(__name__)
```

Then it will take current module and set `__virtualname__` by itself. But this is maybe too much magic...

isbm:
@cachedout bump!

terminalmage (approved):

PR #50360, opened 2018-11-02
INFRA-7721 - boto CloudFront support of various kinds

tkwilliams:
### New functionality
- new boto_cloudfront exec functions
-- get_distribution_v2()
-- get_distribution_config()
-- list_distributions()
-- distribution_exists()
-- get_distributions_by_comment()
-- create_distribution_v2()
-- update_distribution_v2()
-- disable_distribution()
-- delete_distribution()
-- list_cloud_front_origin_access_identities()
-- get_cloud_front_origin_access_identity()
-- get_cloud_front_origin_access_identity_config()
-- get_cloud_front_origin_access_identities_by_comment()
-- create_cloud_front_origin_access_identity()
-- update_cloud_front_origin_access_identity()
-- delete_cloud_front_origin_access_identity()
-- cloud_front_origin_access_identity_exists()
-- list_tags_for_resource()
-- tag_resource()
-- untag_resource()
-- enforce_tags()
- new boto_cloudfront state functions
-- distribution_present()
-- route53_alias_present()
-- distribution_absent()
-- origin_access_identity_present()
-- origin_access_identity_absent()
- new boto_iam exec functions
-- list_server_certificates()
- small bugfix to boto_route53.rr_present() found while developing the above

### Tests written?

No

### Commits signed with GPG?

No


dwoz (changes requested):
@tkwilliams, Looks like there are some test failures related to missing or incomplete documentation for this. It would be nice if we had some unit tests for these additions too.

tkwilliams:
Sorry, I've been down with a nasty cold the past several days.  I definitely want to get this cleaned up - give me a few days and I should have something closer to the mark :)

tkwilliams:
@dwoz  OK, I've fixed up the doc strings all around :)

I won't have any time in the near-term to work on unit tests - super sorry, but my boss has a bunch of stuff in my queue, and getting this changeset ready for PR was already putting me past the delivery date for the underlying project.  HE would have been happy with working code sans any docu at all :)  You know how that goes - if it lets our state runs succeed, merge it into local tree, build our RPMs, and open a backlog to get it "ready for upstream PR" at some far distant date....

tkwilliams:
PS, no clue whatsoever why the windows build (?!?) is failing.  And well, let's be honest, CodeClimate is a lost cause on the best of days...

cachedout:
@dwoz Can you please re-review the changes here?

salt/modules/boto_cloudfront.py
```diff
@@ -53,10 +53,12 @@
 # Import Python libs
 from __future__ import absolute_import, print_function, unicode_literals
 import logging
+import jmespath
```
cachedout: Eh? Should we be conditionally importing this? AFAIK, this won't be present in most installations.

salt/modules/boto_cloudfront.py
```diff
@@ -456,3 +460,1230 @@ def update_distribution(
         )
 
     return {'result': True}
+
+
+def get_distribution_v2(region=None, key=None, keyid=None, profile=None, **kwargs):
```
cachedout: What sort of items should a person be passing here via kwargs? Can they be documented? If not, can this be removed entirely?

salt/modules/boto_cloudfront.py
tkwilliams: Hrm.  It seems to be there on most modern-ish python installs but I see your point.  Ideally we'd add it to the salt.ext bundle, since it's just crazy useful in general, but until that happens I can pretty easily translate the several jmespath calls I make to `xxx.get('foo', {}).get('bar', '')` type calls instead :)  That's what they originally were in my code before I transmuted the to jmespath calls anyway...

tkwilliams: @cachedout  Considering that ALL the AWS API talks is JSON back and forth, having a lightweight but super convenient library on hand to easily dig around inside that JSON payload (rather than the clumsy .get().get().get() style used so often, or even worse, unsafe assumptions like foo['bar'][0]['baz']) would be a HUGE win for everyone writing boto based modules, and provide a standard approach to such things as well...

How would I go about asking for jmespath to be added to the salt.ext module exactly?  :)

tkwilliams:
@cachedout Hey Mike, any thoughts on my question above?  I think it would be of general use to have a json data mining lib such as jmespath available in salt.ext for these sorts of things.
Or @terminalmage if you're a better person to bug for those sorts of questions, I'd understand that too... :)

salt/modules/boto_cloudfront.py
cachedout: We'd have to take our case to @thatch45, @twangboy  and @dmurphy18 and see what they think. The biggest issue here is packaging, which is a pretty major pain. We'd much rather see this as an optional dependency but let's see what the feedback looks like.

tkwilliams: I can definitely make it optional, but that just ends up with two code-paths inside each function - one using JMESPath, and one using raw python (generally something on the order of `xxx.get('foo', {}).get('bar', {})[0].get('baz')` or such crappiness.  If we're going to have to put the latter into each function anyway, there's zero advantage to having the alternate JMESPath version to begin with :)

Fingers crossed ...

dmurphy18: @cachedout It is generally available, e.g. Redhat extras repo, so so long as expect extras to be available, then updating salt requires should be easy enough. Question is would this be from Neon forward, or backport to Fluorine ?

tkwilliams: I have no particular need to backport it right this moment, since I already have JMESPath installed on my salt masters (due to the other modules such as boto_rds which require it).  That said, there ARE the several modules mentioned below which already use it, so if nothing else, adding it to the requires for Fluorine would help allay any confusion which might otherwise arise if someone happened to be missing it, but tried out those modules :)

twangboy: @cachedout the jmespath module installed cleanly on Windows when I just tried it. Just let me know if I need to add it to the deps for Windows.

cachedout:
I am not at all sure why this is causing the docs to fail to compile. :-/

tkwilliams:
Heh, I was puzzled by the exact same issue :)

rallytime:
The doc build is failing because of the following error:
```
11:22:38 Warning, treated as error:
11:22:38 [autosummary] failed to import u'salt.modules.boto_cloudfront': no module named salt.modules.boto_cloudfront
11:22:38 Failed to import 'salt.modules.boto_cloudfront': no module named salt.modules.boto_cloudfront
11:22:38 make: *** [html] Error 2
```
Which is probably due to that new import of `jmespath`.

tkwilliams:
@rallytime What's the path forward here?  :)  Is there something I need to add or change about the import, or.is it a matter of jmespath would need to be added to the image before the doc build can succeed?

gtmanfred:
if jmespath is required. it sounds like it should be added to the block that catches import errors, and also don't load this module if it is not available.

tkwilliams:
OK, so I'm now befuddled by the prdoc error being emitted :)  Why is it interpreting one (JSON string) argument as multiple?

terminalmage:
@tkwilliams Because there needs to be a blank line after a `code-block` directive. A blank line divides the `code-block` directive from any arguments used to further configure the `code-block` plugin. Because you didn't add a blank line, Sphinx is interpreting the code you're trying to format as arguments to the `code-block` plugin.

Add a blank line after the `code-block` and it should be fine.

tkwilliams:
> @tkwilliams Because there needs to be a blank line after a `code-block` directive. A blank line divides the `code-block` directive from any arguments used to further configure the `code-block` plugin. Because you didn't add a blank line, Sphinx is interpreting the code you're trying to format as arguments to the `code-block` plugin.
> 
> Add a blank line after the `code-block` and it should be fine.

Aha!  You foolishly reveal the FORBIDDEN SECRETS and HIDDEN KNOWLEDGE of the Ancient Rosicrucian Order!  Soon, ALL will know and the conspiracy of silence will fall!

Seriously though, THANKS :)  I would never have figured that one out ...

tkwilliams:
Moar befuddlery:
```
11:52:38 Warning, treated as error:
11:52:38 /var/jenkins/workspace/pr-doc_PR-50360-OUENTEDF7WHIJRMME6T6MNX3O6FAKCCKI3O7PVNNWMIEMNTTZUCQ/salt/states/boto_cloudfront.py:docstring of salt.states.boto_cloudfront.distribution_present:27:Unexpected indentation.
11:52:38 make: *** [html] Error 2
```
Nothing at all wrong in the section it SEEMS to be complaining about...  Any thoughts?

terminalmage:
@tkwilliams I just pushed a commit to this PR that should clear up the remaining docs issues. A few notes:

1. We use double-backticks for emphasis. There were likely single-backticks there before you made your changes, but I replaced all of them in the file with doubles to match how we do things.
2. When doing a bulleted list, you should leave a blank line before the start. You should also leave blank lines between each list element. This is not always necessary, but it _is_ necessary if you are making multi-level bulleted lists (that is, bulleted lists with nested bullet indentation levels).
3. I replaced some references to `present()` in your docs with hyperlinks to the function. This will allow one to click on that link and go right to the docs for that function. I also removed the word "above". It may be "above" it in the code file, but won't necessarily be so when rendered in the docs. In fact, if docs are rendered in alphanumeric/lexicographic order, `present()` would actually have appeared _below_. Better just to use a hyperlink.

tkwilliams:
Checking back in :)
Is this happy camper now, or still blocked on something on my side?

terminalmage:
Looks fine to me :smile:

tkwilliams:
Sweet!  Thanks everyone :)

PR #50361, opened 2018-11-02
Add more gets and deletes for aptly execution module

m03:
### What does this PR do?

* Adds additional gets and deletes for mirror, publish, and snapshot commands.
* Moves the parsing of the output from aptly `show` commands to a separate function.
* Improves the output of `list_published` to allow discerning between the distribution and the prefix.

None

### Previous Behavior

The module was lacking this functionality.

### New Behavior

The module includes this functionality.

### Tests written?

No

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/modules/aptly.py
+++ b/salt/modules/aptly.py
@@ -141,7 +141,7 @@ def _parse_show_output(cmd_ret):
 
     if 'sources' in ret:
         # Match lines like "main: xenial [snapshot]" or "test [local]".
-        source_pattern = '((?P<component>\S+):)?\s*(?P<name>\S+)\s+\[(?P<type>\S+)\]'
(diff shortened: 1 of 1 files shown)


salt/modules/aptly.py
```diff
@@ -50,6 +55,26 @@ def _cmd_run(cmd):
     return cmd_ret['stdout']
 
 
+def _convert_to_closest_type(value):
```
cachedout: This is generic enough that I wonder if we should put it in `salt.utils` so that other modules can use it.

m03: @cachedout At least for the conversion of strings to equivalent boolean values, I agree. I can submit a follow-up PR for that and modify the functionality here afterwards.

m03:
Closing this in favor of submitting a new PR with the changes squashed and new functionality added to stringutils.

PR #50366, opened 2018-11-02
2018.3 fix issue50254

t0fik:
Fix for #50254

Tests: YES

Issue #50254, opened 2018-10-26
file.line adding line every time


jmcook1:
### Description of Issue/Question
2018.3.3 appears to have a bug with file.line. The line is always added even if it already exists. This does not happen on 2018.3.2 minions. I've tried a variety of after/before lines and the line is always added.

### Setup
yum.sls:
```
disable_epel_check_mk_install:
  file.line:
    - name: /etc/yum.repos.d/epel.repo
    - backup: True
    - mode: Ensure
    - location:
      - start
    - before: '\[epel-debuginfo\]'
    - indent: True
    - content: 'exclude=check-mk-agent*'
    - onlyif: 'test -e /etc/yum.repos.d/epel.repo'
```
Contents of /etc/yum.repos.d/epel.repo:
```

/etc/yum.repos.d/epel.repo
[epel]
name=Extra Packages for Enterprise Linux 7 - $basearch
#baseurl=http://download.fedoraproject.org/pub/epel/7/$basearch
mirrorlist=https://mirrors.fedoraproject.org/metalink?repo=epel-7&arch=$basearch
failovermethod=priority
enabled=1
gpgcheck=1
gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-EPEL-7

exclude=check-mk-agent*
[epel-debuginfo]
name=Extra Packages for Enterprise Linux 7 - $basearch - Debug
#baseurl=http://download.fedoraproject.org/pub/epel/7/$basearch/debug
mirrorlist=https://mirrors.fedoraproject.org/metalink?repo=epel-debug-7&arch=$basearch
failovermethod=priority
enabled=0
gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-EPEL-7
gpgcheck=1

[epel-source]
name=Extra Packages for Enterprise Linux 7 - $basearch - Source
#baseurl=http://download.fedoraproject.org/pub/epel/7/SRPMS
mirrorlist=https://mirrors.fedoraproject.org/metalink?repo=epel-source-7&arch=$basearch
failovermethod=priority
enabled=0
gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-EPEL-7
gpgcheck=1
```
### Steps to Reproduce Issue
Run the state on the minion:
```
# salt-call state.apply yum
```
### Versions Report
```
Salt Version:
           Salt: 2018.3.3
 
Dependency Versions:
           cffi: Not Installed
       cherrypy: Not Installed
[...]

PR #50367, opened 2018-11-02
Fix of issue #50254

t0fik:
Port of PR #50366


cachedout:
@t0fik Is this fix different than what's in #50366? If not, this should be closed as the fix will automatically be merged forward.

t0fik:
@cachedout Yes it's the same.

PR #50370, opened 2018-11-04
Log message flow to publish daemon

dwoz:
Log when a message is pushed to the publish daemon process and when we receive it on the other side.


cachedout:
IMHO these should all be at the `trace` level.

PR #50371, opened 2018-11-04
Fix bad documentation reference for required modules

sdodsley:
### What does this PR do?
Fixes an incorrect reference to a required module in the module documentation section

### Commits signed with GPG?

No


rallytime (approved):
Thanks @sdodsley

PR #50374, opened 2018-11-05
Report proper path remove failure to the error log

isbm:
### What does this PR do?

Cosmetic bugfix: report proper error message in the logs while attempting to delete path in the cache.


rallytime (approved):
👍

PR #50380, opened 2018-11-05
systemd: add optional root parameter

aplanas:
### What does this PR do?
Add the optional `--root` parameter to the systemd calls

### DIFF
--- a/salt/modules/systemd.py
+++ b/salt/modules/systemd.py
@@ -56,13 +56,23 @@ def __virtual__():
     Only work on systems that have been booted with systemd
     '''
     if __grains__['kernel'] == 'Linux' \
-            and salt.utils.systemd.booted(__context__):
+       and salt.utils.systemd.booted(__context__):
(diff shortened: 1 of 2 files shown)


salt/modules/service.py
```diff
@@ -41,6 +41,7 @@ def __virtual__():
         'elementary OS',
         'McAfee  OS Server',
         'Raspbian',
+        'SUSE',
```
cachedout: This seems unrelated to this PR?

cachedout (commented):
IMHO it would be good to add some docstrings indicating what the `root` parameter does.

salt/modules/service.py
aplanas: Very true

aplanas:
@cachedout right, I also removed the parameter for certain actions where this parameter is silently ignored.

cachedout (approved):

Issue #50062, opened 2018-10-15
max_event_size does not always trim large events. 


whytewolf:
### Description of Issue/Question
[max_event_size](https://docs.saltstack.com/en/latest/ref/configuration/master.html#max-event-size) Does not always trim long events. 

If the event has a `key:value` pair where the value is a dict and not a string than the `dict_trim` will run `sys.getsizeof` on the dict. This returned size is much lower than the issue size even in cases where it shouldn't be. 

However just serializing that dict and running the trim on it also doesn't seem like the right answer as that could be trimming important information used for parsing the event data. 

I'm not sure what the correct course is for this. But I can see this causing headaches for very large state runs. 

### Setup
Reducing this down to a more simple test can be done with the following.

create 2 largeish random files and convert them to base64 so that we don't trigger diff binary mode. 
`for i in 1 2; do dd if=/dev/urandom of=/tmp/testfile bs=512 count=10000; base64 /tmp/testfile > /srv/salt/testfile.${i}; rm -f /tmp/testfile; done`

then alternate between these 2 states. 

```
test1:
  file.managed:
    - name: /tmp/testfile
    - source: salt://testfile.1

test2:
  file.managed:
    - name: /tmp/testfile
    - source: salt://testfile.2
```
This will create a huge return dict inside the return data. which will get sent back to the master as an job event will not be trimmed by dict_trim.

### Steps to Reproduce Issue
See above

### Versions Report
```
Salt Version:
           Salt: 2018.3.2

Dependency Versions:
           cffi: Not Installed
       cherrypy: Not Installed
       dateutil: Not Installed
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.7.2
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.5.6
   mysql-python: Not Installed
      pycparser: Not Installed
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: Not Installed
         Python: 2.7.5 (default, Aug  4 2017, 00:39:18)
   python-gnupg: Not Installed
         PyYAML: 3.11
          PyZMQ: 15.3.0
           RAET: Not Installed
          smmap: Not Installed
        timelib: Not Installed
        Tornado: 4.2.1
            ZMQ: 4.1.4

System Versions:
[...]

PR #50382, opened 2018-11-05
[2018.3] Fix to utils/dicttrim.py to honor max_event_size with nested dictionaries

garethgreenaway:
### What does this PR do?
Adding _trim_dict_in_dict to utils/dictrim.py to be called from trim_dict when data contains a nested dictionary.  This will ensure that values will still be trimmed correctly.

#50062 

### Previous Behavior
Event data that contained nested dictionaries that exceeded the max_event_size would not be trimmed because the size of the key was reported, not the actual data inside the dictionary.

### New Behavior
If a nested dictionary is found in the event data dictionary, the `_trim_dict_in_dict` is called recursively, ensuring all values are trimmed correctly.

### Tests written?
Not yet.

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/utils/dicttrim.py
+++ b/salt/utils/dicttrim.py
@@ -6,6 +6,22 @@ import sys
 import salt.payload
 
 
+def _trim_dict_in_dict(data, max_val_size, replace_with):
+    '''
+    Takes a dictionary, max_val_size and replace_with
+    and recursively loops through and replaces any values
+    that are greater than max_val_size.
+    '''
+    for key in data:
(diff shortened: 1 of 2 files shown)


whytewolf (approved):
This change works well. I noticed no reduction in performance.

And enough of the data is saved that it doesn't compromise the return results.

garethgreenaway:
@whytewolf Thanks!  I want to include a simple test just to make everything works as expected.

cachedout (approved):

rallytime (approved):

Issue #50388, opened 2018-11-05
salt-ssh to sparc solaris 64 bit uses 32 bit library path to test for libcrypto


ggates1020:
### Description of Issue/Question
Using salt-ssh and connecting to a sparc solaris 64 bit system running 64 bit python uses the 32 bit library path to test for libcrypto resulting in an OSError.

        Traceback (most recent call last):
          File "/var/tmp/.ajb_303cbc_salt/salt-call", line 15, in <module>
            salt_call()
          File "/var/tmp/.ajb_303cbc_salt/py2/salt/scripts.py", line 405, in salt_call
            import salt.cli.call
          File "/var/tmp/.ajb_303cbc_salt/py2/salt/cli/call.py", line 5, in <module>
            import salt.utils.parsers
          File "/var/tmp/.ajb_303cbc_salt/py2/salt/utils/parsers.py", line 27, in <module>
            import salt.config as config
          File "/var/tmp/.ajb_303cbc_salt/py2/salt/config/__init__.py", line 98, in <module>
            _DFLT_IPC_WBUFFER = _gather_buffer_space() * .5
          File "/var/tmp/.ajb_303cbc_salt/py2/salt/config/__init__.py", line 88, in _gather_buffer_space
            import salt.grains.core
          File "/var/tmp/.ajb_303cbc_salt/py2/salt/grains/core.py", line 44, in <module>
            import salt.utils.dns
          File "/var/tmp/.ajb_303cbc_salt/py2/salt/utils/dns.py", line 32, in <module>
            import salt.modules.cmdmod
          File "/var/tmp/.ajb_303cbc_salt/py2/salt/modules/cmdmod.py", line 34, in <module>
            import salt.utils.templates
          File "/var/tmp/.ajb_303cbc_salt/py2/salt/utils/templates.py", line 32, in <module>
            import salt.utils.http
          File "/var/tmp/.ajb_303cbc_salt/py2/salt/utils/http.py", line 41, in <module>
            import salt.loader
          File "/var/tmp/.ajb_303cbc_salt/py2/salt/loader.py", line 31, in <module>
            import salt.utils.event
          File "/var/tmp/.ajb_303cbc_salt/py2/salt/utils/event.py", line 74, in <module>
            import salt.payload
          File "/var/tmp/.ajb_303cbc_salt/py2/salt/payload.py", line 17, in <module>
            import salt.crypt
[...]

PR #50389, opened 2018-11-05
Fix libcrypto search for sparc solaris 64-bit

ggates1020:
On solaris sparc the python ctypes.util.find_library() method defaults to the 32 bit library path.  This will test if python is running in 64 bit and request the appropriate library search path.

### What does this PR do?

On Solaris sparc systems this will test if python is running in 64 bit and request the appropriate library search path.

#50388 

### Previous Behavior
Always checked only the 32 bit library path

### New Behavior
If the library is not found and the platform is sunos5, it will run the test to see if it is 64 bit and check the appropriate library path.

### Tests written?

No

### Commits signed with GPG?

No

### DIFF
--- a/salt/utils/rsax931.py
+++ b/salt/utils/rsax931.py
@@ -36,6 +36,9 @@ def _load_libcrypto():
             'libcrypto.so*'))[0])
     else:
         lib = find_library('crypto')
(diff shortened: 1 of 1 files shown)


cachedout (approved):
LGTM

PR #50396, opened 2018-11-06
service: SUSE is not based on sysvinit anymore

aplanas:
openSUSE and SLE are systemd based distributions

### DIFF
--- a/salt/modules/service.py
+++ b/salt/modules/service.py
@@ -41,6 +41,7 @@ def __virtual__():
         'elementary OS',
         'McAfee  OS Server',
         'Raspbian',
+        'SUSE',
     ))
     if __grains__.get('os') in disable:
         return (False, 'Your OS is on the disabled list')


brejoc (changes requested):
I think we can't do that. We've still got SLES11 around and that is afaik **not** systemd based. So for SLES11 we would fall back to sysvinit and for all the others systemd would be picked -right?

isbm:
@brejoc but we also not running anything but 2016.11 on SLE11. Why would we stop this in development branch? I think this is good to go.

brejoc:
@isbm Ah, good point! Then we are fine!

brejoc (approved):

isbm (approved):

isbm:
@gtmanfred But pip installed is unnecessary in SUSE world as we have supported packages and _strongly_ recommend to use them. If someone is using pip — on their own risk.

brejoc:
While I generally agree, that this change isn't really needed, running anything but our packaged version on SLES11 is really, really hard. SLES11 doesn't even ship with `pip` and let's not talk about Python versions. So: 🤷

aplanas:
IMHO the reviews are correct and the merger was also correct. And the change, in some sense, is required. It is used to document a fact, that in 2018 the current SUSE family is systemd based.

But I also agree that there is a bigger problem to solve, that this PR is not doing.

The logic behind the decision of using sysv or systemd is quite error prone. IMHO the list of `disable` is not a good approach and the comment about the usage is not precise (this is what drives me to do the change). For example, there are names with two spaces there (`McAfee  OS Server`), that I do not know if this is a bug or not. And the delegation of a more modern module to see if this one needs to be loaded or not is a bit unexpected.

PR #50398, opened 2018-11-06
Expand aptly gets and deletes

m03:
### What does this PR do?

* Adds additional gets and deletes for mirror, publish, and snapshot commands.
* Adds more clarity to the docstrings of existing functions.
* Moves the parsing of the output from aptly `show` commands to a separate function.
* Improves the output of `list_published` to allow discerning between the distribution and the prefix.
* Add `to_bool` function to stringutils.

None

### Previous Behavior

The modules were lacking this functionality.

### New Behavior

The modules include this functionality.

### Tests written?

Yes (for `to_bool`)

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/modules/aptly.py
+++ b/salt/modules/aptly.py
@@ -61,7 +61,7 @@ def _convert_to_closest_type(value):
     '''
     value = salt.utils.stringutils.to_bool(value.strip())
 
-    if type(value) == type(True):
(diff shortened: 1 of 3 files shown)


cachedout (approved):
This is very clean code. Nicely done.

PR #50405, opened 2018-11-06
Fix integration.modules.test_decorators for Windows

twangboy:
### What does this PR do?
Fixes 5 failing tests in `integration.modules.test_decorators` on Windows

Adds a exit_code.cmd file for Windows
Uses `posix=True` for shlex_split on Windows

found in jenkins testing
## Tests written?
Yes

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/utils/decorators/__init__.py
+++ b/salt/utils/decorators/__init__.py
@@ -23,6 +23,10 @@ from salt.log import LOG_LEVELS
 # Import 3rd-party libs
 from salt.ext import six
 
+IS_WINDOWS = False
+if getattr(sys, 'getwindowsversion', False):
+    IS_WINDOWS = True
+
 log = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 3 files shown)


dwoz (approved):

rallytime (approved):

rallytime:
@twangboy That test is still failing on the oy2 windows build. Can you take another look?

salt/utils/decorators/__init__.py
```diff
@@ -23,6 +23,10 @@
 # Import 3rd-party libs
 from salt.ext import six
 
+IS_WINDOWS = False
+if getattr(sys, 'getwindowsversion', False):
```
cachedout: Hrm. I think this might neutralize the gains we get from the memoization that we get in `salt.utils.platform.is_windows` and that would not be great. Is there any way to achieve this without doing OS detection?

twangboy: Using `salt.utils.platform.is_windows` was throwing some memoize errors and the test suite wouldn't even start.

We need to set `posix=False` when doing the shlex_split on Windows, otherwise it removes all the `\` in the file path to the batch script that is run. So, I kind of need to detect Windows. I didn't want to import another library, so I went with this method. I could import sys and do `if sys.name == 'nt'` or something... if that makes a difference.

twangboy:
@rallytime That failure is a timeout issue caused by a dropped message. Dwoz is working on a fix that should address all those.

salt/utils/decorators/__init__.py
cachedout: After thinking about this for a bit, I think this is fine. This shouldn't be per-invocation so we'll only get the overhead on import.

PR #50410, opened 2018-11-07
Add a new drbd monitor function via drbdadm to deprecate the old method of drbd-overview

nick-wang:
### What does this PR do?
Add a new method to monitor the drbd resources status.

drbd-overview is removed in the latest drbd-utils-9.6.0. Refer to [drbd-utils commit](https://github.com/LINBIT/drbd-utils/commit/811b67219101e18eafdaee34a94621042123b404)
The original solution with drbd-overview is deprecated/removed in the latest version, the original salt function only support show 2 nodes with 1 volume. Add another monitor approach with "drbdadm status" to support all drbd versions, new function also support multiple drbd resources, including multiple volumes/peernodes of each resource.

### Previous Behavior
Show drbd status in 2 nodes(1 volume) scenario via drbd-overview. Leave the old method in case some users still using salt with the old drbd version/method.

### New Behavior
Add a new method to show drbd status via drbdadm, which also support more than 2 nodes/volumes.
drbdadm also support multiple drbd versions including the latest version.

### Tests written?

Yes

### Commits signed with GPG?

No

### DIFF
--- a/salt/modules/drbd.py
+++ b/salt/modules/drbd.py
@@ -5,6 +5,7 @@ DRBD administration module
 from __future__ import absolute_import, print_function, unicode_literals
 
 import logging
+from salt.ext import six
 
 log = logging.getLogger(__name__)
 
@@ -37,22 +38,24 @@ def _analyse_status_type(line):
 
     if spaces is None:
         return None
-    elif spaces == 0:
-        return 'RESOURCE'
-    elif spaces == 2:
-        if ' disk:' in line:
-            return 'LOCALDISK'
-        elif ' role:' in line or ' connection:' in line:
-            return 'PEERNODE'
-        else:
-            return 'UNKNOWN'
-    elif spaces == 4:
-        if ' peer-disk:' in line:
-            return 'PEERDISK'
-        else:
-            return 'UNKNOWN'
-    else:
-        return 'UNKNOWN'
+
+    switch = {
+        0: 'RESOURCE',
+        2: {' disk:': 'LOCALDISK', ' role:': 'PEERNODE', ' connection:': 'PEERNODE'},
+        4: {' peer-disk:': 'PEERDISK'}
+    }
+
+    ret = switch.get(spaces, 'UNKNOWN')
+
+    # isinstance(ret, str) only works when run directly, calling need unicode(six)
+    if isinstance(ret, six.text_type):
+        return ret
+
+    for x in ret:
+        if x in line:
+            return ret[x]
+
+    return 'UNKNOWN'
 
 
 def overview():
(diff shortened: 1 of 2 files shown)


cachedout:
@nick-wang Looks like a test is unhappy here: https://jenkinsci.saltstack.com/job/pr-kitchen-centos7-py3/job/PR-50410/7/testReport/junit/unit.modules.test_drbd/DrbdTestCase/test_status/

nick-wang:
> Looks like a test is unhappy here: https://jenkinsci.saltstack.com/job/pr-kitchen-centos7-py3/job/PR-50410/7/testReport/junit/unit.modules.test_drbd/DrbdTestCase/test_status/

Thanks for review. It is a regression when fixing "Cognitive Complexity" of code climate...
I am working on refactoring

cachedout:
@nick-wang Sounds good. Keep us posted. Thanks!

nick-wang:
@cachedout Just fixed all issues, please help to review. 
Thanks in advanced.

salt/modules/drbd.py
```diff
@@ -20,9 +21,117 @@ def _analyse_overview_field(content):
     return content, ""
 
 
+def _count_spaces(line):
```
cachedout: This doesn't seem like it does what I would expect it to. Ex:

```
def counter():
    line = "a b c"
    spaces = 0
    for i in line:
        print(i)
        if i.isspace():
            spaces += 1
        else:
            return spaces

spaces = counter()
print("Found {0} spaces".format(spaces))
```

```
gray% python ./foo.py
a
Found 0 spaces
```

nick-wang: It is my bad of naming the function, the idea is suppose to count the number of spaces before the first alphabeta character.
The output of "drbdadm status" doesn't have a standard format, since it may have multiple resources, each resources may have multiple peer nodes or volumes. So it use two spaces in front of a line as an indentation to figure out the section.

I will rename the function like _count_spaces_startswith()

nick-wang:
The current failure of "jenkins/pr/py2-ubuntu-1604" and "jenkins/pr/py3-windows-2016" may not caused by the request.

nick-wang:
Rebase the commits based on the latest code to make everything in order.

nick-wang:
Rebase to the latest due to jenkins check failure by branch out of date.
Please help to review, thanks.

nick-wang:
@cachedout This pull request has nothing to do with the jenkins failure of docs, actually it failed periodicly even without any changes when rebuilt in this pull request.(Please refer to the previous build log.) 

I rebased the commits based on the latest again, so that all checks (except docs...) can pass.
Please help to check, thanks in advanced.

nick-wang:
@cachedout Could you please have a look on this one? The check for docs is self-healing, so all check passed.

cachedout (approved):

nick-wang:
@cachedout  Thanks for your review on this case.

s0undt3ch:
Port to master https://github.com/saltstack/salt/pull/56768

Issue #47787, opened 2018-05-23
cmd.run on Windows does not default to the "runas" user's HOME directory nor obey the "cwd" parameter..


vernondcole:
### Description of Issue/Question

The documentation for [cmdmod.run](https://docs.saltstack.com/en/latest/ref/modules/all/salt.modules.cmdmod.html#salt.modules.cmdmod.run) states "cwd (str) -- ... Defaults to the home directory of the user specified by `runas`".

In fact, on Windows it defaults to `c:\salt\bin`.

### Setup

install Vagrant and VirtualBox on a suitable workstation...

```(bash)
git clone https://github.com/salt-bevy/salt-bevy.git
cd salt-bevy
pip install pyyaml,ifaddr
./join-bevy.sh
vagrant up bevymaster
vagrant up win10
vagrant up win12
vagrant up win16
```
Note: you may have to repeat "vagrant up" and/or "vagrant provision" to get Windows minions running. 

Also, on my test system I am actually running the three Windows VMs on three
different hosts, using a Raspberry Pi Salt-Master, but write it up here as all four running 
as VMs on the same host.

### Steps to Reproduce Issue

``` 
vagrant ssh bevymaster
```
on the Salt Master...
```
$sudo salt win* user.add bobz password=x1234567!
$sudo salt win* user.addgroup bobz Administrators
$sudo salt win* user.addgroup bobz Users
$sudo salt win* cmd.run "echo xxx > %userprofile%\xxx.txt" runas=bobz password=x1234567!

win10:
win12:
win16:
```
* Wrong answer v v v
```
$sudo salt win* cmd.run "dir" runas=bobz password=x1234567!

win10:
     Volume in drive C is System
     Volume Serial Number is AC86-0FAE
    
     Directory of c:\salt\bin
    
    05/22/2018  08:40 PM    <DIR>          .
    05/22/2018  08:40 PM    <DIR>          ..
    05/22/2018  08:40 PM    <DIR>          DLLs
    05/22/2018  08:40 PM    <DIR>          include
    05/22/2018  08:41 PM    <DIR>          Lib
    04/25/2018  11:28 PM         2,265,600 libeay32.dll
    05/22/2018  08:40 PM    <DIR>          libs
    04/25/2018  11:28 PM           398,336 libsodium.dll
    09/16/2017  08:30 PM            38,580 LICENSE.txt
    04/25/2018  11:28 PM           963,232 msvcr120.dll
[...]

PR #50415, opened 2018-11-07
cwd needs to be passed from runas() to runas_system()

flassman:
Fixes parts of #47787

### What does this PR do?

47787 ends with a comment that something has fixed in develop which could be backported.
Since this might be a major effort, my suggestion is to first merge my little fix which is much easier to do.
While I'm no expert neither in salt nor in windows, my change seems straightforward enough to be harmless ;)

with my fix, the cwd parameter gets gets passed into the runas_system() method

### Previous Behavior

salt myminion cmd.run "dir" runas=myuser password=mysecret cwd=C:/Windows/Temp
would be executed in C:/salt

### New Behavior

salt myminion cmd.run "dir" runas=myuser password=mysecret cwd=C:/Windows/Temp
would be executed in C:/Windows/Temp

### Tests written?

No

### Commits signed with GPG?

No

### DIFF
--- a/salt/utils/win_runas.py
+++ b/salt/utils/win_runas.py
@@ -271,7 +271,7 @@ def make_inheritable(token):
                                     win32con.DUPLICATE_SAME_ACCESS)
 
 
-def runas_system(cmd, username, password):
+def runas_system(cmd, username, password, cwd=None):
     # This only works as system, when salt is running as a service for example
 
     # Check for a domain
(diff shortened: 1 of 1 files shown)


cachedout:
@dwoz Thoughts?

dwoz (approved):

twangboy (approved):

Issue #32245, opened 2016-03-30
git_pillar should permit automatic mapping of environment to branch a-la gitfs 


tkwilliams:
Gitfs (for the salt side of things) can be configured to dynamically map git branches to environments, which is stunningly useful for scenarios where development takes place in numerous, transient, "feature branches"; which are then merged into "well-known" branches such as dev, qa and prd as the code promotion process.

git_pillar, contrariwise, must be configured in the master conf with explicit, exhaustive,  mappings of environments to branches.  Since, at least in our case, good data/code separation has led to about 90% of changes taking place in pillar rather than state files, we're finding the actual USEFULNESS of dynamic branch mapping to be almost completely nullified by the lack of such functionality in pillar.

We've tried a large assortment of things to work around this lack (stupid Jinja tricks, strange combinations of git_pillar_\* config options and so on) but the basic lack of a way to do anything but statically map predefined environments to branches seems a hard limit on our workflow.

If the above is not detailed enough I can certainly give some examples of our desired workflow, but I think the basic usefulness of such a feature is pretty evident :)

Thanks!

jfindlay:
@tkwilliams, thanks for reporting.  This seems like a useful feature to me.

jfindlay:
One thing to be careful about with such a feature is that enabling inexplicit behavior for pillar may reduce clarity on how pillars are assigned to minions and by this cause expose secret information to the wrong system/party.

tkwilliams:
Earlier discussion on the subject in salt-users:
https://groups.google.com/forum/#!topic/salt-users/OXHpBuf28QU

tkwilliams:
@cachedout @rallytime @jfindlay I don't suppose anyone has had a chance to look into this, have they?

rallytime:
@terminalmage Has done quite a bit of work on git pillar. I think he'd be the best person to respond here.

terminalmage:
I don't see this ever being implemented, to be honest.  git_pillar and gitfs aren't the same thing, they have fundamental differences. With gitfs, individual files are requested, which lends itself easily to mapping an environment to a branch, tag or SHA. We never actually check out a branch, we just fetch from the repo and then pull files out of the fetched objects. With git_pillar, each defined repo must be checked out, so that the SLS files within it exist locally and can be processed similar to states.

[...]

PR #50417, opened 2018-11-07
Get the right target when using "__env__" on git ext_pillar to avoid merging problems

meaksh:
### What does this PR do?
This PR fixes an issue when **multiple** Git repositories are set as `ext_pillar` and they're using `__env__` as the branch name.

When `git_pillar_includes` is enabled, the mechanism for extending and merging the different pillars is failing because the `__env__` branch is not yet resolved: https://github.com/saltstack/salt/blob/59894e2c830585f7a0412dcc6b6637797459fe84/salt/pillar/git_pillar.py#L441-L443

Because of that, given we have the following configuration:

- On `/etc/salt/master`:
```yaml
ext_pillar:
  - git:
    - __env__ git@gs.intern.priv:/srv/git/pillar:
      - name: gitinfo
      - privkey: "/etc/keys/id_rsa"
      - pubkey: "/etc/keys/id_rsa.pub"
    - __env__ git@gs.intern.priv:/srv/git/webapps:
      - name: webinfo
      - mountpoint: nowhere
      - privkey: "/etc/keys/id_rsa"
      - pubkey: "/etc/keys/id_rsa.pub"
```

#### First repository (gitinfo)

- On the git server `/srv/git/pillar/top.sls`:
```yaml
"{{saltenv}}":
  '*':
    - mytest 
    - nowhere.webinfo_external
```

- On the git server `/srv/git/pillar/mytest.sls`:
```yaml
mytest: added new stuff
```

#### Second repository (webinfo)

- On the git server `/srv/git/webapps/webinfo_external.sls`:
```yaml
mytest_external: True
```

### Previous Behavior
```yaml
# salt \* pillar.items

minion1.intern.priv:
    ----------
    _errors:
        - Specified SLS 'nowhere.webinfo_external' in environment 'master' is not available on the salt master
    mytest:
        added new stuff
```

### New Behavior
```yaml
# salt \* pillar.items

minion1.intern.priv:
    ----------
    mytest:
        added new stuff
    mytest_external:
        True
```

### Tests written?

Yes

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/utils/gitfs.py
+++ b/salt/utils/gitfs.py
@@ -2980,7 +2980,7 @@ class GitPillar(GitBase):
                 if repo.env:
                     env = repo.env
                 else:
-                    env = 'base' if repo.branch == repo.base else repo.branch
+                    env = 'base' if repo.branch == repo.base else repo.get_checkout_target()
                 if repo._mountpoint:
                     if self.link_mountpoint(repo):
                         self.pillar_dirs[repo.linkdir] = env
--- a/tests/integration/pillar/test_git_pillar.py
+++ b/tests/integration/pillar/test_git_pillar.py
@@ -383,7 +383,7 @@ class GitPythonMixin(object):
         self.assertEqual(
             ret,
             {'branch': 'master',
-             'motd': u'The force will be with you. Always.',
+             'motd': 'The force will be with you. Always.',
              'mylist': ['master'],
              'mydict': {'master': True,
                         'nested_list': ['master'],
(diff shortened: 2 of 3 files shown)


sathieu:
I've been hit by this. Thanks for this PR.

sathieu:
can you add a test with one repo on `top_only` and the other on `__env__`?

sathieu:
Also, beware of the race condition (see #32245)

terminalmage (approved):

terminalmage:
Can you please be more specific than "this breaks"?

terminalmage:
I think I see actually, the target just returns the branch name, which is fine when the pillarenv is the branch, but it doesn't properly get identified as the `base` env when that branch is returned by `get_checkout_target()`.

PR #50418, opened 2018-11-07
Add new chroot module

aplanas:
### What does this PR do?
Add a new chroot module in saltstack.

Based on dockermod.call, this module implement an way to execute random salt modules inside a chroot environment.

Tests are included.

### DIFF
--- a/salt/modules/chroot.py
+++ b/salt/modules/chroot.py
@@ -1,0 +1,165 @@
+# -*- coding: utf-8 -*-
+#
+# Author: Alberto Planas <aplanas@suse.com>
+#
+# Copyright 2018 SUSE LINUX GmbH, Nuernberg, Germany.
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+'''
+:maintainer:    Alberto Planas <aplanas@suse.com>
+:maturity:      new
+:depends:       None
+:platform:      Linux
+'''
+from __future__ import absolute_import, print_function, unicode_literals
+import logging
+import os
+import sys
+import tempfile
+
+from salt.defaults.exitcodes import EX_OK
+from salt.exceptions import CommandExecutionError
+from salt.utils.args import clean_kwargs
+
+log = logging.getLogger(__name__)
+
+
+def __virtual__():
+    '''
+    Chroot command is required.
+    '''
+    if __utils__['path.which']('chroot') is not None:
+        return True
+    else:
+        return (False, 'Module chroot requires the command chroot')
+
+
+def exist(name):
+    '''
+    Return True if the chroot environment is present.
+    '''
+    dev = os.path.join(name, 'dev')
+    proc = os.path.join(name, 'proc')
+    return all(os.path.isdir(i) for i in (name, dev, proc))
+
+
+def create(name):
+    '''
+    Create a basic chroot environment.
+
+    Note that this environment is not functional. The caller needs to
+    install the minimal required binaries, including Python if
+    chroot.call is called.
+
+    name
+        Path to the chroot environment
+
+    CLI Example:
+
+    .. code-block:: bash
+
+        salt myminion chroot.create /chroot
+
+    '''
+    if not exist(name):
+        dev = os.path.join(name, 'dev')
+        proc = os.path.join(name, 'proc')
+        try:
+            os.makedirs(dev, mode=0o755)
+            os.makedirs(proc, mode=0o555)
+        except OSError as e:
+            log.error('Error when trying to create chroot directories: %s', e)
+            return False
+    return True
+
+
+def call(name, function, *args, **kwargs):
+    '''
+    Executes a Salt function inside a chroot environment.
+
+    The chroot does not need to have Salt installed, but Python is
+    required.
+
+    name
+        Path to the chroot environment
+
+    function
+        Salt execution module function
+
+    CLI Example:
+
+    .. code-block:: bash
+
+        salt myminion chroot.call /chroot test.ping
+
+    '''
+
+    if not function:
+        raise CommandExecutionError('Missing function parameter')
+
+    if not exist(name):
+        raise CommandExecutionError('Chroot environment not found')
+
+    # Create a temporary directory inside the chroot where we can
+    # untar salt-thin
+    thin_dest_path = tempfile.mkdtemp(dir=name)
+    thin_path = __utils__['thin.gen_thin'](
+        __opts__['cachedir'],
+        extra_mods=__salt__['config.option']('thin_extra_mods', ''),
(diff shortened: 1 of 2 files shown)


aplanas:
Ready for review. I decided not to include the sls call in the chroot code, as I really do not see as relevant as the one that we have in docker.

But if I find that the sls is indeed useful in this context, I will implement it.

aplanas:
Also note that the header is now under apache license. I copied it from the apache site [1]. If this is not correct, please, point me to a good one.

[1] https://www.apache.org/legal/src-headers.html

cachedout:
@aplanas No need for a source license in each individual file. The entire project is covered under Apache 2.0. It doesn't hurt to have it there, though.

salt/modules/chroot.py
```diff
@@ -0,0 +1,164 @@
...
+
+    # Create a temporary directory inside the chroot where we can
+    # untar salt-thin
+    thin_dest_path = tempfile.mkdtemp(dir=name)
+    thin_path = __utils__['thin.gen_thin'](
+        __opts__['cachedir'],
+        extra_mods=__salt__['config.option']('thin_extra_mods', ''),
+        so_mods=__salt__['config.option']('thin_so_mods', '')
+    )
+    untar_cmd = ['python', '-c', (
```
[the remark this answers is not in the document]
aplanas: Good idea!

salt/modules/chroot.py
```diff
@@ -0,0 +1,164 @@
...
+
+    '''
+    if not exist(name):
+        dev = os.path.join(name, 'dev')
+        proc = os.path.join(name, 'proc')
+        return __salt__['cmd.retcode'](['mkdir', '-p', name, dev, proc]) == 0
+    return True
+
+
+def call(name, function, *args, **kwargs):
```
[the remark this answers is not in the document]
aplanas: I do not get it. You mean something like `def call(name, function, *args, **extra_options):`?

salt/modules/chroot.py
```diff
@@ -0,0 +1,164 @@
...
+            '--metadata',
+            '--local',
+            '--log-file', os.path.join(chroot_path, 'log'),
+            '--cachedir', os.path.join(chroot_path, 'cache'),
+            '--out', 'json',
+            '-l', 'quiet',
+            '--',
+            function
+        ] + list(args) + [
+            '{}={}'.format(k, v) for (k, v) in kwargs.items()
```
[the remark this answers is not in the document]
aplanas: I think so. This command delegate whatever salt execution module to the inner chroot, and the parameters are unknown.

```
salt '*' chroot.call /chroot test.ping
salt '*' chroot.call /chroot test.echo 'hi there'
salt '*' chroot.call /chroot test.random_hash size=200 hash_type=sha1
```

salt/modules/chroot.py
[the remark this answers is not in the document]
isbm: @cachedout but since when this became a rule?

```
$ grep 'def' *py | grep kwargs | wc -l
1333
```

salt/modules/chroot.py
```diff
@@ -0,0 +1,160 @@
...
+import logging
+import os
+import sys
+import tempfile
+
+from salt.exceptions import CommandExecutionError
+
+log = logging.getLogger(__name__)
+
+__virtualname__ = 'chroot'
```
isbm: Again, this is needed only in case there will be another version of this or you want to rename its access name.

salt/modules/chroot.py
```diff
@@ -0,0 +1,160 @@
...
+
+
+def __virtual__():
+    '''
+    Chroot command is required.
+    '''
+    return __utils__['path.which']('chroot') is not None
```
isbm: Otherwise needs to be a message why this module has not been lazyloaded.

salt/modules/chroot.py
```diff
@@ -0,0 +1,160 @@
...
+
+    .. code-block:: bash
+
+        salt myminion chroot.create /chroot
+
+    '''
+    if not exist(name):
+        dev = os.path.join(name, 'dev')
+        proc = os.path.join(name, 'proc')
+        return __salt__['cmd.retcode'](['mkdir', '-p', name, dev, proc]) == 0
```
isbm: Please use python function instead from the module `os`.

salt/modules/chroot.py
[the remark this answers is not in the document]
isbm: @cachedout why? We never preferred that last 4-5 years. Is this changed now?

salt/modules/chroot.py
```diff
@@ -0,0 +1,160 @@
...
+            '--local',
+            '--log-file', os.path.join(chroot_path, 'log'),
+            '--cachedir', os.path.join(chroot_path, 'cache'),
+            '--out', 'json',
+            '-l', 'quiet',
+            '--',
+            function
+        ] + list(args) + [
+            '{}={}'.format(k, v) for (k, v) in kwargs.items()
+            if not k.startswith('__')
```
isbm: I'd rather use, again, utility `salt.utils.args` for cleaning this up instead.

salt/modules/chroot.py
```diff
@@ -0,0 +1,160 @@
...
+        ]
+        ret = __salt__['cmd.run_chroot'](name, [str(x) for x in salt_argv])
+        if ret['retcode'] != 0:
```
isbm: Please do use constants instead:

```python
import salt.defaults.exitcodes
...
if ret['retcode'] == salt.defaults.exitcodes.EX_OK:
```

salt/modules/chroot.py
```diff
@@ -0,0 +1,160 @@
...
+            raise CommandExecutionError(ret['stderr'])
+
+        # Process "real" result in stdout
+        try:
+            data = __utils__['json.find_json'](ret['stdout'])
+            local = data.get('local', data)
+            if isinstance(local, dict):
+                if 'retcode' in local:
```
isbm: No need to nest the same thing:
```python
if isinstance(local, dict) and 'retcode' in local:
   ...
```

tests/unit/modules/test_chroot.py
```diff
@@ -0,0 +1,172 @@
...
+    @patch('tempfile.mkdtemp')
+    def test_call(self, mkdtemp, exist):
+        '''
+        Test execution of Salt functions in chroot.
+        '''
+        # Basic input validation
+        exist.return_value = False
+        self.assertRaises(CommandExecutionError, chroot.call, '/chroot', '')
+        self.assertRaises(CommandExecutionError, chroot.call, '/chroot',
+                          'test.ping')
```
isbm: can be on the same line though

tests/unit/modules/test_chroot.py
```diff
@@ -0,0 +1,172 @@
...
+            salt_mock['cmd.retcode'].assert_not_called()
+
+        exist.return_value = False
+        with patch.dict(chroot.__salt__, salt_mock):
+            self.assertTrue(chroot.create('/chroot'))
+            salt_mock['cmd.retcode'].assert_called_once()
+
+    @patch('salt.modules.chroot.exist')
+    @patch('tempfile.mkdtemp')
+    def test_call(self, mkdtemp, exist):
```
isbm: Again, this UT should be broken into several.

isbm (changes requested):

salt/modules/chroot.py
aplanas: Fixed.

salt/modules/chroot.py
aplanas: done

isbm (approved):

salt/modules/chroot.py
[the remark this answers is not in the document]
cachedout: @isbm We have always preferred this. There are cases where it is not possible, but whenever it is possible, we pretty much always ask that it be changed.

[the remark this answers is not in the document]
aplanas: I think that we can agree that in this case is not possible to set kwargs, as this module enables the execution of any salt module inside a chroot, via salt-thin. So is not possible to know beforehand the all parameters that can be used.

Issue #50393, opened 2018-11-06
zmq socket closed before sending all data


tsaridas:
### Description of Issue/Question
While testing sending command to multiple minions (200+) at the same time we noticed that some syndic randomly did not send publish the event to the minions and tracked down the issue to the ZeroMQPubServerChannel class.

After adding debug log after

https://github.com/saltstack/salt/blob/develop/salt/transport/zeromq.py#L938

the problem seemed to be fixed thinking that the socket is closed before all events are send.

Worth noticing that we have zmq_filtering enabled on the master and minion side so an event is created for each minion.

We switched from zmq socket to normal sockets ( AF_UNIX, SOCK_DGRAM )  for publishing events and the issue seems to have been resolved.

https://github.com/saltstack/salt/blob/develop/salt/transport/zeromq.py#L909
https://github.com/saltstack/salt/blob/develop/salt/transport/zeromq.py#L910

### Setup
(Please provide relevant configs and/or SLS files (Be sure to remove sensitive info).)

### Steps to Reproduce Issue
(Include debug logs if possible and relevant.)

### Versions Report
2016.11.10

garethgreenaway:
@saltstack/team-core Anyone have thoughts on this one?

cachedout:
@garethgreenaway Right now this is something that @dwoz is looking into.

garethgreenaway:
@cachedout Thanks.  Assigned to @dwoz.

dwoz:
@tsaridas can you please provide us the output of `salt --versions-report` ?

tsaridas:
```
Salt Version:

           Salt: 2016.11.10


Dependency Versions:

           cffi: Not Installed

       cherrypy: 3.2.2

       dateutil: 1.4.1

      docker-py: Not Installed

          gitdb: Not Installed

      gitpython: Not Installed

          ioflo: Not Installed

         Jinja2: 2.7.3

        libgit2: Not Installed

        libnacl: Not Installed

       M2Crypto: 0.20.2

           Mako: Not Installed

   msgpack-pure: Not Installed

msgpack-python: 0.4.8

   mysql-python: Not Installed

      pycparser: Not Installed

       pycrypto: 2.6.1

   pycryptodome: 3.4.3

         pygit2: Not Installed

         Python: 2.6.6 (r266:84292, Aug  9 2016, 06:11:56)

   python-gnupg: Not Installed

         PyYAML: 3.11

          PyZMQ: 14.5.0

[...]

Issue #46553, opened 2018-03-15
Salt occasionally lose command when publish a large command


pengyao:
### Description of Issue/Question

If command or target is big, salt occasionally lose the payload in publish_pull socket

### Versions Report
(Provided by running `salt --versions-report`. Please also mention any differences in master/minion versions.)

```
Salt Version:
           Salt: 2017.7.4
 
Dependency Versions:
           cffi: 1.10.0
       cherrypy: Not Installed
       dateutil: Not Installed
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.8.1
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.4.6
   mysql-python: Not Installed
      pycparser: 2.18
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: Not Installed
         Python: 2.7.13 (default, Jul 12 2017, 17:32:34)
   python-gnupg: Not Installed
         PyYAML: 3.11
          PyZMQ: 14.5.0
           RAET: Not Installed
          smmap: Not Installed
        timelib: Not Installed
        Tornado: 4.2.1
            ZMQ: 4.0.5
 
System Versions:
           dist: centos 6.5 Final
         locale: UTF-8
        machine: x86_64
        release: 2.6.32-431.el6.x86_64
         system: Linux
        version: CentOS 6.5 Final
```

pengyao:
https://github.com/saltstack/salt/blob/b5b083fd2655e5082e57d70a03a0f8b850e32ee9/salt/transport/zeromq.py#L854-L856

pub_sock will immediately destroy after payload send, master pull socket not received the payload. I think the reason is: https://github.com/zeromq/libzmq/issues/1922

garethgreenaway:
@pengyao Thanks for the report.  So we have a better change of reproducing this, can you provide more information about the target and command that you're sending along is?

pengyao:
My test case:

* A new salt master, version: 2017.7.4, OS: CentOS 6. installed from salt repo
* Config `zmq_filtering` to `True` for generate large target:
```bash
echo "zmq_filtering: True" >> /etc/salt/master
```
* Add debug statement for publish_pull sock (/usr/lib/python2.7/site-packages/salt/transport/zeromq.py):
```py
# After line 768
[...]

Issue #36469, opened 2016-09-21
Dropped jobs on minion when salt-master has large jobs in the queue.


kingsquirrel152:
### Description of Issue/Question

Sending large payloads the salt-master appears to drop jobs (or never send them).
### Setup

Tested on 2015.8 one master one minion. Apologies for testing against an older rev. I don't have the bandwidth right now to test against latest. My hope is someone out there can run the gist against latest and see if its still an issue. I mainly wanted to get this on the radar to confirm that is not just me. Env is easy enough, one master, one minion with the test + cmd module enabled.
### Steps to Reproduce Issue

Script I use to trigger issue is in the gist link below. It floods a master with small jobs + some jumbo ones. Idea here is tail the minion log. Wait for the all the jobs to go through (ie look for the log to settle). Then press a key to clear the script, it will then auto grep the minion log to see if JID's published are missing from the minion under test.
https://gist.github.com/kingsquirrel152/c9d9fece1d55346e5692de67f990dc57 

You may need to play with the run count and payload size. I suspect that the issue is related to this area of the code https://github.com/saltstack/salt/blob/develop/salt/transport/zeromq.py#L812
My basic theory is that the PUSH socket is being garbage collected since linger is net set on exit of the scope. (For ref: http://stackoverflow.com/questions/6702187/python-zeromq-push-pull-lost-messages) I tried a small refactor that moved the PUSH socket creation out of the publish function and into the init of the class... but I still get dropped messages. 
### Versions Report

``` bash
root@salt_master0_97e605:~# salt --version
salt 2015.8.0 (Beryllium)
root@salt_master0_97e605:~# salt --versions
Salt Version:
           Salt: 2015.8.0

Dependency Versions:
         Jinja2: 2.8
       M2Crypto: Not Installed
           Mako: Not Installed
         PyYAML: 3.12
          PyZMQ: 15.4.0
         Python: 2.7.3 (default, Jun 22 2015, 19:33:41)
           RAET: Not Installed
        Tornado: 4.2.1
            ZMQ: 4.1.5
           cffi: Not Installed
       cherrypy: Not Installed
       dateutil: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
        libnacl: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.4.8
   mysql-python: Not Installed
      pycparser: Not Installed
       pycrypto: 2.6.1
         pygit2: Not Installed
[...]

PR #50424, opened 2018-11-08
Optionally use multiprocessing.Queue for MWorker to publisher

dwoz:
### What does this PR do?

Adds an option to use a `multiprocessing.Queue` instead of ZeroMQ PUSH/PULL sockets for messages relayed from MWorkers to the publisher daemon process.

#36469
#46553
#50393

### Previous Behavior

ZeroMQ's PUSH/PULL sockets are unreliable when dealing with large messages (size) or large number of messages sent quickly.

### New Behavior

Optionally a multiprocessing.Queue for messages moving from MWorkers to the publisher daemon. Messages will get published reliably.

### Tests written?

No - This patch has been tested on via jenkin's branch tests on a fork of 2017.7

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/transport/zeromq.py
+++ b/salt/transport/zeromq.py
@@ -778,32 +778,29 @@ class ZeroMQPubServerChannel(salt.transport.server.PubServerChannel):
         log.info('Starting the Salt Publisher on {0}'.format(pub_uri))
         pub_sock.bind(pub_uri)
 
-        if not msg_queue:
-            # Prepare minion pull socket
-            pull_sock = context.socket(zmq.PULL)
+        # Prepare minion pull socket
+        pull_sock = context.socket(zmq.PULL)
 
-            if self.opts.get('ipc_mode', '') == 'tcp':
-                pull_uri = 'tcp://127.0.0.1:{0}'.format(
-                    self.opts.get('tcp_master_publish_pull', 4514)
-                    )
-            else:
-                pull_uri = 'ipc://{0}'.format(
-                    os.path.join(self.opts['sock_dir'], 'publish_pull.ipc')
-                    )
-            salt.utils.zeromq.check_ipc_path_max_len(pull_uri)
-            # Securely create socket
-            log.info('Starting the Salt Puller on {0}'.format(pull_uri))
-            with salt.utils.files.set_umask(0o177):
-                pull_sock.bind(pull_uri)
-
-
+        if self.opts.get('ipc_mode', '') == 'tcp':
+            pull_uri = 'tcp://127.0.0.1:{0}'.format(
+                self.opts.get('tcp_master_publish_pull', 4514)
+                )
+        else:
+            pull_uri = 'ipc://{0}'.format(
+                os.path.join(self.opts['sock_dir'], 'publish_pull.ipc')
+                )
+        salt.utils.zeromq.check_ipc_path_max_len(pull_uri)
+        # Securely create socket
+        log.info('Starting the Salt Puller on {0}'.format(pull_uri))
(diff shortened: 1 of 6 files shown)


twangboy (approved):

twangboy (approved):

garethgreenaway (approved):

s0undt3ch (approved):

isbm (approved):

salt/transport/zeromq.py
```diff
@@ -905,12 +908,14 @@ def publish(self, load):
             'Sending payload to publish daemon. jid=%s size=%d',
             load.get('jid', None), len(payload),
         )
-        pub_sock.send(payload)
+        if msg_queue:
+            msg_queue.put(payload)
+        else:
+            pub_sock.send(payload)
+            pub_sock.close()
```
rares-pop: should we cache the socket and not create it and close it for each message we publish?

cachedout:
@DmitryKuzmenko What do you think of this?

salt/transport/zeromq.py
dwoz: I did try storing the context and socket on the class so it won't get created each time but I still saw the issue. @kingsquirrel152 also talks about trying that as a fix in #36469

DmitryKuzmenko:
@cachedout I already had a quick lookup of this PR today. Just thought that there is some reason to do this that I don't know. But if there isn't here is my opinion.

ZMQ and Tornado TCP/IPC are always faster than MP Queues. So we need a reasonable argument to use queues instead of ipc/tcp. I'll take a closer look on that related issues tomorrow.

dwoz:
@DmitryKuzmenko I added a muliprocessing.Queue here because I was un-able to make the zeromq PUSH/PULL reliable. In the windows test suite we are seeing tests fail after messages get dropped while moving over the PUSH/PULL [from MWorkers](https://github.com/saltstack/salt/blob/2017.7/salt/transport/zeromq.py#L908) to the [publish daemon](https://github.com/saltstack/salt/blob/2017.7/salt/transport/zeromq.py#L805). I recently added the debug level logs around those calls to more easily spot this when it happens in the test suites.

I tried adjusting the zeromq watermarks. I also moved context and socket to the class level so they only created once, something @rares-pop mentions in his comment. I still saw the dropped messages with these changes. 

After, I moved to a multiprocessing.Queue. I agree that sticking with zmq will be more performant and that it the primary reason I made the Queue optional. I also think it would still be nice if we can make zmq rock solid. I'm re-testing the class level context just to be on the safe side that I didn't miss anything before. The only other thing I can think of is to replace PUSH/PULL with PUB/SUB in hopes of that being more reliable. It seems like that would still work if we only have one subscriber (subscribed to everything) with multiple publisher.

I am looking forward to hearing more of your thoughts on this if you have them. Thanks for looking.

salt/transport/zeromq.py
kingsquirrel152: I fixed this issue in an older version of salt stack (back when it was single threaded and not driven by tornado). The issue iirc is that creating the socket, and sending data, then _tearing the socket down_, breaks the contract with zmq. You have no guarantee the message is ever sent. I think there are some settings for the socket that could help (linger for instance). TLDR though - I think your best bet is to follow @rares-pop suggest and cache the socket avoiding the issue entirely.

dwoz:
I've opened a new PR (#50463) which I think resolves these issues without using `multiprocessing.Queue` and has some tests. I'm going to close this one since the other seems to be reliable. Thanks for everyone's input.

PR #50428, opened 2018-11-08
Port json_query Jinja filter from Ansible

max-arnold:
### What does this PR do?

This is a port of Ansible `json_query` Jinja filter to make complex queries against json data structures. Query language http://jmespath.org/ parser depends on `jmespath` python library (optional).

It could be used to filter `pillar` data, `yaml` maps, and in combination with`http_query`. Can replace lots of ugly Jinja loops and in some cases help to avoid writing trivial custom modules https://docs.saltstack.com/en/latest/topics/tutorials/jinja_to_execution_module.html

Jinja template:

```
  Example 1: {{ [1, 2, 3, 4, [5, 6]] | json_query('[]') }}

  Example 2: {{
  {"machines": [
    {"name": "a", "state": "running"},
    {"name": "b", "state": "stopped"},
    {"name": "b", "state": "running"}
  ]} | json_query("machines[?state=='running'].name") }}

  Example 3: {{
  {"services": [
    {"name": "http", "host": "1.2.3.4", "port": 80},
    {"name": "smtp", "host": "1.2.3.5", "port": 25},
    {"name": "ssh",  "host": "1.2.3.6", "port": 22},
  ]} | json_query("services[].port") }}
```

Rendered output:

```
  Example 1: [1, 2, 3, 4, 5, 6]

  Example 2: ['a', 'b']

  Example 3: [80, 25, 22]
```

### Tests written?

Yes

### Commits signed with GPG?

No

P.S. The feature is not intrusive, how about including it into the Fluorine branch?

### DIFF
--- a/salt/utils/data.py
+++ b/salt/utils/data.py
@@ -29,6 +29,11 @@ from salt.utils.decorators.jinja import jinja_filter
 from salt.ext import six
 from salt.ext.six.moves import range  # pylint: disable=redefined-builtin
 
+try:
+    import jmespath
+except ImportError:
+    jmespath = None
+
 log = logging.getLogger(__name__)
 
 
@@ -894,3 +899,13 @@ def stringify(data):
             item = six.text_type(item)
         ret.append(item)
     return ret
+
+
+@jinja_filter('json_query')
+def json_query(data, expr):
+    '''
+    Query data using JMESPath query language (http://jmespath.org).
+    '''
+    if jmespath is None:
+        raise RuntimeError('json_query filter requires jmespath library which is not installed')
+    return jmespath.search(expr, data)
--- a/salt/utils/data.py
+++ b/salt/utils/data.py
@@ -904,8 +904,10 @@ def stringify(data):
 @jinja_filter('json_query')
 def json_query(data, expr):
     '''
-    Query data using JMESPath query language (http://jmespath.org).
+    Query data using JMESPath language (http://jmespath.org).
     '''
     if jmespath is None:
-        raise RuntimeError('json_query filter requires jmespath library which is not installed')
+        err = 'json_query requires jmespath module installed'
+        log.error(err)
+        raise RuntimeError(err)
     return jmespath.search(expr, data)
--- a/doc/topics/jinja/index.rst
+++ b/doc/topics/jinja/index.rst
@@ -1255,8 +1255,52 @@ Returns:
 
   'default'
 
+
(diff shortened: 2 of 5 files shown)


dwoz (approved):

salt/utils/data.py
```diff
@@ -894,3 +899,13 @@ def stringify(data):
...
     return ret
+
+
+@jinja_filter('json_query')
+def json_query(data, expr):
+    '''
+    Query data using JMESPath query language (http://jmespath.org).
+    '''
+    if jmespath is None:
+        raise RuntimeError('json_query filter requires jmespath library which is not installed')
```
isbm: Please remove "which is not", this is redundant.

salt/utils/data.py
```diff
@@ -894,3 +899,13 @@ def stringify(data):
...
         ret.append(item)
     return ret
+
+
+@jinja_filter('json_query')
+def json_query(data, expr):
+    '''
+    Query data using JMESPath query language (http://jmespath.org).
+    '''
+    if jmespath is None:
```
isbm: Please add `log.error` to record this.

isbm (changes requested):

salt/utils/data.py
max-arnold: Done

salt/utils/data.py
```diff
@@ -894,3 +899,15 @@ def stringify(data):
...
     return ret
+
+
+@jinja_filter('json_query')
+def json_query(data, expr):
+    '''
+    Query data using JMESPath language (http://jmespath.org).
+    '''
+    if jmespath is None:
+        err = 'json_query filter requires installed jmespath module'
```
isbm: "json_query requires jmespath module installed". No?

isbm (changes requested):

salt/utils/data.py
max-arnold: Fixed :)

isbm (approved):

max-arnold:
Are jinja filters accessible in other salt modules as simple functions?

cachedout (approved):
I really like this. Excellent idea.

mchugh19:
Great! Probably needs a mention in the release notes as well.

max-arnold:
@mchugh19 Oops, I completely forgot about that. Thanks for the reminder!  Fixed in #53866

PR #50451, opened 2018-11-09
Introduce a new way of specifying uri + attempt to use oratab

jose1711:
### What does this PR do?

Specifying uri for all databases may be cumbersome, same goes with
maintaining pillars where some might see security concerns having
passwords stored in a central location.

This PR introduces a new uri format spec in form:
 sid[ as {sysdba|sysoper}]

As a result ORACLE_HOME and ORACLE_SID env variables are set based on
the contents of oratab. Due to the fact that this file on exists on
unix-like minions, Windows minions do not support this syntax.

Another change is that presence of oracle:dbs pillar is no longer
required when executing a query. In this case SID is set to the name
of the database argument and /sysdba login is used to run the query.

### Tests written?

No

### Commits signed with GPG?

No

### DIFF
--- a/salt/modules/oracle.py
+++ b/salt/modules/oracle.py
@@ -28,7 +28,10 @@ Oracle DataBase connection module
             <db>:
               uri: connection credentials in format:
             user/password@host[:port]/sid[ servicename as {sysdba|sysoper}]
-              optional keyword servicename will determine whether it is a sid or service_name
+              or
+            sid[ as {sysdba|sysoper}]
+             (ORACLE_HOME is parsed from oratab)
+            optional keyword servicename will determine whether it is a sid or service_name
             <db>:
               uri: .....
 '''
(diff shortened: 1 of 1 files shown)


salt/modules/oracle.py
```diff
@@ -37,6 +40,11 @@
 import os
 import logging
 from salt.utils.decorators import depends
+try:
+    from salt.utils.files import fopen
+except ImportError:
+    from salt.utils import fopen
```
cachedout: Why would this be necessary, considering this is going into the develop branch of Salt?

cachedout:
@jose1711 Did you see the review comment on this?

salt/modules/oracle.py
jose1711: hey mike, you're absolutely right. please let me set up an aix minion with salt from dev branch so that i can make a real test of the proposed changes. until now i've based this on stable minion 2016-xx. i will have the next update til friday 23rd. also.. apart from the import is there anything else that needs to be looked at?

cachedout: This was my only concern. Thanks, @jose1711

jose1711: fixed and tested on a real db system with develop branch

PR #50452, opened 2018-11-09
Add new module freezer

aplanas:
### What does this PR do?

### Tests written?

Yes

### DIFF
--- a/salt/modules/freezer.py
+++ b/salt/modules/freezer.py
@@ -1,0 +1,294 @@
+# -*- coding: utf-8 -*-
+#
+# Author: Alberto Planas <aplanas@suse.com>
+#
+# Copyright 2018 SUSE LINUX GmbH, Nuernberg, Germany.
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+'''
+:maintainer:    Alberto Planas <aplanas@suse.com>
+:maturity:      new
+:depends:       None
+:platform:      Linux
+'''
+from __future__ import absolute_import, print_function, unicode_literals
+import logging
+import os
+
+from salt.exceptions import CommandExecutionError
+from salt.utils.args import clean_kwargs
+from salt.utils.files import fopen
+import salt.utils.json as json
+from salt.ext.six.moves import zip
+
+log = logging.getLogger(__name__)
+
+__func_alias__ = {
+    'list_': 'list',
+}
+
+
+def __virtual__():
+    '''
+    Freezer is based on top of the pkg module.
+
+    Return True as pkg is going to be there, so we can avoid of
+    loading all modules.
+
+    '''
+    return True
+
+
+def _states_path():
+    '''
+    Return the path where we will store the states.
+    '''
+    return os.path.join(__opts__['cachedir'], 'freezer')
+
+
+def _paths(name=None):
+    '''
+    Return the full path for the packages and repository freezer
+    files.
+
+    '''
+    name = 'freezer' if not name else name
+    states_path = _states_path()
+    return (
+        os.path.join(states_path, '{}-pkgs.yml'.format(name)),
+        os.path.join(states_path, '{}-reps.yml'.format(name)),
+    )
+
+
+def status(name=None):
+    '''
+    Return True if there is already a frozen state.
+
+    A frozen state is merely a list of packages (including the
+    version) in a specific time. This information can be used to
+    compare with the current list of packages, and revert the
+    installation of some extra packages that are in the system.
+
+    name
+        Name of the frozen state. Optional.
+
+    CLI Example:
+
+    .. code-block:: bash
+
+        salt '*' freezer.status
+        salt '*' freezer.status pre_install
+
+    '''
+    name = 'freezer' if not name else name
+    return all(os.path.isfile(i) for i in _paths(name))
+
+
+def list_():
+    '''
+    Return the list of frozen states.
+
+    CLI Example:
+
+    .. code-block:: bash
+
+        salt '*' freezer.list
+
+    '''
+    ret = []
+    states_path = _states_path()
+    if not os.path.isdir(states_path):
+        return ret
+
+    for state in os.listdir(states_path):
+        if state.endswith(('-pkgs.yml', '-reps.yml')):
+            # Remove the suffix, as both share the same size
+            ret.append(state[:-9])
+    return sorted(set(ret))
+
+
+def freeze(name=None, force=False, **kwargs):
+    '''
+    Save the list of package and repos in a freeze file.
+
+    As this module is build on top of the pkg module, the user can
+    send extra attributes to the underlying pkg module via kwargs.
+    This function will call ``pkg.list_pkgs`` and ``pkg.list_repos``,
+    and any additional arguments will be passed through to those
+    functions.
+
+    name
+        Name of the frozen state. Optional.
+
+    force
+        If true, overwrite the state. Optional.
+
+    CLI Example:
+
+    .. code-block:: bash
+
+        salt '*' freezer.freeze
+        salt '*' freezer.freeze pre_install
+        salt '*' freezer.freeze force=True root=/chroot
+
+    '''
+    states_path = _states_path()
+
+    try:
+        os.makedirs(states_path)
+    except OSError as e:
+        msg = 'Error when trying to create the freezer storage %s: %s'
+        log.error(msg, states_path, e)
+        raise CommandExecutionError(msg % (states_path, e))
+
+    if status(name) and not force:
+        raise CommandExecutionError('The state is already present. Use '
+                                    'force parameter to overwrite.')
+    safe_kwargs = clean_kwargs(**kwargs)
+    pkgs = __salt__['pkg.list_pkgs'](**safe_kwargs)
+    repos = __salt__['pkg.list_repos'](**safe_kwargs)
+    for name, content in zip(_paths(name), (pkgs, repos)):
+        with fopen(name, 'w') as fp:
+            json.dump(content, fp)
+    return True
+
+
+def restore(name=None, **kwargs):
+    '''
+    Make sure that the system contains the packages and repos from a
+    frozen state.
+
+    Read the list of packages and repositories from the freeze file,
+    and compare it with the current list of packages and repos. If
+    there is any difference, all the missing packages are repos will
+    be installed, and all the extra packages and repos will be
+    removed.
+
+    As this module is build on top of the pkg module, the user can
+    send extra attributes to the underlying pkg module via kwargs.
+    This function will call ``pkg.list_repos``, ``pkg.mod_repo``,
+    ``pkg.list_pkgs``, ``pkg.install``, ``pkg.remove`` and
+    ``pkg.del_repo``, and any additional arguments will be passed
+    through to those functions.
+
+    name
+        Name of the frozen state. Optional.
+
+    CLI Example:
+
+    .. code-block:: bash
+
+        salt '*' freezer.restore
+        salt '*' freezer.restore root=/chroot
+
+    '''
+    if not status(name):
+        raise CommandExecutionError('Frozen state not found.')
+
+    frozen_pkgs = {}
+    frozen_repos = {}
+    for name, content in zip(_paths(name), (frozen_pkgs, frozen_repos)):
+        with fopen(name) as fp:
+            content.update(json.load(fp))
+
+    # The ordering of removing or adding packages and repos can be
+    # relevant, as maybe some missing package comes from a repo that
+    # is also missing, so it cannot be installed. But can also happend
+    # that a missing package comes from a repo that is present, but
+    # will be removed.
+    #
+    # So the proposed order is;
+    #   - Add missing repos
+    #   - Add missing packages
+    #   - Remove extra packages
+    #   - Remove extra repos
+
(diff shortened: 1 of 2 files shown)


aplanas:
This little module is ready for review. I am not sure if the utility is general enough, but is covering my use cases.

salt/modules/freezer.py
```diff
@@ -0,0 +1,281 @@
...
+__func_alias__ = {
+    'list_': 'list',
+}
+
+
+def __virtual__():
+    '''
+    Freezer is based on top of the pkg module.
+    '''
+    return 'pkg.list_pkgs' in __salt__
```
cachedout: This ends up being expensive because the iteration has the side effect of loading _all_ modules. I can't think of a situation where this wouldn't already be the case. Could we just return True here?

salt/modules/freezer.py
```diff
@@ -0,0 +1,281 @@
...
+    safe_kwargs = {k: v for k, v in kwargs.items() if not k.startswith('__')}
+    pkgs = __salt__['pkg.list_pkgs'](**safe_kwargs)
+    repos = __salt__['pkg.list_repos'](**safe_kwargs)
+    for name, content in zip(_paths(name), (pkgs, repos)):
+        with fopen(name, 'w') as fp:
+            json.dump(content, fp)
+    return True
+
+
+def restore(name=None, **kwargs):
```
cachedout: We dislike allowing kwargs because a user can't know what arguments are possible. Can we either make this explicit or remove this entirely?

salt/modules/freezer.py
aplanas: Sure!

salt/modules/freezer.py
aplanas: I see. I have a similar comment in the `chroot` PR (https://github.com/saltstack/salt/pull/50418#issuecomment-437057166). In this case is more clear why `kwargs` is needed, as the user can send any parameter to any of the hundreds of execution modules that can be executed inside the chroot jail.

But in this case is less clear. I agree. The issue here is again _composability_. I need to compose one module, freezer, with several others that depends on. In this case `pkg.list_repos`, `pkg.mod_repo`, `pkg.list_pkgs`, `pkg.remove`, `pkg.install` and `pkg.del_repo`. The user can influence how is calling those modules via the ugly `kwargs`, for example to send the `root` parameter (my personal use case), the `refresh` to force a refresh of the repository information, or many others.

I can simply close the parameters to my single and personal use case (passing root and maybe one or two when in a week I discover that I need it), or I can do a closure of the parameters that are affected for those pkg.* modules that I use (including the code for yum, aptget, zypper, dnf, pacman and windows), and make it explicit in the signature. But this will means to have more than 15 parameters easily.

If the main problem is the user, I added in the documentation a comment about how the user can affect the underlining pkg.* modules via kwargs, but I can try to make the comment more explicit.

aplanas: I added the extra comments.

isbm: @cachedout not sure this is quite right, we actually use `kwargs` everywhere, and we have [utility package for that](https://github.com/saltstack/salt/blob/develop/salt/utils/args.py), using it for example [here](https://github.com/saltstack/salt/blob/develop/salt/minion.py#L359-L360) or [here](https://github.com/saltstack/salt/blob/develop/salt/modules/libcloud_compute.py#L110) or [here](https://github.com/saltstack/salt/blob/develop/salt/modules/aptpkg.py#L2776) or [here](https://github.com/saltstack/salt/blob/develop/salt/modules/git.py#L3909) etc for the same reasons.

What is important is to just document that properly and raise an exception if there are unknown args, according to the examples.

salt/modules/freezer.py
```diff
@@ -0,0 +1,291 @@
...
+    CLI Example:
+
+    .. code-block:: bash
+
+        salt '*' freezer.freeze
+        salt '*' freezer.freeze force=True root=/chroot
+
+    '''
+    states_path = _states_path()
+    ret = __salt__['cmd.retcode'](['mkdir', '-p', states_path])
```
isbm: Why to spin the whole process for just create nested directories?

```python
import os
...
try:
    os.makedirs(states_path)
except OSError as err:
    log.error(err)
```
[More info about `os.makedirs`](https://docs.python.org/3/library/os.html#os.makedirs).

salt/modules/freezer.py
```diff
@@ -0,0 +1,291 @@
...
+    if ret:
+        msg = 'Error when trying to create the freezer ' \
+            'storage: {}'.format(states_path)
```
isbm: We don't have IBM punchcard limit on the line width, so feel free to make it readable, up to ~120 chars easily. :wink:

salt/modules/freezer.py
isbm: As I commented before, we actually using kwargs all over the place, and sometimes we even have to (e.g. when "async" is a parameter, but it is reserved word in Python 3.7). Still we would require:

- Proper documentation
- Validation of them with mentioned utilities

salt/modules/freezer.py
```diff
@@ -0,0 +1,291 @@
...
+    # that a missing package comes from a repo that is present, but
+    # will be removed.
+    #
+    # So the proposed order is;
+    #   - Add missing repos
+    #   - Add missing packages
+    #   - Remove extra packages
+    #   - Remove extra repos
+
+    safe_kwargs = {k: v for k, v in kwargs.items() if not k.startswith('__')}
```
isbm: We have [utility](https://github.com/saltstack/salt/blob/develop/salt/utils/args.py#L30) exactly for that. Please use it instead.

salt/modules/freezer.py
```diff
@@ -0,0 +1,291 @@
...
+
+    # Add missing repositories
+    repos = __salt__['pkg.list_repos'](**safe_kwargs)
+    missing_repos = set(frozen_repos) - set(repos)
+    for repo in missing_repos:
+        try:
+            # In Python 2 we cannot do advance destructuring, so we
+            # need to create a temporal dictionary that will merge all
+            # the parameters
+            _tmp_kwargs = frozen_repos[repo].copy()
```
isbm: It is only a shallow copy.

salt/modules/freezer.py
```diff
@@ -0,0 +1,291 @@
...
+            _tmp_kwargs.update(safe_kwargs)
+            __salt__['pkg.mod_repo'](repo, **_tmp_kwargs)
```
isbm: No checks here?

salt/modules/freezer.py
```diff
@@ -0,0 +1,291 @@
...
+            res['repos']['add'].append(repo)
+            log.info('Added missing repository {}'.format(repo))
+        except Exception as e:
+            msg = 'Error adding {} repository: {}'.format(repo, e)
+            log.error(msg)
+            res['comment'].append(msg)
+
+    # Add missing packages
+    # NOTE(aplanas) we can remove the `for` using `pkgs`. This will
```
isbm: I guess, git log will tell who wrote this comment.

tests/unit/modules/test_freezer.py
```diff
@@ -0,0 +1,239 @@
...
+            'freezer-pkgs.yml', 'freezer-reps.yml',
+            'state-pkgs.yml', 'state-reps.yml',
+            'random-file'
+        ]
+        self.assertEqual(freezer.list_(), ['freezer', 'state'])
+
+    @patch('salt.utils.json.dump')
+    @patch('salt.modules.freezer.fopen')
+    @patch('salt.modules.freezer.isfrozen')
+    def test_freeze(self, isfrozen, fopen, dump):
```
isbm: This would make sense to break this UT into at least four as I see it. UT should be kept as small as possible and as focused as possible to check only one thing.

tests/unit/modules/test_freezer.py
```diff
@@ -0,0 +1,239 @@
...
+        '''
+        Test to freeze a current installation
+        '''
+
+        # Fails when creating the freeze cache directory
+        salt_mock = {
+            'cmd.retcode': MagicMock(return_value=1),
+        }
+        with patch.dict(freezer.__salt__, salt_mock):
```
isbm: Breaking this UT into more, you can also get rid of this `with ...` thing, making overall test easier to comprehend.

tests/unit/modules/test_freezer.py
```diff
@@ -0,0 +1,239 @@
...
+            salt_mock['cmd.retcode'].assert_called_once()
+            salt_mock['pkg.list_pkgs'].assert_called_once()
+            salt_mock['pkg.list_repos'].assert_called_once()
+            fopen.assert_called()
+            dump.asster_called()
+
+    @patch('salt.utils.json.load')
+    @patch('salt.modules.freezer.fopen')
+    @patch('salt.modules.freezer.isfrozen')
+    def test_restore(self, isfrozen, fopen, load):
```
isbm: This UT should be also broken into couple of those.

isbm (changes requested):

salt/modules/freezer.py
aplanas: yes, but this is what I need if I am not wrong. I do not want to change frozen_repos[repo] (the container that I copy and later I do the update), and I will not touch the content (that is not cloned)

salt/modules/freezer.py
aplanas: indeed

salt/modules/freezer.py
aplanas: Uhm, the kind of fails that matters are throwing `CommandExecutionError` or other kind of exception ([here](https://github.com/saltstack/salt/blob/develop/salt/modules/zypper.py#L1039), [or here](https://github.com/saltstack/salt/blob/develop/salt/modules/zypper.py#L1045), ...). That is why this is inside a `try`, and report to the user and via log.error in the `except` part.

salt/modules/freezer.py
aplanas: Yeah, is an old habit from a different project. Removed.

tests/unit/modules/test_freezer.py
aplanas: Sure.

tests/unit/modules/test_freezer.py
aplanas: done

salt/modules/freezer.py
terminalmage: I understand @cachedout's concerns here, but I think we just need this to be better documented. We need to tell the user what underlying functions are being called and refer the user to use documentation to determine which of them to use.

salt/modules/freezer.py
```diff
@@ -0,0 +1,294 @@
...
+    send extra attributes to the underlining pkg module via kwargs.
+
+    name
+        Name of the frozen state. Optional.
+
+    force
+        If true, overwrite the state. Optional.
+
+    kwargs
+        Extra parameters send to the underlining pkg module. Optional.
```
terminalmage: We shouldn't be documenting an argument that isn't used on the CLI. As mentioned elsewhere, we should document how other functions will be called, and let the user know that this function will pass through certain arguments. They can use the documentation from their `pkg` module to know what arguments are supported.

terminalmage (changes requested):

salt/modules/freezer.py
```diff
@@ -0,0 +1,291 @@
...
+            ret.append(state[:-9])
+    return sorted(set(ret))
+
+
+def freeze(name=None, force=False, **kwargs):
+    '''
+    Save the list of package and repos in a freeze file.
+
+    As this module is build on top of the pkg module, the user can
+    send extra attributes to the underlining pkg module via kwargs.
```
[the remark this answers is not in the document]
aplanas: ouch

salt/modules/freezer.py
```diff
@@ -0,0 +1,291 @@
...
+        'comment': [],
+    }
+
+    # Add missing repositories
+    repos = __salt__['pkg.list_repos'](**safe_kwargs)
+    missing_repos = set(frozen_repos) - set(repos)
+    for repo in missing_repos:
+        try:
+            # In Python 2 we cannot do advance destructuring, so we
+            # need to create a temporal dictionary that will merge all
```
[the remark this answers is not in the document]
aplanas: done, thanks

salt/modules/freezer.py
```diff
@@ -0,0 +1,291 @@
...
+    # NOTE(aplanas) we can remove the `for` using `pkgs`. This will
+    # improve performance, but I want to have a more detalied report
+    # of what packages are installed or failled.
+    pkgs = __salt__['pkg.list_pkgs'](**safe_kwargs)
+    missing_pkgs = set(frozen_pkgs) - set(pkgs)
+    for pkg in missing_pkgs:
+        try:
+            __salt__['pkg.install'](name=pkg, **safe_kwargs)
+            res['pkgs']['add'].append(pkg)
+            log.info('Added missing package {}'.format(pkg))
```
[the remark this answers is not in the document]
aplanas: done

salt/modules/freezer.py
aplanas: I think that the new paragraph point this.

salt/modules/freezer.py
```diff
@@ -0,0 +1,290 @@
...
+    return sorted(set(ret))
+
+
+def freeze(name=None, force=False, **kwargs):
+    '''
+    Save the list of package and repos in a freeze file.
+
+    As this module is build on top of the pkg module, the user can
+    send extra attributes to the underlying pkg module via kwargs. For
+    this command we can influence pgk.list_pkgs and pkg.list_repos.
```
terminalmage: `pgk.list_pkgs` should be `pkg.list_pkgs`. Also, the wording in the last sentence doesn't seem right. We should instead say something like:

```
This function will call ``pkg.list_pkgs`` and ``pkg.list_repos``, and any
additional arguments will be passed through to those functions.
```

salt/modules/freezer.py
```diff
@@ -0,0 +1,290 @@
...
+
+    Read the list of packages and repositories from the freeze file,
+    and compare it with the current list of packages and repos. If
+    there is any difference, all the missing packages are repos will
+    be installed, and all the extra packages and repos will be
+    removed.
+
+    As this module is build on top of the pkg module, the user can
+    send extra attributes to the underlying pkg module via kwargs. For
+    this command we can influence pkg.list_repos, pkg.mod_repo,
```
terminalmage: Again, the wording in this sentence is not quote right. Instead, let's use the example I suggested above.

terminalmage (changes requested):

salt/modules/freezer.py
aplanas: thanks, done!

terminalmage (approved):

PR #50453, opened 2018-11-09
Debian info_installed compatibility

isbm:
### What does this PR do?

Brings `info_installed` function compatible with the RPM equivalent

### Tests written?

Yes

### DIFF
--- a/salt/modules/dpkg.py
+++ b/salt/modules/dpkg.py
@@ -447,6 +447,15 @@ def info(*packages, **kwargs):
 
         .. versionadded:: 2016.11.3
 
+    attr
+        Comma-separated package attributes. If no 'attr' is specified, all available attributes returned.
+
+        Valid attributes are:
+            version, vendor, release, build_date, build_date_time_t, install_date, install_date_time_t,
+            build_host, group, source_rpm, arch, epoch, size, license, signature, packager, url, summary, description.
+
+        .. versionadded:: 2018.11
+
     CLI example:
 
     .. code-block:: bash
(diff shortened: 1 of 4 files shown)


salt/modules/dpkg.py
```diff
@@ -274,7 +306,7 @@ def _get_pkg_info(*packages, **kwargs):
     ret = []
     cmd = "dpkg-query -W -f='package:" + bin_var + "\\n" \
           "revision:${binary:Revision}\\n" \
-          "architecture:${Architecture}\\n" \
+          "arch:${Architecture}\\n" \
```
meaksh: IIRC this should not be changed to `arch`, otherwise we won't get the right architecture value here: https://github.com/saltstack/salt/blob/576e8717501e1cec0e2dd968526eb14ec4dc5f0e/salt/modules/dpkg.py#L342 

Am I missing something?

meaksh (commented):
I have a doubt about how this code handles `arch` attribute from `dpkg-query` output.

salt/modules/dpkg.py
isbm: @meaksh sure you missing something:

```
"meaksh:${Architecture}\\n"
```
This is also valid. :wink: That said:
```
$ salt-call --local pkg.info_installed vim attr=arch,build_date,install_date,version
local:
    ----------
    vim:
        ----------
        arch:
            amd64
        build_date:
            2016-04-03T13:25:38Z
        install_date:
            2017-11-02T09:55:37Z
        version:
            2:7.4.1689-3ubuntu1.2
```
Looks OK?

salt/modules/dpkg.py
```diff
@@ -307,9 +339,14 @@ def _get_pkg_info(*packages, **kwargs):
             key, value = pkg_info_line.split(":", 1)
             if value:
                 pkg_data[key] = value
-        install_date = _get_pkg_install_time(pkg_data.get('package'), pkg_data.get('architecture'))
+        install_date, install_date_t = _get_pkg_install_time(pkg_data.get('package'), pkg_data.get('architecture'))
```
isbm: @meaksh ah, here. Right.

meaksh (approved):
Looks good now! Thanks :+1:

isbm:
@cachedout @rallytime @dwoz  bump!

PR #50460, opened 2018-11-10
Add support to highstate_return.py for user authentication

cdalvaro:
### What does this PR do?

This PR adds support for sending emails with `highstate` return out of the same mail server by providing user authentication:

- `smtp_username`: Empty by default
- `smtp_password`: Empty by default

Additionally, more parameters for server configuration have been added:

- `smtp_port`: Default to `0` (same as before)
- `smtp_tls`: Default to `False` (same as before)

Previously this returner was unable to send messages outside the outgoing mail server.

### New Behavior

By providing user authentication parameters you can now send emails outside the outgoing mail server.

### Tests written?

No

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/returners/highstate_return.py
+++ b/salt/returners/highstate_return.py
@@ -44,6 +44,10 @@ Here is an example of what the configuration might look like:
       smtp_success_subject: 'success minion {id} on host {host}'
       smtp_failure_subject: 'failure minion {id} on host {host}'
       smtp_server: smtp.example.com
+      smtp_port: 25
+      smtp_tls: False
+      smtp_username: username
+      smtp_password: password
       smtp_recipients: saltusers@example.com, devops@example.com
       smtp_sender: salt@example.com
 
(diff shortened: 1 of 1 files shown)


dwoz (approved):

salt/returners/highstate_return.py
```diff
@@ -448,6 +457,12 @@ def _produce_output(report, failed, setup):
         sender = setup.get('smtp_sender', '')
         recipients = setup.get('smtp_recipients', '')
 
+        host = setup.get('smtp_server', '')
+        port = int(setup.get('smtp_port', 0))
```
cachedout: Default to 25 perhaps?

cdalvaro: That was my first thought, however, since the port value wasn't set before and the default value in the `smtplib` is `0` (see [official documentation](https://docs.python.org/3.6/library/smtplib.html#module-smtplib)), I thought that for backward compatibility it was better to set the same vale as the default one.

I was really tempted for setting the value to `25` too.

Maybe this could be discussed.

cachedout: I don't think we have backward compat issues here since we're introducing a new feature and in other places in salt we default to 25. IMHO, that is what we should do here.

cachedout (approved):

cdalvaro:
Hi @dwoz, when will be this change packaged into a new release?

PR #50463, opened 2018-11-11
More reliable pub server publishing

dwoz:
- Uses existing zmq context
- Store socket in thread local storage for re-use
- Add tests to verify publish reliability

#36469
#46553
#50393

### Tests written?

Yes

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/transport/zeromq.py
+++ b/salt/transport/zeromq.py
@@ -13,6 +13,7 @@ import signal
 import hashlib
 import logging
 import weakref
+import threading
 from random import randint
 
 # Import Salt Libs
@@ -738,6 +739,9 @@ class ZeroMQPubServerChannel(salt.transport.server.PubServerChannel):
     '''
     Encapsulate synchronous operations for a publisher channel
     '''
+
+    _sock_data = threading.local()
+
     def __init__(self, opts):
         self.opts = opts
         self.serial = salt.payload.Serial(self.opts)  # TODO: in init?
@@ -773,9 +777,11 @@ class ZeroMQPubServerChannel(salt.transport.server.PubServerChannel):
             # IPv6 sockets work for both IPv6 and IPv4 addresses
             pub_sock.setsockopt(zmq.IPV4ONLY, 0)
         pub_sock.setsockopt(zmq.BACKLOG, self.opts.get('zmq_backlog', 1000))
+        pub_sock.setsockopt(zmq.LINGER, -1)
         pub_uri = 'tcp://{interface}:{publish_port}'.format(**self.opts)
         # Prepare minion pull socket
         pull_sock = context.socket(zmq.PULL)
+        pull_sock.setsockopt(zmq.LINGER, -1)
 
         if self.opts.get('ipc_mode', '') == 'tcp':
             pull_uri = 'tcp://127.0.0.1:{0}'.format(
@@ -838,15 +844,14 @@ class ZeroMQPubServerChannel(salt.transport.server.PubServerChannel):
                     raise exc
 
         except KeyboardInterrupt:
-            # Cleanly close the sockets if we're shutting down
-            if pub_sock.closed is False:
-                pub_sock.setsockopt(zmq.LINGER, 1)
-                pub_sock.close()
-            if pull_sock.closed is False:
-                pull_sock.setsockopt(zmq.LINGER, 1)
-                pull_sock.close()
-            if context.closed is False:
-                context.term()
+            log.trace('Publish daemon caught Keyboard interupt, tearing down')
+        # Cleanly close the sockets if we're shutting down
+        if pub_sock.closed is False:
+            pub_sock.close()
+        if pull_sock.closed is False:
+            pull_sock.close()
+        if context.closed is False:
+            context.term()
 
     def pre_fork(self, process_manager, kwargs=None):
         '''
@@ -858,23 +863,29 @@ class ZeroMQPubServerChannel(salt.transport.server.PubServerChannel):
         '''
         process_manager.add_process(self._publish_daemon, kwargs=kwargs)
 
-    def publish(self, load):
+    @property
+    def pub_sock(self):
         '''
-        Publish "load" to minions
-
-        :param dict load: A load to be sent across the wire to minions
+        This thread's zmq publisher socket. This socket is stored on the class
+        so that multiple instantiations in the same thread will re-use a single
+        zmq socket.
         '''
-        payload = {'enc': 'aes'}
+        try:
+            return self._sock_data.sock
+        except AttributeError:
+            pass
 
-        crypticle = salt.crypt.Crypticle(self.opts, salt.master.SMaster.secrets['aes']['secret'].value)
-        payload['load'] = crypticle.dumps(load)
-        if self.opts['sign_pub_messages']:
-            master_pem_path = os.path.join(self.opts['pki_dir'], 'master.pem')
-            log.debug("Signing data packet")
-            payload['sig'] = salt.crypt.sign_message(master_pem_path, payload['load'])
-        # Send 0MQ to the publisher
-        context = zmq.Context(1)
-        pub_sock = context.socket(zmq.PUSH)
+    def pub_connect(self):
+        '''
+        Create and connect this thread's zmq socket. If a publisher socket
+        already exists "pub_close" is called before creating and connecting a
+        new socket.
+        '''
+        if self.pub_sock:
+            self.pub_close()
+        ctx = zmq.Context.instance()
+        self._sock_data.sock = ctx.socket(zmq.PUSH)
+        self.pub_sock.setsockopt(zmq.LINGER, -1)
         if self.opts.get('ipc_mode', '') == 'tcp':
             pull_uri = 'tcp://127.0.0.1:{0}'.format(
                 self.opts.get('tcp_master_publish_pull', 4514)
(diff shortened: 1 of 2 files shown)


dwoz:
This PR replaces the work done in #50424

s0undt3ch (approved):

salt/transport/zeromq.py
```diff
@@ -838,15 +844,14 @@ def _publish_daemon(self, log_queue=None):
...
-            if context.closed is False:
-                context.term()
+            log.trace('Publish daemon caught Keyboard interupt, tearing down')
+        # Cleanly close the sockets if we're shutting down
+        if pub_sock.closed is False:
+            pub_sock.close()
+        if pull_sock.closed is False:
+            pull_sock.close()
+        if context.closed is False:
+            context.term()
```
twangboy: Does the above need to be indented... part of the KeyboardInterrupt error?

dwoz: Instead of only cleaning up the sockets and context on a keyboard error. I moved this so it will always run when the daemon process is ending.

twangboy: Sweet

twangboy (approved):

DmitryKuzmenko (approved):
:+1: :1st_place_medal:

rallytime (approved):

cachedout:
@skizunov @rares-pop or @jacksontj Do any of you have interest in looking at this?

rares-pop (approved):
@cachedout - yeah, it looks good! It's nice we are caching the socket.

salt/transport/zeromq.py
```diff
@@ -773,9 +777,11 @@ def _publish_daemon(self, log_queue=None):
             # IPv6 sockets work for both IPv6 and IPv4 addresses
             pub_sock.setsockopt(zmq.IPV4ONLY, 0)
         pub_sock.setsockopt(zmq.BACKLOG, self.opts.get('zmq_backlog', 1000))
+        pub_sock.setsockopt(zmq.LINGER, -1)
```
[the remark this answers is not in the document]
dwoz: The default is `LINGER == -1`, the change is just to show that's what it is explicitly.

salt/transport/zeromq.py
```diff
@@ -13,6 +13,7 @@
 import hashlib
 import logging
 import weakref
+import threading
```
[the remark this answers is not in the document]
dwoz: We're not creating any threads here. This import is for thread local storage `threading.local`. Since the zmq socket is not thread save, we're caching it in thread local storage rather than just on the `ZeroMQPubServerChannel`. It's safer than storing it on class and facilitates testing (where we use threads). Under normal operation, there will only be one socket per MWorker because the publish method is called from the same thread every time.

As for async, I think we should move this to zmq's async. I figured that should happen in develop after this get's merged up.

salt/transport/zeromq.py
```diff
@@ -858,32 +863,63 @@ def pre_fork(self, process_manager, kwargs=None):
...
+        except AttributeError:
+            pass
+
+    def pub_connect(self):
+        '''
+        Create and connect this thread's zmq socket. If a publisher socket
+        already exists "pub_close" is called before creating and connecting a
+        new socket.
+        '''
+        if self.pub_sock:
```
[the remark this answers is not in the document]
dwoz: I've added a log with the uri it is connecting to. I made it debug since it should only happen once per MWorker.

salt/transport/zeromq.py
```diff
@@ -838,15 +844,14 @@ def _publish_daemon(self, log_queue=None):
...
-                pub_sock.close()
-            if pull_sock.closed is False:
-                pull_sock.setsockopt(zmq.LINGER, 1)
-                pull_sock.close()
-            if context.closed is False:
-                context.term()
+            log.trace('Publish daemon caught Keyboard interupt, tearing down')
+        # Cleanly close the sockets if we're shutting down
+        if pub_sock.closed is False:
+            pub_sock.close()
```
isbm: Hmm... Why not reuse [function you've just created](https://github.com/saltstack/salt/pull/50463/files#diff-7e93fb0d8129a1b378ae1a10559936cbR901)? I would rather have any socket variable closing function that takes a socket variable and closes it with all the precautions and logging, instead of always checking that.

To me it is too much `if/else` and self-repeating, rather then can be just one function reused to all the sockets.

salt/transport/zeromq.py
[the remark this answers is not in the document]
isbm: I'd make this check inside `self.pub_close()` and add logging there.

salt/transport/zeromq.py
```diff
@@ -905,12 +942,11 @@ def publish(self, load):
             'Sending payload to publish daemon. jid=%s size=%d',
             load.get('jid', None), len(payload),
         )
-        pub_sock.send(payload)
+        if not self.pub_sock:
+            self.pub_connect()
+        self.pub_sock.send(payload)
```
isbm: It would be make more sense to send if socket is opened. So I would make `self.pub_connect()` check for socket availability and add logging there too. So then the logic would be:

```python
if self.pub_socket_connect():
    self.pub_sock.send(payload)
```

And if not, you already have an error in the log. You read this better and also one line of code less. :wink: While it is still not pure OOP, yet better.

tests/unit/transport/test_zeromq.py
```diff
@@ -305,3 +309,221 @@ def test_send(self):
...
+            server_channel.publish(load)
+
+    @staticmethod
+    def _send_large(opts, sid, num=10, size=250000 * 3):
+        server_channel = salt.transport.zeromq.ZeroMQPubServerChannel(opts)
+        for i in range(num):
+            load = {'tgt_type': 'glob', 'tgt': '*', 'jid': '{}-{}'.format(sid, i), 'xdata': '0' * size}
+            server_channel.publish(load)
+
+    def test_issue_36469_tcp(self):
```
isbm: Could you add issue URL in the docstring or comments please? Easier to click&open from IDE/editor :wink:

tests/unit/transport/test_zeromq.py
```diff
@@ -305,3 +309,221 @@ def test_send(self):
...
+        server_channel = salt.transport.zeromq.ZeroMQPubServerChannel(opts)
+        server_channel.pre_fork(self.process_manager, kwargs={
+            'log_queue': salt.log.setup.get_multiprocessing_logging_queue()
+        })
+        send_num = 10 * 4
+        expect = []
+        results = []
+        pub_uri = 'tcp://{interface}:{publish_port}'.format(**opts)
+        # Allow time for server channel to start, especially on windows
+        time.sleep(2)
```
isbm: This is not a good sign. What if needed 3 seconds? Or overloaded CI 10?.. Or 1? Or none at all? We are waiting on server channel, so we should rather look for if it is ready.

tests/unit/transport/test_zeromq.py
```diff
@@ -305,3 +309,221 @@ def test_send(self):
...
+        expect = ['{}-{}'.format(a, b) for a in range(10) for b in (1, 2, 3, 4)]
+        server_channel.publish({'tgt_type': 'glob', 'tgt': '*', 'stop': True})
+        gather.join()
+        server_channel.pub_close()
+        assert len(results) == send_num, (len(results), set(expect).difference(results))
+
+    @skipIf(salt.utils.is_windows(), 'Skip on Windows OS')
+    def test_issue_36469_udp(self):
+        '''
+        Test sending both large and small messags to publisher using UDP
```
isbm: Add issue URL?

isbm (changes requested):
I have few points here, maybe you could look at them. Also you seems wrote almost pure integration tests, since they are not looking how code logic works but how different pieces are working together. I would suggest to move them away to the integration tests part and do not call them unit tests.

salt/transport/zeromq.py
dwoz: If pub_sock returns true, the socket is opened and connected.

tests/unit/transport/test_zeromq.py
dwoz: I agree but I am not sure how to check if the daemon process is started without adding some synchronization logic. Unfortunately we can't check if the sockets are connected because zmq will always allow you to connect even if the other side is not bound or connected. I'm open to other ideas.

salt/transport/zeromq.py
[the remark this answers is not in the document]
dwoz: Then we'd be calling pub_close ever time we call pub_connect?

tests/unit/transport/test_zeromq.py
dwoz: btw, generally for this kind of thing I would use a `multiprocessing.Event` and have the get set after the sockets are connected up and we're starting the main daemon loop. We can check the even from the parent process to know if the daemon is 'ready'. This approach has the added benefit being able to signal the daemon to stop from the parent process. Still, If going down that path, I'd rather do it in develop once this gets merged up.

salt/transport/zeromq.py
dwoz: This is not the same socket. The socket here is being created in the daemon process. The `pub_sock` you are talking about here is sending things to the minions. I would have named it something different but not trying to change to much as a part of a bug fix PR.

salt/transport/zeromq.py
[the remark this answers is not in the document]
isbm: Then you'd need to make pub_connect only when it is needed.

salt/transport/zeromq.py
isbm: Point is, right now it will just crash, if there is a network problem. But if you would put all those timeouts and stuff into generic `socket_connect` you could then better handle exceptions/logging. Like I said, it is up to you.

salt/transport/zeromq.py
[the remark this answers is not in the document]
dwoz: We are only calling connect if it's needed right now. Having the close call here is just a convenience. If we want to enforce connect only get's called once, raising an exception here makes the most sense to me.

tests/unit/transport/test_zeromq.py
isbm: OK I can't lie, so I still honestly disagree :rofl: to keep such tests as Unit and have them fail just because of timing that we even don't know how to test.

@cachedout can we agree that we call it "technical debt" and then @dwoz gets an extra-cycle to move these type of tests away from Unit into something like `tests/functional`, that are almost like integration, but not against full-blown minion/master?

salt/transport/zeromq.py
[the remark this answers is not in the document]
isbm: @cachedout @dwoz OK.

tests/unit/transport/test_zeromq.py
dwoz: @isbm I'm happy to start moving things to `tests/functional` in the develop branch. @cachedout We're good here?

cachedout: @isbm That is a reasonable compromise. Let's do that.

Issue #50104, opened 2018-10-18
About pillar_cache only work for pillar_roots but not ext_pillar


rongzeng54:
### Description of Issue/Question
A personal suggestion
Except for different data sources, Pillar_roots and ext_pillar have different frequency of change and real-time data requirements. The latter usually changes more frequently and the real-time requirements are higher.
Pillar_cache has worked fine, we confirming that it can reduce the CPU overhead caused by the pillar request.
Is it possible to control the separate cache of the pillar_roots by adding new configuration items, and ext_pillar will not cache?

Ch3LL:
i dont think this is possible but maybe @saltstack/team-core can correct me if i'm wrong.

PR #50466, opened 2018-11-11
add ext_pillar_cache

rongzeng54:
### What does this PR do?
Add ext_pillar_cache for cacheing pillar data rendered from external pillar sources.
At the same time, pillar_cache only for cacheing pillar data rendered from pillar_roots and pillar_opts.

https://github.com/saltstack/salt/issues/50104

### Previous Behavior
pillar_cache cacheing all pillar data

### New Behavior
pillar_cache cacheing pillar data rendered from pillar_roots and pillar_opts.
ext_pillar_cache cacheing pillar data rendered from external pillar sources.

### Tests written?
No

### Commits signed with GPG?
No


salt/pillar/__init__.py
```diff
@@ -273,83 +273,158 @@ def __init__(self, opts, grains, minion_id, saltenv, ext=None, functions=None,
...
         '''
         Return the path to the cache file for the minion.
 
         Used only for disk-based backends
         '''
-        return os.path.join(self.opts['cachedir'], 'pillar_cache', minion_id)
+        return os.path.join(self.opts['cachedir'], cache_type, minion_id)
 
-    def fetch_pillar(self):
+    def fetch_pillar(self, fetch_pillar=True, fetch_ext_pillar=True):
```
cachedout: It's not clear what these do and should probably be documented in the doc string below.

salt/pillar/__init__.py
```diff
@@ -273,83 +273,158 @@ def __init__(self, opts, grains, minion_id, saltenv, ext=None, functions=None,
...
-                log.debug('Pillar cache hit for minion %s and pillarenv %s', self.minion_id, self.pillarenv)
-                return self.cache[self.minion_id][self.pillarenv]
+        pillar_ret = {}
+        ext_pillar_ret = {}
+        if self.cache:
+            log.debug('Scanning pillar cache for information about minion %s and pillarenv %s',
+                      self.minion_id, self.pillarenv)
+            log.debug('Scanning cache: %s', self.cache._dict)
+            # Check the cache!
+            if self.minion_id in self.cache:  # Keyed by minion_id
```
cachedout: This is getting to be a bit nested. Would it make sense at this point to break these out and refactor into new methods?

conf/suse/master
```diff
@@ -961,6 +963,28 @@ syndic_user: salt
...
+# See the additional ``pillar_cache_*`` configuration options.
+#ext_pillar_cache: False
+
+# If and only if a master has set ``ext_pillar_cache: True``, others same with 
+# pillar_cache_ttl.
+#ext_pillar_cache_ttl: 3600
+
+# If and only if a master has set ``ext_pillar_cache: True``, others same with 
+# pillar_cache_backend.
+#ext_pillar_cache_backend: disk
```
isbm: This optioning is kind of scattered and also can be optimised. E.g. this would enable it and put TTL to 3600:

```yaml
pillar_cache:
   backend: disk
   ttl: 3600
```

Commenting all that out would be equal to disabling it.

isbm (changes requested):
This design raises me some fundamental questions.

First, it makes sense only a little bit when you have extensive repetitive requests. However, if your  access to the minion[s] are rare or has enough time distance and yet you keep pillars unchanged, it would make sense to invalidate the cache by a pillar content modification instead. Second, this design has no way to immediately purge/update the pillar. Third, it seems has a security flaw (CVE, here you come!): if you unregister a minion and get another minion with the same minion ID, you get the content of the pillar, which supposed to be secure!

@cachedout hmm, too much questions on the design (regardless the implementation). This rather has to go through the RFC first, I'd strongly suggest.

isbm:
@rongzeng54 nor I designed this part, neither I deny there are no issues. Currently my point is that your design is not efficient and seems vulnerable. So I only point out that it should be first agreed in RFC beforehand and only _then_ implemented, _according to_ agreed RFC.

isbm:
@rongzeng54 what is in `xxx.py`?

cachedout:
@isbm But we already _have_ pillar caches. I'm not sure I understand how this would introduce a security flaw that would not already be present. (And, to be honest, I'm not sure how caching would turn this into a security hole that wouldn't already be present in uncached pillars given the scenario that you describe.)

While I'm open to the idea of an RFC, given that this models the existing implementation for pillar caches, this seems more like an enhancement to me than a full-blown new feature. Thoughts?

isbm:
> While I'm open to the idea of an RFC, given that this models the existing implementation for pillar caches, this seems more like an enhancement to me than a full-blown new feature. Thoughts?

Well, I know. :wink: I am all for RFC, because while I am **not against** the improvements, I want it through through. E.g. time-based is not entirely nice. How to invalidate the cache? How to protect it from swapping the minion? — etc. I also not sure the configuration is nice as well. If we are configuring pillar cache, then likely we want it nested and grouped, instead of 10 more long options at the root of the config.

cachedout:
@rongzeng54 How do you feel about proposing an RFC for this before we consider merging this code?

cachedout:
We need to make some decisions here.

@isbm Are there _specific_ concerns that you have about this implementation?

isbm:
> @isbm Are there _specific_ concerns that you have about this implementation?

- More root-level options turns current config looks like Windows 3.1 INI file (and even that one is more structured). Since pillars are now _all_ can be cached, maybe it makes sense to have a better configuration:

```yaml
pillar_cache: True
```

Above is valid for both pillars, defaults to everything. Or split, if needed:

```yaml
pillar_cache:
  internal:
    backend: disk
    ttl: 3600
  external: False
```

Or differently configure:

```yaml
pillar_cache:
  internal:
    backend: disk
    ttl: 3600
  external:
    backend: memory
```

It is, of course, easier to have `if __opts__['some_random_opt']:` but we have them so many already that they can be all around the place unstructured and ungrouped.

I also did not get any answer [to the comment I already made](https://github.com/saltstack/salt/pull/50466/files#r233192289).

- We need to think how to instantly invalidate cache. Especially, if cache has very long TTL and we want changes _right now_.

- TTL-based cache isn't the most efficient. If data is changing at random point of time, it may not be just requiring everything to re-read (and thus re-CPU things), but also slow down frequently working minions, which will re-CPU pillars just because TTL is over, while this is not needed. I would think of something like last time update or checksum/hash of the pillar object, and let cache knows about it, instead of TTL. So we can drop TTL at all and update cache only when data is actually changing.

- I don't see any of Unit tests, while changes impact is significant.

cachedout:
@rongzeng54 Do you have feedback on the comments from @isbm ? Also, we need a merge conflict resolved, please.

dwoz:
@rongzeng54 Can you please resolve the merge conflict? Thanks!

isbm:
@dwoz I think @rongzeng54 should also stop ignoring discussion and give a proper responses to other comments.

dwoz:
@rongzeng54 Ping.

dwoz:
@rongzeng54 This PR seems to have gone stale. I am going to close the issue for now. Please feel free to create a new PR with these changes if you find time to work on this again.

PR #50467, opened 2018-11-11
Use patched version of test-kitchen until next upstream release

dwoz:
### What does this PR do?

Point `test-kitchen` at my bug fix branch until next `test-kitchen` release so we can configure the winrm `receive_timeout` option.

We're seeing `[execution expired]` exceptions from test kitchen in our Windows test suites. Point `test-kitchen` at my bug fix branch for upstream issue [test-kitchen 1331](https://github.com/test-kitchen/test-kitchen/issues/1331).

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,7 +2,8 @@
 
 source 'https://rubygems.org'
 
-gem 'test-kitchen', '~>1.21'
+# Point this back at the test-kitchen package after 1.23.3 is relased
+gem 'test-kitchen', :git => 'https://github.com/dwoz/test-kitchen.git', :branch => 'winrm_opts'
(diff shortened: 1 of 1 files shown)


dubb-b:
@dwoz We need to update the gems in each kitchen node after this correct?

dwoz:
@dubb-b I don't think we need to do anything special here since Jenkins runs a bundle install before starting the kitchen builds. Am I missing something?

dubb-b:
@dwoz I have seen the version cache. may have to clear out the .gems folder. Im not quite sure. We can let it go and see if there are issues afterward.

dubb-b (approved):

PR #50471, opened 2018-11-12
Fix debian ip

horakmar:
Fixes handling of space delimited lists in /etc/network/interfaces parser.
Bug: Parser creates from the file unicode strings, but expects str. Result is malformed attributes dns-nameservers and dns-search in /etc/network/interfaces.


max-arnold:
@horakmar Does your issue look like this?

```
----------
          ID: eth0
    Function: network.managed
      Result: None
     Comment: Interface eth0 is set to be updated:
              --- 
              +++ 
              @@ -3,5 +3,5 @@
                   address 1.2.3.4
              
                   netmask 255.255.255.0
              
                   gateway 1.2.3.254
              
              -    dns-nameservers 1 0 . 2 0 0 . 0 . 2 5 4   8 . 8 . 8 . 8
              
              +    dns-nameservers 10.200.0.254 8.8.8.8
     Started: 15:04:00.282830
    Duration: 64.541 ms
     Changes:   
```

We always have this diff on `state.highstate` after updating to `2018.3.x, just haven't filed an issue yet.

horakmar:
Yes, that's it. I can see in merge conflicts, that it's already fixed.

PR #50474, opened 2018-11-12
Allow to specify used trustmodel in `gpg.verify`

kyuubii:
### What does this PR do?
It allows to explicitly specify the trust model used by `gpg.verify`, such as `direct` or `tofu+pgp`

None

### Previous Behavior
`gpg.verify` always used the trust model implicitly defined by python-gnupg/GnuPG

### New Behavior
The previous behaviour can be overridden by explicitly providing a trust model.

### Tests written?
No

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/modules/gpg.py
+++ b/salt/modules/gpg.py
@@ -1031,7 +1031,8 @@ def verify(text=None,
            user=None,
            filename=None,
            gnupghome=None,
-           signature=None):
+           signature=None,
+           trustmodel=None):
     '''
     Verify a message or file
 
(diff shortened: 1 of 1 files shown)


dwoz:
@kyuubii it looks like there are 0 unit and/or integration tests for this module. How do you feel about adding one, either unit or integration?

cachedout (approved):

kyuubii:
> @kyuubii it looks like there are 0 unit and/or integration tests for this module. How do you feel about adding one, either unit or integration?
Due to our current time constraints, I won't get around to provide unit or even integration tests.

PR #41025, opened 2017-05-03
Fix merge order of options in run_wfunc (fix mine.get issues)

ptitdoc:
### What does this PR do?
This fix issues with mine.get when using salt-ssh

### What issues does this PR fix or reference?
https://github.com/saltstack/salt/issues/36796

### Previous Behavior
pillar.Pillar() is called with options where master_opts options are overwritten (because of dict merge/update order).

### New Behavior
pillar.Pillar() is called with options where master_opts take precedence. This allow mine.get over salt-ssh to work properly.

### Tests written?
No

Please review [Salt's Contributing Guide](https://docs.saltstack.com/en/latest/topics/development/contributing.html) for best practices.

Issue #36796, opened 2016-10-05
Problem with salt-ssh and mine.get 


manjiki:
### Description of Issue/Question

Under some circumstances highstate fails when using salt-ssh and `salt['mine.get']` with error:

```
Jinja variable 'salt.utils.templates.AliasedLoader object' has no attribute 'grains.get'
```

Although I am able to successfully run each state individually, I get the above error when running highstate. The problem doesn't occure if we:
- remove the block/state where  `salt['mine.get']` is called
- change the order of states in top.sls
### Setup

**Roster file:**

```
murray2:
  host: 10.10.10.56
  user: dan
  sudo: True
  priv: ~/.ssh/dan.pem
  thin_dir: /opt/salt-thin
  mine_functions:
    network.interfaces: []
    grains.item:
      - roles
      - fqdn
      - applications
      - ip_interfaces
  grains:
    roles:
      - backend
    applications:
      - chuck
```

**states/test_hosts/init.sls:**

```
/etc/hostname:
  file.managed:
    - contents: {{ salt['grains.get']('id') }}

{%- for host, data in salt['mine.get']('*', 'grains.item').items() %}
hostsfile_{{host}}:
  host.present:
    - names:
      - {{host}}
    - ip: {{data.ip_interfaces.eth0[0]}}
{% endfor%}
```

**states/test_apps/init.sls**

```
{%- for app in salt['grains.get']('applications',[]) %}
{%- set app_data = salt['pillar.get']('applications:%s' % app) %}
{%- from 'test_apps/macros.j2' import this_macro  %}
/tmp/{{app}}:
  file.managed:
    - contents: {{ salt['disk.blkid']('')}}

{{this_macro(app)}}
{% endfor %}
```

**states/top.sls:**

```
base:
  '*':
    - test_hosts
    - test_apps
```
### Steps to Reproduce Issue

```
$ salt-ssh 'murray2' state.highstate

[ERROR   ] Rendering exception occurred: Jinja variable 'salt.utils.templates.AliasedLoader object' has no attribute 'grains.get'
[CRITICAL] Rendering SLS 'base:test_apps' failed: Jinja variable 'salt.utils.templates.AliasedLoader object' has no attribute 'grains.get'
murray2:
    - Rendering SLS 'base:test_apps' failed: Jinja variable 'salt.utils.templates.AliasedLoader object' has no attribute 'grains.get'
```

The last lines of the master.log are:

```
[...]

PR #50489, opened 2018-11-13
ssh-client: fix merge order of options in run_wfunc (fix usage of mine.get in pillar)

ptitdoc:
### What does this PR do?
pillar.Pillar() is called with options where master_opts options are overwritten (because of dict merge/update order). As a consequence, most saltutil functions are not available in the pillar when salt-ssh client is used.

This PR change the opts merge order selectively based on specific wfunc routines (eg: mine.get, pillar.items ...)

https://github.com/saltstack/salt/issues/36796
https://github.com/saltstack/salt/pull/41025

### Previous Behavior
master_opts are overwritten when calling mine.get or pillar.items within salt-ssh client.

master_opts takes precedence when calling mine.get or pillar.items within salt-ssh client.

### Tests written?
No

### Commits signed with GPG?
No


thatch45 (approved):
This is a very good solution to the issue here, thanks!

salt/client/ssh/__init__.py
```diff
@@ -1075,8 +1075,22 @@ def run_wfunc(self):
...
+            #
+            # Depending on popts merge order, it will overwrite some options found in opts_pkg['__master_opts__']
+            master_centric_funcs = [
+              "pillar.items",
+              "mine.get"
+            ]
+
+            # Pillar compilation is a master centric operation.
+            # Master options take precedence during Pillar compilation
+            popts.update(opts_pkg['__master_opts__'])
```
isbm: Q: are the **keys** of the `opts_pkg` aren't the same as in `__master_opts__`? I.e. aren't `opts_pkg` just do not override the master opts? In this case the first `update(opts_pkg)` is not really needed. But I am not entirely sure here.

isbm (changes requested):

cachedout:
@ptitdoc Did you see the feedback here from @isbm? Once those issues are resolved we should be able to merge this. Thanks!

salt/client/ssh/__init__.py
```diff
@@ -1085,6 +1099,13 @@ def run_wfunc(self):
                     )
             pillar_data = pillar.compile_pillar()
 
+            # Once pillar has been compiled, restore priority of minion opts
+            if self.fun not in master_centric_funcs:
+                log.debug('%s is a minion function', self.fun)
+                popts.update(opts_pkg)
```
isbm: @cachedout I am still wondering if we need that `popts.update(opts_pkg)` first time. This one is a second.

cachedout: Will it do any harm though?

thatch45: I think this is fine and it makes sense.

thatch45 (approved):

PR #50492, opened 2018-11-13
Back-port #50228 and #50443 to 2018.3

rallytime:
Back-port #50228 and #50443 to 2018.3

### DIFF
--- a/salt/utils/mac_utils.py
+++ b/salt/utils/mac_utils.py
@@ -11,6 +11,7 @@ import subprocess
 import os
 import plistlib
 import time
+import xml.parsers.expat
 
 # Import Salt Libs
 import salt.modules.cmdmod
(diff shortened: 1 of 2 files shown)


salt/utils/mac_utils.py
```diff
@@ -328,39 +349,59 @@ def _available_services(refresh=False):
                 # Follow symbolic links of files in _launchd_paths
                 file_path = os.path.join(root, file_name)
                 true_path = os.path.realpath(file_path)
-
+                log.trace('Gathering service info for {}'.format(true_path))
```
terminalmage: Can you add a commit here to fix this logging? We shouldn't be using `.format()` in logging. And then on merge-forward, it will correct it in the other branches.

salt/utils/mac_utils.py
```diff
@@ -328,39 +349,59 @@ def _available_services(refresh=False):
...
-                    # Handle malformed plist files
-                    _available_services[os.path.basename(file_name).lower()] = {
+                    # not all launchd plists contain a Label key
+                    _available_services[plist['Label'].lower()] = {
                         'file_name': file_name,
                         'file_path': true_path,
                         'plist': plist}
+                except KeyError:
+                    log.debug('Service {} does not contain a'
+                              ' Label key. Skipping.'.format(true_path))
```
terminalmage: Again, the logging should not be using `.format()`

terminalmage (changes requested):
@rallytime I know you didn't introduce these new log messages, but if you can fix them here in 2018.3 then the fixes will propagate forward on merge.

rallytime:
@terminalmage Good eye! I have fixed those logging formats.

terminalmage (approved):

PR #50504, opened 2018-11-13
Set jenkins docs job to treat any warnings as errors (develop branch)

terminalmage:
See https://github.com/saltstack/salt/pull/50500

Also, fix a bunch of build errors.


rallytime (approved):

terminalmage:
These commits are already in master, so this doesn't need to be ported and can be removed from the list of PRs to port.

Issue #50507, opened 2018-11-14
multiple salt-minion process starts on el6


rongzeng54:
### Description of Issue/Question
In el6 environment without systemd, quickly start salt-minion will gain more than one salt-minion instance with same ID.
Then each command sent from the salt-master will run multiple times.
You can easily reproduce the problem by following the steps below.
```
[root@test ~]# ps -ef | grep -i salt-minion
root      236099  188843  0 22:17 pts/0    00:00:00 grep -i salt-minion
[root@test ~]# service salt-minion start &
[1] 236210
[root@test ~]# service salt-minion start &
[2] 236253
[root@test ~]# service salt-minion start &
[3] 236394
[root@test ~]# service salt-minion start &
[4] 236523
[root@test ~]# service salt-minion start &
[5] 236659
[root@test ~]# service salt-minion start &
[6] 236756
[root@test ~]# service salt-minion start &
[7] 236959

[root@test ~]# Starting salt-minion:root:test daemon: Starting salt-minion:root:test daemon: Starting salt-minion:root:testService salt-minion:root:test already running
OK
Service salt-minion:root:test already running
OK
FAIL
Service salt-minion:root:test already running
Service salt-minion:root:test already running

[1]   Done                    service salt-minion start
[2]   Done                    service salt-minion start
[3]   Exit 1                  service salt-minion start
[4]   Done                    service salt-minion start
[5]   Done                    service salt-minion start
[6]-  Done                    service salt-minion start
[7]+  Done                    service salt-minion start

[root@test ~]# ps -ef | grep -i salt-minion
root      237646       1  1 22:17 ?        00:00:01 /usr/bin/python2.7 /usr/bin/salt-minion -c /etc/salt -d
root      237648  237646  0 22:17 ?        00:00:00 /usr/bin/python2.7 /usr/bin/salt-minion -c /etc/salt -d
root      237759       1  2 22:17 ?        00:00:02 /usr/bin/python2.7 /usr/bin/salt-minion -c /etc/salt -d
root      237762  237759  0 22:17 ?        00:00:00 /usr/bin/python2.7 /usr/bin/salt-minion -c /etc/salt -d
root      251064  188843  0 22:19 pts/0    00:00:00 grep -i salt-minion
```

### Setup
No

### Steps to Reproduce Issue
No

### Versions Report
```
[root@test ~]# salt-call --versions-report
Salt Version:
           Salt: 2017.7.5

Dependency Versions:
           cffi: Not Installed
[...]

PR #50513, opened 2018-11-14
Bugfix for multi salt-minion process startup

rongzeng54:
### What does this PR do?
Bugfix for multi salt-minion process startup

https://github.com/saltstack/salt/issues/50507

### Previous Behavior
Starts slowly, does not exclude other startup action during startup.

### New Behavior
Starts slowly, does exclude other startup action during startup by lockfile.

### Tests written?
No

### Commits signed with GPG?
No

### DIFF
--- a/pkg/rpm/salt-minion
+++ b/pkg/rpm/salt-minion
@@ -67,12 +67,14 @@ _su_cmd() {
 
 
 _get_pid() {
-    cat $PID_FILE 2>/dev/null
+    local pid_file="$1"
+    cat $(eval echo '$'$pid_file) 2>/dev/null
 }
 
 
 _is_running() {
-    [ -n "$(_get_pid)" ] && ps wwwaxu | grep '[s]alt-minion' | awk '{print $2}' | grep -qi "\b$(_get_pid)\b"
+    local pid_file="$1"
+    [ -n "$(_get_pid $pid_file)" ] && ps wwwaxu | grep '[s]alt-minion' | awk '{print $2}' | grep -qi "\b$(_get_pid $pid_file)\b"
 }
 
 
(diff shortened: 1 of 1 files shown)


pkg/rpm/salt-minion
```diff
@@ -116,8 +116,16 @@ start() {
     # $1 - config dir
     local retval=0
 
+    if [ -f $LOCK_FILE ]; then
+        echo "Service $SERVICE:$MINION_USER:$MINION_ID is starting"
```
cachedout: What happens if for some reason the minion crashes and the lockfile is not cleaned up? Won't the minion then fail to start correctly until it is manually removed?

rongzeng54: This lock file is different from the usual usage and it only exist within the startup process.
When the startup is completed, whether the result is successful or fails, the lock file will be cleaned up.

terminalmage: I don't like this either, it opens us up for problems later if the script is modified and the lockfile handling is not properly managed.

rongzeng54: hmm, seems in order to solve a small probability problem, a large probability problem is introduced. Before file lock management becomes reliable, i closed it.

rongzeng54: @terminalmage  @cachedout 
Now it seems that the problem can be solved.

cachedout:
@dmurphy18 Do you have any thoughts here?

pkg/rpm/salt-minion
dmurphy18: Sorry for delay in reviewing this, but the changes should be made in the salt-pack repository, see https://github.com/saltstack/salt-pack/blob/develop/file_roots/pkg/salt/2017_7/rhel6/sources/salt-minion for the 2017.7 branch and there is a similar change for the 2018.3 branch.

The contents of salt-minion are similar but different.  Need to update salt repository with a more recent version of packaging or pointers to salt-pack (salt-pack-py3 for Python3).

Further to the problem of cleanup on failure, perhaps should have a cleanup function added based off trapping signals, etc. which would ensure the lock file is removed.

dmurphy18 (changes requested):
Sorry for delay in reviewing this, but the changes should be made in the salt-pack repository, see https://github.com/saltstack/salt-pack/blob/develop/file_roots/pkg/salt/2017_7/rhel6/sources/salt-minion for the 2017.7 branch and there is a similar change for the 2018.3 branch.

The contents of salt-minion are similar but different and any changes should be made in salt-pack.. 
SaltStack needs to update salt repository with a more recent version of packaging or pointers to salt-pack (salt-pack-py3 for Python3).

Further to the problem of cleanup on failure, perhaps there should be an additional cleanup function added based off trapping signals, etc. which would ensure the lock file is removed on exit.

thatch45:
Given the status of this PR I am going to close it out. This change should be made to salt-pack and we should consider removing some of the pkg directory files wholesale

PR #50523, opened 2018-11-15
Bugfix: OS arch fall-back when no RPM installed on inird image

isbm:
### What does this PR do?

Bugfix.

Salt is unable to determine OS architecture grain on `initrd` Linux images for RPM systems.

### Previous Behavior

OS architecture is not detected.

### New Behavior

OS architecture is detected.

### Tests written?

Yes

### DIFF
--- a/tests/unit/utils/test_pkg.py
+++ b/tests/unit/utils/test_pkg.py
@@ -3,7 +3,7 @@
 from __future__ import absolute_import, unicode_literals, print_function
 
 from tests.support.unit import TestCase, skipIf
-from tests.support.mock import Mock, MagicMock, patch, NO_MOCK, NO_MOCK_REASON
+from tests.support.mock import MagicMock, patch, NO_MOCK, NO_MOCK_REASON
 import salt.utils.pkg
 from salt.utils.pkg import rpm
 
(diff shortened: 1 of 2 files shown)


meaksh (approved):

isbm:
Package [is out here](https://build.opensuse.org/package/view_file/systemsmanagement:saltstack:products:testing/salt/salt.changes?expand=1).

tests/unit/utils/test_pkg.py
```diff
@@ -45,3 +50,62 @@ def test_split_comparison(self):
             oper, verstr = salt.utils.pkg.split_comparison(test_parameter[0])
             self.assertEqual(test_parameter[1], oper)
             self.assertEqual(test_parameter[2], verstr)
+
+
+@skipIf(NO_MOCK, NO_MOCK_REASON)
+@skipIf(pytest is None, 'PyTest is missing')
```
terminalmage: I see nothing pytest-specific about these tests, so why is it being skipped?

isbm: Well, this is a general pattern, apparently. Maybe we should stop even doing it, since pytest is always there.

cachedout:
@isbm There are some lint errors here, it seems

terminalmage (approved):

isbm:
@cachedout OK, lint error gone after rebase on `develop`. :confused: So should be OK now.

PR #50541, opened 2018-11-16
btrfs: add all subvolume commands

aplanas:
### What does this PR do?
Add all subvolume commands to the btrfs module

### Tests written?

Yes

### DIFF
--- a/salt/modules/btrfs.py
+++ b/salt/modules/btrfs.py
@@ -20,10 +20,10 @@ Module for managing BTRFS file systems.
 
 # Import Python libs
 from __future__ import absolute_import, print_function, unicode_literals
+import itertools
 import os
 import re
 import uuid
-
 
 # Import Salt libs
 import salt.utils.fsutils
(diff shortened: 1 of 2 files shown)


salt/modules/btrfs.py
```diff
@@ -673,3 +673,465 @@ def properties(obj, type=None, set=None):
...
+
+        salt '*' btrfs.subvolume_find_new /var/volumes/tmp 1024
+
+    '''
+    cmd = ['btrfs', 'subvolume', 'find-new', name, last_gen]
+
+    res = __salt__['cmd.run_all'](cmd)
+    salt.utils.fsutils._verify_run(res)
+
+    lines = res['stdout'].split('\n')
```
isbm:
```python
import os
...
....split(os.linesep)
```

tests/unit/modules/test_btrfs.py
```diff
@@ -362,3 +362,300 @@ def test_properties_error(self):
         '''
         self.assertRaises(CommandExecutionError, btrfs.properties,
                           '/dev/sda1', 'subvol', True)
+
+    def test_subvolume_create(self):
+        '''
+        Test btrfs subvolume create
+        '''
+        # Fails when qgroupids is not a list
+        self.assertRaises(CommandExecutionError, btrfs.subvolume_create,
```
isbm: Please use pytest instead. And for the raising exception should be another UT.

tests/unit/modules/test_btrfs.py
```diff
@@ -362,3 +362,300 @@ def test_properties_error(self):
...
+                          'var', qgroupids='1')
+
+        salt_mock = {
+            'cmd.run_all': MagicMock(return_value={'recode': 0}),
+        }
+        with patch.dict(btrfs.__salt__, salt_mock):
+            self.assertTrue(btrfs.subvolume_create('var', dest='/mnt'))
```
isbm: Please use just plain `assert` from Python.

tests/unit/modules/test_btrfs.py
```diff
@@ -362,3 +362,300 @@ def test_properties_error(self):
...
+        Test btrfs subvolume delete
+        '''
+        # We need to provide name or names
+        self.assertRaises(CommandExecutionError, btrfs.subvolume_delete)
+        self.assertRaises(CommandExecutionError, btrfs.subvolume_delete,
+                          names='var')
+
+        # Parameter commit can be 'after' or 'each'
+        self.assertRaises(CommandExecutionError, btrfs.subvolume_delete,
+                          name='var', commit='maybe')
```
isbm: Please break them into three UT

isbm (changes requested):
Please take a look at these points and please use pytest's ecosystem instead.

salt/modules/btrfs.py
aplanas: I moved to ''.splitlines()

tests/unit/modules/test_btrfs.py
aplanas: Done

tests/unit/modules/test_btrfs.py
aplanas: done

tests/unit/modules/test_btrfs.py
aplanas: I moved into two, as the two first are testing basically the same.

tests/unit/modules/test_btrfs.py
cachedout: @isbm What is the motivation for just suddenly switching to pytest here? We certainly have long-term plans to go in that direction but are we sure we want to just start importing pytest into individual tests like this?

salt/modules/btrfs.py
```diff
@@ -673,3 +673,465 @@ def properties(obj, type=None, set=None):
...
+    if not name and not (names and type(names) is list):
+        raise CommandExecutionError('Provide a value for the name parameter')
+
+    if commit and commit not in ('after', 'each'):
+        raise CommandExecutionError('Value for commit not recognized')
+
+    cmd = ['btrfs', 'subvolume', 'delete']
+    if commit == 'after':
+        cmd.append('--commit-after')
+    if commit == 'each':
```
isbm: `elif`

salt/modules/btrfs.py
```diff
@@ -673,3 +673,465 @@ def properties(obj, type=None, set=None):
...
+    value = None
+    if line.startswith(key):
+        line = line[len(key):]
+        line = line.strip()
+        if use_rest:
+            value = line
+            line = ''
+        else:
+            value, line = line.split(' ', 1)
+    return (value, line.strip())
```
isbm: Extra parenthesis

salt/modules/btrfs.py
```diff
@@ -673,3 +673,465 @@ def properties(obj, type=None, set=None):
...
+
+    cmd = ['btrfs', 'subvolume', 'snapshot']
+    if read_only:
+        cmd.append('-r')
+    if dest and not name:
+        cmd.append(dest)
+    if dest and name:
+        name = os.path.join(dest, name)
+    if name:
+        cmd.append(name)
```
isbm: No need in all these `if if if`:
```python
cmd.append(os.path.join(dest or '', name or '').strip(os.path.sep))
```

salt/modules/btrfs.py
isbm: Or so:
```python
value = None
if line.startswith(key):
    line = line[len(key):].strip()
    value, line = use_rest and (line, '') or line.split(' ', 1)
return value, line.strip()

**Review by isbm — CHANGES_REQUESTED** [review]

File: salt/modules/btrfs.py (hunk 5 above)
Comment (aplanas) [reply salt/modules/btrfs.py]: done

File: tests/unit/modules/test_btrfs.py (hunk 2 above)
Comment (aplanas) [reply tests/unit/modules/test_btrfs.py]: IMHO I do not like pytest. Basically because the magic that involves `assert`. In normal Python `assert` is a language statement that evaluate the expression (True or False) and throws an `AssertionError`.

Pytest change that to include the features that are already in the `self.assertEqual` family. Both code generate very similar report to show the differences of data structures. So I do not see the gain incorporating more magic to this problem.

But I am happy to adhere to the project expectations in regards to tests frameworks.

File: salt/modules/btrfs.py (hunk 6 above)
Comment (aplanas) [reply salt/modules/btrfs.py]: Why to obfuscate the code with `and`s and `or`s like if this is pre-Python 2.4?

I removed the extra line with `.strip()` and removed the optional parents (I usually like to be explicit when I return a tuple, but I guess that here is clear enough)

File: salt/modules/btrfs.py (hunk 7 above)
Comment (aplanas) [reply salt/modules/btrfs.py]: I am not sure why to favor a more obfuscated code.

I am personally not so smart to realize that the last `.strip(os.path.sep)` is needed because if `name` is not assigned, `os.path.join(dest, '')` will add this unneeded extra '/'.

Let me be easy and explicit here!

File: tests/unit/modules/test_btrfs.py (hunk 2 above)
Comment (isbm) [reply tests/unit/modules/test_btrfs.py]: > IMHO I do not like pytest.

`¯\_(ツ)_/¯`. This is what we are using here and pytest gives _better results_ to catch the error.

P.S. I like pytest.

**Comment by aplanas** [comment]
I changed the code a bit. I added one new function, `subvolume_exists`, and I changed `create` and `delete` to first check if the volume is there or is missing.

I updated the tests and added more scenarios to cover the new features.

**Review by isbm — CHANGES_REQUESTED** [review]

File: salt/modules/btrfs.py [hunk 8]
```diff
@@ -673,3 +673,496 @@ def properties(obj, type=None, set=None):
...
+
+    cmd = ['btrfs', 'subvolume', 'create']
+    if type(qgroupids) is list:
+        cmd.append('-i')
+        cmd.extend(qgroupids)
+    cmd.append(name)
+
+    res = __salt__['cmd.run_all'](cmd)
+    salt.utils.fsutils._verify_run(res)
+    return True
```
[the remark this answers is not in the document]
aplanas: `salt.utils.fsutils._verify_run` throws an exception then the command fails and log the reason.

Do you think that makes more sense to catch the CommandExecutionError and return False there?

salt/modules/btrfs.py
```diff
@@ -673,3 +673,496 @@ def properties(obj, type=None, set=None):
...
+    cmd = ['btrfs', 'subvolume', 'delete']
+    if commit == 'after':
+        cmd.append('--commit-after')
+    elif commit == 'each':
+        cmd.append('--commit-each')
+    cmd.extend(names)
+
+    res = __salt__['cmd.run_all'](cmd)
+    salt.utils.fsutils._verify_run(res)
+    return True
```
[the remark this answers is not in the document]
aplanas: same answer

salt/modules/btrfs.py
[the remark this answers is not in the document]
isbm: No, I just saw you return just `True`, so I am wondering if that doesn't depend on how we are going to use it. E.g. if this boolean is going to be reused something like:

```jinja2
{%if __salt__['btrfs.subvolume_create'](...)%}
   {# Do something #}
{% else %}
   {# Do something else #}
{% endif %}
```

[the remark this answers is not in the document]
aplanas: Yeah, puzzle me why we throw exceptions.

In any case subvolume_create and delete are aware of the existence (or not) of the subvolume:

https://github.com/saltstack/salt/pull/50541/files/14004026ad048a20badd4c68f79c43360834c757#diff-af9d04adc25a92fdf8fece1a0762dad7R733

So will return True if the subvolume was already there or was created later. A fail is that something went very bad with the command itself.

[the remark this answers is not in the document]
isbm: But I thought maybe it is correct to return `False` if subvol already existed? Because your function is called "subvolume_**create**".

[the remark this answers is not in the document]
aplanas: I really do not know!

I do not mind to change it if is more expressive, but I wonder if a False can wrongly suggest that the submodule was not created, even if was already there.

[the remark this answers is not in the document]
isbm: Well, I think this is fine, apparently. I hear your point for wrong suggest, but apparently why not: it says _create_. So was it _created_? — no, as it was no-op. Therefore clearly `False`. To know if it exists or not, should be _exists_ function. And so:

```jinja2
{% if not subvol_create('blah') and not subvol_exists('blah') %}
  <h>SCREAM!</h>
{% endif %}
```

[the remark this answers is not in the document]
isbm: OK, I think this is good. Except maybe update the documentation so we know what it returns when. Other then that we're good to go.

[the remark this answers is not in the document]
aplanas: Documented.

aplanas:
I changed the `subvolume_exists` function to avoid logging ERROR in case that a subvolume do not exits, but functionally is equivalent. Also I added the documentation required.

isbm (approved):

cachedout:
Merged! Thanks, @aplanas

Issue #50542, opened 2018-11-16
mysql.user_exists doesn't work with Mysql 8.0.13


doug-stratoscale:
### Description of Issue/Question
Using latest fluorine and running mysql.user_exists results in this:
[ERROR   ] MySQL Error 2005: Unknown MySQL server host '%' (2)

I believe this issue is as a result of this change:
https://github.com/saltstack/salt/commit/9265195deb0fd65f80cca8d1a788ce6d9deb8f43
The user_exists code now calls verify_login to check if the user exists. The problem with this approach is that having "%" as the hostname will no longer work. 

### Setup
Have Mysql 8.0.13 installed. Attempting to execute this:
salt-call mysql.user_exists 'doug' '%' 'aa'

This call works with with Mysql 5.7 and salt v2017.7

### Steps to Reproduce Issue
$ salt-call --log-level debug mysql.user_exists 'doug' '%' 'aa'
<snip>
[DEBUG   ] Loaded minion key: /etc/salt/pki/minion/minion.pem
[DEBUG   ] Doing query: SELECT VERSION()
[DEBUG   ] Doing query: SELECT column_name from information_schema.COLUMNS WHERE table_schema=%(schema)s and table_name=%(table)s and column_name=%(column)s args: {u'column': u'Password', u'table': u'user', u'schema': u'mysql'}
[ERROR   ] MySQL Error 2005: Unknown MySQL server host '%' (2)
<snip>
local:
    False

### Versions Report
Salt Version:
           Salt: 2018.11.0-2172-g0327ea0

Dependency Versions:
           cffi: Not Installed
       cherrypy: Not Installed
       dateutil: Not Installed
      docker-py: 1.8.0
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.7.2
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: 0.28.2
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.5.6
   mysql-python: 1.3.13
      pycparser: Not Installed
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: Not Installed
         Python: 2.7.5 (default, Jul 13 2018, 13:06:57)
   python-gnupg: Not Installed
         PyYAML: 3.11
          PyZMQ: 15.3.0
           RAET: Not Installed
          smmap: Not Installed
        timelib: Not Installed
        Tornado: 4.2.1
[...]

PR #50551, opened 2018-11-16
[2018.3] Fixes to verify_login in mysql module

garethgreenaway:
### What does this PR do?
Ensure that verify_login is using the host from the connection_args and not the host associated with the user.  Adding a test to ensure user_exists when the passed host is the MySQL wildcard %.

#50542 

### Previous Behavior
When `user_exists` was called with a wildcard hostname, `%`, then verify login would fail because it was trying to connect to that as a real hostname.

### New Behavior
We should ensure we're using the hostname from connection_args when verifying the login.

### Tests written?
Yes

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/modules/mysql.py
+++ b/salt/modules/mysql.py
@@ -1247,7 +1247,7 @@ def user_exists(user,
         args['password'] = password_hash
 
     if run_verify:
-        if not verify_login(user, host, password):
+        if not verify_login(user, password, **connection_args):
             return False
     try:
         _execute(cur, qry, args)
@@ -2235,7 +2235,7 @@ def showglobal(**connection_args):
     return rtnv
 
 
-def verify_login(user, host='localhost', password=None, **connection_args):
+def verify_login(user, password=None, **connection_args):
     '''
     Attempt to login using the provided credentials.
     If successful, return true.  Otherwise, return False.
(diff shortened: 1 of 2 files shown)


ghost (commented):
Thanks for taking a look at and fixing my issue..

However, I think there is a problem with the approach here for checking if a user exists. The hostname passed into the user_exists call is supposed to be the client host not the server host. Mysql users are specified by client host so "user1@host1" is different from "user1@host2". Using verify_login to check if "user1@host1" exists won't work because the verify method may be running for a host other than where "user1" intends to connect from. I believe this is why the original code was using a query against the INFORMATION_SCHEMA to determine existence.

garethgreenaway:
@doug-stratoscale the `user_exists` function is still checking the user table that both the user and host are valid, the host parameter is not used to determine which MySQL host to connect to, that is found in `connection_args`.  The `verify_login` function is checking that the password is valid for MySQL versions higher than 8.0.11 when the password hashing mechanism changed and can no longer use the PASSWORD function from MySQL.

doug-stratoscale:
@garethgreenaway Thanks for the clarification - I looked at the code again ( :) ) and it makes sense.

ghost (approved):

garethgreenaway:
@doug-stratoscale No worries! Appreciate the second pair of eyes :smile:

cachedout (approved):

PR #50552, opened 2018-11-17
Remove numbering from RFC files

isbm:
### What does this PR do?

Number suffix in RFC files doesn't make sense, because the PRs aren't removed synchronously. 
Therefore:

- `0000-template.md` is renamed to just `TEMPLATE.md`
- All files with `NNNN-something.md` are renamed to `something.md`


isbm:
@cachedout Ordering is only working when they are merged serial way. Problem is that doesn't work in our case. If you call it `0003-something` and PR stays for long months, something else as `0003-blah` gets in. Then we have couple of same numbers. This is just a noise at the end, that makes bash auto-completion working bad.

damon-atkins:
Maybe just the `Year-Description`  Year being the year it was raised.

dwoz (commented):
I agree with @cachedout about ordering. I wonder if we could just ask that people look at open PRs before assigning a number to their RFC.  We might also look at [how Python manages PEPs](https://github.com/python/peps).

isbm:
@dwoz except this doesn't work. Will you be in willing reconcile that _all the time_? Unless there is a special hook at Git that blocking merger, we will have random numbers again.

damon-atkins:
The original pull request number?

isbm:
> The original pull request number?

Why is this needed??

cachedout:
@isbm That is a fair point. I still think that a numbering scheme is good, though. Perhaps we could assign them numbers at the point at which they are merged?

isbm:
> Perhaps we could assign them numbers at the point at which they are merged?

Well, as I see it, the numbering is fine if someone from documentation team is having an extra-job to make them right. Someone who wants to craft another Git hook that will break after next upgrade... 

But I reason that we already have `NNNN-<name>.md` format. So it seems like `NNNN` part not just confusing as we already had same number for different RFC and thus seems like one RFC is _continuation_ of another RFC, but also totally unnecessary, since we've got date/time there. Also we already have `<NAME>.md`, plain text in English inside and `grep` or `awk` or `perl`. So I would understand some sort of `Keywords:` field in the body, that certainly would make sense in terms of finding something more fast and precise, as long as we have big enough library of those and the filenames no longer efficient. But I doubt we expect ten thousands of RFCs either in next 10 years...

damon-atkins:
PS I have no opinion on with or without a number, just offer a possible solution if a any number is required at all.

isbm:
@cachedout anyways, if this is useful, let's merge. Otherwise close. I am fine with both. :wink:

dwoz:
@cachedout @isbm I am willing to take on the responsibility of making sure RFC id numbers remain consistent.

isbm:
> @cachedout @isbm I am willing to take on the responsibility of making sure RFC id numbers remain consistent. Perhaps I can figure out a git hook to help with this like @isbm suggested.

I am here only asking myself how I would going to do that... 

- File creation time is obviously not working:

```
>>> for f in os.listdir('.'):
...     print(os.path.getctime(f))
... 
1542902944.2764814
1542902944.2764814
1542907407.1942673
1542907023.1594484
1542907023.1594484
1542907023.1594484
1542907023.1594484
1542907023.1594484
```

- Last modified time will always reshuffle numbers, rendering them completely useless.
- Parsing `Start Date` is free form — parsing lots of fun of doing that. It is not.

And for what all that pain?..

cachedout:
I think the thing to do here is to close this for now. If this becomes unmanageable in the future then we can revisit it. Since @isbm says he is fine with this being closed, I will do so now.

PR #50553, opened 2018-11-17
Batch-mode SLS syntax

isbm:
### What does this PR do?

Introduces batch-mode syntax for SLS. Please [read the RFC here](https://github.com/isbm/salt/blob/isbm-rfc-simpler-sls-syntax/rfcs/sls-alternative-syntax.md) for more information.


isbm:
@cachedout the RFC you were asked for.

isbm:
@thatch45 could you please look at this?

thatch45:
While I think that this is a good idea, I think that your proposed loader based solution is highly questionable. This would be a good place to add this capability into the state compiler so that we can unravel this logic as a data structure.

isbm:
But as stated in the RFC, it IS in the state compiler, exactly as you say. Or you meant something else?

thatch45:
I guess I don't understand the `__call__` proposal. I also don't understand why we can't just compile down the syntax to low chunks

isbm:
@thatch45 

> I guess I don't understand the `__call__` proposal. 
> I also don't understand why we can't just compile down the syntax to low chunks

Mainly, because `e = mc2` means "errors = more code ** 2". :wink: 

We _already_ have this done: compiler works just right. So why redoing the same at big scale with lots of changes? I wouldn't do that. All is missing is a tiny bit to let the compiler do not throw "Ouch!, main function is missed!", but just substitute it to a default `__call__`, [like here this already done and works](https://github.com/saltstack/salt/blob/94edf6d238dda143a1d1339c7fdf4797c804d68c/salt/state.py#L2791-L2809). Second, is that `__call__` stub is just a function that is injected into _any_ module, as long as it is not there, so we preserve current rules of developing modules. Developers won't even notice anything. And `__call__` is called only when state asks so (i.e. you not defining main entry function as per RFC already described).

I am also using `OrderedDict` instead of just `dict` in compiler structures, so that gives me priority/ordering of the State steps just for free without special re-ordering. E.g.:

```yaml
foo:
  runme:
    func_1:
       - params
    func_2:
       - params
    func_3:
       - params
```

In current situation, all functions are stored in just a `dict` instance. And so then its order is not controlled. So if `func_2` depends on `func_1`, it might work, and might fail. Using OrderedDict doesn't impact performance (measured already), but keeps it in the order as in SLS, which might even obsolete the idea of [this interesting ordering](https://github.com/saltstack/salt/blob/94edf6d238dda143a1d1339c7fdf4797c804d68c/salt/state.py#L580-L612) in a future.

Benefits of all this are obvious: we winning of _not touching_ the entire compiler code and so the changes are  very minimal and absolutely not intrusive nor to existing SLS, neither to the modules and their development.

thatch45 (approved):
Ok, I think I am good to go here. Just make sure that you use the syntax in the RFC, not the syntax in your last comment. It still needs to be a list.

isbm:
@thatch45 ahh, you are right. Sorry, forgot that it has to be a list.

PR #50555, opened 2018-11-18
add the exit msg for salt-cli

rongzeng54:
### What does this PR do?
When someone press Crtl+C in the state execution module running, it may be that they find something is wrong.
At this time, a stop prompt is very useful.

No

### Previous Behavior
Only remind user how to look up the return data for this job later.

### New Behavior
If user ran the state execution module, added reminders on how to safely set up a state run to die.

### Tests written?
No

### Commits signed with GPG?
No


cachedout (approved):
This is a really nice idea. Thanks, @rongzeng54

salt/client/__init__.py
```diff
@@ -778,16 +800,21 @@ def cmd_cli(
...
-                    )
+                        'remaining minions will return upon completion.\n\n'
+                        'To look up the return data for this job later, run the '
+                        'following command:\n'
+                        'salt-run jobs.lookup_jid {0}'.format(self.pub_data['jid']))
+                    if self.target_data:
+                        exit_msg += (
+                            '\n\n'
+                            'To set up the state run to safely exit, run the following command:\n'
+                            'salt {0} state.soft_kill {1}'.format(self.target_data, jid))
```
cachedout: Do you mean `self.pub_data['jid']` here instead? I think `jid` is undef.

cachedout (changes requested):

salt/client/__init__.py
rongzeng54: > Do you mean `self.pub_data['jid']` here instead? I think `jid` is undef.

already fixed

cachedout (approved):

PR #50563, opened 2018-11-19
Fix inconsistency with `pkg.list_pkgs` when using `attr` on RHEL systems

meaksh:
### What does this PR do?
This PR fixes an issue on the `pkg.list_pkgs` method when using `attr` parameter on RHEL systems.

On past PRs we've introduced the use of `attr` on `pkg.list_pkgs` on the Zypper/Yum modules for collecting not only the package version string but instead getting the desired RPM attributes (like `epoch`, `install_date`, `release`, etc) in an unified and structured way.

So when using `attr` the output from `pkg.list_pkgs` will follow this specific format for `attr='release,arch'`):
```yaml
"package": [
  {
    "version": "value1",
    "release": "value2",
    "arch": "value3",
  },
  {
    "version": "value4",
    "release": "value5",
    "arch": "value6",
  }
]
```
As you can see, the format is meant to group all installed versions of the same package (with its RPM attributes) into a single key `package` which holds a list containing a dict with all attrs for each installed version.

### Previous Behavior

On SUSE systems there is no problem since packages with different architectures always have different names, like `glibc` and `glibc-32bit` :+1: 

But the problem is happening on RHEL systems, where the architecture is internally handled by Salt as part of the package name. This is leading into an inconsistent looking output from `pkg.list_pkgs` according with the proposed format when using `attr`:

```json
$ salt-call --local pkg.list_pkgs attr='arch,release' --output=json

...
        "glibc": [
            {
                "release": "1.212.el6", 
                "version": "2.12", 
                "arch": "x86_64"
            }
        ], 
        "glibc.i686": [
            {
                "release": "1.212.el6", 
                "version": "2.12", 
                "arch": "i686"
            }
        ],
...
```

### New Behavior

The expected output should consider `glibc` and `glibc.i686` as the same package and group them into the same dict key. The output should be:
```json
$ salt-call --local pkg.list_pkgs attr='arch,release' --output=json

...
        "glibc": [
            {
                "release": "1.212.el6", 
                "version": "2.12", 
                "arch": "i686"
            }, 
            {
                "release": "1.212.el6", 
                "version": "2.12", 
                "arch": "x86_64"
            }
        ], 
...
```

### Tests written?

Yes

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/modules/pkg_resource.py
+++ b/salt/modules/pkg_resource.py
@@ -319,22 +319,22 @@ def format_pkg_list(packages, versions_as_list, attr):
             requested_attr &= set(attr + ['version'] + ['arch'])
 
         for name in ret:
+            try:
+                _name, _arch = name.rsplit('.', 1)
+            except ValueError:
+                _name, _arch = None, None
+
             versions = []
-            pkgname = name
+            pkgname = None
             for all_attr in ret[name]:
                 filtered_attr = {}
                 for key in requested_attr:
-                    if all_attr[key]:
+                    if key in all_attr:
                         filtered_attr[key] = all_attr[key]
                 versions.append(filtered_attr)
-            try:
-                namepart, archpart = name.rsplit('.', 1)
-            except ValueError:
-                pass
-            else:
-                if archpart in [attrs['arch'] for attrs in versions]:
-                    pkgname = namepart
-            ret_attr.setdefault(pkgname, []).extend(versions)
+                if _name and filtered_attr.get('arch', None) == _arch:
+                    pkgname = _name
+            ret_attr.setdefault(pkgname or name, []).extend(versions)
         return ret_attr
 
     for name in ret:
--- a/salt/modules/pkg_resource.py
+++ b/salt/modules/pkg_resource.py
@@ -319,10 +319,10 @@ def format_pkg_list(packages, versions_as_list, attr):
             requested_attr &= set(attr + ['version'] + ['arch'])
 
         for name in ret:
-            try:
-                _name, _arch = name.rsplit('.', 1)
-            except ValueError:
-                _name, _arch = None, None
+            _parse_arch_from_name = __salt__.get('pkg.parse_arch_from_name', lambda pkgname: {'name': pkgname, 'arch': None})
+            name_arch_d = _parse_arch_from_name(name)
+            _name = name_arch_d['name']
+            _arch = name_arch_d['arch']
 
             versions = []
             pkgname = None
--- a/salt/modules/pkg_resource.py
+++ b/salt/modules/pkg_resource.py
@@ -311,22 +311,31 @@ def format_pkg_list(packages, versions_as_list, attr):
     '''
     ret = copy.deepcopy(packages)
     if attr:
+        ret_attr = {}
         requested_attr = set(['epoch', 'version', 'release', 'arch',
                               'install_date', 'install_date_time_t'])
 
         if attr != 'all':
-            requested_attr &= set(attr + ['version'])
+            requested_attr &= set(attr + ['version'] + ['arch'])
 
         for name in ret:
             versions = []
+            pkgname = name
             for all_attr in ret[name]:
                 filtered_attr = {}
                 for key in requested_attr:
                     if all_attr[key]:
                         filtered_attr[key] = all_attr[key]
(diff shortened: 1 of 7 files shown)


salt/modules/pkg_resource.py
```diff
@@ -311,22 +311,31 @@ def format_pkg_list(packages, versions_as_list, attr):
...
-        return ret
+            try:
+                namepart, archpart = name.rsplit('.', 1)
+            except ValueError:
+                pass
+            else:
+                if archpart in [attrs['arch'] for attrs in versions]:
+                    pkgname = namepart
+            ret_attr.setdefault(pkgname, []).extend(versions)
+        return ret_attr
```
isbm: This loop over `attrs` is not even needed. Try/except here is also overkill. And `pass` in try/except is just not nice. I would write it differently (and shorter): 

```python
try:
    _name, _arch = name.rsplit('.', 1)
except ValueError:
    _name, _arch = None, None

for all_attr in ret[name]:
    pkgname, filtered_attr = None, {}
    for key in requested_attr:
        if all_attr.get(key):
            filtered_attr[key] = all_attr[key]
    versions.append(filtered_attr)
    if _name and filtered_attr.get('arch', None) == _arch:
        pkgname = _name
    ret_attr.setdefault(pkgname or name, []).extend(versions)
```

This is also covering bug you've missed when `"foo."` is splitted as `['foo', '']`. If `attrs['arch']` is an empty string, you're in trouble. While it is untested, you've got the picture. :wink:

isbm (changes requested):

meaksh:
@isbm this should be ready now :+1:

isbm (approved):

terminalmage (commented):
This is the wrong approach, IMO. This is a yum-specific approach, and will break `pkg` states on distros which use APT-based package managers. This is because our current approach with regard to 32-bit packages on 64-bit distros is to match the package manager's naming convention. That means `pkgname.i686` when using yum/dnf, but `pkgname:i686` when using apt.

Moreover, if we're going to make a change to how `pkg.list_pkgs` works, it should coincide with a different approach in states as well, one where we pass the architecture as its own parameter and not in the package name. Any such change should be made in the `develop` branch, and _not_ in a point release, and it should be made in such a way that we ensure it works on all package managers.

isbm (changes requested):
Ohh, @terminalmage, my fault approving this. Didn't noticed this is against `2018.3`, sorry. While I am not entirely sure this has that much impact, yet it indeed has to be against `develop` branch at first place. 

But seems the roots of these sort of issues are related to [this RFC](https://github.com/saltstack/salt/pull/45948). Seems like it was an attempt by @meaksh to [follow this comment](https://github.com/saltstack/salt/pull/45948#issuecomment-417336326) if I am not mistaken. While it is OK, I'd say first we need the final agreement on this.

@cachedout @terminalmage so what do we do with [this RFC](https://github.com/saltstack/salt/pull/45948)? I find @terminalmage 's [proposal](https://github.com/saltstack/salt/pull/45948#issuecomment-417336326) a nice, acceptable compromise. Can we all agree on that?

If we agreed on the RFC, then we should take care of states too, so not breaking anything, as @terminalmage pointed out.

terminalmage:
I think the format change for pkg.list_pkgs as @isbm and I discussed privately (and linked to [here](https://github.com/saltstack/salt/pull/45948#issuecomment-417336326)) is the right thing to do. However, this needs to be implemented across _all_ `pkg` virtual modules. And like I said in my review, we need accept the architecture as an argument (probably called `arch`) which must be properly handled by all the relevant `pkg` virtual modules in their `install`, `remove`, `purge`, etc. functions. Even those package managers which don't have a concept of alternate architectures (i.e. packages which don't match the OS arch) will need to _accept_ the argument and log something if the value is not `None`.

meaksh:
Hey @cachedout, I'd like to do it as soon as I have some free cycles to work on this. If you don't mind, please leave this here for some more time. Thanks :+1:

cachedout:
@meaksh No problem at all. Thanks for the update.

meaksh:
@isbm @terminalmage - I think this looks better now. I've also created the new PR to `develop`.

Closing this in favor of https://github.com/saltstack/salt/pull/51108

PR #50575, opened 2018-11-19
[doc] Nexus install doc

mikewiebe:
### What does this PR do?

Documentation for installing and configuring NX-OS proxy and native GuestShell minions.

### Commits signed with GPG?

Yes


cachedout (changes requested):
Can you also modify the `index.rst` in the same directory to reference these new instructions?

cachedout (approved):

PR #50583, opened 2018-11-20
Fix lint only changes, full lint on merge forwards

damon-atkins:
### What does this PR do?
* lint only changes, previous diff picked up out of data files, when the branch was out of date.
* full lint on merge forward to pick up changes in the lint checks between versions.

* Errors with pylint itself were not picked up, resulting in a pass.

### Previous Behavior
* When a branch was out of date, more files were linted than required.

### New Behavior
* Lint only the files changes in the PR
* If the lint of "change files only" passed, Lint all files as well if a merger forward i.e. branch name starts with merge-

### DIFF
--- a/.ci/lint
+++ b/.ci/lint
@@ -3,7 +3,7 @@ pipeline {
     options {
         timestamps()
         ansiColor('xterm')
-        timeout(time: 1, unit: 'HOURS') 
+        timeout(time: 3, unit: 'HOURS') 
     }
     environment {
         PYENV_ROOT = "/usr/local/pyenv"
@@ -14,7 +14,7 @@ pipeline {
         stage('github-pending') {
             steps {
                 githubNotify credentialsId: 'test-jenkins-credentials',
-                    description: 'Testing lint...',
+                    description: 'Python lint on changes...',
                     status: 'PENDING',
                     context: "jenkins/pr/lint"
             }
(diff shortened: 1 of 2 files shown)


.ci/lint
```diff
@@ -3,7 +3,7 @@ pipeline {
     options {
         timestamps()
         ansiColor('xterm')
-        timeout(time: 1, unit: 'HOURS') 
+        timeout(time: 3, unit: 'HOURS')
```
dubb-b: Just curious if you ever saw any go beyond an hour?

.ci/lint
```diff
@@ -41,52 +43,117 @@ pipeline {
...
                     }
                     steps {
                         sh '''
                         eval "$(pyenv init - --no-rehash)"
-                        grep -Ei '^salt/.*\\.py$|^setup\\.py$' file-list-changed.log | xargs -r '--delimiter=\\n' tox -e pylint-salt | tee pylint-report-salt.log
+                        # tee makes the exit/return code always 0
+                        grep -Ei '^salt/.*\\.py$|^setup\\.py$' file-list-changed.log | (xargs -r '--delimiter=\\n' tox -e pylint-salt;echo"$?" >  pylint-salt-chg.exit) | tee pylint-report-salt.log
                         # remove color escape coding
                         sed -ri 's/\\x1B\\[([0-9]{1,2}(;[0-9]{1,2})?)?[m|K]//g' pylint-report-salt.log
+                        read rc_exit < pylint-salt-chg.exit
```
dubb-b: Im not sure that these are going to work on our build servers.

Why did you choose this command?

.ci/lint
damon-atkins: Because its got a full lint option.  How long does a full lint take ?

.ci/lint
damon-atkins: The above reads the first line of a file.   `read` is just part of the shell commands like `if` i.e.
```
-bash-4.1$ help read
read: read .....
    Read a line from the standard input and split it into fields.
```
Its a lot better method than `rc_exit=$(head -1 pylint-salt-chg.exit)`   # and read removes the line feed

dubb-b: I was referring to rc_exit. Not sure that is available in Cent7. Isn't that a Suse specific command?

.ci/lint
dubb-b: Gottcha, they didn't take that long although it may be a safe default. I need to ask the team here if they are wanting to do a full lint on merge forwards. I'd imagine they will.

.ci/lint
dubb-b: Going to test this out on your PR and see what we get.

dubb-b: Or are you just using that as a variable?

dubb-b: NM I see it's becoming a variable.

.ci/lint
```diff
@@ -41,52 +43,117 @@ pipeline {
...
+                stage('lint salt') {
                     when {
                         expression { return readFile('file-list-changed.log') =~ /(?i)(^|\n)(salt\/.*\.py|setup\.py)\n/ }
                     }
                     steps {
                         sh '''
                         eval "$(pyenv init - --no-rehash)"
-                        grep -Ei '^salt/.*\\.py$|^setup\\.py$' file-list-changed.log | xargs -r '--delimiter=\\n' tox -e pylint-salt | tee pylint-report-salt.log
+                        # tee makes the exit/return code always 0
+                        grep -Ei '^salt/.*\\.py$|^setup\\.py$' file-list-changed.log | (xargs -r '--delimiter=\\n' tox -e pylint-salt;echo"$?" >  pylint-salt-chg.exit) | tee pylint-report-salt.log
```
dubb-b: https://jenkinsci.saltstack.com/job/pr-lint/job/PR-50583/6/

`echo"$?"` This needs to be fixed.

dubb-b:
@damon-atkins after talking to @cachedout and a few others we may have decided that we only want to do the changes and not the full lint on a merge forward.

@cachedout if you change your mind please let me know. The full lint would only run on a merge forward PR and only if the changes passes. It may be handy to have it run only in that case. I'd like your or @saltstack/team-core 's input here.

damon-atkins:
lint has coding rules, which change from one release to the next. The idea is to try and make sure those changes are picked up. Hence it only happens if the branch starts with `merge-` and only if the change files pass also.  i.e for a `merge-*` branch, lets check the changes and if passed make extra sure the rest of the code is ok as well, for any other branch name else just check the changes.

Happy to pull the code. Just thought it might help the quality of the releases.

The original code which listed out the changes detected "files out of date" when a branch is old. This is fixed in this PR after running some experiments to determine the best git diff args to use.

One choice it to try it out and pull code if it becomes an issue to the merge forward process.

cachedout:
I am fine with trying this out for a while if @dubb-b is.

dubb-b:
I am fine with this.

Proceed with merge if no one has objections.

damon-atkins:
Hi @dubb-b can you approve the PR please.

dubb-b (approved):

PR #50585, opened 2018-11-20
CacheDisk enhancements and bugfixes

isbm:
### What does this PR do?

- Enhances API for all caches (currently CacheDict and CacheDisk)
- Provides `store` function that will trigger persistence manually (in case implementation has one)
- Adds better logging
- "Todoes" TODO hacks
- Guards against IO/OS crashes
- other minor improvements

This all started here as @rongzeng54 spotted a bug, where CacheDisk is not storing pillar data, when its _content_ is changed: https://github.com/saltstack/salt/pull/50179

### New Behavior

- Mentioned bug is fixed
- No crashes on e.g. different permissions (happens when master was previously run as `root`, then as `salt`).
- Everything else should work as before

### Tests written?

- Refactored existing
- Added new

### DIFF
--- a/salt/utils/cache.py
+++ b/salt/utils/cache.py
@@ -43,7 +43,15 @@ class CacheFactory(object):
             log.error('CacheFactory received unrecognized cache type')
 
 
-class CacheDict(dict):
+class CacheAPI(dict):
+    def store(self):
+        '''
+        Store data in the cache persistence.
+        :return:
+        '''
+
+
+class CacheDict(CacheAPI):
     '''
     Subclass of dict that will lazily delete items past ttl
     '''
--- a/salt/utils/cache.py
+++ b/salt/utils/cache.py
@@ -115,6 +115,13 @@ class CacheDisk(CacheDict):
         self._enforce_ttl_key(key)
         return self._dict.__contains__(key)
 
+    def __repr__(self):
+        '''
+        Represent CacheDisk.
+        :return:
+        '''
+        return '<{name} of {length} entries at {memaddr}>'.format(
+            name=self.__class__.__name__, length=len(self), memaddr=hex(id(self)))
     def __getitem__(self, key):
         '''
         Check if the key is ttld out, then do the get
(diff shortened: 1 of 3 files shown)


salt/utils/cache.py
```diff
@@ -130,41 +167,50 @@ def __delitem__(self, key):
...
-            }
-            msgpack.dump(cache, fp_, use_bin_type=True)
+        if msgpack is None:
+            log.error('Cache cannot be read from the disk: msgpack is missing')
+        elif not os.path.exists(self._path):
+            log.debug('Cache path does not exist for reading: %s', self._path)
+        else:
+            try:
+                with salt.utils.files.fopen(self._path, 'rb') as fp_:
+                    cache = salt.utils.data.decode(msgpack.load(fp_, encoding=__salt_system_encoding__))
```
cachedout: This looks like it will stacktrace if msgpack is missing.

cachedout (changes requested):

salt/utils/cache.py
isbm: Hmm, how? Line #176 will just log error about it and return `None`. Unit tests are also explicitly testing this [here](https://github.com/saltstack/salt/pull/50585/files#diff-9fc11a5086d51e0723e81f3cf065bf0aR171) and implicitly [here](https://github.com/saltstack/salt/pull/50585/files#diff-9fc11a5086d51e0723e81f3cf065bf0aR374) and [here](https://github.com/saltstack/salt/pull/50585/files#diff-9fc11a5086d51e0723e81f3cf065bf0aR362). Am I still missing something?

brejoc: I think this should be fine. We are only entering else if `msgpack` is not `None`.

DmitryKuzmenko (approved):

salt/utils/cache.py
cachedout: Wow, what was I thinking there? My fault. Apologies.

cachedout (approved):

Issue #50562, opened 2018-11-19
scheduled highstate ignores blackout


soostdijck:
### Description of Issue/Question
A scheduled job that runs state.highstate ignores blackout mode. I'm 99% sure this is a change since upgrading from 2018.3.2

### Setup
```
$ salt 'parrot*' schedule.add test_blacko function='state.highstate' seconds=20
parrot.ripe.net:
    ----------
    comment:
        Added job: test_blacko to schedule.
    result:
        True
$ salt 'parrot*' grains.setval minion_blackout true
parrot.ripe.net:
    ----------
    minion_blackout:
        True
```
Wait for highstate...
```
$ salt 'parrot*' grains.item minion_blackout

parrot.ripe.net:
    ----------
    minion_blackout:
        False
```

p.s. In this case the blackout grain is being unset by my something in the highstate, this is expected behaviour because the in highstate I have a stanza like this:
```
salt-minion-blackout:
  grains.present:
    - name: minion_blackout
    - value: false
```
 Of course the issue is that the highstate should not be running in the first place...

### Versions Report
Both master and minions are on:
```
Salt Version:
           Salt: 2018.3.3

Dependency Versions:
           cffi: 1.6.0
       cherrypy: unknown
       dateutil: 1.5
      docker-py: Not Installed
          gitdb: 0.6.4
      gitpython: 1.0.1
          ioflo: Not Installed
         Jinja2: 2.7.2
        libgit2: 0.26.3
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.5.1
   mysql-python: Not Installed
      pycparser: 2.14
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: 0.26.4
         Python: 2.7.5 (default, Apr 11 2018, 07:36:10)
   python-gnupg: Not Installed
         PyYAML: 3.11
          PyZMQ: 15.3.0
           RAET: Not Installed
          smmap: 0.9.0
        timelib: Not Installed
        Tornado: 4.2.1
            ZMQ: 4.1.4

System Versions:
           dist: centos 7.5.1804 Core
         locale: UTF-8
        machine: x86_64
        release: 3.10.0-862.3.3.el7.x86_64
         system: Linux
        version: CentOS Linux 7.5.1804 Core
```

garethgreenaway:
[...]

PR #50586, opened 2018-11-20
[2018.3] minion_blackout for scheduled jobs

garethgreenaway:
### What does this PR do?
Adding support for minion_blackout for jobs that are run from the Salt scheduler.

#50562 

### Tests written?
Yes

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/utils/schedule.py
+++ b/salt/utils/schedule.py
@@ -583,6 +583,7 @@ class Schedule(object):
         '''
         Execute this method in a multiprocess or thread
         '''
+        log.debug('=== calling handle_func with data %s ===', data)
         if salt.utils.platform.is_windows() \
                 or self.opts.get('transport') == 'zeromq':
             # Since function references can't be pickled and pickling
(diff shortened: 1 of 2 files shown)


salt/utils/schedule.py
```diff
@@ -713,7 +734,6 @@ def handle_func(self, multiprocessing_enabled, func, data):
 
                 ret['success'] = True
 
-                data_returner = data.get('returner', None)
```
cachedout: I'm not following the reasoning for this change. Help? :)

garethgreenaway: @cachedout Definitely.  I moved it outside the try...except.  When an exception happens, eg. the new one I added for minion_blackout or if an invalid call happens, the code will drop into the `finally` section where data_returner is referenced.  In some situations it hasn't been created yet, so errors appear in the logs.

rares-pop (approved):
LGTM

salt/utils/schedule.py
cachedout: Makes perfect sense. Thank you!

cachedout (approved):

PR #50591, opened 2018-11-20
INFRA-7855 - allow boto_lambda.function_present() to support remote URLs for ZipFile param

tkwilliams:
Crazy simple change.

Specifically added to provide salt:// URL support, but will work with any URL scheme cp.cache_file() recognizes.


cachedout (approved):

terminalmage:

PR #50603, opened 2018-11-21
Remove no-longer-existant test module from whitelist.txt

terminalmage:
This was converted to unit tests in #50528. Since there are no unit tests in the whitelist.txt, I've removed it rather than renaming.

CC: @twangboy


twangboy:
@terminalmage Thanks. I have that removal as part of another PR that may or may not be merged.

PR #50607, opened 2018-11-21
file: update attributes for lsattr and chattr

aplanas:
### What does this PR do?

Update the different attributes that chattr and lsattr can understand. This include new attributes like C, that can disable the copy on write for btrfs subvolumes.

### Tests written?

No, the current ones cover the same case.

### DIFF
--- a/salt/modules/file.py
+++ b/salt/modules/file.py
@@ -570,7 +570,7 @@ def lsattr(path):
     for line in result.splitlines():
         if not line.startswith('lsattr: '):
             vals = line.split(None, 1)
-            results[vals[1]] = re.findall(r"[acdijstuADST]", vals[0])
+            results[vals[1]] = re.findall(r"[aAcCdDeijPsStTu]", vals[0])
 
     return results
 
@@ -587,8 +587,8 @@ def chattr(*files, **kwargs):
         should be added or removed from files
 
     attributes
-        One or more of the following characters: ``acdijstuADST``, representing
-        attributes to add to/remove from files
+        One or more of the following characters: ``aAcCdDeijPsStTu``,
+        representing attributes to add to/remove from files
 
     version
         a version number to assign to the file(s)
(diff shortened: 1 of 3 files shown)


cachedout:
@aplanas It looks like this may be causing some failures on Windows. Could you take a look?

aplanas:
@cachedout Makes me wonder how a code that is excluded in Windows can make a Windows test fail:

https://github.com/saltstack/salt/blob/c8eb3f88a3f972201998fa8859b74936157319de/salt/modules/file.py#L79

The failing tests are most likely related with win_file.py, but my knowledge on this platform makes me a bit useless here.

salt/modules/file.py
```diff
@@ -6243,6 +6243,7 @@ def list_backups(path, limit=None):
         [files[x] for x in sorted(files, reverse=True)[:limit]]
     )))
 
+
```
cachedout: Can we pull newline out please? I'm not sure why it was added.

salt/modules/file.py
```diff
@@ -6415,6 +6416,7 @@ def delete_backup(path, backup_id):
 
     return ret
 
+
```
cachedout: Can we pull newline out please? I'm not sure why it was added.

cachedout (changes requested):

salt/modules/file.py
aplanas: Sure, but this is needed. Comes from a flake8 / pep8 complaint (E305: E305 expected 2 blank lines after class or function definition). After each function we need to add two lines. Looks like that there is no exception for assigments.

salt/modules/file.py
aplanas: same, E305

Is it OK to ignore it?

aplanas:
* Rebased the code
* Fix the documentation in states/file.py and states/netconfig.py

cachedout (approved):

Issue #50536, opened 2018-11-16
ini.options_present separator add extra whitespaces (and pkgrepo is not viable to manage yum)


Poil:
Hi,

When using ini.options_present with a defined separator, salt add extra whitespace around the separator.

If I apply this code, all the parameters of the file are converted from `key=value` to `key = value`
Also I do this because, it is not possible with pkgrepo to only disable an existant repo, you have to set all the parameters and you can't manage a yum repo configuration file with multiple entries (one file, one entry); so when the centos-release package is updated all the entries are recreated; so the only way I've found to have something not dirty when using a local YUM mirror, is to disable all the default repository and to manage each entry of the local mirror via pkgrepo.

```
yum_disable_centos_base_repo:
  ini.options_present:
    - name: /etc/yum.repos.d/CentOS-Base.repo
    - separator: '='
    - sections:
        base:
          enabled: 0
        updates:
          enabled: 0
        extras:
          enabled: 0
        centosplus:
          enabled: 0

CentOS-Base:
  pkgrepo.managed:
    - humanname: 'CentOS-$releasever - Base'
    - name: local-base
    - baseurl: 'https://repos.local/pulp/repos/centos/7/os/$basearch/'
    - enabled: 1
    - proxy: _none_

CentOS-Updates:
  pkgrepo.managed:
    - humanname: 'CentOS-$releasever - Updates'
    - name: local-updates
    - baseurl: 'https://repos.local/pulp/repos/centos/7/updates/$basearch/'
    - enabled: 1
    - proxy: _none_

CentOS-Extras:
  pkgrepo.managed:
    - humanname: 'CentOS-$releasever - Extras'
    - name: local-extras
    - baseurl: 'https://repos.local/pulp/repos/centos/7/extras/$basearch/'
    - enabled: 1
    - proxy: _none_
```

Ch3LL:
can you please include the output of `salt --versions-report`

Poil:
Hi,

```
Salt Version:
           Salt: 2018.3.3
 
Dependency Versions:
           cffi: 1.11.5
       cherrypy: unknown
       dateutil: Not Installed
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.10
        libgit2: 0.26.5
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.5.6
[...]

PR #50614, opened 2018-11-22
Avoid leaving an empty first line on ini edits

boltronics:
### What does this PR do?
This change prevents Salt creating an empty new line to the start of an ini file whenever it first makes an edit.

This is related to issue https://github.com/saltstack/salt/issues/50536 and discovered when writing PR https://github.com/saltstack/salt/pull/50613.

### Previous Behavior
Currently Salt's `ini_manage` execution module writes sections typically like this:
`\n[section name]\n`

Since ini files typically lead with a section, this would cause the file to look like this:
```
$ cat somefile.ini

[section name]
some_option: 1
$
```

### New Behavior
The new behaviour is to avoid writing an initial empty line. eg.
```
$ cat somefile.ini
[section name]
some_option: 1
$
```

### Tests written?
No. Existing tests pass for me.

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/modules/ini_manage.py
+++ b/salt/modules/ini_manage.py
@@ -449,7 +449,11 @@ class _Ini(_Section):
             with salt.utils.files.fopen(self.name, 'wb') as outfile:
                 ini_gen = self.gen_ini()
                 next(ini_gen)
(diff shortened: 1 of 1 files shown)


cachedout (approved):
LGTM

Issue #45867, opened 2018-02-05
Icinga2 Module uses incorrect paths


Nick2253:
### Description of Issue/Question
Icinga2 changed the paths for certificates starting in version 2.8.  The paths are hardcoded in the Icinga2 module, and refer to the paths for versions 2.7 and prior.

https://www.icinga.com/docs/icinga2/latest/doc/16-upgrading-icinga-2/#upgrading-to-2-8-certificate-paths

### Suggestion to Fix
We could update the module and hardcode the new paths.  However, this could break installs for old versions.  In a perfect world, we'd be able to determine the paths automatically based on the version number.

Another option is to give the user control of the path (either create a different version of the module for the new versions, create a flag for the new paths, or an option to directly select the path).  I'm not a fan of this option because it requires the user to micromanage the command/state as clients are upgraded (and is therefore no longer portable).  This is especially problematic in a mixed environment.

I'm pretty comfortable doing the fixing, but I'm not sure what the best direction is.  If we do want to autodetect the version, a canonical example of a different module that does it would be helpful for best practices.

### Versions Report
This is an issue for the current version, and appears to still be an issue in Oxygen.  As of the latest Oxygen commit in Github, the old paths were still hardcoded.

garethgreenaway:
@Nick2253 Thanks for the report.  The paths should definitely be configurable with defaults based one standard locations.  If there is a way to determine the version of icinga2 installed, perhaps with a CLI argument to a binary, that would be one way to determine the appropriate default locations by version.  And we would definitely welcome a PR with this fix.  Thanks!

Nick2253:
I'm working through the changes here, and I just had a quick question about best practices.  I can get the version directly from the icinga binary, but I'm not sure where I should put the logic to get this version and determine the default paths.  Should that go in the __virtual___ function?

PR #50615, opened 2018-11-22
Fixes icinga2 certs path for newer versions 2.8+

ClaudiuLSH:
### What does this PR do?

Bug fixing.

#45867

### Previous Behavior

Icinga2 module and state worked only with versions prior to 2.8 for new deployments.
It also worked for versions newer than 2.8 but if upgraded from older versions without following the official documentation.

### Tests written?

No

### Commits signed with GPG?

No

### Notes

My issues so far:
1. Documentation refers to old certs path and now the path is selected based on icinga2 version. How should the docs be written?
2. If the server was deployed with icinga2 prior to 2.8 and upgraded without following the documentation, this change will break compatibility. This scenario is a bit messy. If configuration files are not updated, certs path is kept like before but new certs will be deployed in new path. So icinga2 will fail to register new clients. Again this will happen only if the upgrade was done without following the documentation;
3. I moved some common parts to salt/utils/icinga2.py so this is a new file. This is not a big change but since it is a bug fix I am not sure if all my changes are acceptable.
4. I created this pull request for branch 2018.3.3 since this should be a bug fix, is it ok?

Tomorrow I will fully test all this changes on an older server following upgrade documentation.

### DIFF
--- a/salt/utils/icinga2.py
+++ b/salt/utils/icinga2.py
@@ -7,8 +7,8 @@ This module provides common functionality for icinga2 module and state.
 
 .. versionadded:: 2018.8.3
 '''
-
 # Import python libs
+from __future__ import absolute_import, print_function, unicode_literals
 import logging
 import re
 
@@ -20,10 +20,9 @@ log = logging.getLogger(__name__)
 
 def get_certs_path():
     icinga2_output = __salt__['cmd.run_all']([salt.utils.path.which('icinga2'), "--version"], python_shell=False)
-    version = re.search('r\d+\.\d+', icinga2_output['stdout']).group(0)
+    version = re.search(r'r\d+\.\d+', icinga2_output['stdout']).group(0)
     # Return new certs path for icinga2 >= 2.8
     if int(version.split('.')[1]) >= 8:
         return '/var/lib/icinga2/certs/'
     # Keep backwords compatibility with older icinga2
     return '/etc/icinga2/pki/'
-
--- a/salt/utils/icinga2.py
+++ b/salt/utils/icinga2.py
@@ -10,7 +10,6 @@ This module provides common functionality for icinga2 module and state.
 
 # Import python libs
 import logging
-import subprocess
 import re
 
 # Import Salt libs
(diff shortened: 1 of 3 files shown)


salt/utils/icinga2.py
```diff
@@ -0,0 +1,42 @@
...
+import subprocess
+import re
+
+# Import Salt libs
+import salt.utils.path
+
+log = logging.getLogger(__name__)
+
+
+def execute(cmd, ret_code=False):
```
cachedout: WIth this, would it perhaps be better just to use Salt's `cmd.run` or `cmd.retcode` module functions that already exist?

ClaudiuPID: Will see if cmd.retcode works, I just realized I did not test that one.

cachedout: Thanks. If we can figure out the exact issue with the `cmd` module, it would be great to have an issue filed against that so we can get it fixed as well. Thanks.

ClaudiuPID:
This worked ok on my old server after I upgraded it.
One issue with debian clients, new certs path is not created by default so one should create it through salt.

cachedout (approved):

ClaudiuPID:
Sure, will push tomorrow.

cachedout:
@ClaudiuPID Thanks. Once those changes are made we should be able to get this in.

ClaudiuPID:
Fixed.

dwoz:
@ClaudiuPID Looks like the linter is still un-happy.

https://jenkinsci.saltstack.com/job/pr-lint/job/PR-50615/3/warnings52Result/new/

ClaudiuPID:
I don't think those are caused by my pull request. I only changed icinga2.py files, none of those reported there.

Ch3LL:
@ClaudiuPID the 2018.3.3 branch was deleted yesterday since its a release branch and its been released. It looks like github automatically closed it since I deleted the branches. Can you push to 2018.3 instead?

ClaudiuPID:
@Ch3LL for some reason when I try to create a new pull request on github it shows branch 2018.3 but not 2018.3.3.
When I try to enter that branch manually it shows:
 Choose different branches or forks above to discuss and review changes. 

Ideas?

Ch3LL:
Yes you want to select 2018.3 not 2018.3.3. The 2018.3.3 branch is deleted because that release has been released. Please submit to the 2018.3 branch please.

PR #50617, opened 2018-11-22
Capabilities RFC

isbm:
### What does this PR do?

Describes presented on Salt Conference 2018 capabilities.

Please [read it here](https://github.com/isbm/salt/blob/isbm-rfc-capabilities/rfcs/minion-capabilitles.md).

### DIFF
--- a/rfcs/minion-capabilitles.md
+++ b/rfcs/minion-capabilitles.md
@@ -143,6 +143,22 @@ Also iterations works the same way:
 {% endif %}
 ```
 
+...or hashes:
+
+```jinja2
+{% if capable.who.knows.what.is.this['something'] %}
+  ...
+{% endif %}
+```
+
+...or even crazy hashes:
+
+```jinja2
+{% if capable.who['knows'].what['is'].this['something'].there() %}
+  ...
+{% endif %}
+```
+
 ## Bonus Feature
 
 The `capable` variable essentially can replace-or-help to `grains`,
--- a/rfcs/minion-capabilitles.md
+++ b/rfcs/minion-capabilitles.md
@@ -1,0 +1,201 @@
+- Feature Name: Minion Capabilitles
+- Start Date: 2018-11-22
+- RFC PR: 
+- Salt Issue:
+- Keywords: capability capabilities introspect introspection
+
+# Summary
+[summary]: #summary
+
+Mechanism to know if a minion is capable of doing something specific,
+that wasnt known during the version update.
+
+# Motivation
+[motivation]: #motivation
+
+Unlike client-less systems, Salt has should deal with the client
+systems. These clients supposed to be compatible with the states, sent
+from the master node. When updates are frequent, two versions
+backward-compatibility sorts out this problem naturally. But there is
+a number of use cases when this does not apply. Few examples:
+
+- if an enterprise system is not updated often, even Salt client stays with
+  old version or yet only required features are back-ported.
+
+- if a new feature wasn't released to the older version but only
+  back-ported by supporting vendor.
+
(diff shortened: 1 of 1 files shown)


isbm:
@cachedout bumpitybump!

dwoz (approved):
I like this. Long term maybe we can combine it with some parameter spec decorators that could handle input validation and normalization. Assuming others agree, we'll probably want that as a separate RFC. 
I approve of this one as it stands now.

isbm:
@dwoz that is certainly another RFC. The input validation/normalisation is another topic and mostly relevant to [the Interface RFC](https://github.com/saltstack/salt/pull/50353). On which you are also welcome to comment. :wink:

cachedout (approved):
I think this is a very smart way to do this. One thought might be that we could perhaps simplify this even further by creating a custom jinja filter?

thatch45:
As much as I don't like having deeply nested objects like this, I do think that this is a good overall design. If there is a way to make this not as nested I would like it more.

cachedout:
@isbm Perhaps we could explore a way to "reserve" shortcuts somewhere in Salt to avoid clashes. It would be good if we could make these as short as possible while still avoiding potential clashes.

isbm:
But how? You need to start somewhere. Either it is a module or state. Then you have to refer to a function **of** that module. Then you should select what area you want to explore and so on. You cannot just say "reboot me a machine if 'foo' in documentation". Because Salt has absolutely no idea what exactly is that you want from it.

Having shortcuts will make it harder because one will need to remember many things instead of one.

Having a single entry also allows us to extend it to other areas I already wrote in the RFC as well as even different places yet to come. We are not going to pollute namespace with the new reserved words, being afraid to screw up someone's SLS.

whytewolf:
I love the idea of this. But I did want to note. That the "bonus feature" of salt.module.function already exists in salt. https://docs.saltstack.com/en/latest/topics/jinja/index.html#calling-salt-functions

https://github.com/saltstack/salt/blob/develop/salt/renderers/jinja.py#L23-L42

However, if dropped in favor of this I would not complain. 

Also if this augmented the sysmod functionality that would be a good bonus. 
https://docs.saltstack.com/en/latest/ref/modules/all/salt.modules.sysmod.html

isbm:
> I love the idea of this. But I did want to note. That the "bonus feature" of salt.module.function already exists in salt. https://docs.saltstack.com/en/latest/topics/jinja/index.html#calling-salt-functions
> 
> https://github.com/saltstack/salt/blob/develop/salt/renderers/jinja.py#L23-L42

Syntax-wise yes. But the way I propose (and POC does already) is basically this:

```python
salt().module().function()
```

So it will load only when you _actually_ access it by calling and loading. Some internal Python tricks allows to get rid of `()` syntax, so it looks to the user like he would be just accessing already pre-loaded structure. However, this bonus feature was only an abstract example how we can build something on top of this principle.

Needless to say, this is not any new. The entire [Zope](http://www.zope.org/en/latest/) is basically built on top of that. :wink: 

> Also if this augmented the sysmod functionality that would be a good bonus.
> https://docs.saltstack.com/en/latest/ref/modules/all/salt.modules.sysmod.html

Primarily we could get rid of `foo['key']['otherkey']['somethingelse']` syntax everywhere in SLS. Especially when you want something like to make sure it won't fail in the middle:

```python
if foo.get('key', {}).get('otherkey', {}).get('somethingelse'):
    ...
```

Much nicer to do just:

```python
if foo.key.otherkey.somethingelse:
    ...
```

But as I said, this is all ideas for the future. Right now important is — capabilities.

PR #50622, opened 2018-11-23
Fix openvswitch_port.present internal=True option

smarsching:
### What does this PR do?

It fixes a bug in `salt.states.openvswitch_port.present` that caused the `internal=True` argument to be ignored when no `tunnel_type` was specified.

### Previous Behavior
When using the following state definition, an OpenVSwitch port would be added to the bridge, but no internal interface would be created for that port (meaning that it could not be used as a network interface in Linux):

```
my_state:
  openvswitch_port.present:
    - name: myport
    - bridge: mybridge
    - internal: True
```

### New Behavior
Now, an internal interface will be created (like it was already before, when specifying `tunnel_type: vlan`).

### Tests written?

No

### Commits signed with GPG?

No


cachedout:
@smarsching Test failures it seems: https://jenkinsci.saltstack.com/job/pr-kitchen-centos7-py2/job/PR-50622/1/testReport/junit/unit.states.test_openvswitch_port/OpenvswitchPortTestCase/test_present/

smarsching:
@cachedout I fixed the problem. The unit test did not provide the `openvswitch.get_interface_type` in the mock version of the `openvswitch` module so that the call to that function failed.
I changed the test to provide that function of part of the mock object and now the tests seem to run successfully.

dwoz (approved):

PR #50627, opened 2018-11-24
states.file: Adding a keyvalue function

teizz:
### What does this PR do?

This PR adds a new function to the file state-module, including documentation of all its parameters. It also includes the addition of an integration test to test the new behaviour added. 

No open issues specifically. 

### New Behavior

Using the new keyvalue_list function (or its keyvalue shorthand version), it's possible to make and maintain changes to files. Some of the features of this function are:

- uncommenting default lines, so all settings don't show up at the top/bottom of a file.
- removing superfluous lines which might have otherwise overridden a setting.
- matching either (or both) the key or value part of a setting case-insensitively as to not make unnecessary changes.
- append or prepend a key/value setting if it wasn't found in the file.
- fully supports a test=True run for detecting potential changes before actually committing them.

### Tests written?

Yes. There is an additional integration test which checks if a commented out line is commented in and has its value updated to reflect the requested setting.

### Commits signed with GPG?

No

### Rationale

Although it is already possible to (un)comment lines, append text, and replace lines through other functions in the file state-module, in some cases it would result in an sls file that's hard to read or a not-quite satisfactory result.

An example. Lots of products come with default configuration files which have the defaults commented out. Not all defaults might be in there though.

- `file.replace` used to be my go-to function, it is capable appending or prepending if the line is not found, can do regex, accepts regex-flags to do case-insensitive matching etc. However, unless one want to go for negated forward lookups, a line would always match, always get replaced and thus always trigger a change.
- `file.append`, and `file.blockreplace` can't really insert at an arbitrary location.
- `file.line` which got added in Beryllium won't replace a line which matches but already has the replacement content, and handles regex. But it can't easily match case-insensitively, and needs a location in order to 'ensure' a line exists.

I've been leaving a trail behind at different companies implementing the functionality in this PR in custom '_states' directories. All in different states of bug-fixing by now I'm sure.
During Saltconf18 talk of this came up and it was suggested I'd try to submit it 'upstream', since it would also be helpful in some of the Salt for Secops check/remediation states.


teizz:
I've gone ahead an put 'versionadded' as 'Neon' for a placeholder in the docs, I suppose that'd need updating should this PR get merged.

teizz:
The windows builds that are failing seem to err on 3 tests unrelated to this PR.
- integration.states.test_file.FileTest.test_file_managed_onchanges
- integration.states.test_file.FileTest.test_file_managed_prereq
- integration.states.test_file.FileTest.test_test_managed

The windows2016-py3 seems to also err on an additional file.managed test:
- integration.states.test_file.FileTest.test_managed_unicode_jinja_with_tojson_filter

I suspect they might have been failing at the last merge touching the file state module as well. At least merge of PR https://github.com/saltstack/salt/pull/49832 seems relevant to running in test, but I can't find even the last Jenkins build-log of when either the file state or test was touched.

salt/states/file.py
```diff
@@ -4570,6 +4570,372 @@ def replace(name,
     return ret
 
 
+def keyvalue_list(
```
cachedout: The name of this function feels a little strange to me. Would it be more straightforward if it were just `file.keyvalue`?

teizz: It is, and that's exactly what I'm doing in the other function below called exactly that. (https://github.com/saltstack/salt/pull/50627/files#diff-3b447dc835943beb1326f98f684fe783R4905-R4936)

 I'd written the first later, then figured it'd be nice to give a list. Since I didn't want to call the original function over and over I'd figured the shorthand version could instead call the list version.

I'd be happy calling it ``file.keyvalues``, and do away with the shorthand version though.

teizz (commented):
I think the additional commits resolve https://github.com/saltstack/salt/pull/50627#discussion_r236339886.

salt/states/file.py
cachedout: That makes a lot of sense. Excellent idea!

salt/states/file.py
```diff
@@ -4570,6 +4570,375 @@ def replace(name,
...
+        Append the key/value to the end of the file if not found. Note that this
+        takes precedence over ``prepend_if_not_found``.
+
+    prepend_if_not_found : False
+        Prepend the key/value to the beginning of the file if not found. Note
+        that ``append_if_not_found`` takes precedence.
+
+    search_only : False
+        Only search, do not actually replace. This is effectively a dry-run.
+        Functionally almost the same as running the state with test=True, which
```
cachedout: Does it make sense to have a separate flag then? What exactly distinguishes this from `test=True`?

teizz: You are absolutely right @cachedout ! Admittedly, I originally wrote this function as an execution module where the need for a ``search_only`` parameter made sense just like it does in the ``file.replace`` execution module.

Back then it was a nice way of seeing which minions would end up with changes and which had already had their config files updated. Ultimately it seemed to make a lot more sense to put this into a state module. I suppose I documented it's behaviour and never realised it really should be taken out.

I'll go ahead and clean this up.

cachedout: Great! Thanks!

cachedout (approved):

s0undt3ch:
Port to master https://github.com/saltstack/salt/pull/56770

Issue #49393, opened 2018-08-29
file.directory recurse not always changing perms


Poil:
### Setup
```
flxccs_fix_apache_errorpages:
  file.directory:
    - name: /data/apache/errorPages
    - user: root
    - group: root
    - file_mode: "0644"
    - dir_mode: "0755"
    - recurse:
      - user
      - group
      - mode
```

### Steps to Reproduce Issue
* Initial permissions
```
[root@flxccs-d03-tst errorPages]# ll
total 204
-rw------- 1 root root  4539 Aug 27 10:39 contact.html.var
-rw------- 1 root root  9941 Aug 27 10:39 HTTP_BAD_GATEWAY.html.var
-rw------- 1 root root  7112 Aug 27 10:39 HTTP_BAD_REQUEST.html.var
-rw------- 1 root root 12578 Aug 27 10:39 HTTP_FORBIDDEN.html.var
-rw------- 1 root root 14338 Aug 27 10:39 HTTP_GONE.html.var
(...)
drwxr-xr-x 2 root root  4096 Aug 27 10:39 include
-rw------- 1 root root  2125 Aug 27 10:39 README
```
* Running state.apply, no changes
```
  Name: /data/orange-apache/errorPages - Function: file.directory - Result: Clean Started: - 09:52:54.988513 Duration: 3.479 ms
```

* Changing manually the include directory permissions
```
[root@flxccs-d03-tst errorPages]# chmod a-x include
[root@flxccs-d03-tst errorPages]# ll
total 204
-rw------- 1 root root  4539 Aug 27 10:39 contact.html.var
-rw------- 1 root root  9941 Aug 27 10:39 HTTP_BAD_GATEWAY.html.var
-rw------- 1 root root  7112 Aug 27 10:39 HTTP_BAD_REQUEST.html.var
-rw------- 1 root root 12578 Aug 27 10:39 HTTP_FORBIDDEN.html.var
-rw------- 1 root root 14338 Aug 27 10:39 HTTP_GONE.html.var
(...)
drw-r--r-- 2 root root  4096 Aug 27 10:39 include
-rw------- 1 root root  2125 Aug 27 10:39 README
```
* Running state.apply, permissions are set correctly
```
----------
          ID: flxccs_fix_apache_errorpages
    Function: file.directory
        Name: /data/orange-apache/errorPages
      Result: True
     Comment: Directory /data/orange-apache/errorPages updated
     Started: 09:54:01.191557
    Duration: 175.081 ms
[...]

PR #50631, opened 2018-11-26
Check file_mode in file.directory and _check_directory

lexvella:
### What does this PR do?

It fixes a bug in salt.states.file.directory that caused the function to return prematurely without evaluating the mode of files in a directory to be recursively evaluated. This PR adds a check for the file_mode parameter in the _check_directory function used by salt.states.file.directory

file.directory recurse not always changing perms saltstack/salt#49393 

### Previous Behavior
_check_directory does not evaluate the mode of files, only the user and group.

### New Behavior
_check_directory evaluates the user, group, and mode of files.

### Tests written?

No

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/states/file.py
+++ b/salt/states/file.py
@@ -677,6 +677,7 @@ def _check_directory(name,
                      group=None,
                      recurse=False,
                      mode=None,
+                     file_mode=None,
                      clean=False,
                      require=False,
                      exclude_pat=None,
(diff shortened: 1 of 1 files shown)


cachedout:
@lexvella Thank you for this but can you please close this PR and re-open one for this change against the 2018.3 branch instead? That's the correct place for this PR. Thanks.

PR #50646, opened 2018-11-27
fix typo for default publish_port

GwiYeong:
### What does this PR do?
fix typo for default publish_port in `zeromq.py`
not match default value publish_port in config https://github.com/saltstack/salt/blob/develop/salt/config/__init__.py#L1508

### Tests written?

No

### Commits signed with GPG?

Yes


cachedout:
This seems right but I want @mattp- to just double-check as well. :)

mattp-:
yep this looks good - this accidentally worked because i think the cases where the master rehomes its own publish_port outside of what is set in self.opts.get('publish_port') is rare, overridden or default.

mattp-:
just occurs to me this also has a parallel in transport.tcp, ill send a pr for that

Issue #49889, opened 2018-10-03
Hosts File is incorrectly handled


jcrsilva:
### Description of Issue/Question
When modifying the hosts file using [states.host](https://github.com/saltstack/salt/blob/develop/salt/states/host.py) the entire file is rewritten to use 1 IP entry and a number of aliases.

Example:
```
127.0.0.1 alias1
127.0.0.1 alias2
127.0.0.1 alias3
```

Gets rewritten into:
```
127.0.0.1 alias1 alias2 alias3
```

Problems start appearing with large numbers of host aliases pointing to the same IP, [since the number of characters in a hosts file line is limited to 256](https://www.uow.edu.au/~blane/netapp/ontap/nag/networking/concept/c_oc_netw_maintaining_host_file_limits.html) (can anyone find an RFC for this?) which essentially becomes 254 because of \n\r line endings under windows.

Besides, there is [undocumented behaviour](https://superuser.com/questions/932112/is-there-a-maxium-number-of-hostname-aliases-per-line-in-a-windows-hosts-file) under Windows that limits the number of aliases in a hosts entry to 8 (1 host + 8 aliases). I was able to reproduce this under Windows Server 2016 with the following:

![image](https://user-images.githubusercontent.com/10566425/46439080-7c856800-c757-11e8-81da-c83f052609df.png)

"alias" host entries stop being loaded into dns cache with alias10 (`ipconfig /displaydns`).

Same behaviour not present in Ubuntu Xenial, where I am able to ping alias10:

![image](https://user-images.githubusercontent.com/10566425/46439831-77291d00-c759-11e8-81b3-6ae27374096f.png)

I am happy to submit a patch for [modules.hosts](https://github.com/saltstack/salt/blob/develop/salt/modules/hosts.py) where this is handled. However I would like to discuss the approach.

Regarding line character limit:
1. Stop using aliases, and have one `IP Hostname` entry per line;
2. Add aliases until the character limit is passed, at which point we break to the following line;

Regarding windows handling of aliases:
1. Identify if we are running under windows (via os grain) and handle hosts file differently;
2. Apply window's undocumented limits to all OSs and err on the safe side;

### Setup
1 sls file with:
```
hosts_updater.set_windows_routes.{{redacted}}: 
  host.present:
    - ip: 127.0.0.1
    - names:
      - "{{redacted}}.redacted.info"
```

### Steps to Reproduce Issue
Mentioned above

### Versions Report
```
Salt Version:
[...]

Issue #29592, opened 2015-12-10
host.present may break /etc/hosts file format


ckoe-bccms:
Dear all,

using host.present to add an host with alias names to /etc/hosts may break the file format as explained below. 

I am not sure whether this is merely a case of unexpected behavior or wrong usage by me or might indeed qualify as a bug.

The format of a line in /etc/hosts on an unixoid system is fixed as
`IP_address canonical_hostname [aliases...]`
so to be more concrete usually `10.10.10.111 test.example.com test`
please confirm with `man (5) hosts`. The FQDN has to be in front of the short-hand alias.

Naively using host.present to add the whole line `10.10.10.111 test.example.com test` in one call generates instead:

``` sls
test_input:
  host.present:
    - ip: 10.10.10.111
    - names:
      - fqdn.domain.com
      - alias
```

Generated line in the minions `/etc/hosts`: `10.10.10.111            alias fqdn.domain.com`

``` sls
test_input:
  host.present:
    - ip: 10.10.10.111
    - names:
      - alias
      - fqdn.domain.com
```

Generated line in the minions `/etc/hosts`: `10.10.10.111            alias fqdn.domain.com`

Both break `/etc/hosts` format, the alias is in front of the FQDN. 

Calling host.present twice can be used to generate the correct order of entries as shown below

``` sls
input_fqdn:
  host.present:
    - ip: 10.10.10.111
    - names:
      - fqdn.domain.com

input_alias:
  host.present:
    - ip: 10.10.10.111
    - names:
      - alias
```

Generated line in the minions `/etc/hosts`: `10.10.10.111            fqdn.domain.com alias`

but is awkward.

In any case I would suggest to mention this behavior in the salt.states.host documentation. If generating an FQDN and several aliases for one IP with one call of host.present is not the intended use case of the option to pass several names as a list (so in case I am using the features of host.present in a wrong way) it would also be nice to document the intended use. 

Best Regards

Christof

Salt Version information:

Master

``` yaml
 Salt Version:
           Salt: 2015.8.3

Dependency Versions:
         Jinja2: 2.7.3
       M2Crypto: Not Installed
           Mako: 1.0.0
         PyYAML: 3.11
          PyZMQ: 14.4.0
         Python: 2.7.9 (default, Mar  1 2015, 12:57:24)
           RAET: Not Installed
        Tornado: 4.2.1
            ZMQ: 4.0.5
           cffi: 0.8.6
       cherrypy: Not Installed
       dateutil: 2.2
[...]

PR #50647, opened 2018-11-27
fixing hosts module making it OS and standard compliant

tacf:
### What does this PR do?

This pull request fixes salt Hosts module (and state) to be Standard compliant and work across OSes, fixing a known Windows limitation. For details views issues referenced bellow.

https://github.com/saltstack/salt/issues/49889
https://github.com/saltstack/salt/issues/29592

### Previous Behavior
- More than 264 chars per line ([Standard Complience](https://www.uow.edu.au/~blane/netapp/ontap/nag/networking/concept/c_oc_netw_maintaining_host_file_limits.html))
- More than 8 alias per line ([Windows Limitation](https://superuser.com/questions/932112/is-there-a-maxium-number-of-hostname-aliases-per-line-in-a-windows-hosts-file))
- Alias and Fqdn mixed together (man (5) hosts)

### New Behavior
- Limited to 264 char per line
- Max 8 alias per line
- fqdn appear before aliases

### Tests written?

No

### Commits signed with GPG?

No


salt/modules/hosts.py
```diff
@@ -34,7 +32,7 @@ def _get_or_create_hostfile():
     if hfn is None:
         hfn = ''
     if not os.path.exists(hfn):
-        with salt.utils.files.fopen(hfn, 'w'):
+        with salt.utils.fopen(hfn, 'w'):
```
damon-atkins: Did you based this on recent code?  Strange to be changing `salt.utils.files.fopen` to `salt.utils.fopen`

damon-atkins (changes requested):
Thanks for raising a PR. The limits you have found are they the same on all platforms? Which platforms have you tested the limits on?

salt/modules/hosts.py
tacf: This had to be done due to the version we're using, so i'm assuming that the one i removed is the newer way of doing it right? I'll gladly change it back.

damon-atkins:
You need to take a copy from the develop branch and update it. OR raise the pull request against the version of salt you updated/changed.

salt/modules/hosts.py
jcrsilva: @damon-atkins these limits are only present in Windows as far as we can tell, but we did test on multiple versions of Windows Server (2016 and 2012).

Applying these limits to *nix OSs will make the hosts file more readable tho.

I couldn't find any info on this on an RFC per se, but [articles like these litter the internet](https://www.uow.edu.au/~blane/netapp/ontap/nag/networking/concept/c_oc_netw_maintaining_host_file_limits.html) (whithout citing sources).
I did test that at least in Ubuntu Xenial, the 255 char limit is not enforced.

cachedout:
@tacf This is failing a series of tests I'm afraid.

thatch45:
@tacf I like what you are trying to do here and we need to get this fix in. But I suspect that the failing tests are relative to changing some of the default paths here.
If we don't hear back in a few days with changes I will close this out, that would nto mean that we don't want this contribution, just that when a PR gets stale we should likely refrsh the process.
I would also ask that if you are going to do dedicated docs/whitespace cleanup that we put that in a separate PR, that makes it easier to get those trivial PRs in and get the main code that needs to be fixed in more easily. :)

dwoz:
@tacf Can you please resolve the conflicts and update the branch with the lates changes from develop? That should take care of some of the test noise.

dwoz (commented):
@tacf Are you able to rebase this to the `master` branch?

leeclemens:
What issue does this PR address? I see two listed with a question mark as to which one to use. My primary concerns are if it addresses the concerns I raised in the PR I was involved in and if sufficient tests will be introduced to ensure consistent behavior (the concerns I raised).

Are we sure we want to introduce this change in behavior:
`Alias and Fqdn mixed together (man (5) hosts)`
to
`fqdn appear before aliases`

Is there an RFC that says FQDN's should appear before aliases? I believe aliases can be FQDN's (please correct me if I'm wrong, but a default install of CentOS 7 has FQDNs as aliases).

Side note: One of the references is to uow.edu.au which states "Maximum number of name servers is 3" in /etc/hosts. I am not aware of nameservers being related to this /etc/hosts change or being declared modernly in /etc/hosts file. Not trying to be pedantic but it also claims "No file size limit", which cannot be true as every file system has a file size limit. It also makes a few other claims I cannot verify from the man page or RFC 952.

salt/modules/hosts.py
leeclemens: NetApp says [Maximum line size is 1022 characters](https://library.netapp.com/ecmdocs/ECMP1368834/html/GUID-C6F3B6D1-232D-44BB-A76C-3304C19607A3.html). Also [Maximum number of aliases is 34](https://library.netapp.com/ecmdocs/ECMP1368834/html/GUID-C6F3B6D1-232D-44BB-A76C-3304C19607A3.html)

leeclemens: NetApp was the 4th google result. I didn't see the link @jcrsilva referenced from uow.edu.au after 5 pages of google search for "hosts file line limit", so "litters the internet" seems a bit dubious.

leeclemens:
> This pull request fixes salt Hosts module (and state) to be Standard compliant and work across OSes, fixing a known Windows limitation

- What is not currently Standard compliant?
- If it needs to be "fixed" to work with a Standard compliant OS's "limitation", how is that Standard compliant?

salt/modules/hosts.py
leeclemens: And there is an RFC referenced by the /etc/hosts man page, 952.

Ch3LL:
@tacf bump here. Any chance we can write some tests and you can respond to @leeclemens 's concerns

sagetherage:
Reviewed in Grooming 2021-APR-05 considering closing, but reopening issues to later fix.

sagetherage:
@tacf can you please rebase against the master branch? we are no longer taking changes against develop.

Discussed in Grooming 2021-APR-19 I will give this 2 more weeks and if no response, will close.

leeclemens:
@sagetherage I am still genuinely interested to know what standard this change is complying to. It appears only to be a personal page on University of Wollongong, Australia's website with no other references.

sagetherage:
> I am still genuinely interested to know what standard this change is complying to. It appears only to be a personal page on University of Wollongong, Australia's website with no other references.

@leeclemens great point!

tacf:
Hi, unfortunately i cannot follow up on this. The issues was found during real world usage and after some major research (google) we found that among several links that reported the limitations we found. They should be easily reproducible but like a mentioned i can't follow up on this atm (since this most likely will require additional testing on newer windows server versions to make sure it still holds).

PR #48598, opened 2018-07-16
Thread safe loader 2

mattp-:
### What does this PR do?
this is a version of #45782 with latest develop merged in. I've not made any changes other than resolving merge conflicts; the original jenkins results have aged out so I'm creating this to re-run. If I figure out a solution I'll send PR to @smarsching and close this.

### Previous Behavior
Remove this section if not relevant

### New Behavior
Remove this section if not relevant

### Tests written?

Yes/No

### Commits signed with GPG?

Yes/No

Please review [Salt's Contributing Guide](https://docs.saltstack.com/en/latest/topics/development/contributing.html) for best practices.

See GitHub's [page on GPG signing](https://help.github.com/articles/signing-commits-using-gpg/) for more information about signing commits with GPG.

PR #50648, opened 2018-11-27
Thread safe loader 2018.3.3

smarsching:
This is a new version of #48598 that is now based against the 2018.3.3 branch. I made a few modifications that might fix (at least some of) the test failures we see in the original PR.

I only created this PR so that I can run Jenkins on my changes. When testing changes locally, I get sporadic errors that do not really seem to be related to my changes (the same code sometimes runs and sometimes fails). I want to debug things in a well-defined environment so that I can concentrate on what effects certain changes have...

I do not intend this PR to be merged and will close it in favor of #48598 once I have finished my tests.

### DIFF
--- a/salt/cloud/clouds/ec2.py
+++ b/salt/cloud/clouds/ec2.py
@@ -82,7 +82,6 @@ import hashlib
 import binascii
 import datetime
 import base64
-import msgpack
 import re
 import decimal
 
@@ -91,6 +90,7 @@ import salt.utils.cloud
 import salt.utils.files
 import salt.utils.hashutils
 import salt.utils.json
+import salt.utils.msgpack
 import salt.utils.stringutils
 import salt.utils.yaml
 from salt._compat import ElementTree as ET
@@ -4859,7 +4859,7 @@ def _parse_pricing(url, name):
         __opts__['cachedir'], 'ec2-pricing-{0}.p'.format(name)
     )
     with salt.utils.files.fopen(outfile, 'w') as fho:
-        msgpack.dump(regions, fho)
+        salt.utils.msgpack.dump(regions, fho)
 
     return True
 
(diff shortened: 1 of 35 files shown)


isbm (changes requested):
This has to be rebased/cherrypicked instead.

smarsching:
@isbm: Sure, I will create a proper PR with my changes neatly squashed into one commit as soon as I have figured out what's causing the problems.

I started this try out on the 2018.3.3 branch, but then decided to switch develop because it seemed like Jenkins was not running the tests at all and I though that it might work with the develop branch. Seems like even with that branch, the tests do not run at all in the Ubuntu 16.04 test environment. 

Incidentally, I also use a Ubuntu 16.04 VM for my test environment and the fact that I am having funny problems in that environment was the main reason why I created this PR in the first place (to have a playground where I can test various approaches without having to worry about a working test environment).

mattp-:
👍 👍

cachedout:
Really glad to see this, @smarsching. We want to be very focused on getting this in as soon as we can. I sincerely apologize for the ridiculous delay on this. It's our fault and we appreciate you sticking with us. :)

smarsching:
Just so that you know: I found the solution to why my changes caused so many problems and an elegant fix for these problems.

Now, I have to clean up my branch (it got really messy from trying so many different things), and the I will force-push this changes to the branch referenced by this PR.

I will keep the branch based against 2018.3.3 (as I need a patch for that version for internal use anyway), but from that point it should be pretty straight forward of porting the changes to the develop branch.

I cannot promise that I will get this done today (its already 7 pm around here), but if not you can expect to have the cleaned up version by tomorrow.

isbm:
@smarsching that is very cool. Looking forward for your final PR!

smarsching:
From my point of view, the code is complete, but Jenkins did not run the tests (most likely because now it is again based on 2018.3.3). Therefore, I created a new PR #50655 that is now based on develop. Let's see whether the tests run now or whether there are still some problems.

PR #50655, opened 2018-11-27
Fix race condition in Salt loader.

smarsching:
### What does this PR do?

This PR is a new version of #39948, #48598, and #50648 so the description has been copied from there. This PR will hopefully resolve the problems that existed in the original PRs. The code is identical to #50648, but this PR is based on the develop branch instead of the 2018.3.3 branch because Jenkins obviously cannot run tests for branches based on the 2018.3.3 branch.

There was a race condition in the salt loader when injecting global values (e.g. "__pillar__" or "__salt__") into modules. One effect of this race condition was that in a setup with multiple threads, some threads may see pillar data intended for other threads or the pillar data seen by a thread might even change spuriously.

There have been earlier attempts to fix this problem (#27937, #29397). These patches tried to fix the problem by storing the dictionary that keeps the relevant data in a thread-local variable and referencing this thread-local variable from the variables that are injected into the modules.

These patches did not fix the problem completely because they only work when a module is loaded through a single loader instance only. When there is more than one loader, there is more than one thread-local variable and the variable injected into a module is changed to point to another thread-local variable when the module is loaded again. Thus, the problem resurfaced while working on #39670.

This patch attempts to solve the problem from a slightly different angle, complementing the earlier patches: The value injected into the modules now is a proxy that internally uses a thread-local variable to decide to which object it points. This means that when loading a module
again through a different loader (possibly passing different pillar data), the data is actually only changed in the thread in which the loader is used. Other threads are not affected by such a change.

This means that it will work correctly in the current situation where loaders are possibly created by many different modules and these modules do not necessary know in which context they are executed. Thus it is much more flexible and reliable than the more explicit approach
used by the two earlier patches.

This PR fixes problems that surfaced when developing the parallel runners feature (#39670), but is also related to problems reported earlier (#23373). The problems described in #29028 might also be related, though this needs further confirmation.

### Previous Behavior
Changes to pillar data in one thread might have influenced pillar data in a different thread. Whether or when this problem actually appeared was unpredictable (due to the nature of a race condition).

### New Behavior
Changes to pillar data in one thread will never affect the pillar data in a different thread.

### Tests written?

Yes

### Regression Potential

The change to the loader code itself is rather small, but it affects a central component. So if there is a problem, the impact might affect nearly all components.


smarsching:
The jenkins/pr/py2-centos-7 run succeeds now. The other ones still fail, but I do not think that this is related to my changes.

@cachedout: Could you maybe have a look and tell me whether any of the failed tests might be related to my changes?

Just a few comments about what's different in comparison to the earlier versions of this PR:

Instead if trying to "unproxy" potentially proxied objects before passing them to serialization code, I know install a handler that is called when the serialization code (json or msgpack) encounters an object of an unsupported type. This is much cleaner than the `unproxy_recursive` function from the original PR (that I removed now). More importantly, it fixes the problem that `unproxy_recursive` was not working as expected in some cases.

There are three commits: The first commit introduces `salt.utils.msgpack` (which is similar to `salt.utils.json`) so that there is a single place where the serialization code can be intercepted.

The second commit includes the actual thread-safety fix. It introduces `salt.utils.thread_local_proxy` which is then used by the Salt loader to proxy injected objects and `salt.utils.json` and `salt.utils.msgpack` to unproxy such objects when a proxy is encountered during serialization.

The third commit basically reverts some changes from the first commit. There are a few places where using the `msgpack` module directly is actually correct, so I made sure that these pieces of code do not (completely) replace `msgpack` with `salt.utils.msgpack`.

cachedout:
@smarsching Something went quite wrong with the test suite there. I'll re-run these.

cachedout:
I'm a little concerned about this failure:

https://jenkinsci.saltstack.com/job/pr-kitchen-ubuntu1604-py3/job/PR-50655/3/testReport/junit/unit.serializers.test_serializers/TestSerializers/test_sls_aggregate/

smarsching:
@cachedout You were right to be concerned, but for the wrong reasons. 😜

The code in salt.serializers.yamlex would indeed fail if a proxied object was passed to it. I pushed a commit that fixes this. I do not think that there currently is any code that tries to serialize a proxied object to YAML, but it is still better to be prepared for that case.

However, this is not the reason why the test fails. I cannot they for sure why it fails, but I **can** say for sure that it is not because of a proxy object: The test fails even when run on the vanilla develop branch, so it is a preexisting bug.

What I found during debugging is that it only fails when PyYAML is installed with bindings to libyaml (C code). That is probably a reason why it only fails in some of the environments. I could not reproduce the problem in my own Ubuntu 16.04 test environment until I specifically installed PyYAML with the C extensions.

cachedout:
We think we have this yamlex issue addressed here: https://github.com/saltstack/salt/pull/50741

smarsching:
@cachedout Thanks for the heads-up. It lookes like that PR has been merged into the `fluorine` branch, not the `develop` branch, so merging the changes from the `develop` branch into my branch probably will not help.

I could certainly cherry-pick the changes in order to check that they fix the failing tests in this branch, but I am afraid this might cause merge conflicts later.

What do you think is the best option to get this changes into my branch for testing them?

cachedout:
@smarsching We'll get the merge-forward done ASAP. Let's plan on waiting for that and then I'll just come over to this PR and rebase it to include the changes. Hopefully the tests all pass after that. :)

thatch45:
The merge forward is done. I updated the branch and if tests pass we can merge

thatch45 (approved):

salt/loader.py
```diff
@@ -1082,6 +1084,76 @@ def _mod_type(module_path):
...
+        been created for the variable.
+
+    force_lock:
+        whether the lock should be acquired before checking whether a proxy
+        object for the specified name has already been injected into the
+        module. If ``False`` (the default), this function checks for the
+        module's variable without acquiring the lock and only acquires the lock
+        if a new proxy has to be created and injected.
+    '''
+    from salt.utils.thread_local_proxy import ThreadLocalProxy
```
s0undt3ch: Why the re-import?

s0undt3ch (approved):

thatch45:
@s0undt3ch do you think these test failures are related?

s0undt3ch:
They seem to, yes.
State module test failures on several of the tested platforms. It really seems related.

thatch45:
This PR fixes a very legit issue, but I don't know why it is causing the states to fail like this

DmitryKuzmenko:
The failing state tests are
```
integration.modules.test_state.StateModuleTest.test_exclude
integration.modules.test_state.StateModuleTest.test_issue_2068_template_st
```

Both are failing on `develop` too. At least on my machine.
Analyzing.

DmitryKuzmenko:
One just done by this PR: https://github.com/saltstack/salt/pull/51119
The second one is in progress.

DmitryKuzmenko:
Both issues are done by #51119, I've just updated that PR with the fix for the second problem. Sorry if confused anybody.

DmitryKuzmenko:
@thatch45 There were 2 tests failures common for all distributions. I've fixed them. Also windows tests were failing. Now I see lint errors and 2 win and 1 linux failures. I've fixed lint errors and updated branch. Let's see..

thatch45:
Fantastic! I think that something is blocking the windows tests right now but this should be good to merge!

mattp-:
👍 going to close my other failed attempt at getting this in

thatch45:
Given the extent of this PR we want to watch it for a bit so it won't be backported into the 2018.3.4 release, but assuming we don't see any issues we should backport it to 2018.3.4 and 2019.2.1

max-arnold:
@smarsching The following pattern (a Jinja check if a module is avalable) which works in a stable release, fails on the latest develop branch deep inside the `thread_local_proxy.py`:

```
{% if 'module.function' in salt %}
{% do salt.log.warning("Module is available") %}
{% endif %}
```

```
  File "/Users/user/.virtualenvs/salt-develop/bin/salt-call", line 7, in <module>
    exec(compile(f.read(), __file__, 'exec'))
  File "/Users/user/saltstack-repo/scripts/salt-call", line 11, in <module>
    salt_call()
  File "/Users/user/saltstack-repo/salt/scripts.py", line 418, in salt_call
    client.run()
  File "/Users/user/saltstack-repo/salt/cli/call.py", line 57, in run
    caller.run()
  File "/Users/user/saltstack-repo/salt/cli/caller.py", line 119, in run
    ret = self.call()
  File "/Users/user/saltstack-repo/salt/cli/caller.py", line 212, in call
  File "/Users/user/saltstack-repo/salt/executors/direct_call.py", line 12, in execute
    return func(*args, **kwargs)
  File "/Users/user/saltstack-repo/salt/modules/state.py", line 778, in apply_
    return sls(mods, **kwargs)
  File "/Users/user/saltstack-repo/salt/modules/state.py", line 1343, in sls
    if __salt__['config.option']('state_data', '') == 'terse' or kwargs.get('terse'):
  File "/Users/user/saltstack-repo/salt/modules/config.py", line 150, in option
    if value in __pillar__.get('master', {}):
  File "/Users/user/.virtualenvs/salt-develop/bin/../lib/python3.6/_collections_abc.py", line 660, in get
    return self[key]
  File "/Users/user/saltstack-repo/salt/utils/context.py", line 236, in __getitem__
    return self._dict()[key]

...

  File "/Users/user/saltstack-repo/salt/utils/thread_local_proxy.py", line 261, in __getitem__
    return reference[key]
  File "/Users/user/saltstack-repo/salt/utils/context.py", line 236, in __getitem__
    return self._dict()[key]
  File "/Users/user/saltstack-repo/salt/utils/thread_local_proxy.py", line 261, in __getitem__
    return reference[key]
  File "/Users/user/saltstack-repo/salt/utils/context.py", line 236, in __getitem__
    return self._dict()[key]
  File "/Users/user/saltstack-repo/salt/utils/thread_local_proxy.py", line 261, in __getitem__
    return reference[key]
  File "/Users/user/saltstack-repo/salt/utils/context.py", line 236, in __getitem__
    return self._dict()[key]
  File "/Users/user/saltstack-repo/salt/utils/context.py", line 223, in _dict
    r = r[k]
  File "/Users/user/saltstack-repo/salt/utils/context.py", line 115, in __getitem__
    if self.active:
  File "/Users/user/saltstack-repo/salt/utils/context.py", line 90, in active
    return self._state.data is not None
RecursionError: maximum recursion depth exceeded in comparison
```

thatch45:
@mattp- what do you make of this?

mattp-:
@thatch45 removing the NamespacedDict wrapper seems to address the issue. I don't really see what purpose it was originally serving, contextDict and ThreadLocalProxy seem to aim to achieve similar things wrt locking access to thread local variables. Will check in on jenkins in the morning. @cachedout saw you in the git blame for namespacedDictWrapper, do you remember any context for what the original intent was?

mattp-:
alright from reading further it seems the idea was to carry the ref of the root stack context instead of the inner dicts for use with tornado.stackcontext. Will try to sort out why thats not playing nice with ThreadLocalProxy

cachedout:
That goes back all the way to @jacksontj IIRC.

PR #50661, opened 2018-11-27
INFRA-7855 - derp, didn't realize the create call used a completely d…

tkwilliams:
Well, just duh!  My previous PR on this only dealt with the "update" case, and I totally failed to clue in that the "create" case was using a different code-path entirely (calling the module function directly rather than preprocessing first before sending to the module).

Just duped my code from the state module over into the `create_function()` function in the module...


cachedout (approved):

terminalmage:

PR #50675, opened 2018-11-29
Reactor: remove thread pool for runner/wheel

mattp-:
### What does this PR do?
rather than execute inline in a separate pool, we can push these off to the master to handle like other work, so lets do that instead. This deprecates/removes the REACTOR_WORKER_HWM and REACTOR_WORKER_THREADS settings.

also included:
add root_key identity check for interacting with the reactor
since actions through the reactor/manage interface can be potentially  destructive, we use the root_key as a identifier to assert it originated  only from the runner as the root user as opposed to any event coming through the bus.

- performance bottleneck in the reactor engine.
- unauthenticated interaction with the reactor management interface

### Tests written?
Yes

### Commits signed with GPG?
No

### DIFF
--- a/salt/utils/reactor.py
+++ b/salt/utils/reactor.py
@@ -267,6 +267,13 @@ class Reactor(salt.utils.process.SignalHandlingMultiprocessingProcess, salt.stat
             if data['data'].get('user') == self.wrap.event_user:
                 continue
 
+            # NOTE: these events must contain the masters key in order to be accepted
+            # see salt.runners.reactor for the requesting interface
+            if 'salt/reactors/manage' in data['tag']:
+                master_key = salt.utils.master.get_master_key('root', self.opts)
+                if data['data'].get('key') != master_key:
+                    log.error('received salt/reactors/manage event without matching master_key. discarding')
+                    continue
             if data['tag'].endswith('salt/reactors/manage/add'):
                 _data = data['data']
                 res = self.add_reactor(_data['event'], _data['reactors'])
--- a/salt/utils/reactor.py
+++ b/salt/utils/reactor.py
@@ -38,9 +38,9 @@ log = logging.getLogger(__name__)
 REACTOR_INTERNAL_KEYWORDS = frozenset([
     '__id__',
     '__sls__',
-    'name',
     'order',
     'fun',
+    'key',
     'state',
 ])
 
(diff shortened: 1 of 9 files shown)


salt/utils/reactor.py
```diff
@@ -442,15 +450,7 @@ def run(self, low):
...
-            ret = l_fun(*args, **kwargs)
-
-            if ret is False:
-                log.error('Reactor \'%s\' failed  to execute %s \'%s\': '
-                            'TaskPool queue is full!'
-                            ' Consider tuning reactor_worker_threads and/or'
-                            ' reactor_worker_hwm',
-                            low['__id__'], low['state'], low['fun']
-                )
+            l_fun(*args, **kwargs)
```
DmitryKuzmenko: Why we're getting to ignore the result here?

DmitryKuzmenko (commented):
I agree at all, just please see one question inline.

salt/utils/reactor.py
mattp-: I guess the question is what should we do with the ret in reactors case? it was previously used to check it the Queue.Queue was full, but otherwise tossed. can we meaningfully do anything that isn't already emitted in the master log?

mattp-:
@DmitryKuzmenko btw had some out of band chat with mike,
```Matthew Phillips [10:50 AM]
@cachedout re; our talk yesterday about the threadpool, i was mistaken it does actually use runnerclient.asynchronous (the fork/daemonize) when it processes runner jobs
i still dont like this though. what id like to do is move the queue idea out of the reactor and into the master itself
that way the zeromq channels can drain to this queue (could even be something like a priority queue, do this job before any other jobs if they come in etc)
it would be very easy to provide metrics on whats going on with jobs coming in if we do that, ie what is the queue depth, what is the contents etc
making the queue size 0 / infinite would leave the behavior backwards compatible, while also stopping runner/wheel jobs from forkbombing the server
the stats in particular have been something we’ve been looking for for quite a while, it just now clued to me how we can implement it. would like to hear your thoughts :+1:
Mike Place [11:07 AM]
I’m a fan of the queuing approach. Moving to that model and providing stats would be a huge win all around.
Matthew Phillips [1:16 PM]
indeed. ok ill try to hack on that at some point. in the interim please slap a hold on that PR
Mike Place [1:23 PM]
10-4```

DmitryKuzmenko:
@mattp- the forks used everywhere is a big pain though.
What Mike said is a kind of what I'm trying to do in Tornado 5.0 work. So I agree.

DmitryKuzmenko (approved):

cachedout:
@mattp- Hrm. We did see some issues with this: https://jenkinsci.saltstack.com/job/pr-kitchen-ubuntu1604-py3/job/PR-50675/13/testReport/junit/integration.reactor.test_reactor/ReactorTest/test_reactor_wheel_reaction/

dwoz:
@mattp- Can you look into the conflicts?

mattp-:
@dwoz please consider this on hold for time being. I need to sort out a similar throttle on master itself before this is safe for merge

dwoz:
@mattp- Can we close this for now and once you get things resolve make another PR?

mattp-:
yep thats fine, closing

Issue #39608, opened 2017-02-23
x509 certificate_managed can't handle managed_private_key


4s1:
### Description of Issue/Question
The x509.certificate_managed state doesn't work when it's configured to handle it's own private key as [explained in the docs](https://docs.saltstack.com/en/latest/ref/states/all/salt.states.x509.html#salt.states.x509.certificate_managed) or [shown in the example](https://docs.saltstack.com/en/latest/ref/states/all/salt.states.x509.html#module-salt.states.x509) and it's not existing yet


### Setup

```SLS
# self signed cert
/etc/nginx/certs/localhost.cert.pem:
  x509.certificate_managed:
    - signing_private_key: /etc/nginx/private/localhost.key.pem
    - CN: localhost
    - days_valid: 365
    - days_remaining: 90
    - managed_private_key:
       name: /etc/nginx/private/localhost.key.pem
       bits: 2048
    - require:
       - pkg: python-m2crypto
```

### Steps to Reproduce Issue

1. configure state as shown above
2. make sure key file doesn't exist, but path is writable
3. and apply state.

```
         ID: /etc/nginx/certs/localhost.cert.pem
    Function: x509.certificate_managed
      Result: False
     Comment: An exception occurred in this state: Traceback (most recent call last):
                File "/usr/lib/python2.7/site-packages/salt/state.py", line 1745, in call
                  **cdata['kwargs'])
                File "/usr/lib/python2.7/site-packages/salt/loader.py", line 1702, in wrapper
                  return f(*args, **kwargs)
                File "/usr/lib/python2.7/site-packages/salt/states/x509.py", line 473, in certificate_managed
                  new = __salt__['x509.create_certificate'](testrun=True, **kwargs)
                File "/usr/lib/python2.7/site-packages/salt/modules/x509.py", line 1484, in create_certificate
                  passphrase=kwargs['signing_private_key_passphrase'])
                File "/usr/lib/python2.7/site-packages/salt/modules/x509.py", line 679, in get_public_key
                  text = get_pem_entry(text)
                File "/usr/lib/python2.7/site-packages/salt/modules/x509.py", line 472, in get_pem_entry
                  raise salt.exceptions.SaltInvocationError(errmsg)
              SaltInvocationError: PEM text not valid:
              /etc/nginx/private/localhost.key.pem
     Started: 16:29:53.585609
    Duration: 99.44 ms
     Changes:   
```

### Versions Report
```
Salt Version:                                                                                                                                                                                      
[...]

PR #49508, opened 2018-09-05
Fix X509 remote signing on python3

bergmannf:
### What does this PR do?

It fixes the `x509` module to work, when using the `sign_remote_certificate` functionality.

### What issues does this PR fix or reference?

Another part for fixing https://bugzilla.suse.com/show_bug.cgi?id=1106382.

### Previous Behavior

Remote signing would not work and fail with an error `salt.exceptions.SaltInvocationError: PEM does not contain a single entry of type CERTIFICATE:`

### New Behavior

Remote signing works again.

### Tests written?

No, however a repository with a reproducer can be found [here](https://github.com/bergmannf/salt_x509_reproducer).

Please review [Salt's Contributing Guide](https://docs.saltstack.com/en/latest/topics/development/contributing.html) for best practices.

PR #50676, opened 2018-11-29
Fix python3 x509

Poil:
### What does this PR do?
Fix x509 usage with Python 3, not sure if it's the right way but working for me

```     Comment: An exception occurred in this state: Traceback (most recent call last):
                File "/usr/lib/python3.4/site-packages/salt/state.py", line 1913, in call
                  **cdata['kwargs'])
                File "/usr/lib/python3.4/site-packages/salt/loader.py", line 1898, in wrapper
                  return f(*args, **kwargs)
                File "/usr/lib/python3.4/site-packages/salt/states/x509.py", line 554, in certificate_managed
                  file_args['contents'] += certificate
              TypeError: Can't convert 'bytes' object to str implicitly
```

My comment in #39608

### Tests written?

No

### Commits signed with GPG?

No

### DIFF
--- a/salt/states/x509.py
+++ b/salt/states/x509.py
@@ -164,6 +164,7 @@ import copy
 # Import Salt Libs
 import salt.exceptions
 import salt.utils.stringutils
+from salt.utils.locales import sdecode
 
 # Import 3rd-party libs
 from salt.ext import six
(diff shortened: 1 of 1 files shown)


Poil:
sorry seems to be already fixed by #49508

PR #50692, opened 2018-11-29
Update a couple test modules to use with_tempdir/with_tempfile decorators

terminalmage:
These prevent us from needing to do manual tempfile cleanup.


cachedout (approved):

tests/integration/states/test_x509.py
```diff
@@ -19,48 +22,54 @@
...
+    def test_issue_50637(self, pemfile):
+        ret = self.run_state(
+            'x509.pem_managed',
+            name=pemfile,
+            text=self.x509_cert_text,
+            show_changes=False,
+            encoding='utf-8',
+            encoding_errors='ignore',
+            mode=640)
+        print ret
```
s0undt3ch: Left over right?

s0undt3ch (changes requested):

tests/integration/states/test_x509.py
s0undt3ch: Also, missing asserts?

terminalmage: It turns out that incomplete test was something I was working on for an issue that was a duplicate of something already being tested. So I have removed it.

s0undt3ch (approved):

cachedout (approved):

PR #50694, opened 2018-11-29
INFRA-7855-3 - fix weird corner case in boto3_sns module :-/

tkwilliams:
OK, so occasionally a subscription will be left in PendingAccept (generally b/c it's to another account and thus has to wait for a human being to click and accept it).  AWS then cleverly RETURNS the string `PendingAccept` as the ARN field when you query the #$#@$#@ topic, which is just breakage because that's NOT a VALID ARN.  Thus any attempt to use it (for instance when looping through all subscriptions attached to a topic to garner their details) gets a None back for the subscription details where a dictionary is expected, leading to eventual stacktraces...

This just tells `describe_topic()` to skip any funky looking ARNs it encounters while looping through them.


cachedout (approved):

terminalmage:

PR #50696, opened 2018-11-29
add asynchronous to states.saltmod

mattp-:
### What does this PR do?
this exposes the cmd_async functionality of each client that saltmod
wraps in the saltmod state. The idea of this being fire and forget,
wherein you can either store that runs jid for later processing, or do
nothing if you don't care. Without this, you must pay the round trip
penalty of waiting for the full job to complete.

### Tests written?
Yes

### Commits signed with GPG?
No


cachedout (approved):
This appears to be very smartly done. I like it!

PR #50706, opened 2018-11-30
blkid: add search by token

aplanas:
### What does this PR do?

Add the `token` parameter in `blkid`, to allow extended search.

### DIFF
--- a/salt/modules/disk.py
+++ b/salt/modules/disk.py
@@ -268,10 +268,16 @@ def percent(args=None):
 
 
 @salt.utils.decorators.path.which('blkid')
-def blkid(device=None):
+def blkid(device=None, token=None):
     '''
     Return block device attributes: UUID, LABEL, etc. This function only works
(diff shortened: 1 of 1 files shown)


salt/modules/disk.py
```diff
@@ -268,24 +268,33 @@ def percent(args=None):
...
+    token
+        Any valid token used for the search
+
     CLI Example:
 
     .. code-block:: bash
 
         salt '*' disk.blkid
         salt '*' disk.blkid /dev/sda
+        salt '*' disk.blkid UUID=6a38ee5-7235-44e7-8b22-816a403bad5d
```
cachedout: Is this the example intended to illustrate the use of a token? If so, I think we should make this more clear.

aplanas: Ouch. Yes.

cachedout (approved):

Issue #50578, opened 2018-11-19
Feature request: add salt.modules.file.tail


cruscio:
### Description of Issue/Question

It'd be useful to have a `tail` method in the `file` module, for retrieving the end of log files from minions when troubleshooting issues - particularly on Windows where `cmd.run 'tail -n <#> <file_name>'` isn't an option. Regardless of platform, it'd be nice to have a python-native tail implementation.  

E.g. `salt.modules.file.tail(path, lines=10)` - [example implementations](https://stackoverflow.com/questions/136168/get-last-n-lines-of-a-file-with-python-similar-to-tail) 

-----

Alternatively, implement/document use of negative `seek` & `offset` parameters in `seek_read` - which doesn't seem possible without a significant change to `seek_read`, given that [internally it uses `os.lseek`](https://github.com/saltstack/salt/blob/0b32461f0f3923193f06c2a68b45ab661246bade/salt/modules/file.py#L3238)

steveno:
I'll be submitting a PR for this later today. Feel free to assign to me.

cruscio:
I was just testing this change, but if you have something you want to submit, go for it. 

```


def tail(path,lines='10'):
    '''
    .. versionadded:: Neon

    Just like the ``tail`` command, return the last lines of a file.

    path
        path to file

    lines
        number of lines to return. Default is 10

    CLI Example:

    .. code-block:: bash

        salt '*' file.tail /path/to/file 100
    '''
    path = os.path.realpath(os.path.expanduser(path))

    if not os.path.isfile(path):
        raise SaltInvocationError('File not found: {0}'.format(path))

    if not __utils__['files.is_text'](path):
        raise SaltInvocationError(
            'Cannot tail a binary file: {0}'.format(path))

    try:
        lines = int(lines)
    except ValueError:
        log.error('file.tail: \'lines\' value must be an integer')
        lines = 10

    try:
        with salt.utils.files.fopen(path,mode='r') as ofile:
            # Source: https://stackoverflow.com/a/13790289
            _buffer=4098
            lines_found = []
            block_counter = -1
            while len(lines_found) <= int(lines):
                try:
                    ofile.seek(block_counter * _buffer, os.SEEK_END)
                except IOError:
                    ofile.seek(0)
                    lines_found = ofile.readlines()
                    break
                lines_found = ofile.readlines()
                block_counter -= 1
            return lines_found[-lines:]
    except (OSError, IOError) as exc:
        raise CommandExecutionError(
[...]

PR #50708, opened 2018-11-30
Issue 50578

steveno:
### What does this PR do?
Add `file.tail`

https://github.com/saltstack/salt/issues/50578

### Previous Behavior
N/A

### New Behavior
Adds a `tail` behaviour that reads from the end of a file instead of the beginning.

### Tests written?
Yes

### Commits signed with GPG?
Yes


steveno:
After updating the branch, it looks like most of the CI jobs are passing now. :+1:

salt/modules/file.py
```diff
@@ -3213,6 +3213,61 @@ def touch(name, atime=None, mtime=None):
...
+            'Cannot tail a binary file: {0}'.format(path))
+
+    try:
+        lines = int(lines)
+    except ValueError:
+        log.error('file.tail: \'lines\' value must be an integer')
+        lines = 10
+
+    try:
+        with open(path) as tail_fh:
```
cachedout: Please change this to `salt.utils.fopen`

cachedout (changes requested):
There is no need to update the man page in this PR. It is automatically generated.

cachedout:
@steveno There are also a few lint errors which need to be addressed before we can move forward with this: https://jenkinsci.saltstack.com/job/pr-lint/job/PR-50708/3/warnings52Result/new/

steveno:
@cachedout Thanks for the quick review! I believe I have corrected the issues you mentioned. Please take another look at your convenience.

steveno:
@cachedout gentle ping

twangboy (approved):

steveno:
@cachedout If there is anything else you need for this MR, please let me know!

salt/modules/file.py
```diff
@@ -3213,6 +3213,62 @@ def touch(name, atime=None, mtime=None):
...
+
+    try:
+        with salt.utils.fopen(path) as tail_fh:
+            tail_fh.seek(0, os.SEEK_END)
+            position = tail_fh.tell()
+            line = ''
+            while position >= 0:
+                tail_fh.seek(position)
+                next_char = tail_fh.read(1)
+                if next_char == os.linesep:
```
skizunov: This won't work on Windows because `os.linesep` is '\r\n' on that platform. I would suggest you use '\n' instead of `os.linesep`.

salt/modules/file.py
```diff
@@ -3213,6 +3213,62 @@ def touch(name, atime=None, mtime=None):
...
+
+    if not __utils__['files.is_text'](path):
+        raise SaltInvocationError(
+            'Cannot tail a binary file: {0}'.format(path))
+
+    try:
+        lines = int(lines)
+    except ValueError:
+        log.error('file.tail: \'lines\' value must be an integer')
+        lines = 10
```
skizunov: I don't understand why you set `lines = 10` here. I would suggest raising an exception instead (in which case you can remove `log.error`).

skizunov (changes requested):

salt/modules/file.py
steveno: It's a safe value to default to in case they formed their command wrong.

I'm not attached to doing this way, I can change it if it's a problem.

salt/modules/file.py
steveno: I don't understand. If it's a Windows system, won't the file have `\r\n` in it? Why would you expect a Windows file to only have `\n` in it?

salt/modules/file.py
skizunov: It is not a good value on Windows because `next_char = tail_fh.read(1)` will always return 1 character and you are comparing a 2-character string, which will never test positive.

salt/modules/file.py
twangboy: `os.linesep` is 2 characters in Windows `\r` and `\n`. You're comparing a single character with 2 characters on Windows in this case.

salt/modules/file.py
skizunov: Sorry, my previous comment belonged to the other issue I opened. As for this case, it doesn't make sense to use a default in the case of invalid input. I think an exception is the best answer for invalid input. Defaults should be used for cases where there is no input, which isn't the case here.

salt/modules/file.py
```diff
@@ -3213,6 +3213,50 @@ def touch(name, atime=None, mtime=None):
...
+    try:
+        lines = int(lines)
+    except ValueError:
+        raise SaltInvocationError('file.tail: \'lines\' value must be an integer')
+
+    try:
+        with salt.utils.fopen(path) as tail_fh:
+            block_counter = -1
+            while len(lines_found) < lines:
+                tail_fh.seek(block_counter * 4098, os.SEEK_END)
```
skizunov: How do you handle the case where one of the lines you want is split across two different blocks of 4098?

salt/modules/file.py
```diff
@@ -3213,6 +3213,50 @@ def touch(name, atime=None, mtime=None):
...
+                lines_found = tail_fh.readlines()
```
skizunov: If you get into a second block, this will overwrite the lines from the first block, since you are re-setting `lines_found`.

skizunov (changes requested):

steveno:
I'm still working on this issue, but I'm having trouble coming with a test that highlights the potential issues pointed out by @skizunov. Specifically the "If you get into a second block, this will overwrite the lines from the first block, since you are re-setting lines_found."

Also, what command should I use to run the test suite for this? I've been running `python tests/runtests.py -m` but that doesn't seem to be working upon closer inspection. 

Any help in this area would be appreciated.

dwoz:
@steveno I've updated the title to make this issue a work in progress. When you are ready for a review you can remove the `[WIP]` prefix.

steveno:
I'll admit I had a lot more trouble with this than I expected to. I ended up adapting an algorithm I found [here](https://github.com/ActiveState/code/blob/73b09edc1b9850c557a79296655f140ce5e853db/recipes/Python/120686_Read_a_text_file_backwards/recipe-120686.py).

steveno:
It looks like the jenkins/pr/lint build is frozen. If it's required, can someone restart it?

PR #50725, opened 2018-12-03
 mount: add fstab_{present,absent} states 

aplanas:
### What does this PR do?

Add two new states in mount:

This PR depends on https://github.com/saltstack/salt/pull/50706 as a helper. But because are not related, I separated it in two. Once that is reviewed and merged I will rebase properly this PR.

### Tests written?

Yes

### DIFF
--- a/salt/states/mount.py
+++ b/salt/states/mount.py
@@ -956,3 +956,295 @@ def mod_watch(name, user=None, **kwargs):
     else:
         ret['comment'] = 'Watch not supported in {0} at this time'.format(kwargs['sfun'])
     return ret
+
+
+def _convert_to(maybe_device, convert_to):
+    '''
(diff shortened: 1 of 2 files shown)


salt/states/mount.py
```diff
@@ -956,3 +956,295 @@ def mod_watch(name, user=None, **kwargs):
     else:
         ret['comment'] = 'Watch not supported in {0} at this time'.format(kwargs['sfun'])
     return ret
+
+
+def _convert_to(maybe_device, convert_to):
+    '''
+    Convert a device name, UUID or LABEL to a devide name, UUID or
```
cachedout: `device name`

cachedout (changes requested):

cachedout (approved):

Issue #50651, opened 2018-11-27
Some states need concurrent when running salt-minion as a non-root user - includes possible solution


bbh-kmd:
### Description of Issue/Question

When salt-minion is configured to run as a non-root user with sudo privileges, some states require concurrent to be set to True, as the minion starts another process to actually execute the state. When using a sudo setup, some states are executed by issuing a sudo command to salt-call, which then detects that another state is already running. This result is then propagated up the chain, and the result of the state is that another state is already running. Unfortunate, as the actual work should have been performed by the child process.

The workaround/solution to this issue is to force concurrent to True for the states that behaves in this manner.

So far, the states state.apply, state.sls, state.highstate has been identified and a workaround has been implemented in executors/sudo.py (https://github.com/saltstack/salt/pull/42021).

I have run into a few other states that need the fix as well, and are unsure whether this is now a conclusive list of all states needing concurrent to run in a sudo setup:

state.show_highstate
state.show_sls

### Setup and steps to Reproduce Issue

Setup and reproduction of errors has been fully described in https://github.com/saltstack/salt/issues/25842

### Versions Report
```
Salt Version:
           Salt: 2018.3.0

Dependency Versions:
           cffi: 1.5.2
       cherrypy: Not Installed
       dateutil: Not Installed
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.8
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.4.6
   mysql-python: Not Installed
      pycparser: 2.10
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: Not Installed
         Python: 2.7.13 (default, Jan 11 2017, 10:56:06) [GCC]
   python-gnupg: Not Installed
         PyYAML: 3.12
          PyZMQ: 14.0.0
           RAET: Not Installed
          smmap: Not Installed
        timelib: Not Installed
        Tornado: 4.2.1
            ZMQ: 4.0.4

System Versions:
           dist: SuSE 12 x86_64
         locale: UTF-8
        machine: x86_64
        release: 4.4.143-94.47-default
         system: Linux
        version: SUSE Linux Enterprise Server  12 x86_64
```

I've also seen the issue on a 2018.3.2 on Debian 9.5, so not unique to the above version.

[...]

PR #50729, opened 2018-12-04
Ignore ancestor processes when checking for concurrent states

bbh-kmd:
### What does this PR do?
Build list of ancestor processes, and check possible concurrent processes against this ancestor list. If a process is found as an ancestor, then we are probably running salt in a sudo setup, which sometimes require that a state be restarted with sudo, which makes it seem that there are multiple processes running the same state.

Issue #50651 

### Previous Behavior
Only check if concurrent is True

### New Behavior
Create list of ancestral processes, then ignore processes that are in this list.

### Tests written?
No

### Commits signed with GPG?
No


DmitryKuzmenko (approved):
I'm okay with this PR just not sure what's happening with the tests. I updated it to develop branch, let's see how tests will pass.

terminalmage:
@DmitryKuzmenko I found the problem. `ppid` is a function, it needs to be invoked to get the PID. Otherwise, you're just passing a function reference.

I pushed a commit to this PR to fix it.

DmitryKuzmenko:
@terminalmage you're awesome you know it? =)

terminalmage:
Actually it looks like I need to add one more commit. Seems that pre-2.0 versions of psutil have most of the functions as simple attributes. @bbh-kmd likely developed this on psutil < 2.0.

bbh-kmd:
The one installed on the system used is the one that comes with an SLES12sp3 (python-psutil-1.2.1-15.3.3), so I think you're right in your assumption about versions of the psutil, and it's just a bit behind on enterprise versions of various Linux distributions. I thought it was psutil_compat that overrode default behaviour, but as I don't know much about python, I may be wrong in that regard.

bbh-kmd:
Hmm... I don't know much about windows, but if a process list on Windows  doesn't end with 0 or 1 as the final originating ancestral process, it might be that the loop tries to continue past the last process in the chain. I don't know what that would result in. Does windows even hava a psutil module?

Is it possible to test for empty as well as 0 and 1 in Python? 

Or, according to stacktrace, it hits a "NoSuchProcess", which to my untrained mind seem to indicate that the chain has stopped, or the process has been removed before we look it up. That would mean no more ancestor processes, and we have thus reached the end of the process line on Windows. I don't know exactly where the exception occur.

If it's in the line just before the try-except, then we might try to catch the "NoSuchProcess" and set init_found = True at that point and then continue (skips rest of loop, right?).

I'm not sure how that would look, but probably something like the following:

...
while not init_found:
   try:
      cur_ppid = psutil.Process(cur_pid).ppid
   except psutil.NoSuchProcess:
      init_found = True
      continue
   try:
      ...

Sanity check, please.

dwoz:
@bbh-kmd We are no longer accepting PRs to the `develop` branch. This needs to be re-opened against the `master` branch. You can link to this PR from the new one to preserve any disscussion history. Sorry for the confusion.

Issue #50731, opened 2018-12-04
zookeeper.present state with no ACL specified shows changes in Test=True mode


rokasz:
### Description of Issue/Question
If no ACL is specified in the zookeeper.present state (because we don't need to enforce any ACLs), ACL settings should be ignored and only values checked. However, if Test=True is specified when running the state, it shows the changes, that are going to be made, even though the difference is just in the ACLs, not the values. 
Empty ACLs were allowed by this PR https://github.com/saltstack/salt/pull/46043, but this does not work for the Test mode. Also, the change was not fully correct, because the message still says 

### Setup
**connection_strings.sls**
```
conn_str_mongodb:
  zookeeper.present:
    - name: /site/connectionstrings/mongodb
    - value: "mongodb://admin:password@mongo1/database"
    - makepath: True
```

### Steps to Reproduce Issue
Running the state with Test=True outputs the following:
```
          ID: conn_str_mongodb
    Function: zookeeper.present
        Name: /site/connectionstrings/mongodb
      Result: None
     Comment: Znode /site/connectionstrings/mongodb is will be updated
     Started: 07:49:39.498238
    Duration: 16.733 ms
     Changes:
              ----------
              new:
                  ----------
                  acls:
              old:
                  ----------
                  acls:
                      |_
                        - 31
                        |_
                          - world
                          - anyone
```
Running the state in normal mode (not Testing) with the same znode already available in the zookeeper cluster, output is the following:
```
          ID: conn_str_mongodb
    Function: zookeeper.present
        Name: /site/connectionstrings/mongodb
      Result: True
     Comment: Znode /site/connectionstrings/mongodb successfully updated
     Started: 07:55:13.788249
    Duration: 18.059 ms
     Changes:
```
As you can see, the values were identical and no changes are shown, but the message is confusing (it says znode was updated), but it was not updated. This should be fixed as well. Having empty ACL settings should just ignore ACL and compare just the values.

### Versions Report
Salt Version:
           Salt: 2018.3.3

Dependency Versions:
           cffi: Not Installed
       cherrypy: Not Installed
       dateutil: 1.5
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.7.2
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: 0.21.1
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.4.6
[...]

PR #50732, opened 2018-12-04
Ignoring ACL check for empty ACL in zookeeper.present

rokasz:
### What does this PR do?
Ignores the ACL differences, if no ACL is specified in the zookeeper.present state, Just values are compared.

https://github.com/saltstack/salt/issues/50731

### Previous Behavior
If zookeeper.present ACL setting is None and znode values are the same, running a state with Test=True always shows ACL changes will be made, while actually running the state would result in the incorrect message "successfully updated", even though the values are the same and no ACL changes would be done as well.

### New Behavior
If zookeeper.present ACL setting is None and znode values are the same, running a state would not show and/or make any changes and respond with the message "is already set to the correct value with the correct acls"


cachedout (approved):
LGTM.

PR #50735, opened 2018-12-04
Add jsonnet salt module

bbinet:
This module provides parsing support of jsonnet format (which is a
data templating language based on json).

This module depends on the jsonnet Python module which can be installed
with "pip install jsonnet".

### DIFF
--- a/salt/modules/jsonnet.py
+++ b/salt/modules/jsonnet.py
@@ -1,0 +1,90 @@
+# -*- coding: utf-8 -*-
+'''
+Jsonnet - A data templating language
+
+Module to provide parsing support of jsonnet format.
+
+:depends:    - jsonnet Python module (>= 0.11.2)
+
+:configuration: This module accepts library paths configuration that
+    can be passed as function parameters or as configuration settings in
+    /etc/salt/minion on the relevant minions::
+
+        jsonnet.library_paths:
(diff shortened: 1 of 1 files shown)


salt/modules/jsonnet.py
```diff
@@ -0,0 +1,90 @@
...
+        raise RuntimeError('Got invalid filename (empty string).')
+    if rel[-1] == '/':
+        raise RuntimeError('Attempted to import a directory')
+    full_paths = [rel] if rel[0] == '/' else \
+            [os.path.realpath(os.path.join(dir, rel))] + \
+            [os.path.realpath(os.path.join(l, rel)) for l in lpaths or []]
+
+    for full_path in full_paths:
+        if os.path.isfile(full_path):
+            with open(full_path) as f:
```
cachedout: Please use `salt.utils.files.fopen`

cachedout (changes requested):

salt/modules/jsonnet.py
bbinet: @cachedout Done in 64b6054.

cachedout (approved):

max-arnold:
Maybe it would be better to convert this into a serializer module?

PR #50741, opened 2018-12-04
Ensure that pure python dumper is used in PY3 for yamlex serializer

terminalmage:
Depending on how PyYAML was built, `yaml.SafeDumper` may actually be `yaml.cyaml.CSafeDumper` (i.e. the C dumper instead of pure Python). This ensures that the pure Python dumper is imported in all cases.

Fixes https://github.com/saltstack/salt-jenkins/issues/1223

**NOTE: This was only failing in Arch, so the py3 Arch branch build for Fluorine will need to run once this is merged to truly confirm this fix.**

### DIFF
--- a/salt/serializers/yamlex.py
+++ b/salt/serializers/yamlex.py
@@ -128,9 +128,18 @@ log = logging.getLogger(__name__)
 available = True
 
 # prefer C bindings over python when available
-# CSafeDumper causes test failures under python3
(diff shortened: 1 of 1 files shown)


dwoz:
There are also failing on the Fluorine Windows (PY3) builds, so we should confirm there too.

dwoz (approved):

PR #50748, opened 2018-12-04
Added NXOS UT support - initially for nxos_upgrade.py

tstoner:
### What does this PR do?
Adds NXOS UT framework for the current nxos_upgrade.py module and for other future NXOS modules.

Enables the community make changes with a level of confidence they did not break expected behavior to the nxos_upgrade.py module. 

### Tests written?

Yes - This PR is the UT for the existing nxos_upgrade.py module.

### Commits signed with GPG?

No


tstoner:
Are the **No name 'nxos' in module 'tests.unit.modules'** lint errors because the **nxos** subdirectory is being added as part of this PR or is there another way to reference the objects in this sub-directory?

tests/unit/modules/test_nxos_upgrade.py
```diff
@@ -0,0 +1,366 @@
...
+        """ UT: nxos_upgrade module:upgrade method - input validation """
+
+        result = nxos_upgrade.upgrade('dummy-platform-image.bin', issu=1)
+        assert 'Input Error' in result
+
+    def test_check_upgrade_impact_backend_processing_error_500(self):
+
+        """ UT: nxos_upgrade module:check_upgrade_impact method - error HTTP code 500 """
+
+        for platform in self.platform_list:
```
mikewiebe: What is the rational for running this test for every platform in `self.platform_list`?  If I am reading this properly `backend_processing_error_500` is defined as part of the base data object `NXOSPlatform`.  It looks like we are running the same test n number of times for each platform.

mikewiebe:
@tstoner I tried to patch these changes in and run them and discovered that you are missing the `__init__.py` file under the `tests/unit/modules/nxos` directory.

I suspect this might help with the `No name 'nxos' in module 'tests.unit.modules' lint errors`

tests/unit/modules/test_nxos_upgrade.py
```diff
@@ -0,0 +1,366 @@
...
+    """ Test cases for salt.modules.nxos_upgrade """
+
+    platform_list = None
+
+    @staticmethod
+    def assert_platform_upgrade(condition, platform):
+
+        """ Assert platform upgrade condition and display appropriate chassis & images upon assertion failure """
+
+        assert bool(condition), "{0}: Upgrade {1} -> {2}".format(platform.chassis, platform.cimage, platform.nimage)
```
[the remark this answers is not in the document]
tstoner: This is how the module information refers to the the current image and the new image.  nimage is 'new image'.  Because of this I'd prefer to stick to the same name and just add a comment.

tests/unit/modules/test_nxos_upgrade.py
```diff
@@ -0,0 +1,366 @@
...
+
+        return {nxos_upgrade: {}}
+
+    def tearDown(self):
+        del self.platform_list
+
+    @staticmethod
+    def test_check_upgrade_impact_input_validation():
+
+        """ UT: nxos_upgrade module:check_upgrade_inpact method - input validation """
```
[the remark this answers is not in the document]
tstoner: Ack

tests/unit/modules/test_nxos_upgrade.py
tstoner: backend_processing_error_500 can be over-ridden by any of the derived platform classes.  Since the UT method must assume it is different from the base class.

mikewiebe: Ok.. so it's more about possible future differences across platforms and enabling the test to automatically pick up and test the changes

tstoner: At the end of the day that is correct.  Most likely this will not be changed but if it is then the intent is this UT method will not need to be changed.

mikewiebe:
👍 Looks good to me.

cachedout (approved):

PR #50753, opened 2018-12-05
Use a process pool to instantiate and start processes

Denislb:
On windows, to instantiate new minion on a new process, for a simple command like `test.ping` it took more than 3 seconds. I tried to find a way to avoid this overload.

### What does this PR do?
Instead of using multiprocessing.Process() to instantiate new minion and run salt command. We use a process pool.

### Previous Behavior
On windows, at every command on the minion, we launch a new process, instantiate the minion, and execute this command.

### New Behavior
At the beginning of salt-minion, we create a process pool sized by `process_count_max`. When a new command has to be run, we just use a process that already exists. If `process_count_max` does not exist, we use the previous behavior.

### Tests written?

No

### Commits signed with GPG?

No

### DIFF
--- a/salt/minion.py
+++ b/salt/minion.py
@@ -1150,7 +1150,7 @@ class Minion(MinionBase):
 
         # We want the process pool on our main process
         # Check if there is a max process count,
-        # if not we will use the same way to create a process as before, Process() 
+        # if not we will use the same way to create a process as before, Process()
         self.pool = None
         process_count_max = self.opts.get('process_count_max')
         if not hasattr(self, '_is_child') and pool and process_count_max > 0:
@@ -1590,7 +1590,6 @@ class Minion(MinionBase):
                 args=(instance, self.opts, data, self.connected),
                 name=data['jid']
             )
-
 
         # Reset current signals before starting the process in
         # order not to inherit the current signal handlers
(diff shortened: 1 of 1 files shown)


cachedout:
If we go this route, I think we want to make it configurable. My concern about using a process pool is the memory overhead that's going to be incurred by having this processes sitting around unused. This might be a totally reasonable approach for Windows, since the performance concerns there are quite valid, but I'm not sure it's a good tradeoff for Linux and it certainly wouldn't be for IoT devices.

Thoughts?

cachedout:
In speaking with @DmitryKuzmenko this morning, we think that https://github.com/saltstack/salt/pull/49398 will address this and may be a better long-term solution.

DmitryKuzmenko:
Yep. @Denislb we have a slightly different approach with the following advantages: it uses tornado IPC sockets to transfer data that is more effective than multiprocessing queues used by the process pool, also it takes care of the event bus handling some of messages like module reload, pillar update and so on.
#49398 contains this approach together with other work on Tornado 5.0. @dwoz is working on splitting it out to a separate branch. So just take us some time we will have a PR for that soon.
Thank you for your work.

cachedout:
@Denislb It looks like the agreement on the core maintainers team is to close this in favor of #49398. We appreciate you sending us this work, however.

PR #50756, opened 2018-12-05
missing minion result data bug.

GwiYeong:
### What does this PR do?
Fix missing minion's return_data problem that occurs timeout exception for api request.
The problem has been occured by timing issue between calling `get_events` and handling return data.
This problem is same #45874

When salt-api received the minion return data, future's result are setted and handle the data. 
If salt-api is receiving another minion return data while handling the data, salt-api missing the data.
Current version of tornado salt-api can miss minions' return events while processing ([here](https://github.com/saltstack/salt/blob/2017.7/salt/netapi/rest_tornado/saltnado.py#L993-L1022)) a lastely received minion return event.

Therefore every single return events from salt minions should be subscribed before firing an event.

And I think this commit should be ported to the 2018.3 branch.

#45874

### Tests written?

No

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/netapi/rest_tornado/saltnado.py
+++ b/salt/netapi/rest_tornado/saltnado.py
@@ -921,14 +921,27 @@ class SaltAPIHandler(BaseSaltAPIHandler):  # pylint: disable=W0223
         '''
         Dispatch local client commands
         '''
-        # Generate jid before triggering a job to subscribe all returns from minions
-        chunk['jid'] = salt.utils.jid.gen_jid()
+        # Generate jid and find all minions before triggering a job to subscribe all returns from minions
+        chunk['jid'] = salt.utils.jid.gen_jid() if not chunk.get('jid', None) else chunk['jid']
+        minions = set(self.ckminions.check_minions(chunk['tgt'], chunk.get('tgt_type', 'glob')))
+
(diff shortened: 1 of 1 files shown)


salt/netapi/rest_tornado/saltnado.py
```diff
@@ -921,14 +921,27 @@ def _disbatch_local(self, chunk):
         '''
         Dispatch local client commands
         '''
-        # Generate jid before triggering a job to subscribe all returns from minions
-        chunk['jid'] = salt.utils.jid.gen_jid()
+        # Generate jid and find all minions before triggering a job to subscribe all returns from minions
+        chunk['jid'] = salt.utils.jid.gen_jid() if not chunk.get('jid', None) else chunk['jid']
```
mattp-: i think you want gen_jid(self.opts)

GwiYeong: the function `def get_jid(opt=None)`  is exist on `develop` branch only.
on `2017.7`, `get_jid` doesn't has `opt` options.

mattp-: my bad, you're right :+1:

PR #50760, opened 2018-12-05
salt.runners.pillar: fix special case for masters pillar

mattp-:
### What does this PR do?
when asking for the pillar of the running minion in master context it
talks on a '_master' to the minion_id, which will be a cache miss for
grains. if this is asked for, just generate it via salt.loader.grains

salt-run pillar.show_pillar foo_master didnt work

### Tests written?
No

### Commits signed with GPG?
No


cachedout (approved):
Makes sense. Nice catch.

PR #50763, opened 2018-12-05
state.file_keyvalue: fix proper recognition of a dict parameter

teizz:
### What does this PR do?
Turns out, when feeding parameters from an .sls file instead of directly to the state module through a test means dicts are actually of type ``salt.utils.odict.OrderedDict`` instead of ``dict``. Though still an instance of ``dict``, I believe this is a better way of checking for correct input.

No issue opened for this.

### Previous Behavior
Running the file.keyvalue state with key_values populated with a dictionary would actually complain no key_values where supplied.

### New Behavior
Correctly identifies key_values as being an instance of a dictionary and will continue to process all key/value pairs.

### Tests written?
Yes, additional checks to also check feeding a dictionary instead of just a key and value parameter.

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/states/file.py
+++ b/salt/states/file.py
@@ -4710,11 +4710,11 @@ def keyvalue(
     if not name:
         return _error(ret, 'Must provide name to file.keyvalue')
     if key is not None and value is not None:
-        if type(key_values) is dict:
+        if isinstance(key_values, dict):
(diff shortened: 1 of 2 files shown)


cachedout (approved):
Seems sensible.

PR #50764, opened 2018-12-05
Non-root building and signing of packages on Debian family

dmurphy18:
### What does this PR do?
Allows for building and signing Debian (and Raspbian and latest Ubuntu 18.04) packages on AWS and VirtualBox, which were failing as root user due to changes in latest versions of gpg-agent.  Ubuntu 14.04 and Ubuntu 16.04 are still built as root, due to gpg-agent version supported in those versions.

With recent Debian 9 (9.5 on VirtualBox, from 9.0 on AWS) the gpg-agent does not start correctly if launched by root user, hence the need to build and sign as a non-root user.  This is due to changes in gpg-agent utilizing systemd and the associated Unix standard sockets failing to start correctly if root.  This results in failing to store the passphrase and hence signing failures.  Also with recent changes the key fingerprint is now used in re-signing with debsign.

### Previous Behavior
Signing packages Debian 9.0 and above would fail on AWS, 9.5 would fail on VirtualBox.  Debian 9.2 on OpenNebula and VirtualBox would succeed.

### New Behavior
Signing Debian 9 packages now succeed on AWS when signing as a non-root user.

### Tests written?
No, use of salt-pack, salt-pack-py3 and salt-auto-pack si sufficient testing

### Commits signed with GPG?
No

### DIFF
--- a/salt/modules/debbuild.py
+++ b/salt/modules/debbuild.py
@@ -877,11 +877,7 @@ def make_repo(repodir,
                     number_retries = timeout / interval
                     times_looped = 0
                     error_msg = 'Failed to debsign file {0}'.format(abs_file)
-                    if __grains__['os'] in ['Debian'] or (__grains__['os'] in ['Ubuntu'] and __grains__['osmajorrelease'] >= 18):
-                        cmd = 'debsign --re-sign -k {0} {1}'.format(local_key_fingerprint, abs_file)
-                        log.debug("DGM make_repo debian gpg2 debsign cmd \'{0}\'".format(cmd))
-                        retrc |= __salt__['cmd.retcode'](cmd, runas=runas, cwd=repodir, use_vt=True, env=env)
-                    else:
+                    if __grains__['os'] in ['Ubuntu'] and __grains__['osmajorrelease'] < 18:
                         cmd = 'debsign --re-sign -k {0} {1}'.format(keyid, abs_file)
                         log.debug("DGM make_repo ubuntu gpg2 debsign cmd \'{0}\'".format(cmd))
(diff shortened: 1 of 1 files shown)


salt/modules/debbuild.py
```diff
@@ -264,6 +273,14 @@ def _create_pbuilders(env):
     if env_overrides and not env_overrides.isspace():
         with salt.utils.files.fopen(pbuilderrc, 'a') as fow:
             fow.write(salt.utils.stringutils.to_str(env_overrides))
+    cmd = "chown {0}:{0} {1}".format(runas, pbuilderrc)
```
cachedout: Is there a reason that `file.chown` isn't being used here instead?

salt/modules/debbuild.py
```diff
@@ -314,13 +331,77 @@ def make_src_pkg(dest_dir, spec, sources, env=None, saltenv='base'):
...
+    _create_pbuilders(env, runas)
     tree_base = _mk_tree()
     ret = []
     if not os.path.isdir(dest_dir):
         os.makedirs(dest_dir)
 
+    # ensure directories are writable
+    root_user = 'root'
+    retrc = 0
+    cmd = "chown {0}:{0} {1}".format(runas, tree_base)
```
cachedout: Same question here about `file.chown`

salt/modules/debbuild.py
dmurphy18: Went looking for it and missed it in the doc's, will make change

salt/modules/debbuild.py
dmurphy18: same as above

dwoz (approved):

salt/modules/debbuild.py
cachedout: Ah, OK. That makes sense.

PR #50771, opened 2018-12-06
Fix failing opensuse tests

terminalmage:
This fixes:

```
integration.states.test_pkg.PkgTest.test_pkg_cap_001_installed
integration.states.test_pkg.PkgTest.test_pkg_cap_002_already_installed
integration.states.test_pkg.PkgTest.test_pkg_cap_003_installed_multipkg_with_version
integration.states.test_pkg.PkgTest.test_pkg_cap_004_latest
integration.states.test_pkg.PkgTest.test_pkg_cap_005_downloaded
integration.states.test_pkg.PkgTest.test_pkg_cap_006_uptodate
```

Resolves https://github.com/saltstack/salt-jenkins/issues/1222.

### DIFF
--- a/tests/integration/states/test_pkg.py
+++ b/tests/integration/states/test_pkg.py
@@ -44,7 +44,7 @@ _PKG_TARGETS = {
 }
 
 _PKG_CAP_TARGETS = {
-    'Suse': [('w3m_ssl', 'w3m')],
+    'Suse': [('perl(ZNC)', 'znc-perl')],
 }
 
 _PKG_TARGETS_32 = {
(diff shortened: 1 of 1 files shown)


terminalmage:
PRs won't show the true fix here, we'll have to confirm once merged and fluorine branch builds are run again on opensuse.

tests/integration/states/test_pkg.py
```diff
@@ -853,18 +855,20 @@ def test_pkg_cap_002_already_installed(self, grains=None):
         self.assertFalse(version)
         self.assertFalse(realver)
 
-        # install the package already
-        ret = self.run_state('pkg.installed', name=realpkg, refresh=False)
+        try:
+            # install the package already
+            ret = self.run_state('pkg.installed', name=realpkg, refresh=False)
```
dwoz: should we check this response for a failure?

terminalmage: I don't think it's 100% necessary. The aim of the two state runs here is to check that we get a no-op result the second time through, since the install should have passed the first time through and installed the package.

It may help in troubleshooting if the first install fails though, so I went ahead and added it :+1:

dwoz (approved):

cachedout (approved):

PR #50772, opened 2018-12-06
create user without password

mikewiebe:
### What does this PR do?

Fixes a bug with the nxos.user_present state. The state should allow a user with no password (password=None).  NX-OS allows creation of a user without a password but it’s essentially unusable

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/states/nxos.py
+++ b/salt/states/nxos.py
@@ -86,6 +86,10 @@ def user_present(name, password=None, roles=None, encrypted=False, crypt_salt=No
 
     old_user = __salt__['nxos.cmd']('get_user', username=name)
 
+    if not old_user and password is None:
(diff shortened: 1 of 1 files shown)


cachedout (approved):
LGTM

mikewiebe:
@cachedout Hold off on merging this.  Chasing down what I just now discovered may not be an issue with this PR but want to be certain.

teizz:
Looking at the code it would disallow creating a user without a password, but wouldn't a user without a password still be usable if it has an ssh-key or x509 certificate associated with it?

mikewiebe:
@teizz This is actually a really good point.  I will close this PR and instead document the possible workflows.  Thanks for reviewing!

Issue #48132, opened 2018-06-14
Dynamic File Roots


BenjaminSchiborr:
### Description of Issue/Question
In our current environment we are using a mixture of information coming out of a git repository and from a local salt file server. Historically all information was always taken from the saltenv base environment (as in the master branch of the git repository and files that were locally on disk configured as the base saltenv)

However, we are moving forward to a new versioning model for git - requiring us to pull information from specific tags/branches. Thus, e.g. we call a salt runner providing a saltenv as a parameter. We have adapted our master configuration to then pull only from a git branch/tag that is equal to the saltenv. 

This, however, does not seem to work with the local file server. It seems that you are required to always specify specific saltenvs for file_roots in the master config.

Example:
Imagine having a branch X in git. You are calling salt via "salt-run [...] saltenv=X". In that case anything in the local salt file server would not be found, unless you also specifically mentioned X in the master config. 

### Question
The question is if something like *__env__* also exists for file_roots - enabling you to specify one path on a local file server for *all* saltenvs?

### Setup
Please find below a sanitized master configuration showing the issue:
```
state_top: salt://top.sls

fileserver_backend:
  - git
  - roots
file_roots:
  base:
    - /srv/salt
    - /srv/tops
pillar_roots:
  base:
    - /srv/pillar
  __env__:
    - /srv/pillar
pillarenv_from_saltenv: True

ext_pillar_first: True
ext_pillar:
  - git:
    - __env__ gitlab:Automation/salt:
      - root: pillar

gitfs_remotes:
  - gitlab:Automation/salt:
    - root: states

# Merge strategies
pillar_source_merging_strategy: recurse
top_file_merging_strategy: same
pillar_merge_lists: True
```

### Steps to Reproduce Issue
1. Have local file server with an sls for a runner
2. Have a git repository with a branch X. Ensure that the runner requires something that is on branch X
3. Ensure master config allows for dynamic usage of git branches via saltenv, but only specify base for file_roots
4. Call salt-run with the runner against a saltenv X
5. Salt-run will fail, because it cannot find the runner

### Versions Report
Salt Version:
           Salt: 2017.7.5
 
Dependency Versions:
           cffi: Not Installed
       cherrypy: Not Installed
       dateutil: Not Installed
      docker-py: Not Installed
          gitdb: 0.6.4
      gitpython: 1.0.1
          ioflo: Not Installed
         Jinja2: 2.8
        libgit2: Not Installed
        libnacl: Not Installed
[...]

PR #46309, opened 2018-03-02
Support dynamic pillar_root environment

bdrung:
Allow to specify a `__env__` pillar_root directory that applies equally to all environments that are not explicitly specified.

fixes #20581

PR #50784, opened 2018-12-07
Allow dynamic file_roots

sathieu:
### What does this PR do?
Implements Dynamic File Roots:
```
file_roots:
  base:
    - /srv/salt
  __env__:
    - /srv/default/salt
```

Fixes: #48132. Similar to #46309 for `pillar_roots`.

### Previous Behavior
Using roots and gitfs with feature branches is not possible, unless you restrict those branch names to match file_roots envs.

### New Behavior
You can add a default file_roots.

### Tests written?
Yes.

### Commits signed with GPG?
Yes

### DIFF
--- a/doc/ref/configuration/master.rst
+++ b/doc/ref/configuration/master.rst
@@ -2599,6 +2599,8 @@ can have multiple root directories. The subdirectories in the multiple file
 roots cannot match, otherwise the downloaded files will not be able to be
 reliably ensured. A base environment is required to house the top file.
 
+As of 2018.3.5 and 2019.2.1, it is possible to have `__env__` as a catch-all environment.
+
 Example:
 
 .. code-block:: yaml
@@ -2612,6 +2614,8 @@ Example:
       prod:
         - /srv/salt/prod/services
         - /srv/salt/prod/states
+      __env__:
+        - /srv/salt/default
 
 .. note::
     For masterless Salt, this parameter must be specified in the minion config
--- a/salt/state.py
+++ b/salt/state.py
@@ -3767,12 +3767,14 @@ class BaseHighState(object):
         statefiles = []
         for saltenv, states in six.iteritems(matches):
             for sls_match in states:
-                try:
+                if saltenv in self.avail:
                     statefiles = fnmatch.filter(self.avail[saltenv], sls_match)
-                except KeyError:
-                    all_errors.extend(
-                        ['No matching salt environment for environment '
-                         '\'{0}\' found'.format(saltenv)]
+                elif '__env__' in self.avail:
+                    statefiles = fnmatch.filter(self.avail['__env__'], sls_match)
+                else:
+                    all_errors.append(
+                        'No matching salt environment for environment '
+                        '\'{0}\' found'.format(saltenv)
                     )
                 # if we did not found any sls in the fileserver listing, this
                 # may be because the sls was generated or added later, we can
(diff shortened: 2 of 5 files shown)


cachedout:
@sathieu We would want to put this into the develop branch. Right now it's on a bug fix branch which is not where it would ultimately get merged. You may wish to switch this. :)

sathieu:
Ok. Anyway it doesn't work. Any idea how to fix this?

does this means that 2018.3.4 is coming soon?

cachedout:
@sathieu What do you mean by "it doesn't work"? Do you mean that this feature isn't doing what you intend it to or are you referring to some other functionality that isn't working as expected?

sathieu:
@cachedout My PR doesn't work. I have other problems but they may come from my specific version (2018.3.3 + several PRs).

sathieu:
Now working. And rebased on top of #51211.

dwoz (approved):

sathieu:
@terminalmage can youbplease review? Without this, using gitfs with file_roots is limited to a list of branches

salt/state.py
```diff
@@ -3768,7 +3768,10 @@ def render_highstate(self, matches):
         for saltenv, states in six.iteritems(matches):
             for sls_match in states:
                 try:
-                    statefiles = fnmatch.filter(self.avail[saltenv], sls_match)
+                    if saltenv not in self.avail and '__env__' in self.avail:
+                        statefiles = fnmatch.filter(self.avail['__env__'], sls_match)
+                    else:
+                        statefiles = fnmatch.filter(self.avail[saltenv], sls_match)
```
terminalmage: The old code used a try/except, but your changes introduce a membership check, which seems to make the try/except redundant. In cases where the saltenv is not present and `__env__` is not configured in `file_roots`, we do both a membership check _and_ a lookup, which we _know will fail_, and we also incur the overhead of raising and handling the exception. IMO it's better to use the results of the membership check to determine whether or not the saltenv could be found, perhaps like this:

```python
if saltenv in self.avail:
    statefiles = fnmatch.filter(self.avail[saltenv], sls_match)
else:
    if '__env__' in self.avail:
        statefiles = fnmatch.filter(self.avail['__env__'], sls_match)
    else:
        all_errors.append(
            'No matching salt environment for environment '
             '\'{0}\' found'.format(saltenv)
        )
```

terminalmage (changes requested):

sathieu:
I've added doc in `doc/ref/configuration/master.rst`. @terminalmage please review...

terminalmage (approved):

sathieu:
@terminalmage @dwoz You both approved. Anything missing?

doc/ref/configuration/master.rst
```diff
@@ -2599,6 +2599,8 @@ can have multiple root directories. The subdirectories in the multiple file
 roots cannot match, otherwise the downloaded files will not be able to be
 reliably ensured. A base environment is required to house the top file.
 
+As of 2018.3.4, it is possible to have `__env__` as a catch-all environment.
```
terminalmage: Can you change this to "2018.3.5 and 2019.2.1"? This missed the branching for both 2018.3.4 and 2019.2.0. Other than that I have no problem merging it.

sathieu:
@terminalmage you can merge, I've updated the versions.

sathieu:
@terminalmage @dwoz thanks to you both!

PR #50786, opened 2018-12-07
fix default ipc_write_buffer always is 0 if not set ipc_write_buffer …

qcha0:
### What does this PR do?
The ipc_write_buffer always is 0 if not set ipc_write_buffer in config file

In my case, the memory of IPCMessagePublisher process use a lot of memory and has been growing
After I review the code, I found that it was caused by not setting the limit of write buffer

### Tests written?
No

### Commits signed with GPG?

Yes/No

### DIFF
--- a/salt/config/__init__.py
+++ b/salt/config/__init__.py
@@ -3812,8 +3812,6 @@ def apply_minion_config(overrides=None,
 
     if overrides.get('ipc_write_buffer', '') == 'dynamic':
         opts['ipc_write_buffer'] = _DFLT_IPC_WBUFFER
-    if 'ipc_write_buffer' not in overrides:
-        opts['ipc_write_buffer'] = 0
 
     # Make sure hash_type is lowercase
     opts['hash_type'] = opts['hash_type'].lower()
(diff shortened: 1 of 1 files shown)


terminalmage (commented):
Can you give more information? We already have a method for calculating the default IPC write buffer size in `salt/config/__init__.py`.

Agnewee:
I knew it, the `ipc_write_buffer` defailt is `_DFLT_IPC_WBUFFER`
but in `apply_master_config()` and `apply_minion_config`, if the `ipc_write_buffer` is not specified in the configuration file, the  `ipc_write_buffer` always was `0`
```
    def apply_minion_config(overrides=None,
                                            defaults=None,
                                            cache_minion_id=False,
                                            minion_id=None):
        ...
        if overrides.get('ipc_write_buffer', '') == 'dynamic':
            opts['ipc_write_buffer'] = _DFLT_IPC_WBUFFER
        if 'ipc_write_buffer' not in overrides:
            opts['ipc_write_buffer'] = 0
```
In this case, the  `_DFLT_IPC_WBUFFER` does not take effect.

cachedout:
I think this fix is correct.

cachedout (approved):

PR #46542, opened 2018-03-14
Fixed GPG pillar decryption.

DmitryKuzmenko:
### What does this PR do?
1. Returned back the ability to decrypt raw encrypted data containing no
begin/end marks.
2. Remove trailing newlines if decrypted in a new 'multi' way
implemented in #45781

### What issues does this PR fix or reference?
Fix https://github.com/saltstack/salt-jenkins/issues/828
Update #45781

### Previous Behavior (before #45781)
pillar:
```
#!yaml|gpg
foo:
  bar: <encrypted data>
  baz: |
    -----BEGIN PGP MESSAGE-----
    <encrypted data>
    -----END PGP MESSAGE-----
  quux: |
    this is my complex pillar value
    it has a password:
    -----BEGIN PGP MESSAGE-----<encrypted password>-----END PGP MESSAGE-----
    and salt:
    -----BEGIN PGP MESSAGE-----<encrypted salt>-----END PGP MESSAGE-----
```
will be rendered as
```
foo:
  bar: <decrypted data>
  baz: <decrypted data>
  quux: <decrypted password>
```
`gpg` will render only the first ciphertext and strip the rest of plaintext and the second ciphertext

### Old Behavior (after #45781)
The same pillar:
```
#!yaml|gpg
foo:
  bar: <encrypted data>
  baz: |
    -----BEGIN PGP MESSAGE-----
    <encrypted data>
    -----END PGP MESSAGE-----
  quux: |
    this is my complex pillar value
    it has a password:
    -----BEGIN PGP MESSAGE-----<encrypted password>-----END PGP MESSAGE-----
    and salt:
    -----BEGIN PGP MESSAGE-----<encrypted salt>-----END PGP MESSAGE-----
```
will be rendered as
```
foo:
  bar: <encrypted data>
  baz: <decrypted data>\n
  quux: |
    this is my complex pillar value
    it has a password:
    <decrypted password>
    and salt:
    <decrypted salt>
```
Raw value for `bar` is not decrypted. Decrypted value for `baz` has unwanted newline symbol at the end.

### New Behavior
pillar:
```
#!yaml|gpg
foo:
  bar: <encrypted data>
  baz: |
    -----BEGIN PGP MESSAGE-----
    <encrypted data>
    -----END PGP MESSAGE-----
  quux: |
    this is my complex pillar value
    it has a password:
    -----BEGIN PGP MESSAGE-----<encrypted password>-----END PGP MESSAGE-----
    and salt:
[...]

PR #50810, opened 2018-12-11
Don't try to decrypt values without PGP header line

### LINKED ISSUE #50809: Warnings from salt.pillar.gpg in Fluorine

It looks like some changes in #45781 or #46542 broke the `salt.pillar.gpg` https://docs.saltstack.com/en/latest/ref/pillar/all/salt.pillar.gpg.html:

```
% tail -2 etc/salt/minion
ext_pillar:
  - gpg: {}

% cat pillar/test.sls
key: value

% LANG=C salt-call --local pillar.items
[WARNING ] Could not decrypt cipher value, received: b'gpg: no valid OpenPGP data found.\n[GNUPG:] NODATA 1\n[GNUPG:] NODATA 2\ngpg: decrypt_message failed: eof\n'
[WARNING ] Could not decrypt cipher value, received: b'gpg: no valid OpenPGP data found.\n[GNUPG:] NODATA 1\n[GNUPG:] NODATA 2\ngpg: decrypt_message failed: eof\n'
local:
    ----------
    key:
        value
```

If I restore the `salt/renderers/gpg.py` back to the `2013.3.3` state, everything works as expected.

### Versions Report

Salt fluorine branch, revision 241741a6d0104ffec9d2aaf29bb579e945c0c39a

CC: @secumod @DmitryKuzmenko

DmitryKuzmenko:
### What does this PR do?
Partially revert logic accidentally introduced in #46542.

Fixes #50809 

### Previous Behavior
I've added the logic that tries to decrypt values having no `-----BEGIN/END PGP MESSAGE-----` header/footer lines in #46542.

### New Behavior
Return the text without marks as is.

### Tests written?
Updated

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/renderers/gpg.py
+++ b/salt/renderers/gpg.py
@@ -372,7 +372,6 @@ def _decrypt_ciphertexts(cipher, translate_newlines=False):
         # it will conain unexpected trailing newline.
         return ret.rstrip('\n')
     else:
-        # Possibly just encrypted data without begin/end marks
         return cipher
 
 
--- a/salt/renderers/gpg.py
+++ b/salt/renderers/gpg.py
@@ -373,7 +373,7 @@ def _decrypt_ciphertexts(cipher, translate_newlines=False):
         return ret.rstrip('\n')
     else:
         # Possibly just encrypted data without begin/end marks
-        return _decrypt_ciphertext(cipher)
+        return cipher
 
 
 def _decrypt_object(obj, translate_newlines=False):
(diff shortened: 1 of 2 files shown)


salt/renderers/gpg.py
```diff
@@ -373,7 +373,7 @@ def _decrypt_ciphertexts(cipher, translate_newlines=False):
         return ret.rstrip('\n')
     else:
         # Possibly just encrypted data without begin/end marks
```
secumod: Do we need this comment here now?

cachedout: Seems like this should be removed.

DmitryKuzmenko: @secumod yes, you're right. Thank you for review.

cachedout (approved):

max-arnold:
@cachedout The issue was reported against fluorine branch, this PR needs to be backported

PR #50826, opened 2018-12-11
salt/modules/opkg: add restart_services kwargs option

dmondrik:
The commands to install and upgrade (and remove) all support the option, which
is useful for clients that will query the restartcheck module and initiate a
system reboot if needed. This omits unnecessary work and makes install faster.

Signed-off-by: Dan Mondrik <dan.mondrik@ni.com>

### What does this PR do?
Some clients will perform a restartcheck after an install, and reboot the system if needed.
For these clients, the service auto-restart during the install operation is unnecessary.

### Previous Behavior
Services using upgraded files would always be restarted, even if a reboot was needed.

### New Behavior
If the user passes --restart_services=False then services will not be restarted.

### Tests written?
N/A

### Commits signed with GPG?
Yes/No

### DIFF
--- a/salt/modules/opkg.py
+++ b/salt/modules/opkg.py
@@ -101,14 +101,18 @@ def _process_restartcheck_result(rs_result):
     '''
     if 'No packages seem to need to be restarted' in rs_result:
         return
+    reboot_required = False
     for rstr in rs_result:
         if 'System restart required' in rstr:
             _update_nilrt_restart_state()
             __salt__['system.set_reboot_required_witnessed']()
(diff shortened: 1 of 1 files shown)


rares-pop (approved):
The change looks good, but the semantic is a bit tricky.

I would expect 'restart-services=False' to mean they never get restarted. Or 'restart-services=True' to mean they always get restarted. I'm not sure what a good name would be though.

cachedout:
I agree with @rares-pop. What about `always_restart_services`?

PR #50828, opened 2018-12-11
Fix issues with lgpo state and util

twangboy:
### What does this PR do?
Fix a py3 issue in the win_lgpo_netsh salt util. It didn't work at all on Py3
Make the state case insensitive for policy lookups. You can set them case-insensitive, but the state was case sensitive, causing `test=True` to report changes erroneously

### Tests written?
No

### Commits signed with GPG?
Yes/No

### DIFF
--- a/salt/states/win_lgpo.py
+++ b/salt/states/win_lgpo.py
@@ -114,6 +114,7 @@ import salt.utils.json
 
 # Import 3rd party libs
 from salt.ext import six
+from requests.structures import CaseInsensitiveDict
 
 log = logging.getLogger(__name__)
 __virtualname__ = 'lgpo'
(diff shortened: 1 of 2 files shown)


salt/states/win_lgpo.py
```diff
@@ -250,6 +251,9 @@ def set_(name,
     for policy_section, policy_data in six.iteritems(pol_data):
         pol_id = None
         if policy_data and policy_data['output_section'] in current_policy:
+            # Make the subdict keys case insensitive
+            current_policy[policy_data['output_section']] = CaseInsensitiveDict(
```
cachedout: You could probably do this with `salt.utils.dictupdate` and avoid the import.

twangboy: The purpose is to do a case-insensitive search of the dictionary keys. I don't see how to implement that with `dictupdate`. Am I missing something?

dwoz (approved):

salt/states/win_lgpo.py
cachedout: You'd have to implement it on top of dictupdate. It's fine if you don't want to though. Either way works.

PR #50838, opened 2018-12-12
[WIP] Cherry-pick minion worker changes

dwoz:
### What does this PR do?

Separate minion worker and tornado 5.0 changes by cherry-picking from #49398. This PR add's a multiprocessing pool for minion jobs.

This is an alternate implementation of #50753 

### Tests written?

No

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/config/__init__.py
+++ b/salt/config/__init__.py
@@ -1501,6 +1501,7 @@ DEFAULT_MINION_OPTS = {
     'ssh_merge_pillar': True,
     'server_id_use_crc': False,
     'disabled_requisites': [],
+    'worker_threads': 3,
 }
 
 DEFAULT_MASTER_OPTS = {
--- a/salt/minion.py
+++ b/salt/minion.py
@@ -870,7 +870,7 @@ class SMinion(MinionBase):
                 self.utils,
                 self.serializers)
         self.rend = salt.loader.render(self.opts, self.functions)
-#        self.matcher = Matcher(self.opts, self.functions)
+        # self.matcher = Matcher(self.opts, self.functions)
         self.matchers = salt.loader.matchers(self.opts)
         self.functions['sys.reload_modules'] = self.gen_modules
         self.executors = salt.loader.executors(self.opts)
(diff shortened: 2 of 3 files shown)


garethgreenaway:
@dwoz Looks like there is a lint failure here.

salt/config/__init__.py
```diff
@@ -1501,6 +1501,7 @@ def _gather_buffer_space():
     'ssh_merge_pillar': True,
     'server_id_use_crc': False,
     'disabled_requisites': [],
+    'worker_threads': 3,
```
cachedout: I still believe this should be 5.

salt/minion.py
```diff
@@ -870,7 +867,7 @@ def gen_modules(self, initial_load=False):
                 self.utils,
                 self.serializers)
         self.rend = salt.loader.render(self.opts, self.functions)
-#        self.matcher = Matcher(self.opts, self.functions)
+        # self.matcher = Matcher(self.opts, self.functions)
```
cachedout: Why not just remove this?

salt/minion.py
```diff
@@ -1532,37 +1548,13 @@ def _handle_decoded_payload(self, data):
...
-
-        # TODO: remove the windows specific check?
-        if multiprocessing_enabled and not salt.utils.platform.is_windows():
-            # we only want to join() immediately if we are daemonizing a process
-            process.join()
-        else:
-            self.win_proc.append(process)
+            pull_uri = os.path.join(self.opts['sock_dir'], 'worker_pull.ipc')
+        log.error("CLINT SEND %s", data)
+        client = salt.transport.ipc.IPCMessageClient(pull_uri)
```
cachedout: Is it really necessary to instantiate a new client on every receipt of a payload? Couldn't this just be reused?

salt/minion.py
```diff
@@ -1601,192 +1596,217 @@ def _target(cls, minion_instance, opts, data, connected):
...
-                    ret['return'] += ' Possible reasons: \'{0}\''.format(
-                        minion_instance.function_errors[mod_name]
+                single_ret = ret
+
+            allow_missing_funcs = any([
+                minion_instance.executors['{0}.allow_missing_func'.format(executor)](function_name)
+                for executor in executors
+                if '{0}.allow_missing_func' in minion_instance.executors
+            ])
+            if function_name in minion_instance.functions or allow_missing_funcs is True:
```
cachedout: This could really use some refactoring. :)

dwoz:
I don't think we need this right now. I will leave the branch around for a while just in case.

Issue #50791, opened 2018-12-07
[fluorine] multi-master failover minion stops when masters not available


Ch3LL:
### Description of Issue/Question
When a minion is configured to use multi-master failover and both masters are down for more then 5 minutes the minion will stop.

### Setup
/etc/salt/minion

```
master_type: failover
master_alive_interval: 10
master:
  - master1
  - master2
```

### Steps to Reproduce Issue
1. Start master1, master2 and minion
2. verify you can run a `test.ping` from master1

```
[root@master1 /]# salt '*' test.ping
minion1:
    True
[root@master1 /]# 
```

3. stop both masters `pkill salt-master`. kill master2 first then master1
4. wait a couple minutes and you will see the minion ends up stopping

```
[DEBUG   ] Trying to connect to: tcp://172.24.0.4:4506
[INFO    ] (u'Master %s could not be reached, trying next next master (if any)', u'master2')
[ERROR   ] No master could be reached or all masters denied the minion's connection attempt.
[ERROR   ] Error while bringing up minion for multi-master. Is master at [u'master1', u'master2'] responding?
[ERROR   ] Minion unable to successfully connect to a Salt Master.  Exiting.
[INFO    ] Stopping the Salt Minion
[ERROR   ] Minion unable to successfully connect to a Salt Master.  Exiting.
[INFO    ] Shutting down the Salt Minion
[DEBUG   ] Stopping the multiprocessing logging queue listener
[DEBUG   ] closing multiprocessing queue
[DEBUG   ] joining multiprocessing queue thread
[DEBUG   ] Stopped the multiprocessing logging queue listener
The Salt Minion is shutdown.
Minion unable to successfully connect to a Salt Master.  Exiting.
[root@minion1 /]# 
```

### Versions Report

```
Salt Version:
           Salt: 2018.11.0-2383-g0ffd1e5

Dependency Versions:
           cffi: Not Installed
       cherrypy: Not Installed
       dateutil: Not Installed
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.7.2
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: 0.28.2
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.5.6
   mysql-python: Not Installed
      pycparser: Not Installed
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: Not Installed
[...]

Issue #50854, opened 2018-12-13
[fluorine] windows stops when cannot connect to master


Ch3LL:
### Description of Issue/Question
When the master is not running and the windows minion attempts to connect to the master the process shutsdown and does not continue to try to connect to the master again.

### Setup
I installed windows like so:

`powershell -file build_env_2.ps1`
`pip install -e .`

### Steps to Reproduce Issue
1. attach a window minion to a master
2. verify you can run commands from the master to windows minion
3. stop both master and minion
4. start minion `salt-minion -ldebug` and wait for about 5 minutes

and you will see it shutdown:

```
[DEBUG   ] SaltReqTimeoutError, retrying. (1/7)
[ERROR   ] Minion unable to successfully connect to a Salt Master.  Exiting.
[INFO    ] Stopping the Salt Minion
[ERROR   ] Minion unable to successfully connect to a Salt Master.  Exiting.
[INFO    ] Shutting down the Salt Minion
The Salt Minion is shutdown.
Minion unable to successfully connect to a Salt Master.  Exiting.
PS C:\salt>
```

### Versions Report

```
PS C:\salt> salt --versions-report
Salt Version:
           Salt: 2018.11.0-166-gc16d0f3

Dependency Versions:
           cffi: 1.10.0
       cherrypy: unknown
       dateutil: 2.6.1
      docker-py: Not Installed
          gitdb: 2.0.5
      gitpython: 2.1.3
          ioflo: Not Installed
         Jinja2: 2.9.6
        libgit2: Not Installed
        libnacl: 1.6.1
       M2Crypto: Not Installed
           Mako: 1.0.6
   msgpack-pure: Not Installed
 msgpack-python: 0.4.8
   mysql-python: Not Installed
      pycparser: 2.17
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: Not Installed
         Python: 2.7.14 (v2.7.14:84471935ed, Sep 16 2017, 20:25:58) [MSC v.1500 64 bit (AMD64)]
   python-gnupg: 0.4.1
         PyYAML: 3.12
          PyZMQ: 16.0.3
           RAET: Not Installed
          smmap: 2.0.5
        timelib: 0.2.4
        Tornado: 4.5.1
            ZMQ: 4.1.6

System Versions:
           dist:
         locale: cp1252
        machine: AMD64
        release: 2016Server
         system: Windows
        version: 2016Server 10.0.14393  Multiprocessor Free
```

[...]

PR #50855, opened 2018-12-13
Let minions try to connect to master indefinitly

dwoz:
### What does this PR do?

Do not exit when the minion can't connect to a master. 

#50854

### Tests written?

No

### Commits signed with GPG?

Yes


dwoz:
@Ch3LL @DmitryKuzmenko @garethgreenaway 

I've tested that this also resolves #50791. Pun intended ;)

damon-atkins:
Random sleep time adjustment?  i.e. to prevent all of the minions trying at about the same time to connect, and causing a wave effect against the master when it returns.   It not I suggest its added to the code. + or - 10% of the interval time. ` interval - (interval * 0.05) + random(interval* 0.1)`

Ch3LL:
@dwoz i believe when @garethgreenaway was trying to fix this initially it was to fix multi-master with dns (dns names as the master in the minion config). Did you happen to test with dns or ip addresses?

Ch3LL:
nevermind. I looked it up here: https://github.com/saltstack/salt/pull/49764  and it seems it should only work when using the new `retry_dns_count` configuratino

dwoz:
@Ch3LL I my testing I used DNS with resolvable and un-resolvable hosts.

@damon-atkins That is a good idea but this is a bugfix. Changes should go in a feature.

cachedout:
> @damon-atkins That is a good idea but this is a bugfix. Changes should go in a feature.

Given that this is still pre-release, this might still be feasible, IMHO.

dwoz:
> > @damon-atkins That is a good idea but this is a bugfix. Changes should go in a feature.
> 
> Given that this is still pre-release, this might still be feasible, IMHO.

I'd still like to see that go in a different PR to keep added functionality separate from the bugfix found here.

DmitryKuzmenko (approved):

cachedout:
We need to ensure this does not regress this fix, IMHO: 

https://github.com/saltstack/salt/commit/e66dc18805812f550d44585aa5d8af954e7eca97

dwoz:
@cachedout We discussed this on slack and @garethgreenaway agreed that this change is what we want. Just waiting for @garethgreenaway to sign off by approving this PR.

garethgreenaway (approved):

PR #50864, opened 2018-12-14
fix typo in file.managed documentation

remijouannet:
Hello,

just a little typo that hide a wonderful feature in the documentation :)

(if that can be backport to 2016/2017/2018 documentation, that would be great)


thatch45:
Thanks for the fix!

Issue #50848, opened 2018-12-13
salt.modules.mysql.py uses `*exc` exception unpacking syntax incompatible with python3


5uper5hoot:
### Description of Issue/Question
`salt.modules.mysql` uses `*` unpacking syntax on caught exceptions which is incompatible with python3.

### Setup
Installed salt with `bootstrap-salt.sh` and args ` -P -x python3 git 2018.3`. MySQL successfully installs.

Used `mysql_user.present` to create a new user.

### Steps to Reproduce Issue
Example state:
```
"mysql_user_sstuser_localhost":
  mysql_user:
    - present
    - name: "sstuser"
    - host: "localhost"
    - password: "password"
    - connection_unix_socket: /var/run/mysqld/mysqld.sock
    - connection_user: root
    - connection_pass: rootpass
    - connection_db: mysql
    - require:
      - service: mysql
```

Exception:

```
[DEBUG   ] An exception occurred in this state: format() argument after * must be an iterable, not OperationalError
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/salt/modules/mysql.py", line 333, in _connect
    dbc = MySQLdb.connect(**connargs)
  File "/usr/lib/python3/dist-packages/MySQLdb/__init__.py", line 86, in Connect
    return Connection(*args, **kwargs)
  File "/usr/lib/python3/dist-packages/MySQLdb/connections.py", line 204, in __init__
    super(Connection, self).__init__(*args, **kwargs2)
_mysql_exceptions.OperationalError: (1045, "Access denied for user 'sstuser'@'localhost' (using password: YES)")

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/salt/state.py", line 1919, in call
    **cdata['kwargs'])
  File "/usr/lib/python3/dist-packages/salt/loader.py", line 1918, in wrapper
    return f(*args, **kwargs)
  File "/usr/lib/python3/dist-packages/salt/states/mysql_user.py", line 142, in present
    **connection_args):
  File "/usr/lib/python3/dist-packages/salt/modules/mysql.py", line 1251, in user_exists
    if not verify_login(user, password, **connection_args):
  File "/usr/lib/python3/dist-packages/salt/modules/mysql.py", line 2254, in verify_login
    dbc = _connect(**connection_args)
  File "/usr/lib/python3/dist-packages/salt/modules/mysql.py", line 335, in _connect
[...]

PR #50869, opened 2018-12-14
[2017.7] Adding a test to ensure exception handling is correct

garethgreenaway:
### What does this PR do?
Adding a test to ensure exception handling is correct.

#50848 

### Tests written?
Yes

### Commits signed with GPG?
Yes

### DIFF
--- a/tests/unit/modules/test_mysql.py
+++ b/tests/unit/modules/test_mysql.py
@@ -282,6 +282,23 @@ class MySQLTestCase(TestCase, LoaderModuleMockMixin):
     def test_query(self):
         self._test_call(mysql.query, 'SELECT * FROM testdb', 'testdb', 'SELECT * FROM testdb')
 
+    def test_query_error(self):
+        def mock_execute(cur, query):
+            '''
+            Mock of info method
+            '''
+            raise MySQLdb.OperationalError(9999, 'Something Went Wrong')
+
+        connect_mock = MagicMock()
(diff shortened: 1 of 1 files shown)


tests/unit/modules/test_mysql.py
```diff
@@ -282,6 +282,23 @@ def test_free_slave(self):
...
+            '''
+            Mock of info method
+            '''
+            raise MySQLdb.OperationalError(9999, 'Something Went Wrong')
+
+        connect_mock = MagicMock()
+        with patch.object(mysql, '_connect', connect_mock):
+            with patch.dict(mysql.__salt__, {'config.option': MagicMock()}):
+                with patch.object(mysql, '_execute') as execute_mock:
+                    execute_mock.side_effect = mock_execute
```
terminalmage: You can actually setup the side_effect right in the init for `MagicMock()`:

```python
MagicMock(side_effect=MySQLdb.OperationalError(9999, 'Something Went Wrong'))
```

This would remove the need to A) write a separate function, and B) set the side_effect manually after the MagicMock is initialized.

It'll work perfectly fine as-is, so I'm approving the PR, but just something to keep in mind.

terminalmage (approved):

tests/unit/modules/test_mysql.py
garethgreenaway: Updating to the more concise approach. Thanks!

PR #50879, opened 2018-12-16
Add new slack returner based on incoming webhooks

cdalvaro:
### What does this PR do?

Add a new slack returner based on incoming webhooks

slack_webhook_returner

### Tests written?

No

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/returners/slack_webhook_return.py
+++ b/salt/returners/slack_webhook_return.py
@@ -57,20 +57,21 @@ To use the alternative configuration, append '--return_config alternative' to th
 from __future__ import absolute_import, print_function, unicode_literals
 
 # Import Python libs
-import pprint
 import logging
+import json
 
 # pylint: disable=import-error,no-name-in-module,redefined-builtin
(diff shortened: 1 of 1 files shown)


damon-atkins:
What is the difference between salt/returners/slack_returner and this code?

cdalvaro:
> So can you keep them compatible? What else needs to be changed in salt to work with the new API?

Do you mean, what changes are necessary to be made in `slack_returner` in order to make compatible with the new API too?

If that is your question, I think that changes should be minimum. Just adding the parameter for the webhook url and based on whether a webhook is passed or an api_token choose the post method.

thatch45 (approved):

cdalvaro:
Will this PR finally be approved? Or it is needed some change?

cdalvaro:
Thank you very much, @thatch45 !

thatch45:
Sure thing, but I likely should not have merged this without tests. If you can get some tests in it would be greatly appreciated!

PR #50885, opened 2018-12-17
fix a del in cloud vmware that crash when the key isn't there

remijouannet:
so I don't really understand what I've patch but this line always throw a KeyError when i try to deploy a machine on vcenter

(if that can be backport to 2017/2018 that would be great )

### DIFF
--- a/salt/cloud/clouds/vmware.py
+++ b/salt/cloud/clouds/vmware.py
@@ -2889,7 +2889,8 @@ def create(vm_):
         log.debug('config_spec set to:\n%s', pprint.pformat(config_spec))
 
     event_kwargs = vm_.copy()
-    del event_kwargs['password']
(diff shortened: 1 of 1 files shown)


gtmanfred (approved):

gtmanfred:
@saltstack/team-core this is ready and should probably be backported.

thatch45:
Thanks @remijouannet !
Sorry, with the holidays we have been a little behind in PR merges and reviews.
Thanks @gtmanfred !

remijouannet:
no problem, any chance for this bugfix to be include in the future 2017.7.9 ? we haven't upgrade to 2018 yet

thatch45:
I will have to double check the release schedule, but that would be a reasonable ask :)

PR #50896, opened 2018-12-18
salt.modules.event tweaks

mattp-:
### What does this PR do?
- use  __opts__['__role'] with salt.utils.event.get_event, so that in master/orch it does the right thing
- expose a timeout kwarg when longer than default is needed
- use keep_loop=True to not close ioloop when done

n/a

## Tests written?
No

### Commits signed with GPG?
No

### DIFF
--- a/salt/modules/event.py
+++ b/salt/modules/event.py
@@ -31,7 +31,7 @@ def _dict_subset(keys, master_dict):
     return dict([(k, v) for k, v in six.iteritems(master_dict) if k in keys])
 
 
-def fire_master(data, tag, preload=None):
+def fire_master(data, tag, preload=None, timeout=60):
     '''
     Fire an event off up to the master server
 
@@ -74,7 +74,7 @@ def fire_master(data, tag, preload=None):
         for master in masters:
             channel = salt.transport.client.ReqChannel.factory(__opts__, master_uri=master)
             try:
-                channel.send(load)
+                channel.send(load, timeout=timeout)
                 # channel.send was successful.
                 # Ensure ret is True.
                 ret = True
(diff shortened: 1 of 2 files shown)


salt/modules/event.py
```diff
@@ -171,6 +177,9 @@ def send(tag,
     :type with_env_opts: Specify ``True`` to include ``saltenv`` and
         ``pillarenv`` values or ``False`` to omit them.
 
+    :param timeout: maximum duration to wait to connect to Salt's
+        IPCMessageServer in seconds. defaults to 60s
```
cachedout: s/defaults/Defaults/

salt/runners/event.py
```diff
@@ -76,5 +76,6 @@ def send(tag, data=None):
     '''
     data = data or {}
     event = salt.utils.event.get_master_event(__opts__, __opts__['sock_dir'],
-                                              listen=False)
+                                              listen=False,
+                                              keep_loop=True)
```
cachedout: Can you explain this change a little more please?

cachedout (changes requested):

dwoz:
@mattp- Ping. Can you responded to @cachedout's comments?

salt/runners/event.py
mattp-: when keep_loop is false the event destroy() also calls io_loop.close(), I didn't see the need for that to happen in this case. that said I don't think leaving this out would cause any issues either.

salt/modules/event.py
mattp-: fixed

mattp-:
@dwoz done, sorry. have a few PRs I need to give some TLC to in the queue

dwoz:
@mattp- Thanks, I've updated the branch so the tests should pass now. We'll get this merged after that happens.

dwoz (approved):

cachedout:
No changelog entry? :)

PR #50905, opened 2018-12-18
add niceness control for subprocesses of salt master

mattp-:
### What does this PR do?
under high load these can be useful to tweak for ensuring critical parts
of salt aren't being resource starved.

### Tests written?
No

### Commits signed with GPG?
No

### DIFF
--- a/salt/master.py
+++ b/salt/master.py
@@ -200,6 +200,10 @@ class Maintenance(salt.utils.process.SignalHandlingMultiprocessingProcess):
         # Init any values needed by the git ext pillar
         self.git_pillar = salt.daemons.masterapi.init_git_pillar(self.opts)
 
+        if self.opts['maintenance_niceness'] and not salt.utils.platform.is_windows():
+            log.info('setting Maintenance niceness to %d', self.opts['maintenance_niceness'])
+            os.nice(self.opts['maintenance_niceness'])
+
         self.presence_events = False
         if self.opts.get('presence_events', False):
             tcp_only = True
@@ -481,6 +485,11 @@ class FileserverUpdate(salt.utils.process.SignalHandlingMultiprocessingProcess):
         Start the update threads
         '''
         salt.utils.process.appendproctitle(self.__class__.__name__)
+
+        if self.opts['fileserver_update_niceness'] and not salt.utils.platform.is_windows():
+            log.info('setting FileServerUpdate niceness to %d', self.opts['fileserver_update_niceness'])
+            os.nice(self.opts['fileserver_update_niceness'])
+
         # Clean out the fileserver backend cache
         salt.daemons.masterapi.clean_fsbackend(self.opts)
 
@@ -920,6 +929,10 @@ class ReqServer(salt.utils.process.SignalHandlingMultiprocessingProcess):
                             'when using Python 2.')
                 self.opts['worker_threads'] = 1
 
+        if self.opts['req_server_niceness'] and not salt.utils.platform.is_windows():
+            log.info('setting ReqServer_ProcessManager niceness to %d', self.opts['req_server_niceness'])
+            os.nice(self.opts['req_server_niceness'])
+
         # Reset signals to default ones before adding processes to the process
         # manager. We don't want the processes being started to inherit those
         # signal handlers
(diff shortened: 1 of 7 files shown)


cachedout:
Way cool!

We'll definitely need docs written before we can get this in, though.

mattp-:
@cachedout working on it, i'll ping ... not you :) (its likely not going to be by eod tomorrow) when tests and docs have this ready to go. thanks

cachedout:
@mattp- I turn into a pumpkin tomorrow at midnight. ;) I'm sure @thatch45 has your back though!

thatch45:
Yes, I really like this addition, a great idea! Once we have docs in for this it should be good to merge!

austinpapp:
for completeness, do we see any value in add niceness tuning for fileserverupdate, maintenance, etc?

salt/master.py
```diff
@@ -1126,6 +1130,18 @@ def run(self):
         Start a Master Worker
         '''
         salt.utils.process.appendproctitle(self.name)
+
+        # if we inherit jeq_server level without our own, reset it
```
austinpapp: s/jeq_server/req_server

dwoz:
Personally I'd like to understand the driver for this better. Specifically which processes are causing the high load and why? Generally, I think of niceness as something that should be set overall for a program by something outside of the said program (not by the program itself).

cachedout:
@dwoz Since we're forking our own procs and those processes have different roles, I think there is a good argument for the application to take control over setting nice values appropriately per-process.

sagetherage:
@mattp, @dwoz raises a valid point to further the discussion and we would like to put this through our RFC process. Can you please look at this template and submit an RFC PR? https://github.com/saltstack/salt/blob/develop/rfcs/0000-template.md

mattp-:
@sagetherage sure I can try to get that together. but basically @dwoz to answer your question any orchestration heavy workload is going to cause cpu contention on the master. introduce orchestrations that call orchestrations, it adds up quickly. what we've seen is the ipcmessageserver and publisher was falling over/significantly falling behind on loads that are on the order of 30-40x core count of the server. using nice to bump priority for it was a large boon in performance under this load.

dwoz:
@mattp Thanks for the explanation. I think this is an easy and immediate fix for the issue you have described.

Longer term, I think we should try to determine what is causing the load. Then look into ways of mitigating the problem. Perhaps we could batch the orchestrations or otherwise optimize the process so it will play nice on it's own.

My concern for having salt set it's own nice settings via the config file is that it seems like a non-standard way for system administrators to set niceness for a process. It'd be nice if we could separate and name processes accordingly so that this could be done outside of salt in the standard way for the specific os. This would be `/etc/security/limits.conf` on Debian Linux.

Overall, I'm not opposed to this getting merged. However, I think we could discuss and possibly implement some of the solutions above which would allow us to remove these setting in the future. We can discuss all this more in the RFC if you make one.

thatch45:
While I agree that limits.conf and doing this from a systems level makes sense, I also think that having this in here is a good idea. Since salt is spinning up so many procs I think it will be wise to allow users to set this up in this way

austinpapp:
Any movement on this? FWIW, this code has been deployed in a production setting and working ... nicely.

cachedout:
Did the ability to nice the maintenance proc get in here? IMHO, it would be great to be able to dial back the priority of that proc in certain cases.

thatch45:
I will update the branch, I am good merging this, but we need the tests to be passing

thatch45:
While I also agree that we should be able to nice the maintenance process I am ok getting that into a future PR

dwoz:
@mattp- What is the status of this?

mattp-:
@dwoz i think the last standing issue was putting together an RFC which I've not had time to do. the original goal of this was to let the mworkers run under a lower priority to not starve the other processes of the salt-master - I don't think you can use limits.conf to do this externally (if you can, I'd happily close this and move to that). if we are ok with foregoing the RFC I can get the maintenance added as a switch. what do you think?

dwoz:
@mattp- Based on the conversation above, I am fine with this provided it's an optional setting. From the looks of the changes it is indeed optional. I think all that is needed is to add the option for the maintenance process.

Ch3LL:

mattp-:
@dwoz added some docs, the opt for maintenance, as well as mworker_queue which I also missed. thanks

salt/master.py
```diff
@@ -1132,6 +1145,24 @@ def run(self):
...
         '''
         salt.utils.process.appendproctitle(self.name)
+
+        # if we inherit req_server level without our own, reset it
+        if not salt.utils.platform.is_windows():
+            enforce_mworker_niceness = True
+            if self.opts['req_server_niceness']:
+                if salt.utils.user.get_user() == 'root':
+                    log.info('%s decrementing inherited ReqServer niceness to 0', self.name)
+                    log.info(os.nice())
```
s0undt3ch: A log call with just the `os.nice()` on the middle of a bunch of logs makes the logged information un-useful.
Either remove this call or join it with the debug call before:
`log.info('%s decrementing inherited ReqServer niceness to 0. Previous value: %s', self.name, os.nice())`

s0undt3ch (changes requested):

salt/master.py
mattp-: this was an accidental include actually, there doesnt seem to be a way to get 'current' nicess. ive removed that log.info(os.nice()) call

mattp-:
rebased on current neon

dwoz:
@mattp- Can you rebase this against the `master` branch please?

dwoz:
@twangboy The `master-port` label is for PRs that *are* porting something to master, not PRs that need to be ported.t

@mattp- This needs rebasing against master.

twangboy:
I rebased this against the master branch: https://github.com/saltstack/salt/pull/57365

sagetherage:
@dwoz it looks like we can close this PR in favor of the rebase from Shane in teh PR https://github.com/saltstack/salt/pull/57365, yes?

PR #50907, opened 2018-12-19
change to correct unit

qcha0:
### What does this PR do?
I think the grains.core get the unit of memory is MB, but there need Bytes


thatch45:
What is the goal of this change? Is there a bug that is being exhibited by the format of the total memory?

Agnewee:
If not set `ipc_write_buffer` in config file, the `ipc_write_buffer` will get from `_gather_buffer_space()`
when `HAS_PSUTIL` was `False`, calculate total_mem by `grains.core._memdata` and unit is MB.
In fact the `ipc_write_buffer` unit must be Bytes. So we need convert MB to Bytes.

dwoz (approved):
LGTM - I validated that IOStream expects MB and that our supported platforms are returning bytes for mem_total.

PR #50922, opened 2018-12-19
Added bypass features to pkg.installed

justindesilets:
### What does this PR do?
This feature adds a bypass_file and bypass_file_contents option to the pkg.installed function of the pkg.py state file. The intent is to offer a method for validating a package is already installed without having to reach all the way back to the package manager. This is to address the 99.99% of state runs we do where we are only applying config changes, but still touch the package manager because of requisite statements for packages.

This issue is to address the issue of state runs taking a long time to run when having to reach out to the package manager to validate packages. In profile runs, on average, I would see 5+ seconds per package that was being validated. At smaller scale this is not a problem. At larger scale, this starts to add up and cause overlapping state runs and timeout issues. In trying to follow best practices, we try and build states so they have all self resolving requisites in them. However, when trying to just push a one line change out, you end up confirming multiple packages spread out between different related states. Below are differences in times between with and without the bypass:

```
# salt \* state.apply profile --out=profile
    ---------------------------------------------------
    |      name     |     mod.fun     | duration (ms) |
    ---------------------------------------------------
    | /etc/ntp.conf |   file.managed  |       57.0170 |
    ---------------------------------------------------
    |      ntpd     | service.running |       85.6650 |
    ---------------------------------------------------
    |      ntp      |  pkg.installed  |     5793.2000 |
    ---------------------------------------------------
# salt \* state.apply profile --out=profile
    ---------------------------------------------------
    |      name     |     mod.fun     | duration (ms) |
    ---------------------------------------------------
    |      ntp      |  pkg.installed  |        0.7380 |
    ---------------------------------------------------
    | /etc/ntp.conf |   file.managed  |       31.7340 |
    ---------------------------------------------------
    |      ntpd     | service.running |       51.0020 |
    ---------------------------------------------------
```

Here is an example of the output from a state run on a vagrant:
```
bslm-022.corp.day.com:
----------
          ID: install_salt_minion
    Function: pkg.installed
        Name: salt-minion
      Result: True
     Comment: pkg.installed was bypassed as bypass_file /run/salt-minion.pid was present
     Started: 15:00:51.483411
    Duration: 6.218 ms
     Changes:
----------
          ID: salt_minion_config
    Function: file.managed
        Name: /etc/salt/minion
      Result: True
     Comment: The file /etc/salt/minion is in the correct state
     Started: 15:00:51.495112
    Duration: 61.426 ms
     Changes:
----------
          ID: start_salt_minion
    Function: service.running
        Name: salt-minion
      Result: True
     Comment: The service salt-minion is already running
     Started: 15:00:51.560070
    Duration: 126.148 ms
     Changes:

Summary for bslm-022.corp.day.com
------------
Succeeded: 3
Failed:    0
------------
Total states run:     3
Total run time: 193.792 ms
```

### Tests written?

No

### Commits signed with GPG?

No


terminalmage (approved):

max-arnold:
@justindesilets Please add some unit-tests in a follow-up PR. Right now (abbfad1) the `bypass_file_contains` argument seems to be broken:

```
                File "/usr/lib/python2.7/dist-packages/salt/state.py", line 1998, in call
                  ret = self.states[cdata['full']](*cdata['args'], **cdata['kwargs'])
                File "/usr/lib/python2.7/dist-packages/salt/loader.py", line 2056, in wrapper
                  return f(*args, **kwargs)
                File "/usr/lib/python2.7/dist-packages/salt/states/pkg.py", line 1587, in installed
                  with salt.utils.fopen(bypass_file) as bypass_file_open:
              AttributeError: 'module' object has no attribute 'fopen'
```

PR #50938, opened 2018-12-20
Implement version_clean & check_extra_requirements in opkg module

andzn:
### What does this PR do?
Implement pkg.version_clean & pkg.check_extra_requirements in the opkg execution module

### Previous Behavior
The time it took to apply a state that contained many (in our case ~400) of packages, when all of the packages were installed was ~11 **minutes**. 
The package state module checks if the pkg module has defined the version_clean and check_extra_requirements_functions in a for loop that iterates through all the packages. Therefore, if the two functions are not defined, the lazy loader will always fail, leading to poor performance.
An alternative would be to cache the function before entering the for loop but that is a little bit trickier to do since, the actual call to version_clean & check_extra_requirements functions are done via a different execution module called pkg_resource. We would not want to cache pkg_resource.version_clean, as that one will always exist. We would want to cache pkg.version_clean because that one might not exist for every 'pkg' module.

### New Behavior
The time it takes to apply a state that contains many (in our case ~400) of packages, when all of the packages are installed is 9 **seconds**. 

### Tests written?
No

### Commits signed with GPG?
No


dwoz (approved):
@andzn Nice find and great ticket description. Thanks!

PR #50950, opened 2018-12-20
user: add subuid/subgid user function

toanju:
This adds subuid/subgid configuration to support user namespaces in Linux.


twangboy (approved):

garethgreenaway:
@toanju Nice addition!  Some unit tests would be appreciated, then we can get this merged in.

toanju:
thanks, I will add them later this week.

salt/modules/useradd.py
```diff
@@ -927,3 +927,111 @@ def _getpwall(root=None):
...
+    CLI Examples:
+
+    .. code-block:: bash
+
+        salt '*' user.add_subuids foo
+        salt '*' user.add_subuids foo first=105000
+        salt '*' user.add_subuids foo first=600000000 last=600100000
+    '''
+    if __grains__['kernel'] != 'Linux':
+        log.warning("'subuids' are only supported on GNU/Linux hosts.")
```
terminalmage: The warnings you've added in each of these functions are unnecessary, as this module will not be loaded on non-Linux minions (see [here](https://github.com/saltstack/salt/blob/501079fffc95f50f338d50eaa9dfd1078c8bc0c5/salt/modules/useradd.py#L40-L47)).

The warnings in each of the new functions you've added should be removed.

terminalmage (changes requested):

salt/modules/useradd.py
terminalmage: Oh, disregard... I was under the impression that pw was used on the other BSD variants, and AIX was also added to the list of OSes which uses this module.

My bad!

terminalmage (approved):

Ch3LL:
It needs to be merged into master. This PR https://github.com/saltstack/salt/pull/57020 attempts to do that but requires a changelog and test coverage as stated here: https://github.com/saltstack/salt/pull/57020#issuecomment-1013442471

PR #50953, opened 2018-12-21
Add new NVMe grain to display minion storage initiator

sdodsley:
### What does this PR do?
Adds new grain to display NVMe storage iniitator for hosts.

Support for Linux only at this time.

With the integration of storage modules to allow configuration
of 3rd party external storage arrays, these values are needed to
enable the full automation of storage provisioning to minions.

### New Behavior
Add new grain:
nvme_grain

### Tests written?

Yes

### Commits signed with GPG?

No

### DIFF
--- a/salt/grains/nvme.py
+++ b/salt/grains/nvme.py
@@ -1,0 +1,64 @@
+# -*- coding: utf-8 -*-
+'''
+Grains for NVMe Qualified Names (NQN).
+
+.. versionadded:: Flourine
+
+To enable these grains set `nvme_grains: True`.
+
+.. code-block:: yaml
+
+    nvme_grains: True
+'''
+# Import Python libs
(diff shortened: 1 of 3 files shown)


salt/grains/nvme.py
```diff
@@ -0,0 +1,65 @@
...
+    nvme_grains: True
+'''
+# Import Python libs
+from __future__ import absolute_import, print_function, unicode_literals
+
+import errno
+import logging
+
+# Import Salt libs
+import salt.modules.cmdmod
```
terminalmage: I can't see where in this module we are running shell commands. Do we actually need this import?

sdodsley: Correct. I'll remove it.

terminalmage (approved):

PR #50954, opened 2018-12-21
Document RFC process

cachedout:
cc: @sagetherage and @isbm

### What does this PR do?

Adds documentation regarding Salt's RFC process

### DIFF
--- a/doc/topics/development/contributing.rst
+++ b/doc/topics/development/contributing.rst
@@ -40,6 +40,27 @@ see the :ref:`Linting<pylint-instructions>` documentation for more information.
 
 .. _github-pull-request:
 
+New Features
+------------
+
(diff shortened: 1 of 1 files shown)


cachedout:
Hrm. Why did the Windows test fail here?

sagetherage:
@cachedout good question, I will need to dig into the test a bit give about an hour.

Ch3LL:
@dwoz or @twangboy can you take a look here. It looks like the windows tests here failed with: `00:22:54        'python.exe' is not recognized as an internal or external command,`

I also checked some other PR windows tests submitted today and it looks like they are hanging right before they attempt to run the tests. They have been running for 5 hours.

isbm (approved):

dwoz (approved):

Issue #50931, opened 2018-12-20
salt-run jobs.list_jobs OOM


Poil:
### Setup
My machine is a Virtual Machine with 16GB of memory, before calling jobs.list_jobs I have 15GB free.
I have 42 nodes connected on the Salt master.
I also use the returner pgjsonb (event_return and master_job_cache)

### Steps to Reproduce Issue
I query jobs list from yesterday and only one node match
```
salt-run jobs.list_jobs start_time='2018-12-20' search_target='tlsslt-p01-mst*'
```

It seems like all the data are get and python is filtering.

### Versions Report
```
Salt Version:
           Salt: 2018.3.3
 
Dependency Versions:
           cffi: 1.11.5
       cherrypy: unknown
       dateutil: 2.4.2
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.10
        libgit2: 0.26.5
        libnacl: Not Installed
       M2Crypto: 0.28.2
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.5.6
   mysql-python: Not Installed
      pycparser: 2.18
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: 0.26.4
         Python: 3.4.9 (default, Aug 14 2018, 21:28:57)
   python-gnupg: Not Installed
         PyYAML: 3.11
          PyZMQ: 15.3.0
           RAET: Not Installed
          smmap: Not Installed
        timelib: Not Installed
        Tornado: 4.4.2
            ZMQ: 4.1.4
 
System Versions:
           dist: centos 7.6.1810 Core
         locale: UTF-8
        machine: x86_64
        release: 3.10.0-957.1.3.el7.x86_64
         system: Linux
        version: CentOS Linux 7.6.1810 Core
```

Poil:
This seems to be already discussed here #21450 some work had been made but the discussion habe been closed before resolving all

Poil:
I've wrote get_jids_filter but this does not fix my issue.
I think we need to add the ability to delegate the filtering to the returners, easy I think for postgresql and other databases that support json

Ch3LL:
thanks for the PRs!

Poil:
I've cleaned #50936, not sure if we can have a better code or not, perhaps a real developer can look at it :)

Poil:
[...]

PR #50960, opened 2018-12-21
Returner native format

Poil:
### What does this PR do?
Use SQL to format the job return when the parameter `job_cache_filter` is set to `native`
Add the missing `get_jids_filter` function
Replace  MySQL keyword in the documentation

#50931

### Previous Behavior
Format the job return via a python function

### New Behavior
Format the job return in SQL

### Tests written?
No


salt/runners/jobs.py
```diff
@@ -305,68 +305,78 @@ def list_jobs(ext_source=None,
...
-    for item in ret:
-        _match = True
-        if search_metadata:
-            _match = False
-            if 'Metadata' in ret[item]:
-                if isinstance(search_metadata, dict):
-                    for key in search_metadata:
-                        if key in ret[item]['Metadata']:
-                            if ret[item]['Metadata'][key] == search_metadata[key]:
+    if returner in [ 'pgjsonb']:
```
cachedout: Coupling the type of returner in use to the jobs runner itself seems undesirable to me.

Poil: Today the jobs API is unusable for a real production, and no one download a database to filter it in his code, so I'm agree it's not very nice to have this hardcoded here but I don't find another easy solution. Have you got an idea ?

cachedout: Hmm. Could we mark certain returners as accepting a job_filter, perhaps via a configuration option? I'm totally in favor of the idea overall, but I want to make sure it's something we can support broadly and not just as a one-off.

Poil: Also, I don't understand why we don't store the job result like this and not how we want it when calling it again ?

Poil:
Hi,

We can add a function in each returners `get_jids_native`
We add an option like `returner.pgjsonb.filter: 'python' or 'native'`
And if the value is native we call `get_jids_native`.
My problem is I don't know how to call the `_get_options of salt/returners/****.py` in `salt/runners/jobs.py`, any tips ?

Best regards

salt/runners/jobs.py
```diff
@@ -305,68 +305,83 @@ def list_jobs(ext_source=None,
...
-        if search_metadata:
-            _match = False
-            if 'Metadata' in ret[item]:
-                if isinstance(search_metadata, dict):
-                    for key in search_metadata:
-                        if key in ret[item]['Metadata']:
-                            if ret[item]['Metadata'][key] == search_metadata[key]:
+    if __opts__.get('filter') == 'native':
+        fun = '{0}.get_jids_native'.format(returner)
+        if fun not in mminion.returners:
```
thatch45: If we have this config value set then this will prevent the returner from running at all even if this function is not in there

thatch45:
I am wondering if there is a way to instead of making a new returner function to just extend the existing returner function. Could we change get_jids to accept these options from the config file just for the postgres returner?

salt/runners/jobs.py
Poil: Yes, I just copy what was done for get_jids_filter

thatch45:
The function itself can remain small and we can break the subsequent calls out into other functions. I am trying to keep the calling interface simple. I am mostly saying that we should be able to make the native optimizations you are looking at native to the function without adding complexity to the returner interface.

DmitryKuzmenko:
@Poil do you plan to follow up Tom's suggestion?

Poil:
I should take some time for that in the next weeks

Poil:
I look, I'm sorry I don't know how to have something more clear, except spliting SQL build and moving the python filtering way outside the current function.

dwoz:
@Poil We are no longer merging to the `develop` branch, can you please rebase this to `master`?

DmitryKuzmenko:
@Poil do you plan to continue work on this?

sagetherage:
closing as abandoned by the author

PR #50961, opened 2018-12-21
Ignore retcodes for linux lvm calls

beornf:
### What does this PR do?
Ensures there is no error logging if the LVM physical volume, volume group or logical volume does not exist.

N/A

### Previous Behavior
Salt errors when creating a new physical volume, volume group or logical volume due to existence check.

### New Behavior
Salt LVM present states work the same without logging errors.

### Tests written?
No

### Commits signed with GPG?
Yes


dwoz (commented):
Would it be possible to detect the error when an lvm volume does not exist and handle the situation appropriately? My concern here is that by ignoring the return code we'll run into other issues that will be hidden and hard to troubleshoot.

beornf:
In this case I see two approaches to a change of the lvm shell command:
`pvdisplay -c pvname || [ $? -eq 5 ]` ignores the case when the physical volume doesn't exist instead of returning linux I/O error 5.
`pvdisplay -c | grep pvname ` lists all partitions and filters on the physical volume name thereby avoiding an error being returned.

dwoz:
@beornf I like `pvdisplay -c pvname || [ $? -eq 5 ]`

beornf:
I've made the change however the merge conflict is due to the return code now being ignored when creating physical volume, volume group or logical volume and also fixes this issue https://github.com/saltstack/salt/commit/922d192bf651150d339ff7f35ac70a6176c3b6cc.

Is this PR redundant or do you still have concerns about ignoring the return code leading to hidden issues?

dwoz:
@beornf We are no longer accepting PRs to the `develop` branch. This needs to be re-opened against the `master` branch. You can link to this PR from the new one to preserve any disscussion history. Sorry for the confusion.

PR #50969, opened 2018-12-21
Fix issue with LGPO when returning empty dict

twangboy:
### What does this PR do?
msgpack doesn't support CaseInsensitiveDict apparently

### Tests written?
No

### Commits signed with GPG?
Yes


s0undt3ch:
If it's that important, you can also teach msgpack I'm how to handle that dict...

twangboy:
I think this is better than importing requests just for CaseInsensitiveDict

s0undt3ch:
True.

PR #50973, opened 2018-12-23
[develop] Merge forward from fluorine to develop

garethgreenaway:
Merge branch 'fluorine' into develop

Conflicts:

  - salt/modules/win_lgpo.py
  - salt/templates/debian_ip/debian_eth.jinja
  - tests/unit/modules/test_debian_ip.py


MTecknology:
At a glance, it looks like @Ch3LL is right and there was a legitimate test
failure, but it could be because I wrote two sets of tests and one was much
better than the other.

Digging into this sounds like an unpleasant challenge... partly because it
seems like there's a whole lot of movement in a few directions and this
kinda thing is unfamiliar territory.

You mentioned the Debian template, but I don't see that file listed in the
merge. Is that because of the conflicts?

To reproduce this issue, should I be able to clone your branch and attempt
a simple merge from develop?

>

garethgreenaway:
@MTecknology In a previous merge-forward attempt there was a conflict with the template and I thought there was gong to be in this one as well, looks like it's all good now though so you can ignore that message for now :) The failing test I have an idea why it's failing and the fix, still need to push that up.

MTecknology:
Excellent! I love hearing that someone else is fixing things!! :P

If you run into troubles, I should have some time this week to help dig in.

dwoz (approved):
The windows tests have been failing for a while. We'll be focused on those when Fluorine is released.

PR #50974, opened 2018-12-24
[Fluorine] Unify salt-call/salt executor arguments

max-arnold:
#48344 adds `--module-executors` and `--executor-opts` arguments to `salt-call`. Unfortunately, the first one does not support lists (as it does in `salt` command):

```
% salt-call --local --module-executors='[direct_call, direct_call]' test.ping

Passed invalid arguments: Executor '[direct_call, direct_call]' is not available.

% salt MINION --module-executors='[direct_call, direct_call]' test.ping
MINION:
    True
```

And the second one seems to be ignored completely.

### New Behavior

This patch allows to specify these CLI arguments for `salt-call` in the same way as for `salt`.

```
% salt-call --local -l debug --module-executors='[splay, direct_call]' --executor-opts='{splaytime: 100}' test.ping 2>&1 | grep Splay
[DEBUG   ] Splay is sleeping 2 secs on test.ping
```

### Tests written?

No

### Commits signed with GPG?

No


max-arnold:
P.S. `salt-ssh` too should support these arguments (for consistency and better UX), but I haven't figured out how to add them.

max-arnold:
@Ch3LL Hi! Any chance of this to get merged into Fluorine?

terminalmage:
Since this was opened well after the feature freeze, I don't think so. It should be re-opened against develop.

max-arnold:
@terminalmage This is a fix against a feature that went into Fluorine

terminalmage:
Ahh, OK. @Ch3LL should this be opened against the RC branch then?

Ch3LL:
yes if it needs to be in for 2019.2.0. If not this will get in for 2019.2.1

Ch3LL (approved):

Issue #44495, opened 2017-11-11
RFE: Minion Authenticated Roles with salt-key & Auto Accept


damon-atkins:
### Part A)
Example of what salt-key could look like
```
salt-key -L
Accepted Keys:
webserver.example.com saltenv role1,role2
Denied Keys:
Unaccepted Keys:
something.example.com saltenv role1,role2,role3 IP  
Rejected Keys:

salt-key -l debug --accept-auto --command my_auto_accept_script.new  # force a run of auto sign and override the command in the master config file
```
(IP address is not stored its used for auto acceptance, or give the user a bit more info)
Roles are not gains or pillars they are a new data type which can be used in a top.sls And because the master stores the roles and the saltenv with the key (pki dir or a role dir), they can be considered secure.  i.e some developers may have access to the pillar dir or pillar git repo, but the roles and saltenv are safe in the pki dir (or its own dir).

### Part B)
The master running a script to see if it should accept or reject, allows a person to do anything they want.
Secure or Insecure.   Here are examples of what a script could do (Various levels of security):
- Check the DHCP server for PXE boot
- LDAP query to AD to see if the "COMPUTER" exists and if the IP it connect from matches. (it would require master recording IP address the connection came from)
- Reverse DNS Lookup of the IP and check the minion name against this (it would require master recording IP address the connection came from)
- AWS Query to see if it exists in AWS
- Check with a 3rd part as described above a few comments ago.  e.g. RightScale, Consul, SoftLayer
- Accept any minion connecting from a list of networks.
- Just accept all 

Something along the lines of  `command <minion id> <input data filename> <return data filename>`  everything else is passed as environment variables. e.g. salt conf dir,  location of the minions data receive (json) by the master (e.g. grains, IP etc) 
Exit Codes Returned:
10 = Accept
11 = Reject
12 = Manual i.e. let the human decide. with `salt-key`  OR Pass it on to the next script (if more then one provided)
13 = No Decision Yet, i.e. Ask again.  e.g. The data need to make a decision with, might be slow at getting populated (e.g. Windows system as not reach the stage it joins AD, or registered in consul). So script would like to be able to ask again, in a minute or two.   The script can implement a count of tries and change the exit code to 12 or 11 if the re-try count is too high.   The minion connecting again would be a good way to trigger auto accept script  again.

`<input data filename>` would be a json file containing secure/verified data and insecure data like grains.
`<return data filename>` would be a json file allowing saltenv to be changed or the roles to be changed.

[...]

PR #50976, opened 2018-12-24
RFC for remote clients

cachedout:
### What does this PR do?

Introduces an RFC to remove the restriction that the `salt` command be run from a local master and allow it to be run from anywhere.


garethgreenaway (approved):

rfcs/remote-client.md
```diff
@@ -0,0 +1,132 @@
...
+`salt-api` which provides a series of REST endpoints which may be accessed
+natively, or by using an alternative CLI designed specifically for use
+with `salt-api`, which is known as `pepper`.
+
+However, even between these option, there remain several serious drawbacks:
+
+ * A user cannot use either the `salt` tool or the `pepper` tool to send
+   commands to a mix of API- and non-API- endpoints.
+ 
+ * Using `pepper`, there is no way to target a given master from the
```
gtmanfred: This is not true, pepper can have multiple salt-api endpoints configured in the pepperrc config file as different sections of the ini config, and that can be used to target different masters based on the profile selected to be used.

gtmanfred (approved):

rfcs/remote-client.md
cachedout: Ah, I didn't know that. I will remove this section.

s0undt3ch (approved):

isbm:
@cachedout @gtmanfred @s0undt3ch @thatch45 Generally calling `salt ...` command to multiple masters is something that is indeed lacking. However, I am not sure the idea to have another pile of keys around to manage is optimal. Maybe it is though... I would go with that idea only if there would be a very good reason to let `salt` command be totally independent from whatever place it is called. Because it anyway _connects to the remote_ master[s]. And thus local master doesn't have to exist on the same host where `salt` command is called from. Potentially, every minion can call master then.

But is it indeed needed?

Instead, I would rather focus on master-to-master[s] grid, so then we already have authenticated multi-master network. And then whatever `salt` command _through_ their local master can send a message to any remote master or to all of them.

Another improvement here is that I would rather... drop the `-m` parameter. I find it redundant. Let me explain it a bit more. Why would you call a remote master at first place? Because it has minions you're interested in. Thus your target not really a _master_ but actually a specific _minion_ to call. That said, we can do it so:

```
salt \*@another-master.lan test.ping
```

Local master takes the `*` pattern, forwards it to `another-master.lan` host. So far, so good.

But then we have another problem with this approach: this still looks like the topology that is only segregating minions-per-master, even boldly assuming that there is only one particular master. Which would mean that it is pretty useless feature with simple multi-master setup, because here only a single master is responsible for specific subset of minions, one can access by `*` at that particular `another-master.lan` master. And if we have/could-have/would-have a multi-master setup that are load-balancing minions, then we rather have no idea to which particular master our minions get connected. And thus calling `salt \*@master` still doesn't mean it will call all the minions we want to reach out, unless messages aren't broadcasted to other masters behind the scenes.

To fix the above, here is a better syntax and it would actually work:

```
salt \* test.ping
```

Oh wait... :laughing:  OK, don't get me wrong here: that is how I would connect to remote master[s]. The difference would be behind the scenes, not in the CLI as actually proposed. Because of the following reasons:

- I don't care to which master I connect, as long as I am reaching particular minion.
- I expect my SLS state files are synchronised across the masters, so I do not operate in hostnames/masters, but environments. And thus I do not care from which master my environment reaches out my particular minions, as long as it does that.
- I configure my grid in the configuration. So my local master knows the network and knows the other peers already and keys are also already known and verified. So I do not need to care about another keys and manage them.

My points do not interfere with the ability to call different APIs either. What I propose is to reach out minions behind other masters, instead of reach out other masters to reach other minions behind them via remote clients to these masters.

Unless I missing something...

isbm (changes requested):
@cachedout @gtmanfred @s0undt3ch @thatch45 Generally calling `salt ...` command to multiple masters is something that is indeed lacking. However, I am not sure the idea to have another pile of keys around to manage is optimal. Maybe it is though... I would go with that idea only if there would be a very good reason to let `salt` command be totally independent from whatever place it is called. Because it anyway _connects to the remote_ master[s]. And thus local master doesn't have to exist on the same host where `salt` command is called from. Potentially, every minion can call master then.

But is it indeed needed?

Instead, I would rather focus on master-to-master[s] grid, so then we already have authenticated multi-master network. And then whatever `salt` command _through_ their local master can send a message to any remote master or to all of them.

Another improvement here is that I would rather... drop the `-m` parameter. I find it redundant. Let me explain it a bit more. Why would you call a remote master at first place? Because it has minions you're interested in. Thus your target not really a _master_ but actually a specific _minion_ to call. That said, we can do it so:

```
salt \*@another-master.lan test.ping
```

Local master takes the `*` pattern, forwards it to `another-master.lan` host. It is even SSHshy, in a way. OK, so far, so good.

But then we have another problem with this approach: this still looks like the topology that is only segregating minions-per-master, even boldly assuming that there is only one particular master. Which would mean that it is pretty useless feature with simple multi-master setup, because here only a single master is responsible for specific subset of minions, one can access by `*` at that particular `another-master.lan` master. And if we have/could-have/would-have a multi-master setup that are load-balancing minions, then we rather have no idea to which particular master our minions get connected. And thus calling `salt \*@master` still doesn't mean it will call all the minions we want to reach out, unless messages aren't broadcasted to other masters behind the scenes.

To fix the above, here is a better syntax and it would actually work:

```
salt \* test.ping
```

Oh wait... :laughing:  OK, don't get me wrong here: that is how I would connect to remote master[s]. The difference would be behind the scenes, not in the CLI as actually proposed. Because of the following reasons:

- I don't care to which master I connect, as long as I am reaching particular minion.
- I expect my SLS state files are synchronised across the masters, so I do not operate in hostnames/masters, but environments. And thus I do not care from which master my environment reaches out my particular minions, as long as it does that.
- I configure my grid in the configuration. So my local master knows the network and knows the other peers already and keys are also already known and verified. So I do not need to care about another keys and manage them.
- I reach out minions by _grains_ and other attributes, not by knowing to which master they are attached. Master is something that is not necessarily needs to drive current minion, especially if we are talking about HA, just like... you know... [here](https://docs.saltstack.com/en/latest/topics/targeting/grains.html), and then you just apply states from your environment.

        salt -G 'ec2_tags:environment:*production*'

How would Salt "grid" react to this? Same as it does now: return me the results of minions. With the difference that it would find the minions as masters would reply to the general client's broadcast with "Yes, I have this minion!" and others would just simply ignore the message.

As simple as that. And no need to change semantics.

My points do not interfere with the ability to call different APIs either. What I propose is to reach out minions behind other masters, instead of reach out other masters to reach other minions behind them via remote clients to these masters.

Unless I am missing something...

damon-atkins:
I have suggest in other issues. That it might be useful to support `username@hostname` which would represent minion not running as root/system.  The keys would be in the format of `username@hostname/minionid`.

damon-atkins:
See #44495 `RFE: Minion Authenticated Roles with salt-key & Auto Accept`

cachedout:
I understand what you're saying here, @isbm, but in my view being able to target a master is actually quite valuable. Think, for example, of a situation where a master plus a set of minions constitutes an application group. Being able to target by the group, as identified by only the master, seems to me like it would be very nice to have.

cachedout:
We'll have to agree to disagree then. :) 

While your workflow might always be limited to "always calling minions", I see cases where addressing targets by master might be desirable and I suspect many others will as well. I suppose these sorts of discussions are the reasons we have an RFC process so I'll be interested to hear what others think.

isbm:
Sure. :-) Not that I opposing here or supporting something. I just trying to understand your reasoning why this is _needed_ and where this is useful. Why we cannot just get minions into a groups during registration and this instantly forming these logical segregations. Especially because you are trying to turn master into something visible, while it supposed to be just a transparent "proxy" between command center (API or console etc) and minions.

mattp-:
I tend to lean toward @isbm's and @gtmanfred's thoughts on this RFC. I've coincidentally put some thoughts about how to approach this problem with what is provided today with salt. I think what makes most sense is to make saltnado/salt-api the integration point for external communication. A RestApiClient could be responsible for executing/returning results over remote (or even local) api. At that point states/saltmod would just need to be augmented for when a request is to be remote vs local, likewise for cli integration. You could use a shared minion data cache to establish if a given request needs to be routed remotely instead of explicitly specifying as well.

astronouth7303:
(disclaimer: I am currently in the process of writing an alternate salt rpc using ssh for transport and authentication.)

To be honest, the only concrete problem I'm seeing this solve is outputters. I don't take it at face value that `salt` cli only works locally is a problem. Especially if one of your goals is to enable a richer client ecosystem. (I already rewrote pepper once.)

In addition, afaict, I don't see a wire format defined. I must assume, then, the the current salt protocol (I believe there's three options, none documented particularly well, and must be configured on each node), each client must depend on salt itself, or salt must refactor to move the protocol code to another package. This effectively makes alternate language implementations impractical, limiting your client ecosystem.

If my rpc client depends on salt, I have the outputters locally and can wire up the outputters with salt-api without significant core changes.

At this point, no salt client I've seen supports dispatching a request to multiple endpoints and merging the results. As mentioned above, the usability of this feature makes topology assumptions.

A UI/UX quibble: salt remote clients to be introduced to the concept of multiple independent clusters (different BUs, different clients, isolated dev/prod).

One thing I don't see mentioned at all is runners&wheels. Afaik, runners&wheels can only be called by the same machine and same local user as the salt master. They also serve as the API entry point for salt cloud, salt virt, salt-key, monitoring, fileserver, and any operation that is not minion-oriented. If your goal is to lessen/remove the need for salt users to ssh/sudo, then I would consider any solution that has not at minimum considered and planned for runners & wheels to be woefully incomplete.

tl;dr: this proposal involves a lot of changes to salt's core for value I'm just not buying. All of this could be accomplished with a salt engine implementing well-designed RPC protocol.

isbm:
`<off-topic>`
> (disclaimer: I am currently in the process of writing an alternate salt rpc using ssh for transport and authentication.)

Not sure what would be the advantage of that in current Salt's architecture, but would be nice to see the RFC to that RPC, explaining motivation and design.

The only value here I would see is to have SaltSSH without copying all possible everything every time you call another machine, so then client-less Salt would be _truly_ one... But that seems not the case, right?

`/<off-topic>`

> At this point, no salt client I've seen supports dispatching a request to multiple endpoints and merging the results. As mentioned above, the usability of this feature makes topology assumptions.

Sounds like a potential quite significant security problem. :wink:

gtmanfred:
> To be honest, the only concrete problem I'm seeing this solve is outputters. I don't take it at face value that `salt` cli only works locally is a problem. Especially if one of your goals is to enable a richer client ecosystem. (I already rewrote pepper once.)

Also worth mentioning that pepper already wires up outputters if salt is installed alongside pepper.

mattp-:
@astronouth7303 I'd be very curious to hear more about your work. are you planning on making it public?

@gtmanfred on my long todo list is a 'wrapped' outputter for salt-api/pepper that just encodes the original requested output so that dependency isn't necessary.

astronouth7303:
@isbm @mattp- The project is https://gitlab.com/spirostack/spiroapi (Alpha: it works, but there's important features missing, and I need to do significant refactoring to enable those features). I would be happy to discuss it and answer any questions you have about it in another medium.

isbm:
@astronouth7303 thanks, but should we discuss this here?  I'm not sure how it is related to @cachedout 's RFC. BTW, Affero GPL requires Salt be also AGPL as far as I understand. In any case I would avoid anything of GPL* in case of Salt. Feel free to create another RFC and bring this up there.

astronouth7303:
I mentioned it because I felt like I should disclose that I'm working on a potentially conflicting project. I didn't want to discuss it here because it's off-topic and not about the RFC at hand.

cachedout:
What is the process for existing RFCs? Will they be migrated to the new repo? Are they subject to the  comment periods as outlined in  https://github.com/saltstack/salt-rfc/pull/1 ? If they are to be migrated, when can we expect that to happen?

cachedout:
@astronouth7303 The only thing that really prevents the existing client working from a remote machine right now, IIRC, is the fact that it reads the master's key off disk in order to include it in the request that it makes. It would be a trivial change as it stands today to allow that key (or keys) to be stored in an arbitrary location and sent along with the request. The real gotcha, though, is that I'm fairly sure that it's an unencrypted exchange right now and changing that would mean updating the master.

astronouth7303:
I'll reiterate, you express the goal:

> Salt could have a richer ecosystem of third-party client tool and libraries to interact with Salt masters if this limitation were removed. These could include smartphone clients, as well as a much easier path for Salt libraries which could interact with a Salt master in a direct manner.

Salt-api accomplishes this by using well-supported HTTP. It's problems come from:
* Insufficient documentation about the low structure
* Not great authentication options (user/password, Kerberos, or external tokens)

This proposal will have the opposite effect, since the wire protocol used internally by salt is basically entirely undocumented (like, it's based on 0MQ, but you can't go from a 0mq library to talking with salt with published documentation).

New criticism: the authentication in this is dependent on the master's private key. The private key that is the cornerstone of salt security. The private key that allows you to issue commands to any connected minion.

Maybe SaltStack should not encourage workflows that distribute this data to much-harder-to-secure user workstations.

cachedout:
> New criticism: the authentication in this is dependent on the master's private key.

Please re-read the original RFC. It discusses the current private key implementation and suggests a way around the issue you reference.

Regarding documentation on the wire protocol that's necessary to send a publication to a master, it could be quite easily documented it's nothing more than a dictionary serialized by message pack.

cachedout:
Closing to migrate to [SEP repo.](https://github.com/saltstack/salt-enhancement-proposals)

PR #50981, opened 2018-12-26
module swarm: catch the Docker API exceptions

madrisan:
### What does this PR do?
Catch the exceptions raised by the Docker API as documented  in [Docker SDK for Python](https://docker-py.readthedocs.io/en/stable/swarm.html).

None.

### Previous Behavior
The minion code ends with an unhandled exception if an error occurs for the Docker API.
For instance adding a node to a Swarm cluster twice leads to an exception. This is an incorrect behavior because does not permit to write idempotent states.

### New Behavior
An error message is reported that can be parsed by the caller code (a runner in my case).

### Tests written?
No

### Commits signed with GPG?
No

### DIFF
--- a/salt/modules/swarm.py
+++ b/salt/modules/swarm.py
@@ -32,6 +32,7 @@ import salt.utils.json
 
 try:
     import docker
+    import docker.errors
     HAS_DOCKER = True
 except ImportError:
     HAS_DOCKER = False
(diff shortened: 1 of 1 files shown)


dwoz (approved):

bryceml:
Having an unknown repository as a source breaks jenkins. If you would like this PR to make it in, please re-submit and link to the original.

PR #50997, opened 2018-12-28
Last singleton instance, closes.

s0undt3ch:
### What does this PR do?
See title.

### Tests written?

No

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/crypt.py
+++ b/salt/crypt.py
@@ -593,64 +593,71 @@ class AsyncAuth(object):
         if not acceptance_wait_time_max:
             acceptance_wait_time_max = acceptance_wait_time
         creds = None
+
         channel = salt.transport.client.AsyncReqChannel.factory(self.opts,
                                                                 crypt='clear',
                                                                 io_loop=self.io_loop)
-        error = None
-        while True:
-            try:
-                creds = yield self.sign_in(channel=channel)
-            except SaltClientError as exc:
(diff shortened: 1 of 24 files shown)


salt/crypt.py
```diff
@@ -593,64 +593,71 @@ def _authenticate(self):
...
+                        # and try the next one in the list.
+                        if self.opts.get('local_masters', None):
+                            error = SaltClientError('Minion failed to authenticate'
+                                                    ' with the master, has the '
+                                                    'minion key been accepted?')
+                            break
+                        else:
+                            print('Minion failed to authenticate with the master, '
+                                  'has the minion key been accepted?')
+                            sys.exit(2)
```
twangboy: Will this cause the minion process to stop trying to connect with the master?

salt/crypt.py
```diff
@@ -1208,31 +1221,34 @@ def authenticate(self, _=None):  # TODO: remove unused var
...
```
twangboy: Also here

terminalmage (approved):

salt/crypt.py
s0undt3ch: I'm not changing behavior. It just makes sure we close the channel when done.
I.e., whatever was the existing reconnection behavior, is still in place.

PR #51016, opened 2018-12-31
[develop] cmdmod OS X fix

garethgreenaway:
### What does this PR do?
Updating cmdmod when run on Darwin (OS X), removing the -l from the call to su so that the session does not immediately change into the user's home directory, when run with runas.

N/A

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/modules/cmdmod.py
+++ b/salt/modules/cmdmod.py
@@ -413,7 +413,7 @@ def _run(cmd,
         if isinstance(cmd, (list, tuple)):
             cmd = ' '.join(map(_cmd_quote, cmd))
 
-        cmd = 'su -l {0} -c "cd {1}; {2}"'.format(runas, cwd, cmd)
+        cmd = 'su {0} -c "{2}"'.format(runas, cmd)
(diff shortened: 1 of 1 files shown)


salt/modules/cmdmod.py
```diff
@@ -413,7 +413,7 @@ def _get_stripped(cmd):
         if isinstance(cmd, (list, tuple)):
             cmd = ' '.join(map(_cmd_quote, cmd))
 
-        cmd = 'su -l {0} -c "cd {1}; {2}"'.format(runas, cwd, cmd)
+        cmd = 'su {0} -c "{2}"'.format(runas, cmd)
```
s0undt3ch: There's no 2 index on the format args.

s0undt3ch (changes requested):

salt/modules/cmdmod.py
s0undt3ch: And the `cd` is no longer needed?
Perhaps only on OSX?

PR #51025, opened 2019-01-02
Update win_task.py

rmarcinik:
In the info() function the win32com client is used to get the result of a scheduled task, and the LastTaskResult is a code without a description. The results dict was created to give the codes some meaning, but it is static and is missing some possible return codes. I think most of the codes can be translated using the win32api like win32api.FormatMessage(task.LastTaskResult), so we wouldn't need to hardcode them.  I have made a comparison between what the win32api returns when given the keys of the results dict.
```
0x0: {
'oldvalue': 'The operation completed successfully',
'win32api': 'The operation completed successfully.'},
0x1:{
'oldvalue': 'Incorrect or unknown function called',
'win32api': 'Incorrect function.'},
0x2:{
'oldvalue': 'File not found',
'win32api': 'The system cannot find the file specified.'},
0xA:{
'oldvalue': 'The environment is incorrect',
'win32api': 'The environment is incorrect.'},
0x41300:{
'oldvalue': 'Task is ready to run at its next scheduled time',
'win32api': 'The task is ready to run at its next scheduled time.'},
0x41301:{
'oldvalue': 'Task is currently running',
'win32api': 'The task is currently running.'},
0x41302:{
'oldvalue': 'Task is disabled',
'win32api': 'The task will not run at the scheduled times because it has been disabled.'},
0x41303:{
'oldvalue': 'Task has not yet run',
'win32api': 'The task has not yet run.'},
0x41304:{
'oldvalue': 'There are no more runs scheduled for this task',
'win32api': 'There are no more runs scheduled for this task.'},
0x41306:{
'oldvalue': 'Task was terminated by the user',
'win32api': 'The last run of the task was terminated by the user.'},
0x8004130F: {
'oldvalue': 'Credentials became corrupted',
'win32api': 'No account information could be found in the Task Scheduler security database for the task indicated.'},
0x8004131F: {
'oldvalue': 'An instance of this task is already running',
'win32api': 'An instance of this task is already running.'},
0x800704DD: {
'oldvalue': 'The service is not available (Run only when logged in?)',
'win32api': 'The operation being requested was not performed because the user has not logged on to the network. The specified service does not exist.'},
```
Two keys were not unrecognized by the win32api however, so I am curious as to where they came from.
```
0xC000013A: {
'oldvalue': 'The application terminated as a result of CTRL+C',
'win32api': 'Unknown code'},
0xC06D007E: {
'oldvalue': 'Unknown software exception',
'win32api': 'Unknown code'}
```
What do you think, is this a worthy change to make?


garethgreenaway:
@rmarcinik Thanks for the PR.  Looks like you've got some lint warnings.

salt/modules/win_task.py
```diff
@@ -143,28 +143,31 @@
...
+          0x41302: 'Task is disabled',
+          0x41303: 'Task has not yet run',
+          0x41304: 'There are no more runs scheduled for this task',
+          0x41306: 'Task was terminated by the user',
+          0x8004130F: 'Credentials became corrupted',
+          0x8004131F: 'An instance of this task is already running',
+          0x800704DD: 'The service is not available (Run only when logged in?)',
+          0x800710E0: 'The operator or administrator has refused the request',
+          0xC000013A: 'The application terminated as a result of CTRL+C',
+          0xC06D007E: 'Unknown software exception'}
```
twangboy: This whole section should probably be indented 2 more spaces...

twangboy (changes requested):

salt/modules/win_task.py
rmarcinik: https://jenkinsci.saltstack.com/job/pr-doc/job/PR-51025/4/console said "docstring of salt.modules.win_task.show_win32api_code:3:Unexpected indentation." So I reformatted it and the tests passed, so I'm not sure whats correct.

twangboy: Maybe there was a mix of tabs and spaces...

twangboy (approved):

PR #51026, opened 2019-01-02
docs: add VIRTUALENV_ALWAYS_COPY option example

binary1230:
Show how a user can specify VIRTUALENV_ALWAYS_COPY to copy files instead of using symlinks

This is necessary for some filesystems (particularly vboxfs mounted in a linux VM from a windows host) where trying to create symlinks results in a 'protocol error'.

See https://github.com/pypa/pipenv/issues/2084 for discussion

In the future, it might be good to explicitly support the --always-copy switch on the virtualenv command so we don't have to use environment vars this way in Salt.

Open for other suggestions on how to document this, or how to handle this situation in Salt.

I didn't have a way to test the doc generation here, will make a fix if it's needed

### Previous Behavior

### New Behavior

### Tests written?

Yes/No

### Commits signed with GPG?

Yes/No

### DIFF
--- a/salt/modules/virtualenv_mod.py
+++ b/salt/modules/virtualenv_mod.py
@@ -134,6 +134,14 @@ def create(path,
     .. code-block:: bash
 
         salt '*' virtualenv.create /path/to/new/virtualenv
+     
+     Example of using --always-copy environment variable (in case your fs doesn't support symlinks).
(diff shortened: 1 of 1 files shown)


binary1230:
actually sorry first-time contributor, I will close and re-open this conforming better to the contrib guidelines

PR #51029, opened 2019-01-02
docs: add VIRTUALENV_ALWAYS_COPY option example

binary1230:
### What does this PR do?
Show how a user can specify VIRTUALENV_ALWAYS_COPY to copy files instead of using symlinks

This is necessary for some filesystems (particularly vboxfs mounted in a linux VM from a windows host) where trying to create symlinks isn't allowed and results in a 'protocol error'.

See https://github.com/pypa/pipenv/issues/2084 for non-salt-related discussion about the underlying issue

In the future, it might be good to explicitly support the --always-copy switch on the virtualenv command so we don't have to use environment vars this way in Salt.

Open for other suggestions on how to document this, or how to handle this situation in Salt.

I didn't have a way to test the doc generation here, will make a fix if it's needed

None

### Tests written?
N/a, documentation only

### Commits signed with GPG?
No

### DIFF
--- a/salt/modules/virtualenv_mod.py
+++ b/salt/modules/virtualenv_mod.py
@@ -134,6 +134,14 @@ def create(path,
     .. code-block:: bash
 
         salt '*' virtualenv.create /path/to/new/virtualenv
+     
+     Example of using --always-copy environment variable (in case your fs doesn't support symlinks).
+     This will copy files into the virtualenv instead of symlinking them.
+     
+     .. code-block:: bash
+        
+         - env:
+           - VIRTUALENV_ALWAYS_COPY: 1
     '''
     if venv_bin is None:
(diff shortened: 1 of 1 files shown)


dwoz:
@binary1230 Can you please resolve the linter failures?

dwoz:
@binary1230 ping

dwoz:
@binary1230 Can you find time to address the linter failures?

binary1230:
Hey, sorry I haven't had a chance to come back to this. I may have time in
next two weeks to give it a look.

dwoz:
@binary1230 pointed this to master

binary1230:
like a year later got around to hopefully fixing the linting errors, I'll ping when this is ready to go.

dwoz (approved):

PR #51031, opened 2019-01-02
Set long_description

### LINKED ISSUE #50964: Add a long description to setup.py for the new pypi.org

Since the upgrade of pypi.org, there is no project description for salt.

The `long_description` needs to be set on the sdist for this to be displayed on https://pypi.org/project/salt

See https://github.com/saltstack/pepper/blob/develop/setup.py#L17-L18 for an example

waynew:
Closes #50964 by setting the `long_description`.

It's good to have information on PyPI if people are looking there, and this does that.

### Previous Behavior
The [salt package on pypi](https://pypi.org/project/salt/) has no project description

screen shot 2019-01-02 at 4 07 21 pm

### New Behavior
Now uploading to pypi should also provide a description. It did on test pypi 😄 :
screen shot 2019-01-02 at 4 08 48 pm

I believe this should also update the description for the `salt-ssh` package.

### Tests written?

No(t relevant?)

### Commits signed with GPG?

Yes


twangboy (approved):

gtmanfred:
@garethgreenaway can you mark this to be backported? this should go into the older releases so that they update the project description the next time they are released.

thanks!

PR #51033, opened 2019-01-02
Add support for GCE accelerators in Salt Cloud

jp-harvey:
### What does this PR do?
This PR enables the Salt Cloud Google Compute Engine module to provision accelerator cards (GPUs) when creating an instance on GCE.
In addition there is a change to the `utils/vt.py` behavior under Python 3. It was found that Python 3 requires an unencoded string be passed to `stream_stdout.write` and Python 2 requires it to be encoded.
These should both be non-breaking changes.
N/A
### Previous Behavior
Previously it was not possible to pass the `ex_accelerator_type` and `ex_accelerator_count` parameters to libcloud to provision a VM instance on GCE with Salt Cloud, and it is not possible to add one after creating the instance.
### New Behavior
Libcloud supports accelerator cards in GCE as of v2.3.0, and now specifying the `ex_accelerator_count` and `ex_accelerator_type` parameters in the Salt Cloud profile will pass the parameters to libcloud and add the requested GPU(s) to the instance on create. 
Adding an accelerator card will also change the `ex_on_host_maintenance` setting to `TERMINATE` as that is required by GCP. An appropriate warning is displayed when provisioning the host.
### Tests written?
No

### Commits signed with GPG?
Yes


dwoz (approved):

terminalmage:
Ported to master in https://github.com/saltstack/salt/pull/56704

PR #51038, opened 2019-01-03
Update zabbix module to avoid exception

thechile:
As seen here https://www.zabbix.com/documentation/current/manual/api/reference/configuration/import 

  >Return values
  (boolean) Returns **true** if importing has been successful.

This PR updates the module to avoid `Exception: object of type 'bool' has no len()` error when using these type of Zabbix server API calls. 

### Tests written?
No

### Commits signed with GPG?
No


dwoz (commented):
@thechile Please have a look at the linter failure

dwoz:
@thechile can you find time to address the linter failures?

salt/modules/zabbix.py
```diff
@@ -2426,7 +2426,7 @@ def run_query(method, params, **kwargs):
             ret = _query(method, params, conn_args['url'], conn_args['auth'])
             if isinstance(ret['result'], bool):
                 return ret['result']
-            return ret['result'] if ret['result'] else False
+            return ret['result'] if (ret['result'] == True or len(ret['result']) > 0) else False
```
mchugh19: To be lint proper this should probably read something like:
`return ret['result'] if (bool(ret['result']) or len(ret['result']) > 0) else False`

dwoz:
I've ported this work to the master branch #55706

PR #51047, opened 2019-01-03
Add new execution module for troubleshooting Jinja map files

### LINKED ISSUE #49619: Add an execution module to test load_yaml/load_json
This would be a function that would import the same code used by our Jinja renderer to handle load_yaml/load_json, to be used as a sanity check so one can see what data that map file contains. This would be very helpful to troubleshoot instances in which the data that is loaded doesn't match what the user expects, resulting in SLS errors.

Additionally, it would be useful to also have an option to use grains.filter_by to see what a given platform would load out of that map file.

terminalmage:
Resolves #49619.

### DIFF
--- a/doc/ref/modules/all/index.rst
+++ b/doc/ref/modules/all/index.rst
@@ -198,6 +198,7 @@ execution modules
     jboss7
     jboss7_cli
     jenkinsmod
+    jinja
     jira_mod
     junos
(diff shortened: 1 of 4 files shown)


max-arnold:
@terminalmage Hi Erik! Do you plan to backport this feature to `master` branch?

PR #51063, opened 2019-01-04
RFC - Sunset Legacy Python Support

waynew:
Or, SaltStack is Python 3 only!

- Feature Name: Sunset Legacy Python 2 Support
- Start Date: 2019-01-04
- RFC PR: (leave this empty)
- Salt Issue: (leave this empty)

# Summary
[summary]: #summary

SaltStack is phasing out Python 2 support, with the develop branch of Salt Open
dropping Python 2 support effective immediately.

We will continue our policy of backporting bug fixes and security patches for
the last two releases, but starting with release Neon we will no longer
support legacy Python versions. For SaltStack, this means we will support
&gt;=Python 3.6.

Previous versions of SaltStack, that are able to run on legacy versions of
Python will continue to be available on http://repo.saltstack.com as well as
https://pypi.org.

Older platforms without native Python 3 packages (e.g. RedHat) will be provided SaltStack
packages that include Python 3. SaltStack will also fully support the creation
of new, Python 3 based packages for all supported platforms.

# Motivation
[motivation]: #motivation

## Python 2 EOL

Python 2 will no longer be [officially supported][1] as of [January 1,
2020][2]. As an organization, we cannot in good faith continue supporting
Python 2 - any lingering bugs in unsupported Python versions could cause
problems with SaltStack, and thus, our users.

### New Features!

By eliminating Python 2 support, it means we will be able to take advantage of
modern Python features, like async/await, async generators, and asyncio (or
other async frameworks).

Additionally, all improvements to the Python language (speed, additional
features, etc.) will come to Python 3. By dropping Python 2 support, SaltStack
users will benefit from these improvements simply by keeping their Python
installation up-to-date.

### Simplify

In addition to the performance benefits, the modern Python features should
allow us to simplify our codebase. As SaltStack internally supports Python 3
already, this simplification includes the ability to eliminate our dependency
on [six][3].

Additionally, but having only Python 3 to support, we significantly decrease
the time it takes to run our tests (having only one version of the language to
run tests against). It also reduces the mental overhead of having to consider
multiple Python versions when writing code.

### Python 3 Minimum Version

In order to take advantage of the new features in Python 3, SaltStack's
baseline version of Python will be Python 3.6.

### Availability

In the past, native Python 3 packages were difficult to find - one usually
needed to compile it themselves. That has changed now with all modern OSes
shipping Python 3.

## Alternatives
[alternatives]: #alternatives

For those who are stuck with Python 2:

- Previous versions of SaltStack, with Python 2 support, will continue to be
  available on http://repo.saltstack.com, as well as https://pypi.org
- Enterprise SaltStack customers will be able to make special arrangements to
  continue to receive Python 2 support for certain platforms.

If SaltStack continues to bridge Python 2 and 3 we will continue to incur the
significant costs associated with maintaining and testing the combined
codebase. We will also suffer the opportunity cost of not being able to take
advantage of the more modern features being introduced with new Python 3
versions.

# Drawbacks
[drawbacks]: #drawbacks

By dropping support for Python 2 it's possible that a significant number of
installations may be impacted. For those who have custom modules that are
Python 2 only, or are stuck on unsupported legacy patforms, they may not be
able to upgrade to the newer versions of Salt.

While this change won't break communication between master and minion, or
change pillars or state files in any way, when SaltStack begins taking
advantage of new Python language features (e.g.  async/await keywords,
f-strings), it will effectively be broken for those who for whatever reason
have not upgraded to modern platforms.

However, in some cases, this change may be tipping point that gives them the
impetus to upgrade to a more modern OS.

# Thank You, Python 2

It wouldn't be right to drop support Python 2 without first saying a hearty,
"Thank you!"

Python 2 was a wonderful language to program in, and has brought us all much
joy and success. Though newer versions of SaltStack will no longer run on
Python 2, we will all have fond memories of the improvements it made in our
life, and the improvements it allowed us to make in the lives of those around
us.

Python 2, so long, and thanks for all the fish!

[1]: https://www.python.org/dev/peps/pep-0373/#maintenance-releases
[2]: https://pythonclock.org/
[3]: https://pypi.org/project/six/

### DIFF
--- a/rfcs/0005-sunset-legacy-python-support.md
+++ b/rfcs/0005-sunset-legacy-python-support.md
@@ -1,0 +1,110 @@
+- Feature Name: Sunset Legacy Python 2 Support
+- Start Date: 2019-01-04
+- RFC PR: (leave this empty)
+- Salt Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+SaltStack is phasing out Python 2 support, with the develop branch of Salt Open
+dropping Python 2 support effective immediately.
+
+We will continue our policy of backporting bug fixes and security patches for
+the last two releases, but starting with release (???) we will no longer
+support legacy Python versions. For SaltStack, this means we will support
+&gt;=Python 3.6.
+
+Previous versions of SaltStack, that are able to run on legacy versions of
+Python will continue to be available on http://repo.saltstack.com as well as
+https://pypi.org.
+
+Older platforms without native Python 3 packages (e.g. RedHat) will be provided SaltStack
+packages that include Python 3.
+
+# Motivation
+[motivation]: #motivation
+
+## Python 2 EOL
+
+Python 2 will no longer be [officially supported][1] as of [January 1,
+2020][2]. As an organization, we cannot in good faith continue supporting
+Python 2 - any lingering bugs in unsupported Python versions could cause
+problems with SaltStack, and thus, our users.
+
+## Python 3
+
+### New Features!
+
+By eliminating Python 2 support, it means we will be able to take advantage of
+modern Python features, like async/await, async generators, and asyncio (or
+other async frameworks).
+
+Additionally, all improvements to the Python language (speed, additional
+features, etc.) will come to Python 3. By dropping Python 2 support, SaltStack
+users will benefit from these improvements simply by keeping their Python
+installation up-to-date.
+
+### Simplify
+
+In addition to the performance benefits, the modern Python features should
+allow us to simplify our codebase. As SaltStack internally supports Python 3
+already, this simplification includes the ability to eliminate our dependency
+on [six][3].
+
+Additionally, but having only Python 3 to support, we significantly decrease
+the time it takes to run our tests (having only one version of the language to
+run tests against). It also reduces the mental overhead of having to consider
+multiple Python versions when writing code.
+
+### Python 3 Minimum Version
+
+In order to take advantage of the new features in Python 3, SaltStack's
+baseline version of Python will be Python 3.6.
+
+### Availability
+
+In the past, native Python 3 packages were difficult to find - one usually
+needed to compile it themselves. That has changed now with all modern OSes
+shipping Python 3.
+
+## Alternatives
+[alternatives]: #alternatives
+
+For those who are stuck with Python 2:
+
+- Previous versions of SaltStack, with Python 2 support, will continue to be
+  available on http://repo.saltstack.com, as well as https://pypi.org
+- Enterprise SaltStack customers will be able to make special arrangements to
+  continue to receive Python 2 support for certain platforms.
+
+If SaltStack continues to bridge Python 2 and 3 we will continue to incur the
+significant costs associated with maintaining and testing the combined
+codebase. We will also suffer the opportunity cost of not being able to take
+advantage of the more modern features being introduced with new Python 3
+versions.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+This is a significant breaking change - many installations are on older
+platforms, and some may have an inclination to not upgrade their systems.
+Though this may be the tipping point that gives them the impetus to upgrade to
+a more modern OS.
+
+# Thank You, Python 2
+
+It wouldn't be right to drop support Python 2 without first saying a hearty,
+"Thank you!"
+
+Python 2 was a wonderful language to program in, and has brought us all much
+joy and success. Though newer versions of SaltStack will no longer run on
+Python 2, we will all have fond memories of the improvements it made in our
+life, and the improvements it allowed us to make in the lives of those around
+us.
+
+Python 2, so long, and thanks for all the fish!
+
+
+[1]: https://www.python.org/dev/peps/pep-0373/#maintenance-releases
+[2]: https://pythonclock.org/
+[3]: https://pypi.org/project/six/
--- a/rfcs/0005-sunset-legacy-python-support.md
+++ b/rfcs/0005-sunset-legacy-python-support.md
@@ -10,7 +10,7 @@ SaltStack is phasing out Python 2 support, with the develop branch of Salt Open
 dropping Python 2 support effective immediately.
 
 We will continue our policy of backporting bug fixes and security patches for
-the last two releases, but starting with release (???) we will no longer
+the last two releases, but starting with release Neon we will no longer
 support legacy Python versions. For SaltStack, this means we will support
 &gt;=Python 3.6.
 
@@ -19,7 +19,8 @@ Python will continue to be available on http://repo.saltstack.com as well as
 https://pypi.org.
 
 Older platforms without native Python 3 packages (e.g. RedHat) will be provided SaltStack
-packages that include Python 3.
+packages that include Python 3. SaltStack will also fully support the creation
+of new, Python 3 based packages for all supported platforms.
 
 # Motivation
 [motivation]: #motivation
@@ -86,10 +87,19 @@ versions.
 # Drawbacks
 [drawbacks]: #drawbacks
 
-This is a significant breaking change - many installations are on older
-platforms, and some may have an inclination to not upgrade their systems.
-Though this may be the tipping point that gives them the impetus to upgrade to
-a more modern OS.
+By dropping support for Python 2 it's possible that a significant number of
+installations may be impacted. For those who have custom modules that are
+Python 2 only, or are stuck on unsupported legacy patforms, they may not be
+able to upgrade to the newer versions of Salt.
+
+While this change won't break communication between master and minion, or
+change pillars or state files in any way, when SaltStack begins taking
+advantage of new Python language features (e.g.  async/await keywords,
+f-strings), it will effectively be broken for those who for whatever reason
+have not upgraded to modern platforms.
+
+However, in some cases, this change may be tipping point that gives them the
+impetus to upgrade to a more modern OS.
 
 # Thank You, Python 2
 


thatch45:
@waynew this is great!
Can we add that SaltStack will fully support the creation of new python3 based packages for all supported platforms?
Also, this is not really a breaking change. We have been pushing people to make modules python3 supported for some time, and the transport and communication between masters and minions will not be broken by this change.
This only means that in a future release (I would push for Neon) we would no longer need to support python2, so it is breaking only inasmuch as people have not updated custom, internally developed salt modules.

waynew:
@thatch45 how do those changes look?

s0undt3ch (approved):
WOW....

terminalmage:
Those are some pretty serious drawbacks...

waynew:
😅 clearly some misunderstandings on my part. We'll get that cleared up 👍

isbm:
> However, in some cases, this change may be tipping point that gives them the impetus to upgrade to a more modern OS.

I think migrating to Ansible, which [is just running](https://github.com/ansible/ansible/blob/devel/setup.py#L247) on 2.7 and 3.5, is way cheaper and less disastrous than touching e.g. production of SAP Hana or an Oracle cluster.

Good luck with that.

cro:
I disagree.  If that becomes an issue we can ship Salt and a Python 3.6 build.  Then customers can leave their system Python untouched.  We already do this for the enterprise packages for AIX and Solaris.

thatch45:
Right the main solution here is to ship with python 3 on systems that don't support python 3

thatch45:
I want to emphasize that Python 2 is EOL in less than 1 year. I cannon in good faith continue to support systems management software on an abandoned platform.
If you want to make a case for python 3.5 over 3.6 I think that should be reasonable, but Python2 needs to die.

isbm:
@s0undt3ch  SaltStack's supported 3.6 likely will render the rest of the vendor supported system as unsupported. That said, if an OS has Python 3.5 as vendor supported and that version is not EOL, there is no valid reason not to support it.

sjorge:
Is there a list of modules/states that currently do not work on py 3.x ?

s0undt3ch:
Here's my reasoning.

Maintaining a py2 and a py3 code base has taken its toll.
But we did it because we wanted to move to py3, and py2 EOL was still not around the corner.
Today, we're in a place where dropping py2 makes sense and that's what we're setting out to do.

Now, to "undo" all of the py2 and py3 logic and move to only py3, is again, going to take its toll, if we want to do it right.
We don't want to leave mixed code laying around, we're after a solid py3 codebase which might require rewriting as opposed to "patch" fixing parts of salt.
We're also after making Salt an even better product, specifically, more maintainable in the long run, and that's why, personally I'm shooting for py36. Honestly, I was shooting for Py37(remember, we're after a better codebase in the long run), but my argument was reasoned to py36, the same way you're trying to reason the Salt project to go only as high as py35.

The reasons to aim for py36 have been laid out in this RFC. There are language features which will make the Salt project better.
SaltStack is aware that some of the officially supported distributions don't yet ship py36, and that's why this RFC also mentions that for such cases, SaltStack will provide the appropriate packaging requirements.
Since a py3 only codebase(because we're migrating from a mixed codebase) is already a huge effort. Aiming to a lower py3 would require yet another big effort when 3.5 goes EOL and then when 3.6 goes EOL, etc... We're just consolidating the work required by skipping the 3.5 EOL "step". If there were no major benefits of aiming to py36, then we would simply aim for py35, but that's not the case. There really are benefits when running under py36, specially with the goals we're setting out to achieve.

thatch45:
@sjorge I think that is a very valid question. I also wish I had a better view of usage on a per-module basis, I fear that we have a lot of basically unused modules.

isbm:
@sjorge not working not. Broken — yes, `ps`, for example.

isbm:
@s0undt3ch why you have to move the entire codebase to the specific version, instead of just  stop caring about specific limitations of particular version when time comes? That way we don't have to have all these efforts.

thatch45:
This would not be for a release that would come out for at least 6 months, or a later release. So we would need to figure out what liabilities are left so we can move modules up.

s0undt3ch:
Sorry, not much time to expand right now but, py2 has limitations, py3 has benefits(not that py3.5 _has_ limitations, it just doesn't bring as many benefits as 3.6).

s0undt3ch:
Well let's agree then to disagree.
We're on different points and we won't converge(not that I'm going to stand by my point no matter what, but the arguments so far haven't made me shift my position).
My personal goal, and why *I* am pushing for 3.6 is to make a better product with a reasonable amount of work(in the long run).
Please note that I made the `I` bold.
Sure we can just stick with 3.5, your point of view, and would that make a better product, perhaps.
To me it would be a stop gap until we can finally enjoy the improvements py36 brings, at which time we'd have to go through Salt's core, again, to make use of the benefits py36 brings.

isbm:
@s0undt3ch if it doesn't run on vendor supported 3.5 — it is not better. There are plenty of places in Salt code that needs  serious  work  investment keeping us busy until 3.5 is phased out.

isbm:
@sjorge for the record, SUSE fixed many of that x509 stuff like couple of months ago. There are more broken things. :-)

thatch45:
The tornado code is very messy and the fix is also proving to be difficult in large part because of some core issues with how it was originally written. I would rather replace it with asyncio, and the async generators in python 3.6 would allow us to make it much better. This could be fixed easier, quicker, and solve many ongoing issues we have with tornado if we move to a better and more modern platform, asyncio.

waynew:
Hey @isbm, thanks for those comments! I'm super new, so be sure to take my opinions with a grain of salt 😉 

I 100% absolutely agree that there are people running *extreme* legacy systems in production (Python 1.4 or earlier, anyone?). Even on a less extreme side, obviously there are tons of people running Python 2.7 in production today, with zero plans to upgrade for the foreseeable future. RedHat has already committed to continue maintenance on Python 2.7 for their RHEL customers for several more years because there are a number of organizations that would prefer the risk of staying the same to the risk of change. And they're willing to pay for it, so why not?

AFAIK, we have similar plans - if an organization wants to trade money for keeping with Python 2.7, great! We're happy to offer support!


But considering everyone in the world is a bit of a strawman. I think we should ask the question, who does this change impact? It impacts people who want to have the latest and greatest Salt, but either do not want or cannot have the latest Python.

People on old versions of Salt will probably happily stay on older versions of Salt. People on archaic platforms that aren't interested in tools like Salt will probably stay there. It's really just the people who 1) want the latest Salt and 2) not the latest Python.

It also impacts the people who are spending their free (or paid) time developing on Salt.

From a developer perspective, I can say that Python 3.6+ is a *far* more enjoyable developer experience - more than 3.5 or 3.4, and definitely more than version `six`. I don't think I've met anyone (well, there was this one guy) who has a contrary opinion.

Assuming all those working on Salt prefer 3.6, then the question becomes - okay, how much of our user base:

1. Keeps up with the latest and greatest Salt
2. **and** has to live with an older Python?

Then again for those users, we also have to ask, is upgrading in the next 6 months impossible?

If they can upgrade to Python 3.6+ within 6 months, for me it's a no brainer - it's likely that they'll all be able to upgrade before it really matters.

Do you have a sense of how much our userbase wants modern Salt and has to live with legacy Python?

cachedout:
``` It impacts people who want to have the latest and greatest Salt, but either do not want or cannot have the latest Python.```

I think this is the crux of this issue and it requires a bit more unpacking. It's not clear from this RFC in which _release_ of Salt this is being planned for.

If Salt drops 2.x support in the next release (2019.2), in a few weeks, there will be two officially supported version of Salt -- one that's 3.x (what will then be the latest) and then the major release prior to that, which will be 2.x.

That's fine, but the issue comes six months _after_ that, when the _next_ 3.x-only release is put out. At that point, in order to get _any_ bug fix for Salt, you've got to move to Py3. If the discussion here is for 2019.2, that puts a drop-dead support for non-security bug fixes for Salt of this August. My guess is that there will be some pretty strong objections to this.

One question that comes to mind is whether or not SaltStack the company will continue to provide paid support *and bugfixes* for running 2.x with Salt after it's cycled out of the supported public releases. If there is literally going to be no recourse for 2.x users at that point, that certainly seems like it's worthy of further discussion.

gtmanfred:
Question, salt-ssh is an important part of salt.  What is the plan to make salt-ssh work on minions that do not have python3 installed on the ssh minion?  We can't just copy over the tarball and have everything work, we can use the new stuff that is coming in fluorine to install an old version of salt, but that doesn't really feel good when you are trying to use the new version of salt, but something like centos 7 doesn't come with python3 installed by default, and still has 5 more years of support, RHEL/CentOS 8 isn't even out of beta yet.

austinpapp:
a quick comment for now... because i do have lots of thoughts here.

expecting a large companies to have to upgrade the OSs of their entire fleet so that they can even _use_ the latest release of salt seems a bit crazy to me. Especially when it comes to bugfixes that may be security vulnerabilities, etc.  You'd end up isolating users and potentially forking salt in-house in an effort to continue to support py2. This also drastically limits salt's ability to compete with other automation tools as requirements to run are incredibly high.

waynew:
> I think this is the crux of this issue and it requires a bit more unpacking. It's not clear from this RFC in which _release_ of Salt this is being planned for.

Quoth the RFC:

> ... but starting with release **Neon** we will no longer
support legacy Python versions...

I guess it wasn't quite loud enough 🙂 

My personal expectation is that we're talking about a release maybe 6 months out? Definitely not the one in a couple of weeks, for sure. Practically speaking that means security patches for 2019.2 for at least what, 12 months?

> One question that comes to mind is whether or not SaltStack the company will continue to provide paid support _and bugfixes_ for running 2.x with Salt after it's cycled out of the supported public releases. If there is literally going to be no recourse for 2.x users at that point, that certainly seems like it's worthy of further discussion.

That's my understanding - that if people are willing to spend dollars to stay on Python 2 (or six, but realistically I'm guessing if you're not going for 3.6+ at this point then you're probably going to be on 2), instead of time/effort to upgrade to Python 3.6+, people are going to be available to take care of that for quite some time.

> expecting a large companies to have to upgrade the OSs of their entire fleet so that they can even use the latest release of salt seems a bit crazy to me.

I'm not sure if that's the expectation, I **think** the expectation is that we have users who are already upgraded (I know there's several old distros that have hit their EOL, unless they're paying for support).

---

I personally may be a more adventurous type, but in the past I've just built my own Python and distributed it (using Salt, of course 🙂). It wasn't difficult at all (and much *much* less difficult than moving to an alternate tool). Plus I had the added advantage that if upstream Python released a bugfix or something I didn't have to wait for the fix to make it to my PPA 😉 But I'm definitely on the earlier adopter side of things...

waynew:
FWIW I just learned that according to https://python3statement.org Tornado will be dropping support for Python2 on or before Jan 1, 2020. Given the fact that we're using Tornado, unless we want to be maintaining Tornado 2.7 for our end users, that seems like a(nother) pretty good reason to be dropping Python 2 support ourselves.

isbm:
I  want to strongly emphasise that one does `zypper install python2` on SLES and it doesn't say "No packages found" but installs it, that means "abandoned platform" claim just plain false. As long as the thing is on an enterprise OS and that OS distro is not EOLed, then it is supported and is **not** abandoned. This has to be very clearly understood. Same applies to RHEL and any other Enterprise OS. Take a look at Oracle Solaris. Where one can see there python 3.6 that is coming from Oracle?

I originally wrote this comment very long, but I decided not to. It just seems that here is quite a problem with understanding how enterprise works at all and RFC completely missed the LTS and LTSS from the vendors. It also misses the fact that major enterprise customers will get support from exactly the distribution vendor, i.e. SUSE/SLE and IBM/RHEL in this case. Simply because otherwise the whole support of such production system will be rendered void. Additionally, the versioning point that @cachedout just mentioned (thanks, Mike!) is also here. And the other pitfalls, like C-bindings for custom modules one can happily have with Salt, Salt SSH that @gtmanfred (thanks, Dan!) mentioned etc. There is zero plan for that so far. It seems we are simply talking different language and categories here, so I doubt that we are on the same page.

Now, SUSE _does not have a problem_ to deliver Py 3.6 on SLE. In fact, we do have 3.6.5 ready. It is just what sane admin will run production with TWO pythons (increasing risk factor in few orders of magnitude), and all just because to run that little Salt client on... e.g. SAP Hana cluster? It runs on an OS with a _decade_ of support, while applications on it are updated (!!!). It is not that like you have an OS and all the apps are also super-old. So if this RFC thinks it is a brilliant idea to drop support on that, then it is a very good plan how to propel them migrate to Ansible — after all, it is client-less and _does_ support 2.7 and 3.5 :laughing: 

That said, nor 2.7 is dead neither 3.5. Unless EOL had happened, the support of it should not be ever dropped. Period. I can understand de-prioritising 2.7 (i.e. not _dropping_ per se, but just getting ready to it), but the 3.5 still has some time till mid of 2020-09 + like I suggested, few months (say, two years at least). Thus I would come back to this RFC in a year to review it for 2.7 and in a year and half for 3.5. 

Until then, have a nice day. :wink:

dmurphy18 (approved):
Open-source wise Python 2.7 ends with 1 Jan 2020, but it is obvious that large customer are going to be using Python 2.7 for some significant period of time.  There are still customers and users using Python 2.6 which is long EOL'd.  Salt packages based on Python 2.7 are going to be around for some time with support, since RHEL 7 and Amazon Linux 2 are going to be around for another few years, however after Python 2.7 has EOL'd, there will be support available for Enterprise customers similar to the current support for Solaris and AIX.

With RHEL 6 there came a time where it was better to support Python 2.7 rather than continually back-porting to Python 2.6, the burden just became too great.  Similarly with RHEL 5 near it's EOL, SaltStack was essentially providing all necessary dependencies since some packages had been dropped, for example: python-jinja2.  Typically SaltStack has provided the packages needed to ensure that Salt functions on the platform and I can see this continuing, but for Enterprise customers once Python 2.7 has EOL'd.

Using Python 3.6 as the basis moving forward makes sense since this allows for asynchronous functionality which is the direction that Python 3 is moving forward with: got to start with the correct initial stake in the ground.  Noting that this Python 3.6  support is scheduled for Neon, which is still some time off, probably Fall of 2019 which implies 2019.2 (code name Fluorine) will still be supported for some time after Neon's release.

I would expect that by the Fall of 2019, that all major Linux distributions will have Python 3.6 support available in their current distributions, currently Python 3.6 is available for RHEL 7 (EPEL based), Ubuntu 18.04 and Debian 8, Amazon Linux 2 has Python 3.7 and upcoming RHEL 8 is Python 3.6, unsure if Debian 10 (Buster) will be Python 3.6 or 3.7 given it's release is not till probably mid-summer 2019.

I understand @isbm  concerns about Python 3.5 but Python 3.6 is a better starting point for leveraging asynchronous use in Python 3, however by Fall 2019 I would expect that SUSE is shipping Python 3.6 too.

gtmanfred:
It is worth mentioning that salt as it is right now doesn't support tornado 5.0 on python3, yet you are still supporting python3 for salt.  Are you maintaining tornado somewhere else so that you can have the latest tornado for python3? From what i remember, you don't, so i don't see the problem with pinning on older releases, to at least maintain compatibility for a little longer.

And again, nothing has been said for what you are going to do about salt-ssh, the whole point of which is not to have to install anything on the remote system.  And you can't exactly go the same route as windows, and ship a whole python distribution because then you will have to pack and ship a different python distribution for each linux version.  Or you can statically compile it, and have to maintain security updates for that.

gtmanfred:
> currently Python 3.6 is available for RHEL 7 (EPEL based), 

Have you considered customers that won't install epel?  Are they just not going to be able to use salt anymore?

gtmanfred:
Also, when deprecating things, does salt not require 2 releases of warnings anymore?

https://docs.saltstack.com/en/latest/topics/development/deprecations.html#deprecating-code

> Salt should remain backwards compatible, though sometimes, this backwards compatibility needs to be broken because a specific feature and/or solution is no longer necessary or required.

> A deprecation warning should be in place for at least two major releases before the deprecated code and its accompanying deprecation warning are removed.

isbm:
> I understand concerns about Python 3.5 but Python 3.6 is a better starting point for leveraging asynchronous use in Python 3

How did you made till today with asynchronous _without_ Python 3.6?

> however by Fall 2019 I would expect that SUSE is shipping Python 3.6 too.

If I were you, I would first [go to OBS](https://build.opensuse.org) and double-check if I my comments are indeed constructive. After finding out that SUSE shipping it for years in [SLE](https://build.opensuse.org/package/show/SUSE:SLE-15:Update/python3) and Leaps [here](https://build.opensuse.org/package/show/openSUSE:Leap:15.0:Update/python3) or [here](https://build.opensuse.org/package/show/openSUSE:Leap:15.1/python3), my next question to myself would be: if this is the case, what is the angle then? Thus if I would not understand something, I would first ask...

> I would expect that by the Fall of 2019, that all major Linux distributions will have Python 3.6 support available in their current distributions

And so you think admins will immediately rush upgrading their expensive clusters to a new systems, just because they will need to run Salt client. :+1: OK, very exciting already, but then so what you think they will have to do with:

- SaltSSH that _will break_?
- C/C++ bindings and dependencies that _will SEGFAULT_?
- Python-written states that _stop work_?
- Custom modules that _won't compute_?
- EOL roadmap to major distros that _will not_ be upgraded just for you?

RFC just doesn't care about this at all.

P.S. Why only @saltstack/team-core is reviewing this, if @saltstack/team-core  wrote this RFC?

damon-atkins:
Linux vendors will fix python2 bugs even if python.org is no longer supporting from 2020. As RHEL and others provide a commitment to provide support. https://access.redhat.com/support/policy/updates/errata

I assume RHEL still do minimal patching of open source. i.e. If a customer reports a bug, they apply only that fix and on-going security fixes.   Open Source Software-1.1 and RHEL Software-1.1-5 will be different, because RHEL version will have security fixes and bug fixes that customers have asked for.

For example salt may increase the required version of zmq from 5.1 to 5.6 because 5.6 has a security bug.  My understanding is RHEL will release 5.1-N with just the security fix from 5.6. Therefore salt should still only require 5.1 on RHEL & CentOS and 5.6 on Fedora. Its been a while since I looked at this.

Maybe SUSE does the same thing.

The alternative is salt to provide python 3 & modules for all platforms and add to the module path to search the vendor provided modules second. e.g    `/opt/salt/lib/python3.6:/usr/lib/python3.6`

damon-atkins:
> This could be fixed easier, quicker, and solve many ongoing issues we have with tornado if we move to a better and more modern platform, asyncio.
https://www.aeracode.org/2018/02/19/python-async-simplified/  Seems to have pros and cons

isbm:
> The alternative is salt to provide python 3 & modules for all platforms and 

...break vendor support. :+1: 

If you've bought an OS subscription a year ago and deployed your production cluster for another 5 years to get supported and maintained, you won't be super-happy hearing from 3rd party "Hey, upgrade your OS to the latest-greatest otherwise our Salt client you're running won't be talking to our new master anymore!".

s0undt3ch:
> Hey, upgrade your OS to the latest-greatest otherwise our Salt client you're running won't be talking to our new master anymore

This RFC never mentions that the backwards compatibility would be broken between different versions of salt master and minion...

isbm:
> This RFC never mentions that the backwards compatibility would be broken between different versions of salt master and minion...

Fine. But it doesn't mention the otherwise as well, so it is naturally to worry about. There should be a clear table what to expect, what combinations will work, timeline, versions and so on. Then we can have a clearer picture what to expect.

Can we have it, please?

s0undt3ch:
You can expect what SaltStack has always delivered.
Backwards compatibility is **never** broken unless **explicitly** necessary by, say, a security fix.

isbm:
So can we please have a detailed plan? I would prefer to see a table, something like (my incorrect proposal):

### Python version compatibility

| Version  | 2.7 | 3.5 | 3.6 | 3.7 |
|---|---|---|---|---|
| 2019.2  | 2020.01  | 2021.09  | 2023.12  | - |
| Neon  | 2020.01  | 2021.09  | 2023.12  | - |
| Sodium  | -  | 2021.09  | 2023.12  | - |
| Magnesium  | -  | -  | 2023.12  | - |

### Components compatibility

| Version | SaltSSH (Target) | Tornado 4 | Tornado 5 | Tornado 6 | asyncio |
|---|---|---|---|---|---|
| 2019.2 |2.7, 3.3, 3.4, 3.5, 3.6, 3.7 | yes | no | no | no
| Neon | 2.7, 3.5, 3.6, 3.7 | no | yes  | no | no
| Sodium | 3.5, 3.6, 3.7 | no | no  | yes | no
| Magnesium | 3.5, 3.6, 3.7 | no | no  | TBD | TBD
| Aluminium | 3.6, 3.7, 3.8 | no | no  | TBD | TBD

And so on. Then this would be some sort of plan, visible and the Universe can get ready for it, nobody will worry that much and no kittens harmed. For me personally that would look like a plan, so we can draw roadmaps, put warnings, prepare users, migrate things, no stress.

Would that work?

thatch45:
Given the responses to this RFC I am going to close it out. We will put together an RFC with deprecation timetables for Python 2.
We can also go over how the transition will be handled by SaltStack and present a more solid case for why we might want 3.6 over 3.5 or 3.4 etc.

Issue #51070, opened 2019-01-07
The network connection is large, causing the cSP to run full when the salt detection is alive.


menglong81:
Reading the /proc/net/tcp file for a machine with a large network connection will consume CPU. So we changed to use the ss command instead of the netstat command to get the network connection first. The ss uses the tcp_diag in the TCP protocol stack so it will be faster than Netstat command is much faster

WalkerWang731:
In fact currently `socket.py` module does not support the NETLINK_INET_DIAG protocol, so it can only be replaced with the ss tool. When the `socket.py` module supports more NETLINK protocols, you can using `socket.py` module to implement

Ch3LL:
just to clarify are you suggesting we migrate from netstat to ss in the socket.py module?

WalkerWang731:
I mean, python's official `socket `library not support NETLINK_INET_DIAG. so can be use ss first. if salt use ss to replace netstat, maybe is nice try. however old release of linux maybe unsupport, so ss before netstat

menglong81:
master_alive_interval   options
Salt will call the netstat command when it performs health check. However, if the number of system connections is large, it will consume cpu because it reads the file  /proc/net/tcp 
So we modified the code and hope to adopt it

Ch3LL:
thanks for clarifying and thanks for teh PR!

PR #51073, opened 2019-01-07
ss command replace netstat command

menglong81:
Reading the /proc/net/tcp file for a machine with a large network connection will consume CPU. So we changed to use the ss command instead of the netstat command to get the network connection first. The ss uses the tcp_diag in the TCP protocol stack so it will be faster than Netstat command is much faster

In fact currently socket.py module does not support the NETLINK_INET_DIAG protocol, so it can only be replaced with the ss tool. When the socket.py module supports more NETLINK protocols, you can using socket.py module to implement

#51070

### DIFF
--- a/salt/utils/network.py
+++ b/salt/utils/network.py
@@ -1399,9 +1399,11 @@ def _remotes_on(port, which_end):
     Return a set of ip addrs active tcp connections
     '''
     port = int(port)
+
     ret = _netlink_tool_remote_on(port, which_end)
-    if ret is not None and len(ret) > 0:    # ss tools may not be valid
+    if ret is not None:
         return ret
+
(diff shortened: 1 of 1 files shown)


menglong81:
ss command replace netstat

dwoz (commented):
@menglong81 I am fine with this but we have some issues that need to be addressed:

- This change should have tests to go along with it
- The linter build is failing and those issues need to be addressed
- The doc build is also failing.
- There are some merge conflicts

pullit:
@menglong81 Are you working on changes in a different branch and planning to open a new PR?

menglong81:
https://github.com/saltstack/salt/pull/51187

PR #51076, opened 2019-01-07
Typos in index_template_present documentation

OrangeDog:
### Previous Behavior

### New Behavior

### Tests written?

Yes/No

### Commits signed with GPG?

Yes/No

### DIFF
--- a/salt/states/elasticsearch.py
+++ b/salt/states/elasticsearch.py
@@ -234,10 +234,10 @@ def index_template_absent(name):
 
 def index_template_present(name, definition, check_definition=False):
     '''
-    Ensure that the named index templat eis present.
+    Ensure that the named index template is present.
 
     name
-        Name of the index to add
+        Name of the index to check
     definition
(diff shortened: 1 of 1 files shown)


salt/states/elasticsearch.py
```diff
@@ -234,10 +234,10 @@ def index_template_absent(name):
 
 def index_template_present(name, definition, check_definition=False):
     '''
-    Ensure that the named index templat eis present.
+    Ensure that the named index template is present.
 
     name
-        Name of the index to add
```
gtmanfred: This is incorrect, it should be the index to add, not check, because if it doesn't exist, it gets added.

gtmanfred (changes requested):

salt/states/elasticsearch.py
OrangeDog: Equally, if it does exist it might get changed instead of added.
I'll leave it as it was if you prefer.

gtmanfred (approved):

dwoz (approved):

PR #51082, opened 2019-01-07
Clarify python requirements for running unit tests

stevewritescode:
### What does this PR do?
I believe that the written recommendation for installing dependencies is out of date. I updated to match the steps I was able to use for installing dependencies and running `./setup.py test`

### Tests written?
No

### Commits signed with GPG?
No

### DIFF
--- a/doc/topics/tutorials/writing_tests.rst
+++ b/doc/topics/tutorials/writing_tests.rst
@@ -34,13 +34,11 @@ Getting Set Up For Tests
 
 There are a couple of requirements, in addition to Salt's requirements, that need
 to be installed in order to run Salt's test suite. You can install these additional
-requirements using the files located in the ``salt/requirements`` directory,
-depending on your relevant version of Python:
+requirements using the files located in the ``salt/requirements`` directory:
 
 .. code-block:: bash
 
-    pip install -r requirements/dev_python27.txt
-    pip install -r requirements/dev_python34.txt
+    pip install -r requirements/tests.txt
 
 To be able to run integration tests which utilizes ZeroMQ transport, you also
 need to install additional requirements for it. Make sure you have installed


doc/topics/tutorials/writing_tests.rst
```diff
@@ -34,13 +34,11 @@ Getting Set Up For Tests
...
 to be installed in order to run Salt's test suite. You can install these additional
-requirements using the files located in the ``salt/requirements`` directory,
-depending on your relevant version of Python:
+requirements using the files located in the ``salt/requirements`` directory:
 
 .. code-block:: bash
 
-    pip install -r requirements/dev_python27.txt
-    pip install -r requirements/dev_python34.txt
+    pip install -r requirements/tests.txt
```
gtmanfred: this should not be changed until salt has changed its test suite over to run with pytest.

as of right now, the way the test dependencies should be installed for runtests.py is to use the [salt-jenkins](https://github.com/saltstack/salt-jenkins) repository.

You can also use the kitchen-salt setup to bootstrap docker containers to run tests in the same manner that jenkins does.

stevewritescode: thanks! should the docs point directly to https://github.com/saltstack/salt-jenkins, then?

gtmanfred: The docs should really have instructions about using test-kitchen to run the test suite.

There are some examples here https://kitchen.saltstack.com/docs/file/docs/gettingstarted.md

but nothing specific to the salt test suite, which uses just `bundle exec kitchen verify` and can be configured with the runtests verifier to only run specific tests.

stevewritescode: Ok thanks for the feedback. I'll close this PR and open an issue re: adding documentation on using test-kitchen. I need to play around with it a bit more before I can write docs on that.

stevewritescode:
https://github.com/saltstack/salt/issues/51136

PR #51094, opened 2019-01-08
cmdmod: add sysfs into the chroot

aplanas:
### What does this PR do?

Some commands executed inside the chroot require access to the /sys
information. For example, /sys/firmware is used to get to information
from UEFI systems.

This patch mount a sysfs file system inside the chroot.

Also normalize the name for devtmpfs, replacing the old udev one.

### Tests written?

No, current ones should cover the case

### DIFF
--- a/salt/modules/cmdmod.py
+++ b/salt/modules/cmdmod.py
@@ -3231,12 +3231,16 @@ def run_chroot(root,
     '''
     __salt__['mount.mount'](
         os.path.join(root, 'dev'),
-        'udev',
+        'devtmpfs',
         fstype='devtmpfs')
     __salt__['mount.mount'](
         os.path.join(root, 'proc'),
         'proc',
         fstype='proc')
+    __salt__['mount.mount'](
+        os.path.join(root, 'sys'),
+        'sysfs',
+        fstype='sysfs')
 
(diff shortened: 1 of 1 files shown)


salt/modules/cmdmod.py
```diff
@@ -3231,12 +3231,16 @@ def run_chroot(root,
     '''
     __salt__['mount.mount'](
         os.path.join(root, 'dev'),
-        'udev',
```
thatch45: Is this change going to create a compatibility issue with some systems?

aplanas: AFAIK this can be any random text as there is not real device named 'udev', nor 'devtmpfs'. Is only the type what is important.

Distributions normalize it to the same name of the type, so a `mount` will show lines like

```
devtmpfs on /dev type devtmpfs ...
tmpfs on /run type tmpfs ...
...
```

So always following `"type" on "mount point" type "type"`, duplicating the name without problems.

thatch45: Thank you for clarifying so I did not have to dig into it more :)

thatch45 (approved):

aplanas:
* Rebased

PR #51108, opened 2019-01-09
[develop] Fix inconsistency with `pkg.list_pkgs` when using `attr` on RHEL systems

meaksh:
### What does this PR do?
As discussed, this PR brings https://github.com/saltstack/salt/pull/50563 to `develop` branch.

Now, `pkg_resource.format_pkg_list` does not directly parse the "arch" from the package name but delegates it to an optional `parse_arch_from_name` particular implementation on each virtual "pkg" module.


thatch45:
At first glance it looks good to me, but @terminalmage should be the one to sign off on it.
I updated the branch, hopefully we can get a cleaner test run

salt/modules/pkg_resource.py
```diff
@@ -311,21 +311,30 @@ def format_pkg_list(packages, versions_as_list, attr):
...
     if attr:
+        ret_attr = {}
         requested_attr = {'epoch', 'version', 'release', 'arch', 'install_date', 'install_date_time_t'}
 
         if attr != 'all':
-            requested_attr &= set(attr + ['version'])
+            requested_attr &= set(attr + ['version'] + ['arch'])
 
         for name in ret:
+            _parse_arch_from_name = __salt__.get('pkg.parse_arch_from_name', lambda pkgname: {'name': pkgname, 'arch': None})
```
terminalmage: Let's not try to get too cute, here. What about something like this?

```python
if 'pkg.parse_arch_from_name' in __salt__:
    _parse_arch_from_name = __salt__['pkg.parse_arch_from_name'](pkgname)
else:
    _parse_arch_from_name = {'name': pkgname, 'arch': None}
```

Also, I have a _strong_ dislike of the function name `parse_arch_from_name`. We should think of something better.

terminalmage (changes requested):
This doesn't even come close to satisfying the requirements outlined in https://github.com/saltstack/salt/pull/50563#issuecomment-442567698

salt/modules/pkg_resource.py
dwoz: I agree with @terminalmage here on both points. We should not be using lambdas for something like this. As for the method name, a simpler `parse_arch` seems fine to me.

meaksh: Hey @dwoz , your feedback is now addressed. No more lambda and function has been renamed to `parse_arch`.

meaksh:
Closing in favor of https://github.com/saltstack/salt/pull/55580

PR #51115, opened 2019-01-09
[2017.7] Introduce pre-commit

s0undt3ch:
### What does this PR do?
Introduce pre-commit into the workflow.
Right now, it has some bare minimum checks because once pre-commit is enabled for a repo, it will complain about missing configuration.
To solve that, we add pre-commit to all active branches(again, a minimal setup).
Additional pre-commit checks will land most likely only on develop and some documentation about it still needs to be written.

### Tests written?

No

### Commits signed with GPG?

Yes


terminalmage:
What happened to the line endings in so many of these files?

s0undt3ch:
The pre-commit fixes were applied.
* No mixed line endings.
* A new line at the end of file
* Strip trailing white-space

s0undt3ch:
Once pre-commit is installed locally(one time setup), the above checks/fixes will be enforced for all the changed files being committed.

garethgreenaway (approved):

terminalmage:
OK thanks... that wasn't clear from the commit message, and I was also (incorrectly) thinking that line-ending/noeol changes would get some other visual distinction in the diff, like you see when removing trailing whitespace.

Ch3LL:
@s0undt3ch does this still need to be merged here since 2017.7.9 will be the last bug fix release?

s0undt3ch:
Nope.

Issue #33309, opened 2016-05-17
RFC: Change Salt to create JIDs as UTC time


whiteinge:
## Background

JIDs in Salt are currently generated from the Salt Master's current timezone. This makes it difficult or (practically) impossible for consumers of Salt to reconcile a job timestamp with the user's relative local time -- specifically programmatic consumers like via returners or salt-api.

(Please comment with corrections if I am mistaken with anything above!)
## Request for comments

What would the potential fallout be for switching Salt to store JIDs in UTC in the next major Salt release?
### Initial thoughts

There would be an increased risk for JID collisions for a window of time after users upgrade to this version (I think just the offset of the user's current time and UTC). That risk is already pretty low, but something to consider anyway.
### Alternate ideas
- Leave the JID as-is and add a new field to the job cache with a UTC timestamp.
  
  Safer but less "single source of truth"-y.
- Expose the Salt Master's timezone via a Runner or a salt-api endpoint and let the end-users do the math.
  
  Safer for Salt but a PITA for users. Also likely fraught with edge-cases and mistakes.

@saltstack/team-core @saltstack/team-riot @saltstack/team-platform @thatch45 thoughts?

cro:
:+1: from me

cro:
Huge benefits for external job caches as well. Also, we already have issues with companies who have masters in different timezones.  If they aggregate to a single external job cache then you can't meaningfully compare JIDs to see when jobs that hit different masters occurred.

cachedout:
I'd want to be sold on the benefit a little more. There are maybe a few billion stored Salt jobs out there in the world. The risk of collisions is higher than I'd like to imagine. ;]

thatch45:
salt checks for a collision before making the jid though....

cro:
Aren't the collisions just going to exist for less than 24 hours after the switch to UTC is made, though?  And only timezones east of UTC are affected since everyone west of UTC will see their JIDs jump forward.

cachedout:
Both good points.

If we have a retry mechanism (which I vaguely recall but haven't looked into) then I'm cool.

thatch45:
right, so it is 12 hours of collisions, which we SHOULD have a check in place for, but we should double check it, since I wrote it 5 years ago....

whiteinge:
I remember that Tom. His voice was a tad more gravelly but he had way, way fewer gray hairs.

whiteinge:
[...]

PR #51126, opened 2019-01-09
Add `utc_jid` option for generating jids based on utc

msteed:
### What does this PR do?

Add the `utc_jid` config option (boolean) for generating jids based on UTC instead of local time.

This is useful in SSE deployments where masters are in different timezones, so that jobs across the infrastructure can easily be sorted chronologically. The workaround is to set each master's clock to UTC.

#33309

### Previous Behavior

Previous behavior is still available and enabled by default.

### New Behavior

If `utc_jid` is `True`, generate jids based on UTC instead of local time.

### Tests written?

Yes

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/config/__init__.py
+++ b/salt/config/__init__.py
@@ -459,6 +459,9 @@ VALID_OPTS = {
     # Tell the client to display the jid when a job is published
     'show_jid': bool,
 
+    # Generate jids based on UTC time instead of local time
+    'utc_jid': bool,
+
     # Ensure that a generated jid is always unique. If this is set, the jid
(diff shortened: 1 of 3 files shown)


garethgreenaway:
Would it make sense to add a new test for this rather than adding code into the existing test?  @saltstack/team-core thoughts?

garethgreenaway (approved):

msteed:
@garethgreenaway : Do I need to do something about the codeclimate issue?

garethgreenaway:
@msteed That codeclimate issue isn't a barrier to the PR being merged.

Issue #29001, opened 2015-11-18
debconf.set doesn't support "prereq" in states


olfway:
```salt
zabbix-server_debconf:
  debconf.set:
    - name: {{ zabbix.server.pkg}}
    - data:
        'zabbix-server-mysql/internal/skip-preseed': {'type': 'boolean', 'value': True}
        'zabbix-server-mysql/dbconfig-install': {'type': 'boolean', 'value': False}
        'zabbix-server-mysql/dbconfig-upgrade': {'type': 'boolean', 'value': False}
    - prereq:
      - pkg: zabbix-server
```
prints warning:
```yaml
----------
          ID: zabbix-server_debconf
    Function: debconf.set
        Name: zabbix-server-mysql
      Result: None
     Comment:
     Started: 14:31:19.554909
    Duration: 291.696 ms
     Changes:
              ----------
              zabbix-server-mysql/dbconfig-install:
                  New value: false
              zabbix-server-mysql/dbconfig-upgrade:
                  New value: false
              zabbix-server-mysql/internal/skip-preseed:
                  New value: true
    Warnings: '__prerequired__' is an invalid keyword argument for
              'debconf.set'. If you were trying to pass additional data to be
              used in a template context, please populate 'context' with 'key:
              value' pairs. Your approach will work until Salt Carbon is out.
              Please update your state files.
```
I'm using salt 2015.8.1 (Beryllium)

jfindlay:
@olfway, thanks for the report.  Can you also post the output of the debug logs of the master and minion when this happens?

olfway:
I'm not sure how to get relevant debug output ("salt-{minion,master} -l debug" gives too much)

but this issue looks same as this one https://github.com/saltstack/salt/issues/16709

and same patch fixes issue for me (no warnings output, order looks right):
```patch
--- salt/states/debconfmod.py.orig	2015-11-18 21:48:48.174990524 +0300
+++ salt/states/debconfmod.py	2015-11-18 21:49:03.715301886 +0300
@@ -132,7 +132,7 @@
     return ret


-def set(name, data):
+def set(name, data, **kwargs):
     '''
     Set debconf selections
```

jfindlay:
ping @cachedout

jfindlay:
@olfway, you are welcome to submit a pull request against 2015.8 with this change.

abednarik:
PR in place #29463 

Please let me know if there are anything else I can do.

Cheers.

jfindlay:
[...]

PR #51134, opened 2019-01-10
Add **kwargs to file.rename

edusperoni:
### What does this PR do?
This will allow using it with prereq

https://github.com/saltstack/salt/issues/29001

### Previous Behavior
file.rename with prereq throws a warning

### New Behavior
file.rename works with prereq without warning

### Tests written?

No

### Commits signed with GPG?

No


salt/states/file.py
```diff
@@ -6810,7 +6810,7 @@ def copy_(name,
     return ret
 
 
-def rename(name, source, force=False, makedirs=False):
+def rename(name, source, force=False, makedirs=False, **kwargs):
```
dwoz: I do not see `kwargs` being used in this method and the issue associated with this PR is closed. Is this PR obsolete? If not, can you please explain why `kwargs` is needed here?

edusperoni: When used with prereq, the same warning from the linked issue is shown, so I applied the same fix. Many other methods add kwargs in the same way, it seems

dwoz (approved):

Issue #45315, opened 2018-01-07
Grains file does not honor --config-dir


lorengordon:
### Description of Issue/Question

When using the `--config-dir` option, grains are not read from or written to `{config-dir}/grains`.

### Setup

Create a config dir:

```
mkdir C:\Foo
```

Also create a grains file in the config-dir with the `bar` grain:

```
PS C:\Users\Administrator> cat C:\Foo\grains
bar: bar
```

### Steps to Reproduce Issue

```
PS C:\Users\Administrator> C:\salt\salt-call.bat --local --config-dir C:\Foo grains.setval foo foo
local:
    ----------
    foo:
        foo
PS C:\Users\Administrator> cat C:\salt\conf\grains
foo: foo
PS C:\Users\Administrator> C:\salt\salt-call.bat --local --config-dir C:\Foo grains.get foo
local:
    foo
PS C:\Users\Administrator> C:\salt\salt-call.bat --local --config-dir C:\Foo grains.get bar
local:
```

Note how the grain `foo` is written to the default config-dir, `C:\salt\conf\grains`, rather than the custom config-dir location, `C:\Foo\grains`.

Also note that the `bar` grain is not read.

### Versions Report

This is Windows, obviously, but I saw the same behavior on Linux (Centos6 and 7).

```
PS C:\Users\Administrator> C:\salt\salt-call.bat --versions
Salt Version:
           Salt: 2016.11.6

Dependency Versions:
           cffi: 1.10.0
       cherrypy: 7.1.0
       dateutil: 2.5.3
      docker-py: Not Installed
          gitdb: 0.6.4
      gitpython: 2.0.8
          ioflo: 1.5.5
         Jinja2: 2.9.4
        libgit2: Not Installed
        libnacl: 1.4.5
       M2Crypto: Not Installed
           Mako: 1.0.4
   msgpack-pure: Not Installed
 msgpack-python: 0.4.8
   mysql-python: Not Installed
      pycparser: 2.17
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: Not Installed
         Python: 2.7.12 (v2.7.12:d33e0cf91556, Jun 27 2016, 15:24:40) [MSC v.1500 64 bit (AMD64)]
   python-gnupg: 0.3.8
         PyYAML: 3.11
          PyZMQ: 16.0.1
           RAET: Not Installed
          smmap: 0.9.0
        timelib: 0.2.4
        Tornado: 4.4.1
[...]

PR #51142, opened 2019-01-10
Warn if config dir is missing config file

waynew:
### What does this PR do?

Adds a warning if the config dir folder does not exist.

Addresses #45315

### Previous Behavior

Currently we just implicitly replace the `conf_file` with the default (`/etc/salt/minion`) if there's no `minion` config file found. Because there are a lot of things that reference the `conf_file` path, like grains, rather than ensuring that the behavior stayed the same, my thought here is that we just display a warning. The real issue for the OP in that issue wasn't the behavior so much as that it was undocumented, so when they ran the code they got unexpected behavior (the file was set to the default without warning).

### New Behavior

Displays a warning on stderr if the config dir is missing, complete with the path to the default config file.

### Tests written?

No(t yet?) - I wanted to make sure that I was headed down a reasonable path (that, and this function in particular isn't unit tested, and though I haven't confirmed it yet, I'm guessing that it's covered by some test or another already)

### Commits signed with GPG?

Yes


dwoz (changes requested):
@waynew looks like there are some linter errors, can you look into them?

waynew:
@dwoz whoops. Looks like there was an extra blank line 😝looks like the lint step passes now 👍

dwoz (approved):

PR #51150, opened 2019-01-11
Back-port #49508 to 2018.3

Ch3LL:
Back-port #49508


dwoz (approved):

isbm:
@meaksh we need to throw away our patch in this case and rebase.

Issue #4010, opened 2013-03-08
Need to provide way to signal dpkg-reconfigure when debconf preseed changes


eugenea:
Problem: if package is already installed and debconf selection for it got changed via debconf.set - the selections will not affect current package state.

Need to provide mechanism to detect changes in debconf selections and signal
dpkg-reconfigure for particular package to enable changes.

grinapo:
So far so good... 😞

grinapo:
A HighSeverity bug autoclosed, mmh. Stoopid bot! Sit! STAY!

waynew:
I know that the `debconf.set` state fires changes - is there any reason that this won't work for you?

```
somestate:
  debconf.set:
    - name: whatever-thing
    - data:
              'whatever-thing': {'type': 'boolean', 'value': True}

reconfigurething:
  cmd.run:
    - name: dpkg-reconfigure -f noninteractive <your thing>
    - onchanges:
      - debconf: somestate
```

boltronics:
@waynew I haven't tested it, but that looks good to me. Thanks for the suggestion! I used a different approach by making sure debconf.set was executed before installing the package, but your suggestion is nicer in some ways.

It would be ideal if Salt provided a more direct and intuitive way to execute that command on changes when the package is already installed. eg.
```
somestate:
  debconf.set:
    - name: whatever-thing
    - data:
        'whatever-thing':
          'type': 'boolean'
          'value': 'true'
    - reconfigure: True
```

waynew:
@boltronics would it help if we added some information to the docs for debconf? To me it seems like that'd be the most helpful thing here.

boltronics:
I agree, the docs should be updated regardless of any change being implemented.

waynew:
Anyone have a reason this shouldn't be closed when that PR gets merged?

PR #51168, opened 2019-01-14
Add onchanges example to debconf docs

waynew:
This should help clear up confusion for people wanting to run
dpkg-reconfigure after setting values in debconf.

### What does this PR do?

Updates the docs for the debconf module to include an example for how to run `dpkg-reconfigure` on changes.

#4010 

### Commits signed with GPG?

Yes

### Tests Written?

I manually built the docs locally using sphinx, and verified that the output was what I intended (i.e. the code was in a YAML codeblock, etc.)

### DIFF
--- a/salt/states/debconfmod.py
+++ b/salt/states/debconfmod.py
@@ -36,6 +36,52 @@ set_file
 .. note::
     Due to how PyYAML imports nested dicts (see :ref:`here <yaml-idiosyncrasies>`),
     the values in the ``data`` dict must be indented four spaces instead of two.
+
+If you're setting debconf values that requires `dpkg-reconfigure`, you can use
+the ``onchanges`` requisite to reconfigure your package:
+
+.. code-block:: yaml
+
+    # or use America
+    set-timezone-area:
+      debconf.set:
+        - name: tzdata
+        - data:
+              'tzdata/Areas': {'type': 'select', 'value': 'US'}
+
+    # Or use tzdata/Zones/America, and Denver
+    set-timezone:
+      debconf.set:
+        - name: tzdata
+        - data:
+              'tzdata/Zones/US': {'type': 'select', 'value': 'Mountain'}
+
+    remove /etc/localtime:
+      file.absent:
+        - name: /etc/localtime
+        - onchanges:
+          - debconf: set-timezone-area
+          - debconf: set-timezone
+
+    remove /etc/timezone:
+      file.absent:
+        - name: /etc/timezone
+        - onchanges:
+          - debconf: set-timezone-area
+          - debconf: set-timezone
+
+    reconfigure-tzdata:
+      cmd.run:
+        - name: dpkg-reconfigure -f noninteractive tzdata
+        - onchanges:
+          - debconf: set-timezone-area
+          - debconf: set-timezone
+        - require:
+          - file: remove /etc/localtime
+          - file: remove /etc/timezone
+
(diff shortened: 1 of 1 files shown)


thatch45 (approved):

boltronics:
I'm surprised to see the removal of `/etc/localtime` and `/etc/timezone` in the example. Is that really necessary - wouldn't dpkg-reconfigure adjust these even if they already exist (which they typically would)? Wouldn't that potentially cause momentary problems for other running applications which rely on those?

waynew:
> I'm surprised to see the removal of `/etc/localtime` and `/etc/timezone` in the example. Is that really necessary - wouldn't dpkg-reconfigure adjust these even if they already exist (which they typically would)?

I thought so, too. You're welcome to try it out without removing those files, but I couldn't get it to work. For some reason, whatever combination of distribution/environment I was using would just not actually change the timezone if those files existed! At least, using dpkg-reconfigure+debconf. I certainly would have preferred not adding those 🙁 

> Wouldn't that potentially cause momentary problems for other running applications which rely on those?

From what I encountered while documenting this... It *shouldn't*. If by some magic miracle you were configuring tzdata like this at the very moment a server started up and you had a race condition between deleting the files and running `dpkg-reconfigure`, that could cause a problem, but most of what I read 1) said to remove those files and 2) some servers wouldn't even pick up the changed timezone until they needed to restart.

I would have even preferred a different example for debconf, but a minute or two of searching only had examples of how to reconfigure tzdata or mysql. I figured that tzdata was a far better choice than throwing a huge example in there for preconfiguring MySQL :trollface:

boltronics:
I just had a look, and sure enough it didn't work. Worse, it seems to be deliberate. I tracked the cause down to this section:

```
# Initializes debconf default values from the ones found in
# configuration files
db_set tzdata/Areas "$AREA"
db_set tzdata/Zones/$AREA "$ZONE"
```
in `/var/lib/dpkg/info/tzdata.config`

Commenting out those lines makes things `dpkg-reconfigure` behave as intended.

Unfortunately, it seems the proposed solution may indeed cause time issues as hinted at in this old bug report: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=516755

From the report:
```        rm -f /etc/localtime && \
            cp -f /usr/share/zoneinfo/$AREA/$ZONE /etc/localtime
```
 * There is a tiny race condition between the rm and the cp, which
   can lead to log files showing wrong timestamps.

With the switch to symlinks, this was addressed most recently by doing the following:
```
	ln -nsf /usr/share/zoneinfo/$AREA/$ZONE /etc/localtime.dpkg-new && \
	    mv -f /etc/localtime.dpkg-new /etc/localtime
```
Of course, I don't know that all Debian-based distributions have switched to using symlinks so that may not be a perfect solution either.

What a mess! Perhaps the `dash` would be a better example? From `debconf-get-selections`:
```
# Use dash as the default system shell (/bin/sh)?
dash	dash/sh	boolean	true
```

So you could instead just have:
```
set-default-shell:
  debconf.set:
    - name: dash
    - data:
          'dash/sh': {'type': 'boolean', 'value': true}

reconfigure-dash:
  cmd.run:
    - name: dpkg-reconfigure -f noninteractive dash
    - onchanges:
      - debconf: set-default-shell
```

which is at least a bit more concise.

boltronics:
FWIW, this is the tzdata bug: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=704089

Issue #51158, opened 2019-01-12
saltmod.state doesn't respect "roster" argument


AndreiPashkin:
### Description of Issue/Question
According to [the documentation](https://github.com/saltstack/salt/blob/c850a94f0e341d68a98851cbda4b93963d8b8f76/salt/states/saltmod.py#L192) `saltmod.state` module must accept `roster` argument:

> roster
>         In the event of using salt-ssh, a roster system can be set

If you analyze [the code](https://github.com/saltstack/salt/blob/c850a94f0e341d68a98851cbda4b93963d8b8f76/salt/states/saltmod.py#L108) you'll see that `roster` argument is not handled in any way.

### Setup
This is a basic orchestrate SLS-file that reproduces the issue.
```
deploy:
  salt.state:
    - tgt: '*'
    - highstate: true
    - pillarenv: live
    - saltenv: live
    - ssh: true
    - roster: my_custom_roster_system
```

### Steps to Reproduce Issue
- Make any Salt-Orchestrate setup that uses `saltmod.state` function.
- Provide an alternative roster system through `roster` argument.
- Observe that the default `flat` roster system is still used.

### Versions Report
```
Salt Version:
           Salt: 2018.3.3
 
Dependency Versions:
           cffi: Not Installed
       cherrypy: Not Installed
       dateutil: 2.6.1
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.10
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: 1.0.7
   msgpack-pure: Not Installed
 msgpack-python: 0.5.6
   mysql-python: Not Installed
      pycparser: Not Installed
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: Not Installed
         Python: 3.6.7 (default, Oct 22 2018, 11:32:17)
   python-gnupg: 0.4.1
         PyYAML: 3.12
          PyZMQ: 16.0.2
           RAET: Not Installed
          smmap: Not Installed
        timelib: Not Installed
        Tornado: 4.5.3
            ZMQ: 4.2.5
 
System Versions:
           dist: Ubuntu 18.04 bionic
         locale: ANSI_X3.4-1968
        machine: x86_64
        release: 4.15.0-43-generic
         system: Linux
        version: Ubuntu 18.04 bionic
```

dwoz:
[...]

PR #51170, opened 2019-01-14
Honor roster system for ssh orchestrations

dwoz:
### What does this PR do?

Honor roster file when passed to orchestration state.

#51158

### Tests written?

Yes

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/states/saltmod.py
+++ b/salt/states/saltmod.py
@@ -285,7 +285,7 @@ def state(name,
     cmd_kw['tgt_type'] = tgt_type
     cmd_kw['ssh'] = ssh
     if 'roster' in kwargs:
-        cmd_kw['roster_file'] = kwargs['roster']
+        cmd_kw['roster'] = kwargs['roster']
     cmd_kw['expect_minions'] = expect_minions
     if highstate:
         fun = 'state.highstate'
--- a/salt/states/saltmod.py
+++ b/salt/states/saltmod.py
@@ -284,6 +284,8 @@ def state(name,
 
     cmd_kw['tgt_type'] = tgt_type
     cmd_kw['ssh'] = ssh
+    if 'roster' in kwargs:
+        cmd_kw['roster_file'] = kwargs['roster']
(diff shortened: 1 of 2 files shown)


twangboy (approved):

salt/states/saltmod.py
```diff
@@ -284,6 +284,8 @@ def state(name,
 
     cmd_kw['tgt_type'] = tgt_type
     cmd_kw['ssh'] = ssh
+    if 'roster' in kwargs:
+        cmd_kw['roster_file'] = kwargs['roster']
```
AndreiPashkin: `roster` argument refers to roster-system, not roster-file. "Roster-file" is relevant within to "flat" roster system, but Salt allows other roster-systems to be used including custom ones.

AndreiPashkin (changes requested):

salt/states/saltmod.py
dwoz: The documentation says that `roster` for the orchestration state refers to a roster file. This PR is just to make the state honor what the documentation says. Passing the `roster_file` keyword arg to `saltutil.cmd` uses the default `roster` value of `flat`

AndreiPashkin: > The documentation says that roster for the orchestration state refers to a roster file.

Not it's not:

> roster
> In the event of using salt-ssh, a **roster system** can be set

https://docs.saltstack.com/en/latest/ref/states/all/salt.states.saltmod.html#salt.states.saltmod.state
Notice - roster **system**, not roster **file**

dwoz: Okay, thanks for pointing that out. I'll look at this again.

dwoz: @AndreiPashkin Can you reply to #51158 with the expected behavior for this option? I will try to re-implement this based on your feedback in the ticket.

dwoz: @AndreiPashkin Are we good here?

AndreiPashkin (approved):

Issue #51138, opened 2019-01-10
Windows 10 Minions use very high CPU if salt-key is rejected / unaccepted on master


tjgruber:
### Description of Issue/Question

My CPU usage is 51-100% + for Python.exe (salt-minion) on any PC who's denied/not accepted on the salt-master.

Background:
Multiple PCs, including my own, have been chewing high CPU resources constantly... fan spinning fast, degraded performance, etc. When I looked, it was python.exe (salt-minion) using over 51% of CPU and showing high power consumption.

Once I checked the salt-master, and removed the PCs via salt-key -d, then accepted them, it got better immediately.

In my case specifically, the computer was reloaded, so the old key was accepted, and the new but current key was denied. I removed both from salt-master, waited a second, then accepted the new current key.

I found an old identical issue here: https://github.com/saltstack/salt/issues/9052

### Versions

salt-minion (2018.3.3)
Windows 10 1809

twangboy:
So, when running the salt-minion in debug mode it looks like the salt-minion process just stops when the key has been denied by the master. That could cause this behavior when salt-minion is running as a service.

The service manager we use monitors the status of the minion. If the process dies for any reason, the process manager will restart the salt-minion service. So, the minion will try to connect, find out it's denied, and die. The service manager will restart the service again and the process repeats. I'm guessing this is what you're seeing.

The service manager does have a `delay restart` option. It's set to 0 by default... no delay. I set it to 10 secs and now I see a cpu spike every 10 seconds. Maybe 60 seconds would be a good default. To set this do the following:

1. Open a comman prompt
2. Navigate to `C:\salt\bin`
3. Type: `ssm set salt-minion AppRestartDelay 60000`

That should fix the issue. We'll need to add this to the installer for situations such as this.

@dwoz @UtahDave Does 60 seconds sound like a sane default?

tjgruber:
> I set it to 10 secs and now I see a cpu spike every 10 seconds.

How long does the spike last?

I'm asking because once the denied key is deleted on the master, it could then take up to 60 seconds for

tjgruber:
Hit the wrong link.

twangboy:
This fix above addresses the issue from Salt code. It adds a random sleep of 10 to 20 seconds before it issues the `sys.exit` command.

PR #51171, opened 2019-01-14
Add random sleep time to avoid overloading Windows

twangboy:
### What does this PR do?
Fixes an issue on Windows where the salt-minion restarts over and over when the key has been denied on the master. This adds a random delay between 10 and 20 seconds before the minion exits.

https://github.com/saltstack/salt/issues/51138

### Tests written?
No

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/crypt.py
+++ b/salt/crypt.py
@@ -9,6 +9,7 @@ authenticating peers
 # the Array class, which has incompatibilities with it.
 from __future__ import absolute_import, print_function
 import os
+import random
 import sys
 import copy
 import time
@@ -727,6 +728,10 @@ class AsyncAuth(object):
                             'minion.\nOr restart the Salt Master in open mode to '
                             'clean out the keys. The Salt Minion will now exit.'
                         )
+                        # Add a random sleep here for systems that are using a
+                        # a service manager to immediately restart the service
+                        # to avoid overloading the system
+                        time.sleep(random.randint(10, 20))
                         sys.exit(salt.defaults.exitcodes.EX_NOPERM)
                 # has the master returned that its maxed out with minions?
                 elif payload['load']['ret'] == 'full':


dwoz (approved):

salt/crypt.py
```diff
@@ -727,6 +728,10 @@ def sign_in(self, timeout=60, safe=True, tries=1, channel=None):
                             'minion.\nOr restart the Salt Master in open mode to '
                             'clean out the keys. The Salt Minion will now exit.'
                         )
+                        # Add a random sleep here for systems that are using a
+                        # a service manager to immediately restart the service
+                        # to avoid overloading the system
+                        time.sleep(random.randint(10, 20))
```
damon-atkins: Are you sure the sleep is in the right spot, the next statement is 
`sys.exit(salt.defaults.exitcodes.EX_NOPERM)`    What is the value in sleep then exit.

damon-atkins:
A better solution on windows is to adjust the service, rather than effect all platforms.
```
nssm set UT2003 AppThrottle 6000  $ 60 sec as python a bit slow to start up
nssm set UT2003 AppExit Default Restart  # already the default.
nssm set UT2003 AppRestartDelay 30000  # 5 min
```

damon-atkins:
ping @saltstack/team-core

damon-atkins:
Most of missed the notification for this discussion #51138

scott-korin:
> A better solution on windows is to adjust the service, rather than effect all platforms.
> 
> ```
> nssm/ssm.exe set salt-minion AppThrottle 6000  # Exits within 60 sec,  as python a bit slow to start up
> nssm/ssm.exe set salt-minion AppExit Default Restart  # already the default.
> nssm/ssm.exe set salt-minion AppRestartDelay 30000  # Then delay restart by 5 min, service will indicate paused
> ```

Is there a way to set these values when salt-minion is installed?  I don't see any of these values documented anywhere.  My google-fu must be very poor, because I can't find any kind of documentation on the Salt Service Manager at all.

Since these are settings for ssm, not the salt-minion itself, I assume these values aren't in minion config file?  

I am just looking for a way to automate setting these during an install or update them once the service is installed.  Just run the ssm command?

scott-korin:
> > A better solution on windows is to adjust the service, rather than effect all platforms.
> > ```
> > nssm/ssm.exe set salt-minion AppThrottle 6000  # Exits within 60 sec,  as python a bit slow to start up
> > nssm/ssm.exe set salt-minion AppExit Default Restart  # already the default.
> > nssm/ssm.exe set salt-minion AppRestartDelay 30000  # Then delay restart by 5 min, service will indicate paused
> > ```
> 
> Is there a way to set these values when salt-minion is installed? I don't see any of these values documented anywhere. My google-fu must be very poor, because I can't find any kind of documentation on the Salt Service Manager at all.
> 
> Since these are settings for ssm, not the salt-minion itself, I assume these values aren't in minion config file?
> 
> I am just looking for a way to automate setting these during an install or update them once the service is installed. Just run the ssm command?
> 
> Are there other related parameters? Is there documentation somewhere on these?

ok, so it looks like ssm is just a SaltStack branded version of https://nssm.cc/ and what it is doing is updating WIndows service settings.

Issue #6237, opened 2013-07-19
require_in overrides excludes


josmardias:
I'm using Fedora 19

```
$ salt --versions-report
           Salt: 0.15.3
         Python: 2.7.5 (default, Jul  8 2013, 09:48:59)
         Jinja2: 2.6
       M2Crypto: 0.21.1
 msgpack-python: 0.1.13
   msgpack-pure: Not Installed
       pycrypto: 2.6
         PyYAML: 3.10
          PyZMQ: 13.0.0
            ZMQ: 3.2.3

```

State:

```
test1:
  file.managed:
    - name: /tmp/nothing

test2:
  file.managed:
    - name: /tmp/nothing2
    - require:
      - file: test1

exclude:
  - id: test1
```

Output

```
$ sudo salt-call state.highstate
[INFO    ] Configuration file path: /etc/salt/minion
[INFO    ] Executing command 'ps -efH' in directory '/root'
[INFO    ] Loading fresh modules for state activity
[INFO    ] Fetching file 'salt://top.sls'
[INFO    ] Creating module dir '/var/cache/salt/minion/extmods/modules'
[INFO    ] Syncing modules for environment 'base'
[INFO    ] Loading cache from salt://_modules, for base)
[INFO    ] Caching directory '_modules/' for environment 'base'
[INFO    ] Creating module dir '/var/cache/salt/minion/extmods/states'
[INFO    ] Syncing states for environment 'base'
[INFO    ] Loading cache from salt://_states, for base)
[INFO    ] Caching directory '_states/' for environment 'base'
[INFO    ] Creating module dir '/var/cache/salt/minion/extmods/grains'
[INFO    ] Syncing grains for environment 'base'
[INFO    ] Loading cache from salt://_grains, for base)
[INFO    ] Caching directory '_grains/' for environment 'base'
[INFO    ] Creating module dir '/var/cache/salt/minion/extmods/renderers'
[INFO    ] Syncing renderers for environment 'base'
[INFO    ] Loading cache from salt://_renderers, for base)
[INFO    ] Caching directory '_renderers/' for environment 'base'
[INFO    ] Creating module dir '/var/cache/salt/minion/extmods/returners'
[INFO    ] Syncing returners for environment 'base'
[INFO    ] Loading cache from salt://_returners, for base)
[INFO    ] Caching directory '_returners/' for environment 'base'
[INFO    ] Creating module dir '/var/cache/salt/minion/extmods/outputters'
[...]

PR #51183, opened 2019-01-16
RFC - changing requisites/excludes

waynew:
RFC text: https://github.com/waynew/salt/blob/rfc-changing-requisites/rfcs/0005-change-requisite-handling.md

There were some suggestions made on #6237 about how we should do requisites/excludes, but no actual decisions made.

I'm making this RFC so if we decide to *not* make code changes we have a place to point and say, "This is why."

Alternatively, if we decide it's something we want to change, well, there we go!

### DIFF
--- a/rfcs/0005-change-requisite-handling.md
+++ b/rfcs/0005-change-requisite-handling.md
@@ -1,0 +1,94 @@
+- Feature Name: `change_requisite_handling`
+- Start Date: 2019-01-15
+- RFC PR: (leave this empty)
+- Salt Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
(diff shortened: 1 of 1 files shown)


thatch45:
I think this is a really good idea, but I also think that we should make sure that this would be configurable. Part of the benefit of having excludes not remove requisites is that if you break something with an exclude then you would want to know about it.

Also, if you want this behavior today then the correct way to make the requisites would be to use requisite_ins from the states that might be excluded to that you are able to apply intent directionally

thatch45:
I think that argument would be fine :)

s0undt3ch:
Both `--ebr`, `--exclude-before-requisites` for the same option?

thatch45:
Sounds resonable

waynew:
Migrated to the SEP repo: https://github.com/saltstack/salt-enhancement-proposals/pull/10

PR #51187, opened 2019-01-16
The network connection is large, causing the to run full when the salt detection is alive

menglong81:
Reading the /proc/net/tcp file for a machine with a large network connection will consume CPU. So we changed to use the ss command instead of the netstat command to get the network connection first. The ss uses the tcp_diag in the TCP protocol stack so it will be faster than Netstat command is much faster
master_alive_interval options
Salt will call the netstat command when it performs health check. However, if the number of system connections is large, it will consume cpu because it reads the file /proc/net/tcp
So we modified the code and hope to adopt it


dwoz (commented):
The `ss` command is not available on MacOS and some other unix variants. Have you validated this change across all OSes Salt supports?

menglong81:
The ss command can be used on most os used by enterprises. If there is no such command on os, then this command will be skipped and the value will continue to run using native code.

menglong81:
for line in lines:
        if 'Address:Port' in line:    # ss tools may not be valid
            valid = True
            continue

dwoz (approved):

dwoz:
@saltstack/team-core Can we get one more set of eyes on this before it's merged?

menglong81:
OK thank u

menglong81:
OK thank u

garethgreenaway (approved):

PR #51204, opened 2019-01-16
Lint: Fix str-format-in-logging, len-as-condition, etc. (5)

terminalmage:
This fixes violations found in the following locations:

- salt/netapi/
- salt/output/
- salt/pillar/
- salt/proxy/
- salt/queues/
- salt/renderers/
- salt/returners/
- salt/runners/
- salt/sdb/
- salt/serializers/
- salt/spm/
- tests/

### DIFF
--- a/salt/netapi/rest_cherrypy/app.py
+++ b/salt/netapi/rest_cherrypy/app.py
@@ -720,11 +720,11 @@ def salt_api_acl_tool(username, request):
     :type request: cherrypy.request
     '''
     failure_str = ("[api_acl] Authentication failed for "
-                   "user {0} from IP {1}")
+                   "user %s from IP %s")
     success_str = ("[api_acl] Authentication sucessful for "
-                   "user {0} from IP {1}")
+                   "user %s from IP %s")
     pass_str = ("[api_acl] Authentication not checked for "
-                "user {0} from IP {1}")
+                "user %s from IP %s")
 
     acl = None
     # Salt Configuration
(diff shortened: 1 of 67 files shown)


salt/netapi/rest_cherrypy/app.py
```diff
@@ -1239,7 +1239,7 @@ def GET(self):
             HTTP/1.1 200 OK
             Content-Type: application/json
         '''
-        import inspect
+        import inspect  # pylint: disable=unused-import
```
s0undt3ch: Actually remove the import?
Or at least document why the import is done since it's not used anywhere in the module?

salt/renderers/hjson.py
```diff
@@ -11,7 +11,7 @@
 
 # Import 3rd party libs
 try:
-    import hjson as hjson
```
s0undt3ch: Really?!

s0undt3ch (approved):

salt/netapi/rest_cherrypy/app.py
terminalmage: I wasn't sure if it was needed somewhere else, so I was being cautious. Sometimes we do late imports because they are implicitly needed by other code in a given function.

salt/renderers/hjson.py
terminalmage: :smiley:

salt/netapi/rest_cherrypy/app.py
s0undt3ch: yeah, but in those cases we should add a comment. Anyway, not a blocker, leave it, it's safer this way.

terminalmage:
There is still a failure in the couchdb returner. I'm going to take another look. It was only failing when I ran pylint on everything and not just on the file itself, so this may be a pylint bug.

terminalmage:
OK that lint failure is fixed now

terminalmage:
Ported to master in https://github.com/saltstack/salt/pull/57439

PR #51205, opened 2019-01-16
Lint: Fix str-format-in-logging, len-as-condition, etc. (6)

terminalmage:
This fixes violations found in the following locations:

- salt/state.py
- salt/states/
- salt/tops/
- salt/transport/

### DIFF
--- a/salt/states/btrfs.py
+++ b/salt/states/btrfs.py
@@ -32,9 +32,9 @@ import functools
 import logging
 import os.path
 import tempfile
-import traceback
 
 from salt.exceptions import CommandExecutionError
+from salt.ext import six
 
 log = logging.getLogger(__name__)
 
@@ -119,8 +119,8 @@ def __mount_device(action):
             kwargs['__dest'] = dest
             ret = action(*args, **kwargs)
         except Exception as e:
-            log.error('''Traceback: {}'''.format(traceback.format_exc()))
-            ret['comment'].append(e)
+            log.exception('Encountered error mounting %s', device)
+            ret['comment'].append(six.text_type(e))
         finally:
             _umount(dest)
         return ret
--- a/salt/states/dvs.py
+++ b/salt/states/dvs.py
@@ -204,7 +204,6 @@ Module was developed against.
 # Import Python Libs
 from __future__ import absolute_import, print_function, unicode_literals
 import logging
-import traceback
 import sys
 
 # Import Salt Libs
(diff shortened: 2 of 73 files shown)


salt/states/dvs.py
```diff
@@ -608,7 +605,7 @@ def portgroups_configured(name, dvs, portgroups):
                                 {'old': current_pg}})
         __salt__['vsphere.disconnect'](si)
     except salt.exceptions.CommandExecutionError as exc:
-        log.error('Error: {0}\n{1}'.format(exc, traceback.format_exc()))
+        log.traceback('Encountered error')
```
s0undt3ch: `log.exception()`?

s0undt3ch (changes requested):

salt/states/dvs.py
terminalmage: d'oh, yes

terminalmage: Should be fixed now.

s0undt3ch (approved):

terminalmage:
I fixed some incorrect mocking that should resolve the mac test failures.

salt/states/btrfs.py
```diff
@@ -119,8 +119,8 @@ def wrapper(*args, **kwargs):
             kwargs['__dest'] = dest
             ret = action(*args, **kwargs)
         except Exception as e:
-            log.error('''Traceback: {}'''.format(traceback.format_exc()))
-            ret['comment'].append(e)
+            log.exception('Encountered error mounting %s', device)
+            ret['comment'].append(six.text_type(e))
```
aplanas: You introduced two error here:

* This error is not a mount error, is the exception produced in the method that is wrapping this code (the call from line 120). The mount error was taken care a bit before.

* 'e' do not contain the traceback, and we do not want to hide it. We need to print it in the log at least (and maybe in the comments too)

I will change that in a different pull request, and I will integrate your fix for log.exception and six.text_type

terminalmage: >     * This error is not a mount error, is the exception produced in the method that is wrapping this code (the call from line 120). The mount error was taken care a bit before.

OK, so the message in the log was inaccurate, fair enough.

>     * 'e' do not contain the traceback, and we do not want to hide it. We need to print it in the log at least (and maybe in the comments too)

I have no idea what you're talking about. Nothing was hidden. Logging at `exception` level is the same thing as doing `log.error('some message', exc_info=True)`. It includes the exception.

terminalmage:
Ported to master in https://github.com/saltstack/salt/pull/57443

PR #51206, opened 2019-01-16
Fix usage of reload for PY3 compatibility

terminalmage:
This also normalizes some hacks that were in place to import reload from
importlib, so that these instances also use the compatibility function.

### DIFF
--- a/salt/modules/ebuild.py
+++ b/salt/modules/ebuild.py
@@ -22,6 +22,7 @@ import re
 
 # Import salt libs
 import salt.utils.args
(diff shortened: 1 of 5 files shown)


s0undt3ch (approved):

terminalmage:
Weird, I had that import there before in develop, I must have removed it accidentally in the process of backporting. Thanks for catching it, @s0undt3ch!

s0undt3ch:
yeah, the other PR's have the import.

Issue #50538, opened 2018-11-16
salt.modules.pdbedit broken on Samba 4.9.x


sjorge:
### Description of Issue/Question
The pdbedit module is broken again after upgrading to Samba 4.9.2

Both pdbedit.list in verbose and non-verbose mode.

```
[atlas :: sjorge][~]
[.]$ salt-call pdbedit.list verbose=False
local:
    - Registered MSG_REQ_POOL_USAGE
    - Registered MSG_REQ_DMALLOC_MARK and LOG_CHANGED
    - lp_load_ex
    - Initialising global parameters
    - Processing section "[global]"
    - tmartine
    - sjorge
```

For non-verbose we should filter out all lines that do not follow the 'login:uid:full name' format,

```
[atlas :: sjorge][~]
[.]$ salt-call pdbedit.list
[ERROR   ] An un-handled exception was caught by salt's global exception handler:
KeyError: u'unix username'
Traceback (most recent call last):
  File "/opt/local/bin/salt-call", line 11, in <module>
    salt_call()
  File "/opt/local/lib/python2.7/site-packages/salt/scripts.py", line 410, in salt_call
    client.run()
  File "/opt/local/lib/python2.7/site-packages/salt/cli/call.py", line 57, in run
    caller.run()
  File "/opt/local/lib/python2.7/site-packages/salt/cli/caller.py", line 134, in run
    ret = self.call()
  File "/opt/local/lib/python2.7/site-packages/salt/cli/caller.py", line 212, in call
    ret['return'] = func(*args, **kwargs)
  File "/var/cache/salt/minion/extmods/modules/pdbedit.py", line 100, in list_users
    users[user_data['unix username']] = user_data
KeyError: u'unix username'
Traceback (most recent call last):
  File "/opt/local/bin/salt-call", line 11, in <module>
    salt_call()
  File "/opt/local/lib/python2.7/site-packages/salt/scripts.py", line 410, in salt_call
    client.run()
  File "/opt/local/lib/python2.7/site-packages/salt/cli/call.py", line 57, in run
    caller.run()
  File "/opt/local/lib/python2.7/site-packages/salt/cli/caller.py", line 134, in run
    ret = self.call()
  File "/opt/local/lib/python2.7/site-packages/salt/cli/caller.py", line 212, in call
    ret['return'] = func(*args, **kwargs)
  File "/var/cache/salt/minion/extmods/modules/pdbedit.py", line 100, in list_users
    users[user_data['unix username']] = user_data
[...]

PR #51221, opened 2019-01-17
Backport #50538 into 2019.2

Ch3LL:
Backport #50538 into 2019.2

conflicts:
  - salt/modules/pdbedit.py

### DIFF
--- a/salt/modules/pdbedit.py
+++ b/salt/modules/pdbedit.py
@@ -92,29 +92,34 @@ def list_users(verbose=True, hashes=False):
 
         if res['retcode'] > 0:
             log.error(res['stderr'] if 'stderr' in res else res['stdout'])
-        else:
-            user_data = {}
-            for user in res['stdout'].splitlines():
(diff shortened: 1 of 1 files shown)


salt/modules/pdbedit.py
```diff
@@ -92,29 +92,34 @@ def list_users(verbose=True, hashes=False):
...
+                if 'unix username' in user_data:
+                    users[user_data['unix username']] = user_data
+                user_data = {}
+            elif ':' in user:
+                label = user[:user.index(':')].strip().lower()
+                data = user[(user.index(':')+1):].strip()
+                user_data[label] = data
+
+        if user_data:
+            users[user_data['unix username']] = user_data
```
sjorge: This probably also applies to what we currently have in develop branch too...

We should probably also check if we actually have ‘unix username’ instead of assuming we do. (See line 100)

sjorge (approved):
LGTM, although I did spot one thing that could make this more robust.

PR #51240, opened 2019-01-18
The mysql version check inverted

waynegemmell:
The mysql version check that was supposed to test whether the version is greater than or equal to  8.0.11 was testing for less than or equal to 8.0.11. This then used the password(_) function in the incorrect context.

### Tests written?

No

### Commits signed with GPG?

No


waynegemmell:
This was supposed to go in against tag 2018.3.3. I'm not sure what went wrong. Selecting the tag on the pull request screen sent me into a code comparison mode.

salt/modules/mysql.py
```diff
@@ -1242,7 +1242,7 @@ def user_exists(user,
         else:
             qry += ' AND ' + password_column + ' = \'\''
     elif password:
-        if salt.utils.versions.version_cmp(server_version, '8.0.11') <= 0:
+        if salt.utils.versions.version_cmp(server_version, '8.0.11') >= 0:
```
rares-pop: what about salt.utils.versions.LooseVersion(server_version) >= '8.0.11'

garethgreenaway:
@waynegemmell Looks like your local branch was against `develop`, switching the base branch in the PR results in a lot of changes beyond your change.  That's okay though, we can merge this into `develop` then back port it to 2018.3.

garethgreenaway:
Looking closer at the code already committed, this was fixed in https://github.com/saltstack/salt/pull/49918.
Closing this PR out.

PR #51243, opened 2019-01-18
Several Test Fixes Found During PyTest Migration

s0undt3ch:
### What does this PR do?
See title

### Commits signed with GPG?

Yes


salt/config/__init__.py
```diff
@@ -4154,10 +4154,10 @@ def api_config(path):
     need to be stubbed out for salt-api
     '''
     # Let's grab a copy of salt-api's required defaults
-    opts = DEFAULT_API_OPTS
+    opts = DEFAULT_API_OPTS.copy()
 
     # Let's override them with salt's master opts
-    opts.update(client_config(path, defaults=DEFAULT_MASTER_OPTS))
+    opts.update(client_config(path, defaults=DEFAULT_MASTER_OPTS.copy()))
```
waynew: Are we requiring a `.copy` in all the cases, or just some?

If it's all the cases, it might be useful to just change to a pattern like [this one](https://stackoverflow.com/a/39440252/344286), which makes it no longer possible for someone to accidentally forget `.copy()`. Thinking about it, one wouldn't even need to use frozenset, something like this would work just fine:

```
DEFAULT_API_OPTS = tuple({'foo': 'bar', 'salt': 'sodium chloride', 'water': 'dihydrogen monoxide'}.items())
...

opts.update(..., defaults=dict(DEFAULT_API_OPTS))
```

Not sure if the change is worth it, but it would at least enforce copying!

salt/modules/beacons.py
```diff
@@ -361,7 +382,7 @@ def save():
     beacons = list_(return_yaml=False, include_pillar=False)
 
     # move this file into an configurable opt
-    sfn = os.path.join(__opts__['config_dir'],
+    sfn = os.path.join(os.path.dirname(__opts__['conf_file']),
```
waynew: Is this a bug fix? I know I was looking at the config dir thing recently, because if no `minion` file exists there then we override it with the default :)

If this is a pattern that we should be following everywhere, FWIW there are a few other locations using `config_dir`:

```
 %  ~/p/salt  ╍╍ rfc-changing-requisites  salt  ag opts__..config_dir .
wheel/config.py
75:    dir_path = os.path.join(__opts__['config_dir'],

cloud/clouds/lxc.py
68:        default_dir = __opts__['config_dir'],
79:        default_dir = __opts__['config_dir'],

modules/beacons.py
364:    sfn = os.path.join(__opts__['config_dir'],

modules/schedule.py
837:    sfn = '{0}/{1}/schedule.conf'.format(__opts__['config_dir'], os.path.dirname(__opts__['default_include']))
```

s0undt3ch: I believe `config_dir` could be an artifact from code that we dropped.
`config_dir` is defined when the CLI scripts run, but that value is not "transposed" to the loaded config.
This change is basically removing those artifacts, but apparently I missed some.

salt/config/__init__.py
s0undt3ch: https://github.com/saltstack/salt/pull/51245

s0undt3ch:
Need to get this in, any new failures are on me to solve, though, the windows failures are unrelated to these changes.

Issue #50221, opened 2018-10-25
file.managed: contents_pillar and pillar.file_tree with keep_newline is broken in 2018.3.3


tkaehn:
### Description of Issue/Question

Using contents_pillar and pillar.file_tree with keep_newline is broken in 2018.3.3 and results in removing lines.

Probably file.managed is broken as `pillar.get` shows the newlines (please see below).

### Setup and steps to reproduce

```
root@minion:~# cat /etc/salt/minion.d/file_tree_pillar.conf 
ext_pillar:
  - file_tree:
      root_dir: /srv/pillar/file_tree
      follow_dir_links: False
      keep_newline: True

root@minion:~# cat /srv/salt/motd.sls 
---
/etc/motd:
  file.managed:
    - contents_pillar: motd

root@minion:~# cp /srv/pillar/file_tree/hosts/minion/motd /etc/motd 

root@minion:~# wc -l /srv/pillar/file_tree/hosts/minion/motd /etc/motd 
 3 /srv/pillar/file_tree/hosts/minion/motd
 3 /etc/motd
 6 total

root@minion:~# salt-call --output=json pillar.get motd
{
    "local": "abc\n\n\n"
}

root@minion:~# salt-call state.apply motd test=False
local:
----------
          ID: /etc/motd
    Function: file.managed
      Result: True
     Comment: File /etc/motd updated
     Started: 10:47:58.830950
    Duration: 24.08 ms
     Changes:   
              ----------
              diff:
                  --- 
                  +++ 
                  @@ -1,3 +1 @@
                   abc
                  -
                  -

Summary for local
------------
Succeeded: 1 (changed=1)
Failed:    0
------------
Total states run:     1
Total run time:  24.080 ms

root@minion:~# wc -l /srv/pillar/file_tree/hosts/minion/motd /etc/motd
 3 /srv/pillar/file_tree/hosts/minion/motd
 1 /etc/motd
 4 total
```

### Versions Report
```
root@minion:~# salt-call --versions-report
Salt Version:
           Salt: 2018.3.3
 
Dependency Versions:
           cffi: Not Installed
       cherrypy: Not Installed
       dateutil: 2.5.3
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.9.4
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: Not Installed
   msgpack-pure: Not Installed
[...]

PR #51252, opened 2019-01-19
Peserve newlines from pillar data for file.managed [2018.3]

dwoz:
### What does this PR do?

Preserve newlines from pillar data when using it for contents of file.managed.

#50221

### Tests written?

Yes

### Commits signed with GPG?

Yes


salt/states/file.py
```diff
@@ -2435,11 +2434,13 @@ def managed(name,
...
-            contents = os.linesep.join(
-                [line.rstrip('\n').rstrip('\r') for line in validated_contents]
-            )
+            contents = ''
+            for part in validated_contents:
+                for line in part.splitlines():
+                    contents += line.rstrip('\n').rstrip('\r') + os.linesep
             if contents_newline and not contents.endswith(os.linesep):
                 contents += os.linesep
+            log.error("USE CONTENTS %s %s %s", repr(use_contents), repr(validated_contents), repr(contents))
```
terminalmage: Instead of manually calling `repr`, you can use `%r` instead of `%s` and string replacement will automagically call `repr()`.

terminalmage (changes requested):

salt/states/file.py
dwoz: Thanks for pointing that out! This log was for debug so I will remove it.

terminalmage (approved):

s0undt3ch (approved):

PR #51253, opened 2019-01-20
boto: Replace some unnecessary usage of locals()

### LINKED ISSUE #51196: Unused variable (possible bug) in boto_lambda state module
While linting the codebase, I found that the `role_arn` variable in [this block of code](https://github.com/saltstack/salt/blob/3273bbd/salt/states/boto_lambda.py#L350-L356) is unused. The`options` dict instead specifies `Role` as `'role_arn'`, that is a quoted string instead of the variable name.

As I'm not familiar with this code, I am looping in @saltstack/team-boto for their input. It seems like if this were really broken, it would be affecting people for several years by now.

Should we be using the `role_arn` variable, or should we keep the quoted string and remove the call to `_get_role_arn()`? It looks to me like we should do the former instead of the latter, but I'd like the input of someone more familiar with this code.

terminalmage:
This makes the code less misleading. Note that I've also changed the variable names used in dict iteration here, as `val` implies that it contains a value and it contains a dictionary _key_ instead.

Resolves #51196.

### DIFF
--- a/salt/modules/boto3_sns.py
+++ b/salt/modules/boto3_sns.py
@@ -319,8 +319,8 @@ def subscribe(TopicArn=None, Protocol=None, Endpoint=None,
     conn = _get_conn(region=region, key=key, keyid=keyid, profile=profile)
     kwargs = {k: v for k, v in kwargs.items() if not k.startswith('_')}
     ## Begin warn_until()
-    if any((locals().get('TopicArn'), locals().get('Protocol'), locals().get('Endpoint'))):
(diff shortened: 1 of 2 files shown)


dwoz (commented):
@terminalmage Looks good to me, there are a few test failures though.

s0undt3ch (approved):

terminalmage:
Actually no, this was just a bug in my commit, I misinterpreted the test failure. I've fixed it now.

s0undt3ch (approved):

terminalmage:
Ported to master in https://github.com/saltstack/salt/pull/56697

PR #51262, opened 2019-01-21
Remove PendingDeprecationWarning

johnvox:
### What does this PR do?
It avoid future deprecated function

None (except pytest's testinfra warning issue)

### Tests written?

No

### Commits signed with GPG?

No


dwoz:
This should resolve #51398

terminalmage (approved):
This will require a backport. Do we know where this popped up the first time? Was it just in the RC?

garethgreenaway:
This should also resolve #51666

Oloremo:
Looking forward for this to be merged.

waynew:
FWIW, the `raw` keyword doesn't exist in older distributions of msgpack, which is why some of those builds are failing. Will we have dropped support for those older versions by the time `develop` is released?

s0undt3ch:
A simple `try/except TypeError` falling back to the old kwargs might be enough to maintain compat with older versions.

waynew:
FWIW, `encoding='utf-8'` is found in test/integration/__init__.py

Not sure if there are any other places we would want to fix this.

waynew:
I'm pretty sure that this has been solved by https://github.com/saltstack/salt/pull/55354

PR #51264, opened 2019-01-21
an argument is not always needed to be passed to open a file

gtmanfred:
### What does this PR do?
Sometimes this is loaded using just the default, which is 'r', and 'r' is not passed.  we should still be able to open the file.

### Tests written?

Yes

### Commits signed with GPG?

Yes


gtmanfred:
@saltstack/team-core this would be super nice to get into 2018.3.4, since it is pretty minor and only broken starting in 2018.3.3

Thanks,
Daniel

gtmanfred:
Broken by https://github.com/saltstack/salt/commit/eaed033cdea35dfb6a19cbf134614b134228593d

@dmurphy18 can you take a look?

Thanks,
Daniel

salt/utils/files.py
```diff
@@ -416,11 +416,13 @@ def flopen(*args, **kwargs):
     '''
     Shortcut for fopen with lock and context manager.
     '''
-    with fopen(*args, **kwargs) as f_handle:
+    filename = args.pop(0)
+    vars = 'wa'
```
s0undt3ch: Could we use something else instead of `vars`? it shadows a built-in.

s0undt3ch (changes requested):

salt/utils/files.py
gtmanfred: yeah

gtmanfred:
ok, fixed.

s0undt3ch (approved):

thatch45 (approved):

thatch45:
Are these test failures related? This is some very central code :)

gtmanfred:
checking.

gtmanfred:
Ok, test should be fixed.

s0undt3ch (approved):

PR #51261, opened 2019-01-21
Fix Zabbix returner

OrangeDog:
I guess nobody ever tested this, because it can never work
```
[ERROR   ] The return failed for job 20190121133315768206: 'zabbix_config'
   File "/usr/lib/python3/dist-packages/salt/minion.py", line 1723, in _thread_return
 Traceback (most recent call last):
     zabbix_send("salt.trap.info", host, 'SALT {0} OK'.format(job_minion_id))
 KeyError: 'zabbix_config'
   File "/usr/lib/python3/dist-packages/salt/returners/zabbix_return.py", line 91, in returner
     minion_instance.returners[returner_str](ret)
   File "/usr/lib/python3/dist-packages/salt/returners/zabbix_return.py", line 59, in zabbix_send
     with salt.utils.files.fopen(zbx()['zabbix_config'], 'r') as file_handle:
```

PR #51265, opened 2019-01-21
Rewrite zabbix returner to be more useful

OrangeDog:
Having fixed the current returner in #51261 I have now discovered that the way in which it worked made it nearly useless in Zabbix.
This changes the whole thing so Zabbix gets the actual return data (to present to others) and is able to set up appropriate problem and resolution triggers (e.g. for scheduled state.apply and test.ping).


OrangeDog:
I'm not familiar with Pygments. What would be the syntax to get a plain code block?

dwoz:
@OrangeDog I mad that docstring a raw string. Lets see if that fixes it.

dwoz (approved):

PR #51271, opened 2019-01-22
Windows: Add support for preventing password changes and expiry in user

syphernl:
We noticed out IIS Application Pools would eventually end up being stopped because the users created via Salt didn't set the "Password never expires" checkbox and the password eventually expired.

This PR addresses this and adds support for this and a few other things that should improve the functionality of the state and module.

### What has been done
- Added module support in `add` for stuff that's currently only possible in `change`
- Added support for disabling a user
- Added support for disabling password expiry of a user
- Added support for disabling password changes by a user
- `unlock_account` is now changed to `lock_account` since it makes more sense

### Notes
- Tested on Windows 2016 (1607) with Salt-minion 2018.3.2 / Salt-master 2018.3.3


salt/modules/win_useradd.py
```diff
@@ -194,7 +229,7 @@ def update(name,
            expiration_date=None,
            expired=None,
            account_disabled=None,
-           unlock_account=None,
+           lock_account=None,
```
rares-pop: While this is cleaner, this breaks backwards compatibilty. Should this go through a deprecation path, or nobody is worried about this. @dwoz ?

dwoz: @rares-pop I agree that it's unlikely many people use this functionality. However, I do think we need to go through the deprecation path, just to be safe.

dwoz: @syphernl We should follow the depreciation path defined here: https://docs.saltstack.com/en/latest/topics/development/deprecations.html

dwoz:
@syphernl Ping!

syphernl:
@dwoz Unfortunately since I'm swamped at work right now this'll take some time until I'll be able to work on this. Feel free to update this PR as you see fit.

salt/modules/win_useradd.py
syphernl: @dwoz The command has been re-added and marked as deprecated.

dwoz:
@syphernl We are no longer accepting PRs to the `develop` branch. This needs to be re-opened against the `master` branch. You can link to this PR from the new one to preserve any disscussion history. Sorry for the confusion.

PR #49755, opened 2018-09-24
fix #49754 minion zmq connecting to master configured explicitly as IPv6 address or IPv6 only master

aphor:
### What does this PR do?
fixes implementation of IPv6 address handling by minion and zeromq transport to enable minions to connect to a master identified only by IPv6 address or address/port

Work:
  - depends on #49705 [MERGED]
  - unit tests failing [FIXED]
  - integration tests failing [FIXED]
  - parsing config master: "host:port" [FIXED]
  - missing coverage for host:port parsing [FIXED]

### What issues does this PR fix or reference?
#49754, possibly #40515

### Tests written?

Yes, or rather test coverage added to existing tests.

### Commits signed with GPG?

Yes

PR #51274, opened 2019-01-22
Fix minion start issue

twangboy:
### What does this PR do?
Fixes an issue introduced by https://github.com/saltstack/salt/pull/49755 which was causing the minion to fail to start on Windows with the following error: 
```
[ERROR   ] Master address: '104.236.251.190' could not be resolved. Invalid or unresolveable address. Set 'master' value in minion config.
[INFO    ] Shutting down the Salt Minion
The Salt Minion is shutdown.
Master address: '104.236.251.190' could not be resolved. Invalid or unresolveable address. Set 'master' value in minion config.
```

The issue is in a new function in the salt.utils.network salt util. It was setting the host to with the following command: 
```
host_ip = ipaddress.ip_address(host)
```
Which sets the host_ip to an IPv4Address object.  This PR adds `.compressed` that sets `host_ip` to the actual string value instead of an `IPv4Address` object
```
host_ip = ipaddress.ip_address(host).compressed
```
It does this on the IPv4 portion and the IPv6 portion as well.

@aphor Please test this.  I don't know how it affects the issue you were trying to fix.

.. note::
    I found this issue by doing a git bisect which brought me to this commit: df7338809bb42

### Tests written?
No

### Commits signed with GPG?
Yes


aphor:
I'll look at it, but I don't have time until next week.

aphor:
Can you add a stack trace for this error? I don't have a Windows test VM handy.

aphor:
I went to a lot of trouble at the request of @isbm to move things towards passing around IP addresses as IP address objects instead of passing around strings. I'd like to understand how your error is happening, and why there is an attempt to do DNS resolution on an IP address.

This change works like duct tape, and I'd like to get closer to the root if we can, and fix the unit tests so we don't end up going in circles trying to fix things naïvely. This is breaking tests which run on Linux.

s0undt3ch:
Leaving ipaddress instances in opts brings several problems.

I tried leaving them and only casting to string(or actually calling `.compressed` as on this PR) when required and see when it kept breaking and got to the port where the minion does start but:
```
11:28:33,187 [salt.minion                                    :1780][WARNING ][ProcessPayload(27042)] [salt-minion/minion] Passed invalid arguments to saltutil.sync_all: can not serialize 'IPv4Address' object 
```

s0undt3ch:
Test failures are valid. Tests need to be updated.

s0undt3ch (approved):

s0undt3ch:
2018.3.4 PR https://github.com/saltstack/salt/pull/51290
This code does not exist on 2019.2.0.rc1

Issue #51248, opened 2019-01-19
"[WARNING ] The file_roots parameter is not properly formatted, using defaults"


johnnybubonic:
### Description of Issue/Question
I keep getting `[WARNING ] The file_roots parameter is not properly formatted, using defaults` in the systemd journal on my master and I can't seem to figure out where it's coming from/why it's happening...

What's weird is I get it even when I comment out the file_roots i have in there (as shown below).

### Setup

```YAML
interface: '::'
ipv6: True
show_jid: True
strip_colors: False
cli_summary: True
keysize: 4096
sign_pub_messages: True
ssl:
  keyfile: /etc/letsencrypt/live/[REDACTED]/privkey.pem
  certfile: /etc/letsencrypt/live/[REDACTED]/fullchain.pem
  ssl_version: PROTOCOL_TLSv1_2
failhard: True
hash_type: sha512
fileserver_backend:
  - roots
  - git
pillar_roots:
  - /srv/pillar
```
I've confirmed it parses to valid YAML with loading it via PyYAML and writing out to a file with pyaml:

```python
import yaml
import pyaml

with open('/etc/salt/master', 'r') as f:
    c = f.read()

y = yaml.load(c)

with open('/root/saltconf.yml', 'wb') as f:
    f.write(pyaml.dumps(y))
```
which yields:

```YAML
cli_summary: true
failhard: true
fileserver_backend:
  - roots
  - git
hash_type: sha512
interface: '::'
ipv6: true
keysize: 4096
pillar_roots:
  - /srv/pillar
show_jid: true
sign_pub_messages: true
ssl:
  certfile: /etc/letsencrypt/live/[REDACTED]/fullchain.pem
  keyfile: /etc/letsencrypt/live/[REDACTED]/privkey.pem
  ssl_version: PROTOCOL_TLSv1_2
strip_colors: false
```

### Steps to Reproduce Issue
See above.

### Versions Report

```YAML
Salt Version:
           Salt: 2018.3.3
 
Dependency Versions:
           cffi: 1.11.5
       cherrypy: unknown
       dateutil: Not Installed
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.8
        libgit2: 0.26.8
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.5.6
   mysql-python: Not Installed
[...]

PR #51278, opened 2019-01-22
Warn about proper validation option

dwoz:
### What does this PR do?

When normalizing `pillar_roots` if the option is not a dictionary, warn about `pillar_roots` and fallback to its default.

#51248 

### Previous Behavior

When validating `pillar_roots` we warned about `file_roots` and used the `file_roots` fallback.

### New Behavior

Warn about the proper option and use proper fallback for `pillar_roots`

### Tests written?

Yes

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/config/__init__.py
+++ b/salt/config/__init__.py
@@ -1938,15 +1938,10 @@ PROVIDER_CONFIG_DEFAULTS = {
 # <---- Salt Cloud Configuration Defaults ------------------------------------
 
 
-def _validate_file_roots(file_roots):
+def _normalize_roots(file_roots):
     '''
-    If the file_roots option has a key that is None then we will error out,
-    just replace it with an empty list
+    Normalize file or pillar roots.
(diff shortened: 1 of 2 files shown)


salt/config/__init__.py
```diff
@@ -1958,6 +1953,30 @@ def _validate_file_roots(file_roots):
     return file_roots
 
 
+def _validate_pillar_roots(pillar_roots):
```
waynew: Does something magically call this function? I'm not seeing a change to call `_validate_pillar_roots`, so just mildly confused 😛

dwoz: @waynew Good catch, fixed.

waynew: Glad I'm not crazy 😀

waynew (approved):
👍

thatch45 (approved):

PR #51325, opened 2019-01-25
Add support for timeout option in pillar file for junos proxy

rsmekala:
### What does this PR do?
Supports timeout option in pillar files, useful for slow devices

https://github.com/Juniper/salt/issues/85

### Tests written?

No


salt/proxy/junos.py
```diff
@@ -113,6 +113,15 @@ def init(opts):
         thisproxy['initialized'] = False
         return
 
+    if 'timeout' in proxy_keys:
+        timeout = int(opts['proxy']['timeout'])
+        try:
+            thisproxy['conn'].timeout = timeout
+        except Exception as ex:
+            log.error('Not able to set timeout due to: %s' % ex)
```
s0undt3ch:
```python
log.error('Not able to set timeout due to: %s', ex)
```

The same for the other log call.

s0undt3ch (changes requested):

rsmekala:
@s0undt3ch I have resolved the lint errors. But CI jobs are still failing due to some unrelated error. Could you please look into it ??

salt/proxy/junos.py
```diff
@@ -113,6 +113,15 @@ def init(opts):
         thisproxy['initialized'] = False
         return
 
+    if 'timeout' in proxy_keys:
+        timeout = int(opts['proxy']['timeout'])
+        try:
+            thisproxy['conn'].timeout = timeout
+        except Exception as ex:
+            log.error('Not able to set timeout due to: {}'.format(ex))
```
s0undt3ch: Sorry, no formatting on log calls either.
```
log.error('Not able to set timeout due to: %s', ex)
```

salt/proxy/junos.py
```diff
@@ -113,6 +113,15 @@ def init(opts):
...
+        else:
+            log.debug('RPC timeout set to {} seconds'.format(timeout))
```
s0undt3ch: Same as above:
```
log.debug('RPC timeout set to %d seconds', timeout)
```

s0undt3ch (changes requested):

salt/proxy/junos.py
rsmekala: Do you mean `format` function call cannot be used in logging ??

s0undt3ch: Yes.
It can, of course, but it shouldn't for performance reasons.

s0undt3ch: See the fixed log calls on my commends.

s0undt3ch (approved):

sagetherage:
already in master

Issue #6922, opened 2013-08-27
Trouble with network.system


smithjm:
I'm trying to use the network module on a Fedora 17 host with a very simple config, using salt 0.16.3 on both the master and minion.  The interface applies correctly, but the network.system fails.

/srv/base/net/init.sls:

```
system:
  network.system:
    - enabled: true
    - hostname: myhost.someplace.com
    - gateway: 10.198.19.254
    - gatewaydev: eth0
    - nozeroconf: true
    - requires_reboot: true

eth0:
  network.managed:
    - enabled: true
    - type: eth
    - proto: none
    - ipaddr: 10.198.16.45
    - netmask:  255.255.252.0
    - dns:
      - 10.192.0.1
      - 10.209.0.233
```

This seems to match the online documentation pretty precisely, however, when I try to apply the state to the host, I get the following error:

```
myhost.someplace.com:

State: - network
Name:      eth0
Function:  managed
    Result:    True
    Comment:   Interface eth0 is up to date.
    Changes:   

State: - network
Name:      system
Function:  system
    Result:    False
    Comment:   An exception occurred in this state: Traceback (most recent call last):
  File "/usr/lib/python2.7/site-packages/salt/state.py", line 1237, in call
    *cdata['args'], **cdata['kwargs'])
  File "/usr/lib/python2.7/site-packages/salt/states/network.py", line 350, in system
    new = __salt__['ip.build_network_settings'](**kwargs)
  File "/usr/lib/python2.7/site-packages/salt/modules/rh_ip.py", line 1017, in build_network_settings
    opts = _parse_network_settings(settings, current_network_settings)
  File "/usr/lib/python2.7/site-packages/salt/modules/rh_ip.py", line 682, in _parse_network_settings
    _raise_error_network('networking', valid)
  File "/usr/lib/python2.7/site-packages/salt/modules/rh_ip.py", line 732, in _raise_error_network
    msg = _error_msg_network(option, expected)
  File "/usr/lib/python2.7/site-packages/salt/modules/rh_ip.py", line 98, in _error_msg_network
    return msg.format(option, '|'.join(expected))
TypeError: sequence item 4: expected string, bool found

    Changes:   
```

The-Loeki:
Replicatable on Ubuntu 14.04 w/ salt 2015.5 and no `/etc/default/networking` file; 
Results:
```
          ID: system-network
    Function: network.system
      Result: False
     Comment: An exception occurred in this state: Traceback (most recent call last):
[...]

PR #51343, opened 2019-01-25
6922 fix network error errors

waynew:
### What does this PR do?

Changes the error message producing code to actually produce the expected behavior (i.e. provide a useful error message).

It would be nice to refactor the debian_ip and rh_ip states, they have a lot of shared code, but I think it's probably fine for now to just make this fix. If I have to touch these modules again, though, I'll probably take the time to refactor out all the shared code 😛 

#6922

### Previous Behavior

Bad options produced a stack trace, whoops.

### New Behavior

Bad options provide a nice helpful comment, e.g.:

```
  Comment: Invalid network setting -- Setting: networking, Expected: [yes|on|true|1|True|no|off|false|0|False]
```
### Tests written?

Yes

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/modules/debian_ip.py
+++ b/salt/modules/debian_ip.py
@@ -158,7 +158,7 @@ def _error_msg_iface(iface, option, expected):
     a list of expected values.
     '''
     msg = 'Invalid option -- Interface: {0}, Option: {1}, Expected: [{2}]'
-    return msg.format(iface, option, '|'.join(expected))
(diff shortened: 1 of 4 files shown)


dwoz (approved):

crlane:
Will this fix be in the next release? It does not appear to be in 2019.2.2. Any idea when that might be?

waynew:
@crlane this should show up in the Neon release - we've already got an RC out that you can try!

PR #51365, opened 2019-01-28
Add Python2 depreciation warning

waynew:
### What does this PR do?

Adds a depreciation warning when anything `salt` runs. I'm not sure if it's better to add it here, or somewhere else. The warning comes out like this:

```
 DeprecationWarning: Python 2.7 will reach the end of its life on January 1st, 2020. Please upgrade your Python as Python 2.7 won't be maintained after that date.  Salt will drop support for Python 2.7 in the Sodium release.
```

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/__init__.py
+++ b/salt/__init__.py
@@ -95,17 +95,3 @@ __define_global_system_encoding_variable__()
 
 # This is now garbage collectable
 del __define_global_system_encoding_variable__
-
-
-# REMOVEME after Python 2.7 support is dropped
-import salt.ext.six as six
-
-if six.PY2:
-    from salt.utils.versions import warn_until
-    # Message borrowed from pip's deprecation warning
-    warn_until('Sodium',
-               'Python 2.7 will reach the end of its life on January 1st, 2020.'
-               ' Please upgrade your Python as Python 2.7 won\'t be maintained'
-               ' after that date.  Salt will drop support for Python 2.7 in the'
-               ' Sodium release.')
-# END REMOVEME
--- a/salt/__init__.py
+++ b/salt/__init__.py
@@ -29,6 +29,7 @@ warnings.filterwarnings(
     UserWarning
 )
 
+
 def __define_global_system_encoding_variable__():
     import sys
     # This is the most trustworthy source of the system encoding, though, if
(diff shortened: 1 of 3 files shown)


Ch3LL:
@waynew some lint failures here. I dont think the other failures are related.

thatch45:
@waynew lets put it in salt/scripts so that it only calls the deprecation warning when you start the minion or master. That way we can avoid it happening for every command and make it happen at the startup of the most critical procs.
Also, change the wording to "No the sodium release or later"

waynew:
So just master & minion? Not salt-call/salt-ssh/salt-cloud, etc.?

thatch45:
No, I think that would just be too noisy, it is a hard balance between notification and noise.

thatch45:
And we can always put it int he release notes and up the notifications in Neon if needs be

s0undt3ch (approved):
This will trigger the warning even for when salt is used as a library(and not on the CLI).

This is the right place.

thatch45:
I am just worried, this will be CRAZY noisy, I am pushing back for just in master and minion startup

s0undt3ch:
Heh, I missed the previous conversation.
Indeed it will be too noisy.
We could filter it to only show once, but again, once per process.
So Tom has a point, though why not all CLIs? Still too noisy?

s0undt3ch:
Hmm. Master and Minion. If we end up feeling like its not getting enough attention, extend to all CLIs....

s0undt3ch (approved):

waynew:
I'm good with putting it in just for salt-master/minion now. Maybe next release add it to the other CLIs, or do some kind of like... drop a timestamp file in `/tmp` or something so it only bugs them every 5 minutes?

I think once we hit the Sodium release, though, we should go ahead and make it extremely noisy. If you're still interested in the latest and greatest salt but you don't want to get a supported version of Python.... 🤷‍♂️ 

Besides, if it *really* bothers someone they can just add a salt `file.managed` state to remove the warning :trollface:

thatch45:
as long as it is not too noisy now. Lets cross the Neon/Sodium bridges later :)

thatch45 (approved):

Ch3LL:
test failures not related but are known and being investigated outside this PR:
  - known gem test failure - currently assigned out for a fix
  - windows tests timing out- currently assigned out for a fix

Issue #51128, opened 2019-01-09
Windows Grains doesn't include default gateway


doesitblend:
### Description of Issue/Question
A comparison between Linux and Windows grains in 2018.3.3
### Setup
Install Salt 2018.3.3 on Windows 2016 server

### Steps to Reproduce Issue
```
[root@winmaster vagrant]# salt '*' grains.item ip4_gw ip_gw ip6_gw
winmaster: # A RHEL Based host and my master 
    ----------
    ip4_gw:
        10.0.2.2
    ip6_gw:
        False
    ip_gw:
        True
win2016-2:
    ----------
    ip4_gw:
    ip6_gw:
    ip_gw:
win2016-3:
    ----------
    ip4_gw:
    ip6_gw:
    ip_gw:
```

### Versions Report

```
win2016-2:
    Salt Version:
               Salt: 2018.3.3
     
    Dependency Versions:
               cffi: 1.10.0
           cherrypy: 10.2.1
           dateutil: 2.6.1
          docker-py: Not Installed
              gitdb: 2.0.5
          gitpython: 2.1.3
              ioflo: Not Installed
             Jinja2: 2.9.6
            libgit2: Not Installed
            libnacl: 1.6.1
           M2Crypto: Not Installed
               Mako: 1.0.6
       msgpack-pure: Not Installed
     msgpack-python: 0.4.8
       mysql-python: Not Installed
          pycparser: 2.17
           pycrypto: 2.6.1
       pycryptodome: Not Installed
             pygit2: Not Installed
             Python: 2.7.14 (v2.7.14:84471935ed, Sep 16 2017, 20:25:58) [MSC v.1500 64 bit (AMD64)]
       python-gnupg: 0.4.1
             PyYAML: 3.12
              PyZMQ: 16.0.3
               RAET: Not Installed
              smmap: 2.0.5
            timelib: 0.2.4
            Tornado: 4.5.1
                ZMQ: 4.1.6
     
    System Versions:
               dist:   
             locale: cp1252
            machine: AMD64
            release: 2016Server
             system: Windows
            version: 2016Server 10.0.14393  Multiprocessor Free
     
win2016-3:
    Salt Version:
               Salt: 2018.3.3
     
    Dependency Versions:
               cffi: 1.10.0
           cherrypy: 10.2.1
           dateutil: 2.6.1
          docker-py: Not Installed
              gitdb: 2.0.5
          gitpython: 2.1.3
              ioflo: Not Installed
[...]

PR #51372, opened 2019-01-28
Use .NET to gather NIC information on newer systems

twangboy:
### What does this PR do?
Uses .NET to gather Network Interface Information. Maintains old code that uses WMI to gather that information for use on Windows 7/2008R2 which has .NET 3.5 which is incompatible with the `pythonnet` library. Though you can install .NET 4 on Windows 7, it's not there out of the box.

This should fix some issues with slowness in Windows as this code is used in generating grains. Where the average run for doing so is 3.5 seconds with WMI, it is ~1 second with .NET.

Also adds the gateway in the return for Windows, which is present on other systems:
https://github.com/saltstack/salt/issues/51128

### Tests written?
Not yet, hence the WIP

### Commits signed with GPG?
Yes


dwoz (approved):

s0undt3ch:
This actually broke develop under Linux
https://github.com/saltstack/salt/pull/53362/commits/fded9da495a3d891294455b223e818471a7c35dc

aplanas:
yep: https://github.com/saltstack/salt/pull/53381

aplanas:
@s0undt3ch where is the PR where this commit is living?

s0undt3ch:
https://github.com/saltstack/salt/pull/53362

s0undt3ch:
This also brings in another python dep `pythonnet`. Please confirm @twangboy

twangboy:
Yes... it was added to the `salt/pkg/windows/req.txt`.  I was unaware of the additional requirement to use the static files.

Issue #51351, opened 2019-01-26
Exception using smartos.vm_present to create docker zone


garethhowell:
### Description of Issue/Question
Getting an exception when trying to create docker version on plexmediaserver on smartos server.

### Setup
(Please provide relevant configs and/or SLS files (Be sure to remove sensitive info).)
**smartos_docker_containers/plex3.sls**
```
include:
  - smartos_docker_container
	
plex3.agdon.net:
  smartos.vm_present:
    - config:
        auto_import: true
        reprovision: true
    - vmconfig:
        image_uuid: plexinc/pms-docker:plexpass
        brand: lx
        alias: plex3
        quota: 5
        max_physical_memory: 1024
        tags:
          label: 'plex3 docker'
          owner: 'garethhowell'
        resolvers:
          - 172.29.12.7
        nics:
          "82:1b:8e:49:e9:19":
            nic_tag: admin
            mtu: 1500
            ips:
              - 172.29.12.44/24
        filesystems:
          "/config":
            source: "/data/config"
            type: lofs
            options:
              - nodevices
          "/data":
            source: "/data/media"
            type: lofs
            options:
              - nodevices
          "/transcode":
            source: "/data/transcode"
            type: lofs
            options:
              - nodevices
```

**smartos_docker_container/init.sls**

```
https://docker.io:
  smartos.source_present:
    - source_type: docker
```

### Steps to Reproduce Issue
(Include debug logs if possible and relevant.)

```
sudo salt global_deneb state.apply smartos_docker_container.plex3
global_deneb:
----------
          ID: https://docker.io
    Function: smartos.source_present
      Result: True
     Comment: image source https://docker.io is present
     Started: 16:08:29.276973
    Duration: 493.941 ms
     Changes:   
----------
          ID: plex3.agdon.net
    Function: smartos.vm_present
      Result: False
     Comment: An exception occurred in this state: Traceback (most recent call last):
                File "/opt/tools/lib/python2.7/site-packages/salt/state.py", line 1951, in call
                  ret = self.states[cdata['full']](*cdata['args'], **cdata['kwargs'])
                File "/opt/tools/lib/python2.7/site-packages/salt/loader.py", line 2033, in wrapper
                  return f(*args, **kwargs)
                File "/opt/tools/lib/python2.7/site-packages/salt/states/smartos.py", line 801, in vm_present
                  if vmconfig['image_uuid'] not in __salt__['imgadm.list']():
[...]

PR #51383, opened 2019-01-29
imgadm module should deal with invalid images better

sjorge:
### What does this PR do?
images should always have a manifest and name tag, but it looks like it is possible to encouter some in the wild that do not have those fields. This PR ensure we do not simple hard fail on those and throw and Exception.

#51351
#51503 (works for me with this applied)

### Previous Behavior
Exception was thrown

### New Behavior
We return `None` ... well technically
{uuid-here: None}

### Tests written?
No

### Commits signed with GPG?
No

### DIFF
--- a/salt/modules/smartos_imgadm.py
+++ b/salt/modules/smartos_imgadm.py
@@ -56,7 +56,7 @@ def _parse_image_meta(image=None, detail=False):
 
     if image and 'Error' in image:
         ret = image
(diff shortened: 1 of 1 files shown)


sjorge:
Might be useful to get this into 2019.3.1 or maybe even .0

Ch3LL (approved):

PR #51389, opened 2019-01-29
Backport proxy fixes to 2018.3

terminalmage:
These issues were discovered while working on PR #51204, and would affect either Salt, the test suite, or both, when the proxy service is running.


s0undt3ch (approved):

s0undt3ch:
The lint error seems serious.

PR #51409, opened 2019-01-30
Add root and no_recommends parameters in the public interface for Zypper and RPM (lowpkg)

aplanas:
### What does this PR do?

New feature

### Tests written?
For the low level RPM yes, for the zypper level I adjusted the current tests.

(backport from #50125)
(backport from #50671)


dwoz:

PR #51412, opened 2019-01-30
cmd: Add root parameter for wait and run states

aplanas:
`run_chroot` module.

(cherry picked from commit cd7b5b2ea81204cea70e73c6db8d270b730b622f)


thatch45:

PR #51413, opened 2019-01-30
systemd: add optional root parameter

aplanas:
With systemctl is possible for certain actions (enable, disable,
mask, and others) indicate a different root directory.  If this
parameter is specified, systemctl will do the work independently
of systemd service.

This patch add the optional root parameter to those actions.

(cherry picked from commit 786688f4d56bb567ed29f5764ff0c2d8e0995176)


thatch45:

PR #51415, opened 2019-01-30
chroot: add new chroot module

aplanas:
Add a new chroot module in saltstack.

Based on dockermod.call, this module implement an way to execute random salt modules inside a chroot environment.

Tests are included.

(cherry picked from commit 984fb62f5df60b5d764788b2833387f04cf7b2cb)

### DIFF
--- a/salt/modules/chroot.py
+++ b/salt/modules/chroot.py
@@ -1,0 +1,165 @@
+# -*- coding: utf-8 -*-
+#
+# Author: Alberto Planas <aplanas@suse.com>
+#
+# Copyright 2018 SUSE LINUX GmbH, Nuernberg, Germany.
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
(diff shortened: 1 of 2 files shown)


dwoz (changes requested):
@aplanas 2019.2 has been feature frozen for weeks now.

aplanas:
@dwoz I know : (((. I was wrongly expecting that my patches to the develop branch from Sept - December will appear in the flourine codebase, but this was not the case.

* Where I can land those pull request that comes from develop, to be sure that they will appear in the next version?

* For the next time, do I need to generate two pull request every time? One for develop and another for the branch that will become the next product?

* There is a huge difference of commits between develop and fluorine, some of them maybe fixing stuff. Is there any janitor process that will backport them to the product branch, for cases like mine where the correct process was not understood / followed?

thatch45:
One of the great things about salt though, is that you can ship these on your install even if it is older by putting these modules into `_modules` directories.

PR #51416, opened 2019-01-30
Add new module freezer

aplanas:

(cherry picked from commit efcce01a78f77c1d4b4f0adf6daa98b8925380e8)


dwoz:

PR #51417, opened 2019-01-30
parted: support variable length output for print

aplanas:
The command print from parted have a variable lenght output,
depending on the units requested, or the kind of partition.

This patch add the logic to support the full range of outputs
that parted 3.1, and early versions, are generating.

(cherry picked from commit 5adf859e3f1c78a34886c065d8540947a7718120)


aplanas:
Uhm the tests are expecting https://github.com/saltstack/salt/pull/51408 to be there. I think that I will add this commit to the other pull request and I will close this, so we can have a green test.

PR #51422, opened 2019-01-30
mount: add fstab_{present,absent} states

aplanas:
### What does this PR do?

Add two new states in mount:

### Tests written?

Yes

(cherry picked from commit 12ae2db12264f8632ccdd1231450699648f4042a)


dwoz:

PR #51424, opened 2019-01-30
disk: support setting FAT size for format_

aplanas:
### What does this PR do?

Add new optional parameter `fat` in disk.format_ to specify the FAT size (12, 16 or 32).

### Tests written?

No, current one should be OK

(cherry picked from commit 5828aa1730022ce7d46a53507dade94dde44fa67)

### DIFF
--- a/salt/modules/disk.py
+++ b/salt/modules/disk.py
@@ -422,6 +422,7 @@ def format_(device,
             fs_type='ext4',
             inode_size=None,
             lazy_itable_init=None,
+            fat=None,
             force=False):
     '''
     Format a filesystem onto a device
@@ -449,6 +450,10 @@ def format_(device,
 
         This option is only enabled for ext filesystems
 
+    fat
+        FAT size option. Can be 12, 16 or 32, and can only be used on
+        fat or vfat filesystems.
+
     force
         Force mke2fs to create a filesystem, even if the specified device is
         not a partition on a block special device. This option is only enabled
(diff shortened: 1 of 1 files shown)


salt/modules/disk.py
```diff
@@ -422,6 +422,7 @@ def format_(device,
             fs_type='ext4',
             inode_size=None,
             lazy_itable_init=None,
+            fat=None,
```
rares-pop: do we really want this new parameter? What about using fs_type for fat, as in fat12, 16, 32?

dwoz:

salt/modules/disk.py
aplanas: We do need it, as mkfs is a frontend for the specific executable mkfs.fs_type. There is not mkfs.fat12, mkfs.fat16 nor mkfs.fat32 executables, only mkfs.fat of mkfs.vfat. They use the -F parameter to indicate the size.

rares-pop: I think it's a wrong route to take, adding each different specific mkfs parameters here. The point of this function is to unify those calls, so probably we should have an abstraction parameter for this, where on ext to mean inode_size and on fat/vfat to mean fat-size. I'm not saying to reuse this particular inode_size, but to avoid adding new parameters for very specific calls.

At worst, we should call it fat-size, as their man page describe it - https://linux.die.net/man/8/mkfs.vfat

PR #51427, opened 2019-01-30
Allow loading executor modules from the fileserver

max-arnold:
### What does this PR do?

Allow loading executor modules from the `_executors` fileserver directory.  This allows overloading standard executors and also adding new ones to wrap state execution calls globally.

### Tests written?

Yes

### Commits signed with GPG?

No

### DIFF
--- a/doc/topics/development/modules/index.rst
+++ b/doc/topics/development/modules/index.rst
@@ -73,7 +73,7 @@ Sync Via the saltutil Module
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 The saltutil module has a number of functions that can be used to sync all
-or specific dynamic modules. The ``saltutil.sync_*`` 
+or specific dynamic modules. The ``saltutil.sync_*``
 :py:mod:`execution functions <salt.modules.saltutil>` and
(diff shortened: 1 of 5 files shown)


max-arnold:
The pylint warning is caused by reserved `async` keyword which was added in #50168

max-arnold:
Waiting several months for the Neon release seems too long. Is it possible to get this merged into 2019.2.1?  If so, I can rebase against 2019.2

gtmanfred:
@saltstack/team-core can yall take a look at this?

thatch45:
I will approve this for Neon, lets discuss moving it to 2019.2.1 after the release of 2019.2.0, but we should avoid feature additions to point releases, it really bites us when we do that

thatch45 (approved):
This looks good, just as it should!

terminalmage (approved):

dwoz:
@max-arnold Please have a look at the linter, it is complaining about `async` or `await` being used in this file.

max-arnold:
@dwoz Already did, see my second comment.

If you want, I can fix it in a separate commit.

PR #51452, opened 2019-01-31
Backport #50963 to 2017.7

terminalmage:
Fixes https://github.com/saltstack/salt-jenkins/issues/1270

### DIFF
--- a/salt/output/nested.py
+++ b/salt/output/nested.py
@@ -34,6 +34,11 @@ from salt.ext.six import string_types
 from salt.utils import get_colors
 import salt.utils.locales
 
+try:
+    from collections.abc import Mapping
+except ImportError:
+    from collections import Mapping
+
 
 class NestDisplay(object):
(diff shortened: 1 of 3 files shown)


salt/payload.py
```diff
@@ -172,6 +173,33 @@ def dumps(self, msg, use_bin_type=False):
                              Since this changes the wire protocol, this
                              option should not be used outside of IPC.
         '''
+<<<<<<< HEAD
```
dwoz: Looks like a merge wart.

dwoz (changes requested):

salt/payload.py
terminalmage: Oh yeah, thought I removed both sides of that, whoops. Fixing.

terminalmage: Fixed and rebased.

dwoz (approved):

PR #51454, opened 2019-01-31
Fix deprecation warning in nested outputter

terminalmage:
Pull request #50963 introduces a deprecation warning on later Python 3 releases, as the abstract base classes now live in `collections.abc`. This fixes that deprecation warning.


terminalmage:
Docs failure is an issue in Jenkins, and this PR didn't touch docstrings anyway. This should be good to merge pending review.

Issue #7424, opened 2013-09-24
feature: salt tls module ssl check


aboe76:
I would like to see an extension to the tls module with:
- certificate expiration date
- validity check

something allong the lines of:
https://gist.github.com/crashdump/5683952

that way we can monitor https/ssl certificate from within salt.

QuinnyPig:
Where is this sitting today? It would be awesome to have nodes able to check cert expiry on their neighbors / ELBs...

terminalmage:
I checked and we already had code in the module that was getting the cert expiration date, so I abstracted that to a helper function and then used that helper function to expose this functionality in a new function `tls.get_expiration_date`. The pull request that adds this new function is https://github.com/saltstack/salt/pull/46922.

As for validation, I don't know pyOpenSSL well enough to do that, would someone else like to tackle it?

PR #51461, opened 2019-02-01
`modules.tls.validate` is now a thing

waynew:
### What does this PR do?

Add 'validate' to the TLS module, and fixes a couple of bugs

#7424

### Previous Behavior

I still need to double check to see how far back it was, but at some point the `digest` argument was required for a couple of features, but we weren't providing it. I've added those. There were also some encoding (text/unicode vs bytes) problems that popped up when I was running it under Python3, whoops. Running those functions would fail.

### New Behavior

Failing functions no longer fail.

`validate` now exists - one can pass a cert, a CA, and a CRL to confirm that a certificate is valid.

### Tests written?

Yes - including some integration tests to check that we're using pyOpenSSL properly.

### Commits signed with GPG?

Yes

### Things to confirm

Before merging this I do need to check when pyOpenSSL introduced the behavior that has our module broken - I don't want my changes to break anything if that was a reasonably new (<1 year?) change.

Also if there's any code style changes that I should make, please point that out.

Oh, also had a couple of ❓ 
```
# TODO: Should we skip on Windows? Or skip if PyOpenSSL is not installed? -W. Werner, 2019-01-30
# TODO: Should any of these tests be marked destructive? -W. Werner, 2019-01-30
```
I wasn't sure what kind of changes were considered `destructive` - these aren't adding or removing users or anything, but it *is* creating and mucking with a CA. I'm guessing that it's entirely fine though?

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -1,13 +1,21 @@
-languages:
-  Ruby: false
-  JavaScript: false
-  Python: true
-  PHP: false
+version: "2"
 
-engines:
+exclude_patterns:
+  - "*.js"
+  - "*.rb"
+  - "*.php"
+
+plugins:
   radon:
     enabled: true
-    exclude_paths:
+    exclude_patterns:
       - "templates/"
     config:
       threshold: "D"
+
+checks:
+  argument-count:
+    enabled: false
+  similar-code:
+    config:
+      threshold: 40
--- a/salt/modules/tls.py
+++ b/salt/modules/tls.py
@@ -563,6 +563,52 @@ def _read_cert(cert):
             return cert
 
 
+def validate(cert, ca_name, crl_file):
+    '''
+    .. versionadded:: Neon
+
+    Validate a certificate against a given CA/CRL.
+
+    cert
+        path to the certifiate PEM file or string
+
+    ca_name
+        name of the CA
+
+    crl_file
+        full path to the CRL file
+    '''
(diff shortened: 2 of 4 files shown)


dwoz (changes requested):
@waynew Can you look at the linter issues?

waynew:
Okay, I went back and apparently [digest has been around for a very, *very* long time](https://github.com/pyca/pyopenssl/blame/8654f785ccd96a0d6cfb33d521dd891a9178380d/OpenSSL/crypto.py#L1838). Like, it was first introduced in [2014](https://github.com/pyca/pyopenssl/commit/2923dc0f475e0d1a7893c7ca264c95c75f8ba9dd) for CRL. At *least* 5 years ago [the lib expected bytes for the serial](https://github.com/pyca/pyopenssl/blame/4d57590bc7fc93430a1fdacc31bc0cbd9778f678/src/OpenSSL/crypto.py#L1947) and no less than 6 for [the time](https://github.com/pyca/pyopenssl/blame/4d57590bc7fc93430a1fdacc31bc0cbd9778f678/src/OpenSSL/crypto.py#L143).

So, I'm totally comfortable with the fact that there aren't any suitably recent versions of pyOpenSSL that expect anything besides what we're providing here 👍 

@dwoz looks like I fixed the lint issues a while back, too :trollface:

dwoz (approved):

PR #51464, opened 2019-02-01
Allow to override git_pillar_base per-remote

sathieu:
### What does this PR do?

Allow to override git_pillar_base per-remote

None.

### New Behavior
Example:
```yaml

git_pillar_base: develop

ext_pillar:
  - git:
    - __env__ https://gitserver/git-pillar.git
    - __env__ https://gitserver/git-pillar2.git:
      - base: master
```

### Tests written?

No

### Commits signed with GPG?

Not yet

### DIFF
--- a/salt/pillar/git_pillar.py
+++ b/salt/pillar/git_pillar.py
@@ -368,9 +368,9 @@ from salt.pillar import Pillar
 # Import third party libs
 from salt.ext import six
 
-PER_REMOTE_OVERRIDES = ('env', 'root', 'ssl_verify', 'refspecs')
+PER_REMOTE_OVERRIDES = ('base', 'env', 'root', 'ssl_verify', 'refspecs')
 PER_REMOTE_ONLY = ('name', 'mountpoint', 'all_saltenvs')
-GLOBAL_ONLY = ('base', 'branch')
+GLOBAL_ONLY = ('branch',)
 
 # Set up logging
 log = logging.getLogger(__name__)


sathieu:
Yes. This is a feature.

Can we have any visibility on the release schedule?

terminalmage:
Can you please help me understand the purpose of this? git_pillar doesn't work like gitfs.

sathieu:
I know, and it's annoying. But with `__env__` it starts to behave similarly.

Our use case: we have different projects (I.e apps) some have develop, qa,staging and prod branches which corresponds to isolated salt master's `git_pillar_base`. But some projects (we call them infra, like DNS) have less branches (develop and master) and every salt-masters of every zones should point to the master branch.

sathieu:
Also, we don't configure saltenv and pillarenv on minions, but we can use them on CLI to test feature branches.

terminalmage (changes requested):
This needs to be changed in the following two ways:

1. It must be opened against `develop`, not `2018.3`.
2. The dunder init for `Pygit2` and `GitPython` provider objects must be modified to log and run `failhard()` when `base` is used in instances where `self.branch != '__env__'`.

dwoz:
@sathieu We are currently working to stay on a 6 month release cadence. Historically we've had trouble hitting this mark but we're actively working on ways to improve in this area.

I'm going to close this PR, you can cherry-pick the change and make a PR to develop.

sathieu:
Replaced by #51640.

PR #51469, opened 2019-02-01
Don't prevent `get_locale` from working when running under Suse docker containers

s0undt3ch:
### What does this PR do?
Don't prevent `get_locale` from working when running under Suse docker containers

None?

### Previous Behavior
When running under docker, even though the system has the required information, salt was prevented from getting it.

### New Behavior
Now also works under Docker containers

### Tests written?

No

### Commits signed with GPG?

Yes


dwoz:
@isbm Ping! Can you take a look and review this one?

isbm (approved):
This seems correct (as of today). On newer systems this will be applied with different tools, while on SLE11 and 12 will work fine.

PR #51470, opened 2019-02-01
Do not treat MSI auth support as hard dependency

jgleissner:
### What does this PR do?
On systems that lack MSI authentication support, only fail if MSI is actually used.

It allows users who are on an older version of msrestazure (i.e. < 0.4.14) to use the azurearm back-end, as long as they do not use MSI authentication.

### Commits signed with GPG?

Yes

### Notes
I had to change the `_call` function of the `communicator` class to catch `BaseException` instead of just `Exception` for this to work. This is because if a `SaltSystemExit` exception is thrown in a worker thread, the `pool.map` call hangs. The other `SaltSystemExit` exception in `get_client` was potentially affected too, but AFAICS that code path could really only be triggered by a bug.

### DIFF
--- a/salt/cloud/__init__.py
+++ b/salt/cloud/__init__.py
@@ -71,10 +71,16 @@ def communicator(func):
             queue.put('KEYBOARDINT')
             queue.put('Keyboard interrupt')
             queue.put('{0}\n{1}\n'.format(ex, trace))
-        except BaseException as ex:
+        except Exception as ex:
             trace = traceback.format_exc()
             queue.put('ERROR')
             queue.put('Exception')
+            queue.put('{0}\n{1}\n'.format(ex, trace))
+        except SystemExit as ex:
+            print("Communicatior: caught exception")
+            trace = traceback.format_exc()
+            queue.put('ERROR')
+            queue.put('System exit')
             queue.put('{0}\n{1}\n'.format(ex, trace))
         return ret
     return _call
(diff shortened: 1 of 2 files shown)


dwoz:
@jgleissner Looks like the linter has one failing check:

https://jenkinsci.saltstack.com/job/pr-lint/job/PR-51470/1/warnings52Result/new/

jgleissner:
Hi Daniel,

The pylint warning was about catching BaseException, which it deems an exception too generic to catch. It was somewhat on purpose to catch generically as not catching a BaseException (which SystemExit is derived of) at this point causes the pool.map call to hang. However, I've changed it to catching SystemExit instead, which silenced the warning.

salt/cloud/__init__.py
```diff
@@ -76,6 +76,12 @@ def _call(queue, args, kwargs):
             queue.put('ERROR')
             queue.put('Exception')
             queue.put('{0}\n{1}\n'.format(ex, trace))
+        except SystemExit as ex:
+            print("Communicatior: caught exception")
```
dwoz: Should this be a print or a log?

jgleissner: That was a left-over debug print, sorry about that. Removed.

dwoz (approved):

terminalmage:
Ported to `master` in https://github.com/saltstack/salt/pull/56694

PR #51471, opened 2019-02-01
Fix the failing dns test on Windows

twangboy:
### What does this PR do?
Gets the name of the first interface on the system. Windows network interfaces don't have the same name across Window systems. You can even go as far as naming them whatever you want. The test was failing because the interface name was hard-coded as 'Ethernet'.

found in jenkins

### Tests written?
Yes

### Commits signed with GPG?
Yes


dwoz (approved):

dwoz (approved):

s0undt3ch (approved):
CentOS 7 Failure unrelated.

PR #51473, opened 2019-02-01
Fix grains metadata wrongly recursing on user-data

### LINKED ISSUE #47543: grains metadata.py should handle user-data as a string
The `169.254.169.254` server provides two endpoints meta-data and user-data.
Currently, both are handled by a recursive function `_search()`
user-data, however, is just a text (YAML ?) and should not be 'walked' like meta-data

jleproust:
Fixes #47543
Some providers return user-data as text/plain, so the condition to
return on application/octet-stream fails. Anyway user-data never needs
recursion.

### Tests written?

No

### Commits signed with GPG?

Yes


dwoz:
@jleproust Are you able to write a regression test for this bug?

jleproust:
> @jleproust Are you able to write a regression test for this bug?

Sorry, I have no idea where to begin. I could add one for this issue if you point me to similar tests, but it seems there are no tests at all for the metadata grains module at the moment.

Ch3LL:
@jleproust you can add a new file nad write either a integration or unit tests. You can take a look at `tests/integration/grains` and `tests/unit/grains` for examples.

Also we just branched 2017.7.9 for the last 2017.7 bug fix release. Are you by chance okay getting this into 2018.3 instead?

dwoz:
@jleproust we are no longer accepting PRs to the `2017.7` branch. If this is still an issue, please re-open this PR against the master branch and note which versions are affected.

Issue #50262, opened 2018-10-27
Saltclass: broken list expansion in Fluorine


max-arnold:
### Description of Issue/Question

This used to work in 2018.3.x:

```
pillars:
  groups:
    - group1
    - group2
  users:
    user1:
      groups: ${groups}
```

It seems that #47799 broke list expansions. The example above fails on Fluorine with the following exception:

```
% salt-ssh example.com pillar.items

[ERROR   ] Exception caught loading ext_pillar 'saltclass':
  File "/Users/user/salt/repos/saltstack-repo/salt/pillar/__init__.py", line 951, in ext_pillar
    key)
  File "/Users/user/salt/repos/saltstack-repo/salt/pillar/__init__.py", line 879, in _external_pillar_data
    *val)
  File "/Users/user/salt/repos/saltstack-repo/salt/pillar/saltclass.py", line 64, in ext_pillar
    return sc.get_pillars(minion_id, salt_data)
  File "/Users/user/salt/repos/saltstack-repo/salt/utils/saltclass.py", line 316, in get_pillars
    pillars_dict_expanded = expand_variables(pillars_dict['pillars'], {}, [])
  File "/Users/user/salt/repos/saltstack-repo/salt/utils/saltclass.py", line 176, in expand_variables
    expand_variables(v, b, expanded, path + [six.text_type(k)])
  File "/Users/user/salt/repos/saltstack-repo/salt/utils/saltclass.py", line 176, in expand_variables
    expand_variables(v, b, expanded, path + [six.text_type(k)])
  File "/Users/user/salt/repos/saltstack-repo/salt/utils/saltclass.py", line 186, in expand_variables
    b = find_and_process_re(v, v, k, b, expanded)
  File "/Users/user/salt/repos/saltstack-repo/salt/utils/saltclass.py", line 158, in find_and_process_re
    v_new = v.replace(re_str, v_expanded)

[CRITICAL] Pillar render error: Failed to load ext_pillar saltclass: replace() argument 2 must be str, not list
example.com:
    ----------
    _errors:
        - Failed to load ext_pillar saltclass: replace() argument 2 must be str, not list
    groups:
        - group1
        - group2
    users:
        ----------
        user1:
            ----------
            groups:
                ${groups}
```

rallytime:
@olivier-mauras Would you be able to look into this issue?

rbthomp:
I can't get variable expansion to work at all on 2018.3.3

Pillar:
```
pillars:
  test:
[...]

Issue #49175, opened 2018-08-17
SaltClass duplicate entries in merged list.


rbthomp:
### Description of Issue/Question
In the situation when a list in contained in both a parent class and child class you get multiples of the same value in the list. This is particularly an issue when running things like beacons. You will get duplicate events and will trigger multiple reactors for the same event. 

### Setup
default.yml
```
pillars:
    data:
        - default 1
        - default 2
```
child.yml
```
pillars:
    data:
        - child 1
        - child 2
```
Result pillar:
```
data:
            - default 1
            - default 2
            - child 1
            - child 2
            - child 1
            - child 2
```

Node.yml
```
classes:
   - default
   - child
```

### Steps to Reproduce Issue
Setup saltclass and node.

### Versions Report
```
Salt Version:
           Salt: 2018.3.0

Dependency Versions:
           cffi: 1.5.2
       cherrypy: 3.6.0
       dateutil: 2.4.2
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.8
        libgit2: 0.24.0
        libnacl: Not Installed
       M2Crypto: 0.21.1
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.4.6
   mysql-python: Not Installed
      pycparser: 2.10
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: 0.24.0
         Python: 2.7.13 (default, Jan 11 2017, 10:56:06) [GCC]
   python-gnupg: Not Installed
         PyYAML: 3.12
          PyZMQ: 14.0.0
           RAET: Not Installed
          smmap: Not Installed
        timelib: Not Installed
        Tornado: 4.2.1
            ZMQ: 4.0.4

System Versions:
           dist: SuSE 12 x86_64
         locale: UTF-8
        machine: x86_64
        release: 4.4.140-94.42-default
         system: Linux
        version: SUSE Linux Enterprise Server  12 x86_64
```
CC: @olivier-mauras

mchugh19:
Not stale. Help us @a-a-abramov

twangboy:
Closing this issue due to age and lack of activity. Please test this on version 3006.2 and create a new issue if the problem persists. The new issue template has more information and will allow us to track and reproduce the issue more effectively. Thanks!

Issue #48170, opened 2018-06-17
Multiple issues with Saltclass variable expansion


max-arnold:
### Description of Issue/Question

I spent more than 2 days trying to migrate existing reclass tree to saltclass (introduced in https://github.com/saltstack/salt/pull/42349) and found multiple issues with saltclass variable expansion.

1. Variable expansion doesn't work at all under Python 2.7

The reason is that checking for `str` is not enough, strings can also be `unicode` with 2.7: https://github.com/saltstack/salt/blob/2018.3.1/salt/utils/saltclass.py#L137

This feature definitely needs more unit tests. Jenkins builds for 2.7 can catch issues like this.

2. Not all variables are expanded

Salt 2017.7, reclass:

```
% salt-call --id example.com pillar.data
local:
    ----------
    __reclass__:
        ----------
        applications:
            - consul
        classes:
            - consul
            - consul.agent
            - site.mydc
        environment:
            base
        nodename:
            example.com
    consul:
        ----------
        agent:
            ----------
            args:
                - datacenter=mydc
                - join=1.2.3.4
        cur_version:
            1.0.3
        dc:
            ----------
            name:
                mydc
            servers:
                1.2.3.4
        versions:
            ----------
            1.0.3:
                ----------
                source:
                    https://releases.hashicorp.com/consul/1.0.3/consul_1.0.3_linux_amd64.zip
                source_hash:
                    sha256=4782e4662de8effe49e97c50b1a1233c03c0026881f6c004144cc3b73f446ec5
```

Salt 2018.3.1, saltclass

```
local:
    ----------
    __saltclass__:
        ----------
        classes:
            - consul
            - consul.agent
            - site.mydc
        environment:
        nodename:
            example.com
        states:
            - consul
    consul:
        ----------
        agent:
            ----------
            args:
                - datacenter=${consul:dc:name}
                - join=${consul:dc:servers}
        cur_version:
            1.0.3
        versions:
            ----------
            1.0.3:
                ----------
                source:
                    https://releases.hashicorp.com/consul/1.0.3/consul_1.0.3_linux_amd64.zip
                source_hash:
                    sha256=4782e4662de8effe49e97c50b1a1233c03c0026881f6c004144cc3b73f446ec5
```

Minimal state tree (both for reclass and saltclass) to reproduce: [example.zip](https://github.com/saltstack/salt/files/2109113/example.zip)

3. Multiple variables on a single line probably won't be expanded

Variable parsing code tells me that things like one below won't work:

[...]

PR #51487, opened 2019-02-04
Fix saltclass

a-a-abramov:
### What does this PR do?
Fixes several saltclass pillar expansion issues.
Note that @olivier-mauras 's commit is essential and I had to cherry pick it into my branch from develop. Also it's worth noting that I had to amend half of @awerner 's commit since it introduced #49175.
And since I didn't have much time code quality is not very high to say the least. Module needs to be heavily refactored.

#50262, #49175, #48170 and possibly others

### Previous Behavior
Check issues from previous section for the examples of malfunctions.  

### New Behavior
Everything works as expected.

### Tests written?
No

### Commits signed with GPG?
No

### DIFF
--- a/salt/utils/saltclass.py
+++ b/salt/utils/saltclass.py
@@ -106,8 +106,20 @@ def dict_search_and_replace(d, old, new, expanded):
     for (k, v) in six.iteritems(d):
         if isinstance(v, dict):
             dict_search_and_replace(d[k], old, new, expanded)
+
+        if isinstance(v, list):
+            x = 0
+            for i in v:
+                if isinstance(i, dict):
+                    dict_search_and_replace(v[x], old, new, expanded)
+                if isinstance(i, six.string_types):
+                    if i == old:
+                        v[x] = new
+                x = x + 1
+
         if v == old:
             d[k] = new
+
     return d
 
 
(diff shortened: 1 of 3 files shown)


rbthomp:
@a-a-abramov 

Can you also remove   

```dict_merge(salt_data['__pillar__'], node_dict[minion_id].get('pillars', {}))```

This breaks, pillars added in node files. It also screws with salts ability to merge and order multiple external pillar sources. There are reason we return the pillars, and not dynamically modify them in place. 

If we need the ability to reference pillars from saltclass in saltclass using jinja there is probably a better way, then modify existing pillars.

**Comment by a-a-abramov** [comment]
@rbthomp 

Could you provide a little more detail on this (preferably with an example)? I don't see how this line screws salt's ability to merge and order external sources.

A few days ago I was thinking exactly about referencing saltclass pillars inside saltclass jinja. I believe we definitely should implement this. Absence of this feature greatly reduces saltclass'es flexibility. Say, for example, we have a state for each version of an application - sls/X/vers8, sls/X/vers11, etc. They provide common functions but are deployed differently and can't simply be parametrized with pillar (X:version) without severe refactoring of states. Now imagine if we also had a dependent application Y. X and Y are deployed onto two hosts - one with X/vers8, another - with X/vers11. If we could simply template states by host pillars with jinja this would be simple. We can do this with simple tops, but not this saltclass. 
So I was thinking that @awerner 's idea was actually very good, but in order for it to work traversing algorithm needs to be changed. Good news - I think I solved this problem, bad news - it would take some time to actually code, debug and test new implementation. I'll let y'all know if I manage to do this.

**Comment by rbthomp** [comment]
@a-a-abramov 

The issue I've ran into is any pillar declared in the node file overwrite existing keys instead of merging which is what we expect, except for lists when "^" is used. I've also encountered #48895. I don't have a complete understanding of what creates this condition, but when  ``` dict_merge(salt_data['__pillar__'], node_dict[minion_id].get('pillars', {}))``` is remove merging works as expected. 

I generally try and keep my logic in the states and not in pillars. I'm able to avoid this by using classes, with a couple of exceptions, like the one you described of using an existing state you don't want to modify. 

### Steps to Reproduce Issue
Pillars from class:
```
pillars:
    test:
      - classdata
```

PIllars from node:
```
pillars:
     test:
        - nodedata
```

Will result in:
```
pillars:
    test:
        - nodedata
```

expected result:
```
pillars:
    test:
        - classdata
        - nodedata

```

a-a-abramov:
@rbthomp 
Yeah, you're right, I see it now. Once I have time I'll fix this and probably refactor the whole `utils/saltclass.py` (if someone except us is interested, really).

a-a-abramov:
Changes from this PR are included in #52407

PR #51492, opened 2019-02-04
state diff.py

ProcenaFr:
### What does this PR do?
A smart diff command to compare a file from the master and the minion :

- Checks the permissions of the file.
- Checks the content of the file.
- Removes the multiple spaces and tabulations.
- Removes commentaries before comparing.
- Can modify or not the content of the file if it is different.
- Can modify or not the permissions of the file if it is different.

### Tests written?

Yes

### Commits signed with GPG?

No


mchugh19:
How is this different to file.managed?

ProcenaFr:
It is smarter because it doesn't compare commentaries in files (With "#").
It removes multi spaces/tabulations before comparing (Useful for files like sudoers).

You can indicate if you want or not to change the file if it is different.
You can indicate to just check the permissions.
You don't need to use "test=True" at the end of your line (for file.managed) which could be forgotten and cause problems.

It sends back an error when the files are different.
We use that everyday on production and parse the output to see how different are servers are.

dwoz:
@ProcenaFr @mchugh19
I wonder if it makes sense to add this as another function under the `file` state. Thoughts?

mchugh19:
First off, thank you for your contribution! Take my opinion with a grain of salt, as I'm just a contributor and don't control the direction of salt at all, but I'm not sure this should be a standalone state. 


> It is smarter because it doesn't compare commentaries in files (With "#").
> It removes multi spaces/tabulations before comparing (Useful for files like sudoers).

I don't see where this would be useful, as I'd expect that salt deploys files so that they match expected state. Perhaps this might be a potentially valid use-case and could be added to the file.managed state, but I'd expect files to be configured by a source of truth like file.managed, or as a managed change like `file.line`, `file.append`, `file.replace`, etc.

> You can indicate if you want or not to change the file if it is different.

It seems sort of not-salty to have a state not put a machine into the desired configuration. This could possibly be used as a run once salt module, but that seems similar to the file.get_diff module.

> You can indicate to just check or modify the permissions.

Again, checking permissions would usually be done by a run once module like file.check_perms, and actually setting the desired state would be done through file.managed. 

> You don't need to use "test=True" at the end of your line (for file.managed) which could be forgotten and cause problems.

I'd expect this to be by design, and working around standard functionality to be a bug. It would place this state into a special category operating differently to all others. 

> It sends back an error when the files are different.

If you intend to error if the contents are different you could raise a failure with onchanges and file.managed. But the way this is described seems like this could be a human managed process and would be better handled by the file module rather than a declaration of intended state as managed by state modules. 


I'd see value in keeping the consistency of the state system through the use of  `test=True` and describing the desired state and expecting states to deploy it. Consolidation around common code also allows for bug fixes and features on that common code, for example adding selinux support to `file.managed` rather than having more state modules control permissions and needing updates for that functionality.


A massive benefit to saltstack is the modular nature of the platform. So if this code is working for you all, by all means keep doing it. But you might also look at why you need such different ways of operating as you might find benefit when standardizing your practices which how others handle these problems. Also, just because I don't see it, there may also be value in adding the comment and line space functionality to the common modules so that they might be more widely discovered and used by others.

max-arnold:
You can use `survey.diff` to compare files on different servers https://docs.saltstack.com/en/latest/ref/runners/all/salt.runners.survey.html#salt.runners.survey.diff

```
sudo salt-run survey.diff "*" file.get_diff salt:///sudoers /etc/sudoers
```

Also, your module seems too specific, and tailored only for your use case:

* The module does not look like a proper state module: it doesn't respect `test=True` option that is expected to work in all states
* It does too many things at once (compares files, applies some specific heuristic, modifies files)
* It checks file permissions but ignores file owner. What if I need to check and enforce a file owner too?
* It assumes that file format uses `#` as a comment symbol. What if a file uses `;` or `//`?
* What if a file format is space/tab sensitive?

I would also like to get better understanding of a process that prompted you to write this module:

1. What specific files you control using this module?
2. What files you **do not** control using this module and why?
3. Why your files keep changing, who changes them and when?
4. Why you can't use Salt as the only source of truth and prevent any other ways to modify the files?
5. Why you want this module to be in core and can't just keep it in your `salt://_states` directory?
6. Are there any other approaches you tried before you wrote this module? What didn't work?

max-arnold:
Also, are you aware of `salt.stages.augeas`? https://docs.saltstack.com/en/latest/ref/states/all/salt.states.augeas.html

It has extensible "lenses" which understand specific file formats and can ignore whitespace and correctly handle the comments: http://augeas.net/stock_lenses.html

ProcenaFr:
All right, I'll try to answer to max-arnold first :
* What specific files you control using this module?
We control /etc/redhat-release , /etc/krb5.conf (Not the content) , /etc/resolv.conf depending on the environnement , /etc/sudoers , some docker files , /etc/hosts (Not the content)
* What files you **do not** control using this module and why?
0.
* Why your files keep changing, who changes them and when?
We have more than 10k VMs, lots of process, bad users and lots of bad scripts changing our stuff.
* Why you can't use Salt as the only source of truth and prevent any other ways to modify the files?
We can't modify the production environnement without an RFC.
* Why you want this module to be in core and can't just keep it in your `salt://_states` directory?
I just wanted to share it. We actually use it in salt://_states/.
* Are there any other approaches you tried before you wrote this module? What didn't work?
We tried file.managed : We didn't like the test=True because we wanted to create a full sls file and adding test: True doesn't work.
We tried file.get_diff : We didn't like the fact that it is a module.

* The module does not look like a proper state module: it doesn't respect `test=True` option that is expected to work in all states
I can modify that.
* It does too many things at once (compares files, applies some specific heuristic, modifies files)
True. We needed something simple with multi-usages for our team.
* It checks file permissions but ignores file owner. What if I need to check and enforce a file owner too?
It is because salt on the master part needs the access of the file. So sometimes for files without read permissions, salt can't access it.
* It assumes that file format uses `#` as a comment symbol. What if a file uses `;` or `//`?
It is not perfect.
* What if a file format is space/tab sensitive?
We didn't have this problem so we coded it that way.

###

I didn't know about the state salt.states.augeas.



###

@mchugh19 
* I don't see where this would be useful, as I'd expect that salt deploys files so that they match expected state
Some sudoers files were modified by hand. Tabulations/spaces were removed and it caused false positive (way too many). We have applications that just add lines to the /etc/sudoers file and we can't make sure the whole park has the same /etc/sudoers file.

* It seems sort of not-salty to have a state not put a machine into the desired configuration
I completely agree with that. The big problem with the state is that it doesn't respect the declarative language of salt but we needed that function.

* Again, checking permissions would usually be done by a run once module like file.check_perms
We didn't want to use 2 states to compare the content and the permissions.

dwoz:
@ProcenaFr Thank you for sharing! It sounds like this functionality tackles a fairly special use case and makes sense to keep as an extension module. Since salt core already handle much of this in a general way, I'm going to close this PR.

PR #51502, opened 2019-02-05
51501 states boto3 route53.py fails to validate alias a records

martin-paulus:
### What does this PR do?

Enable boto3_route53 state to correctly handle ALIAS (A-record) validation.

51501

### Previous Behavior

Previously, boto3_route53.rr_present would crash on validating an existing ALIAS record.

### New Behavior

Now, boto3_route53.rr_present is able to correctly create, update, validate ALIAS records.

### Tests written?

No

### Commits signed with GPG?

No


dwoz (approved):

anitakrueger:
@dwoz Is this going to be merged into 2019.2 as well by any chance? We are managing lots of ALIAS records with this state and it creates them over and over again because they are not correctly validated at the moment.

martin-paulus:
@anitakrueger (and visitors:) The changes in this PR are present in 2019.2: https://github.com/saltstack/salt/blob/779bc179e25be49d3449a56648bf827bcb6d5fb6/salt/states/boto3_route53.py#L637

anitakrueger:
@martin-paulus Unfortunately I don't see this working in 2019.2.2.
I just ran the state that should ensure one of our alias records is present for an ALB, and it got created again despite already being there. Running with test=True also says "Route 53 resource record XYZ with type A would be added."

Should I open another issue and if so, what details would be needed please?

martin-paulus:
@anitakrueger Yes, I suppose opening another issue is the way to go. But could you just confirm if the salt/states/boto3_route53.py file in your installation contains the code in this PR? (near lines 637-693)

If the code is there, then it is definitely a different issue you are experiencing. In that case there is no need to refer to this PR or continue the discussion here.

Thank you!

anitakrueger:
@martin-paulus Yes, the code is definitely in there.
I'm wondering how this was tested then? And if this code isn't entirely working, I would think referencing this case would actually be helpful.
I'll try to open a new issue tonight.

Issue #51481, opened 2019-02-03
grains ip6_interfaces returns ipv4 addresses if secondary


styx-tdo:
### Description of Issue/Question
ip6_interfaces:
  ---------
      interfacename:
            - fe80::1
            - **10.0.0.1**
### Setup
Setup:
Interface has both ipv4 and ipv6 addresses.
i.e.:
v4: 10.0.0.1/24 10.0.0.2/24
v6: fe80::1

### Steps to Reproduce Issue
call grains.get ip6_interfaces

I traced this to:
grains/core.py line 2127
-> this gets ifaces[face].get('secondary',[])
this, in the end, calls
utils/network.py
which has a proto defaulting to inet (in this case, proto needs to be inet6)

### Versions Report
2018.3.3

styx-tdo:
possible fix:

        for secondary in ifaces[face].get('secondary', []):
            try:
              socket.inet_pton(socket.AF_INET6, secondary['address'])
              if 'address' in secondary:
                iface_ips.append(secondary['address'])
            except:
              pass
        ret[face] = iface_ips

Ch3LL:
can you post as a PR so we can get more eyes on the fix. thanks :)

Thoro:
I just ran into this issue, and thought it was fixed in salt 3001.3 - it actually is not, can someone take a look?

Ch3LL:
i went ahead and re-opened since the PR was never merged. Please feel free to push a PR as it is not currently assigned to anyone.

mayrstefan:
This looks like #61370 which should be fixed by #61434

PR #51513, opened 2019-02-06
fix grains: ip6_interfaces returns secondary v4 addresses

styx-tdo:
…1481

### What does this PR do?
Fix ipv6 grains returning ipv4 secondary addresses

#51481 

### Previous Behavior
IPv6 grains returned secondary ipv4 addresses, as the call to the network functions did not specify ipv6

### New Behavior
Filter return list to exclude ipv4

### Tests written?

No

### Commits signed with GPG?

No

### DIFF
--- a/salt/grains/core.py
+++ b/salt/grains/core.py
@@ -2290,8 +2290,12 @@ def ip4_interfaces():
             if 'address' in inet:
                 iface_ips.append(inet['address'])
         for secondary in ifaces[face].get('secondary', []):
-            if 'address' in secondary:
-                iface_ips.append(secondary['address'])
+            try:
+                socket.inet_pton(socket.AF_INET6, secondary['address'])
+                if 'address' in secondary:
+                    iface_ips.append(secondary['address'])
+            except:
+                pass
         ret[face] = iface_ips
     return {'ip4_interfaces': ret}
 


salt/grains/core.py
```diff
@@ -2290,8 +2290,12 @@ def ip4_interfaces():
             if 'address' in inet:
                 iface_ips.append(inet['address'])
         for secondary in ifaces[face].get('secondary', []):
-            if 'address' in secondary:
-                iface_ips.append(secondary['address'])
+            try:
+                socket.inet_pton(socket.AF_INET6, secondary['address'])
+                if 'address' in secondary:
+                    iface_ips.append(secondary['address'])
+            except:
```
dwoz: Please add an explicit exception type here and only except the specific exception expected. This will also fix the linter. :)

dwoz (changes requested):

dwoz:
@styx-tdo We are no longer accepting PRs to the `develop` branch. This needs to be re-opened against the `master` branch. You can link to this PR from the new one to preserve any disscussion history. Sorry for the confusion.

styx-tdo:
nah, it's ok for me. i have fixed it in my version, so whatever. you can close and delete the request.

Issue #51537, opened 2019-02-07
salt.utils.etcd_util emits an EtcdNotFile error (emits not a file to logs) when creating a directory (because it's not a file)


arizvisa:
### Description of Issue/Question
`salt.utils.etcd_util` doesn't properly distinguish between creating directories and files. When creating a directory with `etcd.directory` (or anything else), `salt.utils.etcd_util` stupidly outputs "$path not a file".

### Setup
Setup etcd somewhere.

### Steps to Reproduce Issue
Simply call etcd.directory on a directory that already exists.

### Versions Report
All of them, nobody seems to use `etcd`.

### Report
The issue is that all writes to etcd have been funnelled into a method off of the `salt.utils.etcd_util.EtcdClient` object. This method, .write(), takes a parameter `dir` to distinguish between whether to reset `value` to `None` or not and is later used to distinguish between the property types.

This is fine, but when an exception is raised, the method does not bother to distinguish between the types of exceptions and thus an exception used to distinguish when a key exists (which occurs when applying states) an error gets passed through.

arizvisa:
(whenever that gets merged that is). Sorry, phrased that prior comment in the wrong tense.

dwoz:
@arizvisa Thanks so much for reporting the issue and for the PR!

arizvisa:
Closing this as it was fixed by PR #51538.

PR #51538, opened 2019-02-07
Fixes salt.utils.EtcdClient.write so it distinguishes between the different python-etcd logic for files/directories

arizvisa:
### What does this PR do?
Fixes some oversights in the salt.utils.etcd_util module when creating directories vs files.

Closes issue #51537

### Previous Behavior
When creating a directory an error message is logged saying the path is not a file.

### New Behavior
Fixes the prior mentioned behaviour.

### Tests written?
I'm tired.

### Commits signed with GPG?
No.

### DIFF
--- a/salt/utils/etcd_util.py
+++ b/salt/utils/etcd_util.py
@@ -267,12 +267,14 @@ class EtcdClient(object):
         return self.write(key, value, ttl=ttl, directory=directory)
 
     def write(self, key, value, ttl=None, directory=False):
-        # directories can't have values, but have to have it passed
         if directory:
-            value = None
+            return self.write_directory(key, value, ttl)
+        return self.write_file(key, value, ttl)
+
+    def write_file(self, key, value, ttl=None):
         try:
-            result = self.client.write(key, value, ttl=ttl, dir=directory)
-        except (etcd.EtcdNotFile, etcd.EtcdNotDir, etcd.EtcdRootReadOnly, ValueError) as err:
+            result = self.client.write(key, value, ttl=ttl, dir=False)
+        except (etcd.EtcdNotDir, etcd.EtcdRootReadOnly, ValueError) as err:
             log.error('etcd: %s', err)
             return None
         except MaxRetryError as err:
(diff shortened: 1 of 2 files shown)


arizvisa:
This PR was only written for the file-directory duality in the `salt.utils.etcd_util` module and does not affect error messages that are emitted when an empty directory entry is attempted to be read.

This is done by the `salt.states.etcd_mod` in order to determine the original value to compare to. This function will pass the `recurse=True` option to `salt.utils.etcd_util.read()` in an attempt to read the contents of the directory. Due to the directory being empty, the `etcd` client will raise an `EtcdKeyNotFound` exception, which will then get logged before the `.read()` function properly returns `None`. This is a red herring in that no error has happened, there are just no "files" in the directory.

Nonetheless, this only fixes the `salt.utils.EtcdClient.write()` method since other things depend on it.

dwoz:
@arizvisa are you able to create a test that verifies this functionality? I think it will prevent us from having a regression in the future.

arizvisa:
I'll see what I can do in the future, it'll take some time though as my primary job isn't related to development or even close to this sorta devops stuff so it's hard to find business cause for this sorta thing.

Honestly I expected using saltstack was to be a quick effort, but I ended up spending a lot more time on this due to the Etcd support apparently being underused (Such as #51345 which you were a part of, then PRs #51363 and #51346). Still a fan of the whole design of salt, just wasn't expecting to do so much work for my one very-specific configuration.

arizvisa:
Oh also, I think I mentioned the issue with the `salt.state.etcd_mod.directory` state calling the `.get()` method on a directory that doesn't exist and it not catching the expected error. That will probably need to get caught at some point and demoted to a `log.trace` as well.

I didn't create the issue yet because I don't have too much time to fix and test it. So just a heads up if someone else gets to writing tests for this PR before I do, they'll still see an error in the logs when using `etcd.directory` on a path that needs to be created.

dwoz:
@arizvisa One of the existing Etcd unit tests is failing, This needs to be fixed before we can merge.

arizvisa:
force-push due to a rebase against develop. added unit tests for both keys and directories. we'll see how it turns out..

arizvisa:
Grr...first force-push was against develop instead of origin/develop. This next push will be against the correct branch...

arizvisa:
Um...am I reading this error right from py2-windows-2016? It seems to have terminated before it even ran my tests?
```
>>>>>> ------Exception-------
>>>>>> Class: Kitchen::ActionFailed
>>>>>> Message: 1 actions failed.
>>>>>>     Failed to complete #verify action: [[WSMAN ERROR CODE: 2152992672]: <f:WSManFault Code='2152992672' Machine='10.200.20.199' xmlns:f='http://schemas.microsoft.com/wbem/wsman/1/wsmanfault'><f:Message><f:ProviderFault path='C:\Windows\system32\pwrshplugin.dll' provider='microsoft.powershell'>Exception of type &apos;System.OutOfMemoryException&apos; was thrown.</f:ProviderFault></f:Message></f:WSManFault>] on py2-windows-2016
```

dwoz:
> Um...am I reading this error right from py2-windows-2016? It seems to have terminated before it even ran my tests?
> 
> ```
> >>>>>> ------Exception-------
> >>>>>> Class: Kitchen::ActionFailed
> >>>>>> Message: 1 actions failed.
> >>>>>>     Failed to complete #verify action: [[WSMAN ERROR CODE: 2152992672]: <f:WSManFault Code='2152992672' Machine='10.200.20.199' xmlns:f='http://schemas.microsoft.com/wbem/wsman/1/wsmanfault'><f:Message><f:ProviderFault path='C:\Windows\system32\pwrshplugin.dll' provider='microsoft.powershell'>Exception of type &apos;System.OutOfMemoryException&apos; was thrown.</f:ProviderFault></f:Message></f:WSManFault>] on py2-windows-2016
> ```

Yes, we still have some issues in the test runner. Having said that, it looks like the next test run passed. :) Merging.

dwoz (approved):

arizvisa:
Thx.

PR #51562, opened 2019-02-08
Bugfix : Fix RabbitMQ policy definition change check

gilbrechbuhler:
### What does this PR do?
Fix the way the rabbitmq_policy.present state compares the existing policy definition to the given policy definition to define the changes to apply.

As this field is a json field, compare the dict objects generated by `json.loads` instead of doing a string comparison.

Could not find any related issue.

### Previous Behavior
An example state is defined like the following :
```
rabbit_policy:
  rabbitmq_policy.present:
    - name: HA
    - pattern: '.*'
    - definition: '{"ha-mode": "all"}'
```

Now, `salt "*saltmaster*" rabbitmq.list_policies` returns the policies for the vhost /, with the previous policy defined it returns for the policy definition :
``` 
definition:
                {"ha-mode":"all"}
```

So, in the current state of the code, the definition from the state and the definition returned by `list_policies` are compared as string, and due to the space that is removed between the property name and its value, the state acts like there was indeed a change in the definition.

### New Behavior
Policy definitions are compared as dictionnaries and not as strings, avoiding uneeded re-application of existing policy.

### Tests written?

No

### Commits signed with GPG?

No

### DIFF
--- a/salt/states/rabbitmq_policy.py
+++ b/salt/states/rabbitmq_policy.py
@@ -21,6 +21,7 @@ from __future__ import absolute_import
 
 # Import python libs
 import logging
+import json
 import salt.utils
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


dwoz:
@gilbrechbuhler This works for old and new versions of RabbitMQ?

gilbrechbuhler:
@dwoz From what I can see, policies exist since version 3.0.0 of RabbitMQ (https://www.rabbitmq.com/changelog.html). So I tested with version 3.0 and it works perfectly. I also tested with the last released version of RabbitMQ (3.6.15) and it also works as expected.

Do you want me to run more extensive tests ?

dwoz:
@gilbrechbuhler No, I think this is okay. In the future though, you should create an issue for the bug before starting the work to fix it. :) It just helps us track these types of things and prevent regressions. Thanks for the contribution!

PR #51563, opened 2019-02-08
Imgadm docker

sjorge:
### What does this PR do?
After the recent issues reported with imgadm module/state on SmartOS I did a few more tests and two small improvement should be made.

1. Up until now imgadm module assumed that docker images were always called `account/image:tag`, but this is not always the case, e.g. `fedora:latest` is a valid docker image too.

2. auto_import would not import image for disks, this resulted in confusing errors about image_size.

N/a

### Previous Behavior
Docker images like `fedora:latest` could not be imported. 
Unhelpful error when disk images were missing, even with auto_import set to true.

### New Behavior
Docker images like `fedora:latest` can now be imported. They might still not run properly, but at least now they can be imported.

With auto_import set to true we now also import disks images, this confusing error is still present with auto_import set to false, but we pass it directly from upstream (vmadm). So not much we can do about it in salt.

### Tests written?
No

### Commits signed with GPG?
No

### DIFF
--- a/salt/modules/smartos_imgadm.py
+++ b/salt/modules/smartos_imgadm.py
@@ -110,8 +110,7 @@ def _split_docker_uuid(uuid):
         if len(uuid) == 2:
             tag = uuid[1]
             repo = uuid[0]
-            if len(repo.split('/')) == 2:
-                return repo, tag
+            return repo, tag
     return None, None
 
 
--- a/salt/states/smartos.py
+++ b/salt/states/smartos.py
@@ -157,8 +157,7 @@ def _split_docker_uuid(uuid):
         if len(uuid) == 2:
             tag = uuid[1]
             repo = uuid[0]
-            if len(repo.split('/')) == 2:
-                return repo, tag
+            return repo, tag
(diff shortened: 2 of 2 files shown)


sjorge:
I'm confused by the lint failures, locally they pass lint?

```
[hyperon :: sjorge][~/Desktop/_github/salt][imgadm_docker ✔]
[.]$ ~/.local/bin/pylint --rcfile=.testing.pylintrc --disable=W1307 salt/modules/smartos_imgadm.py
Using config file /Users/sjorge/Desktop/_github/salt/.testing.pylintrc

[hyperon :: sjorge][~/Desktop/_github/salt][imgadm_docker ✔]
[.]$ ~/.local/bin/pylint --rcfile=.testing.pylintrc --disable=W1307 salt/states/smartos.py
Using config file /Users/sjorge/Desktop/_github/salt/.testing.pylintrc

```

damon-atkins:
Just something up with the server/slave which did the lint. Just try again.

dwoz (approved):

Issue #51531, opened 2019-02-06
Docker login additional configuration options for docker authentication


c4t3l:
https://github.com/saltstack/salt/blob/927adb31cbee132e9fba7fcd3e2ed7cd9643177c/salt/modules/dockermod.py#L1393

In reference to the above, would it be possible to make this function use master/minion config files _as well as_ pillar data?  I have used the "trifecta" of configuration options in other modules, but my external pillar setup does not easily support this particular configuration.  

I am running salt version 2018.3.2 on Centos 7.  My external pillar system is etcd.

c4t3l:
profile = __salt__['config.option'](profile) 
- from Cody

dwoz:
@terminalmage What are your thoughts on this?

c4t3l:
I would like to submit the PR for this.  I will have time to do it over the weekend.

terminalmage:
See my comments [here](https://github.com/saltstack/salt/pull/51571#pullrequestreview-202810015).

terminalmage:
Alternate implementation here: https://github.com/saltstack/salt/pull/51677

terminalmage:
This was fixed 11 months ago. Please close this rather than letting it be closed as "stale".

PR #51571, opened 2019-02-09
Docker login 51531

c4t3l:
### What does this PR do?
Adds functionality for setting docker login info from pillar and master/minion config files.

References #51531 

### Previous Behavior
Docker login execution module only reads configuration data from Pillar.

### New Behavior
Docker login execution module now reads configuration data from minion, master, and pillar.

### Tests written?

No

### Commits signed with GPG?

No

### DIFF
--- a/salt/modules/dockermod.py
+++ b/salt/modules/dockermod.py
@@ -1387,7 +1387,7 @@ def login(*registries):
     # NOTE: This function uses the "docker login" CLI command so that login
     # information is added to the config.json, since docker-py isn't designed
     # to do so.
-    registry_auth = __salt__['config.option]('docker-registries', {})
+    registry_auth = __salt__['config.option']('docker-registries', {})
     ret = {'retcode': 0}
     errors = ret.setdefault('Errors', [])
     if not isinstance(registry_auth, dict):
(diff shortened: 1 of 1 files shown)


terminalmage (commented):
I am not a fan of this implementation for a couple reasons:

1. It drops support for using config keys ending in `-docker-registries`, functionality that many users likely rely on.
2. It uses `config.option` instead of `config.get` which is widely used elsewhere in dockermod.py.

In truth, the reason we limited this to Pillar to begin with is because it lets you store the secrets somewhere _other_ than the minion config file. I'm not convinced it's a great idea to store them elsewhere, but as long as it's properly documented and people are aware of the risk they are taking, I guess I'm OK with it.

However, I think it can be done in a way that doesn't break existing users' configurations. I'll see what I can come up with.

Akm0d (approved):

terminalmage:
@Akm0d I'm confused as to why you approved this PR, since I gave it a negative review. Did you read my comment?

Additionally, I submitted a more comprehensive fix 4 months ago (https://github.com/saltstack/salt/pull/51677) which has yet to be merged and which should supersede this PR.

Ch3LL:
ping @c4t3l are you okay closing this one in favor of #51677

c4t3l:
Yes that is fine with me.  :)

PR #51572, opened 2019-02-09
Fix error retry handling for Route53

kaidokert:
### What does this PR do?

Fixes error handling, as not all errors should be retried. Currently, if a wrong AWS key is used with no access, the error is not reported at all.

### Previous Behavior
Configure `route53.key` in pillar to an invalid value, the access errors are not reported

### New Behavior
With invalid key provided, the exception is correctly raised and reported on state run

### Tests written?

No

### Commits signed with GPG?

No


dwoz (commented):
@kaidokert It looks there are some failing tests related to this change. Can you take a look?

kaidokert:
Checked, i'm completely lost on how any of those changes could result in these errors. It seems completely unrelated.

kaidokert:
Ok jenkins finally showed me the actual issue, linting. Cleaned up

Akm0d (approved):

terminalmage:
Ported to master in https://github.com/saltstack/salt/pull/56696

Issue #51600, opened 2019-02-12
Running schedule.enable or schedule.disable breaks schedule.save


TheLocehiliosan:
### Description of Issue/Question

After running `schedule.enable` or `schedule.disable`, one can no longer run `schedule.save` successfully until salt-minion is restarted. The error reported:
```
TypeError: 'bool' object is not iterable
````
I think this is because a key value pair `enabled: True` is added to the schedule data structure, at the same level as the jobs, so it is attempted to be treated as a job to save, but can not be processed successfully. I came across this error while searching for a way to set the global scheduling "enabled/disabled" state within the minion configuration.

### Setup
None

### Steps to Reproduce Issue
1. Create any sort of schedule.
2. Run either `schedule.enable` or `schedule.disable`
3. Attempt to run `schedule.save` (this will result in error)

You can observe the added `enabled: <boolean>` by running `schedule.list show_all=True`. 

### Versions Report
(Provided by running `salt --versions-report`. Please also mention any differences in master/minion versions.)
```Salt Version:
           Salt: 2018.3.3

Dependency Versions:
           cffi: Not Installed
       cherrypy: Not Installed
       dateutil: Not Installed
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.7.2
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.5.6
   mysql-python: Not Installed
      pycparser: Not Installed
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: Not Installed
         Python: 2.7.5 (default, Oct 30 2018, 23:45:53)
   python-gnupg: Not Installed
         PyYAML: 3.11
          PyZMQ: 15.3.0
           RAET: Not Installed
          smmap: Not Installed
        timelib: Not Installed
        Tornado: 4.2.1
            ZMQ: 4.1.4

System Versions:
           dist: centos 7.6.1810 Core
         locale: UTF-8
        machine: x86_64
        release: 4.9.93-linuxkit-aufs
         system: Linux
        version: CentOS Linux 7.6.1810 Core
```

garethgreenaway:
Closing this out as the fix is in the 2018.3 and later branches, and will be available in the next minor releases. If the problem persists, let us know and we can reopen this issue.

Issue #51266, opened 2019-01-21
TypeError: 'bool' object is not iterable when operating on schedule


OrangeDog:
When using the schedule module (e.g. applying `schedule.present` or running `schedule.enable`) a `TypeError` is thrown. The function returns successfully, but the schedule has not been persisted.
```
[ERROR   ] Exception in callback functools.partial(<function wrap.<locals>.null_wrapper at 0x7fee3181ebf8>, <tornado.concurrent.Future object at 0x7fee2972fc18>)
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/tornado/stack_context.py", line 277, in null_wrapper
    yielded = self.gen.throw(*exc_info)
  File "/usr/lib/python3/dist-packages/tornado/concurrent.py", line 238, in result
    raise_exc_info(self._exc_info)
  File "/usr/lib/python3/dist-packages/salt/minion.py", line 946, in handle_event
    yield [minion.handle_event(package) for minion in self.minions]
    future.result()
    ret = callback()
    return fn(*args, **kwargs)
  File "/usr/lib/python3/dist-packages/tornado/ioloop.py", line 626, in _discard_future_result
  File "/usr/lib/python3/dist-packages/tornado/concurrent.py", line 238, in result
    value = future.result()
  File "<string>", line 4, in raise_exc_info
  File "/usr/lib/python3/dist-packages/tornado/gen.py", line 1063, in run
  File "/usr/lib/python3/dist-packages/tornado/gen.py", line 1055, in run
  File "/usr/lib/python3/dist-packages/salt/utils/schedule.py", line 381, in modify_job
    self.delete_job(name, persist)
    raise_exc_info(self._exc_info)
    result_list.append(f.result())
    raise_exc_info(self._exc_info)
  File "/usr/lib/python3/dist-packages/salt/minion.py", line 2135, in manage_schedule
    self.schedule.modify_job(name, schedule, persist)
  File "<string>", line 4, in raise_exc_info
  File "/usr/lib/python3/dist-packages/tornado/concurrent.py", line 238, in result
  File "<string>", line 4, in raise_exc_info
    yielded = next(result)
  File "/usr/lib/python3/dist-packages/salt/minion.py", line 2275, in handle_event
    self.manage_schedule(tag, data)
  File "/usr/lib/python3/dist-packages/tornado/gen.py", line 828, in callback
  File "/usr/lib/python3/dist-packages/tornado/gen.py", line 307, in wrapper
[...]

PR #51610, opened 2019-02-12
[2018.3] Fixing schedule when enabled/disabled and saved

garethgreenaway:
### What does this PR do?
When looping through the schedule to remove hidden attributes, we should only do that if the item in question is a dictionary.  It could be the attribute that determines if the entire schedule is enabled or disable.

#51600 #51266

### Commits signed with GPG?
Yes


terminalmage (approved):

dwoz:
@garethgreenaway It looks like the test failures might be related.

Issue #51585, opened 2019-02-11
Issues configuring event_return: smtp on salt-master


tuxthepenguin84:
### Description of Issue/Question
I want to configure my Salt Master to send jobs to Splunk, however right now before I get access to our internal Splunk server I'm testing by sending emails (to myself). I just want to validate I have the correct config for event returns.

On my salt master I have this config:
event_return: smtp
smtp.from: user@company.com
smtp.to: user@company.com
smtp.host: exch-server
smtp.port: 25

When I restart my salt-master I get this error in the logs:
```
2019-02-11 15:28:44,760 [salt.utils.templates:181 ][ERROR   ][3663] Rendering exception occurred
Traceback (most recent call last):
  File "/usr/lib/python2.7/dist-packages/salt/utils/templates.py", line 170, in render_tmpl
    output = render_str(tmplstr, context, tmplpath)
  File "/usr/lib/python2.7/dist-packages/salt/utils/templates.py", line 403, in render_jinja_tmpl
    buf=tmplstr)
SaltRenderError: Jinja variable 'id' is undefined
2019-02-11 15:28:44,761 [salt.utils.event :1232][ERROR   ][3663] Could not store events - returner 'smtp.event_return' raised exception: Jinja variable 'id' is undefined
```

Salt commands and jobs still run correctly and I get the emails, but it generates this on all salt commands I run.

### Setup
(Please provide relevant configs and/or SLS files (Be sure to remove sensitive info).)

### Steps to Reproduce Issue
(Include debug logs if possible and relevant.)

### Versions Report
(Provided by running `salt --versions-report`. Please also mention any differences in master/minion versions.)
```
Salt Version:
           Salt: 2018.3.3

Dependency Versions:
           cffi: Not Installed
       cherrypy: Not Installed
       dateutil: 2.4.2
      docker-py: Not Installed
          gitdb: 0.6.4
      gitpython: 1.0.1
          ioflo: Not Installed
         Jinja2: 2.8
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: 1.0.3
   msgpack-pure: Not Installed
 msgpack-python: 0.4.6
   mysql-python: Not Installed
      pycparser: Not Installed
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: Not Installed
         Python: 2.7.12 (default, Nov 12 2018, 14:36:49)
   python-gnupg: 0.3.8
         PyYAML: 3.11
[...]

PR #51622, opened 2019-02-13
Prevents crash when there is no job entry

brejoc:
### What does this PR do?

Seems like in rare race conditions it might happen that there is no job entry. This prevents Salt from crashing at that point.

In rare cases we could see this while bootstrapping a minion:

```
Exception occurred in runner jobs.list_jobs: Traceback (most recent call last):
  File "/usr/lib/python2.7/site-packages/salt/client/mixins.py", line 387, in _low
  File "/usr/lib/python2.7/site-packages/salt/runners/jobs.py", line 306, in list_jobs
    ret = mminion.returners['{0}.get_jids'.format(returner)]()
  File "/usr/lib/python2.7/site-packages/salt/returners/local_cache.py", line 376, in get_jids
    for jid, job, _, _ in _walk_through(_job_dir()):
  File "/usr/lib/python2.7/site-packages/salt/returners/local_cache.py", line 80, in _walk_through
    jid = job['jid']
TypeError: 'NoneType' object has no attribute '__getitem__'
```

I'm sure that the cause of this can be found somewhere else. But in the meantime it would be nice if Salt wouldn't crash at that point of time.

### Previous Behavior

Salt threw a `TypeError` like above.

### New Behavior

Salt logs an error instead.

### Tests written?

No

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/returners/local_cache.py
+++ b/salt/returners/local_cache.py
@@ -77,6 +77,9 @@ def _walk_through(job_dir):
                 except Exception:
                     log.exception('Failed to deserialize %s', load_path)
                     continue
+                if not job:
+                    log.error('Deserialization of job succeded but there is no data in %s', load_path)
+                    continue
                 jid = job['jid']
                 yield jid, job, t_path, final
 


dwoz:
Looks like this may fix #51585

dwoz (approved):

garethgreenaway (approved):

Issue #51620, opened 2019-02-13
return code breaks expected behaviour for pkg.info_installed


brejoc:
### Description of Issue/Question

There have been changes in the way the `retcode` is handled when doing a `pkg.info_installed`. With `2018.3.0` (the latest version we use in SUSE) there was no `retcode` when `pkg.info_installed` was invoked. This has changed for `2019.2.0 RC2`. Now a `retcode` is returned and if a pkg is not installed, it is `1`, which produces a `CommandExecutionError`.

Of course the `retcode` is coming from `rpm -q <pkgname>`, which returns `0` if the pkg is installed and `1` if the pkg is not installed. I can see why this has been changed, but this breaks expected behaviour when used in SLS files. My question is: Was this intended?

Related commits:
This introduced `retcode`: https://github.com/saltstack/salt/commit/e995cba7323b6439693fc2d364464bab999fe3e2
And some refactoring: https://github.com/saltstack/salt/commit/4c233442a23211ef310a6ce394746a4c9e0c1b49

### Steps to Reproduce Issue

2018.3.0:
```
bash-4.4# salt -L "C, E"  pkg.info_installed ceph-common
A:
    ERROR: package ceph-common is not installed
```

2019.2.0 RC2:
```
bash-4.4# salt \* pkg.info_installed ceph-common
A:
    ERROR: package ceph-common is not installed
ERROR: Minions returned with non-zero exit code
```
In the second case (2019.2.0 RC2) the job execution is marked as failed.

### Versions Report

This is for Salt 2019.2.0 RC2.

isbm:
Well, the package was not installed, therefore `salt.defaults.exitcodes.EX_GENERIC`. So I am not really sure how this is an _issue_.

brejoc:
@isbm I understand that the new behaviour is better, but this breaks the expected behaviour when coming from an older Salt version and will cause headache for a lot of customers.

terminalmage:
The retcode handling is expected behavior and is not going to be changed.

What seems wrong to me is that a function that is meant to return info about an installed package raises an exception instead of simply returning an empty dictionary.

terminalmage:
It seems to me the correct fix is not to make `lowpkg.info` raise an exception.

terminalmage:
Or, like @isbm suggested to me just now, catch it in `pkg.info_installed`.

dincamihai:
[...]

PR #51623, opened 2019-02-13
Ignore return code in "run_all" and do not raise an exception therefore

isbm:
### What does this PR do?

Addresses concerns of https://github.com/saltstack/salt/issues/51620

### Tests written?

Existing

### DIFF
--- a/salt/modules/rpm.py
+++ b/salt/modules/rpm.py
@@ -545,17 +545,8 @@ def info(*packages, **kwargs):
     query.append("-----\\n")
 
     call = __salt__['cmd.run_all'](cmd + (" --queryformat '{0}'".format(''.join(query))),
-                                   output_loglevel='trace', env={'TZ': 'UTC'}, clean_env=True)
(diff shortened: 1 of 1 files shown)


dincamihai:
Correct me if I'm wrong, but isn't this hiding any error raised by rpm?

```
try:
    rpm()
except:
    pass
```

thatch45:
@dincamihai raises a valid point, we should likely make sure that we at least set the final retcode for the function itself as well as make sure things are logged.

dwoz:
@isbm Ping. We're getting close to freezing 2019.2.1.

dwoz:
@isbm This needs to be re-opened against the `master` branch. We are no

terminalmage:
@dwoz this was intended as a bugfix in the 2019.2 release cycle, from the looks of it. I took the liberty of cherry-picking this onto master in https://github.com/saltstack/salt/pull/55587.

PR #51629, opened 2019-02-13
Locally track JID when executed through async/sync mixin clients

austinpapp:
### What does this PR do?

This is a simple means to keep track of jobs executed through Async and Sync mixin clients (runners). The purpose of this is to better understand what is running on a master. 

As a result, this opens the possibilities to understanding when it is safe to apply new configurations and restart the master service without disrupting current executions. 

This also unifies around where JIDs are stored as this essentially is how minion's do it. 

Key updates
* Ability to track jobs through Async/SyncMixin classes
* Added support functions to handle reading proc files master side
* Added clean proc dir for master side
* Added dateutil and setproctile to tests.txt
* async option for ShellTestCase/ShellCase

### Previous Behavior
The master had no understanding of runners currently executing. 

### New Behavior
The master now has understanding of runners currently executing. 

### Tests written?

Yes

### Commits signed with GPG?

No


austinpapp:
i think integration tests are failing because i needed to add an async option to `runtests.py`. however, im not sure if my `runtests.py` is being used or not.

austinpapp:
A few things here.

1. Will have to look into why cherrypy tests are failing but its not immediately clear to me why
2. In order to test that a proc file is in fact created, i've added the async option to `run_run_plus()` for `ShellCase`. however, it appears the actual execution of my `runner.test.sleep` call is happening *after* the test has wrapped up. not sure why. i gave it a 20s sleep to see if i could catch the delay which works. however, its a 20s sleep and i can't assume that did not have a negative impact to the cherrypy test(s).

waynew:
Hey, thanks for the PR! I'm not 100% positive, but as far as I know Jenkins should be using the code from your branch as-is.

Though, running tests locally seems to succeed - at least `unit.cache.test_imscache` tests. I'm going to try a few other configurations to see if I can get it to fail locally.

austinpapp:
@waynew yea it wouldn't make much sense not to use my updated runtest. anyway, will continue to dig in on why these tests are failing.

waynew:
Well, so far it's looking like the tests are a lot happier here! 👍 Assuming the final test is also A-OK I'll take a look at this later today

tests/unit/utils/test_master.py
```diff
@@ -0,0 +1,106 @@
...
+                assert master.is_pid_healthy(12345) is False
+
+    def tets_is_pid_healthy_raises(self):
+        m_fopen = mock_open(side_effect=IOError)
+        with patch('salt.utils.files.fopen', m_fopen):
+            assert master.is_pid_healthy(12345) is False
+
+        m_fopen = mock_open(side_effect=OSError)
+        with patch('salt.utils.files.fopen', m_fopen):
+            assert master.is_pid_healthy(12345) is False
```
waynew: Do these pytest assertion styles work well on failures with `pytest-salt`? It seems like they should, though I haven't tried it, and have typically just gone with the UnitTest style assertions. I'm not sure if we have a guideline/standard there.

waynew (commented):
Looks like Jenkins was finally happy!

Code wise this looks great, there are a couple of improvements that I think could be made to the docs. I don't have intense feelings about the changes I've asked for, so if anyone has strong feelings otherwise I'm happy to go ahead an approve as-is.

Otherwise, if you agree with suggestions then I'm happy to merge once those are made :+1:

tests/unit/utils/test_master.py
austinpapp: afaik, they work just fine. i did it this way bc its force of habit for me. haha. at least nothing blew up saying it didnt know how to run them. 🤷‍♂️

austinpapp:
feels like i'm playing cat and mouse with these tests... updated comments and now everything fails. sigh

waynew:
And now things are successful (pending py3-windows-2016) :trollface: 

I know it's small comfort in the face of currently having problems, but *are* we working on improving our CI/CD pipeline. Thank you for your patience and persistence!

austinpapp:
@waynew any chance this can be merged?

thatch45 (approved):

Issue #40152, opened 2017-03-20
ssh_known_hosts.present port option requires hashed_known_hosts False with older ssh-keyscan


shallot:
_From @hackel on December 10, 2015 22:13_

This just started happening after upgrading to 2015.8:
I have the following clause in my users pillar:

``` yml
    ssh_known_hosts:
      slsapp.com:
        port: 1234
        fingerprint: xx:xx:xx
        enc: ssh-dss
```

This results in:

```
          ID: users_ssh_known_hosts_username_0
    Function: ssh_known_hosts.present
        Name: slsapp.com
      Result: False
     Comment: argument port can not be used in conjunction with argument hash_known_hosts
```

hash_known_hosts is not present, so I'm not sure what is going on here.  I tried to explicitly set hash_known_hosts to False in case it was the default, but it didn't help.

_Copied from original issue: saltstack-formulas/users-formula#103_

shallot:
_From @hackel on December 10, 2015 22:22_

Adding the deprecated option `hash_hostname: False` seems to do the trick.  This may be a larger issue in the salt ssh_known_hosts module.

shallot:
_From @kbuilds on September 20, 2016 19:50_

I am also running into this same issue in salt-minion 2016.3.3

shallot:
_From @dosercz on January 12, 2017 18:53_

Same in 2016.11.1 and hack with hash_hostname doesn't work there:(

shallot:
_From @dosercz on January 12, 2017 19:3_

But port could be set if hash_known_hosts is set to False

shallot:
This error seems to have been added in https://github.com/saltstack/salt/commit/c7466e789436adb3c8ede811e18a0d04cfd3366a
but there's no explanation for it. Hashed known_hosts files as such do of course have to support arbitrary ports.

I think this is actually documenting an unrelated bug in the underlying `ssh-keyscan` program used by the `recv_known_host()` function, cf. https://bugs.debian.org/824010

It should be removed or replaced with a warning telling people they need to upgrade their ssh-keyscan for port to work.

gtmanfred:
@jahamn since you wrote this, could you provide us information about why the port does not work with hash_known_host (previously hash_hostname)?

It looks like the hash_hostname workaround no longer works because of changes made in this PR https://github.com/saltstack/salt/pull/27726

And that you should be disabling this now?

Thanks,
Daniel

shallot:
[...]

PR #51633, opened 2019-02-13
Workaround broken ssh-keyscan

waynew:
### What does this PR do?

This is a workaround for #40152 - ssh-keyscan hashing for non-standard
(i.e. 22) ports is broken on basically every single OS. Rather than
hashing, say `[example.com]:55555`, when you pass an alternate port, it
hashes `example.com`. Oops!

This is a problem because `ssh-keygen` *expects* to see a hash for
`[example.com]:55555` in the known_hosts file so when you ssh for the
first time you have to accept the host key because it can't find one for
`[example.com]:55555` in your known_hosts.

Fortunately, `ssh-keygen` can properly hash our hostnames for us, so now
we don't tell ssh-keyscan to hash anymore - we tell ssh-keygen to do it.
### Previous Behavior

Use the `ssh_known_hosts.present` state with a server running on a non-standard port (standard is 22, maybe 443 or 2222 or 55555 or something) and then try to `ssh`. Still asks you to confirm the server key, also adds a new entry to your known_hosts file

### New Behavior

Now it behaves as expected. A state like:

```
example.com:
  ssh_known_hosts.present:
    - port: 55555
    - hash_known_hosts: True
```

Will 1) not throw an error, and 2) work. You can `ssh example.com -p 55555` and you won't be asked to confirm the host key. 

### Tests written?

Yes

### Commits signed with GPG?

Yes


dwoz:
@waynew Are the test failures related?

waynew:
Interesting! I'll check them out - not sure why those should be failing... but I'll check them out!

dwoz (approved):

bbinet:
Can this patch be backported to 2018.3 and 2019.2 branches?

Issue #50525, opened 2018-11-15
cp.get_url in _modules not work on salt-ssh


jils2013:
### Description of Issue/Question
i used cp.get_url(other function seems not work on salt-ssh with salt filesystem) in my module,Unlike direct calls, always return False in the module.

any preparation before call cp.get_url in a module?
### Setup
```
[root@saltmaster ~]# cat /srv/salt/_modules/testmod.py 
def get_url(source,dest):
        return __salt__['cp.get_url'](source,dest)
[root@saltmaster ~]# cat /srv/salt/testcp.sls 
#!py
import logging

def run():
        log=logging.getLogger(__name__)
        log.error(__salt__['cp.get_url']('salt://testfile','/tmp/'))
        log.error(__salt__['testmod.get_url']('salt://testfile','/tmp/'))
        return {}
[root@saltmaster ~]# salt-ssh tgthost state.apply testcp
[ERROR   ] True
[ERROR   ] False
tgthost :

Summary for tgthost 
-----------
Succeeded: 0
Failed:   0
-----------
Total states run:    0
Total run time:  0.000 ms
```

### Steps to Reproduce Issue
(Include debug logs if possible and relevant.)

### Versions Report
master version:2018.3.3

dwoz:
@jils2013 Have you tried the `file.managed` state?

```
/home/admin/testfile:
  file.managed:
    - source: salt://testfile
```

jils2013:
@dwoz ,when i call salt.client.ssh.Single like client/ssh/wrapper/cp.py,get following
`[root@saltmaster ~]# cat /srv/salt/testcp.sls 
#!py
import logging

def run():
        log=logging.getLogger(__name__)
        log.error(__salt__['cp.get_url']('salt://testfile','/tmp/'))
        log.error(__salt__['testmod.get_url']('salt://testfile','/tmp/',__salt__.kwargs))
        return {}
[root@saltmaster ~]# cat /srv/salt/_modules/testmod.py  
import logging
import salt.client.ssh.client

def get_url(source,dest,saltkwargs):
        log=logging.getLogger(__name__)
        single = salt.client.ssh.Single(__opts__,'',**saltkwargs)
        ret = single.shell.send(source, dest)
        return ret
[root@saltmaster ~]# salt-ssh tgthost state.apply testcp
[ERROR   ] True
[ERROR   ] [u'', u'ssh: Could not resolve hostname salt: Name or service not known\r\n', 1]
tgthost:

[...]

Issue #50351, opened 2018-11-02
salt-ssh throws exception when minionfs is enabled.


russellballestrini:
### Setup
salt-ssh throws the following exception when minionfs is enabled on salt master::

```
ERROR   ] An Exception occurred while executing state.highstate: [Errno 2] No such file or directory: '/var/tmp/.root_812c13_salt/running_data/var/cache/salt/minion/minions'
Traceback (most recent call last):
  File "/usr/lib/python2.7/dist-packages/salt/client/ssh/__init__.py", line 1156, in run_wfunc
    result = self.wfuncs[self.fun](*self.args, **self.kwargs)
  File "/usr/lib/python2.7/dist-packages/salt/client/ssh/wrapper/state.py", line 705, in highstate
    roster_grains)
  File "/usr/lib/python2.7/dist-packages/salt/client/ssh/state.py", line 229, in prep_trans_tar
    files = file_client.cache_dir(name, saltenv, cachedir=cachedir)
  File "/usr/lib/python2.7/dist-packages/salt/fileclient.py", line 240, in cache_dir
    for fn_ in self.file_list(saltenv):
  File "/usr/lib/python2.7/dist-packages/salt/fileclient.py", line 1283, in file_list
    return salt.utils.data.decode(self.channel.send(load)) if six.PY2 \
  File "/usr/lib/python2.7/dist-packages/salt/fileserver/__init__.py", line 871, in send
    return getattr(self.fs, cmd)(load)
  File "/usr/lib/python2.7/dist-packages/salt/utils/decorators/__init__.py", line 594, in wrapped
    **salt.utils.data.decode_dict(kwargs)
  File "/usr/lib/python2.7/dist-packages/salt/fileserver/__init__.py", line 752, in file_list
    ret.update(self.servers[fstr](load))
  File "/usr/lib/python2.7/dist-packages/salt/fileserver/minionfs.py", line 253, in file_list
    minion_dirs = os.listdir(minions_cache_dir)
OSError: [Errno 2] No such file or directory: '/var/tmp/.root_812c13_salt/running_data/var/cache/salt/minion/minions'
[DEBUG   ] LazyLoaded nested.output
[TRACE   ] data = {u'mx1.example.net': u"An Exception occurred while executing state.highstate: [Errno 2] No such file or directory: '/var/tmp/.root_812c13_salt/running_data/var/cache/salt/minion/minions'"}
```

### Steps to Reproduce Issue

[...]

Issue #50196, opened 2018-10-24
No way to debug templates over salt-ssh


max-arnold:
### Description of Issue/Question

The recommended ways to debug templates are (as seen in the Slack chat):

```
salt 'example.com' slsutil.renderer /path/to/file.jinja 'jinja'
salt 'example.com' cp.get_template salt://path/to/template /dev/stdout template=jinja
```

Unfortunately, I was unable to make it work over `salt-ssh`:

```
salt-ssh -l debug example.com slsutil.renderer path='salt://test.sls' default_renderer='jinja'
...
[DEBUG   ] Could not LazyLoad slsutil.renderer: 'slsutil.renderer' is not available.
...
```

```
salt-ssh example.com cp.get_template salt://test.sls /dev/stdout template=jinja

example.com:
    ----------
    retcode:
        0
    stderr:
        Traceback (most recent call last):
          File "/var/tmp/.root_7e26e1_salt/salt-call", line 15, in <module>
            salt_call()
          File "/var/tmp/.root_7e26e1_salt/py3/salt/scripts.py", line 400, in salt_call
            client.run()
          File "/var/tmp/.root_7e26e1_salt/py3/salt/cli/call.py", line 57, in run
            caller.run()
          File "/var/tmp/.root_7e26e1_salt/py3/salt/cli/caller.py", line 134, in run
            ret = self.call()
          File "/var/tmp/.root_7e26e1_salt/py3/salt/cli/caller.py", line 212, in call
            ret['return'] = func(*args, **kwargs)
          File "/var/tmp/.root_7e26e1_salt/py3/salt/modules/cp.py", line 335, in get_template
            **kwargs)
          File "/var/tmp/.root_7e26e1_salt/py3/salt/fileclient.py", line 782, in get_template
            **kwargs
          File "/var/tmp/.root_7e26e1_salt/py3/salt/utils/templates.py", line 167, in render_tmpl
            tmplstr = tmplsrc.read()
        AttributeError: 'bool' object has no attribute 'read'
    stdout:
```

Third way also doesn't work:

```
salt-ssh example.com cp.get_file salt://test.sls /dev/stdout template=jinja

example.com:
    False
```

Salt version `2018.3.2` on Py3.

The question is: what is the best (and documented) way to debug templates, which works consistently across `salt`, `salt-call` and `salt-ssh` (and maybe `salt-run`)?

dwoz:
@max-arnold Thank you for reporting this.

[...]

Issue #48067, opened 2018-06-12
cmd.script is not executing script on remote server via ssh


guptasakshi01:
### 
while executing script on target server using salt-ssh i am getting following error
cmd:- sudo salt-ssh minion cmd.script source=salt://test.sh 
minion: 
   ----------
    cache_error:
        True
    pid:
        0
    retcode:
        1
    stderr:
    stdout:

### Setup
test.sh script is present at master on the /srv/salt/
We are executing command considering script will be executed remotely on target server(minion).

### Versions Report
Salt Version:
           Salt: 2017.7.1

Dependency Versions:
           cffi: 1.11.5
       cherrypy: 3.5.0
       dateutil: 2.7.0
      docker-py: Not Installed
          gitdb: 0.6.4
      gitpython: 1.0.1
          ioflo: Not Installed
         Jinja2: 2.8
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: 1.0.3
   msgpack-pure: Not Installed
 msgpack-python: 0.4.6
   mysql-python: 1.3.7
      pycparser: 2.18
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: Not Installed
         Python: 2.7.12 (default, Dec  4 2017, 14:50:18)
   python-gnupg: Not Installed
         PyYAML: 3.11
          PyZMQ: 15.2.0
           RAET: Not Installed
          smmap: 0.9.0
        timelib: Not Installed
        Tornado: 4.2.1
            ZMQ: 4.1.4

System Versions:
           dist: Ubuntu 16.04 xenial
         locale: UTF-8
        machine: x86_64
        release: 4.4.0-112-generic
         system: Linux
        version: Ubuntu 16.04 xenial

gtmanfred:
I am unable to replicate this behavior.

```
[root@salt ~]# cat /srv/salt/test.sh
#!/usr/bin/env bash
echo 'Hello, World!'
[root@salt ~]# salt-ssh -i \* cmd.script source=salt://test.sh
server:
    ----------
    pid:
        1744
    retcode:
        0
    stderr:
    stdout:
        Hello, World!
[root@salt ~]# salt-ssh --versions-report
Salt Version:
           Salt: 2017.7.1

Dependency Versions:
           cffi: Not Installed
       cherrypy: Not Installed
       dateutil: Not Installed
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
[...]

PR #51636, opened 2019-02-13
PoC / WiP for #48067 and related

The-Loeki:
### What does this PR do?
Salt-SSH 'wraps' around the `cp` module to enable transport over SSH. This wrap is however incomplete and is done by manually rewriting parts of the original function in the wrapper.
As a number of functions are missing, the direct invocation of salt-ssh will b0rk on them, subtly or less so for range of modules & functions.
See discussion in #48067.

This is a working PoC to take a slightly different tack on that problem, by injecting a modified `fileclient`.
The end result is that the function calls to the cp module can be left untouched & be called directly, increasing stability, reducing code dupliation and erroneous and unexpected behaviour from `salt-ssh` when calling modules directly.

I would really like some input on this solution before proceeding, as I'm not sure I fully understand the consequences of what I'm doing here code/namespace/injection/cleanliness wise.

#48067 
#50525
#50196 
#50351

### Previous Behavior
Salt-SSH Wrapping functions in the `cp` module is cumbersome and difficult

### New Behavior
Salt-SSH Wrapping functions in the `cp` module is a matter of pointing at the actual `cp` module and very maybe add/overload a function.

### Tests written?

No

### Commits signed with GPG?

Yes/No


waynew:
You haven't made any changes to the `cmd.script` part, right? At least when I try that (even after doing a `cp.cache_file`) it fails with a cache error.

I think what we'd want is something similar to what you've done (though maybe someone else in @saltstack/team-core who's more familiar with salt-ssh can weigh in on this), since this appears to actually be asking the master for the file. What I think we'd want is to have the file client injected into the fileclients on the client. If that's not too many clients :trollface:

If you need me to hunt it down, I can find the place I think it's in the cmd module, where it's checking fileclients - that's where we'd want this one, I believe.

The-Loeki:
Well I figured out the wrapper is actually only executed master-side upon direct invocation. 
As you say, for that too work the fileclient/wrapper needs to be active during local execution. That and the minion would be pulling rather than the master pushing.

waynew:
Ahh. That makes some sense, regarding what we've seen. It sounds like we just need to see if it's possible for the salt-ssh master to receive file requests, rather than the existing method of bundling up the file when we start the call.

mchugh19:
Some sort of wrapper or utility function like this would still be very handy for the modules which do include `cp` functions if they want to implement `salt/client/ssh/wrapper/` capabilities. 

I was looking at getting saltcheck to support salt-ssh, and it looks like it will need to override the `cp.cache_dir` calls with a salt/client/ssh/wrapper/saltcheck.py wrapper which copies over the needed files to the ssh-minion's cache directory, then calls the expected functions how that the files are on the minion. 

Relatedly, when testing out this PR, I noticed that `cp.cache_dir` only copies to the ssh user's home directory, and not the salt minion's cache dir:
```
# salt-ssh -l debug minion-ssh cp.cache_dir salt://cron
...
[DEBUG   ] Terminal Command: /bin/sh -c scp -o KbdInteractiveAuthentication=no -o PasswordAuthentication=no -o GSSAPIAuthentication=no -o ConnectTimeout=65 -o Port=22 -o IdentityFile=/etc/salt/pki/master/ssh/salt-ssh.rsa -o User=vagrant  salt-ssh/minion-ssh/files/base/cron/service.sls 192.168.10.3:
...
```

So rather than populating `/var/tmp/.vagrant_526c8b_salt/running_data/var/cache/salt/minion/files/cron/service.sls` it's in `/home/vagrant/service.sls`

dwoz:
@The-Loeki This needs to be re-opened against the `master` branch and link back to this issue for the discussion. We are no longer accepting PRs to `2018.3`. Sorry for the confusion.

The-Loeki:
> 
> 
> @The-Loeki This needs to be re-opened against the `master` branch and link back to this issue for the discussion. We are no longer accepting PRs to `2018.3`. Sorry for the confusion.

Dude this was a test-case in last February while you guys were really busy upgrading your pipelines. 
I've repeatedly asked for comments & suggestions in here, and I'll be happy to rebase/update if need be, because I guarantee this problem still exists and the discussion for the core solution needed too.

So please don't 'just' close this

PR #51652, opened 2019-02-14
modules.postgres: fix function privileges check

emerichunter:
### What does this PR do?

Fixes the column name in the function privileges table `pg_proc`.

See https://www.postgresql.org/docs/current/catalog-pg-proc.html

None

### Previous Behavior

The function privilege is reapplied every time.

### New Behavior

The salt function is idempotent.

### Tests written?

No

### Commits signed with GPG?

Yes


dwoz (approved):
@emerichunter Thanks for linking to the docs. LGMT

PR #51667, opened 2019-02-15
Persist tornado.iostream until module unloads

rares-pop:
There is a race-condition where tornado.iostream
gets unloaded while the code in tcp.py is executed
and when it's doing:
except tornado.iostream.StreamClosedError
it's looking for StreamClosedError on None, and raises.

This happened in the past but got somehow workaround
Sergey's change (to delay things), but if the
Windows has a Japanese or Chinese locale, the bug reproduces.

@dwoz , @twangboy - does this have side-effects? I only validated that it works on Windows with different locales, and is an actual fix until tornado 5.0 support is complete

### Previous Behavior
There was an exception thrown after any salt * --async test.ping or any other async jobs.

### New Behavior
No errors are returned in CLI

### Tests written?

Yes

### Commits signed with GPG?

Yes


twangboy (approved):
So, just changing the exception syntax fixes this? It looks basically like the same code... not really changing any logic. Am I missing something?

rares-pop:
@twangboy - it's the actual import that's fixing the problem (having the global import). Is that keeping a reference and forces tornado.iostream to stay in memory?

It was @skizunov that proposed this fix, and I only tested it and it works. Again, this is supposed to be temporary, until tornado 5.0 is supported, which whouldn't require this. Maybe Sergey can provide more info on this topic.

dwoz (approved):
I can't see how this would hurt anything, approved. :)

PR #51670, opened 2019-02-15
 Fix extract_hash case for top-level binary shasums 

OrangeDog:
For example:

    b45991e49e50d89e8ceb1d6b7b0e881cece39e63 *filename

Such a file is generated with e.g. `shasums -b` or the `checksum-maven-plugin`.
This allows the hash to be matched.


OrangeDog:
Test failures appear to be from the `@flaky` test wrapper, so I assume that wasn't me.

tests/unit/modules/test_file.py
```diff
@@ -907,9 +907,9 @@ def test_extract_hash(self):
         with tempfile.NamedTemporaryFile(mode='w+b', delete=False) as tfile:
             tfile.write(salt.utils.stringutils.to_bytes(
                 'rc.conf ef6e82e4006dee563d98ada2a2a80a27\n'
-                'ead48423703509d37c4a90e6a0d53e143b6fc268 example.tar.gz\n'
-                'fe05bcdcdc4928012781a5f1a2a77cbb5398e106 ./subdir/example.tar.gz\n'
-                'ad782ecdac770fc6eb9a62e44f90873fb97fb26b foo.tar.bz2\n'
+                'ead48423703509d37c4a90e6a0d53e143b6fc268  example.tar.gz\n'
+                'fe05bcdcdc4928012781a5f1a2a77cbb5398e106  ./subdir/example.tar.gz\n'
```
skizunov: Is there a reason why there are 2 spaces after the hash (instead of the original 1 space) for these 2 items? If not, would you change these 2 back?

OrangeDog: That’s the output I get from shasums.

skizunov: Could you test at least one of those two with just one space, to ensure backwards compatibility is preserved?

skizunov (approved):

cmcmarrow (approved):

PR #51674, opened 2019-02-15
[2019.2] Merge forward from 2018.3 to 2019.2

Ch3LL:
Conflicts:
  - pkg/windows/build_env_2.ps1
  - pkg/windows/build_env_3.ps1
  - pkg/windows/req.txt
  - pkg/windows/req_win.txt
  - salt/grains/core.py

ping @s0undt3ch i decided to keep all the changes from 2019.2 instead of 2018.3 for the pkg/windows/* files, because it looked like it was intentional as you created a PR for each branch for the updates. Let me know if this is wrong and i can handle the merge-conflicts again.


s0undt3ch (approved):
You did right!

dwoz (approved):

Ch3LL:
@dwoz i removed that skipif. The test should be able to run on non solaris platforms. The test was failing because in 2019.2 the return was changed here 69ab0b8f167 from `Package already installed` to {} and these new tests where originally written for the 2018.3 branch

PR #51677, opened 2019-02-15
2 enhancements to config.option

### LINKED ISSUE #51531: Docker login additional configuration options for docker authentication

terminalmage:
This makes the following two changes to `config.option`:

1. Matches are now returned from grains
2. Wildcard matches are now supported

For backward compatibility, wildcard matches are only attempted when `wildcard=True` is passed.

Additionally, the Docker registry configuration now uses `config.option`, meaning that registry auth can be configured elsewhere than the pillar data.

Resolves #51531.

### DIFF
--- a/doc/topics/releases/neon.rst
+++ b/doc/topics/releases/neon.rst
@@ -372,9 +372,19 @@ Module Changes
 - Added new :py:func:`boto_ssm <salt.modules.boto_ssm>` module to set and query
   secrets in AWS SSM parameters.
 
-- The :py:func:`file.set_selinux_context <salt.modules.file.set_selinux_context>`
-  module now supports perstant changes with ``persist=True`` by calling the
-  :py:func:`selinux.fcontext_add_policy <salt.modules.selinux.fcontext_add_policy>` module.
+- The :py:func:`file.set_selinux_context
+  <salt.modules.file.set_selinux_context>` module now supports perstant changes
+  with ``persist=True`` by calling the :py:func:`selinux.fcontext_add_policy
+  <salt.modules.selinux.fcontext_add_policy>` module.
+
+- The :py:func:`config.option <salt.modules.config.option>` now also returns
+  matches from the grains, making it align better with :py:func:`config.get
+  <salt.modules.config.get>`.
+
+- Configuration for Docker registries is no longer restricted only to pillar
+  data, and is now loaded using :py:func:`config.option
+  <salt.modules.config.option>`. More information on registry authentication
+  can be found :ref:`here <docker-authentication>`.
 
 - The :py:func:`yumpkg <salt.modules.yumpkg>` module has been updated to support
   VMWare's Photon OS, which uses tdnf (a C implementation of dnf).
--- a/doc/topics/releases/neon.rst
+++ b/doc/topics/releases/neon.rst
@@ -373,8 +373,9 @@ Module Changes
   secrets in AWS SSM parameters.
 
 - The :py:func:`file.set_selinux_context
-  <salt.modules.file.set_selinux_context>` module now supports perstant changes
-  with ``persist=True`` by calling the :py:func:`selinux.fcontext_add_policy
+  <salt.modules.file.set_selinux_context>` module now supports persistant
+  changes with ``persist=True`` by calling the
+  :py:func:`selinux.fcontext_add_policy
   <salt.modules.selinux.fcontext_add_policy>` module.
 
 - The :py:func:`config.option <salt.modules.config.option>` now also returns
--- a/salt/modules/config.py
+++ b/salt/modules/config.py
@@ -242,7 +242,6 @@ def option(
         return ret or default
 
 
-
 def merge(value,
           default='',
           omit_opts=False,
--- a/salt/modules/config.py
+++ b/salt/modules/config.py
@@ -79,6 +79,7 @@ DEFAULTS = {'mongo.db': 'salt',
             'virt': {'tunnel': False,
                      'images': os.path.join(syspaths.SRV_ROOT_DIR, 'salt-images')},
 
+            'docker.exec_driver': 'docker-exec',
             'docker.compare_container_networks': {
                 'static': ['Aliases', 'Links', 'IPAMConfig'],
                 'automatic': ['IPAddress', 'Gateway',
--- a/salt/modules/config.py
+++ b/salt/modules/config.py
@@ -6,6 +6,7 @@ Return config information
 # Import python libs
 from __future__ import absolute_import, print_function, unicode_literals
 import copy
+import fnmatch
 import re
 import os
 import logging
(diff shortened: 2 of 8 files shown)


c4t3l:
+1

This will solve my original issue!  Thanks Erik.

terminalmage:
It looks like Jenkins was in a bad state when I opened this, so since it's been a while I rebased and pushed to trigger new tests.

terminalmage:
@Ch3LL I opened this 4 months ago and had added information to the Neon release notes, assuming that this would have been merged before we branched for Neon. Since that never happened, please let me know if you would like to either A) re-point this PR at Neon (will require a rebase from me, which I'm happy to do), or B) modify this PR to add the release notes to Sodium.

doc/topics/releases/neon.rst
```diff
@@ -372,9 +372,19 @@ Module Changes
 - Added new :py:func:`boto_ssm <salt.modules.boto_ssm>` module to set and query
   secrets in AWS SSM parameters.
 
-- The :py:func:`file.set_selinux_context <salt.modules.file.set_selinux_context>`
-  module now supports perstant changes with ``persist=True`` by calling the
-  :py:func:`selinux.fcontext_add_policy <salt.modules.selinux.fcontext_add_policy>` module.
+- The :py:func:`file.set_selinux_context
+  <salt.modules.file.set_selinux_context>` module now supports perstant changes
```
waynew: s/persistant/persistent

salt/modules/config.py
```diff
@@ -130,31 +136,110 @@ def valid_fileproto(uri):
...
-        omit_pillar=False):
+        omit_all=False,
+        wildcard=False):
     '''
-    Pass in a generic option and receive the value that will be assigned
+    Returns the setting for the specified config value. The priority for
+    matches is the same as in :py:func:`config.get <salt.modules.config.get>`,
+    only this function does not recurse into nested data structures. Another
+    difference between this function and :py:func:`config.get
+    <salt.modules.config.get>` is that it comes with a set of "sane defaults".
```
waynew: Would "reasonable defaults" read better here?

salt/modules/config.py
```diff
@@ -130,31 +136,110 @@ def valid_fileproto(uri):
...
+    if omit_all:
+        omit_opts = omit_grains = omit_pillar = omit_master = True
+
+    if default is None:
+        default = '' if not wildcard else {}
+
+    if not wildcard:
+        if not omit_opts:
+            if value in __opts__:
+                return __opts__[value]
```
waynew: Would it be better to do

```
try:
    return __opts__[value]
except KeyError:
    pass
```
? (and then the same things for the later?)

Actually, if I understand this block correctly, couldn't you write it like this?

```
if not omit_opts:
    return __opts__.get(value, DEFAULTS.get(value, default))
```

Or perhaps

```
if not omit_opts:
    try:
        return __opts__[value]
    except KeyError:
        return DEFAULTS.get(value, default)
```

If you want to avoid always evaluating the `DEFAULTS.get(value, default)`.

waynew (changes requested):

salt/modules/config.py
gtmanfred: exception handling is slower than just checking if the value is in the list.

and .get is also slower.

This is the fastest implementation

gtmanfred:
```
>>> import timeit
>>> setup = '__opts__ = {"one": 1, "two": 2}'
>>> timeit.timeit('''
... if 'one' in __opts__:
...     __opts__['one']
... ''', setup=setup)
0.047407952000000364
>>> timeit.timeit('''
... __opts__.get('one', 'THING')
... ''', setup=setup)
0.07143948800000999
```

waynew: Huh! I wonder if that's something that's changed or if I've just been misremembering that `get` was faster than `in`, and it's just better than a `try/except`.

Well, TIL!

👍

Ch3LL:
ping @s0undt3ch or @twangboy any ideas why the windows tests are not finishing here?

waynew:
@Ch3LL I think it might be the changes we got merged in yesterday. @terminalmage you mind rebasing on neon, I believe that will fix the issue 🤞

terminalmage:
Given that the tests weren't even starting, and I didn't modify any files that affect Windows, I'm not sure I follow your logic. But I rebased again anyway, and sure enough, tests aren't even starting on Windows:

```
15:35:29 >>>>>>     Failed to complete #create action: [AMI 'ami-0b72a5e51f042d1bc' is pending, and cannot be run] on py2-windows-2016
```

By the way, you can see Jenkins' console output by clicking the "Console" link on the Jenkins test results.

waynew:
I've had it happen where I'm trying to merge against say commit A on branch develop, and the build fails. Commits happen on develop and now we're at commit C. Rebasing (or merging C into my branch) triggers a rebuild, which will fix whatever. I'm not sure *why* GitHub says that a branch is out of date with the original... that seems pretty weird, now that you point it out :suspect: 

Regardless, it looks like now the only problem is this known Windows failure (Pedro is looking into it right now). Hopefully that shouldn't take *too* much.

terminalmage:
> I've had it happen where I'm trying to merge against say commit A on branch develop, and the build fails. Commits happen on develop and now we're at commit C. Rebasing (or merging C into my branch) triggers a rebuild, which will fix whatever.

Again, the console was pretty clear that the failure was in just _starting_ the tests. There is actionable information that we can use, so I am honestly confused (and a little dismayed) that the plan seems to be to cross our fingers that "whatever" will be fixed by pushing and triggering another build.

> I'm not sure _why_ GitHub says that a branch is out of date with the original...

Really? You can see this change in the GitHub web interface when a pull request is merged into a branch. All the other PRs which were opened against that same branch are now out-of-date with the base branch since the base branch has commits that the head branch does not.

waynew:
Ah, I had a fuzzy mental model going on, never mind 😝

> Again, the console was pretty clear that the failure was in just starting the tests. There is actionable information that we can use, so I am honestly confused (and a little dismayed) that the plan seems to be to cross our fingers that "whatever" will be fixed by pushing and triggering another build.

Apologies - I didn't actually check the console because I had been noticing failed windows builds for the last few days, and I knew that a couple of commits had gone in to fix other windows failures. I made an (incorrect) educated guess that this windows failure was related to the other failures. It wasn't until today that I learned of *this* particular class of failures. This was bad pattern matching on my part. Probably also driven by the fact that PRs have to be brought up-to-date before merging.

Next time I'll spend the time to ensure that the build failures are what I expect them to be.

Ch3LL:
looks like the tests are passing. @waynew are we just waiting on your re-review?

waynew (approved):

PR #51685, opened 2019-02-17
INFRA-7947 - replace direct calls into boto3 utils module with __utils__[XXX]

tkwilliams:
The direct calls broke recently when the boto3 module was renamed to boto3mod, so using the canonical __utils__ path seems better here.


tkwilliams:
@rallytime Is there anything needed on this?

dwoz (approved):

PR #51695, opened 2019-02-18
Jinja regex_match can return named subgroups dict with as_dict arg

aarnaud:
Signed-off-by: Anthony ARNAUD <github@anthony-arnaud.fr>

### What does this PR do?
Enable dict return for jinja `regex_match` with `as_dict` argument

### Tests written?
No

### Commits signed with GPG?
Yes


aarnaud:
Hi @dwoz 
if you can review this small PR too ;-)

dwoz (changes requested):
@aarnaud Sorry I didn't look at this sooner. It looks good but I do think we should add a test for this case.

dwoz:
@aarnaud We are no longer accepting PRs to the `develop` branch. This needs to be re-opened against the `master` branch. You can link to this PR from the new one to preserve any disscussion history. Sorry for the confusion.

PR #51697, opened 2019-02-18
Jinja regex_match can return named subgroups dict with as_dict arg

aarnaud:
Signed-off-by: Anthony ARNAUD <github@anthony-arnaud.fr>

### What does this PR do?
Enable dict return for jinja `regex_match` with `as_dict` argument

### Tests written?
No

### Commits signed with GPG?
Yes


aarnaud:
I keep only https://github.com/saltstack/salt/pull/51695

PR #51705, opened 2019-02-19
Use salt rather than salt-call when target is set

jschmid1:
### What does this PR do?

Fix documentation for proper ext_pillar usage

### Commits signed with GPG?

Yes


dwoz (changes requested):
@jschmid1  Should this be `salt-call pillar.get external_pillar` without the `'*'`?

jschmid1:
Not neccesarily. I don't think the documentation indicates that we operate on a minion level.

dwoz (approved):

PR #51707, opened 2019-02-19
Bugfix: pillar refresh

isbm:
### What does this PR do?

Bugfix

Pillars are not fully refreshed.

### Previous Behavior

Reproducer:

1. Start master.
2. Start minion.
3. Add some pillar keypair, say `name: toto`.
4. Verify that the new pillar is ready: `salt <yourminion> pillar.items`
5. Refresh it: `salt <yourminion> saltutil.refresh_pillar`
6. Try target your minion by this pillar: `salt -I 'name:toto' test.ping` and it won't work.
7. The targeting above fully works the whole cycle only if you restart the minion (so it refreshes all the instances of the pillar)

While _targeting_ by itself works, the minion won't return the results back and the master is trying to find a running job. The reason is because the minion is verifying itself if it is properly targeted against the _old_ data. This is happening because `opts` instance of the Minion is newer than the copy of it as `__opts__` in the matchers.

For the same reasons beacons do not fully tolerate pillar refresh.

### New Behavior

Targeting and module orchestration calls is working after the `saltutil.refresh_pillar` is issued.

### Tests written?

Existing should apply.

### DIFF
--- a/salt/minion.py
+++ b/salt/minion.py
@@ -2210,6 +2210,8 @@ class Minion(MinionBase):
             finally:
                 async_pillar.destroy()
         self.module_refresh(force_refresh)
+        self.matchers_refresh()
+        self.beacons_refresh()
 
     def manage_schedule(self, tag, data):
         '''


brejoc (approved):
lgtm

thatch45 (approved):

waynew:
Not sure, but would it make sense/is it possible to write an integration test that fails in the current code base, but passes with this fix?

isbm:
@waynew this just _also_ refreshes what still has to be refreshed. LazyLoader still keeps old references to the `__opt__`, so we need to somehow invalidate them so the matchers wont lookup over the old data.

brejoc:
How should we proceed here? Is there a test needed or can it be merged like this?

thatch45:
We proceed like that :)
Yes, this is basically refreshing the loader refs

brejoc:
Thanks!

Issue #51739, opened 2019-02-20
salt.states.file.copy force=true on Windows does not clear the Read-Only attribute during removal


arizvisa:
### Description of Issue/Question
On the windows platform, when using file.copy where the source has a file with a read-only attribute (`attrib +r`), the call to `file.remove` fails with an exception and so the whole state fails.

### Setup
Setup a Windows box, drop a directory with some files that have the readonly attribute set. You can enable the +r attribute via `attrib +r $filename` at the command prompt.

### Steps to Reproduce Issue
Once you have a source directory with the prior mentioned constraints, just use a state with `file.copy` and the `force` parameter set to `true`. After the copy happens, the second time you apply the state you'll get the following exception:

```----------
          ID: Copy harness stage into target path
    Function: file.copy
        Name: R:\harness
      Result: False
     Comment: An exception occurred in this state: Traceback (most recent call last):
                File "c:\salt\bin\lib\site-packages\salt\state.py", line 1913, in call
                  **cdata['kwargs'])
                File "c:\salt\bin\lib\site-packages\salt\loader.py", line 1898, in wrapper
                  return f(*args, **kwargs)
                File "c:\salt\bin\lib\site-packages\salt\states\file.py", line 5757, in copy
                  __salt__['file.remove'](name)
                File "c:\salt\bin\lib\site-packages\salt\modules\win_file.py", line 1093, in remove
                  remove(item, force)
                File "c:\salt\bin\lib\site-packages\salt\modules\win_file.py", line 1093, in remove
                  remove(item, force)
                File "c:\salt\bin\lib\site-packages\salt\modules\win_file.py", line 1093, in remove
                  remove(item, force)
                File "c:\salt\bin\lib\site-packages\salt\modules\win_file.py", line 1093, in remove
                  remove(item, force)
                File "c:\salt\bin\lib\site-packages\salt\modules\win_file.py", line 1093, in remove
                  remove(item, force)
                File "c:\salt\bin\lib\site-packages\salt\modules\win_file.py", line 1093, in remove
                  remove(item, force)
                File "c:\salt\bin\lib\site-packages\salt\modules\win_file.py", line 1093, in remove
                  remove(item, force)
                File "c:\salt\bin\lib\site-packages\salt\modules\win_file.py", line 1093, in remove
                  remove(item, force)
                File "c:\salt\bin\lib\site-packages\salt\modules\win_file.py", line 1093, in remove
                  remove(item, force)
[...]

Issue #25250, opened 2015-07-08
'force' option in copy state deletes target file


wipfs:
```console
[root@minion]# cat /var/cache/salt/minion/files/base/example/init.sls
/tmp/test:
  file.copy:
    - source: /tmp/test_src
    - force: True
```

This will behave as expected if the target does not exist or if it is identical to the source
If however the target exists but is different from the source, then this state simply deletes the target, though it claims not to overwrite it:

```console
[root@minion]# ls -l /tmp/test /tmp/test_src
-rw-r--r--. 1 root root 192 Jul  8 12:20 /tmp/test
-rw-r--r--. 1 root root 198 Jul  8 12:19 /tmp/test_src
```
```console
[root@minion]# salt-call state.sls example
[INFO    ] Loading fresh modules for state activity
[INFO    ] Fetching file from saltenv 'base', ** skipped ** latest already in cache 'salt://example/init.sls'
[INFO    ] Running state [/tmp/test] at time 12:20:58.419543
[INFO    ] Executing state file.copy for /tmp/test
[INFO    ] The target file "/tmp/test" exists and will not be overwritten
[INFO    ] Completed state [/tmp/test] at time 12:20:58.421524
local:

          ID: /tmp/test
    Function: file.copy
      Result: True
     Comment: The target file "/tmp/test" exists and will not be overwritten
     Started: 12:20:58.419543
    Duration: 1.981 ms
     Changes:

Summary

Succeeded: 1
Failed:    0

Total states run:     1
```
```console
[root@minion]# ls -l /tmp/test /tmp/test_src
ls: cannot access /tmp/test: No such file or directory
-rw-r--r--. 1 root root 198 Jul  8 12:19 /tmp/test_src
```
```console
[root@minion-1]# salt --version
salt 2015.5.2 (Lithium)
```

jfindlay:
@wipfs, thanks for the report.

thunderk:
This is a rather critical bug as it may cause missing files, and so broken services. Maybe some unit tests would be good to avoid future regressions on this ?

jfindlay:
@thunderk, good ideas, thanks.

cachedout:
The fix and the test are both merged now. Thanks to all who helped out here! I'll go ahead and close this.

PR #51740, opened 2019-02-20
Updated the salt.states.file.copy function to make the call to file.remove with the force parameter set to the same as one for salt.states.file.copy.

arizvisa:
### What does this PR do?
On the windows platform, the `salt.state.file.copy` function does not handle files with the read-only attribute properly. When `force` is set to `true, the `file.copy` function uses `file.remove` to attempt to remove the target before copying it. This function fails with an exception when the read-only attribute is set. However when `file.remove` has its `force` parameter set to true, this gets handled properly.

This PR passes the value of the `force` parameter from `file.copy` to its call to `file.remove` so that files with the read-only attribute are handled.

This closes issue #51739.

### Previous Behavior
Previously when `file.copy` tried to remove the target directory it would fail when encountering a read-only file.

### New Behavior
Now when `force` is set to `true`, the call to `file.remove` will inherit the same value which allows `file.copy` to successfully remove the target directory. 

### Tests written?
No

### Commits signed with GPG?
No


arizvisa:
This PR seems to conflict with issue #25250. I'll have to look at it the next time I get free time, but in summary it looks like `file.copy` used to not check the differences between files before transferring them.

This fix is essentially a 12-byte patch that passes the `force` parameter to `file.remove`, and so I'm assuming that the condition that calls `file.remove` is not actually checking if the file was different or not before calling it, (or the test for #25250 itself is weird which I'd imagine is unlikely).

salt/states/file.py
```diff
@@ -6754,7 +6754,9 @@ def copy_(name,
         elif not __opts__['test'] and changed:
             # Remove the destination to prevent problems later
             try:
-                __salt__['file.remove'](name)
+                # On windows, if a file has the read-only attribute then we are unable
+                # to complete this copy unless force is set to true.
+                __salt__['file.remove'](name, force=force)
```
twangboy: You'll probably have to gate this for Windows as the Linux version of `file.remove` does not include the `force` option.

```
if salt.utls.platform.is_windows():
  __salt__['file.remove'](name, force=force)
else:
  __salt__['file.remove'](name)
endif
```
Something like that ^^^

twangboy: Or we could modify the linux `file.remove` to accept `kwargs`

twangboy (changes requested):

salt/states/file.py
twangboy: This is probably what's causing all the test failures

arizvisa: Ahh ok. Thx. I'll add this in a minute.

arizvisa: Hey, actually this pattern of calling `file.remove` without the `force` parameter exists in a number of places within this state and so I imagine a number of them will fail with the read-only attribute set.

Is it worth fixing all of these now to solve this condition? Or is it better just to fix them as bugs get reported?

arizvisa: Also I do believe that instead of adding the gate as you mentioned, that it'd be better to add the `force` parameter to `file.remove` as at some point posix acls might interfere with the api on linux (?) and so the semantics of `force` might come into existence on non-windows platforms.

It looks like some prior work also does the gating you suggested too.

arizvisa:
Ok, I added the kwargs support to `salt.modules.file.remove`. This actually fixes a number of issues in a couple of salt.states.file functions that exist due to there being no `force` parameter. So good call.

arizvisa:
@twangboy, when you get a chance next week, lmk if there's anything else I need to do to get these merged upstream as next weekend I won't be able to get back to these saltstack PRs due to them not being even closely related to my regular job.

Thanks for your help!

arizvisa:

```
$ salt --async \* state.apply
    |_
      ----------
      arg:
      fun:
          state.apply
      jid:
          20190225210028096476
      master_id:
      pid:
          508
      ret:
          etcd
      tgt:
          *
      tgt_type:
          glob
      user:
          root
bX_nxPEVRyWOcIDBECwW:
    |_
      ----------
      arg:
      fun:
          state.apply
      jid:
          20190225210028096476
      master_id:
      pid:
          2136
      ret:
          etcd
      tgt:
          *
      tgt_type:
          glob
      user:
          root


Summary
# of minions targeted: 2
# of minions returned: 2
# of minions with errors: 0
$
```

twangboy (approved):

arizvisa:
Thanks!

PR #51781, opened 2019-02-23
Add support for docker mounts with custom drivers

clinta:
### What does this PR do?

Adds the ability to specify [mounts](https://docker-py.readthedocs.io/en/stable/api.html#docker.types.Mount) with custom [driver configs](https://docker-py.readthedocs.io/en/stable/api.html#docker.types.DriverConfig) as part of the `docker.container_running` state.

### Previous Behavior

Attempting to use a state like this:
```sls
test:
  docker_container.running:
    - image: busybox:latest
    - mounts:
      - source: tank/docker-volumes/test_data
        target: /data
        type: volume
        driver_config:
          name: zfs
```

Would result in the driver_config being stripped and a local volume created instead.

### New Behavior

The driver_config section is parsed and passed through to docker-py.

### Tests written?

No. This issue is only apparent when trying to create a mount with a custom volume plugin, and I'm not sure how a driver plugin can be added to the testing environment.

### Commits signed with GPG?

Yes


dwoz (dismissed):

dwoz (commented):
@clinta.. Since this is new functionality, I think the PR should be against develop.

clinta:
This is freshly rebased on develop and passing all CI checks.

clinta:
Is passing codeclimate a new requirement for merging?

I can break  `_create_mounts` into smaller functions which if I understand it correctly should fix the conative complexity check.

But I think making a small function to remove the 5 lines of similar code in `salt/utils/docker/__init__.py` would actually reduce the readability there.

clinta:
@dwoz can you give me any guidance here? Do I need to re-write this to meet the codeclimate requirements? I've made an attempt to rewrite the duplicate code, but IMO it makes the code less readable.

And is codeclimate why the pr-head and pr-merge checks fail? Or is there something else that needs to be changed?

twangboy:
@clinta I think we need some tests for the new functionality.

clinta:
How would I go about getting a custom docker volume driver installed in the CI environment so it can be tested?

cmcmarrow:
@clinta I apologizes for that late responses. Rather than installing docker we can mock all the docker calls and verify that where sent the right `*args` and `**kwargs`. Feel free to message me if you have any questions. If you need help mocking I would give this link a read https://docs.python.org/3/library/unittest.mock.html. Also you don’t need to switch the branch to develop. This PR can stay on develop.

dwoz:
@clinta Do you still desire to get this change merged?

Ch3LL:
Just following up here to see if you would like to come back to this PR?

Ch3LL:
Closing due to inactivity. Please let me know if I need to re-open or open a new PR against master including the requested feedback.

PR #51782, opened 2019-02-23
Reduce the number of days an issue is stale by 10

garethgreenaway:


.github/stale.yml
```diff
@@ -2,7 +2,7 @@
 
 # Number of days of inactivity before an issue becomes stale
 # 480 is approximately 1 year and 4 months
```
waynew: Might want to update at least the 480 in this comment...

dwoz (approved):

Issue #51784, opened 2019-02-23
salt.utils.path.which() on the windows platform can return a non-executable file if a file without an extension already exists (breaks npm module on windows)


arizvisa:
### Description of Issue/Question
Okay...this is getting pretty fucking outrageous. This condition can happen in numerous other places, but I encountered it specifically when installing nodejs via chocolatey and then trying to use the npm modules in salt. The issue is that in order to detect the existence of npm, salt will use `salt.utils.path.which()` to locate the `npm` binary.

The `salt.utils.path.which()` function mixes up the semantics what's considered executable on Windows with what's considered executable on Posix. This mix up occurs when a script with the exact name that is being searched for exists despite the filename not ending with the correct extension. Specifically if there's two scripts..one of which is "blah.cmd" and another which is "blah", salt will return to the path to "blah" instead of "blah.cmd".

This situation results in the script w/o the extension being given priority. i.e. The node.js installer bundles two scripts in your PATH one of which is a bourne-script named `npm`, and another cmd-script named `npm.cmd`. Due to the existence of the first one, `salt.util.path.which()` returns the path to this instead of `npm.cmd` which is the correct command to run since it has an extension in `PATHEXT`.

In the node.js example which calls `salt.utils.path.which('npm')`, it'll log the following error due to `npm` not being being executable from the windows shell.

```
2019-02-23 19:21:06,789 [salt.state       :320 ][ERROR   ][1356] State 'npm.installed' was not found in SLS 'develop'
Reason: 'npm' __virtual__ returned False: 'npm' binary not found on system
```

### Setup
On the Windows platform, drop two files named `test` and `test.cmd` in the search PATH. If you want to execute something, put some cmd-script that redirects some text into a logfile or something in either.

### Steps to Reproduce Issue
After getting two files in your path, simply call `salt.utils.path.which()` and it will return the path to the `test` file instead of the one with the correct extension. As mentioned, this is wrong as `test` cannot be executed from the command shell due to its extension not existing in PATHEXT.

To prove that you can't execute the first script, try passing it to os.system or something.

### Versions Report
```
Salt Version:
           Salt: 2018.3.3
 
Dependency Versions:
           cffi: 1.10.0
       cherrypy: 10.2.1
       dateutil: 2.6.1
      docker-py: Not Installed
          gitdb: 2.0.5
      gitpython: 2.1.3
          ioflo: Not Installed
         Jinja2: 2.10
        libgit2: Not Installed
        libnacl: 1.6.1
[...]

PR #51785, opened 2019-02-23
Re-implemented salt.utils.path.which to properly define executable semantics on both the windows and posix platforms.

arizvisa:
### What does this PR do?
The original `salt.utils.path.which()` function had a number of issues, one of which resulted in #51784. The original implementation did not appear to properly isolate windows and posix semantics which is why the issue came into existence. Another probably non-important issue is that memoization is attempted to be used, but on a function with no arguments which results in the performance benefit not being able to be taken advantage of. I assume the original author intended to use memoization to improve the peformance of the regular expression.

In essence, this PR is the same as the following code but with the ability to follow links.
```
executable = lambda(_): os.path.isfile(_) and os.access(_,os.X_OK)
which = lambda _,envvar="PATH",extvar='PATHEXT':_ if executable(_) else iter(filter(executable,itertools.starmap(os.path.join,itertools.product(os.environ.get(envvar,os.defpath).split(os.pathsep),(_+e for e in os.environ.get(extvar,'').split(os.pathsep)))))).next()
```

This fixes issue #51784, and also re-implements some of the optimizations that the original author desired.

### Previous Behavior
In some instances (see #51784), the `salt.utils.path.which()` function would return non-(shell)-executable programs on the Windows platform, and also not benefit from memoization used to reduce the number of times a regular expression is matched (which it appears that the original author considered expensive)

### New Behavior
Now it splits up the posix and windows logic so that the windows implementation properly searches the path for executables according to the PATHEXT environment variable, posix version will follow links in order to identify whether the link is executable, and the matching of PATHEXT is done in constant-time without using regexes.

### Tests written?
No

### Commits signed with GPG?
No


arizvisa:
Some of these tests need to be punted as they're dying due to a `MemoryError` in code unrelated to mine.

arizvisa:
Still something to fix... just a moment here.

arizvisa:
Force-pushed as the commit specified the wrong module path.

arizvisa:

```
$ salt --async \* state.apply
    |_
      ----------
      arg:
      fun:
          state.apply
      jid:
          20190225210028096476
      master_id:
      pid:
          508
      ret:
          etcd
      tgt:
          *
      tgt_type:
          glob
      user:
          root
bX_nxPEVRyWOcIDBECwW:
    |_
      ----------
      arg:
      fun:
          state.apply
      jid:
          20190225210028096476
      master_id:
      pid:
          2136
      ret:
          etcd
      tgt:
          *
      tgt_type:
          glob
      user:
          root


Summary
# of minions targeted: 2
# of minions returned: 2
# of minions with errors: 0
$
```

arizvisa:
Bah, this one also needs its tests punted as they're having that `saltutil.is_running` error again...

I patched the unit-tests to include an assignment to os.pathsep since it's ':' on linux, and also changed some of the side-effects of os.access since the original implementation had one more syscall than was necessary.

arizvisa:
Wow, am I reading this right? All the tests passed, yay.

arizvisa:
Thx @twangboy

dwoz (approved):

PR #51799, opened 2019-02-25
Catch attribute error

s0undt3ch:
### What does this PR do?
Catches an `AttributeError` found when running `salt-call --local` on Windows.

### Tests written?

No

### Commits signed with GPG?

Yes


twangboy:
Is this also a problem on 2018.3?

s0undt3ch:
@twangboy No.

Issue #51788, opened 2019-02-24
Typo in documentation of _check_cache_minions function


ritishverma:
https://github.com/saltstack/salt/blob/61baab97de511aab9273446062430f02623f5d2d/salt/utils/minions.py#L275

There is a typo in this line in the documentation of function _check_cache_minions.
It should say 'absent' and not 'absend'
This typo exists in all branches.

dwoz:
@ritishverma Thanks for reporting this.

PR #51800, opened 2019-02-25
Fix typo in docstring

dwoz:
Fix typo #51788

### Tests written?

No

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/utils/minions.py
+++ b/salt/utils/minions.py
@@ -256,9 +256,10 @@ class CkMinions(object):
                              regex_match=False,
                              exact_match=False):
         '''
-        Helper function to search for minions in master caches
-        If 'greedy' return accepted minions that matched by the condition or absent in the cache.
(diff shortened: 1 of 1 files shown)


garethgreenaway (approved):

salt/utils/minions.py
```diff
@@ -257,7 +257,7 @@ def _check_cache_minions(self,
                              exact_match=False):
         '''
         Helper function to search for minions in master caches
-        If 'greedy' return accepted minions that matched by the condition or absend in the cache.
+        If 'greedy' return accepted minions that matched by the condition or absent in the cache.
```
cachedout: This could benefit from a slight rewrite:

`If 'greedy', then return accepted minions matched by the condition or those absent from the cache.`

dwoz: Good call, I've updated this docstring.

PR #51804, opened 2019-02-25
bug: status might be a string, coerce it

mattp-:
some backends (requests) return status code as an int, make sure its
coerced as string before doing a comparison.

### What does this PR do?
fix status match when using requests backend. 

n/a

### Tests written?
No

### Commits signed with GPG?

Yes/No

### DIFF
--- a/salt/states/http.py
+++ b/salt/states/http.py
@@ -123,7 +123,7 @@ def query(name, match=None, match_type='string', status=None, status_type='strin
 
     if status is not None:
         if status_type == 'string':
-            if data.get('status', '') == str(status):
+            if str(data.get('status', '')) == str(status):
                 ret['comment'] += ' Status {0} was found.'.format(status)
                 if ret['result'] is None:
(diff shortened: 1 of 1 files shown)


cachedout (approved):
👍

salt/states/http.py
```diff
@@ -123,7 +123,7 @@ def query(name, match=None, match_type='string', status=None, status_type='strin
 
     if status is not None:
         if status_type == 'string':
-            if data.get('status', '') == str(status):
+            if str(data.get('status', '')) == str(status):
```
dwoz: @mattp- Can we use `salt.utils.stringutils.to_str_here`?

mattp-: this actually is a bit of an issue as status can be an int which to_str seems to not coerce. does it make sense to use ?

dwoz: @mattp- Yes, makes sense.

Issue #51811, opened 2019-02-25
Umm...wow. The salt.states.npm module is not checking for the existence of npm correctly.


arizvisa:
### Description of Issue/Question
The `salt.states.npm` module doesn't work because it's checking for npm support incorrectly. Back in 2017, some guy renamed the `salt.modules.list` function to `salt.modules.list_` which broke the npm module entirely.

The reason being is that in order to verify that npm support is available, the following check is made:
```
def __virtual__():
    '''
    Only load if the npm module is available in __salt__
    '''
    return 'npm' if 'npm.list' in __salt__ else False, '\'npm\' binary not found on system'
```

Due to this states module checking for `npm.list` instead of the correct `npm.list_`, the npm state module can't be used.

### Setup
This apparently has been (not) working this way since 2017-06-04. Write up a state that uses `npm.installed` as you will use it to repro this issue.

### Steps to Reproduce Issue
Once you apply the state, you will see that it claims it's unable to locate the npm binary. This is a misnomer because if you try and exec `cmd.which` for npm, it will totally find it (assuming you use my PR #51785 if you're on Windows)

### Versions Report

This is in the develop branch, there's no need for versions here because you can literally see the code discrepancy. As a review look at `salt.states.npm` and then `salt.modules.npm.`

`salt.states.npm:31`
```
def __virtual__():
    '''
    Only load if the npm module is available in __salt__
    '''
    return 'npm' if 'npm.list' in __salt__ else False, '\'npm\' binary not found on system'
```

`salt.modules.npm:237`:
```
def list_(pkg=None, dir=None, runas=None, env=None, depth=None):
    '''
    List installed NPM packages.

    If no directory is specified, this will return the list of globally-
    installed packages.

    pkg
        Limit package listing by name
```

arizvisa:
Jesus christ guys. Will write up another PR in a sec...

arizvisa:
Closed by PR #51812.

arizvisa:
Okay. So I was able to track this down while working on the PR. It's a windows-specific issue and it seems like a common construct. Will write further results on analysis in PR #51812.

arizvisa:
Okay, the PR has been re-created PR #51812 is deprecated in favor of PR #51813. This properly fixes the version check on the Windows platform and has been tested.

[...]

PR #51812, opened 2019-02-25
Fixed the salt.modules.npm module to check the npm version correctly on the Windows platform

arizvisa:
### What does this PR do?
The `salt.states.npm` module verifying that `npm` support is available in the wrong way. It is checking for a function (`salt.modules.npm.list()`) that does not exist. Halfway through 2017, this function was renamed to `list_()` but for some reason this module was not updated.

Instead of fixing this issue by renaming the check to `list_`, we opt to use `install` as due to it being a not-so-generic name it's unlikely to change during another refactor. The error message that is emitted is also rephrased in order to be more clear that it's not that the `npm` binary is missing, but that it's unable to load the `salt.modules.npm` module.

This closes issues #51811.

### Previous Behavior
Prior to this fix, the `salt.states.npm` module could never work (since 2017-06-04.

### New Behavior
Now that it's checking for a function that exists, it should behave correctly and all the states/functions in `salt.states.npm` should work. By checking for `install` instead of `list`, this will hopefully survive a refactor as renaming `install` appears to be a more serious issue than renaming something as general as `list`.

There needs to be a better/safer to check for a successful vs failed module load to avoid these issues in the future, but for now we'll just hope that nobody renames `install` to something else...

### Tests written?
No, but if someone else had actually written tests here and verified them before commit 384570384e then that would've blocked that commit until things like this worked.

### Commits signed with GPG?
No

### DIFF
--- a/salt/modules/npm.py
+++ b/salt/modules/npm.py
@@ -49,10 +49,13 @@ def _check_valid_version():
     Check the version of npm to ensure this module will work. Currently
     npm must be at least version 1.2.
     '''
+
+    # Locate the full path to npm
+    npm_path = salt.utils.path.which('npm')
+
     # pylint: disable=no-member
-    npm_version = _LooseVersion(
-        salt.modules.cmdmod.run('npm --version', output_loglevel='quiet'))
-    valid_version = _LooseVersion('1.2')
+    res = salt.modules.cmdmod.run('{npm} --version'.format(npm=npm_path), output_loglevel='quiet')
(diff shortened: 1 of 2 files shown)


arizvisa:
Demoting this to WIP as that now it's locating the function name (despite the alias), it's encountering an issue with the version check.

arizvisa:
Actually originally I was going to append to this PR, but I'm going to close it as it turns out that it's a Windows-specific issue.

Issue #10723, opened 2014-02-25
Controlled fail in state definitions


cznewt:
Hello, puppet has mechanism to deliberately fail if some conditions are not met (i.e. os_family is not supported) like:

fail("No support for OSfamily \"${::osfamily}\"")

does salt have something similar, it would be very useful

powellchristoph:
+1 I would love to have this functionality.

whiteinge:
I think we have answers to all the asks in this thread now with the upcoming 2015.8 release of Salt. Is there anything anyone would still like to see or can we close this issue out? We should add a "Throwing and Handling Errors" section to the states/formulas docs that lists each approach.

Reacting to, or recovering from, a failed state.

```yaml
create_lots_of_files:
    cmd.run:
        - name: for i in {1..99999}; do touch mydirectory/$i ;done

"Unable to create tmp files. Check disk space?":
  cmd.run:
    - name: rm mydirectory/*
    - onfail:
      - cmd: create_lots_of_files
```

Throwing an exception during the Pillar compilation phase:

```jinja
{% if salt['file.exists']('/path/that/must/exist') %}
...
{% else %}
{{ salt.test.exception("/path/that/must/exist does not exist!!") }}
{% endif %}
```

Requiring certain Pillar values (or Pillar types) with the [test.check_pillar](http://docs.saltstack.com/en/develop/ref/states/all/salt.states.test.html#salt.states.test.check_pillar) state function:

```yaml
is-pillar-foo-present-and-bar-is-int:
  test.check_pillar:
    - present:
        - foo
    - integer:
        - bar
```

powellchristoph:
@whiteinge Thank you for the update. The test.check_pillar is very nice!

mitar:
`{{ salt.test.exception("/path/that/must/exist does not exist!!") }}` to me does not work. It seems like it returns the value with `_error` set to "Failed to return clean data" but it does not fail the compilation of the whole state.

whiteinge:
@mitar Sorry for the bad wording above. Throwing an exception will stop the parsing of only the .sls file containing that error when generating Pillar. It will stop the full execution of a state run, however.

mitar:
So maybe I am not testing this correctly. But i have a state `test.sls` like:

```
group-users:
  group.present:
    - gid: |
        {{ salt.test.exception("/path/that/must/exist does not exist!!") }}
```

I call:

```
salt-ssh '*' state.show_sls test
```

And I get:

```
testserver:
[...]

PR #51820, opened 2019-02-26
Document controlled failure in state files

cebe:
### What does this PR do?

Added documentations for the solution pointed out by @bwesemann-cxp here https://github.com/saltstack/salt/issues/10723#issuecomment-449181053

https://github.com/saltstack/salt/issues/10723

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/states/test.py
+++ b/salt/states/test.py
@@ -43,6 +43,24 @@ calls, e.g. running, calling, logging, output filtering etc.
             - foo
         - integer:
             - bar
+
+You may also use these states for controlled failure in state definitions, for example if certain conditions in
+pillar or grains do not apply. The following state definition will fail with a message "OS not supported!" when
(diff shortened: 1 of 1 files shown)


Ch3LL (changes requested):
thanks for this doc addition :) looks like there is a doc failure that needs to be cleaned up

cebe:
anything to do or change from my side?

Ch3LL:
yeah if you could clean up the doc build failure. since you are using jinja in your code block you need to change the code block from yaml to jinja. That should clean up that error. If you want to test your change before pushing you can try to build the docs locally using these instructions: https://docs.saltstack.com/en/latest/topics/development/conventions/documentation.html#docs-building

cebe:
@Ch3LL thanks for the hint, changed the code block type to `jinja` and rebased on `develop`, ready to merge from my side.

Issue #51604, opened 2019-02-12
[Feature Request] Use modules in requisites


Oloremo:
### Description of Issue/Question

Saltstack is historically and by design is not about working with runtime information. One of the features that'd like to see is a requisite which are based on some runtime info. 

Here is an example:
I want to run some state but only if the result of the different module is True.

It may look like this:
```
cmd.run:
  <...>
  - onlyif:  my_custom_module_which_returns_true_or_false
```
And I want to do it in **runtime**, not before via Jinja execution of the Salt.

I think it's already possible with **slots** and **onlyif** in a way that if slots function will return a string "true" or "false", **onlyif** will execute it on the system and make a decision based on this.

The executing a shell part is really not necessary and this whole approach is hacky so if we may have a new requisite module just for this that'd be really great.

This idea is inspired by [Ansible's condition engine](https://docs.ansible.com/ansible/latest/user_guide/playbooks_conditionals.html), which allows doing this:

```
name: Run command
command: "ls -1 /tmp"
register: result_of_command

name: Do something if you should
when: "'file123' in result_of_command.stdout_lines"
module_name:
<...>
```

Ch3LL:
will approve as a feature request. thanks :)

gtmanfred:
This is a duplicate of #5050

And i am working on implementing it right now.

gtmanfred:
PR https://github.com/saltstack/salt/pull/51846

Issue #5050, opened 2013-05-16
Add unless/ifonly params to module.run/wait + more granular unless/ifonly result evluation


ryba-xek:
> NOTE (added by @basepi): All states now support `unless` and `onlyif`, this issue is still open for the additional request for more power in these statements than just `cmd.retcode`.

Salt state scripting will be _much_ more useful if one can use `module.run/wait` states with `unless` and `ifonly` params.

The whole idea of salt is the aim _not_ to use console commands in state.
So, the whole internal command stack was developed — called modules.
But why `cmd.run/wait` have `unless/ifonly` and `module.run/wait` doesn't?..

The second idea is to somehow to be able to control what `unless/ifonly` module commands are considered having positive result.

I mean for example I want smth like this:

```
debconf.set:
    module.run:
    - package: mysql-server-5.5
    - question: 'mysql-server/root_password'
    - type: 'password'
    - value: {{pillar['mysql.pass']}}
    - unless:
        command: pkg.version   
            - names:
                - mysql-server-5.5
```

Here `pkg.version` command returns empty string if not installed, this should be considered `False`. Any non-empty result should be string.

I will be happy if I will be ablo to use more complex logic like
`- unless: pkg.available_version mysql-server < 5.5`

sivann:
This still does not work.

gtmanfred:
That is not what it is being talked about.

Passing modules to unless and onlyif, instead of just bash commands is the request.

afletch:
How far did this get, can modules currently be passed to unless/onlyif? If so, what is the format for this?

gtmanfred:
no work has been done on this afaik

gtmanfred:
Work done here https://github.com/saltstack/salt/pull/51846

PR #51846, opened 2019-02-26
Allow using salt modules in onlyif and unless

gtmanfred:
### What does this PR do?
Allow using salt modules in onlyif and unless

Implements #5050 
Implements #51604
### Previous Behavior
Can only use commands that run through cmd.retcode

### New Behavior
Can use any salt module, and the result is determined by the trueish or falsish return of the module.

### Tests written?

Yes

### Commits signed with GPG?

Yes


gtmanfred:
@saltstack/team-core can I get a review on this.

Thanks!
Daniel

salt/state.py
```diff
@@ -883,19 +883,31 @@ def _run_check_onlyif(self, low_data, cmd_opts):
...
+                                'skip_watch': True,
+                                'result': True})
+                elif cmd == 0:
+                    ret.update({'comment': 'onlyif condition is true', 'result': False})
+            elif isinstance(entry, dict):
+                if 'fun' not in entry:
+                    ret['comment'] = 'no `fun` argument in onlyif: {0}'.format(entry)
+                    log.warning(ret['comment'])
+                    return ret
+                result = self.functions[entry.pop('fun')](**entry)
```
mattp-: nice 👍 i'd suggest adding a check for __context__['retcode'] and taking that over the return result if set

gtmanfred: ahh good point.

gtmanfred: going to make it only care about if retcode is anything but 0.

gtmanfred: since otherwise, we would always be 0, because the salt minion sets it to 0 if nothing is set.

gtmanfred: Done!

gtmanfred:
Ok, this should be good to merge 😄

dwoz (approved):

mchugh19:
This is pretty major and should get a shout out in the release notes too

max-arnold:
@gtmanfred Unfortunately, this doesn't work consistently across all states and needs to be improved. The following states have their own `mod_run_check` implementations:

* `states.git`
* `states.cmd`
* `states.macpackage`
* `states.file`
* `states.docker_container`

This is where unless/onlyif get skipped: https://github.com/saltstack/salt/blob/develop/salt/state.py#L1955-L1957

I discovered this while trying to fix slots #53307

mchugh19:
> The following states have their own `mod_run_check` implementations

In glancing over those states, it's not immediately obvious why they have their own `mod_run_check` implementations? Would functionality change if they were just removed?

gtmanfred:
They had unless and onlyif implementations before unless and onlyif was implemented, and unless/onlyif was a simpler implementation, so instead of having to make compatible changes for everything that the cmd.run onlyif did, i just let it be configured to be able to customize it.

Issue #51824, opened 2019-02-26
schedule.present state not idempotent in 2019.2.0


OrangeDog:
Master and minion 2019.2.0 (with #51265).
Every alternating application of this state reports changes.
```
highstate:
  schedule.present:
    - enabled: true
    - function: state.apply
    - job_kwargs:
        saltenv: base
    - minutes: 10
    - returner: zabbix
    - run_on_start: true
```
```
          ID: highstate
    Function: schedule.present
      Result: True
     Comment: Modifying job highstate in schedule
     Started: 12:34:57.835010
    Duration: 27.769 ms
     Changes:
              ----------
              diff:
                  ---
                  +++
                  @@ -1,10 +1,9 @@
                   enabled:True
                   function:state.apply
                   jid_include:True
                  -kwargs:{'saltenv': 'base'}
                  +kwargs:OrderedDict([('saltenv', 'base')])
                   maxrunning:1
                   minutes:10
                   name:highstate
                   returner:zabbix
                   run_on_start:True
                  -splay:None
```

max-arnold:
Duplicate of #51467

garethgreenaway:
Closing this out as the fix is in the 2019.2 and later branches, and will be available in the next minor releases. If the problem persists, let us know and we can reopen this issue.

PR #51849, opened 2019-02-26
[2019.2] Fix to schedule.list function to filter out attributes that are None

garethgreenaway:
### What does this PR do?
Splay defaulting to None internally, broke schedule.present causing it to always report differences when a state is run.  Updating the schedule.list function to check if an attribute is None and remove it.

#51824 #51647


sjorge (approved):
LGTM

waynew (approved):
Makes sense to me 👍

PR #51870, opened 2019-02-27
Revert "Removing options from libcrypto.OPENSSL_init_crypto..."

s0undt3ch:
Revert "Removing options from libcrypto.OPENSSL_init_crypto so it will simply use the defaults."

This reverts commit 1831e645fa104a2ec59b16c2a309981f7be14cad.

This change throws core dumps on at least Ubuntu 18.04.


s0undt3ch:
A simple sleep right before `libcrypto.OPENSSL_init_crypto()` also solves the core dump.

So this PR is probably not the best choice and can be closed.

dwoz:
@s0undt3ch We can close this since the other fix #51875 was merged?

s0undt3ch:
Yup.

Issue #51872, opened 2019-02-27
cron state and module not working correctly with salt running from non-root user


Oloremo:
### Description of Issue/Question
Salt support running from **non-root user**. When I trying to do a `cron.present` state I have to specify a `user` arg, it doesn't make much sense from non-root user but ok, I set it to the same user Salt is running from and state acts weird:

1) It trying to execute cron commands from this user using a `runas` in many functions which doesn't make any sense and won't work in non-root deployment.
2) it fails at 1) yet it somehow updates the user cron
3) It does that in non-idempotent way, so this state always produce a change.

Other observations:
If we look at the function `write_cron_file` in cron module we cant see a strange condition:
```
    if _check_instance_uid_match(user) or __grains__.get('os_family') in ('Solaris', 'AIX'):
        return __salt__['cmd.retcode'](_get_cron_cmdstr(path),
                                       runas=user,
                                       python_shell=False) == 0
    else:
        return __salt__['cmd.retcode'](_get_cron_cmdstr(path, user),
                                       python_shell=False) == 0
``` 
`if _check_instance_uid_match(user)` actually compares the user in state with runtime user of salt and, for some reason, if they're the **same** it decides that it need to do a `runas`. But that makes no sense! If they're the same you can actually just fire the `crontab` command and it will be executed from current user - no need for user switch! 

Kinda the same problem I found a few other functions.

One more thing I'd like to add is that cron module and state doesn't have logging at all so debugging them is very hard.

### Setup
Salt running from non-root user.

### Steps to Reproduce Issue
Start Salt from non-root user
Apply `cron.present` state for this user.

### Versions Report
```
Salt Version:
           Salt: 2019.2.0

Dependency Versions:
           cffi: Not Installed
       cherrypy: Not Installed
       dateutil: Not Installed
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.10
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.6.1
   mysql-python: Not Installed
      pycparser: Not Installed
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: Not Installed
         Python: 3.6.6 (default, Aug 13 2018, 18:24:23)
   python-gnupg: 0.4.4
         PyYAML: 3.13
[...]

PR #51873, opened 2019-02-27
Fix cron module logic regarding working from non-root user

Oloremo:
### What does this PR do?
Fix cron module\state execution in case of Salt running from non-root user

Fixes https://github.com/saltstack/salt/issues/51872

### Previous Behavior
cron state trying to run commands using the privilege escalation via `runas` arg in different modules.

### New Behavior
If the current Salt process owner UID equals specified user in cron state it won't do a privilege escalation.

### Tests written?

No

### Commits signed with GPG?

Yes


garethgreenaway:
@Oloremo Thanks for the PR.  I think it might make more sense to break out the other if...else situations as you did for the `raw_cron` function in the others, to ensure that the Solaris/AIX functionality remains consistent.

Oloremo:
@garethgreenaway I felt like I didn't break any logic and `raw_cron` func is just a bit different from the other, but if you like I can make them in the same style

Oloremo:
Ok, so I tried to do some unification of the logic and updated tests according to it.
Problem is now I have new kind of condition which isn't covered by tests - when I check if salt running from a root user:

`_check_instance_uid_match('root'):`

And I'm not sure how to mock it properly to test this condition.

Oloremo:
@garethgreenaway mind to look at this PR again?

garethgreenaway (approved):

Issue #51898, opened 2019-02-28
SSDs grain reports Device Mapper (LVM) devices as SSDs


wido:
### Setup
Linux hosts running Ceph OSDs deployed with ceph-volume which creates Logical Volumes on top of these SSDs.

<pre>salt 'srv-*03-07*' grains.item SSDs</pre>

This returns:

<pre>
    SSDs:
        - sda
        - sdb
        - sdc
        - sdd
        - sde
        - sdf
        - sdg
        - sdh
        - sdi
        - sdj
        - sdk
        - dm-0
        - dm-1
        - dm-2
        - dm-3
        - dm-4
        - dm-5
        - dm-6
        - dm-7
        - dm-8
        - dm-9
</pre>

I checked the code (commit 4cd0680b7f48790cc82a8074695ef8f04b99c74a) and I see that salt/grains/disks.py contains this:

<pre>
def _linux_disks():
    '''
    Return list of disk devices and work out if they are SSD or HDD.
    '''
    ret = {'disks': [], 'SSDs': []}

    for entry in glob.glob('/sys/block/*/queue/rotational'):
        try:
            with salt.utils.files.fopen(entry) as entry_fp:
                device = entry.split('/')[3]
                flag = entry_fp.read(1)
                if flag == '0':
                    ret['SSDs'].append(device)
                    log.trace('Device %s reports itself as an SSD', device)
                elif flag == '1':
                    ret['disks'].append(device)
                    log.trace('Device %s reports itself as an HDD', device)
                else:
                    log.trace(
                        'Unable to identify device %s as an SSD or HDD. It does '
                        'not report 0 or 1', device
                    )
        except IOError:
            pass
    return ret
</pre>

It only checks in */sys/block/X/queue/rotational* if this reports 1 or 0 and for a Device Mapper device it's 0, but this is a virtual device:

<pre>
[salt@ceph-salt (GN3) ~]$ salt 'XXXX' cmd.shell 'cat /sys/block/dm-9/queue/rotational'
srv-XXX03-07.XXXX:
    0
[salt@ceph-salt (GN3) ~]$
</pre>

There should probably be a better check at this point to check if these devices are real and not device mapper devices.

### Versions Report
<pre>
Salt Version:
           Salt: 2018.3.3
 
Dependency Versions:
           cffi: 1.6.0
       cherrypy: unknown
       dateutil: 1.5
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
[...]

PR #51899, opened 2019-02-28
disk grain: Do not report Virtual Device as SSDs

wido:
Device Mapper (LVM), Bcache and VirtIO devices are reported as
SSDs by this grain, but since they are virtual we can not report
them as SSDs.

### What does this PR do?
The SSDs grain no longer reports Virtual Devices like LVM, Bcache and VirtIO as SSDs.

It fixes issue #51898 

### Previous Behavior

<pre>
    SSDs:
        - sda
        - sdb
        - sdc
        - sdd
        - sde
        - sdf
        - sdg
        - sdh
        - sdi
        - dm-0
        - dm-1
        - dm-2
        - dm-3
        - dm-4
        - dm-5
        - dm-6
        - dm-7
</pre>

### New Behavior
<pre>
    SSDs:
        - sda
        - sdb
        - sdc
        - sdd
        - sde
        - sdf
        - sdg
        - sdh
        - sdi
</pre>

### Tests written?
No

### Commits signed with GPG?
Yes


Ch3LL (changes requested):
thanks for the PR! would you mind adding a test so this does not regress in the future?

wido:
Where would I add these tests? Unit Testing isn't my strongest point and I'm not completely sure with how this works within Salt.

Ch3LL:
there is some documentation here: https://docs.saltstack.com/en/latest/topics/tutorials/writing_tests.html on the test runner which should add some more context on how to add tests for salt.

The best place to add this test would be tests/unit/grains/test_disks.py

dwoz:
@wido I've rebased this on top of master and changed the base branch of the PR, this still need tests though.

PR #50197, opened 2018-10-24
add saltutil state to wrap saltutil.sync* modules

mchugh19:
### What does this PR do?
Adds saltutil state to wrap the `saltutil.sync_*` modules.

### What issues does this PR fix or reference?
#45264

### Previous Behavior
`salutil.sync-*` was run in states via `module.run`, and always ran and reported changes

### New Behavior
new saltutil.sync_* states properly report if changes occured supporting on_changes

### Tests written?
Yes

Issue #45264, opened 2018-01-04
Feature request for saltutil.synced_all and related state modules


boltronics:
### Description of Issue/Question
Currently if I need to sync custom states, modules, grains, etc. and have that triggered from a state sls, I need to wrap the request in a `module.run` block. This has the disadvantage of triggering a system "change" every time that state sls executes, even if everything was already in sync.

I feel this is a common (possibly the most common) use case for using `module.run`, but it's far from ideal.

### Setup
I would like to replace something like:
```
update_all:
  module.run:
    - saltutil.sync_all:
      - refresh: True
```
(assuming `use_superseded` for `module.run` is in play) with:
```
update_all:
  saltutil.synced_all:
    - reload_modules: True
```
The output would ideally show only modules that needed to be synced, or otherwise indicate no changes were necessary to the system. I would expect `saltutil.synced_modules`, `saltutil.synced_states`, etc. be implemented similarly.

### Steps to Reproduce Issue
N/A

### Versions Report
Not implemented as of Salt 2017.7.2 (Nitrogen).

Ch3LL:
I think this is a great idea. We are more than willing to accept a PR for this if you want to give it a go

boltronics:
Thanks. I'd be happy to give it a go when I find enough free time. Probably not any time soon though, so equally happy if someone else beats me to it. :wink:

max-arnold:
@boltronics The linked PR was merged. Perhaps this issue could be closed?

boltronics:
@max-arnold Sorry I didn't notice this previously. It looks like @mchugh19 has done wonderful work. Thanks!

I don't believe I have privileges to do so, but I'm happy for this to be closed.

boltronics:
Oh I do. :wink:

PR #50633, opened 2018-11-26
Loader docs

AstraLuma:
### What does this PR do?

Add documentation about module loading and in general attempts to improve documentation about writing modules for salt.

### What issues does this PR fix or reference?

#50594 

### Notes

I already know that there's changes for the 2018.3 branch. I'm not sure how to handle that?

* Add token modules
* File server modules can now be loaded from the file server
* Search system was removed
* Documentation for Executors was added.

For flourine:
* matcher modules added https://github.com/saltstack/salt/blob/develop/doc/topics/matchers/index.rst

PR #51900, opened 2019-02-28
Add missing saltutil.sync_* states

max-arnold:
### What does this PR do?

A follow-up patch for #50197 (issue #45264). Cc: @mchugh19 

* Add missing `saltutil.sync_matchers` and `saltutil.sync_serializers` states 
* Add `states.saltutil` docs
* Also there is a test to make sure this won't happen again.

Question: does it make sense to also ensure function parity in the `saltutil` runner? https://docs.saltstack.com/en/develop/ref/runners/all/salt.runners.saltutil.html. The missing functions are: `sync_beacons`, `sync_engines`, `sync_executors`, `sync_log_handlers`, `sync_matchers`.

### Tests written?

Yes

### Commits signed with GPG?

No


tests/unit/states/test_saltutil.py
```diff
@@ -45,6 +48,8 @@ def test_saltutil_sync_all_nochange(self):
             'proxymodules': [],
             'output': [],
             'pillar': [],
+            'matchers': [],
+            'serializers': [],
```
max-arnold: @mchugh19 Could you please clarify when these tests are supposed to fail? For me they pass even if I remove these two lines (`matchers` and `serializers`).

dwoz: @mchugh19 ping.

max-arnold:
@dwoz Also, it looks like the docs build process doesn't clean removed files. One example is https://docs.saltstack.com/en/develop/ref/file_server/dynamic-modules.html, which was deleted in #50633

tests/unit/states/test_saltutil.py
mchugh19: Just took a look, and it appears that as long as there are no changes to sync, the test succeeds. So for example, if pillar had data: `'pillar': ['somedata'],` that causes the test to fail (since state would interpret that there are changes to sync, when it should return no changes.)

Thus, as of now, the test is only validating that the state evaluates empty sync types as not having any changes. It does not otherwise understand what types *should* be synced. I'm not sure if having that logic included is useful or not. In its current form, the sync state walks the module output looking for changes so it's really only the sync modules that need to be updated with additional types.

max-arnold:
@dwoz I got the answer to my question (and the docs link seems to be fixed now), so the PR could be merged.

max-arnold:
@dwoz I believe this can be merged

dwoz (approved):

Issue #36397, opened 2016-09-18
RFE: postgres_permissions should support ALTER DEFAULT PRIVILEGES


jdelic:
As mentioned in #6558, which hasn't been updated in a long time, the PostgreSQL states (specifically `salt.states.postgres_privileges`) should also allow `ALTER DEFAULT PRIVILEGES` queries. This is necessary since PostgreSQL's `GRANT` implementation only changes privileges on _already existing_ objects in a database/schema. `ALTER DEFAULT PRIVILEGES` makes it so that a user will gain the defined privileges on _all future objects_ as well.

Using Salt to manage database owners and read-only database roles is pretty much impossible currently as you can only grant permissions on already existing objects, thereby creating a chicken/egg problem where Salt has to create a database owner which then creates the tables and then Salt has to run again for the read-only user to gain access to the newly created tables (this is assuming that Salt does not also create the tables).

A simple boolean flag `alter_default_privileges` which then issues _both the `GRANT` and `ALTER DEFAULT PRIVILEGES` query_, would probably solve this problem just fine.

Docs:
https://www.postgresql.org/docs/9.5/static/sql-alterdefaultprivileges.html

lsh-0:
hi guys - has there been any movement since Sept 2016?

mattiasr:
I would like to see this, since i needed it today

koalaurent:
Hello, no news about this option since 2016 ? Thanks.

emerichunter:
Forked the repo today. I need the feature too so I'm looking into it. Any advice on implementation would be welcome.

emerichunter:
Forked the repo today. I need the feature too so I'm looking into it. Any advice on implementation would be welcome.

emerichunter:
Quick update. Managed to implement DEFAULT PRIVILEGES for TABLES, SEQUENCES and FUNCTIONS. In doing so, I also stumbled on an error from a SQL query checking the owner from the object. Will patch this on if it is not already done. GROUP will be done also most likely.

PR #51904, opened 2019-02-28
Add function postgres_default_privileges to grant and revoke default privileges to users in postgres instances

emerichunter:
### What does this PR do?

Add function postgres_default_privileges to grant and revoke default privileges to users in postgres instances

The objects in a schema (with or without group) covered are:
- functions
- tables
- sequences

See
https://www.postgresql.org/docs/current/sql-alterdefaultprivileges.html

issue #36397

### New Behavior
Add default privileges for DBAs

### Tests written?

Yes

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/modules/postgres.py
+++ b/salt/modules/postgres.py
@@ -90,6 +90,28 @@ _PRIVILEGES_MAP = {
     'd': 'DELETE',
     '*': 'GRANT',
 }
+_DEFAULT_PRIVILEGES_MAP = {
+    'a': 'INSERT',
+    'C': 'CREATE',
+    'D': 'TRUNCATE',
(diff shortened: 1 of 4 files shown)


dwoz (approved):

eriko:
Any idea on when then will make it in to master or better yet a release version?

emerichunter:
It doesn't seem like a priority. Standard privileges are too limited, this should be pushed. 

But I'm not the one to decide. 

I have more features to add, but if this does not see the light of day, I'll just stop entirely using SALT.

evkuzin:
FWIW - this bit is not full. I cant see `ALTER DEFAULT PRIVILEGES FOR ROLE`

PR #51905, opened 2019-02-28
mount: fix extra -t parameter

aplanas:
If 'fstype' parameter is not set in Linux environments, salt will
build a mount command with an empty -t value, making the command
fail.

Note that this code needs to be backported to 2019.2, as this bug is also present there.

### DIFF
--- a/salt/modules/mount.py
+++ b/salt/modules/mount.py
@@ -1208,13 +1208,15 @@ def mount(name, device, mkmnt=False, fstype='', opts='defaults', user=None, util
         lopts = ','.join(opts)
         args = '-o {0}'.format(lopts)
 
-    # use of fstype on AIX differs from typical Linux use of -t functionality
-    # AIX uses -v vfsname, -t fstype mounts all with fstype in /etc/filesystems
-    if 'AIX' in __grains__['os']:
-        if fstype:
+    if fstype:
+        # use of fstype on AIX differs from typical Linux use of -t
+        # functionality AIX uses -v vfsname, -t fstype mounts all with
+        # fstype in /etc/filesystems
+        if 'AIX' in __grains__['os']:
             args += ' -v {0}'.format(fstype)
-    else:
-        args += ' -t {0}'.format(fstype)
+        else:
+            args += ' -t {0}'.format(fstype)
+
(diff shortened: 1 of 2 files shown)


aplanas:
For reference, this bug was introduced here:

https://github.com/saltstack/salt/commit/b701e16ad06bec1d4b7273d7f48ad788ab715e44#diff-85764a41d31cd335e1c8d99383999e1e

aplanas:
Ouch, is already in 2018.X, so needs to be backported there too.

aplanas:
Based on IRC comment, and after reading [1] I changed the base branch from 2018.3. I hope that this make the code move to 2019.2 and develop.

[1] https://docs.saltstack.com/en/develop/topics/development/contributing.html#which-salt-branch

brejoc (approved):
lgtm

salt/modules/mount.py
```diff
@@ -1214,7 +1214,8 @@ def mount(name, device, mkmnt=False, fstype='', opts='defaults', user=None, util
         if fstype:
             args += ' -v {0}'.format(fstype)
     else:
-        args += ' -t {0}'.format(fstype)
+        if fstype:
```
dincamihai: @aplanas wouldn't it make sense to refactor a bit here? I can see the `if fstype` being duplicated.

aplanas: Makes sense!

aplanas: done

dwoz:
@aplanas Since this if fixing a bug, can you write a test for it please?

aplanas:
@dwoz makes sense!

salt/modules/mount.py
dincamihai: great. thanks!

aplanas:
@dwoz I updated the test_mount code. I think that this test needs some refactoring, but I also think that is out of scope of this pull request

dwoz:
@aplanas I agree we should break up that monster test :D. Looks good for this PR though, thanks!

aplanas:
Thanks for the reviews!

Do I need to port it to 2019.2 and develop? Or is there any process that will take care of it?

PR #51915, opened 2019-02-28
Remove nulls to prevent Zabbix API errors

timdufrane:
If IP or DNS are null, Zabbix API returns an error. Prefer empty string over null.

### What does this PR do?
Replaces nulls with empty strings in "ip" and "dns" params
N/A
### Previous Behavior
Null ip or dns params if empty

### New Behavior
Empty string ip or dns params if empty

### Tests written?
No

### Commits signed with GPG?
No

### DIFF
--- a/salt/states/zabbix_host.py
+++ b/salt/states/zabbix_host.py
@@ -122,8 +122,8 @@ def present(host, groups, interfaces, **kwargs):
             interface_type = interface_ports[value['type'].lower()][0]
             main = '1' if six.text_type(value.get('main', 'true')).lower() == 'true' else '0'
             useip = '1' if six.text_type(value.get('useip', 'true')).lower() == 'true' else '0'
-            interface_ip = value.get('ip')
-            dns = value.get('dns', key)
+            interface_ip = value.get('ip') or ""
+            dns = value.get('dns', key) or ""
             port = six.text_type(value.get('port', interface_ports[value['type'].lower()][1]))
 
             interfaces_list.append({'type': interface_type,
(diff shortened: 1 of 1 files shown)


salt/states/zabbix_host.py
```diff
@@ -122,8 +122,8 @@ def _interface_format(interfaces_data):
             interface_type = interface_ports[value['type'].lower()][0]
             main = '1' if six.text_type(value.get('main', 'true')).lower() == 'true' else '0'
             useip = '1' if six.text_type(value.get('useip', 'true')).lower() == 'true' else '0'
-            interface_ip = value.get('ip')
-            dns = value.get('dns', key)
+            interface_ip = value.get('ip') or ""
```
twangboy: We prefer single quotes in Salt. Also, we should probably use the optional `value` option.
```
interface_ip = value.get('ip', '')
```

salt/states/zabbix_host.py
```diff
@@ -122,8 +122,8 @@ def _interface_format(interfaces_data):
...
+            dns = value.get('dns', key) or ""
```
twangboy: Why is the `or` statement necessary here?

twangboy (changes requested):

salt/states/zabbix_host.py
timdufrane: My SLS specified "- dns" but nothing for "- ip", which generated this for Zabbix API params:
```
		"interfaces": [{
			"port": "10050",
			"dns": "xxx.yyy",
			"type": "1",
			"useip": "0",
			"ip": null,
			"main": "1"
		}]
```
This caused the Zabbix API to repeatedly fail with invalid parameter to function. After switching to empty strings instead of nulls the API calls began working. I suppose that specifying "- ip: '0.0.0.0'" is a smart idea in any case, but it doesn't seem like that should be a requirement when only using DNS.

salt/states/zabbix_host.py
timdufrane: You're right, that's much cleaner. Changed.

salt/states/zabbix_host.py
timdufrane: I see now that you're probably referring to ```dns = value.get('dns', key) or ""```. To be honest, I did not notice the ```key``` argument. You're right, no reason for an or statement there.

twangboy (approved):

dwoz (approved):

PR #51923, opened 2019-03-01
INFRA-5209 - Avoid throwing exception for missing security group in test mode

tkwilliams:
What it says on the can :)  Currently, if a state run WOULD happen to create the needed Security Group before whichever a later state function would require to be there, it will still throw a CommandExecutionError, even though when run "for realsies" it would have succeeded.

Since there's no way (to my knowledge) to have the module code which tries to dereference the SG see the STATE RUN output (which happens to know at a higher layer that the SG would be created), the next best solution I can see is to simply WARN about a potential failure instead when running with test=True.

@ryan-lane - I'd like to get your take on this change specifically since IIRC this was some of yours, and I don't want to fubar semantics if something downstream is expecting this failure mode on a test run...  I don't imagine anyone is, but better safe than sorry :)


tkwilliams:
@rallytime / @dwoz - Looks like @ryan-lane may be AWOL :)

Is there anything else which would block merging at this point?

Thanks
t.

Akm0d (approved):

terminalmage:

Issue #11995, opened 2014-04-14
Python 3 Support


ryan-lane:
Salt is currently not compatible with python 3. Ubuntu trusty and above default to using python 3, so it would be really nice to be able to run salt in the default python version.

thatch45:
We just merged some code that deprecates the need for m2crypto.  So we can start on full py3 support

tebanep:
Any progress regarding this Issue? Python 3.5 was released yesterday and it will be the default in Ubuntu 16.04 LTS.

s0undt3ch:
Several progresses, including dropping m2crypto, however, there's still work to be done. This is a massive task.

kamikaze:
Please make Salt working with Python 3.5. Python 3 came out in 2008! C'moooon

s0undt3ch:
...And only a couple of months ago we were able to drop m2crypto...

Believe me, all of us want Salt fully running under Py3, but have you looked at the current size of Salt's source?

... Its an ongoing daunting task, but its not forgotten...

One just needs to keep believing and submitting PRs...

blarghmatey:
@s0undt3ch Is there any list of tasks related to Python 3 support? I think that this effort would benefit from having an easy to find set of issues/updates that would move the progress forward toward a 3.x supported release.

s0undt3ch:
Full internal Unicode support, without breaking py2 compat.
And I'm sure quite a few modules would need some syntax updates, ie, `.iteritems()` usage, for example.

But honestly, try to start a master in py3 and fix as you go along... Without breaking py2 compact....

Allowing the master to run in py3 will give us some milieage to this task. Then the minion, etc...

tigrus:
Guys, do you have salt branch? Where we can take a look, to understand how much code we must change to port to python3?

deepakhj:
+1. Ubuntu 16.04 installs Python 3 by default.

damon-atkins:
+1
python library cryptography #30287   OpenStack have developed a library, anyone "platform" support openstack will need to  use this crypto library I suspect.  e.g. It's standard with Solaris 11, when Solaris provided support for OpenStack & Solaris Zones.

It would be nice to see Jenkins do a Python 3 Lint as a starting place

thatch45:
Just as an update here, we have made great progress, but this is complicated. Once we can get tests passing on #32366 then we will be able to run remote execution, pillar and states on python 3, this will signify enough to start trying to run the full test suite on python3

[...]

PR #51935, opened 2019-03-01
Update supported python versions in installation docs

Ch3LL:
Fixes comment from issue: https://github.com/saltstack/salt/issues/11995#issuecomment-468069962


cachedout (commented):
Is `futures` still required under Py3? If not, perhaps note that?

dwoz (approved):

cachedout (approved):

PR #51936, opened 2019-03-01
rename kubernetes module to kubernetesmod

gtmanfred:
### What does this PR do?
This makes sure this doesn't import the kubernetes module from
salt/modules while the kubernetes python module is not installed

Issue brought up in community slack

### Tests written?

Yes

### Commits signed with GPG?

Yes


cachedout (approved):
LGTM

salt/states/kubernetes.py
```diff
@@ -4,7 +4,7 @@
 ==========================================
```
dwoz: Could this have been solved using `absolute_imports` here?

gtmanfred: No, because we add salt/modules to sys.path in the loader.

gtmanfred: https://github.com/saltstack/salt/blob/2018.3/salt/loader.py#L1453

PR #51939, opened 2019-03-01
file.line respect quiet flag

toanju:
Despite passing quiet=True to file.line the result is not quiet. This PR
silences the error returned in case quiet is true. The same pattern as
in the replace function is applied.


twangboy (dismissed):

dwoz:
@toanju Can you please write a regression test for this?

dwoz:
@toanju I've rebased this to the `master` branch.

twangboy (changes requested):
This needs a testcase and a changelog

Ch3LL:
Closing due to inactivity. Please let me know if we need to re-open or you can open a new PR based off of master with the feedback included above.

Issue #51940, opened 2019-03-02
Salt's reg.read_value function does not properly remove nulls from REG_SZ string types


arizvisa:
### Description of Issue/Question
In some critical places, such as determining the current system's timezone at minion start, the `reg.read_value` is assumed to return a properly formatted string. This is actually not the case as in all versions of Windows as you can store '\0' bytes in a REG_SZ string due to REG_SZ not actually being a constraint that is linked to the value, but really just a type that is stored in a registry hive.

This results in some parts of salt which assume a string is being returned not acting correctly when being compared or used as a key to a dictionary.

As mentioned before, the current system's timezone is used at minion start which unfortunately results in a real weird issue on a system I installed a minion on where an exception is raised before the minion constructor actually completes execution which results in an invalid attribute being fetched every single interval (flooding the minion's log). The exception that is raised is due to being unable to determine the timezone as a result of the terminating null bytes, the timezone "Unknown" is returned. This is then searched in `mapper.get_unix` (in the win_timezone.get_zone function) which doesn't exist, thus raising the exception that demonstrates the following symptoms.

```
2019-03-02 05:54:03,575 [salt.minion      :1014][CRITICAL][548] Unexpected error while connecting to 172.33.33.160
Traceback (most recent call last):
  File "c:\salt\bin\lib\site-packages\salt\minion.py", line 997, in _connect_minion
    yield minion.connect_master(failed=failed)
  File "c:\salt\bin\lib\site-packages\tornado\gen.py", line 1055, in run
    value = future.result()
  File "c:\salt\bin\lib\site-packages\tornado\concurrent.py", line 238, in result
    raise_exc_info(self._exc_info)
  File "c:\salt\bin\lib\site-packages\tornado\gen.py", line 1063, in run
    yielded = self.gen.throw(*exc_info)
  File "c:\salt\bin\lib\site-packages\salt\minion.py", line 1189, in connect_master
    yield self._post_master_init(master)
  File "c:\salt\bin\lib\site-packages\tornado\gen.py", line 1055, in run
    value = future.result()
  File "c:\salt\bin\lib\site-packages\tornado\concurrent.py", line 238, in result
    raise_exc_info(self._exc_info)
  File "c:\salt\bin\lib\site-packages\tornado\gen.py", line 1069, in run
    yielded = self.gen.send(value)
  File "c:\salt\bin\lib\site-packages\salt\minion.py", line 1234, in _post_master_init
    cleanup=[master_event(type='alive')])
[...]

PR #51941, opened 2019-03-02
Ensure that terminating null-bytes are being removed from registry values fetched from salt.utils.win_reg.

arizvisa:
### What does this PR do?
Please see the original issue at #51940. I'm really getting tired of creating PRs and filling out forms as I just really want things to work. So, merge this if you want.

This closes issue #51940.

### Previous Behavior
When fetching a REG_SZ value from the registry, terminating null-bytes may be returned. Some calls assume that a correct string is returned such as when dealing with the timezone when a  minion initializes. Since null bytes are returned, these calls can fail as described in the initial report).

### New Behavior
This removes terminating null-bytes. This may not be what the caller intends as Null-bytes are considered part of the registry value, however there are numerous places in salt that assume that values read from the registry are assumed to contain no null bytes and therefore this is the quickest and easiest fix rather than going through all of them and fixing that code.

### Tests written?
No

### Commits signed with GPG?
No


twangboy (approved):

twangboy:
We're going to revert #51941 and check the return from `reg.read_value` at the call in each module. There may be instances where we want to know if there's a trailing Null Byte.

arizvisa:
Okay, cool. Yeah, I was thinking that as well as removing the null-bytes changes the semantics despite most of salt assuming that it returns a well-formed string.

Maybe something like a `reg.read_string` might be more clear perhaps.

arizvisa:
Can you let me know what the PR for that is whenever you guys start working on it? (Either here, or in the related issue)

twangboy:
Hey, here's the PR
https://github.com/saltstack/salt/pull/52013

twangboy:
We need to check the return from the registry when we're calling it. I think it only needs to be done if we're querying REG_SZ and possibly REG_MULTI_SZ.

This no longer applies to the `timezone` issue you fixed this for as you changed it to use the command line tool `tzutil` instead of querying the registry.

arizvisa:
Awesome. Thanks.

Ch3LL:
@twangboy did you still want this backported, seems its getting reverted if i'm reading this correctly?

twangboy:
@ch3ll No, do not back port this... it is being reverted

PR #51947, opened 2019-03-02
Don't show changes for docker_container.running with custom network and hostname

clinta:
### What does this PR do?

This make sure to ignore the docker id aliases when doing network comparisons. This is usually the same as the container hostname, but not when a custom hostname is specified.

### Previous Behavior

When a `docker_container.running` state is run with a custom network and hostname, changes are shown for every run because the compared aliases change.

### New Behavior
Changes are no longer detected because container id aliases are ignored.

### Tests written?

Yes

### Commits signed with GPG?

Yes


clinta:
Test demonstrating the bug in 03198f1. Fixed in d1a2d1a.

clinta:
This is freshly rebased, and all CI tests are passing.

twangboy (approved):

dwoz (approved):

sagetherage:
is in Master no port PR needed

PR #51949, opened 2019-03-02
Remove hipchat engine due to service retirement

mchugh19:
### What does this PR do?
Removes the hipchat engine

None, but hipchat cloud is no more.

### Previous Behavior
There was hipchat, and salt worked with it.

### New Behavior
No more hipchat, so salt has no one to talk to

### Tests written?
No

### Commits signed with GPG?
No


garethgreenaway (approved):

mchugh19:
I don't think we have to do removals very often, is it worth adding a sentence into the release notes?

garethgreenaway:
@mchugh19 Yup.  A note in the release notes about removing the engine would be a good addition.

Issue #49147, opened 2018-08-16
Log filled with "Exception occurred while Subscriber handling stream: Already reading"


furgerf:
### Description of Issue/Question
Hi, on a server running salt-minion and salt-master, the minion's logfile was today filled (quite quickly) with the following message:

```
2018-08-12 06:25:10,762 [salt.transport.ipc][ERROR   ][25008] Exception occurred while Subscriber handling stream: Already reading
2018-08-12 06:25:10,762 [salt.transport.ipc][ERROR   ][25008] Exception occurred while Subscriber handling stream: Already reading
2018-08-12 06:25:10,762 [salt.transport.ipc][ERROR   ][25008] Exception occurred while Subscriber handling stream: Already reading
2018-08-12 06:25:10,763 [salt.transport.ipc][ERROR   ][25008] Exception occurred while Subscriber handling stream: Already reading
2018-08-12 06:25:10,763 [salt.transport.ipc][ERROR   ][25008] Exception occurred while Subscriber handling stream: Already reading
2018-08-12 06:25:10,763 [salt.transport.ipc][ERROR   ][25008] Exception occurred while Subscriber handling stream: Already reading
2018-08-12 06:25:10,763 [salt.transport.ipc][ERROR   ][25008] Exception occurred while Subscriber handling stream: Already reading
2018-08-12 06:25:10,763 [salt.transport.ipc][ERROR   ][25008] Exception occurred while Subscriber handling stream: Already reading
2018-08-12 06:25:10,763 [salt.transport.ipc][ERROR   ][25008] Exception occurred while Subscriber handling stream: Already reading
2018-08-12 06:25:10,764 [salt.transport.ipc][ERROR   ][25008] Exception occurred while Subscriber handling stream: Already reading
```

Stopping the minion service didn't stop the process writing to the logfile, neither did SIGTERM - I only got rid of it with SIGKILL (PID 25008 here):

```
$ ps aux | grep [p]ython
root     24851  0.0  0.1 101600  6048 ?        S    Aug10   0:00 /usr/bin/python /usr/bin/salt-minion
root     24854  0.0  0.2 548420 11264 ?        Sl   Aug10   0:09 /usr/bin/python /usr/bin/salt-minion
root     25005  0.0  0.1 101600  6412 ?        S    Aug10   0:00 /usr/bin/python /usr/bin/salt-minion
[...]

PR #51963, opened 2019-03-04
Allow multiple instances of IPCMessageSubscriber in one process

DmitryKuzmenko:
### What does this PR do?
Updates IPCMessageSubscriber to be instantiated and read simultaneously. To do this I'm introducing an IPCMessageSubscriberService that is the old IPCMessageSubscriber singleton. It's instantiated once and it controls that the socket reading is started once any subscriber want's to read data and stopped when there still no subscribers.
IPCMessageSubscriber is subscribing to data read by the service and reading the data from the async queue not dire

Fixes: #49147

### Previous Behavior
The second parallel try to read data with IPCMessageSubscriber raises socket error while another reading is in progress.

### New Behavior
IPCMessageSubscriber can be used anywhere without taking care of other parts of client code.

### Tests written?
Yes, regression.

### Commits signed with GPG?
Yes


salt/transport/ipc.py
```diff
@@ -582,11 +583,116 @@ def __del__(self):
...
+    def close(self):
+        '''
+        Routines to handle any cleanup before the instance shuts down.
+        Sockets and filehandles should be closed explicitly, to prevent
+        leaks.
+        '''
+        if not self._closing:
+            super(IPCMessageSubscriberService, self).close()
+
+    def __del__(self):
```
dwoz: @DmitryKuzmenko Why do we need to use the `__del__` method? Is it possible to explicitly close these where we use them? Perhaps in a process teardown?

DmitryKuzmenko: Currently it's common way to close all the transport channels, so my opinion if we'd like to change this it shall be done everywhere in a specific PR. I've just leaved things here as they were before.

salt/transport/ipc.py
```diff
@@ -615,147 +721,60 @@ class IPCMessageSubscriber(IPCClient):
...
+        '''
+        Asynchronously read messages and invoke a callback when they are ready.
 
-        if exc_to_raise is not None:
-            raise exc_to_raise  # pylint: disable=E0702
-        self._sync_read_in_progress.release()
-        raise tornado.gen.Return(ret)
+        :param callback: A callback with the received data
+        '''
+        self.service.read(self)
```
s0undt3ch: Isn't the service read also a coroutine?
Missing another yield?

salt/transport/ipc.py
```diff
@@ -582,11 +583,116 @@ def __del__(self):
...
+        while not self.connected():
+            try:
+                yield self.connect(timeout=5)
+            except tornado.iostream.StreamClosedError:
+                log.trace('Subscriber closed stream on IPC %s before connect', self.socket_path)
+                yield tornado.gen.sleep(1)
+            except Exception as exc:
+                log.error('Exception occurred while Subscriber connecting: %s', exc)
+                yield tornado.gen.sleep(1)
+        self._read(timeout)
```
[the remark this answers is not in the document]
DmitryKuzmenko: No, we don't need `yield` here because the `_read` returns nothing, it starts the coroutine reading the data into the async queue. So here we just run the async coroutine that will be put to the IOLoop's execution pipeline.

salt/transport/ipc.py
DmitryKuzmenko: The same as in the previous comment.

salt/transport/ipc.py
```diff
@@ -582,11 +583,116 @@ def __del__(self):
...
 
 
-class IPCMessageSubscriber(IPCClient):
+class IPCMessageSubscriberService(IPCClient):
+    '''
+    IPC message subscriber service that is a standalone singleton class starting once for a number
+    of IPCMessageSubscriber instances feeding all of them with data. It closes automatically when
+    there are no more subscribers.
+
+    To use this rever to IPCMessageSubscriber documentation.
```
twangboy: refer to IPCMessageSubscriberService documentation

twangboy (approved):
Just a documentation change... if you wanna

DmitryKuzmenko:
@dwoz, @s0undt3ch are you ok with this now?

s0undt3ch (approved):

dwoz (approved):

Issue #51816, opened 2019-02-26
pillar_ldap broken in 2019.2.0


jkrauz:
### Description of Issue/Question
This is the same issue as in https://github.com/saltstack/salt/issues/46859. Returned in 2019.2.0.

### Versions Report
Salt Version:
           Salt: 2019.2.0

Dependency Versions:
           cffi: Not Installed
       cherrypy: Not Installed
       dateutil: Not Installed
      docker-py: Not Installed
          gitdb: 2.0.2
      gitpython: 2.1.3
          ioflo: Not Installed
         Jinja2: 2.7.2
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.5.6
   mysql-python: Not Installed
      pycparser: Not Installed
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: Not Installed
         Python: 2.7.5 (default, Oct 30 2018, 23:45:53)
   python-gnupg: Not Installed
         PyYAML: 3.11
          PyZMQ: 15.3.0
           RAET: Not Installed
          smmap: 2.0.2
        timelib: Not Installed
        Tornado: 4.2.1
            ZMQ: 4.1.4

System Versions:
           dist: centos 7.6.1810 Core
         locale: UTF-8
        machine: x86_64
        release: 3.10.0-957.5.1.el7.x86_64
         system: Linux
        version: CentOS Linux 7.6.1810 Core

Ch3LL:
glancing at the code i'm not seeing a lot of change in those files. Can you paste the stack trace here so we can see exactly where its showing up in the code.

jkrauz:
2019-02-27 10:03:27,204 [salt.loaded.int.pillar.pillar_ldap:301 ][CRITICAL][14613] Failed to retrieve pillar data from LDAP:
Traceback (most recent call last):
  File "/usr/lib/python2.7/site-packages/salt/pillar/pillar_ldap.py", line 295, in _do_search
    **connargs)['results']
  File "/usr/lib/python2.7/site-packages/salt/modules/ldapmod.py", line 151, in search
    results = _ldap.search_s(dn, int(scope), filter, attrs)
  File "/usr/lib64/python2.7/site-packages/ldap/ldapobject.py", line 552, in search_s
[...]

PR #51964, opened 2019-03-04
[2019.2] Make sure ldap attrs are string types on python 2

dwoz:
### What does this PR do?

Add optional default to the `_config` method in the pillar_ldap module. This ensures default values are converted to string types on python 2.

Fixes issue #51816

### Tests written?

Yes

### Commits signed with GPG?

Yes


waynew (approved):
Looks good 👍

Issue #51578, opened 2019-02-11
MasterPillarUtil __opts__ assignment seems to be poisoning salt-master options when used in ext_pillar module


kylehuff:
### Description of Issue/Question
Providing unique run-time options (opts) to `MasterPillarUtil` within an external pillar provider seems to be poisoning the configured options of the running salt-master process(es) after invoking `get_minion_pillar()`.  Is this intended/desired behavior?  It doesn't seem desirable that an external instantiation of MasterPillarUtil with custom options would alter the configuration of the running salt-master process(es).

### Setup
Can be demonstrated by instantiating MasterPillarUtil with a custom options dictionary that sets the `ext_pillar` config item to an empty list and then invoking `get_minion_pillar()`.  After doing so, subsequent requests for pillar data no longer invoke ext_pillar providers (this lasts until the salt-master process is restarted).  For expedience, the example master config below sets the WORKER_THREADS value to 1 to quickly illustrate this behavior (otherwise, the second iteration may be executed by a worker thread that has not yet been poisoned).

Master config:
 
    worker_threads: 1
 
    file_roots:
      base:
        - /srv/salt/states
 
    pillar_roots:
      base:
        - /srv/salt/pillar
 
    extension_modules: /srv/salt/modules
 
    ext_pillar:
      - cmd_json: 'echo {\"pillar_from_cmd_json\":\"a_json_value\"}'
      - my_pillar_module: noargs
 
Example ext_pillar provider `/srv/salt/modules/pillar/my_pillar_module.py`:
 
    #!/bin/env python
 
    import salt.utils.master
    import salt.pillar
 
    import logging
 
    log = logging.getLogger('my_pillar_module')
 
 
    def ext_pillar(minion_id, pillar, *args, **kwargs):
      # return pillar
      ret_pillar = {'dual-shared-value': 'no match'}
 
      neighbor = pillar.get('neighbor')
      log.info("ext_pillar 'my_pillar_module' checking relationship between {} and {}".format(minion_id, neighbor))
 
      # disable ext_pillars, we don't want a recursion problem
      master_options = {}
      master_options.update(__opts__)
      master_options['ext_pillar'] = []
 
      # setup pillar request
      pillar_util = salt.utils.master.MasterPillarUtil(
        neighbor,
        "glob",
        use_cached_grains=True,
        grains_fallback=False,
        use_cached_pillar=True,
        pillar_fallback=False,
        opts=master_options,
      )
     
[...]

PR #51991, opened 2019-03-06
Copy opts to avoid cache poisioning

waynew:
### What does this PR do?

Creates a copy of the opts dictionary to avoid external pillars poisoning the opts cache

#51578 

### Previous Behavior

If you create an external pillar, and then have it set `'ext_pillar':[]` and pass it to a `MasterPIllarUtil` class, as `opts` then the second time that master process runs it will fail to load the external pillar, because it has `'ext_pillar': []` in the cache.

### New Behavior

Now it will properly load the external pillar each time.

### Tests written?

Yes

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/loader.py
+++ b/salt/loader.py
@@ -8,10 +8,10 @@ plugin interfaces used by Salt.
 # Import python libs
 from __future__ import absolute_import, print_function, unicode_literals
 
+import copy
 import functools
 import inspect
 import logging
-import copy
 import os
 import re
 import sys
--- a/salt/loader.py
+++ b/salt/loader.py
@@ -1,0 +1,2123 @@
+# -*- coding: utf-8 -*-
+"""
+The Salt loader is the core to Salt's plugin system, the loader scans
+directories for python loadable code and organizes the code into the
+plugin interfaces used by Salt.
+"""
+
+# Import python libs
+from __future__ import absolute_import, print_function, unicode_literals
+
+import functools
+import inspect
+import logging
+import copy
+import os
+import re
+import sys
+import tempfile
+import threading
+import time
+import traceback
+import types
+from collections.abc import MutableMapping
+from zipimport import zipimporter
+
+# Import salt libs
+import salt.config
+import salt.defaults.events
+import salt.defaults.exitcodes
+import salt.syspaths
+import salt.utils.args
+import salt.utils.context
+import salt.utils.data
+import salt.utils.dictupdate
+import salt.utils.event
+import salt.utils.files
+import salt.utils.lazy
+import salt.utils.odict
+import salt.utils.platform
+import salt.utils.stringutils
+import salt.utils.versions
+from salt.exceptions import LoaderError
+
+# Import 3rd-party libs
+from salt.ext import six
+from salt.ext.six.moves import reload_module
+from salt.template import check_render_pipe_str
+from salt.utils.decorators import Depends
+
+if sys.version_info[:2] >= (3, 5):
+    import importlib.machinery  # pylint: disable=no-name-in-module,import-error
+    import importlib.util  # pylint: disable=no-name-in-module,import-error
+
+    USE_IMPORTLIB = True
+else:
+    import imp
+
+    USE_IMPORTLIB = False
+
+
+try:
+    import pkg_resources
+
+    HAS_PKG_RESOURCES = True
+except ImportError:
+    HAS_PKG_RESOURCES = False
+
+log = logging.getLogger(__name__)
+
+SALT_BASE_PATH = os.path.abspath(salt.syspaths.INSTALL_DIR)
+LOADED_BASE_NAME = "salt.loaded"
+
+if USE_IMPORTLIB:
+    # pylint: disable=no-member
+    MODULE_KIND_SOURCE = 1
+    MODULE_KIND_COMPILED = 2
+    MODULE_KIND_EXTENSION = 3
+    MODULE_KIND_PKG_DIRECTORY = 5
+    SUFFIXES = []
+    for suffix in importlib.machinery.EXTENSION_SUFFIXES:
+        SUFFIXES.append((suffix, "rb", MODULE_KIND_EXTENSION))
+    for suffix in importlib.machinery.SOURCE_SUFFIXES:
+        SUFFIXES.append((suffix, "rb", MODULE_KIND_SOURCE))
+    for suffix in importlib.machinery.BYTECODE_SUFFIXES:
+        SUFFIXES.append((suffix, "rb", MODULE_KIND_COMPILED))
+    MODULE_KIND_MAP = {
+        MODULE_KIND_SOURCE: importlib.machinery.SourceFileLoader,
+        MODULE_KIND_COMPILED: importlib.machinery.SourcelessFileLoader,
+        MODULE_KIND_EXTENSION: importlib.machinery.ExtensionFileLoader,
+    }
+    # pylint: enable=no-member
+else:
+    SUFFIXES = imp.get_suffixes()
+
+PY3_PRE_EXT = re.compile(
+    r"\.cpython-{0}{1}(\.opt-[1-9])?".format(*sys.version_info[:2])
+)
+
+# Because on the cloud drivers we do `from salt.cloud.libcloudfuncs import *`
+# which simplifies code readability, it adds some unsupported functions into
+# the driver's module scope.
+# We list un-supported functions here. These will be removed from the loaded.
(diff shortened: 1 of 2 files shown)


twangboy (dismissed):

dwoz (commented):
@waynew Is it possible to create a test for this?

waynew:
@dwoz I had the same question 😕 

I could just test that the dictionary has a different ID, but it seems like that's of limited value?

The other test (not sure exactly what it would take to implement it) looks like this:

start a single-process master, with an ext_pillar that has a MasterPillarUtil, passing it an opts dict with `'ext_pillar': []`, then ensure that after a call to pillar.items, subsequent call to pillar.items loads the external pillar.

It's not clear to me if the tradeoff is worth it.

I *do* think that it's worth commenting this code if not, though.

cachedout:
IMHO this is not a complete fix and may introduce difficult-to-find bugs. 

This looks like you are doing a shallow copy of the opts dictionary which will do the right thing in top-level keys but *not* for nested keys, of which the opts dict contains many.

For example:

```
>>> a = {'1': {'2': '3'}}
>>> b = dict(a)
>>> a
{'1': {'2': '3'}}
>>> b
{'1': {'2': '3'}}
>>> a['1']['2'] = '9'
>>> a
{'1': {'2': '9'}}
>>> b
{'1': {'2': '9'}}```

Compare to a deep copy:

```
>>> import copy
>>> a = {'1': {'2': '3'}}
>>> b = copy.deepcopy(a)
>>> a['1']['2'] = '9'
>>> a
{'1': {'2': '9'}}
>>> b
{'1': {'2': '3'}}```

Even though the original issue is about a top-level key, if this is about `__opts__` needing to be isolated, this is an incomplete approach for solving that for any nested key in the dictionary.

cachedout (changes requested):
See comment on in the main thread of the PR.

waynew:
> Even though the original issue is about a top-level key, if this is about `__opts__` needing to be isolated, this is an incomplete approach for solving that for any nested key in the dictionary.

From my discussions, it seems like any approach short of freezing `__opts__` is an incomplete approach :trollface: 

I guess what I'm ultimately saying is that this PR is not attempting a general fix of `__opts__` - and maybe we need someone to sit down and write a SEP about that, because from what I understand, `__opts__` causes a ton of subtle and hard to understand bugs already. TBH, the reason I *didn't* use `copy.deepcopy` here is because I'm entirely not sure if there is code that expects to be able to communicate via those nested values, and I didn't want to break that code, just because it's doing the right thing the wrong way 😝 

If you think that risk is worth the change, I'm more than happy to change that `dict` call to a `copy.deepcopy`. I was just punting on that until we decide what to do with `__opts__`. I also picked this limited approach due to the fact that when I was trying to figure out this change literally any kind of wrapping of the dict fixed this problem.

I just did a `class FancyDict(dict):` that literally stored the opts dict internally and wrapped getitem and setitem and it fixed this cache poisoning problem. I couldn't reproduce the behavior with any kind of wrapped spying.

---

Maybe it's worth specifying with a comment, and/or here in the PR, that currently `__opts__` can be abused so here be dragons? I very much welcome suggestions and/or information.

cachedout:
My concern in this instance comes from what I still see as a potentially security flaw here. If an external service can modify `__opts__` for a running master that is extremely problematic. 

If this fix is designed to prevent that, it needs to isolate all of `__opts__` and not just top-level keys. If it causes additional problems, so be it, but I just don't see how Salt can allow for a situation in which configuring an external pillar means that whomever can compromise that service can potentially control _any_ part of the running configuration of masters which consume from it.

waynew:
> My concern in this instance comes from what I still see as a potentially security flaw here. If an external service can modify __opts__ for a running master that is extremely problematic.

If you write code to do that, then yes, it can do that 🙂 

This PR does not introduce that capability, which is currently how Salt behaves.

My feeling is that properly ensuring the integrity of `__opts__` for the running master is far beyond the scope of this PR, which is just to fix #51578.

My suggestion is that we should create a new issue addressing the lack of isolation in `__opts__` so that we can properly audit, test, and fix the problems we'll find. I think that trying to address that problem within this PR will be much less effective than making this minor improvement, and then creating an issue that can properly address the broader scope, which I'm pretty sure will benefit from a broader visibility. Or am I missing something?

austinpapp:
just my 2 cents

> __opts__ causes a ton of subtle and hard to understand bugs already.

well, by only partially fixing this problem , i think you are just kicking the can. top level "works" but nested exerts a completely different behavior. that alone is massively confusing. 🤷‍♂️

mattp-:
I don't think that is the right place to make that copy (and I think it should be a deep copy). Why not add it to salt.cache.__init__ instead? 
the place it is being shared is through the masterapi.cache instance (through the mod.__opts__.update() you referenced), so it could probably be added to self.opts for deep copying as well.

waynew:
> well, by only partially fixing this problem , i think you are just kicking the can. 

Yes, I am 🙂 

I could *not* fix this particular issue, because properly fixing `__opts__` is the broadest, most correct fix. And from a purely theoretical point of view, that's the perfect thing to do.

But properly fixing `__opts__` is going to be a non-trivial enhancement. I'm betting there's more than a weeks worth of effort there to understand the impact and implement the change.

> top level "works" but nested exerts a completely different behavior. that alone is massively confusing.

We're both making the assumption that something is using `__opts__` inappropriately :trollface: 

If nothing is abusing opts then the fact that this isn't a comprehensive solution won't actually be a problem.

On the other hand, if there are things abusing opts then the fact that this isn't a comprehensive solution is actually a good thing, because it's making the minimal amount of change required to fix this specific issue so it's unlikely to break *other* things that are relying on doing the wrong things.

I'm all for fixing things that are doing wrong things (i.e. paying down technical debt), but I'm also comfortable saying that this is not the appropriate place or reason to make the payment on `__opts__`. Of course, that's based on the information that I currently have. If I get better information, I'm happy to change my mind, but I currently don't have any information that suggests that we currently need a more comprehensive fix.

This patch makes an improvement, fixes a bug, and *probably* doesn't introduce new bugs. And doesn't introduce a significant amount of technical debt.

cachedout:
> If nothing is abusing opts then the fact that this isn't a comprehensive solution won't actually be a problem.

Respectfully, I think it may be beneficial to have @thatch45 to weigh in here as he has a great deal of experience when it comes to debugging issues caused by bifurcating opts with shallow copies. :)

thatch45:
This change would not be enough, and I think it is covering up deeper issues. Any way about it we need to make a special function to deeply copy a dict (I thought there was one in utils/dict.py). We also need to keep in mind that making it change for just one loader instance can cause confusion when writing plugins in general.

In the end I think it would make more sense to see what happens to the tests when we use the frozenset approach.

As for this, I think it is an ok way to fix the given issue, but we should be doing a propper nested deep dict copy

mattp-:
I take it by frozenset it is _not_ desirable for master opts to be mutable? I know in certain spots (ie minion master tune in), functionality is actually reliant on mutating opts, I'm sure it's elsewhere. Making the deepcopy on the input points (salt.cache constructor, which passes to salt.loader.cache, and salt.masterapi, which passes it to salt.cache and probably others) would be a cleaner spot to squash this bug imo

thatch45:
As I look at this I am torn between making the loader create a deep copy of the opts dict when every loader instance happens, so one copy of opts for all of `__salt__` or to make the frozenset. Both approaches should resolve this in an elegant way

waynew:
@thatch45 do you have any preference for this yet?

FWIW, I think a healthy approach would be to stick the deepcopy in the loader for 2018.3, and then freeze `__opts__` in develop (and potentially wrap the `__opts__` setitem with a warning or something).

thatch45:
You are spot on. Just make sure that you have one copy of opts per loader instance, not one copy per module :)

dwoz:
@waynew Can you port this to the master branch please?

dwoz:
@waynew I've updated this to point to master.

waynew:
@dwoz as soon as tests pass, this should be good to go.

dwoz:
@waynew Sorry, this needs another rebase.

sagetherage:
this isn't failing a test any longer so I removed that label, I am unsure what order we want to update PRs with master since currently we are in code freeze, so I am not touching that piece and will leave that to you @waynew :D

waynew:
This was fixed in https://github.com/saltstack/salt/pull/59832/files#diff-e8686c624191c5109ca13f27e7fee3ca5ac3703958fe784bfb3fd4a7cc08b51eR1317

PR #52001, opened 2019-03-06
extmods: add utils directories in sys.path

### LINKED ISSUE #51958: extmods/utils is not in sys.path after synchronizing _utils directory

When doing a state.hightstate, the initial synchronization works properly and all the _modules, _utils and _grains code from the formula are living in '/var/cache/salt/minion/files/base/' and ('/var/cache/salt/minion/extmods'.

The problem is that when a _module import some code that is living in _utils, this is not found the during this first highstate, as for some reason /var/cache/salt/minion/extmods/utils is not in the sys.path, as only /var/cache/salt/minion/extmods/modules is there.

The extmods/utils is present in the second highstate run, and the import succeed.

### Versions Report

develop branch

aplanas:
### What does this PR do?

When the minion start clean and there are not utils directory in
the extmods cache, any import from a module to any code inside the
utils directory will fail. A second run of the highstate will fix
this issue, as the utils directories are added into sys.path inside
config/__init__.py, as part of the master and minion startup.

This commit add the utils directories during the synchronization of
the custom modules.

Fixes #51958

### DIFF
--- a/salt/utils/extmods.py
+++ b/salt/utils/extmods.py
@@ -8,6 +8,7 @@ from __future__ import absolute_import, unicode_literals, print_function
 import logging
 import os
 import shutil
+import sys
 
 # Import salt libs
 import salt.fileclient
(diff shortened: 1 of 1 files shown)


twangboy:
@thatch @dwoz Will this affect the loader?

aplanas:
I tested this code for a week and I didn't find any problem with the loader. I am only tenting it in Linux.

thatch45:
@aplanas logic is sound. This should be good to merge

thatch45 (approved):

PR #52006, opened 2019-03-06
Allow minions to request the master to clear its pillar cache

buchanan:
Adds a function clear_masterpillar to the saltutil module to allow the minion to send a request to the master to have its pillar cache cleared. Functionality must be enabled in the salt-master conf.

### Tests written?
No

### Commits signed with GPG?
No

### DIFF
--- a/salt/daemons/masterapi.py
+++ b/salt/daemons/masterapi.py
@@ -1034,6 +1034,31 @@ class RemoteFuncs(object):
                                                          False))
         return True
 
+    def clear_pillarcache(self, load):
+        '''
+        Allow a minion to clear its master pillar cache
+        '''
+        if not all(k in load for k in ('id', 'grains', 'saltenv', 'pillarenv')):
+            return False
+        if self.opts.get('pillar_cache', False):
+            try:
+                pillar_cache = salt.pillar.PillarCache(
+                    self.opts,
+                    load['grains'],
+                    load['id'],
+                    saltenv=load['saltenv'],
+                    functions=self.mminion.functions,
+                    pillarenv=load['pillarenv']).cache
+                if load['id'] in pillar_cache:
+                    del pillar_cache[load['id']]
+            except Exception as exc:
+                log.error(
+                    'Exception %s occurred in clear pillarcache for minion %s',
+                    exc, load['id'], exc_info_on_loglevel=logging.DEBUG
+                )
+                return False
+        return True
+
 
 class LocalFuncs(object):
     '''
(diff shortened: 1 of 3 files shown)


afischer-opentext-com:
Just out of curiosity, what is the use case for this? 

Some notes, the PR does not consider masterless minions (no _master\_uri_ parameter defined). It may be no use case to call this on masterless minions, still the exception being raised due to this is not ideal imho. Why is the _master_ config parameter not considered?

buchanan:
> Just out of curiosity, what is the use case for this?
Most of our system administrators don't have access to the salt-master and need a quick way to refresh a minions pillar data when changes are made.

waynew:
Would `salt-call saltutil.pillar_refresh` on the minion not work?

cachedout:
> Would `salt-call saltutil.pillar_refresh` on the minion not work?

@waynew Do you mean `saltutil.refresh_pillar`? If so, this refreshes on the minion side and wouldn't operate as I believe this PR intends.

waynew:
Yes, that's what I meant 😜 I was under the impression that `refresh_pillar` would reload pillar data from the master, but I guess it only reloads the cached pillar data from the master?

buchanan:
> Yes, that's what I meant  I was under the impression that `refresh_pillar` would reload pillar data from the master, but I guess it only reloads the cached pillar data from the master?

If it would be more accepted by the community. I'm not opposed to the idea of using `saltutil.refresh_pillar` to clear the master's pillar_cache if a flag is set in the master config

waynew:
I don't know about anyone else, but think I like that idea - if there was a `refresh_pillar_clears_cache: True` or something (that defaults to the current existing behavior, of course). Or now that I'm kicking it around in my head, maybe a `refresh_pillar_can_clear_cache`, then one does something like `saltutil.refresh_pillar clear_cache=True`?

🤷‍♂️

cachedout:
What happens when a syndic receives this request? If it forwards it (which I assume it does), doesn't that mean that the MoM receives it and acts on it based on the `id` of the sender?

If this is the case, there is a problem. Minion ids must be unique for the master they are connected to but *not* between syndics. Given that, it would be possible for a minion from one syndic to potentially delete the cache from a minion connected to a top-level master if they share the same name. 

It might be good to test this and see what happens.

afischer-opentext-com:
I also second the idea here, honestly I thought that `refresh_pillar` would do exactly this already.

waynew:
@buchanan here's another approach that may work - you could use [reactors](https://docs.saltstack.com/en/latest/topics/reactor/) and fire an event from the minion that requests a cache clear. Would that work for you?

waynew:
@buchanan I'm going to close this PR due to inactivity. If you'd like to have it re-opened, just let us know. Thanks!

Issue #51832, opened 2019-02-26
pillar include stopped working


nocturo:
### Description of Issue/Question
I have used git for ext_pillar for a while now and when upgrading to 2019.2.0 no pillars get rendered for me. As a quick test I downgraded to 2018.3.4 and everything works as expected.

ext_pillar config:
```yaml
ext_pillar:
  - git:
    - master git@github.com:user/pillar.git:
       - root: pillar
       - privkey: /root/.ssh/id_rsa
       - pubkey: /root/.ssh/id_rsa.pub
```

This straight up doesn't load up the pillar top.sls if I have root: specified. If I remove the option and put top.sls in git "root" it then loads it and I get error that now it can't find the SLS (as expected since the path is not relative anymore to the specified root in the config which I've been using).

### Steps to Reproduce Issue
Configure exit_pillar git pillar with root: and top.sls in the root.

### Versions Report
```Salt Version:
           Salt: 2019.2.0

Dependency Versions:
           cffi: 1.9.1
       cherrypy: unknown
       dateutil: Not Installed
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.8
        libgit2: 0.24.6
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.5.6
   mysql-python: Not Installed
      pycparser: 2.14
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: 0.24.2
         Python: 3.4.9 (default, Feb  5 2019, 14:36:09)
   python-gnupg: Not Installed
         PyYAML: 3.11
          PyZMQ: 15.3.0
           RAET: Not Installed
          smmap: Not Installed
        timelib: Not Installed
        Tornado: 4.4.2
            ZMQ: 4.1.4

System Versions:
           dist: centos 7.6.1810 Core
         locale: UTF-8
        machine: x86_64
        release: 3.10.0-957.1.3.el7.x86_64
         system: Linux
        version: CentOS Linux 7.6.1810 Core
```

Ch3LL:
im not able to replicate this with:

```
ext_pillar:                                                                                                                                                                                          
  - git:                                                                                                                                                                                             
    - master https://github.com/Ch3LL/Ch3LLScripts.git:                                                                                                                                              
      - root: salt/pillars
```

[...]

PR #52008, opened 2019-03-06
Allow `/` in pillar includes

waynew:
### What does this PR do?

Re-allows / in pillar includes

#51832 

### Previous Behavior

Using `fnmatch` stopped us from allowing `/` in our pillar includes.

### New Behavior

re-allow `/`

### Tests written?

No(t yet)

### Commits signed with GPG?

Yes

---

This works but I'm not certain if this works too broadly. For instance, if you wrote:

```
  include: foo/bar.bang/thing
```

It would look for `foo.bar.bang.thing`. Is that reasonable, or should I throw in a check to make sure that there is no `.` in the `sub_sls` before I do the replacement?

### DIFF
--- a/salt/pillar/__init__.py
+++ b/salt/pillar/__init__.py
@@ -776,7 +776,7 @@ class Pillar(object):
                             try:
                                 matched_pstates.extend(fnmatch.filter(
                                     self.avail[saltenv],
-                                    sub_sls.replace('/', '.'),
+                                    sub_sls.lstrip('.').replace('/', '.'),
                                 ))
                             except KeyError:
                                 errors.extend(
--- a/salt/pillar/__init__.py
+++ b/salt/pillar/__init__.py
@@ -257,7 +257,7 @@ class RemotePillar(RemotePillarMixin):
         return ret_pillar
 
     def destroy(self):
-        if self._closing:
+        if hasattr(self, '_closing') and self._closing:
             return
 
         self._closing = True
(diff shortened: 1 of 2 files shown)


dwoz (commented):
@waynew does this have a test that verifies the behavior?

waynew:
@dwoz not yet - I wanted to make sure that this behavior makes sense before I went and wrote the test for it.

i.e. should all of these be the same as `include: foo.bar.bang.quux`?

- `include: foo/bar/bang/quux`
- `include: foo.bar/bang.quux`
- `include: foo/bar.bang/quux`

If the answer is yes, I'll get that test knocked out... otherwise... should I just make it so

- `include: foo/bar/bang/quux` is equivalent, while `include: foo.bar/bang.quux` is something different?

dmurphy18:
@waynew also capturing it here, needs to support leading dot as in :

 include:
   - .foo.bar.bang.quux

waynew:
@dmurphy18 So... I have some mixed feelings about my current PR.

On the one hand, yes, it does work for allowing `/` and a leading `.`

The part I'm questioning is that *technically* this introduces a new feature, i.e. relative pillar includes. Before, a leading `.` worked, it was just a nop. Now the include will be relative to the path of the pillar file doing the include.

Should I break this up into multiple PRs, one restoring the behavior, and one with the improvement?

dmurphy18:
@waynew I would be in favor of splitting, an issue per fix is my motto.

dwoz (approved):

PR #52014, opened 2019-03-07
fix some python3 type issue on x509

arsiesys:
fix some python3 type issue on x509
    
    [salt/states/x509.py]
    Behavior:
    When you store the private key and a certificate in a same file, both have to be in the same format.
    But in python3, when the private key is already existing and get using "get_pem_entry", it will return a bytes. A new generated one will be in str. That mean that at the second execution, your state will fail. We have to convert the get_pem_entry return in a str using .decode().
    
    [salt/modules/x509.py]
    Behavior:
    At each state execution, salt is making a comparison of the current key a new generated one.
    If one element (Subject, public key of the issuer...) differ, the key is replaced.
    In python3, the local key is in bytes while the remote key is provided in str. The compraison then fail $
    Fix:
    - use .decode() on the return of get_public_key to be sure that the result will not be a bytes

### DIFF
--- a/salt/modules/x509.py
+++ b/salt/modules/x509.py
@@ -726,7 +726,7 @@ def get_public_key(key, passphrase=None, asObj=False):
         return evppubkey
 
     rsa.save_pub_key_bio(bio)
-    return bio.read_all()
+    return bio.read_all().decode()
 
 
 def get_private_key_size(private_key, passphrase=None):
--- a/salt/modules/x509.py
+++ b/salt/modules/x509.py
@@ -726,7 +726,7 @@ def get_public_key(key, passphrase=None, asObj=False):
         return evppubkey
 
     rsa.save_pub_key_bio(bio)
-    return bio.read_all().decode()
+    return bio.read_all().decode('ascii')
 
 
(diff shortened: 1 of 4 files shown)


arsiesys:
Fix #52026

KChandrashekhar:
@arsiesys Thanks for providing fix. Can you add a test case ? We are now requiring test cases to be written for all PRs

arsiesys:
> @arsiesys Thanks for providing fix. Can you add a test case ? We are now requiring test cases to be written for all PRs

Hello @KChandrashekhar, I am sorry but I don't really have the bandwidth and time to work on it for now :(. As I am not really comfortable with the tests currently deployed, it would take time to learn this part.

dwoz:
re-run full all

waynew:
@arsiesys just FYI, we're working on getting 2019.2.1 stable, and then we'll look at merging these changes in.

alxwr:
@arsiesys Please also have a look at  #52456. There are similar bugs in `modules.x509.read_crl` and `modules.x509.crl_verify`.

PR #52018, opened 2019-03-07
always encode printed data and write it to the raw fd

### LINKED ISSUE #51933: Minion locale "feature" resulting in unparsable json return value

`salt-ssh` would not report back, what it had done. It just says:
```
[ERROR   ] JSON Render failed for: b'{\n    "local": {\n        "jid": "20190301171055036639",\n        "return": {\n            "file_|-write-motd_|.....
[ERROR   ] Expecting value: line 1 column 1 (char 0)
```

The quick reader of course notices, that there's a `b'` in that string! This short piece is of course not `bytes`, but a `str(bytes)`, for example `value = "b'\"blabla\"'"` or even `value = str("blabla".encode())`.
Now the json parser refuses to understand what `b` means. But where does that `b` come from?

After very intense debugging sessions, I've figured out that that a **wrong locale** on a `salt-ssh` minion leads to this simple additional `b'...'` apart from otherwise fully valid json.

### Steps to Reproduce Issue
* Have `LANG="en_US"` on the minion (yes this is wrong but no way should other people need to undertake my debuggings to figure this out in the future)
* Update a file through `file.managed` and include some UTF-8 only codepoints (in practice these were special tab-highlighting characters in my `.vimrc`...)
* In the return-value, specifically the `file.managed`-diff, the unicode characters are returned (and will thus be printed).
* `salt.output.display_output` writes the data to the terminal, using `print`
* In this print, `sys.stdout.encoding` will be `ISO-8859-1` because of the wrong locale (`LANG=en_US`) and thus `print` is unable to print the characters, instead it raises an `UnicodeEncodeError`.
* The glorious `salt.utils.stringutils.print_cli` function has the "awesome" "feature" to do blind auto-encoding:
```python
            try:
                print(msg)
            except UnicodeEncodeError:
                print(msg.encode('utf-8'))
```
* This prints the `bytes` object to the terminal, which json of course can't understand

### Fix suggestions
`salt.output.display_output` should probably do this:
``` python
            sys.stdout.write(display_data)                                      
            sys.stdout.write("\n") 
```
instead of the funny `print_cli` function it actually does crash in case of encoding problems and not silently swallow them and cause infinite pain!

The error message would then be actually helpful and one can investigate why `latin-1` is chosen here.
```
[ERROR   ] JSON Render failed for: 
  File "/var/tmp/.me_3c8639_salt/salt-call", line 27, in <module>
    salt_call()
  File "/var/tmp/.me_3c8639_salt/pyall/salt/scripts.py", line 431, in salt_call
    client.run()
  File "/var/tmp/.me_3c8639_salt/pyall/salt/cli/call.py", line 57, in run
    caller.run()
  File "/var/tmp/.me_3c8639_salt/pyall/salt/cli/caller.py", line 155, in run
    _retcode=ret.get('retcode', 0))
  File "/var/tmp/.me_3c8639_salt/pyall/salt/output/__init__.py", line 134, in display_output
    sys.stdout.write(display_data)
UnicodeEncodeError: 'latin-1' codec can't encode characters in position 20811-20812: ordinal not in range(256)
[ERROR   ] Expecting value: line 1 column 1 (char 0)
```

Which is great btw because the `json` *tries to parse* the Python traceback.

Alternatively, one can even use:
``` python
        if display_data:                                                        
            encoded_data = display_data.encode('utf-8')                                
            sys.stdout.buffer.write(encoded_data)                               
            sys.stdout.buffer.write(b"\n")  
```

Then it actually **does work** even though the locale is wrong: The `encode()` function always uses `utf-8` and `stdout.buffer.write` just writes that data to the raw stdout without caring about encodings/the locale at all.

Of course another option would be an actual sanity check for the minion's locale....

Using the automatic byte-conversion-print of `print_cli` is just mental, though.

### Versions Report
```
% salt-ssh --versions-report
Salt Version:
           Salt: 2019.2.0
 
Dependency Versions:
           cffi: Not Installed
       cherrypy: Not Installed
       dateutil: 2.6.1
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.10
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.5.6
   mysql-python: Not Installed
      pycparser: Not Installed
       pycrypto: Not Installed
   pycryptodome: Not Installed
         pygit2: Not Installed
         Python: 3.6.7 (default, Oct 22 2018, 11:32:17)
   python-gnupg: 0.4.1
         PyYAML: 3.12
          PyZMQ: Not Installed
           RAET: Not Installed
          smmap: Not Installed
        timelib: Not Installed
        Tornado: 4.5.3
            ZMQ: Not Installed
 
System Versions:
           dist: Ubuntu 18.04 bionic
         locale: UTF-8
        machine: x86_64
        release: 4.15.0-33-generic
         system: Linux
        version: Ubuntu 18.04 bionic
```

TheJJ:
When using salt-ssh on a minion with a non-utf8 locale,
the `print_cli` function "automatically" encodes the data if `print`
fails:

``` python
try:
    print(msg)
except UnicodeEncodeError:
    print(msg.encode('utf-8'))
```

Because `json` is transported here, printing a `bytes` object will
lead to an unparsable `b'{"some": "json"}'` because of the starting `b'`.

This patch fixes the issue by always encoding the data as `utf-8` and
printing it to the raw file descriptor of `stdout`, thus bypassing
possible encoding errors because of some incompatible `sys.stdout.encoding`.

### What does this PR do?
Fix a bug where returned json-value of `salt-ssh` is not readable because the output was automatically encoded and then printed because printing it directly fails because of an incompatible `sys.stdout.encoding`.

Fixes #51933

### Previous Behavior
returned json output looked like `b'{"funny": "json"}'` which is unparsable because of the `b'`

### New Behavior
the returned json will look like `{"funny": "json"}`

### Tests written?

No

### Commits signed with GPG?

No

### DIFF
--- a/salt/output/__init__.py
+++ b/salt/output/__init__.py
@@ -125,7 +125,12 @@ def display_output(data, out=None, opts=None, **kwargs):
                     ofh.close()
             return
         if display_data:
-            salt.utils.stringutils.print_cli(display_data)
+            if six.PY2:
+                salt.utils.stringutils.print_cli(display_data)
+            else:
+                encoded_data = display_data.encode('utf-8')
+                sys.stdout.buffer.write(encoded_data)
+                sys.stdout.buffer.write(b"\n")
     except IOError as exc:
         # Only raise if it's NOT a broken pipe
         if exc.errno != errno.EPIPE:


TheJJ:
could anyone please have a look at this PR so we can bring this upstream?

TheJJ:
More than another month has passed now, is anyone willing to have a look at this?

waynew:
@TheJJ Thanks for the PR! Sorry for the wait - I'll take a look at this today!

salt/output/__init__.py
```diff
@@ -125,7 +125,12 @@ def display_output(data, out=None, opts=None, **kwargs):
                     ofh.close()
             return
         if display_data:
-            salt.utils.stringutils.print_cli(display_data)
+            if six.PY2:
+                salt.utils.stringutils.print_cli(display_data)
```
waynew: Would it make more sense to just fix the underlying `print_cli`?

I believe that it would be as simple as changing `print(msg.encode('utf-8'))` to `sys.stdout.buffer.write(msg)`. In either case, I'd love to see some tests for this - it should be pretty trivial to add something in the unit tests dir. Then I'll be a hearty 👍 on this.

waynew (changes requested):

salt/output/__init__.py
waynew: @saltstack/team-core any thoughts on whether this is the right place for the fix, or we should fix the underlying `stringutils.print_cli`?

TheJJ: I changed it here in order to be sure future improvements to `print_cli` won't affect the information transport, because `print_cli` sounds like something relevant for interactive stuff. Also, I didn't want to break other uses of the function unintentionally.

I'm of course open to suggestion for a better fix :)

waynew: @saltstack/team-core any thoughts?

TheJJ:
I think I'll give up now.

PR #52025, opened 2019-03-07
Update schedule.py

vguevara:
The documentation is incongruent with the module documentation. In the module the behaviour is stated as the fix  (https://docs.saltstack.com/en/latest/topics/jobs/#run-on-start), while the current doc suggest the schedule will not run at all if the value is set to False (which is wrong)

### Previous Behavior

### New Behavior

### Tests written?

Yes/No

### Commits signed with GPG?

Yes/No

### DIFF
--- a/salt/states/schedule.py
+++ b/salt/states/schedule.py
@@ -136,9 +136,9 @@ def present(name,
         Requires python-croniter.
 
     run_on_start
-        Whether the job will perform the first run when Salt minion starts,
-        or the job will skip it and wait until the next scheduled run.  Value
-        should be a boolean.
+        Whether the job will run when Salt minion starts, or the job will be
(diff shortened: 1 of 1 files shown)


dwoz (changes requested):
Something about this reads a little weird to me. Does this work?
```
Whether the job will run when Salt minion starts,
or the job will be skipped it and run at the next scheduled run.  Value
should be a boolean.
```

vguevara:
Perhaps better:
```
Whether the job will run when Salt minion starts,
or the job will be skipped **once** and run at the next scheduled run.  Value
should be a boolean.
```

dwoz:
@vguevara I've updated this to point to the `master` branch and have added the working change. We'll get this merged after the docs build passes.

dwoz (approved):

Issue #50667, opened 2018-11-28
Modified Minion installer for Windows is giving Timezone issue and not connecting with master.


nithinh123:
### Description of Issue/Question
Modified Minion installer for Windows is giving Timezone isseue and not connecting with master. Able to add minion key in Master, but not able to connect to minion after that. 
### Setup
Modified the Salt Minion installer for windows. Tried to debug the issue using the salt-minion-debug.bat
Getting the below error:

  File "C:\hivecenter\bin\lib\site-packages\salt\modules\win_timezone.py", line
235, in get_offset
    tz_object = pytz.timezone(get_zone())
  File "C:\hivecenter\bin\lib\site-packages\pytz\__init__.py", line 179, in time
zone
    raise UnknownTimeZoneError(zone)
UnknownTimeZoneError: 'Unknown'

### Steps to Reproduce Issue
Run the salt-minion-debug.bat in windows command line.

### Versions Report
Minion- 2017.7.2

dwoz:
@nithinh123 I do not fully understand the problem you are having. When you call the installer 'modified', does it mean you changed it in some way? If so What modifications did you make to the installer? Otherwise, please explain this further.

nithinh123:
@dwoz  I made changes to the service name, installer directory and installer UI and icons. It works well in Windows Server 2012. Facing issue in Windows Server 2008.

Steltek:
Seeing the same issue on a Win2k8 R2 server. Doesn't connect to master and is continually outputting these errors into the log:

`2019-03-06 10:45:12,377 [salt.minion      :1041][CRITICAL][5852] Unexpected error while connecting to sivour.dyndns.org
Traceback (most recent call last):
  File "c:\salt\bin\lib\site-packages\salt\minion.py", line 1018, in _connect_minion
    yield minion.connect_master(failed=failed)
  File "c:\salt\bin\lib\site-packages\tornado\gen.py", line 1055, in run
    value = future.result()
  File "c:\salt\bin\lib\site-packages\tornado\concurrent.py", line 238, in result
    raise_exc_info(self._exc_info)
  File "c:\salt\bin\lib\site-packages\tornado\gen.py", line 1063, in run
    yielded = self.gen.throw(*exc_info)
  File "c:\salt\bin\lib\site-packages\salt\minion.py", line 1216, in connect_master
    yield self._post_master_init(master)
  File "c:\salt\bin\lib\site-packages\tornado\gen.py", line 1055, in run
    value = future.result()
[...]

Issue #51455, opened 2019-01-31
Exception 'Schedule' object has no attribute 'loop_interval' occurred in scheduled job


afischer-opentext-com:
### Description of Issue/Question

Our Salt Minions (version 2018.3.3) are flood their log file with the following log line

```
2019-01-31 18:59:13,543 [salt.minion      :433 ][ERROR   ][4216] Exception 'Schedule' object has no attribute 'loop_interval' occurred in scheduled job
2019-01-31 18:59:14,230 [salt.minion      :433 ][ERROR   ][4216] Exception 'Schedule' object has no attribute 'loop_interval' occurred in scheduled job
2019-01-31 18:59:14,402 [salt.minion      :433 ][ERROR   ][4216] Exception 'Schedule' object has no attribute 'loop_interval' occurred in scheduled job
2019-01-31 18:59:14,528 [salt.minion      :433 ][ERROR   ][4216] Exception 'Schedule' object has no attribute 'loop_interval' occurred in scheduled job
2019-01-31 18:59:14,544 [salt.minion      :433 ][ERROR   ][4216] Exception 'Schedule' object has no attribute 'loop_interval' occurred in scheduled job
2019-01-31 18:59:15,230 [salt.minion      :433 ][ERROR   ][4216] Exception 'Schedule' object has no attribute 'loop_interval' occurred in scheduled job
2019-01-31 18:59:15,402 [salt.minion      :433 ][ERROR   ][4216] Exception 'Schedule' object has no attribute 'loop_interval' occurred in scheduled job
2019-01-31 18:59:15,528 [salt.minion      :433 ][ERROR   ][4216] Exception 'Schedule' object has no attribute 'loop_interval' occurred in scheduled job
2019-01-31 18:59:15,544 [salt.minion      :433 ][ERROR   ][4216] Exception 'Schedule' object has no attribute 'loop_interval' occurred in scheduled job
2019-01-31 18:59:16,230 [salt.minion      :433 ][ERROR   ][4216] Exception 'Schedule' object has no attribute 'loop_interval' occurred in scheduled job
2019-01-31 18:59:16,403 [salt.minion      :433 ][ERROR   ][4216] Exception 'Schedule' object has no attribute 'loop_interval' occurred in scheduled job
2019-01-31 18:59:16,528 [salt.minion      :433 ][ERROR   ][4216] Exception 'Schedule' object has no attribute 'loop_interval' occurred in scheduled job
[...]

PR #52046, opened 2019-03-07
Use registry to get timezone

twangboy:
### What does this PR do?
Uses the registry to get the timezone (reverts https://github.com/saltstack/salt/pull/51095)
Handles null characters in the registry.
Adds tests to handle null characters in the registry.

Registry is faster than shelling out commands:
```
timeit: cmd - 1000 runs
9.4037297
timeit: reg - 1000 runs
0.0563001
```
Since this is used for grains and is run often we need this to be fast.

https://github.com/saltstack/salt/issues/50667
https://github.com/saltstack/salt/issues/51940

### Tests written?
Yes

### Commits signed with GPG?
Yes


twangboy:
@afischer-opentext-com @arizvisa This should take care of the issue with null values. Please let me know if this solves the issues you are having.

afischer-opentext-com:
Doesn't it make sense to use the command as a fallback in case the registry lookup fails? That way we always would have a legit result.

twangboy:
The transition to `tzutil` was prompted because in some systems some the registry contained some strange data (https://github.com/saltstack/salt/issues/51455#issuecomment-461129536, https://github.com/saltstack/salt/issues/50667, https://github.com/saltstack/salt/issues/51940). In all cases the correct value was followed by a null character and then some additional 'junk' data. This PR will take everything up to the first null character as the timezone in those cases.

The transition to `tzutil` was causing incorrect results as well in instances where DST was disabled. The `tzutil` appends some data to the end in that case which couldn't map to the lookup dict. This was causing the second issue for some folks... or would in the next release.

arizvisa:
Yeah despite this specifically being related to a timezone, really it's just the semantics of the registry apis that's confusing. Fetching the Timezone from the registry is probably safe.

The oversight is that Windows allows you to write arbitrary data to any the registry types. So types such as REG_SZ (and even REG_DWORD), lets you write anything you want (its parameters are a pointer to your data and a length) due to the type really just being an enumeration and not a constraint.

The REG_SZ parameter even mentions that it must be null-terminated in the api documentation.
(https://docs.microsoft.com/en-us/windows/desktop/api/Winreg/nf-winreg-regsetvalueexa). When you get to languages that allow null-bytes in their "string", if the caller includes all the data returned from the api, a trailing null-byte(s) can be returned.

So the proper workflow is, if you want to read arbitrary data from the registry. There's no need to process anything. However.. if you want to read a "string" from the registry, it is up to you to terminate it at the null byte (unless your language already does it for you).

arizvisa:
Ftr, with the prior comment this implies that any place that a string is intended to be fetched from the registry has the potential to include null bytes at the end. As twangboy mentioned, this manifests itself in the timezone on some versions of Windows which can include more than one null byte.

Actually, after posting the prior comment I went to check the size of the timezone which turns out to be a common/constant size. I'm willing to bet the code that Windows was using for writing the timezone looks like:

```
char timezone[1024];
// initialize and populate timezone buffer
hr = RegSetValueEx(hKey, "TimeZoneInformation", 0, REG_SZ, timezone, sizeof(timezone));
...
```

Issue #51860, opened 2019-02-27
The win_dism module does not work on a 64-bit architecture when running inside a 32-bit process


arizvisa:
### Description of Issue/Question
On a 64-bit windows system, there are two versions of the `DISM.exe` binary. The first one (64-bit) is `/Windows/System32/Dism.exe` and the second one (32-bit) is `/Windows/SysWOW64/Dism.exe`. The `win_dism` module simply trusts that the PATH is set correctly as it just tries to execute "dism".

It is prudent to note that on 64-bit versions of windows, the directories for `/Windows/System32` and `/Windows/SysWOW64` have different semantics depending on the architecture of the application that is being used. So the path to the correct instance of `Dism.exe` is different depending on the Python interpreter that is running the salt process.

Normally it is assumed that `/Windows/SysWOW64` contains the 32-bit binaries (such as `Dism.exe`) and `/Windows/System32` contains the 64-bit binaries. However, when running a 32-bit application (such as the 32-bit version of Python) both the `/Windows/SysWOW64` and `/Windows/System32` directories point to the 32-bit binaries for the platform. This is done for compatibility reasons as historically the `/Windows/System32` directory contains the system tools available for the platform.

Unfortunately this means that via a 32-bit process, it would normally be impossible to call the correct 64-bit version of `Dism.exe`. However, on a 32-bit process windows creates a file-system redirection to the 64-bit platform tools at `/Windows/Sysnative`. So depending on the process architecture, the correct path can be either `/Windows/System32` or `/Windows/Sysnative`.

In summary, if one is running salt on a 32-bit version of Windows. Nothing matters, everything is okay.
If one is running salt on a 64-bit version of Windows, then the correct path to `Dism.exe` depends on the architecture of the Python interpreter that is running.

If salt is running in a 32-bit interpreter, then the path to the 64-bit `Dism.exe` is at `/Windows/Sysnative/Dism.exe`. If salt is running in a 64-bit interpreter, then the path to the 64-bit `Dism.exe` is at `/Windows/System32`.

### Symptom (in minion log file)
```
2019-02-27 09:29:52,104 [salt.minion      :1667][WARNING ][2740] The minion function caused an exception
2019-02-27 09:30:02,510 [salt.loaded.int.module.cmdmod:774 ][ERROR   ][3132] Command '[u'DISM', u'/English', u'/Online', u'/Get-Features']' failed with return code: 11
[...]

PR #52065, opened 2019-03-08
Use the dism binary that matches system architecture

twangboy:
### What does this PR do?
Uses the dism binary that matches the system architecture.

https://github.com/saltstack/salt/issues/51860

### Previous Behavior
Dism module would fail if run from 32bit salt on a 64bit system

### New Behavior
Module functions as expected

### Tests written?
No

### Commits signed with GPG?
Yes


dwoz (approved):

twangboy:
@arizvisa Updated with your changes. Thanks!

PR #52117, opened 2019-03-08
Add jenkins test failures script

dwoz:
### What does this PR do?

Adding a script to dump test failures from Jenkins branch jobs.

### Commits signed with GPG?

Yes


Ch3LL (changes requested):

garethgreenaway (approved):

dwoz:
I don't think we need this anymore. I will leave the branch around for a while just in case.

Issue #52118, opened 2019-03-08
GCE doesn't have support for Labels


tehsu:
### Description of Issue/Question
When deploying a server, Labels is set to None, I tried to use metadata and it sets it as metadata, I was able to fix it by adding support to gce.py.

### Setup
cloud profile

cent-7-test:
  image: centos
  size: n1-standard-1
  location: 
  network: 
  subnetwork: 
  use_persistent_disk: True
  delete_boot_pd: False
  deploy: True
  make_master: False
  provider: 
  ssh_interface: private_ips
  ssh_username: 
  ssh_keyfile:
  external_ip: "None"
  labels: '{"environment": "test"}'

### Steps to Reproduce Issue
Set labels as above and nothing is set on the server

### Versions Report
Salt Version:
           Salt: 2019.2.0
 
Dependency Versions:
           cffi: 1.6.0
       cherrypy: unknown
       dateutil: 1.5
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.7.2
        libgit2: 0.26.3
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.5.6
   mysql-python: Not Installed
      pycparser: 2.14
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: 0.26.4
         Python: 2.7.5 (default, Oct 30 2018, 23:45:53)
   python-gnupg: Not Installed
         PyYAML: 3.12
          PyZMQ: 15.3.0
           RAET: Not Installed
          smmap: Not Installed
        timelib: Not Installed
        Tornado: 4.2.1
            ZMQ: 4.1.4
 
System Versions:
           dist: centos 7.6.1810 Core
         locale: UTF-8
        machine: x86_64
        release: 3.10.0-862.3.2.el7.x86_64
         system: Linux
        version: CentOS Linux 7.6.1810 Core

Ch3LL:
thanks for the PR! Since the PR fixes this i will go ahead close here and continue conversation in PR. thanks again!

PR #52119, opened 2019-03-08
Add labels support for GCE

tehsu:
### What does this PR do?
It adds support for Labels into Google Cloud

It allows us to create VMs with labels, Issue #52118

### Previous Behavior
Labels were always set to None

### New Behavior
Able to set key value for labels

### Tests written?
No

### Commits signed with GPG?
No


gtmanfred (approved):

xeacott:
Hi @tehsu since your changes introduce a new function, you'll need to also give it a test (unit/integration). I can find some examples for you if you need any!

salt/cloud/clouds/gce.py
```diff
@@ -400,6 +400,24 @@ def __get_tags(vm_):
...
+    Get configured labels.
+    '''
+    lbl = config.get_cloud_config_value(
+        'labels', vm_, __opts__,
+        default='[]', search_global=False)
+    # Consider warning the user that the labels in the cloud profile
+    # could not be interpreted, bad formatting?
+    try:
+        labels = literal_eval(lbl)
+    except Exception:  # pylint: disable=W0703
```
xeacott: Here you should be catching the exact exception that occurs, otherwise we may want to handle different exceptions with different cases

salt/cloud/clouds/gce.py
```diff
@@ -400,6 +400,24 @@ def __get_tags(vm_):
     return tags
 
 
+def __get_labels(vm_):
```
xeacott: Since this function takes a param it should be stated in the docstring what this param is.

xeacott (changes requested):
See comments in files changed.

salt/cloud/clouds/gce.py
gtmanfred: It doesn't need to for a private function.

Plus the vm_ is standard in salt-cloud

salt/cloud/clouds/gce.py
gtmanfred: This is the exact same block that is used elsewhere for all the other gce functions, i would leave it as is.

xeacott: Is this the rule of thumb here - I am asking so I know what is considered good or bad practice for the salt code base.

gtmanfred: The rule of thumb is usually be consistent in that module.

Good luck trying to get stuff to be consistent on the whole code base.  If we feel like this is wrong, then we should write a pylint plugin for it.

In reality, we don't really care what the lower exception is, because we just know it was unable to get labels for that vm.  It could be any one of a bunch of them, better to just catch them and throw it away and return that there were no labels.

xeacott: Gotcha, thanks for giving some insight

xeacott (changes requested):
Just following up on this PR, let me know if you need anything else!

tehsu:
> Just following up on this PR, let me know if you need anything else!

Will get the test case updated, thanks!

tehsu:
still need to add test

sagetherage:
@tehsu this needs to be rebased against the master branch, we aren't accepting changes to the develop branch any longer.

sagetherage:
@tehsu this needs to be rebased against the master branch, we aren't accepting changes to the develop branch any longer.

sagetherage:

Issue #50660, opened 2018-11-27
win_lgpo.py slow to execute


dan-rus:
When using states in win_lgpo.py, it is very slow to complete each check.

If I add some warning messages at key points of the set_ function in the state folder's win_lgpo, I can see the following if I have a two states each with one configuration item:
[WARNING ] 0.0 : Start win_lgpo.py set_ function call
[WARNING ] 0.0529649065866 : About to call lgpo.get_policy_info
[WARNING ] 0.791142923417 : Returned from lgpo.get_policy_info
[WARNING ] 0.842564259418 : About to call lgpo.get
[WARNING ] 3.94409042198 : Returned from lgpo.get
[WARNING ] 3.99761096055 : End win_lgpo.py set_ function call
[WARNING ] 4.05679556171 : Start win_lgpo.py set_ function call
[WARNING ] 4.10796445914 : About to call lgpo.get_policy_info
[WARNING ] 4.80997786822 : Returned from lgpo.get_policy_info
[WARNING ] 4.85807965928 : About to call lgpo.get
[WARNING ] 7.70147202406 : Returned from lgpo.get
[WARNING ] 7.74824216828 : End win_lgpo.py set_ function call

The lgpo.get call takes approximately 3 seconds every time, and the lgpo.get_policy_info call which made for each configuration item in the state takes approximately 0.5 seconds, so you can roughly guess how long a call is going to take.  The end result of it in my use-case, is "Total run time:  201.482 s" for 92 states (not all are done via win_lgpo).

If I memoize \_processPolicyDefinitions (It only appears to be reading and parsing the same file over and over), and rewrite 'lgpo.get' so that it only gets the configuration items I am targetting (Currently it gets all GPO settings even if I am only comparing one), then the runtime comes down to:
[WARNING ] 0.0 : Start win_lgpo.py set_ function call
[WARNING ] 0.048824824996 : About to call lgpo.get_policy_info
[WARNING ] 0.788979609464 : Returned from lgpo.get_policy_info
[WARNING ] 0.836328982218 : About to call lgpo.get
[WARNING ] 0.891362816556 : Returned from lgpo.get
[WARNING ] 0.94043829925 : End win_lgpo.py set_ function call
[WARNING ] 0.988341968335 : Start win_lgpo.py set_ function call
[WARNING ] 1.03744282845 : About to call lgpo.get_policy_info
[WARNING ] 1.08731168951 : Returned from lgpo.get_policy_info
[...]

PR #52124, opened 2019-03-09
Optimize __context__ in LGPO module

twangboy:
### What does this PR do?
Instead of popping everything out of `__context__` when a single item changes, just change that item in `__context__`... in an attempt to speed things up. When you pop the whole section out, it has to be regenerated when called again. This will only speed up states with multiple LGPO calls.

This was done for `netsh` and `secedit` data.

https://github.com/saltstack/salt/issues/50660

### Tests written?
No

### Commits signed with GPG?
Yes


lomeroe (approved):
LGTM

waynew (approved):

Issue #8875, opened 2013-11-28
Support relative includes for pillar


pniederw:
The dot notation that's supported for relative includes in state files  (`include: - .other`) doesn't seem to work for pillar. Would be just as useful there.

wuxxin:
+1 
because this way you can write:
----
something/common.sls
{% load_yaml as common %}
test:
   whatever: true
   another:
      hu: he
{% endload %}
----
something/another.sls
include:
  - .common

{{ common.something }}

uvsmtid:
To complement this issue, this is how to live now with workaround.

## Workaround ##

Push a variable with location of the pillar into the pillar template being included using `defaults` as described in [doc how to include pillar](http://docs.saltstack.com/en/latest/topics/pillar/#including-other-pillars).

#### Example ####

The top file is not used, but it's important to note that it cannot use `include` directly.

All we need to do is to list the first "entry point" pillar which will `include` others.
```
# top.sls
base:
    '*':
        - main
```
Include pillar and tell it where it is located using `this_pillar`.
```
# main.sls
include:
    - deep.inside.directory.structure.pillar_a:
        defaults:
            this_pillar: deep.inside.directory.structure.pillar_a
```
Include other pillars relatively:
```
# deep.inside.directory.structure.pillar_a.sls
include:
    - {{ this_pillar }}.whatever:
    - {{ this_pillar }}.blah_blah_blah:
    - {{ this_pillar }}.last_one:
```

#### Proper complication ####

Because you don't know (or you don't want to keep this in mind) whether any pillar subsequently includes relatively anything else or not, the safest way is to pass `this_pillar` recursively to every other one being included (just in case).
```
# deep.inside.directory.structure.pillar_a.sls
include:
    - {{ this_pillar }}.whatever:
        defaults:
            this_pillar: {{ this_pillar }}.whatever
    - {{ this_pillar }}.blah_blah_blah:
        defaults:
            this_pillar: {{ this_pillar }}.blah_blah_blah
    - {{ this_pillar }}.last_one:
        defaults:
            this_pillar: {{ this_pillar }}.last_one
```
It's ugly, but it works.

titilambert:
+1

enver:
+1

adithep:
+1

peter-slovak:
+1

damon-atkins:
+1

mlalpho:
+1

damon-atkins:
+1

[...]

PR #52156, opened 2019-03-12
Allow pillar relative includes

waynew:
### What does this PR do?

Implements longstanding request #8875 - allows for pillar relative includes

#8875 - there are probably others

### Previous Behavior

A leading dot in a pillar include was a nop. If you had a pillar like:

**foo/bar/baz.sls**
```
include:
  - .bang
```

That would not refer to `foo/bar/bang.sls`, but `bang.sls` at the pillar root.

### New Behavior

Now leading dots in includes refer to relative imports :tada:

### Tests written?

Yes

### Commits signed with GPG?

Yes

---

I'm operating under the assumption that https://github.com/saltstack/salt/pull/52008 will be merged.. forward?


dmurphy18 (approved):

mchugh19:
Can/should this be added to release notes?

max-arnold:
@waynew It would be nice if this feature worked exactly as in states, i.e. multiple dots should also work: https://docs.saltstack.com/en/latest/ref/states/include.html#relative-include

```
include:
  - ..http
  - ...base
```

Also, no docs at all 😞

waynew:
@max-arnold Do pillar relative includes work that way?

max-arnold:
@waynew 
> Do pillar relative includes work that way?

No, **pillar** includes do not work that way, but it makes sense to implement because **state** includes support this, and feature symmetry is generally a good thing. See the states doc I mentioned earlier: https://docs.saltstack.com/en/latest/ref/states/include.html#relative-include

waynew:
Wow, I'm not sure *at all* what I meant there 🙃 

I think it's probably worth creating an issue/feature request. I don't know if it's significant enough to create a SEP, but I agree with the feature symmetry point.

waynew:
When porting this to master it should be noted that it closes #56186

Issue #52091, opened 2019-03-08
integration.runners.test_nacl.NaclTest.test_sealedbox_enc_dec


dwoz:
```
integration.runners.test_nacl.NaclTest.test_sealedbox_enc_dec
--------------------------------------------------------------------------------
develop failed salt-windows-2019-py2
--------------------------------------------------------------------------------
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
u'pk' not found in {u'retcode': 70}
................................................................................

Traceback (most recent call last):
  File "c:\users\admini~1\appdata\local\temp\kitchen\testing\tests\integration\runners\test_nacl.py", line 100, in test_sealedbox_enc_dec
    self.assertIn('pk', ret['return'])
AssertionError: u'pk' not found in {u'retcode': 70}
```

PR #52158, opened 2019-03-12
Fix runner tests

twangboy:
### What does this PR do?
Fixes failing runner tests
Skips test in test_state.py that was causing pickling errors on Windows
Comment on another test that fails on Windows

Need to get more eyes on this. I don't understand the purpose of the code I removed. But it was causing all the failures on the runner tests. Valid return was being replaced with a dict with a single entry, retcode.

Perhaps @isbm could shed some light on it.

https://github.com/saltstack/salt/issues/52091

### Tests written?
Yes

### Commits signed with GPG?
Yes


cachedout:
IIRC this was the same code that @mattp- was concerned with the other day.

isbm:
Leaves me wondering how that got in. Clearly my commit, according to blame, but more looks like a rebasing mistake.

mattp-:
@dwoz I think the change in https://github.com/saltstack/salt/pull/52258 might make more sense as a fallback when ret is an empty dict, just to ensure a failure retcode is present when something definitely went wrong.

dwoz (approved):

dwoz:
re-run centos-7

Ch3LL:

Issue #51959, opened 2019-03-04
Changes with Linux ACL between 2018.3.3 and 2018.3.4


knine:
### Description of Issue/Question

It appears the acl.present state has a bug between versions 2018.3.3 and 2018.3.4. The following output is showing up during a test run in 2018.3.4:

          ID: Starfleet FACL
    Function: acl.present
        Name: /path/to/directory
      Result: None
     Comment: Updated permissions will be applied for starfleet: 5 -> rx
     Started: 09:11:27.079996
    Duration: 24350.106 ms
     Changes:   
              ----------
              new:
                  ----------
                  acl_name:
                      starfleet
                  acl_type:
                      group
                  perms:
                      rx
              old:
                  ----------
                  acl_name:
                      starfleet
                  acl_type:
                      group
                  perms:
                      5
This does not show up if I roll back to 2018.3.3. My guess is under the hood it is not seeing that "rx" is the same as "5".  I can apply the changes, and it does it, but then the message re-appears again during a test run. I looked at the docs for the state and "perms" uses rwx, not numeric.

### Setup

    'Starfleet FACL':
      acl.present:
        - name: /path/to/directory
        - acl_type: group
        - acl_name: starfleet
        - perms: rx
        - recurse: True
        - require:
          - pkg: 'Starfleet Packages'

### Steps to Reproduce Issue
Update from 2018.3.3 to 2018.3.4

### Versions Report

Has issue:

    Salt Version:
               Salt: 2018.3.4
     
    Dependency Versions:
               cffi: Not Installed
           cherrypy: Not Installed
           dateutil: 1.5
              docker-py: 1.10.6
              gitdb: Not Installed
          gitpython: Not Installed
              ioflo: Not Installed
             Jinja2: 2.7.2
            libgit2: Not Installed
            libnacl: Not Installed
           M2Crypto: 0.31.0
               Mako: Not Installed
       msgpack-pure: Not Installed
     msgpack-python: 0.5.6
       mysql-python: Not Installed
          pycparser: Not Installed
           pycrypto: 2.6.1
       pycryptodome: Not Installed
             pygit2: Not Installed
             Python: 2.7.5 (default, Sep 12 2018, 05:31:16)
       python-gnupg: Not Installed
             PyYAML: 3.11
              PyZMQ: 15.3.0
               RAET: Not Installed
              smmap: Not Installed
            timelib: Not Installed
            Tornado: 4.2.1
                ZMQ: 4.1.4
     
    System Versions:
[...]

PR #52172, opened 2019-03-14
[2018.3] Changes to linux_acl state.

garethgreenaway:
### What does this PR do?
Fixing the output when there are changes for the ACL state so the permissions are shown and not the octal number.

#51959 

### Previous Behavior
Changes would display the octal number for the old value.

### New Behavior
Now displays the permissions, eg. r, rw, rwx, etc.

### Tests written?
Yes.  Tests updated.

### Commits signed with GPG?

Yes/No


waynew:
Just curious - does it make more sense to do this approach or something like:

```
>>> lookup = {0: '-', 1: 'r', 2: 'w', 4: 'x'}
>>> for num in range(8):
...  print('{}{}{}'.format(lookup[num&1], lookup[num&2], lookup[num&4]))
...
---
r--
-w-
rw-
--x
r-x
-wx
rwx
```
?

waynew (approved):
👍

dwoz (approved):

Issue #50462, opened 2018-11-11
How to append string with the new __slot__ syntax?


UniIsland:
### Description of Issue/Question

The `__slot__` syntax #44866 is great. But can we manipulate the string returned, or append another string to it?

I tried the following states syntax. Neither of them works.

```
tmpfile:
  file.managed:
    - name: __slot__:salt:cmd.run('gem env gemdir') ~ `/cache/tmp`
    - contents: example content
```

```
tmpfile:
  file.managed:
    - name: __slot__:salt:cmd.run('gem env gemdir')/cache/tmp
    - contents: example content
```

After reading the source code [salt/state.py#2054](https://github.com/saltstack/salt/pull/44866/files#diff-e5cf22153489a6c40864babf3ca18002R2054), I guess anything after the `)` is ignored.

Is there a way to append sth to the `__slot__` part and use the result in the states file? 

### Versions Report
salt-minion 2018.3.3 (Oxygen)

max-arnold:
@UniIsland You could write a custom grain for this. If you still want to use slots, this workaround could help:

```
- name: __slot__:salt:cmd.run('sh -c "echo $(gem env gemdir)/cache/tmp"')
```

max-arnold:
This feature was added in Salt Neon

Issue #46771, opened 2018-03-29
[Feature Request] inline slot functions (expressions)


srkunze:
based on the observation here: #46516

Salt already provides 1000s of useful standard functions and modules. I actually would want to reuse all of them as slots.

However, as soon as they return something which I cannot use directly, e.g. because it's a dict (like **user.info**), I need to write a whole custom function for a thing such as small as a attribute access.

It would be great if salt provides a way to do inline slot functions in jinja templates, to enable this:

```
__slot__:user.info(user_name).home:
  ...
```

Syntax-wise maybe the colon could be a separator to signify the end of the expression.

gtmanfred:
@DmitryKuzmenko when you have a second can you take a look at this?

Thanks,
Daniel

DmitryKuzmenko:
@srkunze I'm not sure what do you mean saying "in jinja templates"? Templates are rendered earlier. Slots are executed right before the state execution. Slots are the last step.

If you just want to use a slot value as a part of state value like you've described in #46516: technically it's possible we just should update the syntax declaration and the slots parser.
@thatch45 what do you think?

srkunze:
@DmitryKuzmenko I said jinja template because that's where I would use slots. 🙂

guettli:
@DmitryKuzmenko are there any reason making `__salt__` availlable in slots (without writing a custom function) is a bad idea? 

I personally like the syntax of the first comment of this issue.

DmitryKuzmenko:
I agree this feature proposal looks interesting and useful.

mchugh19:
To update on this, the requested functionality was implemented in Neon. 
`__slot__:salt:user.info("root").home` should work as desired

srkunze:
Nice!

PR #52187, opened 2019-03-14
draft support for parsing slot results and appending text

mchugh19:
### What does this PR do?
Extends slot syntax to support parsing resulting dictionaries and appending text. Note: the appending process only functions on slot results which are strings.

Next steps:
- Since python structures are now injected with raw unicode: `{u'shell': u'/bin/bash'}` we could also force to json output if desired.

#50462
#46771

### Previous Behavior
Using `__slots__` with modules which return dictionaries required custom modules to parse result
Also, unable to append text

### New Behavior
Support dictionary results by appending the slot function with period delimited dictionary keys to traverse
`__slot__:salt:user.info(root).shell` returns `/bin/bash`

Support appending data by using a tilde delimiter followed by text
` __slot__:salt:user.info(root).shell ~ /appended~` returns `/bin/bash/appended~`

### Tests written?
Yes

### Commits signed with GPG?
No


dwoz (changes requested):
@mchugh19 Can you please add a test or two that cover this?

mchugh19:
Tests have been added for the traverse and append functions. 

In thinking about this a little bit more, the parsing for the append function looks for a tilde character and appends everything afterwards. It is possible that the slot text itself could contain that character as you might have `cmd.run('ls ~user')` or something. Should this behavior be updated to a more complicated regex?

mchugh19 (commented):
Tests added

dwoz (approved):

twangboy (approved):

mchugh19:
@DmitryKuzmenko you should take a look here as well

DmitryKuzmenko (approved):

thatch45 (approved):

PR #52191, opened 2019-03-14
Fix issue where version doesn't detect when check_remote=True

twangboy:
### What does this PR do?
The `chocolatey.list` function now returns a CaseInsensitiveDict so case is ignored.
Fixes issues where `__context__` and `__salt__` were being used incorrectly
Optimize the `_find_chocolatey` function

https://github.com/saltstack/salt/issues/46150

### Tests written?
No

### Commits signed with GPG?
Yes


s0undt3ch (approved):

dwoz (commented):
@twangboy Are there tests for this module?

PR #52195, opened 2019-03-14
Use encoding when caching pillar data

### LINKED ISSUE #51854: Fluorine: minion_pillar_cache: True leads to exception
After updating to the Fluorine and switching to python36 I run on the problem with `salt-call`. Any execution of the `salt-call` with any module leads to the this stackstrace:
https://gist.github.com/Oloremo/8c1d9da8a66342f0eb2354442825d9ce

Simplest example is somethinf like `cmd.run ls`

I pinpoint this problem to a single minion option - `minion_pillar_cache: True`. Setting it to False solves the issue. 

### Steps to Reproduce Issue
I have a very customized Salt deployment, it's a Centos 7 with python36 from EPEL and salt is installed via pip and runs from non-root user so I'm not sure if it's reproducible on a normal deployment.

But I was able to reproduce it using a docker to mimic mine deployment. So if you like or if you can't reproduce it on your env, you could download docker project from here: https://www.dropbox.com/s/b5j5qp55e0c622j/salt-bug.tar.gz?dl=0
unpack it in some dir and follow 3 steps instruction(included in README file):
```
1) cd into root of this directory
2) docker build -t salt-bug .
3) docker run -it salt-bug bash /start-daemons.sh
```

### Versions Report
```
Salt Version:
           Salt: 2019.2.0

Dependency Versions:
           cffi: Not Installed
       cherrypy: Not Installed
       dateutil: Not Installed
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.10
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.6.1
   mysql-python: Not Installed
      pycparser: Not Installed
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: Not Installed
         Python: 3.6.6 (default, Aug 13 2018, 18:24:23)
   python-gnupg: 0.4.4
         PyYAML: 3.13
          PyZMQ: 18.0.0
           RAET: Not Installed
          smmap: Not Installed
        timelib: Not Installed
        Tornado: 4.5.3
            ZMQ: 4.3.1

System Versions:
           dist: centos 7.4.1708 Core
         locale: UTF-8
        machine: x86_64
        release: 3.10.0-693.11.6.el7.x86_64
         system: Linux
        version: CentOS Linux 7.4.1708 Core
```

waynew:
### What does this PR do?

Fixes #51854 - sets the encoding when using the yaml dump

#51854 

### Previous Behavior

Caching just used the default encoding, which broke on Python3 where it expected to be writing bytes out to the file.

### New Behavior

Now it specifies the encoding so dump expects the pillar in the form that it is.

### Tests written?

Yes

### Commits signed with GPG?

Yes


twangboy (commented):
This looks like a unit test with all the mocking

waynew:
Those are just pieces that are irrelevant to the test. Opportunity to refactor maybe?

salt/minion.py
```diff
@@ -819,11 +819,11 @@ def __init__(self, opts):
                 penv = 'base'
             cache_top = {penv: {self.opts['id']: ['cache']}}
             with salt.utils.files.fopen(ptop, 'wb') as fp_:
-                salt.utils.yaml.safe_dump(cache_top, fp_)
+                salt.utils.yaml.safe_dump(cache_top, fp_, encoding='utf-8')
```
cachedout: What happens when the data is in fact in another encoding? Would it be better to use `__salt_system_encoding__` here?

waynew: That's probably a good idea 👍

dwoz: I think we need too be careful about using `__salt_system_encoding__` here. There is a very good chance that will break on windows where the system encoding will be `cp1252`. In other places we try the system encoding then fall back to `UTF-8` if there is an exception. However in this case, since this is a cache file that is internal to salt, I actually think it's better to stick with `UTF-8`. Salt is the only thing reading/writing the file.

dwoz (changes requested):
See my utf-8 comment and the linter.

salt/minion.py
waynew: @dwoz apparently I need to dig a little further, but I don't see that we're just reading the pillar data using a block like this, so it's very possible that we're using the system encoding to read the pillar data. I'm looking into that.

Oloremo:
@waynew Sorry, I missed this. Just checked - it fixes my issue!

waynew:
@dwoz lint issue is fixed, also it turns out there exists an SLS_ENCODING already, so I changed to using that.

dwoz (approved):

waynew:
re-run all

PR #42317, opened 2017-07-14
Refactoring Beacons

garethgreenaway:
### What does this PR do?
Refactoring beacons to ensuring all of them use list based configuration rather than dictionary based configuration.

### What issues does this PR fix or reference?
Updates all beacons, execution module and state module for list based configuration to be more in line with state files.

### Previous Behavior
Previously some beacons used dictionary based configuration while others used list based configuration.  

### New Behavior
Ensuring all configurations are the same type.

### Tests written?
Yes

Please review [Salt's Contributing Guide](https://docs.saltstack.com/en/latest/topics/development/contributing.html) for best practices.

PR #35374, opened 2016-08-11
Allow for additional module validation functions.

cachedout:
- Enforces validation on load for beacons.

Refs #34434

PR #52201, opened 2019-03-15
Rename obsolete beacon validation function

max-arnold:
### What does this PR do?

Small follow-up bugfix for #42317 to rename two remaining instances of beacon validation function from `__validate__` to `validate`:

* Use proper function name in the docs
* Rename validation function in the `aix_account` beacon (all other beacons are ok)

### Previous Behavior

`aix_account` beacon validator was ignored

### New Behavior

`aix_account` beacon validator should be called. 

### Tests written?

No. I don't have any AIX machine at hand :)

### DIFF
--- a/doc/topics/beacons/index.rst
+++ b/doc/topics/beacons/index.rst
@@ -305,7 +305,7 @@ the minion. The ``beacon`` function therefore cannot block and should be as
 lightweight as possible. The ``beacon`` also must return a list of dicts, each
 dict in the list will be translated into an event on the master.
 
-Beacons may also choose to implement a ``__validate__`` function which
+Beacons may also choose to implement a ``validate`` function which
 takes the beacon configuration as an argument and ensures that it
 is valid prior to continuing. This function is called automatically
 by the Salt loader when a beacon is loaded.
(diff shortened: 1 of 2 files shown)


max-arnold:
@garethgreenaway Aside from this fix, it looks like #35374 is no longer used and can be removed. I can try to roll it back if that is true.

garethgreenaway:
@max-arnold Thanks!  PR looks good.  You are correct, the `__validate__` was renamed back to `validate` as it's called by the various beacon execution & state execution modules when a beacon is added.

max-arnold:
Do you think it is worth undoing #35374? I can submit it as a separate PR, but I need some reassurance that it is not used somewhere else (like Salt Enterprise).

garethgreenaway:
@max-arnold The changes in that PR should have already been undone.  Do you see a change in that PR that still exists somewhere in the code?  All the beacons, except for the aix beacon, have their validate functions in the correct state.  The only thing that still remains is probably the virtual_funcs, which I think we can safely leave as it might find a use at a later date.

max-arnold:
Yep, I'm talking specifically about the `virtual_funcs`.

PR #52209, opened 2019-03-15
Handle Chocolatey's new Enhanced Exit Codes

twangboy:
### What does this PR do?
Chocolatey has started implementing Enhanced Exit Codes that will effect the stability of the Chocolatey module as future versions are released. This PR handles Enhanced Exit Code `2` added to the Chocolatey list function.

https://github.com/chocolatey/choco/issues/1758

### Tests written?
No

### Commits signed with GPG?
Yes


garethgreenaway:
@twangboy unless this is a CVE, this should go into 2018.3 and we can merge forward.

twangboy:
@garethgreenaway You're right...

waynew (approved):
Not sure if we care to use https://pypi.org/project/enum34/ but I, for one, really like using the Enum class (or IntEnum) from Python 3.4+. It will be nice to have that available. Then we could have something like:

```
from enum import IntEnum

class ChoclateyExitCode(IntEnum):
    success = 0
    error = 1
    no_result = 2

    @property
    def failed(self):
        return self not in (ExitCode.success, ExitCode.no_result)


# And then used like
if ChoclateyErrorCode(result['retcode']).failed:
    raise CommandExecutionError(...)
```

There are a number of places having some slightly more user friendly status codes would be nice... but I can wait :slightly_smiling_face:

PR #52213, opened 2019-03-16
[2018.3] Don't save beacons when test=True

garethgreenaway:
### What does this PR do?
Don't save beacons when test=True

N/A

### Previous Behavior
When test=true beacons would be saved to /etc/salt/minion.d/beacons.conf

### New Behavior
When test=True we shouldn't save.

### Commits signed with GPG?
Yes


salt/states/beacon.py
```diff
@@ -100,8 +100,11 @@ def present(name,
                 ret['comment'].append('Adding {0} to beacons'.format(name))
 
     if save:
-        result = __salt__['beacons.save']()
-        ret['comment'].append('Beacon {0} saved'.format(name))
+        if 'test' in __opts__ and __opts__['test']:
```
waynew: What about
```suggestion
if __opts__.get('test'):
```
Or if you wanted to be explicit,
```
if __opts__.get('test', False):
```
But (as you're no doubt aware) `.get` defaults to the falsey `None`.

salt/states/beacon.py
```diff
@@ -146,8 +149,11 @@ def absent(name,
         ret['comment'].append('{0} not configured in beacons'.format(name))
 
     if save:
-        result = __salt__['beacons.save']()
-        ret['comment'].append('Beacon {0} saved'.format(name))
+        if 'test' in __opts__ and __opts__['test']:
```
waynew: Same here,
```suggestion
if __opts__.get('test'):
```

waynew (changes requested):

waynew (approved):

dwoz (approved):

PR #52221, opened 2019-03-18
[develop] First pass at adding mod_aggregate support to the pip_state module.

garethgreenaway:
### What does this PR do?
First pass at adding mod_aggregate support to the pip_state module.

N/A

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/states/pip_state.py
+++ b/salt/states/pip_state.py
@@ -1065,3 +1065,51 @@ def uptodate(name,
         ret['comment'] = 'Upgrade failed.'
 
     return ret
+
+
+def mod_aggregate(low, chunks, running):
+    '''
+    The mod_aggregate function which looks up all packages in the available
+    low chunks and merges them into a single pkgs ref in the present low data
+    '''
+    pkgs = []
+    pkg_type = None
+    agg_enabled = [
+        'installed',
+        'removed',
+    ]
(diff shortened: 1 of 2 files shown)


salt/states/pip_state.py
```diff
@@ -1065,3 +1065,51 @@ def uptodate(name,
         ret['comment'] = 'Upgrade failed.'
 
     return ret
+
+
+def mod_aggregate(low, chunks, running):
+    '''
+    The mod_aggregate function which looks up all packages in the available
+    low chunks and merges them into a single pkgs ref in the present low data
+    '''
```
twangboy: Some docs here for the args... unless you wanna make it a hidden function. (`_mod_aggregate`)

twangboy (changes requested):

twangboy:
And can we get some tests?

salt/states/pip_state.py
garethgreenaway: This is the same doc string that exists in the iptables and pkg state modules, it's not a function that is intended to be called directly but since it's called outside of the state module it can't be a private function.

twangboy (approved):

garethgreenaway:
Test failure fixed with https://github.com/saltstack/salt/pull/52279

Issue #51865, opened 2019-02-27
salt.match.search_by resulting in 'None' since 2019.2


magenbrot:
### Description of Issue/Question

I'm using a setup as described in https://github.com/saltstack/salt/issues/39457#issuecomment-282506189 to apply roles to my minions. This worked fine until I upgraded to 2019.02. 

It seems like the match.search_by isn't working correctly, or maybe it's something I've missed in the release notes. Here's what I have (I've added log.debug for some variables):

### Setup
pillar/roles.jinja
```
{% load_yaml as role_defs %}
salt-master:
  - balder.mydomain.com

cloud:
  - balder.mydomain.com
  - heimdall.mydomain.com
  - loki.mydomain.com

weechat:
  - balder.mydomain.com
{% endload %}

{% set roles = salt.match.search_by(role_defs) %}

# testing
{% set test_roles = role_defs | yaml() %}
{%- do salt.log.debug('MYDEBUG-role_defs: ' ~ role_defs) %}
{%- do salt.log.debug('MYDEBUG-test: ' ~ test_roles) %}
{%- do salt.log.debug('MYDEBUG-roles: ' ~ roles) %}
```

pillar/roles.sls
```
{% from "roles.jinja" import roles with context %}
roles: {{ roles | yaml() }}
```

pillar/top.sls
```
{% from "roles.jinja" import roles with context %}

base:
  '*':
    - roles
    - users
```

### Steps to Reproduce Issue

Setup like above and run "salt 'balder*' pillar.items roles". roles pillar is empty, but should contain this minions roles instead:
```
balder.mydomain.com:
    ----------
    retcode:
        0
    roles:
        None
jid:
    20190227143040611412
retcode:
    0
```

This is the "extra" debug output:
```
2019-02-27 14:22:24,042 [salt.loaded.int.module.logmod:41  ][DEBUG   ][326] MYDEBUG-role_defs: {u'weechat': [u'balder.mydomain.com'], u'salt-master': [u'balder.mydomain.com'], u'cloud': [u'balder.mydomain.com', u'heimdall.mydomain.com', u'loki.mydomain.com']}
2019-02-27 14:22:24,042 [salt.loaded.int.module.logmod:41  ][DEBUG   ][326] MYDEBUG-test: {cloud: [balder.mydomain.com, heimdall.mydomain.com, loki.mydomain.com], salt-master: [balder.mydomain.com],
[...]

PR #52234, opened 2019-03-19
[2019.2] Fixes to matchers when used in pillar

garethgreenaway:
### What does this PR do?
With the change/addition of the matcher subsystem in 2019.2, the match.search_by when used in pillar broke when targeting the minion that was also the salt master.  This was caused by the id in __opts__ being used in all cases.  This change updates the glob_match function to use the preserved minion_id of the master if it is available so that targeting works as expected.

#51865 

### Previous Behavior
When used in pillar with the `salt.match.search_by` function, the result would was empty when targeting a Salt minion that was running on the Salt master. 

### New Behavior
In the event that the target is a Salt minion on the Salt master, match on the preserved `minion_id` value instead of the `id`.

### Tests written?
No.  This scenario only appears when there is a Salt minion running with the same ID as the salt master, which we currently do not have in place in the integration tests.

### Commits signed with GPG?
Yes


salt/matchers/compound_match.py
```diff
@@ -25,11 +25,15 @@ def match(tgt):
     '''
     nodegroups = __opts__.get('nodegroups', {})
     matchers = salt.loader.matchers(__opts__)
+    if 'minion_id' in __opts__:
+        minion_id = __opts__['minion_id']
+    else:
+        minion_id = __opts__['id']
```
waynew: Is LBYL preferred or would

```
try:
    minion_id = __opts__['minion_id']
except KeyError:
    minion_id = __opts__['id']
```

be a better approach? Not that it would probably matter *here* but I'm pretty sure the EAFP version is slightly faster. As another alternative, I think `minion_id = __opts__.get('minion_id', __opts__['id'])` would also work, but not sure if thats funkier lookin' :man_shrugging:

salt/matchers/glob_match.py
```diff
@@ -12,7 +12,11 @@ def match(tgt):
     '''
     Returns true if the passed glob matches the id
     '''
+    if 'minion_id' in __opts__:
+        minion_id = __opts__['minion_id']
+    else:
+        minion_id = __opts__['id']
```
waynew: Same question here?

waynew:
I don't have strong feelings about those changes - if you like it the way it is :+1: from me. Otherwise, :+1: from me after the changes :grinning:

salt/matchers/compound_match.py
garethgreenaway: I like the __opts__.get approach better, cleaner.  Updated the PR.

waynew (approved):

dwoz (approved):

Issue #52102, opened 2019-03-08
integration.runners.test_fileserver.FileserverTest.test_dir_list


dwoz:
```
integration.runners.test_fileserver.FileserverTest.test_dir_list
--------------------------------------------------------------------------------
develop failed salt-centos-7-py2-m2crypto, salt-centos-7-py2-pycryptodomex, salt-centos-7-py3, salt-centos-7-py3-m2crypto, salt-centos-7-py3-pycryptodomex, salt-centos-7-py3-transport, salt-debian-8-py2, salt-debian-8-py3, salt-debian-9-py2, salt-debian-9-py3, salt-fedora-28-py2, salt-fedora-28-py3, salt-fedora-29-py2, salt-fedora-29-py3, salt-opensuse-42-py2, salt-ubuntu-1404-py2, salt-ubuntu-1604-py2, salt-ubuntu-1604-py2-m2crypto, salt-ubuntu-1604-py2-pycryptodomex, salt-ubuntu-1604-py2-tornado, salt-ubuntu-1604-py3, salt-ubuntu-1604-py3-m2crypto, salt-ubuntu-1604-py3-pycryptodomex, salt-ubuntu-1804-py2, salt-windows-2019-py2
--------------------------------------------------------------------------------
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
{u'retcode': 70} is not an instance of <type 'list'>
................................................................................

Traceback (most recent call last):
  File "c:\users\admini~1\appdata\local\temp\kitchen\testing\tests\integration\runners\test_fileserver.py", line 26, in test_dir_list
    self.assertIsInstance(ret['return'], list)
AssertionError: {u'retcode': 70} is not an instance of <type 'list'>
```

waynew:
Still failing on develop:

[...]

Issue #52092, opened 2019-03-08
integration.runners.test_jobs.ManageTest.test_active


dwoz:
```
integration.runners.test_jobs.ManageTest.test_active
--------------------------------------------------------------------------------
develop failed salt-centos-7-py2-m2crypto, salt-centos-7-py2-pycryptodomex, salt-centos-7-py3, salt-centos-7-py3-m2crypto, salt-centos-7-py3-pycryptodomex, salt-centos-7-py3-transport, salt-debian-8-py2, salt-debian-8-py3, salt-debian-9-py2, salt-debian-9-py3, salt-fedora-28-py2, salt-fedora-28-py3, salt-fedora-29-py2, salt-fedora-29-py3, salt-opensuse-42-py2, salt-ubuntu-1404-py2, salt-ubuntu-1604-py2, salt-ubuntu-1604-py2-m2crypto, salt-ubuntu-1604-py2-pycryptodomex, salt-ubuntu-1604-py2-tornado, salt-ubuntu-1604-py3, salt-ubuntu-1604-py3-m2crypto, salt-ubuntu-1604-py3-pycryptodomex, salt-ubuntu-1804-py2, salt-windows-2019-py2
--------------------------------------------------------------------------------
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
{u'retcode': 70} != {} - {u'retcode': 70} + {}
................................................................................

Traceback (most recent call last):
  File "c:\users\admini~1\appdata\local\temp\kitchen\testing\tests\integration\runners\test_jobs.py", line 22, in test_active
    self.assertEqual(ret['return'], {})
AssertionError: {u'retcode': 70} != {}
- {u'retcode': 70}
+ {}
```

Ch3LL:
im still seeing this fail here: https://jenkinsci.saltstack.com/job/develop/job/salt-debian-9-py2/163/

twangboy:
Failing on develop:
May 21, 2019 5:42:00 PM - https://jenkinsci.saltstack.com/job/develop/job/salt-windows-2019-py2/95/
May 21, 2019 9:43:00 PM - https://jenkinsci.saltstack.com/job/develop/job/salt-windows-2016-py2/49/

Issue #52088, opened 2019-03-08
integration.modules.test_vault.VaultTestCase.test_write_raw_read_secret


dwoz:
```
integration.modules.test_vault.VaultTestCase.test_write_raw_read_secret
--------------------------------------------------------------------------------
develop failed salt-centos-7-py2-m2crypto, salt-centos-7-py2-pycryptodomex, salt-centos-7-py3, salt-centos-7-py3-m2crypto, salt-centos-7-py3-pycryptodomex, salt-centos-7-py3-transport, salt-ubuntu-1604-py2, salt-ubuntu-1604-py2-m2crypto, salt-ubuntu-1604-py2-pycryptodomex, salt-ubuntu-1604-py2-tornado, salt-ubuntu-1604-py3, salt-ubuntu-1604-py3-m2crypto, salt-ubuntu-1604-py3-pycryptodomex, salt-ubuntu-1804-py2
--------------------------------------------------------------------------------
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

................................................................................

Traceback (most recent call last):
  File "/tmp/kitchen/testing/tests/support/helpers.py", line 214, in wrap
    raise exc
AssertionError
```

Issue #52056, opened 2019-03-08
integration.runners.test_fileserver.FileserverTest.test_file_list


dwoz:
```
integration.runners.test_fileserver.FileserverTest.test_file_list
--------------------------------------------------------------------------------
develop failed salt-centos-7-py2-m2crypto, salt-centos-7-py2-pycryptodomex, salt-centos-7-py3, salt-centos-7-py3-m2crypto, salt-centos-7-py3-pycryptodomex, salt-centos-7-py3-transport, salt-debian-8-py2, salt-debian-8-py3, salt-debian-9-py2, salt-debian-9-py3, salt-fedora-28-py2, salt-fedora-28-py3, salt-fedora-29-py2, salt-fedora-29-py3, salt-opensuse-42-py2, salt-ubuntu-1404-py2, salt-ubuntu-1604-py2, salt-ubuntu-1604-py2-m2crypto, salt-ubuntu-1604-py2-pycryptodomex, salt-ubuntu-1604-py2-tornado, salt-ubuntu-1604-py3, salt-ubuntu-1604-py3-m2crypto, salt-ubuntu-1604-py3-pycryptodomex, salt-ubuntu-1804-py2, salt-windows-2019-py2
--------------------------------------------------------------------------------
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
{u'retcode': 70} is not an instance of <type 'list'>
................................................................................

Traceback (most recent call last):
  File "c:\users\admini~1\appdata\local\temp\kitchen\testing\tests\integration\runners\test_fileserver.py", line 151, in test_file_list
    self.assertIsInstance(ret['return'], list)
AssertionError: {u'retcode': 70} is not an instance of <type 'list'>

```

PR #52258, opened 2019-03-20
salt.cmd: be cautious with lazyloader

mattp-:
### What does this PR do?
in certain contexts we don't actually need to fire up grains (ie,
saltutil.runner from a pillar render). instead of firing it every time
we first check its if its in opts already, and if not, try and fetch it
from cache before falling back to instantiating a loader.

calling grains from pillar could clobber later calls to grains.get in a pillar render
Various runner tests that are failing:
#52102
#52052
#52056
#52088
#52092

### Tests written?
No

### Commits signed with GPG?

Yes/No


mattp-:
seems like there is an unrelated bug in salt.runner that is failing the tests, I put what looks like an appropriate fix in to get a clean run.

salt/runner.py
```diff
@@ -301,9 +301,10 @@ def _run_runner(self):
             if not self.opts.get('quiet', False):
                 display_output(ret, 'nested', self.opts)
         else:
-            ret = {
-                'retcode': salt.defaults.exitcodes.EX_SOFTWARE,
-            }
+            if not ret:
+                ret = {
+                    'retcode': salt.defaults.exitcodes.EX_SOFTWARE,
+                }
```
waynew: I'd have to look into this to understand the correct behavior - but if I understand things correctly this would skip the `EX_SOFTWARE` retcode if `ret` already exists. Is that intentional? Would it be possible to add some comment to this code explaining why that should be the case?

salt/runners/salt.py
```diff
@@ -85,7 +86,13 @@ def cmd(fun, *args, **kwargs):
     with_pillar = kwargs.pop('with_pillar', False)
 
     opts = copy.deepcopy(__opts__)
-    opts['grains'] = salt.loader.grains(opts)
+    # try to only load grains if we need to, it may already exist from other contexts (ie pillar)
+    if 'grains' not in opts:
+        id_, grains, _ = salt.utils.minions.get_minion_data(__opts__['id'], __opts__)
```
waynew: Is anything supposed to happen with `id_` here?

salt/runners/salt.py
```diff
@@ -85,7 +86,13 @@ def cmd(fun, *args, **kwargs):
     with_pillar = kwargs.pop('with_pillar', False)
 
     opts = copy.deepcopy(__opts__)
-    opts['grains'] = salt.loader.grains(opts)
+    # try to only load grains if we need to, it may already exist from other contexts (ie pillar)
```
waynew: Are pillars the only place where the grains could come from? If there are other places, [e.g.](https://www.grammarly.com/blog/know-your-latin-i-e-vs-e-g/) would be more appropriate.

waynew (changes requested):
Thanks for the PR!

I'm not sure that I understand the full impact of this change, but there are a couple of easy minor improvements that I think would be helpful.

salt/runner.py
mattp-: if ret isn't an empty dict. I think this is a bug introduced at some point in the recent past, as with try, except, else;  the else is always fired. as such it was discarding the ret. i suspect noone noticed because runnerclient is strictly for cli use (runner is for netapi), as such run_plus_plus in the test suite is the only thing actually invoking it and relying on the return in this way.

salt/runners/salt.py
mattp-: nope. i can null that

salt/runners/salt.py
mattp-: any jinja render, though pillar is the only one i can think of that provides a stubbed grain context. i'll update the comment though

salt/runner.py
mattp-: fwiw this change is unrelated to PR, I just made it to fix the failing tests

salt/runner.py
```diff
@@ -301,9 +301,10 @@ def _run_runner(self):
             if not self.opts.get('quiet', False):
                 display_output(ret, 'nested', self.opts)
         else:
-            ret = {
-                'retcode': salt.defaults.exitcodes.EX_SOFTWARE,
-            }
+            if not ret:
```
waynew: > fwiw this change is unrelated to PR, I just made it to fix the failing tests

Ah. Well, took a look at the code and maybe this will be a good improvement (and 🤞 help future edits not introduce bugs :trollface:)

```suggestion
            # If we don't have any values in ret by now, that's a problem.
            # Otherwise, we shouldn't be overwriting the retcode.
            if not ret:
```

waynew (changes requested):

salt/runner.py
mattp-: added comment, thanks!

salt/client/mixins.py
```diff
@@ -382,6 +382,7 @@ def low(self, fun, low, print_event=True, full_return=False):
             data['pid'] = os.getpid()
             with salt.utils.files.fopen(jid_proc_file, 'w+b') as fp_:
                 fp_.write(serial.dumps(data))
+            del data['pid']
```
cachedout: I'm not quite following the relation between this and the rest of the PR. Help me out? :)

mattp-: @cachedout just addressing another incidental test failure. I suspect maybe the jenkins test matcher didn't include the runner tests on the original checkin of the proc handling stuff from Austin.

cachedout: Whew. I was afraid there was some crazy logic linking the grains change to this and I was starting to panic. ;)

waynew (approved):

PR #52259, opened 2019-03-20
Fix win_proc threads leaked on linux

andzn:
### What does this PR do?
Minion win_proc threads were leaked on linux. This happened because processes were added to win_proc on both Linux and Windows, but they were cleaned up only on Windows (see https://github.com/saltstack/salt/blob/develop/salt/minion.py#L2689). Since the win_proc collection was supposed to keep windows threads, the fix was to not add processes to it unless the minion runs on windows. 

### Previous Behavior
linux threads added to win_proc were leaked

### New Behavior
linuxs threads are not added to win_proc anymore

### Tests written?
No

### Commits signed with GPG?
No

### DIFF
--- a/salt/minion.py
+++ b/salt/minion.py
@@ -1575,7 +1575,7 @@ class Minion(MinionBase):
         if multiprocessing_enabled and not salt.utils.platform.is_windows():
             # we only want to join() immediately if we are daemonizing a process
             process.join()
-        else:
+        elif salt.utils.platform.is_windows():
             self.win_proc.append(process)
 
     def ctx(self):


rares-pop (approved):
Nice finding! 
@twangboy , @dwoz - this is fixing part of the leaks we were seeing on our platforms. This should be backported too.

waynew (approved):

twangboy (approved):

dwoz:
@andzn Are you able to write a regression test for this?

andzn:
@dwoz, it would be my first regression test written. Should this be a memory leak test? If so, are there any examples? Or should I just write a test to make sure the win_proc list is always empty on linux? - something like the handle_decoded_payload tests under tests/unit/test_minion.py?

cachedout:
IMHO, it would be a better idea for us to come up with a way just to detect leaks in an automated fashion versus trying to construct a test around this specific PR. I think that will be very difficult and not buy too much in the end.

waynew:
I think that writing a test ensuring the win_proc list is always empty on linux is a reasonable compromise. It sounds like it should be simple enough, and provides at least a simple safety check. Like you say, though, I think it would be great if we had a way to detect memory leaks... though I'm not aware of any valgrind-like Python tools that we could take advantage of.

rares-pop:
I concur with @cachedout here. It seems too defensive to create a test just for this. It might justify one in case this is a recurrent problem.

dwoz:
@cachedout I agree we should try and automate leak detection. I think we might be able to use something like this:

https://pythonhosted.org/Pympler/tutorials/muppy_tutorial.html

dwoz (approved):

cachedout:
IMHO muppy might be overkill here. I think one good strategy could be to just take samples of memory usage during a test run and then turn them into a line graph. Overlay the line graphs on top of each other and if the shape starts to change, boom, you've got a leak.

rares-pop:
@cachedout , @dwoz - As I mentioned here:
https://github.com/saltstack/salt/pull/52471

I used objgraph to find and fix the leaks:
https://mg.pov.lt/objgraph/objgraph.html

It would be nice to have that integrated in the testing infrastructure, having like a mixin fixture which checks that after any test was run, there are no leaks.

Issue #50650, opened 2018-11-27
Windows Salt Minion 2018.3.3 Memory Leak 


binarygraka:
### Description of Issue/Question

I have a bunch of Windows Server VMs (running on VMware vSphere) in my infrastructure and they are all running the same version of Salt Minion. On two of them, the `python.exe` process needs almost all memory that is available on the system.

#### Windows Server 1

* Memory: 6.00 GB

When checking the memory consumption, the process `python.exe` was consuming almost all memory. I then stopped the `salt-minion` process, which took surprisingly long (about 30 seconds), the process `python.exe` was stopped and the 4.5 GB of RAM were free again. 

I then started the salt-minion process again, and till now (5 minutes later) the memory consumption is 63 MB, which seems normal. To me this seems like a memory leak.

#### Windows Server 2

Same behaviour. Regarding this server, I have better statistics. 

The process `python.exe` was consuming the following:

* Commit (KB): 2.883.772
* Working Set (KB): 1.203.412
* Shareable (KB): 2.600
* Private (KB): 1.200.856

Total:

* Available: 95 MB
* Cached: 84 MB
* Total: 4095 MB
* Installed: 4095 MB

After restarting the service:

`python.exe`:

* Commit (KB): 62.688
* Working Set (KB): 75.860
* Shareable (KB): 16.648
* Private (KB): 59.212

Total:

* Available: 1677 MB
* Cached: 207 MB
* Total: 4095 MB
* Installed: 4095 MB

### Setup

My infrastructure is:

* Salt master of masters
* Salt syndic, that is connected to the salt master of masters
* Salt minion, that is connected to the Salt syndic and is running on the mentioned Windows Server

#### Windows Server 1

The following log is available on the Salt minion. The first two lines must be from stopping and/or starting the Salt minion service:

```
2018-11-27 12:48:09,415 [salt.modules.cmdmod:673 ][ERROR   ][7868] Failed to decode stderr from command C:\Windows\System32\Wbem\wmic.EXE /namespace:\\root\microsoft\windows\storage path MSFT_PhysicalDisk where (MediaType=3 or MediaType=4) get DeviceID,MediaType /format:table, non-decodable characters have been replaced
[...]

PR #52260, opened 2019-03-20
Fix leak of SaltMessageClient instances when using tcp transport

andzn:
### What does this PR do?
SaltMessageClient instances are currently leaked on all OS-es (when tcp transport is used). A SaltMessageClient instance is created everytime a new TCP AsyncReqChannel is created which is happening every time a new job is posted (if transport is set to tcp). 
The SaltMessageClient is used to write the request on the stream and then to read the result back from it. The send method creates a future saves it to send_future_map and then returns it. The _stream_return callback then reads the result and obtains the future from the send_future_map and sets the result on it. 
When the result is set the send operation finishes, and because the channel is wrapped into a SyncWrapper that nobody references anymore, the garbage collector calls __del__ on the SyncWrapper which in turn closes the io_loop. The close method on SaltMessageClient is called, but as already pointed out in [this](https://github.com/saltstack/salt/blob/develop/salt/transport/tcp.py#L950) comment, by the time SaltMessageClient.close is called, the io_loop is already closed. This is causing problems because the _stream_return coroutine didn't finish, so tornado will continue to keep a reference on it. since the coroutine has a reference to self, the coroutine also keeps the SaltMessageClient alive. 
This PR gives a (new) chance to the _stream_coroutine to finish and therefore allows tornado to release the reference on it, and hence the garbage collector to collect the whole chain.
Notice that I also added an Exception handler because during the minion startup SaltMessageClient.close is called two times while the io_loop is still running. Therefore calling start again on it throws an exception. Unfortunately there's no public method that can be used to check whether an io_loop is running or not.
Notice that I also had to null out the self.io_loop reference, and also the _read_until_future. If the loop is started again, when _stream_return runs, it throws because the stream is closed, and puts an exception on the _read_until_future that contains a reference to the SaltMessageClient creating a loop, and preventing the garbage collector from collecting the SaltMessageClient object.
Also, because I nulled out the io_loop, the lambda passed in for the _stream_return_future needs to check whether io_loop is None or not before trying to stop it. Therefore I created a helper that does that. It's not clear to me whether why the [add_future call](https://github.com/saltstack/salt/blob/develop/salt/transport/tcp.py#L969) is needed. I tried to remove it but that caused the salt-minion to not work properly anymore.
Although this fixes the issue, a better fix would probably be to synchronize the _stream_return and the io_loop close so that the _stream_return finishes before the io_loop is closed. However with my current understanding I'm not sure how that can be done now - probably would need a refactoring, but i might be wrong.
 
Notice 
#49949
#50650

### Previous Behavior
All SaltMessageClient instances were leaked. Since a SaltMessageClient is created for each new job, the leaked memory can easily grow when there are many jobs created.

### New Behavior
SaltMessageClients instanced are garbage collected.

### Tests written?
No. I've used objgraph in my testing - a library that analyzes the information in the gc collections to give more details about how many instances are at some point of each type. Using that utility I was able to see that the number of the SaltMessageClient instances growed with each call without the fix, and doesn't grow anymore with the fix. I have checked this behavior on Windows 32-bit, Windows 64-bit and on LinuxRT.

### Commits signed with GPG?
No


rares-pop (approved):
I hoped this code goes away when moving to tornado 5.0, but doesn't seems like it. Maybe that's a chance to refactor those things properly.

Without this fix any Windows or 'tcp' minion is pretty much unusable in time. It will have to be backported to release branches.

@twangboy , @dwoz

DmitryKuzmenko:
I like this code as a first glance, but I want to take a deeper look tomorrow since it's night to me and I want to have an adequate look.

andzn:
I've just realized that there might be an easier and (in my opinion) also cleaner fix. 

The ideea would be to not spawn the _stream_return callback from _init_ but from the send and on_recv methods and also pass a stop_condition to it. 
If we spawn the _stream_return callback from send, the stop condition would be: 'send_future_map is empty'.
If we spawn the _stream_return callback from on_recv, the stop condition would be: '_on_recv is None'.
This approach should also allow us to remove the code from the close method that restarts the io_loop since that shouldn't be necessary anymore.
I did a quick experiment and it seems to work and fix the memory leak, but I still see a few exception thrown that I need to look into.

What do you think about this approach?

twangboy:
@andzn I like it. @dwoz? @DmitryKuzmenko?

andzn:
After some more testing that I did yesterday, I noticed that the second approach leaks SaltMessageClients after some time. The problem with the second approach is that there's only one _stream object per SaltMessageClient and I spawned the _stream_return callback two times. If there are two _stream_return callbacks running at the same time, we can't make one _stream_return callback read only responses for send, and the other one read only messages that were initiated by the other end. Both of the callbacks will read whatever comes first on the _stream. Even worse it could happen that one _stream_return callback reads all the messages and the other one doesn't manage to read anything and therefore will hang waiting to read something. And there's also the case where nothing comes back on a _stream and the operation times out. Therefore we still need the close code to stop any hanging _stream_return callbacks, which makes the second approach look not that appealing anymore.

I have set up a ubuntu machine with the salt testing suite, and I'm going to look into why one of the tests crashed on this PR build.

andzn:
I've probably got confused by the stack traces in the output log of the PR builds. I have ran the tests locally on a machine, and they all pass, and the exceptions that are thrown by some of the tests are thrown without my changes too. I'll update the branch so that tests are triggered again.

andzn:
It seems that the builds have passed on linux now, although they seem to have failed on windows. In the first build it was the other way (passed on windows and failed on linux). Since I didn't make any changes to the PR I assume that's just an intermittency in the build infrastructure.
@DmitryKuzmenko did you manage to look deeper over the PR? @twangboy? @dwoz?

DmitryKuzmenko (approved):
Looks good to me.

DmitryKuzmenko:
@andzn sorry for delay. Your changes looks good and reasonable from my viewpoint. Thank you for work.

cmcmarrow:
Looks good to me too.

cmcmarrow (approved):

dwoz:
I've backported this to 2019.2.1 and added a test for the close method. #53542

dwoz:
These changes are already in the new `master` branch via a merge forward from `2019.2.1`

PR #52264, opened 2019-03-20
Use windows state to bootstrap windows builds

dwoz:
This is also removing the use of dev_*.ps1 scripts in the windows builds

### Tests written?

No

### Commits signed with GPG?

Yes

### DIFF
--- a/.kitchen.yml
+++ b/.kitchen.yml
@@ -54,7 +54,7 @@ provisioner:
     base:
       "os:Windows":
         - match: grain
-        - prep_windows
+        - windows
       "*":
         - <%= ENV['KITCHEN_STATE'] || 'git.salt' %>
   pillars:


s0undt3ch (approved):

cachedout:
This appears to possibly break tests:
```
01:14:45        ----------
01:14:45            No matching sls found for 'windows' in env 'base'

s0undt3ch:
It's missing https://github.com/saltstack/salt/pull/52291

cachedout:
@s0undt3ch Should we bring back the `Blocked` label for situations such as this one?

s0undt3ch:
Are those more _tailored_ to issues?
I'm not against though.

cachedout:
@s0undt3ch Seems to me it could apply to both.

s0undt3ch:
Yep, that why I wasn't against :)
Anyway, both linked PRs are blocking each other...
Lets see if the other passes all tests(with the expected exception on Windows) and I'll merge it, then this one will be unblocked.

s0undt3ch:
Lets do it the other way around. Merging.

PR #52274, opened 2019-03-21
Add git_pillar interval timer to control how often git remotes are checked

Tanoti:
### What does this PR do?
Add support for new config option ```git_pillar_interval``` so git_pillar data can be collected less frequently than the main ```loop_interval``` timer allows

N/A

### Previous Behavior
Git pillar remotes would be checked using the ```loop_interval``` timer value (e.g. every 60 seconds).

### New Behavior
Git pillar remotes can be checked using a different interval to the ```loop_interval```. For example, if the git pillar data is not likely to change very often then a higher value such as 300 seconds could be used which would reduce the load on the git repo server by a factor of five.

### Tests written?
No

### Commits signed with GPG?
No

### DIFF
--- a/salt/master.py
+++ b/salt/master.py
@@ -226,7 +226,9 @@ class Maintenance(salt.utils.process.SignalHandlingMultiprocessingProcess):
 
         # Make Start Times
         last = int(time.time())
+        last_git_pillar = last
 
+        git_pillar_interval = self.opts.get('git_pillar_interval', 0)
         old_present = set()
         while True:
             now = int(time.time())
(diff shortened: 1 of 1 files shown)


twangboy (approved):

dwoz:
@Tanoti  can you write some basic tests for this?

Tanoti:
@dwoz Please point me to some similar tests and I'll take a look. Not sure how far the tests have to go (i.e. just checking the option is parsed or that it actually works)?

salt/master.py
```diff
@@ -226,7 +226,9 @@ def run(self):
 
         # Make Start Times
         last = int(time.time())
+        last_git_pillar = last
 
+        git_pillar_interval = self.opts.get('git_pillar_interval', 0)
```
cachedout: Is this intended to be a user-configurable option? If so, it needs documentation. :)

sathieu:
@Tanoti Also the config should be named `git_pillar_update_interval` to match [gitfs_update_interval](https://docs.saltstack.com/en/latest/ref/configuration/master.html#gitfs-update-interval).

sathieu:
Using another loop would be better (like gitfs does), but this is probably harder. See https://github.com/saltstack/salt/commit/958c99b87cf0c3af34e9f8353fc75ba08d1fcd63 and https://github.com/saltstack/salt/blob/a8fc7be7bab5f2d924086696b4960b621aeb6f7e/salt/master.py#L357

sathieu:
Superseeded by #53621.

waynew:
@sathieu so should we go ahead and close this PR then?

sathieu:
@waynew Yes I think. Unless @Tanoti wants to keep working on it.

Tanoti:
@sathieu I'm happy for this to be closed. I've not had time to look at it any further so thanks for moving it on.

PR #51750, opened 2019-02-21
Replaced a dead link for github issue tracker with another github

alan-cugler:
### What does this PR do?
Replaced a dead link for github issue tracker with another github link for the issue tracker and added a link to show more options for closing issues in commit messages.

### What issues does this PR fix or reference?
Didn't bother to make an issue for the deadlink

### Commits signed with GPG?
Yes

PR #52276, opened 2019-03-21
Replace GitHub dead link

alan-cugler:
### What does this PR do?
Adjusted  a URL alias in doc page to properly redirect to the salt issues page on github.
The original URL link is valid but the rendered alias redirects in error.
Just stripped some characters from the alias name and it worked properly!

### Commits signed with GPG?
Yes


alan-cugler:
#51750 is the same PR, but I was instructed to make a new one.

rallytime (approved):

cachedout (approved):
Looked at the rendered markdown and it appears correct.

PR #52279, opened 2019-03-21
[develop] Update to virtualenv module removing pyvenv if Python greater than 3.6

garethgreenaway:
### What does this PR do?
Beginning in Python 3.6, pyvenv has been deprecated in favor of using venv, so remove it from the list if we are using 3.6 or higher.

N/A

### Tests written?
No.  Tests exist.

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/modules/virtualenv_mod.py
+++ b/salt/modules/virtualenv_mod.py
@@ -22,6 +22,7 @@ from salt.exceptions import CommandExecutionError, SaltInvocationError
 from salt.ext.six import string_types
 
 KNOWN_BINARY_NAMES = frozenset([
+    'pyvenv-{0}.{1}'.format(*sys.version_info[:2]),
     'virtualenv-{0}.{1}'.format(*sys.version_info[:2]),
     'pyvenv{0}'.format(sys.version_info[0]),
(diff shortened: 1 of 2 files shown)


twangboy (approved):

gtmanfred:
Can you clarify this, is pyenv deprecated or pyvenv deprecated? Because they are two separate things.

salt/modules/virtualenv_mod.py
```diff
@@ -139,9 +139,17 @@ def create(path,
         salt '*' virtualenv.create /path/to/new/virtualenv
     '''
     if venv_bin is None:
-        venv_bin = __pillar__.get('venv_bin') or __opts__.get('venv_bin')
-
-    cmd = [venv_bin]
+        # Beginning in 3.6, pyenv has been deprecated
```
gtmanfred: this should be `pyvenv`

gtmanfred (commented):
`pyenv` is a tool for installing multiple full versions of python, and is not part of the stdlib

`pyvenv` is a binary that used to be provided by python as a replacement for virtualenv, and in python3.6 was deprecated in favor of `python -m venv`

gtmanfred (approved):
Thanks

PR #52286, opened 2019-03-22
Handle Service dependencies in Windows

ghost:
### What does this PR?

When attempting to start/stop/restart a Windos which has dependencies configured which are not in the needed state, an exception is currently raised. E.g.

```
pywintypes.error: (1051, 'ControlService', 'A stop control has been sent to a service that other running services are dependent on.')

2019-03-20 08:36:09,168 [salt.state       :321 ][ERROR   ][6744] An exception occurred in this state: Traceback (most recent call last):
  File "c:\salt\bin\lib\site-packages\salt\modules\win_service.py", line 518, in stop
    win32serviceutil.StopService(name)
  File "c:\salt\bin\lib\site-packages\win32\lib\win32serviceutil.py", line 409, in StopService
    return ControlService(serviceName, win32service.SERVICE_CONTROL_STOP, machine)
  File "c:\salt\bin\lib\site-packages\win32\lib\win32serviceutil.py", line 320, in ControlService
    status = win32service.ControlService(hs, code)
pywintypes.error: (1051, 'ControlService', 'A stop control has been sent to a service that other running services are dependent on.')

  File "c:\salt\bin\lib\site-packages\salt\state.py", line 1919, in call
    **cdata['kwargs'])
  File "c:\salt\bin\lib\site-packages\salt\loader.py", line 1918, in wrapper
    return f(*args, **kwargs)
  File "c:\salt\bin\lib\site-packages\salt\states\service.py", line 607, in dead
    func_ret = __salt__['service.stop'](name, **stop_kwargs)
  File "c:\salt\bin\lib\site-packages\salt\modules\win_service.py", line 522, in stop
    'Failed To Stop {0}: {1}'.format(name, exc.strerror))
salt.exceptions.CommandExecutionError: Failed To Stop SomeService: A stop control has been sent to a service that other running services are dependent on.
```

This PR attempts to mitigate the situation by adding logic which resolves the dependencies and considers them during the start/stop/restart operations. 

For this purpose a class was written which provides the related services in most convenient order. 

This new feature is currently optional to maintain old behavior. It can be enabled using the two new kwargs `with_deps=bool()` and `with_customers=bool()` which were added to the methods `modules/win_service.[start|stop|restart]`. In case multiple services are affected by the operation, the return value is `True` if all operations succeeded, otherwise it is `False`. 

Additionally the `states/service.py` has been modified the way it passes the parameters to the `win_service` module if configured. 

I failed to find a snappy classifying name for the group of services which depend on the given service. That's not the dependencies of the service but the opposite direction of the dependency hierarchy. I used the term `customers` for this, maybe there is something better available. 

### Previous Behavior

When a service with a not running dependency is started, an exception is raised
When a service with a running dependency is stopped, an exception is raised

### New Behavior

When a service with a not running dependency is started, the dependency can be optionally automatically started as well
When a service with a running dependency is stopped, the dependency can be optionally automatically stopped as well

### Tests written?

Unit tests

### Commits signed with GPG?

Yes


twangboy (approved):

afischer-opentext-com:
I reworked the code now to reduce complexity of the code. Still codeclimate complains. Imho it is not reasonable to further attempt to reduce the complexity. Maybe you can give me a hint or we have to live with it. The Windows tests seem to fail for different reasons at the moment. They fail since the last merge of develop.

salt/modules/win_service.py
```diff
@@ -462,6 +608,15 @@ def start(name, timeout=90):
...
             .. versionadded:: 2017.7.9,2018.3.4
 
+        with_deps (bool):
+            If enabled start the given service and the services the current
+            service depends on.
+
+        with_customers (bool):
+            If enabled and in case other running services depend on the to be start
+            service, this flag indicates that those other services will be started
+            as well.
```
twangboy: Perhaps it could be named `with_parents`:
```
If other running services depend on the service being started, ``True`` will start the parent services as well. Default is ``False``
```

twangboy (approved):
Looks good to me.

salt/modules/win_service.py
afischer-opentext-com: I changed the wording from `customer` to `parent`. I agree that this may better describe the hierarchies we see here, however it does not describe the semantics properly as well. Parents do not depend on their children necessarily, its more the way around, but this dependency exists here.

twangboy: Yeah, it was just an option... I'm fine either way.

afischer-opentext-com:
@twangboy, what are the next steps now?

PR #52292, opened 2019-03-22
Allow byte data to be encoded to base64 if UnicodeDecodeError

Ch3LL:
### What does this PR do?
Currently the tests in this merge forward are failing: https://github.com/saltstack/salt/pull/52190

The reason they are failing is because when you run the two tests:

`python3 tests/runtests.py -n integration.states.test_file.FileTest.test_binary_contents -n integration.netapi.rest_cherrypy.test_app.TestJobs.test_all_jobs`

particularly when you run `test_binary_contents` before `test_all_jobs` the following gets added to the job return info: `b'GIF89a\x01\x00\x01\x00\x80\x00\x00\x05\x04\x04\x00\x00\x00,\x00\x00\x00\x00\x01\x00\x01\x00\x00\x02\x02D\x01\x00;'`
So when we go to run the `test_all_jobs` salt-api is not able to `json.dump` this because json does not work with bytes.

We also cant just decode it, because '\x80' is not a utf-8 unicode code point and results in a `UnicodeDecodeError`. We could possibly just pass `errors='replace'` when decoding, but then the user is no longer able to work with this data. This PR will encode bytes via base64 if its not able to decode it. This way the user will still be able to decode the base64 if they need the data, because a lot of times users that are using the salt-api are trying to consume this information programatically.

If reviewers are against this approach, I see the only other best scenario is just decoding with `errors='replace'`

Fixes failing tests in this PR: https://github.com/saltstack/salt/pull/52190

### Previous Behavior
failing tests

### New Behavior
The tests `integration.states.test_file.FileTest.test_binary_contents` and `integration.netapi.rest_cherrypy.test_app.TestJobs.test_all_jobs` both pass

### Tests written?

Yes

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/netapi/rest_cherrypy/app.py
+++ b/salt/netapi/rest_cherrypy/app.py
@@ -612,6 +612,7 @@ except ImportError:
 import salt
 import salt.auth
 import salt.exceptions
+import salt.utils.data
 import salt.utils.event
 import salt.utils.json
 import salt.utils.stringutils
@@ -898,6 +899,8 @@ def hypermedia_handler(*args, **kwargs):
     # Transform the output from the handler into the requested output format
     cherrypy.response.headers['Content-Type'] = best
     out = cherrypy.response.processors[best]
+    if isinstance(ret, dict):
+        ret = salt.utils.data.decode_dict(ret, encode_bytes=True)
     try:
         response = out(ret)
         if six.PY3:
--- a/salt/utils/data.py
+++ b/salt/utils/data.py
@@ -155,7 +155,7 @@ def compare_lists(old=None, new=None):
 
 def decode(data, encoding=None, errors='strict', keep=False,
            normalize=False, preserve_dict_class=False, preserve_tuples=False,
-           to_str=False):
+           to_str=False, encode_bytes=False):
     '''
     Generic function which will decode whichever type is passed, if necessary.
     Optionally use to_str=True to ensure strings are str types and not unicode
@@ -182,25 +182,30 @@ def decode(data, encoding=None, errors='strict', keep=False,
     two strings above, in which "й" is represented as two code points (i.e. one
     for the base character, and one for the breve mark). Normalizing allows for
     a more reliable test case.
+    :param encode_bytes: base64 encode bytes if it cannot be decoded
     '''
     _decode_func = salt.utils.stringutils.to_unicode \
         if not to_str \
         else salt.utils.stringutils.to_str
     if isinstance(data, Mapping):
         return decode_dict(data, encoding, errors, keep, normalize,
-                           preserve_dict_class, preserve_tuples, to_str)
+                           preserve_dict_class, preserve_tuples, to_str,
+                           encode_bytes=encode_bytes)
     elif isinstance(data, list):
         return decode_list(data, encoding, errors, keep, normalize,
-                           preserve_dict_class, preserve_tuples, to_str)
+                           preserve_dict_class, preserve_tuples, to_str,
+                           encode_bytes=encode_bytes)
     elif isinstance(data, tuple):
         return decode_tuple(data, encoding, errors, keep, normalize,
-                            preserve_dict_class, to_str) \
+                            preserve_dict_class, to_str, encode_bytes=encode_bytes) \
             if preserve_tuples \
             else decode_list(data, encoding, errors, keep, normalize,
-                             preserve_dict_class, preserve_tuples, to_str)
+                             preserve_dict_class, preserve_tuples, to_str,
+                             encode_bytes=encode_bytes)
     else:
         try:
-            data = _decode_func(data, encoding, errors, normalize)
+            data = _decode_func(data, encoding, errors, normalize,
+                                encode_bytes=encode_bytes)
         except TypeError:
             # to_unicode raises a TypeError when input is not a
             # string/bytestring/bytearray. This is expected and simply means we
@@ -214,10 +219,11 @@ def decode(data, encoding=None, errors='strict', keep=False,
 
 def decode_dict(data, encoding=None, errors='strict', keep=False,
                 normalize=False, preserve_dict_class=False,
-                preserve_tuples=False, to_str=False):
+                preserve_tuples=False, to_str=False, encode_bytes=False):
     '''
     Decode all string values to Unicode. Optionally use to_str=True to ensure
     strings are str types and not unicode on Python 2.
+    :param encode_bytes: base64 encode bytes if it cannot be decoded
     '''
     _decode_func = salt.utils.stringutils.to_unicode \
         if not to_str \
@@ -227,13 +233,15 @@ def decode_dict(data, encoding=None, errors='strict', keep=False,
     for key, value in six.iteritems(data):
         if isinstance(key, tuple):
             key = decode_tuple(key, encoding, errors, keep, normalize,
-                               preserve_dict_class, to_str) \
+                               preserve_dict_class, to_str, encode_bytes=encode_bytes) \
                 if preserve_tuples \
                 else decode_list(key, encoding, errors, keep, normalize,
-                                 preserve_dict_class, preserve_tuples, to_str)
+                                 preserve_dict_class, preserve_tuples, to_str,
+                                 encode_bytes=encode_bytes)
         else:
             try:
-                key = _decode_func(key, encoding, errors, normalize)
+                key = _decode_func(key, encoding, errors, normalize,
+                                   encode_bytes=encode_bytes)
             except TypeError:
                 # to_unicode raises a TypeError when input is not a
                 # string/bytestring/bytearray. This is expected and simply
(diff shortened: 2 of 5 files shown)


twangboy (approved):

max-arnold:
Trying to guess the encoding is rarely a good solution.

A binary content may or may not produce the UnicodeDecodeError, so the data may or may not be base64 encoded. This will require guessing on the receiving side because there is no reliable way to tell base64 encoded data from just a string (and the managed file also could be base64 encoded!).

Possible ideas:

1. Add an out-of-band explicit encoding marker to this particular response or field (to indicate that it is necessary to decode the result using base64)
2. Do not send binary data at all (and indicate this in some way, preferably out-of-band)
3. Append a sufficiently unique in-band marker to each base64 encoded field
4. Use a binary-safe protocol

mattp-:
I don't think this makes sense, not all data can be represented in json. returning a 406 if an alternative content-type provided via Accept header is generally what a web service would do, which looking it seems the cherrypy implementation does that. the test should either be changed to use yaml instead, or the job cache should be cleared before running that test to ensure only json compatible jobs are in the cache for the test.

waynew:
> A binary content may or may not produce the UnicodeDecodeError

What cases are you envisioning?

Also, while that may be true in the *general* case, currently Salt (that I'm aware of) doesn't really have a great story for supporting binary data. This is pretty much due to Python2 conflating binary and text data.

> Trying to guess the encoding is rarely a good solution.

That's very true. It's also what Salt does today, uh, everywhere :trollface:

Ultimately, we *should* fix this, and I *do* plan on creating a [Salt Enhancement Proposal](https://github.com/saltstack/salt-enhancement-proposals) but the actual fix is going to be pretty intense, especially if it happens before we drop Python2 support, because basically everywhere in the code base we're expecting Unicode, instead of bytes.

---

All of that to say, from an ideal perspective we should definitely be doing other things, but for now this should be totally fine and is an excellent improvement :slightly_smiling_face:

waynew (approved):

max-arnold:
>> A binary content may or may not produce the UnicodeDecodeError

> What cases are you envisioning?

I'm talking specifically about these lines:

```
            try:
                return _normalize(to_str(s, encoding, errors))
            except UnicodeDecodeError as err:
```

A type of encoding would be dependent on the content and is unpredictable and hard to guess if you want to reliably decode it.

salt/netapi/rest_cherrypy/app.py
```diff
@@ -898,6 +899,8 @@ def hypermedia_handler(*args, **kwargs):
     # Transform the output from the handler into the requested output format
     cherrypy.response.headers['Content-Type'] = best
     out = cherrypy.response.processors[best]
+    if isinstance(ret, dict):
```
waynew:
```suggestion
if best == 'application/json' and isinstance(ret, dict):
```
There may be a better way to write that, but the `ct_out_map` could be `application/x-yaml` which should support binary data, and we shouldn't be encoding the bytes there.

waynew (changes requested):

gtmanfred:
@whiteinge can you help here?

Thanks.
Daniel

gtmanfred:
Adding some discussion from slack
```
Wayne Werner [10:21 AM]
What would you expect to see?


Matt [10:24 AM]
wayne:
```406 Not Acceptable The resource identified by the request is only capable of generating response entities which have content characteristics not acceptable according to the accept headers sent in the request.```

gtmanfred [10:24 AM]
probably a 406 error, and a better handling of the situation as a whole in develop, instead of shoehorning it into 2019.2 (edited) 

Matt [10:28 AM]
@Wayne thats basically http way of saying ‘i cant serve you what you asked for with the content-type you said you can talk with via Accept: header’

gtmanfred [10:28 AM]
if you specify the accept header as yaml, does it return the data correctly?

gtmanfred [10:35 AM]
Looks like a better solution would be to include a multi-part form where you can actually include attachments in the return https://www.ietf.org/rfc/rfc1867.txt

Matt [10:35 AM]
im not actually sure. i doubt many people use salt with binary data like this

gtmanfred [10:36 AM]
it looks like yaml will just return your !!binary “…” whatver, and it just base64s it
so it is handled in the yaml spec
which would mean if people need binary data, they can specify yaml as the content type, and this could be documented, and then in a future release of salt, a multipart attachment based form could be used for handling this usecase

Matt [10:37 AM]
ya. im trying to jimmy a file.managed yaml with that right now

gtmanfred [10:38 AM]
deferring to documentation seems like a better solution that a jury-rigged solution where salt tries to base64 encode anything it cant convert to unicode.
```

whiteinge:
At first glance, I don't think this is specific to salt-api. The failing `test_binary_contents` test looks to be ensuring that running a `file.managed` that returns binary data in the output won't cause a failure and so I would expect that Salt's JSON outputter should have the same bug, but perhaps that outputter doesn't have tests that exercise this code path. (An aside: the JSON outputter is often a good metric to use to decide if something _should_ work in salt-api.)

I'm coming into this issue cold so it's possible I'm overlooking something or missing some history, but it looks to me like support for this should be added to `salt.utils.json` and not to salt-api.

gtmanfred:
Also, from what i remembered, if the file we were managing in file.managed was binary data, we did not show a diff of it, which should handle this error anyway?

https://github.com/saltstack/salt/blob/2018.3/salt/modules/file.py#L112-L131

mattp-:
@whiteinge you are spot on, --out=json I expect would blow up in exactly the same fashion. The only difference is I am arguing it's not a bug, if people want to use salt in a way that is encoding binary in low/ret thats fine, they just need to use encoding and output that can handle it. Doing some ad-hoc base64 encoding of the inner data for _all_ output types seems to be very unexpected behavior unless I asked for base64 data.

whiteinge:
> if people want to use salt in a way that is encoding binary in low/ret that's fine, they just need to use encoding and output that can handle it.

That use-case seems fair, though probably niche. And there may also be payload-size restrictions either in the response payload from minion to master or the job cache (and/or ext job cache and/or returners). I don't know if any of the questions around storing binary data in job returns have been answered already, but if Salt wants to facilitate that it should be defined and documented.

To facilitate that workflow throughout all of Salt I think the outputter modules would be the first line of integration (gracefully handling that format or gracefully not) so that the end-user can expect to be able to safely use the CLI, returners, job cache, salt-api, etc without worry of tracebacks in any of the usual contexts. After that it may make sense to add content-type negotiation to salt-api to deliver binary responses directly.

tl;dr: adding robust minion-to-master binary support to job returns should be defined and spec'ed out in Salt-core first, then salt-api next. And, IMO, only if there's compelling end-user need for doing so.

waynew:
Okay, now that I've really dug into and tried to grok the surrounding code I'm even more confused as to what's going on here (i.e. why these tests would be failing).

I've been working on a different problem that has the same fundamental (overeager decoding of binary data) cause. I'm not sure why the jobs data is returning data from a potentially unrelated test (though I'm guessing it's due to the fact that the integration tests spin up a master and minion, and I'm guessing this cherrpy appserver/api is hitting that minion) but even more I'm not sure why it appears that the `/jobs` would be returning binary data. At least on the regular outputter you don't get the actual file contents when binary files diff, just something like "binary file changed".

I will probably test to see if the json outputter normally barfs when you throw some binary in there.

But these tests actually *should* work as written if the client accepted YAML, though they might need to deserialize the YAML

mattp-:
it looks like output_json passes repr for default: https://github.com/saltstack/salt/blob/develop/salt/output/json_out.py#L80
which would stringify any bytes that it can't deal with. I'm not sure that makes sense to do for salt-api for roughly the same reason I mentioned above; it's serving the raw low data, not a rendered outputter.

Ch3LL:
Thanks for everyone's comments and suggestions here.  To clarify how this is occurring, in the test runner it is using `state.single` to run the `file.managed` state. It calls the `state.single` command against the LocalClients `cmd` method. So ` contents=b'GIF89a\x01\x00\x01\x00\x80\x00\x00\x05\x04\x04\x00\x00\x00,\x00\x00\x00\x00\x01\x00\x01\x00\x00\x02\x02D\x01\x00;'` is added to the state.single call and that is how it is getting added to the job cache. Its not being added by the actual `file.managed` state run, as everyone pointed out the file.managed state would not actually add the GIF data to the return of the state run. As everyone pointed out this is not a very common use case, and i doubt anyone is using this, so the question lies in do we want to actually support this?

I am completely fine taking the approach of defining later on if we want to actually support bytes data in the job cache. It seems this needs to be handled somewhere else outside of salt-api. It might be best to just clean up the test, by just cleaning the job cache, and if it becomes a requirement in the future that we figure out how to handle the byte data in a future SEP in the further context of all of salt, because as @waynew pointed out currently in salt it heavily relies on expecting the data to be unicode and not bytes throughout the code.

Is everyone okay with this approach?

PR #52328, opened 2019-03-26
fix net.load_template bug

wasabi222:
### What does this PR do?

This fixes a bug in the napalm_network module that prevents users from only specifying an inline template with the `template_source` kwarg. 

The `template_name` variable defaults to `None` in this function, so when a user calls `net.load_template` without the `template_name` var defined (for example, if they want to specify an inline template with the `template_source` kwarg), this errors out due to the following:

```
if not isinstance(template_name, (tuple, list)): # this will run since template_name is a NoneType
        for salt_render_prefix in salt_render_prefixes:
            if not salt_render:
                salt_render = salt_render or template_name.startswith(salt_render_prefix) # <-- the bug is here - NoneType has no attribute 'startswith'
```

This currently produces the following traceback:

```
      File "/usr/lib/python2.7/dist-packages/salt/minion.py", line 1607, in _thread_return
        return func(*args, **kwargs)
      File "/usr/lib/python2.7/dist-packages/salt/utils/napalm.py", line 430, in func_wrapper
        return func(*args, **kwargs)
      File "/var/cache/salt/proxy/extmods/junos/modules/napalm_network.py", line 1920, in load_template
        salt_render = salt_render or template_name.startswith(salt_render_prefix)
```

pinging @mirceaulinic as well!

### DIFF
--- a/salt/modules/napalm_network.py
+++ b/salt/modules/napalm_network.py
@@ -1914,7 +1914,7 @@ def load_template(template_name=None,
     salt_render_prefixes = ('salt://', 'http://', 'https://', 'ftp://')
     salt_render = False
     file_exists = False
-    if not isinstance(template_name, (tuple, list)):
+    if template_name and not isinstance(template_name, (tuple, list)):
         for salt_render_prefix in salt_render_prefixes:
             if not salt_render:
                 salt_render = salt_render or template_name.startswith(salt_render_prefix)


mirceaulinic (approved):
LGTM

wasabi222:
@Ch3LL can I get this slated for 2019.2.1?

Ch3LL:
we can try to get it into 2019.2.1 if you target that branch and add a regression test for the change

Ch3LL:
my previous comment is not correct. This PR will need to be added to the master branch with tests. Anything I can help with here to get tests written?

sagetherage:
@wasabi222 this needs to be rebased against the master branch, we are no longer accepting changes to develop. We have not heard from you in some time so this may be closed as abandoned.

Ch3LL (changes requested):
any chance of getting back to this one. needs to be on master branch and a test added

wasabi222:
oh sorry about that. My git game is not very strong so I'm not sure how to do this. my fork of salt doesn't have a master branch. Should I just delete this pr and submit a new one with a new fork?

wasabi222:
Also this has a test here: https://github.com/saltstack/salt/blob/master/tests/unit/modules/test_napalm_network.py#L105 - is there anything else that needs to be done on the testing side? Can I just run the tests locally and make sure they pass?

sagetherage:
@wasabi222 yes, if you can create a new PR that points to the `master` branch in Salt and include the test, we will review for the next release cycle :-)

Issue #42518, opened 2017-07-24
Running python scripts w/ `cmd.run` results in bizarre default encoding


eykd:
### Description of Issue/Question

When running a python script (v2.x or v3.x) via `cmd.run` state, I'm getting a bizarre default encoding:

```
$ sudo salt 'myserver' cmd.run "python3 -c 'import locale; print(locale.getpreferredencoding())'"
myserver:
    ANSI_X3.4-1968

$ sudo salt 'myserver' cmd.run "python2 -c 'import locale; print(locale.getpreferredencoding())'"
myserver:
    ANSI_X3.4-1968
```

However, running this w/ `salt-call` gives me the expected default of `UTF-8`:

```
$ sudo salt-call cmd.run "python3 -c 'import locale; print(locale.getpreferredencoding())'"
local:
    UTF-8
```

### Versions Report

```
$ salt --versions-report                                                                                                                                  [3/384]
Salt Version:
           Salt: 2016.11.5

Dependency Versions:
           cffi: 1.10.0
       cherrypy: Not Installed
       dateutil: 2.4.0
      docker-py: Not Installed
          gitdb: 0.6.4
      gitpython: 0.3.2 RC1
          ioflo: Not Installed
         Jinja2: 2.8
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: 0.21.1
           Mako: 1.0.3
   msgpack-pure: Not Installed
 msgpack-python: 0.4.6
   mysql-python: Not Installed
      pycparser: 2.17
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: Not Installed
         Python: 2.7.12 (default, Nov 19 2016, 06:48:10)
   python-gnupg: Not Installed
         PyYAML: 3.11
          PyZMQ: 15.2.0
           RAET: Not Installed
          smmap: 0.9.0
        timelib: Not Installed
        Tornado: 4.3
            ZMQ: 4.1.4

System Versions:
           dist: Ubuntu 16.04 xenial
        machine: x86_64
        release: 4.4.0-83-generic
         system: Linux
        version: Ubuntu 16.04 xenial
```

astronouth7303:
I just stumbled on this issue.

The default `cmd.run` (and friends) configuration looses the system encoding, setting it to `C`. (Hardcoded in the source.)

[...]

PR #52330, opened 2019-03-26
Correct reset locale in cmd.run

AstraLuma:
### What does this PR do?
Intelligently set the locale in `cmd` to preserve encoding information.

#42518

### Previous Behavior
Unconditionally set the locale to `C`, erasing system encoding information (including especially the filesystem).

### New Behavior
Preserves any set encoding while still setting the language to `C`.

### Tests written?

Could not find existing tests for cmd.run that this would go with.

### Commits signed with GPG?

Yes


astronouth7303:
Ok, I have no idea why windows broke.

I'll look into the linux tests, though.

s0undt3ch (dismissed):

s0undt3ch:
We should try and get tests for this. I know plain `C` works but of the system doesn't have, for example `C.UTF-8` what happens?

astronouth7303:
It should use whatever encoding is declared. It doesn't hardcode an encoding, just forwards it.

twangboy (dismissed):

twangboy:
@dmurphy18 Hey, is this related to the language code stuff you were working on?

dmurphy18 (changes requested):
This needs to be merged up to the master branch and have some tests associated with it, esp. given some Containers and platforms do not have C.UTF-8 for example: see PEP 538 and PEP540. Noting that this will be going into the Sodium release at a minimum at this point, where these PEP's have been back-ported to Python 3.6. Some environments only provide 'C' and POSIX, and would want to ensure that these changes work in those environments.

astronouth7303:
What should happen is that it'll preserve the encoding of the environment.

For example, if the system is `en_US`, it should change it to `C`. If it's `en_US.UTF-8`, it'll change it to `C.UTF-8`.

I think what happened at the time was that it was declared as a UTF-8 locale without having it installed?

astronouth7303:
But yeah, this PR needs a refresh.

sagetherage:
@AstraLuma do you want to rebase this against the `master` branch or close it and base a new PR on this one against `master`?

AstraLuma:
A wild necromancer appears!

I can attempt to rebase it, but I haven't really thought about this code since I wrote it.

AstraLuma:
> Some environments only provide 'C' and POSIX, and would want to ensure that these changes work in those environments.

This should work fine in environments without UTF-8 locales installed--it only preserves the encoding in the declared locale. As written, though, it'll cause a problem if your system is declared `en_US.UTF-8` but you don't have `C.UTF-8` installed. (Which, I would argue, is a pretty unfortunate circumstance that could lead to other problems if you actually have to use unicode >127.)

AstraLuma:
Oh, and to add insult to injury: the `locale` module doesn't provide a way to probe if a locale exists without trying to set it and seeing if it works. Given [the description](https://docs.python.org/3/library/locale.html#background-details-hints-tips-and-caveats) of locales, I would say that this method is a Bad Idea.

AstraLuma:
it does look like there's `cmd.run` tests now, so I'll have to remember to look into that. (Between my other stuff.)

AstraLuma:
it does look like there's `cmd.run` tests now, so I'll have to remember to look into that. (Between my other stuff.)

Ch3LL:
@AstraLuma are you willing to come back to this PR and follow up with @dmurphy18 's feedback about test coverage?

dmurphy18 (changes requested):
@AstraLuma This PR is marked pending-close since unresponsive, if you have lost interested in this PR please let us know, else can you give an estimate as to when you will address the failing tests ?

Ch3LL:
Closing due to inactivity. Please let me know if you want me to re-open or please submit a PR against the master branch with the requested changes.

Issue #49855, opened 2018-10-01
file.line not inserting new line as expected


doesitblend:
### Description of Issue/Question
Given a dynamic state `file.line` does not insert a new line as expected. Given the state provided in the setup section it is expected to get behavior that continually inserts a new line above the line `AAAAAAA`. Instead, the first attempt adds a new line, however subsequent executions of the state do not modify the file at all.

I tried using the `ensure` mode to see if that would be a valid workaround for this issue. This worked, but replaced the line with the new content instead of inserting the new line as expected.

Further testing revealed that if the beginning of the string changes, then the line gets inserted. But changing on the end of the line will not update the line. 

### Setup
Setup a 2018.3.2 master and minion.

Create a file called `/var/log/salt/test.log` with write permissions for the root user:

```
AAAAAAA
```

Add this state:

```
{% set curtime = None | strftime('%Y-%m-%d %H:%M:%S') %} 
update_missing_list: 
  file.line: 
    - name: /var/log/salt/test.log 
    - mode: insert
    - content: {{ curtime }}
    - before: AAAAAAA
```

#### First execution of state
```
f420deb9bac6:
----------
          ID: update_missing_list
    Function: file.line
        Name: /var/log/salt/test.log
      Result: True
     Comment: Changes were made
     Started: 19:20:15.094866
    Duration: 8.982 ms
     Changes:   
              ----------
              diff:
                  --- 
                  +++ 
                  @@ -1 +1,2 @@
                  +2018-10-01 19:20:15
                   AAAAAAA

Summary for f420deb9bac6
------------
Succeeded: 1 (changed=1)
Failed:    0
------------
Total states run:     1
Total run time:   8.982 ms
```

#### Second execution of state

```
f420deb9bac6:
----------
          ID: update_missing_list
    Function: file.line
        Name: /var/log/salt/test.log
      Result: True
     Comment: No changes needed to be made
     Started: 19:20:53.272621
    Duration: 7.865 ms
     Changes:   

Summary for f420deb9bac6
------------
Succeeded: 1
Failed:    0
------------
Total states run:     1
Total run time:   7.865 ms

```

### Versions Report
```
Salt Version:
           Salt: 2018.3.2
 
Dependency Versions:
           cffi: Not Installed
       cherrypy: Not Installed
       dateutil: Not Installed
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.7.2
[...]

PR #52333, opened 2019-03-26
compare probe/src directly issues/49855

jubrad:
### What does this PR do?
This changes the tokenized approach for string comparison
and shifts it to direct comparison of the content and the line in question
to determine if any changes should be made during file.line insert/ensure before or after

https://github.com/saltstack/salt/issues/49855

### Previous Behavior
Weird comparison of words that begin a line to determine if changes should be made

### New Behavior
Direct comparison on lines and content to be added.

### Tests written?
Manually tested by replacing site-packages files.
Adjusted unit tests to match new functionality.

### DIFF
--- a/salt/modules/file.py
+++ b/salt/modules/file.py
@@ -1729,11 +1729,11 @@ def _mkstemp_copy(path,
 
     return temp_file
 
+
 def _starts_till(src, probe):
     '''
     Returns True if the line matches the probe
     '''
-    
     no_match = -1
     equal = 0
 
@@ -1744,6 +1744,7 @@ def _starts_till(src, probe):
         return equal
     
     return no_match
+
 
 def _regex_to_static(src, regex):
     '''
--- a/salt/modules/file.py
+++ b/salt/modules/file.py
@@ -347,7 +347,7 @@ def user_to_uid(user):
         if isinstance(user, int):
             return user
         return pwd.getpwnam(user).pw_uid
-    except KeyError:
+    except (KeyError, NameError):
         return ''
 
 
--- a/salt/modules/file.py
+++ b/salt/modules/file.py
@@ -1743,7 +1743,7 @@ def _regex_to_static(src, regex):
     except Exception as ex:
         raise CommandExecutionError("{0}: '{1}'".format(_get_error_message(ex), regex))
 
-    return src and src or []
+    return src
 
 
 def _assert_occurrence(probe, target, amount=1):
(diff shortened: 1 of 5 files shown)


salt/modules/file.py
```diff
@@ -1680,44 +1680,19 @@ def _mkstemp_copy(path,
...
+    no_match = 1
     equal = 0
+
     if not src or not probe:
         return no_match
 
-    src = src.rstrip('\n\r')
-    probe = probe.rstrip('\n\r')
-    if src == probe:
+    if probe == src.rstrip('\r\n'):
```
Ch3LL: Just one question. Can you elaborate why you are not adding `rstrip('\n\r')` to probe as well?

I also added some tests for this PR here https://github.com/jubrad/salt/pull/1  if you can give that a review and merge so we can ensure this does not regress.

Ch3LL (changes requested):

salt/modules/file.py
jubrad: Thanks, sorry I believe that we've implemented a workaround so there is less pressure on a fix here from us, as such I've completely stopped thinking about it. 

Thanks for adding the tests.

The probe is coming from an argument to the method. I assumed any newlines in it would be intentional thus shouldn't be stripped, although it would probably cause less confusion and better code to just strip both.

Ch3LL:
thanks for those changes. looks like the windows tests are failing because it still references `_starts_till` in salt/modules/win_file.py. Can you update those references to `_line_match`?

Ch3LL:
bump @jubrad

Ch3LL:
also waiting on review from @waynew

waynew:
Oh, yeah, this one!

I'm pretty sure this does what it says it does, but I also discovered uh... several other issues with this module that were introduced. (Assuming tests pass) I'm OK with merging this, because it *is* an improvement, but we also certainly need to fix some of the other historical artifacts here.

@jubrad are you interested in fixing some of the other issues surrounding this bug, or just trying to get this fix in as-is?

waynew:
@jubrad Do you have time to work on this in the next day or two? We'd like to get this fix incorporated in time for our 2019.2.1 release. If not, I can go ahead and build off your work here. If I don't hear from you, I'll assume you're busy 😂

dwoz:
re-run full all

Ch3LL:
ping @waynew looks like there are some failures and we are trying to get this into 2019.2.1

waynew:
re-run py2-centos-7

waynew:
re-run full all

jubrad:
@waynew sorry I was on vacation during your message then I just completely dropped the ball. Looks like you've made some significant changes here. I'm happy to help move this along as needed and I'll make sure to increase the volume of my alerts for this pr.

waynew:
@jubrad now that we've switched to the master branch approach, https://github.com/saltstack/salt/pull/53888 has been rebased for master, so I'm going to go ahead and close this PR - feel free to watch that other PR, or reply on here if you have any questions!

Issue #52339, opened 2019-03-27
osrelease grain is None on Windows Embedded Standard 7 systems


andzn:
### Description of Issue/Question
After change  #50844 the os_release grain is None on Windows Embedded Standard systems. And because there's logic that expects the os_release grain is set, the salt-minion doesn't start.
It seems that the os info.caption is "Windows Embedded Standard ". Since the caption doesn't contain the version in it, the new code resolves the osrelease grain to unknown. 
Note that the platform.release() returned the correct version (7) on windows embedded standard 7. 

### Steps to Reproduce Issue
Run the salt-minion on Windows Embedded Standard 7 and it will crash because the osrelease grain is None. I have tested with salt-minion 2018.3 (python3)

### Versions Report
```
Salt Version:
           Salt: 2018.3.0-547-g4f8fe89

Dependency Versions:
           cffi: 1.11.5
       cherrypy: unknown
       dateutil: 2.7.3
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.10
        libgit2: 0.27.4
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.5.6
   mysql-python: Not Installed
      pycparser: 2.19
       pycrypto: Not Installed
   pycryptodome: Not Installed
         pygit2: 0.27.2
         Python: 3.6.8 (tags/v3.6.8:3c6b436a57, Dec 24 2018, 00:16:47) [MSC v.19
16 64 bit (AMD64)]
   python-gnupg: Not Installed
         PyYAML: 3.12
          PyZMQ: Not Installed
           RAET: Not Installed
          smmap: Not Installed
        timelib: Not Installed
        Tornado: 4.5.3
            ZMQ: Not Installed

System Versions:
           dist:
         locale: cp1252
        machine: AMD64
        release: Unknown
         system: Windows
        version: Unknown 6.1.7601 SP1 Multiprocessor Free
```

twangboy:
Please post the results of the following command on an Windows Embedded Standard 7 system
```
wmic os get /value
```

PR #52344, opened 2019-03-27
Fall back to old way of getting osrelease grain if new way fails

twangboy:
### What does this PR do?
Falls back to the old way of getting `osrelease` if the new way doesn't detect it. The new way wasn't detecting embedded versions of Windows correctly as they have no version number in the caption.
Adds some tests to make sure it's all working

https://github.com/saltstack/salt/issues/52339

### Tests written?
Yes

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/grains/core.py
+++ b/salt/grains/core.py
@@ -1131,6 +1131,57 @@ def _clean_value(key, val):
     return val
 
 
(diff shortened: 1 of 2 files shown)


dwoz (approved):

andzn (approved):
Thanks!

Issue #52230, opened 2019-03-18
_syspaths.py variables defaulted unless explicitly changed


alan-cugler:
### Intent
This is a discussion issue of a possible PR to the "develop" branch of salt. 

### Description of Issue
Currently the `_syspaths.py` file has a set default of "None" for all variables listed.
You can set these variables to custom directories for individual parts of the salt infrastructure.

However, if you don't state *all* of the variables in this file the import in [syspaths.py](https://github.com/saltstack/salt/blob/develop/salt/syspaths.py#L31) will result in an error like in #52228. 

### Proposed Change
The `import` line in the `syspaths.py` file should be modified to assume all non-mentioned variables default to "None" implicitly. This way the user can state only variables they are concerned with and ignore the rest whether they are there or not in the `_syspaths.py` file

A secondary addition would be to have `syspaths.py` add in the `[missing variable] = None` at the bottom of `_syspaths.py` for all variables missing from the import.

### Related Issues
- https://github.com/saltstack/salt/issues/52228
- https://github.com/saltstack/salt/issues/51768

nixjdm:
An additional motive for not having to declare attributes that would just be `None` in the custom `_syspaths.py` is that it would make an existing `_syspaths.py` more forwards compatible. If you have an existing setup where you need the custom module, and then upgrade Salt such that new attributes are looked for, it would likely just work instead of a getting an `AttributeError`.

I don't think we should do your second proposed change though. I don't think an existing `_syspath.py` that a user has customized or version controlled should be automatically written to. It's also not necessary especially if we get the first change.

Ch3LL:
ping @saltstack/team-core can i get your opinions here?

waynew:
On the face of it I agree with @nixjdm - seems like `_syspath.py` should be overriding the defaults.

I don't know if there were other more original ideas/reasons, though.

One counterpoint that I can think of - allowing missing values to pass silently *could* lead to problems with typos, or people missing out when new things are added that they may want to provide.

An approach that might be ideal is that if there *are* missing values that we provide a warning on startup suggesting that we'll set the value to `None` or whatever default value.

alan-cugler:
@waynew, my secondary addition had also been thinking about users missing new variables, but your suggestion of a warning message would be a better way. 

So if we are going to put a warning message then some thoughts for that:
 - The warning message will be at the warning log level? as opposed to debug and info levels
[...]

PR #52352, opened 2019-03-29
syspaths.py warning message

alan-cugler:
### What does this PR do?
- sets missing variable to `None`
- Created a warning level message:
  - lists missing variables
  - "typo?" prompt
  - shows how & where to explicitly set variables
#52230 
### Previous Behavior
- If variable is missing from _syspaths.py file salt breaks and gives a traceback.
### New Behavior
- Default sets variables not explicitly set in the `_syspaths.py` file
- Prompts user at log warning level about said variables.
### Tests written?
- Was manually tested, no automated tests written.
### Commits signed with GPG?
Yes


alan-cugler:
I did not add in salt docs reference or URL since said doc doesn't exist yet, and reviewing other log messages showed no such references or URLs set in the messages.

alan-cugler:
I don't get the space after comma error in pylint. later in the same file the lists don't have spaces after commas? pylint test is just bullying me... (T_T)

alan-cugler:
@Ch3LL can you explain the error?

Ch3LL:
you will want to make sure you point your pylint to the testing pylintrc file. For example I run: `python2 -m pylint --rcfile /home/ch3ll/git/salt/.testing.pylintrc tests/unit/utils/test_stringutils.py`


This should give you similar results as these tests.

But to help point you in the right direction this is how you will need to fix this particular lint error:

```
diff --git a/salt/syspaths.py b/salt/syspaths.py
index 0e61da7639..46c104e9d3 100644
--- a/salt/syspaths.py
+++ b/salt/syspaths.py
@@ -27,7 +27,7 @@ svl = ('ROOT_DIR', 'CONFIG_DIR', 'CACHE_DIR', 'SOCK_DIR',  # salt variable list
     'SRV_ROOT_DIR', 'BASE_FILE_ROOTS_DIR', 'HOME_DIR',
     'BASE_PILLAR_ROOTS_DIR', 'BASE_THORIUM_ROOTS_DIR',
     'BASE_MASTER_ROOTS_DIR', 'LOGS_DIR', 'PIDFILE_DIR',
-    'SPM_PARENT_PATH', 'SPM_FORMULA_PATH','SPM_PILLAR_PATH',
+    'SPM_PARENT_PATH', 'SPM_FORMULA_PATH', 'SPM_PILLAR_PATH',
     'SPM_REACTOR_PATH', 'SHARE_DIR')
 
 try:
```

notice the space after the SPM_FORMULA_PATH comma

alan-cugler:
Hey @Ch3LL, I'm not sure why the Centos test failed. This test passed last time, it was the pylint test that had cried wolf before.

alan-cugler:
Cool, not sure why it failed before, but now it passes all the tests!

salt/syspaths.py
```diff
@@ -21,9 +18,17 @@
 from __future__ import absolute_import, print_function, unicode_literals
 import sys
 import os.path
+import logging
 
 __PLATFORM = sys.platform.lower()
-
+typo_warning = False
+log = logging.getLogger(__name__)
+svl = ('ROOT_DIR', 'CONFIG_DIR', 'CACHE_DIR', 'SOCK_DIR',  # salt variable list
```
waynew: I think I'd prefer to call this `EXPECTED_VARIABLES`, since it shouldn't be changing. It's also more obvious what `EXPECTED_VARIABLES` means, than coming across `svl` somewhere in your code.

salt/syspaths.py
```diff
@@ -32,14 +37,19 @@
...
+else:
+    for key in svl:
+        if hasattr(__generated_syspaths, key):
+            continue
+        else:
+            if not typo_warning:
+                log.warning('Possible Typo?')
+                log.warning('To dissolve this warning add `[variable] = None` to _syspaths.py')
+            typo_warning = True
+            log.warning('Variable "' + key + '" is missing, value set to None')
```
waynew:
```suggestion
            log.warning('Variable %r is missing from `_syspaths.py`, value set to `None`', key)
```

salt/syspaths.py
```diff
@@ -32,14 +37,19 @@
...
         setattr(__generated_syspaths, key, None)
-
+else:
+    for key in svl:
+        if hasattr(__generated_syspaths, key):
+            continue
+        else:
+            if not typo_warning:
+                log.warning('Possible Typo?')
+                log.warning('To dissolve this warning add `[variable] = None` to _syspaths.py')
```
waynew: Was your thought here that we should only show this typo warning if there's more than one missing value? It seems somewhat odd to me that we'd show it for subsequent errors and not for the first (after all, the first one *could* be a typo).

I think I'd either display it for all or for none - though I'd probably just include it in the other log message.

waynew (changes requested):

salt/syspaths.py
alan-cugler: I can change it to that if you'd like. but I had a comment at the end of the line stating what svl stood for.

waynew: That's how I knew what it meant :)

The reason I prefer having it spelled out is that the other places have the full definition as well. And if you have to put a comment you may as well put it in code.

alan-cugler:
@waynew let me know if anything is still dis-satisfying for the PR. I ended up using `%s` instead of `%r` since the `%r` was printing the unicode with the key variable in the print.

waynew (approved):

waynew:
@alan-cugler Looking great! Do you think it would make sense to add a test ensuring that the expected values do get set? I'm not sure if that's a test we *could* automate very well?

alan-cugler:
@waynew I am still learning your guy's test suite and how to use the mock library. We can open an issue to add a test in.

PR #52357, opened 2019-03-29
Add support for trusted repos.

chotea:
Check kwargs in mod_repo for trusted option and set it to
the feed configuration file if needed. Read the trusted option 
from the feed configuration file.
Example: 
src/gz feedname feeduri [trusted=yes]

Signed-off-by: Cristian Hotea <cristian.hotea@ni.com>

### Tests written?
No

### Commits signed with GPG?
No


rares-pop (approved):
Changes looks good!

dwoz (approved):
@chotea Thanks for the contribution. In the future it would be best if you could add a test for this type of change. :)

PR #52365, opened 2019-03-30
Stops the Window installer from hanging

cmcmarrow:
*Removed “-NoNewWindow” which was the cause of the hang when pip ran
*Cleaned up start-process-and-test-exitcode.psm1
*Switched the order of Removed Temporary and script complete in “build_env_3.ps1”
*www.github.com/saltstack/salt/issues/52265

### What does this PR do?
*Removed “-NoNewWindow” which was the cause of the hang when pip ran
*Cleaned up start-process-and-test-exitcode.psm1
*Switched the order of Removed Temporary and script complete in “build_env_3.ps1”

www.github.com/saltstack/salt/issues/52265

### Tests written?

No

### Commits signed with GPG?

Yes


pkg/windows/build_env_3.ps1
```diff
@@ -276,6 +276,15 @@ ForEach($key in $ini[$bitDLLs].Keys) {
...
 
+#------------------------------------------------------------------------------
+# Remove the temporary download directory
+#------------------------------------------------------------------------------
+Write-Output " ----------------------------------------------------------------"
+Write-Output " - $script_name :: Cleaning up downloaded files"
+Write-Output " ----------------------------------------------------------------"
+Write-Output ""
+Remove-Item $($ini['Settings']['DownloadDir']) -Force -Recurse
+
```
twangboy: Can you make the same change in the `build_env_2.ps1` file so that the are basically the same?

twangboy (changes requested):

pkg/windows/build_env_3.ps1
cmcmarrow: I switched it back to normal. Someone might want it to be in that order.

pkg/windows/build_env_2.ps1
```diff
@@ -289,3 +289,4 @@ Write-Output " - $script_name :: Cleaning up downloaded files"
 Write-Output " ----------------------------------------------------------------"
 Write-Output ""
 Remove-Item $($ini['Settings']['DownloadDir']) -Force -Recurse
+
```
twangboy: Since we're not making any changes to this file, go ahead and remove the new blank line at the end.

twangboy (changes requested):

twangboy (approved):

Issue #46902, opened 2018-04-05
Intermittent failure from apt-get locking on 16.04 minion 


chrisportela:
### Description of Issue/Question
All my pkg related states on Ubuntu 16.04 keep failing because of a lock.

Part of my states are to ensure unattended upgrades and other apt services are disabled, but these aren't working. 

### Setup
Building an AMI using Packer with a Salt Masterless provisioner (i.e. `salt-call`)

```
[snip]
    amazon-ebs:     Function: service.dead
    amazon-ebs:         Name: unattended_upgrades
    amazon-ebs:       Result: True
    amazon-ebs:      Comment: The named service unattended_upgrades is not available
    amazon-ebs:      Started: 15:04:59.987639
    amazon-ebs:     Duration: 14.204 ms
    amazon-ebs:      Changes:
    amazon-ebs: ----------
    amazon-ebs:           ID: unattended_disabled
    amazon-ebs:     Function: service.masked
    amazon-ebs:         Name: unattended_upgrades
    amazon-ebs:       Result: True
    amazon-ebs:      Comment: Service unattended_upgrades was runtime masked
    amazon-ebs:      Started: 15:05:00.002455
    amazon-ebs:     Duration: 82.046 ms
    amazon-ebs:      Changes:
    amazon-ebs:               ----------
    amazon-ebs:               runtime masked:
    amazon-ebs:                   ----------
    amazon-ebs:                   new:
    amazon-ebs:                       True
    amazon-ebs:                   old:
    amazon-ebs:                       False
    amazon-ebs: ----------
    amazon-ebs:           ID: apt_daily_timer_dead
    amazon-ebs:     Function: service.dead
    amazon-ebs:       Result: True
    amazon-ebs:      Comment: The named service apt_daily_timer_dead is not available
    amazon-ebs:      Started: 15:05:00.084770
    amazon-ebs:     Duration: 11.926 ms
    amazon-ebs:      Changes:
    amazon-ebs: ----------
    amazon-ebs:           ID: uptodate
    amazon-ebs:     Function: pkg.uptodate
    amazon-ebs:       Result: False
    amazon-ebs:      Comment: Problem encountered upgrading packages. Additional info follows:
    amazon-ebs:
    amazon-ebs:               result:
    amazon-ebs:                   ----------
    amazon-ebs:                   pid:
    amazon-ebs:                       3344
    amazon-ebs:                   retcode:
    amazon-ebs:                       100
    amazon-ebs:                   stderr:
[...]

PR #52378, opened 2019-04-01
fix 46902: wait dpkg lock

aarnaud:
https://github.com/saltstack/salt/issues/46902

### Previous Behavior
All pkg related states failed when dpkg lock cause by unattended-upgrade for example

### New Behavior
aptpkg wait for dpkg lock before continue, it will continue anyway if lock still present after 5min.

### Tests written?

No

### Commits signed with GPG?

Yes


aarnaud:
Thanks to @johanfleury, i updated my PR to use a better way to check the dpkg lock

dwoz (changes requested):
@aarnaud Please see the linter and add a regression test for this bugfix.

aarnaud:
Hi @dwoz
I didn't the lint error. it's fixed now

Ch3LL:
@aarnaud thanks for fixing the lint failures. looks like there are some other apt test failures. Can you also write a regression test for this fix?

aarnaud:
Hi @Ch3LL 

I don't see how can i write regression test for this fix especially. Because i have to launch apt-get in same time with saltstack.

dmurphy18:
@aarnaud 

If you could change
    except OSError as e:
to 
  except (IOError, OSError) as e:

since in Python 2.7, fcntl will throw an IOError, but in Python 3.5+ it throws an OSError and still have to support Python 2.7 at the moment.

aarnaud:
@dmurphy18 I update the PR with `except (IOError, OSError) as e:`

Ch3LL:
@aarnaud you should be able to add some unit tests around your new functions which would not require launching apt-get

aarnaud:
Hi @Ch3LL,

I look into `test_pkg.py`, i don't know how i can lock the file and unlock during `self.run_function` 

like 
```
lock /var/lib/dpkg/lock-frontend
ret = self.run_function(func)
unlock before the timeout in self.run_function
```

Ch3LL:
can you point me to which test you are referring to? And I can dive in and see if we cant iterate upon it

salt/modules/aptpkg.py
```diff
@@ -158,6 +159,35 @@ def _check_apt():
...
+    else:
+        fcntl.lockf(fd, fcntl.F_UNLCK)
+        return False
+
+
+def _wait_dpkg_lock():
+    '''
+    Wait for dpkg lock release
+    '''
+    if os.path.isfile("/var/lib/dpkg/lock-frontend"):
```
rares-pop: This doesn't solve anything.

What if the lock happens after this check, but before the next statement?
According to some documentation, sometimes is /var/lib/dpkg/lock-frontend, sometimes /var/lib/dpkg/lock, depending on which tool is run to update/install stuff.

rares-pop (commented):
I don't think this change should go in, and the case you are trying to solve should be handled at a higher level, by adding error-handling and retry.

salt/modules/aptpkg.py
aarnaud: Since 1 Apr, we are using this patch internally, and we don't have any error anymore. 
Maybe it's not the best fix but a don't know where we can fix that

aarnaud: @Ch3LL https://github.com/saltstack/salt/blob/develop/tests/integration/states/test_pkg.py

rares-pop: You can't avoid the race-condition, that's why I say you should handle the failing of the job by having a retry mechanism.

aarnaud:
> I don't think this change should go in, and the case you are trying to solve should be handled at a higher level, by adding error-handling and retry.

- Error handling have to parse the apt error to catch the lock error, seem to me not a good idea.
- And retry on error, will retry for other error like missing package if the name has typo. 

But i understand our point of `if the lock happens after this check, but before the next statement` and i don't have answer for that.
We you use cloud instance `unattended-upgrades` is installed, and run on boot, saltstack failed 70% of the time

salt/modules/aptpkg.py
aarnaud: - Error handling have to parse the apt error to catch the lock error, seem to me not a good idea.
- And retry on error, will retry for other error like missing package if the name has typo. 

But i understand our point of `if the lock happens after this check, but before the next statement` and i don't have answer for that.
When you use cloud instance `unattended-upgrades` is installed, and run on boot, saltstack failed 70% of the time

Ch3LL:
looking at https://github.com/saltstack/salt/blob/develop/tests/integration/states/test_pkg.py

you could possibly run `cmd.run` to run `apt-get install <pgk>` with `bg`set so it does not wait for the test results. And then run a pkg.install state to ensure another pkg gets installed and waits for the pkg to finish installing previously.

salt/modules/aptpkg.py
Ch3LL: > "Error handling have to parse the apt error to catch the lock error, seem to me not a good idea"

Can you explain why that is not a good idea. Is it not possible to parse for the lock file error?

aarnaud: Error message can changed in future release and break the behaviour,

Put a can update my PR by catching `Could not get lock`
```
E: Could not get lock /var/lib/dpkg/lock-frontend - open (11: Resource temporarily unavailable)
E: Unable to acquire the dpkg frontend lock (/var/lib/dpkg/lock-frontend), is another process using it?
```
```python
def _call_apt(args, scope=True, **kwargs):
    '''
    Call apt* utilities.
    '''
    cmd = []
    if scope and salt.utils.systemd.has_scope(__context__) and __salt__['config.get']('systemd.scope', True):
        cmd.extend(['systemd-run', '--scope'])
    cmd.extend(args)

    params = {'output_loglevel': 'trace',
              'python_shell': False,
              'env': salt.utils.environment.get_module_environment(globals())}
    params.update(kwargs)

    _wait_dpkg_lock()
    return __salt__['cmd.run_all'](cmd, **params)
```

by somethings like that

```python
def _call_apt(args, scope=True, **kwargs):
    '''
    Call apt* utilities.
    '''
    cmd = []
    if scope and salt.utils.systemd.has_scope(__context__) and __salt__['config.get']('systemd.scope', True):
        cmd.extend(['systemd-run', '--scope'])
    cmd.extend(args)

    params = {'output_loglevel': 'trace',
              'python_shell': False,
              'env': salt.utils.environment.get_module_environment(globals())}
    params.update(kwargs)

    cmd_ret = __salt__['cmd.run_all'](cmd, **params)
    count = 0
    while "Could not get lock" in cmd_ret['stderr'] and count < 100:
        count += 1
        log.warning('Waiting for dpkg lock release: retrying... %s/100', count)
        time.sleep(3)
        cmd_ret = __salt__['cmd.run_all'](cmd, **params)
    return cmd_ret
```

rares-pop: Again - your box seems to have acquired the dpkg lock from a graphical interface app, hence the lock-frontend. Can you check if you do a manual install like apt-get/apt/aptitude install foo from a terminal, if it's the same lock-frontend lock that's acquired? Some documentations mention '/var/lib/dpkg/lock'. When is when?

reference: https://askubuntu.com/questions/827212/which-is-the-process-using-apt-get-lock/827239
E: Could not get lock /var/lib/dpkg/lock - open (11: Resource temporarily unavailable)
E: Unable to lock the administration directory (/var/lib/dpkg/), is another process using it?
In theory the error messages can change, but in practice they rarely do, if ever. Salt is returning the underlying executable errors in almost all the cases (execution modules), so things will break pretty bad if that happens.

Doing this change in salt will give you an impression that your function is safe, and will wait, but it's not. I let other decide if it's meaningful for salt or not.

aarnaud: Or if one day salt used python-apt we can handle the exception: 
https://github.com/ansible/ansible/blob/devel/lib/ansible/modules/packaging/os/apt.py#L1193

I will change that by parsing the stderr.

aarnaud: > Again - your box seems to have acquired the dpkg lock from a graphical interface app, hence the lock-frontend. Can you check if you do a manual install like apt-get/apt/aptitude install foo from a terminal, if it's the same lock-frontend lock that's acquired? Some documentations mention '/var/lib/dpkg/lock'. When is when?

`lock-frontend` is when you use apt, it's not link to graphical interface
`lock` is when you use dpkg directly

aarnaud:
@Ch3LL @rares-pop PR updated by parsing stderr.

Ch3LL:
@aarnaud  looks like there are some test failures. Also just looking to get some test coverage for this change. I included a test suggestion above, but let me know if its not clear.

also @rares-pop can you re-review here? thanks

rares-pop:
The change looks good to me now! The behavior is correct - no races can happen.

I'm not sure about the 3 seconds wait for 100 iterations. In terms of, if it would be better to have a more elaborated backoff strategy - https://en.wikipedia.org/wiki/Exponential_backoff , but that's very minor if anything at all.

rares-pop (approved):

aarnaud:
Hi @rares-pop I added an basic exponential backoff.

salt/modules/aptpkg.py
```diff
@@ -172,7 +172,14 @@ def _call_apt(args, scope=True, **kwargs):
               'env': salt.utils.environment.get_module_environment(globals())}
     params.update(kwargs)
 
-    return __salt__['cmd.run_all'](cmd, **params)
+    cmd_ret = __salt__['cmd.run_all'](cmd, **params)
+    count = 0
+    while 'stderr' in cmd_ret and "Could not get lock" in cmd_ret['stderr'] and count < 10:
```
rares-pop: This used to be a python anti-pattern, to do:
`if 'foo' in dict and 'blah' == dict['foo']`
the recommended would be:
`errText = cmd_ret.get('stderr')`
`if 'Could not get lock' in errText`

or directly:
`while 'Could not get lock' in cmd_ret.get('stderr', '')`

rares-pop:
And btw, you'll probably want to make this change against the 'master' branch. There's an effort to move all existing open PRs over to 'master'. Saltstack changed the branching strategy.

salt/modules/aptpkg.py
aarnaud: yes it's better, I changed the condition by your example

aarnaud:
I will rebase my branch on master and change the PR

rares-pop (dismissed):
LGTM!

I hate to dispatch bad news though, but saltstack's policy now require tests for *any* PR. You might get away touching existing tests if any, to exercise the retry.

dmurphy18 (dismissed):

dmurphy18:
@aarnaud Any chance of writing a test for this, perhaps with mock to return that lock is active and ensure it backs off.  Only looking at this again now, apologies, thought this had been merged up long ago, but everything needs tests now in order to stabilize the code better.

aarnaud:
thanks @garethgreenaway for the tests https://github.com/saltstack/salt/commit/43802835f5535db4d92cae8f825c8b5401636707

dwoz (approved):

Issue #34096, opened 2016-06-17
Poor quality control


coen-hyde:
Hey Guys,

This issue isn't about a particular issue I am experiencing. Rather that lack of quality control around salt. I've experienced so many issues in setting up our salt environment; most of which already have tickets in github. It's really disheartening to find out that the project that you championed is so buggy. I'm about to rip Salt out from our infrastructure because I can not get it to work reliably. And I've lost confidence in it.

I hope this ticket might bring to attention the serious state of affairs regarding the quality of salt. 

Thanks,
Coen

githubcdr:
I have to agree, there are a lot of regression bugs lately, some serious some just annoying, some to mention;

- Upgrading Salt from 2015 to 2016 made me lose all my minion because the minion needs double restart bug in the package.
- After upgrading Salt my "pkg upgrade state" failed because of the extra kwargs bug in pkg state.
- Beacons seem to be unreliable, in fact I never got them to work for diskusage. (known issue)
- For some reason salt-cloud is giving 404 errors when deleting instances on Digital Ocean.
- When releasing a new Salt version, the win-repo git data is not in sync, minor thing but these things matter.

I love Salt but I really feel that there should be more focus on bugs because I'm starting to feel the same pain as @coen-hyde, it's getting frustrating.

Keep up all the good work, Salt is amazing but don't lose control over it.

thatch45:
Yes, this is a never ending challenge in software development and one that is particularly difficult in Salt. We do have quite a lot in place to help minimize these issues are are working on more.

Our automated test suite is run against multiple platforms for every pull request and must pass before all releases and all code that is introduced into Salt.
We have an extensive set of manual tests which are run before every release as well.

We are currently working hard on expanding the test suite as well.

One of the main problems we face is the sheer coverage and usage of Salt. In this regard we face many similar problems that are faced by kernel and OS developers, due to the fact that we cannot control the environment in which our software runs, and we need to have such a broad management of the environment.

With all this said, we do not shrink from the challenge and we feel that we have greatly improved! 2016.3.0 is a .0 release and I am very pleased to say that while regressions were introduced it was by far the most stable .0 release we have had in years and that I am very excited about it. Also we have been able to quickly handle almost all of the regressions and many bugs in the .1 release. So I think that our ongoing efforts are paying off!

[...]

PR #52379, opened 2019-04-01
fix 46902: wait dpkg lock

aarnaud:
https://github.com/saltstack/salt/issues/46902

### Previous Behavior
All pkg related states failed when dpkg lock cause by unattended-upgrade for example

### New Behavior
aptpkg wait for dpkg lock before continue, it will continue anyway if lock still present after 5min.

### Tests written?

No

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/modules/aptpkg.py
+++ b/salt/modules/aptpkg.py
@@ -14,6 +14,7 @@ from __future__ import absolute_import, print_function, unicode_literals
 
 # Import python libs
 import copy
+import fcntl
 import os
 import re
 import logging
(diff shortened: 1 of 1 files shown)


aarnaud:
Thanks to @johanfleury, i updated my PR to use a better way to check the dpkg lock

rares-pop (changes requested):
The logic is flawed again. You fixed it in the previous PR, but for some reason you went back.

Again, you have a race-condition - your _wait_dpkg_lock() can say it's not locked, but when you actually do the __salt__['cmd.runall'](cmd, **params) be locked.

johanfleury:
@rares-pop the way this repository is managed forces us to create multiple PRs to be sure our changes are pushed to every existing releases (this is honestly the worst way of doing release management I've seen, and [I'm not the only one](https://github.com/saltstack/salt/issues/34096)).

johanfleury:
The way Salt releases are managed in this repository forces us to create multiple PRs if we want to see our changes published. This is honestly the worst way of doing release management I've seen (multiple release branches, new releases not sync'ed with develop, etc.) and [others share that feeling too](https://github.com/saltstack/salt/issues/34096).

I think @aarnaud only forgot to update this PR. The “correct” one (the one where discussions took place) is #52378.

rares-pop:
> The way Salt releases are managed in this repository forces us to create multiple PRs if we want to see our changes published. This is honestly the worst way of doing release management I've seen (multiple release branches, new releases not sync'ed with develop, etc.) and [others share that feeling too](https://github.com/saltstack/salt/issues/34096).

I think you got it wrong. Creating multiple PRs won't help you much. There is actually a policy for salt changes, which is - bug fixes should go *only* in the oldest supported release that manifest the bug (and from there on it's saltstack's responsibility to forward-port it everywhere). Features should *only* go in 'develop', and it will make it in the next release (not the one that has feature-frozen due to immediate release).

> I think @aarnaud only forgot to update this PR. The “correct” one (the one where discussions took place) is #52378.

Again, you should decide if it's a fix, or a feature, and create just one PR in the right place.

Once you mentioned this, I saw it's an older PR. Apologize.

aarnaud:
It's a bugfix for me. I close this PR and we keep #52378.

Issue #51180, opened 2019-01-15
salt.http.query() causes jinja template render failure, despite "raise_error = false" if server does not respond or response has zero length


TheCatFelix:
### Description of Issue/Question

salt.http.query() causes jinja template render failure, despite "raise_error = false" if server does not respond or response has zero length.

Relevant error message:

local:
    Data failed to compile:
----------
    Rendering SLS 'base:test' failed: Jinja error: object of type 'NoneType' has no len()
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/salt/utils/templates.py", line 393, in render_jinja_tmpl
    output = template.render(**decoded_context)
  File "/usr/lib/python3/dist-packages/jinja2/asyncsupport.py", line 76, in render
    return original_render(self, *args, **kwargs)
  File "/usr/lib/python3/dist-packages/jinja2/environment.py", line 1008, in render
    return self.environment.handle_exception(exc_info, True)
  File "/usr/lib/python3/dist-packages/jinja2/environment.py", line 780, in handle_exception
    reraise(exc_type, exc_value, tb)
  File "/usr/lib/python3/dist-packages/jinja2/_compat.py", line 37, in reraise
    raise value.with_traceback(tb)
  File "<template>", line 2, in top-level template code
  File "/usr/lib/python3/dist-packages/salt/modules/http.py", line 38, in query
    return salt.utils.http.query(url=url, opts=opts, **kwargs)
  File "/usr/lib/python3/dist-packages/salt/utils/http.py", line 613, in query
    result_text = __decompressContent(coding, result_text)
  File "/usr/lib/python3/dist-packages/salt/utils/http.py", line 113, in __decompressContent
    log.trace("Decompressing %s byte content with compression type: %s", len(pgctnt), coding)
TypeError: object of type 'NoneType' has no len()

; line 2

---

{% set test = salt.http.query('https://not.a.resolvable.server', raise_error = False) %}    <======================

/home/test/salt/query.txt:
 file.managed:
 - contents: {{ test }}

[...]

Note that when I comment out the following two lines in /usr/lib/python3/dist-packages/salt/utils/http.py, the state runs just fine:

   113	    #log.trace("Decompressing %s byte content with compression type: %s", len(pgctnt), coding)
   133	    #log.trace("Content size after decompression: %s", len(pgctnt))

It appears that the root of the exception is calling len() on an empty object.

### Setup
(Please provide relevant configs and/or SLS files (Be sure to remove sensitive info).)

[...]

Issue #49602, opened 2018-09-11
[bug] salt.util.http implementations should all provide the same behavior


jdelic:
Salt states and modules that need to make HTTP queries go through the `salt.modules.http` execution module, which itself delegates to `salt.util.http`. `salt.util.http.query` consists of three different implementations depending on what libraries are available, selected by the `backend` keyword parameter. The available backends are:

* `tornado` using tornado's HTTPClient (the default)
* `requests` using the requests Python library
* `urllib2` using the Python built-in urllib2 library

Unfortunately `tornado` subtly behaves differently from the other two. Among other things it's the only backend that supports the `raise_error` parameter and is therefor the only backend that allows a query to actually return any status code that is not `2xx`, since every other backend will throw an exception on any status code that might indicate an error.

An example usage might look like this:
```yaml
    http.wait_for_successful_query:
        - name: https://vault.service.consul:8200/v1/sys/health
        - match: "initialized": false
        - wait_for: 10
        - request_interval: 1
        - raise_error: False  # only exists in 'tornado' backend
        - backend: tornado
        - watch:
            - service: vault-service
        - require_in:
            - cmd: vault-sync
```

[Hashicorp Vault will return status code 501](https://www.vaultproject.io/api/system/health.html#read-health-information) when Vault is running but is not initialized. The above state will wait for Vault to be up, before Salt can then move on to initialize it. With any other backend it will fail, since they will raise exceptions on status code 501 regardless of the `match`.

garethgreenaway:
@saltstack/team-cloud This seems like a bug.  Thoughts?

gtmanfred:
Yeah, i don't see a reason to not add raise_error: support to the other backends.

jdelic:
I had another issue that made me look more into `salt.util.http` and I wrote a longer comment on a PR that summarizes additional issues with this code. I hope that if someone from Salt takes a look at it they'll find the number of inconsistencies listed here useful: https://github.com/saltstack/salt/pull/52383#issuecomment-479680198.

twangboy:
Closing this issue due to age and lack of activity. Please test this on version 3006.2 and create a new issue if the problem persists. The new issue template has more information and will allow us to track and reproduce the issue more effectively. Thanks!

PR #52383, opened 2019-04-02
correctly reraise exceptions in states.http

jdelic:
### What does this PR do?
It improves error reporting for unexpected circumstances in `salt.states.http`. I stumbled over this while investigating saltstack/salt#51180.

As per the contribution guidelines I'm filing this bugfix PR against the lowest supported release, which I believe to be 2018.3.

### Previous Behavior
When the http state returned `None` as the value of the body, the previous code would raise a meaningless stacktrace giving no information about where the actual problem occurred. 

### New Behavior
This PR uses six to reraise the exception correctly in Python 2 and 3 by saving the stack trace first and ensuring that the correct stack trace is then raised.

### Tests written?
No. The behavior has not changed only the exceptions content.

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/states/http.py
+++ b/salt/states/http.py
@@ -12,6 +12,10 @@ from __future__ import absolute_import, print_function, unicode_literals
 import logging
 import re
 import time
+import sys
+
+from salt.ext import six
+
 
 __monitor__ = [
         'query',
@@ -149,18 +153,19 @@ def wait_for_successful_query(name, wait_for=300, **kwargs):
 
     while True:
         caught_exception = None
+        exception_type = None
+        stacktrace = None
         ret = None
         try:
             ret = query(name, **kwargs)
             if ret['result']:
                 return ret
         except Exception as exc:
-            caught_exception = exc
+            exception_type, caught_exception, stacktrace = sys.exc_info()
 
         if time.time() > starttime + wait_for:
             if not ret and caught_exception:
-                # workaround pylint bug https://www.logilab.org/ticket/3207
-                raise caught_exception  # pylint: disable=E0702
+                six.reraise(exception_type, caught_exception, stacktrace)
             return ret
         else:
             # Space requests out by delaying for an interval


dwoz:
@jdelic In cases like these we could add a test that validates the exception content is what we expect.

jdelic:
@dwoz 
As the code above my change catches `Exception` (which is already smelly, isn't it?), I wouldn't know how to check for the "expected exception content", because which ones of the available exceptions do you expect or test for?

(I would also argue that you should never test for exception content, but that's more of a philosophical point about code quality).

The ticket I'm referring to (#51180) describes that functions two trace levels down from `salt.states.http` believe that the returned body of a HTTP request will never be `None`. So simply running with this change from this PR and just returning from `__decompressContent` I immediately ran into the next block of code in `salt.utils.http` that expects that everything returned from either `urllib3`, `tornado.httpclient` or `requests` as a response body will be iterable or have attributes.

So those would be `TypeError`s with two different contents. Of the other three backends:

* only the Tornado HTTP backend respects `raise_error` (I filed this as #49602 a while ago), but if it _is set_, it will raise `tornado.httpclient.HTTPError`
* The Requests backend blindly calls `response.raise_for_status()` which will raise a `RequestException` for any non-2xx status code in addition to other exceptions
* Urllib2 catches all `URLError`s which includes `urllib2.HTTPError` which is actually how urllib2 communicates non-2xx status codes, so it won't even raise an exception for a invalid hostname which both other backends would, but ignores the part of the urllib2 documentation where it states that the returned `result` can be `None`, but let's be generous and assume that the code somewhere up the chain ensures that the URL can be handled by `salt.util.http.query`.

In theory the urllib2 code could bubble `socket.timeout` under Python 2.7 (because 2.7 stopped wrapping `socket.timeout`s in `URLError`s and the code only catches `URLError`), but the urllib2 code doesn't respect the `timeout` parameter either, so it doesn't matter.

So by my count that's `TypeError`, `AttributeError`, Tornado's `HTTPError`, `RequestException`, `URLError` and (only in theory) `socket.timeout`.

I'd rather not :).

dwoz:
@jdelic I don't think it's important to add a test for each exception. I am more referring to a test that validates an exception get's re-raised where expected. It could be any arbitrary exception, just test that the change does what is expected when an exception occurs.

jdelic:
@dwoz 
> I am more referring to a test that validates an exception get's re-raised where expected. It could be any arbitrary exception, just test that the change does what is expected when an exception occurs.

...now I honestly can't quite understand what test you would expect.

If you don't test for the content of an exception, which requires that you expect a specific exception (which will be difficult to trigger given  my explanation above) and therefor test a specific path that yields that specific exception, then what is there to do?

Because if you mock the call to `salt.utils.http.query` so that it raises a guaranteed exception you're effectively just unit testing that `sys.exc_info()` works as documented, which is a useless test in the scope of Salt.

I honestly wouldn't expect this change to be covered by a test. :man_shrugging:

mchugh19:
Since there isn't any logic change here, is a test really necessary to validate `six.reraise`?

dwoz:
@mchugh19 Valid point, I suppose not.

dwoz (approved):

Ch3LL:
closing in favor of https://github.com/saltstack/salt/pull/62595 which ports the fix to hte master branch and includes a changelog file.

Issue #52079, opened 2019-03-08
integration.states.test_cron.CronTest.test_managed


dwoz:
```
integration.states.test_cron.CronTest.test_managed
--------------------------------------------------------------------------------
develop failed salt-fedora-28-py2, salt-fedora-28-py3, salt-fedora-29-py2, salt-fedora-29-py3
--------------------------------------------------------------------------------
--------------------------------------------------------------------------------
2019.2 failed salt-centos-6-py2, salt-fedora-28-py2, salt-fedora-28-py3, salt-fedora-29-py2
--------------------------------------------------------------------------------
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
u'diff' not found in {}
................................................................................

Traceback (most recent call last):
  File "/tmp/kitchen/testing/tests/integration/states/test_cron.py", line 54, in test_managed
    self.assertIn('diff', ret['cron_|-salt://issue-46881/cron_|-salt://issue-46881/cron_|-file']['changes'])
AssertionError: u'diff' not found in {}
```

waynew:
Man... I cannot seem to reproduce this via kitchen -_-

This tests, and everything up to `integration.states.test_cron` all worked fine on Centos. Trying with fedora py2 now...

waynew:
While I couldn't reproduce this issue, I did realize that the test output could be improved by changing the assertion style. Created a PR for the change, so if this test fails again we should be able to get a more useful output than `AssertionError: u'diff' not found in {}`

Ch3LL:
its failing every test run on suse 42 py2: https://jenkinsci.saltstack.com/job/2019.2.1/job/salt-opensuse-42-py2/7/testReport/junit/integration.states.test_cron/CronTest/test_managed/

not sure if that will help you to replicate

waynew:
[...]

PR #52385, opened 2019-04-02
Refactor test_cron test

waynew:
### What does this PR do?

Refactors the one `test_cron` state test that we have, because #52079

It's possible that the dict could contain some more helpful information
when/if this test fails again. This assertion improves the potential for
getting useful information in the test failure.

#52079

### Previous Behavior

The assert statement failed to have any useful information in it.

### New Behavior

Now when the test fails, it should have some useful information in the failure

### Tests written?

Yes (changed the test - verified that changing the expected value failed the test and provided useful information, i.e. the entire contents of the changes, then changed it back and verified that the test passes)

### Commits signed with GPG?

Yes


dwoz (changes requested):
@waynew Please have a look at the test failures.

waynew:
Well, I fixed the failing Ubuntu tests @dwoz - it just completely failed running on win 2016. I'm re-running that job 🤞 (but in general I don't think that that failure is related to my changes)

waynew:
Currently still running, so that's a promising sign

dwoz (approved):

Issue #52087, opened 2019-03-08
integration.modules.test_mine.MineTest.test_mine_delete


dwoz:
```
integration.modules.test_mine.MineTest.test_mine_delete
--------------------------------------------------------------------------------
2018.3 failed salt-fedora-28-py2
--------------------------------------------------------------------------------
--------------------------------------------------------------------------------
2019.2 failed salt-fedora-28-py2
--------------------------------------------------------------------------------
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
[...]

PR #52393, opened 2019-04-02
Use uncommon item in mine delete test

waynew:
### What does this PR do?

As an attempt to prevent #52087 from failing during a (probable) race condition, rather than using `grains.items`, I've refactored this test to use `test.arg` which is less common, and
specifying a value that isn't currently used anywhere else.

#52087 

### Previous Behavior

If a `mine.send \* grains.items` was launched, or for any other reason the mine was updated with grains between `mine.delete` and `mine.get`, this test would fail due to the race condition (see #52087)

### New Behavior

Use `test.arg` with an unlikely value in an attempt to avoid any possibility of race condition

### Tests written?

Yes (and, since this was changing the test, I also manually verified - failing to actually delete the the value in the mine makes the test fail).

### Commits signed with GPG?

Yes

### DIFF
--- a/tests/integration/modules/test_mine.py
+++ b/tests/integration/modules/test_mine.py
@@ -111,35 +111,72 @@ class MineTest(ModuleCase):
                 ['grains.items']
             )
         )
+        # Smoke testing that grains should now exist in the mine
+        ret_grains = self.run_function(
+            'mine.get',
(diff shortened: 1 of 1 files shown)


dwoz (approved):
LGTM, I've re-launched the centos py3 tests which should pass now.

Issue #2130, opened 2012-09-28
cpan function needed to install Perl modules on server via state file


druonysus:
Currently there is no native Salt function to install Perl modules on systems from the state file unless there is a system package. Not every Perl module is packaged as an RPM or DEB. I would like to be able to just use something like cpan.install: in my state file to install Perl modules (similar to the pip function for Python modules).

mkotsbak:
This one should be reopened and implemented now that there is an execution module: https://github.com/saltstack/salt/pull/15214

rallytime:
@mkotsbak Sure thing!

Akm0d:
I'm looking into this right now and the toughest challenge is that cpan doesn't support uninstalling packages.  Would it be too outrageous to require cpanminus instead?  That or exclude the "cpan.absent" function.

Akm0d:
For now I'm going to do it with cpan and not support uninstalling modules with states.  In the future we should consider using cpanminus instead of cpan... although that creates a  whole new slew of problems.  For example, cpanminus can't see some packages installed by cpan.

Akm0d:
Just need to write some tests then will make a PR

Akm0d:
This has been merged to develop.  We can open a new issue if anyone wants to discuss adding functionality for cpanminus, rakudo, perl6 or uninstalling from states.

PR #52395, opened 2019-04-02
Cpan state

Akm0d:
### What does this PR do?
Adds a cpan state module
Does some fixes on 

https://github.com/saltstack/salt/issues/2130

### Tests written?

Yes

### Commits signed with GPG?

Yes


salt/states/cpan_state.py
```diff
@@ -0,0 +1,206 @@
+# -*- coding: utf-8 -*-
```
mchugh19: How about using the filename `cpan`? This would be more inline with other state modules, and make more obvious the state declaration of `cpan.installed:`

Akm0d: I based it of of the pip_state.py module which happens to be the only one that has `_state` in the name, It's a good suggestion and I"ll make the change :+1:

Akm0d: Fixed in https://github.com/saltstack/salt/pull/52395/commits/d8a2a0efc3089d59e54adeb35f79978cd822d5b3

twangboy:
@Akm0d Looks like some of the cpan tests are failing

Akm0d:
> @Akm0d Looks like some of the cpan tests are failing

Yeah, I'm still working those out

Issue #51817, opened 2019-02-26
Saltclass external pillar broken in 2019.2.0


jkrauz:
### Description of Issue/Question
After upgrading to 2019.2.0 our states are playing weirdly. I believe saltclass external pillar expansion is broken.

### Setup
```
# saltclass/nodes/node1.yml
classes:
    - child
pillars:
    foo:
      id: node1
```
```
# saltclass/classes/child.yml
classes:
    - parent
pillars:
    foo:
      id: child
      list:
        - child1
        - child2
```
```
# saltclass/classes/parent.yml
environment: base
pillars:
    foo:
      id: parent
      list:
        - parent1
        - parent2
```

### Steps to Reproduce Issue
salt-call pillar.get foo
```
local:
    ----------
    id:
        parent
    list:
        - parent1
        - parent2
        - child1
        - child2
        - parent1
        - parent2
        - child1
        - child2
        - parent1
        - parent2
```
The output above doesn't look correct. The expected output is:
```
local:
    ----------
    id:
        node1
    list:
        - child1
        - child2
        - parent1
        - parent2
```

### Versions Report
Salt Version:
           Salt: 2019.2.0

Dependency Versions:
           cffi: Not Installed
       cherrypy: Not Installed
       dateutil: Not Installed
      docker-py: Not Installed
          gitdb: 2.0.2
      gitpython: 2.1.3
          ioflo: Not Installed
         Jinja2: 2.7.2
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.5.6
   mysql-python: Not Installed
      pycparser: Not Installed
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: Not Installed
         Python: 2.7.5 (default, Oct 30 2018, 23:45:53)
   python-gnupg: Not Installed
         PyYAML: 3.11
          PyZMQ: 15.3.0
           RAET: Not Installed
          smmap: 2.0.2
        timelib: Not Installed
        Tornado: 4.2.1
            ZMQ: 4.1.4

System Versions:
           dist: centos 7.6.1810 Core
         locale: UTF-8
        machine: x86_64
        release: 3.10.0-957.5.1.el7.x86_64
         system: Linux
        version: CentOS Linux 7.6.1810 Core

Ch3LL:
[...]

Issue #51489, opened 2019-02-04
Saltclass globbing breaks class expansion


a-a-abramov:
### Description of Issue/Question
Saltclass globbing introduced with commit a3c0158188 breaks saltclass. To be specific this line of code inside a recursive function `expand_classes_in_order`:
`classes_to_expand = expand_classes_glob(classes_to_expand, salt_data)`
calls function, where classes_to_expand pointer seems to be overwritten, so we have this picture during debug: 

### Setup
**nodes/node.yml:**
```
classes:
  - test.0-level-class.yml
```

**classes/0-level-class.yml:**
```
classes:
  - test.1-level-class
```

**classes/1-level-class.yml:**
```
classes:
  - test.2-level-class
```

**classes/2-level-class.yml:**
```
classes:
  - test.3-level-class
```

**classes/3-level-class.yml:**
```
pillars:
```

### Steps to Reproduce Issue
Get pillars for a `node` described in saltclass with several (>1) levels of nested classes.
Expected behavior:
```
node:
    ----------
    __saltclass__:
        ----------
        classes:
            - test.3-level-class
            - test.2-level-class
            - test.1-level-class
            - test.0-level-class
```

Actual behavior:
```
node:
    ----------
    __saltclass__:
        ----------
        classes:
            - test.1-level-class
            - test.0-level-class
```

All classes below second level are lost due to the fact that `classes_to_expand` pointer is overwritten in the middle of recursion somewhere inside `expand_classes_glob`. If we log `classes_to_expand` at the beginning and right before the return of function `expand_classes_in_order` we would get this:
```
Start: []
Start: [u'test.1-level-class', u'test.0-level-class']
Start: [u'test.2-level-class', u'test.1-level-class', u'test.0-level-class']
Start: [u'test.3-level-class', u'test.2-level-class', u'test.1-level-class', u'test.0-level-class']
Start: [u'test.3-level-class', u'test.2-level-class', u'test.1-level-class', u'test.0-level-class']
Return: [u'test.3-level-class', u'test.2-level-class', u'test.1-level-class', u'test.0-level-class']
Return: [u'test.3-level-class', u'test.2-level-class', u'test.1-level-class', u'test.0-level-class']
Return: [u'test.3-level-class', u'test.2-level-class', u'test.1-level-class', u'test.0-level-class']
Return: [u'test.2-level-class', u'test.1-level-class', u'test.0-level-class']
Return: [u'test.1-level-class', u'test.0-level-class']
```
[...]

Issue #50755, opened 2018-12-05
saltclass list override '^' is literal for single list


mrichar1:
### Description

When using the override marker (`^`) at the start of a list, this is translated as a literal `^` if there are no other lists to override.

### Setup/Steps to Reproduce Issue

For example, with a state set up to install all packages in a list:

`states/pkg.yml`
```
{% for pkg in pillar.get('pkgs', []) %}
pkg_{{pkg}}:
  pkg.installed:
    - name: {{pkg}}
{% endfor %}
```

`nodes/foo.yml`
```
classes:
  - a
  - b

states:
  - pkgs
```

`classes/a.yml`
```
pillars:
  pkgs:
    - one
    - two
```

`classes/b.yml`
```
pillars:
  pkgs:
    - ^
    - three
```

Works as expected:

```
$ salt 'foo' state.apply | grep updated
Comment: The following packages would be installed/updated: three
```

However, if only the the 'override list' is used:

`nodes/foo.yml`
```
classes:
  - b

states:
  - pkgs
```

We get:
```
$ salt 'foo' state.apply | grep updated
Comment: The following packages would be installed/updated: ^
Comment: The following packages would be installed/updated: three
```

### Versions Report

Salt Version:
           Salt: 2018.3.0
 
Dependency Versions:
           cffi: Not Installed
       cherrypy: unknown
       dateutil: 2.6.1
      docker-py: Not Installed
          gitdb: 2.0.3
      gitpython: 2.1.8
          ioflo: Not Installed
         Jinja2: 2.10
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: 1.0.7
   msgpack-pure: Not Installed
 msgpack-python: 0.5.6
   mysql-python: Not Installed
      pycparser: Not Installed
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: Not Installed
         Python: 2.7.15rc1 (default, Nov 12 2018, 14:31:15)
   python-gnupg: Not Installed
         PyYAML: 3.12
          PyZMQ: 16.0.2
           RAET: Not Installed
          smmap: 2.0.3
        timelib: Not Installed
        Tornado: 4.5.3
            ZMQ: 4.2.5
 
System Versions:
           dist: Ubuntu 18.04 bionic
         locale: UTF-8
        machine: x86_64
        release: 4.15.0-39-generic
         system: Linux
        version: Ubuntu 18.04 bionic

[...]

Issue #50265, opened 2018-10-27
saltclass pillar not available to jinja


ReubenM:
/etc/salt/master
```
ext_pillar_first: True
```

saltclass/nodes/node1.yml
```
pillars:
  foo:
    - ^
    - bar
    - baz
```

pillar/top.sls
```
base:
  'I@foo':
    {%- for x in salt['pillar.get']('foo') %}
    - subdir.{{ x }}
    {%- endfor %}
```

The 'I@foo' pillar targeting works fine.
However, the pillar object made available to jinja when preprocessing the template is empty.

Salt: 2018.3.2
Python: 2.7.5

rbthomp:
@ReubenM I believe you have to use ` {%- for x in salt['__pillars__.get']('foo') %}`

https://github.com/saltstack/salt/blob/033ce00ad0259fd286665b4a19740121cf247135/salt/tops/saltclass.py#L32

rbthomp:
@ ReubenM are you trying to merge saltclass pillars and standard top style pillars? Both are completely different top and pillar generating systems. Maybe you can clarify on the use case for defining pillars for `foo` by saltclass and standard pillar top.sls.

garethgreenaway:
@ReubenM Thanks for the report.
@saltstack/team-core Does anyone whose more familiar with saltclass have any ideas here?

ReubenM:
@rbthomp:

Yes, I'm merging pillar from two sources. The first pass is from saltclass. The second pass using the pillar top sets additional pillar data based on the first pass. The pillar targeting within the pillar top file is working. However using pillar within the Jinja template processing does not. 

The real world use case for it is that I want to define roles using pillar, but I also want to assign pillar data based upon those roles. saltclass makes that possible. I have a large collection of nodes being used for perf testing, and their roles can change quite a bit. Modifying roles using grains on each minion is tedious and time consuming. Use pillar to define roles keeps role assignments all centralized to the master node.

I can do it entirely through pillar matching in the pillar top file instead of using jinja to iterate over sets, but it can make for an unwieldy pillar top file.

rbthomp:
@ReubenM I'm using roles as your are to avoid setting grains on nodes, except I'm keeping everything within saltclass instead of merging. This way you don't have to maintain top files. You can just define role classes that contain the pillars you need. You can even include classes in classes. 

saltclass/nodes/node1.yml
```
classes:
  - role.role1
  - role.role2
```
saltclass/classes/role/role1.yml
```
[...]

PR #52407, opened 2019-04-03
Large refactoring of SaltClass

a-a-abramov:
### What does this PR do?
Major refactor and bugfix of saltclass ext_pillar/master_tops utility.

Almost each one, that could be found by "saltclass" keyword in issues search. Previous discussions are here #51487 and here #51488

### Tests written?

Yes

### Commits signed with GPG?

Yes


a-a-abramov:
Hey everyone! 
I believe this should be a working saltclass. Please don't hesitate to point on my errors - I'll fix it as soon as possible. 
@rbthomp @dwoz @mgomersbach @mchugh19 and others interested.

a-a-abramov:
Next things to do:
- fix bugs you find :)
- update documentation a bit
- refactor find_and_process_re - didn't have time for this
- reintroduce globbing
- give an option to change class traverse mode in options (if class tree is constructed well and without hidden overlaps in pillar subtrees, than nobody will notice anything anyways)
- optional pretty output for __saltclass__ element in pillar.items (make it dict to look like tree, maybe?)
- saltclass matcher to address nodes in CLI by class names - have no idea if it's possible, but will look into it.

max-arnold:
Wow, this is great! I'll definitely carve some time next week to test the PR.

+1 for the matcher

a-a-abramov:
@max-arnold nice!
About CI - I looked through those 3-10M console outputs and didn't quite understand why they fail. It seems like saltclass is okay and it's other modules' issues (well, except linter checks).

mgomersbach:
This looks great already @a-a-abramov !

a-a-abramov:
Just to keep thread alive, here's the list of fixed issues:
#51817 and #49175 - fixed
#51489 - fixed by removal of globbing for now, but I'm working on globbing overhaul right now, so it's due today-tomorrow
#51360 - now you can access `__pillar__` dict from jinja in exactly the same manner you accessed `__grains__`, but with an important caveat - since you need pillars to render jinja while getting new pillars and changing old ones as the result of rendering, you can only access pillars from already rendered at that moment classes. It might sound a bit weird, but if you think more about this you would understand that it's classic chicken and egg problem. To clarify the method, here's a little example:
```
node:
  - A:
    - B
    - C:
      - D
  - E:
    - F
    - B
```
(implying node has classes A and E attached if it's not clear enough)
The order in which classes would be traversed is determined with [BFS](https://en.wikipedia.org/wiki/Breadth-first_search):
`node, A, E, B, C, F, (B), D`
Note, that second occurrence of B will not be processed. So what about `__pillar__` dict for jinja? In each class you can use pillars, which were processed before. For example in B we can use pillars from `node, A, E`, but not from `C, F, D`. I'd also like to add configuration option to choose traversing option - BFS or [DFS](https://en.wikipedia.org/wiki/Depth-first_search). For a well-constructed pillar tree it should not be an issue, but it's better to give user an option.
#50755 - fixed
#50265 - must be resolved, but could anyone check please? Unfortunately I don't have time to dive into integration test, but if the whole ext_pillar works as intended I did fix this.
#50262 - fixed
#48170 - 1 and 2 mentioned above and/or fixed before, surprisingly 3 works fine. 4 - didn't think about it, I suppose you're right about failing hard. 
#48167 - I've already addressed issue with unnecessary filesystem searches locally and it works fine, though I didn't notice any significant performance boost. But neither did I run stress tests or profilers, so everything is possible :)   
---
#50688, #50501 - didn't check yet
Several questions and feature requests - will look into those after this PR would be accepted, otherwise there's not point in doing so :)

rbthomp:
@a-a-abramov Thanks for the commit and addressing the many issues. I've found a small bug, we need to account for files that are blank. It looks like `render_yaml()` returns a nonetype when nothing is rendered. I added a little logic to `get_class()' to check for this.  Here is what I did to correct this.

```    c_rendered = ''

    if straight in l_files:
        c_rendered  =  render_yaml(straight, salt_data)
    elif  sub_straight in l_files:
        c_rendered = render_yaml(sub_straight, salt_data)
    elif  sub_init in l_files:
        c_rendered = render_yaml(sub_init, salt_data)
    else:
        log.warning('%s: Class definition not found', _class)
        return {}

    # Verify rendered yaml is not blank.
    if c_rendered:
        return c_rendered

    return {}
```
Probably need to perform the same check on node.yml files.

a-a-abramov:
Globbing is back! Though I need an additional approve from you on it since I changed a bit of how it works - will explain tomorrow.

a-a-abramov:
So about globbing. I tried to make it intuitive and I hope I succeeded. Basically it works like this:
- A.B.* will resolve into these:
  - A.B.C
  - A.B.C.D 
  - A.B if class A.B is defined with A/B/init.yml.

The last thing is questionable, but ny team decided that it's a lesser evil, because otherwise if you have number of files and folders under A/B all of them would be included EXCEPT init.yml. That's exactly the unintuitivity I wanted to avoid. What do you think?

- .X.Y in A.B will resolve into A.B.X.Y - nothing special

- mixture of both works exactly like mixture of both: .* will resolve into each "neighbour" class, "local" init.yml and all the subtree
- One speciality is . (single dot) includes local init.yml. If used in init.yml itself - safely ignored. 

Order of resolved glob is alphanumeric - I get absolute path for each class yml file, sort it and keep it ordered during resolution. My thought was that instead of implementation of some complex logic I'd better just keep it as simple and predictable as possible. Also I don't think that globbing for deep subtrees is a way to go for anyone.

a-a-abramov:
Could anyone tell me what to do next? @Ch3ll? Failed tests confuse me a bit since I can't understand if it's my fault or not. It's also worth to notice that implementation from all the releases saltclass was in is severely bugged and unusable under any circumstances.

This PR would be completed in a few days when I overcome my procrastination and finish user manual.

max-arnold:
Got some initial roadbumps during my testing (existing reclass-based state/pillar tree with obvious renames for saltclass):

```
% python --version
Python 2.7.15


% salt-call --local --id example.com pillar.data
[ERROR   ] Exception caught loading ext_pillar 'saltclass':
  File "/Users/user/.virtualenvs/test-salt/lib/python2.7/site-packages/salt/pillar/__init__.py", line 974, in ext_pillar
    key)
  File "/Users/user/.virtualenvs/test-salt/lib/python2.7/site-packages/salt/pillar/__init__.py", line 902, in _external_pillar_data
    *val)
  File "/Users/user/.virtualenvs/test-salt/lib/python2.7/site-packages/salt/pillar/saltclass.py", line 65, in ext_pillar
    return sc.get_pillars(minion_id, salt_data)
  File "/Users/user/.virtualenvs/test-salt/lib/python2.7/site-packages/salt/utils/saltclass.py", line 439, in get_pillars
    pillars, classes, states, environment = get_saltclass_data(node_data, salt_data)
  File "/Users/user/.virtualenvs/test-salt/lib/python2.7/site-packages/salt/utils/saltclass.py", line 300, in get_saltclass_data
    .format(dirpath, filename, dirpath, filename[:-4]))

[CRITICAL] Pillar render error: Failed to load ext_pillar saltclass: Conflict in class file structure - file ./reclass/classes/project/meta/example/test.yml and directory ./reclass/classes/project/meta/example/test.
```

The naming conflict above happens when there is a `test.yml` and `test` dir in the same directory. This worked just fine in reclass (if I remember correctly, `test.yml` is preferred over `test/init.yml`).

```
% python --version
Python 3.6.6

% salt-call --local --id example.com pillar.data
[ERROR   ] Failed to import pillar saltclass, this is due most likely to a syntax error:
  File "/Users/user/.virtualenvs/test-salt3/lib/python3.6/site-packages/salt/loader.py", line 1548, in _load_module
    mod = spec.loader.load_module()
  File "<frozen importlib._bootstrap_external>", line 399, in _check_name_wrapper
  File "<frozen importlib._bootstrap_external>", line 823, in load_module
  File "<frozen importlib._bootstrap_external>", line 682, in load_module
  File "<frozen importlib._bootstrap>", line 265, in _load_module_shim
  File "<frozen importlib._bootstrap>", line 684, in _load
  File "<frozen importlib._bootstrap>", line 665, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 678, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "/Users/user/.virtualenvs/test-salt3/lib/python3.6/site-packages/salt/pillar/saltclass.py", line 17, in <module>
    import salt.utils.saltclass as sc
  File "/Users/user/.virtualenvs/test-salt3/lib/python3.6/site-packages/salt/utils/saltclass.py", line 308
    salt_data['class_paths'] = OrderedDict(sorted(salt_data['class_paths'].iteritems(), key=lambda (k, v): k))
                                                                                                   ^
SyntaxError: invalid syntax
[CRITICAL] Specified ext_pillar interface saltclass is unavailable
'example.com' is not available.
```

On a positive note, refactored saltclass is able to detect duplicate keys:

```
parameters:
  network:
    interfaces:
    interfaces:
      eth0: 1.2.3.4

[ERROR   ] YAML rendering exception for file ./reclass/nodes/selectel/example.com.yml:
while constructing a mapping
  in "<unicode striding>", line 44, column 5:
        interfaces:
        ^
found conflicting ID 'interfaces'
  in "<unicode string>", line 45, column 5:
        interfaces:
        ^
[WARNING ] Unable to render yaml from ./reclass/nodes/selectel/example.com.yml
```

Tested against https://github.com/a-a-abramov/saltstack/commit/2b8da7608a37a59644e71e61bcaec91778510b7e

max-arnold:
> This PR would be completed in a few days when I overcome my procrastination and finish user manual.

You could probably borrow some useful bits from the release notes: https://docs.saltstack.com/en/latest/topics/releases/2018.3.0.html#new-saltclass-pillar-master-tops-modules

a-a-abramov:
@max-arnold
As you can probably see I raise conflict error intentionally. I think reclass'es model with priority of one file over the other doesn't really give you benefits but it's ambiguity could lead to unintuitive pillar rendering.I believe we should keep it simple and fail hard.

And for the next error - good point, I completely forgot to test the module against python3. Will do it asap.

max-arnold:
Performance tests (#48167):

### 2018.3.4:
```
% time salt-call --local --id example.com pillar.data > /dev/null
salt-call --local --id example.com pillar.data > /dev/null  2,52s user 1,30s system 91% cpu 4,155 total
```

Profiling data (started at `salt/utils/saltclass.py:get_pillars`):



### 2019.2@2b8da76:
```
% time salt-call --local --id example.com pillar.data > /dev/null
salt-call --local --id example.com pillar.data > /dev/null  1,58s user 0,74s system 87% cpu 2,647 total
```

Profiling data (started at `salt/utils/saltclass.py:get_pillars`):

2019.2@2b8da76

The second graph looks much more dense, with the majority of time spent in `yaml/composer.py:compose_node`.  Maybe there is some minor optimization potential in blue `_compile` box (bottom-left), related to regexp compilation.

Steps to reproduce:

```
% find reclass/classes -type f | wc -l
     285

% find reclass/nodes -type f | wc -l
      56

python -m cProfile -o salt.profile ~/.virtualenvs/test-salt/bin/salt-call --local --id example.com pillar.data
pyprof2calltree -i salt.profile -o salt.calltree
```

The resulting `salt.calltree` could be opened with `kcachegrind`, `qcachegrind` or anything from this list on a Mac: https://stackoverflow.com/questions/5426799/is-there-any-kcachegrind-alternative-for-mac-os-x-outta-there

Overall verdict: FIXED!

a-a-abramov:
Fixed 2to3 issue

a-a-abramov:
Fixed 2to3 issue

dwoz:
@a-a-abramov Great work here but this needs to go to the develop branch instead of 2019.2

max-arnold:
> Great work here but this needs to go to the develop branch instead of 2019.2

@dwoz This is a bugfix (albeit a large one). Saltclass is plagued by multiple problems from the moment it was merged in 2018.3 without proper QA (see the [list of issues](https://github.com/saltstack/salt/pull/52407#issuecomment-480890763) that is fixed by this PR).

Things got even worse in 2019.2.0 - Saltclass is completely broken and [doesn't work at all](#51487) for many people, plus the original author is unresponsive for several months:

2019-04-11 at 17 47

It is unpractical to fix these bugs one-by-one because the implementation has too many defects and it is much simpler to just rewrite it.

By rebasing this PR on top of the `develop` branch, you are effectively delaying those necessary fixes until the Neon release, which is not yet scheduled and won't be shipped in several months.  This leaves Saltclass implementation in a broken state.

Since it is hard to break backward compatibility for the already broken feature, and the feature is experimental (it doesn't even have proper docs and only mentioned in 2018.3 release notes), please consider reverting your decision and allow us to rectify this mess in the nearest bugfix release (2019.2.1)

dwoz:
@max-arnold Thanks for explaining this more. We'll still need to make sure the linter, docs, and PR tests all pass to get this into 2019.2.1.

max-arnold:
@dwoz Thanks for understanding!

How much time do we have to finalize the PR before 2019.2.1 is branched?

mgomersbach:
Can someone decypher the lint error? Because I seem to be reading over it.

mchugh19:
@mgomersbach looks like a bad Jenkins job config. 
```
Error when executing always post condition:
java.lang.IllegalArgumentException: Failed to prepare cleanWs step
```
If anyone has changed Jenkins job steps recently, they might need to test out the lint job.

max-arnold:
QA progress so far:

* FIXED: #48434 Crash while applying state to new minion with saltclass enabled (empty node.yml)
* FIXED: #50262 Saltclass: broken list expansion (2019.2.0)
* FIXED w/issues (no hardfail): #48170 Multiple issues with Saltclass variable expansion (py27 fixed; crashes.zip fixed; multiple expansions on a single line - fixed; undefined variables are still silently ignored)

More to follow. Also, found a few new issues, need more time to reproduce.

a-a-abramov:
@max-arnold 
You mean we must hardfail if we fail to find pillar `xx:yy:zz` during `${xx:yy:zz}` expansion? If that's the case I'll fix it easily. But in #48170 you specifically mentioned type mismatch - that would be hard (if not impossible) to fix. 
Say, you have `salt.pillar.get('xx:yy:zz', {}).get('smth')` or similar expression in your state. If `xx:yy:zz` is not a dict but str, you would get exactly the same jinja error if I'm not mistaken. 
But I think I got your point - fail hard if expression can't be expanded.

max-arnold:
My point was related to just typos in parameter names that are silently ignored. Also, hardfail on missing parameters would be good for CI purposes (does top/pillar actually render at all?).

a-a-abramov:
@max-arnold
Ok, reasonable. Will do it this week alongside documentation draft.

@dwoz
What's our deadline to be included into the upcoming release? If I had time till the end of the month I'd definitively add a matcher - It should be pretty easy since we create a __saltclass__:classes pillar.

a-a-abramov:
> FIXED w/issues (**no hardfail**): #48170 Multiple issues with Saltclass variable expansion (py27 fixed; crashes.zip fixed; multiple expansions on a single line - fixed; **undefined variables are still silently ignored**)

fixed

max-arnold:
> > FIXED w/issues (**no hardfail**): #48170 Multiple issues with Saltclass variable expansion (py27 fixed; crashes.zip fixed; multiple expansions on a single line - fixed; **undefined variables are still silently ignored**)
> 
> fixed

Below is a case that doesn't work for me:

```
pillars:
  network:
    interfaces:
      eth0:
        ipaddr: 1.1.1.1
  rules:
    - '-A PREROUTING -s 1.2.3.4/32 -d ${network:interfaces:eth0:ipaddrXYZ} -p tcp --dport 1234 5.6.7.8:1234'
```

```
local:
    ----------
    __saltclass__:
        ----------
        classes:
        environment:
        nodename:
            example.com
        states:
    network:
        ----------
        interfaces:
            ----------
            eth0:
                ----------
                ipaddr:
                    1.1.1.1
    rules:
        - -A PREROUTING -s 1.2.3.4/32 -d ${network:interfaces:eth0:ipaddrXYZ} -p tcp --dport 1234 5.6.7.8:1234
```

If you remove the `XYZ` suffix from the expression then it expands properly:

```
   rules:
        - -A PREROUTING -s 1.2.3.4/32 -d 1.1.1.1 -p tcp --dport 1234 5.6.7.8:1234
```

max-arnold:
@ReubenM Do you mind double checking if #50265 (saltclass pillar not available to jinja) is fixed by this PR?  I'm getting errors while trying to run your example:

```
[CRITICAL] Pillar render error: Error encountered while rendering pillar top file.
[ERROR   ] Got insufficient arguments for pillar match statement from master
```

max-arnold:
FIXED #50755 saltclass list override '^' is literal for single list
FIXED #51489 Saltclass globbing breaks class expansion (**tested very briefly**)
FIXED #51817 Saltclass external pillar broken in 2019.2.0

max-arnold:
FIXED (**w/issues**) #50755 saltclass list override '^' is literal for single list

The following example fails:

```
# node.yml
classes:
  - class2

# class1.yml
pillars:
  list:
    - foo
    - bar

# class2.yml
classes:
  - class1
pillars:
  list: []
```

```
  File "/Users/users/.virtualenvs/test-salt3/lib/python3.6/site-packages/salt/utils/saltclass.py", line 68, in _dict_merge
    if m_target[key][0] == '^':

[CRITICAL] Pillar render error: Failed to load ext_pillar saltclass: list index out of range
```

But if you remove the `list` key from  the `class1.yml`, then it works as expected.

max-arnold:
#51360 Pillar in jinja

The first issue is that Jinja variable is named `__pillar__` instead of `pillar` as in other parts of Salt (see https://docs.saltstack.com/en/latest/topics/tutorials/pillar.html#parameterizing-states-with-pillar)

Second, I found a bunch of strange edge cases:

```
# node.yml:
classes:
  - class1
  - class2
pillars:
  foo: node_foo

# class1.yml:
pillars:
  foo: foo_class1

# class2.yml:
pillars:
  foo: foo_class2
  bar: {{ __pillar__['foo'] }}
```

And the resulting `pillar.data` is puzzling to me:

```
local:
    ----------
    __saltclass__:
        ----------
        classes:
            - class2
            - class1
        environment:
        nodename:
            example.com
        states:
    bar:
        node_foo
    foo:
        node_foo
```

It obviously can't work at all in node files (because jinja is rendered before yaml):

```
# node.yml
pillars:
  foo: {{ __pillar__['bar'] }}
```

This one triggers an exception:

```
# node.yml
classes:
  - class1

# class1.yml
states:
  - {{ __pillar__['blah'] }}
```

```
File "/Users/user/.virtualenvs/test-salt3/lib/python3.6/site-packages/salt/utils/saltclass.py", line 384, in get_saltclass_data
    if state not in ordered_state_list and state.find('^') == -1:

[CRITICAL] Pillar render error: Failed to load ext_pillar saltclass: 'NoneType' object has no attribute 'find'
```

But this one doesn't:

```
# node.yml
states:
  - {{ __pillar__['blah'] }}
```

Also, in the last snippet the nonexistent `blah` key produces 'None', although it is much better to fail (as Salt does with regular pillar):

```
{{ pillar['blah'] }}

[ERROR   ] Unable to render top file: Jinja variable 'salt.utils.context.NamespacedDictWrapper object' has no attribute 'blah'
```

To be honest, I think it would be better to exclude this feature (at least for now). It is complicated, unintuitive, and has lots of corner cases (you drew a whole inheritance diagram to explain it!). It is highly likely, that once documented it will lead to several Github issues submitted by users. Basically it opens a new can of worms.

Also, it looks like the person who requested this feature found a workaround: https://github.com/saltstack/salt/issues/51360#issuecomment-474118481

Overall, in my experience, the standard reclass/saltclass parameter expansion is flexible enough to cover the majority of cases when I needed to reference previously defined pillar values.

ReubenM:
Using pillar data from salt-class inside another pillar module (in my case using jinja) is a mess of ugly hackery.

To step back for a moment and view this all from a higher level, the reason behind my use of it is that I'm using pillar data to define roles. But I also need to add other pillar data, the content of which is dependent on the assigned role. In my case a role is a tree structure containing multiple role matches anywhere in that tree; and I want to be able to dynamically manage those roles from pillar on the salt master rather than using grains. Using grains to define roles is much less flexible for what I'm doing. (I'm running multiple perf test environments using salt to orchestrate the testing, and roles structures are always changing)

I would honestly recommend not supporting / only offering limited support for using pillar data from one module within another. A better solution would be to be able to support **multi-pass pillar generation**.

Three approaches to this could be: 
- Building multipass into the salt-class module. (Pro: less invasive changes, Con: limited to salt-class module. Yaml files could end up being more complex)
- Building pillar multi-pass into salt. (Pro: you could have multiple instances of salt-class directories and control the order they are evaluated. You could also apply the multi-pass to mix and match multiple instances of other pillar module types. Con: Possibly more invasive code changes into salt. Multi-pass for other pillar module types might not be the most intuitive thing for an end user to configure, and for some modules may just not make sense at all)
- Define a "Roles" data structure that is managed entirely separated from Pillar. (Pros: minimal changes to pillar modules. Cons: brittle design and only slightly more flexible than what is currently possible with just Pillar.)

I would recommend the second option I listed. For example it would be nice to be able to have pillar data built using "salt-class instance 1" -> "etcd pillar module" -> "salt-class instance 2" where pillar is fully evaluated at each step and presented to the module in the next step. Supposedly that is what you should be able to do now, but I don't think there is any way to run multiple instances of the same pillar module.

All this is outside the scope of this fix-up for salt-class. But if this is something that might be considered for a future release , it would allow dropping support for some of the more complex corner cases in exchange for a better design coming down the pipeline in a future release. You don't really loose anything because most of these corner cases haven't ever worked / worked as expected anyway.

max-arnold:
@ReubenM I think I understand your point. Roles/grouping and data inheritance/overriding is a long-term issue in Salt (#2466 and #3991 are just two examples that were created back in 2012-2013), which led to proliferation of different pillar/top implementations (and also the TOFS pattern). Unfortunately, none of these approaches turned into a "blessed" standard in Salt.

Still, this doesn't answer the specific question:

> #50265 - must be resolved, but could anyone check please? Unfortunately I don't have time to dive into integration test, but if the whole ext_pillar works as intended I did fix this.

We need your input on this issue to move forward (even if you just say that it is no longer relevant).

max-arnold:
@a-a-abramov The 2019.2.1 has branched just a few hours ago: https://groups.google.com/forum/#!topic/salt-users/WWJ4hm9U8lk

I still have 2 blocking issues to reproduce (need to strip down a large reclass state tree), but I think I can do that this week. The issues are:

1. Not all expressions are expanded
2. Class ordering that you implemented is a showstopper for us to migrate from reclass. I'll try to provide a compelling example

ReubenM:
@max-arnold I won't be able to test it till sometime next week.

It's no longer directly relevant to me, I made changes to work around it. But it's still something that would be nice to have working.

a-a-abramov:
Wow that's a ton of feedback to digest. I'll fix minor bugs and cornercases today or on Sunday. 
@max-arnold could you elaborate on this?
> * Not all expressions are expanded
> 
> * Class ordering that you implemented is a showstopper for us to migrate from reclass. I'll try to provide a compelling example

As for class ordering - I really can't understand why it's so puzzling for you. But I guess that could easily be the consequence of sheer amount of time I spent over the code... I'll review it once again with my colleagues to figure out if it's intuitive or not for them.

max-arnold:
> could you elaborate on this?

Sure, I just need some time to prepare reproducible test cases

max-arnold:
> Not all expressions are expanded

Below is a minimal reproducible test case (may be the same as in https://github.com/saltstack/salt/pull/52407#issuecomment-484035578):

```
pillars:
  key1: abc
  key2: def
  key3: ${key1}${key2}
```

```
local:
    ----------
    __saltclass__:
        ----------
        classes:
        environment:
        nodename:
            example.com
        states:
    key1:
        abc
    key2:
        def
    key3:
        abc${key2}
```

max-arnold:
Ordering issues:

```
# node.yml
classes:
  - network
  - base

# network.yml
pillars:
  key2: 2
states:
  - network

# base.yml
pillars:
  key1: 1
states:
  - baseline
```

```
% salt-call --local --id example.com pillar.data
local:
    ----------
    __saltclass__:
        ----------
        classes:
            - base
            - network
        environment:
        nodename:
            example.com
        states:
            - baseline
            - network
    key1:
        1
    key2:
        2
```

The problem here is that network (low-level settings) should be configured **before** the baseline (the baseline class assumes that network is already configured). In `reclass` the order of both classes and states is preserved as is:


```
local:
    ----------
    __reclass__:
        ----------
        applications:
            - network
            - baseline
        classes:
            - network
            - base
        environment:
            base
        nodename:
            example.com
    key1:
        1
    key2:
        2
```

I believe this is quite important (and is a showstopper for us to migrate from reclass). Some rationale:

1. This is how Salt behaves by default. Since 0.17.0 it preserves specified state ordering (instead of using lexicographic sorting https://docs.saltstack.com/en/latest/ref/states/ordering.html#state-auto-ordering)
2. This is how SaltClass used to work in 2018.3.x:

```
local:
    ----------
    __saltclass__:
        ----------
        classes:
            - network
            - base
        environment:
        nodename:
            example.com
        states:
            - network
            - baseline
    key1:
        1
    key2:
        2
```

3. IMO, the lexicographic class/state ordering in SaltClass would lead to tight coupling between states, because the only way to specify dependencies would be to use state requsites (different states/formulas would need to have references to each other). This negates much of the layering/inheritance benefits of SaltClass.

max-arnold:
And the final tests:

FIXED (w/ ordering issues): #49175 SaltClass duplicate entries in merged list
FIXED (w/ ordering issues): #51817 Saltclass external pillar broken in 2019.2.0

There are no duplicates (and thus both issues are fixed), but the ordering issues are also visible here. In the #49175 the expected result is:

```
data:
            - default 1
            - default 2
            - child 1
            - child 2
```


The actual behavior is different:

```
    data:
        - child 1
        - child 2
        - default 1
        - default 2
```

This is yet another argument to preserve the class/state ordering.

max-arnold:
So far I'm done with the QA process. To summarize the long thread:

* The most pressing issues are fixed
* There are two critical expansion issues: https://github.com/saltstack/salt/pull/52407#issuecomment-485239602 https://github.com/saltstack/salt/pull/52407#issuecomment-485239602 
* The list merging issue: https://github.com/saltstack/salt/pull/52407#issuecomment-484086058
* The class/state ordering issue (I consider it to be critical and also a regression from 2018.3.x): https://github.com/saltstack/salt/pull/52407#issuecomment-485240912  https://github.com/saltstack/salt/pull/52407#issuecomment-485241449
* A questionable new feature that introduces more problems that it solves: https://github.com/saltstack/salt/pull/52407#issuecomment-484134939. I'm -1 on this, but it is not the thing I would strongly object against.

I haven't tested #50688 and #50501 (maybe @ptitdoc or @cheribral could help here?). Also there are some feature requests that aren't mentioned here (#49493, #49590, #49599), but it is better to consider them after this PR gets merged.

a-a-abramov:
Omg what a silly mistake I made... There's nothing wrong with class ordering in algorithm itself - it's just the code that builds initial classes queue from node yaml. I messed up ordering just there ><
And since we have a rule to keep node yamls as empty as possible (ideally - single class) I missed it in tests. Great job finding it!
Unfortunately I didn't have spare time on Sunday to fix anything, so patch with multiple fixes is coming tomorrow. 
I was also thinking abound jinja-pillar problem with strangely looking results and I think we should just state in docs that using pillar dict in jinja is experimental, it is intended to give additional flexibility to those looking for it, but with some natural limitations - _examples and explanations I gave earlier_ - it's not intended to use instead of ${notation} but to add states based on pillar information etc etc etc.

Likewise I'd like to state that wildcards are experimental too since I implemented alphanumeric ordering for resolved list - one could use it on nested subtrees, but at their own risk. It could easily confuse people so I think it's better to warn them beforehand.

Pillar expansion issues are the hardest ones for me - sadly I'm still mostly unfamiliar with that part of code. I'll try my best to get rid of them though.

a-a-abramov:
> * There are two critical expansion issues: [#52407 (comment)](https://github.com/saltstack/salt/pull/52407#issuecomment-485239602) [#52407 (comment)](https://github.com/saltstack/salt/pull/52407#issuecomment-485239602)
both fixed

> * The list merging issue: [#52407 (comment)](https://github.com/saltstack/salt/pull/52407#issuecomment-484086058)
fixed

> * The class/state ordering issue (I consider it to be critical and also a regression from 2018.3.x): [#52407 (comment)](https://github.com/saltstack/salt/pull/52407#issuecomment-485240912)  [#52407 (comment)](https://github.com/saltstack/salt/pull/52407#issuecomment-485241449)
fixed

a-a-abramov:
Things I forgot:
  * look into this
```
# node.yml 
states: 
  - {{ __pillar__['blah'] }} 
```
  * rename __param__ into param - it bothers me a lot too

max-arnold:
> could you measure execution time against your large tree again? I've finally rewritten variable expansion part and it was... well... not very well optimized.

Previous revision (py2.7):

```
salt-call --local --id example.com pillar.data > /dev/null  1,62s user 0,80s system 87% cpu 2,776 total
```
prev


Last revision (py2.7):

```
salt-call --local --id example.com pillar.data > /dev/null  1,57s user 0,79s system 87% cpu 2,693 total
```
last

It is definitely faster, although not by a large % of the total run time. At least the `_expand_variables` and `_dict_search_and_replace` boxes on the diagram take much less space now.

max-arnold:
Also it is possible to shave off a few milliseconds by precompiling the regex:

```diff
--- a/salt/utils/saltclass.py	2019-04-25 11:59:31.000000000 +0700
+++ b/salt/utils/saltclass.py	2019-04-25 11:56:59.000000000 +0700
@@ -21,6 +21,9 @@
 log = logging.getLogger(__name__)


+VARIABLE_RE = re.compile(r'(\\)?\${.*?}')
+
+
 def _render_jinja(_file, salt_data):
     j_env = Environment(loader=FileSystemLoader(os.path.dirname(_file)))
     j_env.globals.update({
@@ -108,7 +111,7 @@
     :param escaped: should we match \${escaped:reference} or ${not}
     :return: set of matched substrings from pillar
     '''
-    matches_iter = re.finditer(r'(\\)?\${.*?}', six.text_type(text_pillar))
+    matches_iter = VARIABLE_RE.finditer(six.text_type(text_pillar))
     result = set()
     if not matches_iter:
         pass
```

a-a-abramov:
As far as I know python itself compiles regexp and puts it into internal cache for future usage. So I doubt it would make a difference.

max-arnold:
@a-a-abramov I'm still seeing state ordering issues and I hope to come up with another minimal reproducible test case.

a-a-abramov:
Also I forgot to mention yesterday - now you can create large nested pillar structures with variables. ${} now may be expanded into each of dict, list and string (dict was unsupported in previous versions). Expanded pillar value may contain other ${} which would be expanded too. Right now number of expansion iterations is limited to avoid infinite cycles. I've set this number to 5 but will move this parameter into opts later if someone somehow would reach this level of nesting.

max-arnold:
@a-a-abramov It took a while to figure out, but I believe that this is the root of the ordering issues I observe:

A minimal reproducible test case:

```
# example.com.yml
classes:
  - baseline
  - internal

# baseline.yml
classes:
  - minimal

# init.yml
pillars:
  services:
    - service1

# internal.yml
classes:
  - init
pillars:
  services:
    - service3

# minimal.yml
classes:
  - init
pillars:
  services:
    - service2
```

This is the output of `salt-call --local --id example.com pillar.data`:

```
local:
    ----------
    __saltclass__:
        ----------
        classes:
            - baseline
            - minimal
            - internal
            - init
        environment:
        nodename:
            example.com
        states:
    services:
        - service2
        - service1
        - service3
```

When I change my node definition to this (it should be equivalent, because `baseline` is just a reference to `minimal`):

```
classes:
  - minimal
  - internal
```

I get this (note the different service order):

```
local:
    ----------
    __saltclass__:
        ----------
        classes:
            - minimal
            - internal
            - init
        environment:
        nodename:
            example.com
        states:
    services:
        - service1
        - service2
        - service3
```

This alone looks suspicious, but I believe that the class ordering also should be different. This is a data merging strategy in `reclass`: https://reclass.pantsfullofunix.net/operations.html#data-merging

And this is how `reclass` renders the initial (unmodified) test case:

```
local:
    ----------
    __reclass__:
        ----------
        applications:
        classes:
            - init
            - minimal
            - baseline
            - internal
        environment:
            base
        nodename:
            example.com
    services:
        - service1
        - service2
        - service3
```

I believe that this class order is the only one that precisely reflects the inheritance tree:

```
classes:
            - init
            - minimal
            - baseline
            - internal
```

Let's look again to the node definition:

```
classes:
  - baseline
  - internal
```

1. `baseline` inherits from `minimal`, and `minimal` from `init`
2. `internal` inherits from `init`

This results in the following class order (from less specific to more specific):

1. `init` > `minimal` > `baseline`
2. `init` > 'internal`

Merging and deduplication leads to the order which is produced by `reclass`:

```
classes:
            - init
            - minimal
            - baseline
            - internal
```

a-a-abramov:
@max-arnold if understand you correctly this class tree
```
G:
  D:
    A
    B
    C
  E
  F
N:
  H
  I
  M:
    J
    K
    L
U:
  O
  P
  T:
    Q
    R
    S
``` 
would produce list pillar `[A,B,C,D,E,F,G,H,I,J,K,L,M,N,O,P,Q,R,S,T,U]` and the same states list assuming each class has definition like this:
```
# G.yml
pillars:
  list:
    - G
states:
  - G
classes:
  - D
  - E
  - F

# A
pillars:
  list:
    - A
states:
  - A
```

max-arnold:
@a-a-abramov I'm sorry, but I have hard time understanding your example. I can't grasp this diagram:

```
G:
  D:
    A
    B
    C
  E
  F
...
```

Is `G:` a node? What are `A`, `B` and `C`? What nesting means - less specific or more specific level (closer to a node)?

a-a-abramov:
```
# node.yml
classes:
  - G
  - N
  - U

# A.yml
pillars:
  list:
    - A
states:
  - A

# B.yml
pillars:
  list:
    - B
states:
  - B

# C.yml
pillars:
  list:
    - C
states:
  - C

# D.yml
pillars:
  list:
    - D
states:
  - D
classes:
  - A
  - B
  - C

# E.yml
pillars:
  list:
    - E
states:
  - E

# F.yml
pillars:
  list:
    - F
states:
  - F

# G.yml
pillars:
  list:
    - G
states:
  - G
classes:
  - D
  - E
  - F

...

```

max-arnold:
@a-a-abramov 

[saltclass.zip](https://github.com/saltstack/salt/files/3116864/saltclass.zip)

This is your example classes, that are rendered by `reclass` (see the `reclass_pillar_data.yml`). The states and pillar lists are ordered exactly as you said above.

Classes are ordered according to their position in the hierarchy:

```
        classes:
            - A
            - B
            - C
            - D
            - E
            - F
            - J
            - K
            - L
            - H
            - I
            - M
            - Q
            - R
            - S
            - O
            - P
            - T
            - G
            - N
            - U
```

max-arnold:
```
diff --git a/salt/utils/saltclass.py b/salt/utils/saltclass.py
index 616cec0a8c..3d80e3110c 100644
--- a/salt/utils/saltclass.py
+++ b/salt/utils/saltclass.py
@@ -10,7 +10,6 @@ from jinja2 import FileSystemLoader, Environment
 import salt.utils.path
 import salt.utils.yaml
 from salt.exceptions import SaltException
-from yaml.error import YAMLError

 # Import 3rd-party libs
 from salt.ext import six
@@ -38,7 +37,7 @@ def _render_yaml(_file, salt_data):
     result = None
     try:
         result = salt.utils.yaml.safe_load(_render_jinja(_file, salt_data))
-    except YAMLError as e:
+    except salt.utils.yaml.YAMLError as e:
         log.error('YAML rendering exception for file %s:\n%s', _file, e)
     if result is None:
         log.warning('Unable to render yaml from %s', _file)
```

max-arnold:
It would be safer if saltclass was able to prevent duplicate node definitions in the same way as reclass:

```
local:
    ----------
    _errors:
        - Failed to load ext_pillar reclass: ext_pillar.reclass: yaml_fs: Definition of node 'node' in '/tmp/salt/reclass/nodes/datacenter1/node.yml' collides with definition in '/tmp/salt/reclass/nodes/node.yml'. Nodes can only be defined once per inventory.
```

a-a-abramov:
Just FYI - I've been on vacation past weeks, but now I'm back and will finish work on this PR very soon. A few days before my departure me and @max-arnold managed to agree on behavior and eliminate several corner cases. Won't make it before 2019.2.1 though, so I'd aim for the next release.

ptitdoc:
Hello,
I'm just checking this pull request and there is something I don't understand well.
You explain the hierarchy that allow to expand the current generated pillar using ${} or {% __pillar__ %}. I thing I understand, but what about the nodes vs classes.
If I lookup the pillar inside a classes file, I can only see the node pillar data, not the pillar data loaded from other classes.

max-arnold:
@ptitdoc 

> but what about the nodes vs classes.
> If I lookup the pillar inside a classes file, I can only see the node pillar data, not the pillar data loaded from other classes.

Pillar data from other classes is merged automatically, in according with the inheritance order.

ptitdoc:
Merged automatically meaning that I cannot access to it programatically through `__pillar__` or expansion ?

ptitdoc:
Here my use case. This is the hierarchy of files I try to follow:
```
classes:
  domains:
    mydomain.local:
      init.yml
      hosts.yml
      ca.yml
  baseline:
    init.yml
    ssh.yml
    ntp.yml
  roles:
    cloud.yml
    reverseproxy.yml
    caroot.yml
    dns.yml

nodes:
  test:
    dns.test.mydomain.local.yml
  prod:
    dns.mydomain.local.yml
    master.mydomain.local.yml
```
One usage of this hierarchy is to have repository of my systems per domain (in hosts.yml). I would like to use the content of this hosts.yml file either to build a name server configuration or a cloud deployment map. So my initial through was to create a role file such as dns.yml or cloud.yml that run some jinja to construct a pillar that contains the right structure for my salt states. However inside `roles/cloud.yml` I do not seem to be able to expand pillar data from `hosts.yml`, and inside `master.mydomain.local.yml` I do not seem to be able to expand pillra data from `hosts.yml`.

It apparently only works the other way around. Meaning that I can expend the node pillar from classes. Is it normal? Or I'm not using the proper way to do this?

max-arnold:
Variable expansion should work across classes just fine (the same way as it worked in [reclass](https://reclass.pantsfullofunix.net/operations.html#parameter-interpolation)).  `__pillar__` lookups are more tricky - Andrey explained the details here: https://github.com/saltstack/salt/pull/52407#issuecomment-480890763

ptitdoc:
You are right, expansion works properly between classes. It is only `__pillar__` that does not work as expected (at least not in my use case). `__pillar__` apparently only contains node data, not classes data.

ptitdoc:
Here is a real test case for reference, as it helps a lot understanding the situation:
classes/domains/hosts.yml:
```
pillars:
  pillar_inhosts: {{ __pillar__ | tojson }}
  datahosts: blablabla
```

classes/roles/cloud.yml:
```
pillars:
  clouddata: blablabla
  pillar_incloud:
    {{ __pillar__|tojson }}
```

nodes/test/master.test.mydomain.local:
```
classes:
  - domains.hosts
  - roles.cloud

pillars:
  pillar_innode: {{ __pillar__|tojson }}
  datanode: blablabla
```

$ salt master.test.mydomain.local
```
master.test.mydomain.local:
    ----------
    __saltclass__:
        ----------
        classes:
            - domains.hosts
            - roles.cloud
        environment:
        nodename:
            master.test.mydomain.local
        states:
    clouddata:
        blablabla
    datahosts:
        blablabla
    datanode:
        blablabla

    pillar_incloud:
        ----------
        datanode:
            blablabla
        pillar_innode:
            ----------
    pillar_inhosts:
        ----------
        clouddata:
            blablabla
        datanode:
            blablabla
        pillar_incloud:
            ----------
            datanode:
                blablabla
            pillar_innode:
                ----------
        pillar_innode:
            ----------
    pillar_innode:
        ----------
```

If I follow Andrey logic, my pillar generation hierarchy is:
```
node (master.test.mydomain.local)
  domains.hosts
  roles.cloud
```

node should see himself
In domains.hosts I should be able to see node
In roles.cloud I should be able to see node and domains.hosts

But the result is the other way around.

a-a-abramov:
Thank you for a feedback @ptitdoc, I appreciate it!
Everything works as intended in your case! And apparently that's the reason why we should ditch `__pillar__` lookups completely :) (I'll elaborate on it in the next comment)
The reason you can't access _node pillars_ from _node yaml_ is that in order to load information into memory you first have to render yaml. And to render yaml you have to render all the jinja in it. So your `{{ __pillar__|tojson }}` is rendered before `__pillar__` dict has any info in it.

max-arnold:
> And apparently that's the reason why we should ditch __pillar__ lookups completely

🎉

a-a-abramov:
Now about expansions.
@max-arnold changed my mind about `__pillar__` lookups. Now I'm pretty sure they are bad and should not be used - [reclass](https://reclass.pantsfullofunix.net/operations.html#parameter-interpolation) pillar expansion is more convenient, unambiguous and clear way to use pillars is pillars. 

Originally I needed the ability to specify states using pillars. For example 
```
states:
  - A
  - B
{% if __pillar__.get('X', False) %}
  - Y
{% endif %}
```
Simple `top.sls` has this feature, why should we make a step back? So I decided to fix original pillar merging code for this to work. But much (*much*) time later code was still erroneous, became cryptic, and it turned out backward compatibility with reclass was broken.

So now I think that we should ditch this idea of using `__pillar__` through jinja and implement similar mechanism to "inject" pillar into states like this:
```
states:
  - A
  - B
  - ${optional:pillar}
```
somewhere else:
```
pillars:
  optional:
    pillar: X
```
OR
```
pillars:
  optional:
    pillar:
      - Y
      - Z
```
so we would either get `states: [A, B, X]` either `states: [A, B, Y, Z]`. I'm almost done implementing this.

ReubenM:
The thing that sucks about the injector method compared to jinja, is that you can't use it to set key names. There is value to using __pillar__ in jinja when you use another ext_pillar that is evaluated before saltclass is evaluated. In that type of workflow you can set key names using pillar data in jinja.

That said, I don't think it's sane to try to support using __pillar__ data derived from saltclass, unless there is devised a means for saltclass to have multiple ordered instances so that an instance can utilize __pillar__ data from the evaluations of the preceding instances.

a-a-abramov:
> That said, I don't think it's sane to try to support using __pillar__ data derived from saltclass, unless there is devised a means for saltclass to have multiple ordered instances so that an instance can utilize __pillar__ data from the evaluations of the preceding instances.

Can't quite understand the part about "multiple ordered instances". Because right now pillars are merged class-after-class so you could use intermediate results of pillar merging to render next classes in order. And it's the exact thing that brings a ton of confusion.

ReubenM:
> Can't quite understand the part about "multiple ordered instances". Because right now pillars are merged class-after-class so you could use intermediate results of pillar merging to render next classes in order. And it's the exact thing that brings a ton of confusion.

Hmm, apparently I need to dig deeper into how this actually works. For some reason I was thinking that the yaml dictionary content from classes were merged together and then the new pillar data was evaluated from the merged yaml.

a-a-abramov:
> Hmm, apparently I need to dig deeper into how this actually works. For some reason I was thinking that the yaml dictionary content from classes were merged together and then the new pillar data was evaluated from the merged yaml.

It seems you mean multi-pass evaluation. No, this thing *right now* work like this:
- render single yaml
- update `pillars` (add and/or overwrite and/or merge etc. pillars dict contents)
- read `classes`, for each get it's location (yaml file), repeat

rbthomp:
What does everyone think of having "pillars" and "state" parameterised in master saltclass config? This way we can consume native pillars from reclass that use "applications" and "parameters"

mikemartino:
Hey folks. Is this still under active development? I just learned about _saltclass_ and was about to try it out until I read [one of @max-arnold 's comments above](https://github.com/saltstack/salt/pull/52407#issuecomment-482402450). Any thoughts on which release this might make it into? Thanks.

waynew:
Note that with the change to a single branch development model, this should be rebased/edited onto `master`

max-arnold:
@a-a-abramov Any estimations on when you will be able to finalize this PR (if at all)?

a-a-abramov:
> @a-a-abramov Any estimations on when you will be able to finalize this PR (if at all)?

I've rewritten the utility from top to bottom after our [conversation](https://github.com/saltstack/salt/pull/52407#issuecomment-512329387) in summer, but couldn't make myself finalize the whole thing. It was essential to also rewrite the test suite since the old one is abysmal.

I'll try to find some time and make necessary changes at the weekend.

mrichar1:
Thanks for putting in all the work to refactor everything - I've had a quick skim over the changes and it all seems sensible to me! (extra thanks for fixing my bug #50755!)

I'd really like to see this merged in - I started using saltclass early days in dev and really love it, and would love to see it working again as a stable salt add-on.

Is there anything now that I can do to help get this moving again? Happy to help where I can!

max-arnold:
I'm also willing to help! Would love to see this merged.

max-arnold:
This PR needs to be rebased against the master branch (no need to reopen it, just rebase and switch the branch by editing the PR on Github).

Also I guess it would be sensible to wait till the [blacken PR](https://github.com/saltstack/salt/pull/55765) is merged.

max-arnold:
And it looks like SUSE is considering using Saltclass: https://github.com/uyuni-project/uyuni-rfc/pull/20

The linked RFC mentions this PR as a potential roadblock:

> The unknown is if the PR will be ever finished or if we can invest and try to finalize it ourselves. It is possible that the work stopped only temporarily due to changes in salt releases, or that the author being external contributor chose to abandon the effort.

mrichar1:
I've just cloned a-a-abramov/saltstack:refactor-saltclass-sq, and attempted to merge against saltstack/salt:master and the only conflict appears to be `tests/unit/pillar/test_saltclass.py` - which I assume would just be an `ours` merge?

Just thought I'd flag it up in case it's useful to get this done before the `blacken` PR after all...

a-a-abramov:
Hey!
I'm terribly sorry for keeping your hopes high few months back. Obviously I failed to live up to your expectations... Have a lot of work to do besides patching saltstack :(
I'd like to see this PR merged, but I really need a hand here. Single most important thing I needed help with in october was test cases. I just failed to compose a tree that simultaneously makes (at least some!) sense AND covers all important cases in code. 
Also I couldn't compose an algorithm that mimics reclass 1-to-1, there still some artifacts that pop up here and there. Oh yeah, I also lost my remote playground, so now I have to recreate it locally. Some guidance on this topic would definitely ignite my interest again!

sagetherage:
@a-a-abramov can you rebase against master like Max suggested? https://github.com/saltstack/salt/pull/52407#issuecomment-589583819

mrichar1:
That's the [blacken PR](https://github.com/saltstack/salt/pull/55765) now merged so it should be safe to rebase against master now.

mchugh19:
Hi @a-a-abramov. In case it's helpful, rather than having to setup a new dev environment to salt yourself, @waynew has a docker image hosting all the needed bits already: https://hub.docker.com/r/waynew/salt-sprints. It allows you to checkout your git repo and run the test suite without having to worry about the environment setup. 

There is also an effort these days to host testing workshops on the community open hour. If you'd like a hand try shouting out in the [salt slack development channel](https://saltstackcommunity.slack.com/messages/develop/).

afletch:
@a-a-abramov It seems and progress on Saltclass is held up by this refactor, would you be able to share progress or status report on this refactor please?

max-arnold:
Ref https://github.com/saltstack/salt/issues/58969

DmitryKuzmenko:
@sagetherage could you please reassign the reviewer from me?

ptitdoc:
Hello,

Without wanting to add a mess to the mess I customized saltclass for my own need. Maybe I should wait for this to be merged first but here are the general ideas:
- Use base classes from master configuration
- Allow additional filters such as serialization filters and is_dict
- Allow Depth-First search instead of Breath-First for classes (probably I should change the behavior based salt master configuration as suggested in the current thread).

Currently the changes are limited regarding the current pull request, but I would prefer the current to be merged before adding more stuff. More especially I have no test cases :-(.

What do you suggest?

sagetherage:
@ptitdoc this PR is pointing to a branch that is no longer being updated or supported. It will not be merged unless it is rebased and pointing to the current `master` branch and tests will be required to be merged. We are considering closing this PR.

afletch:
@sagetherage If this is closed, where does that leave SaltClass as part of SaltStack? At the very least I think the closure of this PR should lead to some clarification on that front, since as it stands it isn't really usable.

sagetherage:
> @sagetherage If this is closed, where does that leave SaltClass as part of SaltStack? At the very least I think the closure of this PR should lead to some clarification on that front, since as it stands it isn't really usable.

@afletch this is only a PR that is pointing to the incorrect branch, is the reason for the closure. Work can still be accomplished and I can re-open, but someone from the community will need to commit to the work, rebase this against `master` and write tests.

afletch:
@sagetherage sorry, I was speaking about the PR as a whole, which it seems has stalled since @a-a-abramov no longer has the time to work on it. This PR was the 'hope for SaltClass' and without it - as even @a-a-abramov notes - SaltClass as distributed in SaltStack, is pretty much unusable. Therefore, my point was; if this PR is closed it signifies that SaltClass as pretty much abandoned. I raised this in Open Hour a few months back and there seemed little appetite beyond "someone needs to pick it up". Perhaps it should be removed (though I appreciate that's a discussion for elsewhere, not this PR).

sagetherage:
> @sagetherage sorry, I was speaking about the PR as a whole, which it seems has stalled since @a-a-abramov no longer has the time to work on it. This PR was the 'hope for SaltClass' and without it - as even @a-a-abramov notes - SaltClass as distributed in SaltStack, is pretty much unusable. Therefore, my point was; if this PR is closed it signifies that SaltClass as pretty much abandoned. I raised this in Open Hour a few months back and there seemed little appetite beyond "someone needs to pick it up". Perhaps it should be removed (though I appreciate that's a discussion for elsewhere, not this PR).
> edit: that would be a shame, as SaltClass had great promise.

@afletch I hear you! I  have also tried to find someone that is willing to even have this discussion! you can find me on saltstackcommunity.slack.com or keybase or IRC all with the same handle as here or you can email me [sage@saltstack.com](mailto:sage@saltstack.com) if you want to discuss further. I had someone in mind for this work but that fell through, recently. Still looking.

sagetherage:
@a-a-abramov wondering if this may be a good candidate for a [salt-extension](Reviewed in Grooming 2021-MAY-03 closing), unsure, merely an idea.

prannonpendragas:
Are there any updates regarding this issue?

Ch3LL:
>  But I saw a couple of PRs featuring saltclass and don't really know if there's anything left to improve. If there is - let me know! I also think that we should first agree on the scope here. Ideally gather some feedback and/or requests from both the community and salt team. 

Did you want to close this PR and open a new one when issues have been created around the other issues?


@sagetherage 's suggestion is also a good one. If someone in the community is willing to maintain it we could create a saltclass pillar salt-extension project (https://github.com/saltstack/salt-extension) This would allow this module to be managed and release outside of Salt.

tomduijf (approved):
Fwiw, this version works nicely, even with v3004
Tested with various live situations already.

Code wise
- Simpler logic, better to understand
- Faster
- better documented

ReubenM:
I'm not sure why this couldn't be merged as-is. It fixes a lot of outstanding issues. I know there were some ideas of adding more bells and whistles to it, but for the basic functionality this patch is already way more stable and functional than what currently exists.

I would really like to see this merged, I use saltclass to manage orchestration in testing envrionments, and I don't know of any other solution within salt that would allow effective management of large numbers of nodes that need lots of pillar data associated with them.

Ch3LL:
The most recent comment from @a-a-abramov led me to belive there are still issues that need to be resolved.

Also this will require a changelog to be added and it needs to be submitted against the master branch not 2019.2.

Ch3LL:
Closing this PR due to inactivity and the target branch is wrong. If anyone wants to continue this PR, please open a new PR against the master branch with the changes and we can continue progressing these changes.

PR #52412, opened 2019-04-04
zypperpkg: install and update knows about patterns

### LINKED ISSUE #52402: pkg.installed and zypperpkg.py do not recognize installed patterns

In zypper (openSUSE) we can install other kind of objects similar to packages. Currently we recognize package, patch, pattern and product. Eventually all of them are resolved to a set of packages.

In zypper we can specify the type using the `--type` parameter or a prefix before the package name. For example, those two sentences are equivalent:

```bash
zypper install --type pattern SUSE-MicroOS
```

```bash
zypper install pattern:SUSE-MicroOS
```

We can now use the salt state `pkg.installed` to install packages and patterns:

```yaml
example:
  pkg.installed:
    - name: pattern:SUSE-MicroOS
```

The problem is that `pkg.installed` and `zypperpkg.list_pkgs` do not recognize patters, and the `rpm` query do not return them, breaking the logic to identify if the package needs to be installed and to signalize the installation as a success or fail.

### Setup
Any SaltStack version will be OK, on top of any SUSE / openSUSE variant.

### Steps to Reproduce Issue

We can reproduce the issue installing any pattern, for example:

```bash
salt-call -l debug state.single pkg.installed name=pattern:SUSE-MicroOS
```

If the pattern is not installed, it will be correctly installed but the state will fail, as is not able to validate the installation. If the pattern is already present, it will try to reinstall again (as is not able to identify the pattern as already installed), and will fail again in the detection, making the state failing. 

### Versions Report

Any version.

aplanas:
In the openSUSE family we have several kind of packages, like
patterns, patches and products. Currently we can use the `pkg`
state and the `pkg` module to install them, but as they are not
listed under any `rpm` command, the module and the state will
report a Fail.

This patch updates the zypperpkg `list_pkgs` to explicitly include
this kind of objects under the list, so the rest of the commands
from the same execution module (install, update, etc) will report
the correct restul at the end.

Because the way that the `installed` state from `pkg` is designed,
we cannot be implicit about the kind of objects that we are
installing, so we need to indicate in the SLS that at least one
object is a pattern or a patch:

```yaml
  example:
    pkg.installed:
      - name: pattern:SUSE-MicroOS
      - includes: [pattern]
```

The `includes` parameter will be passed to the `list_pkgs` from
the zypper execution module, that will return the package list
using RPM (for packages) and zypper (for patterns). This more
complete report will be used in `pkg.installed` to identify the
correct installation of the new object.

Closes #52402

### DIFF
--- a/salt/modules/zypperpkg.py
+++ b/salt/modules/zypperpkg.py
@@ -726,7 +726,7 @@ def version_cmp(ver1, ver2, ignore_epoch=False):
     return __salt__['lowpkg.version_cmp'](ver1, ver2, ignore_epoch=ignore_epoch)
 
 
-def list_pkgs(versions_as_list=False, root=None, **kwargs):
+def list_pkgs(versions_as_list=False, root=None, includes=None, **kwargs):
     '''
     List the packages currently installed as a dict. By default, the dict
     contains versions as a comma separated string::
@@ -740,6 +740,10 @@ def list_pkgs(versions_as_list=False, root=None, **kwargs):
 
     root:
         operate on a different root directory.
+
+    includes:
+        List of types of packages to include (package, patch, pattern, product)
+        By default packages are always included
 
     attr:
         If a list of package attributes is specified, returned value will
@@ -777,6 +781,8 @@ def list_pkgs(versions_as_list=False, root=None, **kwargs):
     attr = kwargs.get('attr')
     if attr is not None:
         attr = salt.utils.args.split_input(attr)
+
+    includes = includes if includes else []
 
     contextkey = 'pkg.list_pkgs'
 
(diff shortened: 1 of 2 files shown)


aplanas:
I am not in love with this solution, but is the best one that I am able to produce without a big refactoring in `pkg.installed`, something that I am not confident to do due the amount of architectures that I am not able to test.

The big problem of this solution, in my view, is that is explicit in the kind of objects that we want to install / update / uninstall. In the zypperpkg level we are able to discover the nature of the package, but 'pattern' is not a concept that 'pkg.installed' have.

As this is an use case for openSUSE / SUSE I decided to commit the code, but if any one have a more generic proposal that I can implement and test with confidence, I will be glad to read.

dwoz (approved):

PR #52420, opened 2019-04-04
Update state.module docs

### LINKED ISSUE #44882: Documentation for states.state.module is incorrect

The documentation in https://docs.saltstack.com/en/develop/ref/states/all/salt.states.module.html#execution-of-salt-modules-from-within-states is incorrect, and contains no
actual description of the salt.states.module.run call.

The doc gives the example:
```
mine.send:
  module.run:
    - network.interfaces
```

### Setup

Install a fresh salt master, today's develop HEAD,  in a virtual machine.

### Steps to Reproduce Issue
I create a similar state in bevy_master.sls:
```
update_the_grains:
  module.run:
    - saltutil.sync_grains
```
and the result is:
```
vagrant@bevymaster:~$ sudo salt-call state.apply bevy_master
local:
    Data failed to compile:
----------
    Too many functions declared in state 'module' in SLS 'bevy_master'
```
So I remove the "-"
```
update_the_grains:
  module.run:
    saltutil.sync_grains
```
This time I get:
```
vagrant@bevymaster:~$ sudo salt-call state.apply bevy_master
local:
    Data failed to compile:
----------
    State 'update_the_grains' in SLS 'bevy_master' is not formed as a list
```
Oh, well. Go back to the old way of saying it:
```
update_the_grains:
  module.run:
    - name: saltutil.sync_grains
```
and get:
```
vagrant@bevymaster:~$ sudo salt-call state.apply bevy_master
[WARNING ] The function "module.run" is using its deprecated version and will expire in version "Sodium".
```
There is no way to win.

### Versions Report
At the bottom of the html page it says:

> Generated on December 07, 2017 at 07:11:35 MST.
> You are viewing docs built from a recent snapshot of the develop branch. `

```
vagrant@bevymaster:~$ salt --versions-report
Salt Version:
           Salt: 2017.7.0-892-g9865a31
 
Dependency Versions:
           cffi: 1.5.2
       cherrypy: Not Installed
       dateutil: Not Installed
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.8
        libgit2: 0.24.0
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.4.6
   mysql-python: Not Installed
      pycparser: 2.14
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: 0.24.0
         Python: 2.7.12 (default, Nov 20 2017, 18:23:56)
   python-gnupg: Not Installed
         PyYAML: 3.11
          PyZMQ: 15.2.0
           RAET: Not Installed
          smmap: Not Installed
        timelib: Not Installed
        Tornado: 4.2.1
            ZMQ: 4.1.4
 
System Versions:
           dist: Ubuntu 16.04 xenial
         locale: UTF-8
        machine: x86_64
        release: 4.4.0-21-generic
         system: Linux
        version: Ubuntu 16.04 xenial
```

### LINKED ISSUE #51349: salt.module.run documentation improvement?
I wanted to run a module's function via module.run (namely systemd.force_reload).
Instead of the expected behavior I got an error related to the *name* of my state.
After spending some frustrating time, I discovered somehow that the documentation on how to use the module.run functionality forgets to mention that you need to enable the 'new' way of running it, via some 'superseeded' config entry.

Please:
- move at the TOP of the documentation, the part about 'this is new behavior', because you start the example WITH the new behavior syntax and people might not read the whole page to understand how it should be used and that in fact it does NOT work like.
- make the part about the 'superseeded'' configuration be PART of the same TOP paragraph and not even more down below.

Thank you.

### Setup

### Steps to Reproduce Issue

### Versions Report

waynew:
I'm not sure if this should just go against develop, or if I should rebase and direct towards 2018.3.

### What does this PR do?

(Attempts) to clarify the module state docs. It wasn't immediately obvious from the docs that there was an old and new style, and the note about manually enabling them was all the way at the bottom.

fixes #44882
fixes #51349
Maybe some others?

### Tests written?

No - doc only (but new examples were tested manually)

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/states/module.py
+++ b/salt/states/module.py
@@ -5,71 +5,137 @@ Execution of Salt modules from within states
 
 .. note::
 
-    There are two styles of calling ``module.run``. To use the new style
-    you must add the following to your ``/etc/salt/minion`` config file:
+    There are two styles of calling ``module.run``. **The legacy style will no
+    longer be available starting in the Sodium release.** To opt-in early to the
+    new style you must add the following to your ``/etc/salt/minion`` config
+    file:
 
     .. code-block:: yaml
 
         use_superseded:
           - module.run
 
-.. note::
-
-    The old style of ``module.run`` will no longer be available in the Sodium
-    release.
-
 With `module.run` these states allow individual execution module calls to be
-made via states. To call a single module function use a
-:mod:`module.run <salt.states.module.run>` state:
+made via states. Here's a contrived example, to show you how it's done:
 
 .. code-block:: yaml
 
-    # Old Style
-    mine.send:
-      module.run:
-        - network.interfaces
-
     # New Style
-    mine.send:
+    test.random_hash:
       module.run:
-        # Note the trailing `:`
-        - network.interfaces:
+        - test.random_hash:
+          - size: 42
+          - hash_type: sha256
 
+    # Legacy Style
+    test.random_hash:
+      module.run:
+        - size: 42
+        - hash_type: sha256
 
-.. note::
-
-    The previous example is contrived and probably unnecessary to use in practice,
-    since the ``mine_functions`` and ``mine_interval`` config parameters
-    can be used to schedule updates for the mine (see :ref:`here <salt-mine>`
-    for more info).
-
-It is sometimes desirable to trigger a function call after a state is executed,
-for this the :mod:`module.wait <salt.states.module.wait>` state can be used:
+In the new style, the state ID (``test.random_hash``, in this case) is
+irrelevant when using ``module.run``. It could have very well been written:
 
 .. code-block:: yaml
 
-    add example to hosts:
-      file.append:
-        - name: /etc/hosts
-        - text: 203.0.113.13     example.com
+    Generate a random hash:
+      module.run:
+        - test.random_hash:
+          - size: 42
+          - hash_type: sha256
 
-    # Old Style
-    mine.send:
-      module.wait:
-        - name: hosts.list
-        - watch:
-          - file: add example to hosts
+For a simple state like that it's not a big deal, but if the module you're
+using has certain parameters, things can get cluttered, fast. Using the
+contrived custom module (stuck in ``/srv/salt/_modules/foo.py``, or your
+configured file_roots_):
 
-    # New Style
-    mine.send:
-      module.wait:
-        # Again, note the trailing `:`
-        - hosts.list_hosts:
-        - watch:
-          - file: add example to hosts
+.. code-block:: python
 
-In the old style, all arguments that the ``module`` state does not consume are
-passed through to the execution module function being executed:
+    def bar(name, names, fun, state, saltenv):
+        return "Name: {name} Names: {names} Fun: {fun} State: {state} Saltenv: {saltenv}".format(**locals())
+
+Your legacy state has to look like this:
+
+.. code-block:: yaml
+
+    # Legacy style
+    Unfortunate example:
+      module.run:
+      - name: foo.bar
+      - m_name: Some name
+      - m_names:
+        - Such names
+        - very wow
+      - m_state: Arkansas
+      - m_fun: Such fun
+      - m_saltenv: Salty
+
+With the new style it's much cleaner:
+
+.. code-block:: yaml
+
+    # New style
+    Better:
+      module.run:
+      - foo.bar:
+        - name: Some name
+        - names:
+          - Such names
+          - very wow
+        - state: Arkansas
+        - fun: Such fun
+        - saltenv: Salty
+
+The new style also allows multiple modules in one state. For instance, you can
+do this:
+
+.. code-block:: yaml
+
+    Do many things:
+      module.run:
+        - test.random_hash:
+          - size: 10
+          - hash_type: md5
+        # Note the `:` at the end
+        - test.true:
+        - test.arg:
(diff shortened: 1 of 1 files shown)


waynew:
@cr1st1p / @bryceml does this improve the `module.run` docs? I'm hoping it should make things a little nicer 🙂

bryceml:
If one is a newer style, I would mention what salt versions accept the newer style and when/if the old style will deprecated as well.

waynew:
@bryceml AFAICT it's supported by all the current versions of Salt. I couldn't find any specific information about how the `with_deprecated` decorator is really supposed to be used, but if I grok this correctly it should be swapped out in the Sodium release?

I'll update to include that note.

cr1st1p:
Sometimes I'm picky so, feel free to use or not my suggestions :-) 
Also, they're not all related to your changes and, mind you, I haven't worked with the module.run since reporting the issue(s).

My suggestions aside, I really appreciate your efforts! Thank you and keep it rolling!

a) I had a little bit of a problem grasping the purpose of the comment '# note the trailing ':' '
What do you think if we would start by giving the general form of usage, something like:

```
# Old style:
some identifier:
  - module.run:
    - name: module_function_to_call # or use module_function_to_call *as* identifier
    - arg1: value  # for that function
    - arg2: value  # for that function
    - other standard *state* parameters
   
# new style:
some identifier:
  - module.run:
    - module_function_to_call : 
      - arg1: value
      - arg2: value
   - other standard *state* parameters

```
Alternative: do not use examples using a module function call that do not require any argument (network.interfaces), but use an example that *does* require one.
 (Update: Heh...one small mistake on my part, see f) and g) :-) 

b) it is not clear (to me), if with the Sodium release, I *must* remove that configuration snippet. I.e. If I keep it and upgrade to Sodium release, will it error out? Probably a dev should chime in with that info.

c) One example seems bogus to me (untested):
Old style, it seems to call module function  **hosts.lists** ( https://docs.saltstack.com/en/latest/ref/modules/all/salt.modules.hosts.html doesn't show the 'lists' function)
New style example calls  **hosts.lists_hosts** (probably correct)

d) I wouldn't mix usage of  'old' and 'legacy' words
e) Same for 'new style' and 'modern'. 'Modern examples' sounds strange to me, for example.
f) using as starting example a module function call that takes another module function (namely 'mine.send') is very confusing. I had to re-read that part of the current online document several times. Because at first look I thought main _actor_ is module function _network.interfaces_ , then some other things did not look right.
  But in fact, the code is about module function 'mine.send'.
g) side effect of f) - I am not sure if the 1st example, in 'new style', is correct.

Thank you again for your effort!

waynew:
> Sometimes I'm picky so, feel free to use or not my suggestions :-)

I'll be sure to take them with a grain of salt :trollface: 

> a) I had a little bit of a problem grasping the purpose of the comment '# note the trailing ':' '

Mainly to call out the fact that it's necessary, and new. I can't tell you the number of times that I've stared at two similar code samples that were off by one character before I finally realized they were different.

> What do you think if we would start by giving the general form of usage

I'm not against that - though we definitely should have the case where no argument is required, because it's different and not at all intuitive. I certainly never would have guessed at that syntax

> b) it is not clear (to me), if with the Sodium release, I _must_ remove that configuration snippet. I.e. If I keep it and upgrade to Sodium release, will it error out? Probably a dev should chime in with that info.

That's probably because it's not clear to me either :trollface: I'll have to find someone who's familiar with the `with_deprecated` decorator.


> c) One example seems bogus to me (untested):
> Old style, it seems to call module function **hosts.lists** ( https://docs.saltstack.com/en/latest/ref/modules/all/salt.modules.hosts.html doesn't show the 'lists' function)
> New style example calls **hosts.lists_hosts** (probably correct)

Ahhh, good catch - I missed that 👍 
 
> d) I wouldn't mix usage of 'old' and 'legacy' words

I probably agree with that. If I change that I'd definitely change it to `legacy`, if the old style is going to go away, as `legacy` does (or should) communicate that.

> e) Same for 'new style' and 'modern'. 'Modern examples' sounds strange to me, for example.

I could probably poke around in the docs elsewhere to see if there's any kind of convention for indicating new-style syntax.

> f) using as starting example a module function call that takes another module function (namely 'mine.send') is very confusing.

I also thought it was a little weird. I was (and still am) of mixed feelings about it - it's cool that it calls out `mine.send`, since that's a useful thing to think about and use, but I'm not entirely sure if this is the best place for it. I defaulted to leaving the reference in there since that's what the original documentation had.

Oh! Looking a little closer into it, I see what you mean. I did *not* expect that. That is very very weird.

Yeah, I think I'll probably address the docs and start from a slightly different place there. I think it's probably valuable to keep the mine examples, but I'll stick those towards the end.

Thanks for your input!

waynew:
@cr1st1p take a look at my updated docs. Turns out the existing `mine.send` samples weren't even valid syntax for salt, so I just went ahead and removed them. Hopefully it should be a bit cleaner now.

cr1st1p:
@waynew : better ! Thank you!

dwoz (approved):
@waynew I think this would be good to backport to 2018.3

PR #52425, opened 2019-04-05
[2017.7.9] Fix lint issues

s0undt3ch:
### What does this PR do?
Fix lint issues on 2017.7.9

### Tests written?

No

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/pillar/git_pillar.py
+++ b/salt/pillar/git_pillar.py
@@ -518,7 +518,7 @@ def __virtual__():
                     'could not be loaded, is GitPython installed?'
                 )
                 return False
-            if not git.__version__ > '0.3.0':
+            if git.__version__ <= '0.3.0':
                 return False
             return __virtualname__
         else:
--- a/.testing.pylintrc
+++ b/.testing.pylintrc
@@ -154,7 +154,8 @@ disable=R,
   E8266,
   E8402,
   E8731,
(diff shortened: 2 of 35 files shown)


salt/pillar/git_pillar.py
```diff
@@ -518,7 +518,7 @@ def __virtual__():
                     'could not be loaded, is GitPython installed?'
                 )
                 return False
-            if not git.__version__ > '0.3.0':
+            if git.__version__ <= '0.3.0':
```
waynew: I know it's arguably unrelated to this PR, but... do we not have a better way to check version strings? I don't know if git guarantees it's version #s will be 0-9, and only 3 segments, but this seems mildly sketchy to me.

waynew: I know it's arguably unrelated to this PR, but... do we not have a better way to check version strings? I don't know if git guarantees it's version #s will be 0-9, and only 3 segments, but this seems mildly sketchy to me.

s0undt3ch: Yes, unrelated to this PR ;)
And, to add up, this kind of comparison works on py2 and not that well on py3.

waynew:
Plot twist: this breaks some Whitespace program :trollface:

waynew (approved):

PR #52434, opened 2019-04-07
convert get to traverse_dict_and_list

mchugh19:
### What does this PR do?
Replaces value.get with call to traverse_dict_and_list to support list and deeper getting
Also only copy tst files from saltmaster, rather than run `cp.cache_master` on each invocation

### Previous Behavior
saltcheck was only able to process a single depth dict in assertion_section 
`cp.cache_master` always run

### New Behavior
saltcheck's assertion_section can now traverse dicts and lists
`cp.cache_files` run for all *.tst files

### Tests written?
No

### Commits signed with GPG?
No

### DIFF
--- a/salt/modules/saltcheck.py
+++ b/salt/modules/saltcheck.py
@@ -201,6 +201,7 @@ from salt.utils.json import loads, dumps
 import salt.utils.files
 import salt.utils.path
 import salt.utils.yaml
+import salt.utils.data
 import salt.client
 import salt.exceptions
 from salt.utils.odict import OrderedDict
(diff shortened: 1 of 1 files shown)


mchugh19:
Based on some feedback from the slack develop channel, this is still a work in progress.

mchugh19:
Now saltcheck should perform a `cp.cache_dir` for each requested state. I'd appreciate a second opinion on this logic

@gtmanfred

mchugh19:
It appears the included solution does not work with salt-ssh, and tst files are not copied to the minion. @garethgreenaway or @Ch3LL can you recommend a better approach? 

The goal is to run salt-ssh someminion saltcheck.run_state_tests somestate and have that run properly. At the moment salt-check runs `cp.cache-dir` for the given states, and then walks the cache looking for associated .tst files which it then renders. This works fine for the `salt` and `salt-call` cases, but not for `salt-ssh`. What's a better method?

mchugh19:
I've figured out a way to support salt-ssh. It works but it might need feedback, so I've opened it as another PR for comments there. It includes the functionality of this PR, so I'll close this one out. If the saltcheck ssh wrapper isn't suitable, this PR can be re-opened.

PR #52440, opened 2019-04-08
Update my contact details, as the previous email address is no longer in use

mirceaulinic:
Targeting the 2017.7 branch, to ensure the docs will be updated everywhere.


rares-pop:
Is this maintained anymore?

I ask because I see here:
https://docs.saltstack.com/en/latest/contents.html

only 2018.3.4 and 2019.2.0.

Even though here you can download any version:
https://repo.saltstack.com/index.html

Issue #50927, opened 2018-12-20
Salt python client "get_full_returns" seems return data from incorrect jid


Yukinoshita-Yukino:
### Versions:
salt-master: 2018.3.2
salt-minion: 2015.5.10 (not sure if this old version of minion is causing this problem)
python salt client: 2018.3.0 (pip install salt==2018.3.0)

### Problem:
Seems like when I call "get_full_returns" when sending out async cmd(cmd_async) I got some return data not from the jid I was given by the cmd_async call.

### Steps:
Let's say I have two Python program running individually, using different Python venv, program A and program B.
They are both sending cmd to three servers, server 1, 2, and 3.
Program A sends out cmd by using async cmd
```
# Program A
client = salt.client.LocalClient()
jid = client.cmd_async(target, 'cmd.run', [cmd, 'shell="/bin/bash"', 'runas="{}"'.format(runas)])
rv = client.get_full_returns(jid, [target], timeout=180)
```
Program B just sends out cmd using list mode
```
# Program B
client = salt.client.LocalClient()
ret = client.cmd(target_list, 'cmd.run', [cmd, 'shell="/bin/bash"', 'runas="root"'], tgt_type='list')
```

Assuming program A is running job on server 1 with jid 100, program B is running job on server 1, 2, 3.
When I call something like `rv = client.get_full_returns(100, ['server 1'], timeout=180)` in program A, I got return from job triggered by program B on server 1.

Any idea is this due to the old version of salt-minion or misuse of the client call or is it a bug?
I have checked the release note  from 2018.3.1 to 2018.3.3 and did't find something that seems related to this problem.

Ch3LL:
is the program B producing the same JID as program A?

Yukinoshita-Yukino:
> is the program B producing the same JID as program A?

I checked the JID by looking into the debug log from salt and I believe no, they don't have the same JID.

This problem occurs quite often in my situation, as program B is sending out cmd at the frequence around 90calls/2min. In program A each time I call get_full_returns there's around (or maybe higher) 50% chance I got some return from the job created by program B. That's quite a abnormally high chance for 10+digits JIDs I think.

Ch3LL:
thanks for clearing that up. I was able to replicate this  and here was my test case:

```
➜  salt git:(ad0d83c6a8) ✗ cat ~/programa.py 
import salt.client
client = salt.client.LocalClient()
target = 'the_cake_is_a_lie'
cmd = 'echo testa'
[...]

PR #52459, opened 2019-04-09
Checking for jid before returning data

brejoc:
### What does this PR do?

Seems raw can have returns for the same minion, but an other job. In order to not return resutls from the wrong job, we need to check for the jid.

https://github.com/saltstack/salt/issues/50927

### Previous Behavior

Wrong job results might be returned. Please take a look at https://github.com/saltstack/salt/issues/50927

### New Behavior

Only the correct jobs results should be returned.

### Tests written?

No

### Commits signed with GPG?

Yes


Ch3LL (changes requested):
Can you write a test for this change?

brejoc:
@Ch3LL Sorry for dragging my feet here for so long, but finally I've just added two test cases. Could you please check your the test runs? Hudson seems to have problems fetching the changes.

Ch3LL:
I restarted the tests. I went ahead and tested this PR against my test case i pasted in the issue and unfortunately im still seeing the issue:

```
(salt-py2) ➜  ~ for i in {1..100}; do sudo python2 ~/programa.py; done
[sudo] password for ch3ll: 
{u'newrecipe': {u'ret': u'testa'}}
{u'newrecipe': {u'ret': u'testa'}}
{u'newrecipe': {u'ret': u'testa'}}
{u'newrecipe': {u'ret': u'testa'}}
{u'newrecipe': {u'ret': u'testb'}}
{u'newrecipe': {u'ret': u'testa'}}
```

are you seeing it fixed for you? I tried multiple times to make sure.

brejoc:
@Ch3LL I've been re-running the tests several times and for me the issue is gone. As soon as I remove the patch, I see it again. Are you running both scripts on the master?

brejoc:
@Ch3LL I'm closing this PR because I've just opened the same PR against the master branch: https://github.com/saltstack/salt/pull/55216

PR #52471, opened 2019-04-10
Fix logging support for salt-api

rares-pop:

### What does this PR do?
Sets up the multi-processing properly for salt-api.

This fixes the biggest part of the memory leaks I mentioned here:
https://github.com/saltstack/salt/pull/49949

Note that some of the remaining memory leaks (after this change) have to do with this change:
https://github.com/saltstack/salt/pull/52260

Once those two goes in, there are very few cherrypy objects leaked, depending on the version of the cherrypy installed. I'm on my way to track more of these down, in a subsequent change.

@twangboy , @dwoz - I know you guys wanted to track down memory leaks. This is a tool I used when hunting those leaks:
https://mg.pov.lt/objgraph/objgraph.html

Maybe there are ways to integrate this library in the general integration-test infrastructure, so we'll benefit from those reports for all existing tests, and avoid having specific memory-leak tests. I'm talking about calling objgraph.get_new_ids from the main infrastructure and make sure the Count_Deltas are meaningful.

### Previous Behavior
Logging wasn't working for salt-api on Windows, and all the logging messages were being leaked (multiple times - for logging NullHandler, QueueHandler, ...), and in about 3-4000 jobs created through salt-api, the master was bogged down.

### New Behavior
Logging works on Windows now and there are way fewer leaks (some cherrypy classes are still leaked, depending on which version of cherrypy is used)

### Tests written?
No

### Commits signed with GPG?

Yes


mattp-:
does this affect saltnado impl as well?

rares-pop:
@mattp- I'm not aware of saltnado. I did port salt-api on Windows here: https://github.com/saltstack/salt/pull/48001

But that's pretty much unusable without this change, because in a few days salt-api will eat-up the whole memory.

I only played with the cherrypy configuration, on Windows.

BTW: I'm in the process of retargeting this PR to be against '2018.3' branch as opposed to 'develop', since it's a fix and not a feature.

twangboy (dismissed):

twangboy:
Please rebase this against master. Would it be possible to write some tests?

twangboy:
I rebased it. We'll still need those tests though.

rares-pop:
@twangboy - I did move on something else and unfortunately I don't have much bandwidth for salt. I hope one of my colleagues will pick this up and finish it @andzn, @chotea.

That being said, there are two things I'd bring up before mentioning tests here:
 * I know both the logging mechanisms and the multi process class got reworked - maybe this isn't needed anymore?
 * this change prevents memory leaks, and there are no global ways of testing leaks in salt. There was a pending discussion with few folks here where we suggested having the memory-leak capability at the testing framework scale, and not for individual tests.

If you guys still insist on a test, I'd be curious what would that look like? Any suggestions? @dwoz?

DmitryKuzmenko (dismissed):

PR #52485, opened 2019-04-10
Update CodeClimate settings

waynew:
I'm pretty sure this is the right place to merge this to - it should go to 2019.2 and develop as well.

This tweaks the settings for code climate.
https://codeclimate.com/github/saltstack/salt/pull/51461 had a matching
block with mass of 38. Apparently that's too low because that code was
not even the same at all. Bumping it up a bit to avoid false positives.

Also disable argument count checks.

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -1,13 +1,21 @@
-languages:
-  Ruby: false
-  JavaScript: false
-  Python: true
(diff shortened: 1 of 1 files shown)


dwoz (approved):
Sweet.

twangboy (approved):

PR #52495, opened 2019-04-11
Explicitly close IPC connections

dwoz:
### What does this PR do?

- Makes salt event context aware
- Use context manages for events when possible
- When not using context mangers for event, make sure we call destroy method on event instances.

### Tests written?

No - Fixing existing tests

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/client/mixins.py
+++ b/salt/client/mixins.py
@@ -327,7 +327,7 @@ class SyncClientMixin(object):
             ),
             tag,
             print_func=print_func
-        ) as namespaced_event
+        ) as namespaced_event:
 
             # TODO: test that they exist
             # TODO: Other things to inject??
--- a/salt/client/mixins.py
+++ b/salt/client/mixins.py
@@ -307,7 +307,6 @@ class SyncClientMixin(object):
                 'user': low.get('__user__', 'UNKNOWN'),
                }
 
-
         if print_event:
             print_func = self.print_async_event \
                 if hasattr(self, 'print_async_event') \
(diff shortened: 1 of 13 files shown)


Ch3LL (approved):

DmitryKuzmenko (approved):

salt/client/mixins.py
```diff
@@ -323,113 +316,117 @@ def _low(self, fun, low, print_event=True, full_return=False):
...
-        # TODO: test that they exist
-        # TODO: Other things to inject??
-        func_globals = {'__jid__': jid,
-                        '__user__': data['user'],
-                        '__tag__': tag,
-                        # weak ref to avoid the Exception in interpreter
-                        # teardown of event
-                        '__jid_event__': weakref.proxy(namespaced_event),
-                        }
+        ) as namespaced_event:
```
waynew: I'm not sure how this works as the context manager (and maybe it will become obvious as I read further into this code), but does the `NamespacedEvent` close this `get_event`, or should we nest this like:

```
        with salt.utils.event.get_event(
                'master',
                self.opts['sock_dir'],
                self.opts['transport'],
                opts=self.opts,
                listen=False,
        ) as event:
            with salt.utils.event.NamespacedEvent(event, tag, print_func=print_func) as namespaced_event:
                ...
```

waynew:
👍 I'm certainly a fan

salt/client/mixins.py
dwoz: NamespacedEvent is also a context manager with this patch and will destroy the event it's namespacing.

waynew: 👍 Destroy all the things! 🔥 

In that case, I am much in favor of this :)

s0undt3ch (approved):

PR #52498, opened 2019-04-11
Batch: avoid recreation of the local client object

moio:
### What does this PR do?

It avoids the recreation of a LocalClient object at every request when using batching from the API.

In a very heavy API test, this allowed to considerably reduce per-call runtime.

None created.

### Tests written?

Should be already covered by existing tests.

### Commits signed with GPG?

Yes.


moio:
@dincamihai I do not see real downsides, the way code is structured currently seems to be mainly due to history. Can you think of cases in which this fails?

dincamihai:
@moio i'm not aware of any issues. :+1: from me

waynew:
I don't see any issues either - I thought that `get_local_client` might have some sort of caching, but no, it doesn't. I'm not aware of any reason we wouldn't want to do this, unless there's a better approach we could take.

Does anyone else on @saltstack/team-core have an opinion? This is just changing the `cli.batch` to have the option to pass in a `LocalClient` rather than having to go out and get a new one each time. Based on my (*checks watch*) 10 minutes of experience in this area, it looks reasonable :)

DmitryKuzmenko (dismissed):
Looks good to me and basically it's covered by existing tests.

sagetherage:
@moio please rebase this against the master branch

moio:
@sagetherage should be good now

DmitryKuzmenko:
@moio thank you for the update. Unfortunately we're not accepting any PR having no tests anymore. This PR is pretty simple so a regression tests for these changes would be enough. Could you add it please? I'd suggest to add 2 tests:
1. Update `tests/unit/cli/test_batch.py:BatchTestCase`. Add a simple test that calls `batch = Batch(opts, quiet=True, local=something` and assert `batch.local == something`. It's pretty simple.
2. Update `tests/unit/test_client.py`. Add a test calling the `LocalClient.cmd_batch` with `salt.cli.batch.Batch` mocked and assert the `Batch` was constructed with the `local` argument equal to the client object.

sagetherage:
There is a test clinic this week on twitch with @waynew on Thursday https://www.twitch.tv/saltstackinc

waynew:
While for the most part this *was* covered by existing tests, I also went ahead and added tests to ensure that we really are using the cached LocalClient when it's been provided.

moio:
Thank you very much @waynew - I had this on my todo list but didn't have the chance to reach the work item yet!!!

moio:
Sorry, this PR got stale and I will probably not have time to resume it in foreseeable future. Closing.

PR #52499, opened 2019-04-11
Data filter falsey

github-abcde:
### What does this PR do?
This adds a function `filter_falsey` to `salt.utils.data`. The function can be used to filter values from a list or dict that evaluate to false (but are not `bool(False)`). Optionally, a level of recursion can be set to process nested dicts/lists. Also optionally, a list of types to not filter can be passed in `ignore_types`.

None that I know of.

### Previous Behavior
The functionality does not exist.

### New Behavior
We are now able to filter falsey values from dicts or lists.
This is especially handy when having to process a lot of kwargs that are explicitly specified in the function argument list, which default to `None`, but need not be passed on if they are None.
For example:
```
def some_function(foo=None, bar=None, baz=None):
    new_kwargs = {}
    if foo is not None:
        new_kwargs.update('Foo': foo)
    if bar is not None:
        new_kwargs.update('Bar': bar)
    if baz is not None:
        new_kwargs.update('Baz': baz)
    some_other_function(**new_kwargs)
```
will become
```
def some_function(foo=None, bar=None, baz=None):
    new_kwargs = salt.utils.data.filter_falsey({
        'Foo': foo,
        'Bar': bar,
        'Baz': baz,
    })
    some_other_function(**new_kwargs)
```
Which is a typical use-case for most Boto calls.

### Tests written?
Yes

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/utils/data.py
+++ b/salt/utils/data.py
@@ -993,3 +993,40 @@ def json_query(data, expr):
         log.error(err)
         raise RuntimeError(err)
     return jmespath.search(expr, data)
+
+
+def filter_falsey(data, recurse_depth=None, ignore_types=None):
+    '''
+    Helper function to remove items from a dict or list with falsey value.
+    Removes ``None``, ``{}`` and ``[]``, 0, '' (but does not remove ``False``).
+    Recurses into subdicts or sublists if ``recurse`` is set to ``True``.
+
+    :param dict/list data: Source dict or list to process.
+    :param int recurse_depth: Recurse this many levels into values that are dicts
+        or lists to also process those. Default: 0 (do not recurse)
+    :param list ignore_types: Contains types that can be falsey but must not
+        be filtered. Default: Only booleans are not filtered.
+
+    :return type(data)
+    '''
+    if ignore_types is None:
+        ignore_types = []
+    if isinstance(data, dict):
+        ret = {}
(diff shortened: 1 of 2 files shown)


salt/utils/data.py
```diff
@@ -993,3 +993,40 @@ def json_query(data, expr):
...
+    Recurses into subdicts or sublists if ``recurse`` is set to ``True``.
+
+    :param dict/list data: Source dict or list to process.
+    :param int recurse_depth: Recurse this many levels into values that are dicts
+        or lists to also process those. Default: 0 (do not recurse)
+    :param list ignore_types: Contains types that can be falsey but must not
+        be filtered. Default: Only booleans are not filtered.
+
+    :return type(data)
+    '''
```
twangboy: We need to add a `.. version-added:: Neon` since this will be new to the Neon release.

twangboy (changes requested):

salt/utils/data.py
github-abcde: You are right, I totally forgot about that. Added.

github-abcde:
@twangboy I've implemented the changes you requested. Is there anything else I can do for this PR?

twangboy (approved):

twangboy:
@github-abcde Looks good. Thanks!

PR #52502, opened 2019-04-11
Show correct output when archive needs to be extracted w/tar CLI

terminalmage:
In cases where the tar CLI needs to be used (such as when tar options are passed to the state), the `extracted_files` item in the changes dict shows "no tar output so far" because the tar command being run doesn't include the `-v` option to show the paths that were extracted. This fixes that and provides a list just like when the `tarfile` module is used to extract.

### DIFF
--- a/salt/states/archive.py
+++ b/salt/states/archive.py
@@ -1350,7 +1350,10 @@ def extracted(name,
                         )
                         return ret
 
-                    tar_opts = shlex.split(options)
+                    # Ignore verbose file list options as we are already using
+                    # "v" below in tar_shortopts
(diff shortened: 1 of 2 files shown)


twangboy (approved):

dwoz (changes requested):
@terminalmage It looks like the failing tests are related, can you take a look?

terminalmage:
I'll have a look over the weekend, thanks for the heads-up.

terminalmage:
@dwoz The tests caught a flaw in my PR. I fixed it, updated the tests, and rebased onto the head of 2018.3.

dwoz (approved):

PR #52478, opened 2019-04-10
Pygit2: re-read repo on config change.

DmitryKuzmenko:
### What does this PR do?
Fixes pygit2 tests error `index out of range`.
The error is produced because pygit2 doesn't re-read the remotes config in case we access remote by index. But our code changes git config directly adding remote to the config file after pygit2 repository object is already created.

### What issues does this PR fix or reference?
N/A

### Previous Behavior
1. Create a new git repo with pygit2 getting a new pygit2 repo handler: `self.repo = pygit2.init_repository(self.cachedir)`.
2. Update git config file not using the pygit2 handler adding the origin remote.
3. Access the remote using pygit2 handler: `origin = self.repo.remotes[0]`
At step 3 self.repo has no remotes because it doesn't look for changes in the repo filesystem.

### New Behavior
1. Create a new git repo with pygit2 getting a new pygit2 repo handler: `self.repo = pygit2.init_repository(self.cachedir)`.
2. Update git config file not using the pygit2 handler adding the origin remote.
2.1 If config file changed re-init pygit2 repository handler
3. Access the remote using pygit2 handler: `origin = self.repo.remotes[0]`
At step 3 self.repo has all remotes we're added on step 2.

### Tests written?
Current tests cover this case.

### Commits signed with GPG?
Yes

PR #52520, opened 2019-04-12
[WIP] pygit2 fetch test

DmitryKuzmenko:
### What does this PR do?
Here's a test for #52478 that should fail on current 2018.3 code.
I'm doing it against 2018.3 not in my PR branch because the goal is to see it's on the current code.

Test for #52478

### Commits signed with GPG?
Yes


dwoz (approved):

DmitryKuzmenko:
There were some mistakes in the test: I forgot that I've created a branch for it manually. =)))) Now I've checked it well in both cases: with and without the fix and without any local dependencies.

DmitryKuzmenko:
Commented in #52478

Issue #52525, opened 2019-04-12
deserializer_opts are not passed the serializer in file.serialize


lomeroe:
### Description of Issue/Question
The deserializer_opts are not correctly passed to the deserializer module.

### Setup
These options are not sent to the deserializer function due to using the serializer_name instead of the deserializer_name when attempting to retrieve the options from the deserializer_options dict

### Steps to Reproduce Issue
Use the file.serialize state with merge_if_exists = True and specify deserializer_opts.  

### Versions Report
```
Salt Version:
           Salt: 2019.2.0

Dependency Versions:
           cffi: Not Installed
       cherrypy: Not Installed
       dateutil: Not Installed
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.7.2
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.5.6
   mysql-python: Not Installed
      pycparser: Not Installed
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: Not Installed
         Python: 2.7.5 (default, Apr  9 2019, 14:30:50)
   python-gnupg: Not Installed
         PyYAML: 3.11
          PyZMQ: 15.3.0
           RAET: Not Installed
          smmap: Not Installed
        timelib: Not Installed
        Tornado: 4.2.1
            ZMQ: 4.1.4

System Versions:
           dist: centos 7.6.1810 Core
         locale: UTF-8
        machine: x86_64
        release: 3.10.0-957.10.1.el7.x86_64
         system: Linux
        version: CentOS Linux 7.6.1810 Core
```

Ch3LL:
thanks for the PR! appreciate it :)

PR #52526, opened 2019-04-12
Fix use of deserializer_opts in file.serialize

lomeroe:
### What does this PR do?
Corrects code attempting to pull deserialize_opts using the serializer_name instead of deserializer_name

#52525

### Previous Behavior
deserializer_opts are packed into the deserializer_options dict using the deserialzier_name as the key.  However, when retrieved to pass to the deserializer, the serializer_name is used.

### New Behavior
deserializer_name is used to retrieve the options

### Tests written?
No

### Commits signed with GPG?
No

### DIFF
--- a/salt/states/file.py
+++ b/salt/states/file.py
@@ -6846,7 +6846,7 @@ def serialize(name,
                 try:
                     existing_data = __serializers__[deserializer_name](
                         fhr,
-                        **deserializer_options.get(serializer_name, {})
+                        **deserializer_options.get(deserializer_name, {})
                     )
                 except (TypeError, DeserializationError) as exc:
                     ret['result'] = False
--- a/tests/integration/states/test_file.py
+++ b/tests/integration/states/test_file.py
@@ -42,6 +42,7 @@ import salt.utils.json
 import salt.utils.path
 import salt.utils.platform
 import salt.utils.stringutils
+import salt.serializers.configparser
 from salt.utils.versions import LooseVersion as _LooseVersion
 
 HAS_PWD = True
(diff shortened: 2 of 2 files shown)


Ch3LL:
thanks for the PR! Is it possible for you to write a regression test for this change?

lomeroe:
@Ch3LL integration tests updated

dwoz (commented):
@lomeroe It looks like we're removing coverage from the json formatter in your test change. I'm concerned that we'll now catch a bug when using the configparser formatter but miss a bug when using the json formatter. Maybe it's better to make a new test for configparser?

lomeroe:
@dwoz the purpose of that test seemed to only be to validate that the serializer_opts and deserializer_opts being passed in the state were functioning correctly (and being used), which it was not doing.  My opinion was that the particular formatter did not make a difference for the test case.  Configparser is the easiest to assert, from a data perspective, that the serializer/deserializer options are in effect.  For the json formatter, there are no options that can be passed (as a string) to modify the returned data (parameters that can must be passed in as functions) to validate the option's functionality.  The json test could be kept, but modified to pass an encoding such as utf-16 or utf-32 which will raise an error and that could be asserted.  As it was, passing an encoding of latin-1 is no different than the default, and can't be validated that it was actually passed to the deserializer function (which is why the test wasn't catching that the options weren't even being passed).

In my opinion, I don't see that it gains anything to test options to both formatters, but at the same time have no issue putting the test in there if you want it.

dwoz (approved):
@lomeroe that makes sense and thank you for the clarification. Approved! :)

PR #52535, opened 2019-04-12
Configparser item case preservation

lomeroe:
### What does this PR do?
Allows passing options to the configparser serializer/deserializer functions to perserve case of item names (section names and item value case is preserved).

### Previous Behavior
configparser converts all item names to lower case by default

For example, if the following were derserialized
```
[section]
item = value
itemTwo = value
```
it would return the dict
```
{'section': {'item': 'value', 'itemtwo': 'value'}}
```

Conversely, if the following dict were passed through the serialize function
```
{'section':{'item': 'value', 'itemTwo': 'value'}}
```
the following would be returned
```
[section]
item = value
itemtwo = value
```
### New Behavior
Case of item names can be preserved by passing "preserve_case=True" in the options of the functions

```
import salt.serializers.configparser as configparser
data = {'foo': {'someItemWithCase': 'data'}}
serialized = configparser.serialize(data, **{'preserve_case': True})
```

i.e. the following dict
```
{'section':{'item': 'value', 'itemTwo': 'value'}}
```
would return:
```
[section]
item = value
itemTwo = value
```

or more usefully, via the file.serialize state:
```
some/file:
  - file.serialze:
      - formatter: configparser
      - dataset:
          section:
              someValue: someData
              someValue2: someData2
  - merge_if_exists: True
  - deserializer_opts:
      - preserve_case: True
  - serializer_opts:
      - preserve_case: True
```
### Tests written?
Yes

### Commits signed with GPG?
No

### DIFF
--- a/salt/serializers/configparser.py
+++ b/salt/serializers/configparser.py
@@ -20,7 +20,7 @@ __all__ = ['deserialize', 'serialize', 'available']
 
 available = True
 
-def _check_preservecase_option(**options)
+def _check_preservecase_option(**options):
     '''
     Checks options passed to the serialize/deserialze function
     for the 'casesensitive' argument.  If passed as True,
--- a/salt/serializers/configparser.py
+++ b/salt/serializers/configparser.py
@@ -20,6 +20,7 @@ __all__ = ['deserialize', 'serialize', 'available']
 
 available = True
 
+
 def _check_preservecase_option(**options):
     '''
     Checks options passed to the serialize/deserialze function
@@ -33,6 +34,7 @@ def _check_preservecase_option(**options):
         options.pop('preserve_case')
     return options, preserve_case
 
+    
 def deserialize(stream_or_string, **options):
     '''
     Deserialize any string or stream like object into a Python data structure.
--- a/salt/serializers/configparser.py
+++ b/salt/serializers/configparser.py
@@ -42,16 +42,7 @@ def deserialize(stream_or_string, **options):
     :param stream_or_string: stream or string to deserialize.
     :param options: options given to lower configparser module.
     '''
-    options, PRESERVE_CASE = _check_preservecase_option(**options)
-
-    if six.PY3:
-        cp = configparser.ConfigParser(**options)
-        if PRESERVE_CASE:
-            cp.optionxform = str
-    else:
-        cp = configparser.SafeConfigParser(**options)
-        if PRESERVE_CASE:
-            cp.optionxform = str
+    cp = _create_cp_object(**options)
 
     try:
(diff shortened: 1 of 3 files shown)


dwoz (approved):

dwoz:
@lomeroe Please have a look at the linter and docs failures.

salt/serializers/configparser.py
```diff
@@ -21,18 +21,37 @@
...
 
 
+def _check_preservecase_option(**options):
+    '''
+    Checks options passed to the serialize/deserialze function
+    for the 'casesensitive' argument.  If passed as True,
+    the configparser object will preserve case in items
+    '''
+    preserve_case = False
+    if 'preserve_case' in options:
```
rares-pop: this is an anti-pattern that we shouldn't use.

use: preserve_case = options.get('preserve_case', False)

salt/serializers/configparser.py
```diff
@@ -21,18 +21,37 @@
...
+        if options['preserve_case']:
+            preserve_case = True
+        options.pop('preserve_case')
```
rares-pop: actually you can replace the 5 lines with the preserve_case = options.pop('preserve_case', False)

salt/serializers/configparser.py
```diff
@@ -21,18 +21,37 @@
...
+
+
 def deserialize(stream_or_string, **options):
     '''
     Deserialize any string or stream like object into a Python data structure.
 
     :param stream_or_string: stream or string to deserialize.
     :param options: options given to lower configparser module.
     '''
+    options, PRESERVE_CASE = _check_preservecase_option(**options)
```
rares-pop: why upper-case?

salt/serializers/configparser.py
```diff
@@ -21,18 +21,37 @@
...
 
     if six.PY3:
```
rares-pop: you should create a helper function that gets the config parser set up correctly, and use that from both locations.

salt/serializers/configparser.py
lomeroe: This is done b/c ``preserve_case`` isn't a valid keyword argument to the ConfigParser object, which options gets passed to when the ConfigParser is instantiated.  It needs to be taken out of the dict, which is why ``pop`` is used.

rares-pop: @lomeroe - This comment was before seeing that you do a pop. Then I suggested using the pop directly, in the next comment.

lomeroe: yeah, sorry about that, read over that one too fast or something :)

twangboy (approved):

PR #52546, opened 2019-04-15
Async ping_all_connected_minions

pengyao:
### What does this PR do?

Async ping_all_connected_minions function

### Previous Behavior

`ping_all_connected_minions` will block master `Maintenance` process for waiting `client.cmd`  finished when master key rotate and `ping_on_rotate` option set
### New Behavior

`ping_all_connected_minions` will be asynchronous  by `cmd_async`

### Tests written?

No

### Commits signed with GPG?

No

### DIFF
--- a/salt/utils/master.py
+++ b/salt/utils/master.py
@@ -775,7 +775,7 @@ def ping_all_connected_minions(opts):
     else:
         tgt = '*'
         form = 'glob'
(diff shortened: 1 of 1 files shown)


Akm0d (approved):
Tested with a minion swarm, it works :+1:

Issue #52449, opened 2019-04-08
integration.cloud.clouds.test_gce.GCETest.test_instance_extra failure


Ch3LL:
### Description of Issue/Question
The test `integration.cloud.clouds.test_gce.GCETest.test_instance_extra` is failing on 2019.2

https://jenkinsci.saltstack.com/job/2019.2/view/Cloud/job/salt-centos-7-py2-cloud/46/testReport/junit/integration.cloud.clouds.test_gce/GCETest/test_instance_extra/

Here is the failure:

```
Traceback (most recent call last):
  File "/tmp/kitchen/testing/tests/integration/cloud/clouds/test_gce.py", line 128, in test_instance_extra
    self.assertIn(self.INSTANCE_NAME, delete_str)
AssertionError: u'cloud-test-lprpnj' not found in u''
```

sagetherage:
@garethgreenaway assign you, but isn't a release blocker

Ch3LL:
im still seeing this fail repeatedly on python3 but NOT in python2

https://jenkinsci.saltstack.com/job/2019.2.1/view/Cloud/job/salt-centos-7-py3-cloud/7/

Ch3LL:
re-adding as a release blocker.

@garethgreenaway seeing this fail here still fail on python2 as well.

https://jenkinsci.saltstack.com/job/2019.2/view/Cloud/job/salt-centos-7-py2-cloud/73/#showFailuresLink

Ch3LL:
actually this is only failing on 2019.2 not 2019.2.1, so this still needs to stay open but not for 2019.2.1 release

Ch3LL:
duplicate of https://github.com/saltstack/salt/issues/53012

PR #52551, opened 2019-04-15
[2019.2] Marking the two tests in test_gce as flaky

garethgreenaway:
### What does this PR do?
Marking the two tests in test_gce as flaky.

#52449

### Commits signed with GPG?
Yes


waynew (approved):
Per some Slack discussion, this is either because:

a) AWS and GCE are playing silly beggars
b) apache-libcloud is failing to do the right thing Re: GCE
c) A race condition (that probably wouldn't be worth working around)
d) Something else
e) All of the above :trollface:

dwoz (approved):

PR #52553, opened 2019-04-15
fix: #51842

### LINKED ISSUE #51842: salt.auth.django docs use wrong model name
`salt.auth.django` doc uses:

```python
class SaltExternalAuthModel(models.Model):
    user_fk = models.ForeignKey(auth.User)
    minion_matcher = models.CharField()
    minion_fn = models.CharField()
```
as an exemple of a django model to auth users.

There's few problems:

- the actual module tries to to fetch `minion_or_fn_matcher` user attribute.

- `CharField` should have mandatory `max_length` kwarg.

- user foreign key should be:
  ```python
  from django.contrib.auth.models import User
  ...
  user_fk = models.ForeignKey(User, on_delete=models.CASCADE)
  ```

Docs should be updated to:

```python
from django.contrib.auth.models import User
class SaltExternalAuthModel(models.Model):
    user_fk = models.ForeignKey(User, on_delete=models.CASCADE)
    minion_or_fn_matcher = models.CharField(max_length=255)
    minion_fn = models.CharField(max_length=255)
```

mattLLVW:
### What does this PR do?

Fix #51842 Wrong model name
 
#51842

### DIFF
--- a/salt/auth/django.py
+++ b/salt/auth/django.py
@@ -28,9 +28,9 @@ The relevant entry in the ``models.py`` file would look like this:
 .. code-block:: python
 
     class SaltExternalAuthModel(models.Model):
-        user_fk = models.ForeignKey(auth.User)
-        minion_matcher = models.CharField()
-        minion_fn = models.CharField()
+        user_fk = models.ForeignKey(User, on_delete=models.CASCADE)
+        minion_or_fn_matcher = models.CharField(max_length=255)
+        minion_fn = models.CharField(max_length=255)
 
 The :conf_master:`external_auth` clause in the master config would then look
 like this:


salt/auth/django.py
```diff
@@ -28,9 +28,9 @@
 .. code-block:: python
 
     class SaltExternalAuthModel(models.Model):
-        user_fk = models.ForeignKey(auth.User)
-        minion_matcher = models.CharField()
-        minion_fn = models.CharField()
+        user_fk = models.ForeignKey(User, on_delete=models.CASCADE)
+        minion_or_fn_matcher = models.CharField(max_length=255)
+        minion_fn = models.CharField(max_length=255)
```
waynew: If there's an assumption of Postgres as the database, [there's no reason to add max_length](https://www.depesz.com/2010/03/02/charx-vs-varcharx-vs-varchar-vs-text/).

waynew: Ah... I see it's [required in Django's CharField](https://docs.djangoproject.com/en/2.2/ref/models/fields/#django.db.models.CharField)

waynew: Apparently one needs to [make their own `CharField` to avoid Django's limit](https://stackoverflow.com/a/20116327)

waynew (approved):

dwoz (changes requested):
@mattLLVW Is it possible to add a test for this change?

mattLLVW:
It's a documentation change...?

dwoz (approved):
@mattLLVW My bad I took a glance and it looked like a code change. Approved.

PR #52560, opened 2019-04-16
initial docs for ssh wrapper modules

mchugh19:
### What does this PR do?
Adds first draft of documentation for ssh wrapper modules.

Not sure of all content that can be included, but this at least gives a starting point and starts the documentation process. 

### Tests written?
No


mchugh19:
Refers to functionality from #52541. If there's a better way to support, documentation text and code method can be adjusted.

dwoz (approved):
@mchugh19 Looks good, thanks!

Issue #52431, opened 2019-04-06
2019.02 virt.vm_info is failing to execute due to _get_domains


tonyplovich:
### Description of Issue/Question

Looking at 2018.3, vm_info used to iterate through a list of VM names obtained from list_domains.  In 2019.02 vm_info now obtains this list from _get_domains which instead returns a list of domain objects.  When the domain object is passed into the functions referenced by vm_info._info (_get_uuid, _get_on_crash, _get_on_reboot, and _get_on_poweroff), _get_domain is eventually called again but with a domain object in *vms instead of a vm's name.  _get_domain then checks the contents of *vms against the list all_vms, but fails because the domain object (<libvirt.virDomain object at 0x7fddce314a10>) doesn't match any of the vm names.

### Setup
virt.vm_info is failing with the following trace:

> salt-minion[18311]: [ERROR   ] A command in 'virt.vm_info' had a problem: The VM "<libvirt.virDomain object at 0x7fddce314a10>" is not present
salt-minion[18311]: Traceback (most recent call last):
salt-minion[18311]: File "/usr/lib/python2.7/site-packages/salt/minion.py", line 1660, in _thread_return
salt-minion[18311]: return_data = minion_instance.executors[fname](opts, data, func, args, kwargs)
salt-minion[18311]: File "/usr/lib/python2.7/site-packages/salt/executors/direct_call.py", line 12, in execute
salt-minion[18311]: return func(*args, **kwargs)
salt-minion[18311]: File "/usr/lib/python2.7/site-packages/salt/modules/virt.py", line 2034, in vm_info
salt-minion[18311]: info[domain.name()] = _info(domain)
salt-minion[18311]: File "/usr/lib/python2.7/site-packages/salt/modules/virt.py", line 2021, in _info
salt-minion[18311]: 'uuid': _get_uuid(dom),
salt-minion[18311]: File "/usr/lib/python2.7/site-packages/salt/modules/virt.py", line 336, in _get_uuid
salt-minion[18311]: return ElementTree.fromstring(get_xml(dom)).find('uuid').text
salt-minion[18311]: File "/usr/lib/python2.7/site-packages/salt/modules/virt.py", line 2452, in get_xml
salt-minion[18311]: xml_desc = _get_domain(conn, vm_).XMLDesc(0)
salt-minion[18311]: File "/usr/lib/python2.7/site-packages/salt/modules/virt.py", line 273, in _get_domain
[...]

PR #52561, opened 2019-04-16
virt.vm_info and other functions in virt module

waheedi:
### What does this PR do?

- It fixes few bugs with the current module and sticks to the standard way of implementing virt module functions, there were inconsistencies in the implementation of few functions like get_on_poweroff, get_on_crash, get_on_reboot, get_uuid and get_profiles. 

- removed function get_xml as it does not belong here

https://github.com/saltstack/salt/issues/52431

### Previous Behavior
`salt '*' virt.vm_info`
The VM "<libvirt.virDomain object at 0x7f24e9e73198>" is not present
### New Behavior
`salt '*' virt.vm_info`
```
minion:
    ----------
    minion-1:
        ----------
        cpu:
            1
        cputime:
            0
        disks:
            ----------
            vda:
                ----------
                cluster size:
                    65536
                disk size:
                    3065716736
                file:
                    /var/lib/libvirt/images/ubuntu16.04.qcow2
                file format:
                    qcow2
                type:
                    disk
                virtual size:
                    10737418240
        graphics:
            ----------
            autoport:
                yes
            keymap:
                None
            listen:
                None
            port:
                None
            type:
                spice
        loader:
            ----------
            path:
                None
        maxMem:
            819200
        mem:
            819200
        nics:
            ----------
            52:54:00:70:b7:63:
                ----------
                address:
                    ----------
                    bus:
                        0x00
                    domain:
                        0x0000
                    function:
                        0x0
                    slot:
                        0x03
                    type:
                        pci
                mac:
                    52:54:00:70:b7:63
                model:
                    virtio
                source:
                    ----------
                    network:
                        default
                type:
                    network
        on_crash:
            destroy
        on_poweroff:
            destroy
        on_reboot:
            restart
        state:
            shutdown
        uuid:
            e6e3f990-8997-4a5e-8cb7-ea835eae4bbe
```
### Tests written?

No (but tested)

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/modules/virt.py
+++ b/salt/modules/virt.py
@@ -326,7 +326,7 @@ def _get_uuid(dom):
     '''
     Get uuid from a libvirt domain object.
     '''
-    uuid = ElementTree.fromstring(dom.XMLDesc(0)).find('uuid')
+    uuid = ElementTree.fromstring(dom.XMLDesc(0)).find('uuid').text
 
     return uuid
 
--- a/salt/modules/virt.py
+++ b/salt/modules/virt.py
@@ -2626,6 +2626,26 @@ def setvcpus(vm_, vcpus, config=False, **kwargs):
     return ret1 == ret2 == 0
 
 
+def get_xml(vm_, **kwargs):
+    '''
+    Returns the XML for a given vm
+    :param vm_: domain name
+    :param connection: libvirt connection URI, overriding defaults
+        .. versionadded:: 2019.2.0
+    :param username: username to connect with, overriding defaults
+        .. versionadded:: 2019.2.0
+    :param password: password to connect with, overriding defaults
+        .. versionadded:: 2019.2.0
+    CLI Example:
+    .. code-block:: bash
+        salt '*' virt.get_xml <domain>
+    '''
+    conn = __get_conn(**kwargs)
+    xml_desc = _get_domain(conn, vm_).XMLDesc(0)
+    conn.close()
+    return xml_desc
+
+
 def _freemem(conn):
     '''
     Internal variant of freemem taking a libvirt connection as parameter
--- a/salt/modules/virt.py
+++ b/salt/modules/virt.py
@@ -2629,15 +2629,22 @@ def setvcpus(vm_, vcpus, config=False, **kwargs):
 def get_xml(vm_, **kwargs):
     '''
     Returns the XML for a given vm
+
     :param vm_: domain name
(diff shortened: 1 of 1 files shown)


gtmanfred (changes requested):
You can't just delete function, salt maintains backwards compatiblity for at least 2 major releases, and this needs to include deprecation warnings for anything that is being deleted.

waheedi:
@gtmanfred I will re-add that function then, thanks for taking a look.

salt/modules/virt.py
```diff
@@ -324,57 +324,52 @@ def _parse_qemu_img_info(info):
...
-
-    CLI Example:
-
-    .. code-block:: bash
-
-        salt '*' virt.get_uuid <domain>
+    Get uuid from a libvirt domain object.
     '''
-    return ElementTree.fromstring(get_xml(dom)).find('uuid').text
+    uuid = ElementTree.fromstring(dom.XMLDesc(0)).find('uuid').text
```
gtmanfred: Why are all these having to be redefined?

What changed? This has worked in the past.  Do you have a reason to not use the get_xml() function that already exists, and looks like it is doing the same thing?

I want to make sure we are not breaking other peoples environments.

waheedi: Thanks @gtmanfred, I'm pretty sure this will not break other peoples environments, few reasons: 

- the previous version of the code used to reference `dom` as string a "name" while it should be an actual domain object.
- all the other _get functions are expecting a domain object not a string and many functions are raising exceptions because of treating an object as a string, then failing with `ERROR: The VM "<libvirt.virDomain object at 0x7f24e9e73198>" is not present`

take a look at this one here for example:
```
    def _info(dom):
        '''
        Compute the infos of a domain
        '''
        raw = dom.info()
        return {'cpu': raw[3],
                'cputime': int(raw[4]),
                'disks': _get_disks(dom),
                'graphics': _get_graphics(dom),
                'nics': _get_nics(dom),
                'uuid': _get_uuid(dom),
                'loader': _get_loader(dom),
                'on_crash': _get_on_crash(dom),
                'on_reboot': _get_on_reboot(dom),
                'on_poweroff': _get_on_poweroff(dom),
                'maxMem': int(raw[1]),
                'mem': int(raw[2]),
                'state': VIRT_STATE_NAME_MAP.get(raw[0], 'unknown')}
 ```
`dom` is a domain object and not a domain name string.

PR #52564, opened 2019-04-16
Simplify IPCClient and prevent corrupt messages (Turn 2)

DmitryKuzmenko:
### What does this PR do?
Since we decided to remove singletons from IPC transport I've reverted my approach implementing the subscriber pattern as a local singleton service and multiple subscribers for it because if we don't need singleton we can just be subscribers.
Also I've removed singleton from the reverted subscriber and slightly simplified the code (in comparison to the original one, before mine and @dwoz fixes)

Reverts #51963
Related to #52445

### Tests written?
Yes (I haven't reverted the tests because them are useful)

### Commits signed with GPG?
Yes

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -131,7 +131,6 @@ MOCK_MODULES = [
     'tornado.ioloop',
     'tornado.iostream',
     'tornado.netutil',
-    'tornado.queues',
     'tornado.simple_httpclient',
     'tornado.stack_context',
     'tornado.web',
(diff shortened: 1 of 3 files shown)


dwoz:
@DmitryKuzmenko I've launched two full test runs on this branch:

https://jenkinsci.saltstack.com/job/salt-centos-7-py3-dwoz/7/
https://jenkinsci.saltstack.com/job/salt-centos-7-py2-dwoz/5/

dwoz (approved):

Issue #29867, opened 2015-12-19
Not maintaining/enforcing windows passwords


paul-mulvihill:
Since the regkey bug was fixed I can now get it to create me a local user on my Windows server.
 Although, I have noticed that it will not maintain/enforce the password after that.

 snippet:
 pault:
   user.present:
     - createhome: true
     - fullname:  "pault Account"
     - password: 'Password2Q!'
     - home: 'c:\Users\pault'
     - groups:
       - Users
       - Remote Desktop Users

 I've tried changing the password to various strings and it doesn't make any difference on the server when I RDP in it's still using the original password I set.
 I've tried it with and without the ' but nothing seems to make a difference.

ssgward:
@paul-mulvihill - thanks for the report. This may be a duplicate issue, but can't find the original, currently.

cachedout:
@twangboy or @UtahDave Any ideas on this one?

paul-mulvihill:
Any thoughts? unless i'm doing something wrong this really limits the use of salt for windows users.

ssgward:
@paul-mulvihill - what is your salt version and Windows version being used?

paul-mulvihill:
@ssgward  I'm currently running 2015.8.5  both server and client
Centos Server with Windows Clients
I just double checked and created some users and then changed the passwords on the master but the state run still refers to the users as present and up to date. even though i know the passwords are out of sync.

UtahDave:
This is a known issue. We need to come up with a way for Salt to verify if the specified password is what is currently assigned in Windows. Right now Salt has no way to get the Windows password to use for comparison.

In the Linux version of this module Salt is able to compare a hash of the specified password and the hash kept in the /etc/shadow file. No such file exists on Windows. We might be able to run a function to check if the specified password authenticates correctly or not.

damon-atkins:
SaltStack would need to test the local username/password to confirm its correct, i.e. logon or change password everytime. Keep in mine windows password history, password change restrictions.    There are a few examples on the internet e.g. https://gallery.technet.microsoft.com/scriptcenter/Verify-the-Local-User-1e365545/view/Discussions#content   https://support.microsoft.com/en-us/kb/180548

No idea how to integrate them into Salt so they are fast.

Or using Desired State Configuration (DSC) #30523

paul-mulvihill:
anyone had any thoughts on this? it would be really useful to have something that can manage credentials for local windows users.

PR #52566, opened 2019-04-16
Verify password

twangboy:
### What does this PR do?
`shadow.verify_password`:
Adds a new function to `win_shadow.py` called `verify_password`. This function attempts to log on with the supplied username and password. If it is successful, it returns True. If it fails to login, it checks if the account was locked as a result of the attempt. If it is locked, it unlocks the account and returns False.

`user.present`
Uses the new `verify_password` function to detect the status of the password for the state so that it behaves similar to other operating systems that allow you to compare a hash.

Adds some tests

https://github.com/saltstack/salt/issues/29867

### Previous Behavior
There was no way to tell if the password in user.present was already set

### New Behavior
Now we can only set the password if it is incorrect

### Tests written?
Yes

### Commits signed with GPG?
Yes


damon-atkins (changes requested):
I do not think having the state check the password like this is a good idea.   In large companies these audit events are look at and reported on and investigated.  If states are run often like every 10 minutes you giving someone an opportunity to try passwords, and have the fail count reset to 0, and then try again.

twangboy:
@damon-atkins The only time this would cause a `failed logon audit event` would be if the password it's checking is wrong. But that would then cause salt to change the password. I guess if you ran the state every 10 minutes with `test=True` you would see the `failed logon audit events` every time... assuming the password is wrong.

The `failed logon attempt counter` increments each time. There's no way to reset that. Once you reach the failed attempt limit, the account is locked and salt will unlock it. I believe that resets the failed logon attempt counter. The counter is also reset when the `logon attempt timer` reaches it's limit, which is usually 30 minutes on enterprise systems, I believe.

I couldn't find any other way to validate a Windows password without logging in and triggering these events. Strangely, this is Microsofts recommended method for validating credentials.

https://support.microsoft.com/en-us/help/180548/how-to-validate-user-credentials-on-microsoft-operating-systems

Do you have any suggestions on how to better approach this? Should we let the account get locked out if someone runs the state with `test=True` 3 times within 30 minutes?
Or...
Do we continue as we are and just set the password each time the state is run (which has its own problem of constantly updating the password changed date)? I know there are folks a lot smarter than I am. I'm open to suggestions.

damon-atkins:
I suppose if salt only resets failed logon when it needs to change the password is ok. And their are no false positives.  i.e the account is locked, therefore the login fails, so salt mistakenly resets the password.   ????

twangboy:
@damon-atkins You're right... let me see what happens with a locked account.

twangboy:
@damon-atkins So, apparently, you can't verify the password at all if the account is locked. It will always return account locked. That probably makes sense from a security perspective. So, I set it up to raise a CommandExecutionError if the account is locked.

The others I tested returned `ERROR_LOGON_FAILURE` if the password was incorrect. Otherwise, they return their respective errors.

damon-atkins (approved):
Approved based on the change being made.

PR #52578, opened 2019-04-17
Try to resolve user local groups without LG_INCLUDE_INDIRECT flag

### LINKED ISSUE #52577: Salt-minion fails with "Access is denied" on Windows when run under domain account with specific privileges
Salt-minion can sometimes fail on Windows with error message `pywintypes.error: (5, 'NetUserGetLocalGroups' , 'Access is denied.')` when executed under domain account with special permissions set up.

image001

We found this issue when we tried to run the service under account in AD in Object List Mode, when the account doesn't have full visibility over the domain. During debugging how the `NetUserGetLocalGroups` function  actually works, we found out it tries to resolve all indirect groups the user is member of and it tries to resolve domain accounts as well, if the user is domain user (see [NetUserGetLocalGroups returns ERROR_ACCESS_DENIED](https://microsoft.public.win32.programmer.networks.narkive.com/YjVYeLCB/netusergetlocalgroups-returns-error-access-denied) for reference). It's because of flag [LG_INCLUDE_INDIRECT](https://docs.microsoft.com/en-us/windows/desktop/api/lmaccess/nf-lmaccess-netusergetlocalgroups), which is set by default.

The solution is explicitly disabling the `LG_INCLUDE_INDIRECT` flag, but it might be sometimes needed, so wrapping the call in try/except with call without the flag would be better.

### Setup
Windows server joined to domain.

### Steps to Reproduce Issue
Run salt-minion under domain account which doesn't have permissions to all groups it's indirectly member of.

### Versions Report
2019.2, 2018.3, 2017.7 versions are affected.

lukasraska:
### What does this PR do?
Try to resolve user groups without the LG_INCLUDE_INDIRECT flag when it fails for the first time with the flag set

Fix #52577

### Previous Behavior
Salt-minion startup fails with `Access is denied.` message.

image001

### New Behavior
Salt-minion starts up and works.

### Tests written?

No

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/utils/win_functions.py
+++ b/salt/utils/win_functions.py
@@ -84,7 +84,12 @@ def get_user_groups(name, sid=False):
         # 'win32net.NetUserGetLocalGroups' will fail if you pass in 'SYSTEM'.
         groups = [name]
     else:
-        groups = win32net.NetUserGetLocalGroups(None, name)
+        try:
+            groups = win32net.NetUserGetLocalGroups(None, name)
+        except pywintypes.error as exc:
+            # Try without LG_INCLUDE_INDIRECT flag, because the user might not have
+            # permissions for it
+            groups = win32net.NetUserGetLocalGroups(None, name, 0)
 
     if not sid:
         return groups


twangboy:
@lukasraska Since this is a bug fix, would you mind submitting it against the 2018.3 branch?

lukasraska:
Sure, created #52596 targetting 2018.3 branch

PR #52589, opened 2019-04-17
Ignore retcode when checking FS type

terminalmage:
This is used in part to check if the block device being checked actually exists. In these cases, a nonzero exit code is not an error condition, and thus should not have its results logged at the ERROR loglevel.

### DIFF
--- a/salt/states/blockdev.py
+++ b/salt/states/blockdev.py
@@ -193,5 +193,7 @@ def _checkblk(name):
     Check if the blk exists and return its fstype if ok
     '''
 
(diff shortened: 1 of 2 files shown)


dwoz (commented):
@terminalmage This seems like it might be a difficult one to test, but can you think of a way to add a regression test for this bugfix?

terminalmage:
@dwoz How about that??

twangboy:
@terminalmage Looks like you got some lint

s0undt3ch (approved):

dwoz (approved):
@terminalmage Works for me. :)

dwoz:
The docs will pass after #52622 is merged.

Issue #52605, opened 2019-04-18
Salt should not use dpkg database directly


bdrung:
The entire dpkg database, its layout and files are an internal interface and no program or package should be accessing it, other than dpkg itself and the suite of dpkg tools.

Whilst the files may be editable by an admin, that's a supported (but unrecommended) feature reserved for humans and not for automatic tools.

See https://lintian.debian.org/tags/uses-dpkg-database-directly.html

lintian complains that these two files in salt 2018.3 uses the dpkg database directly:

* salt/modules/alternatives.py
 * salt/modules/dpkg_lowpkg.py

Ch3LL:
thanks for the PR :)

bdrung:
The PR is still open.

PR #52606, opened 2019-04-18
Improve dpkg_lowpkg

bdrung:
While trying to work on fixing #52605, I stumbled over some code in `dpkg_lowpkg` that could be improved. This pull requests changes the code to use `dpkg-query` for listing packages instead of `dpkg -L`, because the output is easier to parse. Also remove some useless loop iterations and improve the documentation. Please have a look at the individual commits for details.

### DIFF
--- a/tests/unit/modules/test_dpkg_lowpkg.py
+++ b/tests/unit/modules/test_dpkg_lowpkg.py
@@ -6,6 +6,7 @@
 # Import Python libs
 from __future__ import absolute_import, print_function, unicode_literals
 
+import logging
 import os
 
 # Import Salt Libs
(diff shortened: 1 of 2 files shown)


tests/unit/modules/test_dpkg_lowpkg.py
```diff
@@ -20,8 +20,41 @@
...
+
+
 @skipIf(NO_MOCK, NO_MOCK_REASON)
 class DpkgTestCase(TestCase, LoaderModuleMockMixin):
+    def dpkg_L_side_effect(self, cmd, **kwargs):
+        self.assertEqual(cmd[:2], ["dpkg", "-L"])
+        package = cmd[2]
+        return DPKG_L_OUTPUT[package]
+
     '''
```
dwoz: Looks like the method `dpkg_L_side_effect` should be below the doc string for the test class

dwoz (changes requested):

tests/unit/modules/test_dpkg_lowpkg.py
bdrung: Done.

bdrung:
@dwoz Your remark is addressed (but Github still declares it as requested change).

dwoz (dismissed):

dwoz (approved):

PR #52624, opened 2019-04-19
fix TypeError: argument of type int is not iterable

tanlingyun2005:
### What does this PR do?
fix a batch run bug, when batch_safe_size is set to a integer or percentage,  will resulting in "TypeError: argument of type 'int' is not iterable"

### Tests written?

Yes

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/cli/batch.py
+++ b/salt/cli/batch.py
@@ -85,7 +85,8 @@ class Batch(object):
         '''
         partition = lambda x: float(x) / 100.0 * len(self.minions)
         try:
-            if isinstance(self.opts['batch'], str) and '%' in self.opts['batch']:
+            if (isinstance(self.opts['batch'], str) or isinstance(self.opts['batch'], unicode)) \
+                    and '%' in self.opts['batch']:
                 res = partition(float(self.opts['batch'].strip('%')))
                 if res < 1:
                     return int(math.ceil(res))
--- a/salt/cli/batch.py
+++ b/salt/cli/batch.py
@@ -8,6 +8,7 @@ from __future__ import absolute_import, print_function, unicode_literals
 import math
 import time
 import copy
+import sys
 from datetime import datetime, timedelta
 
 # Import salt libs
@@ -85,8 +86,11 @@ class Batch(object):
         '''
         partition = lambda x: float(x) / 100.0 * len(self.minions)
         try:
-            if (isinstance(self.opts['batch'], str) or isinstance(self.opts['batch'], unicode)) \
-                    and '%' in self.opts['batch']:
+            if sys.version_info.major == 3:
+                batch_is_str = isinstance(self.opts['batch'], str)
+            else:
+                batch_is_str = isinstance(self.opts['batch'], str) or isinstance(self.opts['batch'], unicode)
+            if batch_is_str and '%' in self.opts['batch']:
                 res = partition(float(self.opts['batch'].strip('%')))
                 if res < 1:
                     return int(math.ceil(res))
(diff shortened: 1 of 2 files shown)


Ch3LL:
@tanlingyun2005 thanks for the contribution! Looks like there are some test failures:

```
    unit.cli.test_batch.BatchTestCase.test_get_bnum_high_percentage
    unit.cli.test_batch.BatchTestCase.test_get_bnum_percentage
```

Also is it possible for you to write a regression test for this fix?

tanlingyun2005:
> @tanlingyun2005 thanks for the contribution! Looks like there are some test failures:
> 
> ```
>     unit.cli.test_batch.BatchTestCase.test_get_bnum_high_percentage
>     unit.cli.test_batch.BatchTestCase.test_get_bnum_percentage
> ```
> Also is it possible for you to write a regression test for this fix?

OK，Let me see

tanlingyun2005:
jenkins/pr/lint — Python lint test has failed

Incompatible Python 3 code found. Proposed fix:
--

because unicode type just exists in Python 2, so "isinstance(self.opts['batch'], unicode)" is not compatible to Python 3, but use "sys.version_info.major" as a conditional branch could avoid this question.

waynew:
@tanlingyun2005 there's some samples around the code of checking for bytes vs text using six - you'll probably want to follow one of those patterns.

tanlingyun2005:
> @tanlingyun2005 there's some samples around the code of checking for bytes vs text using six - you'll probably want to follow one of those patterns.

Thanks remind

Akm0d (approved):
Looks good

Ch3LL (changes requested):
would you mind writing a regression test for this so this issue does not pop up again? thanks :)

tanlingyun2005:
> would you mind writing a regression test for this so this issue does not pop up again? thanks :)

sorry, I don't know how to write "regression test", could you give me some samples?
but I add a unit test case to tests/unit/cli/test_batch.py, this is a run result:
test_get_bnum_high_percentage (__main__.BatchTestCase)
[CPU:13.3%|MEM:41.5%]  ... ok
test_get_bnum_int (__main__.BatchTestCase)
[CPU:0.0%|MEM:41.5%]  ... ok
test_get_bnum_invalid_batch_data (__main__.BatchTestCase)
[CPU:25.0%|MEM:41.5%]  ... <type 'unicode'>
ok
test_get_bnum_percentage (__main__.BatchTestCase)
[CPU:0.0%|MEM:41.5%]  ... ok
test_get_bnum_str (__main__.BatchTestCase)
[CPU:0.0%|MEM:41.5%]  ... ok

Ran 5 tests in 0.009s

OK

dwoz (approved):
@tanlingyun2005 

A 'regression test' is basically a tests that shows the bug when the tests fails. Generally, we write a regression test first before fixing the bug in question. The purpose of these types of tests is to make sure that once a bug is fixed it does not happen again in the future.

Your unit test looks good for this case. Thanks a ton!

salt/cli/batch.py
```diff
@@ -85,7 +85,7 @@ def get_bnum(self):
         '''
         partition = lambda x: float(x) / 100.0 * len(self.minions)
         try:
-            if '%' in self.opts['batch']:
+            if isinstance(self.opts['batch'], six.string_types) and '%' in self.opts['batch']:
```
pullit: Love the `six.string_types` for Python 2 and 3 compatibility.

Ch3LL (approved):

PR #52625, opened 2019-04-19
Explicitly close IPC connections

dwoz:
### What does this PR do?

Explicitly close IPC connections.

### Tests written?

No - Covered by existing tests

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/utils/decorators/__init__.py
+++ b/salt/utils/decorators/__init__.py
@@ -560,6 +560,7 @@ class _WithDeprecated(_DeprecationDecorator):
             return self._call_function(kwargs)
 
         _decorate.__doc__ = self._function.__doc__
+        _decorate.__wrapped__ = self._function
         return _decorate
 
 
--- a/salt/fileserver/hgfs.py
+++ b/salt/fileserver/hgfs.py
@@ -558,13 +558,13 @@ def update():
 
     # if there is a change, fire an event
     if __opts__.get('fileserver_events', False):
-        event = salt.utils.event.get_event(
+        with salt.utils.event.get_event(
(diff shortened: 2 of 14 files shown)


waynew (approved):

salt/utils/decorators/__init__.py
```diff
@@ -560,6 +560,7 @@ def _decorate(*args, **kwargs):
             return self._call_function(kwargs)
 
         _decorate.__doc__ = self._function.__doc__
+        _decorate.__wrapped__ = self._function
```
s0undt3ch: `__wrapped__` is added on Py3 by `functools.wraps`, want to check for that?

s0undt3ch (changes requested):
> New in version 3.2: Automatic addition of the `__wrapped__` attribute.
> New in version 3.2: Copying of the `__annotations__` attribute by default.
> Changed in version 3.4: The `__wrapped__` attribute now always refers to the wrapped function, even if that function defined a `__wrapped__` attribute. (see bpo-17482)

salt/utils/decorators/__init__.py
dwoz: Nope, wrapped returns a partial. The `__wrapped__` attribute get's added when the partial is called. This adds a `__wrapped__` attribute to that partial so the test can detect it and behave appropriately.

https://github.com/python/cpython/blob/master/Lib/functools.py#L74

s0undt3ch: :+1:

s0undt3ch (approved):

Issue #51008, opened 2018-12-29
cwd option not working with cmd.run and runas


cdalvaro:
### Description of Issue

`cwd` option from `cmd.run` state is not working in combination with `runas` option (on macOS at least)

I have arrived to this problem through the `git.latest` and `git.cloned` states.

These states are failing with the following comment:

> fatal: not a git repository (or any of the parent directories): .git

(Possible related issues #43185 and #586)

#### SLS recipe

```yaml
running as root:
  cmd.run:
    - name: pwd
    - cwd: /Users/Carlos/Desktop

running as user:
  cmd.run:
    - name: pwd
    - cwd: /Users/Carlos/Desktop
    - runas: Carlos

powerlevel9k present:
  git.latest:
    - name: https://github.com/bhilburn/powerlevel9k.git
    - target: /Users/Carlos/.oh-my-zsh/custom/themes/powerlevel9k
    - user: Carlos
    - require:
      - oh-my-zsh present
```

#### Output

```
----------
          ID: running as root
    Function: cmd.run
        Name: pwd
      Result: True
     Comment: Command "pwd" run
     Started: 16:12:18.764387
    Duration: 9.628 ms
     Changes:
              ----------
              pid:
                  73618
              retcode:
                  0
              stderr:
              stdout:
                  /Users/Carlos/Desktop
----------
          ID: running as user
    Function: cmd.run
        Name: pwd
      Result: True
     Comment: Command "pwd" run
     Started: 16:12:18.774467
    Duration: 55.428 ms
     Changes:
              ----------
              pid:
                  73619
              retcode:
                  0
              stderr:
              stdout:
                  /Users/Carlos
----------
          ID: powerlevel9k present
    Function: git.latest
        Name: https://github.com/bhilburn/powerlevel9k.git
      Result: False
     Comment: fatal: not a git repository (or any of the parent directories): .git
     Started: 16:12:18.831065
    Duration: 1908.322 ms
     Changes:
```

### Versions Report
(Provided by running `salt --versions-report`. Please also mention any differences in master/minion versions.)

```
Salt Version:
           Salt: 2018.3.3

Dependency Versions:
           cffi: 1.11.5
       cherrypy: unknown
       dateutil: 2.4.2
      docker-py: 1.10.6
          gitdb: 2.0.5
      gitpython: 2.1.11
          ioflo: 1.7.5
         Jinja2: 2.8
        libgit2: 0.27.7
        libnacl: 1.6.1
[...]

PR #52632, opened 2019-04-19
Prevent shell injection as root with cmd.run on macOS

cdalvaro:
### What does this PR do?

This commit prevents shell injection as `root` with `cmd.run` on macOS.

Also this commit ensures that the command is executed only if the cwd
directory exists.

#51008

### Previous Behavior

Since this command is executed as root for launching the command as a different user, code injection is possible:

```sh
sudo salt-call cmd.run \
    cmd='echo ''I am $(whoami). I have full privileges inside this subshell.''' \
    cwd='/tmp' runas='Carlos'
local:
    I am root. I have full privileges inside this subshell.
```

### New Behavior

This commit fix this security issue:

```sh
sudo salt-call cmd.run \
    cmd='echo ''I am $(whoami). I can only do what $(whoami) can.''' \
    cwd='/tmp' runas='Carlos'
local:
    I am Carlos. I can only do what Carlos can.
```

### Tests written?

Yes

### Commits signed with GPG?

Yes


twangboy (approved):

twangboy:
@cdalvaro Looks like you got some failing tests.

dwoz (changes requested):
@cdalvaro Please include a test for this change.

twangboy:
These failing tests are being addressed here: https://github.com/saltstack/salt/pull/52635

cdalvaro:
@dwoz I'll prepare a test for this change, but also I want to test this change with `salt.states.git` because I think this change breaks the `git latest` state.

cdalvaro:
I have seen that each element of the command array is escaped with [`shlex_quote`](https://kite.com/python/docs/six.moves.shlex_quote) function.

https://github.com/saltstack/salt/blob/189a8227adeca896b43615acce5b784e81b3e4ad/salt/modules/cmdmod.py#L60

https://github.com/saltstack/salt/blob/189a8227adeca896b43615acce5b784e81b3e4ad/salt/modules/cmdmod.py#L416-L418

This function returns an escaped string with single quotes making it save to be executed on a shell.

The problem now is that having changed the double quotes by single quotes has broken this escaping system, so know we would need to replace the single quotes used for escaping the command by double quotes.

As an example, with the current code this is what happens when you try to update a git repository using `git.latest`, the command executed for updating the repository is:

```sh
[INFO    ] Executing command ['git', 'for-each-ref', '--format', '%(refname:short)', 'refs/heads/'] as user 'Carlos' in directory '/Users/Carlos/.vim/pack/git-plugins/start/salt'
[DEBUG   ] CMD: su -l Carlos -c 'cd /Users/Carlos/.vim/pack/git-plugins/start/salt && git for-each-ref --format '%(refname:short)' refs/heads/'
[ERROR   ] Command '['su', '-l', 'Carlos', '-c', 'cd /Users/Carlos/.vim/pack/git-plugins/start/salt && git for-each-ref --format %(refname:short) refs/heads/']' failed with return code: 1
[ERROR   ] stderr: zsh:1: missing end of string
```

The command breaks at `--format`, one solution to this problem could be replacing the single quotes by double quotes:

```sh
[INFO    ] Executing command ['git', 'for-each-ref', '--format', '%(refname:short)', 'refs/heads/'] as user 'Carlos' in directory '/Users/Carlos/.vim/pack/git-plugins/start/salt'
[DEBUG   ] CMD: su -l Carlos -c 'cd /Users/Carlos/.vim/pack/git-plugins/start/salt && git for-each-ref --format "%(refname:short)" refs/heads/'
[DEBUG   ] stdout: master
```

cdalvaro:
ping @dwoz

cdalvaro:
Please @dwoz, tell me if more changes are required

cdalvaro:
Is there any problem with this PR?

@dwoz, @twangboy

cdalvaro:
> @cdalvaro Please include a test for this change.

@dwoz, I did your requirement.

twangboy (approved):

cdalvaro:
@dwoz, Will this PR be rejected in the end?

cdalvaro:
With all my respect, this is not serious, @dwoz 

I opened this PR one month and a half ago and you requested to me some changes.
I performed that changes as soon as I could and since then you have not answered to any of my questions.

If this PR is not going to be approved, it's OK, but I would like to finish it and think in another thing.

Thank you for your time anyway.

salt/modules/cmdmod.py
```diff
@@ -414,8 +414,10 @@ def _get_stripped(cmd):
         # method doesn't work properly when run as root for certain commands.
         if isinstance(cmd, (list, tuple)):
             cmd = ' '.join(map(_cmd_quote, cmd))
+            # This patch is explained on https://github.com/saltstack/salt/pull/52632#issuecomment-485046644
+            cmd = cmd.replace("'", '"')
 
-        cmd = 'su -l {0} -c "cd {1}; {2}"'.format(runas, cwd, cmd)
+        cmd = "su -l {0} -c 'cd {1} && {2}'".format(runas, cwd, cmd)
```
damon-atkins: A better fix would not to have this run by a shell.  I suspect we have 
`sh -c "su -l {0} -c 'cd {1} && {2}'"` happening.
The code should assume the su will invoke the shell and skip prefix with `sh -c` some where later in the code.

cdalvaro: This is a good point, unfortunately I don't have too much time to investigate this in a short time.

If you prefer not to approve this merge I will understand it, although this change solves a security bug.

cdalvaro:
I don't know where does this error comes from:

> Session lint-salt raised exception UnicodeEncodeError('ascii', u'Carlos D. \xc1lvaro', 10, 11, 'ordinal not in range(128)')

Since I have changed my last name from *Álvaro* to *Alvaro* in all occurrences.

waynew:
Hey @cdalvaro quick question - I just tried this:

```
sudo touch /tmp/test.txt
sudo chmod 700 /tmp/test.txt
sudo salt-call --local cmd.run cmd='touch /tmp/test.txt' cwd='/tmp' runas='Wayne'
```

This failed with ```local:
    touch: /tmp/foo.txt: Permission denied
```

Does that also fail for you? This is on 2019.2, FWIW. From what I can gather it's really only reporting the wrong whoami (which is bad enough), but isn't (im)properly giving the shell wrong permissions.

Definitely 👍 for the fix, just want to double check the severity of the problem.

**Comment by cdalvaro** [comment]
Dear @waynew, thank you very much for your time, based on your example I'll try to explain the bug I'm trying to fix.

**Initial setup**
```sh
sudo touch /tmp/test.txt
sudo chmod 700 /tmp/test.txt

ll /tmp/test.txt
Permissions Size User Group Date Modified Name
.rwx------     0 root wheel 26 Jun  0:38  /tmp/test.txt
```

**salt 2019.2 unfixed**

Now, with the unfixed version of salt 2019.2 I execute the following command:

```sh
sudo salt-call --local cmd.run cmd='$(chmod 644 /tmp/test.txt)' cwd='/tmp' runas='Carlos'
local:
```
 salt finished OK and this is what I get:

```sh
ll /tmp/test.txt
Permissions Size User Group Date Modified Name
.rw-r--r--     0 root   wheel 26 Jun  0:38  test.txt
```

**salt 2019.2 fixed**

If I restore the permissions to `700` and try again this time with the fix applied, this is the output:

```sh
sudo salt-call --local cmd.run cmd='$(touch /tmp/test.txt)' cwd='/tmp' runas='Carlos'
local:
    chmod: Unable to change file mode on /tmp/test.txt: Operation not permitted

ll /tmp/test.txt
Permissions Size User Group Date Modified Name
.rwx------     0 root wheel 26 Jun  0:38  /tmp/test.txt
```

Basically, the command is being interpolated by the `root` user instead of being called from the user's shell.

Please, don't hesitate if you need more information about the fix.

damon-atkins (changes requested):
Fixing the call so cmd is a list and not a string is best option.   https://github.com/saltstack/salt/pull/52632/files/c3109a994adb1b1dae84d1031f325883475ad993#pullrequestreview-244966197

waynew:
Ah! Gotcha! It appears that this is strictly limited to '$(...)' within the `cmd` parameter - at least I tried creating this `/tmp/change_perms.sh`:

```
#!/bin/sh
$(chmod 644 /tmp/test.txt)
```

Running with `cmd='/tmp/change_perms.sh' runas='Me'` failed with permissions, while `cmd='$(/tmp/change_perms.sh)' runas='Me'` succeeded. (Just trying to fully understand the impact of the problem and the fix 👍 )

cdalvaro:
> Ah! Gotcha! It appears that this is strictly limited to '$(...)' within the `cmd` parameter - at least I tried creating this `/tmp/change_perms.sh`:
> 
> ```
> #!/bin/sh
> $(chmod 644 /tmp/test.txt)
> ```
> 
> Running with `cmd='/tmp/change_perms.sh' runas='Me'` failed with permissions, while `cmd='$(/tmp/change_perms.sh)' runas='Me'` succeeded. (Just trying to fully understand the impact of the problem and the fix 👍 )

Yes, the hack is due to the string interpolation of the shell, it does not affect to scripts.

I will try to make the changes that @damon-atkins suggests!

cdalvaro:
I have made some changes.

Now, instead of mixing *double* and *single* quotes, the method `_cmd_quote` (that delegates in `shlex.quote`) is used to properly format the `cmd` string escaping quotes when needed and avoiding the possibility of using shell injection commands.

Following the example proposed by @waynew, this is the output when trying to change the permissions of the *test.txt* file:

```sh
sudo salt-call --local cmd.run cmd='$(chmod 644  /tmp/test.txt)' cwd='/tmp' runas='Carlos'

[ERROR   ] Command '$(chmod 644  /tmp/test.txt)' failed with return code: 1
[ERROR   ] output: chmod: Unable to change file mode on /tmp/test.txt: Operation not permitted
local:
    chmod: Unable to change file mode on /tmp/test.txt: Operation not permitted
```

Also, with these changes, all modules that use the `cmd.run` method won't have to worry about which quotes to use in their commands, since `_cmd_quote` will handle them properly.

salt/modules/cmdmod.py
cdalvaro: Finally I have used `_cmd_quote` to prevent shell injection. (`_cmd_quote` is an alias for `shlex.quote` which handles single quotes properly for shell commands)

cdalvaro:
> Fixing the call so cmd is a list and not a string is best option. https://github.com/saltstack/salt/pull/52632/files/c3109a994adb1b1dae84d1031f325883475ad993#pullrequestreview-244966197

For some reason, the conversion of the `list` to the `string` was done previously on purpose:

https://github.com/saltstack/salt/blob/189a8227adeca896b43615acce5b784e81b3e4ad/salt/modules/cmdmod.py#L415-L416

I would rather not to change it.

cdalvaro:
ping @waynew @damon-atkins

salt/modules/cmdmod.py
```diff
@@ -415,7 +415,8 @@ def _get_stripped(cmd):
         if isinstance(cmd, (list, tuple)):
             cmd = ' '.join(map(_cmd_quote, cmd))
 
-        cmd = 'su -l {0} -c "cd {1}; {2}"'.format(runas, cwd, cmd)
+        cmd = 'cd {0} && {1}'.format(_cmd_quote(cwd), cmd)
```
damon-atkins: What us line 411 trying to do. Seems there is two sections for Darwin.  Is this code it the right spot (not your fault)
If you look at this function eventual for all platforms except for Windows, this is turned into a list.
cmd = [ 'su', '-l', runas, '-c', 'cd \'{0}\' && {1}'.format(_cmd_quote(cwd), cmd) ]

damon-atkins (changes requested):

salt/modules/cmdmod.py
cdalvaro: I understand your proposal and it seems reasonable to me, but I am not sure if `cmd` will be properly quoted, since it could contain single quotes that potentially can break the `-c` part of the command.

By doing this way:

```python
cmd = 'cd {0} && {1}'.format(_cmd_quote(cwd), cmd)
cmd = 'su -l {0} -c {1}'.format(runas, _cmd_quote(cmd))
```

`_cmd_quote` should ensure that, whatever the `cmd` content is, it will be properly escaped. Because previously `cmd` has been converted into a string with all its components escaped.

cdalvaro:
It is a bit frustrating that after three months this PR stills unmerged. And the worst thing is that comes to solve a security hole.

I have seen PR #54079 which improves mine, so maybe my PR can be discarded now.

I hope that @ScoreUnder will be luckier than me and get its PR merged soon.

waynew:
@cdalvaro Apologies - I understand the frustration!

We're currently undergoing an intense process to pay down the technical debt that we've accumulated around our build process - the unfortunate side effect of that is that some of the PRs (like this) end out taking longer to merge than they aught. I believe there will be some more detailed communication during the [Salt Office Hours](https://calendar.google.com/calendar/event?eid=YzRwajRvYjFjOG9tNmJiNGNoaG1jYjlrY2dwajBiOXBjY3NqMGI5bzcwcDY2ZDFoNmtzNmFkcjM2Y18yMDE5MDgwNlQxNTAwMDBaIHNhbHRzdGFjay5jb21fbWQ3M2MzdWZjczJlcWJzbW1uaWtlNGVtODBAZw) about those changes, but the ultimate effect will be that our merge process is much smoother and quicker.

Unfortunately we have to slow a lot of things down to get there. "Pardon our dust", as the saying goes.

cdalvaro:
OK, I hope the review process improves in the future.

I close now this PR since #54079 includes these changes and it looks like it has more attention than this one.

PR #52633, opened 2019-04-19
Update PR Template with tests requirement

Ch3LL:


twangboy (approved):

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -9,6 +9,8 @@ Remove this section if not relevant
 Remove this section if not relevant
 
 ### Tests written?
+**[NOTICE] Bug fixes or features added to Salt require tests.**
+Please review the [test documentation](https://docs.saltstack.com/en/latest/topics/tutorials/writing_tests.html) for details on how to implement tests into Salt's test suite.
```
waynew: Should we throw this and the other bits into `<!-- ... -->` so they don't show up in the PR body?

Ch3LL: I think its fine if its added into the PR body. It only provides more information and does no harm. Additionally, the user would not be able to preview the additional information using the `preview` tab, so the URL and bold would not show up either.

s0undt3ch (approved):

dwoz:
re-run debian
re-run ubuntu

dwoz (approved):

PR #52571, opened 2019-04-17
Merge forward 2019.3 to develop

DmitryKuzmenko:
### What does this PR do?
Here's Merge forward from 2019.3 + #52564 to develop branch.

### What issues does this PR fix or reference?
Merge forward!

PR #52635, opened 2019-04-19
Update the cmdmod test according to #49886 changes.

DmitryKuzmenko:
### What does this PR do?
Fixes these tests failures in #52571 and develop
```
integration.modules.test_cmdmod.CMDModuleTest.test_exec_code
integration.modules.test_cmdmod.CMDModuleTest.test_exec_code_with_multiple_args
integration.modules.test_cmdmod.CMDModuleTest.test_exec_code_with_single_arg
```

1. Don't yamlify  in `parse_input` twice on client subcalls. `parse_input` yamlifies
the data that corrupts it if yamlified twice.
2. Update the test to pass the data that could be yamlified to
multiline.

Fixes issue introduced in #49886
Fixes tests for merge forward #52571


DmitryKuzmenko:
This fix breaks other tests. I've analyzed that and thinking on a better patch.
The problem is that we need to keep all the handling as is, but yamling multiline strings (containing `\n`) removes one `\n` per yaml run. So for instance code `print('asdf')\n\nprint('asdf')` passed as an argument to `run_job` will send `print('asdf')\nprint('asdf')` to master while the same code passed to `cmd` will send `print('asdf') print('asdf')` to master that is incorrect python code.
Any thoughts a very appreciated.
I'll work it around on Monday.

twangboy (changes requested):
Please resolve the conflicts

DmitryKuzmenko:
@twangboy this is already fixed by #52709. Just forgot to drop this one. Thank you for bump.

Issue #43409, opened 2017-09-08
Unclear documentation for mysql returner/Wrong information provided


b3hni4:
### Description of Issue/Question
Unclear documentation for mysql returner

In salt 2016.11.7, mysql returner was updated so it purges old data. In mysql returner documentation it is stated that "Should you wish the returner data to be cleaned out every so often, set keep_jobs to the number of hours for the jobs to live in the tables. **Setting it to 0 or leaving it unset will cause the data to stay in the tables.**"
https://docs.saltstack.com/en/2016.11/ref/returners/all/salt.returners.mysql.html

To me this implies that by default data is **not** cleaned. 
This leads to loosing all returner data, but for the previous 24 hours. 
After digging into this, noticed that default value in saltmaster configuration is set to 24.
https://github.com/saltstack/salt/blob/2016.11/salt/config/__init__.py 

It is expected when such change is applied, at least the documentation is more clear and provides correct information.

### Setup
(Please provide relevant configs and/or SLS files (Be sure to remove sensitive info).)

### Steps to Reproduce Issue
(Include debug logs if possible and relevant.)

### Versions Report
(Provided by running `salt --versions-report`. Please also mention any differences in master/minion versions.)
```shell
Salt Version:
           Salt: 2016.11.7

Dependency Versions:
           cffi: Not Installed
       cherrypy: Not Installed
       dateutil: 1.5
      docker-py: Not Installed
          gitdb: 0.5.4
      gitpython: 0.3.2 RC1
          ioflo: Not Installed
         Jinja2: 2.7.2
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: 0.9.1
   msgpack-pure: Not Installed
 msgpack-python: 0.4.6
   mysql-python: 1.2.3
      pycparser: Not Installed
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: Not Installed
         Python: 2.7.6 (default, Oct 26 2016, 20:30:19)
   python-gnupg: Not Installed
         PyYAML: 3.10
          PyZMQ: 14.0.1
           RAET: Not Installed
          smmap: 0.8.2
        timelib: Not Installed
        Tornado: 4.2.1
            ZMQ: 4.0.4

System Versions:
           dist: Ubuntu 14.04 trusty
        machine: x86_64
        release: 4.4.0-93-generic
         system: Linux
        version: Ubuntu 14.04 trusty
```
@tn8or
@RWejlgaard
[...]

PR #52638, opened 2019-04-19
Corrected documentation to state default keep_job param is set to 24 and not 0.

xeacott:
### What does this PR do?
This PR changes MySQL's documentation to state that the default setting for keep_job is in fact set to 24 and not to 0.

This addresses #43409  

### Tests written?

No

### Commits signed with GPG?

Yes


cachedout (approved):

dwoz (changes requested):
@xeacott The linter and docs builds are unhappy with this change. Please have a look.

xeacott:
@dwoz Looks like both lint and docs are failing 
16:18:52 Session docs raised exception UnicodeEncodeError('ascii', u'Corrected documentation to state default keep_job param is set to 24 \u2026', 69, 70, 'ordinal not in range(128)'), which is the triple dots in the title of the PR to continue the comment. I've never seen a doc fail for that reason, but would that be the case?

Ch3LL (approved):

Ch3LL (changes requested):
just realized this is posted to the 2017.7 branch. Can you add this to the 2018.3 branch instead since >=2018.3 branches are the current branches.

xeacott:
@Ch3LL Yes!

Ch3LL (approved):

dwoz (approved):

Issue #52648, opened 2019-04-20
DDNS Module logging is subpar


DeForce:
### Description of Issue/Question
Debugging DDNS module is meaning going to source code and go line by line on what happens. There is no logging implemented for it, so it's really not easy to debug what is happening and what is wrong.

No logging whatsoever, only thing that is happening is a return boolean (True, False, None)

DeForce:
I have created PR #52649 to address this issue.

Ch3LL:
thanks for the PR! that will be a great improvement :)

PR #52649, opened 2019-04-20
Adding logging to DDNS Module

DeForce:
### What does this PR do?
This PR adds logging to DDNS Module so debugging is easier

#52648 

### Previous Behavior
No logging whatsoever, only thing that is happening is a return boolean (True, False, None)

### New Behavior
Now we have varying degrees of logging. Warning, Info and Debug logging is implemented.

### Tests written?
No

### Commits signed with GPG?
No


salt/modules/ddns.py
```diff
@@ -191,16 +202,25 @@ def update(zone, name, ttl, rdtype, data, nameserver='127.0.0.1', timeout=5,
                     is_exist = True
                     break
 
+    log.debug('record exists: %s', is_exist)
```
dwoz: A bit of a nit-pick but can you make the `r` in `record` upper case please?

salt/modules/ddns.py
```diff
@@ -191,16 +202,25 @@ def update(zone, name, ttl, rdtype, data, nameserver='127.0.0.1', timeout=5,
...
     dns_update = dns.update.Update(zone, keyring=keyring, keyname=keyname,
                                    keyalgorithm=keyalgorithm)
     if replace:
+        log.info('Replacing record %s', name)
         dns_update.replace(name, ttl, rdata)
     elif not is_exist:
+        log.info('Creating record %s', name)
         dns_update.add(name, ttl, rdata)
     else:
+        log.info('Not doing anything, record exists and in correct state', name)
```
dwoz: There is no string format for `name` here.

dwoz (changes requested):

salt/modules/ddns.py
DeForce: Yup! My bad :)

salt/modules/ddns.py
DeForce: Now there is!

DeForce:
@dwoz Changes are in, please check again! :)

oeuftete:
I've re-submitted this against master in https://github.com/saltstack/salt/pull/58634.

Issue #52650, opened 2019-04-20
DDNS Module configuration structure.


DeForce:
### Description of Issue/Question
Currently, DDNS Module configuration documentation is not really understandable to what you need to configure in the minion configuration to run this execution module with TSIG authentication.

Documentation: https://docs.saltstack.com/en/latest/ref/modules/all/salt.modules.ddns.html
looks like documentation specifies how you configure it using CLI, not minion configuration:
```
keyfile: keyring file (default=None)
keyname: key name in file (default=None)
...
```

You can configure module using minion configuration which is not documented and looks like:
```
ddns.keyfile: ...
ddns.keyname: ...
```

Browsing Source code I have found that it is using module `config.option` which doesn't know how to traverse a dict.

I think that current way of configuring is oldschool and we can leverage YAML and nested structure to be like that:
```
ddns:
  keyfile: something
  keyname: something
...
```

DeForce:
PR for the change #52651

PR #52651, opened 2019-04-20
DDNS configuration layout change

DeForce:
### What does this PR do?
Implements different way to store configuration for execution module.
Yes, this adds one line to store config, but it makes it look like all other configuration structures.

#52650 

### Previous Behavior
Minion configuration file snippet:
```
ddns.keyfile: ...
ddns.keyname: ...
```

### New Behavior
Minion configuration file snippet:
```
ddns:
  keyfile: something
  keyname: something
```

### Tests written?
No

### Commits signed with GPG?
No

### DIFF
--- a/salt/modules/ddns.py
+++ b/salt/modules/ddns.py
@@ -7,13 +7,13 @@ Support for RFC 2136 dynamic DNS updates.
     are a couple of optional configuration parameters made available to
     support this (the keyname is only needed if the keyring contains more
     than one key)::
-
-        keyfile: keyring file (default=None)
-        keyname: key name in file (default=None)
-        keyalgorithm: algorithm used to create the key
-                      (default='HMAC-MD5.SIG-ALG.REG.INT').
-            Other possible values: hmac-sha1, hmac-sha224, hmac-sha256,
-                hmac-sha384, hmac-sha512
+        ddns:
+          keyfile: keyring file (default=None)
+          keyname: key name in file (default=None)
+          keyalgorithm: algorithm used to create the key
+                        (default='HMAC-MD5.SIG-ALG.REG.INT').
+              Other possible values: hmac-sha1, hmac-sha224, hmac-sha256,
+                  hmac-sha384, hmac-sha512
 
 
     The keyring file needs to be in json format and the key name needs to end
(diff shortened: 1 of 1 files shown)


salt/modules/ddns.py
```diff
@@ -7,13 +7,13 @@
...
     support this (the keyname is only needed if the keyring contains more
     than one key)::
-
-        keyfile: keyring file (default=None)
-        keyname: key name in file (default=None)
-        keyalgorithm: algorithm used to create the key
-                      (default='HMAC-MD5.SIG-ALG.REG.INT').
-            Other possible values: hmac-sha1, hmac-sha224, hmac-sha256,
-                hmac-sha384, hmac-sha512
+        ddns:
```
Ch3LL:
```suggestion
        ddns.keyfile: keyring file (default=None)
```

Is there a reason not to just change the docs to add `ddns` before each one?  Seems this was the initial intent. And if we change it to `ddns:` instead then it would need to be properly deprecated, because any user using this module would need to update their configuration.

Ch3LL (changes requested):

salt/modules/ddns.py
DeForce: The reason I have opted to do nested dict, is that I have not seen any module except this one use the `module_name.key: value` instead of `module_name: {key1: value}`.

I can make it supported both ways, and send a deprecation message about the "old" dot style, if you are fine with that?

Ch3LL: there are a lot of other modules that use this same format:

https://github.com/saltstack/salt/blob/v2019.2.0/salt/states/boto_lc.py#L34
https://github.com/saltstack/salt/blob/v2019.2.0/salt/states/boto_s3.py#L24
https://github.com/saltstack/salt/blob/v2019.2.0/salt/modules/redismod.py#L12

just to name a few. Do you see any other reason, since this is indeed used throughout salt? 

If there is, this is how you properly deprecate: https://docs.saltstack.com/en/latest/topics/development/deprecations.html

Ch3LL:
@DeForce following up here. One thing we could do is support both configurations so we don't break it for current users. Something like `value = __salt__['config.get']('ddns.{0}'.format(key)) or __salt__['config.get']('ddns:{0}'.format(key))`  This would require test coverage for this change to make sure both use cases are covered. Are you able to make this change and add the tests?

Ch3LL:
Also I forgot to mention that this will need to be migrated to the `master` branch and a changelog entry added. https://docs.saltproject.io/en/latest/topics/development/changelog.html

sagetherage:
@DeForce any follow up here to rebase this against master?

sagetherage:

PR #52665, opened 2019-04-23
Tidy "Changes made" output in git states

ScoreUnder:
### What does this PR do?

Fixes a cosmetic issue in comment output of git-related states.

None as far as a cursory search shows

### Previous Behavior

Sample excerpt from `git.latest` output:
`Changes made: /. h. o. m. e. /. g. i. t. /. r. e. p. o. s. i. t. o. r. i. e. s. /. s. a. l. t. -. c. o. n. f. i. g. .. g. i. t.  . w. a. s.  . f. e. t. c. h. e. d. ,.  . r. e. s. u. l. t. i. n. g.  . i. n.  . u. p. d. a. t. e. d.  . r. e. f. s.`

### New Behavior

`Changes made: /home/git/repositories/salt-config.git was fetched, resulting in updated refs`

### Tests written?

No

Not sure how to test for this either

### Commits signed with GPG?

No

### DIFF
--- a/salt/states/git.py
+++ b/salt/states/git.py
@@ -163,7 +163,7 @@ def _uptodate(ret, target, comments=None, local_changes=False):
         # report on them so we are alerted to potential problems with our
         # logic.
         ret["comment"] += "\n\nChanges {0}made: {1}".format(
-            "that would be " if __opts__["test"] else "", _format_comments(comments)
+            "that would be " if __opts__["test"] else "", comments
         )
     return ret
 
--- a/salt/states/git.py
+++ b/salt/states/git.py
@@ -152,6 +152,8 @@ def _strip_exc(exc):
 
 
 def _uptodate(ret, target, comments=None, local_changes=False):
+    assert comments is None or isinstance(comments, six.string_types)
+
     ret["comment"] = "Repository {0} is up-to-date".format(target)
     if local_changes:
         ret["comment"] += (
--- a/tests/unit/states/test_git.py
+++ b/tests/unit/states/test_git.py
@@ -73,3 +73,86 @@ class GitTestCase(TestCase, LoaderModuleMockMixin):
             )
             assert result["result"] is True, result
             git_diff.assert_not_called()
+
+    def test_uptodate_message_format_gen(self):
+        fetch_url = "https://xyz.invalid/salt/unit/test/repo.git"
+        local_path = "/local/salt/unit/test/repo.git"
+        branch = "master"
+        head_rev = "cafe0000cafe0000cafe0000cafe0000cafe0000"
+        remote_name = "salt_test_remote"
+
+        tags = []
+        remote_refs = {
+            "HEAD": head_rev,
+            "refs/heads/" + branch: head_rev
+        }
+
+        def rev_parse(cwd, rev, *args, **kwargs):
+            assert cwd == local_path
+
+            if "--abbrev-ref" in kwargs.get("opts", []) and rev.endswith("@{upstream}"):
+                return remote_name + "/" + rev.split("@")[0]
+
+            rev = rev.split("^")[0]
+            return remote_refs.get(rev) \
+                or remote_refs.get("refs/heads/" + rev) \
+                or rev
+
+        def ls_remote(cwd, remote, opts, *args, **kwargs):
+            assert cwd == local_path
+            assert remote in [remote_name, fetch_url]
+
+            if opts == "--tags":
+                return tags
+            raise NotImplementedError()
+
+        def merge_base(cwd, refs, is_ancestor=False, **kwargs):
+            assert cwd == local_path
+            assert is_ancestor
+
+            return all(ref == refs[0] for ref in refs)
+
+        dunder_salt = {
+            "git.config_get_regexp": MagicMock(return_value={}),
+            "git.remote_refs": MagicMock(return_value=remote_refs),
+            "git.version": MagicMock(return_value="1.8.3.1"),
+            "git.list_branches": MagicMock(return_value=[branch]),
+            "git.list_tags": MagicMock(return_value=tags),
+            "git.revision": MagicMock(return_value=head_rev),
+            "git.current_branch": MagicMock(return_value=branch),
+            "git.remotes": MagicMock(return_value={
+                remote_name: {
+                    "fetch": fetch_url,
+                },
+            }),
+            "git.diff": MagicMock(return_value=""),
+            "git.rev_parse": MagicMock(side_effect=rev_parse),
+            "git.ls_remote": MagicMock(side_effect=ls_remote),
+            "git.fetch": MagicMock(return_value=True),
+            "git.merge_base": MagicMock(side_effect=merge_base),
+        }
+
+        # Things that would interact with FS on test host
+        isfile_mock = MagicMock(return_value=False)
+        # expanduser reads home directory from system
+        expanduser_mock = MagicMock(side_effect=NotImplementedError)
+        isdir_mock = MagicMock(return_value=True)
+
+        with patch.dict(git_state.__salt__, dunder_salt), \
+                patch("os.path.isfile", isfile_mock), \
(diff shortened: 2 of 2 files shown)


dwoz (commented):
@ScoreUnder I wonder if it would be better to detect if we are dealing with a list type in `_format_comments`. The thought being that there was a reason that comments were being messaged before and the case for that reason might still exist in some contexts.

waynew:
@ScoreUnder We've changed our development process - could you rebase this on `master` branch, and address dwoz's comment? 

We're also requiring tests, so a recommended approach would be to just write some tests & update `_format_comments` to return the appropriate data whether it's list or string.

Does that make sense? If you need pointers on writing tests, we're hosting a test clinic on Thursday, April 2 during our community open hour. Our community event calendar can be found [here](https://calendar.google.com/calendar/embed?src=saltstack.com_md73c3ufcs2eqbsmmnike4em80%40group.calendar.google.com) - you can add to your own Google Calendar by clicking the _totally intuitive_ + at the bottom right of the screen.

ScoreUnder:
I held off on addressing this because I disagree somewhat strongly with the amendment proposed, but here are my current thoughts on the matter.

> I wonder if it would be better to detect if we are dealing with a list type in `_format_comments`

This is fundamentally just a confusion of types, so I think `comments: str` vs `comments: List[str]` would clear up this confusion, though as far as I know that isn't done in salt.

It does feel a little off to me to be making a list-to-string function handle the string case (outside of just throwing an exception), especially when this is currently calling `_format_comments` directly on the return value of `_format_comments` from higher up in the call stack, which is not a semantically sensible thing to be doing.

As a less minimal solution but still without confusing the purpose of `_format_comments`, perhaps I could remove the calls to `_format_comments` at the call sites instead, passing comments as a list into this function.

That would prevent `_format_comments` from becoming a red herring for people debugging the code, and it would prevent it from becoming dead code in future (with only the no-op branch taken) without anyone noticing, as well as avoiding the semantic contradiction alluded to above.

Regarding the assertion that this poses a red herring for people debugging the code, I think that this comment is evidence for that:

> The thought being that there was a reason that comments were being messaged before and the case for that reason might still exist in some contexts.

This suggests that on reading this code you got the impression that there may be at some point a need to convert the argument from a list to a string type. It turns out that that is not the case. Leaving the function in there would allow this misconception to persist and similarly waste time for anyone else trying to understand the code.

> We're also requiring tests

I'm dreading it! I'm sorry for the negative post but it seems like a disproportionate amount of effort to write effective tests for a display format change for something that is visibly incorrect all of the time in current builds. The bug is a discrepancy in the type expected and the type passed to `_uptodate` by `latest`, and to be able to test that it is passing the right value in I would need to write at least 5 different tests (one per call to `_uptodate`) for a 1795-line function that accesses git repositories on the disk, using an unfamiliar test framework.

I just want to be able to make casual contributions, I don't want to require the knowledge and dedication of a full-time employee just to fix a typo to get readable salt output in one state.

If I'm taking the other approach of just adding a no-op case to `_format_comments` (disagreements aside) then writing a test for that is simple, but all the test is saying is that the no-op does a no-op. This isn't really useful and it makes the same level of untested assumptions that I've made in this commit: that `_format_string` shouldn't be doing anything here and that that is the source of the bug.

waynew:
> I held off on addressing this because I disagree somewhat strongly with the amendment proposed, but here are my current thoughts on the matter.
> 
> > I wonder if it would be better to detect if we are dealing with a list type in `_format_comments`
> 
> This is fundamentally just a confusion of types, so I think `comments: str` vs `comments: List[str]` would clear up this confusion, though as far as I know that isn't done in salt.

I think you're probably right - I don't know if anyone is fundamentally opposed to type hints, at least I'm not. In the past we were supporting Python2, but now that we're dropping Python2 support with Sodium, it's much less of an issue.

> It does feel a little off to me to be making a list-to-string function handle the string case (outside of just throwing an exception), especially when this is currently calling `_format_comments` directly on the return value of `_format_comments` from higher up in the call stack, which is not a semantically sensible thing to be doing.

That's a reasonable counterpoint.

> > We're also requiring tests
> 
> I'm dreading it! I'm sorry for the negative post but it seems like a disproportionate amount of effort to write effective tests for a display format change for something that is visibly incorrect all of the time in current builds.

I get it - testing is hard, and testing things that *look* simple can sometimes be the hardest tests of all.

But we don't write tests simply to write tests - we write them to avoid this kind of problem. If this original code came with tests for the output functionality, then we wouldn't even be having this discussion :upside_down_face: 

> I just want to be able to make casual contributions, I don't want to require the knowledge and dedication of a full-time employee just to fix a typo to get readable salt output in one state.

That's definitely understandable, and totally acceptable! We readily welcome all kinds of contributions, though, we don't really yet have a good workflow for these casual contributions - they *are* valuable, but we also need tests to ensure that when we merge *new* changes to the code that we're not introducing regressions. We also _will_ be starting test clinics to help train people who are actually interested in writing tests.

> If I'm taking the other approach of just adding a no-op case to `_format_comments` (disagreements aside) then writing a test for that is simple, but all the test is saying is that the no-op does a no-op. This isn't really useful and it makes the same level of untested assumptions that I've made in this commit: that `_format_string` shouldn't be doing anything here and that that is the source of the bug.

Ah. Yeah, I was not sure what you meant earlier, but I've looked a few more times at this code and... wow :upside_down_face: 

---

On the topic of tests - have you ever written tests for Salt before? TBH, I think you would only have to write *one* test here to make a meaningful improvement:

```
def test_uptodate_should_properly_add_comments(self):
    ... # arrange here

    actual_ret = git._uptodate(ret, target, comments=expected_comments)

     assert expected_comments in actual_ret['comment']
```

Something to that effect. There's already a `tests/unit/states/test_git.py`, so you could totally just add the new testcase there.

Does that seem reasonable to you? That wouldn't require a lot of intensive testing or complicated scenarios... But it would 1) fix the problem correctly - I totally agree with your assessment that the `nop` approach is wrong and 2) ensure that any future changes to `_uptodate` are exposed before they break formatting.

And 3) it doesn't require you to have a lot of formal training :slightly_smiling_face: 

If you haven't written tests yet, and you don't mind a large download, I have a 3GB docker image that's built to make testing locally a pretty easy process.

If you do need help, feel free to ping me on IRC or the community Slack!

ScoreUnder:
Sorry for the frustration in the previous post, and thanks for your understanding.

I have written tests for saltstack before though they have mostly been a matter of reading surrounding tests and hoping that all the tools I need are "in reach" somewhere nearby in the code.

I will write tests for this PR soon, but in the meantime I'd like to ask some advice on what approach would be most appropriate for setting expectations for parameter types:

* `.pyi` files
* Python3-only function signatures
* MyPy comments (might be a problem with importing `typing` on py2?)
* `isinstance` assertion

waynew:
Currently, `isinstance` is the preferred approach.

I, for one, would support the introduction of type annotations, but I'm not sure what the broader community thinks -- it hasn't really come up in discussion yet. With Salt Sodium, we'll be dropping Python 2 support, so in theory it would be possible for us to start using type annotations.

ScoreUnder:
I'm tempted to pull out a lot of the behaviours of `latest` into their own private functions, just for the sake of easier mocking, but I think the parameter list of each will get pretty unwieldy and it would require a new exception class (for when an operation fails the entire state). I'll keep away from that for now, since I've got what I wanted to do done.

cmcmarrow:
Hi, @ScoreUnder. Thanks for the git PR! It looks like you have some failing tests. This is including pre-commit. You can set pre-commit by install it `pip install pre-commit`. You don’t need to worry about codecov. Also thank you for the tests.

cmcmarrow:
re-run full all

cmcmarrow:
You also have merge conflicts.

dwoz:
Closing due to lack of activity. Anyone should feel free to re-open or create a new PR if they want to see this through to the end.

Issue #52617, opened 2019-04-18
[2019.2.0] schedule.run_job throwing exception


tyhunt99:
### Description of Issue/Question

I am trying to set up a new scheduled job and test it, but when I try to run this job I get an exception.

My job:
```
salt minion schedule.add docker_system_prune function='docker.prune' cron='0 10 * * *'
```
This returns an exception:
```
salt schedule.run_job docker_system_prune
```
Exception:
```
2019-04-18 21:44:15,375 [tornado.application:638 ][ERROR   ][19423] Exception in callback <functools.partial object at 0x7ff1526db730>
Traceback (most recent call last):
  File "/usr/lib/python2.7/dist-packages/tornado/ioloop.py", line 605, in _run_callback
    ret = callback()
  File "/usr/lib/python2.7/dist-packages/tornado/stack_context.py", line 277, in null_wrapper
    return fn(*args, **kwargs)
  File "/usr/lib/python2.7/dist-packages/tornado/ioloop.py", line 626, in _discard_future_result
    future.result()
  File "/usr/lib/python2.7/dist-packages/tornado/concurrent.py", line 238, in result
    raise_exc_info(self._exc_info)
  File "/usr/lib/python2.7/dist-packages/tornado/gen.py", line 1063, in run
    yielded = self.gen.throw(*exc_info)
  File "/usr/lib/python2.7/dist-packages/salt/minion.py", line 959, in handle_event
    yield [minion.handle_event(package) for minion in self.minions]
  File "/usr/lib/python2.7/dist-packages/tornado/gen.py", line 1055, in run
    value = future.result()
  File "/usr/lib/python2.7/dist-packages/tornado/concurrent.py", line 238, in result
    raise_exc_info(self._exc_info)
  File "/usr/lib/python2.7/dist-packages/tornado/gen.py", line 828, in callback
    result_list.append(f.result())
  File "/usr/lib/python2.7/dist-packages/tornado/concurrent.py", line 238, in result
    raise_exc_info(self._exc_info)
  File "/usr/lib/python2.7/dist-packages/tornado/gen.py", line 307, in wrapper
    yielded = next(result)
  File "/usr/lib/python2.7/dist-packages/salt/minion.py", line 2371, in handle_event
    self.manage_schedule(tag, data)
  File "/usr/lib/python2.7/dist-packages/salt/minion.py", line 2233, in manage_schedule
    self.schedule.run_job(name)
[...]

PR #52666, opened 2019-04-23
[2019.2] Fixes to run_job in scheduler

garethgreenaway:
### What does this PR do?
Updating the max_running function to assume run=True if the value isnot available, this fixes the run_job function.  Adding a test for run_job.

#52617

### Tests written?
Yes

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/utils/schedule.py
+++ b/salt/utils/schedule.py
@@ -210,7 +210,7 @@ class Schedule(object):
         # dict we treat it like it was there and is True
 
         # Check if we're able to run
-        if not data['run']:
+        if not data.get('run', True):
             return data
         if 'jid_include' not in data or data['jid_include']:
             jobcount = 0
@@ -461,16 +461,18 @@ class Schedule(object):
             data['name'] = name
         log.info('Running Job: %s', name)
 
+        now = datetime.datetime.now()
         if not self.standalone:
             data = self._check_max_running(func,
                                            data,
                                            self.opts,
-                                           datetime.datetime.now())
+                                           now)
 
         # Grab run, assume True
         run = data.get('run', True)
         if run:
             self._run_job(func, data)
+            data['_last_run'] = now
 
     def enable_schedule(self):
         '''
--- a/salt/utils/schedule.py
+++ b/salt/utils/schedule.py
@@ -211,7 +211,8 @@ class Schedule(object):
 
         # Check if we're able to run, default to False
         # in case the `run` value is not present.
-        if not data.get('run'):
+        log.info('==== run %s ===', data['run'])
+        if not data['run']:
             return data
         if 'jid_include' not in data or data['jid_include']:
             jobcount = 0
--- a/salt/utils/schedule.py
+++ b/salt/utils/schedule.py
@@ -211,7 +211,6 @@ class Schedule(object):
 
         # Check if we're able to run, default to False
         # in case the `run` value is not present.
-        log.info('==== run %s ===', data['run'])
         if not data['run']:
             return data
         if 'jid_include' not in data or data['jid_include']:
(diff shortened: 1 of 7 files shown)


salt/utils/schedule.py
```diff
@@ -210,7 +210,7 @@ def _check_max_running(self, func, data, opts, now):
         # dict we treat it like it was there and is True
 
         # Check if we're able to run
-        if not data['run']:
+        if not data.get('run', True):
```
waynew: Why is it that we want to default to `True` here?

garethgreenaway: Updated the code, reverting this one back and setting the value in the run function.

waynew:
Would it also make sense to change that check to `if not data.get('run')`? I haven't really spent enough time grokking this code to know if that's the "right" idea or not, though. 🤷‍♂️

garethgreenaway:
@waynew Probably a good idea, in all cases it should be there but that would be a good "better safe than sorry" check.  Will update the PR.

waynew (approved):

waynew:
Looks like there was a failure but like... :suspect: 2 seconds off? I'm not really sure how to handle that. Rerun the test? Update that test suite to better handle time?

garethgreenaway:
@waynew Looks like 2 minutes off.  I've seen this particular error before. Taking a look.

tests/integration/scheduler/test_eval.py
```diff
@@ -753,15 +769,19 @@ def test_eval_minutes(self):
         ret = self.schedule.job_status(job_name)
         self.assertEqual(ret['_last_run'], run_time)
 
-        time.sleep(2)
+        # Give the job a chance to finish
+        while self.schedule.job_running(ret):
```
dwoz: Should this have some kind of timeout? What are the chances that something could goo wrong and it will run forever?

tests/integration/scheduler/test_postpone.py
```diff
@@ -43,10 +43,13 @@ def setUp(self):
             functions = {'test.ping': ping}
             self.schedule = salt.utils.schedule.Schedule(copy.deepcopy(DEFAULT_CONFIG), functions, returners={})
         self.schedule.opts['loop_interval'] = 1
+        self.schedule.opts['run_schedule_jobs_in_background'] = False
 
     def tearDown(self):
         self.schedule.reset()
 
+        del self.schedule
```
dwoz: Why is this needed? If we need some kind of shutdown, can we call the shutdown method directly, or could we just set self.schedule to None and let it get garbage collected?

garethgreenaway: The `reset` function that is being called does set the schedule to an empty dict.  I added the del because the test suit kept recommending it in a log.  I'll take it out.

tests/integration/scheduler/test_eval.py
garethgreenaway: Since the jobs are no longer running in the background this might not be necessary.  Will check and update.

tests/integration/scheduler/test_postpone.py
dwoz: Oh, right. I forgot the test suite will do that. I'm fine with leaving it in that case

garethgreenaway: I took it out and having it as an empty dict is fine, it will be recreated in between each test run.

tests/integration/scheduler/test_eval.py
garethgreenaway: Looks like the previous failures on py2-centos-7 are passing with the change to not have the jobs run in the background and don't require the check to see if the job is still running.  This makes sense since it's running it right away and returning right away.

dwoz (approved):

PR #52693, opened 2019-04-25
doc: Set script type explicitly to text/javascript

bdrung:
dh_sphinxdoc (used for building the documentation in the Debian package) searches for script lines and expect the type set to text/javascript. Otherwise dh_sphinxdoc won't find the scripts.

Therefore mark the JavaScript scripts explicitly as text/javascript.


s0undt3ch (approved):

bdrung:
Obsoleted by pull request #55956.

PR #52698, opened 2019-04-25
conftest: replace deprecated get_marker call in pytest

aplanas:
### What does this PR do?

The current conftest.py is using deprecated call to `get_marker()` in the item elements. We need to replace them with the new `get_closest_marker()` call, as documented in:

https://docs.pytest.org/en/latest/mark.html#updating-code

I didn't use `iter_markers()` iterator as looks like that there is not overwrite.


dwoz (approved):

s0undt3ch (commented):
:+1:

PR #52705, opened 2019-04-25
Pin kubernetes to <4.0

dwoz:
### What does this PR do?

Update kubernetes pinning in static requirements files. 

### Tests written?

No - Fixing existing tests

### Commits signed with GPG?

Yes


s0undt3ch:
To generate the static requirements:

```
pip install pre-commit
pre-commit -av
```

dwoz:
@s0undt3ch Fixed, thanks!

s0undt3ch (approved):

Ch3LL:
Do we have an issue open so we can track down why salt does not currently work with 4?

s0undt3ch:
Don't think we have.

dwoz:
re-run centos

dwoz:
re-run full all

angrydead (commented):
Can this be simplified to the following?

```
if (name in proc.name() and _search(proc)) or search is None:
    return proc
```

dwoz:
@pullit

It seems like we'd be missing the `if name in proc.name():` after the first if. I think in order for it to be the exact same logic we would need to be something like:

```
if (name in proc.name() and _search(proc)) or (search is None and name in proc.name()):
    return proc
```

However, I also think it's debatable wether or not that's actually simpler. Also, having said that, I'm just trying to prevent some tests from raising and exception in this PR and I feel like changing too much will run the risk of introducing a bug. I'd prefer not to change it.

dwoz:
re-run full all

pullit:
@dwoz That's fair.  I'm new here and don't want to step on any toes.  In hindsight, your original implementation is more readable and scales better if additional functionality is needed.

dwoz:
@pullit Thanks, In any case your review is welcomed. An extra set of eyes no changes never hurts. :)

dwoz:
re-run full all

s0undt3ch (approved):

dwoz:
re-run full all

dwoz:
re-run full all

.ci/kitchen-centos6-py3
```diff
@@ -0,0 +1,119 @@
+properties([
```
s0undt3ch: CentOS 6 under Py3?
We're not even building that on branch builds...

s0undt3ch (changes requested):

s0undt3ch (approved):

s0undt3ch (approved):

PR #47679, opened 2018-05-16
Adding a merge_all function to slsutil

onlyanegg:
The merge_all function merges a list of objects in order. This will make merging multiple objects more readable. Consider the difference between merging three objects with `slsutil.merge` versus merging with `slsutil.merge_all`

With `slsutil.merge`:
```
{{ hadoop.hdfs.data_node.service.name }}_environment_file_installed:
  file.managed:
    - name: /etc/sysconfig/{{ hadoop.hdfs.data_node.service.name }}
    - source: salt://hadoop/files/environment
    - template: jinja
    - context:
        environment: {{
          salt.slsutil.merge(
            hadoop.environment, salt.slsutil.merge(
              hadoop.hdfs.environment,
              hadoop.hdfs.data_node.environment
            )
          )
        }}
```

With `slsutil.merge_all`:
```
{{ hadoop.hdfs.data_node.service.name }}_environment_file_installed:
  file.managed:
    - name: /etc/sysconfig/{{ hadoop.hdfs.data_node.service.name }}
    - source: salt://hadoop/files/environment
    - template: jinja
    - context:
        environment: {{
          salt.slsutil.merge_all([
            hadoop.environment,
            hadoop.hdfs.environment,
            hadoop.hdfs.data_node.environment
          ])
        }}
```

I'm interested to know if you all think it would be better to add this function or to overload the merge function. 

### What does this PR do?

This PR adds a merge_all function to the slsutil module

### What issues does this PR fix or reference?

None

### Tests written?

No

### Commits signed with GPG?

Yes

Please review [Salt's Contributing Guide](https://docs.saltstack.com/en/latest/topics/development/contributing.html) for best practices.

See GitHub's [page on GPG signing](https://help.github.com/articles/signing-commits-using-gpg/) for more information about signing commits with GPG.

PR #52708, opened 2019-04-26
Add the "merge" jinja filter

bbinet:
This filter can be really useful in states to merge two dicts recursively.

The code is trivial as it is just a matter of exposing the existing merge function as a jinja filter through the @jinja_filter('merge') decorator.
That is the reason why it seems safe to merge it in the 2019.2 branch instead of the develop branch.

### DIFF
--- a/doc/topics/jinja/index.rst
+++ b/doc/topics/jinja/index.rst
@@ -1212,6 +1212,28 @@ Returns:
 
   'default'
 
+
+.. jinja_ref:: merge
+
+``merge``
+------------
+
+.. versionadded:: 2019.2.1
+
+Deeply merge a dict with another dict.
+
+Example:
+
+.. code-block:: jinja
+
+  {{ {'a1': {'b1': 'foo'}, 'a2': 'bar'} | merge({'a1': {'b1': 'foo2'}, 'a3': 'bar3'}) }}
+
+Returns:
+
+.. code-block:: python
+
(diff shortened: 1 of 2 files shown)


bbinet:
Also, is there any chance this patch can be published in the 2019.2.1 release? (it would be helpful to me)

dwoz (changes requested):
@bbinet Actually, this change should go to the develop branch as `2019.2.x` is bugfix only. Also, there should be some tests associated with the change.

max-arnold:
@bbinet Are you aware that [`merge_all`](https://github.com/saltstack/salt/pull/47679) function exists Salt since 2019.2.0?

doc/topics/jinja/index.rst
```diff
@@ -1212,6 +1212,28 @@ Returns:
...
+
+.. code-block:: jinja
+
+  {{ {'a1': {'b1': 'foo'}, 'a2': 'bar'} | merge({'a1': {'b1': 'foo2'}, 'a3': 'bar3'}) }}
+
+Returns:
+
+.. code-block:: python
+
+  {'a1': {'b1': 'foo2'}, 'a2': 'bar', 'a3': 'bar3'}
```
pullit: The merged value of `b1` is more like a rewrite than a merge.  I would expect a JSON array.

bbinet:
Thanks @max-arnold, I was not aware of the slsutil module which I can actually use to merge my dicts.
So this PR can be closed.

PR #52709, opened 2019-04-26
LocalClient: don't parse input twice and test fix.

DmitryKuzmenko:
### What does this PR do?
Related to #49886 where passing input args and kwargs through `parse_input` was inroduced.

I've removed passing the args to `parse_input` from all functions except `run_job` because other functions aren't used that args and all are calling `run_job`. This avoids parsing that args multiple times.

Also I've updated the CMDModuleTest according to `parse_input` expectations. It yamlifies each text value in arguments so it expects arguments values to be a YAML code. This means that multiline blocks of python code shall be formatted accordingly.
Reference: https://yaml.org/spec/1.2/spec.html#id2795688
Examples and human friendly description: https://yaml-multiline.info/#block-scalars

Fixes the following tests in develop branch:
integration.modules.test_cmdmod.CMDModuleTest.test_exec_code
integration.modules.test_cmdmod.CMDModuleTest.test_exec_code_with_multiple_args
integration.modules.test_cmdmod.CMDModuleTest.test_exec_code_with_single_arg

Fixes merge forward #52571

### DIFF
--- a/tests/integration/modules/test_cmdmod.py
+++ b/tests/integration/modules/test_cmdmod.py
@@ -6,7 +6,6 @@ import os
 import random
 import sys
 import tempfile
-import textwrap
 
 # Import Salt Testing libs
 from tests.support.case import ModuleCase
@@ -262,9 +261,11 @@ class CMDModuleTest(ModuleCase):
         '''
         cmd.exec_code
         '''
-        code = textwrap.dedent('''\
-               import sys
-               sys.stdout.write('cheese')''')
+        # `code` is a multiline YAML text. Formatting it as a YAML block scalar.
+        code = '''|
+                   import sys
+                   sys.stdout.write('cheese')
+               '''
         self.assertEqual(self.run_function('cmd.exec_code',
                                            [AVAILABLE_PYTHON_EXECUTABLE,
                                             code]).rstrip(),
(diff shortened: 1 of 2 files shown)


tests/integration/modules/test_cmdmod.py
```diff
@@ -262,7 +262,8 @@ def test_exec_code(self):
         '''
         cmd.exec_code
         '''
-        code = textwrap.dedent('''\
+        # `code` is a multiline YAML text. Formatting it as a YAML block scalar.
+        code = textwrap.dedent('''|
```
waynew: Does `dedent` do anything here anymore? I'm pretty sure that with this leading `|` there's no leading spaces to dedent, and the output of this is

```
|
            import sys
            sys.stdout.write('cheese')
```

I'd either remove the misleading `dedent` call or rewrite it as

```
code = textwrap.dedent('''
    |
        import sys
        sys.stdout.write('cheese')
'''').lstrip()
```

(or, something else)

mattp-:
:+1: thank you for catching this, forgot about it on my original PR

tests/integration/modules/test_cmdmod.py
dwoz: @waynew @DmitryKuzmenko We should be using `tests.support.helpers.dedent` over `textwrap.dedent`.

DmitryKuzmenko: Wow. Thank you both!

s0undt3ch (approved):

waynew (approved):

Issue #39875, opened 2017-03-07
Can't use grains.list_present with complex(dict) types


nw0428:
### Description of Issue/Question
A state like this:
```
service_ssh:
  grains.list_present:
    - name: services
    - value:
        - sshd:
            - hostname: localhost
            - port: 22
```
Will work the first time but not the second. The second time it throws the following:
```
        ID: service_ssh
    Function: grains.list_present
        Name: services
      Result: False
     Comment: An exception occurred in this state: Traceback (most recent call last):
                File "/usr/lib/python2.7/site-packages/salt/state.py", line 1734, in call
                  **cdata['kwargs'])
                File "/usr/lib/python2.7/site-packages/salt/loader.py", line 1651, in wrapper
                  return f(*args, **kwargs)
                File "/usr/lib/python2.7/site-packages/salt/states/grains.py", line 150, in list_present
                  if set(value).issubset(set(__salt__['grains.get'](name))):
              TypeError: unhashable type: 'dict'
```

### Versions Report
```
Salt Version:
           Salt: 2016.3.5
 
Dependency Versions:
           cffi: 1.6.0
       cherrypy: 3.2.2
       dateutil: 2.5.3
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.8
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.4.8
   mysql-python: Not Installed
      pycparser: 2.14
       pycrypto: 2.6.1
         pygit2: Not Installed
         Python: 2.7.5 (default, Nov  6 2016, 00:28:07)
   python-gnupg: Not Installed
         PyYAML: 3.11
          PyZMQ: 15.3.0
           RAET: Not Installed
          smmap: Not Installed
        timelib: Not Installed
        Tornado: 4.2.1
            ZMQ: 4.1.4
 
System Versions:
           dist: centos 7.3.1611 Core
        machine: x86_64
        release: 3.10.0-514.6.2.el7.x86_64
         system: Linux
        version: CentOS Linux 7.3.1611 Core
```

Ch3LL:
Looks like i'm able to replicate this.

Here is a docker container for anyone that wants to quickly replicate this:

[...]

PR #52710, opened 2019-04-26
Fix complex grain

mickenordin:
### What does this PR do?
If a grain is a list of complex objects, grains.py is unable to compare the currently set grain to the potentially new one. Instead it fails with TypeError: unhashable type: 'dict' (might be something else aswell depending on what you are trying to set as a grain). This PR will make it so that a list grain that contains complex objects can be compared via the function make_hashable.    
issue #39875

### Tests written?

Yes, in tests/unit/states/test_grains.py there is a test called "test_make_hashable"

### Commits signed with GPG?

No


waynew:
Hi! Thanks for the PR!

I think you either meant to make this PR against a different branch or you want to rebase your changes against develop. If you need any pointers I'm happy to help!

mickenordin:
Hi, you are probably right. I started to make a clean PR from my fork to
development, but then I read your development guidelines and tried to
follow the instructions there...

This is a simple fix in a file that has been untuched for a long time. What
is the best way to fix this PR?

Den fre 26 apr. 2019 16:22Wayne Werner <notifications@github.com> skrev:

waynew:
@mickenordin If you've just tried to merge into the wrong branch you can go to the top right and click "edit" and then change what branch you're trying to merge into (probably 2018.3 is my guess).

If that's not right, let me know where it belongs, and where you branched from, and I can help you rebase or cherry pick your changes.

mickenordin:
@waynew thank you very much, now it looks cleaner I think :)

salt/states/grains.py
```diff
@@ -44,6 +44,13 @@ def exists(name, delimiter=DEFAULT_TARGET_DELIM):
         ret['comment'] = 'Grain does not exist'
     return ret
 
+def flatten(li, flattened = list()):
+    for subli in li:
+        if type(subli) == list:
+            flatten(subli, flattened)
+        else:
+            flattened.append(frozenset(subli))
+    return set(flattened)
```
waynew: Would it be better to just use `set` here? The default arg *definitely* should be changed, though.

I'm thinking something like:

```
def flatten(li, flattened=None):
    flattened = flattened or set()
    for subli in li:
        if type(subli) == list:
            flattened.update(flatten(subli))
        else:
            flattened.add(frozenset(subli))
    return flattened
```

I'm not entirely certain that's the best approach either, though.

waynew (changes requested):
Thanks for the PR! Could you update the function definition, as well as add some tests?

`tests/unit/states/test_grains.py` would be a good place to add (or update) tests for this.

salt/states/grains.py
mickenordin: Great, I will fix those things :)

mickenordin:
@waynew thank you for you patiens and your help. I have now fixed all lint errors, added a test that uses the function and added comments and over all tried to make the code more clear and precise.

waynew (approved):

mickenordin:
What is the next step for this PR? Is there a process for merging or is more review needed?

waynew:
@mickenordin shouldn't be - just keep getting changes before we can merge it in 😑 I'll keep my eye out today

waynew:
🎉

PR #52712, opened 2019-04-26
mount: fix remount when fstype is not set

aplanas:
### What does this PR do?

Fix the remount command when fstype is not set.

Also fix a wrong rebase that happens somewhere when the big forward port dragged changed from the past.


dwoz (changes requested):
@aplanas please make sure this change has test coverage.

dwoz (approved):

dwoz (approved):

PR #52731, opened 2019-04-28
Add replace_on_change option to ddns.update()

awlx:
### What does this PR do? 
Only replace record in DNS if entry content has changed thus avoiding
unnecessary DDNS updates as well as state changes in ddns.present.

Signed-off-by: Annika Wickert <aw@awlnx.space>
Signed-off-by: Maximilian Wilhelm <max@sdn.clinic>

### Previous Behavior
ddns.update with replace option set did always trigger a DDNS update thus a state change.

### New Behavior
With replace_on_change option set a DDNS update is only triggered if records really changed. But we keep the replace option to not break old states or functions.

### Tests written?

No

### Commits signed with GPG?

Yes


dwoz (changes requested):
@awlx Please add some test coverage for this change.

sagetherage:
@awlx 
please rebase this against the master branch

sagetherage:

PR #52735, opened 2019-04-28
Fix salt-cloud traceback in python 3

NiCalderon:
### What does this PR do?
Fix salt/cloud/clouds/ec2.py get_imageid to use sorted's key instead or cmp parameter (removed in python 3)
No issue was filed
### Previous Behavior
Traceback when attempting to start an instance with an image not starting with "ami-"

### New Behavior
No more traceback and expected image is being used.

### Tests written?

No

### Commits signed with GPG?

No

### DIFF
--- a/salt/cloud/clouds/ec2.py
+++ b/salt/cloud/clouds/ec2.py
@@ -1223,10 +1223,9 @@ def get_imageid(vm_):
               'Filter.0.Name': 'name',
               'Filter.0.Value.0': image}
     # Query AWS, sort by 'creationDate' and get the last imageId
-    _t = lambda x: datetime.datetime.strptime(x['creationDate'], '%Y-%m-%dT%H:%M:%S.%fZ')
     image_id = sorted(aws.query(params, location=get_location(),
                                  provider=get_provider(), opts=__opts__, sigver='4'),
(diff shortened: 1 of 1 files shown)


dwoz:
@NiCalderon Was this submitted against the proper base branch? It definitely looks like something is off based on the number of files changed and the commit history.

dwoz (changes requested):
@NiCalderon Was this submitted against the proper base branch? It definitely looks like something is off based on the number of files changed and the commit history.

NiCalderon:
It suggested to merge to develop by default and I forgot to change that to 2019.2, where the modified code was first introduced.

NiCalderon:
I edited it, but should I resubmit?

dwoz:
@NiCalderon might be a good idea, we'll need some test coverage for this change too.

NiCalderon:
I will resubmit. As far as testing, I looked to see if there was anything I could adapt, but the tests for this modules are very thin. Any suggestions as to how I should proceed?

dwoz:
@NiCalderon Just try to write a test that exposes the bug you are fixing. If the test fails when your patch is not present but passes when your patch is applied, that should be good.

NiCalderon:
Created a new PR (https://github.com/saltstack/salt/pull/52936), closing this one

PR #52754, opened 2019-04-30
[2018.3] Fixes to multiprocessing queue when using MacOS

garethgreenaway:
### What does this PR do?
The maximum for the multiprocessing queue on MacOS is 32767, so if we running on MacOS then we use that maximum.

N/A

### Previous Behavior
Tests do not run currently on MacOS

### New Behavior
Allows tests to run

### Tests written?
No.  Fixes to existing tests.

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/log/setup.py
+++ b/salt/log/setup.py
@@ -805,13 +805,17 @@ def setup_extended_logging(opts):
 
 def get_multiprocessing_logging_queue():
     global __MP_LOGGING_QUEUE
+    from salt.utils.platform import is_darwin
 
     if __MP_IN_MAINPROCESS is False:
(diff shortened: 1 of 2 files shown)


s0undt3ch (approved):
Darn, I though that was the max value when leaving the queue size empty...

waynew:
@garethgreenaway not 100% sure if we should backport this to 2017.7 for that final release, but currently it looks like this is what's preventing that branch from building. 2018.3 is running into some other issues that I'm looking into.

s0undt3ch:
We should also bring this to 2019.2.1?

PR #52697, opened 2019-04-25
Fix influx retention_policy overwrite of "default" when not specified and add support for shard_duration

bbinet:
### What does this PR do?

This PR improves influx module and influxdb_retention_policy state:
- c45748b79dc237aa844b305608613985d985fe34: fix unexpected overwrite of "default" param in influxdb.alter_retention_policy
- ded8f4ce1318618bb8723d374c3478a49c03e3ea: add support for shard_duration in influx module
- 3a2eb541046b7e690bebcb69cd190427d5265cb1: add support for shard_duration in influxdb_retention_policy state

### What issues does this PR fix or reference?

This PR fix the unexpected overwrite of the retention policy "default" value when "default" param was omitted in influxdb.alter_retention_policy call.

### Previous Behavior

After a call to influxdb.alter_retention_policy with "default" omitted, the function will overwrite the "default" value to set it to "False".

### New Behavior

After a call to influxdb.alter_retention_policy with "default" omitted, the function will keep the "default" value unchanged.

### Tests written?

No

### Commits signed with GPG?

No

Please review [Salt's Contributing Guide](https://docs.saltstack.com/en/latest/topics/development/contributing.html) for best practices.

See GitHub's [page on GPG signing](https://help.github.com/articles/signing-commits-using-gpg/) for more information about signing commits with GPG.

PR #52757, opened 2019-04-30
Add support for setting influxdb retention policy shard_duration

bbinet:
(See #52697 original PR for 2019.2)

This PR improves influx module and influxdb_retention_policy state:
- 0b8264d7877dc2cded7eeba63b567825baae29c0: fix unexpected overwrite of "default" param in influxdb.alter_retention_policy
- 8fac8a1c31833348c510b3cfc33ea9668952457c: add support for shard_duration in influx module
- bcef2397d47c2a3c9f879651b643557c411b344f: add support for shard_duration in influxdb_retention_policy state
- 6dbab0fcfac6441dcb1bbe27bdb4fcc653fbbcb0: fix some buggy edge cases in the convert_duration function and also support INF duration
- 649fadb48485808fb8f8399cc127b3a958799321: fix some false positive rp update when shard_duration='0s'

### DIFF
--- a/salt/modules/influxdbmod.py
+++ b/salt/modules/influxdbmod.py
@@ -449,9 +449,9 @@ def create_retention_policy(database,
 
 def alter_retention_policy(database,
                            name,
-                           duration,
-                           replication,
-                           default=False,
+                           duration=None,
+                           replication=None,
+                           default=None,
                            **client_args):
     '''
     Modify an existing retention policy.
@@ -462,7 +462,7 @@ def alter_retention_policy(database,
     database
         Name of the database for which the retention policy was defined.
 
-    duration
+    duration : None
         New duration of given retention policy.
 
         Durations such as 1h, 90m, 12h, 7d, and 4w, are all supported
@@ -471,13 +471,13 @@ def alter_retention_policy(database,
         never be deleted – use 'INF' for duration.
         The minimum retention period is 1 hour.
 
-    replication
+    replication : None
         New replication of given retention policy.
 
         This determines how many independent copies of each data point are
         stored in a cluster.
 
-    default : False
+    default : None
         Whether or not to set the modified policy as default.
 
     CLI Example:
--- a/salt/modules/influxdbmod.py
+++ b/salt/modules/influxdbmod.py
@@ -407,6 +407,7 @@ def create_retention_policy(database,
                             duration,
                             replication,
                             default=False,
+                            shard_duration="0s",
                             **client_args):
     '''
     Create a retention policy.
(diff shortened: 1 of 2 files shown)


salt/modules/influxdbmod.py
```diff
@@ -442,16 +455,17 @@ def create_retention_policy(database,
...
 
 
 def alter_retention_policy(database,
                            name,
-                           duration,
-                           replication,
-                           default=False,
+                           duration=None,
+                           replication=None,
+                           default=None,
```
Ch3LL: Can you elaborate as to why having `default=False` is causing issues? I did notice on the api documentation that the default value should be None but just trying to understand the issue this is fixing

salt/modules/influxdbmod.py
```diff
@@ -434,6 +435,18 @@ def create_retention_policy(database,
...
+
+        Durations such as 1h, 90m, 12h, 7d, and 4w, are all supported
+        and mean 1 hour, 90 minutes, 12 hours, 7 day, and 4 weeks,
+        respectively. Infinite retention is not supported.
+        As a workaround, specify a “1000w” duration to achieve an extremely
+        long shard group duration.
+        Defaults to “0s”, which is interpreted by the database to mean the
+        default value given the duration.
+        The minimum shard group duration is 1 hour.
+
```
Ch3LL: will need to add `.. versionadded:: Neon` for each new argument added in this pr.

Ch3LL (changes requested):
Thanks for splitting this out. I appreciate it . Unfortunately we will not be able to merge this until tests are added. Here are the docs about the test runner: https://docs.saltstack.com/en/latest/topics/tutorials/writing_tests.html

Let me know if i can help answer any other questions to help get you started. thanks

salt/modules/influxdbmod.py
bbinet: @Ch3LL You're right, after a few more tests, default=False is the same as default=None. I thought it had fixed my issue probably because of a side effect: the "default" value was already changed in my influxdb database.

bbinet:
> Thanks for splitting this out. I appreciate it . Unfortunately we will not be able to merge this until tests are added. Here are the docs about the test runner: https://docs.saltstack.com/en/latest/topics/tutorials/writing_tests.html

@Ch3LL I understand, but sadly I won't have time right now to create new tests from scratch for influxdb module and states, I'm sorry (I could have done it if some tests would already exist for that module and states).
So let's close this PR. 
If someone wants to take on this task, please do ;-)

Issue #52759, opened 2019-04-30
salt.modules.runit fails 'invalid argument' on non-symlinked service


bbczeuz:
### Description of Issue/Question
Service management via salt-formula in a runit based container fails.

### Setup
state top.sls:
```
base:
  '*':
    - salt.minion
```
(use salt-formula; The whole salt-setup runs within a runit-based container)

pillar salt.sls:
```
salt:
  master_remove_config: True
  minion_remove_config: True
  release: '2018.3'
  version: latest
  minion:
    master: 192.168.0.1
    master_port: 4506
    publish_port: 4505
```

### Steps to Reproduce Issue
```
root@testsystem:~# salt 'saltmaster-test01' state.apply salt.minion
saltmaster-test01:
[...]
          ID: salt-minion
    Function: service.running
      Result: False
     Comment: An exception occurred in this state: Traceback (most recent call last):
                File "/usr/lib/python2.7/dist-packages/salt/state.py", line 1919, in call
                  **cdata['kwargs'])
                File "/usr/lib/python2.7/dist-packages/salt/loader.py", line 1918, in wrapper
                  return f(*args, **kwargs)
                File "/usr/lib/python2.7/dist-packages/salt/states/service.py", line 407, in running
                  if not _available(name, ret):
                File "/usr/lib/python2.7/dist-packages/salt/states/service.py", line 326, in _available
                  avail = __salt__['service.available'](name)
                File "/usr/lib/python2.7/dist-packages/salt/modules/runit.py", line 427, in available
                  return name in _get_svc_list(name)
                File "/usr/lib/python2.7/dist-packages/salt/modules/runit.py", line 390, in _get_svc_list
                  return sorted([os.path.basename(el) for el in _get_svc_path(name, status)])
                File "/usr/lib/python2.7/dist-packages/salt/modules/runit.py", line 354, in _get_svc_path
                  ena.add(os.readlink(el))
              OSError: [Errno 22] Invalid argument: '/etc/service/salt-minion'
     Started: 12:41:32.640261
    Duration: 2.0 ms
     Changes:   
[...]
```

### Versions Report
(same on master and minion)
```
# salt --versions-report
Salt Version:
           Salt: 2018.3.4
 
Dependency Versions:
           cffi: Not Installed
       cherrypy: unknown
       dateutil: 2.6.1
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.10
[...]

PR #52760, opened 2019-04-30
Runit: Only resolve link if path is a link

bbczeuz:
### What does this PR do?
Module runit: Check wether a service's path is really a link. If it is a link, resolve it; else directly add path.

https://github.com/saltstack/salt/issues/52759

### Previous Behavior
service.running throws an exception at `os.readlink(el) (OSError: [Errno 22] Invalid argument: '/etc/service/salt-minion'` (see Issue; /etc/service/salt-minion exists as a normal directory, not as symlink)

### New Behavior
service.running runs without throwing an exception

### Tests written?
No

### Commits signed with GPG?
No

### DIFF
--- a/salt/modules/runit.py
+++ b/salt/modules/runit.py
@@ -350,7 +350,7 @@ def _get_svc_path(name='*', status=None):
     ena = set()
     for el in glob.glob(os.path.join(SERVICE_DIR, name)):
         if _is_svc(el):
-            ena.add(os.readlink(el))
+            ena.add(os.readlink(el) if os.path.islink(el) else el)
             log.trace('found enabled service path: %s', el)
 
     if status == 'ENABLED':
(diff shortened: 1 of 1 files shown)


dwoz (changes requested):
@bbczeuz Thanks for the contribution. Can you please add a unit (or integration) test to cover this change?

waynew:
Maybe we should be using [realpath](https://docs.python.org/3/library/os.path.html#os.path.realpath) instead?

bbczeuz:
> 
> 
> @bbczeuz Thanks for the contribution. Can you please add a unit (or integration) test to cover this change?

I'm thinking about this. Probably the easiest would be to base this on the test_service.py code. Are there any runit based integration test setups available? We're using this code in a phusion/baseimage Docker container.

waynew:
@bbczeuz It doesn't look like it - `grep -Inr runit tests` doesn't show up anything of note. We have an `runit` module (`salt/modules/runit.py`), but it doesn't look like we're testing it.

sagetherage:
@bbczeuz 
please rebase this against the master branch

sagetherage:
@bbczeuz 

PR #52772, opened 2019-05-01
2018.3: bug: executors lazyloader is missing functions and proxy args

mattp-:
### What does this PR do?
these are needed for this functionality to work

### What issues does this PR fix or reference?
n/a

### Tests written?
should be covered by existing tests

### Commits signed with GPG?
No

Please review [Salt's Contributing Guide](https://docs.saltstack.com/en/latest/topics/development/contributing.html) for best practices.

See GitHub's [page on GPG signing](https://help.github.com/articles/signing-commits-using-gpg/) for more information about signing commits with GPG.

PR #52773, opened 2019-05-01
2019.2: bug: executors lazyloader is missing functions and proxy args

mattp-:
What does this PR do?
these are needed for this functionality to work

### Tests written?
should be covered by existing tests

### Commits signed with GPG?
no


waynew:
@mattp- thanks for the PR!

Is there a test that this fixes, or would it be possible to throw together a regression test?

mattp-:
the bug is only seen in the sudo executor in core, for which there are no tests. I am working on a replacement implementation of that executor that doesn't rely on sudo but it is not ready for upstream

mattp-:
@waynew i worked out adding an integration test that exercises the bug, however I can't backport it to #52772 because it relies on salt-call having --module-executors via salt-call, which is new functionality added elsewhere in 2019. while doing this i uncovered that executors weren't actually exposed via saltutil.sync* in runner & module, so added that in both pr's. looks like serializer was also added in 2019 but not 2018, so i added it there as well in the other PR.

mattp-:
@waynew clean now. py3 windows seems hung from unrelated issue, but py2 windows passed along with the others.

waynew:
Looks like last couple passed, just waiting on this current build to finish

waynew:
Oi. Looks like this is out of date again 😜

mattp-:
something seems wrong with that win2k16 build, its been running for 2 hours. I'll update branch maybe it will pass this time

mattp-:
@waynew clean now

waynew (approved):

max-arnold:
I'm pretty sure this could cause some conflicts during a forward-merge, because I added `sync_executors` to develop (Neon): #51427

PR #52844, opened 2019-05-06
zypperpkg: don't consider error a pattern not found

aplanas:
The `rpm -q --provides --whatprovides 'pattern()' command return
error code 1 when any pattern is present. This patch ignore the
return error code, as a not found pattern is not an error.

### DIFF
--- a/salt/modules/zypperpkg.py
+++ b/salt/modules/zypperpkg.py
@@ -2222,7 +2222,9 @@ def _get_installed_patterns(root=None):
     if root:
         cmd.extend(['--root', root])
     cmd.extend(['-q', '--provides', '--whatprovides', 'pattern()'])
-    output = __salt__['cmd.run'](cmd)
+    # If no `pattern()`s are found, RPM returns `1`, but for us is not
+    # a real error.
+    output = __salt__['cmd.run'](cmd, ignore_retcode=True)
 
     installed_patterns = [_pattern_name(line) for line in output.splitlines()
                           if line.startswith('pattern() = ')]


Ch3LL (changes requested):
can you add a regression test?

aplanas:
@Ch3LL Not sure that is needed, as this code is tested already here:

https://github.com/saltstack/salt/blob/develop/tests/unit/modules/test_zypperpkg.py#L1305

The change is a cosmetic one, as affect the logs but in any case affect the behavior.

Ch3LL:
if I'm reading that test correctly it seems to test when the `rpm` command returns when a pattern is found. Might be worthy to add a test when a pattern is not found?

aplanas:
@Ch3LL You are correct, the rpm command search for all the patterns installed, that can be 0 or more. Somehow rpm return an exit status different from 0 when no pattern is available and an ERROR is logged, but this is not a real error.

The patch do not change the behavior, as 0 or more patterns founds is a valid outcome. Only makes salt to not log ERROR when there are not patterns.

Ch3LL:
I understand that it does not change the behavior, I only was looking for a way to add more test coverage to this module since its not currently testing when the pattern is not found  and this fix addresses that situation. But you make a good point, I agree with you that this would technically would not be a regression test, as it would not fail with/without the new change, so I'll go ahead and merge for now.

aplanas:
@Ch3LL Thanks for the review. Detecting and pointing missing tests is really critical.

Issue #52781, opened 2019-05-01
unit.modules.test_win_file.WinFileTestCase.test_issue_52002_check_file_remove_symlink test failure


waynew:
# unit.modules.test_win_file.WinFileTestCase.test_issue_52002_check_file_remove_symlink test failure

2019.2 failed salt-windows-2016-py3, salt-windows-2019-py3
--------------------------------------------------------------------------------

Could not create 'C:\Users\ADMINI~1\AppData\Local\Temp\base-a_dga918\child 2\link' - [1314] A required privilege is not held by the client.

```
Traceback (most recent call last):
  File "C:\\Users\\ADMINI~1\\AppData\\Local\\Temp\\kitchen\\testing\salt\modules\win_file.py", line 1151, in symlink
    win32file.CreateSymbolicLink(link, src, int(is_dir))
pywintypes.error: (1314, 'CreateSymbolicLink', 'A required privilege is not held by the client.')

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "C:\\Users\\ADMINI~1\\AppData\\Local\\Temp\\kitchen\\testing\tests\support\helpers.py", line 129, in wrap
    return caller(cls)
  File "c:\users\admini~1\appdata\local\temp\kitchen\testing\tests\unit\modules\test_win_file.py", line 76, in test_issue_52002_check_file_remove_symlink
    self.assertTrue(win_file.symlink(target, symlink))
  File "C:\\Users\\ADMINI~1\\AppData\\Local\\Temp\\kitchen\\testing\salt\modules\win_file.py", line 1158, in symlink
    exc.strerror
salt.exceptions.CommandExecutionError: Could not create 'C:\Users\ADMINI~1\AppData\Local\Temp\base-a_dga918\child 2\link' - [1314] A required privilege is not held by the client.
```

waynew:
This is failing on develop with a different error

[...]

PR #52849, opened 2019-05-06
Fix symlink test on Windows

twangboy:
### What does this PR do?
Removes redundent test from WinFileCheckPermsTestCase object
Removes, conflicting mock that was causing the following test to fail:
```
unit.modules.test_win_file.WinFileTestCase.test_issue_52002_check_file_remove_symlink
```

https://github.com/saltstack/salt/issues/52781

### Tests written?
Yes

### Commits signed with GPG?
Yes


Ch3LL (changes requested):
just a lint failure to fix

Ch3LL (approved):

PR #52932, opened 2019-05-07
Fix GCE Windows instances being identified as physical by core grains

kvasserman:
### Previous Behavior
'virtual' grain on GCE instances is set to 'physical'

### New Behavior
'virtual' grain on GCE instances is set to 'gce'

### Tests written?

No

### Commits signed with GPG?

No


twangboy (changes requested):
2019.2.0 has been released. Please rebase this PR against the 2019.2.1 branch.

PR #52944, opened 2019-05-08
Autodetect source_hash file encoding

NiCalderon:
### What does this PR do?
Autodetect source_hash file encoding

None

### Previous Behavior
If source_hash's URL referred to a file  with a different encoding than the system's encoding:
  File "/home/nicolasc/salt/tests/unit/modules/test_file.py", line 1013, in test_extract_hash
    hsum = filemod.extract_hash(tfile.name, '', 'Salt-Minion-2019.2.0-Py3-AMD64-Setup.exe')
  File "/home/nicolasc/salt/salt/modules/file.py", line 4448, in extract_hash
    for line in fp_:
  File "/usr/lib/python3.7/codecs.py", line 322, in decode
    (result, consumed) = self._buffer_decode(data, self.errors, final)
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xff in position 0: invalid start byte

### New Behavior
Attempt to detect file encoding before opening it.

### Tests written?

Yes. If the feature is not enabled, tracebacks should be observed on both Windows (on the utf-8 file) and Linux (on the utf-16 file).

### Commits signed with GPG?

Yes


NiCalderon:
It seems python 2's os.open has no support for the encoding argument. I'm therefore left with 3 options:
1. Disable feature in python <3
2. If we do not support python <2.6, make extract_hash use io.open instead of salt.utils.files.fopen, like blockreplace does. Seeing as this is the only use of io.open in the whole code base, I'm thinking this could lead to race conditions, as documented in salt.utils.files.fopen.
3. If we do not support python <2.6, make salt.utils.files.fopen use io.open, but this is a fairly core change, with potentially major impacts (performance, to name only one).
4. If we want to support earlier python versions, there is codecs.open, but it appears to be an even more disruptive change.

What would you recommend?

garethgreenaway:
@NiCalderon Thanks for the PR!  We do not support python older than 2.7 currently and are planning to drop support for Python 2 in future releases.

NiCalderon:
Force pushed to retrigger checks

sagetherage:
@NiCalderon please rebase this against the master branch

sagetherage:
@NiCalderon 

PR #52945, opened 2019-05-08
PR38362 missed that redhat 6 also can use nm_controlled.

whytewolf:
added nm_controlled to the rh6_eth.jinja template.
#38362  did not account for redhat 6

### What does this PR do?
adds NM_CONTROLLED to the redhat 6 rh_ip eth template

### Previous Behavior
nm_controlled on redhat 6 was silently ignored in the configuration of an interface. 

### New Behavior
nm_controlled will now be enabled if it is added to the config. 

### Tests written?

currently no. Not sure how to write tests for the internal templates. And can't find examples in the test suite, Would love help with this. 

### Commits signed with GPG?
no


whytewolf:
ZD-3681

dmurphy18 (approved):

dmurphy18:
re-run full all

whytewolf:
re-run full all

whytewolf:
:/ I'm not sure what to do with this anymore. The tests that are failing have nothing to do with the change. and they are failing on operating systems that should never touch the file changed.

waynew:
I'm not sure if we have any examples, but one way you could test it is by invoking whatever would render this template and then asserting that the contents of the file are as expected.

sagetherage:
@whytewolf is this still pending and wanted for 2018.3? OR does this need to now go to `master`?

whytewolf:
This is still an issue. however, I have not had the time to write tests for it. If there is anyone that does have the time to write tests for it? that would be great. Otherwise, Maybe we can close this and enter it as a bug instead?

bryceml:
Redhat 6 is no longer supported.

Issue #52930, opened 2019-05-07
unit.serializers.test_serializers.TestSerializers.test_serialize_complex_sls


waynew:
2019.2.1 failed [salt-fedora-28-py2](https://jenkinsci.saltstack.com/job/2019.2.1/job/salt-fedora-28-py2/21/testReport/junit/unit.serializers.test_serializers/TestSerializers/test_serialize_complex_sls), [salt-fedora-29-py2](https://jenkinsci.saltstack.com/job/2019.2.1/job/salt-fedora-29-py2/21/testReport/junit/unit.serializers.test_serializers/TestSerializers/test_serialize_complex_sls)
---

foo: 1 bar: 2 baz: true

```

Traceback (most recent call last):
  File "/tmp/kitchen/testing/tests/unit/serializers/test_serializers.py", line 82, in test_serialize_complex_sls
    assert serialized == '{foo: 1, bar: 2, baz: true}', serialized
AssertionError: foo: 1
bar: 2
baz: true
		
```

Ch3LL:
Fixed by https://github.com/saltstack/salt/pull/52957

Issue #52928, opened 2019-05-07
unit.serializers.test_serializers.TestSerializers.test_serialize_yaml - encrypted_data: !encrypted foo


waynew:
2019.2.1 failed [salt-fedora-28-py2](https://jenkinsci.saltstack.com/job/2019.2.1/job/salt-fedora-28-py2/21/testReport/junit/unit.serializers.test_serializers/TestSerializers/test_serialize_yaml), [salt-fedora-29-py2](https://jenkinsci.saltstack.com/job/2019.2.1/job/salt-fedora-29-py2/21/testReport/junit/unit.serializers.test_serializers/TestSerializers/test_serialize_yaml)
---

encrypted_data: !encrypted foo foo: bar

```

Traceback (most recent call last):
  File "/tmp/kitchen/testing/tests/unit/serializers/test_serializers.py", line 58, in test_serialize_yaml
    assert serialized == expected, serialized
AssertionError: encrypted_data: !encrypted foo
foo: bar
		
```

Issue #52927, opened 2019-05-07
unit.serializers.test_serializers.TestSerializers.test_sls_aggregate


waynew:
2019.2.1 failed [salt-fedora-28-py2](https://jenkinsci.saltstack.com/job/2019.2.1/job/salt-fedora-28-py2/21/testReport/junit/unit.serializers.test_serializers/TestSerializers/test_sls_aggregate), [salt-fedora-29-py2](https://jenkinsci.saltstack.com/job/2019.2.1/job/salt-fedora-29-py2/21/testReport/junit/unit.serializers.test_serializers/TestSerializers/test_sls_aggregate)
---

{a: lol, foo: [hello], bar: [1, 2, 3], baz: {a: 42, b: 666, c: the beast}}

```

Traceback (most recent call last):
  File "/tmp/kitchen/testing/tests/unit/serializers/test_serializers.py", line 187, in test_sls_aggregate
    """).strip() == yamlex.serialize(sls_obj), sls_obj
AssertionError: {a: lol, foo: [hello], bar: [1, 2, 3], baz: {a: 42, b: 666, c: the beast}}
		
```

Ch3LL:
this should be fixed by https://github.com/saltstack/salt/pull/52957

PR #52957, opened 2019-05-09
Set default_flow_style=None in yaml.dump calls

Ch3LL:
### What does this PR do?

In PyYaml 5.1 they changed the `default_flow_style` from None to False. Changing this back to None in salt since this is a breaking change to keep backwards compatibility. Here is the PR/commit that added this: https://github.com/yaml/pyyaml/pull/256 https://github.com/yaml/pyyaml/commit/507a464ce62c933bf667b2296a96ad45f0147873

Would appreciate any further review as I'm not sure if we want to keep backwards compatibility or update salt to ensure `default_flow_style=False` works properly. But I think this is the least obtrusive change for now.

Fixes up some tests on Fedora which has PyYaml 5.1 installed:

https://github.com/saltstack/salt/issues/52930
https://github.com/saltstack/salt/issues/52929
https://github.com/saltstack/salt/issues/52928
https://github.com/saltstack/salt/issues/52927

PyYAML 5.1:

```
>>> test = {'foo': 'bar'}
>>> yaml.dump(test)
'foo: bar\n'
```

PyYAML 3.13:

```
>>> test = {'food': 'bar'}
>>> yaml.dump(test)
'{food: bar}\n'
```

With this change it will always dump like so:

PyYAML 5.1:

```
>>> yaml.dump(test, default_flow_style=None)
'{food: bar}\n'
```

PyYAML 3.13:

```
>>> yaml.dump(test, default_flow_style=None)
'{food: bar}\n'
```

### Tests written?

Yes

### Commits signed with GPG?

Yes

### DIFF
--- a/tests/unit/utils/test_yamldumper.py
+++ b/tests/unit/utils/test_yamldumper.py
@@ -17,7 +17,7 @@ from tests.support.mock import NO_MOCK, NO_MOCK_REASON
 @skipIf(NO_MOCK, NO_MOCK_REASON)
 class YamlDumperTestCase(TestCase):
     '''
-    TestCase for salt.utils.yamlloader module
+    TestCase for salt.utils.yamldumper module
     '''
     def test_yaml_dump(self):
         '''
--- a/tests/unit/utils/test_yamldumper.py
+++ b/tests/unit/utils/test_yamldumper.py
@@ -11,7 +11,7 @@ import salt.utils.yamldumper
 
 # Import Salt Testing Libs
 from tests.support.unit import TestCase, skipIf
-from tests.support.mock import patch, NO_MOCK, NO_MOCK_REASON, mock_open
+from tests.support.mock import NO_MOCK, NO_MOCK_REASON
 
 
 @skipIf(NO_MOCK, NO_MOCK_REASON)
--- a/tests/unit/utils/test_yamldumper.py
+++ b/tests/unit/utils/test_yamldumper.py
@@ -7,6 +7,7 @@
 from __future__ import absolute_import, print_function, unicode_literals
 
 # Import Salt Libs
+import salt.ext.six
 import salt.utils.yamldumper
 
 # Import Salt Testing Libs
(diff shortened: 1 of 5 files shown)


tests/unit/utils/test_yamldumper.py
```diff
@@ -0,0 +1,38 @@
...
+
+# Import python libs
+from __future__ import absolute_import, print_function, unicode_literals
+
+# Import Salt Libs
+import salt.utils.yamldumper
+
+# Import Salt Testing Libs
+from tests.support.unit import TestCase, skipIf
+from tests.support.mock import patch, NO_MOCK, NO_MOCK_REASON, mock_open
```
waynew:
```suggestion
from tests.support.mock import NO_MOCK, NO_MOCK_REASON
```

According to pylint 😛

waynew (changes requested):

tests/unit/utils/test_yamldumper.py
Ch3LL: bah! thanks for that :)

Ch3LL:
started a full test run:

https://jenkinsci.saltstack.com/job/2019.2.1/job/ch3ll-yaml-salt-fedora-29-py2/
https://jenkinsci.saltstack.com/job/2019.2.1/job/ch3ll-yaml-salt-fedora-29-py3/

waynew (approved):

tests/unit/utils/test_yamldumper.py
waynew: BTDT. Too many times 🙃

waynew:
I bet there's a unicode issue on py3. When we're using pytest this assertion style will actually give us information 😄 Of course, right now... 😢 

You're probably aware of both these options, but you could do
`self.assertEqual(salt.utils.yamldumper.dump(data), '{!!python/unicode \'foo\': !!python/unicode \'bar\'}\n')` for now. Or you could do

```
actual = salt.utils.yamldumper.dump(data)
assert actual == '{!!python/unicode \'foo\': !!python/unicode \'bar\'}\n', actual
``` to see what the actual text is.

Ch3LL:
i'm fine with just sticking to just `assert`. I know we do not get more information, but this will make it an easier transition to pytest and will get more information then.

garethgreenaway (approved):

PR #52969, opened 2019-05-10
support args in unless and onlyif 

mchugh19:
### What does this PR do?
support args in unless and onlyif requisites module parsing
also document in release notes

### Previous Behavior
Neon functionality enhancing onlyif and unless only works with execution modules supporting kwargs

### New Behavior
Now execution modules run through onlyif or unless can support modules with args
Also includes release note

### Tests written?
Yes

### Commits signed with GPG?
No

@gtmanfred


gtmanfred (approved):
nice tests!

Ch3LL (approved):

PR #52975, opened 2019-05-11
Fix GCE Windows instances being identified as physical by core grains

kvasserman:
### What does this PR do?
Fixes GCE Windows instances being identified as physical by core grains

### Previous Behavior
GCE Windows instances had 'virtual' grain set to 'physical'

### New Behavior
GCE Windows instances will have 'virtual' grain set to 'gce' (the same as Linux GCE instances)

### Tests written?

No

### Commits signed with GPG?

No

### DIFF
--- a/salt/grains/core.py
+++ b/salt/grains/core.py
@@ -667,6 +667,9 @@ def _windows_virtual(osdata):
     elif 'CloudStack KVM Hypervisor' in productname:
         grains['virtual'] = 'kvm'
         grains['virtual_subtype'] = 'cloudstack'
+    # Google Compute Engine
+    elif 'Google Compute Engine' in productname:
+        grains['virtual'] = 'gce'
     return grains
 
 


Ch3LL (changes requested):
can you write a regression test for this? thanks

kvasserman:
@Ch3LL Maybe. Do you have any documentation on how one should approach doing this in your project? Also, if I do write the test, should it be a separate PR with just the test or do I create a completely new one with both the current changes + the test? Thank you.

Ch3LL:
@kvasserman our test documentation is located here:

https://docs.saltstack.com/en/latest/topics/tutorials/writing_tests.html

my suggestion would be a unit test for this change.

Ch3LL:
also to answer your other question:

> Also, if I do write the test, should it be a separate PR with just the test or do I create a completely new one with both the current changes + the test?

Just keep it in this PR so we can make sure the test passes when it runs :)

kvasserman:
Unfortunately, I don't think I will be able to find the time to write the unit tests for this.

PR #52979, opened 2019-05-12
Updating the docs to reflect a recent change in the way GPG works

ksoviero:
While trying to setup SaltStack on Ubuntu 18.04 LTS, I ran into the issue described in the updated docs. This does not appear to be documented anywhere and it cost me quite a lot of time tracking down the issue.

You don't have to use my exact PR, but please update the docs to reflect the workaround.

Thanks!


salt/renderers/gpg.py
```diff
@@ -51,6 +51,13 @@
     increasing the system entropy. On virtualised Linux systems, this can often
     be achieved by installing the ``rng-tools`` package.
 
+.. note::
+    As of Ubuntu 18.04 LTS you cannot create a key without a passphrase using
```
Ch3LL: First of all thanks for the contribution! We love it when more clarification is added to the docs.

Just want to clarifying one thing. Is it only ubuntu 18.04 thats affected or a certain version of gpg? I think it would be more useful to call it out if its a version of gpg that has this requirement or possibly document the error that you were receiving that lead you to realize it required a passphrase.

Ch3LL (changes requested):

salt/renderers/gpg.py
Ch3LL: @ksoviero bump

DmitryKuzmenko: It looks it depends on gpg version. There are a number of references to 2.0 and 2.1 changes of the related behavior in man page. I'm not sure what's exactly. But what I'd like to suggest here is to not add any not. Just change the original `gpg` command to work as expected with any version. Maybe this one would be good enough?
```
gpg --quick-gen-key --batch --homedir /etc/salt/gpgkeys --passphrase '' <username>
```

DmitryKuzmenko: @saltstack/team-core thoughts?

DmitryKuzmenko:
@ksoviero I didn't noticed this first time. Could you rebase this on master? We've changed to a master development branch.

salt/renderers/gpg.py
Ch3LL: im good with that approach

sagetherage:
closing as abandoned

Issue #48473, opened 2018-07-06
salt-ssh returns with minimal unicode related error message, no useful debugging output


chschmitt:
### Description of Issue/Question

A minion returned the following error when a state was applied via the salt-ssh Python API and state.sls:

```
sanitized_minion_id:
    ----------
    return:
        An Exception occurred while executing state.sls: 'utf-8' codec can't decode bytes in position 1022-1023: unexpected end of data
```
**Note:** There is no useful output in the master logs. Even the running_data/salt-call.log on the minion is not helpful. Is there any chance to recover the full stack trace on the minion to see which piece of code causes the error?

### Setup
Salt is installed via pip in a virtual environment (Python 3.6). There is a custom roster implementation. The top file and some pillars use the Python renderer. Salt-ssh is called via the Python API.

### Steps to Reproduce Issue

Execute the following state. It produces binary output within a cmd.run state. In my original states, I am running some commands that (now) include latin1 or cp1252 strings in the output. I seems that the error depends on the exact sequence/position of the "bad" bytes (which is why the error was not reliably reproducible in the first place).

As a workaround, I redirect the cmd.run stdout and stderr to files.

```
put_script:
  file.managed:
    - name: /tmp/script.py
    - contents: |
        import sys
        sys.stdout.buffer.write(b'\xef'*4096)

run_script:
   cmd.run:
     - name: python3 /tmp/script.py
     - require:
       - put_script
```

Original description: Executing the same state on an identical VM created from the same VM snapshot. Executing the state a second time will not trigger the error.

### Versions Report

Also applies to 2018.3.0

```
Salt Version:
           Salt: 2018.3.2
 
Dependency Versions:
           cffi: Not Installed
       cherrypy: Not Installed
       dateutil: 2.7.2
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.10
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.5.6
   mysql-python: 1.3.12
      pycparser: Not Installed
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: Not Installed
         Python: 3.6.3 (default, Jan  4 2018, 16:40:53)
   python-gnupg: Not Installed
         PyYAML: 3.12
          PyZMQ: 17.0.0
[...]

PR #52985, opened 2019-05-13
Bugfix: Handle multibyte characters spanning blocks of received data

LuAPi:
### What does this PR do?

Probably this one:
https://github.com/saltstack/salt/issues/48473

### Previous Behaviour
This process would fail when a multi-byte unicode character occurs at the end of a received block of data with only part of the character in the current block. The decoding would fail with the exception UnicodeDecodeError and the message 'unexpected end of data' and include the position of failure as the end of the block of data.

### New Behaviour
This fix checks for that specific failure case and will then read another block of data and append it to the current block of data so that the unicode character is no longer split across received data blocks. It then attempts to perform decoding on the combined data block. A loop is used to handle the case where the same problem occurred with the new combined block.

The fix checks that the exception is UnicodeDecodeError and the reason is 'unexpected end of data'. This is to prevent an infinite loop that could occurred if some invalid bytes occurred in the output streams (for example if a program spewed some binary data onto stdout) which would cause a UnicodeDecodeError to happen at the first of the encountered invalid characters. The infinite loop is prevented by testing that the reason is 'unexpected end of data' and not 'invalid start byte'.

### Tests written?

No

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/utils/vt.py
+++ b/salt/utils/vt.py
@@ -652,7 +652,7 @@ class Terminal(object):
                             )
                             break
                         except UnicodeDecodeError as ex:
-                            if (ex.reason == 'unexpected end of data'):
+                            if ex.reason == 'unexpected end of data':
                                 bytes_read += os.read(self.child_fde, maxsize)
                             else:
                                 raise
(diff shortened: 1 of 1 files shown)


LuAPi:
Hold off on merging. I think I might need to handle the case where the attempt to read the next block of data returns 0 bytes because the end of the stream has been reached.

Ch3LL:
@LuAPi I just recently deleted the 2019.2.0 branch since we already released 2019.2.0. When I deleted the branch it automatically closed your PR. Apologies for that, but can you open this PR on the 2019.2 branch? Thanks

PR #52986, opened 2019-05-13
Revert "Make sure --run-expensive runtests.py arg works"

Ch3LL:
This reverts commit 56bf06907993202d53ff477ed723277187237416.

This change was causing all cloud tests to get skipped, because expensive tests where not enabled.

### DIFF
--- a/tests/runtests.py
+++ b/tests/runtests.py
@@ -195,7 +195,6 @@ TEST_SUITES = collections.OrderedDict(sorted(TEST_SUITES_UNORDERED.items(),
 class SaltTestsuiteParser(SaltCoverageTestingParser):
     support_docker_execution = True
     support_destructive_tests_selection = True
-    support_expensive_tests_selection = True
     source_code_basedir = SALT_ROOT
 
     def _get_suites(self, include_unit=False, include_cloud_provider=False,


s0undt3ch:
But if we pass `--run-expensive` on the CLI, they get tested?

Ch3LL:
yes that would work, so the question is do we want to require passing `--run-expensive` alongside cloud-tests which was not the previous behavior?

s0undt3ch:
Are the cloud tests the only expensive tests?

Ch3LL:
```
(salt-py3) ➜  salt git:(ee05da5951) grep -ri @expens tests 
tests/support/helpers.py:            @expensiveTest
Binary file tests/support/__pycache__/helpers.cpython-37.pyc matches
tests/integration/externalapi/test_venafiapi.py:@expensiveTest
tests/integration/cloud/clouds/test_gogrid.py:    @expensiveTest
tests/integration/cloud/clouds/test_vultrpy.py:    @expensiveTest
tests/integration/cloud/clouds/test_digitalocean.py:    @expensiveTest
tests/integration/cloud/clouds/test_linode.py:    @expensiveTest
tests/integration/cloud/clouds/test_msazure.py:    @expensiveTest
tests/integration/cloud/clouds/test_vmware.py:    @expensiveTest
tests/integration/cloud/clouds/test_dimensiondata.py:    @expensiveTest
tests/integration/cloud/clouds/test_gce.py:    @expensiveTest
tests/integration/cloud/clouds/test_joyent.py:    @expensiveTest
tests/integration/cloud/clouds/test_openstack.py:    @expensiveTest
tests/integration/cloud/clouds/test_profitbricks.py:    @expensiveTest
tests/integration/cloud/clouds/test_ec2.py:    @expensiveTest
tests/integration/cloud/clouds/test_oneandone.py:    @expensiveTest
tests/integration/cloud/test_cloud.py:    @expensiveTest
tests/integration/runners/test_state.py:    @expensiveTest
tests/integration/runners/test_state.py:    @expensiveTest
tests/integration/runners/test_state.py:    @expensiveTest
```

a couple of runner tests too

s0undt3ch:
Yeah, lets revert this but re-add the commit on develop?
So that it's explicit?

s0undt3ch (approved):

Ch3LL:
I think the best thing would be to figure out why we cannot have both set, because the code is suppose to set the os.environment to `EXPENSIVE_TESTS=True` if cloud-tests are being run. I just wanted to revert this quickly so we could start getting some test results. I also was not sure if it was worth it to delve into this since we will be migrating to pytest soon.

PR #52989, opened 2019-05-13
Fix true/false shell paths, zcbuildout tests

### LINKED ISSUE #51853: integration tests are failing on py2 on mac
https://jenkinsci.saltstack.com/job/2017.7/job/salt-mac-sierra-py2/104/testReport/

Looking at most of these failures it looks like somehow return values are True/False instead of something iterable. I'll need to take a look at this - there's nothing obvious from the messages, and it's happening only on the macs :suspect:

waynew:
### What does this PR do?

Fixes shell paths - some platforms have `/bin/true` and `/bin/false`. Other platforms put them in `/usr/bin` instead. This asks `which` for the correct path.

When updating the references, I discovered the zcbuildout tests were just skipped. I enabled them and fixed several issues.

Closes #51853 

### Previous Behavior

Tests expected `/bin/false` and `/bin/true`

### New Behavior

Tests ask the environment where `true` and `false` live

### Tests written?

Yes (updated existing tests)

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/modules/zcbuildout.py
+++ b/salt/modules/zcbuildout.py
@@ -186,23 +186,20 @@ class _Logger(object):
 LOG = _Logger()
 
 
-def _encode_string(string):
-    if isinstance(string, six.text_type):
-        string = string.encode('utf-8')
-    return string
-
-
 def _encode_status(status):
-    status['out'] = _encode_string(status['out'])
-    status['outlog_by_level'] = _encode_string(status['outlog_by_level'])
+    if status['out'] is None:
+        status['out'] = None
+    else:
+        status['out'] = salt.utils.stringutils.to_unicode(status['out'])
+    status['outlog_by_level'] = salt.utils.stringutils.to_unicode(status['outlog_by_level'])
     if status['logs']:
         for i, data in enumerate(status['logs'][:]):
-            status['logs'][i] = (data[0], _encode_string(data[1]))
+            status['logs'][i] = (data[0], salt.utils.stringutils.to_unicode(data[1]))
         for logger in 'error', 'warn', 'info', 'debug':
             logs = status['logs_by_level'].get(logger, [])[:]
             if logs:
                 for i, log in enumerate(logs):
-                    status['logs_by_level'][logger][i] = _encode_string(log)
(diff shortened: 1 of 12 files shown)


salt/modules/zcbuildout.py
```diff
@@ -186,23 +186,20 @@ def by_level(self):
...
-    if isinstance(string, six.text_type):
-        string = string.encode('utf-8')
-    return string
-
-
 def _encode_status(status):
-    status['out'] = _encode_string(status['out'])
-    status['outlog_by_level'] = _encode_string(status['outlog_by_level'])
+    if status['out'] is None:
+        status['out'] = None
```
rares-pop: what is if this doing?

waynew: @rares-pop Yeah, that's a fair question. Basically my thought was to just be explicit that `None` was a valid/desired `status['out']`.

My other option of course was to flip the logic and just have: `if status['out'] is not None: <do encoding>`, and then just implicitly leave `status['out']` alone. 

Which would certainly work, just like this does.

The biggest problem for me is just one of priority. I wouldn't even have been in this code in the first place if the tests weren't using `/bin/false`, which was breaking tests elsewhere. I went to fix that problem throughout the codebase and discovered that these tests were just not even turned on. So I figured I'd see if I could improve that a little, without spending *too* much time on it.

Combine that with the fact that this doesn't appear to be a widely (at all?) used module, and the tests have been completely disabled and definitely broken, for quite some time, and according to the modules own docstring it's only a beta support. I also found it kind of difficult to tell at a glance the *why* behind most of this module. This was one of those things. Why isn't it just doing `stderr.decode()`? It would take more time than I had to devote.

Of course, if someone *does* have the time and interest, I would 💯% support their efforts to improve this entire module, and be more than happy to review their PRs 👍

Akm0d (approved):

salt/modules/zcbuildout.py
rares-pop: @waynew - Hehe.. a comment would do it too. :) I have nothing against your change, I was just curious about the reasons behind. Thanks for the reply!

Issue #52076, opened 2019-03-08
integration.modules.test_pip.PipModuleTest


dwoz:
```
integration.modules.test_pip.PipModuleTest.test_chained_requirements__absolute_file_path
--------------------------------------------------------------------------------
develop failed salt-fedora-28-py3
--------------------------------------------------------------------------------
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
string indices must be integers
................................................................................

Traceback (most recent call last):
  File "/tmp/kitchen/testing/tests/integration/modules/test_pip.py", line 275, in test_chained_requirements__absolute_file_path
    self.assertEqual(ret['retcode'], 0)
TypeError: string indices must be integers
```

```
integration.modules.test_pip.PipModuleTest.test_requirements_as_list__non_absolute_file_path
--------------------------------------------------------------------------------
develop failed salt-fedora-28-py3
--------------------------------------------------------------------------------
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
string indices must be integers
................................................................................

Traceback (most recent call last):
  File "/tmp/kitchen/testing/tests/integration/modules/test_pip.py", line 247, in test_requirements_as_list__non_absolute_file_path
    self.assertEqual(ret['retcode'], 0)
TypeError: string indices must be integers
```

waynew:
Also failing here:

develop failed [salt-ubuntu-1804-py3](https://jenkinsci.saltstack.com/job/develop/job/salt-ubuntu-1804-py3/203/testReport/junit/integration.modules.test_pip/PipModuleTest/test_requirements_as_list_of_chains__cwd_not_set__absolute_file_path)
---

string indices must be integers

```

Traceback (most recent call last):
  File "/tmp/kitchen/testing/tests/integration/modules/test_pip.py", line 182, in test_requirements_as_list_of_chains__cwd_not_set__absolute_file_path
    self.assertEqual(ret['retcode'], 0)
TypeError: string indices must be integers
		
```

waynew:
And these cases:

- integration.modules.test_pip.PipModuleTest.test_requirements_as_list__non_absolute_file_path
- integration.modules.test_pip.PipModuleTest.test_pip_install_multiple_editables
[...]

PR #53015, opened 2019-05-14
install python3-venv on ubuntu 18.04

Akm0d:
### What does this PR do?
Installs python3-venv to ubuntu 18.04 for kitchen tests

https://github.com/saltstack/salt/issues/52076

### Tests written?

N/A

### Commits signed with GPG?

Yes


garethgreenaway (approved):

dwoz (changes requested):
Good job tracking this down. This may fix #52076 for you when running kitchen locally however:

- It should be noted that our Jenkins builds don't use this `.kitchen.yml` file. Instead, we use the kitchen configs found in [sre-jenkins](https://github.com/saltstack/sre-jenkins/blob/master/states/services/jenkins-node/kitchen-install.sls)

- Dependencies like this are installed via the states in [salt-jenkins](https://github.com/saltstack/salt-jenkins), if needed it should be installed from a state rather than kitchen's provision command. The reason is too keep with the established convention and keep things as kitchen agnostic as possible.

- It's probably worth syncing with @s0undt3ch on this since it may not be an issue anymore once the nox changes get merged.

Akm0d:
re-run full all

Akm0d:
re-run full all

Akm0d:
re-run full all

PR #52687, opened 2019-04-24
[2019.2] Merge forward from 2018.3 (Includes move to nox)

s0undt3ch:
### What does this PR do?
Merge forward from 2018.3
Conflicts:
* doc/man/salt-api.1
* doc/man/salt-call.1
* doc/man/salt-cloud.1
* doc/man/salt-cp.1
* doc/man/salt-key.1
* doc/man/salt-master.1
* doc/man/salt-minion.1
* doc/man/salt-proxy.1
* doc/man/salt-run.1
* doc/man/salt-ssh.1
* doc/man/salt-syndic.1
* doc/man/salt-unity.1
* doc/man/salt.1
* doc/man/salt.7
* doc/man/spm.1
* salt/modules/beacons.py
* salt/modules/state.py
* tests/unit/modules/test_win_file.py

PR #53021, opened 2019-05-14
[2019.2] MacOS X Kitchen Salt Fixes

garethgreenaway:
### What does this PR do?
MacOS X Kitchen Salt Fixes

N/A

### Tests written?
Yes.  Existing tests fixed.

### Commits signed with GPG?
Yes

### DIFF
--- a/tests/integration/states/test_pkg.py
+++ b/tests/integration/states/test_pkg.py
@@ -1059,7 +1059,7 @@ class PkgTest(ModuleCase, SaltReturnAssertsMixin):
             ret = self.run_state('pkg.removed', name=realpkg)
             self.assertSaltTrueReturn(ret)
 
-    @skipIf(True)
+    @skipIf(True, 'Temporarily disabling this for now')
     @requires_salt_modules('pkg.hold', 'pkg.unhold')
     @requires_system_grains
     def test_pkg_015_installed_held(self, grains=None):  # pylint: disable=unused-argument
--- a/tests/integration/states/test_pkg.py
+++ b/tests/integration/states/test_pkg.py
@@ -1059,6 +1059,7 @@ class PkgTest(ModuleCase, SaltReturnAssertsMixin):
             ret = self.run_state('pkg.removed', name=realpkg)
             self.assertSaltTrueReturn(ret)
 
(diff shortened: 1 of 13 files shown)


waynew (approved):

tests/integration/states/test_pkg.py
```diff
@@ -1059,6 +1059,7 @@ def test_pkg_cap_006_uptodate(self, grains=None):
             ret = self.run_state('pkg.removed', name=realpkg)
             self.assertSaltTrueReturn(ret)
 
+    @skipIf(True)
```
waynew: Is this failing on other OS? It looks like it should only be  running on redhat/debian :suspect:

garethgreenaway: Nope.  And it should be skipped, this was one of the spots where the tests were hanging.  The skipIf was just a test, not a permanent solution :smile:

waynew: That's freaky. I don't see anything in here that should be causing anything to hang - grains.get maybe? You could always test that with a `start = time.time()` and then `if time.time() - start > THRESHOLD: assert False, 'Took too long'`, or something to that effect. I haven't been able to get reliable output on my tests without raising an exception.

waynew: This test really would benefit from something like this though:

    @skipIf(not salt.utils.platform.is_darwin(), 'Test only available on macOS')

s0undt3ch (approved):

s0undt3ch (approved):

waynew:
Looks like the current failure is due to another lib no longer supporting Python 3.4 - docker, at least

s0undt3ch:
That issue will go away once 2019.2 is on nox :/

waynew:
@s0undt3ch  Is there currently a merge forward/issue to track for that?

s0undt3ch:
https://github.com/saltstack/salt/pull/52687

s0undt3ch (approved):

s0undt3ch:
Now passing. Merging.

s0undt3ch:
Should this go into 2019.2.1?

PR #53042, opened 2019-05-15
minor tweaks to master initialization / EventReturn

mattp-:
### What does this PR do?
- moved creating a masterMinion object for EventReturn into the child proc to avoid loader churn (funcs, grains, etc). Theres no reason for this to happen in parent proc then get pickled down post fork. This should improve master startup time.
- move EventReturn creation to _after_ ReqServer, since it emits events on initialization.

n/a

### Tests written?
Not really applicable to this change.

### Commits signed with GPG?
No

### DIFF
--- a/salt/master.py
+++ b/salt/master.py
@@ -705,10 +705,6 @@ class Master(SMaster):
             log.info('Creating master maintenance process')
             self.process_manager.add_process(Maintenance, args=(self.opts,))
 
-            if self.opts.get('event_return'):
-                log.info('Creating master event return process')
-                self.process_manager.add_process(salt.utils.event.EventReturn, args=(self.opts,))
-
             ext_procs = self.opts.get('ext_processes', [])
             for proc in ext_procs:
                 log.info('Creating ext_processes process: %s', proc)
@@ -746,6 +742,11 @@ class Master(SMaster):
                 kwargs=kwargs,
                 name='ReqServer')
 
+            if self.opts.get('event_return'):
+                log.info('Creating master event return process')
+                self.process_manager.add_process(salt.utils.event.EventReturn, args=(self.opts,))
+
+            log.info('Creating FileServerUpdate process')
             self.process_manager.add_process(
                 FileserverUpdate,
                 args=(self.opts,))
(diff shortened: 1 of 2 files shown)


DmitryKuzmenko (approved):
Sounds reasonable, looks good.

rares-pop:
I see you did this in 'develop', but all releases can benefit from it. Can you add some flags so this can be backported?

mattp-:
@rares-pop I am not a salt employee / unable to add flags to PRs. maybe @DmitryKuzmenko can help?

DmitryKuzmenko:
To decide what branch this change should go to we need to see the bug description and reproducing steps i.e. a properly formed issue.
Here I just see the code that looks correct for me. I'm not sure how critical is that.
So @mattp- could you please describe the original issue you're fixing here or create a separate GH issue to track this?

@saltstack/team-core could anybody please confirm I'm right here?

mattp-:
@DmitryKuzmenko wasn't actually a bug, behavior should be the same before as after, just something I uncovered tuning our test suite. i think given the usual rules just sticking it in develop for neon and not backporting makes sense.

DmitryKuzmenko:
@mattp- thank you for explanation. Since this touches major business code and it doesn't fix anything critical I see no reason to speed it up to the current releases. Let it leave in develop and show it as good as it looks.

dwoz (changes requested):
@mattp- All too often simple changes like these cause hard to track down bugs. If this is truly a bugfix or is going to get back ported into any stable branch it should have some kind of regression test.

mattp-:
@dwoz it isn't a bugfix, just a minor optimization; i think we can disregard / remove the bugfix backport label. I'm not exactly sure how you would write a test to assert the change, or if you had one it would be very useful

waynew:
Hey @mattp- looks like it's been a while here - would you be able to rebase this on the `master` branch?

One thought about how you could test this is by mocking some of the calls and ensuring that the calls happen in the correct order. I'd probably make liberal use of comments in the test explaining why the order is important.

Does that make sense?

waynew:
@mattp- hope you're doing well - do you have any interest in driving this to completion?

sagetherage:
closing as abandoned

PR #53068, opened 2019-05-16
Saltcheck merge

mchugh19:
### What does this PR do?
supersedes #52434 #52541 and #53024

- Optimize test copying (replace `cp.cache_master` with logic to `cp.cache_dir` on used salt states) @lomeroe
- Create saltcheck wrapper for use in salt-ssh. Handles cp.cache_dir functionality by creating a tarball, copying, and extracting on ssh minion. Then calls the regular saltcheck functions.
- allow custom saltcheck test locations inside a state/formula's folder @lomeroe
- provide more information on invalid tests @lomeroe

### Previous Behavior
1. saltcheck, being unable to run `cp.cache_*` from a salt-ssh minion, was not able to locate test files and did not function.
2. saltcheck was only able to parse an `assertion_section` of a single depth
3. `cp.cache_master` takes more time and I/O than necessary

### New Behavior
1. Wrapper written to gather needed state directories, tarball them, copy tar to salt-ssh minion, extract to local cache directory, and cleanup. Then hands over saltcheck running to the local saltcheck module. 
2. `assertion_section` can now be nested dicts or lists
3. Now uses `cp.cache_dir` on needed state directories

### Tests written?
Yes - new integration and ssh integration tests

### Commits signed with GPG?
No


mchugh19:
@lomeroe does this look like a good merge of the updates?

lomeroe:
A cursory look at it seems good, I'll test it out to make sure I don't find anything.

One other option that I think we should add is a ``assertion_section_delimiter`` to pass to the ``traverse_dict_and_list`` function (in the case where a ':' is in the dict key (which is the case in some output from lgpo.get at least))

```
return_value = salt.utils.data.traverse_dict_and_list(value,
                                                                  assertion_section,
                                                                  default=False,
                                                                  delimiter=assertion_section_delimiter)
```

with a test def example like:
```
Example with a nested assertion_section
---------------------------------------

.. code-block:: yaml

    validate_smb_signing:
      module_and_function: lgpo.get
      args:
        - 'Machine'
      kwargs:
        return_full_policy_names: True
      assertion: assertEqual
      expected-return: Enabled
      assertion_section: "'Computer Configuration'|'Microsoft network client: Digitally sign communications (always)'"
      assertion_section_delimiter: '|'
```

I updated my original PR to use that too, so you can pull it from there into this one too.

mchugh19:
@lomeroe I did see a problem where testing multiple states only finds tests for the first; example `salt-call saltcheck.run_state_tests ntp,ntp.server` only shows results for ntp. If the order is reversed, then it only shows results for ntp.server. 

It seems to be that from line 350, after a state has been added for testing, then tests are run for that state, the next state in the for loop is fails the `state.show_low_sls` around line 878. 

Tests are run with a salt client using --local. So it may be that once tests are run, that --local setting is preserved and subsequent states aren't able to progress past the now failing `state.show_low_sls` dependency. 

Thoughts?

lomeroe:
I haven't been able to duplicate this yet in a test environment (all masterless, so maybe that is playing a part).  I'll test in a mastered minion environment tomorrow and see if I can get the same issue to appear.

I did find a PY3 issue line 409 ```out_lists.sort()``` will fail on PY3.

Changing that to
```
out_list = sorted(out_list, key=lambda x:sorted(x.keys()))
```

Will make both PY2 and PY3 behave the same

mchugh19:
> I did find a PY3 issue line 409 `out_lists.sort()` will fail on PY3.

I wasn't aware of the py3 sort() change. Are you sure of that? https://www.tutorialspoint.com/python3/list_sort.htm makes it look like it should be ok.

mchugh19:
I think I've found the issue with the `ntp,ntp.server` styled multiple tests. The issue was that `file_client` was being set to local after the first test run. Thus, subsequent tests failed at the `state.show_low_sls` as the local change meant no states could be found.

Changing the __opts__ call to deepcopy, resolves the problem. 
```
 conf_file = copy.deepcopy(__opts__['conf_file'])
 local_opts = salt.config.minion_config(conf_file)
 local_opts['file_client'] = 'local'
```

Somewhat relatedly, the use of `state.show_low_sls` and `cp.list_states` also breaks salt-ssh minions as they don't return data for those minions. So I've also worked around that, by running those commands from the saltcheck ssh wrapper, saved the results to a file, and added them into the tarball that is copied to the minion. Then in the normal saltcheck.py module, if an ssh-minion is detected, rather than running the destined to fail modules, it instead reads in the copied results and progresses as (mostly) normal. 

So with this (possible hack), I think all known use cases are now covered. 

Also @lomeroe, I've adjusted the logic in add_test_files_for_sls to only call `cp.cache_dir` when needed rather than for each state file. So this now works, but let me know if this overcomplicates things too much.

lomeroe:
@mchugh19 I was able to duplicate this on a mastered minion and I think you are right as to what is occurring.  I think we will have to render/gather everything first and add a way to track which sls a tst file was pulled in for so all the tests can be executed at once...

lomeroe:
@mchugh19 sorry, I missed your updates before my last response, you can ignore my suggestions since you already got it working...

PY3 - it's not that you can't sort a list, but a list of dicts won't sort...

I'll test out your most recent changes locally and let you know my results

mchugh19:
> PY3 - it's not that you can't sort a list, but a list of dicts won't sort...

Ah, gotcha! Sort fix pushed.

lomeroe (approved):
LGTM and module works as expected in my environment

Ch3LL:

Ch3LL (commented):
any chance you could add some integration tests. we currently have some ssh tests here: tests/integration/ssh where you could add a test to ensure we have test coverage for this new saltcheck wrapper for ssh.

One other note, in order to get the ssh tests to start running make sure you pass `--ssh` to the test runner.

mchugh19:
> any chance you could add some integration tests.

Working on it!

mchugh19:
@Ch3LL I think I need some assistance with the ssh integration tests. I worked up standard integration tests, which run as expected. But the same test run over ssh, it doesn't work properly. 

To explain in more detail...
The `saltcheck.run_test` function takes an kwargument of a single test. Because you pass the test params to the function, there is no other dependency on file hierarchy or anything external. Thus the standard integration test of 
```
saltcheck_test = {"module_and_function": "test.echo",
                            "assertion": "assertEqual",
                            "expected-return": "Test Works",
                            "args":["Test Works"] }
ret = self.run_function('saltcheck.run_test', test=saltcheck_test)
# assert expected to fail, but in failing displays the ret output
self.assertEqual(ret, "run_test")
```
works just fine.

However, when run as an ssh integration test, the same code returns
```
AssertionError: u'Test argument must be a dictionary: None' != u'run_test'
```
Meaning that the test argument is `None` rather than the expected dict. This is weird as the same argument works in the non ssh integration test. 

To help troubleshoot further, I tried running the integration environment in interactive mode. `./tests/runtests.py --interactive --ssh`, I then manually created the `/tmp/salt-tests-tmpdir/scripts/cli_salt_ssh.py` which seems to be created when running the ssh integration tests. I then used that file to manually run the `saltcheck.run_test` function.

```
python /tmp/salt-tests-tmpdir/scripts/cli_salt_ssh.py -ldebug -c /tmp/salt-tests-tmpdir/config -i --priv /tmp/salt-tests-tmpdir/config/key_test --roster-file /tmp/salt-tests-tmpdir/config/roster --out=json localhost saltcheck.run_test 'test={"module_and_function": "test.echo","assertion": "assertEqual","expected-return": "This works!","args":["This works!"] }'
```

At this point, the result comes back as expected! 
```
Result of run_ssh for command ' -W -ldebug -c /tmp/salt-tests-tmpdir/config -i --priv /tmp/salt-tests-tmpdir/config/key_test --roster-file /tmp/salt-tests-tmpdir/config/roster --out=json localhost saltcheck.run_state_tests validate-saltcheck': {
    "localhost": [
        {
            "validate-saltcheck": {}
        },
        {
            "TEST RESULTS": {
                "Failed": 0,
                "Missing Tests": 1,
                "Skipped": 0,
                "Passed": 0,
                "Execution Time": 0.0
            }
        }
    ]
}
```


So in summary, I have an integration test, which works just fine as a standard integration test. The same test can be run from the ssh integration environment manually. However, when run as an ssh integration test, the kwarg content seems to be swallowed and not send to the execution module and function. :(

At this point, I think I'm stuck trying to debug the salt integration framework more than I'm able to progress on integration tests. Can anyone assist?

tests/integration/ssh/test_saltcheck.py
```diff
@@ -0,0 +1,38 @@
...
+    '''
+    def test_saltcheck_run_test(self):
+        '''
+        test saltcheck.run_test with salt-ssh
+        '''
+        saltcheck_test = {"module_and_function": "test.echo",
+                        "assertion": "assertEqual",
+                        "expected-return": "Test Works",
+                        "args":["Test Works"] }
+        ret = self.run_function('saltcheck.run_test', test="saltcheck_test")
```
mchugh19: This is obviously just passing the string "saltcheck_test", but this was set to attempt to debug the `None` test argument as described in the PR comments.

mchugh19:
@Ch3LL Happy to help here, but can anyone assist with the ssh integration tests? It seems the framework operates a bit differently, and I'm having trouble getting any kwargs passed into it. I've uploaded some ssh integration tests that don't pass, so this will need a slight adjustment before progressing further.

Ch3LL:
Looks like `kwargs` needs to be added to `run_ssh` call https://github.com/saltstack/salt/blob/develop/tests/support/case.py#L880

its only accounting for args currently. If that doesn't point you in the right direction let me know and i can dive in a bit deeper to the problem.

mchugh19:
Success! The case script was modified to support kwargs, and the saltcheck logic for detecting if called from salt-ssh was adjusted to support the cli_salt_ssh.py environment. With that, integration tests now succeed for both the standard and salt-ssh cases.

mchugh19:
I noticed tests failing on:
```
Starting unit.test_module_names.BadTestModuleNamesTestCase.test_module_name_source_match Tests
...
  File "/root/tech/gits/salt/tests/unit/test_module_names.py", line 242, in test_module_name_source_match
    assert not errors, _format_errors(errors)
AssertionError: The following 1 test module(s) could not be matched to a source code file:

integration.ssh.test_saltcheck (expected: salt/ssh/saltcheck.py)
```

so I added "integration.ssh.test_saltcheck" to the list of ignored names in test_module_names.py. A comment above referenced using tests/filename_map.yml, but that didn't seem to help with the ssh integration tests. If there's a better method, just let me know.

mchugh19:
Windows test errors seem to be unrelated to this PR. If there's any feedback on some of the above comments, let me know. Otherwise this looks good to me.

Ch3LL:
im going to re-run the tests. The windows issues are fixed now.

mchugh19:
Windows worked (yay!) but what appears to be an unrelated failure failed centos (boo!)

Ch3LL:
We currently are putting all of our efforts into stabilizing the neon and 2019.2.1 tests right now, so it might be a little bit before we can get this merged in. On neon in particular we are skipping the tests that are failing and then will go back through and re-enable them as we fix. By doing this we will be able to now focus on all of our CI issues, such as a SSH timeout that we keep running into sometimes and then fix the tests and ensure they are passing and we have not introduced anything new to the environment. I will go ahead and review this PR right now as its ready, but just wanted to give you a heads up that it might take a while before we can get it merged while we work on the test infrastructure to ensure its stable going forward.

salt/modules/saltcheck.py
```diff
@@ -840,53 +868,123 @@ def _load_file_salt_rendered(self, filepath):
...
-                init_path = base_path + os.sep + sls_split[2] + os.sep + 'saltcheck-tests' + os.sep + 'init.tst'
-                name_path = base_path + os.sep + 'saltcheck-tests' + os.sep + sls_split[2] + '.tst'
-                if os.path.isfile(init_path):
-                    self.test_files.append(init_path)
-                    log.info("Found test init: %s", init_path)
-                if os.path.isfile(name_path):
-                    self.test_files.append(name_path)
-                    log.info("Found test named: %s", name_path)
+        salt_ssh = False
+        if 'running_data/var/run/salt-minion.pid' in __opts__.get('pidfile', False):
```
Ch3LL: this check might not work across different OS platforms

Ch3LL (commented):
thanks for the new tests :)

Ch3LL:
ping @wcannon any chance you have time to look at this?

salt/modules/saltcheck.py
mchugh19: Yep. That line is the only unique info I can find to detect that salt-client was launched from a salt-ssh session. If there's an alternative I'm all ears. While likely not an issue for Windows machines much these days, eventually they may also be ssh managed and this will need to be adjusted.

Ch3LL: I took a glance at opts as well and couldn't find anything that stands out. @thatch45 any better ideas on how to detect an ssh minion here?

mchugh19: Any thoughts team? The requirement is to detect that this is running inside of a salt-ssh session.

salt/modules/saltcheck.py
```diff
@@ -68,6 +96,9 @@
     (str) Optional keyword used to parse the module_and_function return. If a salt module
     returns a dictionary as a result, the ``assertion_section`` value is used to lookup a specific value
     in that return for the assertion comparison.
+**assertion_section_delimiter:**
+    (str) Optional delimiter to use when splitting a neseted structure.
```
max-arnold: Typo: `neseted` -> `nested`

mchugh19: Gah! Thanks, corrected!

Ch3LL:
ping @mchugh19 looks like you have ported some of your PRs to master (which i super appreciate) looks like this one needs to be ported over to the master branch as well. Anything i can do to help you out in this process?

mchugh19:
@Ch3LL there were a few saltcheck PRs in neon which this then builds on top of. Would it be acceptable to just roll them all up into a comprehensive PR to master?

Ch3LL:
I personally always prefer smaller PRs, but if its not too big that should be fine.

mchugh19:
Closing in favor of #55613

PR #53074, opened 2019-05-16
Add visible_name and firstname arguments

timdufrane:
Neither was being respected, resulting in an always blank Visible Name in Zabbix.

### What does this PR do?
Allows specification of Visible Name in Zabbix
N/A
### Previous Behavior
Visible Name was always blank

### New Behavior
Visible Name is now set as specified

### Tests written?

No

### Commits signed with GPG?

No

### DIFF
--- a/salt/states/zabbix_host.py
+++ b/salt/states/zabbix_host.py
@@ -323,7 +323,7 @@ def present(host, groups, interfaces, **kwargs):
                                                      proxy_hostid=proxy_hostid,
                                                      inventory=new_inventory,
                                                      visible_name=new_visible_name,
-                                                     firstname=new_first_name
+                                                     firstname=new_first_name,
                                                      **connection_args)
 
         if 'error' not in host_create:
--- a/salt/states/zabbix_host.py
+++ b/salt/states/zabbix_host.py
@@ -185,10 +185,11 @@ def present(host, groups, interfaces, **kwargs):
     if inventory is None:
         inventory = {}
         
+    new_visible_name = ''
     if 'visible_name' in kwargs:
         new_visible_name = kwargs['visible_name']
     if 'firstname' in kwargs:
-        new_first_name = kwargs['firstname']
+        new_visible_name = kwargs['firstname']
 
     # Create dict of requested inventory items
     new_inventory = {}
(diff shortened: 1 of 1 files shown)


salt/states/zabbix_host.py
```diff
@@ -184,6 +184,13 @@ def _interface_format(interfaces_data):
         inventory = kwargs['inventory']
     if inventory is None:
         inventory = {}
+        
+    new_visible_name = ''
```
rares-pop: you should use:
new_visible_name = kwargs.get('visible_name', '')
because what you do is an anti-pattern in python:
https://docs.quantifiedcode.com/python-anti-patterns/correctness/not_using_get_to_return_a_default_value_from_a_dictionary.html

timdufrane: Fair, and fixed - though I would note that there is literally no use of kwargs.get anywhere else in this file. I was simply following the established pattern of the file.

rares-pop: Maybe that was declared an anti-pattern after. :)

rares-pop (approved):

salt/states/zabbix_host.py
timdufrane: I'll have to grab a Python environment for doing this in the future, apologies for all the commits. Doing this while overtired isn't a great idea.

twangboy (approved):

salt/states/zabbix_host.py
```diff
@@ -184,6 +184,10 @@ def _interface_format(interfaces_data):
         inventory = kwargs['inventory']
     if inventory is None:
         inventory = {}
+
+    new_visible_name = kwargs.get('visible_name', '')
+    new_first_name = kwargs.get('first_name', '')
```
twangboy: Please add some documentation for these two new parameters.

twangboy (changes requested):

twangboy:
@timdufrane Thanks for the new feature. Can we get at least some unit tests for this as well?

salt/states/zabbix_host.py
timdufrane: `visible_name` is documented at the top of the file:
`:param visible_name: Optional - string with visible name of the host, use 'visible_name' instead of 'name' \`
I can comment that this is the mapping for that directive, but it seems a bit redundant...
Will add `firstname` to the documentation of that method as well, however.

timdufrane:
Working on unit tests for states now.

timdufrane:
Resubmitting when test coverage is better

PR #53100, opened 2019-05-17
salt.modules.network: Add functions to get minion's networks

terminalmage:
### What does this PR do?

Two new functions added which return information about the networks to which the minion belongs. Additionally, the `interface` arg to both new functions supports multiple interfaces and globbing. This additional functionality has also been added to the `ip_addrs` and `ip_addrs6` functions, and all four of these functions now use common code for filtering the interface data. DRY.

N/A

### Tests written?

Yes

### Commits signed with GPG?

No

### DIFF
--- a/salt/modules/network.py
+++ b/salt/modules/network.py
@@ -1275,6 +1275,10 @@ def ip_addrs(interface=None, include_loopback=False, cidr=None, type=None):
     which are within that subnet. If 'type' is 'public', then only public
     addresses will be returned. Ditto for 'type'='private'.
 
+    .. versionchanged:: Neon
+        ``interface`` can now be a single interface name or a list of
+        interfaces. Globbing is also supported.
+
     CLI Example:
 
     .. code-block:: bash
@@ -1305,6 +1309,10 @@ def ip_addrs6(interface=None, include_loopback=False, cidr=None):
     then only IP addresses from that interface will be returned.
     Providing a CIDR via 'cidr="2000::/3"' will return only the addresses
     which are within that subnet.
+
+    .. versionchanged:: Neon
+        ``interface`` can now be a single interface name or a list of
+        interfaces. Globbing is also supported.
 
     CLI Example:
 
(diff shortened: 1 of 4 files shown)


salt/modules/network.py
```diff
@@ -1887,3 +1895,55 @@ def iphexval(ip):
...
+    CLI Example:
+
+    .. code-block:: bash
+
+        salt '*' network.list_networks
+        salt '*' network.list_networks interface=docker0
+        salt '*' network.list_networks interface=docker0,enp*
+        salt '*' network.list_networks interface=eth*
+    '''
+    return salt.utils.network.ip_networks(
```
mirceaulinic: Maybe use `__utils__['network.ip_networks']` instead, to make it easier to port the utils functions?

mirceaulinic: Side question: I've always found it a bit strange having functions that simply invoke util functions and only forward arguments. For some time I've been thinking about a module, say `utils.py`, e.g.,

```
def fun(name, *args, **kwargs):
    return __utils__[name](*args, **kwargs)
```

Which gives you access to any util functions (and plenty are quite handy) without having to write almost-dummy module functions like that. What would be your thoughts on this? I can see any arguments against, but you have more experience and let me know if there's anything I'm missing.

terminalmage: That could be useful for troubleshooting purposes.

I get where you're coming from with regard to having a function that just invokes a utils function, but a lot of this code comes from before `__utils__` was a thing in execution modules. The utils libraries were mainly created so that code exposed to the user via execution modules, as well as code elsewhere in salt, could both import from common code. Simply moving it to salt/modules/network.py would make it unavailable to code that does not have access to the remote execution modules.

mirceaulinic: I agree, but this codes goes into the develop branch, where `__utils__` is a thing - or are you referring to something different? In this state it's hard (read: impossible) to port this to older versions.

terminalmage: Yes, I'm referring to something different. I'm saying that putting the code directly into salt/modules/network.py would make unavailable to code that doesn't have access to the execution modules.

mirceaulinic: This is not what I was suggesting: I wasn't objecting about the code being under utils, but the way the util functions are called from the execution module. Or to put this in code:

Instead of

```python
return salt.utils.network.ip_networks(
```

(which cannot be used when porting the module as-is), might be better to use:

```python
return __utils__['network.ip_networks'](
```

Which make the module portable (when porting the utils under `salt://_utils`). Does this make more sense? `salt.utils.network.ip_networks` is really a major PITA when you can't upgrade to the latest Salt version and you need to patch an utils module/function.

terminalmage: Ah, OK. Sorry for the misunderstanding. I 100% agree. I just pushed another commit which changes much of the utils module usage to the `__utils__` dunder.

twangboy:
@terminalmage Looks like have some lint

terminalmage:
@twangboy That was a alse positive, I've pushed an update that ignores that particular failure.

twangboy (approved):

Ch3LL:
@terminalmage looks like there is a merge conflict, would you be willing to fix that? 

also, i plan on creating the neon branch today, i will migrate this PR to the new branch after its created thanks :)

terminalmage:
Rebased.

Ch3LL:

terminalmage:
@Ch3LL It looked fine to me. I noticed a docstring that had IPv4 instead of IPv6 in it, so I fixed that and rebased, and re-ran tests. They look good to me so I pushed. Assuming tests pass I think this is good to merge.

twangboy (dismissed):

terminalmage:
I've rebased and re-pointed at `master`.

terminalmage:
Rebased again. Any chance we can get this merged?

terminalmage:
Can we get this re-reviewed and try to get it merged? It has been nearly 11 months.

DmitryKuzmenko:
@terminalmage thanks for rebasing this, I've also blackened it for you. Could you please address the lint issues?

terminalmage:
One of the lint warnings was due to black itself inserting a comma. Both warnings should now be fixed.

terminalmage:
Rebased again.

dwoz (approved):

PR #53126, opened 2019-05-20
Adds btrfs and xfs to parted module

brejoc:
### What does this PR do?

Both filesystem types are supported by parted and mkfs.

None

### Previous Behavior

btrfs and xfs where not supported for `parted` and `mkfs.xxx`.

### New Behavior

btrfs and xfs are now supported.

### Tests written?

No

### Commits signed with GPG?

Yes


brejoc:
If this change is fine for you, could we also have that in the upcoming Neon branch?

twangboy (approved):

brejoc:
The failure in `py3-centos-7` seems to happen in `install-dev-moto`. I don't think that this is related to this change.

Ch3LL:
We had some issues with the doc/lint tests the other day. I went ahead and restarted them. I plan on creating the neon branch today so I will migrate this PR to that branch after its created to ensure its in that release. thanks

garethgreenaway (approved):

Ch3LL:

Ch3LL (approved):

Ch3LL (commented):
When I initially reviewed this I thought it would be difficult to test this since our testing infrastructure does not include these file system types, but I'm wondering if there would be value in adding atleast a unit test with a call to mkfs(). which calls `_is_fstype`?

aplanas:
If we go for this path, can I suggest to make the full list?

"""
btrfs, nilfs2, ext4, ext3, ext2, fat32, fat16, hfsx, hfs+,
hfs, jfs, swsusp, linux-swap(v1), linux-swap(v0), ntfs, reiserfs, hp-ufs,
sun-ufs, xfs, apfs2, apfs1, asfs, amufs5, amufs4, amufs3, amufs2, amufs1,
amufs0, amufs, affs7, affs6, affs5, affs4, affs3, affs2, affs1, affs0,
linux-swap, linux-swap(new), linux-swap(old)
"""

brejoc:
> When I initially reviewed this I thought it would be difficult to test this since our testing infrastructure does not include these file system types, but I'm wondering if there would be value in adding atleast a unit test with a call to mkfs(). which calls `_is_fstype`?

@Ch3LL What exactly do you have in mind? Do you want to check someone accidentally or on purpose removes some of the fs-types?

We could do an integration test here, skip the device validation and try to create a filesystem in a file. But if this is not supported by the test systems, then this of course not possible.

brejoc:
> If we go for this path, can I suggest to make the full list?

@aplanas Good point! But I'd suggest to do this in a second PR to get this merged now.

Ch3LL:
yeah to be honest I thought my request was probably overkill so I'm okay to not add something here unless anyone from @saltstack/team-core can think of a valuable test to add here?

brejoc:
This PR is replaced by https://github.com/saltstack/salt/pull/55209.

Issue #52922, opened 2019-05-07
integration.modules.test_shadow.ShadowModuleTest.test_unlock_password


waynew:
2019.2.1 failed [salt-fedora-29-py3](https://jenkinsci.saltstack.com/job/2019.2.1/job/salt-fedora-29-py3/21/testReport/junit/integration.modules.test_shadow/ShadowModuleTest/test_unlock_password)
---

False is not true

```

Traceback (most recent call last):
  File "/tmp/kitchen/testing/tests/support/helpers.py", line 129, in wrap
    return caller(cls)
  File "/tmp/kitchen/testing/tests/integration/modules/test_shadow.py", line 173, in test_unlock_password
    self.assertTrue(self.run_function('shadow.unlock_password', [self._test_user]))
AssertionError: False is not true
		
```

Ch3LL:
confirmed they are passing now here:

https://jenkinsci.saltstack.com/job/2019.2.1/job/salt-fedora-29-py3/57/consoleFull

DmitryKuzmenko:
The same is in 2017.7:

https://jenkinsci.saltstack.com/job/2017.7/job/salt-fedora-29-py2/107/testReport/junit/integration.modules.test_shadow/ShadowModuleTest/test_unlock_password/

Need a backport?

PR #53141, opened 2019-05-20
Check for all non-word characters when calling secure_password

Ch3LL:
### What does this PR do?
The `integration.modules.test_shadow` tests are sometimes failing on fedora. What I discovered is `shadow.gen_passwd` is sometimes returning `None`. It returns `None` anytime there is a `_` character in the the `crypt.crypt()` call. below will demonstrate:

```
[root@ip-10-27-42-60 testing]# python -c "import crypt; print(crypt.crypt('blah', '\$6\$3m7s2p4SXgL_mYOL'))"
None
[root@ip-10-27-42-60 testing]# python -c "import crypt; print(crypt.crypt('blah', '\$6\$3m7s2p4SXgLmYOL'))"
$6$3m7s2p4SXgLmYOL$VqLpFYq8D5gy9xhh3x3R00s4NNJgS1JNIcOWguXfIBtKX3oCKZLCbvzrUOQCkt69yMmJSOTdAbImDOYW.OyW51
[root@ip-10-27-42-60 testing]# 
```

As you can see above when we remove the `_` we get whats expected. This PR now checks for all non-word characters and `_` and removes them.

https://github.com/saltstack/salt/issues/52922

### Previous Behavior
`integration.modules.test_shadow` tests sometimes failed. I even saw this behavior on about 1/10 runs when i was running the test suite.

### New Behavior
`integration.modules.test_shadow` tests pass. I have not seen a flaky test failure during my testing so far after this change. Will monitor jenkins results after this is merged.

### Tests written?

Yes - and also hopefully fixes some flaky tests.

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/utils/pycrypto.py
+++ b/salt/utils/pycrypto.py
@@ -54,7 +54,7 @@ def secure_password(length=20, use_random=True):
                     except UnicodeDecodeError:
                         continue
                 pw += re.sub(
-                    salt.utils.stringutils.to_str(r'\W'),
+                    salt.utils.stringutils.to_str(r'[\W_]'),
                     str(),  # future lint: disable=blacklisted-function
                     char
                 )
--- a/tests/unit/utils/test_pycrypto.py
+++ b/tests/unit/utils/test_pycrypto.py
@@ -1,0 +1,43 @@
+# -*- coding: utf-8 -*-
+
+# Import python libs
+from __future__ import absolute_import, print_function, unicode_literals
+import logging
+import re
+
+# Import Salt Libs
+import salt.utils.pycrypto
+
+# Import Salt Testing Libs
+from tests.support.unit import TestCase
+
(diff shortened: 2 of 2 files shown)


twangboy (approved):

Ch3LL:
started a full fedora py3 test run here: https://jenkinsci.saltstack.com/job/2019.2.1/job/ch3ll-shadow-salt-fedora-29-py3/1/

waiting on results

waynew:
It's (obviously?) not part of this PR, and maybe it would be worth opening a new issue or something, but that generation method feels weird to me. I think because it's discarding values outside of the text range instead of maybe doing something like modulo? Or alternatively, [something like this](https://stackoverflow.com/a/17958373/344286).

Anyway, just wanted to capture that thought somewhere 🙂

PR #53142, opened 2019-05-21
Use six.reraise for the greater good

waynew:
### What does this PR do?

Uses `six.reraise` in all the places.

n/a

### Previous Behavior

When re-raising an exception in Python2, `raise exc` obliterates the stack trace. It should be obvious why this is undesirable.

### New Behavior

Now legacy Python (2) and Python 3 both include the underlying stack trace, using six.reraise

### Tests written?

Not that kind of a change

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/cloud/clouds/libvirt.py
+++ b/salt/cloud/clouds/libvirt.py
@@ -58,8 +58,9 @@ Tested on:
 from __future__ import absolute_import, print_function, unicode_literals
 
 import logging
-import uuid
 import os
+import sys
+import uuid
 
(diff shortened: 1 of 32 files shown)


garethgreenaway (approved):

s0undt3ch (approved):

twangboy (approved):

s0undt3ch:
Ported this PR into 2019.2.1 in https://github.com/saltstack/salt/pull/53271

PR #53157, opened 2019-05-21
Fix `unit.modules.test_win_file.WinFileTestCase.test_issue_52002_check_file_remove_symlink`

twangboy:
### What does this PR do?
Attempts to fix the following test on Windows:
```
unit.modules.test_win_file.WinFileTestCase.test_issue_52002_check_file_remove_symlink
```
It's failing with the following error:
```
  File "C:\\Users\\ADMINI~1\\AppData\\Local\\Temp\\kitchen\\testing\tests\support\helpers.py", line 129, in wrap
    return caller(cls)
  File "c:\users\admini~1\appdata\local\temp\kitchen\testing\tests\unit\modules\test_win_file.py", line 76, in test_issue_52002_check_file_remove_symlink
    self.assertTrue(win_file.symlink(target, symlink))
  File "C:\\Users\\ADMINI~1\\AppData\\Local\\Temp\\kitchen\\testing\salt\modules\win_file.py", line 1158, in symlink
    exc.strerror
CommandExecutionError: Could not create 'c:\users\admini~1\appdata\local\temp\base-z7v221\child 2\link' - [1314] A required privilege is not held by the client.	
```
But only when run from kitchen. I believe this has something to do with winrm... it may run as an admin, but without elevating the accounts privileges. This PR attempts to do so before trying to create the symlink.

Jenkins failure

### Tests written?
Fixes a test... hopefully

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/modules/win_file.py
+++ b/salt/modules/win_file.py
@@ -69,9 +69,10 @@ HAS_WINDOWS_MODULES = False
 try:
     if salt.utils.platform.is_windows():
         import win32api
-        import win32file
         import win32con
-        from pywintypes import error as pywinerror
+        import win32file
+        import win32security
+        import salt.platform.win
         HAS_WINDOWS_MODULES = True
 except ImportError:
     HAS_WINDOWS_MODULES = False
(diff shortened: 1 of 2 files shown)


dwoz (approved):

twangboy:
Need to bring this into Neon
https://github.com/saltstack/salt/issues/53639

Issue #50238, opened 2018-10-25
KeyError: u'ret' when running salt -b x ... on salt master-of-masters.


roskens:
### Description of Issue/Question
Running `salt -b 50 '*' state.highstate` on salt MasterOfMasters and getting an exception:
```
[ERROR   ] An un-handled exception was caught by salt's global exception handler:
KeyError: u'ret'
Traceback (most recent call last):
  File "/bin/salt", line 10, in <module>
    salt_main()
  File "/usr/lib/python2.7/site-packages/salt/scripts.py", line 485, in salt_main
    client.run()
  File "/usr/lib/python2.7/site-packages/salt/cli/salt.py", line 65, in run
    self._run_batch()
  File "/usr/lib/python2.7/site-packages/salt/cli/salt.py", line 283, in _run_batch
    for res in batch.run():
  File "/usr/lib/python2.7/site-packages/salt/cli/batch.py", line 262, in run
    ret[minion] = data['ret']
KeyError: u'ret'
Traceback (most recent call last):
  File "/bin/salt", line 10, in <module>
    salt_main()
  File "/usr/lib/python2.7/site-packages/salt/scripts.py", line 485, in salt_main
    client.run()
  File "/usr/lib/python2.7/site-packages/salt/cli/salt.py", line 65, in run
    self._run_batch()
  File "/usr/lib/python2.7/site-packages/salt/cli/salt.py", line 283, in _run_batch
    for res in batch.run():
  File "/usr/lib/python2.7/site-packages/salt/cli/batch.py", line 262, in run
    ret[minion] = data['ret']
KeyError: u'ret'
```

I added a logging statement after `for minion, data in six.iteritems(parts):` in /usr/lib/python2.7/site-packages/salt/cli/batch.py to print the minion & its data value. When this exception happens, data is always `{'failed': True}`. 

The minion that it fails on is random too. 

### Setup
MasterOfMaster(1) -> Syndic(3) -> Minion (800)

Minion's are configured with 3 masters and master_type: "failover" so they only are connected to a single syndic at a time.

### Steps to Reproduce Issue
Exception occurs regardless of running either of these:

# salt -b 50 '*' state.highstate
# salt -b 50 '*' state.highstate --static

### Versions Report
(Provided by running `salt --versions-report`. Please also mention any differences in master/minion versions.)

master# salt --versions-report
Salt Version:
           Salt: 2018.3.2

Dependency Versions:
           cffi: 1.6.0
       cherrypy: Not Installed
[...]

Issue #48509, opened 2018-07-10
Running salt with batch flag causes hard crash


Cull-Methi:
### Description of Issue/Question
Running the master --> syndic --> minion topology trying to use a batch size with any salt command causes an exception which makes the master hard crash. Any command seems to trigger it, but in particular I was trying `salt '*' cmd.run 'uname -v' -b 20%` (it also failed with `-b 100`, I thought it might've been an issue with the `%`)
### Setup
Top level salt master --> Salt syndic == salt syndic master --> Salt minions

Using a pex executable built off of the 2018.3.1 tag with #48308 git cherrypicked in
`/etc/salt/master`: 
```                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                   
master_id: smn-854f3dbb_U003722

file_roots:
  base:
    - /etc/salt/file_root


auto_accept: True

order_masters: True 

log_file: /var/log/salt/master
log_level_logfile: info


transport: tcp
tcp_authentication_retries: -1
tcp_keepalive_idle: 300
tcp_keepalive_cnt: 9
tcp_keepalive_intvl: 30
return_retry_timer: 5
return_retry_timer_max: 30
recon_max: 149000
random_reauth_delay: 150

timeout: 30
cli_summary: True

open_mode: True


worker_threads: 100

```

The error produced: 

```
[ERROR   ] An un-handled exception was caught by salt's global exception handler:
KeyError: 'ret'
Traceback (most recent call last):
  File ".bootstrap/_pex/pex.py", line 367, in execute
    self._wrap_coverage(self._wrap_profiling, self._execute)
  File ".bootstrap/_pex/pex.py", line 293, in _wrap_coverage
    runner(*args)
  File ".bootstrap/_pex/pex.py", line 325, in _wrap_profiling
    runner(*args)
  File ".bootstrap/_pex/pex.py", line 410, in _execute
    return self.execute_entry(self._pex_info.entry_point)
  File ".bootstrap/_pex/pex.py", line 468, in execute_entry
    return runner(entry_point)
  File ".bootstrap/_pex/pex.py", line 486, in execute_pkg_resources
    return runner()
  File "/root/.pex/install/salt-2018.3.1_1_ge1035a6-py3-none-any.whl.1121c685594156d9569db3f436db1abad3b5bae4/salt-2018.3.1_1_ge1035a6-py3-none-any.whl/salt/unity.py", line 45, in salt_unity
[...]

PR #53159, opened 2019-05-21
batch.py: avoid exception when minion does not respond

jvanz:
### What does this PR do?

We have several issues reporting that salt is throwing exception when
the minion does not respond. This change avoid the exception adding a
default data to the minion when it fails to respond. This patch based
on the patch suggested by @roskens.

Issues #46876 #48509 #50238
bsc#1135507

### Previous Behavior

salt -b 50 -L server1.domain.com,fake.domain.com test.ping 

Fail with

KeyError: u'ret'
  File "/usr/bin/salt", line 10, in <module>
    salt_main()
  File "/usr/lib/python2.7/site-packages/salt/scripts.py", line 485, in salt_main
    client.run()
  File "/usr/lib/python2.7/site-packages/salt/cli/salt.py", line 65, in run
    self._run_batch()
  File "/usr/lib/python2.7/site-packages/salt/cli/salt.py", line 283, in _run_batch
    for res in batch.run():
  File "/usr/lib/python2.7/site-packages/salt/cli/batch.py", line 322, in run
    ret[minion] = data['ret']
KeyError: u'ret'
  File "/usr/bin/salt", line 10, in <module>
    salt_main()
  File "/usr/lib/python2.7/site-packages/salt/scripts.py", line 485, in salt_main
    client.run()
  File "/usr/lib/python2.7/site-packages/salt/cli/salt.py", line 65, in run
    self._run_batch()
  File "/usr/lib/python2.7/site-packages/salt/cli/salt.py", line 283, in _run_batch
    for res in batch.run():
  File "/usr/lib/python2.7/site-packages/salt/cli/batch.py", line 322, in run
    ret[minion] = data['ret']
KeyError: u'ret'

### New Behavior

\# salt --out=raw  -L germ109.suse.cz,fake.domain.cz test.ping 
{u'germ109.suse.cz': True}
{u'fake.domain.cz': u'Minion did not return. [Failed]'}

### Tests written?

No

### Commits signed with GPG?

No


jvanz:
I don't know why the the jenkins/pr/py3-centos-7 is failing. As far as I can see, the job is failing because a problem in the python packages installation. How can I fix that?

meaksh (approved):

salt/cli/batch.py
```diff
@@ -312,6 +312,11 @@ def run(self):
                     if self.opts.get('failhard') and data['ret']['retcode'] > 0:
                         failhard = True
 
+                # avoid an exception if the minion does not respond.
+                if data.get("failed") is True:
+                    log.debug('Minion %s failed to respond: data=%s', minion, data)
+                    data = {'ret': 'Minion did not return. [Failed]', 'retcode': 1}
```
DmitryKuzmenko: I'm ok except the hardcoded retcode value. Could you please use a constant instead?
```
'retcode': salt.defaults.exitcodes.EX_GENERIC
```

DmitryKuzmenko (changes requested):

salt/cli/batch.py
jvanz: Indeed. You're right. I forgot to search for a constant.

jvanz: I've pushed a new commit fixing this

DmitryKuzmenko (approved):

dwoz:
re-run full all

salt/cli/batch.py
```diff
@@ -312,6 +312,11 @@ def run(self):
                     if self.opts.get('failhard') and data['ret']['retcode'] > 0:
                         failhard = True
 
+                # avoid an exception if the minion does not respond.
+                if data.get("failed") is True:
+                    log.debug('Minion %s failed to respond: data=%s', minion, data)
+                    data = {'ret': 'Minion did not return. [Failed]', 'retcode': salt.defaults.exitcodes.EX_GENERIC}
```
dincamihai: I've just noticed that here: https://github.com/saltstack/salt/blob/e81e495b2713336f9908e5f4cc15ba8ebcb67306/salt/cli/batch.py#L311 the structure of 'ret' is different. Shouldn't this be unified?

dincamihai: I'm also not sure if completely replacing `data` with a new dictionary is the right thing to do here.

dincamihai (changes requested):

salt/cli/batch.py
DmitryKuzmenko: @dincamihai 

> the structure of 'ret' is different. Shouldn't this be unified?

The snipped you've mentioned is applied in the only case the executed module returns a `dict` value. In  other cases the return value is cusually `bool`. I the error case this PR handles the return value is absent and it makes it a string. So `retcode` is put in the correct place here.

> I'm also not sure if completely replacing data with a new dictionary is the right thing to do here.

This change is making batch mode work similar to non-batch. In the described case non-batch return is implemented here: https://github.com/saltstack/salt/blob/develop/salt/client/__init__.py#L1580 it just returns a dict containing `out`, `ret` and `retcode` values. Probably it'd be more correct to add `'out': 'no_return'` here too. But overall this PR looks good.

DmitryKuzmenko (approved):

salt/cli/batch.py
dincamihai: > The snipped you've mentioned is applied in the only case the executed module returns a dict value. In other cases the return value is cusually bool. I the error case this PR handles the return value is absent and it makes it a string. So retcode is put in the correct place here.

You are right, but shouldn't we align that exception and have an unified return structure? The structure should not change depending on the executed module.

jvanz (commented):
Why do we have the second pull request, #146? Sorry, I don't get it.

salt/cli/batch.py
jvanz: Guys, sorry but I'm still confused. Should I change the return dict? It is still no clear to me what was the decision. =/

bryceml:
Having an unknown repository as a source breaks jenkins.  If you would like this PR to make it in, please re-submit and link to the original.

Issue #52136, opened 2019-03-11
Import from custom _utils


kiemlicz:
### Description of Issue/Question
Is it possible to import custom utilities via python imports?
The doc https://docs.saltstack.com/en/latest/topics/utils/index.html suggests so but I couldn't get it working.

### Setup
Follow `Foo` class example from:  
https://docs.saltstack.com/en/latest/topics/utils/index.html

### Steps to Reproduce Issue
In any custom execution module, e.g.,: `_modules/customModule.py` it is not possible to write:
`import mymodule` (given I've named the `Foo` class file `mymodule.py`)  
It always fails with `No module named 'mymodule'`
Of course `__utils__['mymodule.foo']()` syntax works, the question is about import

Could you please clarify if the doc is wrong?

### Versions Report
```
Salt Version:
           Salt: 2019.2.0

Dependency Versions:
           cffi: 1.12.2
       cherrypy: unknown
       dateutil: 2.8.0
      docker-py: 1.10.6
          gitdb: 2.0.0
      gitpython: 2.1.1
          ioflo: Not Installed
         Jinja2: 2.10
        libgit2: 0.27.4
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.4.8
   mysql-python: Not Installed
      pycparser: 2.19
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: 0.27.4
         Python: 3.5.3 (default, Sep 27 2018, 17:25:39)
   python-gnupg: Not Installed
         PyYAML: 3.13
          PyZMQ: 16.0.2
           RAET: Not Installed
          smmap: 2.0.1
        timelib: Not Installed
        Tornado: 4.4.3
            ZMQ: 4.2.1

System Versions:
           dist: debian 9.8
         locale: ANSI_X3.4-1968
        machine: x86_64
        release: 4.9.0-7-amd64
         system: Linux
        version: debian 9.8
```

Ch3LL:
i thought you had to use `__utils__` because thats how we pack it in  when we load `minion_mods`, but maybe someone on the @saltstack/team-core can correct me? Looks like that doc example makes it seem like you can import utils, but i thought we always wanted to use `__utils__`?

UtahDave:
@Ch3LL There are a lot of place where we're importing from Salt utils.

```
[...]

PR #53167, opened 2019-05-22
loader: support extra modules in sys.path

aplanas:
### What does this PR do?

The LazyLoader class populate the sys.path array to add the path
that contains the Python module that will be executed.  If before
this point sys.path gets extended (like when globally the extmods
utils directory in added in some places, like after a sync) the
code can find all the dependencies.

This patch adds a new parameter to the LazyLoader class,
extra_module_dirs, that will populate locally the sys.path before
executing the the module.

This can be used to guarantee that in utils (and other modules) will be
available to the modules and states in any circumstances and Python
process.

### Tests written?

Yes

### DIFF
--- a/salt/loader.py
+++ b/salt/loader.py
@@ -1547,15 +1547,27 @@ class LazyLoader(salt.utils.lazy.LazyDict):
                 self._clean_module_dirs.append(directory)
 
     def __clean_sys_path(self):
+        invalidate_path_importer_cache = False
         for directory in self._clean_module_dirs:
             if directory in sys.path:
                 sys.path.remove(directory)
+                invalidate_path_importer_cache = True
         self._clean_module_dirs = []
 
         # Be sure that sys.path_importer_cache do not contains any
         # invalid FileFinder references
         if USE_IMPORTLIB:
             importlib.invalidate_caches()
+
+        # Because we are mangling with importlib, we can find from
+        # time to time an invalidation issue with
+        # sys.path_importer_cache, that requires the removal of
+        # FileFinder that remain None for the extra_module_dirs
+        if invalidate_path_importer_cache:
+            for directory in self.extra_module_dirs:
+                if directory in sys.path_importer_cache \
+                   and sys.path_importer_cache[directory] is None:
+                    del sys.path_importer_cache[directory]
 
     def _load_module(self, name):
         mod = None
--- a/salt/loader.py
+++ b/salt/loader.py
@@ -260,6 +260,7 @@ def minion_mods(
         whitelist=whitelist,
         loaded_base_name=loaded_base_name,
         static_modules=static_modules,
+        extra_module_dirs=utils.module_dirs if utils else None,
     )
 
     ret.pack["__salt__"] = ret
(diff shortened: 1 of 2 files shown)


waynew:
There's a test failure happening https://jenkinsci.saltstack.com/job/pr-kitchen-windows2016-py3/job/PR-53167/3/testReport/junit/integration.states.test_file/FileTest/test_file_managed_onchanges/ - I'm not sure if that's related to this PR or not :suspect: 

Also looks like there's a merge conflict 😛 

Otherwise, I think this is looking good.

max-arnold:
How this PR relates to #52001?  Does it extend or replace it?

aplanas:
@max-arnold I think that is a replacement of this one. Maybe is a good idea to add a commit here to revert the other code.

aplanas:
@max-arnold I added the revert commit here

@waynew I rebased the code. Also I do not see the windows fail related with the code. Lets see if is still present after the rebase.

waynew (approved):
I think this is ready to go, assuming tests pass after the rebase 😛

max-arnold:
@aplanas Would this fix #52136 and #53666?

aplanas:
@max-arnold Not sure about salt-ssh (#53666), but this fix #52136. I will not be able to test #53666 for some time (out from the office for a while)

max-arnold:
@aplanas I'm not very familiar with the loader internals, but it seems like other module types also import `salt.utils` and could benefit from this improvement.

In other words, if there is a bug in `utils` and someone shipped a fix via `saltutil.sync_utils`, all module types should be able to pick it up.

Below is a list of functions in `loader.py`, which you might want to change too (I'm not 100% sure though):

```
auth
beacons
cache
call
eauth_tokens
executors
log_handlers
matchers
metaproxy
netapi
outputters
pkgdb
pkgfiles
queues
render
returners
serializers
ssh_wrapper
thorium
tops
wheels
```

What do you think about this idea?

aplanas:
@max-arnold you can be right.

I updated the LazyLoader constructors where there is an explicit call to make `__util__` available, but I left the others untouched.

For example, in `def auth`, I cannot see that is the case:

```python
return LazyLoader(
        _module_dirs(opts, 'auth'),
        opts,
        tag='auth',
        whitelist=whitelist,
        pack={'__salt__': minion_mods(opts)},
    )
```

I would expect that only `__salt__` is available.

Can be that there is a different mechanism where `__util__` becomes available in those cases, if so I would need a bigger pull request.

Can we do that later?

max-arnold:
@aplanas 

> Can we do that later?

We can definitely do this later, and it looks like there will be enough time (a few months?) before the `neon` release. I'm just a community member and this was a suggestion, not a requirement.

aplanas:
Rebased on top of master.

Note that I dropped the commit that reverted #52001, as this is not in master. I did not backported the original #52001 commit either.

waynew:
@aplanas looks like there's a conflict. I think these test failures are resolved on the most up-to-date `master`.

aplanas:
@waynew thanks, I rebased the code.

aplanas:
windows errors are related to utils.pycrypto: AttributeError: 'module' object has no attribute 'crypt'

waynew (dismissed):
I think this looks good, but we have some issues going on with our Windows test suite that we need to figure out. As you point out, I don't think they're related to the changes in this PR.

waynew:
We had some timeouts happening, something going wonky with the infrastructure(?)

I've restarted the Windows tests - assuming they pass this should be good to merge in 👍

waynew:
Hm. I'm a bit confused. I merged the latest master into this PR and yet there's still an issue with the crypt thing? I don't have time to look at this today but I'll stick it on my list.

aplanas:
I rebased, let me check the results.

aplanas:
Yep. Ignore the lint, is not related with my code:

************* Module salt.returners.django_return
salt/returners/django_return.py:42: [E0611(no-name-in-module), ] No name 'dispatch' in module 'django'

aplanas:
no failing test, re-review?

dwoz (approved):

PR #53188, opened 2019-05-22
Update file.chattr

waynew:
### What does this PR do?

Fixes issues with chattr that cause failing tests on RedHat platforms

#52079 (partially, still need to re-install e2fsprogs on the suse boxes)

### Previous Behavior

Before, `check_perms` removed all the file attrs and then added them back, before setting the requested permissions. On certain platforms (RedHat), adding the extent formatting flat (e) is impossible. Removing all flags and then adding them back caused a failure here.

By removing then re-adding the flags, we were also steamrolling over any safety bits someone may have set on their files (e.g. append(a) or immutable(i)).

While this version doesn't actually have the `e` flag, it *does* have the flag removal/readdition behavior.

### New Behavior

Now attributes are only added or removed as requested. If someone is setting attributes on their files, we won't ignore them any more.

### Tests written?

Yes - and many more to cover this section of code.

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/modules/file.py
+++ b/salt/modules/file.py
@@ -29,7 +29,7 @@ import time
 import glob
 import hashlib
 import mmap
-from collections import Iterable, Mapping
+from collections import Iterable, Mapping, namedtuple
 from functools import reduce  # pylint: disable=redefined-builtin
 
 # pylint: disable=import-error,no-name-in-module,redefined-builtin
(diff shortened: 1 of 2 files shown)


waynew:
@dwoz I know you made some of the code changes in here on the develop branch, so you may be familiar with this area of the code base.

twangboy (approved):

twangboy:
@waynew Looks like you got some lint

s0undt3ch (approved):
Let me now what system packages need to go in and into which OS'es so I can add those to salt-jenkins and thus the golden images.

waynew:
@twangboy doh! I knew I was forgetting to check *something* :stuck_out_tongue_closed_eyes:

Issue #51903, opened 2019-02-28
[Fluorine] State file.directory recurse: When both file_mode and dir_mode are supplied, file_mode overrides dir_mode for the directory itself


ssiuhk:
### Description of Issue/Question
Hi, I have an issue with file.directory state, when both file_mode and dir_mode are supplied, file_mode overrides dir_mode for the directory itself

According to [Saltstack Document](https://docs.saltstack.com/en/latest/ref/states/all/salt.states.file.html#salt.states.file.directory), file_mode for permission of file, dir_mode for permission of directories.

### Setup
```
[root@localhost salttest]# pwd
/tmp/salttest
[root@localhost salttest]# tree -up .
.
├── [drwxr-xr-x root    ]  dir_one
│   └── [drwxr-xr-x root    ]  dir_two
│       └── [-rw-r--r-- root    ]  testfile.txt
├── [-rw-r--r-- root    ]  file_perm.sls
└── [-rw-r--r-- root    ]  top.sls

2 directories, 3 files

[root@localhost salttest]# cat file_perm.sls
/tmp/salttest/dir_one:
  file.directory:
    - user: root
    - group: root
    - dir_mode: 755
    - file_mode: 644
    - recurse:
      - user
      - group
      - mode

/tmp/salttest/dir_one/dir_two:
  file.directory:
    - user: root
    - group: root
    - dir_mode: 755
    - makedirs: True

```

### Steps to Reproduce Issue
```
[root@localhost salttest]# salt-call --local --file-root=/tmp/salttest/ state.apply test=True file_perm
local:
----------
          ID: /tmp/salttest/dir_one
    Function: file.directory
      Result: None
     Comment: The following files will be changed:
              /tmp/salttest/dir_one: mode - 0644
     Started: 14:41:07.828673
    Duration: 6.19 ms
     Changes:
----------
          ID: /tmp/salttest/dir_one/dir_two
    Function: file.directory
      Result: True
     Comment: The directory /tmp/salttest/dir_one/dir_two is in the correct state
     Started: 14:41:07.835033
    Duration: 0.738 ms
     Changes:

Summary for local
------------
Succeeded: 2 (unchanged=1)
Failed:    0
------------
Total states run:     2
Total run time:   6.928 ms

```

### Versions Report
```
[root@localhost salttest]# salt-call --versions-report
Salt Version:
           Salt: 2019.2.0

Dependency Versions:
           cffi: Not Installed
       cherrypy: Not Installed
[...]

PR #52024, opened 2019-03-07
State always reports changed when files and dirs exist at the same level

danfoster:
Fixes #51903

### What does this PR do?

Bugfix: When using the `file.directory` state with `resurse=[mode]` and you have directories and files at the same level, the state always reports as changed, even if no changes need to be made

### What issues does this PR fix or reference?

#51903

### Previous Behavior

```
$ umask 007
$ for i in `seq 1 3`; do mkdir -p /tmp/test/d$i; touch /tmp/test/f$i; done        
ls -l /tmp/test
total 0
drwx------ 2 root root 6 Mar  7 13:16 d1
drwx------ 2 root root 6 Mar  7 13:16 d2
drwx------ 2 root root 6 Mar  7 13:16 d3
-rw------- 1 root root 0 Mar  7 13:16 f1
-rw------- 1 root root 0 Mar  7 13:16 f2
-rw------- 1 root root 0 Mar  7 13:16 f3
$ salt-call state.single file.directory name="/tmp/test" dir_mode="700" file_mode="600" recurse="[mode]" test=false
local:
----------
          ID: /tmp/test
    Function: file.directory
      Result: True
     Comment: Directory /tmp/test is in the correct state
     Started: 13:54:50.775917
    Duration: 8.027 ms
     Changes:
              ----------
              /tmp/test/d1:
                  ----------
                  mode:
                      0600
              /tmp/test/d2:
                  ----------
                  mode:
                      0600
              /tmp/test/d3:
                  ----------
                  mode:
                      0600

Summary for local
------------
Succeeded: 1 (changed=1)
Failed:    0
------------
Total states run:     1
Total run time:   8.027 ms
```

### New Behavior

```
$ salt-call state.single file.directory name="/tmp/test" dir_mode="700" file_mode="600" recurse="[mode]" test=false
local:

Summary for local
------------
Succeeded: 1
Failed:    0
------------
Total states run:     1
Total run time:   8.128 ms
$ chmod 644 /tmp/test/f2
$ chmod 755 /tmp/test/d2
$ salt-call state.single file.directory name="/tmp/test" dir_mode="700" file_mode="600" recurse="[mode]" test=false
local:
----------
          ID: /tmp/test
    Function: file.directory
      Result: True
     Comment: Directory /tmp/test updated
     Started: 14:16:34.005604
    Duration: 11.362 ms
     Changes:
              ----------
              mode:
                  0700

Summary for local
------------
Succeeded: 1 (changed=1)
Failed:    0
------------
[...]

PR #53200, opened 2019-05-23
Fix file.directory mode behaviour in def _check_directory

hktb:
### What does this PR do?
This PR definitively fixes the behaviour of the file.directory module so it reflects the expected (documented) behaviour.

It therefore fixes issues #51903 and #52024 aswell.

I have published a [gist](https://gist.github.com/hktb/18ab8a1781fff799de19ed4e876abe69) which details all aspects of this fix.

### Commits signed with GPG?

No

### DIFF
--- a/salt/states/file.py
+++ b/salt/states/file.py
@@ -694,13 +694,14 @@ def _check_directory(name,
                      user=None,
                      group=None,
                      recurse=False,
-                     mode=None,
+                     dir_mode=None,
                      file_mode=None,
                      clean=False,
                      require=False,
                      exclude_pat=None,
                      max_depth=None,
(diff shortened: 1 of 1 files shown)


hktb:
It seems the unit tests assume a static comment on applying or testing the state. Since I changed the comment to better reflect what will be changed this breaks these tests for all Linux distros. It does not affect the Windows tests since this fix is only related to a Linux mechanism.
I will do 1 more commit, leaving the comment unchanged to accommodate the unit tests.

twangboy (approved):

dwoz:
Closing due to lack of activity. Anyone should feel free to re-open or create a new PR if they want to see this through to the end.

PR #53225, opened 2019-05-23
Sync modules by default

cmclaughlin:
I recently ran into an issue where my minion cache was cleared and I
ran `state.apply xxx` only to eventually find modules aren't sync'd.
My `xxx` state called a custom module and it failed with vague jinja2
error.

I assumed modules would be sync'd by that command, since they are sync'd by
`state.apply` with no mod arg and `state.highstate` (which, of course,
are equivalent). I think that's a reasonable assumption, so here I'm
proposing that all modules should be sync'd by default.

In my case here's where this is called:

https://github.com/saltstack/salt/blob/develop/salt/modules/state.py#L778

### What does this PR do?

See above.

N/A

### Previous Behavior

See above.

### New Behavior

See above.

### Tests written?

No.... hopefully already covered?

### Commits signed with GPG?

No

### DIFF
--- a/salt/modules/state.py
+++ b/salt/modules/state.py
@@ -1098,7 +1098,7 @@ def highstate(test=None, queue=False, **kwargs):
     return ret
 
 
-def sls(mods, test=None, exclude=None, queue=False, sync_mods=None, **kwargs):
+def sls(mods, test=None, exclude=None, queue=False, sync_mods=True, **kwargs):
     '''
     Execute the states in one or more SLS files
 


max-arnold:
Sync operation is not “free” to do it each time. This would decrease the state.apply performance for everyone.

cmclaughlin:
Good point, but using that logic highstate shouldn't auto-sync either.  Why sync in one place but not the other?  At my company we run highstate 95% of the time. We typically only run state.apply when doing development - and that's when we're more likely to change custom modules and need them to sync. Sure, we can accept this as how it is... but I opened this PR because I don't find this intuitive.

sagetherage:
@cmclaughlin we are no longer accepting changes to the develop branch, please point this to the master branch

sagetherage:
@cmclaughlin we are no longer accepting changes to the develop branch, please point this to the master branch

sagetherage:

PR #53229, opened 2019-05-23
fixes integration.states.test_cmd on win

cmcmarrow:
### What does this PR do?
fixes integration.states.test_cmd on win

https://github.com/saltstack/salt/issues/53218

### Previous Behavior
integration.states.test_cmd.CMDTest.test_run_hide_output would fail on win

### Tests written?

No... it is a test

### Commits signed with GPG?
Yes

### DIFF
--- a/tests/integration/states/test_cmd.py
+++ b/tests/integration/states/test_cmd.py
@@ -46,9 +46,11 @@ class CMDTest(ModuleCase, SaltReturnAssertsMixin):
         '''
         cmd.run with output hidden
         '''
+
+        cmd = u'dir' if IS_WINDOWS else u'ls'
         ret = self.run_state(
             u'cmd.run',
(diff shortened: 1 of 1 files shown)


tests/integration/states/test_cmd.py
```diff
@@ -46,9 +46,11 @@ def test_run_hide_output(self):
         '''
         cmd.run with output hidden
         '''
+
+        cmd = u'dir' if IS_WINDOWS else u'ls'
```
twangboy: Since we have the following import at the top:
```
```
you don't have to add the `u`. 
`unicode_literals` makes all strings unicode strings.

twangboy: I know you're just following convention here.

twangboy (approved):

PR #53234, opened 2019-05-24
[WIP] Bugfix: Handle multibyte characters spanning blocks of received data

LuAPi:
### What does this PR do?

This fix only applies to unicode since I don't know what the specific exceptions raised by other encoding schemes are. If similar errors occur with other multibyte characters in other encoding schemes then the read_and_decode_fd function can be extended to handle those cases too.

Probably this one: https://github.com/saltstack/salt/issues/48473

### Previous Behaviour
This process would fail when a multibyte character occurs at the end of a received block of data with only part of the character in the current block. The decoding would fail with the exception UnicodeDecodeError and the message 'unexpected end of data' and include the position of failure as the end of the block of data.

### New Behaviour
This fix checks for that specific failure case and will then read another block of data and append it to the current block of data so that the unicode character is no longer split across received data blocks. It then attempts to perform decoding on the combined data block. A loop is used to handle the case where the same problem occurred with the new combined block.

The intent of this change is to only handle this specific error and to raise all other exceptions as before. The code is only intended to remain in the while True loop so long as this specific failure case occurs. This should only loop forever if the program it is receiving data from outputs an infinite stream of valid characters that consistently has multibyte characters split across the block boundaries.

A better fix might be to switch to reading individual bytes rather than whole blocks to attempt to complete only the character that failed. That would avoid the potential infinite loop in the case of a program that runs forever whose output happens to always have a multibyte character split across blocks. That said, the chances of that happening are probably fairly low and you might have other problems if you are receiving data from something that generates infinite output.

Any invalid characters that don't occur at the end of the block will not have the reason 'unexpected end of data' so those exceptions will be raised and will not cause code to loop.

### Tests written?

No

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/utils/vt.py
+++ b/salt/utils/vt.py
@@ -32,7 +32,6 @@ import logging
 
 # Import salt libs
 from salt.ext import six
-from salt.ext.six.moves import range
 
 mswindows = (sys.platform == "win32")
 
--- a/salt/utils/vt.py
+++ b/salt/utils/vt.py
@@ -644,17 +644,25 @@ class Terminal(object):
                 bytes_read = os.read(fd, maxsize)
                 # Loop to handle the case where a multi-byte unicode
                 # character is split across blocks of received bytes
-                while True:
+                # This max length is for unicode, it may need to change if
+                # this function is updated to handle errors in other encodings
+                max_multibyte_char_len = 4
+                for additional_reads in range(max_multibyte_char_len):
                     try:
                         return self._translate_newlines(
                             salt.utils.stringutils.to_unicode(bytes_read)
                         )
                     except UnicodeDecodeError as ex:
+                        if additional_reads == (max_multibyte_char_len - 1):
+                            # We have read enough extra bytes to complete the
+                            # character, give up now
+                            raise
                         if ex.reason == 'unexpected end of data':
-                            new_bytes_read = os.read(fd, maxsize)
+                            # Read an extra byte to try and complete the split
+                            # character
+                            new_bytes_read = os.read(fd, 1)
                             if new_bytes_read == b'':
                                 # End of stream is an incomplete character
-                                # Raise exception to avoid infinite loop
                                 raise
                             bytes_read += new_bytes_read
                         else:
--- a/salt/utils/vt.py
+++ b/salt/utils/vt.py
@@ -658,7 +658,9 @@ class Terminal(object):
             # ----- Helper function for processing STDERR and STDOUT -------->
             def read_and_decode_fd(fd, maxsize, partial_data_attr=None):
                 bytes_read = getattr(self, partial_data_attr, b'')
-                bytes_read += os.read(fd, maxsize)
+                # Only read one byte if we already have some existing data
+                # to try and complete a split multibyte character
+                bytes_read += os.read(fd, 1 if len(bytes_read) else maxsize)
                 try:
                     decoded_data =  self._translate_newlines(
                         salt.utils.stringutils.to_unicode(bytes_read)
--- a/salt/utils/vt.py
+++ b/salt/utils/vt.py
@@ -120,6 +120,9 @@ class Terminal(object):
                  # sys.stdXYZ streaming options
                  stream_stdout=None,
                  stream_stderr=None,
+
+                 # Used for tests
+                 force_receive_encoding=__salt_system_encoding__,
                  ):
 
         # Let's avoid Zombies!!!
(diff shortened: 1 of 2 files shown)


LuAPi:
Having thought about this a bit more I think should change the part where it reads more data on failure to only read one byte at a time to attempt to only complete the character that was split as well as set a limit on the maximum number of extra reads.

s0undt3ch (changes requested):
Could you please add tests for this change?
In particular, I'm worried how it handles streams of different encodings

LuAPi:
I've not had time to properly explore how the integration tests are setup in this repo.

I believe the proper tests would be a salt state that runs some commands that generate appropriate output on stdout or stderr and checks that running the state succeeds.

As for how it handles different encodings it's intended to match the current behavior of not handling this particular failure mode. If I added tests that can trigger this failure mode for other encodings then I expect the salt state to fail both with and without this change.

Is there a list of supported encodings for Salt? I'm not aware of one.

I was looking to fix the problem for the most common encoding and leave a good place to handle other encodings should anyone else ever encounter this problem rather than handle all possible encodings.

LuAPi:
My unit test revealed I'd made an incorrect assumption about what os.read would do since I hadn't seen that previous code had changed the file descriptor into non-blocking mode.
I changed how it so that it stores the data to decode and attempts to complete the split character next time _recv is called. I did that rather than try and determine if the reason os.read returned empty bytes (either EOF or no data available right now) in the helper function because doing that would duplicate much of what _recv does at the start so it kept the code simpler this way.

I added a unit test for a non-unicode encoding (shift-jis) which failed as expected so I changed the method of detecting split multibyte characters to something less accurate but which handles more than one encoding and for which I'm currently not aware of any serious problems it may cause.

LuAPi:
@s0undt3ch 
> Could you please add tests for this change?
> In particular, I'm worried how it handles streams of different encodings

So I added some tests and they work on my machine, but I can't help but notice that all the other tests in test_vt are disabled.

The windows builds are failing because the location the windows builds gets their openssl installer from became a 404 yesterday.
`18:28:28           ID: openssl
18:28:28            Function: module.run
18:28:28         Name: winrepo_pkg.install
18:28:28              Result: False
18:28:28             Comment: Module function winrepo_pkg.install threw an exception. Exception: Error: HTTP 404: Not Found reading https://slproweb.com/download/Win64OpenSSL-1_0_2r.exe`

The python 3 builds are failing because the test environment is weird and sys.stdout.fileno() doesn't work so the VT doesn't do what it's supposed to and seems to be returning some of the stdout of the unit test instead of the stdout of the program it's supposed to be running which doesn't match the blocks of characters expected.
`18:34:55        -        -        Traceback (most recent call last):
18:34:55        -        -          File "/tmp/kitchen/testing/salt/utils/vt.py", line 219, in __init__
18:34:55        -        -            self._spawn()
18:34:55        -        -          File "/tmp/kitchen/testing/salt/utils/vt.py", line 440, in _spawn
18:34:55        -        -            self.stdout = sys.stdout.fileno()
18:34:55        -        -        io.UnsupportedOperation: fileno
18:34:55        -        -
18:34:55        -        -        During handling of the above exception, another exception occurred:
18:34:55        -        -
18:34:55        -        -        Traceback (most recent call last):
18:34:55        -        -          File "/tmp/kitchen/testing/tests/unit/utils/test_vt.py", line 305, in test_split_multibyte_characters_shiftjis
18:34:55        -        -            force_receive_encoding=encoding)
18:34:55        -        -          File "/tmp/kitchen/testing/salt/utils/vt.py", line 228, in __init__
18:34:55        -        -            'Failed to spawn the VT. Error: {0}'.format(err)
18:34:55        -        -        salt.utils.vt.TerminalException: Failed to spawn the VT. Error: fileno`

The python 2 tests are failing for the same reason (I think) as the python 3 tests, but you get a less informative error message
`18:33:27        -        -        Traceback (most recent call last):
18:33:27        -        -          File "/tmp/kitchen/testing/tests/unit/utils/test_vt.py", line 305, in test_split_multibyte_characters_shiftjis
18:33:27        -        -            force_receive_encoding=encoding)
18:33:27        -        -          File "/tmp/kitchen/testing/salt/utils/vt.py", line 228, in __init__
18:33:27        -        -            'Failed to spawn the VT. Error: {0}'.format(err)
18:33:27        -        -        TerminalException: Failed to spawn the VT. Error: fileno`

LuAPi:
re-run full
Jenkins wasn't working properly the last ~6 days but looks like it might be back up now.

LuAPi:
I think this is good now, but I'll hold off removing the [WIP] until the windows openssl installer state ( https://github.com/saltstack/salt/issues/53312 ) is fixed so the windows tests can run and I've seen that they succeed.

LuAPi:
I just noticed I'm merging into the wrong branch because I counted two point releases back not two main releases back. I've rebased and opened https://github.com/saltstack/salt/pull/53404 instead.

PR #53235, opened 2019-05-24
Revert "Update msgpack calls for newer msgpack"

s0undt3ch:
Reverts saltstack/salt#52934


s0undt3ch:
Merging as this broke the whole 2019.2.1 builds.

PR #53271, opened 2019-05-28
Backport #53142 to 2019.2.1

s0undt3ch:
Backport #53142 to 2019.2.1

> Use six.reraise for the greater good

> In Python2, `raise exc` obliterates the stack trace. That's no good.

> Fortunately, six has a useful syntax that works across Python versions.
> It does have the downside of adding `reraise` to the stacktrace, but
> otherwise, this should be better.

Conflicts:
* salt/transport/tcp.py
* salt/utils/cloud.py
* salt/utils/vmware.py
* salt/utils/yamlloader.py
* tests/support/helpers.py

### DIFF
--- a/salt/minion.py
+++ b/salt/minion.py
@@ -719,7 +719,7 @@ class MinionBase(object):
                     if attempts == tries:
                         # Exhausted all attempts. Return exception.
                         self.connected = False
-                        raise exc
+                        six.reraise(*sys.exc_info())
 
     def _discover_masters(self):
         '''
--- a/salt/scripts.py
+++ b/salt/scripts.py
@@ -40,7 +40,7 @@ def _handle_interrupt(exc, original_exc, hardfail=False, trace=''):
             log.error(trace)
         raise original_exc
     else:
-        raise exc
+        six.reraise(*sys.exc_info())
 
 
 def _handle_signals(client, signum, sigframe):
--- a/salt/utils/asynchronous.py
+++ b/salt/utils/asynchronous.py
@@ -5,9 +5,12 @@ Helpers/utils for working with tornado asynchronous stuff
 
 from __future__ import absolute_import, print_function, unicode_literals
 
+import sys
+
 import tornado.ioloop
 import tornado.concurrent
 import contextlib
+from salt.ext import six
 from salt.utils import zeromq
 
 
@@ -53,7 +56,7 @@ class SyncWrapper(object):
             return object.__getattribute__(self, key)
         except AttributeError as ex:
             if key == 'asynchronous':
-                raise ex
+                six.reraise(*sys.exc_info())
         attr = getattr(self.asynchronous, key)
         if hasattr(attr, '__call__'):
             def wrap(*args, **kwargs):
--- a/salt/utils/json.py
+++ b/salt/utils/json.py
@@ -8,6 +8,7 @@ from __future__ import absolute_import, unicode_literals
 # Import Python libs
 import json  # future lint: blacklisted-module
 import logging
+import sys
 
 # Import Salt libs
 import salt.utils.data
@@ -95,7 +96,7 @@ def loads(s, **kwargs):
         if six.PY3 and isinstance(s, bytes):
             return json_module.loads(salt.utils.stringutils.to_unicode(s), **kwargs)
         else:
-            raise exc
+            six.reraise(*sys.exc_info())
 
 
 def dump(obj, fp, **kwargs):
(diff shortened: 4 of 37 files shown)


DmitryKuzmenko:
Looks good for me, just have a couple of questions/comments.
1. There are still a number of places in 2019.2.1 containing the pattern "except .. as ex: raise ex" that is completely removed from 2018.3 by the original commit. Is it OK?
2. This is true for the original PR too: in most of places we're not using the exception variable anymore. So maybe it's better to replace `except ... as ex` with `except ...`?

s0undt3ch:
After you comment @DmitryKuzmenko I went and look for additional occurrences which are now fixed.
As for `except Exception as blah`, if we use `blah` we should keep it, when not using it, we should remove it. See any place where it should be removed?

waynew:
IIRC the linter catches the unused `as blah:`... but I could be wrong!

salt/scripts.py
```diff
@@ -40,7 +40,7 @@ def _handle_interrupt(exc, original_exc, hardfail=False, trace=''):
             log.error(trace)
         raise original_exc
     else:
-        raise exc
+        six.reraise(*sys.exc_info())
```
DmitryKuzmenko: Unused `exc`.

salt/utils/asynchronous.py
```diff
@@ -53,7 +56,7 @@ def __getattribute__(self, key):
             return object.__getattribute__(self, key)
         except AttributeError as ex:
             if key == 'asynchronous':
-                raise ex
+                six.reraise(*sys.exc_info())
```
DmitryKuzmenko: Unused `ex`.

salt/utils/json.py
```diff
@@ -95,7 +96,7 @@ def loads(s, **kwargs):
         if six.PY3 and isinstance(s, bytes):
             return json_module.loads(salt.utils.stringutils.to_unicode(s), **kwargs)
         else:
-            raise exc
+            six.reraise(*sys.exc_info())
```
DmitryKuzmenko: Unused `exc`.

tests/integration/states/test_ssh_known_hosts.py
```diff
@@ -54,8 +59,7 @@ def test_present(self):
                 )
                 self.skipTest('Unable to receive remote host key')
             except AssertionError:
-                # raise initial assertion error
-                raise err
+                six.reraise(*sys.exc_info())
```
DmitryKuzmenko: The original code raises the first exception that is the `err`. The new code reraises the last (nested) AssertionError. Is is expected?

tests/integration/states/test_ssh_known_hosts.py
```diff
@@ -82,7 +86,7 @@ def test_present(self):
                         )
                 self.skipTest('Unable to receive remote host key')
             except AssertionError:
-                raise err
+                six.reraise(*sys.exc_info())
```
DmitryKuzmenko: The same as above.

tests/runtests.py
```diff
@@ -56,7 +57,7 @@
     print('Current sys.path:')
     import pprint
     pprint.pprint(sys.path)
-    raise exc
+    six.reraise(*sys.exc_info())
```
DmitryKuzmenko: Unused `exc`.

tests/support/helpers.py
```diff
@@ -209,7 +209,7 @@ def wrap(cls):
                 return caller(cls)
             except Exception as exc:
                 if attempt >= attempts -1:
-                    raise exc
+                    six.reraise(*sys.exc_info())
```
DmitryKuzmenko: Unused `exc`.

DmitryKuzmenko (changes requested):
Here are all places we got unused variable after this change.

DmitryKuzmenko:
@waynew Just checked in my dev environmant: lint says it's OK to have unused `as exc` variables.. ¯\_(ツ)_/¯

salt/minion.py
```diff
@@ -719,7 +719,7 @@ def eval_master(self,
                     if attempts == tries:
```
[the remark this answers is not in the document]
s0undt3ch: Err, we disabled that pylint rule.
@DmitryKuzmenko I'll address that in a different PR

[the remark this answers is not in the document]
DmitryKuzmenko: Sure! I'm just pointing things I've found. No problem.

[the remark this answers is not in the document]
s0undt3ch: No, no, you did the right thing, Thanks!

tests/integration/states/test_ssh_known_hosts.py
DmitryKuzmenko: @s0undt3ch this one still need to be fixed here. It's minor, but it directly relates.

salt/scripts.py
s0undt3ch: This one needs to be re-thought

Issue #53137, opened 2019-05-20
update doc pr's to be built using python3 and sphinx 2.0.1


bryceml:
### Description of Issue/Question
Doc PR's are tested using python2, I'm not sure of the version of sphinx, however, the actual docs that go on docs.saltstack.com are built using python3 and sphinx 2.0.1.

We need to move the pr builds to run on python3

The means update
* .ci/docs
* requirements/static/docs.in
* update things in sre-jenkins

verify it works.

s0undt3ch:
2018.3, 2019.2.1, neon and develop have migrated to building docs on Py3.
2019.2 will get that when the merge-forward is merged.
2017.7.9 and 2017.7 need extra work to switch.

bryceml:
This is done.

PR #53272, opened 2019-05-28
[2018.3] Switch docs building under Py3

s0undt3ch:
### What does this PR do?
See title.

#53137


.ci/docs
```diff
@@ -27,8 +27,8 @@ pipeline {
                 sh '''
                 eval "$(pyenv init -)"
                 pyenv --version
-                pyenv install --skip-existing 2.7.15
-                pyenv shell 2.7.15
+                pyenv install --skip-existing 3.6.8
```
bryceml: Is there a way to pin to latest 3.6 ?  would that be better or worse than pinning to a specific version?

s0undt3ch: latest would be a moving target(ok, it wouldn't move that much).
I guess a static, that we can update when needed is better.

bryceml (approved):

Akm0d (approved):

PR #53278, opened 2019-05-28
Update signing packages to allow for Redhat 8's use of gpg-agent

### LINKED ISSUE #57657: Update signing packages to allow for Redhat 8's use of gpg-agent
Need a way to allow for Redhat 8's use of gpg-agent when signing packages and creating a repo

dmurphy18:
### What does this PR do?
Allows for Redhat 8's use of gpg-agent when signing packages and creating a repo

https://github.com/saltstack/salt-pack-py3/issues/120
closes https://github.com/saltstack/salt-pack-py3/issues/234

### New Behavior
This allows for signing and creating rpm packages on Redhat 8, which now makes use of gpg-agent and gpg-preset-passphrase to sign packages.   Methods used previously in Redhat 6 & 7 no longer apply for REdhat 8.

### Tests written?
No - nightly packaging exercises this new functionality

### Commits signed with GPG?
No


Ch3LL (changes requested):
Can you add some test coverage here? I know you stated there is no need because nightly builds will catch any issues, but we want to catch issues before we even merge a PR, which writing tests will allow us to do so.

dmurphy18:
@ch3ll 
> Can you add some test coverage here? I know you stated there is no need because nightly builds will catch any issues, but we want to catch issues before we even merge a PR, which writing tests will allow us to do so.

I would like too, but have to fake out I/O to a file, signing it from information from gpg-agent (faking out whatever it needs from that) and then the whole signing cryptographic process. I am not sure that is possible from our testing framework given the number of interactions between various active processes and agents.  Need to read up on pytest and nox as to see if this is possible.

Ch3LL:
sounds like an integration test is out of the picture but you should still be able to mock it out and write a unit test to cover the bug fix here.

Issue #53285, opened 2019-05-28
[2019.2.1] test_kubernetes lint failures


Ch3LL:
There are a couple lint failures on 2019.2.1

```
12:29:38 tests/unit/modules/test_kubernetesmod.py:80: [E1120(no-value-for-parameter), KubernetesTestCase.test_deployments] No value for argument 'namespace' in method call
12:29:38 tests/unit/modules/test_kubernetesmod.py:94: [E1120(no-value-for-parameter), KubernetesTestCase.test_services] No value for argument 'namespace' in method call
12:29:38 tests/unit/modules/test_kubernetesmod.py:108: [E1120(no-value-for-parameter), KubernetesTestCase.test_pods] No value for argument 'namespace' in method call
12:29:38 tests/unit/modules/test_kubernetesmod.py:125: [E1120(no-value-for-parameter), KubernetesTestCase.test_delete_deployments] No value for argument 'name' in method call
12:29:38 tests/unit/modules/test_kubernetesmod.py:125: [E1120(no-value-for-parameter), KubernetesTestCase.test_delete_deployments] No value for argument 'namespace' in method call
12:29:38 tests/unit/modules/test_kubernetesmod.py:125: [E1120(no-value-for-parameter), KubernetesTestCase.test_delete_deployments] No value for argument 'body' in method call
12:29:38 tests/unit/modules/test_kubernetesmod.py:141: [E1120(no-value-for-parameter), KubernetesTestCase.test_create_deployments] No value for argument 'namespace' in method call
12:29:38 tests/unit/modules/test_kubernetesmod.py:141: [E1120(no-value-for-parameter), KubernetesTestCase.test_create_deployments] No value for argument 'body' in method call
```

https://jenkinsci.saltstack.com/job/2019.2.1/view/Lint/job/salt-2019.2.1-lint/68/console

PR #53289, opened 2019-05-28
stops pylint E1120 from showing

cmcmarrow:
### What does this PR do?
silences pylint E1120(no-value-for-parameter)

I don't believe there is a good way to fix this pylint warring. pylint is complaining about a method in a 3rd party library. The 3rd party library did nothing wrong. We add it default arguments to methods in the 3rd party library. 

https://github.com/saltstack/salt/issues/53285

### Tests written?

No

### Commits signed with GPG?

Yes


Akm0d (approved):

tests/unit/modules/test_kubernetesmod.py
```diff
@@ -23,6 +23,8 @@
 import salt.utils.platform
 from salt.modules import kubernetesmod as kubernetes
 
+# pylint: disable=E1120
```
twangboy: You need to enable this after the part that is failing.
```
# pylint: disable=E1120
some code
# pylint: enable=E1120
```

twangboy (changes requested):

garethgreenaway (approved):

twangboy (approved):

Akm0d (approved):

PR #53309, opened 2019-05-29
[WIP] Unit.modules.test win file fix2

cmcmarrow:
### What does this PR do?
Unit.modules.test win file fix2

This is just a test to see if kitchen has the right permissions  

### Tests written?
No

### Commits signed with GPG?
Yes


salt/modules/win_file.py
```diff
@@ -1154,7 +1154,7 @@ def symlink(src, link):
         win32security.TOKEN_ADJUST_PRIVILEGES
     )
     th = win32security.OpenProcessToken(win32api.GetCurrentProcess(),
-                                        desired_access)
+                                        win32security.TOKEN_ALL_ACCESS) # test
```
rares-pop: This doesn't feel right. With this approach you seem not to need desired_access anymore (so why don't you delete that?).

But is this just for fixing a test? Is the issue that a test isn't run as an Administrator?

cmcmarrow: Its just a test to see if kitchen is missing a permission.

cmcmarrow: I say it in the PR.

rares-pop: I'm not sure I follow. 

If it's a local test on your machine, why do you commit this? If it's the test that's broken, why fixing the code and not the test?

Again, if you do go with this route, why don't you remove the 'desired_access', since won't be used anymore.

twangboy: @rares-pop The test we're trying to fix here only fails when run by the test suite, so we're doing it here to observe the failure. There's something different about the environment that we're trying to figure out.

Issue #51741, opened 2019-02-21
encoding is deprecated


Ofloo:
### Description of Issue/Question

[WARNING ] /usr/local/lib/python2.7/site-packages/salt/payload.py:145: DeprecationWarning: encoding is deprecated, Use raw=False instead.
  ret = msgpack.loads(msg, use_list=True, ext_hook=ext_type_decoder, encoding=encoding)
### Setup
(Please provide relevant configs and/or SLS files (Be sure to remove sensitive info).)

### Steps to Reproduce Issue
(Include debug logs if possible and relevant.)
just run salt
### Versions Report
(Provided by running `salt --versions-report`. Please also mention any differences in master/minion versions.)

# salt --versions-report
Salt Version:
           Salt: 2018.3.3
 
Dependency Versions:
           cffi: 1.11.5
       cherrypy: Not Installed
       dateutil: Not Installed
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.10
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.6.1
   mysql-python: Not Installed
      pycparser: 2.18
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: Not Installed
         Python: 2.7.15 (default, Feb  7 2019, 01:15:05)
   python-gnupg: Not Installed
         PyYAML: 3.13
          PyZMQ: 17.1.2
           RAET: Not Installed
          smmap: Not Installed
        timelib: Not Installed
        Tornado: 4.5.3
            ZMQ: 4.3.1
 
System Versions:
           dist:   
         locale: US-ASCII
        machine: amd64
        release: 12.0-RELEASE
         system: FreeBSD
        version: Not Installed

Ofloo:
https://github.com/Ofloo/salt/commit/3e62ad17c1e8af8f03d72b65d3af647b2fb082e5#diff-30108e69fa3d1358454431c6c9b08965 Should fix it

Ch3LL:
mind pushing that as a PR?

Ofloo:
What do you mean,..?

alxwr:
> mind pushing that as a PR?

@Ch3LL Here you go: #52480

alxwr:
@Ofloo Thanks for the initial work!

[...]

Issue #51666, opened 2019-02-15
DeprecationWarning with msgpack v0.6.1


amendlik:
### Description of Issue/Question
After upgrading msgpack to v0.6.1, I get a warning on every remote execution:
```
[WARNING ][77134] /usr/local/lib/python3.6/site-packages/salt/payload.py:145: DeprecationWarning: encoding is deprecated, Use raw=False instead.
  ret = msgpack.loads(msg, use_list=True, ext_hook=ext_type_decoder, encoding=encoding)
```

### Versions Report
```
Salt Version:
           Salt: 2018.3.3
 
Dependency Versions:
           cffi: 1.11.5
       cherrypy: unknown
       dateutil: Not Installed
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.10
        libgit2: 0.27.7
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.6.1
   mysql-python: Not Installed
      pycparser: 2.18
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: 0.27.4
         Python: 3.6.8 (default, Feb  8 2019, 21:14:59)
   python-gnupg: Not Installed
         PyYAML: 3.13
          PyZMQ: 17.1.2
           RAET: Not Installed
          smmap: Not Installed
        timelib: Not Installed
        Tornado: 4.5.3
            ZMQ: 4.3.1
 
System Versions:
           dist:   
         locale: UTF-8
        machine: amd64
        release: 11.2-STABLE
         system: FreeBSD
        version: Not Installed
```

dpizzle:
I have the same issue here #51398

Issue #51398, opened 2019-01-29
Please resolve DeprecationWarning


dpizzle:
### Steps to Reproduce Issue
```
/ # salt 'vsrx01' test.ping
[WARNING ] /usr/lib/python3.6/site-packages/salt/transport/ipc.py:287: DeprecationWarning: encoding is deprecated, Use raw=False instead.
  self.unpacker = msgpack.Unpacker(encoding=encoding)

[WARNING ] /usr/lib/python3.6/site-packages/salt/payload.py:149: DeprecationWarning: encoding is deprecated, Use raw=False instead.
  ret = msgpack.loads(msg, use_list=True, ext_hook=ext_type_decoder, encoding=encoding)

vsrx01:
    True
```

### Versions Report
```
/ # salt --versions-report
Salt Version:
           Salt: 2019.2.0rc1

Dependency Versions:
           cffi: 1.11.5
       cherrypy: Not Installed
       dateutil: Not Installed
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.10
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.6.1
   mysql-python: Not Installed
      pycparser: 2.19
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: Not Installed
         Python: 3.6.6 (default, Aug 24 2018, 05:04:18)
   python-gnupg: Not Installed
         PyYAML: 3.13
          PyZMQ: 17.1.2
           RAET: Not Installed
          smmap: Not Installed
        timelib: Not Installed
        Tornado: 4.5.3
            ZMQ: 4.2.5

System Versions:
           dist:
         locale: UTF-8
        machine: x86_64
        release: 4.9.125-linuxkit
         system: Linux
        version: Not Installed
```

chenull:
i think this PR #51262 will do

keke8273:
Same issue here

[...]

Issue #50994, opened 2018-12-28
Fix this warning please.


wzlee:
```
/usr/local/Cellar/salt/2018.3.3/libexec/lib/python3.7/site-packages/salt/utils/immutabletypes.py:38: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
  class ImmutableList(collections.Sequence):
[WARNING ] /usr/local/Cellar/salt/2018.3.3/libexec/lib/python3.7/site-packages/salt/daemons/__init__.py:9: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
  from collections import namedtuple, Iterable, Sequence, Mapping

[WARNING ] /usr/local/Cellar/salt/2018.3.3/libexec/lib/python3.7/site-packages/salt/transport/ipc.py:280: DeprecationWarning: encoding is deprecated, Use raw=False instead.
  self.unpacker = msgpack.Unpacker(encoding=encoding)

[WARNING ] /usr/local/Cellar/salt/2018.3.3/libexec/lib/python3.7/site-packages/salt/payload.py:145: DeprecationWarning: encoding is deprecated, Use raw=False instead.
  ret = msgpack.loads(msg, use_list=True, ext_hook=ext_type_decoder, encoding=encoding)
```

garethgreenaway:
Fixed in https://github.com/saltstack/salt/pull/49487

BrianSidebotham:
It looks like this was merged into the 2018.3 branch but not into the develop branch? I still get this warning when using the develop branch.

garethgreenaway:
The fix in 2018.3 will be merged forward and make it's way into develop. A merge forward including that fix hasn't happened yet. Most likely sometime this week once the previous merge forward has cleared.

BrianSidebotham:
@garethgreenaway Perfect, thanks for the confirmation. :+1:

jakehilton:
Upgraded to 2018.3.4 and still seeing this:

Feb 28 17:42:41 host-smaster208 salt-master[7695]: [WARNING ] /usr/local/lib/python3.6/dist-packages/salt/payload.py:145: DeprecationWarning: encoding is deprecated, Use raw=False instead.

twangboy:
It doesn't look like that PR addresses the msgpack deprecation warning.

hackacad:
[...]

PR #53310, opened 2019-05-29
 Wrap msgpack

waynew:
### What does this PR do?

**Note:** We definitely want to run the full test suite on this

The msgpack API changed, and is in the process of changing - it will probably stabilize with version 1.0.

We don't [specify a minimum msgpack version](https://docs.saltstack.com/en/latest/topics/installation/index.html), and newer versions warn about future API changes.

This PR replaces all of the existing msgpack references with references to `salt.payload`, which has existed for a while as a msgpack wrapper, but hasn't been globally used. It also extends the wrapper to fully support the warning-free API.

https://github.com/saltstack/salt/issues/50994
https://github.com/saltstack/salt/issues/51398
https://github.com/saltstack/salt/issues/51398
https://github.com/saltstack/salt/issues/51741
https://github.com/saltstack/salt/issues/51666

There may be more...

### Previous Behavior

On more modern versions of msgpack (>=0.5.2), a warning was raised when the depreciated `encoding` parameter was sent.

### New Behavior

  According to https://github.com/msgpack/msgpack-python:

> When msgpack 1.0, I planning these breaking changes:
>
> - packer and unpacker: Remove encoding and unicode_errors option.
> - packer: Change default of use_bin_type option from False to True.
> - unpacker: Change default of raw option from True to False.
> - unpacker: Reduce all max_xxx_len options for typical usage.
> - unpacker: Remove write_bytes option from all methods.

Now we use `raw=True` if an encoding is not specified, otherwise we use `raw=False`.

### Tests written?

Yes - added a couple, and the other changes should be covered by existing tests.

### Commits signed with GPG?

Yes


rares-pop (approved):

waynew:
re-run full all

waynew:
re-run full all

dwoz (approved):

s0undt3ch:
Looks good, though I'd prefer msgpack specific handling to be on its own msgpack utils module or similar, and not mix it with salt's payload module.

Having that would allows us to transparently "fix" the lack of python set's support(see https://github.com/saltstack/salt/pull/53350) by converting it to a list for example.

waynew:
@s0undt3ch I was actually going to do that and then I came across the payload module that had comments explicitly stating that it was wrapping msgpack... it just wasn't completely wrapping it, nor was it used everywhere. 🤷‍♂ 

We *could* make that change here within this module. Though it also sounds like maybe there's a different approach for that? https://github.com/msgpack/msgpack-python/issues/241

twangboy (approved):

mattp-:
this PR confuses me. a few things:
- why is this going into 2018? 
- doesn't roughly similar code exist in develop? 
- how does this interact with threadlocalproxy?

waynew:
@mattp- historically the approach has been to push bug fixes to the oldest affected branches, where they would be continually merged forward.

There's some discussion around changing the practice, as we've seen very real problems due to that approach.

"similar" code, sort of. I've seen probably at least 4-5 PRs patching one instance or another where we patched a particular call to msgpack. We have also had the start of *this* particular wrapping code in question for quite some time, we just haven't been strict in its use. The irony is that if that we had used this payload code in the first place this would have been a 10 second fix to just change the [adapter](https://en.wikipedia.org/wiki/Adapter_pattern) like each of those (incomplete) PRs has done.

waynew:
As far as interaction with threadlocalproxy goes - I'm not sure? That's probably another piece of interaction that needs to be properly tested.

mattp-:
I would encourage you to look at the changes in the develop/neon branch as its doing the same thing, and I think your changes conflict with it

waynew:
@mattp every branch has some subset of these changes.

From the existing code it's clear that at one point we intended to wrap msgpack, it's also clear that was never completely enforced, so now we have a situation where we have several different places wrapping msgpack, but I don't believe that any of them have collected all the places.

Given our history of wanting to support as wide a range of dependencies as possible, and that (afaict) the underlying msgpack hasn't changed, just the API, it makes a lot of sense to me to go ahead and finish the original concept, wrap msgpack like we started, and replace all existing uses with the wrapped version.

I didn't find any more history, or it escapes my memory, but if you have any more light to shed on the subject, I very much welcome it! (All this to say conflicts are expected 🙃)

austinpapp:
to @s0undt3ch's point, the util exists in develop. https://github.com/saltstack/salt/blob/develop/salt/utils/msgpack.py

a few spot checks of the mods you updated use the msgpack util...

waynew:
Ah! That *is* new. And ironically just stuck a wrapped msgpack inside payload instead of unifying them 😂

waynew:
@Akm0d pretty sure you had a PR that changed things a bit - I don't remember if I have anything in this that wasn't present there.

Issue #53274, opened 2019-05-28
integration.states.test_pkg failing on amazon 1 py2


Ch3LL:
The tests:

```
integration.states.test_pkg.PkgTest.test_pkg_009_latest_with_epoch
integration.states.test_pkg.PkgTest.test_pkg_010_latest_with_epoch_and_info_installed
```

are failing on amazon 1 python 2
https://jenkinsci.saltstack.com/job/2019.2.1/job/salt-amazon-1-py2/lastFailedBuild/

dmurphy18:
Amazon Linux 1 does not have bash-completion as a package

[root@ip-10-200-45-57 ec2-user]# yum list | grep bash
bash.x86_64                           4.2.46-28.37.amzn1            installed   
bash-doc.x86_64                       4.2.46-28.37.amzn1            amzn-main   

Test is expecting bash-completion to exist on Amazon Linux 1 when it does not

19:46:02         -----------------------------------------------------------------------------------------------------------------------
19:46:02         --------  Failed Tests  -----------------------------------------------------------------------------------------------
19:46:02           -> integration.states.test_pkg.PkgTest.test_pkg_009_latest_with_epoch  ..............................................
19:46:02        Traceback (most recent call last):
19:46:02          File "/tmp/kitchen/testing/tests/support/helpers.py", line 212, in wrap
19:46:02            raise exc
19:46:02        AssertionError: False is not True. Salt Comment:
19:46:02        Error occurred installing package(s). Additional info follows:
19:46:02        
19:46:02        errors:
19:46:02            - Loaded plugins: priorities, update-motd, upgrade-helper
19:46:02              No package **_bash-completion_** available.
19:46:02              Error: Nothing to do
19:46:02           **..........................
**

dmurphy18:
Adjusted tests, https://github.com/saltstack/salt/pull/53323

dmurphy18:
Updated test after review suggestion to leverage _bash-doc_ instead of _bash-completion_ on Amazon Linux 1

KChandrashekhar:
Ok, closing this as not seeing this failure anymore

PR #53323, opened 2019-05-30
Fix for issue #53274, test on Amazon Linux 1

dmurphy18:
### What does this PR do?
Disable test_pkg tests for bash-completion on Amazon Linux 1, since it does not have that package.

https://github.com/saltstack/salt/issues/53274

### Previous Behavior
Test failures due to lack of bash-completion.

### New Behavior
Skip tests involving  for bash-completion if platform is Amazon Linux 1.
There is no package for bash-completion on Amazon Linux 1

### Tests written?
Yes - test fixed

### Commits signed with GPG?
No


tests/integration/states/test_pkg.py
```diff
@@ -488,13 +489,17 @@ def test_pkg_009_latest_with_epoch(self):
         if not pkgmgr_avail(self.run_function, self.run_function('grains.items')):
             self.skipTest('Package manager is not available')
 
+        if (grains.get('os') == 'Amazon' and grains.get('osmajorrelease') != 2):
```
rares-pop: This doesn't seem future proof. Maybe Amazon Linux 1 will support it in the future.

What if we test against a different package, that's available everywhere, like vim or something?

dmurphy18: Amazon Linux 1 reaches EOL June 20, 2020.  I think it is a done OS at this point.

tests/integration/states/test_pkg.py
```diff
@@ -506,6 +511,9 @@ def test_pkg_010_latest_with_epoch_and_info_installed(self):
...
```
Ch3LL: would it be possible to change the pkg name to something that is available on amazon linux 1, instead of skipping the test so we ensure we are also testing this on amazon linux 1? I believe it just needs to be a package that includes an epoch

Ch3LL (changes requested):

tests/integration/states/test_pkg.py
dmurphy18: I shall take a look at something else that we can install but typically would not be installed by default

dmurphy18: Using bash-doc instead of unavailable bash-completion

dmurphy18:
re-run full all

s0undt3ch (approved):

rares-pop (approved):

dmurphy18:
re-run full all

dmurphy18:
re-run full all

tests/integration/states/test_pkg.py
Ch3LL: maybe to make it more future proof for all OSs, add a check to ensure the package exists, if it does not use bash-doc? what do you think about that approach @rares-pop

PR #53326, opened 2019-05-31
core.py: ignore wrong product_name files

aplanas:
Some firmwares (like some NUC machines), do not provide valid
/sys/class/dmi/id/product_name strings. In those cases an
UnicodeDecodeError exception happens.

This patch ignore this kind of issue during the grains creation.

### DIFF
--- a/salt/grains/core.py
+++ b/salt/grains/core.py
@@ -1066,7 +1066,9 @@ def _virtual(osdata):
             except UnicodeDecodeError:
                 # Some firmwares provide non-valid 'product_name'
                 # files, ignore them
-                pass
+                log.debug(
+                    "The content in /sys/devices/virtual/dmi/id/product_name is not valid"
+                )
             except IOError:
                 pass
     elif osdata["kernel"] == "FreeBSD":
@@ -2717,7 +2719,9 @@ def _hw_data(osdata):
                 except UnicodeDecodeError:
                     # Some firmwares provide non-valid 'product_name'
                     # files, ignore them
-                    pass
+                    log.debug(
+                        "The content in /sys/devices/virtual/dmi/id/product_name is not valid"
+                    )
                 except (IOError, OSError) as err:
                     # PermissionError is new to Python 3, but corresponds to the EACESS and
                     # EPERM error numbers. Use those instead here for PY2 compatibility.
--- a/salt/grains/core.py
+++ b/salt/grains/core.py
@@ -1063,6 +1063,10 @@ def _virtual(osdata):
                         grains["virtual"] = "gce"
                     elif "BHYVE" in output:
                         grains["virtual"] = "bhyve"
+            except UnicodeDecodeError:
+                # Some firmwares provide non-valid 'product_name'
+                # files, ignore them
+                pass
             except IOError:
                 pass
     elif osdata["kernel"] == "FreeBSD":
@@ -2710,6 +2714,10 @@ def _hw_data(osdata):
                         )
                         if key == "uuid":
                             grains["uuid"] = grains["uuid"].lower()
+                except UnicodeDecodeError:
+                    # Some firmwares provide non-valid 'product_name'
+                    # files, ignore them
+                    pass
                 except (IOError, OSError) as err:
                     # PermissionError is new to Python 3, but corresponds to the EACESS and
                     # EPERM error numbers. Use those instead here for PY2 compatibility.
--- a/tests/unit/grains/test_core.py
+++ b/tests/unit/grains/test_core.py
@@ -1791,3 +1791,44 @@ class CoreGrainsTestCase(TestCase, LoaderModuleMockMixin):
         with patch.dict(os.environ, {"PATH": path}):
             result = core.path()
         assert result == {"path": path, "systempath": comps}, result
+
+    @skipIf(not salt.utils.platform.is_linux(), "System is not Linux")
+    @patch("os.path.exists")
+    @patch("salt.utils.platform.is_proxy")
+    def test__hw_data_linux_empty(self, is_proxy, exists):
+        is_proxy.return_value = False
+        exists.return_value = True
+        with patch("salt.utils.files.fopen", mock_open(read_data="")):
+            self.assertEqual(
+                core._hw_data({"kernel": "Linux"}),
+                {
+                    "biosreleasedate": "",
+                    "biosversion": "",
+                    "manufacturer": "",
+                    "productname": "",
+                    "serialnumber": "",
+                    "uuid": "",
+                },
+            )
+
+    @skipIf(not salt.utils.platform.is_linux(), "System is not Linux")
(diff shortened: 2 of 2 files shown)


aplanas:
Added tests and remove duplicate dead code for AIX (maybe comes from a wrong rebase in the merge-forward mega PR : (( )

Ch3LL (dismissed):

Ch3LL (commented):
can you take a look at the lint failures and clean those up?

aplanas:
@Ch3LL indeed, done

salt/grains/core.py
```diff
@@ -2536,6 +2536,10 @@ def _hw_data(osdata):
                         grains[key] = salt.utils.stringutils.to_unicode(ifile.read().strip(), errors='replace')
                         if key == 'uuid':
                             grains['uuid'] = grains['uuid'].lower()
+                except UnicodeDecodeError:
```
Ch3LL: just one more thing for added clarification as i was re-looking at this change. Can you elaborate what is in the file that is causing the `UnicodeDecodeError`?  I am asking because I would assume that `salt.utils.stringutils.to_unicode` should be handling this properly? Just want to make sure we aren't just hiding an actual problem.

aplanas: Right. Can be that the catch is not in the correct place, I am not sure.

The problem is that there are some devices where the /sys/class/dmi/id/product_name file is not correct, is full of 0xFF content and some other random bytes. We reported this error as we are not sure if the problem is in the manufacturer or in the kernel.

In Python 3 installations this produces an UnicodeDecodeError exception in this point when the grain is try to read the file, and this makes the minion fails to connect to the master.

Maybe you can replicate this changing the code in the minion, and reading from a different file that contains non-valid unicode points

Ch3LL: Can you add a unit test similar to https://github.com/saltstack/salt/pull/53326/files#diff-3b5d15bc59b82fc8d4b15f819babf4faR1327 that reads in this type of data and ensure it raises the unicode error?

aplanas: Actually this is how I try to test first. Injecting invalid unicode data via `read_data`, but as this is mocking `salt.utils.files.fopen` is skipping where the problem happens: in the fopen instance itself.

```
Process Process-1:
  File "/usr/lib64/python3.6/multiprocessing/process.py", line 258, in _bootstrap
    self.run()
  File "/usr/lib64/python3.6/multiprocessing/process.py", line 93, in run
    self._target(*self._args, **self._kwargs)
  File "/usr/lib/python3.6/site-packages/salt/scripts.py", line 157, in minion_process
    minion.start()
  File "/usr/lib/python3.6/site-packages/salt/cli/daemons.py", line 346, in start
    self._real_start()
  File "/usr/lib/python3.6/site-packages/salt/cli/daemons.py", line 358, in _real_start
    self.minion.tune_in()
  File "/usr/lib/python3.6/site-packages/salt/minion.py", line 1060, in tune_in
    self._spawn_minions()
  File "/usr/lib/python3.6/site-packages/salt/minion.py", line 1004, in _spawn_minions
    jid_queue=self.jid_queue)
  File "/usr/lib/python3.6/site-packages/salt/minion.py", line 976, in _create_minion_object
    jid_queue=jid_queue)
  File "/usr/lib/python3.6/site-packages/salt/minion.py", line 1130, in __init__
    self.opts['grains'] = salt.loader.grains(opts)
  File "/usr/lib/python3.6/site-packages/salt/loader.py", line 772, in grains
    ret = funcs[key]()
  File "/usr/lib/python3.6/site-packages/salt/grains/core.py", line 2088, in os_data
    grains.update(_hw_data(grains))
  File "/usr/lib/python3.6/site-packages/salt/grains/core.py", line 2529, in _hw_data
    grains[key] = salt.utils.stringutils.to_unicode(ifile.read().strip(), errors='replace')
  File "/usr/lib64/python3.6/codecs.py", line 321, in decode
    (result, consumed) = self._buffer_decode(data, self.errors, final)
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xff in position 0: invalid start byte
```

From the stack trace we can suspect that the problem is in `ifile.read().strip()`, so this is not the object that we need to mock.

Ch3LL: thanks for the clarifications. would just like to get one more review here and should be good to go.

Ch3LL:
ping @dwoz bump can you review?

aplanas:
The ci/lint complain is a bit weird, as complains for some code that this PR do not touch:

```
14:06:56  ************* Module salt.grains.core
14:06:56  salt/grains/core.py:51: [W1505(deprecated-method), linux_distribution] Using deprecated method _deprecated_linux_distribution()
```

Ch3LL:
ping @s0undt3ch any ideas why we aren't only checking the lines changed here? I checked the settings in noxfile and it looks like we are just running pylint against the files changed, not the lines changed.

waynew:
@Ch3LL  IIUC we've always been doing that, since before nox. I know I've had to fix code in unchanged parts of files that I've touched. I tried running pylint against our whole codebase once and it came up with hundreds, if not thousands of errors and warnings.

I'm not sure if a) we should keep it as-is, b) it's possible/we should change it to ignore untouched lines, or c) we should make a massive lint fix.

My initial impression is that a) is a fine approach.... but this PR is definitely not the place for a discussion about that :)

s0undt3ch:
PyLint is not capable of checking changed lines.
I'm on my phone so I can't see right away, but I bet this is against develop.
If that's the cases, I started to clean the hundreds of lint errors(because I had to upgrade PyLint due to switching to py3 only, and also because I removed some of the ignores), but that work is not yet complete.
So, expect lint issues, and please fix if so inclided, on PRs against develop.
I'll try to get that work completed once I'm back from PTO.

s0undt3ch:
By fix, I mean not all of the lint issues, just the ones triggered on PRs on the changed files.

Ch3LL:
thanks for the clarifications, i was mistaken and thought we were previously doing this. 

I put a PR in here: https://github.com/saltstack/salt/pull/54163 to ignore this pylint warning

waynew (dismissed):

salt/grains/core.py
```diff
@@ -2536,6 +2540,10 @@ def _hw_data(osdata):
...
+                    # Some firmwares provide non-valid 'product_name'
+                    # files, ignore them
+                    pass
```
dincamihai: Does it make sense to write a log entry instead of silently `pass`. We should also avoid spamming the log files with a lot of messages. The core grains are executed often.

waynew: Might be good as debug, or trace...

aplanas:
Rebased into master

Ch3LL (dismissed):

Ch3LL:
ping @waynew can you re-review here?

dwoz (approved):

PR #53344, opened 2019-06-03
Run beacons on the only one minion instance. Return to all masters.

### LINKED ISSUE #49663: Beacon only fires to single random master in HA mode
We are seeing an issue where certain beacons are only firing to a single random master. In particular this happens with the `inotify` and the `process` beacon. The problem occurs when the beacon is firing with at least one master down in an HA environment. 

@DmitryKuzmenko is already aware of this issue and some of the details.

What we expect to see is that the beacon would fire to all alive masters.

### Setup
Configure 3-4 masters and connect at least a single minion in the environment. The minions should be connected in HA mode to all masters.

### Steps to Reproduce Issue
1. Configure the `inotify` beacon on the minion to watch any file 
2. Stop the first master configured in the minion
2. Watch the master event bus on the remaining alive masters
3. Update the watched file on the minion to trigger the beacon
4. You should see the event come in only once to a random master
5. Re-triggering the beacon may send the event to a different master

### Versions Report
```
#  Minion Version
    Salt Version:
               Salt: 2018.3.2
     
    Dependency Versions:
               cffi: Not Installed
           cherrypy: Not Installed
           dateutil: Not Installed
          docker-py: Not Installed
              gitdb: Not Installed
          gitpython: Not Installed
              ioflo: Not Installed
             Jinja2: 2.7.2
            libgit2: Not Installed
            libnacl: Not Installed
           M2Crypto: Not Installed
               Mako: Not Installed
       msgpack-pure: Not Installed
     msgpack-python: 0.5.6
       mysql-python: Not Installed
          pycparser: Not Installed
           pycrypto: 2.6.1
       pycryptodome: Not Installed
             pygit2: Not Installed
             Python: 2.7.5 (default, Jul 13 2018, 13:06:57)
       python-gnupg: Not Installed
             PyYAML: 3.11
              PyZMQ: 15.3.0
               RAET: Not Installed
              smmap: Not Installed
            timelib: Not Installed
            Tornado: 4.2.1
                ZMQ: 4.1.4
     
    System Versions:
               dist: centos 7.5.1804 Core
             locale: UTF-8
            machine: x86_64
            release: 4.9.93-linuxkit-aufs
             system: Linux
```

### LINKED ISSUE #49680: Beacon events handled differently than event.send
I would like to open a discussion about how events are fired from the minion side. This discussion is brought up in observance of a difference between how beacons emit events versus how `event.fire` emits events. The difference really shows up in HA environments where the first master listed in the minion configuration is down. The problem was noticed in an HA environment with 4 masters and a couple of minions. This especially shows up with the `inotify` beacon on Linux. So we will focus on this. However, I believe that this behavior is also manifested when using the `process` beacon.

In the above environment, using IP addresses to define the masters, when the inotify beacon fires,  it will fire quickly to a single random master, even if a master is down. This is different from `salt-call event.fire` which seems to attempt to send the event serially to each master in the configured list. This is fine except that it causes a slowdown in sending the events when attempting to connect to the down master. 

I will explain some observances below.

### Setup

For testing: 

1. Configure 4 masters and at least 1 minion connected to all 4 masters in active/active:
  - Use IP address, not DNS, especially in Docker since DNS will not resolve for a stopped container (depending on your configuration)
2. I have attached my files for configuring Docker. Note that the image is built locally.
[Docker_Setup.zip](https://github.com/saltstack/salt/files/2389957/Docker_Setup.zip)
  - If you run this on Linux you may run into permission issues that do not occur on Macos since the Docker daemon runs as an admin user, not root. To solve, just fix the permissions on the keys and files under your local `saltstack` directory.
  - Be sure to review the `docker-compose.yml` file before running `docker-compose up` to ensure that all needed local directories are created.

#### Behavior of beacon (inotify specifically)
1. monitor events on all masters
 - run test.version and check events on all masters
2. Minion runs through the serially and stops at first alive master for sending
the event.
 - Takes timeout period to move on to the next 
 - There will be a log entry showing that the new connection is being made to the
   new master 
3. Bring the master back up and re-run the command and this master will be re-added 
   to the beginning. So the minion is talking to the masters serially every single time
   instead of tracking in memory which master it should talk to.

### Versions report
```
    Salt Version:
               Salt: 2018.3.2
     
    Dependency Versions:
               cffi: Not Installed
           cherrypy: Not Installed
           dateutil: Not Installed
          docker-py: Not Installed
              gitdb: Not Installed
          gitpython: Not Installed
              ioflo: Not Installed
             Jinja2: 2.7.2
            libgit2: Not Installed
            libnacl: Not Installed
           M2Crypto: Not Installed
               Mako: Not Installed
       msgpack-pure: Not Installed
     msgpack-python: 0.5.6
       mysql-python: Not Installed
          pycparser: Not Installed
           pycrypto: 2.6.1
       pycryptodome: Not Installed
             pygit2: Not Installed
             Python: 2.7.5 (default, Jul 13 2018, 13:06:57)
       python-gnupg: Not Installed
             PyYAML: 3.11
              PyZMQ: 15.3.0
               RAET: Not Installed
              smmap: Not Installed
            timelib: Not Installed
            Tornado: 4.2.1
                ZMQ: 4.1.4
     
    System Versions:
               dist: centos 7.5.1804 Core
             locale: UTF-8
            machine: x86_64
            release: 4.9.93-linuxkit-aufs
             system: Linux
```

DmitryKuzmenko:
### What does this PR do?
In multimaster environments run beacons on the only one minion instance. The instance associated with the first master in the list.
The results are reported through minion event bus to all minion instances thus fired to all connected masters.

Fixes #49680 
Fixes #49663 

### Previous Behavior
Beacons were executed on all minion instances. As a result the minion does extra work and sometimes beacons work unexpectedly like in the case of `inotify` beacon that was reporting results to the only one random master.

### New Behavior
Beacons executed once. Results reported to all connected masters.

### Tests written?
Manual.

### Commits signed with GPG?
Yes


waynew:
I'm pretty new to beacons (and the multimaster setup), so forgive me if this question doesn't make sense.

For inotify, isn't the purpose to watch a file/directory for changes? Wouldn't we want to be performing that check on each of the minions? At least the targeted ones? Or is there something with the multimaster/beacon setup that should change the way I'm thinking?

DmitryKuzmenko:
@waynew Here I'm speaking about the minion object instances in one minion process. That is the way minion works in multimaster setup. It's logically one minion in terms of SaltStack.

waynew:
Ah, that was what I was missing! So this is just saying that on each minion there are N minion processes, and only one of them should be doing the beacon work, right?

Is there some check to make sure that there is a beacon leader?

DmitryKuzmenko:
@waynew multiple minion object instances work inside one minion process. Beacon leader is the minion handling the first master in the list. Others are not. It's handled on the creation time here:
https://github.com/saltstack/salt/pull/53344/files#diff-b5732eecf0273ccf6a96e1c720198dc6R996

bryceml:
re-run full all

garethgreenaway (approved):

DmitryKuzmenko:
Replaced with #54247

shorawitz:
I'm curious to understand how the masters are handing reactors being executed on both masters?  What is filtering the same execution so you don't step on something called in the reactor?

DmitryKuzmenko:
@shorawitz If I'm not wrong there's nothing special done for this case. If you have a case where it may be useful please create a new issue.

PR #53347, opened 2019-06-03
Revert "Added bypass features to pkg.installed"

justindesilets:
Reverts saltstack/salt#50922. It was determined that this was a duplicate function of another feature and less native to the salt method. I would like to revert it and will close out my current PR that has fixes in place for it here https://github.com/saltstack/salt/pull/53320. Additionally, I will open up a new PR in the next week to update the pkg.installed documentation to include how to use the unless feature with the aggregate feature.


Ch3LL (approved):

Ch3LL:
I checked that this is not currently in a release so it should be fine to revert. Since this is in develop we will need to backport to neon before we release there to ensure the feature is not released.

sagetherage:
@justindesilets this needs to be rebased against the master branch - or is there a different PR?

justindesilets:
@sagetherage this was just to revert a feature that had been merged in but was then later determined should be left out. Do I still actually need to debase this or can it just be merged in?

sagetherage:
@justindesilets good question - we are no longer accepting PRs against the `develop` branch, so if you want the change it needs to point to the active branch for changes = `master`

justindesilets:
@sagetherage I'm looking at the current master branch for the state/pkg.py and I don't see the bypass code in there. It looks like it was never merged in the transition from develop to master. I believe this PR can be closed at this point as there isn't a change to revert.

PR #53354, opened 2019-06-03
[WIP] [2019.2.1] Tornado 5.0 support

dwoz:
### What does this PR do?

*** DO NOT MERGE ***

### DIFF
--- a/salt/client/__init__.py
+++ b/salt/client/__init__.py
@@ -1732,8 +1732,7 @@ class LocalClient(object):
                 if listen and not self.event.connect_pub(timeout=timeout):
                     raise SaltReqTimeoutError()
                 payload = channel.send(payload_kwargs, timeout=timeout)
-            except SaltReqTimeoutError as err:
-                log.error(err)
+            except SaltReqTimeoutError:
                 raise SaltReqTimeoutError(
                     'Salt request timed out. The master is not responding. You '
                     'may need to run your command with `--async` in order to '
@@ -1754,20 +1753,20 @@ class LocalClient(object):
                 payload_kwargs['key'] = self.key
                 payload = channel.send(payload_kwargs)
 
-            error = payload.pop('error', None)
-            if error is not None:
-                if isinstance(error, dict):
-                    err_name = error.get('name', '')
-                    err_msg = error.get('message', '')
-                    if err_name == 'AuthenticationError':
-                        raise AuthenticationError(err_msg)
-                    elif err_name == 'AuthorizationError':
-                        raise AuthorizationError(err_msg)
+        error = payload.pop('error', None)
+        if error is not None:
+            if isinstance(error, dict):
+                err_name = error.get('name', '')
+                err_msg = error.get('message', '')
+                if err_name == 'AuthenticationError':
+                    raise AuthenticationError(err_msg)
+                elif err_name == 'AuthorizationError':
+                    raise AuthorizationError(err_msg)
 
-                raise PublishError(error)
+            raise PublishError(error)
 
-            if not payload:
-                return payload
+        if not payload:
+            return payload
 
         return {'jid': payload['load']['jid'],
                 'minions': payload['load']['minions']}
@@ -1859,20 +1858,21 @@ class LocalClient(object):
                 payload_kwargs['key'] = self.key
                 payload = yield channel.send(payload_kwargs)
 
-            error = payload.pop('error', None)
-            if error is not None:
-                if isinstance(error, dict):
-                    err_name = error.get('name', '')
-                    err_msg = error.get('message', '')
-                    if err_name == 'AuthenticationError':
-                        raise AuthenticationError(err_msg)
-                    elif err_name == 'AuthorizationError':
-                        raise AuthorizationError(err_msg)
+        error = payload.pop('error', None)
+        if error is not None:
+            if isinstance(error, dict):
+                err_name = error.get('name', '')
+                err_msg = error.get('message', '')
+                if err_name == 'AuthenticationError':
+                    raise AuthenticationError(err_msg)
+                elif err_name == 'AuthorizationError':
+                    raise AuthorizationError(err_msg)
 
-                raise PublishError(error)
+            raise PublishError(error)
 
-            if not payload:
-                raise tornado.gen.Return(payload)
+
+        if not payload:
+            raise tornado.gen.Return(payload)
 
         raise tornado.gen.Return({'jid': payload['load']['jid'],
                                   'minions': payload['load']['minions']})
--- a/salt/client/__init__.py
+++ b/salt/client/__init__.py
@@ -1870,7 +1870,6 @@ class LocalClient(object):
 
             raise PublishError(error)
 
-
         if not payload:
             raise tornado.gen.Return(payload)
 
--- a/salt/crypt.py
+++ b/salt/crypt.py
@@ -1284,7 +1284,7 @@ class SAuth(AsyncAuth):
             close_channel = True
             channel = salt.transport.client.ReqChannel.factory(self.opts, crypt='clear')
         else:
-            cloose_channel = False
+            close_channel = False
 
         sign_in_payload = self.minion_sign_in_payload()
         try:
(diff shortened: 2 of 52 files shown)


salt/crypt.py
```diff
@@ -1291,6 +1292,8 @@ def sign_in(self, timeout=60, safe=True, tries=1, channel=None):
         if not channel:
             close_channel = True
             channel = salt.transport.client.ReqChannel.factory(self.opts, crypt='clear')
+        else:
+            cloose_channel = False
```
Akm0d: change cloose_chanel to close_channel

dwoz:
re-run docs

dwoz:
re-run full all

garethgreenaway (approved):

salt/client/__init__.py
```diff
@@ -1733,6 +1733,7 @@ def pub(self,
                 raise SaltReqTimeoutError()
             payload = channel.send(payload_kwargs, timeout=timeout)
         except SaltReqTimeoutError:
+            channel.close()
```
rares-pop: instead of doing all these channel.close() everywhere (and possibly forget to do some now, or when the code evolves), can you wrap the channel in a RAII object and don't worry about it?

salt/crypt.py
```diff
@@ -1291,6 +1292,8 @@ def sign_in(self, timeout=60, safe=True, tries=1, channel=None):
         if not channel:
             close_channel = True
             channel = salt.transport.client.ReqChannel.factory(self.opts, crypt='clear')
```
rares-pop: this could also benefit from a channel RAII object, because what you mainly want is if you create it here, to close it, otherwise let who created it to close it. That's perfect for RAII

rares-pop (commented):
Looks pretty good!

salt/client/__init__.py
rares-pop: @dwoz - yes, that's what I had in mind. From what I see in the review, you mainly want to close those channels when they are not used anymore. I hope I didn't overlooked some cases where that's not valid.

waynew: @rares-pop You're not wrong.

The only problem is that a lot of our (current) architecture wasn't thinking about context managers (some of the code maybe predates that feature??)

Unfortunately it's not a trivial change to toss in the context managers.

rares-pop: @waynew Yeah, that makes sense. However, as we touch code, it should get modernized assuming the scope of the change match the scope of the PR. 

For this particular case, I'd create the context manager and use it in 2 or 3 places to cover the places where @dwoz added the channel.close(). 

It's fine to do it in a subsequent change, or refactor that without moving to tornado 5. But I'd aim to go in that direction. :)

itamarjp:
tornado 6.x supported needed, reported at https://github.com/saltstack/salt/issues/54392

dmurphy18:
As soon as you get it working on Fedora 31 with Tornado v5.x, Fedora upgrades Tornado to v6.x :)
There is also a bugzilla tracking this on Fedora https://bugzilla.redhat.com/show_bug.cgi?id=1748170

salt/client/__init__.py
dwoz: @rares-pop I agree making the request channels context managers is a good direction to head in. I have added a commit ([c6ad52f](https://github.com/saltstack/salt/pull/53354/commits/c6ad52f6b2037c645877bf87805757480a4b5642)) that makes them context managers and uses them in `salt/client/__init__.py` and `salt/crypt.py`. One thing to note is this pattern does not work everywhere. We have some places like `salt/state.py` and `salt/fileclient.py` that store the instance of the request channel as a class attribute and and re-use it.

Context managers are not exactly the same thing as RAII. There really is no similar functionality in python. You can see [this stack overflow](https://stackoverflow.com/a/5071376) thread for more details

rares-pop: I believe it already looks better!

I agree with python as a standard there's not really an equivalent for RAII. Context managers are the closest things, but won't solve what you mentioned above.

Can we use python implementation specifics? Is salt supposed to work on *any* python implementation (i.e. CPython, Jython, IronPython, ...)?
If it is reasonable to only support CPython, __del__ is deterministic and it really behaves as a RAII, so in addition to what you did, you could write a __del__ that's doing the channel.close(), and solve that case too. I understand if that's not the case though.

salt/transport/client.py
```diff
@@ -38,15 +42,25 @@ def crypted_transfer_decode_dictentry(self, load, dictkey=None, tries=3, timeout
         '''
         raise NotImplementedError()
 
+    def __enter__(self):
+        return self
+
+    def __exit__(self):
```
rares-pop: Beware of this:
https://docs.quantifiedcode.com/python-anti-patterns/correctness/exit_must_accept_three_arguments.html

It won't get called!

salt/utils/asynchronous.py
```diff
@@ -2,95 +2,317 @@
...
+            set(async_methods + getattr(self.obj, 'async_methods', []))
+        )
+        self._close_methods = list(
+            set(close_methods + getattr(self.obj, 'close_methods', []))
+        )
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self):
```
rares-pop: diito:
https://docs.quantifiedcode.com/python-anti-patterns/correctness/exit_must_accept_three_arguments.html

salt/client/__init__.py
waynew: Actually, `__del__` is *not* deterministic - there's no guarantee that it will ever be called. I don't remember if `del <thing>` will always call `__del__` or if it just marks it *available* for deletion.

`with` blocks, on the other hand, are, and unless your program segfaults, the `__exit__` block *will* be called (or, the `finally` if you use the `@contextmanager` decorator).

The correct pattern in Python is to setup communication channels using `with` and then pass those channels to whatever long-running bits require them. There may be some other worthwhile patterns, as well, but that's the one that I've seen and used most often, when I can't isolate my channel creation and usage to a specific block.

It's possible that we could incur the cost of decorators for these things, which I *think* could be a minimally invasive change, and yet provide us with the benefits of a context manager (i.e. deterministically closing channels).

rares-pop: As I mentioned above, the standard (official documentation) says `__del__` is not deterministic. However, there are multiple implementations of python, where the most common one is CPython. In that implementation, the 'garbage collection' is done using refcounted objects. In that one, `__del__` is deterministic. When you do call `del` on a reference, or when all the references are released (goes out of scope).

Another aspect is that `__del__` is going to be called *always*, it's just that it's not specified when. Does it matter when we close the channel, or all we care is not to leak it.

You cannot use context managers in the context @dwoz was mentioning, because you create it in one place, and use it in a totally different place. We'd need something more like a shared_ptr if you are familiar with c++.

dwoz: @rares-pop Using `__del__` is an anti-pattern in most python circles. We've discussed this at length and we already run into all kinds of issues because we are using `__del__` in several places in the codebase. The problem is that the order in which things get garbage collected is non-deterministic. Often we'll be trying to do something like log an exception in `__del__` but the logger object has already been garbage collected.

In short, we're trying to stop the usage of `__del__` and wean of any cases that exist today. Properly written python code will avoid memory leaks without the use of `__del__`.

salt/transport/client.py
dwoz: Yep, I rushed through that and forgot about the `__exit__` signature.

salt/utils/asynchronous.py
dwoz: Yep, I rushed through that and forgot about the `__exit__` signature.

salt/client/__init__.py
rares-pop: @dwoz - that makes sense!

It's understandable not to rely on specific implementation of python when writing python code. I was questioning if it's okay for salt codebase (that's what I wrote in the first reply here), but I agree it's way safer not to do it. 

Disregard the `__del__` comments.

salt/master.py
```diff
@@ -1020,8 +1020,7 @@ def __bind(self):
         Bind to the local port
         '''
         # using ZMQIOLoop since we *might* need zmq in there
```
iggy: comment doesn't match code any longer

dwoz:
re-run full all

dwoz:
re-run full all

dwoz:
re-run full all

dwoz:
closing in favor of #56169

PR #53357, opened 2019-06-04
[2017.7.9] Only make a new log record if it's a dictionary.

s0undt3ch:
### What does this PR do?
Fixes a partial revert of https://github.com/s0undt3ch/salt/commit/4eeedaa38911d8bd9db1fc8d86fe2af70ee02656#diff-018cc41100faccc6dd6246e9e09e6c87 in https://github.com/s0undt3ch/salt/commit/b6cfd144b760671b138df7bd99cb19271e6f53df#diff-018cc41100faccc6dd6246e9e09e6c87

### DIFF
--- a/salt/log/setup.py
+++ b/salt/log/setup.py
@@ -1059,13 +1059,12 @@ def __process_multiprocessing_logging_queue(opts, queue):
         setup_extended_logging(opts)
     while True:
         try:
-            record_dict = queue.get()
-            if record_dict is None:
+            record = queue.get()
+            if record is None:
                 # A sentinel to stop processing the queue
                 break
             # Just log everything, filtering will happen on the main process
             # logging handlers
-            record = logging.makeLogRecord(record_dict)
             logger = logging.getLogger(record.name)
             logger.handle(record)
         except (EOFError, KeyboardInterrupt, SystemExit):


salt/log/setup.py
```diff
@@ -1059,13 +1059,14 @@ def __process_multiprocessing_logging_queue(opts, queue):
...
-            if record_dict is None:
+            record = queue.get()
+            if record is None:
                 # A sentinel to stop processing the queue
                 break
             # Just log everything, filtering will happen on the main process
             # logging handlers
-            record = logging.makeLogRecord(record_dict)
+            if isinstance(record, dict):
+                record = logging.makeLogRecord(record)
```
waynew: Would this be better as a try/except SomeException: pass?

s0undt3ch: We shouldn't actually ever hit that check, but I'm leaving it for future proofing.

waynew: IIRC it's (a microptimization, but still) faster to do a try/except than an if. Huh, interesting - at least it seems to be better if the truthy is expected:

```
Instance     0.431544249
Not instance 0.229624255
Try OK       0.3762739260000001
Try Fail     1.0968513620000002
```

From this code:
```
import timeit


def do_thing(kwargs):
    return dict(**kwargs)


def instance():
    thing = dict()
    if isinstance(thing, dict):
        thing = do_thing(thing)


def not_instance():
    thing = object()
    if isinstance(thing, dict):
        thing = do_thing(thing)

def tryish():
    thing = dict()
    try:
        thing = do_thing(thing)
    except TypeError:
        pass

def tryish_fail():
    thing = object()
    try:
        thing = do_thing(thing)
    except TypeError:
        pass


print('Instance    ', timeit.timeit(instance, number=1_000_000))
print('Not instance', timeit.timeit(not_instance, number=1_000_000))
print('Try OK      ', timeit.timeit(tryish, number=1_000_000))
print('Try Fail    ', timeit.timeit(tryish_fail, number=1_000_000))
```

s0undt3ch: Yeah, try/except is usually faster, but thinking about this again, I think I'd prefer to just remove the check and this will fully revert the initial partial revert....

twangboy (approved):

waynew (approved):

Akm0d (approved):

PR #53366, opened 2019-06-04
Acme state fixup

github-abcde:
### What does this PR do?
Edit: Expanded on the changes in this PR. Clarified the changes to behaviour in test-mode. I earlier stated that test-mode is blatantly ignored, which was incorrect.

It fixes the state `acme` by implementing correct handling of `__opts__['test']` and also properly detecting if a change is needed or not. It also fixes the returned changes (i.e. no changes returned if no changes were made, and only the changed items returned when there were changes.). Also now returns expected changes in test-mode.
Fixed handling of the `window`-parameter in `salt.modules.acme.needs_renewal` and thereby removing the broad `Exception` catching in `acme.cert`.
Moved check if certificate file actually exists into `salt.modules.acme.info`, removing 
Pylint now scores both the module and the state, as well as the unittest files with a 10 :)

None that I'm aware of.

### Previous Behavior
`acme.cert` always starts `salt.modules.acme.cert` regardless of whether a certificate renewal is needed.
No expected changes are reported in testmode, even when this is the case.
`acme.cert` always returns changes, even when no changes have been made (i.e. when a certificate was not due for renewal and no renewal had been forced).

### New Behavior
`acme.cert` does not start `salt.modules.acme.cert` when the certificate specified does not need renewal and no renewal has been forced. It also now properly returns actual changes instead of always dumping the entire certificate information for both the certificate prior to the call to `salt.modules.acme.cert` and after.
Running `acme.cert` with `__opts__['test'] == True` returns changes when a certificate would have been fetched or renewed.

### Tests written?
Yes

### Commits signed with GPG?
Yes


waynew:
Does this problem exist in earlier branches (i.e. 2018.3 and 2019.2)? If so, we should point the PR at the earliest branch with the bug.

github-abcde:
@waynew The behavior with regard to always calling `salt.modules.acme.cert` whether or not a change is needed, and with regard to not returning expected changes in test-mode has been present since the initial commit.
However, I have used new utility-functions in the unittest that were only recently added to develop (PR 52455), so simply pointing this PR to 2018.3 would break those.

waynew:
That makes sense. It would be nice if we could get the fix in the older version(s), but if that's not something that you're able to do, at least fixing it in develop is still 👍

github-abcde:
@waynew How would you suggest I proceed on getting this fix in an older version? Do I have to create a new PR for that, based on the branch to fix, and specifically include the additional utility functions? Or would that better be left for a backport initiated by you?
Or rebase this on the older branch (if that is at all possible/feasible/desired) including the additional utility functions?
Or something else perhaps :)

waynew:
@github-abcde in the past we've encouraged doing this merge forward kind of thing, but over the past couple of weeks it's become painfully obvious that merge forwards are not a good strategy.

The current plan is to encourage rebasing against each branch and creating N pull requests. I've got a [script](https://github.com/waynew/salt/blob/utils/tools/pull-requests.py) in my fork of Salt, that *should* help in the process of creating a couple of PRs. Let me know if you have any problems!

github-abcde:
@waynew Succeeded, however not with the help of your script, which works under the assumption that all goes well (it didn't, had a rebase conflict when rebasing onto 2018.3). On the bright side, I learned a bit more about git today.

waynew:
Unfortunately yes, it *does* assume the happy path 😂 I noticed on the other PR that you had made some maybe more changes? Are those changes that should also apply here?

waynew:
Looks like a minor docstring problem:
```
02:51:11  /var/jenkins/workspace/pr-doc_PR-53366/salt/modules/acme.py:docstring of salt.modules.acme.info:7:Field list ends without a blank line; unexpected unindent.
```

salt/modules/acme.py
```diff
@@ -360,4 +371,13 @@ def needs_renewal(name, window=None):
     if window is not None and window in ('force', 'Force', True):
```
waynew: It's not a problem with your code, but if you'd like to improve this, this seems silly. It's already doing a membership check, and `None` is not one of those options. Though looking at it I wonder if at one point it was `if window is not None and window.lower() in (...)`. It seems to me like this would be better as `if str(window).lower() in ('force', 'true'):`, because that's a lot more simple.

github-abcde: Indeed. Included this change in 8fce3b1d3a100ebb5afa3c97c7ced2c2735437d3.

github-abcde:
Closing this PR as https://github.com/saltstack/salt/pull/55589 has already been merged to master.

Issue #53079, opened 2019-05-16
gpg.py fails on Py 3 due to still using Py 2 conventions for str and bytes


dmurphy18:
### Description of Issue/Question
Using salt/modules/gpg.py in pure Python 3 environment, reading keys fails due to code not having been updated to allow for separation of str and bytes. 

### Setup
Running setup on salt-pack-py3 in a pure Python 3 environment (during development of RHEL 8 support - but should be possible to reproduce with RHEL 7)

### Steps to Reproduce Issue
root@c7master:/srv/salt/setup/redhat# salt c8m state.sls setup.redhat.init  pillar='{ "build_dest" : "/srv/redhat/2019.2/pkgs", "build_release" : "rhel8", "build_version" : "2019_2", "build_py3" : True }'
c8m:
----------
          ID: build_pkgs
    Function: pkg.installed
      Result: True
     Comment: All specified packages are already installed
     Started: 14:16:26.913932
    Duration: 1503.278 ms
     Changes:   
----------
          ID: root
    Function: user.present
      Result: True
     Comment: User root is present and up to date
     Started: 14:16:28.418588
    Duration: 4.778 ms
     Changes:   
----------
          ID: manage_priv_key
    Function: file.managed
        Name: /root/.gnupg/gpg_pkg_key.pem
      Result: True
     Comment: File /root/.gnupg/gpg_pkg_key.pem is in the correct state
     Started: 14:16:28.425409
    Duration: 11.759 ms
     Changes:   
----------
          ID: manage_pub_key
    Function: file.managed
        Name: /root/.gnupg/gpg_pkg_key.pub
      Result: True
     Comment: File /root/.gnupg/gpg_pkg_key.pub is in the correct state
     Started: 14:16:28.437487
    Duration: 1.735 ms
     Changes:   
----------
          ID: gpg_load_pub_key
    Function: module.run
        Name: gpg.import_key
      Result: False
     Comment: Module function gpg.import_key threw an exception. Exception: sequence item 0: expected str instance, bytes found
     Started: 14:16:28.440295
    Duration: 14.901 ms
     Changes:   
----------
          ID: gpg_load_priv_key
    Function: module.run
        Name: gpg.import_key
      Result: False
     Comment: One or more requisite failed: setup.redhat.init.gpg_load_pub_key
     Started: 14:16:28.455668
    Duration: 0.002 ms
     Changes:   

Summary for c8m
------------
Succeeded: 4
Failed:    2
------------
Total states run:     6
[...]

PR #52456, opened 2019-04-09
x509: fix type issues; a 2019.2-based variant of PR #52014, fixes #52026

alxwr:
### What does this PR do?

This is just an updated version of #52014.

### What issues does this PR fix or reference?

#52014 #52026 #52180

### Previous Behavior

`modules.x509` fails to parse existing certificates.

### New Behavior

`modules.x509` 

### Tests written?

No, but tested on FreeBSD 11.2 and on other machines (see #52026)

### Commits signed with GPG?

No

----

I just hope this helps getting #52014 into upstream. All credit goes to @arsiesys.

PR #53370, opened 2019-06-04
Update access to str/bytes with to_unicode/to_bytes for Python 3 compatibility

dmurphy18:
### What does this PR do?
Update access to str/bytes with to_unicode/to_bytes for Python 3 compatibility

https://github.com/saltstack/salt/issues/53079

### Previous Behavior
gpg.import_key would fail on Python3 throwing an error.
Found when supporting Redhat 8 with Python 3 and signing rpms

### New Behavior
Able to import a gpg key with Python 3 based minions

### Tests written?
Yes

### Commits signed with GPG?
No

### DIFF
--- a/salt/modules/gpg.py
+++ b/salt/modules/gpg.py
@@ -23,6 +23,7 @@ import time
 # Import salt libs
 import salt.utils.files
 import salt.utils.path
+import salt.utils.stringutils
 from salt.exceptions import SaltInvocationError
 from salt.utils.versions import LooseVersion as _LooseVersion
 
@@ -722,8 +723,8 @@ def import_key(text=None,
     if filename:
         try:
             with salt.utils.files.flopen(filename, 'rb') as _fp:
-                lines = _fp.readlines()
-                text = ''.join(lines)
+                text = ''.join([salt.utils.stringutils.to_unicode(x)
+                    for x in _fp.readlines()])
         except IOError:
             raise SaltInvocationError('filename does not exist.')
 
@@ -1021,7 +1022,7 @@ def sign(user=None,
                 signed_data = gpg.sign_file(_fp, keyid=keyid, passphrase=gpg_passphrase)
         if output:
             with salt.utils.files.flopen(output, 'w') as fout:
-                fout.write(signed_data.data)
+                fout.write(salt.utils.stringutils.to_bytes(signed_data.data))
     else:
         raise SaltInvocationError('filename or text must be passed.')
     return signed_data.data
@@ -1168,7 +1169,7 @@ def encrypt(user=None,
             # This version does not allow us to encrypt using the
             # file stream # have to read in the contents and encrypt.
             with salt.utils.files.flopen(filename, 'rb') as _fp:
-                _contents = _fp.read()
+                _contents = salt.utils.stringutils.to_unicode(_fp.read())
             result = gpg.encrypt(_contents, recipients, passphrase=gpg_passphrase, output=output)
         else:
             # This version allows encrypting the file stream
--- a/salt/modules/gpg.py
+++ b/salt/modules/gpg.py
@@ -723,8 +723,7 @@ def import_key(text=None,
     if filename:
         try:
             with salt.utils.files.flopen(filename, 'rb') as _fp:
-                text = ''.join([salt.utils.stringutils.to_unicode(x)
-                    for x in _fp.readlines()])
+                text = salt.utils.stringutils.to_unicode(_fp.read())
         except IOError:
             raise SaltInvocationError('filename does not exist.')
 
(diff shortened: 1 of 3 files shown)


dmurphy18:
re-run full all

alxwr:
@dmurphy18 I was able to succeed with your patches on a Ubuntu 18.04 via `salt-ssh`:

State:

```sls
# part of gpg/import.sls

gpg-import-pubkey-import-{{ name }}:
  module.run:
    - gpg.import_key:
      - user: salt
      - filename: {{ pubkey }}
    - onchanges:
      - file: gpg-import-pubkey-{{ name }}
```

Command:

```
salt-ssh -W -t <ID> state.apply gpg.import
```

Fail (before patching):

```
----------
          ID: gpg-import-pubkey-import-user                                               
    Function: module.run
      Result: False
     Comment: 'gpg.import_key' failed: sequence item 0: expected str instance, bytes found               
     Started: 09:44:56.560307
    Duration: 41.894 ms
     Changes:
```

Succeess (after patching):

```
----------
          ID: gpg-import-pubkey-import-user
    Function: module.run
      Result: True
     Comment: gpg.import_key: Success
     Started: 10:05:05.061056
    Duration: 181.364 ms
     Changes:   
              ----------
              gpg.import_key:
                  ----------
                  message:
                      Successfully imported key(s).
                  res:
                      True
```

This might be related: #52456

alxwr:
@dmurphy18 Thank you!

dmurphy18:
re-run full all

dmurphy18:
re-run full all

salt/modules/gpg.py
```diff
@@ -722,8 +723,8 @@ def import_key(text=None,
     if filename:
         try:
             with salt.utils.files.flopen(filename, 'rb') as _fp:
-                lines = _fp.readlines()
-                text = ''.join(lines)
+                text = ''.join([salt.utils.stringutils.to_unicode(x)
```
waynew: I'm not sure why this was done this way originally, but if we don't need lines and we're reading the whole thing into memory anyway this would better just be

```suggestion
                text = salt.utils.stringutils.to_unicode(_fp.read())
```

No point in creating an intermediate list and then joining it if we just want the entire file read in as unicode.

In fact, I'm not familiar with the `flopen` function but I'm assuming that we can just pass it an encoding, so maybe it makes more sense to just do something like `flopen(filename, 'r', encoding='utf-8')` (or whatever it is that we're doing under the hood in `to_unicode`, but I'm pretty sure it's that)

waynew (changes requested):
Just a minor change to reading the file

salt/modules/gpg.py
dmurphy18: Used _to_unicode_ to allow for it working correctly under Windows (Shane's suggestion).

Against pulling apart function to see how they do it, and do it that way, better to use the function and pick up any issues that may be adjust or corrected later in the function. Similar for flopen given it will do different things on Solaris or AIX.

waynew: That makes sense. 👍 for the `to_unicode` part, then, but we definitely shouldn't be doing the `''.join` bit. At least I don't see the need to create that intermediate list - `_fp.read()` should do the trick.

dmurphy18:
re-run full all

waynew (approved):

dmurphy18:
re-run full all

twangboy (approved):

garethgreenaway (approved):

tests/unit/modules/test_gpg.py
```diff
@@ -0,0 +1,221 @@
+# -*- coding: utf-8 -*-
+
+# Import Python libs
+from __future__ import absolute_import, print_function, unicode_literals
+import os
+import shutil
+
+# Posix only
+try:
+    import gpg
```
s0undt3ch: You import `gpg` here and seem to not do anything with it.

And then a bit down...

tests/unit/modules/test_gpg.py
```diff
@@ -0,0 +1,221 @@
...
+    NO_MOCK,
+    NO_MOCK_REASON,
+    MagicMock,
+    patch
+)
+
+# Import Salt libs
+import salt.utils.platform
+import salt.utils.files
+import salt.modules.gpg as gpg
```
s0undt3ch: Here, you shadow the already imported `gpg` module...

s0undt3ch (changes requested):
Just needs a clarification of the `gpg` module shadowing.

tests/unit/modules/test_gpg.py
dmurphy18: cut and paste boilerplate that I missed when creating the test file with Shane.
I shall remove it

tests/unit/modules/test_gpg.py
dmurphy18: unintentional, see previous comment - shall un-shadow.
Thanks for catching this.

tests/unit/modules/test_gpg.py
dmurphy18: Done

s0undt3ch (approved):

Issue #53333, opened 2019-05-31
Update the Aix package documentation 


KChandrashekhar:
### Description of Issue
The Aix package support was delivered in Fluorine 
https://github.com/saltstack/zh/issues/1408 

But the documentation for it is missing from https://docs.saltstack.com/en/latest/ref/modules/all/salt.modules.pkg.html

KChandrashekhar:
ZD- 2602

PR #53372, opened 2019-06-04
Added aixpkg to index

Akm0d:
### What does this PR do?
Add refs to AIXpkg documentation

https://github.com/saltstack/salt/issues/53333

### Commits signed with GPG?

Yes


Akm0d:
We only want this for 2019.2 and above

PR #53381, opened 2019-06-05
systemd_service: add firstboot function

aplanas:
systemd-firstboot is a command line tool from systend that will
set some basic configuration options, like the locale, keyboard
configuration, or the machine-id of the system.

### Tests written?

Yes

### DIFF
--- a/salt/modules/systemd_service.py
+++ b/salt/modules/systemd_service.py
@@ -1346,3 +1346,58 @@ def execs(root=None):
             continue
         ret[service] = data['ExecStart']['path']
     return ret
+
+
+def firstboot(locale=None, locale_message=None, keymap=None,
+              timezone=None, hostname=None, machine_id=None,
+              root=None):
+    '''
+    .. versionadded:: TBD
+
+    Call systemd-firstboot to configure basic settings of the system
+
+    locale
+        Set primary locale (LANG=)
+
+    locale_message
+        Set message locale (LC_MESSAGES=)
+
+    keymap
+        Set keymap
+
+    timezone
+        Set timezone
+
+    hostname
+        Set host name
+
+    machine_id
+        Set machine ID
+
+    root
+        Operate on an alternative filesystem root
+
+    CLI Example:
+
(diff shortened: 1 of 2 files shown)


aplanas:
The commit for win_network is to allow the test running in a non-windows platform. There an alternative fix somewhere: https://github.com/saltstack/salt/pull/51372#issuecomment-499124580

salt/modules/systemd_service.py
```diff
@@ -1346,3 +1346,58 @@ def execs(root=None):
...
+
+    CLI Example:
+
+        salt '*' service.firstboot keymap=jp locale=en_US.UTF-8
+
+    '''
+    cmd = ['systemd-firstboot']
+    parameters = [('locale', locale),
+                  ('locale-message', locale_message),
+                  ('keymap', keymap),
```
waynew: It would appear that keymap is not available on all systems - at least ```
CentOS Linux release 7.6.1810 (Core)
[vagrant@localhost ~]$ systemd-firstboot --keymap
systemd-firstboot: unrecognized option '--keymap'
```

Do you think it would make sense to provide a different message there, or should we just let the `unrecognized option` error bubble up?

aplanas: Ah good point!

IMHO is ok to let it go up. I will show this error message: "systemd-firstboot error: systemd-firstboot: unrecognized option '--keymap'". I think that is informative.

aplanas:
pr-kitchen-windows2016-py2 looks broken

waynew (approved):

waynew:
@aplanas Looks good 👍 Mind rebasing against develop?

aplanas:
Rebased

waynew:
Aaaand I merged one of your other PRs against develop :trollface: Looks like the issue with docs was something from building up the VM :suspect: I'm betting that it will resolve with a rebuild, which will happen with a rebase.

aplanas:
@waynew rebased again, let see! : )

PR #53389, opened 2019-06-05
only keep last 10 builds of every pr on jenkins to reduce disk usage

bryceml:
What does this PR do?

It adjusts the jenkins pipeline files to only keep the last 10 builds of PR's so we don't fill up jenkins when some pr's have around 100 builds sitting around.
What issues does this PR fix or reference?

None
Previous Behavior

Builds for pr jobs were kept until the pr was closed.
New Behavior

Builds for pr's only are kept if they are the 20 most recent builds for a given pr.
Tests written?

N/A
Commits signed with GPG?

Yes

PR #53384, opened 2019-06-05
only keep last 10 builds of every pr on jenkins to reduce disk usage

bryceml:
What does this PR do?

It adjusts the jenkins pipeline files to only keep the last 10 builds of PR's so we don't fill up jenkins when some pr's have around 100 builds sitting around.
What issues does this PR fix or reference?

None
Previous Behavior

Builds for pr jobs were kept until the pr was closed.
New Behavior

Builds for pr's only are kept if they are the 20 most recent builds for a given pr.
Tests written?

N/A
Commits signed with GPG?

Yes

See https://github.com/saltstack/salt/pull/53383

PR #53390, opened 2019-06-05
only keep last 10 builds of every pr on jenkins to reduce disk usage

bryceml:
What does this PR do?

@s0undt3ch said we should only keep 10 rather than 20 builds.

It adjusts the jenkins pipeline files to only keep the last 10 builds of PR's so we don't fill up jenkins when some pr's have around 100 builds sitting around.

None
Previous Behavior

Builds for pr jobs were kept until the pr was closed.
New Behavior

Builds for pr's only are kept if they are the 10 most recent builds for a given pr.
Tests written?

N/A
Commits signed with GPG?

Yes

### DIFF
--- a/.ci/docs
+++ b/.ci/docs
@@ -6,7 +6,7 @@ pipeline {
         timestamps()
         ansiColor('xterm')
         timeout(time: 2, unit: 'HOURS')
-        buildDiscarder(logRotator(numToKeepStr: '20'))
+        buildDiscarder(logRotator(numToKeepStr: '10'))
     }
     environment {
         PYENV_ROOT = "/usr/local/pyenv"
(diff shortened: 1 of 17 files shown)


bryceml:
See https://github.com/saltstack/salt/pull/53387
https://github.com/saltstack/salt/pull/53388
https://github.com/saltstack/salt/pull/53389
https://github.com/saltstack/salt/pull/53384

bryceml:
Note that I haven't made this change for 2019.2 since it doesn't have nox yet and it's easier to let that happen first.

s0undt3ch (approved):

bryceml:
re-run full all

garethgreenaway (approved):

Akm0d (approved):

PR #53394, opened 2019-06-06
Fix traceback when making http query to non-listening host/port and raise_error=False

terminalmage:
### What does this PR do?

When `salt.utils.http.query()` attempts to query a non-listening host/port and socket errors are suppressed, the response is `None`. This results in a traceback when the code attempts to log the length of the response before and after attempting to decompress it.

This PR simply returns without decompressing when the response resolves as `False`.

N/A

### Previous Behavior

```
  File "/testing/salt/cli/caller.py", line 237, in call
  File "/testing/salt/executors/direct_call.py", line 12, in execute
    return func(*args, **kwargs)
  File "/testing/salt/modules/http.py", line 41, in query
    return salt.utils.http.query(url=url, opts=opts, **kwargs)
  File "/testing/salt/utils/http.py", line 618, in query
    result_text = __decompressContent(coding, result_text)
  File "/testing/salt/utils/http.py", line 108, in __decompressContent
    log.trace("Decompressing %s byte content with compression type: %s", len(pgctnt), coding)
TypeError: object of type 'NoneType' has no len()
```

### New Behavior

No traceback, and the return data is:

```python
{'body': None}
```

### Tests written?

Yes

### DIFF
--- a/salt/modules/http.py
+++ b/salt/modules/http.py
@@ -16,9 +16,14 @@ import salt.utils.http
 
 def query(url, **kwargs):
     '''
+    .. versionadded:: 2015.5.0
+
     Query a resource, and decode the return data
 
-    .. versionadded:: 2015.5.0
+    raise_error : True
+        If ``False``, and if a connection cannot be made, the error will be
+        suppressed and the body of the return will simply be ``None``.
+
 
     CLI Example:
 
@@ -35,7 +40,10 @@ def query(url, **kwargs):
         opts.update(kwargs['opts'])
         del kwargs['opts']
 
-    return salt.utils.http.query(url=url, opts=opts, **kwargs)
+    try:
+        return salt.utils.http.query(url=url, opts=opts, **kwargs)
+    except Exception as exc:
+        raise CommandExecutionError(six.text_type(exc))
 
 
 def wait_for_successful_query(url, wait_for=300, **kwargs):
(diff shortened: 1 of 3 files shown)


Akm0d (changes requested):
I really appreciate that you added a test case! Is returning none better than raising a more descriptive error? Maybe returning {'body': None, 'error': 'Non-listening host/port'} so that salt registers this as a failure would be appropriate.

terminalmage:
@Akm0d So, this only happens when `raise_error=False` to suppress the error in the `socket` module when we try to make the HTTP request. If the connection can be made, the body will be a string, instead of `None`. It seems to me like an error would be redundant.

I pushed another commit which updates the docstring for this function, and makes it more clear why `None` is returned.

Akm0d (approved):
Thanks for the clarification in the docs!  It makes sense to me now why you did it this way.  Looks good.

Akm0d (approved):

salt/modules/http.py
```diff
@@ -35,7 +40,10 @@ def query(url, **kwargs):
         opts.update(kwargs['opts'])
         del kwargs['opts']
 
-    return salt.utils.http.query(url=url, opts=opts, **kwargs)
+    try:
+        return salt.utils.http.query(url=url, opts=opts, **kwargs)
+    except Exception as exc:
+        raise CommandExecutionError(six.text_type(exc))
```
Akm0d: Pylint is failing on this line with "undefined variable six" and "undefined variable CommandExecutionError"

terminalmage: Should be fixed now.

PR #53399, opened 2019-06-06
Fix test suite on develop

twangboy:
### What does this PR do?
Tests aren't running because of some issues with the win_network util again... This should take care of it.

### Tests written?
This fixes tests

### Commits signed with GPG?
Yes


salt/utils/network.py
```diff
@@ -1006,9 +1006,6 @@ def win_interfaces():
     '''
     Obtain interface information for Windows systems
     '''
-    if WIN_NETWORK_LOADED is False:
```
rares-pop: If you do remove those, do you still need the WIN_NETWORK_LOADED at all?

salt/utils/network.py
```diff
@@ -1006,9 +1006,6 @@ def win_interfaces():
...
-        # Let's throw the ImportException again
-        import salt.utils.win_network
     return salt.utils.win_network.get_interface_info()
```
rares-pop: what if importing win_network failed (the thing you tried to check against before this change)?

twangboy: This should only run on a Windows box.
For some reaseon, with the second import I was getting the following:
```
   -> unit.utils.test_network.NetworkTestCase.test_generate_minion_id_with_long_hostname
         File "c:\dev\salt\tests\unit\utils\test_network.py", line 659, in test_generate_minion_id_with_long_hostname
           minion_id = network.generate_minion_id()
         File "C:\\dev\\salt\salt\utils\network.py", line 180, in generate_minion_id
           ret = salt.utils.stringutils.to_unicode(_generate_minion_id().first())
         File "C:\\dev\\salt\salt\utils\network.py", line 169, in _generate_minion_id
           return hosts.extend([addr for addr in ip_addrs()
         File "C:\\dev\\salt\salt\utils\network.py", line 1255, in ip_addrs
           return _ip_addrs(interface, include_loopback, interface_data, 'inet')
         File "C:\\dev\\salt\salt\utils\network.py", line 1230, in _ip_addrs
           else interfaces()
         File "C:\\dev\\salt\salt\utils\network.py", line 1019, in interfaces
           return win_interfaces()
         File "C:\\dev\\salt\salt\utils\network.py", line 1011, in win_interfaces
           return salt.utils.win_network.get_interface_info()
       UnboundLocalError: local variable 'salt' referenced before assignment
```

salt/utils/network.py
twangboy: Yesh, I removed WIN_NETWORK_LOADED

rares-pop (approved):

salt/utils/network.py
s0undt3ch: Because unit tests run both on windows and linux even if just to be skipped.
You only need import in that particular function and this will clean up.

twangboy: That's what was causing the test failures.

s0undt3ch: I mean something like https://github.com/saltstack/salt/commit/10ce96161a6b49bc186152d1cca25b788978ffe8

waynew (approved):

waynew:
@twangboy shouldn't these windows tests be passing then? 😝

twangboy:
@s0undt3ch The Windows tests are failing (I believe) because `pythonnet` is not in the `zeromq-windows` static requirements file

s0undt3ch:
It needs https://github.com/saltstack/salt/pull/53409 merged

cmcmarrow (approved):

Akm0d (approved):

PR #53404, opened 2019-06-07
Bugfix: Handle multibyte characters spanning blocks of received data (2018.3)

LuAPi:
### What does this PR do?

Probably this one: https://github.com/saltstack/salt/issues/48473

### Previous Behaviour
This process would fail when a multibyte character occurs at the end of a received block of data with only part of the character in the current block. The decoding would fail with the exception UnicodeDecodeError and the message 'unexpected end of data' and include the start and end positions of failure - the start being the first byte of the split character and the end as the end of the block of data. Non-unicode encoding schemes don't necessarily report 'unexpected end of data' but the one tested (shift-jis) did still generate a UnicodeDecodeError with similar start and end locations.

### New Behaviour
This fix checks for a UnicodeDecodeError with the start position near the end of the block of data (near being within the length of the longest multibyte character in any encoding scheme that may be used [ I checked a couple and 4 seemed to be the longest] ) and the end position being the end of the received data. If such an error occurs then the received block of data is stored and empty string is returned as if no data was received. The next attempt to receive data will only attempt to read one byte and will have the previously stored data prepended to it.
The reason for only reading one byte is to make successive calls to recv attempt to read one byte at a time to complete the split character and successfully decode the received block.
If the previous data had been stored because of a malformed character that happened at the end of the block of data instead of a multibyte character that was split then the error start position will have the same index each time recv is called and after successive calls will no-longer be close enough to the end of the new combined block of received data to look like a possible split multibyte character and the decoding exception will be raised.

If the stream closes and there is stored data from the previous call to recv because decoding failed that time then next time recv is called decoding will be attempted again on the stored data in order to generate the appropriate exception.

### Tests written?

Yes

A new member variable force_receive_encoding was added to the vt to allow tests to force a specific encoding to be used to allow testing with character encodings that aren't utf-8 or whatever `__salt_system_encoding__` happens to be on the testing system.

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/utils/vt.py
+++ b/salt/utils/vt.py
@@ -123,6 +123,8 @@ class Terminal(object):
         # sys.stdXYZ streaming options
         stream_stdout=None,
         stream_stderr=None,
+        # Used for tests
+        force_receive_encoding=__salt_system_encoding__,
     ):
 
         # Let's avoid Zombies!!!
@@ -139,6 +141,7 @@ class Terminal(object):
         self.cwd = cwd
         self.env = env
         self.preexec_fn = preexec_fn
+        self.receive_encoding = force_receive_encoding
 
         # ----- Set the desired terminal size ------------------------------->
         if rows is None and cols is None:
@@ -159,6 +162,9 @@ class Terminal(object):
 
         self.child_fd = None
         self.child_fde = None
+
+        self.partial_data_stdout = b""
+        self.partial_data_stderr = b""
 
         self.closed = True
         self.flag_eof_stdout = False
@@ -604,6 +610,14 @@ class Terminal(object):
                 if not rlist:
                     self.flag_eof_stdout = self.flag_eof_stderr = True
                     log.debug("End of file(EOL). Brain-dead platform.")
+                    if self.partial_data_stdout or self.partial_data_stderr:
+                        # There is data that was received but for which
+                        # decoding failed, attempt decoding again to generate
+                        # relevant exception
+                        return (
+                            salt.utils.stringutils.to_unicode(self.partial_data_stdout),
+                            salt.utils.stringutils.to_unicode(self.partial_data_stderr),
+                        )
                     return None, None
             elif self.__irix_hack:
                 # Irix takes a long time before it realizes a child was
@@ -615,6 +629,14 @@ class Terminal(object):
                 if not rlist:
                     self.flag_eof_stdout = self.flag_eof_stderr = True
                     log.debug("End of file(EOL). Slow platform.")
+                    if self.partial_data_stdout or self.partial_data_stderr:
+                        # There is data that was received but for which
+                        # decoding failed, attempt decoding again to generate
+                        # relevant exception
+                        return (
+                            salt.utils.stringutils.to_unicode(self.partial_data_stdout),
+                            salt.utils.stringutils.to_unicode(self.partial_data_stderr),
+                        )
                     return None, None
 
             stderr = ""
@@ -646,16 +668,56 @@ class Terminal(object):
                     return None, None
             # <---- Nothing to Process!? -------------------------------------
 
-            # ----- Process STDERR ------------------------------------------>
-            if self.child_fde in rlist:
+            # ----- Helper function for processing STDERR and STDOUT -------->
+            def read_and_decode_fd(fd, maxsize, partial_data_attr=None):
+                bytes_read = getattr(self, partial_data_attr, b"")
+                # Only read one byte if we already have some existing data
+                # to try and complete a split multibyte character
+                bytes_read += os.read(fd, maxsize if not bytes_read else 1)
                 try:
-                    stderr = self._translate_newlines(
+                    decoded_data = self._translate_newlines(
                         salt.utils.stringutils.to_unicode(
-                            os.read(self.child_fde, maxsize)
+                            bytes_read, self.receive_encoding
                         )
                     )
+                    if partial_data_attr is not None:
+                        setattr(self, partial_data_attr, b"")
+                    return decoded_data, False
+                except UnicodeDecodeError as ex:
+                    max_multibyte_character_length = 4
+                    if ex.start > (
+                        len(bytes_read) - max_multibyte_character_length
+                    ) and ex.end == len(bytes_read):
+                        # We weren't able to decode the received data possibly
+                        # because it is a multibyte character split across
+                        # blocks. Save what data we have to try and decode
+                        # later. If the error wasn't caused by a multibyte
+                        # character being split then the error start position
+                        # should remain the same each time we get here but the
+                        # length of the bytes_read will increase so we will
+                        # give up and raise an exception instead.
+                        if partial_data_attr is not None:
+                            setattr(self, partial_data_attr, bytes_read)
+                        else:
+                            # We haven't been given anywhere to store partial
+                            # data so raise the exception instead
+                            raise
+                        # No decoded data to return, but indicate that there
+                        # is buffered data
+                        return "", True
+                    else:
+                        raise
 
-                    if not stderr:
+            # <---- Helper function for processing STDERR and STDOUT ---------
+
+            # ----- Process STDERR ------------------------------------------>
+            if self.child_fde in rlist and not self.flag_eof_stderr:
+                try:
+                    stderr, partial_data = read_and_decode_fd(
+                        self.child_fde, maxsize, "partial_data_stderr"
+                    )
+
+                    if not stderr and not partial_data:
                         self.flag_eof_stderr = True
                         stderr = None
                     else:
@@ -682,15 +744,13 @@ class Terminal(object):
             # <---- Process STDERR -------------------------------------------
 
             # ----- Process STDOUT ------------------------------------------>
-            if self.child_fd in rlist:
+            if self.child_fd in rlist and not self.flag_eof_stdout:
                 try:
-                    stdout = self._translate_newlines(
-                        salt.utils.stringutils.to_unicode(
-                            os.read(self.child_fd, maxsize)
-                        )
+                    stdout, partial_data = read_and_decode_fd(
+                        self.child_fd, maxsize, "partial_data_stdout"
                     )
 
-                    if not stdout:
+                    if not stdout and not partial_data:
                         self.flag_eof_stdout = True
                         stdout = None
                     else:
--- a/tests/unit/utils/test_vt.py
+++ b/tests/unit/utils/test_vt.py
@@ -12,6 +12,8 @@
 # Import Python libs
 from __future__ import absolute_import, print_function, unicode_literals
 
+import functools
+import io
 import os
 import random
 import subprocess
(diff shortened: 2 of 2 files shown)


LuAPi:
This PR is a rebase of https://github.com/saltstack/salt/pull/53234 which was merging into the wrong branch. It has the tests that @s0undt3ch requested.
It had successfully built but I think someone modified the Jenkins configurations so it re-ran the CI jobs and one of the jobs stalled for some reason. I don't know if Jenkins will rebuild if I say
re-run full ubuntu

salt/utils/vt.py
```diff
@@ -156,6 +160,9 @@ def __init__(self,
         self.child_fd = None
         self.child_fde = None
 
+        self.patrial_data_stdout = b''
```
twangboy: should this be `partial`

LuAPi: Yes, yes it should. Thanks for catching that typo.

tests/unit/utils/test_vt.py
```diff
@@ -212,3 +273,128 @@ def test_isalive_while_theres_data_to_read(self):
...
+                        buffer_e += stderr
+
+                self.assertEqual(buffer_o, expected_stdout)
+                self.assertEqual(buffer_e, expected_stderr)
+            finally:
+                term.close(terminate=True, kill=True)
+        finally:
+            # Ignore errors because on some platforms the tempdir doesn't exist
+            # which would seem be indicative of deeper problems
+            shutil.rmtree(tempdir, ignore_errors=True)
```
waynew: It would probably be better to use `from tests.support.helpers import with_tempdir` and use the `@with_tempdir` decorator here.

LuAPi: Thanks for the suggestion, that is a neater way of doing it so I've switched to doing that.
I'm currently looking into making it not use the temp files since some of the testing environments seem to be missing the temporary directory so creating the temp directory fails.
I had originally not used the temp-files but found that they were the most reliable way to be sure I got exactly the same stream of bytes on all platforms.
I'm not sure if the missing temp dir a transient error so I'll try rerunning the tests.

LuAPi:
re-run full all

LuAPi:
There seems to be a problem with the `sys.stdout.fileno` method in the CI system. The VT uses this method internally so in order to make the VT work I had to do something to make that method work again.

I previously tried to work around that by detecting if the method was broken using the `stdout_fileno_available` function I added, and if it was broken I spawned a new test process and asserted that it succeeded. This broke the tests because I was unaware that the testing process makes a named temp directory and then deletes it when it has finished. The cleanup of the new testing process deleted the temp dir. Specifying a different temp dir for the subprocess wasn't possible because `tests/support/paths.py` only lets you specify a temp directory on all platforms except MacOS. The subprocess fix also was a bad workaround because it didn't properly forward the current test conditions (like what transport to use).

I looked into the proper solution of actually fixing the problem of `sys.stdout.fileno()` not working. I found that `sys.stdout` (and `sys.stderr`) get replaced in `tests/support/xmlunit.py`. The code there looks like it shouldn't cause a problem. That probably needs more investigation but that seems like scope creep for the PR. I've changed the workaround to replacing `sys.stdout` and `sys.stderr` with `sys.__stdout__` and `sys._stderr__` for the duration of the VT tests and restoring them afterwards. This has fixed the VT tests on all platforms except windows which is currently failing for things unrelated to this PR.

Should I just disable my new VT tests (like the rest of the existing VT tests) or maybe skip them if they're running on windows?

tests/unit/utils/test_vt.py
```diff
@@ -16,19 +16,34 @@
...
 import salt.utils.files
 import salt.utils.platform
 import salt.utils.vt
+import salt.utils.stringutils
 
 # Import 3rd-party libs
 from salt.ext.six.moves import range  # pylint: disable=import-error,redefined-builtin
 
 
+def stdout_fileno_available():
```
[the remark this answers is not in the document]
LuAPi: I think the cleanest option would be to put this test, the switch from the current stdout and stderr if necessary ( https://github.com/saltstack/salt/pull/53404/files#diff-64e62f3c1ef41b8920ec7488db35ec28R246-R252 ) and the restore ( https://github.com/saltstack/salt/pull/53404/files#diff-64e62f3c1ef41b8920ec7488db35ec28R246-R252 ) in a decorator. I'm not sure whether I should put the decorator code in this file since it's probably only needed here or whether to add it to one of the libraries of decorators that already exist.

The intent is to report given the environment the code is executing in whether you will get an exception if you try and call sys.stdout.fileno() at this point. Since we do modify the environment such that this will produce different results if called at different points it's safer to not cache the result since the result would be wrong if someone fails to properly restore the environment after changing it. It's not a particularly expensive operation and it isn't called that often.

LuAPi:
> I'd recommend rebasing and squashing your commits - if you're not comfortable with doing that sort of thing I'd be happy to help out 👍

Thanks for the offer but I think I can manage. Presumably I should force push the rebased and squashed commit to this branch rather than open a new PR?
I'll do that after we've resolved the `def stdout_fileno_available()` conversation you opened and we've decided that there aren't any more changes to be made.

waynew:
Yeah *usually* rebasing isn't a big deal - it's only when you've made changes that will result in a conflict.

I always find the Hitchhiker's Guide advice to be great in that situation though: Don't Panic!

A `git rebase --abort` will get you back to where you were if things go slightly awry. And if *everything* has gone sideways you should still be able to get back your info from the reflog.

Of course, if you want to make sure you still have a place to get back to if you rebase you can just create a new branch at that commit, e.g.

```
git checkout real-work
git branch backup
git rebase <wherever>
```

Then worst case you can always
```
git checkout real-work
git reset --hard backup
```
to force `real-work` to point to whatever commit you set with `backup`.

waynew:
It *appears* that something in here is breaking some other tests - it *might* be that there's a problem when patching stderr/out. I suspect that wrapping the `if` in that try/finally would be a good idea, too.

LuAPi:
Yep I made a bunch of backup branches. I've found rebasing has generally gone a lot better since I learned what it's actually doing.

> It _appears_ that something in here is breaking some other tests - it _might_ be that there's a problem when patching stderr/out. I suspect that wrapping the `if` in that try/finally would be a good idea, too.

I'm not sure about that. Doing that would silence any exceptions from stdout_fileno_available .
When I wrote stdout_fileno_available I thought about not specifying an exception type and just catching everything but then I thought I should catch the specific thing I expect to go wrong and let the system raise any other unexpected things. Not catching everything means that if someone does something weird in some other tests to the stdout and stderr this test should fail and let them know they broke things for other tests.

Are you sure it's caused by my tests because of the tests that failed the errors all seem to be appearing elsewhere:
- Windows Py2 - Tests are disabled on windows
- Windows Py3 - Tests are disabled on windows
- Py2 Debian 9 - Timeout also seen here https://github.com/saltstack/salt/issues/54182
- Py2 Ubuntu 18 - Timeout also seen here https://github.com/saltstack/salt/issues/54182
- Py3 Centos 7 - Timeout also seen here https://github.com/saltstack/salt/issues/54182
- Py2 Ubuntu 16.04 - unit.utils.test_ssdp.SSDPFactoryTestCase.test_datagram_ test failures also observed in this trivial PR https://github.com/saltstack/salt/pull/54354
- Py3 Ubuntu 16.04 - unit.utils.test_ssdp.SSDPFactoryTestCase.test_datagram_ test failures also observed in this trivial PR https://github.com/saltstack/salt/pull/54354
- Py3 Amazon Linux 2  - unit.modules.test_file.FilemodLineTests.test_line_insert_ensure_beforeafter_rangelines test failure also observed in this trivial PR https://github.com/saltstack/salt/pull/54354

I've pushed a commit to disable the tests on all platforms so we can see it the tests are breaking other things. After that I'm thinking I'll push a commit reverting all my changes and just adding a comment to vt.py and test_vt.py since I'm a bit uncertain about how well the CI system is working right now since I've seen the same seemingly unrelated failures in other PRs and I've had to implement workarounds for weirdness in the test environment.

LuAPi:
After disabling the tests these failed:
- Windows Py2 - same as before
- Windows Py3 - same as before
- CentOS 7 Py2 - SSH session could not be established - Transient error? - new error
- CentOS 7 Py3 - SSH session could not be established - Transient error? - different error but same as Py2 version so some transient problem that killed it before it reached the other error?
- Py2 Debian 8 - same error as before
- Py3 Fedora 29 - same error as before
- Py2 Fedora 29 - integration.states.test_pkg.PkgTest.test_pkg_* - This didn't happen when the tests were enabled, random error looks like some package failed to be fetched properly? - new error
- Py2 Ubuntu 16.04  - same error as before
- Py3 Ubuntu 16.04 -  same error as before
- Py3 Amazon Linux 2  -  same error as before

Tests that passed this time:
- Py2 Ubuntu 18 - It didn't have the timeout, but other tests did so maybe the timeout thing is a race condition or random?

I'll try reverting all my changes and just adding comments to both files to trigger the same set of tests to give us a baseline for what the test suite is doing.

LuAPi:
The CI results for just adding comments to both files (I ran that twice):
- Py3 Ubuntu 16.04 - Failed both times with unit.utils.test_ssdp.SSDPFactoryTestCase.test_datagram_ test failures
- Py2 Ubuntu 16.04 - Failed both times with unit.utils.test_ssdp.SSDPFactoryTestCase.test_datagram_ test failures
- Py2 Windows Failed both times
- Py3 Windows Failed both times
- Py3 Amazon Linux 2 - Failed both times with unit.modules.test_file.FilemodLineTests.test_line_insert_ensure_beforeafter_rangelines
- Py3 CentOS 7 - Failed both times due to timeout - "Cancelling nested steps due to timeout"

I think that rules out the changes in this PR as the cause of those particular types of errors. I think the "Cancelling nested steps due to timeout" error is mostly random so should be ignored if it happens on other platforms.

waynew:
> * Py3 Ubuntu 16.04 - Failed both times with unit.utils.test_ssdp.SSDPFactoryTestCase.test_datagram_ test failures
> * Py2 Ubuntu 16.04 - Failed both times with unit.utils.test_ssdp.SSDPFactoryTestCase.test_datagram_ test failures

I'm re-running this py2 build, I want to see if there's anything that we've changed in our build system that would change this (though I suspect not).

If nothing has changed here, I want to investigate whether or not these changes have any impact on these tests.

As a side note - we've been devoting a *lot* of effort to getting our test infrastructure stable and working on 2019.2.1. I may try rebasing your changes on top of that to see if these still fail.

> * Py2 Windows Failed both times
> * Py3 Windows Failed both times

One of those 2019.2.1 fixes that we've made was stabilizing the Windows tests quite a bit.

> * Py3 Amazon Linux 2 - Failed both times with unit.modules.test_file.FilemodLineTests.test_line_insert_ensure_beforeafter_rangelines
> * Py3 CentOS 7 - Failed both times due to timeout - "Cancelling nested steps due to timeout"
> 
> I think that rules out the changes in this PR as the cause of those particular types of errors. I think the "Cancelling nested steps due to timeout" error is mostly random so should be ignored if it happens on other platforms.

You might be right about this PR causing the errors - but ignoring "mostly random" errors in a test framework is something that we've seen cause a tremendous instability in our test systems. It should go without saying, but we don't want to do that anymore 🙃 

So we'll figure out if it's just the instability on the underlying 2018.3 build, or if there are some surprising an unexpected interactions going on here, so we can get this fix in 👍

LuAPi:
re-run full amazon2

LuAPi:
re-run full windows2016

LuAPi:
re-run full windows2019

LuAPi:
re-run full amazon2-py2

LuAPi:
re-run full amazon2-py2

LuAPi:
Looks like the CI system has had enough consistent failures fixed now that all the tests were able to pass so I've removed the [WIP] tag.

waynew:
I've just updated the most recent changes from master, let's see how this works 👍

waynew:
py3/windows2019 failed, but I'm pretty sure that was unrelated. Re-running tests, but then this should be good to go.

waynew:
👀 Well. The py3/windows2019 build is still failing, but I'm pretty sure it's an infrastructure issue. That ci/py3/centos8 build was just a misconfigured test run and can be ignored.

Checking with the team on what we should do for this failing windows build, but otherwise this looks good to me 👍

waynew (dismissed):

LuAPi:
I've rebased on top of master to clean up the history, no other changes.
I agree with your assessment of the py3 windows 2019 failures.
The first one looked to me like what https://github.com/saltstack/salt/pull/55718 attempted to reduce the likeliness of. The second one seemed to be that the instance didn't have enough resources.

LuAPi:
Fixed merge conflicts caused by the blackening.

LuAPi:
I rebased onto master yesterday. It seems that the tests are taking longer to run than the period between updates to master. It doesn't seem easy to keep this in a state where all tests have finished running and it is up to date with master. I don't think it's too far behind but let me know if you want another rebase.

waynew (approved):
🚀

waynew:
@LuAPi no, that's fine - unless there's a merge conflict, this should be ready to go :) (codecov should just be considered informational, we've merged a PR to the latest master that should make that official)

PR #53425, opened 2019-06-10
DOC pkg.installed unless option

justindesilets:
### What does this PR do?
DOC - This adds documentation on how to use the ``pkg.installed`` state with the unless requisite combined with ``file.exists`` and file.search options. I have tested the examples I have provided in the documentation. One particular example is using ``file.search`` function when the file doesn't exist yet. This does not cause any issues in my testing.

There was discussion around the ``pkg.installed`` ``file.bypass`` options I had added. The general opinion was that it seemed more like a hack and less of a salt / python way of solving the problem. I've submitted a revert request here https://github.com/saltstack/salt/pull/53347 to have the bypass feature removed. I have tried to do my best in structuring the text in a reasonable way but am happy to update it with suggestions.


mchugh19:
Thanks for all your work on this front! If the aggregate solution also helped, might it be worth adding a final note linking to those docs as an alternative?

justindesilets:
@mchugh19 I've just been able to implement this in our dev environment but there hasn't been a good opportunity to test and validate. If I have some concrete testing and validation, I will revisit this documentation and update it. I'm hoping to push it to prod in a week or so. That will allow us to get a much better idea on what is a reasonable gain in state run times.

Issue #53276, opened 2019-05-28
unit.modules.test_win_file.WinFileTestCase.test_issue_52002_check_file_remove_symlink failing


Ch3LL:
The test `unit.modules.test_win_file.WinFileTestCase.test_issue_52002_check_file_remove_symlink` is failing on both windows 2016 and 2019 python 2

https://jenkinsci.saltstack.com/job/2019.2.1/job/salt-windows-2016-py2/lastFailedBuild/
https://jenkinsci.saltstack.com/job/2019.2.1/job/salt-windows-2019-py2/lastFailedBuild/

it looks like https://github.com/saltstack/salt/pull/53157 attempted to resolve this error but i'm still seeing it fail on the most recent branch builds.

cmcmarrow:
I have replicate the error

PR #53437, opened 2019-06-11
Fix failing symlink test

twangboy:
### What does this PR do?
Fix symlink tests again...

https://github.com/saltstack/salt/issues/52781
https://github.com/saltstack/salt/issues/53276

### Tests written?
Yes

### Commits signed with GPG?
Yes

### DIFF
--- a/tests/unit/modules/test_win_file.py
+++ b/tests/unit/modules/test_win_file.py
@@ -78,31 +78,6 @@ class WinFileTestCase(TestCase, LoaderModuleMockMixin):
             self.assertRaises(
                 CommandExecutionError, win_file.check_perms, self.FAKE_PATH)
 
-    @destructiveTest
-    @skipIf(not salt.utils.platform.is_windows(), 'Skip on Non-Windows systems')
-    def test_issue_52002_check_file_remove_symlink(self):
-        '''
-        Make sure that directories including symlinks or symlinks can be removed
-        '''
-        base = temp.dir(prefix='base-')
-        target = os.path.join(base, 'child 1', 'target\\')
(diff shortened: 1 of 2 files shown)


tests/unit/modules/test_win_file.py
```diff
@@ -324,6 +320,8 @@ def test_check_perms_reset(self):
             reset=True)
         self.assertDictEqual(expected, ret)
 
+    @destructiveTest
```
s0undt3ch: If the symlink lives on a temporary directory the test suite created then this is not a destructive test.

s0undt3ch (approved):

tests/unit/modules/test_win_file.py
twangboy: Sweet. I thought it was any modification to the system

s0undt3ch: We're free to destroy whatever temp files/dirs we create :)

twangboy: So, those were being created in the user temp directory. I modified all tests in the module to use Salt's temp directory and removed the destructive decorator

s0undt3ch (approved):

s0undt3ch (approved):

cmcmarrow (approved):

Akm0d (approved):

twangboy:
These tests are passing on both Py2 and Py3 on Windows using kitchen. So, if it's failing in the entire test run, there must be something changing the privs of the user running the test. I added priv output to the error message to hopefully demonstrate the issue.

PR #53440, opened 2019-06-11
(Backport 53167) loader: support extra modules in sys.path

aplanas:
### What does this PR do?

The LazyLoader class populate the sys.path array to add the path
that contains the Python module that will be executed.  If before
this point sys.path gets extended (like when globally the extmods
utils directory in added in some places, like after a sync) the
code can find all the dependencies.

This patch adds a new parameter to the LazyLoader class,
extra_module_dirs, that will populate locally the sys.path before
executing the the module.

This can be used to guarantee that in utils (and other modules) will be
available to the modules and states in any circumstances and Python
process.

### Tests written?

Yes

(backport #53167)

### DIFF
--- a/salt/loader.py
+++ b/salt/loader.py
@@ -16,7 +16,6 @@ import inspect
 import tempfile
 import threading
 import functools
-import threading
 import traceback
 import types
 from zipimport import zipimporter
--- a/salt/loader.py
+++ b/salt/loader.py
@@ -256,6 +256,7 @@ def minion_mods(
         whitelist=whitelist,
         loaded_base_name=loaded_base_name,
         static_modules=static_modules,
+        extra_module_dirs=utils.module_dirs if utils else None,
     )
 
     ret.pack['__salt__'] = ret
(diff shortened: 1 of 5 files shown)


waynew (approved):
Pretty sure the code is fine, and the test failure is due to a dependency failing to get installed.

We need to get that sorted out so the tests all pass.

cmcmarrow (approved):

Akm0d (approved):

waynew:
re-run all

waynew:
@aplanas is there any way that these git test failures

- https://jenkinsci.saltstack.com/job/pr-kitchen-centos7-py3/job/PR-53440/6/#showFailuresLink
- https://jenkinsci.saltstack.com/job/pr-kitchen-fedora29-py3/job/PR-53440/5/#showFailuresLink
- https://jenkinsci.saltstack.com/job/pr-kitchen-ubuntu1804-py3/job/PR-53440/5/#showFailuresLink

are related to this change? It seems unlikely to me without further digging, and weird that it's just these couple of platforms on which it's failing... 🤷‍♂

aplanas:
@waynew I rebased the code, to relaunch the CI

aplanas:
Rebased on master as part of https://github.com/saltstack/salt/pull/53167

PR #53447, opened 2019-06-11
xfs: do not fails if type is not present

aplanas:
The command `blkid -o export` not always provides a 'TYPE' output
for all the devices. One example is non-formatted partitions, like for
example the BIOS partition.

This patch do not force the presence of this field in the blkid
output.

### Tests written?

Yes

### DIFF
--- a/tests/unit/modules/test_xfs.py
+++ b/tests/unit/modules/test_xfs.py
@@ -1,0 +1,52 @@
+# -*- coding: utf-8 -*-
+
+# Import Python libs
+from __future__ import absolute_import, print_function, unicode_literals
+
+import textwrap
+
+# Import Salt Libs
+import salt.modules.xfs as xfs
+
+# Import Salt Testing Libs
+from tests.support.mixins import LoaderModuleMockMixin
+from tests.support.mock import MagicMock, patch
+from tests.support.unit import TestCase
+
+
+@patch("salt.modules.xfs._get_mounts", MagicMock(return_value={}))
+class XFSTestCase(TestCase, LoaderModuleMockMixin):
+    """
+    Test cases for salt.modules.xfs
+    """
+
+    def setup_loader_modules(self):
+        return {xfs: {}}
+
+    def test__blkid_output(self):
+        """
+        Test xfs._blkid_output when there is data
+        """
+        blkid_export = textwrap.dedent(
+            """
+            DEVNAME=/dev/sda1
+            UUID=XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXXXX
+            TYPE=xfs
+            PARTUUID=YYYYYYYY-YY
+
+            DEVNAME=/dev/sdb1
+            PARTUUID=ZZZZZZZZ-ZZZZ-ZZZZ-ZZZZ-ZZZZZZZZZZZZ
+            """
+        )
+        # We expect to find only data from /dev/sda1, nothig from
+        # /dev/sdb1
+        self.assertEqual(
+            xfs._blkid_output(blkid_export),
+            {
+                "/dev/sda1": {
+                    "label": None,
+                    "partuuid": "YYYYYYYY-YY",
+                    "uuid": "XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXXXX",
+                }
+            },
+        )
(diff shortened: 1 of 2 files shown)


twangboy (dismissed):

aplanas:
Rebased, in hope that the windows tests are fixed

aplanas:
backport in: https://github.com/saltstack/salt/pull/53941

aplanas:
Rebased into master

waynew (dismissed):
@aplanas looks great so far! Could you add one more test case, basically a dupe of the existing one, but remove the `TYPE=xfs` line?

We want to make sure that if `TYPE` is missing that we don't error, right? Looks like the existing test should pass with or without your change, but without companion test it's possible that someone could introduce a regression in the future.

(If there is already another test that does that, feel free to point me in that direction - just not seeing it near this code.)

aplanas:
> @aplanas looks great so far! Could you add one more test case, basically a dupe of the existing one, but remove the `TYPE=xfs` line?
> 
> We want to make sure that if `TYPE` is missing that we don't error, right? Looks like the existing test should pass with or without your change, but without companion test it's possible that someone could introduce a regression in the future.

I do not understand. This is already tested. You can see the /dev/sdb1 do not have a TYPE parameter, and the assertion is based only on the data of /dev/sda1.

To make it more clear I will change the PARTUUID section from /dev/sdb1 to be explicit that we do not expect it.

aplanas:
> Ah! I didn't realize that those were two separate blocks, my bad! I'm +1 now.

No problem! :+1: 

> Looks like we have some similar test failures to that other PR of yours. Checking on these failures now.

Do you think that is only on my PR? :thinking:

waynew:
The mac issues I know are global, and there are some timeout issues that I believe we *finally* have a resolution for (dhcp leases during long-running suites). The Windows failures I don't think I've seen recently - they might be related, though that would be surprising, I think 🤔

aplanas:
> The Windows failures I don't think I've seen recently - they might be related, though that would be surprising, I think

XFS tests are indeed not expected to be executed on Windows machines. And checking the log I do not see them.

I am very bad on Windows, so I do not think that I can help on the diagnosis.

tests/unit/modules/test_xfs.py
```diff
@@ -0,0 +1,50 @@
...
+# Import Python libs
+from __future__ import absolute_import, print_function, unicode_literals
+import textwrap
+
+# Import Salt Testing Libs
+from tests.support.mixins import LoaderModuleMockMixin
+from tests.support.unit import skipIf, TestCase
+from tests.support.mock import (
+    NO_MOCK,
+    NO_MOCK_REASON,
```
waynew:
```suggestion
```

tests/unit/modules/test_xfs.py
```diff
@@ -0,0 +1,50 @@
...
+    MagicMock,
+    patch)
+
+# Import Salt Libs
+import salt.modules.xfs as xfs
+
+
+@skipIf(NO_MOCK, NO_MOCK_REASON)
```
waynew:
```suggestion
```

waynew (changes requested):
Hey @aplanas, thanks for the PR!

We've recently updated `master` to assume that mock actually exists everywhere. If mock doesn't at this point that should be a big problem. Can you go ahead and remove that from your tests?

tests/unit/modules/test_xfs.py
aplanas: Yes, thanks!!

tests/unit/modules/test_xfs.py
aplanas: Thanks!

aplanas:
Yep. Ignore the lint, is not related with my code:

************* Module salt.returners.django_return
salt/returners/django_return.py:42: [E0611(no-name-in-module), ] No name 'dispatch' in module 'django'

waynew:
Something weird seems to have happened with the tests, probably jenkins+aws+github. Updating your branch to see if that fixes it.

waynew:
looks like things are at least reporting now. A minor isort complaint, but I'm strongly inclined to ignore that if there aren't any other issues here.

aplanas:
Looks like the the `isort` pinned in the project produces a different output that the official release of `isort` in relation to comments on the import. Also I saw how the interaction between black a isort produces inestable results, so one do / undo the change of the other.

waynew:
> Looks like the the `isort` pinned in the project produces a different output that the official release of `isort` in relation to comments on the import

That's why we've pinned to a different version. The pinned version in question provides an option that agrees with Black on imports.

waynew (dismissed):
Looks good 👍

aplanas:
no failing test, re-review?

dwoz (approved):

PR #53461, opened 2019-06-12
Update file.chattr

waynew:
Cherry pick https://github.com/saltstack/salt/pull/53188 into 2019.2.1

PR #53463, opened 2019-06-12
Decode result_text instead

waynew:
### What does this PR do?

Fixes a test failure as a precursor to https://github.com/saltstack/salt/pull/53461

#53461 - the Centos test failure

### Previous Behavior

If we were on Python3 and the request was an HTTPRequest instead of a requests library requests it would blow up trying to re-read the request body, because the API is different.

### New Behavior

No longer blows up. Now it just uses the already existing request body instead of reads it. Also removed the flaky decorator. At least, I'm operating under the assumption that this was the cause of flakiness 😛 

### Tests written?

Fixes existing test

### Commits signed with GPG?

Yes


Ch3LL (approved):

waynew:
re-run lint

waynew:
Looks like the code behind was already merged in, aside from the flaky decorator removed. So... just going to close this.

Issue #53457, opened 2019-06-11
Windows: reg.present failing with "NotImplementedError: Ace type 9 is not supported yet"


ctr49:
### Description of Issue
When setting an (existing) registry value (possibly having permissions containing an unknown user), salt fails to apply.

### Setup
example sls:
```
HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\PolicyManager\current\device\System\AllowExperimentation:
  reg.present:
    - name: HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\PolicyManager\current\device\System
    - vname: AllowExperimentation
    - vdata: 0
    - vtype: REG_DWORD
```
### Steps to Reproduce Issue
Apply above state
```
----------
          ID: HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\PolicyManager\current\device\System\AllowExperimentation
    Function: reg.present
        Name: HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\PolicyManager\current\device\System
      Result: False
     Comment: An exception occurred in this state: Traceback (most recent call last):
                File "c:\salt\bin\lib\site-packages\salt\state.py", line 1933, in call
                  **cdata['kwargs'])
                File "c:\salt\bin\lib\site-packages\salt\loader.py", line 1939, in wrapper
                  return f(*args, **kwargs)
                File "c:\salt\bin\lib\site-packages\salt\states\reg.py", line 414, in present
                  reset=win_perms_reset)
                File "c:\salt\bin\lib\site-packages\salt\utils\win_dacl.py", line 2343, in check_perms
                  cur_perms = get_permissions(obj_name=obj_name, obj_type=obj_type)
                File "c:\salt\bin\lib\site-packages\salt\utils\win_dacl.py", line 1680, in get_permissions
                  return obj_dacl.list_aces()
                File "c:\salt\bin\lib\site-packages\salt\utils\win_dacl.py", line 862, in list_aces
                  ace = self.dacl.GetAce(i)
              NotImplementedError: Ace type 9 is not supported yet
     Started: 00:33:36.070295
    Duration: 51.255 ms
     Changes:   
----------
```
### Versions Report
```
    Salt Version:
               Salt: 2019.2.0
     
    Dependency Versions:
               cffi: 1.11.5
           cherrypy: 17.4.1
           dateutil: 2.7.5
          docker-py: Not Installed
              gitdb: 2.0.5
          gitpython: Not Installed
[...]

PR #53464, opened 2019-06-12
Stops salt from crashing if ace 9 error happens.

cmcmarrow:
### What does this PR do?
Fixes Ace 9 error
https://github.com/saltstack/salt/issues/53457

### Tests written?
No

### Commits signed with GPG?

Yes


twangboy (changes requested):
Looks great. Could you add a link to the PyWin32 docs where it states what Aces are supported to justify why we're just skipping here.

Akm0d (approved):

salt/modules/win_dacl.py
```diff
@@ -3,6 +3,15 @@
 Manage DACLs on Windows
 
 :depends:   - winreg Python module
+
+NOTE: pywin32 does not support some ace types like ac9.
+If a non implemented ace type is give to this module it will just be skipped.
```
twangboy: If the object contains a non-implemented ace type it will just be skipped. You see similar behavior in the Windows GUI when you view the permissions of an object with some ace objects.

twangboy (changes requested):

xeacott (approved):

twangboy (approved):

PR #53467, opened 2019-06-12
Check valid username first (fixes failing symlink test)

twangboy:
### What does this PR do?
The failing symlink test had to do with two tests in `unit.modules.test_cmdmod` that were testing `runas` on Windows. For some reason, this was causing the symlink test to fail later on. It was getting the system token, then checking to see if the username was valid before returning a `CommandExecutionError`. I moved the username validation to be the first thing the function does before grabbing the system token. This fixed the failing symlink test.

Fixes `log/setup.py` so that it doesn't bork if `__salt_system_encoding__` is undefined
Fixes `modules/file.py` `_add_content` function to pass the correct parameters to the `check_perms` function when on Windows
Removes some extra error messages that were being used for troubleshooting in the `modules/win_file.py` `symlink` function
Improves the error messages displayed by functions in `platform/win.py`
Fixes `states/file.py` `copy_` function to pass the correct parameters to the `check_perms` function when on Windows
Fixes the `__virtual__` function on `states/zfs.py` and `states/zpool.py` when the `zfs_support` grain doesn't exist (as it doesn't on Windows)
Wraps some code in the `runas` and `runas_unprov` functions in `utils/win_runas.py` salt util in a `try/finally` block to make sure handles are being closed if an error is encountered.
Fixes some issues with `integration.states.test_file` when trying to remove files that don't exist. Also uses `os.mkdir` on Windows as the mode is just ignored. Additionally, uses more descriptive assert statements.
Fixes more issues with `unit.modules.test_file` for handling the fact that `win_file.py` namespaces many functions from the `file.py` module. So this patches those tests to run the correct `check_file` function. Also skips 2 Linux specific tests on Windows.

This is the rest of https://github.com/saltstack/salt/pull/53437

### Tests written?
Fixes tests

### Commits signed with GPG?
Yes

### DIFF
--- a/tests/unit/modules/test_file.py
+++ b/tests/unit/modules/test_file.py
@@ -33,6 +33,15 @@ import salt.modules.cmdmod as cmdmod
 from salt.exceptions import CommandExecutionError, SaltInvocationError
 from salt.utils.jinja import SaltCacheLoader
 
+if salt.utils.platform.is_windows():
+    import salt.modules.win_file as win_file
+    import salt.utils.win_dacl as win_dacl
+    # We have to monkey patch this here because many of the functions from
+    # `filemod` module are "namespaced" into the `win_file` module. The
+    # `win_file` module has its own `check_perms` function which would be
+    # used under a Windows system
+    filemod.check_perms = win_file.check_perms
+
 SED_CONTENT = '''test
 some
 content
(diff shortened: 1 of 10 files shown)


dwoz (approved):

cmcmarrow (approved):

Akm0d (approved):

tests/unit/modules/test_file.py
```diff
@@ -338,29 +342,53 @@ def test_uncomment(self):
...
+    @classmethod
+    def setUpClass(self):
+        if salt.utils.platform.is_windows():
+            self.original_check_perms = filemod.check_perms
+            filemod.check_perms = win_file.check_perms
+
+    @classmethod
+    def tearDownClass(self):
+        if salt.utils.platform.is_windows():
+            filemod.check_perms = self.original_check_perms
```
Akm0d: Does overriding filemod.check_perms then restoring it from the class actually work?

twangboy: Who knows...

PR #53473, opened 2019-06-13
address elasticsearch lib returning a dict instead of a string

### LINKED ISSUE #53472: Elasticsearch `get_load` throws warning trying to return JSON
This issue has shown up in a previous issue in the repo, but it appears that sometime along the line the `elasticsearch` Python API changed to return dictionaries instead of the current behavior which is to return a dictionary of the job cache.

_Originally posted by @msteed in https://github.com/saltstack/salt/issues/23125#issuecomment-140512958_

```
$ salt --versions-report
Salt Version:
           Salt: 2019.2.0

Dependency Versions:
           cffi: Not Installed
       cherrypy: Not Installed
       dateutil: 2.4.2
      docker-py: Not Installed
          gitdb: 0.6.4
      gitpython: 1.0.1
          ioflo: Not Installed
         Jinja2: 2.8
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: 1.0.3
   msgpack-pure: Not Installed
 msgpack-python: 0.4.6
   mysql-python: Not Installed
      pycparser: Not Installed
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: Not Installed
         Python: 2.7.12 (default, Nov 12 2018, 14:36:49)
   python-gnupg: 0.3.8
         PyYAML: 3.11
          PyZMQ: 15.2.0
           RAET: Not Installed
          smmap: 0.9.0
        timelib: Not Installed
        Tornado: 4.2.1
            ZMQ: 4.1.4

System Versions:
           dist: Ubuntu 16.04 xenial
         locale: UTF-8
        machine: x86_64
        release: 4.4.0-148-generic
         system: Linux
        version: Ubuntu 16.04 xenial
```

```
$ pip freeze | grep elastic
elasticsearch==6.4.0
```

```
$ salt salt elasticsearch.info
salt:
    ----------
    cluster_name:
        elasticsearch_calvin
    cluster_uuid:
        GjYi_ICTQJGolj3Lo42DQg
    name:
        Yv7_FY8
    tagline:
        You Know, for Search
    version:
        ----------
        build_date:
            2019-05-15T20:06:13.172855Z
        build_flavor:
            oss
        build_hash:
            65b6179
        build_snapshot:
            False
        build_type:
            tar
        lucene_version:
            7.7.0
        minimum_index_compatibility_version:
            5.0.0
        minimum_wire_compatibility_version:
            5.6.0
        number:
            6.8.0
```

calvinhp:
### What does this PR do?

 Fixes issue where elasticsearch lib returns a dict instead of a string that needs to be loaded via a JSON library.

Fixes #53472

### Tests written?

Yes

### Commits signed with GPG?

No


Akm0d (changes requested):
We need a testcase for this before we can merge.   I suggest adding it [here](https://github.com/saltstack/salt/blob/2018.3/tests/unit/states/test_elasticsearch.py) and can help if needed.

calvinhp:
> We need a testcase for this before we can merge. I suggest adding it [here](https://github.com/saltstack/salt/blob/2018.3/tests/unit/states/test_elasticsearch.py) and can help if needed.

I don't feel like that is the correct spot for this test case. This isn't an issue with the elasticsearch module or state. This is an issue when using the elasticsearch returner as a job cache. I looked through the code and can't see any tests that exercise the `get_load()` interface of the job cache. Do you have any pointers?  Seems like this should exist, but is out of scope for my small fix.

salt/returners/elasticsearch_return.py
```diff
@@ -405,6 +405,8 @@ def get_load(jid):
     data = __salt__['elasticsearch.document_get'](index=index,
                                                   id=jid,
                                                   doc_type=doc_type)
+    if isinstance(data, dict):
+        return data
```
waynew: I would actually simply rewrite this function as:

```
try:
    return salt.utils.json.loads(data)
except SomeError:
    return data or {}
```

waynew (changes requested):
Hey, thanks for the PR! @calvinhp this should actually be fairly easy (and worthwhile) to write a unit test for. You can simply follow this pattern found here
https://github.com/saltstack/salt/blob/2018.3/tests/unit/states/test_elasticsearch.py#L90

I would test:

- returning `None`
- returning an empty string
- returning json data (i.e. a string containing valid JSON)
- returning dictionary data

I'm pretty sure that we should trust that we're not going to get invalid JSON from `elasticsearch.document_get`. If that's a possibility, though, then how the function is written is definitely fine as written, and we should test that in that case we get an exception.

salt/returners/elasticsearch_return.py
calvinhp: Following back up, I have done more digging in the code and the documentation says that SaltStack supports Elasticsearch 1.5+ and the code for supporting the `master_job_cache` in Elastic landed in Salt 2015.8.1. The `salt.modules.elasticsearch` at that time (1) combined with the version of the python library (2) for even that old of an Elastic instance would still have returned at `dict()`. I think the code above makes the correct code path seem exceptional and we should remove handling of strings all together since this appears to have never worked.

Any error handling should be done at the salt module by ensuring that `__salt__['elasticsearch.document_get']()` should always return a `dict()`.  Plus this is easier to ensure with a test case.

So I think the `get_load()` method should just `return data or {}`. 

#### References
1. https://github.com/saltstack/salt/blob/2015.8/salt/modules/elasticsearch.py#L249
2. https://github.com/elastic/elasticsearch-py/blob/1.5.0/elasticsearch/transport.py#L331

calvinhp: > Hey, thanks for the PR! @calvinhp this should actually be fairly easy (and worthwhile) to write a unit test for. You can simply follow this pattern found here
> https://github.com/saltstack/salt/blob/2018.3/tests/unit/states/test_elasticsearch.py#L90

Since the `get_load()` method is using the `salt.modules.elasticsearch.document_get()` method. I think we should be ensuring that this method always returns a `dict()`. This has tests:

https://github.com/saltstack/salt/blob/develop/tests/unit/modules/test_elasticsearch.py#L676

Do you have a suggestion for what I should add that would ensure the job cache methods work?

waynew:
🤦‍♂️Oh dear.

Yeah, it doesn't look like this function should actually be doing anything. If it *should* be falling back to returning an empty `{}` that's what the test should be doing. Otherwise I'd just change this function to `return ...` and call it good.

Do you grok if it should be returning an empty dict always, over being able to return None, like it looks like the underlying code can?

calvinhp:
> Do you grok if it should be returning an empty dict always, over being able to return None, like it looks like the underlying code can?

I don't believe it will ever return `None`, the elastic search transport will return a `dict()` of somekind even if it is a 404.

If I look at the `returners.local_cache` as an example, the `get_load()` method is expected to return an empty `dict()` if there is no job found in the cache.

So that does indicate that if we get a 404, we need to return an empty `dict()` vs a `dict()` that contains the body from the 404 response.  Based on that, I'd modify the `get_load()` method to handle this case.  Sound good?

waynew:
Sounds and looks good - we definitely want some unit tests thrown in here, though, so we can capture *this* discussion in the code/tests.

calvinhp:
> Sounds and looks good - we definitely want some unit tests thrown in here, though, so we can capture _this_ discussion in the code/tests.

Sounds good, but it would help if there were any tests at all for the job cache stuff so I could make sure the elasticsearch ones matched the expected results.  What do you recommend here?

waynew:
If you're interested in just getting this fix in, it would be totally acceptable to just write what, the two tests to verify that if the dict contains `found` that we return it, otherwise, an empty dict?

If you want to improve things even further though, great! ✨ 🍰 🚀 

I'm not familiar with the part that you're talking about though - are there specific functions/functionality you are thinking about testing? What are they?

calvinhp:
Any feedback on the tests?

waynew:
Assuming that the return values are the kind we can expect, 👍 from me. There are a couple of test failures (py2-ubuntu-1604, py2- and py3-windows-2016, py3-centos-7, py3-fedora-29, and py3-ubunut-1804) I'll take a look and see if there's anything useful in Jenkins there

calvinhp:
I double checked a few of them and they appeared to be transient errors not caused by this code.

waynew:
re-run all

calvinhp:
I'm not sure what the next step is here?  The windows failure that is required also appears to be transient.  Any way we can get this reviewed?

calvinhp:
Ping!

waynew:
From a review perspective, this is 🆗 as far as I can tell. We've got some test/build infrastructure tech debt that we're working hard to pay down. Right now we've got a priority on getting 2019.2.1 released and out the door, we've just gotta go slow to go fast.

Unfortunately that means issues like this that are targeting 2018.3 are not going to get the attention we'd like.

But this code looks good, and we'll make sure to ping you if, once our test infrastructure is a-OK, we need more changes. We appreciate your patience! 🙂

twangboy:
We are not releasing 2018.3... we'll need to get this rebased against master.

twangboy (changes requested):
Please rebase this against the master branch and add a changelog

dwoz:
Closing due to lack of activity. Anyone should feel free to re-open or create a new PR if they want to see this through to the end in the next release cycle.

Issue #52899, opened 2019-05-07
integration.shell.test_enabled.EnabledTest.test_shell_default_enabled - '0\nsaltines' != '3\nsaltines'


waynew:
[...]

Issue #52898, opened 2019-05-07
integration.shell.test_enabled.EnabledTest.test_shell_disabled


waynew:
[...]

PR #53476, opened 2019-06-13
Fix for integration shell tests

xeacott:
test_shell_disabled and test_shell_default_enabled are now behaving
properly with updated test command.

### What does this PR do?
Uses echo instead of printf based on docs for wc l flag

#52899 
#52898 

### Previous Behavior
test_shell_disabled and test_shell_default_enabled would fail as wc had different behavior when using
the l flag

### New Behavior
now these tests pass using echo and newline character

### Tests written?

Yes

### Commits signed with GPG?

No


tests/integration/shell/test_enabled.py
```diff
@@ -20,7 +20,7 @@ class EnabledTest(ModuleCase):
     validate the use of shell processing for cmd.run on the salt command line
     and in templating
     '''
-    cmd = ("printf '%s\n' first second third | wc -l ; "
+    cmd = ("echo -e 'first\\nsecond\\nthird' | wc -l ; "
```
twangboy: Should it be `\\n` as you have it here or `\n`?

xeacott: It should remain as is since echo is being used instead of printf and they handle EOL differently

cmcmarrow (approved):

xeacott:
re-run docs

xeacott:
re-run full all

Akm0d (approved):

sagetherage:
@xeacott is this still valid to merge into develop or should we be putting this into master or maybe we already did?

PR #53482, opened 2019-06-14
[2017.7.9] Workaround nox's install only flag

s0undt3ch:
### What does this PR do?
To get some information from the system, which we then use to choose the appropriate static requirements file, we need to run some commands, something that nox will refuse to do if `--install-only` is passed.

We work around it by manually patching the value of`session._runner.global_config.install_only` for the commands that we MUST run, and only those and then we set it back to the value it had before.

For additional information about why we have to do this, please see:
   https://github.com/theacodes/nox/pull/181

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -57,15 +57,22 @@ def _get_session_python_version_info(session):
     try:
         version_info = session._runner._real_python_version_info
     except AttributeError:
-        session_py_version = session.run(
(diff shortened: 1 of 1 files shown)


waynew (approved):

Akm0d (approved):

s0undt3ch:
Approved and tests pass, merging.

Issue #53488, opened 2019-06-14
unit.utils.test_pycrypto.PycryptoTestCase.test_gen_hash failing


Ch3LL:
The test `unit.utils.test_pycrypto.PycryptoTestCase.test_gen_hash` is failing on macosx

https://jenkinsci.saltstack.com/job/2019.2.1/view/Python3/job/salt-mac-highsierra-py3/2/testReport/junit/unit.utils.test_pycrypto/PycryptoTestCase/test_gen_hash/

```
Traceback (most recent call last):
  File "/tmp/kitchen/testing/tests/unit/utils/test_pycrypto.py", line 30, in test_gen_hash
    self.assertTrue(ret.startswith('$6$'))
AssertionError: False is not true
```

PR #53491, opened 2019-06-14
Update test_gen_hash for macosx

Ch3LL:
### What does this PR do?
This test `unit.utils.test_pycrypto.PycryptoTestCase.test_gen_hash` is failing on macosx with this failure:

```
         File "/private/tmp/kitchen/testing/tests/unit/utils/test_pycrypto.py", line 30, in test_gen_hash
           self.assertTrue(ret.startswith('$6$'))
       AssertionError: False is not true
```

This is because crypt on macosx doesn't support the `$id$` format. A great explanation is shared here: https://bugs.python.org/issue22432

https://github.com/saltstack/salt/issues/53488

### Previous Behavior
`unit.utils.test_pycrypto.PycryptoTestCase.test_gen_hash` test fails on macosx

### New Behavior
`unit.utils.test_pycrypto.PycryptoTestCase.test_gen_hash` passes on macosx

### Tests written?
fixes test

### Commits signed with GPG?

Yes

### DIFF
--- a/tests/unit/utils/test_pycrypto.py
+++ b/tests/unit/utils/test_pycrypto.py
@@ -26,14 +26,17 @@ class PycryptoTestCase(TestCase):
         Test gen_hash
         '''
         passwd = 'test_password'
+        id = '$'
+        if salt.utils.platform.is_darwin():
+            id = ''
         ret = salt.utils.pycrypto.gen_hash(password=passwd)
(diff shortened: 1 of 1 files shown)


cmcmarrow (approved):

garethgreenaway (approved):

s0undt3ch (approved):

waynew:
Nice find!

PR #53495, opened 2019-06-14
Backport 53492 - Fix syndic connection when using tcp transport

dwoz:
### What does this PR do?

Fix syndic connection when using tcp transport

### Tests written?

No - Fixing issue in existing test suite

### Commits signed with GPG?

Yes

### DIFF
--- a/tests/integration/files/log_handlers/runtests_log_handler.py
+++ b/tests/integration/files/log_handlers/runtests_log_handler.py
@@ -34,6 +34,8 @@ __virtualname__ = 'runtests_log_handler'
 def __virtual__():
     if 'runtests_log_port' not in __opts__:
         return False, "'runtests_log_port' not in options"
+    if six.PY3:
+        return False, "runtests external logging handler is temporarily disabled for Python 3 tests"
     return True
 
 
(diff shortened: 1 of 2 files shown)


tests/integration/files/log_handlers/runtests_log_handler.py
```diff
@@ -35,8 +34,6 @@
 def __virtual__():
     if 'runtests_log_port' not in __opts__:
         return False, "'runtests_log_port' not in options"
-    if six.PY3:
```
Ch3LL: can we revert this for now since we showed it hanging tests on the 2019.2.1 branch?

Ch3LL (changes requested):

tests/integration/files/log_handlers/runtests_log_handler.py
dwoz: Done.

tests/integration/files/log_handlers/runtests_log_handler.py
```diff
@@ -23,7 +23,6 @@
 import msgpack
 
 # Import Salt libs
-from salt.ext import six
```
Ch3LL: just need to re-add this import back in

Ch3LL (changes requested):

dwoz:
This PR is no longer needed

PR #53503, opened 2019-06-17
fixes inconsistent return between local and local_async client

GwiYeong:
### What does this PR do?
fixes inconsistent return between `local` and `local_async` client in salt-api when the empty data returned after published

salt-api have two client for running job `local` and `local_async`. Actually they call same function `local_client.run_job_async`. but handling return data from `self.saltclient['local']` and `self.saltclient['local_async']` is different when `pub_data` is empty.

`local` client return `No minions matched the target. No command was sent, no jid was assigned.` message. `local_async` client return empty. 

so i added code for the same return at `local_async` for consistency of salt-api return

### Previous Behavior
if `jid` not exist in pub_data, return empty data when using `local_async` client

### New Behavior
if `jid` not exist in pub_data, return `No minions matched the target. No command was sent, no jid was assigned.` message when using `local_async` client (same as `local` client)

### Tests written?
No

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/netapi/rest_tornado/saltnado.py
+++ b/salt/netapi/rest_tornado/saltnado.py
@@ -1114,7 +1114,10 @@ class SaltAPIHandler(BaseSaltAPIHandler):  # pylint: disable=W0223
         # fire a job off
         pub_data = yield self.saltclients['local_async'](*f_call.get('args', ()), **f_call.get('kwargs', {}))
 
-        raise tornado.gen.Return(pub_data)
+        if 'jid' not in pub_data:
+            raise tornado.gen.Return('No minions matched the target. No command was sent, no jid was assigned.')
+        else:
+            raise tornado.gen.Return(pub_data)
 
     @tornado.gen.coroutine
     def _disbatch_runner(self, chunk):


waynew:
Hi, thanks for the PR!

Looks like there are actually tests which specify this behavior that are failing with this change. Are you sure that we want a consistent return here? I'm not super familiar with this part of Salt - it *does* make more sense to me that they should have a unified return, but I'm also not sure if there's some alternate explanation.

Do you know if there's a good explanation, or does it just look like an oversight?

GwiYeong:
@waynew 
i didn't check that there is a test for this case before write the PR. my bad. i will fix test case after making decision to approve this PR

i don't have alternate explanation. i also think that salt-api returning unified data is make more sense. 
furthermore consistent return is good for using api. more easy, not confusion. 
there is no worldwide standard rule for writing api i think. there is only exist design pattern or guideline for writing api based on people's know-how, but this can not be applied to all projects. 

so... this PR is actually my opinion not bug fix. you can reject this PR if it is not fit for saltstack.
thanks

waynew:
Hey @GwiYeong, thanks again for this PR! I brought it up with the team and we really like unifying the API

**but**

There are many downstream users that will expect the existing API. They would be very sad if we pulled the rug out from under them. We have some [documentation on deprecating functionality](https://docs.saltstack.com/en/latest/topics/development/deprecations.html) in Salt that we would definitely need to follow here. If you're up for modifying the code to follow that pattern, I think it would be great to have these APIs unified!

GwiYeong:
@waynew Ok. i will write the code to follow that pattern. 
i read document carefully but i could not find how to know current release name like `Aluminum` (may be i missed something because my english is not good). if i write warning for this change, i need that.
can you guide for me? 

And i will send new PR to develop branch after close this PR according to the document `Pull requests containing feature additions or non-bug-fix changes should be made against the develop branch.`. can i do it like this?

thanks in advance

waynew:
@GwiYeong 

> i read document carefully but i could not find how to know current release name like Aluminum (may be i missed something because my english is not good).

No problem - the current release, 2019.2, is Fluorine

> can you guide how to know release name for me?

The next release is Neon, then Sodium, followed by Magnesium. We've already hit our feature freeze for Neon, and it's possible that we'll get it released anyway before this is ready with tests. That means that you should plan to introduce the DeprecationWarning in Sodium, it will also be displayed in Magnesium, and then in the Aluminum release this change will go live.

> And i will send new PR to develop branch after close this PR according to the document Pull requests containing feature additions or non-bug-fix changes should be made against the develop branch.. can i do it like this?

Yeah, I think that's the right way to do it! I'll go ahead and close this one for you, feel free to tag me on your upcoming PR. Looking forward to it!

PR #53510, opened 2019-06-18
Allow management of AWS Route53 Health Checks to boto3_route53 state and execution modules.

pruiz:
### What does this PR do?

This PR extendes current boto3_route53 state/execution modules with required methods required in order to be able to handle/manage AWS Route53's Health Checks with salt.

This commit adds the following methods:
  - states/boto3_route53.py:
    - health_check_absent
    - health_check_present
  - modules/boto3/route53.py:
    - delete_health_check_by_name
    - delete_health_check
    - update_health_check_by_name
    - update_health_check
    - create_health_check
    - get_health_check_version_by_name
    - get_health_check_version
    - find_health_check
    - find_health_checks
    - get_health_checks_by_name
    - get_health_check_by_name
    - get_health_check
    - list_health_checks
    - list_tags_for_resource
    - get_health_check_id_by_name
    - get_health_check_ids_by_name

N/A

### Tests written?

I've tried to add some tests, but I've been unable to find any existing test against boto3_route53 to take as reference.

### Commits signed with GPG?

No


pruiz:
Hi, looks like 'continuous-integration/jenkins/pr-head' task has some issues with my second family name (García) due encoding issues with the diacritical í.

However I dont know where it is taking my name from (I guess from the commit message)?

waynew:
@pruiz Oops! Looks like we'll have to fix that. Side note: I will be very glad when everything is Python3 :trollface: 

As far as testing goes, have you looked at `tests/unit/modules/test_boto_route53.py`? It does look like we're missing a `test_boto3_route53.py`. The existing boto_route53 test might be a reasonable start, thought I can't tell if that's *actually* a `unit` test, or if it's a functional or integration test :suspect:

salt/modules/boto3_route53.py
```diff
@@ -1000,3 +1003,448 @@ def change_resource_record_sets(HostedZoneId=None, Name=None,
...
+            conn.change_tags_for_resource(
+                ResourceType='healthcheck',
+                ResourceId=result['Id'],
+                AddTags=[{ 'Key': 'Name', 'Value': Name }],
+                )
+            return {'result': result }
+        except ClientError as e:
+            if tries and e.response.get('Error', {}).get('Code') == 'Throttling':
+                log.debug('Throttled by AWS API.')
+                time.sleep(3)
```
waynew: Should we do an exponential/increasing backoff here? Or is 3 seconds the right way to go?

salt/modules/boto3_route53.py
```diff
@@ -1000,3 +1003,448 @@ def change_resource_record_sets(HostedZoneId=None, Name=None,
...
+            raise e
+
+    return result
+
+def update_health_check(Id, changes, region=None, key=None, keyid=None, profile=None, tries=20):
+    '''
+    Updates a health check with with passed changes request.
+
+    Name
+        The name to assign to newly created health check.
```
waynew: wrong parameter(s) here - should be id and changes

salt/modules/boto3_route53.py
```diff
@@ -1000,3 +1003,448 @@ def change_resource_record_sets(HostedZoneId=None, Name=None,
...
+                log.debug('Throttled by AWS API.')
+                time.sleep(3)
+                tries -= 1
+                continue
+            log.error('Failed to create healthcheck %s: %s', Name, six.text_type(e))
+            raise e
+
+    return result
+
+def update_health_check(Id, changes, region=None, key=None, keyid=None, profile=None, tries=20):
```
waynew: I think we should prefer `id` here, too

salt/modules/boto3_route53.py
```diff
@@ -1000,3 +1003,448 @@ def change_resource_record_sets(HostedZoneId=None, Name=None,
...
+    args = changes or {}
+    args.update({'HealthCheckId': Id})
+
+    while tries:
+        try:
+            return {'result': conn.update_health_check(**args)}
+        except ClientError as e:
+            if tries and e.response.get('Error', {}).get('Code') == 'Throttling':
+                log.debug('Throttled by AWS API.')
+                time.sleep(3)
```
waynew: same question here - should we prefer 3 seconds or increasing backoff?

salt/modules/boto3_route53.py
```diff
@@ -1000,3 +1003,448 @@ def change_resource_record_sets(HostedZoneId=None, Name=None,
...
+                tries -= 1
+                continue
+            log.error('Failed to update healthcheck %s: %s', Id, six.text_type(e))
+            raise e
+
+    return False
+
+def update_health_check_by_name(Name, changes, region=None, key=None, keyid=None, profile=None, tries=20):
```
waynew: Should we prefer `name` here?

salt/modules/boto3_route53.py
```diff
@@ -1000,3 +1003,448 @@ def change_resource_record_sets(HostedZoneId=None, Name=None,
...
+        raise SaltInvocationError('HealthCheck with name {0} not found.'.format(Name))
+    elif len(ids) > 1:
+        raise SaltInvocationError(
+            'Request matched more than one HealthCheck (%s). Refine your '
+            'criteria and try again.'.format(ids)
+        )
+
+    return update_health_check(ids[0], changes, region=region, key=key, keyid=keyid, profile=profile, tries=tries)
+
+def delete_health_check(Id, region=None, key=None, keyid=None, profile=None):
```
waynew: prefer `id` here

salt/modules/boto3_route53.py
```diff
@@ -1000,3 +1003,448 @@ def change_resource_record_sets(HostedZoneId=None, Name=None,
...
+    '''
+    Delete a Route53 health check.
+
+    CLI Example::
+
+        salt myminion boto3_route53.health_check 2405309f-2aa1-4add-9e8c-c47546a199d7
```
waynew: Docstring needs updating parameters and example

salt/modules/boto3_route53.py
```diff
@@ -1000,3 +1003,448 @@ def change_resource_record_sets(HostedZoneId=None, Name=None,
...
+    '''
+    conn = _get_conn(region=region, key=key, keyid=keyid, profile=profile)
+    try:
+        r = conn.delete_health_check(HealthCheckId=Id)
+        return {'result': True}
+    except ClientError as e:
+        log.error('Failed to delete health check %s: %s', Id, e)
+    return False
```
waynew: Should this be `{'result': False}`?

salt/modules/boto3_route53.py
```diff
@@ -1000,3 +1003,448 @@ def change_resource_record_sets(HostedZoneId=None, Name=None,
...
+    CLI Example::
+
+        salt myminion boto3_route53.health_check 2405309f-2aa1-4add-9e8c-c47546a199d7
+    '''
+    conn = _get_conn(region=region, key=key, keyid=keyid, profile=profile)
+    try:
+        r = conn.delete_health_check(HealthCheckId=Id)
+        return {'result': True}
+    except ClientError as e:
+        log.error('Failed to delete health check %s: %s', Id, e)
```
waynew: I think we should prefer `%r` for error types

salt/modules/boto3_route53.py
```diff
@@ -1000,3 +1003,448 @@ def change_resource_record_sets(HostedZoneId=None, Name=None,
...
+    return False
+
+def delete_health_check_by_name(Name, region=None, key=None, keyid=None, profile=None):
```
waynew: I think we should prefer `name` here.

salt/modules/boto3_route53.py
```diff
@@ -1000,3 +1003,448 @@ def change_resource_record_sets(HostedZoneId=None, Name=None,
...
+    '''
+    Delete a Route53 health check (by name).
+
+    CLI Example::
+
+        salt myminion boto3_route53.health_check_by_name A-HEALTH-CHECK
```
waynew: Docstring needs more example, and is missing parameters

salt/modules/boto3_route53.py
```diff
@@ -1000,3 +1003,448 @@ def change_resource_record_sets(HostedZoneId=None, Name=None,
...
+    '''
+    ids = get_health_check_ids_by_name(Name, region=region, key=key, keyid=keyid, profile=profile) or []
+
+    if len(ids) == 0:
```
waynew:
```suggestion
    if not ids:
```

salt/states/boto3_route53.py
```diff
@@ -83,6 +84,11 @@ def __virtual__():
     '''
     return 'boto3_route53' if 'boto3_route53.find_hosted_zone' in __salt__ else False
 
+def _ensure_unicode(obj):
```
waynew: I'm pretty sure the `salt.utils.stringutils.to_unicode` function properly handles this.

salt/states/boto3_route53.py
```diff
@@ -603,6 +620,15 @@ def rr_present(name, HostedZoneId=None, DomainName=None, PrivateZone=False, Name
                 fixed_rrs += [rr]
         ResourceRecords = [{'Value': rr} for rr in sorted(fixed_rrs)]
 
+    # If HostedZoneId is missing from AliasTarget, provide it on our own.
+    if AliasTarget:
+        if not 'HostedZoneId' in AliasTarget:
```
waynew: Would this be better as:

```
if AliasTarget and 'HostedZoneId' not in AliasTarget
```
?

salt/states/boto3_route53.py
```diff
@@ -603,6 +620,15 @@ def rr_present(name, HostedZoneId=None, DomainName=None, PrivateZone=False, Name
...
+            cleanid = HostedZoneId[12:] if HostedZoneId.startswith('/hostedzone/') else HostedZoneId
+            AliasTarget.update({'HostedZoneId': cleanid})
+
+    if HealthCheckName is not None:
```
waynew: Would this read better as just `if HealthCheckName:`? That will skip on `None`, but it will also skip `''` and any other false-y values.

salt/states/boto3_route53.py
```diff
@@ -798,3 +826,204 @@ def rr_absent(name, HostedZoneId=None, DomainName=None, PrivateZone=False,
...
+        Secret key to be used.
+
+    keyid
+        Access key to be used.
+
+    profile
+        Dict, or pillar key pointing to a dict, containing AWS region/key/keyid.
+    '''
+
+    Name = Name if Name else name
```
waynew:
```suggestion
Name = Name or name
```

Though... is it necessary to have both `Name` *and* `name`? I'm pretty sure that other states simply do something like:

```
foo:
   test.nop:
    - name: This is my name now!
```

That seems like it would make this function a bit easier to follow.

salt/states/boto3_route53.py
```diff
@@ -798,3 +826,204 @@ def rr_absent(name, HostedZoneId=None, DomainName=None, PrivateZone=False,
...
+    resettables = [ 'FullyQualifiedDomainName', 'Regions', 'ResourcePath', 'ChildHealthChecks' ]
+
+    checks = __salt__['boto3_route53.find_health_checks'](Name=Name, region=region, key=key, keyid=keyid, profile=profile)
+
+    if len(checks) > 1:
+        ret['comment'] = 'Given criteria matched more than one health check.'
+        log.error(ret['comment'])
+        ret['result'] = False
+        return ret
+    elif len(checks) == 0:
```
waynew: I might rename `checks` to `existing_checks`, then this could read:
```suggestion
    elif not existing_checks:
```

salt/states/boto3_route53.py
```diff
@@ -798,3 +826,204 @@ def rr_absent(name, HostedZoneId=None, DomainName=None, PrivateZone=False,
...
+        CallerReference = CallerReference if CallerReference else str(uuid.uuid4())  # future lint: disable=blacklisted-function
+    else:
+        check = checks[0].get('HealthCheckConfig')
+        for u in updatables:
+            if u in [ 'HealthCheckVersion' ]:
+                continue
+            if _ensure_unicode(params.get(u)) != _ensure_unicode(check.get(u)):
+                update = True
+                break
+            if ((ChildHealthChecks is not None) and (check.get('ChildHealthChecks') is not None)):
```
waynew: If `''` is not valid or irrelevant here, then I would rewrite this as
```suggestion
            if ChildHealthChecks and check.get('ChildHealthChecks'):
````

I'd do the same with the other checks below here

File: salt/states/boto3_route53.py [hunk 19]
```diff
@@ -798,3 +826,204 @@ def rr_absent(name, HostedZoneId=None, DomainName=None, PrivateZone=False,
...
+
+        items = {key: params.get(key) for key in keys if params.get(key) is not None}
+        args = {'Name': Name, 'region': region, 'key': key, 'keyid': keyid, 'profile': profile}
+        args.update(items)
+        log.debug(' ==> {}'.format(items))
+
+        if __opts__['test']:
+            ret['comment'] = 'Route 53 health check {} would be created.'.format(Name)
+            ret['changes'] = items
+            ret['result'] = None
```
waynew: I seem to recall another state where it was returning `None` for the result instead of `True` during a test run. I could be wrong about that part, perhaps it was just `changes`, but I do know that whatever it was doing led to an erroneous reporting (that is, instead of the test saying, yes, something would have changed, it just *said* there were changes but everything was already in the correct state).

If that doesn't seem correct, I can run it down and make sure that what I'm thinking of is correct.

salt/states/boto3_route53.py
```diff
@@ -798,3 +826,204 @@ def rr_absent(name, HostedZoneId=None, DomainName=None, PrivateZone=False,
...
+            ret['comment'] = 'Creation of Route 53 health check {} failed.'.format(Name)
+            log.error(ret['comment'])
+            ret['result'] = False
+            return ret
+    elif update:
+        keys = [
+            'HealthCheckVersion', 'IPAddress', 'Port', 'Type', 'ResourcePath', 'FullyQualifiedDomainName', 'SearchString',
+            'RequestInterval', 'FailureThreshold', 'MeasureLatency', 'Inverted', 'Disabled', 'HealthThreshold',
+            'EnableSNI', 'InsufficientDataHealthStatus', 'Regions', 'AlarmIdentifier', 'ChildHealthChecks', 'CallerReference'
+        ]
```
waynew: It looks like you've got this same keys collection multiple times - is there some better way to do this, maybe by creating a dictionary or using `**kwargs` or something?

salt/states/boto3_route53.py
```diff
@@ -798,3 +826,204 @@ def rr_absent(name, HostedZoneId=None, DomainName=None, PrivateZone=False,
...
+
+        check = checks[0].get('HealthCheckConfig')
+        args = {'region': region, 'key': key, 'keyid': keyid, 'profile': profile}
+        changes = {key: params.get(key) for key in keys if params.get(key) is not None}
+
+        if len(reset_elements) > 0:
```
waynew:
```suggestion
        if reset_elements:
```

salt/states/boto3_route53.py
```diff
@@ -798,3 +826,204 @@ def rr_absent(name, HostedZoneId=None, DomainName=None, PrivateZone=False,
...
+            changes.update({'ResetElements': reset_elements})
+
+        if __opts__['test']:
+            ret['comment'] = 'Route 53 health check {} would be updated.'.format(Name)
+            keys = [key for key in keys if changes.get(key) is not None]
+            ret['changes'] = {key: changes.get(key) for key in keys if _ensure_unicode(changes.get(key)) != _ensure_unicode(check.get(key))}
+            ret['result'] = None
```
waynew: same here - not sure if this should be `None` or `True`

salt/states/boto3_route53.py
```diff
@@ -798,3 +826,204 @@ def rr_absent(name, HostedZoneId=None, DomainName=None, PrivateZone=False,
...
+    return ret
+
+def health_check_absent(name, Name=None, region=None, key=None, keyid=None, profile=None):
+    '''
+    Ensure the Route53 Health Check described is absent
+
+    name
+        The name of the state definition.
+    Name
+        The name of the health check.
```
waynew: Same question here about `Name` vs `name`

salt/states/boto3_route53.py
```diff
@@ -798,3 +826,204 @@ def rr_absent(name, HostedZoneId=None, DomainName=None, PrivateZone=False,
...
+    args = {'Name': Name, 'region': region, 'key': key, 'keyid': keyid, 'profile': profile}
+
+    checks = __salt__['boto3_route53.find_health_checks'](Name=Name, region=region, key=key, keyid=keyid, profile=profile)
+
+    if len(checks) > 1:
+        ret['comment'] = 'Given criteria matched more than one health check.'
+        log.error(ret['comment'])
+        ret['result'] = False
+        return ret
+    elif len(checks) == 0:
```
waynew:
```suggestion
    elif not checks:
```

salt/states/boto3_route53.py
```diff
@@ -798,3 +826,204 @@ def rr_absent(name, HostedZoneId=None, DomainName=None, PrivateZone=False,
...
+        ret['comment'] = 'Route 53 health check {} already absent'.format(Name)
+        log.info(ret['comment'])
+        return ret
+
+    if __opts__['test']:
+        ret['comment'] = 'Route 53 health check {} would be deleted.'.format(Name)
+        ret['result'] = None
```
waynew: same question here on `None` vs `True`

Actually, I think it should have `ret['changes']['old'] = check and whatnot, too, like in the below block

waynew (changes requested):
Hey, thanks for the PR!

There were a bunch of minor improvements that could be made, and I also came across a few issues with docstring mismatches & whatnot.

Aside from the minor issues (and of course, getting tests), the only thing that I *think* could use some improvement is in the state where we've got this set of keys being defined multiple times. I forget where I've read it, but I'm pretty sure anytime you have a list of keys like that that's almost a guarantee that you should be using a dictionary instead.

Let me know if you have any questions or disagree with any of these points! I think these bits will be a valuable addition to salt!

pruiz:
@waynew thanks for the review, I'll try to take a look at it this weekend. And I will try to review what boto_route53 do and wether I can adapt it to this other module :)

salt/modules/boto3_route53.py
```diff
@@ -1000,3 +1003,448 @@ def change_resource_record_sets(HostedZoneId=None, Name=None,
...
+
+    if len(ids) > 1:
+        log.error(
+            'Request matched more than one HealthCheck (%s). Refine your '
+            'criteria and try again.', [z['Id'] for z in ret]
+        )
+
+    return ids[0] if len(ids) > 0 else None
+
+def list_tags_for_resource(Type, Id, region=None, key=None, keyid=None, profile=None):
```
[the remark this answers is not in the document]
pruiz: This was based on this other existing method on this same module:

```
def find_hosted_zone(Id=None, Name=None, PrivateZone=None,
                     region=None, key=None, keyid=None, profile=None):
```

However there 'Id' and 'Name' are optional arguments instead of positional ones. It looks like on boto3 modules, all parameters which have direct relation with AWS's XML API, are kept on PascalCase, so I followed this "convention".

salt/modules/boto3_route53.py
```diff
@@ -1000,3 +1003,448 @@ def change_resource_record_sets(HostedZoneId=None, Name=None,
                     (Name or HostedZoneId), six.text_type(e))
             raise e
     return False
+
+def get_health_check_ids_by_name(name, region=None, key=None, keyid=None, profile=None):
+    '''
+    Find Ids of all health checks with the given name tag.
```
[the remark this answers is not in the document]
pruiz: Oka

salt/modules/boto3_route53.py
```diff
@@ -1000,3 +1003,448 @@ def change_resource_record_sets(HostedZoneId=None, Name=None,
...
+    CLI Example:
+    .. code-block:: bash
+        salt myminion boto3_route53.list_health_checks \
+                profile='{"region": "us-east-1", "keyid": "A12345678AB", "key": "xblahblahblah"}'
+    '''
+    conn = _get_conn(region=region, key=key, keyid=keyid, profile=profile)
+    args = {'DelegationSetId': DelegationSetId} if DelegationSetId else {}
+    return _collect_results(conn.list_health_checks, 'HealthChecks', args)
+
+def get_health_check(Id, region=None, key=None, keyid=None, profile=None):
```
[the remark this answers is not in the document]
pruiz: See my previous comment on keeping parameters directly mapping with AWS's XML API CamelCase. If still the choice is to normalize it all I will go with it.

salt/states/boto3_route53.py
pruiz: On existing methods like hosted_zone_present, hosted_zone_absent, rr_present, rr_absent, etc. What I saw as pattern was setting result to None when test=true. So I replicated it on this new method.

salt/states/boto3_route53.py
pruiz: This is also something taking from exiting methods (rr_present, hosted_zone_absent, hosted_zone_present, etc)

salt/modules/boto3_route53.py
```diff
@@ -1000,3 +1003,448 @@ def change_resource_record_sets(HostedZoneId=None, Name=None,
...
+        log.error(
+            'Request matched more than one HealthCheck (%s). Refine your '
+            'criteria and try again.', [z['Id'] for z in ret]
+        )
+
+    return ids[0] if len(ids) > 0 else None
+
+def list_tags_for_resource(Type, Id, region=None, key=None, keyid=None, profile=None):
+    '''
+    Lists all tags attadhed to given resource.
```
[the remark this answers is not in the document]
pruiz: Fixed on latest commit.

salt/modules/boto3_route53.py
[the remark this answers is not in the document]
pruiz: Fixed on latest commit.

salt/states/boto3_route53.py
pruiz: Fixed on latest commit.

salt/states/boto3_route53.py
pruiz: Fixed on latest commit.

salt/modules/boto3_route53.py
pruiz: This is taken from already existing boto/boto3 modules.

salt/modules/boto3_route53.py
```diff
@@ -1000,3 +1003,448 @@ def change_resource_record_sets(HostedZoneId=None, Name=None,
...
+        'EnableSNI', 'InsufficientDataHealthStatus', 'Regions', 'AlarmIdentifier', 'ChildHealthChecks'
+    ]
+    params = locals()
+    result = None
+
+    conn = _get_conn(region=region, key=key, keyid=keyid, profile=profile)
+    CallerReference = CallerReference if CallerReference else str(uuid.uuid4())  # future lint: disable=blacklisted-function
+    HealthCheckConfig = {}
+
+    for key in [x for x in keys if params.get(x) is not None]:
```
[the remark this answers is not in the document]
pruiz: I got this pattern from another boto3 module, and I found it helps me keep track of the many changes (additions/removals/swaps) that we should make to HealthCheckConfig  in order to format it for aws submission.

salt/modules/boto3_route53.py
pruiz: Dunno, this is taken from another boto3 module. I am not keep on adding sleeps to code, but given it was under an exception handler, and I was not able to repro the actual issue, I simply relied on what was already there.

salt/states/boto3_route53.py
pruiz: This pattern was taken from boto_cloudtrail & boto_cloudwatch_event.

salt/states/boto3_route53.py
pruiz: I think in this case (having a case for > 1, another one for 0 plus a final else), the len() call on each leg makes it more readable.

salt/states/boto3_route53.py
pruiz: This code is quite tricky, as not all values are strings, nor should a 0 be treated like a none.

salt/states/boto3_route53.py
pruiz: This is taken from anothe boto module, but it looks like it is present everywhere:

$ grep -F "ret['result'] = None" salt/states/*|wc -l
     971

salt/states/boto3_route53.py
pruiz: I tried, but there are different sub-sets of keys which should be handled on each case, and after quite many errors (removing/adding the wrong key here and there), I ended up separating each key-set, as to make it easier to understand what is used/required on each case.

pruiz:
@waynew I've made quite a few changes from your review, added some unit-tests and replied to other comments. 
I've found moto to be somewhat limited on what it implements from AWS API, so I've left some unit testes disabled till they have enough coverage.

May you have a chance at reviewing this again?

Regards
Pablo

salt/modules/boto3_route53.py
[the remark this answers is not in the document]
waynew: That seems reasonable, if mildly (personally) distasteful 😂

salt/states/boto3_route53.py
waynew: think this is still valid about the `ret['changes']['old']`. `None` is correct for test mode, though.

salt/modules/boto3_route53.py
```diff
@@ -1000,3 +1003,448 @@ def change_resource_record_sets(HostedZoneId=None, Name=None,
...
+        raise SaltInvocationError('Exactly one of either Id or Name is required.')
+
+    if Id:
+        ret = [get_health_check(Id, region=region, key=key, keyid=keyid, profile=profile)]
+    else:
+        ret = get_health_checks_by_name(Name, region=region, key=key, keyid=keyid, profile=profile)
+
+    return ret
+
+def find_health_check(Id=None, Name=None, region=None, key=None, keyid=None, profile=None):
```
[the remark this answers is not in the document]
pruiz: This is another case of code based on a pattern already existing on this same module (ie. find_hosted_zone)

salt/modules/boto3_route53.py
pruiz: This is another case of code based on a pattern already existing on this same module (ie. find_hosted_zone)

salt/modules/boto3_route53.py
pruiz: This is another case of code based on a pattern already existing on this same module (ie. find_hosted_zone)

salt/modules/boto3_route53.py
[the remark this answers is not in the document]
waynew: I'm not entirely convinced (despite the pattern existing elsewhere - however that does bring up other questions).

Foolish consistency is a hobgoblin, but reasonable consistency is good.

But it feels like there must be a better way.

In this particular case there *are* no additions/removals/swaps - there's only one difference between taking `**kwargs` (or, `**HealthCheckConfig`) - with the `**` approach you *could* receive extra values. 🤔

salt/modules/boto3_route53.py
waynew: C'mon, that's easy, all you need to do is [create 200(and one?) active health checks (per account??)](https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/DNSLimitations.html#limits-api-entities-health-checks) :trollface:

I guess I'm OK with this. I'd prefer something else, like you, but hopefully it shouldn't happen very often. Maybe just add a comment to the code suggesting that we could update the way we do the backoffs to actually check for how long we need to wait, or just report the failure or something.

salt/modules/boto3_route53.py
waynew: See other comment here, I think just a comment suggesting an alternate approach, but for now this code is fine.

waynew:
@pruiz thanks for the updates! I think this is looking pretty good, but there are a couple of unresolved comments - some I think need actual changes, but others are just discussion points.

waynew:
@pruiz are you interested in picking this up again? It will need a rebase against `master` at this point.

pruiz:
I have scarce time right now, but maybe I can take a look at it again on a
few weeks.

waynew:
@pruiz how does your time look like now? Are you able to drive this PR to completion?

If you're not aware, we've blackened the Salt codebase - a good approach that I've found to avoid merge conflicts is to do a:

```
git checkout master
git checkout -b your-new-change
git checkout abcd1234 -- files/you/changed.py in/that/revision.py
git commit
```

And then just repeat that for each of the commits you're interested in. Finally just do a `git --force push your-remote your-new-change` You could also just do it with your final changes if you don't care about having all of the intermediate commits.

Feel free to reach out to me on IRC or community slack if you need some help!

waynew:
@pruiz bump - do you have some free time for this?

pruiz:
Not right now, I will probably have to review this during this summer, once
we have moved over to salt 3000.x

sagetherage:
Reviewed in Grooming 2021-MAY-03
@pruiz are you still looking into this and want to rebase to master branch instead?

sagetherage:
Reviewed in Grooming meeting 2021-MAY-03 - I will leave open for another week or so, but if no comments I am considering closing.

sagetherage:

Issue #53539, opened 2019-06-19
s3fs appears to double encode urls when fetching


benjumanji:
### Description of Issue
S3 urls when fetching for s3fs appear to be double encoded.  I have an item in a bucket containing a `+` symbol that is being rendered as `%252B`. Which is `+` -> `%2B` -> `%252B`.

### Setup
fileserver_backend:
  - s3fs

etc..

### Steps to Reproduce Issue
Have an item in a bucket with a +. Running `salt-run fileserver.update` should yield a log line somewhere like
```
[INFO    ] example - base : examplel/unstable/example_0.1.1-alpha.1+57dfd57_armhf.deb
[DEBUG   ] S3 Request: http://example.s3.amazonaws.com/example/unstable/example_0.1.1-alpha.1%252B57dfd57_armhf.deb?
```
The caching will blow up with a 404, which is not surprising. Note how the `+` has been encoded. Also note that salt got the listing correct, just not the fetching. 

### Versions Report
Salt Version:
           Salt: 2019.2.0

Dependency Versions:
           cffi: 1.6.0
       cherrypy: Not Installed
       dateutil: 1.5
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.7.2
        libgit2: 0.26.3
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.5.6
   mysql-python: Not Installed
      pycparser: 2.14
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: 0.26.4
         Python: 2.7.5 (default, Apr  9 2019, 14:30:50)
   python-gnupg: Not Installed
         PyYAML: 3.10
          PyZMQ: 15.3.0
           RAET: Not Installed
          smmap: Not Installed
        timelib: Not Installed
        Tornado: 4.2.1
            ZMQ: 4.1.4

System Versions:
           dist: centos 7.6.1810 Core
         locale: UTF-8
        machine: x86_64
        release: 3.10.0-957.1.3.el7.x86_64
         system: Linux
        version: CentOS Linux 7.6.1810 Core

benjumanji:
[...]

PR #53540, opened 2019-06-19
Don't double encode s3 paths

benjumanji:
As s3.query already knows how to quote paths, don't quote the passed in
path: s3.query will handle it.

### What does this PR do?
Prevents double urlencoding s3 paths in s3fs fetch code.

#53539
### Previous Behavior
Paths double encoded leading to 404s for legit bucket items

### New Behavior
Paths encoded just the once, items are located as one might expect

### Tests written?
Nope

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/fileserver/s3fs.py
+++ b/salt/fileserver/s3fs.py
@@ -98,7 +98,6 @@ import salt.utils.versions
 # pylint: disable=import-error,no-name-in-module,redefined-builtin
 from salt.ext import six
 from salt.ext.six.moves import filter
-from salt.ext.six.moves.urllib.parse import quote as _quote
 # pylint: enable=import-error,no-name-in-module,redefined-builtin
 
 log = logging.getLogger(__name__)
@@ -696,7 +695,7 @@ def _get_file_from_s3(metadata, saltenv, bucket_name, path, cached_file_path):
                         service_url=service_url,
                         verify_ssl=verify_ssl,
                         location=location,
-                        path=_quote(path),
+                        path=path,
                         local_file=cached_file_path,
                         full_headers=True,
                         path_style=path_style,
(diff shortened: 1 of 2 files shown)


benjumanji:
I'd be happy to do something with testing if someone could point me to the right test infrastructure. I've been using salt for about 3 days.

garethgreenaway (approved):

benjumanji:
What's the standard mo for out of date branches? is it rebase or merge?

benjumanji:
I already have patches in my provisioning for salt-master so I don't mind if this gets merged one way or the other but it would be cool if someone could point me to a similar test that I could copy. AFAICT there are _no_ extant tests for `s3fs` which makes it a little challenging for a drive-by contributor to add the test case that it seems I need to get this merged.

benjumanji:
@garethgreenaway is there any likelihood of this getting merged without further work?

benjumanji:
This is one of the weirder interactions I've had on github. This is a 3 line change, obviously fixes a bug, all status checks pass and was reviewed reviewed super promptly. I've had three contributors interact with it and yet I'm stuck in a cycle of checking once every few days, rebasing and generally talking to myself with no idea if it will ever be merged.

benjumanji:
Ok. I've learnt my lesson. Don't bother with upstream, it's a total waste of time.

garethgreenaway:
@benjumanji Apologies for the delay and thank you for the PR!  We definitely want to get this merged but need some tests associated with the change, I'll comment in a bit with some suggestions.

garethgreenaway:
@benjumanji I've added a test, to ensure that the path argument that is passed along to the `s3.query` function is the same one that is passed into `_get_file_from_s3` function.  One the tests pass, we can go ahead and merge this in.  Thanks for your patience.

dwoz (approved):

waynew:
@garethgreenaway / @benjumanji is there a way to test this and verify that a file with a space in the name works fine with this change?

I did some code archaeology and discovered that was the reason that `quote` was added in the first place. I'd like to be sure that this change doesn't affect that case, at least!

garethgreenaway:
@waynew Updated the test to include a test to ensure URLs with spaces are quoted properly.

waynew (approved):

Issue #53533, opened 2019-06-19
Salt cloud installation fails sporadically - minion sometimes fails to start on windows


Mapel88:
### Description of Issue
I'm running windows VM installation with salt-run (on XEN servers if t matters).
Salt minion is installed with setup.exe, then the setup supposed to stop then start the minion.
It seems that sometimes the service is not stopped in a timely fashion and therefore the start command fails.
Is there a way to add sleep for a few secs after stop command or maybe validate that the service actually stopped before trying to start it?

```
[INFO    ] Running command under pid 382: 'winexe -U 'Administrator%XXX-REDACTED-XXX' //10.0.0.1 "c:\salttemp\Salt-Minion-2018.3.4-Py3-AMD64-Setup.exe /S /master=[u'master1'] /minion-name=server11"'
[DEBUG   ] Executing command(PID 382): 'winexe -U 'Administrator%XXX-REDACTED-XXX' //10.0.0.1 "c:\salttemp\Salt-Minion-2018.3.4-Py3-AMD64-Setup.exe /S /master=[u'master1'] /minion-name=server11"'
[INFO    ] Running command under pid 449: 'winexe -U 'Administrator%XXX-REDACTED-XXX' //10.0.0.1 "sc stop salt-minion"'
[DEBUG   ] Executing command(PID 449): 'winexe -U 'Administrator%XXX-REDACTED-XXX' //10.0.0.1 "sc stop salt-minion"'
[DEBUG   ]
SERVICE_NAME: salt-minion
        TYPE               : 10  WIN32_OWN_PROCESS
        STATE              : 3  STOP_PENDING
                                (NOT_STOPPABLE, NOT_PAUSABLE, IGNORES_SHUTDOWN)
        WIN32_EXIT_CODE    : 0  (0x0)

SERVICE_NAME: salt-minion
        TYPE               : 10  WIN32_OWN_PROCESS
        STATE              : 3  STOP_PENDING
                                (NOT_STOPPABLE, NOT_PAUSABLE, IGNORES_SHUTDOWN)
        WIN32_EXIT_CODE    : 0  (0x0)
[DEBUG   ]    SERVICE_EXIT_CODE  : 0  (0x0)
        CHECKPOINT         : 0x1
        WAIT_HINT          : 0x7d0
        SERVICE_EXIT_CODE  : 0  (0x0)
        CHECKPOINT         : 0x1
        WAIT_HINT          : 0x7d0
[...]

PR #53547, opened 2019-06-20
Increase sleep time after stop minion command

Mapel88:
When installing windows with salt cloud, salt minion fails to start sporadically. 

### What does this PR do?
increase sleep time after stop minion command

bug #53533

### Previous Behavior
Sometimes salt minion takes longer than 5 seconds to stop and therefore the start command fails with error:
```
[DEBUG   ] [SC] StartService FAILED 1056:

An instance of the service is already running.
```

### New Behavior
Hopefully salt-minion will be stopped in 10 sec.

### Tests written?
No

### Commits signed with GPG?
No


xeacott:
If this fixes, I'm okay with this change. For the future, adding optional argument to pass in something larger than 10 seconds may be necessary.

xeacott:
re-run full all

xeacott (approved):

waynew:
Would it make more sense to do an increasing (exponential?) backoff while waiting? e.g. check, then wait 1, 2, 4, 8s, and then finally fail?

Akm0d (approved):

waynew (changes requested):
Hey @Mapel88, thanks for this PR! You may have noticed that we've been changing our development strategy to work off the `master` branch only, as well as requiring tests.

We also discussed this in our architecture meeting and came up with an alternate approach. The result of the discussion was that we should be able to check for the failures that you provided in the log, and there shouldn't be any harm in retrying the start command until it succeeds.

That is, the code should look something like this:

```
# stop minion here
started = False
while not started:
    try:
        # run start command
        started = True
    except SomeException as e:
        if '[SC] StartService FAILED 1056' in e.output:  # or, whatever
            time.sleep(1)  # one second should be plenty here
        else:
            six.reraise(e)


```

This is certainly not the *exact* code that we want, but it should be a helpful starting place. We want to test this too, namely:

- if the command run is successful, it doesn't get called twice
- if it's not successful, we keep trying until it is
- if a different error is raised, we appropriately re-rase it

Does that sound like something that you're interested in? If not, I'm sure we can figure out an alternate approach. Thanks!

It would also be good to have some tests confirming that it will continue to call the start command if the results contain

Mapel88:
@waynew I already changed it to a while loop to validate service stopped properly before starting it in another PR, but no progress there... Waiting for approval.
Please see PR #54911

Thanks

waynew:
@Mapel88 Looks like you'd probably want to close this PR, (and any other dupes) and edit that other PR to point against `master`. The same advice applies there, of course.

Mapel88:
Closing this PR. Newer fix available at PR #54911

PR #53566, opened 2019-06-21
Faster tests

waynew:
### What does this PR do?

Makes test go fast 💨 ⏩ 🚀 

During a some cursory checks, a test class typically went from around 60s to 1s. So, quite a bit of a speedup.

### Previous Behavior

Previously all integration tests went through the entire client -> master -> minion -> master -> client loop. We still will have several tests that go through that cycle, but we have *plenty* of tests for which that's irrelevant. As long as we still have some representative tests making that full cycle, there shouldn't be any problems.

### New Behavior

Now we use the MasterMinion for most of the integration tests. 

### Tests written?

Yes (updating the testrunner)

### Commits signed with GPG?

Yes

---

It's likely that running the full suite here will expose some failures for one reason or another. The tests should either get updated to only care about  what they need, or be explicit about needed the `rem`ote. I'll take care of them as they're reported.

### DIFF
--- a/tests/integration/minion/test_pillar.py
+++ b/tests/integration/minion/test_pillar.py
@@ -228,6 +228,7 @@ class BasePillarTest(ModuleCase):
     '''
     @classmethod
     def setUpClass(cls):
+        cls.remote = True
         os.makedirs(PILLAR_BASE)
         with salt.utils.files.fopen(TOP_SLS, 'w') as fp_:
             fp_.write(textwrap.dedent('''\
@@ -292,6 +293,7 @@ class DecryptGPGPillarTest(ModuleCase):
 
     @classmethod
     def setUpClass(cls):
+        cls.remote = True
         try:
             os.makedirs(GPG_HOMEDIR, mode=0o700)
         except Exception:
--- a/tests/integration/modules/test_cmdmod.py
+++ b/tests/integration/modules/test_cmdmod.py
@@ -5,7 +5,6 @@ from __future__ import absolute_import, print_function, unicode_literals
 import os
 import sys
 import tempfile
-import textwrap
 
 # Import Salt Testing libs
 from tests.support.case import ModuleCase
@@ -38,6 +37,8 @@ class CMDModuleTest(ModuleCase):
     '''
     Validate the cmd module
     '''
+    remote = salt.utils.platform.is_windows()
+
     def setUp(self):
         self.runas_usr = 'nobody'
         if salt.utils.platform.is_darwin():
@@ -70,7 +71,7 @@ class CMDModuleTest(ModuleCase):
         self.assertEqual(self.run_function('cmd.run',
                          ['echo {{grains.id}} | awk "{print $1}"'],
                          template='jinja',
-                         python_shell=True), 'minion')
+                         python_shell=True), 'master_minion')
         self.assertEqual(self.run_function('cmd.run',
                          ['grep f'],
                          stdin='one\ntwo\nthree\nfour\nfive\n'), 'four\nfive')
@@ -157,7 +158,7 @@ class CMDModuleTest(ModuleCase):
         cmd_blacklist_glob
         '''
         self.assertEqual(self.run_function('cmd.run',
-                ['bad_command --foo']).rstrip(),
+                ['bad_command --foo'], rem=True).rstrip(),
                 'ERROR: The shell command "bad_command --foo" is not permitted')
 
     def test_script(self):
@@ -174,7 +175,7 @@ class CMDModuleTest(ModuleCase):
         cmd.script_retcode
         '''
         script = 'salt://script.py'
-        ret = self.run_function('cmd.script_retcode', [script])
+        ret = self.run_function('cmd.script_retcode', [script], rem=True)
         self.assertEqual(ret, 0)
 
     def test_script_cwd(self):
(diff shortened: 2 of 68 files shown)


tests/integration/modules/test_cmdmod.py
```diff
@@ -70,7 +70,7 @@ def test_run(self):
         self.assertEqual(self.run_function('cmd.run',
                          ['echo {{grains.id}} | awk "{print $1}"'],
                          template='jinja',
-                         python_shell=True), 'minion')
+                         python_shell=True), 'master')
```
s0undt3ch: We shouldn't flip to `master`... We should handle the flip, if needed in the function being called?
Or maybe name it `master_minion`?

This way we know better what's being targeted?

tests/support/case.py
```diff
@@ -788,6 +790,17 @@ class ModuleCase(TestCase, SaltClientTestCaseMixin):
...
+    masterminion = None
+
+    def __init__(self, *args, **kwargs):
+        super(ModuleCase, self).__init__(*args, **kwargs)
+        if self.masterminion is None:
+            opts = self.get_config(
+                self._salt_client_config_file_name_,
+                from_scratch=True,
+            )
+            ModuleCase.masterminion = salt.minion.MasterMinion(opts)
```
s0undt3ch: Why not self?

tests/support/case.py
```diff
@@ -808,11 +821,32 @@ def minion_run(self, _function, *args, **kw):
...
         behavior of the raw function call
         '''
+
+        # This not rem and minion_tgt block changes how most integration
+        # tests are run. Instead of running client -> master -> minion -> os
+        # and then back again, it skips all of that by just running the
+        # function from the MasterMinion.  This should work in most cases,
+        # but for those that really do need the round trip, or only make
+        # sense when executed in the context of a minion, `rem=True` (where
+        # rem means remote), allows bypassing this. It incurs the cost of
```
s0undt3ch: And why not `remote` instead of `rem`?
Saves a read on the comment....

tests/support/case.py
s0undt3ch: I know we only instantiate the class object once, but we should either do something in `setUpClass` or... something.
Because we should delete the object once we're done with the test module class.

We'll also have to instantiate it several times...

I guess we can leave as is, and see if we get  object leaks...

waynew: I *wanted* to just make it a classlevel, to avoid creating it umptymillion times, but the `_salt_client_config_filename_` came from self. It might not really be important, though.

tests/support/case.py
waynew: 🤷‍♂ I don't know if I have an *strong* feelings. `rem` was just what Tom suggested while we were working :) I *do* kind of like it being terse and somewhat cryptic as an indication that you probably shouldn't be using it unless you understand why you're using it. But if there are strong feelings the other way, I'm happy to change it.

tests/integration/modules/test_cmdmod.py
waynew: I'm not sure I understand the question? It's not changing the target, it's just the MasterMinion's id grain is `master`, instead of `minion`. It may be that a different grain would be a better choice here :suspect: 

Do you have any recommendations?

s0undt3ch: `master_minion`?

tests/support/case.py
s0undt3ch: Yeah, I understand why you did it.
But we've had issues with not releasing objects before hence my worries.
But lets wait and see.

tests/support/case.py
s0undt3ch: I thought that name would have come from @thatch45 or a similar mind :)

He like short names, I like understanding the names as I read them.
Your cryptic argument sold me.
Leave it.

tests/integration/modules/test_cmdmod.py
waynew: Then we'd have to change the ID in the `master` config file as well. Is that something that we want to do?

waynew: i.e. this line:

https://github.com/saltstack/salt/blob/develop/tests/integration/files/conf/master#L3

s0undt3ch: We're copying the options from a loaded dictionary, we would change the `id` there.
During this process do we ever load options from the config file? If so, we might want to create a temporary one with the changed ID and point salt to that one...

waynew: Ah yeah, that would work.

For the latter, I'm not sure. I'll see about setting the ID before passing the opts in.

tests/integration/minion/test_pillar.py
```diff
@@ -196,6 +196,7 @@ class BasePillarTest(ModuleCase):
     '''
     @classmethod
     def setUpClass(cls):
+        self.remote = True
```
s0undt3ch: `cls.remote`

tests/integration/modules/test_cmdmod.py
waynew: Done 👍

tests/integration/minion/test_pillar.py
waynew: 🤦‍♂

cmcmarrow (approved):

PR #53598, opened 2019-06-25
Some more CI pipelines fixes to support old and new jenkins

s0undt3ch:
### What does this PR do?
See title.

### DIFF
--- a/.ci/docs
+++ b/.ci/docs
@@ -4,12 +4,6 @@ def salt_target_branch = '2017.7.9'
 
 properties([
     buildDiscarder(logRotator(artifactDaysToKeepStr: '', artifactNumToKeepStr: '', daysToKeepStr: '', numToKeepStr: '10')),
-    [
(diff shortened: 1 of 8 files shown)


.ci/docs
```diff
@@ -4,12 +4,6 @@ def salt_target_branch = '2017.7.9'
 
 properties([
     buildDiscarder(logRotator(artifactDaysToKeepStr: '', artifactNumToKeepStr: '', daysToKeepStr: '', numToKeepStr: '10')),
-    [
-        $class: 'ScannerJobProperty', doNotScan: false
-    ],
-    [
-        $class: 'RebuildSettings', autoRebuild: false, rebuildDisabled: false
```
bryceml: It works fine removing this line? are rebuilds allowed by default?

s0undt3ch: yes to both Q's

bryceml: cool, sounds good to me.

PR #53621, opened 2019-06-27
Git pillar update interval

sathieu:
### What does this PR do?

Add support for new config option `git_pillar_update_interval` so git_pillar data can be collected less frequently than the main `loop_interval` timer allows

This is based on work on #52274 by @Tanoti

### Previous Behavior

Git pillar remotes would be checked using the `loop_interval` timer value (e.g. every 60 seconds).

### New Behavior

Git pillar remotes can be checked using a different interval to the `loop_interval`. For example, if the git pillar data is not likely to change very often then a higher value such as 300 seconds could be used which would reduce the load on the git repo server by a factor of five.

### Tests written?

No

### Commits signed with GPG?

Yes


sathieu:
I don't know how to write tests for this...

sathieu:
@twangboy @dwoz @terminalmage @cachedout Please review.

DmitryKuzmenko (approved):

waynew:
I'm not sure if there's any good way of automating the test, but a meaningful test would be to update the interval to something ridiculously short (1s? 3s?) and simply verify that it triggers an update the appropriate number of times within an interval.

salt/pillar/git_pillar.py
```diff
@@ -350,6 +350,21 @@
         - __env__ https://mydomain.tld/pillar-appdata.git:
           - mountpoint: web/server/
 
+.. _git_pillar_update_interval:
+
+git_pillar_update_interval
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. versionadded:: 2020.2.0
```
dwoz: This should be `Neon`

dwoz (changes requested):

salt/pillar/git_pillar.py
sathieu: Done (also in `doc/ref/configuration/master.rst`), rebased and force-pushed.

sathieu:
@dwoz thanks!

PR #53631, opened 2019-06-27
proxy minion __virtual__ error cleanup

dhiltonp:
### What does this PR do?
proxy minions may not have os/os_family/kernel set.

Let's use `__grains__.get('os')` appropriately so errors aren't written to logs.

### DIFF
--- a/salt/modules/gentoolkitmod.py
+++ b/salt/modules/gentoolkitmod.py
@@ -24,7 +24,7 @@ def __virtual__():
     '''
     Only work on Gentoo systems with gentoolkit installed
     '''
-    if __grains__['os_family'] == 'Gentoo' and HAS_GENTOOLKIT:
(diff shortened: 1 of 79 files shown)


salt/modules/gentoolkitmod.py
```diff
@@ -24,7 +24,7 @@ def __virtual__():
     '''
     Only work on Gentoo systems with gentoolkit installed
     '''
-    if __grains__['os_family'] == 'Gentoo' and HAS_GENTOOLKIT:
+    if __grains__('os_family') == 'Gentoo' and HAS_GENTOOLKIT:
```
garethgreenaway: Should this include a get as well?

garethgreenaway (changes requested):

salt/modules/gentoolkitmod.py
dhiltonp: thanks for the catch!

dhiltonp:
merged #53632

PR #49476, opened 2018-08-31
OmniOS also supports pkgin

sjorge:
### What does this PR do?
As discovered in #49469 OmniOS is not loading the pkgin module while it is supported.
We remove SunOS and SmartOS from the OS list and enable pkgin for all members of the
Solaris family instead.

### What issues does this PR fix or reference?
This is a **partial** fix for #49469

### Previous Behavior
Only Sun/Oracle Solaris and SmartOS had pkgin support.

### New Behavior
The entire Solaris family has pkgin support, including OpenIndiana, OmniOS, ...

### Tests written?
No

### Commits signed with GPG?
No

PR #53632, opened 2019-06-27
pkgin is on Solaris...

dhiltonp:
### What does this PR do?
pkgin is the package manager if the os family is Solaris and pkgin is available.

Prior to this, if pkgin was installed, pkgin was set as the default package manager. The OS check was always true:

` __grains__.get('os') in ['NetBSD', 'DragonFly', 'Minix', 'Darwin', 'SmartOS'] or 'pkgin'`

#49476 - previous Solaris family list
#53631 - waiting for this PR to be accepted

### DIFF
--- a/salt/beacons/aix_account.py
+++ b/salt/beacons/aix_account.py
@@ -20,7 +20,7 @@ def __virtual__():
     '''
     Only load if kernel is AIX
     '''
-    if __grains__['kernel'] == ('AIX'):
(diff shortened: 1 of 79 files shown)


dhiltonp:
The only difference between this version and the one that tests passed for is the Gentoo get fix...

cmcmarrow (approved):

PR #53658, opened 2019-06-28
Added version argument to @depends decorator to enforce a minimum ver...

github-abcde:
...sion requirement.

### What does this PR do?
It adds a `version` kwarg to the @depends decorator, which can handle any value that is able to be parsed with `salt.utils.version.LooseVersion`. The value of this version kwarg will be compared to the `__version__`-attribute of the module mentioned in the decorator (if it cannot be found (because it is a command or the module does not have the attribute), it will silently default to `None`). If the version is not equal or larger than the determined version (noting that when the latter is `None`, everything will be either equal (also `None`) or greater (any value)), the specific function will be removed by `enforce_dependencies`.

None that I know of

### Previous Behavior
The @depends decorator can only specify whether or not a module needs to be loaded.

### New Behavior
The @depends decorator can also specify a specific minimum version for the module that needs to be loaded.

### Tests written?
No, there are tests in `tests/integration/modules/test_decorators.py`, but I have no idea how those work.

### Commits signed with GPG?
Yes


max-arnold:
Nice! Could this be merged to Neon?

github-abcde:
Hmm, apparently somewhere, a unicode character (the triple-dots) was introduced into the branch name, and that is causing several tests to fail. Is this something I can fix on my end (and how)?

github-abcde:
Ok, that seems to have fixed the failing builds.
@max-arnold What do I need to do to merge this to Neon?

max-arnold:
Ask someone from SaltStack if that is possible, and if they agree then rebase against neon branch.

twangboy (approved):

github-abcde:
Closed in favor of https://github.com/saltstack/salt/pull/55590

PR #53665, opened 2019-06-29
[WIP] added reg.imported state module function

H20-17:
### What does this PR do?
Adds the state function, _reg.imported_file_, which is supposed to be the stateful correlate of the existing executable module function, _reg.import_file_

This does not fix an existing PR (AFAIK)

### Tests written?

No

### Commits signed with GPG?

No

### DIFF
--- a/salt/states/reg.py
+++ b/salt/states/reg.py
@@ -812,32 +812,33 @@ def _imported_file_do_import(reference_reg_file,use_32bit_registry):
     the module function reg.import_file. If that function then throws
     exceptions, this function will catch some of them and generate
     appropriate messages to use as the comment of the
-    ret dictionary. The function returns an ordered triplet (a,b,c).
+    ret dictionary. The function returns an ordered triplet (a,b).
     The values are defined as follows
-       a: Whether or not the operation succeeded expressed as a boolean. If in test, mode, just None.
-       b: a comment. This will just be None if not in test mode and the operation succeeds,
-          because we will determine a comment later on.
-       c: a changes dictionary. This will just be None if not in test mode and the operation succeeds,
-          because we will create a changes dictionary later on. If the operation fails, this
-          will be an empty dictionary.
+       a: Whether or not the operation succeeded expressed as a boolean,
+          or None if in test mode.
+       b: If not in test mode and the operation succeeds this will have the value of None.
+          Otherwise this will be an ordered pair consisting of a comment and a 
+          suitably populated changes directory (empty if the operation failed).
     '''
     if __opts__['test'] == True:
         comment = "Changes required. Import will proceed."
         changes = {}
         changes['old'] = 'Registry unmodified'
         changes['new'] = 'Registry modified by importing reg file.'
-        return (None, comment, changes)
+        return (None, (comment, changes))
     try:
         __salt__['reg.import_file'](reference_reg_file,use_32bit_registry)
     except ValueError as err:
         comment_fmt = "Call to module function 'reg.import_file' has failed. Error is '{0}'"
         comment = comment_fmt.format(str(err))
-        return (False, comment, {})
+        changes = {}
+        return (False, (comment, changes))
     except CommandExecutionError as err:
         comment_fmt = "Call to module function 'reg.import_file' has failed. Error is '{0}'."
         comment = comment_fmt.format(err.message)
-        return (False, comment, {})
-    return (True, None, None)
+        changes = {}
+        return (False, (comment, changes))
+    return (True, None)
 
 
 def _imported_file_data_and_test(reference_reg_file_url,use_32bit_registry):
@@ -874,6 +875,7 @@ def _imported_file_data_and_test(reference_reg_file_url,use_32bit_registry):
     info = (reg_location, reference_reg_file, reference_data)
     return (None, info)
 
+
 def _get_present_state_data_and_test(reg_location,reference_data,use_32bit_registry):
     r'''
     This is a utility function used by imported_file. 
@@ -888,7 +890,8 @@ def _get_present_state_data_and_test(reg_location,reference_data,use_32bit_regis
            a will be False if the data acquisition succeeds and the test fails,
            indicating that the although the import may have appeared to succeed,
            success was not observed when tested.
-        b: In all cases if b will be a suitable comment for the return dictionary.
+        b: In all cases if b will be a an ordered pair consisting
+           of a suitable comment with a changes dictionary.
     '''
     # acquire new data corresponding with the import.
     (get_present_state_data_success, info) = \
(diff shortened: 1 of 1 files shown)


H20-17:
This is bug free and does exactly what I think it should but I decided that I could improve the code a lot, so I'm refactoring.

cmcmarrow (approved):

cmcmarrow:
Hey H20-17! Thanks for the code. It looks good. Can you please write some test for your code? If you need help with writing the test please let me know. I'll be happy to help you.

H20-17:
> Hey H20-17! Thanks for the code. It looks good. Can you please write some test for your code? If you need help with writing the test please let me know. I'll be happy to help you.

Hi @cmcmarrow, I think that the code is good insofar as it works (i.e. you get the right output for the input you give it), but I think that it's not very elegant and doesn't capture logic of salt states in the most concice way. It's still WIP.

Yes, I confess i know nothing about testing. I may need assistance with that. As soon as this is non-WIP (which should be appxox. tonight or tomorrow some time as soon as I get the time to make the changes I want), I'd like some pointers about what I'm supposed to do about that because I haven't got a clue.

H20-17:
Please no one review the code until this ceases to be WIP. I'm happy to have people test it by running it since its working and the functionality is complete, and the functionality will not change because of the refactoring.

H20-17:
This is no longer WIP. The github lint test is not completing. The function is now just called "imported" instead of "imported_file. The code has been refactored a lot to be more clear.

salt/states/reg.py
```diff
@@ -617,3 +621,317 @@ def key_absent(name, use_32bit_registry=False):
...
+    ConfigParser object. A configparser.Error exception will be thrown implicitly
+    upon failure to parse.
+    '''
+    reg_file_fp = io.open(reg_file, "r", encoding="utf-16")
+    try:
+        reg_file_fp.readline()
+        # The first line of a reg file is english text which we must consume before parsing.
+        # It contains no data.
+        reg_data = configparser.ConfigParser()
+        PY2 = sys.version_info[0] == 2
```
damon-atkins: PY2 is already defined see other salt code for examples.

salt/states/reg.py
```diff
@@ -617,3 +621,317 @@ def key_absent(name, use_32bit_registry=False):
...
+    fails, it raises a CommandExecution error exception. If the function succeeds,
+    it returns a parser object representing the registry at the location.
+    '''
+    if use_32bit_registry:
+        word_sz_txt = "32"
+    else:
+        word_sz_txt = "64"
+    present_reg_file = (__salt__['temp.file'])()
+    try:
+        cmd = 'reg export "{0}" "{1}" /y /reg:{2}'.format(reg_location, present_reg_file, word_sz_txt)
```
damon-atkins: This assumes all items are under a single tree.

salt/states/reg.py
```diff
@@ -617,3 +621,317 @@ def key_absent(name, use_32bit_registry=False):
...
+    reg_location = (reference_parse.sections())[0]
+    try:
+        state_parse = _imported_state_data(reg_location, use_32bit_registry)
+    except CommandExecutionError as err:
+        comment = _imported_compose_cmd_execution_err_msg(err)
+        return (False, comment)
+    changes = {}
+    operation = None
+    if not _imported_test(state_parse, reference_data):
+        changes['old'] = "Registry unmodified."
```
damon-atkins: These should be empty.     The `comment `filed is the only field to have comments.  new and old should only contain what the old values & delete items and new contain changes to existing values & additional items.

damon-atkins (changes requested):
Whats the aim of the code.  I think I known what you are trying to do.  Bit of introduction is needed.

What actions are taken if something did exist in the file and and later in the future it no longer exists.
What are the limitations,  e.g. everything needs to be under the same tree?
Are you better off not using the reg.exe at all, given you need to parse the file anyway?

damon-atkins:
PS   [WIP] at the front of the description to indicate it is work in progress

H20-17:
> Whats the aim of the code. I think I known what you are trying to do. Bit of introduction is needed.
> 
> What actions are taken if something did exist in the file and and later in the future it no longer exists.
I don't understand this question.
> What are the limitations, e.g. everything needs to be under the same tree?
Yes it assumes that the reg file is the consequence of a reg export, which I think implies a single tree. Under what situations might we end up with other types of reg files? I'll clarify that in the text if you think this is an issue.
> Are you better off not using the reg.exe at all, given you need to parse the file anyway?
It keeps the code simpler because I don't have to interpret the meaning of values in reg files. Instead I,m just do a string comparison of parsed values without worrying about types. Also this is _possibly_ faster. If I don't grab a section of the registry and turn it into some kind of data, the alternative is a lot of individual lookups. It sounds slow-ish but perhaps it isn't.

salt/states/reg.py
H20-17: > These should be empty. The `comment `filed is the only field to have comments. new and old should only contain what the old values & delete items and new contain changes to existing values & additional items.

The changes dictionary i'm returning gets used when in test mode or later on if the operation is performed and succeeds (I hang on to to is so I can use it in the final output), so it needs to not be empty in the situations where i'm actually using it.

I realise that the contents of the dictionary isn't standards conformant. I can fix that but it'll be a few minutes of work since I really should be creating the changes dictionary as I'm testing. It's not a big deal.

H20-17:
@damon-atkins I have attempted to address all your issues now apart from the multiple tree roots issue (but i mentioned this a limitation in the documentation) and I am still just exporting and comparing the parse trees which I sitll think is the right approach. I have made sure that the return changes dictionary now has good data. 

The changes dictionary ends up looking like this:
```
              new:
                  ----------
                  added keys:
                      - HKEY_LOCAL_MACHINE\SOFTWARE\SaltTestKey\subKey5
                  added/changed values:
                      ----------
                      HKEY_LOCAL_MACHINE\SOFTWARE\SaltTestKey\subKey3:
                          |_
                            ----------
                            @:
                                oik
                          |_
                            ----------
                            g:
                                ciuh
                      HKEY_LOCAL_MACHINE\SOFTWARE\SaltTestKey\subKey5:
                          |_
                            ----------
                            @:
                                oki2
                          |_
                            ----------
                            g:
                                c
                          |_
                            ----------
                            xterm256colour:
                                dword:10001011
              old:
                  ----------
                  values:
                      ----------
                      HKEY_LOCAL_MACHINE\SOFTWARE\SaltTestKey\subKey3:
                          |_
                            ----------
                            @:
                                ok
                          |_
                            ----------
                            g:
                                ch

```

sagetherage:
@H20-17 please rebase this against the master branch

sagetherage:
@H20-17 please rebase this against the master branch

sagetherage:
Reviewed in Grooming 2021-May-03 considering closing

PR #44830, opened 2017-12-05
Implement variable fileserver update intervals

terminalmage:
This PR moves fileserver updates out of the maintenance process and into their own separate process. Within that process, backends/remotes being updated for a given unique interval are all bunched together, so we end up with one thread spawned for each unique interval.

In the process of implementing this, the virtual names for ``gitfs``, ``hgfs``, ``svnfs``, and ``minionfs`` have all been changed to normalize them with the other non-roots backends, which all end in "fs". To accommodate this change, when the master/minion config is loaded, the fileserver backends are examined and ``git``, ``hg``, ``svn``, ``minion`` are replaced with ``gitfs``, ``hgfs``, etc. This allows old configuration to continue to work, while the examples in the docs now reference the new virtual names.

Renaming the virtual names makes it easier for the ``FileserverUpdate`` process to get the update interval for each of the fileserver backends from the master opts, as we'll be able to look for an option named ``<backend>_update_interval>`` and expect it will be there.

My main concern is that I'm not sure yet how our signal handlers will treat the threads spawned for the different variable intervals. I've only tested this in Docker thus far, so I need to test it in an environment where I have an init system and can see what happens when I shut down the service.

Resolves #40776.

CC: @jmalerbsjr, this should satisfy your concerns with update intervals, as we discussed in #44384.

PR #53674, opened 2019-07-01
salt.utils.gitfs: fix fetching specific remote

terminalmage:
The default behavior for gitfs is to fetch all remotes. The "remotes"
argument to this function is meant for additional flexibility should
anyone be using salt.utils.gitfs outside of the context of
gitfs/git_pillar. This is therefore a low-impact bug, but a bug
nonetheless.

Good catch by @sathieu https://github.com/saltstack/salt/pull/44830#discussion_r298210579

### DIFF
--- a/salt/utils/gitfs.py
+++ b/salt/utils/gitfs.py
@@ -2358,7 +2358,9 @@ class GitBase(object):
         changed = False
         for repo in self.remotes:
             name = getattr(repo, 'name', None)
-            if not remotes or (repo.id, name) in remotes:
+            if not remotes or repo.id in remotes or name in remotes:
+                # Fetch if no specific remote desired, or if the repo's "id" or
+                # "name" attribute matches one of the desired remotes
                 try:
                     if repo.fetch():
                         # We can't just use the return value from repo.fetch()


sathieu:
@terminalmage Is there any link with `update_intervals` method (in `salt/utils/gitfs.py`), which returns a dict with `(id, name)` as key:
```python
    def update_intervals(self):
        '''
        Returns a dictionary mapping remote IDs to their intervals, designed to
        be used for variable update intervals in salt.master.FileserverUpdate.

        A remote's ID is defined here as a tuple of the GitPython/Pygit2
        object's "id" and "name" attributes, with None being assumed as the
        "name" value if the attribute is not present.
        '''
        return {(repo.id, getattr(repo, 'name', None)): repo.update_interval
                for repo in self.remotes}
```

terminalmage:
Oh yeah, I think I was confusing this with another change I made to gitfs at about the same time. When we setup all the update buckets, we append an `id_` value which is a tuple that comes from the key of the return dict from `update_intervals()` in salt/utils/gitfs.py (See [here](https://github.com/saltstack/salt/blob/48609bb/salt/master.py#L405-L418)). Later on, we end up calling the update function with that same tuple as input, building the tuple [here](https://github.com/saltstack/salt/blob/48609bb/salt/master.py#L450-L455).

Based on this, I'm going to close this PR.

PR #53678, opened 2019-07-01
Fix LGPO when string object is None

twangboy:
### What does this PR do?
Return encoded null when string value is None. Encode to `utf-16-le` otherwise. Raise an error if a string value is not passed. Add some tests.

https://github.com/saltstack/lock/issues/1688

### Tests written?
Oui

### Commits signed with GPG?
Yes


amalaguti (approved):
It did work!

dwoz (approved):

lomeroe (approved):

cmcmarrow (approved):

PR #53681, opened 2019-07-02
Fix util users chugid and umask

pkwestm:
### What does this PR do?
When group parameter was added to `salt.utils.user` previous behaviour of the function `chugid_and_umask` changed. When parameter `runas` is provided but group parameter is not the effective group will remain as the executing user of salt. Previously this was changed to the primary group of user in runas parameter. This PR restores the old behaviour so that old code continues to work when group parameter is omitted.

### Previous Behavior
Example below use cmd.run to show behaviour. Notice gid in output. 
```
[saltminion2019] root@saltminion2019:~ # salt-call cmd.run id runas=wpa
local:
    uid=24774(wpa) gid=0(root) groups=0(root),10513(domain_users)
[saltminion2019] root@saltminion2019:~ #
```

### New Behavior
```
[saltminion2019] root@saltminion2019:~ # salt-call cmd.run id runas=wpa
local:
    uid=24774(wpa) gid=10513(domain_users) groups=10513(domain_users)
[saltminion2019] root@saltminion2019:~ #
```
### Tests written?
Yes, could not find any existing test for user.py. New file created unit/utils/test_user.py. Please let me know if tests should be implemented elsewhere.

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/utils/user.py
+++ b/salt/utils/user.py
@@ -238,18 +238,24 @@ def chugid_and_umask(runas, umask, group=None):
     set_grp = False
 
     current_user = getpass.getuser()
+    current_grp = grp.getgrgid(pwd.getpwnam(getpass.getuser()).pw_gid).gr_name
+
     if runas and runas != current_user:
         set_runas = True
         runas_user = runas
     else:
         runas_user = current_user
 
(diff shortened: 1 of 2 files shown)


mattp-:
:+1:

Akm0d (approved):
This change looks good.  The location of the test case seems to be appropriate  :+1:

PR #53682, opened 2019-07-02
Fix compare issue in lgpo state module

twangboy:
### What does this PR do?
The state was failing when the current policy and the new policy were differing types.
Creates a `_compare_policies` helper function to determine if policies are the same.
Compares `dicts` to `dicts` and `lists` to `lists`. Otherwise, returns `False`
Adds some unit tests

https://github.com/saltstack/lock/issues/1740

### Tests written?
Yes

### Commits signed with GPG?
Yes


lomeroe (approved):

amalaguti (approved):
Did work! thanks @twangboy

dwoz (approved):

garethgreenaway (approved):

Issue #53567, opened 2019-06-22
Salt-cloud API completely broken and nonfunctional sometime between April and now


fake-name:
### Description of Issue
I'm using the `salt.cloud.CloudClient()` interface to manage salt minions.

Currently, every invocation of `CloudClient().<function>` by my code now fails.

```
Traceback (most recent call last):
  File "salt_scheduler.py", line 326, in <module>
    refresh()
  File "salt_scheduler.py", line 259, in refresh
    sched.destroy_vm(vm_name)
  File "salt_scheduler.py", line 140, in destroy_vm
    self.interface.destroy_client(vm_name)
  File "/home/client/AutoTriever/marshaller/salt_runner.py", line 659, in destroy_client
    ret = self.cc.destroy([clientname])
  File "/usr/local/lib/python3.6/dist-packages/salt/cloud/__init__.py", line 390, in destroy
    mapper.destroy(names)
  File "/usr/local/lib/python3.6/dist-packages/salt/cloud/__init__.py", line 1023, in destroy
    ret = self.clouds[fun](name)
  File "/usr/local/lib/python3.6/dist-packages/salt/cloud/clouds/linode.py", line 742, in destroy
    transport=__opts__['transport']
  File "/usr/local/lib/python3.6/dist-packages/salt/utils/cloud.py", line 2016, in fire_event
    time.sleep(0.025)
  File "/usr/local/lib/python3.6/dist-packages/salt/utils/event.py", line 905, in __exit__
    self.destroy()
  File "/usr/local/lib/python3.6/dist-packages/salt/utils/event.py", line 790, in destroy
    self.io_loop.close()
  File "/usr/lib/python3/dist-packages/tornado/ioloop.py", line 716, in close
    self.remove_handler(self._waker.fileno())
  File "/usr/lib/python3/dist-packages/tornado/platform/posix.py", line 48, in fileno
    return self.reader.fileno()
ValueError: I/O operation on closed file
```

```
Traceback (most recent call last):
  File "salt_runner.py", line 930, in <module>
    go()
  File "salt_runner.py", line 922, in go
    fmap[command]()
  File "salt_runner.py", line 753, in dtest
    ret = herder.cc.create(names=[clientname], provider=provider, **kwargs)
  File "/usr/local/lib/python3.6/dist-packages/salt/cloud/__init__.py", line 429, in create
    mapper.create(vm_))
[...]

PR #53684, opened 2019-07-02
Resolve https://github.com/saltstack/salt/issues/53567

fake-name:
### What does this PR do?

This removes the reactor teardown in the salt-cloud function calls.

As far as I understand it, the reactor is a global singleton, and should be torn down properly at the interpreter exit anyways. 

Right now, since it's shut down early, any calls to salt-cloud functions will subsequently explode. This makes the salt-cloud API somewhat useless, since you get *one* function call before everything is broken.

https://github.com/saltstack/salt/issues/53567


dwoz:
@fake-name Ideally this would get a test.

fake-name:
I have no idea how to test this without a account at a VPS provider somewhere.

Akm0d:
I might be able to help with a test case, but won't have time until next week to work on it

fake-name:
Note that this now also includes support for deploying vultr VMs using SSH keys.

If you specify a ssh_key in the config file, that also exists in your vultr account, it'll use ssh-key based auth for your VM. If not, it falls back to password auth.

whytewolf:
Since this event_get item also exists in 2019.2.0  this might be better ported to the 2019.2 releases and brought up to neon and develop. 

I can see this event_get closing the event bus in other ways that are not so obvious in other versions of python.

whytewolf:
Also, the ssh keys stuff should be its own PR.

fake-name:
> Since this event_get item also exists in 2019.2.0 this might be better ported to the 2019.2 releases and brought up to neon and develop.

All I want to do is use salt as a library without it exploding. Arrrrgh.

> I can see this event_get closing the event bus in other ways that are not so obvious in other versions of python.

Does that translate to "make the default to not destroy the event loop"?

whytewolf:
> > Since this event_get item also exists in 2019.2.0 this might be better ported to the 2019.2 releases and brought up to neon and develop.
> 
> All I want to do is use salt as a library without it exploding. Arrrrgh.

I can applaud that. However, It is always good to make sure that if something breaks we get the break as early in the code as possible so that when things merge up they don't break again. Just trying to avoid later issues. as well as make sure things in earlier revisions work.

> 
> > I can see this event_get closing the event bus in other ways that are not so obvious in other versions of python.
> 
> Does that translate to "make the default to not destroy the event loop"?

No, Your fix is the right way to go. I'm just saying that while your testing showed the failure in python3. It might be a more silent failure in py2.

DmitryKuzmenko:
@fake-name we don't accept any PRs to develop anymore. Please rebase this to master branch. Also please install the `pre-commit` hook as described in the [contribution quide](https://docs.saltstack.com/en/latest/topics/development/contributing.html)

sagetherage:
@fake-name likely this will be closed as abandoned work, please rebase against master, use the pre-commit hook, resolve merge conflicts, and we have test clinics weekly to help with writing tests, if not then we will close this PR.

waynew (changes requested):
If you need any help with writing tests, please reach out, I'm in #salt on freenode, community Slack, or I stream Test Clinics on twitch on T/Th

ScriptAutomate:
Closing this PR. Changes can be made and submitted again, if wanted, but should now be done from `master` and to `master`. This has otherwise become too indecipherable, and is in a broken state with all the branch conflicts.

PR #53687, opened 2019-07-02
Fix upgrade method to append repo argument to zypper when it is calle…

pkwestm:
### What does this PR do?
Fix in the zyppper pkg module when calling method without dist_upgrade but with fromrepo. PR also make methods list_upgrade and upgrade both accept fromrepo parameter as a list. This needs to be consistent when module is used through states.

Fix upgrade method to correctly append --repo argument to zypper command when called with fromrepo but without dist-upgrade parameter.
 Previously this caused a behaviour on state function pkg.uptodate where test mode showed that a subset of packages would upgrade, but when executution same state without test flag the all packages on machine upgraded!

### Previous Behavior
```
[saltminion] root@saltminion:~ # salt-call -l info pkg.upgrade fromrepo=sle-manager-tools12-updates-x86_64-sp4
....
[INFO    ] Executing command [u'zypper', u'--non-interactive', u'--no-refresh', u'update', u'--auto-agree-with-licenses'] in directory '/root'
....
local:
    ----------
[saltminion] root@saltminion:~ #
```

### New Behavior
```
[saltminion] root@saltminion:~ # salt-call -l info pkg.upgrade fromrepo=sle-manager-tools12-updates-x86_64-sp4
....
[INFO    ] Targeting repos: [u'sle-manager-tools12-updates-x86_64-sp4']
[INFO    ] Executing command [u'zypper', u'--non-interactive', u'--no-refresh', u'update', u'--auto-agree-with-licenses', u'--repo', u'sle-manager-tools12-updates-x86_64-sp4'] in directory '/root'
....
local:
    ----------
[saltminion] root@saltminion:~ #
```

### Tests written?
Yes

### Commits signed with GPG?
Yes


dwoz:
@xetix This needs to be re-opened against the `master` branch. We are no longer
accepting PRs to `2017.7`. You can link back to this PR to preserve the

PR #53693, opened 2019-07-03
Allow passing "kwargs" to "pkg.list_downloaded" on Zypper (bsc#1140193)

meaksh:
### What does this PR do?

This PR fixes an issue with `pkg.list_downloaded` function in the Zypper module. This function currently does not accept any kwarg argument, which makes the logic of the `pkg.downloaded` state to crash.

Given the following example of SLS at `/srv/salt/test_pkg.sls`:

```yaml
test_pkg_downloaded:
  pkg.downloaded:
    - name: salt
```

Running `salt-call --local state.apply test_pkg` will produce a traceback:

```console
local:
----------
          ID: test_pkg_downloaded
    Function: pkg.downloaded
        Name: salt
      Result: False
                File "/usr/lib/python3.7/site-packages/salt/state.py", line 1939, in call
                  ret = self.states[cdata['full']](*cdata['args'], **cdata['kwargs'])
                File "/usr/lib/python3.7/site-packages/salt/loader.py", line 1982, in wrapper
                  return f(*args, **kwargs)
                File "/usr/lib/python3.7/site-packages/salt/states/pkg.py", line 2068, in downloaded
                  **kwargs)
                File "/usr/lib/python3.7/site-packages/salt/states/pkg.py", line 268, in _find_download_targets
                  cur_pkgs = __salt__['pkg.list_downloaded'](**kwargs)
              TypeError: list_downloaded() got an unexpected keyword argument 'fromrepo'
     Started: 09:20:36.896878
    Duration: 0.843 ms
     Changes:   

Summary for local
------------
Succeeded: 0
Failed:    1
------------
Total states run:     1
Total run time:   0.843 ms
```

### Tests written?

Yes

### Commits signed with GPG?

Yes


salt/modules/zypperpkg.py
```diff
@@ -2528,7 +2528,7 @@ def download(*packages, **kwargs):
     )
 
 
-def list_downloaded(root=None):
+def list_downloaded(root=None, **kwargs):
```
cmcmarrow: We don't use kwargs in list_downloaded. 
I believe this would be a better fix.
OLD
file "/usr/lib/python3.7/site-packages/salt/states/pkg.py", line 268, in _find_download_targets
                  cur_pkgs = __salt__['pkg.list_downloaded'](**kwargs)
NEW
file "/usr/lib/python3.7/site-packages/salt/states/pkg.py", line 268, in _find_download_targets
                  cur_pkgs = __salt__['pkg.list_downloaded'](kwargs.get('root'))

cmcmarrow (changes requested):

salt/modules/zypperpkg.py
aplanas: @cmcmarrow I like this idea, but in that case I would also change the signature here: https://github.com/saltstack/salt/blob/develop/salt/modules/yumpkg.py#L991

Also there are other places that require the change in pkg.py:

https://github.com/saltstack/salt/blob/develop/salt/states/pkg.py#L2166

meaksh: Hey @cmcmarrow, I understand your point but I've just added `kwargs` as an accepted argument to `list_downloaded` since that's the way most of the Zypper module (and also yumpkg) is implemented. There are many methods accepting `kwargs` even if they are not used at the moment:

```
 732 def version_cmp(ver1, ver2, ignore_epoch=False, **kwargs):                                                                                                              
1039 def get_repo(repo, root=None, **kwargs):  # pylint: disable=unused-argument                                                                                             
1055 def list_repos(root=None, **kwargs):
1556 def upgrade(refresh=True, dryrun=False, dist_upgrade=False, fromrepo=None, novendorchange=False, skip_verify=False, no_recommends=False, root=None, **kwargs):  # pylint: disable=unused-argument
1753 def remove(name=None, pkgs=None, root=None, **kwargs):  # pylint: disable=unused-argument                                                                               
1800 def purge(name=None, pkgs=None, root=None, **kwargs):  # pylint: disable=unused-argument
1960 def remove_lock(packages, root=None, **kwargs):  # pylint: disable=unused-argument
2044 def add_lock(packages, root=None, **kwargs):  # pylint: disable=unused-argument
2647 def list_patches(refresh=False, root=None, **kwargs):
2673 def list_installed_patches(root=None, **kwargs):
```

One of the reason for this might be to prevent changing the method signature (also for APIs) in case that there is a new argument that wants to be introduced to one of those method.

cmcmarrow: okay that's fair

cmcmarrow (approved):

meaksh:
Closing in favor of https://github.com/saltstack/salt/pull/56490

PR #53715, opened 2019-07-04
New recursive diff

github-abcde:
### What does this PR do?
It adds `recursive_diff` to `salt.utils.data`. This is a recursive diff that can handle mixed nested mappings/iterables. This is an alternative to `salt/utils/dictdiffer.py` or `salt/utils/listdiffer.py` that can only handle nested dicts and list-of-nested-dicts respectively.

None that I'm aware of.

### Previous Behavior
Only nested dicts or list-of-nested-dicts could be recursively diffed. Nothing is available to recursively diff a mix of lists, dicts, OrderedDicts, sets, tuples or other mapping or iterable type.

### New Behavior
`salt.utils.data.recursive_diff` is able to produce a dict with `old` and `new` keys for any combination of nested mapping- or iterable types. Some options are only applicable to certain types, like `ignore_keys` and `ignore_missing_keys` will only work on (nested) mappings, and `ignore_order` will only make a difference on ordered datatypes (like OrderedDict and list).

### Tests written?
Yes

### Commits signed with GPG?
Yes


github-abcde:
Closed in favor of https://github.com/saltstack/salt/pull/55759.

Issue #48500, opened 2018-07-10
Feature: grain ip4_routes


msciciel:
### Description of Issue/Question
It would be nice to have grain containing entries from routing table.  There are sometimes situation when configuration must depends on entries in routing table.

Proposed example output:
```
ip4_routes:
  - dst: 172.20.0.0/16
    dev: br-b472ad8a72d1
    proto: kernel
    scope: link
    src: 172.20.0.1
  - dst: default
    via: 172.27.100.126
    dev: enp0s31f6
    proto: static
    metric: 100
```

Maybe it would be better to always provide unused keys ?
```
ip4_routes:
  - dst: 172.20.0.0/16
    dev: br-b472ad8a72d1
    proto: kernel
    scope: link
    src: 172.20.0.1
    via:
    metric:
```

I could try to provide such grain if approved.

PR #53723, opened 2019-07-05
WIP: Add grain ip4_routes on linux

msciciel:
### What does this PR do?
Provide grain with list of routing table entries using common format.

Supported parameters to provide common output:
- destination
- netmask
- device
- proto
- scope
- source
- gateway
- metric

Because 'ip' and 'route' output does provide the same information
so it tries to provide similar experience. By default 'ip' is used
with fallback to 'route' when 'ip' is not available.

Example output:

ip4_routes:
- destination: 192.168.1.1
  device: eth0
  netmask: 255.255.255.255
  scope: link
- destination: 10.10.10.0
  device: eth0
  netmask: 255.255.255.0
  proto: kernel
  scope: link
  source: 10.10.10.10
- destination: 192.168.1.2
  device: eth0
  gateway: 10.10.10.254
  netmask: 255.255.255.0
- destination: 0.0.0.0
  device: bond0
  gateway: 10.10.10.1
  netmask: 0.0.0.0

#48500

### Tests written?
Yes

### Commits signed with GPG?
No


max-arnold:
This could potentially cause some performance problems on a node with full BGP routing table: https://bgp.potaroo.net/

msciciel:
@max-arnold good point. Similar situation could also occur on systems with a large number of network interfaces ie. network nodes of openstack cloud. I could imagine situations with hundreds or even thousands of interfaces on single machine which affects grains  `hwaddr_interfaces`, `ip4_interfaces`, `ip6_interfaces`, `ip_interfaces` but for sure thousands is not the same as hundreds of thounsands. 

Possible solutions which came to my mind:
- some limit should be enforced to import only `100` or `1000` routing entries? It should be fine propably for 99,99% systems. Because of https://docs.saltstack.com/en/latest/ref/configuration/master.html#max-event-size it would be nice to fit within default limit. During my tests `10k` route entries produce event size near default limit. Event size could be decreased twice using tuple instead of dictionary with fixed fields.
- skip grains if number of entries in `/proc/net/routes`/ exceeds some threshold?
- abandon change but it's sometimes useful to have ability to specify pillar configuration based on routing table instead of using execution module inside formula to operate on routing table
- another idea?

I've have performed tests with `64k` route entries and it added extra `10s` (`0.5s` for `ip route show`) to grains execution time.

max-arnold:
I believe Salt needs a way to disable certain grains by default (the situation when everyone keeps adding grains to the core is not sustainable in the long term). There are a couple of existing config options, but they are somewhat tangential and aren't systematized: https://github.com/saltstack/salt/pull/53804#issuecomment-510359786

As for this particular PR, a good compromise would be to keep the `utils/network.py` part (or improve the `modules/network.py` so it does what you want out of the box), but drop the grain. Some rationale:

1. The use case is not entirely clear - `There are sometimes situation when configuration must depend on entries in routing table`.  What is the actual situation and how common is it?
2. What is the difference between your `utils/network.py:route()` function and the existing `modules/network.py:route()`?
3. If you really need the grain, there is an easy workaround:

```
ip4_routes:
  grains.present:
    - value: {{ salt['network.routes']() | yaml }}
```

(this is untested, just to illustrate the idea).

msciciel:
Another compromise may be to reduce scope of this PR to provide only default gateway with grain `ip4_gateway` or `ip4_default_gateway` instead. In most cases I need name of the default gateway interface to configure additional routes on it or to configure some service to listen only on it.

```
ip4_gateway:
  device: eth0
  gateway: 10.10.10.254
```

or

```
ip4_gateway:
  - 10.10.10.254  
  - eth0
```

Provide `utils/network.py:get_route()` and use it for grains and `modules/network.py:get_route()` to avoid code duplication.

I do not see any sense in keeping `utils/network.py:routes()` without using it somewhere but we may keep it if You consider it useful for the future.

If no one else see sense in such grain let's just drop this PR. I You agree with `ip4_gateway` grain then I could rewrite PR to provide it.

msciciel:
@sagetherage of course I can rebase it but propably before that we should decide about the shape of this change. Current version of salt provides grain `ip4_gw` and module `network.routes`. As mentioned before such grain could be provided using below state:

```
ip4_routes:
  grains.present:
    - value: {{ salt['network.routes']() | yaml }}
```
In my environment grain `ip4_gw` covers 99,9% of cases. I think that most of the systems have limited number of entries in routing table but it may happen that routing table contains hundreds or thousands of entries. I do not feel that I'm good person to decide if it make all people happy to see such grain. I can rewrite it to conform with current code but people with broader scope should decide.

Built-in grain has advantage of being available during first highstate.

Is it possible to define optional grains to enable and do not enable by default such grain with a performance impact?

sagetherage:
Reviewed in Grooming 2021-MAY-03 we would like to consider this behind a feature flag and turn on in a config - perhaps by an "allow list" or a bigger solution @max-arnold @msciciel would you two like to collaborate on a Salt Enhancement Proposal (SEP)?

sagetherage:
Reviewed in Grooming 2021-MAY-24 - closing as this should be a Salt Enhancement Proposal and we can see several ways to move forward on this, but we will not accept this change into the `develop` branch.

msciciel:
>  Salt Enhancement Proposal

@sagetherage I'm ready to serve. Where should we begin?

Issue #53012, opened 2019-05-14
integration.cloud.clouds.test_gce failures


garethgreenaway:
integration.cloud.clouds.test_gce.GCETest.test_instance_extra test failures
=======

2019.2 failed [salt-centos-7-py2-cloud](https://jenkinsci.saltstack.com/job/2019.2/job/salt-centos-7-py2-cloud/77/testReport/junit/integration.cloud.clouds.test_gce/GCETest/test_instance_extra), [salt-centos-7-py3-cloud](https://jenkinsci.saltstack.com/job/2019.2/job/salt-centos-7-py3-cloud/77/testReport/junit/integration.cloud.clouds.test_gce/GCETest/test_instance_extra)
---

'cloud-test-mqge3x:' not found in []

```

Traceback (most recent call last):
  File "/tmp/kitchen/testing/tests/support/helpers.py", line 212, in wrap
    raise exc
  File "/tmp/kitchen/testing/tests/support/helpers.py", line 209, in wrap
    return caller(cls)
  File "/tmp/kitchen/testing/tests/integration/cloud/clouds/test_gce.py", line 118, in test_instance_extra
    self.assertIn(ret_str, instance)
AssertionError: 'cloud-test-mqge3x:' not found in []


integration.cloud.clouds.test_gce.GCETest.test_instance test failures
=======

2019.2 failed [salt-centos-7-py2-cloud](https://jenkinsci.saltstack.com/job/2019.2/job/salt-centos-7-py2-cloud/77/testReport/junit/integration.cloud.clouds.test_gce/GCETest/test_instance), [salt-centos-7-py3-cloud](https://jenkinsci.saltstack.com/job/2019.2/job/salt-centos-7-py3-cloud/77/testReport/junit/integration.cloud.clouds.test_gce/GCETest/test_instance)
---

'cloud-test-fv10bo:' not found in []

```

Traceback (most recent call last):
  File "/tmp/kitchen/testing/tests/support/helpers.py", line 212, in wrap
    raise exc
  File "/tmp/kitchen/testing/tests/support/helpers.py", line 209, in wrap
    return caller(cls)
  File "/tmp/kitchen/testing/tests/integration/cloud/clouds/test_gce.py", line 87, in test_instance
    self.assertIn(ret_str, instance)
AssertionError: 'cloud-test-fv10bo:' not found in []
		
```

Ch3LL:
[...]

PR #53734, opened 2019-07-05
Don't call potentially missing shade library

Akm0d:
### What does this PR do?
See title

https://github.com/saltstack/salt/issues/53012


garethgreenaway (approved):

Akm0d:
Doesn't need a new test case because this PR causes previously failing tests to pass.

Issue #53740, opened 2019-07-06
grains.items throws exception on lastest SmartOS under python3


mamalos:
### Description of Issue
When trying to collect grains, the module fails to run smoothly on a clean install of latest SmartOS (20190703T233036Z). Consequently the global zone (GZ) is incorrectly not identified as a global zone, hence vmadm module won't work.

### Setup
Clean install of SmartOS 20190703T233036Z on VMware vCenterr 6.0, pkgsrc installed on GZ as suggested by the documentation, salt package salt-2019.2.0nb1 installed using pkgin.

### Steps to Reproduce Issue
Just install the SmartOS image on vCenter 6.0, install pkgsrc on the GZ "by the book" (https://pkgsrc.joyent.com/install-on-illumos/), install salt on the GZ using pkgin.

Run:
```
# salt-call grains.items
[WARNING ] /opt/tools/lib/python3.7/site-packages/salt/loader.py:772: DeprecationWarning: dist() and linux_distribution() functions are deprecated in Python 3.5
  ret = funcs[key]()

[WARNING ] /opt/tools/lib/python3.7/site-packages/salt/utils/path.py:265: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
  if not isinstance(exes, collections.Iterable):

[CRITICAL] Failed to load grains defined in grain file mdata.mdata in function <function mdata at 0x3a658f8>, error:
Traceback (most recent call last):
  File "/opt/tools/lib/python3.7/site-packages/salt/loader.py", line 798, in grains
    ret = funcs[key](**kwargs)
  File "/opt/tools/lib/python3.7/site-packages/salt/grains/mdata.py", line 160, in mdata
    grains = salt.utils.dictupdate.update(grains, _user_mdata(mdata_list, mdata_get), merge_lists=True)
  File "/opt/tools/lib/python3.7/site-packages/salt/grains/mdata.py", line 66, in _user_mdata
    mdata_value = __salt__['cmd.run']('{0} {1}'.format(mdata_get, mdata_grain), ignore_retcode=True)
  File "/opt/tools/lib/python3.7/site-packages/salt/modules/cmdmod.py", line 1167, in run
    **kwargs)
  File "/opt/tools/lib/python3.7/site-packages/salt/modules/cmdmod.py", line 633, in _run
    cmd = salt.utils.args.shlex_split(cmd)
[...]

PR #53794, opened 2019-07-10
Fixes SmartOS grains when using py3

sjorge:
### What does this PR do?
The salt package in pkgsrc was moved to py3, this broke some grains on SmartOS.
For now the current release 2019.3 is back on py27. But in preperation for the next release, this PR fixes those issues.

#53740

### Previous Behavior
Missing grains on SmartOS (Global Zone) when using py3

### New Behavior
Grains are no longer missing

### Tests written?
Yes

### Commits signed with GPG?
No

### DIFF
--- a/salt/grains/mdata.py
+++ b/salt/grains/mdata.py
@@ -63,6 +63,8 @@ def _user_mdata(mdata_list=None, mdata_get=None):
         return grains
 
     for mdata_grain in __salt__['cmd.run'](mdata_list, ignore_retcode=True).splitlines():
+        if mdata_grain.startswith("ERROR:"):
+            continue
         mdata_value = __salt__['cmd.run']('{0} {1}'.format(mdata_get, mdata_grain), ignore_retcode=True)
 
         if not mdata_grain.startswith('sdc:'):
@@ -89,6 +91,7 @@ def _sdc_mdata(mdata_list=None, mdata_get=None):
         'datacenter_name',
         'hostname',
         'dns_domain',
+        'alias',
     ]
     sdc_json_keys = [
         'resolvers',
@@ -107,6 +110,8 @@ def _sdc_mdata(mdata_list=None, mdata_get=None):
 
     for mdata_grain in sdc_text_keys+sdc_json_keys:
         mdata_value = __salt__['cmd.run']('{0} sdc:{1}'.format(mdata_get, mdata_grain), ignore_retcode=True)
+        if mdata_value.startswith("ERROR:"):
+            continue
 
         if not mdata_value.startswith('No metadata for '):
             if 'mdata' not in grains:
--- a/salt/grains/mdata.py
+++ b/salt/grains/mdata.py
@@ -64,6 +64,7 @@ def _user_mdata(mdata_list=None, mdata_get=None):
 
     for mdata_grain in __salt__['cmd.run'](mdata_list, ignore_retcode=True).splitlines():
         if mdata_grain.startswith("ERROR:"):
+            log.warning("mdata-list returned an error, skipping mdata grains.")
             continue
         mdata_value = __salt__['cmd.run']('{0} {1}'.format(mdata_get, mdata_grain), ignore_retcode=True)
 
@@ -111,6 +112,9 @@ def _sdc_mdata(mdata_list=None, mdata_get=None):
     for mdata_grain in sdc_text_keys+sdc_json_keys:
         mdata_value = __salt__['cmd.run']('{0} sdc:{1}'.format(mdata_get, mdata_grain), ignore_retcode=True)
         if mdata_value.startswith("ERROR:"):
+            log.warning("unable to read sdc:{0} via mdata-get, mdata grain may be incomplete.".format(
+                mdata_grain,
+            ))
             continue
 
         if not mdata_value.startswith('No metadata for '):
--- a/salt/grains/smartos.py
+++ b/salt/grains/smartos.py
@@ -107,14 +107,14 @@ def _smartos_zone_data():
     #   zoneid
     #   zonename
     #   imageversion
+    grains = {}
 
-    grains = {
-        'zoneid': __salt__['cmd.run']('zoneadm list -p | awk -F: \'{ print $1 }\'', python_shell=True),
-        'zonename': __salt__['cmd.run']('zonename'),
-        'imageversion': 'Unknown',
-    }
+    zoneinfo = __salt__['cmd.run']('zoneadm list -p').strip().split(":")
+    grains["zoneid"] = zoneinfo[0]
+    grains["zonename"] = zoneinfo[1]
 
     imageversion = re.compile('Image:\\s(.+)')
+    grains["imageversion"] = "Unknown"
     if os.path.isfile('/etc/product'):
         with salt.utils.files.fopen('/etc/product', 'r') as fp_:
             for line in fp_:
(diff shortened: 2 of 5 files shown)


sjorge:
Some data we can use to mock mdata-list:

1. failure case
```
[root@carbon ~]# mdata-list > /tmp/out 2> /tmp/err
[root@carbon ~]# cat /tmp/out
[root@carbon ~]# cat /tmp/err
ERROR: could not initialise protocol: I don't know how to get metadata on this system.
```

2. success case
```
isotope# mdata-list > /tmp/out 2> /tmp/err
isotope# cat /tmp/out
primary_mac
user-script
minion_key
minion_pub
isotope# cat /tmp/err
```

(yes, sdc:XXX values are not listed)

3. mdata-get failure missing prop
```
isotope# mdata-get sdc:alaisx > /tmp/out 2> /tmp/err
isotope# cat /tmp/out
isotope# cat /tmp/err
No metadata for 'sdc:alaisx'
```

4. mdata-get OK
```
isotope# mdata-get sdc:alias > /tmp/out 2> /tmp/err
isotope# cat /tmp/err
isotope# cat /tmp/out
isotope
```

5. mdata-get hard errro
```
[isotope :: sjorge][~]
[.]$ mdata-get sdc:alias > /tmp/out 2> /tmp/err
[isotope :: sjorge][~]
[!]$ cat /tmp/out
[isotope :: sjorge][~]
[.]$ cat /tmp/err
ERROR: could not initialise protocol: Could not find metadata socket.
```

sjorge:
Ideally the test should just run through the grain mdata grain parsing and with all the output combinations and we should get the expected grains or lack of the grains (but no exceptions)

sjorge:
I got some feedback there are still some grains missing, will dig a bit deeper and update the PR.

sjorge:
Some more fixes were needed in salt.utils.platform as the is_smartos_zone and is_smartos_globalzone were failing on bytes vs string literal on py3.

sjorge:
So I talked with @Ch3LL and we need the following tests:
- unit for grains.mdata
- unit for grains.smartos

Probably something for salt.utils.platform too as that was the read cause of the issue, it just expose the other to grains to some undefined behavior which we cannot hit otherwise.

sjorge:
Tests for the functions in both grain modules are done, and some small more tweaks to salt.grains.smartos to make it more efficient.

Ch3LL (approved):

salt/grains/mdata.py
```diff
@@ -63,6 +63,8 @@ def _user_mdata(mdata_list=None, mdata_get=None):
         return grains
 
     for mdata_grain in __salt__['cmd.run'](mdata_list, ignore_retcode=True).splitlines():
+        if mdata_grain.startswith("ERROR:"):
```
waynew: Would it be a good idea to log here? Or should this data only be in the `mdata_grain` with logging elsewhere?

salt/grains/smartos.py
```diff
@@ -139,30 +139,31 @@ def _smartos_zone_pkgsrc_data():
...
-
     pkgsrcpath = re.compile('PKG_PATH=(.+)')
-    if os.path.isfile('/opt/local/etc/pkg_install.conf'):
-        with salt.utils.files.fopen('/opt/local/etc/pkg_install.conf', 'r') as fp_:
-            for line in fp_:
-                line = salt.utils.stringutils.to_unicode(line)
-                match = pkgsrcpath.match(line)
-                if match:
-                    grains['pkgsrcpath'] = match.group(1)
+    pkgsrcversion = re.compile('^https?://pkgsrc.joyent.com/packages/SmartOS/(.+)/(.+)/All$')
```
waynew: Is this the only URL this will ever be? Also, is there a better place to put this - maybe as a "constant" at the toplevel of the file? I'm just trying to think of where would be the best place to put this in the case that they change up their URL schemes.

salt/utils/platform.py
```diff
@@ -91,16 +91,17 @@ def is_smartos_globalzone():
...
     else:
-        cmd = ['zonename']
         try:
-            zonename = subprocess.Popen(
-                cmd, shell=False,
-                stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+            zonename_proc = subprocess.Popen(
+                ['zonename'], stdout=subprocess.PIPE, stderr=subprocess.STDOUT,
+            )
+            zonename_output = zonename_proc.communicate()[0].strip().decode("utf-8")
```
waynew: pretty sure this should be [`__salt_system_encoding__`](https://docs.saltstack.com/en/latest/ref/states/writing.html#strings-and-unicode)

salt/utils/platform.py
```diff
@@ -114,16 +115,17 @@ def is_smartos_zone():
...
```
waynew: same here - `__salt_system_encoding__`

salt/grains/mdata.py
sjorge: We were only getting here in a one case that was due to the broken is_smartos_XXX functions in py3, and this was throwing exceptions. With this if in place if we ever get here we skip the invalid date. I don't think this needs to be logged, as the only way to get here is if is_smartos_XXX() is broken again, just having the grain missing is OK in that case.

In theory we could remove this one and the one below on L113 again and we'll have exceptions -> error log entries again.

waynew (changes requested):

salt/grains/smartos.py
sjorge: On SmartOS (which these grains are specific for) this should be the only URL we'll encounter. The URL is probably different on NetBSD which also uses pkgsrc, but I'm not sure they even use the same release tagging methode that SmartOS has.

salt/utils/platform.py
sjorge: The docs seems to indicate this is specifically for encoding when writing out str/unicode data to disks? Here we are ingesting data on a lower level (aka not wrapping salt.module.cmdmod.run() because it is not yet available, which takes care of these encoding bits for us), so we do the decode to ensure we feed unicode data to the rest of salt.

sjorge: The docs seems to indicate this is specifically for encoding when writing out str/unicode data to disks? Here we are ingesting data on a lower level (aka not wrapping salt.module.cmdmod.run() because it is not yet available, which takes care of these encoding bits for us), so we do the decode to ensure we feed unicode data to the rest of salt.

sjorge: I found a few locations were this is also use with decode, I will change it and retest.

sjorge: At a quick glance it looks like `__salt_system_encoding__` is undefiend for salt/utils/platform.py, maybe because it is so low level most of the salt objects like __salt__ and friends also don't exist yet?

waynew: If the system isn't utf-8, then decoding as utf-8 *could* bork up the encoding. I mean, it *should* be utf-8 (and use UTC), but not all sysadmins have the luxury of dealing with sane systems ;)

salt/grains/smartos.py
sjorge: I have added a note above that explains why 'Unknown' would the the correct value of a sysadmin decided to create there own 'spin' of a SmartOS native zone. (Sort of like creating an entirely new linux distribution. e.g. debian <-> redhat <-> suse)

So it is clear why there is only one URL provided from the code + comment.

salt/grains/mdata.py
sjorge: Added some 'warning' level locking indicating that the mdata grains may be incomplete or missing, but this should not cause an message of level error or higher as it did before when an exception was thrown.

salt/grains/smartos.py
```diff
@@ -139,30 +139,38 @@ def _smartos_zone_pkgsrc_data():
...
-                if match:
-                    grains['pkgsrcversion'] = match.group(1)
-
+    # NOTE: we are specifically interested in the SmartOS pkgsrc version and path
+    #       - PKG_PATH MAY be different on non-SmartOS systems, but they will not
+    #         use this grains module.
+    #       - A sysadmin with advanced needs COULD create a 'spin' with a totally
+    #         different URL. But at that point the value would be meaning less in
+    #         the context of the pkgsrcversion grain as it will not followed the
+    #         SmartOS pkgsrc versioning. So 'Unknown' would be appropriet.
```
waynew: s/appropriet/appropriate

waynew (changes requested):
Small typo, otherwise this looks good from my perspective.

salt/grains/smartos.py
sjorge: Fixed

waynew (approved):

PR #53802, opened 2019-07-10
SaltClass classes jinja inheritance

ReubenM:
### What does this PR do?
Adds the parent directories all the way down to the root for saltclass classes to jinja's FileSystemLoader so that elements from parent directories can be accessed by the jinja engine. 

### New Behavior
For classes in saltclass, the jinja engine currently searches the directory of the class yaml file when using import or include. This PR adds the parent directories, so that if the import or include is not found in the current directory, it steps down through parent directories toward the saltclass classes root till it finds a match.

For my personal use-case, I have extensive tree structures for saltclass classes that share a lot of key names that need to be consistent across all classes. Setting those common key-names using salt-class variable expansion is not an option. (it only sets values, not key names, and for good reason) Using jinja without this PR offers little benefit since it only has access to the local directory of the class file. With this PR I can set a variable once, and propagate that variable to all classes in child directories using jinja's "from...import..." and use the variable for the key names.

### Tests written?
No: If needed I would prefer to wait till #52407 is merged since it pretty much re-writes all the saltclass integration tests. Not yet sure if there will be more churn in that space before it gets merged.

### Commits signed with GPG?
No


garethgreenaway:
@ReubenM Would you be able to rebase this PR against the master branch.  Thanks!

sagetherage:
Reviewed in Grooming 2021-MAR-03 and this is now pointing to master and updated it with `master` and letting tests run and new reviewer

Ch3LL (changes requested):
this will require tests and a changelog

dwoz:

PR #53837, opened 2019-07-12
Revert "Use winapi.com instead of coinitialize"

rares-pop:
This reverts commit a38300b156f6c2df78e72c34bb59613215342a5b.

### What does this PR do?
Fixes Windows behavior on the salt-minion, because the above commit uses pythoncom.CoInitialize() / CoUninitialize() in a flawed way.

salt.utils.winapi.Com() is a context manager for pythoncom - a way to use Windows COM interface for queries. That means, when you get out of the context, you can't touch any COM object in any way, because on `__exit__` it will call pythoncom.CoUninitilize() which releases all resources, which ends up in an immediat crash.
The above change added code like:
`with salt.utils.winapi.Com():`
`....conn = wmi.WMI()`
` comp = conn.Win32_ComputerSystem()[0]`

where the last line happens after pythoncom.CoUninitialize() happens, and end up in a crash.

I thought a quick fix would be put the whole code inside the 'with context', but there are cases where from the middle of a 'with context' calls a function that's again doing the 'with salt.utils.winapi.Com()', and while it's not a problem to call pythoncom.CoInitialize() multiple times, when the first pythoncom.CoUninitialize() happens, the whole COM interface that happens on the thread gets destroyed, so after returning from this function, calling anything from the COM object will crash.

Safest thing is to revert this and leave with possible leaks due to the fact that we never call pythoncom.CoUninitialize(), until the code is refactored properly.

A small test I did on Windows from an installed minion is, start a 'python' instance and execute the following:

`import pythoncom`
`import wmi`
`pythoncom.CoInitialize()`
`c = wmi.WMI()`
`pythoncom.CoInitialize()`
`pythoncom.CoUninitialize()`
`system = c.Win32_OperatingSystem()`
`pythoncom.CoUninitialize()`

It will crash on the 'system = ' line.

This has to be merged in all branches that contained the above change.

### Previous Behavior
Certain aspects of windows salt-minion were broken - anything that used salt/modules/win_system.py was crashing.

### New Behavior
Things to back to normal.

### Tests written?
No

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/modules/win_groupadd.py
+++ b/salt/modules/win_groupadd.py
@@ -13,11 +13,11 @@ from __future__ import absolute_import, unicode_literals, print_function
 # Import Salt libs
 import salt.utils.platform
 import salt.utils.win_functions
-import salt.utils.winapi
 
 
 try:
     import win32com.client
+    import pythoncom
     import pywintypes
     HAS_DEPENDENCIES = True
 except ImportError:
@@ -43,8 +43,8 @@ def _get_computer_object():
     Returns:
         object: Returns the computer object for the local machine
     '''
-    with salt.utils.winapi.Com():
-        nt = win32com.client.Dispatch('AdsNameSpaces')
+    pythoncom.CoInitialize()
+    nt = win32com.client.Dispatch('AdsNameSpaces')
     return nt.GetObject('', 'WinNT://.,computer')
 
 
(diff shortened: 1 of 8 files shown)


rares-pop:
Attn @twangboy, @dwoz.

rares-pop:
Another options is not to revert the change, but make the salt.utils.winapi.Com() refcounted, and put all the code that appears in the aforementioned commit under the with statements, not just the first call.

twangboy (approved):

rares-pop:
This is a revert of:
https://github.com/saltstack/salt/commit/a38300b156f6c2df78e72c34bb59613215342a5b

which is buggy, and I explained in the PR why.

If that particular change won't make it on the 'master', then we can safely close this PR and not port it over. The 'master' has the right behavior as of now.

attn: @dwoz

dwoz:

Issue #53838, opened 2019-07-12
zabbix_host module does not process additional host properties


slivik:
### Description of Issue
I try to create a new Zabbix host with saltstack.
The description of Zabbix_host.present stated this:
`This function accepts all standard host properties: keyword argument names differ depending on your zabbix version, see:https://www.zabbix.com/documentation/2.4/manual/api/reference/host/object#host`
I have added for example  tls keywords to the  Zabbix_host.present  command in the sls file:

### Setup
```
zabbix_create_host:
  zabbix_host.present:
    - host: {{ salt['grains.get']('nodename') }}
    - groups:
      - 20
    - interfaces:
        - default:
            - ip: '0.0.0.0'
            - dns: ''
            - type: 'agent'
            - port: '10050'
    - inventory_mode: 1
    - tls_connect: 2
    - tls_accept: 2
    - tls_psk_identity: PSK_{{ salt['grains.get']('nodename') }}
    - tls_psk: {{ salt['pillar.get']('zabbix:psk') }}
```
### Steps to Reproduce Issue
minion log file:
```
data: {"jsonrpc": "2.0", "params": {"interfaces": [{"ip": "0.0.0.0", "useip": "1", "dns": "", "main": "1", "type": "1", "port": "10050"}], "proxy_hostid": "0", "host": "<hostname>", "groups": [{"groupid": 20}], "inventory": {}}, "id": 0, "auth": "<auth_token>", "method": "host.create"}
```

### Versions Report
All versions affected

cmcmarrow:
Thank for letting us know

OrangeDog:
I've looked at the code and it ignores all properties other than `host`, `groups`, `interfaces`, `inventory`, and `proxy_host`. Even the documented `visible_name` is ignored.

As well as encryption settings, this also locks out `inventory_mode` and `description`.

Also missing, not directly related to this API but logically part of a host, is the ability to configure [host macros](https://www.zabbix.com/documentation/2.4/manual/api/reference/usermacro/object#host_macro).

slivik:
It ignores nothing. It is only preprocessing of all parameters to form them properly (pop them out of the dictionary) for next run of the state module and for passing them properly to zabbix execution module. If this is an issue could you, please, send some output?

OrangeDog:
@slivik it's trivial to demonstrate that they're ignored. Here's one example:
If the host doesn't exist it will be created, but the name not set. If it already exists, then the name will not be changed,

```yaml
[...]

PR #53550, opened 2019-06-20
[2019.2]Adjust zabbix_template state module to Zabbix 4.x

slivik:
### What does this PR do?
Zabbix version 4 introduces item value preprocessing process. This little adjustment fixes the template creation and is backward compatible with Zabbix version 3.

### Tests written?
No

### Commits signed with GPG?

Yes

PR #53839, opened 2019-07-12
fixed: zabbix_host module does not process additional host properties

slivik:
### What does this PR do?
Fixes proper processing of additional host properties in zabbix_host state module.

https://github.com/saltstack/salt/issues/53838

### Tests written?
No

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/states/zabbix_host.py
+++ b/salt/states/zabbix_host.py
@@ -68,7 +68,11 @@ def present(host, groups, interfaces, **kwargs):
 
 
     '''
-    connection_args = {}
+    connection_args = dict(kwargs)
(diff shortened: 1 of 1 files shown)


OrangeDog:
I'm not sure this is quite the way to do it. The `connection_args` is supposed to be the args used to connect to zabbix, and not sent in the actual API call.

(this will also need retargeting to master, and tests)

waynew:
@slivik thanks for the PR - my [comment on your other PR applies](https://github.com/saltstack/salt/pull/53550#issuecomment-575218904) could you rebase against master?

piterpunk:
@sagetherage the bug referenced by this PR was fixed and closed by #58604 so I guess this PR can be closed, too.

PR #53845, opened 2019-07-13
Optimized pip state tests

Akm0d:
### What does this PR do?
Centos7 pip module test improvements


twangboy:
What are you trying to fix with this? Which OS is failing? Is it still failing on this PR? https://github.com/saltstack/salt/pull/53825

twangboy:
Brought into https://github.com/saltstack/salt/pull/53825

Akm0d:
re-run docs

garethgreenaway (approved):

tests/support/helpers.py
```diff
@@ -1153,7 +1153,7 @@ def skip_if_not_root(func):
     if RUNTIME_VARS.PYTEST_SESSION:
         setattr(func, '__skip_if_not_root__', True)
 
-    if not sys.platform.startswith('win'):
+    if not salt.utils.platform.is_windows() and hasattr(os, 'getuid'):
```
dwoz: This seems weird. Rather than working around the os module not having `getuid` we should get to the root of what's causing that. Maybe sync with @s0undt3ch on that.

dwoz (changes requested):

tests/integration/modules/test_cmdmod.py
```diff
@@ -383,44 +383,44 @@ def test_hide_output(self):
...
             hide_output=True)
-        self.assertEqual(out, '')
+        self.assertFalse(out)
 
         # cmd.run_all (command should have produced stdout)
         out = self.run_function(
             'cmd.run_all',
             ls_command,
             hide_output=True)
-        self.assertEqual(out['stdout'], '')
```
cmcmarrow: I think self.assertEqual(out['stderr'], '') should not be changed.
A'lot of things can be False.

cmcmarrow (changes requested):
Nice Job. I just have one question/request

tests/integration/modules/test_beacons.py
```diff
@@ -44,15 +44,15 @@ def test_add_and_delete(self):
             ['ps', [{'processes': {'apache2': 'stopped'}}]],
             f_timeout=300
         )
-        self.assertTrue(_add['result'])
+        self.assertTrue(_add if isinstance(_add, bool) else _add['result'], _add)
```
dwoz: This shouldn't be needed since we resolved the test failure. Also, if `_add` is not a dictionary something went wrong. I think it makes more sense to use `assertSaltTrueReturn` rather than doing the inline if else.

tests/integration/modules/test_cmdmod.py
```diff
@@ -383,44 +383,44 @@ def test_hide_output(self):
             'cmd.run',
             ls_command,
             hide_output=True)
-        self.assertEqual(out, '')
+        self.assertFalse(out)
```
dwoz: Can we explain why this change is needed?

dwoz (changes requested):

tests/integration/modules/test_cmdmod.py
dwoz: Agreed, something seems weird about this change. Why would the test be returning a bool on 2019.2.1 but a string here?

Akm0d: That's true, there is a deeper problem that needs to be addressed if the same function is returning different types on different OSs.  I'll revert some of those changes and stick to the pip module test changes for now

tests/integration/modules/test_pip.py
```diff
@@ -6,18 +6,18 @@
 
 # Import python libs
 from __future__ import absolute_import, print_function, unicode_literals
+from git.compat import PY3
```
s0undt3ch: Hmm, `salt.ext.six.PY3`?

tests/integration/modules/test_pip.py
```diff
@@ -40,6 +40,48 @@ def cleanup_environ(environ):
...
+        for func in ('pip.freeze', 'pip.list'):
+            ret = self.run_function(func, bin_env=os.path.join(RUNTIME_VARS.TMP, 'no_pip'))
+            self.assertNotIsInstance(ret, list)
+            self.assertNotIsInstance(ret, dict)
+            self.assertIn('command required for \'{0}\' not found: '.format(func), ret.lower())
+            self.assertIn('could not find a pip binary', ret)
+
+    @skipIf(salt.utils.path.which('deactivate'), 'Must not be in a virtual environment for this test')
+    @destructiveTest
+    def test_system_pip3(self):
```
s0undt3ch: Why does it have to be the system's pip3?
We'll never run these under nox, it's always a virrualenv...

tests/integration/modules/test_pip.py
```diff
@@ -59,85 +101,57 @@ def cleanup_environ(environ):
...
+                if salt.utils.platform.is_windows():
+                    python = os.path.join(sys.real_prefix, os.path.basename(sys.executable))
+                else:
+                    python = os.path.join(sys.real_prefix, 'bin', os.path.basename(sys.executable))
+                # We're running off a virtualenv, and we don't want to create a virtualenv off of a virtualenv
+
+                # pyvenv-3.4 does not support the --python option
+                if not salt.utils.path.which('pyvenv-3.4'):
+                    self._venv_kwargs = {'python': python}
+        return self._venv_kwargs
```
s0undt3ch: Its safer if you return a `.copy()`

s0undt3ch: But good call!

s0undt3ch (commented):
Loved the code de-duplication efforts.
I have some more comments about moving methods(or properties) into test classes just because under pytest we should really move away from test classes. However, these tests still use the base TestCase class so, its not a blocker.

tests/support/helpers.py
s0undt3ch: Yeah, let's not work around that.

tests/integration/modules/test_pip.py
Akm0d: @soundtech Should the test be removed then if it will always be a virtualenv?  If it's always going to be skipped because it will always run in a virtualenv it isn't valuable

s0undt3ch: Well, Nox runs tests in a virtualenv, but salt usually works against the system python, so, my question is rather, does it matter if its the system pip or the virtualenv pip.
I'll have a better look at the test tomorrow since I'm writing from the phone and the tests could actually be really explicit as to why.

Akm0d: @s0undt3ch  It seems to me that this test is using the environment's pip to test that salt actually installed things correctly through pip.  I don't think the test makes sense to leave in there.

s0undt3ch: Do we have another test that covers installing a python package?

Akm0d: yes, within a virtual environment, I don't actually see the point of testing this with the system pip vs a virtualenv pip

Akm0d:
re-run py3/centos7/tcp

Akm0d:
re-run fedora29

Akm0d:
re-run centos7/tcp

Akm0d:
re-run centos7

Akm0d:
re-run centos7

Akm0d:
re-run centos7

Akm0d:
re-run full all

Akm0d:
re-run centos7-tcp

Akm0d:
re-run py2 ubuntu1604

cmcmarrow (approved):

Akm0d:
re-run windows

Akm0d:
re-run debian9

Akm0d:
re-run debian

Akm0d:
re-run debian

Akm0d:
re-run debian

Akm0d:
re-run full debian

Ch3LL (approved):

Ch3LL (approved):

Ch3LL (approved):

PR #53847, opened 2019-07-13
Always skip test, not just on suse

dwoz:
More test skips for war room efforts


twangboy (approved):

s0undt3ch:
We can close this PR.
The PyTest enhancements can land on https://github.com/saltstack/salt/pull/53869
and the pipeline fixes are in https://github.com/saltstack/salt/pull/53879

PR #53849, opened 2019-07-15
Fix container detection on virtual machines

nkukard:
When creating an LXC container on a virtual machine which has access to /proc/cpuinfo and potentially paths in /sys, the latter part of detection in core.py overrides the correct 'virtual' grain with the virtual machine type.

This fixes the issue by skipping the inspection of /proc and /sys files if we have already determined the virtualization type being used.

### What does this PR do?

This PR fixes the issue of container detection on Linux systems when the container is running on a virtual machine.

I think the patch is self explanatory. The only line being added is...
`if 'virtual' not in grains:`

### Previous Behavior

When running an LXC container on a VirtualBox virtual machine the following is observed...

> $ salt-call grains.items
> ...
>     virtual:
>        VirtualBox
> ...

### New Behavior

After applying the patch, the following is observed...

> $ salt-call grains.items
> ...
>     virtual:
>        LXC
> ...

### Tests written?

No

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/grains/core.py
+++ b/salt/grains/core.py
@@ -1034,38 +1034,42 @@ def _virtual(osdata):
                         grains["virtual_subtype"] = "Docker"
             except IOError:
                 pass
-        if os.path.isfile("/proc/cpuinfo"):
-            with salt.utils.files.fopen("/proc/cpuinfo", "r") as fhr:
-                if "QEMU Virtual CPU" in fhr.read():
-                    grains["virtual"] = "kvm"
-        if os.path.isfile("/sys/devices/virtual/dmi/id/product_name"):
-            try:
-                with salt.utils.files.fopen(
-                    "/sys/devices/virtual/dmi/id/product_name", "r"
-                ) as fhr:
-                    output = salt.utils.stringutils.to_unicode(
-                        fhr.read(), errors="replace"
+        # When creating an LXC container on a virtual machine which has access
+        # to /proc/cpuinfo and potentially paths in /sys, the below overrides
+        # the correct 'virtual' grain with the virtual machine type.
+        if ("virtual" not in grains) or (grains["virtual"] != "LXC"):
+            if os.path.isfile("/proc/cpuinfo"):
+                with salt.utils.files.fopen("/proc/cpuinfo", "r") as fhr:
+                    if "QEMU Virtual CPU" in fhr.read():
+                        grains["virtual"] = "kvm"
+            if os.path.isfile("/sys/devices/virtual/dmi/id/product_name"):
+                try:
+                    with salt.utils.files.fopen(
+                        "/sys/devices/virtual/dmi/id/product_name", "r"
+                    ) as fhr:
+                        output = salt.utils.stringutils.to_unicode(
+                            fhr.read(), errors="replace"
+                        )
+                        if "VirtualBox" in output:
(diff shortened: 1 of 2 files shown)


nkukard:
Updated patch to only catch LXC case

nkukard:
Rebased against develop

xeacott:
Thanks for submitting a PR and, I will also ask for some tests for this because this introduces some new code. Also, the lint job failed here however, the complaint is because a deprecated function is being used in the file - I'll ask and see how to remedy this.

xeacott:
Thanks for submitting a PR and, I will also ask for some tests for this because this introduces some new code. Also, the lint job failed here however, the complaint is because a deprecated function is being used in the file - I'll ask and see how to remedy this.

nkukard:
> Thanks for submitting a PR and, I will also ask for some tests for this because this introduces some new code. Also, the lint job failed here however, the complaint is because a deprecated function is being used in the file - I'll ask and see how to remedy this.

The latest force push I made should target only when virtual = LXC.

I've tested on VirtualBox myself (with and without LXC) and on barebones systems, but I hope you have more tests you can do from your side.

xeacott:
Unfortunately, we don't have exhaustive unit/integration tests for grains/core.py right now. Moreover, we don't have a test that would cover this change, let alone the function itself :-1: So, as part of generating a PR we do ask that you provide a written test too, which can be simple and only test your change. It helps us maintain a stable testing environment and increases our test coverage, which never hurts. I can point you over to `tests/integration/grains/text_core.py` for some quick examples in there.

sagetherage (changes requested):
please point this to the master branch instead of develop

nkukard:
> please point this to the master branch instead of develop

Done as requested. Effectively all this is is pretty much a 1 line change encapsulating the VM test if LXC was already detected.

xeacott:
This will also need pre-commit ran in order to pass, I can run that for you and push the changes if you cannot.

nkukard:
> This will also need pre-commit ran in order to pass, I can run that for you and push the changes if you cannot.

I installed pre-commit, re-committed the changes, it ran with no errors what I could see and I re-pushed.

I hope this is right? I was following this https://docs.saltstack.com/en/latest/topics/development/contributing.html

I'm not sure if you want me to perhaps post the output of the pre-commit? :)

xeacott:
Ahhh I'll babysit this PR and wait for pre-commit to finish and see what's up with it...

nkukard:
> Ahhh I'll babysit this PR and wait for pre-commit to finish and see what's up with it...

Ah I see an black fail on the CI system, I'll fix it quick.

xeacott:
Cool looks like pre-commit passed. At this point we'll just need to get this PR a test and we'll merge it in.

xeacott (approved):

sagetherage:
@toanju 
Looks like there is another PR that will fix the same issue here: https://github.com/saltstack/salt/pull/60196 - if you would and could please give a review to that PR - love to have your feedback!

toanju:
@sagetherage, you wanted to ping me?

sagetherage:
@toanju I must have made a mistake in pinging you, here. I can't follow my own comment - so sorry! Definitely my mistake!

PR #53858, opened 2019-07-15
Fixed logic around false negative in file.patch state with test=True

raddessi:
### What does this PR do?
Changes one conditional from bool() to if False so that when in test mode salt does not think it needs to make changes that it actually will not make when not in test mode.

I'm not sure this is the correct way to fix this, I can just say it does fix my issue. In fact I do not understand why previous behaviour examples below numbers 2 and 3 have different output so this very well may not be the correct fix.

We have some states that self patch salt, and this works except that when running in test mode salt always says it will patch the file even when the file has been patched successfully.

### Previous Behavior
1. `salt 'myminion' state.apply` will result in no changes
1. `salt 'myminion' state.apply test=True` will show that all file.patched states will have changes, AND that the changes are then entire contents of the patch file, which is quite alarming to look at. IE:
    ```
    ----------
              ID: patch /my-conda-env/lib/python3.6/site-packages/salt/beacons/napalm_beacon.py in 2019.2.0
        Function: file.patch
            Name: /my-conda-env/lib/python3.6/site-packages/salt/beacons/napalm_beacon.py
          Result: None
         Comment: File /tmp/__salt.tmp.4w45xiks updated
         Started: 20:45:21.051892
        Duration: 10.492 ms
         Changes:   
                  ----------
                  diff:
                      --- 
                      +++ 
                      @@ -0,0 +1,24 @@
                      +--- a/salt/beacons/napalm_beacon.py
                      ++++ b/salt/beacons/napalm_beacon.py
                      +@@ -286,8 +286,8 @@ def validate(config):
                      +     if not isinstance(config, list):
                      +         return False, 'Configuration for napalm beacon must be a list.'
                      +     for mod in config:
                      +-        fun = mod.keys()[0]
                      +-        fun_cfg = mod.values()[0]
                      ++        fun = list(mod.keys())[0]
                      ++        fun_cfg = list(mod.values())[0]
                      +         if not isinstance(fun_cfg, dict):
                      +             return False, 'The match structure for the {} execution function output must be a dictionary'.format(fun)
                      +         if fun not in __salt__:
                      +@@ -306,8 +306,8 @@ def beacon(config):
                      +         if not mod:
                      +             continue
                      +         event = {}
                      +-        fun = mod.keys()[0]
                      +-        fun_cfg = mod.values()[0]
                      ++        fun = list(mod.keys())[0]
                      ++        fun_cfg = list(mod.values())[0]
                      +         args = fun_cfg.pop('_args', [])
                      +         kwargs = fun_cfg.pop('_kwargs', {})
                      +         log.debug('Executing %s with %s and %s', fun, args, kwargs)
    ```
    I got here from this log entry when running the minion in debug mode:
    ```
    [DEBUG   ] failed to download salt://patches/2019-2-0/napalm_beacon.patch
    ```
1. `salt 'myminion' state.apply patches test=True` THIS however shows no changes, correctly.

### New Behavior
1. `salt 'myminion' state.apply` will result in no changes
1. `salt 'myminion' state.apply test=True` Shows no changes, correctly.
1. `salt 'myminion' state.apply patches test=True` Shows no changes, correctly.

### Tests written?
No

### Commits signed with GPG?

No

### DIFF
--- a/salt/states/file.py
+++ b/salt/states/file.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-"""
+'''
 Operations on regular files, special files, directories, and symlinks
 =====================================================================
 
@@ -276,7 +276,7 @@ For example:
             first_of_month: 6
             first_of_year: all
 
-"""
+'''
 
 # Import python libs
 from __future__ import absolute_import, print_function, unicode_literals
@@ -292,7 +292,7 @@ import sys
 import time
 import traceback
 from collections import Iterable, Mapping, defaultdict
-from datetime import datetime, date  # python3 problem in the making?
+from datetime import datetime, date   # python3 problem in the making?
 
 # Import salt libs
 import salt.loader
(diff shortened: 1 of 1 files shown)


dwoz:
@raddessi We are no longer accepting PRs to the `develop` branch. This needs to be re-opened against the `master` branch. You can link to this PR from the new one to preserve any disscussion history. Sorry for the confusion.

PR #53882, opened 2019-07-17
Fixed Bug  LVM has no Filesystem Type , no Filesystem is not allowed

tyl0re:
In the moment (when fs_type is empty) the else 
```
 if fs_type:
        cmd = ('parted', '-m', '-s', '--', device, 'mkpart', part_type, fs_type, start, end)
    else:
        cmd = ('parted', '-m', '-s', '--', device, 'mkpart', part_type, start, end)
```
 is never reached.Since
```
if not _is_fstype(fs_type):
        raise CommandExecutionError(
```
Rises the condtion earlier.

An Empty Filesystem is needed for LVM Volumes

 The Check was in earlier Versions of salt it was correct in parted.py:
```
443     if fs_type and fs_type not in set(['ext2', 'fat32', 'fat16', 'linux-swap', 'reiserfs',
444                           'hfs', 'hfs+', 'hfsx', 'NTFS', 'ufs', 'xfs', 'zfs']):
```

So the check on not defined fs_type was forgotten on recoding 

### What does this PR do?
It Fixed the Bug , that Partion without Filesystem can be created like LVM
It Fixed the Bug , that Partion without Filesystem can be created like LVM
### Previous Behavior
The Exception is rised when Filesystem type is empty

### New Behavior
Empty Filesystem type is Handled correct

### Tests written?

No

### Commits signed with GPG?

No

### DIFF
--- a/salt/modules/parted_partition.py
+++ b/salt/modules/parted_partition.py
@@ -515,7 +515,7 @@ def mkpartfs(device, part_type, fs_type, start, end):
             'Invalid part_type passed to partition.mkpartfs'
         )
 
-    if not _is_fstype(fs_type):
+    if fs_type and not _is_fstype(fs_type):
         raise CommandExecutionError(
             'Invalid fs_type passed to partition.mkpartfs'
         )


cmcmarrow:
Thanks @tyl0re for the PR. Can you also add a few tests for just the changes you made? This is so salt wont run into this bug again.

tyl0re:
im not a good Programmer, it was a bug we ran in SALT in our Susemanger, We compared it with an earlier Version which behaved correct. So we have patched it with the condition from the old code and tested it in our susemanager . SO i thought it is a good idea to push it pack for all users.
Im not Familiar with writing test since i normally am only an user not an programmer.
Do you have an example test for partitioning?

Ch3LL:
@tyl0re our documentation on our test suite is here: https://docs.saltstack.com/en/latest/topics/tutorials/writing_tests.html
There are currently tests here: tests/unit/modules/test_parted_partition.py that you can look at. The key is to write a test that fails before your fix and passes with it. Let me know if i can give more direction.

daks:
ref #54456

daks:
We were hit by the same bug, trying to create a non-fstype partition for LUKS. The proposed patch works fine, thanks @tyl0re

waynew:
@tyl0re I will make a couple of times this week available for showing how to write and (currently) run salt tests. I'll make sure to come back here and provide the calendar for it.

waynew:
@tyl0re I believe you should be able to schedule a time [on my calendar](https://calendar.google.com/calendar/selfsched?sstoken=UURWQ2dRZDlKU0twfGRlZmF1bHR8OWU3N2NmZjgwODY5ODVjNGU3NWRhMTRhYTI4MDM3ZmY)

If nothing works for you during that time period, let me know what days work best for you and I'll add some more time next week!

monofumado:
I forgot about this issue and updated to Salt 2019.2.2 and...the problem still happens, any news on this?

9numbernine9:
I just ran into this issue myself using Salt `2019.2.3`, but I'm assuming it's still an issue in `3000.0`. Is there any chance of this patch getting merged into `master`? As previously stated, it's basically impossible to create an LVM partition with `partition.mkpart` otherwise.

Ch3LL:
looks like this just needs a test with the associated fix and added to the master branch instead of develop. Anyone willing to write a test here to ensure this does not regress and add to master and we can get this merged in? I'm more than willing to help out here.

sagetherage (changes requested):
please rebase against the master branch

piterpunk:
@sagetherage , I just wrote the test for this fix and rebased it to master in #58213

sagetherage:
> @sagetherage , I just wrote the test for this fix and rebased it to master in #58213

great thank you! I will close this PR in favor of the new one rebased to master :)

PR #53900, opened 2019-07-18
Fix inconsistent full names in LGPO

twangboy:
### What does this PR do?
Fixes an issue with setting the `Allow Telemetry` group policy. Ensures the `_checkAllAdmxPolicies` function always returns full names when `return_full_policy_names=True`

Also adds some speed optimizations to the XPath queries.

Adds some tests.

https://github.com/saltstack/lock/issues/1826

### Previous Behavior
After setting the `Allow Telemetry` group policy, all subsequent group policy states would show `Allow Telemetry` also being changed.

The state module gets the current applied policies by running `lgpo.get`. This would return `AllowTelemetry`. After applying the state, the state module gets the current applied policies again and then compares the results with a `dict.diff`. The second `lgpo.get` would return `Allow Telemetry` (with a space). This was causing the state module to report an additional change.

### New Behavior
Since the returns from `_checkAllAdmxPolicies` are more consistent, subsequent state runs no longer report a change.

### Tests written?
Yes

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/modules/win_lgpo.py
+++ b/salt/modules/win_lgpo.py
@@ -47,6 +47,7 @@ import re
 import locale
 import ctypes
 import tempfile
+import time
 import uuid
 
 # Import Salt libs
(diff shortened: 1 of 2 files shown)


lomeroe (approved):
looks good

PR #53935, opened 2019-07-19
Power-off when shutting down FreeBSD, NetBSD, and OpenBSD

morganwillcock:
### What does this PR do?

Adds platform checks into the shutdown function of the system execution module, since FreeBSD, NetBSD, and OpenBSD have different shutdown results to other platforms.

These platforms don't power-off by default when halted, instead they prompt for input at the console. The system.shutdown currently uses 'shutdown -h' to shutdown these platforms,which means they are not powered off.

### Previous Behavior
Calling system.shutdown on a physical minion leaves the hardware powered on. The only way to restart it at this point is by being at the console or by using out-of-band power management.

Calling system.shutdown on a virtual minion leaves the VM using resources on the hypervisor. This can cause excessive CPU usage depending on how the hypervisor deals with the halt state, or lead to later failures when trying to programmatically start the minion (since it is technically still running).

### New Behavior
If the platform is FreeBSD, NetBSD, or OpenBSD, use 'shutdown -p' to shutdown, instead of the default 'shutdown -h'. This aligns with the system.shutdown result of other platforms: power off is requested.

### Tests written?
Yes?/~~No~~
~~I imagine this is probably out-of-scope for the test suite?~~

### Commits signed with GPG?

~~Yes~~/No

### DIFF
--- a/salt/modules/system.py
+++ b/salt/modules/system.py
@@ -122,7 +122,15 @@ def shutdown(at_time=None):
 
         salt '*' system.shutdown 5
     '''
-    cmd = ['shutdown', '-h', ('{0}'.format(at_time) if at_time else 'now')]
+    if (salt.utils.platform.is_freebsd() or
+            salt.utils.platform.is_netbsd() or
+            salt.utils.platform.is_openbsd()):
+        # these platforms don't power off by default when halted
+        flag = '-p'
+    else:
+        flag = '-h'
+
+    cmd = ['shutdown', flag, ('{0}'.format(at_time) if at_time else 'now')]
     ret = __salt__['cmd.run'](cmd, python_shell=False)
     return ret
 
--- a/tests/unit/modules/test_system.py
+++ b/tests/unit/modules/test_system.py
@@ -72,6 +72,46 @@ class SystemTestCase(TestCase, LoaderModuleMockMixin):
         '''
         Test to shutdown a running system
         '''
-        with patch.dict(system.__salt__,
-                        {'cmd.run': MagicMock(return_value='A')}):
+        cmd_mock = MagicMock(return_value='A')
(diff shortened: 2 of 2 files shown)


twangboy:
@morganwillcock There's actually a unit test for this... but it's pretty crappy. It needs to be changed to make sure the `cmd.run` function is called with the parameters you're expecting. You can do this with the `assert_called_once_with` assertion. There are many examples in salt code, for example:
```
        mock = MagicMock()
        with patch.dict(macpackage.__salt__, {'cmd.run': mock}):
            macpackage.install_app('/path/to/file.app/')
            mock.assert_called_once_with('rsync -a --delete "/path/to/file.app/" '
                                         '"/Applications/file.app"')
```

morganwillcock:
I've searched the documentation and I can't find an obvious answer...
How would I fake the platform to test both outcomes?

waynew:
Should be something like `with patch('salt.utils.platform.is_freebsd', MagicMock(return_value=True)):`. There may be other examples throughout the tests.

morganwillcock:
@twangboy is https://github.com/saltstack/salt/pull/53935/commits/05da4c5204ce72af9fb3140a6dd05c25abe4cf60 the sort of thing you would be expecting for the tests?

waynew:
@morganwillcock Yeah, I think that's reasonable

morganwillcock:
Should I submit this against a different branch?

waynew:
@morganwillcock Yes, please - against `master`. If you'd like, you can simply rebase these changes against master, force push, and then edit this PR.

Alternatively, you can close this PR and just open a new one. Either approach is fine.

morganwillcock:
@waynew I've rebased and switched the target to master

dwoz (approved):

PR #53936, opened 2019-07-20
Harcoded ARNS Bad: There are more partitions than standard.

thenoid:
### What does this PR do?

All throughout the boto salt code are hard coded strings starting with:
'arn:aws:'.  This means this code will only ever support the standard partition.

There are at least 4 'known' partitions in the world:
* standard - aws
* govcloud - aws-us-gov
* chinacloud - aws-cn
* D-O-Dcloud - ???

Limting salt's ability to be an enterprise grade tool to 1/4 partitions.

Additionally throughout the code are queries to 'ec2.get_user' and other services to return the accountID. Which leads to chicken and the egg scenarios and granting additional permissions beyond what is needed to function

Thankfully AWS provides the `sts` endpoint which allows you to ask the philosophical questions of 'who am i?', 'where am i?', but saddly not 'are you my mother?'.

This MR implements a 'boto_sts' module so that we can get introspective and patches the boto_iam module to make use of it.  There are close to a hundred additional hardcoded strings to be fixed, but for now this solves our use cases.

https://docs.aws.amazon.com/general/latest/gr/aws-arns-and-namespaces.html

Salt's ability to support partitions beyond standard.
Having to grant additional permissions when managing other services.

### Previous Behavior
Attempting manage any resource in any partition other than standard would result in failure.  In addition in the standard partition, you had to grant the IAM/LAMBDA module access to EC2 api endpoints....even if it wasn't managing EC2

### New Behavior
You can now create IAM/LAMBDA resources in non-standard partitions.  Also no longer have to grant superfluous permissions.

### Tests written?
No

### Commits signed with GPG?

No

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,6 +33,8 @@ Pipfile.lock
 /tests/cachedir/
 /tests/unit/templates/roots/
 /var/
+.direnv
+.envrc
 
 # setuptools stuff
 *.egg-info
--- a/.gitignore
+++ b/.gitignore
@@ -54,6 +54,7 @@ htmlcov/
 /*.iml
 *.sublime-project
 *.sublime-workspace
+.vscode
 
 # ignore ctags file
 tags
--- a/salt/modules/boto_iam.py
+++ b/salt/modules/boto_iam.py
@@ -1647,7 +1647,7 @@ def _get_policy_arn(name, region=None, key=None, keyid=None, profile=None):
     if name.startswith('arn:{0}:iam'.format(partition)):
         return name
 
-    return 'arn:{0}:iam::{1}:policy/{2}'.format(partition,account_id, name)
+    return 'arn:{0}:iam::{1}:policy/{2}'.format(partition, account_id, name)
 
 
 def policy_exists(policy_name,
--- a/salt/modules/boto_iam.py
+++ b/salt/modules/boto_iam.py
@@ -1283,10 +1283,33 @@ def build_policy(region=None, key=None, keyid=None, profile=None):
 
 
 def get_account_id(region=None, key=None, keyid=None, profile=None):
+    '''
+    Wrap boto_sts.get_account_id for backwards compatability with modules that may call
+    boto_iam.get_account_id
+
+    .. versionadded:: 2017.7.9
+
+    CLI Example:
+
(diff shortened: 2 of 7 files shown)


tkwilliams (changes requested):
I like the generalization of the ARN strings -- good catch!  :)

WRT the IAM->STS changes, I distinctly recall TRYING to go this route before (in fact the URL to that boto3 function call is still red, showing "recently visited", in my google search results); but IIRC, it behaved unreliably in different situations.  Note that the boto modules all support a variety of bind methods:  IAM instance creds, STS creds, key-pair, etc.  I don't remember exactly which, but the STS call failed with one or the other of these.  Please do verify all the possible auth methods before merge (hopefully updating the test cases appropriately).

I've not used the get_partition function, and it looks useful, but the same concerns would be there WRT whether these structures are actually populated in all authentication scenarios.  Please do verify, and create test-cases if possible.

Also, on a stupidly pedantic note, the comments still mention `get_user` even though you're using `get_caller_identity` :)

PS, grabbing the info from the FAILURE return data if it gets an error back?!?  GENIUS!!

thenoid:
So we have been using this code in production for over a year now, using all those authentication schemes you mentioned, with 0 issues.  the get_caller_identity function is there explicitly for asking `who am I`, which is nice because it doesn't require authorization, just authentication, to access.

Thanks for the comments catch - looking at the CI failures guess I need to become adept at writing salt tests.

dwoz:
@fattybenji This needs to be re-opened against the `master` branch and link to this PR. We are no longer accepting PRs to `2017.7.x`. Sorry for the confusion.

thenoid:
@dwoz @Ch3LL This PR sat for 5 months, passing tests, unmerged and now you are asking me to do MORE work to fix bugs in your product.  This is how you kill a community.

Ch3LL:
@thenoid we apologize for the delay. We have been doing a lot of planning to change our release and branch strategy to increase our release velocity, which unfortunately required taking a delay. All of the details of these changes are in this SEP here https://github.com/saltstack/salt-enhancement-proposals/pull/20 

I'm more than willing to help you port this to the master branch as well just let me know.

PR #53941, opened 2019-07-22
(Backport 53447) xfs: do not fails if type is not present

aplanas:
The command `blkid -o export` not always provides a 'TYPE' output
for all the devices. One example is non-formatted partitions, like for
example the BIOS partition.

This patch do not force the presence of this field in the blkid
output.

(cherry picked from commit 42e72265fef5a52cfcd3fe760ce9f1c0c7723ef3)

(backport #53447)

### DIFF
--- a/salt/modules/xfs.py
+++ b/salt/modules/xfs.py
@@ -329,7 +329,7 @@ def _blkid_output(out):
         for items in flt(dev_meta.strip().split("\n")):
             key, val = items.split("=", 1)
             dev[key.lower()] = val
-        if dev.pop("type") == "xfs":
+        if dev.pop("type", None) == "xfs":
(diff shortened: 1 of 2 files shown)


waynew (changes requested):
Hey @aplanas thanks for the PR! Would it be possible to toss together a small regression test for this?

Looks like this function is (fortunately) fairly minimal and the only things you might have to mock is `_get_mounts()` and `flt()`.

aplanas:
@waynew done, here and in https://github.com/saltstack/salt/pull/53447

waynew (approved):

aplanas:
Rebased on top of master in https://github.com/saltstack/salt/pull/53447

PR #53958, opened 2019-07-23
Loop until no eval

github-abcde:
### What does this PR do?
It provides an alternative to `loop.until` that does not use `eval`.
Added example for using this in the faq. Made some pylint-inspired changes.
I'm not entirely sure that the name of the new state `until_no_eval` is good enough (even though it is accurate), since it feels wrong to name a function with something that it does *not* have. I'm open to suggestions on this.

None

### Previous Behavior
The generic state that allows you to (blockingly) execute a salt function until it produces a desired output hinged on using `eval` with user-input. Since this is potentially very dangerous, an alternative has been supplied.

### New Behavior
A state exists (`loop.until_no_eval`) that will block execution until a specific salt execution module function returns an expected result. This produced result is passed together with the expected result to a function (either a method of python's `operator`, a function from `__salt__` or `__utils__`) to produce a boolean indicating success (or failure).
Additionally, a parameter `init_wait` is available to configure a waiting period before the 1st call of the salt function is made.

### Tests written?
Yes

### Commits signed with GPG?
Yes

### DIFF
--- a/doc/faq.rst
+++ b/doc/faq.rst
@@ -394,6 +394,35 @@ Restart the Minion from the command line:
     salt -G kernel:Windows cmd.run_bg 'C:\salt\salt-call.bat service.restart salt-minion'
     salt -C 'not G@kernel:Windows' cmd.run_bg 'salt-call service.restart salt-minion'
 
+Waiting for minions to come back online
+***************************************
+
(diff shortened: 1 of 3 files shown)


github-abcde:
Closed in favor of https://github.com/saltstack/salt/pull/55639

PR #53968, opened 2019-07-23
Backport #53192 into 2018.3

Ch3LL:
Backport #53192 into 2018.3


Akm0d (approved):

DmitryKuzmenko (approved):

DmitryKuzmenko:
2018.3 already contains disabling of the test. Just different case. It disables it if crypto is not available instead of checking the OS version. That works in the same way for windows where crypto is not available. So we don't need this anymore.

PR #53974, opened 2019-07-23
[2019.2.1] Add pygit2 requirement

Ch3LL:


garethgreenaway (approved):

dwoz (approved):

twangboy (approved):

waynew (approved):

twangboy (approved):

terminalmage:
This is likely a Bad Idea:tm:.

pygit2 has a number of non-Python dependencies, especially if you want SSH support. libffi, libssh, and libgit2 at minimum. _Then_ there's the fact that pygit2 is developed alongside libgit2, and its minor releases must track with the version of libgit2 which is installed. Assuming you have installed libgit2, if a 0.24.x libgit2 release is what you have installed (because that was what was available from the upstream repos), and pip installs pygit2 0.28.2.... in the immortal words of Thumper, "you're gonna have a bad time".

One other thing to consider is that the wheels built for pygit2 may or may not have been built against libssh2 (I haven't tested them), and if they haven't then installing them would leave you with a pygit2 that can't do ssh authentication.

Also, are these "static" requirements files only being used for the test suite?

Ch3LL:
Are you stating its a bad idea to use the pygit2 pip packages? Are you suggesting we only use system packages which is what we were doing previously in our test suite with our salt-jenkins states. All of the pygit2 tests are passing on this PR. 

 And yes these requirements files are only for tests. Without this change the pygit2 tests are not running.

terminalmage:
I just installed the wheel into a virtualenv. The good news is that the wheel bundles its own `libgit2.so`. That `libgit2.so` doesn't link against libssh2, but from the test results it seems that it's successfully authenticating via SSH, so it seems that they're maybe statically compiling libssh support into libgit2? If that's the case then this is probably fine. Didn't mean to needlessly alarm anyone but I've seen version differences between libgit2/pygit2 bite people before and that raised a big red flag for me.

Ch3LL:
thanks for pointing it out. Looks like we are okay for now, but something to be aware of in the future for the tests.

PR #53975, opened 2019-07-23
Fix a breaking typo in zonecfg module

ghost:
The zone net resource properties vlan-id and defrouter were combined into a single array element, which prevented Salt from setting either of them on Solaris and illumos zones.  This commit splits them into their correct seperate elements and allows Salt to set the properties.


Akm0d (approved):

ninefathom:
@Akm0d thanks!

PR #53987, opened 2019-07-24
Bring changes from 2019.2.1 PR #53735

twangboy:
https://github.com/saltstack/salt/pull/53735

### What does this PR do?
Try to account for the fact that the system time may be obtained just as the clock increments minutes
Make sure the `w32time` service is stopped by wrapping it in some asserts and checking the status of the service after stopping
Put the `service.stop` inside the `try/finally` block so that it will be restarted after if something fails

Skips tests that set date/time for now

This has been failing in PR tests

### Tests written?
Yes

### Commits signed with GPG?
Yes

### DIFF
--- a/tests/integration/modules/test_system.py
+++ b/tests/integration/modules/test_system.py
@@ -7,6 +7,7 @@ import logging
 import os
 import signal
 import subprocess
+import time
 import textwrap
 
 # Import Salt Testing libs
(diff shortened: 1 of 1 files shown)


dwoz (approved):

Akm0d (approved):

waynew (approved):

dwoz:
Merge without flaky PR tests

Issue #24338, opened 2015-06-02
tls module cert_info should convert extension objects into dicts


joejulian:
The tls module's cert_info function returns extensions as objects instead of as dicts. The results are not very functional. The extension tree should be walked and the objects converted to dicts.

jfindlay:
https://github.com/saltstack/salt/pull/24371/files#diff-3b99ecc50e324d73fab1c21caf844276R259

jfindlay:
@joejulian, thanks for the report.

joejulian:
Closing this to get it off my issues tab. After this long I doubt it's even valid anymore.

PR #53989, opened 2019-07-24
Fix tls cert unknown extensions

jleproust:
When analyzing a TLS certificate with unknown extensions, for example from
https://datanova.laposte.fr/, OpenSSL raises an OpenSSL.crypto.Error on
str(cert.get_extension()). The only useful thing to do with the
extension is then to print the content as a string.

No issue created yet. #24338 may be relevant but I'm not sure, it seems already fixed.

### Previous Behavior
Throws an uncaught OpenSSL.crypto.Error and fails

### New Behavior
Return relevant information about the certificate, with strings for unknown extensions, and no uncaught exception

### Tests written?
Yes

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/modules/tls.py
+++ b/salt/modules/tls.py
@@ -1642,7 +1642,12 @@ def cert_info(cert, digest='sha256'):
             try:
                 ext = cert.get_extension(i)
                 key = salt.utils.stringutils.to_unicode(ext.get_short_name())
-                ret['extensions'][key] = str(ext).strip()
+                value = None
+                try:
+                    value = str(ext)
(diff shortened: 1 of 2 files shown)


jleproust:
Hello, can we have an idea of the timeline for merging this? Thanks in advance.

jleproust:
Any news on this please? The fix is trivial, I added a unit test, and it's blocking for us. I'd really like to stop patching salt code in production. Forking and re-packaging shouldn't be necessary for this...

jleproust:
Ping?

dwoz:

PR #53990, opened 2019-07-24
auth: log details in case of user auth issue

eliasp:
When a local SaltClient sends a job as "root" while the Master runs as a non-root user, make it more clear why the message `Authentication failure of type "user" occurred.` is logged.

### What does this PR do?

Improve logging, to save the next one affected a few hours of debugging 😄 

None

### Previous Behavior

When the Master runs as non-root user, but `root` is used locally on the CLI for any kind of operations (`salt \* test.ping`, ...), no Job could be created. The only message that was logged without any further indication what might cause the failure was `Authentication failure of type "user" occurred.`

### New Behavior
An additional log message (level: `warning`) was added which states:
```
Master runs as "some-other-user", but user in payload is "root"
```

### Tests written?

No - doesn't really apply, as it's just an added log message.

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/auth/__init__.py
+++ b/salt/auth/__init__.py
@@ -330,7 +330,7 @@ class LoadAuth(object):
                 return auth_user.sudo_name()
             elif load['user'] == self.opts.get('user', 'root') or load['user'] == 'root':
                 if auth_key != key[self.opts.get('user', 'root')]:
-                    log.warning('Master runs as "{}", but user in payload is "{}"'.format(self.opts.get('user', 'root'), load['user']))
+                    log.warning('Master runs as %r, but user in payload is %r', self.opts.get('user', 'root'), load['user'])
                     log.warning(error_msg)
                     return False
             elif auth_user.is_running_user():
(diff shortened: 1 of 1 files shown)


salt/auth/__init__.py
```diff
@@ -330,6 +330,7 @@ def authenticate_key(self, load, key):
                 return auth_user.sudo_name()
             elif load['user'] == self.opts.get('user', 'root') or load['user'] == 'root':
                 if auth_key != key[self.opts.get('user', 'root')]:
+                    log.warning('Master runs as "{}", but user in payload is "{}"'.format(self.opts.get('user', 'root'), load['user']))
```
waynew:
```suggestion
                    log.warning('Master runs as %r, but user in payload is %r', self.opts.get('user', 'root'), load['user'])
```

Salt prefers log formatting for logs. I'm surprised that the linter didn't complain about this one :suspect: Usually it does.

waynew (changes requested):
Thanks for the PR!

This mostly looks great, just a minor change to the logging code.

waynew (approved):

twangboy (approved):

Akm0d (approved):

cmcmarrow (approved):

terminalmage:
Ported to master in https://github.com/saltstack/salt/pull/56692

PR #53992, opened 2019-07-24
cmdmod: fix runas and group in run_chroot

aplanas:
The parameters runas and group for cmdmod.run() will change the efective
user and group before executing the command. But in a chroot environment is
expected that the change happends inside the chroot, not outside, as the
user and groups are refering to objects that can only exist inside the
environment.

This patch add the userspec parameter to the chroot command, to change
the user in the correct place.

PR #53993, opened 2019-07-24
(Backport 53992) cmdmod: fix runas and group in run_chroot

aplanas:
The parameters runas and group for cmdmod.run() will change the efective
user and group before executing the command. But in a chroot environment is
expected that the change happends inside the chroot, not outside, as the
user and groups are refering to objects that can only exist inside the
environment.

This patch add the userspec parameter to the chroot command, to change
the user in the correct place.

(backport #53992)


cmcmarrow (approved):

aplanas:
Rebased on top of master as part of https://github.com/saltstack/salt/pull/53992

PR #53996, opened 2019-07-24
Fix kwargs iteration for chroot.call

aplanas:
### What does this PR do?

Fix a bug during the call of the kwargs parameters for chroot.call(). In the original code the iteration was done only on the keys, but the code expect the key and the value too. Simply adding `.items()` fix the issue.

### Tests written?

Yes

### DIFF
--- a/tests/unit/modules/test_chroot.py
+++ b/tests/unit/modules/test_chroot.py
@@ -150,7 +150,12 @@ class ChrootTestCase(TestCase, LoaderModuleMockMixin):
             utils_mock['thin.gen_thin'].assert_called_once()
             salt_mock['config.option'].assert_called()
             salt_mock['archive.tar'].assert_called_once()
-            salt_mock['cmd.run_chroot'].assert_called_once()
+            salt_mock['cmd.run_chroot'].assert_called_with(
+                '/chroot',
+                ['python{}'.format(sys.version_info[0]), '/tmp01/salt-call',
+                 '--metadata', '--local',
+                 '--log-file', '/tmp01/log', '--cachedir', '/tmp01/cache',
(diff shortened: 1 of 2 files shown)


tests/unit/modules/test_chroot.py
```diff
@@ -182,3 +182,35 @@ def test_call_success(self, mkdtemp, exist):
...
+        }
+        with patch.dict(chroot.__utils__, utils_mock), \
+                patch.dict(chroot.__salt__, salt_mock):
+            self.assertEqual(chroot.call('/chroot', 'ssh.set_auth_key',
+                                         user='user', key='key'), 'result')
+            utils_mock['thin.gen_thin'].assert_called_once()
+            salt_mock['config.option'].assert_called()
+            salt_mock['archive.tar'].assert_called_once()
+            salt_mock['cmd.run_chroot'].assert_called_once()
+            utils_mock['files.rm_rf'].assert_called_once()
```
waynew: I'm not sure that I understand - what is it that this is testing here?

aplanas: `test_call_success_parameters` is testing that the call succeed when args and kwargs are provided, that is where the bug was living.

Those asserts validate that the mocks are called, so we are sure that the full process takes place (create the salt-thin, the call to cmd.run_chroot, the clean at the end, etc)

waynew: I suppose that's reasonable. I'd be happier if we had a functional test (i.e. a test that calls `chroot.call('/chroot', 'ssh.set_auth_key', user='user', key='key')`, and it really creates the chroot & does the filesystem stuff), but it's *probably* fine to test like this 🤷‍♂

aplanas:
This patch is rebased in https://github.com/saltstack/salt/pull/54959

PR #53999, opened 2019-07-24
Generate new key each time for digital ocean key test

Ch3LL:
### What does this PR do?
The test `integration.cloud.clouds.test_digitalocean.DigitalOceanTest.test_key_management` sometimes fail. I believe it might be failing when we are running the test at the same time since it attempts to add the same key to digital ocean at the same time so there are conflicts.


garethgreenaway (approved):

twangboy (approved):

waynew (approved):

dwoz:
Merging without PR tests due to flakiness in the suite.

PR #54006, opened 2019-07-25
Acme state fixup 2018.3

github-abcde:
### What does this PR do?
(2018.3 version of https://github.com/saltstack/salt/pull/53366)
Edit: Expanded on the changes in this PR. Clarified the changes to behaviour in test-mode. I earlier stated that test-mode is blatantly ignored, which was incorrect.

It fixes the state `acme` by implementing correct handling of `__opts__['test']` and also properly detecting if a change is needed or not. It also fixes the returned changes (i.e. no changes returned if no changes were made, and only the changed items returned when there were changes.). Also now returns expected changes in test-mode.
Fixed handling of the `window`-parameter in `salt.modules.acme.needs_renewal` and thereby removing the broad `Exception` catching in `acme.cert`.
Moved check if certificate file actually exists into `salt.modules.acme.info`, removing 
Pylint now scores both the module and the state, as well as the unittest files with a 10 :)

None that I'm aware of.

### Previous Behavior
`acme.cert` always starts `salt.modules.acme.cert` regardless of whether a certificate renewal is needed.
No expected changes are reported in testmode, even when this is the case.
`acme.cert` always returns changes, even when no changes have been made (i.e. when a certificate was not due for renewal and no renewal had been forced).

### New Behavior
`acme.cert` does not start `salt.modules.acme.cert` when the certificate specified does not need renewal and no renewal has been forced. It also now properly returns actual changes instead of always dumping the entire certificate information for both the certificate prior to the call to `salt.modules.acme.cert` and after.
Running `acme.cert` with `__opts__['test'] == True` returns changes when a certificate would have been fetched or renewed.

### Tests written?
Yes

### Commits signed with GPG?
Yes

### DIFF
--- a/doc/topics/jinja/index.rst
+++ b/doc/topics/jinja/index.rst
@@ -1019,6 +1019,137 @@ Returns:
   d94a45acd81f8e3107d237dbc0d5d195f6a52a0d188bc0284c0763ece1eac9f9496fb6a531a296074c87b3540398dace1222b42e150e67c9301383fde3d66ae5
 
 
+.. jinja_ref:: set_dict_key_value
+
+``set_dict_key_value``
+----------------------
+
+..versionadded:: Neon
+
+Allows you to set a value in a nested dictionary without having to worry if all the nested keys actually exist.
+Missing keys will be automatically created if they do not exist.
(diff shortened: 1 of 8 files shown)


github-abcde:
Closing this in favor of https://github.com/saltstack/salt/pull/55589

Issue #54012, opened 2019-07-25
[Feature request] Archive.extracted to be able to remove top level directory


Oloremo:
### Description of Issue
I have the following workflow in the ansible:

1) Download archive md5 sum file.
2) Check local archive(if exists) against md5 in 1)
3) If it differs(or no old archive) - download the new archive on top of the old one
4) If we downloaded archive in 3) we check for a specific directory(with archive name) and if it exists - we delete it.
5) We unarchive archive, it will have the same name as we deleted at step 4)

The idea here is to release applications with a version like "**master**". It constantly updated and we can't use something like
`onlyif: path/to/file`

Since it can exist but it could be old. So we rely on the md5 sum and if it's changes - we overwrite old directory by deleting it first and unarchive after.

And since slots are not a thing yet in requisites I'm not particularly sure how to implement this logic in the current state of Salt.

What I tried:
`clean: True` argument - it's only removing the diff between current and new files based on file names.
`overwrite: True` - leads to the state to lose idempotency 
`prereq` - with `file.absent` - leads to the state to be run every time

### Versions Report
2019.02

waynew:
So you have an archive, say `master.zip` that you've deployed, and then you want to change that `master.zip` and then deploy it, but only if there are changes from the `master.zip` that exists on the minion?

Oloremo:
@waynew 
The idea here is to be able to deploy a new release with the same name. For example "master". Imagine we deployed it once and now we have a directory called "master" with the contents of this archive.

Now new commit was pushed to the master branch of something and the new artifact was created and we want to deploy it again. But now we have no means to check if the update is actually required since there is no sane version, just "master". So our approach here is to create a md5 sum file for each archive and if this md5 differs from the one we already have on a minion - we do a rollout\extraction.

So basically we use a 
`source_hash_update: True`

But the problem is that current arg `clean: True` is not suitable for that case since it just removes the diff between a local directory and the one that will be unarchived - based on filenames. And in our case filenames could be the same but they're could be very different,

So my proposal is to add the ability for a full cleanup for archive module. Not the diff between, but ensure that extraction will be done in the empty top directory.

waynew:
@Oloremo have you looked into the `onchanges` or `watch` requisites? They might allow you to do what you want without having to add new functionality.

Oloremo:
@waynew I looked at `prereq` with `file.absent` since I need to remove directory before the extraction. This leads to the state to be run every time.

It was something like this:
```
remove:
[...]

PR #54013, opened 2019-07-25
[WIP] Added clean_top argument for the archive state

Oloremo:
### What does this PR do?

Adding new argument `clean_top` to the archive state. This allows the state to remove top level directory before the extraction.

#54012 

### Previous Behavior
archive state was unable to do so by itself.

### New Behavior
archive state now able to remove top level directory if requested

### Tests written?

No(yet) - RFC required first

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/states/archive.py
+++ b/salt/states/archive.py
@@ -183,6 +183,7 @@ def extracted(name,
               force=False,
               overwrite=False,
               clean=False,
+              clean_top=False,
               user=None,
               group=None,
               if_missing=None,
@@ -525,6 +526,9 @@ def extracted(name,
             ``overwrite`` is set to ``True``.
 
         .. versionadded:: 2016.11.1
+
+    clean_top : False
+        Set this to ``True`` to remove a top level directory defore the extration.
 
     user
         The user to own each extracted file. Not available on Windows.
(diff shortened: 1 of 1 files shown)


Oloremo:
Folks, could you please look at the mentioned issue and in my implementation of this feature and tell if it's ok(both issue and implementation). If it's ok for you I'll write tests(not yet sure how)

salt/states/archive.py
```diff
@@ -526,6 +527,9 @@ def extracted(name,
 
         .. versionadded:: 2016.11.1
 
+    clean_top : False
+        Set this to ``True`` to remove a top level directory before the extration.
```
twangboy: extraction

salt/states/archive.py
```diff
@@ -1220,6 +1238,26 @@ def extracted(name,
             _add_explanation(ret, source_hash_trigger, contents_missing)
             return ret
 
+        if clean_top and contents is not None:
+            errors = []
+            log.debug('Removing directory %s due to clean_top set to True', name)
+            try:
+                salt.utils.files.rm_rf(name.rstrip(os.sep))
```
twangboy: Maybe rename the directory, if the archive.extract is successful, then delete, otherwise roll it back

twangboy (changes requested):

salt/states/archive.py
Oloremo: I'm not sure here.
First of all, it adds complexity + a hidden requirement for a 2x space(to keep both old and new dir at some point of time).
Second is that the current `clean: True` logic doesn't do it and I feel like it's more or less the same case.

Oloremo:
Closed in favor of https://github.com/saltstack/salt/pull/55418

PR #54017, opened 2019-07-25
Allow for main thread having terminated pid, before ThreadPoolExecutor threads

dmurphy18:
### What does this PR do?
Checks if pid exists before using psutil to kill any children of the pid

https://jenkinsci.saltstack.com/job/2019.2.1/job/salt-debian-9-py2/90/testReport/junit/unit.transport.test_zeromq/PubServerChannel/test_publish_to_pubserv_tcp

### Previous Behavior
Test would fail.

### New Behavior
test should pass.  The problem was that the ThreadExecutorPool executor.submit (4 of them) are handling the signal for the main thread too, which by the time they are called, is terminated.  Fix is to check if the pid exists before having psutil attempt to use it and kill any existing children.

### Tests written?
Yes, using existing tests, e.g. unit/transport/test_zeromq.py

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/utils/process.py
+++ b/salt/utils/process.py
@@ -807,15 +807,15 @@ class SignalHandlingMultiprocessingProcess(MultiprocessingProcess):
         msg += '. Exiting'
         log.debug(msg)
         if HAS_PSUTIL:
-            for process in psutil.process_iter():
-                if process == self.pid and hasattr(process, 'children'):
+            if psutil.pid_exists(self.pid):
+                process = psutil.Process(self.pid)
+                if hasattr(process, 'children'):
                     try:
                         for child in process.children(recursive=True):
                             if child.is_running():
                                 child.terminate()
                     except psutil.NoSuchProcess:
                         pass
-                    break
         sys.exit(salt.defaults.exitcodes.EX_OK)
 
     def start(self):
--- a/salt/utils/process.py
+++ b/salt/utils/process.py
@@ -1,4 +1,4 @@
-# -*- coding: utf-8 -*-
+# ndle-*- coding: utf-8 -*-
 '''
 Functions for daemonizing and otherwise modifying running processes
 '''
@@ -807,11 +807,15 @@ class SignalHandlingMultiprocessingProcess(MultiprocessingProcess):
         msg += '. Exiting'
         log.debug(msg)
         if HAS_PSUTIL:
-            process = psutil.Process(self.pid)
-            if hasattr(process, 'children'):
-                for child in process.children(recursive=True):
-                    if child.is_running():
-                        child.terminate()
+            for process in psutil.process_iter():
+                if process == self.pid and hasattr(process, 'children'):
+                    try:
+                        for child in process.children(recursive=True):
+                            if child.is_running():
+                                child.terminate()
+                    except psutil.NoSuchProcess:
+                        pass
+                    break
         sys.exit(salt.defaults.exitcodes.EX_OK)
 
     def start(self):
--- a/salt/utils/process.py
+++ b/salt/utils/process.py
@@ -815,9 +815,17 @@ class SignalHandlingMultiprocessingProcess(MultiprocessingProcess):
                             if child.is_running():
                                 child.terminate()
                         except psutil.NoSuchProcess:
-                            pass
+                            log.warn(
+                                'Unable to kill child of process %d, it does '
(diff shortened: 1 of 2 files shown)


salt/utils/process.py
```diff
@@ -807,11 +807,15 @@ def _handle_signals(self, signum, sigframe):
         msg += '. Exiting'
         log.debug(msg)
         if HAS_PSUTIL:
-            process = psutil.Process(self.pid)
-            if hasattr(process, 'children'):
-                for child in process.children(recursive=True):
-                    if child.is_running():
-                        child.terminate()
+            for process in psutil.process_iter():
```
twangboy: Couldn't you do
```
if psutil.pid_exists(self.pid):
    process = psutil.Process(self.pid)
    if hasattr(process, 'children')
    try:
        ...
```
Instead of looping over every process on the system?

dmurphy18: Thanks for the tip, was waiting to see results of tests before cleaning up. Will be sure to implement it that way

dmurphy18: done

salt/utils/process.py
```diff
@@ -1,4 +1,4 @@
-# -*- coding: utf-8 -*-
+# ndle-*- coding: utf-8 -*-
```
dwoz: Looks like some extra cruft got added here.

dwoz (changes requested):

salt/utils/process.py
dmurphy18: darn, will clean it

dmurphy18: done

salt/utils/process.py
```diff
@@ -807,11 +807,15 @@ def _handle_signals(self, signum, sigframe):
         msg += '. Exiting'
         log.debug(msg)
         if HAS_PSUTIL:
-            process = psutil.Process(self.pid)
-            if hasattr(process, 'children'):
-                for child in process.children(recursive=True):
-                    if child.is_running():
-                        child.terminate()
+            if psutil.pid_exists(self.pid):
```
rares-pop: I'd avoid checking for this, because it can mean nothing. What happens if gets stopped after the check?

I'd add the try-except block around the original code and that's it.

dmurphy18: adjusted try/except block

salt/utils/process.py
```diff
@@ -807,11 +807,15 @@ def _handle_signals(self, signum, sigframe):
...
+                try:
+                    process = psutil.Process(self.pid)
+                    if hasattr(process, 'children'):
+                        for child in process.children(recursive=True):
+                            if child.is_running():
+                                child.terminate()
```
rares-pop: Sorry for not mentioning this earlier, but maybe it's worth adding a try-catch here as well, if any children fail to terminate (due to race-conditions), to still try to terminate the rest of the child (not to stop on the first one that fails). Or maybe it's okay to stop on the first one?

dwoz: @rares-pop @dmurphy18 I don't think we need this `if psutil.pid_exists(self.pid):` at all. Just let it raise an exception if the parent doesn't exist. However, when iterating through the children I think we should except each one individually in case one goes away but others don't.

Maybe move this functionality to a utility method and add some unit tests for it.

dwoz (changes requested):

salt/utils/process.py
rares-pop: @dwoz - that was exactly my feedback too.  👍 

The first comment was about the unneeded pid_exists (I mentioned the possible race-condition), and the second about having the try-raise for children.

dmurphy18:
@rares-pop @dwoz While just executing the try/except and catching the exception is still functional code, testing and skipping is faster code than having to handle an exception case. I shall adjust it so each child is tried rather than exiting after the first one, as for making it a utility and associated unit tests, I believe that is over-complicating code.

twangboy (approved):

Ch3LL:
ping @rares-pop can you re-review here? thanks

waynew (approved):

Ch3LL:
ping @dmurphy18 some lint failures in the new tests.

rares-pop (approved):
LGTM!

waynew (approved):

dwoz (approved):

PR #54022, opened 2019-07-26
Some improvements and fixes to the freezer module

aplanas:
### What does this PR do?

Fix a bug when the module try to re-create the directory where the package information is stored, and add a new parameter in the restore function to clean the YAML document that contain this information after a successful recovery.

### Tests written?

Yes/No

### DIFF
--- a/salt/modules/freezer.py
+++ b/salt/modules/freezer.py
@@ -164,13 +164,13 @@ def freeze(name=None, force=False, **kwargs):
     safe_kwargs = clean_kwargs(**kwargs)
     pkgs = __salt__['pkg.list_pkgs'](**safe_kwargs)
     repos = __salt__['pkg.list_repos'](**safe_kwargs)
-    for name, content in zip(_paths(name), (pkgs, repos)):
-        with fopen(name, 'w') as fp:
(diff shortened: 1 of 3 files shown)


cmcmarrow (approved):

salt/modules/freezer.py
```diff
@@ -190,6 +190,9 @@ def restore(name=None, **kwargs):
     name
         Name of the frozen state. Optional.
 
+    clean
+        In True remove the frozen information YAML from the cache
+
```
twangboy: Please add a versionadded comment here for the new parameter. I think it's something like this:
```
.. version-added:: Sodium
```

twangboy (changes requested):
Should also probably add this to the release notes for sodium (create the file if it doesn't exist)

salt/modules/freezer.py
aplanas: I have the backport to Neon in https://github.com/saltstack/salt/pull/54023, do I need to put Neon just in case?

aplanas: I also added the release note in neon

aplanas:
Rebased on top of master as part of https://github.com/saltstack/salt/pull/54960

PR #54025, opened 2019-07-26
(Backport 54024) zypperpkg: fix pkg.list_pkgs cache

aplanas:
### What does this PR do?

The cache from pkg.list_pkgs for the zypper installer is too aggresive.
Some parameters will deliver different package lists, like root and
includes. The current cache do not take those parameters into
consideration, so the next time that this function is called, the last
list of packages will be returned, without checking if the current
parameters match the old one.

This patch create a different cache key for each parameter combination,
so the cached data will be separated too.

### Tests written?

Yes

(backport #54024)

### DIFF
--- a/salt/modules/zypperpkg.py
+++ b/salt/modules/zypperpkg.py
@@ -448,8 +448,14 @@ def _clean_cache():
     '''
     Clean cached results
     '''
+    keys = []
     for cache_name in ['pkg.list_pkgs', 'pkg.list_provides']:
-        __context__.pop(cache_name, None)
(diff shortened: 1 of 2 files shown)


aplanas:
@waynew sure thing! done!

waynew:
Py3 Fedora 29 build timed out - I just kicked off a rebuild, so hopefully that will get all squared away 👍 

If not, that's kind of promising because we've been hunting this problem for a while and maybe we can get some insight from this particular process 😁

PR #54048, opened 2019-07-29
Avoid traceback on "http.query" when there are errors with the requested URL

meaksh:
### What does this PR do?
This PR prevents traceback on `http.query` and improves error logging in cases where there are errors with the requested URL:

- Connection refused
- No route to host
- Errors with DNS servers

### Previous Behavior
In cases where the requested URL refuses the connection:

```
# salt-call --local http.query http://127.0.0.1/
  File "salt-call", line 27, in <module>
    salt_call()
  File "pyall/salt/scripts.py", line 431, in salt_call
    client.run()
  File "pyall/salt/cli/call.py", line 57, in run
    caller.run()
  File "pyall/salt/cli/caller.py", line 138, in run
    ret = self.call()
  File "pyall/salt/cli/caller.py", line 237, in call
  File "/var/tmp/.root_fc9778_salt/pyall/salt/executors/direct_call.py", line 12, in execute
    return func(*args, **kwargs)
  File "/var/tmp/.root_fc9778_salt/pyall/salt/modules/http.py", line 41, in query
    return salt.utils.http.query(url=url, opts=opts, **kwargs)
  File "pyall/salt/utils/http.py", line 566, in query
    result = download_client.fetch(url_full, **req_kwargs)
  File "py2/tornado/httpclient.py", line 102, in fetch
    self._async_client.fetch, request, **kwargs))
  File "py2/tornado/ioloop.py", line 457, in run_sync
    return future_cell[0].result()
  File "py2/tornado/concurrent.py", line 237, in result
  File "<string>", line 3, in raise_exc_info
error: [Errno 111] Connection refused
```

### New Behavior
```
# salt-call --local http.query http://127.0.0.1/
[ERROR   ] Cannot perform 'http.query': http://127.0.0.1/ - [Errno 111] Connection refused
local:
    ----------
    error:
        [Errno 111] Connection refused
```

### Tests written?

No - This PR doesn't include unit tests to cover this small issue since the whole `utils.http.query` is currently not being tested and complexity is really high (function is bigger than 500 LOC).

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/utils/http.py
+++ b/salt/utils/http.py
@@ -568,7 +568,7 @@ def query(url,
             ret['status'] = exc.code
             ret['error'] = six.text_type(exc)
             return ret
-        except socket.gaierror as exc:
+        except (socket.herror, socket.error, socket.timeout, socket.gaierror) as exc:
             if status is True:
                 ret['status'] = 0
             ret['error'] = six.text_type(exc)
--- a/salt/utils/http.py
+++ b/salt/utils/http.py
@@ -567,11 +567,13 @@ def query(url,
         except tornado.httpclient.HTTPError as exc:
             ret['status'] = exc.code
             ret['error'] = six.text_type(exc)
+            log.error("Cannot perform 'http.query': {0} - {1}".format(url_full, ret['error']))
             return ret
         except (socket.herror, socket.error, socket.timeout, socket.gaierror) as exc:
             if status is True:
                 ret['status'] = 0
             ret['error'] = six.text_type(exc)
+            log.error("Cannot perform 'http.query': {0} - {1}".format(url_full, ret['error']))
             return ret
 
         if stream is True or handle is True:
--- a/salt/utils/http.py
+++ b/salt/utils/http.py
@@ -567,7 +567,6 @@ def query(url,
         except tornado.httpclient.HTTPError as exc:
             ret['status'] = exc.code
             ret['error'] = six.text_type(exc)
-            log.error("Cannot perform 'http.query': {0} - {1}".format(url_full, ret['error']))
             return ret
         except (socket.herror, socket.error, socket.timeout, socket.gaierror) as exc:
             if status is True:
--- a/salt/utils/http.py
+++ b/salt/utils/http.py
@@ -572,7 +572,7 @@ def query(url,
             if status is True:
                 ret['status'] = 0
             ret['error'] = six.text_type(exc)
-            log.error("Cannot perform 'http.query': {0} - {1}".format(url_full, ret['error']))
+            log.debug("Cannot perform 'http.query': {0} - {1}".format(url_full, ret['error']))
             return ret
 
         if stream is True or handle is True:


salt/utils/http.py
```diff
@@ -567,11 +567,13 @@ def query(url,
         except tornado.httpclient.HTTPError as exc:
             ret['status'] = exc.code
             ret['error'] = six.text_type(exc)
+            log.error("Cannot perform 'http.query': {0} - {1}".format(url_full, ret['error']))
```
rares-pop: I'd avoid logging these, because you do return the errors. Passing an invalid URI is expected too.

salt/utils/http.py
```diff
@@ -567,11 +567,13 @@ def query(url,
...
             return ret
-        except socket.gaierror as exc:
+        except (socket.herror, socket.error, socket.timeout, socket.gaierror) as exc:
             if status is True:
                 ret['status'] = 0
             ret['error'] = six.text_type(exc)
+            log.error("Cannot perform 'http.query': {0} - {1}".format(url_full, ret['error']))
```
rares-pop: diito

salt/utils/http.py
meaksh: Thanks for your review @rares-pop, IMHO it does make sense to add this `log.error` message here. There are few reasons for that:

- It keeps consistency with the rest of the `http.query` method, where every time an error is produce, we return with `ret['error']` and also log the error in the minion log:

https://github.com/saltstack/salt/blob/310a30e8648ccc52d5bc91f0cda5e477821cbae3/salt/utils/http.py#L207-L208
https://github.com/saltstack/salt/blob/310a30e8648ccc52d5bc91f0cda5e477821cbae3/salt/utils/http.py#L525-L526
https://github.com/saltstack/salt/blob/310a30e8648ccc52d5bc91f0cda5e477821cbae3/salt/utils/http.py#L677-L678

- It also makes this situation easier for debugging since the error message is kept on the minion logs and not only displayed on the return (which can be lost or covered in case you're not directly calling this method, i.a. as part of Jinja inside of an SLS file).

xeacott: I am also in favor of consistency in this case.

xeacott (approved):

salt/utils/http.py
rares-pop: @meaksh, @xeacott - this is an execution module. If you pass invalid parameters, it will return an error.

Logging is for when something internally went wrong.

How is that logging useful in any way? You don't even have to debug that code, because it returns the error immediately. You only have to check the status of the function.

To be consistent you would have to remove all the logs from there, or at least downgrade them to log.debug but I fail to see any meaning in those.

waynew: I agree with @rares-pop here - if you get a 400 or 403 or 500, presumably each one of those will show up in this log message. That's not terribly useful to dump to a log.

meaksh: Thanks for the discussion here. @rares-pop @waynew I see your point here, and I agree that when we get a HTTP Error from the remote hosts, then it would be ok not logging it.

However, IMHO I would keep the `log.error` when a `socket` exception is raised since, on that case, the HTTP request is not even executed due some error. Then I consider useful to produce an error message that reaches the minion log files.

I'm removing logging for the `HTTPError` exception.

rares-pop: Yes, the HTTP request is not executed in that case, but the function will return saying it wasn't executed and why. I still think it's pointless to log it. Logging is useful when you don't know what happens, to avoid silent fails. But here you do. The function will return the exact error you want to duplicate in the log file.

meaksh: @rares-pop, what happens when you're calling this inside a Jinja? For example, I have an SLS file which does something like:

```
{% if salt['http.query'](my_url, status=True).get('status') == 200 %}
...
{% endif %}
```

In this case, if something happens on that `http.query` execution and we don't log anything, it will not be easy to know what happened. Neither looking at the minion logs.

rares-pop: I'm not an expert on jinja, but that doesn't look very good.

You should be doing more like:
`ret = salt['http.query'](my_url, status=True)`
`if (ret.get('status') == 200)`:
`   ...`
`else`
`  log the error % ret`

The function is returning everything you need. If you choose to cherry-pick certain info, then you will loose details. But you shouldn't rely on the log to tell you what you ignored.

If I'm the only one who see things in this way, please disregard my comment.

xeacott: After looking back into this, I do agree with @rares-pop here. The `return` should have everything useful about what happened during the call and its given back to you.

xeacott:
Hey guys @rares-pop @meaksh let's find a good resolution for this PR. If im reading this correctly, it sounds like these changes aren't really necessary to be added?

meaksh:
The changes introduced by this PR are necessary, but the `log.error` might be something to discuss. I'd also agree on showing them as "DEBUG" instead of "ERROR" if that makes the consensus.

waynew:
I have no problem with the messages being DEBUG level. If you're not getting information back from your master, you're probably going to want to be debugging anyways. And that will prevent it from noising up the logs for cases that could be entirely reasonable.

rares-pop (approved):
LGTM

meaksh:
Closing in favor of https://github.com/saltstack/salt/pull/55586

Issue #52764, opened 2019-04-30
VMWare Driver does not support tag info


doesitblend:
### Description of Issue/Question
The current cloud vmware driver does not support retrieving information about tags.
Example `salt-cloud -a show_instance <instancename>` does not display the information about tags. The instance when viewing in vSphere does have tags. If we use the same command in AWS tag info is retrieved and displayed. The tag values are crucial for managing the cloud infra structure.

### Setup
(Please provide relevant configs and/or SLS files (Be sure to remove sensitive info).)

### Steps to Reproduce Issue
(Include debug logs if possible and relevant.)

### Versions Report
Version info:
Salt: 2018.3.4
Vsphere: 6.5

doesitblend:
ZD-3633

waynew:
https://vdc-download.vmware.com/vmwb-repository/dcr-public/6b586ed2-655c-49d9-9029-bc416323cb22/fa0b429a-a695-4c11-b7d2-2cbc284049dc/doc/vim.ManagedEntity.html#tag

According to this doc it seems like that might be (currently) an experimental feature. Not sure if it's supported in pyvmomi yet?

rdutch:
The set of tags associated with this managed entity. Experimental. Subject to change.
**Since vSphere API 4.0**
I would not call this Experimental

doesitblend:
Reviewing the current version of pyvmomi (master branch) the tests pull an object containing a key tag.

```
[...trimmed for brevity..]
"dynamicProperty": [], "dynamicType": null, "key": "host"}, "tag": [], "triggeredAlarmState": [],
[...trimmed for brevity..]
```

xeacott:
Upon much investigation, VMWare PyVmomi does not support tagging. This has been stated by tianhao64 in a few github issues that have been opened in pyvmomi, for example:
 
- [Issue 307](https://github.com/vmware/pyvmomi/issues/307)
- [Issue 733](https://github.com/vmware/pyvmomi/issues/733) 

To support this feature, tagging requires vSphere Automation SDK (formerly known as vCloud Suite SDK, which can be viewed [here)](https://code.vmware.com/web/sdk/60/vcloudsuite-python?h=VMware%20vCloud%20Suite%20SDK%20for%20Python) of version 6.0 at minimum. This is a feature that is being worked on at the moment and will be implemented using the new SDK suite.

xeacott:
[...]

PR #54058, opened 2019-07-29
New tagging vsphere feature

xeacott:
### What does this PR do?
Adds tagging ability to vcenter proxy through new functions.
.
.

https://github.com/saltstack/salt/issues/52764

### Tests written?
Yes

### Commits signed with GPG?

Yes


tests/unit/modules/test_vsphere.py
```diff
@@ -1918,3 +1948,728 @@ def test_vcenter_proxy_return(self):
...
+                                  'get_service_instance',
+                                  return_value=None
+                                  ) as get_service_instance:
+                    with patch.dict(vsphere.__salt__,
+                                    {'vcenter.get_details': get_details},
+                                    clear=True) as get_vcenter_details:
+                        with patch.object(salt.utils.vmware,
+                                          'get_vsphere_client',
+                                          return_value=None
+                                          ) as get_vsphere_client:
```
waynew: FWIW you can write this like:

```
patch_proxy_type = patch.object(vsphere, 'get_proxy_type', ...)
patch_proxy_connection = patch.object(...)

with patch_proxy_type as get_proxy_type:
    with patch_proxy_connection as get_proxy_connection:
        ...
```

I think you can also write it as

```
with patch_proxy_type as get_proxy_type, \
        patch_proxy_connection as get_proxy_connection, \
        ...:
    ...
```

If that feels cleaner

waynew (approved):

waynew (changes requested):
Looks like there's a bunch of lint issues 😛

cmcmarrow (approved):
Nice work Joe!

tests/unit/modules/test_vsphere.py
xeacott: I may, simply because these lines are SO long and I would prefer a cleaner way of writing it!

twangboy (approved):

requirements/tests.txt
```diff
@@ -20,6 +20,7 @@ supervisor; python_version < '3'
 kubernetes<4.0
 psutil
 pyvmomi
+vSphere-Automation-SDK
```
s0undt3ch: This should instead be added to `requirements/static/*.in` as we talked about, likely where `pyvmomi` is also present, and then the requirements need to be recompiled, `pre-commit run -av`

xeacott: Yep, I'll get that going...

cro (approved):

xeacott:
Changes are waiting on a response from this thread: https://github.com/vmware/vsphere-automation-sdk-python/issues/38
@s0undt3ch this is the thread from which our previous conversation originates from - the outcome of this will determine the fate of the installation process :)

Akm0d:
re-run full tornado

Akm0d:
re-run full pycryptodomex

Akm0d:
re-run full macosxmojave

dwoz (approved):

salt/modules/vsphere.py
```diff
@@ -9050,6 +9094,492 @@ def _delete_device(device):
...
+        # User didn't provide CLI args so use proxy information
+        details = __salt__['vcenter.get_details']()
+        server = details['vcenter']
+        username = details['username']
+        password = details['password']
+
+    # Establish connection with client
+    client = salt.utils.vmware.get_vsphere_client(server=server,
+                                                  username=username,
+                                                  password=password)
```
s0undt3ch: Getting a client instance should be factored in it's own private function, since its used more than once.

```python
def _get_client(server, username, password):
    return blah
```
or
```python
def _get_client(details):
    server = details['vcenter']
    .... etc ....
    return client_instance
```
No?

s0undt3ch (changes requested):

salt/modules/vsphere.py
xeacott: Fixing that now...

salt/modules/vsphere.py
```diff
@@ -218,6 +238,30 @@
 except ImportError:
     HAS_PYVMOMI = False
 
+# vSphere SDK Automation
+try:
+    # pylint: disable=unused-import
```
s0undt3ch: Should be:
```python
# pylint: disable=unused-import
... a bunch of imports
# pylint: enable=unused-import
```
Or
```python
import blah  # pylint: disable=unused-import
```

Ie, inline disabled are ok, but if the disable is on its own line, you've disabled the rule for the rest of the file. Which is ok if that's the purpose, if not, you then need to enable the rule again.

s0undt3ch (changes requested):
The PyLint disable/enable is the only thing to point out. The rest is good.

salt/modules/vsphere.py
xeacott: Understood, I'll drop a change for that by end of day. Thanks for clarifying!

xeacott: Should be updated now, if there's no other changes I'll get this asked to be merged/ 👍

s0undt3ch (approved):

waynew (approved):

waynew:
You may want to rebase this on master.

dwoz:
Windows tests are passing but the builds failed.

PR #54061, opened 2019-07-30
Skip pygit2 fileserver tests on Windows

twangboy:
### What does this PR do?
Skips the pygit2 fileserver tests on Windows due to issues with file locks not being released by git.

This is being done on 2019.2.1 via https://github.com/saltstack/salt/pull/53974

### Tests written?
Yes

### Commits signed with GPG?
Yes

### DIFF
--- a/tests/unit/fileserver/test_gitfs.py
+++ b/tests/unit/fileserver/test_gitfs.py
@@ -500,6 +500,7 @@ class GitPythonTest(GitFSTestBase, GitFSTestFuncs, TestCase, LoaderModuleMockMix
 
 @skipIf(not HAS_GITPYTHON, 'GitPython >= {0} required for temp repo setup'.format(GITPYTHON_MINVER))
 @skipIf(not HAS_PYGIT2, 'pygit2 >= {0} and libgit2 >= {1} required'.format(PYGIT2_MINVER, LIBGIT2_MINVER))
+@skipIf(salt.utils.platform.is_windows(), 'Skip Pygit2 on windows, due to pygit2 access error on windows')
 @skipIf(NO_MOCK, NO_MOCK_REASON)
 class Pygit2Test(GitFSTestBase, GitFSTestFuncs, TestCase, LoaderModuleMockMixin):
 


Ch3LL (approved):

cmcmarrow (approved):

garethgreenaway (approved):

cachedout:
Is pygit2 support on Windows being dropped?

waynew:
FWIW (from what I understand) pygit/gitfs are both broken, leaking memory like a sieve. From what I gather that's something that's fundamental to the way both libraries work and would basically require re-writing them. I could be wrong!

This (along with many others) is a temporary skip as we focus on stabilizing our test suite and automating our entire build process.

Issues will be addressed, and users will be informed if we change any kind of behavior 👍

terminalmage:
> FWIW (from what I understand) pygit/gitfs are both broken, leaking memory like a sieve.


`[citation needed]`

waynew:
https://github.com/libgit2/pygit2/issues/625 https://github.com/gitpython-developers/GitPython/issues/289

I also know that the reason that we're skipping tests right now are because the file handles are still open - it may be that we're doing something improperly in Salt, or it may be something in the underlying library, but the fact is that that the test suite is failing because file handles are open and Windows doesn't like it when you try to delete open files, or access open files.

Fixing it is important to do, but not the *most important* thing on our plate... yet. So we're squelching the noise from this particular test failure until we can address the problem.

Of course, if this particular problem is an itch for you or anyone else, we gladly welcome contributions! Otherwise, we'll devote time to it once we've addressed the more pressing issues.

terminalmage:
So, a few notes on this.

1. The issue you linked is from GitPython, not Pygit2. Can you explain how this serves as a justification for disabling Pygit2 tests?

2. That issue is also 4 years old, and from before a major reworking of GitPython, so its explanatory power seems dubious at best.

3. Can you share the profiling work that has been done which led you to the conclusion that Pygit2 "leaks memory like a sieve"?

terminalmage:
I noticed that there was a Pygit2 issue linked in your comment as well. However, this issue is:

1. 3 years old
2. from 4 release cycles ago
3. for a method that salt _doesn't even use_

waynew:
It's entirely possible that I'm wrong about the underlying git libraries - I haven't performed exhaustive tests.

I know that *this* test is failing, and it's failing in a way that indicates file handles are still open when we're trying to perform cleanup.

I can't find the Salt issue where I did some profiling so I can't confirm the facts, but there was some issue around resource usage, and from what I recall of the process when using either backend for gitfs the memory usage grew and grew - exceeding hundreds of MB over 24h, with the master just idling. At the time, when I searched for reasons why, I came across issues on both pygit2 and gitpython issue trackers in regards to memory leaks, and the impression that I got from what I read is that the fundamental architecture of both libraries led them to be susceptible to memory leaks if they were used within long-running processes.

Is that an inaccurate impression? I don't know. I have seen with gitfs, Salt has a memory leak. I have not seen a memory leak without gitfs (or, if there is one it's not nearly as intense as with gitfs).

----

Ultimately, though, I feel like discussion has turned to a discussion for discussion's sake. Correct me if I'm wrong, but your concern is that we're skipping these tests, and though I didn't notice it stated outright, there's probably a concern that the tests will **remain** skipped on Windows, leading to instability and other problems. Perhaps even a sense that we're just skipping the tests because that's the easy thing to do, rather than solve the underlying problem. Is that correct?

If that is the concern, hopefully it will allay your fears to know that we are only **TEMPORARILY** skipping the test(s) as part of a triage process which will only be completed once these tests are either fixed and re-enabled (best case), or some other satisfactory solution.

twangboy:
The problem is that pygit2 doesn't release the file handles for the `.idx` files and when the tests try to clean up they fail.

terminalmage:
> It's entirely possible that I'm wrong about the underlying git libraries - I haven't performed exhaustive tests.
> 

This being the case, I wish you would refrain from matter-of-fact pejoratives ("broken", "leaking memory like a sieve") with regard to Python modules you admittedly don't understand.

> I can't find the Salt issue where I did some profiling so I can't confirm the facts, but there was some issue around resource usage, and from what I recall of the process when using either backend for gitfs the memory usage grew and grew - exceeding hundreds of MB over 24h, with the master just idling. At the time, when I searched for reasons why, I came across issues on both pygit2 and gitpython issue trackers in regards to memory leaks, and the impression that I got from what I read is that the fundamental architecture of both libraries led them to be susceptible to memory leaks if they were used within long-running processes.
> 
> Is that an inaccurate impression? I don't know. I have seen with gitfs, Salt has a memory leak. I have not seen a memory leak without gitfs (or, if there is one it's not nearly as intense as with gitfs).
> 

This is confusing. Running a profiler tool would show you what is using the memory, what methods are called, and other useful information. You shouldn't _need_ to scour the internet for possible explanations (as it seems you did with the two links you posted earlier, neither of which correspond to code that Salt actually uses).

What profiler did you use? [cProfile](https://docs.python.org/2/library/profile.html)?

> Ultimately, though, I feel like discussion has turned to a discussion for discussion's sake.

Perhaps, though I think you _invited_ the discussion by making unsupported claims. :man_shrugging:

> Correct me if I'm wrong, but your concern is that we're skipping these tests, and though I didn't notice it stated outright, there's probably a concern that the tests will **remain** skipped on Windows, leading to instability and other problems. Perhaps even a sense that we're just skipping the tests because that's the easy thing to do, rather than solve the underlying problem. Is that correct?
> 
> If that is the concern, hopefully it will allay your fears to know that we are only **TEMPORARILY** skipping the test(s) as part of a triage process which will only be completed once these tests are either fixed and re-enabled (best case), or some other satisfactory solution.

I hope so, though the discussion in this thread doesn't inspire much in the way of confidence.

cachedout:
> If that is the concern, hopefully it will allay your fears to know that we are only TEMPORARILY skipping the test(s) as part of a triage process which will only be completed once these tests are either fixed and re-enabled (best case), or some other satisfactory solution.

The concern of the community, as I understand it, is that details surrounding this approach have not been communicated well. I suspect that people would be very happy to know the following:

1. Are new PRs being reviewed and merged during this test triage process? What experience should a submitter expect while this process is ongoing?

2. What. is the state of releases? Are they going forward as planned? Are there target dates? What resource can Salt developers look at to know what. stage of the release a. given branch is at without having to ask somebody employed by SaltStack?

3. Are all releases suspended until the temporary skips are addressed or is the plan to do releases with the skips. in place and fix them incrementally over time?

I think that if you can find a way to communicate answers to the above, many in the community will be pleased to try to help however they can. Thanks!

PR #54068, opened 2019-07-30
pillar_stack: add *fetch* method as an alias to salt.utils.data.traverse_dict_and_list

n-rodriguez:
### What does this PR do?

Add *fetch* method as an alias to salt.utils.data.traverse_dict_and_list in pillar stack.

### Tests written?

No

### Commits signed with GPG?

No

### DIFF
--- a/salt/pillar/stack.py
+++ b/salt/pillar/stack.py
@@ -438,6 +438,7 @@ def _process_stack_cfg(cfg, stack, minion_id, pillar):
             },
         "minion_id": minion_id,
         "pillar": pillar,
+        "fetch": salt.utils.data.traverse_dict_and_list,
         })
     for item in _parse_stack_cfg(
             jenv.get_template(filename).render(stack=stack)):


n-rodriguez:
This simple addition allows to easily fetch pillars in pillar stack :

```jinja
{% set foo = fetch(stack, 'foo:bar:baz', []) %}
{% set bar = fetch(pillar, 'bar:bar:baz', {}) %}
{% set baz = fetch(__grains__, 'baz:bar:baz', false) %}
```

n-rodriguez:
ping @dwoz

n-rodriguez:
hi there! any news?

waynew:
Hi @n-rodriguez thanks for the PR! We like this feature, but we would also like a test or two, just something that ensures that the filter is present in the jinja environment. I was looking around, I'm not sure but I *think* that `tests/support/jinja_filters.py` would be the appropriate place to add a test for that. @garethgreenaway or @dwoz - do either of you have any other recommendations?

n-rodriguez:
@waynew thank for your reply! actually I was thinking about another approach/naming to match with the existing [`traverse`](https://docs.saltstack.com/en/latest/topics/jinja/index.html#traverse)

n-rodriguez:
>  I was looking around, I'm not sure but I _think_ that `tests/support/jinja_filters.py` would be the appropriate place to add a test for that. @garethgreenaway or @dwoz - do either of you have any other recommendations?

Hi there any news?

PR #54073, opened 2019-07-31
Fix deprecation warnings for msgpack > 0.5.2

angeloudy:
I am getting the following warnings
```
09:58:35,223 [WARNING ] /usr/local/lib/python3.6/site-packages/salt/transport/ipc.py:280: DeprecationWarning: encoding is deprecated, Use raw=False instead.
  self.unpacker = msgpack.Unpacker(encoding=encoding)
```
This commit fixed the problem

### DIFF
--- a/salt/transport/ipc.py
+++ b/salt/transport/ipc.py
@@ -265,7 +265,12 @@ class IPCClient(object):
             encoding = None
         else:
             encoding = 'utf-8'
-        self.unpacker = msgpack.Unpacker(encoding=encoding)
+        unpack_args = {}
+        if msgpack.version >= (0, 5, 2):
+            unpack_args['raw'] = False
+        else:
+            unpack_args['encoding'] = encoding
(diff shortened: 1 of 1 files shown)


waynew:
We have an [existing PR](https://github.com/saltstack/salt/pull/53310) that will fix the problem across the board. However, every single message from Salt master <-> minion gets piped through this transport layer, and while we have enough tests to give me a 90% confidence level that this won't break everything...

Given the potential risk, I would like to get more complete tests in place ensuring that these changes will do nothing wrong when it comes to Salt's messages.

Unfortunately, I've got some other, higher-priority items on my plate currently so I haven't been able to devote the time I would like to getting that fixed. If you're interested, more complete transport level tests would be most welcome!

waynew:
@angeloudy I'm going to go ahead and close this PR, since it's pointing to `develop`, and @Akm0d I know made some changes to this recently on `master`. If you have any questions, please feel free to reach out!

Issue #25915, opened 2015-07-31
FreeBSD pkg install fails


ari:
Installing some packages fails without any clues in the logs at to why. On the command line, this succeeds:

```
pkg install  -y databases/py-MySQLdb
```

This is FreeBSD 10.1-RELEASE-p15.

However ```salt-call -l debug state.highstate``` fails with a state like this:

```
salt-mysql:
  pkg.installed:
    - name: databases/py-MySQLdb
```

Here's the relevant snippet from the logs. Everything looks like it worked perfectly, except for the ERROR on the last line.
```
[INFO    ] Running state [databases/py-MySQLdb] at time 11:02:07.503711
[INFO    ] Executing state pkg.installed for databases/py-MySQLdb
[DEBUG   ] Could not LazyLoad pkg.normalize_name
[DEBUG   ] Could not LazyLoad pkg.check_db
[DEBUG   ] Could not LazyLoad pkg.normalize_name
[INFO    ] Executing command 'pkg install  -y databases/py-MySQLdb' in directory '/root'
[INFO    ] Executing command 'pkg info -ao' in directory '/root'
[DEBUG   ] Could not LazyLoad pkg.hold
[ERROR   ] Installed Packages:
py27-MySQLdb changed from absent to 1.2.5
```
I get a very similar error if the package is already installed, so it appears to be an issue with the parsing of ```pkg info -ao``` but I can't see what:

```
# pkg info -ao | grep py-MySQLdb
py27-MySQLdb-1.2.5             databases/py-MySQLdb
```

This is the salt environment on the minion.
```
           Salt: 2015.5.3
         Python: 2.7.10 (default, Jul 10 2015, 21:46:13)
         Jinja2: 2.7.3
       M2Crypto: 0.22
 msgpack-python: 0.4.2
   msgpack-pure: Not Installed
       pycrypto: 2.6.1
        libnacl: Not Installed
         PyYAML: 3.11
          ioflo: Not Installed
          PyZMQ: 14.6.0
           RAET: Not Installed
            ZMQ: 4.1.2
           Mako: Not Installed
        Tornado: Not Installed
```

jfindlay:
@ari, thanks for the report.  @cedwards, @kim0, any ideas here?

amontalban:
I'm facing the same issue, I think the problem is when you use `/` in the package name:

```yaml
java:
  pkg.installed:
    - name: java/openjdk7
```
```yaml
ID: java
Function: pkg.installed
[...]

PR #54074, opened 2019-07-31
Make sure to strip the origin and flatten the dict before calling _…

gmsoft-tuxicoman:
…parse_version_string().

### What does this PR do?

This fixes an issue occurring when using pkg.latest on FreeBSD. On that OS, each versionis returned as a dict with "origin" and "version". This get rid of the dict before calling _parse_version_string().

This behavior is similar to what is done in _fulfills_version_string().

### Previous Behavior
pkg.latest() fails complaining that version "origin" doesn't match version "version".

### New Behavior
pkg.latest() works.


asomers:
IMHO this isn't the correct solution to the problem, because it doesn't fix the divergence in datatypes within pkg.py.  Even with this patch, much of pkg.py would be left processing one type of dict on FreeBSD, and another on Linux.  A better solution would be to make the format of `pkg.list_pkgs` uniform across OSes.

On Linux it would look like this: `{'version' = '1.1.0' }` and on FreeBSD `{'origin' = 'security', 'version' = '1.1.0'}`

waynew:
Yeah, @asomers makes a great point - we should have a unified return format that we expect here.

Also, we have since changed to a `master`-only branch, and we require tests for merging PRs.

asomers:
I'm working on the tests.  I should have something soon.

waynew:
This is definitely not the desired approach. It looks like ultimately the issue traces back to https://github.com/saltstack/salt/issues/25915 and possibly with freebsdpkg vs pkgng.

On FreeBSD it lists packages like so:

```
{u'py27-ipaddress': {u'origin': u'net/py-ipaddress', u'version': [u'1.0.23']}, u'py27-msgpack': {u'origin':
 u'devel/py-msgpack', u'version': [u'0.6.2']}, u'py27-cryptography': {u'origin': u'security/py-cryptography', u'version': [u'2.6.1']}, u'sudo': {u'origin': u'security/sudo', u'version': [u'1.8.30']}, u'virtualbox-ose-additions-nox11
': {u'origin': u'emulators/virtualbox-ose-additions-nox11', u'version': [u'5.2.34_1']}, u'py27-pyinotify': {u'origin': u'devel/py-pyinotify', u'version': [u'0.9.6']}, u'py27-Babel': {u'origin': u'devel/py-babel', u'version': [u'2.7.0']}, u'py27-backports_abc': {u'origin': u'devel/py-backports_abc', u'version': [u'0.5']}, u'py27-idna': {u'origin': u'dns/py-idna', u'version': [u'2.8']}, ...}
```

It does that because of [this PR](https://github.com/saltstack/salt/commit/6b0f4fca05c6aed37cef588e710a2b8f6dba4bc0#diff-e8ecdef70c9c678790ca9e0e07c9ab77R133)

On CentOS I get this kind of result:

```
{'git': ['1.8.3.1-20.el7'], 'tzdata': ['2018g-1.el7'], ...}
```

Note that one provides the origin/version combo, while the other provides just the list, since `versions_as_list=True` is passed.

We definitely need to dig a bit deeper and decide if the correct fix is simply to just remove these two lines:

https://github.com/saltstack/salt/blob/4fed7e0/salt/states/pkg.py#L1844-L1845

Or to adjust this line:

https://github.com/saltstack/salt/blob/4fed7e00aa6e63586059ca51e64d0ab1ccf45bc2/salt/modules/pkgng.py#L433

And change it to 

```
    if salt.utils.data.is_true(with_origin) and not versions_as_list:
```

I'm pretty sure that one of those two approaches are the correct one, but we should definitely understand why `with_origin` was added in the first place.

waynew:
@asomers ☝️ please check the above discussion as you make your changes. HTH!

asomers:
I find this PR to be incomplete.  It fixes the problem when specifying a fixed version of a package, but not the latest version.  Here's an updated patch:
```diff
--- salt/states/pkg.py.orig	2020-01-14 22:57:27 UTC
+++ salt/states/pkg.py
@@ -207,6 +207,10 @@ def _fulfills_version_string(installed_versions, versi
         the repository without enforcing a re-installation of the package.
         (Only applicable if only one strict version condition is specified E.G. version: 2.0.6~ubuntu3)
     '''
+        # stripping "with_origin" dict wrapper
+    if salt.utils.platform.is_freebsd():
+        if isinstance(installed_versions, dict) and 'version' in installed_versions:
+            installed_versions = installed_versions['version']
     version_conditions = _parse_version_string(version_conditions_string)
     for installed_version in installed_versions:
         fullfills_all = True
@@ -704,7 +708,7 @@ def _find_install_targets(name=None,
         for key in resolved_latest:
             if not resolved_latest[key]:
                 if key in cur_pkgs:
-                    resolved_latest[key] = cur_pkgs[key][-1]
+                    resolved_latest[key] = cur_pkgs[key]['version'][-1]
                 else:
                     resolved_latest[key] = None
         # Update the desired versions with the ones we resolved
```

Akm0d:
> I find this PR to be incomplete. It fixes the problem when specifying a fixed version of a package, but not the latest version. Here's an updated patch:
> 
> ```diff
> --- salt/states/pkg.py.orig	2020-01-14 22:57:27 UTC
> +++ salt/states/pkg.py
> @@ -207,6 +207,10 @@ def _fulfills_version_string(installed_versions, versi
>          the repository without enforcing a re-installation of the package.
>          (Only applicable if only one strict version condition is specified E.G. version: 2.0.6~ubuntu3)
>      '''
> +        # stripping "with_origin" dict wrapper
> +    if salt.utils.platform.is_freebsd():
> +        if isinstance(installed_versions, dict) and 'version' in installed_versions:
> +            installed_versions = installed_versions['version']
>      version_conditions = _parse_version_string(version_conditions_string)
>      for installed_version in installed_versions:
>          fullfills_all = True
> @@ -704,7 +708,7 @@ def _find_install_targets(name=None,
>          for key in resolved_latest:
>              if not resolved_latest[key]:
>                  if key in cur_pkgs:
> -                    resolved_latest[key] = cur_pkgs[key][-1]
> +                    resolved_latest[key] = cur_pkgs[key]['version'][-1]
>                  else:
>                      resolved_latest[key] = None
>          # Update the desired versions with the ones we resolved
> ```

Can we get these changes merged in and a basic unit test for this PR?

sagetherage (changes requested):
please rebase against the master branch

sagetherage:
@gmsoft-tuxicoman please rebase this against the `master` branch

sagetherage:
@gmsoft-tuxicoman 

Discussed in Grooming 2021-APR-19 no response to our requests to rebase against master, closing.

PR #54077, opened 2019-07-31
Fix aptpkg systemd call (bsc#1143301)

dincamihai:
### What does this PR do?
Sets a `--description` for the systemd-run command called by aptpkg to avoid long description strings that would cause crashes.
The description comes from `__name__` and it would be `salt.modules.aptpkg`

Fixes: aptpkg module causes systemd to crash in Ubuntu 18.04.
When no description is provided to systemd-run, the description is autogenerated from the apt-get command.
If we have a long list of packages the description will be longer than what systemd-run supports and it will crash (https://github.com/systemd/systemd/commit/a7419dbc59da5c8cc9e90b3d96bc947cad91ae16)

Setting the description ourselves prevents this even if the systemd doesn't have the fix.

### Previous Behavior
See above

### New Behavior
Sets the description to `salt.modules.aptpkg`

### Tests written?

Yes

### DIFF
--- a/salt/modules/aptpkg.py
+++ b/salt/modules/aptpkg.py
@@ -164,7 +164,7 @@ def _call_apt(args, scope=True, **kwargs):
     '''
     cmd = []
     if scope and salt.utils.systemd.has_scope(__context__) and __salt__['config.get']('systemd.scope', True):
-        cmd.extend(['systemd-run', '--scope', '--description "{0}"'.format(__name__)])
+        cmd.extend(['systemd-run', '--scope', '--description', '"{0}"'.format(__name__)])
     cmd.extend(args)
 
     params = {'output_loglevel': 'trace',
--- a/salt/modules/aptpkg.py
+++ b/salt/modules/aptpkg.py
@@ -164,7 +164,7 @@ def _call_apt(args, scope=True, **kwargs):
     '''
     cmd = []
     if scope and salt.utils.systemd.has_scope(__context__) and __salt__['config.get']('systemd.scope', True):
-        cmd.extend(['systemd-run', '--scope'])
+        cmd.extend(['systemd-run', '--scope', '--description "{0}"'.format(__name__)])
     cmd.extend(args)
 
     params = {'output_loglevel': 'trace',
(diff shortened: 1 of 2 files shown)


salt/modules/aptpkg.py
```diff
@@ -164,7 +164,7 @@ def _call_apt(args, scope=True, **kwargs):
     '''
     cmd = []
     if scope and salt.utils.systemd.has_scope(__context__) and __salt__['config.get']('systemd.scope', True):
-        cmd.extend(['systemd-run', '--scope'])
+        cmd.extend(['systemd-run', '--scope', '--description "{0}"'.format(__name__)])
```
meaksh: This is making `apt` calls to produce the following issue on Ubuntu 1804:

```
    - systemd-run: unrecognized option '--description "salt.loaded.int.module.aptpkg"'
```

It seems the issue is caused due the "whitespace" introduced on the list of "args". This should be fixed like this in order to prevent the issue:

```suggestion
        cmd.extend(['systemd-run', '--scope', '--description', '"{0}"'.format(__name__)])
```

tests/unit/modules/test_aptpkg.py
```diff
@@ -539,7 +539,7 @@ def test_call_apt_in_scope(self):
         with patch.dict(aptpkg.__salt__, {'cmd.run_all': MagicMock(), 'config.get': MagicMock(return_value=True)}):
             aptpkg._call_apt(['apt-get', 'purge', 'vim'])  # pylint: disable=W0106
             aptpkg.__salt__['cmd.run_all'].assert_called_once_with(
-                ['systemd-run', '--scope', 'apt-get', 'purge', 'vim'], env={},
+                ['systemd-run', '--scope', '--description "salt.modules.aptpkg"', 'apt-get', 'purge', 'vim'], env={},
```
meaksh: Same here:

```suggestion
                ['systemd-run', '--scope', '--description', '"salt.modules.aptpkg"', 'apt-get', 'purge', 'vim'], env={},
```

meaksh (changes requested):

dwoz:
@dincamihai Can you rebase this PR against the master branch please?

sagetherage:
@dincamihai Can you rebase this PR against the master branch please?

dincamihai:
Closing in favor of: https://github.com/saltstack/salt/pull/56396

PR #54078, opened 2019-07-31
Fixed salt key management with eauth

### LINKED ISSUE #50544: salt-key: key deletion not working with eauth
When using external auth is not possible to delete the minion key.

It looks that key name is not passed to `_run_cmd` function in https://github.com/saltstack/salt/blob/develop/salt/key.py#L179
```
_run_cmd: {'fun': 'key.delete', 'token': '7590720fd41faa2e98ddb3cd6947934bcf4c9fc5f74928a41ab08c375db22271', 'kwarg': {}, 'arg': [None]}
```
I was able to fix it by changing:
https://github.com/saltstack/salt/blob/develop/salt/key.py#L268
```
 ret = self._run_cmd(cmd)
```
Into:
```
                if self.opts.get('eauth') and cmd in ('accept', 'reject', 'delete'):
                    ret = self._run_cmd(cmd, args=ret['minions'])
                else:
                    ret = self._run_cmd(cmd)
```
It looks rather like workaround than a proper solution for me but it's working and I was unable to fix it in better way.

When `eauth` is not used then key name is fetched here: https://github.com/saltstack/salt/blob/develop/salt/key.py#L168

If You agree to such fix I could provide PR or provide better fix with some guidance how to do this in a proper way.

### Setup
Enable eauth in /etc/salt/master
```
external_auth:
  pam:
    users%:
      - '.*'
      - '@wheel'
      - '@runner'
      - '@jobs'
```
Add user to group `users`

### Steps to Reproduce Issue
Log on you user account and execute:
```
salt-key -T --auth=pam -d example_minion
```

### Versions Report
```
Salt Version:
           Salt: 2017.7.8

Dependency Versions:
           cffi: Not Installed
       cherrypy: Not Installed
       dateutil: Not Installed
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.7.2
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.5.6
   mysql-python: Not Installed
      pycparser: Not Installed
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: Not Installed
         Python: 2.7.5 (default, Jul 13 2018, 13:06:57)
   python-gnupg: Not Installed
         PyYAML: 3.11
          PyZMQ: 15.3.0
           RAET: Not Installed
          smmap: Not Installed
        timelib: Not Installed
        Tornado: 4.2.1
            ZMQ: 4.1.4

System Versions:
           dist: centos 7.5.1804 Core
         locale: UTF-8
        machine: x86_64
        release: 3.10.0-862.11.6.el7.x86_64
         system: Linux

```

DmitryKuzmenko:
### What does this PR do?
Properly pass minion list to wheel command in case of using eauth with `salt-key` command.

Fix #50544 

### Previous Behavior
`salt-key` reports operation successful but actually nothing happens with the key on accept/delete/reject operations with eauth.

### New Behavior
Eauth operations work as expected (like local).

### Tests written?
Yes. Also updated the eauth `salt-key` tests: use decorator to create a pam user for the test.

### Commits signed with GPG?
Yes

### DIFF
--- a/tests/support/helpers.py
+++ b/tests/support/helpers.py
@@ -38,6 +38,7 @@ import salt.ext.tornado.ioloop
 import salt.ext.tornado.web
 import salt.utils.files
 import salt.utils.platform
+import salt.utils.pycrypto
 import salt.utils.stringutils
 import salt.utils.versions
 from salt.ext import six
@@ -747,6 +748,17 @@ def with_system_user(
                                 username
                             )
                         )
+            if not salt.utils.platform.is_windows() and password is not None:
+                if salt.utils.platform.is_darwin():
+                    hashed_password = password
+                else:
+                    hashed_password = salt.utils.pycrypto.gen_hash(
+                        crypt_salt="SALTsalt", password=password
+                    )
+                hashed_password = "'{0}'".format(hashed_password)
+                add_pwd = cls.run_function(
+                    "shadow.set_password", [username, hashed_password]
+                )
 
             failure = None
             try:
--- a/salt/auth/__init__.py
+++ b/salt/auth/__init__.py
@@ -541,7 +541,7 @@ class Resolver(object):
             )
             print(
                 "Available eauth types: {0}".format(
-                    ", ".join(self.auth.file_mapping.keys())
+                    ", ".join([k[:-5] for k in self.auth if k.endswith(".auth")])
                 )
             )
             return ret
--- a/salt/key.py
+++ b/salt/key.py
@@ -184,6 +184,8 @@ class KeyCLI(object):
                 ret.pop("local", None)
             return ret
 
+        if cmd in ("accept", "reject", "delete") and args is None:
+            args = self.opts.get("match_dict", {}).get("minions")
         fstr = "key.{0}".format(cmd)
         fun = self.client.functions[fstr]
         args, kwargs = self._get_args_kwargs(fun, args)
--- a/tests/integration/shell/test_key.py
+++ b/tests/integration/shell/test_key.py
@@ -11,15 +11,33 @@ import pytest
 import salt.utils.files
 import salt.utils.platform
 import salt.utils.yaml
-from salt.ext import six
 from tests.support.case import ShellCase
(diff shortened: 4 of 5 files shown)


waynew (changes requested):

msciciel:
Any chances to backport this fix also to `2017.7` and `2018.3`?

DmitryKuzmenko:
@waynew could you please re-review?

dwoz:
@DmitryKuzmenko there are merge conflicts here now.

DmitryKuzmenko:
@waynew one more review plesae?

DmitryKuzmenko:
I've re-based squashed all to one one commit to not increase the complexity in the future. Now it should be good. @waynew please-please!

waynew (dismissed):

dwoz (approved):

tests/support/helpers.py
```diff
@@ -747,6 +748,17 @@ def wrap(cls):
                                 username
                             )
                         )
+            if not salt.utils.platform.is_windows() and password is not None:
+                if salt.utils.platform.is_darwin():
+                    hashed_password = password
+                else:
+                    hashed_password = salt.utils.pycrypto.gen_hash(
+                        crypt_salt="SALTsalt", password=password
```
dwoz, line 756: @DmitryKuzmenko We should probably not use a static salt here. It'd be better to have this as a random string in a config file which a user can update. That way they'll actually get some security from using a salted hash ;)

DmitryKuzmenko: =)))

DmitryKuzmenko: Always be as safe as you can ))

DmitryKuzmenko: Actyally we can not set the salt here, but this will force `crypt` to generate more secure salt that will make tests a little bit longer. So my way is more productive for tests. :-P

Issue #53353, opened 2019-06-03
iptables states fail when state_aggregate option is enabled


dstoliker:
### Description of Issue
State files with iptables states will fail with a runtime error `RuntimeError: maximum recursion depth exceeded in cmp` when running Salt 2019.2 with `state_aggregate` enabled in `/etc/salt/minion`.

### Setup
Enable `state_aggregate` in `/etc/salt/minion`:
```
state_aggregate: True
```
Create a state file like this example:
```
# cat /srv/salt/iptables-test.sls
append_accept_local_interface:
  iptables.append:
    - name: accept_local_interface
    - table: filter
    - chain: INPUT
    - in-interface: lo
    - jump: ACCEPT
    - save: True

append_accept_loopback_output:
  iptables.append:
    - table: filter
    - chain: OUTPUT
    - out-interface: lo
    - jump: ACCEPT
    - save: True

append_drop_non_loopback:
  iptables.append:
    - table: filter
    - chain: INPUT
    - source: 127.0.0.0/8
    - jump: DROP
    - save: True
```

### Steps to Reproduce Issue
Apply the state file:
```
# salt-call --local state.sls iptables-test -l debug
[DEBUG   ] Reading configuration from /etc/salt/minion
[DEBUG   ] Using cached minion ID from /etc/salt/minion_id: 10.0.2.15
[DEBUG   ] Configuration file path: /etc/salt/minion
[WARNING ] Insecure logging configuration detected! Sensitive data may be logged.
[DEBUG   ] Grains refresh requested. Refreshing grains.
[DEBUG   ] Reading configuration from /etc/salt/minion
[DEBUG   ] Determining pillar cache
[DEBUG   ] LazyLoaded jinja.render
[DEBUG   ] LazyLoaded yaml.render
[DEBUG   ] LazyLoaded jinja.render
[DEBUG   ] LazyLoaded yaml.render
[DEBUG   ] LazyLoaded state.sls
[DEBUG   ] LazyLoaded direct_call.execute
[DEBUG   ] LazyLoaded saltutil.is_running
[DEBUG   ] LazyLoaded grains.get
[DEBUG   ] LazyLoaded config.get
[DEBUG   ] key: test, ret: _|-
[DEBUG   ] Updating roots fileserver cache
[DEBUG   ] Gathering pillar data for state run
[DEBUG   ] Determining pillar cache
[DEBUG   ] LazyLoaded jinja.render
[DEBUG   ] LazyLoaded yaml.render
[...]

PR #54081, opened 2019-07-31
Fix for iptables states fail when state_aggregate option is enabled

xeacott:
### What does this PR do?
Add method to remove circular references to objects that must be decoded/encoded.

Fix https://github.com/saltstack/salt/issues/53353 by cleaning data object in case of circular references.

### Tests written?
Yes

### Commits signed with GPG?
No


xeacott:
Forgot to squash when I merged :-1:

waynew:
I'm not entirely sure that I like just removing circular refs. I mean, it probably solves the problem, but it obscures the fact that we shouldn't be creating circular data structures in the first place. At least, I'm not aware of any place in salt where that would be a good thing to have 🤷‍♂

xeacott:
I agree. This just puts in a fail safe that if we do have a circular reference, clean it up and let it pass through. However, I'm not entirely sure where an object would have gained a circular reference, which would be the real root cause. I can investigate it further for sure!

waynew:
I've noticed that we have occasions where this is the case. I wouldn't be opposed to putting this PR in *if* we also logged (probably as an error, probably with the stack trace?) that a circular reference was detected, asking users to report it.

At least, I don't think I would 😂

xeacott:
Let me find the correct place to log and I'll get that in with this PR!

twangboy (approved):

tests/unit/utils/test_data.py
```diff
@@ -268,6 +268,25 @@ def test_compare_lists_changes_old(self):
...
+        test_dict = {"key": "value", "type": "test1"}
+        test_dict["self"] = test_dict
+        ret = salt.utils.data._remove_circular_refs(ob=test_dict)
+        self.assertDictEqual(ret, {'key': 'value', 'type': 'test1', 'self': None})
+
+    def test_circular_refs_lists(self):
+        test_list = {'foo': [],}
+        test_list['foo'].append((test_list,))
+        ret = salt.utils.data._remove_circular_refs(ob=test_list)
+        self.assertDictEqual(ret, {'foo': [(None,)]})
```
twangboy: Looks like this is failing lint. You need a space after the comma.

twangboy (changes requested):

PR #54087, opened 2019-08-01
Pass grains in minion start event

admd:
### What does this PR do?
Adds a configuration option to have a selection of grains in the minion start event.

Reason behind this is better integration with [Uyuni]( https://github.com/uyuni-project/uyuni), especially when a lot of minions are starting in a short time frame - we need to get some minimal data (at least the `machine_id`) and getting it separately as a reaction to the minion start event (via a separate call or a startup state) generates considerable load on the master.

By default, there will be no grains passed to the event.

### Previous Behavior
```
salt/minion/abid-minion.tf.local/start	{
    "_stamp": "2019-08-01T10:46:22.840597", 
    "cmd": "_minion_event", 
    "data": "Minion abid-minion.tf.local started at Thu Aug  1 12:46:22 2019", 
    "id": "abid-minion.tf.local", 
    "pretag": null, 
    "tag": "salt/minion/abid-minion.tf.local/start"
}
```

### New Behavior
```
salt/minion/abid-minion.tf.local/start	{
    "_stamp": "2019-08-01T10:46:58.696390", 
    "cmd": "_minion_event", 
    "data": "Minion abid-minion.tf.local started at Thu Aug  1 12:46:58 2019", 
    "grains": {
        "id": "abid-minion.tf.local", 
        "machine_id": "21ee22c6906887c2dd6adeed5d404145"
    }, 
    "id": "abid-minion.tf.local", 
    "pretag": null, 
    "tag": "salt/minion/abid-minion.tf.local/start"
}
```

### Tests written?

_Tests will be provided once upstream agreed to approach, this draft PR was opened to make sure the overall approach is OK_

Yes/No

### Commits signed with GPG?

Yes

### DIFF
--- a/conf/minion
+++ b/conf/minion
@@ -548,6 +548,11 @@
 #  - edit.vim
 #  - hyper
 #
+# List of grains to pass in start event when minion starts up:
+#start_event_grains:
+#  - machine_id
+#  - id
+#
 # Top file to execute if startup_states is 'top':
 #top_file: ''
 
--- a/doc/ref/configuration/minion.rst
+++ b/doc/ref/configuration/minion.rst
@@ -2066,6 +2066,21 @@ List of states to run when the minion starts up if ``startup_states`` is set to 
       - edit.vim
       - hyper
 
+.. conf_minion:: start_event_grains
+
+``start_event_grains``
+----------------------
+
+Default: ``[]``
+
+List of grains to pass in start event when minion starts up.
+
+.. code-block:: yaml
+
+    start_event_grains:
+      - machine_id
+      - id
+
 .. conf_minion:: top_file
 
 ``top_file``
--- a/salt/config/__init__.py
+++ b/salt/config/__init__.py
@@ -1283,6 +1283,7 @@ DEFAULT_MINION_OPTS = immutabletypes.freeze({
     'state_top_saltenv': None,
     'startup_states': '',
     'sls_list': [],
+    'start_event_grains': [],
     'top_file': '',
     'thoriumenv': None,
     'thorium_top': 'top.sls',
--- a/salt/minion.py
+++ b/salt/minion.py
@@ -1475,6 +1475,11 @@ class Minion(MinionBase):
         else:
             return
 
+        if self.opts['start_event_grains']:
+            grains_to_add = dict(
+                [(k, v) for k, v in six.iteritems(self.opts.get('grains', {})) if k in self.opts['start_event_grains']])
+            load['grains'] = grains_to_add
+
         if sync:
             try:
                 self._send_req_sync(load, timeout)


admd:
@waynew @dincamihai

dincamihai:
Makes sense to me. I don't know if there are side effects to adding a new key to the start events.
Maybe there is already some test that we could update to incorporate this change? @admd

waynew:
Same here. I would also be interested to know if there's any other performance impact. Obviously it's healthier for the master since it doesn't have to process some other events, but are there any other performance implications?

admd:
Thank you @waynew & @dincamihai for the input.
@waynew can you please ping someone who might have more experience in this particular part of the code and might point us to any possible side effects.

max-arnold:
My two cents:

* I would remove the `id` grain from the examples and replace it with any other grain. All minion events already have the `id` attribute that could be trusted (unlike the custom added one).
* Why just the start event? Maybe it would be useful to be able to do this:

```
event_grains:
  - 'salt/*/start':
    - machine_id
    - saltversion
  - 'salt/cloud/*/created':
    - virtual
```

P.S. Also it would be nice to maybe unify this with `autosign_grains`, but I wasn't able to figure out how these grains are passed in the payload (`salt-run state.event` doesn't show them).

moio:
> I would remove the id grain from the examples and replace it with any other grain. All minion events already have the id attribute that could be trusted (unlike the custom added one).

+1

> I don't know if there are side effects to adding a new key to the start events.

At least none we are aware of at this point, and we are going to test this properly before asking for the merge.

> Why just the start event?

Well, this is the use case we need in Uyuni, so we didn't want to propose an overgeneral solution that's potentially more disruptive/invasive. Of course, if that is deemed a necessary condition for merging we can probably do that.


@waynew: is there any veto from you or any other team member, or any necessary condition you want to see here? Basically we would like to know if the approach is fundamentally acceptable, so that we could proceed to finalize this PR for merging (and base Uyuni code on the assumption this change in Salt will be present at some point).

Thanks!

max-arnold:
>> Why just the start event?
>  Of course, if that is deemed a necessary condition for merging we can probably do that.

@moio I'm just a community member and this was simply a suggestion. Recently I wrote a couple of states that send grains via custom events, and I thought it would be nice to automatically attach grains to any event.

KChandrashekhar:
@admd any reason this PR is still in draft ? We can have assign someone from Salt team to review this once you publish

admd:
Hi @KChandrashekhar , only reason it's still in the draft because I wanted to get Ok on the approach,  from the salt team, before proceeding. Of course, I can publish it if that's what it takes to get more attention.

waynew:
@admd discussed this with the team - the biggest concern was that adding to grains can add a *significant* performance overhead. But if we can specify the grains that should be added to the startup event, and probably default to none - the current behavior - this could be worthwhile.

So, as long as we can "first do no harm", then it should be OK to pass grains on start event (i.e. only when it's explicitly requested).

admd:
Thank you @waynew for the update. This is exactly how we approached this problem in this PR. By default, there would be no grains passed in the start event as it is today. Only after the user will explicitly add `start_event_grains` in conf file with the needed grains, we will pass those grains in start event.

waynew:
@admd in that case, I'd say it's safe to go ahead and flesh this PR out so it's ready for a review 👍

admd:
Hi @waynew, I have created a new PR https://github.com/saltstack/salt/pull/54948 against master. I will be closing this PR.

PR #54094, opened 2019-08-02
Fix Vault Request Amplification and Latency

thenoid:
### What does this PR do?
This patch does the following:
 1. Allows you set set minion or master configuration values for the TTL and Number of uses of the generated vault auth tokens.
 2. Updates the token generation function to utilize those values
 3. Updates the make_request function to appropriately cache/monitor the life cycles.
 4. Defaults to the old inefficient way of doing things

Two related issues currently exist with the design of the vault token generate/make request functionality.

Currently when a minion requests a secret the following occurs
 1. Minion generates a signature and contacts the Master for a new token, blocking while it does so
 2. Salt-master verifies the signature and sends a request to the vault server for a new, **1 use** token,
 3. Vault server generates the token, returns it to the salt-master
 4. Salt-master returns the new token to the salt-minion
 5. Salt-minion uses token and contacts the vault server, **1 use** token is now dead
 6. Vault returns secret to salt-minion.

This results in two issues:
 1. Latency - every secret lookup involves multiple network hops.  Which can quickly add up, over the course of just 10 secrets in testing, this processed averaged 42.65 seconds.
 2. Licensing/Amplification - Vault enterprise is licensed on a [per-request basis](https://medium.com/@rmolsen/monitoring-hashicorp-vault-requests-with-datadog-fa73ae53229d) With the current design you are doubling the number of requests needed to take any action.  Maybe the original author secretly had stock in hashicorp....the truth is out there 👽

### Previous Behavior
Slow salt run's and paying twice as much as you should.  See above for description of how Salt interacts with Vault

### New Behavior
716.80% speed increase of salt runs (using the 10 secret test above) and paying way less than you do now.  

Default behavior will remain the same as described above.

If uses/ttl is specified, the salt-master will generate a longer lived token.  The salt-minion will then cache that token and utilize it until the TTL or uses are exhausted.  At which point it removes the token from cache and requests a new one.

### Tests written?
No

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/runners/vault.py
+++ b/salt/runners/vault.py
@@ -14,6 +14,7 @@ import json
 import logging
 import string
 import requests
+import time
 
 # Import Salt libs
 import salt.crypt
@@ -25,7 +26,7 @@ from salt.ext import six
 log = logging.getLogger(__name__)
 
 
-def generate_token(minion_id, signature, impersonated_by_master=False):
+def generate_token(minion_id, signature, impersonated_by_master=False, ttl=None, uses=None):
     '''
     Generate a Vault token for minion minion_id
 
@@ -39,6 +40,12 @@ def generate_token(minion_id, signature, impersonated_by_master=False):
     impersonated_by_master
         If the master needs to create a token on behalf of the minion, this is
         True. This happens when the master generates minion pillars.
+
+    ttl
+        time to live in seconds, 1m minutes, or 2h hrs
+
+    uses
+        number of times a token can be use
     '''
     log.debug(
         'Token generation request for %s (impersonated by master: %s)',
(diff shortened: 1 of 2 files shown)


Ch3LL (changes requested):
Can we get some regression tests for this change?

Also looks like there is a couple of test failures related to this change.

mchugh19:
@thenoid when you say that the minion caches the token, this code uses the minion's `__context__` and, at least in my testing, it is only valid for the individual run. Meaning credentials are saved for `salt-call vault.read_secret /secret/mysecret` but after that run, `__context__` is flushed. The next call of `salt-call vault.read_secret /secret/mysecret` must generate a vault token all over again from the master. So credentails are not actually cached between `vault.read_secret` calls.

Since `__context__` does not persist, how did you test your speed increase? 

Might it make sense to write out the token to disk, similar to vault cli's .vault-token file?

thenoid:
@mchugh19 That speed up was over the course of a single `state.highstate` that was retrieving just **10** secrets from vault.  Since new tokens weren't being provisioned for *every* secret lookup, only the first, it cut network-roundtrip/provisioning down that much.

I would opt not to store the token on disk personally.  There is too much chance for error and also creates a possible security issue since the token would be on disk, unless you do some salty voodoo to create an encrypted file.



> @thenoid when you say that the minion caches the token, this code uses the minion's `__context__` and, at least in my testing, it is only valid for the individual run. Meaning credentials are saved for `salt-call vault.read_secret /secret/mysecret` but after that run, `__context__` is flushed. The next call of `salt-call vault.read_secret /secret/mysecret` must generate a vault token all over again from the master. So credentails are not actually cached between `vault.read_secret` calls.
> 
> Since `__context__` does not persist, how did you test your speed increase?
> 
> Might it make sense to write out the token to disk, similar to vault cli's .vault-token file?

Issue #39420, opened 2017-02-15
Inconsistent pillar data from git_pillar


peter-slovak:
### Description of Issue/Question
I am using git_pillar as the only pillar provider, with "dynamic" environments (`__env__` instead of hardcoded branch name). The way I understand it, this setup allows you to fork a branch in your Git repo and change values in Pillar so that only minions in the same `pillarenv` see that change.

Unfortunately, pillar representation is flapping between different branches; here's a minion in `environment: dev` and `pillarenv: dev` to whose pillar I added a new key with a value that occasionally disappears, either when getting the key from master or from the minion. In the `dev` branch this key exists with the value you see in first two attempts; in another branch `production` the key doesn't exist. 
```bash
(dev) root@node-3:~$ salt-call pillar.item ovs_bridges:br-ex-vdc                                                                                                                        
local:
    ----------
    ovs_bridges:br-ex-vdc:
        vlan430
(dev) root@node-3:~$ salt-call pillar.item ovs_bridges:br-ex-vdc
local:
    ----------
    ovs_bridges:br-ex-vdc:
        vlan430
(dev) root@node-3:~$ salt-call pillar.item ovs_bridges:br-ex-vdc
local:
    ----------
    ovs_bridges:br-ex-vdc:
``` 

I suspect this might be caused by a race between simultaneous pillar renderings on the master (since Git pillar data is saved on the master in the form of a single directory in which checkouts are performed), but have not verified it.

There is no pillar data merging configured, because I'd like to have pillar data between environments completely separated:
```bash
...
top_file_merging_strategy: same
pillar_source_merging_strategy: none
...
```

Maybe it would help if the Pillar repository on salt master got spread into different directories (one for each branch) instead of checking out inside one directory. Otherwise we have to be sure that all operations that can potentially perform `git checkout` obey a lock set by the process that is currently compiling the pillar.

### Steps to Reproduce Issue
- Install Salt 2016.11.2 on both master and minion, configure new-style Git ext_pillar with dynamic `__env__` mapping. I'm using `pygit2` right now, but the issue was identical while using `gitpython`:
```bash
git_pillar_global_lock: False
git_pillar_provider: pygit2
git_pillar_privkey: /root/.ssh/id_rsa
git_pillar_pubkey: /root/.ssh/id_rsa.pub

ext_pillar:
  - git:
    - __env__ ssh://git@git.domain.tld/group/salt.git:
      - root: pillar
```
[...]

PR #54097, opened 2019-08-02
Add a global lock when processing git_pillar

gmsoft-tuxicoman:
### What does this PR do?

This PR introduces a global lock on git pillars to make sure that the checkout and the processing operation are done atomically.

As per #39420, there is a race condition when processing git_pillar and using __env__.

The race exists in pillar/pillar.py:ext_pillar() :

```
Process 1               |  Process 2 
checkout  branch devel  |
                        | checkout branch master
read files from tree    |
                        | read files from tree
```

This occurs when the salt master is processing pillars from different branches no matter what host.

### Previous Behavior

state.apply would randomly fail. For example a file not present in the master branch but present in the devel branch would be reported as missing even if pillarenv=devel.

### New Behavior

It works !

### Tests written?

No

### Commits signed with GPG?

No

### DIFF
--- a/salt/pillar/git_pillar.py
+++ b/salt/pillar/git_pillar.py
@@ -362,7 +362,7 @@ import salt.utils.gitfs
 import salt.utils.dictupdate
 import salt.utils.stringutils
 import salt.utils.versions
-from salt.exceptions import ( FileserverConfigError, GitLockError )
+from salt.exceptions import (FileserverConfigError, GitLockError)
 from salt.pillar import Pillar
 
 # Import third party libs
@@ -481,6 +481,7 @@ def ext_pillar(minion_id, pillar, *repos):  # pylint: disable=unused-argument
     except GitLockError:
         log.error("Could not obtain the global lock for pillar procssing")
     return False
+
 
 def _extract_key_val(kv, delimiter='='):
     '''Extract key and value from key=val string.
--- a/salt/pillar/git_pillar.py
+++ b/salt/pillar/git_pillar.py
@@ -415,69 +415,72 @@ def ext_pillar(minion_id, pillar, *repos):  # pylint: disable=unused-argument
         # If masterless, fetch the remotes. We'll need to remove this once
         # we make the minion daemon able to run standalone.
         git_pillar.fetch_remotes()
-    git_pillar.global_lock()
-    git_pillar.checkout()
-    ret = {}
-    merge_strategy = __opts__.get(
-        'pillar_source_merging_strategy',
-        'smart'
-    )
-    merge_lists = __opts__.get(
-        'pillar_merge_lists',
-        False
-    )
-    for pillar_dir, env in six.iteritems(git_pillar.pillar_dirs):
-        # Map env if env == '__env__' before checking the env value
-        if env == '__env__':
-            env = opts.get('pillarenv') \
-                or opts.get('saltenv') \
-                or opts.get('git_pillar_base')
-            log.debug('__env__ maps to %s', env)
-
-        # If pillarenv is set, only grab pillars with that match pillarenv
-        if opts['pillarenv'] and env != opts['pillarenv']:
-            log.debug(
-                'env \'%s\' for pillar dir \'%s\' does not match '
-                'pillarenv \'%s\', skipping',
-                env, pillar_dir, opts['pillarenv']
+    try:
+        with git_pillar.global_lock():
+            git_pillar.checkout()
+            ret = {}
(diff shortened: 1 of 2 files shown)


waynew:
Thanks for the PR!

Looks like some lint failures - there may be some issues with this code in the two test failures, too.

Would you be willing/able to write tests for this?

gmsoft-tuxicoman:
Sure I can write some tests but I probably need some guidance. I guess we want to test only the global_lock() function.

Regarding the lint errors, I'm not sure what to do about them. The code is definitely similar to the gen_lock() function that I used as a template. Is it worth refactoring or should those errors be ignore ?

salt/pillar/git_pillar.py
```diff
@@ -415,6 +415,7 @@ def ext_pillar(minion_id, pillar, *repos):  # pylint: disable=unused-argument
         # If masterless, fetch the remotes. We'll need to remove this once
         # we make the minion daemon able to run standalone.
         git_pillar.fetch_remotes()
+    git_pillar.global_lock()
```
waynew: I did just notice that you're not using a `with` block, which you need since you're using `contextlib.contextmanager`.

salt/utils/gitfs.py
```diff
@@ -3045,6 +3045,61 @@ def checkout(self):
                 else:
                     self.pillar_dirs[cachedir] = env
 
+    @contextlib.contextmanager
+    def global_lock(self, timeout=0, poll_interval=0.5):
+        '''
+        Set and automatically clear a lock on all the repos
+        '''
+        if not isinstance(lock_type, six.string_types):
```
waynew: `lock_type` isn't actually defined here - did you mean to take it as a parameter in the function?

waynew (changes requested):
Regarding the lint, at least the one issue I looked at seems valid. Can you go ahead and review the other ones?

As far as testing goes, I assume that there's some scenario that this is supposed to protect against. If you can duplicate that scenario in an automated fashion, that would be ideal. The goals are to make sure:

1. We can replicate the problem
2. This code change fixes the problem
3. If we change code in the future, we don't create a regression (i.e. reintroduce this problem)

I'm also curious if there are any performance implications behind introducing this lock. I'm guessing not much... but if it's something severe we may need to revisit what's going on here.

gmsoft-tuxicoman:
Thanks. Indeed lock_type wasn't used at all and I had to use the with statement.
I fixed those issues.

Regarding the test, I'm not familiar with the test framework for salt.
The problem doesn't seem easy to reproduce within a test suite. To reproduce, we need to create a pillar git repo and configure the salt master accordingly. The pillar repo needs to have at least 2 branches with at least one different pillar in each branch.
Then 2 salt process need to call ext_pillar (or at least git_pillar.global_lock) at the same time and check that the resulting pillar is correct.


The problem that I see with the above is that I cannot reliably make the test fail if global_lock() wasn't there. It would just be luck if it was to fail.

Also the whole test setup feels overwhelming but it might be something that the test framework might have easy functions for.

sathieu:
Please look at #57540, fixing this without a global lock (better performance).

salt/utils/gitfs.py
```diff
@@ -3045,6 +3045,56 @@ def checkout(self):
...
+                        if not timeout or time.time() - time_start > timeout:
+                            raise GitLockError(exc.errno, exc.strerror)
+                        else:
+                            log.debug(
+                                'A global lock is already present for %s remote '
+                                '\'%s\', sleeping %f second(s)',
+                                self.role, remote.id, poll_interval
+                            )
+                            time.sleep(poll_interval)
+                            continue
```
electromatter, line 3093: The yield needs to be outside of the for loop after you have locked all of the repos. Otherwise you yield multiple times once for each repo. The purpose of this function is to yield once after all repos are locked.
```suggestion
                        break
                    except (OSError, IOError, GitLockError) as exc:
                        if not timeout or time.time() - time_start > timeout:
                            raise GitLockError(exc.errno, exc.strerror)
                        else:
                            log.debug(
                                'A global lock is already present for %s remote '
                                '\'%s\', sleeping %f second(s)',
                                self.role, remote.id, poll_interval
                            )
                            time.sleep(poll_interval)
                            continue
            yield
```

sagetherage:
@gmsoft-tuxicoman can you please rebase against the master branch, please?

sagetherage:
Reviewed in Grooming 2021-MAY-03

sagetherage:

PR #53841, opened 2019-07-12
Merge forward 2017.7 to 2018.3

dwoz:
Merge forward 2017.7 to 2018.3

PR #54107, opened 2019-08-02
Fix for file module test

DmitryKuzmenko:
### What does this PR do?
Fixes errors like
```
AssertionError: 'Found more than one line between boundaries "before" and "after"' not found in '<ExceptionInfo CommandExecutionError tblen=2>'
```
in `unit.modules.test_file` tests.

I've actually didn't got why it gets `repr` string instead of `str` string there. But what I've found is: `with raises(Exc) as exc` puts an `ExceptionInfo` value to the `exc` variable instead of the caught exception. So if we want to assert something in the exception we have to use `exc.value` instead of `exc`. It'd be more correct.

So I've made the least change let's see what will happen.

New failure. No issue.
A part of tests fixes for #53841 

### Tests written?
No, Tests fix.

### Commits signed with GPG?
Yes

### DIFF
--- a/tests/unit/modules/test_file.py
+++ b/tests/unit/modules/test_file.py
@@ -1128,9 +1128,9 @@ class FilemodLineTests(TestCase, LoaderModuleMockMixin):
         :return:
         '''
         for mode, err_msg in [(None, 'How to process the file'), ('nonsense', 'Unknown mode')]:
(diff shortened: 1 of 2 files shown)


xeacott (approved):

DmitryKuzmenko:
Merged with others to one PR #54162

Issue #54115, opened 2019-08-05
stop depending on pycrypto


JanZerebecki:
### Description of Issue
PyCrypto is unmaintained and has open security issues. This was already reported in e.g. https://github.com/saltstack/salt/issues/52674 but it is currently closed.

### Steps to Reproduce Issue
Run `pip download salt` and check if it mentions pycrypto. It currently does, but it should not. It currently comes from https://github.com/saltstack/salt/blob/develop/requirements/zeromq.txt .

### Versions Report
Checked for 2019.2.0 via pip, and on branch develop by inspecting the above file.

OrangeDog:
Note that OS packages (e.g. python3-crypto) should already have the security issues patched on currently-supported systems.

PR #54119, opened 2019-08-05
replace pycrypto with pycryptodomex [2017.7]

JanZerebecki:
pycrypto is unmaintained and has open security issues.

Fixes: #54115

### What does this PR do?
replace pycrypto with pycryptodomex, like it already is on many platforms

https://github.com/saltstack/salt/issues/54115

### Previous Behavior
require pycrpto

### New Behavior
require pycryptodomex

### Tests written?
No, drop in replacement of functionality.

### Commits signed with GPG?
Yes

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -10,7 +10,7 @@ PyYAML = "*"
 MarkupSafe = "*"
 requests = ">=1.0.0"
 tornado = ">=4.2.1,<5.0"
-pycrypto = ">=2.6.1"
+pycryptodomex = "*"
 pyzmq = ">=2.2.0"
 
 [dev-packages]
--- a/pkg/arch/PKGBUILD-local
+++ b/pkg/arch/PKGBUILD-local
@@ -13,7 +13,7 @@ depends=('python2'
          'python2-pyzmq'
          'python-m2crypto'
          'python2-yaml'
-         'pycrypto'
+         'pycryptodomex'
          'python2-psutil')
 makedepends=('git')
 provides=()
--- a/pkg/smartos/esky/zeromq_requirements.txt
+++ b/pkg/smartos/esky/zeromq_requirements.txt
@@ -1,6 +1,6 @@
 # Need to set a specific version of pyzmq, so can't use the main project's requirements file... have to copy it in and modify...
 #-r ../../../requirements/zeromq.txt
 -r ../../../requirements/base.txt
-pycrypto>=2.6.1
+pycryptodomex
 pyzmq
 -r requirements.txt
--- a/pkg/suse/salt.spec
+++ b/pkg/suse/salt.spec
@@ -74,7 +74,7 @@ BuildRequires:  python-ioflo >= 1.1.7
 BuildRequires:  python-raet >= 0.6.0
 %endif
 # requirements/zeromq.txt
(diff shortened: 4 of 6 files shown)


Akm0d (approved):

dmurphy18:
@JanZerebecki There is a plan to make Salt cryptography a plug-able entity.  Currently pycryptodomex is not available on all platforms that Salt provides for.

Currently there is a hierarchy as follows, in preferred order:
M2Crypto
pycryptodomex
pycrypto

Hence, if pycryptodomex is already installed, it will be preferred  over pycrypto.
This also answers #54115

Lastly this only addresses Suse spec file, and pip requirements, actual packaging occurs with https://github.com/saltstack/salt-pack and https://github.com/saltstack/salt-pack-py3

JanZerebecki:
The main goal of this PR was fixing this for pip requirements, the changes under pkg were just easy to add while I was at it.

pycrypto should not be installed on a platform that has one of your stated alternatives. Do you agree with that?

@dmurphy18 I couldn't read from your comment if you think something about this PR needs to be changed or if it is acceptable as is. If it needs to be changed, please point out how.

dmurphy18:
@JanZerebecki re: _pycrypto should not be installed on a platform that has one of your stated alternatives. Do you agree with that?_

I personally prefer that a customer use a better alternative than pycrypto when it is available, however cannot dictate customer circumstances, hence the alternative of providing a preference for now with Python 2.  As for the PR I cannot see it passing all tests since as previously stated, pycryptodomex is not available on all platforms that SaltStack supports.

JanZerebecki:
I'm not saying that salt should stop working with pycrypto. But it should only stop preferring it over the others when being installed using pip. Currently with pip it installs pycrypto even when pycryptodomex is already installed.

On which platforms is pycryptodomex not available via pip, where pycrypto is?

It is hard to read the CI log, but the only failure seems unrelated https://jenkinsci.saltstack.com/job/pr-kitchen-ubuntu1604-py3/job/PR-54119/1/consoleText (same one on py2/centos7 and py2/ubuntu1604):
> test_issue_2731_masterless (integration.shell.test_call.CallTest) ... FAIL

Any idea regarding the failure?

dmurphy18:
@JanZerebecki The package does not build on Debian/Ubuntu is the most probable cause.

Understand your point with pip preferences but you need to ensure works on Debian 8/9, Ubuntu 16.04/18.04, Redhat 6/7

JanZerebecki:
Earlier in the log for all the platforms CI runs on one can see that installing pycryptodomex worked just fine on each of them. So that is not the cause for the CI failure.

dmurphy18:
@JanZerebecki Need to review if M2Crypto is the preferred crypto library, have to discuss this within the core team, next week, and shall get back to you

dmurphy18:
@JanZerebecki pycryptodomex shall be the pip install preferred library, will take a deeper look into this PR and see why tests failing.  Can you re-submit these changes against the 2018.3 branch since that is the branch which is currently supported.  The 2017.7 branch is only for CVE fixes at this point.

Thanks

JanZerebecki:
The main commit cherry picked on 2018.3: https://github.com/saltstack/salt/pull/54241

Issue #53912, opened 2019-07-18
Flaky cloud tests 


KChandrashekhar:
### Description of Issue
Cloud Tests are not passing consistently 

### Setup
NA

### Steps to Reproduce Issue
NA

### Versions Report
NA

KChandrashekhar:
Might be a cloud flaky test failure 
https://jenkinsci.saltstack.com/job/2019.2.1/view/All/job/salt-centos-7-py2-cloud/100/

Akm0d:
I'll check it out

Akm0d:
For reference: https://jenkinsci.saltstack.com/job/2019.2.1/view/All/job/salt-centos-7-py2-cloud/100/testReport/junit/integration.cloud.clouds.test_digitalocean/DigitalOceanTest/test_list_sizes/

PR #54121, opened 2019-08-05
Refactored cloud tests

Akm0d:
### What does this PR do?
Clearly check for instance creation and deletion at the beginning and end of cloud tests

### Related to issue
https://github.com/saltstack/salt/issues/53912

### Commits signed with GPG?

Yes


Akm0d:
re-run full all

tests/integration/cloud/clouds/test_digitalocean.py
```diff
@@ -122,53 +77,33 @@ def test_key_management(self):
...
-            raise
+        ret_str = self.run_cloud('-p digitalocean-test {0}'.format(self.instance_name), timeout=TIMEOUT)
+        self.assertInstanceExists(ret_str)
 
-        # Final clean-up of created instance, in case something went wrong.
-        # This was originally in a tearDown function, but that didn't make sense
-        # To run this for each test when not all tests create instances.
-        if INSTANCE_NAME in [i.strip() for i in self.run_cloud('--query')]:
-            self.run_cloud('-d {0} --assume-yes'.format(INSTANCE_NAME), timeout=500)
+        self._destroy_instance()
```
s0undt3ch: Probably, `self.addCleanup(self._destroy_instance)` before calling the assert method. This way, even if the assertion method fails, the cleanup code will run.

tests/integration/cloud/clouds/test_ec2.py
```diff
@@ -9,115 +9,65 @@
...
 
-    def _ensure_installer(self):
-        '''
-        Make sure the testing environment has a Windows installer executbale.
-        '''
-        name = self._installer_name()
-        if name:
-            return name
-        return self._fetch_latest_installer()
+INSTALLER = _fetch_installer()
```
s0undt3ch: Does this need to be at the module global level?
Than means, for example, when pytest collects tests(I know 2019.2.1 does not run pytest, but the "problem" remains), this code will run, even if the test is to be skipped.
Stick it in `setUp` or `setUpClass`?

s0undt3ch (changes requested):

tests/integration/cloud/clouds/test_ec2.py
Akm0d: That makes sense :+1:

tests/integration/cloud/clouds/test_digitalocean.py
Akm0d: the tearDown inherited from the CloudTest metaclass will make sure that `self._destroy_instance()` is run.  It will also fail the test if the instance is destroyed from the tearDown because it should be successful the first time `self._destroy_instance()` is called

twangboy (approved):

tests/integration/cloud/helpers/cloud_test_base.py
```diff
@@ -0,0 +1,175 @@
...
+
+    def tearDown(self):
+        '''
+        Clean up after tests, If the instance still exists for any reason, delete it.
+        Instances should be destroyed before the tearDown, _destroy_instance() should be called exactly
+        one time in a test for each instance created.  This is a failSafe and something went wrong
+        if the tearDown is where an instance is destroyed.
+        '''
+        instance_deleted = True
+        tries = 0
```
dwoz: - Why define tries here? It'll always start a 0 when calling range.
- We might consider making the number of tries and the sleep time variables in case we want to tweak them in the future.

tests/integration/cloud/clouds/test_digitalocean.py
dwoz: Why bother calling `_destroy_instance` in the test when it's alway going to be done by the `tearDown` method?

tests/integration/cloud/clouds/test_joyent.py
```diff
@@ -5,95 +5,29 @@
...
 # Create the cloud instance name to be used throughout the tests
-INSTANCE_NAME = generate_random_name('CLOUD-TEST-')
+from tests.integration.cloud.helpers.cloud_test_base import CloudTest, TIMEOUT
+
 PROVIDER_NAME = 'joyent'
 
 
-class JoyentTest(ShellCase):
+@skipIf(True, 'Joyent is EOL as of November 9th, 2019.  It will no longer be supported in salt-cloud')
+class JoyentTest(CloudTest):
```
dwoz: I thought we agreed to remove these `joyent` tests.

dwoz (changes requested):

tests/integration/cloud/helpers/cloud_test_base.py
Akm0d: That was so that I could include it in a test failure message outside of the for loop

tests/integration/cloud/clouds/test_joyent.py
Akm0d: :+1: I'll delete the file

tests/integration/cloud/clouds/test_digitalocean.py
Akm0d: The _destroy_instance in the tearDown is just a contingency so that we don't accidentally end up spinning up instances that don't get cleaned up.  The test fails if _destroy_instance happens in the tearDown.  After a test that creates an instance is over I want to explicitly destroy the instance and assert that it was destroyed on the first try.

dwoz (approved):

tests/integration/cloud/clouds/test_gce.py
```diff
@@ -6,140 +6,46 @@
...
-from tests.support.paths import FILES
-from tests.support.helpers import expensiveTest, generate_random_name, flaky
-
-TIMEOUT = 500
+from tests.integration.cloud.helpers.cloud_test_base import TIMEOUT, CloudTest
+from tests.support.unit import skipIf
 
 
-class GCETest(ShellCase):
+@skipIf(True, 'WAR ROOM TEMPORARY SKIP')  # needs to be rewritten to allow for dnf on Fedora 30 and RHEL 8
```
Ch3LL: why are we skipping this test? Wasn't this passing previously?

tests/integration/cloud/clouds/test_oneandone.py
```diff
@@ -5,78 +5,33 @@
...
 from tests.support.unit import skipIf
-from tests.support.helpers import expensiveTest, generate_random_name
-
-# Import Salt Libs
-from salt.config import cloud_providers_config
 
 # Import Third-Party Libs
 try:
     from oneandone.client import OneAndOneService  # pylint: disable=unused-import
+
```
Ch3LL: do we need a space here?

tests/integration/cloud/clouds/test_openstack.py
```diff
@@ -6,40 +6,35 @@
...
-from salt.config import cloud_providers_config
+from tests.integration.cloud.helpers.cloud_test_base import TIMEOUT, CloudTest
 
 log = logging.getLogger(__name__)
 
 try:
     import keystoneclient  # pylint: disable=import-error,unused-import
     from libcloud.common.openstack_identity import OpenStackIdentity_3_0_Connection
     from libcloud.common.openstack_identity import OpenStackIdentityTokenScope
+
```
Ch3LL: space here too?

tests/integration/files/conf/cloud.providers.d/profitbricks.conf
```diff
@@ -1,6 +1,6 @@
 profitbricks-config:
-  username: 'foo'
-  password: 'bar'
+  username: ''
```
Ch3LL: This was added because of the details in here: https://github.com/saltstack/salt/pull/46265  If we aren't seeing this issue anymore we can remove this. If you are not seeing that issue with this change I am guessing that we arent installing the profitbrick library or its been updated and fixed?

Ch3LL (changes requested):

tests/integration/cloud/clouds/test_gce.py
Akm0d: I was having trouble with GCE when two builds were running simultaneously, I wanted to skip until it was figured out so I could get the rest of this merged

tests/integration/files/conf/cloud.providers.d/profitbricks.conf
Akm0d: I'm glad you pointed this out because it was confusing to me.  I'll verify that the issue you referenced won't come back

Ch3LL (approved):

PR #54146, opened 2019-08-08
cherrypy: fix handling of raw application/x-www-form-urlencoded traffic

nicholasmhughes:
### What does this PR do?
handle the preservation of raw application/x-www-form-urlencoded traffic in cherrypy

N/A

### Previous Behavior
cherrypy natively parsed the incoming x-www-form-urlencoded traffic into a dictionary, which was then put into the `post` key when the event fired. the `raw_body` was manually set to an empty string, because cherrypy will leave `raw_body` as `None` when parsing x-www-form-urlencoded traffic. therefore, the `body` key of the event was always an empty string in the case of x-www-form-urlencoded traffic, where that key would normally contain the raw request when processing other types of traffic.

I consider this "broken" because it doesn't provide a way to handle the raw data _at all_ when parsing x-www-form-urlencoded traffic. This functionality is required when attempting to validate [Slack webhooks](https://api.slack.com/docs/verifying-requests-from-slack#a_recipe_for_security) since the raw x-www-form-urlencoded string is hashed for a signature and cannot be deserialized because it will destroy the order of the parameters and provide no mechanism to validate the traffic.

### New Behavior
instead of using the native cherrypy parsing, which throws away the raw request when creating the `params` dictionary, we read the raw data into a string (which ends up in the `body` key in the event) and then perform manual parsing of the tokens into a dictionary using parse_qsl to set the contents of the `post` key in the event. since the `post` key output functionality isn't changed and the `body` key was always empty and therefore wasn't relied on, this is a low impact change.

### Tests written?
No - existing tests cover functionality

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/netapi/rest_cherrypy/app.py
+++ b/salt/netapi/rest_cherrypy/app.py
@@ -636,6 +636,15 @@ except ImportError:
 
     HAS_WEBSOCKETS = False
 
+# Python version-specific imports
+try:
+    if six.PY2:
+        from urlparse import parse_qsl
+    if six.PY3:
+        from urllib.parse import parse_qsl
+except ImportError as exc:
+    logger.warn('Import failed: %s', exc)
+
 
 def html_override_tool():
     '''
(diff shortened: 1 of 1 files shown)


nicholasmhughes:
think this might've flown in under the radar... @waynew , can you help find an reviewer for this?

waynew:
@nicholasmhughes sure thing - would you be able to create a PR for this and rebase/edit your PR to point against master? 🙏

nicholasmhughes:
my world has been turned upside down... I'm moving this to #54901

PR #54147, opened 2019-08-08
remove stuff that was used on old jenkins

bryceml:
### What does this PR do?
This removes a bunch of stuff that made it so we could run on the old jenkins and the new jenkins at the same time.

None

### Previous Behavior
We had conditionals to run on both jenkins

### New Behavior
We only run on new jenkins, we can delete some unneeded credentials on the new jenkins now.

This probably needs approved by @s0undt3ch

### DIFF
--- a/.ci/docs
+++ b/.ci/docs
@@ -15,25 +15,11 @@ if (buildNumber > 1) {
 // Define a milestone for this build so that, if another build starts, this one will be aborted
 milestone(buildNumber)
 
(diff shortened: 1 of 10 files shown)


cmcmarrow (approved):

bryceml:
conflicts with https://github.com/saltstack/salt/pull/53921

s0undt3ch (approved):

PR #54148, opened 2019-08-08
Fix plugin in user_chpass with unix_socket True

scambra:
### What does this PR do?

Plugin should be set to unix_socket, as user_create does.
user_exists and user_create checks/sets plugin as unix_socket

### Previous Behavior

user_chpass is setting 'auth_socket', which fails to login with error:

ERROR 1524 (HY000): Plugin 'auth_socket' is not loaded

### New Behavior

user_chpass sets 'unix_socket' plugin, then login works.

### Tests written?

No


garethgreenaway (approved):

scambra:
Why is it closed? Master still have wrong line:
https://github.com/saltstack/salt/blob/master/salt/modules/mysql.py#L1562

PR #54159, opened 2019-08-09
Test mode should return changes

angeloudy:
If extraction is needed, the state should return some changes, otherwise, requisites such as `prereq`, `onchanges` won't work.

It's hard to determine the exact changes, but the archive state should at least return some thing non-empty as long as extraction is necessary.


angeloudy:
It is actually more complicated to return changes in test mode for this state. Close it for now.

PR #54164, opened 2019-08-09
Add hmac_compute

Ajnbro:
### What does this PR do?
Creates a util and a module called hmac_compute for computing a HMAC SHA256 hexdigest using a string and secret.

N/A

### Tests written?
Yes

### Commits signed with GPG?
Yes


waynew (approved):

waynew:
This looks good - and the lint issues aren't actually with your code, just existing code. 

Would you mind fixing those?

nicholasmhughes:
@waynew , I think the right way to go here is to mark `base64_encodestring` and `base64_decodestring` for deprecation. Any attempts to fix the lint issues will end up duplicating the functionality already present in `base64_b64encode` and `base64_b64decode`. What do you think?

waynew:
I don't understand what you mean.

base64_b64* are wrappers around base64.b64(encode/decode)

base64.encodebytes does *not* produce the same output:

```
>>> import base64
>>> string = b'salt'*100
>>> print(base64.b64encode(string).decode().count('\n'))
0
>>> print(base64.encodebytes(string).decode().count('\n'))
8
```

What am I missing here?

nicholasmhughes:
haha! my bad. I wasn't using a long enough test. `*100` ftw.

Ajnbro:
@waynew I apologize for the delay, I haven't been able to get to this till now. I will be fixing it later tonight most likely. Correct me if I am wrong here, but Python2 only has support for encodestring/decodestring and not for encodebytes/decodebyes. While Python3 has support for both (with encodestring/decodestring being depreciated). Is Python2 support anticipated to be dropped by the next version of Salt? If it isn't then should I just use both encodebytes/decodebytes and encodestring/decodestring in the module and utils files and ensure that the correct function is being used based on the Python version Salt supports at that moment? If it is going to be dropped by the next version then I can just set it up with only encodebytes/decodebytes.

waynew:
> @waynew I apologize for the delay, I haven't been able to get to this until now.

No problem. Life gets busy for all of us :)

> Python2 only has support for encodestring/decodestring and not for encodebytes/decodebyes. While Python3 has support for both (with encodestring/decodestring being depreciated). Is Python2 support anticipated to be dropped by the next version of Salt? If it isn't then should I just use both encodebytes/decodebytes

Right - with the Sodium release we are dropping Python2 support. That means that you can safely use `encodebytes/decodebytes` instead. 

> Edit: Also I accidentally requested that review, sorry about that.

Heh, no problem - Sometimes it's not entirely clear what does what in GitHub 😂 

---

Summary: Yes, use `encodebytes/decodebytes` instead of `...string`.

Ajnbro:
@waynew Finished fixing that all up. Sorry about the multiple commit mess-ups, this is my first time ever submitting a PR :P

salt/modules/hashutil.py
```diff
@@ -113,23 +113,23 @@ def base64_b64decode(instr):
     return salt.utils.hashutils.base64_b64decode(instr)
 
 
-def base64_encodestring(instr):
+def base64_encodebytes(instr):
```
waynew: We don't want to just drop the existing name - because whatever code that's currently using it will fail without warning.

What I would do is actually just 1) create encodebytes like you have, but 2) leave the encodestring and it should basically be:

    return salt.utils.stringutils.to_str(salt.utils.hashutils.base64_encodebytes(instr))

(It might be that we need to also `salt.utils.stringutils.to_bytes(instr)` there - I'd have to look a bit closer)

waynew (changes requested):
No problems 😀 We've all been there, I'm sure 🙃 

Only one little tweak I think, and this should be good to go!

salt/modules/hashutil.py
Ajnbro: Completed

waynew (approved):

waynew:
@Ajnbro looks great, thanks! We're working through some internal changes to exactly how we merge, so the *actual* merge will be coming soon(ish). Just translate that `merge ready` label to `merged` mentally if you need to 😃

dwoz:
@Ajnbro We are no longer accepting PRs to the `develop` branch. This needs to be re-opened against the `master` branch and reference this PR so we have the conversation history. Sorry for the confusion.

Issue #54174, opened 2019-08-09
SignalHandlingMultiprocessingProcess bugs


dwoz:
Misuse of `multiprocessing.utils.register_after_fork` causes multiple bugs in `salt.utils.process`. Every fork will call the after fork methods registered, not just the fork associated with the process it is intended for. The following code illustrates the issue. An un-related process can cause a `SignalHandlingMultiprocessingProcess` to get killed off prematurely.

```
import os
import signal
import logging
from salt.utils.process import SignalHandlingMultiprocessingProcess, log
import time
import multiprocessing

log.addHandler(logging.StreamHandler())
log.setLevel(logging.DEBUG)

evt = multiprocessing.Event()

def target():
    def run_forever():
        while not evt.is_set():
            time.sleep(1)
    p = multiprocessing.Process(target=run_forever)
    p.start()
    p.join()

sh_proc = SignalHandlingMultiprocessingProcess(target=target)
sh_proc.start()

def kill_target_sub_proc():
    pid = os.fork()
    if pid == 0:
        return
    pid = os.fork()
    if pid == 0:
        return
    time.sleep(.1)
    try:
        os.kill(os.getpid(), signal.SIGINT)
    except KeyboardInterrupt:
        pass

proc = multiprocessing.Process(target=kill_target_sub_proc)
proc.start()
proc.join()

# When the bug exists, the kill_target_sub_proc signal will kill both
# processes. sh_proc will be alive if the bug is fixed
assert sh_proc.is_alive()

evt.set()
sh_proc.join()
```

### Versions Report

All current versions are affected.

dmurphy18:
@dwoz Giving this to you from Triage since you have fix for it

sagetherage:
looks like this is fixed, can we close the ticket @dwoz ?

sagetherage:
closing as fixed

PR #54176, opened 2019-08-10
Fix up transport tests and bugs in process utils

dwoz:
### What does this PR do?

- Run loops in a single thread during transport unit tests
- Fix bugs in `salt/utils/process.py` caused by the use of `multiprocessing.utils.register_after_fork`

#54174

### Tests written?

Yes

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/utils/process.py
+++ b/salt/utils/process.py
@@ -753,24 +753,20 @@ class MultiprocessingProcess(multiprocessing.Process, NewStyleClassMixIn):
 class SignalHandlingMultiprocessingProcess(MultiprocessingProcess):
     def __init__(self, *args, **kwargs):
         super(SignalHandlingMultiprocessingProcess, self).__init__(*args, **kwargs)
-        self._signal_handled = multiprocessing.Value('i', 0)
-        self._signal_handled.value = 0
+        self._signal_handled = multiprocessing.Event()
         self._after_fork_methods.append(
             (SignalHandlingMultiprocessingProcess._setup_signals, [self], {})
         )
 
     def signal_handled(self):
-        if self._signal_handled.value == 1:
-            return True
-        return False
+        return self._signal_handled.is_set()
 
     def _setup_signals(self):
         signal.signal(signal.SIGINT, self._handle_signals)
         signal.signal(signal.SIGTERM, self._handle_signals)
 
     def _handle_signals(self, signum, sigframe):
-        if self._signal_handled.value == 0:
-            self._signal_handled.value = 1
+        self._signal_handled.set()
         msg = '{0} received a '.format(self.__class__.__name__)
         if signum == signal.SIGINT:
             msg += 'SIGINT'
--- a/salt/utils/process.py
+++ b/salt/utils/process.py
@@ -669,22 +669,9 @@ class MultiprocessingProcess(multiprocessing.Process, NewStyleClassMixIn):
         return instance
 
     def __init__(self, *args, **kwargs):
-        if (salt.utils.platform.is_windows() and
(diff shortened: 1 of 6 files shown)


tests/unit/utils/test_process.py
```diff
@@ -236,35 +236,158 @@ def test_daemonize_if(self):
...
+        start = time.time()
+        while time.time() - start < 10:
+            if proc.signal_handled():
+                break
+            time.sleep(.3)
+
+        assert proc.signal_handled()
+        # Reap the signaled process
+        proc.join(1)
+        assert psutil.pid_exists(proc2.pid)
```
s0undt3ch: If the assert fails the event set and proc join won't run...

s0undt3ch: Safe assertion?

salt/utils/process.py
```diff
@@ -755,12 +721,17 @@ def __getstate__(self):
...
+                '_finalize_methods': self._finalize_methods,
+                }
 
-    def __setup_process_logging(self):
+    def _setup_process_logging(self):
         salt.log.setup.setup_multiprocessing_logging(self.log_queue)
 
     def _run(self):
+        for method, args, kwargs in self._after_fork_methods:
+            method(*args, **kwargs)
```
s0undt3ch: No need to wrap in a try/except to make sure all run?

salt/utils/process.py
```diff
@@ -774,31 +745,32 @@ def _run(self):
             # sys.stderr and set the proper exitcode and we have already logged
             # it above.
             raise
+        finally:
+            for method, args, kwargs in self._finalize_methods:
+                method(*args, **kwargs)
```
s0undt3ch: Same try/except Q as above.

salt/utils/process.py
```diff
@@ -774,31 +745,32 @@ def _run(self):
...
-                # register_after_fork() would only work on Windows if called
-                # from the child process anyway. Since we know this is the
-                # child process, call __setup_signals() directly.
-                self.__setup_signals()
-        else:
-            multiprocessing.util.register_after_fork(
-                self,
-                SignalHandlingMultiprocessingProcess.__setup_signals
-            )
+        self._signal_handled = multiprocessing.Value('i', 0)
```
s0undt3ch: Just curious, why a `Value` instead of for example `Event`?

s0undt3ch (approved):

salt/utils/process.py
dwoz: No particular reason. I can change this to event if you prefer.

s0undt3ch: Nope, no need to change. Just though there could be a reason behind it, like performance or similar and I just wanted to know so I could also do better next time I was faced with a similar problem.

salt/utils/process.py
dwoz: Since this api is internal to the class only I think we're safer to fail fast and let the exception get raised if there is one. We'll be more likely see and fix any issues that come up.

salt/utils/process.py
dwoz: Same for this one.

tests/unit/utils/test_process.py
dwoz: Yeah, something is weird since we'd never reap proc2. I will fix.

salt/utils/process.py
s0undt3ch: Ok

dwoz:
Closing this in favor of #54203 and #54204

PR #54205, opened 2019-08-13
Remove noisy debug logging from config.get

terminalmage:
This is not helpful and results in a bunch of noise in debug logging, since config.get is used in a lot of places.

### DIFF
--- a/salt/modules/config.py
+++ b/salt/modules/config.py
@@ -480,7 +480,6 @@ def get(
         ret = salt.utils.data.traverse_dict_and_list(
             DEFAULTS, key, "_|-", delimiter=delimiter
         )
-        log.debug("key: %s, ret: %s", key, ret)
         if ret != "_|-":
             return sdb.sdb_get(ret, __opts__)
     else:


dwoz (dismissed):

Ch3LL (dismissed):

waynew:
Don't we have garbage or maybe trace level we could use? Maybe just removing it is still better, I just have a nagging feeling there are some esoteric reasons one may want to see that info. Also that we may run into those cases sooner than expected.

But maybe I'm just paranoid 🙃🤷‍♂️

terminalmage:
I'd argue that the message is actively unhelpful, as the "ret" message there suggests that what is being returned by `config.get` is `_|-` in cases where that config key is not defined. In reality, it's showing you the return data of a function call made _within_ `config.get`. This is not very useful information.

To be honest, this has the look of something someone added while working on changes to the function, and forgot to remove.

waynew:
Fair enough 👍

waynew (dismissed):

s0undt3ch (dismissed):

terminalmage:
Since this was never merged into 2019.2, I rebased against master.

terminalmage:
Been a few months, so I've rebased again.

terminalmage:
This PR was opened almost 8 months ago. It is a one-line-change that just removes a noisy log entry. It was approved twice within a day of being opened.

Can we please get this merged?

DmitryKuzmenko (approved):

Issue #54206, opened 2019-08-14
[2019.2.0] git.present not honoring unless


tyhunt99:
### Description of Issue
The git.present state does not honor the unless or only if options.

### Setup
The following will always run regardless of the result of `ls /etc/salt/.git`
State:
```
init_repo:
  git.present:
    - name: /etc/salt
    - unless: ls /etc/salt/.git
```

### Versions Report
```
Salt Version:
           Salt: 2019.2.0

Dependency Versions:
           cffi: 1.12.3
       cherrypy: Not Installed
       dateutil: 2.6.1
      docker-py: 1.10.6
          gitdb: 2.0.3
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.10
        libgit2: 0.28.2
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: 1.0.7
   msgpack-pure: Not Installed
 msgpack-python: 0.5.6
   mysql-python: Not Installed
      pycparser: 2.19
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: 0.28.2
         Python: 2.7.15+ (default, Nov 27 2018, 23:36:35)
   python-gnupg: 0.4.1
         PyYAML: 3.12
          PyZMQ: 16.0.2
           RAET: Not Installed
          smmap: 2.0.3
        timelib: Not Installed
        Tornado: 4.5.3
            ZMQ: 4.2.5

System Versions:
           dist: Ubuntu 18.04 bionic
         locale: UTF-8
        machine: x86_64
        release: 4.15.0-1044-aws
         system: Linux
        version: Ubuntu 18.04 bionic
```

dmurphy18:
@tyhunt99 Thanks for the issue and the PR to fix it, comments left in the PR about adding tests.

PR #54207, opened 2019-08-14
Git present mod run check

tyhunt99:
### What does this PR do?
Adds ability to specify unless and onlyif for git.present

#54206 

### Previous Behavior

### New Behavior

### Tests written?

No

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/states/git.py
+++ b/salt/states/git.py
@@ -2130,6 +2130,14 @@ def present(name,
 
       .. versionadded:: 2016.3.4
 
+    onlyif
+        A command to run as a check, run the named command only if the command
+        passed to the ``onlyif`` option returns true
+
+    unless
+        A command to run as a check, only run the named command if the command
+        passed to the ``unless`` option returns false
+
     output_encoding
         Use this option to specify which encoding to use to decode the output
         from any git commands which are run. This should not be needed in most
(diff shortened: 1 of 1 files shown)


dmurphy18 (changes requested):
@tyhunt99 Can you also supply some tests to exercise your PR's changes.  SaltStack is now asking for tests with PRs to ensure higher stability of Salt

waynew:
@tyhunt99 thanks for the PR! As David mentioned, we are asking for tests to go with PRs - you don't have to test the rest of this function, just the changes that you've made. If you're unsure where to start, please let us know and we'll do what we can to help. Thanks again!

mchugh19:
If #55974 is accepted, this can be closed out. Various state modules had their own requisites system. As the global requisites have gained functionality, these individual ones have not kept up (as this PR attempts to partially correct). #55974 removes the local state requisites in favor of the super set globals and thus this PR would no longer be needed.

sagetherage:
> If #55974 is accepted, this can be closed out. Various state modules had their own requisites system. As the global requisites have gained functionality, these individual ones have not kept up (as this PR attempts to partially correct). #55974 removes the local state requisites in favor of the super set globals and thus this PR would no longer be needed.

closing

PR #54210, opened 2019-08-14
Fix zpool state's return values for test mode

vector-of-bool:
## What does this PR do?

Fix test mode for `zpool.present` and `zpool.absent`.

(None)

### Previous Behavior

The `zpool` state would return changes when executed in test mode when no changes are pending and return ``result=True` in test mode when changes were pending.

### New Behavior

- `result` should be `None` for pending changes
- `changes` should be empty when run in test mode.
- Tweak test `comment` for present/absent

### Tests written?

No - I tried to get tests setup but something appears awry on `develop` when I try to run them. Rather than delay this PR, I'll open it and try to get tests in-place if they are required for this change.

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/states/zpool.py
+++ b/salt/states/zpool.py
@@ -295,11 +295,9 @@ def present(name, properties=None, filesystem_properties=None, layout=None, conf
     # don't do anything because this is a test
     if __opts__['test']:
         ret['result'] = True
-        if __salt__['zpool.exists'](name):
-            ret['changes'][name] = 'uptodate'
-        else:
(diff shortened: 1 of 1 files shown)


vector-of-bool:
I got tests running and was able to do some heavier refactoring to get `zpool` to behave much more nicely in test mode.

vector-of-bool:
I was able to get tests running, and decided to do a big cleanup and refactor of the `zpool` state. I'll open a new PR.

PR #54221, opened 2019-08-15
Napalm: Fixing loading template from source

nodebanger:
### What does this PR do?
This PR picks up where https://github.com/saltstack/salt/pull/52328 left off and fixes a bug where a user is unable to pass an inline template to net.load_template.

There is a bug in the the load_template call of the napalm_network module where we attempt to eval a NoneType as a string when template_name == None. This prevents a user from passing an inline template to net.load_template:

```
      File "/usr/lib/python2.7/dist-packages/salt/minion.py", line 1607, in _thread_return
        return func(*args, **kwargs)
      File "/usr/lib/python2.7/dist-packages/salt/utils/napalm.py", line 430, in func_wrapper
        return func(*args, **kwargs)
      File "/var/cache/salt/proxy/extmods/junos/modules/napalm_network.py", line 1920, in load_template
        salt_render = salt_render or template_name.startswith(salt_render_prefix)
```

### Tests written?

Yes

### Commits signed with GPG?

Yes


nodebanger:
@mirceaulinic Does this PR look good to you?

@twangboy Is there anything outstanding left to be done in order to get this PR merged?

cmcmarrow (approved):

cmcmarrow:
Looks good to me

cmcmarrow:
@nodebanger

tests/support/napalm.py
```diff
@@ -391,7 +391,7 @@ def get_environment(self):
     def get_arp_table(self):
         return TEST_ARP_TABLE
 
-    def get(self, key, default=None, *args, **kwargs):
+    def get(self, key, *args, default=None, **kwargs):
```
twangboy: I don't think this is gonna work. `*args` and `**kwargs` go at the end.

twangboy (changes requested):

tests/support/napalm.py
waynew: It will work on modern Python. Considering this is going into develop, I believe it should be fine, though I'd have to look it up to be sure when it was added to Python.

salt/modules/napalm_network.py
```diff
@@ -1914,7 +1914,7 @@ def load_template(template_name=None,
     salt_render_prefixes = ('salt://', 'http://', 'https://', 'ftp://')
     salt_render = False
     file_exists = False
-    if not isinstance(template_name, (tuple, list)):
+    if template_name and not isinstance(template_name, (tuple, list)):
```
waynew: To make sure I understand, what we're looking for here is that we have something *in* the tuple or list? Because None will already fail the check, but empty lists will pass.

Wouldn't it make more sense to make this positively check that's text? I'm pretty sure we could use something like `six.text_type`, I forget what the name is exactly.

tests/support/napalm.py
cmcmarrow: It works

salt/modules/napalm_network.py
nodebanger: None type will not fail the type check and cause the if block to be eval'd, which is the problem since at the function signature `template_name` defaults to None:
```
>>> template_name = None
>>> if not isinstance(template_name, (tuple, list)):
...   print(True)
...
True
```
I'm happy to explicitly check for text if you would like. The way this diff is written, anything other than a None, tuple or list type will cause the if block to be eval'd.

mirceaulinic (approved):
LGTM. Thanks @nodebanger!

tests/support/napalm.py
cmcmarrow: Yea @nodebanger you'r right it does not work in python2 :( but you can add a commit that will disable the pylint warning

dwoz:
@nodebanger We are no longer accepting PRs to the `develop` branch. This needs to be re-opened against the `master` branch and reference this PR so we have the discussion history. Sorry for the confusion.

PR #54222, opened 2019-08-15
Refactor the zpool state to behave correctly in test mode

vector-of-bool:
### What does this PR do?

- Heavily simplify zpool.present
- Property updates will trigger changes in test mode
- `None` is only returned for pending changes
- `changes` is empty when run in test mode
- New tests for test mode
- Tweak many state result comments

(None)

### Previous Behavior

zpool `changes` and `result` returned incorrect values when `test=true`, and pending changes to pool properties were missed in test mode.

### New Behavior

`test=true` now behaves correctly in more cases.

### Tests written?

Yes

### Commits signed with GPG?

Yes

### Other:

When `test=true`, there seems to be a disagreement in existing states on whether `changes` should be populated or not (with pending changes). Some states also use `pchanges`, but this does not appear to actually have any effect in any of the rest of the codebase. I opted to leave `changes` empty in all cases when `test=true`.

### DIFF
--- a/salt/states/zpool.py
+++ b/salt/states/zpool.py
@@ -72,11 +72,14 @@ States for managing zpools
 from __future__ import absolute_import, print_function, unicode_literals
 
 # Import Python libs
+from collections import namedtuple
+import itertools
 import os
 import logging
 
 # Import Salt libs
 from salt.utils.odict import OrderedDict
+from salt.ext.six import iteritems
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


garethgreenaway:
@vector-of-bool Would you be able to rebase this PR against the master branch.  Thanks!

sagetherage:
@vector-of-bool Would you be able to rebase this PR against the master branch. Thanks!

sagetherage:
Reviewed in Grooming 20201-MAY-03 considering closing

PR #54228, opened 2019-08-16
Docfix adding a note about masterless saltutil.sync_all calls

ixs:
### What does this PR do?
Small documentation change clarifying functionality of saltutil.sync_all in masterless mode.

### DIFF
--- a/doc/ref/modules/all/salt.modules.pkg.rst
+++ b/doc/ref/modules/all/salt.modules.pkg.rst
@@ -15,26 +15,26 @@ Execution Module                       Used for
 :py:mod:`~salt.modules.aixpkg`         AIX OS using ``installp`` and ``rpm``
 :py:mod:`~salt.modules.aptpkg`         Debian/Ubuntu-based distros which use
                                        ``apt-get(8)`` for package management
-:py:mod:`~salt.modules.brew`           Mac OS software management using
+:py:mod:`~salt.modules.mac_brew_pkg`   Mac OS software management using
                                        `Homebrew`_
-:py:mod:`~salt.modules.ebuild`         Gentoo-based systems (utilizes the
+:py:mod:`~salt.modules.ebuildpkg`      Gentoo-based systems (utilizes the
                                        ``portage`` python module as well as
                                        ``emerge(1)``)
 :py:mod:`~salt.modules.freebsdpkg`     FreeBSD-based OSes using ``pkg_add(1)``
 :py:mod:`~salt.modules.openbsdpkg`     OpenBSD-based OSes using ``pkg_add(1)``
-:py:mod:`~salt.modules.pacman`         Arch Linux-based distros using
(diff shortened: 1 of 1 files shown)


salt/modules/saltutil.py
```diff
@@ -1006,6 +1006,10 @@ def sync_all(saltenv=None, refresh=True, extmod_whitelist=None, extmod_blacklist
         salt '*' saltutil.sync_all saltenv=dev
         salt '*' saltutil.sync_all saltenv=base,dev
         salt '*' saltutil.sync_all extmod_whitelist={'modules': ['custom_module']}
+
+    .. note::
+        In masterless setups this function will only sync pillars when the `file_client`
+        option is set to `local` or `salt-call` is executed with the `--local` option.
```
Ch3LL: The definition of masterless is when a minion is run with either `file_client`  or `--local` set, so i'm not sure why this note is needed, since if you are in masterless mode these options are already set. Can you clarify the issue you ran into where the documentation did not make this clear enough?

Ch3LL (changes requested):

salt/modules/saltutil.py
ixs: You are normally running with a master.
But because you are trying to debug some pillar rendering problem with an ext_pillar you are running `salt-call --local --id=other.host.example.com pillar.get whatever`.
For this to work, you need to have the data from _pillar synced. So you do that by calling `salt-call saltutil.sync_all` and wonder why everything _but_ the _pillar dir are synced.

So you talk on slack and have the following conversation:

```
ixs 2:00 AM
I am trying to get an external pillar working. I have the code in /srv/salt/_pillar and salt-run saltutil.sync_all nicely copies it to /var/cache/salt/extmods/master/pillar/.
But sometimes I want to execute something masterless on the master as it gives way better debug info. Which is when I noticed that salt-call saltutil.sync_all does copy everything but pillars.
Is that the right behavior? I kinda feel it should copy pillars too... (edited) 

Matt 4:23 AM
Pillar is master only
Not exactly sure how masterless works though

gtmanfred 5:04 AM
You should be able to sync them to minions.
Either a) call saltutil.sync_pillar seperately, or patch the saltutil.sync_all to add sync_pillar to it, and then push a pr upstream, I think it is left out because the majority of setups don't need pillars synced on minions.

ixs 6:06 AM
@Matt Pillar works equally well in masterless setups. They just get executed same as any other modules. The only thing is, that you of course need to have access to the underlying data and the ext_pillar needs to be defined in the minion config.
@gtmanfred thanks. PR it is...

ixs 6:15 AM
ahh. The code is already there. So it will be a documentation patch...
    if __opts__['file_client'] == 'local':
        ret['pillar'] = sync_pillar(saltenv, False, extmod_whitelist, extmod_blacklist)
I did not run salt-call --local...
```

So that makes multiple people who were confused about masterless and pillars. A doc patch might be useful to save the next person from having to dig into the source.

Issue #46034, opened 2018-02-14
Using `!` as comment char with `file.comment_line` results in 'None' being used instead


hop:
The command

    salt '*' file.comment_line /etc/ca-certificates.conf '^mozilla/DigiCert_Global_Root_CA\.crt' '!'

results in

    Nonemozilla/DigiCert_Global_Root_CA.crt

in the file. In order to get the correct result, one has to use

    salt '*' file.comment_line /etc/ca-certificates.conf '^mozilla/DigiCert_Global_Root_CA\.crt' '"!"'

(the order of `'` and `"` is important to keep bash from interpreting the `!`).

I could trace this back you `yaml.load('!')` returning `None`, but I was not able to determine why the problem doesn't occur with `#`.

If this behaviour is intentional, it should at least be mentioned in the documentation.

hop:
Sorry, forget to mention that I use versions `2016.11.5` and `2016.11.9`.

I could not find a previous bug report for this, but maybe I couldn't come up with the right search terms.

garethgreenaway:
@hop As you pointed out this is actually your shell interpreting the `!` before Salt can get to it.  Since the `!` is used to call previous commands I'm not sure this is a scenario that we can work around and the behavior is expected.  Quoting the `!` as you've done would be the recommended solution.

hop:
No, it is not. Using single quotes prevents bash from interpreting the `!`. You can verify this by using double quotes instead.

github-abcde:
I have traced this to `salt/utils/args.py:yamlify_arg`.
In 2018.3, this happens:
```
>>> import salt.utils.yaml
>>> print(salt.utils.yaml.safe_load('!'))
None
>>>
```
In 2019.2, neon and develop, this happens:
```
>>> import salt.utils.yaml
>>> print(salt.utils.yaml.safe_load('!'))

>>>
```
This *only* happens when an arg consists of just an exclamation mark.
I'll see about creating PRs that fix this.

PR #54231, opened 2019-08-16
2018.3 fix 46034

github-abcde:
### What does this PR do?
This PR prevents `salt.utils.yaml.safe_load` from stomping arguments that only consist of a single exclamation mark.
Note that this behavior is different between 2018.3 on one side and 2019.2 and newer on the other. This PR should not be merged-forward. Different PR's will be created for newer versions of salt.

https://github.com/saltstack/salt/issues/46034

### Previous Behavior
When using `!` as an argument, like in 46034, `safe_load` changes this to `None` (NoneType), which gets coerced into the string `None` in `file.comment`.

### New Behavior
The argument `!` is kept as is.

### Tests written?
Yes

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/utils/args.py
+++ b/salt/utils/args.py
@@ -201,6 +201,9 @@ def yamlify_arg(arg):
             # will become None. Keep this value from being stomped as well.
             elif arg is None and original_arg.strip().startswith('#'):
                 return original_arg
+            # The condition below happens in 2018.3 only. Do not merge-forward.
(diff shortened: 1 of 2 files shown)


dwoz:
@github-abcde his needs to be re-opened against the `master` branch. We are no longer accepting PRs to `2018.3`. Sorry for the confusion.

PR #54232, opened 2019-08-16
2019.2 fix 46034

github-abcde:
### What does this PR do?
This PR prevents `salt.utils.yaml.safe_load` from stomping arguments that only consist of a single exclamation mark.

https://github.com/saltstack/salt/issues/46034

### Previous Behavior
When using `!` as an argument, like in 46034, `safe_load` changes this to `''` (empty string).

### New Behavior
The argument `!` is kept as is.

### Tests written?
Yes

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/utils/args.py
+++ b/salt/utils/args.py
@@ -198,7 +198,7 @@ def yamlify_arg(arg):
         elif arg is None \
                 or isinstance(arg, (list, float, six.integer_types, six.string_types)):
             # yaml.safe_load will load '|' as '', don't let it do that.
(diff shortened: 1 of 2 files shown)


github-abcde:
Ok, this is weird, locally the added tests succeed for python2.7 and python3.6.8.
However, the behavior I'm seeing here is similar to 2018.3 (`'!'` becomes `None`). Apparently this behavior is not linked to a certain salt version, but to a python version?

dwoz:
@github-abcde This needs to be re-opened against the `master` branch. We are no

PR #54241, opened 2019-08-16
replace pycrypto with pycryptodomex [2018.3]

JanZerebecki:
pycrypto is unmaintained and has open security issues.

Fixes: #54115

Forward port of https://github.com/saltstack/salt/pull/54119

### What does this PR do?
replace pycrypto with pycryptodomex, like it already is on many platforms

https://github.com/saltstack/salt/issues/54115

### Previous Behavior
require pycrpto

### New Behavior
require pycryptodomex

### Tests written?
No, drop in replacement of functionality.

### Commits signed with GPG?
Yes

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -10,7 +10,7 @@ PyYAML = "*"
 MarkupSafe = "*"
 requests = ">=1.0.0"
 tornado = ">=4.2.1,<5.0"
-pycrypto = ">=2.6.1"
+pycryptodomex = "*"
 pyzmq = ">=2.2.0"
 
 [dev-packages]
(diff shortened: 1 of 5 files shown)


JanZerebecki:
Many CI failures also happen without this patch.

s0undt3ch (changes requested):
Yes, we'd need to regenerate all static requirements.
```
pip install pre-commit
pre-commit run -av
```
Take a few cups of coffee.
First pre-commit run will show failures because files were updated, this is ok.
Second run should be all green.

dmurphy18 (changes requested):
These changes need to be migrated to 'master' branch for Neon or later. the 2018.3 branch is CVE only.
Some tests for a PR are required, if only to confirm that pycryptodomex installed, noting that have run into issues with pycryptodomex in 32-bit environments

s0undt3ch:
We have switched to pycryptodome for the upcoming Sodium release on https://github.com/saltstack/salt/pull/56625

Issue #54219, opened 2019-08-15
integration.shell.test_call.py and tests/integration/shell/test_proxy.py


KChandrashekhar:
### Description of Issue
integration.shell.test_call.py:    def test_issue_2731_masterless(self):
tests/integration/shell/test_proxy.py:            def test_exit_status_correct_usage

Issue #54153, opened 2019-08-08
Ubuntu 16.04 Proxy failures 


KChandrashekhar:
### Description of Issue
https://jenkinsci.saltstack.com/job/2019.2.1/view/Proxy/job/salt-ubuntu-1604-py2-proxy/

PR #54259, opened 2019-08-19
Call os.fork less to avoid race conditions

dwoz:

#54153
#54219

### Previous Behavior

- The minions' `_threaded_return` and `_threaded_multi_return` methods did a double fork so that we could join the parent process immediately, not blocking anything. 
- Schedules run by masters and minions did a double fork so that we could join the parent process immediately, not blocking anything.

### New Behavior

- The minions do not call join right away on job subprocesses instead they track and reap them when they are finished.
- Masters and minions do not join schedule processes right away, instead the Schedule class tracks it's subprocesses and provide a method for the users of this class to call periodically to reap them.

### Tests written?

Yes - Unit tests added for new classes

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/minion.py
+++ b/salt/minion.py
@@ -422,6 +422,41 @@ class MinionBase(object):
     def __init__(self, opts):
         self.opts = opts
 
+    def gen_modules(self, initial_load=False):
+        '''
+        Tell the minion to reload the execution modules
+
+        CLI Example:
+
+        .. code-block:: bash
+
+            salt '*' sys.reload_modules
+        '''
+        self.opts['pillar'] = salt.pillar.get_pillar(
+            self.opts,
+            self.opts['grains'],
+            self.opts['id'],
+            self.opts['saltenv'],
+            pillarenv=self.opts.get('pillarenv'),
+        ).compile_pillar()
+
+        self.utils = salt.loader.utils(self.opts)
+        self.functions = salt.loader.minion_mods(self.opts, utils=self.utils)
+        self.serializers = salt.loader.serializers(self.opts)
+        self.returners = salt.loader.returners(self.opts, self.functions)
+        self.proxy = salt.loader.proxy(self.opts, self.functions, self.returners, None)
+        # TODO: remove
+        self.function_errors = {}  # Keep the funcs clean
+        self.states = salt.loader.states(self.opts,
+                self.functions,
+                self.utils,
+                self.serializers)
+        self.rend = salt.loader.render(self.opts, self.functions)
+#        self.matcher = Matcher(self.opts, self.functions)
+        self.matchers = salt.loader.matchers(self.opts)
+        self.functions['sys.reload_modules'] = self.gen_modules
+        self.executors = salt.loader.executors(self.opts)
+
     @staticmethod
     def process_schedule(minion, loop_interval):
         try:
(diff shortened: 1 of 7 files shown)


salt/minion.py
```diff
@@ -422,6 +422,41 @@ class MinionBase(object):
...
+            self.opts['saltenv'],
+            pillarenv=self.opts.get('pillarenv'),
+        ).compile_pillar()
+
+        self.utils = salt.loader.utils(self.opts)
+        self.functions = salt.loader.minion_mods(self.opts, utils=self.utils)
+        self.serializers = salt.loader.serializers(self.opts)
+        self.returners = salt.loader.returners(self.opts, self.functions)
+        self.proxy = salt.loader.proxy(self.opts, self.functions, self.returners, None)
+        # TODO: remove
```
waynew: Should you remove this now, or could you provide a better explanation here about exactly why we want to remove it?

Akm0d (approved):

salt/minion.py
```diff
@@ -422,6 +422,41 @@ class MinionBase(object):
...
+        self.function_errors = {}  # Keep the funcs clean
+        self.states = salt.loader.states(self.opts,
+                self.functions,
+                self.utils,
+                self.serializers)
+        self.rend = salt.loader.render(self.opts, self.functions)
+#        self.matcher = Matcher(self.opts, self.functions)
```
waynew: Does this commented code need to be in here?

salt/minion.py
waynew: Aaaand  now I see that this block was just moved from somewhere else.

Which just validates my reasoning :trollface:

salt/minion.py
```diff
@@ -2659,21 +2616,36 @@ def setup_scheduler(self, before_connect=False):
...
+        return True
 
-        self.periodic_callbacks.update(new_periodic_callbacks)
+    def remove_periodic_callback(self, name):
+        '''
+        Remove a periodic callback.
+        If a callback by the given name does not exist this method returns False
+        '''
+        callback = self.periodic_callbacks.pop(name, None)
+        if callbback is None:
```
waynew:
```suggestion
        if callback is None:
```
too many b's! 🐝 🐝

salt/utils/schedule.py
```diff
@@ -659,9 +664,9 @@ def handle_func(self, multiprocessing_enabled, func, data):
             # Reconfigure multiprocessing logging after daemonizing
             log_setup.setup_multiprocessing_logging()
 
-        if multiprocessing_enabled:
-            # Don't *BEFORE* to go into try to don't let it triple execute the finally section.
-            salt.utils.process.daemonize_if(self.opts)
+        #if multiprocessing_enabled:
+        #    # Don't *BEFORE* to go into try to don't let it triple execute the finally section.
+        #    salt.utils.process.daemonize_if(self.opts)
```
waynew: Do we need this commented-out code around?

waynew (changes requested):

salt/minion.py
dwoz: I don't know the history of this line. What I did was move the method from the `Minion` class to the `MinionBase` class. So that it's available to all of the `MinionBase` subclasses. I didn't remove the commented out line so that whomever commented out can address it.

Having said that, I don't think I ended up using this method anywhere other than on the `Minion` so it's probably best to put it back where it was for now.

salt/minion.py
```diff
@@ -2659,21 +2616,36 @@ def setup_scheduler(self, before_connect=False):
...
-            new_periodic_callbacks['schedule'] = tornado.ioloop.PeriodicCallback(handle_schedule, 1000)
 
             if before_connect:
                 # Make sure there is a chance for one iteration to occur before connect
                 handle_schedule()
 
-        if 'cleanup' not in self.periodic_callbacks:
-            new_periodic_callbacks['cleanup'] = tornado.ioloop.PeriodicCallback(
-                    self._fallback_cleanups, loop_interval * 1000)
+            self.add_periodic_callback('schedue', handle_schedule)
```
s0undt3ch: `schedule`

salt/utils/process.py
```diff
@@ -850,3 +854,30 @@ def default_signals(*signals):
...
+            self.processes = processes
+        if lock is None:
+            self.lock = multiprocessing.Lock()
+        else:
+            self.lock = lock
+
+    def add(self, proc):
+        with self.lock:
+            self.processes.append(proc)
+            log.warning('Subprocess %s added', proc.name)
```
s0undt3ch: Keep warning log level?

salt/utils/process.py
```diff
@@ -850,3 +854,30 @@ def default_signals(*signals):
...
+
+    def cleanup(self):
+        with self.lock:
+            for proc in self.processes:
+                if proc.is_alive():
+                    continue
+                proc.join()
+                self.processes.remove(proc)
+                log.warning('Subprocess %s cleaned up', proc.name)
```
s0undt3ch: Same as above

s0undt3ch (changes requested):

salt/minion.py
dwoz: Actually, I moved it from `SMinion` to `MinionBase` so that instances of `Minion` have access to it.

salt/minion.py
dwoz: Again, I simply moved what was on `SMinion` to `MinionBase`

salt/utils/schedule.py
dwoz: Nope, good catch. Fixed

salt/minion.py
dwoz: Good catch, fixed.

salt/minion.py
dwoz: Good catch, fixed.

salt/utils/process.py
dwoz: No this should be debug

twangboy (approved):

waynew (approved):

s0undt3ch (approved):

PR #54264, opened 2019-08-20
Fix unit utils jinja when run on their own

dwoz:
### What does this PR do?

Fix an issue to make sure `unit.utils.test_jinja` tests will pass when they are run on their own

### Tests written?

No - Fixing unit tests

### Commits signed with GPG?

Yes

### DIFF
--- a/tests/unit/utils/test_jinja.py
+++ b/tests/unit/utils/test_jinja.py
@@ -116,6 +116,7 @@ class TestSaltCacheLoader(TestCase):
             self.template_dir
         )
         self.opts = {
+            'file_buffer_size': 1048576,
             'cachedir': self.tempdir,
             'file_roots': {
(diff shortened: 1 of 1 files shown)


s0undt3ch (approved):
This suggests a bigger problem.
Ie, if this addition makes tests pass when they run on their own, what "fills in" the missing options when they run with the rest of the test suite.

dwoz:
> This suggests a bigger problem.
> Ie, if this addition makes tests pass when they run on their own, what "fills in" the missing options when they run with the rest of the test suite.

That's definitely an issue. I've created a ticket (#54265) to go back and figure out how this is getting set.

Ch3LL (approved):

Akm0d (approved):

PR #54279, opened 2019-08-21
[2019.2.1] Don't even go through the transport to get the system grains

s0undt3ch:
### What does this PR do?
See title.

### Previous Behavior
Get's system grains from a running minion

### New Behavior
Get's system grains without daemons running

### Tests written?

No - Improves testing helpers decorator

### Commits signed with GPG?

Yes

### DIFF
--- a/tests/support/helpers.py
+++ b/tests/support/helpers.py
@@ -1066,7 +1066,9 @@ def requires_system_grains(func):
     @functools.wraps(func)
     def decorator(*args, **kwargs):
         if not hasattr(requires_system_grains, '__grains__'):
+            # Late import
             import salt.config
+            minion_id = 'requires_system_grains_minion_id'
             root_dir = tempfile.mkdtemp(dir=TMP)
             defaults = salt.config.DEFAULT_MINION_OPTS.copy()
             defaults.pop('conf_file')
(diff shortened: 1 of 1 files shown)


s0undt3ch:
Refs https://github.com/saltstack/salt/pull/54270

Akm0d (approved):

waynew (approved):

tests/support/helpers.py
```diff
@@ -1064,17 +1064,25 @@ def requires_system_grains(func):
...
-            if not hasattr(cls, 'run_function'):
-                raise RuntimeError(
-                    '{0} does not have the \'run_function\' method which is '
-                    'necessary to collect the system grains'.format(
-                        cls.__class__.__name__
-                    )
-                )
-            requires_system_grains.__grains__ = func(cls, grains=cls.run_function('grains.items'))
-        return requires_system_grains.__grains__
+            import salt.config
```
dwoz: Can we move this import to the top of the module?

dwoz (changes requested):

tests/support/helpers.py
s0undt3ch: The intent is to not trigger the import of any salt module when importing the helpers, which is also in use on other code in the module.

PR #54304, opened 2019-08-24
Add `salt.version.reqs`

dwoz:
### What does this PR do?

Adds `salt.version.reqs` to cut down on HAS_* usage. Once a requirement is defined in `salt.version` we can do this
```
if salt.version.reqs.msgpack:
    import msgpack
```
instead of
```
try:
     import msgpack
     HAS_MSGPACK = True
except ImportError:
    HAS_MSGPACK = False
```
The reqs class also supports version matching
```
if salt.version.reqs.msgpack >= 4.6:
     ...
```

### Tests written?

Yes

### Commits signed with GPG?

Yes


dwoz:
re-run full all

twangboy (dismissed):

cmcmarrow (dismissed):

salt/utils/__init__.py
```diff
@@ -13,6 +13,7 @@
 
 # Import Salt libs
 from salt.defaults import DEFAULT_TARGET_DELIM
+from salt.utils.dependencies import depends
```
s0undt3ch: You're suggesting we should import depends from `salt.utils`.
Besides namespacing, the reason why we now have `salt.utils.files` and the like is to improve import time speed.
This would go against that work.

salt/utils/dependencies.py
```diff
@@ -0,0 +1,144 @@
...
+            raise DependencyNotRegistered('Unknown dependency: {}'.format(val))
+        module_getter, version_getter = self.deps_map[val]
+        return Dependency(
+            val,
+            module_getter or default_module_getter,
+            version_getter or default_version_getter,
+        )
+
+
+depends = Dependencies()
```
s0undt3ch: there's already a [`depends`](https://github.com/saltstack/salt/blob/neon/salt/utils/decorators/__init__.py#L33-L217) decorator, `dependencies`? `requirements`?

```
from salt.utils.dependencies import requirements
if requirements.gnupg >= '1.3.1':
   ...
```
doesn't look that bad...

s0undt3ch (commented):
I also think this should not go into 2019.2.1

s0undt3ch:
For non complicated requirements, those that already provide for example `__version__` in the package, I'd like to be able to use something like:
```python
fobbarmod = requirements.pkg('foobarmod', version='2.0')
```
Or even better:
```python
foobarmod = requirements.foobarmod('2.0')
```
ie, without an specific entry in the packages mapping, in case we really don't need any special checks of the package at import time.

salt/utils/__init__.py
dwoz: I think this will be used often enough (450+ times across the code base) to warrant allowing this import.

dwoz:
> For non complicated requirements, those that already provide for example `__version__` in the package, I'd like to be able to use something like:
> 
> ```python
> fobbarmod = requirements.pkg('foobarmod', version='2.0')
> ```
> 
> Or even better:
> 
> ```python
> foobarmod = requirements.foobarmod('2.0')
> ```
> 
> ie, without an specific entry in the packages mapping, in case we really don't need any special checks of the package at import time.

I made the packages mapping explicit rather than implicit. It might mean we the mapping grows quite large over time but it has the benefit of being able to see all of our dependencies in one place.

salt/utils/__init__.py
s0undt3ch: Sure, often enough, but it still doesn't remove the fact that it will trigger a bunch of other imports which might not matter for most of those, 450+ occurrences.

dwoz: The only other thing its causing to be imported is `salt.utils.versions.LooseVersion`

salt/utils/dependencies.py
dwoz: We also have `salt.versions` I could see putting this there too.

s0undt3ch: That would also make sense, and I wouldn't think it would cause the same speed penalty as importing it in `salt.utils`, ie the `from salt.utils.dependencies import depends` there

salt/utils/__init__.py
dwoz: I moved this from `salt.utils.depends` to `salt.version.reqs`

salt/utils/dependencies.py
dwoz: I moved this from salt.utils.depends to salt.version.reqs

tests/unit/test_version.py
```diff
@@ -71,3 +83,109 @@ def test_unparsable_version(self):
...
+
+    def test_default_version_getter___version__(self):
+        class mock:
+            __version__ = '1.2.1'
+        assert default_version_getter(mock) == '1.2.1'
+
+    def test_default_version_getter_str(self):
+        class mock:
+            __version__ = '1.2.1'
+        assert default_version_getter(mock) == '1.2.1'
```
waynew: @dwoz Am I missing something? What's the difference between `getter_str(self)` and `getter__version__(self)` (the test before this?

Ch3LL (changes requested):

salt/version.py
```diff
@@ -788,3 +840,141 @@ def msi_conformant_version():
...
+
+    def __init__(self, deps_map=None):
+        if deps_map is None:
+            self.deps_map = DEPS_MAP
+        else:
+            self.deps_map = deps_map
+
+    def __getattr__(self, val):
+        if val not in self.deps_map:
+            raise RequirementNotRegistered('Unknown dependency: {}'.format(val))
```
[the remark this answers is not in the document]
dwoz: This is *not* raising an exception if the requirement failed to import. It's raising an exception if we try to use a requirement that has not been defined in `deps_map`. I think failing fast in this case is the correct route to take.

Again for the exception above. If we run into that case it means we've poorly coded something. We expect a version variable to exist based on what we define in the `version_getter` for any module. If it does not, it's a coding error and so failing fast is intended. Although, it should be a more descriptive Exception, I will change that.

Ch3LL (dismissed):

tests/unit/test_version.py
dwoz: Yeah, good catch. I just removed the duplicate test.

raddessi:
Is this dead or still something that would be welcomed?

dwoz:
> Is this dead or still something that would be welcomed?

Still not dead, I (or someone) just needs to find the time to get it across the finish line

Ch3LL:
I think this would be good to get in. How about we close this PR and open an issue, assigned to a developer and slot it to get fixed for a release?

salt/version.py
```diff
@@ -935,6 +939,156 @@ def versions_report(include_salt_cloud=False, include_extensions=True):
...
+                self.has_depend = False
+
+    def __nonzero__(self):
+        return self.has_depend
+
+    def __bool__(self):
+        return self.has_depend
+
+    def __eq__(self, other):
+        print(self.version)
```
twangboy, line 1006: Probably left over from debugging

tests/pytests/unit/test_version.py
```diff
@@ -558,3 +568,116 @@ def test_parsed_version_name(version_str, expected_str, expected_name):
...
+
+def test_default_version_getter_tuple():
+    class mock:
+        __version__ = (1, 2, 1)
+
+    assert default_version_getter(mock) == "1.2.1"
+
+
+@pytest.fixture
+def test_module(tmp_path):
```
twangboy, line 606: Is this actually a test? Or just a fixture? I thought only tests should begin with `test_`

twangboy (changes requested):

twangboy (dismissed):

twangboy (dismissed):

twangboy:
Seeing this in all the unit tests on Windows:
```
tests/pytests/unit/serializers/test_serializers.py - AttributeError: module 'salt.serializers.msgpack' has no attribute 'available'
```

twangboy (dismissed):

salt/version.py
s0undt3ch, line 1: I'd prefer to have this code somewhere in utils, `salt.utils.versions` ? 
This module is already too long and does too much.

dwoz: I'm good with that. Moved.

twangboy:
Looks like the lint failure is because it's trying to write contents to a file from a variable that no longer exists (`TEST_MOD`). I'm assuming that's supposed to contain the contents of a module to be used for testing.

twangboy (dismissed):

Issue #53931, opened 2019-07-19
integration.states.test_pip_state.PipStateTest.test_issue_6912_wrong_owner_requirements_file


KChandrashekhar:
### Description of Issue
**Find a solution when artifacts isn't available** 
Failure on https://jenkinsci.saltstack.com/job/2019.2.1/job/salt-ubuntu-1604-py2/76/

bryceml:
I believe the issue was codecov was down.  This is fixed on branch jobs, we need to fix it on pr jobs at some point.  This was the change to fix it: https://github.com/saltstack/sre-jenkins/commit/04a1daa86e2f6d2fcab6e0c3e5e66ff9e4b4fd09

DmitryKuzmenko:
It looks it's not fixed.
https://jenkinsci.saltstack.com/job/2019.2.1/job/salt-windows-2016-py2/179/
https://jenkinsci.saltstack.com/job/2019.2.1/job/salt-windows-2019-py2/180/

sagetherage:
closing if this isn't fixed - then open a new ticket with latest from Jenkins

PR #54319, opened 2019-08-27
#53931 never fail on codecov curls

felippeb:
What does this PR do?

adds || true to codecov curls so we don't fail if they are down.

#53931
Previous Behavior

codecov curl statements could cause a test run to fail if codecov is down
New Behavior

bash will now return 0 for this curl even if codecov is down
Tests written?

[NOTICE] Bug fixes or features added to Salt require tests.
Please review the test documentation for details on how to implement tests into Salt's test suite.

No
Commits signed with GPG?

Yes

Please review Salt's Contributing Guide for best practices.

See GitHub's page on GPG signing for more information about signing commits with GPG.


bryceml (approved):

dwoz:
@felippeb @bryceml If this is still something that is needed please re-open it against master

Issue #54256, opened 2019-08-19
Empty salt api token file breaks maintenance process.


sleminov-tc:
### Description of Issue
Maintenance process that runs every minute (guided by loop_interval) is failing

```
[salt.utils.process:754 ][ERROR   ][5766] An un-handled exception from the multiprocessing process 'Maintenance-4443' was caught:
Traceback (most recent call last):
File "/usr/lib/python2.7/dist-packages/salt/utils/process.py", line 747, in _run
return self._original_run()
File "/usr/lib/python2.7/dist-packages/salt/master.py", line 224, in run
salt.daemons.masterapi.clean_expired_tokens(self.opts)
File "/usr/lib/python2.7/dist-packages/salt/daemons/masterapi.py", line 138, in clean_expired_tokens
token_data = loadauth.get_tok(tok)
File "/usr/lib/python2.7/dist-packages/salt/auth/__init__.py", line 235, in get_tok
tdata = self.tokens["{0}.get_token".format(self.opts['eauth_tokens'])](self.opts, tok)
File "/usr/lib/python2.7/dist-packages/salt/tokens/localfs.py", line 67, in get_token
tdata = serial.loads(fp_.read())
File "/usr/lib/python2.7/dist-packages/salt/payload.py", line 145, in loads
ret = msgpack.loads(msg, use_list=True, ext_hook=ext_type_decoder, encoding=encoding)
File "msgpack/_unpacker.pyx", line 146, in msgpack._unpacker.unpackb (msgpack/_unpacker.cpp:2231)
UnpackValueError: Unpack failed: error = 0
[salt.utils.process:435 ][INFO    ][30130] Process <class 'salt.master.Maintenance'> (5766) died with exit status 1, restarting...
```

As result non of the scheduled task are run. The issue is caused by 0 bytes token file in
`/var/cache/salt/master/tokens`

Removing the file fixes the problem

The related issue https://github.com/saltstack/salt/issues/37945 which was fixed by 
https://github.com/saltstack/salt/commit/330021cd8b6205861a4a51982da4bc90125a1be1.

The routine however was rewritten to not handle this Exception. 

So looks like regression. 

### Steps to Reproduce Issue
Periodically strikes. Not clear what causes these empty token files to be created.  

### Versions Report
```
Salt Version:
           Salt: 2018.3.4
 
Dependency Versions:
           cffi: Not Installed
       cherrypy: 3.2.2
       dateutil: 2.5.3
      docker-py: Not Installed
[...]

Issue #37945, opened 2016-11-29
Missing exception handling in salt.master.Maintenance. Process never completes.


gstachowiak:
### Description of Issue/Question

I noticed the following error in salt master logs. It was repeated many times.

```
2016-11-27 03:12:16,885 [salt.config                                          ][DEBUG   ][21595] Reading configuration from /etc/salt/master
2016-11-27 03:12:16,933 [salt.config                                          ][DEBUG   ][21595] Using cached minion ID from /etc/salt/minion_id: rbalvcdstestm.bas.roche.com
2016-11-27 03:12:16,935 [salt.config                                          ][DEBUG   ][21595] Reading configuration from /etc/salt/master
2016-11-27 03:12:17,167 [salt.utils.lazy                                      ][DEBUG   ][21595] LazyLoaded local_cache.clean_old_jobs
2016-11-27 03:12:17,168 [salt.payload                                         ][CRITICAL][21595] Could not deserialize msgpack message.This often happens when trying to read a file not in binary modeTo see message payload, enable debug logging and retry. Exception: Unpack failed: error = 0
2016-11-27 03:12:17,168 [salt.payload                                         ][DEBUG   ][21595] Msgpack deserialization failure on message: 
2016-11-27 03:12:17,168 [salt.utils.process                                   ][ERROR   ][21595] An un-handled exception from the multiprocessing process 'Maintenance-2441' was caught:
Traceback (most recent call last):
  File "/usr/lib/python2.6/site-packages/salt/utils/process.py", line 616, in _run
    return self._original_run()
  File "/usr/lib/python2.6/site-packages/salt/master.py", line 238, in run
    salt.daemons.masterapi.clean_expired_tokens(self.opts)
  File "/usr/lib/python2.6/site-packages/salt/daemons/masterapi.py", line 149, in clean_expired_tokens
    token_data = serializer.loads(token_file.read())
  File "/usr/lib/python2.6/site-packages/salt/payload.py", line 117, in loads
    return msgpack.loads(msg, use_list=True)
  File "msgpack/_unpacker.pyx", line 145, in msgpack._unpacker.unpackb (msgpack/_unpacker.cpp:145)
UnpackValueError: Unpack failed: error = 0
2016-11-27 03:12:17,173 [salt.utils.process                                   ][INFO    ][7911] Process <class 'salt.master.Maintenance'> (21595) died with exit status None, restarting...
[...]

PR #54325, opened 2019-08-27
Delete bad API token files

waynew:
### What does this PR do?

#54324 to 2019.2.1 - Deletes eauth token files if they're empty (or otherwise fail to be read)

#54256 
#37945

### Previous Behavior

If the token file was empty or we otherwise failed to deserialize the token it would cause an exception. This was a regression.

### New Behavior

If the token file is empty or we otherwise fail to deserialize the token we delete the invalid token file.

### Tests written?

Yes - not just ones to cover this regression, but also added a couple more to cover this whole function.

### Commits signed with GPG?

Yes

---

I have one question about this, though - right now I'm capturing `Exception` when trying to load the token file. This seems like it *could* be overly broad - OTOH, it's also probably accurate. For an IOError or OSError it will end out taking a different path (i.e. that load will just return an empty dict that gets returned, which bypasses the file removal process).

Any thoughts?


garethgreenaway (approved):

dwoz (approved):

Akm0d (approved):

salt/tokens/localfs.py
```diff
@@ -10,9 +10,10 @@
 import os
 import logging
 
+import salt.exceptions
```
s0undt3ch: Is this now an unused import?

twangboy (approved):

KChandrashekhar:
Close the Pull request

PR #54345, opened 2019-08-29
Handle exception in `salt.utils.args.format_call()`

eliasp:
### What does this PR do?

Without handling this exception, a SaltClient might hang in a loop waiting for a timeout and providing no clue as to why the call hangs/fails.
Handling this exception and logging a corresponding error msg improves this situation.

### Previous Behavior
A SaltClient would hang/loop like this:
```
[DEBUG   ] Connecting the Minion to the Master URI (for the return server): tcp://127.0.0.1:54519
[DEBUG   ] Trying to connect to: tcp://127.0.0.1:54519
[DEBUG   ] SaltReqTimeoutError, retrying. (1/3)
[DEBUG   ] SaltReqTimeoutError, retrying. (2/3)
[DEBUG   ] SaltReqTimeoutError, retrying. (3/3)
```
### New Behavior
**CLI:**
```
Authentication failure of type "eauth" occurred for user $USER
```
**Master Log:**
```
[ERROR   ][354] 'format_call' failed: SaltInvocationError('auth takes at least 2 arguments (1 given)',)
```

### Tests written?
No

As it's still quite unclear why we're seeing a certain external auth issue here, there's no way yet to define a reproducable positive/negative scenario which could be used for testing this.
This fix just helps us to further narrow down our problem and ensures no-one else will waste time on the red herring `SaltReqTimeoutError` in this scenario anymore.

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/auth/__init__.py
+++ b/salt/auth/__init__.py
@@ -20,6 +20,7 @@ import time
 import logging
 import random
 import getpass
+from salt.exceptions import SaltInvocationError
 from salt.ext.six.moves import input
 from salt.ext import six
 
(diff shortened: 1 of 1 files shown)


eliasp:
The [2 codeclimate issues](https://codeclimate.com/github/saltstack/salt/pull/54345) are IMHO out-of-scope for this PR.

cmcmarrow (approved):
Thanks for adding this

eliasp:
I added 2 more cases in `salt/auth/__init__.py` of "blind" calls of `salt.utils.args.format_call()`.

dwoz:
@eliasp We are no longer accepting PRs to the `develop` branch. This needs to be re-opened against the `master` branch. You can link to this PR from the new one to preserve any disscussion history. Sorry for the confusion.

eliasp:
Now filed against `master` in #55731

Issue #47488, opened 2018-05-04
MySQL Comments Can Cause mysql_query.run_file to Fail


doc987:
**Comments with ```--```:** Comments in sql files that start with ```--``` and continue to the end of the line are not recognized when by ```salt.states.mysql_query.run_file```.  That state calls ```salt.modules.mysql.file_query```, which uses ```re.match``` to ignore lines that _begin_ with ```--```.  However, comments that are later on the line are not ignored, but instead are included in the ```query_string```.  Since ```query_string``` is a concatenation of the lines in the file, a ```--``` will comment out the rest of the concatenated query.

**Comments with ```#```:** These can't be used with ```run_file``` because they are not recognized by ```run_file```.  MySQL will recognize them, but I don't think they're standard SQL.

**Comments with ```/* ... */```:** These seem to work okay because concatenating lines doesn't break this type of comment.

**Comments Containing a Semicolon:** These can't be used inside a comment in a query with ```run_file``` because they're interpreted as ending the query, whether using ```--```, ```#```, or ```/* ... */```.

**Error Messages:** One result of the lines being concatenated is the any syntax error (caused by this, or otherwise) always reports a problem on line 1, which can make it difficult to locate the problem.  When I tried running a file directly (```mysql < input.sql```), it ran with no issues, but running the same file with ```run_file``` failed.

Salt Version: 2017.7.3

gtmanfred:
Thanks for reporting this.  A pr to fix it would be greatly appreciated.

Thanks
Daniel

doc987:
Some additional code could be written to parse the input better (which may be able to handle most cases).  If one wanted to handle *any* SQL provided, I think an actual SQL parser would be needed.  Some possible approaches are:
1. Write a new SQL parser.
2. Use an existing SQL parser.  One option is https://github.com/andialbrecht/sqlparse.  Splitting a string into individual statements can be done as shown below.
```python
import sqlparse
statements = sqlparse.split(sql_string)
```
3. Don't parse the SQL.  Send the whole file directly to MySQL, and let MySQL do the parsing.

gtmanfred:
I think that 2 or 3 seems fine, if you can switch it to just sending the whole sql file in python i think that would be the better option.

doc987:
[...]

PR #54371, opened 2019-08-31
Better support comments in mysql files

mchugh19:
### What does this PR do?
Better handles comments in mysql files. 
Also adds a no_parse flag to pass the entirety of the mysql file contents to mysql. This allows mysql to handle its own parsing, which might work better for some complex setups, but also prevents salt from returning specifics of what has changed.

#47488

### Previous Behavior
A file like
```
-- some comment
INSERT INTO users (username) VALUES ("sampleuser2");  # after
#another comment
INSERT INTO users (username) VALUES ("sampleuser3"); 
```
would be interpreted as
```
INSERT INTO users (username) VALUES ("sampleuser2");
# after #another comment INSERT INTO users (username) VALUES ("sampleuser3");
```

### New Behavior
Now that file is interpreted as
```
INSERT INTO users (username) VALUES ("sampleuser2");
INSERT INTO users (username) VALUES ("sampleuser3");
```

When no_parse is used, the execution module will return something like the following for files which contain many create, insert, and update statements.
```
local:
    ----------
    query time:
        ----------
        human:
            0.0s
        raw:
            0.00042
    rows affected:
        1  
```
### Tests written?
Yes - existing test updated to support this functionality

### Commits signed with GPG?
No


mchugh19:
Rebased on master

DmitryKuzmenko:
@mchugh19 thank you for this work! It's awesome!
It's not a full support of MySQL/MariaDB comments style, but it's much better than it was.
The only thing I want to ask you. You're using regex that is as powerful as hardly readable and supportable. Meanwhile our tests don't cover your functionality. You've added a generic regression update to an existing integration test. But we definitely need tests:
1. checking your regular expressions in all combinations of data;
2. covering the `no_parse` case.

There's already a unit test for mysql module so it'd be pretty simple to extend this.
Are you able to extend the test to your functionality please?

DmitryKuzmenko:
Additional thanks for blacken your code!

mchugh19:
Never say nothing good came out of the test cases! Since getting back to this, I was not able to get the `no_parse` feature to work with these libraries. Since it's that flakey, I just removed the whole thing. An additional unit test was written for the comment handling.

tests/unit/modules/test_mysql.py
```diff
@@ -620,6 +620,29 @@ def test_plugin_status(self):
...
+        Test comment sanitization
+        """
+        input_data = """/*
+        multiline
+        comment
+        */
+        CREATE TABLE test_update (a INT); # end of line comment
+        # example comment
+        insert into test_update values (1); -- ending comment
+        -- another comment type
```
DmitryKuzmenko, line 634: Just one more case please-please. A comment sign inside the quotes.

DmitryKuzmenko (changes requested):
Please understand me correctly I'm not trying to bore you. =) Thanks for understanding!

tests/unit/modules/test_mysql.py
mchugh19: Is this what you meant? (see commit)

DmitryKuzmenko: Perfect! Thank you!

DmitryKuzmenko (approved):

DmitryKuzmenko:
This PR is blacken and merge ready.

PR #54379, opened 2019-09-02
zypperpkg: do not interfere with advisory_ids

aplanas:
When a patch is installed via advisory_ids, we do not want to change the
behavior of list_pkgs, via the includes field.

Currently, if a patch is installed via advisory_ids (state
pkg.patch_installed), it will call zypperpkg.install with this parameter
inside kwargs. This parameter is parsed by pkg_resource.parse_targets,
and the prefix `patch:` will be added very early, just before taking the
list of the current packages.

This last step will change the behavior of list_pkgs, as all the other
patches will be listed.

This sound OK except when we want to maintain the old behavior, when
patches are not listed in the `install` function by default.

This patch will add the `patch:` prefix in the list of targets later,
once we have the list of packages, restoring the old behavior.

Note that we can still install a patch as a normal package, if we add
the prefix in the name before calling `install`, and we do not use the
`advisory_ids` field.

### Tests written?

No. This is a partial revert of https://github.com/saltstack/salt/pull/52412, so this should be covered by the current tests.

### DIFF
--- a/salt/modules/zypperpkg.py
+++ b/salt/modules/zypperpkg.py
@@ -1467,7 +1467,12 @@ def install(name=None,
             if advisory_id not in cur_patches:
                 raise CommandExecutionError('Advisory id "{0}" not found'.format(advisory_id))
             else:
-                targets.append('patch:{}'.format(advisory_id))
+                # If we add here the `patch:` prefix, the
+                # `_find_types` helper will take the patches into the
+                # list of packages. Usually this is the correct thing
+                # to do, but we can break software the depends on the
+                # old behaviour.
+                targets.append(advisory_id)
     else:
         targets = pkg_params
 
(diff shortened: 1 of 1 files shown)


aplanas:
@brejoc @dincamihai alternative fix without adding an extra parameter (not validated in SUMA)

dincamihai:
> Note that we can still install a patch as a normal package, if we add
the prefix in the name before calling install, and we do not use the
advisory_ids field.

does this mean we have to adapt all the calls to pkg.install and append `patch:` prefix when we install a patch?

aplanas:
> > Note that we can still install a patch as a normal package, if we add
> > the prefix in the name before calling install, and we do not use the
> > advisory_ids field.
> 
> does this mean we have to adapt all the calls to pkg.install and append `patch:` prefix when we install a patch?

No.

This patch revert the part from https://github.com/saltstack/salt/pull/52412 that interfere with the `pkg.patch_installed` state.

Before this patch or the previous one you always can use `salt-call pkg.install 'patch:mypatch` or `salt-call pkg.install 'pattern:mypatch'`. The pattern or the patch will be installed, but the comparison at the end of the module will fail. This is what was fixed in https://github.com/saltstack/salt/pull/52412

xeacott (approved):

aplanas:
I moved the patch into https://github.com/saltstack/salt/pull/54954, as this is a fix related with this change. With this I consider this patch rebased into develop. Please, review on the new place.

PR #54384, opened 2019-09-03
Fix pyenv installation into the state.pyenv.install_pyenv

### LINKED ISSUE #37648: salt.states.pyenv.install_pyenv broken since at least 2015.8.8

The `pyenv.install_pyenv` state [uses the wrong signature when calling `_check_and_install_python()`](https://github.com/saltstack/salt/blob/develop/salt/states/pyenv.py#L219), passing `user` in for the `python` arg. Since `user` defaults to `None`, this causes a fun-to-debug traceback when it later tries to run a regex search against `python`:

```
----------
          ID: install-pyenv
    Function: pyenv.install_pyenv
      Result: False
                File "/usr/lib/python2.7/dist-packages/salt/state.py", line 1626, in call
                  **cdata['kwargs'])
                File "/usr/lib/python2.7/dist-packages/salt/loader.py", line 1492, in wrapper
                  return f(*args, **kwargs)
                File "/usr/lib/python2.7/dist-packages/salt/states/pyenv.py", line 215, in install_pyenv
                  return _check_and_install_python(ret, user)
                File "/usr/lib/python2.7/dist-packages/salt/states/pyenv.py", line 84, in _check_and_install_python
                  if __salt__['pyenv.install_python'](python, runas=user):
                File "/usr/lib/python2.7/dist-packages/salt/modules/pyenv.py", line 159, in install_python
                  python = re.sub(r'^python-', '', python)
                File "/usr/lib/python2.7/re.py", line 151, in sub
                  return _compile(pattern, flags).sub(repl, string, count)
              TypeError: expected string or buffer
     Started: 12:14:18.916348
    Duration: 2.586 ms
     Changes:   
```

### Setup

This state will adequately demonstrate the bug:

```
    pyenv-install:
      pyenv.install_pyenv
```

### Versions Report

```
$ salt --versions-report
Salt Version:
           Salt: 2015.8.8
 
Dependency Versions:
         Jinja2: 2.7.2
       M2Crypto: Not Installed
           Mako: Not Installed
         PyYAML: 3.10
          PyZMQ: 14.0.1
         Python: 2.7.6 (default, Jun 22 2015, 17:58:13)
           RAET: Not Installed
        Tornado: 4.2.1
            ZMQ: 4.0.5
           cffi: 1.8.3
       cherrypy: Not Installed
       dateutil: 2.6.0
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
        libgit2: Not Installed
        libnacl: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.4.6
   mysql-python: 1.2.3
      pycparser: 2.17
       pycrypto: 2.6.1
         pygit2: Not Installed
   python-gnupg: Not Installed
          smmap: Not Installed
        timelib: Not Installed
 
System Versions:
           dist: Ubuntu 14.04 trusty
        machine: x86_64
        release: 3.13.0-30-generic
         system: Ubuntu 14.04 trusty
```

We ran into the problem on a masterless VM.

/cc @ryanmrubin

bayaro:
### What does this PR do?
Add real request for pyenv installation inside then states.pyenv.install_pyenv

Fixes #37648

### Previous Behavior
The function states.pyenv.install_pyenv just requested an installation of python with the username for pyenv instead of a real python version. So when it was called it always returns error.

### New Behavior
if pyenv is not installed the function requests the installation

### Tests written?
No

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/states/pyenv.py
+++ b/salt/states/pyenv.py
@@ -61,6 +61,9 @@ def _check_pyenv(ret, user=None):
     if not __salt__['pyenv.is_installed'](user):
         ret['result'] = False
         ret['comment'] = 'pyenv is not installed.'
+    else:
+        ret['result'] = True
+        ret['comment'] = 'pyenv is installed.'
     return ret
 
 
--- a/salt/states/pyenv.py
+++ b/salt/states/pyenv.py
@@ -216,4 +216,10 @@ def install_pyenv(name, user=None):
         ret['comment'] = 'pyenv is set to be installed'
         return ret
 
-    return _check_and_install_python(ret, user)
+    ret =  _check_pyenv(ret, user)
+    if not ret['result'] is True:
+        if __salt__['pyenv.install'](user):
+            ret['result'] = True
+            ret['comment'] = 'Successfully installed pyenv'
+
+    return ret
--- a/salt/states/pyenv.py
+++ b/salt/states/pyenv.py
@@ -220,7 +220,7 @@ def install_pyenv(name, user=None):
         return ret
 
     ret =  _check_pyenv(ret, user)
-    if not ret['result'] is True:
+    if not ret['result']:
         if __salt__['pyenv.install'](user):
             ret['result'] = True
             ret['comment'] = 'Successfully installed pyenv'


bayaro:
I have started branch from merge-base for develop & 2018.3 branches

salt/states/pyenv.py
```diff
@@ -216,4 +216,10 @@ def install_pyenv(name, user=None):
         ret['comment'] = 'pyenv is set to be installed'
         return ret
 
-    return _check_and_install_python(ret, user)
+    ret =  _check_pyenv(ret, user)
+    if not ret['result'] is True:
```
waynew: This would be better written:

```suggestion
    if not ret['result']:
```

No need to check against `True`. Also, `is` is an identity comparison.

waynew (changes requested):
Thanks for the PR!

A minor change here, and also would you be able to write some regression test(s) for this? It looks like you could probably just mock out a few things to get a basic test in there.

salt/states/pyenv.py
bayaro: Thanks for the review!

Yeap, philosophically you are right, but in practice the result must be equal only  the True, not any non empty string, not a number, etc. And such situation is possible in any non-typed language.

Unfortunately, I'm not familiar with the project test system. Could you please add some dummy test to the PR, so I'll add a real test.

waynew: Philosophically, you're also right, but practically speaking Salt internals should (MUST?) return `True`, `False`, or `None` for the result.

In fact this code is trivial enough to follow - https://github.com/saltstack/salt/blob/8eecd9c69f41c2ef227318f73c228d0f541f58cb/salt/states/pyenv.py#L57

You're setting `result` to `None`, and it can only ever be set to `False` in that function, which actually makes me wonder what this test is even meant to check? It will never ever be `True` or anything even remotely truthy, until after this line, if I'm reading this correctly.

bayaro: > Philosophically, you're also right, but practically speaking Salt internals should (MUST?) return `True`, `False`, or `None` for the result.

 The whole file is _not_ my code.  I just added as short as possible patch. I do not understand who must or should?
> 
> In fact this code is trivial enough to follow -
> 
> https://github.com/saltstack/salt/blob/8eecd9c69f41c2ef227318f73c228d0f541f58cb/salt/states/pyenv.py#L57
> 
> You're setting `result` to `None`, and it can only ever be set to `False` in that function, which actually makes me wonder what this test is even meant to check? It will never ever be `True` or anything even remotely truthy, until after this line, if I'm reading this correctly.

 It's NOT my code.

Anyway, thank you, this's a good note. I forgot to add another part of patch ))) Another part returns True in case pyenv is installed.

Generally speaking, nobody knows what other function will return.

waynew: https://docs.saltstack.com/en/latest/ref/states/writing.html#return-data

> result: A tristate value. True if the action was successful, False if it was not, or None if the state was run in test mode, test=True, and changes would have been made if the state was not run in test mode.

We shouldn't worry if `result` is ever anything but `True`, `False`, or `None`. 

```
$ ag 'if ret\[.result.\]:' -c
tests/unit/states/test_module.py:1
salt/cloud/clouds/lxc.py:1
salt/states/memcached.py:2
salt/states/elasticsearch_index_template.py:1
salt/states/elasticsearch.py:5
salt/states/elasticsearch_index.py:1
salt/states/user.py:1
salt/states/webutil.py:1
salt/states/selinux.py:1
salt/states/win_path.py:2
salt/states/cron.py:1
salt/states/ports.py:1
salt/states/logrotate.py:2
salt/states/virt.py:1
salt/states/at.py:1
salt/states/zfs.py:9
salt/states/win_system.py:1
salt/states/win_pki.py:2
salt/states/reg.py:1
salt/states/http.py:1
salt/states/zone.py:12
salt/states/zpool.py:4
salt/states/group.py:1
salt/states/file.py:3
salt/states/logadm.py:2
salt/states/rbenv.py:2
salt/states/mdadm_raid.py:2
salt/states/grains.py:2
salt/states/pyenv.py:1
salt/states/smartos.py:5
salt/states/jboss7.py:1
salt/modules/nagios_rpc.py:2
salt/modules/dockermod.py:1
```

There *are* other places in the code where we use `is True`:

```
$ ag 'if ret\[.result.\] is True:' -c
salt/states/lxc.py:1
salt/states/heat.py:2
salt/states/timezone.py:1
salt/modules/heat.py:2
```

Out of those, I'd say that only one of those files *might* be following a reasonable pattern, but even that one still feels a bit odd.

> Generally speaking, nobody knows what other function will return.

My understanding is that any salt functions that are returning a `'result'` value outside of `True, False, or None` should be considered buggy and fixed.

bayaro: Just for interest. Could you please also grep for 'if.*is True'?

bayaro: It's a first time in my life when somebody states that implicit check is better than explicit.

waynew: https://docs.quantifiedcode.com/python-anti-patterns/readability/comparison_to_true.html

Very *very* rarely will you see Python code checking for *absolute* "Truth". As a matter of fact, I've seen more `is True` in Salt that I've seen since Python 2.4!

In other languages you check for equality with true and false, but Python code has a strong preference for relying on the boolean nature of things, because it's rare that you need to be so explicit. And I would argue that it means that you wrote your code wrong.

bayaro: 1) I have used not a comparison but the strong check!
2) 
> I've seen more is True in Salt that I've seen since Python 2.4!
because Salt was the devops tool (and now I'm doubt if it will stay as it is and not become a garbage of the states). We can suggest everything that we want, but if it's possible to go by a wrong way it will definitely go by the wrong way. So an explicit declaration is always preferable (despite most people are lazy and don't use it).
3) in other languages they also have special mechanisms for the strong comparison with True or False. And because of many languages (especially nontyped) have such mechanism I think they know something )))

Issue #54393, opened 2019-09-03
setUpClass (integration.states.test_file.RemoteFileTest)


KChandrashekhar:
### Description of Issue
https://jenkinsci.saltstack.com/job/2019.2.1/view/Tornado/job/salt-centos-7-py2-tornado/191/

Error Message

Failed to start tornado webserver on 127.0.0.1:51564 within 5 seconds
Stacktrace

Traceback (most recent call last):
  File "/tmp/kitchen/testing/tests/integration/states/test_file.py", line 4024, in setUpClass
    cls.webserver.start()
  File "/tmp/kitchen/testing/tests/support/helpers.py", line 1611, in start
    '{1} seconds'.format(self.port, self.wait)
Exception: Failed to start tornado webserver on 127.0.0.1:51564 within 5 seconds

PR #54400, opened 2019-09-04
[2019.2.1] Retry starting the testing webserver at least 3 times before failing

s0undt3ch:
### What does this PR do?
See title.

#54393 

### Previous Behavior
We would only try once

### New Behavior
We try at least 3 times

### Tests written?

No - This is testing support code

### Commits signed with GPG?

Yes


sagetherage:
closes: https://github.com/saltstack/salt/issues/54393

PR #54411, opened 2019-09-05
Manage wusa 3010 return code

tlemarchand:
...and add additional comment in state in case of return code different from 0.

### What does this PR do?
This PR fixes a bug.
If a .msu file installation requires reboot, currently wusa state will fail with an un-catched CommandExecutionError, even if installation is done and requires only a reboot.
Also, this PR adds an additional comment in case of CommandExecutionError, so that we can easily know that a reboot is needed, or in case of failure what is the problem.

### Previous Behavior

> ----------
> ID: KB4054590
> Function: wusa.installed
> Result: False
> Comment: An exception occurred in this state: Traceback (most recent call last):
>             File "C:\salt\bin\lib\site-packages\salt\state.py", line 1919, in call
>                 **cdata['kwargs'])
>             File "C:\salt\bin\lib\site-packages\salt\loader.py", line 1918, in wrapper
>                 return f(*args, **kwargs)
>             File "C:\salt\bin\lib\site-packages\salt\states\win_wusa.py", line 90, in installed
>                 __salt__['wusa.install'](cached_source_path)
>             File "C:\salt\bin\lib\site-packages\salt\modules\win_wusa.py", line 143, in install
>                 raise CommandExecutionError('Unknown error: {0}'.format(ret_code))
>             CommandExecutionError: Unknown error: 3010

### New Behavior
wusa

### Tests written?
No

### Commits signed with GPG?

No


salt/states/win_wusa.py
```diff
@@ -87,15 +88,21 @@ def installed(name, source):
         return ret
 
     # Install the KB
-    __salt__['wusa.install'](cached_source_path)
+
+    additional_comment = ''
```
waynew: Maybe make this `'.'` by default, so if there's no additional comment, the original output will be produced?

waynew (changes requested):

salt/states/win_wusa.py
tlemarchand: I'm not sure I understand.
I think that's the case already, since I "append" additional_comment to the "old" comment. Or maybe you have something else in mind ?

waynew: You've removed the trailing period from the message a few lines down. When there's no additional comment instead of something like `The sentence.` you get `The sentence` without a period.

If you set this, then when there is no other additional comment, you'll get the original output (i.e. with the period).

Does that make sense?

tlemarchand: Well I think I understand what you want to say, I agree that there should always be a trailing dot in the comment, but actually I **added** a trailing dot. It's missing in the current version.
With this PR, the trailing dot will always be present, with or without additional comment.

waynew: 😝 doh. You're right. I must have been tired when I was looking at this earlier.

tlemarchand:
Hi,

Is it possible to move forward on this ?

twangboy (approved):

twangboy:
@tlemarchand Could you rebase this against master. And please add some tests.

tlemarchand:
> @tlemarchand Could you rebase this against master. And please add some tests.

@twangboy , I rebased against master, fixed existing tests and added a new test for the 3010 return code.

dwoz (approved):

PR #54454, opened 2019-09-11
Handle bytes when using gitpython with python3.x

vin01:
### What does this PR do?

Use BytesIO fake file object instead of StringIO to support gitpython with Python3.x

https://github.com/saltstack/salt/issues/54402

### Commits signed with GPG?

Yes/No


vin01:
Closing in favor of https://github.com/saltstack/salt/pull/54900

Issue #54459, opened 2019-09-11
modules.iptables optimistically tries to read /etc/sysconfig/iptables


vladionescu:
https://github.com/saltstack/salt/blob/705e8cca0b9ad6441c3884e76798cac1ff7859d7/salt/modules/iptables.py#L987

If modules.iptables.get_saved_rules() is called without an explicit dump file path (generated by iptables -S or iptables-save), then it assumes the default OS' iptables config path and tries to read it, but does not handle the case where that file doesn't exist.

All this happens at the top of modules.iptables._parse_conf().

Here's the unhandled exception when this happens:
```
sys-firewall:
  ----------
            ID: QBS-FORWARD-exists
      Function: iptables.chain_present
          Name: QBS-FORWARD
        Result: True
       Comment: iptables QBS-FORWARD chain is already exist in filter table for ipv4
       Started: 12:39:56.244931
      Duration: 6.656 ms
       Changes:   
  ----------
            ID: set-downstream-mtu
      Function: iptables.insert
        Result: False
       Comment: An exception occurred in this state: Traceback (most recent call last):
                  File /var/tmp/.root_62a99a_salt/pyall/salt/state.py, line 1933, in call
                    **cdata['kwargs'])
                  File /var/tmp/.root_62a99a_salt/pyall/salt/loader.py, line 1939, in wrapper
                    return f(*args, **kwargs)
                  File /var/tmp/.root_62a99a_salt/pyall/salt/states/iptables.py, line 573, in insert
                    saved_rules = __salt__['iptables.get_saved_rules'](family=family)
                  File /var/tmp/.root_62a99a_salt/pyall/salt/modules/iptables.py, line 561, in get_saved_rules
                    return _parse_conf(conf_file=conf_file, family=family)
                  File /var/tmp/.root_62a99a_salt/pyall/salt/modules/iptables.py, line 990, in _parse_conf
                    with salt.utils.files.fopen(conf_file, 'r') as ifile:
                  File /var/tmp/.root_62a99a_salt/pyall/salt/utils/files.py, line 399, in fopen
                    f_handle = open(*args, **kwargs)  # pylint: disable=resource-leakage
                IOError: [Errno 2] No such file or directory: u'/etc/sysconfig/iptables'
       Started: 12:39:56.252038
      Duration: 14.878 ms
       Changes:   
```

And the state that triggers it:
```
QBS-FORWARD-exists:
  iptables.chain_present:
    - name: QBS-FORWARD

[...]

PR #54460, opened 2019-09-11
Fix Issue #54459

vladionescu:
### What does this PR do?
When a minion doesn't have the default file where
iptables stores it's current state, iptables.insert
will throw an exception. This change makes it so if
that file is not found, the current state is read
from memory instead of hard failing.

#54459 
### Previous Behavior
Exception thrown during `iptables.insert` when the default saved configuration file (/etc/sysconfig/iptables on Fedora) didn't exist.

### New Behavior
Iptables state is read from memory (`iptables-save`) when the file is not found.

### Tests written?

No

### Commits signed with GPG?

No

### DIFF
--- a/salt/modules/iptables.py
+++ b/salt/modules/iptables.py
@@ -984,8 +984,12 @@ def _parse_conf(conf_file=None, in_mem=False, family='ipv4'):
 
     rules = ''
     if conf_file:
-        with salt.utils.files.fopen(conf_file, 'r') as ifile:
-            rules = ifile.read()
+        try:
+            with salt.utils.files.fopen(conf_file, 'r') as ifile:
+                rules = ifile.read()
+        except IOError:
+            # Fix GitHub Issue #54459
+            return _parse_conf(conf_file=None, in_mem=True, family=family)
     elif in_mem:
         cmd = '{0}-save' . format(_iptables_cmd(family))
         rules = __salt__['cmd.run'](cmd)


waynew:
Hi @vladionescu thanks for the PR! I like it - looks good so far.

Do you think you would be able to put together a test for this? It should be pretty straightforward, I think.

vladionescu:
Truthfully I don't think I'll be writing a test for this because I don't have a dev Salt deployment I can use. I made this change in my prod environment and verified it worked so wanted to share.

The only place I'm running Salt right now is in QubesOS which has a bunch of stuff built around Salt, and I'm not sure (or have the time to figure out) how I could set up a dev environment within all that. Things are weird. https://www.qubes-os.org/doc/salt/

waynew:
If you have docker available, you can run:

    docker run --rm -it -v "$PWD/salt":/usr/lib/python2.7/dist-packages/salt -v "$PWD/salt":/usr/local/lib64/python3.6/site-packages/salt -v "$PWD":/testing waynew/salt-sprints python3  /testing/tests/runtests.py -n unit.modules.test_iptables

from within your salt directory (you could probably skip out on the python2 part - I just have a shell alias). That should run the existing unit tests. You could update `tests/unit/modules/test_iptables.py` with some appropriate tests. I'd at least write a test that ensures that if the config file is missing that it doesn't error, but instead does what you've configured it to do here.

dwoz:
@vladionescu We are no longer accepting PRs to the `develop` branch. This needs to be re-opened against the `master` branch.

PR #54468, opened 2019-09-13
Lazily gather the lists of available sls data from the master

sathieu:
### What does this PR do?

This improves significantly the performance of a state.apply when using
GitFS and a lot of branches.

none

### Previous Behavior
3 minutes with 276 branches.

### New Behavior
25 seconds

### Tests written?

No. Already covered by existing tests.

### Commits signed with GPG?

No

### DIFF
--- a/salt/state.py
+++ b/salt/state.py
@@ -3275,6 +3275,49 @@ class State(object):
         return self.call_high(high)
 
 
+class LazyAvailStates(object):
+    """
+    The LazyAvailStates lazily loads the list of states of available
+    environments.
+
+    This is particularly usefull when top_file_merging_strategy=same and there
+    are many environments.
+    """
+
+    def __init__(self, hs):
+        self._hs = hs
+        self._avail = {"base": None}
+        self._filled = False
+
+    def _fill(self):
+        if self._filled:
+            return
+        for saltenv in self._hs._get_envs():
+            if saltenv not in self._avail:
+                self._avail[saltenv] = None
+        self._filled = True
+
+    def __contains__(self, saltenv):
+        if saltenv == "base":
+            return True
+        self._fill()
+        return saltenv in self._avail
+
+    def __getitem__(self, saltenv):
+        if saltenv != "base":
+            self._fill()
+        if self._avail[saltenv] is None:
+            self._avail[saltenv] = self._hs.client.list_states(saltenv)
+        return self._avail[saltenv]
+
+    def items(self):
+        self._fill()
+        ret = []
+        for saltenv, states in self._avail:
+            ret.append((saltenv, self.__getitem__(saltenv)))
+        return ret
+
+
 class BaseHighState(object):
     """
     The BaseHighState is an abstract base class that is the foundation of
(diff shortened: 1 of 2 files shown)


sathieu:
Note : I have tested a similar patch for pillars (in `salt/pillar/__init__.py:462` `Pillar.__gather_avail()`), but the performance improvement is not that significant. Probably because the loop in master-only, while the loop in `BaseHighState` is on the minion but talks to the master for each env.

sathieu:
@twangboy Could you please review?

sathieu:
@dwoz I've added tests

salt/state.py
```diff
@@ -362,7 +362,6 @@ class StateError(Exception):
     '''
     Custom exception class.
     '''
-    pass
```
twangboy: Isn't this supposed to remain?

twangboy (changes requested):

salt/state.py
sathieu: As written in 332de26df886a442ccbfa43ccce1d73dd1b3bf96, this is to fix:
```
salt/state.py:370: [W0107(unnecessary-pass), StateError] Unnecessary pass statement
```

Ref:
- http://pylint.pycqa.org/en/latest/technical_reference/features.html
- https://help.semmle.com/wiki/display/PYTHON/Unnecessary+pass

DmitryKuzmenko (dismissed):

waynew:
@thatch45 can you take a look at this? It *looks* pretty straightforward to me but I'm not 100% familiar with the implications of futzing with the states.

thatch45 (dismissed):

thatch45:
I really like this approach, it is clean, and it can be pulled out if there are issues. But this does look like it should work.
I love simple, clean additions :)

DmitryKuzmenko:
@twangboy your turn!

DmitryKuzmenko:
@sathieu could you please resolve merge conflicts?

DmitryKuzmenko:
@sathieu could you please resolve the merge conflicts and run the pre-commit script on your code to re-format it with `isort` and `black` we're using now.
```
pip install pre-commit
pre-commit install
FILES="$(git show --pretty="" --name-only master..HEAD)"
pre-commit run isort $FILES
pre-commit run black $FILES
```
next time you'll commit it will be ran automatically.

sathieu:
@DmitryKuzmenko I've rebased and used pre-commit hook. I had to use `virtualenv -p /usr/bin/python3 DESTDIR`, otherwise python2 was used.

Then, it failed with:
```
  Attempting uninstall: wrapt
    Found existing installation: wrapt 1.10.11
ERROR: Cannot uninstall 'wrapt'. It is a distutils installed project and thus we cannot accurately determine which files belong to it which would lead to only a partial uninstall.
Session lint-tests-pre-commit failed.
```
I've commited again (black and isort already runt) with `--no-verify` and pushed here.

DmitryKuzmenko:
@sathieu thank you for report. We'll work on it.
Meanwhile I'll check your code for you. Thank you for contribution.

waynew:
@sathieu did you use `pip` or `python3 -m pip install --user pre-commit`?  It *should* be [defaulting to python3](https://github.com/saltstack/salt/pull/56586/files#diff-82c3b36709926de3a1b27bea0d8258ccR2) when installing the hooks.

If just installing pre-commit that way doesn't work running `rm -rf ~/.cache/pre-commit/` may help.

DmitryKuzmenko:
@waynew ah, it looks it's my fault. I should recommend `python3 pip` next time.

dwoz (approved):

Issue #54505, opened 2019-09-17
[Feature request] Add a Elasticsearch cluster settings methods


Oloremo:
### Description of Issue
Please add a `get_settings` and `put_settings` methods from the Python elasticsearch library: https://elasticsearch-py.readthedocs.io/en/master/api.html

Also, the `flush_synced` API call is good to have.

Having these methods are critical for the implementation of a rolling update logic: https://www.elastic.co/guide/en/elasticsearch/reference/current/rolling-upgrades.html

PR #54506, opened 2019-09-17
Added Elasticsearch cluster settings get and put methods plus the flush_synced method

Oloremo:
### What does this PR do?
Adding `get_settings` and `put_settings` cluster method for Elasticsearch module plus the `flush_synced` method.

#54505

### Previous Behavior
Elasticsearch module doesn't support these methods

### New Behavior
Elasticsearch module does support these methods

### Tests written?

Yes

### Commits signed with GPG?

Yes

### DIFF
--- a/tests/unit/modules/test_elasticsearch.py
+++ b/tests/unit/modules/test_elasticsearch.py
@@ -2339,3 +2339,37 @@ class ElasticsearchTestCase(TestCase):
         '''
 
         self.assertRaises(SaltInvocationError, elasticsearch.cluster_put_settings)
+
+    # flush_synced tests below.
+    # We're assuming that _get_instance is properly tested
+    # These tests are very simple in nature, mostly checking default arguments.
+    def test_flush_synced_succeess(self):
+        '''
+        Test if flush_synced succeeds
+        '''
+
+        expected_return = {'_shards': {'failed': 0, 'successful': 0, 'total': 0}}
+        fake_es = MagicMock()
+        fake_es.indices = MagicMock()
+        fake_es.indices.flush_synced = MagicMock(return_value=expected_return)
+        fake_instance = MagicMock(return_value=fake_es)
+
+        with patch.object(elasticsearch, '_get_instance', fake_instance):
+            output = elasticsearch.flush_synced(index='_all', ignore_unavailable=True, allow_no_indices=True, expand_wildcards='all')
+            fake_es.indices.flush_synced.assert_called_with({'index': '_all', 'ignore_unavailable': True, 'allow_no_indices': True, 'expand_wildcards': 'all'})
+            assert output == expected_return
+
+    def test_flush_synced_failure(self):
+        '''
+        Test if flush_synced fails with CommandExecutionError
+        '''
+
+        fake_es = MagicMock()
+        fake_es.indices = MagicMock()
+        fake_es.indices.flush_synced = MagicMock()
+        fake_es.indices.flush_synced.side_effect = TransportError("custom error", 123)
+        fake_instance = MagicMock(return_value=fake_es)
+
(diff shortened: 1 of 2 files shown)


tests/unit/modules/test_elasticsearch.py
```diff
@@ -2264,3 +2264,115 @@ def delete_template(self, id=None):
...
+
+        class MockElastic(object):
+            '''
+            Mock of Elasticsearch client
+            '''
+            cluster = MockElasticCluster()
+
+        with patch.object(elasticsearch, '_get_instance',
+                          MagicMock(return_value=MockElastic())):
+            self.assertDictEqual(elasticsearch.cluster_get_settings(), {"transient": {}, "persistent": {}})
```
waynew: I'm not sure if you understood what I was talking about on Slack, but because there really isn't *behavior* to test here it's probably reasonable to just make sure that interfaces aren't changing without warning. You actually don't need to write your own mock classes - you can just use a (series) of MagicMocks for that

```
expected_settings = {"transient": {}, "persistent": {}}
fake_instance = MagicMock()
fake_instance.cluster = MagicMock()
fake_instance.cluster.get_settings = MagicMock(return_value=expected_settings)
with patch.object(elasticsearch, '_get_instance', fake_instance):
     actual_settings = elasticsearch.cluster_get_settings()
     fake_instance.cluster.get_settings.assert_called_with(flat_settings=None, include_defaults=None)
```

I'm not a *huge* fan of deep mocking like this - ideally you shouldn't be mocking anything - but I'm not aware of any better patterns to follow for cases like this.

If I had the [infinity gauntlet](https://en.wikipedia.org/wiki/The_Infinity_Gauntlet), I would write a functional test here. That is, I would have an Elasticsearch cluster for real that I could test with, that I would spin up at the beginning of the Elasticsearch tests and down when I am finished testing. Then I would test that in fact, settings were set and could also be retrieved.

That might be prohibitive, though, so the pragmatist in me says that just making sure that we're passing through the values that we expect to pass in, and we're returning the values that we expect to return is good enough, for now.

I'm open to other suggestions, though!

Oloremo: Well, I'm tried to make changes which are aligned with the rest of the changes already present ie "copy-paste" :-D I do hope they make at least some sense.

waynew: Yeah, it does make some sense. Do you understand what I meant Re: mocks, though? Does that feel like a better approach than the one that we've got here?

I'd also recommend adding some comment to the effect of, "Yes, we're aware that these tests aren't *super* useful - more of just a smoke test to make sure we don't accidentally break things". Seem reasonable?

Oloremo: Well yes, I'm trying to understand your example which is not working for me(yet). I mean i get the idea now I'm trying to understand what is wrong with the syntax(I don't have much experience with this whole mock thingy)

Oloremo: tests were updated based on your proposal

Oloremo:
Also added the flush_synced method which is required for the original issue.

Oloremo:
Tests failing with
```
13:11:19  rbenv: version `2.4.2' is not installed (set by RBENV_VERSION environment variable)
```
`/shrug`

Oloremo:
@waynew after a long fight I've managed to pass all tests :-)

waynew (approved):

waynew:
@Oloremo thanks for sticking with it! Looks good and we'll get this merged in as soon as we can 👍

Oloremo:
PR to master: https://github.com/saltstack/salt/pull/54917

dwoz:
@Oloremo Closing this since there has been a port to master already.

PR #54531, opened 2019-09-18
[master] Porting #50555 to master

garethgreenaway:
Porting #50555 to master


garethgreenaway:
re-run full all

dwoz (changes requested):
@rongzeng54 We are in the process of migrating this PR to the new `master` branch and believe these changes have the potential to become out of date because of the mapping [here](https://github.com/saltstack/salt/pull/54531/files#diff-afe7362c6e0817c26aaf1e3c4351ba97R755).

I think the best approach is to just detect if it is a state running and point the user to the `state.soft_kill` documentation.

Are you able to make these changes?

dwoz:

PR #54533, opened 2019-09-18
[master] Porting #50696 to master

garethgreenaway:
Porting #50696 to master


garethgreenaway:
re-run full all

salt/modules/saltutil.py
```diff
@@ -1444,6 +1445,14 @@ def _exec(client, tgt, fun, arg, timeout, tgt_type, ret, kwarg, **kwargs):
...
         }
         del kwargs['subset']
+    elif kwargs.get('asynchronous'):
+        cmd_kwargs = {
+            'tgt': tgt, 'fun': fun, 'arg': arg, 'tgt_type': tgt_type,
+            'ret': ret, 'kwarg': kwarg
+        }
+        del kwargs['asynchronous']
+        # run_job doesnt need processing like the others
+        return client.run_job(**cmd_kwargs)
```
dhiltonp: Passed kwargs aren't ever applied to cmd_kwargs, unlike the other cases. Keeping consistent behavior would be good.

A thing I'm not clear on - why does this chunk bypass calling `_cmd` like the other ways the function is called?

salt/modules/saltutil.py
```diff
@@ -1444,6 +1445,14 @@ def _exec(client, tgt, fun, arg, timeout, tgt_type, ret, kwarg, **kwargs):
             'ret': ret, 'cli': True, 'kwarg': kwarg, 'sub': kwargs['subset'],
         }
         del kwargs['subset']
+    elif kwargs.get('asynchronous'):
```
dhiltonp: We should change the function signature to:

`def _exec(client, tgt, fun, arg, timeout, tgt_type, ret, kwarg, batch=False, subset=False, asynchronous=False, **kwargs):`

That way we don't need to remove the flags from kwargs manually - any excess options will be passed to the command.

salt/modules/saltutil.py
```diff
@@ -1444,6 +1445,14 @@ def _exec(client, tgt, fun, arg, timeout, tgt_type, ret, kwarg, **kwargs):
...
+        cmd_kwargs = {
```
dhiltonp: Maybe we should define `cmd_kwargs` above the if statements, then just modify it with the differences in the respective if clauses.

On a related note, are the clauses `batch`, `subset` and `asynchronous` mutually exclusive? @waynew what do you think?

salt/modules/saltutil.py
```diff
@@ -1608,11 +1622,16 @@ def runner(name, arg=None, kwarg=None, full_return=False, saltenv='base', jid=No
...
 
-    return rclient.cmd(name,
-                       arg=arg,
-                       kwarg=kwarg,
-                       print_event=False,
-                       full_return=full_return)
+    if asynchronous:
+        master_key = salt.utils.master.get_master_key('root', __opts__)
+        low = {'arg': arg, 'kwarg': kwarg, 'fun': name, 'key': master_key}
+        return rclient.cmd_async(low)
```
dhiltonp: `rclient.cmd` and `rclient.cmd_async` seem to have inconsistent interfaces - a quick check and `LocalClient.cmd_async` accepts full parameters, but other `cmd_async`s accept a dictionary...

salt/modules/saltutil.py
dhiltonp: Doing that would also require removing a few lines from `saltmod.py:function` which determines if batch, subset and asynchronous are in kwargs.

DmitryKuzmenko:
@garethgreenaway do you plan to follow @dhiltonp recommendations in scope of this PR?

dhiltonp:
I've cleaned up the `_exec` implementation in #56203, finding a bug along the way.

It doesn't include this PR, though.

salt/modules/saltutil.py
dhiltonp: What's the deal with `low`? Is it normal elsewhere or something?

dhiltonp:
I've merged the fixes I suggested with this code into master: #56480

DmitryKuzmenko:
Thank you @dhiltonp

dhiltonp:
I just realized I miscommunicated. That PR hasn't yet been merged, but it is based on current master and combines the two PRs.

DmitryKuzmenko:
@dhiltonp this is how I got you. I've checked your PR and found it containing everything from this one. Am I wrong in closing this?

PR #54535, opened 2019-09-18
[master] Porting #52502 to master

garethgreenaway:
Porting #52502 to master


garethgreenaway:
re-run full all

dwoz (approved):

salt/states/archive.py
```diff
@@ -1350,10 +1350,13 @@ def extracted(name,
                         )
                         return ret
 
-                    tar_opts = shlex.split(options)
+                    # Ignore verbose file list options as we are already using
+                    # "v" below in tar_shortopts
+                    tar_opts = [x for x in shlex.split(options)
+                                if x not in ('v', '-v', '--verbose')]
```
dwoz: What happens if someone passes `xv`? Should this check be something like:

```
if not x.startswith('-') and 'v' not in x
```

in addition to what it is checking now?

salt/states/archive.py
```diff
@@ -1383,9 +1386,9 @@ def extracted(name,
                         ret['changes'] = results
                         return ret
                     if _is_bsdtar():
-                        files = results['stderr']
+                        files = results['stderr'].splitlines()
                     else:
-                        files = results['stdout']
+                        files = results['stdout'].splitlines()
```
dwoz: What happens if someone extracts an archive with 100K + files?

dwoz (changes requested):

dwoz:

PR #54547, opened 2019-09-18
[master] Porting #49482 to master

garethgreenaway:
Porting #49482 to master

### DIFF
--- a/salt/utils/network.py
+++ b/salt/utils/network.py
@@ -1904,7 +1904,7 @@ def connection_check(addr, port=80, safe=False, ipv6=None):
 
 
 @jinja_filter('dns_check')
-def dns_check(addr, port=80, safe=False, ipv6=None):
+def dns_check(addr, port=None, safe=False, ipv6=None):
     '''
     Return an ip address resolved by dns in a format usable in URLs (ipv6 in brackets).
     Obeys system preference for IPv4/6 address resolution - this can be overridden by
--- a/salt/utils/network.py
+++ b/salt/utils/network.py
@@ -1895,8 +1895,16 @@ def refresh_dns():
         pass
 
 
+@jinja_filter('connection_check')
+def connection_check(addr, port=80, safe=False, ipv6=None):
+    '''
+    Provides a convenient alias for the dns_check filter.
+    '''
+    return dns_check(addr, port, safe, ipv6)
+
+
 @jinja_filter('dns_check')
-def dns_check(addr, port, safe=False, ipv6=None):
+def dns_check(addr, port=80, safe=False, ipv6=None):
     '''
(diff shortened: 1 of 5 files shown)


salt/utils/network.py
```diff
@@ -1896,8 +1896,16 @@ def refresh_dns():
         pass
 
 
+@jinja_filter('connection_check')
+def connection_check(addr, port=80, safe=False, ipv6=None):
```
dwoz: I don't see the need for `connection_check` when `dns_check` does the *exact* same thing.

salt/utils/network.py
```diff
@@ -1896,8 +1896,16 @@ def refresh_dns():
...
+    '''
+    Provides a convenient alias for the dns_check filter.
+    '''
+    return dns_check(addr, port, safe, ipv6)
+
+
 @jinja_filter('dns_check')
-def dns_check(addr, port, safe=False, ipv6=None):
+def dns_check(addr, port=80, safe=False, ipv6=None):
```
dwoz: Should this default to `None` instead of 80?

Also, this change should be tested.

dwoz (changes requested):

salt/utils/network.py
garethgreenaway: Added a test case to the dictionary that does not include the port.  Updated the call to dns_check to not pass port if it's not in the dictionary.  Updated the default to be None.

salt/utils/network.py
garethgreenaway: @MTecknology Looks like you originally added the `connection_check` alias, would you care to comment on it's need versus `dns_check`?

MTecknology: @garethgreenaway I don't recall adding it and I'm having a hard time finding a commit where I added it. Is there any chance you could point me toward it so maybe my memory can be triggered?

twangboy: https://github.com/saltstack/salt/pull/49482/commits/026a5763e31731cfb66509f3ac7eea20f87d078a

twangboy: @MTecknology ^^^

MTecknology: Oh, hey friends! How have you fellas been?

Yes, now that I'm digging back into things, I do recall this; ``connection_check()`` was a "convenience alias" in [#46053](https://github.com/saltstack/salt/issues/46053#issuecomment-366277516).

If we prefer not have aliases, then I see no reason to keep it.

MTecknology: Removing this function would effectively finish reverting commits 2e0ae24 and 026a576. I believe this is a good idea.

MTecknology (approved):

dwoz:

MTecknology:
So, was this abandoned due to staff turnover, or what happened?

twangboy:
Pretty much

PR #54551, opened 2019-09-18
[master] Porting #50150 to master

garethgreenaway:
Porting #50150 to master


salt/states/http.py
```diff
@@ -94,28 +107,40 @@ def query(name, match=None, match_type='string', status=None, wait_for=None, **k
...
     if status is not None:
-        if data.get('status', '') == status:
-            ret['comment'] += 'Status {0} was found, as specified.'.format(status)
-            if ret['result'] is None:
-                ret['result'] = True
-        else:
-            ret['comment'] += 'Status {0} was not found, as specified.'.format(status)
-            ret['result'] = False
+        if status_type == 'string':
+            if data.get('status', '') == str(status):
```
dwoz: Should this be `str(data.get('status', '')) == str(status)`? Assuming so, we should have a test for this case.

dwoz (changes requested):

max-arnold:
I believe there is no need to backport this PR, because it was included into https://github.com/saltstack/salt/pull/55145

dwoz:

PR #53080, opened 2019-05-16
Pkg installed flag fix

cmcmarrow:
### What does this PR do?
listen to extra flags when running pkg.installed state

### What issues does this PR fix or reference?
https://github.com/saltstack/salt/issues/50041

### Tests written?

yes

### Commits signed with GPG?

Yes

PR #54566, opened 2019-09-18
[master] Porting #53080 to master

garethgreenaway:
Porting #53080 to master


salt/modules/win_pkg.py
```diff
@@ -1447,6 +1447,11 @@ def install(name=None, refresh=False, pkgs=None, **kwargs):
                 'extra_install_flags': kwargs.get('extra_install_flags')
             }
         }
+    elif len(pkg_params) == 1:
+        # A dict of packages was passed, but it contains only 1 key, so we need
+        # to add the 'extra_install_flags'
+        for pkg in pkg_params:
```
cmcmarrow: remove for loop

cmcmarrow (changes requested):

salt/modules/win_pkg.py
waynew: would it be better as `pkg = next(iter(pkg_params))`? Not sure if there's another, better approach.

dafyddj:
I hope this can make it into the next release, as it is an annoying bug.

twangboy (approved):

salt/modules/win_pkg.py
cmcmarrow: I like that

cmcmarrow (approved):

s0undt3ch (approved):

PR #52838, opened 2019-05-03
Json to xml psget fix

cmcmarrow:
### What does this PR do?
win_psget.py drops Json to use Xml. 
This is because Power-Shell Json does not write to file/buffer correctly.

### What issues does this PR fix or reference?
https://github.com/saltstack/salt/issues/52216

### Tests written?
Yes 
"test_win_psget.py"

### Commits signed with GPG?
Yes

PR #54568, opened 2019-09-18
[master] Porting #52838 to master

garethgreenaway:
Porting #52838 to master


salt/modules/win_psget.py
```diff
@@ -50,14 +49,42 @@ def __virtual__():
     return __virtualname__
 
 
-def _pshell(cmd, cwd=None, json_depth=2):
+def _ps_xml_to_dict(parent, dic=None):
+    '''
+    Formats powershell Xml to a dict.
+    Note: This _ps_xml_to_dict is not perfect with powershell Xml.
```
dwoz: @cmcmarrow 
What does this mean `is not perfect with powershell Xml`?

Ch3LL: bump @cmcmarrow

Ch3LL:
Closing since this was included in the 2018.3 merge forward here https://github.com/saltstack/salt/pull/55027

salt/modules/win_psget.py
cmcmarrow: XML and Dictionaries are not one to one. But in this uses cases it is. We should be using Json here but at the time there was a upstream bug in with the PowerShell program.

cmcmarrow: @dwoz

PR #54570, opened 2019-09-18
[master] Porting #52044 to master

garethgreenaway:
Porting #52044 to master

### DIFF
--- a/salt/modules/baredoc.py
+++ b/salt/modules/baredoc.py
@@ -1,0 +1,183 @@
+# -*- coding: utf-8 -*-
+'''
+Baredoc walks the installed module and state directories and generates
+dictionaries and lists of the function names and their arguments.
+
+.. versionadded:: Neon
+
+'''
+from __future__ import absolute_import, unicode_literals, print_function
+
+# Import python libs
+import logging
+import os
+import re
+
+# Import salt libs
+import salt.loader
+import salt.runner
+import salt.state
+import salt.utils.data
+import salt.utils.files
+import salt.utils.args
+import salt.utils.schema
+
+# Import 3rd-party libs
+from salt.ext import six
+from salt.utils.odict import OrderedDict
+
+log = logging.getLogger(__name__)
+
+
+def _parse_function_definition(fn_def, modulename, ret):
+    args = []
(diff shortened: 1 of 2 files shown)


salt/modules/baredoc.py
```diff
@@ -0,0 +1,183 @@
...
+# Import 3rd-party libs
+from salt.ext import six
+from salt.utils.odict import OrderedDict
+
+log = logging.getLogger(__name__)
+
+
+def _parse_function_definition(fn_def, modulename, ret):
+    args = []
+    match = re.match(r'def\s+(.*?)\((.*)\):$', fn_def)
```
dwoz: Does this handle function definitions that span multiple lines?
```
def foo(
    bar,
    bang):
    pass
``` 

Does this handle function aliases in modules?

s0undt3ch (commented):
Why exactly do we need this kind of docs parsing?

mchugh19:
> Why exactly do we need this kind of docs parsing?

Actually, I have a use-case for listing state functions. It might not be common, but in order to generate autocomplete listings for vscode (korekontrol/vscode-saltstack#6), I need to know what possible state functions might exist. So far I've tried installing a bunch of dependencies and run on linux and windows to use `sys.list_state_functions`. This is pretty good, but I'm probably missing out on lots of states which won't load due to dependencies. 

I'm not sure if this necessitates maintaining baredoc forever, although if it were merged it should really be updated to use ast rather than regex. 

Iterating possible state and module arguments without loading all modules might also be helpful for [saltcheck](https://github.com/saltstack/salt/blob/master/salt/modules/saltcheck.py#L530) so that you don't need to load up all modules and blow any lazyloading cache. Further thought may be needed though, as you may want to alert if a dependency can't be run.

@cro

mchugh19:
I've tried making a version of baredoc which uses ast parsing rather than regex: #56685. If it's decided to be not useful enough to merge, I understand. But this idea does seem to crop up ever so often.

mchugh19:
With the merge of #56685 this can be closed.

garethgreenaway:
@mchugh19 awesome. Thanks!

PR #54574, opened 2019-09-19
[master] Porting #52749 to master

garethgreenaway:
Porting #52749 to master


tests/unit/test_version.py
```diff
@@ -71,3 +71,15 @@ def test_unparsable_version(self):
 
         with self.assertRaises(ValueError):
             SaltStackVersion.parse('Drunk')
+
+    def test_version_report_lines(self):
```
dhiltonp: Good PR all in all.

One concern I have is the execution time of versions_report - basically, this is an integration test, not a unit test.

Moving this test to integration, then having a unit test version that runs with known static data could be good.

s0undt3ch (approved):

PR #54576, opened 2019-09-19
[master] Porting #52933 to master

garethgreenaway:
Porting #52933 to master


dwoz (approved):

salt/beacons/inotify.py
```diff
@@ -117,6 +117,10 @@ def validate(config):
         if 'files' not in _config:
             return False, 'Configuration for inotify beacon must include files.'
         else:
+            if not isinstance(_config['files'], dict):
```
s0undt3ch: I would remove the above `else` and dedent this logic.

tests/unit/beacons/test_inotify.py
```diff
@@ -39,10 +39,33 @@ def setUp(self):
     def tearDown(self):
         shutil.rmtree(self.tmpdir, ignore_errors=True)
 
+    def test_non_list_config(self):
+        config = {}
```
s0undt3ch: Unnecessary variable, just pass `{}` to `inotify.validate()`.

tests/unit/beacons/test_inotify.py
```diff
@@ -39,10 +39,33 @@ def setUp(self):
...
+
+        ret = inotify.validate(config)
+
+        self.assertEqual(ret, (False, 'Configuration for inotify beacon must'
+                                      ' be a list.'))
+
     def test_empty_config(self):
         config = [{}]
```
s0undt3ch: Same as above.

tests/unit/beacons/test_inotify.py
```diff
@@ -39,10 +39,33 @@ def setUp(self):
...
-        ret = inotify.beacon(config)
-        self.assertEqual(ret, [])
+        ret = inotify.validate(config)
+        _expected = (False, 'Configuration for inotify beacon must include files.')
+        self.assertEqual(ret, _expected)
+
+    def test_files_none_config(self):
+        config = [{'files': None}]
```
s0undt3ch: I won't say same as above here, although we don't actually use `config` anywhere else.
But for the sake of readability, let's leave it.
The same for the next test.

s0undt3ch (approved):
None of my review comments are blockers. It would just make the code better, more readable.

salt/beacons/inotify.py
```diff
@@ -117,6 +117,10 @@ def validate(config):
...
+                return False, ('Configuration for inotify beacon invalid, '
+                               'files must be a dict.')
+
             for path in _config.get('files'):
 
                 if not isinstance(_config['files'][path], dict):
```
dhiltonp: `Configuration for inotify beacon must be a list of dictionaries.`

This error is not correct, it must be a dict of dicts (which this PR validates).

dhiltonp (approved):

PR #54577, opened 2019-09-19
[master] Porting #52715 to master

garethgreenaway:
Porting #52715 to master

### DIFF
--- a/tests/unit/states/test_slack.py
+++ b/tests/unit/states/test_slack.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 '''
     :codeauthor: Jayesh Kariya <jayeshk@saltstack.com>
+    :codeauthor: Gareth J. Greenaway <gareth@saltstack.com>
 '''
 # Import Python libs
 from __future__ import absolute_import, unicode_literals, print_function
@@ -42,8 +43,7 @@ class SlackTestCase(TestCase, LoaderModuleMockMixin):
                'comment': ''}
 
         with patch.dict(slack.__opts__, {'test': True}):
-            comt = ('The following message is to be sent to Slack: {0}'
-                    .format(message))
+            comt = 'The following message is to be sent to Slack: {0}'.format(message)
             ret.update({'comment': comt})
             self.assertDictEqual(slack.post_message(name,
                                                     channel=channel,
@@ -52,7 +52,7 @@ class SlackTestCase(TestCase, LoaderModuleMockMixin):
                                                     api_key=api_key), ret)
 
         with patch.dict(slack.__opts__, {'test': False}):
-            comt = ('Please specify api_key or webhook.')
+            comt = 'Please specify api_key or webhook.'
             ret.update({'comment': comt, 'result': False})
             self.assertDictEqual(slack.post_message(name,
                                                     channel=None,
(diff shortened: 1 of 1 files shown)


dwoz (approved):

salt/states/slack.py
```diff
@@ -39,12 +39,7 @@ def __virtual__():
     return 'slack' if 'slack.post_message' in __salt__ else False
 
 
-def post_message(name,
-                 channel,
-                 from_name,
-                 message,
-                 api_key=None,
-                 icon=None):
+def post_message(name, **kwargs):
```
s0undt3ch: Changing the signature to `**kwargs` is an immediate red flag.
Let's not.
Or better, why do we need it?
To support old and new signature?
Add old and new and deprecate old, etc...

tests/unit/states/test_slack.py
```diff
@@ -47,29 +48,115 @@ def test_post_message(self):
...
-                                                    message), ret)
+            self.assertDictEqual(slack.post_message(name,
+                                                    channel=channel,
+                                                    from_name=from_name,
+                                                    message=message,
+                                                    api_key=api_key), ret)
 
         with patch.dict(slack.__opts__, {'test': False}):
-            comt = ('Slack channel is missing: None')
+            comt = ('Please specify api_key or webhook.')
```
s0undt3ch: Please remove the `()` around the string. It's not needed.

tests/unit/states/test_slack.py
```diff
@@ -47,29 +48,115 @@ def test_post_message(self):
...
-            self.assertDictEqual(slack.post_message(name, None, from_name,
-                                                    message), ret)
+            self.assertDictEqual(slack.post_message(name,
+                                                    channel=None,
+                                                    from_name=from_name,
+                                                    message=message,
+                                                    api_key=None), ret)
 
-            comt = ('Slack from name is missing: None')
+            comt = ('Slack channel is missing.')
```
s0undt3ch: Same as above.

tests/unit/states/test_slack.py
```diff
@@ -47,29 +48,115 @@ def test_post_message(self):
...
-            self.assertDictEqual(slack.post_message(name, channel, None,
-                                                    message), ret)
+            self.assertDictEqual(slack.post_message(name,
+                                                    channel=None,
+                                                    from_name=from_name,
+                                                    message=message,
+                                                    api_key=api_key), ret)
 
-            comt = ('Slack message is missing: None')
+            comt = ('Slack from name is missing.')
```
s0undt3ch: Instead of same as above, any string which fits in a single line and yet has `()` around it. Let's remove the `()`.

s0undt3ch (changes requested):

salt/states/slack.py
dhiltonp: This is weird. The modified function is doing some validation then dispatching calls to slack_notify.py's `call_hook` and `post_message` (both of which do their own validation, by the way).

This function is called `post_message`, so why would it ever call `call_hook`?

I can see adding a generic function called "send_message" if it made sense, but it doesn't:

The only arguments in common are `message` and `channel`.

So long as the caller needs to keep track of the method of dispatch and the several unique arguments, it is better to have 2 functions that are named differently.

-----------

This could be 2 functions, `post_message` and `call_hook` which just pass through to the the `slack_notify` module. 

Why would someone use this instead of `module.run`? Is that less common?

dhiltonp (changes requested):
Can be simplified quite a bit, I think.

Ch3LL:
ping @garethgreenaway looks like you were the original committer of this PR. Can you take a look at the review comments?

garethgreenaway:
Looks like most of the changes for this are already on master, most likely from a merge forward from the 2018.3.x. branch.  This will update the test with @s0undt3ch's requested changes.

PR #49166, opened 2018-08-16
[fluorine] salt enterprise scheduler tests

pizzapanther:
### What does this PR do?

Added tests to simulate how Salt Enterprise uses the scheduler to prevent breakage in the future.

### What issues does this PR fix or reference?

PR #49104

### Tests written?

Yes

### Commits signed with GPG?

Yes

Please review [Salt's Contributing Guide](https://docs.saltstack.com/en/latest/topics/development/contributing.html) for best practices.

See GitHub's [page on GPG signing](https://help.github.com/articles/signing-commits-using-gpg/) for more information about signing commits with GPG.

PR #54578, opened 2019-09-19
[master] Porting #52666 to master

garethgreenaway:
Porting #52666 to master


tests/integration/scheduler/test_eval.py
```diff
@@ -44,21 +43,25 @@
 DEFAULT_CONFIG['pki_dir'] = os.path.join(ROOT_DIR, 'pki')
```
s0undt3ch: These kind of globals should be avoided.
`setUp` or `setUpClass` is the place to define them.

s0undt3ch (changes requested):

salt/utils/schedule.py
```diff
@@ -463,18 +464,22 @@ def run_job(self, name):
...
+        now = datetime.datetime.now()
         if not self.standalone:
             data = self._check_max_running(func,
                                            data,
                                            self.opts,
-                                           datetime.datetime.now())
+                                           now)
 
         # Grab run, assume True
         run = data.get('run', True)
```
dhiltonp: now duplicate?

salt/utils/schedule.py
```diff
@@ -615,6 +620,34 @@ def job_status(self, name):
...
+        if 'jid_include' not in data or data['jid_include']:
+            if self.opts['__role'] == 'master':
+                current_jobs = salt.utils.master.get_running_jobs(self.opts)
+            else:
+                current_jobs = salt.utils.minion.running(self.opts)
+            for job in current_jobs:
+                if 'schedule' in job:
+                    log.info('=== data %s ===', data)
+                    log.info('=== job %s ===', job)
+                    if data['name'] in job['schedule'] \
```
dhiltonp: is this enough to be unique?

salt/utils/schedule.py
```diff
@@ -615,6 +620,34 @@ def job_status(self, name):
         schedule = self._get_schedule()
         return schedule.get(name, {})
 
+    def job_running(self, data):
```
dhiltonp: What is this actually used for?

salt/utils/schedule.py
```diff
@@ -213,7 +213,8 @@ def _check_max_running(self, func, data, opts, now):
         # NOTE--jid_include defaults to True, thus if it is missing from the data
```
dhiltonp: this sounds terrible.

salt/utils/schedule.py
dhiltonp: This seems to be a copy of `_check_max_running` - why don't we call that and then return True/False based on the result?

dhiltonp (changes requested):

waynew:
@garethgreenaway looks like you were the OP on this, so I'll let you address the updated comments 🙃

salt/utils/schedule.py
garethgreenaway: `now` is used in multiple places.  It's used in the call to `_check_max_running` and then it's used again when updating `_last_run` in data.  It should be the same value in both places so using a variable instead of a call to `datetime.datetime.now()`.

salt/utils/schedule.py
garethgreenaway: They are not identical, the code is similar but `job_running` takes different arguments than `_check_maxrunning` and they have different return values.  We're also setting various things in `_check_maxrunning` in data that is used later, that is not being set in `job_running`.  That being said I can't find where `job_running` is being used and I suspect it might be something that the Enterprise team might have needed, @pizzapanther can you comment if you're using `job_running` for anything?

salt/utils/schedule.py
dhiltonp: The comment is about something I noticed in the area, not specific to your commit (my comments were a brain dump, intended to be discussed in the daily meeting).

The comment was about `run = data.get('run', True)` - it was no longer necessary to do that because line 470 does the same thing:

```
if 'run' not in data:
            data['run'] = True
```

The merge from master changed this line to `if data.get('run'):`, which is again unnecessary as 'run' is guaranteed to be set on line 470.

salt/utils/schedule.py
pizzapanther: we do not use `job_running` because we can't exclusively rely on one master to check for running schedules, so we have other measures.

I would suggest resurrecting: https://github.com/saltstack/salt/pull/49166

Then you would have tests on how we use the scheduler which would catch any API breakage we our relying on.

waynew: @dhiltonp Do you have any further comments here?

waynew:
@garethgreenaway looks like there's a merge conflict here

salt/utils/schedule.py
dhiltonp: I haven't found any references to 'job_running' in the enterprise code, maybe we should pull this out.

dhiltonp: There aren't tests for this, either.

dwoz:

PR #54595, opened 2019-09-19
[master] Porting #51344 to master

garethgreenaway:
Porting #51344 to master

### DIFF
--- a/salt/states/network.py
+++ b/salt/states/network.py
@@ -164,7 +164,7 @@ all interfaces are ignored unless specified.
         - max_bonds: 1
         - updelay: 0
         - use_carrier: on
-        - xmit_hash_policy: layer2
+        - hashing-algorithm: layer2
         - mtu: 9000
         - autoneg: on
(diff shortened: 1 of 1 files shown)


doc/man/salt.7
```diff
@@ -383685,7 +383685,7 @@ bond0:
     \- max_bonds: 1
     \- updelay: 0
     \- use_carrier: on
-    \- xmit_hash_policy: layer2
+    \- hashing-algorithm: layer2
```
Ch3LL: this file is generated when we build the man pages for each release, so we should never directly edit this file

Ch3LL (changes requested):

doc/man/salt.7
Ch3LL: ping @waynew looks like you made this change originally, mind taking a look here? Or i can make the change

waynew:
@dwoz doc-only change here

PR #54617, opened 2019-09-19
[master] Porting #50072 to master

garethgreenaway:
Porting #50072 to master


DmitryKuzmenko (approved):

DmitryKuzmenko:
The test will fail until #54618 be merged.

dwoz:
This test failed on multiple builds during the last runs:

https://jenkinsci.saltstack.com/job/pr-amazon2-py2/job/PR-54617/5/testReport/junit/unit.states.test_module/ModuleStateTest/test_call_function_named_args/

It's flaky and we'll need to fix it before merging this.

DmitryKuzmenko:
@dwoz this is a test for #50049 that would fail without that fix. The fix is ported in #54618 but the PR isn't merged yet.
Both the logic and the test for it are isolated and have no side effects as far as I can see so it shouldn't be flaky at all.

PR #50668, opened 2018-11-28
Adds 'keyspace' option for returner to make returner available to specify Cassandra keyspace.

kstreee:
### What does this PR do?

Adds 'keyspace' option for returner to make returner available to specify Cassandra keyspace.

### What issues does this PR fix or reference?

N/A

### Previous Behavior

Cassandra returner cannot be specified the keyspace.

### New Behavior

Cassandra returner can be specified which keyspace to use.

### Tests written?

Yes

### Commits signed with GPG?

Yes

Please review [Salt's Contributing Guide](https://docs.saltstack.com/en/latest/topics/development/contributing.html) for best practices.

See GitHub's [page on GPG signing](https://help.github.com/articles/signing-commits-using-gpg/) for more information about signing commits with GPG.

PR #54635, opened 2019-09-19
[master] Porting #50668 to master

garethgreenaway:
Porting #50668 to master


waynew (changes requested):
I'm not sure if one can parametrize the keyspace here, or if that's not possible in Cassandra.

I'd like to see either:

a. Parametrize if possible, or 
b. A warning comment within `_get_keyspace`, as well as before each of the `query` calls that mentions the possibility of cql(?) injection - care should be taken when modifying this code.

waynew:
Not sure if this is related at all to #57129, but should check that out

dwoz:

PR #54644, opened 2019-09-19
[master] Porting #49507 to master

garethgreenaway:
Porting #49507 to master

### DIFF
--- a/salt/modules/artifactory.py
+++ b/salt/modules/artifactory.py
@@ -452,6 +452,7 @@ def __save_artifact(artifact_url, target_file, headers):
         checksum_url = artifact_url + ".sha1"
 
         checksum_success, artifact_sum, checksum_comment = __download(checksum_url, headers)
+        artifact_sum = salt.utils.stringutils.to_str(artifact_sum)
         if checksum_success:
             log.debug("Downloaded SHA1 SUM: %s", artifact_sum)
             file_sum = __salt__['file.get_hash'](path=target_file, form='sha1')


salt/modules/artifactory.py
```diff
@@ -452,6 +452,7 @@ def __save_artifact(artifact_url, target_file, headers):
         checksum_url = artifact_url + ".sha1"
 
         checksum_success, artifact_sum, checksum_comment = __download(checksum_url, headers)
+        artifact_sum = salt.utils.stringutils.to_str(artifact_sum)
```
waynew: I'm not convinced that this is the correct fix for this problem.

The issue appears to be that `__download` is producing (sometimes) binary data as the result of a `url.read()` command. I'm pretty sure that we should be checking the `Content-type` header for `text/*`, which we can do with 
```
info = url.info()
# python3
main_type = info.get_content_maintype()
# python2
main_type = info.maintype

if main_type == 'text':
    content = salt.utils.stringutils.to_str(content)
```

Granted, this is the only place that's using this `__download`, but that should correctly fix it for all possible uses.

Note that this `__download` code is *also* found in `salt/modules/nexus.py` - so it's possible that it's got the same sort of failure. Not sure if we should extract `__download`. Looks like the original code was from back in 2014 so maybe we should just refactor these usages to use `salt.utils.http.query`?

waynew (changes requested):

salt/modules/artifactory.py
waynew: https://stackoverflow.com/a/36882727/344286 for more info about python2/3

acornett:
@waynew , I've taken a swing at resolving this issue (which is impacting me) by using the built in http util, PR#56291

waynew:
I'm going to go ahead and close this in favor of #56291 - feel free to reopen or ping me if anyone disagrees.

PR #54648, opened 2019-09-19
[master] Porting #51740 to master

garethgreenaway:
Porting #51740 to master


salt/modules/file.py
```diff
@@ -3747,7 +3747,7 @@ def rmdir(path):
         return exc.strerror
 
 
-def remove(path):
+def remove(path, **kwargs):
```
s0undt3ch: This isn't your code, but why `kwargs` and not explicit keyword arguments?

s0undt3ch (changes requested):

salt/modules/file.py
DmitryKuzmenko: What's the reason of passing the force option that is ignored in `file.remove`?

waynew: FWIW, force can also be passed to the states: `file.absent` and `file.tidied` which are also merrily ignored :trollface:

waynew:
Updated this PR, added tests, ran the drop python 2 changes. Lint had a complaint about the use of `open` - that was a historical artifact, changed that to use a proper `with` statement.

DmitryKuzmenko:
I still not understand how this PR helps. This passes `force=True` from `state.file` calls to `module.file.remove` that receives this argument to `kwargs` and does NOT use it at all. So this PR does not change any behavior.

waynew:
@DmitryKuzmenko that's not actually the part that's doing anything.

This module is a virtual module, so on Windows passing `force=True` does something. Something super necessary, actually. On Linux when you have a file that's chmod 000, and you do `os.unlink` it will remove the file. On Windows it will die in a fire, because I guess deletion is the same thing as writing.

I agree that the `**kwargs` is a useless addition to `file.remove`. Though a couple of places only use partial **kwargs:

- chattr pops operator, flags, attributes, and versions
- manage_file uses it for win_owner, win_perms, win_deny_perms, win_inheritance, win_perms_reset

I would welcome the update that removes `**kwargs` and updates the tests to check win vs. Linux. You interested? ;)

DmitryKuzmenko:
@waynew sorry, my bad. I've missed the `__virtual__` method in `file` module. Thank you!

DmitryKuzmenko (dismissed):

Ch3LL (changes requested):
@waynew looks like you last updated the code here. Would you mind resolving the merge conflicts and adding a changelog file?

Ch3LL:
bump @waynew did you see my review comments?

Ch3LL (approved):

PR #54649, opened 2019-09-19
[master] Porting #51701 to master

garethgreenaway:
Porting #51701 to master


salt/modules/chocolatey.py
```diff
@@ -930,8 +930,14 @@ def version(name, check_remote=False, source=None, pre_versions=False):
...
-            packages[pkg] = {'installed': installed[pkg],
-                             'available': available[pkg]}
+            # Grab the current version from the package that was installed
+            packages[pkg] = {'installed': installed[pkg]}
+
+            # If there's a remote package available, then also include that
+            # in the dictionary that we return.
+            if pkg in available:
+                packages[pkg]['available'] = available[pkg]
+            continue
```
waynew: No need for implicit continue

salt/modules/chocolatey.py
```diff
@@ -930,8 +930,14 @@ def version(name, check_remote=False, source=None, pre_versions=False):
         available = list_(narrow=name, pre_versions=pre_versions, source=source)
 
         for pkg in packages:
-            packages[pkg] = {'installed': installed[pkg],
-                             'available': available[pkg]}
+            # Grab the current version from the package that was installed
+            packages[pkg] = {'installed': installed[pkg]}
```
waynew: I wonder - would this be better as `'available': available.get(pkg)`?

It produces an output like this:
```
local:
    ----------
    hosts:
        ----------
        available:
            None
        installed:
            - 1.1.0
```

Which feels better than:

```
local:
    ----------
    hosts:
        ----------
        installed:
            - 1.1.0
```

waynew (changes requested):
Definitely needs tests, could use some updates.

garethgreenaway:
Replaced with https://github.com/saltstack/salt/pull/56870

PR #54709, opened 2019-09-20
2017.7 remove sshagent

bryceml:
### What does this PR do?


waynew (approved):

waynew:
Should this just go into `master` now?

bryceml:
yeah, there is a pr open to every branch for this.
we could close the old ones if they're not going to make it in before we delete them.

bryceml:
https://github.com/saltstack/salt/pull/54735

bryceml:
This is no longer being tested.

PR #54712, opened 2019-09-20
2019.2.1 remove sshagent

bryceml:
### What does this PR do?


bryceml:
this doesn't matter since 2019.2.1 and 2019.2.2 have been tagged.

PR #54715, opened 2019-09-20
remove usage of sshagent plugin like the branch tests

bryceml:
### What does this PR do?


dwoz:
re-run full all

bryceml:
We don't test this branch anymore.

PR #54731, opened 2019-09-24
Fix returners not loading properly

pizzapanther:
### What does this PR do?

The `__new__` method was removed from the parent class so it must be removed here too, to load the returner properly.

So I:

- removed `__new__`
- always import `salt.minion`. Since the import changed from `__new__` to `__init__` you seem to need it always for every platform now. Also doesn't hurt to load it anyways.

### Tests written?
No

### Commits signed with GPG?
Yes


twangboy (approved):

s0undt3ch:
Tests are in

waynew (approved):

PR #54079, opened 2019-07-31
Fix cmd.run on MacOS -- wrong environment variables

ScoreUnder:
### What does this PR do?
**I am trying to merge this as a fix to 2019.2. Not sure if this is how things are done so please correct me if that is wrong**

Arranges the environment and working directory correctly when invoking cmd.run on MacOS.

### What issues does this PR fix or reference?

*PRs*
Relates: #47212 -- Keeps `su -l` which works around launchctl bootstrapping weirdness
Improves upon: #52632 -- Takes test and adds a few more.

*Issues*
Fixes: #51008
Fixes an apparently unreported issue where environment variables like PATH are not set correctly, resulting in `brew` and `pkg.installed` failing to work.

### Previous Behavior
With `runas`, the command was invoked with `su -l`, which blanked the environment and reset the cwd.

### New Behavior
With `runas`, the command is invoked with `su -l`, then the environment is corrected by running `bash -l`, then `cd` is used to correct the cwd.

### Tests written?
Yes

### Commits signed with GPG?
No

PR #54216, opened 2019-08-15
Improve macOS homebrew cask support

cdalvaro:
### What does this PR do?

When listing Homebrew packages, cask packages do not have the right namespace when they are from a tap different from the default one.

This PR fixes this issue recovering the right namespace for each cask package.

**Note:** This PR comes from #53756 

### Previous Behavior

```sh
local:
----------
          ID: FiraCode Nerd Font
    Function: pkg.installed
        Name: homebrew/cask-fonts/font-firacode-nerd-font
      Result: False
     Comment: The following packages failed to install/update: homebrew/cask-fonts/font-firacode-nerd-font
     Started: 17:59:08.334054
    Duration: 14324.087 ms
     Changes:
```

### New Behavior

```sh
local:
----------
          ID: FiraCode Nerd Font
    Function: pkg.installed
        Name: homebrew/cask-fonts/font-firacode-nerd-font
      Result: True
     Comment: All specified packages are already installed
     Started: 18:56:36.057920
    Duration: 48438.613 ms
     Changes:
```

### Tests written?

Yes

### Commits signed with GPG?

Yes

PR #54769, opened 2019-09-26
Call bash only when necessary on macOS

cdalvaro:
### What does this PR do?

This PR fixes a bug introduced on PR #54136 (and possibly on PR #54079 too) that calls `/bin/bash` when the user's default shell is not bash.

Also, users that have set a different version of bash as their default shell, they might be calling the wrong shell.

For example, for people with `zsh` as default shell the current behaviour is not right, neither for people with `/usr/local/bin/bash` as their main `shell` since the current code is invoking `/bin/bash` instead the right one.

### New Behavior

The new behaviour introduced in this PR invokes the right bash shell only when the users default shell is bash.

This is done in this way to load the right environment as it was intended in PR #54136.

### Tests written?

✅

### Commits signed with GPG?

✅


cdalvaro:
Following @waynew [advice]( https://github.com/saltstack/salt/pull/54216#issuecomment-538504386) from PR #54216 

> Hey @cdalvaro - as you may have heard, we're changing to a single branch release strategy. What does that mean for this PR?
> 
> Well, you've got tests so 👍
> 
> We will be working to migrate all PRs from non-`master` branches to `master`. If you'd like to get your PR in sooner, please rebase it against `master`, force push to your branch, and edit this PR to be against `master`. If you need any help with any of those steps, please let us know!

I have modified this PR too in order to fulfil the new release strategy.

cdalvaro:
I would like to know if this PR will be finally merged

cdalvaro:
Please @xeacott, could you revise this PR, it is open for nearly two months.

Thank you in advance.

weswhet (approved):
Looks good to me!

cdalvaro:
Please @xeacott, could you revise this PR, it solves an issue introduced in #54136.

By accepting this pull request, #54079 could be closed since it was duplicated in #54136 in order to rebased commits into 2019.2.1 instead of 2019.2

I mention @weswhet too, because I don't know if @xeacott is actively working on this project since I haven't heard from him for a long time.

Thank you all in advance!!

dwoz (approved):

Issue #54762, opened 2019-09-26
2019.2.1: Breaks Minion-Master Communication


margau:
### Description of Issue
After our minions were upgraded from 2019.2.0 to 2019.2.1, the communication (IPv6 only) with the master (either 2019.2.0 or 2019.2.1) is broken.
Minion is stuck at:
`
SaltReqTimeoutError, retrying. (2/7)
`

Via tcpdump communication - but with length "0" - is observable.

Master says:
`
Sep 26 11:59:36 salt salt-master[9266]: [WARNING ] /usr/lib/python3/dist-packages/zmq/eventloop/ioloop.py:211: RuntimeWarning: IOLoop.current expected instance of <class 'zmq.eventloop.ioloop.ZMQIOLoop'>, got <tornado.platform.epoll.EPollIOLoop object at 0x7fd9e798feb8>
Sep 26 11:59:36 salt salt-master[9266]:   ioloop.IOLoop.instance() is IOLoop.instance(), "tornado IOLoop already initialized"
`

### Setup
IPv6-only communication, no special configuration.

### Steps to Reproduce Issue
After minion is started:
`Sep 26 12:33:52  salt-minion[15491]: [DEBUG   ] Initializing new AsyncZeroMQReqChannel for ('/etc/salt/pki/minion', 'minion', 'tcp://<master-ip>', 'clear')
Sep 26 12:33:52  salt-minion[15491]: [DEBUG   ] Connecting the Minion to the Master URI (for the return server): tcp://<master-ip>
Sep 26 12:33:52  salt-minion[15491]: [DEBUG   ] Trying to connect to: tcp://<master-ip>
Sep 26 12:33:52  salt-minion[15491]: [DEBUG   ] salt.crypt.get_rsa_pub_key: Loading public key
Sep 26 12:33:57  salt-minion[15491]: [DEBUG   ] SaltReqTimeoutError, retrying. (1/7)
Sep 26 12:34:02  salt-minion[15491]: [DEBUG   ] SaltReqTimeoutError, retrying. (2/7)
Sep 26 12:34:07  salt-minion[15491]: [DEBUG   ] SaltReqTimeoutError, retrying. (3/7)
Sep 26 12:34:12  salt-minion[15491]: [DEBUG   ] SaltReqTimeoutError, retrying. (4/7)
Sep 26 12:34:17  salt-minion[15491]: [DEBUG   ] SaltReqTimeoutError, retrying. (5/7)
[...]

PR #54772, opened 2019-09-26
Fix import of pip modules

Orlando-Richards:
### What does this PR do?

When using pip versions between 10.0 and 18.0, the import of the pip.exceptions modules in pip_state will fail, as it was moved to pip._internal.exceptions in pip version 10.0.

Ref pip 10.0.0b1 release notes:
https://pip.pypa.io/en/stable/news/#b1-2018-03-31

This PR changes 18.1 to 10.0

https://github.com/saltstack/salt/issues/54762

### Previous Behavior
pip.installed state will fail to load with pip >=10.0  and <=18.0

### New Behavior
pip.installed state will load with pip >=10.0  and <=18.0

### Tests written?

No

### Commits signed with GPG?

No


OrlandoArcapix:
Initial PR had an error in it - updated now.

OrlandoArcapix:
Resolves https://github.com/saltstack/salt/issues/54773

cmcmarrow:
@OrlandoArcapix could you add a warning? If the pip version is lower than 18.1 have a warning "pip versions lower than 18.1 may not function currently". Also all changes need test now. Even if its a simple regression test. I can help you with the test if you need it.

OrlandoArcapix:
Hey @cmcmarrow -
would it be correct to put a warning in for versions < 18.1? I'm not sure that's relevant to the pip import of the exceptions module - certainly not something I could make a call on! If there's a requirement for salt to have pip >= 18.1, I'd have thought that should go elsewhere rather than in the exceptions import...

For testing - what would a suitable test be? Something like:
 * Install pip v 18.0
 * check it doesn't fail to import the module and/or do `pip install chicken`

? 

Would be great if you can give me a boost into getting a test in there!

dwoz:
@OrlandoArcapix @cmcmarrow this needs tests to get in.

dwoz:
Closing this in favor of #54807 which adds tests for the change

PR #54774, opened 2019-09-26
Fix broken jinja imports in masterless salt-call

terminalmage:
The addition of fileclient caching in the JinjaCacheLoader has broken
all jinja imports in state files, in masterless salt-call runs. This is
because masterless Pillar compilation uses the `PillarClient` while
regular file caching (states, etc.) use the `LocalClient`. Since Pillar
is compiled first in the masterless run, the class-level attribute where
the cached fileclient is stored only ever uses the PillarClient, and
thus any Jinja imports will cache the templates to the wrong location
(under `pillar_roots`).

This expands the cached fileclient support to distinguish between pillar
and non-pillar, and keep a separate cached fileclient for each.

### DIFF
--- a/salt/utils/jinja.py
+++ b/salt/utils/jinja.py
@@ -56,16 +56,18 @@ class SaltCacheLoader(BaseLoader):
     and only loaded once per loader instance.
     '''
 
+    _cached_pillar_client = None
     _cached_client = None
 
     @classmethod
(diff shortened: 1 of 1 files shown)


twangboy (approved):

dwoz:
Closing this in favor of #54780 which adds tests

Issue #54768, opened 2019-09-26
2019.2.1 Some Jinja imports not found


paul-palmer:
### Description of Issue

Starting with 2019.2.1, some, but not all Jinja imports fail with TemplateNotFound. It seems to be consistent by file. That is, the files that cannot be imported always fail to import, the files that can be imported always seem to succeed.

For example,

    vsphere: ----------
    vsphere:     Rendering SLS 'base:service.firewalld.enable' failed: Jinja error: macro/add-service.jinja
    vsphere: Traceback (most recent call last):
    vsphere:   File "/usr/lib/python2.7/site-packages/salt/utils/templates.py", line 394, in render_jinja_tmpl
    vsphere:     output = template.render(**decoded_context)
    vsphere:   File "/usr/lib/python2.7/site-packages/jinja2/environment.py", line 1008, in render
    vsphere:     return self.environment.handle_exception(exc_info, True)
    vsphere:   File "/usr/lib/python2.7/site-packages/jinja2/environment.py", line 780, in handle_exception
    vsphere:     reraise(exc_type, exc_value, tb)
    vsphere:   File "<template>", line 6, in top-level template code
    vsphere:   File "/usr/lib/python2.7/site-packages/salt/utils/jinja.py", line 189, in get_source
    vsphere:     raise TemplateNotFound(template)
    vsphere: TemplateNotFound: macro/add-service.jinja
    vsphere:
    vsphere: ; line 6
    vsphere:
    vsphere: ---
    vsphere: {% set template = 'service.firewalld.enable' %}
    vsphere:
    vsphere:
    vsphere:
    vsphere: {% from 'macro/add-dst-port.jinja' import add_dst_port %}
    vsphere: {% from 'macro/add-service.jinja' import add_service %}    <======================
    vsphere: {% from 'macro/vars.jinja' import host, nic, zone %}
    vsphere:
    vsphere: {% set dst_ports_path = pillar.global.firewalld.dst_ports_path %}
    vsphere: {% set services_path = pillar.global.firewalld.services_path %}

### Setup
Running standalone, initiated via packer.

### Steps to Reproduce Issue
Will provide full debug logs if needed (after sanitization effort).

### Versions Report

Salt Version:
           Salt: 2019.2.1
 
Dependency Versions:
           cffi: Not Installed
       cherrypy: Not Installed
       dateutil: Not Installed
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
[...]

Issue #54765, opened 2019-09-26
2019.2.1: Jinja from import broken


awerner:
### Description of Issue
Starting with salt version 2019.2.1, the following statement is broken:
```jinja
{% from "formula/map.jinja" import defaults with context %}
```
This is the error that occurs:
```
           Rendering SLS 'base:formula' failed: Jinja error: formula/map.jinja
       Traceback (most recent call last):
         File "/usr/lib/python2.7/site-packages/salt/utils/templates.py", line 394, in render_jinja_tmpl
           output = template.render(**decoded_context)
         File "/usr/lib/python2.7/site-packages/jinja2/environment.py", line 969, in render
           return self.environment.handle_exception(exc_info, True)
         File "/usr/lib/python2.7/site-packages/jinja2/environment.py", line 742, in handle_exception
           reraise(exc_type, exc_value, tb)
         File "<template>", line 1, in top-level template code
         File "/usr/lib/python2.7/site-packages/salt/utils/jinja.py", line 189, in get_source
           raise TemplateNotFound(template)
       TemplateNotFound: formula/map.jinja
       
       ; line 1
       
       ---
       {% from "formula/map.jinja" import defaults with context %}    <======================
```
The same setup is working with 2019.2.0.

### Setup

`formula/map.jinja`:
```jinja
{% set defaults = {'foo': 'bar'} %}
```
`formula/init.sls`:
```jinja
{% from "formula/map.jinja" import defaults with context %}
```

```bash
salt-call --local state.apply formula
```

### Steps to Reproduce Issue

This happens for us using our kitchen-salt based setup. So only salt-call and salt-minion are installed in the vms.

### Versions Report
```
Salt Version:
           Salt: 2019.2.1
 
Dependency Versions:
           cffi: Not Installed
       cherrypy: Not Installed
       dateutil: Not Installed
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.7.2
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.5.6
   mysql-python: Not Installed
      pycparser: Not Installed
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: Not Installed
         Python: 2.7.5 (default, Oct 30 2018, 23:45:53)
   python-gnupg: Not Installed
[...]

PR #54780, opened 2019-09-26
Fix masterless jinja imports

dwoz:
### What does this PR do?

Fixes and adds tests for #54765 #54768

### Tests written?

Yes

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/utils/jinja.py
+++ b/salt/utils/jinja.py
@@ -56,16 +56,18 @@ class SaltCacheLoader(BaseLoader):
     and only loaded once per loader instance.
     '''
 
+    _cached_pillar_client = None
     _cached_client = None
 
     @classmethod
     def shutdown(cls):
-        if cls._cached_client is None:
-            return
-        # PillarClient and LocalClient objects do not have a destroy method
-        if hasattr(cls._cached_client, 'destroy'):
-            cls._cached_client.destroy()
-        cls._cached_client = None
+        for attr in ('_cached_client', '_cached_pillar_client'):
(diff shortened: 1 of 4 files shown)


dwoz:
re-run full all

cmcmarrow (approved):

kiemlicz:
I've pulled the `jinja.py` and added to `_utils` dir in state tree root

Still getting:
```
{% from "_common/util.jinja" import retry with context %}    <======================
```

s0undt3ch:
@kiemlicz this PR adds a test which covers the proposed fix.

It seems that what you're trying to do is to "push" the fix to the minions through your state tree. That should be filed as a separate issue so we can better track it. It's not related to the change this PR is addressing.

dwoz:
re-run full all

dwoz:
This PR has the same patch as #54774 and adds tests

kiemlicz:
@s0undt3ch thanks for reply, now I'm truly disappointed

I know I can override the execution modules, state modules etc. using e.g. `_utils` dir from local state tree. 
It was always the fastest way to 'port' the fixes to the running salt version, however it turns that it doesn't work for `utils/jinja.py`: **why?**

Overriding the local installed file: `cp /srv/salt/base/_utils/jinja.py /usr/lib/python3/dist-packages/salt/utils/jinja.py` proves that the fix is working, however the sudden inability to use `_utils/` dir is frightening! Could you explain this please?

twangboy (approved):

garethgreenaway (approved):

PR #52990, opened 2019-05-13
Fallback to image id if we don't have an image name on query

sblaisot:
### What does this PR do?

When an openstack VM has been instanciated with an image that gets deleted from glance (openstack image store) while the VM is still running, salt-cloud cannot get back the image name. In this case, openstack, only returns the image id but cannot get the nae back. This leads to `salt-cloud -Q` not working.

### What issues does this PR fix or reference?

This PR fixes #52984 and #49438 

### Previous Behavior

```
# salt-cloud -Q -m my-cloud.map 
ovh-gra3:
    ----------
    openstack:
        ----------
        server1.company.tld:
            Absent
```

### New Behavior

```
# salt-cloud -Q -m my-cloud.map 
ovh-gra3:
    ----------
    openstack:
        ----------
        server1.company.tld:
            ----------
            fixed_ips:
                - 1.2.3.4
                - 2001:8d5e:104:9a64::bde:3c46
                - 10.1.2.3
            floating_ips:
            id:
                609467bf-6082-454b-a999-0f3512548260
            image:
                04787255-134b-4c96-b4ac-ac7d81e4467d
            name:
                server1.company.tld
            private_ips:
                - 10.1.2.3
            public_ips:
                - 1.2.3.4
                - 2001:8d5e:104:9a64::bde:3c46
            size:
                b2-7
            state:
                ACTIVE
```

### Tests written?

No

### Commits signed with GPG?

No

Issue #52984, opened 2019-05-13
salt-cloud partially fails to query openstack


sblaisot:
### Description of Issue/Question

Using salt-cloud 2019.2 with openstack driver configured, I can instanciate VMs without problem.

If I use a cloud-map file with 3 vms, two of which are up and running and the third is to be created, `salt-cloud -m my-cloud.map` correctly excludes the two running vms and creates the third one. So it appears in some circumstances running servers could be queried.

But after the server has been instanciated, `salt-cloud -Q` is unable to query openstack on this server and tells servers are absent despite they are up and running.

```
# salt-cloud -Q -m my-cloud.map 
ovh-gra3:
    ----------
    openstack:
        ----------
        server1.company.tld:
            Absent
        server2.company.tld:
            Absent
        server3.company.tld:
            Absent
```

Destroying a server with `salt-cloud -d server3.company.tld` leads to an error despite the server is correctly destroyed at the end.

```
# salt-cloud -d server3.company.tld
The following virtual machines are set to be destroyed:
  ovh-gra3:
    openstack:
      server3.company.tld

Proceed? [N/y] y
... proceeding
[ERROR   ] There was an error destroying machines: 'dict' object has no attribute 'public_ips'
```

### Setup

salt-cloud 2019.2 on debian 9 (installed using deb packages from official repo)

`/etc/salt/cloud.providers.d/ovh-gra3.conf `:
```
ovh-gra3:
  driver: openstack
  region_name: GRA3
  profile: ovh
  auth:
    username: REDACTED
    password: 'REDACTED'
    project_id: REDACTED
    project_name: 'REDATED'

  ssh_interface: private_ips
  ssh_key_name: my_ssh_key
  ssh_key_file: /path/to/keyfile

  minion:
    master: salt-master
	
  security_groups:
    - no-input

  nics:
    - net-id: xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx	
    - net-id: yyyyyyyy-yyyy-yyyy-yyyy-yyyyyyyyyyyy
```

`/etc/salt/cloud.profiles.d/ovh-gra3.conf`:
```
ovh-gra3_s1-2_deb9base:
  provider: ovh-gra3
  size: s1-2
  image: 'Debian 9'
  ssh_username: 'debian'
  script: bootstrap-salt
  script_args: |
    -j '{"master": "salt-master"}' stable 2019.2
```

`/srv/cloud/my-cloud.map`:
```
[...]

Issue #49438, opened 2018-08-30
salt-cloud fails to query openstack


markododa:
### Description of Issue/Question
Querying openstack doesn't work, (salt-cloud -Q)
Listing sizes and images works just fine

### Setup
This is the config file:

openstack_provider:
  minion:
    master: salt
    master_type: str
  driver: openstack
  region_name: RegionOne
  auth:
    username: 'admin'
    password: admin
    project_name: 'admin'
    auth_url: 'http://192.168.80.16:5000'

### Steps to Reproduce Issue
  File "/usr/lib/python2.7/dist-packages/salt/cloud/cli.py", line 271, in run
    self.function_provider, self.function_name, kwargs
  File "/usr/lib/python2.7/dist-packages/salt/cloud/__init__.py", line 1612, in do_function
    driver: self.clouds[fun](call='function')
  File "/usr/lib/python2.7/dist-packages/salt/cloud/clouds/openstack.py", line 413, in list_nodes_full
    ret[node.name]['image'] = node.image.name
AttributeError: 'dict' object has no attribute 'name'

### Versions Report
salt 2018.3.2 (Oxygen

Ch3LL:
im not able to replicate this just running `sudo salt-cloud -Q` with openstack configured. was this working previously for you or is this a new setup?

pzskc383:
I ran into this on rackspace connected via openstack driver.
I see no `name` in `node.image` dict there, only id.
For servers started from block storage, dict contsists of empty id only.

Catching raised attribute error gives no further errors when listing.

Hope this helps. 

Version report:
```
Salt Version:
            Salt: 2018.3.2
 
Dependency Versions:
 Apache Libcloud: 1.2.1
            cffi: 1.11.4
        cherrypy: 3.8.0
        dateutil: 2.6.1
       docker-py: Not Installed
           gitdb: Not Installed
       gitpython: Not Installed
           ioflo: Not Installed
          Jinja2: 2.10
         libgit2: Not Installed
         libnacl: 1.4.4
        M2Crypto: 0.24.0
            Mako: 1.0.6
    msgpack-pure: Not Installed
  msgpack-python: 0.5.1
    mysql-python: Not Installed
       pycparser: 2.14
        pycrypto: Not Installed
    pycryptodome: Not Installed
          pygit2: Not Installed
          Python: 2.7.14 (default, Apr 16 2018, 06:30:00)
    python-gnupg: 0.4.1
          PyYAML: 3.12
[...]

PR #54778, opened 2019-09-26
fix broken salt-cloud openstack query

Akm0d:
### What does this PR do?
Cherry pick https://github.com/saltstack/salt/pull/54761 into Master

PR #54782, opened 2019-09-26
Fallback to image id if we don't have an image name

Akm0d:
### What does this PR do?
Cherry-picks https://github.com/saltstack/salt/pull/52990 into Master

Solves
https://github.com/saltstack/salt/issues/52984
https://github.com/saltstack/salt/issues/49438


Akm0d:
Cherry-picking into https://github.com/saltstack/salt/pull/54778

Issue #54758, opened 2019-09-26
salt-call state.show_states gives error on 2019.2.1-1


verhage:
### Description of Issue
I just updated my salt-minion on CentOS 7 to version 2019.2.1-1.el7. Now when I run `salt-call state.show_states`, I get the following error;

```
Passed invalid arguments: string indices must be integers.

Usage:

    Returns the list of states that will be applied on highstate.

    CLI Example:

    .. code-block:: bash

        salt '*' state.show_states

    .. versionadded:: 2019.2.0
```

On a minion still running the previous version (2019.2.0-1.el7) everything works fine.

### Setup
salt-master and minion both running on CentOS 7 7.6.1810, salt version 2019.2.1-1.el7

### Steps to Reproduce Issue

Update minion to latest version and call state.show_states:

```
yum update
salt-call state.show_states
```

### Versions Report
```
Salt Version:
           Salt: 2019.2.1

Dependency Versions:
           cffi: 1.6.0
       cherrypy: Not Installed
       dateutil: Not Installed
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.7.2
        libgit2: 0.26.3
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.5.6
   mysql-python: Not Installed
      pycparser: 2.14
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: 0.26.4
         Python: 2.7.5 (default, Jun 20 2019, 20:27:34)
   python-gnupg: Not Installed
         PyYAML: 3.11
          PyZMQ: 15.3.0
           RAET: Not Installed
          smmap: Not Installed
        timelib: Not Installed
        Tornado: 4.2.1
            ZMQ: 4.1.4

System Versions:
           dist: centos 7.6.1810 Core
         locale: UTF-8
        machine: x86_64
        release: 3.10.0-957.27.2.el7.x86_64
         system: Linux
        version: CentOS Linux 7.6.1810 Core
```

waynew:
Hi @verhage thanks for the report - I'm trying to reproduce this, and it's not failing for me yet.

Question: do you have any Jinja in your states?

verhage:
Yes, we use Jinja templates in almost every state. I'll try and see if I can reproduce this in Vagrant.

Ch3LL:
[...]

PR #54785, opened 2019-09-26
Fix state.show_states when sls file missing in top file

Ch3LL:
### What does this PR do?
When a sls file is defined in a top.sls file that is missing a TypeError was raised when using `state.show_states`. This PR ensures we do not stacktrace and handle this situation

https://github.com/saltstack/salt/issues/54758

### Previous Behavior

```
Passed invalid arguments: string indices must be integers.

Usage:

    Returns the list of states that will be applied on highstate.

    CLI Example:

    .. code-block:: bash

        salt '*' state.show_states

    .. versionadded:: 2019.2.0
```

### New Behavior

```
local:
       - No matching sls found for 'cloud' in env 'base'
```

### Tests written?

Yes

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/modules/state.py
+++ b/salt/modules/state.py
@@ -1634,6 +1634,9 @@ def show_states(queue=False, **kwargs):
             raise Exception(result)
 
         for s in result:
+            if not isinstance(s, dict):
+                _set_retcode(result)
+                return result
             states[s['__sls__']] = True
     finally:
         st_.pop_active()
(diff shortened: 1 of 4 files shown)


Ch3LL:
re-run full all

garethgreenaway (approved):

s0undt3ch (approved):

dwoz (approved):

Issue #54741, opened 2019-09-25
Schedulers Fail to Run


kjkeane:
### Description of Issue
Schedulers are failing to run with a `KeyError: u'run`

I am testing this by running it manually with 
`salt-call schedule.run_job test-scheduler-present -l debug`

Error Ouput:
```Python
2019-09-25 14:06:41,391 [salt.utils.schedule:461 ][INFO    ][6910] Running Job: test-scheduler-present
2019-09-25 14:06:41,391 [tornado.application:611 ][ERROR   ][6910] Exception in callback <functools.partial object at 0x7f5bcc528520>
Traceback (most recent call last):
  File "/usr/lib64/python2.7/site-packages/tornado/ioloop.py", line 591, in _run_callback
    ret = callback()
  File "/usr/lib64/python2.7/site-packages/tornado/stack_context.py", line 274, in null_wrapper
    return fn(*args, **kwargs)
  File "/usr/lib64/python2.7/site-packages/tornado/ioloop.py", line 597, in <lambda>
    self.add_future(ret, lambda f: f.result())
  File "/usr/lib64/python2.7/site-packages/tornado/concurrent.py", line 214, in result
    raise_exc_info(self._exc_info)
  File "/usr/lib64/python2.7/site-packages/tornado/gen.py", line 888, in run
    yielded = self.gen.throw(*exc_info)
  File "/usr/lib/python2.7/site-packages/salt/minion.py", line 959, in handle_event
    yield [minion.handle_event(package) for minion in self.minions]
  File "/usr/lib64/python2.7/site-packages/tornado/gen.py", line 882, in run
    value = future.result()
  File "/usr/lib64/python2.7/site-packages/tornado/concurrent.py", line 214, in result
    raise_exc_info(self._exc_info)
  File "/usr/lib64/python2.7/site-packages/tornado/gen.py", line 668, in callback
    result_list.append(f.result())
  File "/usr/lib64/python2.7/site-packages/tornado/concurrent.py", line 214, in result
    raise_exc_info(self._exc_info)
  File "/usr/lib64/python2.7/site-packages/tornado/gen.py", line 232, in wrapper
    yielded = next(result)
  File "/usr/lib/python2.7/site-packages/salt/minion.py", line 2371, in handle_event
    self.manage_schedule(tag, data)
  File "/usr/lib/python2.7/site-packages/salt/minion.py", line 2233, in manage_schedule
    self.schedule.run_job(name)
[...]

PR #54799, opened 2019-09-28
Fix to scheduler when job without a time element is run with schedule.run_job

garethgreenaway:
### What does this PR do?
If a scheduled job does not contains a time element parameter, running that job with schedule.run_job fails and causes a traceback because data['run'] does not exist.

#54741 

### Previous Behavior
Jobs without time elements when run with "schedule.run_job" failed with traceback.

### New Behavior
Ensure data['run'] exists and handle situations when it does not.  Jobs run as expected.

### Tests written?
Yes

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/utils/schedule.py
+++ b/salt/utils/schedule.py
@@ -475,8 +475,7 @@ class Schedule(object):
                                            datetime.datetime.now())
 
         # Grab run, assume True
-        run = data.get('run', True)
-        if run:
+        if data.get('run'):
             log.info('Running Job: %s', name)
             self._run_job(func, data)
 
--- a/salt/utils/schedule.py
+++ b/salt/utils/schedule.py
@@ -214,7 +214,7 @@ class Schedule(object):
         # dict we treat it like it was there and is True
 
         # Check if we're able to run
-        if not data['run']:
+        if 'run' not in data or not data['run']:
             return data
         if 'jid_include' not in data or data['jid_include']:
             jobcount = 0
@@ -463,7 +463,10 @@ class Schedule(object):
 
         if 'name' not in data:
             data['name'] = name
-        log.info('Running Job: %s', name)
+
+        # Assume run should be True until we check max_running
+        if 'run' not in data:
+            data['run'] = True
 
         if not self.standalone:
             data = self._check_max_running(func,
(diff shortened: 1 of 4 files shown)


dwoz (approved):

salt/utils/schedule.py
```diff
@@ -474,6 +477,7 @@ def run_job(self, name):
         # Grab run, assume True
         run = data.get('run', True)
```
dhiltonp: This line is no longer needed

salt/utils/schedule.py
```diff
@@ -474,6 +477,7 @@ def run_job(self, name):
...
         if run:
```
dhiltonp: Just check `data['run']` as 10 lines up we guarantee data['run'] exists?

dhiltonp (changes requested):
There's a tiny bit of duplication, maybe we can clean it up?

salt/utils/schedule.py
```diff
@@ -214,7 +214,7 @@ def _check_max_running(self, func, data, opts, now):
         # dict we treat it like it was there and is True
 
         # Check if we're able to run
-        if not data['run']:
+        if 'run' not in data or not data['run']:
```
dhiltonp: `_check_max_running` is called in 2 places (`eval` and `run_job`). `eval` guarantees that `data['run']` is set, now `run_job` does as well.

With current code, `'run' not in data` will always be False. I'm ok with defensive programming, just thought I'd point that out.

dhiltonp: If we change the PR for the other issue, maybe `not data.get('run', False)` would be better? Just emphasizing the important bit is 'run' being False, with a "just in case" second return.

dhiltonp: Don't worry about that last bit, I was overthinking it.

dhiltonp (approved):

twangboy (approved):

salt/utils/schedule.py
garethgreenaway: :+1: Good catch, definitely more efficient.

salt/utils/schedule.py
garethgreenaway: Yeah. That makes more sense.  My fix was just a quick fix to get the functionality working again but this is a cleaner approach.

PR #54823, opened 2019-09-30
ip_bracket can now accept ipv6 addresses with brackets

dhiltonp:
### What does this PR do?
`ip_bracket` is now more liberal in its input - attempting to add brackets to an IPv6 address that already has brackets is not fatal.

#54762

### Tests written?
Yes

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/utils/zeromq.py
+++ b/salt/utils/zeromq.py
@@ -80,8 +80,11 @@ def check_ipc_path_max_len(uri):
 
 def ip_bracket(addr):
     '''
-    Convert IP address representation to ZMQ (URL) format. ZMQ expects
-    brackets around IPv6 literals, since they are used in URLs.
+    Ensure IP addresses are URI-compatible - specifically, add brackets
+    around IPv6 literals if they are not already present.
     '''
+    addr = str(addr)
+    addr = addr.lstrip('[')
+    addr = addr.rstrip(']')
     addr = ipaddress.ip_address(addr)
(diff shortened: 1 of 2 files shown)


salt/utils/zeromq.py
```diff
@@ -80,8 +80,11 @@ def check_ipc_path_max_len(uri):
...
 def ip_bracket(addr):
     '''
-    Convert IP address representation to ZMQ (URL) format. ZMQ expects
-    brackets around IPv6 literals, since they are used in URLs.
+    Ensure IP addresses are URI-compatible - specifically, add brackets
+    around IPv6 literals if they are not already present.
     '''
+    addr = str(addr)
+    addr = addr.lstrip('[')
+    addr = addr.rstrip(']')
```
waynew: Isn't it possible to have a partially bracketed address? I don't remember the exact syntax, but can't you have something like `[::]:beef`?

waynew: To answer my question, [nope](https://en.wikipedia.org/wiki/IPv6_address#Literal_IPv6_addresses_in_network_resource_identifiers), not according to Wikipedia 👍

waynew (approved):

salt/utils/zeromq.py
dhiltonp: We're assuming this is just an ip address (no port attached) :)

Akm0d (approved):

waynew (approved):

garethgreenaway (approved):

Issue #54755, opened 2019-09-26
2019.2.1/2019.2.0 pip failures even when not using pip


Reiner030:
While testing my setup with Debian Buster and "stable" Saltstack version (2019.2.0) on a vagrant instance I run into weird problems... 
Especially i didn't use pip and most of the other modules notified in the failure messages I was search for over half hour for the cause:

````
# salt-call state.apply
[ERROR   ] Failed to import states pip_state, this is due most likely to a syntax error:
Traceback (most recent call last):
  File "/usr/lib/python2.7/dist-packages/salt/loader.py", line 1567, in _load_module
    mod = imp.load_module(mod_namespace, fn_, fpath, desc)
  File "/usr/lib/python2.7/dist-packages/salt/states/pip_state.py", line 58, in <module>
    del pip
NameError: name 'pip' is not defined
[ERROR   ] Failed to import states pkg, this is due most likely to a syntax error:
Traceback (most recent call last):
  File "/usr/lib/python2.7/dist-packages/salt/loader.py", line 1567, in _load_module
    mod = imp.load_module(mod_namespace, fn_, fpath, desc)
  File "/usr/lib/python2.7/dist-packages/salt/states/pkg.py", line 79, in <module>
    import logging
  File "/usr/lib/python2.7/logging/__init__.py", line 26, in <module>
    import sys, os, time, cStringIO, traceback, warnings, weakref, collections
  File "/usr/lib/python2.7/os.py", line 119, in <module>
    sys.modules['os.path'] = path
AttributeError: 'module' object has no attribute 'modules'
[ERROR   ] Failed to import states pkgbuild, this is due most likely to a syntax error:
Traceback (most recent call last):
  File "/usr/lib/python2.7/dist-packages/salt/loader.py", line 1567, in _load_module
    mod = imp.load_module(mod_namespace, fn_, fpath, desc)
  File "/usr/lib/python2.7/dist-packages/salt/states/pkgbuild.py", line 48, in <module>
    import logging
  File "/usr/lib/python2.7/logging/__init__.py", line 26, in <module>
    import sys, os, time, cStringIO, traceback, warnings, weakref, collections
  File "/usr/lib/python2.7/os.py", line 119, in <module>
    sys.modules['os.path'] = path
AttributeError: 'module' object has no attribute 'modules'
[ERROR   ] Failed to import states pkgrepo, this is due most likely to a syntax error:
Traceback (most recent call last):
[...]

PR #54826, opened 2019-09-30
Fix issue 54755 and add regression tests

dwoz:
### What does this PR do?

Address #54755 and add regression tests.

In general we should try to address this kind of thing without mucking around with already imported modules. Ideally we'd recognize that pip was upgraded and re-start the minion automatically. This will however address the bugs reported against the pip_state and other states using the `unless` requisite. 

#54755

### Previous Behavior

If no pip module is in the python environment imports of the pip state failed with multiple tracebacks. This is due to us trying to `del pip` when there is no `pip` variable defined.

### New Behavior

Use a `purge_pip` method before trying to import `pip` this method will only try to remove the `pip` module and/or `pip` variable if it exists. 

### Tests written?

Yes

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/states/pip_state.py
+++ b/salt/states/pip_state.py
@@ -58,8 +58,10 @@ def purge_pip():
         sys.modules.pop(name)
         del entry
 
-    if 'pip' in globals() or 'pip' in locals():
-        del pip
+    if 'pip' in globals():
+        del globals()['pip']
+    if 'pip' in locals():
+        del locals()['pip']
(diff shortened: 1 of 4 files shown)


cmcmarrow (approved):

Akm0d (approved):

garethgreenaway (approved):

s0undt3ch (approved):

dwoz:
re-run full all

dwoz:
re-run full all

PR #54827, opened 2019-09-30
24501 - add Note to docs explaining when the behavior is non-deterministic

wwalker:
### What does this PR do?
Documents the behavior until 24501 is fixed

24501

### Tests written?
No

### Commits signed with GPG?

Yes


wwalker:
typo "in" rather than "is" 
Closing

PR #54735, opened 2019-09-25
Master spot instances

bryceml:
### What does this PR do?
It uses spot instances if they are available at the moment

### Previous Behavior
We always used on-demand instances

### New Behavior
It attempts to use a spot instance before trying an on-demand instance

### Commits signed with GPG?

Yes

PR #54840, opened 2019-10-01
Changed the default installed package for zypper

Akm0d:
### What does this PR do?
Fix failing test because of non-existant package on OpenSuse


Akm0d:
included in https://github.com/saltstack/salt/pull/54735

Issue #54694, opened 2019-09-19
Core napalm grains assigned after custom grains


jtraub91:
### Description of Issue
According to https://docs.saltstack.com/en/latest/topics/grains/#precedence core grains are assigned first and thus should be overwritten by custom grains in _grains/. However, this is not the case for core grains included in salt/grains/napalm.py

### Setup
Need an actual or virtual device capable of being controlled by napalm proxy minion. Need to then configure proxy nominally via pillar data.

### Steps to Reproduce Issue
Create custom grain 'interfaces' to override salt-napalm included 'interfaces'. Does not work as described in link above

jtraub91:
PR opened here https://github.com/saltstack/salt/pull/54853

sagetherage:
@Akm0d you did triage on this ticket, is it a regression? what is the criticality?

PR #54853, opened 2019-10-02
sorted keys for non core. grains

jtraub91:
### What does this PR do?
Fixes problem due to grains module evaluation. Specifically for proxyenabled grains modules, the salt-core included non-core grains were being evaluated after custom grains. So for example, the grains included in grains/napalm.py were being evaluated _after_ my custom grains included in /srv/salt/_grains. This was a problem since I wished to overwrite grains according to the precedence define in the [docs](https://docs.saltstack.com/en/latest/topics/grains/#precedence) and it was not behaving the way it should.

The way I solved this (while not perfect) was to sort the non-core (not core.xxx) grains so that if you wish to overwrite a non-core salt-core included grain's key, simply use a \_\_virtualname\_\_ that has a later key alphabetically. E.g. my custom grain module 'napalm_mod' will overwrite similar keys in 'napalm' grains module

https://github.com/saltstack/salt/issues/54694

### Previous Behavior
Non-core salt-core include grains module would be evaluated last, thus taking precedence over custom grains

### New Behavior
Custom grain will take precedence if virtualname has a later key alphabetically

### Tests written?

No

### Commits signed with GPG?

No

### DIFF
--- a/tests/integration/loader/test_ext_grains.py
+++ b/tests/integration/loader/test_ext_grains.py
@@ -24,6 +24,7 @@ from tests.support.unit import skipIf
 # Import salt libs
 import salt.config
 import salt.loader
+import salt.utils
 
 log = logging.getLogger(__name__)
 
@@ -79,7 +80,7 @@ class GrainsPrecedenceTest(ModuleCase):
         Test that /etc/salt/grains overrides core
         '''
         etc_salt_grains_file = os.path.join(TMP_CONF_DIR, 'grains')
-        with open(etc_salt_grains_file, 'w') as etc_:
+        with salt.utils.fopen(etc_salt_grains_file, 'w') as etc_:
             etc_.write('localhost: cosmos')
 
         self.run_function('saltutil.refresh_grains')
@@ -98,12 +99,15 @@ class GrainsPrecedenceTest(ModuleCase):
         self.assertEqual(grains.get('localhost'), 'alien')
         self.run_function('grains.delkey', arg=('localhost'))
 
-    def test_minion_config_override(self):
+    def test_minion_config_grains(self):
+        '''
+        Test that minion config overrides core
+        '''
         minion_config_filename = os.path.join(TMP_CONF_DIR, 'minion')
         # read original contents
         opts = salt.config.minion_config(minion_config_filename)
         opts['grains']['localhost'] = 'not-local'
-        with open(minion_config_filename, 'w') as cfg_:
+        with salt.utils.fopen(minion_config_filename, 'w') as cfg_:
             yaml.safe_dump(opts, cfg_, default_flow_style=False)
 
         self.run_function('saltutil.refresh_grains')
@@ -113,18 +117,18 @@ class GrainsPrecedenceTest(ModuleCase):
 
         # restore original contents
         del opts['grains']['localhost']
-        with open(minion_config_filename, 'w') as cfg_:
+        with salt.utils.fopen(minion_config_filename, 'w') as cfg_:
             yaml.safe_dump(opts, cfg_, default_flow_style=False)
 
         self.run_function('saltutil.refresh_grains')
 
-    def test_overwrite_localhost_custom_grains_module(self):
+    def test_custom_overwrite(self):
         '''
         Test that custom grain module overwrites core 'localhost' grain
         '''
         filename = 'overwrite_localhost_{0}.py'.format(int(time.time()))
         custom_grain_abs_path = os.path.join(BASE_FILES, '_grains', filename)
-        with open(custom_grain_abs_path, 'w') as file_:
+        with salt.utils.fopen(custom_grain_abs_path, 'w') as file_:
             file_.write("def overwrite_localhost():{0}".format(os.linesep))
             file_.write("    return {{'localhost': 'overwrite-localhost'}}{0}".format(os.linesep))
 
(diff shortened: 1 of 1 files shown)


xeacott:
I do not know the implications of this change, but there are some lint issues you'll need to fix first. Then I'll have someone else on the core team review this.

jtraub91:
Thanks @xeacott ! I fixed the lint and build errors. Looking forward to feedback.

xeacott:
Before we can get this merged, you'll need to change the target branch and add a test. This means re-basing and targeting `master` branch. I believe there should be some unit tests for example purposes in tests\unit\test_loader.py

cmcmarrow:
This should go into 2019.2.1.

salt/loader.py
```diff
@@ -815,7 +815,9 @@ def grains(opts, force_refresh=False, proxy=None):
             grains_data.update(ret)
 
     # Run the rest of the grains
-    for key in funcs:
+    # sort keys so that custom grains for proxyenabled modules can overwrite salt-core included grains
+    sorted_keys = sorted(funcs.keys())
```
cmcmarrow: `for key in sorted(funcs.keys()):` sorted_keys is never used again. Also it does not change the readability of the code.

cmcmarrow (changes requested):
Thanks for the code, but salt now want tests with every code change now.

salt/loader.py
jtraub91: Absolutely, was planning to do this. Less lines is better 🐍

jtraub91:
Rebased on master. Will add test asap

jtraub91:
May I propose that this commit doesn't, in fact, need a test? The commit simply sorts the keys to execute grains modules in alphabetical order. What type of test is warranted here?

waynew:
@jtraub91 you kind of stated it yourself:

> So for example, the grains included in grains/napalm.py were being evaluated after my custom grains included in /srv/salt/_grains. This was a problem since I wished to overwrite grains according to the precedence define in the docs and it was not behaving the way it should.

I would assert that there should be tests written to ensure that grains are evaluated in the correct order. TBH it's really weird that there aren't (apparently) any tests covering this situation.

But if there's a faulty behavior, we shouldn't add code to salt that fixes the faulty behavior unless we can write a test that fail when the faulty behavior exists. Otherwise someone may come in and change the code later, re-introducing this error.

jtraub91:
@waynew understood.

I added better testing for this situation. The real issue is that there are 'core' grains that aren't in the 'core' grains module. If you look at loader.py core grains will always run first however 'non-core' grains that are included in the repo can only be overwritten with an alphabetically-superseding module name (with this change).

Perhaps the better way to handle this is to change all the non-core grains included in the repo to have \_\_virtualname\_\_ = 'core'. Then they would always get run first no matter custom module name. <- that sounds like a good way to go now that I say it, although there may be other complications introduced. I'm going to take a break for today, but interested to hear your thought if any, as I mull this over and test it soon.

jtraub91:
Another issue with grains precedence, /etc/salt/minion grains take precedence over custom _grains
Appeared to be this way via testing but also quite evident by examining loader.py (https://github.com/saltstack/salt/blob/master/salt/loader.py#L846)

We will want to either make a change to the code, or the docs (https://docs.saltstack.com/en/latest/topics/grains/#precedence). This probably belongs in a separate issue / PR.

waynew:
I'll bring this up with the core team - it seems like we will probably want to take a step back and figure out a long-term strategy for these.

jtraub91:
I've spoke with some of my colleagues and the path forward that makes most sense to us, is the following:

Since there are grains included in this repo which aren't under the modules scope 'core.' (e.g. napalm, zfs, etc) and changing these to be ```__virtualname__ = 'core'``` could be a breaking change, the grains evaluation should be as follows:
1. Evaluate grains modules contained in ```salt/grains```
  a. First execute ```core.``` grains
  b. Then execute other grains modules in alphabetical order
2. Evaluate custom ```_grains``` in alphabetical order

jtraub91:
I'm willing to investigate further and make this change (and add a test), but have been unsuccessful so far. The logic to execute the modules within ```salt.loader.grains``` is not sufficient, and perhaps this could be fixed there, but maybe it is better to be fixed before this point, perhaps in the ordering of modules via the ```_module_dirs``` function in loader.py ?

waynew:
@jtraub91 Sorry for the delay - we finally had a chance to discuss things, and from what I can tell what you're heading after seems like the right direction.

I'd say it's "sort of" a breaking change - those grains *are* salt core grains even though they don't live in the `.py` file. But it could break some things for people who are relying on the incorrect behavior. Because of that we'll make sure to call special attention to it when we release the fixes - hopefully that's enough for people to be aware that we're fixing broken behavior they could have been relying on.

We'd love to see a solid set of unit tests that verify that grains are evaluated in the order the documentation says they are. If you're not able to, that's understandable. But it would be awesome  and appreciated!

If not, at least some minimal test to verify that your code does what it says it will will be 👍 .

jtraub91:
But if those grains are 'core' grains then there should be no worries aliasing them under the ```__virtualname__ = 'core'``` namespace. Any reliance would be purely an error in overwriting an existing grain of the same name. 

Accidental reliance aside, from what you are saying, it sounds like, to me, all 'core' grains that aren't namespaced under 'core' _should_ be. After all that is the criteria that ```salt.loader.grains``` is using when determining if it is in fact a core grain: https://github.com/saltstack/salt/blob/master/salt/loader.py#L783

waynew:
> Accidental reliance aside, from what you are saying, it sounds like, to me, all 'core' grains that aren't namespaced under 'core' should be.

That's the way I understand it, yes!

xeacott:
@jtraub91 Would you mind also resolving the merge conflict?

dwoz:

PR #54888, opened 2019-10-04
.modules.pacmanpkg.purge: Do not cascade

ayekat:
### What does this PR do?

Remove cascade option from the pacmanpkg module.

### Previous Behavior
`pacmanpkg.purge` invokes pacman with `-Rsc`.

The `-c` option there is "cascade", meaning that if a package is required by any other packages, those will be removed, too. This is not the default behaviour, and can lead to the removal of likely still required packages ([pacman(8)][man:pacman] also warns about using this option).

Based on the module description for this action, the user does generally not expect this to remove packages like this.

[man:pacman]: http://jlk.fjfi.cvut.cz/arch/manpages/man/core/pacman/pacman.8.en#REMOVE_OPTIONS_(APPLY_TO__FI_-R_FR)

### New Behavior
`pacmanpkg.purge` invokes pacman with `-Rs`. This does what the description claims: Remove packages and its dependencies.

This will error out if the package is still required by another package (rather than "force-remove this package no matter what").

### Tests written?

No

### Commits signed with GPG?

No

### DIFF
--- a/salt/modules/pacmanpkg.py
+++ b/salt/modules/pacmanpkg.py
@@ -718,7 +718,7 @@ def _uninstall(action='remove', name=None, pkgs=None, **kwargs):
     if not targets:
         return {}
 
-    remove_arg = '-Rsc' if action == 'purge' else '-R'
+    remove_arg = '-Rs' if action == 'purge' else '-R'
 
     cmd = []
     if salt.utils.systemd.has_scope(__context__) \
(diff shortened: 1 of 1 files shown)


ayekat:
This PR may not be required if #54889 gets merged, as this would be a non-issue then.

Akm0d:
Related to https://github.com/saltstack/salt/pull/55375

sagetherage:
@ayekat please rebase this against the master branch

sagetherage:
@ayekat can you please rebase this against the `master` branch

ayekat:
@sagetherage Actually I think there is no point in keeping this PR open if there's already https://github.com/saltstack/salt/pull/55375, which renders this PR obsolete. Unfortunately I haven't yet found any time to read into writing tests for Salt (but that would block both PRs anyway).

So I'm going to close this one here and focus on the other one.

But thanks!

PR #54893, opened 2019-10-04
Adds an example of running states on the Salt Master during an orchestration

grandmasterv:
### What does this PR do?
Adds an example of running states on the Salt Master during an orchestration.

There was no example of running states on the Salt Master directly during an orchestration in the docs.  This led me to believe that in order to have something run on the Salt Master I had to either use a runner or target the Salt Master specifically which I did not want to do.

### Previous Behavior
Only provided an example of running a state on a minion.

### New Behavior
Provides an example of running states on Minions and the Salt Master running the orchestration.

### Tests written?
No

### Commits signed with GPG?
Yes

### DIFF
--- a/doc/topics/orchestrate/orchestrate_runner.rst
+++ b/doc/topics/orchestrate/orchestrate_runner.rst
@@ -140,7 +140,7 @@ can specify the "name" argument to avoid conflicting IDs:
 State
 ^^^^^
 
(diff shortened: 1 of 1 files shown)


dwoz:
@grandmasterv We are no longer accepting PRs against `2017.7` Please re-open this against the `master` branch if it is still an issue there.

PR #54903, opened 2019-10-05
adding multipart/form-data file posting to http.query util

nicholasmhughes:
### What does this PR do?
This PR adds the ability for the requests backend to post files as multipart/form-data. I'm not really familiar with tornado, or else I'd add that too... updates welcome!

N/A

### Previous Behavior
The http.query utility could not handle multipart/form-data, which is commonly used in file uploads to web servers.

### New Behavior
Now the http.query utility supports that capability in the requests backend.

### Tests written?
Yes

### Commits signed with GPG?
Yes


salt/utils/http.py
```diff
@@ -70,6 +70,11 @@
 import tornado.simple_httpclient
 from tornado.httpclient import HTTPClient
 
+if six.PY2:
+    from StringIO import StringIO
+elif six.PY3:
+    from io import StringIO
```
ScoreUnder: `from salt.ext.six import StringIO` should cover this

dwoz (approved):

Issue #54910, opened 2019-10-07
Bug with winexe in cloud.py module in version 2019.2.1


Mapel88:
### Description of Issue
After upgrade from 2018.3.4 to 2019.2.1 winexe command got messed up.

### Setup
Install windows server via cloud module.
my cloud profile:
```
nl1-win2016-backend:
  provider: nlxen
  image: windows_2016_backend_template
  clone: True
  win_installer: /srv/salt/win/files/Salt-Minion-2018.3.4-Py3-AMD64-Setup.exe
  win_username: Administrator
  win_password: '1qaz2wsX'
  minion:
    master:
      - myMaster.my.domain,com
  ipv4_cidr: 1.1.1.1/24
  ipv4_gw: 1.1.1.254
  use_winrm: False
  winrm_verify_ssl: False
  port_timeout: 30

```
### Steps to Reproduce Issue
1. The winexe command got messed up and presented the actual password instead just "REDUCTED"
2. The command syntax also got messed up and missing "" for the sc command itself.
```
[INFO    ] Running command under pid 10714: 'winexe -U 'Administrator%XXX-REDACTED-XXX' //1.1.1.1 winexe -U 'Administrator%<plainTextPassword>' //1.1.1.1 sc query winexesvc query winexesvc'
[DEBUG   ] Executing command(PID 10714): 'winexe -U 'Administrator%XXX-REDACTED-XXX' //172.16.31.178 winexe -U 'Administrator%plainTextPassword' //1.1.1.1 sc query winexesvc query winexesvc'
[DEBUG   ] Return code was 1

```
### Versions Report
Salt Version:
           Salt: 2019.2.1

Dependency Versions:
           cffi: 1.6.0
       cherrypy: unknown
       dateutil: 1.5
      docker-py: Not Installed
          gitdb: 0.6.4
      gitpython: 1.0.1
          ioflo: Not Installed
         Jinja2: 2.7.2
        libgit2: 0.24.6
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.4.8
   mysql-python: Not Installed
      pycparser: 2.14
       pycrypto: 2.6.1
   pycryptodome: 3.6.1
         pygit2: 0.24.2
         Python: 2.7.5 (default, Oct 30 2018, 23:45:53)
   python-gnupg: Not Installed
         PyYAML: 3.11
          PyZMQ: 15.3.0
[...]

Issue #54904, opened 2019-10-06
salt cloud fails to start minion on windows after upgrade from 2018.3.4 to 2019.2.1


Mapel88:
### Description of Issue
After upgrading from 2018.3.4 to 2019.2.1 salt-minion fails to start on windows when deploying server via salt cloud.
Service is still stopping while start command is activated.

### Setup
I'm using salt minion version 2018.3.4 setup due to a bug in "system.join_domain" #54869 but this happens in version 2019.2.1 as well.

My cloud profile:
```
nl1-win2016:
  provider: nlxen
  image: windows_2016_web_template
  clone: True
  win_installer: /srv/salt/win/files/Salt-Minion-2018.3.4-Py3-AMD64-Setup.exe
  win_username: AdminUser
  win_password: 'AdminPassword'
  minion:
    master:
      - myMaster.my.domain.com
  ipv4_cidr: 1.1.1.1/24
  ipv4_gw: 1.1.1.254
  use_winrm: False
  winrm_verify_ssl: False
  winrm_port: 5986
  port_timeout: 30

```
### Steps to Reproduce Issue
```
[INFO    ] Connecting to SMB Tree \\1.1.1.1\ADMIN$
[INFO    ] Creating open to PAExec file with delete on close flags
[INFO    ] Disconnecting from SMB Tree \\1.1.1.1\ADMIN$
[DEBUG   ] psexec connected...
[DEBUG   ] SMB port 445 on 1.1.1.1 is available
[DEBUG   ] Logging into 1.1.1.1:445 as AdminUser
[INFO    ] Get connection smbprotocol
[WARNING ] Failed auth for mech NEGOEX (1.3.6.1.4.1.311.2.2.30): Mech Type 1.3.6.1.4.1.311.2.2.30 is not yet supported
[WARNING ] Failed auth for mech NEGOEX (1.3.6.1.4.1.311.2.2.30): Mech Type 1.3.6.1.4.1.311.2.2.30 is not yet supported
[DEBUG   ] Deleting PAExec service at the end of the process
[INFO    ] Connecting to SMB Tree \\1.1.1.1\ADMIN$
[INFO    ] Creating open to PAExec file with delete on close flags
[INFO    ] Disconnecting from SMB Tree \\1.1.1.1\ADMIN$
[DEBUG   ] _delete_directory_smbprotocol - share: C$, path: salttemp
[...]

PR #54911, opened 2019-10-07
Fix winexe and psexec bugs #54910 #54904

Mapel88:
minion fails to start after upgrade. start action is performed while minion is still shutting down.
- increased sleep time from 5 to 10 for winrm usage
- added while loop to verify minion stopped before starting it for psexec
- fixed run_winexe_command function

### What does this PR do?
Fix minion start issues

fix bugs #54910 #54904

### Previous Behavior
minion fails to start after upgrade. start action is performed while minion is still shutting down.

### New Behavior
wait for minion to stop / increase sleep - depends on the used module

### Tests written?
No

### Commits signed with GPG?
No

### DIFF
--- a/salt/utils/cloud.py
+++ b/salt/utils/cloud.py
@@ -1356,12 +1356,10 @@ def deploy_windows(host,
                 salt.utils.smb.delete_directory('salttemp', 'C$', conn=smb_conn)
         # Shell out to psexec to ensure salt-minion service started
         if use_winrm:
-            
             winrm_cmd(winrm_session, 'sc', ['stop', 'salt-minion'])
             time.sleep(10)
             winrm_cmd(winrm_session, 'sc', ['start', 'salt-minion'])
-        else:
-            
+        else:            
             stdout, stderr, ret_code = run_psexec_command(
                 'cmd.exe', '/c sc stop salt-minion', host, username, password
             )
@@ -1371,15 +1369,13 @@ def deploy_windows(host,
             stdout, stderr, ret_code = run_psexec_command(
                 'cmd.exe', '/c sc query salt-minion', host, username, password
             )
-            log.debug('%s', stdout) 
-            
+            log.debug('%s', stdout)
             retry = 0
             while 'STOP_PENDING' in stdout and retry < 100:
                 stdout, stderr, ret_code = run_psexec_command(
                     'cmd.exe', '/c sc query salt-minion', host, username, password
                 )
                 retry += 1
-            
             log.debug('Run psexec: sc start salt-minion')
             stdout, stderr, ret_code = run_psexec_command(
                 'cmd.exe', '/c sc start salt-minion', host, username, password
--- a/salt/utils/cloud.py
+++ b/salt/utils/cloud.py
@@ -1371,7 +1371,7 @@ def deploy_windows(host,
             stdout, stderr, ret_code = run_psexec_command(
                 'cmd.exe', '/c sc query salt-minion', host, username, password
             )
-            log.debug('%s',stdout) 
+            log.debug('%s', stdout) 
             
             retry = 0
             while 'STOP_PENDING' in stdout and retry < 100:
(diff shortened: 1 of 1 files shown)


salt/utils/cloud.py
```diff
@@ -935,7 +935,7 @@ def remove_service(self, wait_timeout=10, sleep_wait=1):
 def run_winexe_command(cmd, args, host, username, password, port=445):
     '''
     Run a command remotly via the winexe executable
-    '''
+    '''
```
dwoz: Pease remove the extra spaces after the doc string. The linter is probably complaining about this one.

salt/utils/cloud.py
```diff
@@ -1356,18 +1356,28 @@ def deploy_windows(host,
...
                 return False
 
-            time.sleep(5)
-
+            stdout, stderr, ret_code = run_psexec_command(
+                'cmd.exe', '/c sc query salt-minion', host, username, password
+            )
+            log.info('%s',stdout) 
+            
+            while "STOP_PENDING" in stdout:
```
dwoz: - This should be using single quotes.
- Should we also have a timeout here?
- Should we wait at the end of this loop?

salt/utils/cloud.py
```diff
@@ -1356,18 +1356,28 @@ def deploy_windows(host,
...
             )
             if ret_code != 0:
                 return False
 
-            time.sleep(5)
-
+            stdout, stderr, ret_code = run_psexec_command(
+                'cmd.exe', '/c sc query salt-minion', host, username, password
+            )
+            log.info('%s',stdout)
```
dwoz: I don't think we need this log statement.

salt/utils/cloud.py
```diff
@@ -1356,18 +1356,28 @@ def deploy_windows(host,
                 salt.utils.smb.delete_directory('salttemp', 'C$', conn=smb_conn)
         # Shell out to psexec to ensure salt-minion service started
         if use_winrm:
+            
             winrm_cmd(winrm_session, 'sc', ['stop', 'salt-minion'])
-            time.sleep(5)
+            time.sleep(10)
```
dwoz: Ideally we'd come up with a way to check and not need the sleep time, similar to what you are doing with the while loop below.

dwoz (changes requested):

salt/utils/cloud.py
Mapel88: OK

salt/utils/cloud.py
Mapel88: - Ok. Will fix it.
- You're right, I will add some timeout in case service fails to stop.
- No need to wait. The while loop does the waiting.

salt/utils/cloud.py
Mapel88: I will change to debug cause I think it's informative for debugging.

salt/utils/cloud.py
Mapel88: Ideally yes. But not really sure what winrm returns. Don't know how to handle the result / output, so just increased the timeout.

Mapel88:
@dwoz Hi,
Could you please help me figure out why the tests are failing?
I fixed all the lint errors I saw in the log, but still no success at committing.

Thanks

salt/utils/cloud.py
waynew: A continuation of my [other comments](https://github.com/saltstack/salt/pull/53547#issuecomment-552915073), one of these PRs should be pointed to master, and have tests applied.

If you don't have access to a windows environment to figure out what is returned here, I'm sure we can find someone who can get that information.

Mapel88: @waynew I figured it out...
It returns the status code. see:

```
def winrm_cmd(session, command, flags, **kwargs):
    '''
    Wrapper for commands to be run against Windows boxes using WinRM.
    '''
    log.debug('Executing WinRM command: %s %s', command, flags)
    r = session.run_cmd(command, flags)
    return r.status_code
```
Not really sure what to do with it.


What do you mean by "one of these PRs should be pointed to master"?
Should I change the branch and merge to master?

Thanks

PR #54929, opened 2019-10-08
Address file.managed with contents_newline flag

xeacott:
### What does this PR do?
Address the issue of the file.managed state not using contents_newline correctly.

Issue #54177

### Previous Behavior
```
            for part in validated_contents:
                for line in part.splitlines():
                    contents += line.rstrip('\n').rstrip('\r') + os.linesep
            if contents_newline and not contents.endswith(os.linesep):
                contents += os.linesep
```
Because of the for loop EOF would contain a platform-specific line-ending, even if the user set
contents_newline = False. According to the documentation, contents_newline should allow the user
to specify that they don't want the file to end with a newline.

### New Behavior
Setting `contents_newline = False` will now determine if the last entry in the file should end with
a newline character

### Tests written?

Yes

### Commits signed with GPG?

Yes


dwoz (approved):

tests/integration/states/test_file.py
```diff
@@ -573,6 +573,36 @@ def test_managed_contents(self):
...
+        test file.managed with contents by using the default content_newline
+        flag.
+        '''
+        contents = r'test_managed_contents_with_newline_one'
+        name = os.path.join(TMP, 'foo')
+
+        # Create a file named foo with contents as above but with a \n at EOF
+        self.run_state('file.managed', name=name, contents=contents,
+                       contents_newline=True)
+        with salt.utils.files.fopen(name, 'r') as fp_:
```
waynew: Wouldn't it be appropriate to open this in `rb` mode instead? 🤔

xeacott: Hmm, you aren't wrong! I must have forgotten the `b`

salt/states/file.py
```diff
@@ -2634,9 +2636,17 @@ def managed(name,
             contents = ''
             for part in validated_contents:
                 for line in part.splitlines():
-                    contents += line.rstrip('\n').rstrip('\r') + os.linesep
-            if contents_newline and not contents.endswith(os.linesep):
-                contents += os.linesep
+                    if not line == validated_contents[-1]:
```
DmitryKuzmenko: Some logical error here. Probably correct one is:
```python
if not line == part[-1]:
```
But it would be also wrong to compare the contents because the content could contains equal data, say
```
line
line
line
```
So you probably will want something like this:
```python
lines = part.splitlines()
for num, line in enumerate(lines):
    if not num == len(lines) - 1
```

DmitryKuzmenko (changes requested):

tests/integration/states/test_file.py
xeacott: Actually no I believe it should remain as r. I thought about this and the PR was incorrectly named (definitely my fault) but opening the file in binary mode and then reading it will produce different results based on the platform, i.e. Window's will read the contents differently than a linux platform. So I decided to just open the file in r as it's more in-line with the change in logic itself. Does that make sense?

dwoz:
@waynew @DmitryKuzmenko Are we good here?

DmitryKuzmenko (approved):

defanator:
@dwoz @DmitryKuzmenko which release will include this particular fix? TIA!

defanator:
@xeacott what is the nearest planned salt release with this fix included? TIA!

oeuftete:
@defanator This fix will be part of the upcoming Salt 3000 (Neon) release.

PR #51698, opened 2019-02-18
Nxos/input validation

mikewiebe:
### What does this PR do?

This PR adds input validation for a couple of the methods in `salt/modules/nxos.py` and updates the install doc to use `systemctl` to start the nxos guestshell minion.

### Previous Behavior
```
(py2-salt) n9kuser@n9kuser:~/Projects/salt$salt n9k nxos.cmd sendline show version
n9k:
    The minion function caused an exception: Traceback (most recent call last):
      File "/Users/n9kuser/Projects/salt/salt/metaproxy/proxy.py", line 409, in thread_return
        return_data = minion_instance.executors[fname](opts, data, func, args, kwargs)
      File "/Users/n9kuser/Projects/salt/salt/executors/direct_call.py", line 12, in execute
        return func(*args, **kwargs)
      File "/Users/n9kuser/Virtualenvs/py2-salt/var/cache/salt/proxy/extmods/n9k/modules/nxos.py", line 228, in cmd
        return __salt__[local_command](*args, **kwargs)
      File "/Users/n9kuser/Virtualenvs/py2-salt/var/cache/salt/proxy/extmods/n9k/modules/nxos.py", line 335, in sendline
        return __proxy__['nxos.sendline'](command, method, **kwargs)
      File "/Users/n9kuser/Virtualenvs/py2-salt/var/cache/salt/proxy/extmods/n9k/proxy/nxos.py", line 323, in sendline
        result = _nxapi_request(command, method, **kwargs)
      File "/Users/n9kuser/Virtualenvs/py2-salt/var/cache/salt/proxy/extmods/n9k/proxy/nxos.py", line 538, in _nxapi_request
        data = __utils__['nxos.nxapi_request'](commands, method=method, **conn_args)
      File "/Users/n9kuser/Virtualenvs/py2-salt/var/cache/salt/proxy/extmods/n9k/utils/nxos.py", line 313, in nxapi_request
        return client.request(method, commands)
      File "/Users/n9kuser/Virtualenvs/py2-salt/var/cache/salt/proxy/extmods/n9k/utils/nxos.py", line 199, in request
        return self.parse_response(response, command_list)
      File "/Users/n9kuser/Virtualenvs/py2-salt/var/cache/salt/proxy/extmods/n9k/utils/nxos.py", line 263, in parse_response
        raise NxosError('Unknown Error: {}, Code: {}'.format(msg, code))
    NxosError: Unknown Error: Message not supported, Code: 501

```
### New Behavior

```
[...]

PR #50588, opened 2018-11-20
New NX-OS Upgrade Execution and State Modules

mikewiebe:
### What does this PR do?

This PR adds a new execution and state module to upgrade/downgrade the software on NX-OS devices.

Note: This PR replaces https://github.com/saltstack/salt/pull/50395

Summary of changes:
  * Adds new execution module `salt/modules/nxos_upgrade.py`
  * Adds new state module `salt/states/nxos_upgrade.py`
  * Adds new `NxosError` exception classes

### Tests written?

Manual tests executed on the following platform types.

* N9k (disruptive and non-disruptive upgrades)
    * Tested both traditional ISSU (Dual Supervisors) and Enhanced ISSU (LXC boot mode)
    * SSH and NX-API Proxy Minion
    * Native GuestShell Minion
* N7k
    * SSH and NX-API Proxy Minion
* N5k
    * SSH Proxy Minion
* N3k
    * SSH and NX-API Proxy Minion
    * Native GuestShell Minion

### Commits signed with GPG?

Yes

PR #54931, opened 2019-10-08
merge nxos-specific work from develop into master.

dhiltonp:
### What does this PR do?
Combines commits from the following PRs into master (original merges were into develop).

https://github.com/saltstack/salt/pull/49496
https://github.com/saltstack/salt/pull/49676
https://github.com/saltstack/salt/pull/49769
https://github.com/saltstack/salt/pull/50588
https://github.com/saltstack/salt/pull/50575
https://github.com/saltstack/salt/pull/50748
https://github.com/saltstack/salt/pull/51059
https://github.com/saltstack/salt/pull/51698

### Tests written?

WIP/TBD.

I'm currently verifying test coverage etc.

### Commits signed with GPG?

Yes


mikewiebe:
@dhiltonp Do you know when this will be merged?

dhiltonp:
I'm still working on it, I hope to be done early next week.

mikewiebe:
@dhiltonp Any updates on this?

salt/proxy/nxos.py
```diff
@@ -402,7 +401,7 @@ def _init_ssh(opts):
         log.error('-- Verify that "feature ssh" is enabled on your NX-OS device: %s', opts['proxy']['host'])
         log.error('-- Exception Generated: %s', ex)
         log.error(ex)
-        exit()
+        raise
```
mikewiebe: @dhiltonp Curious about this change.  Without the `exit()` we will see the message but the proxy minion will remain in an initialization / connection loop trying over and over to establish a connection making the message really hard to read.

tests/unit/modules/test_nxos.py
```diff
@@ -0,0 +1,1024 @@
...
+    def test_config_nxos_error_ssh(self):
+
+        """ UT: nxos module:config method - nxos device error over ssh transport """
+
+        commands = ['feature bgp', 'router bgp 57']
+        kwargs = {}
+        config_result = [['feature bgp', 'router bgp 57'], u'bgp instance is already running; Tag is 55']
+        expected_output = 'COMMAND_LIST: feature bgp ; router bgp 57\nbgp instance is already running; Tag is 55\n'
+
+        with patch.object(nxos, 'show', MagicMock(return_value=initial_config[0])):
```
dhiltonp: we're patching 'show' twice, we should change this to be `side_effect = [initial_config[0], modified_config[0]]`

This applies to several of these tests.

salt/proxy/nxos.py
dhiltonp: With exit(), a ssh connection issue prevents the salt minion from starting up at all (I think). Am I wrong?

salt/proxy/nxos.py
```diff
@@ -317,7 +316,7 @@ def sendline(command, method='cli_show_ascii', **kwargs):
     return result
 
 
-def proxy_config(commands, **kwargs):
+def proxy_config(commands, no_save_config=None):
```
dhiltonp: `no_save_config=None`, then we negate it everywhere.

`save_config` defaulting to True seems to make more sense logically. 

Do you see any issues with us flipping the logic?

sagetherage:
@mikewiebe looks like there are merge conflicts to be addressed, please have all the changes been addressed? I can have it reviewed again once you have completed both those. Thank you!

mikewiebe:
> @mikewiebe looks like there are merge conflicts to be addressed, please have all the changes been addressed? I can have it reviewed again once you have completed both those. Thank you!

Hi @sagetherage.  I don't own this PR.  @dhiltonp owns it.  Thanks!

tests/unit/modules/test_nxos.py
dhiltonp: @mikewiebe, could you make this change?

salt/proxy/nxos.py
dhiltonp: @mikewiebe, do you see any issues with flipping the logic?

dhiltonp: To be clear, I'm happy to do the flipping, and am just asking if you see problems with the flip.

dhiltonp:
I do own this PR, don't worry about keeping it up to date with master. I do have a few questions that I've been waiting for input on, and I've tagged you in them to make sure they're easy to find.

dhiltonp:
If you'd like, we can certainly have a quick video chat about them, too!

salt/proxy/nxos.py
mikewiebe: No, you are not wrong but that was the intent.  We did not want the minion to start up if there was an ssh connection issue but exit with guidance about how to fix the problem and then restart the minion.

tests/unit/modules/test_nxos.py
mikewiebe: Yes, I can make this change.

salt/proxy/nxos.py
mikewiebe: I am fine with the flip.

salt/proxy/nxos.py
dhiltonp: `exit()` doesn't return a special error code or anything. We'd need to at least change this to be `exit(1)`, but raise is much preferred by the team.

salt/modules/nxos.py
```diff
@@ -1,47 +1,208 @@
...
+                       '6': 'sha512', }
+    password_line = get_user(username, **kwargs)
+    if not password_line:
+        return None
+    if '!' in password_line:
+        return False
+    cur_hash = re.search(r'(\$[0-6](?:\$[^$ ]+)+)', password_line).group(0)
+    if encrypted is False:
+        hash_type, cur_salt, hashed_pass = re.search(r'^\$([0-6])\$([^$]+)\$(.*)$', cur_hash).groups()
+        new_hash = gen_hash(crypt_salt=cur_salt, password=password, algorithm=hash_algorithms[hash_type])
```
dhiltonp, line 174: gen_hash uses python's `crypt.crypt`, which says:

> This module implements an interface to the crypt(3) routine, which is a one-way hash function based upon a modified DES algorithm; see the Unix man page for further details. Possible uses include storing hashed passwords so you can check passwords without storing the actual password, or attempting to crack Unix passwords with a dictionary.
>
> Notice that the behavior of this module depends on the actual implementation of the crypt(3) routine in the running system. Therefore, any extensions available on the current implementation will also be available on this module.

[Looking on wikipedia, I see this](https://en.wikipedia.org/wiki/Crypt_(C)#Support_in_operating_systems):
image

I need to modify `gen_hash` so that it reflects the crypt methods available on the current system.

I'll also need to modify `set_password` to make sure that the hash method requested by the user is available on NXOS. I would prefer to raise an error reporting the problem to the user rather than silently setting a hash not usable on NXOS or lowering the hash method to the lowest common denominator.

I did a quick search, and it appears that base OS for NXOS is linux, but your list of hash_algorithms includes 2a - which isn't available on my Linux test box.

@mikewiebe, can you please verify what password hash algorithms are supported on nxos? If python can run, [`crypt.methods` will show the list of supported algorithms](https://docs.python.org/3/library/crypt.html#crypt.methods).

dhiltonp: I've done the gen_hash fixes and am planning on restricting the usable hashes to those normally available on linux.

mikewiebe: @dhiltonp Here is the list of supported algorithms on nxos.

```python
>>> import crypt
>>> crypt.methods
[<crypt.METHOD_SHA512>, <crypt.METHOD_SHA256>, <crypt.METHOD_MD5>, <crypt.METHOD_CRYPT>]
>>>
```

mikewiebe: @dhiltonp I should point out that much of this functionality existed before my changes to combine (ssh and nxapi) proxy minion and nxos guestshell native minion workflows.  I simply moved this code from the proxy source code file into the module source code file and wanted to preserve existing functionality since I did not add it orginally. 

See: https://github.com/saltstack/salt/pull/49496/files#diff-a9bea6509dcf922bcddad8022f67a474L257

dhiltonp: Thanks!

salt/proxy/nxos.py
dhiltonp: done

salt/proxy/nxos.py
dhiltonp: @mikewiebe I've made just about all the changes I know should be made, could you do some manual/integration tests against nxos.

tests/unit/modules/test_nxos.py
dhiltonp: (these are the currently failing tests)

salt/modules/nxos.py
```diff
@@ -1,49 +1,217 @@
 # -*- coding: utf-8 -*-
 """
-Execution module for Cisco NX OS Switches Proxy minions
+Execution module for Cisco NX OS Switches.
 
 .. versionadded:: 2016.11.0
 
-For documentation on setting up the nxos proxy minion look in the documentation
-for :mod:`salt.proxy.nxos <salt.proxy.nxos>`.
+This module supports execution using a Proxy Minion or Native Minion:
```
dhiltonp, line 7: We have an issue with docs here. 

I did look in the official sphinx docs for examples at a file level and didn't find anything that seemed applicable here.

waynew: What's the docs issue?

dhiltonp: From the build: `/var/jenkins/workspace/pr-docs_PR-54931/salt/modules/nxos.py:docstring of salt.modules.nxos:7:Unexpected indentation.`

bryceml: ```/var/jenkins/workspace/pr-docs_PR-54931/salt/modules/nxos.py:docstring of salt.modules.nxos:7:Unexpected indentation.```
it's line 7 of the docstring I'm pretty sure, not the file, so who knows which doc string it's from.

salt/modules/nxos.py
```diff
@@ -57,11 +225,636 @@ def cmd(command, *args, **kwargs):
...
+
+def sendline(command, method="cli_show_ascii", **kwargs):
+    """
+    Send arbitray commands to the NX-OS device.
+
+    command
+        The command to be sent.
+
+    method:
+        ``cli_show_ascii``: Return raw test or unstructured output.
```
bryceml, line 329: is this supposed to be indented 3 maybe?  I think 3 spaces might be normal for sphinx, not sure though.

mikewiebe:
@dhiltonp I have done quite a bit of manual testing today with our nxos lab devices and for the most part thinks are looking ok.  I did discover a couple of odd failures though that I would like to discuss with you.  Do you have time to jump on a call?

dhiltonp:
Sure, do you want to do it right now?

salt/modules/nxos.py
bryceml:
```
/code/salt/modules/nxos.py:docstring of salt.modules.nxos:9: WARNING: Unexpected indentation.
/code/salt/modules/nxos.py:docstring of salt.modules.nxos:56: WARNING: Definition list ends without a blank line; unexpected unindent.
/code/salt/proxy/nxos.py:docstring of salt.proxy.nxos:63: WARNING: Definition list ends without a blank line; unexpected unindent.
/code/doc/topics/installation/nxos.rst:154: WARNING: Title underline too short.
```

Is the warnings I get when running without `-W` so you will likely hit those too

mikewiebe:
> Sure, do you want to do it right now?

Sure, now works.

salt/modules/nxos.py
bryceml:
```
diff --git a/doc/topics/installation/nxos.rst b/doc/topics/installation/nxos.rst
index a753c051ef..91c7bfecf3 100644
--- a/doc/topics/installation/nxos.rst
+++ b/doc/topics/installation/nxos.rst
@@ -151,7 +151,7 @@ GuestShell Salt Minion Installation
 This section is only required when running the SaltStack Minion from the ``guestshell``.
 
 STEP 1a: Enable the Guestshell on low footprint N3ks
----------------------------------------------------
+----------------------------------------------------
 
 **NOTE:** Skip down to **STEP 1b** if the target system is not a low footprint N3k.
 
diff --git a/salt/modules/nxos.py b/salt/modules/nxos.py
index 293508fcdc..8770004f5e 100644
--- a/salt/modules/nxos.py
+++ b/salt/modules/nxos.py
@@ -5,9 +5,9 @@ Execution module for Cisco NX OS Switches.
 .. versionadded:: 2016.11.0
 
 This module supports execution using a Proxy Minion or Native Minion:
-1) Proxy Minion: Connect over SSH or NX-API HTTP(S).
+   1) Proxy Minion: Connect over SSH or NX-API HTTP(S).
    See :mod:`salt.proxy.nxos <salt.proxy.nxos>` for proxy minion setup details.
-2) Native Minion: Connect over NX-API Unix Domain Socket (UDS).
+   2) Native Minion: Connect over NX-API Unix Domain Socket (UDS).
    Install the minion inside the GuestShell running on the NX-OS device.
 
 :maturity:   new
@@ -54,7 +54,7 @@ cookie
 
 save_config:
     If True, 'copy running-config starting-config' is issues for every
-        configuration command.
+    configuration command.
     If False, Running config is not saved to startup config
     Default: True
 
diff --git a/salt/proxy/nxos.py b/salt/proxy/nxos.py
index fc62bce7e2..3bd921273d 100644
--- a/salt/proxy/nxos.py
+++ b/salt/proxy/nxos.py
@@ -61,7 +61,7 @@ password:
 
 save_config:
     If True, 'copy running-config starting-config' is issues for every
-        configuration command.
+    configuration command.
     If False, Running config is not saved to startup config
     Default: True
 
```
seems to fix it.

salt/modules/nxos.py
```diff
@@ -1,49 +1,217 @@
...
     """
-    return cmd("system_info")
+    return role in get_roles(username, **kwargs)
 
 
 def cmd(command, *args, **kwargs):
     """
-    run commands from __proxy__
-    :mod:`salt.proxy.nxos<salt.proxy.nxos>`
+    NOTE: This function is preserved for backwards compatibilty.  This allows
```
dhiltonp, line 204: @mikewiebe: which is the preferred form? It seems that `nxos.function` is preferred, but most of the documentation for the commands says to use `nxos.cmd <function>`?

salt/modules/nxos.py
dhiltonp: Thanks, that did it!

salt/modules/nxos.py
```diff
@@ -57,11 +225,636 @@ def cmd(command, *args, **kwargs):
...
+        Whether to return raw text or structured data.
+        NOTE: raw_text option is ignored for SSH proxy minion.  Data is
+        returned unstructured.
+
+    CLI Example:
+
+    .. code-block:: bash
+
+        salt-call --local nxos.show 'show version'
+        salt '*' nxos.show 'show bgp sessions ; show processes' raw_text=False
```
dhiltonp, line 379: The repeating "show" seems weird. 

Also, the default of raw_text is the same default as for sendline, so the flag seems less important?

salt/proxy/nxos.py
```diff
@@ -120,462 +195,352 @@ def __virtual__():
...
+            result = _nxapi_request(command, method, **kwargs)
+    except (TerminalException, NxosCliError) as e:
         log.error(e)
-        return "Failed to set password"
+        raise
+    return result
 
 
-def unset_role(username, role):
+def proxy_config(commands, save_config=None):
```
dhiltonp, line 320: Why are we calling this proxy_config instead of config?

To be consistent, we would also call the proxy.sendline proxy_sendline, etc.

salt/modules/nxos.py
mikewiebe: @dhiltonp `nxos.function` is definitely preferred.  I am updating the docs now.

salt/modules/nxos.py
mikewiebe: This is legacy behavior.  See:

https://github.com/saltstack/salt/blob/master/salt/modules/nxos_api.py#L300

salt/proxy/nxos.py
mikewiebe: Ok, I know where some of the confusion is coming from.  The only functions that are really callable are the functions defined in the execution module `salt/modules/nxos.py`.  The functions in `salt/proxy/nxos.py` are called by functions in `salt/modules/nxos.py` when a proxy is needed.

That means that `proxy_config` is not directly callable.

```shell
(salt_py375) ➜  salt_py375 salt -c ./etc/salt 'n9k-proxy' nxos.p
n9k-proxy:
    ----------
    nxos.p:
        'nxos.p' is not available.
    nxos.ping:

            Ping the device on the other end of the connection.

            .. code-block: bash

                salt '*' nxos.ping

ERROR: Minions returned with non-zero exit code
(salt_py375) ➜  salt_py375 salt -c ./etc/salt 'n9k-proxy' nxos.proxy_config
n9k-proxy:
    'nxos.proxy_config' is not available.
ERROR: Minions returned with non-zero exit code
(salt_py375) ➜  salt_py375
```

Before my refactoring most of the functions were in the proxy source file and where directly callable but to streamline the workflows for all minion types I refactored the execution flow to always go through the execution module.  In order to `expose` a function in the proxy source file we need a corresponding function in the execution module source file.

From my perspective we need to do the following.

1. Remove function level documentation in the proxy source file and/or add a note to indicate that functions in the proxy file are documented in the module source file.
2. Add a function in the module source file for the `shutdown` function and return a message when `_shutdown_nxapi` is called to indicate that shutting down the nxapi minion connection is not possible.

mikewiebe: **Update:** I decided to remove the `shutdown` function because it does not provide any value.

mikewiebe: Never mind.  Turns out that a `shutdown` function is required but since it's not adding value I turned them into no-ops.

```
The Salt ProxyMinion is shutdown.
Proxymodule nxos is missing an init() or a shutdown() or both. Check your proxymodule.  Salt-proxy aborted.
```

salt/modules/nxos.py
dhiltonp: Should we modify salt/modules/nxos_api.py to call the functions in modules/nxos?

dhiltonp:
Maybe we should mark some of the older API for deprecation?

`cmd`, `system_info`, `add_config` and `show` seem potential candidates.

```
salt.utils.versions.warn_until("Silicon", "nxos.cmd COMMAND is deprecated in favor of nxos.COMMAND")
```

We're on Neon, next up are Sodium, Magnesium, Aluminium, then Silicon.

salt/modules/nxos.py
mikewiebe: I don't really feel comfortable modifying the functions in `nxos_api` and also believe that work is beyond the scope of what we are doing here.  I would prefer to mark the entire `nxos_api` module for deprecation if that is possible.

dhiltonp: Sure, we can do that :)

salt/proxy/nxos.py
dhiltonp: I guess my confusion with the name is because every other function in proxy/nxos has the same name as the equivalent in modules/nxos (like `proxy.sendline`), while this is named `proxy.proxy_config`.

dhiltonp: I believe the shutdown function is called when shutting down a proxy minion. It seems that explicitly shutting down the connection may be better... You didn't notice any issues when you restarted the ssh proxy minion, I guess?

https://docs.saltstack.com/en/latest/topics/proxyminion/index.html#proxymodules

mikewiebe: > I believe the shutdown function is called when shutting down a proxy minion. It seems that explicitly shutting down the connection may be better... You didn't notice any issues when you restarted the ssh proxy minion, I guess?
> 
> https://docs.saltstack.com/en/latest/topics/proxyminion/index.html#proxymodules

What exactly do you mean by restart?  I was able to stop and start the minion without any issues.  I also see the following in the guide you sent.

```
shutdown(): Code to cleanly shut down or close a connection to a controlled device goes here. This function must exist, but can contain only the keyword pass if there is no shutdown logic required.
```

I could add the code back that we had for the ssh proxy minion since it's not hurting anything just to be safe but just use `pass` for the nxapi proxy minion.

mikewiebe: > I guess my confusion with the name is because every other function in proxy/nxos has the same name as the equivalent in modules/nxos (like `proxy.sendline`), while this is named `proxy.proxy_config`.

Just took another look at the execution module functions and I believe we were trying to avoid confusion because of the following.

Functions in `modules/nxos` call `_configure_device` which in turn calls `proxy.proxy_config` if a proxy is needed.  This is also true for the function in `modules/nxos` that is named `config`.  I think we were trying to make it clear that `config` and `proxy_config` are different.  I suppose we could still rename `proxy_config` in `proxy/nxos` to be `config` and there would not be a conflict but it might be more confusing?

mikewiebe:
> Maybe we should mark some of the older API for deprecation?
> 
> `cmd`, `system_info`, `add_config` and `show` seem potential candidates.
> 
> https://docs.saltstack.com/en/latest/topics/development/deprecations.html
> 
> ```
> salt.utils.versions.warn_until("Silicon", "nxos.cmd COMMAND is deprecated in favor of nxos.COMMAND")
> ```
> 
> We're on Neon, next up are Sodium, Magnesium, Aluminium, then Silicon.

I am fine with adding this for `cmd` and `show` but not sure I would add them for `system_info` and `add_config`.

dhiltonp:
Ok! I'm not sure I completely understand the design.

It looks to me like modules/nxos.py's `system_info` returns the exact same thing as `grains`. Should one be deprecated, or is there some reason to keep them separate and active?

`add_config` just calls `config`, but cannot take a file.

`template_engine`, `context` and `defaults` are only documented as being applicable to `config`, but `add_config` will accept them and merrily pass them along to `config`, where they will behave just as if `config` had been called...

It seems like these functions don't really add much - but maybe you see something I don't?

mikewiebe:
> Ok! I'm not sure I completely understand the design.
> 
> It looks to me like modules/nxos.py's `system_info` returns the exact same thing as `grains`. Should one be deprecated, or is there some reason to keep them separate and active?
> 
> `add_config` just calls `config`, but cannot take a file.
> 
> `template_engine`, `context` and `defaults` are only documented as being applicable to `config`, but `add_config` will accept them and merrily pass them along to `config`, where they will behave just as if `config` had been called...
> 
> It seems like these functions don't really add much - but maybe you see something I don't?

I completely agree with your assessment. They don't add much value at all.  The only reason to keep them is because they are part of a released version of SaltStack and I don't know what impact removing them would have for existing customers (if any) that are using them.  I suppose that adding the deprecation statement would give them more than enough time to adjust however.

dhiltonp:
@mikewiebe: I'd like to get on the same page about kwargs and where to use them.

They are very convenient when you are passing arguments through a long chain, or doing some sort of function dispatch (like `nxos.cmd COMMAND args`).

However, the last function (one that doesn't call another function with with all remaining kwargs) shouldn't use kwargs:

```python
def external(**kwargs):
    intermediate(**kwargs)
    ...

def intermediate(my_arg=None, **kwargs):
    last(**kwargs)
    ...

def last(arg1, arg2=None):
    ...
```

The 'last' function might be the function that passes args to 2 other functions:

```python
def external(**kwargs):
    intermediate(**kwargs)
    ...

def intermediate(my_arg=None, **kwargs):
    last(**kwargs)
    ...

def last(arg1, arg2=None):
    foo(arg1)
    bar(arg2)
```

We want *all* arguments to be consumed, and if an argument isn't used, it should be an error.

This protects the user against typos or compatibility issues where they haven't noticed an argument being deprecated.

dhiltonp:
It looks like we have an inconsistency between `modules/nxos` `show` and `sendline` functions.

Ultimately, `SSHConnection.sendline` knows how to handle one command, while `NxapiClient.request` joins commands (`'; '.join(commands)`).

I suggest we subclass `SSHConnection` and update all the nxos sendline functions and docs to report they can take multiple commands.

salt/modules/nxos.py
dhiltonp: `napalm` also uses `nxos_api`, should we update it to use the equivalent `nxos` functions?

mikewiebe: @dhiltonp I was not aware of the fact that `napalm` was using `nxos_api`.  In that case maybe we don't want to deprecate `nxos_api`.  I do believe that `napalm` should be updated at some point to use the `nxos` functions but that should involve the `napalm` folks and is beyond the scope of this PR.

mikewiebe:
> @mikewiebe: I'd like to get on the same page about kwargs and where to use them.
> 
> They are very convenient when you are passing arguments through a long chain, or doing some sort of function dispatch (like `nxos.cmd COMMAND args`).
> 
> However, the last function (one that doesn't call another function with with all remaining kwargs) shouldn't use kwargs:
> 
> ```python
> def external(**kwargs):
>     intermediate(**kwargs)
>     ...
> 
> def intermediate(my_arg=None, **kwargs):
>     last(**kwargs)
>     ...
> 
> def last(arg1, arg2=None):
>     ...
> ```
> 
> The 'last' function might be a function that passes args to 2 other functions:
> 
> ```python
> def external(**kwargs):
>     intermediate(**kwargs)
>     ...
> 
> def intermediate(my_arg=None, **kwargs):
>     last(**kwargs)
>     ...
> 
> def last(arg1, arg2=None):
>     foo(arg1)
>     bar(arg2)
> ```
> 
> We want _all_ arguments to be consumed, and if an argument isn't used, it should be an error.
> 
> This protects the user against typos or compatibility issues where they haven't noticed an argument being deprecated.

@dhiltonp I went back through the various execution module and proxy functions.  I believe the only 3 functions that apply are the following 3 in the proxy source file.

* `def grains, def grains_refresh, def _parse_output_for_errors`
    * https://github.com/dhiltonp/salt/blob/with-nxos/salt/proxy/nxos.py#L245
    * https://github.com/dhiltonp/salt/blob/with-nxos/salt/proxy/nxos.py#L260
    * https://github.com/dhiltonp/salt/blob/with-nxos/salt/proxy/nxos.py#L410

I already explained in the PR that you merged today why I added it back for `_parse_output_for_errors` but I can probably safely remove them for `grains` and `grains_refresh`.

mikewiebe:
> It looks like we have an inconsistency between `modules/nxos` `show` and `sendline` functions.
> 
> Ultimately, `SSHConnection.sendline` knows how to handle one command, while `NxapiClient.request` joins commands (`'; '.join(commands)`).
> 
> I suggest we subclass `SSHConnection` and update all the nxos sendline functions and docs to report they can take multiple commands.

@dhiltonp Not sure I follow.  The `sendline` function can handle multiple commands for both the ssh and nxapi proxy minion. Although it's not obvious, the following call to will handle multiple `;` separated commands.

https://github.com/dhiltonp/salt/blob/with-nxos/salt/proxy/nxos.py#L402

For example.  The following calls will process both the `show version` and `show inventory` commands and return the output of each for the SSH proxy minion.

```
salt -c ./etc/salt 'n7k-proxy' nxos.show 'show version ; show inventory'
salt -c ./etc/salt 'n7k-proxy' nxos.sendline 'show version ; show inventory'
```

The same behavior holds true for the nxapi proxy minion.

dhiltonp:
Correct, this does reliably work:

`salt -c ./etc/salt 'n7k-proxy' nxos.sendline 'show version ; show inventory'`

The issue comes from state files. The following will work if the minion is using NxapiClient, but not if it is using SSH:

```yaml
versinv:
  module.run:
    - nxos.sendline:
      - command:
        - show version
        - show inventory
```

dhiltonp:
I've handled `grains` and `grains_refresh` in `proxy/nxos.py`, take a look.

I've modified the SSH sendline to work with a string or list like NxapiClient, but I can easily roll that back if I'm in left field.

I'm sorry, I've forgotten the exact details of our discussion of `_parse_output_for_errors`, could you please refresh my memory? 
I know we had a bunch of excess arguments, but we're now using `clean_kwargs`, which fixes that, right? I do recall there being an argument that wasn't used for anything but I can't recall what it was or why it was passed to this function...

salt/proxy/nxos.py
dhiltonp: Let's leave it as is for now :)

dhiltonp:
Ah, the reason for the SSH vs. NxapiClient difference is because we're merging show and sendline.

Another fix might be to make `sendline` reject `command` if it isn't a string. People should mainly hit the bug when migrating from `show` to `sendline`.

I looked in to how NxapiClient handles the list of commands, and it's a bit more than just merging the command list and running it - every command in the list has separate error handling.

dhiltonp:
I'm not quite sure when this came about, but the current PR fails because nxos_upgrade docs are missing, we'll need to have something there too.

mikewiebe:
> Correct, this does reliably work:
> 
> `salt -c ./etc/salt 'n7k-proxy' nxos.sendline 'show version ; show inventory'`
> 
> The issue comes from state files. The following will work if the minion is using NxapiClient, but not if it is using SSH:
> 
> ```yaml
> versinv:
>   module.run:
>     - nxos.sendline:
>       - command:
>         - show version
>         - show inventory
> ```

Ah, I see.  Looks like you made changes for this to work so I will test them out.

mikewiebe:
> I've handled `grains` and `grains_refresh` in `proxy/nxos.py`, take a look.
> 
> I've modified the SSH sendline to work with a string or list like NxapiClient, but I can easily roll that back if I'm in left field.
> 
> I'm sorry, I've forgotten the exact details of our discussion of `_parse_output_for_errors`, could you please refresh my memory?
> I know we had a bunch of excess arguments, but we're now using `clean_kwargs`, which fixes that, right? I do recall there being an argument that wasn't used for anything but I can't recall what it was or why it was passed to this function...

I had opened up the following PR to handle grains a few days ago and add a minor doc update.
* https://github.com/saltstack/salt/pull/54931#issuecomment-617393023

Since you already handled the grains change I will remove them from my PR diff.

mikewiebe:
> I'm not quite sure when this came about, but the current PR fails because nxos_upgrade docs are missing, we'll need to have something there too.

Yeah, true.  I will add these!

dwoz (approved):

PR #54935, opened 2019-10-09
[master] Add missing 'fun' for returned events from "wfunc" SSH executions

meaksh:
### What does this PR do?
This PR fixes an issue when executing `wfuncs`, for example, `state.show_highstate` over `salt-ssh`.

While the `state.show_highstate` response is successfully received, the event that is fired to the "EventBus" is missing the `fun` attribute.

This PR will add the `fun` to the fired event in case this is missing.

### Tests written?

Yes

### Commits signed with GPG?

Yes


meaksh:
I'll work on the tests ASAP

salt/client/ssh/__init__.py
```diff
@@ -797,6 +799,8 @@ def run(self, jid=None):
                     data = {'return': data}
                 if 'id' not in data:
                     data['id'] = id_
+                if 'fun' not in data:
+                    data['fun'] = fun
```
isbm: This can be just written as:

```python
data.setdefault("fun", fun)
```

But I am still not sure this is the right place of doing it. In `handle_ssh` the function might not appear due to various reasons there as well as the next element is just a text, so the data was just created above in line 682. In that case I would rather find out why `fun` key did not appeared at all and if it is a bug, then should be done in the mentioned method instead.

Maybe @thatch45 might help here.

thatch45: This looks right since this data is only part of the event it should not interfere with any other aspect of execution. Using  data.setdefault("fun", fun) would also be an acceptable approach but it is not something that I am too concerned with here.

isbm: My main concern is why _sometimes_ "fun" and "Id" in this case needs to be checked. Shouldn't we rather move that above where "data" variable is redefined from string type to dict instead? I mean:

```python
if isinstance(...
    data = {"id": ...., "fun":...
```
Instead of double-checking.

meaksh:
@dwoz added missing unit tests

Ch3LL (dismissed):

Ch3LL:
@meaksh looks like this needs a rebase and theres some failing test although not certain they are related to your change.

meaksh:
@Ch3LL I've updated the branch with latest changes from `master`. Let see how tests are looking now :+1:

Ch3LL:
looks like pre-commit is failing and that should do it :)

meaksh:
@Ch3LL done! :+1: 

Now it looks correct apart from the `ci/lint` failure which I think it's spurious in this case.

Ch3LL (approved):

dwoz (approved):

Issue #54941, opened 2019-10-09
Pillar data is refreshed for EVERY salt command in 2019.2.1 and 2019.2.2


UtahDave:
### Description of Issue
In Salt 2019.2.1 and 2019.2.2 the pillar top file and all pillar data are re-rendered for every single salt command that is run.  Including things like `salt \* test.ping`

### Setup
Install Salt 2019.2.1 or 2019.2.2.  Set the master log to debug and tail /var/log/salt/master
Create a `/srv/pillar/top.sls` and add a pillar sls file so that your minion will be assigned pillar data.

### Steps to Reproduce Issue
Run `salt \* saltutil.refresh_pillar`
Run `salt \* pillar.items`
Run `salt \* test.ping`

You'll see that the pillar top file and pillar data are rendered for every command, when pillar data should only be rendered with refresh_pillar, highstate and the other proscribed times.

```



2019-10-09 22:52:14,844 [salt.config      :2185][DEBUG   ][2682] Reading configuration from /etc/salt/master
2019-10-09 22:52:14,846 [salt.config      :3666][DEBUG   ][2682] Using cached minion ID from /etc/salt/minion_id: xxPing
2019-10-09 22:52:14,848 [salt.config      :2289][DEBUG   ][2682] Missing configuration file: /root/.saltrc
2019-10-09 22:52:14,851 [salt.utils.event :323 ][DEBUG   ][2682] MasterEvent PUB socket URI: /var/run/salt/master/master_event_pub.ipc
2019-10-09 22:52:14,851 [salt.utils.event :324 ][DEBUG   ][2682] MasterEvent PULL socket URI: /var/run/salt/master/master_event_pull.ipc
2019-10-09 22:52:14,902 [salt.transport.zeromq:138 ][DEBUG   ][2682] Initializing new AsyncZeroMQReqChannel for (u'/etc/salt/pki/master', u'xxPing_master', u'tcp://127.0.0.1:4506', u'clear')
2019-10-09 22:52:14,902 [salt.transport.zeromq:209 ][DEBUG   ][2682] Connecting the Minion to the Master URI (for the return server): tcp://127.0.0.1:4506
2019-10-09 22:52:14,903 [salt.transport.zeromq:1189][DEBUG   ][2682] Trying to connect to: tcp://127.0.0.1:4506
[...]

PR #54942, opened 2019-10-09
Fix for 54941 pillar_refresh regression

dwoz:
### What does this PR do?

Only refresh pillar data when `BaseMinion.gen_modules` is called for the first time. Subsequent pillar refreshes are done when the `pillar_refresh` event if fired. 

#54941 

### Tests written?

Yes

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/minion.py
+++ b/salt/minion.py
@@ -433,13 +433,14 @@ class MinionBase(object):
 
             salt '*' sys.reload_modules
         '''
-        self.opts['pillar'] = salt.pillar.get_pillar(
-            self.opts,
-            self.opts['grains'],
-            self.opts['id'],
-            self.opts['saltenv'],
-            pillarenv=self.opts.get('pillarenv'),
-        ).compile_pillar()
+        if initial_load:
+            self.opts['pillar'] = salt.pillar.get_pillar(
+                self.opts,
+                self.opts['grains'],
+                self.opts['id'],
+                self.opts['saltenv'],
+                pillarenv=self.opts.get('pillarenv'),
+            ).compile_pillar()
 
         self.utils = salt.loader.utils(self.opts)
         self.functions = salt.loader.minion_mods(self.opts, utils=self.utils)
--- a/salt/minion.py
+++ b/salt/minion.py
@@ -2206,6 +2206,8 @@ class Minion(MinionBase):
         self.module_refresh(force_refresh)
         self.matchers_refresh()
         self.beacons_refresh()
+        evt = salt.utils.event.get_event('minion', opts=self.opts)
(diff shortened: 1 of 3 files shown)


s0undt3ch (approved):

Ch3LL (approved):

UtahDave (changes requested):
There's one more scenario I think I'd love to see tested. I'm not sure exactly where you would add this, but it would be great if in one of these spots you add this workflow:

1. Add a key/value pair to the pillar sls file
2. Execute a `test.ping`
3. Test that the pillar data has NOT been updated with the new key/value pair yet.
4. Run a `saltutil.refresh_pillar`
5. Then verify that the pillar data has been updated.

The reason for this is it would be great to make sure that a full pillar refresh isn't happening during a regular salt command.

Is this scenario already covered by one of these tests you already have?

dwoz:
> There's one more scenario I think I'd love to see tested. I'm not sure exactly where you would add this, but it would be great if in one of these spots you add this workflow:
> 
> 1. Add a key/value pair to the pillar sls file
> 2. Execute a `test.ping`
> 3. Test that the pillar data has NOT been updated with the new key/value pair yet.
> 4. Run a `saltutil.refresh_pillar`
> 5. Then verify that the pillar data has been updated.
> 
> The reason for this is it would be great to make sure that a full pillar refresh isn't happening during a regular salt command.
> 
> Is this scenario already covered by one of these tests you already have?

I thought about adding this exact test because it most closely mirrors the way we found the regression. The only reason I didn't is that we're essentially using `pillar.get`, `pillar.item`, and `pillar.raw` calls in place of `test.ping`.

UtahDave:
> > There's one more scenario I think I'd love to see tested. I'm not sure exactly where you would add this, but it would be great if in one of these spots you add this workflow:
> > 
> > 1. Add a key/value pair to the pillar sls file
> > 2. Execute a `test.ping`
> > 3. Test that the pillar data has NOT been updated with the new key/value pair yet.
> > 4. Run a `saltutil.refresh_pillar`
> > 5. Then verify that the pillar data has been updated.
> > 
> > The reason for this is it would be great to make sure that a full pillar refresh isn't happening during a regular salt command.
> > Is this scenario already covered by one of these tests you already have?
> 
> I thought about adding this exact test because it most closely mirrors the way we found the regression. The only reason I didn't is that we're essentially using `pillar.get`, `pillar.item`, and `pillar.raw` calls in place of `test.ping`.

OK, valid point.  You are correct. Those calls would trigger the bad behavior, if present.

UtahDave (approved):

waynew (approved):

PR #54956, opened 2019-10-11
(Backport 50302) cmd: Add root parameter for wait and run states

aplanas:
### What does this PR do?

`run_chroot` module.

### Tests written?

Yes. Were missing in develop, added for master

(backport #50302, already merged in develop)

### DIFF
--- a/salt/states/cmd.py
+++ b/salt/states/cmd.py
@@ -402,6 +402,7 @@ def wait(name,
          unless=None,
          creates=None,
          cwd=None,
+         root=None,
          runas=None,
          shell=None,
          env=(),
@@ -435,6 +436,10 @@ def wait(name,
     cwd
         The current working directory to execute the command in, defaults to
         /root
+
+    root
+        Path to the root of the jail to use. If this parameter is set, the command
+        will run inside a chroot
 
     runas
         The user name to run the command as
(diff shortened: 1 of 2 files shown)


salt/states/cmd.py
```diff
@@ -912,10 +923,11 @@ def run(name,
...
     # Wow, we passed the test, run this sucker!
     try:
-        cmd_all = __salt__['cmd.run_all'](
-            name, timeout=timeout, python_shell=True, **cmd_kwargs
+        run_cmd = 'cmd.run_all' if not root else 'cmd.run_chroot'
+        cmd_all = __salt__[run_cmd](
+            cmd=name, timeout=timeout, python_shell=True, **cmd_kwargs
         )
-    except CommandExecutionError as err:
+    except Exception as err:
```
waynew: Should we be catching all `Exceptions` here? 🤔

waynew (commented):
Hey @aplanas thanks for the PR and tests! Looks good just the one question about exception types.

salt/states/cmd.py
aplanas: Yes, this comes from the review of the original PR:

https://github.com/saltstack/salt/pull/50302#discussion_r229833477

The logic here is that we are making a call to **any** salt module (e.g mkdir, file.managed, etc), so we can have any kind of exception, including some declared by the user (via a _module)

dwoz (approved):

PR #54960, opened 2019-10-11
(Backport 50452, 54022) Add new module freezer

aplanas:
### What does this PR do?

### Tests written?

Yes

(backport #50452, already merged in develop)
(backport #54022, under review)

### DIFF
--- a/salt/modules/freezer.py
+++ b/salt/modules/freezer.py
@@ -164,13 +164,13 @@ def freeze(name=None, force=False, **kwargs):
     safe_kwargs = clean_kwargs(**kwargs)
     pkgs = __salt__['pkg.list_pkgs'](**safe_kwargs)
     repos = __salt__['pkg.list_repos'](**safe_kwargs)
-    for name, content in zip(_paths(name), (pkgs, repos)):
-        with fopen(name, 'w') as fp:
+    for fname, content in zip(_paths(name), (pkgs, repos)):
+        with fopen(fname, 'w') as fp:
             json.dump(content, fp)
     return True
 
 
-def restore(name=None, **kwargs):
+def restore(name=None, clean=False, **kwargs):
     '''
     Make sure that the system contains the packages and repos from a
     frozen state.
@@ -191,6 +191,11 @@ def restore(name=None, **kwargs):
     name
         Name of the frozen state. Optional.
 
+    clean
+        If True remove the frozen information YAML from the cache
+
+        .. version-added:: Neon
+
     CLI Example:
 
     .. code-block:: bash
@@ -204,8 +209,8 @@ def restore(name=None, **kwargs):
 
     frozen_pkgs = {}
     frozen_repos = {}
-    for name, content in zip(_paths(name), (frozen_pkgs, frozen_repos)):
-        with fopen(name) as fp:
+    for fname, content in zip(_paths(name), (frozen_pkgs, frozen_repos)):
+        with fopen(fname) as fp:
             content.update(json.load(fp))
 
     # The ordering of removing or adding packages and repos can be
@@ -291,5 +296,10 @@ def restore(name=None, **kwargs):
             msg = 'Error removing %s repository: %s'
             log.error(msg, repo, e)
             res['comment'].append(msg % (repo, e))
+
+    # Clean the cached YAML files
+    if clean and not res['comment']:
+        for fname in _paths(name):
+            os.remove(fname)
 
     return res
--- a/salt/modules/freezer.py
+++ b/salt/modules/freezer.py
@@ -151,7 +151,8 @@ def freeze(name=None, force=False, **kwargs):
     states_path = _states_path()
 
     try:
-        os.makedirs(states_path)
+        if not os.path.exists(states_path):
+            os.makedirs(states_path)
     except OSError as e:
         msg = 'Error when trying to create the freezer storage %s: %s'
         log.error(msg, states_path, e)
--- a/salt/modules/freezer.py
+++ b/salt/modules/freezer.py
@@ -1,0 +1,294 @@
+# -*- coding: utf-8 -*-
+#
+# Author: Alberto Planas <aplanas@suse.com>
+#
+# Copyright 2018 SUSE LINUX GmbH, Nuernberg, Germany.
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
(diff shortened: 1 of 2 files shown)


waynew:
Hey @aplanas thanks for the PR and backport - checking this one out now 😃

salt/modules/freezer.py
```diff
@@ -0,0 +1,305 @@
...
+    'list_': 'list',
+}
+
+
+def __virtual__():
+    '''
+    Freezer is based on top of the pkg module.
+
+    Return True as pkg is going to be there, so we can avoid of
+    loading all modules.
```
waynew: I'm not sure I understand this comment. Are you just referring to the fact that `pkg` should always be an available module, and is the only dependency?

> so we can avoid of loading all modules

That's the part I'm most confused about (might just be something lost in translation) 😕 

Do you mean that since we're assuming that `pkg` exists we don't have to worry about trying to import it first? Also that we're avoiding the overhead?

If so, maybe this could be

```
Return True since the `pkg` module should always exist. This avoids the overhead of loading all modules.
```
(or "the" modules?)

salt/modules/freezer.py
```diff
@@ -0,0 +1,305 @@
...
+
+    '''
+    ret = []
+    states_path = _states_path()
+    if not os.path.isdir(states_path):
+        return ret
+
+    for state in os.listdir(states_path):
+        if state.endswith(('-pkgs.yml', '-reps.yml')):
+            # Remove the suffix, as both share the same size
```
waynew: I'm not sure that I understand this comment either. "as both share the same size" is somewhat confusing. Do you mean that a `freezer` (or whatever the term is) has one or both `-pkgs.yml` and `-reps.yml` for one name?

If so, perhaps 
```suggestion
            # Remove the suffix - both files start with the freezer name
```
would be more clear.

salt/modules/freezer.py
```diff
@@ -0,0 +1,305 @@
...
+            # all the parameters
+            _tmp_kwargs = frozen_repos[repo].copy()
+            _tmp_kwargs.update(safe_kwargs)
+            __salt__['pkg.mod_repo'](repo, **_tmp_kwargs)
+            res['repos']['add'].append(repo)
+            log.info('Added missing repository %s', repo)
+        except Exception as e:
+            msg = 'Error adding %s repository: %s'
+            log.error(msg, repo, e)
+            res['comment'].append(msg % (repo, e))
```
waynew: Would it be a good idea to just pull each of these out into their own functions? That would make easier to verify that we're following the proposed order:

```
    ... = _add_missing_repos(...)
    ... = _add_missing_packages(...)
    ... = _remove_extra_packages(...)
    ... = _remove_extra_repos(...)
```

I don't know if it's *necessary*, but it might make things a bit clearer.

salt/modules/freezer.py
```diff
@@ -0,0 +1,305 @@
...
+
+    # Add missing packages
+    # NOTE: we can remove the `for` using `pkgs`. This will improve
+    # performance, but I want to have a more detalied report of what
+    # packages are installed or failled.
```
waynew: 👍 for calling out potential performance gains.

Also typo: `s/failled/failed` 🙃

tests/unit/modules/test_freezer.py
```diff
@@ -0,0 +1,324 @@
...
+        Test if a frozen state exist.
+        '''
+        isfile.side_effect = (True, True)
+        self.assertTrue(freezer.status())
+
+        isfile.side_effect = (True, False)
+        self.assertFalse(freezer.status())
+
+    @patch('os.listdir')
+    @patch('os.path.isdir')
```
waynew: @s0undt3ch do I remember correctly that we should be using `with patch` instead of the decorator approach?

s0undt3ch: As much as possible, there are cases where zo remember having to use the decorator, but that was likely because I didn't try harder.

https://engblog.nextdoor.com/what-your-mocks-do-when-you-aren-t-looking-b278e0d9e201

doc/topics/releases/neon.rst
```diff
@@ -3,3 +3,9 @@
 ==================================
 Salt Release Notes - Codename Neon
 ==================================
+
+Module Changes
+==============
+
+- The :py:func:`freezer.restore <salt.modules.freezer.restore>` function provides a
+  new parameter, `clean`, that will remove the frozen data after a successful recovery.
```
waynew: We can probably remove this comment here, since the `freezer` module was never released, right?

waynew (approved):
Code-wise, I think this looks great. There are a couple of comments that I don't *think* are blockers, but I think they could be improved.

salt/modules/freezer.py
aplanas: Ouch yes. Thanks, is more clear now.

salt/modules/freezer.py
aplanas: I agree. I take your suggestion. Thanks!!

salt/modules/freezer.py
aplanas: I think that this is a sensitive refactoring. Done!

salt/modules/freezer.py
aplanas: Ouch. Fixed, thanks!!

tests/unit/modules/test_freezer.py
aplanas: I will follow this rule in future tests. I understand some of the potential issues with side-effects when a `with` context is not used, and I am not reusing the mocks on those tests.

doc/topics/releases/neon.rst
aplanas: You are right. Done

salt/modules/freezer.py
```diff
@@ -0,0 +1,319 @@
+# -*- coding: utf-8 -*-
+#
+# Author: Alberto Planas <aplanas@suse.com>
+#
+# Copyright 2018 SUSE LINUX GmbH, Nuernberg, Germany.
```
dwoz: Please remove this copyright.

dwoz (changes requested):

s0undt3ch (approved):
Aside from the Copyright removal, I'm cool with this PR.

salt/modules/freezer.py
aplanas: Done.

aplanas:
If I split it in two, will help the reviews?

PR #54964, opened 2019-10-11
(Backport 50706) blkid: add search by token

aplanas:
### What does this PR do?

Add the `token` parameter in `blkid`, to allow extended search.

### Tests written?

Yes

(backport #50706, already merged in develop)

### DIFF
--- a/tests/unit/modules/test_disk.py
+++ b/tests/unit/modules/test_disk.py
@@ -188,6 +188,16 @@ class DiskTestCase(TestCase, LoaderModuleMockMixin):
         ), patch("salt.modules.disk.blkid", MagicMock(return_value=STUB_DISK_BLKID)):
             self.assertDictEqual(STUB_DISK_BLKID, disk.blkid())
 
+    @skipIf(salt.utils.platform.is_windows(), "Skip on Windows")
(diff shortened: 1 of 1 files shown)


tests/unit/modules/test_disk.py
```diff
@@ -99,6 +99,13 @@ def test_blkid(self):
                 patch('salt.modules.disk.blkid', MagicMock(return_value=STUB_DISK_BLKID)):
             self.assertDictEqual(STUB_DISK_BLKID, disk.blkid())
 
+    def test_blkid_token(self):
```
dwoz: @aplanas Looks like these tests fail on windows, skip them?

dwoz (changes requested):

tests/unit/modules/test_disk.py
aplanas: Ouch. Yes!

The real problem is the interaction with the decorator in the disk module, that search for the blkid binary. I think that is better to skip than to add a new mock for this decorator.

Thanks!

aplanas:
no falling tests, re-review please?

dwoz (approved):

PR #54965, opened 2019-10-11
(Backport 50725) mount: add fstab_{present,absent} states

aplanas:
### What does this PR do?

Add two new states in mount:

### Tests written?

Yes

(backport #50725, already merged in develop)

### DIFF
--- a/salt/states/mount.py
+++ b/salt/states/mount.py
@@ -956,3 +956,295 @@ def mod_watch(name, user=None, **kwargs):
     else:
         ret['comment'] = 'Watch not supported in {0} at this time'.format(kwargs['sfun'])
     return ret
+
+
+def _convert_to(maybe_device, convert_to):
+    '''
+    Convert a device name, UUID or LABEL to a device name, UUID or
+    LABEL.
+
+    Return the fs_spec required for fstab.
+
+    '''
+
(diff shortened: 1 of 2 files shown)


dhiltonp:
Thank you so much for this PR with these 2 backports. Could you split them up in to two separate PRs?

The functionality of these 2 PRs overlaps and I'm not sure both should go in.

dhiltonp (changes requested):
Thank you so much for this PR with these 2 backports. Could you split them up in to two separate PRs?

The functionality of these 2 PRs overlaps and I'm not sure both should go in.

aplanas:
@dhiltonp they overlap, as the second one is build as a change for the first one. You can see it, for example, here: https://github.com/saltstack/salt/pull/54965/commits/ae341590ce7a4993ad55f679b3479bbd4f5382e0#diff-ba2bfb2cb504d0deb9fc5af6584c9032L994

dhiltonp:
Yeah. It is annoying having PRs that depend on each other.

I certainly see why the 52962 port PR would include the 50725 commits, but there is value in also having an individual port PR for 50725.

aplanas:
@dhiltonp very well. I split the PR in two, the other backport is living here in #55516

dwoz (approved):

PR #50834, opened 2018-12-12
Add disk_set and disk_toggle functions, and update valid partition flags

aplanas:
### What does this PR do?

* Add disk_set and disk_toggle parted functions
* Update the list of valid flags for partitions, based on parted 3.2

### Tests written?

Yes, for the new code

PR #54966, opened 2019-10-11
(Backport 50834) Add disk_set and disk_toggle functions, and update valid partition flags

aplanas:
### What does this PR do?

* Add disk_set and disk_toggle parted functions
* Update the list of valid flags for partitions, based on parted 3.2

### Tests written?

Yes

(backport #50834 already merged in develop)


aplanas:
I split the backports here and in #55515

waynew:
Looks like this is already included in #55515 - let me know if that's not correct

PR #54970, opened 2019-10-11
[wip] event throttling mechanism

pizzapanther:
### What does this PR do?

Mechanism to throttle events. When in a complex environment, this can help you put in a failsafe to prevent duplicate work.

Example Config:

```
event_throttles:
  cache_path: /dev/shm/event_throttles.dict
  cache_ttl: 300
  rules:
    - method: startswith
      match: minion/refresh/
```

Note: for the cache path, you probably want to use something like `/dev/shm` so that disk operations are fast an in memory only. 

Example Test:
(with the config above)

1. Run several times: `salt 'zort' saltutil.refresh_grains` 
2. The master should skip minion refresh events for 300 seconds after the first event. Can be verified in the log.

### DIFF
--- a/salt/utils/event.py
+++ b/salt/utils/event.py
@@ -722,7 +722,6 @@ class SaltEvent(object):
 
         yield self.pusher.send(msg)
 
-
     def fire_event(self, data, tag, timeout=1000):
         '''
         Send a single event into the publisher with payload dict "data" and
--- a/salt/utils/event.py
+++ b/salt/utils/event.py
@@ -71,6 +71,7 @@ from salt.ext.six.moves import range
 
 # Import third party libs
 from salt.ext import six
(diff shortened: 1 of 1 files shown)


pizzapanther:
This is a prototype right now, looking for feedback from @dwoz 

No review need yet.

amalaguti:
From a recent conversation in Slack, this feature and event filtering into raas-master would be great addition to Enterprise customers. Some customers have expressed opinion that there is way too much info and events recorded in SSE. 
Ideally, there should be a feature to filter events at the raas plugin level AND at the SSE Console (dynamically by selecting fields). Would be great if the options to both filters can be managed by SSE directly (instead of requiring modifying conf files).

dwoz:

PR #54996, opened 2019-10-14
(Backport 51871) cmdmod: add 'binds' parameter in run_chroot

aplanas:
### What does this PR do?

Sometimes we want to export directories from the system to the
chroot environment. For example, some services will require the
presence of /run inside the chroot, to request some PID information.

This patch add a new 'binds' parameter in the run_chroot function,
that will export those system directories into the chroot.

### Tests written?

Yes, added in this PR

(backport #51871, already merged in develop)

### DIFF
--- a/salt/modules/cmdmod.py
+++ b/salt/modules/cmdmod.py
@@ -2904,6 +2904,7 @@ def run_chroot(root,
                group=None,
                shell=DEFAULT_SHELL,
                python_shell=True,
+               binds=None,
                env=None,
                clean_env=False,
                template=None,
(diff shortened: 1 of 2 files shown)


salt/modules/cmdmod.py
```diff
@@ -2965,6 +2964,9 @@ def run_chroot(root,
         arguments. Set to True to use shell features, such as pipes or
         redirection.
 
+    :param list binds: List of directories that will be exported inside
```
waynew: Should we have a versionadded `Sodium` (i.e. the next major release) here? I don't remember the precise syntax for it.

waynew:
@aplanas thanks for adding the tests and getting the backport. Looks like this is out of date with `master`, would you mind rebasing?

I know at least one of these failures is a flaky test that we've been working on, that shouldn't block the PR from merging. Hopefully with the rebase we should see that test pass.

salt/modules/cmdmod.py
aplanas: Indeed. Added. I used the version name, instead the number (https://docs.saltstack.com/en/latest/topics/releases/version_numbers.html have a TBD for the version number in Sodium)

Thanks!

dwoz (approved):

PR #54998, opened 2019-10-14
(Backport 51956) blockdev: hide blkid errors when are expected

aplanas:
### What does this PR do?

The function _checkblk is called to detect if the device is already
formatted. The first time that we call blkid -o value -s TYPE will
will fail, as the device is still not formatted.

This patch hide this error from the logs (only the ERROR status,
not the stderr output and the retcode), as is an expected error.

### Tests written?

No, no change in behavior

(backport #51956, already merged in develop)

### DIFF
--- a/salt/states/blockdev.py
+++ b/salt/states/blockdev.py
@@ -193,5 +193,6 @@ def _checkblk(name):
     Check if the blk exists and return its fstype if ok
     '''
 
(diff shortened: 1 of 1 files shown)


dwoz:
@aplanas Can you please resolve the conflicts here?

aplanas:
The fix is already in master! So this PR is not needed anymore. Closing and thanks for the reviews!!

Issue #39475, opened 2017-02-17
[Feature] Create hard link salt state to go with the file.link module.


weirdbricks:
### Description of Issue/Question

Hello,

I'm very new to Salt so hopefully I'm not doing something dumb.

The issue I'm having is that I can create a symlink by using `file.symlink` but I cannot create a hard link by using `file.link`.

### Setup

The symlink works:

```
create stunnel link:
  file.symlink:
    - name: /opt/init/stunnel-alpine.init
    - target: /root/tools-monit-configs/stunnel-alpine.init
```

The hard link does not:

```
create stunnel link:
  file.link:
    - name: /opt/init/stunnel-alpine.init
    - target: /root/tools-monit-configs/stunnel-alpine.init
```

Interestingly enough, this *does* work:

```
salt '*' file.link /root/tools-monit-configs/stunnel-alpine.init /opt/init/stunnel-alpine.init
```

### Steps to Reproduce Issue
The error I get is:

```
          ID: create stunnel link
    Function: file.link
        Name: /opt/init/stunnel-alpine.init
      Result: False
     Comment: State 'file.link' was not found in SLS 'telegraf.docker'
              Reason: 'file.link' is not available.
     Changes: 
```

### Versions Report
(Provided by running `salt --versions-report`. Please also mention any differences in master/minion versions.)

Master: 

```
salt --versions-report
Salt Version:
           Salt: 2016.11.0
 
Dependency Versions:
           cffi: 1.8.3
       cherrypy: Not Installed
       dateutil: 2.5.3
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.8
        libgit2: 0.24.3
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.4.8
   mysql-python: Not Installed
      pycparser: 2.14
       pycrypto: 2.6.1
         pygit2: 0.24.2
         Python: 2.7.13 (default, Dec 22 2016, 09:22:15)
   python-gnupg: Not Installed
         PyYAML: 3.12
          PyZMQ: 16.0.2
           RAET: Not Installed
          smmap: Not Installed
        timelib: Not Installed
        Tornado: 4.4.2
            ZMQ: 4.1.5
 
System Versions:
           dist:   
        machine: x86_64
        release: 3.10.0-327.36.3.el7.x86_64
         system: Linux
        version: Not Installed
```

[...]

PR #20428, opened 2015-02-05
GH-16780: add hardlink function to file exec/state modules

systembell:
This change resolves #16780, adding a `hardlink` function to the `file` state and execution modules. It makes does not perform any mount point checking, thus assumes that the operator knows if the `target` and `dest` reside on the same filesystem (it will throw an `OSError` if an attempt is made to create a hard link across devices).

PR #55000, opened 2019-10-14
Implement hardlink() for both salt.modules.file and salt.states.file

arizvisa:
### What does this PR do?
This adds the `hardlink()` function as intended by PR #20428 but was dropped by the author. PR #20428 seemed to be a copy-paste job from `symlink()`, and although this PR is based on it. The semantics in this PR were changed so that it made sense for hard links. (i.e. you can't hard link a directory, lack of windows support, etc.) Hard link symmetry is determined by comparing the inode of the source and target paths.

This closes issue #39475 which was actually closed originally due to the issue going stale.

### Previous Behavior
There was no way to create hard links on posixy environments.

### New Behavior
There are now two functions in `salt.modules.file`. These are the `.is_hardlink()` which checks the st_nlink field to determine if something is a hard link, and `.hardlink()` which can be used to actually make the link. There is also a new state in `salt.states.file` which allows one to create a hard link without overwriting a non-hardlink unless the `force` option is provided.

### Tests written?
Yes. It was very painful. It is 4am over here.

### Commits signed with GPG?
No.


arizvisa:
(Force-push'd to cull out reverts)

arizvisa:
Okay. I really need help on this. Somebody halp plz.

In this PR, I haven't touched any of the following lines. I do use `CommandExecutionError`, but that shouldn't matter between Python2/3. The issue is that empty space between the imports and the if-test results in a lint error of `incompatible-py3-code`. I developed this and the testcases using Python3.7, and all of the Python3 tests passed, so this doesn't appear to be correct...

These are the relevant lines (according to the lint check) in salt.states.file at line +314. Again, line +314 is the white space in the center.
```
import salt.utils.versions
from salt.serializers import DeserializationError
from salt.state import get_accumulator_dir as _get_accumulator_dir

if salt.utils.platform.is_windows():
    import salt.utils.win_dacl
    import salt.utils.win_functions
    import salt.utils.winapi
```

dwoz (changes requested):
@arizvisa The linter is un-happy

arizvisa:
> @arizvisa The linter is un-happy

Yeah, my previous message mentions the following, and I'm not sure how to proceed here because I haven't touched the relevant lines at all, and the PR works in Python3 for me.

```
Okay. I really need help on this. Somebody halp plz.

In this PR, I haven't touched any of the following lines. I do use CommandExecutionError, but that shouldn't matter between Python2/3. The issue is that empty space between the imports and the if-test results in a lint error of incompatible-py3-code. I developed this and the testcases using Python3.7, and all of the Python3 tests passed, so this doesn't appear to be correct...

These are the relevant lines (according to the lint check) in salt.states.file at line +314. Again, line +314 is the white space in the center.

import salt.utils.versions
from salt.serializers import DeserializationError
from salt.state import get_accumulator_dir as _get_accumulator_dir

if salt.utils.platform.is_windows():
    import salt.utils.win_dacl
    import salt.utils.win_functions
    import salt.utils.winapi
```

arizvisa:
@dwoz, apparently from debugging PyLint it believes that
`for direction, item in zip(['to', 'from'], [name, target]):` is not being iterated through and is hence failing `pylint.checkers.python3._in_iterating_context`.

This is because pylint's checker is looking for `For` to distinguish this via the following code in `_in_iterating_context`.
```
    if isinstance(parent, astroid.For):
        return True
```
However, the parent for `zip` there is actually
```
>>> list(b[1].get_children())[0]
<Name.zip l.1 at 0x7f522f3b55d0>
>>> list(b[1].get_children())[0].parent
<Call l.1 at 0x7f522f3b5590>
```

which instead takes this branch and `*_ACCEPTS_ITERATOR' doesn't include `zip`:
```
    elif isinstance(parent, astroid.Call):
        if isinstance(parent.func, astroid.Name):
            if parent.func.name in _ACCEPTS_ITERATOR:
                return True
        elif isinstance(parent.func, astroid.Attribute):
            if parent.func.attrname in ATTRIBUTES_ACCEPTS_ITERATOR:
                return True
```

Instead of sitting around and waiting for PyLint to fix their issue which would then block this PR,
I replaced my usage of `zip` with `zip_longest` (not referenced at all in PyLint) so that PyLint doesn't even bother to try and execute this busted code and will thus not even try to determine whether it's being iterated through or not.

```
   if node.func.name in ("filter", "map", "range", "zip"):
                    if not _in_iterating_context(node):
                        checker = "{}-builtin-not-iterating".format(node.func.name)
                        self.add_message(checker, node=node)
```

All our tools are fucked..

dwoz:
@arizvisa In the future you can also skip that check for the line pylint doesn't understand `# pylint: check-name-or-id`

dwoz (approved):

arizvisa:
@dwoz, gotcha. thanks for the heads up.

PR #55003, opened 2019-10-14
(Backport 52304) utils/path.py: fix a DeprecationWarning

aplanas:
### What does this PR do?

Fix DeprecationWarning about using or importing ABCs from collections,
instead of from 'collections.abc'

### Tests written?

No, as this is a deprecation warning emitted by the interpreter. No behavior change.

(backport #52304, already merged in develop)


dhiltonp (approved):

dhiltonp:
More importantly, the deprecated collections reference was removed in 3.9.

We have another ~60 raw imports of collections that need similar treatment.

PR #55004, opened 2019-10-14
(Backport 52698) conftest: replace deprecated get_marker call in pytest

aplanas:
### What does this PR do?

The current conftest.py is using deprecated call to `get_marker()` in the item elements. We need to replace them with the new `get_closest_marker()` call, as documented in:

https://docs.pytest.org/en/latest/mark.html#updating-code

I didn't use `iter_markers()` iterator as looks like that there is not overwrite.

### Tests written?

Is part of the test code

(backport #52698, already merged in develop)


waynew:
Looks like we've got this change in already, so I'm closing this - let me know if you have any questions!

aplanas:
oh thanks! this makes sense.

Issue #54413, opened 2019-09-05
hosts.add_host rewrites the entire /etc/hosts by separating fields with tab chars only


sonickenbaker:
### Description of Issue
Using `hosts.add_host` all the entry in `/etc/hosts` are separated by using 2 tab characters. Checking hosts' man entry:
_Fields of the entry are separated by any number of blanks and/or tab characters._

This creates issues if a directive like `$include` is used inside the `/etc/hosts` file. By using tabs that directive does not work.

The issue is present even by using the `host.present` state, since makes use of the `hosts.add_host` module.

[Checking the code](https://github.com/saltstack/salt/blob/v2019.2.0/salt/modules/hosts.py#L300), tab characters are used a separator to rewrite alle the hosts

### Setup
Salt installed on a Ubuntu 18.04.2 LTS via bootstrap script. Master and minion on the same machine. Simple module execution is involved.

### Steps to Reproduce Issue

- create an entry in `/etc/hosts` and separate "IP_address" and "canonical_hostname" using spaces instead of tabs: `echo "111.111.111.111 a_hostname" >> /etc/hosts`
- `cat -T /etc/hosts | grep a_hostname` to check that a blank char have been inserted
- launch: `salt-call hosts.add_host 222.222.222.222 another_hostname`
- `cat -T /etc/hosts | grep a_hostname` . Blank char is now replaced by 2 tabs characters (Tabs are printed as "^I")

### Versions Report
Salt Version:
           Salt: 2019.2.0

Dependency Versions:
           cffi: Not Installed
       cherrypy: Not Installed
       dateutil: Not Installed
      docker-py: 2.7.0
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.10
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: 0.27.0
           Mako: 1.0.14
   msgpack-pure: Not Installed
 msgpack-python: 0.5.6
   mysql-python: 1.2.5
      pycparser: Not Installed
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: Not Installed
         Python: 2.7.15+ (default, Nov 27 2018, 23:36:35)
   python-gnupg: Not Installed
         PyYAML: 3.13
          PyZMQ: 16.0.2
           RAET: Not Installed
          smmap: Not Installed
        timelib: Not Installed
        Tornado: 4.2.1
            ZMQ: 4.2.5

System Versions:
[...]

PR #55030, opened 2019-10-16
Expire0 patch 1

Expire0:
removing the \t tabs from the host file . This is causing a issue under saltstack#54413

Reference to issue [Issue](https://github.com/saltstack/salt/issues/54413)

What does this PR do?
Using hosts.add_host all the entry in /etc/hosts are separated by using 2 tab characters. Checking hosts' man entry:
Fields of the entry are separated by any number of blanks and/or tab characters.

This creates issues if a directive like $include is used inside the /etc/hosts file. By using tabs that directive does not work.

saltstack#54413 Previous Behavior

Tests written?
yes , I confirm the issue has been resolve and double spacing is being used. # # # # #

### DIFF
--- a/doc/ref/states/ordering.rst
+++ b/doc/ref/states/ordering.rst
@@ -5,10 +5,10 @@ Ordering States
 ===============
 
 The way in which configuration management systems are executed is a hotly
-debated topic in the configuration management world. Two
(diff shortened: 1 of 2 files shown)


dwoz:
@Expire0 This needs to be re-opened against the `master` branch. We are no

PR #55046, opened 2019-10-17
[WIP] Windows Server 2012r2 winrm and psexec test failure fixes

Akm0d:
### What does this PR do?
Add delay after rename in EC2 test


tests/integration/cloud/clouds/test_ec2.py
```diff
@@ -122,11 +123,10 @@ def test_instance_rename(self):
 
         changed_name = self.instance_name + '-changed'
 
-        rename_result = self.run_cloud(
-            '-a rename {0} newname={1} --assume-yes'.format(self.instance_name, changed_name), timeout=TIMEOUT)
-        self.assertFalse(self._instance_exists(), 'Instance wasn\'t renamed: |\n{}'.format(rename_result))
+        self.run_cloud('-a rename {0} newname={1} --assume-yes'.format(self.instance_name, changed_name), timeout=TIMEOUT)
+        sleep(20)
```
waynew: I think that what I'd prefer to see here , rather than a simple `sleep(20)`, is do something like:

```
tries = 10
while tries > 0:
    tries -= 1
    if not self._instance_exists():
        tries = 0
    else:
        time.sleep(2)
```

Or something to that effect (that would wait for 20s total, max). One could increase the timeout length (e.g. 5s between checks) or whatever, to tune things. That would allow it to return faster if able, or take longer, if there are network issues.

Obviously one would want to have a reasonable maximum, and a reasonable time between checks...

I don't know if it's worth holding up the PR, but it might be a better choice.

Akm0d: Cloud test PRs aren't blocking since cloud tests don't run on PR builds.  Your suggestion is good and I'll implement that real quick :+1:

Akm0d:
re-run full ubuntu 1804

Akm0d:
re-run full ubuntu1604

tests/integration/cloud/clouds/test_ec2.py
waynew: I think that what I'd prefer to see here , rather than a simple `sleep(20)`, is do something like:

```
tries = 10
while tries > 0:
    tries -= 1
    if not self._instance_exists():
        tries = 0
    else:
        time.sleep(2)
```

Or something to that effect (that would wait for 20s total, max). One could increase the timeout length (e.g. 5s between checks) or whatever, to tune things. That would allow it to return faster if able, or take longer, if there are network issues.

Obviously one would want to have a reasonable maximum, and a reasonable time between checks...

I don't know if it's worth holding up the PR, but it might be a better choice.

waynew (approved):

Akm0d:
re-run full all

Akm0d:
Cherry-picking https://github.com/saltstack/salt/pull/54911/commits into this PR for master

Akm0d:
Tracked the cloud test error down to `The request must contain the parameter AWSAccessKeyId`

salt/utils/cloud.py
```diff
@@ -1366,7 +1366,13 @@ def deploy_windows(host,
...
                 return False
 
-            time.sleep(5)
+            for retry in range(100):
+                stdout, stderr, ret_code = run_psexec_command(
+                    'cmd.exe', '/c sc query salt-minion', host, username, password
+                )
+                log.debug('Cloud winrm command #{}: {}'.format(retry, stdout))
+                if 'STOP_PENDING' not in stdout:
+                    break
```
waynew: Would it make sense to put a momentary or increasing delay in here? I don't know that much about the psexec thing, but it's not obvious if this will just burn through the calls as fast as it can.

Akm0d:
This needs to be paired with a change to sre-jenkins: https://github.com/saltstack/sre-jenkins/pull/214

s0undt3ch (approved):

Akm0d:
`Salt-Minion-2019.2.2-Py3-AMD64-Setup.exe` is causing this test to hang

Akm0d:
This issue seems to be the root cause of the test failures: https://github.com/saltstack/salt/issues/55119

PR #55047, opened 2019-10-17
[master] Allow deprecating by date

s0undt3ch:
### What does this PR do?
Allow deprecating by date. Includes tests.

### Previous Behavior
We could only deprecated by version names

### New Behavior
We can now deprecate based on dates

### Tests written?

Yes

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/utils/versions.py
+++ b/salt/utils/versions.py
@@ -17,6 +17,9 @@ import logging
 import numbers
 import sys
 import warnings
+import datetime
+import inspect
+import contextlib
 # pylint: disable=blacklisted-module,no-name-in-module
 from distutils.version import StrictVersion as _StrictVersion
 from distutils.version import LooseVersion as _LooseVersion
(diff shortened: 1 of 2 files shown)


salt/utils/versions.py
```diff
@@ -162,6 +163,92 @@ def _formatwarning(message,
...
+            '{today}. Please remove the warning.'.format(
+                filename=caller.filename,
+                lineno=caller.lineno,
+                date=date,
+                today=today,
+            ),
+        )
+
+    if _dont_call_warnings is False:
+        def _formatwarning(message,
```
dwoz: Does this have to be a nested function?

dwoz: Does `_formatwarning` need to be a nested function?

dwoz (changes requested):

salt/utils/versions.py
s0undt3ch: That was copied from the other `warn_until` function, but I guess I can be reused and factored out.

s0undt3ch: Fixed

dwoz (approved):

twangboy (approved):

s0undt3ch:
Updated PR. We don't have to patch under Py3.

Akm0d (approved):

waynew (approved):

salt/utils/versions.py
```diff
@@ -140,26 +164,92 @@ def warn_until(version,
...
-                filename, lineno, category.__name__, message
+        with _patched_format_warning():
+            warnings.warn(
+                message.format(version=version.formatted_version),
+                category,
+                stacklevel=stacklevel
+            )
+
+
+def _get_utcnow_date():
```
dwoz: This can also be solved by making this and other time methods injectable dependencies to `warn_until_date`. This way tests can use a different version of the method without needing a separate mockable method in this module.

s0undt3ch: Let's address that in a future PR.

dwoz (approved):

PR #55043, opened 2019-10-17
[WIP][master] Greatly Simplify(refactor) Salt's Logging

s0undt3ch:
### What does this PR do?
This PR simplifies Salt's logging setup, in particular it changes how we handle multiprocessing logging.
It also deprecates the old logging setup and set's it's removal for the first release after 2022/Jan/01.

### Previous Behavior
Multiprocessing worked OK in Linux, and not so OK in Windows.

### New Behavior
Logging setup simplified and multiprocessing logging working great on all distributions

### Tests written?

Yes, and some fixed.

### Commits signed with GPG?

Yes

PR #55053, opened 2019-10-18
[master] Deprecate the multiprocessing process long names

s0undt3ch:
### What does this PR do?
Deprecate the multiprocessing process long names.

Who on earth named these classes?!

Oh. It was me.
I must have been insane at the time because these names are absurd.

This PR should only be merged after https://github.com/saltstack/salt/pull/55047 since it includes code from that PR to provide the deprecation by date feature.

### Previous Behavior
We had insane names two classes.

### New Behavior
We have proper names for those two classes.
The previous names still exist but are deprecated and set to be removed on the first release after Jan, 1st, 2022.

### Tests written?

No, this is just deprecating class names.

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/client/mixins.py
+++ b/salt/client/mixins.py
@@ -508,8 +508,7 @@ class AsyncClientMixin(object):
         to watch for the return
         '''
         async_pub = pub if pub is not None else self._gen_async_pub()
-
-        proc = salt.utils.process.SignalHandlingMultiprocessingProcess(
+        proc = salt.utils.process.SignalHandlingProcess(
                 target=self._proc_function,
                 args=(fun, low, user, async_pub['tag'], async_pub['jid']))
         with salt.utils.process.default_signals(signal.SIGINT, signal.SIGTERM):
--- a/salt/client/netapi.py
+++ b/salt/client/netapi.py
@@ -14,7 +14,7 @@ import salt.utils.process
 log = logging.getLogger(__name__)
 
 
-class RunNetapi(salt.utils.process.SignalHandlingMultiprocessingProcess):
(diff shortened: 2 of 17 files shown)


rares-pop:
The change looks good, but we can do more cleanup vis-à-vis of this class. More specifically regarding the _is_child abstraction leakage (where any class that inherits from this has to specify '_is_child = True' in order to behave correctly.

s0undt3ch:
Well, this PR is pre-work for https://github.com/saltstack/salt/pull/55043 which refactors logging, mostly because it has grew too complicated and on windows it does leak(we can't switch to pytest on windows right now because on this multiprocessing leakage).

However, `_is_child` is no longer in use(at least on the master branch), it just looks like some leftovers from whatever tweaks we once did.
I'm removing those keys in this PR.

rares-pop:
Oh, I wasn't aware of the other one. I'll take a look on it.

I do use both salt-master and salt-minion, along with salt-api on Windows, and have systems running for months. I can't say they are not leaking, but I can say that in case they do, it's not obvious and a showstopper.

If you get rid of '_is_child', please make sure it still works on Windows.

This is the aforementioned PR: https://github.com/saltstack/salt/pull/52471

Akm0d (approved):

Akm0d (approved):

waynew (approved):

dwoz:
The `is_child` attribute was removed by me in 2019.2.1. It is safe to remove it from these classes and everything still works on windows.

dwoz (approved):

PR #55055, opened 2019-10-18
[master] Initial work for the new logging setup for salt

s0undt3ch:
### What does this PR do?
Starts moving code around to support the new Salt logging support.
This is part of https://github.com/saltstack/salt/pull/55043 so that it gets implemented/reviewed in smaller chunks.

https://github.com/saltstack/salt/pull/55043

### Commits signed with GPG?

Yes/No


Akm0d (approved):

salt/_logging/__init__.py
```diff
@@ -0,0 +1,18 @@
+# -*- coding: utf-8 -*-
+'''
+    salt._logging
+    ~~~~~~~~~~~~~
+
+    This is salt's new logging setup.
+    As the name suggests, this is considered an internal API which can change without notice,
+    although, as best effort, we'll try not to break code relyin on it.
```
waynew:
```suggestion
    although, as best effort, we'll try not to break code relying on it.
```

salt/_logging/__init__.py
```diff
@@ -0,0 +1,18 @@
...
+
+    The ``salt._logging`` package should be imported as soon as possible since salt teaks
```
waynew:
```suggestion
    The ``salt._logging`` package should be imported as soon as possible since salt tweaks
```

waynew (changes requested):
I see that your PR is in chunks now 😂

salt/_logging/__init__.py
s0undt3ch: Will fix

s0undt3ch: Fixed

salt/_logging/__init__.py
s0undt3ch: Fixed

waynew (approved):
😞emoji, y u no have :log:???

dwoz (approved):

PR #55065, opened 2019-10-20
Fix multiprocessing process after fork and finalize regression

s0undt3ch:
### What does this PR do?
Fix multiprocessing process after fork and finalize regression

PR #54277 introduced a regression where `salt.utils.process.Process`
subclasses which override their `run()` method would not have the
`_after_fork_methods` and `_finalize_methods` executed.

This PR **must** be merged after #55043 since it relies on the new class names.

### Previous Behavior
Classes which sub-classed `salt.utils.process.Process` or `salt.utils.process.SignalHandlingProcess` which also defined a `run()` method would not run through the `_after_fork_methods` and `_finalize_methods`.

### New Behavior

Classes which subclass `salt.utils.process.Process` or `salt.utils.process.SignalHandlingProcess` which also defines a `run()` **will** run through the `_after_fork_methods` and `_finalize_methods`.

### Tests written?

Yes

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/utils/process.py
+++ b/salt/utils/process.py
@@ -720,7 +720,10 @@ class Process(multiprocessing.Process, NewStyleClassMixIn):
         # we must wrap this class run method to allow for these extra steps
         # to be executed pre and post calling the actual run method,
         # having subclasses call super would just not work.
-        self.run = self.__decorate_run(self.run)
+        #
+        # We use setattr here to fool pylint not to complain that we're
+        # overriding run from the subclass here
+        setattr(self, 'run', self.__decorate_run(self.run))
 
     # __setstate__ and __getstate__ are only used on Windows.
     def __setstate__(self, state):
--- a/salt/utils/process.py
+++ b/salt/utils/process.py
@@ -14,13 +14,13 @@ import errno
 import types
 import signal
 import logging
+import functools
 import threading
 import contextlib
 import subprocess
 import multiprocessing
 import multiprocessing.util
 import socket
-
 
 # Import salt libs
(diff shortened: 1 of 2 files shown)


salt/utils/process.py
```diff
@@ -740,33 +753,51 @@ def __getstate__(self):
     def _setup_process_logging(self):
         salt.log.setup.setup_multiprocessing_logging(self.log_queue)
 
-    def run(self):
+    def _run(self):
         for method, args, kwargs in self._after_fork_methods:
             method(*args, **kwargs)
         try:
-            return super(MultiprocessingProcess, self).run()
```
s0undt3ch: Hmm, tests pass , weirdly, but I actually have to still do some fixing, and find out why the tests passed.
We should be calling `self._origonal_run`

s0undt3ch: Probably because tests are calling `run` and not `start`...

s0undt3ch: Of course, the test suite didn't even start.
Will still investigate the unit tests passing.

s0undt3ch:
Not Ready.

salt/utils/process.py
s0undt3ch: Of course we can't call `.start()`, the mocks will get inherited by the new process, on Linux, but we can no longer assert against them, at least like we usually do.

s0undt3ch:
PR is now ready.

dwoz (approved):

waynew (approved):

Akm0d (approved):

PR #55068, opened 2019-10-20
Prevent grains calculation of master in MasterMinion

### LINKED ISSUE #55067: MasterMinion calculates master grains on every usage
### Description of Issue
High CPU on Salt Master and slow minion responses caused by an extreme number of commands being run on the master. Tracking this down we see that MasterMinion calculates grains when instantiated, however it does not calculate the minions grains - it calculates the master’s grains, and the results are uncashed and unused (grains_cache defaults to false and the grains I cannot find references anywhere)

When setting up the MasterMinion pillar calculation is disabled and it seems grains should be too

### Setup
Salt Master using GifFS and GitPillar
10 Minions

### Steps to Reproduce Issue

Running salt asterisk test.ping takes about a minute
After disabling the grains in MasterMinion it takes seconds 

### Versions Report
2019.2.0

driskell:
These grains do not appear used and are calculated on the master meaning they are not applicable to the requested minion anyway. They also slow down actions significantly in some cases due the enormous number of commands that run.

### What does this PR do?
Prevents grains calculation during MasterMinion setup

Fixes #55067 

### Previous Behavior
For each minion a MasterMinion would be created and it would calculate the grains of the master (not the minion) each time with no cache as default grains_cache is False. The great number of commands needed to calculate grains (CPU type and GPU etc) means this can take an enormous amount of resource, especially if #44758 impacts the system.

### New Behavior
Do not calculate grains for MasterMinion as they appear unused and are incorrect anyway

### Tests written?

No

### Commits signed with GPG?

No


xeacott:
Collecting grains on the master seems to be incorrect behavior here, I wonder when this may have been introduced. Anyways, I realize that this is an _extremely_ small change but because this behavior has been addressed can you also write a partnering test for the change? This can just be to test whether or not a master is calculating grains during runtime, or see if master is storing grains like how a minion would behave.

cmcmarrow (changes requested):
Can we still get a test? SaltStack wants at least a test for every change done to the code repository.

driskell:
I won’t have time at all in the coming weeks I’m afraid so if someone else or a SaltStack employee could do the test that’ll be great.

cmcmarrow (approved):

cmcmarrow:
@driskell this looks good to merge now! Thanks for finding and fixing this grain problem.

driskell:
@xeacott @cmcmarrow Thanks!

sagetherage:
@driskell can you address the merge conflict here, please?

driskell:
@sagetherage Sorry, I won’t be able to do anything for several weeks due to other commitments and personal things. It’s a one line change and a brand new test though so shouldn’t be a problem for someone else to resolve.

driskell:
@sagetherage judging by the conflict this May have been solved already so I’ll close this. I haven’t had many performance issues lately so likely it is the case

PR #55075, opened 2019-10-21
Dedup parser callback registration on parsers code

s0undt3ch:
### What does this PR do?
See title.
This actually only impacted tests since we created several instances of the parser, something regular operations don't do.
Additionally, we now only register functions or unbound methods because we pass the class instance when calling the functions.

### Previous Behavior
We would register the same callback more than once when instantiating more than one parser instance.
We would also register bound methods and yet, when calling them, still pass the class instance.

### New Behavior
We only register the callback once.
We only register functions or unbound methods.
We no longer need to be root to run the parser tests.

### Tests written?

Yes

### Commits signed with GPG?

Yes


salt/utils/parsers.py
```diff
@@ -304,7 +316,7 @@ def _mixin_setup(self):
         # the config options and if needed override them
         self._mixin_after_parsed_funcs.append(self.__merge_config_with_cli)
 
-    def __merge_config_with_cli(self, *args):  # pylint: disable=unused-argument
```
waynew: why do we need to take `*args` here and everywhere else? Wouldn't they just be overridden in the subclass? 🤔

dwoz (approved):

salt/utils/parsers.py
s0undt3ch: We don't, hence it was removed.
Previously, because we were keeping bound methods on the call back lists, we were actually passing `self` twice.

Akm0d (approved):

waynew (approved):

PR #55079, opened 2019-10-21
salt.utils.args: make parse_input non-destructive on its input args

mattp-:
popping __kwargs__ off the input data invalidates the inputs for future
calls, this breaks salt.cli.batch among other things; possibly other
usecases that save low data and re-use it.

### Previous Behavior

### New Behavior

### Tests written?

Yes/No

### Commits signed with GPG?

Yes/No


mattp-:
this should very likely be made a bugfix but I don't know what the current target branch for that is.

PR #55088, opened 2019-10-22
Add is_installed in various pkg manager

Poulpatine:
### What does this PR do?
Add is_installed function 

### New Behavior
is_installed returns True if all packages in names are installed, else returns False.

### Tests written?
Yes where modules are tested


dwoz:

Issue #55094, opened 2019-10-22
docs: Libvirt docs should contain ssh key auth


HG00:
### Description of Issue
"Getting started with libvirt" doesnt mention the use of ssh keys. A VM base image can be made with a root key instead of having to specify a password. 

### Setup
N/A

xeacott:
Thanks for submitting!

Issue #40965, opened 2017-04-29
salt-cloud sudo failing


weirdbricks:
### Description of Issue/Question

Hello,
I am trying to launch some droplets in Digital Ocean using `salt-cloud`. The process seems to be breaking when `salt-cloud` attempts to use sudo.

Do you see anything wrong with my setup ? Thanks!

### Setup

```
#provider
digitalocean-lampros:
  driver: digital_ocean
  ssh_key_name: digital-ocean-salt-cloud.pub
  ssh_key_file: /root/.ssh/id_rsa 
  ssh_username: tools
  ssh_interface: private
  ssh_port: 4910
  sudo: True 
  sudo_password: "--redacted--"
  personal_access_token: --redacted--
  location: San Francisco 2
  minion:
    master: salt 
```

```
#profile
tools:
  provider: digitalocean-lampros
  image: my-super-image
  size: 16GB
  location: San Francisco 2
  private_networking: True
```

### Steps to Reproduce Issue

This is what I use to create the droplet:

```
salt-cloud --profile tools lobsterman4 -l all
```

And this is what I get:

```
[DEBUG   ] Logging into 10.138.136.234:4910 as tools
[DEBUG   ] Using /root/.ssh/id_rsa as the key_filename
[DEBUG   ] Using sudo to run command sudo -S "XXX-REDACTED-XXX" test -e '/tmp/.saltcloud-4ba4b02e-5a85-40ae-8192-cf2dcb527c9d'
[DEBUG   ] SSH command: 'ssh -t -t -oStrictHostKeyChecking=no -oUserKnownHostsFile=/dev/null -oControlPath=none -oPasswordAuthentication=no -oChallengeResponseAuthentication=no -oPubkeyAuthentication=yes -oIdentitiesOnly=yes -oKbdInteractiveAuthentication=no -i /root/.ssh/id_rsa -p 4910 tools@10.138.136.234 sudo -S "XXX-REDACTED-XXX" test -e '/tmp/.saltcloud-4ba4b02e-5a85-40ae-8192-cf2dcb527c9d''
[DEBUG   ] Child Forked! PID: 4921  STDOUT_FD: 4  STDERR_FD: 8
[...]

PR #55095, opened 2019-10-22
Docs: Add SSH Key auth section to libvirt #55094

HG00:
### What does this PR do?
Adds SSH Key authentication section to libvirt doc

#55094 

### Commits signed with GPG?
No


doc/topics/cloud/libvirt.rst
```diff
@@ -106,6 +106,21 @@ When using ``ip-learning`` the IP is requested from libvirt. This needs a recent
 version and may only work for NAT networks. Another option is to use ``qemu-agent`` this requires
 that the qemu-agent is installed and configured to run at startup in the base domain.
 
+SSH Key Authentication
+======================
+Instead of specifying a password, an authorized key can be used for the minion setup. Ensure that
+the root user of your base image has the public key you want to use in .ssh/authorized_keys, then
```
bryceml: a more complex example using sudo and a non root user might be more helpful similar to the config given here: https://github.com/saltstack/salt/issues/40965#issue-225232406
```yaml
#provider
digitalocean-lampros:
  driver: digital_ocean
  ssh_key_name: digital-ocean-salt-cloud.pub
  ssh_key_file: /root/.ssh/id_rsa 
  ssh_username: tools
  ssh_interface: private
  ssh_port: 4910
  sudo: True 
  sudo_password: "--redacted--"
  personal_access_token: --redacted--
  location: San Francisco 2
  minion:
    master: salt 
```
an example with root as well would be good.  Ideally we should have examples that show all options that could be configured.

dwoz:
@HG00 Can you address @bryceml's comment please?

doc/topics/cloud/libvirt.rst
HG00: Sorry for long silence! This PR was for libvirt specifically - I had trouble with other auth methods here - the others work fine like your example, libvirt not so much

HG00: What shall we do here guys - the doc is "Getting started with Libvirt", I dont think any other provider makes sense here

bryceml: If @Akm0d has time to verify these, that would be ideal.  I probably take a look some time this month if not.

bryceml: I didn't mean other providers, I just meant document use of alternative ports, usernames, sudo usage, etc.  Any option that can be used should be documented.

HG00: ok, i put this in because i wanted to use key rather than password login and figured others might, but we can do all libvirt providers options as well - where do i look up every single option so it can be documented?

bryceml: You would likely have to look through the code to get them all, you could get most of them by looking at the documentation for other providers and seeing what works.

bryceml: If you got most of them, that would probably be enough to merge for now.  Mainly it's just misleading in your pr as it is now to make people think that logging in as root is the only way possible.

bryceml:
I haven't tested all the documentation I added yet, but if it's correct, it should be good enough to get this merged.

bryceml:
This should be ready now.

Akm0d (approved):

PR #55099, opened 2019-10-23
flat roster: pass master options to apply_sdb

jhass:
### What does this PR do?

Despite being more convenient and consistent, in avoiding to having to specify the sdb options
again for each host, `apply_sdb` (or actually `sdb_get` I think) also somehow manages to mess up
the previously passed options in the next iteration.

### Previous Behavior

Using sdb in the flat roster copied the entry of the current host into the key of the current one. It also required the sdb configuration to be present in each roster entry.

### New Behavior

Pass the master config available in `__opts__` to `apply_sdb`, centralizing the sdb configuration properly and keeping `sdb_get` from messing things up.

### Tests written?

is a bit beyond my experience with this project.

### DIFF
--- a/salt/roster/flat.py
+++ b/salt/roster/flat.py
@@ -32,5 +32,5 @@ def targets(tgt, tgt_type='glob', **kwargs):
                            **kwargs)
     conditioned_raw = {}
     for minion in raw:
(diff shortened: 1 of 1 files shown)


dwoz:
@jhass We are no longer accepting PRs to the `develop` branch. This needs to be re-opened against the `master` branch. Sorry for the confusion.

PR #55101, opened 2019-10-23
Enable Coverage Reporting on PRs

s0undt3ch:
### What does this PR do?
Enable Coverage Reporting on PRs
Additionally, it also enabled code coverage for tests, not just salt, so that we can start detecting dead code under `tests/`

### Previous Behavior
No coverage commented where added to PRs

### New Behavior
PR's will now have code coverage comments

### Tests written?
No - Unrelated to tests

### Commits signed with GPG?

Yes


waynew (approved):

cmcmarrow (approved):

Akm0d (approved):

s0undt3ch:
The reason why there's also a change in a unit test module is because those broke for the pycryptodome builds, dunno why since no dep's changed. Anyway, they now pass.

PR #55126, opened 2019-10-25
Respect argline passed into jinja renderer

### LINKED ISSUE #55124: Jinja renderer ignores argline
### Description of Issue

The Jinja renderer accepts only a single arg using the `argline` method: `-s`. This argline is designed to allow the renderer to be able to convey to `salt.utils.templates` that the source of the data being rendered is a string rather than a file-like object. The problem here is that while the jinja renderer inspects the argline, it does not relay the result of that inspection when rendering. The result of this is that the argline is ignored.

The reason an issue like this has been able to go unnoticed thus far is because there are very few use cases for it. Jinja is almost always at the _beginning_ of the render pipe, so it's nearly always receiving its input from a StringIO object. However, when one GPG-encrypts an SLS file and uses the GPG renderer to decrypt prior to rendering the jinja, rendering fails because the jinja renderer is expecting a StringIO or other file-like object.

### Steps to Reproduce Issue

**NOTE: I have a fix for this issue already and will be opening an accompanying PR once I have a testcase written. (EDIT: #55126)**

1. Set up GPG renderer
2. Create an SLS file with Jinja blocks in it
3. Encrypt that file using the key set up for the GPG renderer
4. Make sure to add a render pipeline of `#!gpg|jinja -s|yaml` at the beginning of the encrypted copy
5. Attempt to run that SLS file.

**OR**

From the root of a git checkout of salt, use the Docker image I've prepared, like so:

```bash
docker run --rm -it -v $PWD:/testing terminalmage/issues:55124 salt-call state.apply encrypted
```

### Expected Behavior

The SLS file executes and all is right with the world.

### Actual Behavior:

```
local:
    Data failed to compile:
----------
    Rendering SLS encrypted failed, render error: [Errno 20] Not a directory: u'/var/cache/salt/minion/files/base/encrypted.sls/#!gpg|jinja -s|yaml\n{#\n\n   To update the encrypted copy, run the following one-liner:\n\n   echo \'#!gpg|jinja -s|yaml\' >/srv/salt/encrypted.sls; cat /srv/salt/unencrypted.sls | gpg --homedir /etc/salt/gpgkeys --armor --batch --trust-model always --encrypt -r "GPG Renderer" >>/srv/salt/encrypted.sls\n\n#}\n\nDisplay current date:\n  cmd.run:\n    - name: echo {{ salt.cmd.run(\'date\') }}'
  File "/testing/salt/state.py", line 3481, in render_state
    rendered_sls=mods
  File "/testing/salt/template.py", line 101, in compile_template
    ret = render(input_data, saltenv, sls, **render_kwargs)
  File "/testing/salt/renderers/jinja.py", line 70, in render
    **kws)
  File "/testing/salt/utils/templates.py", line 164, in render_tmpl
    raise exc
IOError: [Errno 20] Not a directory: u'/var/cache/salt/minion/files/base/encrypted.sls/#!gpg|jinja -s|yaml\n{#\n\n   To update the encrypted copy, run the following one-liner:\n\n   echo \'#!gpg|jinja -s|yaml\' >/srv/salt/encrypted.sls; cat /srv/salt/unencrypted.sls | gpg --homedir /etc/salt/gpgkeys --armor --batch --trust-model always --encrypt -r "GPG Renderer" >>/srv/salt/encrypted.sls\n\n#}\n\nDisplay current date:\n  cmd.run:\n    - name: echo {{ salt.cmd.run(\'date\') }}'
```

### Versions Report

This affects all salt versions I've tested and has likely been hiding under-the-radar for years.

terminalmage:
Resolves #55124.


dwoz (dismissed):

terminalmage:
I just realized that I opened this PR against master. The goal is to have this fix in 2019.2.3, does it need to be rebased against a different branch for this?

tests/integration/modules/test_state.py
```diff
@@ -2256,6 +2256,18 @@ def test_state_sls_integer_name(self):
                          'Success!')
         self.assertTrue(state_run[state_id]['result'])
 
+
```
waynew: Lint says too many blank lines (should be 1)

waynew (changes requested):

waynew:
Looks like there are some possibly related test failures?

```
  File "/tmp/kitchen/testing/tests/unit/utils/test_jinja.py", line 1492, in test_normlookup
    ret = self.render(tmpl_str)
  File "/tmp/kitchen/testing/tests/unit/utils/test_jinja.py", line 1483, in render
    return self.jinja(tmpl_str, context=context or {}, from_str=True).read()
  File "/tmp/kitchen/testing/salt/renderers/jinja.py", line 71, in render
    **kws)
TypeError: render_tmpl() got multiple values for keyword argument 'from_str'
	
```

terminalmage:
The tests that were failing were failing because they were not designed properly. `from_str` is supposed to be derived from the argline, not passed in via the double-asterisk catch-all. I fixed the failing tests.

dwoz:
@terminalmage @waynew the windows failures look legit.

terminalmage:
@dwoz I pushed an update to the tests, and rebased to bring it up to date with master.

sagetherage:
@terminalmage looks like there are pending conflicts when you get time :) I'm  updating the labels to next release so that it isn't confusing

terminalmage:
Rebased again.

terminalmage:
Rebased yet again. It has been 5 1/2 months since this PR was opened. It's a one-line code change, and was submitted _with tests_.

Can we please get this re-re-re-reviewed and merged?

DmitryKuzmenko (approved):

dwoz (approved):

PR #55130, opened 2019-10-25
Pip test flaky

cmcmarrow:
### What does this PR do?
add @flaky to some test

### Previous Behavior
win 2019 py3 fails on install but works on second try. Something hangs or a recourse is not let go.

### Commits signed with GPG?
Yes


tests/integration/states/test_pip_state.py
```diff
@@ -335,6 +337,7 @@ def test_issue_6912_wrong_owner(self, temp_dir, username):
                     self.assertEqual(
                         salt.utils.win_dacl.get_owner(path), username)
 
+    @flaky
```
s0undt3ch: Instead of flaky, and if this is likely a problem with windows not releasing the file locks, let's set the venv_dir different from the previous test. Instead of `6912-wrong-owner`, let's make it `6912-wrong-owner-reqs-file`.
Although, the tempdir we get shouldn't be using the same path, but let's at least try that.

s0undt3ch (changes requested):
If like for us to start removing flaky and not adding.
If my suggestion leaves us nowhere and flaky does it, sure let's keep it.

rares-pop:
I'd also like to at least understand what's going on. From what I've seen, running the Windows CI is already over 6 hours. Let's try and at least not increase it.

Ideally would be to fix the issue, if there's an issue, or rewrite the test to remain valid but not to hang. Or worst, remove the test on Windows.

Issue #55136, opened 2019-10-26
smartos_imgadm cannot deal with orphan images


sjorge:
### Description of Issue
While testing 2019.2.2 I noticed that smartos_imgadm cannot deal with orphaned images due to them not having a 'name' key.

According to the SmartOS folks this should not happend unless you messed up with image creation. I'll try and write a fix for this to go into master.

### Setup
n/a

### Steps to Reproduce Issue
Have a SmartOS node with a broken image, rather hard to do unfortunately.

### Versions Report
```
Salt Version:
           Salt: 2019.2.2

Dependency Versions:
           cffi: 1.11.5
       cherrypy: 17.3.0
       dateutil: 2.7.3
      docker-py: Not Installed
          gitdb: 2.0.5
      gitpython: 2.1.11
          ioflo: Not Installed
         Jinja2: 2.10
        libgit2: Not Installed
        libnacl: 1.6.1
       M2Crypto: 0.30.1
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.5.6
   mysql-python: Not Installed
      pycparser: 2.19
       pycrypto: 3.6.6
   pycryptodome: Not Installed
         pygit2: Not Installed
         Python: 2.7.15 (default, Oct 23 2018, 17:11:49)
   python-gnupg: 0.4.3
         PyYAML: 3.13
          PyZMQ: 17.1.2
           RAET: Not Installed
          smmap: 0.9.0
        timelib: Not Installed
        Tornado: 5.1.1
            ZMQ: 4.2.5

System Versions:
           dist:
         locale: UTF-8
        machine: i86pc
        release: 5.11
         system: SunOS
        version: Not Installed
```

sjorge:
Extra data to help with writing a unit test, output of 'imgadm list -j' with a broken image:

```json
[
  {
    "manifest": {
      "v": 2,
      "uuid": "05140a7e-279f-11e6-aedf-47d4f69d2887",
      "owner": "00000000-0000-0000-0000-000000000000",
      "name": "ubuntu-16.04",
      "version": "20160601",
      "state": "active",
      "disabled": false,
      "public": true,
      "published_at": "2016-06-01T02:17:41Z",
      "type": "lx-dataset",
      "os": "linux",
      "files": [
        {
          "sha1": "d342f137c5ccef0702ec479acb63c196cf81b38a",
          "size": 134969110,
          "compression": "gzip"
        }
      ],
[...]

PR #55137, opened 2019-10-26
Fix smartos_imgadm doesn't handle orphan images

sjorge:
### What does this PR do?
Update `_parse_image_meta` so that it handles orphan images.
Additionally we also pull in the changes from #51563 to master as those are also needed.

#55136 

### Previous Behavior
`_parse_image_meta`  would not handle orphan image data e.g.
```json
{
  "manifest": {
    "uuid": "07f360fd-12d5-e624-a279-eb8a15b630f6"
  },
  "zpool": "zones",
  "cloneNames": [],
  "clones": 0
}
```

### New Behavior
`_parse_image_meta` will now check if the 'name' field is present.

### Tests written?
Yes, there were no tests for smartos_imgadm yet. Tests for `_parse_image_meta` were added that feed is various kinds of image data. Currently those are 'orphan', 'native', 'lx', and 'zvol'.

Those cover the currently available image types available in SmartOS

### Commits signed with GPG?
No


sjorge:
I'm not sure how to fix the ci/py3 failures?

sjorge:
@Ch3LL aside from a rebase, anything blocking this? CI seems to always fail for a bunch of distros but it are just unit tests which is weird as they work fine locally and on some targets?

sjorge:
@Ch3LL bump

dwoz (approved):

sjorge:
Thanks for merging!

PR #55144, opened 2019-10-28
Add function to format powershell paths in utils.powershell module

rmarcinik:
The new function here fixes a cornercase where utils.powershell.get_modules() doesnt handle environment variables in paths. There is logic to test the paths before adding them to root_paths and os.path.exists() will return False when the path contains a variable like %SystemRoot%. If get_modules() returns nothing then modules like win_pki will be unavailable because the __virtual__ function will return False.


dwoz:
@rmarcinik We are no longer accepting PRs to the `develop` branch. This needs to be re-opened against the `master` branch. Sorry for the confusion.

rmarcinik:
It's my mistake, I think I was trying to make a draft request and submitted it by accident.

PR #55132, opened 2019-10-26
Added Status Code Lists for HTTP Query State

Ajnbro:
### What does this PR do?
This PR allows for individuals to pass a list of strings representing multiple different status codes to the HTTP Query State in order to check if any of the provided status codes are returned by the query. It also creates two additional tests for the HTTP Query State, one for testing this new functionality and one for testing prior functionality.

### What issues does this PR fix or reference?
N/A

### Previous Behavior
HTTP Query only allowed individuals to specify the status parameter as a string representing a status code or as a regex used to search for specific status codes.

### New Behavior
HTTP Query allows individuals to specify the status parameter as a string representing a status code, a list of strings representing multiple status codes, or as a regex used to search for specific status codes.

### Tests written?
Yes

### Commits signed with GPG?
Yes

PR #55145, opened 2019-10-28
Added Status Code Lists for HTTP Query State

Ajnbro:
### What does this PR do?
This PR allows for individuals to pass a list of strings representing multiple different status codes to the HTTP State for querying in order to check if any of the provided status codes are returned by the query. It also creates two additional tests for the HTTP Query State, one for testing this new functionality and one for testing prior functionality. Repost of #55132 because I had opened the PR against the develop branch initially.

N/A

### Previous Behavior
HTTP Query only allowed individuals to specify the status parameter as a string representing a status code or as a regex used to search for specific status codes.

### New Behavior
HTTP Query allows individuals to specify the status parameter as a string representing a status code, a list of strings representing multiple status codes, or as a regex used to search for specific status codes.

### Tests written?
Yes

### Commits signed with GPG?
Yes


dwoz (approved):
@Ajnbro This would be a good case to start refactoring things into smaller checks. If the match and status checks were individual methods, they could be more easily unit tested without having to mock the http response. If you have some time to dedicated to this, it would be fantastic.

Otherwise, approved for now.

PR #55148, opened 2019-10-29
[WIP] Warn that the "py" renderer can't be preceded by any other renderer

max-arnold:
### What does this PR do?

Add a warning that any renderer that comes before the "py" one is ignored.

### Previous Behavior

The `jinja` part in `jinja|py` renderer pipeline is silently ignored. This happens because the `py` renderer can only handle files:

https://github.com/saltstack/salt/blob/0ac7520c670a8d77b479355c38c19678a85a59ce/salt/renderers/py.py#L129
https://github.com/saltstack/salt/blob/0ac7520c670a8d77b479355c38c19678a85a59ce/salt/utils/templates.py#L535

### New Behavior

A warning has been added:

`The "py" renderer ignores previously processed results. The "#!jinja|py" pipeline would be equivalent to just "#!py"`

### Tests written?

Yes

### Commits signed with GPG?

No

### DIFF
--- a/salt/template.py
+++ b/salt/template.py
@@ -232,6 +232,13 @@ def check_render_pipe_str(pipestr, renderers, blacklist, whitelist):
     try:
         if parts[0] == pipestr and pipestr in OLD_STYLE_RENDERERS:
             parts = OLD_STYLE_RENDERERS[pipestr].split("|")
+        if "py" in parts[1:]:
+            log.warning(
+                'The "py" renderer ignores previously processed results. '
+                'The "#!%s" pipeline would be equivalent to just "#!%s"',
+                pipestr,
+                "|".join(parts[parts.index("py") :]),
+            )
         for part in parts:
             name, argline = (part + " ").split(" ", 1)
             if whitelist and name not in whitelist or blacklist and name in blacklist:
--- a/tests/unit/test_template.py
+++ b/tests/unit/test_template.py
@@ -9,7 +9,7 @@ from __future__ import absolute_import, print_function, unicode_literals
 # Import Salt libs
 from salt import template
 from salt.ext.six.moves import StringIO
-from tests.support.mock import MagicMock
+from tests.support.mock import MagicMock, patch
 
 # Import Salt Testing libs
 from tests.support.unit import TestCase
@@ -21,6 +21,7 @@ class TemplateTestCase(TestCase):
         "jinja": "fake_jinja_func",
         "json": "fake_json_func",
         "mako": "fake_make_func",
+        "py": "fake_py_func",
     }
 
     def test_compile_template_bad_type(self):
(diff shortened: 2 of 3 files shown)


salt/template.py
```diff
@@ -221,6 +221,12 @@ def check_render_pipe_str(pipestr, renderers, blacklist, whitelist):
     try:
         if parts[0] == pipestr and pipestr in OLD_STYLE_RENDERERS:
             parts = OLD_STYLE_RENDERERS[pipestr].split('|')
+        if len(parts) > 1 and 'py' in parts[1:]:
```
waynew: Slices don't raise an IndexError, you can simplify this to just
```suggestion
        if 'py' in parts[1:]:
```

waynew (dismissed):
Love it - just one suggestion to clean things up. Would it also make sense to add a `.. note::` to the renderer docs in the `py` section? Like, in https://docs.saltstack.com/en/latest/ref/renderers/all/salt.renderers.py.html#module-salt.renderers.py ?

max-arnold:
@waynew Both suggestions were adressed

max-arnold:
This is obviously a very minor usability fix (I've seen exactly two people who spent some time trying to understand why `jinja|py` doesn't work; one of them was me). However, there are several (mostly rhetorical) points to add:

1. The proper solution is to fix the `py` renderer, so it works the same way as any other renderer (i.e. can take a string as input)
2. There are no existing tests for the `py` renderer, so the entry barrier to fix it is higher (as per SEP10)
3. This PR contains good example of a test that is not an interface spec (it just documents the currently broken behavior)
4. The test could be safely removed to keep the test suite lean (i.e., the fix is an one-liner)

max-arnold:
It looks like the https://github.com/saltstack/salt/pull/55390 makes my PR unnecessary!

tests/unit/test_template.py
```diff
@@ -113,3 +114,31 @@ def test_check_renderer_blacklisting(self):
...
         )
         self.assertListEqual([("fake_json_func", "")], ret)
+
+    def test_check_renderer_py_warning(self):
+        """
+        Check that warning is emitted if "py" has preceding renderers.
+        """
+        warn = "ignores previously processed results"
+        mock_log = MagicMock()
+        with patch.object(template.log, "warning", mock_log):
```
waynew, line 124: Not sure, but would it be possible to dynamically grab the list of renderers from somewhere within Salt? I think we should at minimum show jinja/json (or, assert that there are at least one renderer in the list), but that would ensure that we're testing any potential future renderers.

If that's not possible with a reasonable amount of effort, then feel free to ignore the suggestion :slightly_smiling_face:

max-arnold: @waynew Can you help to improve this PR instead (it supersedes mine)? https://github.com/saltstack/salt/pull/55390

waynew: Should we go ahead and close this one then?

max-arnold: I’ll close mine once the other one is merged. That is why I set the title to WIP

dwoz:
@waynew We good here now?

waynew:
@dwoz I think that's entirely up to @max-arnold (and the other PR) - as he [mentioned](https://github.com/saltstack/salt/pull/55148#discussion_r408939919) there's another PR that actually makes it so py renderer *can* be be preceded by other renderers, so this warning won't be necessary.

So if that one gets done & merged, this one will no longer be necessary :)

Issue #55152, opened 2019-10-29
file.get_selinux_context incorrect works with directories


alexminder:
### Description of Issue
`file.get_selinux_context` incorrect works with directory as path.

### Steps to Reproduce Issue
```
# salt-call --local file.get_selinux_context /run --out=json
14:50:25 [INFO    ] Executing command ['ls', '-Z', '/run'] in directory '/root' [salt.loaded.int.module.cmdmod]
{
    "local": "system_u:object_r:getty_runtime_t:s0"
}
```
Try to manualy lookup:
```
# ls -Z /run|head -1
      system_u:object_r:getty_runtime_t:s0 agetty.reload

# ls -dZ /run
system_u:object_r:var_run_t:s0 /run
```
As you can see the `file.get_selinux_context` returns selinux context for first file in directory. It is wrong.

### Versions Report
2019.2.2

sagetherage:
PR merged and will be released in Sodium mid-June 2020

PR #55153, opened 2019-10-29
file.get_selinux_context: fix directory path behaviour

alexminder:
### What does this PR do?

Fix directory path behaviour for module `file.get_selinux_context`

issue #55152


waynew:
@alexminder Thanks for the PR!

Looks like there aren't any tests written here. Could you add tests to verify that this is working correctly?

alexminder:
> @alexminder Thanks for the PR!
> 
> Looks like there aren't any tests written here. Could you add tests to verify that this is working correctly?

There are no new functionality in this PR. I don't know what tests do you want.

waynew:
In the case of this change the minimum is to verify that `cmd.run` is called with the correct arguments.

Yes, it's a minimal test, but it still protects against regressions, and is worth writing.

waynew:
It looks like this was also fixed in a different way in https://github.com/saltstack/salt/pull/56558/files which also had tests, so I'm going to go ahead and close this PR

Issue #55057, opened 2019-10-18
Stalekey event fails to delete keys for multiple minions


jayshah11:
### Description of Issue

The event fails to keys of multiple minions that have not checked in for a configured interval. It successfully deletes the key if one minion has not checked in but reports the following error in case of multiple minions. 

```
[salt.utils.event :742 ][DEBUG   ][781] Sending event: tag = salt/wheel/20191018132606446871/ret; data = {u'fun_args': [u'saltminion1', u'saltminion2'], u'jid': u'20191018132606446871', u'return': u"\nPassed invalid arguments: delete() takes exactly 1 argument (2 given)\n\nUsage:\n\n    Delete keys based on a glob match. Returns a dictionary.\n\n    match\n        The glob match of keys to delete.\n\n    .. code-block:: python\n\n        >>> wheel.cmd_async({'fun': 'key.delete', 'match': 'minion1'})\n        {'jid': '20160826201244808521', 'tag': 'salt/wheel/20160826201244808521'}\n    ", u'success': True, u'_stamp': '2019-10-18T13:26:06.450593', u'user': u'UNKNOWN', u'fun': u'wheel.key.delete'}
```

### Versions Report
```
Salt Version:
           Salt: 2019.2.0

Dependency Versions:
           cffi: 1.6.0
       cherrypy: 3.2.2
       dateutil: 1.5
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.7.2
        libgit2: 0.26.3
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.4.6
   mysql-python: Not Installed
      pycparser: 2.14
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: 0.26.4
         Python: 2.7.5 (default, Jun 11 2019, 14:33:56)
   python-gnupg: Not Installed
         PyYAML: 3.11
          PyZMQ: 15.3.0
           RAET: Not Installed
          smmap: Not Installed
        timelib: Not Installed
        Tornado: 4.2.1
            ZMQ: 4.1.4
```

garethgreenaway:
@jayshah11 Just to confirm, you're using the `stalekeys` Salt engine?

jayshah11:
[...]

PR #55168, opened 2019-10-30
[master] stalekey engine fixes

garethgreenaway:
### What does this PR do?

Fixing stalekey engine so it deletes the keys when they are a list.

Breaking out various functionality so it's easier to test.

Writing tests.

#55057 

### Tests written?
Yes
### Commits signed with GPG?
Yes


salt/engines/stalekey.py
```diff
@@ -39,41 +39,103 @@
...
+    '''
+    Indicate if engine should continue running
+    Useful when running tests and we want to ensure
+    the engine only makes one pass
+    '''
+    return True
+
+
 def start(interval=3600, expire=604800):
+    '''
```
s0undt3ch: I'm going to pull a @dwoz on you :)
For reference, https://github.com/saltstack/salt/pull/55047#discussion_r336696340

Let's make start accept `_check_running_callable=None`

Then the `while` code Bellow becomes:
```python
while True:
    # do stuff
    If callable(_check_running_callable) and _check_running_callable() is False:
        break
```

No?

s0undt3ch (changes requested):
You need to pull master and rebase too.

salt/engines/stalekey.py
garethgreenaway: Not against the idea, just curious, what is the advantage?

garethgreenaway: Actually the `_running` function was being used when I was attempting to run `start` from tests, which didn't work out.  So I can remove that.

s0undt3ch: Not having a function for us just to patch during tests, ie, dependency injection.

s0undt3ch (approved):

dwoz (approved):

tests/unit/engines/test_stalekey.py
```diff
@@ -0,0 +1,93 @@
...
+
+import logging
+import msgpack
+
+# Import Salt Testing Libs
+from tests.support.mixins import LoaderModuleMockMixin
+from tests.support.unit import skipIf, TestCase
+from tests.support.mock import (
+    NO_MOCK,
+    NO_MOCK_REASON,
```
waynew: Gonna need to remove `NO_MOCK` and `NO_MOCK_REASON` here.

tests/unit/engines/test_stalekey.py
```diff
@@ -0,0 +1,93 @@
...
+
+class MockKey(object):
+    def __init__(self, opts):
+        pass
+
+    def delete_key(self, key):
+        pass
+
+
+@skipIf(NO_MOCK, NO_MOCK_REASON)
```
waynew: This no mock skip should also be removed

PR #55183, opened 2019-11-01
Fix PY3 dict_keys TypeError in modules/capirca_acl

basdusee:
### What does this PR do?
It fixes a PY3 incompatibility in modules/capirca_acl.py.

In modules/capirca_acl.py there are some places where the first key of a dict is referenced.
Like so: dict.keys()[0]
This works in PY2 because .keys() gives back a list.
In PY3 .keys() returns a dict_keys object, which can't be indexed

### Previous Behavior
modules/capirca_acl.py is throwing:
TypeError: 'dict_keys' object does not support indexing
for the lines changed in this commit.

### New Behavior
Sunshine, Unicorns and Rainbows. modules/capirca_acl.py is working on PY3 installation.

### Tests written?
No. Small bug/typo. 

### Commits signed with GPG?
No

### DIFF
--- a/salt/modules/capirca_acl.py
+++ b/salt/modules/capirca_acl.py
@@ -401,8 +401,8 @@ def _lookup_element(lst, key):
     for ele in lst:
         if not ele or not isinstance(ele, dict):
             continue
-        if ele.keys()[0] == key:
-            return ele.values()[0]
+        if list(ele.keys())[0] == key:
+            return list(ele.values())[0]
     return {}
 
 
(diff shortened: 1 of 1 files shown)


basdusee:
Wait, I have some indications that this fix actually induce a (or: more) bugs. I need to check that out first. Source_address is not processed in capirca_acl at the moment. This could be a Caprica problem, or a Saltstack problem. First need to sort that out.

basdusee:
Closing this request for now due to all the testing errors and the doubtfullness if it will be fixed properly (PY3 code runs after patching, but has strange behaviour).

PR #49891, opened 2018-10-03
Add salt_version execution module

rallytime:
### What does this PR do?
This execution module exposes Salt's release codenames as functions that can be used to make comparison's to the minion's Salt version.

This is useful when writing states that handle deprecations. By exposing the release codenames, people can use these comparisons in Jinja statements in state files. This is only one example and the use case I wanted to cover here. There may be more that I haven't thought of, but this is a start!

### What issues does this PR fix or reference?
None that I know of.

### Tests written?

Yes

### Commits signed with GPG?

Yes

Please review [Salt's Contributing Guide](https://docs.saltstack.com/en/latest/topics/development/contributing.html) for best practices.

See GitHub's [page on GPG signing](https://help.github.com/articles/signing-commits-using-gpg/) for more information about signing commits with GPG.

PR #55195, opened 2019-11-04
[master] Port #49891 (salt_version module)

max-arnold:
Master backport of #49891 (`salt_version` execution module). 

Also, fixed the `LNAMES` mocks (indexes should be lowercase, the mocked tests passed by accident) and changed the function names and semantics:

2019-11-05 at 00 06


salt/modules/salt_version.py
```diff
@@ -0,0 +1,176 @@
...
+Access Salt's elemental release code-names.
+
+.. versionadded:: Neon
+
+Salt's feature release schedule is based on the Periodic Table, as described
+in the :ref:`Version Numbers <version-numbers>` documentation.
+
+Since deprecation notices often use the elemental release code-name when warning
+users about deprecated changes, it can be difficult to build out future-proof
+functionality that are dependent on a naming scheme that moves.
```
waynew: Since we're just moving to a 1-year (minimum) deprecation cycle, we can probably remove this paragraph. The next paragraph may need some tweaking to make it flow better.

salt/modules/salt_version.py
```diff
@@ -0,0 +1,176 @@
...
+    '''
+    Only work on POSIX-like systems
+    '''
+    return __virtualname__
+
+
+def get_release_number(name):
+    '''
+    Returns the release number of a given release code name in a
+    ``<year>.<month>`` context.
```
waynew: Probably should just change this to `Returns the release number of a given codename`.

waynew (changes requested):
The code looks good, but with the change to 3000-based versions, we probably should update the docs.

salt/modules/salt_version.py
max-arnold: Quotes from the SEP:

> Our plan is to have a typical release cadence of 3 months. We expect to have an intermediate release cadence of 4 months.

> With the increased release cadence in order to allow proper planning for feature deprecation, Salt will introduce a minimum 1-year, expected 3 major release cycle. If `neon` ships in January 2020, any new deprecations will not be removed until January 2021 at the earliest.

I do not see why the paragraph in question is no longer relevant.

salt/modules/salt_version.py
max-arnold: I changed it to

> Returns the release number of a given release codename in a ``MAJOR.PATCH`` format.

in line with the SEP:

> With the `neon` release, to indicate the new change in release process, Salt will change to a new, non-date based version schema beginning at 3000. The version will be MAJOR.PATCH. For a planned release containing features and/or bug fixes the MAJOR version will be incremented

salt/modules/salt_version.py
max-arnold: @waynew I rephrased the paragraph

max-arnold:
@waynew Does the tweaked first paragraph look good?

dwoz (approved):

Issue #54906, opened 2019-10-06
Question: salt --async emitting jid and methods of capturing it


arizvisa:
### Description of Issue
This is not an issue, but probably a question that leads towards a feature request. When using the salt client with the `--async` (or the `--show-jid`) option, extra text is emitted in order to describe to the caller what is being returned. As an example, using the --async parameter the following message is emitted.
```
$ salt --async $id test.ping
Executed command with job ID: 20191006185730665675
```

If one wants to capture the job id from their shell, this extra text will need to be culled out through some means. For some shells, such as on a bourne-like shell they'll need to combine it with other commands like in the following:
```
$ jid=`salt --async $id test.ping | cut -d: -f2 | tr -d' '`
```

If using the windows command prompt, one'll probably use something similar to the following to extract the job id.
```
> for /f "tokens=2 delims=:" %T in ('salt --async %id% test.ping') do @set jid=%T
```

Then Powershell which might be considered simpler by some...
```
PS> (salt --async $id test.ping).Split(":")[-1].Trim() | Set-Variable jid
```

However, If one uses the `--show-jid` option, the `--out=quiet` parameter can be passed as in the following which makes capturing a little simpler perhaps.
```
$ salt --out=quiet --show-jid $id test.ping
jid: 20191006193152611760
```

One minor issue of `--show-jid` is that the job id is being emitted in a hardcoded format (can't convert to json/yaml/etc. w/ an outputter). Ftr the format of this is at `salt/client/__init__.py:990`. The primary issue with this, however, is that it appears that this format is unable to be combined with `--async` in order to run a job in the background which makes it not possible for saving the jid for a long-running job.

So...Is there some way to make the salt client emit just the job id and nothing else? If not, would you guys accept a PR for something like a `-j` option that emits just the job-id? This way scripters can simply just assign the job id for a job to a variable (generally) across platforms.

Rather than emitting the job-id by considering another parameter, perhaps the regular output of `--async` could be split across stdout+stderr so its simpler to capture the job id by discarding stderr. 

Or perhaps another (more consistent?) way would be to feed the output of the job id from `--async` through the outputter that was specified. This way one can include a minimalistic outputter that emits just the jid, and specify `--async --out=my-jid-emitter` if they desire that particular capability.

Ftr, the line of code that emits the async job id is at `salt/cli/salt.py:149`.

### Setup
Not really relevant.

### Steps to Reproduce Issue
Please see description.

### Versions Report
Since 2015.5

[...]

PR #55199, opened 2019-11-05
Split up the --async option's emitting of the jid into two flushing writes to stderr and stdout

arizvisa:
### What does this PR do?
This splits up the emitting of the jid when passing the `--async` option to the `salt` command. This allows the user to use their shell to immediately capture the jid without needing to do any extra shell gymnastics (which is a pita on Windows).

This simplifies the method to capture the jid in a bourne-compliant shell:
```
$ jid=`salt \* test.ping`
...
$ salt-run jobs.print_job $jid
$ salt-run state.event "tagmatch=$jid"
```

..and in the windows command-prompt:
```
C> for /f %j in ('salt * test.ping') do set jid=%j
...
C> salt-run jobs.print_job %jid%
C> salt-run sate.event tagmatch=%jid%
```

This closes issue #54906.

### Previous Behavior
Prior to this the `salt.stringutils.print_cli` command was used to emit the jid which is used to prevent scrollback with `less`. This is unnecessary because when using the `--async` option, only one line is emitted before terminating and so someone would literally need to explicitly close the fd, or use explicitly use a non-linebuffered fd to interrupt less.

### New Behavior
The emitting of the jid is split between two writes. The first is to stderr for the "Executed command with job ID:" message, and the second is the jid being written to stdout. Both writes are being flushed so that the behaviour appears the same, but when explicitly using i/o redirection to capture the jid only the identifier will be caught.

### Tests written?
No. The previous test should still work.

### Commits signed with GPG?
No.

### DIFF
--- a/salt/cli/salt.py
+++ b/salt/cli/salt.py
@@ -148,7 +148,8 @@ class SaltCMD(salt.utils.parsers.SaltCMDOptionParser):
 
         if self.config['async']:
             jid = self.local_client.cmd_async(**kwargs)
-            salt.utils.stringutils.print_cli('Executed command with job ID: {0}'.format(jid))
+            six.print_('Executed command with job ID:', end=' ', file=sys.stderr, flush=True)
+            six.print_('{0}'.format(jid), end='\n', file=sys.stdout, flush=True)
             return
 
         # local will be None when there was an error


dwoz:

arizvisa:
What was the issue with this anyways, it just needed a maintainer to review it and nobody wanted to?

dwoz:
@arizvisa it may have been closed in error. If you want to get it, re-open it with changelog and test. I'll review it.

PR #55206, opened 2019-11-05
Update to more current SLE version numbers

brejoc:
### What does this PR do?

SLE 15 and SLE 15 SP1 are more current. Also those versions ship with Salt in the base repos. No need to add additional repos. Adding additional repos would actually be harmful, since those would have skipped the openQA checks.


brejoc:
These two PRs are related: https://github.com/saltstack/salt/pull/55205

brejoc:
@thatch45 Could you please take a look here? Thanks!

brejoc:
@dwoz Could you please have a look!?

DmitryKuzmenko (approved):
I see no problem. SUSE guys know better how to install Salt on SUSE =)

PR #55216, opened 2019-11-06
Checking for jid before returning data

brejoc:
### What does this PR do?

Seems raw can have returns for the same minion, but an other job. In order to not return resutls from the wrong job, we need to check for the jid.

#50927

This is the successor of https://github.com/saltstack/salt/pull/52459.

### Previous Behavior
Wrong job results might be returned. Please take a look at #50927

### New Behavior
Only the correct jobs results should be returned.

### Tests written?

Yes

### Commits signed with GPG?

Yes


brejoc:
@Ch3LL Could you please take a look here? https://github.com/saltstack/salt/pull/52459#issuecomment-550340020
Thanks!

Ch3LL (approved):
not sure what i did wrong previously, but when i patch my local setup i dont see the issue anymore :) thanks for this

Ch3LL:
ping @brejoc looks like there is a conflict. mind resolving that? if not I can help out

brejoc:
@Ch3LL Thanks for looking into this again!

Seems like the conflicting imports where just removed, so I did the same here. Conflict is solved. Let's see if the tests are still passing for this.

brejoc:
@dwoz Thanks for taking care of this! :bowing_man:

Issue #52503, opened 2019-04-11
loop.until state cannot run without kwargs


fattybenji:
### Setup
version v2018.3.2

### Steps to Reproduce Issue
Run something like this :

```
Wait for something:
  loop_until:
    - name: test.sleep
    - condition: number_of_whatever == n
    - args:
      - 5
```
And it should break, saying that kwargs cannot be NoneType. In that case, test.sleep only has one arg, "length". Same when we only give it as kwarg :

```
     Comment: An exception occurred in this state: Traceback (most recent call last):
                File "/usr/lib/python2.7/dist-packages/salt/state.py", line 1913, in call
                  cdata['kwargs'])
                File "/usr/lib/python2.7/dist-packages/salt/loader.py", line 1898, in wrapper
                  return f(*args, kwargs)
                File "/usr/lib/python2.7/dist-packages/salt/states/loop.py", line 106, in until
                  m_ret = saltname
              TypeError: sleep() argument after ** must be a mapping, not NoneType
     Started: 18:02:06.437522
    Duration: 9.991 ms
```

I think it's because the loop.until state doesn't test null args or kwargs. I think it has not been modified in later version. We could maybe test whether args or kwargs are null and adapt the call, or create a default value...

If no one has the time I can probably give it a shot (if someone agrees on an implementation)

### Versions Report

```
Salt Version:
           Salt: 2018.3.2
 
Dependency Versions:
           cffi: 1.6.0
       cherrypy: unknown
       dateutil: 1.5
      docker-py: Not Installed
          gitdb: 0.6.4
      gitpython: 1.0.1
          ioflo: Not Installed
         Jinja2: 2.7.2
        libgit2: 0.26.3
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.4.8
   mysql-python: Not Installed
      pycparser: 2.14
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: 0.26.4
         Python: 2.7.5 (default, Aug  4 2017, 00:39:18)
   python-gnupg: Not Installed
         PyYAML: 3.11
          PyZMQ: 15.3.0
           RAET: Not Installed
          smmap: 0.9.0
        timelib: Not Installed
        Tornado: 4.2.1
            ZMQ: 4.1.4
 
System Versions:
           dist: centos 7.4.1708 Core
         locale: UTF-8
        machine: x86_64
[...]

PR #55227, opened 2019-11-07
Fix case where args or kwargs are not specified

fattybenji:

### What does this PR do?

Integrate this fix commit in the main 2018.3 branch. Not sure this is the right process, see #52503 for more information.

#52503 

Not 100% sure about the process there, I am completely open to criticism and suggestions.


dwoz:
@fattybenji This needs to be re-opened against the `master` branch. We are no longer accepting PRs to `2018.3`. Sorry for the confusion.

PR #55235, opened 2019-11-08
Don't set path explicitly

bryceml:
### What does this PR do?
This is to check that we can skip specifically setting the path so we can move to ubuntu jenkins slaves.  It will not get merged but is just a test.


s0undt3ch (approved):
It works!

PR #55254, opened 2019-11-11
Update to net.py and bgp.py

aegiacometti:
### What does this PR do?
Redirect to start using new "napalm" module instead of "napalm-base"

### Previous Behavior
sudo salt-run net.find 1.1.1.1
'net' __virtual__ False: The napalm-base module could not be imported

### New Behavior
Return the findings


aegiacometti:
Added the modififications at another PR

PR #55257, opened 2019-11-11
Strip trailing "/" from repo.uri when comparing repos in apktpkg.mod_repo

mateiw:
### What does this PR do?

Port of #51654 to master.

None.

### Previous Behavior
`salt-call --local pkg.mod_repo repo="deb http://es.archive.ubuntu.com/ubuntu xenial main restricted" ...`

did not find the target repo if it had a trailing `/` in the URI because repo URIs were compared 
verbatim.
E.g. from `/etc/apt/sorces.list` on Ubuntu 16.04:
```
deb http://es.archive.ubuntu.com/ubuntu/ xenial restricted main
```

### New Behavior
To be consistent with the rest of functions from `aptpkg` the trailing `/` are removed when comparing repos URIs in `mod_repo`.

### Tests written?

No.

### Commits signed with GPG?

No.

### DIFF
--- a/salt/modules/aptpkg.py
+++ b/salt/modules/aptpkg.py
@@ -2438,7 +2438,7 @@ def mod_repo(repo, saltenv="base", **kwargs):
         # has already been modified on a previous run.
         repo_matches = (
             source.type == repo_type
-            and source.uri == repo_uri
+            and source.uri.rstrip('/') == repo_uri.rstrip('/')
             and source.dist == repo_dist
         )
         kw_matches = source.dist == kw_dist and source.type == kw_type
--- a/tests/unit/modules/test_aptpkg.py
+++ b/tests/unit/modules/test_aptpkg.py
@@ -155,14 +155,16 @@ def _get_uri(repo):
 
 
 class MockSourceEntry:
-    def __init__(self, uri, source_type, line, invalid, file=None):
+    def __init__(self, uri, source_type, line, invalid, dist="", file=None):
         self.uri = uri
         self.type = source_type
         self.line = line
         self.invalid = invalid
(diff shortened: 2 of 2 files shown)


dwoz:
@mateiw are you able to write a unit test to cover this change?

mateiw:
@dwoz I'm afraid writing a unit test is complicated because `sourceslist.SourcesList` needs to be mocked but only a very old version of `python-apt` exists in PyPI.

Installing `python-apt` from PyPI throws:
```
. . . .
      File "<string>", line 1, in <module>
      File "/tmp/pip-install-n6uwdr7k/python-apt/setup.py", line 6, in <module>
        from DistUtilsExtra.command import *
    ModuleNotFoundError: No module named 'DistUtilsExtra'
```
More about the problem here https://github.com/sdispater/poetry/issues/1363

mateiw:
@dwoz do you see any alternative to mocking `sourceslist.SourcesList` ? Or is there any chance of relaxing a bit the rules in this case and merge it ?

cmcmarrow:
@mateiw All merges must come with at lest a test "Unless Doc" because this has burn us before. I know it can be very annoying when the change seems simple. Mocking sourceslist.SourcesList is your best option.

cmcmarrow:
@mateiw we cant felax the rules. I also don’t see any alternative to mocking `sourceslist.SourcesList`. You only need to test that `//` was striped.

mateiw:
@cmcmarrow I've added a unit test to test that `/` is striped.

meaksh:
This PR can be closed in favor of https://github.com/saltstack/salt/pull/58962

PR #55258, opened 2019-11-11
Make aptpkg.info return only installed packages

mateiw:
### What does this PR do?

Port of #51576 to master.

None.

### Previous Behavior
`pkg.info_instaled` returned also packages in status `rc` or `pn`.

### New Behavior
`pkg.info_instaled` returnes only packages in `ii` or `hi`.

### Tests written?

Yes

### Commits signed with GPG?

No

### DIFF
--- a/tests/unit/modules/test_aptpkg.py
+++ b/tests/unit/modules/test_aptpkg.py
@@ -90,7 +90,26 @@ LOWPKG_INFO = {
         'name': 'wget',
         'section': 'web',
         'source': 'wget',
-        'version': '1.15-1ubuntu1.14.04.2'
+        'version': '1.15-1ubuntu1.14.04.2',
+        'status': 'ii',
+    },
+    'apache2': {
+        'architecture': 'amd64',
+        'description': """Apache HTTP Server
+ The Apache HTTP Server Project's goal is to build a secure, efficient and
+ extensible HTTP server as standards-compliant open source software. The
+ result has long been the number one web server on the Internet.
+ .
+ Installing this package results in a full installation, including the
+ configuration files, init scripts and support scripts.""",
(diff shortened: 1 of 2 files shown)


tests/unit/modules/test_aptpkg.py
```diff
@@ -90,7 +90,26 @@
         'name': 'wget',
         'section': 'web',
         'source': 'wget',
-        'version': '1.15-1ubuntu1.14.04.2'
+        'version': '1.15-1ubuntu1.14.04.2',
+        'status': 'ii'
```
waynew: Nitpick, but if you make this 
```suggestion
        'status': 'ii',
```
then if something else has to be added, it's only one line changed instead of two.

tests/unit/modules/test_aptpkg.py
```diff
@@ -90,7 +90,26 @@
...
+ Installing this package results in a full installation, including the
+ configuration files, init scripts and support scripts.""",
+        'homepage': 'http://httpd.apache.org/',
+        'install_date': '2016-08-30T22:20:15Z',
+        'maintainer': 'Ubuntu Developers <ubuntu-devel-discuss@lists.ubuntu.com>',
+        'name': 'apache2',
+        'section': 'httpd',
+        'source': 'apache2',
+        'version': '2.4.18-2ubuntu3.9',
+        'status': 'rc'
```
waynew: Same here:
```suggestion
        'status': 'rc',
```

waynew (commented):
Just some minor picks, otherwise this seems reasonable.

mateiw:
@waynew thanks for the suggestions, I've updated the unit test.

dwoz (approved):

PR #55260, opened 2019-11-11
Only look at local packages when removing from a venv

danfoster:
### What does this PR do?

Corrects the behaviour of the `pip.absent` state when using a virtualenv with global access.

If a virtualenv was created with:
```
virtualenv --system-site-packages /venv/
```
Then you a python package in the system location and the virtualenv. For example, consider `paramiko`
```
yum install python-paramiko # Install to the global site-packages
/venv/bin/pip install paramiko # Install to the venv
```

When using the `pip.absent` state with this virtualenv, it will always mark the state as changed, even after it uninstalls the package after the first run:

```
remove package in venv:
  pip.removed:
    - name: paramiko
    - bin_env: /venv
```

This is because the call to `pip.list` returns pip packages from both the global and local spaces.

This PR adds the `local` attribute to the `pip.list` and `pip.freeze` execution modules, which matches the `pip -l` option:
```
-l, --local                 If in a virtualenv that has global access, do not output globally-installed packages.
```

The `pip.absent` state then sets `local=True` as it should always be operating on the local package list only.

N/A

### Tests written?

Unit tests written for the  `pip.list` and `pip.freeze` execution modules.
No tests written for the `pip.absent` state. It feels like only integration tests would be useful here and I'm not sure how to go about this. Could anyone help?

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/modules/pip.py
+++ b/salt/modules/pip.py
@@ -1045,6 +1045,7 @@ def freeze(bin_env=None,
            cwd=None,
            use_vt=False,
            env_vars=None,
+           local=False,
            **kwargs):
     '''
     Return a list of installed packages either globally or in the specified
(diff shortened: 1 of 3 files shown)


dwoz:
@danfoster This needs to be re-opened against the `master` branch. We are no longer accepting PRs to the `2019.2`. If this fixes a bug in a `2019.2.x` release, please make note of it in your PR to `master`. Thanks!

PR #55263, opened 2019-11-12
salt/cloud: utils+ec2: fix bugs, add new profile option: ssh_gateway_test_command

icy:
### What does this PR do?

When gateway is set to launch new ec2 instance, salt-cloud would expect a few commands
to continue: `date`, and `netcat`. Before this commit, `netcat` requires option `-q`
(which is often provided by `openbsd-netcat`), however, this is not always possible:

https://github.com/saltstack/salt/issues/55219

This commit introduces new option `ssh_gateway_test_command` in cloud profile, that
allows administrator to pick up the right command for testing if some port is open.
The default command string is `nc -z -w5 -q0 {0} {1}`; this is Python format string,
where `{0}` would be replaced by the target hostname, and `{1}` by the ssh port number.

This new option would fix #55219.

This commit also fixes some minor bugs when getting values of some gateway options.

Question: Is that possible to use gateway without its shell? In some strict
environment, the gateway only forwards connection and it doesn't allow to execute
any command (that means, `date` and `netcat` are not available.)

https://github.com/saltstack/salt/issues/55219

### Previous Behavior

`salt-cloud` may not launch new server if gateway is set with inappropriate netcat command.

### New Behavior

netcat options on gateway can be changed in cloud profile settings.

### Tests written?

No.

Test done manually.

### Commits signed with GPG?

Yes


icy:
@Ch3LL Can you please have a look and restart the builds? Some tests weren't started . Thanks

Ch3LL (changes requested):
would you mind adding some tests to this change so we can ensure it does not end up regressing in the future?

Ch3LL:
Following up here. Are you willing to add test coverage and a changelog?

icy:
> Following up here. Are you willing to add test coverage and a changelog?

Hi @Ch3LL , my apologies. I haven't used salt-stack for a while. I will have a look at this PR this weekend.

Cheers.

Ch3LL:
any updates here?

twangboy (changes requested):
Please rebase this and fix the conflicts. This will also need a testcase and a changelog.

Ch3LL:
Closing due to inactivity. Please let me know if you want me to re-open or open a new PR against master with the requested feedback.

PR #55270, opened 2019-11-12
Use Nox to drive MacOS tests

s0undt3ch:
### What does this PR do?
See title.

### DIFF
--- a/.ci/kitchen-macosxhighsierra-py2
+++ b/.ci/kitchen-macosxhighsierra-py2
@@ -14,6 +14,9 @@ def golden_images_branch = 'master'
 def nox_passthrough_opts = ''
 def concurrent_builds = 1
 def jenkins_slave_label = 'kitchen-slave-mac'
+def kitchen_driver_file = '/var/jenkins/workspace/driver.yml'
+def kitchen_verifier_file = '/var/jenkins/workspace/nox-verifier.yml'
+def kitchen_platforms_file = '/var/jenkins/workspace/pre-golden-nox-platforms.yml'
 
 properties([
     buildDiscarder(logRotator(artifactDaysToKeepStr: '', artifactNumToKeepStr: '', daysToKeepStr: '', numToKeepStr: '30')),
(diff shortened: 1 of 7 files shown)


Akm0d (approved):

.ci/kitchen-macosxhighsierra-py2
```diff
@@ -14,6 +14,9 @@ def golden_images_branch = 'master'
 def nox_passthrough_opts = ''
 def concurrent_builds = 1
 def jenkins_slave_label = 'kitchen-slave-mac'
+def kitchen_driver_file = '/var/jenkins/workspace/driver.yml'
+def kitchen_verifier_file = '/var/jenkins/workspace/nox-verifier.yml'
+def kitchen_platforms_file = '/var/jenkins/workspace/pre-golden-nox-platforms.yml'
```
bryceml: why not call it nox-platforms.yml so it's the same as the others?  just wondering.

s0undt3ch: We already have that one setup in sre-jenkins, but for proper MacOS golden images.

s0undt3ch: Just named it differently so that proper golden images can continue without affecting this switch to nox.

bryceml: what is different between the files?

waynew (approved):
Aside from Bryce's valid nit, 👍

.ci/kitchen-macosxhighsierra-py2
s0undt3ch: Proper golden images, [this](https://github.com/saltstack/sre-jenkins/blob/develop/states/services/jenkins-node/mac-install.sls#L457-L461), pre golden images [this](https://github.com/saltstack/sre-jenkins/blob/develop/states/services/jenkins-node/mac-install.sls#L540-L541) and [this](https://github.com/saltstack/sre-jenkins/blob/develop/states/services/jenkins-node/mac-install.sls#L566), as an example.

bryceml: nice

bryceml (approved):

s0undt3ch:
Follow up on https://github.com/saltstack/salt/pull/55272

PR #55278, opened 2019-11-13
Update net.py

aegiacometti:
Reflect new module name napalm instead of napalm.base.
Catch exception for IPv6
Support for ip secondary address

### What does this PR do?
Reflect new module name napalm intead of napalm-base
Added try/except when the interface does not have an IPv6 address.
Added support for secondary IP interface address
### Previous Behavior
net.find cannot load the correct napalm module name
Net.find not working because there was an dict_key error at line 228
AttributeError: 'dict_keys' object has no attribute 'extend'.
It should be attribute update, but anyway still having and error because the device don't have an IPv6 address (key), if IPv6 is not enabled at the device/interface.
Also add support for secondary IP address.

### DIFF
--- a/salt/runners/net.py
+++ b/salt/runners/net.py
@@ -78,12 +78,12 @@ from salt.ext.six.moves import map
 
 # Import third party libs
 try:
-    from netaddr import IPNetwork  # netaddr is already required by napalm-base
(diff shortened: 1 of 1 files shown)


dwoz:
@aegiacometti please submit this change against `master`

PR #55281, opened 2019-11-13
Aegiacometti patch 1

aegiacometti:
### What does this PR do?
1.- Reflect new module name napalm instead of napalm-base. (runners/net.py and runners/bgp.py)
2.- Catch exception for IPv6 (runners/bgp.py)
3.- Support for IP secondary address (runners/bgp.py)
4.- the module (runners/bgp.py) was trowing an exemption because a variable list named "compare" was initiated with [None] instead of []. After adding items to the list there was a max(compare), and None is not supported for comparing of the list with types None and IPNetwork.

### Previous Behavior
1.-
sudo salt-run net.find 1.1.1.1
'net' virtual False: The napalm-base module could not be imported
4.-
sudo salt-run net.find 10.100.12.2
Passed invalid arguments: '>' not supported between instances of 'IPNetwork' and 'NoneType'

### New Behavior
Finding/using the module napalm and not asking for napalm-base.
Returns the findings of net.find module/function.

### DIFF
--- a/salt/runners/bgp.py
+++ b/salt/runners/bgp.py
@@ -104,11 +104,11 @@ try:
     from netaddr import IPNetwork
     from netaddr import IPAddress
     # pylint: disable=unused-import
-    from napalm_base import helpers as napalm_helpers
+    from napalm.base import helpers as napalm_helpers
     # pylint: enable=unused-import
-    HAS_NAPALM_BASE = True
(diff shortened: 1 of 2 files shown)


aegiacometti:
cancelling this PR, will open a new one with some corrections

PR #55290, opened 2019-11-13
updating boto_elbv2 to work without full target group arns

buroa:
### What does this PR do?

Allows registering instances to target groups easier.

```register targets:
  boto_elbv2.targets_registered:
    - name: xxx-not-an-arn
    - targets:
      - i-xxx
```

N/A

### Tests written?

No

### Commits signed with GPG?

Yes


cmcmarrow (changes requested):
Looks good! But we still need some test.

PR #55321, opened 2019-11-15
Fix case where args or kwargs are not specified

fattybenji:

#52503 

This fixes a bug that is present in versions 2018.3 and 2019.2 so this should ideally be reintegrated in the associated branches for future releases.


dwoz:
@fattybenji The linter is failing and there are no tests for this change.

fattybenji:
I'll try and find the original PR to include everything...

waynew:
Thanks for the PR @fattybenji  but these changes are actually already present in `master`.

PR #55322, opened 2019-11-15
flat roster: pass master options to apply_sdb

jhass:
### What does this PR do?

Despite being more convenient and consistent, in avoiding to having to specify the sdb options
again for each host, `apply_sdb` (or actually `sdb_get` I think) also somehow manages to mess up
the previously passed options in the next iteration.

### Previous Behavior

Using sdb in the flat roster copied the entry of the current host into the key of the current one. It also required the sdb configuration to be present in each roster entry.

### New Behavior

Pass the master config available in `__opts__` to `apply_sdb`, centralizing the sdb configuration properly and keeping `sdb_get` from messing things up.

### Tests written?

is a bit beyond my experience with this project.

### DIFF
--- a/salt/roster/flat.py
+++ b/salt/roster/flat.py
@@ -36,5 +36,7 @@ def targets(tgt, tgt_type="glob", **kwargs):
     )
     conditioned_raw = {}
     for minion in raw:
-        conditioned_raw[six.text_type(minion)] = salt.config.apply_sdb(raw[minion])
+        conditioned_raw[six.text_type(minion)] = salt.config.apply_sdb(
+            __opts__, raw[minion]
+        )
     return __utils__["roster_matcher.targets"](conditioned_raw, tgt, tgt_type, "ipv4")
--- a/tests/unit/roster/test_flat.py
+++ b/tests/unit/roster/test_flat.py
@@ -1,0 +1,49 @@
+# -*- coding: utf-8 -*-
+"""
+unit tests for flat roster
+"""
+from __future__ import absolute_import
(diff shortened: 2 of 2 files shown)


xeacott:
Hey @jhass thanks for submitting this PR! Unfortunately we won't be able to merge without tests for this. I realize there aren't any unit or integration tests, but maybe we can work together to get some written!

jhass:
Sure, any pointers?

xeacott:
So it looks like there aren't any tests for the flat file itself, however there are tests residing in tests/unit/roster/test_clustershell.py and i think that's a pretty good example of how to test flat.py :)

xeacott:
Also @jhass I would also ask if you could resolve the conflict please and also run pre-commit on this. If you are unsure of pre-commit I can run that for ya.

jhass:
Gladly take over, I have no clue what I'm doing, whether this is any sensible test. It feels like it's just asserting a particular implementation, which in my book is pretty useless. Also let's see what CI says, locally I just get

```
$ python tests/runtests.py --from-filenames=salt/roster/flat.py
[1]    90592 abort      python tests/runtests.py --from-filenames=salt/roster/flat.py 
```

xeacott:
Thanks for getting that file together :) I'll get pre-commit ran for ya and then check out those tests! Sometimes they do seem really meaningless, especially when the change is _that_ minor but it helps increase our coverage!

xeacott (approved):

dwoz:

PR #55325, opened 2019-11-15
Add support for usage of headers in http_json ext pillar.

ni3mm4nd:
Add support of request headers for http_json ext pillar.
Such headers could be used for authentication purposes for example.

### What does this PR do?
Add support for headers propagation via http_json ext pillar module.

none

### Previous Behavior

### New Behavior

### Tests written?
No

### Commits signed with GPG?
No

### DIFF
--- a/salt/pillar/http_json.py
+++ b/salt/pillar/http_json.py
@@ -13,9 +13,8 @@ Set the following Salt config to setup http json result as external pillar sourc
   ext_pillar:
     - http_json:
         url: http://example.com/api/minion_id
-        ::TODO::
-        username: username
-        password: password
+        headers:
+            X-AUTH-TOKEN: 123456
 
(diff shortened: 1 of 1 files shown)


ni3mm4nd:
Hi

What kind of test would you like?
It depends only on salt http.query which is coverd already. The only thing I've done is just add of one parameter.

Thanks

cmcmarrow:
Hi, @ni3mm4nd. We only expect you to test your new parameter and check that it behaves expectedly. We will never expect you to test all of a function unless your wrote the function. We only expect you to test what you changed.

ni3mm4nd:
@cmcmarrow Would you please be so kind and point me to file where are some related tests already written to be able to add another one which could cover it? Thanks a lot.

dwoz:

Issue #55292, opened 2019-11-13
Kubernetes Secrets Fail To Base64 Encode Under Python 3


morinap:
### Description of Issue
Kubernetes secrets are not correctly base64 encoded under Python 3. An error is generated:
```
local:
----------
          ID: test-secret
    Function: kubernetes.secret_present
      Result: False
     Comment: An exception occurred in this state: Traceback (most recent call last):
                File "/Users/<redacted>g/tools/saltstack/salt/state.py", line 1933, in call
                  **cdata['kwargs'])
                File "/Users/<redacted>g/tools/saltstack/salt/loader.py", line 1939, in wrapper
                  return f(*args, **kwargs)
                File "/Users/<redacted>g/tools/saltstack/salt/states/kubernetes.py", line 583, in secret_present
                  **kwargs)
                File "/Users/<redacted>g/tools/saltstack/salt/modules/kubernetesmod.py", line 1358, in replace_secret
                  data[key] = base64.b64encode(data[key])
                File "/Users/<redacted>g/tools/python/lib/python3.7/base64.py", line 58, in b64encode
                  encoded = binascii.b2a_base64(s, newline=False)
              TypeError: a bytes-like object is required, not 'str'
     Started: 13:32:56.827595
    Duration: 22.805 ms
     Changes:
```

### Setup
Example sls:
```
test-secret:
  kubernetes.secret_present:
    - namespace: default
      metadata:
        name: test-secret
      data:
        tls.crt: test cert data
        tls.key: test key data
```

### Steps to Reproduce Issue
Execute the above sls using state.sls and receive the error

### Versions Report
This bug has been reproduced against `develop`, `2018.3`, and `2019.2` branches:
```
Salt Version:
           Salt: 2019.2.0-339-g779bc17

Dependency Versions:
           cffi: Not Installed
       cherrypy: Not Installed
       dateutil: 2.8.0
      docker-py: 4.0.2
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.10.1
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.6.2
   mysql-python: Not Installed
      pycparser: Not Installed
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: Not Installed
         Python: 3.7.4 (default, Sep 23 2019, 10:32:33)
[...]

PR #55326, opened 2019-11-15
Fix Kubernetes Secret Encoding Under Python 3

morinap:
### What does this PR do?
Base64 encode secret data properly under Python3. Secret data sent to the Kubernetes API must be a base64 encoded String object.

https://github.com/saltstack/salt/issues/55292

### Previous Behavior
A fatal error occurred

### New Behavior
Resolve fatal error

### Tests written?
No

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/modules/kubernetesmod.py
+++ b/salt/modules/kubernetesmod.py
@@ -1118,7 +1118,7 @@ def create_secret(
 
     # encode the secrets using base64 as required by kubernetes
     for key in data:
-        data[key] = base64.b64encode(data[key])
+        data[key] = __base64_str(data[key])
 
     body = kubernetes.client.V1Secret(
         metadata=__dict_to_object_meta(name, namespace, {}),
@@ -1355,7 +1355,7 @@ def replace_secret(name,
 
     # encode the secrets using base64 as required by kubernetes
     for key in data:
-        data[key] = base64.b64encode(data[key])
+        data[key] = __base64_str(data[key])
 
     body = kubernetes.client.V1Secret(
         metadata=__dict_to_object_meta(name, namespace, {}),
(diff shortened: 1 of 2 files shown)


salt/modules/kubernetesmod.py
```diff
@@ -1593,3 +1593,13 @@ def __enforce_only_strings_dict(dictionary):
...
 
     return ret
+
+
+def __base64_str(string):
+    '''
+    Converts a string of data into a base64 encoded string based on python version
+    '''
+    if sys.version_info >= (3, 0):
+        return str(base64.b64encode(string.encode('utf-8')), 'utf-8')
```
dwoz: How do we know this will always be `utf-8`?

morinap: @dwoz If `string` in this case is already a `str` object and I'm simply trying to get it to a bytes representation so that can be passed to the `b64encode` call, does it matter what the underlying encoding of the string is? Perhaps I am misunderstanding the problem. Are you suggesting there may be a problem if there are characters in `string` that cannot be encoded by `utf-8`?

morinap:
@dwoz I've added the appropriate unit test cases to this PR, rebased, and pushed everything up. I'll await a response from you regarding the encoding question.

salt/modules/kubernetesmod.py
DmitryKuzmenko: @morinap There are 2 more salty ways to do this depending on your goals. If you need a `unicode` string as a result on both py2 and py3 you can use
```python
salt.utils.hashutils.base64_b64encode(string)
```
If you need a native `str` object on all platforms you can do either
```python
salt.utils.stringutils.to_str(salt.utils.hashutils.base64_b64encode(string))
```

DmitryKuzmenko (changes requested):

DmitryKuzmenko:
@morinap thank you for contribution. Special thanks for the test!
Now this needs the following steps to be applied:
1. Update your master branch and install pre-commit hook as described in the contribution guide https://docs.saltstack.com/en/latest/topics/development/contributing.html
2. Rebase your commit resolving conflicts and update your fix with my comment suggestion.
3. Run pre-commit run on your branch to re-format your code (quotes and spacing) it will do it automatically.

Issue #55287, opened 2019-11-13
Creation of Kubernetes Service Fails With Invalid Port


morinap:
### Description of Issue
Calling `kubernetes.service_present` with any port configuration fails with an error in creating a kubernetes client library object:
```
local:
----------
          ID: web-service
    Function: kubernetes.service_present
      Result: False
     Comment: An exception occurred in this state: Traceback (most recent call last):
                File "/Users/<redacted>/tools/saltstack/salt/state.py", line 1933, in call
                  **cdata['kwargs'])
                File "/Users/<redacted>/tools/saltstack/salt/loader.py", line 1939, in wrapper
                  return f(*args, **kwargs)
                File "/Users/<redacted>/tools/saltstack/salt/states/kubernetes.py", line 312, in service_present
                  **kwargs)
                File "/Users/<redacted>/tools/saltstack/salt/modules/kubernetesmod.py", line 1070, in create_service
                  saltenv=saltenv)
                File "/Users/<redacted>/tools/saltstack/salt/modules/kubernetesmod.py", line 1464, in __create_object_body
                  spec=spec_creator(spec))
                File "/Users/<redacted>/tools/saltstack/salt/modules/kubernetesmod.py", line 1572, in __dict_to_service_spec
                  kube_port = kubernetes.client.V1ServicePort()
                File "/Users/<redacted>/tools/python/lib/python3.7/site-packages/kubernetes/client/models/v1_service_port.py", line 65, in __init__
                  self.port = port
                File "/Users/<redacted>/tools/python/lib/python3.7/site-packages/kubernetes/client/models/v1_service_port.py", line 138, in port
                  raise ValueError("Invalid value for `port`, must not be `None`")
              ValueError: Invalid value for `port`, must not be `None`
     Started: 09:33:32.025176
    Duration: 24.63 ms
     Changes:
```

### Setup
Example sls:
```
web-service:
  kubernetes.service_present:
    - namespace: default
      metadata:
        name: web-service
      spec:
        selector:
          app: web
        type: NodePort
        ports:
        - port: 3001
          target_port: 80
          protocol: TCP
```

### Steps to Reproduce Issue
Execute the above sls using `state.sls` and receive the error

### Versions Report
This has been reproduced against current develop and 2019.2. It works against 2018.3 because it uses an older version of the Kubernetes python client (3.0.0) for which this call does not fail. For any Kubernetes python client >= 4.0.0, it does fail.

```
Salt Version:
[...]

PR #55327, opened 2019-11-15
Fix Error in Creation of Kubernetes Service

morinap:
### What does this PR do?
Resolves an error in creating a kubernetes service by passing a port value into the Kubernetes V1ServicePort constructor.

https://github.com/saltstack/salt/issues/55287

### Previous Behavior
Initialize the port object with an empty argument lsit

### New Behavior
Pull the port value off and pass it to the argument list

### Tests written?
No

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/modules/kubernetesmod.py
+++ b/salt/modules/kubernetesmod.py
@@ -1569,7 +1569,7 @@ def __dict_to_service_spec(spec):
         if key == 'ports':
             spec_obj.ports = []
             for port in value:
-                kube_port = kubernetes.client.V1ServicePort()
+                kube_port = kubernetes.client.V1ServicePort(port=port.get('port', ''))
                 if isinstance(port, dict):
                     for port_key, port_value in iteritems(port):
                         if hasattr(kube_port, port_key):
--- a/tests/unit/modules/test_kubernetesmod.py
+++ b/tests/unit/modules/test_kubernetesmod.py
@@ -24,6 +24,14 @@ from salt.modules import kubernetesmod as kubernetes
 
 
 @contextmanager
+def real_kubernetes_library():
+    """
+    Get the actual kubernetes lib instance for testing
+    """
+    yield kubernetes.kubernetes
+
+
+@contextmanager
 def mock_kubernetes_library():
     """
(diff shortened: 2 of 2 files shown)


morinap:
@dwoz I've updated this branch to include a base test case for the method being modified and a case that tests the specific failure being addressed

salt/modules/kubernetesmod.py
```diff
@@ -1569,7 +1569,7 @@ def __dict_to_service_spec(spec):
         if key == 'ports':
             spec_obj.ports = []
             for port in value:
-                kube_port = kubernetes.client.V1ServicePort()
+                kube_port = kubernetes.client.V1ServicePort(port=port.get('port', ''))
```
Ch3LL: does this only work with kubernetes >=4.0.0 now? if so is it possible to keep compatibility for both 3 and 4 kubernetes versions?

morinap: A good question, thank you! I did just verify now that this still works for kubernetes-python below 4.0.0 as well.

jnschaeffer: This works for setting `port`, but not `node_port` or `target_port` (which map to `nodePort` and `targetPort` in the spec respectively). The constructor accepts a limited enough set of arguments that we could ditch the key iteration logic below and do something like this:

```python
kube_port = kubernetes.client.V1ServicePort(
    name=port.get('name'),
    node_port=port.get('nodePort'),
    port=port.get('port'),
    protocol=port.get('protocol'),
    target_port=port.get('targetPort'),
)
```
It seems like those options are all available in v3.0.0 ([link](https://github.com/kubernetes-client/python/blob/v3.0.0/kubernetes/client/models/v1_service_port.py)).

I found this while testing this exact PR myself with `targetPort` set on a service, so thought it might be worth mentioning. The code above also sets `port` to `None` if it's not defined in the spec, which will result in an error on applying the service by design.

Ch3LL:
Just following up here. This will need a changelog and did you see @jnschaeffer 's comment?

morinap:
I'm going to close this since it's all stale anyway and investigate if this is still an issue and if so, fixing against current code.

PR #55329, opened 2019-11-15
Filesystem specific mount parameter for Solaris

brejoc:
### What does this PR do?

The mount command of Solaris based operating systems is using -F to specify a filesystem type.

### Previous Behavior

Mount command with fstype wouldn't work for Solaris based operating systems.

### New Behavior

Mount command with fstype works for Solaris based operating systems.

### Tests written

No

### Commits signed with GPG?

No


brejoc:
Can be closed. Changes found its way into master.

PR #55375, opened 2019-11-20
WIP: .modules.pacmanpkg.purge: Behave like .modules.aptpkg.purge

ayekat:
*(Note: this is #54889, reopened against master)*

### What does this PR do?
Make `pacmanpkg.purge` behave like `aptpkg.purge`, i.e. remove a package along with its configuration files.

### Previous Behavior
`pacmanpkg.purge` invokes pacman with `-Rs`, i.e. it removes packages along with its dependencies.

The word "purge" in the context of package manager operations (to my knowledge) describes something else: apt uses `purge` for removing packages and its configuration files, whereas pacman does not have any meaning attributed to that word.

More importantly, the same action (`pkg.purge`) does different things on different systems.

### New Behavior
`pacmanpkg.purge` now invokes pacman with `-Rn`, i.e. it removes packages along with its configuration files.

This is more in line with what at least Debian users would expect, and does not conflict with any predisposed understanding that Arch users might have.

It does, however, introduce a significant, backward-incompatible change for pacman-based systems.
Also, this removes the possibility for recursively removing unrequired dependencies on pacman-based systems, so it might be wise to add a new action for pacman in the same way as `aptpkg.autoremove`.

### Tests written?

No

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/modules/pacmanpkg.py
+++ b/salt/modules/pacmanpkg.py
@@ -709,7 +709,7 @@ def _uninstall(action="remove", name=None, pkgs=None, **kwargs):
     if not targets:
         return {}
 
-    remove_arg = "-Rsc" if action == "purge" else "-R"
+    remove_arg = '-Rn' if action == 'purge' else '-R'
 
     cmd = []
     if salt.utils.systemd.has_scope(__context__) and __salt__["config.get"](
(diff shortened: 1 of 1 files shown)


Akm0d:
@ayekat Can you write a test for this PR?

ayekat:
Hey @Akm0d, thanks for the reply!

Sure, I can do that (I'll just have to read up on how to write tests first).

Just out of curiosity, what is that new merge commit? When I rebase onto master again, should I take care to retain that merge commit, or can I clean up the commit history again?

ayekat:
Sorry for the delay on this. Unfortunately, it doesn't look like I'll get around to working into how to write tests for this anytime soon.

As I don't use Salt on Arch Linux, this is not relevant for me personally. And in the interest of keeping the PR list clean, I'm going to close this.

PR #55379, opened 2019-11-20
fix logging on python3.8

bryceml:
### What does this PR do?
This lets the test-suite start on python3.8.  There are more issues like the minion not starting, but this is at least one thing that will need to be fixed.


dwoz (approved):

s0undt3ch:
https://github.com/saltstack/salt/pull/55441 is a more complete fix

bryceml:
closed as @s0undt3ch's fix is better.

PR #55383, opened 2019-11-21
31022 fix rbenv build env parsing

### LINKED ISSUE #31022: [Rbenv] rbenv:build_env doesn't appear to make it to the minion
When swapping from RVM managed outside of saltstack to rbenv with staltstack I hit issues in that static libraries weren't being compiled by default.

With careful reading of:
https://docs.saltstack.com/en/develop/ref/modules/all/salt.modules.rbenv.html

```
Additional environment variables can be configured in pillar / grains / master:

rbenv:
  build_env: 'CONFIGURE_OPTS="--no-tcmalloc" CFLAGS="-fno-tree-dce"'
```

( which was harder to find than I thought starting from the state )

And putting it together with [this ruby-build thread](https://github.com/rbenv/ruby-build/issues/35) it appeared there was a way to have shared libraries built.

Running a similar command in a shell seems to work:

```
$ CONFIGURE_OPTS="--enable-shared" rbenv install 2.3.0

# eventually generated these files among others
/usr/local/rbenv/versions/2.3.0/lib/libruby-static.a
/usr/local/rbenv/versions/2.3.0/lib/libruby.so
/usr/local/rbenv/versions/2.3.0/lib/libruby.so.2.3
/usr/local/rbenv/versions/2.3.0/lib/libruby.so.2.3.0
```

So, in my pillar I have:

``` yaml

---
rbenv:
  build_env: 'CONFIGURE_OPTS="--enable-shared"'
  rubies:
    - 2.3.0
```

and among my states I have:

``` yaml
rbenv-deps:
  pkg.installed:
    - names:
      - bash
      - git
      - openssl
      - libssl-dev
      - make
      - curl
      - autoconf
      - bison
      - build-essential
      - libffi-dev
      - libyaml-dev
      - libreadline6-dev
      - zlib1g-dev
      - libncurses5-dev

{% for ruby in salt['pillar.get']("rbenv:rubies", []) %}
{{ ruby }}:
  rbenv.installed:
    - default: True
    - require:
      - pkg: rbenv-deps
{% endfor %}

/etc/profile.d/rbenv.sh:
  file.append:
    - text:
      - 'export RBENV_ROOT="/usr/local/rbenv"'
      - 'export PATH="$RBENV_ROOT/bin:$PATH"'
      - 'eval "$(rbenv init -)"'
```

Deep in the rbenv module it does `config.get("rbenv:build_env")` so I tried that.

```
salt '*' config.get "rbenv:build_env"
production-1:
    CONFIGURE_OPTS="--enable-shared"
```

When running the high state, no shared libraries are generated as they were when I ran the command directly. Is it possible they are getting scrubbed from the environment somewhere below the rbenv module?

No related events appeared in minion or master salt logs. 
Is there anything else for me to double check?

```
# salt-minion --versions
Salt Version:
           Salt: 2015.8.4

Dependency Versions:
         Jinja2: 2.7.2
       M2Crypto: Not Installed
           Mako: 0.9.1
         PyYAML: 3.10
          PyZMQ: 14.0.1
         Python: 2.7.6 (default, Mar 22 2014, 22:59:56)
           RAET: Not Installed
        Tornado: 4.2.1
            ZMQ: 4.0.4
           cffi: Not Installed
       cherrypy: Not Installed
       dateutil: 1.5
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
        libgit2: Not Installed
        libnacl: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.3.0
   mysql-python: 1.2.3
      pycparser: Not Installed
       pycrypto: 2.6.1
         pygit2: Not Installed
   python-gnupg: Not Installed
          smmap: Not Installed
        timelib: Not Installed

System Versions:
           dist: Ubuntu 14.04 trusty
        machine: x86_64
        release: 3.13.0-55-generic
         system: Ubuntu 14.04 trusty
```

brlanier:
### What does this PR do?
Fixes the rbenv module issue with not passing the build_config compile environment options to the command line for use. They were getting lost along the way. At some point in the past a change was made on how the options where being handled as strings and I believe that is when this stopped working. I don't know that many people use this option and my use case was trivial so I never bothered to debug. Please see the referenced issue for more details on the problem.

Originally, the build_env was passed as a string, converted to a list, back to a string and then expected to be a list or dictionary. If it was a list, it was converted to a dictionary. This PR changes the expectation of the pillar/grain input to be in a dictionary format, while retaining compatibility for previous setups. 

Also, not a python expert so guessing on some of the use of the built in tools and practices I worked with. 

Fixes #31022 

### Previous Behavior
The rbenv state would work fine but any compile time options you might have tried to set via pillar or grains would be ignored. For instance, if you tried to compile a ruby with `RUBY_CONFIGURE_OPTS="--with-jemalloc`, that compile time option never made it to the ruby build process.

### New Behavior
The build_env pillar/grains options are now handled properly and not lost in the process. If you use the state to compile a ruby with extra options, you will now see the compiled in options present.

### Tests written?
No, tests updated. I know enough python to get this far.

I did test this works for my old configuration that was previously getting silently ignored and I tested under the new dictionary setup I propose for going forward since this appears to be what is asked for in the code anyway.

### Commits signed with GPG?

No

### DIFF
--- a/salt/modules/rbenv.py
+++ b/salt/modules/rbenv.py
@@ -217,12 +217,14 @@ def install_ruby(ruby, runas=None):
         run as the user under which Salt is running.
 
     Additional environment variables can be configured in pillar /
-    grains / master:
+    grains / master as a dictionary:
 
     .. code-block:: yaml
 
         rbenv:
-          build_env: 'CONFIGURE_OPTS="--no-tcmalloc" CFLAGS="-fno-tree-dce"'
+          build_env:
+            CONFIGURE_OPTS: "--no-tcmalloc"
+            CFLAGS: "-fno-tree-dce"
 
     CLI Example:
 
@@ -232,19 +234,21 @@ def install_ruby(ruby, runas=None):
     '''
     ruby = re.sub(r'^ruby-', '', ruby)
 
-    env = None
-    env_list = []
+    env = {}
+    build_env = None
 
     if __grains__['os'] in ('FreeBSD', 'NetBSD', 'OpenBSD'):
-        env_list.append('MAKE=gmake')
+        env['MAKE'] = 'gmake'
 
     if __salt__['config.get']('rbenv:build_env'):
-        env_list.append(__salt__['config.get']('rbenv:build_env'))
+        build_env = __salt__['config.get']('rbenv:build_env')
     elif __salt__['config.option']('rbenv.build_env'):
-        env_list.append(__salt__['config.option']('rbenv.build_env'))
+        build_env = __salt__['config.option']('rbenv.build_env')
 
-    if env_list:
-        env = ' '.join(env_list)
+    if isinstance(build_env, dict):
+        env.update(build_env)
+    elif isinstance(build_env, str):
+        env.update(dict(build_arg.split('=', 1) for build_arg in salt.utils.args.shlex_split(six.text_type(build_env))))
 
     ret = {}
     ret = _rbenv_exec(['install', ruby], env=env, runas=runas, ret=ret)
--- a/salt/states/rbenv.py
+++ b/salt/states/rbenv.py
@@ -18,6 +18,18 @@ directories to the users PATH. If you are installing as root and want other
 users to be able to access rbenv then you will need to add RBENV_ROOT to
 their environment.
 
+To pass additional environment based compile time options to rbenv_build, you
+can setup a build_env dictionary in the pillar/grains/master configuration. The
+following example would pass `CONFIGURE_OPTS="--no-tcmalloc" CFLAGS="-fno-tree-dce"``
+to the rbenv_build process.
+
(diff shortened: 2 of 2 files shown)


brlanier:
I actually was using this in salt version 2019.2.2 but opening against 2018.3 per the contrib docs as the bug exists there but I don't fully understand the lifecycle process, so please let me know if you prefer this against a different branch.

dwoz:
@brlanier We are no longer accepting PRs to the `2018.3` branch. Sorry we have not update the docs to reflect our new branching strategy yet. This change should be made against `master` and we'll label it as something that should get back-ported to a bugfix release.

brlanier:
Thanks... wasn't sure.. Done on #55496

PR #55391, opened 2019-11-21
Replace inspect.getargspec -- it's deprecated in Python 3.0

philodavies:
### What does this PR do?
Replaces `inspect.getargspec` with `inspect.getfullargspec` (it's deprecated in Python3.0)
Removes netmiko_defaults as it was not being used in netmiko_mod

My IDE also automatically runs Black on write, so the code was also formatted.

N/A

### Tests written?
No (this is an improvement that shouldn't modify the values returned by the previously used function)

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/modules/napalm_mod.py
+++ b/salt/modules/napalm_mod.py
@@ -1,12 +1,12 @@
 # -*- coding: utf-8 -*-
-'''
+"""
 NAPALM helpers
 ==============
 
 Helpers for the NAPALM modules.
 
 .. versionadded:: 2017.7.0
-'''
+"""
 from __future__ import absolute_import, unicode_literals, print_function
 
 # Import python stdlib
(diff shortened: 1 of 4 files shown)


philodavies:
I think I goofed up and created my branch from master when it should have been from develop.

s0undt3ch:
First of all, Thank You for the work you put in to get us even closer to a Py3 codebase.
Second, new changes should be made, based, and against, the master branch.
Third, though Salt is also moving to black, https://github.com/saltstack/salt-enhancement-proposals/pull/21, it hasn't yet, so, you will actually need to disable black formatting on save in order for use to review your PR. Makes it hard to actually see and review your fix amongst so much changes.

Sorry for the extra work.

philodavies:
That's no problem. I'll redo this commit and remove the automated black formatting and rebase from master.

I also noticed one of the libraries did a check for getfullargspec and fell back on getargspec which is probably how I should have done it. Will use that template to keep things from breaking on any systems still on python2

philodavies:
I wasn't able to make this change and do not use saltstack anymore. I don't think I'll be able to fix this so I'll close the issue

PR #54878, opened 2019-10-03
Master thread safe loader

Akm0d:
### What does this PR do?
Merge PR from develop https://github.com/saltstack/salt/pull/50655 which was a new version of #39948, #48598, and #50648 so the description has been copied from there. This PR will hopefully resolve the problems that existed in the original PRs. The code is identical to #50648, but this PR is based on the develop branch instead of the 2018.3.3 branch because Jenkins obviously cannot run tests for branches based on the 2018.3.3 branch.

There was a race condition in the salt loader when injecting global values (e.g. "pillar" or "salt") into modules. One effect of this race condition was that in a setup with multiple threads, some threads may see pillar data intended for other threads or the pillar data seen by a thread might even change spuriously.

There have been earlier attempts to fix this problem (#27937, #29397). These patches tried to fix the problem by storing the dictionary that keeps the relevant data in a thread-local variable and referencing this thread-local variable from the variables that are injected into the modules.

These patches did not fix the problem completely because they only work when a module is loaded through a single loader instance only. When there is more than one loader, there is more than one thread-local variable and the variable injected into a module is changed to point to another thread-local variable when the module is loaded again. Thus, the problem resurfaced while working on #39670.

This patch attempts to solve the problem from a slightly different angle, complementing the earlier patches: The value injected into the modules now is a proxy that internally uses a thread-local variable to decide to which object it points. This means that when loading a module
again through a different loader (possibly passing different pillar data), the data is actually only changed in the thread in which the loader is used. Other threads are not affected by such a change.

This means that it will work correctly in the current situation where loaders are possibly created by many different modules and these modules do not necessary know in which context they are executed. Thus it is much more flexible and reliable than the more explicit approach
used by the two earlier patches.
What issues does this PR fix or reference?

This PR fixes problems that surfaced when developing the parallel runners feature (#39670), but is also related to problems reported earlier (#23373). The problems described in #29028 might also be related, though this needs further confirmation.
Previous Behavior

Changes to pillar data in one thread might have influenced pillar data in a different thread. Whether or when this problem actually appeared was unpredictable (due to the nature of a race condition).
New Behavior

Changes to pillar data in one thread will never affect the pillar data in a different thread.
Tests written?

Yes
Regression Potential

The change to the loader code itself is rather small, but it affects a central component. So if there is a problem, the impact might affect nearly all components.

PR #55398, opened 2019-11-21
Port #51499 to master

oeuftete:
TODOs:
* [x] figure out how to port this or if it's required.  Things have changed somewhat,

### What does this PR do?

Ports #51499 to master.  See that PR for additional information.

It looks like the original loader changes are no longer necessary with refactoring that has occurred since the original issue was addressed.  The original author's tests look like they will show whether that is the case.

### Commits signed with GPG?

No.

### DIFF
--- a/tests/integration/files/file/base/issue-51499.sls
+++ b/tests/integration/files/file/base/issue-51499.sls
@@ -1,0 +1,6 @@
+{% if 'nonexistent_module.function' in salt %}
+{% do salt.log.warning("Module is available") %}
+{% endif %}
+always-passes:
+  test.succeed_without_changes:
+    - name: foo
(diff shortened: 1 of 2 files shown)


sagetherage:
this looks like it just needs the WIP title removed and all other checks pass - but also needs a rebase, is that right @waynew ?

sagetherage:
@mattp- @oeuftete this looks like it needs to be rebased and WIP in the title removed.

oeuftete:
I've rebased.  Note that the original change to the main code base was removed, since it looked not applicable with refactors to the loader since.  The added tests have been left, though, and appeared to be passing (before this rebase, anyway.)

oeuftete:
@Akm0d  It looks like you might be able to assess whether no change in the loader here is appropriate, based on https://github.com/saltstack/salt/pull/54878#issuecomment-538604777?

dwoz (approved):

PR #55401, opened 2019-11-21
salt.runners.pillar: fix special case for masters pillar

salty-blue-mango:
when asking for the pillar of the running minion in master context it
talks on a '_master' to the minion_id, which will be a cache miss for
grains. if this is asked for, just generate it via salt.loader.grains

### What does this PR do?
Port of #50760 to master

salt-run pillar.show_pillar foo_master didn't work

### Tests written?
No

### Commits signed with GPG?

No


Oloremo:
Any plans to finish this PR?

twangboy (changes requested):
This needs a testcase and a changelog. Also needs to be rebased and conflicts fixed.

PR #55410, opened 2019-11-22
Remove _supported_dists import for python-3.8 support 

mfussenegger:
Replace an import to a structure that was removed in Python3.8

See commit message for details

### Previous Behavior

ImportError under Python 3.8

### New Behavior

No ImportError

### Tests written?

No

I suspect that some tests already fail if the CI contains python 3.8

### Commits signed with GPG?

Yes


s0undt3ch (commented):
The right move is to actually switch to using `distro`

PR #55418, opened 2019-11-23
Added clean_parent argument for the archive state

Oloremo:
### What does this PR do?

Adding new argument `clean_parent` to the archive state. This allows the state to remove parent directory before the extraction.

#54012 

### Previous Behavior
archive state was unable to do so by itself.

### New Behavior
archive state now able to remove parent directory if requested

### Tests written?

Yes... kinda.

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/states/archive.py
+++ b/salt/states/archive.py
@@ -183,6 +183,7 @@ def extracted(name,
               force=False,
               overwrite=False,
               clean=False,
+              clean_parent=False,
               user=None,
               group=None,
               if_missing=None,
@@ -521,6 +522,12 @@ def extracted(name,
             ``overwrite`` is set to ``True``.
 
         .. versionadded:: 2016.11.1
+
+    clean_parent : False
+        Set this to ``True`` to remove a parent archive directory before the extration.
+        ``clean`` and ``clean_parent`` args are mutually exclusive.
+
+        .. versionadded:: Sodium
 
     user
         The user to own each extracted file. Not available on Windows.
@@ -1076,6 +1083,11 @@ def extracted(name,
                           ))
         return ret
 
+    if clean and clean_parent:
+        ret['comment'] = "You can't set both 'clean' and 'clean_parent' to True."
+        ret['result'] = False
+        return ret
+
     extraction_needed = overwrite
     contents_missing = False
 
@@ -1148,6 +1160,15 @@ def extracted(name,
                         )
                     )
                     return ret
+                if __opts__['test'] and clean_parent and contents is not None:
+                    ret['result'] = None
+                    ret['comment'] += (
+                        ' Since the \'clean_parent\' option is enabled, the '
+                        'destination parent directory would be removed first '
+                        'and than re-created and the archive would be '
+                        'extracted'
+                    )
+                    return ret
 
                 # Skip notices of incorrect types if we're cleaning
                 if not (clean and contents is not None):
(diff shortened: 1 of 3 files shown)


Oloremo:
The only thing that I test is the conflict check between clean and clean_parent. 
The archive.extracted state is huge and only have 3 test cases, right now I don't really have an idea how to test this new logic. Any advice welcome.

salt/states/archive.py
```diff
@@ -1076,6 +1083,11 @@ def extracted(name,
                           ))
         return ret
 
+    if clean and clean_parent:
+        ret['comment'] = "You can't set both 'clean' and 'clean_parent' to True."
```
terminalmage: We don't typically use the phrasing "you can't". Instead, can you change this to `Only one of 'clean' and 'clean_parent' can be set to True`?

salt/states/archive.py
```diff
@@ -1216,6 +1237,26 @@ def extracted(name,
             _add_explanation(ret, source_hash_trigger, contents_missing)
             return ret
 
+        if clean_parent and contents is not None:
+            errors = []
+            log.debug('Removing directory %s due to clean_parent set to True', name)
+            try:
+                salt.utils.files.rm_rf(name.rstrip(os.sep))
+                ret['changes'].setdefault(
+                    'removed', "Directory {} was removed prior to the extraction".format(name))
```
terminalmage: Since you've removed the directory, you need to recreate it now. And you also need to take into account the value of the `makedirs` argument when you do. Otherwise, the destination directory won't exist when it comes time to extract, and the extraction will fail.

terminalmage (changes requested):
Just a few minor changes.

salt/states/archive.py
Oloremo: Correct me if I'm wrong but parent directory going to be created no matter what:
```
1286         if not os.path.isdir(name):
1287             __states__['file.directory'](name, user=user, makedirs=True)
1288             created_destdir = True
```

salt/states/archive.py
Oloremo: thanks, done!

salt/states/archive.py
```diff
@@ -522,6 +523,12 @@ def extracted(name,
 
         .. versionadded:: 2016.11.1
 
+    clean_parent : False
+        Set this to ``True`` to remove a parent archive directory before the extration.
```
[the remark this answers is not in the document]
Oloremo: thanks, done!

salt/states/archive.py
terminalmage: Ahh yes, you are correct. I mistakenly thought that this state used a `makedirs` param, I must have been confusing it with something else.

terminalmage (approved):

Oloremo:
Did a rebase + added a changelog

Oloremo:
re-run full centos7-py2-m2crypto

Oloremo:
re-run full  amazon2-py2

Akm0d (approved):

salt/states/archive.py
```diff
@@ -1148,6 +1162,15 @@ def extracted(name,
                         )
                     )
                     return ret
+                if __opts__['test'] and clean_parent and contents is not None:
+                    ret['result'] = None
+                    ret['comment'] += (
+                        ' Since the \'clean_parent\' option is enabled, the '
+                        'destination parent directory would be removed first '
+                        'and than re-created and the archive would be '
```
waynew: Typo:
```suggestion
                        'and then re-created and the archive would be '
```

Could also be re-written to include the actual path that will be purged (using `.format`). But definitely fix the typo.

waynew (changes requested):
Small typo

salt/states/archive.py
Oloremo: thanks, fixed

PR #55434, opened 2019-11-25
Don't remove one directory level from slspath (master branch)

terminalmage:
This is the same fix as https://github.com/saltstack/salt/pull/55433, opened against the master branch.


dwoz:
re-run full centos

dwoz (approved):

PR #55441, opened 2019-11-26
Initial logging Refactoring Bugfix

s0undt3ch:
### What does this PR do?
See title.

### DIFF
--- a/salt/_logging/impl.py
+++ b/salt/_logging/impl.py
@@ -121,9 +121,9 @@ LOGGING_STORE_HANDLER = __StoreLoggingHandler()
 class SaltLogRecord(logging.LogRecord):
     def __init__(self, *args, **kwargs):
         logging.LogRecord.__init__(self, *args, **kwargs)
-        self.bracketname = '[{:<17}]'.format(self.name)
-        self.bracketlevel = '[{:<8}]'.format(self.levelname)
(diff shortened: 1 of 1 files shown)


salt/_logging/impl.py
```diff
@@ -121,9 +121,9 @@ class __StoreLoggingHandler(TemporaryLoggingHandler):
 class SaltLogRecord(logging.LogRecord):
     def __init__(self, *args, **kwargs):
         logging.LogRecord.__init__(self, *args, **kwargs)
-        self.bracketname = '[{:<17}]'.format(self.name)
-        self.bracketlevel = '[{:<8}]'.format(self.levelname)
-        self.bracketprocess = '[{:>5}]'.format(self.process)
+        self.bracketname = '[{:<17}]'.format(str(self.name))
+        self.bracketlevel = '[{:<8}]'.format(str(self.levelname))
+        self.bracketprocess = '[{:>5}]'.format(str(self.process))
```
Akm0d: Isn't casting the args to `format` to a string redundant?

s0undt3ch: Yes, but when python creates a new log record from a dictionary it passes `None` initially and format blows up with that.
It then updates the classes `__dict__` with valid data.
This cast is just to avoid that traceback.

And kudos for the review.

PR #55450, opened 2019-11-27
Improve robustness for invalid dmidecode characters and add test.

mmulqueen:
### What does this PR do?

Ensures that salt does not crash when trying to load DMI data to build the core grains. DMI data is expected to be valid text, however on at least some hardware (Intel NUC marketed around 2014), it contains characters that aren't valid (\xff). This causes salt to crash during start up.

### Previous Behavior
With Python 2.7, there is no bug because it was resolved in https://github.com/saltstack/salt/pull/48440

In Python 3, the difference in default behaviour of file opening causes the bug to resurface.

### New Behavior

Files are explicitly opened as bytes in both Python 2.7 and Python 3 versions, this prevents the bug from occurring and makes the previous bug fix work as intended.

### Tests written?

Yes. 

Test previously written by @Ch3LL has been adapted and new test written for invalid character. test_core_virtual_unicode and test_core_virtual_invalid

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/grains/core.py
+++ b/salt/grains/core.py
@@ -960,7 +960,7 @@ def _virtual(osdata):
                     grains['virtual'] = 'kvm'
         if os.path.isfile('/sys/devices/virtual/dmi/id/product_name'):
             try:
-                with salt.utils.files.fopen('/sys/devices/virtual/dmi/id/product_name', 'r') as fhr:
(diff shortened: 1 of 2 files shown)


dwoz:
@mmulqueen This needs to be re-opened against the `master` branch. We are no longer accepting PRs to `2018.3`. Sorry for the confusion.

PR #55453, opened 2019-11-27
we don't pass run_full anymore since the property is set in the function we are calling

bryceml:
### What does this PR do?
moves parameter fetching into the function rather than the jenkins file


s0undt3ch (commented):
:+1:

Issue #55481, opened 2019-12-01
test_network_ping fails on windows with ipv6


bryceml:
### Description of Issue
https://jenkinsci.saltstack.com/job/pr-windows2019-py2/job/master/153/testReport/junit/integration.modules.test_network/NetworkTest/test_network_ping/
fails if the instance has ipv6.  We most likely just need to change the value it's checking for in the output.

PR #55482, opened 2019-12-01
fix network ping test on windows with ipv6

bryceml:
### What does this PR do?
fixes windows ping test when it has ipv6

https://github.com/saltstack/salt/issues/55481

### Previous Behavior
https://jenkinsci.saltstack.com/job/pr-windows2019-py2/job/master/153/testReport/junit/integration.modules.test_network/NetworkTest/test_network_ping/

### DIFF
--- a/tests/integration/modules/test_network.py
+++ b/tests/integration/modules/test_network.py
@@ -23,7 +23,7 @@ class NetworkTest(ModuleCase):
         network.ping
         '''
         ret = self.run_function('network.ping', [URL])
-        exp_out = ['ping', URL, 'ttl', 'time']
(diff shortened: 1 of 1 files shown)


s0undt3ch (approved):

dwoz (changes requested):
Don't we still need `ttl` if this test runs in a non-ipv6 environment?

bryceml:
it just parses the output to make sure it ran the command, we don't even need ms to detect that, but it's extra insurance.

it basically runs `ping [url]` on windows or `ping -c 4` on linux and makes sure it was successful by parsing the output.

PR #52327, opened 2019-03-26
Allow arbitrary arguments to be passed to pip through pip.installed

Akm0d:
### What does this PR do?
Makes it possible to pass arbitrary arguments to pip from pip.installed

### What issues does this PR fix or reference?
https://github.com/saltstack/salt/issues/24751

### Previous Behavior
N/A

### New Behavior
 pip keyword and positional arguments not yet implemented in salt can be passed to pip through pip.installed

    .. code-block:: yaml

        pandas:
          pip.installed:
            - name: pandas
            - pip_future:
              - --latest-pip-kwarg:
                - param1
                - param2
              - --latest-pip-arg

    Will be translated into the following pip command:

    .. code-block:: bash

        pip install pandas --latest-pip-kwarg param1 --latest-pip-kwarg parm2 --latest-pip-arg

### Tests written?

Yes

### Commits signed with GPG?

Yes

PR #55492, opened 2019-12-02
Allow arbitrary arguments to be bassed through the pip module

Akm0d:
### What does this PR do?
Port https://github.com/saltstack/salt/pull/52327 to master

### DIFF
--- a/salt/modules/pip.py
+++ b/salt/modules/pip.py
@@ -608,7 +608,7 @@ def install(pkgs=None,  # pylint: disable=R0912,R0913,R0914
 
     extra_args
         pip keyword and positional arguments not yet implemented in salt
-        
+
         .. code-block:: yaml
 
             salt '*' pip.install pandas extra_args="[{'--latest-pip-kwarg':'param'}, '--latest-pip-arg']"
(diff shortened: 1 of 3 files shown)


dwoz (changes requested):
Seems like we need to validate that we are not introducing the potential for a shell escape here

Akm0d:
> Seems like we need to validate that we are not introducing the potential for a shell escape here

The python documentation on `subprocess` has this to say: 
```
Unlike some other popen functions, this implementation will never implicitly call a system shell. This means that all characters, including shell metacharacters, can safely be passed to child processes. If the shell is invoked explicitly, via shell=True, it is the application’s responsibility to ensure that all whitespace and metacharacters are quoted appropriately to avoid shell injection vulnerabilities.

When using shell=True, the shlex.quote() function can be used to properly escape whitespace and shell metacharacters in strings that are going to be used to construct shell commands.
```
Perhaps it would be good to wrap all options in `shlex.quote()` before appending them to the pip command

Akm0d:
Actually, this is already handled in `cmdmod.py` which is where the pip command is ultimately handed to:
```    if python_shell is not True \
            and not salt.utils.platform.is_windows() \
            and not isinstance(cmd, list):
        cmd = salt.utils.args.shlex_split(cmd)```

xeacott (approved):

PR #55522, opened 2019-12-05
[PyTest #20] Reduce code duplicattion between salt and pytest-salt

s0undt3ch:
### What does this PR do?
See title.
Plus some additional cleanup.


waynew:
currently looks like some lint issues 😛

dwoz (approved):

Issue #49422, opened 2018-08-29
Handle deprecated code in cmd state for Neon release


rallytime:
- [ ] salt/states/cmd.py

Ch3LL:
this PR https://github.com/saltstack/salt/pull/44840 added the initial deprecation warning for added context

PR #55573, opened 2019-12-09
Deprecate quiet kwarg in cmd.run state module

Ch3LL:
### What does this PR do?
Deprecate quiet kwarg in cmd.run state module

https://github.com/saltstack/salt/issues/49422

### Previous Behavior
You could pass the kwarg `quiet: True` to the cmd.run state module.

### New Behavior
You can no longer use `quiet: True` in the cmd.run state module and need to use `output_loglevel: quiet` instead.

### Tests written?
Yes

### Commits signed with GPG?

Yes


dwoz:
@Ch3LL Are the cmdmod test failures related?

Ch3LL:
yes they are. looks like `TestsLoggingHandler` got renamed to `TstSuiteLoggingHandler` in a recent PR merge here: https://github.com/saltstack/salt/pull/55533 so it was not able to load the test module. Just pushed a fix, we will see how this goes :)

dwoz (approved):

PR #55580, opened 2019-12-10
[master] Fix inconsistency with `pkg.list_pkgs` when using `attr` on RHEL systems

meaksh:
This PR ports #51108 to master branch

Now, `pkg_resource.format_pkg_list` does not directly parse the "arch" from the package name but delegates it to an optional `parse_arch_from_name` particular implementation on each virtual "pkg" module.

---

### What does this PR do?
This PR fixes an issue on the `pkg.list_pkgs` method when using `attr` parameter on RHEL systems.

On past PRs we've introduced the use of `attr` on `pkg.list_pkgs` on the Zypper/Yum modules for collecting not only the package version string but instead getting the desired RPM attributes (like `epoch`, `install_date`, `release`, etc) in an unified and structured way.

So when using `attr` the output from `pkg.list_pkgs` will follow this specific format for `attr='release,arch'`):
```yaml
"package": [
  {
    "version": "value1",
    "release": "value2",
    "arch": "value3",
  },
  {
    "version": "value4",
    "release": "value5",
    "arch": "value6",
  }
]
```
As you can see, the format is meant to group all installed versions of the same package (with its RPM attributes) into a single key `package` which holds a list containing a dict with all attrs for each installed version.

### Previous Behavior

On SUSE systems there is no problem since packages with different architectures always have different names, like `glibc` and `glibc-32bit` :+1: 

But the problem is happening on RHEL systems, where the architecture is internally handled by Salt as part of the package name. This is leading into an inconsistent looking output from `pkg.list_pkgs` according with the proposed format when using `attr`:

```json
$ salt-call --local pkg.list_pkgs attr='arch,release' --output=json

...
        "glibc": [
            {
                "release": "1.212.el6", 
                "version": "2.12", 
                "arch": "x86_64"
            }
        ], 
        "glibc.i686": [
            {
                "release": "1.212.el6", 
                "version": "2.12", 
                "arch": "i686"
            }
        ],
...
```

### New Behavior

The expected output should consider `glibc` and `glibc.i686` as the same package and group them into the same dict key. The output should be:
```json
$ salt-call --local pkg.list_pkgs attr='arch,release' --output=json

...
        "glibc": [
            {
                "release": "1.212.el6", 
                "version": "2.12", 
                "arch": "i686"
            }, 
            {
                "release": "1.212.el6", 
                "version": "2.12", 
                "arch": "x86_64"
            }
        ], 
...
```

### Tests written?

Yes

### Commits signed with GPG?

Yes


terminalmage:
This broke 32-bit package installs on 64-bit Debian/Ubuntu. I'm not sure why a modification to aptpkg.py was included in a PR meant to address RHEL, or how it got merged without a test, though given that it was incorrectly implemented, a test would likely only have served to confirm the incorrect behavior anyway.

At any rate, I've fixed this in https://github.com/saltstack/salt/pull/57084.

PR #55587, opened 2019-12-10
Ignore return code in "run_all" and do not raise an exception therefore

terminalmage:
This is https://github.com/saltstack/salt/pull/51623, reopened against master. It is a bugfix for https://github.com/saltstack/salt/issues/51620, so it should be targeted for 2019.2.3.

### DIFF
--- a/changelog/51620.fixed
+++ b/changelog/51620.fixed
@@ -1,0 +1,1 @@
+Do not raise an exception in pkg.info_installed on nonzero return code
--- a/salt/modules/rpm_lowpkg.py
+++ b/salt/modules/rpm_lowpkg.py
@@ -615,17 +615,9 @@ def info(*packages, **kwargs):
         output_loglevel="trace",
         env={"TZ": "UTC"},
         clean_env=True,
+        ignore_retcode=True,
     )
-    if call["retcode"] != 0:
-        comment = ""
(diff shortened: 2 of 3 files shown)


terminalmage:
Test case added.

terminalmage:
re-run full all

waynew (dismissed):
:rocket:

Ch3LL:
@terminalmage would you mind fixing the merge conflicts and add a changelog. thanks!

Ch3LL:
thanks for resolving that conflict. It looks like there are failing tests on the OS tests and pre-commit.

Ch3LL:
Thanks. Let me know if I can help. I'm more than wiling to help wherever needed since this is not your original code.

terminalmage:
Thanks! I think this is probably fixed now, though. The unit tests added in this PR were migrated to the `pytests` subdir and those tests are not class-based like they were originally when I added the test case. So, `self` was being referenced in the test function, and that needed to be removed.

Ch3LL:
looks like the windows test failures are related

terminalmage:
There's no reason to run that test on windows, so I marked it to skip on Windows.

Ch3LL (approved):

PR #55590, opened 2019-12-10
Depends decorator version master

github-abcde:
This is https://github.com/saltstack/salt/pull/53658, rebased on master.

### What does this PR do?
It adds a `version` kwarg to the @depends decorator, which can handle any value that is able to be parsed with `salt.utils.version.LooseVersion`. The value of this version kwarg will be compared to the `__version__`-attribute of the module mentioned in the decorator (if it cannot be found (because it is a command or the module does not have the attribute), it will silently default to `None`). If the version is not equal or larger than the determined version (noting that when the latter is `None`, everything will be either equal (also `None`) or greater (any value)), the specific function will be removed by `enforce_dependencies`.

None that I know of

### Previous Behavior
The @depends decorator can only specify whether or not a module needs to be loaded.

### New Behavior
The @depends decorator can also specify a specific minimum version for the module that needs to be loaded.

### Tests written?
No, there are tests in `tests/integration/modules/test_decorators.py`, but I have no idea how those work.

### Commits signed with GPG?
Yes


salt/utils/decorators/__init__.py
```diff
@@ -179,17 +184,31 @@ def enforce_dependencies(cls, functions, kind, tgt_mod):
...
+                        log.trace(
+                            'Dependency (%s) already loaded inside %s, skipping',
+                            dependency, mod_name
+                        )
+                        continue
 
                 log.trace(
-                    'Unloading %s.%s because dependency (%s) is not met',
-                    mod_name, func_name, dependency
+                    'Unloading %s.%s because dependency (%s%s) is not met',
```
waynew: Was this missing a `.`?
```suggestion
                    'Unloading %s.%s because dependency (%s.%s) is not met',
```

waynew:
Hey, thanks for this PR!

I went a-looking to see what I could see Re: tests:

In `tests/integration/files/file/base/_modules/runtests_decorators.py` there are a bunch of functions that have been decorated. The tests in `tests/integration/modules/test_decorators.py` use salt to call those functions when testing to make sure that they do the right thing.

I think similar functions could be added to the existing `depends` ones in there. In fact, I think adding an additional `depends` (or something) module in there (`tests/integration/files/file/base/_modules`) with a `__version__` attribute would make it possible to test that the version check works as intended. Though might need a `verisionless_depends` as well, given the check for `__version__` in this PR.

Let me know if you need any more help on this!

salt/utils/decorators/__init__.py
github-abcde: Nope, the 4th `%s` placeholder is filled with https://github.com/saltstack/salt/pull/55590/files#diff-7bd4f0af6512326b8c5a14e4ecb70b40R211 which starts with a space. Alternatively, I could move the space to this line (207) if you prefer.

dwoz (approved):

Issue #49414, opened 2018-08-29
Handle deprecated glance module for Neon release


rallytime:
- [ ] salt/modules/glance.py

The glance module has been replaced with glanceng.py.

Don't forget to remove the autodoc files + references, as well as the stale doc files on the doc build server.

PR #55592, opened 2019-12-10
Add deprecation warning path for glance state and execution module 

Ch3LL:
### What does this PR do?
Add deprecation warning path for glance state and execution module 

https://github.com/saltstack/salt/issues/49414

### Previous Behavior
glance module was set to be deprecated in the neon release

### New Behavior
glance module will not be deprecated till the Aluminium release. This is because the glance state module which uses the glance module was never set on a deprecation path. So both the glance execution and state modules are set for deprecation in Aluminium.

### Tests written?
No - adding deprecation warning

### Commits signed with GPG?

Yes


salt/modules/glance.py
```diff
@@ -104,7 +104,7 @@ def _auth(profile=None, api_version=2, **connection_args):
     Only intended to be used within glance-enabled modules
     '''
     __utils__['versions.warn_until'](
```
s0undt3ch: I know this is not your change but why are we preferring `__opts__` instead of the direct function call?
You know the reason behind that?

s0undt3ch (approved):

salt/modules/glance.py
s0undt3ch: `__utils__` not `__opts__`

Ch3LL: Personally I thought we always preferred `__utils__` where possible since its packed in via the loader instead of importing directly. Is there a reason not to?

s0undt3ch: Guess there really isn't a reason, I apparently just dislike it's usage, seems like mixing loader stuff with regular utils modules, but it's just my personal preference.

Issue #55593, opened 2019-12-10
pip state doesn't support salt global proxy


Oloremo:
### Description of Issue
There is an ability in Salt to sent minion wide http proxy:
#29322
#43764

https://docs.saltstack.com/en/latest/ref/configuration/minion.html#proxy-host

Setting this args will affect every module and state which is using a utils.http since it uses tornado by default and tornado is aware of those proxy options. pip module supports the separate `proxy` argument and I think kinda should support the system http_proxy variable but people report it's broken: https://github.com/saltstack/salt/issues/52186

I think the idea of global proxy is great I'd like other states to support it. So my proposal is to add a global proxy awareness for pip state.

**Right now:**
1) If `proxy` arg passed - add the proxy to the pip command

**Will be:**
1) If `proxy` arg passed and it's 'truthy" - add the proxy to the pip command
2) If `proxy` arg passed and it's `False` - disable proxy all together 
3) If proxy not passed BUT global proxy set - use it to set up pip proxy

So the `proxy` arg still will be top priority and global proxy as a fallback plus the option to disable the proxy in a strange case when you have global proxy set but you dont want it to be applied to this particular state.

I already have a working patch and willing to implement it.

Oloremo:
Feature merged!

PR #55596, opened 2019-12-10
Added a support for a global proxy to pip module

Oloremo:
### What does this PR do?
Added support for a global Salt proxy to pip module

#55593

### Previous Behavior
pip module and state doesn't support salt global proxy

### New Behavior
pip module and state does support salt global proxy

### Tests written?

Yes

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/modules/pip.py
+++ b/salt/modules/pip.py
@@ -481,6 +481,11 @@ def install(pkgs=None,  # pylint: disable=R0912,R0913,R0914
         behind an authenticated proxy. If you provide
         ``user@proxy.server:port`` then you will be prompted for a password.
 
+        .. note::
+            If the the Minion has a globaly configured proxy - it will be used
+            even if no proxy was set here. To explicitly disable proxy for pip
+            you should pass ``False`` as a value.
+
     timeout
         Set the socket timeout (default 15 seconds)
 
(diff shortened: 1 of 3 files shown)


salt/modules/pip.py
```diff
@@ -995,6 +1012,11 @@ def uninstall(pkgs=None,
         behind an authenticated proxy.  If you provide
         ``user@proxy.server:port`` then you will be prompted for a password.
 
+        .. note::
+            If the the Minion has a globaly configured proxy - it will be used
+            even if no proxy was set here. To explicitly disable proxy for pip
+            you should pass the ``False`` as a value.
```
s0undt3ch: 'you should pass ``False`` as'...

s0undt3ch (changes requested):

salt/modules/pip.py
Oloremo: tx. done

s0undt3ch (commented):
I would also do something like:
```python
    elif proxy is not False and config.get('proxy_host') and config.get('proxy_port'):
        if config.get('proxy_username') and config.get('proxy_password'):
            http_proxy_url = 'http://{proxy_username}:{proxy_password}@{proxy_host}:{proxy_port}'.format(**config)
        else:
            http_proxy_url = 'http://{proxy_host}:{proxy_port}'.format(**config)
```

Oloremo:
@s0undt3ch  Very cool, added!

s0undt3ch (approved):

Oloremo:
re-run full  opensuse15-py2

Oloremo:
re-run full ubuntu1604-py2

Oloremo:
re-run full macosxmojave-py2

Oloremo:
re-run full windows2019-py3

max-arnold:
Nitpick: it is worth documenting that the `no_proxy` global setting has no effect on pip

Issue #49428, opened 2018-08-29
Handle jinja filter deprecation notices for Neon release


rallytime:
- [ ] salt/utils/data.py
- [ ] salt/utils/hashutils.py
- [ ] salt/utils/stringutils.py
- [ ] salt/utils/jinja.py

PR #55614, opened 2019-12-11
Deprecate jinja filters for neon

Ch3LL:
### What does this PR do?
- in this pr https://github.com/saltstack/salt/pull/48339 a note was added in the jinja docs that the `tojson` jinja filter would replace `json_encode_dict` and `json_encode_list`, but it was never added to the release notes for a release to properly deprecate it. This PR will add this notification in the neon release notes for `json_decode_*` as this was improperly named. we will keep `json_encode_*` as it still makes sense and we want to decrease users having to update their states.

- we want to keep `str_to_num` and `rand_str` to ensure users do not have to update their states, but its already documented they these are no the preferred method.

Fixes issue https://github.com/saltstack/salt/issues/49428


s0undt3ch (approved):

dwoz:
re-run full all

s0undt3ch (approved):

Issue #52567, opened 2019-04-16
pillar targeting using pillar with git_pillar backend


vquiering:
### Description of Issue/Question
Whenever I try to target minions using pillar in pillar top.sls with `git_pillar` as backend I get the following error message.

```
2019-04-16 22:58:51,680 [salt.pillar      :984 ][ERROR   ][17607] Exception caught loading ext_pillar 'git':
  File "/usr/lib/python2.7/dist-packages/salt/pillar/__init__.py", line 974, in ext_pillar
    key)
  File "/usr/lib/python2.7/dist-packages/salt/pillar/__init__.py", line 902, in _external_pillar_data
    *val)
  File "/usr/lib/python2.7/dist-packages/salt/pillar/git_pillar.py", line 474, in ext_pillar
    local_pillar.compile_pillar(ext=False),
  File "/usr/lib/python2.7/dist-packages/salt/pillar/__init__.py", line 1018, in compile_pillar
    matches = self.top_matches(top)
  File "/usr/lib/python2.7/dist-packages/salt/pillar/__init__.py", line 675, in top_matches
    self.opts.get('nodegroups', {}),
  File "/usr/lib/python2.7/dist-packages/salt/matchers/confirm_top.py", line 36, in confirm_top
    return m(match)
  File "/usr/lib/python2.7/dist-packages/salt/matchers/compound_match.py", line 97, in match
    six.text_type(matchers['{0}_match.match'.format(engine)](*engine_args, **engine_kwargs))
  File "/usr/lib/python2.7/dist-packages/salt/matchers/pillar_match.py", line 26, in match
    __opts__['pillar'], tgt, delimiter=delimiter

2019-04-16 22:58:51,682 [salt.pillar      :1036][CRITICAL][17607] Pillar render error: Failed to load ext_pillar git: u'pillar'
```
In `salt/matchers/pillar_match.py` I edited the corresponding line as following since I couldn‘t find `'pillar'` in `__opts__`.
```
diff --git a/salt/matchers/pillar_match.py b/salt/matchers/pillar_match.py
index 5bb9cc8ae2..37aeb643f3 100644
--- a/salt/matchers/pillar_match.py
+++ b/salt/matchers/pillar_match.py
@@ -21,5 +21,5 @@ def match(tgt, delimiter=DEFAULT_TARGET_DELIM):
                   'statement from master')
         return False
     return salt.utils.data.subdict_match(
-        __opts__['pillar'], tgt, delimiter=delimiter
[...]

PR #55624, opened 2019-12-12
fallback to ext_pillar in pillar match if there is no pillar in __opts__

vquie:
### What does this PR do?
If there is 'pillar' in `__opts__` the matchers would break. This PR determines if there is a 'pillar' object and uses 'ext_pillar' if there is none.

#52567

### Previous Behavior
Previously the matcher would just call `__opts__['pillar']`.

### New Behavior
The matcher checks for `__opts__['pillar']`. If there is none, it calls `__opts__['ext_pillar']` instead.

### Tests written?
No

### Commits signed with GPG?
No

### DIFF
--- a/salt/matchers/pillar_exact_match.py
+++ b/salt/matchers/pillar_exact_match.py
@@ -22,7 +22,12 @@ def match(tgt, delimiter=':', opts=None):
                   'statement from master')
         return False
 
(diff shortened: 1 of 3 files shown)


dwoz:
re-run full centos

dwoz (approved):

PR #55636, opened 2019-12-13
Backport #49338 to master

terminalmage:
See #49338.


dwoz:
The windows tests passed but the build failed due to some file handle issues we need to resolve in the test suite.

```
OK (total=9293, skipped=1213, passed=8080, failures=0, errors=0)
```

Merging

PR #55639, opened 2019-12-13
Loop until no eval master

github-abcde:
This is https://github.com/saltstack/salt/pull/53958, rebased on master.

### What does this PR do?
It provides an alternative to `loop.until` that does not use `eval`.
Added example for using this in the faq. Made some pylint-inspired changes.
I'm not entirely sure that the name of the new state `until_no_eval` is good enough (even though it is accurate), since it feels wrong to name a function with something that it does *not* have. I'm open to suggestions on this.

None

### Previous Behavior
The generic state that allows you to (blockingly) execute a salt function until it produces a desired output hinged on using `eval` with user-input. Since this is potentially very dangerous, an alternative has been supplied.

### New Behavior
A state exists (`loop.until_no_eval`) that will block execution until a specific salt execution module function returns an expected result. This produced result is passed together with the expected result to a function (either a method of python's `operator`, a function from `__salt__` or `__utils__`) to produce a boolean indicating success (or failure).
Additionally, a parameter `init_wait` is available to configure a waiting period before the 1st call of the salt function is made.

### Tests written?
Yes

### Commits signed with GPG?
Yes


waynew (approved):
Looks great! Just merged the most recent changes in, but this should be ready to go once tests pass.

salt/states/loop.py
```diff
@@ -85,39 +104,121 @@ def until(name,
...
+    :param timeout: Abort after this amount of seconds (excluding init_wait).
+    :type timeout: int or float
+    :param period: Time (in seconds) to wait between attempts.
+    :type period: int or float
+    :param init_wait: Time (in seconds) to wait before trying anything.
+    :type init_wait: int or float
+    :param list args: args to pass to the salt module.function.
+    :param dict kwargs: kwargs to pass to the salt module.function.
+
+    .. versionadded:: Natrium
```
dwoz: This version should be `3000`

dwoz (changes requested):

salt/states/loop.py
github-abcde: Changed in cb6b22c29c34f3081b7767875d3061d1ea32341b

github-abcde:
re-test pr-amazon2-py2

github-abcde:
re-run pr-amazon2-py2

github-abcde:
re-run pr-windows2019-py3

dwoz (approved):

PR #54808, opened 2019-09-30
support recursive globbing

angeloudy:
I believe `os.path.basename()` limited the functionality of glob.
By removing `os.path.basename(fi)`, it becomes more flexible and can now support the following matching

`salt-call cp.push_dir /a/b glob='**/b/*.txt'`, while still supports the old matching

`salt-call cp.push_dir /a/b glob='*.txt'`

In my case, I have many exported files(with extension `*.groovy`)from my application inside FreeBSD Jails. 
```
/jails/jail1/usr/local/myapp/export
/jails/jail2/usr/local/myapp/export
/jails/jail3/usr/local/myapp/export
/jails/jail3/usr/local/myapp/export
```
I wanted to do
`salt minion1 cp.push_dir /jails glob='**/export/*.groovy'`,
which is achieved by removing `os.path.basename`

PR #55688, opened 2019-12-18
glob on absolute path instead of relative path

angeloudy:
### What does this PR do?
Resubmitting PR #54808 to master branch

### Tests written?

Not needed

### Commits signed with GPG?

Yes


dwoz:
re-run full centos

dwoz:
@angeloudy Can you write a test for this change?

DmitryKuzmenko:
Here's the PR where the basename match was added https://github.com/saltstack/salt/pull/27917
There's no details wy. I'm just not sure how the change could affect others. Maybe it's better to make the full path match an option? @saltstack/team-core

DmitryKuzmenko:
@angeloudy thank you for contribution. We definitely need tests here as DWos asked. Also could you please install the pre-commit hook on your repo and update this PR?

waynew:
I was able to track the discussion down to https://groups.google.com/g/salt-users/c/qZvmtdGxLKg/m/awL8FwfiCQAJ

> eg
`salt-call cp.push_dir /etc/sysconfig/network-scripts glob='ifcfg-*'`
> does not work, but:
`salt-call cp.push_dir /etc/salt/minion.d glob='*.conf'`
> does.
> Bug or feature? (It looks buggy to me, and it sure is annoying.)
Using 2015.5.3 (Lithium).

I'll probably go ahead and write some tests for this during Tuesday's test clinic

waynew:
Writing tests currently - just discovered that this desired behavior would definitely break backwards compatibility, the aforementioned intended behavior. However, we can introduce a flag - `glob_recurse=True` for example, that would allow *this* desired behavior.

waynew:
I went ahead and added test cases for this. As previously mentioned, it would be impossible/unfeasible to try and both maintain backwards compatibility along with having the recursive capabilities. This introduces the ability to use recursive globbing, but doesn't clobber any of the existing intentional functionality.

Akm0d (dismissed):

waynew:
Doc build failed. Fixed docs as well as went ahead and Boy Scout'ed the rest of the docs in this module. Should be good *now*.

Ch3LL:
Looks like some test failures on windows here still

sagetherage:
Removing from `Aluminium` for now

dwoz:

Issue #55662, opened 2019-12-16
S3 External Pillar Missing Data


Linuturk:
### Description of Issue
Pillar data stored in S3 isn't found for all minions, even after clearing cache and forcing a refresh. https://docs.saltstack.com/en/latest/ref/pillar/all/salt.pillar.s3.html 

### Setup
Configure salt master to use S3 as a pillar source:
```
ext_pillar:
  - git:
    - __env__ ssh://git@mygitserver.com/pillar.git:
      - pubkey: '/etc/salt/ssh/saltmaster-id_rsa.pub'
      - privkey: '/etc/salt/ssh/saltmaster-id_rsa'
  - s3:
      bucket: mys3bucket
      prefix: pillar
```

Write your pillar data to S3. We have a unique pillar setup for each minion in the format of `minionid/init.sls`. We currently have 2249 pillar folder/file combinations in the bucket.

top.sls:
```
{{saltenv}}:
  {{opts.id}}:
    - {{opts.id}}
    - ignore_missing: True
```

### Steps to Reproduce Issue
Call `pillar.items` on a given minion and see that all the pillar data from S3 isn't present. Only the data from our other pillars.

Use `saltutil.refresh_pillar` return clean, and still show empty for the S3 data.

Use `salt-run cache.pillar` and see the S3 data isn't cached. Clear the cache with `salt-run cache.clear_pillar` and refresh again, and still don't see the data.

Notice that this data is available for some minions, but not all.

### Versions Report
```
Salt Version:
           Salt: 2019.2.2
 
Dependency Versions:
           cffi: 1.11.5
       cherrypy: unknown
       dateutil: 2.6.1
      docker-py: Not Installed
          gitdb: 2.0.3
      gitpython: 2.1.8
          ioflo: Not Installed
         Jinja2: 2.10
        libgit2: 0.26.0
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: 1.0.7
   msgpack-pure: Not Installed
 msgpack-python: 0.5.6
   mysql-python: Not Installed
      pycparser: 2.18
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: 0.26.2
         Python: 2.7.15+ (default, Oct  7 2019, 17:39:04)
   python-gnupg: 0.4.1
         PyYAML: 3.12
          PyZMQ: 16.0.2
           RAET: Not Installed
          smmap: 2.0.3
        timelib: Not Installed
        Tornado: 4.5.3
[...]

PR #55694, opened 2019-12-18
[master] S3 Pillar Pagination Fixes

garethgreenaway:
### What does this PR do?
Fixing S3 pillar when the keys in a bucket exceed 1000, ensure we are handling pagination properly

#55662 

### Tests written?

Not Yet.

### Commits signed with GPG?
Yes


salt/pillar/s3.py
```diff
@@ -312,6 +324,14 @@ def __get_pillar_environments(files):
         if s3_meta:
             bucket_files[bucket] = __get_pillar_files_from_s3_meta(s3_meta)
 
+            # Check if we have a NextContinuationToken and loop until we don't
+            while True:
+                continuation_token = __get_continuation_token(s3_meta)
```
waynew: Could this be rewritten as an iterator?
```
for continuation_token in __get_continuation_tokens(s3_meta):
    ...
```

I think the function would just be:

```
for item in s3_meta:
    if item.get('NextContinuationToken'):
        yield item('NextContinuationToken')
```

Or something to that effect.

garethgreenaway: Not sure if that approach will work.  There is only ever one `NextContinuationToken` in s3_meta, but it could be different when s3_meta is updated after subsequent calls to `__get_s3_meta`.  Once it's not longer in s3_meta then it should exist the loop since there are no more results to pull down.

waynew: Oooooh. I see!

What about something like this?

```

    # helper s3 query function
    def __get_s3_meta():
        params = {'prefix': prefix, 'list-type': 2}
        while True:
            result = __utils__['s3.query'](
                key=creds.key,
                keyid=creds.keyid,
                kms_keyid=creds.kms_keyid,
                bucket=creds.bucket,
                service_url=creds.service_url,
                verify_ssl=creds.verify_ssl,
                location=creds.location,
                return_bin=False,
                params={'prefix': prefix},
                path_style=creds.path_style,
                https_enable=creds.https_enable)
            yield result
            continuation_token = next((item.get('NextContinuationToken')
                                       for item in s3_meta
                                       if item.get('NextContinuationToken')),
                                      None)
            if continuation_token is None:
                break
            else:
                params['continuation-token'] = continuation_token
```

Then it could simply be

```
for s3_meta in __get_s3_meta():
    ...
```

Would that work?

waynew (approved):

Issue #55443, opened 2019-11-26
archive.extracted downloads the file even if checksums are the same


Oloremo:
### Description of Issue
Right now `archive.extracted` looks for cached copy of the file on a minion and cache it(download file) if it's not present in the minion local cache to do an archive list and check the archive files against the destination directory. I think it's overkill in some cases.

Example:
```
extract:
  archive.extracted:
    - name: /tmp/archive
    - source: http://path/to/archive.gz
    - source_hash: http://path/to/archive.gz.md5
    - source_hash_update: True
    - clean: True
    - keep_source: False
```
What is going to happen in the example above first time you run it:

1. Salt will search for the `http://path/to/archive.gz.hash` in local minion cache. There is none.
2. Salt will download `http://path/to/archive.gz` into minion cache and will create a `http://path/to/archive.gz.hash` file with checksum of the file.
3. Salt will call a `archive.list` module to compare the output of the archive contents with the destination directory and if there is a difference(there is in our case) it will extract it.
4. Since `keep_source` set to  `False` local cached file `http://path/to/archive.gz` will be removed, yet `http://path/to/archive.gz.hash` won't be.

So far so good. Now let's see what is going to happen if we run this state once again assuming nor `archive.gz` nor `archive.gz.md5` changed:

1. Salt will search for the `http://path/to/archive.gz.hash` in local minion cache. There is such file.
2. Salt will compare it with the hash from the `http://path/to/archive.gz.md5`. **They will match**.
3. Salt will download `http://path/to/archive.gz` into minion cache since there is no such file due to  `keep_source: False`
4. Salt will call a `archive.list` module to compare the output of the archive contents with the destination directory and if there is a difference(there is none in our case).
5. Since `keep_source` set to  `False` local cached file `http://path/to/archive.gz` will be removed, yet `http://path/to/archive.gz.hash` won't be.

And this re-download of the `archive.gz` will be done over and over again. And now imagine it's 1000 nodes and archives are 1Gb+.

What is the alternative? `keep_source: True`. It solves the issue by the cost of having file cache on all minions which **will never be cleaned** due to https://github.com/saltstack/salt/issues/34369 not implemented. And now imagine it's 1000 nodes and archives are 1Gb+.

So right now it's a choice between re-downloading it every time OR keeping all archives in the cache without any ability to purge it efficiently.

My proposal is to add an optional argument like `trust_source_hash: BOOL` which will change the behavior to this one:
[...]

PR #55700, opened 2019-12-19
Added a skip_files_list_verify argument to archive.extracted state

Oloremo:
### What does this PR do?

This PR adding an ability to improve the archive.extracted state performance and traffic efficiency by adding an optional argument `skip_files_list_verify` which will allow to skip the archive file list validation replacing it by checksum validation.

#55443

### Previous Behavior

Before that, it was a choice between re-downloading source every time OR keeping all archives sources in the cache without any ability to purge it efficiently.

### New Behavior

Setting `skip_files_list_verify` to `True` will enforce the checksum check and if the checksums are match - we're done here.

### Tests written?

Yes

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/states/archive.py
+++ b/salt/states/archive.py
@@ -437,7 +437,7 @@ def extracted(name,
             minions ``hash_type`` config option to the same one that you're going
             to pass via ``source_hash`` argument.
 
-        .. versionadded:: Sodium
+        .. versionadded:: Neon
 
     skip_verify : False
         If ``True``, hash verification of remote file sources (``http://``,
@@ -553,7 +553,7 @@ def extracted(name,
         then re-create that directory before extracting. Note that ``clean``
         and ``clean_parent`` are mutually exclusive.
 
-        .. versionadded:: Sodium
+        .. versionadded:: Neon
 
     user
         The user to own each extracted file. Not available on Windows.
--- a/salt/states/archive.py
+++ b/salt/states/archive.py
@@ -176,6 +176,7 @@ def extracted(name,
               source_hash=None,
               source_hash_name=None,
               source_hash_update=False,
+              skip_files_list_verify=False,
               skip_verify=False,
               password=None,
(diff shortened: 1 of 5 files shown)


Oloremo:
```
salt/utils/ctx.py | 87.5% <0%> (-8.33)
```
So I'm reducing code coverage of files I didn't even touch. That's my secret superpower!

salt/states/archive.py
```diff
@@ -411,14 +412,33 @@ def extracted(name,
...
+        Set this to ``True`` if archive should be extracted if source_hash has
+        changed but only checksums of the archive will be checked to determine if
+        the extraction is required.
+
+        .. note::
+            The current limitation of this logic is that you have to set
+            minions ``hash_type`` config option to the same one that you're going
+            to pass via ``source_hash`` argument.
+
+        .. versionadded:: Sodium
```
Oloremo: Neon?

dwoz: yes

salt/states/archive.py
```diff
@@ -533,7 +553,7 @@ def extracted(name,
         then re-create that directory before extracting. Note that ``clean``
         and ``clean_parent`` are mutually exclusive.
 
-        .. versionadded:: Sodium
+        .. versionadded:: Neon
```
Oloremo: not related to this PR, just a mistake from my previous one. I hope it's ok to fix it here too

Oloremo:
I tested this logic on some scenarios like a remote source, local source, remote source hash, source_hash as a string - all seems good. not sure about all possible edge cases - salt file caching is complicated. But this is an optional feature disabled by default so I hope it's ok.

What I'm worried about in the docs. First of all, I'm not sure if I described it clearly.

Oloremo:
re-run full  macosxmojave-py2

dwoz (approved):

PR #55707, opened 2019-12-20
Create a wrapper that checks for the presence of state modules for tests

Akm0d:
### What does this PR do?
Create a wrapper that checks for the presence of state modules for tests


tests/support/helpers.py
```diff
@@ -1064,6 +1084,69 @@ def decorator(*args, **kwargs):
...
         return func(*args, **kwargs)
+
+    return decorator
+
+
+def requires_salt_state_modules(*names):
+    '''
+    Makes sure the passed salt state module is available. Skips the test if not
+
+    .. versionadded:: 0.5.2
```
s0undt3ch: Version added, 3000

s0undt3ch (changes requested):

Akm0d:
re-run full all

tests/integration/states/test_pkg.py
```diff
@@ -28,12 +29,9 @@
...
 from salt.ext import six
 from salt.ext.six.moves import range  # pylint: disable=import-error,redefined-builtin
+from tests.support.sminion import build_minion_opts
 
-try:
-    from distro import LinuxDistribution
-    pre_grains = LinuxDistribution()
-except ImportError:
-    pre_grains = None
+_grains = salt.loader.grains(build_minion_opts(minion_id='runtests-internal-sminion'))
```
s0undt3ch: Please, no dictionaries at the global scope.
If you need them to build the package list, maybe that should actually live in `setUpClass`?

s0undt3ch (changes requested):

tests/integration/states/test_pkg.py
s0undt3ch: On top of being a dictionary, loading grains is not a blazing fast operation and will run just by importing the test module, this alone will slow down pytest test collecting.

tests/integration/states/test_pkg.py
```diff
@@ -29,73 +31,50 @@
...
-            return caller(cls)
-
-        return wrapper
-
-    return decorator
-
 
 @destructiveTest
 class PkgTest(ModuleCase, SaltReturnAssertsMixin):
+    _PKG_EPOCH_TARGETS = []
```
s0undt3ch: There's really no reason why these class attribute aren't also set in `setUpClass` right?

tests/integration/states/test_pkg.py
```diff
@@ -583,46 +562,52 @@ def test_pkg_015_installed_held(self):
             refresh=False,
         )
 
+        if versionlock_pkg and '-versionlock is not installed' in str(ret):
+            self.skipTest(str(ret) + '  `{}` is installed'.format(versionlock_pkg))
```
s0undt3ch: Single formatted string here too.

s0undt3ch (changes requested):

tests/integration/states/test_pkg.py
Akm0d: The `skipIf` decorator doesn't have access to `self`.  These need to be declared outside of `setUpClass` so that certain tests can be skipped if they are empty.

tests/support/helpers.py
```diff
@@ -1067,6 +1067,111 @@ def decorator(*args, **kwargs):
...
+                        cls.skipTest('This test does not run on {}={}'.format(kw, grains.get(kw)))
+                else:
+                    if str(grains.get(kw)).lower() == str(value).lower():
+                        cls.skipTest('This test does not run on {}={}, got {}'.format(kw, value, grains.get(kw)))
+            return caller(cls)
+        return wrapper
+    return decorator
+
+
+def requires_salt_state_modules(*names):
```
dwoz: Can this be `requires_salt_states` instead?

tests/support/helpers.py
```diff
@@ -1067,6 +1067,111 @@ def decorator(*args, **kwargs):
...
+    '''
+    Makes sure the passed salt state module is available. Skips the test if not
+
+    .. versionadded:: 3000
+    '''
+
+    def _check_required_salt_state_modules(*required_salt_modules):
```
dwoz: Does this have to be a nested helper function? I'd prefer something like this live at the top level of the module.

dwoz (changes requested):

tests/support/helpers.py
Akm0d: Was just following the format of `requires_salt_modules`, they share quite a bit of code though and it certainly could sit at the top level

tests/integration/states/test_pkg.py
s0undt3ch: Ok

tests/support/helpers.py
```diff
@@ -1067,50 +1067,135 @@ def decorator(*args, **kwargs):
...
+                        cls.skipTest('This test does not run on {}={}'.format(kw, grains.get(kw)))
+                else:
+                    if str(grains.get(kw)).lower() == str(value).lower():
+                        cls.skipTest('This test does not run on {}={}, got {}'.format(kw, value, grains.get(kw)))
+            return caller(cls)
+        return wrapper
+    return decorator
+
+
+def check_required_sminion_attributes(sminion_attr, *required_items):
```
s0undt3ch: I understand why Daniel wanted this as a top level function, specially now that its used on 2 different helpers, but let's prefix it with a `_`.
This way people aren't tempted to use it directly and will create the right `requires_salt_*` helper function.

s0undt3ch (changes requested):

tests/support/helpers.py
Akm0d: I noticed this function might be usable in conftest.py since it contains an identical code block there, but since helpers.py imports from unittest is that a good idea?  Maybe there is another place this can go so that it is usable by conftest.py

s0undt3ch: For now, leave it here.
Once we go PyTest, we'll rearrange.

Akm0d:
re-run full windows2019

s0undt3ch (approved):

waynew (approved):

terminalmage:
This PR caused several tests to always be skipped. I discovered this while writing another test earlier tonight. Details are in https://github.com/saltstack/salt/pull/56681, and I've fixed the problem there.

Just a friendly reminder that whenever changes are made to test skipping behavior, it's that much more important to closely check the results for the affected tests, on *all* platforms, and not to assume that everything is OK because tests are green. :smiley:

PR #55717, opened 2019-12-20
[WIP] Async `run_function` with extra debugging information in case of failures

s0undt3ch:
### What does this PR do?
See title.


tests/support/case.py
```diff
@@ -737,7 +737,7 @@ def minion_run(self, _function, *args, **kw):
         '''
         return self.run_function(_function, args, **kw)
 
-    def run_function(self, function, arg=(), minion_tgt='minion', timeout=300, master_tgt=None, **kwargs):
+    def run_function(self, function, arg=(), minion_tgt='minion', timeout=60, master_tgt=None, **kwargs):
```
dwoz: Increasing this timeout drastically reduced flakiness in the test suite. I'm not sure that I am on board with changing it.

s0undt3ch: Opened mostly as a test to see what breaks. And it broke a few tests.

My problem with the timeout, is not actually the timeout. It's the timeout plus the flaky decorator which defaults to 4 times. That means waiting for 20 minutes for a failure. I'm seeing that happen on the 2019.2.3 branch.

In fact this PR should have been submitted two fold, one with the timeout reduction, to see what actually needs this much time and explcitly pass a bigger timeout, the other trying the async approach.

My ultimate goal is to see if we can identify the stages of the job and act acordingly. When an assert fails we of get true is not false. So, if we could identify that, first, the job was published, second, the job was picked up by the minion and third, did the job complete or not, that would help with these unuseful assertions...

So, take this with a grain of salt, I'm not ready to argue for this cause yet, I'm still not sold on it :smile:

dwoz: I agree we need to remove the flaky decorators and reduce the overall flakiness of tests. Ideally, we would not *need* such large timeouts. However, I'm not in favor of changing the timeout time unless we're sure it won't introduce the flakiness which it resolved back into the test suite.

s0undt3ch: I've renamed the PR.
Right now, it's not supposed to be merged.
I'll let you know if this changes.

s0undt3ch: Removed the timeout reduction changes.

PR #55720, opened 2019-12-21
attempt at fixing 25721 on master

### LINKED ISSUE #25721: Strange "Permission denied" error with "/etc/apache2/sites-available/default-ssl.conf" with salt-ssh
I noticed a strange error when I tried to write a state which manages `/etc/apache2/site-available/default-ssl.conf`.
The state failed with:

```
Permission denied for host <hostname>, do you want to deploy the salt-ssh key? (password required):
```

This can be more easily reproduced by simply reading the file from the remote system: `salt-ssh '*' -r 'cat /etc/apache2/sites-available/default-ssl.conf'` this gives also:

```
Permission denied for host <hostname>, do you want to deploy the salt-ssh key? (password required):
[Y/n]
```

I can read other files easily this way, and on the remote system, I can read the default-ssl.conf also with the user I'm using with salt-ssh, so it is not a permission problem.
It rather seems to me that something in this file is kind of interpreted by salt-ssh or during the transfer or so. Renaming the file and also moving into another location didn't help, so I guess it is related to the contents but couldn't find anything obvious which is causing this.

Version information:
- Client: `salt-ssh 2015.5.3 (Lithium)` installed in a virtualenv using pip
- Remote system: Ubuntu 14.04 LTS

Full debug log:

```
[INFO    ] Loading Saltfile from '/tmp/salttest/Saltfile'
[DEBUG   ] Reading configuration from /tmp/salttest/Saltfile
[DEBUG   ] Reading configuration from /tmp/salttest/master
[DEBUG   ] Configuration file path: /tmp/salttest/master
[DEBUG   ] LazyLoaded flat.targets
[DEBUG   ] LazyLoaded jinja.render
[DEBUG   ] LazyLoaded yaml.render
[DEBUG   ] Jinja search path: ['/tmp/salttest/cache/files/base']
[DEBUG   ] Rendered data from file: /tmp/salttest/roster:
<hostname>:
  host: <ip_address>
  port: 22
  user: enrico

[DEBUG   ] Results of YAML rendering: 
OrderedDict([('<hostname>', OrderedDict([('host', '<ip_address>'), ('port', 22), ('user', 'enrico')]))])
[INFO    ] minions list: {'<hostname>': OrderedDict([('host', '<ip_address>'), ('port', 22), ('user', 'enrico')])}
[DEBUG   ] LazyLoaded roots.envs
[DEBUG   ] Could not LazyLoad roots.init
[DEBUG   ] Updating roots fileserver cache
[DEBUG   ] LazyLoaded local_cache.prep_jid
[DEBUG   ] Executing command: ssh <ip_address>  -o KbdInteractiveAuthentication=no -o PasswordAuthentication=no -o GSSAPIAuthentication=no -o ConnectTimeout=65 -o Port=22 -o IdentityFile=/tmp/salttest/pki/ssh/salt-ssh.rsa -o User=enrico  cat\ \/etc\/apache2\/sites-available\/default\-ssl\.conf
[DEBUG   ] Child Forked! PID: 25328  STDOUT_FD: 8  STDERR_FD: 10
[DEBUG   ] Terminal Command: /bin/sh -c ssh <ip_address>  -o KbdInteractiveAuthentication=no -o PasswordAuthentication=no -o GSSAPIAuthentication=no -o ConnectTimeout=65 -o Port=22 -o IdentityFile=/tmp/salttest/pki/ssh/salt-ssh.rsa -o User=enrico  cat\ \/etc\/apache2\/sites-available\/default\-ssl\.conf
Permission denied for host <hostname>, do you want to deploy the salt-ssh key? (password required):
[Y/n] 
```

This might be related to https://github.com/saltstack/salt/issues/18310#issuecomment-98094716 as the error message is the same.

mchugh19:
### What does this PR do?
Adjusts salt-ssh detection with an updated regex. New behavior assumes password prompt will not have content after `password:`
Port of #53714 to master

Attempts to sort of fix #25721

### Previous Behavior
See diff

### New Behavior
See diff

### Tests written?
Yes - **but please comment as there is a new pip dependency**
Existing test adjusted to trigger this behavior. This is complicated by salt-ssh commands being spawned by a subprocess. Thus, the usual `self.run_ssh(whatever, timeout=60)` doesn't help as the previous behavior of pausing for input for an ssh password: prompt would hang the test processes. To progress, I used the timeout-decorator pip module, but this adds an additional dependency to the test framework. Perhaps there's a better method? 

### Commits signed with GPG?
No


mchugh19:
As mentioned in the PR description, this currently adds a new test dependency of timeout-decorator. Comments welcome on if this is the best way to support ssh test hangs.

tests/integration/ssh/test_raw.py
```diff
@@ -10,16 +10,19 @@
...
     '''
     testing salt-ssh with raw calls
     '''
+    @timeout_decorator.timeout(60, use_signals=False)
     def test_ssh_raw(self):
         '''
         test salt-ssh with -r argument
         '''
-        msg = 'running raw msg'
+        msg = 'password: foo'
```
Ch3LL, line 26: can we ensure we are testing for both of these msg types?

Also can you elaborate why you need this decorator?

Ch3LL (changes requested):

tests/integration/ssh/test_raw.py
mchugh19: Is there any need to test for the text `running raw msg`? It seems to just be some text to test for, where as this PR puts it to use as also triggering the #25721 bug.

mchugh19:
The decorator dependency isn't about the decorator itself, so much as the need to have the functionality to kill a hung subprocess in the test suite. So far this PR does it with the timeout-decorator pip module, but perhaps there is already some functionality like this in the salt test suite someplace? 

The issue is simply that the pre-PR behavior causes pausing for input for an ssh password: prompt which then hangs the test processes. I used the timeout-decorator module to detect this scenario and kill off the waiting for input code. If there's a better method, happy to switch over.

sagetherage:
@mchugh19 looks like there are some merge conflicts, are you able to address those this week?

s0undt3ch:
Could you please rebase and resolve the conflicts?

mchugh19:
Updated for master.
@s0undt3ch the open question on this PR is how you want to handle timeouts of ssh tests. The bug this PR is attempting to correct is one where ssh connections hang when finding a string which appears to be a password prompt. Since this also hung the test, I used the timeout-decorator import which seems to kill off the stalled salt-ssh process. Is that new dependency the best way of handling this situation?

mchugh19:
@dwoz @sagetherage @s0undt3ch this could really use some feedback.

I've added a new test dependency in order to detect a hung ssh session. If that's the proper fix, then that dependency needs to be added to the test suite for the tests to pass. If there's a better way to detect, then I need to know and update the test. 

I've been asking about this since the original PR was opened, but have not received a response yet and I don't think the `failing-test` label applies here. Just let me know how you want to proceed.

sagetherage:
@mchugh19 I am still working to get you some feedback on this from yesterday, and will continue. updating here to let you know, this isn't going unnoticed.

s0undt3ch:
Sorry @mchugh19, I'll get to this first thing tomorrow morning.

s0undt3ch:
Added commit which ads `timeout_decorator` as a testing dependency

mchugh19:
Thanks @s0undt3ch!

There are test failures, but they don't seem to be related to this PR. Maybe there's still some gremlins to be worked out.

s0undt3ch (dismissed):

sagetherage:
rebasing

sagetherage:
I have to remove this from Magnesium, but would love to get it into the next release cycle

Ch3LL (changes requested):
looks like there is just a merge conflict that needs to be resolved on this one. Anything I can help with?

Ch3LL:
Also forgot to mention this will need a changelog file.

garethgreenaway (approved):

PR #55729, opened 2019-12-23
[DOCS] Add runas info to cmd.script for windows server

xeacott:
### What does this PR do?
Add documentation to cmd.script that indicates Window's Server users may need to use
a different format when specifying runas parameters.

Specifically, using the style `domain\user` has caused issues, whereas using the style `user@domain.local` is the correct way. 

https://github.com/saltstack/salt/issues/55080

### Tests written?
No

### Commits signed with GPG?
Yes


dwoz (approved):

dwoz:
@xeacott Docs are failing

dwoz (approved):

PR #55731, opened 2019-12-23
Handle exception in `salt.utils.args.format_call()`

eliasp:
Re-based #54345 onto `master`…

### What does this PR do?

Without handling this exception, a SaltClient might hang in a loop waiting for a timeout and providing no clue as to why the call hangs/fails.
Handling this exception and logging a corresponding error msg improves this situation.

### Previous Behavior
A SaltClient would hang/loop like this:
```
[DEBUG   ] Connecting the Minion to the Master URI (for the return server): tcp://127.0.0.1:54519
[DEBUG   ] Trying to connect to: tcp://127.0.0.1:54519
[DEBUG   ] SaltReqTimeoutError, retrying. (1/3)
[DEBUG   ] SaltReqTimeoutError, retrying. (2/3)
[DEBUG   ] SaltReqTimeoutError, retrying. (3/3)
```
### New Behavior
**CLI:**
```
Authentication failure of type "eauth" occurred for user $USER
```
**Master Log:**
```
[ERROR   ][354] 'format_call' failed: SaltInvocationError('auth takes at least 2 arguments (1 given)',)
```

### Tests written?
No

As it's still quite unclear why we're seeing a certain external auth issue here, there's no way yet to define a reproducable positive/negative scenario which could be used for testing this.
This fix just helps us to further narrow down our problem and ensures no-one else will waste time on the red herring `SaltReqTimeoutError` in this scenario anymore.

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/auth/__init__.py
+++ b/salt/auth/__init__.py
@@ -30,6 +30,7 @@ import salt.utils.minions
 import salt.utils.user
 import salt.utils.versions
 import salt.utils.zeromq
+from salt.exceptions import SaltInvocationError
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


dwoz:
@eliasp Please address the test failures.

dwoz:

PR #55732, opened 2019-12-24
Enable salt-minion service auto restart (Linux).

Denistro:
### What does this PR do?
Enables salt-minion service auto restart
Salt-minion doesn't get up after falling down.
### Commits signed with GPG?
No


vtolstov:
2s is too fast
```
Restart=always        
RestartSec=5s         
StartLimitInterval=0  
```

dwoz:
@Denistro Why is the minion crashing?

Denistro:
> 
> 
> @Denistro Why is the minion crashing?
Didn't investigate it - just found that almost all my test Linux minions (>20) with long uptime (>1 month) went down. 
Minions OS setup - mostly CentOS 7 and couple  openSUSE 15.0
salt-minion 2019.2.0 (Fluorine)

It was a surprise for me that salt-minion service has no restart option, cause it could be really critical for prod.

nicholasmhughes:
I don't think `always` is a good idea...

> If set to `always`, the service will be restarted regardless of whether it exited cleanly or not, got terminated abnormally by a signal, or hit a timeout.

`on-failure` is probably a better bet so we're only restarting with problems are encountered, instead of when we're stopping the service for a valid reason.

> If set to `on-failure`, the service will be restarted when the process exits with a non-zero exit code, is terminated by a signal (including on core dump, but excluding the aforementioned four signals), when an operation (such as service reload) times out, and when the configured watchdog timeout is triggered.

FWIW, a lot of unit files don't make the assumption of criticality within your infrastructure. That's what [drop-ins](https://wiki.archlinux.org/index.php/systemd#Drop-in_files) are for.

vtolstov:
Restart always is very useful, you can change config from salt, shutdown minion from salt, and it restarted automatic with never config.

nicholasmhughes:
I don't doubt it's usefulness. I agree that some environments would benefit from that kind of configuration. I just don't think it should be the **default** because having the service restart magically when a user stopped it would catch most people off guard.

vtolstov:
Can you provide info, how user can stop salt minion, without using systemd? If he is stop it via systemctl it dont start magically after that

dwoz:
Maybe just documenting what is needed to add this configuration is a better way to go, rather than include the config by default. My concern is that having the minion restart by default might cause some critical issues to get overlooked.

hedrickbt:
We have run into the same issue.  The service.restart is just too convenient for people not to use it!

dwoz:

PR #55733, opened 2019-12-24
Data filter falsey-master

github-abcde:
This PR ports the changes from https://github.com/saltstack/salt/pull/52499 to master.

### What does this PR do?
This adds a function `filter_falsey` to `salt.utils.data`. The function can be used to filter values from a list or dict that evaluate to false (but are not `bool(False)`). Optionally, a level of recursion can be set to process nested dicts/lists. Also optionally, a list of types to not filter can be passed in `ignore_types`.

None that I know of.

### Previous Behavior
The functionality does not exist.

### New Behavior
We are now able to filter falsey values from dicts or lists.
This is especially handy when having to process a lot of kwargs that are explicitly specified in the function argument list, which default to `None`, but need not be passed on if they are None.
For example:
```
def some_function(foo=None, bar=None, baz=None):
    new_kwargs = {}
    if foo is not None:
        new_kwargs.update('Foo': foo)
    if bar is not None:
        new_kwargs.update('Bar': bar)
    if baz is not None:
        new_kwargs.update('Baz': baz)
    some_other_function(**new_kwargs)
```
will become
```
def some_function(foo=None, bar=None, baz=None):
    new_kwargs = salt.utils.data.filter_falsey({
        'Foo': foo,
        'Bar': bar,
        'Baz': baz,
    })
    some_other_function(**new_kwargs)
```
Which is a typical use-case for most Boto calls.

### Tests written?
Yes

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/utils/data.py
+++ b/salt/utils/data.py
@@ -11,6 +11,7 @@ import copy
 import fnmatch
 import logging
 import re
+import functools
 
 try:
     from collections.abc import Mapping, MutableMapping, Sequence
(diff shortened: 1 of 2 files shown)


dwoz (approved):
@github-abcde Approved, though I would like to see this broken up into two separate methods in the future. One that handles dictionaries and on that handles list like objects. `filter_falsey` could stick around as a convenience method.

github-abcde:
@dwoz Do you have a preference on how those two separate methods should be called?

PR #55745, opened 2019-12-27
[PyTest #33] Final Stretch

s0undt3ch:
### What does this PR do?
See title


.ci/kitchen-centos7-py2-cloud
```diff
@@ -1,14 +1,14 @@
-@Library('salt@master-1.5') _
+@Library('salt@master-1.6') _
 
 runTestSuite(
-    concurrent_builds: 0,
+    concurrent_builds: 0,  // Don't cancel builds still running. Wait until they end.
```
waynew: That means that if a new cloud build rolls in it won't cancel the existing one, right? i.e. if I create a PR it will launch one of these tests, and then I make a change, it will have two tests?

Or are these only run on branch builds?

noxfile.py
```diff
@@ -401,263 +401,6 @@ def _run_with_coverage(session, *test_cmd):
...
-    _install_requirements(session, 'zeromq', 'unittest-xml-reporting==2.2.1')
-    session.install('--progress-bar=off', 'tornado==5.0.2', silent=PIP_INSTALL_SILENT)
-    session.install('--progress-bar=off', 'pyzmq==17.0.0', silent=PIP_INSTALL_SILENT)
-
-    cmd_args = [
-        '--tests-logfile={}'.format(RUNTESTS_LOGFILE)
-    ] + session.posargs
-    _runtests(session, coverage, cmd_args)
-
-
```
waynew: Do I understand correctly that this is completely removing the "runtests" support?

requirements/static/py3.7/windows.txt
```diff
@@ -79,10 +79,11 @@ pymysql==0.9.3
 pyopenssl==19.0.0
 pyparsing==2.4.5          # via packaging
 pytest-helpers-namespace==2019.1.8
-pytest-salt-runtests-bridge==2019.7.10
-pytest-salt==2019.12.18
+pytest-salt-runtests-bridge==2019.12.5
```
waynew: I'm sure there's a reasonable reason that we're using `pytest-salt-runtests-bridge` even though we've removed all the runtests stuff(??). Could we have an explanation for those of us unfamiliar with it?

waynew (changes requested):
Looking great! I think there's only one spot that I'm confident needs a change (the `_logger`). Then just a couple of points of confusion on my part.

tests/integration/modules/test_linux_acl.py
```diff
@@ -5,22 +5,24 @@
...
 
 
 # Acl package should be installed to test linux_acl module
-@skip_if_binaries_missing(['getfacl'])
+@pytest.mark.skip_if_binaries_missing('getfacl')
 # Doesn't work. Why?
-# @requires_salt_modules('acl')
-# @requires_salt_modules('linux_acl')
+# @pytest.mark.requires_salt_modules('acl')
+# @pytest.mark.requires_salt_modules('linux_acl')
```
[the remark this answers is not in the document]
waynew: Though, after reviewing the rest of the code, I think either this is not clearly named and could be `requires_salt_function` - because everything else is like `requires_salt_modules('pkg.list_pkgs')`, which is different than `requires_salt_modules('pkg')` 🤷‍♂ 

I'm not familiar with that helper, though - just observations based on reviewing *this* code 😝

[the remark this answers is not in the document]
waynew: Hm. Looking at that code, I'm not sure *why* this would not work.

If it still is broken, what I would do is just add `if 'acl' in required_salt_modules: raise SkipTest(...)` (same thing for `'linux_acl'`) in that function to make sure *that* works. That would tell us that the skip is correctly working.

Then if that part is working I'd start to look into why the sminion isn't producing the desired information here.

.ci/kitchen-centos7-py2-cloud
s0undt3ch: Yes, we're not canceling previous runs on purpose, and yes, cloud tests only run on branch builds.

noxfile.py
s0undt3ch: You understand correctly.

requirements/static/py3.7/windows.txt
s0undt3ch: It provides test listing from files and test whitelist from a file compatability layer.
Eventually we might drop it.

salt/log/setup.py
```diff
@@ -866,19 +869,39 @@ def __global_logging_exception_handler(exc_type, exc_value, exc_traceback):
...
         )
-        # Call the original sys.excepthook
+    except Exception:  # pylint: disable=broad-except
+        msg = '{}\n{}: {}\n(UNABLE TO FORMAT TRACEBACK)'.format(
+            msg,
+            exc_type.__name__,
+            exc_value,
+        )
+    try:
+        _logger(__name__).error(msg)
```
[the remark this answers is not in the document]
s0undt3ch: Duh!
Yes, you are correct.

cachedout:
This will break the existing method of running tests. 

It should not be merged [without first re-writing the developer documentation.](https://docs.saltstack.com/en/latest/topics/development/tests/index.html#salt-test-suite)

tests/integration/modules/test_linux_acl.py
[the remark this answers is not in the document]
s0undt3ch: This is a `sed`ed change.
I really don't want to address making this test work, or not in this PR.
Perhaps on a follow-up.

s0undt3ch:
Yes, it will and documentation will be a follow-up, and by follow-up, right after.
I really don't think it will be a blocker, that documentation has been out of date for months now, since we switched to nox, and you're the first to mention it(on a ticket).
Doesn't make the lack of documentation better, just to make my point that it isn't a hard blocker.

Anyway, there are still issues to address on this PR, so I'll try to at least update the bit on how to run tests.

.ci/kitchen-macosxhighsierra-py3
```diff
@@ -10,10 +10,10 @@ runTestSuite(
     jenkins_slave_label: 'kitchen-slave-mac',
     kitchen_platforms_file: '/var/jenkins/workspace/pre-golden-nox-platforms.yml',
     kitchen_verifier_file: '/var/jenkins/workspace/nox-verifier.yml',
-    nox_env_name: 'runtests-zeromq',
+    nox_env_name: 'pytest-zeromq',
     nox_passthrough_opts: '',
     python_version: 'py3',
-    run_full: params.runFull,
+    //splits: ['unit', 'integration', 'multimaster'],
```
waynew: Why is this just commented out instead of either live or completely removed? There are several of these lines.

s0undt3ch: Ran into issues when running split at some point in time and just commented it out.
I can remove on the final commit.

s0undt3ch: I'm actually leaving it in since they are not all equal and getting them to this point required some trial and error which I don't want to go through again, for when we split it up again.

Akm0d (approved):

Akm0d:
Is't saying "destructive_test" redundant since pytest markers will be only be applied to tests?  can the marker just be called "destructive"
I.E. `@pytest.mark.destructive`

PR #55750, opened 2019-12-28
Fix the linux_acl module on FreeBSD (and probably Illumos)

asomers:
### What does this PR do?
Fixes all of the commands in the linux_acl module on FreeBSD.  It probably fixes them for Illumos too, but I haven't tested on that platform.

No open issues

### Previous Behavior
`linux_acl.getfacl` would return an error on FreeBSD because `--absolute-names` was unrecognized.  `linux_acl.{wipefacls,modfacl,delfacl}` with the `recursive` flag would probably return an error on Illumos (but I didn't check) because `-R` was unrecognized.

### New Behavior
`linux_acl.getfacl` now works on FreeBSD, as long as `recursive` isn't specified.  If it is, a `CommandExecutionError` will be raised.  The other commands, when used with the `recursive` flag on Illumos, will now also raise a `CommandExecutionError`

### Tests written?
No.  The existing tests in tests/unit/modules/test_linux_acl.py and tests/integration/modules/test_linu_acl.py provide adequate coverage.  I modified them so that all will pass on FreeBSD.

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/modules/linux_acl.py
+++ b/salt/modules/linux_acl.py
@@ -40,6 +40,30 @@ def version():
     return ret[1].strip()
 
 
+def _getfacl_dash_r(kwargs):
+    """ Append the -R flag if requested and supported """
+    recursive = kwargs.pop('recursive', False)
+    if recursive:
+        if salt.utils.platform.is_linux():
+            return ' -R'
+        else:
+            raise CommandExecutionError('recursive getfacl not available')
+    else:
+        return ''
+
+
+def _setfacl_dash_r(kwargs):
+    """ Append the -R flag if requested and supported """
+    recursive = kwargs.pop('recursive', False)
+    if recursive:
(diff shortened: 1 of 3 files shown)


dwoz (changes requested):
@asomers please look into the linter failure

asomers:
I haven't forgotten @dwoz;I've just been on vacation.

asomers:
Lint error fixed.  Also, rebased onto master.

asomers:
The test failures are timeouts.  They are unrelated to my changes.

sjorge:
Do you have an example state? If so I could probably pull in the patch and run the state on a SmartOS zone.

asomers:
For example state just look at the integration tests, @sjorge .  I don't have anything beyond that.

sjorge:
I guess from looking at the module, it makes sense. `dentry` is set to `''`

dwoz:

PR #55751, opened 2019-12-28
Set the osfullname grain on FreeBSD

asomers:
### What does this PR do?
Sets the osfullname to "FreeBSD" on FreeBSD

none

### Previous Behavior
`grains.get osfullname` would return nothing

### New Behavior
`grains.get osfullname` will return "FreeBSD"

### Tests written?
No.  The existing TestModulesGrains.test_get_core_grains test will now pass on FreeBSD

### Commits signed with GPG?
Yes


s0undt3ch (changes requested):
Thank You for your contribution!

Could you please add a small unittest around this change?

asomers:
@s0undt3ch are we good here?  The test failure looks unrelated.

s0undt3ch (approved):

PR #55752, opened 2019-12-28
Misc fixes to the localemod module

asomers:
### What does this PR do?
1) Fixes the error message for locale.get_locale on unsupported platforms.  Previously it would raise a KeyError
2) Skips the localemod tests on FreeBSD, where that module is unsupported
3) Fixes some localemod tests whose pass conditions were too generous.

none

### Previous Behavior
Calling localemod.get_locale on an unsupported platform would raise an unhelpful KeyError

### New Behavior
Calling localemod.get_locale on an unsupported platform would raise a CommandExecutionError, as it should

### Tests written?
No.  The existing localemod tests are now skipped on FreeBSD.  If any new errors pop up in CI, it's probably because I fixed test_gen_locale and test_set_locale.  Previously, those tests weren't actually asserting anything

### Commits signed with GPG?
Yes

### DIFF
--- a/tests/integration/modules/test_localemod.py
+++ b/tests/integration/modules/test_localemod.py
@@ -45,9 +45,9 @@ class LocaleModuleTest(ModuleCase):
             )
 
         locale = self.run_function('locale.get_locale')
-        self.assertEqual(ret['retcode'], 0)
         new_locale = _find_new_locale(locale)
         ret = self.run_function('locale.gen_locale', [new_locale])
+        self.assertEqual(ret['retcode'], 0)
         self.assertNotIn('ERROR:', locale)
 
     @destructiveTest
--- a/tests/integration/modules/test_localemod.py
+++ b/tests/integration/modules/test_localemod.py
@@ -26,12 +26,7 @@ def _find_new_locale(current_locale):
 @skipIf(salt.utils.platform.is_freebsd(), 'locale module is not supported on FreeBSD')
 @requires_salt_modules('locale')
 class LocaleModuleTest(ModuleCase):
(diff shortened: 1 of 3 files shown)


dwoz:
@asomers looks like the test failures are related to the change.

asomers:
Yeah.  I'll set up a Linux development environment so I can debug those.

tests/integration/modules/test_localemod.py
```diff
@@ -46,7 +47,7 @@ def test_gen_locale(self):
         locale = self.run_function('locale.get_locale')
         new_locale = _find_new_locale(locale)
         ret = self.run_function('locale.gen_locale', [new_locale])
-        self.assertTrue(ret)
```
s0undt3ch: We should still assert against `ret`, we're testing against `gen_locale`.
Probably `get_locale` again using the new locale, and then assert against the error not being present.

s0undt3ch (changes requested):

asomers:
The failure on ci/py3/centos7/proxy was due to bundler, not my changes.

tests/integration/modules/test_localemod.py
asomers: asserting against `ret` is useless.  It always succeeds, even if the command failed.

s0undt3ch: Hmm, that suggests a bad implementation.
Not saying it should be addressed in this PR, at least for now.

asomers:
I believe I fixed the failures on CentOS and Fedora

dwoz:

PR #54100, opened 2019-08-02
Mine minion acl

github-abcde:
### What does this PR do?
It adds a new format for defining mine_functions that is equal to the format used for `module.run`.
It adds deprecation warnings for the old style format until version `Aluminium`.
It adds minion-side ACL when sending items to the mine, or updating the mine, which uses `allow_tgt` and `allow_tgt_type` to determine the list of minions that is allowed to retrieve this function from the salt mine. Minions that do not have access get an empty result, just as if the requested function was not present.
Also a slight refactor has been performed, cleaning up code in `salt.modules.mine` significantly.

### What issues does this PR fix or reference?
I have mentioned the minion-side ACL as a possible solution for https://github.com/saltstack/salt/issues/45882. Whether or not this will resolve that specific issue is up for debate though :)
https://github.com/saltstack/salt/issues/6437

### Previous Behavior
All mine functions (and their data) are retrievable by all minions.

### New Behavior
Mine functions (and their data) can be made available to targeted minions only.

### Tests written?
Yes, plenty. There were none for the salt mine in `salt.modules.mine` to begin with.

### Commits signed with GPG?
Yes

Issue #45882, opened 2018-02-06
Canonical solution for securely transferring data between minions


Nick2253:
## Description of Question
I've been trying to determine the canonical solution to the following problem:
- Generate secure data on a first minion.
- Pass the secured data from the first minion to a second minion.

An example of this process is generating a Let's Encrypt certificate from a web-accessible client (first minion) for a web-inaccessible client (second minion), and passing the certificate and key to the second minion.  This particular example will become even more important once Let's Encrypt allows wildcard certs at the end of February.

Another example of this process is generating an Icinga2 client ticket.  When a new minion (the second minion) is added to the network, the Icinga2 master (the first minion) generates a ticket, and the second minion needs that ticket to complete client setup.

A further example of this process is a username/password.  Many legacy services are only accessible via usernames/passwords, and they have requirements for regular password changes (usually for regulatory reasons).  These services can be configured to automatically generate a new password (on the first minion), and these passwords need to be send to clients that need them (the second minion).

### Possible Solutions, and Problems
I have seen a few possible solutions discussed on the internet, but no solution seems to provide all these features:
- Handled entirely by Salt
- Secure the data so that only the master and the two minions know about it

Generally speaking, the first part of the problem is easily done in Salt.  Whether with a custom method, a command, etc, it's fairly easy to either create a file, or even a pillar file, containing the secure data on the first minion.  The challenge is the second part:  moving the file securely using Salt.

I'm assuming that the best way to handle this is transfer the file from the first minion to the master, and then from the master to the second minion.  This assumption is also shared with some of the solutions I've seen below.

#### `cp.push`
*As suggested in #31863*

The general workflow is to use [`cp.push`](https://docs.saltstack.com/en/latest/ref/modules/all/salt.modules.cp.html#salt.modules.cp.push) to send the file from the first minion to the master.  However, `cp.push` uses `salt.transport.Channel.send()`, which is [not guaranteed to be confidential](https://docs.saltstack.com/en/develop/topics/transports/index.html#req-channel).

If we could add a "secure" flag to the `cp.push` method to use the `crypted_transfer_decode_dictentry`, that would enable a means to securely pass the files to the master.

#### `event.send`
*As suggested in #27156*

The general workflow is to generate the secure data on the first minion, and then trigger an event, passing the secure data from the first minion to the master.  A reactor is triggered on the event, and applies the data to the second minion.

However, this method has the same problems as `cp.push`.  All data passed in events, including pillar data, uses the `.send()` transport.

[...]

Issue #6437, opened 2013-07-30
[Feature]: Salt-mine access restrictions / permission granularity


ptonelli:
## Problem

 Salt-mine (http://docs.saltstack.com/topics/mine/index.html) is a new great tool in Salt. However, the acces to salt-mine authorization is all or nothing. A minion either has access to the mine or has no access.
## Proposition

 It would be really useful to have a permission mechanism which would:
- limit access for a given minion only to certains grains from other minion

and / or
-  permissions to  access only specific minions and not to all the infrastructure.

paul-mulvihill:
+1 for this

lubyou:
+1 for this feature

pierluca:
+1

The-Loeki:
https://docs.saltstack.com/en/latest/ref/configuration/examples.html

```
#####         Mine settings     #####
#####################################
# Restrict mine.get access from minions. By default any minion has a full access
# to get all mine data from master cache. In acl definion below, only pcre matches
# are allowed.
# mine_get:
#   .*:
#     - .*
#
# The example below enables minion foo.example.com to get 'network.interfaces' mine
# data only, minions web* to get all network.* and disk.* mine data and all other
# minions won't get any mine data.
# mine_get:
#   foo.example.com:
#     - network.interfaces
#   web.*:
#     - network.*
#     - disk.*
```

paul-mulvihill:
That does cover the first half of the request. For me personally the 2nd half of the request would be the most useful. The ability to limit the infrastructure it pulls from.
I have a different departments in my estate and a single network of salt masters.. being able to forcibly limit the results of mine queries to the minions department would be very useful (maybe via department level grain/pillar value).

The-Loeki:
Well the docs are pretty limited, but the following oughta kinda work:

```yaml
mine_get:
  mine-getters*.example.com:
    - mine-pusher*.example.com:
      - network.interfaces
```

paul-mulvihill:
Would that enable the restriction on results?
Department A, Minion S1 S2 S4 S6 S8
Department B, Minion S3 S7 S9 S10

If they two servers (S1, S3) query for all minion IP's. I want S1 to only return other minions in department A (S1, S2, S4, S6, S8) and S3 to return only minions in department B (S3, S7, S9, S10). 
I'm not sure it currently allows that level of restrictions but I've not looked at mines for a long time after I saw the original lack of limitation.
Unfortunately I can't have minions from other departments appearing in a minions results.
[...]

PR #55760, opened 2019-12-30
Mine minion acl master

github-abcde:
Master port of https://github.com/saltstack/salt/pull/54100

### What does this PR do?
It adds a new format for defining mine_functions that is equal to the format used for `module.run`.
It adds deprecation warnings for the old style format until version `Aluminium`.
It adds minion-side ACL when sending items to the mine, or updating the mine, which uses `allow_tgt` and `allow_tgt_type` to determine the list of minions that is allowed to retrieve this function from the salt mine. Minions that do not have access get an empty result, just as if the requested function was not present.
Also a slight refactor has been performed, cleaning up code in `salt.modules.mine` significantly.

I have mentioned the minion-side ACL as a possible solution for https://github.com/saltstack/salt/issues/45882. Whether or not this will resolve that specific issue is up for debate though :)
https://github.com/saltstack/salt/issues/6437

### Previous Behavior
All mine functions (and their data) are retrievable by all minions.

### New Behavior
Mine functions (and their data) can be made available to targeted minions only.

### Tests written?
Yes, plenty. There were none for the salt mine in `salt.modules.mine` to begin with.

### Commits signed with GPG?
Yes


dwoz (approved):

salt/utils/mine.py
```diff
@@ -0,0 +1,128 @@
...
+
+    :rtype: tuple
+    :return: Tuple with function_name, function_args, function_kwargs, minion_acl (dict)
+    '''
+    function_name = None
+    function_args = []
+    function_kwargs = {}
+    minion_acl = {}
+    if isinstance(function_definition, dict):
+        salt.utils.versions.warn_until(
```
s0undt3ch: Salt has taken a strict approach when changes might eventually break people's states, which was the case with the new `module.run` syntax and also this case.

It's ok to support both old and new syntax.
It's not ok to deprecate the old syntax.

Please remove any deprecation information about the old syntax, it won't be dropped, and update(if needed) the code to support both syntaxes.

s0undt3ch (changes requested):

salt/utils/mine.py
github-abcde: Removed deprecation warnings in 4f4c562e612230874bb61ffd762977378dd174ba

doc/topics/mine/index.rst
```diff
@@ -42,13 +42,57 @@ the example below:
...
 
 In the example above :py:mod:`salt.modules.network.ip_addrs` has additional
 filters to help narrow down the results.  In the above example IP addresses
 are only returned if they are on a eth0 interface and in the 10.0.0.0/8 IP
 range.
 
+.. versionchanged:: Sodium
+
+The format to define mine_functions has been changed to allow the same format
+as used for module.run. The old format (above) will still be supported until
```
s0undt3ch: The old format will still be supported. Please remove the until.... bit.

doc/topics/releases/sodium.rst
```diff
@@ -0,0 +1,28 @@
...
+====================================
+
+
+Salt mine updates
+=================
+
+Syntax update
+-------------
+
+The syntax for defining salt functions in config or pillar files has changed to
```
s0undt3ch: Also needs updating

s0undt3ch (changes requested):

doc/topics/releases/sodium.rst
github-abcde: Updated in bfee1c8d9bcab7685bbc95f5d05a493db52560e9

doc/topics/mine/index.rst
github-abcde: Updated in bfee1c8d9bcab7685bbc95f5d05a493db52560e9

s0undt3ch (approved):
Thank You!

dwoz:
@github-abcde Can you take a look at the merge conflicts please?

github-abcde:
Ok, looks like I forgot to add a few pylint-disable statements. Will do that tomorrow :)

github-abcde:
re-run pr-amazon1-py2

mchugh19:
@dwoz and @github-abcde looks like the versionadded and release notes reference sodium, but it looks like this is in Neon. Should that be updated?

OrangeDog:
Turns out this was a breaking change, as the arg `func` was renamed `name`.

See the examples here, that use `func` as a kwarg.
https://docs.saltstack.com/en/3000/ref/states/all/salt.states.x509.html

PR #55772, opened 2019-12-31
salt.renderers.gpg: Cache GPG data if gpg_cache=True

sathieu:
### What does this PR do?
- Cache gpg data if requested

NB : `pillar_cache ` doesn't work reliably as it can return obsolete data.

### Previous Behavior

Doing `salt  <matcher> pillar.items` matching 60 minions with several pillars encrypted, we got timeout error after 45s.

### New Behavior

The same command returns within 2 to 4 seconds.

### Tests written?

Yes

### Commits signed with GPG?

No

### DIFF
--- a/conf/master
+++ b/conf/master
@@ -1024,6 +1024,24 @@
 #
 #pillar_cache_backend: disk
 
+# A master can also cache GPG data locally to bypass the expense of having to render them
+# for each minion on every request. This feature should only be enabled in cases
+# where pillar rendering time is known to be unsatisfactory and any attendant security
+# concerns about storing decrypted GPG data in a master cache have been addressed.
+#
+# When enabling this feature, be certain to read through the additional ``gpg_cache_*``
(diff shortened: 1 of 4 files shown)


dwoz (changes requested):
@sathieu It looks like the test failures are related to this change, please take a look.

sathieu:
@dwoz Yes, this `AttributeError: 'module' object has no attribute 'errors'` error is strange because I don't have it, and this is [per doc](http://files.au.adversary.org/crypto/gpgme-python-howto.html#Decryption).

sathieu:
@dwoz I've removed the _Use Python bindings when available_, the most usefull feature being `gpg_cache`.

sathieu:
@dwoz Please review again, the failing tests are not related to this PR.

sathieu:
@dwoz @Akm0d Please review, now al checks have passed :v: .

Akm0d (approved):

Issue #54809, opened 2019-09-30
Salt Minion constantly restarting when using AD user


kriss0011:
We are having a windows server (2019 Standart) in remote location. Therefore we are using a different domain that we haven't use for Salt until now. 
After installing the Salt minion, the minion registers in the master and it works just fine. When we change the Salt-Minion Service user from "Local System" to domain user that is local admin the servicestart to restart constantly and it has no longer a connection to master.
This is the error we get in the Minion log:

> [salt.log.setup   :1217][ERROR   ][256748] An un-handled exception was caught by salt's global exception handler:
error: (1722, 'NetUserGetLocalGroups', 'The RPC server is unavailable.')
Traceback (most recent call last):
  File "c:\salt\bin\Scripts\salt-minion", line 26, in <module>
    salt_minion()
  File "c:\salt\bin\lib\site-packages\salt\scripts.py", line 191, in salt_minion
    minion.start()
  File "c:\salt\bin\lib\site-packages\salt\cli\daemons.py", line 343, in start
    super(Minion, self).start()
  File "c:\salt\bin\lib\site-packages\salt\utils\parsers.py", line 1062, in start
    self.prepare()
  File "c:\salt\bin\lib\site-packages\salt\cli\daemons.py", line 291, in prepare
    pki_dir=self.config['pki_dir'],
  File "c:\salt\bin\lib\site-packages\salt\utils\verify.py", line 215, in verify_env
    skip_extra=skip_extra)
  File "c:\salt\bin\lib\site-packages\salt\utils\verify.py", line 572, in win_verify_env
    if salt.utils.win_functions.is_admin(current_user):
  File "c:\salt\bin\lib\site-packages\salt\utils\win_functions.py", line 62, in is_admin
    groups = get_user_groups(name, True)
  File "c:\salt\bin\lib\site-packages\salt\utils\win_functions.py", line 87, in get_user_groups
    groups = win32net.NetUserGetLocalGroups(None, name)
pywintypes.error: (1722, 'NetUserGetLocalGroups', 'The RPC server is unavailable.')

This error doesn't appear until we change the service user.

### Steps to Reproduce Issue
(Include debug logs if possible and relevant.)

The master is:

Dependency Versions:
           cffi: Not Installed
       cherrypy: 3.5.0
       dateutil: 2.4.2
      docker-py: Not Installed
          gitdb: 0.6.4
      gitpython: 1.0.1
          ioflo: Not Installed
         Jinja2: 2.8
[...]

PR #55780, opened 2020-01-03
Fallback to disabled LG_INCLUDE_INDIRECT when DC is unavailable

lukasraska:
### What does this PR do?
Fixes issue when DC is unavailable and Salt Minion tries to retrieve list of groups the running user is member of. 

#54809

### Previous Behavior
Minion also tries to resolve indirect groups, which causes minion to fail with exception which causes the service to restart indefinitely.

### New Behavior
Minion tries to resolve only direct groups if indirect resolving fails, causing service to start properly if there is any issue with DC

### Tests written?
Yes

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/utils/win_functions.py
+++ b/salt/utils/win_functions.py
@@ -88,9 +88,10 @@ def get_user_groups(name, sid=False):
         try:
             groups = win32net.NetUserGetLocalGroups(None, name)
         except win32net.error as exc:
-            if exc.winerror == 5:
+            # ERROR_ACCESS_DENIED, NERR_DCNotFound, RPC_S_SERVER_UNAVAILABLE
+            if exc.winerror in (5, 1722, 2453):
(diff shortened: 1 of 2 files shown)


lukasraska:
I've used ERROR_ACCESS_DENIED, NERR_DCNotFound and RPC_S_SERVER_UNAVAILABLE (taken from https://docs.microsoft.com/en-us/windows/win32/api/lmaccess/nf-lmaccess-netusergetlocalgroups) to check the exception, as other don't make much sense for checking.

While NERR_DCNotFound probably won't occur, might be also good to check, so I've included it apart from what was mentioned in #54809

lukasraska:
re-run pr-windows-2019-py3
re-run pr-archits-py3

lukasraska:
re-run pr-windows2019-py3

dwoz (approved):

PR #55793, opened 2020-01-06
`__getstate__` and `__setstate__` consolidation

s0undt3ch:
### What does this PR do?
The `Process` class will now take care of all `__getstate__` and `__setstate__`

This PR will then ensure that any forked process, on windows, and of course all platforms which support forking, will behave properly **without** having to implement their own get and set state functions, unless they need to extend the base functionality provided in out Process class.

### Previous Behavior
All subclasses had to implement `__getstate__` and `__setstate__`

### New Behavior
The `Process` class will now take care of all `__getstate__` and `__setstate__`

### Tests written?

No

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/engines/__init__.py
+++ b/salt/engines/__init__.py
@@ -2,7 +2,6 @@
 Initialize the engines system. This plugin system allows for
 complex services to be encapsulated within the salt plugin environment
 """
-
 import logging
 
 import salt
@@ -82,35 +81,6 @@ class Engine(salt.utils.process.SignalHandlingProcess):
         self.funcs = funcs
         self.runners = runners
         self.proxy = proxy
-
-    # __setstate__ and __getstate__ are only used on Windows.
-    # We do this so that __init__ will be invoked on Windows in the child
-    # process so that a register_after_fork() equivalent will work on Windows.
-    def __setstate__(self, state):
-        self.__init__(
-            state["name"],
-            state["opts"],
-            state["fun"],
-            state["config"],
-            state["funcs"],
-            state["runners"],
-            state["proxy"],
-            log_queue=state["log_queue"],
-            log_queue_level=state["log_queue_level"],
-        )
-
-    def __getstate__(self):
-        return {
-            "name": self.name,
-            "opts": self.opts,
-            "fun": self.fun,
-            "config": self.config,
-            "funcs": self.funcs,
-            "runners": self.runners,
-            "proxy": self.proxy,
-            "log_queue": self.log_queue,
-            "log_queue_level": self.log_queue_level,
-        }
 
     def run(self):
         """
--- a/salt/utils/platform.py
+++ b/salt/utils/platform.py
@@ -2,6 +2,7 @@
 Functions for identifying which platform a machine is
 """
 
+import multiprocessing
 import os
 import platform
 import subprocess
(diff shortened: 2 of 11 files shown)


salt/engines/__init__.py
```diff
@@ -81,33 +81,6 @@ def __init__(self, opts, fun, config, funcs, runners, proxy, **kwargs):
...
-            'opts': self.opts,
-            'fun': self.fun,
-            'config': self.config,
-            'funcs': self.funcs,
-            'runners': self.runners,
-            'proxy': self.proxy,
-            'log_queue': self.log_queue,
-            'log_queue_level': self.log_queue_level
-        }
-
```
waynew: Not sure that I understand - how do all of these things end out getting called? They're quite different than what's in the other files.

s0undt3ch: On windows, and windows alone, the `__new__` method will keep a copy of the `*args` and `**kwargs` used to instantiate the class.
Then, again, only on windows, the get and set state methods will use that information to re-instsntiate the class.

s0undt3ch: [Here](https://github.com/saltstack/salt/pull/55793/files#diff-0eb908cc193e4bf8dd9f96567a3abfebR741)

s0undt3ch:
This PR will then ensure that any forked process, on windows, and of course all platforms which support forking, will behave properly **without** having to implement their own get and set state functions.

s0undt3ch:
Unless they need to extend the base functionality provided in out Process class.

twangboy (dismissed):

s0undt3ch:
I need to make sure any extra methods for post fork and pre shutdown are also carried over to the new process.

s0undt3ch:
re-run full all

dwoz (commented):
I'm not a fan of this. It seems reasonable to me to expect sub-classes of Process to implement their own `__getstate__` and `__setstate__` rather than trying to be smart about it in the constructor.

s0undt3ch:
As we discussed, it gets us, at least, logging on those processes for free.

twangboy (dismissed):

s0undt3ch:
Conflicts resolved

dwoz (dismissed):

krionbsd (dismissed):

s0undt3ch:
The pre-commit issue is being solved in https://github.com/saltstack/salt/pull/60254

s0undt3ch:
re-run full all

s0undt3ch:
re-run full all

s0undt3ch:
re-run full all

s0undt3ch:
re-run full all

s0undt3ch:
re-run full all

s0undt3ch:
re-run full all

s0undt3ch:
re-run full all

s0undt3ch:
re-run full all

s0undt3ch:
re-run full all

s0undt3ch:
re-run full all

salt/utils/platform.py
```diff
@@ -209,3 +210,12 @@ def is_aarch64():
     Simple function to return if host is AArch64 or not
     """
     return platform.machine().startswith("aarch64")
+
+
+def spawning_platform():
```
waynew, line 215: **question** is it better to gate this behind a function or do something like this at the module level?

```
SPAWNING_PLATFORM = multiprocessing.get_start_method(allow_none=False) == "spawn"
```

I don't know if it's called enough to make a difference :shrug:

s0undt3ch: No, the method can be changed programmatically at run time. A global variable would only pick the initial config value.

s0undt3ch: @waynew are we good to resolve this review comment?

twangboy (approved):

salt/utils/platform.py
s0undt3ch: Also, if you notice, this is probably the only function which doesnt have the cached decorator, that was the reason behind it.

dwoz (approved):

Issue #54908, opened 2019-10-06
Type issue in salt.modules.dpkg_lowpkg._get_pkg_info() that results in pkg.info_installed not working against Ubuntu


arizvisa:
### Description of Issue
The `_get_pkg_info` function from `salt.modules.dpkg_lowpkg` is comparing the `osrelease_info` grain against a tuple in order to determine the correct keywords to use when querying dpkg. The issue is that the `osrelease_info` grain returns a list (or at least it does on my systems) and is being compared against a tuple. This is shown in the following snippet from `salt.modules.dpkg_lowpkg`:

This is grabbed from near
https://github.com/saltstack/salt/blob/develop/salt/modules/dpkg_lowpkg.py#L301
```
def _get_pkg_info(*packages, **kwargs):
    '''
    Return list of package information. If 'packages' parameter is empty,
    then data about all installed packages will be returned.

    :param packages: Specified packages.
    :param failhard: Throw an exception if no packages found.
    :return:
    '''
...
    if __grains__['os'] == 'Ubuntu' and __grains__['osrelease_info'] < (12, 4):
        bin_var = '${binary}'
    else:
        bin_var = '${Package}'
```

On Python2, comparing a list against a tuple with `<` will always return `True`. On Python3, this will raise an exception. On ubuntu releases >= 12.4 (assuming based on the code), dpkg uses `${Package}` to insert the package name, whereas on earlier versions `${binary}` represents the package name. Due to the success of this check, the `else` case will never get executed on Python2 if the `osrelease_info` grain returns a list.

Some callers of `_get_pkg_info` depend on the "package" key being available in the dictionary that is returned. Due to this test always succeeding, this results in the "package" key potentially being non-existent on releases later than 12.4.

### Setup
Install Ubuntu newer than 12.4 and try salt-call

### Steps to Reproduce Issue
```
$ salt --async $id pkg.info_installed
Executed command with job ID: 20191006202640409841
$ salt-run jobs.lookup_jid 20191006202640409841
hyEowqOAC7PQnONCkaMI:
    The minion function caused an exception: Traceback (most recent call last):
      File "/usr/lib/python2.7/dist-packages/salt/minion.py", line 1663, in _thread_return
        return_data = minion_instance.executors[fname](opts, data, func, args, kwargs)
      File "/usr/lib/python2.7/dist-packages/salt/executors/direct_call.py", line 12, in execute
        return func(*args, **kwargs)
[...]

Issue #54882, opened 2019-10-04
When using an empty list in the top file "Received bad data when setting the match..."


arizvisa:
### Description of Issue
When using an empty list as a placeholder in the topfile, the following error is displayed:
```
2019-10-03 19:44:24,865 [/usr/lib/python2.7/site-packages/salt/matchers/confirm_top.pyc:22  ][ERROR   ][176] Received bad data when setting the match from the top file
```
This happens because the following definition of `salt.matchers.confirm_top` which assumes that an implicit bool means the data is bad. This should probably be checking the type as an empty list is a valid structure to come from yaml. This is a matter of opinion however.

```
def confirm_top(match, data, nodegroups=None):
    '''
    Takes the data passed to a top file environment and determines if the
    data matches this minion
    '''
    matcher = 'compound'
    if not data:
        log.error('Received bad data when setting the match from the top '
                  'file')
        return False
```

### Setup
Define a topfile that looks like the following:
```
base:
  '*':
   []
```

### Steps to Reproduce Issue
You could run anything that processes the topfile, but since it's related you can try `state.show_top`.

### Versions Report
```
Salt Version:
           Salt: 2019.2.1

Dependency Versions:
           cffi: 1.11.5
       cherrypy: Not Installed
       dateutil: 2.8.0
      docker-py: Not Installed
          gitdb: 2.0.3
      gitpython: 2.1.14
          ioflo: 1.7.6
         Jinja2: 2.10.1
        libgit2: 0.27.8
        libnacl: 1.6.1
       M2Crypto: Not Installed
           Mako: 1.1.0
   msgpack-pure: Not Installed
 msgpack-python: 0.6.1
   mysql-python: Not Installed
      pycparser: 2.14
       pycrypto: 2.6.1
   pycryptodome: 3.9.0
         pygit2: 0.27.4
         Python: 2.7.16 (default, Apr 30 2019, 15:54:43)
   python-gnupg: Not Installed
         PyYAML: 5.1
          PyZMQ: 17.0.0
           RAET: 0.6.8
          smmap: 2.0.3
        timelib: Not Installed
        Tornado: 5.0.2
            ZMQ: 4.3.2

System Versions:
           dist: fedora 30 Thirty
         locale: ANSI_X3.4-1968
        machine: x86_64
        release: 4.14.96-coreos
         system: Linux
        version: Fedora 30 Thirty
```

arizvisa:
Closed by #54883.

arizvisa:
[...]

PR #55796, opened 2020-01-06
Fix cached osrelease_info grain type

srg91:
### What does this PR do?
This pull-request fixes problems with operating on `osrelease_info` grain if it was loaded from cache. Now it loads correctly as `tuple`.

I found the one: #54908

### Previous Behavior
`loader.py` loads grain `osrelease_info` as `list`.
It broke modules like `aptpkg.info_installed` and similar.

### New Behavior
Now `osrelease_info` loads as tuple 

### Tests written?
Yes, `unit.test_loader.LoaderLoadCachedGrainsTest`

### Commits signed with GPG?
No

### DIFF
--- a/salt/loader.py
+++ b/salt/loader.py
@@ -688,6 +688,17 @@ def grain_funcs(opts, proxy=None):
     return ret
 
 
+def _format_cached_grains(cached_grains):
+    """
+    Returns cached grains with fixed types, like tuples.
+    """
+    if cached_grains.get('osrelease_info'):
+        osrelease_info = cached_grains['osrelease_info']
+        if isinstance(osrelease_info, list):
+            cached_grains['osrelease_info'] = tuple(osrelease_info)
+    return cached_grains
+
+
 def _load_cached_grains(opts, cfn):
     '''
     Returns the grains cached in cfn, or None if the cache is too old or is
@@ -720,7 +731,7 @@ def _load_cached_grains(opts, cfn):
             log.debug('Cached grains are empty, cache might be corrupted. Refreshing.')
             return None
 
-        return cached_grains
+        return _format_cached_grains(cached_grains)
     except (IOError, OSError):
         return None
 
--- a/tests/unit/test_loader.py
+++ b/tests/unit/test_loader.py
@@ -1353,3 +1353,31 @@ class LazyLoaderOptimizationOrderTest(TestCase):
         basename = os.path.basename(filename)
         expected = 'lazyloadertest.py' if six.PY3 else 'lazyloadertest.pyc'
         assert basename == expected, basename
+
+
+class LoaderLoadCachedGrainsTest(TestCase):
+    '''
+    Test how the loader works with cached grains
+    '''
+
+    @classmethod
+    def setUpClass(cls):
+        cls.opts = salt.config.minion_config(None)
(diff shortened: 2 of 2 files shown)


rares-pop:
This seems to fix an effect, but not the root cause of the problem. 

Also, if it's a general mechanism that 'flattens' the tuples, are there other cases that needs to be fixed?

srg91:
> This seems to fix an effect, but not the root cause of the problem.
I think it is, because we have the common func which process loaded cache. We can fix it in another way, but it seems to be more complex, because (if I understand problem correctly) it will require to change the store cache engine (from msgpack to something else).

> Also, if it's a general mechanism that 'flattens' the tuples, are there other cases that needs to be fixed?
I looked through `grains/core.py` and some other modules before and found the only that case.

arizvisa:
Although it might not be the root cause of the problem, when performing an implicit operation that can have different semantics between sets, lists, or tuples, it's much safer to impose the type at the time of use rather than let the type pass through a callstack without knowing what else that type might taint.

Nonetheless, I think it's a maybe good idea to assert the type during development, and impose the type w/ a warning during production to help reduce and track down these sorts of issues. That mis-typing symptom occurs in a few of my other bug reports (#54882 come to mind).

Just my 2cents, and thanks for tracking this down to a caching issue. I literally thought I was crazy when I couldn't repro.

dwoz:
@srg91 Thanks for the contribution. This change needs to be made against `master`.

srg91:
> @srg91 Thanks for the contribution. This change needs to be made against `master`.

Thanks for noticing, I just followed the contribution docs. Could you tell, please, the best way to do this? Is it will be enough if I just force replace my branch and change this pull request or I should create the new one?

srg91:
I've replaced commit with force push and switched destination branch to master, but forgot to update master before, sorry. I've updated branch to master with merge, but maybe I should do this with rebase?

arizvisa:
@srg91, not sure if you'll get a response from the maintainers as I asked questions similar to yours before and didn't get an answer to my questin for like a year. But yeah, once you change the branch compare for your PR to master, the commits are a lot cleaner w/ a rebase.

For now though, just wait for the tests to complete before doing the rebase and I think you should be good.

srg91:
@arizvisa thanks for your comment. All checks are done instead one - `ci/py2/centos7` (https://jenkinsci.saltstack.com/job/pr-centos7-py2/job/PR-55796/3/console). I had no abilities to parse exception and decided to just do rebase :) Maybe restart will help, if no I will try check it later.

dwoz:
@srg91 @arizvisa Yes, it's been a fairly painful transition to `master`. The updates to the contributor docs have been committed and will be released with `neon`.

dwoz (approved):

arizvisa:
@dwoz, if you guys make tags for prioritizing the PRs your community can likely help with some of the small or submitted ones. Maybe something simple like "PR low-priority", "PR high-priority", "PR medium-priority", etc.

Like I didn't know that you guys were going through the huge process of phasing out the `development` branch entirely and I can at least help with some of the PRs that me or others have submitted so it's less work on you guys.

But again, I hope I'm not speaking out-of-turn for others, but please let us know if you guys need help with porting one of our PRs. Silence makes us contributors feel hopeless and also whiny.

PR #55801, opened 2020-01-07
Reduce code duplication(complication)

s0undt3ch:
### What does this PR do?
See title

### DIFF
--- a/tests/support/case.py
+++ b/tests/support/case.py
@@ -33,7 +33,6 @@ from tests.support.mixins import (
     SaltClientTestCaseMixin,
     SaltMultimasterClientTestCaseMixin,
 )
-from tests.support.paths import CODE_DIR, INTEGRATION_TEST_DIR, PYEXEC, SCRIPT_DIR
 from tests.support.processes import terminate_process
 from tests.support.runtests import RUNTIME_VARS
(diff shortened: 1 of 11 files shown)


cmcmarrow (dismissed):

tests/support/case.py
```diff
@@ -46,18 +45,27 @@
...
 
 
-class ShellTestCase(TestCase, AdaptedConfigurationTestCaseMixin, ScriptPathMixin):
+class ShellCase(TestCase, AdaptedConfigurationTestCaseMixin, ScriptPathMixin):
     """
     Execute a test for a shell command
     """
 
-    def run_salt(self, arg_str, with_retcode=False, catch_stderr=False, timeout=15):
+    RUN_TIMEOUT = 30
```
dwoz, line 53: Looks like we are changing timeouts from 60 to 30. Most of these long timeouts were added to address flakiness in the windows test suite.

s0undt3ch: Even before we started skipping slow tests, these passed on windows. This is just a rebase againsr latest changes.
If we do get some failures on windows, we can just increase those and for windows alone giving us yet another chance to identify slow tests on windows.

dwoz (approved):

PR #55807, opened 2020-01-07
Mac group tests

Akm0d:
### What does this PR do?
Refactors the PAM auth tests and skips one that doesn't run properly with the threadsafe loader.  
separated from https://github.com/saltstack/salt/pull/54878

### DIFF
--- a/tests/integration/shell/test_auth.py
+++ b/tests/integration/shell/test_auth.py
@@ -13,11 +13,18 @@ try:
 except ImportError:
     pwd, grp = None, None
 import random
+import string
 
 # Import Salt Testing libs
-from tests.support.case import ShellCase
 from tests.support.unit import skipIf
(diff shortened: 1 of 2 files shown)


waynew (approved):

garethgreenaway (approved):

dwoz:
@Akm0d I don't like the idea of skipping a test because we don't understand why it's failing. Let's take another look at this. Should we close this one while we are investigating more?

Akm0d:
> @Akm0d I don't like the idea of skipping a test because we don't understand why it's failing. Let's take another look at this. Should we close this one while we are investigating more?

It shouldn't be closed yet, there may be a bug that the thread safe loader PR doesn't create, but exposes.  I'm looking deeper and getting close, removed the test skipping.

Akm0d:
re-run full windows2016

dwoz (approved):

PR #55817, opened 2020-01-08
Bring #51372 to Master branch

twangboy:
### What does this PR do?
Brings #51372 to master

### Tests written?
Yes

### Commits signed with GPG?
Yes


salt/utils/network.py
```diff
@@ -997,46 +996,7 @@ def win_interfaces():
...
-                                item['broadcast'] = broadcast
-                        if iface.IPSubnet:
-                            netmask = next((i for i in iface.IPSubnet if ':' in i), '')
-                            if netmask:
-                                item['netmask'] = netmask
-                        ifaces[iface.Description]['inet6'].append(item)
-            else:
-                ifaces[iface.Description]['up'] = False
-    return ifaces
+    return salt.utils.win_network.get_interface_info()
```
rares-pop: is it safe to do this, if importing salt.utils.win_network failed? (we swallow the exception if that fails)

salt/utils/win_network.py
```diff
@@ -0,0 +1,337 @@
...
+    Uses .NET 4.0+ to gather Network Interface information. Should only run on
+    Windows systems newer than Windows 7/Server 2008R2
+    Returns:
+        dict: A dictionary of information about all interfaces on the system
+    '''
+    clr.AddReference('System.Net')
+    interfaces = _get_network_interfaces()
+
+    int_dict = {}
+    for i_face in interfaces:
```
rares-pop: is it common to abbreviate something using underscore?

everywhere I've seen was 'iface' from 'interface'

salt/utils/win_network.py
```diff
@@ -0,0 +1,337 @@
...
+        int_dict[i_face.Name] = _get_base_properties(i_face)
+        int_dict[i_face.Name].update(_get_ip_base_properties(i_face))
```
rares-pop: doing int_dict[i_face.Name] on all following lines are doing the lookups each time.

Maybe it's worth having a separate dictionary and set it as int_dict[i_face.Name] once, at the end.

salt/utils/win_network.py
```diff
@@ -0,0 +1,337 @@
...
+    below will use WMI. Newer systems will use .NET.
+    Returns:
+        dict: A dictionary of information about the Network interfaces
+    '''
+    # On Windows 7 machines, use WMI as dotnet 4.0 is not available by default
+    if USE_WMI:
+        return get_interface_info_wmi()
+
+    # Massage the data returned by dotnet to mirror that returned by wmi
+    interfaces = get_interface_info_dot_net()
```
rares-pop: you are not using the same level of abstraction here.

In one case (wmi), you return what that worker is doing. In the other (.net), you call the worker than you do more extra stuff. Maybe you should have that worker do that extra stuff? Or maybe you can have a subworker that's doing what get_interface_info_dot_net() is doing, and have get_interface_info_dot_net() call that and do this extra stuff?

salt/utils/win_network.py
twangboy: I went with `i_face`

rares-pop (approved):
LGTM!

dwoz (approved):

PR #55820, opened 2020-01-08
SEP-17: Job acknowledgement events

lukasraska:
### What does this PR do?
Implements new event type `salt/job/<JID>/ack/<MID>` that is fired everytime minion receives job for execution.

saltstack/salt-enhancement-proposals#23

### Tests written?
Yes, but unit tests only. I've prepared integration tests as well, but they are too flaky to be used steadily (they use multithreaded Queue and it doesn't play well with these tests on various platforms).

### Commits signed with GPG?
Yes

TODO:
- Update release notes before release (not done, since not sure when will it be merged)


lukasraska:
re-run pr-opensuse15-py3

dwoz:
@lukasraska Please add your integration tests. I will take a look at making them less flaky.

lukasraska:
@dwoz Thanks for looking into this, I've pushed them and rebased to latest master. From my tries, the `setUpClass` method can sometimes raise `Empty` for no apparent reason (some environments work almost all the time, some raise it almost every time). Also I don't know if the Win test failures are related to this PR, but running just these tests is working fine on my win-server setup.

It also might be worth noting, before the change in logging, these tests using multithreaded Queue were completely failing on Win and OSX (they worked when print was put strategically, as I mentioned on SaltConf).

If you can think of how to make them more successful, `tests.integration.modules.test_event` can reuse the same mixin and can be re-enabled.

waynew:
> It also might be worth noting, before the change in logging, these tests using multithreaded Queue were completely failing on Win and OSX (they worked when print was put strategically, as I mentioned on SaltConf).

That's interesting, and hints at a race condition. It may be something else, but `print` involves a pretty hefty wait on that thread along with the system call(s). You should check if a `time.sleep(1)` accomplishes the same thing. That would be a strong indication that we're running into a race condition. If it's only the `print` that works, though, that's pretty confusing.

dwoz:
@lukasraska  are you still interested in pushing this over the finish line?

doc/ref/configuration/minion.rst
```diff
@@ -3586,3 +3586,14 @@ global pillar data.
...
 If set to ``False`` only the overriding pillar data will be available
 to the ``showpillar`` state.
+
+.. conf_minion:: acknowledge_jobs
+
+``acknowledge_jobs``
+--------------------
+
+.. versionadded:: Sodium
+
```
twangboy, line 3596: Chlorine or Argon

twangboy (changes requested):
The tests need to be migrated to pytest

twangboy:
Also needs a changelog

dwoz:
This can be re-based and re-opened if it should get into `3008.x`

PR #55823, opened 2020-01-09
Fix issue with overly long names in the LGPO module

twangboy:
### What does this PR do?
Fixes an issue in the LGPO module where policy parameters are extremely long. Some policy names were entire sentences pull from the adml file and prepended to the actual policy name.

This change does not affect ALL policies, only those that had prepended additional text. I estimate that to be 5 to 10 percent of policies. I added some logic to the state that will attempt to replace the old element name with the new element name. If this occurs, a Deprecation warning will be displayed in the comments. The warning will be displayed until Phosphorus.

This PR also replaces smart quotes, smart single quotes, em-dashes, and en-dashes with normal quotes and dashes. The unrecognized characters in the `Previous Behavior` section below contains smart quotes and an em dash. Those will be replaced and display properly. It will also make it so you don't have to add em dashes and smart quotes when setting policies. You will be able to use standard characters.

### Previous Behavior
 A recent update to the Windows Update Config policy recently gives us this as a policy setting:
```
            If you have selected ΓÇ£4 ΓÇô Auto download and schedule the installΓÇ¥ for your scheduled install day and specified a schedule, you also have the option to limit updating to a weekly, bi-weekly or monthly occurrence, using the options below: Every week:
                True
```

### New Behavior
With this PR the above policy setting becomes:
```
            Every week:
                True
```

### Tests written?
Yes

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/modules/win_lgpo.py
+++ b/salt/modules/win_lgpo.py
@@ -5651,33 +5651,15 @@ def _getAdmlPresentationRefId(adml_data, ref_id):
     helper function to check for a presentation label for a policy element
     '''
     search_results = adml_data.xpath('//*[@*[local-name() = "refId"] = "{0}"]'.format(ref_id))
-    prepended_text = ''
     if search_results:
         for result in search_results:
             the_localname = etree.QName(result.tag).localname
-            presentation_element = PRESENTATION_ANCESTOR_XPATH(result)
-            if presentation_element:
-                presentation_element = presentation_element[0]
-                if TEXT_ELEMENT_XPATH(presentation_element):
-                    for p_item in presentation_element.getchildren():
-                        if p_item == result:
-                            break
-                        else:
-                            if etree.QName(p_item.tag).localname == 'text':
-                                if prepended_text:
-                                    prepended_text = ' '.join((text for text in (prepended_text, getattr(p_item, 'text', '').rstrip()) if text))
-                                else:
-                                    prepended_text = getattr(p_item, 'text', '').rstrip() if getattr(p_item, 'text', '') else ''
-                            else:
-                                prepended_text = ''
-                    if prepended_text.endswith('.'):
-                        prepended_text = ''
             if the_localname == 'textBox' \
                     or the_localname == 'comboBox':
                 label_items = result.xpath('.//*[local-name() = "label"]')
                 for label_item in label_items:
                     if label_item.text:
-                        return (prepended_text + ' ' + label_item.text.rstrip().rstrip(':')).lstrip()
+                        return label_item.text.rstrip().rstrip(':').strip()
             elif the_localname == 'decimalTextBox' \
                     or the_localname == 'longDecimalTextBox' \
                     or the_localname == 'dropdownList' \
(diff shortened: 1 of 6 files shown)


twangboy:
@lomeroe What are your thoughts?

lomeroe:
Some of the "prepended text" unfortunately makes the policy item names make sense (in some cases you'll get the tail end of it only -- I don't remember what policy drove me to add the prepended text, but there are some), that would be the only downside I can come up with off the top of my head.

It looks like prepended text would be cleared if there was a period at the end of it (line 5673).  Which was probably for cases similar to this on other policies....Maybe a few other characters should be added, like the colon in your example (or other sentence ending characters) as opposed to dropping it completely...

twangboy:
Maybe this one: 
image
It renders as with old code:
```
        Network\Offline Files\Configure Slow link speed:
            ----------
            Value entered is [ bps / 100 ] --> Example: 128,000bps, enter 1280 Value:
                640
```
new code:
```
        Network\Offline Files\Configure Slow link speed:
            ----------
            Value:
                640
```
I think I still like the short one

twangboy:
Here's another one that renders pretty bad:
image

old code:
```
        Network\Offline Files\Configure slow-link mode:
            ----------
            Allows the administrator to configure when the automatic slow-link transition should happen for the specified UNC paths. The UNC path should be specified in the Value Name colume. The thresholds for throughput (in bits per second) and/or latency (in milliseconds) should be specified in the Value column. Examples: To enable the slow-link mode to apply when network throughput is less than 10000 bits per second and network latency is greater than 50 milliseconds:  Value Name="*" Value="Throughput=10000, Latency=50" To enable the slow-link mode to apply to all shares on a server named "server" when network latency is greater than 50 milliseconds:  Value Name="\\server\*" Value="Latency=50" To enable the slow-link mode to apply to the share named "\\server\share" when network throughput is less than 10000 bits per second:  Value Name="\\server\share\*" Value="Throughput=10000" UNC Paths:
                ----------
                **delvals.:

                \\server\share:
                    Throughput=10000
```
new code:
```
        Network\Offline Files\Configure slow-link mode:
            ----------
            UNC Paths:
                ----------
                **delvals.:

                \\server\share:
                    Throughput=10000
```

twangboy:
I added back the old code, but commented out so we can test it if we find one that seems like it would benefit from the old code.

lomeroe:
Yep, agree that the shorter names are easier and maybe items that needed the prepended text is the edge case (I just don't remember what policy element drove that being added -- so we could look at an alternative in those scenarios).

salt/modules/win_lgpo.py
```diff
@@ -5655,23 +5662,25 @@ def _getAdmlPresentationRefId(adml_data, ref_id):
...
-                            if etree.QName(p_item.tag).localname == 'text':
-                                if prepended_text:
-                                    prepended_text = ' '.join((text for text in (prepended_text, getattr(p_item, 'text', '').rstrip()) if text))
-                                else:
-                                    prepended_text = getattr(p_item, 'text', '').rstrip() if getattr(p_item, 'text', '') else ''
-                            else:
-                                prepended_text = ''
-                    if prepended_text.endswith('.'):
-                        prepended_text = ''
+            # presentation_element = PRESENTATION_ANCESTOR_XPATH(result)
```
dwoz: We should remove code, not comment it out.

dwoz (changes requested):

salt/modules/win_lgpo.py
twangboy: I put that back in, but commented out in case we found the policy that code was written for.

twangboy: But, we can dig that out of an older release.

dwoz (approved):

PR #55859, opened 2020-01-14
[PyTest #34] Switch to plain asserts

s0undt3ch:
### What does this PR do?
See title


mchugh19:
I had actually just written some new tests and adjusted the old ones to be in the self.assert<whatever> model. What's the reason for the conversation to plain asserts? When looking at test output, I do find the diff model rather helpful.

s0undt3ch:
The PR is to be merged after we switch to PyTest.

PyTest is smart enough to show us proper diff's, at least for built-in types.

Asserting dictionary equality in pytest is as simple as:
```python
assert {1: 1} == {2: 2}
```
Which would fail and we'd get a nice diff printed.

s0undt3ch:
https://docs.pytest.org/en/latest/assert.html

s0undt3ch:
So, in conclusion, we write more readable code without loosing the diff

s0undt3ch:
Removed from Aluminum and actually closing. We now have salt-rewrite in place and once we're ready, it will rewrite the whole test suite. Right now, it's only sensible to `tests/pytests`

Issue #55816, opened 2020-01-08
smartos.vm_present failing if image_uuid is an "official" docker image


garethhowell:
### Description of Issue
When creating a SmartOS zone from a docker image that is a single name (e.g. influxdb rather than influxdata/influxdb), vmadm create fails with `Comment: provisioning dataset null with brand lx is not supported`

### Setup
State file contains
```
influxdb.agdon.net:
  smartos.vm_present:
    - config:
        auto_import: true
        reprovision: true
    - vmconfig:
        image_uuid: influxdb:latest
        brand: lx
        alias: influxdb
...
```
### Steps to Reproduce Issue
`salt-call state.apply influxdb.sls`

### Versions Report
Salt Version:
           Salt: 2019.2.0-566-g79217f3
 
Dependency Versions:
           cffi: 1.12.3
       cherrypy: Not Installed
       dateutil: Not Installed
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
         Jinja2: 2.10.1
        libgit2: Not Installed
       M2Crypto: Not Installed
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.5.6
   mysql-python: Not Installed
      pycparser: 2.19
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: Not Installed
         Python: 2.7.16 (default, Jul  3 2019, 09:06:51)
   python-gnupg: Not Installed
         PyYAML: 5.1.1
          PyZMQ: 18.0.2
          smmap: Not Installed
        timelib: Not Installed
        Tornado: 5.1.1
            ZMQ: 4.3.1
 
System Versions:
           dist:   
         locale: UTF-8
        machine: i86pc
        release: 5.11
         system: SunOS
        version: Not Installed

Ch3LL:
ping @sjorge any ideas here?

sjorge:
I have one local fix I need to get in eventually, it might fix it... although I am not sure.
Rather busy with work the last halfish year so didn't have time to get stuff in or better test them.

```
--- /opt/local/lib/python2.7/site-packages/salt/modules/smartos_imgadm.py    2019-10-26 11:38:15.000000000 +0000
+++ smartos_imgadm.py   2019-10-26 14:46:38.765958032 +0000
@@ -56,7 +56,7 @@

     if image and 'Error' in image:
         ret = image
-    elif image and 'manifest' in image:
[...]

Issue #55185, opened 2019-11-01
salt.modules.pdbedit doesn't work on samba older than 4.8


sjorge:
### Description of Issue
Someone contacted me via email that pdbedit module/state wasn't working.
The person is using samba 4.3.11 and the required fields in the pdbedit output are missing.

I was using samba 4.8.x when writing this module, there should probably be a version check that doesn't load the module when the version is older.

### Setup
N/a

### Steps to Reproduce Issue
Try and use pdbedit module or state on samba older than 4.8.x

### Versions Report
(from e-mail)
```
Salt Version:
           Salt: 2018.3.2
 
Dependency Versions:
           cffi: Not Installed
       cherrypy: Not Installed
       dateutil: 2.5.3
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.8
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: 1.0.3
   msgpack-pure: Not Installed
 msgpack-python: 0.4.6
   mysql-python: Not Installed
      pycparser: Not Installed
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: Not Installed
         Python: 2.7.12 (default, Dec  4 2017, 14:50:18)
   python-gnupg: 0.3.8
         PyYAML: 3.11
          PyZMQ: 15.2.0
           RAET: Not Installed
          smmap: Not Installed
        timelib: Not Installed
        Tornado: 4.2.1
            ZMQ: 4.1.4
 
System Versions:
           dist: Ubuntu 16.04 xenial
         locale: UTF-8
        machine: x86_64
        release: 4.4.0-116-generic
         system: Linux
        version: Ubuntu 16.04 xenial
```

sjorge:
```
[atlas :: sjorge][~]
[!]$ pdbedit -V
Version 4.9.8
```

Seems to print the version.

Ch3LL:
we will probably want to add this check to salt/modules/pdbedit.py in `__virtual__`. Did you want to take a stab at it or i can?

sjorge:
I’m a bit busy atm, I was hoping to take a stab at it after my current open PR gets in though.

But I’d you do, let me know and I will give it a test spin.

~ sjorge

sjorge:
Nope, still an issue... might have some time to work on this later this month

sjorge:
[...]

PR #55861, opened 2020-01-14
Add test framework for pdbedit and state smartos tests

Ch3LL:
### What does this PR do?
Add test framework for smartos module and pdbedit state/execution modules.

Also ensure we add a comment in `config_present` when we cannot write to the config file.

fyi @sjorge

### DIFF
--- a/tests/unit/modules/test_pdbedit.py
+++ b/tests/unit/modules/test_pdbedit.py
@@ -1,0 +1,27 @@
+# -*- coding: utf-8 -*-
+
+# Import python libs
+from __future__ import absolute_import, print_function, unicode_literals
+
+# Import Salt Libs
+import salt.modules.pdbedit as pdbedit
+
+# Import Salt Testing Libs
+from tests.support.mixins import LoaderModuleMockMixin
+from tests.support.unit import TestCase
+
+
+class PdbeditTestCase(TestCase, LoaderModuleMockMixin):
+    '''
+    TestCase for salt.modules.pdbedit module
+    '''
+
+    def setup_loader_modules(self):
+        return {pdbedit: {}}
+
+    def test_generate_nt_hash(self):
+        '''
+        Test salt.modules.pdbedit.generate_nt_hash
+        '''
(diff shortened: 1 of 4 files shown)


tests/unit/states/test_pdbedit.py
```diff
@@ -0,0 +1,31 @@
...
+    def setup_loader_modules(self):
+        return {pdbedit: {}}
+
+    def test_generate_absent(self):
+        '''
+        Test salt.states.pdbedit.absent when
+        user is already absent
+        '''
+        name = 'testname'
+        with patch.object(pdbedit, '__salt__', return_value=MagicMock()):
```
sjorge: What does MagicMock() doe in this case? It sets __salt__ on pdbedit, but to what?

tests/unit/modules/test_pdbedit.py
```diff
@@ -0,0 +1,27 @@
...
+
+class PdbeditTestCase(TestCase, LoaderModuleMockMixin):
+    '''
+    TestCase for salt.modules.pdbedit module
+    '''
+
+    def setup_loader_modules(self):
+        return {pdbedit: {}}
+
+    def test_generate_nt_hash(self):
```
sjorge: For #55185 I will need a similar test but need to mock the result of cmd.run... can I still include the mock data in the test file or does it need to go into tests.support.<something> ?

sjorge:
👍 I'll try to get PR's open for #55816 and #55185 soonish,

tests/unit/states/test_pdbedit.py
Ch3LL: i actually intended on mocking `pdbedit.list`  specificly, note the entire `__salt__` dict. this was a mistake on my part. thanks for pointing out. working on a fix.

tests/unit/modules/test_pdbedit.py
Ch3LL: the test file should work

sjorge (approved):

dwoz (approved):

Issue #55870, opened 2020-01-15
[Neon] Grains in minion startup event #54948


max-arnold:
I just discovered that [this feature](https://github.com/saltstack/salt/pull/54948) (when enabled) will append grains to other events sent from a minion (not just the start event):

Both commands:

```
sudo salt minion1 state.single event.send name='tag' data='{"data":"my event data"}'
sudo salt minion1 event.fire_master '{"data":"my event data"}' 'tag'
```

result in the following master event:

```
tag	{
    "_stamp": "2020-01-15T07:55:24.548188",
    "cmd": "_minion_event",
    "data": {
        "data": "my event data"
    },
    "grains": {
        "machine_id": "599118d5b9619443ac3166fb0e59349e",
        "uuid": "599118d5-b961-9443-ac31-66fb0e59349e"
    },
    "id": "minion1",
    "pretag": null,
    "tag": "tag"
}
```

`def _fire_master_minion_start(self)` seems to be better suited for this code.

> was that a suggestion that we should create a new PR to update this behavior? If so, would you mind opening an issue so it doesn't get lost here in the PR? Otherwise feel free to ignore this comment

@waynew Yes, that was a suggestion that something should be fixed - either the actual behavior, or the setting name and docs.

CC: @admd

Ch3LL:
thanks for opening up the issue with the details. really appreciate it.

sagetherage:
@max-arnold @admd looks like the PR is merged ok to close this ticket knowing it is getting into the Neon release?

PR #54948, opened 2019-10-10
Pass grains in minion startup event

admd:
### Note
The previous PR against develop branch https://github.com/saltstack/salt/pull/54087 has more discussion about the background for this change.

### What does this PR do?
Adds a configuration option to have a selection of grains in the minion start event.

Reason behind this is better integration with [Uyuni]( https://github.com/uyuni-project/uyuni), especially when a lot of minions are starting in a short time frame - we need to get some minimal data (at least the `machine_id`) and getting it separately as a reaction to the minion start event (via a separate call or a startup state) generates considerable load on the master.

By default, there will be no grains passed to the event.

### Previous Behavior
```
salt/minion/test-minion.tf.local/start	{
    "_stamp": "2019-08-01T10:46:22.840597", 
    "cmd": "_minion_event", 
    "data": "Minion test-minion.tf.local started at Thu Aug  1 12:46:22 2019", 
    "id": "test-minion.tf.local", 
    "pretag": null, 
    "tag": "salt/minion/test-minion.tf.local/start"
}
```

### New Behavior
```
salt/minion/test-minion.tf.local/start	{
    "_stamp": "2019-08-01T10:46:58.696390", 
    "cmd": "_minion_event", 
    "data": "Minion test-minion.tf.local started at Thu Aug  1 12:46:58 2019", 
    "grains": {
        "uuid": "f4cd2a8c-eb73-11e9-81b4-2a2ae2dbcce4", 
        "machine_id": "21ee22c6906887c2dd6adeed5d404145"
    }, 
    "id": "test-minion.tf.local", 
    "pretag": null, 
    "tag": "salt/minion/test-minion.tf.local/start"
}
```

### Tests written?

Yes

### Commits signed with GPG?

No

PR #55885, opened 2020-01-16
Restrict the 'start_event_grains' only to the start events

admd:
### What does this PR do?
https://github.com/saltstack/salt/pull/54948 introduced an unintended effect where `start_event_grains` were being passed to events other than start event, which wasn't the idea. This PR restrict these grains to startup events.

https://github.com/saltstack/salt/issues/55870
### Previous Behavior
Grains defined against `start_event_grains` were being passed to events other that start events.

### New Behavior
Grains defined against `start_event_grains` are being passed to only start events

### Tests written?
The existing test has been modified 

Yes

### Commits signed with GPG?

No

### DIFF
--- a/tests/unit/test_minion.py
+++ b/tests/unit/test_minion.py
@@ -341,6 +341,22 @@ class MinionTestCase(TestCase, AdaptedConfigurationTestCaseMixin):
         finally:
             minion.destroy()
 
+    def test_when_other_events_fired_and_start_event_grains_are_set(self):
+        mock_opts = self.get_config('minion', from_scratch=True)
+        mock_opts['start_event_grains'] = ["os"]
+        io_loop = tornado.ioloop.IOLoop()
+        io_loop.make_current()
+        minion = salt.minion.Minion(mock_opts, io_loop=io_loop)
+        try:
+            minion.tok = MagicMock()
+            minion._send_req_sync = MagicMock()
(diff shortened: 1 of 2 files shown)


admd:
Thank you @max-arnold for noticing it. It would be nice if you could test it as well. 
cc @waynew

tests/unit/test_minion.py
```diff
@@ -318,7 +318,7 @@ def test_when_passed_start_event_grains(self):
         try:
             minion.tok = MagicMock()
             minion._send_req_sync = MagicMock()
-            minion._fire_master('Minion has started', 'minion_start')
+            minion._fire_master('Minion has started', 'minion_start', include_startup_grains=True)
```
Ch3LL: any chance we can also add a test that mirrors the test case found in issue https://github.com/saltstack/salt/issues/55870

with `start_event_grains` set, fire a non-startup event and ensure grains data is not in the event?

also heads up you have a lint failure

Ch3LL (changes requested):

tests/unit/test_minion.py
max-arnold: Yep, +1 for the test to check that other events do not contain any grains.

admd: @Ch3LL thank you for having a look. I have added a test for this case as well.

Ch3LL (approved):

PR #55893, opened 2020-01-16
Remove references to `include_default`

twangboy:
### What does this PR do?
The `include_default` parameter was removed from the `salt.utils.win_reg.list_values` function. It wasn't removed from other sources that were calling that function with `include_default`.

This PR removes references to `include_default`

Found in testing

### Previous Behavior
reg.list_values would stacktrace

### New Behavior
Now it doesn't stacktrace

### Tests written?
No

### Commits signed with GPG?
Yes


waynew (approved):
Were these changes already included in something listed in the CHANGELOG or should they be listed in there, probably as a fix? (i.e. is this bug in 2019.2.X or is it only master?)

cmcmarrow (approved):

dwoz (approved):

PR #55856, opened 2020-01-13
note only applies to `salt` and `salt-ssh` commands

vernondcole:
### What does this PR do?
Removes a NOTE which does not apply to several CLI commands

### What issues does this PR fix or reference?
#40888

### Old Behavior
Note was included with salt-call, salt-key, and salt-cloud commands, which do not have a --static option. Attempt to add the option suggested by the Note would give an error.

### New Behavior
Note is only included in docs for `salt` and `salt-ssh` commands.

### Tests written?
No.  PR addresses documentation only.

`make html` was run and correct results appear on the appropriate pages.

### Commits signed with GPG?

Yes

Issue #40888, opened 2017-04-26
salt-call has no --static option (documentation error?)


vernondcole:
### Setup

Salt minion installed.

Read documentation for CLI of salt-call...

 https://docs.saltstack.com/en/latest/ref/cli/salt-call.html#output-options

... which says: `If using --out=json, you will probably want --static as well.`

### Steps to Reproduce Issue

`sudo salt-call --local grains.get master --out=json --static`

```
Usage: salt-call [options] <function> [arguments]

salt-call: error: no such option: --static
```

### Versions Report

Salt Version:
           Salt: 2016.11.4
 
Dependency Versions:
           cffi: 1.5.2
       cherrypy: Not Installed
       dateutil: 2.4.2
      docker-py: Not Installed
          gitdb: 0.6.4
      gitpython: 1.0.1
          ioflo: Not Installed
         Jinja2: 2.8
        libgit2: 0.24.0
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: 1.0.3
   msgpack-pure: Not Installed
 msgpack-python: 0.4.6
   mysql-python: Not Installed
      pycparser: 2.14
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: 0.24.0
         Python: 2.7.12 (default, Nov 19 2016, 06:48:10)
   python-gnupg: Not Installed
         PyYAML: 3.12
          PyZMQ: 15.2.0
           RAET: Not Installed
          smmap: 0.9.0
        timelib: Not Installed
        Tornado: 4.2.1
            ZMQ: 4.1.4
 
System Versions:
           dist: Ubuntu 16.04 xenial
        machine: x86_64
        release: 4.4.0-21-generic
         system: Linux
        version: Ubuntu 16.04 xenial

gtmanfred:
I believe that this is caused by the fact that we use the same parser for salt and salt-call.

We should get the documentation fixed, because salt-call does not have a --static option.

vernondcole:
This document still has the error. [develop version](https://docs.saltstack.com/en/develop/ref/cli/salt-call.html#output-options)

gtmanfred:
a pr to fix it would be greatly appreciated.

vernondcole:
Ungh! Okay.  I'll try to find a round tuit.

vernondcole:
issue is still current -- has not been corrected

vernondcole:
@gtmanfred Please review PR #55856

PR #55897, opened 2020-01-16
the note only applies to SALT and SALT_SSH commands

vernondcole:
What does this PR do?

Removes a NOTE which does not apply to several CLI commands

#40888

Old Behavior

Note was included with salt-call, salt-key, and salt-cloud commands, which do not have a --static option. Attempt to add the option suggested by the Note would give an error.
New Behavior

Note is only included in docs for salt and salt-ssh commands.
Tests written?

No. PR addresses documentation only.

make html was run and correct results appear on the appropriate pages.

Commits signed with GPG?

Apparently not. Commits from my MacBook seem to be broken, now, But the GPG key is present. Will debug now that I see it is no longer working.

This is a re-issue of #55856 which was applied to the 2019.2.3 branch (as instructed by https://docs.saltstack.com/en/latest/topics/development/contributing.html#which-salt-branch) rather than master. My attempt to rebase failed in git hell.

### DIFF
--- a/doc/ref/cli/_includes/output-options.rst
+++ b/doc/ref/cli/_includes/output-options.rst
@@ -13,13 +13,6 @@ Output Options
     Salt will fall back on the ``pprint`` outputter and display the return data
     using the Python ``pprint`` standard library module.
 
-    .. note::
-        If using ``--out=json``, you will probably want ``--static`` as well.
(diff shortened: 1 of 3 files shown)


vernondcole:
@Ch3LL this is a new PR, now against master branch.

dwoz (approved):

PR #55906, opened 2020-01-17
smartos.vm_present could not handle nics with vrrp_vrid property

sjorge:
### What does this PR do?
SmartOS will automatically calculate (and override) the mac property of any nics that have `vrrp_vrid` property set.

This causes a lot of issues:
- we keep trying to add a new nic if the mac that was provided is incorrect
- we keep trying to delete the existing nic with the mac based on the vrid
- we keep trying to incorrect update a nic if the vrid was change (cannot be update, remove and add new nic)

n/a

### Previous Behavior
We would try to add/remove/update the incorrect nic when `vrrp_vrid` property was present making the state module fail.

### New Behavior
We now account for this in `_parse_vmconfig()`, this function is used to map unique identifiers for various smartos resource types (nics, disks, ...)

We will use the provided mac address from the resource key **unless* the `vrrp_vrid` property is present, then generate the mac based on the provided vrid.

This results in the nic being correctly added, when the vrid is changes this results in a remove of the old nic and adding of the new nic as is expected by vmadm.

### Tests written?
Yes, I've added tests to test the mac handling of `_parse_vmconfig()`

### Commits signed with GPG?
No

### DIFF
--- a/salt/states/smartos.py
+++ b/salt/states/smartos.py
@@ -246,6 +246,11 @@ def _parse_vmconfig(config, instances):
                     ## some property are lowercase
                     if 'mac' in instance_config:
                         instance_config['mac'] = instance_config['mac'].lower()
+                    ## calculate mac from vrrp_vrid
+                    if 'vrrp_vrid' in instance_config:
+                        instance_config['mac'] = "00:00:5e:00:01:{0}".format(
+                            hex(int(instance_config['vrrp_vrid'])).split('x')[-1].zfill(2),
+                        )
                     vmconfig[prop].append(instance_config)
     else:
         log.error('smartos.vm_present::parse_vmconfig - failed to parse')
@@ -768,7 +773,8 @@ def vm_present(name, vmconfig, config=None):
         'instance': {
             'nics': 'mac',
             'disks': 'path',
-            'filesystems': 'target'
+            'filesystems': 'target',
+            'pci_devices': 'path',
         },
         'create_only': [
             'filesystems'
--- a/tests/unit/states/test_smartos.py
+++ b/tests/unit/states/test_smartos.py
@@ -5,6 +5,7 @@ from __future__ import absolute_import, print_function, unicode_literals
 
 # Import Salt Libs
 import salt.states.smartos as smartos
+from salt.utils.odict import OrderedDict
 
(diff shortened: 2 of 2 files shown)


sjorge:
Tests are working local, but that was with py3... will look into py2

sjorge:
Interesting, the order is different between versions!

python3:
```
00:00:5e:00:01:01
00:00:5e:00:01:f0
00:22:06:00:00:01
```

```OrderedDict([('nics', [OrderedDict([('vrrp_vrid', 1), ('vrrp_primary_ip', '12.34.5.6'), ('mac', '00:00:5e:00:01:01')]), OrderedDict([('vrrp_vrid', 240), ('vrrp_primary_ip', '12.34.5.6'), ('mac', '00:00:5e:00:01:f0')]), OrderedDict([('ips', ['12.34.5.6/24']), ('mac', '00:22:06:00:00:01')])])])```

python2
```
00:22:06:00:00:01
00:00:5e:00:01:01
00:00:5e:00:01:f0
```

```OrderedDict([(u'nics', [OrderedDict([(u'ips', [u'12.34.5.6/24']), (u'mac', u'00:22:06:00:00:01')]), OrderedDict([(u'vrrp_vrid', 1), (u'vrrp_primary_ip', u'12.34.5.6'), (u'mac', u'00:00:5e:00:01:01')]), OrderedDict([(u'vrrp_vrid', 240), (u'vrrp_primary_ip', u'12.34.5.6'), (u'mac', u'00:00:5e:00:01:f0')])])])```

This is very weird, because the input is also an ordereddict and the order should not change, as we see on python3... I am unsure why it is changing on python2...

@Ch3LL any ideas? Have you seen this before?

sjorge:
I can change the tests and call the function 3 times with a seperate ret object to test... as the order is not important for this test... but I do believe OrderedDict might be broken?

sjorge:
nvm, I am dumb as a rock and incorrectly create the OrderedDict.

sjorge:
Current test failures look to be for different modules?

Ch3LL (approved):

cmcmarrow (approved):

sjorge:
@Ch3LL bump

PR #51672, opened 2019-02-15
Allow minions to return job information to any configured master

sbrennan4:
### What does this PR do?

This PR introduces a new minion configuration option, `master_return_strategy`. Can be `source` or `any`. The default of `source` preserves the current behavior, namely, the minion will only attempt to return the job results to the master that sent the job. If set to `any`, then the minion will first attempt to return the job results to the master that sent the job. If that fails then the minion will attempt to return the job results to the other configured masters, one by one, until successful or the master list has been exhausted. `master_alive_interval` must also be set. This is used to keep track of which masters are currently connected.

### Previous Behavior

When a master becomes unavailable after sending a job but before receiving the return information, that return info will never make it back to the masters. If relying on `master_job_cache` or `event_returns` on the master side then the job return information is lost forever.

### New Behavior

When the initiating master becomes unavailable after firing a job to a minion then the minion will attempt to return the job information to another configured master.

### Tests written?

Hoping current tests are sufficient.

### Commits signed with GPG?

No

Please review [Salt's Contributing Guide](https://docs.saltstack.com/en/latest/topics/development/contributing.html) for best practices.

See GitHub's [page on GPG signing](https://help.github.com/articles/signing-commits-using-gpg/) for more information about signing commits with GPG.

PR #55907, opened 2020-01-17
Allow minions to return job information to any configured master

sbrennan4:
### What does this PR do?

This PR introduces a new minion configuration option, master_return_strategy. Can be source or any. The default of source preserves the current behavior, namely, the minion will only attempt to return the job results to the master that sent the job. If set to any, then the minion will first attempt to return the job results to the master that sent the job. If that fails then the minion will attempt to return the job results to the other configured masters, one by one, until successful or the master list has been exhausted. master_alive_interval must also be set. This is used to keep track of which masters are currently connected.

### Previous Behavior

When a master becomes unavailable after sending a job but before receiving the return information, that return info will never make it back to the masters. If relying on master_job_cache or event_returns on the master side then the job return information is lost forever.

### New Behavior

When the initiating master becomes unavailable after firing a job to a minion then the minion will attempt to return the job information to another configured master.

### Tests written?

Not yet

### Commits signed with GPG?

No

### DIFF
--- a/doc/ref/configuration/minion.rst
+++ b/doc/ref/configuration/minion.rst
@@ -280,6 +280,28 @@ to the next master in the list if it finds the existing one is dead.
 
     master_alive_interval: 30
 
+.. conf_minion:: master_return_strategy
+
+``master_return_strategy``
+--------------------------
+
+.. versionadded:: 2018.x.x
+
+Default: ``source``
+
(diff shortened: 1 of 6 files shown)


sbrennan4:
This is a copy of https://github.com/saltstack/salt/pull/51672 but for the master branch.

dwoz:

PR #52368, opened 2019-03-30
Unit test to require execution module documentation

mchugh19:
### What does this PR do?
Adds unit test to validate that salt module, state, returners, wheel, etc files have a matching rst in docs and checks that doc rst files have matching py.
Also adds missing docs so that the tests pass

### Previous Behavior
Modules could be added without documentation link

### New Behavior
Unit test will catch that modules do not have associated docs
```
python tests\runtests.py -n unit.test_doc.DocTestCase.test_states_doc_files

AssertionError: 'zabbix_usermacro' not found in ['acme', 'alias', '...
'zone', 'zookeeper', 'zpool'] : State file zabbix_usermacro is missing documentation in ...salt/doc/ref/states/all
```

### Tests written?
Yes

### Commits signed with GPG?
No

PR #53296, opened 2019-05-29
rename shadow to linux_shadow for consistency

gtmanfred:
### What does this PR do?
Right now, the docs for shadow.rst just talk about there being a shadow module, but it is impossible to see the shadow docs for linux.  This moves the linux module to linux_shadow.py and links them
### Commits signed with GPG?

Yes

Issue #52000, opened 2019-03-06
Messed up module documentation


OrangeDog:
https://docs.saltstack.com/en/2018.3/ref/modules/all/salt.modules.service.html

There are 6 modules (group, kernelpkg, pkg, service, shadow, user) duplicated at the top of the list, and half the service modules have been renamed (launchctl, smf, systemd, upstart) so the links don't work.

![screenshot](https://user-images.githubusercontent.com/675056/53879715-0d15ec00-4007-11e9-836c-0f151f8a8c88.png)

twangboy:
@jacobhammons @newwebash ^^^

newwebash:
I'm tagging @KChandrashekhar and @Ch3LL since they're more familiar with the open Salt docs and processes than I am.

Ch3LL:
i can definitely fix up the services issue, but I'm not quit sure what you mean by "There are 6 modules (group, kernelpkg, pkg, service, shadow, user) duplicated at the top of the list," I think you are referencing the virtual models? That should stay as we are documenting which of these modules are virtual modules

OrangeDog:
@Ch3LL I mean they're both listed on their own at the top of the list, and also in their correct alphabetical order. If it's deliberate that they're listed separately there should be some sort of divider, and possibly no duplicate in the rest of the list.

Ch3LL:
![image](https://user-images.githubusercontent.com/6886822/54548443-50187d80-497e-11e9-8fdc-70f048867208.png)

I believe you are referring to these virtual modules? In my opinion there is a clear division here.

OrangeDog:
@Ch3LL see my screenshot.
There is no division in the sidebar that appears on every page.

sagetherage:
fixed with port to master -- will be released in sodium

Issue #51153, opened 2019-01-12
Shadow documentation missing


MTecknology:
### Description of Issue/Question

The documentation produced for `salt.modules.shadow` [1] mentions that `shadow` [2] is the resource that provides `shadow` support for Linux.

Unfortunately, these names conflict and it's not possible to get to documentation for the linux/shadow module [2].

I suspect this is an easy fix (rename the module), but I don't have the time to provide a patch... just the complaint.

[1] https://docs.saltstack.com/en/latest/ref/modules/all/salt.modules.shadow.html
[2] https://github.com/saltstack/salt/blob/develop/salt/modules/shadow.py

max-arnold:
Duplicate of #49335

racooper:
This is still an ongoing issue.  Web docs for the shadow module are not available.  Clicking on "shadow" for linux on https://docs.saltstack.com/en/latest/ref/modules/all/salt.modules.shadow.html#module-salt.modules.shadow just returns to the same page.

MTecknology:
Can someone add that label that prevents marking this as stale?

johnnybubonic:
not stale; pending merge of https://github.com/saltstack/salt/pull/55914

johnnybubonic:
keep open; this problem still exists.

johnnybubonic:
issue still exists.

PR #55914, opened 2020-01-19
Port #52368 doc requirement tests to master

mchugh19:
### What does this PR do?
Ports #52368 to master

Adds tests to ensure modules have needed doc entries and ensures all defined doc entries have associated modules

### Tests written?
Yes

### Commits signed with GPG?
No


max-arnold:
Do you think it is possible to also have a test to detect unreachable documentation pages?

https://github.com/saltstack/salt/issues/49335
https://github.com/saltstack/salt/issues/51153
https://github.com/saltstack/salt/pull/53296 (fix)

mchugh19:
@max-arnold Turns out we sure can! 

Added a test to check each module's rst for the text ".. _virtual" which successfully detects problems with shadow and service execution module naming.

With this additional check, the doc test is expected to fail based on the mentioned issues with shadow and service. I don't think correcting those should be part of his PR, so waiting for input from core team on next steps. I suggest merging #53296 and an equivalent for service.

xeacott:
Looks like some builds are failing

mchugh19:
@xeacott yep! Please review the posted comments. This is creating new tests, some of which are detecting currently broken things and I asked for how the core team wants it handled. 

I can remove the test for broken stuff and you can merge the doc fixes. Or the broken stuff can be fixed and you can merge the whole thing. Just let me know what you want here.

mchugh19:
To verify tests, I merged #53296 into this branch and performed a similar move for the service module.

max-arnold:
@mchugh19 I added more checks to your PR to verify that index files have no duplicates and are sorted (see https://github.com/saltstack/docs-hub/issues/15). It also contains a partial attempt to check for nested modules (currently fails, not sure if it makes sense to fix it). If you decide to fix it, then you can delete the check for stray module docs (the docs won't build if the `py` file is missing).

It is mostly a PoC, feel free to discard/rewrite it.

CC: @myii

<details><summary>Patch</summary>

```diff
diff --git a/tests/unit/test_doc.py b/tests/unit/test_doc.py
index bbfd34da10..5bfaa9e679 100644
--- a/tests/unit/test_doc.py
+++ b/tests/unit/test_doc.py
@@ -9,6 +9,7 @@ from __future__ import absolute_import
 import os
 import re
 import logging
+import collections
 
 # Import Salt Testing libs
 from tests.support.unit import TestCase
@@ -119,6 +120,12 @@ class DocTestCase(TestCase):
                 module_name = os.path.splitext(file)[0]
                 if module_name not in skip_module_files:
                     module_files.append(module_name)
+            elif (os.path.isdir(os.path.join(full_module_dir, file)) and
+                  not file.startswith('_') and
+                  os.path.isfile(os.path.join(full_module_dir, file, '__init__.py'))):
+                module_name = file
+                if module_name not in skip_module_files:
+                    module_files.append(module_name)
 
         # Build list of documentation files
         module_docs = []
@@ -133,6 +140,22 @@ class DocTestCase(TestCase):
                 if doc_name not in skip_doc_files:
                     module_docs.append(doc_name)
 
+        module_index_file = os.path.join(full_module_doc_dir, 'index.rst')
+        with open(module_index_file, 'rb') as fp:
+            module_index_contents = fp.read().decode('utf-8')
+
+        module_index_block = re.search(r"""
+            \.\.\s+autosummary::\s*\n
+            (\s+:[a-z]+:.*\n)*
+            (\s*\n)+
+            (?P<mods>(\s*[a-z0-9_\.]+\s*\n)+)
+        """, module_index_contents, flags=re.VERBOSE)
+
+        module_index = re.findall(
+            r"""\s*([a-z0-9_\.]+)\s*\n""",
+            module_index_block.group('mods')
+        )
+
         # Check that every module has associated documentation file
         for module in module_files:
             self.assertIn(module,
@@ -140,6 +163,13 @@ class DocTestCase(TestCase):
                           'module file {0} is missing documentation in {1}'.format(module,
                                                                                    full_module_doc_dir))
 
+            # Check that every module is listed in the index file
+            self.assertIn(
+                module,
+                module_index,
+                'module file {0} is missing in {1}'.format(module, module_index_file)
+            )
+
             # Check if .rst file for this module contains the text
             # ".. _virtual" indicating it is a virtual doc page
             full_module_doc_name = os.path.join(full_module_doc_dir, doc_prefix + module + '.rst')
@@ -155,6 +185,26 @@ class DocTestCase(TestCase):
                           module_files,
                           'Doc file {0} is missing associated module in {1}'.format(doc_file,
                                                                                     full_module_dir))
+        # Check that a module index is sorted
+        sorted_module_index = sorted(module_index)
+        self.assertEqual(
+            module_index,
+            sorted_module_index,
+            msg='Module index is not sorted: {}'.format(module_index_file)
+        )
+
+        # Check for duplicates inside of a module index
+        module_index_duplicates = [mod for mod, count in collections.Counter(module_index).items() if count > 1]
+        if module_index_duplicates:
+            self.fail('Module index {0} contains duplicates: {1}'.format(module_index_file, module_index_duplicates))
+
+        # Check for stray module docs
+        stray_modules = set(module_index).difference(module_files)
+        if stray_modules:
+            self.fail('Stray module names {0} in the doc index {1}'.format(sorted(list(stray_modules)), module_index_file))
+        stray_modules = set(module_docs).difference(module_files)
+        if stray_modules:
+            self.fail('Stray module doc files {0} in the doc folder {1}'.format(sorted(list(stray_modules)), full_module_doc_dir))
 
     def test_auth_doc_files(self):
         '''
@@ -226,6 +276,20 @@ class DocTestCase(TestCase):
         doc_dir = ['doc', 'ref', 'engines', 'all']
         self._check_doc_files(skip_module_files, module_dir, skip_doc_files, doc_dir)
 
+    def test_executors_doc_files(self):
+        '''
+        Ensure executor modules have associated documentation
+
+        doc example: doc/ref/executors/all/salt.executors.docker.rst
+        engine module example: salt/executors/docker.py
+        '''
+
+        skip_module_files = ['__init__']
+        module_dir = ['salt', 'executors']
+        skip_doc_files = ['index', 'all']
+        doc_dir = ['doc', 'ref', 'executors', 'all']
+        self._check_doc_files(skip_module_files, module_dir, skip_doc_files, doc_dir)
+
     def test_fileserver_doc_files(self):
         '''
         Ensure fileserver modules have associated documentation
@@ -262,7 +326,7 @@ class DocTestCase(TestCase):
         execution module example: salt/modules/zabbix.py
         '''
 
-        skip_module_files = ['__init__']
+        skip_module_files = ['__init__', 'inspectlib']
         module_dir = ['salt', 'modules']
         skip_doc_files = ['index', 'group', 'inspectlib', 'inspectlib.collector', 'inspectlib.dbhandle',
                           'inspectlib.entities', 'inspectlib.exceptions', 'inspectlib.fsdb',
@@ -271,6 +335,20 @@ class DocTestCase(TestCase):
         doc_dir = ['doc', 'ref', 'modules', 'all']
         self._check_doc_files(skip_module_files, module_dir, skip_doc_files, doc_dir)
 
+    def test_netapi_doc_files(self):
+        '''
+        Ensure netapi modules have associated documentation
+
+        doc example: doc/ref/netapi/all/salt.netapi.rest_cherrypy.rst
+        module example: salt/netapi/rest_cherrypy
+        '''
+
+        skip_module_files = ['__init__']
+        module_dir = ['salt', 'netapi']
+        skip_doc_files = ['index', 'all']
+        doc_dir = ['doc', 'ref', 'netapi', 'all']
+        self._check_doc_files(skip_module_files, module_dir, skip_doc_files, doc_dir)
+
     def test_output_doc_files(self):
         '''
         Ensure output modules have associated documentation
@@ -425,6 +503,20 @@ class DocTestCase(TestCase):
         doc_dir = ['doc', 'ref', 'thorium', 'all']
         self._check_doc_files(skip_module_files, module_dir, skip_doc_files, doc_dir)
 
+    def test_token_doc_files(self):
+        '''
+        Ensure token modules have associated documentation
+
+        doc example: doc/ref/tokens/all/salt.tokens.localfs.rst
+        module example: salt/tokens/localfs.py
+        '''
+
+        skip_module_files = ['__init__']
+        module_dir = ['salt', 'tokens']
+        skip_doc_files = ['index', 'all']
+        doc_dir = ['doc', 'ref', 'tokens', 'all']
+        self._check_doc_files(skip_module_files, module_dir, skip_doc_files, doc_dir)
+
     def test_tops_doc_files(self):
         '''
         Ensure top modules have associated documentation
```
</details>

mchugh19:
Thanks @max-arnold! The subdirectory module check is also needed for the netapi modules, so that was helpful!

frogunder:
@mchugh19 There are some test failures here. 
If possible, it would be great if those could be looked at / fixed.

Thanks.

mchugh19:
@frogunder I'll need some help with this one. I don't think the test failure are related to this PR. 

amazon2 reports no test failures and the console reports a ton of:
`[ERROR   ] An extra return was detected from minion minion, please verify the minion, this could be a replay attack`

centos7 reports tests errors with:
```
unit.test_beacons.BeaconsTestCase.test_beacon_module
unit.test_module_names.BadTestModuleNamesTestCase.test_module_name_source_match
```
neither of which (nor the specific failures) have been touched in this PR

mchugh19:
The testsuite used to report:
>Error Message
module 'salt.beacons.log' has no attribute 'trace'
Stacktrace
  File "/tmp/kitchen/testing/tests/unit/test_beacons.py", line 39, in test_beacon_module
    ret = salt.beacons.Beacon(mock_opts, []).process(mock_opts['beacons'], mock_opts['grains'])
  File "/tmp/kitchen/testing/salt/beacons/\_\_init\_\_.py", line 71, in process
    log.trace('Beacon processing: %s', mod)
AttributeError: module 'salt.beacons.log' has no attribute 'trace'

Since there isn't a "trace" level of the standard python logging module, this is to be expected, but I'm not sure why only this PR displays the error. I've converted all `log.trace(` calls in salt/beacons/\_\_init\_\_.py to `log.debug(` to correct, but some additional eyes on this wouldn't be bad.

frogunder:
@mchugh19 Thanks for your work on this. It is on my list to follow up on :-)

waynew:
Interesting. I'm guessing that something may have incorrectly patched the beacons log, unless it's simply not being set. I'll take a look at this.

waynew:
Hrm. Still seeing that same error, grr. Test passes on its own, no problem.

tests/unit/beacons/test_log.py
```diff
@@ -6,7 +6,7 @@
 import logging
 
 # Salt libs
-import salt.beacons.log_beacon as log_beacon
+import salt.beacons.log as log_beacon
```
waynew, line 9: I think this would be part of that. Why did we rename `salt/beacons/log_beacon.py` to `log.py`?

FWIW, it's failing because https://github.com/saltstack/salt/blob/master/salt/beacons/__init__.py#L19 is defined so this attempt at import doesn't work.

waynew (changes requested):
Found the problem!

salt/beacons/__init__.py
```diff
@@ -170,19 +170,19 @@ def _process_interval(self, mod, interval):
...
-            log.trace("Interval counter: %s", counter)
+            log.debug("Interval counter: %s", counter)
             if counter * loop_interval >= interval:
                 self.interval_map[mod] = 1
                 return True
             else:
                 self.interval_map[mod] += 1
         else:
-            log.trace("Interval process inserting mod: %s", mod)
+            log.debug("Interval process inserting mod: %s", mod)
```
waynew, line 185: We should probably change all of these back to `trace`

salt/modules/linux_shadow.py
```diff
@@ -37,9 +37,11 @@
 except ImportError:
     HAS_CRYPT = False
 
+__virtualname__ = "shadow"
```
waynew, line 40: Should we also be renaming this file?

waynew: And I guess the `linux_service.py`

tests/unit/beacons/test_log.py
mchugh19: Rename was done to match the name to what the tests expected. Great catch. Will revert and update the tests instead.

waynew: Probably the general pattern to follow - I think there were  a couple of other file renames, too.

salt/beacons/__init__.py
mchugh19: Done

salt/modules/linux_shadow.py
mchugh19: As in, is it appropriate to mv shadow to linux_shadow and service to linux_service? Yes, they have to be updated. See the previous comments for more context.

From above:
> Added a test to check each module's rst for the text ".. _virtual" which successfully detects problems with shadow and service execution module naming.
> With this additional check, the doc test is expected to fail based on the mentioned issues with shadow and service. I don't think correcting those should be part of his PR, so waiting for input from core team on next steps. I suggest merging #53296 and an equivalent for service.

(The referenced PR was since merged into this one to get the mentioned tests passing)

The issue is #49335 and also #52000. Salt docs generate index pages for virtual modules. Since service is a virtual module name, having a module also named service means that any links to the service module, instead are directed to the virtual module index page. This means that on 
https://docs.saltstack.com/en/2018.3/ref/modules/all/salt.modules.service.html you are unable to click on the service module link name in the middle, as it just loads the current page rather than the module doc. 

This was a problem for the generically named shadow and service modules. A test was added in this PR to detect modules which duplicate virtual module index names, so they had to be renamed in order to keep tests passing.

waynew: Okay, I'm pretty sure I'm down with that. If we're renaming things only *within* salt, but they're not part of the API, I think it's a great idea to name them consistently.

If it borked with the public API then we'd have to deprecate things, but if people shouldn't be importing this code themselves then I'm :100:

mchugh19:
tests are green! (pre-commit looks like a new, not yet working one)

tests/unit/beacons/test_log.py
mchugh19: log beacon changes have been reverted, and tests pass now

salt/modules/linux_shadow.py
mchugh19: As far as I know, module file names are not part of the salt api. Advice has always been to call other modules via \_\_salt\_\_['whatever'] rather than direct imports (https://docs.saltstack.com/en/latest/ref/modules/index.html#cross-calling-execution-modules).

Feel free to discuss with the everyone, but as far as I know, this should be a-ok!

mchugh19:
@waynew thoughts on getting this merged? Already seeing merge conflicts.

dwoz:
@waynew LGTM, you good on this now?

waynew (approved):

waynew:
@dwoz yeah I'm good 👍

waynew:
@mchugh19 :tada: :rocket: :book:

Issue #46946, opened 2018-04-08
2018.3.0 State 'apache_module.enable' was not found in SLS. Reason: 'apache_module' __virtual__ returned False. 


topikettunen:
### Description of Issue/Question

Similar issues:

- #26284
- #19090
- #14045

`apache_module` won't allow enabling modules. Master is running on Ubuntu 16.04 and minion on OpenSUSE Tumbleweed.

### Setup
(Please provide relevant configs and/or SLS files (Be sure to remove sensitive info).)

```
apache:
  pkg.installed:
    - pkgs:
      - apache2
      - apache2-mod_php7

mod_php7:
  apache_module.enable:
    - name: php7
```

### Steps to Reproduce Issue
(Include debug logs if possible and relevant.)

```
# salt tumble state.apply suse_apache2
tumble:
----------
          ID: apache
    Function: pkg.installed
      Result: True
     Comment: 2 targeted packages were installed/updated.
     Started: 12:45:52.445081
    Duration: 37336.818 ms
     Changes:   
              ----------
              apache2:
                  ----------
                  new:
                      2.4.29-7.1
                  old:
              apache2-mod_php7:
                  ----------
                  new:
                      7.2.3-3.1
                  old:
              apache2-prefork:
                  ----------
                  new:
                      2.4.29-7.1
                  old:
              gpg-pubkey:
                  ----------
                  new:
                      a0e46e11-5911fc35
                  old:
                      1abd1afb-54176598,307e3d54-4be01a65,3dbdc284-53674dd4,a0e46e11-5911fc35
              kernel-default:
                  ----------
                  new:
                      4.15.13-1.7
                  old:
                      4.15.10-1.5,4.15.13-1.7
              kernel-default-devel:
                  ----------
                  new:
                      4.15.13-1.7
                  old:
                      4.15.10-1.5,4.15.13-1.7
              kernel-devel:
                  ----------
                  new:
                      4.15.13-1.7
                  old:
                      4.15.10-1.5,4.15.13-1.6,4.15.13-1.7
              kernel-source:
                  ----------
                  new:
                      4.15.13-1.7
                  old:
                      4.15.10-1.5,4.15.13-1.6,4.15.13-1.7
              kernel-syms:
                  ----------
                  new:
                      4.15.13-1.7
                  old:
                      4.15.13-1.6,4.15.13-1.7
----------
          ID: mod_php7
    Function: apache_module.enable
        Name: php7
      Result: False
[...]

Issue #26284, opened 2015-08-13
apache_module.enable fails on SUSE (SLES 11 SP3)


storner:
I am running Salt 2015.5.2 on a SUSE Linux Enterprise installation, and trying to use the Salt apache_module to enable Apache's proxy-module.

The state files is:

    proxyfunction:
      apache_module.enable:
        - names:
          - proxy
          - proxy_http

This fails with the following error:

              ID: proxyfunction
        Function: apache_module.enable
            Name: proxy
          Result: False
         Comment: State 'apache_module.enable' was not found in SLS 'apache2.proxy'

Searching for similar issues, I found issue #19090 which sounds similar, but no real cause was found for the problem.

It looks as if the entire "apache" module is somewhat broken on SUSE. Just trying to get the Apache version fails:

    # salt web001 apache.version
    web001:
        'apache' __virtual__ returned False

So it may be a more general problem on SUSE Linux. There is an "apache2ctl" binary installed on SUSE, though.

Version info:

               Salt: 2015.5.2
             Python: 2.6.9 (unknown, Apr  7 2015, 08:28:12)
             Jinja2: 2.7.3
           M2Crypto: 0.22
     msgpack-python: 0.4.6
       msgpack-pure: Not Installed
           pycrypto: 2.6.1
            libnacl: Not Installed
             PyYAML: 3.11
              ioflo: Not Installed
              PyZMQ: 14.5.0
               RAET: Not Installed
                ZMQ: 4.0.4
               Mako: Not Installed

cachedout:
We probably just need to add an entry for SuSE here: https://github.com/saltstack/salt/blob/develop/salt/modules/apache.py#L47

Sounds like it's look for `apachectl` instead of `apache2ctl`, which is what you say is installed. Would you like to submit a PR making this change?

rallytime:
I've fixed this in #27878. This fix will be available in the 2015.8.2 release.

Issue #19090, opened 2014-12-18
State apache_module.enable found in sls apache.sls is unavailable


chdeliens:
Hi,

When using the following state:
```yaml
Apache mod_rewrite:
  apache_module.enable:
    - name: rewrite
```

--> I got the following output "State apache_module.enable found in sls apache.sls is unavailable"

I'm using SaltStack 2014.7 on Ubuntu 14.04 LTS.

```bash
# salt --versions-report
           Salt: 2014.7.0
         Python: 2.7.6 (default, Mar 22 2014, 22:59:56)
         Jinja2: 2.7.2
       M2Crypto: 0.21.1
 msgpack-python: 0.3.0
   msgpack-pure: Not Installed
       pycrypto: 2.6.1
        libnacl: Not Installed
         PyYAML: 3.10
          ioflo: Not Installed
          PyZMQ: 14.0.1
           RAET: Not Installed
            ZMQ: 4.0.4
           Mako: Not Installed
```

As a workaround, I'm using a good old cmd.run to enable the mod_rewrite module:
```
yaml
apache mod_rewrite:
  cmd.run:
    - name: /usr/sbin/a2enmod rewrite
    - unless: "[ -f /etc/apache2/mods-enabled/rewrite.load ]"
```

garethgreenaway:
Tested this in a fresh 14.04 VM and I was unable to duplicate the error.  Do you see a similar error if you try and run the apache module (not a state call)?  Eg. salt 'hostname' apache.a2enmod rewrite

chdeliens:
Running the apache module works:

```bash
# salt 'a2enmod-test' apache.a2enmod rewrite
a2enmod-test:
    ----------
    Mod:
        rewrite
    Name:
        Apache2 Enable Mod
    Status:
        Mod rewrite enabled
```

Note that Apache is not installed on the SaltStack master, could it be the issue?

garethgreenaway:
That's a good sign, the state checks for the presence of the a2enmod function in the module to check availability.  Apache shouldn't need to be installed on the salt master.

chdeliens:
Hi, quick update on the issue. I decided to give it another try, I uncommented the state formula that wasn't working and now... it does work. No idea what changed, I always tested with a fresh Ubuntu 14.04 LTS AMI from AWS. I guess we can close this issue!

basepi:
Very strange indeed. Glad it's working, keep us posted if the problem returns!

L4rS6:
[...]

Issue #14045, opened 2014-07-09
State apache_module not working


cznewt:
Hello,
i'm trying replace cmd calls to setup apache modules with something more salt standard, but apache_module state requires module apache.a2enmod which does not exist.

am I missing something or is this a bug?

PR #55923, opened 2020-01-21
Fixing State 'apache_module' __virtual__ returned False.

gutttlt:
#46946
#26284
#19090
#14045

### What does this PR do?
It fixes "State 'apache_module.enable' was not found in SLS. Reason: 'apache_module' \_\_virtual\_\_ returned False." error that appears on SLES systems when using apache_module.enable. There is a typo: os_family grain is "Suse" but not "SUSE" on at least SLES12 and SLES15.

For sure: #46946
Probably (wasn't tested): #26284, #19090, #14045

### Tests written?
No

### Commits signed with GPG?
No

### DIFF
--- a/salt/modules/suse_apache.py
+++ b/salt/modules/suse_apache.py
@@ -23,7 +23,7 @@ def __virtual__():
     '''
     Only load the module if apache is installed.
     '''
-    if salt.utils.path.which('apache2ctl') and __grains__['os_family'] == 'SUSE':
+    if salt.utils.path.which('apache2ctl') and __grains__['os_family'] == 'Suse':
         return __virtualname__
     return (False, 'apache execution module not loaded: apache not installed.')
 


salt/modules/suse_apache.py
```diff
@@ -23,7 +23,7 @@ def __virtual__():
     '''
     Only load the module if apache is installed.
     '''
-    if salt.utils.path.which('apache2ctl') and __grains__['os_family'] == 'SUSE':
+    if salt.utils.path.which('apache2ctl') and __grains__['os_family'] == 'Suse':
```
Ch3LL: could we just ignore case, by adding lower() call there? or something similar?

Also can we get a test added here. We do run branch tests on suse 15, so we can ensure it still loads there with a test.

Ch3LL (changes requested):

salt/modules/suse_apache.py
Gutttlt: Sure, something like 
`if salt.utils.path.which('apache2ctl') and __grains__['os_family'].casefold() == 'suse':`
would also work. In this case I suppose changing all grains matching code in the module to avoid problems with case changes in the future and provide uniformity.
Should I test this and adjust the commit?

Ch3LL: yep that will work. let me know if you have any questions about tests.

Ch3LL:
Looks like this is already fixed here  https://github.com/saltstack/salt/pull/46441

Will go ahead and close this issue since we are no longer accepting PRs on 2018.3 branch and its fixed in the newer releases of Salt. Please let me know if there is a reason to re-open though and I will.

PR #55946, opened 2020-01-23
Mysql grant by column

sizgiyaev:
### What does this PR do?
The current MYSQL module can't handle grants with per column defined privileges:
More info: https://dev.mysql.com/doc/refman/8.0/en/grant.html#grant-column-privileges

### Previous Behavior
Works: `GRANT SELECT ON database.table TO user@host`
**Doesn't work**:  `GRANT SELECT(column1, column2) ON database.table TO user@host`

### New Behavior
Works: `GRANT SELECT ON database.table TO user@host`
Works: `GRANT SELECT(column1, column2) ON database.table TO user@host`

### Tests written?
Yes

### Commits signed with GPG?
No

### DIFF
--- a/salt/modules/mysql.py
+++ b/salt/modules/mysql.py
@@ -477,13 +477,25 @@ def _grant_to_tokens(grant):
     position_tracker = 1  # Skip the initial 'GRANT' word token
     database = ''
     phrase = 'grants'
-
+    column = False
+    current_grant = ''
+    
     for token in exploded_grant[position_tracker:]:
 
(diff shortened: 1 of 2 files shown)


dwoz:
@sizgiyaev This change should be against the `master` branch, not `2018.3`

sizgiyaev:
@dwoz So i need initially checkout to the new branch  from `master`?

sizgiyaev:
Closed that one and opened another PR #55980

PR #55952, opened 2020-01-23
Use vendor tornado instead of system tornado

dwoz:
### Tests written?

Yes

### Commits signed with GPG?

Yes

PR #55960, opened 2020-01-24
Refactor import of tornado

bdrung:
salt requires tornado 4 and does not work with tornado >= 5. Debian/Ubuntu ships newer versions of tornado, but renamed the old version to `tornado4`. So we changed the import statements for the Debian package.

Here I am propose to merge the two of the three commits that changes the way the tornado import is used.

### DIFF
--- a/salt/client/__init__.py
+++ b/salt/client/__init__.py
@@ -69,7 +69,7 @@ except ImportError:
 # pylint: enable=import-error
 
 # Import tornado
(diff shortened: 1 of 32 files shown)


s0undt3ch:
FYI, We are vendoring tornado for Neon.
Please see https://github.com/saltstack/salt/pull/55952

bdrung:
Then this pull request can be closed.

PR #55974, opened 2020-01-25
Consolidate some state requisites

### LINKED ISSUE #55938: Refactoring onlyif/unless - generalized with fallback?
As I was going through the release notes for Neon/3000 I came across a feature that I thought was missing, but it turned out that there were just some issues with the way a [few states](https://github.com/saltstack/salt/pull/51846#issuecomment-497913069) already had onlyif/unless implementations.

This causes some issues with the new functionality to allow states in the `onlyif/unless` declarations.

It would be great if we could figure out a clean way to refactor those so they take advantage of that new functionality, but we don't break backwards compatibility. Though it may be worth breaking that compatibility in favor of having a unified onlyif/unless flow 🤔 

In any case, the current codebase won't support the enhancements for a couple of fairly major states, such as `cmd` and `file`. 

Example:

```
do_a_thing:
  cmd.run:
    - name: echo "what?"
    - onlyif:
      - fun: file.file_exists
        path: /tmp/fnord.txt
```

Fails like this:

```
local:
----------
          ID: do_a_thing
    Function: cmd.run
        Name: echo "what?"
      Result: True
     Comment: onlyif condition is false: OrderedDict([('fun', 'file.file_exists'), ('path', '/tmp/fnord.txt')])
     Started: 16:32:53.827416
    Duration: 157.563 ms
     Changes:   
Summary for local
------------
Succeeded: 1
Failed:    0
------------
Total states run:     1
Total run time: 157.563 ms
```

Where:

``` 
do_a_thing:
  http.query:
    - name: https://correcthorsebatterystaple.waynewerner.com
    - status: 200
    - onlyif:
      - fun: file.file_exists
        path: /tmp/fnord.txt
```

works correctly.

### LINKED ISSUE #56329: Requisites 'onlyif'/'unless' ignored in some cases
### Description of Issue
I have some states with 'unless' that are running **regardless** of the unless condition exit code.

### Setup
After lot of lost time, I deduced that the problem is that in some conditions, when a salt python exception is thrown, the onlyif/unless conditions are not evaluated anymore in **next** state entries.

I succeeded in reproducing with a smaller state:
```
zsh_fail_1:
  cmd.run:
    - name: echo hey
    - runas: cristi
    - unless: test -d ~/.zinit/bin    

zsh_test_1:
  cmd.run:
    - name: echo "t1"
    - onlyif: /bin/false
```
When I used some 'test.fail_without_changes' state for 'zsh_fail_1', 'zsh_test_1' is working ok.
But if you try using it as above, the zsh_test_1 command is executed!

The important part here is that user 'cristi' does not exist, so code gets some python exception like below:
```
----------
          ID: zsh_fail_1
    Function: cmd.run
        Name: echo hey
      Result: False
                File "/usr/lib/python3/dist-packages/salt/modules/cmdmod.py", line 442, in _run
                  pwd.getpwnam(runas)
              KeyError: "getpwnam(): name not found: 'cristi'"
              
                File "/usr/lib/python3/dist-packages/salt/state.py", line 1981, in call
                  **cdata['kwargs'])
                File "/usr/lib/python3/dist-packages/salt/loader.py", line 1977, in wrapper
                  return f(*args, **kwargs)
                File "/usr/lib/python3/dist-packages/salt/states/cmd.py", line 892, in run
                  cret = mod_run_check(cmd_kwargs, onlyif, unless, creates)
                File "/usr/lib/python3/dist-packages/salt/states/cmd.py", line 366, in mod_run_check
                  cmd = __salt__['cmd.retcode'](unless, ignore_retcode=True, python_shell=True, **cmd_kwargs)
                File "/usr/lib/python3/dist-packages/salt/modules/cmdmod.py", line 2277, in retcode
                  **kwargs)
                File "/usr/lib/python3/dist-packages/salt/modules/cmdmod.py", line 445, in _run
                  'User \'{0}\' is not available'.format(runas)
              salt.exceptions.CommandExecutionError: User 'cristi' is not available
     Started: 11:52:32.097022
    Duration: 9.035 ms
     Changes:   
----------
          ID: zsh_test_1
    Function: cmd.run
        Name: echo "t1"
      Result: True
     Comment: Command "echo "t1"" run
     Started: 11:52:32.106490
    Duration: 37.838 ms
     Changes:   
              ----------
              pid:
                  30057
              retcode:
                  0
              stderr:
              stdout:
                  t1
```

### Steps to Reproduce Issue
See above.

### Versions Report
```
Salt Version:
           Salt: 3000
 
Dependency Versions:
           cffi: Not Installed
       cherrypy: Not Installed
       dateutil: 2.6.1
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
         Jinja2: 2.10
        libgit2: Not Installed
       M2Crypto: Not Installed
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.5.6
   mysql-python: Not Installed
      pycparser: Not Installed
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: Not Installed
         Python: 3.6.9 (default, Nov  7 2019, 10:44:02)
   python-gnupg: 0.4.1
         PyYAML: 3.12
          PyZMQ: 16.0.2
          smmap: Not Installed
        timelib: Not Installed
        Tornado: 4.5.3
            ZMQ: 4.2.5
 
System Versions:
           dist: Ubuntu 18.04 bionic
         locale: UTF-8
        machine: x86_64
        release: 5.4.18-1-MANJARO
         system: Linux
        version: Ubuntu 18.04 bionic
```
Strange value for the 'System Versions' 'release' field. Looks more like the kernel's name  - I'm running salt inside an ubuntu LXD container on a Manjaro distro.

mchugh19:
### What does this PR do?
Attempts to unify some of the `mod_run_check` state implementations. Global requisites onlyif and unless now support use of salt modules as well as slots, and custom implementations were falling behind (especially the commonly used cmd.run state).

As this adjusts some specific state behavior, review is encouraged. 

- states.cmd
**Done**
Included unique `creates` feature. Shorthand for unless: `test -f <filename>` or list of names. 
PR implements `creates` as a top level state requisite and remove special mod_run_check from cmd. (Left undocumented as the few states using it have mention it individually)

- states.docker_container
**Done**
Also supported unless, onlyif, and creates.
Removes special mod_run_check in favor of global creates.

- states.git
**Done**
I might have missed something, but this didn't appear to implement anything other than unless and onlyif. Removed special mod_run_check.

- states.macpackage
~Allows passing user runas to onlyif and unless command. Kept the unique implementation.~
See comment https://github.com/saltstack/salt/pull/55974#issuecomment-583897651
I think the macpackage implementation is broken and should be removed along with the local references to unless, onlyif, and user.

- states.file
~Runs command specified in `check_cmd` but also allows for additional `tmp_dir` and `tmp_ext`. Kept the unique implementation.~
Nope, file just has a mod_run_check_**cmd**

Fixes: #55938
https://github.com/saltstack/salt/pull/51846#issuecomment-497913069
Fixes: #56329

### Previous Behavior
Unique implementations of onlyif and unless act differently for different states

### New Behavior
cmd, docker_container, and git state behavior is now more consistent. 

### Tests written?
Yes - with high likelihood others need adjustment

### Commits signed with GPG?
No

@waynew @max-arnold thoughts?


max-arnold:
Neat!

I will review the refactored implementation more thoroughly in a couple of days.

* Am I correct in that `unless/onlyif` should work as expected (support module calls) with `file` state too?
* Could/should this change be included in Neon?

max-arnold:
* I'm not 100% sure, but it might be necessary to add the `creates` keyword to the following list: https://github.com/saltstack/salt/blob/master/salt/state.py#L92-L127
* Maybe it is worth documenting it as a global state argument (to help in use=cases like this: https://github.com/saltstack/salt/pull/53320 https://github.com/saltstack/salt/pull/53425)

mchugh19:
I left the file and macpackage custom implementations, as they looked somewhat state specific. If it was decided to aim for the full superset, those features could be ported too.

max-arnold:
@mchugh19 

I finally had some time to review/test the patch, and so far, it looks quite good! There are minor differences (for example, how exceptions in `states.git.mod_run_check` were handled), but I believe there is no sense to port them to the global implementation.

I still think it makes sense to add the `creates` keyword to the `STATE_RUNTIME_KEYWORDS` list and document it as a global requisite: https://github.com/saltstack/salt/blob/master/salt/state.py#L92-L127.  Otherwise, it won't be filtered here https://github.com/saltstack/salt/blob/master/salt/utils/args.py#L467-L485 and will be passed down to any state that accepts `**kwargs`.

> I left the file and macpackage custom implementations, as they looked somewhat state specific. If it was decided to aim for the full superset, those features could be ported too.

I believe that `states.file` already supports global unless/onlyif options (it has a `mod_run_check_cmd` implementation and no `mod_run_check` one)

As for `macpackage`, the code seems to be quite similar, so it might be possible to drop the module-specific implementation if you have the energy to do so.

And finally, it makes sense to mention this in the release notes, because there is a small risk that the change is not 100% compatible.

mchugh19:
Thanks for the review @max-arnold! I think I've taken care of the STATE_RUNTIME_KEYWORDS adding and some of the needed doc updates. 

You're right on the file state. So that's an easy one off the list. 

However, while the macpackage state's `mod_run_check` is simplistic, the fanciness is in that it accepts as user argument under which to run the `only_if` and `unless` requisites. 
https://github.com/saltstack/salt/blob/01b9405b61cd416d4c852c87bd484759f5ec9c96/salt/states/macpackage.py#L109

Since the global `only_if` and `unless` can do things other than just run shell commands, I'm not sure if it would make the most sense to attempt to port this to the global version, as it may introduce some confusion. 

Thoughts, and did I catch all the needed docs?

max-arnold:
Found an interesting cheatsheet that states the following:

> This logic is specific to the cmd state and may not apply to others in the same way or at all since salt.states.cmd implements its own onlyif and unless requisites.

https://gist.github.com/jfindlay/331a4394281e778a1ae9217edb5728b7

Not sure if the statement is true, but maybe it worth to double check the `cmd.run` invariants by manually running the linked state twice (before and after the patch is applied).

mchugh19:
Nice! With the patch, those states still execute as described. (Seems like the expected behavior anyway)

Currently this still leaves macpackage as the only state with specialized unless/onlyif and missing out on other requisite capabilities.

max-arnold:
> Currently this still leaves macpackage as the only state with specialized unless/onlyif and missing out on other requisite capabilities.

If you really want to fix the `macpackage` state, it could be done in the opposite way - by calling these advanced unless/onlyif features from `macpackage._mod_run_check`. 

Not sure if it is worth the trouble, though. And it is still possible that someone will introduce new state-specific `mod_run_check` functions in the future.  Unless you remove this possibilty altogether 😎

mchugh19:
Zowie! On further investigation, it appears macpackage doesn't use its local version at all. 

Here we go:
The macpackage function name is `_mod_run_check` with a leading underscore. https://github.com/saltstack/salt/blob/01b9405b61cd416d4c852c87bd484759f5ec9c96/salt/states/macpackage.py#L248

As such, the state specific version doesn't cause state.py to not run its own global versions, since it doesn't match the name `mod_run_check` https://github.com/saltstack/salt/pull/55974/files#diff-e5cf22153489a6c40864babf3ca18002R1974
and since this global requisite check occurs before the macpackage state is actually executed, it seems any `onlyif` or `unless` checks will have already occured. Should they meet whatever conditions are being tested, then the state will never be run and the local `_mod_run_check` will never execute. 

The only way I foresee the local version ever functioning is in the very special circumstance where `user` is specified on the state, but the `unless` (`onlyif` by nature would be even more difficult to work this way) check command is done in such a way where it fails if run by the global implementation, but later succeeds when run by the user specified in the state. 

This seems a bit far fetched, and as it looks like there aren't any tests for the macpackage state's _mod_run_check, I suspect it has simply never been called and the user functionality is just broken.

Can anyone from the MacOS working group confirm? If unused, we might as well remove the confusing code and documentation.

For this PR, I think that wraps up the outstanding bits.

weswhet:
I can take a look at the macpackage state tomorrow. cc @sheagcraig

weswhet:
@mchugh19  per ```The only way I foresee the local version ever functioning is in the very special circumstance where user is specified on the state,```

macOS packages can only be installed as root and attempting to do so as a user will fail. So with that being said, I'm okay with removing the code and the documentation for the macpackage state.

weswhet:
Actually, after discussion with the macOS workgroup, the change to macpackage state would break this formula. https://github.com/saltstack-formulas/salt-formula/blob/efc5d7c632834fa7e14f1c2410645a4a6594df6f/salt/minion.sls#L69-L70

@garethgreenaway recommends adding a deprecation notice.

mchugh19:
I think that should still work just fine. That code will still work, and in my testing, currently does work through the global unless requisite. Due to the function name in the macpackage file, the local unless is never actually used.

max-arnold:
Copying from Slack so it won't be lost:

> marnold
> I just watched the part about dropping the macpackage-specific unless/onlyif implementation in favor of the global one https://youtu.be/AP4Oevoqp24?t=1300
> I reviewed the PR and I believe it shouldn't affect any existing formulas (hence no need to issue any deprecation warnings), because the global unless/onlyif code basically does the same (plus supports the new way of using module calls) and in fact the local implementation is just ignored.  See the PR for more details https://github.com/saltstack/salt/pull/55974#issuecomment-587628472
> Clould you take a closer look at the code (and tests) and clarify if the deprecation process is really required?

> Christan McHugh
> Agreed. As far as I can tell, the macpackage unless code did the same as the global unless code in that it checks the output of shell commands. (The discussed feature to use salt modules in unless is a new feature in 3000, and is added functionality to the shell exit code check)
> Most importantly I don't see how the macpackage function could ever be called. Any unless statements are being rendered by the global unless handling, because the macpackage function was not named in a way which turned off the global version.
> The macpackage requisite appears to be dead code and can simply be removed.

> This can even be tested by using the new module handling with unless in a macpackage state. You will find that it works as expected, which would fail if the local macpackage version was used.

> Gareth J. Greenaway

> Yeah.  I'm confident that's the case but I'd personally rather error on the side of caution and do a proper deprecation.  Even if people aren't using in formulas they might be using it in their state files.

> wes
> I think the key point is they are saying the current implementation doesn’t work at all anyway. 

> Christan McHugh
> Right. If you use the macpackage state today, and specify an unless statement using the new module syntax, it works. If the module was evaluating anything it shouldn't. By every metric I can think of, that's just dead code. If any states use that syntax today, it was already using the global unless requisite, so there isn't any harm removing the local function.

> Gareth J. Greenaway
> Oh. If it doesn't work at all. That's different. :slightly_smiling_face:

> Gareth J. Greenaway
> If I remember right it was in a PR that included other changes?

mchugh19:
re-run pr-ubuntu2004-py3
re-run pr-fedora32-py3

garethgreenaway (approved):

Issue #34369, opened 2016-06-29
Feature request: Execution module with function(s) to clear files cached from the salt fileserver


terminalmage:
Currently, the minion fileserver cache is never purged. Adding functionality to control the minion fileserver cache would have a couple benefits:

* Reclaim space on the minion
* If a file with a password in it is cached to the minion, and later removed from the salt fileserver, its counterpart cached to the minion will remain there.

terminalmage:
ZD-702

meggiebot:
@terminalmage how big is this feature add?

UtahDave:
I think some of this may already be done. I think there may be a runner function that clears the minion cache

terminalmage:
@UtahDave To my knowledge that only clears grains/pillar/mine data. There is a ``saltutil.clear_cache`` function already, but it blows away all cached files in the minion cachedir.

The likely solution for this is a modification of this function that allows one to limit the function to just the fileserver cache. I might also add additional arguments to clean only files with a filename matching a pattern, or of a certain age, so that people can schedule a cache cleanup via cron to purge files older than N days.

@meggiebot this shouldn't take more than a day or two (famous last words), just haven't gotten around to it yet.

Oloremo:
This is a very much needed feature.

max-arnold:
See https://github.com/saltstack/salt/pull/51431

Oloremo:
#51431 will remove everything in cache dir so it's far from perfect. I don't want to remove my custom modules\states etc.

Oloremo:
I'm working on some archive.extracted state enchantments and cache control is the problem that I run on over and over.  I'd really like to have an option to "keep N last releases" for archive.extracted case. 

I understand that the cache is not really related to archive.extracted but they're interconnected a lot.

PR #55975, opened 2020-01-25
Fixed the skip_files_list_verify in archive.extracted state

Oloremo:
### What does this PR do?
It fixes the `skip_files_list_verify` logic in `archive.extracted`state for the case if `keep_source` is set to `False`.

#55443

### Previous Behavior
Before that this logic only worked if we keep the source which is not how I originally intended to implement this. See #55443

### New Behavior
Now, this logic works for both cases of `keep_source`, yet there is a bunch of limitations that has to be taken into account.

### Implementation details
I don't fancy current implementation. There is a lot of magic going on in this state, especially regarding caching. 

Also, there is a huge edge case that I highlighted in the warning section of the `skip_files_list_verify` argument. To properly fix that we need to introduce some smart Minion cache controls, see https://github.com/saltstack/salt/issues/34369

### Tests written?

Yes

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/states/archive.py
+++ b/salt/states/archive.py
@@ -430,14 +430,25 @@ def extracted(
         .. versionadded:: 2016.3.0
 
     skip_files_list_verify : False
-        Set this to ``True`` if archive should be extracted if source_hash has
+        Set this to ``True`` if archive should be extracted if ``source_hash`` has
         changed but only checksums of the archive will be checked to determine if
         the extraction is required.
+
+        It will try to find a local cache of the ``source`` and check its hash agains
+        the ``source_hash``. If there is no local cache available, for example if you
+        set the ``keep_source`` to ``False``,  it will try to find a cached source hash
+        file in the Minion archives cache directory.
 
         .. note::
             The current limitation of this logic is that you have to set
             minions ``hash_type`` config option to the same one that you're going
             to pass via ``source_hash`` argument.
+
+        .. warning::
+            With this argument set to ``True`` Salt will only check for the ``source_hash``
+            agains the local hash of the ``sourse``. So if you, for example, remove extracted
+            files without clearing the Salt Minion cache next time you execute the state Salt
+            will not notice that extraction is required if the hashes are still match.
 
         .. versionadded:: 3000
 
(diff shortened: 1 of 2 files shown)


salt/states/archive.py
```diff
@@ -978,21 +989,36 @@ def extracted(name,
...
-                            salt.utils.url.redact_http_basic_auth(source_match)
-                        )
+
+        # If file was not cached we still could have a pre-existing hash file which would be
+        # generated if update_source was set to True.
+        else:
+            parsed = _urlparse(source_match)
+            # This path would be generated if this file would be cached by file.cached state
+            # We have to mimic it due to questionable logic in the _checksum_file_path function
+            expected_cached_path = salt.utils.path.join(__opts__['cachedir'],
```
Oloremo: I don't like it, but I alternative was to change the approach for this whole caching thing in the state which sounds scary for me. Right now it's a bit magical like this archive cache **extrn_files** is only exists in a single line in the state and no other states or modules aware of it. So I have to mimic that logic to generate the expected path to the file.

tests/integration/states/test_archive.py
```diff
@@ -265,10 +265,51 @@ def test_local_archive_extracted_with_skip_files_list_verify(self):
         expected_comment = ('existing source sum is the same as the expected one and '
                             'skip_files_list_verify argument was set to True. '
                             'Extraction is not needed')
+
+        # Clearing the minion cache at the start to ensure that different tests of
+        # skip_files_list_verify won't affect each other
+        self.run_function('saltutil.clear_cache')
```
Oloremo: This is required due to cache limitation. Right now any test with skip_files_list_verify=True and source_hash_update=True will create a hash file in minion cache and other tests will use it when they shouldn't.

The alternative was to pass a different source for each test which sounds worse.

Oloremo:
re-run full ubuntu1604-py2-m2crypto

Oloremo:
re-run full centos7-py2

cmcmarrow (dismissed):

Oloremo:
@cmcmarrow Mind to look at it one more time? No changes were made, just rebased to be up to date with the current master.

Oloremo:
re-run full fedora32-py3

Oloremo:
re-run full fedora32-py3

dwoz:
re-run fedora32

dwoz (approved):

PR #55982, opened 2020-01-26
Tornado requirements and change log

dwoz:
### What does this PR do?

Update Tornado requirements and changelog

### Tests written?

Yes - In a previous PR

### Commits signed with GPG?

Yes.

### DIFF
--- a/pkg/osx/req.txt
+++ b/pkg/osx/req.txt
@@ -31,5 +31,4 @@ setproctitle
 singledispatch==3.4.0.3
 smmap==0.9.0
 timelib==0.2.4
-tornado==4.5.3
 vultr==1.0.1
--- a/pkg/windows/req.txt
+++ b/pkg/windows/req.txt
@@ -36,6 +36,10 @@ setproctitle
 singledispatch==3.4.0.3
 smmap==0.9.0
 timelib==0.2.4
-tornado==4.5.3
 watchdog==0.9.0
 wheel==0.33.4
+# Requirements for Tornado 4.5.3 (vendored as salt.ext.tornado)
+backports-abc==0.5
+singledispatch==3.4.0.3
+# Required by Tornado to handle threads stuff.
+futures>=2.0; python_version < '3.0'
--- a/requirements/static/py3.7/windows.txt
+++ b/requirements/static/py3.7/windows.txt
@@ -108,7 +108,7 @@ smmap==0.9.0
 strict-rfc3339==0.7
 tempora==1.14.1           # via portend
 timelib==0.2.4
-tornado==4.5.3 ; python_version >= "3.4"
+tornado==4.5.3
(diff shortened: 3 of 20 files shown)


requirements/static/py3.7/windows.txt
```diff
@@ -108,7 +108,7 @@ smmap==0.9.0
 strict-rfc3339==0.7
 tempora==1.14.1           # via portend
 timelib==0.2.4
-tornado==4.5.3 ; python_version >= "3.4"
+tornado==4.5.3
```
Ch3LL: shouldn't tornado be removed in these windows.txt files and the darwin.txt files since tornado was removed from their `*.in` files?

Ch3LL (changes requested):

s0undt3ch (commented):
I think you should add the tornado requirements to the base.txt file since they are now a hard dependency(they were before but they were pulled from tornado).
And remove tornado from base.txt.

If `pkg/win/req.txt` or `pkg/osx/req.txt` doesn't pull from `base.txt`, we also need to pull out tornado from those and add the tornado dependencies.

Let me know if you need me to update the static requirements after three changes.

requirements/static/py3.7/windows.txt
dwoz: This is auto generated, so something weird happened. I'll look into it.

pkg/osx/req.txt
```diff
@@ -31,5 +31,4 @@ setproctitle
 singledispatch==3.4.0.3
 smmap==0.9.0
 timelib==0.2.4
-tornado==4.5.3
```
s0undt3ch: You need to add the tornado requirements here too.
I don't believe either the osx or windows packages look at base.txt

pkg/windows/req.txt
```diff
@@ -36,6 +36,10 @@ setproctitle
 singledispatch==3.4.0.3
 smmap==0.9.0
 timelib==0.2.4
-tornado==4.5.3
 watchdog==0.9.0
 wheel==0.33.4
+# Requirements for Tornado 4.5.3 (vendored as salt.ext.tornado)
```
s0undt3ch: Like this, refs osx reqs.txt

s0undt3ch (changes requested):

s0undt3ch:
I know our current requirement lock process is bad and broken, I'm looking at either fixing it or getting an alternative post neon.

pkg/osx/req.txt
dwoz: The tornado requirements are already in this file for some reason.

s0undt3ch: Oh?!
Sorry, Should have checked.

s0undt3ch (approved):

Ch3LL (approved):

waynew (approved):

PR #55315, opened 2019-11-15
Fixes 'docker_container does not support ipv6 port bindings' https://…

HenriWahl:
### What does this PR do?
Allows to define IPv6 addresses in port_bindings directive of docker_container state. Use brackets as widely done: 
```
  - port_bindings: 
      - 172.16.0.1:443:443
      - '[2001:db8::1]:443:443'
```

### What issues does this PR fix or reference?
Closes https://github.com/saltstack/salt/issues/53709

### Previous Behavior
Error thrown because of too many ':' in address+ports-combo

### New Behavior
Accept IPv6-addresses in brackets by temporarily hiding the IPv6-related ':' to keep number of maximal binding elements on 3 - 1 address, 1 external and 1 internal port.

### Tests written?
Surely already exists for previously written module - this fix does not change anything big, just makes things work.

**Edit:** Yes, extended existing file tests/unit/utils/docker_test.py.

### Commits signed with GPG?

Yes

PR #55985, opened 2020-01-27
Avoid gaierror: [Errno -9] Address family for hostname not supported

HenriWahl:
### What does this PR do?

Avoids failing deployment of new Windows VM when using IPv6 or the new VM has only IP address:

```
win2019-test3:
    ----------
    Provisioning:
        VM being provisioned in parallel. PID: 18177
[WARNING ] Failed auth for mech NEGOEX (1.3.6.1.4.1.311.2.2.30): Mech Type 1.3.6.1.4.1.311.2.2.30 is not yet supported
Process Process-4:
  File "/usr/lib64/python2.7/multiprocessing/process.py", line 258, in _bootstrap
    self.run()
  File "/usr/lib64/python2.7/multiprocessing/process.py", line 114, in run
    self._target(*self._args, **self._kwargs)
  File "/usr/lib/python2.7/site-packages/salt/cloud/__init__.py", line 1253, in create
    output = self.clouds[func](vm_)
  File "/usr/lib/python2.7/site-packages/salt/cloud/clouds/opennebula.py", line 1155, in create
    ret = __utils__['cloud.bootstrap'](vm_, __opts__)
  File "/usr/lib/python2.7/site-packages/salt/utils/cloud.py", line 642, in bootstrap
    deployed = deploy_windows(**deploy_kwargs)
  File "/usr/lib/python2.7/site-packages/salt/utils/cloud.py", line 1263, in deploy_windows
    smb_conn = salt.utils.smb.get_conn(host, username, password)
  File "/usr/lib/python2.7/site-packages/salt/utils/smb.py", line 189, in get_conn
    return _get_conn_smbprotocol(host, username, password, port=port)
  File "/usr/lib/python2.7/site-packages/salt/utils/smb.py", line 172, in _get_conn_smbprotocol
    conn = SMBProto(host, username, password, port)
  File "/usr/lib/python2.7/site-packages/salt/utils/smb.py", line 57, in __init__
    addr = socket.gethostbyname(server)
gaierror: [Errno -9] Address family for hostname not supported
```

### Previous Behavior
Breaking Windows VM deployment when IPv6 is used

### New Behavior
Working Windows VM deployment when IPv6 is used

### Tests written?
Not really necessary - if input is an IP address its output will be an IP address. More robust as before.

### Commits signed with GPG?

Yes


bryceml:
This looks like it's probably fine, I would think it would be better to switch to using socket.getaddrinfo so we have full ipv6 support with hostnames too if I understand this correctly.

HenriWahl:
Good idea!
Like:
```
addr = socket.getaddrinfo(server, None, 0, 0, socket.IPPROTO_TCP)[0][4][0]

```
I tried it and it works.

bryceml:
This needs tests testing hard-coded ipv4, ipv6, a hostname that points to only ipv4, a hostname that points to only ipv6, and a dualstack hostname.  The value of addr could be mocked.

HenriWahl:
How to point to IPv4-only or IPv6-only hostnames? Therefore some external DNS will be needed, right? Won't the automatic test then be too dependent by something not under it's control?

bryceml:
> How to point to IPv4-only or IPv6-only hostnames? Therefore some external DNS will be needed, right? Won't the automatic test then be too dependent by something not under it's control?

I'll probably have to help, we're planning on throwing some addresses that point to either 127.0.0.1 or ::1 with all these conditions in /etc/hosts.  I'll have to make sure they all make it in the images properly.  I don't want to use ip6-localhost or localhost because they might be different between oses and I don't want to change any default behavior around those.  I was thinking something like dualstack.saltstack-test, ipv6.saltstack-test, etc.

Any thoughts?  /etc/hosts seem to work properly with everything we are doing.

bryceml:
@HenriWahl We also have several other parts of the salt codebase that needs to test all of these as well so it makes sense to have a shared set of names they can all use.

HenriWahl:
@bryceml using /etc/hosts with dualstack.saltstack-test and ipv6.saltstack-test sounds good. But to be honest I have no clue where to start with. I only once extended some test for saltstack - https://github.com/saltstack/salt/pull/55315/files#diff-b6789e2356ba3fce7cb32731400a5504. Still not merged though :-(.

But how to start a new test? How will /etc/hosts be manipulated?

bryceml:
https://github.com/saltstack/salt-jenkins/ are the states we use to make "golden images", you're probably best off waiting for me to add the entries to /etc/hosts via that first.

As far as the tests go, https://github.com/saltstack/salt/blob/master/tests/integration/utils/test_smb.py#L144 should work just copied and pasted basically with the 5 different types should be good after I've made the changes to /etc/hosts.  @dwoz said that if it can fetch a file over each of them, that's good enough.

HenriWahl:
OK. So mostly https://github.com/saltstack/salt/blob/master/tests/integration/utils/test_smb.py#L152 could be changed to test against more cases than 127.0.0.1, right?

bryceml:
> OK. So mostly https://github.com/saltstack/salt/blob/master/tests/integration/utils/test_smb.py#L152 could be changed to test against more cases than 127.0.0.1, right?

Yeah, as far as I can tell.

HenriWahl:
Many other tests there also try to test against 127.0.0.1 - I guess they all have to be modified?

bryceml:
> Many other tests there also try to test against 127.0.0.1 - I guess they all have to be modified?

for salt/utils/smb.py, @dwoz said for ipv6 testing, if it can fetch a file, that's good enough and that not every test needs an ipv6 version.

bryceml:
@HenriWahl btw, we can't build images until a bit after neon is released, so probably at least until February is over.

bryceml:
@HenriWahl these records should exist on ubuntu 20.04 images now, with more to come if you want to start writing tests.  You'll have to rebase/merge master to get the ubuntu 20.04 tests on your pr. fedora-32 should have them as well as soon as https://github.com/saltstack/salt/pull/56946 is merged.

dwoz (approved):

PR #55998, opened 2020-01-28
Remove redundant reqs from windows

dwoz:
### Commits signed with GPG?

Yes


twangboy (approved):

garethgreenaway (approved):

waynew (approved):

waynew:
Just for posterity: these are being removed because we've vendored the deps already.

s0undt3ch (approved):

dwoz:
> Just for posterity: these are being removed because we've vendored the deps already.

Not really, They are being removed because we already have these dependencies in this file.

Issue #56007, opened 2020-01-28
salt incompatible with msgpack >= 1.0.0rc1


Ch3LL:
### Description of Issue
When using msgpack 1.0.0rc1 I see this error:

```
Process salt.transport.zeromq.<class 'method'>._publish_daemon:
Traceback (most recent call last):
  File "/home/ch3ll/.pyenv/versions/3.6.9/lib/python3.6/multiprocessing/process.py", line 258, in _bootstrap
    self.run()
  File "/home/ch3ll/.pyenv/versions/3.6.9/lib/python3.6/multiprocessing/process.py", line 93, in run
    self._target(*self._args, **self._kwargs)
  File "/home/ch3ll/git/salt/salt/transport/zeromq.py", line 920, in _publish_daemon
    unpacked_package = salt.payload.unpackage(package)
  File "/home/ch3ll/git/salt/salt/payload.py", line 47, in unpackage
    return salt.utils.msgpack.loads(package_, use_list=True)
  File "/home/ch3ll/git/salt/salt/utils/msgpack.py", line 128, in unpackb
    return msgpack.unpackb(packed, **_sanitize_msgpack_kwargs(kwargs))
  File "msgpack/_unpacker.pyx", line 195, in msgpack._cmsgpack.unpackb
UnicodeDecodeError: 'utf-8' codec can't decode byte 0x83 in position 0: invalid start byte
```

when trying to run any commands against a minion.

### Setup
`pip install --pre --upgrade msgpack` or `python setup.py install` in salt repo.

### Steps to Reproduce Issue
1. install msgpack 1.0.0rc1
2. `salt '*' test.ping`

### Versions Report

```
Salt Version:
           Salt: 3000rc2-29-g796265c5be

Dependency Versions:
           cffi: Not Installed
       cherrypy: Not Installed
       dateutil: Not Installed
      docker-py: 4.1.0
          gitdb: Not Installed
      gitpython: Not Installed
         Jinja2: 2.10.3
        libgit2: Not Installed
       M2Crypto: Not Installed
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.6.2
   mysql-python: Not Installed
      pycparser: Not Installed
       pycrypto: 2.6.1
   pycryptodome: 3.9.4
         pygit2: Not Installed
         Python: 2.7.16 (default, Nov 21 2019, 13:15:23)
   python-gnupg: Not Installed
         PyYAML: 3.13
[...]

PR #56009, opened 2020-01-28
Do not allow msgpack 1.0.0rc1

dwoz:
This pins msgpack to lower than 1.0.0 until we can address #56007

### Tests written?

No

### Commits signed with GPG?

Yes

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -1,7 +1,5 @@
 Jinja2
-# This should be changed to msgpack-python for Packages
-# msgpack-python>0.3,!=0.5.5
-msgpack>=0.5,!=0.5.5
+msgpack>=0.5,!=0.5.5,<1.0.0
 PyYAML
 MarkupSafe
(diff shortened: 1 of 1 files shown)


s0undt3ch (approved):

garethgreenaway (approved):

waynew (approved):

twangboy (approved):

twangboy (changes requested):
Probably need to update the seperate osx and windows req files. Nope, looks like they're pinned to 0.5.6

twangboy (approved):

prometheanfire:
when can this be reverted?

s0undt3ch:
@prometheanfire, when we fix our code to support 1.0.0, but why?

prometheanfire:
because the cap prevents salt from being co-installed with other things that require msgpack.  The world does not consist of only salt.

PR #56014, opened 2020-01-29
Skip test_check_virtualname if source directory is not found

bdrung:
Running the unittest with autopkgtest against the installed version of
salt fails (log from salt 2019.2.3):

```
ERROR: test_check_virtualname (unit.test_virtualname.VirtualNameTestCase)
[CPU:0.0%|MEM:60.8%]
  File "/tmp/autopkgtest.O28A17/autopkgtest_tmp/tests/unit/test_virtualname.py", line 89, in test_check_virtualname
    for entry in os.listdir(os.path.join(CODE_DIR, 'salt/')):
FileNotFoundError: [Errno 2] No such file or directory: '/tmp/autopkgtest.O28A17/autopkgtest_tmp/salt/'

```

Therefore skip `test_check_virtualname` if the source directory `salt` is not
found.

### DIFF
--- a/tests/unit/test_virtualname.py
+++ b/tests/unit/test_virtualname.py
@@ -12,9 +12,9 @@ import os
 
 # Import Salt libs
 import salt.ext.six as six
-from tests.support.runtests import RUNTIME_VARS
 
 # Import Salt Testing libs
+from tests.support.runtests import RUNTIME_VARS
 from tests.support.unit import TestCase
 
 try:
--- a/tests/unit/test_virtualname.py
+++ b/tests/unit/test_virtualname.py
@@ -15,7 +15,7 @@ import salt.ext.six as six
 
 # Import Salt Testing libs
 from tests.support.runtests import RUNTIME_VARS
-from tests.support.unit import TestCase
+from tests.support.unit import TestCase, skipIf
 
 try:
     import importlib.util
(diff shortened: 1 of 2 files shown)


tests/unit/test_virtualname.py
```diff
@@ -81,6 +81,8 @@ def _check_modules(self, path):
                     )
         return ret
 
+    @skipIf(not os.path.isdir(os.path.join(RUNTIME_VARS.CODE_DIR, 'salt/')),
```
Akm0d: The slash should be removed from `salt/`

Akm0d (changes requested):

tests/unit/test_virtualname.py
bdrung: Done.

Akm0d (dismissed):

DmitryKuzmenko:
@bdrung you have to run the pre-commit script on your code to re-format it with isort and black we're using now. This will fix the `ci/pre-commit` build failure.
```
pip install pre-commit
pre-commit install
FILES="$(git show --pretty="" --name-only master..HEAD)"
pre-commit run isort $FILES
pre-commit run black $FILES
```
Next time you'll commit it will be ran automatically.

bdrung:
Thanks for pointing that out. I have rebased this merge request on master and did the blackening. I had to adjust the blackening commits, because black did not take the comments into account.

dwoz:
@bdrung I wonder, is it possible to fix `CODEDIR` for autopkg tests instead of skipping these?

bdrung:
`CODEDIR` should be `/usr/lib/python3/dist-packages` on Ubuntu for the installed salt package in this case.

s0undt3ch (approved):
Both of these tests should in fact be pre-commit hooks, now that we use pre-commit.
I'm good skipping them in this particular situation.

PR #56020, opened 2020-01-29
add kwargs to be passed from grains.setval to saltutil.refresh_grains

jtraub91:
### What does this PR do?
Allows kwargs to be passed thru to grains.setval

This allows the user to specify refresh_pillar=False

### DIFF
--- a/salt/modules/grains.py
+++ b/salt/modules/grains.py
@@ -211,7 +211,7 @@ def item(*args, **kwargs):
     return ret
 
 
-def setvals(grains, destructive=False):
+def setvals(grains, destructive=False, **kwargs):
     '''
     Set new grains values in the grains config file
 
@@ -305,12 +305,12 @@ def setvals(grains, destructive=False):
         )
     if not __opts__.get('local', False):
         # Refresh the grains
-        __salt__['saltutil.refresh_grains']()
+        __salt__['saltutil.refresh_grains'](**kwargs)
     # Return the grains we just set to confirm everything was OK
     return new_grains
 
 
-def setval(key, val, destructive=False):
+def setval(key, val, destructive=False, **kwargs):
     '''
     Set a grains value in the grains config file
 
@@ -331,7 +331,7 @@ def setval(key, val, destructive=False):
         salt '*' grains.setval key val
         salt '*' grains.setval key "{'sub-key': 'val', 'sub-key2': 'val2'}"
     '''
-    return setvals({key: val}, destructive)
+    return setvals({key: val}, destructive, **kwargs)
 
 
 def append(key, val, convert=False, delimiter=DEFAULT_TARGET_DELIM):


max-arnold:
How a user (who is not familiar with the code) is supposed to discover that it is possible to pass `refresh_pillar=False` to these functions?

alan-cugler:
This does not need test case since its a bug fix not a new feature request.

As for user discovery of possible `key:value` to pass, that can be added in a one liner to the docs.

waynew:
> This does not need test case since its a bug fix not a new feature request.

(Un)fortunately that statement is incorrect.

Tests *are* required for bug fixes to prevent regressions. If we can't reproduce a problem with a test case, it's likely that we haven't fully understood the problem.

Adding tests that exhibit the existing problem will at least notify us if we re-introduce that problem in the future.

jtraub91:
The "problem" that's occurring is that pillar is refreshed when setting grains with `set_val`. I added the ability to pass kwargs as a "quick fix" to allow for bypassing the pillar refresh. I would argue that the bigger issue is that the default behavior of `saltutil.refresh_grains` is to also refresh the pillar. Changing this would likely break things for people, however..

sagetherage:
talking to @ch3ll if this is intended https://github.com/saltstack/salt/blob/master/salt/modules/saltutil.py#L383-L385 and the better fix would be to add refresh_pillar as a specific kwarg, and yes as @waynew stated please add a test

DmitryKuzmenko:
@jtraub91 are you able to follow up the suggestions?

DmitryKuzmenko:
@jtraub91 please let me know if I can help here.

jtraub91:
@DmitryKuzmenko I can re-implement this by specifying `refresh_pillar` as a specific kwarg, but I still don't think this warrants a new test case. Not sure what I would even test in that scenario.

DmitryKuzmenko:
@jtraub91 yes please if possible.
What about test, no problem with it. All you need is just to add a test to `tests/unit/modules/test_grains.py` that mocks `saltutil.refresh_grains` and checks it's called with proper `refresh_pillar` value.

jtraub91:
@DmitryKuzmenko Ok. Will do.

DmitryKuzmenko:
Thank you!

jtraub91:
Changes have been pushed. Not sure why it's saying "unknown repository" above, so I opened a new PR to resolve this. Further discussion should ensure there (https://github.com/saltstack/salt/pull/56573) and this PR can be closed.

DmitryKuzmenko:
@jtraub91 am I right we can close this now?

jtraub91:
@DmitryKuzmenko Yes and let's make sure we get https://github.com/saltstack/salt/pull/56573 merged.

DmitryKuzmenko:
I will.

PR #56021, opened 2020-01-29
/bin/sh to /bin/bash

chalbersma:
For the virtual terminal on Linux change the default from /bin/sh to /bin/bash

Commits signed with GPG? - Yes

### DIFF
--- a/salt/utils/vt.py
+++ b/salt/utils/vt.py
@@ -414,10 +414,15 @@ class Terminal(object):
             else:
                 args = []
 
+            if os.path.exists("/bin/bash"):
+                shell_path = "/bin/bash"
+            else:
+                shell_path = "/bin/sh"
+
             if self.shell and self.args:
-                self.args = ['/bin/bash', '-c', ' '.join(args)]
+                self.args = [shell_path, '-c', ' '.join(args)]
             elif self.shell:
-                self.args = ['/bin/bash']
+                self.args = [shell_path]
             else:
                 self.args = args
 
--- a/salt/utils/vt.py
+++ b/salt/utils/vt.py
@@ -420,7 +420,8 @@ class Terminal(object):
             # /bin/sh -> To Start
             # /bin/bash -> For Most Linux
             # /bin/ksh -> For AIX/HP-UX and the Like
-            # /usr/local/bin/tcsh -> For FreeBSDers
+            # /bin/tcsh -> For FreeBSDers
+            # /bin/csh -> For some generic BSD/Unix fun
             for best_shell in ["/bin/bash", "/bin/ksh", "/bin/tcsh", "/bin/csh"]:
                 if os.path.exists(best_shell):
                     shell_path = best_shell
--- a/salt/utils/vt.py
+++ b/salt/utils/vt.py
@@ -421,7 +421,7 @@ class Terminal(object):
             # /bin/bash -> For Most Linux
             # /bin/ksh -> For AIX/HP-UX and the Like
             # /usr/local/bin/tcsh -> For FreeBSDers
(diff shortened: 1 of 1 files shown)


dwoz:
@chalbersma What is the driver for this? I'm surprised to see this PR when there is no associated issue and discussion around it.

chalbersma:
I've been doing some testing and I've noticed that sometimes salt-ssh
will hang when doing a /bin/sh but the same command with /bin/bash
works seemingly flawlessly. Plus it's 2020 and shit so I was thinking
why not just bump this to bash?

Akm0d:
Not all systems have bash, I'm thinking of some routers and such.   This would work with something like `shell_path = '/bin/bash' if os.path.exists('/bin/bash') else '/bin/sh'

chalbersma:
I guess I only considered the salt-ssh usecase. Is the vt object used
by the minion also?

dmurphy18:
@chalbersma Some OS's default to ksh, older UNIX's generally

chalbersma:
Man I forgot about AIX. I'll change to an if bash elif ksh. Do you guys
want an if for `/usr/local/bin/tcsh` for the BSD world?

CRH

dmurphy18 (commented):
Looks reasonable, cannot speak for BSD

Don't forget tests for PRs

salt/utils/vt.py
```diff
@@ -414,10 +414,21 @@ def _spawn(self):
...
                 args = []
 
+            # Decide Which Shell To Use
+            shell_path = "/bin/sh"
+
+            # /bin/sh -> To Start
+            # /bin/bash -> For Most Linux
+            # /bin/ksh -> For AIX/HP-UX and the Like
+            # /usr/local/bin/tcsh -> For FreeBSDers
+            for best_shell in ["/bin/bash", "/bin/ksh", "/usr/local/bin/tcsh"]:
```
cro: tcsh is not in /usr/local/bin on FreeBSD.  It's in /bin because it is part of the default install.  However if you are looking for root's default shell it's `/bin/csh`.

cro (changes requested):

salt/utils/vt.py
chalbersma: Okay I believe I fixed that. Sorry for the delay.

twangboy (approved):

twangboy:
@chalbersma Please rebase and address conflicts

salt/utils/vt.py
```diff
@@ -414,10 +414,23 @@ def _spawn(self):
...
 
+            # Decide Which Shell To Use
+            shell_path = "/bin/sh"
+
+            # /bin/sh -> To Start
+            # /bin/bash -> For Most Linux
+            # /bin/ksh -> For AIX/HP-UX and the Like
+            # /bin/tcsh -> For FreeBSDers
+            # /bin/csh -> For some generic BSD/Unix fun
+            for best_shell in ["/bin/bash", "/bin/ksh", "/bin/tcsh", "/bin/csh"]:
```
s0undt3ch, line 425:
```python
shell = salt.utils.path.which_bin(["sh", "bash", ...etc...])
```

s0undt3ch (changes requested):

dmurphy18 (changes requested):
@chalbersma Are you still interested in this PR ?, if unresponsive the PR shall be closed after several days.
Test cases need addressing as well as resolve conflicts
Also don't see a changelog entry

Ch3LL:
Closing due to inactivity.  Please let me know if you want me to re-open, or open a new PR against the master branch including the feedback from this PR.

Issue #55911, opened 2020-01-18
3000 release annouced the deprecation of json_decode_dict is not equalivant as tojson


sjorge:
### Description of Issue
With the 3000rc dropping the deprecation of `json_decode_dict` in the next release was announced. The new way to do it is to use `tojson`.

In prepare my states and pillars before upgrading from 2018.2.3 to 3000rc2 I noticed issues when replacing `json_decode_dict` with `tojson`.

json_decode_dict would output an in the order it was received, tojson will output the keys sorted alphabetically!

### Setup
```
{% load_yaml as x %}
z: some
y: text
x: here
{% endload %}
/tmp/tojson.out:
  file.managed:
    - contents: |
        {{ x|tojson }}

/tmp/json_decode_dict.out:
  file.managed:
    - contents: |
        {{ x|json_decode_dict }}
```

### Steps to Reproduce Issue
I provided a basic test state to see it in action, notice the different order between tosjon and json_decode_dict

### Versions Report
```
Salt Version:
           Salt: 2019.2.3

Dependency Versions:
           cffi: 1.11.5
       cherrypy: 17.3.0
       dateutil: 2.7.3
      docker-py: Not Installed
          gitdb: 2.0.5
      gitpython: 2.1.11
          ioflo: Not Installed
         Jinja2: 2.10
        libgit2: Not Installed
        libnacl: 1.6.1
       M2Crypto: 0.30.1
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.5.6
   mysql-python: Not Installed
      pycparser: 2.19
       pycrypto: 3.6.6
   pycryptodome: Not Installed
         pygit2: Not Installed
         Python: 2.7.15 (default, Oct 23 2018, 17:11:49)
   python-gnupg: 0.4.3
         PyYAML: 3.13
          PyZMQ: 17.1.2
           RAET: Not Installed
          smmap: 0.9.0
        timelib: Not Installed
        Tornado: 5.1.1
            ZMQ: 4.2.5

System Versions:
           dist:
         locale: UTF-8
        machine: i86pc
        release: 5.11
         system: SunOS
        version: Not Installed
```

xeacott:
I was looking at the documentation and it doesn't appear to have anything indicating that the output between those 2 flags are different, just that `tojson` is suppose to replace it. Is having them sorted unwanted behavior?

Here's the docs page for reference too... https://jinja.palletsprojects.com/en/2.10.x/templates/#tojson

[...]

PR #56022, opened 2020-01-29
Update `tojson` to allow arbitrary args

### LINKED ISSUE #56012: No way to sort keys (or pass extra args) on `tojson` in Python3/`json_encode_dict` doesn't produce valid JSON
# Context

https://github.com/saltstack/salt/issues/55911 raised an issue that `tojson`, the recommended replacement for `json_decode_dict` changed to sort keys, which was undesired behavior. As it turns out, `json_decode_dict` was misnamed - it should have been `json_encode_dict`. That functionality works for their desired use-case.

# Problem 1 - bytes aren't JSON

In Python2, just using `json_encode_dict` works as expected, because text and bytes are the same thing.

In Python3, text and bytes are different, so `json_encode_dict({'z': 'zaz', 'y': 'yay', 'x': 'xax'})` produces: `{b'z': b'zaz', b'y':b'yay': b'x': b'xax'}`, which is not valid JSON. This is probably a bug.

# Problem (feature?) 2 - `tojson` sorts

The other part of this issue is that while `json.dump(s)` provides a `sort_keys` argument, the underlying Jinja2 implementation of `tojson` fails to expose the ability to pass arbitrary kwargs to the underlying JSON dump, so at least `tojson(sort_keys=False)` could be used to mimic the old behavior of `json_encode_dict`. We default to the Jinja2 [tojson](https://jinja.palletsprojects.com/en/2.10.x/templates/#tojson) function, but fall back to our own internal implementation. It may be, in order to preserve feature parity, that we should go ahead and override the builtin `tojson` always, with our own implementation that allows passing arbitrary args.

The implementation would actually be *super simple*. We [already have the code](https://github.com/saltstack/salt/blob/c6871f7355790bc3ddeb135c7a48700efb77e3b0/salt/utils/jinja.py#L337-L354), all we would have to do is change the function signature from

```
def tojson(val, indent=None):
```

to 

```
def tojson(val, indent=None, **options):
```

Then we would change https://github.com/saltstack/salt/blob/c6871f7355790bc3ddeb135c7a48700efb77e3b0/salt/utils/jinja.py#L344

to 

```
options.setdefault('ensure_ascii', True)
```

That's all that would be require to change our filter to provide the capability to do

```
{{ {'z': 'zzz', 'y': 'yyy', 'x': 'xxx'}|to_json(sort_keys=False) }}
```

Though, we'll have to also change things to always use our filter.

waynew:
### What does this PR do?

The [builtin Jinja tojson filter](https://jinja.palletsprojects.com/en/2.11.x/templates/#tojson) only supports `indent` as an arg that can be passed to the [underlying json dump function](https://docs.python.org/3/library/json.html#json.dumps). This doesn't allow [all the functionality people might want](https://github.com/saltstack/salt/issues/55911).

This updates the `tojson` filter to allow arbitrary keyword arguments to be passed along to the underlying JSON library.

#55911 and closes #56012 

### Previous Behavior

It was impossible to get unsorted JSON, or use other functionality.

### New Behavior

It's possible to specify sorted or unsorted JSON, as well as pass arbitrary arguments to the underlying JSON dump function.

### Tests written?

Yes

### Commits signed with GPG?

Yes


cmcmarrow (dismissed):

DmitryKuzmenko (dismissed):

DmitryKuzmenko:
@waynew it looks failures are related.

dwoz (dismissed):

DmitryKuzmenko (dismissed):

waynew:
rebased + py2 drop precommit. Should be good *now* (or in a couple hours when all tests are done)

Ch3LL:
ping @waynew theres some failing tests here

sagetherage:
moving this to another release

waynew:
@sagetherage  hopefully this should pass the broken tests -- I  fell victim to one of the classic blunders - no, not "Never get involved in a land war in Asia,"... and no, not "Never go up against a Sicilian when death is on the line"  This one was, "Earlier versions of Python are arbitrary with their dictionary ordering no matter how lucky you think you are."

Anyway, this should be better now, please let me know if it isn't!

Ch3LL (approved):

garethgreenaway (approved):

s0undt3ch (approved):

myii:
@waynew Note that the `tojson` section in https://github.com/saltstack/salt/blob/master/doc/topics/jinja/index.rst also needs to be updated.

PR #56034, opened 2020-01-30
Extend bond hashing algorithms with encaps

mitroko:
### What does this PR do?
This PR adds encap hashing algorithms to bond valid opts

According to https://www.kernel.org/doc/Documentation/networking/bonding.txt encap options are also valid. This PR will remove limitations

### Tests written?
No

### Commits signed with GPG?
No


dwoz:

PR #56036, opened 2020-01-31
singledispatch only needed for < 3.4

dwoz:
### Commits signed with GPG?

Yes

### DIFF
--- a/requirements/static/py3.5/linux.txt
+++ b/requirements/static/py3.5/linux.txt
@@ -105,8 +105,7 @@ salttesting==2017.6.1
 scp==0.13.2               # via junos-eznc
 setproctitle==1.1.10
 setuptools-scm==3.2.0
-singledispatch==3.4.0.3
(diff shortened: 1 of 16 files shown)


requirements/static/py3.5/linux.txt
```diff
@@ -105,8 +105,7 @@ salttesting==2017.6.1
 scp==0.13.2               # via junos-eznc
 setproctitle==1.1.10
 setuptools-scm==3.2.0
-singledispatch==3.4.0.3
```
waynew: The title of this PR says it's not needed for < 3.4 - is that pointing the wrong way?

dwoz: Yeah, the commit message was wrong. Fixed.

waynew (approved):

dmurphy18 (changes requested):
backports_abc need the change too ?, I believe it does

dwoz:
> backports_abc need the change too ?, I believe it does

Yes, fixed now.

s0undt3ch (approved):

dmurphy18 (approved):

PR #56037, opened 2020-01-31
Add example for jinja template inheritance

inthecloud247:
### What does this PR do?
Adds an example showing how jinja template inheritance can work in practice. My particular example shows how template inheritance can help encourage code reuse and reduce duplication of pillar variable import commands.

It fixes a gap in saltstack documentation where it doesn't give a working example of jinja template inheritance.

### Commits signed with GPG?

Yes

### DIFF
--- a/doc/topics/jinja/index.rst
+++ b/doc/topics/jinja/index.rst
@@ -148,8 +148,40 @@ breaks, using `whitespace control`_.
 Template Inheritance
 ====================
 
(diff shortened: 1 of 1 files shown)


inthecloud247:
`19:15:13  nohup: failed to run command 'sh': Permission denied` looks like some issue with the `ci/py3` tests not running properly.

PR #56045, opened 2020-01-31
Minion maintenance process

sbrennan4:
### What does this PR do?

This PR replaces the `grains_refresh_every` mechanism with a new sub-process that will manage the cache independently of the main salt-minion process. This process has been modeled after the maintenance process that runs on the master and could be extended with additional features if desired.

### Previous Behavior

Refreshing grains on the minion was a blocking process. This was causing sporadic timeouts for us when running jobs on some minions if that job initiated a grain refresh, particularly on Solaris minions.

### New Behavior

By turning on `grains_cache` and setting `grains_refresh_every` to a value less than `grains_cache_expiration`, the grains will be refreshed by a non-blocking process on a periodic basis. This way, the minion will never need to refresh its grains as part of a job initiated on the minion, which will prevent the timeouts seen as a result.

### Tests written?

Yes

### Commits signed with GPG?

No

### DIFF
--- a/conf/minion
+++ b/conf/minion
@@ -341,17 +341,13 @@
 # to ``True``.
 #grains_deep_merge: False
 
-# The grains_refresh_every setting allows for a minion to periodically check
-# its grains to see if they have changed and, if so, to inform the master
-# of the new grains. This operation is moderately expensive, therefore
-# care should be taken not to set this value too low.
+# The grains_refresh_every setting allows for a minion to periodically
+# refresh its grains cache. Will have no effect if 'grains_cache' is not enabled.
 #
 # Note: This value is expressed in __minutes__!
 #
-# A value of 10 minutes is a reasonable default.
-#
 # If the value is set to zero, this check is disabled.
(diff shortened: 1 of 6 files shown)


thatch45:
My concern here is making the minion run an additional process, this will double the latent memory use of the minion. We have avoided running multiple procs on the minion and when we have it has created serious problems for many deployments.
I would strongly discourage running an additional process like this on the minion

sbrennan4:
@thatch45 thanks for your feedback. To avoid running a permanent process, how would you feel about adding the grains refresh to `self.periodic_callbacks` as a `MultiprocessingProcess`?

thatch45:
I think that would be a better path. We might want to make it configurable

thatch45:
I also think it should be configurable because most installs don't need this if there are specific grains that are taking too long to generate that should be considered a bug in the grain itself

sbrennan4:
https://github.com/saltstack/salt/pull/56130

PR #56046, opened 2020-01-31
Update kubernetesmod.py to use V1Deployment/V1DeploymentSpec

jnschaeffer:
### What does this PR do?

Deployment and DeploymentSpec have been available in the apps/v1 API since Kubernetes v1.9, and were removed from apps/v1beta1 in Kubernetes v1.16. This commit updates the kubernetesmod module to use those instead of the apps/v1beta1 Deployment and DeploymentSpec objects.

Additionally, V1DeploymentSpec also requires a selector object during construction, so that has been added, and the default empty string value of template on construction has been removed.

### Previous Behavior
For Kubernetes clusters of version v1.16 or higher, Deployments would fail to be created.

### New Behavior
For Kubernetes clusters of version v1.9 or higher, Deployments will be created.

### Tests written?
No

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/modules/kubernetesmod.py
+++ b/salt/modules/kubernetesmod.py
@@ -72,15 +72,7 @@ try:
     import kubernetes.client
     from kubernetes.client.rest import ApiException
     from urllib3.exceptions import HTTPError
-    # pylint: disable=no-name-in-module
-    try:
-        # There is an API change in Kubernetes >= 2.0.0.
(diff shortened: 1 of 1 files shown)


cmcmarrow:
Looks like you have some conflicts now. Also I still don’t see tests. If you need help with test writing me please message me.

cmcmarrow:
Also you need to setup pre-commit. You do this by running `pip install  pre-commit`. This should also fix your lint errors.

dwoz:

PR #56047, opened 2020-02-01
Move the `salt-unity` implementation to `salt.scripts`

s0undt3ch:
### What does this PR do?
See title.

With the move to setuptools there was still one distutils script `scripts/salt-unity` which wasn't implemented in `salt.scripts`, this PR addresses that.


twangboy (approved):

dwoz (approved):

dmurphy18 (approved):
My only grip is '#!/usr/bin/env python' is not the recommended way, explicit is preferred
But I want to build packages and the rest of the scripts are as bad, so approve

PR #56048, opened 2020-02-01
[WIP] Setuptools is smart enough to understand requirement markers

s0undt3ch:
### What does this PR do?
See title

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -115,78 +115,6 @@ exec(compile(open(SALT_VERSION).read(), SALT_VERSION, 'exec'))
 
 # ----- Helper Functions -------------------------------------------------------------------------------------------->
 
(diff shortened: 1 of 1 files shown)


dmurphy18 (commented):
This looks good, but are there version limitations with setuptools for parsing line, for example: using v10.2.1 of setuptools, otherwise usually v33+ or v36+

s0undt3ch:
@dmurphy18 good call. setuptools 10.2.1 doesn't know how to parse these requirements

s0undt3ch:
This change will bump the setuptools requirement to `>=20` when doing a `python setup.py install`

s0undt3ch:
Added `WIP` since this setuptools minimum version bump is most likely not desirable.

PR #56060, opened 2020-02-04
Fix regex string for **Del and **DelVals

### LINKED ISSUE #56062: LGPO Module doesn't handle **Del and **DelVals properly
### Description of Issue
The LGPO module will fail to enable some policies if they are set to `disabled`.

### Setup
Set the following Group Policy Setting to `Disabled`:
```
Computer Policy->Administrative Templates->Windows Components->Windows Error Reporting->Configure Default consent
```

### Steps to Reproduce Issue
Run the following state to enable the policy:
```
ERRORREP_config_default_consent:
  lgpo.set:
    - computer_policy:
        "Configure Default consent":
            "Consent level": 'Always ask before sending data'
```
The following will always be the result. The actual value never changes, even though there are no errors:
```
C:\Users\vagrant>salt-call --local state.apply test
[ERROR   ] policy WerDefaultConsent_1 was found but it does not appear to be valid for the class Machine
local:
----------
          ID: ERRORREP_config_default_consent
    Function: lgpo.set
      Result: True
     Comment: The following policies are in the correct state:
              Configure Default consent
     Started: 01:12:24.828000
    Duration: 2062.0 ms
     Changes:
Summary for local
------------
Succeeded: 1
Failed:    0
------------
Total states run:     1
Total run time:   2.062 s
```
### Versions Report
Salt 3000

twangboy:
### What does this PR do?
The regex string that updates the Registry.pol file wasn't properly matching instances where the reg_key was prepended with **Del or **DelVals due to the utf-16-le encoding of the Registry.pol file.

This adds the null byte characters to the regex portion that searches for those values.

### Related issue
https://github.com/saltstack/salt/issues/56062

### Previous Behavior
The example I was working with was the `Default Consent` key which renders in the Registry.pol file like this: "`D e f a u l t C o n s e n t`". Where the spaces between the characters are null bytes. (utf-16-le)

When this policy is disabled there is some additional text prepended, like this: "`* * d e l . D e f a u l t C o n s e n t`"

The original regex did not include the null byte characters and therefore could not find the policy that needed to be changed, so it appended the new setting to the file. This caused there to be two conflicting entries in the file for the same setting.

### New Behavior
Null byte characters are accounted for and you can now set this policy.

### Tests written?
Yes

### Commits signed with GPG?
Yes


twangboy:
@lomeroe Would you mind taking a look at this?

lomeroe (approved):
LGTM

Ch3LL (approved):

waynew (approved):

dwoz (approved):

lorengordon:
Heh, thanks for this @twangboy! I was just troubleshooting the same issue, but in the `_policyFileReplaceOrAppend` and `_policyFileReplaceOrAppendList` functions. My solution was a little more... violent. I just redefined the function in my own custom module, using pure `re` functionality...

```
def _policyFileReplaceOrAppend(policy, policy_data, append=True):
    '''
    helper function to take a ADMX policy string for registry.pol file data and
    update existing string or append the string to the data

    Cut from win_lgpo.py due to bugs when there are DELETE policies
    '''
    # token to match policy start = encoded [
    policy_start = re.escape('['.encode('utf-16-le'))

    # token to match policy end = encoded ]
    policy_end = re.escape(']'.encode('utf-16-le'))

    # token to match policy delimiter = encoded null + encoded semicolon
    policy_delimiter = b''.join([
        chr(0).encode('utf-16-le'),
        ';'.encode('utf-16-le'),
    ])

    # pattern group to OPTIONALLY match delete instructions in value token
    policy_pattern_del = b''.join([
        '(',
        re.escape('**Del.'.encode('utf-16-le')),
        '|',
        re.escape('**DelVals.'.encode('utf-16-le')),
        '){0,1}',
    ])

    # pattern group to match one delimited token in a policy
    policy_token = b''.join([
        '(',
        '.*?',  # non-greedy match, up to next policy delimiter
        policy_delimiter,
        ')',
    ])

    # pattern to capture the key and value tokens from `policy`
    policy_pattern = b''.join([
        policy_start,
        policy_token, # this is the registry key
        policy_pattern_del,
        policy_token, # this is the value
        '.*?',        # this is the remainder of the policy tokens
        policy_end,
    ])

    # parse the tokens from `policy`
    policy_match = re.search(
        policy_pattern,
        policy,
        flags=re.IGNORECASE|re.DOTALL,
    )

    # pattern to match `policy` in `policy_data`
    policy_match_groups = policy_match.groups()
    policy_data_pattern = b''.join([
        policy_start,
        re.escape(policy_match_groups[0]),  # key
        policy_pattern_del,
        re.escape(policy_match_groups[2]),  # value
        '.*?',
        policy_end,
    ])

    # search for `policy` in `policy_data`
    policy_data_match = re.search(
        policy_data_pattern,
        policy_data,
        flags=re.IGNORECASE|re.DOTALL,
    )

    if policy_data_match:
        # replace a match with the policy
        policy_repl = policy_data_match.group()
        log.debug('replacing "%s" with "%s"', policy_repl, policy)
        return policy_data.replace(policy_repl, policy)
    elif append:
        # append the policy
        log.debug('appending "%s"', policy)
        return b''.join([policy_data, policy])
    else:
        # no match, no append, just return what we were given
        return policy_data
```

PR #56079, opened 2020-02-06
Allow bypassing requirements installattion.

s0undt3ch:
### What does this PR do?
Allow bypassing requirements installation.

This shall be considered an expert feature since you need to be aware of
what's going on.
This is useful when writing tests and running them over and over again
where on the first run the requirements get properly installed but on
subsequent runs, if no requirements are changed, the install procedure
does not need to run, thus, saving valuable seconds which can quick turn
into several minutes when repeating test runs.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -42,6 +42,7 @@ CI_RUN = (
     or os.environ.get("DRONE") is not None
 )
 PIP_INSTALL_SILENT = CI_RUN is False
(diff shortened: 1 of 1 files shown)


twangboy (dismissed):

DmitryKuzmenko (dismissed):
Must have for development! :+1:

twangboy (approved):

dwoz (approved):

PR #56095, opened 2020-02-08
Provide security advisory for PyCrypto

### LINKED ISSUE #56080: Update release notes, pip download page and install docs on pycrpto
### Description of Issue
Inform via release notes, pip download page and install docs of pycrpto issues

### Related
https://github.com/saltstack/salt/pull/56039

### Steps to Reproduce Issue

### Versions Report

waynew:
Fixes #56080

@cachedout @dmurphy18 @sagetherage @frogunder 

Does this look reasonable? Anything I should add/take away/adjust?

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -34,6 +34,14 @@ documentation.
 
 `<https://docs.saltstack.com/en/latest/>`_
 
+Security Advisory
+=================
+
+For historical reasons, Salt requires PyCrypto as a "lowest common
+denominator". However, `PyCrypto is unmaintained`_ and best practice is to
+manually upgrade to use a more maintained library such as `PyCryptodome`_. See
+`Issue #52674`_ and `Issue #54115`_ for more info
+
 Engage SaltStack
 ================
 
@@ -66,3 +74,7 @@ services`_ offerings.
 .. _SaltStack education offerings: http://saltstack.com/training/
 .. _SaltStack Certified Engineer (SSCE): http://saltstack.com/certification/
 .. _SaltStack professional services: http://saltstack.com/services/
+.. _PyCrypto is unmaintained: https://github.com/dlitz/pycrypto/issues/301#issue-551975699
+.. _PyCryptodome: https://pypi.org/project/pycryptodome/
+.. _Issue #52674: https://github.com/saltstack/salt/issues/52674
+.. _Issue #54115: https://github.com/saltstack/salt/issues/54115
--- a/doc/topics/hardening.rst
+++ b/doc/topics/hardening.rst
@@ -10,6 +10,20 @@ heavily on how you use Salt, where you use Salt, how your team is structured,
 where you get data from, and what kinds of access (internal and external) you
 require.
 
+.. warning::
+
+    For historical reasons, Salt requires PyCrypto as a "lowest common
+    denominator". However, `PyCrypto is unmaintained`_ and best practice is to
+    manually upgrade to use a more maintained library such as `PyCryptodome`_. See
+    `Issue #52674`_ and `Issue #54115`_ for more info
+
+
+.. _PyCrypto is unmaintained: https://github.com/dlitz/pycrypto/issues/301#issue-551975699
+.. _PyCryptodome: https://pypi.org/project/pycryptodome/
+.. _Issue #52674: https://github.com/saltstack/salt/issues/52674
+.. _Issue #54115: https://github.com/saltstack/salt/issues/54115
+
+
 General hardening tips
 ======================
 
--- a/doc/topics/installation/index.rst
+++ b/doc/topics/installation/index.rst
@@ -104,6 +104,21 @@ Salt should run on any Unix-like platform so long as the dependencies are met.
   * `PyCrypto`_ - The Python cryptography toolkit
 
 
+.. warning::
+
+    For historical reasons, Salt requires PyCrypto as a "lowest common
+    denominator". However, `PyCrypto is unmaintained`_ and best practice is to
+    manually upgrade to use a more maintained library such as `PyCryptodome`_. See
+    `Issue #52674`_ and `Issue #54115`_ for more info
+
+
+.. _PyCrypto is unmaintained: https://github.com/dlitz/pycrypto/issues/301#issue-551975699
+.. _PyCryptodome: https://pypi.org/project/pycryptodome/
+.. _Issue #52674: https://github.com/saltstack/salt/issues/52674
+.. _Issue #54115: https://github.com/saltstack/salt/issues/54115
+
+
+
 Salt defaults to the `ZeroMQ`_ transport. The ``--salt-transport`` installation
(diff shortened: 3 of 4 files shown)


sagetherage:
This looks pretty good, the pip download page is what I wrote in the ticket. I am not sure if that is a page I am familiar with unless that is the repo dot saltstack dot com. If it is that page, or another dot com page then we will put the same security wording there (once approved and merged) and review it as we stage then promote and that will be the 3. Please correct me if I am wrong @cachedout @dwoz @dmurphy18 @frogunder

waynew:
The README.rst file gets read in our setup.py, which... if I understand correctly, is what gets set in the package build that's uploaded to PyPI. At least that's how typical packages work 🙂

sagetherage:
got it, pypi = pip install page, I am all set

cachedout:
Hi @waynew. Thanks for doing this!

There are a few ways I think this could be improved:

1. Documentation [in the installation instructions](https://docs.saltstack.com/en/latest/topics/installation/index.html#dependencies).

1. A mention of this caveat in [the bootstrap instructions](https://github.com/saltstack/salt-bootstrap/blob/develop/README.rst) for cases where people are installing from git or passing the `-P` flag.

1. [Fedora instructions which use pip](https://docs.saltstack.com/en/latest/topics/installation/fedora.html#installation-using-pip).

1. [OS X instructions which use pip](https://docs.saltstack.com/en/latest/topics/installation/osx.html#installation-from-pip)

1. [RHEL instructions which use pip](https://docs.saltstack.com/en/latest/topics/installation/osx.html#installation-from-pip)

1. It may be worth mentioning this in [the developer documentation](https://docs.saltstack.com/en/latest/topics/development/hacking.html)? (Up to you.)

1. It should probably be added to the documentation which describes [how to harden salt](https://docs.saltstack.com/en/latest/topics/hardening.html).

1. There is an [old guide on packaging for Salt](https://docs.saltstack.com/en/latest/topics/development/conventions/packaging.html) that should probably include this, or the page should be updated/removed.

It's not quite clear to me whether any packages in the repo will use pycrypto out of the box. I know that @dmurphy18 mentioned a few RHEL cases where a supported crypto library is used but just so I can understand are there any packages delivered from the Salt repo which would install a pycrypto dependency and not a supported crypto lib?

Thanks again!

dmurphy18:
@cachedout Here are the following platforms which still have pycrypto installed, noting that this was the same for the previous Salt 2019.2.3 release, therefore no change in what Salt was installing.

Python 2:
Debian 8 (inc. Raspbian 8)
Debian 9 (inc. Raspbian 9)
Ubuntu 16.04
Ubuntu 18.04
RHEL 7
RHEL 6
Amazon Linux 1
Amazon Linux 2

Python 3:
Debian 9 (inc. Raspbian 9)
Ubuntu 16.04
Ubuntu 18.04
Amazon Linux 2

Python 3 RHEL 7 and RHEL 8 utilize M2Crypto
Python 3 Amazon Linux 2 provides pycryptodomex but still makes use of pycrypto, an oversight that will be fixed in next point release.

Will also adjust Python 2 & 3 Debian and Ubuntu to support minimum of pycryptodome in the next point release, where possible and time allowing.

cachedout:
> Here are the following platforms which still have pycrypto installed

Thanks! This is very helpful! I just want to understand what you mean by the above statement. For each item in the list you provided, will an out-of-the-box install of the package from the Salt repo install pycrypto and use it as the default, assuming that neither m2crypto nor pycryptodome were previously installed?

waynew:
If I understand correctly, the `python-crypto` package dependency from the vendor repos will provide pycrypto.

dmurphy18:
see https://github.com/saltstack/salt-pack/issues/724
https://github.com/saltstack/salt-pack-py3/issues/189

And yes, clean box will install pycrypto, noting that as before in Salt releases that if pycryptodomex or M2crypto is present , then they will be used, order of preference:

Lowest:. pycrypto
Medium: pycryptodomex
Highest: M2Crypto

This  is unchanged from the last few years since I ported pycryptodomex to RHEL 7.
Note that RHEL 6 32-bit does not work with pycryptodomex v3.6.3 that Salt supplies (due to 64-bit assumptions in their code), this may have  changed in later versions, recalling that SaltStack provides minimum versions of a package such that it will function with Salt, expecting that later OS versions of the package will supersede the version supplied by SaltStack.

cachedout:
> If I understand correctly, the python-crypto package dependency from the vendor repos will provide pycrypto.

Gotcha. But the end result in the same, no? That in cases for the machine listed above, stock installs of the Salt package on a clean machine result in Salt defaulting to using pycrypto during runtime. Is that correct?

dmurphy18:
@cachedout Yes, unchanged from previous releases, for the last few years

cachedout:
Thanks for the clarification.

It seems like this notice needs to be expanded to a few other places. Would it make sense to put it in the repo?

https://repo.saltstack.com/#ubuntu
https://repo.saltstack.com/#rhel
https://repo.saltstack.com/#amzn
https://repo.saltstack.com/#debian
https://repo.saltstack.com/#raspbian

@waynew What do you think about putting up changes in those locations, correlating to the specific targets that @dmurphy18 has mentioned?

dwoz (approved):

waynew:
@cachedout since each of those OS/pip install instructions come from the overall installation instructions I only added the warning to that. I've also opened a PR against bootstrap's readme to link to the hardening Salt docs.

cachedout:
> @cachedout since each of those OS/pip install instructions come from the overall installation instructions

I'm not sure I understand. The [front page of the documentation site](https://docs.saltstack.com/en/latest/) links directly to the repo under the section "Installing Salt". Why wouldn't the per-distribution installation instructions which follow include this advisory? 

install_salt

I don't recall exactly, either, but if memory serves, this page might also be where users are directed to after [filling out SaltStack's marketing collection form ](https://www.saltstack.com/download-saltstack-automation-engine/)for downloading Salt.

waynew:
https://docs.saltstack.com/en/latest/topics/installation/index.html links to the platform-specific pip instructions, e.g.

> 3. [Fedora instructions which use pip.](https://docs.saltstack.com/en/latest/topics/installation/fedora.html#installation-using-pip)

For the repo.saltstack.com, I haven't touched those docs yet, but IIUC they're in another repo, right? It would be good to update those to warn that the platform ships a broken version of PyCrypto, which we will use, but we *strongly* recommend upgrading to something like pycryptodome, for instance.

cachedout:
>  IIUC they're in another repo, right? 

Yes, that is correct. I don't remember offhand where they are but @Ch3LL might remember.

> It would be good to update those to warn that the platform ships a broken version of PyCrypto, which we will use, but we strongly recommend upgrading to something like pycryptodome, for instance.

Sounds good. Thanks!

Ch3LL:
yeah @waynew feel free to reach out to me or @frogunder and we can point you in the right direction

PR #56100, opened 2020-02-08
Add feature request 51919

### LINKED ISSUE #51919: Feature request: salt-cloud vmware ip handaling
Currently With the current way saltstack handles `wait_for_ip_timeout` the connection it will grab the first IP that is returned by vmware tools and hold onto it for trying to connect. 

However, in mixed environments where DHCP and static IPs are used The first IP might not be the one that salt needs. And DNS might also not cover the host in question. 

What would be good is if there was another option that says if we fail to connect to the first address we got from VMware tools query VMware tools again for the IP. Or an option to only accept the IP if it matches a list/cidr or is in the list of IPs set in the network devices.

alfotos:
### What does this PR do?
Implement new feature in the VMware cloud module to give the option to wait for a pre-defined static IP when creating a new VM. The new parameter "wait_for_ip_priority" can be set to "any", "dhcp", "static" or "ip" to control the behaviour of the "_wait_for_ip" function.

Fixes #51919

### Previous Behavior
At the moment, the _wait_for_ip uses the first IP configured in the VM during the deployment. This can be an issue if you are deploying a VM in a DHCP-enabled subnet but you want to use a given static IP. This can lead to a timeout later, as Salt will try to connect to an IP which was only used fore few seconds (time between the DHCP IP and the static IP configured by VMware tools). More details are explained in the issue #51919

### New Behavior
The current behaviour is maintained by default, or by setting the "wait_for_ip_priority" to "dhcp" or "any". On the other hand, if you have a static IP and want to be sure that one is used the parameter can be set to "static", this will filter the registered IP and will only accept one of the IPs defined in the network devices. The specific use case of this is "ip", when you also have to use the parameter "wait_for_ip_address".

### Tests written?
No.

### Commits signed with GPG?
No


Ch3LL:
looks like there is a mishap here as there are a ton of commits. can you re-apply your fix on the head of master and re-push? 

For tests it would be good to take a look at the tests/unit/cloud/clouds/test_vmware.py file for inspiration.

cmcmarrow (changes requested):
Hey @alfotos. I would just make a new branch from master and put your changes into it.

alfotos:
Hi @cmcmarrow . I am closing this PR as I have just created a new on for this #56701 .

PR #56125, opened 2020-02-11
Add Astra Linux Common Edition to the OS Family list

juliogonzalez:
### What does this PR do?

It adds [Astra Linux Common Edition](https://astralinux.ru/en/products/astra-linux-common-edition/) to the OS Family list, so it's recognized by salt.

Astra Linux is a Russian distribution based on Debian, currently based on Debian 9.0:
```
$ cat /etc/debian_version 
9.0
```
```
$ cat /etc/os-release 
PRETTY_NAME="Astra Linux (Orel 2.12.22)"
NAME="Astra Linux (Orel)"
ID=astra
ID_LIKE=debian
ANSI_COLOR="1;31"
HOME_URL="http://astralinux.ru"
SUPPORT_URL="http://astralinux.ru/support"
VARIANT_ID=orel
VARIANT=Orel
VERSION_ID=2.12.22
```

None. I discovered this problem because I am trying to add support for this distribution at [Uyuni](https://www.uyuni-project.org/).

### Previous Behavior

`os_family` grain value was `AstraLinuxCE`, so the distribution was not recognized as Debian derivative and things like installing packages failed:
```
# salt-call --local pkg.install htop
'pkg.install' is not available.
```
### New Behavior

`os_family` grain value is now `Debian`, so the distribution is recognized and it can be used, and salt can be used -for example- to install packages:
```
# salt-call --local pkg.install htop
local:
    ----------
    htop:
        ----------
        new:
            2.0.2-1
        old:
```

### Tests written?

Yes, one new unit test.

### Commits signed with GPG?

Yes


salt/grains/core.py
```diff
@@ -1513,6 +1513,7 @@ def id_():
     'Funtoo': 'Gentoo',
     'AIX': 'AIX',
     'TurnKey': 'Debian',
+    'AstraLinuxCE': 'Debian',
```
juliogonzalez: I couldn't really understand if the list has any order, so I hope adding it at the end will be fine. Let me know otherwise.

waynew: Doesn't look like there's any order. I would have alphabetized it :joy: but I definitely don't expect you to.

waynew:
@juliogonzalez thanks for the PR!

> As far as I can see, not all distributions have tests at `tests/unit/grains/test_core.py`, but if you want me to add them anyway, please let me know.

Especially if there already tests for other mappings, please do go ahead and add tests for this one. It may be simple, but it should help prevent someone accidentally removing/editing this one.

tests/unit/grains/test_core.py
```diff
@@ -605,6 +605,26 @@ def test_ubuntu_artful_os_grains(self):
         }
         self._run_os_grains_tests("ubuntu-17.10", _os_release_map, expectation)
 
+    @skipIf(not salt.utils.platform.is_linux(), 'System is not Linux')
+    def test_astralinuxce_2_os_grains(self):
```
juliogonzalez: The value of `osrelease` is `2.12.22`, but I don't think we should use it for the testname.

juliogonzalez:
Not 100% sure, but I'd say the failures are not really related to the PR? :-?

tests/unit/grains/test_core.py
waynew: Seems reasonable enough for me. If we need to get more specific down the road, we can handle things then.

waynew (dismissed):
Looks great - tried running tests locally without the change and the failed, success with the change 👍

juliogonzalez:
@waynew @sagetherage, since this is waiting since Feb (when tests were passing), can we get a review and merge it if there are no issues?

I already changed the code and the tests to the latest changes at master, and tests are passing (with the exception of ci/py3/centos7/m2crypto/pytes), but I'd say this is unrelated to my changes.

sagetherage:
I am re-running the failed test @juliogonzalez can you address the merge conflict, please? thank you!

juliogonzalez:
> I am re-running the failed test @juliogonzalez can you address the merge conflict, please? thank you!

Done

sagetherage:
not sure why I applied those labels here - I think I was on the wrong tab - my mistake!

juliogonzalez:
Seems some tests are failing, but I can don't see any relationship with the PR.

sagetherage:
I updated your PR with master and let those tests run again, I will check back in a bit.

juliogonzalez:
Well, macOS is still failing... but don't think it's this PR's fault :-)

juliogonzalez:
@sagetherage  can we get this merged? :-)

sagetherage:
Please address the merge conflicts and then we will need to get more than 3 reviews

juliogonzalez:
@sagetherage I was going to do it (for the third time), but looks like somebody else added Astra: https://github.com/saltstack/salt/commit/ec63d2cff5c16b6d9528f3dbb796cb58dbd6361e

So I guess this PR can be closed now. Please go ahead and close it if you agree.

PR #56126, opened 2020-02-11
update release process docs

Ch3LL:
update release process docs

ping @frogunder can you review here please since you are more involved with this process now?


frogunder (approved):

max-arnold:
Sorry for jumping in, but I think the following (removed) part is quite important:

> Notify salt-users and salt-announce google groups when the feature freeze branch creation will occur so they can try to get their work merged.

Even if there is no separate branch, it is still important to notify contributors:

1. Before feature freeze, so people can get their PRs in before it is too late
2. Feature freeze start
3. Feature freeze end

Ch3LL:
ya thats a fair point. ping @sagetherage just want to make sure that is on the current release process to notify of feature freeze still? obviously we dont need to notify of new branch anymore, just when we plan to not allow any more features in.

sagetherage:
yes, we will get the communication msg corrected in the auto msg (and will look if there is more). we can keep that verbiage in this doc and update it to be something more like:
"Notify salt-users and salt-announce google groups when the start of the freeze will occur so they can try to get their work merged, and when it will end and merging to continue."  Open to suggestions though.

PR #56145, opened 2020-02-12
The service.status function doesn't accept kwargs so don't pass them

blarghmatey:
### What does this PR do?
Removes **kwargs from calls to the `service.status` function

N/A

### Previous Behavior
There are a few places in the service execution module where `**kwargs` are getting passed to calls for `service.status`, but that function signature doesn't accept arbitrary kwargs which results in exceptions being thrown. This removes those kwargs from the call to service.status since they are not useful anyway.

### New Behavior
Calls to `service.status` will not throw an exception

### Tests written?
No

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/states/service.py
+++ b/salt/states/service.py
@@ -428,7 +428,7 @@ def running(name,
 
     # lot of custom init script won't or mis implement the status
     # command, so it is just an indicator but can not be fully trusted
-    before_toggle_status = __salt__['service.status'](name, sig, **status_kwargs)
+    before_toggle_status = __salt__['service.status'](name, sig)
     if 'service.enabled' in __salt__:
         before_toggle_enable_status = __salt__['service.enabled'](name)
     else:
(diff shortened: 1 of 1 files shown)


Ch3LL:
Do you have an issue opened associated with this PR so I can see what the error an use case that you are running into? It looks like some of the service execution modules do accept more kwargs then just name and sig.

blarghmatey:
I don't have an issue opened for this but I know that it has been identified by others as well and there is an alternative PR that is not as far reaching. The short version is that the `service.status` function only accepts 2 arguments and has no place for `*kwargs`. The proper solution is to either change its function signature to accept and ignore kwargs, or to remove all calls to the function that pass kwargs which is what this PR does.

DmitryKuzmenko:
We definitely need more complicated solution just because some implementations of service, like `systemd_service`, `mac_service` and `freebsd_service` are accepting additional arguments. So we have to unify the `status` interface instead of cutting the `kwargs`.

twangboy (approved):

twangboy:
@blarghmatey This needs a rebase

Ch3LL:
I agree with @DmitryKuzmenko here. We need to allow kwargs here in order to ensure the proper kwargs are passed to the multiple service moduels. If you could open an issue with the details it would help me to see the issue you are running into and figure out a solution.

Ch3LL:
Closing due to inactivity. Please let me know if this needs to be re-opened or feel free to open a new PR with the feedback from the PR.

Issue #56091, opened 2020-02-07
salt-cloud not working with azure and unmanaged disk


tkykm:
### Description of Issue

The azurearm module fail to create a unmanaged disk.
Because it attempt to create VirtualHardDisk instance with string arugment, but it should pass it as keyword argument like VirtualHardDisk(uri='https://~~~')
https://github.com/saltstack/salt/blob/master/salt/cloud/clouds/azurearm.py#L1162

### Setup
stable 2019.2.2 python3

### Steps to Reproduce Issue
set `vhd: unmanaged` to profile or mapfile and then deploy

```
Traceback (most recent call last):
  File "/usr/lib/python3.6/site-packages/salt/cloud/__init__.py", line 67, in _call
    ret = func(*args, **kwargs)
  File "/usr/lib/python3.6/site-packages/salt/cloud/__init__.py", line 2294, in create_multiprocessing
    local_master=parallel_data['local_master']
  File "/usr/lib/python3.6/site-packages/salt/cloud/__init__.py", line 1253, in create
    output = self.clouds[func](vm_)
  File "/usr/lib/python3.6/site-packages/salt/cloud/clouds/azurearm.py", line 1404, in create
    vm_request = request_instance(vm_=vm_)
  File "/usr/lib/python3.6/site-packages/salt/cloud/clouds/azurearm.py", line 1167, in request_instance
    volume['lun'],
TypeError: __init__() takes 1 positional argument but 2 were given
```

### Versions Report
I've used this pip modules: 
https://raw.githubusercontent.com/saltstack/salt/v2019.2.2/requirements/static/py3.6/cloud.txt

Salt Version:
           Salt: 2019.2.2

Dependency Versions:
           cffi: 1.13.2
       cherrypy: Not Installed
       dateutil: 2.8.1
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.8.1
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: 0.35.2
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.5.6
   mysql-python: Not Installed
      pycparser: 2.19
       pycrypto: Not Installed
   pycryptodome: Not Installed
         pygit2: Not Installed
         Python: 3.6.8 (default, Aug  7 2019, 17:28:10)
   python-gnupg: Not Installed
         PyYAML: 3.12
[...]

PR #56155, opened 2020-02-13
fix azure vm creation, added a function-signature check for azure.request_instance

dhiltonp:
### What does this PR do?
Fixes azure vm creation with azure modules on Python3 starting around 2018?

Also adds a test that detects api changes. The test itself is kind of terrible, but I think a good part of the blame lies on the 450+ loc function. 

I've manually verified the other uses of the azure models API pass kwargs as well.

#56091 

### Previous Behavior
Code was broken on Python3 and recent API versions due to a change in API parameters.

### Tests written?
Yes, unit tests verify *function/method signatures only*

### Commits signed with GPG?

Yes


dhiltonp:
As of 2018.10.01, the VirtualHardDisk definition is:

```python
    def __init__(self, *, uri: str=None, **kwargs) -> None:
        super(VirtualHardDisk, self).__init__(**kwargs)
        self.uri = uri
```

dwoz (approved):

PR #56156, opened 2020-02-13
Fix incorrect list of installed packages via dpkg

initd:
### What does this PR do?
Fix incorrect list of installed packages (dpkg) with status "hold" and sub-status "triggers-pending"

**usual situation:**
`root@test ~ # dpkg-query --showformat '${Status} ${Package} ${Version} ${Architecture}' -W vim`
`install ok installed vim 2:8.0.1453-1ubuntu1.1 amd64`

**trouble case:**
`root@test ~ # dpkg-query --showformat '${Status} ${Package} ${Version} ${Architecture}' -W systemd`
`hold ok triggers-pending systemd 237-3ubuntu10.38 amd64`
In trouble case systemd is already installed but has status "triggers-pending" and will be skipped by function list_pkgs so the Salt will try to install it again.

### Previous Behavior
Utility apt tried to repeatedly install target held package because function list_pkgs return its incorrect status

### New Behavior
Packages with a state "hold triggers-pending" will handle correctly

### Tests written?
No

### Commits signed with GPG?
No

### DIFF
--- a/salt/modules/aptpkg.py
+++ b/salt/modules/aptpkg.py
@@ -1330,7 +1330,7 @@ def list_pkgs(versions_as_list=False,
                 name += ':{0}'.format(arch)
         if cols:
             if ('install' in linetype or 'hold' in linetype) and \
-                    'installed' in status:
+                    ('installed' in status or 'triggers-pending' in status):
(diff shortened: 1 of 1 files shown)


dwoz:

PR #56159, opened 2020-02-13
Change x509.certficate_managed

svenseeberg:
### What does this PR do?
This change decodes a bytes variable to an utf-8 string before concatenating to another string.

On Python 3.7 I get the following error for `x509.certificate_managed`, which is fixed in this PR:
```
                File "/usr/lib/python3/dist-packages/salt/state.py", line 1919, in call
                  **cdata['kwargs'])
                File "/usr/lib/python3/dist-packages/salt/loader.py", line 1918, in wrapper
                  return f(*args, **kwargs)
                File "/usr/lib/python3/dist-packages/salt/states/x509.py", line 560, in certificate_managed
                  'contents'] += __salt__['x509.get_pem_entry'](append_cert, pem_type='CERTIFICATE')
              TypeError: can only concatenate str (not "bytes") to str
```

### Tests written?

No

### Commits signed with GPG?

Yes

### DIFF
--- a/salt/states/x509.py
+++ b/salt/states/x509.py
@@ -718,7 +718,7 @@ def certificate_managed(
             append_file_contents = __salt__["x509.get_pem_entry"](
                 append_file, pem_type="CERTIFICATE"
             )
-            contents += append_file_contents
+            contents += salt.utils.stringutils.to_str(append_file_contents)
         except salt.exceptions.SaltInvocationError as e:
             ret["result"] = False
             ret[


salt/states/x509.py
```diff
@@ -578,7 +578,7 @@ def certificate_managed(
     for append_cert in append_certs:
         file_args["contents"] += __salt__["x509.get_pem_entry"](
             append_cert, pem_type="CERTIFICATE"
-        )
+        ).decode("utf-8")
```
DmitryKuzmenko, line 581: We're trying to avoid direct calls to `decode` and `encode` functions. Previous code appending `file_args["contents"]` uses `salt.utils.stringutils.to_str` for data in this field. I think it would be more correct to use the same utility function here.

DmitryKuzmenko (changes requested):

DmitryKuzmenko:
@svenseeberg thank you for contribution! Could you please take a look at my review comment?
Also could you please consider a possibility to write a test for your change? Ask for any help if you need an advice on that.

salt/states/x509.py
svenseeberg: Alright, that should be pretty straight forward.

svenseeberg:
@DmitryKuzmenko I will have a look at the tests. I have to admit I'm a little short on time though.

svenseeberg:
Two of my Debian Buster minions failed to refresh their certificates due to this. I rebased the branch to the current `master`.

Not sure how to test this. The problem seems to be specific to Debian Buster.

svenseeberg:
Oh it seems that these two nodes still used the versions packaged in the Debian main repos and the problem itself has been solved in current releases. I'll close the PR.

svenseeberg:
Oh no, the problem persists after upgrading to the current release:

```
----------
          ID: pki-client_cert
    Function: x509.certificate_managed
        Name: /etc/pki/client.crt
      Result: False
                File "/usr/lib/python3/dist-packages/salt/state.py", line 2154, in call
                File "/usr/lib/python3/dist-packages/salt/loader.py", line 2106, in wrapper
                  return f(*args, **kwargs)
                File "/usr/lib/python3/dist-packages/salt/states/x509.py", line 721, in certificate_managed
                  contents += append_file_contents
              TypeError: can only concatenate str (not "bytes") to str
     Started: 14:54:50.564209
    Duration: 368.706 ms
     Changes:   
```

The change that comes with this PR definitely fixes the problem.

svenseeberg:
It seems that this was now fixed with some other commits. With the newest version I can deploy minions without patching the `x509.py`.

svenseeberg:
Now this error is occuring on Ubuntu 18.04 nodes.

svenseeberg:
It would be really cool to have this merged at some point. I need to manually fix the x509.py for each new minion.

Ch3LL (changes requested):
this will require a changelog and test coverage to ensure this does not regress

svenseeberg:
I'll close the PR due to the rewrite: https://github.com/saltstack/salt/issues/58481#issuecomment-1415847600

Issue #55284, opened 2019-11-13
reg.present claims change even though there is none


flassman:
### Description of Issue
Running a reg.present state will always say something has changed even though it has not. 
This makes highstate monitoring impossible.
The state is repeatedly executed which should not be the case.
This only happens if the vdata is passed as a string (see below)

### Setup
```
HKEY_LOCAL_MACHINE:
  reg.present:
    - vname: MyTest
    - vtype: REG_DWORD
    - vdata: '00000001'
```
### Steps to Reproduce Issue
use the state above and execute it twice. it claims something has changed even if nothing has changed during the second run:

>           ID: HKEY_LOCAL_MACHINE
>     Function: reg.present
>       Result: True
>      Comment: Added HKEY_LOCAL_MACHINE to HKEY_LOCAL_MACHINE\
>      Started: 12:08:47.424619
>     Duration: 8.998 ms
>      Changes:   
>               ----------
>               inheritance:
>                   True
>               reg:
>                   ----------
>                   Added:
>                       ----------
>                       Entry:
>                           MyTest
>                       Inheritance:
>                           True
>                       Key:
>                           HKEY_LOCAL_MACHINE\
>                       Owner:
>                           None
>                       Perms:
>                           ----------
>                           Deny:
>                               None
>                           Grant:
>                               None
>                       Value:
>                           1

### Versions Report
the problem was observed both with 2018.3.4 and 2019.2.2. Anyway, here comes the versions report:

> Salt Version:
>            Salt: 2019.2.2
>  
> Dependency Versions:
>            cffi: Not Installed
>        cherrypy: Not Installed
>        dateutil: 1.5
>       docker-py: Not Installed
>           gitdb: 2.0.3
>       gitpython: 2.1.8
>           ioflo: Not Installed
>          Jinja2: 2.7.2
>         libgit2: Not Installed
>         libnacl: Not Installed
>        M2Crypto: 0.31.0
>            Mako: Not Installed
>    msgpack-pure: Not Installed
>  msgpack-python: 0.5.6
>    mysql-python: Not Installed
>       pycparser: Not Installed
>        pycrypto: 2.6.1
>    pycryptodome: Not Installed
>          pygit2: Not Installed
>          Python: 2.7.5 (default, Jun 11 2019, 14:33:56)
>    python-gnupg: Not Installed
>          PyYAML: 3.11
>           PyZMQ: 15.3.0
>            RAET: Not Installed
>           smmap: 2.0.3
[...]

PR #56160, opened 2020-02-13
Fix issue with existing reg_dword entries

twangboy:
### What does this PR do?
When passing an invalid formatted reg_dword data in the state file, such as '000001', the registry code casts it to the proper type. In the case of '000001' it would be cast to an integer `1`. This conversion was happening after the comparison to see if changes were to be made, thus comparing
'000001' to the existing integer `1`.

This PR moves the casting of the state file data before the comparison so that it is comparing proper types.

Also fixes some issues with the comments being returned by the state file where it was returning incorrect duplicate strings. `<key name> added to <key name>`. Now it returns `<key value> added to <key name>`

Added tests for the integer comparison.

Fixed tests for the new comments

https://github.com/saltstack/salt/issues/55284

### Tests written?
Yes

### Commits signed with GPG?
Yes

### DIFF
--- a/tests/unit/utils/test_win_reg.py
+++ b/tests/unit/utils/test_win_reg.py
@@ -637,6 +637,10 @@ class WinFunctionsTestCase(TestCase):
         result = win_reg.cast_vdata(vdata=vdata, vtype='REG_DWORD')
         self.assertTrue(isinstance(result, six.integer_types))
 
+        vdata = '0000001'
+        result = win_reg.cast_vdata(vdata=vdata, vtype='REG_DWORD')
+        self.assertTrue(isinstance(result, six.integer_types))
+
     def test_cast_vdata_reg_expand_sz(self):
         '''
         Test the cast_vdata function with REG_EXPAND_SZ
--- a/tests/unit/utils/test_win_reg.py
+++ b/tests/unit/utils/test_win_reg.py
@@ -630,16 +630,17 @@ class WinFunctionsTestCase(TestCase):
         Should always return integer
         '''
         vdata = 1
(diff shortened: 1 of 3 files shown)


tests/unit/utils/test_win_reg.py
```diff
@@ -637,6 +637,10 @@ def test_cast_vdata_reg_dword(self):
         result = win_reg.cast_vdata(vdata=vdata, vtype='REG_DWORD')
         self.assertTrue(isinstance(result, six.integer_types))
 
+        vdata = '0000001'
+        result = win_reg.cast_vdata(vdata=vdata, vtype='REG_DWORD')
+        self.assertTrue(isinstance(result, six.integer_types))
```
waynew: Should the expected value be `1`? If so, this test would be more useful like this:

```suggestion
        vdata = '0000001'
        expected_result = 1
        result = win_reg.cast_vdata(vdata=vdata, vtype='REG_DWORD')
        self.assertEqual(result, expected_result)
```
Rewritten this way the test already checks that the types are correct but *also* that the value gets correctly converted.

waynew (approved):

tests/unit/utils/test_win_reg.py
twangboy: Damn... I'll make the change. Why you gotta be so smart...?

waynew (approved):
🚀

dwoz (approved):

cmcmarrow (approved):

s0undt3ch (approved):

Akm0d (approved):

PR #56161, opened 2020-02-14
set `up` in alive function

vnitinv:
### What does this PR do?
fix persistent connection to ssh connection lost

ping checks for `up` status, so alive should set that

### Previous Behavior
When ssh connection was broken, it was taking huge time to come back

### New Behavior
next keep-alive will make sure new connection is established.

### Tests written?

No

### Commits signed with GPG?

No

### DIFF
--- a/salt/proxy/netmiko_px.py
+++ b/salt/proxy/netmiko_px.py
@@ -262,7 +262,8 @@ def alive(opts):
     if not netmiko_device['always_alive']:
         return True
     if ping() and initialized():
-        return netmiko_device['connection'].remote_conn.transport.is_alive()
+        netmiko_device['up'] = netmiko_device['connection'].remote_conn.transport.is_alive()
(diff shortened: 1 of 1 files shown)


dhiltonp:
This looks like a good change. 

[netmiko's is_alive never attempts to re-connect](https://github.com/ktbyers/netmiko/blob/0ff8953b08291d483df72ed2b939c1b4a0077c04/netmiko/base_connection.py#L432), so updating `netmiko_device['up']` to reflect that is a good move.

Subsequent calls to `ping()` will be false, so the function explicitly returns `False`.

---------

Could you add a test file `tests/unit/proxy/test_netmiko_px.py` with a test that verifies the expected behavior? Specifically, that the short-circuit works as expected.

Obviously we know it works right now, but we'd like to prevent any regression!

dwoz:

Issue #55687, opened 2019-12-17
selinux.port_policy_present state can add but not modify


knine:
### Description of Issue
The selinux.port_policy_present state does not modify SELinux Manage Policy if needed, only adds (or deletes using when using selinux.port_policy_absent) the selinux port.

### Setup
Two example states for selinux.port_policy_present:

```
'Add Port 8000 To SELinux Port Policy':
  selinux.port_policy_present:
    - name: example1
    - protocol: tcp
    - port: 8000
    - sel_type: http_port_t
```
```
'Add Port 8007 To SELinux Port Policy':
  selinux.port_policy_present:
    - name: example2
    - protocol: tcp
    - port: 8007
    - sel_type: http_port_t
```
### Steps to Reproduce Issue

Applying the states results in:
```
          ID: Add Port 8000 To SELinux Port Policy
    Function: selinux.port_policy_present
        Name: example1
      Result: False
     Comment: Error adding new policy: {u'pid': 36419, u'retcode': 1, u'stderr': u'ValueError: Port tcp/8000 already defined', u'stdout': u''}
     Started: 14:31:03.330186
    Duration: 522.656 ms
     Changes: 
```
```
          ID: Add Port 8000 To SELinux Port Policy
    Function: selinux.port_policy_present
        Name: example2
      Result: True
     Comment: 
     Started: 14:31:38.357825
    Duration: 2997.401 ms
     Changes:   
              ----------
              new:
                  ----------
                  port:
                      8007, 80, 81, 443, 488, 8008, 8009, 8443, 9000
                  protocol:
                      tcp
                  sel_type:
                      http_port_t
              old:
                  None
```
Inspecting the system prior to applying the highstate:

```
# semanage port -l | grep 8000
soundd_port_t                  tcp      8000, 9433, 16001
```
```
# semanage port -l | grep 8007
<no output>
```
Running the system commands by hand:

```
# semanage port --add --proto tcp 8000 -t http_port_t
ValueError: Port tcp/8000 already defined
```
```
# semanage port --modify --proto tcp 8000 -t http_port_t
# semanage port -l | grep http_port_t
http_port_t                    tcp      8000, 80, 81, 443, 488, 8008, 8009, 8443, 9000
```
After running the semanage port command by hand with the --modify option, then the highstate for port 8000 is true.

### Versions Report

Master:

```
Salt Version:
           Salt: 2019.2.2
 
Dependency Versions:
           cffi: Not Installed
       cherrypy: Not Installed
[...]

PR #56164, opened 2020-02-14
Enable port modification in state selinux.port_policy_present

driskell:
### What does this PR do?

Allow port_policy_present to modify the SELinux type for a port that already has a type, such as a port that exists in the system policy.

#55687

### Previous Behavior
port_policy_present would call port_add_policy even though the port already exists under a different type, which calls semanage -a. This will fail if the port already exists in the policy. Default ports such as 8080 already exist in the system policy so previously you could not set additional types for any default port as it would always fail.

### New Behavior
When a port already exists in the policy, port_modify_policy is used which will call semanage -m and successfully modify the port type.

Because policy modifications results in port_get_policy finding two policies if sel_type is omitted and only returning the first, most likely the default policy and not the modification, there is first a check for an exact matching policy including sel_type, and then if that doesn't find one, it then checks for ANY policy regardless of sel_type, and thus performs a modification. If the second check finds no policy it then performs an addition.

### Tests written?
Yes

### Commits signed with GPG?
No

### DIFF
--- a/changelog/55687.fixed.md
+++ b/changelog/55687.fixed.md
@@ -1,0 +1,1 @@
+Fix the ability of the 'selinux.port_policy_present' state to modify.
--- a/salt/modules/selinux.py
+++ b/salt/modules/selinux.py
@@ -10,7 +10,6 @@ Execute calls on selinux
     proper packages are installed.
 """
 
-
 import os
 import re
 
@@ -807,10 +806,45 @@ def port_add_policy(name, sel_type=None, protocol=None, port=None, sel_range=Non
 
     .. code-block:: bash
 
-        salt '*' selinux.port_add_policy add tcp/8080 http_port_t
-        salt '*' selinux.port_add_policy add foobar http_port_t protocol=tcp port=8091
+        salt '*' selinux.port_add_policy tcp/8080 http_port_t
+        salt '*' selinux.port_add_policy foobar http_port_t protocol=tcp port=8091
     """
     return _port_add_or_delete_policy("add", name, sel_type, protocol, port, sel_range)
+
+
+def port_modify_policy(name, sel_type=None, protocol=None, port=None, sel_range=None):
+    """
+    .. versionadded:: 2019.2.0
+
+    Modifies the SELinux policy for a given protocol and port.
+
+    Returns the result of the call to semanage.
+
+    name
+        The protocol and port spec. Can be formatted as ``(tcp|udp)/(port|port-range)``.
+
+    sel_type
+        The SELinux Type. Required.
+
+    protocol
+        The protocol for the port, ``tcp`` or ``udp``. Required if name is not formatted.
+
+    port
+        The port or port range. Required if name is not formatted.
+
+    sel_range
+        The SELinux MLS/MCS Security Range.
+
+    CLI Example:
+
+    .. code-block:: bash
+
+        salt '*' selinux.port_modify_policy tcp/8080 http_port_t
+        salt '*' selinux.port_modify_policy foobar http_port_t protocol=tcp port=8091
+    """
+    return _port_add_or_delete_policy(
+        "modify", name, sel_type, protocol, port, sel_range
+    )
 
 
 def port_delete_policy(name, protocol=None, port=None):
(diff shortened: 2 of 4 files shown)


driskell:
Couldn't run the tests locally on OSX as the requirements/dev_python27.txt doesn't seem to bring in pytest-salt, and when I tried to use requirements/tests.txt, it will now run but seems to abort... so will monitor this if they fail in the CI.

driskell:
Not sure why tests aren't running:
`ERROR: Could not find credentials entry with ID 'AWS_ACCESS_KEY_ID'`

bryceml:
> Not sure why tests aren't running:
> `ERROR: Could not find credentials entry with ID 'AWS_ACCESS_KEY_ID'`

This is because our CI is not configured to work against that branch.  You should change your pr to go into master instead of 2019.2. See: https://docs.saltstack.com/en/master/topics/development/contributing.html#salt-s-branch-topology

driskell:
Aha my bad. I’ll rebase

duckfez:
Any update available on this?

bradpoulton:
I am in need of this fix for the selinux.port_policy_present. Thanks for your hard work!

driskell:
Sorry I didn't get around to the rebase and then completely forgot about it!
I've just rebased now onto master and pushed again so hopefully this unblocks things from merge for @bryceml

driskell:
I noticed the None appearing in test failure and it's thrown up another existing issue - that if you use name to specify protocol/port it reports "None" in the output. So I have expanded test to include the protocol and port parameter version of the call, and removes the text that can report None. Didn't want to bring the parser into the state module.

Same issue exists in fcontext but I don't want to creep the scope of the PR as it's been a long time and this really needs to be merged.

driskell:
Managed to get `pre-commit` to work to run the blacken/lints and also got `nox` to work from the new documentation really easily! Much less stress than before - though I didn't use docs for `pre-commit` I stole the command line from Jenkins.

Anyhows...

- Rebased on top of latest master
- Pre-commit hooks all pass and blacken / lint has completed on all modified files
- Tests are now fixed and working correctly, there were some gruesome errors in there, my apologies for that

Hopefully this is now ready to go!

driskell:
@dwoz Can we remove the Has-Failing-Test label? All tests are now working.

Thanks!

twangboy (dismissed):

twangboy:
@driskell Please address the issues in pre-commit

driskell:
@twangboy Sorry I do not have bandwidth to contribute further. Pre-commit was passing in October 2020. https://jenkinsci.saltstack.com/job/pr-pre-commit/job/PR-56164/5/display/redirect Could someone at SaltStack pick this up? I have no personal need for it anymore (I maintain my own internal module now.)

andreasdijkman:
What do we need to do to get this PR merged? I'm running into the same problem as state in the issue and this PR.
I've tried looking through these checks but the links are dead. Also some of them are marked as required but cannot be completed. I don't think macOS, FreeBSD or Windows have SELinux capabilities. Not sure about Ubuntu/Debian though.

driskell:
@twangboy I managed to find some time - migrated the test to pytests folder and updated. I can see there's some dodgy change in master at the moment where protocol will be output as `None` if the policy already exists and you didn't explicitly specify the protocol and allowed it to be parsed from the `name` instead. I chose not to fix that here - it should be fixed elsewhere - specifically it likely means port and protocol parsing needs to be public from the module... but then it grays responsibilities so I think needs a discussion first

driskell:
@twangboy Just to let you know I've tested this now on 3006.4 as an override drop in to the tree in `_modules` and `_states` and it's all working as expected. Please do let me know if you need something to get this merged. Really keen not to lose the momentum again as this is the second time it's been ready :+1:

driskell:
I've split to two commits so that the lint changes from pre-commit are applied first, and then the code changes, which hopefully will ease the code review. First commit should be zero change just syntax. Second commit is the fix for this PR.

I've confirmed that the resulting code is (almost) exactly the same as the previous one I've tested. It's just no longer got line wrap on the long line (`+ f'with specified sel_type "{sel_type}", protocol "{protocol}" and port "{port}".',`) previously it was split over two lines but looks like it isn't needed to be split anymore.

dwoz:
@driskell Are you still wanting to get this in?

s0undt3ch (approved):

Issue #55376, opened 2019-11-20
RabbitMQ Module Incompatible with Rabbit 3.8+


homelinen:
### Description of Issue

When creating a user with Salt in RabbitMQ 3.8.0 I receive the following error:

```
2019-11-20 10:56:30,492 [salt.state       :322 ][ERROR   ][28696] Error: RabbitMQ command failed: Error:
{:undef, [{:rabbit_auth_backend_internal, :check_user_login, ["admin", [password: "adminPassword1"]], []}, {:erl_eval, :do_apply, 6, [file: 'erl_eval.erl', line: 680]}, {:rpc, :"-handle_call_call/6-fun-0-", 5, [file: 'rpc.erl', line: 197]}]}
```

This issue is partly because `rabbit_auth_backend` was renamed in 3.5.7 but it is being used in 3.8 because of an output format change.

The test for version is here: https://github.com/saltstack/salt/blob/master/salt/modules/rabbitmq.py#L459 (But it fails for 3.8)

This looks like one of the problems, as the format of `rabbitmqctl status` changed in 3.8 so this regex no longer works:
https://github.com/saltstack/salt/blob/master/salt/modules/rabbitmq.py#L445

And it falls back to https://github.com/saltstack/salt/blob/master/salt/modules/rabbitmq.py#L486

Rabbitmq Status before 3.8
```
$ rabbitmqctl status
[{pid,305}, {running_applications,
    [{rabbit,"RabbitMQ","3.7.21"},
```

RabbitMQ status in 3.8
```
$ rabbitmqctl status
Status of node rabbit ...
Runtime

OS PID: 4712
OS: Linux
Uptime (seconds): 3294202
RabbitMQ version: 3.8.0
Node name: rabbit@dev-g-rbt1-01
Erlang configuration: Erlang/OTP 22 [erts-10.5.2] [source] [64-bit] [smp:2:2] [ds:2:2:10] [async-threads:64] [hipe]
Erlang processes: 539 used, 1048576 limit
Scheduler run queue: 1
Cluster heartbeat timeout (net_ticktime): 60
```

One thing that works is:
```
rabbitmqctl eval 'rabbit_access_control:check_user_login(<<"admin">>, [{password, <<"adminPassword1">>}]).'
```

### Setup

```sls
rabbitadminUser:
  rabbitmq_user.present:
    - name: admin
    - password: "adminPassword1"
    - tags:
      - administrator
    - perms:
      - '/':
        - ".*"
        - ".*"
[...]

Issue #33588, opened 2016-05-27
rabbitmq_user.present error 


whytewolf:
### Description of Issue/Question

with 2 users being maintained by rabbitmq_user.present from older builds. on the latest release 2016.3.0 rabbitmq_user.present has an odd error. this setup worked before since at least 2015.5.5

### Setup
(Please provide relevant configs and/or SLS files (Be sure to remove sensitive info).)
```
openstack rabbit user:
  rabbitmq_user.present:
    - name: openstack
    - password: password1
    - perms: [{/: [.*, .*, .*]}]


whytewolf rabbit user:
  rabbitmq_user.present:
    - name: whytewolf
    - password: password2
    - perms: [{/: [.*, .*, .*]}]
    - tags: [administrator]

[DEBUG   ] Results of YAML rendering:
OrderedDict([('openstack rabbit user', OrderedDict([('rabbitmq_user.present', [OrderedDict([('name', 'openstack')]), OrderedDict([('password', 'password1')]), OrderedDict([('perms', [OrderedDict([('/', ['.*', '.*', '.*'])])])])])])), ('whytewolf rabbit user', OrderedDict
([('rabbitmq_user.present', [OrderedDict([('name', 'whytewolf')]), OrderedDict([('password', 'password2')]), OrderedDict([('perms', [OrderedDict([('/', ['.*', '.*', '.*'])])])]), OrderedDict([('tags', ['administrator'])])])]))])

[DEBUG   ] LazyLoaded rabbitmq_user.present
[INFO    ] Running state [openstack] at time 09:02:54.112737
[INFO    ] Executing state rabbitmq_user.present for openstack
[INFO    ] Executing command ['rabbitmqctl', 'list_users'] as user 'root' in directory '/root'
[DEBUG   ] output: Listing users ...
whytewolf       [administrator]
openstack       []
guest   [administrator]
[DEBUG   ] {'openstack': set(['']), 'whytewolf': set(['administrator']), 'guest': set(['administrator'])}
[ERROR   ] Error: RabbitMQ command failed: Error: {undef,
           [{rabbit_auth_backend_internal,check_user_login,
                [<<"openstack">>,[{password,<<"password1">>}]],
                []},
            {erl_eval,do_apply,6,[{file,"erl_eval.erl"},{line,569}]},
            {rpc,'-handle_call_call/6-fun-0-',5,
                [{file,"rpc.erl"},{line,205}]}]}
[...]

PR #56193, opened 2020-02-18
Update check_password function for newer RabbitMQ versions.

11chrisadams11:
### What does this PR do?
Newer versions of RabbitMQ have a differing output to older versions that breaks the search regex in the check_password function. This caused rabbitmq_user states to fail. This PR adds a check for the new status output as well, so the function will work for all versions of RabbitMQ.

Follow up to #33588
Fixes issues #56258 & #55376

### Previous Behavior
The check_password function would only search the regex against the status output of older versions of RabbitMQ, which would cause errors on newer versions of RabbitMQ.

### New Behavior
The check_password function will now search for the regex of the status output for both new and old versions of RabbitMQ.

### Tests written?
Yes

### Commits signed with GPG?
No


DmitryKuzmenko (dismissed):

DmitryKuzmenko:
@11chrisadams11  Could you please add a test case for it?

11chrisadams11:
@DmitryKuzmenko Added test cases for the different versions.

DmitryKuzmenko (approved):

DmitryKuzmenko:
Perfect! Thank you for updating this with tests. I've blackened the code now tests should be green. Let's wait for ci.

Issue #56131, opened 2020-02-11
Salt v3000 crashes with unless clause if pip is missing


thusoy:
### Description of Issue
`archive.extracted` (and possibly others?) fails if setting an `unless` clause and pip is not installed. Works on 2019.2.3.

### Setup
```yaml
# archive-test
vault:
    archive.extracted:
        - name: /usr/local/bin/
        - source: https://releases.hashicorp.com/vault/1.3.2/vault_1.3.2_linux_amd64.zip
        - source_hash: sha256=b49de8fd508eb9c2c222fa0f38e23546fb28991af2e8bfdb9bbe381a380f9baa
        - archive_format: zip
        - enforce_toplevel: False
        - unless:
            - echo hello
```

### Steps to Reproduce Issue
`salt-call state.sls archive-test` without pip installed. Removing the `unless` clause or installing `python-pip` resolves the issue.

### Versions Report

Masterless.
```
Salt Version:
           Salt: 3000

Dependency Versions:
           cffi: Not Installed
       cherrypy: Not Installed
       dateutil: 2.5.3
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
         Jinja2: 2.9.4
        libgit2: Not Installed
       M2Crypto: Not Installed
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.4.8
   mysql-python: Not Installed
      pycparser: Not Installed
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: Not Installed
         Python: 2.7.13 (default, Sep 26 2018, 18:42:22)
   python-gnupg: Not Installed
         PyYAML: 3.12
          PyZMQ: 16.0.2
          smmap: Not Installed
        timelib: Not Installed
        Tornado: 4.5.3
            ZMQ: 4.2.1

System Versions:
           dist: debian 9.4
         locale: UTF-8
        machine: x86_64
        release: 4.9.0-6-amd64
         system: Linux
        version: debian 9.4
```

There's a very long stacktrace when it crashes, this is the beginning and end:
```
[ERROR   ] Failed to import states pip_state, this is due most likely to a syntax error:
Traceback (most recent call last):
  File "/usr/lib/python2.7/dist-packages/salt/loader.py", line 1607, in _load_module
    mod = imp.load_module(mod_namespace, fn_, fpath, desc)
[...]

PR #56215, opened 2020-02-20
Fix for unless requisite when pip is not installed

dwoz:
### What does this PR do?

Fix for #56131

### Tests written?

Yes

### Commits signed with GPG?

Yes


tests/integration/modules/test_state.py
```diff
@@ -2268,3 +2268,27 @@ def test_state_sls_lazyloader_allows_recursion(self):
...
                          'Success!')
         self.assertTrue(state_run[state_id]['result'])
+
+    def test_issue_56131(self):
+        module_path = os.path.join(RUNTIME_VARS.CODE_DIR, 'pip.py')
+        if six.PY3:
+            modulec_path = os.path.join(RUNTIME_VARS.CODE_DIR, '__pycache__', 'pip.pyc')
+        else:
+            modulec_path = os.path.join(RUNTIME_VARS.CODE_DIR, 'pip.pyc')
+        print('*' * 80)
```
Ch3LL: do we want to keep these prints here?

Ch3LL (changes requested):

max-arnold:
@dwoz I believe this line should start with `if` (not `or`), otherwise it will iterate and delete all modules:

https://github.com/saltstack/salt/blob/1746e6ac60d1b6f9f2d3cb372e043d2aad66fcc0/salt/states/pip_state.py#L54

dwoz:
> @dwoz I believe this line should start with `if` (not `or`), otherwise it will iterate and delete all modules:
> 
> https://github.com/saltstack/salt/blob/1746e6ac60d1b6f9f2d3cb372e043d2aad66fcc0/salt/states/pip_state.py#L54

Good catch, thanks!

tests/integration/modules/test_state.py
dwoz: Nope, will resolve this

Ch3LL (approved):

Ch3LL:
looks like there are some lint failures that need to be cleaned up

DmitryKuzmenko (approved):

waynew (approved):

PR #56220, opened 2020-02-21
config.option returns empty string not None

mirceaulinic:
``__salt__['config.option']`` can never return `None`:

```bash
# salt tgt config.option fake --out=raw
{u'tgt': u''}
```

Therefore, those statements are never evaluated, which
results in surprising behaviour and wasted hours of debugging.

### DIFF
--- a/salt/modules/s3.py
+++ b/salt/modules/s3.py
@@ -303,19 +303,19 @@ def _get_key(key, keyid, service_url, verify_ssl, kms_keyid, location, role_arn,
     if not service_url:
         service_url = 's3.amazonaws.com'
 
-    if verify_ssl is None and __salt__['config.option']('s3.verify_ssl') is not None:
+    if verify_ssl is None and __salt__['config.option']('s3.verify_ssl'):
         verify_ssl = __salt__['config.option']('s3.verify_ssl')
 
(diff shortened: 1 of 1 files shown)


twangboy (approved):

salt/modules/s3.py
```diff
@@ -303,19 +303,19 @@ def _get_key(key, keyid, service_url, verify_ssl, kms_keyid, location, role_arn,
...
         location = __salt__['config.option']('s3.location')
 
     if role_arn is None and __salt__['config.option']('s3.role_arn'):
         role_arn = __salt__['config.option']('s3.role_arn')
 
-    if path_style is None and __salt__['config.option']('s3.path_style') is not None:
+    if path_style is None and __salt__['config.option']('s3.path_style'):
         path_style = __salt__['config.option']('s3.path_style')
 
     if path_style is None:
```
dhiltonp: There is another of these checks 3 lines down, that should also be included!

dhiltonp (changes requested):
This looks good.

I did a quick check through the rest of the code base and we seem fine, except for one more occurrence just a few lines down!

dwoz:

PR #56226, opened 2020-02-21
Replace log.warn by log.warning and deprecated ElementTree.getchildren()

bdrung:
When running the test suite, following deprecation warnings appear:

* DeprecationWarning: The `warn` method is deprecated, use `warning` instead.
* DeprecationWarning: This method will be removed in future versions. Use `list(elem)` or iteration over elem instead.

Replace `log.warn` by `log.warning` and deprecated `ElementTree.getchildren()`.

### DIFF
--- a/salt/modules/win_lgpo.py
+++ b/salt/modules/win_lgpo.py
@@ -5576,7 +5576,7 @@ def _getAdmlPresentationRefId(adml_data, ref_id):
                 if presentation_element:
                     presentation_element = presentation_element[0]
                     if TEXT_ELEMENT_XPATH(presentation_element):
-                        for p_item in presentation_element.getchildren():
+                        for p_item in presentation_element:
                             if p_item == result:
                                 break
                             if etree.QName(p_item.tag).localname == "text":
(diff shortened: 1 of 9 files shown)


salt/modules/win_lgpo.py
```diff
@@ -8131,7 +8131,7 @@ def _get_policy_adm_setting(admx_policy,
             configured_elements = {}
             policy_disabled_elements = 0
             for elements_item in ELEMENTS_XPATH(admx_policy):
-                for child_item in elements_item.getchildren():
+                for child_item in elements_itemHashable:
```
mchugh19: I could be wrong, but I don't see a variable of `elements_itemHashable`, so this seems broken. There are a few more instances of this below as well.

bdrung: Ups. That's a copy-paste failure. I'll fix it.

twangboy (dismissed):

s0undt3ch (dismissed):
Approved, though splitting the fixed would have been preferred, though, not a blocking reason not to approve.

Ch3LL (dismissed):

dwoz (approved):

Issue #44618, opened 2017-11-20
Allow Configuration of "Let Windows apps" Registry Settings


doesitblend:
### Description of Issue/Question
Allowing settings to be configured for Application security. These settings are found under:

### Setup

```
configure_app_privacy:
  lgpo.set:
    - computer_policy:
        Let Windows apps access account information: Enabled
        Default for all apps: Force Deny
        Put user in control of these specific apps (use Package Family Names): ""
        Force allow these specific apps (use Package Family Names): ""
        Force deny these specific apps (use Package Family Names): ""
```

### Steps to Reproduce Issue
Run the above state on a Windows 2016 server. You should see the following error:

```
win2016-2:
----------
          ID: configure_app_privacy
    Function: lgpo.set
      Result: False
     Comment: Unable to find Machine policy Default for all apps Unable to find Machine policy Put user in control of these specific apps (use Package Family Names) Unable to find Machine policy Force allow these specific apps (use Package Family Names) Unable to find Machine policy Force deny these specific apps (use Package Family Names)
     Started: 23:06:00.223000
    Duration: 2672.0 ms
     Changes:   

```

### Versions Report
    Salt Version:
               Salt: 2017.7.0-16-gab1b099
     
    Dependency Versions:
               cffi: 1.10.0
           cherrypy: unknown
           dateutil: 2.6.0
          docker-py: Not Installed
              gitdb: 2.0.3
          gitpython: 2.1.3
              ioflo: Not Installed
             Jinja2: 2.9.6
            libgit2: Not Installed
            libnacl: Not Installed
           M2Crypto: Not Installed
               Mako: 1.0.6
       msgpack-pure: Not Installed
     msgpack-python: 0.4.8
       mysql-python: Not Installed
          pycparser: 2.17
           pycrypto: 2.6.1
       pycryptodome: Not Installed
             pygit2: Not Installed
             Python: 2.7.13 (v2.7.13:a06454b1afa1, Dec 17 2016, 20:53:40) [MSC v.1500 64 bit (AMD64)]
       python-gnupg: 0.4.0
             PyYAML: 3.11
              PyZMQ: 16.0.2
               RAET: Not Installed
              smmap: 2.0.3
            timelib: 0.2.4
            Tornado: 4.5.1
                ZMQ: 4.1.6
     
    System Versions:
               dist:   
             locale: cp1252
            machine: AMD64
            release: 2016Server
             system: Windows
            version: 2016Server 10.0.14393  Multiprocessor Free

[...]

PR #56272, opened 2020-02-28
Properly resolve the policy name

twangboy:
# What does this PR do?
Some states were not resolving the policy name from the refId.
Additionally, there was an issue where the element value wasn't being encoded properly

https://github.com/saltstack/salt/issues/44618

### Previous Behavior
Policy name could only be set using refId and not the policy name. `lgpo.get_policy_info` would return only the refId's and not the policy name.

```
local:
    ----------
    message:
    policy_aliases:
        - Let Windows apps access account information
        - LetAppsAccessAccountInfo
        - Windows Components\App Privacy\Let Windows apps access account information
    policy_class:
        machine
    policy_elements:
        |_
          ----------
          element_aliases:
              - LetAppsAccessAccountInfo_Enum
              - Default for all apps
          element_id:
              LetAppsAccessAccountInfo_Enum
        |_
          ----------
          element_aliases:
              - LetAppsAccessAccountInfo_UserInControlOfTheseApps_List
              - LetAppsAccessAccountInfo_UserInControlOfTheseApps_List
          element_id:
              LetAppsAccessAccountInfo_UserInControlOfTheseApps_List
        |_
          ----------
          element_aliases:
              - LetAppsAccessAccountInfo_ForceAllowTheseApps_List
              - LetAppsAccessAccountInfo_ForceAllowTheseApps_List
          element_id:
              LetAppsAccessAccountInfo_ForceAllowTheseApps_List
        |_
          ----------
          element_aliases:
              - LetAppsAccessAccountInfo_ForceDenyTheseApps_List
              - LetAppsAccessAccountInfo_ForceDenyTheseApps_List
          element_id:
              LetAppsAccessAccountInfo_ForceDenyTheseApps_List
    policy_found:
        True
    policy_name:
        Let Windows apps access account information
    rights_assignment:
        False
```

### New Behavior
Policy names are now properly resolved. `lgpo.get_policy_info` now returns policy names:
```
local:
    ----------
    message:
    policy_aliases:
        - Let Windows apps access account information
        - LetAppsAccessAccountInfo
        - Windows Components\App Privacy\Let Windows apps access account information
    policy_class:
        machine
    policy_elements:
        |_
          ----------
          element_aliases:
              - LetAppsAccessAccountInfo_Enum
              - Default for all apps
          element_id:
              LetAppsAccessAccountInfo_Enum
        |_
          ----------
          element_aliases:
              - LetAppsAccessAccountInfo_UserInControlOfTheseApps_List
              - Put user in control of these specific apps (use Package Family Names)
          element_id:
              LetAppsAccessAccountInfo_UserInControlOfTheseApps_List
        |_
          ----------
          element_aliases:
              - LetAppsAccessAccountInfo_ForceAllowTheseApps_List
              - Force allow these specific apps (use Package Family Names)
          element_id:
              LetAppsAccessAccountInfo_ForceAllowTheseApps_List
        |_
          ----------
          element_aliases:
              - LetAppsAccessAccountInfo_ForceDenyTheseApps_List
              - Force deny these specific apps (use Package Family Names)
          element_id:
              LetAppsAccessAccountInfo_ForceDenyTheseApps_List
    policy_found:
        True
    policy_name:
        Let Windows apps access account information
    rights_assignment:
        False
```

### Tests written?
They're coming

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/states/win_lgpo.py
+++ b/salt/states/win_lgpo.py
@@ -112,6 +112,7 @@ import logging
 import salt.utils.data
 import salt.utils.dictdiffer
 import salt.utils.json
+import salt.utils.stringutils
 import salt.utils.versions
 import salt.utils.win_functions
 
@@ -153,6 +154,21 @@ def _compare_policies(new_policy, current_policy):
 
 
 def _convert_to_unicode(data):
+    '''
+    Helper function that makes sure all items in the dictionary are unicode for
+    comparing the existing state with the desired state. This function is only
+    needed for Python 2 and can be removed once we've migrated to Python 3.
+
+    The data returned by the current settings sometimes has a mix of unicode and
+    string values (these don't matter in Py3). This causes the comparison to
+    say it's not in the correct state even though it is. They basically compares
+    apples to apples, etc.
+
+    Also, in Python 2, the utf-16 encoded strings remain utf-16 encoded (each
+    character separated by `/x00`) In Python 3 it returns a utf-8 string. This
+    will just remove all the null bytes (`/x00`), again comparing apples to
+    apples.
+    '''
     if isinstance(data, six.string_types):
         data = data.replace('\x00', '')
         return salt.utils.stringutils.to_unicode(data)
--- a/salt/states/win_lgpo.py
+++ b/salt/states/win_lgpo.py
@@ -152,6 +152,20 @@ def _compare_policies(new_policy, current_policy):
             return False
 
 
+def _convert_to_unicode(data):
+    if isinstance(data, six.string_types):
+        data = data.replace('\x00', '')
+        return salt.utils.stringutils.to_unicode(data)
+    elif isinstance(data, dict):
+        return dict((_convert_to_unicode(k),
+                     _convert_to_unicode(v))
+                    for k, v in data.items())
+    elif isinstance(data, list):
+        return list(_convert_to_unicode(v) for v in data)
+    else:
+        return data
+
+
 def set_(name,
          setting=None,
          policy_class=None,
@@ -341,6 +355,9 @@ def set_(name,
 
                     requested_policy_check = salt.utils.json.loads(requested_policy_json)
                     current_policy_check = salt.utils.json.loads(current_policy_json)
+
+                    if six.PY2:
+                        current_policy_check = _convert_to_unicode(current_policy_check)
 
                     # Are the requested and current policies identical
                     policies_are_equal = _compare_policies(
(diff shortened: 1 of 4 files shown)


dwoz (approved):

tests/unit/states/test_win_lgpo.py
```diff
@@ -174,6 +163,19 @@ def setUp(self):
         with patch.dict(win_lgpo.__opts__, {'test': False}):
             win_lgpo.set_(name='nc_state', computer_policy=computer_policy)
 
+    def _convert_to_unicode(self, data):
```
Ch3LL: looks like this same code is being used 3 times. anyway way we can condense this to one function that we call?

twangboy: Used the function in the state

tests/unit/modules/test_win_lgpo.py
```diff
@@ -594,6 +589,19 @@ def setUp(self):
             win_lgpo.set_(computer_policy=computer_policy)
             self.configured = True
 
+    def _convert_to_unicode(self, data):
```
waynew: Still declaring the function here.

salt/states/win_lgpo.py
```diff
@@ -152,6 +152,20 @@ def _compare_policies(new_policy, current_policy):
             return False
 
 
+def _convert_to_unicode(data):
+    if isinstance(data, six.string_types):
+        data = data.replace('\x00', '')
```
waynew: It's not clear exactly why this code is present. I know, but that's just because we had a conversation about it this last week 😂 

I'm sure I will forget in 6 months when I have to come back here to look 🙃 

Should we add a comment here about why we're doing the `.replace('\x00', '')`? Probably good to call out that if we correctly handle bytes vs. text that this functionality should become obsolete.

waynew (changes requested):
Looks like you missed one of the instances @Ch3LL was referring to 😝

tests/unit/modules/test_win_lgpo.py
twangboy: Yep... I'll get it

salt/states/win_lgpo.py
twangboy: Yeah, I meant to comment the code... but moved on before doing that... I'll add that

twangboy: Dun

waynew (approved):

PR #56278, opened 2020-03-01
use full option name instead of undocumented abbreviation

mcalmer:
### What does this PR do?

Change an option name when call zypper. We used an undocumented abbreviation which will be removed in future versions of zypper. Use the full option name instead.

Prevents:
```
"Module function pkg.list_products threw an exception. Exception: Zypper command failure: The flag --disable-repos is not known.",
```
errors.

### Tests written?

No

Behaviour did not change, we just use the documented option now.

### Commits signed with GPG?

No

### DIFF
--- a/salt/modules/zypperpkg.py
+++ b/salt/modules/zypperpkg.py
@@ -2244,7 +2244,7 @@ def list_products(all=False, refresh=False):
     OEM_PATH = "/var/lib/suseRegister/OEM"
     cmd = list()
     if not all:
-        cmd.append("--disable-repos")
+        cmd.append("--disable-repositories")
     cmd.append("products")
     if not all:
         cmd.append("-i")
--- a/tests/unit/modules/test_zypperpkg.py
+++ b/tests/unit/modules/test_zypperpkg.py
@@ -304,9 +304,18 @@ class ZypperTestCase(TestCase, LoaderModuleMockMixin):
 
             ref_out = {"retcode": 0, "stdout": get_test_data(filename)}
 
-            with patch.dict(
-                zypper.__salt__, {"cmd.run_all": MagicMock(return_value=ref_out)}
-            ):
+            cmd_run_all = MagicMock(return_value=ref_out)
+            mock_call = call(["zypper",
+                              "--non-interactive",
+                              "--xmlout",
+                              "--no-refresh",
+                              "--disable-repositories",
+                              "products", u"-i"],
+                              env={"ZYPP_READONLY_HACK": "1"},
+                              output_loglevel="trace",
+                              python_shell=False)
+
+            with patch.dict(zypper.__salt__, {"cmd.run_all": cmd_run_all}):
                 products = zypper.list_products()
                 self.assertEqual(len(products), 7)
                 self.assertIn(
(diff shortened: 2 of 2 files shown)


brejoc:
Change in zypper is documented here: https://github.com/openSUSE/zypper/blob/fe91bd04de5b2aaead6fe2ed6766c9fafdb1340d/package/zypper.changes#L4

This affects SLE15 SP2 and most probably also Leap 15.2.

mcalmer:
> This affects SLE15 SP2 and most probably also Leap 15.2.

Will be backported to SP1 and GA as well. Maybe also to SLE12 if sombody request it.
Sooner or later it might be required on all SUSE systems.

garethgreenaway:
Looks good.  Needs a test to ensure we don't end up with a regression, something like this:
```
diff --git a/tests/unit/modules/test_zypperpkg.py b/tests/unit/modules/test_zypperpkg.py
index 78414ca4ac..6b488049ce 100644
--- a/tests/unit/modules/test_zypperpkg.py
+++ b/tests/unit/modules/test_zypperpkg.py
@@ -6,6 +6,7 @@
 # Import Python Libs
 from __future__ import absolute_import
 import os
+import logging
 from xml.dom import minidom
 
 # Import Salt Testing Libs
@@ -29,6 +30,8 @@ from salt.ext.six.moves import configparser
 from salt.ext import six
 import salt.utils.pkg
 
+
 
 class ZyppCallMock(object):
     def __init__(self, return_value=None):
@@ -235,7 +238,18 @@ class ZypperTestCase(TestCase, LoaderModuleMockMixin):
                     'stdout': get_test_data(filename)
             }
 
-            with patch.dict(zypper.__salt__, {'cmd.run_all': MagicMock(return_value=ref_out)}):
+            cmd_run_all = MagicMock(return_value=ref_out)
+            mock_call = call(['zypper',
+                              u'--non-interactive',
+                              '--xmlout',
+                              '--no-refresh',
+                              '--disable-repositories',
+                              'products', u'-i'],
+                              env={'ZYPP_READONLY_HACK': '1'},
+                              output_loglevel='trace',
+                              python_shell=False)
+
+            with patch.dict(zypper.__salt__, {'cmd.run_all': cmd_run_all}):
                 products = zypper.list_products()
                 self.assertEqual(len(products), 7)
                 self.assertIn(test_data['vendor'], [product['vendor'] for product in products])
@@ -244,6 +258,7 @@ class ZypperTestCase(TestCase, LoaderModuleMockMixin):
                         self.assertCountEqual(test_data[kwd], [prod.get(kwd) for prod in products])
                     else:
                         self.assertEqual(test_data[kwd], sorted([prod.get(kwd) for prod in products]))
+                cmd_run_all.assert_has_calls([mock_call])
 
     def test_refresh_db(self):
         '''
```

mcalmer:
@garethgreenaway unit test added.

twangboy (dismissed):

garethgreenaway:
@mcalmer Awesome.  Thanks!

garethgreenaway (dismissed):

Ch3LL:
looks like there is a merge conflict that needs to be resolved and pre-commit is failing.

mcalmer:
@Ch3LL rebased - let's see if the tests still succeed

Ch3LL:
ping @garethgreenaway and @twangboy looks like this is also awaiting your review approvals

dwoz (approved):

PR #56280, opened 2020-03-02
Allow skipping test cases that need Internet access

bdrung:
Mark test cases that need Internet access and allow skipping them by setting the environment variable `NO_INTERNET`.

### DIFF
--- a/tests/support/helpers.py
+++ b/tests/support/helpers.py
@@ -622,6 +622,9 @@ def requires_network(only_local_network=False):
                     cls.skipTest("No local network was detected")
                 return func(cls)
 
+            if os.environ.get("NO_INTERNET"):
+                cls.skipTest("Environment variable NO_INTERNET is set.")
+
             # We are using the google.com DNS records as numerical IPs to avoid
             # DNS lookups which could greatly slow down this check
             for addr in (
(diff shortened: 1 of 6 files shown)


twangboy (dismissed):

frogunder (dismissed):

frogunder:
@bdrung seems like there is a conflict here, can you take a look?

Thanks.

bdrung:
Rebased and resolved the conflict.

DmitryKuzmenko (dismissed):

dwoz (changes requested):
I'd prefer we try to detect if we have internet access and then skip the tests automatically. It's a more holistic approach IMO.

bdrung:
For our use-case (reproducible running tests), we want to disable all tests that need Internet access – regardless if the system has actually Internet access or not. On some buildsystems pip is configured to work locally. So a Internet check could be tricked to assume Internet is there, but pip would fail.

frogunder:
@dwoz Is this one good now?

bdrung:
Rebased the merge request on master and resolved the merged conflicts.

dwoz (approved):

Issue #39479, opened 2017-02-17
Dbus not running error.


daanlevi:
### Description of Issue/Question
When service messagebus is stopped on a rhel based system we get the error that salt can not connect to dbus
### Setup
(Please provide relevant configs and/or SLS files (Be sure to remove sensitive info).)
[ERROR   ] org.freedesktop.DBus.Error.NoServer: Failed to connect to socket /var/run/dbus/system_bus_socket: Connection refused
### Steps to Reproduce Issue
(Include debug logs if possible and relevant.)
service messagebus stop and run any salt state
### Versions Report
(Provided by running `salt --versions-report`. Please also mention any differences in master/minion versions.)
salt 2016.11.2-1.el6

NavidZisky:
PR created to fix it:

https://github.com/saltstack/salt/pull/39481

gtmanfred:
@daanlevi does the above PR satisfy your request?

Thanks,
Daniel

daanlevi:
Looks good to me, Thank's.

gtmanfred:
Cool, thanks! once that gets merged, this ticket will be closed.

Thanks!
Daniel

mahrton:
Hi @gtmanfred,

Let me chip in here as I was the one who complained about this issue at our company. I think this is a fundamentally wrong approach to display ERROR or WARNING for this in the Salt logs. The DBus as I understand is solution that delivers messages for window management systems, when none is running it shouldn't even try to place messages into the messagebus (as it is called in RedHat).

It is step forward but now we'll stil get useless warnings for something we don't even need reporting on.

Does it sound reasonable?

Regards
Marton Tatai

gtmanfred:
I disagree there, dbus is no longer just for window management systems, now that systemd requires it for starting services.

Thus, i think it is a good warning to know that certain things that would usually be running are no longer running because dbus is not started.

Thanks,
Daniel

mahrton:
I am not a RedHat expert but we don't have this service running on our servers (RHEL 6.5) and it does not seem to be vital so I believe it's fine like that.
Salt is reporting on something that shouldn't be reported, if no dbus service is running why try post on it? These errors/warnings seem to be irrelevant.

gtmanfred:
Ahh, my appologies, I understand what is actually happening now.

Yeah this shouldn't be logging an error message there IMO, because the error message is printed if someone tries to run the snapper module and the __virtual__ returns the error.

I would support removing that log.warning all together. :+1:

[...]

PR #56286, opened 2020-03-02
Remove dbus warnings from snapper __virtual__

oeuftete:
### What does this PR do?

Related to #39479/#39481.  The original solution didn't go far enough IMO to remove needless warnings.  The user comment at https://github.com/saltstack/salt/issues/39479#issuecomment-280669729 summarizes the situation well.

The exceptions here are already being propagated to a message that comes from a `__virtual__` failure presented at an appropriate severity for the context.  They don't need to be logged as warnings here too.

One context where the warning is particularly unwelcome is running the master from a docker container where there is no `dbus`.  If the python package for `dbus` is installed, a warning-level message like:

```
2020-03-02 13:22:15,082 [salt.loaded.int.module.snapper :67 ][WARNING ][2397] org.freedesktop.DBus.Error.FileNotFound: Failed to connect to socket /run/dbus/system_bus_socket: No such file or directory
```

... will appear in the logs every minute.

### Previous Behavior

See the descriptions above.

### New Behavior

No additional `WARNING`-level message.

### Tests written?

No.  I didn't find any other modules trying to test logs (or absence of) in `__virtual__`.  Maybe there's a way if this is important.

### Commits signed with GPG?

No.


twangboy (approved):

Ch3LL (approved):

dwoz:
@oeuftete We should be catching the exceptions explicitly. I merged this but didn't see that it was an except all. The warnings serve a purpose here, they show which exceptions we should be handling explicitly. Can you please take this task on?

oeuftete:
> @oeuftete We should be catching the exceptions explicitly. I merged this but didn't see that it was an except all. The warnings serve a purpose here, they show which exceptions we should be handling explicitly. Can you please take this task on?

@dwoz Can you clarify for me?  The exceptions involved here are limited to `dbus.DBusException` in one case and `(dbus.DBusException, ValueError)` in the other, and the exception text is still being logged as part of the `__virtual__` failure.  Everything is being logged somewhere still, but now not a second time at a noisy level.

PR #56290, opened 2020-03-03
Ensures popping lgpo.secedit_data does not throw KeyError

### LINKED ISSUE #56288: win_lgpo: `lgpo.set` with secedit policy always generates `KeyError: u'lgpo.secedit_data'`
### Description of Issue

Running the `lgpo.set` execution module with a secedit policy always generates an error: `KeyError: u'lgpo.secedit_data'`.

This happens because `set` calls [_write_secedit_data()](https://github.com/saltstack/salt/blob/9adc2214c3bb7c68f820f7bd5fe5e132b7b3fbc9/salt/modules/win_lgpo.py#L5529) which attempts to pop `lgpo.secedit_data` off the `__context__`, but `set` never calls `_get_secedit_data()` which is where it is added to context.

### Setup
None really required. Just install salt. Tested both 2018.3.4 and 2019.2.3.

### Steps to Reproduce Issue
```
C:\salt\salt-call.bat lgpo.set computer_policy="{'GuestAccountStatus': 'Enabled'}"
```

### Versions Report

lorengordon:
### What does this PR do?

Ensures popping lgpo.secedit_data does not throw KeyError

Fixes #56288

### Previous Behavior
See linked issue

### New Behavior
Avoids a KeyError when using `lgpo.set` with secedit policies

### Tests written?

Yes

### Commits signed with GPG?

Yes


lorengordon:
Just added a test. Turns out this wasn't caught before because the tests were only exercising the `_modal_sets` logic path in `set_`... This bug is in the `_secedits` logic path.

twangboy (approved):

Ch3LL (approved):

DmitryKuzmenko (approved):

tests/integration/modules/test_win_lgpo.py
```diff
@@ -506,6 +506,22 @@ def test_set_computer_policy_ClipboardRedirection(self):
...
+    @destructiveTest
+    def test_set_computer_policy_GuestAccountStatus(self):
+        '''
+        Test setting/unsetting/changing GuestAccountStatus
+        '''
+        # disable GuestAccountStatus
+        self._testSeceditPolicy(
+            'GuestAccountStatus',
+            'Disabled',
+            [r'^EnableGuestAccount = 0'])
```
waynew: Would this function exhibit the error if we called it a second time? Or should it be failing currently?

We definitely should have a test that shows the KeyError when the change in 5529 is **not** present.

waynew (changes requested):

tests/integration/modules/test_win_lgpo.py
lorengordon: I don't understand what you're saying. This test *will* throw the KeyError if the change in `win_lgpo.py` is *not* present. 5529 is a meaningless number to me, without some kind of reference link.

Myself, I've worked around this differently by using my own custom execution module. If you want to write another test or change the code, I've ticked the box to allow edits from maintainers. Code is easier to understand than words.

waynew: To clarify:

if someone changes the code back to:

      __context__.pop('lgpo.secedit_data')

Or something equivalent, we need a regression test to show that.

waynew: Sorry, 5529 was the line of code that you changed 🙃

lorengordon: Yes, that is _exactly_ what this test does.

waynew (approved):

tests/integration/modules/test_win_lgpo.py
lorengordon: Sorry, I was trying to get at that in my prior comment:

> Just added a test. Turns out this wasn't caught before because the tests were only exercising the `_modal_sets` logic path in `set_`... This bug is in the `_secedits` logic path.

There were tests for this module, and even this function, but the tests were not exercising this logic path in `set_`. The test I've added _does_ exercise the logic path, so without this change in `win_lgpo.py` the test will fail.

waynew: Gotcha - It wasn't obvious that `self._testSeceditPolicy(` was in the `_secedits` logic path, from what was shown in the PR here on GitHub. I *guessed* that your test would expose it, but without checking out the PR and futzing with the tests, or just asking you directly, I wasn't sure.

Now we're both on the same page :) Thanks for responding!

lorengordon: I get it. It took me a while to trace the code and figure out how the tests worked and why this wasn't caught before. The `set_` function could use some serious refactoring to separate these concerns and focus on doing one thing...

waynew: There are definitely a number of places that could use similar refactoring. If you're interested in doing any of that refactoring for this section of the code, we've got a couple of months before the Sodium release - if you're available during one of the Community Open Hours to discuss what you're thinking. If you're not available during that period, I think just opening up an issue documenting the changes you'd like to make would be good enough.

lorengordon:
@waynew Is salt still doing bug fixes to 2018.3? Would it be possible to backport this?

waynew:
@lorengordon No - per https://github.com/saltstack/salt-enhancement-proposals/blob/master/accepted/0014-dev-overhaul.md (and https://docs.google.com/document/d/1DfOGVmsQaqr3rZrFqo-gf9wm1mYM-nwa5a1zROatCXQ/edit) old branches are supported, but they no longer receive active development.

To receive new fixes and features one will either need to upgrade, or backport changes themselves.

But to continue to receive new fixes and features, the recommended approach is to upgrade (after testing, of course).

lorengordon:
Ok, it's just going to be a loooooong road before 3000 stable enough for us to switch. Sigh.

PR #56334, opened 2020-03-09
Fix bad RST in consul_pillar docstring

oeuftete:
### What does this PR do?

Fixes some broken RST in the consul_pillar docs.  I couldn't find a way to simultaneously satisfy RST and lint to actually render "`pillarenv`s", so I rephrased the problem away.

Before:

image

After:

image

No open issues.

### Tests written?

No.  Docs-only change.

### Commits signed with GPG?

No.

### DIFF
--- a/salt/pillar/consul_pillar.py
+++ b/salt/pillar/consul_pillar.py
@@ -22,9 +22,9 @@ All parameters are optional.
 
 The ``consul.token`` requires python-consul >= 0.4.7.
 
-If you have a multi-datacenter Consul cluster you can map your ``pillarenv``s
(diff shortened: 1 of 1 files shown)


DmitryKuzmenko (dismissed):

oeuftete:
Looks like this is legit breaking the lint job, so I probably should take this back and either ignore or fix it better.

```
************* Module salt.pillar.consul_pillar
salt/pillar/consul_pillar.py:2: [W1401(anomalous-backslash-in-string), ] Anomalous backslash in string: '\s'. String constant might be missing an r prefix.
```

dwoz (approved):

PR #56099, opened 2020-02-08
Fix Windows and macOS requirements handling in setup.py

s0undt3ch:
### What does this PR do?
See title.

PR #56335, opened 2020-03-09
add -r support to setup.py

cmcmarrow:
### What does this PR do?
lets salt-bin build on windows

### New Behavior
"-r <file>" is now ran in setup.py

### Tests written?
no

### Commits signed with GPG?
Yes


Akm0d (dismissed):

xeacott (dismissed):

dmurphy18 (commented):
Worried that fixing this issue for 3000.1 will change current mode of operation (even if it is wrong) and now building will have a slew of new requirements appear, similar to what happened with the change from distutils to setuptools in 3000.  Perhaps this would be better postponed till Sodium.  But open to persuasion on the matter

s0undt3ch (commented):
I would advise against merging this for 3000.1

There's an open PR which fixes the requirements for 3000.1 in a less intrusive way.

https://github.com/saltstack/salt/pull/56099

For Sodium, as long as the packaging is not broken, I'm ok with this change.

cmcmarrow:
Windows and Apple are the only two that have -r from what I can see. The normal Windows installer catches all missing modules that this catches. The only place I can see go wrong is Apple and that's just because I don't know how the processes.

setup.py
```diff
@@ -192,7 +192,12 @@ def _parse_requirements_file(requirements_file):
     with open(requirements_file) as rfh:
         for line in rfh.readlines():
             line = line.strip()
-            if not line or line.startswith(('#', '-r')):
+            if not line or line.startswith('#'):
+                continue
+            if not line or line.startswith('-r'):
```
twangboy: Would `not line` ever be evaluated here considering the if statement above?

twangboy (changes requested):

DmitryKuzmenko (dismissed):

s0undt3ch:
This is superseded by https://github.com/saltstack/salt/pull/56904

DmitryKuzmenko:
@s0undt3ch maybe close it then?

s0undt3ch:
It will get closed when the linked PR is merged.

DmitryKuzmenko:
Hm... So this one or #56904?

DmitryKuzmenko:
@s0undt3ch ^^^

DmitryKuzmenko:
Just checked this part in #56904 it also handles an inline hash-comments.
I'd better like #56904

Akm0d (approved):

dmurphy18 (approved):

s0undt3ch:
None for Sodium.
If the issue is salt-bin, we'll address it when building(again for Sodium).

DmitryKuzmenko (approved):

cmcmarrow:
@s0undt3ch the PR is open due to salt-bin.

Issue #56330, opened 2020-03-09
test_issue_2594_non_invalidated_cache failing on multiple platforms


Ch3LL:
### Description of Issue
This test `integration.states.test_virtualenv_mod.VirtualenvTest.test_issue_2594_non_invalidated_cache`

failed on multiple platforms. https://jenkinsci.saltstack.com/job/pr-fedora30-py3/job/master/621/

```
Traceback (most recent call last):
  File "/tmp/kitchen/testing/tests/integration/states/test_virtualenv_mod.py", line 122, in test_issue_2594_non_invalidated_cache
    self.assertSaltTrueReturn(ret)
  File "/tmp/kitchen/testing/tests/support/mixins.py", line 644, in assertSaltTrueReturn
    **(next(six.itervalues(ret)))
AssertionError: False is not True. Salt Comment:
virtualenv exists
Collecting zope.interface==4.0.1
  Downloading zope.interface-4.0.1.tar.gz (136 kB)
    ERROR: Command errored out with exit status 1:
     command: /tmp/salt-tests-tmpdir/issue-2594-ve/bin/python -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/tmp/pip-install-4af8gbu0/zope.interface/setup.py'"'"'; __file__='"'"'/tmp/pip-install-4af8gbu0/zope.interface/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' egg_info --egg-base /tmp/pip-install-4af8gbu0/zope.interface/pip-egg-info
         cwd: /tmp/pip-install-4af8gbu0/zope.interface/
    Complete output (5 lines):
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/tmp/pip-install-4af8gbu0/zope.interface/setup.py", line 138, in <module>
        **extra)
    NameError: name 'extra' is not defined
    ----------------------------------------
ERROR: Command errored out with exit status 1: python setup.py egg_info Check the logs for full command output.
```

PR #56336, opened 2020-03-09
Fix test_issue_2594_non_invalidated_cache test (update zope.interface)

Ch3LL:
### What does this PR do?
Setuptools removed `Features` with their latest 46 release and we are indeed installing the latest setuptools. If you look at the console output:

`Requirement already satisfied: setuptools>=21.0.0 in ./.nox/runtests-parametrized-3-coverage-true-crypto-none-transport-zeromq/lib/python3.6/site-packages (from kubernetes==3.0.0->-c requirements/static/py3.6/linux.txt (line 55)) (46.0.0)`

we are seeing this error `NameError: name 'extra' is not defined` on trying to install `zope.interace` because in their setup.py on the older version only sets `extra` if setuptools is able to be installed. Since zope tries to import Features `from setuptools import setup, Extension, Feature` but it no longer exists the setup.py install fails. This upgrades the zope version that does not fail on pip install with latest setuptools.

https://github.com/saltstack/salt/issues/56330

ping @s0undt3ch its probably a better fix to ensure we are installing a specific version of setuptools, would you agree? I can prepare another PR for sodium that ensures we are setting the same version of setuptools for each test run, but wanted to get your opinion first.

### DIFF
--- a/tests/integration/states/test_virtualenv_mod.py
+++ b/tests/integration/states/test_virtualenv_mod.py
@@ -81,9 +81,10 @@ class VirtualenvTest(ModuleCase, SaltReturnAssertsMixin):
             '    - requirements: salt://issue-2594-requirements.txt',
         ]
 
+        reqs = ['pep8==1.3.3', 'zope.interface==4.7.1']
         # Let's populate the requirements file, just pep-8 for now
         with salt.utils.files.fopen(requirements_file_path, 'a') as fhw:
-            fhw.write('pep8==1.3.3\n')
+            fhw.write(reqs[0] + '\n')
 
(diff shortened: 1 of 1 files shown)


s0undt3ch (approved):
For this particular test I don't mind locking the setuptools version, but outside the test we really shouldn't lock setuptools since when, for example, a new virtualenv is created, the lastest pip and setuptools will also be installed and we should be prepared for that.

Ch3LL:
okay i'll leave this fix as is then and not worry about locking setuptools. thanks

dwoz (approved):

waynew (approved):

Issue #56320, opened 2020-03-06
AttributeError: module 'salt.utils' has no attribute 'winapi' with state win_dns_client.dns_exists and schedule in Pillar


asmialoski:
### Description of Issue
I am using a state win_dns_client.dns_exists to setup DNS Servers in Network Interface of a Windows Server 2016. This state works perfectly when I run "salt '*' state.apply", but when I using a schedule through Pillar, the following error is reported:
```
2020-03-06 15:32:08,164 [salt.state       :322 ][ERROR   ][4036] An exception occurred in this state: Traceback (most recent call last):
  File "c:\salt\bin\lib\site-packages\salt-3000-py3.5.egg\salt\state.py", line 1981, in call
    **cdata['kwargs'])
  File "c:\salt\bin\lib\site-packages\salt-3000-py3.5.egg\salt\loader.py", line 1977, in wrapper
    return f(*args, **kwargs)
  File "c:\salt\bin\lib\site-packages\salt-3000-py3.5.egg\salt\states\win_dns_client.py", line 48, in dns_exists
    configured_list = __salt__['win_dns_client.get_dns_servers'](interface)
  File "c:\salt\bin\lib\site-packages\salt-3000-py3.5.egg\salt\modules\win_dns_client.py", line 44, in get_dns_servers
    with salt.utils.winapi.Com():
AttributeError: module 'salt.utils' has no attribute 'winapi'
```

### Setup
Schedule in Pillar:
```
schedule:
  state_dns:
    function: state.apply
    minutes: 1
```

State dns:
```
{% from "lib/tofs.jinja" import file_switch with context -%}
{% if pillar.get('dns_state', True) != False %}

{% if grains['kernel'] == 'Windows' %}
  {% if pillar.get('dns_static', False) == True %}
config_dns_servers:
  win_dns_client.dns_exists:
    - interface: Ethernet0
    - replace: True #remove any servers not in the "servers" list, defau
    - servers:
        {% for dnsserver in pillar.get('dns-servers', {}) %}
      - {{ dnsserver }}
        {% endfor %}
  {% endif %}
{% endif %}
{% endif %}
```

### Steps to Reproduce Issue
1. Create a state dns;
2. Create a schedule through Pillar;
3. Wait the schedule to run.

### Versions Report
```
SALT-MASTER:
Salt Version:
           Salt: 3000

Dependency Versions:
           cffi: 1.14.0
       cherrypy: unknown
       dateutil: Not Installed
[...]

PR #56339, opened 2020-03-09
Fix win_dns_client when used with scheduler

twangboy:
### What does this PR do?
Add missing import to `modules\win_dns_client.py`
Improve docs for a couple functions.

https://github.com/saltstack/salt/issues/56320

### Tests written?
Yes

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/modules/win_dns_client.py
+++ b/salt/modules/win_dns_client.py
@@ -12,8 +12,10 @@ import salt.utils.platform
 
 try:
     import wmi
+    import salt.utils.winapi
+    HAS_LIBS = True
 except ImportError:
-    pass
+    HAS_LIBS = False
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


salt/modules/win_dns_client.py
```diff
@@ -22,15 +24,25 @@ def __virtual__():
...
     '''
-    if salt.utils.platform.is_windows():
-        return 'win_dns_client'
-    return (False, "Module win_dns_client: module only works on Windows systems")
+    if not salt.utils.platform.is_windows():
+        return False, 'Module win_dns_client: module only works on Windows ' \
+                      'systems'
+    if not HAS_LIBS:
+        return False, 'Module win_dns_client: missing required libraries'
+    return 'win_dns_client'
```
waynew: You could at least test that the `__virtual__` function returns the correct things when it's windows (or not) and has libs (or not)

s0undt3ch (approved):

salt/modules/win_dns_client.py
twangboy: Added some tests... hope you like them

Issue #56119, opened 2020-02-10
Release notes for v3000 do not mention changes to `slspath` variable


finalduty:
### Description of Issue
[Release notes for v3000 ](https://docs.saltstack.com/en/latest/topics/releases/3000.html) do not mention changes to `slspath` variable by #55434 / #55433 , which previously returned the parent directory instead of the path to the state file itself.

This functionality appears to be covered by the `tpldir` variable, but this is not documented on https://docs.saltstack.com/en/latest/ref/states/vars.html.

Can both the v3000 release notes and the SLS Template Variable Reference be updated to make this more visible to others?

### Setup
Tree:
```
/srv/salt/users/
|-- files
|   |-- bash_profile
|   |-- bashrc
|   |-- login.defs
|   |-- root_profile
|   `-- vimrc
|-- init.sls
|-- root.sls
`-- staff.sls
```
root.sls
```
/root/.bashrc:
  file.managed:
    - source: salt://{{ slspath }}/files/bashrc
    - template: jinja
    - defaults:
        sls: {{ sls }}
```

### Steps to Reproduce Issue
1. Apply users.root state

Output:
```
----------
          ID: /root/.bashrc
    Function: file.managed
      Result: False
     Comment: Source file salt://users/root/files/bashrc not found in saltenv 'base'
     Started: 20:46:08.752499
    Duration: 2.315 ms
     Changes:   
----------
```
Replacing `slspath` with `tpldir` resolves the issue, but this is not an obvious fix.

### Versions Report
```
Salt Version:
           Salt: 3000
 
Dependency Versions:
           cffi: Not Installed
       cherrypy: Not Installed
       dateutil: Not Installed
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
         Jinja2: 2.8.1
        libgit2: Not Installed
       M2Crypto: 0.33.0
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.6.2
   mysql-python: Not Installed
      pycparser: Not Installed
       pycrypto: 2.6.1
   pycryptodome: 3.7.3
         pygit2: Not Installed
         Python: 3.6.8 (default, Aug  7 2019, 17:28:10)
   python-gnupg: Not Installed
         PyYAML: 3.11
          PyZMQ: 15.3.0
          smmap: Not Installed
        timelib: Not Installed
        Tornado: 4.5.3
            ZMQ: 4.1.4
 
System Versions:
[...]

PR #56341, opened 2020-03-10
Revert "Don't remove one directory level from slspath"

dwoz:
This reverts commit 7c28995a0a535f4db2f76db39f09754c7ceda2f5.

This change is being reverted because it broke states without going through any kind of deprecation path.

#56119

### Tests written?

Yes

### Commits signed with GPG?

Yes

### DIFF
--- a/doc/ref/states/vars.rst
+++ b/doc/ref/states/vars.rst
@@ -96,10 +96,11 @@ include option.
 slspath
 =======
 
-The `slspath` variable contains the path to the current sls file. The value
-of `slspath` in files referenced in the current sls depends on the reference
-method. For jinja includes `slspath` is the path to the current file. For
-salt includes `slspath` is the path to the included file.
+The `slspath` variable contains the path to the directory of the current sls
+file. The value of `slspath` in files referenced in the current sls depends on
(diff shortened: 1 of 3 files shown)


terminalmage:
That change should absolutely _not_ be reverted. It was a bugfix.

From the [documentation](https://docs.saltstack.com/en/latest/ref/states/vars.html#slspath):

```
The slspath variable contains the path to the current sls file. The value of slspath
in files referenced in the current sls depends on the reference method. For jinja
includes slspath is the path to the current file. For salt includes slspath is the path
to the included file.
```

"the path to the current SLS file". Not a directory.

thatch45:
Those docs should be fixed, not breaking everyone's states

thatch45 (approved):

terminalmage:
Actually, after looking at the code, I think that we should revert, but not without making other changes. See [here](https://github.com/saltstack/salt/issues/56119#issuecomment-597160470) for my reasoning on this.

twangboy (approved):

DmitryKuzmenko (approved):

Ch3LL (approved):

Issue #51868, opened 2019-02-27
Windows reg.present salt-call win_dacl.py 2019.2.0 'LookupAccountSid' Error


jeremylarose:
### Description of Issue/Question
The reg.present states work, but presents an ugly error when running salt-call from a minion running version 2019.2.0... the error is not present with 2018.3.4 or if running the state from the master... only with a salt-call from the minion.

### Setup
example sls:
```
'HKLM\SOFTWARE\Policies\Microsoft\\Windows\TabletPC\PreventHandwritingDataSharing':
  reg.present:
    - name: 'HKLM\SOFTWARE\Policies\Microsoft\Windows\TabletPC'
    - vname: PreventHandwritingDataSharing
    - vdata: 1
    - vtype: REG_DWORD
```
### Steps to Reproduce Issue
Run salt-call on multiple Windows 10 machines
```
[ERROR   ] Error resolving "S-1-1-1-1111-1111111111111-1111111111-11-111-111111-11111-1111-1111": No mapping between account names and security IDs was done.
Traceback (most recent call last):
  File "C:\salt\bin\lib\site-packages\salt\utils\win_dacl.py", line 1179, in get_name
    return win32security.LookupAccountSid(None, sid_obj)[0]
error: (1332, 'LookupAccountSid', 'No mapping between account names and security IDs was done.')
[ERROR   ] Error resolving "S-1-1-1-1111-1111111111111-1111111111-11-111-111111-11111-1111-1111": No mapping between account names and security IDs was done.
Traceback (most recent call last):
  File "C:\salt\bin\lib\site-packages\salt\utils\win_dacl.py", line 1179, in get_name
    return win32security.LookupAccountSid(None, sid_obj)[0]
error: (1332, 'LookupAccountSid', 'No mapping between account names and security IDs was done.')
```
... but the reg.present state runs and succeeds

### Versions Report

salt 2019.2.0 (Fluorine)
salt-call 2019.2.0 (Fluorine)... ERROR
salt-call 2018.3.4 (Oxygen)... no error

twangboy:
@jeremylarose Thanks for reporting. I'll take a look at this.
That GUID looks suspect though... I wonder where that came from...
That code is trying to match the GUID it received to an Account name.

jeremylarose:
Thanks, yes, i made up that id, it was a proper one

twangboy:
Oh, OK. That makes sense. I'll take a look at it.

twangboy:
[...]

PR #56351, opened 2020-03-10
Skip Capability SIDs, don't stack trace

twangboy:
### What does this PR do?
Since Capability SIDs do not resolve, we'll just skip the error. Only show the error if it's not a Capability SID.

https://github.com/saltstack/salt/issues/51868

### Previous Behavior
Any state that dealt with permissions on Windows would display a stack trace saying that it could not resolve a SID that started with `S-1-15-3`. This would be states like:
```
file.managed
reg.present
```
If any object had a Capability SID in the ACL it would throw the stack trace, though the state would generally still apply successfully.

### New Behavior
Capability SIDs are ignored.

### Tests written?
No
Not sure how to test if something is no longer showing up in the Log

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/states/win_wua.py
+++ b/salt/states/win_wua.py
@@ -183,7 +183,7 @@ def installed(name, updates=None):
     wua.install(install)
 
     # Refresh windows update info
-    wua.refresh()
+    wua.load_search()
     post_info = wua.updates().list()
 
     # Verify the installation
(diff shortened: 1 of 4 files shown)


s0undt3ch (commented):
Here's an example of asserting against logs.
https://github.com/saltstack/salt/blob/master/tests/integration/logging/test_jid_logging.py

dwoz (approved):

PR #56362, opened 2020-03-12
Fix list handling, all dict types in mongo returner

MasonEdgel:
### What does this PR do?
Updates the mongo_future_return module to handle key-fixing correctly in dicts with list values that themselves contain dicts, as well as handling OrderedDict types.

### Previous Behavior
Previously the returner would only fix nested dicts and their keys, so whenever a dict nested in a list was passed to mongo with invalid keys it would cause the return to fail.

### New Behavior
It now correctly fixes dict keys in both those embedded in lists and OrderedDict types.

### Tests written?
No (doesn't seem to be any existing tests for the mongo returner)

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/returners/mongo_future_return.py
+++ b/salt/returners/mongo_future_return.py
@@ -96,6 +96,7 @@ from __future__ import absolute_import, print_function, unicode_literals
 
 # Import python libs
 import logging
+import collections
 
 # Import Salt libs
 import salt.utils.jid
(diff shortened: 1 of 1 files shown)


dwoz:
@MasonEdgel This needs to be opened against `master` not `develop`.

sagetherage:
closing as abandoned work

Issue #56377, opened 2020-03-14
installing bdist_wheel package returns incorrect version when using minor version


Ch3LL:
### Description of Issue
When installing salt from the bdist_wheel package it returns the wrong version when the tag includes a minor version. You can quickly replicate this by creating a tag and pushing it up to your personal repo and running the bootstrap:

```
wget -O bootstrap.sh https://bootstrap.saltstack.com
sh bootstrap.sh -g https://github.com/ch3ll/salt.git -M -P git v3000.1
```

```
[root@ch3ll ~]# salt --version
salt 3000
```

The reason this is occurring is because we do not include the `_version.py` file when we build the wheel package.

PR #56378, opened 2020-03-14
 Include _version.py if building wheel 

Ch3LL:
### What does this PR do?
Ensures the `_version.py` file is included when building the bdist_wheel package. While writing tests and running tests for this found a couple more issues that this PR cleans up:

1. name is not included in repr for the new version scheme, and it includes bugfix even if we are using new version scheme:

before:

```
>>> salt.version.SaltStackVersion(3000, 1, None, None, '', 0, 0, None)
<SaltStackVersion major=3000 minor=1 bugfix=None>
```

after:

```
>>> salt.version.SaltStackVersion(3000, 1, None, None, '', 0, 0, None)
<SaltStackVersion name='Neon' major=3000 minor=1>
```

2. When adding the version to `_version.py` it uses `full_info` and then instantiates another instance of SaltStackVersion in that file using full_info. The problem with this is `full_info` will not return the bugfix, mbugfix or minor if it doesn't exist with the new versioning scheme. This leads to the pre_type, pre_num, noc, or sha then becoming the bufix version for example. This change adds a new property method `full_info_all_versions` which returns the old versioning scheme so when we instantiate a new object we return the correct major,minor,etc.

To display this:

before:

```
In [7]: first = salt.version.SaltStackVersion(3000, 1, None, None, '', 0, 0, None)        

In [8]: first.pre_type                       
Out[8]: ''

In [9]: second = salt.version.SaltStackVersion(*first.full_info)                          

In [10]: second.pre_type                     
Out[10]: 0
```

as you can see above, full_info changed what we were expecting because full_info should only return the new versioning scheme. But this causes problems when we use this `full_info` to instantiate another SaltStackVersion. So i created another propertymethod for this use case:

```
In [11]: second = salt.version.SaltStackVersion(*first.full_info_all_versions)            

In [12]: second.pre_type                     
Out[12]: ''
```

fyi @s0undt3ch thats why you were seeing this: https://github.com/saltstack/salt/pull/56358
this should also fix it so we dont get `''` as the minor type. we are now using `full_info_all_versions` during setup.py when creating `_version.py`

https://github.com/saltstack/salt/issues/56377

### Previous Behavior

```
[root@ch3ll ~]# salt --version
salt 3000
```

### New Behavior

```
[root@ch3ll ~]# salt --version
salt 3000.1
```

### Tests written?
Yes

### Commits signed with GPG?
Yes


tests/support/virtualenv.py
```diff
@@ -0,0 +1,77 @@
+# -*- coding: utf-8 -*-
```
s0undt3ch: We already have [this helper](https://github.com/saltstack/salt/blob/master/tests/support/helpers.py#L1620).

tests/support/virtualenv.py
```diff
@@ -0,0 +1,77 @@
...
+                            python_binary_names
+                        )
+                    )
+            # We're running off a virtualenv, and we don't want to create a virtualenv off of
+            # a virtualenv
+            kwargs = {'python': python}
+        except AttributeError:
+            # We're running off of the system python
+            kwargs = {}
+        self.run_function('virtualenv.create', [path], **kwargs)
```
s0undt3ch: `run_function` requires a master and minion running, last time I checked.
The existing virtualenv helper does not because it uses the `sminion` approach.

s0undt3ch (changes requested):
Let use the existing helper instead.
Since you also patch the environ, the existing helper functionality could be extended to support that too.

In order to get .1 out faster I'm cool with switching my review but we should do the right thing then after the .1 release.

s0undt3ch:
Also, great job finding out the real reason behind the empty string version minor issue.

s0undt3ch (approved):

twangboy (approved):

dwoz (approved):

cmcmarrow (approved):

PR #56379, opened 2020-03-14
Fixes AzureRM create_object_model util

Ajnbro:
### What does this PR do?
This PR fixes the Microsoft Azure Resource Manager create_object_model util. Currently, the util is not properly building some object models. The issue is caused by the util improperly handling the existence of parameters because it does not account for the existence of parameters that are set as Boolean values (specially False). 

N/A

### Previous Behavior
Object models with parameters that may have contained Boolean values (specially False) were being built improperly. 

### New Behavior
Object models with parameters that may contain Boolean values (specially False) are being built properly. 

### Tests written?
No

### Commits signed with GPG?
Yes


Ch3LL (changes requested):
can we get some tests added here? and can you create an issue with details of the issue and tie it to this PR, to make it easier for others running into the issue to see it.

nicholasmhughes:
@Ch3LL test case added. This issue was actually something that we ran into in Idem, but also could apply to Salt... we just don't have any concrete errors to report.

Ch3LL (dismissed):

dwoz:
Merge conflicts need to be resolved.

Ch3LL (approved):

Issue #52800, opened 2019-05-02
2019.2.0 some function throws in test=True mode exception "Passed invalid arguments: sequence item 0: expected string or Unicode, list found."


Reiner030:
As from Github found already some years ago thrown in https://github.com/saltstack/salt/issues/31591

All I get it this not very useful thrown error:

```
[...]
[DEBUG   ] Could not LazyLoad pkg.normalize_name: 'pkg.normalize_name' is not available.
[INFO    ] All specified packages are already installed
[INFO    ] Completed state [salt-minion] at time 16:13:45.596428 (duration_in_ms=28.239)
[INFO    ] Running state [salt-minion] at time 16:13:45.599755
[INFO    ] Executing state service.running for [salt-minion]
[INFO    ] Executing command [u'systemctl', 'is-active', u'salt-minion.service'] in directory '/root'
[DEBUG   ] stdout: active
[INFO    ] Executing command [u'systemctl', 'is-enabled', u'salt-minion.service'] in directory '/root'
[DEBUG   ] stdout: enabled
[INFO    ] The service salt-minion is already running
[INFO    ] Completed state [salt-minion] at time 16:13:45.628544 (duration_in_ms=28.789)
[INFO    ] Running state [salt-cloud] at time 16:13:45.629082
[INFO    ] Executing state pkg.installed for [salt-cloud]
[DEBUG   ] Could not LazyLoad pkg.resolve_capabilities: 'pkg.resolve_capabilities' is not available.
[DEBUG   ] Could not LazyLoad pkg.normalize_name: 'pkg.normalize_name' is not available.
[INFO    ] All specified packages are already installed
[INFO    ] Completed state [salt-cloud] at time 16:13:45.669880 (duration_in_ms=40.797)
[INFO    ] Running state [awscli] at time 16:13:45.670374
[INFO    ] Executing state pip.installed for [awscli]
[DEBUG   ] Could not LazyLoad pip.mod_run_check: 'pip.mod_run_check' is not available.
[INFO    ] Executing command 'type aws' in directory '/root'
[DEBUG   ] stdout: aws is /usr/local/bin/aws
[DEBUG   ] Last command return code: 0
[INFO    ] [u'unless condition is true']
[INFO    ] Completed state [awscli] at time 16:13:45.694873 (duration_in_ms=24.498)

Passed invalid arguments: sequence item 0: expected string or Unicode, list found.

Usage:

    .. versionadded:: 2015.5.0

    This function will call :mod:`state.highstate
[...]

Issue #54501, opened 2019-09-16
2019.2.0 - test mode crashes with both "unless:" and "retry:" declarations in a file.managed state


OrlandoArcapix:
### Description of Issue

When running a file.managed state in test mode, which is configured to download a file over https, salt-call will crash if the state has both a `retry:` and an `unless:` declaration which evaluates to true.

Command line returns:
```Passed invalid arguments: sequence item 0: expected string or Unicode, list found.```

followed by the `Usage:` statement for `salt-call`:
[crash.log](https://github.com/saltstack/salt/files/3616723/crash.log)

### Setup

Create a state `test.sls`:

```
download sample data:
  file.managed:
    - name: /tmp/saltstack.README.rst
    - retry:
        attempts: 5
        interval: 5
    - unless:
      - test -f /tmp/saltstack.README.rst
    - source:
      - https://raw.githubusercontent.com/saltstack/salt/develop/README.rst
    - source_hash: f2bc8c0aa2ae4f5bb5c2051686016b48
```

### Steps to Reproduce Issue

* `touch /tmp/saltstack.README.rst`
* `salt-call state.apply test -l info test=true 2>&1 | tee /tmp/crash.log`

### Versions Report
```
Salt Version:
           Salt: 2019.2.0

Dependency Versions:
           cffi: 1.12.3
       cherrypy: unknown
       dateutil: 2.7.2
      docker-py: 3.5.0
          gitdb: 2.0.5
      gitpython: 2.1.11
          ioflo: Not Installed
         Jinja2: 2.9.6
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: 0.31.0
           Mako: 1.1.0
   msgpack-pure: Not Installed
 msgpack-python: 0.5.6
   mysql-python: 1.2.5
      pycparser: 2.19
       pycrypto: 2.6.1
   pycryptodome: 3.7.3
         pygit2: Not Installed
         Python: 2.7.5 (default, Apr  9 2019, 14:30:50)
   python-gnupg: Not Installed
         PyYAML: 3.11
          PyZMQ: 15.3.0
           RAET: Not Installed
          smmap: 2.0.5
        timelib: Not Installed
        Tornado: 4.2.1
            ZMQ: 4.1.4

System Versions:
           dist: centos 7.6.1810 Core
         locale: UTF-8
        machine: x86_64
        release: 3.10.0-957.12.2.el7.x86_64
         system: Linux
[...]

PR #56381, opened 2020-03-15
Fix 54501

mchugh19:
### What does this PR do?
When state.py runs the _run_check function to validate unless/onlyif, the resulting output can include a list as a comment:
```
{'result': True, 'comment': ['unless condition is true'], 'skip_watch': True}
```
This list was breaking the state output when test=True with retry, which assumed any existing comments would already be strings. This PR add `six.text_type` casting to to string formatting to support other input types.

#54501

### Previous Behavior
Running a state with included unless and retry with `test=True` would result in crash:
```
         File "/root/gits/salt/tests/unit/test_state.py", line 265, in test_verify_retry_parsing
           self.assertDictContainsSubset(expected_result, state_obj.call(low_data))
         File "/root/gits/salt/salt/utils/decorators/state.py", line 30, in _func
           result = func(*args, **kwargs)
         File "/root/gits/salt/salt/state.py", line 2084, in call
           low['retry']['splay'])])
       TypeError: sequence item 0: expected str instance, list found
```

### New Behavior
Output is rendered as expected:
```
['unless condition is true']  The state would be retried every 5 'seconds (with a splay of up to 0 seconds) a maximum of 5 times or until a result of True is returned
```

### Tests written?
Yes

### Commits signed with GPG?
No


mchugh19:
Possibly fixes #52800 as well. An example state was not supplied for that issue, but the line referenced in the exception is the same issue as #54501.

DmitryKuzmenko (dismissed):

DmitryKuzmenko:
@mchugh19 could you please resolve merge conflicts?

dwoz (approved):

PR #56382, opened 2020-03-15
doc: Update kiwi url

rst0git:
### What does this PR do?
Updates an URL for the KIWI Image System in the docs.


waynew (approved):

waynew (changes requested):
Whoops, clicked that too soon - the man page is built from the docs, so no need to update the links there.

waynew (approved):

PR #56394, opened 2020-03-17
Add IP filtering by network

agraul:
#### Add IP filtering by network

IPs are filtered out if they don't belong to any of the given networks.
If no network is passed, all items are returned.

Example:

	{% set networks = ['192.168.0.0/24', 'fe80::/64'] %}
	{{ grains['ip_interfaces'] | filter_by_networks(networks) }}
	{{ grains['ipv6'] | filter_by_networks(networks) }}
	{{ grains['ipv4'] | filter_by_networks(networks) }}

### Tests written?

Not yet, I will do it next.

### Commits signed with GPG?

Yes


salt/utils/network.py
```diff
@@ -1958,3 +1959,45 @@ def parse_host_port(host_port):
...
+    '''
+    filtered_addresses = []
+
+    for value in values:
+        address = ipaddress.ip_address(value)
+        for network in networks:
+            if address in network:
+                filtered_addresses.append(value)
+
+    return filtered_addresses
```
aplanas: All this can be replaced with

```python
return [i for n in networks for i in values if ipaddress.ip_address(i) in n]
```

You can put that in a`lambda` inside the main function (or a inner function)

salt/utils/network.py
```diff
@@ -1958,3 +1959,45 @@ def parse_host_port(host_port):
...
+    {{ grains['ipv6'] | filter_by_networks(networks) }}
+    {{ grains['ipv4'] | filter_by_networks(networks) }}
+    '''
+    if filters:
+        networks = [ipaddress.ip_network(network) for network in filters]
+        new_values = None
+        if isinstance(values, collections.Mapping):
+            new_values = {}
+            for interface in values:
+                new_values[interface] = _filter_by_networks(values[interface], networks)
```
aplanas: there is also a dict comprehension expression that can collapse the full `Mapping` case

salt/utils/network.py
```diff
@@ -1958,3 +1959,45 @@ def parse_host_port(host_port):
...
+    '''
+    Returns a the list of IPs filtered by the optional network list.
+    If no filter is specified all all items are returned.
+
+    {% set networks = ['192.168.0.0/24', 'fe80::/64'] %}
+    {{ grains['ip_interfaces'] | filter_by_networks(networks) }}
+    {{ grains['ipv6'] | filter_by_networks(networks) }}
+    {{ grains['ipv4'] | filter_by_networks(networks) }}
+    '''
+    if filters:
```
aplanas: I would not make filters optional, as this is counter-intuitive. For example if `filters=[]`, the empty array, there is not possible match, and we would expect to return also the empty array. `None` and `[]` are semantically similar, but in this case would to just the opposite: will return the full set of values.

salt/utils/network.py
```diff
@@ -1958,3 +1959,45 @@ def parse_host_port(host_port):
...
+        address = ipaddress.ip_address(value)
+        for network in networks:
+            if address in network:
+                filtered_addresses.append(value)
+
+    return filtered_addresses
+
+
+@jinja_filter('filter_by_networks')
+def filter_by_networks(values, filters=None):
```
aplanas: Nit, maybe is better to rename `values` as `addresses`, and `filters` as `networks`

salt/utils/network.py
darix: For me the loop is more readable than the list comprehension expression.

salt/utils/network.py
darix: This makes code with filters loaded from a pillar easier. "if no filter in pillar gimme everything" otherwise you would need to stick a default filter into the pillar.

salt/utils/network.py
agraul: I think a list comprehension like this is both readable and simpler:
```python
    return [ip for ip in ips
               for net in networks
               if ipaddress.ip_address(ip) in net]
```
If I don't overlook something the runtime of this is `len(ips) * len(networks)` in either case, so switching the loops (compared to @aplanas suggestion) does not matter.

aplanas: @agraul the complexity is the same, yes. Choose whatever you want : )

salt/utils/network.py
agraul: We could check `if filters is None` to treat the `[]` case differently, but I don't know if that is the best approach. Passing an explicit default¹ as the filter conveys intention better than leaving the filter out in my opinion.

¹ such as `['0.0.0.0/0', '::/0']`

salt/utils/network.py
agraul: I agree, I'll update the names

salt/utils/network.py
aplanas: I see, but do not make the filter parameter optional. By default the pillar data is not optional (KeyError exception), and to make it optional you need to provide an explicit fallback value: `pillar.get('networks', None)`

I would expect that an empty list of networks will produce no match, but I guess that is OK to go to the opposite semantic if is documented.

salt/utils/network.py
```diff
@@ -1958,3 +1959,33 @@ def parse_host_port(host_port):
             raise ValueError('bad hostname: "{}"'.format(host))
 
     return host, port
+
+
+@jinja_filter("filter_by_networks")
+def filter_by_networks(values, networks=None):
```
aplanas: Very cool changes. Still do not make this parameter optional. The argument about the pillar was discussed in another comment. IMHO this parameter is no optional.

salt/utils/network.py
```diff
@@ -1958,3 +1959,33 @@ def parse_host_port(host_port):
...
+    """
+    Returns a the list of IPs filtered by the optional network list.
+    If no filter is specified all all items are returned.
```
aplanas: Maybe to fulfill @darix user case and the current version of the code (where an `is not None` make it explicit), we need to document that for None all items will be returned, and for an empty list of networks, no match is found.

I still find this weird, but I guess that is OK.

tests/unit/utils/test_network.py
```diff
@@ -701,3 +701,15 @@ def test_generate_minion_id_with_long_hostname(self):
             # An exception is raised if unicode is passed to socket.getfqdn
             minion_id = network.generate_minion_id()
         assert minion_id != '', minion_id
+
+    def test_filter_by_network(self):
```
aplanas: In different tests, test the corner cases (empty, None). I would split the tests more, one for each case (dict, no dict for IPs,, None, empty list, list with matches, list without matches for network) Is more work, but each test will be like 2 lines and you will be super sure that all cases are covered, for whatever semantic you choose.

salt/utils/network.py
agraul: >  do not make the filter parameter optional. By default the pillar data is not optional (KeyError exception)

okay, in that case I don't see a reason not to require a filter (now `networks`) parameter.

About the sematics of `networks=None` vs `networks=[]`: I definitely think that `[]` should return `[]`, but I am not sure about `None`. I am leaning towards treating it the same as `[]`, but I can see the argument that `None` is the absence of filtering.

tests/unit/utils/test_network.py
agraul: I looked at the `test_is_ipv(4|6)` as a reference which do it like this, but I agree that it is better to split them up.

salt/utils/network.py
agraul: Yes, I just read your comment about that after I did the changes. I'll have it required.

salt/utils/network.py
aplanas: Yes. The semantic confuses me. For me `[]` and `None` are the same, so I would expect the same return value: `[]`.

If we make it different, we need to document it, and we cannot change it later. @darix argument is not very convincing for me, as you can be explicit in the SLS and an `{% if %}` to apply the filter. But I can see how this is more verbose. Is OK to maintain the semantic where both a different, and see if another review have a stronger opinion.

tests/unit/utils/test_network.py
aplanas: Right. Do not use the current tests as a reference for good behavior : ( The plan is try to do the right thing.

For example, the new code is moving out of `unittest` and moving to `pytest`. This means that instead of `self.assertEqual(a, b)` the code will be `assert a == b`. But you will see tons of code using the first one (I still write it like that, to be honest, that is why I didn't complain in this code)

aplanas (dismissed):

tests/unit/utils/test_network.py
```diff
@@ -701,3 +701,50 @@ def test_generate_minion_id_with_long_hostname(self):
             # An exception is raised if unicode is passed to socket.getfqdn
             minion_id = network.generate_minion_id()
         assert minion_id != '', minion_id
+
+    def test_filter_by_networks_with_no_filter(self):
+        ips = ["10.0.123.200", "10.10.10.10"]
+        self.assertRaises(TypeError, network.filter_by_networks, ips)
```
agraul: I've not used `with pytest.raises` here because I was not sure if it is ok to "`import pytest` or skip"¹ this test. Although it just came to my mind that if `assert a == b` requires `pytest`, it would make sense to skip these all tests if `pytest` is unavailable.

¹ like it is done [here](https://github.com/saltstack/salt/blob/master/tests/unit/utils/test_thin.py#L24)

aplanas: This try is two years old. I would assume pytest available.

salt/utils/network.py
agraul: I've updated the docstring according to what the code does at the moment. Let's discuss the behavior of `None` vs `[]` in the other thread.

salt/utils/network.py
```diff
@@ -1958,3 +1959,34 @@ def parse_host_port(host_port):
             raise ValueError('bad hostname: "{}"'.format(host))
 
     return host, port
+
+
+@jinja_filter("filter_by_networks")
+def filter_by_networks(values, networks):
+    """
+    Returns a the list of IPs filtered by the network list.
```
DmitryKuzmenko, line 1967: Typo: `Returns a the list...`.

DmitryKuzmenko (changes requested):

salt/utils/network.py
agraul: Right, that needs to be fixed. Did you have a look at the [discussion](https://github.com/saltstack/salt/pull/56394#discussion_r394536151) about an empty list vs None ? If we decide to change the meaning (from the current implementation), I need to update the docstring anyway.

salt/utils/network.py
DmitryKuzmenko: My opinion on this is the following: I see no need for the filters argument to be optional since if someone uses the filter they definitely wants to filter by some condition. The same is about empty filter values like `None` or `[]`. So since it's documented I'm okay with this implementation.
@saltstack/team-core comments on this?

salt/utils/network.py
DmitryKuzmenko: @agraul I've just left a comment on that discussion.

DmitryKuzmenko (approved):

dwoz:
I've re-started the failing windows build. When it passes this will get merged.

DmitryKuzmenko:
Just re-started it once more because it timed out again.

dwoz (approved):

Issue #54702, opened 2019-09-20
__virtual__ returned False is cryptic


ksalman:
The error is cryptic while it just needed python3-mysqldb package. So `apt-get install -y python3-mysqldb` fixed it. 

# Setup
```
vlab local user:
  mysql_user.present:
    - name: user
    - host: localhost
    - allow_passwordless: True
    - require:
      - pkg: mysql-server
```

### Steps to Reproduce Issue
(Include debug logs if possible and relevant.)
Tried to create user with the example pasted above and got this error
```
<snip>
[DEBUG   ] Could not LazyLoad mysql_user.present: 'mysql_user' __virtual__ returned False
[INFO    ] Running state [vlab] at time 08:06:37.711519
[DEBUG   ] LazyLoaded mysql_user.present
[ERROR   ] State 'mysql_user.present' was not found in SLS 'roles.vlab.database'
Reason: 'mysql_user' __virtual__ returned False
<snip>
----------
          ID: vlab local user
    Function: mysql_user.present
        Name: user
      Result: False
     Comment: State 'mysql_user.present' was not found in SLS 'roles.vlab.database'
              Reason: 'mysql_user' __virtual__ returned False
     Changes:
```

### Versions Report
salt-minion --versions
Salt Version:
           Salt: 2019.2.0

Dependency Versions:
           cffi: Not Installed
       cherrypy: Not Installed
       dateutil: 2.4.2
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.8
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.4.6
   mysql-python: 1.3.7
      pycparser: Not Installed
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: Not Installed
         Python: 3.5.2 (default, Nov 12 2018, 13:43:14)
   python-gnupg: 0.3.8
         PyYAML: 3.11
          PyZMQ: 15.2.0
           RAET: Not Installed
          smmap: Not Installed
        timelib: Not Installed
        Tornado: 4.2.1
            ZMQ: 4.1.4

System Versions:
           dist: Ubuntu 16.04 xenial
         locale: UTF-8
        machine: x86_64
        release: 4.4.0-21-generic
         system: Linux
        version: Ubuntu 16.04 xenial

[...]

PR #56395, opened 2020-03-17
Add __virtual__ fail text

mchugh19:
### What does this PR do?
Add __virtual__ fail text for execution and state modules. 

Fixes wording for #54702

### Previous Behavior
Various modules and states logged `False` without explanation.

### New Behavior
For most modules and states, a reason is logged as to why they are not loaded.

### Tests written?
No
Generally a doc fix. 
Unless salt were to standardize on how messages should be logged, there isn't a single method suitable for testing.

### Commits signed with GPG?
No


mchugh19:
Ugh! Don't know why I started fixing PRs from the black merge with the one that touched all the files. Grrr!

waynew (approved):

waynew:
Maybe you like pain? :joy: 

Just re-ran the couple of failing tests, I think they just failed b/c we're using aws spot instances. I believe we're putting in a change to fix that, but this *should* run.

waynew:
@Ch3LL all success here 🚀

PR #56396, opened 2020-03-17
Master fix aptpkg systemd call

brejoc:
### What does this PR do?

Sets a `--description` for the systemd-run command called by aptpkg to avoid long description strings that would cause crashes.
The description comes from `__name__` and it would be `salt.modules.aptpkg`

Fixes: aptpkg module causes systemd to crash in Ubuntu 18.04.
When no description is provided to systemd-run, the description is autogenerated from the apt-get command.
If we have a long list of packages the description will be longer than what systemd-run supports and it will crash (https://github.com/systemd/systemd/commit/a7419dbc59da5c8cc9e90b3d96bc947cad91ae16)

Setting the description ourselves prevents this even if the systemd doesn't have the fix.

### Previous Behavior

See above

### New Behavior

Sets the description to `salt.modules.aptpkg`

### Tests written?

Yes


brejoc:
This PR replaces https://github.com/saltstack/salt/pull/54077.

cmcmarrow (dismissed):

brejoc:
I've just merged master again. Anything else that needs to be done?

DmitryKuzmenko (dismissed):

DmitryKuzmenko:
@brejoc could you please resolve merge conflicts?

brejoc:
@DmitryKuzmenko PR now complies with black formatting.

DmitryKuzmenko (approved):

Issue #50422, opened 2018-11-07
It's impossible to restart salt-minion from the master on MacOS minions


V3XATI0N:
### Description of Issue/Question
A command like `salt MINION service.restart salt-minion` works fine on Linux, Windows, and even Solaris. But on MacOS, it just kills the service. There's no apparent workaround either, because any process (including a local script that would start the service) started by the minion dies with the main process when it restarts, killing any would-be restart commands anywhere.

Also, I can't use `at` because every attempt for any command in the `at` module just results in a dump and a `list index out of range` error from the minion.

weswhet:
Looks like service.restart on macOS does a stop/start. Which is an issue because when it the stops the service, it kills the process thats running the command so its unable to start it again. Looks like we should be able to change this to run `launchctl kickstart -k system/com.saltstack.salt.minion` and we should be good to go.

ThisGuyCodes:
bump.

Seems like the fix for this is already known (to the point that people are doing it themselves.

PR #56400, opened 2020-03-17
Use `kickstart` on macos instead of stop + start

thisguycodes:
### What does this PR do?
Changes `service.restart` on MacOS to use `launchctl kickstart -k` instead of `launchctl stop` + `launchctl start`

This is more atomic as both actions are handled by the OS, rather than potentially having the salt-minion die mid-restart (such as restarting itself).

#50422 

### Previous Behavior
Salt-minion couldn't restart itself, as stop + start would cause it to die before the start.

### New Behavior
Salt-minion can restart itself.

### Tests written?
No.

This is strictly a refactor of existing behavior, I believe existing tests cover this.

...I suppose a test specifically covering salt-minion restarting itself would be prudent. I'll look into this, help / directional advice appreciated!

### Commits signed with GPG?

No (I'll force push with signed commit(s) in a bit).

---
This change is [Reviewable](https://reviewable.io/reviews/saltstack/salt/56400)


ThisGuyCodes:
I'm reading the contributor guidelines now and will make changes as necessary. I just figured this might be pretty easy to fix and jumped on it!

ThisGuyCodes:
I'm actually not clear on what the process is here: do I need to gpg sign my commits?

dwoz:
@ThisGuyCodes Signing you commits is recommended but not required. The fastest way to get this merged in will be to add a test case to cover the change.

waynew:
@ThisGuyCodes thanks for the PR! Are you able to write some tests for this? An integration test might be the most meaningful, but I think you could also write some decently meaningful unit tests.

waynew:
@ThisGuyCodes following up again - would you be able to add some tests here? If you're unsure about writing tests, I'm running a test clinic tomorrow AM American times on our [Twitch stream](https://www.twitch.tv/saltstackinc) where I'll be talking about and writing tests.

waynew (changes requested):
If you need help writing tests, ping me in #salt on freenode or Slack, or attend one of our Test Clinics!

dwoz:

Issue #56328, opened 2020-03-09
mismatch function argument between cassandra_cql and cassandra_cql_return


GwiYeong:
### Description of Issue
there is mismatch keyword argument between `cassandra_cql` and `cassandra_cql_return`

the function [cql_query_with_prepare](https://github.com/saltstack/salt/blob/master/salt/modules/cassandra_cql.py#L365) is defined like below

```python
def cql_query_with_prepare(query, statement_name, statement_arguments, callback_errors=None, contact_points=None, port=None, cql_user=None, cql_pass=None, **kwargs):

    # Backward-compatibility with Python 3.7: "async" is a reserved word
    asynchronous = kwargs.get('async', False)
```

the first line of the function gets `async` param from `**kwargs`. 

however the returner, `cassandra_cql_return` calls this function with argument that is named `asynchronous`.

See the below code line.

https://github.com/saltstack/salt/blob/master/salt/returners/cassandra_cql_return.py#L205

https://github.com/saltstack/salt/blob/master/salt/returners/cassandra_cql_return.py#L226

https://github.com/saltstack/salt/blob/master/salt/returners/cassandra_cql_return.py#L268

https://github.com/saltstack/salt/blob/master/salt/returners/cassandra_cql_return.py#L298

these functions call `cassandra_cql.cql_query_with_prepare` with `asynchronous=True`, but actually `cql_query_with_prepare` not work asynchronous

maybe missing in this commit https://github.com/saltstack/salt/commit/377fbbc7b760bad1fd267747bf463e7cc8f3c1ae

### Setup
None

### Steps to Reproduce Issue
just see code.

### Versions Report
master branch

Ch3LL:
nice catch. seems you have a handle here, would you mind sending in a PR with a test case as well?

GwiYeong:
@Ch3LL 
thanks to response. Ok, I will.
should I send pr to the `master` branch?

GwiYeong:
@Ch3LL hi I found this commit 377fbbc7b760bad1fd267747bf463e7cc8f3c1ae. I could not understand this commit. 
since python 3.7, `async` is reserved keyword, so use `asynchronous` for function args is correct. 
mayby this issues will fixed after rollback 377fbbc7b760bad1fd267747bf463e7cc8f3c1ae commit
can you give me a comment?

Ch3LL:
yep master. looks like you got that right :) i'll leave comments on your PR. thanks for this!

PR #56404, opened 2020-03-18
Fix mismatch function argument between cassandra_cql and cassandra_cql_return

GwiYeong:
### What does this PR do?
there is mismatch keyword argument between `cassandra_cql` and `cassandra_cql_return`

the function [cql_query_with_prepare](https://github.com/saltstack/salt/blob/master/salt/modules/cassandra_cql.py#L365) is defined like below

```python
def cql_query_with_prepare(query, statement_name, statement_arguments, callback_errors=None, contact_points=None, port=None, cql_user=None, cql_pass=None, **kwargs):

    # Backward-compatibility with Python 3.7: "async" is a reserved word
    asynchronous = kwargs.get('async', False)
```

the first line of the function gets `async` param from `**kwargs`. 

however the returner, `cassandra_cql_return` calls this function with argument that is named `asynchronous`.

See the below code line.

https://github.com/saltstack/salt/blob/master/salt/returners/cassandra_cql_return.py#L205
https://github.com/saltstack/salt/blob/master/salt/returners/cassandra_cql_return.py#L226
https://github.com/saltstack/salt/blob/master/salt/returners/cassandra_cql_return.py#L268
https://github.com/saltstack/salt/blob/master/salt/returners/cassandra_cql_return.py#L298

these functions call `cassandra_cql.cql_query_with_prepare` with `asynchronous=True`, but actually `cql_query_with_prepare` not work asynchronous

I found this commit 377fbbc7b760bad1fd267747bf463e7cc8f3c1ae. I could not understand this commit. 
since python 3.7, `async` is reserved keyword, so use `asynchronous` for function args looks like correct.  but this commit remove `asynchronous` arg.
therefore I revert that commit and write test case for this bug.

#56328

### Previous Behavior
below code work synchronously
```
__salt__['cassandra_cql.cql_query_with_prepare'](query,
                                                         'returner_return',
                                                         tuple(statement_arguments),
                                                         asynchronous=True)
```

### New Behavior
above code work asynchronously

### Tests written?
Yes

### Commits signed with GPG?

Yes


salt/modules/cassandra_cql.py
```diff
@@ -407,8 +411,13 @@ def cql_query_with_prepare(query, statement_name, statement_arguments, callback_
         salt this-node cassandra_cql.cql_query_with_prepare "name_select" "SELECT * FROM USERS WHERE first_name=?" \
             statement_arguments=['John']
     '''
-    # Backward-compatibility with Python 3.7: "async" is a reserved word
-    asynchronous = kwargs.get('async', False)
+    # Backward-compatibility. This should be removed in Sodium version
+    if 'async' in kwargs:
+        asynchronous = bool(kwargs['async'])
+        msg = 'Use of "async" parameter detected. The "asynchronous" parameter should be used instead.'
```
Ch3LL: I dont think we should deprecate async, since this is such a small change. So we can ensure users dont need to update their states. Can we remove the deprecation messages?

tests/unit/modules/test_cassandra_cql.py
```diff
@@ -79,3 +79,19 @@ def test_unspecified_opts(self):
...
+            'cassandra_cql_returner_cluster': MagicMock(return_value={}),
+            'cassandra_cql_returner_session': MagicMock(execute=mock_execute,
+                                                        execute_async=mock_execute_async,
+                                                        prepare=lambda _: MagicMock(bind=lambda _: None),  # mock prepared_statement
+                                                        row_factory=None),
+            'cassandra_cql_prepared': {}
+        }
+
+        with patch.dict(cassandra_cql.__context__, mock_context):
+            cassandra_cql.cql_query_with_prepare('SELECT now() from system.local;', 'select_now', [], asynchronous=True)
```
Ch3LL: can we also ensure `async` works as well here?

Ch3LL (changes requested):

Ch3LL:
also looks like you have a legit test failure as well

salt/modules/cassandra_cql.py
GwiYeong: ok, this message removed

tests/unit/modules/test_cassandra_cql.py
GwiYeong: I add test for ensuring `async` works. 
but since python 3.7, async is reserved keyword so I have no choice but to pass async param as dict.

GwiYeong:
I don't know why the test is failed.
this is the failure log.
```
  File "/tmp/kitchen/testing/tests/unit/modules/test_cassandra_cql.py", line 96, in test_valid_async_args
    cassandra_cql.cql_query_with_prepare('SELECT now() from system.local;', 'select_now', [], asynchronous=True)
  File "/tmp/kitchen/testing/salt/modules/cassandra_cql.py", line 441, in cql_query_with_prepare
    session.row_factory = dict_factory
NameError: global name 'dict_factory' is not defined
```
`dict_factory` is imported from `cassandra.query`.  this is prerequisite for import `cassandra_cql` module, so if import error occurred for dict_factory, this test should not be run. 
how can I fix this import error..?
In my local environments, this test is working fine.

GwiYeong:
@Ch3LL 
I tried to mock `from cassandra.query import dict_factory` but.. i don't know how to mock this statement. could you upgrade `cassandra-driver` version in the test environment?
`dict_factory` is moved from cassandra.decoder to cassandra.query. maybe `cassandra-driver` in the test environment is 1.x.
this is document about `dict_factory`. [doc](https://docs.datastax.com/en/developer/python-driver/3.21/api/cassandra/query/)
```
cassandra.query.dict_factory (colnames, rows)
Returns each row as a dict.

Example:

>>> from cassandra.query import dict_factory
>>> session = cluster.connect('mykeyspace')
>>> session.row_factory = dict_factory
>>> rows = session.execute("SELECT name, age FROM users LIMIT 1")
>>> print rows[0]
{u'age': 42, u'name': u'Bob'}
Changed in version 2.0.0: moved from cassandra.decoder to cassandra.query
```

Ch3LL:
looks like your right. I pulled down your pr and was able to run the tests just fine. Weirdly I'm having a hard time finding out how we are installing that requirement. I dont see cassandra in our requirements directory or in our https://github.com/saltstack/salt-jenkins repo which is where we install system dependencies. 

ping @s0undt3ch any ideas on how cassandra is even getting installed? I'm seeing the tests pass on jenkins: https://jenkinsci.saltstack.com/job/pr-centos7-py3/job/master/813/testReport/unit.modules.test_cassandra_cql/CassandraCQLReturnerTestCase/  so they aren't getting skipped and the test does try to import cassandra so this package is getting installed somehow. 

I think we should add this to our requirements directly instead, but still want to understand first how this is working currently.

s0undt3ch:
I see nowhere `cassandra-driver` being installed. Might we be mocking the module?

s0undt3ch:
@gwiyeong, how are you running tests?

GwiYeong:
@s0undt3ch 
I didn't do anything now... I don't know how cassandra is installed in test environments...
should I mock all of modules in cassandra?

s0undt3ch:
The jenkins builds are gone now.
There's also conflicts that need to be solved.
Could you fix those and push so we can get another set of builds please?

GwiYeong:
@s0undt3ch 
I just fixed conflict from master. these conflicts are just code formatting.

Ch3LL:
@s0undt3ch i dont believe we are mocking the module. When i tested it it skips the test since the module is not installed. Even when i build the centos-py3 vm that jenkins uses as well. I also thought maybe the package was getting installed in an earlier test but i do not see that happening as well.

Lets try this @GwiYeong -> Can you add the cassandra requirement in this file: `requirements/static/linux.in` and when you run pre-commit it will run the tool that will compile all the requirements files for the tests.

s0undt3ch:
My guess is that @GwiYeong has cassandra already installed?
Ah, wait, `AttributeError: <module 'cassandra' from '/tmp/kitchen/testing/salt/modules/cassandra.py'> does not have the attribute 'query'`, rename `salt/modules/cassandra.py` to ``salt/modules/cassandra_mod.py`, on `__virtualname__` set it to `cassandra`, cleanup any `__pycache__` directories and `.pyc` files and run tests again.
That will fix it.

GwiYeong:
@s0undt3ch oh you are right. I also think `from cassandra import query` is try to import from`salt/modules/cassandra` instends of cassandra module from `cassandra-driver`.  so `salt/modules/cassandra` doesn't have `query module`, test is failed... 
`salt/modules/cassandra` is use pycassa that is maybe deprecated because [documents](https://github.com/pycassa/pycassa) says `it is highly recommended that you use the newer DataStax python driver instead of pycassa`

hmm. i changed **virtualname** of `salt/modules/cassandra` to `cassandra_mod`. is it ok..? no one use `salt/modules/cassandra`?? 

there is already tests for `salt/modules/cassandra`.. if this module doesn't use any more, you can remove this module and test cases from salt.

@Ch3LL  i added cassandra-driver in `salt/requirements/linux.in` for running cassandra test.

Ch3LL:
@s0undt3ch nice find.

@GwiYeong do you have pre-commit setup? When you edit the requirements/static files it should generate the appropriate appropriate requirements/static/py3{5,6,7} files for the tests. If you dont want to run pre-commit the diff is included in the jenkins output here: https://jenkinsci.saltstack.com/job/pr-pre-commit/job/PR-56404/7/console

GwiYeong:
@Ch3LL 
I'm sorry. I missed that you said `when you run pre-commit it will run the tool that will compile all the requirements files for the tests.` 
but I don't know what is pre-commit setup means?  how to run pre-commit?
I would appreciate it if you let me know
thanks

Ch3LL:
Heres the documentation on how to setup pre-commit: https://docs.saltstack.com/en/latest/topics/development/contributing.html#quickstart  pre-commit will run a couple of different tools against your code when you run `git commit`. One of those tools checks to see if any requirement files have been changed and generates the needed files to add those dependencies for the tests.

If you dont want to set up pre-commit you can copy/paste the diff from the jenkins output and apply that instead.

GwiYeong:
@Ch3LL 
I just added the diff from jenkins output instead of running pre-commit.

thanks for your help

salt/modules/cassandra.py
```diff
@@ -43,7 +43,7 @@ def __virtual__():
         )
 
     if HAS_PYCASSA and salt.utils.path.which("nodetool"):
-        return "cassandra"
+        return "cassandra_mod"
```
Ch3LL, line 46: from @s0undt3ch 's comment:  "rename salt/modules/cassandra.py to salt/modules/cassandra_mod.py, on `__virtualname__` set it to cassandra"

So we will need to rename the file and set `__virtualname__` to cassandra.

Also it looks like there are some mac/windows failures.

Ch3LL (changes requested):

salt/modules/cassandra.py
GwiYeong: i fixed this.

xeacott:
Im gunna give these a fresh run. Everything else looks good here.

xeacott:
@GwiYeong Anything else this needs? Otherwise i'll mark as merge ready.

s0undt3ch (dismissed):

GwiYeong:
@xeacott nothing to do anymore. thanks.

dwoz (approved):

Issue #56415, opened 2020-03-19
3000.1 Windows Packages return saltversion 3000


frogunder:
The Windows packages for 3000.1 returns saltversion 3000

PR #56418, opened 2020-03-19
Ensure version.py included before we install

Ch3LL:
### What does this PR do?
Ensures _version.py included before we run install when running `python setup.py install`

https://github.com/saltstack/salt/issues/56415

### Previous Behavior
`_version.py` would not be included in the build if you run `python setup.py install`. This is because setuptools does not call `run_command('build')`, whereas distutils did when we did an install and we recently migrated to setuptools.

### New Behavior
`_version.py` included and `salt --version` reporting correctly.

### Tests written?
Yes

### Commits signed with GPG?

Yes

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -745,6 +745,12 @@ class Install(install):
             self.distribution.salt_download_windows_dlls = True
             self.run_command('download-windows-dlls')
             self.distribution.salt_download_windows_dlls = None
+        # need to ensure _version.py is created in build dir before install
+        if not os.path.exists(os.path.join(self.build_lib)):
+            if not self.skip_build:
+                self.run_command('build')
+        else:
+            self.run_command('write_salt_version')
         # Run install.run
         install.run(self)
 
--- a/tests/integration/setup/test_egg.py
+++ b/tests/integration/setup/test_egg.py
@@ -28,6 +28,22 @@ class EggSetupTest(ModuleCase):
     '''
     Tests for building and installing egg packages
     '''
+
+    def setUp(self):
+        # ensure we have a clean build dir
+        self._clean_build()
+
+    def _clean_build(self):
+        '''
+        helper method to clean the build dir
+        '''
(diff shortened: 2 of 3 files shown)


s0undt3ch (approved):

cmcmarrow (approved):

dwoz (approved):

markuskramerIgitt:
@Ch3LL, I suspect another root cause. 

[salt/version.py](https://github.com/saltstack/salt/blob/e1cf6ec7dbce2a56fb3b05dc0e37cf28bfa58c71/salt/version.py#L567) contains "tag describe":

    git describe --tags --first-parent --match v[0-9]* --always

I think this was appropriate for the abandoned "develop" / multi releases git scheme.

Problem with the current "master" git scheme is that "tag describe" returns from head

    v3000_docs-33-g4249d7f1e1

This is maybe is the **last tag**, but it should be **3001**, as @frogunder noted 26 days ago.

When I execute simply `git describe` (on git head master), I get    

    v3000.1-206-g4249d7f1e1


When I build the Windows installer (on git head master) with the current "tag describe" in salt/version.py I read

    build :: Install Current Version of salt...
    ---------------------------------------------------------------------
    3000-33-g4249d7f1e1

When I build the Windows installer with modified  and simple `git describe` in salt/version.py I read

    build :: Install Current Version of salt...
    ---------------------------------------------------------------------
    3000.1-206-g4249d7f1e1

Though I fail to understand where the file content for `_version.py` comes from, I assume it is "git describe" and would I suggest that version.py uses it too.

I wrote to @twangboy on Slack before.

Ch3LL:
@markuskramerIgitt would you mind opening a new issue with these details. This is breaking because we have never before used a `<version>_docs` tag.

markuskramerIgitt:
Hi @Ch3LL , I opened https://github.com/saltstack/salt/issues/56634

By the way, I looked for and found other `<version>_docs` tags.

PR #56419, opened 2020-03-20
Add setting for forcing refresh in zypper

agraul:
### What does this PR do?
Currently `salt '*' pkg.refresh_db` will always call `zypper refresh` with `--force` when `zypper` is used as the backend. With this PR, this behavior can be controlled. It leaves the default as it is, but allows a CLI flag and/or a pillar entry to control if a repository is refreshed with `--force`.

### Previous Behavior
`salt '*' pkg.refresh_db` -> `zypper refresh --force`

### New Behavior
`salt '*' pkg.refresh_db` -> `zypper refresh --force`
`salt '*' pkg.refresh_db force=true` -> `zypper refresh --force`
`salt '*' pkg.refresh_db force=false` -> `zypper refresh`

Alternatively, a pillar setting can be used (which is overridden by the CLI option):
```yml
zypper:
  refreshdb_force: false
```

### Tests written?

Yes

### Commits signed with GPG?

Yes


brejoc (approved):
~~lgtm!~~ 

Sorry, had to revert my initial review.

salt/modules/zypperpkg.py
```diff
@@ -1171,22 +1171,39 @@ def mod_repo(repo, **kwargs):
...
     # Remove rtag file to keep multiple refreshes from happening in pkg states
     salt.utils.pkg.clear_rtag(__opts__)
     ret = {}
-    out = __zypper__.refreshable.call('refresh', '--force')
+    refresh_opts = ['refresh']
+    if force is None:
+        force = __pillar__.get('zypper', {}).get('refreshdb_force', True)
+    if force:
+        refresh_opts.append('--force')
+    out = __zypper__(root=root).refreshable.call(*refresh_opts)
```
brejoc: Oh, wait a sec. I think the root kwarg is not yet introducted here. Could you check that please?

brejoc (changes requested):

salt/modules/zypperpkg.py
agraul: You are correct, that will be introduced with https://github.com/saltstack/salt/pull/54954

brejoc (dismissed):
:+1:

brejoc (approved):

dwoz (approved):

Issue #56425, opened 2020-03-22
rest_cherrypy's POST /keys does not allow selection of keysize


johnnybubonic:
### Description of Issue
[The documentation for /keys POST method](https://docs.saltstack.com/en/latest/ref/netapi/all/salt.netapi.rest_cherrypy.html#salt.netapi.rest_cherrypy.app.Keys.POST) states that:

> Accepts all the same parameters as the [`key.gen_accept`](https://docs.saltstack.com/en/latest/ref/wheel/all/salt.wheel.key.html#salt.wheel.key.gen_accept).

However, it seems to be impossible to specify a keysize larger than 2048.

Here's a snippet I'm using:

```python
import requests

user = 'someuser'
password = 'somepasswd'

api_uri = 'https://my_api.tld'
api = requests.Session()
auth = api.post('{0}/login'.format(api_uri),
                headers = {'Accept': 'application/x-yaml'},
                data = {'username': user,
                        'password': password,
                        'eauth': 'ldap'})
# ...
keypair = api.post('{0}/keys'.format(api_uri),
                   data = {'mid': 'testminion',
                           'username': user,
                           'password': password,
                           # 'keysize': 4096,  # "A"
                           'keysize': '4096',  # "B"
                           'eauth': 'ldap'})
with open('/tmp/keys.tar', 'wb') as fh:
    fh.write(keypair.content)
# ...
```

Now, /tmp/keys.tar is indeed a tarball that contains a private and public key with method "B" - but it is always only 2048 bits (checked with `openssl rsa -text -noout -in minion.pem`).

With method "A", though, not only is a keypair not created, but the API throws the following exception:

```python
[22/Mar/2020:02:21:01] HTTP 
Traceback (most recent call last):
  File "/usr/lib/python3.6/site-packages/cherrypy/_cprequest.py", line 670, in respond
    response.body = self.handler()
  File "/usr/lib/python3.6/site-packages/cherrypy/lib/encoding.py", line 217, in __call__
    self.body = self.oldhandler(*args, **kwargs)
  File "/usr/lib/python3.6/site-packages/cherrypy/_cpdispatch.py", line 60, in __call__
    return self.callable(*self.args, **self.kwargs)
  File "/usr/lib/python3.6/site-packages/salt/netapi/rest_cherrypy/app.py", line 1717, in POST
    pub_key = ret.get('pub', '')
AttributeError: 'str' object has no attribute 'get'
```

returning the following with a `500`:

```html
[...]

PR #56426, opened 2020-03-22
Improve salt rest_cherrypy gen_keys tarball

johnnybubonic:
### What does this PR do?
This commit contains two *very* small improvements to the tarball generated by POST /keys, but they were bugging me:

* Add a time (mtime) to the minion.pem and minion.pub files.
* Set mode to 0600 for the minion.pem, due to it being a private key.

Shameless low-hanging-fruit for https://github.com/saltstack/salt/issues/56425#issuecomment-602138263

### Previous Behavior
* `POST /keys` would return (if successful) a tarball containing files with no timestamps.
* `POST /keys` would return (if successful) a tarball containing a private key with no defined octal mode.

### New Behavior
* Files in tarball now contain timestamps (should be *very* close to time of generation; more closely matches when they're added to the TarFile object)
* File `minion.pem` now has mode `0600` inside the TarFile.

### Tests written?

No, but if you find a bug in the three lines I added I'd be very impressed.

### Commits signed with GPG?

Yes


DmitryKuzmenko:
@johnnybubonic 
> Tests written?
> No, but if you find a bug in the three lines I added I'd be very impressed.

Missing `import time`.
I won! Write a test as you promised ;-)

dwoz:

Issue #52140, opened 2019-03-11
type hints not supported


kiemlicz:
### Description of Issue/Question
It is impossible to run Execution Module which contains type hints from CLI  
Is there any workaround available now?

### Setup
Create any custom Execution Module with type hints, e.g. `_modules/my.py`:  
```
def info(ip: str, port: int):
    return {}
```

### Steps to Reproduce Issue
Run from CLI (sync_all first):
`salt '*' my.info 127.0.0.1 11111`
It will fail with:
```
salt-minion-mnwlq:
    The minion function caused an exception: Traceback (most recent call last):
      File "/usr/lib/python3/dist-packages/salt/minion.py", line 1641, in _thread_return
        data)
      File "/usr/lib/python3/dist-packages/salt/minion.py", line 321, in load_args_and_kwargs
        argspec = salt.utils.args.get_function_argspec(func)
      File "/usr/lib/python3/dist-packages/salt/utils/args.py", line 271, in get_function_argspec
        aspec = _getargspec(func)  # pylint: disable=redefined-variable-type
      File "/usr/lib/python3/dist-packages/salt/utils/args.py", line 231, in _getargspec
        raise ValueError('Function has keyword-only arguments or annotations'
    ValueError: Function has keyword-only arguments or annotations, use getfullargspec() API which can support them
ERROR: Minions returned with non-zero exit code
```

### Versions Report
```
Salt Version:
           Salt: 2019.2.0

Dependency Versions:
           cffi: 1.12.2
       cherrypy: unknown
       dateutil: 2.8.0
      docker-py: 1.10.6
          gitdb: 2.0.0
      gitpython: 2.1.1
          ioflo: Not Installed
         Jinja2: 2.10
        libgit2: 0.27.4
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.4.8
   mysql-python: Not Installed
      pycparser: 2.19
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: 0.27.4
         Python: 3.5.3 (default, Sep 27 2018, 17:25:39)
   python-gnupg: Not Installed
         PyYAML: 3.13
          PyZMQ: 16.0.2
           RAET: Not Installed
          smmap: 2.0.1
        timelib: Not Installed
        Tornado: 4.4.3
            ZMQ: 4.2.1

System Versions:
           dist: debian 9.8
[...]

Issue #50679, opened 2018-11-29
salt doesn't support LXC 3.0+


mced:
### Description of Issue/Question
Creation of LXC with network_profile with LXC 3.0+ results to unusable containers.

_The problem is that salt doesn't support the new config keys supporting nic settings_: 
```
lxc.net.[i].type
lxc.net.[i].flags 
lxc.net.[i].link 
lxc.net.[i].mtu 
lxc.net.[i].name 
lxc.net.[i].hwaddr 
lxc.net.[i].ipv4.address 
lxc.net.[i].ipv4.gateway 
lxc.net.[i].ipv6.address 
lxc.net.[i].ipv6.gateway 
lxc.net.[i].script.up 
lxc.net.[i].script.down 
```

### Setup
salt/lxc.sls
```yaml
create-lxc-foo:
  lxc.present:
    - name:  foo
    - profile: ubuntu1804
    - network_profile: foo_network
```
pillar/profile.sls
```yaml
lxc.container_profile:
  ubuntu1804:
    template: download
    options:
      dist: ubuntu
      release: bionic
      arch: amd64
```
pillar/network.sls
```yaml
lxc.network_profile:
  foo_network:
    eth1:
      link: lxcbr0
      type: veth
      flags: up
```

### Versions Report
```
Salt Version:
           Salt: 2017.7.8
 
Dependency Versions:
           cffi: Not Installed
       cherrypy: Not Installed
       dateutil: 2.4.2
      docker-py: Not Installed
          gitdb: 0.6.4
      gitpython: 1.0.1
          ioflo: Not Installed
         Jinja2: 2.8
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: 1.0.3
   msgpack-pure: Not Installed
 msgpack-python: 0.4.6
   mysql-python: Not Installed
      pycparser: Not Installed
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: Not Installed
         Python: 2.7.12 (default, Nov 12 2018, 14:36:49)
   python-gnupg: Not Installed
         PyYAML: 3.11
          PyZMQ: 15.2.0
           RAET: Not Installed
          smmap: 0.9.0
        timelib: Not Installed
        Tornado: 4.2.1
            ZMQ: 4.1.4
 
System Versions:
           dist: Ubuntu 16.04 xenial
         locale: UTF-8
        machine: x86_64
        release: 4.4.0-139-generic
         system: Linux
[...]

PR #56427, opened 2020-03-22
Python 3 LXC config dump fix

kiemlicz:
Fixes Python3 compatibility

### What does this PR do?
Fix Python3 issue with:
```
    def tempfile(self):
        # this might look like the function name is shadowing the
        # module, but it's not since the method belongs to the class
        ntf = tempfile.NamedTemporaryFile()
        ntf.write(self.as_string())  # <----
        ntf.flush()
        return ntf
```

https://github.com/saltstack/salt/issues/52219

### Previous Behavior
`Passed invalid arguments to lxc.init: a bytes-like object is required, not 'str'`

### New Behavior
Config dumps properly  
However LXC module still won't work due to: https://github.com/saltstack/salt/issues/50679

### Tests written?
No and please **don't** make me write tests for this since there should be at least some "type checker" to catch errors of this kind
I'm still amazed why https://github.com/saltstack/salt/issues/52140 was marked as 'feature' and not highest priority bug.   
Moreover due to LXC 3 (default version even on Debian) this module still doesn't work

### Commits signed with GPG?
No


twangboy (changes requested):
This needs a testcase and a changelog

dwoz:

PR #56430, opened 2020-03-22
Module and state virt change

antropova-m:
### What does this PR do?
Add console and serial parameters to module virt.update and virt.running state.

### Tests written?
Yes

### Commits signed with GPG?
No

### DIFF
--- a/salt/modules/virt.py
+++ b/salt/modules/virt.py
@@ -143,6 +143,10 @@ VIRT_STATE_NAME_MAP = {
     6: "crashed",
 }
 
+default_port = 23023
+default_serial_type = "tcp"
+default_console_type = "tcp"
+
 
 def __virtual__():
     if not HAS_LIBVIRT:
@@ -780,6 +784,8 @@ def _gen_xml(
     graphics=None,
     boot=None,
     boot_dev=None,
+    serials=None,
+    consoles=None,
     **kwargs
 ):
     """
@@ -831,18 +837,76 @@ def _gen_xml(
                 context["boot"]["kernel"] = "/usr/lib/grub2/x86_64-xen/grub.xen"
                 context["boot_dev"] = []
 
-    if "serial_type" in kwargs:
-        context["serial_type"] = kwargs["serial_type"]
-    if "serial_type" in context and context["serial_type"] == "tcp":
-        if "telnet_port" in kwargs:
-            context["telnet_port"] = kwargs["telnet_port"]
-        else:
-            context["telnet_port"] = 23023  # FIXME: use random unused port
-    if "serial_type" in context:
-        if "console" in kwargs:
-            context["console"] = kwargs["console"]
-        else:
-            context["console"] = True
+    context["serials"] = []
+    if serials is not None:
+        for serial in serials:
+            serial_type = serial.get("type") or default_serial_type
+            if not serial_type:
+                raise SaltInvocationError("Missing type in serial")
+
+            serial_context = {
+                "type": serial_type
+            }
+            if serial_context["type"] == "tcp":
+                serial_context["port"] = serial.get("port", default_port)
+                serial_context["protocol"] = serial.get("protocol", "telnet")
+            context["serials"].append(serial_context)
+
+    context["consoles"] = []
+    if consoles is not None:
+        for console in consoles:
+            console_type = console.get("type") or default_console_type
+            if not console_type:
+                raise SaltInvocationError("Missing type in console")
+
+            console_context = {
+                "type": console_type
+            }
+            if console_context["type"] == "tcp":
+                console_context["port"] = console.get("port", default_port)
+                console_context["protocol"] = console.get("protocol", "telnet")
+            context["consoles"].append(console_context)
+
+    # processing of deprecated parameters
+    old_port = kwargs.get("telnet_port")
+    if old_port:
+        salt.utils.versions.warn_until(
+            "Aluminium",
+            "'telnet_port' parameter has been deprecated, use the 'serials' and 'consoles' parameters instead. "
+            "It will be removed in {version}.",
+        )
+
+    old_serial_type = kwargs.get("serial_type")
+    if old_serial_type:
+        salt.utils.versions.warn_until(
+            "Aluminium",
+            "'serial_type' parameter has been deprecated, use the 'serials' and 'consoles' parameters instead. "
+            "It will be removed in {version}.",
+        )
+        serial_context = {
+            "type": serial_type
+        }
+        if serial_context["type"] == "tcp":
+            serial_context["port"] = old_port if old_port is not None else default_port
+            serial_context["protocol"] = serial.get("protocol", "telnet")
+        context["serials"].append(serial_context)
+
+    old_console = kwargs.get("console")
+    if old_console:
+        salt.utils.versions.warn_until(
+            "Aluminium",
+            "'console' parameter has been deprecated, use the 'serials' and 'consoles' parameters instead. "
+            "It will be removed in {version}.",
+        )
+        if old_console is True:
+            console_context = {
+                "type": old_serial_type if old_serial_type else default_console_type
+            }
+            if console_context["type"] == "tcp":
+                console_context["port"] = old_port if old_port else default_port
+                console_context["protocol"] = console.get("protocol", "telnet")
+            context["consoles"].append(console_context)
+    # end processing of deprecated parameters
 
     context["disks"] = []
     disk_bus_map = {"virtio": "vd", "xen": "xvd", "fdc": "fd", "ide": "hd"}
@@ -1737,6 +1801,8 @@ def init(
     arch=None,
     boot=None,
     boot_dev=None,
+    serials=None,
+    consoles=None,
     **kwargs
 ):
     """
@@ -1856,6 +1922,17 @@ def init(
        A boolean value.
 
        .. versionadded:: sodium
+
+    :param serials:
+        Dictionary providing details on the serials connection to create. (Default: ``None``)
+        See :ref:`init-serials-def` for more details on the possible values.
+
+        .. versionadded:: Sodium
+    :param consoles:
+        Dictionary providing details on the consoles device to create. (Default: ``None``)
+        See :ref:`init-consoles-def` for more details on the possible values.
+
+        .. versionadded:: Sodium
 
     .. _init-nic-def:
 
@@ -2001,6 +2078,40 @@ def init(
 
         By default, not setting the ``listen`` part of the dictionary will default to
         listen on all addresses.
+
+    .. _init-serials-def:
+
+    .. rubric:: Serials Definitions
+
+    Serial dictionaries can contain the following properties:
+
+    type
+        Type of the serial connection, like ``'tcp'``, ``'pty'``.
+
+    port
+        The serial port number.
+        (Default: 23023)
+
+    Protocol
+        Name of connection protocol.
+        (Default: telnet)
+
+    .. _init-consoles-def:
+
+    .. rubric:: Consoles Definitions
+
+    Consol dictionaries can contain the following properties:
+
+    type
+        Type of the serial connection, like ``'tcp'``, ``'pty'``.
+
+    port
+        The serial port number.
+        (Default: 23023)
+
+    Protocol
+        Name of connection protocol.
+        (Default: telnet)
 
     .. rubric:: CLI Example
 
@@ -2151,6 +2262,8 @@ def init(
             graphics,
             boot,
             boot_dev,
+            serials,
+            consoles,
             **kwargs
         )
         log.debug("New virtual machine definition: %s", vm_xml)
@@ -2367,6 +2480,45 @@ def _diff_graphics_lists(old, new):
     return _diff_lists(old, new, _graphics_equal)
 
 
+def _serial_or_concole_equal(old, new):
+
+    def _filter_serial_or_concole(item):
+        """
+        Filter out elements to ignore when comparing items
+        """
+        return {
+            "type": item.attrib["type"],
+            "port": item.find("source").attrib["service"]
+            if item.find("source") is not None
+            else None,
+            "protocol": item.find("protocol").attrib["type"]
+            if item.find("protocol") is not None
+            else None,
+        }
+
+    return _filter_serial_or_concole(old) == _filter_serial_or_concole(new)
+
+
+def _diff_serial_list(old, new):
+    """
+        Compare serial definitions to extract the changes
+
+        :param old: list of ElementTree nodes representing the old serials
+        :param new: list of ElementTree nodes representing the new serials
+        """
+    return _diff_lists(old, new, _serial_or_concole_equal)
+
+
+def _diff_console_list(old, new):
+    """
+        Compare console definitions to extract the changes
+
+        :param old: list of ElementTree nodes representing the old consoles
+        :param new: list of ElementTree nodes representing the new consoles
+        """
+    return _diff_lists(old, new, _serial_or_concole_equal)
+
+
 def update(
     name,
     cpu=0,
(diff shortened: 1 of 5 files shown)


salt/modules/virt.py
```diff
@@ -1967,6 +1967,42 @@ def update(name,
                 devices_node.extend(changes[dev_type]['sorted'])
                 need_update = True
 
+    # Update the serial and the console definitions
+    if 'serial_type' in kwargs:
```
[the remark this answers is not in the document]
cbosdo: Also add the parameters documentation to the `update` function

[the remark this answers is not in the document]
cbosdo: According to [the libvirt domain documentation](https://libvirt.org/formatdomain.html#elementsConsole) the serial elements could be of various types and can have a source path, sometimes a mode and various other parameters.
There can also be multiple serial devices.

Along the lines of the NICs and disks, I would see a `serial` parameter as a list of dictionaries. An empty list means no serial device and the dictionary could hold properties like `type`. This way even if you don't implement all the properties of the serial devices this will be easy to extend it later on without breaking the existing behavior.

Of course the same applies for consoles.

I know these parameters are already in `virt.init()` but I think those should be deprecated in favor of the idea I described above. And thus only the news parameters would be propagated to `virt.defined()`, `virt.running()` and `virt.update()`

salt/modules/virt.py
```diff
@@ -1967,6 +1967,42 @@ def update(name,
...
+        serial_node = devices_node.find('serial')
+        if serial_node and serial_node.attrib['type'] != kwargs['serial_type']:
+            serial_node.set('type', kwargs['serial_type'])
+            need_update = True
+        if serial_node is None:
+            devices_node.append('serial').set('type', kwargs['serial_type'])
+            need_update = True
```
cbosdo: You got the idea, but now let's handle the possible multiple serial devices. I think it would be better to mimic the disk or interfaces code to do this since it handles lists and strives to normalize the XML elements to compare them. See the first commit of PR #56414 for a fix in that area.

salt/modules/virt.py
```diff
@@ -1967,6 +1967,42 @@ def update(name,
...
+            if console_node == None:
+                console_node = devices_node.append('console')
+
+            if console_node.attrib('type') != kwargs['serial_type']:
+                console_node.set('type', kwargs['serial_type'])
+                need_update = True
+
+            source_node = console_node.find('source')
+            if kwargs['serial_type'] == 'tcp':
+                telnet = kwargs['telnet_port'] if 'telnet_port' in kwargs else 23023 # FIXME: use random unused port
```
cbosdo: I know that code comes from `virt.init()` but I'm honestly not at ease with a hard-coded port number. Even picking a random unused port seems rather fragile. What about simply ignoring devices where that value is missing and logging an error?

salt/states/virt.py
```diff
@@ -265,7 +265,10 @@ def running(name,
             password=None,
             os_type=None,
             arch=None,
-            boot=None):
+            boot=None,
+            serial_type=None,
+            console=None,
+            telnet_port=None):
```
cbosdo: As mentioned in the `virt.update` change, those parameters should be deprecated in favor of `serials` and `consoles`

salt/states/virt.py
```diff
@@ -367,6 +370,18 @@ def running(name,
 
         .. versionadded:: 3000
 
+    :param serial_type: type of serial connection
+
+        .. versionadded:: 3000
```
cbosdo: When adding new parameters now, use `.. versionadded:: Sodium`.
The `Sodium` will be replaced by the proper version number at release time

tests/unit/modules/test_virt.py
```diff
@@ -1392,6 +1394,18 @@ def test_update(self):
         setxml = ET.fromstring(define_mock.call_args[0][0])
         self.assertEqual('vnc', setxml.find('devices/graphics').get('type'))
 
+        # Serial and console test case
+        self.assertEqual({
+                'definition': True,
+                'mem': True,
+                'disk': {'attached': [], 'detached': []},
+                'interface': {'attached': [], 'detached': []}
+            }, virt.update('my vm', serial_type='tcp', console=True, telnet_port=22223))
```
cbosdo: with the new parameters concept that would typically be:
`}, virt.update('my vm', serials=[{'type': 'tcp'}], consoles=[{'type': 'tcp', 'port': 22223, 'protocol': 'telnet'}]))`

cbosdo (changes requested):
Thanks Marina for the pull request. This is a really good start, even though I think we could start revamping the serial and console related parameters to be more future-proof, but that you couldn't know in advance ;)

The first commits for the concurrent change looks fishy to me and I suspect this is the think that causes a lot of tests to break. I would rather remove them. Why did you want to change that in the first place? May be we could solve the problem differently

salt/modules/virt.py
```diff
@@ -599,18 +601,29 @@ def _gen_xml(
...
-            context["telnet_port"] = 23023  # FIXME: use random unused port
-    if "serial_type" in context:
-        if "console" in kwargs:
-            context["console"] = kwargs["console"]
-        else:
-            context["console"] = True
+    context["serials"] = []
+    for serial in enumerate(serials):
+        serial_context = {
+            "type": serial.type
```
cbosdo, line 607: Don't assume the `type` property will always be in the `serial` dict... users may forget it and it would explode here. You should guard the whole loop with something like

```python
if not serial.get("type"):
    throw SaltInvocationError("Missing type in serial")
```

salt/modules/virt.py
```diff
@@ -599,18 +601,29 @@ def _gen_xml(
...
+        }
+        if serial_context["type"] == "tcp":
+            if (serial["port"] is None) or (serial["protocol"] is None):
+                raise CommandExecutionError("Absent serial parameter for TCP connection")
+            serial_context["port"] = serial["port"]
+            serial_context["protocol"] = serial["protocol"]
```
cbosdo, line 613: Use `get` since it provides a default value even if the property you are looking for is not set.

And I don't agree with this check. The old code was using a default port `23023` if not set and forced the protocol to `telnet` in the Jinja file. Even if that is not perfect you should provide the same feature.

``` suggestion
            serial_context["port"] = serial.get("port", default_port)
            serial_context["protocol"] = serial.get("protocol", "telnet")
```

salt/modules/virt.py
```diff
@@ -599,18 +601,29 @@ def _gen_xml(
...
-            context["telnet_port"] = kwargs["telnet_port"]
-        else:
-            context["telnet_port"] = 23023  # FIXME: use random unused port
-    if "serial_type" in context:
-        if "console" in kwargs:
-            context["console"] = kwargs["console"]
-        else:
-            context["console"] = True
+    context["serials"] = []
+    for serial in enumerate(serials):
```
cbosdo, line 605: not sure you need `enumerate()` here

salt/modules/virt.py
```diff
@@ -599,18 +601,29 @@ def _gen_xml(
...
+        }
+        if serial_context["type"] == "tcp":
+            if (serial["port"] is None) or (serial["protocol"] is None):
+                raise CommandExecutionError("Absent serial parameter for TCP connection")
+            serial_context["port"] = serial["port"]
+            serial_context["protocol"] = serial["protocol"]
+        context["serials"].append(serial_context)
+
+    context["consoles"] = []
+    for console in enumerate(consoles):
```
cbosdo, line 617: not sure about `enumerate()` here too

salt/modules/virt.py
```diff
@@ -1278,6 +1293,11 @@ def init(
 
         .. versionadded:: 3000
 
+    :param serials: list of serial connections
+        .. versionadded:: Sodium
+    :param consoles: list of console parameters
+        .. versionadded:: Sodium
```
cbosdo, line 1299: You need to document what properties are allowed in the items of the lists

salt/modules/virt.py
```diff
@@ -1759,6 +1783,11 @@ def update(
...
```
cbosdo, line 1789: Refer to the documentation of the properties of items in the list contained in `init()`. See how this is done for disks, network interfaces and graphics already.

salt/modules/virt.py
```diff
@@ -1903,6 +1934,8 @@ def update(
         "disk": ["disks", "disk_profile"],
         "interface": ["interfaces", "nic_profile"],
         "graphics": ["graphics"],
+        "serial": ["serial"],
+        "console": ["console"]
```
cbosdo, line 1938: You may need to enhance the diffing function as done in `_diff_graphics_list()`. Be aware that libvirt will slightly change the XML when running the VM. I expect something like this to be needed:

```python
def _serial_equal(serial1, serial2):
    # TODO diff the XML fragments based only on the meaningful elements and properties

def _diff_serial_list(old, new):
    return _diff_lists(old, new, _serial_equal)
```

And surely the same for consoles... unless you can share some of that code for both since the data are fairly similar

salt/states/virt.py
```diff
@@ -355,6 +357,11 @@ def defined(
 
         .. deprecated:: sodium
 
+    :param serials: list of serial connections
+        .. versionadded:: Sodium
+    :param consoles: list of console parameters
+        .. versionadded:: Sodium
+
```
cbosdo, line 364: Like for the `update()` documentation, add a link to the list items definition that you will add to `init()`

salt/states/virt.py
```diff
@@ -582,6 +595,12 @@ def running(
             }
 
         .. versionadded:: 3000
+    :param serials: list of serial connections
+
+        .. versionadded:: Sodium
+    :param consoles: list of console parameters
+
+        .. versionadded:: Sodium
```
cbosdo, line 603: Doc link here too

salt/templates/virt/libvirt_domain.jinja
```diff
@@ -71,31 +71,37 @@
...
-                {% if serial_type == 'pty' %}
+
+                {% for serial in serials %}
+                {% if serial.type == 'pty' %}
                 <serial type='pty'>
                         <target port='0'/>
                 </serial>
-                {% if console %}
+                {% endif %}
+                {% if serial.type == 'tcp' %}
```
cbosdo, line 81: [Jinja](https://jinja.palletsprojects.com/en/2.10.x/templates/#if) has an `{% elif %}` block, please use it here.
```suggestion
                {% elif serial.type == 'tcp' %}
```

salt/templates/virt/libvirt_domain.jinja
```diff
@@ -71,31 +71,37 @@
...
-                        <protocol type='telnet'/>
-                        <target port='0'/>
-                </serial>
-                {% if console %}
+                {% if console.type == 'tcp' %}
                 <console type='tcp'>
-                        <source mode='bind' host='' service='{{ telnet_port }}'/>
-                        <protocol type='telnet'/>
+                        <source mode='bind' host='' service='{{ console.port }}'/>
+                        <protocol type='{{ console.protocol }}'/>
```
cbosdo, line 100: Either you use a default value for protocol here or you define it in the `_gen_xml()`. I would prefer the latter one.

salt/modules/virt.py
```diff
@@ -599,18 +601,29 @@ def _gen_xml(
...
+    context["consoles"] = []
+    for console in enumerate(consoles):
+        console_context = {
+            "type": console.type
+        }
+        if console_context["type"] == "tcp":
+            if (console["port"] is None) or (console["protocol"] is None):
+                raise CommandExecutionError("Absent console parameters for TCP connection")
+            console_context["port"] = console["port"]
+            console_context["protocol"] = console["protocol"]
```
cbosdo, line 625: In the old code the protocol is forced to `telnet`. The port was also set to the same default that the serial port (`23023`). You need to use it as a default values.
```suggestion
            console_context["port"] = console.get("port", default_port)
            console_context["protocol"] = console.get("protocol", "telnet")
```

salt/modules/virt.py
```diff
@@ -599,18 +601,29 @@ def _gen_xml(
...
-    if "serial_type" in context and context["serial_type"] == "tcp":
-        if "telnet_port" in kwargs:
-            context["telnet_port"] = kwargs["telnet_port"]
-        else:
-            context["telnet_port"] = 23023  # FIXME: use random unused port
-    if "serial_type" in context:
-        if "console" in kwargs:
-            context["console"] = kwargs["console"]
-        else:
-            context["console"] = True
```
cbosdo, line 613: Here you cannot get rid of the old parameters so easily since we need to flag them as deprecated before removing them to avoid being to harsh to people using them. You need to do something like this:

```python
default_serial_type = kwargs.get("serial_type")
if default_serial_type:
    salt.utils.versions.warn_until(
        "Aluminium",
        "'serial_type' parameter has been deprecated, use the 'serials' and 'consoles' parameters instead. "
        "It will be removed in {version}.",
    )

default_port = kwargs.get("telnet_port")
if telnet_port:
    salt.utils.versions.warn_until(
        "Aluminium",
        "'telnet_port' parameter has been deprecated, use the 'serials' and 'consoles' parameters instead. "
        "It will be removed in {version}.",
    )
else:
    default_port = 23023

# TODO do something similar for 'console' parameter
```

salt/modules/virt.py
```diff
@@ -599,18 +601,29 @@ def _gen_xml(
...
+    context["serials"] = []
+    for serial in enumerate(serials):
+        serial_context = {
+            "type": serial.type
+        }
```
cbosdo, line 608: You need to take into account the old way to give the type, use the above mentioned `default_serial_type` too

Also don't assume the `type` property will always be in the `serial` dict... users may forget it and it would explode here. You should guard the whole loop with something like:

```suggestion
        serial_type = serial.get("type") or default_serial_type
        if not serial_type:
            throw SaltInvocationError("Missing type in serial")
            
        serial_context = {
            "type": serial_type
        }
```

cbosdo (changes requested):
See my comments inline. I also saw a lot of unit test failures that you need to fix for me to review them.
Ensure that running these in virtualenv runs nicely:

```sh
./runtests.py -n unit.states.test_virt
./runtests.py -n unit.modules.test_virt
```

You should also install the pre-commit hook on your machine since this would run the linter, isort and black checks as you commit:
```
pre-commit install
```

cbosdo:
Looks better, but could you rebase on latest master and fix all the red tests?

cbosdo:
@marina2209 maybe you should rebase your PR on this pull request of mine: #58332 it makes handling new elements in `virt.update()` way easier. Don't hesitate to reach out to me on Slack if you need help on it. See the newly introduced `params_mapping` variable in the `virt.update` function for examples.

marina2209:
Ok. I am going to make a rebase.

пн, 31 авг. 2020 г. в 19:00, Cedric Bosdonnat <notifications@github.com>:

cbosdo:
I finally found time to go through your code in more details, rebased it, cleaned the code a little bit and added support for more serial/console types. You can check it out here:

https://github.com/cbosdo/salt/tree/virt-console

Do you wish me to submit this branch as a pull request or you prefer continue working on it?

marina2209:
Thanks for your help. It would be great if you submit this branch as a pull
request.

пт, 30 окт. 2020 г. в 14:15, Cedric Bosdonnat <notifications@github.com>:

cbosdo:
> Thanks for your help. It would be great if you submit this branch as a pull request

Ok, I'll handle it then.

cbosdo:
@marina2209 you can now close this PR: I have submitted another one with your rebased commit

marina2209:
Ok.

пн, 2 нояб. 2020 г. в 16:16, Cedric Bosdonnat <notifications@github.com>:

marina2209:
All changes moved to another PR.

PR #56493, opened 2020-04-01
[docs] Fix incorrect smartos.vm_present example

sjorge:
### What does this PR do?
When looking up my own docs because I forgot how it worked, I noticed and example was broken.

I also added an extra warning as I wasted half an hour on silly mistake.

N/a

### Previous Behavior
N/a

### New Behavior
N/a

### Tests written?
No, as this is a docstring only change

### Commits signed with GPG?
No


DmitryKuzmenko (dismissed):

DmitryKuzmenko:
Codecov link isn't opening that's a known issue. So this is merge ready.

Issue #56258, opened 2020-02-26
rabbitmq_user.present Errors if user exists.


rtaft:
### Description of Issue
When using rabbitmq_user.present, the user is created the first time around.  After that, the salt function errors with 
```
Error: RabbitMQ command failed: Error:
              {:undef, [{:rabbit_auth_backend_internal, :check_user_login, ["USERNAME", [password: "PASSWORD"]], []}, {:erl_eval, :do_apply, 6, [file: 'erl_eval.erl', line: 680]}, {:rpc, :"-handle_call_call/6-fun-0-", 5, [file: 'rpc.erl', line: 197]}]}
```
### Setup
```
rabbitmq_add_myuser_user:
  rabbitmq_user.present:
    - name: USERNAME
    - password: PASSWORD
    - tags:
      - administrator
    - runas: rabbitmq
    - perms:
      - 'vhost1':
        - '.*'
        - '.*'
        - '.*'
      - 'vhost2':
        - '.*'
        - '.*'
        - '.*'
```

### Steps to Reproduce Issue
```
2020-02-26 15:52:57,531 [salt.template    :30  ][PROFILE ][4855] Time (in seconds) to render '/var/cache/salt/minion/files/base/rabbitmq/state.sls' using 'yaml' renderer: 0.00201606750488
2020-02-26 15:52:57,532 [salt.utils.lazy  :104 ][DEBUG   ][4855] LazyLoaded config.get
2020-02-26 15:52:57,532 [salt.loaded.int.module.config:483 ][DEBUG   ][4855] key: gpg_keydir, ret: _|-
2020-02-26 15:52:57,532 [salt.loaded.int.render.gpg:355 ][DEBUG   ][4855] Reading GPG keys from: /etc/salt/gpgkeys
2020-02-26 15:52:57,533 [salt.template    :30  ][PROFILE ][4855] Time (in seconds) to render '/var/cache/salt/minion/files/base/rabbitmq/state.sls' using 'gpg' renderer: 0.00186109542847
2020-02-26 15:52:57,535 [salt.utils.lazy  :104 ][DEBUG   ][4855] LazyLoaded rabbitmq_user.present
2020-02-26 15:52:57,535 [salt.state       :1867][INFO    ][4855] Running state [USERNAME] at time 15:52:57.535303
2020-02-26 15:52:57,535 [salt.state       :1900][INFO    ][4855] Executing state rabbitmq_user.present for [USERNAME]
[...]

PR #56497, opened 2020-04-02
Support newer versions of RabbitMQ in check_password

ScoreUnder:
### What does this PR do?

Adds support for newer versions of RabbitMQ. Not entirely sure as to the exact range of versions it adds support for.

#56258
Duplicates #56193 (This was originally written as a quick fix for our network without checking existing PRs, so it is somewhat unsurprising that someone else has the same PR ready)

### Previous Behavior
1. Support versions of rabbitmq which provide an ETS-form output from `rabbitmqctl status`
2. Default to trying old approaches when version is unclear

### New Behavior
1. Also support newer versions of rabbitmq with human-readable `rabbitmq status` output
2. Default to trying the newer approach when version is unclear

### Tests written?
Yes

### Commits signed with GPG?
Yes


Akm0d:
The failing tests are directly related to this PR
```
  File "/tmp/kitchen/testing/tests/unit/modules/test_rabbitmq.py", line 1065, in test__get_server_version_old
    self.assertListEqual(rabbitmq._get_server_version(), [3, 5, 7])
AssertionError: First sequence is not a list: None
```

ScoreUnder:
Tests were broken from the merge (`_get_server_version` got mangled by unrelated changes) but now that #56193 has merged I think this PR is no longer needed so I will close it

PR #56547, opened 2020-04-03
testing only running few tests if this is a doc pr

bryceml:
### What does this PR do?
Makes it so the pr-tests only test docs, lint, and ubuntu (since there are doc tests as part of a normal run)

### DIFF
--- a/.ci/docs
+++ b/.ci/docs
@@ -1,6 +1,7 @@
-@Library('salt@master-1.6') _
+@Library('salt@bryce-test') _
 
 runDocs(
-    env: env)
+    env: env,
+    run_on_docs_pr: true)
 
 // vim: ft=groovy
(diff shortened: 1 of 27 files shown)


s0undt3ch:
I would suggest running on either Debian 9 or 10 since those are the fastest builds?

bryceml:
Sure, I just wasn't sure all the doc tests run on anything but ubuntu, they probably do.

s0undt3ch:
We would also have to disable the required Py3 statuses, unless the builds just maked as passed? Ie, all other builds not building docs.
An alternative would be to run all builds, but only doc related tests?

bryceml:
that would be nice, how do we do that?

s0undt3ch:
Once pytest lands we ciuld create a doc tests nox session and use that instead of the regular test session if it's only a doc related change.

bryceml:
yeah, lets do that, should I leave this here until we get that work done?

s0undt3ch:
Either that or an issue.

bryceml:
We would need a way to know that the tests were run with doc only change easily, so if someone removes the label after the test runs, we still know.

I'll make an issue referencing this and then close this.

Issue #56431, opened 2020-03-23
Module service.stop use StopService instead of StopServiceWithDeps


Denislb:
### Description of Issue

Module `service.stop` does not stop recursively services on windows.
We got an error when we try to stop  service with dependencies: `ERROR: Failed To Stop ServiceName: A stop control has been sent to a service that other running services are dependent on.`
The service and his dependencies are not stopped 

### Setup
- Windows 10
- Salt Minion 3000 py2

### Steps to Reproduce Issue
You can run:
`salt 'minion' service.stop "ServiceName"`  where ServiceName has dependencies

Ch3LL:
we could probably add an arg to that state to call `StopServiceWithDeps` if its set. will approve as a feature request thanks.

afischer-opentext-com:
On the [develop](https://github.com/saltstack/salt/blob/develop/salt/modules/win_service.py#L657 ) branch the modules support the requested functionality. I am not sure why this is not present in Salt 3000 (or master branch) as this is not totally new.

Ch3LL:
yeah we recently changed our release and branch strategy detailed here https://github.com/saltstack/salt-enhancement-proposals/pull/20

so yes porting those changes to master would be the right approach. thanks for doing that.

richarddbarnett:
Would this also affect `mod_watch` restarting a service?

I have a state
```
datadog-agent-service:
  service.running:
    - name: datadogagent
    - listen:
      - file: c:/ProgramData/DataDog/datadog.conf
      - file: c:/ProgramData/DataDog/conf.d/process.d/conf.yaml
```
and when one of the listened-to files changes Salt is unable to restart the service because there's a running `datadog-trace-service` that's dependent on `datadogagent`.

brandonRoom:
I was having the same issue with this piece of code
`datadogagent:
  service.running:
    - enable: True
    - watch:
      - file: datadog_config
      - file: datadog_design_online_config`

I searched high and low for an answer to this issue. I settled for looking through the code and found the following solution
`datadogagent:
  service.running:
    - enable: True
    - force: True
    - watch:
      - file: datadog_config
      - file: datadog_design_online_config`

The addition of `-force: True` seems so obvious now.

PR #56574, opened 2020-04-07
Port service dependencies feature from develop to master

ghost:
### What does this PR do?

This is a port of pull request #52286 which for some reason did not make it into master. This came up with issue #56431 again. 

See PR #52286

### Tests written?

Yes

### Commits signed with GPG?

Yes


salt/modules/win_service.py
```diff
@@ -484,6 +625,15 @@ def start(name, timeout=90):
 
             .. versionadded:: 2017.7.9,2018.3.4
 
+        with_deps (bool):
+            If enabled start the given service and the services the current
+            service depends on.
+
+        with_parents (bool):
```
twangboy, line 632: This name is confusing... sounds more like Children... but not quite. Maybe use `with_depends_on`, though I'm not sold on that either.

twangboy (changes requested):

salt/modules/win_service.py
afischer-opentext-com: Sure, I can do that. The fun fact here is that there was a discussion about that name already and now guess who suggested to use "parents" ;)

Issue #56554, opened 2020-04-06
state.orch highstate output only indents first line under Changes:Minion


leifliddy:
### Description of Issue

```
server.example.com_master:
----------
          ID: command_1
    Function: salt.function
        Name: cmd.run
      Result: True
     Comment: Function ran successfully. Function cmd.run ran on cloud.
     Started: 05:11:33.376113
    Duration: 441.463 ms
     Changes:   
              cloud:
                  file1
              file2
              file3
```
In this output you can the that ```file1``` was indented, but ```file2``` and ```file3``` were not.

### Steps to Reproduce Issue

run an orchestration state that uses ```cmd.run``` that results in multiple ```Changes``` lines for a minion in the output.

ie 
```
command_1:
  salt.function:
    - name: cmd.run
    - tgt: 'server'
    - arg:
      - ls /etc
```

```salt-run state.orch orch.test --out highstate```

### Versions Report

```
Salt Version:
           Salt: 3000
 
Dependency Versions:
           cffi: 1.12.3
       cherrypy: Not Installed
       dateutil: 2.8.0
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
         Jinja2: 2.10.1
        libgit2: Not Installed
       M2Crypto: 0.35.2
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.6.1
   mysql-python: Not Installed
      pycparser: 2.14
       pycrypto: Not Installed
   pycryptodome: 3.9.7
         pygit2: Not Installed
         Python: 3.7.6 (default, Jan 30 2020, 09:44:41)
   python-gnupg: Not Installed
         PyYAML: 5.3.1
          PyZMQ: 18.0.2
          smmap: Not Installed
        timelib: Not Installed
        Tornado: 4.5.3
            ZMQ: 4.3.2
 
System Versions:
           dist: fedora 31 Thirty One
         locale: UTF-8
        machine: x86_64
        release: 5.5.13-200.fc31.x86_64
         system: Linux
        version: Fedora 31 Thirty One
```

### Resolution:
This issue can be resolved with the following patch:
```
--- /usr/lib/python3.7/site-packages/salt/output/highstate.py.orig	2020-04-09 21:20:11.099958656 +0200
+++ /usr/lib/python3.7/site-packages/salt/output/highstate.py		2020-04-09 20:26:55.524264605 +0200
[...]

PR #56583, opened 2020-04-08
correct formatting issue with highstate output module

leifliddy:
### What does this PR do?
Corrects a formatting issue when ```cmd.run``` is run from an orchestration and returns multiple lines of output. This PR ensures every line of output under ```Changes``` is indented properly.

#56554

### Previous Behavior

given the orchestration state

```
command_1:
  salt.function:
    - name: cmd.run
    - tgt: 'server'
    - arg:
      - ls /tmp
```
the files listed under ```Changes``` would not be indented properly
```
[root@server ~]# salt-run state.orch orch.test --out highstate
server.example.com_master:
----------
          ID: command_1
    Function: salt.function
        Name: cmd.run
      Result: True
     Comment: Function ran successfully. Function cmd.run ran on server.
     Started: 19:02:44.195687
    Duration: 413.993 ms
     Changes:   
              server:
                  file1
              file2
              file3
```

### New Behavior
```
[root@server ~]# salt-run state.orch orch.test --out highstate
server.example.com_master:
----------
          ID: command_1
    Function: salt.function
        Name: cmd.run
      Result: True
     Comment: Function ran successfully. Function cmd.run ran on server.
     Started: 19:03:19.668603
    Duration: 418.08 ms
     Changes:   
              server:
                  file1
                  file2
                  file3
```

### DIFF
--- a/salt/output/highstate.py
+++ b/salt/output/highstate.py
@@ -186,7 +186,7 @@ def _format_host(host, data, indent_level=1):
     nchanges = 0
     strip_colors = __opts__.get("strip_colors", True)
 
-    if data is None or data is True or data is False:
+    if isinstance(data, int) or data is None:
         nchanges = 1
         hstrs.append(('{0}    {1}{2[ENDC]}'.format(hcolor, data, colors)))
         hcolor = colors["CYAN"]  # Print the minion name in cyan
--- a/salt/output/highstate.py
+++ b/salt/output/highstate.py
@@ -190,7 +190,7 @@ def _format_host(host, data, indent_level=1):
         nchanges = 1
         hstrs.append(('{0}    {1}{2[ENDC]}'.format(hcolor, data, colors)))
         hcolor = colors["CYAN"]  # Print the minion name in cyan
-    elif isinstance(data, int) or isinstance(data, six.string_types):
+    elif isinstance(data, six.string_types):
         # Data in this format is from saltmod.function,
         # so it is always a 'change'
         nchanges = 1
--- a/salt/output/highstate.py
+++ b/salt/output/highstate.py
@@ -188,15 +188,15 @@ def _format_host(host, data, indent_level=1):
 
     if isinstance(data, int):
         nchanges = 1
-        hcolor = colors["CYAN"]  # Print the minion name in cyan
         hstrs.append(('{0}    {1}{2[ENDC]}'.format(hcolor, data, colors)))
+        hcolor = colors["CYAN"]  # Print the minion name in cyan
     elif isinstance(data, six.string_types):
         # Data in this format is from saltmod.function,
         # so it is always a 'change'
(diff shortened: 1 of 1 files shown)


salt/output/highstate.py
```diff
@@ -186,13 +186,18 @@ def _format_host(host, data, indent_level=1):
...
+    if data is None or data is True or data is False:
+        nchanges = 1
+        hstrs.append(('{0}    {1}{2[ENDC]}'.format(hcolor, data, colors)))
+        hcolor = colors["CYAN"]  # Print the minion name in cyan
+    elif isinstance(data, int) or isinstance(data, six.string_types):
         # Data in this format is from saltmod.function,
         # so it is always a 'change'
         nchanges = 1
-        hstrs.append(("{0}    {1}{2[ENDC]}".format(hcolor, data, colors)))
+        for data in data.splitlines():
```
DmitryKuzmenko, line 197: The `int` type has no attribute `splitlines`.

DmitryKuzmenko: Possibly you've meant to handle `int` in the first block handling `None` and `bool`?

DmitryKuzmenko (changes requested):

DmitryKuzmenko:
@leifliddy could you please write a unit test for it?

salt/output/highstate.py
leifliddy: That was oversight on my part.
Odd, I never realized that True and False are numerically equivalent to  1 and 0 in python
So, the first check just needs to be
```if isinstance(data, int) or data is None:```

In the future, if someone wanted to apply different colors to different data types, then it might make sense to do a ```if data is True or data is False```  check. 
But, there's no need to perform that check in the current module (it's a bit wordy anyway) 

Ok, I've just pushed the new commit, this one should pass no problem. 

> could you please write a unit test for it?
I'm not sure how to do that, I'll research that a bit later.

leifliddy: I'm going to close this PR and submit a new one, the current PR is too messy given the relative simplicity of the change requested.

leifliddy:
closing PR, will open a new one with less commits.

salt/output/highstate.py
DmitryKuzmenko: It's okay. Thank you for your work!

DmitryKuzmenko:
@leifliddy thank you for your work. Let me know if you need any help on this.

salt/output/highstate.py
leifliddy: the new PR# is 56598
The ```highstate.py``` file changed since my last pull, which is the reason for the 2nd commit.

leifliddy:
the new PR# is 56598

DmitryKuzmenko:
Adding a link to associate them to each other #56598

Issue #53340, opened 2019-06-02
Windows pkg.list_pkgs and grain encoding issue


max-arnold:
### `®` of death

Somewhere deep in Salt an output of `pkg.list_pkgs` is not decoded properly and passed as a binary string (not unicode)

This binary string gets stored in grain

When Salt tries to read the grain, it expects it to be stored as unicode, and fails miserably while trying to decode it, because the characters are from CP437

### Setup

1. Windows 10
2. Install the Intel driver (msi) https://downloadmirror.intel.com/28688/a08/WiFi_21.0.0_PROSet64.zip (link to the page: https://downloadcenter.intel.com/download/28688/Intel-PROSet-Wireless-Software-and-Drivers-for-IT-Admins)  This also happens with MS Office localization package names that contain diacritics, but this Intel driver is the simplest way to reproduce the problem.

### Steps to Reproduce Issue

1. Apply the following state:

```
{% set pkgs = salt['pkg.list_pkgs']() %}

windows_packages:
   grains.present:
     - value : {{ pkgs }}
     - force: True
```

The output should look like this

```
local:
----------
          ID: windows_packages
    Function: grains.present
...
Intel\xc2\xae PROSet/Wireless WiFi Software:
                      21.0.0.2990

...
Succeeded: 1 (changed=1)
Failed:    0
------------
Total states run:     1
Total run time:   8.316 ms
```

2. Apply the same state second time. It should fail like this:

```
[WARNING ] Bad syntax in grains file! Skipping.
[ERROR   ] Unable to read existing grains file: unacceptable character #x00ae: invalid start byte
  in "c:\salt\conf\grains", position 548
local:
----------
          ID: windows_packages
    Function: grains.present
      Result: False
     Comment: Unable to read existing grains file: unacceptable character #x00ae: invalid start byte
                in "c:\salt\conf\grains", position 548
     Started: 00:39:02.515980
    Duration: 7.801 ms
     Changes:

Summary for local
------------
Succeeded: 0
Failed:    1
------------
Total states run:     1
Total run time:   7.801 ms
```

```
c:\salt>chcp
Active code page: 437
```

### Versions Report

<details><summary>c:\salt>salt-call --versions-report</summary>

```
Salt Version:
           Salt: 2018.3.4

Dependency Versions:
           cffi: 1.10.0
       cherrypy: 10.2.1
       dateutil: 2.6.1
      docker-py: Not Installed
          gitdb: 2.0.5
      gitpython: 2.1.3
[...]

PR #56077, opened 2020-02-06
Update to python 3.7

twangboy:
### What does this PR do?
Updates Windows installer scripts to use Python 3.7.4 as the version of python to be bundled for Windows. Removes references to Python 2. Updates a few dependencies required by Python 3.7.

Merge this one first: https://github.com/saltstack/salt/pull/56120

Fixes: https://github.com/saltstack/salt/issues/56873

### Previous Behavior
Python 2.7.15 and Python 3.5.4

### New Behavior
Python 3.7.4

### Tests written?
No

### Commits signed with GPG?
Yes

PR #56585, opened 2020-04-08
Handle unicode custom grains on Windows

twangboy:
### What does this PR do?
Handles the creation of custom grains with Unicode Characters properly in Windows. The grains config file needs to be encoded UTF-8 in order to write and read the grains properly. If you don't specify the encoding with fopen, then it will use the system encoding which is usually cp437. This issue may pop up elsewhere in salt as we move to Python 3.7+

This will need to go in after the following:
https://github.com/saltstack/salt/pull/56077

### Previous Behavior
Unable to decode a custom grain with unicode characters if the file encoding isn't UTF-8

### New Behavior
Custom grains are written using UTF-8 and is decoded properly as a result

### Tests written?
No

### Commits signed with GPG?
Yes


waynew:
Are you planning to write tests for this?

max-arnold:
Does that fix https://github.com/saltstack/salt/issues/53340 ?

twangboy:
@max-arnold The transition to Python37 should take care of https://github.com/saltstack/salt/issues/53340. This handles an issue when creating custom grains.

@waynew I do intend to write some tests for this... they're gonna be awesome.

max-arnold:
>  The transition to Python37 should take care of #53340

Transition from Python27 or from Python35? This issue happened on 3.5.3 and I don’t understand what is going to change on 3.7.

waynew (changes requested):
Needs tests :joy:

PR #56586, opened 2020-04-08
Pre-commit fixes

waynew:
### What does this PR do?

This is not, strictly speaking, necessary. Removing the language version
allows use of one's default supported python (3.6+).

Also consolidates isort config in pyproject.toml and fixes excludes, because the way pre-commit provides black/isort files  on `pre-commit -a` overrides their default exclude settings

### Previous Behavior

Contributors not using Python3.7 in their dev environment would have to add pyenv or similiar

### New Behavior

Now contributors with Python >=3.6 should be A-OK

### Tests written?

Not relevant

### Commits signed with GPG?

Yes

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -5,3 +5,12 @@ exclude= '''
   | tests/kitchen
   | templates
 )/
+
+
+[tool.isort]
+multi_line_output = 3
+include_trailing_comma = true
+force_grid_wrap = 0
+use_parentheses = true
+line_length = 88
+skip=salt/ext,tests/kitchen,templates
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -5,12 +5,3 @@ exclude= '''
   | tests/kitchen
   | templates
 )/
-
-
-[tool.isort]
-multi_line_output = 3
-include_trailing_comma = true
-force_grid_wrap = 0
-use_parentheses = true
-line_length = 88
-skip=salt/ext,tests/kitchen,templates
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,7 +1,18 @@
 [tool.black]
-exclude= '''
-\(
+exclude= """
+/(
     salt/ext
   | tests/kitchen
   | templates
 )/
+"""
+
+[tool.isort]
+multi_line_output = 3
+include_trailing_comma = true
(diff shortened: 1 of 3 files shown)


s0undt3ch (dismissed):
Should be `python3` at least?

DmitryKuzmenko (dismissed):
Let's try =)

waynew:
@s0undt3ch I think you're right - I should go ahead and just set the [`default_language_version`](https://pre-commit.com/#pre-commit-configyaml---top-level) to python3.

s0undt3ch (dismissed):

pyproject.toml
```diff
@@ -5,3 +5,12 @@ exclude= '''
   | tests/kitchen
   | templates
 )/
+
+
+[tool.isort]
```
s0undt3ch, line 10: This actually makes isort and black fight over imports.
Let's leave the old `.isort.cfg` as it was?

s0undt3ch (changes requested):

pyproject.toml
waynew: I should have added the `ensure_newline_before_comments` flag, but I couldn't get it to work correctly :suspect: 

I'm down with switching back, though when there's a version 5 release of isort that has mo' bettah support, I would prefer heading back to pyproject.toml

Also FWIW when toml hits the version 1 release, Python will be including a toml parser in the stdlib :tada:

s0undt3ch: Cool, we'll switch when it lands :)

s0undt3ch (dismissed):

Ch3LL (dismissed):

waynew:
Doh, don't merge yet!

pyproject.toml
waynew: Okay, turns out I figured out how to fix it. My bad :upside_down_face:

waynew:
Okay, *now* it should be good.

s0undt3ch:
Run it by tests/conftest.py and see if both agree

s0undt3ch:
Or at least the same file in https://github.com/saltstack/salt/pull/56460

s0undt3ch (approved):

dwoz (approved):

Issue #56322, opened 2020-03-07
Returner does not work in client batch mode


sundowner69:
### Description of Issue
While using the LocalClient with cmd_batch() I was wondering why I got no client returner output. It looks like there is a small issue in /usr/lib/python2.7/site-packages/salt/cli/batch.py in the class Batch, function run().
The function Batch.run() reads the returner option like this:
<code>
ret=self.opts.get('return', '')</code>  ( line 241 )
This works fine as long as you use the cli salt/salt-call command in batch mode but not if you use LocalClient.cmd_batch() since this is using 'ret' for an optional argument.
If I change the code in batch.py by changing <code>ret=self.opts.get('return', '')</code> to <code>ret=self.opts.get('ret', '')</code>, the LocalClient.cmd_batch() call works fine but the cli stops working by not using the returner. This behaviour is noticed in all versions of salt from 2018.3 or even earlier.
I am not sure if this is the way it is supposed to work, can you please advise?

### Setup
Take an arbitrary returner e.g. /srv/salt/_returners/test.py:
<pre><code>
def returner(ret) :
    try :
        with open('/tmp/test.log','w') as fh :
            fh.write('Returner started...\n{0}'.format(str(ret)))
    except :
        pass
</code></pre>

Take a small python script using the LocalClient() /root/bin/test_batch.py:
<pre><code>
import salt.client
local = salt.client.LocalClient()
local.cmd_batch('*','test.ping',ret='test',batch='10%')
</code></pre>

Compare the results by running CLI in batch mode and the small script. Change the 'return' into 'ret' in the code and run again.

### Versions Report
mars:/srv/salt/_returners # salt --versions-report
Salt Version:
           Salt: 2019.2.0
 
Dependency Versions:
           cffi: 1.5.2
       cherrypy: Not Installed
       dateutil: Not Installed
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: Not Installed
         Jinja2: 2.8
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: Not Installed
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.4.6
   mysql-python: Not Installed
      pycparser: 2.10
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: Not Installed
         Python: 2.7.13 (default, Jan 11 2017, 10:56:06) [GCC]
   python-gnupg: Not Installed
[...]

PR #56595, opened 2020-04-09
Fix for return value ret vs return in batch mode

brejoc:
### What does this PR do?

The least intrusive fix for `ret` vs `return` in batch mode.

https://github.com/saltstack/salt/issues/56322

### Previous Behavior

Returner would only work via cli and not via `LocalClient`.

### New Behavior

Returner works with cli and `LocalClient`.

### Tests written?

Not yet - I'd like to discuss this first.

### Commits signed with GPG?

Yes


brejoc:
Has a test case now. Please re-check. @dwoz

waynew:
@brejoc guess not 🎉

dwoz (approved):

PR #56602, opened 2020-04-10
Sodium is Py3 only. Disable our Pylint Py3 compatible code plugin.

s0undt3ch:
### What does this PR do?
See title

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -128,7 +128,8 @@ disable=R,
   import-outside-toplevel,
   deprecated-method,
   repr-flag-used-in-string,
-  keyword-arg-before-vararg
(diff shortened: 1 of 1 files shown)


s0undt3ch:
This needs a fix

s0undt3ch:
Fix pushed

dwoz (approved):

dwoz:
Don't we want to catch incompatible py3 code since that's all we support?

s0undt3ch:
This plugin was probably badly named.
The purpose was to have py2 code which was compatible with py3.
We now run on py3 only, so, no need for these checks.

dwoz (approved):

PR #56625, opened 2020-04-13
Switch crypto requirements to pycryptodomex on all platforms

### LINKED ISSUE #54115: stop depending on pycrypto

### LINKED ISSUE #56313: salt pip package is using wrong and super outdated crypto library
### Description of Issue
I am working on Manjaro - Arch based Linux distribution.

I wanted to use salt-lint, which unfortunately, asks for salt being installed through python packaging (pip3)

The problem is that the salt package depends on pycrypto, which is at version 2.6.1, which is said, on its home page to have been tested up to python 3.3, and on github, there is no commit since 2014 (!!)

And of course, Arch is using now python 3.8, so I get, from Crypto code:
```
AttributeError: module 'time' has no attribute 'clock'
```

Checking what the distro is using instead of pycrypto - it seems to be pycryptodome, which is a fork/replacement of pycrypto.

You should update the dependency on the python pip package.

### Setup
none

### Steps to Reproduce Issue
have salt-minion try to connect to a master, on a python 3.8 setup, and installed from the python pip package.

### Versions Report
```
Salt Version:
           Salt: 3000
 
Dependency Versions:
           cffi: 1.13.2
       cherrypy: Not Installed
       dateutil: 2.8.1
      docker-py: 4.2.0
          gitdb: 2.0.6
      gitpython: 2.1.1
         Jinja2: 2.11.1
        libgit2: 0.28.4
       M2Crypto: Not Installed
           Mako: 1.1.1
   msgpack-pure: Not Installed
 msgpack-python: 0.6.2
   mysql-python: Not Installed
      pycparser: 2.19
       pycrypto: 3.9.7
   pycryptodome: Not Installed
         pygit2: 1.0.3
         Python: 3.8.1 (default, Jan 22 2020, 06:38:00)
   python-gnupg: Not Installed
         PyYAML: 5.3
          PyZMQ: 18.1.1
          smmap: 2.0.5
        timelib: Not Installed
        Tornado: 4.5.3
            ZMQ: 4.3.2
 
System Versions:
           dist: manjaro 19.0.0 Kyria
         locale: utf-8
        machine: x86_64
        release: 5.4.18-1-MANJARO
         system: Linux
        version: Manjaro Linux 19.0.0 Kyria
```

By the way, it looks to wrongly detect pycrypto as being installed and NOT pycryptodome! (look also at the versions)

s0undt3ch:
### What does this PR do?
See title.
Fixes https://github.com/saltstack/salt/issues/56313
Fixes https://github.com/saltstack/salt/issues/54115

### DIFF
--- a/.ci/kitchen-centos7-py3-pycryptodome
+++ b/.ci/kitchen-centos7-py3-pycryptodome
@@ -2,12 +2,12 @@
 
 runTestSuite(
     concurrent_builds: 1,
-    distro_name: 'ubuntu',
-    distro_version: '1604',
+    distro_name: 'centos',
+    distro_version: '7',
     env: env,
     golden_images_branch: 'master',
     jenkins_slave_label: 'kitchen-slave',
-    nox_env_name: 'runtests-zeromq-pycryptodomex',
+    nox_env_name: 'runtests-zeromq-pycryptodome',
     nox_passthrough_opts: '--ssh-tests',
     python_version: 'py3',
     testrun_timeout: 6,
--- a/.ci/kitchen-centos7-py3-pycryptodomex
+++ b/.ci/kitchen-centos7-py3-pycryptodomex
@@ -1,0 +1,16 @@
+@Library('salt@master-1.7') _
+
+runTestSuite(
+    concurrent_builds: 1,
+    distro_name: 'centos',
+    distro_version: '7',
+    env: env,
+    golden_images_branch: 'master',
+    jenkins_slave_label: 'kitchen-slave',
+    nox_env_name: 'runtests-zeromq-pycryptodome',
+    nox_passthrough_opts: '--ssh-tests',
+    python_version: 'py3',
+    testrun_timeout: 6,
+    use_spot_instances: true)
+
+// vim: ft=groovy
--- a/.ci/kitchen-ubuntu1604-py3-pycryptodome
+++ b/.ci/kitchen-ubuntu1604-py3-pycryptodome
@@ -2,12 +2,12 @@
 
 runTestSuite(
     concurrent_builds: 1,
-    distro_name: 'centos',
-    distro_version: '7',
+    distro_name: 'ubuntu',
+    distro_version: '1604',
     env: env,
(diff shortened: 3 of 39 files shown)


dwoz (dismissed):

.ci/kitchen-centos7-py3-m2crypto
```diff
@@ -1,16 +0,0 @@
-@Library('salt@master-1.7') _
```
Ch3LL, line 1: Why are we removing m2crypto tests as well? A user can still use the m2crypto library or pycryptodomex.

s0undt3ch: From what I understood, the goal, with or without salt-bin, but specially with salt-bin, is to focus on **one** supported crypto library, no?

Ch3LL: If it was a goal to only support only one library I just wasn't aware. But users can still use m2crypto if they are using pip or if they pip install it into the binary. And we try to import it first (https://github.com/saltstack/salt/blob/master/salt/crypt.py#L57).

s0undt3ch: Hmm, maybe it was just wishful thinking.
Let's bring this up in standup today.

dwoz: We still want to support M2Crypto for the time being as far as I recall.

s0undt3ch: PR updated

thatch45 (approved):

PR #56641, opened 2020-04-14
[WIP] Add towncrier to manage CHANGELOG.md

### LINKED ISSUE #55836: research change log use of tools
### Description of Issue
Research ticket task for exploring the use of a change log tools such as towncrier

Ch3LL:
### What does this PR do?
Add towncrier to help manage CHANGELOG.md

Fixes: https://github.com/saltstack/salt/issues/55836

### Previous Behavior
Merge conflicts would occur anytime someone edits the changelog

### New Behavior
no more merge conflicts and at release time we just need to run ` towncrier  --version=3001` for example.

This PR require a contributor to create a new file under the `changelog` dir in the salt repo. The filename will need to be `<issue number>.<type>`. All given types are added, changed, deprecated, fixed, removed. Come release time we will run towncrier and this will combine all the files into the changelog and remove them.

Here is towncrier run with `---draft` with these changes:

```
(salt)  ch3ll@thecakeisalie  ~/git/salt   towncrier  towncrier --draft --version=3001
Loading template...
Finding news fragments...
Rendering news fragments...
Draft only -- nothing has been written.
What is seen below is what would be written.

Salt 3001 (2020-04-14)
======================


Removed
-------

- test remove (#1234)


Deprecated
----------

- test deprecated (#1234)


Changed
-------

- test changed (#1234)


Fixed
-----

- test fixed (#1234)


Added
-----

- test added (#1234)
```

For the Sodium release we will have to do a combination of things. We will have to manually add the PRs we missed before this is added, and run this before the release for hte PRs that do use this.


waynew:
Is it possible/easy to move the issue # before the description? It may not be necessary - I just have a mild preference for that. Probably because I like it lining up better :joy:

waynew:
Otherwise... great job! I was pretty concerned about towncrier based on the readme, but it looks like you've got a solid PoC out of it.

Ch3LL:
I didn't see a way to move the issue number based off the towncrier docs. Since this does not seem to be a big concern I will not worry about it. This is ready for review now. I just need to prep and announcement for this as it will change the workflow for contributors.

garethgreenaway (dismissed):

waynew:
I didn't see it anywhere, but if this removes everything from the changelog dir when building the new changelog, does it just continually prepend to the file then? Or do we have to do something special there?

Ch3LL:
>I didn't see it anywhere, but if this removes everything from the changelog dir when building the new changelog, does it just continually prepend to the file then? Or do we have to do something special there?

I'm not sure i'm following 100%, but let me explain how it works and maybe that will answer your question.

When you run `towncrier` in non-draft mode it compiles all the files in the `changelog` dir and adds it right below `# CHANGELOG`, it then asks you if you want to remove all of the files, which yes we do want to do each release. If those files dont get deleted then they would be added for every release going forward. So we want to make sure we delete those files for each release after they have been added to the CHANGELOG.md file. Then in the next release it will only add any of the new files that were added.

waynew:
Okay, that's what I *thought* would happen, it just wasn't clear - since it was deleting files I wasn't sure if it was *also* deleting/emptying CHANGELOG.md.

waynew (dismissed):

Ch3LL:
should i make it more clear in the docs I wrote?

noxfile.py
```diff
@@ -1129,3 +1129,26 @@ def docs_man(session, compress, update):
...
+@nox.parametrize("draft", [False, True])
+def changelog(session, draft):
+    """
+    Generate salt's changelog
+    """
+
+    pydir = _get_pydir(session)
+    requirements_file = "requirements/static/changelog.in"
+    distro_constraints = [
+        "requirements/static/{}/changelog.txt".format(_get_pydir(session))
```
s0undt3ch, line 1144: Use `pydir` defined above?

noxfile.py
```diff
@@ -1129,3 +1129,26 @@ def docs_man(session, compress, update):
...
+    ]
+    install_command = ["--progress-bar=off", "-r", requirements_file]
+    for distro_constraint in distro_constraints:
+        install_command.extend(["--constraint", distro_constraint])
+    session.install(*install_command, silent=PIP_INSTALL_SILENT)
+
+    town_cmd = ["towncrier", "--version={0}".format(session.posargs[0])]
```
s0undt3ch, line 1151: We're on python 3, and, unless you want the first format item to be repeated, let's drop the formatting index, `{0}`.

s0undt3ch (changes requested):

noxfile.py
Ch3LL: i removed `pydir` instead since i dont use it anywhere else. I did update the calls from the tests above that do use `pydir` but not with `distro_constraints`. I also updated the format call. thanks

s0undt3ch (approved):

PR #56647, opened 2020-04-15
Add gitfs and git_pillar fallback branch

bdrung:
### What does this PR do?

Add a fallback branch when no matching branch is found. This is a forward port of pull request #51971.

### New Behavior

Example:

```yaml
gitfs_remotes:
  - http://foo.com/states/top.git:
    - all_saltenvs: master
  - http://foo.com/states/apache.git:
    - fallback: master
  - http://foo.com/states/nginx.git:
    - fallback: master
  - http://foo.com/states/webapp.git:
    - fallback: develop
```

Where a state in `webapp` includes a state of `apache`.

With this PR, one can test a feature branch of `webapp` without needing to create the same branch in `apache`. Similarly, one can test a feature branch of `apache` without needing to create the same branch in `webapp`.

- [x] Docs
- [ ] Changelog
- [x] Tests written/updated

### DIFF
--- a/salt/utils/gitfs.py
+++ b/salt/utils/gitfs.py
@@ -1073,6 +1073,22 @@ class GitProvider(object):
                 if candidate is not None:
                     return candidate
 
+        if self.fallback:
+            for ref_type in self.ref_types:
+                try:
+                    func_name = "get_tree_from_{0}".format(ref_type)
+                    func = getattr(self, func_name)
+                except AttributeError:
+                    log.error(
+                        "%s class is missing function '%s'",
+                        self.__class__.__name__,
+                        func_name,
+                    )
+                else:
+                    candidate = func(self.fallback)
+                    if candidate is not None:
+                        return candidate
+
         # No matches found
         return None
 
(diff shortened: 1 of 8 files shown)


waynew (approved):

waynew:
Excellent, thanks for porting this to master!

salt/utils/gitfs.py
```diff
@@ -2890,9 +2918,7 @@ def find_file(self, path, tgt_env="base", **kwargs):  # pylint: disable=W0613
         and send the path to the newly cached file
         """
         fnd = {"path": "", "rel": ""}
-        if os.path.isabs(path) or (
-            not salt.utils.stringutils.is_hex(tgt_env) and tgt_env not in self.envs()
-        ):
+        if os.path.isabs(path):
```
terminalmage, line 2921: @dwoz Can you explain this change? This would appear to break existing functionality which allows people to use a SHA as a saltenv in SLS files (or when using `cp.cache_file`).

PR #56031, opened 2020-01-30
Fix unit test on Python 3.8

bdrung:
Salt fails on Python 3.8:

```
======================================================================
ERROR: unit.grains.test_core (unittest.loader._FailedTest)
----------------------------------------------------------------------
ImportError: Failed to import test module: unit.grains.test_core
Traceback (most recent call last):
  File "/usr/lib/python3.8/unittest/loader.py", line 436, in _find_test_path
    module = self._get_module_from_name(name)
  File "/usr/lib/python3.8/unittest/loader.py", line 377, in _get_module_from_name
    __import__(name)
  File "tests/unit/grains/test_core.py", line 37, in <module>
    import salt.grains.core as core
  File "salt/grains/core.py", line 40, in <module>
    from platform import _supported_dists
ImportError: cannot import name '_supported_dists' from 'platform' (/usr/lib/python3.8/platform.py)
```

So only try to import `_supported_dists` from `platform` for Python <=
3.7. Otherwise rely on the external `distro` module to  not need any
special handling.

Addresses parts of #55835

PR #56651, opened 2020-04-15
Add preliminary Python 3.8 static requirements.

s0undt3ch:
### What does this PR do?
See title.
Needs to be revisited for Windows 3.8 support since pywin32 and pymssql do not have packages or support for Py3.8


s0undt3ch:
The requirements files changes are included in https://github.com/saltstack/salt/pull/56031

PR #56659, opened 2020-04-16
Add windows test whitelisting as pytest markers

s0undt3ch:
### What does this PR do?
Add windows test whitelisting as pytest markers

Taken out from the pytest migration PR, https://github.com/saltstack/salt/pull/56460

### DIFF
--- a/tests/integration/modules/test_archive.py
+++ b/tests/integration/modules/test_archive.py
@@ -2,30 +2,25 @@
 """
 Tests for the archive state
 """
-# Import python libs
 from __future__ import absolute_import, print_function, unicode_literals
 
 import os
 import shutil
 import textwrap
 
-# Import salt libs
+import pytest
 import salt.utils.files
 import salt.utils.path
 import salt.utils.platform
 import salt.utils.stringutils
-
-# Import 3rd party libs
 from salt.ext import six
 from tests.support.case import ModuleCase
 from tests.support.helpers import destructiveTest
-
-# Import Salt Testing libs
 from tests.support.runtests import RUNTIME_VARS
 from tests.support.unit import skipIf
 
 try:
-    import zipfile  # pylint: disable=W0611
+    import zipfile  # pylint: disable=unused-import
 
     HAS_ZIPFILE = True
 except ImportError:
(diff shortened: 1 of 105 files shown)


tests/integration/modules/test_archive.py
```diff
@@ -2,44 +2,46 @@
...
-    # Base path used for test artifacts
-    base_path = os.path.join(RUNTIME_VARS.TMP, "modules", "archive")
+    @classmethod
+    def setUpClass(cls):
+        # Base path used for test artifacts
+        cls.base_path = os.path.join(RUNTIME_VARS.TMP, "modules", "archive")
+
+    @classmethod
+    def tearDownClass(cls):
+        cls.base_path = None
```
waynew, line 44: Why does this need to be setup or torn down?

tests/integration/returners/test_librato_return.py
```diff
@@ -2,62 +2,65 @@
...
+    @classmethod
+    def tearDownClass(cls):
+        cls.mock_ret_obj = None
+
     def test_count_runtimes(self):
         """
         Test the calculations
         """
-        results = librato_return._calculate_runtimes(MOCK_RET_OBJ["return"])
+        results = librato_return._calculate_runtimes(self.mock_ret_obj["return"])
```
waynew, line 63: Does this `mock_ret_obj` change need to be in this PR? Also, since pytest, shouldn't this just be a pytest fixture?

waynew (changes requested):
Just a couple of changes I'm not sure are necessary in this PR, otherwise :+1:

tests/integration/modules/test_archive.py
s0undt3ch: Since it's a class attribute, it's less "garbage" to keep in memory until the class is, eventually, GC'ed

tests/integration/returners/test_librato_return.py
s0undt3ch: > Does this mock_ret_obj change need to be in this PR? 

Module level globals, specially dictionaries, are evil. If you really want me to take this into the "BIG PR", I'll take it out.

> Also, since pytest, shouldn't this just be a pytest fixture?
When we have a full pytest only test suite, sure. I'm not going to migrate every single test I touch, at least just yet :grinning:

waynew: I think if you pull these two changes into a smaller PR - or even one that's a bit larger, that's doing a bunch of this kind of cleanup - that would be better. If this one is *just* adding windows whitelisting markers, that's :+1: 

Then we can have other(s) that are doing other assorted cleanup.

s0undt3ch: backed out

waynew (approved):

PR #56661, opened 2020-04-16
Last bit to disable codecov's PR status checks

s0undt3ch:
### What does this PR do?
See title

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -7,9 +7,8 @@ codecov:
 #  require_ci_to_pass: yes      # Less spammy. Only notify on passing builds.
 
 
-  notify: off                    # Disable Notifications
 #  notify:
-#    after_n_builds: 46           # Only notify after N builds
+#    after_n_builds: 25           # Only notify after N builds
 #                                 # This value is the output of:
 #                                 #    sh -c 'echo "$(ls .ci/ | grep kitchen | wc -l)"'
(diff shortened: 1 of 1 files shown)


Ch3LL (dismissed):

Ch3LL (dismissed):

garethgreenaway (dismissed):

waynew:
I would prefer changing it to `informational: True` - then we still get the information, but it doesn't (shouldn't?) block the PR

s0undt3ch:
@waynew agreed and updated

waynew (approved):
🚀

Ch3LL:
will this still show failing tests? If i remember previously, when I try to click on codecov's results I couldn't even see them, so the failing test seemed unnecessary to include

garethgreenaway (approved):

s0undt3ch:
Not tests, _per se_, you're referring to codecov's reports.

s0undt3ch:
Some will render fine, most will, still, timeout

Ch3LL:
yeah tests wasn't the right word to say. I mean a failing job on the PR- are we going to see a RED job if codecov fails. If this is the case and they are timing out and we can't rely on them why keep them to informational then and instead disable it all together until its resolved?

s0undt3ch:
We should stop seeing the RED for codecov on PRs, according to their docs.

Ch3LL (approved):

PR #56674, opened 2020-04-16
test updating docs and lint to python 3.7.6

bryceml:
### What does this PR do?
tests if we can update docs and lint tests to python 3.7.6


s0undt3ch:
Test successful.

PR #56678, opened 2020-04-17
Add new function for comparing package versions

terminalmage:
This exposes the functionality from `salt.utils.versions.compare()`, allowing it to be used in places like jinja templates.


Nerigal:
@terminalmage  thank you !

dwoz (approved):

Issue #56654, opened 2020-04-15
[BUG] Centos7 pkg.installed issue when trying to install a package that is already at its latest version


saltybaker:
**Description**
In Centos7 pkg.installed issue when trying to install a package that is already at its latest version

**Setup**
stuff.sls

bind-libs-lite:
  pkg:
    - installed
    - version: '9.11.4'

**Steps to Reproduce the behavior**
[root@master2 salt]# salt minion2 pkg.latest_version bind-libs-lite
[WARNING ][salt.client                              :1073][MainProcess:12806] jid does not exist
[WARNING ][salt.client                              :1078][MainProcess:12806] Returner unavailable: 
minion2:

This is expected if bind-libs-lite is already installed with the version latest available version.

So we try pkg.info_installed to get the version of the installed package
[root@master2 salt]# salt minion2 pkg.info_installed bind-libs-lite
[WARNING ][salt.client                              :1073][MainProcess:12818] jid does not exist
[WARNING ][salt.client                              :1078][MainProcess:12818] Returner unavailable: 
minion2:
    ----------
    bind-libs-lite:
        ----------
        arch:
            x86_64
        build_date:
            2019-08-08T12:20:30Z
        build_date_time_t:
            1565266830
        build_host:
            x86-02.bsys.centos.org
        description:
            Contains lite version of BIND suite libraries which are used by various
            programs to work with DNS protocol.
        epoch:
            32
        group:
            Unspecified
        install_date:
            2020-04-13T18:56:29Z
        install_date_time_t:
            1586804189
        license:
            MPLv2.0
        packager:
            CentOS BuildSystem <http://bugs.centos.org>
        release:
            9.P2.el7
        relocations:
            (not relocatable)
        signature:
            RSA/SHA256, Thu Aug 22 21:20:49 2019, Key ID 24c6a8a7f4a80eb5
        size:
            3091672
        source:
            bind-9.11.4-9.P2.el7.src.rpm
        summary:
            Libraries for working with the DNS protocol
        url:
            http://www.isc.org/products/BIND/
        vendor:
            CentOS
        version:
            9.11.4
[root@master2 salt]#

This confirms that bind-libs-lite is already installed with version 9.11.4
So lets apply the state.

[root@master2 salt]# salt minion2 state.apply stuff
[WARNING ][salt.client                              :1073][MainProcess:12772] jid does not exist
[WARNING ][salt.client                              :1078][MainProcess:12772] Returner unavailable: 
minion2:
----------
[...]

PR #56681, opened 2020-04-17
Implicitly ignore epoch when no epoch in desired version, and ignore_epoch not passed

terminalmage:
### What does this PR do?

A usability tweak for `pkg` states on platforms that support pkg epochs.

In the process of adding a test case for this, I encountered a major testing oversight that was merged on Jan 9th in https://github.com/saltstack/salt/pull/55707. By moving the lists of package targets into the class, they weren't defined until _after_ the decorators had already ran. This means that most of these decorated tests were being skipped all the time, on all platforms. Not great, Bob!

I've moved the skip logic back within the actual tests, where they were before the decorators were added last autumn in 291579c02b630e453a4ec90936218964bd5c5f89.

#56654 

https://github.com/saltstack/salt/pull/55707

### Previous Behavior

Epochs were strictly enforced in package comparisons (i.e. `2.3-4` would not be a match if the actual version of the package was `1:2.3-4`), unless `ignore_epoch` was set to `True`.

### New Behavior

Now, when no value is explicitly passed this argument, the epoch will be implicitly skipped if the desired version does not itself contain an epoch. This will allow for much more natural version handling, instead of requiring the user to always know beforehand when they are dealing with a package that has an epoch in the version.

- [x] Docs
- [ ] Changelog
- [x] Tests written/updated

### Commits signed with GPG?
No

### DIFF
--- a/salt/modules/yumpkg.py
+++ b/salt/modules/yumpkg.py
@@ -1722,7 +1722,7 @@ def install(
             cmd.extend(targets)
             out = _call_yum(cmd, ignore_retcode=False, redirect_stderr=True)
             if out["retcode"] != 0:
-                errors.append(out["stdout"])
+                errors.append(out["stderr"])
 
     targets = []
     with _temporarily_unhold(to_downgrade, targets):
@@ -1733,7 +1733,7 @@ def install(
             cmd.extend(targets)
             out = _call_yum(cmd)
             if out["retcode"] != 0:
-                errors.append(out["stdout"])
+                errors.append(out["stderr"])
 
     targets = []
     with _temporarily_unhold(to_reinstall, targets):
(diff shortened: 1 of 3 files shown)


terminalmage:
One other thought: should I make similar changes in the execution modules (and salt.utils.versions) for their `version_cmp` funcs? I don't know if it's that important, since I doubt many people are going to be running these functions directly. I did however just add a new function for package version comparison in https://github.com/saltstack/salt/pull/56678. Maybe it would make sense to replicate the same `ignore_epoch` behavior from this PR into that function, since it's what people are most likely to be using to do version comparisons, in jinja.

What do you think?

terminalmage:
On the other hand, epochs are specifically designed for version comparison, so maybe they don't make much sense to ignore. There's a much bigger case to be made to implicitly ignore them in states where one is likely to put a version number into their `pkg.installed` state and not necessarily know that the package has an epoch.

If you are explicitly making a version comparison in Jinja, I think it's more reasonable to expect you to take epochs into account. So perhaps I shouldn't make that change in the other PR.

cmcmarrow (approved):
Looks good. Thanks for the work

PR #52516, opened 2019-04-11
make multimaster minionmanager create master_uri_list

mattp-:
### What does this PR do?
a similar snippet of code exists in minion tune_in for letting
masterless mode work / callerClient, unfortunately it isn't hit in
multimaster because multimaster creates N minion objects with a single master: ip
value. This fills in that deficiency.

### Tests written?
no

### Commits signed with GPG?
No

Please review [Salt's Contributing Guide](https://docs.saltstack.com/en/latest/topics/development/contributing.html) for best practices.

See GitHub's [page on GPG signing](https://help.github.com/articles/signing-commits-using-gpg/) for more information about signing commits with GPG.

PR #56693, opened 2020-04-18
Port #52516 to master

terminalmage:
See #52516.


terminalmage:
Closing since there is already a pending backport

PR #56701, opened 2020-04-18
Add feature request 51919 (wait for static IP during VM creation in VMware)

### LINKED ISSUE #51919: Feature request: salt-cloud vmware ip handaling

alfotos:
Fixes #51919

This change adds the functionality of waiting for predefined static IPs
when deploying a new VM in VMWare. Instead of picking the first available
IP assigned to a VM, it waits for a certain IP to be used in the VM.
This is controlled by the parameter "wait_for_ip_priority".
Old behaviour still remains if parameter is set to "any".

### What does this PR do?
Implement new feature in the VMware cloud module to give the option to wait for a pre-defined static IP when creating a new VM. The new parameter "wait_for_ip_priority" can be set to "any", "dhcp", "static" or "ip" to control the behaviour of the "_wait_for_ip" function.

Fixes #51919

### Previous Behavior
At the moment, the _wait_for_ip uses the first IP configured in the VM during the deployment. This can be an issue if you are deploying a VM in a DHCP-enabled subnet but you want to use a given static IP. This can lead to a timeout later, as Salt will try to connect to an IP which was only used fore few seconds (time between the DHCP IP and the static IP configured by VMware tools). More details are explained in the issue #51919

### New Behavior
The current behaviour is maintained by default, or by setting the "wait_for_ip_priority" to "dhcp" or "any". On the other hand, if you have a static IP and want to be sure that one is used the parameter can be set to "static", this will filter the registered IP and will only accept one of the IPs defined in the network devices. The specific use case of this is "ip", when you also have to use the parameter "wait_for_ip_address".

- [ ] Docs
- [ ] Changelog - https://docs.saltstack.com/en/latest/topics/development/changelog.html
- [ ] Tests written/updated

### Commits signed with GPG?
No


Ch3LL:
looks like there is some lint failures here. Can you also write a regression test for this?

DmitryKuzmenko:
@alfotos could you please follow this up?

Ch3LL (changes requested):
bump for the requests in this comment https://github.com/saltstack/salt/pull/56701#issuecomment-622402984

Ch3LL:
@alfotos any updates here?

dwoz:

PR #51202, opened 2019-01-16
Lint: Fix str-format-in-logging, len-as-condition, etc. (3)

terminalmage:
This fixes violations found in salt/modules/ (A-M)

PR #56702, opened 2020-04-18
Port #51202 to master

terminalmage:
See https://github.com/saltstack/salt/pull/51202


dwoz:
Looks like the file test failure is related, bad test.

terminalmage:
OK, I'll have a look tonight, thanks

terminalmage:
@dwoz Should be fixed now. The test was using a mocked logger and checking what was passed to the `log.warning()` call, which before this PR would have been the full log message (since the logged string was formatted using `.format()`), but is now just a string with `%s` placeholders.

dwoz (approved):

PR #51096, opened 2019-01-08
Feature to enable/disable termination protection for ec2 instances

AmitChotaliya:
### What does this PR do?

- Adds termination protection option to salt-cloud ec2 cofiguration

### What issues does this PR fix or reference?
Fixes: #45496

### Tests written?

No

### Commits signed with GPG?

Yes

Please review [Salt's Contributing Guide](https://docs.saltstack.com/en/latest/topics/development/contributing.html) for best practices.

See GitHub's [page on GPG signing](https://help.github.com/articles/signing-commits-using-gpg/) for more information about signing commits with GPG.

PR #56703, opened 2020-04-18
Port #51096 to master

terminalmage:
See https://github.com/saltstack/salt/pull/51096


terminalmage:
NOTE that this PR has already been ported to master, with the exception of the small docstring change that remains.

dwoz (approved):

PR #56704, opened 2020-04-18
Port #51033 to master

terminalmage:
See https://github.com/saltstack/salt/pull/51033


terminalmage:
I've added a testcase and also made a slight tweak. This was limited to Python 3 only because it looks like the original author of this PR was unaware of how to force `salt.utils.vt` to write a `str` type irrespective of which version of Python (something that was fixed by @garethgreenaway in ffdae27877b23dcb06a7f8b76682548f2198845e around the same time the original PR was merged).

dwoz (approved):

PR #56027, opened 2020-01-30
Replace deprecated inspect.formatargspec

bdrung:
Python 3.7 raises a deprecation warning:

salt/utils/decorators/signature.py:31: DeprecationWarning: `formatargspec` is deprecated since Python 3.5. Use `signature` and the `Signature` object directly

`inspect.formatargspec` is only used in `salt.utils.decorators.signature.identical_signature_wrapper` which is only used in `salt.utils.decorators.path` for decorating the `which` and `which_bin` functions. The function `identical_signature_wrapper` can be simply replaced by Python's `functools.wraps` which is available since at least Python 2.7.

When inspecting those wrapped functions, the underlying function (stored in the `__wrapped__` attribute) needs to be inspect instead.

fixes #50911

PR #56713, opened 2020-04-19
Rewrite PY3-specific code that uses argspec functions from inspect module

### LINKED ISSUE #48735: getfullargspec and annotations
I've got a strange error when I use type hinting for my salt module:

```
  The minion function caused an exception: Traceback (most recent call last):  File "/opt/alt/python35/lib/python3.5/site-packages/salt/minion.py", line 1589, in _thread_return
    data)
  File "/opt/alt/python35/lib/python3.5/site-packages/salt/minion.py", line 323, in load_args_and_kwargs
    argspec = salt.utils.args.get_function_argspec(func)
  File "/opt/alt/python35/lib/python3.5/site-packages/salt/utils/args.py", line 265, in get_function_argspec
    aspec = _getargspec(func)  # pylint: disable=redefined-variable-type
  File "/opt/alt/python35/lib/python3.5/site-packages/salt/utils/args.py", line 225, in _getargspec
    raise ValueError(\'Function has keyword-only arguments or annotations\'
   ValueError: Function has keyword-only arguments or annotations, use getfullargspec() API which can support them
   
'}}
```

The problem is that you check function annotations and raise an error when it has some annotations:

`site-packages/salt/utils/args.py`
```
   def _getargspec(func):
        '''
        Python 3 wrapper for inspect.getargsspec

        inspect.getargsspec is deprecated and will be removed in Python 3.6.
        '''
        args, varargs, varkw, defaults, kwonlyargs, _, ann = \
            inspect.getfullargspec(func)  # pylint: disable=no-member
        if kwonlyargs or ann:
            raise ValueError('Function has keyword-only arguments or annotations'
                             ', use getfullargspec() API which can support them')
        return _ArgSpec(args, varargs, varkw, defaults)
```

The code looks strange! There is a warning about `getargsspec` but you don't use `getargsspec`. There is exception message that give you advice to use `getfullargspec`, but you already use `getfullargspec`.

And I still don't know why do you raise an error if the function has an annotations. What is the reason?

### Versions Report
salt-minion 2018.3.0 (Oxygen)

terminalmage:
*This pull request was inspired by https://github.com/saltstack/salt/pull/56027.*

### What does this PR do?

[`inspect.getargspec()`](https://docs.python.org/3/library/inspect.html#inspect.getargspec) and [`inspect.getfullargspec()`](https://docs.python.org/3/library/inspect.html#inspect.getfullargspec) are deprecated in Python 3 in favor of [`inspect.Signature`](https://docs.python.org/3/library/inspect.html#inspect.Signature).

To ensure consistent behavior moving forward with Python 3, this PR does the following:

1. Modifies `salt.utils.args` such that the PY3-specific code to build the namedtuple now uses [`inspect.signature()`](https://docs.python.org/3/library/inspect.html#inspect.signature).

2. Updates all calls to [`inspect.getargspec()`](https://docs.python.org/3/library/inspect.html#inspect.getargspec) and [`inspect.getfullargspec()`](https://docs.python.org/3/library/inspect.html#inspect.getfullargspec) elsewhere in the code to ensure that they use the proper PY3-specific code when running under PY3.

Refs: https://github.com/saltstack/salt/pull/56027

Docs are unnecessary, as this doesn't change how one uses Salt.

Additionally, new tests are not needed because there are already unit tests that confirm proper operation of `salt.utils.args.get_function_argspec()`, and these tests pass with the new code.

### Commits signed with GPG?
No

### DIFF
--- a/salt/utils/args.py
+++ b/salt/utils/args.py
@@ -290,6 +290,8 @@ def get_function_argspec(func, is_class_method=None):
                     varargs = param.name
                 elif param.kind == param.VAR_KEYWORD:
                     keywords = param.name
+            if is_class_method:
+                del args[0]
             aspec = _ArgSpec(args, varargs, keywords, tuple(defaults) or None)
 
     return aspec
--- a/salt/utils/args.py
+++ b/salt/utils/args.py
@@ -268,7 +268,30 @@ def get_function_argspec(func, is_class_method=None):
         aspec = _getargspec(func.__call__)
         del aspec.args[0]  # self
     else:
-        raise TypeError("Cannot inspect argument list for '{0}'".format(func))
+        try:
+            sig = inspect.signature(func)
+        except TypeError:
+            raise TypeError("Cannot inspect argument list for '{0}'".format(func))
+        else:
+            # argspec-related functions are deprecated in Python 3 in favor of
+            # the new inspect.Signature class, and will be removed at some
+            # point in the Python 3 lifecycle. So, build a namedtuple which
+            # looks like the result of a Python 2 argspec.
+            _ArgSpec = namedtuple("ArgSpec", "args varargs keywords defaults")
+            args = []
+            defaults = []
+            varargs = keywords = None
+            for param in sig.parameters.values():
+                if param.kind == param.POSITIONAL_OR_KEYWORD:
+                    args.append(param.name)
+                    if param.default is not inspect._empty:
+                        defaults.append(param.default)
+                elif param.kind == param.VAR_POSITIONAL:
+                    varargs = param.name
+                elif param.kind == param.VAR_KEYWORD:
+                    keywords = param.name
+            aspec = _ArgSpec(args, varargs, keywords, tuple(defaults) or None)
+
     return aspec
 
 
--- a/salt/modules/napalm_mod.py
+++ b/salt/modules/napalm_mod.py
@@ -10,7 +10,6 @@ Helpers for the NAPALM modules.
 from __future__ import absolute_import, print_function, unicode_literals
 
 # Import python stdlib
-import inspect
 import logging
 
 # import NAPALM utils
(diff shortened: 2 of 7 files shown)


DmitryKuzmenko:
@terminalmage thank you for PR! It looks some failures are related.

terminalmage:
@DmitryKuzmenko I pushed a commit to fix the failing test case. I also checked and realized the testing for this was not as good as I first thought, so I expanded the test cases for get_function_argspec.

DmitryKuzmenko:
Thank you! Let's check it in the morning.

salt/utils/args.py
```diff
@@ -23,6 +23,9 @@
 from salt.ext import six
 from salt.ext.six.moves import map, zip
 
+if six.PY3:
```
s0undt3ch, line 26: Master is PY3.
So just import it.

s0undt3ch (changes requested):

s0undt3ch:
On a side note, what's the gain of using `__utils__` instead of just importing the module and calling the function?
It's not like the function uses `__context__`...
Not blocking the merge because of this, I'm genuinely interested in the why.

salt/utils/args.py
terminalmage: Sorry, master what?

It's being imported because it's only used in for PY3 within `get_function_argspec()`. Maybe it makes more sense to do a late import within `get_function_argspec()`.

terminalmage:
> On a side note, what's the gain of using `__utils__` instead of just importing the module and calling the function?
> It's not like the function uses `__context__`...
> Not blocking the merge because of this, I'm genuinely interested in the why.

Because this allows for people to deploy their own versions of the utils modules, to do things like deploying bugfixes to those modules via `saltutil.sync_utils`.

salt/utils/args.py
s0undt3ch: The master branch is no longer a Py2 branch, though _it can_ support Py2.
So, this code will, in fact, always run under Py3, that's why I said not to gate it behind `six.PY3`

s0undt3ch:
> Because this allows for people to deploy their own versions of the utils modules, to do things like deploying bugfixes to those modules via saltutil.sync_utils.
Meh, ok.

salt/utils/args.py
terminalmage: OK, I was confused and thought you were referring somehow to `salt-master`. I've updated the import.

s0undt3ch (dismissed):

DmitryKuzmenko (dismissed):
I see nothing that could be related to the `integration.modules.test_cmdmod.CMDModuleTest` so I'm approving this and re-starting the builds.

dwoz (approved):

dwoz:
re-run full all

PR #56723, opened 2020-04-20
Terminate subprocesses when main proces is terminated.

### LINKED ISSUE #48390: SIGTERM -- Return error when salt-minion service/process receives a SIGTERM -- Feature Request

Currently, when salt-minion service/process receives a SIGTERM, it just exits. 
It leaves all the running jobs in orphan state. 

**Feature-Request**: Please return some kind of errors for those jobs before stopping the minion.

DmitryKuzmenko:
### What does this PR do?
When minion's main process receives the SIGTERM or SIGINT it doesn't shut down the worker processes executing jobs.
While if minion is stopped by "Ctrl-C" from terminal it shuts down all the processes.
This PR removes this inconsistency.

Also I've fixed the `SubprocessList.cleanup` method where iteration was broken.
Also fixed Lint failure: `third party module 'psutil' is not surrounded by try/except` that appeared from pre-commit hook.

Fixes: https://github.com/saltstack/salt/issues/48390

### Previous Behavior
Run
```bash
for i in $(seq 3); do salt '*' test.sleep 100 --async; done
```
This will run 3 async jobs on the salt minion. Now send the SIGTERM to the only main process:
```bash
pkill -f salt-minion$
```
Result: the minion reports about stopping the logging queue and the main minion process. Then it hangs for 100 seconds until the last job is done. Then exits.
Re-run the same but run minion in foreground and hit `Ctrl-C`. Minion exits immediately.

### New Behavior
Result: minion exits immediately terminating the jobs.

- [ ] Docs
- [ ] Changelog - https://docs.saltstack.com/en/latest/topics/development/changelog.html
- [x] Tests written/updated

### Commits signed with GPG?
Yes


bryceml:
re-run all full

Ch3LL (dismissed):

Ch3LL:
re-run all full

DmitryKuzmenko:
The tests are broken by `filename_map.yml`. Fix is here: #57004

DmitryKuzmenko:
It looks tests timeout issue could be related. I'm checking it.

s0undt3ch (dismissed):

tests/unit/utils/test_process.py
```diff
@@ -535,16 +531,42 @@ def test_process_list_cleanup(self):
...
+        assert proc2 in plist.processes
         self.wait_for_proc(proc)
+        self.wait_for_proc(proc2)
         assert not proc.is_alive()
+        assert not proc2.is_alive()
         plist.cleanup()
         assert proc not in plist.processes
+        assert proc2 not in plist.processes
+
+    @skipIf(True, "SLOWTEST skip")
```
s0undt3ch, line 553: Let's add instead the slow test decorator?

s0undt3ch (changes requested):

tests/unit/utils/test_process.py
DmitryKuzmenko: Sure! Thank you @s0undt3ch !

DmitryKuzmenko: I wonder how it could be missed. :shrug:. Fixed.

s0undt3ch (dismissed):

dwoz:

PR #56741, opened 2020-04-20
[PyTest #33 - Extract] There's really no need to setup clear_func for each test.

s0undt3ch:
### What does this PR do?
This makes it so that all tests in this module take ~30 seconds to run
as opposed to ~20s for each test


dwoz (approved):

dwoz:
10 seconds to setup ClearFuncs is alarming.

s0undt3ch:
When I made the change, yep, that was about the time it took.

PR #56745, opened 2020-04-20
[PyTest #33 - Extract] Better handling of `close()` when called from `__del__`

s0undt3ch:
### What does this PR do?
See title and:
```
Exception AttributeError: "'AsyncReqMessageClient' object has no attribute '_closing'" in <bound method AsyncReqMessageClient.__del__ of <salt.transport.zeromq.AsyncReqMessageClient object at 0x1a0e82e90>> ignored
```

### DIFF
--- a/salt/transport/zeromq.py
+++ b/salt/transport/zeromq.py
@@ -1234,25 +1234,30 @@ class AsyncReqMessageClient(object):
 
     # TODO: timeout all in-flight sessions, or error
     def close(self):
-        if self._closing:
+        try:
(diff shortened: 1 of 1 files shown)


dwoz:
We really need to do away with __del__, it shouldn't be that far of a stretch to find all the instantiations of transports and handle them explicitly (along with doing away with the singleton stuff).

dwoz (approved):

s0undt3ch:
> We really need to do away with `__del__`

Yes, we do.

PR #56773, opened 2020-04-21
Port #51466 to master

s0undt3ch:
### What does this PR do?
See title

#51466

- [x] Tests written/updated

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/modules/nilrt_ip.py
+++ b/salt/modules/nilrt_ip.py
@@ -488,14 +488,12 @@ def _dict_to_string(dictionary):
 
 def _get_info(interface):
     """
-    Return information about an interface even if it's not associated with a service.
+    Return information about an interface if it's associated with a service.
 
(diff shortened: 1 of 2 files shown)


waynew:
This PR is rather large and involved, and looking at the history of the original PR I'm not sure there's enough information or details for me to be able to write meaningful tests here.

I *could* write tests against this PR, but lacking a higher level understanding, the tests wouldn't be much more than just cargo cult tests. They would just be based on the implementation, regardless of how correct it is, and really just be brittle tests.

If someone is more familiar with the underlying system, so we could write some meaningful tests here, I'd be happy to help - otherwise I'm going to drop this to the bottom of my priority list.

twangboy (changes requested):
This needs a changelog. I don't think the modified test is testing this change. Also needs to be rebased and conflicts fixed

PR #56843, opened 2020-04-22
Fix SyntaxWarning in salt.utils.schedule

terminalmage:
This fixes the following warning:

```
/testing/salt/utils/schedule.py:775: SyntaxWarning: "is not" with a literal. Did you mean "!="?
  if key is not "kwargs":
```

### DIFF
--- a/.ci/kitchen-fedora32-py3
+++ b/.ci/kitchen-fedora32-py3
@@ -1,0 +1,16 @@
+@Library('salt@master-1.7') _
+
+runTestSuite(
+    concurrent_builds: 1,
+    distro_name: 'fedora',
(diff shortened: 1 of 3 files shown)


terminalmage:
@dwoz This is a single-line change to a comparison, for which the incorrect syntax still works (but generates a warning). I'm not sure how one would test this. Do you have ideas? There's no function call to mock check how it was invoked.

The best way to address something like this, I think, would be to use pylint.

terminalmage:
It looks like this was fixed separately by Pedro in https://github.com/saltstack/salt/pull/56931 (oddly enough, without a testcase), so this PR is now redundant.

PR #56852, opened 2020-04-22
Master port 52156

joechainz:
https://github.com/saltstack/salt/pull/52156
### What does this PR do?
Implements longstanding request #8875 - allows for pillar relative includes

#8875 - there are probably others

### Previous Behavior
A leading dot in a pillar include was a nop. If you had a pillar like:

foo/bar/baz.sls

include:
  - .bang
That would not refer to foo/bar/bang.sls, but bang.sls at the pillar root.

### New Behavior
Now leading dots in includes refer to relative imports 🎉

### Tests written?
Yes

### Commits signed with GPG?
Yes


waynew:
I actually just ported this myself 😬 - https://github.com/saltstack/salt/pull/56851

dwoz (approved):

PR #56853, opened 2020-04-22
Port #49886 and #52709 to master

DmitryKuzmenko:
Port #49886 and #52709 to master

### localclient: pass input args/kwargs through salt.utils.args.parse_input #49886

this functionality brings localclient in line with what happens on the
cli, as well as what runnerclient does. this ensures that any foo=bar
kwargs passed in as args are handled accordingly. right now localclient
relies on SaltCMDOptionParser to do this handling for us, but that
leaves a hole when salt-api interactions are brought in.

### LocalClient: don't parse input twice and test fix. #52709

Related to #49886 where passing input args and kwargs through parse_input was inroduced.

I've removed passing the args to parse_input from all functions except run_job because other functions aren't used that args and all are calling run_job. This avoids parsing that args multiple times.

Also I've updated the CMDModuleTest according to parse_input expectations. It yamlifies each text value in arguments so it expects arguments values to be a YAML code. This means that multiline blocks of python code shall be formatted accordingly.
Reference: https://yaml.org/spec/1.2/spec.html#id2795688
Examples and human friendly description: https://yaml-multiline.info/#block-scalars

### DIFF
--- a/salt/client/__init__.py
+++ b/salt/client/__init__.py
@@ -340,7 +340,7 @@ class LocalClient(object):
             >>> local.run_job('*', 'test.sleep', [300])
             {'jid': '20131219215650131543', 'minions': ['jerry']}
         """
-        arg = salt.utils.args.condition_input(arg, kwarg)
+        arg = salt.utils.args.parse_input(arg, kwargs=kwarg)
 
         try:
             pub_data = self.pub(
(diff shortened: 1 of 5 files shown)


DmitryKuzmenko:
This PR introduces a difference in parsing the input that is the reason of the `cmd.run` tests failures. And this behavior change looks desired so I'm planning to update the test.
@mattp- @saltstack/team-core am I right?

mattp-:
@DmitryKuzmenko yep that sounds about right

dwoz (approved):

PR #56880, opened 2020-04-24
Always run check-docs hook to detect deleted files.

s0undt3ch:
### What does this PR do?
See title.
https://github.com/pre-commit/pre-commit/issues/749#issuecomment-618964503

 /cc @max-arnold

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -39,6 +39,9 @@ jobs:
           repo:
             - added|modified:
               - '**'
+          deleted:
+            - deleted:
+              - '**'
 
     - name: Check ALL Files On Branch
       uses: pre-commit/action@v2.0.0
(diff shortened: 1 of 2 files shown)


s0undt3ch:
I actually would prefer we don't set this to always run because the hook is not fast enough.
Let's see what reply we get from @asottile about https://github.com/pre-commit/pre-commit/issues/749#issuecomment-618964503

s0undt3ch:
This was reworked to only run under CI since it's an expensive hook to run locally

s0undt3ch:
Current behavior:
* Only run `check-docs` on deleted files on CI.
* The hook fails when we delete a module and don't delete the corresponding docs

twangboy (approved):

ScriptAutomate (approved):

Ch3LL (approved):

PR #56904, opened 2020-04-27
Rework static requirements

s0undt3ch:
### What does this PR do?
All requirements are now compiled to static requirements.
This includes Windows and Darwin packaging requirements.

Refs https://github.com/saltstack/release-planning/issues/238
Fixes https://github.com/saltstack/salt/pull/56335
Fixes https://github.com/saltstack/salt/pull/56048


s0undt3ch:
Conflicts solved again.

twangboy:
Rebase?

requirements/static/ci/py3.5/windows-crypto.txt
```diff
@@ -0,0 +1,9 @@
+#
+# This file is autogenerated by pip-compile
+# To update, run:
+#
+#    pip-compile -o requirements/static/ci/py3.5/windows-crypto.txt -v requirements/crypto.txt requirements/static/ci/crypto.in
+#
+m2crypto==0.35.2
```
twangboy, line 7: I don't think there are wheels available for 35.2 on Windows...

twangboy: Nope. Pretty sure M2Crypto won't build off the source with pip on Windows. That's why I've been bundling pycryptodome on Windows packages.

setup.py
```diff
@@ -117,140 +118,81 @@
...
-            if IS_WINDOWS_PLATFORM:
-                if "libcloud" in line:
-                    continue
-            if IS_PY3 and "futures" in line.lower():
-                # Python 3 already has futures, installing it will only break
-                # the current python installation whenever futures is imported
+            if line.startswith("-r"):
+                include_requirements_file = os.path.join(
+                    os.path.dirname(requirements_file),
+                    line.split("-r")[1].split("#")[0].strip(),
```
DmitryKuzmenko, line 189: I'm not sure, but isn't there a chance to have a file name with spaces? Maybe `line.split(' ', 1)`?

Ch3LL (changes requested):
looks like this requires a rebase

s0undt3ch:
@twangboy I need your review on the updated macOS and Windows builds scripts.

twangboy (dismissed):

twangboy (dismissed):
Looks good on Windows and Mac

s0undt3ch:
re-run full all

s0undt3ch:
re-run full all

s0undt3ch:
re-run full all

s0undt3ch:
re-run full all

Ch3LL (dismissed):
Is there documentation that covers how to update dependencies for packages and test dependencies? If not we should add some, although does not need to be in this PR.

Also, will salt-pkg need to be udpated, because it is currently pulling in the requirements here: https://gitlab.com/saltstack/open/salt-pkg/-/blob/master/redhat/sources/requirements.txt

s0undt3ch:
> Is there documentation that covers how to update dependencies for packages and test dependencies? If not we should add some, although does not need to be in this PR.
No and we should add some.

> Also, will salt-pkg need to be udpated, because it is currently pulling in the requirements here: https://gitlab.com/saltstack/open/salt-pkg/-/blob/master/redhat/sources/requirements.txt
Yes.

PR #56914, opened 2020-04-27
 [PyTest #33 - Extract] Add invoke task to mark slow tests based on jenkins test report data

s0undt3ch:
### What does this PR do?
See title

### DIFF
--- a/tasks/tests.py
+++ b/tasks/tests.py
@@ -1,0 +1,348 @@
+# -*- coding: utf-8 -*-
+"""
+    tasks.tests
+    ~~~~~~~~~~~
+
+    Mark slow tests based on Jenkins test reports
+"""
+
+import os
+import pathlib
+import sys
+
+from invoke import task  # pylint: disable=3rd-party-module-not-gated
+from tasks import utils
+
+try:
+    import jenkins
+except ImportError:
+    utils.error("pip install python-jenkins")
+    utils.exit_invoke(1)
+
+CODE_DIR = pathlib.Path(__file__).resolve().parent.parent
+SALT_CODE_DIR = CODE_DIR / "salt"
+TESTS_CODE_DIR = CODE_DIR / "tests"
+
+
+@task
+def mark_slow(ctx, pr_number=None, branch="master", username=None, password=None):
+    slowest_test_times = {}
+
+    if username is None:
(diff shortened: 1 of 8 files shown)


tasks/tests.py
```diff
@@ -0,0 +1,348 @@
...
+
+    if username is None:
+        username = os.environ.get("JENKINS_USERNAME")
+
+    if password is None:
+        password = os.environ.get("JENKINS_PASSWORD")
+
+    client = jenkins.Jenkins(
+        "https://jenkinsci.saltstack.com/", username=username, password=password
+    )
```
waynew, line 39: Do we really need to collect data from Jenkins?

While I do think it's a neat idea the point of skipping slow tests on PR is that we should have solid test coverage within the fast tests.

If we're looking at Jenkins then we're basically codifying the idea that slow tests are :ok_hand: and should be expected as part of our PR infrastructure. I don't agree with that idea.

The goal is that **we should have zero speed skips** on PR tests.

Skipping tests *should* be a manual marker because they're just temporary until we have the time to either turn them into proper quick/unit tests, or get a more comprehensive test suite as part of our build pipeline.

I think we should just "manually" mark the tests and leave it at that.

waynew (changes requested):

tasks/tests.py
s0undt3ch: This is basically your script, but it's pushing latest data.
This invoke task is also meant to be temporary as I also agree that we should have zero speed skips, but right now, we have a few.
This work also comes from the fact that I had already rewritten some unittest2 tests to pytest, so, your data file would not match.
So, instead of building a mapping, and because the speed of unittest2 is not the same as pytest's, this script made sense.
It also allows us to test where we want to set the threshold without having to delete the blunt `@skipIf(True, ...)` and re-adding.

Again, this is being commited into the repo but by no means it alters the notion of the goal we wan to pursue, no speed related skips, ie, it's temporary, but I don't want it to get lost.

waynew: We should remove the Jenkins processing code from this PR and just rely on manually added skip flags. If you don't want it to get lost, we can keep the commits on the PR, but the end result shouldn't be doing anything "smart".

If the skip markers are there, we should skip, if they're not, we should run the test :+1:

Issue #41195, opened 2017-05-11
Can't use tpldir or slspath in template file


hatifnatt:
### Can't use tpldir or slspath in template file
I want to load defaults with construction like this
```
{%- import_yaml slspath ~ "/defaults.yaml" as defaults %}
```
Which work flawless in sls state file. But fails in template file. On #salt IRC channel I received a recommendation to use 'tpldir' variable instead, but it doesn't work as well. 

### Setup
init.sls
```
{%- import_yaml slspath ~ "/defaults.yaml" as defaults %}
config:
  file.managed:
    - name: "/tmp/test.cfg"
    - source: salt://{{ slspath }}/test.cfg.jinja
    - template: jinja
```

test.cfg.jinja
```
{%- import_yaml tpldir ~ "/defaults.yaml" as defaults %}
opt1 = var1
opt2 = var2
```

With setup like this I receive error:
```
          ID: config
    Function: file.managed
        Name: /tmp/test.cfg
      Result: False
     Comment: An exception occurred in this state: Traceback (most recent call last):
                File "/usr/lib/python2.7/dist-packages/salt/state.py", line 1746, in call
                  **cdata['kwargs'])
                File "/usr/lib/python2.7/dist-packages/salt/loader.py", line 1704, in wrapper
                  return f(*args, **kwargs)
                File "/usr/lib/python2.7/dist-packages/salt/states/file.py", line 1821, in managed
                  **kwargs
                File "/usr/lib/python2.7/dist-packages/salt/modules/file.py", line 4340, in check_managed_changes
                  **kwargs)
                File "/usr/lib/python2.7/dist-packages/salt/modules/file.py", line 3834, in get_managed
                  **kwargs)
                File "/usr/lib/python2.7/dist-packages/salt/utils/templates.py", line 178, in render_tmpl
                  output = render_str(tmplstr, context, tmplpath)
                File "/usr/lib/python2.7/dist-packages/salt/utils/templates.py", line 386, in render_jinja_tmpl
                  buf=tmplstr)
              SaltRenderError: Jinja variable 'tpldir' is undefined
     Started: 18:20:39.233032
    Duration: 52.652 ms
     Changes:
```
Expected result is 'tpldir' will contain path to current template file.

### Versions Report
```
Salt Version:
           Salt: 2016.11.4

Dependency Versions:
           cffi: 0.8.6
       cherrypy: Not Installed
       dateutil: 2.2
      docker-py: Not Installed
          gitdb: 0.5.4
[...]

PR #56918, opened 2020-04-27
Port #51814 to master

alexey-zhukovin:
### What does this PR do?
Master port #51814


bryceml:
re-run all full

myii:
Thanks for porting this @alexey-zhukovin.  Unfortunately, as @max-arnold has pointed out in https://github.com/saltstack/salt/issues/41195#issuecomment-577159561, there is a least one situation where `tplroot` won't evaluate in a useful manner:

> One potential workaround is to introduce the `tplroot` variable (see #51814, however if you place a formula into a deeply nested dir (e.g., `blah/blah/blah/blah/template-formula`), the `tplroot` variable will be just `blah` and all the import paths inside the formula will be broken.

From a purely formulas' perspective, the workaround being considered is to work back through the directory tree from `tpldir` to find `map.jinja` and use that directory as the `tplroot`.  That's not going to cut it here, though.  I'd love to hear any further suggestions because `tpldir` is too unwieldy, as mentioned in the original PR.

waynew:
@myii Thanks for checking this!

---

Tests *should* be passing now - apparently you can `'foo'.join(Path(...), Path(...))` on newer Python but not on our Ubuntu 16.04 :upside_down_face: :tada:

waynew:
re-run archlts

twangboy (dismissed):

waynew:
re-run centos7-py3-pycryptodome

myii:
@waynew I was just commenting in Slack about how `tplroot` in now available in Salt, so we could start using it directly -- then I remembered that this PR might have not been merged -- which it isn't!  Is there anything else to be done to get this in?

Ch3LL:
I resolved the merge conflict and added the 3006 label so we do not loose track of this fix.

twangboy:
Failing tests:
```
    def file_client(self):
        """
        Return a file client. Instantiates on first call.
        """
        # If there was no file_client passed to the class, create a cache_client
        # and use that. This avoids opening a new file_client every time this
        # class is instantiated
        if (
            self._file_client is None
            or not hasattr(self._file_client, "opts")
            or self._file_client.opts["file_roots"] != self.opts["file_roots"]
        ):
            attr = "_cached_pillar_client" if self.pillar_rend else "_cached_client"
            cached_client = getattr(self, attr, None)
            if (
                cached_client is None
                or not hasattr(cached_client, "opts")
>               or cached_client.opts["file_roots"] != self.opts["file_roots"]
            ):
E           KeyError: 'file_roots'
```

dwoz:

PR #56937, opened 2020-04-28
Port #51803 to master

alexey-zhukovin:
### What does this PR do?

Port #51803 to master


cmcmarrow (approved):

cmcmarrow (changes requested):
Imap and range is never used in this port. Is their another port coming that uses imap and range?

waynew:
This is not actually necessary. Instead, we need to remove salt/ext/ipaddress.py since that only existed for Python2 support, which has been gone for a while.

PR #56954, opened 2020-04-28
add a plist serializer

weswhet:
### What does this PR do?
Adds a plist serializer, allowing you to manage plist files with `file.serialize`.

### Previous Behavior
Before the only way to manage and manipulate plists, a common file type on macOS, was to use `file.manage` with a source file.

### New Behavior
you can now use a `formatter` of `plist` in a `file.serialize` state to manage xml and binary plist files. 

```
test_plist_serial:
  file.serialize:
    - name: /tmp/test_plist.plist
    - serializer_opts:
      - fmt: FMT_BINARY
    - formatter: plist
    - user: root
    - group: wheel
    - mode: 0644
    - makedirs: True
    - merge_if_exists: False
    - dataset:
        Label: com.saltstack.salt.master
        RunAtLoad: True
        KeepAlive: True
        ProgramArguments:
          - /opt/salt/bin/start-salt-master.sh
        SoftResourceLimits:
          NumberOfFiles: 100000
        HardResourceLimits:
          NumberOfFiles: 100000
        StandardOutPath: /tmp/salt-minion.out
        StandardErrorPath: /tmp/salt-minion.err
``` 

- [x] Docs
- [x] Changelog
- [x] Tests written/updated

### Commits signed with GPG?
Yes


weswhet:
cc @sheagcraig @garethgreenaway members of the macOS working group.

salt/states/file.py
```diff
@@ -7805,7 +7805,8 @@ def serialize(
         dataset, **serializer_options.get(serializer_name, {})
     )
 
-    contents += "\n"
+    if isinstance(contents, str):
```
weswhet, line 7808: the previous implementation would attempt to add a new line to the end of a binary plist. which would cause sadness.

salt/states/file.py
```diff
@@ -7773,7 +7773,7 @@ def serialize(
                     "result": False,
                 }
 
-            with salt.utils.files.fopen(name, "r") as fhr:
+            with salt.utils.files.fopen(name, "rb") as fhr:
```
weswhet, line 7776: trying to read a binary plist without the "b" would cause an exception. Also appears all the other implementations of  `salt.utils.files.fopen` in this file use `"rb"` as arguments.

salt/serializers/plist.py
```diff
@@ -0,0 +1,69 @@
+# -*- coding: utf-8 -*-
+"""
+    salt.serializers.plist
+    ~~~~~~~~~~~~~~~~~~~~~~~~~~~
+    .. versionadded:: TODO
```
weswhet, line 5: Would this be `3001`?

sheagcraig:
@wes and I wargamed this one and I think it looks ✨🤘🏻✨

salt/serializers/plist.py
waynew: Should be, assuming all is 🥇

salt/states/file.py
waynew: My guess is that was written using Py2, where binary == str == text. In Py3, binary == bytes and text == str. Definitely a good move here - unless we should have an 
```
else:
   contents += b"\n"
```

weswhet: @waynew Would it make sense to check to see if a new line already exists? I know `plistlib` automatically adds a newline.

waynew: I'm not familiar with other tools out there, but I would assume we should do whatever the "correct" thing is.

weswhet: I've updated with the recommendation as well as skipping new line concatenation on plists as this will create an invalid plist file.

salt/states/file.py
weswhet: I moved this back to its previous behavior and only add a "b" if the formatter is a plist. this change was causing the `integration.states.test_file.FileTest.test_serializer_deserializer_opts` test to fail.

salt/serializers/plist.py
weswhet: added

waynew (changes requested):
> The module at salt/serializers/plist.py does not have a sphinx stub at doc/ref/serializers/all/salt.serializers.plist.rst

Looks like you're missing some docs. I *think* you can autogen - just take a look at the other docs around there.

weswhet:
> > The module at salt/serializers/plist.py does not have a sphinx stub at doc/ref/serializers/all/salt.serializers.plist.rst
> 
> Looks like you're missing some docs. I _think_ you can autogen - just take a look at the other docs around there.

Should be added. I think I did that right :)

weswhet:
wooo, thanks all!

PR #56988, opened 2020-04-29
Return pip binary without -m flag when using onedir

xeacott:
### What does this PR do?
When using the onedir inside dist/run from pop-build, pip is provided and therefore cannot handle the -m flag. The pip module needs to handle this case.

Fixes:
https://github.com/saltstack/release-planning/issues/239

### Previous Behavior

### New Behavior

- [ ] Docs
- [ ] Changelog
- [x] Tests written/updated

### Commits signed with GPG?
Yes/No


s0undt3ch (commented):
Should we keep functionality as it was and provide an alternate logic if the app is frozen and/or has the `sys._MEIPASS` attribute?

xeacott:
@s0undt3ch I'm open to taking this in a different direction. The heart of the issue is that the executable that sys.executable is reporting is *not* python, just something else; in the case of the onedir it's `./run`. So the `-m` that gets passed around isn't going to a python binary but rather the pip binary.

s0undt3ch:
Sure, and we should handle that. Draw your own conclusions and solution _when_ running under pyinstaller.
If it's not running under pyinstaller, then the logic should remain as it is today.

Makes sense?

s0undt3ch:
```
if under_pyinstaller:
     ...
else:
    ...
```

salt/modules/pip.py
```diff
@@ -144,14 +144,30 @@ def _clear_context(bin_env=None):
...
 
+def _check_bundled():
+    """
+    Gather run-time information to indicate if we are running from source or bundled.
+    """
+    frozen = False
+    if getattr(sys, "frozen", False) and hasattr(sys, "_MEIPASS"):
+        frozen = True
+    return frozen
+
```
s0undt3ch, line 155:
```
ef _check_bundled():
    """
    Gather run-time information to indicate if we are running from source or bundled.
    """
    if getattr(sys, "frozen", False) and hasattr(sys, "_MEIPASS"):
        return True
    return False
```
One less local variable

tests/unit/modules/test_pip.py
```diff
@@ -38,6 +38,11 @@ def test__pip_bin_env_no_bin_env(self):
         ret = pip._pip_bin_env(None, None)
         self.assertIsNone(ret)
 
+    def test__pip_bin_bundled_app(self):
+        sys_info = hasattr(sys, "sys._MEIPASS")
```
s0undt3ch, line 42: `sys_info = hasattr(sys, "_MEIPASS")`?

s0undt3ch (approved):

PR #57002, opened 2020-04-30
Fix spelling mistakes

bdrung:
Fix some more / new spelling mistakes found by lintian.


bdrung:
This typo fix does not need a test case IMO.

dwoz (approved):

PR #57039, opened 2020-05-01
Port #50617 to master

alexey-zhukovin:
### What does this PR do?

Port #50617 to master


mchugh19:
As this is an RFC, the `Needs Testcase` label isn't needed.

Since this was not submitted, by @isbm, is the intent to continue this work? While I'd be in full support of developing capabilities, if there isn't intent to complete the RFC then there probably isn't much point in discussing it.

isbm:
I have no idea what "needs testcase" means here.

dwoz:
The should be a SEP if it is still needed.

dwoz:

Issue #23501, opened 2015-05-08
aptpkg does not support passing --force-confnew


anitakrueger:
In multiple places in aptpkg.py the option of --force-confold is set when installing packages with apt:
https://github.com/saltstack/salt/blob/develop/salt/modules/aptpkg.py#L507
https://github.com/saltstack/salt/blob/develop/salt/modules/aptpkg.py#L536
https://github.com/saltstack/salt/blob/develop/salt/modules/aptpkg.py#L738
https://github.com/saltstack/salt/blob/develop/salt/modules/aptpkg.py#L741

We have a requirement where we would like to always stomp local changes when installing a package, but I can't find a way to pass --force-confnew instead of --force-confold to pkg.install or the state pkg.

Is this possible with the current aptpkg.py and if not, can it be incorporated?

Thanks a lot
- Anita

jfindlay:
@anitakrueger, this does not seem to be currently possible.  Thanks for the report.

rallytime:
@anitakrueger Are you in a position to give the pull request above a try? Does it resolve this issue for you?

anitakrueger:
Hey...sorry for the late response. I'll try to test this tonight and let you know

acatton:
It looks like this has been fixed by #23684, no?

jfindlay:
@acatton, yes.  That change will be in 2015.8.0.

rallytime:
I am going to close this since it looks to me like this is fixed. If more work should be done here, please let us know and we'd be happy to re-open this and take another look. Thanks!

harmv:
This should be reopened, the fix is not working correctly. See comments in #23684.

Issue #40558, opened 2017-04-06
aptpkg does not correctly pass  force_conf_new


harmv:
### Description of Issue/Question

Ticket #23501 is not correctly fixed and should be reopened.
I filed a new ticket as I could not reopen that ticket. 

I think the merged fix from  9d0bb55a135a52d900f25ec  is not correct.

If  `force_conf_new: True` is given,  salt adds _both_ options: 
 `'-o', 'DPkg::Options::=--force-confnew', '-o', 'DPkg::Options::=--force-confdef'`

As net effect apt-get will then _not_ use force-confnew, but force-confdef.

The following patch fixed the problem for me
```diff
diff --git a/salt/modules/aptpkg.py b/salt/modules/aptpkg.py
index 3a96d63..87ca989 100644
--- a/salt/modules/aptpkg.py
+++ b/salt/modules/aptpkg.py
@@ -635,7 +635,7 @@ def install(name=None,
             cmd_prefix.extend(['-o', 'DPkg::Options::=--force-confnew'])
         else:
             cmd_prefix.extend(['-o', 'DPkg::Options::=--force-confold'])
-        cmd_prefix += ['-o', 'DPkg::Options::=--force-confdef']
+            cmd_prefix += ['-o', 'DPkg::Options::=--force-confdef']
         if 'install_recommends' in kwargs:
             if not kwargs['install_recommends']:
                 cmd_prefix.append('--no-install-recommends')
```
That patch is not complete though..
The following code is wrong too  salt/modules/aptpkg.py
```python
 cmd.extend(['apt-get', '-q', '-y',
                '-o', 'DPkg::Options::={0}'.format(force_conf),
                '-o', 'DPkg::Options::=--force-confdef'])
```

The patch above is just to demonstrate the problem, a better fix is probably needed.

### Setup
(Please provide relevant configs and/or SLS files (Be sure to remove sensitive info).)

### Steps to Reproduce Issue
use force_conf_new, and see it fail

### Versions Report
2016.11

Ch3LL:
I had to look up these options and the man pages [here](http://manpages.ubuntu.com/manpages/trusty/man1/dpkg.1.html) was useful. It states:

>  confnew: If a conffile has been modified and the version in  the
              package  did  change,  always  install  the  new version without
              prompting, unless the  --force-confdef  is  also  specified,  in
              which case the default action is preferred. 

[...]

PR #57051, opened 2020-05-02
fix force-confnew/force-confdef options in aptpkg

terrible-broom:
### What does this PR do?
Fix of force-confnew/force-confdef/force-confold dpkg options in aptpkg module.

Fixes:
https://github.com/saltstack/salt/issues/23501
https://github.com/saltstack/salt/issues/40558

### Previous Behavior
Options force-confnew and force-confdef were used together.

### New Behavior
Option force-confnew is used on its own, and force-confold and force-confdef are used together.

### Commits signed with GPG?
No


terrible-broom:
Honestly have no idea why code coverage changed. Please give a hint how to fix.

cmcmarrow (changes requested):
Thank you for your fix, but can we get some tests?

terrible-broom:
> Thank you for your fix, but can we get some tests?

Okay, i've made very basic tests. I hope this is the right way to do it.

dwoz (approved):

PR #57070, opened 2020-05-04
Port #50389 to master

alexey-zhukovin:
### What does this PR do?

Port #50389 to master

### DIFF
--- a/salt/utils/rsax931.py
+++ b/salt/utils/rsax931.py
@@ -39,6 +39,9 @@ def _load_libcrypto():
         )
     else:
         lib = find_library("crypto")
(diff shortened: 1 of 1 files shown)


waynew (changes requested):
needs tests/help wanted

waynew:
It looks like the API for this changed, I don't know if this is still necessary - if it is we should re-open this, but it should be pretty straightforward.

dmurphy18:
Submitted a patch to Python to allow for Solaris 11.4 and how CRLE is detected, last I saw, still waiting for patch to be accepted.

Problem was with how CRLE was detected, the output string changed with Solaris 11 and the fix resulted in how crypto was found.

PR #57119, opened 2020-05-07
Make "LazyLoader.__init__" thread-safe when calling "_refresh_file_mapping"

meaksh:
### What does this PR do?
This PR fixes spurious segfault we've been experiencing when heavily using the "ssh" client via "salt-api".

After analyzing generated "coredumps", the segfault is always coming from the same place:

```
#0  0x00007f4551e88bd4 in visit_decref (op=<unknown at remote 0x7f4490040e58>, data=data@entry=0x0) at Modules/gcmodule.c:374
#1  0x00007f4551ddd3eb in gen_traverse (gen=0x7f451c7ece60, visit=0x7f4551e88bd0 <visit_decref>, arg=0x0) at Objects/genobject.c:21
#2  0x00007f4551e88273 in subtract_refs (containers=<optimized out>) at Modules/gcmodule.c:399
#3  collect (generation=1, n_collected=0x7f445a7f2ff8, n_uncollectable=0x7f445a7f3000, nofail=0) at Modules/gcmodule.c:953
#4  0x00007f4551ed97ed in collect_with_callback (generation=1) at Modules/gcmodule.c:1128
#5  0x00007f4551e87fcb in _PyObject_GC_Alloc (basicsize=<optimized out>, use_calloc=0) at Modules/gcmodule.c:1729
#6  _PyObject_GC_Malloc (basicsize=<optimized out>) at Modules/gcmodule.c:1739
#7  0x00007f4551e8801e in _PyObject_GC_NewVar (tp=tp@entry=0x7f45521bae00 <PyTuple_Type>, nitems=nitems@entry=3) at Modules/gcmodule.c:1768
#8  0x00007f4551e0e05f in PyTuple_New (size=size@entry=3) at Objects/tupleobject.c:114
#9  0x00007f4551e50f52 in _PyEval_EvalFrameDefault (f=<optimized out>, throwflag=<optimized out>) at Python/ceval.c:2558
#10 0x00007f4551e582f5 in _PyEval_EvalCodeWithName (qualname='LazyLoader._refresh_file_mapping', name=<optimized out>, closure=0x0, kwdefs=0x0, defcount=<optimized out>, 
    defs=0x0, kwstep=1, kwcount=<optimized out>, kwargs=<optimized out>, kwnames=0x0, argcount=<optimized out>, args=<optimized out>, locals=0x0, globals=<optimized out>, 
    _co=<code at remote 0x7f45476e4660>) at Python/ceval.c:4166
#11 fast_function (func=<optimized out>, stack=<optimized out>, nargs=<optimized out>, kwnames=<optimized out>) at Python/ceval.c:4992
#12 0x00007f4551e578d6 in call_function (pp_stack=0x7f445a7f3368, oparg=<optimized out>, kwnames=0x0) at Python/ceval.c:4872
#13 0x00007f4551e5089a in _PyEval_EvalFrameDefault (f=<optimized out>, throwflag=<optimized out>) at Python/ceval.c:3335
#14 0x00007f4551e5981a in _PyEval_EvalCodeWithName (qualname='LazyLoader.__init__', name=<optimized out>, closure=(<cell at remote 0x7f45472226d8>,), kwdefs=0x0, 
    defcount=<optimized out>, defs=0x7f45476e91f8, kwstep=2, kwcount=<optimized out>, kwargs=<optimized out>, kwnames=<optimized out>, argcount=<optimized out>, 
    args=0x7f453c1105f8, locals=0x0, globals=<optimized out>, _co=<code at remote 0x7f45476e4150>) at Python/ceval.c:4166
```

So, basically the "segfault" is always triggered by the Python GC (garbage collector) when running on the context of `LazyLoader.__init__` at the time of executing `LazyLoader._refresh_file_mapping` (frame 10).

This function is apparently not thread-safe, in consequence, there is already a RLock mechanism around to ensure thread-safety when this function is called.

It seems the RLock mechanism is not triggered on this particular "path" of code, and that might produce the "segfault" in some cases.

This PR makes sure the RLock is used on this particular path of the code.

- [ ] Docs
- [ ] Changelog
- [ ] Tests written/updated

### Commits signed with GPG?
Yes


dwoz:
@meaksh Is it possible to write a test case to cover this? I'm basically looking for something that will cause a segfault if this lock gets reverted in the future.

meaksh:
@dwoz it's actually not easy to reproduce the `segfault` in a simple way, but I've added a unit test to ensure the "lock" is used when calling `LazyLoader._refresh_file_mapping` from `LazyLoader.__init__`.

dwoz (approved):

PR #57122, opened 2020-05-07
msgpack: support versions >= 1.0.0

aplanas:
### What does this PR do?

A recent change in msgpack >= 1.0.0, update the default value for the
parameter `raw` to False.  This change breaks Salt for those versions.

This patch add the parameter `raw=True` to all the unpack operations,
restoring the old default.

### DIFF
--- a/salt/utils/msgpack.py
+++ b/salt/utils/msgpack.py
@@ -75,13 +75,29 @@ def _sanitize_msgpack_kwargs(kwargs):
     return kwargs
 
 
+def _sanitize_msgpack_unpack_kwargs(kwargs):
+    """
+    Clean up msgpack keyword arguments for unpack operations, based on
+    the version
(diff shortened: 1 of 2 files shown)


aplanas:
Added tests for both helpers

dmurphy18:
Fixed memory leak using msgpack 1.0.0 experiencing on Junos, thanks timely PR

dmurphy18 (approved):

dmurphy18 (approved):

dwoz:
#56007

prometheanfire:
this missed removing the cap from `requirements/base.txt`

prometheanfire:
from what I can see this is low hanging fruit for sodium

aplanas:
@prometheanfire right: https://github.com/saltstack/salt/pull/57539

PR #57123, opened 2020-05-07
Prevent logging deadlock on "salt-api" subprocesses dealing with SSH minions

meaksh:
### What does this PR do?

This PR fixes a deadlock that sometimes happens under heavy load (lot of parallel requests to Salt API) performing operations with SSH minions. This situation gets even worse when "DEBUG" logging is enabled on the master.

Let say we have an environment with 5 ssh minions, (we use a "roster file"), and we execute multiple `curl` requests to "salt-api" `/run` endpoint, i.a. running: `cmd.run date`, in parallel.

The "salt-api" starts processing the requests, most of the requests eventually finish but for some of them, the `salt-api` subprocesses created for handling those requests get into a deadlock during the very beginning of the process' bootstrapping, at Python's `logging._acquireLock()`. Before the "ssh" call is even executed.

Those processes are stuck forever.

This PR fixes the deadlock by doing some refactors on how Salt handles the Python's logging locks in order to manage them properly and avoid the deadlock.

Sorry for not attaching any dump or debugging information at this moment.

(I would need to think how this could be to tested)

- [ ] Docs
- [ ] Changelog
- [ ] Tests written/updated

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/_logging/impl.py
+++ b/salt/_logging/impl.py
@@ -245,7 +245,7 @@ class SaltLoggingClass(
             finally:
                 try:
                     handler.release()
-                except:
+                except Exception:  # pylint: disable=broad-except
                     pass
         return instance
 
@@ -323,7 +323,7 @@ class SaltLoggingClass(
                     stack_info=stack_info,
                     stacklevel=stacklevel,
                 )
-        except:
+        except Exception:  # pylint: disable=broad-except
             pass
         finally:
             logging._releaseLock()
--- a/salt/_logging/impl.py
+++ b/salt/_logging/impl.py
@@ -1,4 +1,3 @@
-# -*- coding: utf-8 -*-
 """
     salt._logging.impl
     ~~~~~~~~~~~~~~~~~~
@@ -6,16 +5,11 @@
     Salt's logging implementation classes/functionality
 """
 
-# Import python libs
-from __future__ import absolute_import, print_function, unicode_literals
 
 import logging
 import re
 import sys
 import types
-
-# Import 3rd-party libs
-import salt.ext.six as six
 
 # Let's define these custom logging levels before importing the salt._logging.mixins
 # since they will be used there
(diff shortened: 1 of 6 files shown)


salt/_logging/impl.py
```diff
@@ -419,6 +426,7 @@ def makeRecord(
     logging.addLevelName(PROFILE, "PROFILE")
     logging.addLevelName(TRACE, "TRACE")
     logging.addLevelName(GARBAGE, "GARBAGE")
+    logging.addLevelName(DEBUG, "DEBUG")
```
s0undt3ch, line 429: The DEBUG level isn't present?!
`addLevelName` doesn't complain if there's already a level name by the same name?

salt/client/ssh/wrapper/cp.py
```diff
@@ -5,11 +5,11 @@
 # Import Python libs
 from __future__ import absolute_import, print_function
 
-import logging
 import os
 
 # Import salt libs
 import salt.client.ssh
+import salt.log.setup as logging
```
s0undt3ch, line 12: Hmm, why are we doing this?

salt/loader.py
```diff
@@ -26,6 +25,7 @@
 import salt.config
 import salt.defaults.events
 import salt.defaults.exitcodes
+import salt.log.setup as logging
```
s0undt3ch, line 28: Same as above

waynew (changes requested):
Definitely need some tests, though... not sure how one would test this?

salt/client/ssh/wrapper/cp.py
meaksh: My understanding is that, if default Python's logging is used here, it would contribute to the deadlock when this code is executed on the context of bootstrapping the new sub-processes for handling the SSH work.

salt/loader.py
meaksh: Same answer as above :)

salt/_logging/impl.py
meaksh: Hmm, we've been running this patch for a while and it doesn't complain, but I'll double check if this is strictly necessary on the context of the deadlock.

meaksh:
> Definitely need some tests, though... not sure how one would test this?

Yeah, I'm wondering the same right now :smile: 
I'll try to figure that out!

dwoz:

PR #57149, opened 2020-05-08
Fix pycrypto.gen_hash and update the test.

### LINKED ISSUE #57268: [Test Failure]  integration.shell.test_key.KeyTest.test_remove_key_eauth
integration.shell.test_key.KeyTest.test_remove_key_eauth

Fixed in https://github.com/saltstack/salt/pull/57149

### LINKED ISSUE #57269: [Test Failure]  integration.shell.test_key.KeyTest.test_list_acc_eauth
integration.shell.test_key.KeyTest.test_list_acc_eauth

Fixed in https://github.com/saltstack/salt/pull/57149

DmitryKuzmenko:
### What does this PR do?
Make `salt.utils.pycrypto.gen_hash` work again `algorithm=None` setting it to the strongest one of available if possible.
Update the corresponding test to actually run the `gen_hash` test, to pass it and to check this particular case.
This is needed for #50544 to pass the key tests.

Fixes salt_key tests for https://github.com/saltstack/salt/issues/50544
Fixes https://github.com/saltstack/salt/issues/57269
Fixes https://github.com/saltstack/salt/issues/57268

### Previous Behavior
Before https://github.com/saltstack/salt/pull/54931 the `gen_hash` worked without the `algorithm` argument passing using the strongest available algorithm.
After #54931 it stopped to work raising an error.

### New Behavior
Works again.

- [ ] Docs
- [x] Tests written/updated

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/utils/pycrypto.py
+++ b/salt/utils/pycrypto.py
@@ -103,6 +103,7 @@ def gen_hash(crypt_salt=None, password=None, algorithm=None, force=False):
     """
     Generate /etc/shadow hash
     """
+    log.debug("gen_hash start with: %s", (crypt_salt, password, algorithm, force))
     if password is None:
         password = secure_password()
 
@@ -112,8 +113,10 @@ def gen_hash(crypt_salt=None, password=None, algorithm=None, force=False):
 
     if algorithm not in methods:
         if force and HAS_PASSLIB:
+            log.debug("gen_hash return fallback")
             return _fallback_gen_hash(crypt_salt, password, algorithm)
         else:
+            log.debug("gen_hash raise SaltInvocationError")
             raise SaltInvocationError(
                 "Algorithm '{}' is not natively supported by this platform, use force=True with passlib installed to override.".format(
                     algorithm
@@ -126,8 +129,10 @@ def gen_hash(crypt_salt=None, password=None, algorithm=None, force=False):
         crypt_salt = "${}${}".format(methods[algorithm].ident, crypt_salt)
     else:  # method is crypt (DES)
         if len(crypt_salt) != 2:
+            log.debug("gen_hash raise ValueError")
             raise ValueError(
                 "Invalid salt for hash, 'crypt' salt must be 2 characters."
             )
 
+    log.debug("gen_hash returning crypt for: %s, %s", password, crypt_salt)
     return crypt.crypt(password, crypt_salt)
--- a/salt/utils/pycrypto.py
+++ b/salt/utils/pycrypto.py
@@ -106,6 +106,10 @@ def gen_hash(crypt_salt=None, password=None, algorithm=None, force=False):
     if password is None:
         password = secure_password()
 
+    if algorithm is None:
+        # use the most secure natively supported method
+        algorithm = crypt.methods[0].name.lower() if HAS_CRYPT else known_methods[0]
+
     if algorithm not in methods:
         if force and HAS_PASSLIB:
             return _fallback_gen_hash(crypt_salt, password, algorithm)
@@ -115,10 +119,6 @@ def gen_hash(crypt_salt=None, password=None, algorithm=None, force=False):
                     algorithm
                 )
             )
-
-    if algorithm is None:
-        # use the most secure natively supported method
-        algorithm = crypt.methods[0].name.lower()
 
     if crypt_salt is None:
         crypt_salt = methods[algorithm]
--- a/salt/utils/pycrypto.py
+++ b/salt/utils/pycrypto.py
@@ -87,9 +87,6 @@ def _fallback_gen_hash(crypt_salt=None, password=None, algorithm=None):
     """
     Generate a /etc/shadow-compatible hash for a non-local system
     """
-    if algorithm is None:
-        algorithm = 0
-
     # these are the passlib equivalents to the 'known_methods' defined in crypt
     schemes = ["sha512_crypt", "sha256_crypt", "bcrypt", "md5_crypt", "des_crypt"]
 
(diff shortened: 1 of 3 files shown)


salt/utils/pycrypto.py
```diff
@@ -103,31 +103,36 @@ def gen_hash(crypt_salt=None, password=None, algorithm=None, force=False):
     """
     Generate /etc/shadow hash
     """
+    log.debug("gen_hash start with: %s", (crypt_salt, password, algorithm, force))
     if password is None:
         password = secure_password()
 
+    if algorithm is None:
+        # use the most secure natively supported method
+        algorithm = crypt.methods[0].name.lower() if HAS_CRYPT else known_methods[0]
```
dhiltonp, line 112: We should either:

1.  keep the `known_methods[0]` here, and delete `if algorithm is None` in `_fallback_gen_hash`

or

2. `if algorithm is None and HAS_CRYPT`, and modify `if algorithm is None` in `_fallback_gen_hash` to set algorithm to `known_methods[0]`.

tests/unit/utils/test_pycrypto.py
```diff
@@ -22,10 +22,7 @@ class PycryptoTestCase(TestCase):
     TestCase for salt.utils.pycrypto module
     """
 
-    @skipIf(
-        not salt.utils.pycrypto.HAS_CRYPT or not salt.utils.pycrypto.HAS_PASSLIB,
-        "crypt not available",
-    )
+    @skipIf(not salt.utils.pycrypto.HAS_CRYPT, "crypt not available")
```
dhiltonp, line 25: keeping `or not HAS_PASSLIB` allows these tests to run if passlib is installed and the platform is windows.

I suggest we keep the condition as is, install passlib by default for tests, and modify the `force` below (line 56) to only be active if HAS_CRYPT is False (so the HAS_PASSLIB code is exercised on windows).

dhiltonp: (Right now the passlib code is *only* tested if passlib is manually installed, and these test changes break passlib).

tests/unit/utils/test_pycrypto.py
```diff
@@ -86,18 +83,24 @@ def test_gen_hash(self):
...
 
-            with self.assertRaises(ValueError, msg=algorithm):
-                ret = salt.utils.pycrypto.gen_hash(
-                    crypt_salt=invalid_salt,
-                    password=passwd,
-                    algorithm=algorithm,
-                    force=force,
-                )
-                self.assertNotEqual(ret, expected["hashed"])
+        with self.assertRaises(ValueError):
```
dhiltonp, line 86: I don't understand this test change...

dhiltonp (changes requested):

dhiltonp:
Also, thanks for pinging me on this. I should have thought of including passlib as a test dependency and executing it on windows earlier.

tests/unit/utils/test_pycrypto.py
DmitryKuzmenko: Hm... Skip if not crypt or not passlib makes the test running only if both installed.
Did you mean `@skipIf(not HAS_CRYPT and not HAS_PASSLIB)` to run it ever if CRYPT is not found but we have passlib?

salt/utils/pycrypto.py
```diff
@@ -103,31 +103,36 @@ def gen_hash(crypt_salt=None, password=None, algorithm=None, force=False):
     """
     Generate /etc/shadow hash
     """
+    log.debug("gen_hash start with: %s", (crypt_salt, password, algorithm, force))
```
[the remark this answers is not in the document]
DmitryKuzmenko: It's a temporary commit to check what's happening with those 3 failures. I've planned to remove this.

salt/utils/pycrypto.py
DmitryKuzmenko: I'll re-check, thank you.

tests/unit/utils/test_pycrypto.py
DmitryKuzmenko: `gen_hash` raises nothing unless it's `blowfish` algorithm and `salt` is longer that `2` chars. I've checked the `crypt` docs, there's nothing about raising `ValueError` too. So I don't know where did you get this. Maybe on windows?

DmitryKuzmenko: Oh yes. Got it, it's from passlib. I'll update the test accordingly.

DmitryKuzmenko:
@dhiltonp I've reviewed the code and the test. Wanted to unify the `gen_hash` interface but crypt and passlib interfaces are significantly different. So I've just decided to split the test to check passlib, crypt and absence of any in different tests. Locally all passed. Could you please review?

dhiltonp:
I feel kind of lame saying this, because I'm the person who wrote it this way.

Maybe we should modify `gen_hash` so that it doesn't call passlib if `force=True`, and instead `gen_hash` would raise if an incompatible method is requested, something like `raise HashError("Hash method '{}' unavailable, install hashlib and use gen_hash_passlib instead")`. 

nxos, which currently uses `force=True`, could do something like:

```
try:
    gen_hash()
except HashError:
    if not salt.utils.pycrypto.PASSLIB:
        raise
    gen_hash_passlib()
```

s0undt3ch:
Fixes https://github.com/saltstack/salt/issues/57269
Fixes https://github.com/saltstack/salt/issues/57268

dwoz (approved):

Issue #55835, opened 2020-01-10
Epic: Support for Python 3.8


sagetherage:
### Description of Issue
Epic or bucket ticket for tasks for supporting Python 3.8 changes going forward, as an epic this ticket may span more than one release over time.

### Related ticket(s):
- https://github.com/saltstack/salt/issues/55310
- https://github.com/saltstack/salt/issues/55847
- https://github.com/saltstack/salt/issues/56998
- #55976
- #56027
- https://github.com/saltstack/salt/pull/56031

Ch3LL:
- [ ] https://github.com/saltstack/salt/issues/55976

Deprecation Warnings that need to be cleaned up

bdrung:
#55976 is a duplicate of #50911, which is addressed in pull request #56027

bdrung:
I pushed more commits to pull request #56031. Now it fixes all Python 3.8 unit test failures. So after backporting these commits, the Debian package 2019.2.3+dfsg1-2 will build on Ubuntu 20.04 (which contains Python 3.8).

s0undt3ch:
Related https://github.com/saltstack/salt/pull/55900

s0undt3ch:
For anyone working on Py3.8 support, some work has been done in https://github.com/saltstack/salt/tree/features/py38-compat?files=1

baby-gnu:
I think this should be excluded from stalebot.

twangboy:
- [ ] https://github.com/saltstack/salt/issues/56998

Issue #55310, opened 2019-11-14
Python 3.8 Support


s0undt3ch:
### Description of Issue
Support Python 3.8

Arch Linux updated to python 3.8 and one of the tests upgrades the system, which means, while the test suite runs, we upgrade from 3.7 to 3.8, thus failing at the very end.

https://jenkinsci.saltstack.com/job/pr-kitchen-archlts-py3/job/PR-55276/4/

Current stopgap: https://github.com/saltstack/salt/pull/55309

s0undt3ch:
Problematic code:
- [ ] [`from platform import _supported_dists`](https://github.com/saltstack/salt/blob/master/salt/grains/core.py#L40)

Ch3LL:
- [ ] Need to have a running list of deprecation's which will affect Salt, for example: abc from collections in Py 3.8.  Plan to have public list in GitHub ? where contributors / Salt can add to the list and generate Issues such that Salt can be updated.

Ikke:
For Alpine, the lack of python3.8 support prevents salt-minion from running, so this is not just a test failure, but actual run-time errors:

```
$ salt-minion
Traceback (most recent call last):
[..]
  File "/usr/lib/python3.8/site-packages/salt/grains/core.py", line 40, in <module>
    from platform import _supported_dists
ImportError: cannot import name '_supported_dists' from 'platform' (/usr/lib/python3.8/platform.py)

itamarjp:
Python 3.8 will land fedora soon. (already there for Fedora 32)
https://koji.fedoraproject.org/koji/packageinfo?packageID=9781

Vaelatern:
This crops up after working around problems in issue  #56551

chalbersma:
I think this will block support for Ubuntu 20.04 as it uses python3.8 by default.

Ch3LL:
can you try this PR https://github.com/saltstack/salt/pull/56031?

brianthelion:
@Ch3LL The `platform._supported_dists` issue had not be fixed as of yesterday. See #56752

terminalmage:
It's not just `platform._supported_dists` that is gone in Python 3.8, it is also `platform.linux_distribution()`. This code was not being maintained and now no longer exists.

The core grains already support using [`distro`](https://pypi.org/project/distro/), which is the recommended replacement for `platform.linux_distribution()`.

[...]

PR #57151, opened 2020-05-08
Distro is now a hard dependency

s0undt3ch:
### What does this PR do?
See title

Refs #55310
Refs #55835
Closes https://github.com/saltstack/salt/pull/55410

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -684,7 +684,8 @@ allowed-3rd-party-modules=msgpack,
                           pytest,
                           setuptools,
                           pytestsalt,
-                          saltfactories
+                          saltfactories,
+                          distro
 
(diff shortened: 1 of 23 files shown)


garethgreenaway (approved):

dmurphy18 (approved):

agraul:
Could you share the reason for making `distro` a hard dependency for all supported Python versions, even those who support `from platform import linux_distribution` (e.g. Python 3.6)? Just curios, I did not expect the old way to go away so soon.

s0undt3ch:
Removed in Py3.8 and since we will have to move to `distro`, we just stop maintaining two different ways of getting the data.

PR #57190, opened 2020-05-10
Fix #57189

### LINKED ISSUE #57189: [BUG] file.managed state’s does not handle extents attribute correctly
**Description**

The `attrs` option of `file.managed` attempts to change file attributes, even if they match except for the `e` (extents) attribute.

**Setup**

```
/tmp/immutable-file:
   file.managed:
     - contents: ''
     - attrs: 'i'
```

Whether this problem can be reproduced depends on details of the file system used on the minion. When the file system stores the file in a way that it uses extents, the following behavior will be visible:

When apply the state for the first time, everything looks fine. When subsequently applying the state in test mode, the output will look like this:

```
          ID: /tmp/immutable-file
    Function: file.managed
        Name: /tmp/immutable-file
      Result: None
     Comment: The file /tmp/immutable-file is set to be changed
              Note: No changes made, actual changes may
              be different due to other states.
     Started: 16:49:48.657780
    Duration: 29.804 ms
     Changes:
              ----------
              attrs:
                  i
```

When actually applying the change again, the state will fail because the immutable attribute is already set and thus `chattr` cannot be used (except for removing the immutable attribute).

```
          ID: /tmp/immutable-file
    Function: file.managed
        Name: /tmp/immutable-file
      Result: False
     Comment: Failed to change attributes to i
     Started: 16:31:04.937077
    Duration: 113.853 ms
     Changes:
```

**Expected behavior**

The `file.managed` state should ignore the `e` (extents) attribute if it is set but not specified as part of the `attrs` option.

From the manpage of `chattr (1)`:

       The 'e' attribute indicates that the file is using extents for  mapping
       the blocks on disk.  It may not be removed using chattr(1).

**Versions Report**
<details><summary>salt --versions-report</summary>

```
Salt Version:
           Salt: 3000.2

Dependency Versions:
           cffi: Not Installed
       cherrypy: Not Installed
       dateutil: 2.6.1
      docker-py: Not Installed
          gitdb: 2.0.3
      gitpython: 2.1.8
         Jinja2: 2.10
        libgit2: Not Installed
       M2Crypto: 0.31.0
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.5.6
   mysql-python: Not Installed
      pycparser: Not Installed
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: Not Installed
         Python: 3.6.9 (default, Apr 18 2020, 01:56:04)
   python-gnupg: 0.4.1
         PyYAML: 3.12
          PyZMQ: 16.0.2
          smmap: 2.0.3
        timelib: Not Installed
        Tornado: 4.5.3
            ZMQ: 4.2.5

System Versions:
           dist: Ubuntu 18.04 bionic
         locale: UTF-8
        machine: x86_64
        release: 4.15.0-99-generic
         system: Linux
        version: Ubuntu 18.04 bionic
```
</details>

smarsching:
### What does this PR do?

This patch ensures that the extents file attribute is ignored when comparing file attributes. This is necessary because this attribute is set by the underlying file system and can never be reset.

Fixes: #57189

### Previous Behavior

The `file.managed` state would call `chattr` if there was a mismatch in the file attributes because the file had the `e` (extents) attribute set, but this attribute wasn’t specified in the `attrs` option of the `file.managed` state.

### New Behavior

The `file.managed` state will not consider the `e` attribute being present for the actual file, but not as part of the `attrs` option a difference. Thus, no changes will be reported in test mode and `chattr` will not be called when applying the state.

**Please note:** I didn’t add a test because the bug being visible depends on behavior of the underyling file system (the extents attribute being present) and so there is no reliable way to test for it.

- [ ] Docs
- [x] Tests written/updated

### Commits signed with GPG?
No


waynew:
@smarsching thanks for the PR! I feel like I've done something with this before...

waynew:
Ah, I guess it was just https://github.com/saltstack/salt/pull/53188 that I did :thinking: 

In order to merge this PR we'll definitely needs some tests, as well as [a changelog entry](https://docs.saltstack.com/en/master/topics/development/changelog.html).

smarsching:
@waynew Well, as I wrote above, writing a test is tricky, because we cannot really know whether the `e` flag will be present in the test environment or not. But if you have an idea how to test this, please let me know.

waynew:
@smarsching the easiest way would be to patch `lsattr`. It *would* be possible to do a bunch of setup and checks on the underlying system... but at this point, I trust that if lsattr says 'e' is there, then 'e' is really there :slightly_smiling_face:

smarsching:
@waynew Maybe you already saw it: I have added a test case and all the CI tasks seem to pass, so I guess this might now be ready for merging.

waynew (dismissed):

Ch3LL (dismissed):

Ch3LL:
looks like there is a pre-commit failure that needs to be cleaned up

smarsching:
I don’t think I understand this failure. The console output has a very long diff, but it seems to be about lines that I didn’t even touch…

Maybe someone with a better understanding of the build pipeline could have a look?

Ch3LL:
I went ahead and added the changes. Pre-commit is being used right now for any files you change to help with the conversion from python2 > python3 code.  

Here is how to get pre-commit setup for your next contribution: https://docs.saltstack.com/en/latest/topics/development/contributing.html#pre-commit

thanks!

PR #52756, opened 2019-04-30
Try to address dict changing during iteration

s0undt3ch:
### What does this PR do?
```
06:10:57        RuntimeError: dictionary changed size during iteration
06:10:57        Traceback (most recent call last):
06:10:57          File "/usr/lib64/python2.7/multiprocessing/queues.py", line 266, in _feed
06:10:57            send(obj)
06:10:57        RuntimeError: dictionary changed size during iteration
```

PR #57192, opened 2020-05-11
Port #52756 to master

terminalmage:
See #52756

### DIFF
--- a/changelog/57192.changed
+++ b/changelog/57192.changed
@@ -1,0 +1,1 @@
+Try to address dict changing during iteration
--- a/salt/log/setup.py
+++ b/salt/log/setup.py
@@ -83,6 +83,10 @@ class SaltLogQueueHandler(QueueHandler):
     Subclassed just to differentiate when debugging
     """
 
+    def prepare(self, record):
+        record = QueueHandler.prepare(self, record)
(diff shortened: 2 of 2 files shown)


s0undt3ch (dismissed):

waynew:
Looked at this for a few minutes during the Testing WG 2021-06-03 - at least with the `prepare` function it looks like maybe we had solved that problem in a different way(?) Probably need to take a closer look. Also found a block of code in `salt/_logging/handlers` that should just be flat out removed (only used on Python <3.2, which is no longer supported). That should probably be part of another effort(?)

Ch3LL:
@terminalmage would you mind adding a changelog

Also 
@waynew its not clear to me if your statement is blocking the merge of this PR or not?
And @s0undt3ch you still okay with this change since its been a while since your previous approval?

terminalmage:
@Ch3LL done

s0undt3ch:
We no longer use queue handlers.

s0undt3ch:
And logging going forwards is in `salt/_logging` the old log stuff is in deprecation.

Ch3LL:
thanks @s0undt3ch you okay closing this then @terminalmage ?

terminalmage:
Of course!

PR #57197, opened 2020-05-11
Shielding port variable from type errors

brejoc:
## What does this PR do?

Casting `port` variable to `int` to prevent error when port is passed as string.

### Previous Behavior

When passing `port` as `string`, you'd get a `TypeError`.

### New Behavior

`port` can be passed as `string` and `int`.

- [ ] Docs
- [x] Tests written/updated

### Commits signed with GPG?
Yes


brejoc:
Added some tests. Waiting for the outcome.

Issue #57088, opened 2020-05-05
A safe solution after the CVE vulnerability


rongzeng54:
**Is your feature request related to a problem? Please describe.**
A safe solution after the CVE vulnerability

**Describe the solution you'd like**
In some case, salt-master cannot been complete defensed by iptables (maybe someone think iptables may also not enough, or the attack may come from a client that you trusted), at the same time the long-term automation work Built on SaltStack.

One possible solution is as follows:
**server**: Publish port is deny, return port is allowed, then the server become a server for distributing anything except command, such as state or pillar (you also need to ensure that these distribution files are trusted, but the master is no longer a God node, because it can no longer control all servers by push mode)

**client**: Send the correct salt-minion configuration and key pair by a more safely way, but make sure that your salt-minion is not started

**publish**: Now we need salt-call work, you can do it by some tools based on ssh(pull mode, so you can publish from a credible network) or other private tools based on P2P (when the network is not good, the success rate of command delivery can be improved)

**Describe alternatives you've considered**
- Iptables
- Black and white lists in server-side and client-side

**Additional context**
salt-call can run without salt-minion is running

**Please Note**
If this feature request would be considered a substantial change or addition, this should go through a SEP process here https://github.com/saltstack/salt-enhancement-proposals, instead of a feature request.

cruscio:
Is this issue a request for proposals?

How about:

[1] hardening the `TCP Transport` so that it can be configured to complete an mTLS handshake on connection to the master? (the configuration options are there today, but I don't believe they work)

I have significant misgivings about the security of the TCP transport based on [this line](https://github.com/saltstack/salt/blob/f52a2f24902ca25d90e12029f84f0635dba4d51f/salt/transport/tcp.py#L957) in transport/tcp.py:
```
# Always connect in plaintext; we'll convert to ssl if necessary
```

[2] An HTTPS WebSocket transport, with a clear separation of the creation of the WebSocket from the rest of the transport's communication. I know functionally this isn't much different from today's TCP transport, but in my experience "HTTPS WebSocket" sells a lot better to security teams. It's more familiar.

[3] An External Transport (BYO Message Bus): Configure both the master & minions to connect to a tertiary Kafka/Pulsar/RabbitMQ/Other pub/sub system. Plugable publishing, just like plugable returners.

Skeen:
[...]

PR #57210, opened 2020-05-11
WIP: HTTP Transport

Skeen:
### What does this PR do?
* Introduces `salt/transport/abstract,py`, which provides abstract implementations of the channels by refactoring out sharable behavior from the `tcp` transport (`salt/transport/tcp.py`).
This is still work-in-progress, and I'd like to refactor it to a mixin-based architecture, where the channel flows through mixins / layers / decorators towards the transport itself, such that the transport layer itself should merely provide an `rx` and a `tx` function for each direction of the data-flow.
* Introduces `salt/transport/http.py`, which utilizes the abstract implementations from above, to implement a `http` transport using a message queue and long-polling.

The arguments for the HTTP channel versus TCP or ZeroMQ is:
* It is possible to run multiple masters behind a single secure HTTPS load-balancer via hostname routing.
* It allows for reuse of the entire HTTP(S) ecosystem of tools, such as for throtttling, logging, tracing, etc.
* The protocol is widespread, and generally firewalls will accept outgoing HTTPS traffic, while random TCP trafic might be blocked.

Currently outstanding tasks:
* Configuration of the `http` channel, such as `master_uri`, potentially with a new format.
* Further refactoring of `salt/transport/abstract,py`.
* Minion connection / reconnection handling events within the `http` channel.
* Tests

Relates to: #57088 

- [ ] Docs
- [ ] Tests written/updated

### Commits signed with GPG?
Yes/No

### DIFF
--- a/salt/transport/http.py
+++ b/salt/transport/http.py
@@ -1,0 +1,363 @@
+# -*- coding: utf-8 -*-
+"""
+HTTP transport classes
+"""
+
+# Import Python Libs
+from __future__ import absolute_import, print_function, unicode_literals
+
+import json
+import base64
+import logging
+import socket
+import uuid
+
+# Import Tornado Libs
+import salt.ext.tornado
+import salt.ext.tornado.web
+import salt.ext.tornado.locks
+import salt.ext.tornado.gen
+import salt.ext.tornado.httpclient
+import salt.ext.tornado.httpserver
+import salt.transport.frame
+import salt.transport.abstract
+import salt.utils.msgpack
+from salt.transport.tcp import _set_tcp_keepalive
+from salt.ext import six
+
+log = logging.getLogger(__name__)
+
+if six.PY2:
+    import urllib
+    urlencode = urllib.urlencode
+else:
+    import urllib.parse
+    urlencode = urllib.parse.urlencode
+
+# pylint: disable=import-error,no-name-in-module
+if six.PY2:
+    import urlparse
+else:
+    import urllib.parse as urlparse
+# pylint: enable=import-error,no-name-in-module
+
+
+class AsyncHTTPReqChannel(salt.transport.abstract.AbstractAsyncReqChannel):
+    def __new__(cls, opts, **kwargs):
+        return super().__new__(cls, opts, **kwargs)
+
+    @classmethod
+    def __key(cls, opts, **kwargs):
+        return super().__key(cls, opts, **kwargs)
+
+    def start_channel(self, io_loop, **kwargs):
+        parse = urlparse.urlparse(self.opts["master_uri"])
+        master_host, master_port = parse.netloc.rsplit(":", 1)
+        self.url = "http://" + master_host + ":" + str(master_port) + "/req"
+        self.http_client = salt.ext.tornado.httpclient.AsyncHTTPClient(io_loop=self.io_loop)
+
+    def close(self):
+        super().close()
+        if hasattr(self, "http_client"):
+            self.http_client.close()
+
+    def publish_string(self, spayload):
+        post_data = {"payload": spayload}
+        body = urlencode(post_data)
+        http_request = salt.ext.tornado.httpclient.HTTPRequest(self.url, method="POST", headers=None, body=body)
+
+        return_future = salt.ext.tornado.gen.Future()
+        def callback(response):
+            if response.error:
+                # TODO: Deal with errors
+                print("!!", __class__, "Error:", response.error)
+            else:
+                spayload = response.body
+                b64payload = spayload.decode("ascii")
+                bpayload = base64.b64decode(b64payload)
+                unpacker = salt.utils.msgpack.Unpacker()
+                unpacker.feed(bpayload)
+                for framed_msg in unpacker:
+                    if six.PY3:
+                        framed_msg = salt.transport.frame.decode_embedded_strs(
+                            framed_msg
+                        )
+                    header = framed_msg['head']
+                    payload = framed_msg['body']
+                    return_future.set_result(payload)
+                return
+            return_future.set_result(None)
+        self.http_client.fetch(http_request, callback)
+
+        return return_future
+
+
+class MessageRequestHandler(salt.ext.tornado.web.RequestHandler):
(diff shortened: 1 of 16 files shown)


Skeen:
Should an RFC be made before this has any chance of being merged?

lukasraska:
Great initiative, my 2 cents here (note that I haven't tested the implementation, just took a look at the source):
- long polling seems like generally a bad idea, there might be multiple connections at the same time, there can be time period between one request ends and second is established - resulting in missed messages, websockets would be better here in my option
- the general implementation very closely mimics what ZMQ transport does - publishes events to all connected minions, exactly what TCP transport avoids (see `publish_payload`) - while this does make sense for ZMQ as it's designed for that, HTTP transport could handle that better
- If I'm not mistaken, the payload of the HTTP message is URL-encoded JSON with base64-encoded msgpack data as `payload` field - not that it would necessarily be a bad idea (that way it reuses a LOT of existing code), but I don't understand the reasoning behind the implementation itself - the content cannot be easily validated / checked (for example by IDS/IPS systems)... essentially it's just TCP transport wrapped in HTTP session without the benefits of TCP transport itself - using something like JWT would be much better, but also much harder to implement

Overall I think it's a good idea to implement HTTP transport, but having it just a wrapper around existing logic doesn't help, it makes it more prone to errors. Having SEP for this would be probably good way to start, so Salt Core team can look into the proposal (I'm not affiliated with SaltStack in any way).

Skeen:
HI, thanks you for your feedback!

> long polling seems like generally a bad idea, there might be multiple connections at the same time, there can be time period between one request ends and second is established - resulting in missed messages, ...

There shouldn't be missed messages, as the implementation uses a [`MessageBuffer`](https://github.com/saltstack/salt/pull/57210/files#diff-65f7ddbebf866e56c1903120ba9b530eR274-R298) to store messages as they come in, and as [clients request the server, with the id of the last message they've consumed](https://github.com/saltstack/salt/pull/57210/files#diff-65f7ddbebf866e56c1903120ba9b530eR250-R252), thus getting all messages that occured since their last connection (if any).
Because of this I do not think messages will be missed, unless the client checks in so infrequently, that the entire message buffer has been flushed through since their last connection. - Making the size of the message buffer configurable should solve this problem though.

> ... websockets would be better here in my option

I agree that websockets will be a good transport too, but I do not believe that one should exclude the other. As I see it, a polling HTTP transport can coexist with a websocket transport.

> the general implementation very closely mimics what ZMQ transport does - publishes events to all connected minions, exactly what TCP transport avoids (see `publish_payload`) - while this does make sense for ZMQ as it's designed for that, HTTP transport could handle that better

Agreed, and it is indeed very related to one of my outstanding tasks, namely 'minion connection / reconnection handling events within the `http` channel.'. The TCP tasks sends id's over to keep track of minions, while this implementation currently does not.

>If I'm not mistaken, the payload of the HTTP message is URL-encoded JSON with base64-encoded msgpack data as `payload` field - not that it would necessarily be a bad idea (that way it reuses a LOT of existing code), but I don't understand the reasoning behind the implementation itself - the content cannot be easily validated / checked (for example by IDS/IPS systems)... essentially it's just TCP transport wrapped in HTTP session without the benefits of TCP transport itself - using something like JWT would be much better, but also much harder to implement

You are not mistaken, what I went for was exactly the low-hanging fruit of just reusing as much existing code as possible. The reasoning behing this, was to get a fully functional implementation as soon as possible, as to test out routing and tracing using standard HTTP software.

The reason I created this PR early (and as a work-in-progress), is exactly to have these discussions and get this input, and it seems that here we've hit upon an interesting discussion indeed. Namely whether the transports should just do exactly what it says on the tin: Transport the data across the wire, or whether they should be specialized for their specific flavor, strengths and weaknesses.

I can see compelling arguments for both sides. Implementing the HTTP transport using JWT would utliize the strengths of HTTP better, however it would share less code and thus be harder to maintain, and would move a lot of the channel protocol into the transport instead of in an intermediate shared layer, as I suggest with a layer based architecture, where data flows through the layers towards the transport itself.

> Overall I think it's a good idea to implement HTTP transport, but having it just a wrapper around existing logic doesn't help, it makes it more prone to errors.

This is interesting, as i would argue for the opposite, the more code which is shared, the more well tested that code will be, and thus the less prone to errors, therefore minimizing the transport specific code and simultaneously minimizing the amount of transport specific bugs.

> Having SEP for this would be probably good way to start, so Salt Core team can look into the proposal (I'm not affiliated with SaltStack in any way).

Yea I thought it might be a good idea too.

salt/transport/http.py
```diff
@@ -0,0 +1,363 @@
...
+    def __new__(cls, opts, **kwargs):
+        return super().__new__(cls, opts, **kwargs)
+
+    @classmethod
+    def __key(cls, opts, **kwargs):
+        return super().__key(cls, opts, **kwargs)
+
+    def start_channel(self, io_loop, **kwargs):
+        parse = urlparse.urlparse(self.opts["master_uri"])
+        master_host, master_port = parse.netloc.rsplit(":", 1)
```
frebib, line 55: This doesn't work for IPv6 addresses. Use `parse.hostname` and `parse.port` instead

salt/transport/http.py
```diff
@@ -0,0 +1,363 @@
...
+                self.http_server.shutdown()
+            except Exception as exc:  # pylint: disable=broad-except
+                log.exception(
+                    "HTTPReqServerChannel close generated an exception: %s", str(exc)
+                )
+        elif hasattr(self.http_server, "stop"):
+            self.http_server.stop()
+
+    def _start_socket(self):
+        sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
```
frebib, line 159: No IPv6 love here

waynew:
Is the goal here simply to add a new/abstract transport process, or are we explicitly concerned with security?

If it's the latter we should (MUST?) add tests for well-known attacks. Granted, that might be a /large/ suite of tests, but assuming they can be run in <24h that wouldn't be onerous to run periodically, and we would add it as a prerequisite for every release.

PR #57237, opened 2020-05-13
Allow set higher priority failhard on the client-side

rongzeng54:
### What does this PR do?
Allow set higher priority failhard on the client-side.
Failhard is a cool feature, because in many case fast fail is better.
Global failhard control is hard to use, client-side can control itself seems better.

Fixes: No

### Previous Behavior
Server-side can set global failhard.

### New Behavior
Client-side can set higher priority failhard.

- [ ] Docs
- [ ] Tests written/updated

### Commits signed with GPG?
No

### DIFF
--- a/salt/state.py
+++ b/salt/state.py
@@ -3145,13 +3145,15 @@ class BaseHighState(object):
         if not isinstance(mopts, dict):
             # An error happened on the master
             opts['renderer'] = 'jinja|yaml'
-            opts['failhard'] = False
+            if 'failhard' not in opts:
+                opts['failhard'] = False
             opts['state_top'] = salt.utils.url.create('top.sls')
             opts['nodegroups'] = {}
             opts['file_roots'] = {'base': [syspaths.BASE_FILE_ROOTS_DIR]}
         else:
             opts['renderer'] = mopts['renderer']
-            opts['failhard'] = mopts.get('failhard', False)
+            if 'failhard' not in opts:
+                opts['failhard'] = mopts.get('failhard', False)
             if mopts['state_top'].startswith('salt://'):
                 opts['state_top'] = mopts['state_top']
             elif mopts['state_top'].startswith('/'):


salt/state.py
```diff
@@ -3145,13 +3145,15 @@ def __gen_opts(self, opts):
         if not isinstance(mopts, dict):
             # An error happened on the master
             opts['renderer'] = 'jinja|yaml'
-            opts['failhard'] = False
+            if 'failhard' not in opts:
+                opts['failhard'] = False
```
waynew, line 3149:
```suggestion
          opts.setdefault('failhard', False)
```
This will only set failhard if it's not already set.

salt/state.py
```diff
@@ -3145,13 +3145,15 @@ def __gen_opts(self, opts):
...
             opts['state_top'] = salt.utils.url.create('top.sls')
             opts['nodegroups'] = {}
             opts['file_roots'] = {'base': [syspaths.BASE_FILE_ROOTS_DIR]}
         else:
             opts['renderer'] = mopts['renderer']
-            opts['failhard'] = mopts.get('failhard', False)
+            if 'failhard' not in opts:
+                opts['failhard'] = mopts.get('failhard', False)
```
waynew, line 3156:
```suggestion
            opts.setdefault('failhard', mopts.get('failhard', False))
```
Same - will only set failhard if not already set.

waynew (changes requested):
@rongzeng54 thanks for the PR! We require tests to merge PRs, would you be able to write some tests for this? It should be relatively straightforward.

Also please add a [changelog entry](https://docs.saltstack.com/en/master/topics/development/changelog.html)

If you're able to get this PR updated before Monday we will be able to get it merged in for the Sodium release (3001).

Thanks again!

dwoz:
@rongzeng54 This should not have been opened against `develop`. Please point this to `master` instead.

sagetherage:
@rongzeng54 This should not have been opened against develop. Please point this to master instead.

sagetherage:

PR #57239, opened 2020-05-13
Adjust the test suite to have a separate pytest style tests directory

s0undt3ch:
### What does this PR do?
See title


cmcmarrow (dismissed):

tests/nts/README.md
```diff
@@ -0,0 +1,14 @@
+# New Test Suite
+
+Welcome to the *new* test suite for Salt!
+
+Any test under this directory shall be written exploring the full capabilities of PyTest.
+That means, no occurrences of the [TestCase](https://docs.python.org/3/library/unittest.html#unittest.TestCase) class
+shall be used, neither our [customizations to it](../support/case.py).
```
waynew, line 7: Would it be possible/simple to enforce this via a test? I'd be 👍 for that.

s0undt3ch: You mean force new tests to be on the PyTest only directory?
Once we fully switch we could make a pre-commit hook to enforce that.

waynew: no - you say that `TestCase` and `support.case` shouldn't be used in the new directory. Can we easily enforce *that*?

s0undt3ch:
@dwoz I assume the label was automation?
Because yes, there's a failure under PyTest and it doesn't need to be fixed on this PR.

tests/nts/README.md
s0undt3ch: Ah! Yes we can!

s0undt3ch: Done!

dwoz (approved):

Issue #43079, opened 2017-08-21
How to obtain more details of error message?


guettli:
### Description of Issue/Question
I run `salt-ssh` and receive this error message.

```
          ID: necessary_packages
    Function: pkg.installed
      Result: False
     Comment: Problem encountered installing package(s). Additional info follows:
              
              errors:
                  - Running scope as unit run-r3338b54eff4140afad9df7dc73c05c93.scope.
                    No apport report written because the error message indicates its a followup error from a previous failure.
                    E: Sub-process /usr/bin/dpkg returned an error code (1)
     Started: 09:15:32.101015
    Duration: 5534.169 ms
     Changes:   


```

How to obtain more details? I think some heavily needed stdout/stderr gets lost.

### Versions Report

```
salt:/srv # salt --versions-report
Salt Version:
           Salt: 2016.11.3
 
Dependency Versions:
           cffi: 1.1.0
       cherrypy: 3.6.0
       dateutil: 2.1
          gitdb: 0.6.4
      gitpython: 1.0.1
          ioflo: Not Installed
         Jinja2: 2.9.5
        libgit2: 0.22.0
        libnacl: Not Installed
       M2Crypto: 0.22
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.4.8
   mysql-python: Not Installed
      pycparser: 2.12
       pycrypto: 2.6.1
         pygit2: 0.22.0
         Python: 2.7.3 (default, Apr 14 2012, 08:58:41) [GCC]
   python-gnupg: Not Installed
         PyYAML: 3.12
          PyZMQ: 16.0.2
           RAET: Not Installed
          smmap: 0.9.0
        timelib: Not Installed
        Tornado: 4.4.2
            ZMQ: 4.1.6
 
System Versions:
           dist: SuSE 12.3 x86_64
        machine: x86_64
        release: 3.7.10-1.45-default
         system: Linux
        version: openSUSE  12.3 x86_64

```

garethgreenaway:
@guettli Thanks for the report. It looks like there is an issue with the packaging system on the client side, possibly a failed attempt that left apt or dpkg in a failed state.

guettli:
@garethgreenaway  thank you for looking at this. You labeled this issue "info needed". What can I do now?

guettli:
[...]

PR #57252, opened 2020-05-13
Print stdout on apt install failure

piotr1212:
When apt install fails it is not clear from the stderr output what went wrong, ex:

 ```
  errors:
     - Running scope as unit: run-r5851ebada9ac4ae0a1ef7288de7c4b69.scope
       E: Unable to correct problems, you have held broken packages.
```
After adding the stdout it shows why it failed:

```
  errors:
      - Reading package lists...
        Building dependency tree...
        Reading state information...
        Some packages could not be installed. This may mean that you have
        requested an impossible situation or if you are using the unstable
        distribution that some required packages have not yet been created
        or been moved out of Incoming.
        The following information may help to resolve the situation:

        The following packages have unmet dependencies:
         bla:i386 : Depends: libsomethingorrather:i386 (>= 1.2.13) but it is not installable
                    Depends: anotherDependency:i386 (>= 1.2.6) but it is not installable
      - Running scope as unit: run-r0097166ab0f9412c879647df0955982f.scope
        E: Unable to correct problems, you have held broken packages.
```

### What does this PR do?
Print stdout when apt install fails.

Fixes:

### Previous Behavior
Only print stderr

### New Behavior
print stderr and stdout

- [ ] Docs
- [ ] Tests written/updated

### Commits signed with GPG?
Yes


piotr1212:
https://github.com/saltstack/salt/issues/43079

piotr1212:
Wondering if this would be useful for all failures of apt, not only install or that this should be configurable with a loglevel.

Ch3LL (changes requested):
Can you add some test coverage. Also the changelog file only needs the PR number and I would not label this as a new feature. Can you rename the file to 57252.fixed, since its fixing an issue where not all output is included in the error?

dwoz:

PR #57307, opened 2020-05-15
Fix proxy shell tests

DmitryKuzmenko:
Fixes these tests failures:
```
integration.proxy.test_shell.ProxyCallerSimpleTestCase.test_can_it_ping
integration.proxy.test_shell.ProxyCallerSimpleTestCase.test_grains_items
integration.proxy.test_shell.ProxyCallerSimpleTestCase.test_list_pkgs
integration.proxy.test_shell.ProxyCallerSimpleTestCase.test_service_get_all
integration.proxy.test_shell.ProxyCallerSimpleTestCase.test_service_list
integration.proxy.test_shell.ProxyCallerSimpleTestCase.test_service_start
integration.proxy.test_shell.ProxyCallerSimpleTestCase.test_upgrade
```
https://jenkinsci.saltstack.com/job/pr-centos7-py3-proxy-slow/job/master/13/

Create 'pki' directory for proxy minion.
Fixed path to proxy config dir.

Still fail on pytest. I'm checking this.

PR #57310, opened 2020-05-15
Additional fix for proxytests in pytest

DmitryKuzmenko:
### What does this PR do?
This fixes proxy tests running with pytest in addition to #57307
This is not needed for Sodium and makes no sense for runtests or salt logic. So no need to merge right now.

This prevents of running `salt_factories._get_root_dir_for_daemon(proxy_minion_id)` twice for the same daemon. Running this twice produces 2 different roots: one is set to `RUNTIME_VARS` the second one is used by daemon that breaks tests that use `RUNTIME_VARS`.


s0undt3ch (approved):
Yes, this is the right fix. Thanks!

Issue #57318, opened 2020-05-15
[BUG] junos.conn seems to go stale


jtraub91:
**Description**
Sometimes, proxy minions utilizing `junos.cli` will return without any newlines. This makes the output hard to read. 

**Setup**
Configure and start a junos proxy minion. The bug _might_ be due to long open connections, so wait, if necessary.

**Steps to Reproduce the behavior**
Execute a cli command.

```
$ salt <minion-id> junos.cli "show ntp associations no-resolve"
<minion-id>:
    ----------
    message:
        remote refid st t when poll reach delay offset jitter =============================================================================== +xxx.xxx.xxx.xxx xxx.xxx.xxx.xxx x - xx xx xx xx xx xx *xxx.xxx.xxx.xxx  xxx.xxx.xxx.xxx x - xx xx xx xx xx xx
    out:
        True

```

**Expected behavior**
Command should output with newlines. Ex:
```
<minion-id>:
    ----------
    message:

           remote         refid           st t when poll reach   delay   offset  jitter
        ===============================================================================
        +xxx.xxx.xxx.xxx      xxx.xxx.xxx.xxx      x -  xx xx  xx    xx    xx   xx
        *xxx.xxx.xxx.xxx      xxx.xxx.xxx.xxx      x -  xx xx  xx    xx    xx   xx
    out:
        True

```

**Versions Report**
```
$ salt --versions-report
Salt Version:
           Salt: 3000.3

Dependency Versions:
           cffi: 1.13.2
       cherrypy: unknown
       dateutil: 2.8.1
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
         Jinja2: 2.10.3
        libgit2: 0.28.2
       M2Crypto: Not Installed
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.5.6
   mysql-python: Not Installed
      pycparser: 2.19
       pycrypto: 2.6.1
   pycryptodome: 3.9.7
         pygit2: 1.0.2
         Python: 3.7.5 (default, Oct 17 2019, 12:25:15)
   python-gnupg: 0.4.5
         PyYAML: 5.3.1
          PyZMQ: 19.0.1
          smmap: Not Installed
        timelib: Not Installed
        Tornado: 4.5.3
            ZMQ: 4.3.2

System Versions:
           dist:
         locale: UTF-8
        machine: x86_64
        release: 3.10.0-1062.el7.x86_64
         system: Linux
[...]

PR #57328, opened 2020-05-15
saltstack#57318 add conn.open to junos.cli

jtraub91:
### What does this PR do?
Adds `conn.open` call to `junos.cli` to refresh the stale ssh connection

https://github.com/saltstack/salt/issues/57318

### Previous Behavior
Stale connections sometimes do not display newlines as it should.

### New Behavior
Connection is effectively refreshed and output is as expected.


vnitinv:
This change is not required. Stale connection should be fixed by keep-alive feature.

jtraub91:
@vnitinv Then there seems to be a bug in the keep alive feature. The default config should achieve this but it does not. Can we please take a closer look here https://github.com/saltstack/salt/issues/57318 ?

vnitinv:
@jtraub91 Not getting o/p in multiple lines has nothing to do with connectivity. Must be something to do with Normalize.

jtraub91:
@vnitinv Do you happen to know why calling conn.open() again resolves the problem? I could refactor `proxy.junos.alive` to call this, but it sounds like you're suggesting a more elegant solution?

vnitinv:
@jtraub91 Can you share with me the state file which was executed before encountering this issue?

jtraub91:
@vnitinv No state file. The details are documented in https://github.com/saltstack/salt/issues/57318. The command I ran was 
```
$ salt <minion-id> junos.cli "show ntp associations no-resolve"
```

dmurphy18 (commented):
While this appears to fix the issue, from the comments it appears that this resolves the symptom but is not the cure.  Better to fix the underlying reason, also fix would need a test-case, which in providing would high-light were the correct fix needs to be made.

dmurphy18:
@jtraub91 Any movement on this issue, there have been updates to junos support since then and if they have resolved your issue since this is issue should really be why keep_alive not working rather than fixing the symptom.

Note have a fix to make to salt/modules/junos.py which is under test at the moment https://github.com/saltstack/salt/issues/59020

dmurphy18 (changes requested):
@jtraub91 Are you still interested in this PR ?, if unresponsive, this PR will be closed after several days

jtraub91:
@dmurphy18 I agree that this probably belongs in the `keep_alive` functionality if the problem persists. I'm OK with closing it.

dmurphy18:
@jtraub91 Going to close this, as the Juniper native minion is going to be opened to the public, going through the VMware processes at the moment, so cannot give a timeline but hopefully soon, already approved to do so.
Also since 27 failing checks and PR over 2 years old.

PR #57399, opened 2020-05-20
consistent file location handling across APIs

### LINKED ISSUE #146: have consistency in path evaluation across APIs
### Description of Issue
currently, some API looks for the file on master and some on proxy server itself. bring the consistency

vnitinv:
### What does this PR do?
Fix https://github.com/Juniper/salt/issues/146

Fixes:
  * all API should have the same behavior for the user-provided file.
  * In case of proxy model, we can have file located either on master or proxy server itself. So we should handle both cases.

### Previous Behavior
* install_os, load, install_config took file from master, file_copy get_table used to take file from proxy. 
* Also if master/proxy running at same machine, install_os would unnecessarily copy huge file to same server to some tmp file. 

### New Behavior
* if master/proxy running at same allocation, file is not copied and can use local file location
* All APIs now can take file from master using salt://

- [ ] Docs
- [ y ] Tests written/updated

### Commits signed with GPG?
Yes/No

### DIFF
--- a/salt/modules/junos.py
+++ b/salt/modules/junos.py
@@ -33,6 +33,7 @@ import salt.utils.stringutils
 import yaml
 from salt.exceptions import MinionError
 from salt.ext import six
+import tempfile
 
 try:
     from lxml import etree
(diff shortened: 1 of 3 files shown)


vnitinv:
@dmurphy18 Would you help in reviewing this?

salt/modules/junos.py
```diff
@@ -72,6 +71,81 @@
...
+            )
+            if self._kwargs:
+                self._cached_file = __salt__["cp.get_template"](
+                    self._file_path, self._cached_folder, **self._kwargs
+                )
+            else:
+                self._cached_file = __salt__["cp.get_file"](
+                    self._file_path, self._cached_folder
+                )
+            if self._cached_file != "":
```
dmurphy18, line 124: Wonder what other values self._cached_file might have other than "" and None, or a better way to know if it is usable.

salt/modules/junos.py
```diff
@@ -72,6 +71,81 @@
 __proxyenabled__ = ["junos"]
 
 
+class HandleFileCopy:
```
dmurphy18, line 74: could this class be placed after the __virtual__, since that is usually kept to the top of a file

salt/modules/junos.py
vnitinv: get_file and get_template return full path on proxy server (if copied from master is a susccess) else ''

salt/modules/junos.py
vnitinv: will do this change

vnitinv: Fixed via https://github.com/saltstack/salt/pull/57399/commits/f308a577738f4e70d27927594cf3c26bab79b4b8

dmurphy18 (approved):

PR #57415, opened 2020-05-21
use hashlib instead of shelling out to generate a server_id

dhiltonp:
### What does this PR do?
Streamlines get_server_id, generating the hash inside the current process.

### Commits signed with GPG?
Yes/No


cmcmarrow (dismissed):

dhiltonp:
This change will alter `server_id`. I chose sha256, while the previous implementation used SipHash24.

dwoz (approved):

Issue #44878, opened 2017-12-07
nodeps doesn't work


jdshewey:
#33145 added the ability to specify nodeps when installing rpms which is documented [here](https://docs.saltstack.com/en/latest/ref/states/all/salt.states.pkg.html). Unfortunately this does not seem to work in 2017.7.2:

    [root@slik01.example.com example]$ rpm -qa | egrep "^salt-"
    salt-repo-latest-2.el7.noarch
    salt-master-2017.7.2-1.el7.noarch
    salt-api-2017.7.2-1.el7.noarch
    salt-minion-2017.7.2-1.el7.noarch
    salt-2017.7.2-1.el7.noarch

If I configure an example state:

    example_install:
      pkg.installed:
        - sources:
          - examle: salt://example/files/example-1.2.3-1.x86_64.rpm
        - pkg_verify:
          - verify_options: nodeps

    example_install:
      pkg.installed:
        - sources:
          - examle: salt://example/files/example-1.2.3-1.x86_64.rpm
        - pkg_verify:
          - verify_options: 
            - nodeps

This results in the error:

          ID: example_install
    Function: pkg.installed
      Result: False
     Comment: Error occurred installing package(s). Additional info follows:

              errors:
                  - Running scope as unit run-67034.scope.
                    Loaded plugins: fastestmirror
                    Examining /var/cache/salt/minion/files/example/files/example-1.2.3-1.x86_64.rpm: example-1.2.3-1.x86_64
                    Marking /var/cache/salt/minion/files/example/files/example-1.2.3-1.x86_64.rpm to be installed
                    Resolving Dependencies
                    --> Running transaction check
                    ---> Package example.x86_64 0:1.2.3 will be installed
                    --> Processing Dependency: example2 >= 3.0.1 for package:  example-1.2.3-1.x86_64
                    Loading mirror speeds from cached hostfile
                    --> Processing Dependency: libstdc++.so.5(CXXABI_1.2)(64bit) for package:  example-1.2.3-1.x86_64
                    --> Processing Dependency: libstdc++.so.5()(64bit) for package:  example-1.2.3-1.x86_64
                    --> Running transaction check
                    ---> Package example.x86_64 0:1.2.3 will be installed
                    --> Processing Dependency: example2 >= 3.0.1 for package: example-1.2.3-1.x86_64
                    ---> Package compat-libstdc++-33.x86_64 0:3.2.3-72.el7 will be installed
                    --> Finished Dependency Resolution
[...]

PR #57426, opened 2020-05-22
Improve pkg.installed pkg_verify documentation

### LINKED ISSUE #62848: [DOCS] pkg.installed pkg_verify docs could be a bit clearer
**Description**
`pkg.installed`'s `pkg_verify` is sometimes misunderstood, e.g. in https://github.com/saltstack/salt/issues/44878

**Suggested Fix**
See https://github.com/saltstack/salt/pull/57426

**Type of documentation**
See https://github.com/saltstack/salt/pull/57426

agraul:
### What does this PR do?
Improves the documentation of the `pkg_verify` parameter in `pkg.installed`.

Refers to https://github.com/saltstack/salt/issues/44878
Fixes https://github.com/saltstack/salt/issues/62848

- [x] Docs
- [X] ~~Tests written/updated~~ docstring changes only

For the changelog I am not sure if I should use this PR's number or something else. I could also open an issue for the bugfix that I then reference, not sure how to go about that.

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/states/pkg.py
+++ b/salt/states/pkg.py
@@ -1279,14 +1279,15 @@ def installed(
 
         .. versionadded:: 2014.7.0
 
-        For requested packages that are already installed and would not be
-        targeted for upgrade or downgrade, use pkg.verify to determine if any
-        of the files installed by the package have been altered. If files have
-        been altered, the reinstall option of pkg.install is used to force a
-        reinstall. Types to ignore can be passed to pkg.verify. Additionally,
-        ``verify_options`` can be used to modify further the behavior of
-        pkg.verify. See examples below.  Currently, this option is supported
-        for the following pkg providers: :mod:`yum <salt.modules.yumpkg>`.
+        Use pkg.verify to check if already installed packages require
+        reinstallion. Requested packages that are already installed and not
+        targeted for up- or downgrade are verified with pkg.verify to determine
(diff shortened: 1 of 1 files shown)


agraul:
@dwoz I have a few questions on how to proceed here, I hope you can help me. 
1. What type of test is more suitable for this type of bug?
2. For the changelog, should I open a new bug to reference or is it okay to use the PR number from this PR
3. Is it expected that the arguments passed to a state module function via `salt-call * state.single` is a `dict`, while the arguments are an `OrderedDict` when the same state module function is called inside a SLS? If not the core issue is somewhere else (although it does not hurt to have the `isinstance` check against the more abstract class)

agraul:
I would like to move this forward, but I'm not clear on how to proceed. Do you have any ideas about he questions from [above](https://github.com/saltstack/salt/pull/57426#issuecomment-649548198) @dwoz ?

meaksh:
@Ch3LL ^^

agraul:
I've removed the code changes that required testing. With 3006, Python 3.5 is not supported anymore and the custom `OrderedDict` can be dropped from Salt. But that is best done in a separate PR.

barbaricyawps (approved):
These changes look great to me. Thanks for your PR, @agraul .

PR #57428, opened 2020-05-22
Late import `salt.client.ssh.client`

s0undt3ch:
### What does this PR do?

The reason behind this late import is that Salt's thin generation, if
Python 2 is available, tries to get the required dependencies, as such,
it tries to import `concurrent` provided by the `futures` package.
Having that import at the module scope will try to import `concurrent`,
if not available, the master will fail to start, even if the user is not
working with salt-ssh at all.

See also:
* https://github.com/saltstack/salt-bootstrap/pull/1466/checks?check_run_id=700200552
* https://github.com/saltstack/salt-bootstrap/pull/1466/checks?check_run_id=700200628

### DIFF
--- a/salt/master.py
+++ b/salt/master.py
@@ -22,7 +22,6 @@ import time
 import salt.acl
 import salt.auth
 import salt.client
-import salt.client.ssh.client
 
 # Import salt libs
 import salt.crypt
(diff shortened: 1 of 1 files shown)


sagetherage:
@saltstack/team-ssh anyone willing to take stab at testing this?

cmcmarrow:
Only one test fail for windows `tests.integration.modules.test_chocolatey.ChocolateyModuleTest.test_list_` it has a `__context__` error.

s0undt3ch:
Yes, and that one should be fixed in master now.
Let me pull latest changes.

PR #57437, opened 2020-05-23
Fix `warn_until` calls for the Sodium release

s0undt3ch:
### What does this PR do?
See title

### DIFF
--- a/salt/modules/keystone.py
+++ b/salt/modules/keystone.py
@@ -175,7 +175,7 @@ def auth(profile=None, **connection_args):
         salt '*' keystone.auth
     """
     __utils__["versions.warn_until"](
-        "3001",
+        "Sodium",
         (
             "The keystone module has been deprecated and will be removed in {version}.  "
             "Please update to using the keystoneng module"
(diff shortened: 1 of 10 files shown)


krionbsd (dismissed):

salt/modules/keystone.py
```diff
@@ -175,7 +175,7 @@ def auth(profile=None, **connection_args):
         salt '*' keystone.auth
     """
     __utils__["versions.warn_until"](
-        "3001",
+        "Sodium",
```
waynew, line 178:
```suggestion
        "Magnesium",
```

waynew (changes requested):
If they're supposed to be removed in Sodium, we'll need to remove those, too. I didn't think to look for `3001` in the deprecations - I was just looking for Sodium, otherwise I would have removed these, too.

Note that we'll also need to update the changelog with the removals. My experience is that it's best to do one commit per deprecation removal. And create an associated issue to track, so an appropriate changelog entry can be generated.

salt/modules/keystone.py
waynew: This one should *not* be removed yet - there's about a week's worth of work (at least) to write tests for they keystone state and have it use the keystoneng module instead. That will have to happen for the Magnesium release at this point.

dwoz (approved):

PR #57460, opened 2020-05-26
Fix test under PyTest.

s0undt3ch:
### What does this PR do?
The `setUp` function relied on a module function which used
`__context__` which is only available when the module is loaded
through Salt's loader.
The test should have, in fact, failed when running under
`runtests.py` but it didn't, which suggests the module's global
scope is _dirty_.

This PR stops the test case from relying on the `__context__` dunder.

https://jenkinsci.saltstack.com/view/Master%20Branch%20Pytest%20Jobs/job/pr-windows2016-py3-pytest/job/master/215/testReport/tests.integration.modules.test_chocolatey/ChocolateyModuleTest/test_list_/

### DIFF
--- a/tests/integration/modules/test_chocolatey.py
+++ b/tests/integration/modules/test_chocolatey.py
@@ -3,30 +3,30 @@
 from __future__ import absolute_import
 
 import pytest
-import salt.modules.chocolatey as choco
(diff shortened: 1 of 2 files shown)


s0undt3ch:
re-run full all

PR #57481, opened 2020-05-27
[Wip] win reg

cmcmarrow:
### What does this PR do?
_winreg = None on windows should help with re.compile time. So every time we MP we should see 0.5 speed improvement. For 1000s of test we should see a 10 to 15 min speed improvement.
 
Note: _winreg is None on non windows platforms. 

### Commits signed with GPG?
Yes


cmcmarrow:
Looks like because of the salt lazy loader the _winreg trick does not work

PR #57489, opened 2020-05-28
Avoid HAS_DOCKER true if import messes with "salt.utils.docker"

meaksh:
### What does this PR do?
This PR should fix an issue on `swarm` execution module when executing a `salt-call` in debug mode and python docker libraries are not installed:

```console
2606fc6a24c5:~ # salt-call --local -ldebug swarm.foobar
[DEBUG   ] Missing configuration file: /etc/salt/minion
[DEBUG   ] Using cached minion ID from /etc/salt/minion_id: 2606fc6a24c5
[DEBUG   ] Configuration file path: /etc/salt/minion
[WARNING ] Insecure logging configuration detected! Sensitive data may be logged.
[DEBUG   ] Grains refresh requested. Refreshing grains.
[DEBUG   ] Missing configuration file: /etc/salt/minion
[DEBUG   ] Elapsed time getting FQDNs: 0.118145942688 seconds
[DEBUG   ] Could not determine init system from command line: (/bin/bash)
[DEBUG   ] The `lspci` binary is not available on the system. GPU grains will not be available.
[DEBUG   ] LazyLoaded zfs.is_supported
[DEBUG   ] Determining pillar cache
[DEBUG   ] LazyLoaded jinja.render
[DEBUG   ] LazyLoaded yaml.render
[DEBUG   ] LazyLoaded jinja.render
[DEBUG   ] LazyLoaded yaml.render
[DEBUG   ] Error loading module.swarm: __init__ failed
  File "/usr/lib/python2.7/site-packages/salt/loader.py", line 1716, in _load_module
    module_init(self.opts)
  File "/usr/lib/python2.7/site-packages/salt/modules/swarm.py", line 55, in __init__
    __context__['client'] = docker.from_env()
AttributeError: 'module' object has no attribute 'from_env'
[DEBUG   ] Missing configuration file: /etc/salt/minion
[DEBUG   ] Using cached minion ID from /etc/salt/minion_id: 2606fc6a24c5
...
```

The `HAS_DOCKER` flag is not working as expected when python docker library is not installed on the minion. It seems `import docker` works but ends up importing `salt.utils.docker` instead, so the `__init__` function will fail because `docker.from_env` does not exist.

This PR increase the checks for setting `HAS_DOCKER` to `true`.

- [ ] Docs
- [ ] Tests written/updated - no tests at all for `swarm` module.

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/modules/swarm.py
+++ b/salt/modules/swarm.py
@@ -32,10 +32,8 @@ import salt.utils.json
 
 
 def _is_docker_module(mod):
-    if hasattr(mod, "APIClient") and hasattr(mod, "from_env"):
-        return True
-    else:
-        return False
+    required_attrs = ["APIClient", "from_env"]
+    return all(hasattr(attr) for attr in required_attrs)
 
 
 try:
--- a/salt/modules/swarm.py
+++ b/salt/modules/swarm.py
@@ -33,7 +33,7 @@ import salt.utils.json
 
 def _is_docker_module(mod):
     required_attrs = ["APIClient", "from_env"]
-    return all(hasattr(attr) for attr in required_attrs)
+    return all(hasattr(mod, attr) for attr in required_attrs)
 
 
 try:
--- a/salt/modules/swarm.py
+++ b/salt/modules/swarm.py
@@ -30,13 +30,18 @@ from __future__ import absolute_import, print_function, unicode_literals
 # Import Salt libs
 import salt.utils.json
 
-HAS_DOCKER = False
+
+def _is_docker_module(mod):
+    if hasattr(mod, "APIClient") and hasattr(mod, "from_env"):
+        return True
+    else:
+        return False
+
 
 try:
     import docker
 
-    if hasattr(docker, "from_env"):
-        HAS_DOCKER = True
+    HAS_DOCKER = _is_docker_module(docker)
 except ImportError:
     HAS_DOCKER = False
 
--- a/salt/modules/swarm.py
+++ b/salt/modules/swarm.py
@@ -1,4 +1,3 @@
-# -*- coding: utf-8 -*-
 """
 Docker Swarm Module using Docker's Python SDK
 =============================================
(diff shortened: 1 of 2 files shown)


salt/modules/swarm.py
```diff
@@ -30,10 +30,13 @@
 # Import Salt libs
 import salt.utils.json
 
+HAS_DOCKER = False
+
 try:
     import docker
 
-    HAS_DOCKER = True
+    if hasattr(docker, "from_env"):
```
waynew, line 38: I would like to see this a standalone function for checking attributes that we use for this file.

The other approach that we could take is determining if this is a NamespacePackage vs an actual file. Though if we dump this check into a function, e.g. `HAS_DOCKER = _is_docker_module(docker)`, we could do multiple checks to verify that the docker module looks like what it should look like.

The advantage of that is easily being able to test the function in question, and being able to expand or narrow the definition as needed.

waynew (changes requested):

salt/modules/swarm.py
meaksh: Thanks for the suggestions @waynew - I've introduce that function and added unit tests :+1:

salt/modules/swarm.py
```diff
@@ -30,10 +30,18 @@
 # Import Salt libs
 import salt.utils.json
 
+
+def _is_docker_module(mod):
+    if hasattr(mod, "APIClient") and hasattr(mod, "from_env"):
+        return True
+    else:
+        return False
```
waynew, line 38:
```suggestion
    required_attrs = ["APIClient", "from_env"]
    return all(hasattr(attr) for attr in required_attrs)
```
Pretty sure that's the right logic. Not a *necessary* change, but it makes it easier to add required_attrs in the future.

waynew (dismissed):
This looks good to me. I've noted an optional improvement to the `_is_docker_module` function if you'd like to change that, but it's not required.

waynew (dismissed):
Thanks for those changes, looks great! This should be able to get merged after the code freeze is over.

dwoz (approved):

max-arnold:
Is this change really required? It looks like `salt.utils.docker` was renamed to `salt.utils.dockermod` on Apr 23 and released in Salt Sodium: https://github.com/saltstack/salt/pull/56857

s0undt3ch:
Hmm, don't think it's needed anymore...

PR #57491, opened 2020-05-28
Accept nested namespaces in spacewalk.api

### LINKED ISSUE #57442: [BUG] Spacewalk runner does not properly parse method namespaces
**Description**
The spacewalk runner does not properly parse the method namespaces. It assumes that a simple .split('.') will result in a two element list [ namespace, method ]. This works in simple cases such as 'system.listSystems' but fails for namespaces such as 'channel.software.listChildren' which results in three elements and causes the following exception:
```
Exception occurred in runner spacewalk.api: Traceback (most recent call last):
  File "/usr/lib/python3.6/site-packages/salt/client/mixins.py", line 377, in low
  File "/usr/lib/python3.6/site-packages/salt/runners/spacewalk.py", line 175, in api
    namespace, method = command.split('.')
ValueError: too many values to unpack (expected 2)
```

**Setup**
N/A

salt-run $spacewalkserver spacewalk.api channel.software.listChildren $basechannel

**Expected behavior**
Should not cause a stack trace

**Screenshots**

**Versions Report**
Irrelevant. Bug has been present since runner was initially committed.
I'm running a 2019.2 branch and it's still present in Salt 3000.1

**Proposed Fix**

```
salt/runners/spacewalk.py
168c168,169
<     namespace, method = command.split(".")
---
>     namespace = ".".join(command.split('.')[:-1])
>     method = command.split('.')[-1]
```

agraul:
### What does this PR do?
`salt-run $server spacewalk.api` allows users to run arbitrary Spacewalk API functions through Salt. These are passed in a namespace.method notation and may use nested namespaces. Previously only methods in a top-level namespace were supported.

Fixes: https://github.com/saltstack/salt/issues/57442

### Previous Behavior
```
Exception occurred in runner spacewalk.api: Traceback (most recent call last):
  File "/usr/lib/python3.6/site-packages/salt/client/mixins.py", line 377, in low
  File "/usr/lib/python3.6/site-packages/salt/runners/spacewalk.py", line 175, in api
    namespace, method = command.split('.')
ValueError: too many values to unpack (expected 2)
```

### New Behavior
API gets called as expected, e.g.
```
channel.software.listChildren ('sle-product-sles15-sp1-pool-x86_64',):
    |_
      ----------
      arch_label:
          channel-x86_64
      arch_name:
          x86_64
[...]
```

- [ ] Docs
- [x] Tests written/updated

Does this need to be documented in addition to the changelog? As there are no spacewalk runner tests yet I will need some time writing them, I opened the PR already to get feedback on the fix.

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/runners/spacewalk.py
+++ b/salt/runners/spacewalk.py
@@ -1,4 +1,3 @@
-# -*- coding: utf-8 -*-
 """
 Spacewalk Runner
 ================
@@ -29,13 +28,10 @@ master configuration at ``/etc/salt/master`` or ``/etc/salt/master.d/spacewalk.c
     not using the defaults. Default is ``protocol: https``.
 
 """
-from __future__ import absolute_import, print_function, unicode_literals
 
-# Import python libs
 import atexit
 import logging
 
-# Import third party libs
 from salt.ext import six
 
 log = logging.getLogger(__name__)
@@ -62,7 +58,7 @@ def _get_spacewalk_configuration(spacewalk_url=""):
 
     if spacewalk_config:
         try:
-            for spacewalk_server, service_config in six.iteritems(spacewalk_config):
+            for spacewalk_server, service_config in spacewalk_config.items():
                 username = service_config.get("username", None)
                 password = service_config.get("password", None)
                 protocol = service_config.get("protocol", "https")
@@ -76,7 +72,7 @@ def _get_spacewalk_configuration(spacewalk_url=""):
                     return False
 
                 ret = {
-                    "api_url": "{0}://{1}/rpc/api".format(protocol, spacewalk_server),
+                    "api_url": "{}://{}/rpc/api".format(protocol, spacewalk_server),
                     "username": username,
                     "password": password,
                 }
@@ -127,7 +123,7 @@ def _get_session(server):
 
     config = _get_spacewalk_configuration(server)
     if not config:
-        raise Exception("No config for '{0}' found on master".format(server))
+        raise Exception("No config for '{}' found on master".format(server))
 
     session = _get_client_and_key(
         config["api_url"], config["username"], config["password"]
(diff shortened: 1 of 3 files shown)


Akm0d:
@agraul Thank you for opening a PR so quickly!  Let me or @waynew know if you need help with writing the tests

salt/runners/spacewalk.py
```diff
@@ -180,7 +180,14 @@ def api(server, command, *args, **kwargs):
...
 
-    namespace, method = command.split(".")
+    command = command.split(".")
+    if len(command) == 1:
+        return {
+            call: "Error: command must use the following format: 'namespace.method'"
+        }
+
+    method = command.pop()
+    namespace = ".".join(command)
```
waynew, line 190: This could be simplified somewhat:

```suggestion
    namespace, _, command = command.rpartition(".")
    if namespace:
        return {
            call: "Error: command must use the following format: 'namespace.method'"
        }
```

agraul: that is elegant! Did not know about `rpartition`, thanks :+1:

waynew: FWIW most string methods have equivalent `r` methods that do things from the right side: split, rsplit, partition, rpartition, find, rfind, etc.

Granted, split and rsplit are identical if you don't provide a `maxsplit`, but... yeah, lots of super useful string methods :slightly_smiling_face:

waynew:
An odd failure on centos happening here - I've asked the team if they've seen anything. I guess I can kick these off again and see if it was just an infrastructure thing

agraul:
I'll push the tests on Tuesday (Monday is a public holiday where I live), I have something that works but I want to clean it up a little.

agraul:
Tests are there now, I got a pylint warning for the test:
> tests/unit/runners/test_spacewalk.py:1: [W9903(no-encoding-in-file), ] PEP263: Use UTF-8 file encoding

With Python3 this can be left out in my opinion (with UTF-8 being the default), but I am not sure if this is an oversight w.r.t. the pylintrc or you want to keep it even for a Python3-only codebase. Let me know if I should add it.

waynew (changes requested):
It's probably easiest for now to just go ahead and add the encoding line.

If we change it, it'll just be one more file for `sed` or `awk` to change. And that's the only thing that looks to need any changes here :+1:

agraul:
I've added the line, thanks for your review @waynew

waynew (dismissed):
:rocket:

PR #57507, opened 2020-05-31
[RC1 Sodium] sphinx 3.0.2 has bugs

bryceml:
### What does this PR do?
it fixes building the docs on python3.8 and probably python3.9

### Previous Behavior
Sphinx would get stuck building the docs on python3.8 and probably python3.9

### New Behavior
Sphinx does not get stuck building the docs on python3.8 and probably python3.9


bryceml:
See also https://saltstackcommunity.slack.com/archives/CRJPRHHT3/p1590891872113800

bryceml:
See https://www.sphinx-doc.org/en/master/changes.html#release-3-0-3-released-apr-26-2020
as well.

s0undt3ch (approved):

sagetherage:
@bryceml is this a legit test failure?

bryceml:
> @bryceml is this a legit test failure?

we don't test building the docs on python3.8, so we wouldn't have caught this with the current tests.

Akm0d (approved):

PR #57525, opened 2020-06-02
Update IPMI module to support local device

nanderson94:
### What does this PR do?

Pyghmi supports local devices natively, and you could even get it to
*kinda* work as-is, for example:
```
salt-call ipmi.get_health api_host=None
```

It will successfully pull the health information, however it'll throw
fatal when trying to cleanup the session, since localsession doesn't
have a logout capability.

The three primary fixes:
  * Don't logout for a localsession in IpmiCommand / IpmiSession
  * IpmiSession needed a bit more work, the private session class
    is used only for LAN connections, so I updated it to follow the same
    'bmc is None' logic as pyghmi's Command.
  * Following convention of ipmitool, pyghmi, racadm, hponcfg, etc,
    default to the local device.

That last point could potentially cause issues with current deployments,
but I don't think anyone is successfully hitting an ipmi device at
127.0.0.1.

I tested these changes on Dell 14G and HP Gen9.

- [ ] Docs
- [ ] Tests written/updated

### Commits signed with GPG?
Yes


DmitryKuzmenko:
@nanderson94 thank you for contribution. Your changes looks good for me as far as I can judge by quick lookup the `pyghmi.ipmi` code.
Unfortunately we do not merge any code not covered by tests anymore. Could you please add a unit test covering your changes?

nanderson94:
@DmitryKuzmenko There appears to be no test cases for this module currently. Are you looking for a test of just the functionality I implemented or for the general ipmi module?

DmitryKuzmenko:
@nanderson94 We can not enforce contributors to write tests for all the code. So if it is not interesting for you just test your changes. Thanks for asking.

garethgreenaway (approved):

dwoz:

Issue #57532, opened 2020-06-03
[BUG] [RC1 Sodium] we can't run on less than pyzmq 17 anymore


bryceml:
**Description**
We only seem to work on pyzmq 17 and newer, can't remember why exactly.  We should update our requirements to a minimum of pyzmq 17.0.0.

dmurphy18:
let me add the requirement for pyzmq >= 19.0.0 for Py 3.8 is also onerous and is there only to make Salt testing with wheels easier.

Fedora 32 ships with python3-zmq v18.1.2, v19.0.0 is in Rawhide
Ubuntu 20.04 ships with 17.1.2-3ubuntu1, again an issue with Py 3.8  and zmq being less than requirements

sagetherage:
@bryceml @s0undt3ch can we close this?

s0undt3ch:
I'll @bryceml handle this since he's the one that worked on it.

PR #57548, opened 2020-06-04
[BUG] [RC1 Sodium] update pyzmq requirement

bryceml:
### What does this PR do?
it updates the requirements on pyzmq since we are using functions that are only present in 17 and newer.
#57532

### DIFF
--- a/requirements/zeromq.txt
+++ b/requirements/zeromq.txt
@@ -1,5 +1,4 @@
 -r base.txt
 -r crypto.txt
 
-pyzmq>=2.2.0; python_version < '3.8'
-pyzmq>=19.0.0; python_version >= '3.8'
+pyzmq>=17.0.0


s0undt3ch (dismissed):

garethgreenaway (dismissed):

requirements/zeromq.txt
```diff
@@ -1,5 +1,5 @@
 -r base.txt
 -r crypto.txt
 
-pyzmq>=2.2.0; python_version < '3.8'
+pyzmq>=17.0.0; python_version < '3.8'
 pyzmq>=19.0.0; python_version >= '3.8'
```
dmurphy18, line 5: Can we change this too since any pip installing into Fedora or Ubuntu 20.04 will fail.  F32 is v18.1, Ubuntu 20.04 I believe is 17.1.2+

dmurphy18 (changes requested):

requirements/zeromq.txt
bryceml: I believe the purpose in pinning to 19 or higher was that there are wheel files available. It looks like there are wheels for previous versions though.

bryceml: https://pypi.org/project/pyzmq/17.0.0/#files

bryceml: oh, they aren't python3.8 wheels

dmurphy18: I know, but breaking people just because it makes our testing life easier is asinine - @s0undt3ch and I argued it out till blue in the face and no resolution.  However, I get Fedora 32, and salt will not pip3 install out of the box.

bryceml: looks like 18.1.0 has them though.

bryceml: 18.1.1 if you want to include windows.

dmurphy18: Fedora 31 is 18.0.2, what is Ubuntu 20.04

bryceml: 18.1.1 https://pkgs.org/search/?q=python3-zmq

bryceml: ubuntu/debian didn't complain though.

s0undt3ch: Remove both lines and leave just `pyzmq>=17.0.0`.
Locked requirements won't change and everyone stays less blue :grin:

bryceml: k, did that, force pushed the change.

dmurphy18: Debian / Ubuntu don't appear to check setuptools the way the red hat family do
But someone is going to complain about it

PR #57558, opened 2020-06-05
use salt.cmd.run_stdout

bochi:
usually, one does not want the stderr here because it may result in non-parsable yaml

### What does this PR do?
It uses cmd.run_stdout to surpress messages on stderr

Fixes:
As an example, when using the foreman salt plugin, foreman-node (which is used as command to get the ENC data for a minion) may display error messages or even only warnings for rubygems on stderr, which results in broken yaml and makes salt fail. 

I strongly believe stderr is never required here, any command acting as ENC command should output relevant data on stdout.

### Previous Behavior
YAML Data failed to parse when there are any warnings

### New Behavior
YAML Data parses fine even with warnings

- [ ] Docs
- [ ] Tests written/updated

### Commits signed with GPG?
No


waynew (changes requested):
Hey @bochi I definitely agree that we don't want to be parsing other pieces of data here.

I *do* think that we should probably dump the stderr to `log.debug` maybe.

We also need a changelog entry and some tests to cover this. If you need help writing tests I run a Test Clinic [on Twitch](https://www.twitch.tv/saltstackinc/schedule) Tuesday & Thursday and I'd be happy to help out here. For the changelog entry - https://docs.saltstack.com/en/master/topics/development/changelog.html describes how to add that.

You can also get me in #salt on freenode, as well as the community Slack. Let me know if you have any questions!

dwoz:

PR #57563, opened 2020-06-05
Daemons wrongly set six.string_types to (str, bytes)

DmitryKuzmenko:
### What does this PR do?
All here is about Python3 specifics, but I'm not breaking Python2.

Daemons wrongly set `six.string_types` to `(str, bytes)` that takes permanent effect.
I've analyzed the historical reason of that behavior, it was a part of initial py3 support and then wasn't updated properly while moving to unicode. I want to check the complete tests run here to understand what could be broken by this.

Fixes: N/A. I've found it during reproducing and fixing another issue but this change is not related.

### Previous Behavior
This produces failures in where we are using `str`-specific methods after `if isinstance(something, six.string_types)` because `bytes` strings are passing the check.

### New Behavior
Leave `six.string_types == (str,)`.

- [ ] Docs
- [ ] Tests written/updated

### Commits signed with GPG?
Yes


DmitryKuzmenko:
re-run full all

s0undt3ch (dismissed):
Erm... Why was this done in the first place!?

salt/beacons/btmp.py
```diff
@@ -295,7 +281,7 @@ def beacon(config):
             event = {}
             for ind, field in enumerate(FIELDS):
                 event[field] = pack[ind]
-                if isinstance(event[field], salt.ext.six.string_types):
```
DmitryKuzmenko, line 298: Fix for test. Here's an example of wrong usage of `six.string_types` that is expected by this logic to pass for `bytes` as we can see on the next line.

salt/beacons/wtmp.py
```diff
@@ -338,22 +323,22 @@ def beacon(config):
             event = {}
             for ind, field in enumerate(FIELDS):
                 event[field] = pack[ind]
-                if isinstance(event[field], six.string_types):
```
DmitryKuzmenko, line 341: The same as above.

salt/daemons/__init__.py
```diff
@@ -1,24 +1,16 @@
...
 import sys
 from collections.abc import Iterable, Mapping, Sequence
 
-# Import Salt Libs
 from salt.ext import six
 
 log = logging.getLogger(__name__)
 
-if sys.version_info[0] == 3:
-    six.string_types = (six.text_type, six.binary_type)
```
DmitryKuzmenko, line 20: This is the only thing I wanted to change by this PR. `six.string_types` is set in `salt.ext.six` and only there. It must not be updated outside the `six` module. Moreover this like updates the `six.string_types` value for the loaded module so it changes the `string_types` for all the code in the current process.

One more interesting thing is: runtests is vulnerable for this while pytest is not.

salt/renderers/aws_kms.py
```diff
@@ -239,14 +231,14 @@ def _decrypt_object(obj, translate_newlines=False):
     """
     if salt.utils.stringio.is_readable(obj):
         return _decrypt_object(obj.getvalue(), translate_newlines)
-    if isinstance(obj, six.string_types):
```
DmitryKuzmenko, line 242: The same as in first two comments.

DmitryKuzmenko (commented):
I've commented out all 4 things I've changed for this PR. Others are automated pre-commit updates.

DmitryKuzmenko:
I hope tests would pass now. @dwoz could you please take a look to my comments when you'll have time? I don't see a field for writing tests here: 3 fixes are fixing the code that is already covered by tests, 1 fix is removing wrong code that logically breaks things.

Writing a test like
```python
import daemons
assert six.string_types == str,
```
looks silly because I don't see any person who will return this back. Moreover we're dropping six from salt. It's just another one small step.
@s0undt3ch your review is dismissed. Could you please re-review this?

salt/daemons/__init__.py
s0undt3ch: Heh, usually it's the other way around.....

s0undt3ch (approved):

salt/beacons/btmp.py
waynew: I don't know if it would be worthwhile to write the test, but you *could* write a test, probably mocking a bunch of stuff to make it to this point, that would get you some bytes instead of string types/text.

waynew: The failing test would fail to convert from bytes to unicode/text, of course.

salt/daemons/__init__.py
waynew: I uh... why are we setting `six.string_types` in *this* module? 😬 

I mean, we're not now but... yikes!

dwoz (approved):

salt/daemons/__init__.py
frebib: I spent several hours second-guessing my own existence over this. I thought I was going crazy!

PR #52362, opened 2019-03-29
Fixes for acme.cert

OrangeDog:
Various fixes for `acme.cert`.

- Correct implementation of `certname`
- Fix for `py3`
- Only report changes if something actually changed

fixes https://github.com/saltstack/salt/issues/56473

PR #57569, opened 2020-06-05
Acme patch

### LINKED ISSUE #52361: [BUG] certname in acme module doesn't work
The acme state and module functions take an optional `certname` argument that's supposed to change the default CN-based name.

However, it's only use is to pass the `--cert-name` parameter to certbot, causing it to create the certificate correctly at the `certname` path.
Everything else then fails, because it looks for files using the `name` instead of `certname`.

As far as I can tell, this feature has never worked since it was added by @mprenditore in https://github.com/saltstack/salt/commit/665803a

nee2c:
### What does this PR do?
Corrects the use of the variable certname in module.acme and state.acme.

Fixes: #52361

### Previous Behavior
Fails to find cert files to update if certname variable is unused.

### New Behavior
Locates cert files to be tested and update

- [ x ] Docs
- [ x ] Tests written/updated

### Commits signed with GPG?
No


OrangeDog:
[This change](https://github.com/saltstack/salt/pull/52362/commits/512c1f6c9ac8abfcaf3a8d66184f7e16582074ca) may still be needed too.

nee2c:
I dropped that change due to conflicts no longer needed it.

cmcmarrow (approved):

cmcmarrow:
@nee2c thanks for the PR. Everything looks good. This PR should be merged soon.

PR #57607, opened 2020-06-09
Fix py3 mac os build script

weswhet:
### What does this PR do?
This PR adds some improvements to the macOS pkg build script, they are as follows.
- Link the compiled OpenSSL to the python build, this broke with the py3 change. It now leverages the included pkg-config to do the proper linking.
- Remove python2 code.
- Builds python with the ``--enable-optimizations`` flag for increased performance, (this is optional with a new flag to easily test the script without waiting for the optimized python)
- move the `pyobjc` library back into req.txt and update to version 6.2.
- The script will now fail early if the build_env.sh script fails. Before it would continue on.
- Updates the `maxfiles` set to better handle the zeromq tests (these are still flaky on Catalina)

- [ ] Docs
- [ ] Tests written/updated

### Commits signed with GPG?
Yes


Akm0d (dismissed):

pkg/osx/req.txt
```diff
@@ -18,6 +18,7 @@ psutil==5.6.6
 pyasn1==0.4.8
 pycparser==2.19
 pycryptodomex==3.9.7
+pyobjc==6.2
```
s0undt3ch, line 21: This needs to be as it was, though the version can/should be upgraded and the `.txt` compiled again.
The reason being that static requirements compilation doesn't not handle pyobjc because it tries to inspect the system at install/build time.

s0undt3ch (changes requested):

pkg/osx/req.txt
weswhet: okay, can do.

s0undt3ch: Thank You!

s0undt3ch (dismissed):

twangboy (dismissed):

Akm0d (dismissed):

weswhet:
@dwoz all the changes here are shell script related. Is there a testing framework for bash that I'm missing?

weswhet:
@sagetherage would love to get this into 3001.1 🙏

weswhet:
re-run centos8

PR #57625, opened 2020-06-10
Install KB2999226 from .cab instead of .msu

berkut-174:
### What does this PR do?

This fixes the installation of update (KB2999226) when the Windows Update service is disabled.

Fixes:

### Previous Behavior
If the Windows Update service is disabled, then the update (KB2999226) will not be installed, as in the current version this is done in `wusa`.

### New Behavior
Update (KB2999226) will be installed regardless of the status of the Windows Update service.

- [ ] Docs
- [ ] Tests written/updated

### Commits signed with GPG?
No

### DIFF
--- a/pkg/windows/build_env.ps1
+++ b/pkg/windows/build_env.ps1
@@ -169,7 +169,7 @@ If (Test-Path "$($ini[$bitPaths]['VCppBuildToolsDir'])\vcbuildtools.bat") {
 #------------------------------------------------------------------------------
 Write-Output " - Checking for Python 3 installation . . ."
 If (Test-Path "$($ini['Settings']['Python3Dir'])\python.exe") {
-    # Found Python 3.5, do nothing
+    # Found Python 3, do nothing
     Write-Output " - Python 3 Found . . ."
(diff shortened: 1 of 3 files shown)


twangboy (approved):
Have you tested this?

cmcmarrow (approved):

berkut-174:
> Have you tested this?

Hi, yes of course I checked on my Windows 7 64-bit and Windows 8.1 32 & 64-bit.
Thank.

twangboy:
@berkut-174 After discussion with the Core team and the Windows Working group we have come to the conclusion that it is not good practice to patch the operating system from the Salt installer. The System Administrator should be in control of system patches and not have a 3rd party installer unknowingly update their systems.

As a result of this determination, we will be removing the code that applies KB2999226 from the `.exe` installer. We will be adding documentation to the Windows download page that specifies the Python requirement for KB2999226. We will submit an issue to Python to add that requirement to their site as well.

However, your solution to this issue is appreciated. I did test it as well and it does work. Sorry we decided to go another direction here.

PR #56353, opened 2020-03-11
[cherrybomb] MTecknology's Masterless Contributions

MTecknology:
### What does this PR do?

:musical_note: Hello Salt. Hello Friends. Here's your... ch-ch-ch-cherry bomb! :musical_note:

This PR cherry-picks of all my previously-accepted PRs which have not been merged from ``develop`` to ``master`` and do not appear to be part of any current merge effort. These previously accepted PRs are also present in the [PRs to port to master](https://github.com/saltstack/salt/projects/5) project board and have been (personally) reviewed for current relevancy/accuracy.

The two areas of salt impacted by these changes are: ['Documentation', 'Debian Networking]

### What PRs* does this PR fix or reference?

- #50070
- #49962
- #49940
- #49906
- #49887
- #49843
- #49462
- #49355

### Tests included*?

Yes; I may have gotten a tiny bit carried away with creativity, but I also tried to break my own code and documented even the silly attempts as tests.

### Commits signed with GPG?

Yes, absolutely.

PR #54572, opened 2019-09-19
[master] Porting #52670 to master

garethgreenaway:
Porting #52670 to master

Issue #57541, opened 2020-06-04
[Tech-Debt] Avoid use of deprecated net-tools


OrangeDog:
**Is your feature request related to a problem? Please describe.**
The linux `net-tools` package has been deprecated since 2011, and provides commands including `arp`, `ifconfig`, `iptunnel`, `iwconfig`, `nameif`, `netstat`, and `route`.

Some salt modules make use of these commands, causing failures when they're not available (#57513).

**Describe the solution you'd like**
Code should be rewritten, preferring first native python (e.g. `pyroute2`) then the `iproute2` tools (e.g. `ip`), then only falling back to `net-tools` if that's all that is available. If every supported platform has already migrated to `iproute2` then the fallback is not required.

**Describe alternatives you've considered**
Continue to depend on systems providing the `net-tools` package. This won't work forever.

**Additional context**
Replacement commands:

net-tools | iproute2
-- | --
arp | ip neighbor
ifconfig | ip addr, ip link, ip -stats
iptunnel | ip tunnel
iwconfig | iw
nameif | ip link
netstat | ss
route | ip route

sagetherage:
I don't know if I would called this a feature - I see it as an enhancement to something that currently exists and is not working due to an upstream dependency deprecation some time ago (2011?!), but I don't have a template for `enhancement` :) so this works

OrangeDog:
@sagetherage the only options are Bug and Feature

sagetherage:
yes, I put those in place, mostly putting notes for myself :)

OrangeDog:
Closely related: Ubuntu doesn't use `ifupdown` any more, but `netplan`. The file `/etc/network/interfaces` is also not used.

OrangeDog:
I think the whole of `debian_ip` needs rewriting. Every config file and tool it uses is obsolete.
Related: #54791

dmurphy18:
Note: the old net-tools cannot be totally removed since still have to support older OS's, for example: Solaris 10, some of the underlying OS's on some switch/routers, AIX.  But as stated above, have the net-tools as a fall back for when more modern implementations are not available.

dmurphy18:
Capturing a thread from community slack here

From: OrangeDog
```
also NetworkManager vs. systemd-networkd / ifupdown vs. netplat, etc.
```

From terminalmage
```@OrangeDog for a second I thought you were referring to network teaming for network.managed states on EL7 and 8
I actually have implemented that at work and plan to submit upstream soon
```

[...]

PR #57637, opened 2020-06-11
[master] Porting #52670 to master again

OrangeDog:
Porting #52670 to master again, because the last one (#54572) got reverted by #56353.

When this is merged, https://github.com/saltstack/salt-pack/issues/682 needs reverting.


waynew (approved):

waynew (changes requested):
Ah, boo, looks like needs pre-commit hooks :upside_down_face:

OrangeDog:
> Ah, boo, looks like needs pre-commit hooks 🙃

Apparently so, but it already ran them when I committed,

waynew (approved):
:rocket:

bryceml:
don't ifup and ifdown set things up according to what's specified in /etc/network/interfaces as opposed to `ip link set up` which doesn't have anything to do with /etc/network/interfaces ?

This seems like it doesn't keep the same behavior at all if I understand it correctly.

s0undt3ch (approved):

bryceml (changes requested):
This would break existing use of ifupdown on debian 10, this probably needs a much bigger PR to do the correct thing.

OrangeDog:
@bryceml if that's true, and nobody noticed on the previous two PRs, it should be tracked with #57541 and related.

OrangeDog:
> It would probably be more correct to do something like nmcli con up ... or nmcli con down ... if using network manager. I'm not sure how you would get the equivilant on netplan since it only really has netplan apply

`netplan` is a higher-level config management system. The alternative to `NetworkManager` is `systemd-networkd`. I don't think there is any CLI other than `ip`. You just edit the config files and reload.

sagetherage:
@OrangeDog we want to have a deep discussion on this since likely needs to be bigger than this PR and we need to discuss a broader solution. I am putting it into blocked for now, but I will schedule a planning meeting to dive deeper. I would love your input, I know our US meeting are difficult to attend, but I am open to other ideas the use of other platforms if that is helpful.

OrangeDog:
I have no particular investment. I just noticed this PR has at least twice already been merged and then lost.

The concept is not difficult. You should be using the recommended commands on each of your supported operating systems and not relying on deprecated and removed packages. You need to work out what they are and then write that code. I think that `ip` is the default method on every linux OS version you support, but I may be mistaken.

dwoz:

PR #57662, opened 2020-06-13
add tiamat run.py script

bryceml:
### What does this PR do?
adds the run.py that tiamat uses to the salt repo so we don't have to maintain it separately


dmurphy18 (commented):
I wonder about adding packaging to salt, given the Redhat and Debian stuff is way out of date and no one has complained much.  Also packaging is always fixing up for when ever was needed to make it work on a platform post-tag, Solaris and AIX, and now native minion too.

Will look at tiamat next week and see if this is good or not then.

bryceml:
I think long term the goal is to have the packaging stuff merge with salt and every pr will be built.  I figured this was a good starting point so we could all be using the same run.py script and incorporating each other's changes.  This probably needs more discussion with everyone involved.

thatch45:
I think that this should be merged in, primarily because the run.py concept from POP is not JUST for packaging. It is also to make it easy to run the code directly from the repo. This is also a complicated run.py file, they are normally not this complicated, but Salt is big. So it does make sense that the `run.py` file is in the repo.

As for packaging components, I think that they should be removed from the Salt repo. Those components should be placed inside of repos that are used for packaging and kept separate from the Salt repo.

The packaging components inside of the Salt repo are mostly out of date and we don't build and manage most of our packages from them. So they are misleading.

thatch45:
Also, this should be in the project's root directory

felippeb (dismissed):

bryceml:
Building with tiamat on pr will be a good enough test.  It would be good to make sure we can pip install something too.

twangboy (approved):

ScriptAutomate (approved):

Akm0d (approved):

Issue #49672, opened 2018-09-17
pkg.installed not working correctly on ppc64le architecture


sjtbham:
### Description of Issue/Question
When running on `ppc64le` systems, with architecture based packages, salt is trying to repeatedly install the package when it is already installed. For example we want to ensure the `iptables` package is installed on the system via salt. This works as expected on `x86_64` arch systems, but doesn't work properly on `ppc64le` systems.

### Setup
Example state for package installation
```
iptables-pkg:
  pkg:
    - installed
    - name: {{ pillar.packages.names.iptables | default('iptables') }}
```

### Steps to Reproduce Issue
```# salt-call state.highstate
[ERROR   ] The following packages failed to install/update: iptables
```

On ppc64le salt client (trimmed to relevant output):
```
# salt-call state.highstate test=true
----------
          ID: iptables-pkg
    Function: pkg.installed
        Name: iptables
      Result: None
     Comment: The following packages would be installed/updated: iptables
     Started: 15:40:47.161451
    Duration: 91.094 ms
     Changes:
```

Listing the versions of the packages installed:
```
# salt-call pkg.list_pkgs
local:
    ----------
    iptables.ppc64le:
        1.4.21-24.1.el7_5
```

Whilst salt is running in non test mode, you can see it running processes like `yum -y install iptables` (in output from `ps -ef`) which salt things has failed (though the package is actually installed) I assume due to some architecture mismatching?

Checking what yum thinks is installed:
```
# yum list iptables
Loaded plugins: product-id, search-disabled-repos, subscription-manager, versionlock
Installed Packages
iptables.ppc64le                                                 1.4.21-24.1.el7_5                                                  @pkg1/7.5
```

### Versions Report
```
# salt-call --versions-report
Salt Version:
           Salt: 2018.3.2

Dependency Versions:
           cffi: 1.6.0
       cherrypy: Not Installed
       dateutil: 1.5
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
          ioflo: 1.3.8
         Jinja2: 2.7.2
        libgit2: Not Installed
        libnacl: Not Installed
       M2Crypto: 0.28.2
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.5.6
   mysql-python: Not Installed
      pycparser: 2.14
       pycrypto: 2.6.1
   pycryptodome: Not Installed
         pygit2: Not Installed
[...]

PR #57685, opened 2020-06-16
Fix yumpkg.py incorrect arch parsing for yum

CJCShadowsan:
This uses the cpuarch grain as opposed to the osarch grain, which reports incorrectly on ppc64le as powerpc64le - causing pkg.installed and pkg.latest checks to fail.

cpuarch also seems to work fine on x86_64 - don't have an aarch64 system to test on but I suspect it will also be fine. 

This also works within kvm.

### What does this PR do?
Fixes RPM arch parsing for multi-arch

Fixes: ??? There's a few issues this addresses, notably as an example people working around yumpkg.py elsewhere:

[https://wiki.nikhef.nl/grid/A_workaround_for_a_powerpc_saltstack_bug](https://wiki.nikhef.nl/grid/A_workaround_for_a_powerpc_saltstack_bug
)
### Previous Behavior
pkg installs on ppc64le do not work.

### New Behavior
pkg installs now work with no modifications for arch.

- [ ] Docs
- [ ] Tests written/updated

### Commits signed with GPG?
No

### DIFF
--- a/salt/modules/yumpkg.py
+++ b/salt/modules/yumpkg.py
@@ -186,7 +186,7 @@ def _yum_pkginfo(output):
     cur = {}
     keys = itertools.cycle(("name", "version", "repoid"))
     values = salt.utils.itertools.split(_strip_headers(output))
-    osarch = __grains__["osarch"]
+    osarch = __grains__["cpuarch"]
     for (key, value) in zip(keys, values):
         if key == "name":
             try:
@@ -441,8 +441,8 @@ def normalize_name(name):
             return name
     except ValueError:
         return name
-    if arch in (__grains__["osarch"], "noarch") or salt.utils.pkg.rpm.check_32(
-        arch, osarch=__grains__["osarch"]
+    if arch in (__grains__["cpuarch"], "noarch") or salt.utils.pkg.rpm.check_32(
+        arch, osarch=__grains__["cpuarch"]
     ):
         return name[: -(len(arch) + 1)]
     return name
@@ -562,9 +562,9 @@ def latest_version(*names, **kwargs):
         try:
             arch = name.rsplit(".", 1)[-1]
             if arch not in salt.utils.pkg.rpm.ARCHES:
-                arch = __grains__["osarch"]
+                arch = __grains__["cpuarch"]
         except ValueError:
-            arch = __grains__["osarch"]
+            arch = __grains__["cpuarch"]
 
         # This loop will iterate over the updates derived by _yum_pkginfo()
         # above, which have been sorted descendingly by version number,
@@ -709,7 +709,7 @@ def list_pkgs(versions_as_list=False, **kwargs):
         output = __salt__["cmd.run"](cmd, python_shell=False, output_loglevel="trace")
         for line in output.splitlines():
             pkginfo = salt.utils.pkg.rpm.parse_pkginfo(
-                line, osarch=__grains__["osarch"]
+                line, osarch=__grains__["cpuarch"]
             )
             if pkginfo is not None:
                 # see rpm version string rules available at https://goo.gl/UGKPNd


krionbsd:
Could you please show the output of `uname -p` on ppc64le?

CJCShadowsan:
> Could you please show the output of `uname -p` on ppc64le?

```
[root@gpu001 ~]# uname -p
ppc64le
```

But if I check the grains...

```
[root@gpu001 ~]# salt-call grains.item osarch
local:
    ----------
    osarch:
        powerpc64le
```

This breaks all sorts of package installations and means that normalize doesn't work also.

But if I use cpuarch:

```
[root@gpu001 ~]# salt-call grains.item cpuarch
local:
    ----------
    cpuarch:
        ppc64le
```

All is well with the world and everything salt wise works without faffing around with creating different formulas for different architectures.

krionbsd:
Hmm, salt uses `rpm --eval "%{_host_cpu}"` to determine osarch grain, but it seems RH mistakenly produces `powerpc64le` for this command. Did you bug RH with it before?

CJCShadowsan:
> Hmm, salt uses `rpm --eval "%{_host_cpu}"` to determine osarch grain, but it seems RH mistakenly produces `powerpc64le` for this command. Did you bug RH with it before?

I didn't but considering that yum itself works fine, they will say it is most definitely a salt issue, hence the proposed fix.

I'm happy to hear any other proposed fixes that work - i'm not precious about how it gets fixed as long as it gets fixed... 

...I just know this method works and is really low-touch so low impact elsewhere.

CJCShadowsan:
Note that this issue has been raised previously, and been closed previously with no fix:

[https://github.com/saltstack/salt/issues/49672](https://github.com/saltstack/salt/issues/49672)

It'd be nice to get a fix in, as opposed to people working around it!

krionbsd:
As processor type from `uname -p` and `rpm` differs I'd definitely bug RH to fix this behavior on their side, because it's wrong. Did you try to use a custom grain in the meantime?

CJCShadowsan:
Also:

This commit seems to not have been submitted as a pull request:

[https://github.com/waynew/salt/commit/7a21c6d7991856c871851f4683dc6ec31295ebe5](https://github.com/waynew/salt/commit/7a21c6d7991856c871851f4683dc6ec31295ebe5)

Which may fix it in a different manner - either way RH will say their OS works as-is and as such isn't a RH bug and i'd be inclined to agree with them.

The way Salt handles what the OS reports is a fundamental issue that needs fixing in Salt.

CJCShadowsan:
> As processor type from `uname -p` and `rpm` differs I'd definitely bug RH to fix this behavior on their side, because it's wrong. Did you try to use a custom grain in the meantime?

And therein lies the problem... Why should a custom grain be used for something that Salt should handle out of the box?

RH will say that all of the functions within their OS that have been put together for it as-is... Work. So it's not strictly a bug, else it would break functionality within the OS, which it doesn't.

This breaks a piece of software which handles package installation, which fails to interpret what the OS reports in a way that allows packages to install - of which the standard package manager successfully installs that software.

That, by definition is a problem with the software - which is in this case Salt.

CJCShadowsan:
> > As processor type from `uname -p` and `rpm` differs I'd definitely bug RH to fix this behavior on their side, because it's wrong. Did you try to use a custom grain in the meantime?
> 
> And therein lies the problem... Why should a custom grain be used for something that Salt should handle out of the box?
> 
> RH will say that all of the functions within their OS that have been put together for it as-is... Work. So it's not strictly a bug, else it would break functionality within the OS, which it doesn't.
> 
> This breaks a piece of software which handles package installation, which fails to interpret what the OS reports in a way that allows packages to install - of which the standard package manager successfully installs that software.
> 
> That, by definition is a problem with the software - which is in this case Salt.

Just playing devils advocate here but I mean, you know that the conversation will go "Well Ansible handles this without a problem..."

CJCShadowsan:
Just a thought - is this ever going to get looked at instead of being told "go raise something with RH"?

The CI checks not successful preventing merging will be because they need a test - but as I don't know what test(s) you would want to run for this, it becomes a bit of a moot point.

If the only thing preventing a merge is a valid test, then let me know the test you want to run and i'll put something together - but just being advised to go raise a bug with RH for something that:

1. Is a couple of words change here
2. That RH will say this works perfectly fine for the OS...
3. RH will come back to me with "Ansible works with this, so what's the problem?" 

Is a pretty sad state of affairs, is it not?

I'm trying to justify continued usage of SaltStack here, but to have something that realistically should work out of the box without hacks or modifications to grains, is making that justification pretty difficult - especially in the presence of a clear fix submitted back to yourselves to remedy the situation.

We'd be throwing away over 200k lines of configuration management to move away, but if we're going to be fighting the good fight here then fundamental things like this need to be addressed - else we'll have little choice. We don't have the time to be fighting the tide like this. Configuration management is meant to save time, not add time picking apart why the configuration management software doesn't work.

Sort it out guys!

dwoz:

CJCShadowsan:
As much as I wanted to help in the beginning (I mean, I submitted a PR...) - To be then told that this is a RedHat problem (even though it doesn't break any _other_ configuration manager) when a fix was right in front of you meant that the action was taken to move away from Salt. 

So even if I wanted to help and "see it through to the end in one release cycle", I no longer have the time nor the need to.

It's a shame, if it hadn't been for the "This is RedHats problem, not ours" attitude, this fix would have gone in over 3 years ago.

PR #57698, opened 2020-06-17
Fix current_zone_only not working in solarispkg

jnieuwen:
current_zone_only is a boolean that can be passed on the command line as a
keyword argument when using pkg on Solaris. This makes sure the code checks it
against the boolean value True instead of the string "True" (True is not
"True"). The 'in' construct to not break workarounds that people might have in
place. Further more the += '-G' option to append a string to a list causes
problems as it results in the '-' and 'G' being added as separate command line
arguments to pkgadd. Therefor change to using the list.append function.

### What does this PR do?
It makes sure that current_zone_only can be used when using pkg.install on Solaris.

Fixes:

### Commits signed with GPG?
No


cmcmarrow:
@jnieuwen can we get a test with  `current_zone_only` set as True? To help prevent the flag from breaking again? Thanks for the PR!

cmcmarrow (changes requested):
We can make line 360 `if kwargs.get("current_zone_only"):` `current_zone_only` should only ever be a Boolean and because we know this flag was 100% broken before your fix we know no one is using the flag. So we don’t need to vet the strings `“False”` or `”True”`. I also checked if salt passed `“True”` or “`False`” which it does not.

jnieuwen:
> We can make line 360 `if kwargs.get("current_zone_only"):` `current_zone_only` should only ever be a Boolean and because we know this flag was 100% broken before your fix we know no one is using the flag. So we don’t need to vet the strings `“False”` or `”True”`. I also checked if salt passed `“True”` or “`False`” which it does not.

Changed

jnieuwen:
> @jnieuwen can we get a test with `current_zone_only` set as True? To help prevent the flag from breaking again? Thanks for the PR!

I'll look into what is possible from my side.

cmcmarrow:
@jnieuwen Debian9 and pycryptodme tests fails are not related to your PR. Next time you push they should pass. If you need help with tests feel free to ping me. Here is a doc to help you with test writing https://docs.saltstack.com/en/master/topics/tutorials/writing_tests.html as well. Salt requires tests for all PRs that has code changes.

dmurphy18 (commented):
Changes looks good, note tests in tests/unit/modules/test_solarisipspkg.py could be used as an example for the requested test changes.

dmurphy18:
@jnieuwen Wondering what versions of Solaris and for which platforms you are running on ?

dmurphy18:
@jnieuwen Reminder, what versions of Solaris and platforms

sagetherage:
Since this is missing a test case we will need to move to the next release - updating.

dmurphy18 (changes requested):
@jnieuwen Are you still interested in this PR ?, if unresponsive the PR shall be closed after several days.
Test cases need addressing as well as review comments

Ch3LL:
Closing due to inactivity. Please let me know if you want me to re-open or feel free to open a new PR against the master branch with the included feedback.

Issue #56296, opened 2020-03-03
salt-minion fails startup on Windows Server 2012R2


xiaohuazi123:
### Description of Issue

salt-minion fails startup on Windows Server 2012R2，due to api-ms-win-crt-runtime-l1-1-0.dll is missing
![QQ截图20200303214632](https://user-images.githubusercontent.com/45115671/75782322-b8cc5600-5d99-11ea-94dc-f86029d76227.jpg)

### Setup
Nothing modified after installing the Salt-Minion-3000-Py3-AMD64-Setup.exe on Windows Server 2012R2
Basic setup, with default configuration files (except for minion id and master id, set during the setup).

### Steps to Reproduce Issue
1/ install Salt-Minion-3000-Py3-AMD64-Setup.exeon a Windows Server 2012R2
2/ start manually C:\salt\salt-minion.bat or C:\salt\salt-minion-debug.bat in a cmd in admin mode.
3/Pop up a dialog box saying (api-ms-win-crt-runtime-l1-1-0.dll) is missing

### Versions Report
Client : Microsoft Windows Server 2012 R2 Datacenter 6.3.9600  Build 9600
Salt-minion : Salt-Minion-3000-Py3-AMD64-Setup.exe

garethgreenaway:
@twangboy FYI.

twangboy:
Looks like the VCRedist is corrupted. Those are bundled in the installer. Try doing a repair in the Add/Remove programs.

xiaohuazi123:
@twangboy  Thanks for your reply, I have tried before

twangboy:
One of these was possibly corrupted during install...
http://repo.saltstack.com/windows/dependencies/64/ucrt/

xiaohuazi123:
Hi @twangboy ，
In another detailed test yesterday, I found that the salt-minion service could not be run on an unactivated Windows machine.

Here are my test results:

- Salt-Minion-3000.1-Py2-AMD64-Setup.exe:Can run the salt-minion service without activating Windows without any errors.
- Salt-Minion-3000.1-Py2-x86-Setup.exe:Can run the salt-minion service without activating Windows without any errors.
- Salt-Minion-3000.1-Py3-AMD64-Setup.exe:Cannot run the salt-minion service without activating Windows. The error message is that there is a lack of api-ms-win-crt-runtime-l1-1-0.dll .
[...]

PR #57700, opened 2020-06-17
Don't apply KB2999226 to Windows with NSIS Installer

### LINKED ISSUE #57594: salt-minion fails startup on Windows 8.1 x86 and x64, and likely any that doesn't include the relevant vc++ runtime
**Description**
salt-minion fails startup on Windows 8.1 x86 and x64 when using the Salt-Minion-3000.1-Py3-x86-Setup.exe or Salt-Minion-3000.1-Py3-x64-Setup.exe, due to api-ms-win-crt-runtime-l1-1-0.dll is missing.

**Setup**
This is actually due to wusa.. Stop the Windows Update service as if you were planning on using saltstack to bootstrap the windows updates.

If its easier, you can use the following packer template from github.com/boxcutter/windows: https://github.com/boxcutter/windows/blob/master/eval-win81x86-enterprise.json Most of the templates using the evaluation iso will have this issue actually.

Do something like `packer build eval-win81x86-enterprise.json`. The eval iso is at http://care.dlservice.microsoft.com/dl/download/B/9/9/B999286E-0A47-406D-8B3D-5B5AD7373A4A/9600.17050.WINBLUE_REFRESH.140317-1640_X64FRE_ENTERPRISE_EVAL_EN-US-IR3_CENA_X64FREE_EN-US_DV9.ISO if you want to pre-download it.

Run the installer (I've been using saltify-windows from salt-cloud, but you can run it normally too). It will fail to install the universal crt (http://repo.saltstack.com/windows/dependencies/64/ucrt/) due to using wusa which requires specific windows updates to be installed in order for wusa to work.

One symptom is that you'll see the salt-minion service hasn't automatically started and remains in the pause state. This is due to Python being unable to be executed.

Event log (Setup) will show the following depending on the patch that the salt-minion installer tried to install:
```
Windows update  could not be installed because of error 2147942402 "The system cannot find the file specified." (Command li ne: "wusa  "C:\Users\root\AppData\Local\Temp\nssB607.tmp\Windows8.1-KB2999226-x86.msu" /quiet /norestart")
```

You can prove that the crt didn't install by running `C:\salt\bin\python --version` and it will display an error from the runtime linker: "api-ms-win-crt-runtime-l1-1-0.dll is missing".

**Expected behavior**
Salt-minion service should be running, so then updates can be bootstrapped by salt...since we're trying to use salt-cloud after all. But because Python CRT couldn't install due to wusa the service is "paused" or really dead which requires manually repairing things.

**Versions Report**
salt-Minion-3000.3-Py3-x86-Setup.exe
salt-Minion-3000.3-Py3-x64-Setup.exe

This issue is the exact same one as #56296, and I documented a potential fix for it on that ticket.

twangboy:
### What does this PR do?
Removes code that applies KB2999226 to Windows using the NSIS installer. KB2999226 updates the Universal C Runtimes (UCRT) which is a requirement for Python 3 to run. For the past 2 years or so we have been patching Windows systems that were missing KB2999226. This applied to unpatched Win8.1/2012R2 and below using `wusa.exe`. There were some specific scenarios where the `wusa.exe` installation would fail. Specifically when the `wuauserv` service was not able to run for whatever reason. We decided that patching the operating system should be handled by the System admin.

However, looking at how Python handles this scenario, we noticed that on unpatched systems python dumps all the UCRT dlls into the root python directory. So, with this PR, we take the same approach. On unpatched systems, we will include the UCRT files in the `C:\salt\bin` directory so that python can run.

Fixes: https://github.com/saltstack/salt/issues/57594, https://github.com/saltstack/salt/issues/56296

- [ ] Docs
- [ ] Tests written/updated

### Commits signed with GPG?
Yes


cmcmarrow (dismissed):

dwoz (approved):

berkut-174:
Hi, how can I install `vcredist` in silent mode now?
The `Salt-Minion-3002-Py3-AMD64-Setup.exe /S` is not enough.

PR #57703, opened 2020-06-17
Tweak regex in salt-minion is_running to be more specific

mtslzr:
We run SaltStack on our customer's servers, and occasionally have customers that themselves wish to run SaltStack. So our implementation involved renaming our salt-minion binaries and startup script to `cd-salt-minion` to avoid confusion. The check in `is_running` falsely matches any other process containing `salt-minion`, causing the customer's service to fail to start (because it sees ours running in `ps` and exits out).

It's a super ticky-tack change, but didn't know if this might be helpful since it'll continue to match `salt-minion` as intended, and not (on the extreme off-chance) match something else unintentionally.

### What does this PR do?

Updates the grep called in `is_running` to match a leading slash and space after `salt-minion`. Intent is to match `/salt-minion` without. matching other process names containing `salt-minion` (e.g. `ab-salt-minion` or `salt-minion-xy`).

Fixes: N/A

((Should not change any known functionality, so marked these done. But open to changes, if needed.))

- [x] Docs
- [x] Changelog
- [x] Tests written/updated

### Commits signed with GPG?
No

### DIFF
--- a/pkg/rpm/salt-minion
+++ b/pkg/rpm/salt-minion
@@ -72,7 +72,7 @@ _get_pid() {
 
 
 _is_running() {
-    [ -n "$(_get_pid)" ] && ps wwwaxu | grep '[s]alt-minion' | awk '{print $2}' | grep -qi "\b$(_get_pid)\b"
+    [ -n "$(_get_pid)" ] && ps wwwaxu | grep '\/[s]alt-minion ' | awk '{print $2}' | grep -qi "\b$(_get_pid)\b"
 }
 
 


waynew:
@mtslzr thanks for the PR! I like this approach in principle, I'll add it to the core team discussion - someone may have other thoughts on it.

bryceml:
So everyone's aware this file also exists here: https://github.com/saltstack/salt-pack-py3/blob/develop/file_roots/pkg/salt/master/rhel8/sources/salt-minion

I haven't checked which one gets used, but my guess is that it's the one in salt-pack-py3.

waynew (changes requested):
We had a discussion internally - based on that discussion and the available information the recommendation from the team is that if you're running your own Salt, potentially along-side a customers Salt, you should actually be producing your own distribution of Salt for them to install. At that point you can call the process whatever you want, and make whatever necessary changes you need to your own distro.

If you've got some other information that you'd like to share or discuss, or this approach doesn't make sense, please come to the Open Hour on Thursday - You can see more details on the [Community Events calendar](https://calendar.google.com/calendar/embed?src=saltstack.com_md73c3ufcs2eqbsmmnike4em80%40group.calendar.google.com). Otherwise, we'll probably go ahead and close this PR, but thanks for bringing this up!

PR #57723, opened 2020-06-18
dont log errors in virtuals.

whytewolf:
### What does this PR do?
changes two log.errors into log.debug in a virtual function. as longing errors can get noisy here. 

### Previous Behavior
Noisy log from errors being logged for an unused returner. 

### New Behavior
The log will only be noisy in debug mode.

- [ ] Docs
- [ ] Tests written/updated - no, only changed two log lines.

Does this change require tests? It is literally only changing log.error to log.debug for two items. 

### Commits signed with GPG?
No

### DIFF
--- a/salt/returners/appoptics_return.py
+++ b/salt/returners/appoptics_return.py
@@ -96,10 +96,6 @@ log = logging.getLogger(__name__)
 
 def __virtual__():
     if not HAS_APPOPTICS:
-        log.debug(
-            "The appoptics_return module couldn't load the appoptics_metrics module."
-        )
-        log.debug("please make sure it is installed and is in the PYTHON_PATH.")
         return (
             False,
             "Could not import appoptics_metrics module; "
--- a/salt/returners/appoptics_return.py
+++ b/salt/returners/appoptics_return.py
@@ -96,10 +96,10 @@ log = logging.getLogger(__name__)
 
 def __virtual__():
     if not HAS_APPOPTICS:
-        log.error(
+        log.debug(
             "The appoptics_return module couldn't load the appoptics_metrics module."
         )
-        log.error("please make sure it is installed and is in the PYTHON_PATH.")
+        log.debug("please make sure it is installed and is in the PYTHON_PATH.")
         return (
             False,
             "Could not import appoptics_metrics module; "
--- a/salt/returners/appoptics_return.py
+++ b/salt/returners/appoptics_return.py
@@ -1,4 +1,3 @@
-# -*- coding: utf-8 -*-
 """Salt returner to return highstate stats to AppOptics Metrics
 
 To enable this returner the minion will need the AppOptics Metrics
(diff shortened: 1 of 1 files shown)


max-arnold:
https://docs.saltstack.com/en/latest/ref/modules/#logging-restrictions

max-arnold:
https://github.com/saltstack/salt/issues/57730

oeuftete:
[One workaround](https://help.saltstack.com/hc/en-us/articles/360044736252-Suppressing-appoptics-return-ERROR-messages-in-the-logs) for the master configuration:

```
log_granular_levels:
  'salt.loaded.int.returner.appoptics_return': 'critical'
```

garethgreenaway (changes requested):
As @max-arnold pointed out we shouldn't be logging in the __virtual__ function at all.  This should be removed completely.  The return of False will indicate why the module couldn't be loaded in the logs.

whytewolf:
noted. changed to nuke the logging in virtual.

garethgreenaway (dismissed):

salt/returners/appoptics_return.py
```diff
@@ -187,7 +177,7 @@ def _state_metrics(ret, options, tags):
 
     log.debug(
         "Batching with  Metric total states {}".format(
-            (stats["num_failed_states"] + stats["num_passed_states"])
+            stats["num_failed_states"] + stats["num_passed_states"]
```
s0undt3ch, line 180: It should instead read:
```python
    log.debug(
        "Batching with Metric total states %s",
        stats["num_failed_states"] + stats["num_passed_states"]
    )
```

s0undt3ch (changes requested):

salt/returners/appoptics_return.py
oeuftete: Is there a tool out there that can automatically update these via pre-commit?

This PR only touches this code because of the pre-commit, and the bug it fixes is really awful for anyone who needs to use the logs.  Please let's do what it takes to get this into the point release.

oeuftete: Enable http://pylint.pycqa.org/en/latest/technical_reference/features.html?highlight=w1202#logging-checker-messages, rather than manual review of these?

s0undt3ch: > Is there a tool out there that can automatically update these via pre-commit?

Unfortunately, no.

> Please let's do what it takes to get this into the point release.

It's not getting out of the point release. If you can't make the change, I'm happy to do it.

oeuftete: Done, fixed them all in this file. I'd forgotten Thomas had given me push access on his fork.

What do you think about [adding `W1202`](https://github.com/saltstack/salt/pull/57723#discussion_r517343286) so this doesn't have to be addressed arbitrarily in the future?

s0undt3ch: Unless we're explicitly disabling those messages, it's enabled by default and PyLint somehow missed them.

If they're disabled, yeah, we should enable them.

s0undt3ch (approved):

salt/returners/appoptics_return.py
oeuftete: They are disabled.

https://github.com/saltstack/salt/blob/a2f6e1bcf0a7728202f061c041e011d6bf553f7a/.pylintrc#L106

dwoz (approved):

PR #57726, opened 2020-06-18
Do not render grains twice if sentry is disabled

bobrik:
We're seeing costly grains being rendered twice due sentry handler, which unconditionally renders grains for own use. We're not using Sentry with salt, so it seems quite wasteful.


Akm0d:
Looks like this can be merged as soon as it has a test written for the changes

PR #57727, opened 2020-06-19
Strip brackets from IPv6 addresses

bobrik:
When connecting via tcp transport to an IPv6 address, salt uses `[a:b::c]` address format, when it should be using `a:b::c`. This silently makes salt client hang.


twangboy:
Could you write some tests for this

twangboy:
Looking at the conflict, this seems to be fixed on the master branch. If you believe this is still a problem please submit a new PR. Thank you.

Issue #57714, opened 2020-06-18
[BUG] `virtual` grains broken on illumos on 3001


sjorge:
**Description**
Somewhere between 3000.1 and 3001 the `virtual` grains got broken on illumos.
It only ever gets set to `physical` which is the default fallback if i am not mistaken.

**Setup**
Running salt 3001 inside a SmartOS zone.

**Steps to Reproduce the behavior**
```
[root@cronos ~]# /opt/salt3001/bin/salt-call grains.get virtual
local:
    physical
```

**Expected behavior**
```
[root@cronos ~]# /opt/salt3000/bin/salt-call grains.get virtual
local:
    zone
```

**Screenshots**
n/a

**Versions Report**
<details><summary>salt --versions-report</summary>
(Provided by running salt --versions-report. Please also mention any differences in master/minion versions.) 

```
[root@cronos ~]# /opt/salt/bin/salt-call --versions-report
Salt Version:
           Salt: 3001

Dependency Versions:
           cffi: 1.11.5
       cherrypy: unknown
       dateutil: Not Installed
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
         Jinja2: 2.11.2
        libgit2: Not Installed
       M2Crypto: 0.30.1
           Mako: Not Installed
   msgpack-pure: Not Installed
 msgpack-python: 0.6.2
   mysql-python: Not Installed
      pycparser: 2.19
       pycrypto: 2.6.1
   pycryptodome: 3.9.7
         pygit2: Not Installed
         Python: 3.6.6 (default, Oct 23 2018, 17:13:47)
   python-gnupg: Not Installed
         PyYAML: 5.3.1
          PyZMQ: 17.1.2
          smmap: Not Installed
        timelib: Not Installed
        Tornado: 4.5.3
            ZMQ: 4.2.5

System Versions:
           dist: pkgsrc 3
         locale: UTF-8
        machine: i86pc
        release: 5.11
         system: SunOS
        version: release: 2018Q 3
```
</details>

No master involved.

**Additional context**
I tried tracing it back but the diff between the too was rather confusing, there were some changes for Solaris LDOM which might have effected this, but AFAIK they don't.

I did do some debugging with pudb and it seems we now ever make it to  https://github.com/saltstack/salt/blob/b95213ec903402f25c1e0aeb3990fe8452ab63ce/salt/grains/core.py#L912 which we did in the past.

sjorge:
[...]

PR #57728, opened 2020-06-19
[Fix] #57714 virtual grain should work on illumos

sjorge:
### What does this PR do?
This fixes the `virtual` grain on illumos that got broken between 3000.1 and 3001, I didn't have time to hunt down the actually commit but after about an hour of analysis.

The loop over _cmds was slightly modified to abort differently if a command had a non-zero exit. On illumos/solaris `prtdiag` has a non-zero exit in the zones. Before we would avoid addning it to the failed_commands list but still continued. 

In the block dealing with prtdiag output we had a fallback that would do some other checks to figure out if we were a global zone(=physical) or zone (=zone).

This PR moves the fallback case outside of the loop, which is a better location for it. So we avoid the problem.

This also fixed resolved my other issues I saw when testing 3001 on SmartOS, which were because some modules depend on the virtual grain.

Fixes:
- #57714 

### Previous Behavior
Fallback case was skipped.

### New Behavior
Fileback case gets executes, resulting in the correct (pre 3001) virtual grain.

- [ ] Docs
- [x] Tests written/updated

### Commits signed with GPG?
No

### DIFF
--- a/changelog/57714.fixed
+++ b/changelog/57714.fixed
@@ -1,0 +1,1 @@
+Due to some optimization the `virtual` grain was never updated on illumos. Move the fallback in prtdiag output parsing outside the loop that now gets skipped due to the command exiting non-zero.
--- a/salt/grains/core.py
+++ b/salt/grains/core.py
@@ -919,16 +919,6 @@ def _virtual(osdata):
                 grains["virtual"] = "kvm"
             elif "joyent smartdc hvm" in model:
                 grains["virtual"] = "kvm"
-            else:
-                # Check if it's a "regular" zone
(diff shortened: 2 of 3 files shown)


sjorge:
As mentioned in #57714, I have a pretty busy few weeks ahead so not sure I will have the time to write the tests. But I was asked to open the PR with my analysis and fix anyway.

sjorge:
Still super busy with the move, hope to be back in action around august. But a lot of people are complaining about this to me on IRC.

I'm not really happy with the tests, but they do seem to work.

dwoz (approved):

PR #57519, opened 2020-06-02
Fixed ec2 cloud unit test.

DmitryKuzmenko:
### What does this PR do?
I've found these tests failures in PR #56565
```
Traceback (most recent call last):
  File "/tmp/kitchen/testing/.nox/runtests-parametrized-3-coverage-true-crypto-none-transport-zeromq/lib/python3.8/site-packages/mock/mock.py", line 1330, in patched
    return func(*args, **keywargs)
  File "/tmp/kitchen/testing/tests/unit/cloud/clouds/test_ec2.py", line 90, in test_get_password_data
    ret = ec2.get_password_data(
  File "/tmp/kitchen/testing/salt/cloud/clouds/ec2.py", line 4970, in get_password_data
    key_obj = Crypto.PublicKey.RSA.importKey(rsa_key)
AttributeError: module 'Crypto' has no attribute 'PublicKey'
```
https://jenkinsci.saltstack.com/job/pr-ubuntu2004-py3/job/PR-56565/2/testReport/junit/unit.cloud.clouds.test_ec2/EC2TestCase/test_get_password_data/
https://jenkinsci.saltstack.com/job/pr-fedora32-py3/job/PR-56565/2/testReport/junit/unit.cloud.clouds.test_ec2/EC2TestCase/test_get_password_data/

I don't know why the test isn't failing in other builds but if I can easily reproduce it using python interpreter:
```
$ python
Python 3.7.7 (default, Apr 13 2020, 14:31:36) 
[GCC 9.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import Crypto
>>> Crypto.PublicKey
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: module 'Crypto' has no attribute 'PublicKey'
>>> Crypto.__version__
'3.9.7'
```

I've moved the crypto related logic to `salt.crypt` package. Also fixed the code to work with Cryptodome.

### What issues does this PR fix or reference?
Fixes test failures for #56565

### Merge requirements satisfied?
**[NOTICE] Bug fixes or features added to Salt require tests.**

- [ ] Docs
- [ ] Changelog - https://docs.saltstack.com/en/master/topics/development/changelog.html
- [x] Tests written/updated

### Commits signed with GPG?
Yes

PR #57763, opened 2020-06-23
Fix import of Crypto in salt.cloud.clouds.ec2

bdrung:
When running the salt tests against the installed salt module (with `Crypto`
available, but not `M2Crypto`), `test_get_password_data` failed:

```
ERROR: test_get_password_data (unit.cloud.clouds.test_ec2.EC2TestCase)
[CPU:0.0%|MEM:29.4%]
  File "/usr/lib/python3.8/unittest/mock.py", line 1325, in patched
    return func(*newargs, **newkeywargs)
  File "tests/unit/cloud/clouds/test_ec2.py", line 90, in test_get_password_data
    ret = ec2.get_password_data(
  File "/usr/lib/python3/dist-packages/salt/cloud/clouds/ec2.py", line 4970, in get_password_data
    key_obj = Crypto.PublicKey.RSA.importKey(rsa_key)
AttributeError: module 'Crypto.PublicKey' has no attribute 'RSA'
```

When importing `Crypto`, only `Crypto.version_info` was available. Therefore
import the needed submodules of `Crypto` explicitly. Split
`test_get_password_data` into two tests: One to test the code path for
`M2Crypto` and one for `Crypto`.

### DIFF
--- a/salt/cloud/clouds/ec2.py
+++ b/salt/cloud/clouds/ec2.py
@@ -138,7 +138,9 @@ except ImportError:
     HAS_M2 = False
 
 try:
-    import Crypto
+    import Crypto.Hash.SHA
+    import Crypto.PublicKey.RSA
+    import Crypto.Random
 
     # PKCS1_v1_5 was added in PyCrypto 2.5
     from Crypto.Cipher import PKCS1_v1_5  # pylint: disable=E0611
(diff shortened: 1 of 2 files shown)


DmitryKuzmenko:
@bdrung  #57519 fixes the same issue. I'd better like my approach moving the crypt related logic to `salt/crypt.py`. But I like your test.
What do you think about it?

bdrung:
@DmitryKuzmenko  I don't care which solution is taken. Feel free to combine your pull requests with my test changes (adapted to the function move).

DmitryKuzmenko:
Will do, thank you!

Akm0d (approved):

DmitryKuzmenko:
Just got it: we don't need to split m2 and pycrypto tests because our CI has different runs with pycrypto, pycryptodomex and m2.
Anyway thank you for your work.

PR #57770, opened 2020-06-23
modules.postgres: pass waldir to init

### LINKED ISSUE #57769: [BUG] postgres.datadir_init parameter "waldir" nonfunctional
**Description**
postgres.datadir_init does not create an external WAL directory when requested via the "waldir" parameter - the "waldir" parameter is not passed through to _run_initdb(): https://github.com/saltstack/salt/blob/master/salt/modules/postgres.py#L3596

**Setup**
no special setup neccessary

look at the code and it's obvious

**Expected behavior**
pass the given parameters

**Screenshots**

**Versions Report**
Support for "waldir" and "checksums" parameters to datadir_init() was added in 2019.2 but both parameters were not passed through to _run_initdb() (which makes me wonder if I screwed up back then or if there was some merge disaster?). The "checksums" parameter has been fixed in commit 35e2cacac6e4dfe21208434b2dea295c62561e9e but "waldir" has once again been missed, so here we go.
```
PASTE HERE
```
</details>

**Additional context**

moench-tegeder:
Since 2019.2.0 postgres.datadir_init() takes a 'waldir' keyword
argument to specify a location for an external wal directory -
but this parameter is not passed to _run_initdb (l. 3596),
rendering 'waldir' effectively nonfunctional.
The fix is obvious - pass the parameter. The unit test is
being modified to at least catch 'waldir' set to None (unset).
Testing the generated initdb command line or even final data
directory properties would be desirable, but seems to be much
more invasive and thus not that suitable for a "small bugfix".

Fixes: #57769


moench-tegeder:
do I need to do anything to get this backpatched to older (supported) releases?

moench-tegeder:
Oh cool, I never got the mail with the changes.
Anyways, both files test were infra failures:
```
09:21:11  ERROR: Maximum checkout retry attempts reached, aborting
09:21:11  Finished: FAILURE
```
And whatever that gitlab pipeline is doing.
Let's pull this up to speed and try again.

garethgreenaway (approved):

Ch3LL (approved):

PR #57772, opened 2020-06-23
Fix cascading KeyError in rh_ip

ixs:
### What does this PR do?
Fix a cascading KeyError exception

### Intended behavior
If configuring an IPIP interface, two options are mandatory.
If one option is missing, the _raise_error_iface() function is called.

This function should raise an AttributeError: Invalid option.

### Previous behavior
With the current code, what will happen though is that the error is
raised, the opts dictionary is searched for the missing key which in
turn raises another KeyError during the handling of the first exception.

The intention was to print the key (opt) and not the value (opts['opt']).

### New behavior
Fix this, new error will correctly print the AttributeError


ixs:
Change got folded into #57775.  Closing

PR #57776, opened 2020-06-24
Add teamed network interface support to rh_ip.py

ixs:
### What does this PR do?
Salt currently allows to configure bonded network interfaces for
redundancy.

With RHEL7 RedHat has added an alterantive approach called teamd.
The website for this project is at <http://libteam.org>.
Teamd is a userspace daemon that controls the trunked interfaces
via the netlink layer allowing for more flexibility than the
bonding module.

Add support for teamd configuration to salt rh_ip.

### DIFF
--- a/salt/modules/rh_ip.py
+++ b/salt/modules/rh_ip.py
@@ -112,6 +112,8 @@ _IFACE_TYPES = [
     "vlan",
     "ipip",
     "ib",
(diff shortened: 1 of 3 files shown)


ixs:
Can probably be closed as #57775 seems to do the same change but also improves bonding support.

PR #57777, opened 2020-06-24
Update Winrepo Docs

### LINKED ISSUE #32780: win_pkg: msiexec parameter does not allow MSP patch files to be applied
When creating a package definition, setting "msiexec: True" forces the msiexec option "/i" for installs, or "/x" for uninstalls. This prevents the use of the "/p" which is used to apply a patch to an already installed MSI package. A patch may increase the version number and so is potentially equivalent to a package upgrade. 
- "msiexec /i /p" is not a valid combination of switches so you cannot override "/i" by setting "/p" in the install_flag string
- setting "msiexec: False" and using "msiexec.exe" as the installer will result in cache_path not being set so you cannot reference a cached patch file in the install_flags string. It also means the cache_dir parameter cannot be used (since there is no inferred salt:// installer path)
- setting "msiexec: True" and applying a patch with "/update" or "PATCH=" requires a fully qualified path to the patch files, which isn't available as a core grain or environment variable

As a workaround, without resorting to hard-coding paths or running batch files, you can glue together the fully qualified location of the minion cache and reference that to apply a patch:

```
{% set minion_cache_path = (opts['cachedir'] ~ '\\files\\' ~ saltenv ~ '\\win\\repo-ng\\my_app') | replace('\\','\\\\') %}

my_app:
  '5.5.3':
    full_name: 'My App'
    installer: 'salt://win/repo-ng/my_app/MyApp.msi'
    install_flags: 'ALLUSERS=2 /update {{ minion_cache_path }}\\MyApp553.msp /qn /norestart'
    uninstaller: '{C7AD892D-4D6C-4444-B6D7-30AFA6147466}'
    uninstall_flags: '/qn /norestart'
    msiexec: True
    cache_dir: True
```

...but this doesn't look too pretty when all that was required was to run `msiexec /p MyApp553.msp /qn /norestart`. It also results in the whole directory being re-hashed since both the original installer and the patch file must be referenced.

I would like to suggest the addition of an extra package definition parameter, which when set would imply the msiexec option "/p" in-place of option "/i", e.g. "msiexec_patch: True".

Tested on Salt version 2015.8.8.2

twangboy:
### What does this PR do?
Organize the document.
Provide additional examples to cover additional scenarios.
Clear out all the Legacy stuff and put it at the end as it just muddled things up,

Fixes: https://github.com/saltstack/salt/issues/32780

- [x] Docs
- [ ] Tests written/updated

### Commits signed with GPG?
Yes


Akm0d (dismissed):

doc/ref/configuration/minion.rst
```diff
@@ -2713,6 +2713,34 @@ the ``extra_minion_data`` parameter will be
...
+      module.run:
+        - name: state.apply
+        - mods:
+          - showpillar
+        - kwargs:
+              pillar:
+                  test: "foo bar"
+
+If set to ``True`` the ``showpillar`` state will have access to the
+global pillar data.
```
barbaricyawps, line 2739: Totally minor grammar point here, but you need to insert a command after True: If set to ``True``, the ``showpillar`` state will...

doc/ref/configuration/minion.rst
```diff
@@ -2713,6 +2713,34 @@ the ``extra_minion_data`` parameter will be
...
+
+If set to ``False`` only the overriding pillar data will be available
+to the ``showpillar`` state.
+
```
barbaricyawps, line 2743: Same grammar point here as well: If set to ``False``, only the overriding pillar data will be available...

doc/ref/configuration/minion.rst
```diff
@@ -3413,6 +3441,13 @@ have other services that need to go with it.
 
     update_restart_services: ['salt-minion']
 
+.. _winrepo-minion-config-opts:
+
+Windows Software Repo Settings
+==============================
+
+These settings apply to all minions whether in masterless or master-minion mode.
+
```
barbaricyawps, line 3450: Maybe instead: "These settings apply to all minions, whether running in masterless or master-minion mode."

doc/ref/configuration/minion.rst
```diff
@@ -3475,9 +3500,13 @@ The source location for the winrepo sls files.
 Standalone Minion Windows Software Repo Settings
 ================================================
 
+The following settings are for configuring the Windows Software Repository
+(winrepo) on a masterless minion. To run in masterless minion mode set the
+:conf_minion:`file_client` to ``local`` or run ``salt-call`` with the
+``--local`` option
+
```
barbaricyawps, line 3507: Same need for a comma to set off your introductory phrase: To run in masterless minion mode, set the..."

salt/modules/win_pkg.py
```diff
@@ -2215,9 +2215,14 @@ def remove(name=None, pkgs=None, **kwargs):
...
     """
-    Package purges are not supported, this function is identical to
+    Package purges are not supported on Windows, this function is identical to
     ``remove()``.
 
+    .. note::
+        At some point in the future ``pkg.purge`` may direct the installer to
+        remove all configs and settings for software packages that support that
+        option.
+
```
barbaricyawps, line 2225: Just another need for a comma after your introductory phrase again: "At some point in the future, ``pkg.purge`` may direct the installer..."

barbaricyawps (dismissed):
I've finished my review! Nice work on these docs! I've made a few grammar and phrasing suggestions that you can choose to adopt. Since they're minor, I'll go ahead and approve this since there won't be a need for me to review it a second time. Keep up the great work!

salt/config/__init__.py
```diff
@@ -1173,7 +1173,7 @@ def _gather_buffer_space():
             salt.syspaths.BASE_FILE_ROOTS_DIR, "win", "repo-ng"
         ),
         "winrepo_cachefile": "winrepo.p",
-        "winrepo_cache_expire_max": 21600,
+        "winrepo_cache_expire_max": 604800,
```
damon-atkins, line 1176: This pull request title was about updating Docs, not changing default values.  As the author of the caching code this should not have been changed..... from the default of 6 hours, which is less than a business day to 7 days.  It will force people people to run refresh_db, even if they have run an high-state.  FYI @twangboy

damon-atkins: Also need to consider some devices will not be running when one one needs to run refresh_db. It's easy for people to roll out a config file to the minion to change this setting.   The impact of this is Clients will not see new versions of software, and may install the wrong software.  FYI @dwoz

damon-atkins: FYI @saltstack/team-windows

damon-atkins: Mmm I suspect this para needs fixing. @saltstack/team-windows 
Software package definitions are automatically refreshed if stale after winrepo_cache_expire_max. Running a highstate normal forces the refresh of the package definition and generation of the metadata, unless the metadata is younger than winrepo_cache_expire**_min**.

With winrepo_cache_expire_max set to 7 days, people will get inconsistent times. Lets say you run `salt * pkg.list_pkgs`  half the minions might return in under 20 seconds (<7d) and the other half return in under 2 minutes (>7d).  An hour letter another minion could it the >7d.   The end result is the first run of  `salt * pkg.list_pkgs`  will still take 2 minutes, as its the slowest minion that holds up the results. And for the rest of the day if your unlucky you might get more minions go past the 7 day mark.    And if one is old than 7d, people will think why is this slow now?

Set to 6 hours the timing of the results is consistent and the output of the results is consistent.  Next time its run it will be quicker for all minions.   The cache was design so people running the commands had consistent results, and if they ran it multiple times within 6 hour period the results would come back faster (we are assuming they have not change the package definition files for 6 hour period).

The their the impact to `list_available` which could return different results.

doc/ref/configuration/minion.rst
```diff
@@ -3506,10 +3537,11 @@ out.
 .. versionadded:: 2015.8.0
     A new :ref:`ng <windows-package-manager>` repo was added.
 
-Default: ``/srv/salt/win/repo-ng``
+Default: ``C:\salt\srv\salt\win\repo-ng``
```
damon-atkins, line 3540: Is this change correct?

PR #57779, opened 2020-06-24
info_installed works without status attr now

brejoc:
### What does this PR do?

If `status` was excluded via `attr`, `info_installed` was no longer able to detect if a package was installed or not. Now `info_installed` adds the `status` for the 'lowpkg.info' request again.

### Previous Behavior

`info_installed` would also return uninstalled packages if `status` wasn't in the `attr` list.

### New Behavior

`info_installed` only returns installed packages.

- [ ] Docs
- [x] Tests written/updated

### Commits signed with GPG?
Yes


garethgreenaway:
@brejoc Looks like your new test is failing.  Also can we get a changelog file for this change?  Thanks!

PR #57803, opened 2020-06-25
Add umask as a global state argument

terminalmage:
This removes the umask argument from cmd states and makes it a global state argument.

### DIFF
--- a/changelog/57803.changed
+++ b/changelog/57803.changed
@@ -1,0 +1,2 @@
+``umask`` is now a global state argument, instead of only applying to ``cmd``
+states.
--- a/doc/ref/states/requisites.rst
+++ b/doc/ref/states/requisites.rst
@@ -1269,3 +1269,22 @@ times with a 2 second interval, but the file specified did not exist on any run)
      Started: 09:08:12.903000
     Duration: 47000.0 ms
      Changes:
+
+Run State With a Different Umask
+================================
+
+.. versionadded:: 3002
+   NOTE: not available on Windows
+
+The ``umask`` state argument can be used to run a state with a different umask.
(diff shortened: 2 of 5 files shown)


garethgreenaway:
@terminalmage Looks good.  My only comment is on the change in doc/ref/states/requisites.rst, why this is temporary?  Also we need a changelog file added and the pre-commit test is failing.

terminalmage:
It's temporary because the umask is only changed for the lifetime of a single state.

Perhaps the wording can be improved?

garethgreenaway:
@terminalmage Thanks.  That makes more sense.  My first impression was that it was being changed for just the 3002 release and would change to a different approach in the future.

terminalmage:
@garethgreenaway Cool. Should this just be considered an "added" changelog entry even though `umask` argument has been removed from the "cmd" state module here? The functionality will remain the same.

terminalmage:
Also, can the changelog files contain markdown? This is not clear from the docs.

garethgreenaway:
@terminalmage I would make it an "changed" change since you're remove the `umask` argument from `cmd` in favor of it using the global one.  It looks like town crier does have support for using markdown.

terminalmage:
OK, thanks. Changelog file added.

garethgreenaway (dismissed):

Ch3LL:
@terminalmage mind resolving the merge conflicts?

dwoz (approved):

MKLeb (approved):

terminalmage:
Merge conflicts removed, tests re-done and passing

max-arnold:
The 3002 version number in the docs need to be changed to 3006: https://docs.saltproject.io/en/master/ref/states/requisites.html#run-state-with-a-different-umask

PR #57814, opened 2020-06-26
Support inter-region peering in boto_vpc.

JamesSnow:
### What does this PR do?
Existing boto_vpc.request_vpc_peering_connection doesn't handle
inter-region VPC peering. This enables inter-region peering but
assumes intra-region peering unless peer_region is specified.

Fixes: n/a

### Previous Behavior
Attempting to issue a peering request would succeed as a salt state, but fail at AWS due to the peer vpc_id not existing in the region the peering request is issued from.

### New Behavior
Inter-region VPC peering requests work now. Yay!

### DIFF
--- a/salt/modules/boto_vpc.py
+++ b/salt/modules/boto_vpc.py
@@ -3713,6 +3713,7 @@ def request_vpc_peering_connection(
     peer_vpc_name=None,
     name=None,
     peer_owner_id=None,
+    peer_region=None,
     region=None,
     key=None,
     keyid=None,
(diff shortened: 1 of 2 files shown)


waynew (changes requested):
Hey @JamesSnow thanks for the PR! We're requiring tests before merging PRs, there might already be some existing boto tests that can be updated for these cases.

This would also need a [changelog entry](https://docs.saltstack.com/en/master/topics/development/changelog.html).

JamesSnow:
Hi Wayne,

Happy to create tests. Since these are AWS calls, does Saltstack have a preferred mock framework?

I have also discovered that the vpc peering accept code needs some reworking to support multi-region. Have had to move onto some other things for now but will circle back to this and tests soonish.

dwoz:

PR #57841, opened 2020-06-30
Only run sysctl ignore when configured

### LINKED ISSUE #57826: [BUG] sysctl.present broken in FreeBSD
```
          ID: security.jail.sysvipc_allowed
    Function: sysctl.present
      Result: False
                File "/usr/local/lib/python3.7/site-packages/salt/state.py", line 2154, in call
                File "/usr/local/lib/python3.7/site-packages/salt/loader.py", line 2085, in wrapper
                  return f(*args, **kwargs)
                File "/usr/local/lib/python3.7/site-packages/salt/states/sysctl.py", line 117, in present
                  update = __salt__["sysctl.persist"](name, value, config, ignore)
              TypeError: persist() takes from 2 to 3 positional arguments but 4 were given
```

Looks like it was broken by this https://github.com/saltstack/salt/pull/55719 since the author of that commit didn't do any testing or review on any platform other than linux.

mchugh19:
### What does this PR do?
The ignore option added to linux sysctl in #55719 was done where it assumed the feature had been added to all OS's using sysctl. This broke sysctl state handling for non linux hosts. This PR corrects this behavior and only passes the ignore option to sysctl execution modules when it has been configured.

Fixes: #57826

### Previous Behavior
The sysctl state would always pass an ignore option to the sysctl execution module

### New Behavior
The sysctl state now only passes the ignore option to the sysctl execution module when it has been configured to do so

- [ ] Docs
- [ ] Tests written/updated - The proper test for this is to run the test suite on non-linux OSs as mentioned https://github.com/saltstack/salt/issues/57826#issuecomment-650747404

### Commits signed with GPG?
No


krionbsd (dismissed):

ari:
I've added some comments to the related bug issue #57826 which might be relevant before this is merged.

sagetherage:
this needs a regression test @krionbsd @mchugh19

mchugh19:
How would you all like a regression test? The logic really boils down to
```
if ignore:
    # ignore is a linux only sysctl setting
    update = __salt__["sysctl.persist"](name, value, config, ignore)
else:
    update = __salt__["sysctl.persist"](name, value, config)
```

Since freebsd and macosx have their own sysctl modules, but don't support the ignore value, this PR attempts to only pass the ignore value if it is defined in the state. 

I see having the test suite run on freebsd and osx as the proper regression test for this. Is there anything else the coreteam thinks makes sense?

mchugh19:
@dwoz any feedback here? This PR is to fix that the linux sysctl execution module supports an additional `ignore` parameter which isn't present in the freebsd and osx versions. How would you like this tested? Or would it be easier to add ignore as a noop to the other modules?

salt/states/sysctl.py
```diff
@@ -102,7 +103,11 @@ def present(name, value, config=None):
         return ret
 
     try:
-        update = __salt__["sysctl.persist"](name, value, config)
+        if ignore:
+            # ignore is a linux only sysctl setting
+            update = __salt__["sysctl.persist"](name, value, config, ignore)
```
twangboy, line 108: The `sysctl.persist` function in `salt/modules/linux_sysctl.py` does not have an `ignore` option. None of the `sysctl.persist` functions have an `ignore` option.

If you want to support ignore on Linux, it will need to be added.

twangboy (changes requested):

salt/states/sysctl.py
mchugh19: It used to ;) #55719

If PRs are being reviewed again, I can take a look at re-adding.

dwoz:

PR #57859, opened 2020-07-02
Fix macOS BigSur

### LINKED ISSUE #57787: [BUG][MacOS] Salt errors on macOS Big Sur
**Description**
When attempting to run salt on macOS Big Sur the following error is produced.

```
  File "/opt/local/bin/salt-call", line 11, in <module>
    load_entry_point('salt==3001', 'console_scripts', 'salt-call')()
  File "/opt/salt/lib/python3.7/site-packages/salt-3001-py3.7.egg/salt/scripts.py", line 466, in salt_call
    import salt.cli.call
  File "/opt/salt/lib/python3.7/site-packages/salt-3001-py3.7.egg/salt/cli/call.py", line 6, in <module>
    import salt.cli.caller
  File "/opt/salt/lib/python3.7/site-packages/salt-3001-py3.7.egg/salt/cli/caller.py", line 19, in <module>
    import salt.minion
  File "/opt/salt/lib/python3.7/site-packages/salt-3001-py3.7.egg/salt/minion.py", line 30, in <module>
    import salt.crypt
  File "/opt/salt/lib/python3.7/site-packages/salt-3001-py3.7.egg/salt/crypt.py", line 39, in <module>
    import salt.utils.rsax931
  File "/opt/salt/lib/python3.7/site-packages/salt-3001-py3.7.egg/salt/utils/rsax931.py", line 131, in <module>
    libcrypto = _init_libcrypto()
  File "/opt/salt/lib/python3.7/site-packages/salt-3001-py3.7.egg/salt/utils/rsax931.py", line 84, in _init_libcrypto
    libcrypto = _load_libcrypto()
  File "/opt/salt/lib/python3.7/site-packages/salt-3001-py3.7.egg/salt/utils/rsax931.py", line 77, in _load_libcrypto
    return cdll.LoadLibrary(_find_libcrypto())
  File "/opt/salt/lib/python3.7/site-packages/salt-3001-py3.7.egg/salt/utils/rsax931.py", line 69, in _find_libcrypto
    raise OSError("Cannot locate OpenSSL libcrypto")
OSError: Cannot locate OpenSSL libcrypto
```

This is caused by a change in how dynamic libraries are made available on macOS. [More info in the snippet bellow from the Big Sur Release Notes.](https://developer.apple.com/documentation/macos-release-notes/macos-big-sur-11-beta-release-notes)

```
New in macOS Big Sur 11 beta, the system ships with a built-in dynamic linker cache of all system-provided libraries.
As part of this change, copies of dynamic libraries are no longer present on the filesystem.
Code that attempts to check for dynamic library presence by looking for a file at a path or enumerating a directory will fail.
Instead, check for library presence by attempting to dlopen() the path, which will correctly check for the library in the cache. (62986286)
```

The way salt currently works is it's looking for the path of the `libcrypto.dylib` which is missing on Big Sur. Which is why we fail.

My recommendation is to no longer rely on a system provided OpenSSL in `salt/utils/rsax931.py` and move to using the OpenSSL that is provided by the salt installer in #57607. (We can also add needed paths for Homebrew and Macports installations) and only fall back to the System as the last resort. 

**Setup**
A system running macOS Big Sur

run `sudo salt-call --version`

**version report**
After applying work around.
```
Salt Version:
           Salt: 3001
 
Dependency Versions:
           cffi: 1.12.2
       cherrypy: unknown
       dateutil: 2.8.0
      docker-py: Not Installed
          gitdb: 2.0.6
      gitpython: 2.1.15
         Jinja2: 2.10.1
        libgit2: Not Installed
       M2Crypto: Not Installed
           Mako: 1.0.7
   msgpack-pure: Not Installed
 msgpack-python: 0.5.6
   mysql-python: Not Installed
      pycparser: 2.19
       pycrypto: Not Installed
   pycryptodome: 3.9.7
         pygit2: Not Installed
         Python: 3.7.4 (default, Jun  9 2020, 17:22:20)
   python-gnupg: 0.4.4
         PyYAML: 5.1.2
          PyZMQ: 18.0.1
          smmap: 3.0.4
        timelib: 0.2.4
        Tornado: 4.5.3
            ZMQ: 4.3.1
 
System Versions:
           dist: darwin 20.0.0 
         locale: UTF-8
        machine: x86_64
        release: 20.0.0
         system: Darwin
        version: 10.16 x86_64
```

weswhet:
### What does this PR do?
Adds basic support for macOS Big Sur. 

Fixes #57787 

### Previous Behavior
Salt would crash on Big Sur and get 😭. See the issue above for more details.
 
### New Behavior
salt + Big Sur = ❤️ 

- [x] Docs
- [x] Tests written/updated

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/utils/rsax931.py
+++ b/salt/utils/rsax931.py
@@ -15,6 +15,7 @@ import sys
 from ctypes import c_char_p, c_int, c_void_p, cdll, create_string_buffer, pointer
 
 # Import Salt libs
+import salt.utils.mac_utils
 import salt.utils.platform
 import salt.utils.stringutils
 
@@ -30,6 +31,24 @@ def _find_libcrypto():
     """
     if sys.platform.startswith("win"):
         lib = str("libeay32")
+    elif salt.utils.platform.is_darwin():
+        # will look for several different location on the system,
+        # Search in the following order. salts pkg, homebrew, macports, finnally
+        # system.
+        # look in salts pkg install location.
+        lib = glob.glob("/opt/salt/lib/libcrypto.dylib")
+        # Find library symlinks in Homebrew locations.
+        lib = lib or glob.glob("/usr/local/opt/openssl/lib/libcrypto.dylib")
+        lib = lib or glob.glob("/usr/local/opt/openssl@*/lib/libcrypto.dylib")
+        # look in macports.
+        lib = lib or glob.glob("/opt/local/lib/libcrypto.dylib")
+        # check if 10.15, regular libcrypto.dylib is just a false pointer.
+        if salt.utils.mac_utils.os_version(as_tuple=True) == (10, 15):
+            lib = lib or glob.glob("/usr/lib/libcrypto.*.dylib")
+            lib = list(reversed(sorted(lib)))
+        # last but not least all the other macOS versions should work here.
+        # including Big Sur
+        lib = lib[0] if lib else str("/usr/lib/libcrypto.dylib")
     elif getattr(sys, "frozen", False) and salt.utils.platform.is_smartos():
         lib = glob.glob(os.path.join(os.path.dirname(sys.executable), "libcrypto.so*"))
         lib = lib[0] if lib else None
@@ -53,18 +72,6 @@ def _find_libcrypto():
                 else:
                     lib = glob.glob("/opt/freeware/lib/libcrypto.so*")
                 lib = lib[0] if lib else None
-        elif salt.utils.platform.is_darwin():
-            # Find versioned libraries in system locations, being careful
-            # to avoid the unversioned stub which is no longer permitted.
-            lib = glob.glob("/usr/lib/libcrypto.*.dylib")
-            if lib:
-                # Sort so as to prefer the newest version.
-                lib = list(reversed(sorted(lib)))
-            else:
-                # Find library symlinks in Homebrew locations.
-                lib = glob.glob("/usr/local/opt/openssl/lib/libcrypto.dylib")
-                lib = lib or glob.glob("/usr/local/opt/openssl@*/lib/libcrypto.dylib")
-            lib = lib[0] if lib else None
     if not lib:
         raise OSError("Cannot locate OpenSSL libcrypto")
     return lib
--- a/salt/utils/rsax931.py
+++ b/salt/utils/rsax931.py
@@ -31,7 +31,7 @@ def _find_libcrypto():
     """
     if sys.platform.startswith("win"):
         # cdll.LoadLibrary on windows requires a 'str' argument
-        lib = str("libeay32")
+        lib = "libeay32"
     elif salt.utils.platform.is_darwin():
         # will look for several different location on the system,
         # Search in the following order. salts pkg, homebrew, macports, finnally
--- a/salt/utils/rsax931.py
+++ b/salt/utils/rsax931.py
@@ -9,13 +9,13 @@ from __future__ import absolute_import, print_function, unicode_literals
 import ctypes.util
 import glob
 import os
+import platform
 import sys
 
 # Import 3rd-party libs
 from ctypes import c_char_p, c_int, c_void_p, cdll, create_string_buffer, pointer
 
 # Import Salt libs
-import salt.utils.mac_utils
 import salt.utils.platform
 import salt.utils.stringutils
 
@@ -43,7 +43,7 @@ def _find_libcrypto():
         # look in macports.
         lib = lib or glob.glob("/opt/local/lib/libcrypto.dylib")
         # check if 10.15, regular libcrypto.dylib is just a false pointer.
-        if salt.utils.mac_utils.os_version(as_tuple=True) == (10, 15):
+        if platform.mac_ver()[0].split(".")[:2] == ["10", "15"]:
             lib = lib or glob.glob("/usr/lib/libcrypto.*.dylib")
             lib = list(reversed(sorted(lib)))
         # last but not least all the other macOS versions should work here.
--- a/salt/utils/rsax931.py
+++ b/salt/utils/rsax931.py
@@ -30,7 +30,6 @@ def _find_libcrypto():
     Find the path (or return the short name) of libcrypto.
     """
     if sys.platform.startswith("win"):
-        # cdll.LoadLibrary on windows requires a 'str' argument
         lib = "libeay32"
     elif salt.utils.platform.is_darwin():
         # will look for several different location on the system,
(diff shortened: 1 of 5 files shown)


weswhet:
@kaorihinata mind taking a look. I tried to clean it up into a single `if darwin` statement.

@sagetherage Can we get this into 3001.1 so more folks can test Big Sur?

weswhet:
Pairs well with #57607 but not required.

weswhet:
re-run all

kaorihinata:
`os_version(...)` doesn't appear to be necessary. Is there a reason you'd like to add another function and tests when you only use 1 form of it?

Something like:

`platform.mac_ver()[0].startswith('10.15')`
(or `'10.15.'` if `mac_ver()` includes the 0.)

...or:

`platform.mac_ver()[0].split('.')[:2] == ['10', '15']`
(Python itself uses this in places now that I check.)

...would also suffice, aye? I don't see `os_version(...)` as enough of a departure from the functionality `platform` already provides for us to need another method of mangling the version.

kaorihinata:
@weswhet Now that I've had time to go over the build logs, it seems like relying on `salt.utils.mac_utils` to get to `os_version` is running into issues when rolled into a thin tarball (possibly a circular import involving `salt.modules.cmdmod`.) I can reproduce the issue by building a thin tarball, then attempting to import `salt.modules.cmdmod`. Any ideas? Alternatively, you could probably just drop `os_version` and use a 1 liner.

sagetherage:
@weswhet 
> Can we get this into 3001.1 so more folks can test Big Sur?

we need to get it passing tests here, it may need some fixes to get merged to do that, I will review today.

tests/unit/utils/test_rsax931.py
```diff
@@ -174,25 +174,55 @@ def test_find_libcrypto_aix(self):
...
             "/usr/local/opt/openssl@*/lib/libcrypto.dylib",
+            "/opt/local/lib/libcrypto.dylib",
+            "/usr/lib/libcrypto.*.dylib",
         ):
             if fnmatch.fnmatch(lib_path, i):
                 passed = True
                 break
         self.assertTrue(passed)
+        # we should never return the stub on catalina
+        self.assertNotEqual(lib_path, "/usr/lib/libcrypto.dylib")
```
kaorihinata, line 220: If it ever did return `"/usr/lib/libcrypto.dylib"` on Catalina, then you would have already crashed at this point because both a full and partial (no slow) test run automatically starts minions/master instances. Additionally, even if by some bizarre twist it made it through, and you were (for example) running the tests alone, the `assertTrue(...)` will have already failed the test. What does this call to `assertNotEqual(...)` achieve at this point?

kaorihinata:
Remove the dependency on `salt.utils.mac_utils` and you fix `unit.utils.test_thin.SSHThinTestCase.test_thin_dir`. This may also fix `integration.ssh.test_deploy.SSHTest.test_set_path` as a side effect.

tests/unit/utils/test_rsax931.py
weswhet: good catch! I've updated the PR. :)

weswhet:
> Remove the dependency on `salt.utils.mac_utils` and you fix `unit.utils.test_thin.SSHThinTestCase.test_thin_dir`. This may also fix `integration.ssh.test_deploy.SSHTest.test_set_path` as a side effect.

Done, I added the mac_utils os piece because I have some other code I was going to PR but I've scratched it for this case. Thanks for the review/feedback. Appreciate it!

kaorihinata:
`os_version` is fine if it's going to be used more widely in the future, but we'll need to find somewhere to put it that doesn't require importer/loader magic to make it work. Also looks like we're already going to need to re-run for one of the Windows hosts.

weswhet:
rerun windows

weswhet:
re-run windows2016

kaorihinata:
re-run opensuse15

krionbsd (dismissed):

salt/utils/rsax931.py
```diff
@@ -30,6 +31,24 @@ def _find_libcrypto():
...
+        lib = lib or glob.glob("/usr/local/opt/openssl@*/lib/libcrypto.dylib")
+        # look in macports.
+        lib = lib or glob.glob("/opt/local/lib/libcrypto.dylib")
+        # check if 10.15, regular libcrypto.dylib is just a false pointer.
+        if platform.mac_ver()[0].split(".")[:2] == ["10", "15"]:
+            lib = lib or glob.glob("/usr/lib/libcrypto.*.dylib")
+            lib = list(reversed(sorted(lib)))
+        # last but not least all the other macOS versions should work here.
+        # including Big Sur
+        lib = lib[0] if lib else str("/usr/lib/libcrypto.dylib")
```
s0undt3ch, line 51: Why the `str()` cast here?

salt/utils/rsax931.py
```diff
@@ -30,6 +31,24 @@ def _find_libcrypto():
     """
     if sys.platform.startswith("win"):
         lib = str("libeay32")
```
kaorihinata, line 33: Now that @s0undt3ch mentions it, here as well.

kaorihinata: Ahh, found it. There was a comment here previously noting that Windows required the `str` call. Is that still the case, or was that an old artifact of Python 2?

salt/utils/rsax931.py
weswhet: Removed unnecessary string cast.

salt/utils/rsax931.py
weswhet: add the comment back in for windows and removed the str cast for macOS as it's not required.

weswhet:
@krionbsd ready for re-review. I made changes based on recommendations from @s0undt3ch.

salt/utils/rsax931.py
s0undt3ch: @twangboy do we still need this `str()` cast? I believe this was only necessary for Py2, right?

weswhet: I validated that on windows salt with py3 the cast is not needed. @s0undt3ch how do you want to proceed?

s0undt3ch: Nuke it too :explosion: :)

s0undt3ch: Sorry, Please.

weswhet: @s0undt3ch changed and ready to go :)

s0undt3ch:
It's properly labeled and when tests pass, queued for the 3001.1 release.

dwoz:
re-run full mac

PR #57861, opened 2020-07-03
Sync OpenBSD installation documentation with reality

jasperla:
# What does this PR do?

OpenBSD's salt package moved to python3 a while ago and dependencies
have changed since the installation documentation was last updated.

While here tweak a reference to the OpenBSD package, not pkg, repo.

### Commits signed with GPG?
Yes


jasperla:
@dwoz This is also a documentation fix, so I'm curious what testcase I should be writing here?

jasperla:
@dwoz Could you please explain why you added the "Needs Testcase" label because I honestly cannot think of a testcase to write.

waynew (approved):

PR #57876, opened 2020-07-06
Support for groups in publisher_acl access control list.

### LINKED ISSUE #42060: Publisher ACL System for Groups
### Feature request

Currently the Publisher ACL System allow users other than root to execute salt commands, but only users. I it would be really helpful if groups could be given permissions rather than a per user basis.

rene-dekkers:
- Feature Name: Publisher_acl group support
- Start Date: 2020-07-06
- RFC PR: 
- Salt Issue: 

# Summary
[summary]: #summary

Support for groups in publisher_acl access control list.

# Motivation
[motivation]: #motivation

Groups are currently not supported in the Master' config publisher_acl entries. Salt master will treat all items ending with % as a group and extracts its members from it. Function get_group_members is added to get members from a group.

Fixes #42060

## Alternatives
[alternatives]: #alternatives

As a workaround a Salt engineer can template a /etc/salt/master.d/acl.conf file like:
```
{%- set salt_group = salt['pillar.get']('salt_developers_group') %}
publisher_acl:
  {{ '\|'.join(salt['group.info'](salt_group)['members']) }}:
    - .*
```

## Unresolved questions
[unresolved]: #unresolved-questions

When a group member is added or removed from a group, the Salt master won't notice this. The keys in masterapi.py are written once (at startup). Key maintenance is not implemented.

# Drawbacks
[drawbacks]: #drawbacks

- Maybe it will affects acl with users ending with %
- Not able to combine a regex with a group name % entry. For example: user1|group1%.

If there is any need to fix one of these drawbacks I'm happy to change this proposal / PR!

### DIFF
--- a/salt/auth/__init__.py
+++ b/salt/auth/__init__.py
@@ -486,6 +486,16 @@ class LoadAuth(object):
                 auth_list = salt.utils.master.get_values_of_matching_keys(
                     self.opts["publisher_acl"], auth_ret
                 )
+
+                # Extend the auth_list with all acl items ending with % and the user is a member of it
+                groups = salt.utils.user.get_group_dict(auth_ret).keys()
+                for expr in [
+                    expr
+                    for expr in self.opts["publisher_acl"]
+                    if expr.endswith("%") and expr[:-1] in groups
+                ]:
+                    auth_list.extend(self.opts["publisher_acl"][expr])
+
                 if not auth_list:
                     ret["error"] = {"name": "UserAuthenticationError", "message": msg}
                     return ret
--- a/salt/daemons/masterapi.py
+++ b/salt/daemons/masterapi.py
@@ -221,6 +221,16 @@ def access_keys(opts):
     if opts.get("user"):
         acl_users.add(opts["user"])
     acl_users.add(salt.utils.user.get_user())
+
+    # Loop all entries which ends with % and add group members to acl_users
+    for group in [group for group in acl_users if group.endswith("%")]:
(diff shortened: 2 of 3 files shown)


Akm0d:
This looks good! Thanks for the PR. Once there are tests for the changes and all the tests are passing we can merge

rene-dekkers:
Thanks! Looking forward to see the testresults.

rene-dekkers:
Sorry, maybe I didn't understand correct. Do I have to provide these tests by myself? If so, could you please put me in the right direction?

Akm0d:
@rene-dekkers yes, that is the expectation.  A unit test in salt/tests/unit/utils/test_user.py and in salt/tests/unit/daemons/test_masterapi.py that tests all the changes are what's required.  I'm not sure where a test for  salt/auth/__init__.py should go... in the modified function it doesn't look like there is an existing test, @waynew do you have any advice?

adcpdk:
@dwoz Is there anything we can do to proceed with this feature?
This would be a huge benefit for Salt as a product.

Regards,
Dmitriy

davama:
sorry for bump
but just want to see if this is still moving along... would love to see this feature
Thank you for the PR

tazaki:
Another bump for this!

mdschmitt:
And another bump. @gtmanfred , you were one of the original contacts on #42060 , can you weigh in here?

Ch3LL:
This PR would require a changelog and test coverage to get this past the line, if anyone is willing to do this that would be appreciated.

dwoz:

Issue #56498, opened 2020-04-02
service.running doesn't properly detect systemd in 3000.1


dkacar-oradian:
### Description of Issue
I didn't look at the code for previous versions. In v3000.1  `salt/states/service.py` has `SYSTEMD_ONLY` variable with the list of arguments which are valid only for systems running `systemd`. And then the code checks if those arguments are present in the state functions.

However, the check doesn't work correctly on CentOS 7 at least. That OS is using `systemd`, but salt code doesn't detect it properly and warns that `systemd` specific arguments are not supported.

I don't know if there are more serious issues than warnings.

### Setup
(Please provide relevant configs and/or SLS files (Be sure to remove sensitive info).)

`test.sls`:
```
firewall_service_run:
  service.running:
    - name: "iptables.service"
    - enable: True
    - unmask: True
```
When I run this on CentOS 7 minion with salt v3000.1 I get:

```
# salt test-dbsnap-02.xdc state.apply test         
test-dbsnap-02.xdc:
----------
          ID: firewall_service_run
    Function: service.running
        Name: iptables.service
      Result: True
     Comment: The service iptables.service is already running
              Service iptables.service was already unmasked
     Started: 14:37:52.031491
    Duration: 46.795 ms
     Changes:   
    Warnings: The 'unmask' argument is not supported by this platform/action

Summary for test-dbsnap-02.xdc
------------
Succeeded: 1
Failed:    0
Warnings:  1
------------
Total states run:     1
Total run time:  46.795 ms
```
The warning shouldn't be there because the OS has `systemd` and that service is being managed by `systemd`.

OS info:
```
{test-dbsnap-02}~> cat /etc/redhat-release 
CentOS Linux release 7.7.1908 (Core)
```

For a test, I'll stop and mask the service:
```
{test-dbsnap-02}~# systemctl stop iptables.service    
{test-dbsnap-02}~# systemctl disable iptables.service 
Removed symlink /etc/systemd/system/basic.target.wants/iptables.service.
{test-dbsnap-02}~# systemctl mask iptables.service   
Created symlink from /etc/systemd/system/iptables.service to /dev/null.
```
And if I run `test.sls` from above I get:
```
# salt test-dbsnap-02.xdc state.apply test
test-dbsnap-02.xdc:
----------
          ID: firewall_service_run
    Function: service.running
        Name: iptables.service
      Result: True
     Comment: Service iptables.service has been enabled, and is running
              Service iptables.service was unmasked
[...]

PR #57891, opened 2020-07-08
fix Service false positive

cmcmarrow:
What does this PR do?
When service is filterering kwargs it warns the users that a kwargs is not supported by their OS/Action. 

Warning the user about a bad kwarg for a action is pointless and confusing because the action doesn't need it and the filter is doing its job. 

The only time when we should warn the User is when the OS does not support the kwarg because that can cause failures. 

Fixes:
#56498

GPG:
Yes

### DIFF
--- a/salt/utils/systemd.py
+++ b/salt/utils/systemd.py
@@ -18,17 +18,6 @@ from salt.exceptions import SaltInvocationError
 log = logging.getLogger(__name__)
 
 
-def _has_systemd_support():
-    try:
-        output = subprocess.check_output(("systemctl",), shell=False).decode().lower()
-        return "systemd" in output
-    except FileNotFoundError:
-        return False
-
-
-HAS_SYSTEMD = _has_systemd_support()
-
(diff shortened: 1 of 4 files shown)


salt/utils/systemd.py
```diff
@@ -18,6 +18,17 @@
...
 
+def _has_systemd_support():
+    try:
+        output = subprocess.check_output(("systemctl",), shell=False).decode().lower()
+        return "systemd" in output
+    except FileNotFoundError:
+        return False
+
+
+HAS_SYSTEMD = _has_systemd_support()
```
DmitryKuzmenko, line 29: `salt.modules.systemd` uses another approach to check systemd presence on the system.
```
salt.utils.systemd.booted(__context__)
```
Is this an expected difference?

cmcmarrow: yea your right

DmitryKuzmenko (approved):

PR #57896, opened 2020-07-09
Revert "Add ShiftLeft Scan to the pipeline"

s0undt3ch:
### What does this PR do?
This reverts commit 8a2c331a4752b98632d1fa231adb6918f483a14f.

This shouldn't have been merged before fixing all the issues it reports,
if we want to accept and fix all issues it reports, or the reporter.

Fixes the GitLab pipeline


bryceml:
We can also just change this to "Allow Failures" on the gitlab job then it will always run  but not mess up the rest of the pipeline.

s0undt3ch:
No. This check, if going in, needs to address(fix or skip) the several hundred warnings it raises.

s0undt3ch:
After we remove it.

bryceml (dismissed):

bryceml (approved):

PR #57897, opened 2020-07-09
[PyTest] Remove `runtests.py` and it's supporting code

s0undt3ch:
### What does this PR do?
Switches from `runtests.py` to PyTest


rst0git:
Hi @s0undt3ch, it looks like this PR may has a conflict with the changes in https://github.com/saltstack/salt/pull/57947 / https://github.com/saltstack/salt/pull/57736. In particular, commit [tests: Add virt minion integration targets](https://github.com/saltstack/salt/pull/57947/commits/cc0a00f861750ab926657f6167b9d88710096750) has several changes in `tests/support/processes.py` that are used to start a salt-minion within a Docker container with pre-installed libvirt+qemu.

Do you have any suggestions on how to resolve the conflict? 

(cc @cbosdo)

s0undt3ch:
Hello @rst0git.

First of all, sorry for the long wait on the reply.

I have bad and good news for you.

The bad news(for you) is that, salt **is** switching to Pytest and **will** drop `runtests.py`. As such, a fair chunk of your work was wasted(more or less), will conflict and will stop working.

The good news is, Salt will finally be switching to PyTest!

No, seriously, the good news is that, this move will actually make some of the work you did unnecessary going forward.
Also good news is that I've been working on pytest-salt-factories to simplify running docker containers in Salt's tests suite, more, running salt daemons in docker containers which is part of what you had to do.
Some things need to happen first, ie, some PRs need to be merged first, but I will personally look into you're PRs and adjust them(to *try* and balance your lost work), either by pushing code to the PR, or opening PRs against your fork so that the required daemons you need on your tests work.

This will take some time, but, as far as your PRs working with pytest, will happen before the next major release is out, I'm aiming to be quite before we release. However, it will take some time.

I hope this bump on your valuable contributions don't discourage you from continuing to contribute in the future.

noxfile.py
```diff
@@ -352,314 +352,6 @@ def _run_with_coverage(session, *test_cmd):
...
-
-        for idx, flag in enumerate(cmd_args[:]):
-            if "--names-file=" in flag:
-                cmd_args.pop(idx)
-                break
-            elif flag == "--names-file":
-                cmd_args.pop(idx)  # pop --names-file
-                cmd_args.pop(idx)  # pop the actual names file
-                break
-        cmd_args.append("--names-file={}".format(names_file_path))
```
rst0git, line 425: perhaps `--names-file` in `tests/conftest.py` can be dropped as well?

s0undt3ch: There are a few CLI flags, which were also migrated to PyTest for backwards compat reasons, that can be removed.
It should go in another stage/PR and needs a careful review since our CI process might rely on them.

.ci/kitchen-macosxmojave-py3
```diff
@@ -9,11 +9,12 @@ runTestSuite(
     golden_images_branch: 'master',
     jenkins_slave_label: 'kitchen-slave-mac',
     kitchen_platforms_file: '/var/jenkins/workspace/pre-golden-nox-platforms.yml',
-    nox_env_name: 'runtests-zeromq',
+    nox_env_name: 'pytest-zeromq',
     nox_passthrough_opts: '--ssh-tests',
     python_version: 'py3',
-    run_full: params.runFull,
+    //splits: ['unit', 'integration', 'multimaster'],
```
waynew, line 15: Why are these in here commented out?

waynew (changes requested):

.ci/kitchen-macosxmojave-py3
s0undt3ch: Because not all of them are split and I don't want to test and try again to get that right, when/if we split

waynew: Would it be useful to add that comment into the text here (and elsewhere we have the splits)?

s0undt3ch: Removed. Git history will reveal them back

waynew (dismissed):

s0undt3ch:
re-run full all

krionbsd (dismissed):

rst0git (dismissed):

s0undt3ch:
re-run full all

s0undt3ch:
re-run full all

s0undt3ch:
re-run full all

s0undt3ch:
re-run full all

s0undt3ch:
re-run full all

s0undt3ch:
re-run full all

s0undt3ch:
re-run full all

PR #57925, opened 2020-07-13
Remove repeated call to _refresh_file_mapping

mbirtwell:
### What does this PR do?

* _refresh_file_mapping is always called via super().__init__() -> clear() so doesn't need to be called directly in the constructor.
* The constructor always sets self.opts, so the hasattr in clear will always pass. So I've removed it.

### Previous Behavior
Ran _refresh_file_mapping twice every time a LazyLoader was constructed. I was doing some profiling and this would be responsible for about half the run time of some shorter state executions. For example saltutil.find_jobs.

### New Behavior
Don't run the extra call to _refresh_file_mapping. Shouldn't change the observable behaviour at all. 

- [ ] Docs - Not relevant.
- [ ] Changelog - https://docs.saltstack.com/en/master/topics/development/changelog.html - done
- [ ] Tests written/updated - Not relevant.

### Commits signed with GPG?
No


mbirtwell:
@dwoz I'm not sure how to go about testing this. The behaviour should be entirely unchanged and all existing tests pass for me (and on most OSs). Any suggestions?

mbirtwell:
I stand by this PR, my former employer had been using this in production for more than a year when I left and I believe that the check failures were flakes and not to do with the changes in this PR. But I want to close this to get it off my dashboard.

PR #49678, opened 2018-09-17
Master stats latency

sbrennan4:
What does this PR do?
Adds a latency measurement to the master_stats data. Latency is measured as the time between the JID being created and the job being picked up by an mworker process.

Previous Behavior
Currently, only mean job duration is collected.

New Behavior
Adds mean job latency to master stats collection.

Tests written?
No

Commits signed with GPG?
No

Please review Salt's Contributing Guide for best practices.

See GitHub's page on GPG signing for more information about signing commits with GPG.

PR #57929, opened 2020-07-13
port #49678, #53823, #53835, #53832

cmcmarrow:
### What does this PR do?
port #49678, #53823, #53835, #53832
https://github.com/saltstack/salt/issues/54813
commits signed with GPG?
Yes


Ch3LL (changes requested):
this will need some test coverage and looks like there is a pre-commit failure and a merge conflict that needs to be cleaned up.

sagetherage:
these ports can be de-scoped from Magnesium and later we can focus on writing tests @cmcmarrow

dwoz:

PR #57942, opened 2020-07-14
Fix bugs with macOS and Windows services.

### LINKED ISSUE #49237: service.status doesn't behave the same way on MacOS as on other platforms
On most platforms, the `service.status` module returns either `True` or `False` depending on the status of a given service. But on MacOS, it returns `False` if the service isn't running, or a list of process IDs if the service is running. This makes it difficult to make platform-agnostic tools on top of Salt, since special consideration has to be made for Mac OS.

### Steps to Reproduce Issue
Run `salt MACOS-MINION service.status servicename` against a running service

### Versions Report
2018.3.2

### LINKED ISSUE #57907: macOS highstate takes a long time with multiple service.dead calls that have missing services.
**Description**
On macOS if you have a service.dead state that is looking for a missing service. Salt will look twice for that missing service. This behavior was introduced in a change to fix a different issue #48414, the method salt uses to find all available services is pretty extensive which causes a delay every time it has to double-check a service should be missing. We should not double check if a service is present if we are checking from a service.dead state.

I'll be attaching a PR to fix shortly.

### LINKED ISSUE #57908: macOS and Windows service states fail to start if disabled, even with enable: true
**Description**
If you attempt to start a service on macOS or Windows and the service is disabled `service.running` with fail to start the service even if the `enable: true` option is set.

This is caused by the service state wanting to enable a service after it attempts to start the state. On macOS and Windows the service state should attempt to enable a service prior to starting it. 

PR coming soon.

cc @twangboy

weswhet:
### What does this PR do?
Fixes multiple issues with macOS and Windows services.

Fixes: #57908 
Fixes #57907 
Fixes #49237 

### Previous Behavior
salt would get really sad when trying to start some services on macOS and Windows.

### New Behavior
They work as intended... 😄 

- [x] Docs
- [x] Tests written/updated

### Commits signed with GPG?
Yes


noelmcloughlin (dismissed):
LGTM, just some quick eyes.

tests/unit/modules/test_mac_service.py
```diff
@@ -206,3 +248,165 @@ def test_service_keep_alive_wrong_setting(self):
...
+            assert mac_service.status("com.salt") is False
+
+    @patch.object(mac_service, "_launch_agent", lambda _: False)
+    @patch.object(mac_service, "_get_service", lambda _: {"": ""})
+    @patch.object(mac_service, "_always_running_service", lambda _: True)
+    def test_service_status_on_daemon_with_pid(self):
+        """
+        Test service status on dameon with PID.
+        """
+        mock_servie_list = '{\n\t"LimitLoadToSessionType" = "System";\n\t"Label" = "com.salt";\n\t"OnDemand" = false;\n\t"LastExitStatus" = 0;\n\t"PID" = 218;\n\t"Program" = "/opt/salt";\n\t\t"--disable-keepalive";\n\t);\n};'
```
sheagcraig, line 304: Do you really want "mock_servie_list" here?

Akm0d:
re-run full all

tests/unit/modules/test_mac_service.py
weswhet: handled

weswhet:
re-run full all

Akm0d (approved):

dwoz (approved):

PR #57946, opened 2020-07-15
Use brew path from which in mac_brew_pkg module

arsa-dev:
Call brew directly to fetch brew full path will throw command not found
error. Fetching with which as in other parts of module code works fine.

### What does this PR do?
I checked a bit deeply and I see that, as @Ch3LL explain at PR #48804, after this fix #47212 we aren't passing the PATH via the env anymore so we need to run which brew to get the entire path or else we see this error. So call brew directly to fetch brew full path will throw previous error, and I think that a valid solution can be fetch full path as @Ch3LL proposes in the past.

Related with previously merged PR #48804. See saltstack/salt-jenkins#1033 conversation for details.

### Previous Behavior
I was getting the following error trying to run 'salt <minion> pkg.install wget':
```
ERROR: Unable to run command '['brew', '--prefix']' with the context '{'env': {'LC_NAME': 'C', 'XPC_FLAGS': '0x0', 'LC_MONETARY': 'C', 'LC_MESSAGES': 'C', 'LC_IDENTIFICATION': 'C', 'LC_MEASUREMENT': 'C', 'LC_COLLATE': 'C', 'SHLVL': '0', 'LC_PAPER': 'C', 'LC_NUMERIC': 'C', 'LC_CTYPE': 'C', 'LC_TELEPHONE': 'C', 'LC_ADDRESS': 'C', 'PATH': '/usr/bin:/bin:/usr/sbin:/sbin', 'LANGUAGE': 'C', 'PWD': '/', 'XPC_SERVICE_NAME': 'com.saltstack.salt.minion', 'LC_TIME': 'C'}, 'close_fds': True, 'bg': False, 'with_communicate': True, 'shell': False, 'stderr': -2, 'timeout': None, 'cwd': '/var/root', 'stdout': -1, 'stdin': None}', reason: command not found
ERROR: Minions returned with non-zero exit code
```

### New Behavior
Command 'salt <minion> pkg.install wget' runs well and do his work.

- [x] Docs
- [x] Tests written/updated

### Commits signed with GPG?
Yes


arsa-dev:
Hi, I've reopened this PR since a long time ago without using salt with macos, I've realised that I'm experiencing the same error behaviour when trying to pkg.install something with brew on macos:
```
ERROR: Unable to run command '['brew', '--prefix']' with the context '{'cwd': '/var/root', 'shell': False, 'env': {'OLDPWD': '/', 'PATH': '/usr/bin:/bin:/usr/sbin:/sbin', 'PWD': '/opt/salt', 'XPC_FLAGS': '0x0', 'XPC_SERVICE_NAME': '0', 'SHLVL': '0', 'LC_CTYPE': 'C', 'SSL_CERT_DIR': '/private/etc/ssl/certs', 'SSL_CERT_FILE': '/private/etc/ssl/cert.pem', 'LC_NUMERIC': 'C', 'LC_TIME': 'C', 'LC_COLLATE': 'C', 'LC_MONETARY': 'C', 'LC_MESSAGES': 'C', 'LC_PAPER': 'C', 'LC_NAME': 'C', 'LC_ADDRESS': 'C', 'LC_TELEPHONE': 'C', 'LC_MEASUREMENT': 'C', 'LC_IDENTIFICATION': 'C', 'LANGUAGE': 'C'}, 'stdin': None, 'stdout': -1, 'stderr': -2, 'with_communicate': True, 'timeout': None, 'bg': False, 'close_fds': True}', reason: [Errno 2] No such file or directory: 'brew'
```

This is reproduced with latest released (3006.1) salt-master & salt-minion on macOS Ventura (also latest) with a just installed homebrew.

s0undt3ch (dismissed):

arsa-dev:
Hi @s0undt3ch, thanks for approving previous changes!
After reviewing ci logs I've pushed two new commits, one for apply diff suggested by this CI job: https://github.com/saltstack/salt/actions/runs/5298983857/jobs/9593574724?pr=57946
And use brew path format at `def _homebrew_bin()` instead of '+' operator in order to avoid coverage `TypeError: unsupported operand type(s) for +: 'NoneType' and 'str'` shown at this CI job: https://github.com/saltstack/salt/actions/runs/5298983857/jobs/9593610712?pr=57946

s0undt3ch (dismissed):
Please also add a changelog file regarding your fix.

arsa-dev:
> Please also add a changelog file regarding your fix.

Thx for noticing about changelog @s0undt3ch, I've added it in the latest commit, together to a new lint fix that appeared at: https://github.com/saltstack/salt/actions/runs/5302730009/jobs/9597939217?pr=57946

Let me know if I can do anything else

s0undt3ch (dismissed):

s0undt3ch:
Thanks @arsa-dev for the prompt fixes.

salt/modules/mac_brew_pkg.py
```diff
@@ -31,7 +31,7 @@ def __virtual__():
     """
     if __grains__["os"] != "MacOS":
         return False, "brew module is macos specific"
-    if not salt.utils.path.which("brew"):
+    if not _homebrew_bin():
```
Ch3LL, line 34: @s0undt3ch this function calls `__salt__['cmd.run']` Can we rely on this module being loaded and availalbe in this virtual function?

s0undt3ch: Good question. I'm not sure, but I'm leaning we can't guarantee.

Now, the question also is, do we have to use` __salt__` for that command?

Ch3LL: @arsa-dev since we are just ensuring `brew` is available and not actually running a brew command in the `__virtual__` func, couldn't we just keep this call in the `__virtual__` function to `salt.utils.path.which`?

arsa-dev: Hi @Ch3LL, first of all thanks you so much for your review and sorry for my delay, last week was a difficult one, I promise to be here more time this week so we can hopefully merge it 😊.

You are totally right, we shouldn't perform salt.run when we cannot fetch the os executable path with `salt.utils.path.which` and also there is no need to get the installation prefix when checking that `brew` is available since running `brew --prefix` should everytime return a valid output if we have the os executable path (except if brew installation is corrupt but this is something that I think it's out of scope).

I've rebased the branch with the latest master version and implemented a new `_homebrew_os_bin` func for fetching the os path with `which` without duplicating `salt.utils.path.which("brew")` instruction, used it at `__virutal__` func and also `_homebrew_bin` func that now comes with a safety extra check just in case the module is loaded and brew not available anymore, we are not calling cmd.run of `None --prefix`

salt/modules/mac_brew_pkg.py
```diff
@@ -93,13 +93,26 @@ def _tap(tap, runas=None):
...
 def _homebrew_bin():
     """
-    Returns the full path to the homebrew binary in the PATH
+    Returns the full path to the homebrew binary in the homebrew installation folder
     """
-    ret = __salt__["cmd.run"]("brew --prefix", output_loglevel="trace")
-    ret += "/bin/brew"
-    return ret
+    brew = _homebrew_os_bin()
+    if not brew:
```
Ch3LL, line 108: `salt.utils.path.which("brew")` returns None already if `brew` doesn't already exist. This isn't a hard ask, so you can implement it if you want, but we could continue just returning the result of the command.

Ch3LL (dismissed):

salt/modules/mac_brew_pkg.py
arsa-dev: Hi again @Ch3LL, I think that I understood correctly your suggestion, it's now returning `_homebrew_os_bin()` value directly in case there is no a true value there and in case there is a true value, it's fetching the installation path and overriding the variable to return. I've also rebased the branch again with latest master version.

Please if I misunderstood your suggestion or you have a new one do not hesitate to comment and I'll be happy to apply it.

Ch3LL (dismissed):

garethgreenaway (dismissed):

s0undt3ch (dismissed):

twangboy (dismissed):

arsa-dev:
Hi @Ch3LL, I've added a new commit (and also rebased again with latest master) that should fix the failing test. The problem was that we wasn't mocking path.which function and this is causing to return None instead of the expected value. Would be great if someone can approve to confirm that all the tests passes on ci.

arsa-dev:
> Would be great if someone can approve to confirm that all the tests passes on ci.

cc: @s0undt3ch @twangboy @garethgreenaway

s0undt3ch (approved):

arsa-dev:
Hi @s0undt3ch, thanks for approving, but seems that the pipeline was canceled without attempting it again (see https://github.com/saltstack/salt/actions/runs/5418501547/jobs/9875210581?pr=57946). So I've rebased the branch again with latest changes, can you or someone else like @Ch3LL approve to execute the pipeline and run tests? I think that should be everything there right now.

s0undt3ch (approved):

arsa-dev:
Hi @s0undt3ch, I've checked the test problems for the previous execution, and seems unrelated to the changes here, maybe you or someone can confirm (I've checked specifically this job: https://github.com/saltstack/salt/actions/runs/5435566699/jobs/9885898981?pr=57946). I've rebased the branch again with latest master version but as I've seen this version also have failing tests, so probably tests here will also fail due to this reason as we can see at this job: https://github.com/saltstack/salt/actions/runs/5447608111/jobs/9911153644. Since I think that everything at this PR is ready ¿Should I continue pinging back when I rebase new changes from master in order to get it merged? ¿or your (or the team) will be checking to merge before the upcoming release?

On the other hand,  @Ch3LL I think that both review thread are resolved, should I mark resolved or you will do that?

s0undt3ch:
The package tests issues are something that needs to be resolved asap.

If prefer to get that fixed before merging your PR. If however I get three reviews from the team, we can merge this(though this 4th of July week, most are out).

Don't worry, you've done your part and this will get merged 

Thank You!

Ch3LL (approved):

PR #57959, opened 2020-07-16
Archive glob with cwd

github-abcde:
### What does this PR do?
This PR fixes the use-case where `sources` contain a wildcard (and thus are globbed) and at the same time `cwd` is used to ensure only relative paths are stored in the archive.

None that I'm aware of

### Previous Behavior
The use of `cwd` could only be combined with providing a (list of) absolute path(s) for `sources` to be archived.

### New Behavior
The `cwd` argument can now be used together with wildcards in `sources` entries.
For example:
```
salt-call --local archive.zip /tmp/test.zip 'ar*' cwd=/home/user/src/salt/salt/modules
```
will now result in the following files being added to `/tmp/test.zip`:
```
local:
    - archive.py
    - arista_pyeapi.py
    - artifactory.py
```
where it previously would result in an empty zipfile.

- [X] Docs
- [X] Tests written/updated

### Commits signed with GPG?
Yes


dwoz:

PR #57966, opened 2020-07-16
[Documentation] Cleanup docs related to Pushover

jasperla:
## What does this PR do?

- the service name is *Pushover*, not *PushOver* so update the documentation to reflect that
- add missing newline states.pushover which broke the codeblock: see [this page](https://docs.saltstack.com/en/master/ref/states/all/salt.states.pushover.html)

### Commits signed with GPG?
Yes


jasperla:
@dwoz Here too, what kind of testcases should I be writing for these comment/documentation fixes? Or are you saying I should write new tests for these four modules?

jasperla:
re-run full all

waynew (changes requested):
@jasperla can you go ahead and revert the changes to the manpage? They should be generated automagically (or partially manually during the release process). Then this should be 🚀

waynew (approved):
🚀

jasperla:
re-run macos

jasperla:

PR #57936, opened 2020-07-14
Start upgrading codebase to Py3 for every changed file

s0undt3ch:
### What does this PR do?
Starts upgrading the codebase to Py3 for every changed file by running [`pyupgrade`](https://github.com/asottile/pyupgrade) against it.
We only do this for each changed file and not the whole code base to avoid MASSIVE code churn.

PR #57188, opened 2020-05-10
Refactor SyncWrapper

dwoz:
Refactor SyncWrapper so that it uses an alternate ioloop and runs it in
a thread. This is needed for torando 5.0 and greater on python 3 where
asyncio's event loop is the default. The asyncio default loop can only
have on loop running at a time in any given thread. So, the old
SyncWrapper doesn't work because it tires to run multiple ioloops in the
main execution thread.

### Merge requirements satisfied?

- [ ] Docs
- [ ] Changelog - https://docs.saltstack.com/en/master/topics/development/changelog.html
- [x] Tests written/updated

### Commits signed with GPG?
Yes

PR #57969, opened 2020-07-17
Added an idem exec and state module

### LINKED ISSUE #57980: [FEATURE REQUEST] Run idem states and modules from within salt
Salt needs the ability to run idem states.  It can manage puppet, chef, and ansible, support for the next-gen idempotent state system makes a lot of sense.

An execution module and state module that can run idem execution modules and states respectively

**Describe alternatives you've considered**
One thought was to include a dunder hub in the loader and then allow states to specify which engine to use, the salt engine or the idem engine.  idem/pop is fundamentally different from salt and execution/state modules calling idem code is a more uninvolved way to get idem functionality in salt.

Akm0d:
### What does this PR do?
fixes #57980
depends #57993

Adds an exec module and a state module for idem.
Idem execution modules and states can now be run through salt in a non-obtrusive way

- [x] Docs
- [x] Tests written/updated

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/modules/idem.py
+++ b/salt/modules/idem.py
@@ -1,0 +1,72 @@
+# -*- coding: utf-8 -*-
+#
+# Author: Tyler Johnson <tjohnson@saltstack.com>
+#
+
+"""
+Idem Support
+============
+
+This module provides access to idem execution modules
+
+.. versionadded:: Magnesium
+"""
+# Function alias to make sure not to shadow built-in's
+__func_alias__ = {"exec_": "exec"}
+__virtualname__ = "idem"
+
+
+def __virtual__():
+    if "idem.hub" in __utils__:
+        return __virtualname__
+    else:
+        return False, "idem is not available"
+
+
+def exec_(path, acct_file=None, acct_key=None, acct_profile=None, *args, **kwargs):
+    """
+    Call an idem execution module
+
+    path
+        The idem path of the idem execution module to run
+
+    acct_file
+        Path to the acct file used in generating idem ctx parameters.
+        Defaults to the value in the ACCT_FILE environment variable.
+
+    acct_key
+        Key used to decrypt the acct file.
+        Defaults to the value in the ACCT_KEY environment variable.
+
+    acct_profile
+        Name of the profile to add to idem's ctx.acct parameter.
+        Defaults to the value in the ACCT_PROFILE environment variable.
+
+    args
+        Any positional arguments to pass to the idem exec function
+
+    kwargs
+        Any keyword arguments to pass to the idem exec function
+
+    CLI Example:
+
+    .. code-block:: bash
+
+        salt '*' idem.exec test.ping
+
+    :maturity:      new
+    :depends:       acct, pop, pop-config, idem
+    :platform:      all
+    """
+    hub = __utils__["idem.hub"]()
+
+    coro = hub.idem.ex.run(
+        path,
+        args,
+        {k: v for k, v in kwargs.items() if not k.startswith("__")},
+        acct_file=acct_file or hub.OPT.acct.acct_file,
+        acct_key=acct_key or hub.OPT.acct.acct_key,
+        acct_profile=acct_profile or hub.OPT.acct.acct_profile or "default",
+    )
+
+    return hub.pop.Loop.run_until_complete(coro)
--- a/salt/states/idem.py
+++ b/salt/states/idem.py
@@ -1,0 +1,159 @@
+# -*- coding: utf-8 -*-
+#
+# Author: Tyler Johnson <tjohnson@saltstack.com>
+#
+
+"""
+Idem Support
+============
+
+This state provides access to idem states
+
+.. versionadded:: Magnesium
+"""
+import pathlib
+import re
+
+__virtualname__ = "idem"
+
+
+def __virtual__():
+    if "idem.hub" in __utils__:
+        return __virtualname__
+    else:
+        return False, "idem is not available"
+
+
+def _get_refs(sources, tree):
+    """
+    Determine where the sls sources are
+    """
+    sls_sources = []
+    SLSs = []
+    if tree:
+        sls_sources.append("file://{}".format(tree))
+    for sls in sources:
+        path = pathlib.Path(sls)
+        if path.is_file():
+            ref = str(path.stem if path.suffix == ".sls" else path.name)
+            SLSs.append(ref)
+            implied = "file://{}".format(path.parent)
+            if implied not in sls_sources:
+                sls_sources.append(implied)
+        else:
+            SLSs.append(sls)
+    return sls_sources, SLSs
+
+
+def _get_low_data(low_data):
+    """
+    Get salt-style low data from an idem state name
+    """
+    # state_|-id_|-name_|-function
+    match = re.match(r"(\w+)_\|-(\w+)\|-(\w+)_\|-(\w+)", low_data)
+    return {
+        "state": match.group(1),
+        "__id__": match.group(2),
+        "name": match.group(3),
+        "fun": match.group(4),
+    }
+
+
+def state(
+    name,
+    sls,
+    acct_file=None,
+    acct_key=None,
+    acct_profile=None,
+    cache_dir=None,
+    render=None,
+    runtime=None,
+    source_dir=None,
+    test=False,
+):
+    """
+    Call an idem state through a salt state
+
+    sls
+        A list of idem sls files or sources
+
+    acct_file
+        Path to the acct file used in generating idem ctx parameters.
+        Defaults to the value in the ACCT_FILE environment variable.
+
+    acct_key
+        Key used to decrypt the acct file.
+        Defaults to the value in the ACCT_KEY environment variable.
+
+    acct_profile
+        Name of the profile to add to idem's ctx.acct parameter
+        Defaults to the value in the ACCT_PROFILE environment variable.
+
+    cache_dir
+        The location to use for the cache directory
+
+    render
+        The render pipe to use, this allows for the language to be specified (jinja|yaml)
+
+    runtime
+        Select which execution runtime to use (serial|parallel)
+
+    source_dir
+        The directory containing sls files
+
+    .. code-block:: yaml
+
+    cheese:
+      idem.state:
+        - runtime: parallel
+        - sls:
+          - idem_state.sls
+          - sls_source
+
+    :maturity:      new
+    :depends:       acct, pop, pop-config, idem
+    :platform:      all
+    """
+    hub = __utils__["idem.hub"]()
+
+    if isinstance(sls, str):
+        sls = [sls]
+
+    sls_sources, SLSs = _get_refs(sls, source_dir or hub.OPT.idem.tree)
+
+    coro = hub.idem.state.apply(
+        name=name,
+        sls_sources=sls_sources,
+        render=render or hub.OPT.idem.render,
+        runtime=runtime or hub.OPT.idem.runtime,
+        subs=["states"],
+        cache_dir=cache_dir or hub.OPT.idem.cache_dir,
+        sls=SLSs,
+        test=test,
+        acct_file=acct_file or hub.OPT.acct.acct_file,
+        acct_key=acct_key or hub.OPT.acct.acct_key,
+        acct_profile=acct_profile or hub.OPT.acct.acct_profile or "default",
+    )
+    hub.pop.Loop.run_until_complete(coro)
+
+    errors = hub.idem.RUNS[name]["errors"]
+    success = not errors
+
+    running = []
+    for idem_name, idem_return in hub.idem.RUNS[name]["running"].items():
+        standardized_idem_return = {
+            "name": idem_return["name"],
+            "changes": idem_return["changes"],
+            "result": idem_return["result"],
+            "comment": idem_return.get("comment"),
+            "low": _get_low_data(idem_name),
+        }
+        running.append(standardized_idem_return)
+
+    return {
+        "name": name,
+        "result": success,
+        "comment": "Ran {} idem states".format(len(running)) if success else errors,
+        "changes": {},
+        "sub_state_run": running,
+    }
--- a/salt/utils/idem.py
+++ b/salt/utils/idem.py
@@ -1,0 +1,58 @@
+# -*- coding: utf-8 -*-
+"""
+Idem Support
+============
+
+This util provides access to an idem-ready hub
+
+.. versionadded:: Magnesium
+"""
+import logging
+import sys
+
+try:
+    import pop.hub
+
+    HAS_POP = True, None
+except ImportError as e:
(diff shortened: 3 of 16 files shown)


salt/states/idem.py
```diff
@@ -0,0 +1,148 @@
...
+        hub.pop.sub.add(dyne_name="config")
+        hub.pop.sub.add(dyne_name="idem")
+        hub.pop.sub.add(dyne_name="exec")
+        hub.pop.sub.add(dyne_name="states")
+        log.debug("Reading idem config options")
+        hub.config.integrate.load(["acct", "idem"], "idem", parse_cli=False, logs=False)
+        __context__["idem.hub"] = hub
+
+
+def _get_refs(hub, sources):
```
mchugh19, line 59: In case it's desired, this doesn't support salt environments

Akm0d: Idem does not support salt environments.  Instead it uses `--sls-sources`, `--tree`, and `acct`.  I added support for acct, but maybe I need to add a way to set sls-sources and tree

mchugh19: Ah, I wasn't sure if you intended to support salt:// urls for file serving. Integration with salt could mean that idem sls files live on the master (like other salt states), and are transparently copied to the minion for idem to render/run. But this integration is aiming for only aiming to pass local file paths to idem. 👍

Akm0d: For now idem states are only intended to be run on the master minion

salt/states/idem.py
```diff
@@ -0,0 +1,149 @@
...
+
+# Function alias to make sure not to shadow built-in's
+__func_alias__ = {"exec_": "exec"}
+__virtualname__ = "idem"
+
+
+def __virtual__():
+    if sys.version_info < (3, 6):
+        return False, "idem only works on python3.6 and later"
+    if not salt.modules.idem.HAS_POP[0]:
```
s0undt3ch, line 30: Instead of cross-importing a state just to get it's Boolean value if it loaded, look for `idem.<a func name that should have loaded>` in `__salt__`

salt/states/idem.py
```diff
@@ -0,0 +1,149 @@
...
+
+
+def _get_refs(sources, tree):
+    """
+    Determine where the sls sources are
+    """
+    sls_sources = []
+    SLSs = []
+    for sls in sources:
+        if os.path.isfile(sls):
```
s0undt3ch, line 47: I'm not asking for a change, but have you looked at `pathlib` from the stdlib?

tests/integration/modules/test_idem.py
```diff
@@ -0,0 +1,32 @@
+# -*- coding: utf-8 -*-
+"""
+Integration tests for the idem execution module
+"""
+
+# Import futures
```
s0undt3ch, line 6: Please remove the import comments, we should get rid of all of them.

s0undt3ch (changes requested):

CHANGELOG.md
```diff
@@ -63,6 +63,7 @@ Added
 -----
 
 - Added docs demonstrating how to apply an MSI patch with winrepo (#32780)
+- Added an execution and state module for running idem modules and states (#57969)
```
waynew, line 66: This should just be a [changelog entry via towncrier](https://docs.saltstack.com/en/master/topics/development/changelog.html)

salt/states/idem.py
waynew: It's not just the loaded - it also provides the correct error message, FWIW.

salt/states/idem.py
waynew: I'd ask for a change 😂 

Especially since this requires Python 3.6+, there's no reason to *not* use pathlib.

tests/integration/modules/test_idem.py
```diff
@@ -0,0 +1,32 @@
...
+from __future__ import absolute_import, print_function, unicode_literals
```
waynew, line 7: @s0undt3ch do we still need these? That might be an obsolete lint rule we need to remove?

tests/integration/modules/test_idem.py
```diff
@@ -0,0 +1,32 @@
...
+import salt.utils.path
+
+# Import salt testing libs
+from tests.support.case import ModuleCase
+from tests.support.unit import skipIf
+
+log = logging.getLogger(__name__)
+
+
+@skipIf(not salt.modules.idem.HAS_POP[0], salt.modules.idem.HAS_POP[1])
```
waynew, line 23: Would this be better as `@skipIf(not all(salt.modules.idem.HAS_POP))`?

waynew (changes requested):

tests/integration/modules/test_idem.py
Akm0d: The second one gives the reason why, it's a string representation of the import error

waynew: Oh, never mind, I misread that 😂

salt/states/idem.py
s0undt3ch: Which the exec module would have logged, the message.

tests/integration/modules/test_idem.py
s0undt3ch: I have a draft PR that also takes care of those, but yes, both the coding line and the future imports.

salt/states/idem.py
waynew: Logged, or presented to the user?

tests/integration/modules/test_idem.py
waynew: Should this merge after that?

s0undt3ch: https://github.com/saltstack/salt/pull/57936

s0undt3ch: Not required. I'm not removing any code, I added a pre-commit rule that removes them for each edited file, to avoid massive changes in a single PR.

salt/states/idem.py
s0undt3ch: Salt doesn't present anything to the user when a module does not load. Or state module. It logs and wines that the module could not be found

tests/integration/modules/test_idem.py
s0undt3ch: https://github.com/saltstack/salt/issues/57979

salt/states/idem.py
```diff
@@ -0,0 +1,150 @@
...
+def _get_refs(sources, tree):
+    """
+    Determine where the sls sources are
+    """
+    sls_sources = []
+    SLSs = []
+    for sls in sources:
+        path = pathlib.Path(sls)
+        if path.is_file():
+            if sls.endswith(".sls"):
```
s0undt3ch, line 45: `path.suffix`
https://docs.python.org/3.5/library/pathlib.html#pathlib.PurePath.suffix

salt/states/idem.py
```diff
@@ -0,0 +1,150 @@
...
+                ref = sls[:-4]
```
s0undt3ch, line 46: https://docs.python.org/3.5/library/pathlib.html#pathlib.PurePath.stem ?
`path.parent / path.stem`

salt/states/idem.py
```diff
@@ -0,0 +1,150 @@
...
+            else:
+                ref = sls
```
s0undt3ch, line 48: Or `srt(path)`

s0undt3ch (changes requested):

salt/states/idem.py
```diff
@@ -0,0 +1,144 @@
...
+    Determine where the sls sources are
+    """
+    sls_sources = []
+    SLSs = []
+    if tree:
+        sls_sources.append("file://{}".format(tree))
+    for sls in sources:
+        path = pathlib.Path(sls)
+        if path.is_file():
+            SLSs.append(str(path.stem if path.suffix == ".sls" else path))
```
s0undt3ch, line 47: I think you want `(path.parent / path.stem) if path.suffix == ".sls" else path`
You don't want to loose the parent path do you?

Akm0d: We do want to lose the parent path in this part.  I actually need to update `path` here to be path.name

salt/modules/idem.py
```diff
@@ -0,0 +1,83 @@
+# -*- coding: utf-8 -*-
+#
+# Author: Tyler Johnson <tjohnson@saltstack.com>
+#
+
+"""
+Idem Support
+============
+
+This module provides access to idem execution modules
```
max-arnold, line 10: The docstring for a new module [should have](https://docs.saltstack.com/en/master/topics/development/conventions/style.html#docstring-conventions) the `.. versionadded:: Magnesium` tag

salt/states/idem.py
```diff
@@ -0,0 +1,147 @@
+# -*- coding: utf-8 -*-
+#
+# Author: Tyler Johnson <tjohnson@saltstack.com>
+#
+
+"""
+Idem Support
+============
+
+This state provides access to idem states
```
max-arnold, line 10: The docstring for a new module [should have](https://docs.saltstack.com/en/master/topics/development/conventions/style.html#docstring-conventions) the `.. versionadded:: Magnesium` tag

salt/states/idem.py
```diff
@@ -0,0 +1,147 @@
...
+"""
+import logging
+import pathlib
+import sys
+
+import salt.utils.idem as idem
+
```
max-arnold, line 17: Maybe utils should be called through the `__utils__` dunder instead of direct imports?

@s0undt3ch What is the recommended convention for that?

max-arnold:
If I understand this correctly, idem is asynchronous, right? How this would work with the planned Tornado upgrade where it is only possible to have a [single event loop per thread](https://github.com/saltstack/salt/pull/57188#issue-415745939)?

Akm0d:
> If I understand this correctly, idem is asynchronous, right? How this would work with the planned Tornado upgrade where it is only possible to have a [single event loop per thread](https://github.com/saltstack/salt/pull/57188#issue-415745939)?

This is executed in a separate process that is isolated from the original tornado loop

max-arnold:
> This is executed in a separate process that is isolated from the original tornado loop

Does idem start this process somewhere, or it is a minion worker process?  (I'm mostly asking this to educate myself, feel free to ignore)

max-arnold:
Can Salt run states and/or execution modules iin the main thread that has an existing event loop?

Akm0d:
re-fun full debian10

Akm0d:
re-run full all

thatch45:
Since this is an execution module it will run inside of a worker process. Idem will run using the available loop that should be present when the worker is spun up.

salt/utils/idem.py
```diff
@@ -0,0 +1,56 @@
...
+"""
+import logging
+import sys
+
+try:
+    import pop.hub
+
+    HAS_POP = True, None
+except ImportError as e:
+    HAS_POP = False, e
```
dhiltonp, line 18: Why not assign `False, str(e)`? All uses of `HAS_POP[1]` do a string conversion.

Just storing `e`, the stack frames and variables leading to this import are forced to persist for the lifetime of the program.

salt/states/idem.py
```diff
@@ -0,0 +1,136 @@
...
+    name,
+    sls,
+    acct_file=None,
+    acct_key=None,
+    acct_profile=None,
+    cache_dir=None,
+    render=None,
+    runtime=None,
+    test=False,
+    tree=None,
```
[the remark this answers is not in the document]
Akm0d: That's following the convention set in `idem`, but a more clear name could be used here

tests/integration/states/test_idem.py
```diff
@@ -0,0 +1,49 @@
...
+state_name:
+  test.succeed_without_changes:
+    - name: idem_test
+    - foo: bar
+"""
+
+
+@skipIf(not idem.HAS_POP[0], str(idem.HAS_POP[1]))
+@skipIf(not salt.utils.path.which("idem"), "idem is not installed")
+@contextmanager
```
s0undt3ch, line 24: Why do you need the context manager decorator?

s0undt3ch (commented):
Just a question about the context manager usage.

tests/integration/states/test_idem.py
Akm0d: The state being tested uses `__context__` and that doesn't seem to get generated properly without the contextmanager decorator

s0undt3ch: Huh?! Really?! ..... Ok....

I'm going to be a jerk down the line and I'll ask you to rewrite this for PyTest. Or I'll do it myself, but don't worry about that now.

salt/states/idem.py
```diff
@@ -0,0 +1,136 @@
...
+        "{}[{}]: {}".format(v.get("name", k), v["result"], v["comment"])
+        for k, v in running.items()
+        if v.get("comment")
+    )
+    return {
+        "name": name,
+        "result": not errors,
+        "comment": errors or comments,
+        "changes": {k: v["changes"] for k, v in running.items() if v.get("changes")},
+    }
```
Akm0d, line 136: I really want some feedback on this bit of code.  `hub.idem.state.apply` runs a whole set of idem idem states then the returns (which look exactly like salt state returns) are stored in `hub.idem.RUNS[name]["running"].  I toyed with a couple ways of returning all of those state returns and in the end decided to concatenate all the comments from all the runs with their idem-state-run name and result and then group all the changes on top of the salt state "changes" return.  Does anyone have any better ideas or does this look OK?

Akm0d: An example of what shows up in  `hub.idem.RUNS[name]["running"]`:
```
{'test_|-state_name2_|-idem_test_|-none_without_changes': {'__run_num': 2,
                                                           'changes': {},
                                                           'comment': 'Success!',
                                                           'name': 'idem_test',
                                                           'result': None},
 'test_|-state_name_|-idem_test_|-none_without_changes': {'__run_num': 1,
                                                          'changes': {},
                                                          'comment': 'Success!',
                                                          'name': 'idem_test',
                                                          'result': None}}
```

Akm0d:
re-run lint

thatch45 (dismissed):

salt/states/idem.py
Akm0d: Maybe a way to do this is to modify states in general so that they can return a list of salt returns.  That way puppet, chef, and ansible calls could return each individual thing they ran as an individual return as well :thinking:

Akm0d: Resolved by #57993

waynew:
I'd really like to do a deeper dive here (and maybe get some of the other more experienced team to review as well). Maybe we could review during your Twitch stream on Monday? That way you could walk us through the changes, your thoughts, any questions that you had as you were putting this together, and maybe we'd come up with questions as well. Or it will all make perfect sense and everyone will be on board 😂

s0undt3ch (changes requested):
This now has conflicts

PR #57974, opened 2020-07-17
Altfix 56635

github-abcde:
### What does this PR do?
It fixes the problems regarding mine.get (more specifically, filling in args with a kwarg with the same name) without modifying `call_function` (and `_call_function`) to accept and pass kwargs where it really shouldn't.
Added tests to indicate that all situations where args get moved to kwargs, kwargs fill in for missing args, kwargs get passed as varargs or varkwargs, too many args, too few args, multiple values for the same (kw)args now either get handled properly, or raise an exception.

Fixes: 59496, 57055, 56584

### Previous Behavior
Supplied single-dict-kwargs did not correctly replace missing args, causing confusion why a function would not execute properly.

### New Behavior
The argument-style of `module.run` is now correctly parsed, args and kwargs can be interchanged to some degree.

- [X] Docs; sort of. Added comments on what is going on in `call_function`.
- [X] Tests written/updated

### Commits signed with GPG?
Yes

Edit: Fixed broken indicator for Tests written/updated


krionbsd (dismissed):

dwoz:
@github-abcde None of the issues you mentioned in the description are valid

github-abcde:
@dwoz If by 'valid' you mean 'open', then indeed. However, that is exactly why this PR was made. The current PR that attempts to fix these issues (56635) does so in a way that is inconsistent with how `call_function` is used and only causes more confusion.
If kwargs do not get processed properly, it is not good to just slap on a `function_kwargs`-argument to fix it, especially if the function is not *supposed* to get kwargs that way (but instead, like `module.run`, as single-parameter dicts in *args).

dwoz:

PR #57977, opened 2020-07-17
Add hpilo proxy minion and execution modules

jasperla:
### What does this PR do?

This module interacts with HP iLO via the network, providing
functionality not covered by the existing salt.modules.ilo
module which requires a minion to run on the host and interfaces
with the iLO via the hponcfg utility. This new module therefore
allows for extra functionality, primarily power management.

- [x] Docs
- [ ] Tests written/updated

### Commits signed with GPG?
Yes

### DIFF
--- a/changelog/57977.added
+++ b/changelog/57977.added
@@ -1,0 +1,1 @@
+Added new hpilo proxy minion and execution modules.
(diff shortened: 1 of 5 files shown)


mchugh19:
Since it contacts ilo endponts over the network, have you given any consideration to running this as a proxy minion? That might make for a slightly more natural interface, and allow contacting multiple ilo devices from a single host.

jasperla:
@mchugh19 Good point, I hadn't considered that. I will adjust it accordingly.

jasperla:

PR #57981, opened 2020-07-17
Jid in state

### LINKED ISSUE #49890: Please add JID to Jinja Context
The job id is not presently available via Jinja during state runs.

Note for starting point from @gtmanfred:
  https://github.com/saltstack/salt/blob/2018.3/salt/modules/state.py#L1053

### Setup
N/A

### Steps to Reproduce Issue
N/A

### Versions Report
New request.

kajigga:
### What does this PR do?
Adds `jid` to renderer context.

Fixes #49890 

### Previous Behavior

`jid` was not available in context.

### New Behavior
the `jid` is now available as a contextual variable. For example, in jinja it can be accessed as `{{ jid }}`. This works in both normal states and orchestration states. In the case of an orchestration state, the variable is the `jid` of the orchestration.

### Commits signed with GPG?
Yes


max-arnold:
Have you considered this comment? https://github.com/saltstack/salt/issues/49890#issuecomment-427033633

Salt can run states through different means:

* `salt MINION state.apply`
* `salt-call state.apply`
* `salt-call --local state.apply`
* `salt-run state.cmd state.apply`
* `salt-run state.orchestrate`
* `salt-ssh HOSTNAME state.apply`
* `salt-ssh` minions (via [`enable_ssh_minions`](https://docs.saltstack.com/en/latest/ref/configuration/master.html#enable-ssh-minions))
* Proxy minions?

How this new `jid` context variable behaves in each case?

Also, there is a commonly used pattern to debug Jinja templates:

* `salt MINION slsutil.renderer default_renderer=jinja path=salt://mystate.sls`
* `salt-call slsutil.renderer default_renderer=jinja path=salt://mystate.sls`

How it will work if a template contains the new `{{ jid }}` variable?

Also a few standard things:

* The new variable needs to be documented [here](https://docs.saltstack.com/en/latest/ref/states/vars.html) (do not forget the [`.. versionadded:: Magnesium`](https://docs.saltstack.com/en/master/topics/development/conventions/style.html#docstring-conventions) tag)
* A changelog entry should be added to the `changelog` folder (see [examples](https://github.com/saltstack/salt/commits/master/changelog))
* Unit-tests

sagetherage:
@waynew maybe we can consider this one for the next Test Clinic on Thursday this week?

dwoz:

PR #57983, opened 2020-07-18
Stop using deprecated cgi.escape

jasperla:
## What does this PR do?

[cgi.escape has been deprecated](https://docs.python.org/3.2/library/cgi.html#cgi.escape) since python 3.2 and removed in 3.8. Its usage should be replaced with `html.escape`.

This issue was originally discovered and fixed in OpenBSD ports by @fobser .

### Previous Behavior

Salt would crash:

```
2020-07-09 08:38:44,604 [salt.utils.schedule                                                   :853 ][ERROR   ][26771] Unhandled exception running stat
e.highstate
 File "/usr/local/lib/python3.8/site-packages/salt/utils/schedule.py", line 844, in handle_func
   self.returners[ret_str](ret)
 File "/usr/local/lib/python3.8/site-packages/salt/returners/highstate_return.py", line 480, in returner
   _produce_output(report, failed, setup)
 File "/usr/local/lib/python3.8/site-packages/salt/returners/highstate_return.py", line 433, in _produce_output
   _generate_html(report, string_file)
 File "/usr/local/lib/python3.8/site-packages/salt/returners/highstate_return.py", line 269, in _generate_html
   _generate_html_table(data, out, 0)
 File "/usr/local/lib/python3.8/site-packages/salt/returners/highstate_return.py", line 220, in _generate_html_table
   _generate_html_table(value, out, level + 1, new_extra_style)
 File "/usr/local/lib/python3.8/site-packages/salt/returners/highstate_return.py", line 229, in _generate_html_table
   cgi.escape(six.text_type(value)),
AttributeError: module 'cgi' has no attribute 'escape'
```

### New Behavior

It no longer crashes.

### Commits signed with GPG?
Yes

### DIFF
--- a/changelog/57983.fixed
+++ b/changelog/57983.fixed
@@ -1,0 +1,1 @@
+Replace deprecated `cgi.escape()` with `html.escape()` after it was removed from Python 3.8.
--- a/salt/returners/highstate_return.py
+++ b/salt/returners/highstate_return.py
@@ -79,7 +79,7 @@ the time of execution.
 """
 from __future__ import absolute_import, print_function, unicode_literals
 
-import cgi
+import html
 import logging
 import smtplib
 from email.mime.text import MIMEText
@@ -226,7 +226,7 @@ def _generate_html_table(data, out, level=0, extra_style=""):
                                 "td",
                                 [cell_style, second_style, "value", new_extra_style],
                             ),
-                            cgi.escape(six.text_type(value)),
+                            html.escape(six.text_type(value)),
                         ),
                         file=out,
                     )
@@ -251,7 +251,7 @@ def _generate_html_table(data, out, level=0, extra_style=""):
                     _lookup_style(
                         "td", [cell_style, first_style, "value", extra_style]
                     ),
-                    cgi.escape(six.text_type(subdata)),
+                    html.escape(six.text_type(subdata)),
                 ),
                 file=out,
             )
--- a/salt/returners/highstate_return.py
+++ b/salt/returners/highstate_return.py
@@ -1,4 +1,3 @@
-# -*- coding: utf-8 -*-
 """
 Return the results of a highstate (or any other state function that returns
 data in a compatible format) via an HTML email or HTML file.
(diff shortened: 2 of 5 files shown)


Akm0d:
re-run full all

Akm0d:
This is merge-ready as soon as a test is written for the changes :+1:

jasperla:
@Akm0d Thanks for having a look -- but what type of tests should be written exactly? This isn't any new functionality but merely fixing a deprecation in Python itself. Unless you're looking for new tests which exercise the current code but which I think is really out of scope for this particular PR, no?

Akm0d:
@jsaperla A smoke test would do the job. Clearly we don’t have any tests testing `_generate_html_table`.
Basically, what should be done is write two tests, one that triggers a failure on each of the bad lines in that block then simply assert that the result data looks the way you expect it to look. 
It should be pretty basic, as long as `_lookup_style` isn’t doing anything sketchy.
I’d write another couple of tests on the nagios return, they wouldn’t have to be much more than:
```
def test_this_thing_produces_correct_output():
    data = some_sort_of_data
    expected_output = "something like this"
    actual_output = module._generate_html_table(data)
    assert actual_output == expected_output
```
4 tests that are roughly that simple, should do it.

waynew:
@jasperla any progress writing tests? If you're struggling with getting started, I can definitely help get you moving in the right direction during my Test Clinic Twitch stream - Tuesday Afternoon, Thursday morning, Central Time - https://www.twitch.tv/saltstackinc/schedule

jasperla:
@waynew Thanks, I'll be sure to join this week.

waynew:
Hey @jasperla I wasn't sure if you were planning today or Thursday - I went ahead and put together some tests during [today's stream](https://www.twitch.tv/videos/700335072) and opened a PR against your branch: https://github.com/jasperla/salt/pull/1 

It should be relatively straightforward to add some tests for the other returner - if you have any problems, just let me know and I'll be happy to help!

waynew (approved):
I'm 👍  on this

One thing that we discovered during the Test Clinic as I was reviewing these changes - cgi.escape uses `quote=False` by default, but html.escape uses `quote=True`, changing `"` to `&quot;`. This seems like a reasonable change to me - especially since the highstate_return is explicitly returning HTML - and `&quot;` will be totally valid in that scenario.

In the nagios returner it's for the hostname, and if you're putting `"` in your hostname, you have bigger problems 🙃  (and escaping to `&quot;` may help reveal them?)

cmcmarrow (approved):

PR #57984, opened 2020-07-18
Fix VMCTL broken module

FinweVI:
### What does this PR do?
VMCTL command line arguments order changed since OpenBSD 6.6.
The changes are described in the upgrade process: https://www.openbsd.org/faq/upgrade66.html
This PR is fixing the arguments order according to the documentation.

Fixes: The VMCTL Salt Module

- [ ] Docs
- [X] Tests written/updated

### Commits signed with GPG?
Yes


salt/modules/vmctl.py
```diff
@@ -257,6 +249,14 @@ def start(
     if disks and len(disks) > 0:
         cmd.extend(["-d", x] for x in disks)
 
+    if not (name or id):
+        raise SaltInvocationError('Must provide either "name" or "id"')
```
jasperla, line 253: I'd suggest keeping this particular check at the original location. This way we can error out early if the `name` or `id` are missing.

The `elif name [..]` can stay where you moved them to.

FinweVI: Thanks for the feedback, just added another commit (#d080992) to fix that.

jasperla: That looks fine to me. Could you please add a changelog entry too?

jasperla:
As the original author of this module I'd like to point out there are no concerns about backwards compat concerning this change. OpenBSD 6.6 has been out for some time now and with having moved to Python 3 one'd need OpenBSD 6.7 anyway.
So this looks like a worthwhile change to me.

salt/modules/vmctl.py
FinweVI: Yep, that's done :)

waynew (changes requested):
Hey @FinweVI thanks for the changes! We've recently added a pre-commit hook that will help us refactor out legacy Python code - would you mind addressing those changes? If you haven't setup pre-commit locally, it's pretty easy to do, and should automatically apply those changes, then we should be able to get this merged in!

FinweVI:
Hello @waynew,
I never used pre-commit before, but this should be fixed with #fe7f4e2 
Let me know if anything else need to be improved.

waynew (approved):
there are currently a couple of test failures, I think we've added some new OS pipelines. I'd trigger a rebuild, but right now there is an upstream issue with setuptools, so there's that.

We should have a workaround in today, and then we will be able to re-run the tests on this.

👍

FinweVI:
It seems all the tests are passing now.
Can we merge it? :)

dwoz:

PR #57993, opened 2020-07-21
Sub state return

Akm0d:
### What does this PR do?
Adds the ability for states to have sub state returns.  For example, if you are running a state that runs several external states under a different engine, you can now add those external state runs individually to the "sub_state_run" key of the state return.  They will be parsed and printed alongside the salt state runs.

- [x] Docs
- [x] Tests written/updated

### Commits signed with GPG?
Yes


Akm0d:
re-run full all

Akm0d:
re-run full archlts

krionbsd (dismissed):

salt/state.py
```diff
@@ -3082,6 +3082,18 @@ def call_chunk(self, low, running, chunks):
...
+                running[tag].pop("sub_state_run", {}).items()
+            ):
+                sub_tag, sub_return = item_data
+                sub_run_num = sub_return.get("__run_num__", i)
+                sub_return["__run_num__"] = float(
+                    "{}.{}".format(running[tag]["__run_num__"], sub_run_num)
+                )
+                sub_return["__sls__"] = running[tag]["__sls__"]
+                running[sub_tag] = sub_return
+
```
waynew, line 3096: Are we intentionally assuming at this point that the data here is well-formed? For instance, `item_data` will always contain a tuple (or, whatever) of data?

tests/unit/test_state.py
```diff
@@ -775,6 +802,138 @@ def test_state_output_unifier_result_converted_to_false(self):
         assert statedecorators.OutputUnifier("unify")(lambda: data)()["result"] is False
 
 
+@skipIf(pytest is None, "PyTest is missing")
```
waynew, line 805: I think that you could just write these as bare functions and get the same effect

tests/unit/test_state.py
```diff
@@ -775,6 +802,138 @@ def test_state_output_unifier_result_converted_to_false(self):
...
+        """
+        data = {"sub_state_run": {"": ["whatever"]}}
+        out = statedecorators.OutputUnifier("content_check")(lambda: data)()
+        assert (
+            "Malformed state return. Data must be a dictionary type"
+            in out["sub_state_run"][""]["comment"]
+        )
+        assert not out["sub_state_run"][""]["result"]
+
+    def test_sub_state_output_check_return_has_nrc(self):
```
waynew, line 834: Can we expand `nrc`? I guess it is expanded in the comment which might be good enough.

doc/ref/states/writing.rst
```diff
@@ -259,6 +259,49 @@ A State Module must return a dict containing the following keys/values:
...
+        "comment": None,  # Comments on the overall external state engine run
+        "sub_state_run": {
+            "{external_module}_|-{state_id}_|-{state_name}_|-{external_module_function}": {
+                "result": None,  # The external state run name
+                "comment": None,  # Comment on the external state run
+                "changes": {},  # A dictionary describing the changes made in the external state run
+                "name": "{state_name}",  # Name of the external state run
+                "__id__": "{state_id}",  # ID of the external state run
+                "__run_num__": None,  # run number of the external state run
+            }
```
Akm0d, line 289: Is there a better way to expect this data to be formatted?  I'm not super familiar with high/low data and analyzing "chunks" from state returns

salt/state.py
Akm0d: yes, item_data will always be a tuple because it is just a result of using `enumerate` while iterating over a dictionary's items.

tests/unit/test_state.py
Akm0d: I just copied the test for states and modified it to verify sub_states.  If both states need this modification I think it should be done separately

krionbsd (dismissed):

thatch45 (dismissed):

waynew:
@Akm0d looks like this one needs pre-commit hooks

waynew:
Also waiting on https://github.com/saltstack/salt-enhancement-proposals/pull/32

twangboy (approved):

Ch3LL (approved):

krionbsd (approved):

PR #57997, opened 2020-07-21
Add ARM64 support for Ubuntu20 pipeline

krionbsd:
### What does this PR do?
Add ARM64 support for Ubuntu20 pipeline


bryceml (approved):
This looks good after we get the kitchen config and jenkins multi-branch pipeline in for it.

bryceml (changes requested):
We need to hard-code the ami, since we're making that change elsewhere.

Ch3LL:
@bryceml can you re-review?

bryceml (dismissed):

Ch3LL:
thanks. @krionbsd looks like there is some ub20 arm64 errors here that need to be cleaned up: https://jenkins.saltproject.io/job/pr-arm64-ubuntu2004-py3-pytest/job/PR-57997/4/#showFailuresLink

garethgreenaway (dismissed):

bryceml:
we decided we can skip the failing tests on arm64 with a link to https://github.com/saltstack/salt/issues/59884

Need to add that, we haven't skipped based on architecture before so we need to figure out the best way to do that.  Most thing use salt.utils.platform but arch isn't in there.

tests/integration/pillar/test_git_pillar.py
```diff
@@ -106,6 +106,10 @@
 HAS_NGINX = bool(salt.utils.path.which("nginx"))
 HAS_VIRTUALENV = bool(salt.utils.path.which_bin(VIRTUALENV_NAMES))
 
+pytestmark = [
```
bryceml, line 109: I don't think we want to skip the whole file, just the SSH pygit2 tests.  Or if we want all the pygit2 tests.  We shouldn't skip the GitPython tests.

And we use the same version of libgit2-dev on x86_64, so it's not that it's outdated, it's just buggy.

bryceml (approved):

twangboy (approved):

Issue #57901, opened 2020-07-10
[BUG] Salt 3001: ä and ü are "non-decodable characters" on Windows


markuskramerIgitt:
**Description**
Letter ä is decoded as `„` (Double Low-9 Quotation Mark, [Unicode letter 201E](https://unicode-table.com/en/search/?q=201E)), 

Letter ü is not decodable (�)

**Setup**
- Install Salt-Minion-3001-Py3-AMD64.msi or
- Install from Git (pip install -e .)

**Steps to Reproduce the behavior**

On Windows, create a file containing the letter ä which occurs in German Datenträger (volume), or set up Windows in German, which will name its volumes "Datenträger".

    mkdir \windows\temp\salt-test
    echo Datenträger> C:\windows\temp\salt-test\Datenträger.txt

On the master execute 

```
salt bob cmd.run 'dir \windows\temp\salt-test'
bob:
     Datentr„ger in Laufwerk C: ist Windows
     Volumeseriennummer: XXXX-XXX

     Verzeichnis von C:\windows\temp\salt-test

    10.07.2020  17:55    <DIR>          .
    10.07.2020  17:55    <DIR>          ..
    10.07.2020  17:51                13 Datentr„ger.txt
                   1 Datei(en),             13 Bytes
                   2 Verzeichnis(se), 328.304.738.304 Bytes frei

```
The letter ä is decoded as double quotation mark in both the file name and the localized word for volume.

On Windows add a file containing the letter ü which occurs in the word ausgeführt (executed), and is therefore often part of the Windows response, e.g. "service is executed":

    echo ausgeführt> C:\windows\temp\salt-test\ausgeführt.txt

On the master execute 

```
salt bob cmd.run 'dir \windows\temp\salt-test'
bob:
     Datentr�ger in Laufwerk C: ist Windows
     Volumeseriennummer: XXXX-XXXX

     Verzeichnis von C:\windows\temp\salt-test

    10.07.2020  17:51    <DIR>          .
    10.07.2020  17:51    <DIR>          ..
    10.07.2020  17:51                12 ausgef�hrt.txt
    10.07.2020  17:51                13 Datentr�ger.txt
                   2 Datei(en),             25 Bytes
                   2 Verzeichnis(se), 328.305.393.664 Bytes frei

```

ä and ü are not decoded

With salt-call on Windows, you get the same output and an error:

```
\salt\salt-call  cmd.run "dir \windows\temp\salt-test"
[...]

PR #57515, opened 2020-06-01
Py3 unicode

twangboy:
### What does this PR do?
Adds some tests for Unicode Support with Py3
Adds the ability to pass file content to the `@with_tempfile` helper
Adds the ability to pass the code page to use with run commands on Windows (`chcp`)

### What issues does this PR fix or reference?
Ref https://github.com/saltstack/salt/pull/56028
fixes https://github.com/saltstack/salt/issues/54447
fixes https://github.com/saltstack/salt/issues/36031

### Merge requirements satisfied?
**[NOTICE] Bug fixes or features added to Salt require tests.**

- [x] Docs
- [x] Changelog - https://docs.saltstack.com/en/master/topics/development/changelog.html
- [x] Tests written/updated

### Commits signed with GPG?
Yes

Please review [Salt's Contributing Guide](https://docs.saltstack.com/en/master/topics/development/contributing.html) for best practices.

See GitHub's [page on GPG signing](https://help.github.com/articles/signing-commits-using-gpg/) for more information about signing commits with GPG.

PR #58008, opened 2020-07-22
let_users_change_the_code_page

cmcmarrow:
What does this PR do?
Lets cmd/powershell go to utf-8
#57901
#57515

Commits signed with GPG?
Yes


markuskramerIgitt:
Please remove lines 607-610 that prepend `chcp 437 > nul &` to the command 
```
        else:
            # On Windows set the codepage to US English.
            if python_shell:
                cmd = "chcp 437 > nul & " + cmd

```

image

markuskramerIgitt:
Function chcp() returns a string, chcp_code is an integer.
    
    chcp(chcp_code) != chcp_code

is always true

Please change to 

    chcp(chcp_code) != str(chcp_code)

markuskramerIgitt:
As negative test, I replaced chcp_code=65001 by chcp_code=437 and got the double quotes instead of ä

markuskramerIgitt:
You find my modified cmdmod.py here:
https://github.com/markuskramerIgitt/LearnSalt/tree/master/patch_salt_minion_3001/modules

markuskramerIgitt:
@cmcmarrow looks good for me

Could you please take over my changes, I would like to mass rollout a version we agree upon.

cmcmarrow:
@markuskramerlgitt I’ll take over your changes and add this feature and add it to the startup of salt.

DmitryKuzmenko (dismissed):

markuskramerIgitt:
@cmcmarrow I patched my 3001 Minions with this

salt/utils/chcp.py
```diff
@@ -0,0 +1,59 @@
...
+            if raise_error:
+                raise CodePageError()
+            else:
+                log.error("Code page failed to change to %s!", page_id)
+
+    if chcp_ret == "":
+        chcp_ret = chcp()
+
+    log.debug("Code page is %s", chcp_ret)
+    return chcp_ret
```
twangboy, line 59: Can I assume this entire file was replaced by `win_chcp.py`?

twangboy: And the test with it?

twangboy (changes requested):

salt/utils/chcp.py
markuskramerIgitt: Yes, @twangboy you assume correctly:    test/chcp.py should be replaced by utils/win_chcp.py
It seems both files exist now, how strange.

Also the test should be replaced, not copied.

markuskramerIgitt: Yes, @twangboy you assume correctly: test/chcp.py should be replaced by utils/win_chcp.py
It seems both files exist now, how strange.

Also the test should be replaced, not copied.

twangboy (approved):

DmitryKuzmenko (approved):

s0undt3ch:
This needs to be partially reverted.
It was correct to use `chcp xxx & cmd`, we want to tweak the code page of the command we're going to run, but not necessarily, all other commands after that one.

As for changing code page when salt starts, seen in a comment but not in any code. The admin should be responsible for properly setting his machine, and, if we already support changing the code page to a command we want to run, we're good, by we, I mean salt.

We could check the current code page in order not to chain the `chcp` code page when running the command, but that means an extra shell out per command.

I took a stab at trying to reuse this code, the right way, setting code page, running command, resetting codepage, but that meant 3 shells to run a single command, not good.

Thoughts? Alternatives? Why did we actually needed to shell out to set the code page in the first place and opposed to chaining the commands?

s0undt3ch:
Actually, now that I think about this a bit more, even chaining the chcp command will change the code page for the following commands.....
We should instead do something like `chcp xxx > nul & the actual command & chcp previous_xxx > nul`
How can we grab the current `chcp` in a variable to reset at the end? Using command chaining....

twangboy:
@s0undt3ch There was a lot of discussion about why we did this here: https://github.com/saltstack/salt/pull/57515

s0undt3ch:
Thanks @twangboy, read it.
I still think the merged PR, when setting the codepage, should set it back to how it was.
The system should be properly configured in the first place. At most, our windows bat scripts should do the codepage change at the start of each of our CLIs, not salt.
Anyway, https://github.com/saltstack/salt/pull/58478 seems to correct this behavior, but I'll try testing with the data from the issue you linked.

markuskramerIgitt:
For years Salt was setting code page 437

For year  the codepage, was not set back to how it was.

A change to set it back to how it was should be optional, opt-in, on
request.

I probably sit on the largest minion basis for non 437 Windows boxes, and I
am worried.

I suggest we change the default code page to 437 as it was for years, and
make 65001 an opt-in

PR #58013, opened 2020-07-23
Fix __mount_device wrapper

### LINKED ISSUE #58012: [BUG] Exception when creating a btrfs submodule
**Description**

When creating a btrfs submodule using the btrfs state, the minion produces an exception inside the __moint_device decorator.

aplanas:
### What does this PR do?

Some recent change in Salt is now doing the right thing, and calling the
different states with separated args and kwargs.  This change trigger a
hidden bug in the __mount_device decorator, that expect those parameter
to be in kwargs, as is happening during the test.

This patch change the way that the wrapper inside the decorator search
for the name and device parameters, first looking into kwargs and later
in args if possible.  A new test is introduced to exercise both cases.

Fix #58012

- [ ] Docs
- [X] Tests written/updated

### Commits signed with GPG?
No

### DIFF
--- a/changelog/58012.fixed
+++ b/changelog/58012.fixed
@@ -1,0 +1,1 @@
+Fix btrfs state decorator, that produces exceptions when creating subvolumes.
--- a/salt/states/btrfs.py
+++ b/salt/states/btrfs.py
@@ -1,12 +1,9 @@
-# -*- coding: utf-8 -*-
-
 """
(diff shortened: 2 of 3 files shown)


garethgreenaway:
re-run all

krionbsd (dismissed):

aplanas:
Ping?

krionbsd (approved):

PR #53844, opened 2019-07-12
Fix executors not being called for multi-jobs

andzn:
### What does this PR do?

Fixes a bug where custom executors were not being called for multi-jobs. This PR is based on 2019.2 and also includes a test. There's another [PR](https://github.com/saltstack/salt/pull/52939) based on the 2018.3 branch that doesn't contain a test, because the base class for executors integration tests was only added in 2019.3.

### Previous Behavior
Executors were not being called for multi-jobs

### New Behavior
Executors are now called for multi-jobs as well as for single function jobs

### Tests written?
Yes

### Commits signed with GPG?

No

PR #58040, opened 2020-07-28
Fix multijobs executors

andzn:
### What does this PR do?

Fixes a bug where custom executors were not being called for multi-jobs. This change was already approved once (a year ago) in this [PR](https://github.com/saltstack/salt/pull/53844) but for some reason it didn't got merged. I rewrote the change based on master this time.

### Previous Behavior
Executors were not being called for multi-jobs

### New Behavior
Executors are now called for multi-jobs as well as for single function jobs

### Tests written?
Yes

### Commits signed with GPG?

No


andzn:
All the changes in commit 429577e0939e7491f66703d71fdfafcf907269b6 were done automatically by pre-commit.

andzn:
@dwoz I have no ideea why the blackout_tests hang intermittently on different machines. I've tried to reproduce the issue locally and couldn't. Also out of 3 runs, once it failed on debian9, once on fedora and ubuntu1604 and now they seem to hang on ubuntu 2004.

I tend to think that the hang is not related to my changes. It's just that those tests are automatically picked for my changes since I touched the blackout logic. 

On the other side, is there some other way I could try to reproduce the issue locally?

krionbsd (approved):

PR #58042, opened 2020-07-28
Port #49063 to master

nicholasmhughes:
### What does this PR do?
Port #49063 to master

Fixes: N/A

- [x] Docs
- [x] Tests written/updated

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/cloud/clouds/azurearm.py
+++ b/salt/cloud/clouds/azurearm.py
@@ -34,6 +34,9 @@ The Azure ARM cloud module is used to control access to Microsoft Azure Resource
       * ``client_id``
       * ``secret``
 
+    if using Managed Service Identity authentication:
+      * ``subscription_id``
+
     Optional provider parameters:
 
     **cloud_environment**: Used to point the cloud driver to different API endpoints, such as Azure GovCloud. Possible values:
@@ -260,6 +263,14 @@ def get_configured_provider():
             ("subscription_id", "username", "password"),
         )
 
+    if provider is False:
+        # check if using MSI style credentials...
+        provider = config.is_provider_configured(
+            __opts__,
+            __active_provider_name__ or __virtualname__,
+            required_keys=("subscription_id",),
+        )
+
     return provider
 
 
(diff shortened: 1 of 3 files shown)


salt/cloud/clouds/azurearm.py
```diff
@@ -260,6 +263,14 @@ def __is_provider_configured(opts, provider, required_keys=()):
...
         )
 
+    if provider is False:
+        # check if using MSI style credentials...
+        provider = config.is_provider_configured(
+            __opts__,
+            __active_provider_name__ or __virtualname__,
+            required_keys=("subscription_id",),
+        )
+
```
waynew, line 273: From your question in Slack - we should probably refactor this code. But at minimum, we need to test:

1. if provider is found by `("subscription_id", "tenant", "client_id", "secret")`, that provider is returned
2. if provider is *not* found that way, but is found via `("subscription_id", "username", "password")`, it is returned
3. if provider is not found either of those ways, then whatever is returned from `config.is_provider_configured(...)` is returned.

Presumably that's False, but it would be more consistent and expected for these things to return None than False.

salt/cloud/clouds/azurearm.py
```diff
@@ -301,10 +312,12 @@ def get_conn(client_type):
...
-    else:
-        username = config.get_cloud_config_value(
-            "username", get_configured_provider(), __opts__, search_global=False
-        )
+
+    username = config.get_cloud_config_value(
+        "username", get_configured_provider(), __opts__, search_global=False
+    )
+
+    if username:
```
waynew, line 320: This one should be much more straightforward - there should be a test that returns none for username and enforces that neither username or password was passed to get_client. And then a test that ensures that if the username was returned that whatever username/password values were provided *were* passed to get_client.

Akm0d (dismissed):

salt/cloud/clouds/azurearm.py
waynew: @nicholasmhughes Excellent job! I think we do still need the other test to be able to merge this, but that should be much more straightforward? 😂

waynew (changes requested):
This looks great! I think just adding a changelog entry, perhaps `49063.added` with `Added support for MSI-style authentication for azurearm.` would be reasonable?

Thanks!

nicholasmhughes:
@waynew unfortunately, that message would be misleading. the main reason for the original PR has since been merged for a while... I basically just ported over some additional housekeeping that got left behind. after all that, the main thing that's been added was tests... :laughing:  I don't even know what to call this change...

waynew:
Well, it *does* have the `else` case, so maybe it's just a fixed? 🤷

nicholasmhughes:
:shrug: took a stab at it

waynew (approved):
Good enough for me! That's at least enough information for someone reading the changelog to be able to say, "Ah, yeah, I need to know more about that". Which is the point of the changelog 😂 

🚀

PR #58054, opened 2020-07-28
[PR port] Port #49737 - Isbm pkg resources crash

roaldnefs:
PR port #49737
- [ ] Docs
- [ ] Changelog
- [ ] Tests written/updated

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/states/pip_state.py
+++ b/salt/states/pip_state.py
@@ -150,7 +150,11 @@ def __virtual__():
     """
     Only load if the pip module is available in __salt__
     """
-    return "pip.list" in __salt__ and __virtualname__ or False
+    if not HAS_PKG_RESOURCES:
+        return False, "The pkg_resources python library is not installed"
+    if "pip.list" not in __salt__:
+        return False, "pip.list is not available"
+    return __virtualname__
 
 
 def _fulfills_version_spec(version, version_spec):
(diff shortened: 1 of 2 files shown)


salt/states/pip_state.py
```diff
@@ -144,11 +150,7 @@ def __virtual__():
     """
     Only load if the pip module is available in __salt__
     """
-    if HAS_PKG_RESOURCES is False:
-        return False, "The pkg_resources python library is not installed"
-    if "pip.list" in __salt__:
-        return __virtualname__
-    return False
+    return "pip.list" in __salt__ and __virtualname__ or False
```
twangboy, line 153: We prefer explicit failures here... something like this:
```
if not HAS_PKG_RESOURCES:
    return False, "The pkg_resources python library is not installed"
if not "pip.list" in __salt__:
    return False, "pip.list is not available"
return __virtualname__
```
Only return the virtualname if all conditions are met. When conditions aren't met, return False, and why it was not met.

twangboy (changes requested):
Just one change requested.

salt/states/pip_state.py
roaldnefs: Thanks @twangboy for the feedback! Fixed in 328b457.

dwoz:

PR #58066, opened 2020-07-29
Update url link in documentation in vault.py

unix196:
Old link (http://docs.python-requests.org/en/master/user/advanced/#ssl-cert-verification) in vault doc no longer exist. Need specify fresh link.

### Commits signed with GPG?
No


cmcmarrow (changes requested):
@unix196 thanks for the PR! If you need help with pre-commit please read https://pre-commit.com/. Also this is a doc change so you don’t need tests! 😊

cmcmarrow (approved):

PR #58070, opened 2020-07-29
[merge jam] 48812 - Use Adler32 algorithm to compute server id

waynew:
# What does this PR do?

Ports #48812

In Python3 the generated server_id has changed due to the way that Salt
seeds the hash function. Using the Adler32 instead of PYTHONHASHSEED=0
allows us to avoid a shell call.

This is an alternative to #46649

### Previous Behavior

Had to shell out with PYTHONHASHSEED to get the correct server ID hash

### New Behavior

Uses Adler32 instead of CRC32 to generate the server hash

- [ ] Docs
- [ ] Tests written/updated

TODO: add docs/changelog/tests, if we really do want this functionality.

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/config/__init__.py
+++ b/salt/config/__init__.py
@@ -1,10 +1,8 @@
-# -*- coding: utf-8 -*-
 """
 All salt configuration loading and defaults should be in this module
 """
 
 # Import python libs
-from __future__ import absolute_import, generators, print_function, unicode_literals
 
 import codecs
 import glob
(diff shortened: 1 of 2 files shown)


max-arnold:
@waynew Isn't https://github.com/saltstack/salt/pull/57415 supersedes the PR that you ported?

waynew:
> @waynew Isn't #57415 supersedes the PR that you ported?

🤦 yup, that'd be it

PR #58086, opened 2020-07-30
[PR port] Port #51382 - Don't call zypper with more than one --no-refresh

piterpunk:
PR port #51382
- [X] Changelog
- [X] Tests written/updated

### DIFF
--- a/changelog/51382.fixed
+++ b/changelog/51382.fixed
@@ -1,0 +1,1 @@
+Ensure that zypper is called with only one --no-refresh parameter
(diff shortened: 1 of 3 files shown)


piterpunk:
All the failed tests seems to fail on unit.states.test_network.NetworkTestCase.test_managed, which is not part of this PR.

DmitryKuzmenko (dismissed):

DmitryKuzmenko (dismissed):

piterpunk:
@sagetherage I guess this PR slipped to the oblivion (even I forgot of it until hit by the bug again) as it is an old one "port-to-master" but the problem with double `--no-refresh` is still in `3003rc1`

Ch3LL (approved):

PR #58088, opened 2020-07-30
Disable tests on FreeBSD

krionbsd:
### What does this PR do?
Disable tests on FreeBSD

### DIFF
--- a/tests/integration/pillar/test_git_pillar.py
+++ b/tests/integration/pillar/test_git_pillar.py
@@ -1,4 +1,3 @@
-# -*- coding: utf-8 -*-
 """
 Integration tests for git_pillar
 
@@ -65,7 +64,6 @@ https://github.com/unbit/uwsgi/commit/ac1e354
 """
 
 # Import Python libs
-from __future__ import absolute_import, print_function, unicode_literals
 
 import random
 import string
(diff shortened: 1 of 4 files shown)


waynew:
Why do we want to disable these tests on FreeBSD? Are they just irrelevant on that platform?

krionbsd:
@waynew exactly, inotify/watchdog have very limited support on FreeBSD at this moment. I temporary disabled shell tests as FreeBSD uses csh as root shell.

waynew (changes requested):
Should we change the message to indicate that? Like`"Skip on FreeBSD - does not yet have inotify/watchdog support"`?

krionbsd:
@waynew that's a bit complicated, FreeBSD has kind of limited inotify support, it doesn't support working with symlinks, as well as directory notifications with `IN_MOVED_FROM` / `IN_MOVED_TO`, and some open/close notifications like `IN_OPEN` / `IN_CLOSE_NOWRITE`. Maybe the skipif like `Skip on FreeBSD - does not yet have full inotify/watchdog support` ?

waynew (approved):
👍 looks great, thanks!

I like to have a minimum, but complete amount of information available - that way when future Wayne/Kirill comes looking into these tests after we've learned that there's good inotify support on FreeBSD that we can say, "Oh! We need to re-enable these tests!"

Otherwise we just assume that they need to be skipped forever, or have to dig into why the heck we skipped them in the first place 🙃

PR #58095, opened 2020-07-30
master function unable to detect connection to master on ipv6 address

trootnev:
### What does this PR do?
If minion and master are connected only using ipv6 addresses then this function always returns false marking minion disconnected. This causes minion stopping returning any results back to master after 1 hour because by default once in an hour the scheduled job calls state.master.
The root cause is that the function is using netstat command output with parameter -n -p TCP. It shows only IPv4 connections. To see both IPv4 and IPv6 we need to omit -p parameter completely and filter out non-TCP connections afterwards. Also we need to remove square brackets in IPv6 addresses returned as we will compare those to the ones without brackets later in code.

Fixes:

### Previous Behavior
Every time the master function run on windows minion with only IPv6 address connected to salt-master it (function) used to conclude that minion is disconnected and responses were not sent since that moment until minion restart.

### New Behavior
Minion connection status gets evaluated correctly on both IPv6 and IPv4 addresses

No. Sorry, Windows environment is used, test setup is too complicated for Windows. This change is very easy to check and I checked it in my test environment manually.

- [ ] Docs
- [ ] Tests written/updated

### Commits signed with GPG?
No

### DIFF
--- a/salt/modules/win_status.py
+++ b/salt/modules/win_status.py
@@ -553,7 +553,7 @@ def master(master=None, connected=True):
             remote_host, remote_port = chunks[2].rsplit(":", 1)
             if int(remote_port) != port:
                 continue
-            remotes.add(remote_host.replace('[','').replace(']',''))
+            remotes.add(remote_host.replace('[', '').replace(']', ''))
         return remotes
 
     # the default publishing port
--- a/salt/modules/win_status.py
+++ b/salt/modules/win_status.py
@@ -539,7 +539,7 @@ def master(master=None, connected=True):
         remotes = set()
         try:
             data = subprocess.check_output(
-                ["netstat", "-n", "-p", "TCP"]
+                ["netstat", "-n"]
             )  # pylint: disable=minimum-python-version
         except subprocess.CalledProcessError:
             log.error("Failed netstat")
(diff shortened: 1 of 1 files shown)


trootnev:
something's clearly wrong with those tests... Linux-related tests are failing on something unrelated to what has been changed and the module itself is aiming windows systems only.

cmcmarrow (changes requested):
@trootnev thanks for the PR! If you need help with pre-commit please read https://pre-commit.com/. Also this PR still needs a test.

salt/modules/win_status.py
```diff
@@ -539,21 +539,21 @@ def _win_remotes_on(port):
...
         for line in lines:
-            if "ESTABLISHED" not in line:
+            if "ESTABLISHED" not in line or "TCP" not in line:
                 continue
             chunks = line.split()
             remote_host, remote_port = chunks[2].rsplit(":", 1)
             if int(remote_port) != port:
                 continue
-            remotes.add(remote_host)
+            remotes.add(remote_host.replace('[', '').replace(']', ''))
```
waynew, line 556: **question** would it be a good idea to use the [`ipaddress` module](https://docs.python.org/3/library/ipaddress.html) here, or is that overkill?

garethgreenaway:
@trootnev Following up on this one and checking that you saw the comments above, as well as the request for a automated test.  Thanks!

dwoz:

PR #58140, opened 2020-08-05
Remove security advisor now that we depend on PyCryptodomex

s0undt3ch:
### What does this PR do?
See title

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -34,14 +34,6 @@ documentation.
 
 `<https://docs.saltstack.com/en/latest/>`_
 
-Security Advisory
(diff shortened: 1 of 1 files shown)


sagetherage:
I do think simply removing this is the answer, only asking if we want to say anything about it briefly or maybe point to the detailed HTML docs? Maybe not, you tell me. We are looking to update the readme file a bit in this release so at that time we can also review any additions. thank you!!

waynew (approved):
I guess it depends on whether or not we support pycrypto still

PR #58155, opened 2020-08-07
Unbreak swap facts on OpenBSD

jasperla:
### What does this PR do?

It fixes two issues:
- on systems without swap the minion would crash attempting to gather the grains:
```
      File "/usr/local/lib/python3.8/site-packages/salt/grains/core.py", line 534, in _bsd_memdata
            grains["swap_total"] = int(swap_total) // 1024 // 1024
            ValueError: invalid literal for int() with base 10: 'no'
```
- additionally the `swap_total` fact was broken for systems which did have swap as it was divided by 1024 too many.

### Previous Behavior

Minion crash or incorrect `swap_total`.

### New Behavior

No more crash and correct `swap_total`.

- [x] Docs
- [x] Tests written/updated

### Commits signed with GPG?
Yes


jasperla:

PR #58161, opened 2020-08-10
Disable locale related test on FreeBSD

krionbsd:
### What does this PR do?
Disable locale related test on FreeBSD. locale method is supported only within login classes or environment variables.


DmitryKuzmenko (approved):

krionbsd:
re-run all

PR #58192, opened 2020-08-13
hard-code ami ids

bryceml:
### What does this PR do?
This changes to hard-coded amis, this reduces the complexity of introducing new batches of golden images.  It also makes it easier for cve releases to use the proper amis they used when the branch was active.


krionbsd (dismissed):

s0undt3ch (dismissed):

Ch3LL (dismissed):

s0undt3ch (approved):

Ch3LL (approved):
just a thought, but is there a way to verify that these are indeed the OSs we intend on testing. I could see a mistake being made in the future where someone updates the ami but its for the wrong OS and we are not aware.

bryceml:
> just a thought, but is there a way to verify that these are indeed the OSs we intend on testing. I could see a mistake being made in the future where someone updates the ami but its for the wrong OS and we are not aware.

it wouldn't be hard to write a script that for the ami in each file, checks the ami name of each of them using the aws cli locally before committing.


```bash
IFS=$'\n'
for i in `grep -r ami-`; do echo $i; aws ec2 describe-images --image-id $(echo $i|cut -d\' -f2) --query 'Images[].ImageLocation' --output text;  done
```

bryceml:
Megan and I talked and we should add a check, probably part of init_environment https://gitlab.com/saltstack/open/cicd/kitchen-conf/-/blob/master/nox-platforms.yml#L69

or sometime close to the kitchen converge that makes sure we're running the right os.

krionbsd (approved):

s0undt3ch:
We should a really use a pre-commit hook, even if we only run it on PRs?

bryceml:
yeah, that could work well once we move to the new images that are named correctly, then we could problematically make sure the names match up.

The names like 540082622920/saltstack/ci/ubuntu/xenial.1604/master/1573754156
would be annoying to work around.  

After the next batch of golden images are ready to go, we should talk about adding a pre-commit job.

Issue #58114, opened 2020-08-03
[FEATURE REQUEST] Support checking/modifying the return of state.module.run


Timbus:
**Problem**
Today I was struggling to use the win_service.modify command from state.module.run in a way that would allow for some result-based followup actions. I ended up writing a custom state but I feel like it should have been possible without needing to use python.

**Solution**
If state.module.run offered an interface to allow for checking return values and using them to signal success/change/failure, I could have called win_service.info and checked the values, then chained the state using `onchange`. 
Strangely enough, loop.until_no_eval seems to offer something like this with the compare operator? But it's hardly fit for purpose.

**Alternatives + SEP thoughts**
Guess I'll also mention that exploring through other options like slots (yuck), state.cmd+`stateful`, and most of the requisites, seems to show there is a need for runtime variable-passing and logical operators (yes like registers from ansible), and having such a system would probably simplify a lot of states? Even being able to use thorium + custom events in my state mostly does the job. Dunno, just musing at this point.

garethgreenaway:
@Timbus Thanks for the report.  Can you elaborate a bit more on what exactly you're looking to do?  Do you want to capture the results from state.module.run and run something based on that?  Is this within the same state run or multiple state runs?

Timbus:
Well, this is a bit of an 'XY' problem I guess. 
For my very specific use case, problem 'X', I wanted the ability to manipulate/check the return value from module.run and use that value to possibly trigger an onchange for the next state. In short, it would allow building custom states without needing to write python.
I'm not a big fan of the design of the new `module.run` as it's not very extensible, but I used it as a guide to come up with something along the lines of:
```yaml
      some_id_of_state:
        module.run_with_check:
          - service.get_info:
            - name: salt-minion
          - tests:
            # Return whatever matches first. If none match return Changed, like module.run would
            - unchanged:
                compare_operator: data.subdict_match  # stolen from the 'loop' module
                expected: 'ServiceAccount:SaltyAdmin'
            - changed:
                compare_operator: data.subdict_match
                expected: 'ServiceAccount:LocalAdmin'
            - error:
                compare_operator: data.subdict_match
                expected: 'Startup:None'
```
Though it might be easier to just make a `state`.`function` module for this lower-level stuff.

[...]

PR #58198, opened 2020-08-14
[Feedback wanted] Add basic thorium register interpolation

Timbus:
### What does this PR do?
This is a quick-n-dirty, but pretty effective implementation of being able to access thorium registers when invoking a 'local.cmd' from thorium.
Without this feature thorium is fairly impotent at reacting to events compared to the default reactor, which is a shame because it is (imo) a better framework for building reactive components.
I tried to follow the 'slots' convention but its just so _extra_ in its implementation and syntax, I couldn't do it. I just let jinja do the work instead.
I am willing to add this feature to other thorium modules. I'm also very open to change suggestions.

Do I think this is the best approach? Probably not. If thorium .sls files were compiled per-run this change would be unneeded. And if the register concept were more global (hint hint) then this would feel like less of a bespoke system tacked onto salt. Another approach might be to have thorium have separate runtime and compiletime .sls files, but that's a huge change.

While I have your attention: I also think registers in thorium are gimped by having a 'val', 'count' and 'total' forced into them for every usage. It's just such an insanely specialized case that detracts from their general purpose nature. Why not only add a count+total when needed?

This is related to #58114 , though only slightly. 

### New Behavior
local.cmd now runs 'tgt', 'func', 'args' and 'kwarg' through jinja before passing them onto LocalClient.cmd_async

- [x] Docs
- [ ] Tests written/updated

### Commits signed with GPG?
No


DmitryKuzmenko:
@Timbus could you please fix the `lint` and `pre-commit` issues? Also this change needs to be covered with some tests.
@saltstack/team-core I'm basically good with this change but could anyone please also take a look at this code too?

Timbus:
I think I corrected the issues as reported, but I guess something else is missing.

After thinking on it a bit, I was wondering if the use of NativeEnvironment like this would be acceptable? Should it be put into the template util? Also, is the bracket usage acceptable? There could be better alternatives that more experienced python folks might know about.

thatch45:
This is a fantastic feature, I love it. But it is a one-off language change, and we need to be really careful there. We will also need more docs for it and tests.
I am wondering if there is another way to process the jinja, and perhaps process the jinja via the renderer sequence.
In a nutshell, I love this, but I think we need to figure out a better way to make it happen

Timbus:
> This is a fantastic feature, I love it. But it is a one-off language change, and we need to be really careful there. We will also need more docs for it and tests.

Awesome to hear, honestly. My goal was to pursue a feature that would also see utility if ported to states and orchestrations, because it ultimately allows functions to pass variables, a sore spot for me and my team as we get used to this tool.
So though it is one-off right now, I'd like to think of it as something that can be expanded.

> I am wondering if there is another way to process the jinja, and perhaps process the jinja via the renderer sequence.

The jinja NativeEnvironment seems to be a perfect way to deal with the difficulties in parsing and typing, so I was hoping this would just be a case of finding a delimiter that works for everyone..
How would you go about preparsing it? Something that takes the {| |} a placeholder at compiletime and mangles it into some long delimiters? I don't mind that idea but it forces a new renderer to be used, and I was hoping it would work with the default setup.
.. Would it be better to use a flag? A named arg that you use to 'inject' the named variables into your states, if they need them? Then you don't need to brute-force through so many jinja calls as well.
Or, how about something similar to python's new f"formatted {string}s" ? Again it narrows down the things we filter through jinja.

Thoughts on those two options?

> In a nutshell, I love this, but I think we need to figure out a better way to make it happen

:+1:

dwoz:

PR #58213, opened 2020-08-17
[PR Port] Port #53882 - Fixed Bug LVM has no Filesystem Type , no Filesystem is not allowed

### LINKED ISSUE #54456: partition.mkpart doesn't work without fstype making it useless to create partitions without filesystem
### Description of Issue
The module partition.mkpart doesn't work without fstype making it useless to create partitions without filesystem (needed for example for LVM)

### Steps to Reproduce Issue
call partition.mkpart without giving a fstype always return the error "Invalid fs_type passed to partition.mkpart"

### Versions Report
Salt Version:
Salt: 2019.2.0

Dependency Versions:
cffi: 1.11.5
cherrypy: Not Installed
dateutil: 2.8.0
docker-py: Not Installed
gitdb: Not Installed
gitpython: Not Installed
ioflo: Not Installed
Jinja2: 2.10
libgit2: Not Installed
libnacl: Not Installed
M2Crypto: 0.30.1
Mako: Not Installed
msgpack-pure: Not Installed
msgpack-python: 0.6.1
mysql-python: Not Installed
pycparser: 2.19
pycrypto: 2.6.1
pycryptodome: Not Installed
pygit2: Not Installed
Python: 3.5.3 (default, Sep 27 2018, 17:25:39)
python-gnupg: Not Installed
PyYAML: 3.13
PyZMQ: 18.0.2
RAET: Not Installed
smmap: Not Installed
timelib: Not Installed
Tornado: 4.5.3
ZMQ: 4.3.1

System Versions:
dist: debian 9.9
locale: UTF-8
machine: x86_64
release: 4.9.0-9-amd64
system: Linux
version: debian 9.9

In the code fstype is initialized to None, but then is checked against _is_fstype before checking if it's not None, which results in returning always and error
```python
if not _is_fstype(fs_type):
        raise CommandExecutionError(
            'Invalid fs_type passed to partition.mkpart'
        )
```

piterpunk:
### What does this PR do?
Port #53882 to master:

- Allow the creation of a partition without filesystem (for LVM, ASM, etc)
- Added unit tests to check if the mkpart allows the empty fs_type case.

Removed mkpartfs:

- Removed mkpartfs which was exactly the same function of mkpart, with an slightly different code which calls the same parted command with the same parameters. It's not used by other execution or state modules.

Fixes: #54456

### Previous Behavior
An error when calling mkpart without fs_type specified

### New Behavior
When mkpart is called without a fs_type, the partition is created. 

- [x] Tests written/updated

### Commits signed with GPG?
No

### DIFF
--- a/salt/modules/parted_partition.py
+++ b/salt/modules/parted_partition.py
@@ -509,6 +509,8 @@ def mkpart(device, part_type, fs_type=None, start=None, end=None):
         salt '*' partition.mkpart /dev/sda primary fs_type=fat32 start=0 end=639
         salt '*' partition.mkpart /dev/sda primary start=0 end=639
     """
+    _validate_device(device)
+
     if part_type not in {"primary", "logical", "extended"}:
         raise CommandExecutionError("Invalid part_type passed to partition.mkpart")
 
--- a/salt/modules/parted_partition.py
+++ b/salt/modules/parted_partition.py
@@ -545,6 +545,22 @@ def mkpart(device, part_type, fs_type=None, start=None, end=None):
     return out
 
 
+def mkpartfs(device, part_type, fs_type=None, start=None, end=None):
+    """
+    The mkpartfs actually is an alias to mkpart and is kept for compatibility.
+    To know the valid options and usage syntax read mkpart documentation.
+
+    CLI Examples:
+
+    .. code-block:: bash
+
+        salt '*' partition.mkpartfs /dev/sda primary fs_type=fat32 start=0 end=639
+        salt '*' partition.mkpartfs /dev/sda primary start=0 end=639
+    """
+    out = mkpart(device, part_type, fs_type, start, end)
+    return out
+
+
 def name(device, partition, name):
     """
     Set the name of partition to name. This option works only on Mac, PC98, and
--- a/salt/modules/parted_partition.py
+++ b/salt/modules/parted_partition.py
@@ -1,4 +1,3 @@
-# -*- coding: utf-8 -*-
 """
 Module for managing partitions on POSIX-like systems.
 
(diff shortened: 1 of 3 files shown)


salt/modules/parted_partition.py
```diff
@@ -542,39 +545,6 @@ def mkpart(device, part_type, fs_type=None, start=None, end=None):
     return out
 
 
-def mkpartfs(device, part_type, fs_type, start, end):
```
Ch3LL, line 545: We cannot remove this function, because a user could be using this even if its not being called from another state/execution module. Could you possibly have this function call the other to ensure backwards compatibility?

Ch3LL (changes requested):

salt/modules/parted_partition.py
piterpunk: Of course. I will do that tonight.

piterpunk: Re-added the mkpartfs which now is just a call to mkpart. Also added a test to check if all arguments from mkpartfs are reaching mkpart.

salt/modules/parted_partition.py
```diff
@@ -542,36 +545,19 @@ def mkpart(device, part_type, fs_type=None, start=None, end=None):
...
-    CLI Example:
+    CLI Examples:
 
     .. code-block:: bash
 
-        salt '*' partition.mkpartfs /dev/sda logical ext2 440 670
+        salt '*' partition.mkpartfs /dev/sda primary fs_type=fat32 start=0 end=639
+        salt '*' partition.mkpartfs /dev/sda primary start=0 end=639
     """
-    _validate_device(device)
```
Ch3LL, line 560: thanks for updating this. looks like this is the difference between the two. Is there a reason we dont want to add this to `mkpart`?

Ch3LL (changes requested):
just one more question here, everything else looks great thanks

salt/modules/parted_partition.py
piterpunk: No reason at all. I just re-added the _validate_device and updated the tests accordingly.

Ch3LL (approved):

PR #58216, opened 2020-08-17
Force quiet on all RunnerClients made by the salt-master

### LINKED ISSUE #58203: [BUG] salt-master wastes time writing event data to standard out when executing runner functions through the api
**Description**
I was doing some profiling to try and understand why some of my salt api requests were so slow and I noticed that the salt-master spends some considerable time in display_output printing data to standard out, which for me is /dev/null. 

I've worked around this by setting quiet: True in my salt master configuration. But this doesn't seem to be documented anywhere. 

In addition I notice that when the RunnerClient is constructed in salt.master.Maintenance._post_fork_init quiet is forced to True, but when the runner client used by the api is constructed in salt.master.ClearFuncs.runner it isn't. I wonder if it should be.

**Versions Report**
<details><summary>salt --versions-report</summary>
(Provided by running salt --versions-report. Please also mention any differences in master/minion versions.) 

```
Salt Version:
           Salt: 3001

Dependency Versions:
           cffi: 1.11.5
       cherrypy: unknown
       dateutil: Not Installed
      docker-py: Not Installed
          gitdb: Not Installed
      gitpython: Not Installed
         Jinja2: 2.10.1
        libgit2: Not Installed
       M2Crypto: Not Installed
           Mako: 1.0.9
   msgpack-pure: Not Installed
 msgpack-python: 0.5.6
   mysql-python: Not Installed
      pycparser: 2.14
       pycrypto: Not Installed
   pycryptodome: 3.9.8
         pygit2: Not Installed
         Python: 3.6.8 (default, Mar 21 2019, 10:08:12)
   python-gnupg: Not Installed
         PyYAML: 5.1
          PyZMQ: 17.0.0
          smmap: Not Installed
        timelib: Not Installed
        Tornado: 4.5.3
            ZMQ: 4.1.6

System Versions:
           dist: fedora 28 Twenty Eight
         locale: utf-8
        machine: x86_64
        release: 4.7.10
         system: Linux
        version: Fedora 28 Twenty Eight
```
</details>

mbirtwell:
### What does this PR do?
Ensure that quiet is set when creating RunnerClient objects inside the salt master so that it doesn't spend time writing the results of the runner command to stdout (which probably is /dev/null). This can be significant when the command is jobs.list_job on a large highstate.

Fixes: #58203

- [ ] Docs
- [ ] Tests written/updated

### Commits signed with GPG?
Yes/No


salt/master.py
```diff
@@ -103,6 +103,12 @@
 log = logging.getLogger(__name__)
 
 
+def _make_master_daemon_runner_client(opts):
+    ropts = dict(opts)
+    ropts["quiet"] = True
```
waynew, line 108: I feel like this should be `ropts.setdefault("quiet", True)`, otherwise it's going to override the option. Unless that's the intention 🤔

waynew (changes requested):
Hey @mbirtwell thanks for the PR - we'll definitely need some tests for this, though I expect this should also get the rest of the @saltstack/team-core to take a look.

salt/master.py
mbirtwell: Certainly that was the behaviour of the code that I was inspired by (line 198), but I don't know if that was the intention of that code. I thought it best to be consistent in the new place that I was applying this to. But if we're happy to change the behaviour for that bit. Then what you suggest works for me. Or maybe there's a reason that's not visible to me that means that we actually want different behaviour in these two places.

mbirtwell:
I'm not sure how to go about testing this. In tests/unit/test_master.py there are several tests for `salt.master.ClearFuncs.runner` but none of them actually execute the runner function. They all test `salt.master.ClearFuncs.runner` returning early with some kind of error (e.g. authorisation) before it gets to execute the runner requested. `salt.master.ClearFuncs.runner` would then generate a sub process to actually execute the runner function. It's this subprocess that I've changed the behaviour of. 

Maybe an integration test is more appropriate. I can't find anything that seems to invoke the salt master's runner interface directly. There is `tests.integration.netapi.test_client.NetapiClientTest.test_runner` which does so through the net api, which incidentally is relevant to my use case because I've come at this from the salt api. But it's very indirect at this point. And that test is already marked as skipped because there's something else broken that prevents it from actually testing anything (according to the comment).

Whilst looking for testing examples I came across `salt.daemons.masterapi.LocalFuncs.runner`, which looks very similar to the function that I was modifying (`salt.master.ClearFuncs.runner`). The only usage I can find (other than tests) of  `salt.daemons.masterapi.LocalFuncs` is to construct the `masterapi` attribute of `salt.master.ClearFuncs`, which in turn seems to be unused.

salt/master.py
waynew: Well, we're on the same page there, then 😂 I'm inclined towards consistency, I'd just like to know the original /why/, to know whether or not it's a foolish consistency 😝

mbirtwell: I'm going to take a punt at some logic to support the way that this was and therefore keeping this the same and making the other place where we create a RunnerClient consistent with the first:

The quiet flag seems only to be used to control the main human facing output for the RunnerClient this is the result of the salt runner that you are executing. It obviously makes sense to display this when the RunnerClient is being used from inside salt-run. However when the RunnerClient is being used from with in the salt-master it never make sense to display this because the salt-master is acting as a server and what ever client initiated the request is responsible for displaying this to the user. Therefore the RunnerClient should always be used with quiet=False in the salt-master.

sagetherage:
@mbirtwell looks like this has a merge conflict to address and it still needs a test case, please. : )

mbirtwell:
@sagetherage I've updated to fix the merge conflict. I previously wrote a bunch of stuff about testing this and would like a response to that before I try and tackle it. In summary I can't see a good example of testing this kind of change in the current test suite. Especially given that most of the change in behaviour occurs in a different process to one executing the code that I've changed. Please could you point me in the right direction.

sagetherage:
@mbirtwell A great way to tackle tests that are particularly complicated or like this one, where it affects another part of the code base, is to attend one of the test clinics during the week. You can subscribe or see them on the public [Salt Community Events Calendar](https://calendar.google.com/calendar/u/0?cid=c2FsdHN0YWNrLmNvbV9tZDczYzN1ZmNzMmVxYnNtbW5pa2U0ZW04MEBncm91cC5jYWxlbmRhci5nb29nbGUuY29t) and if that does not work with your schedule @waynew or @garethgreenaway can be reached on IRC or Slack for direct messages about this PR, as well. I hope that helps.

Also, this PR is failing on pre-commit, we require pre-commit to be run prior, but many PRs exists prior to the requirement -- so to fix that after the PR is created youcan try to pull down the PR locally, make a change (such as adding at test), stage the changed files, and then run the pre-commit manually:
```
    pre-commit run isort - run all the hooks
    pre-commit run black - run black
```
the force a push. This has worked for others, but not always, so we are working on globally addressing this, but in the mean time if you get help with tests also ask about pre-commit. Let me know if I can help more, happy to.

waynew (approved):
Yeah, I'm not sure how you could really test the particular places that you're changing. You *could*... do something. Lol. I mean, you can test your single function that you're creating.

One other approach that might work, but might be a pain to setup - you could add an integration test to the pytest suite, and use `capsys`, the pytest fixture to capture stdout/err. Then you just ensure that the thing is configured to use stdout/err and then just ensure that it doesn't dump whatever messages to out/err 🤷 

But... that's somewhat complicated and I'm leaning to recommending a 3-approval override. But if you *do* implement that test then 🚀  and we don't need to worry about that! 😁

mbirtwell:
I stand by this PR, my former employer had been using this in production for more than a year when I left, I believe that the check failures were flakes and not to do with the changes in this PR and I don't really see this a sensibly testable. But I want to close this to get it off my dashboard.

PR #58217, opened 2020-08-17
Allow cherrypy to gzip json and yaml responses

mbirtwell:
### What does this PR do?

Sets some cherrypy configuration so that it will gzip json and yaml responses from the salt API. Only applies if the client sends Accept-Encoding: gzip

- [ ] Docs
- [x] Tests written/updated

### Commits signed with GPG?
No


krionbsd (changes requested):
@mbirtwell thanks for contribution, could you please run pre-commit hook on your changes and re-commit it? It would be nice if you could add a test case for this change as well, if possible.

mbirtwell:
I think that the unit tests were interacting with cherrypy at a level which didn't run the middleware responsible for gzipping responses. So I've gone for an integration test.

krionbsd:
> I think that the unit tests were interacting with cherrypy at a level which didn't run the middleware responsible for gzipping responses. So I've gone for an integration test.

Awesome @mbirtwell, thanks a lot, we'll merge it soon

krionbsd (approved):

mbirtwell:
I stand by this PR, my former employer had been using this in production for more than a year when I left and I believe that the check failures were flakes and not to do with the changes in this PR. But I want to close this to get it off my dashboard.

PR #58238, opened 2020-08-18
Cleanup calculation of template sls/tpl context

### LINKED ISSUE #56410: tpldir calculation is broken in some cases
### Description of Issue

Buggy code is here:
https://github.com/saltstack/salt/blob/84c60708cd355f0b24742e8bcdba39073d24c9f8/salt/utils/templates.py#L127-L131

Short version, if name of the directory containing the sls is contained in sls name, tpldir calculation will return wrong result.

Slightly longer version, it seems to be trying to find the tplpath by looking for last occurrence of slspath in template name - without realizing it may occur later in the full template path.

This can be fixed by simply adding the basename of the template to the slspath - although I am not really sure what is the point of this code - seems like a complicated way to get what you already have in slspath....

### Setup

setup as file x/x.sls

```
test:
  test.nop:
    - name: {{ tpldir }}
```
### Steps to Reproduce Issue

* if you run the state above, you will get output of name to be `.` - which is incorrect
* if you rename that state file to anything without "x" in the name, it will return name 'x' (correct)
* if you name it anything with x in the name, you will have a problem
* if you move it to x/x/x.sls - you will get "x" - which is, again, incorrect
* if you move it to x/x/y.sls - you will get "x/x" - which is correct

### Versions Report

bug is still present in the master branch as evident by code link above

### LINKED ISSUE #58249: [BUG] `tplfile` variable not set in some sls files
**Description**
When processing sls in root directory of file roots, `tplfile` context variable  is not set

**Setup**
standard file roots

create a `tplfile.sls` on the top level with following contents:
```
test:
  test.nop:
    - name: tplfile is "{{ tplfile }}"
```

run `salt-call state.show_sls tplfile --output=yaml`

result:
```
# [INFO    ] Loading fresh modules for state activity
local:
  test:
    test:
    - name: tplfile is ""
    - nop
    - order: 10000
    __sls__: tplfile
    __env__: base
```
As you see, tplfile, which should be `tplfile.sls` is not set

**Expected behavior**
expect tplfile to be set to `tplfile.sls` 

**Screenshots**
N/A

### LINKED ISSUE #58250: [BUG] When calling init.sls via explicit init name, the name is treated as a directory name in context
**Description**

if "init" SLS is access using it's explicit name - i.e. `path.to.init` instead of `path.to` - the "init" part shows up as a directory component in various `sls*` context parmeters

**Setup**
Set up following file `slspath/init.sls`:
```
test:
  test.nop:
    - context:
        slspath: "{{ slspath }}"
        slsdotpath: "{{ slsdotpath }}"
        slscolonpath: "{{ slscolonpath }}"
        sls_path: "{{sls_path}}"
```

Run: `salt-call state.show_sls slspath.init --output=yaml`

Result:
```
# [INFO    ] Loading fresh modules for state activity
local:
  test:
    test:
    - context:
        slspath: slspath/init
        slsdotpath: slspath.init
        slscolonpath: slspath:init
        sls_path: slspath_init
    - nop
    - order: 10000
    __sls__: slspath.init
    __env__: base
```

As you see, "init" becomes a directory name, even though no such directory exists

**Expected behavior**

Expecting `slspath` to alway reflect actual path of the sls - i.e. in this case it should be `slspath`

mlasevich:
### What does this PR do?
Cleanup calculation of sls* and tpl* context parameters when parsing sls files. Fix bug which under some circumstances produced wrong value for tpldir

* Fixes: #56410 - `tpldir` and related are calculated incorrectly when name repeats in the path
* Fixes #58250 - inclusion of "init" in `sls*` values when sls path explicitly contains it. i.e. if sls is foo.bar.init - expected path is `foo/bar`and not `foo/bar/init` as previously
* Fixes #58249 - Blank `tplfile` value  when loading top level sls files 

### Previous Behavior
See issue #56410. If directory name matched part of the path name, wrong value was calculated

### New Behavior
There should be no functional difference, all the same values are being produced, the key difference is how the template path and related variables are calculated.

- [x] Docs
- [x] Tests written/updated

### Commits signed with GPG?
No


salt/utils/templates.py
```diff
@@ -99,6 +97,54 @@ def __getattr__(self, name):
 
 
 def wrap_tmpl_func(render_str):
+    def _sls_context(tmplpath, sls):
```
waynew, line 100: No need to make this a local function. That just makes it harder to test, and additionally has to re-create the function every time `wrap_tmpl_func` is called.

salt/utils/templates.py
```diff
@@ -99,6 +97,54 @@ def __getattr__(self, name):
...
+            template = tmplpath.replace("\\", "/")
+
+            # Determine proper template name without root
+            if template.endswith("{}.sls".format(slspath)):
+                template = template[-(4 + len(slspath)) :]
+            elif template.endswith("{}/init.sls".format(slspath)):
+                template = template[-(9 + len(slspath)) :]
+            else:
+                # Something went wrong
+                log.warning("Failed to determine proper template path")
```
waynew, line 121: Generally speaking, I bet `pathlib.Path` would be useful here. It already contains methods for renaming, getting the parent, etc. - it would make the logic throughout this function much more apparent.

waynew (changes requested):
Thanks for the PR! I know you're already aware of the need for tests 🙃 

I've added a couple of points of feedback that should help. If you need help getting some tests going on this, let me know!

salt/utils/templates.py
mlasevich: Ok, any issue with making it a module function. I stuck it here for now because I hated the large block of self contained code embedded, but was not sure where is the right place for it.

salt/utils/templates.py
mlasevich: Is there any scenario under which the sls would not match the template path name? This else is a bit of my paranoia, but I am not sure if it is ever going to be reached. I will check out `pathlib.Path` though. If there is a better way than comparing sls to template path, I'd rather do that and not have all these conditions

mlasevich:
Been trying to satisfy the quality gates, but this is starting to spin out of control...

salt/utils/templates.py
mlasevich: Moved it to module level

salt/utils/templates.py
mlasevich: I looked into pathlib - but I am not sure how it is useful here. The issue is that if sls does not map to template path, there is no way to figure out which part of the template path is the "file root" and which is the template within said root. I am not sure any standard library would be able to do this as this is something only meaningful within the code. I am not sure if there is ever a scenario where it would fall through to the `else` section, so it is mostly there for defensive purposes in case either I do not understand things correctly or something changes in the future.

sagetherage:
@mlasevich looks like a merge conflict needs to be addressed, here, can you take a look, please?

mlasevich:
Ugh. This is why it is nice to merge PRs quickly rather than letting them sit for a months  ;-) I will take a look :-)

mlasevich:
I am a bit concerned that I could not find the actual conflict. I do not see any code changes in the block indicated from master. I am assuming this came from the pre-commit hook reformatting the code, but can someone else take a quick glance at this to make sure I did not miss something. Thanks

mlasevich:
Also, @waynew - you have it marked as "changes requested" - are there any changes being requested?

waynew (dismissed):

dwoz:
We need to get lots of eyes on this change. See [@terminalmage's comment](https://github.com/saltstack/salt/issues/56410#issuecomment-675244854) about potentially breaking people's existing states.

mlasevich:
> We need to get lots of eyes on this change. See [@terminalmage's comment](https://github.com/saltstack/salt/issues/56410#issuecomment-675244854) about potentially breaking people's existing states.

To be fair, this is a bugfix and not at all a functionality change like the previous attempt. This Implementation does exactly what previous implementation was doing, but fixes some situations where the value of `tpldir` may be unexpected. While it is possible that it will break some existing states, I cannot imagine a scenario where someone would expect `tpdir` to point to an unrelated directory on purpose, so I would argue that those states are already broken - they just don't know that yet. This would be true of any bugfix...

While I am not against more review of this, this PR has been "done" for over a month now, and my desire to merge it sooner rather than later it is primarily out of concern that with all the slow pre-commit changes, we are going to end up with a lot more merge conflicts over time.

dwoz:
It's been a while since I looked a this specific issue but what I recall is there was a consensus that we should not change tpldir. I thought we agreed to add a new variable that has all the fixes and send tpldir down a deprecation path.

We can not merge this if we know it will break people's existing states even if people are relying on 'broken' functionality.

mlasevich:
@dwoz I am not really following - are you saying no more bugfixes for any bugs in salt ever in case someone has a typo? That seems crazy. 

We are not talking about changing one functionally for another, we are talking about a function that under some conditions will returns something completely wrong, unrelated value. While it is theoretically possible that someone has this bug in their state AND has another bug that hides this bug, nobody is actually RELYING on this functionality. It is a straight up edge case bug. if you have a bad combination of names in your state, it will garble the output. Yes there have been attempts to change functionality, but this is not it. This fixes specifically the situation  that if you have a state called `x/x/y.sls` it will return  "x/x and if you have `x/x/x.sls` it will return "x" - who would expect/count on that?

garethgreenaway (dismissed):

dwoz:
Because we definitely want these fixes but we need to avoid breaking existing state trees I have added a feature flag for this functionality. To use these fixes add the following to your minion's config:

```
features:
  use_slsvars_fixes: True
```

This feature flag will go away in the Phosphorus (3005) and the fixed functionality will become the default.

twangboy (approved):

Akm0d (dismissed):

dwoz:
re-run windows

dwoz:
re-run full all

twangboy (dismissed):

dwoz:
re-run full all

max-arnold:
I spent some time actually trying to use this feature, and it resulted in nothing but frustration:

1. Either the docs or the code is wrong. In the docs the feature toggle is named `use_slsvars_fixes`, in the code and changelog it is named `enable_slsvars_fixes`.  If you follow the docs, it won't work.
2. There supposed to be a warning (`There have been significant improvement to template variables...`), but it isn't shown in the `salt-call -l warning` console output. Maybe it is because I installed Salt from git, maybe for some other reason...
3. When I finally enabled it, I got this:

```
% sudo salt-call state.apply example_state -l warning

[CRITICAL] Rendering SLS formula1 failed, render error: 'NoneType' object is not iterable
local:
    Data failed to compile:
----------
    Rendering SLS formula1 failed, render error: 'NoneType' object is not iterable
  File "/usr/local/lib/python3.8/dist-packages/salt/state.py", line 3959, in render_state
    state = compile_template(
  File "/usr/local/lib/python3.8/dist-packages/salt/template.py", line 101, in compile_template
    ret = render(input_data, saltenv, sls, **render_kwargs)
  File "/usr/local/lib/python3.8/dist-packages/salt/renderers/jinja.py", line 66, in render
    tmp_data = salt.utils.templates.JINJA(
  File "/usr/local/lib/python3.8/dist-packages/salt/utils/templates.py", line 234, in render_tmpl
    context.update(sls_context)
TypeError: 'NoneType' object is not iterable
```

The example state looks like this (it doesn't matter if it is placed in the root or in a subfolder):

```yaml
{% do salt.log.warning('VAR tpldir="{}"'.format(tpldir)) %}
{% do salt.log.warning('VAR slspath="{}"'.format(slspath)) %}
{% do salt.log.warning('VAR tplfile="{}"'.format(tplfile)) %}
{% do salt.log.warning('VAR tpldot="{}"'.format(tpldot)) %}
```

4. Why we need another feature toggle subsystem? 

We already have the (badly named) `use_superseded` config namespace for that. Maybe it is better to give it a more intuitive name alias and avoid introducing another one (`features`)?

5. Have [these complications](https://github.com/saltstack/salt/issues/41195#issuecomment-577159561) been considered/tested with this PR?

mlasevich:
Can't speak to the feature flags, that was not part of my original PR and was added by @dwoz as a compromise to get things merged "safely" to not affect existing functionality. Unfortunately the issue you see in #3 is a direct result of a bug that was apparently introduced with the feature flags - a simple, but devastating bug. When the old  implementation was re-introduced in https://github.com/saltstack/salt/pull/58238/commits/9380b56692ca04a043731fcaeb0ca049a7251e9b, a bug was added that completely broke the new implementation.  It is a simple enough fix(missing return statement) -  I will open a new PR to fix it. :-(

As for your last point, as the names of those variables suggest, they are designed with SLS files in mind - so mapping them to non-sls files may not produce what you would expect. That said, remember that you are loading those files from an sls file, and as such, passing in the sls context into it - so you can probably pass in whatever context items you want. I will glance at it when I have a sec. FWIW, we moved away from using map.jinja files in general, as there are much better ( at least for us), ways to handle those needs.

max-arnold:
Thanks for the quick response! Hopefully the fix will be merged before the Magnesium is out of the doors 🤞

I also Cc'ed @myii, maybe he will provide some additional feedback on these changes after running the `saltstack-formulas` test suite

changelog/58652.added
```diff
@@ -0,0 +1,4 @@
+Added features config option for feature flags. Added a feature flag
+`enable_slsvars_fixes` to enable fixes to tpldir, tplfile and sls_path.
+This flag will be depricated in the Phosphorus release when this functionality
```
myii, line 3: Typo:

```suggestion
This flag will be deprecated in the Phosphorus release when this functionality
```

doc/ref/states/vars.rst
```diff
@@ -2,6 +2,21 @@
 SLS Template Variable Reference
 ===============================
 
+
+.. warning::
+   In the 3002 release sls_path, tplfile, and tpldir have had some significate
```
myii, line 7: Typo mainly, formatting for consistency:

```suggestion
   In the 3002 release ``sls_path``, ``tplfile``, and ``tpldir`` have had some significant
```

doc/ref/states/vars.rst
```diff
@@ -94,14 +114,96 @@ include option.
...
+.. code-block:: jinja
+
+    {{ slscolonpath }}
+
+tplpath
+-------
+
+Full path to sls template file being process on local disk. This is usually
+pointing to a copy of the sls file in a cache directory. This will be in OS
+specific format (windows vs posix). (It is probably best not to use this.)
```
myii, line 168:
```suggestion
specific format (Windows vs POSIX). (It is probably best not to use this.)
```

tests/unit/utils/test_templates.py
```diff
@@ -218,26 +211,197 @@ def __call__(self, tplstr, context, tmplpath=None):
...
+            tpldir=".",
+            tpldot="",
+            slsdotpath="",
+            slscolonpath="",
+            sls_path="",
+            slspath="",
+        )
+
+    def test_generate_sls_context__one_level_init_implicit(self):
+        """ generate_sls_context - Basic one level with impliocit init.sls """
```
myii, line 276: Typo:

```suggestion
        """ generate_sls_context - Basic one level with implicit init.sls """
```

myii (commented):
@mlasevich When providing the fix you've been discussing with @max-arnold, please consider these items as well.

PR #58245, opened 2020-08-19
Add extra xml states.

Timbus:
### What does this PR do?
Adds two new xml states, `contains_tree` and `contains_fragment`. They're documented (need to verify the documentation formatting though)

It didn't seem possible to make this into execution modules? Because you have to keep opening the file.. Wonder how ansible does it..

Anyway this needs tests I guess, but wanted to kick this off before I went to sleep and I have no idea how the existing tests work. Will figure it out soon enough. Thanks

### New Behavior
Allows one to statefully manage fragments of XML inside XML files.

- [x] Docs
- [ ] Tests written/updated

### Commits signed with GPG?
No

### DIFF
--- a/salt/states/xml.py
+++ b/salt/states/xml.py
@@ -85,7 +85,7 @@ def value_present(name, xpath, value, **kwargs):
     return ret
 
 
-def merge_fragment(name, fragment, xpath=None):
+def contains_tree(name, fragment, xpath=None):
     """
     .. versionadded:: NEXT_RELEASE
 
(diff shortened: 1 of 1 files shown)


salt/states/xml.py
```diff
@@ -54,20 +67,216 @@ def value_present(name, xpath, value, **kwargs):
...
+        the fragment's tags match, they will be merged -- the fragment's
+        properties will overwrite any existing properties with the same key.
+        This means attributes are merged, text, and tail values are replaced.
+
+    .. code-block:: yaml
+
+        ensure_tree_exists:
+        xml.contains_tree:
+            - name: /etc/data.xml # MUST already have valid XML.
+            - resolve: merge
```
mchugh19, line 119: Looks good!

The resolve key appears to no longer be used, so it can be removed from the docs.

Timbus: Fixed! Thanks for noticing.

dwoz:

PR #58262, opened 2020-08-21
Add pkg.services_need_restart

### LINKED ISSUE #58261: [FEATURE REQUEST] Ability to list system services that were affected by package manager operations
Not all services that run on a system are managed by Salt and can be restarted automatically. One example is the salt-minion service itself, but there are others that are unlikely to be managed by Salt. Reinstalling, upgrading, or downgrading packages might affect running services which can cause further problems. 

One example is an update to Python, which touches `/usr/lib64/python2.7/lib-dynload/mmap.so` and causes a stacktrace like this, e.g. on highstate:
```
  File "/usr/lib/python2.7/site-packages/salt/loader.py", line 1606, in _load_module
    mod = imp.load_module(mod_namespace, fn_, fpath, desc)
  File "/usr/lib/python2.7/site-packages/salt/modules/file.py", line 31, in <module>
    import mmap
ImportError: /usr/lib64/python2.7/lib-dynload/mmap.so: undefined symbol: _PySlice_Unpack
```
Restarting `salt-minion` fixed this issue.

I would like to have a execution function that lists services that might be need a restart. In case highstate is not working, this allows the user to check if a restart to salt-minion is needed. This list can also be used for other services, not only for salt-minion.

**Describe alternatives you've considered**
Automatically restarting salt-minion on updating Python is a possibility, but not in the middle of a highstate.

agraul:
### What does this PR do?
Add `pkg.services_need_restarting` to Zypper, Yum(DNF) and Apt.

Current name proposal is `pkg.services_need_restarting`, but I don't mind changing the name to a better one if suggested.

Fixes: #58261 

### New Behavior
```
server:~ # salt 192.168.100.194 pkg.services_need_restarting
192.168.100.194:
    - salt-minion
    - qemu-guest-agent
    - sssd
    - polkit
    - gssproxy
    - getty@tty1
    - firewalld
    - NetworkManager
    - ModemManager
```

- [ ] Docs
- [x] Tests written/updated

### Commits signed with GPG?
Yes (always ;))

### DIFF
--- a/salt/modules/aptpkg.py
+++ b/salt/modules/aptpkg.py
@@ -37,7 +37,12 @@ import salt.utils.stringutils
 import salt.utils.systemd
 import salt.utils.versions
 import salt.utils.yaml
-from salt.exceptions import CommandExecutionError, MinionError, SaltInvocationError
+from salt.exceptions import (
+    CommandExecutionError,
+    CommandNotFoundError,
+    MinionError,
+    SaltInvocationError,
+)
 from salt.modules.cmdmod import _parse_env
 
 log = logging.getLogger(__name__)
@@ -3009,3 +3014,38 @@ def list_downloaded(root=None, **kwargs):
                 ).isoformat(),
             }
     return ret
+
+
+def services_need_restart(**kwargs):
+    """
+    .. versionadded:: 3003
+
+    List services that use files which have been changed by the
+    package manager. It might be needed to restart them.
+
+    Requires checkrestart from the debian-goodies package.
+
+    CLI Examples:
+
+    .. code-block:: bash
+
+        salt '*' pkg.services_need_restart
+    """
+    if not salt.utils.path.which_bin(["checkrestart"]):
+        raise CommandNotFoundError(
+            "'checkrestart' is needed. It is part of the 'debian-goodies' "
+            "package which can be installed from official repositories."
+        )
+
+    cmd = ["checkrestart", "--machine", "--package"]
+    services = set()
+
+    cr_output = __salt__["cmd.run_stdout"](cmd, python_shell=False)
+    for line in cr_output.split("\n"):
+        if not line.startswith("SERVICE:"):
+            continue
+        end_of_name = line.find(",")
+        service = line[8:end_of_name]  # skip "SERVICE:"
+        services.add(service)
+
+    return list(services)
--- a/salt/modules/yumpkg.py
+++ b/salt/modules/yumpkg.py
@@ -3399,13 +3399,10 @@ def services_need_restart(**kwargs):
 
     services = set()
     for line in dnf_output.split("\n"):
-        try:
-            pid, _ = line.split(":")
-        # Skip lines that can't be parsed
-        except ValueError:
-            continue
-        service = salt.utils.systemd.pid_to_service(pid.strip())
-        if service:
-            services.add(service)
+        pid, has_delim, _ = line.partition(":")
+        if has_delim:
+            service = salt.utils.systemd.pid_to_service(pid.strip())
+            if service:
+                services.add(service)
 
     return list(services)
--- a/salt/modules/yumpkg.py
+++ b/salt/modules/yumpkg.py
@@ -3369,3 +3369,43 @@ def list_installed_patches(**kwargs):
         salt '*' pkg.list_installed_patches
     """
     return _get_patches(installed_only=True)
+
+
+def services_need_restart(**kwargs):
+    """
+    .. versionadded:: 3003
+
+    List services that use files which have been changed by the
+    package manager. It might be needed to restart them.
+
+    Requires systemd.
+
+
+    CLI Examples:
+
+    .. code-block:: bash
+
+        salt '*' pkg.services_need_restart
+    """
+    if _yum() != "dnf":
+        raise CommandExecutionError("dnf is required to list outdated services.")
+    if not salt.utils.systemd.booted(__context__):
+        raise CommandExecutionError("systemd is required to list outdated services.")
+
+    cmd = ["dnf", "--quiet", "needs-restarting"]
+    dnf_output = __salt__["cmd.run_stdout"](cmd, python_shell=False)
+    if not dnf_output:
+        return []
+
+    services = set()
+    for line in dnf_output.split("\n"):
+        try:
+            pid, _ = line.split(":")
+        # Skip lines that can't be parsed
+        except ValueError:
+            continue
+        service = salt.utils.systemd.pid_to_service(pid.strip())
+        if service:
+            services.add(service)
+
+    return list(services)
--- a/salt/modules/zypperpkg.py
+++ b/salt/modules/zypperpkg.py
@@ -3014,3 +3014,27 @@ def resolve_capabilities(pkgs, refresh=False, root=None, **kwargs):
         else:
             ret.append(name)
     return ret
+
+
+def services_need_restart(root=None, **kwargs):
+    """
+    .. versionadded:: 3003
+
+    List services that use files which have been changed by the
+    package manager. It might be needed to restart them.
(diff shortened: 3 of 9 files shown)


agraul:
I'll fix the CI complains due to the Python2 removal and push the changes asap, hopefully tomorrow.

salt/modules/zypperpkg.py
```diff
@@ -539,14 +534,11 @@ def list_upgrades(refresh=True, root=None, **kwargs):
...
-        if isinstance(repos, six.string_types):
+        if isinstance(repos, str):
             repos = [repos]
         for repo in repos:
             cmd.extend(
-                [
-                    "--repo",
-                    repo if isinstance(repo, six.string_types) else six.text_type(repo),
-                ]
+                [("--repo", repo) if isinstance(repo, str) else (str(repo), )]
```
agraul, line 541: This needs to be `(str(repo), )` because I get a conflict between `pylint` and `black` otherwise. For consistency I've also added the parentheses to the first tuple.

`pylint` reports 
```
salt/modules/zypperpkg.py:541: [E8231(missing-whitespace-after-comma), ] PEP8 E231: missing whitespace after ','
```
but I can't add the whitespace -- `black` will remove it.

aplanas: ah no, remove the parenthesis:

```python
[("--repo", repo) if isinstance(repo, str) else str(repo)]
```

agraul: Thanks, I've adjusted that now!

salt/modules/yumpkg.py
```diff
@@ -3335,3 +3335,43 @@ def list_installed_patches(**kwargs):
...
+    services = set()
+    for line in dnf_output.split("\n"):
+        try:
+            pid, _ = line.split(":")
+        # Skip lines that can't be parsed
+        except ValueError:
+            continue
+        service = salt.utils.systemd.pid_to_service(pid.strip())
+        if service:
+            services.add(service)
```
waynew, line 3375:
```suggestion
        pid, has_delim, _ = line.partition(':')
        if has_delim:
            service = salt.utils.systemd.pid_to_service(pid.strip())
            if service:
                services.add(service)
```

**suggestion:** this may be a bit cleaner way to write this - exception handling can be avoided, since partition will *always* return 3 values - and if there is no `:` then it won't even show up. The output may never have extra `:`, but this will also always limit to the first `:` per line. This also avoids having the `continue` in there as well.

waynew (dismissed):
👍  @agraul nice! Thanks for the PR - I think this is excellent. I'm not sure if "services_need_restart" would read better? 🤷 

I also left a suggestion - I don't think it's necessary, but it might be nicer?

Anyway, looks great!

salt/modules/yumpkg.py
agraul: Agreed, that is a nicer way!

agraul:
I'm fine with either `services_need_restart` or `services_need_restarting`. Given you're the native speaker between the two of us, I'll use your suggestions if no one else joins the discussion with a reason not to use it in the upcoming days :)

I'll rebase and fix the `black` failure once I change the name.

waynew:
> I'm fine with either `services_need_restart` or `services_need_restarting`. Given you're the native speaker between the two of us, I'll use your suggestions if no one else joins the discussion with a reason not to use it in the upcoming days :)

TBH both of them sound a bit weird. The only reason I suggest the alternative is that we have [`needs_reboot`](https://docs.saltstack.com/en/develop/ref/modules/all/salt.modules.kernelpkg_linux_yum.html#salt.modules.kernelpkg_linux_yum.needs_reboot) already. That's the closest analog I was able to find within the existing codebase. 🤷 

But it's a bit different so I wasn't 100% convinced that it made sense to try and match that naming.

agraul:
> TBH both of them sound a bit weird.

I agree. Do you have a better idea? My initial idea was `list_outdated_services` but then a colleague of mine said it was misleading because someone might think of it being related to a `systemctl daemon-reload` situation. 

In SUSE systems there is a `zypper needs-rebooting` command to check for needed reboots, on RHEL systems we got `dnf needs-restarting` so from that perspective the `-ing` is consistent. But now that you mention `needs_reboot` in Salt, I think we should use the same verb form.

waynew:
re-run windows2019-py3

waynew:
:thinking: Hm. That win2019 fail seems to be reliably happening, but I think that's an infrastructure issue. I've pinged the rest of @saltstack/team-core to see if they know anything.

waynew:
Let's see if updating the branch will get tests passing

waynew (dismissed):

agraul:
I hope this latest 'merge branch master ...' did not spoil it, the test were green before :sweat_smile:

sagetherage:
@agraul looks like a merge conflict, but then it should be good to go!

salt/modules/aptpkg.py
```diff
@@ -2995,3 +3000,38 @@ def list_downloaded(root=None, **kwargs):
...
+
+        salt '*' pkg.services_need_restart
+    """
+    if not salt.utils.path.which_bin(["checkrestart"]):
+        raise CommandNotFoundError(
+            "'checkrestart' is needed. It is part of the 'debian-goodies' "
+            "package which can be installed from official repositories."
+        )
+
+    cmd = ["checkrestart", "--machine"]
```
OrangeDog, line 3026: I think a `--package` is necessary to avoid certain false-positives. Or a way to take additional arguments.

OrangeDog (changes requested):

salt/modules/aptpkg.py
agraul: I think you're right with `--package` ([man page](https://manpages.debian.org/stretch/debian-goodies/checkrestart.1.en.html)).

Do you think it is needed to have other arguments?

OrangeDog: > Do you think it is needed to have other arguments?

I don't think so, but it's nice to be flexible if people have other ideas or new ones are added.

agraul: I only worry that adding arbitrary arguments will move this from an implementation detail for gathering a list of services to something that will be used in a way that breaks the parsing done blow.
Parsing the output is already pretty brittle and not something I like doing, but it seems like the easiest way for achieving the goal for Debian systems (which I have to admit I have very little experience with).

OrangeDog: That's a fair point. Just adding `--package` would be sufficient.

salt/modules/yumpkg.py
```diff
@@ -3371,3 +3371,40 @@ def list_installed_patches(**kwargs):
         salt '*' pkg.list_installed_patches
     """
     return _get_patches(installed_only=True)
+
+
+def services_need_restart(**kwargs):
+    """
+    .. versionadded:: NEXT
```
Ch3LL, line 3378: 3003

salt/modules/zypperpkg.py
```diff
@@ -2890,3 +2890,27 @@ def resolve_capabilities(pkgs, refresh=False, root=None, **kwargs):
         else:
             ret.append(name)
     return ret
+
+
+def services_need_restart(root=None, **kwargs):
+    """
+    .. versionadded:: NEXT
```
Ch3LL, line 2897: 3003

Ch3LL (changes requested):
One small change. I would love to get this into 3003, but we will be tagging any day now so I cannot guarantee this.

agraul:
@Ch3LL good catch! I've changed the NEXT to 3003 in all three pkg modules

Ch3LL (approved):

s0undt3ch (approved):

twangboy (approved):

PR #58268, opened 2020-08-21
Skip test which is incompatible with FreeBSD

krionbsd:
### What does this PR do?
Skip test which is incompatible with FreeBSD


DmitryKuzmenko (dismissed):

krionbsd:
re-run ubuntu2004

dwoz (approved):

PR #58284, opened 2020-08-26
[PyTest Migration] Minor fixes & Improvements to PyTest runs

s0undt3ch:
### What does this PR do?
See title

### DIFF
--- a/tests/pytests/unit/utils/test_msgpack.py
+++ b/tests/pytests/unit/utils/test_msgpack.py
@@ -1,5 +1,3 @@
-import os
-
 import pytest
 import salt.utils.msgpack
 from tests.support.mock import MagicMock, patch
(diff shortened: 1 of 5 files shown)


krionbsd (dismissed):

tests/pytests/unit/utils/test_msgpack.py
```diff
@@ -14,17 +12,16 @@ def test_load_encoding(tmpdir):
 
     https://github.com/msgpack/msgpack-python/blob/master/ChangeLog.rst
     """
-    fn_ = os.path.join(tmpdir, "test_load_encoding.txt")
-    salt.utils.msgpack.version = (1, 0, 0)
```
s0undt3ch, line 18: @Ch3LL :facepalm: missed this in my review of your PR....
We can't just change attributes while testing, we need to `patch` it.

Ch3LL: makes total sense. I actually got that from the current msgpack unit test so that will need to be updated. I'll push up a fix in a sec.

s0undt3ch: I know we have a bunch of this wrong way of doing things, we need to clean those up as we find them :/

Ch3LL: https://github.com/saltstack/salt/pull/58289  looks like we were actually setting the version and then setting it back to the original version, but i wrote it to use patch anyways since that is better.

PR #58287, opened 2020-08-26
do not del roster_file from opts for ssh netapi

Ch3LL:
### What does this PR do?
Fixes test failure: `integration.netapi.test_client.NetapiSSHClientTest.test_ssh`

This started failing after removing ip requirement since we are no longer converting ip to dns: https://github.com/saltstack/salt/pull/58163/files#diff-57bb6fe4c0e17531b2558ba86e91401fL356 
This was not failing previously on the tests simple because the target name was not an ip address. If it was then this test would have failed previously.

this resulted in `salt.roster.get_roster_file` getting called multiple times and `roster_file` would be removed from opts.

### DIFF
--- a/salt/roster/__init__.py
+++ b/salt/roster/__init__.py
@@ -1,11 +1,9 @@
-# -*- coding: utf-8 -*-
 """
 Generate roster data, this data is used by non-minion devices which need to be
 hit from the master rather than acting as an independent entity. This covers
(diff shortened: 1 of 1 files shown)


Ch3LL:
re-run full all

s0undt3ch:
The macOS test failure is fixed in https://github.com/saltstack/salt/pull/58288

s0undt3ch (approved):

PR #58319, opened 2020-08-28
update arch ami

bryceml:
### What does this PR do?
updates the arch ami to latest
It has problems that need fixed: https://jenkinsci.saltstack.com/job/Full-PR/job/full-pr-arch-lts/job/master/15/


bryceml:
re-run full archlts

s0undt3ch:
The failures suggest valid failures which need to be fixed in order to use this new AMI

DmitryKuzmenko:
Merged master to pull this change to my PR properly. Just want to check will it fix both tests or just one.

PR #58350, opened 2020-09-01
Revert "Bump black to version 20.8b1"

s0undt3ch:
# What does this PR do?
This reverts commit e22fa636f5cf66de000e1a527dc933a053c62b17.

We will have to update a bunch of files to upgrade to this version of black, which we should, but it should all be in one go.


Ch3LL (approved):

s0undt3ch:
Pre commit passed. Merging.

PR #58368, opened 2020-09-03
[PyTest Migration] PyTest Fixes & Improvements

s0undt3ch:
### What does this PR do?
See title


s0undt3ch:
re-run full all

s0undt3ch:
re-run full all

DmitryKuzmenko (dismissed):

waynew:
FWIW it would be easier to review shifting tests to pytest if you have a single PR per migration.

If the migration requires a bunch of updates to the test harness, I think it's totally reasonable to set the subsequent PRs as draft until the necessary PRs get merged.

s0undt3ch:
> If the migration requires a bunch of updates to the test harness, I think it's totally reasonable to set the subsequent PRs as draft until the necessary PRs get merged.

Unfortunately there's a catch 22. Some migrations will only be possible after upgrading to the most recent pytest-salt-factories.
And pytest-salt-factories can only be upgraded with the migrations or we'll either have test failures, or added flakiness.

If you get past the 50+ files modified which are just the requirements, there's way less to review.
I also made it that we can still keep running runtests, while updating the test suite for PyTest for improved stability, untill we actually switch to PyTest alone.
So, on this note, how hard would it be to fullfil a review and not stale stabilizing pytest before we migrate to it?
Come one @waynew, I know you can do it ;)

s0undt3ch:
re-run full all

s0undt3ch:
Bah! New AMIs have issues with pytest....
Anyway, I'm of to PTO, it can wait until I'm back.

PR #58380, opened 2020-09-04
Use default timezone offset in scheduler when correct timezone cannot be determined

### LINKED ISSUE #58379: [BUG] Minion fails to start on Windows if timezone cannot be determined
**Description**
When Windows timezone cannot be determined for any reason, if defaults to `Unknown`, which is invalid timezone for pytz and exception is raised. Due to scheduler relying on `timezone.get_offset` not raising exception, it then fails in initialization, hence whole minion fails to start.

**Setup**
Windows system where timezone is unknown (the exact reason is not clear).

Possibly empty values in following registry keys:
```
HKLM\SOFTWARE\Microsoft\Windows NT\CurrentVersion\Time Zones 
HKLM\System\CurrentControlSet\Control\TimeZoneInformation
```

Then start minion and encounter:
```
2020-09-02 11:10:37,702 [salt.minion      :1050][CRITICAL][972] Unexpected error while connecting to <master>
  File "c:\salt\bin\lib\site-packages\salt\minion.py", line 1027, in _connect_minion
    yield minion.connect_master(failed=failed)
  File "c:\salt\bin\lib\site-packages\tornado\gen.py", line 1055, in run
    value = future.result()
  File "c:\salt\bin\lib\site-packages\tornado\concurrent.py", line 238, in result
  File "<string>", line 4, in raise_exc_info
  File "c:\salt\bin\lib\site-packages\tornado\gen.py", line 1063, in run
    yielded = self.gen.throw(*exc_info)
  File "c:\salt\bin\lib\site-packages\salt\minion.py", line 1226, in connect_master
    yield self._post_master_init(master)
  File "c:\salt\bin\lib\site-packages\tornado\gen.py", line 1055, in run
    value = future.result()
  File "c:\salt\bin\lib\site-packages\tornado\concurrent.py", line 238, in result
  File "<string>", line 4, in raise_exc_info
  File "c:\salt\bin\lib\site-packages\tornado\gen.py", line 1069, in run
    yielded = self.gen.send(value)
  File "c:\salt\bin\lib\site-packages\salt\minion.py", line 1273, in _post_master_init
    cleanup=[master_event(type='alive')])
  File "c:\salt\bin\lib\site-packages\salt\utils\schedule.py", line 105, in __new__
    utils=utils)
  File "c:\salt\bin\lib\site-packages\salt\utils\schedule.py", line 152, in __singleton_init__
    self.time_offset = self.functions.get('timezone.get_offset', lambda: '0000')()
  File "c:\salt\bin\lib\site-packages\salt\modules\win_timezone.py", line 241, in get_offset
    tz_object = pytz.timezone(get_zone())
  File "c:\salt\bin\lib\site-packages\pytz\__init__.py", line 181, in timezone
    raise UnknownTimeZoneError(zone)
pytz.exceptions.UnknownTimeZoneError: 'Unknown'
```

**Expected behavior**
Minion starts and connects to master, even with default timezone.

**Versions Report**
At least 2019.2 through 3001 minions on Windows (found on Windows 2008 R2)

**Additional context**
Related to https://support.microsoft.com/en-us/help/2562433/the-timezonekeyname-registry-entry-value-changes-to-a-null-value-in-wi and https://github.com/saltstack/salt/issues/50667

lukasraska:
### What does this PR do?
Defaults timezone offset to `0000` if `timezone.get_offset` raises and exception when scheduler is instantiated.

There can be also destructive integration tests written, to test this scenario fully (which will mess with Win registries), but from grepping the Salt codebase, it seems `timezone` module is only used in `salt.util.schedule` and `salt.states.timezone` 

Fixes: #58379

### Previous Behavior
Minion fails hard with `[salt.minion      :1050][CRITICAL][972] Unexpected error while connecting to <master>` if timezone cannot be determined.

### New Behavior
Minions starts up and connects to master, although using default offset. Note that `win_timezone.*` modules can still fail with `pytz.exceptions.UnknownTimeZoneError` error and `win_timezone.get_timezone` still returns `Unknown`.

- [ ] Docs
- [x] Tests written/updated

### Commits signed with GPG?
Yes

### DIFF
--- a/salt/utils/schedule.py
+++ b/salt/utils/schedule.py
@@ -160,7 +160,18 @@ class Schedule:
                 self.returners = returners
             else:
                 self.returners = returners.loader.gen_functions()
-        self.time_offset = self.functions.get("timezone.get_offset", lambda: "0000")()
+        try:
+            self.time_offset = self.functions.get(
+                "timezone.get_offset", lambda: "0000"
+            )()
+        except Exception:  # pylint: disable=W0703
+            # get_offset can fail, if that happens, default to 0000
+            log.warning(
+                "Unable to obtain correct timezone offset, defaulting to 0000",
+                exc_info_on_loglevel=logging.DEBUG,
+            )
+            self.time_offset = "0000"
+
         self.schedule_returner = self.option("schedule_returner")
         # Keep track of the lowest loop interval needed in this variable
         self.loop_interval = six.MAXSIZE
--- a/changelog/58379.fixed
+++ b/changelog/58379.fixed
@@ -1,2 +1,3 @@
 Fixed issue with win_timezone when dst is turned off. This was causing the
 minion not to start
+Use default timezone offset in scheduler when correct timezone cannot be determined
--- a/tests/pytests/unit/utils/test_schedule.py
+++ b/tests/pytests/unit/utils/test_schedule.py
@@ -1,0 +1,57 @@
+import pytest
+import salt.utils.schedule
+from tests.support.mock import MagicMock
+
+try:
+    import pytz.exceptions
+
+    HAS_LIBS = True
+except ImportError:
+    HAS_LIBS = False
+
+
+def test_correct_tz_offset():
+    kwargs = {
+        "opts": {},
+        "functions": {"timezone.get_offset": lambda: "0200"},
+        "cleanup": False,
+        "standalone": True,
+        "new_instance": True,
+        "utils": {"k": "v"},
+    }
+
+    s = salt.utils.schedule.Schedule(**kwargs)
+    assert s.time_offset == "0200"
+
+
+def test_default_tz_offset():
+    kwargs = {
+        "opts": {},
+        "functions": {},
+        "cleanup": False,
+        "standalone": True,
+        "new_instance": True,
+        "utils": {"k": "v"},
+    }
+
+    s = salt.utils.schedule.Schedule(**kwargs)
+    assert s.time_offset == "0000"
+
+
+@pytest.mark.skipif(not HAS_LIBS, reason="pytz is not installed")
+def test_default_tz_offset_exc():
+    kwargs = {
+        "opts": {},
+        "functions": {
+            "timezone.get_offset": MagicMock(
+                side_effect=pytz.exceptions.UnknownTimeZoneError("Unknown")
+            )
+        },
+        "cleanup": False,
+        "standalone": True,
+        "new_instance": True,
+        "utils": {"k": "v"},
+    }
+
+    s = salt.utils.schedule.Schedule(**kwargs)
+    assert s.time_offset == "0000"


cmcmarrow (dismissed):

lukasraska:
@dwoz / @waynew any ideas what could be tested here? I assumed the unit test would be enough to ensure the constructor doesn't fail hard on exception.

Is there also any reason why this can't be merged into Magnesium when it's high severity bug? Thanks

lukasraska:
Rebased and conflicts fixed. Anything else needs to be done here since it has `Needs Testcase` flag, @dwoz? Thanks

krionbsd:
re-run fedora32

twangboy (dismissed):

Ch3LL (dismissed):

Ch3LL:
@twangboy looks like your changelog entry in this PR https://github.com/saltstack/salt/pull/59341 is conflicting with this one as well. Seems both PRs are fixing two different issues:
1. @twangboy 's PR is resolving the last comment in the issue
2. This PR resolves the original reported issue.
I think what we can do is just combine the descriptions into one changelog entry. Am i understanding this conflict correclty?

garethgreenaway (dismissed):

sagetherage:
@twangboy can you combine the changelogs here pretty please?

salt/utils/schedule.py
```diff
@@ -160,7 +160,18 @@ def __singleton_init__(
                 self.returners = returners
             else:
                 self.returners = returners.loader.gen_functions()
-        self.time_offset = self.functions.get("timezone.get_offset", lambda: "0000")()
+        try:
+            self.time_offset = self.functions.get(
+                "timezone.get_offset", lambda: "0000"
+            )()
+        except Exception:  # pylint: disable=W0703
```
Ch3LL, line 167: Do we know what Exception is raised here? Its usually better to be explicit about the exception we are catching.

lukasraska: In the linked issue it was `pytz.exceptions.UnknownTimeZoneError`. But since scheduler is using `timezone.get_offset` (and if there is exception, minion will fail on start), it should be possible to get any exception here in theory (such as with wrong custom module)

From the core modules it could be:
 - pytz.exceptions.* from win_timezone
 - CommandExecutionError from mac_timezone
 - CommandExecutionError from timezone (or potentially lots of others, depending on the state of OS)

So I thought it would be good idea to ensure minion can start and function if there is any immediate timezone-related problem (although it violates W0703 rule). The second problem with catching pytz.exceptions.* directly might be whether it's really available on all supported platforms by default. It's hard dependency on Windows and it's in static deps for some other platforms, but it's not directly used outside win_timezone.

But I can change it to specific list of exceptions that might be relevant, if it's necessary.

Ch3LL: Thanks for the explanation. I'm good with your approach. :)

Ch3LL (approved):

twangboy (approved):

Issue #38990, opened 2017-01-27
Regarding Proxy process memory and CPU usage


mirceaulinic:
### Description of Issue/Question

I'm afraid the proxy process consumes more than expected (I recall @cro said they would need ~40M of memory), but this is what I constantly see (in the `top`):

```yaml
  PID USER      PR  NI    VIRT    RES    SHR S  %CPU %MEM     TIME+ COMMAND
20134 root      20   0 1189492  85628  13764 S   0.3  0.1   2:03.23 salt-proxy
20224 root      20   0 1119728  89104  14256 S   0.3  0.1   2:15.44 salt-proxy
20241 root      20   0 1261148 103232  14044 S   0.3  0.1   2:25.58 salt-proxy
20413 root      20   0 1203788 111008  13676 S   0.3  0.1   2:36.26 salt-proxy
20419 root      20   0 1126996  83564  14108 S   0.3  0.1   2:10.13 salt-proxy
20434 root      20   0 1475828 142124  13676 S   0.3  0.1   4:21.18 salt-proxy
20479 root      20   0 1258680  92728  13980 S   0.3  0.1   2:13.64 salt-proxy
20480 root      20   0 1127228  85156  13812 S   0.3  0.1   2:03.96 salt-proxy
20523 root      20   0 1393708  95884  13932 S   0.3  0.1   2:48.58 salt-proxy
```

When doing a `sync_all`, the CPU usage spikes:

```yaml
  PID USER      PR  NI    VIRT    RES    SHR S  %CPU %MEM     TIME+ COMMAND
20704 root      20   0 1275184 102724  13848 S 108.3  0.1   4:29.43 salt-proxy
20950 root      20   0 1602884 216072  13888 S 105.7  0.2   8:01.68 salt-proxy
21039 root      20   0 1198564 149368  13956 S 103.7  0.1   9:59.86 salt-proxy
20954 root      20   0 1421144 132168  13968 S 103.1  0.1   5:53.28 salt-proxy
21044 root      20   0 1532552 169760  14052 S 103.1  0.1   7:46.55 salt-proxy
21648 root      20   0 1635956 154768  13924 S 103.1  0.1   7:05.68 salt-proxy
20643 root      20   0 1418324 154004  13812 S 102.7  0.1   5:20.54 salt-proxy
```

And the master:

```yaml
  PID USER      PR  NI    VIRT    RES    SHR S  %CPU %MEM     TIME+ COMMAND
[...]

PR #31218, opened 2016-02-15
Allow loading external modules using setuptools entry points

s0undt3ch:
Since I was unable to find the right document to properly document this addition, here's what should be added to said document.
# External Modules Setuptools Entry-Points Support

The salt loader was enhanced to look for external modules by looking at the `salt.loader` [entry-point](https://pythonhosted.org/setuptools/setuptools.html#dynamic-discovery-of-services-and-plugins).
If [`pkg_resources`](https://pythonhosted.org/setuptools/pkg_resources.html) is available(which it should if setuptools is installed).

The package which has custom engines, minion modules, outputters, etc, should require setuptools and should define the following entry points in it's setup function:

``` python
from setuptools import setup, find_packages

setup(name=<NAME>,
      version=<VERSION>,
      description=<DESC>,
      author=<AUTHOR>,
      author_email=<AUTHOR-EMAIL>,
      url=' ... ',
      packages=find_packages(),
      entry_points='''
        [salt.loader]
        engines_dirs = <package>.<loader-module>:engines_dirs
        fileserver_dirs = <package>.<loader-module>:fileserver_dirs
        pillar_dirs = <package>.<loader-module>:pillar_dirs
        returner_dirs = <package>.<loader-module>:returner_dirs
        roster_dirs = <package>.<loader-module>:roster_dirs
      ''')
```

The above setup script example mentions a loader module, here's an example of how `<package>/<loader-module>.py` should look like:

``` python
# -*- coding: utf-8 -*-

# Import python libs
import os

PKG_DIR = os.path.abspath(os.path.dirname(__file__))


def engines_dirs():
    '''
    yield one path per parent directory of where engines can be found
    '''
    yield os.path.join(PKG_DIR, 'engines_1')
    yield os.path.join(PKG_DIR, 'engines_2')


def fileserver_dirs():
    '''
    yield one path per parent directory of where fileserver modules can be found
    '''
    yield os.path.join(PKG_DIR, 'fileserver')


def pillar_dirs():
    '''
    yield one path per parent directory of where external pillar modules can be found
    '''
    yield os.path.join(PKG_DIR, 'pillar')


def returner_dirs():
    '''
    yield one path per parent directory of where returner modules can be found
    '''
    yield os.path.join(PKG_DIR, 'returners')


def roster_dirs():
    '''
    yield one path per parent directory of where roster modules can be found
    '''
    yield os.path.join(PKG_DIR, 'roster')
```

---

Refs https://github.com/saltstack/salt/issues/31206

PR #58393, opened 2020-09-06
Added Nornir based proxy-minion module.

dmulyalin:
## What does this PR do?

Added new SALT proxy-minion module to manage network devices using [Nornir ](https://nornir.readthedocs.io/en/latest/index.html) library

Added new execution module to interact with Nornir salt-proxy.

- [x] Docs
- [ ] Tests written/updated

### Commits signed with GPG?
No

## Motivation

Primary reason for developing Nornir proxy is scaling issues. 

Not a secret that normally for each network device we want to manage with SALT, there is a dedicated proxy-minion process must be spawned and configured. Each such a process consumes from 40 to 70 MByte or RAM. Meaning, to managed 1000 network devices, 40-70 Gbyte of RAM required. That is a big amount of resources to spent. Problem aggravated by the fact that SSH based proxy-minions have to run with multiprocessing mode off, resulting in use of threading to run tasks, but threading prone to memory leaks [issue ](https://github.com/saltstack/salt/issues/38990) and the cure is to periodically restart processes, raising operational costs of overall system.

## Design

[Nornir ](https://nornir.readthedocs.io/en/latest/index.html) is an open-source automation software that uses same libraries under the hood as some other Network proxy-minion - NAPALM, Netmiko, Paramiko and NCClient. The key differentiator of Norninr is that it uses Python to express task execution workflows and runs them in parallel using threading. 

Wrapping Norninr in salt-proxy allowing us to use SALT systems such as schedulers, reactors, returners, mines, pillars, API etc. together with Nornir's capability to effectively run jobs for multiple devices. As a result, single proxy process can deliver configuration or retrieve state from multiple devices simultaneously. Moreover, due to Nornir stateless nature, proxy can and should operate with multiprocessing set to True.

To illustrate, say we want to manage same 1000 devices, but with Nornir proxy. For that we start 10 Nornir processes proxying for 100 devices each, say each process consumes 100 MByte of RAM, resulting in 1 GByte of RAM to run 10 process. Every time we want to run task, new process started, consuming another 100 MByte or RAM. Assume extreme case, we have 4 tasks running in parallel to 1000 devices with process dedicated to each task, that would consume 5 GByte of RAM, once tasks completed and processes destroyed, RAM consumption will fall back to 1 GByte. That is, in author's opinion, quite an improvement compared to above example where we used 40-70 GByte or RAM to manage same number of devices.

## Drawbacks

- Double targeting required to narrow down tasks execution to a subset of hosts
- No long running connections exists to devices, each new task creates dedicated connections to devices, increasing overall execution time
- Load on AAA system (Tacacs, Radius) will increase as more tasks will result in more authentication requests from devices
- In addition to knowing how pillar works, one will have to know how [Nornir inventory](https://nornir.readthedocs.io/en/3.0.0/tutorial/inventory.html) structured to use it effectively, as Nornir inventory integrated in proxy-minion pillar


dmulyalin:
Does not see much appetite for this PR from dev Team. 

Published Nornir based modules as standalone packages on PyPi so that they can be installed via pip and SALT can detect and run them using entry points as per this PR - https://github.com/saltstack/salt/pull/31218

SALT-NORNIR modules - https://pypi.org/project/salt-nornir/

Above modules ported to Nornir 3.x.x and does not work with 2.5.0

Also created a collection of functions/runners/inventory etc. plugins that can be run standalone with Nornir - https://pypi.org/project/nornir-salt/ - and used by SALT-NORNIR proxy.


Leaving this PR open just case, but stop working on it at least for now.

network-shark:
We need this in master :)

dmulyalin:
Not needed, [Salt-Nornir](https://github.com/dmulyalin/salt-nornir) developed as a standalone plugin.

PR #58405, opened 2020-09-09
Fix name returned by list_vdirs

### LINKED ISSUE #58404: [BUG] win_iis state + module for vdirs doesn't work for nested names?
**Description**
`list_vdirs` from the win_iis module seems to fetch virtual directories using the powershell 'Get-WebVirtualDirectory' command. Cool. [Then it splits the name and fetches the last part of the path](https://github.com/saltstack/salt/blob/master/salt/modules/win_iis.py#L1722). 
This means any virtual directory that is has a name of (for example) Content/Packages/Archive, will come back as 'Archive'

Subsequently, the [matching state won't work](https://github.com/saltstack/salt/blob/master/salt/states/win_iis.py#L701), because it keeps trying to 'clean up' the name.

**Setup**
Standard salt setup, applying the following .sls as a state:
```sls
ensure the nested vdir exists:
  win_iis.create_vdir:
    - name: Content/Packages/Archive
    - site: Default Web Site
    - sourcepath: 'C:\tmp'
```

Apply the state twice. It won't work because the name that comes back will never match the name you put in.

Timbus:
Fixes: #58404

### Previous Behavior
list_vdirs returned the last part of the path in the vdir name.

### New Behavior
list_vdirs returns the entire path, minus the initial forward-slash.
This shouldn't affect people using it since there's no way it ever worked for nested paths. If there is some hidden reason for this trim command removing nested paths, then it needs documentation or something, or an IIS expert needs to step in and tell me what up.

- [ ] Docs
- [ ] Tests written/updated

### Commits signed with GPG?
No

### DIFF
--- a/salt/modules/win_iis.py
+++ b/salt/modules/win_iis.py
@@ -1719,7 +1719,7 @@ def list_vdirs(site, app=_DEFAULT_APP):
         r"'{0}'".format(app),
         "|",
         "Select-Object PhysicalPath, @{ Name = 'name';",
-        r"Expression = { $_.path.Split('/')[-1] } }",
+        r"Expression = { $_.path.Trim('/') } }",
     ]
 
     cmd_ret = _srvmgr(cmd=ps_cmd, return_json=True)
--- a/salt/modules/win_iis.py
+++ b/salt/modules/win_iis.py
@@ -8,17 +8,14 @@ Microsoft IIS site management via WebAdministration powershell module
 
 .. versionadded:: 2016.3.0
 """
-# Import python libs
-
 import decimal
 import logging
 import os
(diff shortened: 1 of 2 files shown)


twangboy:
@Timbus Could we please get some tests for this.

Timbus:
Oh yuck, I copied the existing tests but.. It's mostly just a test of your mocks system. 
Unless there are tests to actually run the powershell it's all just fake =/

twangboy (dismissed):

twangboy (dismissed):

twangboy (approved):

dwoz (approved):

PR #58423, opened 2020-09-10
Fix malformed shadow entries

terminalmage:
If a shadow entry for a user has too many (or too few) fields, due to
having been hand-edited for some silly reason, this will cause Salt not
to be able to successfully run a user.present state (with password
argument) against it, as spwd will not be able to retrieve a hash for
the user, causing Salt to always think that the password hash still
needs changes.

This fixes these two cases.


Ch3LL:
Do you mind fixing the merge conflicts

Ch3LL (approved):

PR #58451, opened 2020-09-14
Add verify_ssl to file.managed state

### LINKED ISSUE #52663: Feature Request: file.managed verify_ssl option.
Currently, there is not a way of disabling the verification of SSL in `https` sources. This could lead to code breakages while trying to install certs. Or even during the first run before a system is setup fully if using `https` sources 

```
example state:
  file.managed:
    - name: /tmp/file
    - source: https://badcert.url/good/file.crt
    - verify_ssl: False
```

Ch3LL:
### What does this PR do?
adds new option to file modules - `verify_ssl`. This  adds the ability to skip validation check of an SSL certificate for https sources and source_hash.

Fixes: https://github.com/saltstack/salt/issues/52663

### Previous Behavior
Could not skip ssl certification validation check

### New Behavior
can use new option `verify_ssl` for the file states/execution modules to skip SSL certificate check.

- [x] Docs
- [x] Tests written/updated

### Commits signed with GPG?
Yes


DmitryKuzmenko (dismissed):

dwoz (approved):

tests/pytests/integration/modules/test_file.py
```diff
@@ -0,0 +1,162 @@
...
+        "",
+        "",
+        "",
+        "",
+        "base",
+        "{}",
+        "",
+        "True",
+        "verify_ssl={}".format(verify_ssl),
+    )
```
s0undt3ch, line 60: FYI,
If you pass these as keyword arguments to `run()` you would likely not need all of the `"",` in the call.
```
ret = salt_call_cli.run(
    "--local",
    "file.get_managed",
    tmpdir.join("test_managed.txt").strpath,
   saltenv="base"
```
Bah!
Nevermind, it's all arguments to the function, not keyword arguments.
_Bad code, bad bad code...._
Anyway, you can still do ` .run(... args..., verify_ssl=True)` , no need to pass `verify_ssl=True` as a string argument.

PR #58461, opened 2020-09-15
[PyTest Migration] Skip tests which randomly fail under Py3.5

s0undt3ch:
### What does this PR do?
Skip tests which randomly fail under Py3.5
There's no point in fixing this since Py3.5 is EOL and support will be dropped soon.

### DIFF
--- a/tests/pytests/unit/utils/test_data.py
+++ b/tests/pytests/unit/utils/test_data.py
@@ -1,3 +1,6 @@
+import sys
+
+import pytest
 import salt.utils.data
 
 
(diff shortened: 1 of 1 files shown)


cbosdo (approved):
@s0undt3ch These tests are working pretty well with 3.6+. I think it's is indeed a good idea to disable them for 3.5

krionbsd (approved):

s0undt3ch:
The windows failures are being addressed in https://github.com/saltstack/salt/pull/58464

PR #58472, opened 2020-09-16
Implement file protocol subsystem

gtmanfred:
### What does this PR do?
Implement a different fileproto handlers for the fileclient

We have some basic ones, but there are other possible ones that people might want to implement, like rsync, and this will make it easier to touch those without having to fiddle with everything else in the fileclient.

### New Behavior
There should be no change of behavior.

- [ ] Docs
- [ ] Tests written/updated

### Commits signed with GPG?
Yes


gtmanfred:
Nice, that is fewer errors than i expected.

gtmanfred:
Why does this need other test cases? shouldn't the regular salt test suite handle this? It would signify that the current file protocols continue to work.

dwoz:

